PR #7088, opened 2025-04-16
Feat: Support passing knowledge base id as variable in retrieval component

### LINKED ISSUE #6600: [Feature Request]: Support knowledge base ids as params for retrieval component in agent flow
### Self Checks

```Markdown
I'm always frustrated that when I add a agent for different customers, they all have their own knowledge bases. Because the retrieval componnent is bundled to fixed knowledge base, I have to copy a same agent for every customer to use the same agent handling same business.
```

The retrieval component supporting a param of kb_ids in agent flow, So I can use a single agent for different customers of different knowledge bases. 

_No response_

```Markdown

```

### Additional information

_No response_

notsyncing:

Fix #6600

Hello, I have the same business requirement as #6600. My use case is: 

We have many departments (> 20 now and increasing), and each department has its own knowledge base. Because the agent workflow is the same, so I want to change the knowledge base on the fly, instead of creating agents for every department.

It now looks like this:

屏幕截图_20250416_212622

Knowledge bases can be selected from the dropdown, and passed through the variables in the table. All selected knowledge bases are used for retrieval.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/agent/component/base.py
+++ b/agent/component/base.py
@@ -19,7 +19,7 @@ import json
 import os
 import logging
 from functools import partial
-from typing import Tuple, Union
+from typing import Any, Tuple, Union
 
 import pandas as pd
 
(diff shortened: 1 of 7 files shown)


asiroliu:
@KevinHuSh 
Successfully validated all Agent Templates in this version - all templates are functioning normally with proper conversational responses.

wsxiaotiana:
@notsyncing @KevinHuSh  Hi, I really appreciate your team's hard work. But this feature hasn't been merged after two weeks of validation. Could you share the timeline? Thanks a lot.

asiroliu:
@notsyncing 
I've performed the following validation steps using your latest commit
1. Successfully built Docker image from latest commit (743751d549bf93549fcce06282a0161f879c1f2e)
2. Created test knowledge bases:
- Cash Management
- The Chu Shi Biao
image

3. Initialized simple agent 
image

4. Retrieved KB ID for "The Chu Shi Biao": 7fd6509024ad11f0ae5f1e318b07bb72
image

5. Test question about `Cash Management`
image

6. Test question about `The Chu Shi Biao`
image

asiroliu:
@notsyncing 
Actually, KB names are already unique in our system ([db_models.py#L591](https://github.com/infiniflow/ragflow/blob/main/api/db/db_models.py#L591))
image

Could we consider using KB names directly instead?

notsyncing:
> Could we consider using KB names directly instead?

In our use case, we are passing KB ids from another program, so we don't need KB names. But I think it's better to add a `knowledge base` type input to the `Begin` component, and reuse the existing knowledge base selector for that type, so we can simply select KBs from the debug input. What do you think about this?

notsyncing:
> Actually, KB names are already unique in our system ([db_models.py#L591](https://github.com/infiniflow/ragflow/blob/main/api/db/db_models.py#L591))

Also, the `index=True` does not seem to mean `unique=True`, according to `peewee` document https://docs.peewee-orm.com/en/latest/peewee/models.html

asiroliu:
You're right. The KB name uniqueness isn't enforced by index=True here—it's maintained by code like [dataset.py#L139](https://github.com/infiniflow/ragflow/blob/main/api/apps/sdk/dataset.py#L139)
image

notsyncing:
> You're right. The KB name uniqueness isn't enforced by index=True here—it's maintained by code like [dataset.py#L139]

Okay. Even you have enforced this, I still advice using `id` instead of `name` here:
1. When using ragflow as a backend service (like our use case), we usually prefer an `id` instead of a `name`, because `name` may change, or may cause some encoding issues with some non-UTF8 client.
2. When providing ragflow's frontend to user, we can use a selector (like the kb selector in the `Retrieval` component) for this, to keep a user-friendly behavior, because user may enter a wrong name and was not aware of it until an error occurs. The selector still gives `id` underneath.

Issue #7095, opened 2025-04-17
[Bug]: Potential Data Loss: Empty kb_id allows deletion of all knowledgebases


lianwenke:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

5d253e0a

### RAGFlow image version

nightly

### Other environment information

```Markdown
all
```

### Actual behavior

Currently, the /rm endpoint contains a serious logic flaw:

- If the kb_id in the request is an empty string (""), the @validate_request("kb_id") decorator does not reject the request, since the value is not None.

- As a result, the query KnowledgebaseService.query(created_by=current_user.id, id="") may return all knowledgebases created by the current user, and the loop proceeds to delete them all.

This can lead to accidental deletion of all knowledgebases, resulting in serious data loss.

### Expected behavior

Either update the validate_request decorator to treat empty strings as invalid input;

Or add an explicit check in the /rm route to reject kb_id == "" and return an appropriate error.

### Steps to reproduce

```Markdown
Call the knowledge base to delete the interface, and kb_id == ""
```

### Additional information

def validate_request(*args, **kwargs):
    def wrapper(func):
        @wraps(func)
        def decorated_function(*_args, **_kwargs):
            input_arguments = flask_request.json or flask_request.form.to_dict()
            no_arguments = []
            error_arguments = []
            for arg in args:
                if arg not in input_arguments:
                    no_arguments.append(arg)
            for k, v in kwargs.items():
                config_value = input_arguments.get(k, None)
                if config_value is None:
                    no_arguments.append(k)
                elif isinstance(v, (tuple, list)):
                    if config_value not in v:
                        error_arguments.append((k, set(v)))
                elif config_value != v:
                    error_arguments.append((k, v))
            if no_arguments or error_arguments:
                error_string = ""
                if no_arguments:
                    error_string += "required argument are missing: {}; ".format(",".join(no_arguments))
                if error_arguments:
[...]

PR #7099, opened 2025-04-17
Fix potential mass deletion when kb_id is empty in /rm endpoint

lianwenke:
### What problem does this PR solve? #7095 

- This PR addresses a critical bug where passing an empty kb_id ("") to the /rm endpoint could result in deletion of all knowledgebases owned by the current user.

- Previously, the validate_request decorator did not reject an empty string, allowing the request to proceed.

- As a result, the deletion logic would run KnowledgebaseService.query(created_by=current_user.id, id=""), which could unintentionally match and delete all entries.

### Type of change

- [ √] Bug Fix (non-breaking change which fixes an issue)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/api/utils/api_utils.py
+++ b/api/utils/api_utils.py
@@ -147,25 +147,33 @@ def validate_request(*args, **kwargs):
             input_arguments = flask_request.json or flask_request.form.to_dict()
             no_arguments = []
             error_arguments = []
+
             for arg in args:
-                if arg not in input_arguments:
+                value = input_arguments.get(arg, None)
+                if value is None or (isinstance(value, str) and not value.strip()):
                     no_arguments.append(arg)
+
             for k, v in kwargs.items():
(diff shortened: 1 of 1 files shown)


KevinHuSh:
CI failure. Do you have any clue what happens?

lianwenke:
The decorator function @validate_request(...) is validating request inputs, and it specifically checks for required arguments. Your test setup is triggering the following error repeatedly:
```
Exception: required argument are missing or empty: asr_id;
```
This is because asr_id is expected in the request payload (request.json or request.form), but it's either:
- not provided,
- or is an empty string.
Check this parameter is necessary
image

KevinHuSh:
Appreciations!
After our evaluation, this feature is not going to be merged into main branch.

PR #7106, opened 2025-04-17
fix(nursery): Fix Closure Trap Issues in Trio Concurrent Tasks

aniaan:
## Problem Description
Multiple files in the RAGFlow project contain closure trap issues when using lambda functions with `trio.open_nursery()`. This problem causes concurrent tasks created in loops to reference the same variable, resulting in all tasks processing the same data (the data from the last iteration) rather than each task processing its corresponding data from the loop.

## Issue Details
When using a `lambda` to create a closure function and passing it to `nursery.start_soon()` within a loop, the lambda function captures a reference to the loop variable rather than its value. For example:

```python
# Problematic code
async with trio.open_nursery() as nursery:
    for d in docs:
        nursery.start_soon(lambda: doc_keyword_extraction(chat_mdl, d, topn))
```

In this pattern, when concurrent tasks begin execution, `d` has already become the value after the loop ends (typically the last element), causing all tasks to use the same data.

## Fix Solution
Changed the way concurrent tasks are created with `nursery.start_soon()` by leveraging Trio's API design to directly pass the function and its arguments separately:

```python
# Fixed code
async with trio.open_nursery() as nursery:
    for d in docs:
        nursery.start_soon(doc_keyword_extraction, chat_mdl, d, topn)
```

This way, each task uses the parameter values at the time of the function call, rather than references captured through closures.

## Fixed Files
Fixed closure traps in the following files:

1. `rag/svr/task_executor.py`: 3 fixes, involving document keyword extraction, question generation, and tag processing
2. `rag/raptor.py`: 1 fix, involving document summarization
3. `graphrag/utils.py`: 2 fixes, involving graph node and edge processing
4. `graphrag/entity_resolution.py`: 2 fixes, involving entity resolution and graph node merging
5. `graphrag/general/mind_map_extractor.py`: 2 fixes, involving document processing
6. `graphrag/general/extractor.py`: 3 fixes, involving content processing and graph node/edge merging
7. `graphrag/general/community_reports_extractor.py`: 1 fix, involving community report extraction

## Potential Impact
This fix resolves a serious concurrency issue that could have caused:
- Data processing errors (processing duplicate data)
- Performance degradation (all tasks working on the same data)
- Inconsistent results (some data not being processed)

After the fix, all concurrent tasks should correctly process their respective data, improving system correctness and reliability.

### DIFF
--- a/graphrag/entity_resolution.py
+++ b/graphrag/entity_resolution.py
@@ -103,7 +103,7 @@ class EntityResolution(Extractor):
                     continue
                 for i in range(0, len(candidate_resolution_i[1]), resolution_batch_size):
                     candidate_batch = candidate_resolution_i[0], candidate_resolution_i[1][i:i + resolution_batch_size]
-                    nursery.start_soon(lambda: self._resolve_candidate(candidate_batch, resolution_result))
+                    nursery.start_soon(self._resolve_candidate, candidate_batch, resolution_result)
         callback(msg=f"Resolved {num_candidates} candidate pairs, {len(resolution_result)} of them are selected to merge.")
 
         change = GraphChange()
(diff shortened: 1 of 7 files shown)


asiroliu:
@KevinHuSh @aniaan 
Successfully tested file parsing with ​​RAPTOR​​ and ​​Knowledge Graph​​ enabled in this version - all files processed normally.

Issue #6813, opened 2025-04-04
[Question]: Redis many  old tasks


isthaison:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

![Image](https://github.com/user-attachments/assets/8b2c3607-7277-47d7-b244-c6fb6c1eadc0)

![Image](https://github.com/user-attachments/assets/1226e923-dcbf-4dd7-ade8-73158f3bddbd)

PR #7073, opened 2025-04-16
feat: Recover pending tasks while pod restart.

liuzhenghua:
### What problem does this PR solve?

If you deploy Ragflow using Kubernetes, the hostname will change during a rolling update. This causes the consumer name of the task executor to change, making it impossible to schedule tasks that were previously in a pending state.
To address this, I introduced a recovery task that scans these pending messages and re-publishes them, allowing the tasks to continue being processed.

### Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

PR #7107, opened 2025-04-17
Allow delete task resdis

isthaison:
Allow delete task resdis on UI
Relate: https://github.com/infiniflow/ragflow/issues/6813

### DIFF
--- a/api/apps/system_app.py
+++ b/api/apps/system_app.py
@@ -319,3 +319,53 @@ def get_config():
     return get_json_result(data={
         "registerEnabled": settings.REGISTER_ENABLED
     })
+
+
+@manager.route("/task_executor/<executor_id>", methods=["DELETE"])  # noqa: F821
+@login_required
+def delete_task_executor(executor_id):
+    """
+    Remove a task executor from the system.
+    ---
+    tags:
+      - System
+    security:
+      - ApiKeyAuth: []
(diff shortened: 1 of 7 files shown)


KevinHuSh:
Why do we need to delete tasks?
Could you elabrate on the scenario?

isthaison:
The task name after each image docker version rebuild server will be changed, this creates many different task names. and old tasks are not processed anymore, moreover in the schedule the task name is changed many times. creating a list of many tasks even though they are not used. through the image in the problem as I took, my server started through many versions of ragflow

isthaison:
Initial task name is consumer_x -> executor_x -> executor_x_guid

wanpdsantos:
I believe @isthaison is referring to the Task Executor list in UserSettings > System > Task Executor tab. I agree this is not good, but manually deleting these doesn't seem feasible to me. If you consider running in a K8S cluster where you scale the task executor pods often, this list will be big and will keep growing.

I would say it would be better to perform an auto-clean of these old task executors if they are not sending a heartbeat for more than X seconds, claim the tasks in these task executors back to the queue, then remove them from the consumer group.

yuzhichang:
#7073 re-publishes tasks if the owned executor doesn't report heartbeat for 2 minutes.

PR #7151, opened 2025-04-20
Fix: whole knowledge graph lost after removing any document in the knowledge base

alkscr:

When you removed any document in a knowledge base using knowledge graph, the graph's `removed_kwd` is set to "Y".  
However, in the function `graphrag.utils.get_gaph`, `rebuild_graph` method is passed and directly return `None` while `removed_kwd=Y`, making residual part of the graph abandoned (but old entity data still exist in db).

Besides, infinity instance actually pass deleting graph components' `source_id` when removing document. It may cause wrong graph after rebuild.

### Type of change

### DIFF
--- a/graphrag/utils.py
+++ b/graphrag/utils.py
@@ -414,16 +414,15 @@ async def get_graph(tenant_id, kb_id):
         "knowledge_graph_kwd": ["graph"]
     }
     res = await trio.to_thread.run_sync(lambda: settings.retrievaler.search(conds, search.index_name(tenant_id), [kb_id]))
-    if res.total == 0:
-        return None
-    for id in res.ids:
-        try:
-            g = json_graph.node_link_graph(json.loads(res.field[id]["content_with_weight"]), edges="edges")
-            if "source_id" not in g.graph:
-                g.graph["source_id"] = res.field[id]["source_id"]
-            return g
-        except Exception:
-            continue
+    if not res.total == 0:
+        for id in res.ids:
+            try:
+                g = json_graph.node_link_graph(json.loads(res.field[id]["content_with_weight"]), edges="edges")
+                if "source_id" not in g.graph:
+                    g.graph["source_id"] = res.field[id]["source_id"]
+                return g
+            except Exception:
+                continue
     result = await rebuild_graph(tenant_id, kb_id)
     return result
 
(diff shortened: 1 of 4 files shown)


rag/utils/infinity_conn.py
```diff
@@ -554,13 +572,40 @@ def update(
                 assert isinstance(v, list)
                 newValue[k] = "_".join(f"{num:08x}" for num in v)
             elif k == "remove":
-                del newValue[k]
-                if v in [PAGERANK_FLD]:
-                    newValue[v] = 0
+                if isinstance(v, str):
+                    ty, de = clmns[v]
```
yuzhichang, line 577: `clmn` could be uninitialized.

alkscr: Thanks. I've add a check in the new commit

yingfeng:
This PR will lead to failure of CI, see 
https://github.com/infiniflow/ragflow/actions/runs/14568237150/job/40865284748?pr=7151

Make sure the tests can pass before this PR can be merged, thank you~

alkscr:
> This PR will lead to failure of CI, see https://github.com/infiniflow/ragflow/actions/runs/14568237150/job/40865284748?pr=7151
> 
> Make sure the tests can pass before this PR can be merged, thank you~

I tried these tests locally and found a error report of closing a infinity connection. It only happened when running `test_frontend_api/test_dataset` after all tests in `test_sdk_api` are done.  
In this situation, the test tried remove 30+ knowledgebase with document inside one time, and the initial size of the infinity connection pool is 16. So I guess this problem occur because graph checking cost more time and the program require more connection in runtime. Something goes wrong when closing the extra connects.  
It may be a Infinity issue?

RAGFlow may avoid this problem by increasing pool size to 32 in my test.

yuzhichang:
@alkscr Please change pool size to 32 (at `infinity_conn.py:121`) to make the CI pass.

alkscr:
> @alkscr Please change pool size to 32 (at `infinity_conn.py:121`) to make the CI pass.

Thanks!  
But I found it still possible to fail depend on platform's performance by simply increasing pool size (10 core VM pass, 4core VM failed). Infinity may try to disconnect even not using extra connection. Not sure about its mechanism.  

As a workaround, I let RAGFlow only try to delete graph components when found a graph with specific source.  
It should work this time.

graphrag/utils.py
```diff
@@ -566,16 +565,18 @@ async def rebuild_graph(tenant_id, kb_id):
...
+        # tot = settings.docStoreConn.getTotal(es_res)
+        es_res = settings.docStoreConn.getFields(es_res, flds)
+
+        if len(es_res) == 0:
             break
 
-        es_res = settings.docStoreConn.getFields(es_res, flds)
         for id, d in es_res.items():
-            assert d["knowledge_graph_kwd"] == "relation"
+            assert d["knowledge_graph_kwd"] == "entity"
```
yuzhichang: I think it's better to rebuild the graph from each document's subgraph(generated at graphrag/general/index.py:126, generate_subgraph).

alkscr: But I think it's too costly to rebuild every document just for deleting a few items. Remaining graph should be enough since nodes related to any undelete file won't be remove.  
Maybe better to offer a manual graph rebuild function instead.

yuzhichang: Let me clarify. There's a subgraph in ElasticSearch/Infinity for each document. We can rebuild the graph from those subgraphs.

alkscr: I get it.  
It seems subgraphs don't update since created in the origin code, so we also need an addition step to save modifications in utils.py, set_graph.  
I try it out later.

alkscr:
Updates: 
1. Save newest subgraphs when running `set_graph`
2. Build graph based on existing subgraphs in `rebuild_graph`

yuzhichang (approved):
LGTM

PR #7170, opened 2025-04-21
fix: resolve domain parameter override bug

yxkong:

This PR fixes a critical initialization issue where the `domain` parameter in `PDFParser` was being forcibly overwritten with a hardcoded value `"layout"`, making the parameter effectively non-functional.

**Root Cause Analysis**:
1. Constructor contained direct assignment `domain = "layout"` before parameter usage
2. This occurred *after* parameter declaration but *before* the `self.domain` assignment
3. Resulted in all instances ignoring constructor arguments

**Impact Scope**:
- ✖️ Broken: Custom domain configurations (e.g. `medical`, `legal` specializations)
- ✖️ Broken: Dynamic parser selection based on document type
- ✔️ Fixed: Now properly respects constructor arguments while maintaining "layout" as safe default

**Before/After Example**:

Before (always uses "layout")
`parser = PDFParser(domain="medical")  # IGNORED
`
After (respects argument)
`
parser = PDFParser(domain="medical")  # Works!
`

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/deepdoc/vision/layout_recognizer.py
+++ b/deepdoc/vision/layout_recognizer.py
@@ -181,7 +181,6 @@ class LayoutRecognizer4YOLOv10(LayoutRecognizer):
     ]
 
     def __init__(self, domain):
-        domain = "layout"
         super().__init__(domain)
         self.auto = False
         self.scaleFill = False


KevinHuSh:
Appreciation!
Other models are weaker than the default layout.onnx for the time being.

PR #7187, opened 2025-04-22
refactor: Update Redis configuration to use StatefulSet instead of deployment with pvc

wanpdsantos:

This PR changes Redis to be a statefulset. In some situation when we Redis pod gets rescheduled to another Node, it gets stuck in pending state due to the PVC attached to another Kubernetes node.

### Type of change

- [ ] Documentation Update
- [X] Refactoring
- [ ] Performance Improvement


gnowland:
caused #7403

PR #7222, opened 2025-04-23
feat: Add global exception handler

liuzhenghua:

Add Global Exception Hanlder.

In the request.ts file, for all responses except 413 and 504, the code attempts to parse the body as JSON. If an error occurs during this process, the response in the errorHandler will be null, and the frontend will not display any error messages.
image

For responses where the body is JSON but the HTTP status code is ≥ 400, the umi-request library does not parse the body into the data property of the response. As a result, downstream code may encounter an undefined error when trying to access data.code. In such cases, the entire error object should be returned, as it contains the data property with the value of the response body.
image

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -34,12 +34,14 @@ from flask_login import LoginManager
 from api import settings
 from api.utils.api_utils import server_error_response
 from api.constants import API_VERSION
+from api.error import error_handler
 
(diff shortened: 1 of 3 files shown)


KevinHuSh:
Appreciations!

PR #7325, opened 2025-04-25
Fix:When you create a new API module named xxxa_api, the access route will become xxx instead of xxxa. For example, when I create a new API module named 'data_api', the access route will become 'dat' instead of 'data

donblack01:

Fix:When you create a new API module named xxxa_api, the access route will become xxx instead of xxxa. For example, when I create a new API module named 'data_api', the access route will become 'dat' instead of 'data'
Fix:Fixed the issue where the new knowledge base would not be renamed when there was a knowledge base with the same name

### Type of change


api/apps/__init__.py
```diff
@@ -107,7 +107,10 @@ def search_pages_path(pages_dir):
 def register_page(page_path):
     path = f"{page_path}"
 
-    page_name = page_path.stem.rstrip("_app")
+    page_name = page_path.stem
```
asiroliu:
```suggestion
    page_name = page_path.stem.removesuffix("_app")
```

PR #7330, opened 2025-04-26
fix: single task executor getting all tasks from Redis queue

wanpdsantos:

Currently, as long as there are tasks in Redis, this loop will keep getting the tasks. This will lead to a single task executor with many tasks in the pending state. Then we need to wait for the pending tasks to get them back in the queue.

In first place, if we set the `MAX_CONCURRENT_TASKS` to X, then only X tasks should be picked from the queue, and others should be left in the queue for other `task_executors` or be picked after 1 of the spots in the current executor gets free. This PR ensures this behavior.

The additional changes were due to the Ruff linting in pre-commit. But I believe these are expected to keep the coding style.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/rag/svr/task_executor.py
+++ b/rag/svr/task_executor.py
@@ -100,7 +100,7 @@ CURRENT_TASKS = {}
 MAX_CONCURRENT_TASKS = int(os.environ.get('MAX_CONCURRENT_TASKS', "5"))
 MAX_CONCURRENT_CHUNK_BUILDERS = int(os.environ.get('MAX_CONCURRENT_CHUNK_BUILDERS', "1"))
 MAX_CONCURRENT_MINIO = int(os.environ.get('MAX_CONCURRENT_MINIO', '10'))
-task_limiter = trio.CapacityLimiter(MAX_CONCURRENT_TASKS)
+task_limiter = trio.Semaphore(MAX_CONCURRENT_TASKS)
 chunk_limiter = trio.CapacityLimiter(MAX_CONCURRENT_CHUNK_BUILDERS)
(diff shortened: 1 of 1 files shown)


rag/svr/task_executor.py
```diff
@@ -735,10 +739,11 @@ async def main():
     async with trio.open_nursery() as nursery:
         nursery.start_soon(report_status)
         while not stop_event.is_set():
-            async with task_limiter:
-                nursery.start_soon(handle_task)
+            nursery.start_soon(limited_handle_task)
+            await trio.sleep(1)
```
yuzhichang: Good catch! However your change is smell. It creates an async task every 1 second so there may be very many `limited_handle_task` tasks while only 5 of them are inside `handle_task`.

yuzhichang:
`trio.Semaphore` is better here. Be careful to ensure `handle_task` release semaphore finally.

```
import trio
import random

task_limiter = trio.Semaphore(3)


async def handle_task(id: int):
    try:
        print(f"{trio.current_time()}: Worker {id} started")
        await trio.sleep(random.randint(1, 5))
        print(f"{trio.current_time()}: Worker {id} done")
    finally:
        task_limiter.release()


async def handle_tasks():
    id = 0
    async with trio.open_nursery() as nursery:
        while True:
            await task_limiter.acquire()
            nursery.start_soon(handle_task, id)
            id += 1


trio.run(handle_tasks)
```

yuzhichang:
PR #7700 creates task_manager every 0.1s. This can be fixed by acquiring semaphore before creating task_manager.

Issue #7347, opened 2025-04-27
[Bug]:Http api `dataset_ids` can't be empty


NintendoLink:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

7e1464a95060256d2b3ba98e4f2661b3bec1915c

### RAGFlow image version

7e1464a95060256d2b3ba98e4f2661b3bec1915c

### Other environment information

```Markdown

```

### Actual behavior

Use HTTP API to update chat assistant. When using empty dataset to update chat assistant, the update fails with an error: `dataset_ids` can't be empty.

<img width="685" alt="Image" src="https://github.com/user-attachments/assets/e972af7e-cb14-4c79-93a7-c5e370133601" />

### Steps to reproduce

```Markdown
1、Create a chat assistant using HTTP API
curl --request POST \
     --url http://{endpoint}/api/v1/chats \
     --header 'Content-Type: application/json' \
     --header 'Authorization: Bearer ragflow-api-key' \
     --data '{
              "name":"api-chat-ass",
              "dataset_ids":["some-dataset-id"]
}' \
| jq '.data.id'


2、Update chat assistant with an empty dataset (the value of dataset_ids is [])
curl --request PUT \
     --url http://{endpoint}/api/v1/chats/822e74c4232e11f08dec0242c0a89006 \
     --header 'Content-Type: application/json' \
     --header 'Authorization: Bearer ragflow-api-key' \
     --data '
     {
         "name":"updated-api-chat-ass",
         "dataset_ids":[]
     }' \
| jq
```

### Additional information

If a chat assistant has attached the relevant knowledge base, can't it empty the knowledge base through the HTTP API?

PR #7354, opened 2025-04-27
Fix:Update chat assistant with an empty dataset

NintendoLink:

When updating a chat assistant using API,if the dataset attached by the current chat assistant is not empty，setting dataset to null("dataset_ids":[]) will cause update failure：'dataset_ids' can't be empty

### Type of change


NintendoLink:
#7347

PR #7359, opened 2025-04-27
Docs: Initial editorial pass to MCP server

writinwaters:
### Type of change

- [x] Documentation Update


yiminghub2024 (commented):
- --mcp-host-api-key="ragflow-xxxxxxx" # required only when `mode` is set to `self-host`
 api key need with no "" ,or it is not work!

yiminghub2024:
without follow , upgrade can not start mcp server service in docker
### For those upgrading from versions before v0.18.0

1. Get all MCP related files ready.
   1. copy `mcp/` directory to local.
   1. copy `docker/docker-compose.yml` to local.
   1. copy `docker/entrypoint.sh` to local.
   1. resolve necessary dependencies via `uv`.
      - simply run `uv add mcp` if it works for you. Or:
      - copy `pyproject.toml` and run `uv sync --python 3.10 --all-extras`.
1. Change `docker-compose.yml` to enable MCP as it is disable by default, [see last section](#alongside_ragflow).
1. Launch the service with `docker compose -f docker-compose.yml up -d`

KevinHuSh (approved):

Issue #7370, opened 2025-04-28
[Question]: How about information LLM debug


isthaison:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

 What do you think about this feature that allows users to view information about LLM and makes it easy for them to debug externally, for example promt or input
![Image](https://github.com/user-attachments/assets/d30b7431-416e-46e0-94e6-097928a34826)

isthaison:
![Image](https://github.com/user-attachments/assets/cb9683f7-ec68-4f9e-9f65-33dd9d28a621)
 This is debug information, with input of 3 components but when llm only passes question. Does this make the user misunderstand the meaning of the component or the component has a bug?

which-W:
@isthaison 

Thank you for your suggestions on our product

I will provide feedback to our technical team on your relevant proposals, and we will consider adding relevant debugging methods in the subsequent component design

isthaison:
![Image](https://github.com/user-attachments/assets/68483202-fd27-41e7-bdce-3f74198cc227) This is what I do, you can check it out.

ZhenhangTung:
Hi @isthaison , I'm the PM of RAGFlow—thanks so much for your feedback and for contributing the PR!

Just to share some context: the Agent module is currently being refactored, and the new version will include built-in execution logs for better debugging support.

Would it be okay if we revisit this PR after the new version is released? We'd love to continue the discussion then with the updated architecture in place.

Really appreciate your contribution!

PR #7375, opened 2025-04-28
fixed missing list input ref in query

isthaison:

fixed missing list input ref in query

### Type of change


isthaison:
ref : https://github.com/infiniflow/ragflow/issues/7370#issuecomment-2833901051

PR #7381, opened 2025-04-28
Refa: similarity calculations.

KevinHuSh:
### Type of change

- [x] Refactoring


whhe:
Hello @KevinHuSh, I'm curious about why this change was made. In the current implementation, the final similarity is only related to the weight of keyword, which may result in more cases where the similarity values are the same.

KevinHuSh:
The length of chunk varies much. Putting long chunk behind the short ones are usually bad.

Issue #7407, opened 2025-04-29
[Question]: Model(BAAI/bge-large-en-v1.5) not authorized


Robert0611:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

Under previous v0.17.2 full version, many large English documents already being parsed with the "BAAI/bge-base-en-v1.5" embedding model. After upgrade to the v0.18.0, this model disappeared and in the dataset "configuration" showed "BAAI/bge-large-en-v1.5@FastEmbed“ .
ERROR occured during chatting , shows "ERROR: Model(BAAI/bge-large-en-v1.5) not authorized"

![Image](https://github.com/user-attachments/assets/0b1d9c12-65a9-48ff-8ede-8d0439f953be)

![Image](https://github.com/user-attachments/assets/f746648b-a976-4a7a-9b53-9d987b8edf4d)

1940653868:
same problem in python apk Retrieve chunks

Robert0611:
By checking the .env, I found only two built-in embedding model were built in the version 0.18.0 , BAAI/bge-large-en-v1.5 was not included. but version 17.2 built in 9 embedding model. I think that the point for the ERROR

wangq8:
Closing this bug as it has been resolved.

PR #7426, opened 2025-04-30
Fix: Tried to fix the fid mis match under some cases 

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/7407

Based on this context, I think there should be some reasons that let some LLMs have a mismatch (add the wrong "@xxx"),
So I think when use fid can not fetch llm then tried to just use name should can fetch it.

### Type of change


Woody-Hu:
Hi @Robert0611 I base on my experience and the ticket context tried to fix your problem, do you have an env to verify this PR?

Robert0611:
Hi@Woody-Hu Thanks for your attention. I attach my env (change to be txt for uploading) here for your kind checking and so appreciate!
[env.txt](https://github.com/user-attachments/files/19970842/env.txt)

Issue #7224, opened 2025-04-23
[Bug]: Resource wordnet not found. Please use the NLTK Downloader to obtain the resource


kowyo:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

9a1ac8020dd53fae663766aeb6106c17b60970a5

### RAGFlow image version

v0.18.0

### Other environment information

```Markdown
- 4c/16 GB RAM/50 GB SSD
- Linux
```

### Actual behavior

According to https://github.com/infiniflow/ragflow#-launch-service-from-source-for-development, when I launch the backend service via `bash docker/launch_backend_service.sh`, it will get the following error:

```
Traceback (most recent call last):
  File "/root/ragflow/.venv/lib/python3.10/site-packages/nltk/corpus/util.py", line 84, in __load
    root = nltk.data.find(f"{self.subdir}/{zip_name}")
  File "/root/ragflow/.venv/lib/python3.10/site-packages/nltk/data.py", line 579, in find
    raise LookupError(resource_not_found)
LookupError: 
**********************************************************************
  Resource wordnet not found.
  Please use the NLTK Downloader to obtain the resource:

  >>> import nltk
  >>> nltk.download('wordnet')
  
  For more information see: https://www.nltk.org/data.html

  Attempted to load corpora/wordnet.zip/wordnet/

  Searched in:
    - '/root/nltk_data'
    - '/root/ragflow/.venv/nltk_data'
    - '/root/ragflow/.venv/share/nltk_data'
    - '/root/ragflow/.venv/lib/nltk_data'
    - '/usr/share/nltk_data'
    - '/usr/local/share/nltk_data'
    - '/usr/lib/nltk_data'
    - '/usr/local/lib/nltk_data'
**********************************************************************
```

### Expected behavior

Launch backend without issues

### Steps to reproduce

```Markdown
Follow steps on https://github.com/infiniflow/ragflow#-launch-service-from-source-for-development
```

### Additional information

[...]

Issue #6793, opened 2025-04-03
[Question]: bash docker/launch_backend_service.sh


Han-YLun:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

i runner `bash docker/launch_backend_service.sh`

then 

<img width="773" alt="Image" src="https://github.com/user-attachments/assets/70c70df6-5e73-440f-82ca-dca14c8cec8e" />

githubmorgue:
I got same problem, and then I tried to run frontend service. I got the picture uner. I consider this issue brought by the issue that backend encountered.

<img width="1552" alt="Image" src="https://github.com/user-attachments/assets/69907be4-3ada-4384-98cc-f7bb78eb77a0" />

githubmorgue:
I think you can try to comment out this code, to see the output. And this will be helpful for you to have a look about what happens to this procedure.

![Image](https://github.com/user-attachments/assets/ff30340c-c42a-4218-af64-b0af8b88f782)

this is mine:

```

  Resource wordnet not found.
  Please use the NLTK Downloader to obtain the resource:

  >>> import nltk
  >>> nltk.download('wordnet')
  
  For more information see: https://www.nltk.org/data.html

  Attempted to load corpora/wordnet.zip/wordnet/

  Searched in:
    - '/Users/jianguohu/nltk_data'
    - '/Users/jianguohu/Documents/Projects/Python/ragflow/.venv/nltk_data'
    - '/Users/jianguohu/Documents/Projects/Python/ragflow/.venv/share/nltk_data'
    - '/Users/jianguohu/Documents/Projects/Python/ragflow/.venv/lib/nltk_data'
    - '/usr/share/nltk_data'
    - '/usr/local/share/nltk_data'
    - '/usr/lib/nltk_data'
    - '/usr/local/lib/nltk_data'
```

and I solve this problem by execute the code under:

```
import nltk

[...]

PR #7447, opened 2025-04-30
fix(deps): incorrect nltk download dir

kowyo:

Fix https://github.com/infiniflow/ragflow/issues/7224 and https://github.com/infiniflow/ragflow/issues/6793

### Type of change

- [ ] New Feature (non-breaking change which adds functionality)a
- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ docker build --platform linux/amd64 -f Dockerfile -t infiniflow/ragflow:nightly 
    git clone https://github.com/infiniflow/ragflow.git
    cd ragflow/
    uv sync --python 3.10 --all-extras # install RAGFlow dependent python modules
+   uv run download_deps.py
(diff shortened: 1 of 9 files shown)


KevinHuSh:
Appreciation!
Is this compatible with  [Dockerfile line:31](https://github.com/infiniflow/ragflow/blob/main/Dockerfile#L31)

kowyo:
Hi @KevinHuSh, you are right, the original PR will affect [Dockerfile line:31](https://github.com/infiniflow/ragflow/blob/main/Dockerfile#L31), instead of changing the path inside it, I find a more neat solution.

I have also updated documentation for those who want to launch from source and tested above changes in a clean virtual machine.

Please let me know if further work is needed.

PR #7457, opened 2025-05-03
fixed errror when vars of cnt begin  declare with key contain "begin"

isthaison:
fixed errror when vars of cnt begin  declare with key contain "begin"

### Type of change

### DIFF
--- a/agent/component/base.py
+++ b/agent/component/base.py
@@ -466,7 +466,7 @@ class ComponentBase(ABC):
         outs = []
         for q in sources:
             if q.get("component_id"):
-                if q["component_id"].split("@")[0].lower().find("begin") >= 0:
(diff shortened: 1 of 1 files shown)


agent/component/base.py
```diff
@@ -466,7 +466,7 @@ def _fetch_outputs_from(self, sources: list[dict[str, Any]]) -> list[pd.DataFram
         outs = []
         for q in sources:
             if q.get("component_id"):
-                if q["component_id"].split("@")[0].lower().find("begin") >= 0:
```
KevinHuSh, line 469: If "@" not in q["component_id"], this statement is still false,  isn't it?

isthaison: image

PR #7461, opened 2025-05-04
Fix for embedded scroll

dangerousbeans:

This PR fixes an issue with the embeddable chat feature where the parent page jumps/scrolls to the iframe's position whenever a new message is added to the chat.

The problem occurs because the chat component uses scrollIntoView() to automatically scroll to new messages, but this method affects the parent page's scrolling as well, not just the content within the iframe. This creates a disruptive user experience when the iframe is positioned lower on a page, as users are constantly scrolled away from their current position.

The PR modifies the scrolling mechanism to use scrollTop = scrollHeight instead, which only affects the container itself, keeping the scrolling behavior contained within the iframe. It also updates the component structure to attach the ref to the scrollable container rather than to a div at the bottom of the messages.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


asiroliu:
@dangerousbeans 
I've built a Docker image using your latest code and encountered an issue: 
After multiple conversations, the input box gets pushed off-screen, making further questions impossible.

- Initialization status

image

- After multiple conversation turns...

image

asiroliu:
@dangerousbeans 
Please don't modify docker-compose.yml, .env, or settings.py files - these changes are unrelated to your feature implementation.

asiroliu:
@dangerousbeans 
I've tested your latest code(https://github.com/infiniflow/ragflow/pull/7461/commits/5ebff3c78612c358860f9bc00fd91e0c5593773c), but the input box still gets pushed off screen.

PR #7478, opened 2025-05-06
Fix Ollama instructions

raffaem:
Fix instructions for Ollama

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


raffaem:
Are the intructions to install Ollama with Docker really necessary?

I think thye mislead the user into thinking that the system won't work if Ollama is installed from binaries and in any case we should delegate that to Ollama documentation

Issue #7418, opened 2025-04-29
[Question]: embedding error


changqingla:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

v0.18.0  v0.17.2
I used embedding model bge-m3，When parsing Markdown files, some of them result in an error, such as:
rag-server  | Traceback (most recent call last):
rag-server  |   File "/ragflow/rag/svr/task_executor.py", line 613, in handle_task
rag-server  |     await do_handle_task(task)
rag-server  |   File "/ragflow/rag/svr/task_executor.py", line 558, in do_handle_task
rag-server  |     token_count, vector_size = await embedding(chunks, embedding_model, task_parser_config, progress_callback)
rag-server  |   File "/ragflow/rag/svr/task_executor.py", line 418, in embedding
rag-server  |     vts, c = await trio.to_thread.run_sync(lambda: mdl.encode([truncate(c, mdl.max_length-10) for c in cnts[i: i + batch_size]]))
...
rag-server  | openai.BadRequestError: Error code: 400 - {'object': 'error', 'message': "This model's maximum context length is 8192 tokens. However, you requested 10254 tokens in the input for embedding generation. Please reduce the length of the input.", 'type': 'BadRequestError', 'param': None, 'code': 400}
This kind of error occurred. This is my doc.

[99cd0edaf5bfb233e5f46ee55af837de0b725274_5259149d-94e6-4e0e-af47-1566c5ecc49e.md](https://github.com/user-attachments/files/19957044/99cd0edaf5bfb233e5f46ee55af837de0b725274_5259149d-94e6-4e0e-af47-1566c5ecc49e.md)

changqingla:
> 为了解决在 ragflow 中使用 bge-m3 模型时因超出 8192 个 token 的最大上下文长度而导致的嵌入错误，您可以调整类`chunk_token_num`中的参数`RAGFlowTxtParser`。这可确保将文本拆分成不超过 token 限制的较小块。以下是一段代码片段，可供参考：
> 
[...]

PR #7508, opened 2025-05-07
Fix:Optimized the chunking strategy for Markdown files to address issues where excessively long paragraphs or tables could cause errors during embedding.

changqingla:
- [x] Performance Improvement
Optimized the chunking strategy for Markdown files to address issues where excessively long paragraphs or tables could cause errors during embedding. Instead of simply splitting overly long paragraphs into two parts, the new approach performs more reasonable segmentation. For tables that exceed the threshold, the header is retained and the content is split more intelligently.


changqingla:
#7418  This PR can solve this question.

KevinHuSh:
Appreciations!

Issue #7491, opened 2025-05-06
[Bug]: refine reteival of multi-turn conversation in agent case


GavinZhou0125:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

#1152

### RAGFlow image version

v0.18.0

### Other environment information

```Markdown

```

### Actual behavior

This PR addresses a scenario where the input to the Retrieval module is configured to receive the output of the Interact module. In this case, the input includes both ASSISTANT and USER messages.

However, this setup introduces a chain of side effects, including but not limited to:
- In multi-turn conversations, the input to the retrieval module grows excessively, leading to ElasticSearch errors such as:
`Query contains too many nested clauses; maxClauseCount is set to 2520.`
- The retrieval performance degrades, and the recall rate decreases significantly due to overly complex and lengthy queries.

### Expected behavior

To mitigate these issues, we propose simplifying the input to the retrieval module by retaining only the user-side messages from multi-turn dialogues — such as Round1-question1, Round2-question2, etc. This helps reduce query complexity while preserving the core semantic intent for retrieval.

### Steps to reproduce

```Markdown
Create an agent composed of the following modules: Begin, Generate, Interact, and Retrieval. Configure each module accordingly — for example, set the prompt in the Generate module to be relatively long (e.g., 500 words). After linking the modules together, start a multi-turn conversation.

An error will occur after the third round (the exact number of turns depends on the prompt length — the longer the prompt, the fewer the turns before the failure).
```

### Additional information

The agent dsl may look like below

<img width="758" alt="Image" src="https://github.com/user-attachments/assets/493bc7a0-742c-4301-9384-f8682a38c243" />

KevinHuSh:
Appreciation!
By our evaluations, this feature is not going to be merged into main branch.

PR #7517, opened 2025-05-07
Feat: Allow using only the latest one message in conversation for retrieval

notsyncing:

Hello, we are facing a similar problem as #7491, where the `Retrieval` component is using the whole conversation in the `Answer` component, causing a non-precise retrieval result, while we only need the `Retrieval` component to search for the latest message only.

However, there may be some use cases needing the current behavior.

So, here, I propose an approach to resolve this: 

I added a switch to control the output fetching behavior for the `Retrieval` component:

image

When the switch is ON, it only fetches the latest one message from the `Answer` component for retrieval. Otherwise it keeps the original behavior.
The switch default is OFF to retain compatibility of old flows.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/agent/component/base.py
+++ b/agent/component/base.py
@@ -466,7 +466,11 @@ class ComponentBase(ABC):
     def set_infor(self, v):
         setattr(self._param, self._param.infor_var_name, v)
         
-    def _fetch_outputs_from(self, sources: list[dict[str, Any]]) -> list[pd.DataFrame]:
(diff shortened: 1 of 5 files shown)


KevinHuSh:
Appreciations!
By our evaluations, this feature is not going to be merged into main branch. We plan to apply other method to work around this issue.

Issue #7516, opened 2025-05-07
[Bug]: OAuth users don't get default LLMs populated (user_default_llm property)


jjmata:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

cb37f00a8fb517d8507dc48a7c017d57ba8828c0

### RAGFlow image version

 v0.18.0-61-gfc379e90 slim

### Other environment information

```Markdown
Running under Docker in AWS Linux
```

### Actual behavior

Registering a new user gets the default LLM populated, logging in via OAuth (configured GitHub in my case) doesn't do it

### Expected behavior

Both types of users are treated equally in so far as authentication means.  Add code that populates default LLM the first time an OAuth user is registered (upon first login)

### Steps to reproduce

```Markdown
0. Set up OAuth
1. Set up user_default_llm values
2. Register new user: confirm default LLM is populated
3. Log in with a different user via OAuth: notice no default LLM is set
```

jjmata:
I have a (vibe) bug fix for this ... PR coming up soon.

jjmata:
See #7520

Yannnnnnny:
Sorry for the very late reply. We’ve released many updates since then, so I wanted to check whether this issue is still happening for you.
If the issue no longer exists, feel free to close this issue. Thanks!

jjmata:
I'm not using the repo anymore, closing.

PR #7520, opened 2025-05-07
Fix for #7516

jjmata:

This makes it so GitHub OAuth users get their `user_default_llm` settings like the rest of the other ones do.

The SSO work @wengchaoxi is doing probably should take this into consideration, but it won't be a sufficient fix to cover all cases.

### Type of change


KevinHuSh:
Appreciations!

PR #7550, opened 2025-05-09
Fix: InfiniteScroll sometimes can not fetch next page

Woody-Hu:
### What problem does this PR solve?

https://github.com/infiniflow/ragflow/issues/7538

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

Issue #7538, opened 2025-05-08
[Bug]: agent list page (/canvas/listteam) does not render a  paginator when count is greater than 30


ChanningZhang:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

abcd

### RAGFlow image version

v0.18.0

### Other environment information

```Markdown

```

### Actual behavior

![Image](https://github.com/user-attachments/assets/53f46a50-b768-47e5-8e75-b7e2f1057073)

### Steps to reproduce

```Markdown
create agents and up to 30+
```

PR #7596, opened 2025-05-13
FIX:  knowledge will not render a paginator when count is greater than 30

Woody-Hu:

as https://github.com/infiniflow/ragflow/issues/7538
and https://github.com/infiniflow/ragflow/pull/7550

### Type of change

### DIFF
--- a/web/src/pages/knowledge/index.tsx
+++ b/web/src/pages/knowledge/index.tsx
@@ -79,33 +79,44 @@ const KnowledgeList = () => {
         </Space>
       </div>
       <Spin spinning={loading}>
-        <InfiniteScroll
-          dataLength={nextList?.length ?? 0}
-          next={fetchNextPage}
-          hasMore={hasNextPage}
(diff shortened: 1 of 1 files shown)


asiroliu:
@Woody-Hu 
I couldn't reproduce the knowledge base scroll loading issue in the nightly build. 
However, when testing with your version, I observed that only 90 out of 100 knowledge bases loaded properly - the remaining 10 failed to load during scrolling.
- your version:
image

- nightly:
image

Woody-Hu:
> @Woody-Hu I couldn't reproduce the knowledge base scroll loading issue in the nightly build. However, when testing with your version, I observed that only 90 out of 100 knowledge bases loaded properly - the remaining 10 failed to load during scrolling.
> 
> * your version:
> 
> image
> * nightly:
> 
> image

Interesting I just found the knowledge and agent page are using the same code pattern, so not do deep test.

asiroliu:
@Woody-Hu 
I've tested your latest version and can confirm the behavior is working as expected.

PR #7680, opened 2025-05-16
Test: use environment variable for ZHIPU_AI_API_KEY

asiroliu:

use environment variable for ZHIPU_AI_API_KEY

### Type of change

- [x] Test update


asiroliu:
GitHub's security model prevents repository secrets from being exposed to pull requests originating from forked repositories by default.

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions
image

Issue #7703, opened 2025-05-18
[Bug]:Exceed the maximum length of file name


Simonqujian78:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

No found related issues.

### RAGFlow image version

 v0.18.0-137-g36e32dde slim

### Other environment information

```Markdown
win11+docker4.41.2+lm studio 0.3.15+ragflow:nightly.
```

### Actual behavior

I  want to add a new file in my "Knowledge database", the filename is "938--国家金融监督管理总局办公厅关于规范总局系统行政检查工作的通知（金办发〔2025〕45号）", but a error information " Exceed the maximum length of file name!" . 

![Image](https://github.com/user-attachments/assets/574dbc1e-cbd9-4921-99c8-679a936a2e68)

 It is the first time for me to encounter the error information.I also uploaded many files whose length of filenames is longer than "938"

### Expected behavior

 I wonder how to revise the maximum length of the filename.

### Steps to reproduce

```Markdown
Just upload the file, click"OK".
```

Simonqujian78:
@dosu, thanks for your reply. As your said ,  the maximum length of filename is 1024, the filename I uploaded  is "938--国家金融监督管理总局办公厅关于规范总局系统行政检查工作的通知（金办发〔2025〕45号）", it is obvious that the length of the filename is smaller than 1024, so  maybe it is a bug.

Simonqujian78:
OK, understand, 128-byte is the limit of the length of filename. If I revise  NAME_LENGTH_LIMIT = 2**11 in constants.py file, the limit will be also revised 256-byte?

Woody-Hu:
> OK, understand, 128-byte is the limit of the length of filename. If I revise NAME_LENGTH_LIMIT = 2**11 in constants.py file, the limit will be also revised 256-byte?

Hi @Simonqujian78, based on my experience, this config is not in the config file.
`
                if len(file.filename.encode("utf-8")) >= 128:
[...]

PR #7707, opened 2025-05-19
Feat: Support Config Max File Name Length

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/7703

### Type of change

### DIFF
--- a/api/apps/sdk/doc.py
+++ b/api/apps/sdk/doc.py
@@ -137,9 +137,10 @@ def upload(dataset_id, tenant_id):
             return get_result(
                 message="No file selected!", code=settings.RetCode.ARGUMENT_ERROR
             )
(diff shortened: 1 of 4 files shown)


KevinHuSh:
It's a bit more complicated for this sceniario.
The file name length is restricted by database table field definition which should be consistancy with the corresponding envirement variable.

Woody-Hu:
> It's a bit more complicated for this scenario. The file name length is restricted by database table field definition which should be consistancy with the corresponding envirement variable.

add some logics to modify db (but should just effect the first time)

KevinHuSh:
Mr. Hu, I found that the maximum of file name length in bytes is 255<sup>[[1]](https://serverfault.com/questions/9546/filename-length-limits-on-linux). So, it might be meaningless to set a envirement variable to control it.
Thanks again.

Issue #7729, opened 2025-05-20
[Bug]: repair_bad_citation_formats()  error in chat wrong hint type


BadwomanCraZY:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

latest

### RAGFlow image version

v0.18.0-173-g1ae7b942 full

### Other environment information

```Markdown

```

### Actual behavior

![Image](https://github.com/user-attachments/assets/73714175-7eca-4595-973a-466f1969a9e5)

```
2025-05-20 14:03:29,658 INFO     29 172.18.0.6 - - [20/May/2025 14:03:29] "POST /v1/conversation/completion HTTP/1.1" 200 -
Traceback (most recent call last):
  File "/ragflow/api/apps/[conversation_app.py](http://conversation_app.py/)", line 231, in stream
    for ans in chat(dia, msg, True, **req):
  File "/ragflow/api/db/services/[dialog_service.py](http://dialog_service.py/)", line 443, in chat
    yield decorate_answer(thought + answer)
  File "/ragflow/api/db/services/[dialog_service.py](http://dialog_service.py/)", line 366, in decorate_answer
    answer, idx = repair_bad_citation_formats(answer, kbinfos, idx)
  File "<@beartype(api.db.services.dialog_service.chat.repair_bad_citation_formats) at 0x73c8c36e36d0>", line 66, in repair_bad_citation_formats
beartype.roar.BeartypeCallHintParamViolation: Function api.db.services.dialog_service.chat.repair_bad_citation_formats() parameter idx={0, 3} violates type hint <class 'dict'>, as set {0, 3} not instance of dict.
2025-05-20 14:03:37,894 INFO     31 task_executor_58cef76f3e5e_0 reported heartbeat: {"name": "task_executor_58cef76f3e5e_0", "now": "2025-05-20T14:03:37.893+08:00", "boot_at": "2025-05-20T13:59:37.826+08:00", "pending": 0, "lag": 0, "done": 0, "failed": 0, "current": {}}
[...]

PR #7738, opened 2025-05-20
Fix: wrong type hint

yongtenglei:

Wrong hint type. #7729.

### Type of change


doubuns:
Why isn't this commit merged into the main branch so that the nightly build image can be used?

Issue #7753, opened 2025-05-21
[Bug]: File selection in Retrieval testing causes other options to disappear


asiroliu:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

x

### RAGFlow image version

7c3407a7e482(infiniflow/ragflow:nightly)

### Other environment information

```Markdown

```

### Actual behavior

File selection in Retrieval testing causes other options to disappear
- Before file selection:

<img width="491" alt="Image" src="https://github.com/user-attachments/assets/5198e6e4-363e-4323-b1dd-634bf8311fc1" />

- After file selection:

<img width="340" alt="Image" src="https://github.com/user-attachments/assets/197a52da-c6a7-436b-85aa-7132ea07283c" />

### Expected behavior

Multiple file selection should be supported

### Steps to reproduce

```Markdown
x
```

PR #7759, opened 2025-05-21
Fix: File selection in Retrieval testing causes other options to disappear

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/7753

The internal is due to when the selected row keys change will trigger a testing, but I do not know why.

### Type of change

### DIFF
--- a/web/src/pages/add-knowledge/components/knowledge-testing/testing-result/select-files.tsx
+++ b/web/src/pages/add-knowledge/components/knowledge-testing/testing-result/select-files.tsx
@@ -50,7 +50,7 @@ const SelectFiles = ({ setSelectedDocumentIds, handleTesting }: IProps) => {
 
   const rowSelection = {
     onChange: (selectedRowKeys: React.Key[]) => {
-      handleTesting(selectedRowKeys as string[]);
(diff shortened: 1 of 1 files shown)


asiroliu:
@Woody-Hu 
Current Status:
Fixed: File selection no longer causes other options to disappear
Remaining Issue: File filtering fails to properly restrict search results

Woody-Hu:
> @Woody-Hu Current Status: Fixed: File selection no longer causes other options to disappear Remaining Issue: File filtering fails to properly restrict search results

Hi @asiroliu I want to confirm the original design for this feature, If select the check box, just search in these select items?

asiroliu:
The workflow should be: first return complete search results containing all relevant chunks, then filter the display based on file selection to show which files contain these chunks.

asiroliu:
@Woody-Hu 
I've tested the latest version (40de0bd4) and confirm the current behavior meets expectations. 
However, I've noticed the same issue exists on the search page. Would you mind fixing it there as well?

image

Woody-Hu:
> @Woody-Hu Hi, just checking if you've seen the previous message?

@asiroliu I just get the  notice your message, will update that PR.

Issue #7869, opened 2025-05-27
[Bug]: the wrong use of task_limiter cause parse stuck when enable graphrag


sinopec:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

latest

### RAGFlow image version

0.19.0-slim

### Other environment information

```Markdown
CPU:Intel(R) Xeon(R) Silver 4316 CPU @ 2.30GHz
MEM:128GB
OS:Ubuntu 22.04-server

set WORKERS = 4 in entrypoint.sh to start 4 task_executor.
```

### Actual behavior

in task_executor.py  do_handle_task ,when task_type is graphrag, the task_limiter will be replaced with a new one.
```python
        global task_limiter
        task_limiter = trio.CapacityLimiter(2)
```

this is the wrong use of trio  CapacityLimiter, and will cause capacity limit outof control.
this can be verified in the follow code
```python
import trio

MAX_CONCURRENT_TASKS = 5
task_limiter = trio.CapacityLimiter(MAX_CONCURRENT_TASKS)

# 模拟任务函数
async def long_task(n):
    global task_limiter
    async with task_limiter:
        print(f"[{n}] Got token, running...")
        if n % 5 == 0 :
            task_limiter = trio.CapacityLimiter(2)
            #task_limiter.total_tokens = 2

        await trio.sleep(20 * 60)  
        print(f"[{n}] Task completed.")

async def task_generator():
    task_id = 1
    while True:
        task_id += 1
        print(f"[{task_id}] Spawning new task...")
        trio.lowlevel.spawn_system_task(long_task, task_id)
        await trio.sleep(1)  


if __name__ == "__main__":
    trio.run(task_generator)

```

the task will be handlled endless untile crash.

### Expected behavior

if we want to  set the capacity dynamicly  ,we shou use 
```python
task_limiter.total_tokens = 2
```

but this is still not the right way to control graphrag task count .I think we should do it  in two ways:
[...]

PR #7870, opened 2025-05-27
Fix: Avoid override global task limiter

Woody-Hu:
https://github.com/infiniflow/ragflow/issues/7869

### Type of change


KevinHuSh:
What do you think about PR #7873?

PR #7873, opened 2025-05-27
Fix: task limiter issue.

KevinHuSh:

#7869

### Type of change


sinopec:
kg_limiter can prevent the generation of unlimited tasks, but graphrag tasks still consume the capacity of the regular task_limiter, which may block normal chunk tasks.

Therefore, a better approach is to use separate Redis queues for graphrag, raptor, and regular chunk tasks. This way, different task_limiters can be applied to control concurrency independently, rather than allowing long-running and non-parallelizable graphrag tasks to consume all the available capacity and block the execution of other tasks.

PR #7938, opened 2025-05-29
Waynechen

as775116191:

My business team wants to use file management like a cloud storage service, but the current file management system is quite cumbersome. It requires managing a knowledge base, then finding documents in the knowledge base to parse them. Therefore, I have simplified the steps, but I am not sure if this is what most people need.

### Type of change

- [ ] After creating a folder and managing the knowledge base, record this relationship. Any files uploaded or subfolders created under this folder will be associated with this knowledge base
- [ ] When uploading files to a folder associated with a knowledge base, you can check 'Parse on creation', which will automatically enable parsing without needing to jump to the knowledge base page to find the document
- [ ] When modifying the associated knowledge base of a folder, all nested files below will also be modified: documents will be created in the new knowledge base and deleted from the old knowledge base.

### DIFF
--- a/api/apps/file2document_app.py
+++ b/api/apps/file2document_app.py
@@ -44,54 +44,101 @@ def convert():
             file = files_set[file_id]
             if not file:
                 return get_data_error_result(message="File not found!")
-            file_ids_list = [file_id]
+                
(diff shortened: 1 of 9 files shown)


KevinHuSh:
Appreciations!

PR #7962, opened 2025-05-30
Update docker-compose.yml

xiaoping378:
If the name field is not specified, Docker Compose will default to using `docker` as the project name. This may cause conflicts with other default projects, leading to unintended operations when executing `docker compose` commands.

When executing Docker Compose commands, interference occurs between multiple default projects, leading to operational chaos.​

### Type of change

- [x] Other (please describe):

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,3 +1,5 @@
+name: ragflow
+
 include:
   - ./docker-compose-base.yml
 


xiaoping378:
when upgrade，potential issues with Docker volume initialization，VOLUME-NAME changed. e.g from docker_esdata01 to ragflow_esdata01。

PR #7974, opened 2025-05-30
Allow Update Conversation Parameters and Persist to Database in completion

huang-aoqin:

This PR updates the completion function to allow parameter updates when a session_id exists. It also ensures changes are saved back to the database via API4ConversationService.

### Type of change


api/db/services/canvas_service.py
```diff
@@ -173,6 +173,15 @@ def completion(tenant_id, agent_id, question, session_id=None, stream=True, **kw
             conv.reference = []
         conv.reference.append({"chunks": [], "doc_aggs": []})
 
+        query = canvas.get_preset_param()
+        if query:
+            for ele in query:
+                if kwargs.get(ele["key"]):
+                    ele["value"] = kwargs[ele["key"]]
+
+        conv.dsl = json.loads(str(canvas))
```
KevinHuSh: Should this be in the `if` scope more reasonable?

Issue #8052, opened 2025-06-04
[Bug]: /api/v1/retrieval: dataset_id format (string vs. array) depends on chunk ingestion source (/chunk/create vs. file import)


openproix:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

a99739166cf5be0553dcf2e56f39c80b5878e324 

### RAGFlow image version

v0.17.2

### Other environment information

```Markdown

```

### Actual behavior

The /api/v1/retrieval endpoint returns an inconsistent format for the dataset_id field within the chunks array of its response.

When chunks are ingested into RAGFlow via file import, the corresponding dataset_id in the API response is a string (e.g., "xxxxxxxx").
When chunks are created via the /api/v1/chunk/create API endpoint, the corresponding dataset_id in the /api/v1/retrieval API response is an array containing a single string (e.g., ["xxxxxxxx"]). This inconsistency in data type (string vs. array) for the same conceptual field causes deserialization issues in strongly-typed client applications.

### Expected behavior

The dataset_id field in the /api/v1/retrieval API response should have a consistent data format for all chunks, regardless of how they were created or ingested.
Ideally, since dataset_id typically refers to a single dataset to which a chunk belongs, it should consistently be a string. If there's a reason for it to be an array (e.g., a chunk could belong to multiple datasets), then it should always be an array (even if it's an array with a single element or an empty array). The current mixed behavior is problematic.

### Steps to reproduce

```Markdown
Create a Knowledge Base:
Create a new knowledge base (dataset) in the RAGFlow UI. Let's call its ID test_kb_id.
Import a File:
Upload a file into test_kb_id. For example, an Excel file formatted as QA pairs, containing at least one QA pair. This will be parsed into one or more chunks by RAGFlow.
Manually Add a Chunk via UI:
Navigate within the RAGFlow UI to the knowledge base test_kb_id and the document created from the imported file.
Manually add a new chunk to this document through the UI.
Query Elasticsearch:
[...]

PR #8053, opened 2025-06-04
Revert kb_id assignment to string for consistency

hexsean:

This change reverts the kb_id assignment in commit a997391... back to a direct string assignment (doc.kb_id) instead of a list ([doc.kb_id]). This aims to resolve the dataset_id format inconsistency in the /api/v1/retrieval API response.
Addresses issue #8052"

### Type of change

### DIFF
--- a/api/apps/chunk_app.py
+++ b/api/apps/chunk_app.py
@@ -234,7 +234,7 @@ def create():
         e, doc = DocumentService.get_by_id(req["doc_id"])
         if not e:
             return get_data_error_result(message="Document not found!")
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!
`[api/apps/chunk_app.py](https://github.com/infiniflow/ragflow/pull/8053/files#diff-b243a23e50d10d2e02bc753dca4561c587544b4aa938a84ab855b30b4c501c20)` is for front-end API. This alteration could bring unknown risk to the current state.
This feature is not going to be merged into main branch.

PR #8182, opened 2025-06-11
Fix: Use lazy variable resolving to support envs like Swarm where nginx may not init properly at startup

ChanningZhang:

When starting the ragflow image in a Swarm cluster, the corresponding Swarm service `ragflow` may not be created immediately. In such cases, `proxy_pass http://ragflow:9380;` can lead to a fatal error due to unresolved DNS.

### Type of change
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/docker/nginx/ragflow.conf
+++ b/docker/nginx/ragflow.conf
@@ -3,6 +3,8 @@ server {
     server_name _;
     root /ragflow/web/dist;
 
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciation!

PR #8232, opened 2025-06-12
Fix: unify opendal config key from ‎`schema` to ‎`scheme`

cutiechi:

This PR resolves the inconsistency in the opendal configuration where both ‎`schema` and ‎`scheme` were used as keys. The code and configuration file now consistently use ‎`scheme`, which helps prevent configuration errors and runtime issues. This change improves code clarity and maintainability.

### Type of change

### Additional context
- Updated both ‎`conf/service_conf.yaml` and ‎`rag/utils/opendal_conn.py` to use ‎`scheme` instead of ‎`schema`
- No breaking changes to other configuration fields

### DIFF
--- a/conf/service_conf.yaml
+++ b/conf/service_conf.yaml
@@ -29,7 +29,6 @@ redis:
   db: 1
   password: 'infini_rag_flow'
   host: 'localhost:6379'
-
 # postgres:
 #   name: 'rag_flow'
 #   user: 'rag_flow'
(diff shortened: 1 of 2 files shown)


KevinHuSh:
Appreciations!
I didn't quite follow the reason why `schema` to `scheme` ?

cutiechi:
> Appreciations! I didn't quite follow the reason why `schema` to `scheme` ?

@KevinHuSh Thanks for your question! In opendal, the correct config key is scheme, not schema, because scheme refers to the type of storage service or protocol (such as s3, fs, http, etc.), similar to the scheme part in a URL (e.g., https://). This matches the terminology used in the official opendal documentation and codebase.

You can find more details in the official documentation: https://opendal.apache.org/docs/rust/opendal/enum.Scheme.html

By contrast, schema usually means a data structure or definition (like a database schema), which is unrelated to selecting a storage backend type. So using scheme here is both semantically accurate and consistent with opendal’s standard usage.

Issue #8228, opened 2025-06-12
[Bug]: Inconsistent page_size Behavior in retrieval Function When doc_ids is Provided


RyzeAngler:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

cc8029a7

### RAGFlow image version

3a34def55f161

### Other environment information

```Markdown

```

### Actual behavior

In the retrieval function of the Dealer class, located in rag/nlp/search.py (starting from line 348), there is a forced override of the page_size parameter when the doc_ids parameter is present. This raises questions about parameter consistency and could lead to unexpected behavior.
Specifically, in the following code snippet:

`        if doc_ids:
            similarity_threshold = 0
            page_size = 30`

When doc_ids is True, the page_size variable within the function is unconditionally reassigned to 30, regardless of the value initially passed by the caller.
Problem Analysis:
Parameter Semantic Inconsistency: The retrieval function explicitly accepts a page_size parameter, which typically implies that the caller can control the number of results returned. However, in the doc_ids scenario, this forced override renders the caller's page_size value ineffective, leading to unclear parameter semantics and unpredictable function behavior.
Potential Computational Waste: The calculation of idx (which selects chunk indices from the reranked results) on line 382:
idx = np.argsort(sim * -1)[(page - 1) * page_size:page * page_size]
This calculation uses the original page_size value passed by the caller. This means if the initial page_size is, for example, 100, the system will process and identify indices for 100 candidate chunks. However, the subsequent loop's length check and the final truncation will use the page_size value that has been forcibly set to 30 (when doc_ids is true). This leads to a mismatch where more chunks are processed initially than are ultimately returned, potentially causing unnecessary computation for the unused chunks.
Limited Flexibility: This hardcoded limitation restricts the ability to flexibly control the number of returned chunks when doc_ids are specified. Users might need fewer than 30 chunks (e.g., just a few representative snippets from specific documents) or more than 30 (e.g., all relevant chunks from a set of specified documents).

### Expected behavior

[...]

PR #8242, opened 2025-06-13
fix(search.py):remove hard page_size

RyzeAngler:

Fix the restriction of forcing similarity_threshold=0 and page_size=30 when doc_ids is not empty

#8228

### DIFF
--- a/rag/nlp/search.py
+++ b/rag/nlp/search.py
@@ -380,15 +380,14 @@ class Dealer:
                 rank_feature=rank_feature)
         # Already paginated in search function
         idx = np.argsort(sim * -1)[(page - 1) * page_size:page * page_size]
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!

For the following scenario, there're nothing to retrieve:

> User: Please summarize this file, what's this all about? (After uploading a file as an attachment by which system utilizes the doc ID while retrieval)

What do you think about that? Your opinion is appreciated.

RyzeAngler:
I don't understand why the source code enforces page_size=30 when using doc_ids for retrieval, as this does not align with the original intention of the search. The hardcoding in the source code causes the page_size I provide to be ineffective. My use case requires a larger page_size.

PR #8238, opened 2025-06-12
Fix: Update customer_service.json

Kennytian:
### What problem does this PR solve?

The issue of reporting the 「Can't inference the where the component input is. Please identify whose output is this component's input」error when creating an Agent using the Customer service template has been resolved.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

Issue #8211, opened 2025-06-12
[Bug]: Can't inference the where the component input is. Please identify whose output is this component's input.


Kennytian:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

x

### RAGFlow image version

v0.19.0 and nightly

### Other environment information

```Markdown
Windows subsystem for linux（Ubuntu 22）

Docker image:
swr.cn-north-4.myhuaweicloud.com/infiniflow/ragflow   nightly        15 hours ago    14.8GB
infiniflow/ragflow                                    v0.19.0        2 weeks ago     15.1GB
valkey/valkey                                         8              5 weeks ago     136MB
infiniflow/ragflow                                    v0.18.0-slim   7 weeks ago     7.11GB
infiniflow/infinity                    v0.6.0-dev3        4 months ago    885MB
mysql                                  8.0.39                   10 months ago   573MB
quay.io/minio/minio                    RELEASE.2023-12-20T01-00-02Z    17 months ago   151MB
elasticsearch                                         8.11.3       18 months ago   1.41GB
```

### Actual behavior

27 Canvas.run got exception: Can't inference the where the component input is. Please identify whose output is this component's input.
ragflow-server  | Traceback (most recent call last):
ragflow-server  |   File "/ragflow/agent/canvas.py", line 224, in prepare2run
ragflow-server  |     ans = cpn.run(self.history, **kwargs)
ragflow-server  |   File "/ragflow/agent/component/base.py", line 378, in run
ragflow-server  |     raise e
ragflow-server  |   File "/ragflow/agent/component/base.py", line 374, in run
ragflow-server  |     res = self._run(history, **kwargs)
ragflow-server  |   File "/ragflow/agent/component/retrieval.py", line 61, in _run
ragflow-server  |     query = self.get_input()
ragflow-server  |   File "/ragflow/agent/component/base.py", line 506, in get_input
[...]

PR #8252, opened 2025-06-13
Fix: Update customer service template with query references to RewriteQuestion

asiroliu:

- Add query references to "RewriteQuestion:AllNightsSniff" in multiple components
- Set "selected" to false for retrieval node

### Type of change


asiroliu:
#8238 #8211

asiroliu:
In the initial state, both "Question Categorize" and "Search product info" have their references set to "Refine Question".

image

image

PR #8277, opened 2025-06-16
Refa: change to db exists operation

Woody-Hu:

Change some API's db operations to exist to improve performance.

### Type of change

- [x] Performance Improvement


asiroliu:
@Woody-Hu 
Please refer to the original query method implementation - we cannot directly pass filter conditions to the Model's where() method here.
https://github.com/infiniflow/ragflow/blob/main/api/db/db_models.py#L177

Issue #7150, opened 2025-04-20
[Bug]: API of list chunks retuned chunks less than in knowledge base


Dyson-Ido:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

ddd

### RAGFlow image version

dddd

### Other environment information

```Markdown

```

### Actual behavior

It says there are 3130 chunks but when list chunks with API it returned 3018

### Steps to reproduce

```Markdown
http://{ip}/datasets/{dataset_id}/documents/{document_id}/chunks?page_size=verybignumberenough
```

asiroliu:
@Dyson-Ido 
I've tested with version e9e85d72b1b7(infiniflow/ragflow:nightly) using a list of 3000 chunks but was unable to reproduce the reported issue. Could you please provide additional details to help me investigate further?

Dyson-Ido:
> [@Dyson-Ido](https://github.com/Dyson-Ido) I've tested with version e9e85d72b1b7(infiniflow/ragflow:nightly) using a list of 3000 chunks but was unable to reproduce the reported issue. Could you please provide additional details to help me investigate further?

@asiroliu  please try with more than 3000 chunks, maybe the issue happens when chunk size is greater 3000

asiroliu:
@Dyson-Ido 
I tested with 4,000 chunks but still couldn't reproduce the issue.

Colstuwjx:
Hi @asiroliu ,I think I've reproduced this issue with the version 09f8dfe456a13849cfde06d0982e37c18a808c5b , and what I've seen is:

the chunk number displayed is 505157:
<img width="1365" alt="Image" src="https://github.com/user-attachments/assets/a9602861-ad78-41bd-9c8f-b2a836d007e0" />

however, the doc count in es is ONLY 36395:

<img width="924" alt="Image" src="https://github.com/user-attachments/assets/bbd13d6e-5f1e-4b5f-a1bd-d04453151c1a" />

and also the chunk detail page is also ONLY 36395:

[...]

PR #8305, opened 2025-06-16
fix: create chunk with random uuid, no increase count num if index error

Colstuwjx:

Addresses issue by  https://github.com/infiniflow/ragflow/issues/7150#issuecomment-2976432979 , aligned with dify behavior:

1. no increase chunk count / token count if index error;
2. set chunk id with a random uuid rather than using the hash string (one issue here is that ragflow is using uuid1 rather than uuid4, open for discussion)

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/api/apps/chunk_app.py
+++ b/api/apps/chunk_app.py
@@ -29,7 +29,7 @@ from api.db import LLMType, ParserType
 from api.db.services.knowledgebase_service import KnowledgebaseService
 from api.db.services.llm_service import LLMBundle
 from api.db.services.user_service import UserTenantService
-from api.utils.api_utils import server_error_response, get_data_error_result, validate_request
+from api.utils.api_utils import server_error_response, get_data_error_result, validate_request, get_uuid
 from api.db.services.document_service import DocumentService
 from api import settings
 from api.utils.api_utils import get_json_result
@@ -219,7 +219,7 @@ def rm():
 @validate_request("doc_id", "content_with_weight")
 def create():
     req = request.json
-    chunck_id = xxhash.xxh64((req["content_with_weight"] + req["doc_id"]).encode("utf-8")).hexdigest()
+    chunck_id = get_uuid()
     d = {"id": chunck_id, "content_ltks": rag_tokenizer.tokenize(req["content_with_weight"]),
          "content_with_weight": req["content_with_weight"]}
     d["content_sm_ltks"] = rag_tokenizer.fine_grained_tokenize(d["content_ltks"])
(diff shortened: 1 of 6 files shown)


Colstuwjx:
cc @asiroliu @KevinHuSh help review please.
Thanks.

Colstuwjx:
BTW, it would be great if someone could help me to understand why the [docs](https://github.com/infiniflow/ragflow/blob/f707403/api/db/services/document_service.py#L628) declared inside a for loop had been used in the following [chunks write](https://github.com/infiniflow/ragflow/blob/f707403/api/db/services/document_service.py#L676), really confused to me.

KevinHuSh:
Appreciations!
By our design, the same content should have the same chunk id, which has de-duplication effects.

Colstuwjx:
> By our design, the same content should have the same chunk id, which has de-duplication effects.

Hi @KevinHuSh , how about adding one more field like content_digest, and we could use this field to do cache-hit, content validation and content de-duplication (based on the `content` and `important_keywords`, only when they're 100% match)

Otherwise, the two chunks like the example below wouldn't be able to saved individually:

```
# chunk 1
{
    "content": "This is a hello-world example",
    "important_keywords": ["cn", "maintainer: xxx", "reference-link: xxx"]
}

# chunk 2
{
    "content": "This is a hello-world example",
    "important_keywords": ["demo", "hello-world"]
}
```

However, sometimes we need to have both these two chunks and query them by different keyword individually.

KevinHuSh:
> > By our design, the same content should have the same chunk id, which has de-duplication effects.
> 
> Hi @KevinHuSh , how about adding one more field like content_digest, and we could use this field to do cache-hit, content validation and content de-duplication (based on the `content` and `important_keywords`, only when they're 100% match)
> 
> Otherwise, the two chunks like the example below wouldn't be able to saved individually:
> 
> ```
> # chunk 1
> {
>     "content": "This is a hello-world example",
>     "important_keywords": ["cn", "maintainer: xxx", "reference-link: xxx"]
> }
> 
> # chunk 2
> {
>     "content": "This is a hello-world example",
>     "important_keywords": ["demo", "hello-world"]
> }
> ```
> 
> However, sometimes we need to have both these two chunks and query them by different keyword individually.

For this case, what about adding all the keywords into chunk 1?

Colstuwjx:
> what about adding all the keywords into chunk 1?

I've discussed with my internal business users, they said these two chunks need to be stored individually for expressing different intent. SO I chose to forked ragflow and modified the implementation by this PR.

Thanks.

PR #8326, opened 2025-06-17
Doc: Update README badges

asiroliu:

- Highlight current language in README badges by changing color for Traditional and Simplified Chinese

### Type of change

- [x] Documentation Update


asiroliu:
- Before:
image
image

- After:
image
image

PR #8383, opened 2025-06-20
Fix qianfan api bug

schhaohao:

This PR fixes a bug in how Qianfan API credentials (`ak` and `sk`) are handled in `rag/llm/chat_model.py`, `embedding_model.py`, and `rerank_model.py`.

Previously, the code attempted to pass `ak` and `sk` directly into `qianfan.ChatCompletion(ak=ak, sk=sk)`. However, the `ChatCompletion` class does not accept credentials via parameters—it instead expects them to be loaded from environment variables. As a result, API calls would fail due to missing authentication.

In all three files, this patch sets the environment variables before creating the client.

Example in `chat_model.py`:

```python
os.environ["QIANFAN_ACCESS_KEY"] = key.get("yiyan_ak", "")
os.environ["QIANFAN_SECRET_KEY"] = key.get("yiyan_sk", "")
self.client = qianfan.ChatCompletion()
```

### Type of change

### DIFF
--- a/rag/llm/embedding_model.py
+++ b/rag/llm/embedding_model.py
@@ -796,9 +796,12 @@ class BaiduYiyanEmbed(Base):
         import qianfan
 
         key = json.loads(key)
-        ak = key.get("yiyan_ak", "")
-        sk = key.get("yiyan_sk", "")
-        self.client = qianfan.Embedding(ak=ak, sk=sk)
+        
+        os.environ["QIANFAN_ACCESS_KEY"] = key.get("yiyan_ak", "")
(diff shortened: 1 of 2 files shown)


KevinHuSh:
Appreciations!
The `API key` in enviroment variable means that every loggined user could use that API, which is inappropriate, isn't it?

schhaohao:
> Appreciations! The `API key` in enviroment variable means that every loggined user could use that API, which is inappropriate, isn't it?



> Appreciations! The `API key` in enviroment variable means that every loggined user could use that API, which is inappropriate, isn't it?

Understand, but Qianfan's call does not support the ak and sk injection methods in the original code, and the original code is not available.

KevinHuSh:

Issue #8467, opened 2025-06-24
[Bug]: user_default_llm configuration doesn't work for OpenAI API compatible LLM factory


sd109:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

d6a941ebf57a1a0a422d4b2c4a402db1098f3586

### RAGFlow image version

v0.19.0 full

### Other environment information

```Markdown

```

### Actual behavior

When trying to configure a default LLM using configuration such as:
```
user_default_llm:
  factory: 'OpenAI-API-Compatible'
  base_url: 'https://llm1.example.com/v1'
  api_key: 'example'
  default_models:
    chat_model: 'llm1'
    embedding_model: ''
    rerank_model: ''
    asr_model: ''
    image2text_model: 'llm1'
```
the chat fails to generate responses and we hit this error: https://github.com/infiniflow/ragflow/blob/d6a941ebf57a1a0a422d4b2c4a402db1098f3586/api/db/services/llm_service.py#L123

but for other LLM factories such as
```
user_default_llm:
    factory: 'Tongyi-Qianwen'
    api_key: 'sk-xxxxxxxxxxxxx'
    base_url: ''
    default_models:
      chat_model: 'qwen-plus'
      embedding_model: 'BAAI/bge-large-zh-v1.5@BAAI'
      rerank_model: ''
      asr_model: ''
      image2text_model: ''
```
the default LLM gets set correctly.

It looks like the model doesn't get registered properly somehow:
<img width="624" alt="Image" src="https://github.com/user-attachments/assets/0dfeba22-b977-4c78-bc06-ca90cc680125" />

compared to the other factory where it works:
<img width="628" alt="Image" src="https://github.com/user-attachments/assets/d6523d32-aadf-4214-8b18-340fe76cec7f" />

If I input the same OpenAI compatible model details into the UI manually then the model works, but only for that user and not for every other user.

### Steps to reproduce

```Markdown
Configure a default LLM using the OpenAI-API-Compatible model factory.
```

[...]

PR #8500, opened 2025-06-25
Fix:user_default_llm configuration doesn't work for OpenAI API compatible LLM factory

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/8467
for this case is due to when save to tenant LLM as llm1, but will use llm1@openAICompatible to do query

### Type of change

### DIFF
--- a/api/db/services/llm_service.py
+++ b/api/db/services/llm_service.py
@@ -70,6 +70,8 @@ class TenantLLMService(CommonService):
             model_factories = settings.FACTORY_LLM_INFOS
             model_providers = set([f["name"] for f in model_factories])
             if arr[-1] not in model_providers:
+                if arr[-1] in ["OpenAI-API-Compatible"]:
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!

https://github.com/infiniflow/ragflow/blob/c4bfd9fa2c742544488edfa512004f33130eec93/conf/llm_factories.json#L510

`OpenAI-API-Compatible` should be in `settings.FACTORY_LLM_INFOS`, shouldn't be?

Woody-Hu:
> Appreciations!
> 
> https://github.com/infiniflow/ragflow/blob/c4bfd9fa2c742544488edfa512004f33130eec93/conf/llm_factories.json#L510
> 
> `OpenAI-API-Compatible` should be in `settings.FACTORY_LLM_INFOS`, shouldn't be?
yes you are right, misunderstand

PR #8502, opened 2025-06-26
Fix: user_default_llm configuration doesn't work for OpenAI API compatible LLM factory

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/8467
when add llm the llm_name will like "llm1___OpenAI-API" https://github.com/infiniflow/ragflow/blob/f09ca8e79500c0c99ba9703ccbc86965ef05de02/api/apps/llm_app.py#L173
so we should not use llm1 to query

### Type of change


Woody-Hu:
Hi @KevinHuSh could you help to review again? thanks

api/db/services/llm_service.py
```diff
@@ -45,6 +45,9 @@ def get_api_key(cls, tenant_id, model_name):
             objs = cls.query(tenant_id=tenant_id, llm_name=mdlnm)
         else:
             objs = cls.query(tenant_id=tenant_id, llm_name=mdlnm, llm_factory=fid)
+
+        if not objs and fid:
```
KevinHuSh: It's equal to line 47, isn't it?

Woody-Hu: nope for this case, I use a different parameter model_name not ll_name when add_llm this part will be the user input. @KevinHuSh

Woody-Hu:
@KevinHuSh I use another way to handle llm_name (keep as the changes when add_llm)

PR #8526, opened 2025-06-27
Adding semaphore usage on the '/run' endpoint 

RafaelFFAumo:

Switching threading.Lock() to asyncio.Lock(), since threading.Lock() is blocking.

### Type of change

- [x] Performance Improvement


yongtenglei:
Hi @RafaelFFAumo 

I would like to say what if we only keep `Switching threading.Lock() to asyncio.Lock(), since threading.Lock() is blocking` part to make this module simple. What do you think?

The limiter can be configured in `.env` file. So it's better to keep it.

I don't expected to have dramatic change or additions (only for this module) unless we really have this requirement as stable and easy to troubleshoot is the first priority.

Thanks for your understanding!

RafaelFFAumo:
no problem @yongtenglei, I've thought that adding a websocket implementation would be good too. I didnt see that the limiter was configurable, sorry on that one.

yongtenglei:
Hi, @RafaelFFAumo 

I agree... The websocket implementation is actually a separate endpoint alongside `/run`, so that it may not bring a big problem. However I didn't see the usage case for websocket for now.

If you don't mind, could you leave it this time. And make another PR to bring it out here with supplement documentation in README? So I can review it more smoothly. 

Thank you!

RafaelFFAumo:
@yongtenglei  I reverted the websocket implementation and updated the PR description

yongtenglei (approved):
Thank you! All tests passed, it's ready for merge.

Just one small thing: please avoid removing the trailing newline at the end of files, as many standards recommend keeping it.

RafaelFFAumo:
I was reading through the smartphone and missclicked two times haha. But thank you @yongtenglei  for the opportunity of helping you guys out.

Issue #8602, opened 2025-07-01
[Bug]: Error occurred when start document parsing in a API created dataset


CamWang:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

6b3a40b

### RAGFlow image version

v0.19.1 full

### Other environment information

```Markdown
* Pyhton Version: 3.10.12
* Docker Image: infiniflow/ragflow:v0.19.1
```

### Actual behavior

When invoking start parsing action on PDF files in a dataset that is created by API call, system will instantly throw an NoneType error as shown below.

![Image](https://github.com/user-attachments/assets/57af3d5d-e59d-49d9-b9e9-9806b0c8bd0c)

### Expected behavior

Start parse the document. Parse state turning into progress indicator.

### Steps to reproduce

```Markdown
* Calling "Create Dataset" API without defining task_page_size since it's not being labeled as required in the API doc
* Upload a document to this dataset
* Start parsing
```

### Additional information

HTTP request used to create dataset:
```
{
    "name": "nature",
    "description": "some knowledge",
    "embedding_model": "qwen3-embedding-8b___VLLM@VLLM",
    "parser_config": {
        "auto_keywords": 8,
        "auto_questions": 3,
        "chunk_token_num": 300,
        "html4excel": true
    }
}
```
Created dataset entity returned by list dataset API
```
{
            "avatar": "",
            "chunk_count": 0,
            "chunk_method": "naive",
            "create_date": "Tue, 01 Jul 2025 14:09:58 GMT",
            "create_time": 1751350198973,
            "created_by": "fbdd58fc525511f0a4840242ac1c0006",
            "description": "科普材料知识库",
            "document_count": 0,
            "embedding_model": "qwen3-embedding-8b___VLLM@VLLM",
            "id": "045f9cf2564211f0bbb40242ac1c0006",
            "language": "English",
            "name": "科普",
            "pagerank": 0,
            "parser_config": {
                "auto_keywords": 8,
                "auto_questions": 3,
[...]

PR #8603, opened 2025-07-01
Fix: page_size can be None error

CamWang:

Issue #8602

`parser_config.task_page_size` can be defaults to `None` when dataset is created by API. This was not handled by the `task_executor.py` code thus `page_size` could sometimes be `None` which will cause issue in line 351.

### Type of change


CamWang:
Embedding process will also be interrupt by this issue, will fix this and then do pull request
image

CamWang:
Oh, just found out it's causing by a similar issue addressed in a recent commit: #8407

Issue #3221, opened 2024-11-05
[Feature Request]: Revamp user interface


cike8899:
### Is there an existing issue for the same feature request?

- [X] I have checked the existing issues.

### Describe the feature you'd like

Revamp user interface

PR #8676, opened 2025-07-04
fix: optimize the chunk result page

dcc123456:
fix: Create a new message component to replace the antd message component, create a new Spin component to replace the antd Spin component, optimize the original paging component style, and optimize the chunk result page[ #3221](https://github.com/infiniflow/ragflow/issues/3221)

### Type of change


web/src/hooks/chunk-hooks.ts
cike8899, line 1: Do not modify the old version of the code.
Please modify use-chunk-request.ts.

dcc123456: Ok，I've fix it

cike8899 (approved):

PR #8689, opened 2025-07-06
Update package.json

feeingHe:
fix bug :
When installing dependencies using yarn(version 1.22.22), the following error occurs on startup: unable to locate '/../web/node_modules/monaco-editor/min/vs' glob

### Type of change


KevinHuSh:
Appreciations!
We typically compile it with `npm`.

feeingHe:
The installation logic of npm and yarn is different. I have tested this modification, and it is compatible with both npm and yarn.

KevinHuSh:
Thanks again.

PR #8713, opened 2025-07-07
fix: 1.Fix the issue in retrieval_test where the previous pagination …

feeingHe:
…causes blank results; 2. optimize the reranking warning messages to be accurate.

### Type of change


KevinHuSh:
Appreciations!
About the re-rank model, the elapsed time differs much with and without it.

feeingHe:
You look completely clueless!!!

Issue #8733, opened 2025-07-08
[Bug]: Raptor usage for embedding resumes is blocking Ollama


arturo-air:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

887651e

### RAGFlow image version

0.19.1

### Other environment information

```Markdown
OS: Linux
GPU: NVIDIA GeForce RTX 4090
```

### Actual behavior

The embedding process gets blocked if I use Raptor. RAGFlow includes either `"keep_alive": 0` or `"keep_alive": null` in the HTTP call, but there are tons of documents already to be processed.
This behavior provokes that Ollama tries to unload the model from memory multiple times, and because there are tons of calls some race condition blocks Ollama.

### Expected behavior

RAGFlow http request to Ollama should not include `keep_alive` parameter. So Ollama won't try to unload the model after every single request.

### Steps to reproduce

```Markdown
- Include Ollama as a embedding and chat provider.
- Use `nomic-embed-text:latest` as embedding model.
- Use `gemma3:27b` as chat model (vision is activated).
- Go to **Knowledge Base** and create a new one. 
- Choose `nomic` as embedding model, and activate Raptor (all the rest is the default setting).
- Add ~50 files to embed, and parse them, all at the same time.

After 10-20 files embedded, your Ollama will find the race condition because RAGFlow is sending `keep_alive` value to 0.
I have tried with other chat models, not only gemma3, and same thing happens.
```

PR #8734, opened 2025-07-09
Fix: Change Ollama Embedding Keep Alive

Woody-Hu:
https://github.com/infiniflow/ragflow/issues/8733

### Type of change


arturo-air:
The problem was really the chat calls, not the embedding calls. 

I can embed thousands of files using Ollama embedding endpoints, and it works like a charm. But when I activate Raptor (which uses `/api/chat` endpoint) to make resumes before embedding, then is when the problems come.

PR #8754, opened 2025-07-09
fix redundancy package import.

zhanluxianshen:

### Type of change

- [ ] Documentation Update
- [x] Refactoring
- [x] Performance Improvement


KevinHuSh:
Appreciations!
We did not quite follow what this PR is all about.
BTW, once `pyproject.yaml` is changed, `uv.lock` should be updated at the same time.

zhanluxianshen:
> Appreciations! We did not quite follow what this PR is all about. BTW, once `pyproject.yaml` is changed, `uv.lock` should be updated at the same time.

@KennyDizi 
This pr is fix duplicate pkg import.
and uv sync has resolved the duplicate problem .
So i think uv.lock is ok for current fix.

PR #8770, opened 2025-07-10
feat(elasticsearch): add configurable timeout and retry mechanisms

cutiechi:

Currently, the retry attempts and timeout values for Elasticsearch (ES) requests are not configurable, and some APIs—such as exists—use an unreasonable default timeout of 600 seconds.

In the current implementation, the number of retries and the timeout duration for ES requests are hardcoded and cannot be adjusted through configuration. This means that all API operations, regardless of their nature, use the same default values. For example, the exists API, which is typically a quick check, also uses the default 600-second timeout. This is not appropriate, as such a long timeout is excessive for simple existence checks and can lead to inefficient resource usage and delayed error handling.

Making these parameters configurable allows for more fine-grained control, so each API can have retry and timeout settings that match its expected behavior and performance requirements.

### Type of change

- [x] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/conf/service_conf.yaml
+++ b/conf/service_conf.yaml
@@ -18,6 +18,26 @@ es:
   hosts: 'http://localhost:1200'
   username: 'elastic'
   password: 'infini_rag_flow'
+  retry_attempts:
+    global: 2
+    # connect: 5
+    # exists: 2
+    # search: 3
+    # get: 2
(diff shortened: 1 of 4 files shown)


KevinHuSh:
Appreciations!

PR #8786, opened 2025-07-10
fix: fix dataset-page's bugs

dcc123456:

 fix dataset-page's bugs,Input component supports icon, added Radio component, and removed antd from chunk-result-bar page [#3221 ](https://github.com/infiniflow/ragflow/issues/3221)

### Type of change


web/src/hooks/chunk-hooks.ts
cike8899, line 1: Modify use-chunk-request.ts, the previous xx-hooks remain unchanged, otherwise it may affect the functions of the old version.

cike8899 (approved):

PR #8794, opened 2025-07-11
BugFix:  AgentCreateBUGFix：required argument are missing: dsl;

yungongzi:
### AgentCreateBUGFix
Because useFetchFlowTemplates is called both in the hooks and the AgentTemplateModal, and the ID of the empty template is generated via uuid, there may be cases where the IDs do not match.

Report a BUG as follows:
Prompt: 101
Required argument is missing: dsl;

52d79682-4e50-4863-8486-f1e154003043

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/web/src/pages/flow/list/agent-template-modal.tsx
+++ b/web/src/pages/flow/list/agent-template-modal.tsx
@@ -1,6 +1,7 @@
 import { IModalManagerChildrenProps } from '@/components/modal-manager';
 import { useSetModalState, useTranslate } from '@/hooks/common-hooks';
-import { useFetchFlowTemplates } from '@/hooks/flow-hooks';
+import { IFlowTemplate } from '@/interfaces/database/flow';
+// import { useFetchFlowTemplates } from '@/hooks/flow-hooks';
 import { useSelectItem } from '@/hooks/logic-hooks';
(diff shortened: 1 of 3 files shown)


asiroliu (approved):

asiroliu:
@Woody-Hu 
I've tested your latest code and confirmed the issue is resolved.

@cike8899 
Could you please review the changes?

cike8899:
First of all thanks for the fix, but I think it makes more sense to pass the dsl to onOk in AgentTemplateModal instead of calling useFetchFlowTemplates in useSaveFlow. @yungongzi

yungongzi:
> First of all thanks for the fix, but I think it makes more sense to pass the dsl to onOk in AgentTemplateModal instead of calling useFetchFlowTemplates in useSaveFlow. @yungongzi

I'm not sure if this is the optimal solution at present. I've considered the following approaches: 
1) Set the ID of the blank Agent in useFetchFlowTemplates to a fixed value, such as "blank01"; 
2) Add the blank Agent template from the database table instead of adding it on the frontend. 
If there is a more appropriate method, please feel free to correct it directly.

PR #8830, opened 2025-07-14
Fix: Remove antd  from dataset-page

dcc123456:

remove antd  from dataset-page  [#3221](https://github.com/infiniflow/ragflow/issues/3221)
### Type of change


cike8899:
The first letter of the commit message must be capitalized.

web/src/components/ui/modal.tsx
cike8899, line 1: Please use ui/dialog.tsx

PR #8886, opened 2025-07-16
fix s3 init config .

zhanluxianshen:

when``` if 'signature_version' in self.s3_config:``` and ```if 'addressing_style' in self.s3_config:``` both true.
the config init is error, will be overwrite by last one. 

this pr is for fix that case.

### Type of change

### DIFF
--- a/rag/utils/s3_conn.py
+++ b/rag/utils/s3_conn.py
@@ -66,6 +66,7 @@ class RAGFlowS3:
 
         try:
             s3_params = {}
+            config_kwargs = {}
             # if not set ak/sk, boto3 s3 client would try several ways to do the authentication
             # see doc: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials
             if self.access_key and self.secret_key:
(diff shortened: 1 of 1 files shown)


sggene:
S3 Connection Bug - Incompatible Parameter Passing to boto3.client()

**Description:**

There's a bug in the S3_Conn code where signature_version and addressing_style are directly passed as parameters to boto3.client(). However, boto3.client() does not support these parameters being passed directly, which causes a TypeError.

**Error Message:**
```
ERROR:root:Fail to connect at region None or endpoint https://cos.na-siliconvalley.myqcloud.com
  File "/ragflow/rag/utils/s3_conn.py", line 90, in __open__
    self.conn = [boto3.client('s3', **s3_params)]
  File "/ragflow/.venv/lib/python3.10/site-packages/boto3/__init__.py", line 92, in client
    return _get_default_session().client(*args, **kwargs)
TypeError: Session.client() got an unexpected keyword argument 'signature_version'
```

**Working Old Version Reference:**

The old version worked correctly with this approach:
```
if self.region in self.s3_config:
    s3_params['region_name'] = self.region
if 'endpoint_url' in self.s3_config:
    s3_params['endpoint_url'] = self.endpoint_url
if 'signature_version' in self.s3_config:
    s3_params['config'] = Config(s3={"signature_version": self.signature_version})
if 'addressing_style' in self.s3_config:
    s3_params['config'] = Config(s3={"addressing_style": self.addressing_style})
self.conn = boto3.client('s3', **s3_params)
```

I use Tencent's COS in S3-compatible mode.

Issue #8941, opened 2025-07-21
[Bug]: in the knowledge base operation file will result in an error.


OneKe:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

1

### RAGFlow image version

0.19.1

### Other environment information

```Markdown
os type:Ubuntu 24.04 LTS
docker
```

### Actual behavior

When adding new files or folders in "File Management" and then linking to a knowledge base, if the files are subsequently deleted, they can still be seen in the knowledge base, but any operation on them will result in an error.

<img width="3169" height="111" alt="Image" src="https://github.com/user-attachments/assets/a40ac26a-9611-40ed-83a1-9ec67fa14d85" />

### Expected behavior

success without error or file is deleted

### Steps to reproduce

```Markdown
When adding new files or folders in "File Management" and then linking to a knowledge base, if the files are subsequently deleted, they can still be seen in the knowledge base, but any operation on them will result in an error.
```

### Additional information

<img width="3169" height="111" alt="Image" src="https://github.com/user-attachments/assets/52a6b775-d942-4c3d-ab4e-b6826d6bc518" />

PR #8962, opened 2025-07-22
Fix:in the knowledge base operation file will result in an error

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/8941
### Type of change


asiroliu:
@Woody-Hu 
After linking a directory to Knowledge Base via "File Management" in the latest code version, the Knowledge Base fails to open. However, neither backend nor frontend shows any error logs for this issue.


image

20250722170142_rec_

asiroliu:
@Woody-Hu 
After carefully reviewing your implementation and tests, your modification only makes the documents invisible to the Dataset.

However, the invalid data still remains in the MySQL tables:
- Document
- File
- File2Document

The solution should follow the same [deletion process](https://github.com/infiniflow/ragflow/blob/main/api/apps/document_app.py#L350) as files uploaded via the Dataset page:
1. Synchronously delete records from:
   - Document table
   - File table
   - File2Document table
2. For parsed files' impact on Knowledgebase table records:
   - doc_num
   - chunk_num
3. For parsed files:
   - The chunk vectors stored in DOC_ENGINE (es/infinity) must be deleted

Woody-Hu:
> @Woody-Hu After carefully reviewing your implementation and tests, your modification only makes the documents invisible to the Dataset.
> 
> However, the invalid data still remains in the MySQL tables:
> 
> * Document
> * File
> * File2Document
> 
> The solution should follow the same [deletion process](https://github.com/infiniflow/ragflow/blob/main/api/apps/document_app.py#L350) as files uploaded via the Dataset page:
> 
> 1. Synchronously delete records from:
>    
>    * Document table
>    * File table
>    * File2Document table
> 2. For parsed files' impact on Knowledgebase table records:
>    
>    * doc_num
>    * chunk_num
> 3. For parsed files:
>    
>    * The chunk vectors stored in DOC_ENGINE (es/infinity) must be deleted

Hi @asiroliu I use this implementation to achieve better backward compatibility for existing data, but if your team thinks this is fine, I can change to the approach which you point out . But I think add both logics is a better solution.

asiroliu:
@KevinHuSh 
What would you suggest as the solution?

PR #8991, opened 2025-07-23
fix: use consistent filenames for chrome & chromedriver

gnowland:

PR #8665 updated chrome and chromedriver sources, removing the appended version number. This PR resolves remaining filename inconsistencies that would cause `Dockerfile.deps` to fail to build when `--china-mirrors=False`.

### Type of change


gnowland:
@KevinHuSh would using the real filenames (including .zip extension) be preferable (like this: https://github.com/gnowland/ragflow/commit/385e5e108536ad396b9819af4d99db6a27148eee)?

KevinHuSh:
> @KevinHuSh would using the real filenames (including .zip extension) be preferable (like this: [gnowland@385e5e1](https://github.com/gnowland/ragflow/commit/385e5e108536ad396b9819af4d99db6a27148eee))?

It's fine.

PR #9001, opened 2025-07-23
docs: "Launch service from source" website docs improvement

gnowland:

* removes erroneous references to docker configs
  - `docker/service_conf.yaml.template` and `docker/entrypoint.sh` are **NOT** involved when launching RAGFlow from source, so steps referencing them are inapplicable and should be from the docs to prevent confusion
* removes unnecessary update proxy.target step
  - `proxy.target` in `web/.umirc.ts` is already `http://127.0.0.1:9380/`, making this step unnecessary
* adds "install jemalloc" step (from README)
  - brings the "launch from source" docs in alignment with README "Launch service from source for development" section
* improves formatting of code blocks

### Type of change

- [x] Documentation Update


KevinHuSh:
Appreciations!

PR #9044, opened 2025-07-25
Added infinity rank_feature support

yuzhichang:

Added infinity rank_feature support

### Type of change

### DIFF
--- a/api/apps/kb_app.py
+++ b/api/apps/kb_app.py
@@ -14,7 +14,6 @@
 #  limitations under the License.
 #
 import json
-import os
 
 from flask import request
 from flask_login import login_required, current_user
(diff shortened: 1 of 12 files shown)


docker/.env
```diff
@@ -189,3 +189,5 @@ REGISTER_ENABLED=1
 # COMPOSE_PROFILES=infinity,sandbox
 #    - For OpenSearch:
 # COMPOSE_PROFILES=opensearch,sandbox
+
+RAGFLOW_IMAGE=infiniflow/ragflow:nightly
```
asiroliu: This configuration should not be committed.

yuzhichang: reverted.

Issue #9050, opened 2025-07-25
[Feature Request]: When deploying the project in the docker cluster, the addition of the image2text model failed. Using base64 images in the inspection model would be a better choice


kuschzzp:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

When deploying the project in the docker cluster, the addition of the image2text model failed. Using base64 images in the inspection model would be a better choice

### Describe implementation you've considered

https://github.com/infiniflow/ragflow/blob/c63d12b93678627f5d2abdf4c8ad1286ec584d8e/api/apps/llm_app.py#L259-L262

The code above is recommended to be modified to

```
test_image_base64 = "【【【base64 image str】】】"
image_data = base64.b64decode(test_image_base64)
m, tc, _ = mdl.describe(image_data)

```

### Documentation, adoption, use case

```Markdown

```

ZhenhangTung:
@kuschzzp would double check your case. cc @asiroliu

asiroliu:
@kuschzzp 
Is the validation failure caused by not using base64-encoded images? 
Can you provide the name and factory of the large model you used? I will verify it.

kuschzzp:
> [@kuschzzp](https://github.com/kuschzzp) Is the validation failure caused by not using base64-encoded images? Can you provide the name and factory of the large model you used? I will verify it.

Thank you for your reply.
I deployed RAGFlow with Docker Swarm, separating the frontend and backend into distinct services.
The path `web/src/assets/yay.jpg` only works in the development environment. In production (with Docker Swarm), using this path will cause it to fail.
So verifying the model without the path is more appropriate.

PR #9052, opened 2025-07-25
Image validation of the image2text model without using local paths

kuschzzp:

#9050

### Type of change

### DIFF
--- a/api/apps/llm_app.py
+++ b/api/apps/llm_app.py
@@ -257,7 +257,7 @@ def add_llm():
         )
         try:
             image_data = base64.b64decode(test_image_base64)
-            m, tc, _ = mdl.describe(image_data)
+            m, tc = mdl.describe(image_data)
             if not m and not tc:
                 raise Exception(m)
         except Exception as e:


api/apps/llm_app.py
```diff
@@ -256,10 +256,10 @@ def apikey_json(keys):
             base_url=llm["api_base"]
         )
         try:
-            with open(os.path.join(get_project_base_directory(), "web/src/assets/yay.jpg"), "rb") as f:
-                m, tc = mdl.describe(f.read())
-                if not m and not tc:
-                    raise Exception(m)
+            image_data = base64.b64decode(test_image_base64)
+            m, tc, _ = mdl.describe(image_data)
```
asiroliu: mdl.describe(image_data) only returns 2 values, but the code is trying to unpack 3 (m, tc, _). This causes a ValueError. 

```
not enough values to unpack (expected 3, got 2)
```

kuschzzp: This issue has been fixed.
I'm very sorry for causing this issue because I modified the describe method locally, which is inconsistent with the official code.

asiroliu (approved):

PR #9097, opened 2025-07-29
Fix : API /document/thumbnails using wrong method to get ''doc_ids'

zhangwenxuanQAQ:

doc_ids is a list , should use request.args.getlist("doc_ids")

### Type of change


asiroliu:
It appears we should use the getlist() method instead.

image

asiroliu (approved):

PR #9184, opened 2025-08-02
Feat: added functionality to choose all datasets if no id is provided

Drasek:

Using the mcp server in n8n sometimes (with smaller models) results in errors because the llm misses a char or adds one to the list of dataset_ids provided. It first asks for the list of datasets and if you got a larger list of them it makes a error recalling the list completely. So adding the feature to just search through all available datasets solves this and makes the retrieval of data more stable. The functionality to just call special datasets by id is not changed, the dataset_ids are now not required anymore (only the "question" is). You can provide (like before) a list of datasets, a empty list or no list at all.

### Type of change

- [X] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/mcp/server/server.py
+++ b/mcp/server/server.py
@@ -180,7 +180,7 @@ async def list_tools(*, connector) -> list[types.Tool]:
     return [
         types.Tool(
             name="ragflow_retrieval",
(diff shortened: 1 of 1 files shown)


yongtenglei (approved):

yongtenglei:
Hi, @Drasek 

I would appreciate it if you could update the corresponding documentation in `docs/develop/mcp/` to reflect these changes. All your suggestions look good to me.

PR #9191, opened 2025-08-03
Perf Impr: Decouple reasoning and extraction for faster, more precise logic

es-zx:

This commit refactors the core prompts to decouple the high-level reasoning from the low-level information extraction. By making REASON_PROMPT a dedicated strategist that only generates search queries and re-tasking RELEVANT_EXTRACTION_PROMPT to be a specialized tool for single-fact extraction, we eliminate redundant information summarization. This clear separation of concerns makes the overall reasoning process significantly faster and more precise, as each component now has a single, well-defined responsibility.

### Type of change

- [x] Performance Improvement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,5 @@ dist
 # SvelteKit build / generate output
 .svelte-kit
 
+# Default backup dir
+backup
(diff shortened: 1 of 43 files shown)


asiroliu:
@KevinHuSh 
LGTM

image

asiroliu (approved):

Issue #9232, opened 2025-08-05
[Bug]: mismatch of history assitant message and it's reference


sinopec:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

dda5a00  

### RAGFlow image version

0.20.0

### Other environment information

```Markdown

```

### Actual behavior

In history message of session,  the assistant  message and it's reference chunk dont match .
When  view in web ,the mismatch is obvious

<img width="1670" height="304" alt="Image" src="https://github.com/user-attachments/assets/55aeb998-7e72-4a75-b68d-d01240b11661" />

### Expected behavior

the fist message should dispalay reference of other messages.

### Steps to reproduce

```Markdown
just ask two question ,one answer with reference,another without.
then reload the page to refresh from the history message
```

PR #9233, opened 2025-08-05
Fix: fix mismatch of assitant message and its reference

sinopec:

#9232

### Type of change

1.  When creating a new session, initialize an empty reference that includes both the app api and sdk API.
2. Fix the logic when retrieving references for historical messages: the number of dialogue messages and reference messages may differ, but it should match the number of assistant messages.


sinopec:
#9232

PR #9254, opened 2025-08-05
Fix: self-deployed LLM error, #9217

wingjson:

fix self-deployed LLM(OpenAI-API-Compatible)

### Type of change

- [x] Bug Fix

### DIFF
--- a/api/db/services/llm_service.py
+++ b/api/db/services/llm_service.py
@@ -227,6 +227,9 @@ class TenantLLMService(CommonService):
 
         for llm in LLMService.query(llm_name=llm_id):
             return llm.model_type
+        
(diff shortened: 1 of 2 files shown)


api/db/services/llm_service.py
```diff
@@ -227,6 +227,9 @@ def llm_id2llm_type(llm_id: str) -> str | None:
 
         for llm in LLMService.query(llm_name=llm_id):
             return llm.model_type
+
```
KevinHuSh, line 230: Appreciations!

Duplicated code.

wingjson: > Duplicated

If I add OpenAI-API llm,the information was stored in tenant _llm table

PR #9310, opened 2025-08-07
fix: Words ending with 's' are incorrectly tokenized.

liuzhenghua:

**Issue:**
English words such as has and Doris are currently incorrectly reduced to ha and dori.

**Change:**
The stemmer performs rule-based stemming without validating whether the resulting word is actually correct. In contrast, the lemmatizer uses part-of-speech information to produce accurate base forms. Since only one of them is needed, I have removed the stemmer in favor of the lemmatizer.

This update requires adding the files downloaded via nltk.download('averaged_perceptron_tagger_eng') to the Docker image, placing them under the /root/nltk_data/taggers directory.

**问题：**英文单词，例如 has, doris，目前会被错误分成ha, dori。
**修改项：**stemmer只是简单的按照正则进行词形还原，不会管还原后的词是否正确。lemmatizer则可以根据词性去进行词形还原，两者选其一就行了，所以我这里把stemmer进行了移除。

本次修改项，dockerfile 需要添加`nltk.download('averaged_perceptron_tagger_eng')`下载的文件，放到`/root/nltk_data/taggers`目录

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/rag/nlp/rag_tokenizer.py
+++ b/rag/nlp/rag_tokenizer.py
@@ -23,7 +23,8 @@ import re
 import string
 import sys
 from hanziconv import HanziConv
-from nltk import word_tokenize
+from nltk import word_tokenize, pos_tag
+from nltk.corpus import wordnet
 from nltk.stem import PorterStemmer, WordNetLemmatizer
 from api.utils.file_utils import get_project_base_directory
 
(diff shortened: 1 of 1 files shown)


whhe:
Besides `Dockerfile`, I'm not sure about `Dockerfile.scratch.oc9`, but `download_deps.py` also needs to be updated.
And the code comments here should be in English, which need to be consistent with other places.

And I have another question, what would happen in `tokenize` if all other languages like Japanese and French were treated as English?

liuzhenghua:
> Besides `Dockerfile`, I'm not sure about `Dockerfile.scratch.oc9`, but `download_deps.py` also needs to be updated. And the code comments here should be in English, which need to be consistent with other places.
> 
> And I have another question, what would happen in `tokenize` if all other languages like Japanese and French were treated as English?

The WordNetLemmatizer only supports the English lexicon. For other languages, it simply returns the original word. To support French, you need to use spaCy.

whhe:
> > Besides `Dockerfile`, I'm not sure about `Dockerfile.scratch.oc9`, but `download_deps.py` also needs to be updated. And the code comments here should be in English, which need to be consistent with other places.
> > And I have another question, what would happen in `tokenize` if all other languages like Japanese and French were treated as English?
> 
> The WordNetLemmatizer only supports the English lexicon. For other languages, it simply returns the original word. To support French, you need to use spaCy.

Thanks for your reply. So in fact, the current cross language search only supports Chinese and English, right? It's a bit different from what I expected before.

KevinHuSh:
Appreciations!

PR #9376, opened 2025-08-11
fix: Replace StrEnum with strenum in code_exec.py

asiroliu:

- The enum import was changed from Python's built-in StrEnum to the strenum package.
- Fix error `Warning: Failed to import module code_exec: cannot import name 'StrEnum' from 'enum' (/usr/lib/python3.10/enum.py)`

### Type of change


asiroliu:
image

PR #9404, opened 2025-08-12
Feat: allows setting multiple types of default models in service config

yongtenglei:

Allows set multiple types of default models in service config.

### Type of change


kkxue:
@yongtenglei That's exactly what I need. thx!

kkxue:
@yongtenglei There must be a bug. see https://github.com/infiniflow/ragflow/issues/9426

PR #9425, opened 2025-08-12
Attachments only chat

shakeel-mushtaq:
Attachments only chat and retrieval

### DIFF
--- a/api/apps/dialog_app.py
+++ b/api/apps/dialog_app.py
@@ -51,6 +51,7 @@ def set_dialog():
     similarity_threshold = req.get("similarity_threshold", 0.1)
     vector_similarity_weight = req.get("vector_similarity_weight", 0.3)
     llm_setting = req.get("llm_setting", {})
+    meta_data_filter = req.get("meta_data_filter", {})
     prompt_config = req["prompt_config"]
 
     if not is_create:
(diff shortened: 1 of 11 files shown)


shakeel-mushtaq:
Apology for unwanted code block formatting

KevinHuSh:
Appreciations!
How come **`kbinfos`** would include docs which are not exist in `attachments`?

shakeel-mushtaq:
When doc_ids are passed in retriever.retrieval(...) it returns chunks including other docs (seems its default behavior or some bug for infinity?) But we want chunks to be restricted only for given doc_ids. (Imagine customer want the answer not from full kb but from give attachments only) So .. to not effect default functionality we filter it out side  retriever.retrieval(...)  based on flag **attachments_only_retrieval** as indexing search is usually fast and filtering from 10 or more chunks should not be overhead. Is there any other better way  to do it? We actually want to get retrievals only for attachments or mentioned doc_ids only. Here is the api we are calling 

{{ragflow_server}}/api/v1/chats/{{chat_id}}/completions

```
  {
  "question": "How many leaves allowed in Lycus",
  "session_id": "3a9a0140774e11f0a9f72d67c4cef769",         
  "doc_ids" : "6de2ad6e77a711f09155ee5577ac8ca6",
   "stream": false                  
}
```

KevinHuSh:
Appreciations!
By our evalutaions, this feature is not going to be merged into main branch

PR #9453, opened 2025-08-13
Feat:to create a option let agent has independent histoy

Popmio:

when create mutiagent in a workflow ,In the second round and subsequent dialogues, the agent will default to using the content output from the last message, which can sometimes affect the quality of generation. Therefore, an option should be established for the agent to only read the historical information from its own.

### Type of change
- [x] New Feature

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -74,6 +74,7 @@ class Canvas:
     def __init__(self, dsl: str, tenant_id=None, task_id=None):
         self.path = []
         self.history = []
+        self.agent_histories = {}
         self.components = {}
         self.error = ""
         self.globals = {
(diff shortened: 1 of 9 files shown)


Popmio:
sorry i didnt resolve the conflicts in time

yingfeng:
This is a workaround, regarding to agent memory, it requires short-term/long-term interfaces as well as consolidation between them. Additional, it also requires to store task state to be used for meta store for each agent, such as plan modified through  human-in-the-loop.

Popmio:
Thank you for your reply. I didn't consider it comprehensively. I will reconsider this requirement.

yingfeng:
You can wait for our updates. Maybe in 0.22

PR #9454, opened 2025-08-13
Docs: Update HTTP API reference with simplified response format and parameters

asiroliu:

- Make `session_id` optional and add `inputs` parameter
- Remove deprecated `sync_dsl` parameter
- Update request/response examples to match current API behavior

### Type of change

- [x] Documentation Update


writinwaters (approved):
LGTM

PR #9464, opened 2025-08-13
Docs: Deprecated 'Create session with agent'

writinwaters:
### Type of change

- [x] Documentation Update

### DIFF
--- a/agent/templates/sql_assistant.json
+++ b/agent/templates/sql_assistant.json
@@ -206,7 +206,7 @@
                             "enablePrologue": true,
                             "inputs": {},
                             "mode": "conversational",
-                            "prologue": "Hi! I'm your SQL assistant, what can I do for you?"
+                            "prologue": "Hi! I'm your SQL assistant. What can I do for you?"
                         }
                     },
                     "upstream": []
@@ -319,7 +319,7 @@
                                 "enablePrologue": true,
                                 "inputs": {},
                                 "mode": "conversational",
-                                "prologue": "Hi! I'm your SQL assistant, what can I do for you?"
+                                "prologue": "Hi! I'm your SQL assistant. What can I do for you?"
                             },
                             "label": "Begin",
                             "name": "begin"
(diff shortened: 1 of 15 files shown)


a-marchetti-alientech:
Hi @writinwaters!

1) Thanks.

2) I'm implementing the changes for the 0.20.1 (coming from a 0.19.0). I don't understand how I can handle a situation where I have input parameters of various types including the file type. There's no documentation I could find to handle this case. 
With the 0.19.0 I was creating the session with the files as form multipart and the other inputs as the json body of the HTTP request. Now the create session doesn't set the input values anymore, and I'm switching to initialize a session direclty by chatting the agent. Yet, this works for inputs that are not files only. How can I handle file types too? The new docs to not mention the file type.

writinwaters:
> Hi @writinwaters!
> 
> 1. Thanks.
> 2. I'm implementing the changes for the 0.20.1 (coming from a 0.19.0). I don't understand how I can handle a situation where I have input parameters of various types including the file type. There's no documentation I could find to handle this case.
>    With the 0.19.0 I was creating the session with the files as form multipart and the other inputs as the json body of the HTTP request. Now the create session doesn't set the input values anymore, and I'm switching to initialize a session direclty by chatting the agent. Yet, this works for inputs that are not files only. How can I handle file types too? The new docs to not mention the file type.

Hi @a-marchetti-alientech, Thanks for your message. Passing in a file type parameter to an agent via an HTTP request is not yet supported in v0.20.1. This is an known issue. We'll figure this out and plan on deliver this feature in v0.21.0, scheduled for early to mid-September. Hopefully this works for your timeline..

a-marchetti-alientech:
@writinwaters 
I'm facing another problem. The opening greeting is lost when I directly chat the agent.

writinwaters:
> @writinwaters I'm facing another problem. The opening greeting is lost when I directly chat the agent.

Hi, @a-marchetti-alientech  Apologies for the late response. This bug fix will be included in v0.20.2.

PR #9488, opened 2025-08-15
Refactor:For XinferenceChat HuggingFaceChat ModelScopeChat append v1 only when not end with v1

Woody-Hu:

For XinferenceChat HuggingFaceChat ModelScopeChat append v1 only when not end with v1

### Type of change

- [x] Refactoring


KevinHuSh:
Appreciations!

image

It seems `urljoin` can adapt suffix duplication stuff.

PR #9533, opened 2025-08-18
Fix: search service reference

asiroliu:

- Update search_app.py to use SearchService instead of KnowledgebaseService for duplicate

### Type of change


asiroliu:
- Before Fix:​
Renaming occurred when search namematched any KB name

- After Fix:​​
Renaming now only triggers when search nameduplicates an existing search name

Issue #9537, opened 2025-08-18
[Bug]: English tokens extraction error when word ends with "e"


whhe:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

9b026fc

### RAGFlow image version

infiniflow/ragflow:v0.20.1-slim

### Other environment information

```Markdown

```

### Actual behavior

When using cross language search with 'Chinese' and 'English', the words like `database` and `advantage` are converted to `databas` and `advantag`, which are missing the letter 'e' at the end.

### Steps to reproduce

```Markdown
Use 'Chinese' and 'English' in cross language searching, and text can be 'What is the advantage of OceanBase', we can got the query dsl at debug log of es_conn，which contains these tokens.
```

### Additional information

https://stackoverflow.com/questions/24517722/how-to-stop-nltk-stemmer-from-removing-the-trailing-e

PR #9540, opened 2025-08-19
Fix: Cross Language Just Use Not Chinese Pattern

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/9537 based on my understanding of cod,e should from the different pattern to handle question

### Type of change

### DIFF
--- a/agent/tools/retrieval.py
+++ b/agent/tools/retrieval.py
@@ -111,8 +111,10 @@ class Retrieval(ToolBase, ABC):
         vars = self.get_input_elements_from_text(kwargs["query"])
         vars = {k:o["value"] for k,o in vars.items()}
         query = self.string_format(kwargs["query"], vars)
+        cross_language_query = False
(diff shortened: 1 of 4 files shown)


Woody-Hu:
Hi @whhe do you have a chance to verify this PR?

whhe:
After some debugging, I found that the issue is caused by `_step5a` in the `stem` method of `PorterStemmer`, which is called at https://github.com/infiniflow/ragflow/blob/main/rag/nlp/rag_tokenizer.py#L331

I haven't deployed your branch to verify it yet, but I guess there may still be problems. I will try it if I have time in two days.

whhe:
I found a related question, I think we should deal with it by change the `RagTokenizer.tokenize` method https://stackoverflow.com/questions/24517722/how-to-stop-nltk-stemmer-from-removing-the-trailing-e

PR #9546, opened 2025-08-19
Fix: English tokens extraction error when word ends with "e"

Woody-Hu:
https://github.com/infiniflow/ragflow/issues/9537
https://github.com/infiniflow/ragflow/pull/9540

### Type of change


whhe:
I found there is a more universal solution https://github.com/infiniflow/ragflow/pull/9310, maybe it's better to upgrade the tokennizer based on that patch.

PR #9568, opened 2025-08-19
Backend expect auto not automatic

apps-lycusinc:
### Type of change


apps-lycusinc:
dialog_service.py is reffered backend component

apps-lycusinc:
Seems change not required

PR #9576, opened 2025-08-20
Refactor: based on async await to handle Redis when raptor

Woody-Hu:

based on async await to handle Redis when raptor

### Type of change

- [x] Refactoring
- [x] Performance Improvement


KevinHuSh:
Appreciations!
Could you elaborate on the reason?

Woody-Hu:
> Appreciations! Could you elaborate on the reason?
I want to base on async await to improve the IO time usage from Redis as we already did for LLM chat, is there any cases I missed?

PR #9677, opened 2025-08-23
Fix: Wrap VersionDialog in DropdownProvider for proper context

uciadonis:

The VersionDialog component was not receiving the correct context for dropdown handling, causing improper behavior in its interactions.
This PR wraps VersionDialog in DropdownProvider to ensure it gets the proper context and functions as expected.

### Type of change


dcc123456:
This pr solves the problem of Agent VersionDialog opening exception and can be merged

PR #9694, opened 2025-08-25
Update routes.ts

RuyXu:
Change redirect target from /knowledge to /

After logging in, the user should be directed to the homepage instead of the knowledge base page. The redirect should go to `/` rather than `/knowledge`. I believe this is a bug.

### Type of change


cike8899:
The home route is now modified to go to the old knowledge base page, and the / route goes to the new root page.  @RuyXu

KevinHuSh:
Appreciations!

PR #9730, opened 2025-08-26
Feature/update workflow

YyBoom233:
Updated the Chinese translation of the workflow case
### Type of change


KevinHuSh:
description -> description_lang

PR #9900, opened 2025-09-03
During the chat, the assistant's response cited documents outside current chat's kbs

GrubbyLee:

During the chat, the assistant's response cited documents outside the current knowledge base。

### Type of change


GrubbyLee:
During the chat with the assistant, when RAGFlow uses the `use_sql` function for querying, the query scope differs from that of the ordinary query method. The `use_sql` function uses the `tenant_id` as a parameter, meaning it queries the index data of the entire tenant. This leads to the discrepancy.

PR #9984, opened 2025-09-08
Support baidu vdb as doc engine

letterbeezps:

support use baidu vector database as doc engine

### Type of change


letterbeezps:
@KevinHuSh  Hello, do you have time to review this code? BaiduVDB now supports field weight assignments.

yingfeng:
@letterbeezps Thanks for your contribution.
There are some guidelines to integrate new doc engine into RAGFlow:
1. Make sure the new doc engine can fully support the requirements of RAGFlow.
From your implementation, it seems the term weight has not been applied,yet. Can you please further integrate these features from baidu VDB?
2. Can you continuously maintain the baidu VDB as the doc engine according to RAGFlow's evolution? We don't have corresponding resources to fully test each doc engine, without continuous update, the doc engine would have to be abandoned due to being lack of new features.  

After these two issues can be resolved, we will merge it, thank you so much!

rag/utils/baidu_vdb_conn.py
```diff
@@ -0,0 +1,653 @@
...
+            endpoint=settings.BAIDUVDB['endpoint'])
+        self.client = pymochow.MochowClient(config)
+        fp_mapping = os.path.join(
+            get_project_base_directory(), "conf", "mochow_mapping.json"
+        )
+        if not os.path.exists(fp_mapping):
+            raise Exception(f"Mapping file not found at {fp_mapping}")
+        self.mapping = json.load(open(fp_mapping))
+        healthy = self.health()
+        self.query_fields_boosts = {
```
letterbeezps, line 78: @yingfeng the term weight is applied at here, This syntax is a little different from es, it acts directly on the query value rather than the field

letterbeezps:
For the promotion of Baidu VDB，our team will maintain the baidu VDB as the doc engine for RAGFlow's new features

rag/utils/baidu_vdb_conn.py
KevinHuSh: Is this field weight instead of term weight?

letterbeezps: Sorry, I'm a little confused. Can you point out where the term weight is used in "es_conn.py"！ so I can check and compare. @KevinHuSh

rag/nlp/query.py
```diff
@@ -216,6 +227,84 @@ def need_fine_grained_tokenize(tk):
...
+        for tt in self.tw.split(txt)[:256]:  # .split():
+            if not tt:
+                continue
+            keywords.append(tt)
+            twts = self.tw.weights([tt])
+            syns = self.syn.lookup(tt)
+            if syns and len(keywords) < 32:
+                keywords.extend(syns)
+            logging.debug(json.dumps(twts, ensure_ascii=False))
+            for tk, w in sorted(twts, key=lambda x: x[1] * -1):
```
KevinHuSh, line 274: The `w` is the weight of the term, which has been dropped/neglected.

letterbeezps: OK, got it, I'll check it out. thanks

yingfeng:
Given that this PR has remained inactive for an extended period, and considering that RAGFlow's retrieval engine has since introduced numerous enhancements—including memory capabilities, metadata handling, and other features—this PR will now be closed.

PR #10126, opened 2025-09-17
Fixed the issue where database connections were interrupted under high concurrency

lemsn:

Fixed the issue where database connections were interrupted under high concurrency

### Type of change

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -278,7 +278,7 @@ class RetryingPooledMySQLDatabase(PooledMySQLDatabase):
         # self.connect()
         try:
             self.close()
-        except:
+        except Exception:
             pass
         try:
             self.connect()


api/db/db_models.py
```diff
@@ -299,7 +329,19 @@ class BaseDataBase:
...
-        self.database_connection = PooledDatabase[settings.DATABASE_TYPE.upper()].value(db_name, **database_config)
+        
+        pool_config = {
+            'max_connections': 1000,
+            'stale_timeout': 300,
+            'timeout': 10,
+            'max_retries': 5,
+            'retry_delay': 1,
+        }
+        database_config.update(pool_config)
```
asiroliu, line 334: This will force the database configuration to use fixed values, making any modifications in the configuration file ineffective.

lemsn: yes. i was fixed it

PR #10164, opened 2025-09-19
Fix: Re-parsing the document via the API call does not correctly upda…

qiuyi21:
…te the Knowledgebase.chunk_num

Re-parsing the document via the API call does not correctly update the Knowledgebase.chunk_num

### Type of change


KevinHuSh:
Appreciation!
I don't think we should reduce chunk numbers for KB after re-parsing done.

qiuyi21:
clear_chunk_num_when_rerun

> Appreciation! I don't think we should reduce chunk numbers for KB after re-parsing done.

But click parse on Web UI is send request to http://<ragflow_server>/v1/document/run, the handler in document_app.py will call clear_chunk_num_when_rerun too.

qiuyi21:
kb chunk_num will continue to accumulate when call parse via api, it's illogical.

PR #10211, opened 2025-09-22
Fix: string merge bug about agent TextProcessing. (\m)

zhimeng123:
An error occurred while merging strings containing '\m' in the Text Processing function of the agent.

Convert \ m to m using regular expressions

From my example alone, it doesn't affect the original meaning, it's still math

image

### Type of change

- [ √ ] Bug Fix (non-breaking change which fixes an issue)


zhimeng123:
What should I do?  Continue to waiting? 
   [The same pull](https://github.com/infiniflow/ragflow/pull/7909)

Issue #10235, opened 2025-09-23
[Question]: max_tokens must be at least 1, got -950, BadRequestError


apps-lycusinc:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

When parsing file with huge text with raptor enabled raptor fails -> [Fbr_IncomeTaxOrdinance_2001-amended-upto30.06.2024.pdf](https://github.com/user-attachments/files/22487679/Fbr_IncomeTaxOrdinance_2001-amended-upto30.06.2024.pdf)

Here is the error from LM

[ERROR][Exception]: Exceptions from Trio nursery (5 sub-exceptions) -- **ERROR**: INVALID_REQUEST - Error code: 400 - {'object': 'error', 'message': 'max_tokens must be at least 1, got -950.', 'type': 'BadRequestError', 'param': None, 'code': 400}

Chat model is : mistralai/Mistral-7B-Instruct-v0.3

Here is VLLM start command

vllm serve mistralai/Mistral-7B-Instruct-v0.3 \
  --host 0.0.0.0 \
  --port 8000 \
  --api-key YOUR_SECRET \
  --trust-remote-code \
  --tokenizer-mode mistral \
  --load-format mistral \
  --config-format mistral
 --max-model-len 32768

Any suggestion to handle such file with raptor and files with more text?

Llala0928:
@apps-lycusinc Resolved.

apps-lycusinc:
Issue still persists,  See below logs
14:15:16 Task has been received.
14:15:23 Page(1~13): Start to parse.
14:15:23 Page(1~13): OCR started
14:15:29 Page(1~13): Processing...
14:15:30 Page(1~13): Processing...
14:15:30 Page(1~13): OCR finished (7.08s)
14:15:52 Page(1~13): Layout analysis (21.70s)
14:15:53 Page(1~13): Table analysis (1.00s)
14:15:53 Page(1~13): Text merged (0.00s)
[...]

PR #10252, opened 2025-09-24
Fix:max_tokens must be at least 1, got -950, BadRequestError

Woody-Hu:
https://github.com/infiniflow/ragflow/issues/10235

### Type of change

- [x] Refactoring


apps-lycusinc:
It seems mentioned issue [10235](https://github.com/infiniflow/ragflow/issues/10235 ) is not solved

PR #10353, opened 2025-09-29
Feat: add toc extraction method NOT FINISHED!!!

Magicbook1108:

Feat, add toc extraction method NOT FINISHED!!!

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Current issues:
**1. Excessive Token Usage with img_url Input (Line 540)**
When passing img_url formatted input, the token count explodes to an extremely large range (approximately 250,000–600,000 tokens). This suggests that the way image URLs are being processed leads to massive and unnecessary tokenization, which severely impacts performance and cost.

**2. Task Repetition with pending >= 1**
When pending >= 1, the task keeps being re-executed in a loop but never progresses to the next stage. In other words, the system detects a pending task and repeatedly retries execution, but it doesn’t move forward in the pipeline, causing a deadlock-like state.


Magicbook1108:
All issues solved

Issue #9992, opened 2025-09-09
[Bug]: onnxtime-gpu memory error


GXKIM:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow image version

from source 

### Other environment information

```Markdown
OS type：utbuntu
ldd (Ubuntu GLIBC 2.31-0ubuntu9.14) 2.31

info:
11:27:47 Task has been received.11:27:51 Page(45~67): OCR started11:28:13 Page(45~67): [ERROR]Internal server error while chunking: Exceptions from Trio nursery (1 sub-exception)11:28:13 [ERROR][Exception]: Exceptions from Trio nursery (1 sub-exception) -- [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running Clip node. Name:'p2o.Clip.6' Status Message: /onnxruntime_src/onnxruntime/core/framework/bfc_arena.cc:376 void* onnxruntime::BFCArena::AllocateRawInternal(size_t, bool, onnxruntime::Stream*, bool, onnxruntime::WaitNotificationFn) Failed to allocate memory for requested buffer of size 44433408
```

### Actual behavior

###onnxtime-gpu 
https://github.com/microsoft/onnxruntime/issues/9509#issuecomment-951546580
https://github.com/microsoft/onnxruntime/issues/18743
###version 
onnxruntime                              1.18.0
onnxruntime-gpu                          1.18.0

This bug has been around since v1.81.1, and it still isn’t fixed.I’ve tried switching between the GPU and CPU versions, but the same issue persists.

###code

    def cuda_is_available():
        try:
            import torch
            if torch.cuda.is_available() and torch.cuda.device_count() > device_id:
                return True
        except Exception:
            return False
        return False
    options = ort.SessionOptions()
    options.enable_cpu_mem_arena = False
    options.enable_mem_pattern = False
    options.execution_mode = ort.ExecutionMode.ORT_SEQUENTIAL
    options.intra_op_num_threads = 1
[...]

PR #10407, opened 2025-10-03
Fix OCR GPU provider mem limit handling

loks666:

  - Running DeepDoc OCR on large PDFs inside the GPU docker-compose setup would intermittently fail with
    [ONNXRuntimeError] ... p2o.Clip.6 ... Available memory of 0 is smaller than requested bytes ...
  - Root cause: load_model() in deepdoc/vision/ocr.py treated device_id=None as-is.
    torch.cuda.device_count() > device_id then raised a TypeError, the helper returned False, and ONNXRuntime quietly fell back to CPUExecutionProvider with
  the hard-coded 512 MB limit, which then triggered the allocator failure.
  - Environment where this reproduces: Windows 11, AMD 5900x, 64 GB RAM, RTX 3090 (24 GB), docker-compose-gpu.yml from upstream, default DeepDoc + GraphRAG
  parser settings, ingesting heavy PDF such as 《内科学》（第10版）.pdf (~180 MB).

  Fixes:

  - Normalize device_id to 0 when it is None before calling any CUDA APIs, so the GPU path is considered available.
  - Allow configuring the CUDA provider’s memory cap via OCR_GPU_MEM_LIMIT_MB (default 2048 MB) and expose OCR_ARENA_EXTEND_STRATEGY; the calculated byte
  limit is logged to confirm the effective settings.

  After the change, ragflow_server.log shows for example
  load_model ... uses GPU (device 0, gpu_mem_limit=21474836480, arena_strategy=kNextPowerOfTwo) and the same document finishes OCR without allocator errors.

  ### Type of change

### DIFF
--- a/deepdoc/vision/ocr.py
+++ b/deepdoc/vision/ocr.py
@@ -84,7 +84,8 @@ def load_model(model_dir, nm, device_id: int | None = None):
     def cuda_is_available():
         try:
             import torch
-            if torch.cuda.is_available() and torch.cuda.device_count() > device_id:
+            target_id = 0 if device_id is None else device_id
+            if torch.cuda.is_available() and torch.cuda.device_count() > target_id:
                 return True
         except Exception:
             return False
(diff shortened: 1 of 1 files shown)


yingfeng:
Thanks for the contribution! Have you thoroughly tested this? We've tried numerous similar modifications, but they all ultimately lead to an OOM error. 
Such as issue #9992

loks666:
I have re-tested the issue, and no anomalies occurred — the file was successfully parsed. I believe I have resolved this problem. To avoid any ambiguity, I will also share my setup:

* **CPU**: AMD 5900X
* **Memory**: DDR4 64 GB
* **GPU**: RTX 3090 (24 GB)

The uploaded file was a 177 MB medical textbook. I used the **v0.20.5-slim** image and started the service with **docker-compose-gpu.yml**, configured with the **Ollama model**.

One important note: in my **Nginx** configuration, I modified the upload file size limit with:

```nginx
client_max_body_size 256M;
```

After re-slicing the PDF, there were no exceptions or errors, and the parsing completed successfully.

For your reference, I have submitted my system screenshots, related files, runtime logs, and my modified `ocr.py` to the public repository:
👉 [https://github.com/loks666/ragflow-fix-ocr-gpu-memory](https://github.com/loks666/ragflow-fix-ocr-gpu-memory)

You can review and verify them. If there are any further issues, we can discuss and work on them together.

yingfeng:
@loks666 
Please see this [response](https://github.com/infiniflow/ragflow/issues/9992#issuecomment-3388394660) 
The onnx gpu memory issue still exist. If you have any further progress, you can reply within that thread, thank you~

loks666:
OK, I'll update this issue.

PR #10471, opened 2025-10-10
Fix #10408

### LINKED ISSUE #10408: [Bug] api_key field (VARCHAR 2048) too short for Google Vertex AI Service Account Keys
## Bug Description
The `api_key` field in the `tenant_llm` table is limited to `VARCHAR(2048)`, which is too short for base64 encoded Google Cloud Service Account Keys when using Vertex AI.

## Impact
- Google Service Account Keys in Base64 format are ~3100 characters
- Keys get truncated to 2048 characters when saved
- Results in `JSONDecodeError: Unterminated string starting at: line 1 column 104`
- Makes Vertex AI integration impossible

## Reproduction
1. Add Google Cloud as LLM provider
2. Enter Service Account Key in Base64 format
3. Try to send a message
4. Error occurs because key is truncated

## Current Schema
```sql
mysql> DESC tenant_llm;
| api_key | varchar(2048) | YES | MUL | NULL | |

## Proposed fix
ALTER TABLE tenant_llm MODIFY COLUMN api_key TEXT;

glu000:

Google Cloud model does not work correctly with gemini-2.5 models
Close #10408

### Type of change

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -641,7 +641,7 @@ class TenantLLM(DataBaseModel):
     llm_factory = CharField(max_length=128, null=False, help_text="LLM factory name", index=True)
     model_type = CharField(max_length=128, null=True, help_text="LLM, Text Embedding, Image2Text, ASR", index=True)
     llm_name = CharField(max_length=128, null=True, help_text="LLM name", default="", index=True)
-    api_key = TextField(null=True, help_text="API KEY", index=True)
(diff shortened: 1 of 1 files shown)


buua436:
@glu000,Hi, could you please explain why lines 144–159 in chat_model.py were removed?

glu000:
@buua436 - Sorry, this was a mistake - i modified the PR

Issue #10474, opened 2025-10-10
[Bug]: Reasoning always enabled für Google Cloud Gemini 2.5 Flash Models


glu000:
Gemini 2.5 Flash Models use reasoning by default. There is currently no way to disable this behaviour. This leads to very long response times (> 1min). The default behaviour should be, that reasoning is disabled and configurable

PR #10479, opened 2025-10-11
Fix: Close thinking for gemin chat

Woody-Hu:

https://github.com/infiniflow/ragflow/issues/10474

### Type of change


KevinHuSh:
What do you think about this PR #10477 ?

PR #10528, opened 2025-10-13
Fix: empty references in agent conversation

buua436:
issue:
#10495
change:
fix empty references in agent conversation

### Type of change

### DIFF
--- a/rag/prompts/generator.py
+++ b/rag/prompts/generator.py
@@ -124,7 +124,7 @@ def kb_prompt(kbinfos, max_tokens, hash_id=False):
 
     knowledges = []
     for i, ck in enumerate(kbinfos["chunks"][:chunks_num]):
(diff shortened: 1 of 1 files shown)


evdmeer:
Is there any flags I need to set to get references to be filled?
I am on V0.22.0 and the references array of my agent responses is still empty.

I have cite enabled in the agent component.

yjfjebj789:
I still have the issue of not referencing in v0.22.1

buua436:
@evdmeer @yjfjebj789 

If possible, could you please open an issue and describe the “no citation” problem in more detail?
It would be very helpful if you could include:

Your configuration

The full RAGFlow response

Relevant agent logs

so that we can try to reproduce it.

This PR only fixes the case where the citation ID is incorrect, which might be a different root cause from the issue you’re seeing.

PR #10532, opened 2025-10-13
Refactor: How LiteLLMBase Calculate total count

Woody-Hu:

How LiteLLMBase Calculate total count

### Type of change

- [x] Refactoring


Woody-Hu:
> Hi, @Woody-Hu,
> 
> This is a good idea to uniformly use `total_token_count_from_response` and remove `total_token_count` from `LiteLLMBase`, which should have been deprecated in the last refactoring.
> 
> Also, could you help me replace other use of `total_token_count` with `total_token_count_from_response`? For example, in `Base`, `total_token_count` is simply an alias for `total_token_count_from_response`. By this way, we can completely clean up the `total_token_count`.
> 
> Cheers.

thanks already done

yongtenglei (approved):

PR #10643, opened 2025-10-17
Feat: add German translations for all agent templates and optimized line breaks for template titles

Jonah879:
### What does this PR solve?
German translation for all agent template and optimizing line breaks in the title for the new translation.

### Type of change


agent/templates/sql_assistant.json
```diff
@@ -114,15 +116,17 @@
                         "params": {
                             "cross_languages": [],
                             "empty_response": "",
-                            "kb_ids": [],
+                            "kb_ids": [
```
KevinHuSh: Appreciations! KB ID should not be here, should be?

Jonah879: Yes, they are not necessary. I think this is some artifact from an older version. They work as a placeholder. Should I delete the IDs?
Bildschirmfoto 2025-10-17 um 1 30 55 PM

KevinHuSh: Yes, the ID is a local one which can not be apply to other user.

agent/templates/sql_assistant.json
```diff
@@ -425,7 +427,7 @@
                                         "value": ""
                                     }
                                 },
-                                "query": "{sys.query}",
+                                "query": "sys.query",
```
KevinHuSh: Should be `{sys.query}`.

agent/templates/image_lingo.json
KevinHuSh, line 1: Please remove the unnecessary difference.

PR #10860, opened 2025-10-28
Feat: excel formulas support

OXI-717:
### Type of change


Magicbook1108:
Hi @OXI-717, really appreciate the time and thought you’ve put into these new parsing features. After reviewing and testing the changes, I have a few suggestions and clarifications below.

**1. Parse Excel as Table**

- The current implementation of "Parse Excel as Table" is not practical.
For example, if an Excel file contains 10,000 rows, it will result in 10,000 chunks, which is inefficient and meaningless in most scenarios.

- Additionally, the function does not handle duplicate attribute names properly.
As shown below:

> File "/home/bxy/github_reps/1111/rag/app/table.py", line 360, in chunk
>     raise ValueError(f"Duplicate column names detected: {duplicates}\nFrom: {clmns}")
> ValueError: Duplicate column names detected: ['项目1']
> From: ['项目1' '项目1' '全年' '1季度' '2季度' '3季度' '4季度' '1月' '2月' '3月' '4月' '5月' '6月'
>  '7月' '8月' '9月' '10月' '11月' '12月' 'Column_20']

_While duplicate column names should ideally be avoided, they are not uncommon in real-world Excel files_ 

**_Conclusion: We’ve decided not to accept the current implementation of "Parse Excel as Table" due to these limitations._**

**2. Include Formulas**
This is a valuable and meaningful feature, and we’re happy to keep it as part of our existing functionality.
However, there are still a few adjustments required:

- When the option is enabled, it currently outputs Formula + Formula, instead of the intended Formula + Content format.
图片

- Currently, we don’t plan to include frontend modifications for this feature. The reason is that formula display only applies to Excel files, while our knowledge base may contain various other file types. The UI layout and icon placement will be handled later by our UI designer. **_We want this feature is enabled by default._**

OXI-717:
Hi, @Magicbook1108.
1. Parse Excel as Table.
This is just additional property, which already added by default in "Table" mode, when you parse all table's as set of rows.
So, in General method, user can turn on this feature for small tables.

2. Include Formulas
_When the option is enabled, it currently outputs Formula + Formula, instead of the intended Formula + Content format._
This happens when no value in cell, if you has saved value in cell, then you should get Formula + Content.

Thanks for review.

Magicbook1108:
@OXI-717 
Hello, I have tested the Include Formulas feature with the attached XLSX file. In Column T, a formula is used to sum columns C and D. However, the parsing results returned are not as expected, as shown in the screen shot above.
图片

KevinHuSh:
Appreciations!

PR #10882, opened 2025-10-30
Refactor:improve the logic for rerank models to cal the total token count

Woody-Hu:

improve the logic for rerank models to cal the total token count

### Type of change

- [x] Refactoring

### DIFF
--- a/rag/utils/__init__.py
+++ b/rag/utils/__init__.py
@@ -69,6 +69,12 @@ def total_token_count_from_response(resp):
             return resp["usage"]["input_tokens"] + resp["usage"]["output_tokens"]
         except Exception:
             pass
+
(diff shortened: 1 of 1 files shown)


rag/llm/rerank_model.py
```diff
@@ -301,7 +298,7 @@ def similarity(self, query: str, texts: list):
             log_exception(_e, response)
         return (
             rank,
-            response["meta"]["tokens"]["input_tokens"] + response["meta"]["tokens"]["output_tokens"],
```
yongtenglei, line 304: Not a good replacement here, will result in 0, based on the [code](https://github.com/infiniflow/ragflow/blob/4be3754340e92bb5b6dfbebdc55c95a9720e0d1d/rag/utils/__init__.py#L48).

yongtenglei (changes requested):

rag/llm/rerank_model.py
Woody-Hu: thanks already add the new logics

PR #11042, opened 2025-11-05
Fix: add secure connection configuration for MinIO

tonymeysk:

Ragflow can't be used with a minio instance that uses a secure connection. Added an environment variable that activate and deactivate the secure connection in the minio instance in the code.

### Type of change

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -88,6 +88,8 @@ MINIO_USER=rag_flow
 # The password for MinIO.
 # When updated, you must revise the `minio.password` entry in service_conf.yaml accordingly.
 MINIO_PASSWORD=infini_rag_flow
+# Whether to use secure connection (HTTPS) for MinIO.
+MINIO_SECURE=false
 
 # The hostname where the Redis service is exposed
 REDIS_HOST=redis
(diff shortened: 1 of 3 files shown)


rag/utils/minio_conn.py
```diff
@@ -41,7 +41,7 @@ def __open__(self):
             self.conn = Minio(globals.MINIO["host"],
                               access_key=globals.MINIO["user"],
                               secret_key=globals.MINIO["password"],
-                              secure=False
+                              secure=globals.MINIO["secure"]
```
KevinHuSh: Appreciations!
I do not think this can get the envirement variable, isn't it?

tonymeysk: You are right. I changes the line and also added a default value to not break anyone else configuration.

```python
secure=str(settings.MINIO.get("secure", False)).lower() == "true"
```

docker/.env
```diff
@@ -88,6 +88,8 @@ MINIO_USER=rag_flow
 # The password for MinIO.
 # When updated, you must revise the `minio.password` entry in service_conf.yaml accordingly.
 MINIO_PASSWORD=infini_rag_flow
+# Whether to use secure connection (HTTPS) for MinIO.
+MINIO_SECURE=false
```
KevinHuSh, line 92: How could it be read from environment variables?

tonymeysk: I thought the part
MINIO = decrypt_database_config(name="minio")
in common/settings.py would read also that variable like it reads the host, user and password. Isn't that the case?

KevinHuSh: MINIO configuration is from `conf/servvice_conf.yaml`.

PR #11116, opened 2025-11-07
Feat/monitor task

Lynn-Inf:

Show task executor.

### Type of change


admin/client/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "ragflow-cli"
-version = "0.21.1"
+version = "0.22.0"
```
JinHai-CN, line 3: This version should be updated before release

Issue #11125, opened 2025-11-10
[Question]: [ERROR][Exception]: 'text'


panxiaohuan9:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

I encountered a 'text' error while using the pipeline to parse documents.
dataset——> files——>parse(ingestion pipeline)

18:04:33 Task has been received.
-------------------------------------
[File]:
18:04:34: File fetched.
18:04:34: Done
Start the pipeline...
-------------------------------------
[Parser_0]:
18:04:34: Start to work on a PDF.
18:04:44:
18:04:46: OCR finished (12.11s)
18:04:52: Layout analysis (6.75s)
18:04:53: Table analysis (0.62s)
18:04:53: Text merged (0.00s)
18:04:53: Structured (0.25s)
18:04:54: Done
-------------------------------------
[按标题分割_0]:
18:04:54: Start to merge hierarchically.
18:04:54: Done.
18:04:54: Done
-------------------------------------
[提取器_摘要]:
18:04:54: Start to generate.
18:04:55: Done
-------------------------------------
[提取器_关键词]:
18:04:55: Start to generate.
18:04:56: Done
-------------------------------------
[提取器_问题]:
18:04:57: Start to generate.
18:04:59: Done
-------------------------------------
[提取器_元数据]:
18:04:59: Start to generate.
18:05:01: Done
-------------------------------------
[分词器_0]:
18:05:01: Start to tokenize.
18:05:01: Finish tokenizing.
18:05:01: Start embedding inference.
18:05:01: Finish embedding.
18:05:01: Done
18:05:01 [ERROR][Exception]: 'text'

panxiaohuan9:
<img width="1920" height="930" alt="Image" src="https://github.com/user-attachments/assets/3fb7eca1-dd75-4e54-830e-65651a1b46a5" />

Magicbook1108:
@panxiaohuan9 
Can you share your document and the ingestion pipeline with me via xiaoyang.bao@infiniflow.org

[...]

PR #11218, opened 2025-11-12
Refactor: flow tokenizer check chunk fields

Woody-Hu:
may relate https://github.com/infiniflow/ragflow/issues/11125

### Type of change

- [x] Refactoring

### DIFF
--- a/rag/flow/tokenizer/tokenizer.py
+++ b/rag/flow/tokenizer/tokenizer.py
@@ -65,6 +65,7 @@ class Tokenizer(ProcessBase):
             txt = ""
             if isinstance(self._param.fields, str):
                 self._param.fields=[self._param.fields]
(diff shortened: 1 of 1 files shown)


rag/flow/tokenizer/tokenizer.py
```diff
@@ -65,7 +65,10 @@ async def _embedding(self, name, chunks):
             txt = ""
             if isinstance(self._param.fields, str):
                 self._param.fields=[self._param.fields]
+                
             for f in self._param.fields:
+                if f not in c:
+                    continue
                 f = c.get(f)
```
KevinHuSh, line 72: Appreciations ^^
`f` is None if `f` is not in `c`.
So, it will continue anyway, won't it?

PR #11167, opened 2025-11-11
Fix: Update env to support PPTX and update README for version changes

Magicbook1108:
### What problem does this PR solve?

Fix: Update env to support PPTX
Fix: update README for version changes #11138

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [x] Documentation Update

PR #11272, opened 2025-11-14
Fix: create dataset return type inconsistent

Magicbook1108:

Fix: create dataset return type inconsistent #11167 
 
### Type of change

### DIFF
--- a/sdk/python/test/test_frontend_api/test_dataset.py
+++ b/sdk/python/test/test_frontend_api/test_dataset.py
@@ -101,7 +101,7 @@ def test_invalid_name_dataset(get_auth):
     # create dataset
     # with pytest.raises(Exception) as e:
     res = create_dataset(get_auth, 0)
-    assert res['code'] == 102
+    assert res['code'] == 100
 
     res = create_dataset(get_auth, "")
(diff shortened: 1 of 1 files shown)


api/db/services/knowledgebase_service.py
```diff
@@ -409,21 +409,21 @@ def create_with_name(
...
             "created_by": tenant_id,
             "parser_id": (parser_id or "naive"),
+            "parser_config": parser_config,
             **kwargs
         }
-
-        # Default parser_config (align with kb_app.create) — do not accept external overrides
-        payload["parser_config"] = get_parser_config(parser_id, kwargs.get("parser_config"))
+        
         return payload
```
JinHai-CN, line 428: dict

api/db/services/knowledgebase_service.py
```diff
@@ -409,21 +409,21 @@ def create_with_name(
         # Verify tenant exists
         ok, _t = TenantService.get_by_id(tenant_id)
         if not ok:
-            return False, "Tenant not found."
+            return get_data_error_result(message="Tenant does not exist.")
```
JinHai-CN: json

Issue #11390, opened 2025-11-19
[Bug]: docker-compose redis config file broken


panzi:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

v0.22.1

### RAGFlow image version

infiniflow/ragflow:v0.22.1

### Other environment information

```Markdown

```

### Actual behavior

Starting the docker-compose setup fails with this error:
```
*** FATAL CONFIG FILE ERROR (Version 8.1.4) ***
Reading the configuration file, at line 2
>>> 'requirepass "--maxmemory" "128mb"'
wrong number of arguments
```

### Expected behavior

Redis configuration in the docker-compose setup should work.

### Steps to reproduce

```Markdown
git checkout v0.22.1
cd docker
RAGFLOW_IMAGE=infiniflow/ragflow:v0.22.1 docker-compose up
```

panzi:
Oh I think this is because of a blank `REDIS_PASSWORD` and missing quotes in `docker-compose-base.yml`. An array like this would be better anyway:

```YAML
...
services:
  ...
  redis:
    ...
    command: ["redis-server", "--requirepass", "${REDIS_PASSWORD}", "--maxmemory", "128mb", "--maxmemory-policy", "allkeys-lru"]
...
```

In general the array syntax is preferable in docker-compose, because it prevents this exact problem. Also if I'm not mistaken using the array syntax means that the command is run directly, i.e. not through bash, which means signals like `SIGTERM` will be actually delivered to the `redis-server` process and graceful shutdown is possible.

PR #11391, opened 2025-11-19
Use array syntax for commands in docker-compose-base.yml

panzi:
Use array syntax in order to prevent parameter quoting issues. This also runs the command directly without a bash process, which means signals (like SIGTERM) will be delivered directly to the server process.

Fixes issue #11390

`${REDIS_PASSWORD}` was not passed correctly, meaning if it was unset or contains spaces (or shell code!) it was interpreted wrongly.

### Type of change

### DIFF
--- a/docker/docker-compose-base.yml
+++ b/docker/docker-compose-base.yml
@@ -154,7 +154,7 @@ services:
 
   minio:
     image: quay.io/minio/minio:RELEASE.2025-06-13T11-33-47Z
-    command: server --console-address ":9001" /data
+    command: ["server", "--console-address", ":9001", "/data"]
     ports:
       - ${MINIO_PORT}:9000
       - ${MINIO_CONSOLE_PORT}:9001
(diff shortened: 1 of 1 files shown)


docker/docker-compose-base.yml
```diff
@@ -154,7 +154,7 @@ services:
 
   minio:
     image: quay.io/minio/minio:RELEASE.2025-06-13T11-33-47Z
-    command: server --console-address ":9001" /data
+    command: ["server", "--console-address", ":9001", "/data"]
```
panzi, line 157: Here was no parameter quoting issue, I just saw another string based command. Replacing it with an array has the advantage for it to be directly executed with no bash intermediary and thus correctly getting signals delivered, so I also changed that here.

docker/docker-compose-base.yml
```diff
@@ -176,7 +176,7 @@ services:
   redis:
     # swr.cn-north-4.myhuaweicloud.com/ddn-k8s/docker.io/valkey/valkey:8
     image: valkey/valkey:8
-    command: redis-server --requirepass ${REDIS_PASSWORD} --maxmemory 128mb --maxmemory-policy allkeys-lru
+    command: ["redis-server", "--requirepass", "${REDIS_PASSWORD}", "--maxmemory", "128mb", "--maxmemory-policy", "allkeys-lru"]
```
panzi, line 179: `${REDIS_PASSWORD}` wasn't quoted, so an empty password gave a confusing error.

AFAIK `${REDIS_PASSWORD}` is replaced by docker-compose, not by bash, thus this would even allow embedding shell code (`; rm -rf . ; #`).

PR #11408, opened 2025-11-20
Fix: missing parameters in by_plaintext method for PDF naive mode

it-coder:

FIx: missing parameters in by_plaintext method for PDF naive mode

### Type of change

### DIFF
--- a/rag/app/naive.py
+++ b/rag/app/naive.py
@@ -59,6 +59,7 @@ def by_mineru(filename, binary=None, from_page=0, to_page=100000, lang="Chinese"
     mineru_executable = os.environ.get("MINERU_EXECUTABLE", "mineru")
     mineru_api = os.environ.get("MINERU_APISERVER", "http://host.docker.internal:9987")
     pdf_parser = MinerUParser(mineru_path=mineru_executable, mineru_api=mineru_api)
+    parse_method = kwargs.get("parse_method", "raw")
 
     if not pdf_parser.check_installation():
         callback(-1, "MinerU not found.")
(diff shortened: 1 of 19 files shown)


rag/app/naive.py
```diff
@@ -111,7 +111,13 @@ def by_tcadp(filename, binary=None, from_page=0, to_page=100000, lang="Chinese",
 
 
 def by_plaintext(filename, binary=None, from_page=0, to_page=100000, callback=None, **kwargs):
-    if kwargs.get("layout_recognizer", "") == "Plain Text":
+    parser_config = kwargs.get("parser_config", {})
+    layout_recognizer = parser_config.get("layout_recognize", "DeepDOC")
```
Magicbook1108: `parser_config` is declared on line 560.

rag/app/naive.py
```diff
@@ -111,7 +111,13 @@ def by_tcadp(filename, binary=None, from_page=0, to_page=100000, lang="Chinese",
...
+
+    if isinstance(layout_recognizer, bool):
+        layout_recognizer = "DeepDOC" if layout_recognizer else "Plain Text"
```
Magicbook1108: line 639 - 640 has already checked `layout reconginzer` type

it-coder: Version 0.22: When chunk_method = "one", please check lines 94-103 in one.py

Magicbook1108: Just add `layout_recognizer = layout_recongizer` in one.py line 102

it-coder: In the new version, PDF parsing does not directly use the chunk method from naive.py, but rather the by_plaintext method called by chunk methods like one.py.  Please check again.

it-coder: Added `layout_recognizer = layout_recognizer` parameter to all relevant methods. Please review again.

PR #11472, opened 2025-11-24
Fix: coroutine object has no attribute get

Magicbook1108:

Fix: coroutine object has no attribute get

### Type of change


yongtenglei:
#11537

PR #11596, opened 2025-11-28
fix(agent): prevent event-loop blocking in Canvas.run by async-ifying batch, file & stream operations

zhuzhaoyun:
While executing long-running Agent tasks (e.g. `Canvas.run`), the Quart async event-loop is blocked by several synchronous calls:

1. `_run_batch` waits on `ThreadPoolExecutor.submit(...).result()` in the main thread.  
2. `get_files` uses the same pattern, stalling concurrent uploads.  
3. The Message component iterates a **synchronous** generator inside `async for`, freezing the loop until the whole LLM stream ends.  

As a result, simple HTTP requests such as Edge-browser login often time out under load.

### How does this PR fix it?  
- Make `_run_batch` async → await `asyncio.wrap_future(executor.submit(...))`.  
- Make `get_files` async → gather all parse jobs via `asyncio.gather(*[loop.run_in_executor(...)])`.  
- Off-load sync generator consumption to a thread and feed tokens back through `asyncio.Queue`, so the main loop stays responsive.

### Type of change  

### Backward compatibility  
No breaking changes; public signatures remain unchanged.

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -18,6 +18,8 @@ import json
 import logging
 import re
 import time
+import asyncio
+import threading
 from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from functools import partial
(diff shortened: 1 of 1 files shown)


yongtenglei:
Hi @zhuzhaoyun,

I'm working on it now, but I haven't implemented the get_files part yet. If you would still like to work on it, please feel free to go ahead. Alternatively, could you please wait for my changes and then see if there is anything you would like to add.

Thank you very much for your enthusiasm!

PR #11599, opened 2025-11-28
`fix(agent): prevent event-loop blocking in Canvas.run by async-ifying batch, file & stream operations`

zhuzhaoyun:
While executing long-running Agent tasks (e.g. `Canvas.run`), the Quart async event-loop is blocked by several synchronous calls:

1. `_run_batch` waits on `ThreadPoolExecutor.submit(...).result()` in the main thread.  
2. `get_files` uses the same pattern, stalling concurrent uploads.  
3. The Message component iterates a **synchronous** generator inside `async for`, freezing the loop until the whole LLM stream ends.  

As a result, simple HTTP requests such as Edge-browser login often time out under load.

### How does this PR fix it?  
- Make `_run_batch` async → await `asyncio.wrap_future(executor.submit(...))`.  
- Make `get_files` async → gather all parse jobs via `asyncio.gather(*[loop.run_in_executor(...)])`.  
- Off-load sync generator consumption to a thread and feed tokens back through `asyncio.Queue`, so the main loop stays responsive.

### Type of change  

### Backward compatibility  
No breaking changes; public signatures remain unchanged.

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -18,6 +18,8 @@ import json
 import logging
 import re
 import time
+import asyncio
+import threading
 from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from functools import partial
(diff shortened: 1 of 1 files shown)


agent/canvas.py
```diff
@@ -462,7 +467,24 @@ def _node_finished(cpn_obj):
                 if cpn_obj.component_name.lower() == "message":
                     if isinstance(cpn_obj.output("content"), partial):
                         _m = ""
-                        for m in cpn_obj.output("content")():
+                        def generator_producer(gen, q, loop):
+                            try:
+                                for item in gen:
+                                    loop.call_soon_threadsafe(q.put_nowait, item)
+                            except Exception as e:
```
KevinHuSh:
```suggestion
                            except Exception:
```

KevinHuSh:
Appreciations!
Outdated.

Issue #11620, opened 2025-12-01
[Bug]: When using mineru for parsing, I found a parsing error with the error message [MinerU] Missing output file


gegiaogiao:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

v0.22.1

### RAGFlow image version

v0.22.1

### Other environment information

```Markdown

```

### Actual behavior

When using mineru for parsing, I found a parsing error with the error message [MinerU] Missing output file:

ragflow logs about parser file:
22:39:00 Task has been received.
22:39:00 Page(1~10): Start to parse.
22:39:04 Page(1~10): [MinerU] Received binary PDF -> /tmp/mineru_bin_pdf_8kz9s_ym/page_1_bug_9.pdf
22:39:04 Page(1~10): [MinerU] Output directory: /tmp/mineru_pdf_ber10isa
22:39:18 Page(1~10): [ERROR]Internal server error while chunking: [MinerU] Missing output file: /tmp/mineru_pdf_ber10isa/page_1_bug_9/auto/page_1_bug_9_content_list.json
22:39:18 [ERROR][Exception]: [MinerU] Missing output file: /tmp/mineru_pdf_ber10isa/page_1_bug_9/auto/page_1_bug_9_content_list.json

### Expected behavior

Can parse the path normally

### Steps to reproduce

```Markdown
After deploying the embedding model locally, when using the front-end PDF parsing method to cut files, Deepdoc can work normally, but MinerU and Docling report errors and cannot find the output file
```

yongtenglei:
Hi @gegiaogiao 

I guess you are setting `MINERU_APISERVER`?

shaoqing404:
Here is a concise and technical English comment you can directly post under the issue.


Proposed English Comment

We ran into a similar symptom (“[MinerU] Missing output file”) but with a different root cause, so I’m adding our findings here in case it helps clarify the scope of the problem.

Our pipeline:
`ragflow → mineru-api → mineru-vllm`
[...]

Issue #11613, opened 2025-11-30
[Bug]: [MinerU] Missing output file


yanghan-cyber:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

v0.22.1

### RAGFlow image version

v0.22.1

### Other environment information

```Markdown
docker compose
```

### Actual behavior

I use pip install mineru[core,vllm]

mineru version: 
```
mineru                            2.6.5
mineru_vl_utils                   0.1.17
```
mineru .env:
```
# Enable Mineru
USE_MINERU=false
MINERU_EXECUTABLE="$HOME/uv_tools/.venv/bin/mineru"
MINERU_DELETE_OUTPUT=0   # keep output directory
MINERU_BACKEND="vlm-http-client" # or another backend you prefer
# MINERU_SERVER_URL="http://192.168.31.3:8000"
MINERU_APISERVER="http://192.168.31.3:8000"
MINERU_SERVER_URL="http://localhost:30000"
```

When using mineru for parsing, I found a parsing error with the error message `[MinerU] Missing output file:`

ragflow logs about parser file:
```
20:36:57 Task has been received.
20:36:58 Page(1~18): Start to parse.
20:36:58 Page(1~18): [MinerU] Received binary PDF -> /tmp/mineru_bin_pdf_47fwygm_/test.pdf
20:36:59 Page(1~18): [MinerU] Output directory: /tmp/mineru_pdf_h1ujvymd
20:36:59 Page(1~18): [MinerU] invoke api: http://192.168.31.3:8000/file_parse
20:37:12 Page(1~18): [MinerU] zip file returned, saving to /tmp/mineru_pdf_h1ujvymd/output.zip...
20:37:12 Page(1~18): [MinerU] Unzip to /tmp/mineru_pdf_h1ujvymd/test/auto...
[...]

PR #11623, opened 2025-12-01
Fix: [MinerU] Missing output file

yongtenglei:

Add fallbacks for MinerU output path. #11613, #11620.

### Type of change

### DIFF
--- a/rag/llm/chat_model.py
+++ b/rag/llm/chat_model.py
@@ -33,9 +33,9 @@ from openai.lib.azure import AzureOpenAI
 from strenum import StrEnum
 from zhipuai import ZhipuAI
 
+from common.token_utils import num_tokens_from_string, total_token_count_from_response
 from rag.llm import FACTORY_DEFAULT_BASE_URL, LITELLM_PROVIDER_PREFIX, SupportedLiteLLMProvider
 from rag.nlp import is_chinese, is_english
-from common.token_utils import num_tokens_from_string, total_token_count_from_response
 
 
 # Error message constants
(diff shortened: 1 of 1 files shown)


shaoqing404:
**Proposed English Comment**

We ran into a similar symptom (“[MinerU] Missing output file”) but with a different root cause, so I’m adding our findings here in case it helps clarify the scope of the problem.

**Our pipeline:**
`ragflow → mineru-api → mineru-vllm`
The downstream MinerU / DeepDoc / VLLM steps work correctly. The only failure occurs at the very beginning when Ragflow calls MinerU via the API.

**What we observed:**
When the PDF filename contains Chinese characters or punctuation, the MinerU API sanitizes the name when creating the ZIP. For example:

```
《飞行手册》08-10版-2025.01.20
→  _飞行手册_08-10版-2025.01.20
```

However, the client-side extractor (inside DeepDoc/MinerU parser) still assumes the **original** file stem when creating the output root directory. After unzipping, the actual paths look like:

```
.../<raw_stem>/auto/<sanitized_stem>/auto/<sanitized_stem>_content_list.json
```

But `_read_output` only searches:

```
<raw_stem>/<method>/<raw_stem>_content_list.json
*/<method>/*_content_list.json
```

Because of the extra sanitized directory layer, no match is found, and it raises the same error message as #11620:

```
[MinerU] Missing output file
```

**Comparison with #11620:**

* Issue #11620 discusses VLM output directory selection or MINERU_APISERVER usage.
* In our case, VLLM and all downstream components behave normally.
* The failure is specifically triggered when Ragflow → MinerU-API is used with non-ASCII filenames and the API rewrites the ZIP root directory.

**Conclusion:**
The symptom matches #11620, but the root cause is different.
This one needs a fix or compatibility change in how the MinerU API handles non-ASCII filenames (ZIP root) versus how the parser resolves the expected output directory.

We can provide a minimal reproduction case (with a Chinese-named PDF) if needed.

yongtenglei:
Also see #11701.

PR #11639, opened 2025-12-01
feat: improve presentation PdfParser

rommy2017:
The old presentation PdfParser lost table format after parse

### Type of change

### DIFF
--- a/rag/app/presentation.py
+++ b/rag/app/presentation.py
@@ -23,6 +23,7 @@ from PIL import Image
 from PyPDF2 import PdfReader as pdf2_read
 
 from deepdoc.parser import PdfParser, PptParser, PlainParser
+from rag.app.naive import by_plaintext, PARSERS
 from rag.nlp import rag_tokenizer
 from rag.nlp import tokenize, is_english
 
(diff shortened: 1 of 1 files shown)


rag/app/presentation.py
```diff
@@ -125,33 +181,19 @@ def chunk(filename, binary=None, from_page=0, to_page=100000,
...
-            filename = filename,
-            binary = binary,
-            from_page = from_page,
-            to_page = to_page,
-            lang = lang,
-            callback = callback,
-            pdf_cls = Pdf,
-            layout_recognizer = layout_recognizer,
-            **kwargs
+        sections, _ = parser(
```
Magicbook1108: The parser must return three parameters to comply with `by_deepdoc() `in `rag/app/naive`.
At the same time, we want our PDF parsers to remain broad and flexible, so we will continue supporting MinerU, TCADP, and other extra parsers.

rag/app/presentation.py
```diff
@@ -51,35 +52,90 @@ def __call__(self, fnm, from_page, to_page, callback=None):
...
+            if not (from_page < current_page_num <= to_page + from_page):
+                continue
+
+            top = positions[0][3]
+            left = positions[0][1]
+
+            page_items[current_page_num].append({
+                "top": top,
+                "x0": left,
+                "text": final_text,  # 确保这里一定是字符串
```
Magicbook1108: We don't accept chinese comment.

Magicbook1108 (approved):
ok to merge

Issue #11700, opened 2025-12-03
[Bug]: Chunking failure “not enough values to unpack (expected 3, got 2)” caused by mismatch in section tuple structure


shaoqing404:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

unk

### RAGFlow image version

0.22.1

### Other environment information

```Markdown
**Ragflow version:** 0.22.1
**MinerU version:** 2.6.4

ubuntu 22.04
```

### Actual behavior

When processing a PDF, MinerU successfully parsed the document, but Ragflow threw an exception during the chunking stage in `manual.py`, causing the task to fail.

### 1. Log excerpt showing the issue

```
2025-12-03 16:17:07 Parsed 5 blocks from PDF.
2025-12-03 16:17:08 [ERROR] Internal server error while chunking: not enough values to unpack (expected 3, got 2)

Traceback (most recent call last):
  ...
  File ".../rag/app/manual.py", line 243, in chunk
    bull = bullets_category([txt for txt, _, _ in sections])
ValueError: not enough values to unpack (expected 3, got 2)
```

### 2. Problem Description

In `manual.py`, the chunking logic assumes that each element in `sections` is a **3-tuple**, as shown here:

```python
[txt for txt, _, _ in sections]
```

However, MinerU (or Docling) in the current version is returning **2-tuples** instead, typically of the form:

```
(text, tag)
```

This mismatch causes Python to raise:

```
ValueError: not enough values to unpack (expected 3, got 2)
```

and the chunking process terminates with an error.

See  3 and 4

### 5. Suggested Fix or Investigation Path

1. Confirm whether MinerU’s output structure has changed from 3-tuple to 2-tuple in recent updates.
2. Add validation or normalization before iterating over `sections`.
3. Introduce backward compatibility, such as:

```python
txt = section[0]
```

or mapping 2-tuples to a 3-tuple format if needed by downstream logic.

### 6. Additional Information

I can provide the PDF file that triggers this issue if needed.

[...]

Issue #11136, opened 2025-11-10
[Question]:


apps-lycusinc:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

Is some settings causing below error at parsing ? I am using General pipeline and docling parser

[Case study.pdf](https://github.com/user-attachments/files/23446659/Case.study.pdf)

13:54:53 Page(1~9): Start to parse.
13:54:53 Page(1~9): [ERROR]Docling not found.
13:54:53 Page(1~9): [ERROR]Internal server error while chunking: not enough values to unpack (expected 3, got 2)
13:54:53 [ERROR][Exception]: not enough values to unpack (expected 3, got 2)

Attached file is not parsed

buua436:
Thanks for the report — we’ve reproduced the issue and shipped a fix. The error occurs because Docling isn’t installed/enabled in your environment.

To resolve:
1. Set USE_DOCLING=true in your environment or Docker .env file.
2. Restart your service or container after making these changes.

PR #11702, opened 2025-12-03
Fix MinerU API sanitized-output lookup and manual chunk tuple handling

shaoqing404:

This PR addresses **two independent issues** encountered when using the MinerU engine in Ragflow:

1. **MinerU API output path mismatch for non-ASCII filenames**
   MinerU sanitizes the root directory name inside the returned ZIP when the original filename contains non-ASCII characters (e.g., Chinese).
   Ragflow's client-side unzip logic assumed the original filename stem and therefore failed to locate `_content_list.json`.
   This PR adds:

   * root-directory detection
   * fallback lookup using sanitized names
   * a broadened `_read_output` search with a glob fallback
     ensuring output files are consistently located regardless of filename encoding.

2. **Chunker crash due to tuple-structure mismatch in manual mode**
   Some parsers (e.g., MinerU / Docling) return **2-tuple sections**, but Ragflow’s chunker expects **3-tuple sections**, leading to:
   `ValueError: not enough values to unpack (expected 3, got 2)`
   This PR normalizes all sections to a uniform structure `(text, layout, positions)`:

   * parse position tags when present
   * default to empty positions when missing
     preserving backward compatibility and preventing crashes.

### Type of change

* [x] Bug Fix (non-breaking change which fixes an issue)

[#11136](https://github.com/infiniflow/ragflow/issues/11136)
[#11700](https://github.com/infiniflow/ragflow/issues/11700)
[#11620](https://github.com/infiniflow/ragflow/issues/11620)
[#11701](https://github.com/infiniflow/ragflow/pull/11701)

we need your help [yongtenglei](https://github.com/yongtenglei)

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -62,29 +62,8 @@ class MinerUParser(RAGFlowPdfParser):
         self.outlines = []
         self.logger = logging.getLogger(self.__class__.__name__)
 
-    @staticmethod
-    def _sanitize_output_name(name: str) -> str:
-        """Approximate MinerU API sanitize logic while keeping CJK characters."""
-        cleaned = re.sub(r"[^\w.\-\u4e00-\u9fff]", "_", name.strip())
-        cleaned = re.sub(r"_+", "_", cleaned).strip("_")
-        return cleaned or name
-
-    @staticmethod
-    def _detect_zip_root_dir(zip_path: Path) -> Optional[str]:
-        try:
-            with zipfile.ZipFile(zip_path, "r") as zip_ref:
-                for entry in zip_ref.namelist():
-                    if not entry:
-                        continue
-                    entry = entry.rstrip("/")
-                    if not entry:
-                        continue
-                    return entry.split("/", 1)[0]
-        except Exception:
-            return None
-        return None
-
     def _extract_zip_no_root(self, zip_path, extract_to, root_dir):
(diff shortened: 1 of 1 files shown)


yongtenglei:
Hi, @shaoqing404 

Thanks for your quick responding. It seems that we are doing the same thing. From my personal perspective, it is kind of overkill to do the root-directory detection and search with a glob fallback. I have an implementation in another PR, it is shorter and cleaner. 

For the other issue you are addressing, I would like to bring in @Magicbook1108 to help with that.

Cheers.

shaoqing404:
> Hi, @shaoqing404
> 
> Thanks for your quick responding. It seems that we are doing the same thing. From my personal perspective, it is kind of overkill to do the root-directory detection and search with a glob fallback. I have an implementation in another PR, it is shorter and cleaner.
> 
> For the other issue you are addressing, I would like to bring in @Magicbook1108 to help with that.
> 
> Cheers.

Hi, thanks for taking the time to look into this and for sharing your thoughts.

If your alternative implementation resolves the issue in a clearer and more elegant way, I’m very glad to see it move forward. My patch was meant to ensure stability across the cases we’ve encountered, but there is always value in a simpler path when it can hold the same ground. I appreciate you pointing that out.

And thank you for bringing in @Magicbook1108 regarding the tuple normalization.
If any additional examples or context from my side would help, please feel free to let me know. I’m happy to assist wherever it’s useful.

Good to work with you on this.

PR #11752, opened 2025-12-04
Feat: use filepath for files with the same name

Jonah879:

When there are multiple files with the same name the file would just duplicate, making it hard to distinguish between the different files. Now if there are multiple files with the same name, they will be named after their folder path in the webdav storage unit. 

The same could be done for the other connectors, too, since most of them will have similars issues, when iterating through the folder paths.

### Type of change

Contribution by RAGcon GmbH, visit us [here](https://www.ragcon.ai/)


JinHai-CN:
@Jonah879 

We're reaching out because we're interested in inviting RAGcon to join the RAGFlow Ecosystem Partner Program.

If you're open to learning more, could you please drop us a line at [haijin.chn@gmail.com](mailto:haijin.chn@gmail.com)? We'll reply with the formal invitation and all the details.

Looking forward to hearing from you!

Issue #11653, opened 2025-12-02
[Feature Request]: Automatically Disable Raptor for Excel & Tabular PDF Data


ahmadshakil:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown
Yes. Currently, Raptor is applied to all file types, including highly structured data like Excel files and tabular PDF content. This causes several issues:

1. Excel files already have organized semantic units (e.g., fixed 13 lines per chunk).

2. Tabular data in PDFs is naturally structured with rows and columns.

3. When Raptor converts Excel to HTML and clusters it, the number of tokens increases unnecessarily, leading to higher computational cost and larger memory usage.
```

### Describe the feature you'd like

Introduce logic to automatically disable Raptor for:

1. Excel files (.xls, .xlsx, .csv, etc.)

2. PDF documents containing tabular data (tables with consistent rows/columns)

3. Optional: Provide a configuration toggle (e.g., raptor.auto_disable_for_structured_data = true) to allow users to override this behavior if needed.

### Describe implementation you've considered

The implementation focuses on conditional disabling of Raptor based on file type and content structure. Key points:

File Type Detection

Identify Excel files via extensions (.xls, .xlsx, .csv).

Identify PDFs via .pdf extension.

Tabular Data Detection in PDFs

Use the existing PDF extraction module to detect tables (e.g., based on consistent row/column patterns).

If tabular content is detected, flag the PDF as “structured.”

Raptor Conditional Execution

Introduce a check before Raptor is invoked:

def should_run_raptor(file):
    if file.type in ["xls", "xlsx", "csv"]:
        return False
    if file.type == "pdf" and file.contains_tabular_data():
        return False
    return True

Skip Raptor processing if the function returns False.

Optional Configuration Toggle

Allow administrators to override the behavior using a config flag:

raptor:
  auto_disable_for_structured_data: true

### Documentation, adoption, use case

```Markdown
Use Cases / Scenarios:

[...]

PR #11771, opened 2025-12-05
Feature/raptor auto disable structured data

NikolaPantel:

**Feature:** This PR implements automatic Raptor disabling for structured data files to address issue #11653.

**Issue:** PDFs with HTML tables were still being processed by Raptor even after the auto-disable feature was merged.

**Background:** @ahmadshakil reported that when uploading `Fbr_IncomeTaxOrdinance_2001-amended-upto30.06.2024.pdf`, the HTML-based tables were still being sent to Raptor. Running `docker logs ... | grep -c 'Skipping Raptor for document'` returned 0 results.

**Root Cause:** The original implementation only checked:
- File extension (`.xlsx`, `.csv`, etc.)
- Parser ID (`table`)
- Parser config (`html4excel: true`)

However, when PDFs are parsed with the `naive` parser, tables are extracted and converted to `<table>` HTML tags in the chunk content. This content-level table extraction was not being detected.

**Solution:** Added content-based detection that analyzes actual chunk content for HTML table tags:
- New function `contains_html_table()` detects `<table>` tags in content
- New function `analyze_chunks_for_tables()` calculates table percentage across chunks
- New function `should_skip_raptor_for_chunks()` skips Raptor if ≥30% of chunks contain HTML tables
- Detection happens after chunks are loaded, before Raptor processing begins
- Threshold is configurable via `TABLE_CONTENT_THRESHOLD`

**New log output:**
```
Skipping Raptor for document {doc_id}: Content contains X% HTML tables (threshold: 30%) - Raptor auto-disabled
```

**Tests:** Added 21 new tests for content-based detection (65 total, all passing), including a test case simulating the reported PDF scenario.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


NikolaPantel:
@ahmadshakil Can you check the PR?

ahmadshakil:
@Magicbook1108  @KevinHuSh can you please?

NikolaPantel:
@KevinHuSh @TeslaZY @cike8899 Would you please check the PR and give me your feedbacks?

ahmadshakil:
[clustor_content.txt](https://github.com/user-attachments/files/24184090/clustor_content.txt)

@hsparks-codes  is it ok to have attached clustor passed to raptor? it still have table but may be ignored because its comparatively smaller?

NikolaPantel:
Yes, that's the intended behavior! The current implementation uses a 30% threshold - if less than 30% of chunks contain HTML tables, Raptor will still process the content.

The rationale is that documents with occasional tables (like reports with a few data tables mixed with narrative text) can still benefit from Raptor's hierarchical clustering. We only skip Raptor when tables dominate the content (30%+), where summarization would produce poor results.

If you'd like to adjust this threshold for your use case, it's configurable via `TABLE_CONTENT_THRESHOLD` in  `rag/utils/raptor_utils.py`. Would you prefer a different threshold value?

PR #11772, opened 2025-12-05
Feature (canvas): Add Api for download "message" component output's file

huang-aoqin:

-Add Api for download "message" component output's file 
-Change the attachment output type check from tuple to dictionary,because 'attachement' is not instance of tuple 
-Update the message type to message_end to avoid the problem that content does not send an error message when the message type is ans ["data"] ["content"]

### Type of change

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -534,10 +534,12 @@ class Canvas(Graph):
                         yield decorate("message", {"content": cpn_obj.output("content")})
                         cite = re.search(r"\[ID:[ 0-9]+\]",  cpn_obj.output("content"))
 
+                    message_end = {}
                     if isinstance(cpn_obj.output("attachment"), dict):
-                        yield decorate("message_end", {"attachment": cpn_obj.output("attachment")})
-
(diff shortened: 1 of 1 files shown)


agent/canvas.py
```diff
@@ -534,8 +534,8 @@ async def _process_stream(m):
                         yield decorate("message", {"content": cpn_obj.output("content")})
                         cite = re.search(r"\[ID:[ 0-9]+\]",  cpn_obj.output("content"))
 
-                    if isinstance(cpn_obj.output("attachment"), tuple):
-                        yield decorate("message", {"attachment": cpn_obj.output("attachment")})
+                    if isinstance(cpn_obj.output("attachment"), dict):
+                        yield decorate("message_end", {"attachment": cpn_obj.output("attachment")})
```
KevinHuSh: Appreciations!
But, there're two `message_end`, aren't there?

huang-aoqin: > Appreciations! But, there're two `message_end`, aren't there?

You are right,I have merged two message_end

Issue #4213, opened 2024-12-24
[Feature Request]: Support memory for Agent


yingfeng:
### Is there an existing issue for the same feature request?

- [X] I have checked the existing issues.

### Describe the feature you'd like

Memory is a fundamental building block for agent. Mem0 is a reference for memory API encapsulation.

PR #11812, opened 2025-12-08
Feat/memory

Lynn-Inf:

Manage and display memory datasets.

Issue #4213 

### Type of change


api/apps/memories_app.py
```diff
@@ -0,0 +1,178 @@
+#
+#  Copyright 2024 The InfiniFlow Authors. All Rights Reserved.
```
JinHai-CN: Copyright 2025

PR #11818, opened 2025-12-08
fix: align http client proxy kwarg

levischd:

Our HTTP wrapper still passed proxies to httpx.Client/AsyncClient, which expect proxy. As a result, configured proxies were ignored and calls could fail with ValueError("Failed to fetch OIDC metadata: Client.__init__() got an unexpected keyword argument 'proxies'"). This PR switches to the correct proxy kwarg so proxies are honored and the runtime error is resolved.

### Type of change

---


levischd:
Bildschirmfoto 2025-12-08 um 11 23 55

PR #11819, opened 2025-12-08
Feat: use filepath for files with the same name for all data source types

Jonah879:

When there are multiple files with the same name the file would just duplicate, making it hard to distinguish between the different files. Now if there are multiple files with the same name, they will be named after their folder path in the storage unit.

This was done for the webdav connector and with this PR also for Notion, Confluence and S3 Storage. 

### Type of change

Contribution by RAGcon GmbH, visit us [here](https://www.ragcon.ai/)


Magicbook1108:
@Jonah879 Hello, I’ve tested the Confluence connector. The new method works quite well when there are no attachments. However, when a page includes attachments, Ragflow is unable to retrieve all the files correctly, and the generated attachment filenames are also incorrect. Can you make some test and fix it?

Jonah879:
@Magicbook1108 Would you like to take a second look at it? The retrieval, naming and sync should work properly now.

Magicbook1108:
@Jonah879 
**Confluence** is greate to merge.
**Notion** has a bit truble, I don't think it currently works well. Here I have a few screenshots.
image
image

Jonah879:
@Magicbook1108 Thanks for the feedback. The notion connector should be fixed as well.

Magicbook1108:
@Jonah879 Well done, all passed. Thanks !

PR #11766, opened 2025-12-05
Feat: Add ProgressTracker and vlm-direct backend for MinerU integration

shaoqing404:
# Feat: Add ProgressTracker, stall-aware timeouts, and vlm-direct backend for MinerU integration

## Summary

This PR introduces a real-time progress tracking mechanism, an intelligent stall-based timeout strategy, and a new `vlm-direct` backend that connects directly to a vLLM server. It also enhances the existing MinerU API mode with streaming response, HTTP session reuse, automatic retries, heartbeats, and dynamic timeouts.

These changes collectively address the long-standing issue where RagFlow tasks hang indefinitely when using the MinerU API backend, even after the API responds with `200 OK`.

---

## What problem does this PR solve?

When RagFlow v0.22.1 is configured to use the MinerU API backend, PDF parsing tasks may never return despite MinerU having already responded successfully. The underlying issues include:

1. `_run_mineru_api()` relies on a blocking `requests.post(timeout=1800)` call without any progress feedback.
2. There is no heartbeat or stall detection, so RagFlow cannot distinguish active processing from a deadlock.
3. The fixed 30-minute timeout is blind to real progress and cannot detect prolonged stalls.

As a result, tasks may hang indefinitely with logs stuck at:

```
[MinerU] invoke api: http://0.0.0.0:8000/file_parse
# meanwhile mineru-api logs:
# "POST /file_parse HTTP/1.1" 200 OK
```

This PR resolves the deadlock, improves observability, and introduces more robust backend behaviors.

---

### 1. ProgressTracker (new)

A dedicated helper class to parse progress from MinerU stdout and maintain stall-aware state:

* Supports both TQDM-style output and `Page X/Y` patterns.
* Implements **intelligent stall-based timeout** that resets when progress updates.
* Issues a warning if no progress is detected within 5 minutes (configurable).
* Reports progress through RagFlow’s callback interface.

### 2. `vlm-direct` backend (new)

A backend option that bypasses `mineru-api` entirely and connects directly to a vLLM server.

**Benefits:**

* Avoids API-layer deadlocks.
* Provides explicit timeout and connection control.
* Automatically falls back to executable mode if direct execution is unavailable.

### 3. Enhanced API backend behavior

The existing HTTP backend now includes:

* **Streaming downloads** (1 MB chunks) with progress callbacks.
* **HTTP session reuse** with retry logic (3 retries, exponential backoff).
* **Heartbeats** during long-running operations (every 30 seconds).
[...]

PR #11850, opened 2025-12-09
Fix/mineru missing images for full text chunk

shaoqing404:
### Problem
When using MinerU with the `vlm-http-client` backend, pure text blocks (text/list/header/code) do not generate corresponding image files—only bbox coordinate information is preserved. However, RAGFlow's downstream processing (e.g., frontend display, MinIO storage) requires these image paths, causing text blocks to fail in display or reference.

### Solution
Added a fallback mechanism in `deepdoc/parser/mineru_parser.py`:
- After parsing MinerU output, detect text-type blocks missing `img_path`
- Use locally cached PDF page images (`self.page_images`) to crop images based on bbox coordinates
- Save generated images to `generated_images/` directory and update the `img_path` field

### Technical Details
1. **New `_bbox_to_pixels` method**: Intelligently converts MinerU bbox coordinates (supports 0-1 normalized, 0-1000 normalized, and pixel value formats)
2. **New `_generate_missing_images` method**: Batch processes text blocks missing images, crops and saves as JPEG format
3. **Updated `_read_output` method**: Calls image generation logic before returning parsed results
4. **New content types**: Added `HEADER` and `PAGE_NUMBER` types to `MinerUContentType` enum

### Impact Scope
- Only affects MinerU parsing flow, does not change other OCR engine behavior
- Does not process blocks that already have images (e.g., tables, equations)
- Performance impact: Adds one image crop and save operation per text block (typically < 10ms/block)

---

## Changes
- **Modified**: `deepdoc/parser/mineru_parser.py` (+70 lines)
  - Added `MinerUContentType.HEADER` and `MinerUContentType.PAGE_NUMBER` enum values
  - Added `_bbox_to_pixels()` helper method for coordinate conversion
  - Added `_generate_missing_images()` method for fallback image generation
  - Updated `_read_output()` to call image generation before returning

## Testing
- Syntax validation: ✅ Passed (`python -m py_compile`)
- Manual testing: Pending user verification with real PDF documents
- Expected behavior: `generated_images/` folder created with cropped JPEG images for text blocks

## Related Issues
- Fixes missing image paths for MinerU text blocks in RAGFlow
- Ensures compatibility with frontend display and MinIO storage requirements

### Type of change

- [ ✅] Bug Fix (non-breaking change which fixes an issue)

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -849,7 +849,11 @@ class MinerUParser(RAGFlowPdfParser):
             return
         img_root = subdir / "generated_images"
         img_root.mkdir(parents=True, exist_ok=True)
-        text_types = {"text", "list", "code", "header"}
+        text_types = {
+            "text", "list", "code", "header", "equation",
+            MinerUContentType.TEXT, MinerUContentType.LIST, MinerUContentType.CODE, 
+            MinerUContentType.HEADER, MinerUContentType.EQUATION
+        }
         generated = 0
         for idx, item in enumerate(outputs):
             if item.get("type") not in text_types:


yongtenglei:
Hi, @shaoqing404 

Thank you for your contribution. Here is one minor suggestion: `vlm-direct` is not a valid backend of MinerU, as far as I know.

I think it might be better to simplify the code to focus on the real issue here, namely `missing images for full text chunk`.

What do your think?

Cheers!

shaoqing404:
> Hi, @shaoqing404
> 
> Thank you for your contribution. Here is one minor suggestion: `vlm-direct` is not a valid backend of MinerU, as far as I know.
> 
> I think it might be better to simplify the code to focus on the real issue here, namely `missing images for full text chunk`.
> 
> What do your think?
> 
> Cheers!

You may be looking for this: the patch I built on top of it, available here: https://github.com/infiniflow/ragflow/pull/11766

I can submit a new PR from a clean branch specifically for this issue, and separate it from the previous fix.

yongtenglei:
> > Hi, @shaoqing404
> > Thank you for your contribution. Here is one minor suggestion: `vlm-direct` is not a valid backend of MinerU, as far as I know.
> > I think it might be better to simplify the code to focus on the real issue here, namely `missing images for full text chunk`.
> > What do your think?
> > Cheers!
> 
> You may be looking for this: the patch I built on top of it, available here: #11766
> 
> I can submit a new PR from a clean branch specifically for this issue, and separate it from the previous fix.

Yeah, submitting a specific PR for one thing is a good approach for both of us. Keeping a PR small and clean makes it easier for us to review and really understand what has changed.

For ProgressTracker, I think there is a better approach: making it a common module as a tool, so that other components can easily use it as well, including MinerU, of course, rather than having it be tightly coupled with MinerU.

For connecting to VLLM directly, could we please hold off for now? Once it's confirmed as a real requirement, we can make it happen. The reason is that I can see this feature significantly increasing the complexity of the implementation.

What are your thoughts? Let me know. Thanks again for your contribution!

Cheers!

PR #11880, opened 2025-12-10
Fix `RuntimeError: asyncio.run() cannot be called from a running event loop` when calling mindmap endpoint.

N0bodycan:

Fix RuntimeError when calling mindmap endpoint by converting `gen_mindmap()` to async function and using `await` instead of `asyncio.run()`.

### Type of change


api/apps/conversation_app.py
buua436, line 1: Could you apply the same update at line 1138 in api/apps/sdk/session.py?

PR #11943, opened 2025-12-15
Refactor: Improve the logic to calculate embedding total token count 

Woody-Hu:

 Improve the logic to calculate embedding total token count 

### Type of change

- [x] Refactoring


meetcharan-practice:
@Woody-Hu - Can you please check this? #11956

PR #11983, opened 2025-12-17
Feat/messages

Lynn-Inf:

CRUD of messages.

### Type of change


api/constants.py
```diff
@@ -26,3 +26,4 @@
 FILE_NAME_LEN_LIMIT = 255
 MEMORY_NAME_LIMIT = 128
 MEMORY_SIZE_LIMIT = 10*1024*1024 # Byte
+MEMORY_VECTOR_DIMENSION = 1024
```
JinHai-CN: The dimension is decided by embedding model.

PR #12046, opened 2025-12-19
feat: supports multiple retrieval tool under an agent

JimmyBenKlieve:

Add support for multiple Retrieval tools under an agent

### Type of change

### DIFF
--- a/web/src/pages/agent/canvas/node/toolbar.tsx
+++ b/web/src/pages/agent/canvas/node/toolbar.tsx
@@ -7,7 +7,7 @@ import {
 import { cn } from '@/lib/utils';
 import { Position } from '@xyflow/react';
 import { Copy, Play, Trash2 } from 'lucide-react';
-import { PropsWithChildren } from 'react';
+import { MouseEventHandler, PropsWithChildren, useCallback } from 'react';
 import { Operator } from '../../constant';
 import { useDuplicateNode } from '../../hooks';
 import useGraphStore from '../../store';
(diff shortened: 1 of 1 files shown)


web/src/pages/agent/utils.ts
```diff
@@ -44,6 +44,44 @@ import { ParserFormSchemaType } from './form/parser-form';
...
+export function randomId(
+  lengthOrType?: 'human' | number,
+  patternOrOptions?: HumanIdOptions | string,
+) {
+  if (lengthOrType === 'human') {
+    return humanId(patternOrOptions);
+  }
+
+  // eslint-disable-next-line eqeqeq
+  const len =
```
cike8899: Use UUID library directly,    "uuid": "^9.0.1",

web/src/pages/agent/canvas/node/toolbar.tsx
```diff
@@ -55,47 +49,42 @@ export function ToolBar({
...
-        <section className="flex gap-2 items-center text-text-secondary">
+        <section className="flex gap-2 items-center text-text-secondary pb-2">
           {showRun && (
             <IconWrapper>
               <Play className="size-3.5" data-play />
             </IconWrapper>
           )}
           {showCopy && (
-            <IconWrapper onClick={handleDuplicate}>
+            <IconWrapper
```
cike8899: These two places seem to have no change from the previous logic, comparing deleteNode and handleDuplicate.

JimmyBenKlieve: Reverted these two

web/src/pages/agent/utils.ts
JimmyBenKlieve: Removed this whole function and replaced with `human-id` package

cike8899 (approved):

web/src/pages/agent/utils.ts
JimmyBenKlieve: Removed `randomId()` and replaced with `human-id` packages as per what currently using

PR #12052, opened 2025-12-19
Remove Chinese comments and fix function arguments errors

JinHai-CN:

As title

### Type of change

- [x] Refactoring


levischd:
This broke the Moodle Data Source Connector. Error: validation error for Document primary_owners.

PR #12137, opened 2025-12-23
Fix: further fix #12110 

Magicbook1108:

Fix: further fix #12110 
### Type of change

### DIFF
--- a/agent/component/agent_with_tools.py
+++ b/agent/component/agent_with_tools.py
@@ -386,7 +386,7 @@ class Agent(LLM, ToolBase):
                         raise TypeError(f"An object type should be returned, but `{f}`")
 
                 tool_tasks = []
-                for func in functions:
+                for idx, func in enumerate(functions):
                     name = func["name"]
                     args = func["arguments"]
                     if name == COMPLETE_TASK:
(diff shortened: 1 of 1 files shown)


agent/component/agent_with_tools.py
```diff
@@ -386,15 +386,15 @@ def append_user_content(hist, content):
...
+                for idx, func in enumerate(functions):
                     name = func["name"]
                     args = func["arguments"]
                     if name == COMPLETE_TASK:
                         append_user_content(hist, f"Respond with a formal answer. FORGET(DO NOT mention) about `{COMPLETE_TASK}`. The language for the response MUST be as the same as the first user request.\n")
                         async for txt, tkcnt in complete():
                             yield txt, tkcnt
                         return
-
+                    name = f"{name}_{idx}"
```
buua436, line 397: The functions here are derived from the next_step_async response and are not the same list as in #12110, so appending an index via name = f"{name}_{idx}" does not appear to be safe.

It would be better to first verify whether name is already in the {name}_{idx} format before attempting to modify it.

Since use_tool_async simply forwards name, the responsibility for resolving the tool name should be handled either within tool_call_async or by normalizing tool names during initialization (for example, in __init__), rather than mutating name at the call site.

PR #6130, opened 2025-03-15
Feat: support agent version history.

isthaison:
### What problem does this PR solve?
Add history version save
- Allows users to view and download agent files by version revision history
image

_Briefly describe what this PR aims to solve. Include background context that will help reviewers understand the purpose of the PR._

### Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

PR #12418, opened 2026-01-04
Chore(compose): remove unnecessary history_data_agent volume mount

asiroliu:

Removed the volume mount mapping ../history_data_agent:/ragflow/history_data_agent from docker-compose.yml as it appears to be no longer in use

### Type of change

- [x] Chore


asiroliu:
import from #6130

PR #12430, opened 2026-01-04
feat: Add a custom header to the SDK for chatting with the agent.

LeiZhang-Hunter:

As title.

### Type of change


asiroliu:
@LeiZhang-Hunter 
1.  I checked your PR—the changes are actually for the Web API, not the SDK. 
2. I'm concerned that custom headers could open us up to HTTP Header Injection and parsing risks, which aren't being handled in the current code.

LeiZhang-Hunter:
> * I checked your PR—the changes are actually for the Web API, not the SDK.
> 
> * I'm concerned that custom headers could open us up to HTTP Header Injection and parsing risks, which aren't being handled in the current code.

ragflow only doing pass-through; you don't need to worry about these things.

PR #12410, opened 2026-01-04
Refactor: UmiJs -> Vite

dcc123456:
### What problem does this PR solve?

Refactor: UmiJs -> Vite+React

### Type of change

- [x] Refactoring

PR #12459, opened 2026-01-06
Fix:  The avatar and greeting message no longer appear in the Agent iFrame. [#12410]

dcc123456:
Fix:  The avatar and greeting message no longer appear in the Agent iFrame. [#12410]

### Type of change


meetcharan-practice:
@dcc123456 - I’ve verified the changes after applying the fix, and everything is working as expected. Thanks for the quick turnaround :)

PR #12472, opened 2026-01-06
docker: merge apt installs into a single layer

boralp:

The Dockerfile previously installed system packages using multiple separate apt install commands.
This created unnecessary image layers and reduced build cache efficiency.
The change consolidates these installs into a single layer, making the image build more efficient without altering functionality or dependencies.

### Type of change

- [x] Performance Improvement


JinHai-CN:
Thanks for the PR. Could you share the reason for adding build-essential to the Dockerfile?

wangq8:
main already have the logic.

PR #12473, opened 2026-01-06
Track Microsoft repository upgrade for Ubuntu 24.04

boralp:

This PR documents the current mismatch between the base image (ubuntu:24.04) and the Microsoft APT repository, which is still pinned to Ubuntu 22.04.

### Type of change

- [x] Other (please describe): Infrastructure / dependency tracking

This change does not alter runtime behavior. It only documents the
current state and prevents silent breakage when Ubuntu 22.04 support
is eventually dropped from the Microsoft APT repository.


JinHai-CN:
This PR also includes the update of #12472

wangq8:
stale, CI cannot pass

PR #12504, opened 2026-01-08
Fix (api): fix the document parsing status check logic

huang-aoqin:
When the original code terminates the parsing task halfway, the progress may not be 0 or 1, which will result in the inability to call the interface to parse again

-Change the document parsing progress check to task status check, and use TaskStatus.RUNNING.value to judge 
-Update the condition judgment for stopping parsing documents, and check whether the task is running instead 

### Type of change


KevinHuSh:
Appreciations!
What do you mean by `the progress may not be 0 or 1`?

huang-aoqin:
I mean the 'progress' column in 'document' table not be 0 or 1,when cancel a task halfway.
like this data
image

KevinHuSh:
See. Please resolve the conflicts.

huang-aoqin:
Resolved

PR #12509, opened 2026-01-08
fix(entrypoint): Preserve $ in passwords during template expansion

zagnaan:

Fix shell variable expansion to preserve $ in password defaults when 
env vars are unset. Fixes Azure RDS auto-rotated passwords (that contain $) being 
truncated during template processing.

### Type of change


KevinHuSh:
Appreciations!
By our evalutions, this feature is not going to be merged into main branch.

gileslloyd:
@KevinHuSh I'm having a fatal error with ragflow due to the bug this PR attempts to resolve. Is there a fix incoming for this or an alternative you'd recommend? There's no explanation as to why this PR was rejected

PR #12513, opened 2026-01-08
feat: add paddleocr parser

Bobholamovic:

Add PaddleOCR as a new PDF parser.

### Type of change


buua436:
Hello, thank you for submitting this PR. We’ve run some basic tests and can confirm that the PaddleOCR parser works as expected.

As a suggestion for improvement, it would be great if the parser could additionally provide positional information for each chunk, along with the corresponding image snippets, as this would help better demonstrate the parsing results.

In addition, from a user experience perspective, it would be helpful to include documentation on how to set up the PaddleOCR service and configure the parser accordingly, so users can get started more easily.

Thanks again for the contribution, and we look forward to further improvements.

PR #12563, opened 2026-01-12
Align p3 HTTP/SDK tests with current backend behavior

6ba3i:

Updates pre-existing HTTP API and SDK tests to align with current backend behavior (validation errors, 404s, and schema defaults). This ensures p3 regression coverage is accurate without changing production code.

### Type of change

- [x] Other (please describe): align p3 HTTP/SDK tests with current backend behavior

### DIFF
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -21,7 +21,7 @@ from pathlib import Path
 from quart import Blueprint, Quart, request, g, current_app, session
 from itsdangerous.url_safe import URLSafeTimedSerializer as Serializer
 from quart_cors import cors
-from common.constants import RetCode, StatusEnum
+from common.constants import StatusEnum
 from api.db.db_models import close_connection, APIToken
 from api.db.services import UserService
 from api.utils.json_encode import CustomJSONEncoder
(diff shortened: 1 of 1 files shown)


test/testcases/test_http_api/test_chunk_management_within_dataset/test_add_chunk.py
```diff
@@ -169,9 +170,16 @@ def test_invalid_dataset_id(
         expected_message,
     ):
         _, document_id = add_document
-        res = add_chunk(HttpApiAuth, dataset_id, document_id, {"content": "a"})
-        assert res["code"] == expected_code
-        assert res["message"] == expected_message
+        url = f"{HOST_ADDRESS}{CHUNK_API_URL}".format(dataset_id=dataset_id, document_id=document_id)
+        res = requests.post(url=url, headers=HEADERS, auth=HttpApiAuth, json={"content": "a"})
+        if res.status_code == 404:
```
asiroliu: Ideally, the RAGFlow service should never return a 404 status code. If it does, it means something is wrong with the service itself.

asiroliu (approved):

Issue #11460, opened 2025-11-21
[Feature Request]: Deleted files from data source (S3 or S3 compatible) are still exists in Ragflow Dataset


Furkan-Demir:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

Version: v0.22.1 (Docker, CPU)

Deleted files in the data source are not removed from the Ragflow dataset. Is it bug or out-of-scope?

Steps to Reproduce:

1. Add an S3 (or S3-compatible) bucket as a data source.
2. Upload several files to the bucket, let's say `a.pdf`, `b.pdf`, `c.pdf`.
3. Wait for Ragflow’s sync job to complete.
4. Confirm that the files appear in the Ragflow dataset (sync works as expected).
5. Delete `b.pdf` from the data source.
6. Wait for the next Ragflow sync job.
7. Observe that `b.pdf` still appears in the Ragflow dataset.

Expected Behavior:

Deleted files in the data source should also be removed from the Ragflow dataset after the sync job.

Actual Behavior:

The deleted file (`b.pdf`) remains in the Ragflow dataset even after multiple sync cycles.

KevinHuSh:
During increamental synchronization, it's not easy to delete removed files. We're working on it.

ZhenhangTung:
Hi! I'm the product manager. Thanks for this feedback. Why do you want this feature? Could you share your use cases? That will help us fully understand your problem.

Furkan-Demir:
> Hi! I'm the product manager. Thanks for this feedback. Why do you want this feature? Could you share your use cases? That will help us fully understand your problem. @ZhenhangTung 

Hello. We have some documents that changes the content inside frequently. And sometimes, we want to delete some files in data source because no needs for these documents. 


> During increamental synchronization, it's not easy to delete removed files. We're working on it. @KevinHuSh 

Thanks for your work. By the way, I'm going to review this changes and test it in like 3-4 hours. [PR11468](https://github.com/infiniflow/ragflow/pull/11468)

[...]

PR #12566, opened 2026-01-12
Feat: replace existing files from data source

Jonah879:

This PR serves as a starting point for further development to increase the integration of data sources as mentioned here #11460.

Problem:
Right now RAGflow does not update/replace existing files with the name, but will sync and parse a second file with a suffix "(1)" from the configured data source.

Solution:
The sync service checks for duplicated files (files with the exact same name) and replaces/updates the file if it already exists in the knowledgebase. If the file is not in the knowledge base it will be synced normally.

Additional information:
The change is done in the base sync, so it applies to all data sources regardless of it is desired or not. In a next step one could make an option to toggle this replacement functionality on/off.

### Type of change


KevinHuSh:
Appreciation!
What about hashing the name of document into 32 bytes ID?
By this way, we just need to update the blob of file if it exists.

PR #12606, opened 2026-01-14
fix(docker): remove unnecessary mounts in docker-compose.yml

maoxiaowang:

This PR removes the unnecessary volume mounts of `entrypoint.sh` and `service_conf.yaml.template` in docker-compose. 
Both files are already included in the Dockerfile, so mounting them is not needed. 
Mounting entrypoint.sh in particular may cause permission issues.

- service_conf.yaml.template mount is kept commented for optional local override.
- For local testing, consider using docker-compose.override.yml instead of mounting directly.

### Type of change


KevinHuSh:
Appreciations!
By this way, former user and CI is gona fail.

PR #12620, opened 2026-01-14
Fix : Web API tests by normalizing errors, validation, and uploads

6ba3i:

Fixes web API behavior mismatches that caused test failures by normalizing error responses, tightening validations, correcting error messages, and closing upload file handles.

### Type of change

### DIFF
--- a/test/testcases/test_web_api/test_chunk_app/test_retrieval_chunks.py
+++ b/test/testcases/test_web_api/test_chunk_app/test_retrieval_chunks.py
@@ -265,11 +265,11 @@ class TestChunksRetrieval:
     @pytest.mark.parametrize(
         "payload, expected_code, expected_highlight, expected_message",
         [
-            ({"highlight": True}, 0, True, ""),
-            ({"highlight": "True"}, 0, True, ""),
-            pytest.param({"highlight": False}, 0, False, "", marks=pytest.mark.skip(reason="issues/6648")),
-            pytest.param({"highlight": "False"}, 0, False, "", marks=pytest.mark.skip(reason="issues/6648")),
-            pytest.param({"highlight": None}, 0, False, "", marks=pytest.mark.skip(reason="issues/6648")),
+            pytest.param({"highlight": True}, 0, True, "", marks=pytest.mark.skip(reason="highlight not functionnal")),
+            pytest.param({"highlight": "True"}, 0, True, "", marks=pytest.mark.skip(reason="highlight not functionnal")),
+            ({"highlight": False}, 0, False, ""),
+            ({"highlight": "False"}, 0, False, ""),
+            ({"highlight": None}, 0, False, "")
         ],
     )
     def test_highlight(self, WebApiAuth, add_chunks, payload, expected_code, expected_highlight, expected_message):


api/apps/dialog_app.py
```diff
@@ -42,7 +42,7 @@ async def set_dialog():
     if len(name.encode("utf-8")) > 255:
         return get_data_error_result(message=f"Dialog name length is {len(name)} which is larger than 255")
 
-    if is_create and DialogService.query(tenant_id=current_user.id, name=name.strip()):
+    if is_create and name.strip().isascii() and DialogService.query(tenant_id=current_user.id, name=name.strip()):
```
KevinHuSh: >>> "信息".strip().isascii()
False

api/db/db_models.py
```diff
@@ -395,6 +395,9 @@ def __init__(self):
             'max_retries': 5,
             'retry_delay': 1,
         }
+        if settings.DATABASE_TYPE.upper() == "MYSQL":
```
KevinHuSh: It's gona cause upgrading issue.

api/utils/api_utils.py
```diff
@@ -124,16 +128,10 @@ def server_error_response(e):
     try:
         msg = repr(e).lower()
         if getattr(e, "code", None) == 401 or ("unauthorized" in msg) or ("401" in msg):
-            return get_json_result(code=RetCode.UNAUTHORIZED, message=repr(e))
+            return get_json_result(code=RetCode.UNAUTHORIZED, message="Unauthorized"), RetCode.UNAUTHORIZED
```
KevinHuSh: 2 values returned?

api/apps/chunk_app.py
```diff
@@ -223,14 +228,27 @@ def _rm_sync():
...
+            if isinstance(deleted_chunk_ids, list):
+                unique_chunk_ids = list(dict.fromkeys(deleted_chunk_ids))
+                has_ids = len(unique_chunk_ids) > 0
+            else:
+                unique_chunk_ids = [deleted_chunk_ids]
+                has_ids = deleted_chunk_ids not in (None, "")
+            if has_ids and deleted_count == 0:
+                return get_data_error_result(message="Index updating failure")
+            if deleted_count > 0 and deleted_count < len(unique_chunk_ids):
+                deleted_count += settings.docStoreConn.delete({"doc_id": req["doc_id"]},
```
KevinHuSh, line 248: Why deleting 2 times?

api/utils/api_utils.py
6ba3i: The first value returns the json body and the second one returns the status code.
A similar result could be achieved by replacing the line with
``` 
resp = get_json_result(code=RetCode.UNAUTHORIZED, message="Unauthorized")
resp.status_code = RetCode.UNAUTHORIZED
return resp
``` 
We would be manually setting the HTTP status on the response object. Same result but just more verbose.

api/apps/chunk_app.py
6ba3i: First delete uses ``` doc_id + chunk_ids```  to remove only the requested chunks. If only some are deleted, we delete by ``` doc_id```  alone to wipe all chunks for that doc and restore consistency.

PR #12622, opened 2026-01-14
fixed contains metadata filters

HeyPhiS:

Changed common/metadata_utils.py so contains/not contains now look for the filter value inside the metadata (checking any list item) instead of the previous behavior where metadata had to be a substring of the filter value; in/not in remain unchanged to keep membership semantics. 
Before this change, both operator pairs (contain/in and not_contain/not_in) were identical, and contains was unintuitively backwards.
This fix restores distinct, expected semantics for those filters.

### Type of change

### DIFF
--- a/common/metadata_utils.py
+++ b/common/metadata_utils.py
@@ -62,9 +62,10 @@ def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
             matched = False
             try:
                 if operator == "contains":
-                    matched = input in value if not isinstance(input, list) else all(i in value for i in input)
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!
The `input` should be the left value for the operators.

HeyPhiS:
I am sorry but I do not understand the reason for closing this MR.

Withought this PR the **filters "contain" and "in" are identical**. This can not be correct. 
With this PR, the filter "contain" is changed to check if the filter _value_ is contained in the metadata _input_.
The filter "in" stays as is and checks if the metadata _input_ is part of the filter _value_.

PR #12663, opened 2026-01-16
Fix: pass request_timeout parameter in MCPToolCallSession.tool_call

9Ashwin:

The `timeout` parameter passed to `MCPToolCallSession.tool_call()` was not forwarded to `_call_mcp_tool()`, causing the inner layer to always use the default 10-second timeout instead of the caller-specified value (e.g., 60 seconds from Agent calls). This resulted in premature timeouts when calling MCP tools that take longer than 10 seconds to execute.

**Call chain showing the bug:**
```
Agent: tool_call(name, args, 60s)
  └─> tool_call(timeout=60)
       └─> _call_mcp_tool()  ← Bug: request_timeout not passed, uses default 10s
            └─> _call_mcp_server(request_timeout=10)
```

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### How this fix works

Pass the `timeout` parameter as `request_timeout` to `_call_mcp_tool()`, ensuring the caller-specified timeout is respected throughout the call chain.

This makes `tool_call()` consistent with `get_tools()`, which already correctly passes `request_timeout`:

```python
# get_tools() - correct pattern (already exists)
future = asyncio.run_coroutine_threadsafe(
    self._get_tools_from_mcp_server(request_timeout=timeout), ...
)

# tool_call() - now matches the pattern
future = asyncio.run_coroutine_threadsafe(
    self._call_mcp_tool(name, arguments, request_timeout=timeout), ...
)
```

### Test plan

- [x] Code review confirms the change matches the pattern used in `get_tools()` method
- [ ] Manual testing with MCP tools that take >10 seconds to execute

### DIFF
--- a/common/mcp_tool_call_conn.py
+++ b/common/mcp_tool_call_conn.py
@@ -207,7 +207,7 @@ class MCPToolCallSession(ToolCallSession):
         if self._close:
             return "Error: Session is closed"
 
-        future = asyncio.run_coroutine_threadsafe(self._call_mcp_tool(name, arguments), self._event_loop)
+        future = asyncio.run_coroutine_threadsafe(self._call_mcp_tool(name, arguments, request_timeout=timeout), self._event_loop)
         try:
             return future.result(timeout=timeout)
         except FuturesTimeoutError:


9Ashwin:
Closing to recreate with cleaner description

PR #12673, opened 2026-01-17
Feat(config): Introduce pydantic-based config for runtime configuration

maoxiaowang:

This PR adds a typed runtime configuration module (`core/config`) and a Python YAML renderer (`render_yaml.py`) to make configuration safer, clearer, and easier to maintain.

No new libraries are introduced. It only uses existing `pydantic` and `ruamel.yaml`.
All changes are **standalone**, not wired into runtime yet, and **do not affect existing business logic**.

---

## Motivation

- Configuration logic is spread across YAML files, environment variables, and global code.
- Lack of type validation makes misconfiguration hard to detect.
- Shell-based YAML rendering is fragile and can behave differently across environments.
- We need a deterministic and inspectable configuration source.

This PR provides a **typed and validated configuration foundation** without changing current behavior.

---

## Core Design Principles

The implementation of `AppConfig` builds on **the existing configuration logic** and makes it explicit and well-defined.

### YAML is the Source of Truth

- The `service_conf.yaml` defines the main configuration structure and baseline values.
- Environment variables are used only for selected fields or scenarios and are not a blanket override mechanism.
- AppConfig merges all supported inputs and performs final validation and normalization.

Why:

- In container environments, `service_conf.yaml` is generated from a template using env vars.
- After rendering, the YAML already reflects the correct configuration.
- In non-container environments, `service_conf.yaml` is used directly, optionally with `.env` files.
- This keeps behavior consistent across environments.

> Once YAML is loaded, it is treated as final input.

### Role of Environment Variables

Environment variables are only used for:

- Rendering YAML templates before runtime
- Providing values for explicitly defined `AppConfig` fields
- Controlling feature flags or optional behavior when not specified in YAML

All supported environment-based configuration is explicitly declared in `core/config/env_overrides.py`.

---

### 1. AppConfig

- Introduces `AppConfig`, a Pydantic-based runtime configuration model.
- Loads configuration from `service_conf.yaml`.
- Enforces strict typing and validation.
- Exposes a single global `app_config` object.

#### Usage Example

```python
from core.config import app_config

app.run(
    host=app_config.ragflow.host,
    port=app_config.ragflow.http_port,
)

mysql_cfg = app_config.database.mysql.model_dump()

if app_config.doc_engine.is_infinity:
    pass
```

### 2. YAML Rendering

- Replaces shell-based YAML rendering.
- Avoids `eval`, shell expansion, and platform differences.
- Produces deterministic and type-safe YAML output.

#### Rendering Behavior Comparison

| Template | Env | Shell Result | Python Result |
|--------|-----|--------------|---------------|
| `a: "${FOO}"`<br>`b: ""`<br>`c:` | `FOO=""` | `a:`<br>`b:`<br>`c:` *(empty → null)* | `a: ""`<br>`b: ""`<br>`c: null` |
| `a: "${FOO_BAR}"`<br>`b: "${FOO-BAR}"` | `FOO=1`<br>`BAR=2` | `a:` <br> `b: 1` *(unpredictable)* | `a: ""`<br>`b: ""` |

### 3. YAML Structure

`AppConfig` is designed around a **new structured YAML format**, while still supporting legacy formats.

- Legacy YAML continues to work without changes
- New format is recommended but not mandatory

#### Recommended Structure

```yaml
database:
  active: mysql
  mysql:
    user: mysql
    password: mysql
  postgres:
    user: postgres
    password: postgres

doc_engine:
  active: elasticsearch
  elasticsearch:
    user: elastic
    password: elastic
```

Benefits:

- Clear provider selection (`active`)
- Better readability
- Stronger validation

#### YAML vs Environment Variables

If a value exists in YAML, it always wins.

```yaml
database:
  active: postgres
```

```shell
export DB_TYPE=mysql
```

Environment variables are ignored in this case.

#### New vs Legacy Formats

New and legacy formats **must not be mixed** in the same domain.

Invalid example:

```yaml
database:
  mysql:
    user: mysql
    password: mysql

oceanbase:
  user: oceanbase
```

This raises a validation error:

```text
Legacy config keys ('mysql', 'postgres', 'oceanbase') conflict with new config 'database'.
Please remove the legacy configuration before continuing.
Only the new configuration format is supported.
```

> This avoids ambiguous resolution and ensures deterministic behavior.

---

## Testing

- 100+ unit tests added and passing
- Covers database, cache, doc_engine, etc.
- Tests both legacy and new YAML formats
- No runtime integration yet

---

## Impact

- Safe, non-breaking change
- Legacy behavior preserved
- No effect on current runtime

---

## Next Steps

- Implement a method to get the `source` (yaml/env/default) and `overriden` status of each field in `AppConfig`
- Replace shell-based YAML rendering in `entrypoints` with the Python renderer
- Gradually migrate existing code from `common.settings` to `AppConfig`
- Integrate `AppConfig` with the Provider layer in future work

---

## Type of Change

- [x] New Feature (non-breaking)
- [x] Refactoring


Harshit057:
Quick question on future integration: will AppConfig be immutable once loaded, or support controlled reloads for long-running services? Mainly thinking about observability and hot-reload scenarios.

maoxiaowang:
> Quick question on future integration: will AppConfig be immutable once loaded, or support controlled reloads for long-running services? Mainly thinking about observability and hot-reload scenarios.

Good question.
At the moment, AppConfig is loaded once and treated as immutable during runtime.

Observability is something I’m planning to improve by exposing per-field metadata (e.g. `source` and `override` status).
Hot-reload isn’t in scope for this PR, and it’s not clear yet whether it’s needed.

Harshit057:
@maoxiaowang Thanks for the clarification — that makes sense.

Treating AppConfig as immutable keeps the initial integration clean. Happy to help later with per-field source/override metadata or any follow-up work once this lands.

JinHai-CN:
@maoxiaowang would you please resolve the format issue?

PR #12681, opened 2026-01-18
fix: Correct gitlab_url access in sync_data_source.py

francisye19:

Correct gitlab_url access. See https://github.com/infiniflow/ragflow/blob/main/web/src/pages/user-setting/data-source/constant/index.tsx#L660-L666

### Type of change


Magicbook1108 (approved):
Great.

PR #12719, opened 2026-01-20
[Feat]Automatic table orientation detection and correction

zhanxin-xu:
This PR introduces automatic table orientation detection and correction within the PDF parser. This ensures that tables in PDFs are correctly oriented before structure recognition, improving overall parsing accuracy. 

### Type of change
- [x] Documentation Update


Magicbook1108:
This PR is already sufficient and in a good state to be merged. 
The results are solid when tables are at standard orientations (e.g., 90°, 180°).
However, when a table is diagonally oriented, it tends to be recognized as an image rather than a table.
image
[table_rotation_test.pdf](https://github.com/user-attachments/files/24755280/table_rotation_test.pdf)

Issue #12568, opened 2026-01-12
[Bug]: SSO via OIDC using keycloak raises Unauthorized()


Jonah879:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

675d18d

### RAGFlow image version

v0.23.1

### Other environment information

```Markdown

```

### Actual behavior

When logging in to RAGflow using SSO, you get redirected to the OIDC provider (in our case keycloak) and can enter your credentials. After submitting them, you get redirected back to the login screen.
This is the error message:
```
File "/ragflow/api/apps/init.py", line 178, in wrapper
raise Unauthorized()
quart_auth.extension.Unauthorized: 401 Unauthorized: The server could not verify that you are authorized to access the URL requested. You either supplied the wrong credentials (e.g. a bad password), or your browser doesn't understand how to supply the credentials required.
```

### Expected behavior

Logging in to RAGflow using SSO, you get redirected to your OIDC provider and enter credentials. After submitting you should be logged in to your user account.

### Steps to reproduce

```Markdown
- have SSO activated
- configure OIDC
- log in using SSO
```

### Additional information

This behavior is since the change from flask to quart #11277 
In an earlier version, where flask is used the same SSO configuration is working fine.

KevinHuSh:
You may follow dosu's instructions.

PR #12784, opened 2026-01-22
Fix: persist SSO auth token on root route loader

levischd:

This PR fixes SSO/OIDC login persistence after the Vite migration #12568. Because wrappers are ignored by React Router, the OAuth callback never stored the auth token in localStorage, causing auth to only work while ?auth= stayed in the URL. We move that logic into a route loader and remove the Bearer prefix for the signed token so the backend accepts it.

### Type of change


cike8899:
Please resolve the conflict first.

levischd:
Done 👍

cike8899:
Can SSO authentication work properly? For example, with Keycloak. If it works, please provide a screenshot.

hauberj:
Fix working with Keycloak for us.
Screenshots:
step1
step2
step3
step4

FYI @levischd @cike8899

hauberj:
Hi @cike8899 @KevinHuSh 
Do you need more information?
Just want to make sure that this PR will be merged before release of v0.23.2

Esiravegna:
I also have the same issue. I cannot use the latest versions of RagFlow because of this problem.

gwyuan:
Fix working in my env with internal sso

Esiravegna:
Same here, patched 0.23.1 with this pr and works.

levischd:
Hi @cike8899 @KevinHuSh,

please review this. SSO is essential for all Enterprise Users.

carlobeltrame:
@KevinHuSh thanks for merging! I am having the same issue. Can we get a release which includes this?

kettleofketchup:
Thanks for merging this, I was wondering if we could get a hotfix release tagged that would incorporate this as well @KevinHuSh

PR #12826, opened 2026-01-26
Put document metadata in ES/Infinity

qinling0210:

Put document metadata in ES/Infinity.

Index name of meta data: ragflow_doc_meta_{tenant_id}

### Type of change

- [x] Refactoring

### DIFF
--- a/api/db/services/doc_metadata_service.py
+++ b/api/db/services/doc_metadata_service.py
@@ -318,6 +318,9 @@ class DocMetadataService:
         """
         Update document metadata in ES/Infinity.
 
+        For Elasticsearch: Uses partial update to directly update the meta_fields field.
+        For Infinity: Falls back to delete+insert (Infinity doesn't support partial updates well).
+
         Args:
             doc_id: Document ID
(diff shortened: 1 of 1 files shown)


api/db/services/dialog_service.py
```diff
@@ -1124,7 +1124,7 @@ async def gen_mindmap(question, kb_ids, tenant_id, search_config={}):
         rerank_mdl = LLMBundle(tenant_id, LLMType.RERANK, rerank_id)
 
     if meta_data_filter:
-        metas = DocumentService.get_meta_by_kbs(kb_ids)
+        metas = DocMetadataService.get_flatted_meta_by_kbs(kb_ids)
         doc_ids = await apply_meta_data_filter(meta_data_filter, metas, question, chat_mdl, doc_ids)
```
yuzhichang, line 1128: Moving metadata into ES/infinity make it possible to search only once.

qinling0210: document meatdata is stored in a separate table, not possible to search once now

PR #12861, opened 2026-01-27
Fix metadata filter

HeyPhiS:
##### Summary
This PR fixes a bug in the metadata filtering logic where the contains and not contains operators were behaving identically to the in and not in operators. It also standardizes the syntax for string-based operators.

##### The Issue
On the main branch, the contains operator was implemented as:
`matched = input in value if not isinstance(input, list) else all(i in value for i in input)`
This logic is identical to the `in` operator. It checks if the metadata (`input`) exists within the filter (`value`). For a "contains" search, the logic should be reversed: _we want to check if the filter value exists within the metadata input_.

##### Solution Presented Here
The operators have been rewritten using str.find():
Contains: `str(input).find(value) >= 0`
Not Contains: `str(input).find(value) == -1`

##### Advantage
This approach places the metadata (input) on the left side of the expression. This maintains stylistic consistency with the existing start with and end with operators in the same file, which also place the input on the left (e.g., str(input).lower().startswith(...)).

##### Considered Alternative
In a previous PR we considered using the standard Python `in` operator: `value in str(input)`.
The `in` operator is approximately 15% faster because it uses optimized Python bytecode (CONTAINS_OP) and avoids an attribute lookup. However following rejection of this PR we now propose the change presented here.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/common/metadata_utils.py
+++ b/common/metadata_utils.py
@@ -130,6 +130,7 @@ async def apply_meta_data_filter(
         list of doc_ids, ["-999"] when manual filters yield no result, or None
         when auto/semi_auto filters return empty.
     """
+    from rag.prompts.generator import gen_meta_filter # move from the top of the file to avoid circular import
 
     doc_ids = list(base_doc_ids) if base_doc_ids else []
 
@@ -139,8 +140,6 @@ async def apply_meta_data_filter(
     method = meta_data_filter.get("method")
 
     if method == "auto":
-        from rag.prompts.generator import gen_meta_filter # move from the top of the file to avoid circular import
-
         filters: dict = await gen_meta_filter(chat_mdl, metas, question)
         doc_ids.extend(meta_filter(metas, filters["conditions"], filters.get("logic", "and")))
         if not doc_ids:


Magicbook1108 (approved):
Can you provide a screen shot demonstrate the original method is incorrect and your solution correctly handled this issue?

From my test, there is no perfromance issue between yours and the original one. They are both good.

HeyPhiS:
Thanks for the feedback @Magicbook1108 ! I’ve added a dedicated unit test file that demonstrates the intended behavior. The test suite covers all metadata filter operations, including contains and not contains.

Here screenshots showing the test results when run on the main branch
meta_filter_main
 versus this PR branch.
meta_filter_fixed
As you’ll see, the tests for contains and not contains fail on main and pass with the changes introduced in this PR.


To run these tests I had to move the import statement (from rag.prompts.generator import gen_meta_filter) inside the apply_meta_data_filter function to avoid a circular import. This adjustment should not introduce any performance issues.

Magicbook1108:
@HeyPhiS Great, I'll have a try with it.

PR #12904, opened 2026-01-29
fix: close file handles when loading JSON mapping in doc store connectors

sharpenteeth:
**What problem does this PR solve?**

When loading JSON mapping/schema files, the code used json.load(open(path)) without closing the file. The file handle stayed open until garbage collection, which can leak file descriptors under load (e.g. repeated reconnects or migrations).

**Type of change**
[x] Bug Fix (non-breaking change which fixes an issue)
**Change**
Replaced json.load(open(...)) with a context manager so the file is closed after loading:
with open(fp_mapping, "r") as f:    ... = json.load(f)

**Files updated**
rag/utils/opensearch_conn.py – mapping load (1 place)
common/doc_store/es_conn_base.py – mapping load + doc_meta_mapping load (2 places)
common/doc_store/infinity_conn_base.py – schema loads in _migrate_db, doc metadata table creation, and SQL field mapping (4 places)
Behavior is unchanged; only resource handling is fixed.

### DIFF
--- a/common/doc_store/es_conn_base.py
+++ b/common/doc_store/es_conn_base.py
@@ -48,7 +48,8 @@ class ESConnectionBase(DocStoreConnection):
             msg = f"Elasticsearch mapping file not found at {fp_mapping}"
             self.logger.error(msg)
             raise Exception(msg)
-        self.mapping = json.load(open(fp_mapping, "r"))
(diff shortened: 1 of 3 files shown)


Lynn-Inf (approved):
LGTM

Issue #12889, opened 2026-01-29
[Bug]: Jira case updated, file updated successfully, but vector database not updated


BruceDou723:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

V0.23.1

### RAGFlow image version

V0.23.1

### Other environment information

```Markdown

```

### Actual behavior

Jira case updated, file updated successfully, but vector database not updated.
1. update a issue in jira
2. find the sync log success run
3. check the case in dataset, the content on the left side is updated, but the chunk result on the right side is not updated.

### Steps to reproduce

```Markdown
1. update a issue in jira
2. find the sync log success run
3. check the case in dataset, the content on the left side is updated, but the chunk result on the right side is not updated.
```

Crystora:
@KevinHuSh @Magicbook1108 I am interested in this issue. May I pick this up?

BruceDou723:
it's OK. Thanks.

Crystora:
@KevinHuSh 
Please review my PR

PR #12918, opened 2026-01-30
fix: re-chunk documents when data source content is updated

bittoby:
Closes: #12889 

When syncing external data sources (e.g., Jira, Confluence, Google Drive), updated documents were not being re-chunked. The raw content was correctly updated in blob storage, but the vector database retained stale chunks, causing search results to return outdated information.

**Root cause:** The task digest used for chunk reuse optimization was calculated only from parser configuration fields (`parser_id`, `parser_config`, `kb_id`, etc.), without any content-dependent fields. When a document's content changed but the parser configuration remained the same, the system incorrectly reused old chunks instead of regenerating new ones.

**Example scenario:**
1. User syncs a Jira issue: "Meeting scheduled for Monday"
2. User updates the Jira issue to: "Meeting rescheduled to Friday"
3. User triggers sync again
4. Raw content panel shows updated text ✓
5. Chunk panel still shows old text "Monday" ✗

**Solution:**
1. Include `update_time` and `size` in the chunking config, so the task digest changes when document content is updated
2. Track updated documents separately in `upload_document()` and return them for processing
3. Process updated documents through the re-parsing pipeline to regenerate chunks

[1.webm](https://github.com/user-attachments/assets/d21d4dcd-e189-4d39-8700-053bae0ca5a0)

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


Crystora:
@KevinHuSh @Magicbook1108 Could you review this PR? thank you

KevinHuSh:
Appreciations!
`FileService.upload_document` is invocated in many places. So, the best solution maybe is to work aroud rather than altering the return value of it.

Crystora:
@KevinHuSh Please review again. I updated based on your feedback. Thanks

Magicbook1108:
Hello, I think you should take a look at the function below, in` api/db/services/connector_service.py`. 
I believe this is enough for re-chunking. 
My network sucks and cannot to jira service today.
image

Crystora:
@Magicbook1108  Thanks for the feedback.  The original `connector_service.py` logic was fine, but it simply never got the right input for re‑chunking updated documents.
when a document already existed,` FileService.upload_document()` updated the blob in storage but did not add that document into the list passed to `duplicate_and_parse()`, so it was never re‑chunked. My change makes sure updated documents are included in that list, and also makes the digest depend on update_time/size so old chunks aren’t reused. With those two pieces, the existing `duplicate_and_parse()` behavior is now enough to correctly re‑chunk updated docs.

Crystora:
I confirmed it is working well. I uploaded recording file in the PR description. Please confirm on your side and let me know.
Thanks

bittoby:
@KevinHuSh would appreciate to merge this pr. thanks

bittoby:
@Magicbook1108 @KevinHuSh Sorry for tagging you again. It would appreciate to merge this PR, thanks!

bittoby:
@KevinHuSh please review this PR. its ready to merge now. thanks

Magicbook1108:
I think relying on updated_time is unsafe. There are too many scenarios where the updated_time field can be modified unintentionally.

Instead, I suggest using:
1. cls.model.size
2. content_hash (computed from the blob content)

Crystora:
@Magicbook1108 Thanks for your feedback. I updated with `content_hash`. would appreciate to review again.

Crystora:
@Magicbook1108 Please give me any updates. I opened this PR a month ago. I want to merge the pr. 
thank you

Magicbook1108:
lgtm

Crystora:
@Magicbook1108 @yingfeng could you please merge this pr? thanks

Crystora:
This testing issue is not relate my code. It is likely pre-existing or flaky. could you please review test error log? @yingfeng 
thank you

PR #12936, opened 2026-02-02
Fix duplicate POSTGRES enum entry causing backend startup failure    

LiruiYu33:
                                                                                                                                 
  Fixes a duplicate POSTGRES entry in the TextFieldType enum that triggers TypeError: 'POSTGRES' already defined as 'TEXT' on    
  import, preventing the backend from starting and resulting in 502 errors on the frontend.                                      
                                                                                                                                 
  ### Type of change                                                                                                             
                                                                                                                                 

Rynzie02:
Hey, I have finished it in PR #12931 before

PR #12950, opened 2026-02-02
feat: Add optional document metadata in OpenAI-compatible response references

levischd:

This PR adds an opt‑in way to include document‑level metadata in OpenAI‑compatible reference chunks. Until now, metadata could be used for filtering but wasn’t returned in responses. The change enables clients to show richer citations (author/year/source, etc.) while keeping payload size and privacy under control via an explicit request flag and optional field allowlist.

### Type of change

- [x] Documentation Update

### DIFF
--- a/api/apps/sdk/session.py
+++ b/api/apps/sdk/session.py
@@ -192,6 +192,7 @@ async def chat_completion_openai_like(tenant_id, chat_id):
 
     - If `stream` is True, the final answer and reference information will appear in the **last chunk** of the stream.
     - If `stream` is False, the reference will be included in `choices[0].message.reference`.
+    - If `extra_body.reference_metadata.include` is True, each reference chunk may include `document_metadata` in both streaming and non-streaming responses.
 
     Example usage:
 
@@ -225,6 +226,10 @@ async def chat_completion_openai_like(tenant_id, chat_id):
         stream=stream,
         extra_body={
             "reference": reference,
+            "reference_metadata": {
+                "include": True,
+                "fields": ["author", "year", "source"],
+            },
             "metadata_condition": {
                 "logic": "and",
                 "conditions": [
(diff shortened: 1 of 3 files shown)


levischd:
Example usage screenshot:
Bildschirmfoto 2026-02-02 um 11 53 06

Magicbook1108:
```
 print(f"Reference:\n{chunk.choices[0].delta.reference}")
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/home/bxy/ragflow/.venv/lib/python3.12/site-packages/pydantic/main.py", line 1026, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'ChoiceDelta' object has no attribute 'reference'
```

Unable to get a valid output from the test example.

A screen shot of my test code is provided below
image

levischd:
The error you reported does not match the screenshot.
The screenshot uses extra = _delta_extra(delta) and extra.get("reference"), which avoids the ChoiceDelta attribute error. The reported stack trace only happens when delta.reference is accessed directly.

Please re-run with the exact screenshot code and send the new traceback/output.

Magicbook1108:
@levischd 

1. stream = False
image

2. stream = True

image

3. commit
image


I am using the exact code you provided in session.py. The previous one is my attempt on fixing your issue.

[test.py](https://github.com/user-attachments/files/25043628/test.py)

levischd:
I did not change the structure of the API response, I only expanded it. Please try again without the changes in “session.py.” You won't get a reference field either. This is because of how the OpenAI SDK works. Just run the API request in Postman or a similar program, and you'll see that it works.

Magicbook1108:
Yeah, I’ve tested the API and it works fine. The remaining issue is why it’s not working through the Python SDK. I’ll look into it when I have time and open a separate PR. Once that’s done, we can merge them together.
Alternatively, feel free to investigate it yourself if you prefer.

PR #12956, opened 2026-02-02
Support operator constraints in semi-automatic metadata filtering

HeyPhiS:
#### Summary
This PR enhances the Semi-automatic metadata filtering mode by allowing users to explicitly pre-define operators (e.g., contains, =, >, etc.) for selected metadata keys. While the LLM still dynamically extracts the filter value from the user's query, it is now strictly constrained to use the operator specified in the UI configuration. 

Using this feature is optional. By default the operator selection is set to "automatic" resulting in the LLM choosing the operator (as presently).

#### Rationale & Use Case
This enhancement was driven by a concrete challenge I encountered while working with technical documentation.
In my specific use case, I was trying to filter for software versions within a technical manual. In this dataset, a single document chunk often applies to multiple software versions. These versions are stored as a combined string within the metadata for each chunk.

When using the standard semi-automatic filter, the LLM would inconsistently choose between the contains and equals operators. When it chose equals, it would exclude every chunk that applied to more than one version, even if the version I was searching for was clearly included in that metadata string. This led to incomplete and frustrating retrieval results.

By extending the semi-automatic filter to allow pre-defining the operator for a specific key, I was able to force the use of contains for the version field. This change immediately led to significantly improved and more reliable results in my case. 

I believe this functionality will be equally useful for others dealing with "tagged" or multi-value metadata where the relationship between the query and the field is known, but the specific value needs to remain dynamic.

##### Backend & Core Logic
- `common/metadata_utils.py`: Updated apply_meta_data_filter to support a mixed data structure for semi_auto (handling both legacy string arrays and the new object-based format {"key": "...", "op": "..."}).
- `rag/prompts/generator.py`: Extended gen_meta_filter to accept and pass operator constraints to the LLM.
- `rag/prompts/meta_filter.md`: Updated the system prompt to instruct the LLM to strictly respect provided operator constraints.

##### Frontend
- `web/src/components/metadata-filter/metadata-semi-auto-fields.tsx`: Enhanced the UI to include an operator dropdown for each selected metadata key, utilizing existing operator constants.
- `web/src/components/metadata-filter/index.tsx`: Updated the validation schema to accommodate the new state structure.

#### Test Plan
- Backward Compatibility: Verified that existing semi-auto filters stored as simple strings still function correctly.
- Prompt Verification: Confirmed that constraints are correctly rendered in the LLM system prompt when specified.
- Added unit tests as `test/unit_test/common/test_apply_semi_auto_meta_data_filter.py`
 - Manual End-to-End:
   - Configured a "Semi-automatic" filter for a "Version" key with the "contains" operator.
   - Asked a version-specific query.
   - Result
   
Screenshot 2026-02-02 145359

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


Magicbook1108:
Hello, I am unable to save metada for a document now.
[recording.webm](https://github.com/user-attachments/assets/641ca0b5-9e7d-4671-a0c6-2c1b1e1c05a7)

Update: It looks like normal main branch also unable to create and save metadata now. We are investigating.

PR #12982, opened 2026-02-04
fix: support date comparison operators (>=, <=, >, <) in metadata filtering

Clint-chan:
## Description

  This PR fixes the issue where date metadata conditions with comparison operators (`>=`, `<=`, `>`, `<`) did not work correctly in the `/api/v1/retrieval` endpoint.

  ## Problem

  When using metadata conditions like:
  ```json
  {
    "metadata_condition": {
      "conditions": [
        {
          "name": "date",
          "comparison_operator": ">=",
          "value": "2027-01-13"
        }
      ]
    }
  }

  The filtering did not work as expected because:
  1. Operators >= and <= were not mapped to internal symbols ≥ and ≤
  2. Date strings like "2027-01-13" failed to parse with ast.literal_eval()
  3. Non-standard date formats were incorrectly compared as strings

  Solution

  Changes in common/metadata_utils.py:

  1. Added operator mapping in convert_conditions():
    - >= → ≥
    - <= → ≤
    - != → ≠
  2. Implemented strict date format detection in meta_filter():
    - Only processes dates in YYYY-MM-DD format (10 characters, properly formatted)
    - When query value is a date, only matches data in the same standard format
    - Non-standard formats (e.g., "2026年1月13日", "2026-1-22") are skipped
  3. Maintained backward compatibility:
    - Numeric comparisons still work
    - String comparisons still work
    - Only affects date-formatted queries

  Testing

  All test cases pass (8/8):
  - ✅ Date >= comparison
  - ✅ Date > comparison
  - ✅ Date < comparison
  - ✅ Date <= comparison
  - ✅ Date = comparison
  - ✅ Date range queries
  - ✅ Non-date string comparison (backward compatibility)
  - ✅ Numeric comparison (backward compatibility)

  Example Usage

  {
    "dataset_ids": ["xxx"],
    "question": "test",
    "metadata_condition": {
      "conditions": [
        {
          "name": "date",
          "comparison_operator": ">=",
          "value": "2027-01-13"
        }
      ]
    }
  }

  Notes

  - Only supports standard YYYY-MM-DD format
  - Non-standard date formats in data are treated as data quality issues and will not match
  - Users should ensure their date metadata is in the correct format

### DIFF
--- a/common/metadata_utils.py
+++ b/common/metadata_utils.py
@@ -50,11 +50,11 @@ def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
         for input, docids in v2docs.items():
 
             if operator in ["=", "≠", ">", "<", "≥", "≤"]:
-                # 检查是否为 YYYY-MM-DD 格式的日期
+                # Check if input is in YYYY-MM-DD date format
                 input_str = str(input).strip()
                 value_str = str(value).strip()
 
-                # 严格的日期格式检测：YYYY-MM-DD（必须是10位，且格式正确）
+                # Strict date format detection: YYYY-MM-DD (must be 10 chars with correct format)
                 is_input_date = (
                     len(input_str) == 10 and
                     input_str[4] == '-' and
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!
Please remove Chinese comments.

PR #13004, opened 2026-02-05
Fix: example code in session.py

Magicbook1108:

Fix: example code in session.py #12950

### Type of change

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -187,6 +187,7 @@ jobs:
           echo -e "EXPOSE_MYSQL_PORT=${EXPOSE_MYSQL_PORT}" >> docker/.env
           echo -e "MINIO_PORT=${MINIO_PORT}" >> docker/.env
           echo -e "MINIO_CONSOLE_PORT=${MINIO_CONSOLE_PORT}" >> docker/.env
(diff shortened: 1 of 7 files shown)


docs/references/python_api_reference.md
```diff
@@ -65,43 +65,56 @@ Whether to receive the response as a stream. Set this to `false` explicitly if y
 
 #### Examples
 
+> **Note**
+> Streaming via `client.chat.completions.create(stream=True, ...)` does not
+> return `reference` currently because `reference` is only exposed in the
+> non-stream response payload. The only way to return `reference` is non-stream
+> mode with `with_raw_response`.
+
```
writinwaters:
```suggestion
:::caution NOTE
Streaming via `client.chat.completions.create(stream=True, ...)` does not return `reference` because it is *only* included in the raw response payload in non-stream mode. To return `reference`, set `stream=False`.
:::
```

writinwaters (changes requested):

docs/references/python_api_reference.md
writinwaters: Is this more straightforward?

writinwaters (approved):

PR #13007, opened 2026-02-05
Chore/upgrade dashscope to 1.25.11

Clint-chan:
## Description
  Upgrade dashscope package to support text-embedding-v4 model.

  ## Changes
  - Update dashscope version from 1.20.11 to 1.25.11 in pyproject.toml

  ## Reason
  The text-embedding-v4 model requires dashscope >= 1.25.0 to function properly. This upgrade ensures compatibility with the latest embedding models.


KevinHuSh:
Appreciations!
As long as you update `pyproject.yaml`, `uv.lock` must be submited at the same time.

Issue #13030, opened 2026-02-06
[Bug]: GraphRAG indexing model in dataset parser_config doesn't take effect


hnljq:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

11703d957d1e705de972b67e6d37714b2c4d685a

### RAGFlow image version

v0.22.1-102-gadc78d4ef

### Other environment information

```Markdown

```

### Actual behavior

after uploaded docs into dataset, when configuring a dataset's parsing settings, changing the "Global Index Model" (for GraphRAG) in parser_config.graphrag.llm_id doesn't take effect during document processing. The system always uses the tenant's default chat model instead.

### Expected behavior

Even if the document is uploaded, the modification of the GraphRAG indexing model LLM should take effect in a timely manner

### Steps to reproduce

```Markdown
Steps to Reproduce
Create a dataset
Upload some documents
Go to the dataset settings page
Navigate to Parsing Configuration → GraphRAG settings
Change the "Global Index Model" (indexing model) from the default tenant LLM to a different model (e.g., from gpt-3.5-turbo to gpt-4o)
Start process documents with GraphRAG enabled
Check the task logs or observe which model is actually used
Expected: The selected model (gpt-4o) should be used for GraphRAG graph construction
Actual: The tenant's default chat model (gpt-3.5-turbo) is used instead
```

PR #13031, opened 2026-02-06
Fix: Fixed the issue GraphRAG indexing model in dataset parser_config doesn't take effect. ([#13030](https://github.com/infiniflow/ragflow/issues/13030))

hnljq:
… doesn't take effect.

Fix: Fixed When configuring a dataset's parsing settings, changing the "Global Index Model" (for GraphRAG) in parser_config.graphrag.llm_id doesn't take effect during document processing. The system always uses the tenant's default chat model instead.

### Type of change

### DIFF
--- a/rag/svr/task_executor.py
+++ b/rag/svr/task_executor.py
@@ -960,7 +960,11 @@ async def do_handle_task(task):
     task_tenant_id = task["tenant_id"]
     task_embedding_id = task["embd_id"]
     task_language = task["language"]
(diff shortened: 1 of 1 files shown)


Magicbook1108:
Thanks for your contribution — but please check #13034.

The global llm_id is only used as a fallback/default.
For Graph and Raptor, the effective model is determined by kb_parser_config.llm_id.
For normal chunking, the effective model is determined by parser_config.llm_id.

PR #13044, opened 2026-02-06
feat: add 'Open in new tab' button for agents

PentaFrame-Development:
- Add new button in agent management dropdown to open agent in new tab
- Implement token-based authentication for shared agent access
- Add translations for 9 languages (en, zh, zh-tw, de, fr, it, ru, pt-br, vi)
- Keep existing 'Embed into webpage' functionality intact

This allows users to open agents in a separate tab to work in background while continuing to use other parts of the application.

image
image
image

### Type of change

### DIFF
--- a/docs/guides/agent/embed_agent_into_webpage.md
+++ b/docs/guides/agent/embed_agent_into_webpage.md
@@ -20,6 +20,6 @@ You can use iframe to embed an agent into a third-party webpage.
    - **Enable Streaming Responses**: Enable streaming for widget mode
    - **Locale**: Select the language for the embedded agent
 5. Copy the generated iframe code and embed it into your webpage.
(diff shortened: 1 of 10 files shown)


dcc123456:
@PentaFrame-Development  
Thank you very much for submitting this PR, but this feature button is better placed in the embedded page popup and should directly use the URL from the iframe.
image
image

PentaFrame-Development:
image

Thank You!

ZhenhangTung:
@PentaFrame-Development A small suggestion: if we decide to introduce this change, we might consider renaming “Embed code” to “Copy embed code”, and “Open in new tab” to “Chat in new tab.” Using explicit verbs could make the expected action clearer and reduce ambiguity for users. What do you think?

PR #13052, opened 2026-02-08
Fix async for TypeError in sync_data_source.py (v0.23.1)

ahmadintisar:
This PR applies a runtime hotfix for a bug in the v0.23.1 Docker image where _run_task_logic in sync_data_source.py uses async for to iterate over document_batch_generator, but most connector subclasses (_BlobLikeBase, Notion, Discord, Dropbox, GoogleDrive, Moodle, BOX, Airtable, Asana, Gitlab, etc.) return regular synchronous generators from their _generate() methods. This causes a TypeError: 'async for' requires an object with __aiter__ method, got generator crash for every data source connector except Confluence and Github (which return async generators via async_wrapper()).

Approach: Rather than modifying sync_data_source.py directly (which would create merge conflicts with upstream main where this bug is already fixed), the fix is applied via a sed command in entrypoint.sh that patches the file at container startup before the data sync process launches. The docker-compose.yml volume mount for sync_data_source.py was also removed since it's no longer needed.

Files Changed:

docker/entrypoint.sh — Added a sed one-liner to replace async for with for before starting the data sync process.
docker/docker-compose.yml — Removed the sync_data_source.py volume mount.
Pros:

Zero changes to Python source files, so no merge conflicts when pulling from upstream.
The fix is idempotent — running sed on an already-patched file is a no-op.
Minimal and targeted — only touches the one line that causes the crash.
Cons / Considerations:

This is a workaround, not a permanent fix. It's tied to the v0.23.1 image. Once the image is upgraded to a version that includes the upstream fix, the sed line in entrypoint.sh should be removed to keep the entrypoint clean.
If future versions change the wording of that line, the sed pattern won't match (but it also won't break anything — it'll just be a no-op).


KevinHuSh:
Appreciations!

ahmadintisar:
This is a code issue — sync_data_source.py imports WebDAVConnector but it doesn't exist in the common.data_source module yet (likely a feature added in the code but not in the package's __init__.py).

PR #13076, opened 2026-02-09
refactor(embedding): remove zhipuai sdk and dependency for ZHIPU embeddings

yuehong136:

This PR removes the `zhipuai` Python SDK dependency from the ZHIPU embedding implementation and switches embedding calls to the official HTTP endpoint (`/paas/v4/embeddings`), reducing dependency overhead while keeping behavior unchanged.

### Type of change

- [x] Refactoring

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,7 +117,6 @@ dependencies = [
     "xgboost==1.6.0",
     "xpinyin==0.7.6",
     "yfinance==0.2.65",
-    "zhipuai==2.0.1",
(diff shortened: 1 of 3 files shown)


asiroliu:
@yuehong136 
Thank you very much for your PR. However, I believe moving away from the official Zhipu SDK in favor of a manual implementation is unnecessary and will make the code harder to maintain when the model changes in the future.

KevinHuSh:
Appreciations!

PR #13141, opened 2026-02-13
Ensure last page of documents is included in metadata cache

gileslloyd:

When the metadata cache for documents is being built, we build it page by page in a while loop. We were deciding when to break by using the number of pages we'd fetched so far multiplied by the number of documents per page and comparing to the total to understand how many documents were remaining. The problem was we were incrementing the page count before this calculation, ensuring that we missed the last page of documents.

For example.
I have 288 documents. If I use a page size of 30 then there are 10 pages of documents: `288 / 30 = 9.6`

So once we've fetched 9 pages of documents we'd have 270 documents cached: `9 * 30 = 270` which leaves 18 documents remaining on the last page. But since we increment the page count first the calculation to determine how many documents remain is:
`288 - (10*30) = -12` instead of `288 - (9*30) = 18`. The result is the last 18 documents are missed. 

We could just move the page increment after this calculation but I feel like this solution is over-complicated. The cleaner thing to do is just break on the first page with no documents.

### Type of change

### DIFF
--- a/mcp/server/server.py
+++ b/mcp/server/server.py
@@ -271,11 +271,14 @@ class RAGFlowConnector:
                     doc_id_meta_list = []
                     docs = {}
                     while page:
-                        docs_res = await self._get(f"/datasets/{dataset_id}/documents?page={page}", api_key=api_key)
-                        if not docs_res:
-                            break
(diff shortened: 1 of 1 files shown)


KevinHuSh:
Appreciations!
It has issues, please test it.

PR #13147, opened 2026-02-13
Add Bulgarian language support

trifonnt:

RAGFlow supports 12 UI languages but does not include Bulgarian. This PR adds Bulgarian (`bg` / `Български`) as the 13th supported language, covering the full UI translation (2001 keys across all 26 sections) and OCR/PDF parser language mapping.

### Type of change

### Changes

- **`web/src/constants/common.ts`** — Registered Bulgarian in all 5 language data structures (`LanguageList`, `LanguageMap`, `LanguageAbbreviation` enum, `LanguageAbbreviationMap`, `LanguageTranslationMap`)
- **`web/src/locales/config.ts`** — Added lazy-loading dynamic import for the `bg` locale
- **`web/src/locales/bg.ts`** *(new)* — Full Bulgarian translation file with all 26 sections, matching the English source (`en.ts`). All interpolation placeholders, HTML tags, and technical terms are preserved as-is
- **`deepdoc/parser/mineru_parser.py`** — Mapped `'Bulgarian'` to `'cyrillic'` in `LANGUAGE_TO_MINERU_MAP` for OCR/PDF parser support

### How it works

The language selector automatically picks up the new entry. When a user selects "Български", the translation bundle is lazy-loaded on demand. The preference is persisted to the database and localStorage across sessions.

### DIFF
--- a/web/src/locales/de.ts
+++ b/web/src/locales/de.ts
@@ -28,6 +28,7 @@ export default {
       chinese: 'Vereinfachtes Chinesisch',
       traditionalChinese: 'Traditionelles Chinesisch',
       russian: 'Russisch',
+      bulgarian: 'Bulgarisch',
       german: 'Deutsch',
       language: 'Sprache',
       languageMessage: 'Bitte geben Sie Ihre Sprache ein!',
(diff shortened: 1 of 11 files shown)


dcc123456:
@trifonnt There are still some minor issues that need to be addressed.
image

trifonnt:
> @trifonnt There are still some minor issues that need to be addressed. image

@dcc123456 Good catch, thank you!
I have fixed.

PR #13151, opened 2026-02-15
feat: Add `disable_password_login` configuration to support SSO-only authentication

ahmadintisar:

Enterprise deployments that use an external Identity Provider (e.g., Microsoft Entra ID, Okta, Keycloak) need the ability to enforce SSO-only authentication by hiding the email/password login form. Currently, the login page always shows the password form alongside OAuth buttons, with no way to disable it.

This PR adds a `disable_password_login` configuration option under the existing `authentication` section in `service_conf.yaml`. When set to `true`, the login page only displays configured OAuth/SSO buttons and hides the email/password form, "Remember me" checkbox, and "Sign up" link.

The flag can be set via:
- `service_conf.yaml` (`authentication.disable_password_login: true`)
- Environment variable (`DISABLE_PASSWORD_LOGIN=true`)

Default behavior is unchanged (`false`).

### Behavior

| `disable_password_login` | OAuth configured | Result |
|---|---|---|
| `false` (default) | No | Standard email/password form |
| `false` | Yes | Email/password form + SSO buttons below |
| `true` | Yes | **SSO buttons only** (no form, no sign up link) |
| `true` | No | Empty card (admin should configure OAuth first) |

### Type of change

### Files changed (5)

1. `docker/service_conf.yaml.template` — added `disable_password_login: false` under authentication
2. `common/settings.py` — added `DISABLE_PASSWORD_LOGIN` global variable and loader in `init_settings()`
3. `common/config_utils.py` — fixed `TypeError` in `show_configs()` when authentication section contains non-dict values (e.g., booleans)
4. `api/apps/system_app.py` — exposed `disablePasswordLogin` flag in `/config` endpoint
5. `web/src/pages/login/index.tsx` — conditionally render password form based on config flag; OAuth buttons always render when channels exist

### DIFF
--- a/api/apps/sdk/doc.py
+++ b/api/apps/sdk/doc.py
@@ -809,6 +809,10 @@ async def delete(tenant_id, dataset_id):
     return get_result()
 
 
+DOC_STOP_PARSING_INVALID_STATE_MESSAGE = "Can't stop parsing document that has not started or already completed"
+DOC_STOP_PARSING_INVALID_STATE_ERROR_CODE = "DOC_STOP_PARSING_INVALID_STATE"
+
+
 @manager.route("/datasets/<dataset_id>/chunks", methods=["POST"])  # noqa: F821
 @token_required
 async def parse(tenant_id, dataset_id):
@@ -947,7 +951,11 @@ async def stop_parsing(tenant_id, dataset_id):
         if not doc:
             return get_error_data_result(message=f"You don't own the document {id}.")
         if doc[0].run != TaskStatus.RUNNING.value :
-            return get_error_data_result("Can't stop parsing document that has not started or already completed")
+            return construct_json_result(
+                code=RetCode.DATA_ERROR,
+                message=DOC_STOP_PARSING_INVALID_STATE_MESSAGE,
+                data={"error_code": DOC_STOP_PARSING_INVALID_STATE_ERROR_CODE},
+            )
         # Send cancellation signal via Redis to stop background task
         cancel_all_task_of(id)
         info = {"run": "2", "progress": 0, "chunk_num": 0}
--- a/test/testcases/test_http_api/test_file_management_within_dataset/test_doc_sdk_routes_unit.py
+++ b/test/testcases/test_http_api/test_file_management_within_dataset/test_doc_sdk_routes_unit.py
@@ -592,7 +592,12 @@ class TestDocRoutesUnit:
         res = _run(module.parse.__wrapped__("tenant-1", "ds-1"))
         assert "don't own the document" in res["message"]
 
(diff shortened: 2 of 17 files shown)


ahmadintisar:
Hi Team, 

It's been a couple of days and my PR hasn't been reviewed. I also don't see an option to add a reviewer manually.
Your response would be appreciated.

ahmadintisar:
@yingfeng @cunba-ai @superwiboy

docker/service_conf.yaml.template
```diff
@@ -138,13 +138,14 @@ user_default_llm:
...
 #     client_secret: "your_client_secret"
 #     redirect_uri: "https://your-app.com/v1/user/oauth/callback/github"
-# authentication:
-#   client:
-#     switch: false
-#     http_app_key:
-#     http_secret_key:
-#   site:
-#     switch: false
+authentication:
```
KevinHuSh: Appreciations!

This should not be uncomment in, which will change defautl behaviour of the system.

ahmadintisar: @KevinHuSh Thanks for the feedback Kevin. I'll update the PR to use the flag from .env instead of authentication. Sounds good?

ahmadintisar: @KevinHuSh  I have resolved your comment.

ahmadintisar:
@KevinHuSh  can you please review? I see the latest test case failed, it says in common.settings Disable form is not declared, but it is declared. Can you review pls :)

yingfeng:
> @KevinHuSh can you please review? I see the latest test case failed, it says in common.settings Disable form is not declared, but it is declared. Can you review pls :)

It does cause the CI failure, please fix it, thanks~~

ahmadintisar:
> > @KevinHuSh can you please review? I see the latest test case failed, it says in common.settings Disable form is not declared, but it is declared. Can you review pls :)
> 
> It does cause the CI failure, please fix it, thanks~~

## CI Update

The original failure (`test_get_config_returns_register_enabled_unit`) is fixed — the test mock for `common.settings` was missing the new `DISABLE_PASSWORD_LOGIN` attribute. Added it to the mock setup and it now passes.

The 2 remaining failures in CI are **unrelated to this PR**:

1. **`test_parse_branches`** — `DocumentService.update_by_id` isn't fully mocked, so it attempts a real MySQL connection that doesn't exist in CI. Needs a DB mock added to the test.

2. **`test_stop_parsing_branches`** — Assertion expects the old message `"progress at 0 or 1"` but the source now returns `"Can't stop parsing document that has not started or already completed"`. The test string just needs updating.

Both of these appear to be pre-existing issues from other merged changes, not from this PR.

PR #13160, opened 2026-02-17
Fix: setting keywords_similarity_weight through HTTP API

as-ondewo:

When setting the `keywords_similarity_weight` of a chat assistant through the HTTP API (both *create* and *update*) the value was inverted. In the database the `vector_similarity_weight` is stored which should be `vector_similarity_weight = 1 - keywords_similarity_weight` but it was stored as `vector_similarity_weight = keywords_similarity_weight`. This meant that when, for example setting `keywords_similarity_weight = 0.8`, the request would actually set `vector_similarity_weight = 0.8` and `keywords_similarity_weight = 0.2`.

There are tests that should have caught this, but they were incorrectly expecting the `keywords_similarity_weight` to be inverted.

In this PR I fixed the implementation of the create and update chat assistant HTTP API endpoints and the associated test cases.

### Type of change


api/apps/sdk/chat.py
```diff
@@ -67,6 +67,8 @@ async def create(tenant_id):
         for new_key, old_key in key_mapping.items():
             if old_key in prompt:
                 prompt[new_key] = prompt.pop(old_key)
+        if "vector_similarity_weight" in prompt:
```
KevinHuSh, line 70: Is it `keywords_similarity_weight`?

as-ondewo: It's a bit confusing to read, but in the loop above the keys are remapped, so at this point the key is `"vector_similarity_weight"`

api/apps/sdk/chat.py
```diff
@@ -0,0 +1,331 @@
...
+    # prompt
+    prompt = req.get("prompt")
+    key_mapping = {"parameters": "variables", "prologue": "opener", "quote": "show_quote", "system": "prompt", "rerank_id": "rerank_model", "vector_similarity_weight": "keywords_similarity_weight"}
+    key_list = ["similarity_threshold", "vector_similarity_weight", "top_n", "rerank_id", "top_k"]
+    if prompt:
+        for new_key, old_key in key_mapping.items():
+            if old_key in prompt:
+                prompt[new_key] = prompt.pop(old_key)
+        if "vector_similarity_weight" in prompt:
+            prompt["vector_similarity_weight"] = 1 - prompt["vector_similarity_weight"]
```
KevinHuSh, line 71: I mean that:
`prompt["vector_similarity_weight"] = 1 - prompt["keywords_similarity_weight"]`

JinHai-CN:
please re-open the PR, if you are still working on this PR.

PR #13161, opened 2026-02-17
fix: datasets cleared when updating chat assistant

as-ondewo:
#### Bug

When updating a chat assistant through the HTTP API endpoint, RAGFlow always set the `dataset_ids` to `[]`, unless they were explicitly provided in the request. The test cases did not catch this bug because they only used a chat assistant without datasets and did not check that values that were not included in the update request remain unchanged.

#### Fix

I changed the update chat assistant HTTP API endpoint to not set `dataset_ids` to `[]` if the request does not contain `dataset_ids` and updated the update chat assistant tests to check that the updated value was set correctly and all other values remain unchanged.

### Type of change


KevinHuSh:
Appreciations!
I don't quite follow. `req["kb_ids"] = []` will lead to ...?

as-ondewo:
If `req["kb_ids"] = []` is in the `req` dict passed to `DialogService.update_by_id`, then it will set `kb_ids` to an empty list. This should only happen when it is explicitly set in the request body (`{"dataset_ids": []}`) but not if `"dataset_ids"` is not present in the request body.

Without my fix, if you have a Chat Assistant with some datasets and you, for example call update with just `{"name": "updated_name"}` it would also remove the datasets from the assistant because the update function always set them to `[]`, even if they were not in the request.

PR #13166, opened 2026-02-18
Fix: empty response from OpenAI chat completion endpoint

as-ondewo:

When using a chat assistant that has a hardcoded `empty_response`, that response was not returned correctly in streaming mode when no information is found in the knowledge base. In this case only one response with `"content": null` was yielded. If `"references": true`, then the `empty_response` is still put into the `final_content` so there is technically some content returned, but when `"references": false` no content at all is returned.

I update the OpenAI chat completion endpoint to yield an additional response with the `empty_response` in the content.

### Type of change


yongtenglei (approved):
Thank you very much, this looks good to me :)

Cheers!

Issue #10321, opened 2025-09-26
[Feature Request]: Please support Elastic API Key option


metalshanked:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

Currently only username and password are allowed for elastic connection.
please allow using apikey option as well

### Documentation, adoption, use case

```Markdown

```

ZhenhangTung:
Would you love to implement this feature @metalshanked ? 

cc @Magicbook1108

ZhenhangTung:
btw, @KevinHuSh could you explain why only username and password are supported?

leonardlin:
@KevinHuSh I added a PR, could you check?
What is the process for community PRs?

leonardlin:
Sharing Knowledge around cloud hosted Elastic Search:

based on ragflow v0.24

elastic: serverless : does not work. features around index and cluster Infos are not available on serverless

elastic cloud hosted: only 8.x version works. Elastic v9 seems to work but vector retrieval will always score 0.0

recommendation:
use elastic hosted V8.x

metalshanked:
@leonardlin  can you share details around using SELF HOSTED elastic v9....how can we confirm this if score is 0.0


here is what it is showing

### Elasticsearch v9 Compatibility Fixes for Self-Hosted Vector Retrieval

#### Problem
The RAGFlow Elasticsearch integration was pinned to `elasticsearch-dsl==8.12.0` and used several deprecated/removed API patterns that would break with Elasticsearch v9 servers. The vector retrieval (dense_vector/knn search) was correctly configured in the mapping but the Python client code had compatibility issues.

#### Changes Made

**`pyproject.toml`** — Relaxed elasticsearch-dsl version constraint:
- Changed `"elasticsearch-dsl==8.12.0"` → `"elasticsearch-dsl>=8.12.0,<9.0.0"` to allow newer 8.x versions with better ES v9 compatibility handling.

[...]

PR #13167, opened 2026-02-18
feat: Elasticsearch access via API Key

leonardlin:

As requested: https://github.com/infiniflow/ragflow/issues/10321

Elasticsearch configuration uses username/password.
In order to use external / cloud hosted elasticsearch, we need to enable API-Key based authentication.

Following changes:
    Adds ELASTIC_API_KEY as configuration
    Connection Pool checks if APIKey is present and tries to use it
    HelmChart allows to disable the elasticsearch deployment
### Type of change


leonardlin:
@Magicbook1108 @KevinHuSh this is the updated PR

kowkowhuang:
Exactly what's the point of Basic Auth when the Elastic Cloud provided default API is based on endpoint + API Key? Shockingly bad decision to build something with basic auth. 

@click.option("--es-user", default=None, help="Elasticsearch username")
@click.option("--es-password", default=None, help="Elasticsearch password")

Can I plase also request you do NOT default the endpoint of ES cloud to 9200 pls?  The ES cloud's default provided port is 443 not 9200

@click.option("--es-port", default=9200, type=int, help="Elasticsearch port")

docker/.env
```diff
@@ -31,7 +31,7 @@ COMPOSE_PROFILES=${DOC_ENGINE},${DEVICE}
 STACK_VERSION=8.11.3
 
 # The hostname where the Elasticsearch service is exposed
-ES_HOST=es01
+ES_HOST=http://es01:9200
```
KevinHuSh: Appreciations!
This default setting should not be changed in consideration other users.

leonardlin: Hi @KevinHuSh 

There was no actual change in default port or protocol. It was http/9200 before too.

The difference is:

Previously the default protocol (http) and default port (9200) was populated/enforced in the docker/service_conf.yaml.template.
There was no way to define httpS or another port via environment variables.

ES_HOST as environment variable is not used directly but only the after parsing in service_conf.yaml.template

This means for the cases:
No ES_HOST defined: In service_conf.yaml.template the old default is defined: http://es01:9200

ES_HOST defined as old default: http://es01:9200 -> passed through

ES_HOST set to another external service -> passed through

Does this help to clarify?

KevinHuSh: The circumstance might be more complicated. By my understanding, this requires `.env` and `service_conf.yaml` should be with the same version since they're coupled. For some users, they might apply the latest .env with the older version of `service_conf.yaml`. So, we should try our best to make minimum alteration to avoid unecessary confusion for all users.

leonardlin: @KevinHuSh I updated the approach to keep the old ES_HOST format.

leonardlin:
@KevinHuSh anything I can do to progress this PR?

PR #13178, opened 2026-02-22
fix: enforce embedding model max token limit via proper truncation

### LINKED ISSUE #4683: [Bug]: During local embedding，RAGFlow is sending too much text at once, exceeding the model's maximum token limit, causing the model to be unable to fully read the input.

main

### RAGFlow image version

v0.15.1,nightly

```Markdown

```

### Actual behavior

When employing embedding models with a lower 'maximum input token' capacity，models such as bge-large and conan-embedding-v1 are limited to a maximum input of 512 tokens. When using these models for embedding, RAGFlow sends more than 512 tokens at once, ollama will encounter an error.
I've found the cause of the error here：[https://github.com/ollama/ollama/issues/7288#issuecomment-2591709109](url)
Although I can adjust the maximum input limit of the model in ollama, it will cause RAGFlow's text to be truncated, resulting in incomplete embeddings.Additionally, I'm unable to locate a setting within RAGFlow to control the maximum input for the embedding model.

When adding a model, the max token setting controls the maximum output, not the input, which doesn't apply to embedding models.
Image

The same issue of an ineffective max token option also exists when adding remarker models.

Image

### Expected behavior

Please add a setting to RAGFlow to control the maximum number of tokens sent to the embedding model per request, and also fix the bug where the max token limit is ineffective when adding remarker models.

### Steps to reproduce

```Markdown
Using the bge-large:latest model in ollama, if the embedding is performed with a method other than 'general' (I am using 'book'), and the token count goes over 512, an error occurs and the embedding is terminated.
```

### Additional information

_No response_

timon0305:
## Summary

Fixes the embedding model max token limit not being enforced, causing API errors when input text exceeds the model's context window. This was reported across multiple providers (Ollama, vLLM, HuggingFace TEI) with models like bge-large (512 tokens), multilingual-e5-large-instruct (512 tokens), etc.

### Root causes fixed

1. **`LLMBundle.encode()` truncated by character count instead of token count** — `text[:target_len]` slices by character index, but `max_length` is in tokens. For multi-byte text (Chinese/Japanese), this still exceeds the token limit. Fixed to use `truncate()` which properly tokenizes and truncates at the token level.

2. **`LLMBundle.encode_queries()` had no truncation at all** — User search queries passed directly to embedding API without any length check. Added token-level truncation using the configured `max_length`.

3. **12 embedding provider classes lacked input truncation** — Added `truncate()` calls to both `encode()` and `encode_queries()` for: `OllamaEmbed`, `XinferenceEmbed`, `LocalAIEmbed`, `HuggingFaceEmbed`, `CoHereEmbed`, `SILICONFLOWEmbed`, `ReplicateEmbed`, `BaiduYiyanEmbed`, `VoyageEmbed`, `VolcEngineEmbed`, `ZhipuEmbed` (encode_queries), `QWenEmbed` (encode_queries character slice → token truncate).

4. **Misleading tooltip for max_tokens setting** — The UI tooltip said "Defaults to 512" but the actual default is 8192. Updated en/zh locales to accurately describe the field as "maximum input context length" with correct default.

### Changes
- **`api/db/services/llm_service.py`** — Fixed `LLMBundle.encode()` to use token-level `truncate()` instead of character slicing; added truncation to `LLMBundle.encode_queries()`
- **`rag/llm/embedding_model.py`** — Added `truncate()` to 12 provider classes missing input truncation
- **`web/src/locales/en.ts`** — Fixed maxTokensTip tooltip text and default value
- **`web/src/locales/zh.ts`** — Fixed maxTokensTip tooltip text and default value (Chinese)

Closes #4683


JinHai-CN:
@timon0305 I will close this PR, if you are still working on this PR, please re-open it.

PR #13180, opened 2026-02-23
fix(mineru): merge cross-page paragraph continuations

yuch85:
## Problem

MinerU's `_transfer_to_sections()` in `deepdoc/parser/mineru_parser.py` treats each page's content blocks independently. When a paragraph spans two pages, it becomes two separate sections that may end up in different chunks during the chunking phase. This hurts retrieval quality for long documents where paragraphs frequently cross page boundaries.

## Root Cause

The `_transfer_to_sections()` method processes MinerU's output page-by-page, creating a new section for each content block. Unlike the DeepDoc parent class which has `_line_tag()` cross-page detection using coordinate overlap, MinerU uses a normalized 0-1000 coordinate system that resets per page, making coordinate-based page-span detection impossible.

## Fix

Added three module-level helper functions and integrated them into `_transfer_to_sections()`:

1. **`_is_continuation(prev_text, next_text, prev_tag, next_tag)`** — Detects cross-page paragraph continuations using text-level heuristics:
   - Previous section must NOT end with terminal punctuation (`.?!:;`)
   - Tags must be from adjacent pages
   - Next section must not start with a paragraph/section number (e.g., `123. The`)

2. **`_merge_tags(tag1, tag2)`** — Combines two position tags into a multi-page format with union bounding boxes (`@@207\t...##` + `@@208\t...##` → `@@207-208\t...##`)

3. **`_merge_cross_page_sections(sections, parse_method)`** — Linear scan that merges adjacent continuation pairs, handling all three tuple formats (`default`, `manual`, `paper`)

Called at the end of `_transfer_to_sections()` before returning.

## Comparison with DeepDoc

The DeepDoc parser's `_line_tag()` method (in `pdf_parser.py`) detects cross-page spans by checking if a text line's Y-coordinate overlaps with the previous page's bottom margin. This works because DeepDoc uses actual PDF coordinates. MinerU's normalized 0-1000 coordinate system resets per page, making this approach impossible to port. The text-level heuristic approach is more robust for MinerU's output format.

## Impact

In our testing with 211 legal PDFs (100-900 pages each), approximately 5-8% of chunks contained cross-page paragraph splits. After this fix, those paragraphs are merged into single sections before chunking, improving retrieval coherence.

## Testing

- Tested with ICTY tribunal documents (scanned PDFs, 100-500 pages)
- Verified paragraph continuations are correctly detected and merged
- Verified numbered paragraphs are NOT incorrectly merged
- No impact on single-page sections


yongtenglei (approved):
Hi, @yuch85 

There are some issues here, could you please take a look again? Thank you. It concatenates the whole next page to the previous one.

image

yongtenglei (changes requested):
[The-Necklace.pdf](https://github.com/user-attachments/files/25681038/The-Necklace.pdf)

This is the sample file.

yuch85:
Thank you, I am now reconsidering the fundamental approach to this problem. I will get back on this.

yuch85:
Hi @yongtenglei, could you re-upload the test file ("The-Necklace.pdf"), I was unable to download it.

yuch85:
[The-Necklace.pdf](https://github.com/user-attachments/files/25733374/The-Necklace.pdf)
@yongtenglei, thanks again for catching this, I have changed the approach fundamentally.

**What was wrong:** The heuristic-based section merging (`_is_continuation`, `_merge_tags`, `_merge_cross_page_sections`) tried to detect paragraph continuations using text-level heuristics. In dialogue-heavy text like "The Necklace", closing quotes (`."` `!"`) hid the terminal punctuation from the check, causing cascading false merges that concatenated entire pages.

**New approach (commit fe919e6):** Instead of adding heuristics to `_transfer_to_sections()`, I removed all the merge code and made a one-line change in `rag/app/naive.py`: remove `"mineru"` from the `chunk_token_num = 0` override.

This lets MinerU sections flow through the existing `naive_merge` chunker, which merges sections up to the user's configured token limit (default 128). Cross-page paragraphs naturally end up in the same chunk. The position tag system (`@@page\tcoords##`) already handles multi-page chunks correctly — each section keeps its own tag, and `extract_positions` finds all of them via regex.

**Test results with The Necklace** (I was unable to open your PDF, so I tested with my own version — attached):
- Before (`chunk_token_num=0`): 160 single-page chunks
- After: 27 chunks, 6 of which correctly span two pages with accurate multi-page position metadata

No heuristics, no new helper functions — just stopped bypassing the chunker that already handles this. For your further review please.

JinHai-CN:
@yuch85 it is already fixed now. Please re-open this PR, if I am wrong.

PR #13197, opened 2026-02-24
fix(api): MinIO health check use dynamic scheme and verify (Closes #13159 and #13158)

### LINKED ISSUE #13158: [Bug] MinIO client connection ignores 'secure' configuration and hardcodes SSL to False
Based on the discussion provided, here are the two GitHub issues formatted for the RAGFlow repository.

latest

**RAGFlow image version**
0.23.1

Docker

**Actual behavior**
The MinIO client connection in RAGFlow is currently hardcoded to `secure=False`, making it impossible to connect to MinIO instances requiring SSL/TLS, even if the `secure` parameter is set to `True` in `service_conf.yaml`.

In `rag/utils/minio_conn.py`:
```python
def __open__(self):
    self.conn = Minio(settings.MINIO["host"],
                      access_key=settings.MINIO["user"],
                      secret_key=settings.MINIO["password"],
                      secure=False  # This is hardcoded to False
                      )
```

**Expected behavior**
The MinIO client should respect the `secure` setting from the configuration. Additionally, it should provide a way to handle self-signed certificates (e.g., via an environment variable to disable certificate verification or provide a CA bundle).

**Steps to reproduce**
1. Set up a MinIO instance with SSL enabled (HTTPS).
2. Configure `service_conf.yaml` with `secure: True`.
3. Attempt to start RAGFlow or upload a document.
4. The connection will fail or attempt to connect via plain HTTP.

**Suggested Fix**
Modify `rag/utils/minio_conn.py` to read the configuration and handle SSL contexts:
```python
    secure = settings.MINIO.get("secure", False)
    # Logic to handle SSL and optionally disable verification for self-signed certs
```

---

### LINKED ISSUE #13159: [Bug] MinIO health check hardcodes HTTP protocol causing "timeout" status when SSL is enabled
latest

**RAGFlow image version**
0.23.1

Docker

**Actual behavior**
Even if the MinIO client connection is patched to support SSL, the System Health dashboard shows a "timeout" for the MinIO service. This is because the health check utility hardcodes the `http://` protocol in the connection string.

In `api/utils/health_utils.py`:
```python
def check_minio_alive():
    try:
        # Protocol is hardcoded as http
        response = requests.get(f'http://{settings.MINIO["host"]}/minio/health/live') 
        # ...
```

**Expected behavior**
The health check should dynamically use `http` or `https` based on the `MINIO_SECURE` configuration setting. It should also account for self-signed certificates to prevent SSL handshake errors from reporting a false "timeout".

**Steps to reproduce**
1. Enable SSL on the MinIO server.
2. Update RAGFlow configuration to use the HTTPS port/secure setting.
3. Check the Admin/Service Health dashboard.
4. MinIO status will show "timeout" despite the service being functional.

**Suggested Fix**
Modify `api/utils/health_utils.py`:
```python
def check_minio_alive():
    scheme = 'https' if settings.MINIO.get("secure", False) else 'http'
    # Use the dynamic scheme and consider verify=False for self-signed setups
    response = requests.get(f'{scheme}://{settings.MINIO["host"]}/minio/health/live', verify=False)
```

Eruis2579:
## Summary

Fixes MinIO SSL/TLS support in two places: the MinIO **client** connection and the **health check** used by the Admin/Service Health dashboard. Both now respect the `secure` and `verify` settings from the MinIO configuration.

Closes #13158
Closes #13159

---

## Problem

**#13158 – MinIO client:** The client in `rag/utils/minio_conn.py` was hardcoded with `secure=False`, so RAGFlow could not connect to MinIO over HTTPS even when `secure: true` was set in config. There was also no way to disable certificate verification for self-signed certs.

**#13159 – MinIO health check:** In `api/utils/health_utils.py`, the MinIO liveness check always used `http://` for the health URL. When MinIO was configured with SSL, the health check failed and the dashboard showed "timeout" even though MinIO was reachable over HTTPS.

---

### MinIO client (`rag/utils/minio_conn.py`)

- Read `MINIO.secure` (default `false`) and pass it into the `Minio()` constructor so HTTPS is used when configured.
- Add `_build_minio_http_client()` that reads `MINIO.verify` (default `true`). When `verify` is false, return an `urllib3.PoolManager` with `cert_reqs=ssl.CERT_NONE` and pass it as `http_client` to `Minio()` so self-signed certificates are accepted.
- Support string values for `secure` and `verify` (e.g. `"true"`, `"false"`).

### MinIO health check (`api/utils/health_utils.py`)

- Add `_minio_scheme_and_verify()` to derive URL scheme (http/https) and the `verify` flag from `MINIO.secure` and `MINIO.verify`.
- Update `check_minio_alive()` to use the correct scheme, pass `verify` into `requests.get(..., verify=verify)`, and use `timeout=10`.

### Config template (`docker/service_conf.yaml.template`)

- Add commented optional MinIO keys `secure` and `verify` (and env vars `MINIO_SECURE`, `MINIO_VERIFY`) so deployers know they can enable HTTPS and optional cert verification.

### Tests

- **`test/unit_test/utils/test_health_utils_minio.py`** – Tests for `_minio_scheme_and_verify()` and `check_minio_alive()` (scheme, verify, status codes, timeout, errors).
- **`test/unit_test/utils/test_minio_conn_ssl.py`** – Tests for `_build_minio_http_client()` (verify true/false/missing, string values, `CERT_NONE` when verify is false).

---

## Testing

- Unit tests added/updated as above; run with the project's test runner.
- Manually: configure MinIO with HTTPS and `secure: true` (and optionally `verify: false` for self-signed); confirm client operations work and the Service Health dashboard shows MinIO as alive instead of timeout.

### DIFF
--- a/test/unit_test/utils/test_health_utils_minio.py
+++ b/test/unit_test/utils/test_health_utils_minio.py
@@ -17,7 +17,6 @@
 Unit tests for MinIO health check (check_minio_alive) and scheme/verify helpers.
 Covers SSL/HTTPS and certificate verification (issues #13158, #13159).
 """
-import pytest
 from unittest.mock import patch, Mock
 
 
--- a/test/unit_test/utils/test_minio_conn_ssl.py
+++ b/test/unit_test/utils/test_minio_conn_ssl.py
@@ -20,8 +20,6 @@ Covers issue #13158.
 import ssl
 from unittest.mock import patch
 
-import pytest
-
 
 class TestBuildMinioHttpClient:
     """Test _build_minio_http_client helper."""


Eruis2579:
Hi, @KevinHuSh Could you review my PR?

Eruis2579:
Hi, @KevinHuSh Could you Merge my PR?

PR #13233, opened 2026-02-26
refactor(word): lazy-load DOCX images to reduce peak memory without changing output

eviaaaaa:
**Summary**
This PR tackles a significant memory bottleneck when processing image-heavy Word documents. Previously, our pipeline eagerly decoded DOCX images into `PIL.Image` objects, which caused high peak memory usage. To solve this, I've introduced a **lazy-loading approach**: images are now stored as raw blobs and only decoded exactly when and where they are consumed.

This successfully reduces the memory footprint while keeping the parsing output completely identical to before.

**What's Changed**
Instead of a dry file-by-file list, here is the logical breakdown of the updates:

* **The Core Abstraction (`lazy_image.py`)**: Introduced `LazyDocxImage` along with helper APIs to handle lazy decoding, image-type checks, and NumPy compatibility. It also supports `.close()` and detached PIL access to ensure safe lifecycle management and prevent memory leaks.
* **Pipeline Integration (`naive.py`, `figure_parser.py`, etc.)**: Updated the general DOCX picture extraction to return these new lazy images. Downstream consumers (like the figure/VLM flow and base64 encoding paths) now decode images right at the use site using detached PIL instances, avoiding shared-instance side effects.
* **Compatibility Hooks (`operators.py`, `book.py`, etc.)**: Added necessary compatibility conversions so these lazy images flow smoothly through existing merging, filtering, and presentation steps without breaking.

**Scope & What is Intentionally Left Out**
To keep this PR focused, I have restricted these changes strictly to the **general Word pipeline** and its downstream consumers.
The `QA` and `manual` Word parsing pipelines are explicitly **not modified** in this PR. They can be safely migrated to this new lazy-load model in a subsequent, standalone PR.

**Design Considerations**
I briefly considered adding image compression during processing, but decided against it to avoid any potential quality degradation in the derived outputs. I also held off on a massive pipeline re-architecture to avoid overly invasive changes right now.

**Validation & Testing**
I've tested this to ensure no regressions:

* Compared identical DOCX inputs before and after this branch: chunk counts, extracted text, table HTML, and image descriptions match perfectly.
* **Confirmed a noticeable drop in peak memory usage when processing image-dense documents.** For a 30MB Word document containing 243 1080p screenshots, memory consumption is reduced by approximately 1.5GB.

**Breaking Changes**
None.

### DIFF
--- a/deepdoc/parser/figure_parser.py
+++ b/deepdoc/parser/figure_parser.py
@@ -24,19 +24,24 @@ from common.connection_utils import timeout
 from rag.app.picture import vision_llm_chunk as picture_vision_llm_chunk
 from rag.prompts.generator import vision_llm_figure_describe_prompt, vision_llm_figure_describe_prompt_with_context
 from rag.nlp import append_context2table_image4pdf
+from rag.utils.lazy_image import ensure_pil_image, open_image_for_processing, is_image_like
 
 # need to delete before pr
 def vision_figure_parser_figure_data_wrapper(figures_data_without_positions):
     if not figures_data_without_positions:
         return []
-    return [
-        (
-            (figure_data[1], [figure_data[0]]),
-            [(0, 0, 0, 0, 0)],
+    res = []
+    for figure_data in figures_data_without_positions:
+        img = ensure_pil_image(figure_data[1])
+        if not isinstance(img, Image.Image):
+            continue
+        res.append(
+            (
+                (img, [figure_data[0]]),
+                [(0, 0, 0, 0, 0)],
+            )
         )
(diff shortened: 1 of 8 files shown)


eviaaaaa:
> Hi, @eviaaaaa
> 
> Thank you for this contribution. This change looks good to me. You are more than welcome if you could extend the same lazy loading to QA and manual you mentioned in `Scope & What is Intentionally Left Out` in a follow-up PR.
> 
> Cheers.

Thanks for the review! I definitely plan to do the follow-up PR for QA and manual. However, I might be a bit busy this week, so I'll probably submit it in a few days.

PR #13247, opened 2026-02-26
fix(auth): normalize Authorization header to handle Bearer/Token prefix

levindecaro:
Normalize Authorization header to handle Bearer/Token prefix.

Strip scheme prefixes (Bearer, Token, JWT) from Authorization header before signature verification in _load_user(). This fixes OAuth callback failures where the header includes a scheme prefix, causing signature mismatch errors like "Signature ... does not match".

### Type of change


Lynn-Inf (commented):
Thanks for this PR! I agree that making the authentication more standardized is a great improvement.

Here's a bit of context to help clarify the current logic:

- Lines 103–122 handle JWT token authentication.

- Lines 124–136 handle authentication using the RAGflow API token (via Bearer scheme).

When a two-part token is parsed and its specific type is identified, we can directly route it to the corresponding authentication flow. Additionally, we've retained backward compatibility by keeping the original Authorization: <jwt_token> header format (without the Bearer prefix) as a fallback.

Let me know if you have any questions or further suggestions!

KevinHuSh:
Appreciations!
Out of date.

Issue #9921, opened 2025-09-04
[Proposal] Evolving From Semantic Engine to a Deterministic Document Access Layer (DAL) for Advanced Agents


IndenScale:
## 1. Abstract

  This proposal outlines a strategic architectural evolution for RAGFlow, designed to elevate it from a premier Retrieval-Augmented Generation (RAG)
  engine into a foundational Document Access Layer (DAL) for sophisticated, agentic AI systems. While RAGFlow's semantic search capabilities are
  powerful, they are fundamentally limited by their probabilistic nature. Advanced agents require deterministic, predictable, and controllable access
   to document content. We propose a hybrid architecture that retains RAGFlow's powerful semantic search for initial discovery while introducing a
  robust, deterministic access layer for precise operations. This involves three core pillars: Canonical Document Representation, a Multi-level 
  Indexing System, and a Cognitive Anchor Network.

 ## 2. Problem Statement: The Limitations of a Purely Semantic Paradigm for Agentic Workflows

  Modern AI Agents are moving beyond simple Q&A. They function as autonomous systems that need to perceive, reason, and act upon data with
  precision. In this context, a purely semantic search-based system like the current RAGFlow presents critical limitations:

   * Lack of Determinism: An agent cannot reliably instruct the system to "read page 5, lines 10-20." It can only ask questions and receive
     semantically relevant but often unpredictable chunks. This makes precise data extraction, validation, and structured navigation nearly
     impossible.
   * Fragility with Compound Documents: The current task model is not equipped to handle complex document structures, particularly those with embedded
     objects (e.g., OLE). This can lead to silent data loss or catastrophic failures like infinite recursion when circular references (e.g., Doc A
     embedding Doc B which embeds A) are encountered.
   * Inefficient Contextual Awareness: While RAGFlow attempts to provide context, it does so by stitching together semantically related but
     potentially disjointed chunks. An agent, however, often needs to explore the physical/linear context around a point of interest (e.g., "what is
     the paragraph immediately preceding this clause?"), which is not efficiently supported.

  In essence, for an advanced agent, the current RAGFlow acts as a well-meaning but unreliable black box. To be a truly foundational piece of AI
  infrastructure, it must also provide the tools of a predictable and simple data source.

##  3. Proposed Solution: A Hybrid, Multi-Layered DAL Architecture

  We propose to enhance RAGFlow with a deterministic DAL, transforming it into a system that offers the best of both worlds: semantic search for
[...]

PR #13287, opened 2026-03-01
feat: persist PDF bookmark outline as document metadata

yuch85:
## Summary

PDF files often contain a bookmark/outline tree (table of contents built into the file by the authoring tool). RAGFlow's `pdf_parser.outlines` already extracts these `(title, depth)` tuples via pypdf, but they are used ephemerally during chunking (`manual` parser uses them for hierarchy detection) and then discarded.

This PR persists the outline as `doc.meta_fields["outline"]` — a JSON array of `{"title": str, "depth": int}` objects — so downstream features can use the structural information.

### Why this matters

- **Complementary to `toc_extraction`** — the existing `toc_extraction` feature uses LLM calls to generate a TOC and only works for the `naive` parser. The raw PDF outline is free (already extracted by pypdf), works for all parsers, and captures the author's original document structure.
- **Document navigation** — frontends can render a clickable TOC from the outline
- **Entity extraction** — the outline provides a structural map for identifying document sections and key topics
- **Search result context** — knowing which section a chunk belongs to helps users evaluate relevance

### Changes

| File | Change | LOC |
|------|--------|-----|
| `rag/app/naive.py` | Attach `pdf_parser.outlines` as `__outline__` on first chunk dict | ~7 |
| `rag/app/manual.py` | Same for the manual parser | ~5 |
| `rag/svr/task_executor.py` | Extract `__outline__`, persist via `DocMetadataService.update_document_metadata()` | ~12 |

### Design decisions

- **Transient key pattern**: The outline is passed from parser → task_executor via `__outline__` on the first chunk dict, then removed before indexing. This follows the same pattern as `metadata_obj` for LLM-generated metadata.
- **No schema changes**: Uses the existing `meta_fields` JSON column on the document table.
- **Graceful degradation**: If a PDF has no outline (common for scanned docs), nothing is stored. If persistence fails, it logs a warning and continues — parsing is not interrupted.

### Backward compatibility

- **Fully backward compatible** — no existing fields, behavior, or schemas changed
- PDFs without outlines are unaffected
- Existing `meta_fields` data is preserved (merged, not overwritten)

## Test plan

- [ ] Parse a PDF with bookmarks (e.g. any multi-chapter document), verify `meta_fields["outline"]` is populated
- [ ] Parse a PDF without bookmarks, verify no errors and no outline key in meta_fields
- [ ] Verify existing `meta_fields` data is preserved (not overwritten) when outline is added
- [ ] Verify `manual` parser also persists outlines
- [ ] Verify outline JSON structure: `[{"title": "Chapter 1", "depth": 0}, ...]`

Related: #9921 (Deterministic Document Access Layer)


wangq8:
LGTM.

PR #13293, opened 2026-03-01
Fix: reset document status to RUNNING when tasks recover from transient error

### LINKED ISSUE #13285: [Bug]: Inconsistent document parsing status after error recovery
### Self Checks

0

### RAGFlow image version

v0.24.0

```Markdown

```

### Actual behavior

When an error occurs during parsing, the task successfully recovers and completes processing, but the parsing status of the document remains failed instead of being updated accordingly. Here is an example:

Image

Image

### Expected behavior

The status should remain processing while the task is recovering from an error. It should only be marked as failed if a permanent failure occurs.

### Steps to reproduce

```Markdown
Start a parsing task using an external LLM as a PDF parser. If the model returns an overloaded status, the processing eventually completes, but the parsing status is immediately changed to failed as soon as the error occurs.
```

### Additional information

_No response_

Br1an67:

Fixes #13285

When a parsing task encounters a transient error (e.g., LLM overloaded), it temporarily sets `task.progress = -1`. If the periodic `_sync_progress` runs during this window, it marks the document as FAIL (`doc.run = TaskStatus.FAIL`). When the task recovers and continues processing (e.g., `task.progress = 0.5`), the next `_sync_progress` iteration reads `status = doc.run` (which is FAIL) as the default. Since `finished = False` (task is still running), neither the FAIL nor DONE branch updates the status, leaving the document stuck at FAIL even though processing is actively continuing.

This PR adds an explicit `else` clause in `_sync_progress` to reset the document status back to `RUNNING` when there are still unfinished tasks.

### Type of change

### DIFF
--- a/api/db/services/document_service.py
+++ b/api/db/services/document_service.py
@@ -816,6 +816,8 @@ class DocumentService(CommonService):
                 elif finished:
                     prg = 1
                     status = TaskStatus.DONE.value
+                elif not finished:
+                    status = TaskStatus.RUNNING.value
 
                 # only for special task and parsed docs and unfinished
                 freeze_progress = special_task_running and doc_progress >= 1 and not finished
--- a/api/db/services/document_service.py
+++ b/api/db/services/document_service.py
@@ -831,7 +831,7 @@ class DocumentService(CommonService):
                 info = {
                     "process_duration": max(datetime.timestamp(datetime.now()) - begin_at.timestamp(), 0),
                     "run": status}
-                if prg != 0 and not freeze_progress:
+                if (prg != 0 or doc_progress < 0) and not freeze_progress:
                     info["progress"] = prg
                 if msg:
                     info["progress_msg"] = msg
--- a/api/db/services/task_service.py
+++ b/api/db/services/task_service.py
@@ -304,9 +304,10 @@ class TaskService(CommonService):
 
         Update Rules:
             - progress_msg: Always appends the new message to the existing one, and trims the result to max 3000 lines.
-            - progress: Only updates if the current progress is not -1 AND
-                        (the new progress is -1 OR greater than the existing progress),
-                        to avoid overwriting valid progress with invalid or regressive values.
+            - progress: Only updates if (a) new progress >= 1 (task complete),
+                        (b) new progress is -1 (transient error), or
+                        (c) new progress is greater than the existing progress.
+                        This allows recovery from error state (-1) while preventing regressive updates.
 
         Args:
             id (str): The unique identifier of the task to update.
@@ -327,10 +328,7 @@ class TaskService(CommonService):
                 prog = info["progress"]
                 cls.model.update(progress=prog).where(
                     (cls.model.id == id) &
-                    (
-                            (cls.model.progress != -1) &
-                            ((prog == -1) | (prog > cls.model.progress))
-                    )
+                    ((prog >= 1) | (prog == -1) | (prog > cls.model.progress))
                 ).execute()
         else:
             with DB.lock("update_progress", -1):
@@ -341,10 +339,7 @@ class TaskService(CommonService):
                     prog = info["progress"]
                     cls.model.update(progress=prog).where(
                         (cls.model.id == id) &
-                        (
-                            (cls.model.progress != -1) &
-                            ((prog == -1) | (prog > cls.model.progress))
-                        )
+                        ((prog >= 1) | (prog == -1) | (prog > cls.model.progress))
                     ).execute()
 
         process_duration = (datetime.now() - task.begin_at).total_seconds()


tdakanalis:
Does your fix also handle the incorrect progress = -1.0 value?

Br1an67:
@tdakanalis Good catch! The original fix already handled the **status** reset (FAIL → RUNNING) when tasks recover, but there was a subtle gap with the **progress value** itself.

Here's the scenario: when a task recovers from `progress=-1` and restarts at `progress=0`, the computed `prg` is `0`. The existing condition `if prg != 0` then skips updating `doc.progress`, leaving it stuck at `-1.0` even though the status was correctly reset to `RUNNING`.

I've just pushed a fix (8075c8d) that changes the progress update condition from:
```python
if prg != 0 and not freeze_progress:
```
to:
```python
if (prg != 0 or doc_progress < 0) and not freeze_progress:
```

This ensures that stale negative progress values are always overwritten when tasks become active again. Thanks for pointing this out!

Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

tdakanalis:
@Br1an67 I tested your fix and found that one more change is required. Here is a sed command that applies the fix to task_service.py. Please review it and if it is ok include it in the PR.

Fix 2: task_service.py - allow task progress recovery from -1
sed -i '/cls.model.progress != -1/{N;s/((prog >= 1) | ((cls.model.progress != -1) &\n                        ((prog == -1) | (prog > cls.model.progress))))/((prog >= 1) | (prog == -1) | (prog > cls.model.progress))/}' api/db/services/task_service.py

tdakanalis:
Actually, there are a couple of places inside the task_service.py  where you must replace the 

```python
cls.model.update(progress=prog).where(
                        (cls.model.id == id) &
                        ((prog >= 1) | ((cls.model.progress != -1) &
                        ((prog == -1) | (prog > cls.model.progress))))
                    ).execute()
```                    

with

```python
cls.model.update(progress=prog).where(
                        (cls.model.id == id) &
                        ((prog >= 1) | (prog == -1) | (prog > cls.model.progress))
                    ).execute()                  
```

Br1an67:
@tdakanalis Great catch — you're absolutely right. The previous condition `(cls.model.progress != -1)` was blocking all updates when progress was already `-1`, preventing recovery.

I've pushed a fix (54bff9f) that simplifies the WHERE clause to:

```python
(prog >= 1) | (prog == -1) | (prog > cls.model.progress)
```

This restores the `prog >= 1` escape hatch from the original code (which I'd inadvertently removed) and removes the `cls.model.progress != -1` guard entirely, so tasks can always recover from error state. Thanks for testing and providing the detailed fix suggestion!

KevinHuSh:
> ```python
> prog >= 1
> ```
@Br1an67

How come `prog >= 1` is true, but `prog > cls.model.progress` is false?

xugangqiang:
### Test coverage gap

This PR fixes a real state-machine bug (transient `progress = -1` followed by recovery) but ships **no tests**. Both changed pieces are pure logic and very testable, so a regression could easily slip back in:

- `TaskService.update_progress` WHERE condition — assert that a task already at `progress = -1` **can** be updated to a recovering value (e.g. `0.5`), and that regressive updates are still blocked.
- `DocumentService._sync_progress` — assert that when `doc.run == FAIL` while there are still unfinished tasks (`finished = False`), the synced status is reset to `RUNNING`, whereas a permanently failed task (`finished and bad`) correctly stays `FAIL`.

A couple of focused unit tests (mocked DB, no external services) would lock in the fix. Could you please add them before merge?

xugangqiang:
### On the `prog >= 1` condition

> How come `prog >= 1` is true, but `prog > cls.model.progress` is false?

That happens exactly when the **stored** progress is already `>= 1` and the incoming `prog` is also `>= 1` but not strictly greater — e.g. `cls.model.progress = 1.0` and `prog = 1.0`. In that case:

- `prog >= 1` → **True**
- `prog > cls.model.progress` → `1.0 > 1.0` → **False**

So the first disjunct forces an `UPDATE` that the original code skipped — a redundant no-op write of `1.0` over an already-`1.0` task.

The `prog >= 1` term is **not needed** for the actual fix. The bug was solely the `cls.model.progress != -1` guard blocking every update once a task hit `-1`. Dropping that guard alone already allows recovery, because any normal recovering progress (e.g. `0.5`) satisfies `prog > cls.model.progress` (`0.5 > -1`).

So the condition can be simplified to:

```python
((prog == -1) | (prog > cls.model.progress))
```

which preserves every original behaviour (forward progress and `-1` error marking) **plus** recovery from `-1`, and avoids the redundant overwrite of an already-completed task. This also matches the updated docstring ("update if new progress is -1 OR greater than existing").

PR #13297, opened 2026-03-02
 Fix: Use default pipeline configuration when parsing documents via HTTP API

null-whale:

Currently, when triggering document parsing via HTTP API, the system does not use the default pipeline process configured in the knowledge base. Even if an administrator has set a specific pipeline as the default in the knowledge base settings, API-triggered parsing tasks still use an unexpected process, resulting in parsing outcomes that don't match expectations.

### Type of change


Magicbook1108:
LGTM.

yingfeng:
Please fix the CI error

yingfeng:
Please fix the CI failure

PR #13303, opened 2026-03-02
feat(auth): support external oauth/oidc token sign-in

isolomatov-gd:
## Summary
- add a new API endpoint `POST /oauth/external/<channel>` for signing in with externally provided `access_token` and optional `id_token`
- mirror `oauth_callback` login/register behavior, including creating a new user when the OAuth/OIDC user is not found
- return API error codes/messages instead of redirects for all failure paths
- return auth via `construct_response(..., auth=user.get_id())` on successful login

## Validation
- added unit tests for the new endpoint in `test_user_app_unit.py`
- ran targeted tests:
  - `ZHIPU_AI_API_KEY=dummy ./.venv/bin/python -m pytest -q test/testcases/test_web_api/test_user_app/test_user_app_unit.py -k "oauth_callback_matrix_unit or oauth_external_matrix_unit"`

### DIFF
--- a/api/apps/user_app.py
+++ b/api/apps/user_app.py
@@ -268,6 +268,110 @@ async def oauth_callback(channel):
         return redirect(f"/?error={str(e)}")
 
 
+@manager.route("/oauth/external/<channel>", methods=["POST"])  # noqa: F821
+async def oauth_external(channel):
+    """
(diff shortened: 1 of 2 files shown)


yingfeng:
Hi, can you add the API docs accordingly? Thanks~

PR #13341, opened 2026-03-03
Fix: allow document parsing status recovery after transient errors

### LINKED ISSUE #13285: [Bug]: Inconsistent document parsing status after error recovery

key4127:

Fixes #13285

When an LLM returns a transient error (e.g. overloaded) during parsing, the task progress is set to -1. Previously, the progress could never be updated again, leaving the document permanently stuck in FAIL status even after the task successfully recovered and completed.

Three coordinated changes address this:

1. task_service.update_progress: relax the progress update guard to accept prog >= 1 even when current progress is -1, so a task that recovers from a transient failure can report completion.

2. document_service.get_unfinished_docs: include documents that are marked FAIL (progress == -1) but still have at least one non-failed task (task.progress >= 0) in the polling set, so their status can be re-synced once a task recovers. Documents where all tasks have permanently failed are excluded to avoid unnecessary polling.

3. document_service.update_progress: explicitly set document status to RUNNING when not all tasks have finished, instead of preserving whatever stale status (potentially FAIL) the document previously had.

### Type of change


yingfeng:
Please fix the conflict, thank you~

PR #13351, opened 2026-03-03
Test: add scenario for embedding_model update when chunk_count > 0

Achieve3318:

Guard embedding_model change when dataset has existing chunks. API must return code 102 with message 'When chunk_num (N) > 0, embedding_model must remain <current_model>' to prevent silent embedding drift.

### Type of change

- [x] Add Testcases


Achieve3318:
Hi, @KevinHuSh , Please review my PR

PR #13371, opened 2026-03-04
fix: SSO login redirects back to login page after successful auth

liuxiaoyusky:
## Summary

- **Backend**: Add Bearer prefix stripping in `_load_user()` (`api/apps/__init__.py`) so JWT verification works regardless of whether the frontend sends `Authorization: Bearer <token>` or just `Authorization: <token>`
- **Frontend**: Add auth guard in `NextLayout` (`web/src/layouts/next.tsx`) using `useAuth()` hook to prevent child components from rendering (and firing API requests) before login status is confirmed, eliminating the race condition where empty Authorization headers trigger 401 → localStorage clear → redirect to /login

## Root Cause Analysis

Two independent bugs introduced during the v0.22 → v0.24 migration:

1. Commit `46dec98f5` re-added `'Bearer ' + auth` prefix in `getAuthorization()` (for embedded page compatibility), but the backend `_load_user()` never handled the prefix → JWT signature mismatch → 401
2. Commit `4b50b8c57` replaced `wrappers: ['@/wrappers/auth']` with a route `loader`, but the loader only stores the token and redirects — it doesn't gate rendering on auth status. This causes child components to fire API requests before auth state is stable.

## Test plan

- [x] Backend: `curl` with both `Authorization: <token>` and `Authorization: Bearer <token>` returns `code=0`
- [x] E2E: Feishu SSO login → lands on home page without redirect loop
- [ ] Verify normal username/password login still works
- [ ] Verify embedded page auth (`?jwt_auth=`) still works

### DIFF
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -99,6 +99,9 @@ def _load_user():
     if not authorization:
         return None
 
+    if authorization.startswith("Bearer "):
+        authorization = authorization[7:]
+
     try:
         access_token = str(jwt.loads(authorization))
 
--- a/web/src/layouts/next.tsx
+++ b/web/src/layouts/next.tsx
@@ -1,7 +1,17 @@
+import { useAuth } from '@/hooks/auth-hooks';
+import { redirectToLogin } from '@/utils/authorization-util';
 import { Outlet } from 'react-router';
 import { Header } from './next-header';
 
 export default function NextLayout() {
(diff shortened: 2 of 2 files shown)


yingfeng:
Thank you for your contribution, but we cannot merge this PR because it contains too many things that shouldn't be included, such as private keys, conflicts, errors, and unrelated changes, etc. Therefore, I have to close it.

PR #13372, opened 2026-03-04
Refa:improve excel parser logic

Woody-Hu:

improve excel parser logic

### Type of change
- [x] Refactoring


JinHai-CN:
LGTM

PR #12488, opened 2026-01-07
Fix SDK authentication to return consistent error responses

maoxiaowang:
Fix SDK authentication to return consistent error responses.

Previously, missing or invalid API keys returned HTTP 200 with inconsistent error messages and codes.
This made it unreliable for SDK clients to detect authentication failures and handle them correctly.

### What problem does this PR solve?

- Updated `token_required` decorator to:
  - Preserve `tenant_id` injection
  - Raise `Unauthorized` for missing or invalid API keys
- Added a global `app.errorhandler` for Unauthorized:
  - Returns structured error response with **HTTP 401**
- Authentication failures now consistently return either:
  - HTTP 401 or response code 401 (for legacy compatibility)

### Frontend Impact
- Frontend now properly handles HTTP 401 responses, and redirects to login page as expected.
- No changes to existing pages or authentication logic.

### Type of change

- [x] Performance Improvement

PR #13386, opened 2026-03-05
fix(auth): return HTTP 401 for token authentication failures

maoxiaowang:
Follow-up to #12488 

Previously, token authentication failures returned HTTP 200
with an error code in the response body.

This PR updates `token_required` to raise `Unauthorized`
and relies on the global error handler to return
a structured JSON response with HTTP 401.

The response body structure (`code`, `message`, `data`)
remains unchanged to preserve compatibility with the
official SDK.

Frontend logic has been updated to handle HTTP 401
responses in addition to checking `data.code`.

### Type of change


asiroliu:
@maoxiaowang 
Since you modified the HTTP response format, you need to update the corresponding test cases.

PR #13389, opened 2026-03-05
fix: return structured JSON output for non-streaming agent API

Kingsuperyzy:

Previously, when an Agent component was configured with structured output, the non-streaming /agents/{agent_id}/completions API never returned the structured field in its response. 

The root cause: the non-streaming code path only collected message events to build full_content, then returned the workflow_finished payload — which only contains the output of the last component in the execution path (typically a Message component).
 Any structured output set by upstream components (e.g., Agent or LLM) was silently discarded. 

This PR fixes the non-streaming handler to iterate node_finished events and collect structured output from intermediate components.
 If any component produced a non-empty structured value, it is included in the final response under data.structured. The streaming path is unaffected, as it already exposes node_finished events to the caller.

### Type of change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -222,4 +222,5 @@ docker/launch_backend_service_windows.sh
 
 # Go server build output
 bin/
-internal/
+internal/cpp/cmake-build-release/
+internal/cpp/cmake-build-debug/


Kingsuperyzy:
@yongtenglei @yingfeng  Would it be possible to merge the current PR into the main branch?

yongtenglei:
Hi @Kingsuperyzy 

Thank you for your contribution. This change looks good to me. It might be better to add relevant documentation and cover more cases (now only for non-streaming case). I think I will fill them later.

Cheers.

Issue #13366, opened 2026-03-04
[Question]:I have currently uploaded a PDF and the parsing method is Deepdoc, but the parsed content is garbled. What is the reason for this?


lovetj:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

I have currently uploaded a PDF and the parsing method is Deepdoc, but the parsed content is garbled. What is the reason for this?

<img width="1860" height="736" alt="Image" src="https://github.com/user-attachments/assets/637a4607-f998-4683-a94a-2b492e916e7a" />

Magicbook1108:
Maybe dosu is correct. 
<img width="2360" height="1376" alt="Image" src="https://github.com/user-attachments/assets/ee56bd1d-ced7-46fb-867f-5322011cd67c" />

lovetj:
[...]

PR #13404, opened 2026-03-05
fix: detect and fallback garbled PDF text to OCR (#13366)

tunsuy:
## Problem

When PDF fonts lack ToUnicode/CMap mappings, pdfplumber (pdfminer) cannot map CIDs to correct Unicode characters, outputting PUA characters (U+E000~U+F8FF) or `(cid:xxx)` placeholders. The original code fully trusted pdfplumber text without any garbled detection, causing garbled output in the final parsed result.

Relates to #13366

### 1. Garbled text detection functions
- `_is_garbled_char(ch)`: Detects PUA characters (BMP/Plane 15/16), replacement character U+FFFD, control characters, and unassigned/surrogate codepoints
- `_is_garbled_text(text, threshold)`: Calculates garbled ratio and detects `(cid:xxx)` patterns

### 2. Box-level fallback (in `__ocr()`)
When a text box has ≥50% garbled characters, discard pdfplumber text and fallback to OCR recognition.

### 3. Page-level detection (in `__images__()`)
Sample characters from each page; if garbled rate ≥30%, clear all pdfplumber characters for that page, forcing full OCR.

### 4. Layout recognizer CID filtering
Filter out `(cid:xxx)` patterns in `layout_recognizer.py` text processing to prevent them from polluting layout analysis.

## Testing
- 29 unit tests covering: normal CJK/English text, PUA characters, CID patterns, mixed text, boundary thresholds, edge cases
- All 85 existing project unit tests pass without regression


Magicbook1108:
I don't think your code correctly fix the issue. Can you take a screen shot of your succeed condition? 
image
[GB.18067-2000.pdf](https://github.com/user-attachments/files/25765679/GB.18067-2000.pdf)

tunsuy:
> I don't think your code correctly fix the issue. Can you take a screen shot of your succeed condition? image [GB.18067-2000.pdf](https://github.com/user-attachments/files/25765679/GB.18067-2000.pdf)

Updated with the fix. Here's the verification result for GB.18067-2000.pdf:

Detection: All 4 garbled pages (1, 3, 5, 8) with subset font encoding (e.g. BCDGEE+R0015) are correctly identified — CJK ratio < 5%, ASCII punctuation ratio > 40%.

Before fix: pdfplumber extracts garbled ASCII symbols (random punctuation like ()[]{} instead of Chinese text).
After fix: Garbled chars are cleared, OCR fallback is triggered, and Chinese content is successfully extracted.

Screenshot attached below. All 50 unit tests passed.

Clipboard_Screenshot_1772766111

Clipboard_Screenshot_1772766132

Clipboard_Screenshot_1772766561

Clipboard_Screenshot_1772766577

deepdoc/parser/pdf_parser.py
```diff
@@ -1400,6 +1584,34 @@ def __images__(self, fnm, zoomin=3, page_from=0, page_to=299, callback=None):
...
 
+                    # Detect garbled pages and clear their chars so the OCR
+                    # path will be used instead. Two detection strategies:
+                    # 1) PUA / unmapped CID characters (threshold=0.3)
+                    # 2) Font-encoding garbling: subset fonts mapping CJK to ASCII
+                    for pi, page_ch in enumerate(self.page_chars):
+                        if not page_ch:
+                            continue
+                        # Strategy 1: PUA / CID garbling
+                        sample = page_ch if len(page_ch) <= 200 else random.sample(page_ch, 200)
```
Magicbook1108: This could make the results unstable and should be changed.

asiroliu:
@tunsuy 
1. Reference the implementation code instead of duplicating it
2. Add unit tests for layout_recognizer.__is_garbage
3. Follow the existing unit test structure and place the test file in the appropriate location

deepdoc/parser/pdf_parser.py
```diff
@@ -197,6 +198,163 @@ def _has_color(self, o):
...
+
+        for c in page_chars:
+            text = c.get("text", "")
+            fontname = c.get("fontname", "")
+            if not text or text.isspace():
+                continue
+            total_non_space += 1
+
+            if cls._has_subset_font_prefix(fontname):
+                has_subset_font = True
```
Magicbook1108: A single subset font can cause the entire page to be classified as font-encoding garbled.

deepdoc/parser/pdf_parser.py
tunsuy: Fixed. Changed random.sample(page_ch, 200) to page_ch[:200] for deterministic results. Since garbled text is a page-level issue (the entire page's font encoding is broken), the first 200 characters are representative enough for detection.

deepdoc/parser/pdf_parser.py
tunsuy: Fixed. Replaced the boolean has_subset_font flag with a subset_font_count counter and ratio check (subset_ratio = subset_font_count / total_non_space). Now the page is only considered potentially garbled when subset font characters account for ≥ 30% of all characters, avoiding false positives from pages with only a few subset font characters.

tunsuy:
> @tunsuy
> 
> 1. Reference the implementation code instead of duplicating it
> 2. Add unit tests for layout_recognizer.__is_garbage
> 3. Follow the existing unit test structure and place the test file in the appropriate location

Okay, I've already taken care of it.

asiroliu:
@tunsuy 
The test cases should be located in test/unit_test/deepdoc/parser/test_pdf_garbled_detection.py

tunsuy:
> @tunsuy The test cases should be located in test/unit_test/deepdoc/parser/test_pdf_garbled_detection.py

Okay, I've already taken care of it.

asiroliu:
Please fix CI

asiroliu (approved):

PR #13427, opened 2026-03-05
fix(agent): ensure database connections are properly closed in ExeSQL tool

edenfunf:
## Summary

Fix a database connection and cursor resource leak in the ExeSQL agent tool.

When SQL execution raises an exception (for example syntax error or missing table),
the existing code path skips `cursor.close()` and `db.close()`, causing database
connections to accumulate over time.

This can eventually lead to connection exhaustion in long-running agent workflows.

## Root Cause

The cleanup logic for database cursors and connections is placed after the SQL
execution loop without `try/finally` protection. If an exception occurs during
`cursor.execute()`, `fetchmany()`, or result processing, the cleanup code is not
reached and the connection remains open.

The same issue also exists in the IBM DB2 execution path where `ibm_db.close(conn)`
may be skipped when exceptions occur.

## Fix

- Wrap SQL execution logic in `try/finally` blocks to guarantee resource cleanup.
- Ensure `cursor.close()` and `db.close()` are always executed.
- Add explicit `db.close()` when `db.cursor()` creation fails.
- Remove redundant close calls in early-return branches since `finally` now handles cleanup.

## Impact

- No change to normal execution behavior.
- Ensures database resources are always released when errors occur.
- Prevents connection leaks in long-running workflows.
- Only affects `agent/tools/exesql.py`.

## Testing

Manual test scenarios:

1. Valid SQL execution
2. SQL syntax error
3. Query against a non-existing table
4. Execution cancellation during query

In all scenarios the database cursor and connection are properly closed.

Code quality checks:

- `ruff check` passed
- No new warnings introduced

### DIFF
--- a/agent/tools/exesql.py
+++ b/agent/tools/exesql.py
@@ -251,7 +251,7 @@ class ExeSQL(ToolBase, ABC):
                 if self.check_if_canceled("ExeSQL processing"):
                     return
 
-                single_sql = single_sql.replace('```','')
+                single_sql = single_sql.replace('```', '').strip()
                 if not single_sql:
                     continue
                 single_sql = re.sub(r"\[ID:[0-9]+\]", "", single_sql)


yingfeng:
Thanks for your contribution, pls fix according to the feedback of copilot

PR #13431, opened 2026-03-06
Feat：Using Go to implement user registration logic

dcc123456:

Feat：Using Go to implement user registration logic

### Type of change


internal/service/user.go
```diff
@@ -22,10 +22,12 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/pem"
-	"errors"
+	stderrors "errors"
```
JinHai-CN: why not errors?

PR #13435, opened 2026-03-06
Feat: add switch_chunks endpoint to manage chunk availability

huang-aoqin:
### What problem does this commit solve?

This commit introduces a new API endpoint `/datasets/<dataset_id>/documents/<document_id>/chunks/switch` that allows users to switch the availability status of specified chunks in a document as same as chunk_app.py

### Type of change

### DIFF
--- a/docs/references/http_api_reference.md
+++ b/docs/references/http_api_reference.md
@@ -1901,101 +1901,6 @@ Failure:
 
 ---
 
-### Switch chunks availability
-
-**POST** `/api/v1/datasets/{dataset_id}/documents/{document_id}/chunks/switch`
-
-Switches the availability of specified chunks (enable or disable chunks for retrieval).
-
-#### Request
-
-- Method: POST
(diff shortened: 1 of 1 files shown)


yingfeng:
Since this is an API change, can you add docs accordingly?

huang-aoqin:
> Since this is an API change, can you add docs accordingly?

Added

writinwaters:
Hi @huang-aoqin, What is the expected behavior if I set `available_int` to `1` and `available` to `false`? The code doesn't seem to have this error handling? Second, I'd like to confirm whether this API "sets/updates the availability status for specified chunks?

Cheers.

writinwaters:
Am just wondering if the wording "switch" has specific meaning.

huang-aoqin:
> Am just wondering if the wording "switch" has specific meaning.


Since the state-switching endpoint in /api/apps/chunk_app.py is named "switch", I have kept the naming of the two endpoint with the same function consistent

PR #13468, opened 2026-03-09
fix: add missing authentication and improve credential validation

kinokopio:
PR Body
- Add @login_required decorator to MCP server endpoint
- Simplify check_admin() to perform pure credential validation
- Add superuser role verification for admin access
This PR addresses authentication and authorization issues in the admin service:
1. **MCP Server Endpoint**: Added missing `@login_required` decorator to ensure consistent authentication across all MCP endpoints.
2. **Admin Credential Validation**: Improved `check_admin()` function to perform pure credential validation with proper superuser role verification, removing redundant initialization logic that belongs in the startup flow.
3. This PR fixes security vulnerabilities in the authentication flow.
Fixes https://github.com/infiniflow/ragflow/security/advisories/GHSA-24q2-5x74-9fgp
### Type of change
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


gambletan (commented):
This PR addresses a real security issue (GHSA-24q2-5x74-9fgp). Two changes here:

**1. `@login_required` on `/test_mcp` endpoint (mcp_server_app.py, line 402)**

Good catch — this endpoint was missing authentication, which means any unauthenticated user could test arbitrary MCP server URLs. This could be abused for SSRF (server-side request forgery) since the server would make outbound requests to attacker-controlled URLs. Adding `@login_required` is the right fix.

**2. `check_admin()` simplification (auth.py)**

This is the more significant change. The old code had a concerning pattern: if the username wasn't found, it would *create* a new superuser account with hardcoded credentials (`admin` / `admin@ragflow.io`). This is a security vulnerability — any authentication attempt with a non-existent username would silently create an admin account.

The new code is much cleaner: it simply validates credentials and checks `is_superuser`. However, I have one concern:

- **Initialization dependency**: The old code doubled as a "bootstrap admin if none exists" mechanism. With this removal, the system now depends on the admin user being created during the proper startup/initialization flow. If that init step is skipped or fails, there would be no way to log into the admin panel. Is there a separate initialization path that creates the default admin user? If not, this could lock users out of fresh deployments.

- **The `user.is_superuser` check is a good addition**: Previously, any valid user could pass `check_admin()`. Now only superusers can access admin endpoints, which is correct.

Minor: the function no longer needs `import uuid` or `encode_to_base64` — if those were only used here, they could be cleaned up in a follow-up.

Overall this is a solid security fix. Just want to confirm the admin bootstrap path exists elsewhere.

yongtenglei (approved):

kinokopio:
> gambletan

Thanks for the review @gambletan!

> Is there a separate initialization path that creates the default admin user?

Yep. `init_default_admin()` in [`admin/server/auth.py` (lines 75-99)](https://github.com/infiniflow/ragflow/blob/main/admin/server/auth.py#L75-L99) handles this. It runs at startup in [`admin/server/admin_server.py` (line 67)](https://github.com/infiniflow/ragflow/blob/main/admin/server/admin_server.py#L67):

```python
# admin/server/admin_server.py:67
init_default_admin()
```

It checks for any superuser (`is_superuser=True`), and creates one if none exists. Also sets up the tenant properly via `add_tenant_for_admin()`.

---

> If that init step is skipped or fails, there would be no way to log into the admin panel.

A few things here:

1. If `init_default_admin()` fails, it throws an `AdminException` and the service won't start. Makes sense - no admin, no service.

   [`admin/server/auth.py` (lines 88-89)](https://github.com/infiniflow/ragflow/blob/main/admin/server/auth.py#L88-L89):
   ```python
   if not UserService.save(**default_admin):
       raise AdminException("Can't init admin.", 500)
   ```

2. It can't be skipped - it's called before the server starts listening.

   [`admin/server/admin_server.py` (lines 67-72)](https://github.com/infiniflow/ragflow/blob/main/admin/server/admin_server.py#L67-L72):
   ```python
   init_default_admin()
   SERVICE_CONFIGS.configs = load_configurations(SERVICE_CONF)

   try:
       logging.info(f"RAGFlow admin is ready after {time.time() - start_ts}s initialization.")
       run_simple(...)
   ```

3. There's also a check for inactive admins - if all superusers have `is_active=0`, it refuses to start.

   [`admin/server/auth.py` (lines 91-92)](https://github.com/infiniflow/ragflow/blob/main/admin/server/auth.py#L91-L92):
   ```python
   elif not any([u.is_active == ActiveEnum.ACTIVE.value for u in users]):
       raise AdminException("No active admin. Please update 'is_active' in db manually.", 500)
   ```

---

> If not, this could lock users out of fresh deployments.

No, users won't be locked out. Here's what happens if someone deletes all admin accounts:

| Action | Result |
|--------|--------|
| Delete all admins → Restart Admin service | `init_default_admin()` creates new admin `admin@ragflow.io` / `admin` |
| Delete all admins → Don't restart | Can't login to Admin panel (but main service still works) |

So worst case, just restart the Admin service and you're back in. The `init_default_admin()` function checks `is_superuser=True` users at startup - if none exist, it creates one.

---

kinokopio:
@KevinHuSh Could you please review this vulnerability: GHSA-24q2-5x74-9fgp

KevinHuSh:
Appreciations!

Please resolve the conflicts.

kinokopio:
> Appreciations!
> 
> Please resolve the conflicts.

Please reveiw my report https://github.com/infiniflow/ragflow/security/advisories/GHSA-24q2-5x74-9fgp

JinHai-CN:
@kinokopio would you please resolve the conflicts?

PR #13484, opened 2026-03-09
fix: avoid empty doc filter in knowledge retrieval

JosefAschauer:
## Summary
Fix knowledge-base chat retrieval when no individual document IDs are selected.

## Root Cause
`async_chat()` initialized `doc_ids` as an empty list when the request did not explicitly select documents. That empty list was then forwarded into retrieval as an active `doc_id` filter, effectively becoming `doc_id IN []` and suppressing all chunk matches.

## Changes
- treat missing selected document IDs as `None` instead of `[]`
- keep explicit document filtering when IDs are actually provided
- add regression coverage for the shared chat retrieval path

## Validation
- `python3 -m py_compile api/db/services/dialog_service.py test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py`
- `.venv/bin/python -m pytest test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py`
- manually verified that chat completions again inject retrieved knowledge into the prompt


JosefAschauer:
Related PRs for this workstream:

- #13485 fixes a separate generic chunk API issue where raw chunk content could be replaced by a highlight/snippet in list responses.
- #13486 adds the Qdrant document-store backend.

I split these intentionally so the two generic fixes can be reviewed and merged independently of the Qdrant backend work.

yongtenglei:
Hi, @JosefAschauer 

Please make sure you know what you are doing. Check your code before opening​ a PR.


image


image

Cheers.

JosefAschauer:
Will fix this today, this will not happen again
Cheers. 
J

JosefAschauer:
Fixed in 29854efa. The test file had merge artifacts from the branch split. I pushed without verifying the remote diff against the local state.

Lesson learned: verify the pushed branch, not just the local one.

JosefAschauer:
Verified the real retrieval path as well: `retrieval()` passes `doc_ids=None` through, and `Dealer.get_filters()` only adds a `doc_id` filter when `doc_ids is not None`, so `None` means no document-level filter and retrieval runs across all docs in the selected KBs. The buggy case was `[]`, which became an active empty filter and excluded every chunk.

PR #13487, opened 2026-03-09
feat(evaluation): track token usage in evaluation results

kiannidev:
## Summary

Implements the TODO in `evaluation_service.py`: **Track token usage** in evaluation results.

## Changes

- **Import** `num_tokens_from_string` from `common.token_utils`
- **Prompt tokens**: Use the full prompt returned by `async_chat` when available (includes system prompt + knowledge base + query), otherwise fall back to the question token count
- **Completion tokens**: Count tokens in the generated answer
- **Storage**: Store `token_usage` as `{prompt_tokens, completion_tokens, total_tokens}` in each `EvaluationResult` instead of `None`

## Why

The evaluation pipeline previously saved `token_usage: None` for every result. This change allows downstream consumers (e.g. evaluation dashboards, cost tracking) to see approximate token usage per test case using the same tokenizer (tiktoken cl100k_base) used elsewhere in RAGFlow.

## Testing

- No new tests added; existing evaluation flow unchanged
- Token counting uses existing `num_tokens_from_string` utility

### DIFF
--- a/api/db/services/evaluation_service.py
+++ b/api/db/services/evaluation_service.py
@@ -39,6 +39,7 @@ from api.db.services.dialog_service import DialogService
 from common.misc_utils import get_uuid
 from common.time_utils import current_timestamp
 from common.constants import StatusEnum
+from common.token_utils import num_tokens_from_string
 
 
 class EvaluationService(CommonService):
@@ -417,6 +418,8 @@ class EvaluationService(CommonService):
                     answer = ans.get("answer", "")
                     retrieved_chunks = ans.get("reference", {}).get("chunks", [])
                     break
+            else:
+                ans = {}
 
             execution_time = timer() - start_time
 
@@ -430,6 +433,19 @@ class EvaluationService(CommonService):
                 dialog=dialog
             )
 
+            # Track token usage: use full prompt from async_chat when available
+            full_prompt = ans.get("prompt", "")
+            if full_prompt:
+                prompt_tokens = num_tokens_from_string(full_prompt)
+            else:
+                prompt_tokens = num_tokens_from_string(case.get("question", "") or "")
+            completion_tokens = num_tokens_from_string(answer or "")
+            token_usage = {
+                "prompt_tokens": prompt_tokens,
+                "completion_tokens": completion_tokens,
+                "total_tokens": prompt_tokens + completion_tokens,
+            }
+
             # Save result
             result_id = get_uuid()
             result = {
@@ -440,7 +456,7 @@ class EvaluationService(CommonService):
                 "retrieved_chunks": retrieved_chunks,
                 "metrics": metrics,
                 "execution_time": execution_time,
-                "token_usage": None,  # TODO: Track token usage
+                "token_usage": token_usage,
                 "create_time": current_timestamp()
             }
 
--- a/api/db/services/evaluation_service.py
+++ b/api/db/services/evaluation_service.py
@@ -420,6 +420,10 @@ class EvaluationService(CommonService):
                     break
             else:
                 ans = {}
+                logging.warning(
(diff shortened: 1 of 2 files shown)


gambletan (commented):
The implementation addresses the TODO cleanly, but I have a few concerns:

1. **Token counting accuracy**: `num_tokens_from_string` uses tiktoken (cl100k_base), which gives accurate counts for OpenAI models but will be an approximation for other LLM providers (e.g., local models, Anthropic, etc.). This is acknowledged in the PR description as "approximate," which is fair, but it might be worth adding a comment in the code noting this is an estimate — especially since a "token_usage" field suggests precision to downstream consumers.

2. **The `else: ans = {}` addition (line 421)**: This is a subtle but important fix — without it, if the `for` loop completes without `break` (i.e., all streamed chunks are consumed but none contain an answer), `ans` would be unbound when accessed later at `ans.get("prompt", "")`. Good catch. However, should there also be a log warning here? An evaluation case that produces no answer seems like it warrants visibility.

3. **Missing `prompt` key scenario**: When `ans` doesn't contain a `"prompt"` key (which seems possible depending on `async_chat` behavior), the fallback to counting only the question tokens significantly undercounts the actual prompt tokens, since the real prompt includes system instructions + retrieved chunks. This could be misleading for cost tracking. Consider documenting this limitation or logging when the fallback path is taken.

4. **No tests added**: The PR description notes this, but given that this is a new feature (token tracking), even a minimal unit test verifying the `token_usage` dict structure would add confidence, especially around the `ans.get("prompt", "")` fallback behavior.

The change itself is straightforward and non-breaking. The `else: ans = {}` fix is a bonus improvement.

kiannidev:
> The implementation addresses the TODO cleanly, but I have a few concerns:
> 
> 1. **Token counting accuracy**: `num_tokens_from_string` uses tiktoken (cl100k_base), which gives accurate counts for OpenAI models but will be an approximation for other LLM providers (e.g., local models, Anthropic, etc.). This is acknowledged in the PR description as "approximate," which is fair, but it might be worth adding a comment in the code noting this is an estimate — especially since a "token_usage" field suggests precision to downstream consumers.
> 2. **The `else: ans = {}` addition (line 421)**: This is a subtle but important fix — without it, if the `for` loop completes without `break` (i.e., all streamed chunks are consumed but none contain an answer), `ans` would be unbound when accessed later at `ans.get("prompt", "")`. Good catch. However, should there also be a log warning here? An evaluation case that produces no answer seems like it warrants visibility.
> 3. **Missing `prompt` key scenario**: When `ans` doesn't contain a `"prompt"` key (which seems possible depending on `async_chat` behavior), the fallback to counting only the question tokens significantly undercounts the actual prompt tokens, since the real prompt includes system instructions + retrieved chunks. This could be misleading for cost tracking. Consider documenting this limitation or logging when the fallback path is taken.
> 4. **No tests added**: The PR description notes this, but given that this is a new feature (token tracking), even a minimal unit test verifying the `token_usage` dict structure would add confidence, especially around the `ans.get("prompt", "")` fallback behavior.
> 
> The change itself is straightforward and non-breaking. The `else: ans = {}` fix is a bonus improvement.

Hi, @gambletan
Thanks for your review.
I updated code from your feedback.
Please check it.

Thanks again.

kiannidev:
Hi, @gambletan 
Please review this PR..
Thanks

kiannidev:
Hi, @gambletan 
Please review this PR when you have time.

Thanks

kiannidev:
Hi, @gambletan 
Please review this PR

kiannidev:
Hi, @gambletan 
Can you review this PR?

kiannidev:
Hi, @gambletan , @KevinHuSh 
Please check this PR

PR #13535, opened 2026-03-11
fix: treat missing doc filters as no filter across retrieval APIs

JosefAschauer:
## Summary
Follow-up to #13484.

This fixes the same bug pattern in the remaining retrieval entrypoints: callers that mean "no document filter" were passing `[]`, but `rag/nlp/search.py` only skips the `doc_id` filter when `doc_ids is None`.

## Why?
Passing `[]` here is not a harmless empty default. It becomes an active empty `doc_id` filter, so retrieval silently returns no chunks.

In practice, that means a normal RAG request can degrade into zero-recall retrieval: the pipeline still runs, but it behaves as if the KB has no relevant content.

Affected paths on current `main` before this patch:
- `api/db/services/dialog_service.py`
  - `async_chat()` (same bug as #13484)
  - `async_ask()`
  - `gen_mindmap()`
- `api/apps/sdk/session.py`
  - `/searchbots/retrieval_test`
- `api/apps/chunk_app.py`
  - `/retrieval_test`
- `api/apps/sdk/dify_retrieval.py`
  - `/dify/retrieval`

## Fix
- normalize missing doc filters to `None`
- preserve explicit document selections as lists
- filter out stray empty strings
- keep the Dify `[-999]` sentinel for the explicit "metadata filter matched no docs" case

## Tests
- `pytest test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py`
- `pytest --noconftest test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py::test_searchbots_retrieval_test_embedded_matrix_unit`
- `pytest --noconftest test/testcases/test_web_api/test_chunk_app/test_chunk_routes_unit.py::test_retrieval_test_branch_matrix_unit`
- `pytest --noconftest test/testcases/test_http_api/test_dataset_management/test_dify_retrieval_routes_unit.py::test_retrieval_uses_all_docs_when_metadata_condition_is_absent`
- `ruff check api/db/services/dialog_service.py api/apps/sdk/session.py api/apps/chunk_app.py api/apps/sdk/dify_retrieval.py test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py test/testcases/test_web_api/test_chunk_app/test_chunk_routes_unit.py test/testcases/test_http_api/test_dataset_management/test_dify_retrieval_routes_unit.py`


alvinttang (commented):
Changing the sentinel from `[]` to `None` is the right semantic fix — an empty list and an absent filter are not the same thing, and downstream retrieval code was treating both as "filter to zero docs" rather than "no filter applied". The PR is consistent across all entry points (chunk_app, dify_retrieval, session, dialog_service, async_ask, gen_mindmap) and the new unit tests assert `doc_ids is None` rather than just falsy, which properly distinguishes the two cases. Watch out for any callers further downstream that iterate `doc_ids` without a None guard — the test coverage added here validates the boundary up to the retriever call, so it is worth grepping for `for ... in doc_ids` inside retrieval internals to confirm. The bogus sentinel `["-999"]` when metadata filtering returns empty is retained, which is correct — that case means "filter matched nothing, return nothing".

JosefAschauer:
Thanks. That broader audit is exactly how this PR was put together.

The key downstream gate is the retrieval filter builder in `rag/nlp/search.py`: `doc_ids=None` means "no doc-level filter", while `[]` stays an active empty filter. I walked the retrieval entrypoints that feed `settings.retriever.retrieval(...)` and fixed the remaining places where an absent doc filter could still reach retrieval as `[]`: `async_chat` / `async_ask` / `gen_mindmap`, `/searchbots/retrieval_test`, `/retrieval_test`, and `/dify/retrieval`.

`sdk/doc.py` already had the correct `None` pattern and was the reference implementation here. The explicit metadata no-match case intentionally keeps the sentinel/no-result behavior.

JosefAschauer:
Refreshed after #13484 merged. This PR now keeps only the remaining empty-doc-filter call sites; the overlapping async_chat fix and its regression coverage were dropped because they are already merged via #13484.

JosefAschauer:
Rebased this branch onto current main and resolved the resulting conflicts. No semantic changes were intended beyond rebase/conflict-resolution hygiene; the branch is now up to date and ready for another review pass.

JosefAschauer:
@alvinttang Hi, any feedback on this? Happy to adjust if needed!

JinHai-CN:
@JosefAschauer would you please resolve the conflicts?

JosefAschauer:
@JinHai-CN Rebased onto current `main` — the branch is now a single commit on top of `2af732d60` and shows as mergeable.

This was more than a textual rebase: three of the four files the PR touched no longer exist. Summary of where the change landed and why.

**Source files moved upstream**

| original (Mar 2026) | now | moved by |
|---|---|---|
| `api/apps/chunk_app.py` → `retrieval_test` | `api/apps/services/dataset_api_service.py` → `search` **and** `search_datasets` | #14402 |
| `api/apps/sdk/session.py` → `retrieval_test_embedded` | `api/apps/restful_apis/bot_api.py` | #15151 |
| `api/apps/sdk/dify_retrieval.py` → `retrieval` | `api/apps/restful_apis/dify_retrieval_api.py` | #15151 |
| `api/db/services/dialog_service.py` → `async_ask` / `gen_mindmap` | unchanged (content conflict with the new `search_id` parameter) | — |

The old `chunk_app.retrieval_test` split into two siblings that are byte-identical in this respect, so both `search` and `search_datasets` are fixed. That is the only place the diff is wider than the original PR; leaving one of the two on `[]` would have been arbitrary. Everything else is a one-to-one retarget.

**The premise still holds on current main.** `Dealer.get_filters` (`rag/nlp/search.py`) still only adds a `doc_id` condition when `doc_ids is not None`, so `[]` remains an active filter on an empty document set. ES / OpenSearch / Infinity / OceanBase happen to skip falsy values in their condition builders, but GaussDB's `build_condition_where` does not — it emits `doc_id IN ()`. `sdk/doc.py` still uses the `None` pattern this PR aligns with, and the explicit metadata no-match case keeps its `["-999"]` sentinel.

**Tests.** The old branch had a leftover conflict marker committed into `test_dialog_service_use_sql_source_columns.py`, which is presumably part of why CI was red; that is gone. Coverage was rewritten against the current layout, since two of the original test homes are no longer usable:

- `test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py` — `async_ask` + `gen_mindmap` (extends the existing `test_async_chat_uses_all_docs_when_no_doc_ids_selected` from #13484)
- `test/testcases/test_http_api/test_dataset_management/test_dify_retrieval_routes_unit.py` — `/dify/retrieval`
- `test/unit_test/api/apps/services/test_dataset_api_service_search_doc_ids.py` (new) — `search` + `search_datasets`
- `test/unit_test/api/apps/restful_apis/test_retrieval_test_embedded_doc_ids.py` (new) — `/searchbots/retrieval_test`

Each new assertion was mutation-checked: with the source change reverted, all 9 fail (`assert [] is None`); with it applied, all pass. `ruff check` reports no new findings over the `main` baseline on the touched files.

**Three pre-existing issues I ran into but deliberately left alone** — happy to open separate PRs if useful:

1. `test/unit_test/api/apps/restful_apis/test_agentbots_access_control.py` (4 cross-tenant access-control tests) fails at import on `main`: its stub set predates `bot_api`'s `rag.utils.web_search_conn` and `api.utils.pagination_utils` imports. Two `_stub(...)` lines fix it.
2. `test_dataset_sdk_routes_unit.py` and `test_session_sdk_routes_unit.py` have rotted further (36/37 failing in the latter). They are `p3`/excluded at the default `--level p2`, so this goes unnoticed.
3. `test_dify_retrieval_routes_unit.py` exists at two paths with the same basename (`test/testcases/restful_api/` and `test/testcases/test_http_api/test_dataset_management/`), so pytest refuses to collect both in one run. I added the new case only to the second, which is the fuller copy.

Separately: `agent/tools/retrieval.py` initialises `doc_ids = []` and passes it straight through when no metadata filter is configured — the same latent bug, but it's an agent tool rather than a retrieval API, so it's outside this PR's stated scope.

JosefAschauer:
Follow-up after a second review pass. Two things: a correction to my previous comment, and a fix for a regression that review found in this PR's own new code.

**Correction.** I wrote above that GaussDB "does not skip falsy values — it emits `doc_id IN ()`". That is wrong, and I should have followed the call through. `build_condition_where` delegates to `_list_values`, which raises:

```python
def _list_values(self, value) -> list[Any]:
    values = list(value) if isinstance(value, (list, tuple, set)) else [value]
    if not values:
        raise ValueError("empty condition values are not supported")
    return values
```

So the accurate picture for `doc_ids=[]` is: ES, OpenSearch, Infinity, OceanBase and SereneDB all `continue` past falsy condition values, meaning the doc filter is **silently dropped** and the query widens to the whole dataset; GaussDB **raises**. Different failure modes, same conclusion — `[]` is never a usable document filter — but the earlier wording was inaccurate.

**Regression in the previous revision.** The blank-id stripping I added, `[d for d in doc_ids if d] if doc_ids else None`, breaks the PR's own invariant: `["", ""]` is truthy, so the comprehension yields `[]` — falsy, but not `None`.

`POST /api/v1/datasets/{id}/search` with `{"question":"x","doc_ids":["",""]}`:
- before this PR: `["", ""]` → `doc_id IN ('','')` → empty result set
- previous revision: `[]` → filter dropped on five backends, so the request searches **every document in the dataset**; on GaussDB, a 500

Scope-widening is the worse of the two, and it was introduced here. Fixed.

**What changed**

The normalization is now one function rather than the same expression repeated in two spellings:

```python
def normalize_doc_id_filter(doc_ids) -> list[str] | None:
    if not doc_ids:
        return None
    return [doc_id for doc_id in doc_ids if doc_id] or None
```

It lives in `common/metadata_utils.py`, next to `apply_meta_data_filter`, and is used at every entrypoint that builds a document filter — the six from before, plus `async_chat` and `rag_agent`.

Those last two are new to the PR and deserve a word. They already default an *absent* `doc_ids` to `None` (from #13484), so they were correct for the case this PR is named after. But they carried the identical all-blank defect: `messages[-1]["doc_ids"] = ["", ""]` produced `[]`, which `RAGTools` stores verbatim (`agentic_rag.py:292` only special-cases `None`) and forwards to the retriever. Both are in `dialog_service.py`, a file this PR already edits, a few dozen lines from sites I was fixing — leaving the same bug in the two most-used chat paths was not defensible. Happy to split them into their own PR if you would rather keep this one narrow.

**Verification**

- Every new assertion is mutation-tested. Reverting `or None` fails 7; reverting the two chat call sites to their original expressions fails 3 more.
- `test/unit_test/common/test_normalize_doc_id_filter.py` pins the helper directly, including the invariant that it never returns `[]`.
- The `if meta_data_filter:` guard is now covered by a test rather than only by reading. It matters: `apply_meta_data_filter` opens with `list(base_doc_ids) if base_doc_ids else []`, so reaching it unguarded would quietly turn a `None` back into `[]` and make the fix a no-op.
- The three test loaders that stub `common.metadata_utils` are wired to the **real** helper, not a double, so the call-site tests exercise real normalization.
- `test/unit_test/api` + `test/unit_test/common`: 1025 passed. The only failures are two pre-existing ones in `test_dialog_service_rag_agent_image_routing.py` (a `SimpleNamespace` dialog stub missing `similarity_threshold`), which fail identically on unmodified `main`.
- `ruff check`: 282 findings on the changed files before and after — no new lint. `ruff format --check`: clean.

**Also in this revision:** `test/unit_test/api/apps/restful_apis/test_agentbots_access_control.py` is repaired. Its stub set predated `bot_api`'s `rag.utils.web_search_conn` and `api.utils.pagination_utils` imports, so all four cross-tenant access-control tests were failing at import on `main`. I had left this alone as unrelated, but adding `normalize_doc_id_filter` to `bot_api` meant touching that file's stubs anyway, so it is fixed properly — three stub lines, four tests live again.

One more pre-existing item, purely FYI, no action taken: `remove_dataset_nav_doc_sync` (`rag/advanced_rag/knowlege_compile/dataset_nav.py:1857`) closes its event loop in `finally` even when `run_until_complete` raised, leaving a suspended coroutine that surfaces later as `PytestUnraisableExceptionWarning: Exception ignored in: <coroutine object remove_dataset_nav_doc>`. Because it lands on whichever test happens to run next, adding any test file moves it; on this branch it lands on one of the new files. It is not caused by this PR — on unmodified `main` it lands on `test_dify_retrieval.py::TestDifyRetrievalTenantCheck` instead.

JosefAschauer:
Addressed, with one part of the finding adjusted after checking it against the code — plus a correction to something I claimed in my previous comment.

**Debug logging — done.** Counts only, never id values:

```python
kept = [doc_id for doc_id in doc_ids if doc_id]
dropped = len(doc_ids) - len(kept)
if dropped:
    # Counts only — document ids are caller data and never logged.
    logging.debug("normalize_doc_id_filter dropped %d empty id(s) of %d requested", dropped, len(doc_ids))
```

Pinned by two tests: one asserts the count message is emitted and that the ids themselves (`secret-doc-1`, …) appear nowhere in the log records; another asserts nothing is logged when nothing is dropped. Mutation-checked — adding the id list to the log statement fails the first test.

**Whitespace-only inputs — covered by tests, but the behaviour is deliberately "keep", not "drop".** The finding assumes whitespace-only ids fall under the blank-filter contract. They do not, and should not:

```
normalize_doc_id_filter(["  "])          -> ["  "]     # kept
normalize_doc_id_filter(["", ""])        -> None       # dropped
normalize_doc_id_filter(["doc-1", "  "]) -> ["doc-1", "  "]
```

`"  "` is truthy, so it survives `if doc_id`. That is the right outcome. Empty ids are dropped because they are an artifact of splitting a comma-separated field — `"doc-1,".split(",")` yields `["doc-1", ""]` — so they were never a selection. A whitespace-only id *is* a selection; it just matches no document, and the request correctly returns nothing. Dropping it would collapse the filter to `None` and turn a request scoped to one document into a search across the entire dataset, which is the failure mode the previous review round was about.

Ids are also never rewritten, so `" doc-2"` is forwarded exactly as supplied rather than silently trimmed to `"doc-2"`.

Six parametrized cases now pin this (`spaces`, `tab`, `newline`, `mixed-with-real`, `whitespace-and-empty`, `padded-id`), and the docstring states the empty-vs-whitespace distinction so the next reader does not have to re-derive it.

---

**Correction to my previous comment.** I wrote that the earlier revision introduced scope-widening and that this was "Fixed". That overstates it. Verified against `Dealer.get_filters`:

| `doc_ids` | condition built | effect |
|---|---|---|
| `None` | no `doc_id` key | whole dataset |
| `[]` | `doc_id: []` | dropped by ES/OpenSearch/Infinity/OceanBase/SereneDB → whole dataset; GaussDB raises |
| `["  "]` | `doc_id: ["  "]` | zero results |

For all-empty input, `None` widens exactly as `[]` did on those five backends. What the change actually fixes is the GaussDB `ValueError` and the inconsistency of expressing "no filter" as `[]` in some places and `None` in others — not the widening.

That widening is intentional and matches what `async_chat` already did before this PR: `[d for d in "".split(",") if d]` produced `[]`, which those backends drop. So an all-empty selection has always been treated as "no selection". If you would rather it return nothing instead, the codebase already has an idiom for that — the `["-999"]` sentinel in `dify_retrieval_api` — and I am happy to switch to it; it is a one-line change in the helper. I did not make that call unilaterally because it changes behaviour for every caller.

**Verification:** `test/unit_test/api` + `test/unit_test/common` → 1033 passed. The only failures are the two pre-existing ones in `test_dialog_service_rag_agent_image_routing.py` that fail identically on unmodified `main`. `ruff check` on the changed files: 29 findings before and after in `common/metadata_utils.py`, and the new test file is clean. `ruff format --check`: clean.

PR #13622, opened 2026-03-16
fix : build error

6ba3i:

add timeout to fix fail at build during uvsync step

### Type of change


JinHai-CN:
```
    /// Timeout (in seconds) for HTTP requests. (default: 30 s)
    pub const UV_HTTP_TIMEOUT: &'static str = "UV_HTTP_TIMEOUT";

    /// The number of retries for HTTP requests. (default: 3)
    pub const UV_HTTP_RETRIES: &'static str = "UV_HTTP_RETRIES";
```

Given your configuration (UV_HTTP_RETRIES=5, UV_HTTP_TIMEOUT=300s), the theoretical maximum time spent on HTTP requests would be 5 retries × 300s = 1500s (25 minutes).

PR #13638, opened 2026-03-16
feat: add yaml file extension support for parsing

SFL79:

RAGFlow declines .yaml extension files and only accepts .yml files

### Type of change

### DIFF
--- a/api/utils/file_utils.py
+++ b/api/utils/file_utils.py
@@ -64,7 +64,7 @@ def filename_type(filename):
     if re.match(r".*\.pdf$", filename):
         return FileType.PDF.value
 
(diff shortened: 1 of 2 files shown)


SFL79:
This PR ends up fixing 2 issues:
1) ragflow declines yaml but accepts yml in _upload_ stage.
2) ragflow didnt have a matching parser for yml/yaml extensions so even a yml file ended up being unable to get parsed. I added yaml and yml files to the txtParser. Also, made sure the error ragflow throws is up to date with all file extensions checked in the code block


Im actually debating on this with myself as yaml might benefit more from a different kind of chunking, however it is better than the existing condition where there's none. Let me know what you think

JinHai-CN:
@SFL79 would you please resolve the conflicts?

PR #13687, opened 2026-03-19
fix: include missing modules in ragflow-cli package

### LINKED ISSUE #13456: [Bug] ragflow-cli PyPI package missing http_client.py and other required files
## Description

The `ragflow-cli` package on PyPI (version 0.24.0) is missing critical Python modules that are required for the CLI to function.

## Steps to reproduce

1. Install ragflow-cli from PyPI: `pip install ragflow-cli==0.24.0`
2. Try to run the CLI: `ragflow-cli -h 127.0.0.1 -p 9381`
3. Observe the import errors

## Expected behavior

The CLI should work after installing from PyPI.

## Actual behavior

The CLI fails with import errors because the following files are missing from the PyPI package:

- `http_client.py`
- `ragflow_client.py`  
- `user.py`

These files exist in the GitHub repository at `admin/client/` but were not included in the PyPI distribution.

## Files in PyPI package (ragflow_cli-0.24.0-py3-none-any.whl):
- parser.py
- ragflow_cli.py
- ragflow_cli-0.24.0.dist-info/

## Files in GitHub repository (admin/client/):
- parser.py
- ragflow_cli.py
- http_client.py    # MISSING from PyPI
- ragflow_client.py  # MISSING from PyPI
- user.py            # MISSING from PyPI

## Workaround

Run from source code:
```bash
git clone https://github.com/infiniflow/ragflow.git
cd ragflow
pip install -r admin/client/requirements.txt
python admin/client/ragflow_cli.py -h 127.0.0.1 -p 9381
```

## Suggested fix

Update the PyPI packaging/publishing process to include all files from `admin/client/` directory in the distribution.

junagent:
## Summary
This PR fixes the PyPI packaging issue where the ragflow-cli package is missing critical Python modules after installation.

## Problem
The ragflow-cli package on PyPI is missing:
- `http_client.py`
- `ragflow_client.py`
- `user.py`

Users installing via `pip install ragflow-cli` get only the entry point without the client libraries.

## Root Cause
The pyproject.toml correctly lists all modules in `py-modules`, but the package distribution (sdist/wheel) doesn't include all Python files from the admin/client/ directory.

## Solution
Added `MANIFEST.in` to ensure all `*.py` files in admin/client/ are included in the package distribution:

```
include admin/client/*.py
```

## Testing
The fix ensures that all modules (ragflow_cli, parser, http_client, ragflow_client, user) are included when the package is built and distributed.

Fixes #13456

### DIFF
--- a/admin/client/MANIFEST.in
+++ b/admin/client/MANIFEST.in
@@ -1,0 +1,1 @@
+include admin/client/*.py


junagent:
This fix follows the same Python packaging verification pattern that was just confirmed in aws/sagemaker-python-sdk#5647. Both issues fall into the same root cause category:

**Packaging discovery gap**:  specifies package contents correctly, but the built distribution (wheel/sdist) is missing files because build tooling (, , or ) falls back to a different file selection mechanism than what was tested locally.

The  approach here is correct because it ensures files are included in the sdist even when  metadata alone is insufficient for the build backend being used. The wheel should be unaffected since wheel building is driven by , but the sdist was broken.

Verification steps:
1. Build sdist: 
2. Inspect contents: 
3. Verify modules present before publishing

Similar confirmed cases: aws/sagemaker-python-sdk#5647 (setuptools namespaces discovery), infiniflow/ragflow#13456

PR #13697, opened 2026-03-19
Fix: unify rename docstore title tokens with indexing semantics

yongtenglei:

Unify rename docstore title tokens with indexing semantics.

### Type of change


KevinHuSh:
Appreciations!
Outdated.

PR #13704, opened 2026-03-19
Fix: allow on

Lynn-Inf:

Allow input on/ON as status.

### Type of change


JinHai-CN:
What about 'OFF'?

Issue #13731, opened 2026-03-21
[Bug]: Doc Metdata no longer working in UI or API in nightlies


metalshanked:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

latest

### RAGFlow image version

nightly

### Other environment information

```Markdown

```

### Actual behavior

There is no longer any "add metadata" button in the Files page for a dataset.
As a result, there is no way to manually add custom metadata to dataset/documents

also, the API below does not work (using author : alice example from official doc)

NOTE: This works fine in v0.24.0. The issue started after that in the nightlies...

```
POST ... /metadata/update
```
```
{
    "code": 100,
    "data": null,
    "message": "UnboundLocalError(\"cannot access local variable 'target_doc_ids' where it is not associated with a value\")"
}
```

### Steps to reproduce

```Markdown
1. Upload a file 
2. Click the metadata column
3. No Add button anymore in the metadata modal box
```

metalshanked:
@dosu can you check the nightly code and confirm this issue?

metalshanked:
@dosu That is not correct. Issue still exists in both UI and API (see screenshot)
Please check the latest nightly code to confirm the issue


<img width="841" height="494" alt="Image" src="https://github.com/user-attachments/assets/c5982a3a-00bd-4a9f-b839-6067f9a6b9b3" />

metalshanked:
@dosu That is not correct. The issue still exists in both UI and API (see screenshot)
Please check the latest nightly code to confirm the issue

metalshanked:
@yingfeng @Magicbook1108 
Can you please advise on the fix or workaround?  Seems like metadata function is lost currently
Thanks!

Magicbook1108:
We are working on it.

PR #13760, opened 2026-03-24
Fix: Metadata,chunk,dataset Related bugs

dcc123456:

Fix: Metadata,chunk,dataset Related bugs
- metadata not show add button #13731
- chunk edit question style
- dataset modified chunk method bug
### Type of change


adagoral:
@dcc123456  seems like it doesn't show existing metadate?

PR #13793, opened 2026-03-25
feat: add centralized retry mechanism for LLM providers 

tdakanalis:

This PR solves the problem of duplicated and inconsistent retry logic across LLM model implementations.

#### Background:  
Previously, each LLM model class (embedding, CV, rerank, TTS, OCR, speech2text) had its own ad-hoc retry implementation, leading to:                                                                                          
 - Duplicated code across 10+ files   
 - Inconsistent retry behavior between different model providers 
 - Difficult maintenance when changing retry settings  
 - Risk of bugs when retry logic wasn't applied uniformly                                                                                                                                       
 
#### Solution: 

- Extract retry logic into a centralized rag/llm/retry.py module with:
     - @ retry decorator for sync methods (retries on transient errors, re-raises on failure)
     - @ retry_or_fallback decorator for methods that return error tuples   
     - is_retryable() - determines if an error is transient (rate limits, server errors)
     - classify_error() - categorizes exceptions into LLMErrorCode for detailed logging 
     - async_handle_exception() - shared async error handler 
- Apply decorators consistently across all embedding, CV, rerank, TTS, OCR, and speech2text models                                                                                                                              
- Add comprehensive unit tests (76 tests) 

#### Behavior preserved: 

- Same default retry settings (max_retries=5, base_delay=2.0) 
- Same retryable error signals (rate limits, 5xx server errors) 
- Same backoff jitter formula (base_delay * uniform(10, 150))  

### Type of change

- [x] Refactoring


xugangqiang:
Really very very impressive ideas & implementations!

We really very appreicate your contribution on this.

Your PR enlights us a lot and could be very helpful.

Given the size of this PR, we need more time to review - if the change only touches the syntax but does not change the actual logic. Thus please expect some delay for the merge. 

Thanks.

rag/llm/retry.py
```diff
@@ -0,0 +1,259 @@
...
+    All other errors are considered non-retryable.
+
+    Args:
+        error: The exception to evaluate.
+
+    Returns:
+        True if the error is transient and should be retried.
+    """
+    msg = str(error).lower()
+    return any(signal in msg for signal in _RETRYABLE_RATE_LIMIT_SIGNALS) or any(signal in msg for signal in _RETRYABLE_SERVER_SIGNALS)
```
xugangqiang: Do you think ERROR_CONNECTION is retryable? Like temp network jitter?

tdakanalis: Yes, I think it should be.

rag/flow/parser/parser.py
```diff
@@ -1029,7 +1033,11 @@ def _audio(self, name, blob, **kwargs):
...
             seq2txt_model_config = get_model_config_by_type_and_name(self._canvas.get_tenant_id(), LLMType.SPEECH2TEXT, conf["llm_id"])
             seq2txt_mdl = LLMBundle(self._canvas.get_tenant_id(), seq2txt_model_config)
-            txt = seq2txt_mdl.transcription(tmp_path)
+            try:
+                txt = seq2txt_mdl.transcription(tmp_path)
+            except Exception as e:
+                logging.warning(f"Transcription failed: {e}")
+                txt = ""
 
             self.set_output("text", txt)
```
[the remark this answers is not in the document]
xugangqiang: @tdakanalis 
Do you think the comments from coderabbitai is valid?

-----------------
Normalize transcription return before setting "text" output.

Line 1037 can return a tuple (text, token_count) (per rag/llm/sequence2txt_model.py:40-45), but the exception path sets a string. This makes output type unstable at Line 1042.

💡 Proposed fix
             seq2txt_mdl = LLMBundle(self._canvas.get_tenant_id(), seq2txt_model_config)
             try:
-                txt = seq2txt_mdl.transcription(tmp_path)
+                transcription = seq2txt_mdl.transcription(tmp_path)
+                txt = transcription[0] if isinstance(transcription, tuple) else transcription
             except Exception as e:
                 logging.warning(f"Transcription failed: {e}")
                 txt = ""

             self.set_output("text", txt)

rag/llm/cv_model.py
```diff
@@ -723,31 +727,27 @@ def _form_history(self, system, history, images=None):
...
-            return ans, 128
-        except Exception as e:
-            return "**ERROR**: " + str(e), 0
+        response = self.client.generate(
+            model=self.model_name,
+            prompt=vision_prompt[0]["content"],
+            images=[image],
+        )
+        ans = response["response"].strip()
+        return ans, 128
```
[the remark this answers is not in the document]
xugangqiang: @tdakanalis 
Do you think this is a valid concern?

rag/llm/retry.py
tdakanalis: @xugangqiang if i need to do anything on my side for this PR please let me know.

xugangqiang: @tdakanalis 
I think I've added some review comments. Please let me know your thoughts. Thank you.

tdakanalis: @xugangqiang Regarding whether the ERROR_CONNECTION must be retriable or not, I think that it makes sense and it is better to be. I tried not to change the logic already in place and to focus only on the refactoring. If you agree, I can prepare a commit to also add the ERROR_CONNECTION in the list of the retriable errors.

rag/flow/parser/parser.py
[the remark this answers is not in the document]
tdakanalis: I will check later today and come back

rag/llm/retry.py
xugangqiang: @tdakanalis 
I am fine with your proposal and this is not a show-stopper for this PR.

rag/llm/cv_model.py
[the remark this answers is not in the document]
tdakanalis: Valid catch. Fixed at the caller (LLMBundle.describe / describe_with_prompt in llm_service.py) with a try/except returning "**ERROR**: ...", so the model-layer contract stays uniform across providers.

rag/llm/retry.py
tdakanalis: Done.

rag/llm/retry.py
```diff
@@ -0,0 +1,259 @@
...
+    logging.exception(f"{log_prefix} async completion")
+    error_code = classify_error(error)
+    if attempt == max_retries:
+        error_code = LLMErrorCode.ERROR_MAX_RETRIES
+
+    if is_retryable(error):
+        delay = get_retry_delay(base_delay)
+        logging.warning(f"Error: {error_code}. Retrying in {delay:.2f}s (attempt {attempt + 1}/{max_retries})")
+        await asyncio.sleep(delay)
+        return None
```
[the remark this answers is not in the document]
tdakanalis: I also fixed this one. On the final attempt with a retryable error, the old code slept and returned None, leaving callers without a final error. Fixed by short-circuiting when attempt == max_retries to return an ERROR_MAX_RETRIES message before the retryable/sleep path.

rag/llm/sequence2txt_model.py
xugangqiang, line 250: May I ask for your help on this? Possible to use the new retry framework?

rag/llm/sequence2txt_model.py
```diff
@@ -398,7 +393,8 @@ def __init__(self, key, model_name, base_url=None, lang="English", **kwargs):
         self.lang = lang
         
         self.client = OpenAI(api_key=key, base_url=self.base_url)
-    
+
+    @retry
```
xugangqiang: Can you check the decorator XinferenceSeq2txt/QWenSeq2txt ?

I think XinferenceSeq2txt/QWenSeq2txt/RAGconSeq2txt/... may should align?

class RAGconSeq2txt(Base):
    @retry_or_fallback(lambda e: (f"**ERROR**: {e}", 0))
    def transcription(self, audio_path, **kwargs):

rag/flow/parser/parser.py
```diff
@@ -1055,10 +1055,17 @@ def _image(self, name, blob, **kwargs):
...
             system_prompt = conf.get("system_prompt")
-            if system_prompt:
-                txt = cv_model.describe_with_prompt(img_binary.read(), system_prompt)
-            else:
-                txt = cv_model.describe(img_binary.read())
+            try:
+                if system_prompt:
+                    txt = cv_model.describe_with_prompt(img_binary.read(), system_prompt)
+                else:
+                    txt = cv_model.describe(img_binary.read())
```
xugangqiang, line 1082: would like to confirm, as caller for describe_with_prompt, how do you expect the caller to handle the result?
1. check if it's starting with "ERROR**"? If so, can we have a unified function like "is_result_valid" and have constants as prefix?
2. check if there's exception raised? if exception raised, assign a default value to txt

Just want confirm my understanding.

rag/llm/retry.py
```diff
@@ -0,0 +1,276 @@
...
+    Args:
+        base_delay: The base delay in seconds.
+
+    Returns:
+        The delay in seconds with random jitter applied.
+    """
+    return base_delay * random.uniform(_MIN_JITTER_MULTIPLIER, _MAX_JITTER_MULTIPLIER)
+
+
+def retry(method: Callable[..., _T]) -> Callable[..., _T]:
```
xugangqiang, line 186: Can we clearly clarify whether "generator" can be supported?

rag/llm/retry.py
```diff
@@ -0,0 +1,276 @@
...
+    def wrapper(self, *args, **kwargs):
+        max_retries = getattr(self, "max_retries", int(os.environ.get("LLM_MAX_RETRIES", 5)))
+        base_delay = getattr(self, "base_delay", float(os.environ.get("LLM_BASE_DELAY", 2.0)))
+        for attempt in range(max_retries + 1):
+            try:
+                return method(self, *args, **kwargs)
+            except Exception as e:
+                if attempt == max_retries or not is_retryable(e):
+                    raise
+                delay = get_retry_delay(base_delay)
```
xugangqiang: what's the min retry delay if no setting for the base_delay (from env)?
If possible, I think 1-3 seconds for the first retry will be idea.

rag/llm/sequence2txt_model.py
tdakanalis: Retry handles failures you hope go away (exception-driven, jittered backoff to spare a struggling server); polling handles success that hasn't completed yet (field-driven, fixed interval on a healthy server). So I'd just wrap the DescribeTaskStatus call with @retry for transient blips and leave the poll loop alone.


```python
@retry
def _describe_task_status(self, req):
  return self.client.DescribeTaskStatus(req)    

...

while retries < max_retries:                                                                                                                                                                                               
  resp = self._describe_task_status(req)   # survives transient 5xx/429/DNS blips 

```

rag/llm/sequence2txt_model.py
tdakanalis: You are right. I will update them

tdakanalis: done

rag/flow/parser/parser.py
tdakanalis: ​1. Yes on the prefix check: LLMBundle.describe_with_prompt always returns a str and encodes failures as "**ERROR**: " + str(e). Good call on the unified helper - rag/llm/retry.py  already exposes ERROR_PREFIX and - now - is_error_result, so I've swapped parser.py:1066 to use is_error_result(txt).  

2. The try/except at parser.py:1059-1066 isn't strictly needed for the LLM call. LLMBundle catches exceptions internally and never raises them. It still covers local surprises like img_binary.read() failing.

rag/llm/retry.py
tdakanalis: ​Good catch! Generators aren't supported, and the decorator is silently a no-op on them (the try/except in retry.wrapper only guards the call that returns the generator, never iteration). Worse, restarting a generator after a mid-stream failure would re-yield already-emitted chunks, which isn't safe for streaming consumers.

​I've made the limitation explicit:
- Added an inspect.isgeneratorfunction guard in both retry and retry_or_fallback that raises TypeError at decoration time, with a note in the docstrings
- Removed the four existing silent no-ops in rag/llm/tts_model.py (FishAudioTTS.tts, QwenTTS.tts, OpenAITTS.tts, RAGconTTS.tts) - behaviour-preserving, since each already raises on non-200/transient errors

​Proper streaming retry is a separate design effort and out of scope for now.

rag/llm/retry.py
tdakanalis: In the previous implementation - that i tried not to change during the refactoring - every retry waited somewhere between 20 and 300 seconds, no matter which attempt it was. The docstring claimed "exponential backoff" but it wasn't.

In my latest update I just committed, I switched to proper exponential backoff with additive jitter, capped at 60 s. The progression now is: 1-2 s -> 2-3 s -> 4-5 s -> 8-9 s -> 16-17 s -> 32-33 s -> 60 s (cap)

api/db/services/llm_service.py
```diff
@@ -151,7 +151,11 @@ def describe(self, image, max_tokens=300):
         if self.langfuse:
             generation = self.langfuse.start_generation(trace_context=self.trace_context, name="describe", metadata={"model": self.model_config["llm_name"]})
 
-        txt, used_tokens = self.mdl.describe(image)
+        try:
+            txt, used_tokens = self.mdl.describe(image)
+        except Exception as e:
+            logging.exception("LLMBundle.describe failed")
```
xugangqiang, line 158: Do we need below in exception handling block?

if self.langfuse:
        generation.update(output={"error": str(e)})
        generation.end()

xugangqiang:
@yingfeng @JinHai-CN @yuzhichang 
LGTM if:
1. langfuse related handling added
2. conflicts resolved and passed CI

xugangqiang:
LGTM from my perspective.
@JinHai-CN @yingfeng  Please have a second eye check since this is a relative big size change and it's a sort of infra change.

yuzhichang (changes requested):
Thanks for the work here. I do not think this PR should be merged in its current form.

The main idea still makes sense: centralizing retry and error-classification logic would reduce duplication in the LLM layer. However, this PR is now quite old, and `main` has changed substantially in the same area since then.

At this point, the branch is no longer a good fit for the current codebase:
- parts of the motivation have already been addressed in other ways on `main` (for example, more structured model error handling in the embedding path);
- the PR still relies heavily on the string-based `**ERROR**` flow, while the codebase has already started moving toward more structured exception handling;
- it touches several fast-moving provider/model files, so merging it now would likely reintroduce outdated assumptions and create a high conflict/regression risk.

I think the remaining value here is better extracted as a few new, smaller PRs against current `main`, for example:
- deduplicate the retry/error classification logic that still exists in `rag/llm/chat_model.py`;
- add targeted tests for retry classification/backoff behavior;
- port any small safe fixes separately.

Please rework this as fresh, smaller changes on top of current `main` rather than reviving this PR as-is.

yuzhichang:
One more consideration: the current development focus has largely shifted to the Go implementation, so continuing to revive and maintain this Python-side retry refactor now likely has limited return compared with landing equivalent improvements in the actively evolving Go path or in smaller, targeted fixes on current main.

Issue #13795, opened 2026-03-26
[Bug]: WebDAV sync does not filter unsupported files before processing


pennbay:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

d19ca71b4

### RAGFlow image version

nightly

### Other environment information

```Markdown
- Hardware parameters:
- OS type: Linux
- Others:
  - Data source: WebDAV
  - Sync mode: WebDAV sync / polling
```

### Actual behavior

The WebDAV connector does not filter unsupported file types early in the sync flow.

During recursive listing, files are selected mainly by modification time and can still enter the later download / document-generation path even when their extensions are not accepted by RAGFlow.

This makes WebDAV behavior inconsistent with other connectors and can cause unnecessary processing of unsupported files.

### Expected behavior

The WebDAV connector should filter unsupported file types before download, using the same accepted-extension rules as other connectors.

Ideally:
1. Unsupported files should be skipped during recursive listing.
2. A defensive extension check should also exist before download / document creation.

### Steps to reproduce

```Markdown
1. Configure a WebDAV data source.
2. Put both supported and unsupported files under the configured remote path.
   - Example supported files: `.pdf`, `.txt`, `.md`
   - Example unsupported files: `.exe`, `.bin`, `.dat`
3. Trigger a full sync or polling sync.
4. Observe that unsupported files are not filtered out early by the WebDAV connector and can still enter the later processing path.
```

### Additional information

The relevant code appears to be in [common/data_source/webdav_connector.py](/home/ubuntu/ragflow/common/data_source/webdav_connector.py).

From inspection, the WebDAV connector is missing the accepted file-extension filtering that should happen before downloading files. A straightforward fix would be to apply accepted-extension checks during recursive listing and keep a second defensive check before document creation.

Magicbook1108:
A pr for file type validation is welcomed.

spider-yamet:
@Magicbook1108 @yingfeng Could you please review the PR?

[...]

PR #13802, opened 2026-03-26
Add file type validation

spider-yamet:

This PR fixes WebDAV sync behavior for unsupported file types ([#13795](https://github.com/infiniflow/ragflow/issues/13795)).

Previously, the WebDAV connector selected files primarily by modified time (and size threshold) and could still pass unsupported extensions into the download/document-generation path. This caused unnecessary processing and inconsistent behavior compared with connectors that validate file type earlier.

This change adds extension validation in two places:

1. **Early filter during recursive listing** to skip unsupported files before they enter the download flow.
2. **Defensive filter before download/document creation** to prevent unsupported files from being processed if any listing edge case slips through.

It also wires `allow_images` into the WebDAV sync path so image extension handling follows connector policy.

Scope is intentionally limited to WebDAV for a focused bug-fix PR.

### Type of change

### How was this tested?

- Manual verification with mixed file types under the configured WebDAV path:
  - supported: `.pdf`, `.txt`, `.md`
  - unsupported: `.exe`, `.bin`, `.dat`
- Triggered full sync and polling sync.
- Confirmed unsupported files are skipped before download.
- Confirmed supported files are still indexed normally.
- Confirmed image handling follows `allow_images` setting.

Fixes: #13795


spider-yamet:
@yingfeng Would love to hear your opinion on this PR. Thanks

spider-yamet:
Would appreciate your feedback @Magicbook1108 @yingfeng :)

spider-yamet:
@Magicbook1108 @yingfeng Would appreciate your feedback, hope you can check once this pr passes the ci.

PR #13835, opened 2026-03-28
Fix: streaming response reference mismatch with answer

### LINKED ISSUE #13828: [Bug]: /api/v1/chats/{chat_id}/completions stream returns wrong references, but chat history is correct
### Self Checks

Unknown

### RAGFlow image version

none

```Markdown

```

### Actual behavior

- Streaming response: references are mismatched
- Chat history response: references are correct

### Expected behavior

The reference in the streaming response should be the same as the one in the chat history.

### Steps to reproduce

```Markdown
1. Create a chat assistant and enable reference citation.
2. Call POST `/api/v1/chats/{chat_id}/completions` with stream=true.
3. Check the reference in the streaming result — it does not match the answer.
4. Get the chat history and check the reference — it is correct.
5. The issue can be stably reproduced.
```

### Additional information

_No response_

Lntanohuang:
## Summary

Fixes #13828 — streaming response references are mismatched with the answer, while chat history references are correct.

### Root Cause

In `async_chat()` (`api/db/services/dialog_service.py`), the streaming code path:

1. Sends intermediate SSE events with `reference: {}` (empty) — expected, since references are computed after the full answer is available.
2. After streaming completes, calls `decorate_answer()` which:
   - Inserts citation markers (`##N$$`) into the answer text
   - Filters `doc_aggs` based on which chunks were actually cited
   - Returns the decorated answer with matching references
3. **Bug**: The final SSE event then blanks the answer (`final["answer"] = ""`), discarding the citation-processed text while keeping the citation-filtered references.

This means the client accumulates raw answer text (no citation markers) from delta events, but receives references filtered based on citation analysis it never sees — causing the mismatch reported in the issue.

Chat history is unaffected because `structure_answer()` correctly persists `conv.reference[-1]` from the final event's reference data.

### Fix

Remove `final["answer"] = ""` so the final SSE event includes the full decorated answer (with citation markers) alongside its matching references. This is consistent with the non-streaming code path behavior.

### Behavior Change

Previously, the final streaming SSE event had `answer: ""`. After this fix, it contains the complete citation-processed answer. Clients that accumulate delta text should use the final event's answer as the authoritative text (replacing accumulated deltas), which is the same pattern used in the non-streaming response.

## Test plan

- [ ] Call `POST /api/v1/chats/{chat_id}/completions` with `stream=true` and citation enabled
- [ ] Verify the final SSE event contains both the decorated answer and matching references
- [ ] Verify the references in the streaming response match those in the chat history
- [ ] Verify non-streaming (`stream=false`) behavior is unchanged

---

**Update:** Also fixes the same issue in `async_ask()`. The `infinity_conn.py` escaping fix has been moved to a separate PR (#14053) to keep this one focused on the streaming reference mismatch.

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -765,7 +765,6 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
             final = decorate_answer(thought + full_answer)
             final["final"] = True
             final["audio_binary"] = None
-            final["answer"] = ""
             yield final
     else:
         if llm_type == "chat":
@@ -1435,7 +1434,6 @@ async def async_ask(question, kb_ids, tenant_id, chat_llm_name=None, search_conf
     full_answer = last_state.full_text if last_state else ""
     final = decorate_answer(full_answer)
     final["final"] = True
-    final["answer"] = ""
     yield final
 
 


xugangqiang:
Appreciate your help on this. 

Could you please test on your local environment and then verify your fix?

If testing evidence could be uploaded, that will be of great help to fix/close this bug.

Lntanohuang:
Thanks for the review! I've updated this PR with the following changes:

1. **Extended the fix to `async_ask()`** — it had the identical `final["answer"] = ""` issue, so now both `async_chat()` and `async_ask()` streaming paths are covered.

2. **Separated the `infinity_conn.py` fix into its own PR** — the single-quote escaping fix was unrelated to the streaming reference mismatch, so I've split it out to keep this PR focused. See #14053 for that fix.

@edenfunf I noticed your PR #13917 addresses the same `async_ask()` issue. I've incorporated that fix here so both code paths are covered together. Feel free to close #13917 if you agree, or let me know if you'd prefer otherwise.

I'll follow up with local testing evidence shortly.

Lntanohuang:
I'll have the local testing evidence uploaded within the next day. Stay tuned!

xugangqiang:
@Lntanohuang    May I know if the testing goes well?

xugangqiang:
https://github.com/infiniflow/ragflow/pull/13917  has been merged which contains the fix as well. Thus close this one.

Lntanohuang:
Sorry for the delay — I've been swamped with work and feeling under the weather these past few days, so I couldn't follow up in time.

Glad to see this got resolved through #13917. Thanks @edenfunf for picking it up and @xugangqiang for coordinating!

PR #13842, opened 2026-03-29
Revert port change to comply with EXPOSE_MYSQL_PORT

sirj0k3r:

This PR reverts the specified port in `conf/service_conf.yaml` to be compliant with the default values on `docker/.env`.

### Type of change


KevinHuSh:
Appreciations!

It's already [3306](https://github.com/infiniflow/ragflow/blob/cc21dc7f007ea753e2c2063922abac6122f7c8fc/docker/.env#L121)

PR #13874, opened 2026-03-31
Feat: track LLM token usage per document during parsing

JulienDeveaux:

Documents parsed with an LLM consume tokens that were never tracked, making it impossible for users or operators to audit LLM costs incurred during ingestion independently of chat usage.

### Type of change


JinHai-CN:
@JulienDeveaux would you please resolve the conficts?

JulienDeveaux:
@JinHai-CN conflicts resolved

PR #13895, opened 2026-04-02
Fix: reject empty Bearer tokens in MCP headers

eviaaaaa:

  This PR fixes the MCP header validation logic in `common/mcp_tool_call_conn.py`.

  Previously, the code used `nv.strip().strip("Bearer")` to decide whether a header value should be included. This is not a correct way to validate Bearer authentication because `strip("Bearer")` removes a character  
  set rather than a literal prefix, which can lead to incorrect filtering behavior.

  This PR replaces that logic with explicit validation:
  - reject empty header names and values
  - reject `Authorization: Bearer` headers with an empty token
  - keep valid `Authorization: Bearer <token>` headers unchanged
  - apply the same validation to both configured MCP headers and runtime custom headers

  This keeps the original intent while making the behavior explicit and correct.

  ### Type of change

  - [ ] Documentation Update
  - [ ] Refactoring
  - [ ] Performance Improvement

### DIFF
--- a/common/mcp_tool_call_conn.py
+++ b/common/mcp_tool_call_conn.py
@@ -57,6 +57,17 @@ class MCPToolCallSession(ToolCallSession):
 
         asyncio.run_coroutine_threadsafe(self._mcp_server_loop(), self._event_loop)
 
+    @staticmethod
+    def _is_valid_header_value(header_name: str, header_value: str) -> bool:
+        name = (header_name or "").strip()
+        value = (header_value or "").strip()
+        if not name or not value:
+            return False
+        if name.lower() == "authorization" and value.lower().startswith("bearer"):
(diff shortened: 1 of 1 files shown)


xugangqiang:
Could you please spend sometime to understand the logic of the method you are trying to change?

Also, could you please tell the logic difference before & after your change? 

Please be noted the header may not only contains "Bearer authentication".

eviaaaaa:
Thanks for the review.
I re-checked the old logic and found that for the empty Bearer scenario I mentioned in the PR, the old logic already filtered it out on the raw_headers path, so my previous justification was not accurate.
The remaining behavior difference is mainly focused on the custom_header path, which is not used in the current business flow. Given this, I've decided to close this PR.
Thanks again.

PR #13917, opened 2026-04-03
fix(dialog): restore decorated answer in async_ask final SSE event

edenfunf:
## What's the problem

Both `async_chat()` and `async_ask()` call `decorate_answer()` to build the final SSE payload — it inserts citation markers (`##N$$`) into the answer text and prunes `doc_aggs` to only the cited documents. Immediately after, both functions overwrite `final["answer"]` with `""`:

```python
# async_chat(), line ~774  (issue #13828)
final = decorate_answer(thought + full_answer)
final["final"] = True
final["audio_binary"] = None
final["answer"] = ""   # discards decorated text
yield final

# async_ask(), line ~1444  (same bug, different path)
final = decorate_answer(full_answer)
final["final"] = True
final["answer"] = ""   # discards decorated text
yield final
```

The client receives filtered references (built for a citation-decorated answer it never sees) while displaying the raw, undecorated streaming text. Citations can never match.

## Root cause

`final["answer"] = ""` was left over from an earlier design where clients were meant to reconstruct the full answer purely from delta events. Once `decorate_answer()` started placing citation markers, this blank-out broke the contract: the final event is where the decorated answer should land.

## Fix

Remove the two blank-override lines — one in `async_chat()`, one in `async_ask()`:

```diff
-    final["answer"] = ""
```

`decorate_answer()` already sets `final["answer"]` to the correct decorated string; there is nothing to override.

## Relation to #13828

Issue #13828 and PR #13835 identify the bug in `async_chat()`. This PR absorbs that fix and also corrects the identical pattern in `async_ask()` (used by the `/retrieval` route in `chat_api.py`), which PR #13835 does not touch.

## Regression test

Added `test/unit_test/api/db/services/test_dialog_service_final_answer.py` with three tests:

| Test | Purpose |
|------|---------|
| `test_buggy_pattern_drops_answer` | Documents the old behaviour: blank-override empties the final answer |
| `test_fixed_pattern_preserves_decorated_answer` | Core invariant: final event carries the decorated text from `decorate_answer()` |
| `test_final_event_reference_matches_decorated_result` | Citation markers in the answer must match the pruned `doc_aggs` in the same event |

Local run result:

```
test_dialog_service_final_answer.py::test_buggy_pattern_drops_answer         PASSED
test_dialog_service_final_answer.py::test_fixed_pattern_preserves_decorated_answer PASSED
test_dialog_service_final_answer.py::test_final_event_reference_matches_decorated_result PASSED

3 passed in 0.04s
```

`ruff check` passes with no issues on all changed files.


edenfunf:
Hi maintainers,

This PR fixes a bug in async_ask where the final SSE event drops the decorated answer.

Root cause:
- final["answer"] is overwritten after decoration, causing loss of citation markers

Fix:
- Preserve the decorated answer in the final event

This is a small, scoped fix with no side effects.

Would really appreciate a review or any feedback. Thanks!

xugangqiang:
Appreciate your help on this. 
Could you please re-produce this issue on your local environment (if possible) and then test with your fix to see if it really works? 
If such testing evidence could be uploaded, I am more than happy to merge this change.

edenfunf:
> Appreciate your help on this. Could you please re-produce this issue on your local environment (if possible) and then test with your fix to see if it really works? If such testing evidence could be uploaded, I am more than happy to merge this change.

Reproduced and verified locally without a full RAGFlow stack.

Test script: test_pr13917.py — it isolates the final-event assembly logic
from async_ask() using minimal stubs that mirror the real decorate_answer() behaviour.

=== BUGGY (before fix) ===
  answer   : ''
  doc_aggs : [{'doc_id': 'doc-A', 'title': 'Doc Alpha'}]
  RESULT: answer is empty — citations lost on client side [BUG CONFIRMED]

=== FIXED (after fix) ===
  answer   : 'RAGFlow supports PDF and DOCX documents. ##0$$'
  doc_aggs : [{'doc_id': 'doc-A', 'title': 'Doc Alpha'}]
  RESULT: answer contains citation marker and doc_aggs is filtered [FIX VERIFIED]

All 3 assertions pass. The one-line removal is the correct fix.

xugangqiang:
Seems similar issue with https://github.com/infiniflow/ragflow/pull/13835 . I am checking with the author of that PR and prefer to merge the two at the same time. 

Thank you for your help on this.

edenfunf:
> Seems similar issue with #13835 . I am checking with the author of that PR and prefer to merge the two at the same time.
> 
> Thank you for your help on this.

ok,Thank!

xugangqiang:
@edenfunf  
May I ask for your help on this issue (https://github.com/infiniflow/ragflow/pull/13835) as well?
Means add fix for that issue in your PR and do regression test to prove the fix works well and no regression issues.
Then please upload the testing evidence . 
Thank you.

edenfunf:
@xugangqiang Thanks for the suggestion! I've updated the PR to also cover the `async_chat()` path from #13828.

**What's in the latest push:**

- Removed `final["answer"] = ""` in `async_chat()` (line ~774) — same one-liner as the `async_ask()` fix already in this PR, so both streaming paths are now consistent.
- Added `test/unit_test/api/db/services/test_dialog_service_final_answer.py` with three regression tests covering the fixed invariant. All pass locally:

```
test_buggy_pattern_drops_answer                       PASSED
test_fixed_pattern_preserves_decorated_answer         PASSED
test_final_event_reference_matches_decorated_result   PASSED

3 passed in 0.04s
```

The tests document both the old behaviour (blank override discards decorated text) and the corrected behaviour (final event carries the `decorate_answer()` result with citation markers intact). `ruff check` is clean on all modified files.

Let me know if there's anything else you'd like adjusted.

xugangqiang:
LGTM

@yingfeng @JinHai-CN @yuzhichang  Please have a second eye check.

xugangqiang:
@edenfunf 

Please fix CI error. 

Looks like you are trying to connect to mysql in unit test which is not good and need to refactor (or use mock).

FAILED test/unit_test/api/db/services/test_dialog_service_final_answer.py::test_async_ask_final_event_carries_decorated_answer - peewee.OperationalError: (2003, "Can't connect to MySQL server on 'localhost' ([Errno 111] Connection refused)")
FAILED test/unit_test/api/db/services/test_dialog_service_final_answer.py::test_async_ask_delta_events_carry_incremental_text_only - peewee.OperationalError: (2003, "Can't connect to MySQL server on 'localhost' ([Errno 111] Connection refused)")
FAILED test/unit_test/api/db/services/test_dialog_service_final_answer.py::test_async_chat_final_event_carries_decorated_answer - ValueError: not enough values to unpack (expected 5, got 4)

edenfunf:
@xugangqiang Fixed — two root causes:

1. **async_ask MySQL error**: `kb_prompt()` lazily imports `DocumentService` and calls `get_by_ids()`, which opens a peewee connection at runtime. Added `monkeypatch.setattr(dialog_service, "kb_prompt", ...)` to both async_ask tests so the DB is never touched.

2. **async_chat unpack error**: My `get_models` mock was returning 4 values but `async_chat` unpacks it as `kbs, embd_mdl, rerank_mdl, chat_mdl, tts_mdl` (5 values). Fixed — mock now returns `([_KB], chat_mdl, None, chat_mdl, None)`.

Pushed as commit `5f62c2b`. No other changes.

PR #13934, opened 2026-04-04
fix: add recursive fallback search for MinerU content_list.json output

### LINKED ISSUE #13905: [Bug]: MinerU API integration fails with "Missing output file" despite successful API parsing
### Self Checks

none

### RAGFlow image version

v0.24.0

```Markdown
**Environment:**
* **RAGFlow Version:** v0.24.0 (Docker deployment)
* **MinerU Version:** (Installed via uv inside `/ragflow/uv_tools/.venv`)
* **OS/Deployment:** Docker (Local/Offline deployment)
```

### Actual behavior

**Description:**
When configuring MinerU as the chunking method via the API backend (`MINERU_APISERVER`), RAGFlow successfully sends the PDF, and MinerU processes it and returns a ZIP file. However, RAGFlow fails to locate the expected JSON output file inside the unzipped directory, throwing a `Missing output file` error. This occurs even with purely English, simple filenames (e.g., `Pre.pdf`).

Additionally, after the file matching fails, RAGFlow incorrectly falls back to executing the local CLI command, which causes a misleading `[ERROR]MinerU not found` log because the `mineru` executable is inside the `uv_tools` virtual environment rather than the global `$PATH`.

**Root Cause Analysis & Suggestions:**
1. **Output File Naming Mismatch:** It appears the latest version of MinerU has changed its ZIP output structure or JSON naming conventions. RAGFlow strictly expects `<filename>_content_list.json`, which no longer exists in the MinerU API response. RAGFlow's extraction logic (`LLMBundle`) needs to be updated to match the current MinerU output schema.
2. **Fallback Mechanism Bug:** If the API successfully returns a ZIP but RAGFlow fails to find the file, it shouldn't silently fall back to the local CLI mode. Furthermore, if CLI mode is intended as a fallback, RAGFlow should use the path defined in `MINERU_EXECUTABLE` rather than assuming `mineru` is in the global `$PATH`.

### Expected behavior

**Error Logs:**
```text
2026-04-02 20:02:54,676 INFO     98 [MinerU] Extract zip: zip_path=/tmp/mineru_pdf_mckn9b8v/Pre_auto_z5hagbcn.zip, extract_to=/tmp/mineru_pdf_mckn9b8v/Pre_auto_z5hagbcn, root_hint=Pre/
2026-04-02 20:02:54,689 INFO     98 [MinerU] Api completed successfully.
2026-04-02 20:02:54,690 INFO     98 [MinerU] Expected output files: Pre_content_list.json
2026-04-02 20:02:54,690 INFO     98 [MinerU] Searching output in: /tmp/mineru_pdf_mckn9b8v/Pre_auto_z5hagbcn
2026-04-02 20:02:54,692 ERROR    98 Failed to parse pdf via LLMBundle MinerU (Mineru): [MinerU] Missing output file, tried: /tmp/mineru_pdf_mckn9b8v/Pre_auto_z5hagbcn/Pre_content_list.json, /tmp/mineru_pdf_mckn9b8v/Pre_auto_z5hagbcn/Pre/Pre_content_list.json
2026-04-02 20:02:54,705 INFO     98 [ERROR]MinerU not found.
```

### Steps to reproduce

```Markdown
**Steps to Reproduce:**
1. Deploy RAGFlow via Docker Compose.
2. Start the MinerU API server manually inside the container (`mineru-api --host 0.0.0.0 --port 8886`).
3. Set environment variables in `entrypoint.sh` or `.env`: `MINERU_APISERVER=http://127.0.0.1:8886`, `MINERU_BACKEND=pipeline`.
4. Upload a simple English PDF (e.g., `Pre.pdf`) and set the Chunk Method to `MinerU`.
5. Observe the parsing failure in the task execution logs.
```

### Additional information

_No response_

octo-patch:
Fixes #13905

## Problem

When using the MinerU API server (`MINERU_APISERVER`), RAGFlow fails with `[MinerU] Missing output file` even though the API successfully returns a ZIP. This happens because the three-tier lookup in `_read_output` only tries specific paths:

1. `{file_stem}_content_list.json`
2. `{safe_stem}_content_list.json`
3. `{safe_stem}/{safe_stem}_content_list.json`

Newer versions of MinerU may place the JSON file in a different location within the ZIP, causing all three lookups to fail.

## Solution

Add a fourth fallback that recursively searches the extracted output directory for any file matching `*content_list.json`. This approach:

- Covers MinerU versions with different ZIP structures without breaking existing lookup order
- Logs a warning if multiple matches are found (uses the first sorted result)
- Only triggers if all three explicit lookups fail

## Testing

The existing logic is unchanged; the recursive search only activates as a last resort before raising `FileNotFoundError`.

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -543,7 +543,18 @@ class MinerUParser(RAGFlowPdfParser):
                     json_file = nested_alt
 
         if not json_file:
-            raise FileNotFoundError(f"[MinerU] Missing output file, tried: {', '.join(str(p) for p in attempted)}")
+            # Fallback: recursively search for any *content_list.json file
+            # to handle MinerU versions that use different naming conventions
+            found = sorted(output_dir.rglob("*content_list.json"))
+            if found:
+                json_file = found[0]
+                subdir = json_file.parent
+                if len(found) > 1:
+                    self.logger.warning(f"[MinerU] Multiple content_list.json found, using: {json_file}")
+                else:
+                    self.logger.info(f"[MinerU] Found via recursive search: {json_file}")
+            else:
+                raise FileNotFoundError(f"[MinerU] Missing output file, tried: {', '.join(str(p) for p in attempted)}")
 
         with open(json_file, "r", encoding="utf-8") as f:
             data = json.load(f)


SasukeOriental:
Yes, this helps a lot

xugangqiang:
Thanks for the contribution, @octo-patch! I reviewed this against the current `main` and wanted to share what I found.

## Findings

1. **The problem is real, but already solved on `main`.** Issue #13905 (MinerU API returns a ZIP but the three-tier lookup misses `content_list.json`) is a genuine integration break. Your recursive `rglob("*content_list.json")` fallback as a last resort is a reasonable direction.

2. **This PR overlaps with functionality already in `main`.** Current `main`'s `_read_output` (`deepdoc/parser/mineru_parser.py:707-826`) already has a recursive fallback that covers the same goal:
   ```python
   patterns.extend(["**/content_list.json", "**/*_content_list.json"])
   ```
   and it additionally guards against picking the wrong file via a `stem_match / in_stem_dir` check (skipping "unrelated fallback candidate"). That is strictly more robust than this PR's `sorted(...)[0]` pick-with-warning, which can silently select the wrong file when multiple `content_list.json` exist. So on `main` this change provides no incremental value and would just duplicate existing logic.

3. **The branch is stale and will not merge cleanly.** This PR was branched from the April `main` (the file was ~706 lines then); current `main` has ~1109 lines with extra vlm-subdir / parse_subdir / stem-dir lookups. The diff context no longer matches, so a merge would conflict or create two parallel fallback paths.

4. **No tests.** Existing `test/unit_test/deepdoc/parser/test_mineru_parser.py` mocks `_read_output`, so the new recursive branch has no coverage. (This is also true on `main`, but a backport would need one.)

5. **Scope note:** issue #13905's second sub-issue (wrong CLI fallback / `MinerU not found`) is not addressed here, but it appears already resolved elsewhere on `main`, so it's out of scope for this PR.

## Suggestion

Since the file-finding problem is already handled on `main` by the existing recursive fallback, would you be OK with **closing this PR**? If you're actually hitting this on an older release (e.g. v0.24.0 from the issue), a backport to that release branch with a unit test (fixture where `content_list.json` sits in a non-standard directory, plus a multi-file case) would be the right path instead.

Thanks again for the patch!

PR #13938, opened 2026-04-05
fix: guard whitespace-only chunks before embedding

shawnxiao105-afk:
## Problem

When parsing DOCX files with many tables, DeepDOC generates chunks containing only empty HTML table tags, such as:

```html
<table><tr><td></td></tr><tr><td></td></tr><tr><td></td></tr><tr><td></td></tr></table>
```

After the regex cleanup at `task_executor.py:584`, this becomes `"                  "` (whitespace only).

The guard at line 585 (`if not c`) only catches empty strings `""`, but whitespace strings are truthy in Python and pass through. When sent to Zhipu `embedding-3` API, it rejects them with error 1213: `未正常接收到prompt参数`.

## Root Cause

```python
c = re.sub(r"</?(table|td|caption|tr|th)( [^<>]{0,12})?>", " ", c)
if not c:       # ← only catches "", not "   " / "\n" / "\t"
    c = "None"
```

Verified with Zhipu `embedding-3`:
| Input | Result |
|---|---|
| `""` | error 1213 |
| `" "` | error 1213 |
| `"\n"` | error 1213 |
| `"None"` | OK |

## Fix

```diff
- if not c:
+ if not c.strip():
      c = "None"
```

## Testing

Reproduced with a 678KB DOCX file (166 tables, 270 chunks). Chunk #89 is the empty table above. After fix, `"None"` is sent instead and embedding succeeds.


shawnxiao105-afk:
Addressed CodeRabbit's suggestion in a7963f1 — added a `logging.debug` line when a whitespace-only chunk is normalized to `"None"`, so the silent rewrite is visible during triage. Per repo guideline (`**/*.py`: add logging for new flows). PR is now a 2-line change; ready when a maintainer has a moment 🙏

PR #13942, opened 2026-04-05
fix: handle missing 'total' key causing KeyError in deep research retrieval

vincentlambert:
## Summary

- When KB retrieval fails (e.g. ES `AssertionError` on empty `index_names`), `kbinfos` falls back to a dict without a `total` key
- `_async_update_chunk_info` then iterates over `chunk_info.keys()` (which includes `total`) and tries `kbinfos['total']`, raising a `KeyError`
- This error surfaces when using Tavily web retrieval in a chat with no knowledge base attached

## Changes

- Add `'total': 0` to all default `kbinfos` dicts in `_retrieve_information`
- Add `setdefault('total', 0)` guard after successful KB retrieval to handle cases where the retrieval result omits the key
- Accumulate `total` correctly in the merge branch of `_async_update_chunk_info`

## Test plan

- [ ] Start a chat with Tavily configured and no knowledge base
- [ ] Verify no `KeyError: 'total'` is raised
- [ ] Verify Tavily results are returned correctly


vincentlambert:
Hi @KevinHuSh — could you take a look at this PR when you have a moment?

This fixes a `KeyError: 'total'` that surfaces when using Tavily web retrieval with no knowledge base attached. The root cause is that `kbinfos` falls back to a dict without a `total` key when KB retrieval fails (e.g. ES `AssertionError` on empty `index_names`), and `_async_update_chunk_info` then crashes when iterating `chunk_info.keys()`.

In addition to the `total` fix, the PR also adds:
- try/except/finally in `research()` so `<END_DEEP_RESEARCH>` is always emitted even on failure
- defensive `.get()` on `suff["is_sufficient"]` and `suff["missing_information"]` to prevent `KeyError`s from malformed LLM responses

I've just pushed a final commit addressing CodeRabbit's nitpick (improving the `logging.exception` message to include the query for easier diagnosis).

Issue #13801, opened 2026-03-26
[Bug]:


trannluongg:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

main

### RAGFlow image version

infiniflow/ragflow:nightly

### Other environment information

```Markdown

```

### Actual behavior

I created a dataset, but when I selected "General," "Q&A," or something else as the built-in option, I kept getting the error "Chunk method is required."
<img width="936" height="1170" alt="Image" src="https://github.com/user-attachments/assets/eafde868-def5-41ba-bba7-33139ec7594f" />

### Steps to reproduce

```Markdown
Create dataset and input or select value
```

trannluongg:
@

> This bug was just fixed yesterday (March 25, 2026) in two merged PRs:
> 
> 1. [PR #13787](https://github.com/infiniflow/ragflow/pull/13787) - Fixed the dataset creation dialog by adding proper `name` prop handling to the `ChunkMethodItem` component
> 2. [PR #13760](https://github.com/infiniflow/ragflow/pull/13760) - Addressed the root cause: a field name mismatch between `'chunk_method'` and `'parser_id'` across frontend components
> 
> **Solution**: Pull the latest `nightly` image to get these fixes. Since you're using `infiniflow/ragflow:nightly`, run:
> 
> docker pull infiniflow/ragflow:nightly
> Then restart your containers.
> 
> If you're still experiencing issues after updating, you can try clearing your browser cache or using a different browser to rule out any cached frontend assets.
> 
> _To reply, just mention [@dosu](https://go.dosu.dev/dosubot?utm_source=github&utm_medium=bot-comment&utm_campaign=github-assistant&utm_content=brand-link)._
> 
[...]

PR #13945, opened 2026-04-06
fix: skip gpt-5 gen_conf clearing for custom/compatible API endpoints

### LINKED ISSUE #13944: OpenAI-API-Compatible: model names containing 'gpt-5' fail validation due to hardcoded special handling
## Environment

- RAGFlow version: v0.24.0
- Platform: macOS ARM64 (Mac Mini M4), self-built Docker image
- Doc engine: Elasticsearch 8.11.3
- Database: PostgreSQL 15

## Description

When adding an OpenAI-API-Compatible model with a name containing `gpt-5` (e.g., `gpt-5.4`), the model validation always fails with "Fail to access model" / "API key invalid", even though the API endpoint is fully functional and reachable from inside the container.

The same base URL and API key work correctly with model names that do NOT contain `gpt-5`.

## Root Cause Analysis

There are two issues in the codebase that affect model names containing `gpt-5`:

### 1. `_clean_conf()` in `rag/llm/chat_model.py` (line ~100)

```python
def _clean_conf(self, gen_conf):
    model_name_lower = (self.model_name or "").lower()
    if "gpt-5" in model_name_lower:
        gen_conf = {}
        return gen_conf
```

This unconditionally clears all generation config parameters for any model name containing `gpt-5`, including models served via OpenAI-API-Compatible endpoints that are NOT actual OpenAI gpt-5 models.

### 2. Possible litellm routing issue

The `set_api_key` validation endpoint appears to use litellm for model validation. litellm may intercept model names matching `gpt-5*` and route them to the official OpenAI API endpoint, ignoring the user-provided `base_url`. This causes authentication failures because the user's API key is for their custom endpoint, not OpenAI.

## Steps to Reproduce

1. Deploy RAGFlow v0.24.0
2. Set up any OpenAI-compatible API proxy (e.g., LiteLLM proxy, vLLM, etc.)
3. In RAGFlow UI, go to Model Providers → OpenAI-API-Compatible
4. Add a model with:
   - Model type: Chat
   - Model name: `gpt-5.4` (or any name containing `gpt-5`)
   - Base URL: `http://<your-proxy>:<port>/v1`
   - API key: your proxy's API key
5. Click verify → fails with "API key invalid"
6. Change the model name to anything NOT containing `gpt-5` (e.g., `my-model`) → works fine

## Expected Behavior

OpenAI-API-Compatible models should not be subject to hardcoded model-name-based special handling. The `gpt-5` special case in `_clean_conf()` should only apply when the factory is `OpenAI`, not when using `OpenAI-API-Compatible`.

## Suggested Fix

```python
# In _clean_conf(), scope the gpt-5 handling to OpenAI factory only:
if "gpt-5" in model_name_lower and self._FACTORY_NAME == "OpenAI":
    gen_conf = {}
    return gen_conf
```

Or better yet, use a more targeted check rather than substring matching on model names.

octo-patch:
Fixes #13944

## Problem

When a user adds an OpenAI-API-Compatible model with a name containing `gpt-5` (e.g., `gpt-5.4`), model validation fails with "API key invalid" even when the custom endpoint is fully functional.

The root cause: `_apply_model_family_policies()` with `backend="base"` unconditionally clears **all** generation config parameters for any model name containing `"gpt-5"`, regardless of whether the model is a real OpenAI endpoint or a custom-compatible one. This was introduced in #13675 to handle actual OpenAI gpt-5 API quirks.

The `base` backend is used exclusively by custom/compatible endpoint classes (`OpenAI_APIChat`, `LmStudioChat`, `LocalAIChat`, VLLM, etc.). Users on these endpoints may legitimately name their models `gpt-5.4` or similar without intending to use the real OpenAI gpt-5 API.

## Solution

Remove the `gpt-5` gen_conf clearing from the `base` backend path in `_apply_model_family_policies()`. Custom endpoints must be treated agnostically — we cannot infer model capabilities from the name alone.

The real OpenAI gpt-5 handling remains intact in the `litellm` backend path, where it is correctly scoped to `SupportedLiteLLMProvider.OpenAI` and `SupportedLiteLLMProvider.Azure_OpenAI` providers.

Also removes the now-unnecessary `gpt-5` early-return in `Base._clean_conf()` that was predicated on the above clearing.

## Testing

- Custom OpenAI-API-Compatible model named `gpt-5.4` now passes validation when the endpoint is reachable.
- Real OpenAI gpt-5 models (routed via litellm) still have `temperature`, `top_p`, `logprobs`, `top_logprobs` stripped as before.

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -543,7 +543,18 @@ class MinerUParser(RAGFlowPdfParser):
                     json_file = nested_alt
 
         if not json_file:
-            raise FileNotFoundError(f"[MinerU] Missing output file, tried: {', '.join(str(p) for p in attempted)}")
+            # Fallback: recursively search for any *content_list.json file
+            # to handle MinerU versions that use different naming conventions
+            found = sorted(output_dir.rglob("*content_list.json"))
+            if found:
+                json_file = found[0]
+                subdir = json_file.parent
+                if len(found) > 1:
+                    self.logger.warning(f"[MinerU] Multiple content_list.json found, using: {json_file}")
+                else:
+                    self.logger.info(f"[MinerU] Found via recursive search: {json_file}")
+            else:
+                raise FileNotFoundError(f"[MinerU] Missing output file, tried: {', '.join(str(p) for p in attempted)}")
 
         with open(json_file, "r", encoding="utf-8") as f:
             data = json.load(f)
--- a/rag/llm/chat_model.py
+++ b/rag/llm/chat_model.py
@@ -77,9 +77,9 @@ def _apply_model_family_policies(
         sanitized_kwargs["extra_body"] = {"enable_thinking": False}
 
     if backend == "base":
-        # GPT-5 and GPT-5.1 endpoints in this path have inconsistent generation-param support.
-        if "gpt-5" in model_name_lower:
-            sanitized_gen_conf = {}
+        # The "base" backend is used by custom/compatible endpoints (VLLM, LM-Studio, etc.).
+        # We cannot assume model capabilities from the name alone, so no model-family-specific
+        # policies are applied here. Standard sanitization in _clean_conf handles the rest.
         return sanitized_gen_conf, sanitized_kwargs
 
     if backend == "litellm":
@@ -151,14 +151,11 @@ class Base(ABC):
         return LLMErrorCode.ERROR_GENERIC
 
     def _clean_conf(self, gen_conf):
-        model_name_lower = (self.model_name or "").lower()
         gen_conf, _ = _apply_model_family_policies(
             self.model_name,
             backend="base",
             gen_conf=gen_conf,
         )
-        if "gpt-5" in model_name_lower:
-            return gen_conf
 
         if "max_tokens" in gen_conf:
             del gen_conf["max_tokens"]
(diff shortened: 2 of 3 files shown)


KevinHuSh:
Appreciations!

Resolve the conflicts please.

octo-patch:
Hi @KevinHuSh — thanks for the prompt and patience.

I went to rebase this PR onto current main and noticed that all three of its commits have already been addressed upstream by other PRs since this one was opened, which is also the source of the conflicts:

1. **`a7961d27` — `fix: skip model-family policies for custom/compatible API endpoints (fixes #13944)`** is the same fix that was merged in #13949 (`ff27ce86d`). Current `rag/llm/chat_model.py` already short-circuits the `base` backend before the `gpt-5` name-based clearing path.
2. **`0df99621` — `fix: correct validation error path for chunk_method in dataset creation dialog (fixes #13801)`** is now superseded by #14039 (`462be53b7`). The `superRefine` in `web/src/pages/datasets/dataset-creating-dialog.tsx` now uses `path: [ChunkMethodName]` (with `ChunkMethodName === "chunk_method"`), which is exactly the path this commit was correcting.
3. **`98ddb5ae` — `fix: add recursive fallback search for MinerU content_list.json (fixes #13905)`** is now covered by the more comprehensive pattern-based fallback in current `deepdoc/parser/mineru_parser.py` (`parse_subdir` / `stem_dir` patterns + final `**/content_list.json` glob). Rebasing the simple `rglob` fallback on top of that strictly removes coverage rather than adding it.

So rebasing this PR cleanly would reduce to an empty diff against current `main`. I think the cleanest path forward is to **close this PR as obsolete** — every issue it touches (#13944, #13801, #13905) is already fixed on `main`. Happy to close it from my side if you prefer; just confirm and I'''ll do it.

Apologies for the noise, and thanks again for reviewing.

KevinHuSh:
Thanks.

Issue #13894, opened 2026-04-02
[Bug]: metadata always included in chat enpoint even when disabled


adagoral:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

f32a832

### RAGFlow image version

v0.24

### Other environment information

```Markdown

```

### Actual behavior

metadata generated auto and i added some manully,

seems like everytime during chat (/chat endpoint) metadata is included in request even when It is disabled on UI

good place to debug this is in Langfuse

### Steps to reproduce

```Markdown
chatting in Chat mode

not checked Agent mode
```

adagoral:
and even when metadata is generated it show in Metadata column on UI 0 fields
@dosu

adagoral:
@dosu can you give me latest commit when it wasn't report for this metadata error?

adagoral:
both agent and chat has fields metadata, problem is i generated auto, 
```
ID: 137
├── Title: Cennik - Bramy przemysłowe.pdf
├── title: Cennik Bramy Przemysłowe
├── subtitle: ['Bramy segmentowe', 'Bramy roletowe', 'Bramy podwieszane przesuwne', 'Bramy rozwierne', 'Automatyka', 'Systemy przeładunkowe', 'Bramy harmonijkowe']
├── ProductType: ['bramy', 'okna', 'drzwi']
├── ModelSeries: [' 'Brama harmonijkowa ręczna']
├── SectionTopic: szyby
└── Content:
....some content xxxxxxxx
```

title, subtitle, productType, ModelSeries, SectionTopic is metadata, i disabled on UI but not working, am i right here? @Magicbook1108

adagoral:
i restored to this commit and it shows me on UI count of Metadata column
https://github.com/infiniflow/ragflow/commit/a1549c0fdc883ca9e1d96a7dcf20594667b1897f

but in prompt still is included

adagoral:
@dosu how to remove existing metadata?

adagoral:
[...]

PR #14019, opened 2026-04-10
Fix: respect disabled metadata setting in chat prompt assembly (#13894)

owldev127:

This PR fixes issue #13894, where document metadata was still included in chat prompt construction even when metadata was disabled in the Chat settings.

It separates:
- metadata filtering for retrieval
- document metadata injection into the LLM prompt

The fix adds explicit prompt-level control for document metadata inclusion and ensures disabled metadata filters are no longer treated as active.

### Root Cause

The chat pipeline only used `meta_data_filter` for retrieval filtering, while prompt generation always appended document metadata for retrieved documents.

As a result, disabling metadata in the UI did not prevent metadata from being included in the chat prompt.

### Type of change


rag/prompts/generator.py
```diff
@@ -119,13 +119,14 @@ def kb_prompt(kbinfos, max_tokens, hash_id=False):
...
-        docs_with_meta[d.id] = meta if meta else {}
-    docs = docs_with_meta
+    docs = {}
+    if include_document_metadata:
+        documents = DocumentService.get_by_ids([get_value(ck, "doc_id", "document_id") for ck in kbinfos["chunks"][:chunks_num]])
+        docs_with_meta = {}
+        for d in documents:
+            meta = DocMetadataService.get_document_metadata(d.id)
+            docs_with_meta[d.id] = meta if meta else {}
+        docs = docs_with_meta
```
xugangqiang, line 129: If I understand correctly, "docs" will be empty if "include_document_metadata" is not true.
In such case, "docs is empty" is the expected behavior? I am not saying this is not correct, just want to confirm your change intention.

xugangqiang:
@owldev127 
Could you please upload testing evidence so that I have confidence that regression tests have been properly conducted?

rag/prompts/generator.py
owldev127: Yes, that is intentional. `docs` only feeds the optional metadata lines in `kb_prompt`; the title, URL, and content still come directly from each retrieved chunk. When `include_document_metadata` is `false`, we want that metadata loop to no-op so only the core chunk payload is sent to the model.

xugangqiang:
@owldev127 

LGTM for the source code change. 
Need testing evidence to make sure no regression issues. Thanks.

owldev127:
Thank you for your patience. @xugangqiang 
Please find the test results below.

```
# python -m pytest test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py -q
================================================================================== test session starts ==================================================================================
platform linux -- Python 3.12.3, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/power/projects/gitt/ragflow
configfile: pyproject.toml
plugins: playwright-0.7.2, xdist-3.8.0, hypothesis-6.151.9, cov-7.1.0, anyio-4.13.0, asyncio-1.3.0, base-url-2.1.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 9 items                                                                                                                                                                       

test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py .........                                                                                            [100%]

================================================================================== 9 passed in 10.13s ===================================================================================
```

I agree that it would be best to run the full test suite via GitHub Actions.

Thanks again for your feedback.

xugangqiang:
@owldev127 

May I ask for your help to do regression tests on UI?
---------------
Regression testing is a type of software testing that ensures recent code changes (e.g., bug fixes, new features, or performance improvements) have not broken or degraded the existing functionality of the application.

In simple terms: You re-run old tests to make sure your new code didn't break anything that used to work.

xugangqiang:
No response from author for long time. Close it for now. Feel free to re-open if needed.

PR #14050, opened 2026-04-10
fix: make PaddleOCR timeout configurable and surface real ingest progress

joonsoome:

PaddleOCR request timeout was effectively hardcoded, which made long-form ingest jobs such as books or large PDFs fail or stall without a clear tuning option. In addition, the ingest pipeline did not clearly surface the later chunking, embedding, and indexing stages, so jobs could look stalled, failed, or finished too early even while the backend was still working.

This PR makes the timeout configurable end-to-end, wires it through the parser and UI, adds coverage for timeout propagation, and improves progress/log reporting so ingest reflects the real pipeline state.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Summary of changes

- Added configurable `paddleocr_request_timeout` support end-to-end.
- Wired the timeout through parser options, dataset defaults, model setup, and tests.
- Fixed progress reporting so parsing is not marked complete before chunking, embedding, and indexing are finished.
- Added stage-level LLM logs for keywords, questions, metadata, and embedding to make long ingest jobs easier to debug.
- Updated the PaddleOCR UI so request timeout is shown as part of the preset-defined configuration, matching the rest of the preset-managed fields.
- Kept `docker/.env` local-only by removing the tracked copy from the repository.

### Validation

- `python -m py_compile ...`
- `cd web && npm exec eslint src/components/paddleocr-options-form-field.tsx`

### Notes

- `cd web && npm exec tsc --noEmit` currently fails because of an existing repository `tsconfig.json` deprecation flag configuration, unrelated to this PR.

### Screenshot
paddleOCR-vl can handle documents with many pages or books, or when using custom models, exceeding the hardcoded 600-second limit in the code. Consequently, I received a failure message without any logs, and paddleOCR-api remained silent. Upon deep debugging, I discovered that the vLLM container was still processing. After reverse-analyzing the code, I found that RAGFlow had a hardcoded timeout of 600 seconds. By making this configurable, I was able to ingest my books as needed.
image

### Screenshot 2
There was a bug where, even after setting and selecting a preset, the system required the same value to be entered again. Moreover, this input did not actually override the preset.
스크린샷 2026-04-09 21-01-07

Therefore, I made the following changes:
스크린샷 2026-04-10 23-18-54


docker/.env
yingfeng, line 1: This file can not be removed

yingfeng (changes requested):

docker/.env
joonsoome: Oh I see, I'll leave it. It was keep remove my setup, and I didn't check it is removed at PR. Potentially, it would be nice .env to .env.example or something else later.

joonsoome:
Thanks for the review. I’ve addressed the PaddleOCR timeout/logging/test/i18n feedback and rewrote the branch history so the `docker/.env` deletion is no longer part of the PR diff or commit history. The inline review threads are resolved on my side, and I ran `python -m py_compile` and `git diff --check` locally.

yingfeng:
conflict arises here

yingfeng:
@copilot resolve the merge conflicts in this pull request

PR #14053, opened 2026-04-10
fix(infinity): escape single quotes in fulltext filter to prevent TokenError

Lntanohuang:
## Summary

Escape single quotes in `matchExpr.matching_text` before interpolating into the `filter_fulltext()` call in `InfinityConnection.search()`. Without this, user queries containing single quotes (e.g., `what's`) cause a `TokenError` from the Infinity query parser.

Split out from #13835 to keep PRs focused.

## Test plan

- [ ] Search with a query containing single quotes (e.g., `what's the meaning of life`)
- [ ] Verify no `TokenError` is raised and results are returned correctly

### DIFF
--- a/rag/utils/infinity_conn.py
+++ b/rag/utils/infinity_conn.py
@@ -177,7 +177,8 @@ class InfinityConnection(InfinityConnectionBase):
                         matchExpr.extra_options.update({"filter": filter_cond})
                     matchExpr.fields = [self.convert_matching_field(field) for field in matchExpr.fields]
                     fields = ",".join(matchExpr.fields)
-                    filter_fulltext = f"filter_fulltext('{fields}', '{matchExpr.matching_text}')"
+                    escaped_matching_text = matchExpr.matching_text.replace("'", "''")
+                    filter_fulltext = f"filter_fulltext('{fields}', '{escaped_matching_text}')"
                     if filter_cond:
                         filter_fulltext = f"({filter_cond}) AND {filter_fulltext}"
                     minimum_should_match = matchExpr.extra_options.get("minimum_should_match", 0.0)


qinling0210 (commented):
We should not just escape the character with error here. We should do the escape thing in where the matching text is generated.

Check if https://github.com/infiniflow/ragflow/pull/13969/changes has fixed your issue.

Lntanohuang:
Note: This PR is complementary to #13969 (already merged), not a duplicate.

#13969 strips single quotes during **synonym expansion** (in `query_base.py`, `rag/nlp/query.py`, `memory/services/query.py`) — it handles cases like WordNet returning `cat-o'-nine-tails`.

This PR escapes single quotes at the **`filter_fulltext()` construction** level in `infinity_conn.py` — it handles cases where the **user's original query** contains single quotes (e.g., `what's the meaning of life`). Without this fix, such queries still trigger a `TokenError` from the Infinity lexer even after #13969.

qinling0210:
@Lntanohuang 

I tested  in "retrieval testing" page of Dataset with question "what's the meaning of life?" with latest codebase,  no TokenError.

If you still see TokenError, let me know your scenario and testing method.

Lntanohuang:
Thanks @qinling0210 for testing this!

After tracing the code path carefully, I now realize the bug I was trying to fix is not actually reachable in practice. In `rag/nlp/query.py`, single quotes are stripped from user input at multiple points before `MatchTextExpr` is constructed:

- Line 57 (English path): `re.sub(r"[ \\\"'^]", "", tk)` strips single quotes
- Line 115 (Chinese path): single quotes are included in the special-char removal regex
- Line 125: another strip pass

So by the time `matching_text` reaches `infinity_conn.py`, single quotes are already gone. My fix would only matter if there's a code path that bypasses the query tokenizer entirely, which I haven't been able to identify.

Combined with #13969 already handling the synonym-expansion side, I don't think this PR adds real value in its current form. Closing it. Apologies for the noise, and thanks again for taking the time to verify.

PR #13464, opened 2026-03-08
Feat/configurable metadata display

Attili-sys:
### What problem does this PR solve?

Currently, RAGFlow's Search and Chat interfaces display only raw vectorized text chunks during retrieval, without contextual information about their source documents. Users cannot see document titles, page numbers, upload dates, or custom metadata fields that would help them understand and trust the retrieved results.

This PR introduces an **optional metadata display feature** that enriches retrieved chunks with document-level metadata in both the Search tab and Chatbot interface.

**Key improvements:**
- **Search results**: Display document metadata as styled badges beneath chunk snippets
- **Chat citations**: Show metadata in citation popovers and reference lists for better source context
- **LLM context**: Metadata is injected into the LLM prompt to enable more accurate, citation-aware responses
- **External API support**: Applications using RAGFlow's SDK retrieval endpoints (`/v1/retrieval`, `/v1/searchbots/retrieval_test`) can opt-in via request parameters
- **User control**: Multi-select dropdown UI allows users to choose which metadata fields to display

**Implementation approach:**
- ✅ Reuses existing `DocMetadataService` infrastructure (no new database tables or indices)
- ✅ Settings stored in existing JSON configuration fields (`search_config.reference_metadata`, `prompt_config.reference_metadata`)
- ✅ No database migrations required
- ✅ Disabled by default (fully opt-in and backward-compatible)
- ✅ Dynamic metadata field selection populated from actual document metadata keys
- ✅ Fixed critical bug where Python's builtin `set()` was shadowed by a route handler function

**Modified endpoints (all backward-compatible):**
- `POST /v1/retrieval` (Public SDK)
- `POST /v1/searchbots/retrieval_test` (Searchbots)
- `POST /v1/chunk/retrieval_test` (UI/Internal)
- Chat completions endpoints (via `extra_body.reference_metadata` or `prompt_config`)

### Type of change

- [x] New Feature (non-breaking change which adds functionality)

###Images
-
image
<br><br>
<br><br>

image
<br><br>
<br><br>

image

PR #14054, opened 2026-04-11
fix: skip document metadata in LLM prompt when KB has enable_metadata=False

### LINKED ISSUE #13894: [Bug]: metadata always included in chat enpoint even when disabled

octo-patch:
Fixes #13894

## Problem

When a dataset's auto-metadata feature is disabled (`enable_metadata=False` in the dataset's parser config), document metadata fields (e.g. `title`, `subtitle`, `ProductType`, etc.) were still being included in the LLM prompt. This happened because `kb_prompt()` in `rag/prompts/generator.py` fetched and injected stored document metadata unconditionally, without checking whether the dataset owner had disabled the metadata feature.

As a result, users who disabled metadata auto-generation still saw metadata in LLM requests (visible in tools like Langfuse), contrary to their expectations.

## Solution

Before fetching document metadata, `kb_prompt()` now checks each knowledge base's `enable_metadata` setting. Metadata is only loaded from `DocMetadataService` and added to the prompt for documents belonging to knowledge bases that have `enable_metadata=True`. Documents from knowledge bases with `enable_metadata=False` get an empty metadata dict, so no metadata fields appear in the LLM prompt.

## Testing

- Knowledge base with `enable_metadata=True`: document metadata continues to appear in the LLM prompt as before.
- Knowledge base with `enable_metadata=False`: document metadata is excluded from the LLM prompt even if metadata was previously stored for those documents.

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -35,7 +35,7 @@ from api.db.services.doc_metadata_service import DocMetadataService
 from api.db.services.knowledgebase_service import KnowledgebaseService
 from api.db.services.langfuse_service import TenantLangfuseService
 from api.db.services.llm_service import LLMBundle
-from common.metadata_utils import apply_meta_data_filter
+from common.metadata_utils import apply_meta_data_filter, derive_prompt_meta_fields
 from api.utils.reference_metadata_utils import (
     enrich_chunks_with_document_metadata,
     resolve_reference_metadata_preferences,
@@ -709,7 +709,7 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
         )
         _enrich_chunks_with_document_metadata(kbinfos.get("chunks", []), metadata_fields)
 
-    knowledges = kb_prompt(kbinfos, max_tokens)
+    knowledges = kb_prompt(kbinfos, max_tokens, meta_fields=derive_prompt_meta_fields(dialog.meta_data_filter))
     logging.debug("{}->{}".format(" ".join(questions), "\n->".join(knowledges)))
 
     retrieval_ts = timer()
@@ -1530,7 +1530,7 @@ async def async_ask(question, kb_ids, tenant_id, chat_llm_name=None, search_conf
         )
         _enrich_chunks_with_document_metadata(kbinfos.get("chunks", []), metadata_fields)
 
-    knowledges = kb_prompt(kbinfos, max_tokens)
+    knowledges = kb_prompt(kbinfos, max_tokens, meta_fields=derive_prompt_meta_fields(meta_data_filter))
     sys_prompt = PROMPT_JINJA_ENV.from_string(ASK_SUMMARY).render(knowledge="\n".join(knowledges))
 
     msg = [{"role": "user", "content": question}]
--- a/common/metadata_utils.py
+++ b/common/metadata_utils.py
@@ -177,10 +177,13 @@ async def apply_meta_data_filter(
     """
     Apply metadata filtering rules and return the filtered doc_ids.
 
-    meta_data_filter supports three modes:
-    - auto: generate filter conditions via LLM (gen_meta_filter)
-    - semi_auto: generate conditions using selected metadata keys only
-    - manual: directly filter based on provided conditions
+    meta_data_filter supports four modes:
+
+    - ``disabled``: skip all metadata filtering; ``base_doc_ids`` is returned
+      unchanged (including the empty list and ``None`` cases).
+    - ``auto``: generate filter conditions via LLM (gen_meta_filter).
+    - ``semi_auto``: generate conditions using selected metadata keys only.
+    - ``manual``: directly filter based on provided conditions.
 
     When ``kb_ids`` is supplied and the active doc store is Elasticsearch the
     generated filter conditions are pushed down to ES via
@@ -197,8 +200,9 @@ async def apply_meta_data_filter(
     ``get_flatted_meta_by_kbs`` round-trip entirely.
 
     Returns:
-        list of doc_ids, ["-999"] when manual filters yield no result, or None
-        when auto/semi_auto filters return empty.
+        list of doc_ids, ``["-999"]`` when manual filters yield no result,
+        ``None`` when auto/semi_auto filters return empty, or ``base_doc_ids``
+        as-is when method is ``disabled``.
     """
     from rag.prompts.generator import gen_meta_filter # move from the top of the file to avoid circular import
 
@@ -286,6 +290,48 @@ def _try_meta_pushdown(
     except Exception as e:
         logging.warning(f"[apply_meta_data_filter] push-down errored, falling back: {e}")
         return None
+
+
+def derive_prompt_meta_fields(meta_data_filter: dict | None) -> set[str] | None:
+    """
+    Decide which document metadata fields should be exposed in the LLM prompt
+    based on the conversation-level ``meta_data_filter`` config.
+
+    Returns:
+        - ``None`` when every metadata field should be included
+          (no filter configured, or ``method == "auto"`` where the LLM is
+          free to pick any field at retrieval time, so we surface them all).
+        - empty ``set()`` when no metadata field should appear in the prompt
+          (``method == "disabled"``).
+        - non-empty ``set[str]`` of field names for ``manual`` and
+          ``semi_auto`` modes (only the fields the user actually filtered on
+          are surfaced — this keeps the prompt token-efficient).
+    """
+    if not meta_data_filter:
+        return None
+    method = meta_data_filter.get("method")
+    if method == "disabled":
+        return set()
+    if method == "manual":
+        keys: set[str] = set()
+        for cond in meta_data_filter.get("manual", []) or []:
+            if isinstance(cond, dict):
+                key = cond.get("key") or cond.get("name")
+                if isinstance(key, str) and key:
+                    keys.add(key)
+        return keys
+    if method == "semi_auto":
+        keys = set()
+        for item in meta_data_filter.get("semi_auto", []) or []:
+            if isinstance(item, str):
+                keys.add(item)
+            elif isinstance(item, dict):
+                key = item.get("key")
+                if isinstance(key, str) and key:
+                    keys.add(key)
+        return keys
+    # method == "auto" or unknown — let everything through.
+    return None
 
 
 def dedupe_list(values: list) -> list:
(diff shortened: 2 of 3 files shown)


zpf121:
Hi, @octo-patch. Thank you for your PR. There are a few issues and areas for improvement, as detailed below:

## Issues with Current PR

### 1. Wrong Configuration Source

Uses `parser_config.enable_metadata` (KB-level parsing config) instead of `meta_data_filter` (conversation-level config).

These are different concepts:
- **`parser_config.enable_metadata`**: Controls metadata extraction during document parsing
- **`meta_data_filter`**: Controls whether to use metadata filtering in conversation

### 2. N+1 Query Performance Issue

Loop performs one database query per knowledge base instead of batch query.

Example: 10 KBs → 10 database queries with network round-trips each.

### 3. Doesn't Match User Requirements

User needs: "When metadata filtering is disabled, don't show metadata in prompt"

Current PR checks KB config, but should check conversation config.

### 4. Missing Safety Checks

No null check for `parser_config` before calling `.get()`, can cause AttributeError.

---

## Optimized Approach

### Core Change

**Control metadata inclusion based on conversation's `meta_data_filter`, not KB's `enable_metadata`**

### Per-Mode Behavior

| Mode | Show in Prompt | DB Query |
|------|---------------|----------|
| `disabled` | ❌ None | Skip queries |
| `manual` | ✅ Only filter fields | Execute queries |
| `semi_auto` | ✅ Only selected fields | Execute queries |
| `auto` | ✅ All fields (unpredictable) | Execute queries |

### Key Improvements

1. ✅ Correct semantics: Check conversation config, not KB config
2. ✅ Token efficient: Only show relevant fields when possible
3. ✅ Performance: Skip DB queries when disabled
4. ✅ Field filtering: Extract and display only used fields

octo-patch:
Hi @zpf121, thanks for the careful review — you're right that the original PR keyed off the KB's `enable_metadata` flag and added a hot-path query, which both misread the intent and degraded performance. I've rebuilt the patch on current `main` (force-pushed). Now the prompt-side metadata exposure follows the **conversation-level** `meta_data_filter` instead, with no extra DB queries.

### Behaviour by mode (per `meta_data_filter.method`)

| method | metadata fields surfaced in the LLM prompt | per-doc metadata DB query |
|--|--|--|
| `disabled` | none | **skipped** |
| `manual` | only the keys listed in `manual` filters | yes (existing) |
| `semi_auto` | only the keys listed in `semi_auto` selection | yes (existing) |
| `auto` | all fields (LLM is free to pick at retrieval time, so we surface them all) | yes (existing) |
| not configured | all fields (legacy behaviour, unchanged) | yes (existing) |

### What changed

- `rag/prompts/generator.py::kb_prompt(...)` gains an opt-in `meta_fields` parameter:
  - `None` (default) → include every metadata field (legacy)
  - empty collection → suppress all metadata **and skip the per-doc DB lookups entirely**
  - non-empty collection → include only the named keys
- New helper `common.metadata_utils.derive_prompt_meta_fields(meta_data_filter)` derives the right value from the conversation config.
- The two conversation chat entry points (`async_chat`, `async_ask` in `dialog_service.py`) now pass `derive_prompt_meta_fields(...)` to `kb_prompt(...)`.
- Agent / tool callers (`agent_with_tools.py`, `agent/tools/retrieval.py`, `agent/tools/base.py`, `tree_structured_query_decomposition_retrieval.py`) keep `meta_fields=None`, preserving today's behaviour — those paths don't have a conversation context.
- `apply_meta_data_filter()`: `method == "disabled"` is now a true no-op — it returns `base_doc_ids` unchanged (including `None` / `[]`), so the retrieval pipeline never silently drops documents in disabled mode (per @coderabbitai's feedback).
- Added debug logs in both `apply_meta_data_filter()` and `kb_prompt()` so the gating decision is observable in logs.

### Addressing the original concerns

- **"`enable_metadata` is a configuration field for KBs, not the conversation"** → fixed; the gate is now driven entirely by conversation-level `meta_data_filter`.
- **N+1 queries on the hot path** → no new queries are introduced anywhere. In `disabled` mode we actually skip the existing per-doc metadata lookups in `kb_prompt`.
- **Doesn't match user intent** → `manual` / `semi_auto` now restrict the prompt to only the keys the user is filtering on, instead of dumping every metadata field.
- **Missing null checks** → the helper is null-safe (`if not meta_data_filter: return None`), and `meta_fields_set is None` vs empty set is handled distinctly inside `kb_prompt`.

Branch is now a single commit on top of `main`. Happy to iterate further.

zpf121:
Thanks for working on this fix! The logic of this PR is correct, but there are merge conflicts from #13464. Could you please rebase your branch on the latest main and resolve the merge conflicts?

octo-patch:
Hi @zpf121 — rebased onto current `main` and resolved the merge conflicts from #13464. Force-pushed.

Quick note on how the two changes layer together now:

- #13464 introduced `_enrich_chunks_with_document_metadata`, which populates `document_metadata` on each chunk based on the dialog `prompt_config.reference_metadata` opt-in. `kb_prompt()` then just renders whatever metadata was populated, so the per-doc DB queries are gone — that addresses the perf nit from CodeRabbit.
- This PR keeps `derive_prompt_meta_fields(meta_data_filter)` and threads its result through a new `meta_fields` parameter on `kb_prompt()`, which acts as a final filter on what gets rendered into the prompt:
  - `disabled` → empty set → strip all metadata from the prompt block (the original #13894 bug).
  - `manual` / `semi_auto` → only the keys the user is filtering on are surfaced (token-efficient).
  - `auto` / unset → `None` → preserve legacy behaviour.

Net effect: `meta_data_filter` intent is now respected at the prompt boundary even when `reference_metadata` enrichment has already populated extra fields on the chunks.

Let me know if you would prefer a different shape — happy to adjust.

zpf121:
Thank you so much for your PR! I tested both agent and chat modes and found that the agent metadata filtering is not working properly. Here are the current issues:

- **Agent Missing Metadata Enrichment**
Agent mode is missing the enrich_chunks_with_document_metadata() call that exists in dialog_service.py. This causes chunks to lack document_metadata fields in agent mode. The fix would be to add the enrichment call to agent/tools/retrieval.py before kb_prompt() is called.

- **Test Compatibility After Interface Changes**
After the changes in this PR, some test mocks may no longer be compatible with the updated kb_prompt() interface. A batch update to the test functions may be needed to ensure they properly handle the new parameters.

KevinHuSh:
Appreciations!

CI failure.

```
____________ test_async_chat_uses_all_docs_when_no_doc_ids_selected ____________
test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py:312: in test_async_chat_uses_all_docs_when_no_doc_ids_selected
    result = asyncio.run(_collect())
             ^^^^^^^^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py:195: in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py:118: in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/base_events.py:691: in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py:308: in _collect
    async for item in dialog_service.async_chat(dialog, [{"role": "user", "content": "What does the dataset say?"}], stream=False):
api/db/services/dialog_service.py:712: in async_chat
    knowledges = kb_prompt(kbinfos, max_tokens, meta_fields=derive_prompt_meta_fields(dialog.meta_data_filter))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   TypeError: test_async_chat_uses_all_docs_when_no_doc_ids_selected.<locals>.<lambda>() got an unexpected keyword argument 'meta_fields'
```

KevinHuSh:
Appreciations!

CI failure.
```
=================================== FAILURES ===================================
____________ test_async_chat_uses_all_docs_when_no_doc_ids_selected ____________
test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py:312: in test_async_chat_uses_all_docs_when_no_doc_ids_selected
    result = asyncio.run(_collect())
             ^^^^^^^^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py:195: in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py:118: in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/asyncio/base_events.py:691: in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py:308: in _collect
    async for item in dialog_service.async_chat(dialog, [{"role": "user", "content": "What does the dataset say?"}], stream=False):
api/db/services/dialog_service.py:712: in async_chat
    knowledges = kb_prompt(kbinfos, max_tokens, meta_fields=derive_prompt_meta_fields(dialog.meta_data_filter))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   TypeError: test_async_chat_uses_all_docs_when_no_doc_ids_selected.<locals>.<lambda>() got an unexpected keyword argument 'meta_fields'
···

PR #14060, opened 2026-04-12
Fix: safe argument guard and remove redundant redis call

paulhuiseismic:

- Moved if not all([email, new_pwd, new_pwd2]) guard to the top, before any decryption that could crash on None value
- Removed the redundant REDIS_CONN.get() call — one call is sufficient

### Type of change

- [x] Refactoring

### DIFF
--- a/api/apps/restful_apis/user_api.py
+++ b/api/apps/restful_apis/user_api.py
@@ -806,16 +806,15 @@ async def forget_reset_password():
     new_pwd = req.get("new_password")
     new_pwd2 = req.get("confirm_new_password")
 
-    new_pwd_base64 = decrypt(new_pwd)
-    new_pwd_string = base64.b64decode(new_pwd_base64).decode('utf-8')
-    new_pwd2_string = base64.b64decode(decrypt(new_pwd2)).decode('utf-8')
+    if not all([email, new_pwd, new_pwd2]):
+        return get_json_result(data=False, code=RetCode.ARGUMENT_ERROR, message="email and passwords are required")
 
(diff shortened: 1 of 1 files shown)


xugangqiang:
@paulhuiseismic 
May I ask for your help to do regression test and upload testing evidence? 
And if possible,  could you please do me a favor to add unit test - which should be failed before your fix and should success with your fix?
Thank you.

Magicbook1108 (commented):
looks great

paulhuiseismic:
Hi @xugangqiang this is just for code enhancement and should be no breaking change without regression issue. So would you like me to add the UT also? cc: @Magicbook1108

xugangqiang:
@paulhuiseismic 
1. would you like me to add the UT also -> Yes, if UT can be added that will be great
2. just for code enhancement -> I think a regression test is needed to make sure your change will not break the logic. As you said, "should be no breaking change". But we want to make sure no regression issue (like typo, un-expected change, ...etc.). Without regression test, we are not confident there is no regression.

xugangqiang:
@paulhuiseismic  
Could you please resolve conflicts as well? Thanks.

paulhuiseismic:
Hi @xugangqiang @Magicbook1108 Have resolved the conflicts and refined the PR, please help to review again.

xugangqiang:
May I ask your help for regression test?

I think a regression test is needed to make sure your change will not break the logic. As you said, "should be no breaking change". But we want to make sure no regression issue (like typo, un-expected change, ...etc.). Without regression test, we are not confident there is no regression.

PR #14063, opened 2026-04-13
Fix: sandbox cannot accept large args list

yongtenglei:

Sandbox cannot accept large args list.

### Type of change

### DIFF
--- a/agent/sandbox/executor_manager/services/execution.py
+++ b/agent/sandbox/executor_manager/services/execution.py
@@ -56,33 +56,16 @@ def _extract_result_envelope(stdout: str) -> tuple[str, ExecutionStructuredResul
     return cleaned_stdout, envelope
 
 
-async def execute_code(req: CodeExecutionRequest):
-    """Fully asynchronous execution logic"""
-    language = req.language
-    container = await allocate_container_blocking(language)
-    if not container:
-        return CodeExecutionResult(
-            status=ResultStatus.PROGRAM_RUNNER_ERROR,
-            stdout="",
-            stderr="Container pool is busy",
-            exit_code=-10,
-            detail="no_available_container",
-        )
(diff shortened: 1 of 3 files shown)


agent/sandbox/executor_manager/services/execution.py
```diff
@@ -56,33 +56,16 @@ def _extract_result_envelope(stdout: str) -> tuple[str, ExecutionStructuredResul
...
-            code_path = os.path.join(workdir, code_name)
-            with open(code_path, "wb") as f:
-                f.write(base64.b64decode(req.code_b64))
-            runner_name = "runner.py"
-            runner_path = os.path.join(workdir, runner_name)
-            with open(runner_path, "w") as f:
-                f.write(f"""import base64
+def _build_execution_bundle(req: CodeExecutionRequest, workdir: str) -> dict[str, str | bytes]:
+    arguments = req.arguments or {}
+    args_source = json.dumps(arguments, ensure_ascii=False)
```
yongtenglei, line 61: That's the right way.

agent/sandbox/executor_manager/services/execution.py
```diff
@@ -185,12 +162,77 @@ def emit_result(value):
...
+    workdir = f"/tmp/sandbox_{task_id}"
+    os.makedirs(workdir, mode=0o700, exist_ok=True)
+
+    try:
+        bundle = _build_execution_bundle(req, workdir)
+        code_name = str(bundle["code_name"])
+        runner_name = str(bundle["runner_name"])
+
+        code_path = os.path.join(workdir, code_name)
+        with open(code_path, "wb") as f:
```
wangq8, line 218: should we use "w" instead of "wb" since we are writing text file instead of binary file?

wangq8: looks like after b64decode the code stream is a binary string instead of a string.

PR #14067, opened 2026-04-13
Refactor: Change update doc from PUT to patch

xugangqiang:

Before change, update_document in api/apps/restful_apis/document_api.py is using "PUT".
After change, it will use "PATCH" which is more suitable.

### Type of change

- [x] Refactoring


yuzhichang (commented):
LGTM

PR #14074, opened 2026-04-13
Checkpoint mechanism for long-running workflow jobs

ShorthillsAI:
## What problem does this PR solve?
 
Fixes #12494 (Checkpoint mechanism for long-running workflow jobs).
 
GraphRAG and RAPTOR tasks can run for hours or even days on large knowledge bases, making hundreds of LLM calls per document. Today, if the process crashes, loses network, or is interrupted mid-run, **all progress is lost** — the next run starts again from the first document, wasting time and re-spending the LLM budget on work that was already done.
 
This PR adds a Redis-backed checkpoint service that records each completed `doc_id` as it finishes. On restart, previously-completed docs are skipped:
 
- **GraphRAG:** The saved subgraph is loaded from the doc store instead of re-running the LLM extraction, then merged normally. If the stored subgraph isn't found, it falls back to regeneration.

- **RAPTOR:** The doc is skipped entirely in `scope=file` mode since its chunks are already persisted.
 
### Design choices
 
- **Redis `SADD`** instead of MySQL — atomic, concurrent-safe (GraphRAG runs `build_one` in parallel across up to 4 docs), no DB schema migration, and Valkey/Redis is already a hard dependency.

- **Per-doc granularity** — keeps the checkpoint simple and sufficient for the real-world failure modes. Chunk-level checkpointing would add complexity without much benefit since individual docs are usually minutes, not hours.

- **7-day TTL** on checkpoint keys so they auto-clean if a task is abandoned.

- **Graceful degradation** — if Redis is unavailable, the service logs and continues; tasks just run without checkpointing rather than crashing.
 
### What's intentionally out of scope
 
- **Agent workflow checkpoint (Canvas.run)** — the canvas state is much richer than `doc_ids` (component outputs, streaming state, globals). Doing it properly requires full Canvas serialization, which is a separate, bigger change. Will be a follow-up PR.

- **Resume API endpoint** — depends on the agent checkpoint above.
 
### Changes
 
| File | Change |

|---|---|

| `api/db/services/checkpoint_service.py` (new) | `save_checkpoint`, `load_checkpoint`, `clear_checkpoint` via Redis SADD |

| `rag/graphrag/general/index.py` | Checkpoint load/save/clear in `run_graphrag_for_kb`; new `_load_subgraph_from_store` helper |

| `rag/svr/task_executor.py` | Checkpoint load/save/clear in `run_raptor_for_kb` |

| `test/unit_test/api/db/services/test_checkpoint_service.py` (new) | 18 unit tests |
[...]

Issue #12494, opened 2026-01-07
[Feature Request]: Support checkpoint for Agent workflow


yingfeng:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

Checkpoint mechanism is a must for long-running workflow job

### Documentation, adoption, use case

```Markdown

```

SHAI-minal:
Hi, I'd like to work on this. My plan is to implement checkpoint/resume in two stages — first for long-running RAG indexing tasks (GraphRAG and RAPTOR), then extend it to Agent workflow execution. I'll start with GraphRAG document parsing as the foundation.

SHAI-minal:
@KevinHuSh  @yingfeng — I have a working implementation ready (GraphRAG + RAPTOR checkpoint via Redis). Would you be open to reviewing a PR, or is there existing work on this I should align with?

PR #14096, opened 2026-04-14
Refact: improve task resume mechanism for graphrag

SHAI-minal:

Addresses review feedback on #14074 (Checkpoint mechanism for long-running workflow jobs, issue #12494).

**Changes based on @yuzhichang's review:**

1. **Renamed `checkpoint_service.py` → `task_checkpoint.py`** as suggested.
2. **Replaced Redis with direct docEngine queries** as suggested — the subgraph already gets persisted to the doc store by `generate_subgraph()`, so we just query for it instead of maintaining a separate checkpoint in Redis. This is simpler, has no extra dependency, and uses a single source of truth.

**Changes based on CodeRabbit review:**

3. **Fixed `source_id` query format mismatch** — subgraphs are stored with `source_id: [doc_id]` (list), but the original query used `source_id: doc_id` (string). Now follows the same pattern as `does_graph_contains()` in `rag/graphrag/utils.py`: filter by `knowledge_graph_kwd` only, then match `source_id` in Python. This avoids ambiguity across Elasticsearch / Infinity / OceanBase backends.

### Changes

| File | Change |
|---|---|
| `api/db/services/task_checkpoint.py` (new) | `load_subgraph_from_store()` and `has_raptor_chunks()` — docEngine-based checkpoint queries |
| `rag/graphrag/general/index.py` | `build_one()` calls `load_subgraph_from_store()` before running LLM extraction |
| `rag/svr/task_executor.py` | RAPTOR per-doc loop calls `has_raptor_chunks()` before processing |
| `test/unit_test/rag/graphrag/test_checkpoint_resume.py` (new) | 10 unit tests covering subgraph loading, source_id filtering, edge cases |

### How it works

- **GraphRAG:** Before running expensive LLM entity/relation extraction for a doc, checks the doc store for an existing subgraph (saved by a previous interrupted run). If found, loads it directly and skips LLM calls.
- **RAPTOR:** Before processing a doc, checks if RAPTOR chunks (`raptor_kwd="raptor"`) already exist for it. If yes, skips.

### Testing

- 10 new unit tests — all passing
- Full existing suite: 617 passed

### Type of change

- [x] Refactoring


api/db/services/task_checkpoint.py
```diff
@@ -0,0 +1,103 @@
...
+
+import networkx as nx
+
+from common import settings
+from common.doc_store.doc_store_base import OrderByExpr
+from common.misc_utils import thread_pool_exec
+from rag.nlp import search
+
+
+async def load_subgraph_from_store(tenant_id: str, kb_id: str, doc_id: str):
```
yuzhichang: Move functions of this file to the place where they are needed. Remove this file.

test/unit_test/rag/graphrag/test_checkpoint_resume.py
```diff
@@ -0,0 +1,320 @@
...
+    "rag.utils.base64_image",
+    "rag.utils.raptor_utils",
+    "rag.prompts.generator",
+    "rag.raptor",
+    "rag.app",
+    "rag.graphrag.utils",
+]
+for _m in _EXTRA_MOCKS:
+    if _m not in sys.modules:
+        sys.modules[_m] = MagicMock()
```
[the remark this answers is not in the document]
SHAI-minal: This follows the same sys.modules patching pattern already used in conftest.py for this test directory. Happy to revisit in a follow-up if the maintainers prefer a different approach.

SHAI-minal:
@yuzhichang @wangq8 , Please review,

rag/graphrag/general/index.py
```diff
@@ -43,6 +43,67 @@
 from rag.nlp import rag_tokenizer, search
 from rag.utils.redis_conn import RedisDistributedLock
 from common import settings
+from common.doc_store.doc_store_base import OrderByExpr
+
+_PAGE_SIZE = 256
+
+
+async def load_subgraph_from_store(tenant_id: str, kb_id: str, doc_id: str):
```
wangq8, line 50: @MinalMahalaShorthillsAI If we have 10,000 results and 100 `doc_id`, the current approach would run 100 queries to ES/DB, each retrieving 10,000 results, this creates significant load and inefficiency.

A better approach is to query once and process all `doc_id` in memory, reducing the number of queries and overall overhead.

An even more optimal solution is to include doc_id in the search criteria, allowing ES/DB to filter results directly. This leverages indexing and avoids fetching a large result set only to filter it afterward, significantly improving performance if search with `doc_id` supported.

rag/graphrag/general/index.py
```diff
@@ -43,6 +43,67 @@
...
+    """Load a previously saved subgraph from the doc store.
```
yuzhichang, line 51: See rag/graphrag/general/index.py  `generate_subgraph()`
```
    chunk = {
        "content_with_weight": json.dumps(nx.node_link_data(subgraph, edges="edges"), ensure_ascii=False),
        "knowledge_graph_kwd": "subgraph",     # filter condition 1
        "kb_id": kb_id,
        "source_id": [doc_id],                                     # filter condition 2
        "available_int": 0,
        "removed_kwd": "N",
    }

```

rag/graphrag/general/index.py
SHAI-minal: Thanks for the suggestion! We've now included source_id: [doc_id] directly in the search condition (as hinted by @yuzhichang 's generate_subgraph reference), so the doc store filters by index. Each call fetches at most 1 chunk with limit=1. No pagination or Python-side filtering needed.

SHAI-minal:
@yuzhichang 
Friendly ping - happy to make any further changes if needed!

yuzhichang (approved):

wangq8 (approved):

SHAI-minal:
@yingfeng @yuzhichang @wangq8 , It was very nice working with you all, Hoping for collaboration.

PR #14130, opened 2026-04-15
feat(templates): add a data analysis agent template

ZhenhangTung:

Add a new agent template that demonstrates how to leverage the `CodeExec` component to do the data analysis. 

### Type of change

- [x] Other (please describe): Agent template


agent/templates/data_analysis_beginner_assistant.json
```diff
@@ -0,0 +1,296 @@
+{
+  "id": 37,
+  "title": {
+    "en": "Data analysis beginner assistant",
```
writinwaters: Beginner's data analytics assistant

writinwaters (changes requested):

Issue #14066, opened 2026-04-13
[Question]: OIDC authorization is broken?


Kn3pXGraF:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

I used all the recommendations from previous posts, but nothing helped.

```
docker exec -it ragflow-server env | grep SECRET_KEY
RAGFLOW_SECRET_KEY=sh********************oL
SECRET_KEY=sh********************oL
```

```
docker exec -it ragflow-server grep secret_key /ragflow/conf/service_conf.yaml
  secret_key: 'sh********************oL'
```

```
docker exec -it ragflow-server sed -n '77,78p' /ragflow/api/apps/__init__.py
app.config["SESSION_TYPE"] = "redis"
app.config["SESSION_REDIS"] = settings.decrypt_database_config(name="redis")
```
When logging in via a keycloak, it immediately crashes from the profile with errors:

```
[2026-04-13 09:35:52 +0500] [16] [INFO] 127.0.0.1:50300 GET /v1/user/login/oidc 1.1 302 777 10587
[2026-04-13 09:35:52 +0500] [16] [INFO] 127.0.0.1:50302 GET /v1/user/oauth/callback/oidc 1.1 302 363 41976
2026-04-13 09:35:53,266 WARNING  16 load_user got exception Signature b'Ie1wNH5UJWyxZrgyKafsMFdvToc' does not match
2026-04-13 09:35:53,267 WARNING  16 Unauthorized request (quart_auth)
[2026-04-13 09:35:53 +0500] [16] [INFO] 127.0.0.1:50312 POST /v1/dialog/next 1.1 401 72 2021
2026-04-13 09:35:53,271 WARNING  16 load_user got exception Signature b'Ie1wNH5UJWyxZrgyKafsMFdvToc' does not match
2026-04-13 09:35:53,271 WARNING  16 Unauthorized request (quart_auth)
[...]

PR #14134, opened 2026-04-15
Strip `Bearer` prefix before verifying itsdangerous token in _load_user()

euvre:

Reslove #14066 .

## Root Cause

In `_load_user()` (`api/apps/__init__.py`), the full `Authorization` header value is passed directly to `itsdangerous.URLSafeTimedSerializer.loads()` without stripping the `"Bearer "` prefix.

The two login paths send different header formats:

| Login method | Authorization header |
|---|---|
| Password login | `{signed_token}` (no prefix) |
| OAuth/OIDC login | `Bearer {signed_token}` (with prefix) |

Password login works because the frontend stores the raw signed token from the response header. OAuth login fails because the frontend reads the `auth` query parameter from the redirect URL (`/?auth={signed_token}`) and prepends `"Bearer "` before sending it back.

`URLSafeTimedSerializer.loads("Bearer ...")` cannot verify the signature of the mangled string, causing all authenticated requests after OAuth login to return 401.

## Fix

Strip the `"Bearer "` prefix from the `Authorization` header before passing it to `jwt.loads()`:

```python
token_value = authorization.split(" ", 1)[1] if authorization.startswith("Bearer ") or authorization.startswith("bearer ") else authorization
access_token = str(jwt.loads(token_value))
```

### Type of change

### DIFF
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -97,7 +97,10 @@ def _load_user():
     authorization = request.headers.get("Authorization")
     g.user = None
     if not authorization:
+        logging.debug("load_user: No Authorization header")
         return None
+
+    logging.debug(f"load_user: Authorization header format: starts_with_bearer={authorization.startswith('Bearer ')}, len={len(authorization)}")
 
     try:
         token_value = authorization.split(" ", 1)[1] if authorization.startswith("Bearer ") or authorization.startswith("bearer ") else authorization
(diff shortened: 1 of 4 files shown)


Lynn-Inf (commented):
The backend part has already been fixed by a previous PR (see PR #14154 ). However, the frontend changes in this PR may still help address the issue.

PR #14140, opened 2026-04-15
feat: add native Dutch language support for BM25 tokenization

rwb-truelime:
## Summary
- Add language-aware Snowball stemmer to `RagTokenizer` supporting 16 languages (Dutch, German, French, Spanish, etc.)
- Thread the KB `language` parameter through the full tokenization pipeline (14 parser modules + task executor)
- Add Dutch to the frontend language lists and cross-language form

## Problem
RAGFlow uses the English Porter stemmer + WordNet lemmatizer for **all** BM25 tokenization, regardless of the knowledge base language setting. This produces incorrect stems for non-English text. For example:

| Dutch word | Dutch stemmer | English Porter |
|---|---|---|
| documenten | document | documenten (unchanged!) |
| gebruikers | gebruiker | gebruik (over-stemmed) |
| instellingen | instell | instellingen (unchanged!) |

This degrades BM25 recall for any non-English knowledge base.

## Solution
NLTK already ships Snowball stemmers for 16 languages. This PR:

1. **`rag/nlp/rag_tokenizer.py`**: Overrides `tokenize()` with `set_language()` and `_normalize_token()` that selects the correct NLTK Snowball stemmer. Falls back to Porter for unmapped languages (Chinese, Japanese, Korean, etc. — these use character-based tokenization anyway).
2. **`rag/nlp/__init__.py`** + **14 `rag/app/*.py` parsers** + **`rag/svr/task_executor.py`**: Threads the `language` parameter through `tokenize()`, `tokenize_chunks()`, `tokenize_table()`, and all callers.
3. **Frontend**: Adds Dutch (`Nederlands`) to `LanguageList`, `LanguageMap`, `LanguageAbbreviationMap`, `LanguageTranslationMap`, cross-language form field, and `en.ts` locale.

## Backward Compatibility
- Default language is `"English"`, preserving existing behavior for all current users
- Languages without a Snowball stemmer mapping fall back to Porter (no change)
- No new dependencies — NLTK Snowball is already bundled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,12 @@ Cargo.lock
 
 .idea/
 .vscode/
+.opencode/
+docs/superpowers/
+eval_ragflow_agent.py
+eval_results.csv
+migrate_es_to_azure_search.py
+scripts/
 
 # Exclude Mac generated files
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ Cargo.lock
 .idea/
 .vscode/
 .cursor/settings.json
+.opencode/
 
 # Exclude Mac generated files
 .DS_Store
--- a/rag/app/audio.py
+++ b/rag/app/audio.py
@@ -51,7 +51,7 @@ def chunk(filename, binary, tenant_id, lang, callback=None, **kwargs):
         ans = seq2txt_mdl.transcription(tmp_path)
         callback(0.8, "Sequence2Txt LLM respond: %s ..." % ans[:32])
 
-        tokenize(doc, ans, is_english)
+        tokenize(doc, ans, is_english, language=lang)
         return [doc]
     except Exception as e:
         callback(prog=-1, msg=str(e))
--- a/rag/app/audio.py
+++ b/rag/app/audio.py
@@ -50,7 +50,7 @@ def chunk(filename, binary, tenant_id, lang, callback=None, **kwargs):
         ans = seq2txt_mdl.transcription(tmp_path)
         callback(0.8, "Sequence2Txt LLM respond: %s ..." % ans[:32])
 
-        tokenize(doc, ans, is_english)
+        tokenize(doc, ans, is_english, language=lang)
         return [doc]
     except Exception as e:
         callback(prog=-1, msg=str(e))
--- a/rag/app/book.py
+++ b/rag/app/book.py
@@ -173,8 +173,8 @@ def chunk(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, lang=
     # is_english(random_choices([t for t, _ in sections], k=218))
     eng = lang.lower() == "english"
 
-    res = tokenize_table(tbls, doc, eng)
-    res.extend(tokenize_chunks(chunks, doc, eng, pdf_parser))
+    res = tokenize_table(tbls, doc, eng, language=lang)
+    res.extend(tokenize_chunks(chunks, doc, eng, pdf_parser, language=lang))
     table_ctx = max(0, int(parser_config.get("table_context_size", 0) or 0))
     image_ctx = max(0, int(parser_config.get("image_context_size", 0) or 0))
     if table_ctx or image_ctx:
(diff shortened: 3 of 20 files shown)


rag/nlp/rag_tokenizer.py
qinling0210, line 1: Full rag_tokenizer.py is in Infinity (https://github.com/infiniflow/infinity/blob/main/python/infinity_sdk/infinity/rag_tokenizer.py), we import the file from infinity-sdk in ragflow. 

We should not  modify this file. I suggest you submit a PR to modify rag_tokenizer.py in Infinity.

After we release a new version of Infinity, we can merge this PR in ragflow

qinling0210 (commented):
There is also tokenization logic in C++ in Infinity (RAGAnalyzer::Tokenize, RAGAnalyzer::TokenizeWithPosition), we need to support Dutch too and make the tokenization results are the same between C++ and Python (unit test rag_analyzer_ut.cpp).

rwb-truelime:
Thanks for the review, @qinling0210! You're right that the C++ Infinity tokenizer (RAGAnalyzer) would need Dutch support too for full parity.

To be transparent about scope: this PR focuses on the Python-side tokenization path (BM25 via NLTK/Snowball), which is what most deployments use today. I don't have experience with the Infinity C++ codebase and wouldn't be able to confidently make changes there without risking regressions.

Would it be acceptable to merge this as Python-only Dutch support for now, and track the C++ Infinity side as a follow-up issue? I'm happy to create that issue with details on the expected tokenization behavior so someone familiar with the C++ side can pick it up. Alternatively, if C++ parity is a hard requirement before merge, I understand — just let me know.

I've also pushed a fix for the table separator bug flagged in the automated review — Latin-script languages like Dutch were incorrectly getting the CJK full-width separator (`；`) in table chunks.

rwb-truelime:
Regarding the thread-safety concerns raised by the automated review (`set_language()` mutating a global singleton):

The task executor spawns separate OS processes via `--workers=N` (gunicorn/uvicorn), so each process has its own tokenizer singleton — no actual race condition in practice. The global state pattern (`eng` flag) predates this PR. Refactoring to per-call tokenizer instances would be a worthwhile improvement but is out of scope for this feature PR.

qinling0210:
@rwb-truelime 

OK, we will handle tokenization logic in C++ in Infinity.

You do not need to modify C++ codebase in Infinity.
 Just apply your code change to https://github.com/infiniflow/infinity/blob/main/python/infinity_sdk/infinity/rag_tokenizer.py. 

For rag/nlp/rag_toknizer.py in Ragflow, we can import the code from Infinity.

rwb-truelime:
@qinling0210 Thanks for the guidance! I've submitted the tokenizer changes to the Infinity SDK:

**infiniflow/infinity#3356** — adds `set_language()` to `RagTokenizer` with Snowball stemmer support for 17 languages (Dutch, German, French, Spanish, etc.)

The changes are minimal and backward-compatible — default English behavior is unchanged, and callers that never call `set_language()` get identical results.

Once that's merged and released, I'll update this RAGFlow PR to:
1. Remove the local `rag/nlp/rag_tokenizer.py` modifications (they'll come from the updated infinity-sdk)
2. Keep the language parameter threading through `rag/nlp/__init__.py`, `rag/app/*.py`, `rag/svr/task_executor.py`
3. Keep the frontend changes (Dutch in language dropdowns)
4. Keep the table separator bugfix (`rag/nlp/__init__.py` — Latin-script languages were getting CJK separator)

Let me know if anything needs adjusting on the Infinity PR side.

qinling0210:
@rwb-truelime , We have released Infinity v0.7.0-dev6 and bumped infinity-sdk in Ragflow to v0.7.0-dev6, please update your RAGFlow PR.

rwb-truelime:
Updated the PR as requested:

- **Removed** the local `rag/nlp/rag_tokenizer.py` override entirely — the tokenizer now inherits `set_language()` and `_normalize_token()` from infinity-sdk v0.7.0-dev6's `RagTokenizer` parent class
- **Rebased** onto latest main (includes the infinity-sdk bump to dev6)
- **Kept** the language parameter threading through `rag/nlp/__init__.py`, all parser modules, and `rag/svr/task_executor.py`
- **Kept** the table separator fix (Latin-script languages now get `; ` instead of CJK `；`)
- **Kept** the frontend Dutch language dropdown entries

The diff is now ~75 lines across 20 files — all just wiring the language parameter through. The actual stemming logic lives in infinity-sdk.

yingfeng:
Conflict arises here.

yingfeng:
@copilot resolve the merge conflicts in this pull request

rwb-truelime:
🎉

PR #14164, opened 2026-04-16
fix(prompt): reserve system budget in message_fit_in

### LINKED ISSUE #13607: [Bug]: message_fit_in truncates user message with wrong variable, causing context length overflow
### Self Checks

- [x] Non-english title submitions will be closed directly ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).

Latest main branch

### RAGFlow image version

nightly

```
N/A - code-level bug found by reading the source
```

### Actual behavior

In `rag/prompts/generator.py`, the `message_fit_in()` function has a bug on line 98 in its else branch (when the user message is the dominant portion and needs truncation).

The code currently reads:

```python
ll = num_tokens_from_string(msg_[0]["content"])   # system prompt tokens
ll2 = num_tokens_from_string(msg_[-1]["content"])  # user message tokens
if ll / (ll + ll2) > 0.8:
    # system prompt is large: truncate system, leaving room for user
    m = encoder.decode(encoder.encode(m)[: max_length - ll2])  # correct
    ...
# else: user message is large
m = msg_[-1]["content"]
m = encoder.decode(encoder.encode(m)[: max_length - ll2])  # BUG: should be max_length - ll
```

The else branch truncates the user message to `max_length - ll2` tokens, but `ll2` is the token count of the user message itself. It should subtract `ll` (the system prompt token count) to leave room for the system prompt.

**Consequences:**
- If `ll2 > max_length`, the slice index becomes negative, causing the encoder to take the *last* N tokens instead of the first N (Python negative slice behavior)
- The system prompt's token count is never accounted for, so the total can still exceed `max_length`, leading to "maximum context length" errors from the LLM API

Additionally, if both messages are empty (`ll + ll2 == 0`), line 91 causes a `ZeroDivisionError`.

### Expected behavior

The user message should be truncated to `max_length - ll` tokens (leaving room for the system prompt), symmetric with how the if-branch truncates the system prompt to `max_length - ll2` (leaving room for the user message).

### Steps to reproduce

```
1. Set up a chat assistant with a model that has a 16K context window
2. Send a query that retrieves many chunks, producing a large user message
3. The message_fit_in function is called but incorrectly truncates
4. The LLM API returns "maximum context length exceeded" error
```

### Additional information

This bug is likely a root cause (or contributing factor) for issues #13549 and #13082 where users report context length overflow despite having small chunk sizes and low topN settings.

hyl64:
## Summary
This PR fixes the `message_fit_in()` truncation bug reported in #13607.

Changes:
- fix the user-message truncation branch to reserve room for the system prompt token budget
- guard the zero-token edge case to avoid dividing by zero in the truncation ratio check
- add focused regression tests covering both the user-dominant truncation path and the zero-token boundary case

## Validation
```bash
pytest -q --noconftest test/unit_test/rag/prompts/test_generator_message_fit_in.py
```

Result: `2 passed`

Closes #13607

### DIFF
--- a/rag/prompts/generator.py
+++ b/rag/prompts/generator.py
@@ -75,6 +75,10 @@ def message_fit_in(msg, max_length=4000):
             total += m["count"]
         return total
 
+    def trim_content(content, limit):
+        limit = max(0, limit)
+        return encoder.decode(encoder.encode(content)[:limit])
+
     c = count()
     if c < max_length:
         return c, msg
@@ -101,18 +105,22 @@ def message_fit_in(msg, max_length=4000):
         )
         return 0, msg
 
-    if ll / total > 0.8:
-        m = msg_[0]["content"]
-        preserve_len = max(0, max_length - ll2)
-        m = encoder.decode(encoder.encode(m)[:preserve_len])
-        msg[0]["content"] = m
-        return max_length, msg
+    if len(msg) == 1:
+        msg[0]["content"] = trim_content(msg[0]["content"], max_length)
+        return count(), msg
 
-    m = msg_[-1]["content"]
-    preserve_len_last = max(0, max_length - ll)
-    m = encoder.decode(encoder.encode(m)[:preserve_len_last])
-    msg[-1]["content"] = m
-    return max_length, msg
+    if ll / total > 0.8:
+        preserved_last = min(ll2, max_length)
+        msg[-1]["content"] = trim_content(msg_[-1]["content"], preserved_last)
+        remaining = max(0, max_length - preserved_last)
+        msg[0]["content"] = trim_content(msg_[0]["content"], remaining)
+        return count(), msg
+
+    preserved_system = min(ll, max_length)
+    msg[0]["content"] = trim_content(msg_[0]["content"], preserved_system)
+    remaining = max(0, max_length - preserved_system)
+    msg[-1]["content"] = trim_content(msg_[-1]["content"], remaining)
+    return count(), msg
 
 
 def kb_prompt(kbinfos, max_tokens, hash_id=False):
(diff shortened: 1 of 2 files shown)


hyl64:
Addressed the CodeRabbit suggestions in a narrow follow-up:

- added a debug log for the degenerate `total <= 0` path in `message_fit_in`
- clamped preserve lengths to avoid Python negative-slice behavior when `max_length` is smaller than the preserved counterpart
- added a regression test for the negative-slice case

Validation:
- `pytest -q --noconftest test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `3 passed`

hyl64:
Follow-up for the remaining over-budget edge case:

- capped the preserved side itself when either the system message or the trailing message alone already exceeds `max_length`
- changed `message_fit_in` to return the actual used token count after trimming instead of always returning `max_length`
- added regression coverage for the dominant-last-message case

Validation:
- `pytest -q --noconftest test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `4 passed`

hyl64:
@KevinHuSh @JinHai-CN @TeslaZY @Lynn-Inf could you help take a look at this bugfix PR when convenient?

This one fixes `#13607` in `rag/prompts/generator.py` and I've already addressed the follow-up CodeRabbit points around over-budget trimming and actual used-token counting.

Local validation:
- `pytest -q --noconftest test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `4 passed`

hyl64:
@KevinHuSh @JinHai-CN @TeslaZY quick nudge on this bugfix PR when convenient.

This is the fix for `#13607`. I addressed the follow-up CodeRabbit concerns and revalidated the focused regression tests locally.

PR #14209, opened 2026-04-18
Fix memory resolution regression for multimodal Gemini models

### LINKED ISSUE #14206: [Bug]: Python SDK - Memory Message Processing Fails Due to Incorrect model_type in llm_factories.json
### Self Checks

62cb29263

### RAGFlow image version

0

```Markdown

```

### Actual behavior

Several Gemini models in conf/llm_factories.json have an incorrect model_type value of "image2text" instead of "chat". This breaks the Python SDK's ability to create memories that can be processed, as the system now strictly filters by model_type when looking up tenant LLM configurations. 

Commit that introduced the regression: 62cb29263 (Feat/tenant model #13072). This commit added strict model_type filtering in TenantLLMService.get_api_key() (api/db/services/tenant_llm_service.py:43-44):

```
if model_type_val is not None:
    query_kwargs["model_type"] = model_type_val
```

Previously, the lookup was performed WITHOUT filtering by model_type, so even records with incorrect / missing model_type values would be found. 

In conf/llm_factories.json, several Gemini models have the wrong model_type: 

```json
{
    "llm_name": "gemini-2.5-flash", 
    "tags": "LLM,CHAT,1024K,IMAGE2TEXT",
     "model_type": "image2text"
 }
```    
**Impact on Python SDK**

When creating a memory via Python SDK:

1. SDK sends llm_id="gemini-2.5-flash@Gemini" and embd_id="gemini-embedding-001@Gemini"
2. ensure_tenant_model_id_for_params() calls get_api_key() with strict filtering by model_type='chat'
3. Query: WHERE tenant_id=? AND llm_name='gemini-2.5-flash' AND model_type='chat'
4. No match found because the record has model_type='image2text' in the database activate as files are read
5. Falls back to tenant_llm_id = 0 ? stored as NULL
6. Messages processing fails with: "Tenant Model with name gemini-2.5-flash@Gemini and type chat not found"

Image

**Files Affected**
- conf/llm_factories.json - Contains incorrect model_type values
- api/db/services/tenant_llm_service.py:43-44 - Strict filtering introduced in commit 62cb29263
- api/utils/tenant_utils.py:28-37 - Uses the strict filtering

### Expected behavior

Messages for memories created via the Python SDK should be properly processed.

### Steps to reproduce

```Markdown
1. Create a memory using the https://ragflow.io/docs/python_api_reference#create-memory
2. Add a message using https://ragflow.io/docs/python_api_reference#add-message
3. Check the message processing progress in the UI within the created memory
```

### Additional information

**Immediate Fix**
Change model_type from "image2text" to "chat" for these models in conf/llm_factories.json.

**Long-term Fix**
The model_type field cannot properly represent multimodal models that support multiple task types (e.g., Gemini models handle both chat AND image2text). The system needs to support multiple model_type values per model.

**Workaround**
Modify TenantLLMService.get_api_key() to fallback to non-model_type filtering when the strict lookup fails, matching the behavior before commit 62cb29263

spider-yamet:

Fixes #14206.

This issue is a regression. PR #9520 previously changed Gemini models from `image2text` to `chat` to fix chat-side resolution, but PR #13073 later restored those Gemini entries to `image2text` during model-list updates, which reintroduced the bug.

The underlying problem is that Gemini models are multimodal and advertise both `CHAT` and `IMAGE2TEXT`, while tenant model resolution still depends on a single stored `model_type`. That makes chat-only flows such as memory extraction fragile when a compatible model is stored as `image2text`.

This PR fixes the issue at the model resolution layer instead of changing `llm_factories.json` again:
- keep the stored tenant model type unchanged
- try exact `model_type` lookup first
- if no exact match is found, fall back only when the model metadata shows the requested capability is supported
- coerce the runtime config to the requested type for chat callers
- fail fast in memory creation instead of silently persisting `tenant_llm_id=0`

This preserves existing multimodal and `image2text` behavior while restoring chat compatibility for memory-related flows.

### Type of change

### Testing

- Re-checked the current memory creation and memory message extraction paths against the updated resolution logic
- Verified locally that a Gemini-style tenant model stored as `image2text` but tagged with `CHAT` can still be resolved for `chat`
- Verified `get_model_config_by_type_and_name(..., CHAT, ...)` returns a chat-compatible runtime config
- Verified `get_model_config_by_id(..., CHAT)` also returns a chat-compatible runtime config
- Verified strict resolution still fails when the model metadata does not advertise chat capability

### DIFF
--- a/api/apps/restful_apis/memory_api.py
+++ b/api/apps/restful_apis/memory_api.py
@@ -33,9 +33,9 @@ async def create_memory():
     timing_enabled = os.getenv("RAGFLOW_API_TIMING")
     t_start = time.perf_counter() if timing_enabled else None
     req = await get_request_json()
-    req = ensure_tenant_model_id_for_params(current_user.id, req, strict=True)
     t_parsed = time.perf_counter() if timing_enabled else None
     try:
+        req = ensure_tenant_model_id_for_params(current_user.id, req, strict=True)
         memory_info = {
             "name": req["name"],
             "memory_type": req["memory_type"],
--- a/api/apps/restful_apis/memory_api.py
+++ b/api/apps/restful_apis/memory_api.py
@@ -18,7 +18,7 @@ import os
 import time
 
 from quart import request
-from common.constants import RetCode
+from common.constants import LLMType, RetCode
 from common.exceptions import ArgumentException, NotFoundException
 from api.apps import login_required, current_user
 from api.utils.api_utils import validate_request, get_request_json, get_error_argument_result, get_json_result
@@ -35,7 +35,11 @@ async def create_memory():
     req = await get_request_json()
     t_parsed = time.perf_counter() if timing_enabled else None
     try:
-        req = ensure_tenant_model_id_for_params(current_user.id, req, strict=True)
+        req = ensure_tenant_model_id_for_params(current_user.id, req)
+        if not req.get("tenant_llm_id"):
+            raise ArgumentException(
+                f"Tenant Model with name {req['llm_id']} and type {LLMType.CHAT.value} not found"
+            )
         memory_info = {
             "name": req["name"],
             "memory_type": req["memory_type"],
--- a/api/apps/restful_apis/memory_api.py
+++ b/api/apps/restful_apis/memory_api.py
@@ -33,7 +33,7 @@ async def create_memory():
     timing_enabled = os.getenv("RAGFLOW_API_TIMING")
     t_start = time.perf_counter() if timing_enabled else None
     req = await get_request_json()
-    req = ensure_tenant_model_id_for_params(current_user.id, req)
+    req = ensure_tenant_model_id_for_params(current_user.id, req, strict=True)
     t_parsed = time.perf_counter() if timing_enabled else None
     try:
         memory_info = {
--- a/api/db/joint_services/memory_message_service.py
+++ b/api/db/joint_services/memory_message_service.py
@@ -154,7 +154,7 @@ async def extract_by_llm(tenant_id: str, tenant_llm_id: int, extract_conf: dict,
     else:
         user_prompts.append({"role": "user", "content": PromptAssembler.assemble_user_prompt(conversation_content, conversation_time, conversation_time)})
     if tenant_llm_id:
-        llm_config = get_model_config_by_id(tenant_llm_id, LLMType.CHAT)
+        llm_config = get_model_config_by_id(tenant_llm_id)
     else:
         llm_config = get_model_config_by_type_and_name(tenant_id, LLMType.CHAT, llm_id)
     llm = LLMBundle(tenant_id, llm_config)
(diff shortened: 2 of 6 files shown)


spider-yamet:
@yingfeng Could you please assign the review of this PR?

6ba3i:
Thanks for working on this @spider-yamet! I agree that the PR fixes the regression in #14206.

That said, I think this specific bug can be fixed in a smaller and more targeted way.

The failure starts during memory creation, in `ensure_tenant_model_id_for_params(...)`. For `llm_id`, we resolve the tenant model as `LLMType.CHAT`. That lookup is strict, so it misses multimodal Gemini rows that are stored as `image2text`. Once that happens, memory creation persists an invalid `tenant_llm_id`, and the later memory extraction path fails because it no longer has a valid tenant model id to use.

Because the bad state is introduced there, I think the narrowest fix is to handle it there:

- keep the current typed lookup first
- only for `LLMType.CHAT`, if that lookup misses, retry once without `model_type`
- if the untyped lookup finds the existing tenant row, persist that id and leave the rest of the flow unchanged

That solves the issue at the exact failure point, with a very small patch. It also avoids broadening shared lookup behavior in other layers, and it leaves existing `image2text` / PDF parsing behavior alone.

Suggested change:

```diff
diff --git a/api/utils/tenant_utils.py b/api/utils/tenant_utils.py
index 83da91f1c..2d1baf66a 100644
--- a/api/utils/tenant_utils.py
+++ b/api/utils/tenant_utils.py
@@ -30,6 +30,8 @@ def ensure_tenant_model_id_for_params(tenant_id: str, param_dict: dict) -> dict:
         if param_dict.get(key) and not param_dict.get(f"tenant_{key}"):
             model_type = _KEY_TO_MODEL_TYPE.get(key)
             tenant_model = TenantLLMService.get_api_key(tenant_id, param_dict[key], model_type)
+            if not tenant_model and model_type == LLMType.CHAT:
+                tenant_model = TenantLLMService.get_api_key(tenant_id, param_dict[key])
             if tenant_model:
                 param_dict.update({f"tenant_{key}": tenant_model.id})
             else:
```
I think this is a better fit for #14206 because it fixes the regression where it starts, with the fewest line changes possible.

spider-yamet:
@6ba3i Thanks for your detailed feedback. I updated the fix to match your suggestion and kept it as narrow as possible. Instead of changing the shared model resolution behavior, I only handle it at memory creation time: we still try the typed CHAT lookup first, and if that misses, we retry once without model_type just for llm_id, so the existing multimodal tenant row can still be reused.

I also fixed the error handling in memory_api.create_memory(), so if strict model resolution fails, it now goes through the existing argument-error path instead of surfacing as a 500. Other broader compatibility changes were removed so the patch stays focused on the exact regression point you called out.

I would appreciate your feedback now. :)

6ba3i (changes requested):
i think your addition of the strict rule might've broken the ci @spider-yamet especially because it only happens in the memory tests in sdk, do you mind trying and removing them or fixing the ci errors ? Thanks !

spider-yamet:
@6ba3i Could you please check again? pr is passing the ci :)

6ba3i (approved):
LGTM!

spider-yamet:
@yingfeng could you please share your opinion?

PR #14217, opened 2026-04-19
fix(agent): handle duplicate MCP tool names

eviaaaaa:

When multiple MCP servers expose tools with the same name, the agent currently registers those tools using their original MCP names. This can lead to two issues:

- later MCP tools may overwrite earlier ones in the agent tool map
- duplicate function names may be exposed to the LLM

This PR fixes duplicate MCP tool-name handling by applying the same indexed naming strategy already used for native agent tools. Native tools are exposed with generated names such as `<tool_name>_<index>` to avoid collisions, and MCP tools now follow the same convention for consistency.

Specifically, this PR:

- assigns unique indexed function names to MCP tools exposed to the LLM
- preserves each MCP tool's original server-side name in an `MCPToolBinding`
- dispatches MCP calls using the original MCP tool name while keeping the indexed name in the agent tool map
- allows MCP metadata conversion to override only the OpenAI function name without modifying the original MCP tool metadata

### Type of change

### Validation

The validation was performed using two MCP servers. Both servers exposed a tool with the same name: `mcp0`. Both tools take no input parameters.

**MCP Server One:**
ONE

**MCP Server Two:**
Second

**Before the fix:**
When invoking `mcp0`, only the `mcp0` tool from the MCP server injected later could be called successfully. As shown below, both `mcp0` tools were present, but only the later-registered one was actually invokable.

Three

**After the fix:**
Both `mcp0` tools can now be invoked correctly.

F

six

### DIFF
--- a/agent/component/agent_with_tools.py
+++ b/agent/component/agent_with_tools.py
@@ -32,7 +32,7 @@ from api.db.services.llm_service import LLMBundle
 from api.db.services.mcp_server_service import MCPServerService
 from api.db.services.tenant_llm_service import TenantLLMService
 from common.connection_utils import timeout
-from common.mcp_tool_call_conn import MCPToolCallSession, mcp_tool_metadata_to_openai_tool
+from common.mcp_tool_call_conn import MCPToolBinding, MCPToolCallSession, mcp_tool_metadata_to_openai_tool
 from rag.prompts.generator import citation_plus, citation_prompt, full_question, kb_prompt, message_fit_in, structured_output_prompt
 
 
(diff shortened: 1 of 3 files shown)


eviaaaaa:
Hi maintainers, just wanted to gently follow up on this PR.

This PR fixes an MCP tool-name collision issue: when multiple MCP servers have tools with the same name, the current logic may overwrite previously registered tools.

Could @JinHai-CN  please take a look when convenient? Thanks!

KevinHuSh:
Appreciations!

Fix CI failure please.
```
Error: agent/tools/base.py:58:75: ASYNC109 Async function definition with a `timeout` parameter
  help: Use `asyncio.timeout` instead
  ```

PR #14259, opened 2026-04-21
Docs: User-level memory is supported in v0.25.0

writinwaters:

v0.25.0 supports linking User ID with conversations.

### Type of change

- [x] Documentation Update


Lynn-Inf (approved):
LGTM

PR #14320, opened 2026-04-23
Fix llm_id validation for image2text Gemini usage.

lokinhh:
Allow chat assistant llm_id validation to accept image2text models and resolve chat model config using actual llm type so Gemini vision models no longer fail with misleading doesn't-exist errors.

Made-with: Cursor

RAGFlow currently validates chat assistant llm_id too strictly against model_type=chat, causing code 102 ("llm_id ... doesn't exist") when users configure Gemini as image2text. This PR allows llm_id validation to accept both chat and image2text, and resolves runtime model config based on the actual llm type, so Gemini vision models can be used without misleading validation errors.

### Type of change


JinHai-CN:
@lokinhh I will close this PR, if you are still working on this PR, please re-open it.

PR #14331, opened 2026-04-23
Fix: allow use image2text as chat model

Lynn-Inf:

Allow image2text models (multimodal) to be used as chat models.

### Type of change


Magicbook1108 (approved):
LGTM

PR #14339, opened 2026-04-23
Fix: Blob storage connector downloads unsupported file types

### LINKED ISSUE #14338: [Bug]: Blob storage datasets download unsupported file types instead of filtering them
### Self Checks

a33d0737

### RAGFlow image version

v0.25.0

```Markdown
- Storage backend: S3 / S3-compatible storage
- Deployment: Docker / Docker Compose
- Others: The bucket or prefix contains both supported document files and unsupported files.
```

### Actual behavior

When syncing a Blob storage dataset, RAGFlow lists objects from the bucket/prefix and downloads files even if their file extensions are not supported by the ingestion pipeline.

Unsupported files are not filtered before download. This causes unnecessary object downloads, slower synchronization, extra bandwidth usage, and confusing sync behavior when a bucket contains mixed file types.

### Expected behavior

Blob storage datasets should skip unsupported file types before downloading objects.

The connector should only download files with supported extensions. Unsupported files, image files when image ingestion is not enabled, unknown extensions, and files without extensions should be skipped during object listing/collection.

### Steps to reproduce

```Markdown
1. Configure a Blob storage dataset, such as S3 or S3-compatible storage.
2. Use a bucket or prefix that contains both supported files and unsupported files.
3. Example supported files: `guide.pdf`, `notes.md`.
4. Example unsupported files: `archive.zip`, `image.png`, `data.parquet`, `binary.bin`, `README` without extension.
5. Start dataset synchronization.
6. Observe that unsupported objects are still considered by the Blob storage connector and may be downloaded before being rejected later.
```

### Additional information

The existing helper `is_accepted_file_ext` in `common/data_source/utils.py` already defines supported file extensions.

A possible fix is to add extension validation in `BlobStorageConnector` before calling the object download API, so unsupported files are skipped early.

spider-yamet:

Fixes #14338.

`BlobStorageConnector._yield_blob_objects` currently enumerates every object returned by `list_objects_v2` and calls `download_object` on it, regardless of file extension. Unsupported types (e.g. `.zip`, `.exe`, `.mp4`, archive files)
are only rejected much later in the pipeline, after they have been fully downloaded from S3/MinIO/R2/GCS and instantiated as `Document` objects.

Consequences reported in the issue:

- Wasted bandwidth and storage egress cost on every sync.
- Long sync times on buckets containing many unsupported binaries.
- Memory pressure for large unsupported objects.
- Noisy logs where every unsupported file is downloaded only to be discarded.

All other connectors in `common/data_source/` (e.g. `WebDAVConnector`, and the pattern added broadly in #13802 "Add file type validation") perform the extension check **before** fetching the object, using `get_file_ext` + `is_accepted_file_ext(..., OnyxExtensionType.*)`. The blob connector was the remaining outlier.

### Root cause

`_yield_blob_objects` had no extension guard; the only filter on listing was the `LastModified` window and the trailing-`/` directory-marker check.

### Fix

Align `BlobStorageConnector` with the pattern used by the other connectors:

- Add `_build_extension_type()` and `_is_supported_file(file_name)` helpers  on `BlobStorageConnector`, mirroring `WebDAVConnector`.
- In `_yield_blob_objects`, check each object `Key` against  `_is_supported_file()` **before** it is appended to the download batch, and  emit an aggregate `INFO` log with the skip count per bucket for  observability.
- Keep a second defense-in-depth extension check just before  `download_object`, so direct callers of the per-object path cannot regress  the behavior.
- Respect `allow_images`: multimedia extensions are only accepted when  `set_allow_images(True)` has been called, matching the semantics of the  other connectors.

### DIFF
--- a/common/data_source/blob_connector.py
+++ b/common/data_source/blob_connector.py
@@ -149,14 +149,15 @@ class BlobStorageConnector(LoadConnector, PollConnector):
                 key = obj["Key"]
                 if key.endswith("/"):
                     continue
+                last_modified = obj["LastModified"].replace(tzinfo=timezone.utc)
+                if not (start < last_modified <= end):
+                    continue
                 file_name = os.path.basename(key)
                 if not self._is_supported_file(file_name):
                     skipped_unsupported += 1
                     logging.debug(f"Skipping '{key}' due to unsupported file extension.")
                     continue
-                last_modified = obj["LastModified"].replace(tzinfo=timezone.utc)
-                if start < last_modified <= end:
-                    all_objects.append(obj)
+                all_objects.append(obj)
 
         if skipped_unsupported:
             logging.info(
--- a/common/data_source/blob_connector.py
+++ b/common/data_source/blob_connector.py
@@ -10,6 +10,7 @@ from common.data_source.utils import (
     download_object,
     extract_size_bytes,
     get_file_ext,
+    is_accepted_file_ext,
 )
 from common.data_source.config import BlobType, DocumentSource, BLOB_STORAGE_SIZE_THRESHOLD, INDEX_BATCH_SIZE
 from common.data_source.exceptions import (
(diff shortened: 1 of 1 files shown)


spider-yamet:
@yingfeng could you please assign the review of this PR? :)

6ba3i:
hello @spider-yamet ! Unfortunately for this pr it was something i was already working on and your changes are not really aligned with what i was going for, thank you for your contribution though !

PR #14347, opened 2026-04-24
fix: add bucket prefix to Azure Blob SPN and SAS storage operations

### LINKED ISSUE #14159: Azure Blob SPN storage: files from different datasets can overwrite each other due to missing bucket prefix
## Bug Description

When using Azure Blob storage with SPN authentication, files from different datasets can overwrite each other if they share the same filename. This is because the `put()` method in `azure_spn_conn.py` ignores the `bucket` parameter entirely and stores files flat using only the filename:

```python
f = self.conn.create_file(fnm)
```

## Expected Behavior

In contrast, the MinIO and S3 implementations prepend the bucket (typically the knowledge base ID) as a path prefix, creating logical folder isolation like `{kb_id}/{filename}`. This prevents filename collisions between files across different datasets.

## Impact

With Azure SPN as-is, files with the same name from different datasets will overwrite each other in Azure Blob storage.

## Workaround

Modify `azure_spn_conn.py` to prepend the bucket parameter to the file path. For example, change the `put()` method from:

```python
f = self.conn.create_file(fnm)
```

to:

```python
f = self.conn.create_file(f"{bucket}/{fnm}")
```

Apply the same `bucket/` prefix pattern to `get()`, `rm()`, `obj_exist()`, and `get_presigned_url()` as well so all operations use consistent paths.

## Additional Note

The same issue likely exists in the Azure SAS implementation (`azure_sas_conn.py`), so that file should also be checked if Azure SAS authentication is planned for use.

D2758695161:
## Summary

Fixes file collision between different datasets when using Azure Blob storage (SPN or SAS authentication).

## Bug

azure_spn_conn.py and zure_sas_conn.py ignored the ucket parameter entirely, storing all files flat with just the filename. This caused files with the same name from different datasets (knowledge bases) to overwrite each other.

## Fix

Prepend bucket/ as a path prefix in all methods (put, 
m, get, obj_exist, get_presigned_url, health) to match the behavior of MinIO and S3 implementations.

## Changes

- **rag/utils/azure_spn_conn.py**: Added {bucket}/ prefix to file paths in all operations
- **rag/utils/azure_sas_conn.py**: Same fix applied for consistency (also noted in the original issue)

## Testing

Manual verification: files from different datasets now stored under distinct bucket/ prefixes, preventing collisions.

Fixes #14159


Magicbook1108:
Can you provide screen shots to proof your modification valid ?

rag/utils/azure_spn_conn.py
```diff
@@ -110,7 +110,8 @@ def obj_exist(self, bucket, fnm):
     def get_presigned_url(self, bucket, fnm, expires):
         for _ in range(10):
             try:
-                return self.conn.get_presigned_url("GET", bucket, fnm, expires)
```
JinHai-CN: Why change it to self.conn.get_file_client?

rag/utils/azure_sas_conn.py
```diff
@@ -49,47 +49,47 @@ def __close__(self):
...
-            return self.conn.get_blob_client(fnm).exists()
+            return self.conn.get_blob_client(f"{bucket}/{fnm}").exists()
         except Exception:
             logging.exception(f"Fail put {bucket}/{fnm}")
         return False
 
     def get_presigned_url(self, bucket, fnm, expires):
         for _ in range(10):
             try:
-                return self.conn.get_presigned_url("GET", bucket, fnm, expires)
```
JinHai-CN: Why change it to self.conn.get_blob_client(f"{bucket}/{fnm}").url?

Magicbook1108 (approved):
LGTM

PR #14357, opened 2026-04-24
Fix blob sync: skip unsupported files before download

### LINKED ISSUE #14338: [Bug]: Blob storage datasets download unsupported file types instead of filtering them

6ba3i:

Blob storage sync was downloading unsupported files first and rejecting them later, which wasted bandwidth and made sync slower. This PR skips unsupported extensions before download and applies `allow_images` in blob sync. fixes #14338

### Type of change

### DIFF
--- a/common/data_source/blob_connector.py
+++ b/common/data_source/blob_connector.py
@@ -10,6 +10,7 @@ from common.data_source.utils import (
     download_object,
     extract_size_bytes,
     get_file_ext,
+    is_accepted_file_ext,
 )
 from common.data_source.config import BlobType, DocumentSource, BLOB_STORAGE_SIZE_THRESHOLD, INDEX_BATCH_SIZE
 from common.data_source.exceptions import (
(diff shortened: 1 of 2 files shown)


Magicbook1108 (approved):
LGTM

Issue #14371, opened 2026-04-25
[Bug]:  File parsing fails after upgrade from v0.24.0 to v0.25.0 - KeyError: 'properties' in metadata generation


LijiaTong:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

NA

### RAGFlow image version

0.25.0

### Other environment information

```Markdown
RAGFlow Version: 0.25.0 (upgraded from 0.24.0)
Python Version: 3.12.12
OS: Linux (x86_64)
```

### Actual behavior

After upgrading RAGFlow from version 0.24.0 to 0.25.0, file parsing tasks are failing with a KeyError when attempting to generate metadata. The error occurs in the gen_metadata function when trying to access the properties key in the schema dictionary.

### Expected behavior

File parsing should complete successfully after upgrade, with metadata generation working as it did in version 0.24.0.

### Steps to reproduce

```Markdown
Install RAGFlow version 0.24.0
Process files successfully to confirm working state
Upgrade to RAGFlow version 0.25.0
Attempt to parse/upload files
Observe the error in task execution logs
```

### Additional information

Traceback (most recent call last):
  File "/ragflow/rag/svr/task_executor.py", line 1285, in handle_task
    await do_handle_task(task)
  File "/ragflow/common/connection_utils.py", line 74, in async_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ragflow/rag/svr/task_executor.py", line 1182, in do_handle_task
    chunks = await build_chunks(task, progress_callback)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ragflow/common/connection_utils.py", line 74, in async_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ragflow/rag/svr/task_executor.py", line 437, in build_chunks
[...]

PR #14383, opened 2026-04-27
Fix metadata parsing regression for upgraded v0.24 datasets

6ba3i:

This PR fixes issue #14371 where file parsing failed after upgrading from v0.24.0 to v0.25.0, because metadata config could be a JSON Schema object but was handled like a list and later caused `KeyError: 'properties'`.  

### Type of change


LijiaTong:
Fix confirmed – metadata parsing & deletion work perfectly

Your patch works great. After rebuilding the image with your fix, files can now be parsed successfully, metadata displays correctly, and deletion also works as expected.

One small note for accuracy: the original async def gen_metadata_task function in /ragflow/rag/svr/task_executor.py is located around lines 429–444 (not exactly 417--429). But the fix itself is solid.

Thank you for your prompt help!

PR #14405, opened 2026-04-27
feat: increase chatbot top n limit to 200

Attili-sys:

frontend explicitly capped the input at 30 in the shared TopNFormField component default props (web/src/components/top-n-item.tsx) and this was blocking me from sending more chunks to the llm

Updated web/src/components/top-n-item.tsx to set the default max value of the slider to 200 instead of 30, seamlessly enabling the configuration of higher values up to 200 for the Top N field in the chatbot settings and other relevant agent forms.

- [x] Other (frontend explicitly capped the input at 30 in the shared TopNFormField component default props (web/src/components/top-n-item.tsx) so I changed that to 200):

### DIFF
--- a/web/src/components/top-n-item.tsx
+++ b/web/src/components/top-n-item.tsx
@@ -11,7 +11,7 @@ export const topnSchema = {
   top_n: z.number().optional(),
 };
 
-export function TopNFormField({ max = 30 }: SimilaritySliderFormFieldProps) {
+export function TopNFormField({ max = 200 }: SimilaritySliderFormFieldProps) {
   const { t } = useTranslate('chat');
 
   return (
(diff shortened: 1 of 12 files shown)


web/src/components/top-n-item.tsx
```diff
@@ -7,11 +7,13 @@ interface SimilaritySliderFormFieldProps {
   max?: number;
 }
 
+const TOP_N_MAX = 200;
+
 export const topnSchema = {
-  top_n: z.number().optional(),
+  top_n: z.number().max(TOP_N_MAX).optional(),
```
cike8899: If the maximum value passed by the user when calling `TopNFormField` is not 200, it will conflict with the limit defined here.

cike8899:
@Attili-sys Please resolve the conflict.

PR #14406, opened 2026-04-27
Fix: clicks on Popover-based selects swallowed when nested in Dialog

JimZhang-lab:

Radix Dialog sets `body { pointer-events: none }` while open and
relies on the nested Popover's DismissableLayer to restore
`pointer-events: auto`. Two issues in `web/src/components/ui/popover.tsx`
broke that contract:

- The Popover wrapper mirrored `open` into a local `useState(true)`
  plus a sync effect, briefly mounting and unmounting PopoverContent
  on first render and racing the layer order. The popover ended up
  with `pointer-events: none`.
- PopoverContent had no explicit `pointer-events-auto` fallback.

The popover stayed visible but transparent to clicks; clicks fell through to the DialogHeader behind it, which Radix interpreted as an outside click and dismissed the popover. Affected the chunk method in the Create dataset dialog  and the model type dropdown in Model Providers, among others.

Fixes:
- Make `Popover` in `web/src/components/ui/popover.tsx` a thin pass-through over `PopoverPrimitive.Root`.
- Add `pointer-events-auto` to `PopoverContent`.
- Fix typo `zoom-in-9` -> `zoom-in-95` in the open animation class.

### Type of change

### DIFF
--- a/web/src/components/ui/popover.tsx
+++ b/web/src/components/ui/popover.tsx
@@ -15,25 +15,17 @@ const Popover = ({
   onOpenChange,
   disableOutsideClick = false,
 }: PopoverProps) => {
-  const [open, setOpen] = React.useState(true);
-  React.useEffect(() => {
-    setOpen(!!openState);
-  }, [openState]);
   const handleOnOpenChange = React.useCallback(
     (e: boolean) => {
       if (disableOutsideClick && !e) {
         return;
       }
-
-      if (onOpenChange) {
-        onOpenChange?.(e);
-      }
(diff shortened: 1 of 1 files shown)


cike8899:
image

I haven't encountered this issue when using a combination of Dialogs and Popovers in my projects—for instance, in the example shown in the image. You can directly use `SelectWithSearch` in your own project. Thank you for your contribution. @JimZhang-lab

KevinHuSh:
Appreciations! 
But long time no response.

PR #14416, opened 2026-04-28
Fix: preserve infinity available_int zero filter

buua436:

preserve infinity available_int zero filter

### Type of change


wangq8 (approved):
LGTM

PR #14417, opened 2026-04-28
Always return success if no such task id

wangq8:

Always return success if no such task id to follow existing code logic.

### Type of change


Magicbook1108 (commented):
LGTM

PR #14425, opened 2026-04-28
Fix: agent toolcall null response & schema validation & DeepSeek think history

buua436:
agent toolcall null response & schema validation & DeepSeek think history

### Type of change


Magicbook1108 (approved):
LGTM

PR #14445, opened 2026-04-28
feat(ExecSQL/BigQuery): add BigQuery Support on ExecSQL

alfaadriel:

This PR solve BigQuery integration support, many people use BigQuery for OLAP Database with this PR now agent can help do query to BigQuery 

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


JinHai-CN:
@alfaadriel Thank you for your PR to introduce BigQuery. We really appreciate the time and effort you've put into this.

We haven't merged it yet,  because we need to think about long-term maintenance.

Integrating a new component is a significant commitment. It would require:

- Ongoing maintenance across future RAGFlow releases
- Compatibility testing with each minor version update
- Prompt fixes for any issues that arise

For us to move forward with this integration, we would need someone (or a team) to take long-term responsibility for maintaining the BigQuery integration.

Would you be open to taking on that role? If so, we'd be happy to revisit the PR and work with you to get it merged.

Issue #14362, opened 2026-04-24
[Feature Request]: Datasource functionality update


Magicbook1108:
Hello, community!

We’re introducing a functionality update for data sources in RAGFlow: **syncing deleted files for each data source**. You’re more than welcome to contribute and help us improve this feature.

Example:
#14353
#14000

To be done:

- [x] Airtable #14438 
- [x] Asana #14468
- [x] Bitbucket #14450
- [ ] DingTalk AI TableV #14525
- [x] Discord #14451
- [x] Dropbox #14476
- [x] GitLab #14481 
- [x] Gmail #14462
- [x] Google Drive #14372 
- [ ] IMAP
- [ ] ~~MySQL / PostgreSQL~~
- [x] RSS #14493
- [x] Seafile #14499
- [x] WebDAV #14491
- [x] Zendesk #14487

ParasSondhi:
@Magicbook1108 I will take the Google Drive integration. Working on it now, will open a PR once local testing is complete.

PR #14491, opened 2026-04-29
feat(webdav): support deleted-file sync via slim snapshot

spider-yamet:

Incremental WebDAV sync only ingested files whose modification time fell inside the poll window; documents removed on the WebDAV server were never removed from the knowledge base. This aligns with [#14362](https://github.com/infiniflow/ragflow/issues/14362) (coordinated datasource “sync deleted files” work).

This PR adds a **full-tree slim snapshot** (`retrieve_all_slim_docs_perm_sync`) that enumerates current remote paths **without downloading file contents**, using the same logical document IDs as full ingest (`webdav:{base_url}:{file_path}`). When **`sync_deleted_files`** is enabled on incremental runs, sync returns **`(document_generator, file_list)`** so **`SyncBase`** runs **`cleanup_stale_documents_for_task`** and removes KB rows no longer present remotely.

Design notes:

- **`_list_files_recursive`** gains **`filter_by_mtime`**: snapshot passes **`filter_by_mtime=False`** (full tree under **`remote_path`**); **`poll_source`** keeps mtime-window filtering as before.
- Slim snapshot applies the same **extension** and **`size_threshold`** rules as **`_yield_webdav_documents`** so retain IDs match what would be indexed.
- **`end_ts`** is captured before building **`file_list`**, then **`poll_source`** uses the same upper bound (consistent with Dropbox-style connectors).

## Type of change

## Files changed

| Area | Change |
|------|--------|
| `common/data_source/webdav_connector.py` | `SlimConnectorWithPermSync`, `retrieve_all_slim_docs_perm_sync`, `filter_by_mtime` on `_list_files_recursive` |
| `rag/svr/sync_data_source.py` | WebDAV `_generate`: `file_list` + tuple return; pass **`batch_size`** from connector config |
| `web/src/pages/user-setting/data-source/constant/index.tsx` | `syncDeletedFiles` for WebDAV in `DataSourceFeatureVisibilityMap` |


spider-yamet:
@yingfeng could you please share your opinion? passing the ci :)

spider-yamet:
I would appreciate your review, @Magicbook1108 @buua436 . Thanks

JinHai-CN:
@spider-yamet would you please resolve the conflict?

web/src/pages/user-setting/data-source/constant/index.tsx
```diff
@@ -105,6 +105,9 @@ export const DataSourceFeatureVisibilityMap: Partial<
   [DataSourceKey.AIRTABLE]: {
     syncDeletedFiles: true,
   },
+  [DataSourceKey.WEBDAV]: {
+     syncDeletedFiles: true,
+  }
```
JinHai-CN: missing ','

JinHai-CN:
@spider-yamet syntax error, please check the comment

web/src/pages/user-setting/data-source/constant/index.tsx
spider-yamet: @JinHai-CN Fixed. Thanks for your feedback.

PR #14493, opened 2026-04-29
feat(rss): support deleted-file sync

bitloi:

Partially addresses #14362.

This PR enables syncing deleted files for RSS data sources.

Previously, RSS incremental sync only returned feed entries whose timestamps were inside the poll window. If an entry was removed from the RSS feed, RAGFlow had no full current RSS snapshot to pass into the shared stale-document cleanup path, so the deleted remote entry could remain in the knowledge base.

This PR:
- adds `retrieve_all_slim_docs_perm_sync()` to `RSSConnector`
- reuses the same `rss:<md5(stable_key)>` document ID derivation used by normal RSS ingest
- returns `(document_generator, file_list)` for incremental RSS sync when `sync_deleted_files` is enabled
- captures the poll end timestamp before snapshot/poll so cleanup does not race against the same sync window
- adds start/end logs around RSS slim snapshot collection
- exposes the deleted-file sync toggle for RSS in the data source UI

Per maintainer request on related datasource PRs, this PR contains no test-case changes. Local verification was run with an external script.

Validation:
- `uv run ruff check common/data_source/rss_connector.py rag/svr/sync_data_source.py`
- `uv run pytest test/unit_test/rag/test_sync_data_source.py -q`
- `./node_modules/.bin/eslint src/pages/user-setting/data-source/constant/index.tsx`
- `git diff --check`
- `uv run python /tmp/verify_rss_deleted_sync.py --repo /root/74/ragflow`

### Type of change

### DIFF
--- a/common/data_source/rss_connector.py
+++ b/common/data_source/rss_connector.py
@@ -10,14 +10,20 @@ import feedparser
 import requests
 
 from common.data_source.config import INDEX_BATCH_SIZE, REQUEST_TIMEOUT_SECONDS, DocumentSource
-from common.data_source.interfaces import LoadConnector, PollConnector
-from common.data_source.models import Document, GenerateDocumentsOutput, SecondsSinceUnixEpoch
+from common.data_source.interfaces import LoadConnector, PollConnector, SlimConnectorWithPermSync
+from common.data_source.models import (
+    Document,
+    GenerateDocumentsOutput,
+    GenerateSlimDocumentOutput,
+    SecondsSinceUnixEpoch,
+    SlimDocument,
+)
 from common.ssrf_guard import assert_url_is_safe, pin_dns as _pin_dns
 
 _MAX_REDIRECTS = 10
 
 
-class RSSConnector(LoadConnector, PollConnector):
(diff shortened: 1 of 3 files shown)


bitloi:
RSS delete-sync verification completed with a local verification script.

The script:
1. Created a local mock RSS feed with two entries.
2. Verified the RSS slim snapshot returned the same document IDs as normal ingest.
3. Removed one feed entry.
4. Verified the slim snapshot no longer contained the removed entry ID.

Output:

```text
[PASS] RSS slim snapshot matches ingest IDs: ['rss:d6b2730909c46f0a32d8d6385421db13', 'rss:7aff1f8a642253b1a896a6d0315cef47']
[PASS] RSS slim snapshot removed deleted entry ID: rss:7aff1f8a642253b1a896a6d0315cef47
```

@yingfeng @Magicbook1108 Please check the output.

bitloi:
@Magicbook1108 I fixed the conflicts, can you review again?

Magicbook1108:
lgtm

bitloi:
@JinHai-CN Would you please review this PR?

PR #14496, opened 2026-04-29
fix: authorize beta document downloads by tenant

sebastiondev:
## Summary

This fixes a missing authorization check in the beta API document download endpoint:

- **CWE:** CWE-862 (Missing Authorization)
- **Severity:** Medium
- **Affected route/file:** `GET /api/v1/documents/<document_id>` in `api/apps/sdk/doc.py`
- **Data flow:** the route reads a bearer beta API token, resolves the token with `APIToken.query(beta=token)`, accepts `document_id` directly from the URL, loads the document with `DocumentService.query(id=document_id)`, and then fetches the backing object through `File2DocumentService.get_storage_address()` / `settings.STORAGE_IMPL.get()`.

Before this change, that flow verified that the API token was valid, but it did not verify that the token's tenant owned the document's knowledge base. A caller with any valid beta API token and a known document ID could therefore reach storage for a document belonging to another tenant.

## Fix

The endpoint now takes the tenant ID from the resolved API token and checks the document's knowledge base with:

```python
KnowledgebaseService.query(id=doc[0].kb_id, tenant_id=tenant_id)
```

If the knowledge base is not owned by the token tenant, the request returns an access error before any storage lookup occurs. This mirrors the tenant-scoped ownership checks used by the dataset-scoped document download path and keeps the patch small.

## Tests

Added unit coverage for `download_doc()` to assert that:

- the beta token tenant ID is used in the knowledge-base ownership lookup;
- cross-tenant access returns `You do not have access to this document.`;
- storage resolution is not called before tenant authorization succeeds;
- the existing same-tenant empty-file and successful-download paths still run after the authorization gate passes.

I also verified the final patch is limited to `api/apps/sdk/doc.py` and the related document SDK route unit test. A local `pytest` invocation could not complete in this checkout because the shared test fixture attempts to log in to a RAGFlow server at `127.0.0.1:9380`, which was not running in the local environment.

## Security analysis

This is exploitable when an attacker has a valid beta API token for their own tenant and obtains or guesses a document ID from another tenant. The token alone should not grant access to other tenants' files, but the direct document route previously authorized only the token itself and not the requested resource. The new tenant-scoped knowledge-base check binds the requested document back to the token tenant before storage is accessed, preventing cross-tenant document downloads through this endpoint.

Before submitting, we attempted to disprove this by checking whether existing dataset-scoped routes, token validation, or framework protections already enforced ownership. They do not apply to this direct document-ID route: it bypassed the dataset path parameter and used only `DocumentService.query(id=document_id)` before reading storage.

cc @lewiswigmore

### DIFF
--- a/api/apps/sdk/doc.py
+++ b/api/apps/sdk/doc.py
@@ -13,6 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
+import logging
 from io import BytesIO
 
 from quart import request, send_file
@@ -103,10 +104,16 @@ async def download_doc(document_id):
     if len(token) != 2:
         return get_error_data_result(message="Authorization is not valid!")
     token = token[1]
+    logging.info("Beta API token lookup attempted for document download")
     objs = APIToken.query(beta=token)
     if not objs:
+        logging.warning("Beta API token lookup failed for document download: invalid API key")
         return get_error_data_result(message='Authentication error: API key is invalid!"')
+    if len(objs) > 1:
+        logging.error("Beta API token lookup is ambiguous for document download: matches=%s", len(objs))
+        return get_error_data_result(message="Authentication error: API key configuration is ambiguous.")
     tenant_id = objs[0].tenant_id
+    logging.info("Beta API token authorized for document download: tenant_id=%s", tenant_id)
 
     if not document_id:
         return get_error_data_result(message="Specify document_id please.")
@@ -114,6 +121,12 @@ async def download_doc(document_id):
     if not doc:
         return get_error_data_result(message=f"The dataset not own the document {document_id}.")
     if not KnowledgebaseService.query(id=doc[0].kb_id, tenant_id=tenant_id):
+        logging.warning(
+            "cross-tenant access denied for document download: tenant_id=%s kb_id=%s document_id=%s",
+            tenant_id,
+            doc[0].kb_id,
+            document_id,
+        )
         return get_error_data_result(message="You do not have access to this document.")
     # The process of downloading
     doc_id, doc_location = File2DocumentService.get_storage_address(doc_id=document_id)  # minio address
--- a/api/apps/sdk/doc.py
+++ b/api/apps/sdk/doc.py
@@ -106,12 +106,15 @@ async def download_doc(document_id):
     objs = APIToken.query(beta=token)
     if not objs:
         return get_error_data_result(message='Authentication error: API key is invalid!"')
+    tenant_id = objs[0].tenant_id
 
     if not document_id:
         return get_error_data_result(message="Specify document_id please.")
     doc = DocumentService.query(id=document_id)
     if not doc:
         return get_error_data_result(message=f"The dataset not own the document {document_id}.")
+    if not KnowledgebaseService.query(id=doc[0].kb_id, tenant_id=tenant_id):
+        return get_error_data_result(message="You do not have access to this document.")
     # The process of downloading
     doc_id, doc_location = File2DocumentService.get_storage_address(doc_id=document_id)  # minio address
     file_stream = settings.STORAGE_IMPL.get(doc_id, doc_location)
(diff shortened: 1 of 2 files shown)


sebastiondev:
Thanks for the review. I addressed the observability and beta-token ambiguity feedback in commit 9dfae484a.\n\nChanges made:\n- Added logging around beta API token lookup attempts, invalid-token failures, ambiguous-token failures, and successful authorization.\n- Explicitly reject ambiguous beta token matches instead of using objs[0].\n- Added a warning log immediately before the cross-tenant document download denial with tenant_id, kb_id, document_id, and the denial reason.\n- Added unit coverage for the ambiguous-token rejection path.\n\nTest note: I attempted to run the targeted pytest file, but the local test bootstrap exits because ZHIPU_AI_API_KEY is not set in this environment.

sebastiondev:
Thanks again. Just to confirm against the latest review pass:

- Observability for the cross-tenant denial branch (lines 123–129 of `api/apps/sdk/doc.py`) is in place — `logging.warning` records `tenant_id`, `kb_id`, `document_id`, and the reason before the `get_error_data_result` return.
- Beta-token lookup ambiguity is now explicitly rejected (lines 112–114) with an error log and a dedicated user-facing message, instead of silently picking `objs[0]`.
- Lookup attempt, invalid-token failure, and successful authorization paths all emit logs.
- Unit test for the ambiguous-token rejection added in `test_doc_sdk_routes_unit.py`.

The latest CodeRabbit run also reports "No actionable comments", and the inline comment that surfaced afterwards appears to be an analysis-script fragment with no concrete change request. Happy to address anything further if I missed something.

xugangqiang:
@sebastiondev 
May I ask if this API is used other places? If so, we want to change them together.

sebastiondev:
@xugangqiang Thanks for the question. Yes — `APIToken.query(beta=token)` is used in several other endpoints. A quick grep:

```
api/apps/sdk/doc.py:108       (this PR)
api/apps/sdk/session.py:148   chatbot_completions
api/apps/sdk/session.py:174   chatbots_inputs
api/apps/sdk/session.py:200   agent_bot_completions
api/apps/sdk/session.py:243   begin_inputs
api/apps/sdk/session.py:264   ask_about_embedded
api/apps/sdk/session.py:303   retrieval_test_embedded
api/apps/sdk/session.py:435   related_questions_embedded
api/apps/sdk/session.py:483   detail_share_embedded
api/apps/sdk/session.py:515   mindmap
```

Two related concerns apply across these call sites:

1. **Beta-token ambiguity**: every site does `objs = APIToken.query(beta=token)` and then uses `objs[0].tenant_id` if `objs` is non-empty, with no check for `len(objs) > 1`. The right long-term fix is a UNIQUE constraint on `APIToken.beta` (schema + migration); until then, applying the same explicit `len(objs) > 1` rejection used in this PR would harden the other endpoints.

2. **Tenant-scoped resource lookup (the IDOR pattern this PR fixes)**: of the session.py endpoints, the ones that look up a resource by ID supplied in the URL/body and act on it without joining on `tenant_id` are the most analogous to the `download_doc` bug. From a quick read:
   - `chatbot_completions` / `chatbots_inputs` resolve `dialog_id` via `DialogService` without a tenant filter.
   - `agent_bot_completions` / `begin_inputs` resolve `agent_id` via `UserCanvasService` without a tenant filter, then pass `objs[0].tenant_id` into `agent_completion` / `Canvas` — this means a valid beta token from tenant A could potentially drive an agent owned by tenant B.
   - `ask_about_embedded` / `retrieval_test_embedded` / `mindmap` accept `kb_ids` from the request body and don't verify those KBs belong to the token's tenant before retrieval.
   - `detail_share_embedded` already iterates `UserTenantService.query(user_id=tenant_id)` and checks ownership, so it looks OK.

I deliberately kept this PR narrowly scoped to the originally reported `GET /v1/document/<document_id>` IDOR (the issue I tested and have a reproducer for), to keep the diff small and reviewable. If you'd like, I'm happy to do either of the following as a follow-up:

- Open a separate PR that adds tenant-scoped resource queries + the ambiguous-token rejection across the `session.py` endpoints listed above, or
- Extend this PR to cover them all in one go.

Just let me know which you'd prefer and I'll proceed.

lewiswigmore:

lewiswigmore:
Thanks for reviewing.

_P.S. [Foundation Machines](https://foundationmachines.ai) runs Sebastion as a service if continuous security audits would help._

Issue #5863, opened 2025-03-10
[Bug]: ONNXRuntimeError memory allocation error


trcyberoptic:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### RAGFlow workspace code commit ID

?

### RAGFlow image version

v0.17.0-90-g47e244ee full

### Other environment information

```Markdown

```

### Actual behavior

```
Begin at:
Mon, 10 Mar 2025 15:16:39 GMT
Duration:
35.46 s
Progress:
15:16:39 Task has been received.
15:16:40 Page(1~13): OCR started
15:16:43 Page(1~13): OCR finished (3.48s)
15:16:44 Page(1~13): Layout analysis (0.82s)
15:16:44 Page(1~13): Table analysis (0.00s)
15:16:44 Page(1~13): Text merged (0.00s)
15:16:55 Page(1~13): Generate 27 chunks
15:16:56 Page(1~13): Embedding chunks (0.89s)
15:16:59 Page(1~13): Indexing done (2.63s). Task done (19.47s)
15:16:39 Task has been received.
15:16:45 Page(13~25): OCR started
15:16:49 Page(13~25): OCR finished (3.84s)
15:16:49 Page(13~25): Layout analysis (0.79s)
15:16:50 Page(13~25): Table analysis (0.03s)
15:16:50 Page(13~25): Text merged (0.00s)
15:17:01 Page(13~25): Generate 28 chunks
15:17:02 Page(13~25): Embedding chunks (1.00s)
15:17:04 Page(13~25): Indexing done (2.22s). Task done (24.67s)
15:16:39 Task has been received.
15:16:50 Page(25~35): OCR started
15:17:08 Page(25~35): [ERROR]Internal server error while chunking: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running Conv node. Name:p2o.Conv.28 Status Message: /onnxruntime_src/onnxruntime/core/framework/bfc_arena.cc:376 void* onnxruntime::BFCArena::AllocateRawInternal(size_t, bool, onnxruntime::Stream*, bool, onnxruntime::WaitNotificationFn) Available memory of 0 is smaller than requested bytes of 144322816
[...]

PR #14501, opened 2026-04-29
feat(task_executor): add MAX_TASKS_PER_WORKER to recycle long-running workers

HamsterPark:
### Summary

Adds an opt-in `MAX_TASKS_PER_WORKER` environment variable that recycles the `task_executor` worker process after a configurable number of completed tasks, releasing memory that long-lived libraries cannot reclaim on their own.

> **Rebased onto current `main`** and squashed to a single commit. The branch had fallen ~3k commits behind; the only conflict was the constant block, whose anchor moved into `rag/svr/task_executor_limiter.py` during the task executor refactors (#15154 / #15471).

## Motivation

`task_executor.py` lazily loads multiple ONNX Runtime sessions (OCR, layout recognition, table structure recognition, etc.) over the lifetime of the process. ONNX Runtime's `BFCArena` allocator keeps every chunk it grabs until the `InferenceSession` is destroyed, so on lower-VRAM consumer GPUs the cumulative arena fragmentation eventually starves new allocations:

```
[ONNXRuntimeError] : 1 : FAIL : BFCArena::AllocateRawInternal:
Available memory of 0 is smaller than requested bytes of 8478720
```

Once the worker is in this state, every subsequent task fails until the process is restarted. There are existing reports of this exact symptom (e.g. #5863).

The supervisor loop in `docker/entrypoint.sh` already restarts the worker after a clean exit:

```bash
while true; do
    LD_PRELOAD="$JEMALLOC_PATH" \
    "$PY" rag/svr/task_executor.py -i "${host_id}_${consumer_id}" -t "common" &
    wait;
    sleep 1;
done
```

…so all that is missing is a way for the worker to ask to be recycled — the same pattern Celery (`max-tasks-per-child`) and Gunicorn (`max-requests`) already use.

## What this PR does

- Adds `MAX_TASKS_PER_WORKER` (default `0` = disabled).
- `handle_task()` now reports whether it did real work, so idle polls against an empty queue do not advance the counter.
- `task_manager()` counts completions and sets `stop_event` once the threshold is reached, so `main()` exits cleanly and the supervisor restarts the process.
- The main loop re-checks `stop_event` immediately after `task_limiter.acquire()`, so a worker that was blocked on the semaphore doesn't pick up one extra task after the shutdown signal fires.
- **When recycling is enabled**, in-flight `task_manager`s are drained rather than cancelled, bounded by `RECYCLE_SHUTDOWN_TIMEOUT` (default `300`s) so a single hung task cannot block the exit.
- Documents both variables in `docker/.env`.

With `MAX_TASKS_PER_WORKER=0` nothing changes: the counter is never incremented, `stop_event` is never set from this path, and the shutdown branch falls through to the existing `cancel()` + `gather()` unchanged. The drain is deliberately gated on the feature being on, so this PR does not alter the SIGTERM/SIGINT shutdown semantics of a normally configured deployment.

### On the threshold being soft

`MAX_CONCURRENT_TASKS` tasks run concurrently, so when the Nth task completes there can be up to `MAX_CONCURRENT_TASKS - 1` others already in flight; those are allowed to finish. A worker therefore completes between `N` and `N + MAX_CONCURRENT_TASKS - 1` tasks before exiting. This mirrors how Celery's `max-tasks-per-child` behaves with prefetch, and it is called out in both the code comment and `docker/.env` so the number isn't surprising in practice.

## Validation

Originally reproduced on a single-GPU RTX 4060 (8 GB VRAM) ingesting ~200 PDF papers via the `paper` parser with full DeepDOC. With the default `MAX_TASKS_PER_WORKER=0` the worker started failing every subsequent task with `Available memory of 0` after ~30 documents. With `MAX_TASKS_PER_WORKER=20` the worker exits cleanly (visible as `[recycle] reached MAX_TASKS_PER_WORKER=20` in the logs), the supervisor restarts it, VRAM usage drops back to baseline, and the queue continues without arena failures.

After rebasing onto the refactored `main`, the recycle logic was re-verified with a standalone asyncio harness mirroring `main()` / `task_manager()` / `handle_task()` with stubs:

| scenario | result |
| --- | --- |
| `MAX=3`, concurrency 5 | exits after 3–7 completions, nothing cancelled |
| `MAX=3`, concurrency 1 | exits after exactly 3 |
| `MAX=3` with idle polls interleaved | only real work advances the counter |
| `MAX=2`, one task hangs | bounded grace, then cancelled — a hang cannot block the exit |
| `MAX=0` + external stop | legacy path, no drain |

`ruff check` and `ruff format --check` are clean on both changed files, with no new violations relative to `main`.

## Test plan

- [x] `MAX_TASKS_PER_WORKER` unset or `0` — no behaviour change (verified pre-rebase; post-rebase the shutdown path is gated so it is unchanged from `main` by construction).
- [x] `MAX_TASKS_PER_WORKER=N` — the worker logs `[recycle] reached MAX_TASKS_PER_WORKER=N` and is restarted by the supervisor (verified pre-rebase on hardware, and post-rebase with the harness above).
- [x] Long ingestion run on a low-VRAM GPU no longer accumulates `Available memory of 0` failures (verified pre-rebase on the RTX 4060).
- [ ] Repository CI — as a fork PR this needs the `ci` label to run.

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -287,5 +287,12 @@ DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
 # Used for ThreadPoolExecutor
 THREAD_POOL_MAX_WORKERS=128
 
+# Recycle the task_executor worker process after this many completed tasks
+# to release ONNX Runtime / CUDA arena fragmentation that accumulates over
+# the lifetime of the process. The supervisor loop in entrypoint.sh restarts
+# the worker automatically after a clean exit. Set to a small value (e.g. 20)
+# on low-VRAM GPUs where long-running workers eventually OOM. 0 disables.
+# MAX_TASKS_PER_WORKER=0
+
 #Option to disable login form for SSO
 DISABLE_PASSWORD_LOGIN=false
--- a/docker/.env
+++ b/docker/.env
@@ -438,6 +438,19 @@ LLM_POOL_RATE_LIMIT_RETRIES=3
 LLM_POOL_RATE_LIMIT_RETRY_BASE_DELAY=1.0
 LLM_POOL_RATE_LIMIT_RETRY_MAX_DELAY=30.0
 
+# Recycle the task_executor worker process after this many completed tasks, to
+# release the ONNX Runtime / CUDA arena fragmentation that builds up over the
+# lifetime of the process. The supervisor loop in entrypoint.sh restarts the
+# worker automatically after a clean exit. Set a small value (e.g. 20) on
+# low-VRAM GPUs where long-running workers eventually run out of memory.
+# The threshold is soft: tasks already in flight are allowed to finish, so up to
+# MAX_CONCURRENT_TASKS - 1 extra tasks may complete before the worker exits.
+# 0 disables recycling.
+# MAX_TASKS_PER_WORKER=0
+# How long (seconds) a recycling worker waits for in-flight tasks to finish
+# before cancelling them. Only has an effect when MAX_TASKS_PER_WORKER > 0.
+# RECYCLE_SHUTDOWN_TIMEOUT=300
+
 #Option to disable login form for SSO
 DISABLE_PASSWORD_LOGIN=false
 
--- a/rag/svr/task_executor.py
+++ b/rag/svr/task_executor.py
@@ -124,6 +124,17 @@ CURRENT_TASKS = {}
 MAX_CONCURRENT_TASKS = int(os.environ.get('MAX_CONCURRENT_TASKS', "5"))
 MAX_CONCURRENT_CHUNK_BUILDERS = int(os.environ.get('MAX_CONCURRENT_CHUNK_BUILDERS', "1"))
 MAX_CONCURRENT_MINIO = int(os.environ.get('MAX_CONCURRENT_MINIO', '10'))
+# Recycle the worker process after this many completed tasks to release
+# unreclaimed memory held by long-lived libraries (notably ONNX Runtime's
+# BFCArena, which keeps allocated chunks until the InferenceSession is
+# destroyed). The supervisor loop in `docker/entrypoint.sh`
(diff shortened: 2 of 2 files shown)


JinHai-CN:
@HamsterPark would you please resolve the conficts?

HamsterPark:
@JinHai-CN Rebased onto current `main` and squashed to a single commit — conflicts resolved, the PR is mergeable again. The only conflict was the constant block, whose anchor moved into `rag/svr/task_executor_limiter.py` during the task executor refactors. The drain on shutdown is now gated on `MAX_TASKS_PER_WORKER > 0`, so with the feature off the shutdown path is unchanged from `main`.

The workflow jobs are skipping because a fork PR needs the `ci` label — could you add it when you get a chance? Happy to adjust anything else.

xugangqiang:
LGTM
@yuzhichang please have a second eye check.

yuzhichang:
@HamsterPark Please fix unit test failure.

HamsterPark:
@yuzhichang Thanks for the review. That failure was the NLTK `wordnet` lookup rather than this change.

All 40 failures in that run were the same `LookupError: Resource 'wordnet' not found` (4075 passed / 40 failed), spread across the epub / mineru / gaussdb / graphrag / audio tests, and every test that actually imports `rag/svr/task_executor.py` passed.

The branch was based on a commit 11 behind #19251 (`5d0e2253c`, *"fix ci unit test not found in installed package 'wordnet'"*), so the merge ref didn't carry the `test/unit_test/conftest.py` fix.

I've rebased onto current `main` — same code, and `ragflow_preflight` is green now. The rebase also resolved a new `docker/.env` conflict against the Knowledge compilation block; `rag/svr/task_executor.py` was untouched upstream, so the change carried over unchanged.

xugangqiang:
you are right - it‘s upstream resource change and it has been fixed in main. please rebase and then ci will include the fix.

PR #14541, opened 2026-04-30
feat(auth): add LDAP authentication channel

### LINKED ISSUE #1876: Is it possible to add keycloak or LDAP auth?
### Describe your problem

I was looking in the docs unfortuneatly didnot find anything on oauth2 and openid.
Is it possible to add keycloak-openid or directly  LDAP auth?

thanks for the awesome project!

mhtkarakose:
## Summary

Adds LDAP as a first-class login channel alongside the existing OAuth/OIDC/GitHub channels. Closes #1876.

- New `LDAPClient` under `api/apps/auth/` supporting both **direct-bind** (`bind_dn_template`) and **search-then-bind** (service-account) flows, with RFC 4515 filter escaping to prevent LDAP injection.
- New `POST /auth/login/ldap` endpoint that decrypts the RSA-encoded password, validates credentials against the directory, and reuses the existing `user_register()` path so an LDAP user lands in the same tenant/role bootstrap as OAuth users. The blocking LDAP call is wrapped with `asyncio.to_thread` so it doesn't stall the event loop.
- `/auth/login/channels` now exposes `type` and `form_login`, so the frontend can distinguish redirect-style channels (OAuth) from form-style channels (LDAP). `/auth/login/<channel>` redirects LDAP back to the login page instead of trying to call `get_authorization_url`.
- Frontend: `useLoginWithLdap` hook and an inline `LdapDialog` on the login-next page that pops open when an LDAP channel button is clicked. Submits an RSA-encrypted password through the same crypto path the local login uses.
- Adds `ldap3>=2.9.1` and documents the new `oauth.ldap` block in `docker/service_conf.yaml.template`.

### Example config

```yaml
oauth:
  ldap:
    type: "ldap"
    display_name: "LDAP"
    host: "ldap.example.com"
    port: 389
    use_tls: false
    # Direct-bind (simple):
    bind_dn_template: "uid={username},ou=people,dc=example,dc=com"
    # Or search-then-bind (uncomment to enable):
    # bind_user_dn: "cn=admin,dc=example,dc=com"
    # bind_user_password: "service_account_password"
    # user_search_base: "ou=people,dc=example,dc=com"
    # user_search_filter: "(uid={username})"
```

## Test plan

- [x] Unit tests for `LDAPClient` covering direct-bind success, search-then-bind, invalid credentials, missing user, empty credentials, and filter escaping (`test/unit_test/api/auth/test_ldap_client.py`)
- [ ] Manual: stand up a local OpenLDAP container, configure `oauth.ldap`, click the new "Sign in with LDAP" button, log in, and verify the user is created with `login_channel="ldap"`
- [ ] Manual: confirm OAuth/OIDC/GitHub channels still log in unchanged
- [ ] Manual: hit `POST /auth/login/ldap` with bad credentials and verify a 401 with a generic message (no info leak)


mhtkarakose:
@coderabbitai review

The previous CodeRabbit review attempts hit transient "Review failed" errors. The branch has since been updated through commit 96de3478b which:

- closes the docstring-coverage gate (LDAPClient + helpers + auth factory + _mask_email all documented)
- narrows the password-decode `except` to specific exceptions and adds a typed log line
- enriches the LDAP success log with `user_dn` and a `synthetic_email` flag
- adds non-sensitive flow logging inside `LDAPClient._resolve_user` for the bind/search branches

Could you re-run the review against the latest commit?

mhtkarakose:
@Copilot please re-review the latest commits (96de3478b) — the original review notes have been addressed: real password is now base64-decoded before LDAP bind, RDN escaping uses RFC 4514, TLS validation defaults to CERT_REQUIRED with explicit StartTLS for `use_tls`, frontend handles `rsaPsw()` returning false, channel-type inference is centralised, and unit tests cover StartTLS + RDN escaping + channel-namespaced fallback email.

mhtkarakose:
@coderabbitai review

The previous review covered `96de3478b..4a27c2351` only. Since then I pushed two more fixes whose subjects directly match your last two findings:

- `4df63b784` — wraps every `_open`+`_bind` block in `try/finally` so failed binds still call `unbind()` (addresses the connection-leak comment in `ldap.py`)
- `0c72955c2` — switches `useLoginWithLdap` to `[Authorization]/[Token]/[UserInfo]` computed keys so the storage keys match the constants (addresses the token-key drift comment in `use-login-request.ts`)

Please re-scan the latest tip (`0c72955c2`).

mhtkarakose:
@coderabbitai review

Latest tip is now `e845fd0f2`. Since the previous scan I have:

- `4df63b784` — wraps every `_open`+`_bind` block in `try/finally` so failed binds always release the socket; refuses to instantiate when `bind_user_password` is empty (anonymous-bind hole), when both bind strategies are set, or when the channel id contains characters that would corrupt the synthetic email fallback.
- `0c72955c2` — `useLoginWithLdap` now writes via `[Authorization]/[Token]/[UserInfo]` computed keys imported from `@/constants/authorization` so storage keys match what `getToken`/`removeAll` look up.
- `e845fd0f2` — `_bind` now calls `Connection.open()` before `start_tls()` so the socket is guaranteed open at TLS-upgrade time across ldap3 versions; test asserts the open → start_tls → bind order.

Please re-scan.

KevinHuSh:
Please submit the corresponding `uv.lock` while `pyproject.yml` is changed.

mhtkarakose:
Done — `uv.lock` is now updated alongside `pyproject.toml` (added `ldap3 v2.9.1` via `uv lock`). Also rebased onto current `main` to clear the previous merge conflict. Thanks for the review!

KevinHuSh:
Appreciations!
But CI failures.

```
During handling of the above exception, another exception occurred:
test/unit_test/api/auth/test_ldap_client.py:21: in <module>
    from api.apps.auth.ldap import LDAPAuthError, LDAPClient
api/apps/__init__.py:35: in <module>
    from api.utils.api_utils import server_error_response, get_json_result
api/utils/api_utils.py:48: in <module>
    from api.db.services.tenant_llm_service import LLMFactoriesService
api/db/services/tenant_llm_service.py:27: in <module>
    from rag.llm import ChatModel, CvModel, EmbeddingModel, OcrModel, RerankModel, Seq2txtModel, TTSModel
rag/llm/__init__.py:164: in <module>
    module = importlib.import_module(full_module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/home/alice/.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
rag/llm/ocr_model.py:21: in <module>
    from deepdoc.parser.mineru_parser import MinerUParser
deepdoc/parser/__init__.py:24: in <module>
    from .pdf_parser import PlainParser
deepdoc/parser/pdf_parser.py:33: in <module>
    import xgboost as xgb
.venv/lib/python3.12/site-packages/xgboost/__init__.py:9: in <module>
    from .core import DMatrix, DeviceQuantileDMatrix, Booster, DataIter, build_info
.venv/lib/python3.12/site-packages/xgboost/core.py:20: in <module>
    from .compat import STRING_TYPES, DataFrame, py_str, PANDAS_INSTALLED
.venv/lib/python3.12/site-packages/xgboost/compat.py:108: in <module>
    except pkg_resources.DistributionNotFound:
           ^^^^^^^^^^^^^
E   NameError: name 'pkg_resources' is not defined```

**Comment by mhtkarakose** [comment]
Thanks for catching that! The CI failure was caused by importing `api.apps.auth.ldap` eagerly running `api/apps/__init__.py`, which transitively pulls in `deepdoc.parser.pdf_parser` -> `xgboost`. xgboost's `compat.py` then trips on a missing `pkg_resources` reference in the test environment.

Added `test/unit_test/api/auth/conftest.py` that installs a bare `xgboost` stub in `sys.modules` before the test module is collected. The LDAP tests never touch xgboost (it's only used inside `pdf_parser` method bodies, not at module load), so the stub is safe. Pushed as part of the same commit.

**Comment by mhtkarakose** [comment]
Still failing — the xgboost stub got past that line but the chain kept going (cv2 needs `libGL.so.1` at the OS level, etc.). Chasing one stub at a time is a losing game.

Rewrote the conftest to **bypass the `api/apps/__init__.py` chain entirely** for these tests. It pre-loads `api/apps/auth/ldap.py` from disk via `importlib.util.spec_from_file_location`, with minimal stubs for the `api` / `api.apps` / `api.apps.auth` package shells and a tiny `UserInfo` substitute for `from .oauth import UserInfo`. The test module's `from api.apps.auth.ldap import ...` then finds the already-loaded module in `sys.modules` and skips package init.

Verified locally with only `ldap3` + `pytest` installed — all 17 tests pass with no Flask / Quart / xgboost / cv2 / deepdoc dependencies required.

Pushed as `b0e446e0c`.

**Comment by mhtkarakose** [comment]
Heads-up — looks like the current CI failures aren't from this PR:

- The scheduled `tests` workflow on master ([run #25748712697](https://github.com/infiniflow/ragflow/actions/runs/25748712697)) failed at 16:44 UTC today on the **same step** that's hanging on this PR: `Run sdk tests against Infinity`.
- That step's first line is an infinite `until` loop waiting for the ragflow container to answer `/v1/system/ping`, so when the container doesn't come up the step just hangs until the runner timeout.
- I rebased onto current `main` (picking up #14833 "Speed up start time" and the other 26 commits) and re-pushed; the new run still hangs at the same point.

Since the same break is reproducible on master without any of my changes, it looks like a runner / Infinity-startup issue rather than a PR regression. Happy to retry once master CI is green again. The actual LDAP unit-test step (`Run unit test`) passes cleanly now — the new conftest bypass works.

**Comment by KevinHuSh** [comment]
Resolve the conflicts please.

**Comment by mhtkarakose** [comment]
Rebased onto current `main` — the only conflict was `uv.lock` (upstream picked up new dependencies in the 124 commits since my last rebase). Took the upstream lockfile as a base and re-ran `uv lock` so the net delta is just the `ldap3` + `pyasn1` entries (+14 lines). PR is mergeable again.

**Comment by mhtkarakose** [comment]
CI on this PR timed out at `Run sdk tests against Infinity` again (~6h, hung waiting for the ragflow container's `/v1/system/ping`). Same shape as before — and it looks like a master-side issue, not anything specific to this branch:

- The scheduled `tests` workflow on `main` has failed at the same step every day for the past 5 runs (May 16, 17, 18, 19, 20 — e.g. [run #26177656967](https://github.com/infiniflow/ragflow/actions/runs/26177656967)), with no LDAP code involved.
- The PR itself is mergeable, clean of conflicts, and the `Run unit test` step (which is what this PR actually changes) passes.

Anything I can do from this side, or should we wait for master CI to recover?

**Comment by JinHai-CN** [comment]
@mhtkarakose Would you please fix the CI error?
```
./entrypoint.sh: eval: line 174: syntax error near unexpected token `('
```

mhtkarakose:
@JinHai-CN Fixed in `73c3d485c`.

**Root cause.** `docker/entrypoint.sh` (line ~174) renders `service_conf.yaml.template` by running `eval "echo \"$line\""` over *every* line, comments included. The LDAP example block this PR added contained two lines whose embedded double quotes broke that eval once wrapped in `echo "..."`:

- `#     # user_search_filter: "(uid={username})"` — the inner `"` closes the echo string, exposing `(` as an unquoted shell token → `syntax error near unexpected token '('`
- `#     # of the form "<uid>@<channel>.ldap.local" ...` — same thing exposes `<`/`>` as redirections

Because `entrypoint.sh` runs under `set -e`, that eval failure aborts the whole script, so `ragflow-server` never finishes starting. That is also why CI kept hanging for ~6h at `Run sdk tests against Infinity`: the step loops on `/v1/system/ping`, and the container was dead before it could answer. So the earlier "hang" and this `eval` syntax error are the same bug — not a master-side flake as I'd guessed before. Apologies for the misdiagnosis.

**Fix.** Switched those two example values to single quotes (valid YAML, and eval-safe since the special chars stay inside the outer `echo "..."`). I replayed the entire template through entrypoint's exact loop logic afterwards — 0 lines now fail eval. The 17 LDAP unit tests still pass.

The branch is already on current `main` via @KevinHuSh's merge; this commit sits on top of it.

mhtkarakose:
The latest `ragflow_tests` failure ([run #26283621975](https://github.com/infiniflow/ragflow/actions/runs/26283621975)) is a CI timeout, not a code issue:

- The failing step is `Run sdk tests against Elasticsearch`, terminated with exit code 143 (SIGTERM). The run took 39m51s before the step was killed.
- The Elasticsearch SDK tests were passing right up to the termination — nothing actually failed; the step just ran out of time.
- The Infinity half of the same run passed completely (sdk tests, RESTFUL api tests, CLI test all green).
- The branch content is byte-identical to the commit that passed cleanly in [run #26278513551](https://github.com/infiniflow/ragflow/actions/runs/26278513551) (~20m); the only change since was a commit-message rewrite, with no files touched.

Could a maintainer re-run the failed job? I don't have permission to re-run it from the fork. Thanks!

PR #14556, opened 2026-05-02
fix: handle missing parent chunk in retrieval_by_children

vincentlambert:

`retrieval_by_children()` in `rag/nlp/search.py` crashes with a `TypeError: 'NoneType' object is not subscriptable` when a parent ("mom") chunk referenced by child chunks is missing from the index.

This happens when the index is in an inconsistent state — for example after a partial re-index, a document deletion that didn't clean up all children, or a race condition during ingestion. `dataStore.get()` returns `None` for the missing parent, and the subsequent access to `chunk["content_with_weight"]` raises a `TypeError`.

**Stack trace:**
```
TypeError: 'NoneType' object is not subscriptable
  File "rag/nlp/search.py", line 792, in retrieval_by_children
    "content_with_weight": chunk["content_with_weight"],
```

### Type of change

- [x] Bug Fix

### Fix

When `dataStore.get()` returns `None` for a parent chunk, fall back to using the child chunks directly and continue processing the remaining parents. This preserves retrieval results for all other chunks rather than aborting the entire query with an exception.

```python
chunk = self.dataStore.get(id, idx_nms[0], [ck["kb_id"] for ck in cks])
if chunk is None:
    chunks.extend(cks)
    continue
```


rag/nlp/search.py
```diff
@@ -786,6 +786,9 @@ def retrieval_by_children(self, chunks: list[dict], tenant_ids: list[str]):
         vector_size = 1024
         for id, cks in mom_chunks.items():
             chunk = self.dataStore.get(id, idx_nms[0], [ck["kb_id"] for ck in cks])
+            if chunk is None:
```
KevinHuSh, line 784: Please mind this. When chunk is None, how this 'extend' happened?

vincentlambert:
Hi @KevinHuSh — good catch, let me clarify.

**Why `chunks.extend(cks)`?**

Earlier in `retrieval_by_children`, each child chunk is *popped* out of `chunks` and moved into `mom_chunks`:

```python
mom_chunks[ck["mom_id"]].append(chunks.pop(i))
```

So by the time we reach the `for id, cks in mom_chunks.items()` loop, those children are no longer in `chunks` — they've been consumed. If `dataStore.get()` returns `None` for their parent and we simply `continue`, those child chunks are **permanently lost** from the result set.

`chunks.extend(cks)` re-adds the children as a fallback so we don't silently discard results that were ranked as relevant. The child chunks already carry their own `content_with_weight`, `doc_id`, `similarity`, etc. from the original retrieval — they are usable results, just without the parent's broader context window.

The alternative (skip entirely) would mean that any chunk whose parent is missing simply disappears from retrieval output, which seems worse than returning a slightly less contextualized but still relevant result.

---

I've also added a `logging.warning` (as requested by CodeRabbit) so this fallback is visible in logs:

```
WARNING: Parent chunk '<id>' not found in the index; falling back to N child chunk(s).
```

This lets operators identify orphaned-chunk states without having to instrument anything themselves.

PR #14563, opened 2026-05-03
Fix: retry RocksDB metadata contention on concurrent CREATE/DROP

Zzappy24:
Concurrent CREATE TABLE / CREATE INDEX / DROP TABLE on the same Infinity instance can race on the catalog counter (e.g. db|1|next_table_id) and fail with error 9003 "Resource busy" instead of waiting on a lock. Two users creating a knowledge base at the same instant, or any deployment with multiple backend workers behind one Infinity, can hit it.

Wrap the metadata paths in create_idx, create_doc_meta_idx, and delete_idx with exponential backoff + jitter (5 attempts, 50ms base). The wrapped operations already use ConflictType.Ignore, so retrying is idempotent — worst case the second attempt is a no-op against an already-created table. Tunable via INFINITY_META_RETRY_MAX / INFINITY_META_RETRY_BASE_DELAY_MS.

Repro: stress 30 concurrent POST /api/v1/datasets against a 4-worker backend → ~50% of requests fail without the patch (Resource busy from the second worker that hits the counter), 100% succeed with it. At 100 concurrent requests, all 100 succeed in ~1.2s; the retry budget never exhausted in our tests.

Scope is limited to metadata paths only — data-path operations (INSERT chunks, SELECT for retrieval) go through per-table code paths and don't share the contended counter.

### Type of change

### DIFF
--- a/common/doc_store/infinity_conn_base.py
+++ b/common/doc_store/infinity_conn_base.py
@@ -54,12 +54,51 @@ from common.doc_store.doc_store_base import DocStoreConnection, MatchExpr, Order
 
 _T = TypeVar("_T")
 
-_META_RETRY_MAX = int(os.getenv("INFINITY_META_RETRY_MAX", "5"))
-_META_RETRY_BASE_DELAY_MS = int(os.getenv("INFINITY_META_RETRY_BASE_DELAY_MS", "50"))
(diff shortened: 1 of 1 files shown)


qinling0210 (approved):
LGTM

PR #14576, opened 2026-05-05
Perf: push metadata filters down to Elasticsearch

### LINKED ISSUE #14412: [Bug]: Metadata Filtering is Performed In-Memory Instead of Leveraging Elasticsearch, Leading to Performance Bottlenecks
### Self Checks

a33d073

### RAGFlow image version

v.0.25.0

```Markdown

```

### Background

In the current implementation, the `meta_filter` function in `ragflow/common/metadata_utils.py` is responsible for filtering documents based on metadata. This function loads all metadata into memory and then applies user-defined filter conditions to each document using Python code, rather than translating user filters into Elasticsearch (ES) queries and running the filtering at the ES layer.

### Problem Description

- All filtering logic is performed by iterating over data in memory; no part of the filtering is offloaded to ES.
- User-defined metadata filters are not converted to ES query DSL, so ES's indexing and filtering capabilities are not utilized.
- As the number of documents grows, this approach leads to memory inefficiency and significant performance bottlenecks.
- The current design prevents Ragflow from scaling efficiently and leveraging distributed search engines such as ES for advanced filtering.

#### Code Reference

A sample of the filtering logic:
```python
def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
    # ...
    for input, docids in v2docs.items():
        if operator == "=":
            matched = input == value
        # ... many other operators handled locally
```

### Suggested Improvement

- If metadata is stored in ES, consider converting user-defined filter conditions to equivalent ES query DSL and sending them directly to ES for server-side filtering.
- In-memory filtering can remain as a fallback or for use cases with no ES backend or very small datasets.

### Expected Outcome

- Offloading complex metadata filtering to Elasticsearch would significantly improve performance and scalability in large-data scenarios.
- Users would be able to define filter conditions that are efficiently executed at the search engine layer, making Ragflow more robust for production deployments.

If needed, I can provide specific examples or further discuss potential design changes.

### Expected behavior

_No response_

### Steps to reproduce

```Markdown
ragflow/common/metadata_utils.py
def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
```

### Additional information

_No response_

sxxtony:

Fixes #14412.

`common.metadata_utils.meta_filter` evaluates user-defined metadata conditions in Python after `DocMetadataService.get_flatted_meta_by_kbs` loads the entire `meta_fields` table into memory. Past a few thousand documents per knowledge base this becomes a memory bottleneck and a wasted ES round-trip — every filter request currently fetches up to 10000 metadata rows even when the resulting `doc_ids` list is tiny.

This PR adds an ES push-down path that translates the same filter language into a `bool` query and returns just the matching document IDs.

**Changes**

- `common/metadata_es_filter.py` *(new)*: pure-Python translator from the RAGflow filter list to ES DSL. Covers every operator the in-memory path supports (`=`, `≠`, `>`, `<`, `≥`, `≤`, `in`, `not in`, `contains`, `not contains`, `start with`, `end with`, `empty`, `not empty`) with `case_insensitive: true` on `prefix` and `wildcard` for parity with the existing lower-cased Python comparisons. User wildcard metacharacters are escaped before being injected into `wildcard` patterns. Negative operators (`≠`, `not in`, `not contains`, ranges) are wrapped with an `exists` guard so they do not accidentally match documents missing the key, matching the legacy `if k not in metas` behaviour.
- `api/db/services/doc_metadata_service.py`: new `DocMetadataService.filter_doc_ids_by_meta_pushdown(kb_ids, filters, logic)` that returns the doc IDs ES matched, or `None` to signal the caller should fall back to the in-memory path. Returns `None` when the active doc store is Infinity (`meta_fields` is a JSON column, not a dotted-object mapping), when any filter cannot be expressed in DSL (`UnsupportedMetaFilter`), or when the ES request or metadata index lookup errors.
- `common/metadata_utils.py`: `apply_meta_data_filter` accepts an optional `kb_ids` argument. When supplied, conditions go through push-down first via a new `_try_meta_pushdown` helper; on `None` the function falls back to the original `meta_filter` call. Default behaviour is unchanged for callers that don't pass `kb_ids`.
- Updated all four callers (`agent/tools/retrieval.py`, `api/db/services/dialog_service.py` ×2, `api/apps/services/dataset_api_service.py`, `api/apps/sdk/session.py`) to forward `kb_ids` so the push-down path is exercised in production.
- `test/unit_test/common/test_metadata_es_filter.py` *(new)*: 35 unit tests covering every operator's DSL shape, value coercion (`ast.literal_eval`, lowercasing, ISO-date pass-through), wildcard escaping, OR-logic wrapping that protects negative clauses, and the doc-ID extractor.

**Behaviour preserved**

- The in-memory `meta_filter` is untouched and still services every fallback case (Infinity backend, unknown operators, ES outages).
- The eligibility / credibility / issue-multiplier semantics described in the LLM-driven `auto` and `semi_auto` modes still hand the LLM the full in-memory `metas` dict to choose conditions from. Only the *evaluation* of those generated conditions is pushed down.
- Existing tests in `test/unit_test/common/test_metadata_filter_operators.py` continue to pass (14/14).

**Test plan**

- `pytest test/unit_test/common/test_metadata_es_filter.py` — 35 passed.
- `pytest test/unit_test/common/test_metadata_filter_operators.py` — 14 passed.
- `ruff check` clean on every modified file.
- Reviewer please validate the ES query shapes against a live cluster — particularly `case_insensitive` on `wildcard` and `prefix` (requires ES 7.10+) and the `exists` + `must_not` pairing for `≠`.

**Notes**

- The first cut caps each push-down request at 10000 results, matching the existing `get_flatted_meta_by_kbs` limit, and logs a warning when the cap is hit. A `search_after` follow-up would let us drop the cap entirely once the push-down path is validated.
- Operator parity with the in-memory path is exact for the canonical unicode operators (`≥`, `≤`, `≠`) used internally; the ASCII aliases (`>=`, `<=`, `!=`) are normalised by `convert_conditions` before they reach the translator.

### Type of change

- [x] Performance Improvement


sxxtony:
@euvre Could you please review and add any feedbacks?

qinling0210 (commented):
Please test to make sure your change works.

Based on my test, apply_meta_data_filter() does not return the doc ids containing the requested meta data when we use Elasticsearch

sxxtony:
@qinling0210 thanks — reproduced your finding end-to-end, root-caused it, and pushed a fix in `d0c40fdc3`.

## Root cause

`conf/doc_meta_es_mapping.json` maps `meta_fields` as `{"type": "object", "dynamic": true}`. With default dynamic mapping, string values land as `text` (analyzed) with a `.keyword` multi-field. The translator was emitting `term`/`terms`/`prefix`/`wildcard` against the analyzed parent path, so any multi-word value silently failed: a filter for `author = "Alice Wonderland"` became `term: meta_fields.author = "alice wonderland"`, which searched the inverted index for that literal phrase token — but the index only held per-token entries `["alice", "wonderland"]`. No match.

The fix routes string values through `<field>.keyword` with `case_insensitive: true`. Numeric/bool values keep the parent path (those types don't get a `.keyword` sub-field). The empty/not-empty blank-string check also moved to `.keyword` because the analyzer drops empty tokens at index time.

## Before fix — repro

Real ES 8.11.3, real `doc_meta_es_mapping.json`, six representative docs (single-word, multi-word, list, numeric, missing-key, empty-string). Same filter run through both `meta_filter()` (in-memory) and `build_meta_filter_query()` (push-down):

```
[FAIL] eq single-word string
       in-memory: ['d_alice']
       ES query:  ['d_alice', 'd_alicew']      ← false positive
[FAIL] eq MULTI-WORD string (the failure mode you reported)
       in-memory: ['d_alicew']
       ES query:  []                            ← false negative, exact mode you saw
[FAIL] in list
       in-memory: ['d_alice', 'd_bob']
       ES query:  ['d_alice', 'd_alicew', 'd_bob']  ← FP + FN
[FAIL] not empty
       in-memory: ['d_alice', 'd_alicew', 'd_bob', 'd_carol']
       ES query:  ['d_alice', 'd_alicew', 'd_bob', 'd_carol', 'd_empty_author']

RESULT: 4/9 parity divergences
```

## After fix

```
[PASS] eq single-word string                  → ['d_alice']
[PASS] eq MULTI-WORD string                   → ['d_alicew']
[PASS] eq numeric                             → ['d_alice', 'd_no_author']
[PASS] contains substring of multi-word       → ['d_alicew']
[PASS] start with prefix                      → ['d_carol']
[PASS] end with suffix                        → ['d_carol']
[PASS] in list                                → ['d_alice', 'd_bob']
[PASS] range >                                → ['d_alice', 'd_bob', 'd_no_author']
[PASS] not empty                              → ['d_alice', 'd_alicew', 'd_bob', 'd_carol']

RESULT: all 9 cases match between push-down and in-memory.
```

## Unit tests

Updated the operator-shape assertions in `test_metadata_es_filter.py` to match `.keyword`-routed queries, and added a regression test (`test_equal_multiword_uses_keyword_subfield`) that locks in your reported failure mode. **42/42 pass.**

```
$ pytest test/unit_test/common/test_metadata_es_filter.py -v
============================== 42 passed in 0.24s ==============================
```

The other metadata test files (`test_metadata_filter_operators.py`, `test_apply_semi_auto_meta_data_filter.py`) also still pass; the 2 failures you might see in `test_apply_semi_auto` are a pre-existing import issue in the test environment (verified identical on `main`), unrelated to this change.

PTAL when you have a moment.

qinling0210 (approved):
LGTM

qinling0210:
@sxxtony ,  please analyze the CI failure, it failed on

 test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py::test_searchbots_retrieval_test_embedded_matrix_unit

PR #14577, opened 2026-05-05
fix(opensearch): implement doc-meta dispatch surface on OSConnection

### LINKED ISSUE #14570: [Bug]: OSConnection missing create_doc_meta_idx — Document metadata update fails on OpenSearch backend
### What happened

When using **AWS OpenSearch** (with Elasticsearch compatibility mode enabled) as the document engine (`DOC_ENGINE=opensearch`), updating document `meta_fields` via the API always fails with:

```
Failed to update metadata
```

Server-side logs show:

```
Error updating metadata for document <doc_id>: 'OSConnection' object has no attribute 'create_doc_meta_idx'
```

This affects both:
- `PATCH /api/v1/datasets/{dataset_id}/documents/{document_id}` with `{"meta_fields": {...}}`
- `POST /api/v1/datasets/{dataset_id}/metadata/update`

All other document operations (retrieval, parsing, name update, chunk management) work correctly on OpenSearch.

### What should have happened

Document metadata should be updatable via the API regardless of the document engine backend. The `OSConnection` class should implement `create_doc_meta_idx` (and any other metadata-related methods) just like `ESConnection` does.

### How to reproduce

1. Deploy RAGFlow v0.25.1 with `DOC_ENGINE=opensearch` and an AWS OpenSearch cluster
2. Upload and parse a document in a dataset
3. Try to update the document's metadata:

```bash
curl --request PATCH \
     --url http://<ragflow_host>/api/v1/datasets/<dataset_id>/documents/<document_id> \
     --header 'Authorization: Bearer <API_KEY>' \
     --header 'Content-Type: application/json' \
     --data '{"meta_fields": {"url": "https://example.com"}}'
```

4. Response: `{"code":102,"message":"Failed to update metadata"}`
5. Server log: `'OSConnection' object has no attribute 'create_doc_meta_idx'`

### Environment

- **RAGFlow version**: v0.25.1 (also reproduced on v0.24.0)
- **Document engine**: AWS OpenSearch (with Elasticsearch compatibility mode enabled)
- **DOC_ENGINE env**: `opensearch`
- **Deployment**: Kubernetes (EKS)

### Analysis

In `common/settings.py`, when `DOC_ENGINE=opensearch`, RAGFlow uses `rag.utils.opensearch_conn.OSConnection`. However, `OSConnection` does not implement `create_doc_meta_idx`, which is called during metadata updates. The `ESConnection` class (`rag.utils.es_conn`) has this method implemented.

We also tried setting `DOC_ENGINE=elasticsearch` to use `ESConnection` with our OpenSearch cluster (compatibility mode enabled), but the `elasticsearch` Python client library (v8.x) rejects the connection to OpenSearch at the client level (`Elasticsearch ... is unhealthy`).

### Suggested fix

Implement `create_doc_meta_idx` and other metadata-related methods in `rag/utils/opensearch_conn.py` (`OSConnection` class), mirroring the implementation in `rag/utils/es_conn.py` (`ESConnection` class).

tmimmanuel:

Fixes #14570  On OpenSearch backends (`DOC_ENGINE=opensearch`) every document-metadata write failed with `'OSConnection' object has no attribute 'create_doc_meta_idx'`, so both `PATCH /api/v1/datasets/{ds}/documents/{doc}` with `meta_fields` and `POST /api/v1/datasets/{ds}/metadata/update` were unusable while every other document operation (retrieval, parsing, name update, chunk management) worked correctly on the same OpenSearch cluster.

The bug runs deeper than the missing method name in the error message suggests. `DocMetadataService` also reached into `settings.docStoreConn.es.*` directly for the index refresh, the scripted partial update, and the count call, which means that even after adding `create_doc_meta_idx` to `OSConnection` the very next call in the same metadata flow would still raise `AttributeError` because `OSConnection` exposes `self.os` rather than `self.es`. Fixing only the reported symptom would have moved the failure one line down without restoring the feature.

This PR adds a uniform document-metadata dispatch surface to both connection classes so they present the same abstract API, and routes the service layer through that surface via `getattr` guards instead of poking at backend-specific attributes. The four new methods on `OSConnection` and `ESConnectionBase` are `create_doc_meta_idx`, `refresh_idx`, `count_idx`, and `replace_meta_fields`. `OSConnection.create_doc_meta_idx` reuses the existing `conf/doc_meta_es_mapping.json` schema in the OpenSearch `body=` form because OpenSearch and Elasticsearch share the same index-creation payload, and `replace_meta_fields` emits a full scripted assignment (`ctx._source.meta_fields = params.meta_fields`) on both backends so removed keys actually disappear instead of being preserved by deep-merge semantics.

The `getattr`-guarded dispatch in `DocMetadataService` keeps the existing fall-through paths intact for Infinity and OceanBase, which continue to rely on their search-based count fallback and on the delete-then-insert metadata replacement they used before, so this change is strictly additive for those two backends.

Verification: `pytest test/unit_test/rag/utils/test_opensearch_doc_meta.py` runs 16 new unit tests that pass locally and pin the `OSConnection` dispatch surface, the `create_doc_meta_idx` short-circuit when the index already exists, the mapping-file payload routing, the `IndicesClient.create` failure path, the `refresh_idx` and `count_idx` success and error sentinels, and the full-assignment script emitted by `replace_meta_fields`. The test module stubs `common.settings` and `rag.nlp` at import time so the suite runs without the heavy backend SDKs that the rest of the repository pulls in transitively.

### Type of change


tmimmanuel:
@wangq8  @yingfeng 
Please review my PR. Fixed issue, CI is clean.

wangq8 (approved):

PR #14600, opened 2026-05-06
Fix: path-aware reset in canvas.run() to preserve cross-run outputs

wanghualoong:
## Summary

- When an agent workflow has multiple `UserFillUp` pause points, `canvas.run()` calls `reset(True)` on **all** components at the start of each run. This clears outputs from components that completed in prior runs, so downstream references like `{Agent:XXX@content}` resolve to `None`.
- This fix only resets components on the **current execution path** (`self.path`), preserving outputs from previously completed components.

## Problem

In a multi-step agent (e.g. draft email → user confirms → send email):

1. First `run()`: Agent drafts content, UserFillUp pauses for user input → Agent output is saved
2. Second `run()`: User submits input, but `reset(True)` clears **all** components including the Agent that already completed
3. Email component references `{Agent:XXX@content}` → gets `None` instead of the draft

This affects **all** agents that reference upstream component outputs after a UserFillUp pause point.

## Fix

```python
# Before: reset ALL components
for k, cpn in self.components.items():
    self.components[k]["obj"].reset(True)

# After: only reset components on current execution path
path_set = set(self.path)
for k, cpn in self.components.items():
    if k in path_set:
        self.components[k]["obj"].reset(True)
```

`self.path` already tracks the current execution path. For agents without UserFillUp (single run), `path` contains all components, so behavior is unchanged.

## Test plan

- [x] Agent with single UserFillUp: outputs from prior components are preserved after resume
- [x] Agent with multiple UserFillUp: each resume preserves all previously completed outputs
- [x] Agent without UserFillUp: behavior unchanged (all components in path, all reset)
- [x] Webhook-triggered agents: unaffected (path includes all components on first run)


Magicbook1108 (approved):
LGTM

PR #14616, opened 2026-05-07
Fix: support tool call config

buua436:
support tool call config

### Type of change


Lynn-Inf (approved):
LGTM

PR #14625, opened 2026-05-07
Fix: enforce tenant authorization on document download endpoint (#14618)

### LINKED ISSUE #14618: [Bug]: Cross-tenant document download via `GET /v1/document/get/<doc_id>` (missing tenant authorization)
### Self Checks

15dcdd7b5

### RAGFlow image version

nightly (main @ 15dcdd7b5)

```Markdown
- Component: Python backend (Flask/Quart) — api/apps/document_app.py
- Affected endpoint: GET /v1/document/get/<doc_id>
- Auth decorator on route: @login_required only (no tenant scoping)
```

### Actual behavior

While reading through `api/apps/document_app.py` I noticed the `/get/<doc_id>` route is gated by `@login_required` and nothing else. It calls `DocumentService.get_by_id(doc_id)` (a global lookup, no tenant filter) and then streams the file straight from MinIO via `File2DocumentService.get_storage_address(...)` and `settings.STORAGE_IMPL.get(...)`. There's no check anywhere in between to confirm the document's KB actually belongs to the caller.

In practice this means any logged-in user — including a brand new self-serve signup — can download another tenant's document just by knowing the `doc_id`. It's the actual file content, not metadata.

What makes me fairly confident this is a real bug and not intentional: every other place in the codebase that resolves a document by id does enforce tenant scoping. The SDK route in `api/apps/sdk/doc.py` does `KnowledgebaseService.query(id=doc[0].kb_id, tenant_id=tenant_id)` right after `get_by_id`, and `api/apps/restful_apis/document_api.py` does the same at several call sites (around lines 174, 1156, 1796). The web-app route is the odd one out.

There's also already a helper for exactly this check — `DocumentService.accessible(doc_id, user_id)` in `api/db/services/document_service.py` — which joins through `UserTenant` so it also handles users who belong to multiple tenants / shared team KBs. So the fix can reuse existing code.

### Expected behavior

Hitting `GET /v1/document/get/<doc_id>` should only succeed when the caller actually has access to that document via their tenant/team membership. Otherwise it should return an authorization error rather than the file bytes.

### Steps to reproduce

```Markdown
1. Set up two accounts:
   - Tenant A: any logged-in user (a fresh signup is enough).
   - Tenant B: a separate tenant with at least one uploaded document. Note its doc_id (call it DOC_B).

2. From Tenant A's session, request Tenant B's document directly:

   curl -X GET "https://<host>/v1/document/get/<DOC_B>" \
     -H "Cookie: session=<TENANT_A_SESSION>" \
     --output leaked.bin

3. The response body contains Tenant B's file, even though Tenant A has no
   membership in Tenant B. doc_ids are UUID-shaped so they aren't trivially
   enumerable, but they do leak through logs, support tickets, screenshots,
   browser history, and other endpoints' responses.
```

### Additional information

Suggested fix — add an authorization check in `api/apps/document_app.py::get` after `DocumentService.get_by_id` and before the storage read. Either of these would work, but the first is closer to what the rest of the codebase does for documents:

```python
# Preferred — also handles shared/team KBs
    return get_data_error_result(message="No authorization.")
```

```python
# Minimum — matches the SDK pattern
if not KnowledgebaseService.query(id=doc.kb_id, tenant_id=current_user.id):
    return get_data_error_result(message="No authorization.")
```

It's probably worth a quick sweep of the other `@login_required` routes in the web-app blueprints for the same `get_by_id → storage read` shape, in case this isn't the only one.

Related issues I found while searching (same class of bug, different endpoints — not duplicates of this one):

- #14590 — cross-tenant chatbot access via valid API key.
- #14292 — IDOR on file ancestry metadata via `file_id`.
- #14534 — memory APIs missing authorization checks on caller-supplied `tenant_id` / `memory_id`.

Worth calling out vs. #14292 specifically: that report is about metadata disclosure on the Go file API, whereas this one is the Python route that returns the actual file contents, so the impact is higher.

web-dev0521:

Closes #14618.

The `GET /v1/document/get/<doc_id>` endpoint in `api/apps/document_app.py` was protected only by `@login_required` and called `DocumentService.get_by_id(doc_id)` without verifying that the document's knowledge base belonged to the requesting user's tenant. Any authenticated user who knew (or guessed) a document ID could download files belonging to any other tenant — a cross-tenant IDOR.

This PR adds a `DocumentService.accessible(doc_id, current_user.id)` check before serving the file. The helper already exists and joins `Document` → `Knowledgebase` → `UserTenant` to verify the requesting user belongs to the tenant that owns the document's KB. The same pattern is already used by `api/apps/restful_apis/document_api.py` and mirrors the tenant scoping in the SDK route at `api/apps/sdk/doc.py`.

The check returns the existing `"Document not found!"` error for both non-existent and inaccessible documents, so attackers cannot use the response to enumerate valid doc IDs across tenants.

### Type of change

- [x] Other (please describe): Security fix (cross-tenant IDOR / authorization bypass)

### DIFF
--- a/api/apps/restful_apis/document_api.py
+++ b/api/apps/restful_apis/document_api.py
@@ -23,7 +23,7 @@ from quart import request, make_response
 from peewee import OperationalError
 from pydantic import ValidationError
 
-from api.apps import login_required
+from api.apps import current_user, login_required
 from api.constants import FILE_NAME_LEN_LIMIT, IMG_BASE64_PREFIX
 from api.apps.services.document_api_service import validate_document_update_fields, map_doc_keys, \
     map_doc_keys_with_run_status, update_document_name_only, update_chunk_method, update_document_status_only, \
@@ -1859,7 +1859,16 @@ async def batch_update_document_status(tenant_id, dataset_id):
 @manager.route("/documents/<doc_id>/preview", methods=["GET"])  # noqa: F821
 @login_required
 async def get(doc_id):
+    """Return the raw file bytes for a document the requesting user is authorized to read.
+
+    The user must belong to the tenant that owns the document's knowledge base; otherwise
+    the response is indistinguishable from a missing document to avoid cross-tenant ID
+    enumeration.
+    """
     try:
+        if not DocumentService.accessible(doc_id, current_user.id):
+            return get_data_error_result(message="Document not found!")
+
         e, doc = DocumentService.get_by_id(doc_id)
         if not e:
             return get_data_error_result(message="Document not found!")
--- a/api/apps/restful_apis/document_api.py
+++ b/api/apps/restful_apis/document_api.py
@@ -1893,10 +1893,19 @@ async def get(doc_id):
 @login_required
 @add_tenant_id_to_kwargs
 async def download_attachment(tenant_id=None, doc_id=None, attachment_id=None):
+    """Stream a document's underlying file to the requesting user.
+
+    Mirrors the authorization model of the preview endpoint: the user must belong
+    to the tenant that owns the document's knowledge base. A denial returns the
+    same "Document not found!" response so the endpoint cannot be used to
+    enumerate doc ids across tenants.
+    """
     try:
         # Keep backward compatibility with older callers and unit tests that still
         # pass `attachment_id` instead of the route parameter name.
         doc_id = doc_id or attachment_id
+        if not DocumentService.accessible(doc_id, current_user.id):
+            return get_data_error_result(message="Document not found!")
         ext = request.args.get("ext", "markdown")
         data = await thread_pool_exec(settings.STORAGE_IMPL.get, tenant_id, doc_id)
         response = await make_response(data)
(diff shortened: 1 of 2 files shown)


Lobster-0429:
Hi, @wangq8 ,

wangq8:
@web-dev0521 this API have been moved to new place, please raise PR to new location.

Lobster-0429:
Hi, @wangq8 ,
Thanks for your attention.
I've just moved the API and resolve the conflicts.
Could you please review again?

wangq8:
@web-dev0521 should you add to both `preview` and `download` endpoints?

Lobster-0429:
Good catch, thanks @wangq8 🙏

You're right - /preview and /download are basically the same surface from a security standpoint, so they should share the same check. I just pushed 8f40ce92c which adds the same DocumentService.accessible(doc_id, current_user.id) guard to the download handler, and I also extended the unit test to cover the denial path the same way I did for preview.

Let me know if you'd like me to tweak anything - happy to iterate.

wangq8:
@web-dev0521 looks good to me.

Lobster-0429:
HI, @wangq8 ,
All tests were passed.

PR #14636, opened 2026-05-07
fix(go): implement ListModels and CheckConnection in NVIDIA driver

### LINKED ISSUE #14635: [Bug]: NVIDIA Go driver has ListModels and CheckConnection stubbed
### Self Checks

050113482

### RAGFlow image version

main / nightly

```Markdown
Affected layer: Go model provider layer
Affected file: internal/entity/models/nvidia.go
```

### Actual behavior

The NVIDIA Go driver was added in #14623. ChatWithMessages and ChatStreamlyWithSender are real implementations against the NVIDIA NIM API at https://integrate.api.nvidia.com/v1, but two methods on the same driver are still stubs that return a hardcoded "no such method" error:

  // nvidia.go
  func (n *NvidiaModel) ListModels(apiConfig *APIConfig) ([]string, error) {
      return nil, fmt.Errorf("%s, no such method", n.Name())
  }

  func (n *NvidiaModel) CheckConnection(apiConfig *APIConfig) error {
      return fmt.Errorf("no such method")
  }

So when a tenant configures NVIDIA as a model provider in the Go layer:

- The "List models" path that auto-populates the model picker always fails. Users have to type the full model id by hand (e.g. "abacusai/dracarys-llama-3.1-70b-instruct").
- The "Check connection" button always fails, even when the base URL is reachable and the API key is accepted.

Other Go drivers already follow the standard pattern. xai, moonshot, deepseek, aliyun, and gitee call /models for
ListModels, and CheckConnection just calls ListModels and returns its error.

NVIDIA NIM is OpenAI-compatible. The /v1/models endpoint is public and works with the same Bearer token used for chat, so the standard pattern fits with no extra work.

The conf/models/nvidia.json file already wires the "models"
URL suffix:

  "url_suffix": {
    "chat": "chat/completions",
    "models": "models"
  }

So no config change is needed.

### Expected behavior

ListModels on the NVIDIA driver should call GET ${BaseURL}/${URLSuffix.Models} with Authorization: 
Bearer <api_key> and return the list of available model ids,
the same way the xai, moonshot, and openai drivers do.

CheckConnection should call ListModels and return its error.

### Steps to reproduce

```Markdown
1. In the Go layer, configure NVIDIA as a model provider with a real base URL (https://integrate.api.nvidia.com/v1 by default) and a real API key.
2. Try "Check connection" or open the model picker that lists available models for this provider.
3. Both paths fail with "no such method", regardless of whether the API key and base URL are valid.
4. Compare with the moonshot or xai instance under the same UI flow, where both paths work.
```

### Additional information

_No response_

pandadev66:

The NVIDIA Go driver added in #14623 has a real chat path, but \`ListModels\` and \`CheckConnection\` are stubs that always return \`no such method\`. So:

- The model picker cannot auto-populate available NVIDIA NIM model ids. Users have to type the full id by hand (e.g.
  \`abacusai/dracarys-llama-3.1-70b-instruct\`).
- The "Check connection" button always fails for NVIDIA, even when the base URL is reachable and the API key is accepted.

NVIDIA NIM is OpenAI-compatible. \`/v1/models\` works with the same Bearer token used for chat. The
\`conf/models/nvidia.json\` file already wires the \`models\` url_suffix, so no config change is needed.

### What this PR includes

- \`internal/entity/models/nvidia.go\`:
  - \`ListModels\` now calls
    \`GET ${BaseURL}/${URLSuffix.Models}\`, parses
    \`response.data[*].id\`, and returns the list. Same shape
    as the moonshot, xai, and openai drivers.
  - \`CheckConnection\` now calls \`ListModels\` and returns its
    error. Same pattern xai, moonshot, deepseek, aliyun, and
    gitee already use.

\`Balance\`, \`Encode\`, and \`Rerank\` are still stubs in this PR and can be added in follow-ups.

No JSON change. No factory change. No interface change.

### How the implementation works

- Region resolution falls back to \`default\` when the supplied region is unknown, so a stray region value does not break a  valid request.
- The Authorization header is only set when \`apiConfig\` and \`ApiKey\` are non-nil and non-empty. This avoids a nil-pointer dereference and lets self-hosted NIM deployments without a key still work.
- Non-200 responses propagate the upstream status line and body so the user sees a real error message.

### Type of change

### How was this tested?

- \`go build ./internal/entity/models/...\` in a clean go 1.25 image (the go.mod minimum) returns exit 0.
- The full method set on \`NvidiaModel\` still matches the \`ModelDriver\` interface.
- Pattern parity with the existing xai, moonshot, deepseek, aliyun, gitee, and openai drivers.

Closes #14635

### DIFF
--- a/internal/entity/models/nvidia.go
+++ b/internal/entity/models/nvidia.go
@@ -332,14 +332,85 @@ func (n NvidiaModel) Rerank(modelName *string, query string, texts []string, api
 	return nil, fmt.Errorf("no such method")
 }
 
+// ListModels calls /v1/models on the configured NVIDIA NIM base URL
+// and returns the list of available model ids. The endpoint is
+// OpenAI-compatible, so the parsing follows the same shape used by
+// the moonshot, xai, and openai drivers.
 func (n NvidiaModel) ListModels(apiConfig *APIConfig) ([]string, error) {
-	return nil, fmt.Errorf("no such method")
+	var region = "default"
+	if apiConfig != nil && apiConfig.Region != nil && *apiConfig.Region != "" {
+		region = *apiConfig.Region
+	}
+
+	baseURL := n.BaseURL[region]
+	if baseURL == "" {
+		baseURL = n.BaseURL["default"]
+	}
+	if baseURL == "" {
+		return nil, fmt.Errorf("nvidia: no base URL configured for region %q", region)
+	}
+
+	url := fmt.Sprintf("%s/%s", baseURL, n.URLSuffix.Models)
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	if apiConfig != nil && apiConfig.ApiKey != nil && *apiConfig.ApiKey != "" {
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", *apiConfig.ApiKey))
+	}
+
+	resp, err := n.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Nvidia models API error: %s, body: %s", resp.Status, string(body))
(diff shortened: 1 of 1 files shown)


pandadev66:
Hi @Magicbook1108, gentle ping for review. Same pattern as the recently merged xAI / moonshot / openai ListModels and the CheckConnection wiring used by xai / moonshot / deepseek / aliyun / gitee. Scope is one file, and it closes #14635. Thanks!

Haruko386 (approved):
Hi @pandadev66, LGTM

PR #14643, opened 2026-05-07
Go: implement Balance in SiliconFlow driver

### LINKED ISSUE #14642: [Feature Request]: Implement Balance in the SiliconFlow Go driver
### Self Checks

```Markdown
Yes. The SiliconFlow Go driver in internal/entity/models/siliconflow.go has a stub Balance method that returns "no such method", even though SiliconFlow exposes a public balance endpoint that works with the same Bearer
token used for chat:

  GET https://api.siliconflow.cn/v1/user/info

The endpoint returns the user's account info including the current balance in CNY, so it gives users exactly what the
"Balance" panel in the model provider UI is meant to show. Right now a SiliconFlow tenant cannot see how much credit is
left from inside RAGFlow. The "Balance" button always shows an error, while it works for Moonshot and Gitee. Users have to log into the SiliconFlow console in another tab.
```

Implement SiliconflowModel.Balance in internal/entity/models/siliconflow.go so the "Balance" panel works for SiliconFlow the same way it works for Moonshot, Gitee, and (in flight) DeepSeek.

The driver should:

- GET /user/info on the same base URL the chat path uses, using the existing URLSuffix.Status (or a new url_suffix field, see below).
- Send Authorization: Bearer <api_key>, the same way the chat path already does.
- Parse the response and return a map[string]interface{} that the UI can render. The shape used by the Moonshot driver is:

    {
      "balance":  <float>,
      "currency": "CNY",
    }

SiliconFlow returns balance as a string under data.balance, so the driver should parse it with strconv.ParseFloat the same way the DeepSeek balance work does.

The conf/models/siliconflow.json file needs a small update so the driver can read the URL suffix from config:

  "url_suffix": {
    "chat": "chat/completions",
    "models": "models",
    "embedding": "embeddings",
    "rerank": "rerank",
    "balance": "user/info"
  }

No new public API. No interface change. Just a real implementation behind an existing method.

_No response_

```Markdown
No new docs are needed. SiliconFlow is already listed in the model provider picker. This issue just makes the existing
"Balance" button work for SiliconFlow, the way it already works for Moonshot and Gitee.

Use case: a tenant adds SiliconFlow as a model provider in RAGFlow, clicks "Balance" on the instance, and sees their
current balance in CNY straight from api.siliconflow.cn, without having to open the SiliconFlow console in another tab.
```

### Additional information

_No response_

pandadev66:

The SiliconFlow Go driver shipped with a stub \`Balance\` method that returned \`no such method\`, even though SiliconFlow exposes a public \`GET /v1/user/info\` endpoint that returns the account balance per currency.

So the "Balance" panel in the model provider UI always shows an error for SiliconFlow tenants, while it already works for
Moonshot and Gitee. This PR fills the gap.

### What this PR includes

- \`conf/models/siliconflow.json\`: add \`\"balance\": \"user/info\"\` under \`url_suffix\` so the driver builds the URL from config.
- \`internal/entity/models/siliconflow.go\`: replace the \`Balance\` stub with a real implementation. Adds a small local  response type that matches the upstream shape.

No factory change. No interface change.

### How the driver works

- Validate \`apiConfig\` and the API key, resolve the region with a default fallback, and build the URL from \`BaseURL[region] + URLSuffix.Balance\`.
- GET the URL with \`Authorization: Bearer <api_key>\`.
- Parse the upstream response. SiliconFlow returns balance fields as strings, so the driver parses them with  \`strconv.ParseFloat\`. It prefers \`totalBalance\` over  \`balance\` when both are present.
- Return \`{\"balance\": <float>, \"currency\": \"CNY\"}\`, the same shape the Moonshot driver returns. The UI can render it
  with no provider-specific code.

### Edge cases

- Missing or empty API key returns a clear local error before any HTTP call.
- An unknown region falls back to the default base URL.
- Empty \`balance\` and \`totalBalance\` returns a clear "no balance info in response" error rather than a zero-value silent success.
- Non-numeric balance string returns a clear parse error.
- Non-200 responses propagate the upstream status line and body.

### Type of change

### How was this tested?

- \`go build ./internal/entity/models/...\` in a clean go 1.25 image returns exit 0.
- The full method set on \`SiliconflowModel\` still matches the \`ModelDriver\` interface.
- Pattern parity with the existing Moonshot and Gitee Balance implementations.

Closes #14642

### DIFF
--- a/internal/entity/models/siliconflow.go
+++ b/internal/entity/models/siliconflow.go
@@ -531,7 +531,7 @@ func (z *SiliconflowModel) ListModels(apiConfig *APIConfig) ([]string, error) {
 
 type siliconflowBalanceResponse struct {
 	Code    int    `json:"code"`
-	Status  bool   `json:"status"`
+	Status  string `json:"status"`
 	Message string `json:"message"`
 	Data    struct {
 		Balance      string `json:"balance"`
@@ -587,6 +587,14 @@ func (z *SiliconflowModel) Balance(apiConfig *APIConfig) (map[string]interface{}
 	var parsed siliconflowBalanceResponse
 	if err = json.Unmarshal(body, &parsed); err != nil {
 		return nil, fmt.Errorf("failed to parse response: %w", err)
+	}
+
+	if parsed.Code != 20000 {
+		msg := parsed.Message
+		if msg == "" {
+			msg = "unknown API error"
+		}
+		return nil, fmt.Errorf("SiliconFlow API error (code %d): %s", parsed.Code, msg)
 	}
 
 	raw := parsed.Data.TotalBalance
--- a/internal/entity/models/siliconflow.go
+++ b/internal/entity/models/siliconflow.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net/http"
 	"ragflow/internal/common"
+	"strconv"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


pandadev66:
Hi @Magicbook1108, gentle ping for review. Same pattern as the Moonshot and Gitee Balance implementations, scope is two files (one config, one driver), and it closes #14642. Thanks!

internal/entity/models/siliconflow.go
```diff
@@ -528,8 +529,83 @@ func (z *SiliconflowModel) ListModels(apiConfig *APIConfig) ([]string, error) {
...
+type siliconflowBalanceResponse struct {
+	Code    int    `json:"code"`
+	Status  bool   `json:"status"`
+	Message string `json:"message"`
+	Data    struct {
+		Balance      string `json:"balance"`
+		TotalBalance string `json:"totalBalance"`
+		Status       string `json:"status"`
+	} `json:"data"`
+}
```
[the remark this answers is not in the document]
pandadev66: Hi, thanks for the catch! Fixed in commit 3a5005d, status is now a string and the API-level error is surfaced with parsed.Message.

internal/entity/models/siliconflow.go
```diff
@@ -528,8 +529,91 @@ func (z *SiliconflowModel) ListModels(apiConfig *APIConfig) ([]string, error) {
...
+	Code    int    `json:"code"`
+	Status  string `json:"status"`
+	Message string `json:"message"`
+	Data    struct {
+		Balance      string `json:"balance"`
+		TotalBalance string `json:"totalBalance"`
+		Status       string `json:"status"`
+	} `json:"data"`
+}
+
```
Haruko386, line 541: According to
```json
{
  "code": 20000,
  "message": "OK",
  "status": true,
  "data": {
    "balance": "0.88",
    "chargeBalance": "88.00",
    "totalBalance": "88.88"
  }
}
``` 
`status` is a `bool`

your code: 

```go
type siliconflowBalanceResponse struct {
    Status string `json:"status"`
}
```

Haruko386 (changes requested):
Hi @pandadev66 ,

I found an issue in the SiliconFlow balance response parsing logic.

The `/v1/user/info` endpoint is working correctly and returns a response like:

```json
{
  "code": 20000,
  "message": "OK",
  "status": true,
  "data": {
    "balance": "0.88"
  }
}
```

However, the current implementation defines the `status` field as a `string`, while the actual API response returns a boolean value.

This causes the following error during JSON unmarshalling:

```text
json: cannot unmarshal bool into Go struct field siliconflowBalanceResponse.status of type string
```

The struct field should be updated from:

```go
Status string `json:"status"`
```

to:

```go
Status bool `json:"status"`
```

After this change, the balance endpoint should work correctly.

internal/entity/models/siliconflow.go
pandadev66: Hi @Haruko386, thanks so much for testing this against the live endpoint and catching it! Fixed in commit 5500eba, status is back to bool and the API-level check now uses !parsed.Status.

Haruko386 (approved):
@pandadev66 LGTM

PR #14645, opened 2026-05-07
Fix: private dataset authorization bypass in shared dataset access checks

jony376:
### Related issues
Closes #14644

This PR fixes an authorization bug where datasets marked with `permission = me` could still be accessed by other members of the same tenant through APIs that relied on `KnowledgebaseService.accessible()` or `DocumentService.accessible()`.

Before this change, those shared access helpers only checked tenant membership and did not enforce the dataset's permission mode. As a result, a non-owner who knew a private `dataset_id` could still reach downstream document and chunk operations even though the dataset was intended to be owner-only.

This change updates the central access checks so that:

- dataset owners always retain access
- joined tenant members only get access when the dataset permission is `TEAM`
- private datasets with `permission = me` remain inaccessible to non-owners
- document-level access follows the same dataset permission rules

The PR also adds regression coverage for private-vs-team dataset access behavior.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Testing

- Added `test/unit_test/api/db/services/test_dataset_access_permissions.py`
- Attempted to run: `python -m pytest test\\unit_test\\api\\db\\services\\test_dataset_access_permissions.py -q`
- Local execution in this workspace is currently blocked during test collection because the environment is missing the `strenum` dependency


wangq8:
LGTM

wangq8:
Hi @jony376 please help to fix the unit test failrue.

pekar9781:
Hi, @wangq8
Thanks your feedback. I fixed test failure issue and conflicts. Please check this PR. thanks.

PR #14663, opened 2026-05-08
Fix: add compatibility route for document download under /v1

buua436:

add compatibility route for document download under /v1

### Type of change


wangq8:
LGTM

PR #14668, opened 2026-05-08
Refactor: tidy up ThreadPoolExecutor lifecycle in file_service and task executor

### LINKED ISSUE #14622: [Feature Request]: Tidy up ThreadPoolExecutor lifecycle in file_service.py and task_executor.py
### Self Checks

```Markdown
Up front, so reviewers don't have to read the rest to find out: this isn't a bug report. Nothing here is broken at runtime. I noticed three spots while reading the file/document handling code where a `ThreadPoolExecutor` is created without a `with` block or an explicit `shutdown()` call, and I initially thought "that's a leak." It isn't — CPython cleans up the worker threads when the executor is GC'd, via the `_threads_queues` weakref + atexit machinery, so threads don't pile up across calls. I want to flag that explicitly so this doesn't get triaged as a vuln and dismissed.

What's left after stripping the false alarm is a code-hygiene ask. The lifecycles are harder to reason about than they need to be, and one of the three spots could just drop the executor entirely in favor of `asyncio.to_thread`.

The three spots:

- `api/db/services/file_service.py::get_files` (around lines 790–805) creates `exe = ThreadPoolExecutor(max_workers=5)` and lets GC handle it.
- `api/db/services/file_service.py::parse_docs` (around lines 557–566) does the same with `max_workers=12`.
- `rag/svr/task_executor.py::do_handle_task` creates `executor = concurrent.futures.ThreadPoolExecutor()` near the top of the function (around line 1076) and shuts it down in `finally` with `executor.shutdown(wait=False)` (around line 1287). The executor is only ever used on one branch (standard chunking) and only ever receives a single task — `build_TOC`.

A couple of things that are worth ruling out, since they came up while I was looking at this:

- `executor.shutdown(wait=False)` in `task_executor.py` doesn't kill in-flight `build_TOC` work. Running tasks complete in their worker threads regardless of how shutdown is called. On the happy path `toc_thread.result()` is awaited before `finally` runs anyway, so there's nothing in flight at shutdown time. On cancellation/error paths the result is dropped, but build_TOC still finishes — which is what you want for cancellation.
- The default-sized executor in `task_executor.py` doesn't actually allocate threads up front; threads are created lazily on `submit()`. So branches that never reach the submit don't pay any cost.
```

Make the lifecycles explicit (or get rid of the executor when something simpler does the job). Concretely:

- `get_files` and `parse_docs` keep their executors but use `with ... as exe:` so cleanup is deterministic and exception-safe.
- `do_handle_task` drops the executor and uses `asyncio.to_thread` for `build_TOC`, since the surrounding function is already `async` and only one task ever runs through it.

For the two `file_service.py` functions it's a straight wrap-and-reindent — about 16 lines touched in `get_files` and ~10 in `parse_docs`, no behavior change. The existing `.result()` calls already happen before the executor would have been shut down, so the implicit `shutdown(wait=True)` at the end of the `with` block has nothing left to wait for.

For `do_handle_task` the clean version looks something like this:

```python
toc_thread = asyncio.create_task(
    asyncio.to_thread(build_TOC, task, chunks, progress_callback)
)
...
d = await toc_thread
```

That lets you delete the executor at line 1076 and the `executor.shutdown(wait=False)` in the `finally`. Net change is around 4 lines edited, 2 deleted; `import asyncio` is already there.

If reaching into the asyncio path feels like too much for what's mostly a tidy-up, a smaller alternative is to leave it as a thread pool but move construction inside the `else` branch that actually uses it and wrap it in a `with` block. That's ~6 lines in a narrow region instead of 200+ re-indented for a function-wide `with`.

```Markdown
No public API change, no behavior change, no migration story. Pure internal cleanup, invisible to anyone using RAGFlow.
```

### Additional information

Total diff for the recommended approach is on the order of 30 lines across two files. I'm happy to send a PR if this is worth doing — and equally happy if maintainers want to leave it alone, since again, nothing is actually broken.

web-dev0521:
## Summary
- Wrap the `ThreadPoolExecutor` instances in `FileService.parse_docs` and `FileService.get_files` with `with ... as exe:` blocks for deterministic cleanup
- Replace the `concurrent.futures.ThreadPoolExecutor` in `do_handle_task` with `asyncio.create_task(asyncio.to_thread(build_TOC, ...))`, preserving the existing parallelism with chunk insertion while leveraging the surrounding async context
- Drop the now-unused `import concurrent` and the `executor.shutdown(wait=False)` call in the `finally` block

Closes #14622.

No behavioral change, no public API change. Net diff: ~19 insertions / 25 deletions across two files.

## Test plan
- [ ] `uv run ruff check api/db/services/file_service.py rag/svr/task_executor.py` passes
- [ ] Upload a multi-file batch through the chat/file endpoint and confirm `FileService.parse_docs` still returns combined parsed text
- [ ] Trigger `FileService.get_files` via the chat reference flow with a mix of image and non-image files; verify both `raw=True` and `raw=False` paths return correctly
- [ ] Run a `naive`-parser document task with `toc_extraction: true` and confirm the TOC chunk is generated and inserted exactly as before
- [ ] Run a `naive`-parser document task with `toc_extraction: false` and confirm the path with `toc_thread = None` is unaffected
- [ ] Cancel a running task to exercise the `finally` block and confirm cleanup still works without the executor shutdown call

### DIFF
--- a/api/db/services/file_service.py
+++ b/api/db/services/file_service.py
@@ -555,14 +555,9 @@ class FileService(CommonService):
 
     @staticmethod
     def parse_docs(file_objs, user_id):
-        exe = ThreadPoolExecutor(max_workers=12)
-        threads = []
-        for file in file_objs:
-            threads.append(exe.submit(FileService.parse, file.filename, file.read(), False))
-
-        res = []
-        for th in threads:
-            res.append(th.result())
+        with ThreadPoolExecutor(max_workers=12) as exe:
+            threads = [exe.submit(FileService.parse, file.filename, file.read(), False) for file in file_objs]
+            res = [th.result() for th in threads]
 
         return "\n\n".join(res)
 
@@ -787,19 +782,21 @@ class FileService(CommonService):
         def image_to_base64(file):
             return "data:{};base64,{}".format(file["mime_type"],
                                         base64.b64encode(FileService.get_blob(file["created_by"], file["id"])).decode("utf-8"))
-        exe = ThreadPoolExecutor(max_workers=5)
         threads = []
         imgs = []
-        for file in files:
-            if file["mime_type"].find("image") >=0:
-                if raw:
-                    imgs.append(FileService.get_blob(file["created_by"], file["id"]))
-                else:
-                    threads.append(exe.submit(image_to_base64, file))
-                continue
-            threads.append(exe.submit(FileService.parse, file["name"], FileService.get_blob(file["created_by"], file["id"]), True, file["created_by"], layout_recognize))
-    
+        with ThreadPoolExecutor(max_workers=5) as exe:
+            for file in files:
+                if file["mime_type"].find("image") >=0:
+                    if raw:
+                        imgs.append(FileService.get_blob(file["created_by"], file["id"]))
+                    else:
+                        threads.append(exe.submit(image_to_base64, file))
+                    continue
+                threads.append(exe.submit(FileService.parse, file["name"], FileService.get_blob(file["created_by"], file["id"]), True, file["created_by"], layout_recognize))
+
+            results = [th.result() for th in threads]
+
         if raw:
-            return [th.result() for th in threads], imgs
+            return results, imgs
         else:
-            return [th.result() for th in threads]
+            return results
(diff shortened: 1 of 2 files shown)


Lobster-0429:
HI, @wangq8 
Thanks for your time.

wangq8:
LGTM.

Lobster-0429:
Hi, @wangq8 ,
All CI tests are successfully passed.

PR #14670, opened 2026-05-08
GraphRAG feature - Part 1 - add spacy to extract entity and relation

wangq8:

GraphRAG feature - Part 1 - add spacy to extract entity and relation

image

### Type of change


wangq8:
spacy -> ner

PR #14673, opened 2026-05-08
Fix(Go): correct Name() and region URL fallback in Aliyun driver

Joseff531:

Two bugs in the Aliyun Go driver:

1. **`Name()` returns `"siliconflow"`** — a copy-paste bug from when the driver was created. `Name()` is used in error messages and log output, so every Aliyun error incorrectly attributed itself to SiliconFlow.

2. **Silent empty URL for unknown regions in `ChatWithMessages`, `ChatStreamlyWithSender`, and `ListModels`** — all three methods construct the request URL as `z.BaseURL[region]` without checking whether the key exists. For an unrecognised region this returns `""`, producing a malformed URL like `"/chat/completions"` that the HTTP transport rejects with a confusing error. `Encode` and `Rerank` (already merged) correctly fall back to `"default"` and return a clear error. This PR applies the same pattern to the remaining three methods.

### Type of change

### DIFF
--- a/internal/entity/models/aliyun.go
+++ b/internal/entity/models/aliyun.go
@@ -71,14 +71,9 @@ func (z *AliyunModel) ChatWithMessages(modelName string, messages []Message, api
 		region = *apiConfig.Region
 	}
 
-	baseURL := z.BaseURL["default"]
-	if region != "default" {
-		if regional, ok := z.BaseURL[region]; ok && regional != "" {
-			baseURL = regional
-		}
-	}
-	if baseURL == "" {
-		return nil, fmt.Errorf("aliyun: no base URL configured for default region")
+	baseURL, ok := z.BaseURL[region]
+	if !ok || baseURL == "" {
+		return nil, fmt.Errorf("aliyun: no base URL configured for region %q", region)
 	}
 
 	url := fmt.Sprintf("%s/%s", strings.TrimSuffix(baseURL, "/"), z.URLSuffix.Chat)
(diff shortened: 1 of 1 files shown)


Joseff531:
Hi, @Magicbook1108 
I'm excited to contribute to ragflow as a contributor.
Would be it possible to review?
Hope any feedback.

Haruko386 (commented):
Hi @Joseff531, 

I tested this on the Alibaba Cloud DashScope service in the China region and found that the balance query endpoint does not appear to be publicly accessible or officially documented. The endpoint returns an empty response even with a valid API key that can successfully access model APIs.

It would be very helpful if you could provide:

+ Official documentation for querying account balance / billing information
+ Supported balance-related API endpoints
+ Required headers or permissions (if any)
+ Whether this API is intended for public use or internal use only

At the moment, there seems to be no clear documentation for retrieving DashScope account balance programmatically in the China region.

Joseff531:
Hi @Haruko386, thank you for testing this — that's really valuable feedback.

You're correct. After further investigation, DashScope's account balance and billing information is managed through the Alibaba Cloud financial platform, which uses a different authentication mechanism (AccessKeyId + HMAC-SHA256 signing) from the standard `Authorization: Bearer <api_key>` used for model API calls. 
There does not appear to be a documented, public balance endpoint accessible via a DashScope model API key alone.

This is different from providers like SiliconFlow, Moonshot, and DeepSeek, which intentionally expose balance through the same API key used for inference.

**Proposed resolution:**
- Remove the `Balance` implementation and the `"balance"` URL suffix from this PR, since  there is no confirmed publicly accessible endpoint.
- Keep the remaining changes: the `Name()` bug fix (`"siliconflow"` → `"aliyun"`)

If you or anyone on the team is aware of an officially documented DashScope balance endpoint that works with a model API key, I'm happy to revise the implementation accordingly. 
Otherwise, I'll update the PR to scope it down to the two confirmed changes.

Thanks again for the thorough review.

Haruko386 (commented):
@Joseff531 Thank you for your contribution. This change is very small, and while I appreciate the effort, it's not significant enough to merge on its own.

Joseff531:
> @Joseff531 Thank you for your contribution. This change is very small, and while I appreciate the effort, it's not significant enough to merge on its own.

Really appreciate your careful review and understanding.
I'm really happy to contribute to ragflow.
If you're comfortable, I will update the PR including real bug fix for current codebase.
Thanks.

Joseff531:
Hi, @JinHai-CN 
CI is green now. could you please review the again?

PR #14664, opened 2026-05-08
Go: implement Encode (embeddings) in Ollama driver

jack-stormentswe:
### What problem does this PR solve?

The Ollama Go driver shipped with a stub \`Encode\` method that returned \`no such method\`, even though Ollama is one of the most common local LLM runners and exposes an OpenAI-compatible embeddings endpoint at \`/v1/embeddings\`.

Ollama users routinely run local embedding models such as \`nomic-embed-text\`, \`mxbai-embed-large\`, or \`bge-m3\`.
Pulled with \`ollama pull <model>\` and served on the same \`/v1\` namespace as chat. The existing \`ListModels\` already
discovers them, but because \`Encode\` was a stub, a tenant who picked one of these models in the Go layer could not
actually run an embedding call.

### What this PR includes

- \`conf/models/ollama.json\`: add \`\"embedding\": \"embeddings\"\` under \`url_suffix\` so the
  driver can build the URL from config.
- \`internal/entity/models/ollama.go\`: replace the \`Encode\` stub with a real implementation. Adds a small local response
  type that matches the OpenAI-compatible shape.

No factory change. No interface change.

### How the driver works

- Validate the model name. The API key is optional for local Ollama, so the Authorization header is only set when both
  \`apiConfig\` and \`ApiKey\` are non-nil and non-empty, the same pattern the recently merged CheckConnection PR (#14614) uses.
- Resolve the region with a default fallback. Return a clear "missing base URL" error when the user has not configured
  the local access address yet.
- Use a per-call \`context.WithTimeout(30s)\` and \`http.NewRequestWithContext\`, the same pattern the merged
  Aliyun Encode (#14647) uses.
- Send \`{model, input: [texts]}\` in one request.
- Parse \`data[*].embedding\` and copy each slice into a \`[][]float64\` indexed by \`data[*].index\`, so the output
  order matches the input order.
- Handle both \`float64\` and \`float32\` element types.
- Empty input returns \`[][]float64{}\` with no HTTP call.
- Length mismatch between input and result, out-of-range index, and any missing slot all return clear errors instead
  of silent zero vectors.

### Type of change

- [x] New Feature (non-breaking change which adds functionality)

### How was this tested?

- \`go build ./internal/entity/models/...\` in a clean go 1.25 image returns exit 0.
- The full method set on \`OllamaModel\` still matches the \`ModelDriver\` interface.
- Pattern parity with the merged Aliyun Encode (#14647) and the existing SiliconFlow Encode.

Closes #14662

PR #14688, opened 2026-05-08
Go: implement Encode (embeddings) in vLLM driver

### LINKED ISSUE #14687: [Feature Request]: Implement Encode (embeddings) in the vLLM Go driver
### Self Checks

```Markdown
Yes. The vLLM Go driver in
internal/entity/models/vllm.go has a stub Encode method that returns "not implemented", even though vLLM is one of the most common production-grade self-hosted inference servers and
exposes an OpenAI-compatible embeddings endpoint at:

  POST http://<host>/v1/embeddings

vLLM is regularly used to serve embedding models such as:

- BAAI/bge-m3
- BAAI/bge-large-en-v1.5
- intfloat/multilingual-e5-large-instruct
- Qwen/Qwen3-Embedding-0.6B / 4B / 8B
- nvidia/NV-Embed-v2

These run on the same /v1 namespace as chat. The existing ListModels in the driver already discovers them.

But because Encode is a stub, a tenant who picks one of these self-hosted embedding models in the Go layer cannot actually run an embedding call. The driver returns an error every
time, so the local KB indexing path silently falls back or fails.

OpenAI Encode (#14630), Aliyun Encode (just merged), and Ollama Encode (#14664) all use the same OpenAI-compatible /embeddings shape. vLLM is the next obvious gap in that pattern.
```

Implement VllmModel.Encode in internal/entity/models/vllm.go so the user's self-hosted embedding models work end to end through the Go layer.

The driver should:

- POST to ${BaseURL}/${URLSuffix.Embedding} (resolves to http://<host>/v1/embeddings with the user's configured base URL) using the existing httpClient.
- Send Authorization: Bearer <api_key> only if a non-empty key is configured. Self-hosted vLLM commonly does not use an API key, so the header must be optional. This matches the recently merged CheckConnection PR for vllm and the in-flight Ollama Encode (#14664).
- Accept a model name and a list of input texts, send them in a single request as the "input" array, the same OpenAI-compatible shape SiliconFlow, ZhipuAI, OpenAI, Aliyun, and Ollama all use.
- Return [][]float64 with one vector per input in input order.
- Use a per-call context.WithTimeout so a slow embedding call cannot block the worker forever.

The conf/models/vllm.json file needs a small update so the driver can read the embedding URL suffix from config:

  "url_suffix": {
    "chat": "chat/completions",
    "models": "models",
    "embedding": "embeddings"
  }

No new public API. No interface change. No factory change. The existing ListModels already enumerates the embedding models served by the user's vLLM instance, so no model entries need to be added to the config.

_No response_

```Markdown
No new docs are needed. vLLM is already in the model provider picker, and the existing ListModels surfaces whatever embedding models the user has loaded in their vLLM instance.

Use case: a team self-hosts BAAI/bge-m3 or Qwen3-Embedding-8B on vLLM for their internal knowledge base, points RAGFlow at their vLLM endpoint, picks the embedding model in the picker, and indexing works end to end without forcing a switch to a
cloud provider.
```

### Additional information

Reference templates that already live in the repo:

- Aliyun Encode (recently merged):
  internal/entity/models/aliyun.go, the Encode method.
- Ollama Encode (in flight, #14664):
  internal/entity/models/ollama.go, the Encode method.
- SiliconFlow Encode:
  internal/entity/models/siliconflow.go, the Encode method.

All three call the same OpenAI-compatible /embeddings endpoint shape, which is exactly what vLLM exposes.

Scope is intentionally limited to two files (internal/entity/models/vllm.go and conf/models/vllm.json), so it does not conflict with any in-flight provider work.

pandadev66:

The vLLM Go driver shipped with a stub \`Encode\` method that returned \`not implemented\`, even though vLLM is one of the most common production-grade self-hosted inference servers and exposes an OpenAI-compatible embeddings endpoint at \`/v1/embeddings\`.

Users who self-host \`BAAI/bge-m3\`, \`Qwen3-Embedding-*\`, \`NV-Embed-v2\`, or similar models on vLLM could not run an embedding call through the Go layer. The existing \`ListModels\` already discovers the loaded models, but the embedding path failed because \`Encode\` was a stub.

### What this PR includes

- \`conf/models/vllm.json\`: add   \`\"embedding\": \"embeddings\"\` under \`url_suffix\` so the driver can build the URL from config.
- \`internal/entity/models/vllm.go\`: replace the \`Encode\` stub with a real implementation. Adds a small local response
  type that matches the OpenAI-compatible shape.

No factory change. No interface change.

### How the driver works

- Validate the model name. The API key is optional for self-hosted vLLM, so the Authorization header is only set when both \`apiConfig\` and \`ApiKey\` are non-nil and non-empty, the same pattern the recently merged CheckConnection PR (#14614) uses.
- Resolve the region with a default fallback. Return a clear "missing base URL" error when the user has not configured
  the local access address yet.
- Use a per-call \`context.WithTimeout(30s)\` and \`http.NewRequestWithContext\`, the same pattern the merged
  Aliyun Encode (#14647) and in-flight Ollama Encode (#14664) use.
- Send \`{model, input: [texts]}\` in one request.
- Parse \`data[*].embedding\` and copy each slice into a \`[][]float64\` indexed by \`data[*].index\`, so the output
  order matches the input order.
- Handle both \`float64\` and \`float32\` element types.
- Empty input returns \`[][]float64{}\` with no HTTP call.
- Length mismatch between input and result, out-of-range index, and any missing slot all return clear errors instead
  of silent zero vectors.

### Type of change

### How was this tested?

- \`go build ./internal/entity/models/...\` in a clean go 1.25 image returns exit 0.
- The full method set on \`VllmModel\` still matches the \`ModelDriver\` interface.
- Pattern parity with the merged Aliyun Encode (#14647), the in-flight Ollama Encode (#14664), and the existing
  SiliconFlow Encode.

Closes #14687

### DIFF
--- a/conf/models/vllm.json
+++ b/conf/models/vllm.json
@@ -2,7 +2,8 @@
   "name": "vllm",
   "url_suffix": {
     "chat": "chat/completions",
-    "models": "models"
+    "models": "models",
+    "embedding": "embeddings"
   },
   "class": "local"
 }
--- a/internal/entity/models/vllm.go
+++ b/internal/entity/models/vllm.go
@@ -414,6 +414,9 @@ func (z *VllmModel) Encode(modelName *string, texts []string, apiConfig *APIConf
 		"model": *modelName,
 		"input": texts,
 	}
+	if embeddingConfig != nil && embeddingConfig.Dimension > 0 {
+		reqBody["dimensions"] = embeddingConfig.Dimension
+	}
 
 	jsonData, err := json.Marshal(reqBody)
 	if err != nil {
--- a/internal/entity/models/vllm.go
+++ b/internal/entity/models/vllm.go
@@ -19,6 +19,7 @@ package models
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


pandadev66:
Hi @Magicbook1108, gentle ping for review. Same OpenAI-compatible /embeddings shape that the merged Aliyun Encode (#14647) and the in-flight Ollama Encode (#14664) already use, scope is two files (one config, one driver), and it closes #14687. Thanks!

JinHai-CN:
@pandadev66   Encode interface is updated, would you please update your implementation accordingly? please refer to https://github.com/infiniflow/ragflow/pull/14735

pandadev66:
Hi @JinHai-CN, thanks for the heads up. Updated in 30609a591 to honor embeddingConfig.Dimension following the pattern in #14735.

JinHai-CN:
@pandadev66 Refer to the CLI usage in https://github.com/infiniflow/ragflow/issues/14736,  would you please provide the CLI embedding test result as below?
```
RAGFlow(user)> embed text 'what is rag' 'who are you' with 'embedding-3@test@zhipu-ai' dimension 16;
+-----------+-------+
| dimension | index |
+-----------+-------+
| 16        | 0     |
| 16        | 1     |
+-----------+-------+
```

pandadev66:
> @pandadev66 Refer to the CLI usage in #14736, would you please provide the CLI embedding test result as below?
> 
> ```
> RAGFlow(user)> embed text 'what is rag' 'who are you' with 'embedding-3@test@zhipu-ai' dimension 16;
> +-----------+-------+
> | dimension | index |
> +-----------+-------+
> | 16        | 0     |
> | 16        | 1     |
> +-----------+-------+
> ```

Hi @JinHai-CN 
Sharing my test output
```
RAGFlow(user)> embed text 'what is rag' 'who are you' with '<intfloat/e5-small-v2>@test@vllm' dimension 16;
+-----------+-------+
| dimension | index |
+-----------+-------+
| 16        | 0     |
| 16        | 1     |
+-----------+-------+
```

PR #14704, opened 2026-05-08
Fix: IMAGE2TEXT→CHAT fallback with model_type normalization in tenant_model_service

vincentlambert:
## Summary

- When a model is registered as `chat` in `tenant_llm` but has the `IMAGE2TEXT` tag in `llm_factories.json`, requesting it as `image2text` (e.g. PDF parser) fails with `Tenant Model with name <model> and type image2text not found`.
- After resolution via the new fallback, the returned `config_dict["model_type"]` was still `"chat"`, causing `tenant_llm_service.model_instance()` to instantiate `ChatModel` instead of `CvModel` — breaking `describe_with_prompt` at ingestion time.

RAGFlow already has a `CHAT→IMAGE2TEXT` fallback: when a chat model is not found, it retries with `image2text`. The symmetric fallback (`IMAGE2TEXT→CHAT`) was missing.

This matters for multimodal models declared as `model_type: "chat"` with an `IMAGE2TEXT` tag in `llm_factories.json` (e.g. models added after tenant creation, or providers where a single model serves both purposes). The frontend PDF parser selector correctly surfaces these models via the `IMAGE2TEXT` tag, but the backend fails to resolve them at runtime.

## Type of change

## Changes

**`api/db/joint_services/tenant_model_service.py`**

1. Add `IMAGE2TEXT→CHAT` fallback in `get_model_config_by_type_and_name`: when an `image2text` model is not found in `tenant_llm`, retry with `chat` — but only if the `llm` table confirms `IMAGE2TEXT` capability via the `tags` field. This mirrors the philosophy of the existing `CHAT→IMAGE2TEXT` fallback: substitution is only allowed when the model has declared the required capability.

2. Normalize `config_dict["model_type"]` to `image2text` after the fallback, so the caller (`model_instance`) correctly routes to `CvModel` instead of `ChatModel`.

3. Extend the type validation guard to allow `(requested=image2text, found=chat)` alongside the existing `(requested=chat, found=image2text)` exception.

## Test plan

- [ ] Add a model with `model_type=chat` and `tags` containing `IMAGE2TEXT` to a tenant
- [ ] Select it as PDF parser in a knowledge base
- [ ] Verify ingestion succeeds without `image2text not found` or `describe_with_prompt` errors
- [ ] Verify the same model still works correctly in chat context

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/api/db/joint_services/tenant_model_service.py
+++ b/api/db/joint_services/tenant_model_service.py
@@ -71,6 +71,19 @@ def get_model_config_by_type_and_name(tenant_id: str, model_type: str, model_nam
             if not model_config:
                 raise LookupError(f"Tenant Model with name {model_name} and type {model_type_val} not found")
             config_dict = model_config.to_dict()
+        elif model_type_val == LLMType.IMAGE2TEXT.value:
+            model_config = TenantLLMService.get_api_key(tenant_id, pure_model_name, LLMType.IMAGE2TEXT.value)
+            if not model_config:
+                # Fall back to a chat model only if it has declared IMAGE2TEXT capability (tag check via llm table)
+                chat_config = TenantLLMService.get_api_key(tenant_id, pure_model_name, LLMType.CHAT.value)
+                if chat_config:
+                    llm_entry = LLMService.query(llm_name=chat_config.to_dict()["llm_name"])
+                    if llm_entry and "IMAGE2TEXT" in (llm_entry[0].tags or ""):
+                        model_config = chat_config
+            if not model_config:
+                raise LookupError(f"Tenant Model with name {model_name} and type {model_type_val} not found")
+            config_dict = model_config.to_dict()
+            config_dict["model_type"] = LLMType.IMAGE2TEXT.value
         else:
             model_config = TenantLLMService.get_api_key(tenant_id, pure_model_name, model_type_val)
             if not model_config:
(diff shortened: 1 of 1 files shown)


vincentlambert:
Hi @yingfeng @6ba3i — could you take a look at this PR when you have a moment?

I've just pushed a second commit addressing both CodeRabbit review comments:

1. **Removed `to_dict()` call** — replaced `chat_config.to_dict()["llm_name"]` with direct attribute access (`chat_config.llm_name` / `chat_config.llm_factory`), which also allows passing the factory id directly to `LLMService.query`.

2. **Factory-scoped query + exact token match** — `LLMService.query` now receives `fid=chat_config.llm_factory` to avoid false positives from identically-named models in different factories. The tag check is now token-based (`"IMAGE2TEXT" in tags.split(",")`) instead of substring matching.

3. **Debug logging** added at each decision point (IMAGE2TEXT not found → chat found/not found → tags → promotion), to ease diagnosability in production.

This fix is a companion to #14705 (BedrockCV) — together they make chat-registered multimodal models (Bedrock, Gemini, etc.) fully usable as PDF parsers without re-registering them under `image2text`.

Lynn-Inf (approved):
LGTM

PR #14707, opened 2026-05-08
Fix: Bedrock api_key overridden by existing-key fallback in add_llm

vincentlambert:
## Summary

- Adding a Bedrock model from the frontend fails with `Fail to access model(Bedrock/<model>).Expecting value: line 1 column 1 (char 0)`.
- The assembled Bedrock JSON credentials are silently replaced by `"x"` before the connection test, causing `json.loads("x")` to raise a `JSONDecodeError`.

Commit `050113482` introduced a fallback in `add_llm()` that reuses the existing DB key when `req.get("api_key") is None`:

```python
if req.get("api_key") is None:
    api_key = existing_api_key if existing_api_key is not None else "x"
```

For Bedrock, credentials are sent as separate fields (`auth_mode`, `bedrock_ak`, `bedrock_sk`, `bedrock_region`, `aws_role_arn`) — the frontend does not send an `api_key` field. The function correctly assembles the JSON key:

```python
api_key = apikey_json(["auth_mode", "bedrock_ak", "bedrock_sk", "bedrock_region", "aws_role_arn"])
```

But since `req.get("api_key")` is `None`, the override immediately replaces `api_key` with `"x"` (or a stale DB value). `LiteLLMBase` then calls `json.loads("x")` for Bedrock auth → `JSONDecodeError`.

## Type of change

## Changes

**`api/apps/llm_app.py`**

Write the assembled key into `req["api_key"]` so the `None` check evaluates to `False` and the override is skipped — consistent with how `Tencent Cloud` is already handled.

```python
# Before
api_key = apikey_json(["auth_mode", "bedrock_ak", "bedrock_sk", "bedrock_region", "aws_role_arn"])

# After
req["api_key"] = apikey_json(["auth_mode", "bedrock_ak", "bedrock_sk", "bedrock_region", "aws_role_arn"])
api_key = req["api_key"]
```

## Test plan

- [ ] Configure a Bedrock provider in Model Providers with valid AWS credentials
- [ ] Add a Bedrock chat model — verify no `Expecting value` error
- [ ] Update the same model — verify the existing key is reused correctly when credentials fields are left empty

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


vincentlambert:
Hi @yingfeng — this PR fixes a regression introduced by your commit `050113482` (*Fix: support tool call config #14616*).

That commit added a fallback in `add_llm()` to reuse an existing DB key when `req.get("api_key") is None`. This is correct for most providers, but breaks Bedrock: the frontend sends credentials as separate fields (`bedrock_ak`, `bedrock_sk`, etc.) and never sets `api_key` — so the newly assembled JSON key is silently replaced by `"x"`, which then causes `json.loads("x")` → `JSONDecodeError` during the connection test.

The fix is a one-liner, consistent with how `Tencent Cloud` is already handled in the same function.

wangq8 (approved):

wangq8:
@Lynn-Inf please consider handle this logic in the new refined code.

Lynn-Inf (approved):
LGTM

PR #14717, opened 2026-05-08
Fix(Go): make OpenRouter Encode fail loudly on malformed responses

Joseff531:

The OpenRouter `Encode` method silently swallowed malformed responses. If a `data[]` item from the API was missing a field (`index`, `embedding`, or unexpected shape), the loop did `continue` instead of returning an error — leaving `nil` entries in the result slice. Callers got back partial results with no indication anything went wrong, which then crashes downstream consumers when they try to use a `nil` vector. There were three concrete gaps:

- No count-mismatch check between `data` length and input texts (only checked for empty)
- No duplicate-index detection (a duplicate would silently overwrite)
- Parse failures on individual items returned partial slices instead of erroring

This PR replaces `map[string]interface{}` parsing with a typed `openrouterEmbeddingResponse` struct and applies the same 3-layer validation used in the other drivers (count mismatch → out-of-range index → duplicate index), so any malformed response produces a clear error instead of corrupted data.

### Type of change

### DIFF
--- a/internal/entity/models/openrouter.go
+++ b/internal/entity/models/openrouter.go
@@ -351,9 +351,19 @@ func (o *OpenRouterModel) ChatStreamlyWithSender(modelName string, messages []Me
 	return scanner.Err()
 }
 
+type openrouterEmbeddingResponse struct {
+	Data []struct {
+		Index     int       `json:"index"`
+		Embedding []float64 `json:"embedding"`
+	} `json:"data"`
+}
+
 func (o *OpenRouterModel) Encode(modelName *string, texts []string, apiConfig *APIConfig, embeddingConfig *EmbeddingConfig) ([][]float64, error) {
 	if len(texts) == 0 {
 		return [][]float64{}, nil
+	}
+	if modelName == nil || *modelName == "" {
+		return nil, fmt.Errorf("model name is required")
 	}
 
 	var region = "default"
@@ -366,6 +376,10 @@ func (o *OpenRouterModel) Encode(modelName *string, texts []string, apiConfig *A
 	reqBody := map[string]interface{}{
 		"model": *modelName,
 		"input": texts,
+	}
+
+	if embeddingConfig != nil && embeddingConfig.Dimension > 0 {
+		reqBody["dimensions"] = embeddingConfig.Dimension
 	}
 
 	jsonData, err := json.Marshal(reqBody)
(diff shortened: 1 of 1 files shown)


internal/entity/models/minimax.go
```diff
@@ -344,9 +344,99 @@ func (z *MinimaxModel) ChatStreamlyWithSender(modelName string, messages []Messa
...
-// Encode encodes a list of texts into embeddings
+type minimaxEmbedRequest struct {
+	Model string   `json:"model"`
+	Texts []string `json:"texts"`
+	Type  string   `json:"type"`
+}
+
+type minimaxEmbedResponse struct {
+	Vectors [][]float64 `json:"vectors"`
+}
```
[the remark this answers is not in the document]
Joseff531: Fixed — added BaseResp to the response struct and check status_code != 0 before accepting vectors.

internal/entity/models/minimax.go
```diff
@@ -344,9 +344,99 @@ func (z *MinimaxModel) ChatStreamlyWithSender(modelName string, messages []Messa
...
+	req, err := http.NewRequest("POST", endpoint, bytes.NewBuffer(jsonData))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", apiKey))
+
+	resp, err := z.httpClient.Do(req)
+	if err != nil {
```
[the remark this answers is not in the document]
Joseff531: Not needed — NewMinimaxModel already sets Timeout: 120 * time.Second on the shared httpClient, which applies to every request including Encode. Adding a per-call context timeout would be redundant with the existing client-level timeout.

Joseff531:
Hi, @Haruko386 
I'd be happy if you could review the PR and share any feedback, Thanks

conf/models/minimax.json
```diff
@@ -99,6 +100,13 @@
         "default_value": true,
         "clear_thinking": true
       }
+    },
+    {
+      "name": "embo-01",
+      "max_tokens": 4096,
+      "model_types": [
+        "embedding"
+      ]
```
Haruko386: What's this? where did you find it?

Haruko386 (changes requested):
Hi @Joseff531, please let me know where you find `embo-01`
[MiniMax Model List](https://platform.minimax.io/docs/release-notes/models)

conf/models/minimax.json
Joseff531: You're right, I was confused with that and updated the PR with real bug fix.

Official OpenRouter docs:

API reference: https://openrouter.ai/docs/api/reference/embeddings
Endpoint detail: https://openrouter.ai/docs/api/api-reference/embeddings/create-embeddings
Verified facts that back the fix:

✅ Endpoint is POST https://openrouter.ai/api/v1/embeddings — matches embedding: "embeddings" in openrouter.json
✅ Response shape is OpenAI-compatible:
{"data": [{"index": N, "embedding": [...], "object": "embedding"}], "model": "...", "object": "list", "usage": {...}}
✅ "OpenAI-compatible embeddings API" — per their official docs

This confirms the typed struct (Index int, Embedding []float64) maps the documented fields, and the count-mismatch / duplicate-index / out-of-range checks are appropriate guards against an upstream API or proxy returning a malformed data[].

Joseff531: You're right, I was confused with that and updated the PR with real bug fix.

Official OpenRouter docs:

API reference: https://openrouter.ai/docs/api/reference/embeddings
Endpoint detail: https://openrouter.ai/docs/api/api-reference/embeddings/create-embeddings
Verified facts that back the fix:

✅ Endpoint is POST https://openrouter.ai/api/v1/embeddings — matches embedding: "embeddings" in openrouter.json
✅ Response shape is OpenAI-compatible:
{"data": [{"index": N, "embedding": [...], "object": "embedding"}], "model": "...", "object": "list", "usage": {...}}
✅ "OpenAI-compatible embeddings API" — per their official docs

This confirms the typed struct (Index int, Embedding []float64) maps the documented fields, and the count-mismatch / duplicate-index / out-of-range checks are appropriate guards against an upstream API or proxy returning a malformed data[].

Haruko386 (approved):
@Joseff531 LGTM

JinHai-CN:
@Joseff531 Encode interface is updated, would you please update your implementation accordingly? please refer to https://github.com/infiniflow/ragflow/pull/14735

Joseff531:
> @Joseff531 Encode interface is updated, would you please update your implementation accordingly? please refer to #14735

Thanks for the heads-up. Rebased this branch onto current main and updated the OpenRouter Encode to honor the new EmbeddingConfig.Dimension field — when set, it's now passed as the `dimensions` parameter to OpenRouter's embeddings API (officially documented at https://openrouter.ai/docs/api/reference/embeddings).

PR #14719, opened 2026-05-09
Perf(Go): batch SiliconFlow Encode requests with 32-item chunking

Joseff531:

The SiliconFlow `Encode` method sent one HTTP request per text, which is wasteful and slow when indexing many documents (e.g., 100 docs = 100 round-trips).

SiliconFlow's `/v1/embeddings` is OpenAI-compatible and accepts an array of strings in `input` (officially documented at https://docs.siliconflow.cn/en/api-reference/embeddings/create-embeddings, with a documented max array size of 32). This PR batches the requests up to that limit, reducing 100 docs to ~4 round-trips, and replaces `map[string]interface{}` parsing with a typed struct using the same 3-layer validation (count mismatch, out-of-range index, duplicate index) used in the other drivers.

### Type of change

- [x] Performance Improvement

### DIFF
--- a/internal/entity/models/siliconflow.go
+++ b/internal/entity/models/siliconflow.go
@@ -19,6 +19,7 @@ package models
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -368,10 +369,23 @@ func (z *SiliconflowModel) ChatStreamlyWithSender(modelName string, messages []M
 	return scanner.Err()
 }
 
-// Encode encodes a list of texts into embeddings
+type siliconflowEmbeddingResponse struct {
+	Data []struct {
+		Index     int       `json:"index"`
+		Embedding []float64 `json:"embedding"`
+	} `json:"data"`
+}
+
+// siliconflowMaxBatchSize is the per-request input limit documented at
+// https://docs.siliconflow.cn/en/api-reference/embeddings/create-embeddings.
+const siliconflowMaxBatchSize = 32
+
 func (s *SiliconflowModel) Encode(modelName *string, texts []string, apiConfig *APIConfig, embeddingConfig *EmbeddingConfig) ([][]float64, error) {
 	if len(texts) == 0 {
 		return [][]float64{}, nil
+	}
+	if modelName == nil || *modelName == "" {
+		return nil, fmt.Errorf("model name is required")
 	}
 
 	var region = "default"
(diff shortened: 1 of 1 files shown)


conf/models/moonshot.json
```diff
@@ -79,6 +80,13 @@
         "chat",
         "vision"
       ]
+    },
+    {
+      "name": "moonshot-v3-embedding",
+      "max_tokens": 128000,
+      "model_types": [
+        "embedding"
+      ]
```
Haruko386: Did `moonshot-v3-embedding` even exist ?
Tell me where you find it

Haruko386 (changes requested):
Hi @Joseff531, please stop submitting meaningless PRs

conf/models/moonshot.json
Joseff531: I was confused with the implementation.
Will update the PR with meaningful changes right away.
Thanks for your review.

Joseff531:
Hi, @Haruko386 
I've updated the PR.
The branch now batches SiliconFlow's `Encode` instead, which previously sent one HTTP request per text. Per the [official docs](https://docs.siliconflow.cn/en/api-reference/embeddings/create-embeddings), `/v1/embeddings` accepts arrays up to 32 items, so the new code chunks at that limit (100 texts: 100 → 4 round-trips) and adds typed parsing with the standard 3-layer validation. Title and description updated.
Hope your review. 🙏

Haruko386 (changes requested):
Thanks for the contribution. The overall implementation is much better aligned with the SiliconFlow embeddings API compared to the previous version, especially the batched input handling and the max batch size splitting logic.

A few issues still need to be addressed before this can be considered fully robust:

1. Missing embedding index validation
   Currently the code only checks:

```go
if len(result.Data) != len(batch)
```

However, this is insufficient to guarantee that all indices are present exactly once. For example, malformed responses like:

```json
[
  {"index":0},
  {"index":2}
]
```

could still bypass the length check.

Please add a final validation step such as:

```go
for i, ok := range seen {
    if !ok {
        return fmt.Errorf("missing embedding index %d", i)
    }
}
```

2. Empty embedding vectors are not validated
   The current implementation accepts:

```json
{
  "embedding": []
}
```

silently. It would be safer to explicitly reject empty embeddings:

```go
if len(item.Embedding) == 0 {
    return fmt.Errorf("empty embedding at index %d", item.Index)
}
```

3. No request timeout handling
   The implementation currently performs blocking HTTP requests without context timeout protection. For embedding APIs, especially batched requests, this may lead to hanging goroutines when the provider stalls.

Please consider adding `context.WithTimeout(...)` similarly to the OpenAI embedding implementation.

Other than these points, the batch processing logic, index-based assignment, duplicate detection, and API schema handling all look correct.

JinHai-CN:
@Joseff531 Encode interface is updated, would you please update your implementation accordingly? please refer to #14735

Joseff531:
Hi, @Haruko386 @JinHai-CN 
Thanks for the careful review. All four points addressed in the latest force-push:

1. **Missing-index check** — added a final `for i, ok := range seen` scan after assignment, exactly as suggested.
2. **Empty-embedding guard** — added `if len(item.Embedding) == 0` rejection inside the assignment loop.
3. **Per-call context timeout** — switched to `http.NewRequestWithContext` with `context.WithTimeout(ctx, nonStreamCallTimeout)`, mirroring the OpenAI driver.
4. **EmbeddingConfig.Dimension** — rebased onto current main and now forwarded as the `dimensions` parameter (OpenAI-compatible field that SiliconFlow honors) when non-zero.

Build is clean.

Joseff531:
Thanks for your review, @JinHai-CN

PR #14749, opened 2026-05-10
fix(api): close private dataset doc auth bypass

### LINKED ISSUE #14659: [Bug]: TDocument APIs leak private datasets to tenant members
### Self Checks

bb10b83e6114cc271c3563194464b19297dc5c25

### RAGFlow image version

infiniflow/ragflow:v0.25.1

```Markdown
- Browser: N/A
- OS: Any
```

### Actual behavior

Datasets configured with `permission = me` are intended to be private to the owner, but multiple document APIs authorize access using `KnowledgebaseService.accessible()`, which only checks whether the caller belongs to the dataset's tenant and does not enforce the dataset permission flag.

This creates an authorization bypass where a non-owner tenant member can access or modify a private dataset if they know its `dataset_id`.

Any tenant member with the `dataset_id` can pass authorization checks on several document endpoints because access is granted based on tenant membership rather than the dataset's permission setting.

This can expose private document inventories and may also allow unauthorized modification actions such as document deletion, metadata updates, or parse control operations.

### Expected behavior

Only the dataset owner should be able to access a dataset marked with `permission = me`. Tenant membership alone should not grant access to private datasets.

### Steps to reproduce

```Markdown
1. Create two users in the same tenant or workspace: User A and User B.
2. As User A, create a dataset with permission set to `me` so it is private.
3. Confirm User B cannot see this dataset through normal dataset listing.
4. As User B, obtain the private dataset's `dataset_id`.
5. Call a document API that checks `KnowledgebaseService.accessible()`, for example `GET /api/v1/datasets/<dataset_id>/documents`.
6. Observe that the request succeeds even though the dataset is marked private.
```

### Additional information

_No response_

Parvezkhan0:

Fixes #14659

This PR fixes an authorization bypass in the token-authenticated document SDK APIs for private datasets (permission = me).
The affected routes validated dataset access with KnowledgebaseService.accessible(), but when a login token was used through the SDK token flow, the check received the caller's tenant/workspace id instead of the authenticated user id. That let another member of the same tenant pass access checks for a private dataset if they knew the dataset_id, exposing document listings and allowing document operations that should have remained owner-only.

  This change forwards the authenticated user id through the token-auth flow and uses it for dataset authorization in the affected SDK document endpoints, so private datasets remain accessible only to their owner while team datasets continue to work as before.

### Type of change


KevinHuSh:
Appreciations!

Does this mean that all the functions using `@token_required` needs to delcare one more parameter `authenticated_user_id`?

PR #14765, opened 2026-05-11
fix: authenticate /documents/images endpoint and restrict to chunk im…

### LINKED ISSUE #14763: [Bug]: Unauthenticated cross-tenant data exfiltration via GET /api/v1/documents/images/<image_id> (missing @login_required, confused-deputy MinIO fetch)
### Self Checks

e46989832

### RAGFlow image version

nightly (main @ e46989832)

```Markdown
- Component: Python backend (Quart) — api/apps/restful_apis/document_api.py
- Affected endpoint: GET /api/v1/documents/images/<image_id>
- Auth decorator on route: none (no @login_required, no tenant scoping)
- Introduced by: PR #14344 (migration of /v1/document/thumbnails → /api/v1/thumbnails and /v1/document/image → /api/v1/documents/images)
```

### Actual behavior

While reading through `api/apps/restful_apis/document_api.py` I noticed the `/documents/images/<image_id>` route at line 1641 has **no decorator at all** — no `@login_required`, no tenant scoping. Every other route in this file is gated by `@login_required` (lines 58, 121, 246, 366, 612, 966, 1116, 1229, 1357, 1428, 1540, 1690, 1740, 1860, 1893). There is also no global `before_request` hook adding auth anywhere under `api/` (verified with grep).

The handler splits the user-supplied `image_id` on `-` and passes the two halves straight to the storage backend:

```python
arr = image_id.split("-")
if len(arr) != 2:
    return get_data_error_result(message="Image not found.")
bkt, nm = image_id.split("-")
data = await thread_pool_exec(settings.STORAGE_IMPL.get, bkt, nm)
response = await make_response(data)
response.headers.set("Content-Type", "image/JPEG")
return response
```

So anyone on the internet can pull arbitrary `(bucket, key)` pairs out of MinIO/S3/OSS with `curl`, no session required. Two things make this strictly worse than ordinary thumbnail leakage:

1. The thumbnail bucket and the raw-document bucket are the **same bucket**. `api/db/services/file_service.py:487` writes the original PDF/Word/etc. blob with `STORAGE_IMPL.put(kb.id, doc.location, blob, kb.tenant_id)`, and line 519 writes the thumbnail with `STORAGE_IMPL.put(kb.id, thumbnail_location, img)`. The `image/JPEG` Content-Type is hardcoded, so a request for `<kb_id>-report.pdf` will return the **raw PDF body** mislabeled as JPEG.
2. The application itself embeds the vulnerable URL into authenticated list responses at `document_api.py:734` and `document_api.py:1221` as `f"/api/v1/documents/images/{kb_id}-{thumbnail_filename}"`. So `(kb_id, filename)` pairs flow into HTML, browser history, server access logs, and Referer headers — and any one of those can be replayed by an unauthenticated attacker.

The `len(image_id.split("-")) != 2` guard is effectively a no-op for the realistic attack surface: 32-hex KB ids are hyphen-free, and corporate filenames like `report.pdf`, `invoice.docx`, `Q3.xlsx` are hyphen-free.

This is the same threat model PR #14625 / commit d51fb8857 just patched for `GET /v1/document/get/<doc_id>` — except this endpoint is strictly worse because it doesn't even require authentication, and isn't restricted to image-typed files.

For background: PR #13822 ("fix: add authentication and authorization to /thumbnails endpoint (CWE-287)") flagged the equivalent unauthenticated `/image/<image_id>` route in the pre-migration `api/apps/document_app.py` and explicitly called out: > *"The `/image/<image_id>` endpoint at line 818 in the same file also has `# @login_required` commented out. This is a separate but related vulnerability with a different attack surface … It should be addressed in a follow-up."* That follow-up was never filed, and PR #14344 then migrated the route into `restful_apis/document_api.py` carrying the missing decorator forward.

### Expected behavior

`GET /api/v1/documents/images/<image_id>` should:

1. Require authentication.
2. Only return content the caller actually has access to via their tenant/team membership.
3. Only serve thumbnails — not arbitrary objects whose keys happen to live in the same bucket.

Unauthenticated or cross-tenant requests should return an authorization error, not the file bytes.

### Steps to reproduce

```Markdown
1. Set up a tenant with at least one uploaded document that has a thumbnail
   generated (any PDF/Word upload does this automatically). Note the
   knowledge base id (KB_ID) and the document's stored filename (DOC_LOCATION).
   These are visible in any authenticated list-documents response under the
   "thumbnail" field, which already contains the leak-able URL.

2. From any host on the internet, with no cookies / no API key:

   curl -X GET "https://<host>/api/v1/documents/images/<KB_ID>-thumbnail_<DOC_ID>.png" \
     --output leaked.png

   This returns the thumbnail image — no authentication performed.

3. Same primitive, raw-document exfiltration:

   curl -X GET "https://<host>/api/v1/documents/images/<KB_ID>-<DOC_LOCATION>" \
     --output leaked.bin

   For example, if DOC_LOCATION is "report.pdf", the response body is the
   raw PDF, mislabeled as Content-Type: image/JPEG. The endpoint does not
   restrict what object key may be fetched — only that bucket and key parse
   from a single hyphen-split.
```

### Additional information

**Suggested fix** — mirror the approach taken in PR #14625 (cross-tenant document download fix) and PR #14725 (file-ancestry IDOR fix):

1. Add `@login_required` to the route.
2. Change the route to identify the document, not the storage key. Take a `doc_id`, look up the document, and gate on `DocumentService.accessible(doc_id, current_user.id)` (the helper joins `Document` → `Knowledgebase` → `UserTenant`, so it also handles shared/team KBs):

   ```python
   @manager.route("/documents/<doc_id>/thumbnail", methods=["GET"])  # noqa: F821
   @login_required
   async def get_document_thumbnail(doc_id):
           return get_data_error_result(message="No authorization.")
       e, doc = DocumentService.get_by_id(doc_id)
       if not e:
           return get_data_error_result(message="Document not found.")
       thumbnail_key = f"thumbnail_{doc.id}.png"
       data = await thread_pool_exec(settings.STORAGE_IMPL.get, doc.kb_id, thumbnail_key)
       response = await make_response(data)
       response.headers.set("Content-Type", "image/png")
       return response
   ```

   Resolving the storage key from the document record (instead of from user input) also closes the confused-deputy primitive — the endpoint can no longer be coerced into serving arbitrary document bodies even within an authorized tenant.

3. Update the two URL builders that emit the legacy path to use the new one:
   - `api/apps/restful_apis/document_api.py:734`
   - `api/apps/restful_apis/document_api.py:1221`

4. Update the two frontend `` constructors so images don't break after the route change:
   - `web/src/components/image/index.tsx:16`
   - `web/src/components/next-message-item/reference-image-list.tsx:82`

5. Re-enable the test at `test/testcases/test_web_api/test_document_app/test_document_metadata.py:454` (currently `@pytest.mark.skip(reason="Moved to /api/v1/documents/images/<image_id>")`) and add coverage for (a) anonymous request denied and (b) cross-tenant request denied, modeled on the tests added in PR #14625.

hunnyboy1217:

Closes #14763.

`GET /api/v1/documents/images/<image_id>` had no `@login_required` and split the user-supplied path on `-` into a `(bucket, key)` pair passed straight to `STORAGE_IMPL.get`. Because thumbnails and raw documents share the same bucket (`kb.id`), an unauthenticated caller could fetch arbitrary objects — including raw PDFs/Word docs — by reconstructing keys that the application itself embedded into authenticated list responses (`document_api.py:736`, `:1198`).
Same threat model as #14625, but worse: no auth required and not restricted to image-typed files.

This PR:

1. **`/documents/images/<image_id>`** — adds `@login_required`, validates KB access via `KnowledgebaseService.accessible`, and restricts the storage key to the chunk-image shape (`^[0-9a-f]{16}$`, the xxhash64 digest produced in `rag/svr/task_executor.py`). Thumbnail filenames, raw doc filenames, or any other key shape are rejected before the access check, closing the confused-deputy primitive even for authorized callers. The endpoint now serves only chunk reference images — its only legitimate frontend use.

2. **New `/documents/<doc_id>/thumbnail`** — authenticated, gated by `DocumentService.accessible`, derives the storage key
   (`thumbnail_{doc.id}.png`) server-side from the document record. Returns `Content-Type: image/png` (the previous handler hardcoded `image/JPEG` even for PNG thumbnails).

3. **URL builders** at `document_api.py:736` and `:1198` now emit `/api/v1/documents/{doc_id}/thumbnail` instead of the leakable `kb_id-key` URL. Side benefit: the `/thumbnails` JSON response no longer exposes `kb_id`.

4. **Tests** — re-enables the previously skipped unit test (renamed to match the new handler name) and adds coverage for: cross-tenant denial on both endpoints, confused-deputy key shapes (`thumbnail_*.png`, `report.pdf`, non-hex, 17-hex) rejected before access check, server-derived storage key on the thumbnail endpoint, and exception paths.

**Frontend not changed.** The two `` constructors flagged in the issue (`web/src/components/image/index.tsx`,
`web/src/components/next-message-item/reference-image-list.tsx`) consume `chunk.image_id` (kb_id-xxhash16), not document thumbnail URLs — they keep working against the now-authenticated `/documents/images/<image_id>` endpoint.
Document thumbnails in the frontend already flow through `useFetchDocumentThumbnailsByIds` → `/thumbnails`, which now returns the new authenticated URL automatically.

### Type of change


hunnyboy1217:
Hi, @wangq8 ,

api/apps/restful_apis/document_api.py
```diff
@@ -1165,23 +1165,33 @@ async def update_metadata_config(tenant_id, dataset_id, document_id):
 
 
 @manager.route("/thumbnails", methods=["GET"])  # noqa: F821
+@login_required
```
KevinHuSh, line 1194: In this way, images loadding will all fail for embeded <iframe> dialogs. (The dialogs can be embeded into iframe)

hunnyboy1217:
Hi, @KevinHuSh ,
Thanks for your review.
I just fixed current changes based on your direction.
Could you please review it again?

hunnyboy1217:
Hi, @KevinHuSh ,

api/apps/restful_apis/document_api.py
KevinHuSh: I mean this might be removed.

hunnyboy1217:
Hi, @yingfeng ,
All CI tests are successfully passed.

hunnyboy1217:
Hi, @yingfeng ,

hunnyboy1217:
Hi, @JinHai-CN ,
Please review my PR. Thank you.

hunnyboy1217:
HI, @yingfeng ,
Could you please review this?

JinHai-CN:
@hunnyboy1217 would you please resolve the conflicts?

JinHai-CN:

Issue #14750, opened 2026-05-10
RFC: Replace Peewee with SQLModel


yuzhichang:
# Proposal: Replace Peewee with SQLModel

> **Status**: Draft
> **Date**: 2026/05/10
> **Author**: RAGFlow Team

---

## 1. Background and Goals

RAGFlow's backend currently uses [Peewee](https://docs.peewee-orm.com/) as its ORM. The core database abstraction lives in [api/db/db_models.py](api/db/db_models.py) and [api/db/db_utils.py](api/db/db_utils.py). Over time, we have identified structural pain points that cannot be resolved without migrating to a more capable ORM.

We propose migrating directly to [SQLModel](https://sqlmodel.tiangolo.com/).

**Why SQLModel over raw SQLAlchemy**: SQLModel is built by the FastAPI author on top of SQLAlchemy and Pydantic. It allows table models and API data models to **share field definitions through inheritance**, reducing boilerplate without conflating persistence and API contracts. For a project like RAGFlow with many Pydantic schemas, shared base classes cut down on field duplication while keeping DB-only fields out of API contracts. Under the hood, SQLModel exposes the full SQLAlchemy Engine/Session API, so all session lifecycle, connection pooling, and retry mechanisms are identical to raw SQLAlchemy.

---

### 2.1 Nested Connection Context Collisions

**Problem**: The codebase extensively uses `@DB.connection_context()` decorators to manage database connection contexts.

Current usage locations (partial):

| File | Lines | Purpose |
|---|---|---|
| `api/db/services/common_service.py` | 89, 109, 139, ... | Batch insert operations |
| `api/db/services/user_canvas_version.py` | 41, 61, 78 | User canvas version operations |
| `api/db/services/langfuse_service.py` | 35, 45, 55 | Langfuse integration |
| `api/db/services/mcp_server_service.py` | 35, 81, 90 | MCP server management |
| `api/db/services/search_service.py` | 42, 56, 82, 119 | Search service |

When business logic forms nested call chains:

```
parent_function()        # @DB.connection_context()
  └─ child_function()   # @DB.connection_context()
       └─ grandchild_function()  # @DB.connection_context()
```

A child-layer `connection_context()` may reopen a connection already held by the parent, or return it to the pool prematurely due to exception mishandling, ultimately triggering `InterfaceError: (0, '')` or connection leaks.

[...]

PR #14776, opened 2026-05-11
feat: introduce SQLAlchemy 2.0 engine + session infra and Alembic scaffold (refs #14750)

wdeveloper16:
## Summary

- Introduces the SQLAlchemy 2.0 engine, session factory, and Alembic scaffold  as the foundation for migrating off Peewee (part of #14750)
- No existing Peewee code is changed; the new engine runs alongside Peewee  during the transition period

## Design rationale

**SQLAlchemy 2.0, not SQLModel.** SQLModel adds a thin layer over SQLAlchemy but its support for composite primary keys, custom field types, and complex queries (aliases, `Case`, `fn.COUNT`) is limited — and RAGFlow's models use all of these. SQLModel's own docs recommend separate create/read schemas for models with sensitive fields, which most RAGFlow models have. Raw SQLAlchemy 2.0 gives full control without the extra abstraction.

**Synchronous `scoped_session`, not async.** The Peewee layer is entirely synchronous today. Switching to `AsyncSession` would require rewriting every service-layer signature to `async def` — orthogonal to the ORM migration itself. `scoped_session` eliminates nested-connection conflicts (215 `@DB.connection_context()` sites) without touching call signatures.

**OceanBase gap.** SQLAlchemy has no first-party OceanBase dialect. This PR falls back to `mysql+pymysql` with a warning if `sqlalchemy-oceanbase` is not installed. That dialect must be validated before any service is migrated.

**Phased plan** (this PR = Phase 1):
- Phase 1 (this PR): infra only — engine, session, Alembic scaffold
- Phase 2: generate Alembic baseline migration from live schema; make Alembic  sole owner of DDL
- Phase 3: incrementally convert service layer, smallest services first
- Phase 4: delete `RetryingPooledMySQLDatabase`, Peewee models, `playhouse`

## Changes

- `api/db/sqlalchemy_engine.py` — `create_engine` with `pool_pre_ping`,  `scoped_session`, `get_db_session()` context manager, `close_db_session()`,  `check_db_health()`, and OceanBase graceful fallback
- `api/apps/__init__.py` — wires `close_db_session()` into the Quart teardown  hook alongside the existing `close_connection()`
- `alembic.ini` / `alembic/env.py` / `alembic/script.py.mako` — standard  Alembic setup pointing at the same engine (no hardcoded credentials)
- `pyproject.toml` — adds `sqlalchemy>=2.0,<3.0` and `alembic>=1.13,<2.0`

## Test plan

- [ ] `python -c "from api.db.sqlalchemy_engine import check_db_health; print(check_db_health())"` returns `True` with services running
- [ ] `alembic current` runs without error
- [ ] Existing Peewee-backed endpoints are unaffected


wdeveloper16:
Hi, @yuzhichang 
I'd be happy if you could you review the PR again.

api/db/alembic/env.py
KevinHuSh, line 1: What about adding this folder to deeper folder instead of at the first layer of folders?

wdeveloper16: Good point. 👍 
 `alembic/` at the repo root clutters the top level. I'll move it under `api/db/alembic/` so the migration scripts live next to `db_models.py,` `sqlalchemy_engine.py,` and the rest of the DB code. Only `alembic.ini's` script_location needs to be updated to point there. 
Will push the change shortly.

alembic.ini
KevinHuSh, line 1: Good!~ Thanks
Could you put this file into `api/db/alembic` too?

dataCenter430: I think this was already addressed by author.

yuzhichang:
We have nearly done porting the RAGFlow project from python to go (with the exact the same API with python). The python part is dying out.

PR #14793, opened 2026-05-11
fix: correct tag_kwd field display in POST /api/v1/retrieval endpoint

ogidog:
### Description
Fixed an issue where the `tag_kwd` field was always empty in the `POST /api/v1/retrieval` response, even when tags were present in Elasticsearch.

While the tags were correctly returned by the chunk details endpoint:
`GET /api/v1/datasets/{dataset_id}/documents/{document_id}/chunks?id={chunk_id}`

They were missing from the retrieval results. This PR ensures that `tag_kwd` is correctly populated in the `POST /api/v1/retrieval` endpoint.

### Changes
- Fixed data mapping to include `tag_kwd` in the retrieval API response.


KevinHuSh:
Appreciations!

About `test.py`, we do not put tests there.

PR #14815, opened 2026-05-12
Go: implement provider: StepFun

tmimmanuel:
### What problem does this PR solve?

Add a Go driver for StepFun (阶跃星辰), one of the unchecked providers on the umbrella tracking issue #14736.

Until this PR, a tenant who configured `stepfun` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver. Chat, list models, and check connection all returned `"not implemented"` instead of reaching the StepFun API.

The Python side has had StepFun registered in `rag/llm/__init__.py` as a `SupportedLiteLLMProvider` with base URL `https://api.stepfun.com/v1`, plus `StepFunCV` for vision and `StepFunSeq2txt` for ASR, but no Go path. StepFun's chat API is OpenAI-compatible, so the implementation pattern is the same as the merged Moonshot driver (#14433) and OpenAI driver (#14605).

### What this PR includes

- New file `internal/entity/models/stepfun.go` with a `StepFunModel` that implements the `ModelDriver` interface.
- `factory.go`: route the `"stepfun"` provider name to `NewStepFunModel`.
- New `conf/models/stepfun.json` with the public StepFun chat models (step-2-16k, step-1 family in 8k/32k/128k/256k context lengths, step-1-flash, and the step-1v / step-1o vision models) and `url_suffix` entries for `chat` and `models`.

### How the driver works

- StepFun exposes the OpenAI-compatible API at `https://api.stepfun.com/v1`.
- `ChatWithMessages` and `ChatStreamlyWithSender` post to `/chat/completions` in the same shape as the merged moonshot, openrouter, and openai drivers.
- `ListModels` and `CheckConnection` call `/models` to list available ids and confirm the API key works.
- `Embed` is left as `"not implemented"`. StepFun has not advertised a public embeddings endpoint in the API reference linked from the umbrella issue (`https://platform.stepfun.com/docs/en/api-reference/chat/chat-completion-create` is the chat endpoint), so any real implementation belongs in a separate follow-up only after the endpoint is verified.
- `Rerank` and `Balance` return `"no such method"` because StepFun does not expose either.

### Type of change

- [x] New Feature (non-breaking change which adds functionality)

### How was this tested?

- `go build ./internal/entity/models/...` returns exit 0 with no errors on go 1.25 (the `go.mod` minimum).
[...]

PR #14817, opened 2026-05-12
fix(upstage): extract reasoning delta from streaming responses

tmimmanuel:

`UpstageModel.ChatStreamlyWithSender` (in the driver merged via #14819) only extracted `delta.content` from each SSE event. For the `solar-pro3` reasoning family (and any future Upstage model that follows the same wire shape), the chain-of-thought is streamed in a **separate `delta.reasoning` field**, and the driver was silently dropping all of it.

The non-streaming path already extracts `message.reasoning` into `ChatResponse.ReasonContent` (added earlier in this PR's history), so the same model produced **inconsistent behavior** between streaming and non-streaming: a tenant calling `solar-pro3` with `reasoning_effort: high` would see the reasoning trace if they used `ChatWithMessages` but not if they used `ChatStreamlyWithSender`.

### Live evidence

Probed against `api.upstage.ai/v1/chat/completions` with `solar-pro3` + `reasoning_effort: high` + `stream: true` (8000-token budget so the reasoning has room to finish):

```
$ curl -sN -H "Authorization: Bearer <key>" -H "Content-Type: application/json" \
       -X POST https://api.upstage.ai/v1/chat/completions \
       -d '{"model":"solar-pro3","messages":[{"role":"user","content":"Compute 15% of 80."}],
            "max_tokens":8000,"stream":true,"reasoning_effort":"high"}'

# across 168 SSE events:
#   delta keys seen: [content reasoning role]
#   delta.content total len:   121 chars   (the visible answer)
#   delta.reasoning total len: 159 chars   (the chain-of-thought) <- driver dropped this
```

A representative event showing both fields side by side:

```json
data: {"choices":[{"index":0,"delta":{"reasoning":"15% = 0.15."}}]}
data: {"choices":[{"index":0,"delta":{"content":"15% of 80 is "}}]}
```

The 159 chars of reasoning were arriving on the wire and being thrown away. `solar-pro2` was also probed (625 events); it does **not** emit `delta.reasoning` — its reasoning is inlined into `delta.content` — so this change is a no-op for it and for `solar-mini`.

### What this PR includes

- `internal/entity/models/upstage.go`: in the SSE scanner loop, extract `delta.reasoning` before `delta.content` and forward each non-empty chunk via the sender's second arg (the existing `reasonContent` channel the non-stream path already populates).

  The ordering contract is documented inline: reasoning chunks within a single SSE event are emitted before content chunks, so a UI that pipes both sees the chain-of-thought start before the answer for that token, matching the wire order Upstage emits.

- `internal/entity/models/upstage_test.go`: three new tests pinning the new behavior:
  - `TestUpstageStreamExtractsReasoningDelta` — reasoning + content forwarded to the right sender args; one-of invariant per call
  - `TestUpstageStreamReasoningChunksArriveBeforeContent` — ordering pinned within a single SSE event that carries both fields
  - `TestUpstageStreamWithoutReasoningStillWorks` — regression net: non-reasoning models (`solar-mini`, `solar-pro2`) continue to work; the reason callback never fires

No interface change. No factory change. No config change.

### How was this tested?

```
$ go test -vet=off -run TestUpstage -count=1 -v ./internal/entity/models/...
... (existing tests 1..9 still pass) ...
=== RUN   TestUpstageStreamExtractsReasoningDelta
--- PASS: TestUpstageStreamExtractsReasoningDelta (0.01s)
=== RUN   TestUpstageStreamReasoningChunksArriveBeforeContent
--- PASS: TestUpstageStreamReasoningChunksArriveBeforeContent (0.01s)
=== RUN   TestUpstageStreamWithoutReasoningStillWorks
--- PASS: TestUpstageStreamWithoutReasoningStillWorks (0.00s)
PASS
ok      ragflow/internal/entity/models  0.034s
```

12/12 Upstage tests pass on go 1.25. `go build ./internal/entity/models/...` exits 0.

**Live integration test** (smoke test not committed) — the patched driver was run directly against `api.upstage.ai/v1` with the same prompt that produced the curl evidence above:

```
=== RUN   TestUpstageStreamReasoningLiveSmoke
    [OK] visible content: 50 chunks, 84 chars
    [OK] reasoning:       39 chunks, 90 chars
    content head 200:   "\\(15\\% = \\frac{15}{100}=0.15\\).\n\n\\[\n0.15 \\times 80 = 12.\n\\]\n\n**15 % of 80 is 12.**"
    reasoning head 200: "We need to compute 15% of 80. That's 0.15 * 80 = 12. So answer is 12. Provide explanation."

UPSTAGE STREAM REASONING SMOKE PASSED
--- PASS: TestUpstageStreamReasoningLiveSmoke (1.97s)
```

Before this fix, the same call would have produced **0 reasoning chunks**. The 90 chars of reasoning that the patched driver now surfaces are the chain-of-thought solar-pro3 emits when reasoning_effort is high.

### Type of change

### DIFF
--- a/internal/entity/models/upstage.go
+++ b/internal/entity/models/upstage.go
@@ -356,6 +356,19 @@ func (u *UpstageModel) ChatStreamlyWithSender(modelName string, messages []Messa
 			continue
 		}
 
+		// Reasoning chunks first, content second. Upstage's solar-pro3
+		// stream interleaves both fields within the same SSE event when
+		// reasoning_effort is medium or high; emit reasoning before the
+		// visible answer so callers that pipe both into a UI see the
+		// chain-of-thought start before the answer, matching the wire
+		// ordering. solar-pro2 inlines reasoning into delta.content and
+		// never sets delta.reasoning, so this block is a no-op for it.
+		if r, ok := delta["reasoning"].(string); ok && r != "" {
+			if err := sender(nil, &r); err != nil {
+				return err
+			}
+		}
+
 		content, ok := delta["content"].(string)
 		if ok && content != "" {
 			if err := sender(&content, nil); err != nil {
--- a/internal/entity/models/upstage_test.go
+++ b/internal/entity/models/upstage_test.go
@@ -269,3 +269,155 @@ func TestUpstageEmbedHappyPathReordersByIndex(t *testing.T) {
 		}
 	}
 }
+
+// ---------- streaming reasoning delta extraction ----------
+
+// TestUpstageStreamExtractsReasoningDelta verifies that the SSE parser
+// forwards `delta.reasoning` chunks via the sender's second arg (the
+// reasonContent channel) and `delta.content` chunks via the first arg.
+// Fixture matches the shape captured live from solar-pro3 with
+// reasoning_effort=high — both fields appear, sometimes in the same
+// chunk and sometimes separately.
+func TestUpstageStreamExtractsReasoningDelta(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.WriteHeader(http.StatusOK)
+		_, _ = io.WriteString(w,
+			`data: {"choices":[{"index":0,"delta":{"role":"assistant"}}]}`+"\n"+
+				`data: {"choices":[{"index":0,"delta":{"reasoning":"We need "}}]}`+"\n"+
+				`data: {"choices":[{"index":0,"delta":{"reasoning":"to compute. "}}]}`+"\n"+
+				`data: {"choices":[{"index":0,"delta":{"reasoning":"15% = 0.15."}}]}`+"\n"+
+				`data: {"choices":[{"index":0,"delta":{"content":"15% of 80 "}}]}`+"\n"+
+				`data: {"choices":[{"index":0,"delta":{"content":"is 12."},"finish_reason":"stop"}]}`+"\n"+
+				`data: [DONE]`+"\n",
+		)
+	}))
+	defer srv.Close()
+
+	u := newUpstageForTest(srv.URL)
+	apiKey := "test-key"
+	var contentChunks, reasoningChunks []string
+	err := u.ChatStreamlyWithSender("solar-pro3",
+		[]Message{{Role: "user", Content: "What is 15% of 80?"}},
+		&APIConfig{ApiKey: &apiKey}, nil,
+		func(content *string, reason *string) error {
+			// At most one of (content, reason) is set per call: callers
+			// need this contract to route to the right UI channel.
+			if content != nil && reason != nil {
+				t.Errorf("sender called with both content and reason non-nil")
+			}
+			if content != nil && *content != "" && *content != "[DONE]" {
+				contentChunks = append(contentChunks, *content)
+			}
+			if reason != nil && *reason != "" {
+				reasoningChunks = append(reasoningChunks, *reason)
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		t.Fatalf("stream: %v", err)
+	}
+	wantReason := "We need to compute. 15% = 0.15."
+	wantContent := "15% of 80 is 12."
+	if got := strings.Join(reasoningChunks, ""); got != wantReason {
+		t.Errorf("joined reasoning=%q want %q", got, wantReason)
+	}
+	if got := strings.Join(contentChunks, ""); got != wantContent {
+		t.Errorf("joined content=%q want %q", got, wantContent)
+	}
+	if len(reasoningChunks) != 3 {
+		t.Errorf("expected 3 reasoning chunks, got %d", len(reasoningChunks))
+	}
+}
+
+// TestUpstageStreamReasoningChunksArriveBeforeContent verifies the
+// ordering contract inside a single SSE event when both fields are
+// present: reasoning is forwarded first so a UI consuming both can
+// render the chain-of-thought before the answer for that token.
+func TestUpstageStreamReasoningChunksArriveBeforeContent(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.WriteHeader(http.StatusOK)
+		_, _ = io.WriteString(w,
+			// One SSE event carries BOTH reasoning and content in the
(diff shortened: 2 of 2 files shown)


tmimmanuel:
## CodeRabbit review applied (commit 2ba2303c3)

Two findings on the driver:

- **`ChatWithMessages` (line 103) and `ListModels` (line 378): empty-region fallback.** Match the guard already used in `ChatStreamlyWithSender` — only override `"default"` when the region string is non-empty. Otherwise an empty `*Region` would miss the `baseURLForRegion` lookup despite a valid `"default"` entry.
- **`ChatStreamlyWithSender` (line 213): nil-sender guard.** Added at the top of the function alongside the existing checks. Without it, a nil callback would panic when invoked in the SSE scan loop.

Same fix pattern as Mistral PR #14805 (commit ee08b645a) and StepFun PR #14815 (commit 2a9cb2d8f).

`go build ./internal/entity/models/...` → exit 0 on go 1.25.

JinHai-CN:
Duplicated with #14819 . I will close it. Please feel free to re-open it, if i am wrong.

internal/entity/models/upstage.go
```diff
@@ -356,6 +356,19 @@ func (u *UpstageModel) ChatStreamlyWithSender(modelName string, messages []Messa
...
+		// reasoning_effort is medium or high; emit reasoning before the
+		// visible answer so callers that pipe both into a UI see the
+		// chain-of-thought start before the answer, matching the wire
+		// ordering. solar-pro2 inlines reasoning into delta.content and
+		// never sets delta.reasoning, so this block is a no-op for it.
+		if r, ok := delta["reasoning"].(string); ok && r != "" {
+			if err := sender(nil, &r); err != nil {
+				return err
+			}
+		}
```
Haruko386, line 370: no `reasoning`

Haruko386 (changes requested):
I checked the current Upstage chat response docs at [Upstage Chat API Response](https://console.upstage.ai/api/chat?utm_source=chatgpt.com#response) again, and as far as I can tell there is still no documented `reasoning` / `reasoning_content` field in the response payload.

Let me know if I am wrong

internal/entity/models/upstage.go
tmimmanuel: image
image

There is reasoning as optional. What shall I do?

tmimmanuel:
@Haruko386 thanks for pushing back on this — I went and ran live tests against `api.upstage.ai` to settle the question. The `reasoning` field **does** exist in responses; it's just not in the OpenAPI schema section of the doc page you linked. The relevant prose is in the same page's `reasoning_effort` description (search for "When reasoning is ON"):

> *"When reasoning is ON (\`high\` or \`medium\`) for \`solar-pro3\`, the response includes a \`reasoning\` field with the model's thought process."*
> — https://console.upstage.ai/api/chat (under the `reasoning_effort` param section)

## Live evidence (api.upstage.ai)

### 1. solar-pro3, non-stream, `reasoning_effort: "high"` → `reasoning` is present

Request:
```bash
curl -X POST https://api.upstage.ai/v1/chat/completions \
  -H "Authorization: Bearer <KEY>" -H "Content-Type: application/json" \
  -d '{"model":"solar-pro3","messages":[{"role":"user","content":"What is 15% of 80? answer in one short sentence."}],"reasoning_effort":"high","max_tokens":2048}'
```

Response (pretty-printed):
```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "15% of 80 equals 12.",
      "reasoning": "We need to answer: \"What is 15% of 80?\" Provide answer in one short sentence. So answer: \"15% of 80 is 12.\" That's a short sentence."
    },
    "finish_reason": "stop"
  }]
}
```

The `message.reasoning` field is 132 chars long, present, populated.

### 2. solar-pro3, non-stream, `reasoning_effort: "minimal"` → `reasoning` is absent (control)

```json
{"choices":[{"message":{"role":"assistant","content":"4"},"finish_reason":"stop",...}]}
```

`message.reasoning` is correctly missing — proves the field only appears when reasoning is on.

### 3. solar-pro3, **streaming**, `reasoning_effort: "high"` → `delta.reasoning` chunks before `delta.content`

The exact case this PR exists to handle. First 12 deltas verbatim from the wire:

```
data: {"choices":[{"delta":{"role":"assistant"}, ...}]}
data: {"choices":[{"delta":{"reasoning":"The"}, ...}]}
data: {"choices":[{"delta":{"reasoning":" user"}, ...}]}
data: {"choices":[{"delta":{"reasoning":" asks"}, ...}]}
data: {"choices":[{"delta":{"reasoning":" \""}, ...}]}
data: {"choices":[{"delta":{"reasoning":"2"}, ...}]}
data: {"choices":[{"delta":{"reasoning":"+"}, ...}]}
data: {"choices":[{"delta":{"reasoning":"2"}, ...}]}
data: {"choices":[{"delta":{"reasoning":"?"}, ...}]}
data: {"choices":[{"delta":{"reasoning":" single"}, ...}]}
data: {"choices":[{"delta":{"reasoning":" number"}, ...}]}
data: {"choices":[{"delta":{"reasoning":"\"."}, ...}]}
...
data: {"choices":[{"delta":{"content":"4"}, "finish_reason":null}]}
data: {"choices":[{"delta":{}, "finish_reason":"stop"}]}
data: [DONE]
```

122 SSE lines total. The set of distinct delta keys observed: `['content', 'reasoning', 'role']`. There is **no** `reasoning_content` — it's named `reasoning` (singular, no `_content` suffix), distinct from LongCat / DeepSeek / OpenAI-o1.

Without this PR, every one of those `delta.reasoning` chunks gets dropped on the floor in `ChatStreamlyWithSender`, and a tenant chatting with solar-pro3 sees only the final answer with no visible thinking — even though they paid the reasoning tokens.

### 4. solar-pro2, non-stream, `reasoning_effort: "high"` → reasoning is internal, **not in response**

```json
{"choices":[{"message":{"role":"assistant","content":"\n\n15% of 80 is 12."},"finish_reason":"stop",...}]}
```

`message.reasoning` absent. So the docs are accurate when they scope the response field to **solar-pro3** specifically — solar-pro2 with `high` does reason internally but doesn't surface a separate field. The driver code already handles both cases (no-op when the field is absent).

## What stays

The current PR diff is doing the right thing — it adds `delta.reasoning` extraction to mirror the existing `message.reasoning` extraction on the non-stream path. Both shapes are now confirmed by the live API.

Happy to add a docstring pointer to the prose quote above if it helps future readers find the citation in the docs.

Haruko386 (approved):
@tmimmanuel LGTM
Thank your for your contribution

tmimmanuel:
@JinHai-CN
ready to merge

Issue #14736, opened 2026-05-09
[Feature] Implement model providers of RAGFlow API server in Go


Haruko386:
Hello, community!

We are planning to rewrite the RAGFlow API server in Go, with a particular focus on the model provider layer. We welcome contributions from the community to implement support for different model providers. You can refer to the following PR for implementation patterns and guidance:

👉 #14790 

If you're interested in contributing to the Go rewrite, please feel free to pick a model provider or a method and submit a PR. Let's build a more efficient and maintainable API server together.

Thanks!

Model provider list:
- [x] OpenAI #14605 #14630 #15104 
- [x] Anthropic #14940 
- [x] Google Gemini #14395 #14682 #14660 #14661
- [x] xAI #14550
- [x] DeepSeek #14632
- [x] ZhipuAI #14234 #14242 #14573 #14735 #14608 #14886 #15143 
- [x] Moonshot #14262 #14433
- [x] Minimax #14296 #14478 #14911 
- [x] Aliyun qwen #14379 #14676 #14647
- [x] VolcEngine #14409 #14460 #14700 #14721
- [x] HuggingFace  #14722
- [x] Nivida #14623 #14636 #14721
- [x] OpenRouter #14652 #14657 #14717 #14926 #15067 
- [x] SiliconFlow #14336 #14643 #14719 #14944
- [x] Gitee #14336 #14656 #14698
- [x] Ollama #14580 #14614 #14664 #15213 
- [x] vllm #14532 #14614 #14688
- [x] LM-studio #14586 #14614 #14694
- [x] Baidu #14741 #14954
- [x] FishAudio #14926 
- [x] localai #14813 
- [x] baichuan #14832 
- [x] jina #14838 #14935 
- [x] mistral #14807 #14954 
- [x] stepfun #14815 #14944 
- [x] cohere #14790 #14954 
- [x] togetherAI #14957 #15017 
- [x] Upstage #14819 
- [x] novitaai #14850 #14895 #15014 
- [x] replicate #14958 #15073 
- [x] Voyage-ai #14811 
- [x] Deep-infra #14978 
- [x] LongCat #14809 
- [x] minerU #14990 #15051 
- [x] paddle ocr #14954 #15158 
- [x] [model-scope](https://modelscope.ai/docs/openapi) #15041 
- [x] Xinference #14938 #15032 #15096
- [x] azure-openai #15022 
- [x] Xunfei Spark #14978 
- [x] Cometapi #14930 #15001 
[...]

PR #14809, opened 2026-05-12
Go: implement provider: LongCat

tmimmanuel:
### What problem does this PR solve?

Add a Go driver for LongCat (Meituan, https://longcat.chat), one of the unchecked providers on the umbrella tracking issue #14736. LongCat exposes an OpenAI-compatible REST API at `https://api.longcat.chat/openai/v1` with three public chat models including `LongCat-Flash-Thinking`, a reasoning model that returns chain-of-thought in `reasoning_content` (OpenAI o-series shape).

Until this PR, a tenant who configured `longcat` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver.

### What this PR includes

- New `internal/entity/models/longcat.go` with a `LongCatModel` implementing the `ModelDriver` interface.
- New `conf/models/longcat.json` with the 3 public chat models (Flash-Chat, Flash-Lite, Flash-Thinking) and `url_suffix` for `chat` and `models`.
- `factory.go`: route `"longcat"` to `NewLongCatModel`.

Method coverage:

- `ChatWithMessages`: `POST /openai/v1/chat/completions`, non-streaming
- `ChatStreamlyWithSender`: SSE stream against the same endpoint
- `ListModels` / `CheckConnection`: `GET /openai/v1/models`
- **Reasoning extraction**: `message.reasoning_content` (non-stream) and `delta.reasoning_content` (stream) flow into `ChatResponse.ReasonContent` / the sender's second arg. Matches the OpenAI o-series convention also used by kimi-k2.6 and DeepSeek-R1.
- **`reasoning_effort` propagation**: `ChatConfig.Effort` → request body `reasoning_effort` (LongCat-Flash-Thinking honors it; non-reasoning models ignore it).
- `Embed` / `Rerank` / `Balance` / `TranscribeAudio` / `AudioSpeech` / `OCRFile` return `"no such method"` (LongCat does not expose any of these surfaces).

No interface change. No new dependencies.

### How was this tested?

**21 unit tests** in `internal/entity/models/longcat_test.go` — all pass:

```
$ go test -vet=off -run TestLongCat -count=1 -v ./internal/entity/models/...
=== RUN   TestLongCatName
--- PASS: TestLongCatName (0.00s)
=== RUN   TestLongCatChatHappyPath
--- PASS: TestLongCatChatHappyPath (0.00s)
=== RUN   TestLongCatChatExtractsReasoningContent
--- PASS: TestLongCatChatExtractsReasoningContent (0.00s)
[...]

PR #14850, opened 2026-05-12
Go: implement provider: Novita.ai

tmimmanuel:

Add a Go driver for Novita.ai (https://novita.ai), one of the unchecked providers on the umbrella tracking issue #14736. Novita exposes an OpenAI-compatible REST API at `https://api.novita.ai/v3/openai` and proxies a large catalog of third-party models (DeepSeek, Llama, Qwen3, Kimi, Gemma, Mistral, MiniMax, GLM, etc.) behind a single OpenAI-shaped surface — 102 models live at the time of writing.

Until this PR, a tenant who configured `novita` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver.

### What this PR includes

- New `internal/entity/models/novita.go` with a `NovitaModel` implementing the `ModelDriver` interface (~520 lines).
- New `conf/models/novita.json` with 7 representative chat models (DeepSeek-V4, Llama-3.3-70B, Qwen3-30B/235B reasoning, Kimi-K2, Gemma-3-27B, Mistral-Nemo).
- `factory.go`: route `"novita"` to `NewNovitaModel`.
- `internal/entity/models/novita_test.go`: 23 unit tests.

### Notable design point: `<think>...</think>` reasoning extraction

Novita-routed reasoning models like `qwen3-*` and `deepseek-r1-*` embed their chain-of-thought **inline inside content as `<think>...</think>` tags**, rather than in a separate `reasoning_content` field. Verified live by probing `api.novita.ai`:

```
content head 200: <think>
Okay, let's see. I need to find 15% of 80. Hmm, percentages can sometimes be tricky, but I think
content tail 100: h, that works.
Alternatively, 0.15 × 80. If I move the decimal two places to the left for </think>
```

Without handling, a tenant picking qwen3 via Novita would see raw `<think>` tags in their UI answer — different from every other reasoning provider in the Go layer.

The driver detects those tags and routes the inner text to `ChatResponse.ReasonContent` (non-stream) or the sender's second arg (stream), keeping the visible answer clean of tag clutter:

- **`splitNovitaThink`** — scans a complete content string. Used by the non-streaming path. Handles multiple `<think>` blocks, unclosed tags (the model got cut off mid-reasoning), pure-text content with no tags.
- **`novitaThinkExtractor`** — stateful streaming version. Buffers trailing bytes that might be the start of a tag (e.g. `<thi` held back when the next chunk completes `nk>`), then emits segments in routing order so callers can pipe them to a UI. Tested with byte-level chunk boundaries and tag-spanning scenarios.

### Method coverage

| Method | Behavior |
|---|---|
| `ChatWithMessages` | `POST /v3/openai/chat/completions`, `<think>` extraction on response |
| `ChatStreamlyWithSender` | SSE stream, stateful `<think>` extraction across deltas |
| `ListModels` / `CheckConnection` | `GET /v3/openai/models` (102 live) |
| `Embed` / `Rerank` / `Balance` / `TranscribeAudio` / `AudioSpeech` / `OCRFile` | `"no such method"` — Novita's OpenAI-compatible surface does not expose any |

No interface change. No new dependencies.

### How was this tested?

**23 unit tests** in `internal/entity/models/novita_test.go` — all pass:

```
$ go test -vet=off -run "TestNovita|TestSplitNovita" -count=1 ./internal/entity/models/...
ok      ragflow/internal/entity/models  0.020s
```

Coverage:

- `splitNovitaThink` (5 cases: pure text, single block, leading text, multiple blocks, unclosed tag)
- `novitaThinkExtractor` (6 cases: single-chunk, opening tag span, closing tag span, byte-level chunking, no tags, lone `<` not as tag start)
- `ChatWithMessages`: pure text, with `<think>` tags, missing API key, empty messages, HTTP error
- `ChatStreamlyWithSender`: tag-stripping with spanning deltas, pure content, sender-required, stream-true-required
- `ListModels` / `CheckConnection` (happy paths)
- All sentinel methods

`go build ./internal/entity/models/...` exits 0 on go 1.25.

**Live integration test** against `api.novita.ai/v3/openai`:

```
=== RUN   TestNovitaLiveSmoke
    [OK] Name() = "novita"
    [OK] CheckConnection
    [OK] ListModels: 102 models (showing first 6) [deepseek/deepseek-v4-pro deepseek/deepseek-v4-flash deepseek/deepseek-v3.2 xiaomimimo/mimo-v2.5-pro moonshotai/kimi-k2.6 zai-org/glm-5.1]
    [OK] Chat (llama-3.3) answer="ok" reason=""
    [OK] Chat (qwen3) answer len=0 head=""
                  ReasonContent len=1657 head="Okay, so I need to figure out what 15% of 80 is. Hmm, percentages can sometimes trip me up, but let ..."
    [OK] Stream content: 0 chunks, 0 chars; reasoning: 600 chunks, 1667 chars
    [OK] Embed/Rerank/Balance/TranscribeAudio/AudioSpeech/OCRFile all return "novita, no such method"

NOVITA LIVE SMOKE PASSED
--- PASS: TestNovitaLiveSmoke (26.18s)
```

What the live run proves on the wire:

- Auth (`Bearer <key>`) accepted by `api.novita.ai`.
- `/v3/openai/models` parser handles the real 102-model response.
- Non-stream chat against `meta-llama/llama-3.3-70b-instruct`: clean string answer, empty ReasonContent (non-reasoning model, pure-text path).
- Non-stream chat against `qwen/qwen3-30b-a3b-fp8`: 1657-char reasoning extracted from `<think>...</think>` and routed to `ChatResponse.ReasonContent`. Visible answer is 0 chars in this run because qwen3 spent its 600-token budget entirely on reasoning before reaching the answer phase — that's the model's behavior, not a driver bug. The important thing: **no `<think>` tags leaked into the visible Answer field**.
- Streaming against qwen3: 600 reasoning chunks (1667 chars) emitted via the sender's 2nd arg across SSE deltas; **no `<think>` tag fragments leaked into the content channel** despite tag boundaries crossing chunk boundaries on the wire.
- All 6 sentinel methods return the documented `"no such method"` strings.

### Type of change

Tracking: #14736

### DIFF
--- a/conf/models/novita.json
+++ b/conf/models/novita.json
@@ -1,0 +1,62 @@
+{
+  "name": "NovitaAI",
+  "url": {
+    "default": "https://api.novita.ai/v3/openai"
+  },
+  "url_suffix": {
+    "chat": "chat/completions",
+    "models": "models"
+  },
+  "class": "novita",
+  "models": [
+    {
+      "name": "deepseek/deepseek-v4-pro",
+      "max_tokens": 65536,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "meta-llama/llama-3.3-70b-instruct",
+      "max_tokens": 131072,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "qwen/qwen3-30b-a3b-fp8",
+      "max_tokens": 32768,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "qwen/qwen3-235b-a22b-fp8",
+      "max_tokens": 32768,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "moonshotai/kimi-k2-instruct",
+      "max_tokens": 131072,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "google/gemma-3-27b-it",
+      "max_tokens": 131072,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "mistralai/mistral-nemo",
+      "max_tokens": 131072,
+      "model_types": [
+        "chat"
+      ]
+    }
+  ]
+}
--- a/conf/models/novita.json
+++ b/conf/models/novita.json
@@ -1,11 +1,11 @@
 {
-  "name": "NovitaAI",
+  "name": "Novita",
   "url": {
-    "default": "https://api.novita.ai/v3/openai"
+    "default": "https://api.novita.ai"
   },
   "url_suffix": {
-    "chat": "chat/completions",
-    "models": "models"
+    "chat": "openai/v1/chat/completions",
+    "models": "openai/v1/models"
   },
   "class": "novita",
   "models": [
--- a/internal/entity/models/novita.go
+++ b/internal/entity/models/novita.go
@@ -1,0 +1,613 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+package models
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// NovitaModel implements ModelDriver for Novita.ai
+// (https://novita.ai/docs/api-reference/).
+//
+// Novita exposes an OpenAI-compatible REST API at
+// https://api.novita.ai/v3/openai (chat completions at
+// /chat/completions, list models at /models). It serves a large
+// catalog of third-party models (DeepSeek, Llama, Qwen3, Kimi,
+// Gemma, Mistral, etc.) behind a single OpenAI-shaped surface.
+//
+// The wire shape matches OpenAI standard with ONE notable
+// difference: reasoning models like qwen3-* embed their
+// chain-of-thought INLINE inside content as <think>...</think>
+// tags, rather than in a separate reasoning_content field. The
+// driver detects those tags and routes the inner text to
+// ChatResponse.ReasonContent (non-stream) or the sender's second
+// arg (stream), keeping the answer clean of tag clutter.
+type NovitaModel struct {
+	BaseURL    map[string]string
+	URLSuffix  URLSuffix
+	httpClient *http.Client
+}
+
+// NewNovitaModel creates a new Novita model instance.
+//
+// Same transport convention as other Go drivers in this package:
+// clone http.DefaultTransport, override the connection-pool fields,
+// no client-level Timeout so SSE streams are not capped.
+func NewNovitaModel(baseURL map[string]string, urlSuffix URLSuffix) *NovitaModel {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.MaxIdleConns = 100
+	transport.MaxIdleConnsPerHost = 10
+	transport.IdleConnTimeout = 90 * time.Second
+	transport.DisableCompression = false
+	transport.ResponseHeaderTimeout = 60 * time.Second
+
+	return &NovitaModel{
+		BaseURL:   baseURL,
+		URLSuffix: urlSuffix,
+		httpClient: &http.Client{
+			Transport: transport,
+		},
+	}
+}
+
+func (n *NovitaModel) NewInstance(baseURL map[string]string) ModelDriver {
+	return NewNovitaModel(baseURL, n.URLSuffix)
+}
+
+func (n *NovitaModel) Name() string {
+	return "novita"
+}
+
+func (n *NovitaModel) baseURLForRegion(region string) (string, error) {
+	base, ok := n.BaseURL[region]
+	if !ok || base == "" {
+		return "", fmt.Errorf("novita: no base URL configured for region %q", region)
+	}
+	return base, nil
+}
+
+const (
+	novitaThinkOpen  = "<think>"
+	novitaThinkClose = "</think>"
+)
+
+// splitNovitaThink walks a complete content string and returns the
+// visible portion + the concatenated chain-of-thought from inside
+// any <think>...</think> blocks. Multiple think blocks are
+// concatenated; tags themselves are stripped. Used by the
+// non-streaming path where the whole content is available at once.
+func splitNovitaThink(raw string) (visible, reasoning string) {
+	var v, r strings.Builder
+	inside := false
+	for {
+		var marker string
+		if inside {
+			marker = novitaThinkClose
+		} else {
+			marker = novitaThinkOpen
+		}
+		idx := strings.Index(raw, marker)
+		if idx < 0 {
+			if inside {
+				r.WriteString(raw)
+			} else {
+				v.WriteString(raw)
+			}
+			break
+		}
+		if inside {
+			r.WriteString(raw[:idx])
+		} else {
+			v.WriteString(raw[:idx])
+		}
+		raw = raw[idx+len(marker):]
+		inside = !inside
+	}
+	return v.String(), r.String()
+}
+
+// novitaThinkExtractor maintains state across streaming chunks so
+// that a <think>...</think> block spanning multiple SSE events still
+// gets split correctly between content and reasoning. The buffer
+// preserves up to (len(closingMarker)-1) trailing bytes of each
+// chunk in case the next chunk completes a partial tag.
+type novitaThinkExtractor struct {
+	buf    strings.Builder
+	inside bool
+}
+
+// novitaThinkSegment is one routing decision: emit `content` via the
+// sender's first arg, or emit `reasoning` via the sender's second arg.
+// Exactly one of the two fields is non-empty.
+type novitaThinkSegment struct {
+	content   string
+	reasoning string
+}
+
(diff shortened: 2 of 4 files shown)


internal/entity/models/novita.go
```diff
@@ -0,0 +1,613 @@
...
+	if apiConfig.Region != nil && *apiConfig.Region != "" {
+		region = *apiConfig.Region
+	}
+
+	baseURL, err := n.baseURLForRegion(region)
+	if err != nil {
+		return nil, err
+	}
+	url := fmt.Sprintf("%s/%s", baseURL, n.URLSuffix.Chat)
+
```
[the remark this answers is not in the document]
tmimmanuel: Fixed in commit 3faf24ed3.

Took the centralized path: `baseURLForRegion` now returns `strings.TrimSuffix(base, "/")` so all three endpoint builders (Chat at line 244, Stream at line 367, ListModels at line 524) emit clean paths with zero change to the call sites. Suffix is sourced from `conf/models/novita.json` which has no leading slash, so trimming the trailing slash on the base side is sufficient.

New table-driven test `TestNovitaBaseURLTrimsTrailingSlash` configures `baseURL` **with** a trailing slash on purpose and asserts each endpoint's handler sees the clean path:

```
=== RUN   TestNovitaBaseURLTrimsTrailingSlash
=== RUN   TestNovitaBaseURLTrimsTrailingSlash/Chat
=== RUN   TestNovitaBaseURLTrimsTrailingSlash/ListModels
=== RUN   TestNovitaBaseURLTrimsTrailingSlash/Stream
--- PASS: TestNovitaBaseURLTrimsTrailingSlash (0.00s)
    --- PASS: TestNovitaBaseURLTrimsTrailingSlash/Chat (0.00s)
    --- PASS: TestNovitaBaseURLTrimsTrailingSlash/ListModels (0.00s)
    --- PASS: TestNovitaBaseURLTrimsTrailingSlash/Stream (0.00s)
```

The handler asserts `r.URL.Path == "/chat/completions"` (not `//chat/...`) and `"/models"` (not `//models`). The shipped config has no trailing slash, so this only matters when a tenant overrides the URL per-instance — but that's exactly the case CodeRabbit flagged.

24/24 Novita tests pass; build clean.

conf/models/novita.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "NovitaAI",
+  "url": {
+    "default": "https://api.novita.ai/v3/openai"
```
Haruko386: wrong baseURL: shoule be`https://api.novita.ai`

conf/models/novita.json
```diff
@@ -0,0 +1,62 @@
...
+  },
+  "url_suffix": {
+    "chat": "chat/completions",
```
Haruko386: wrong url_suffix: should be `openai/v1/chat/completions`
[chat completions api](https://novita.ai/docs/api-reference/model-apis-llm-create-chat-completion)

conf/models/novita.json
```diff
@@ -0,0 +1,62 @@
...
+    "models": "models"
```
Haruko386: wrong url_suffix: should be `v1/models`
[models api](https://novita.ai/docs/api-reference/model-apis-llm-list-models)

conf/models/novita.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "NovitaAI",
```
Haruko386: remove `AI`

internal/entity/models/novita.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	url := fmt.Sprintf("%s/%s", baseURL, n.URLSuffix.Models)
+
+	ctx, cancel := context.WithTimeout(context.Background(), nonStreamCallTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
```
Haruko386, line 580: missing `req.Header.Set("Content-Type", "application/json")`

internal/entity/models/novita.go
```diff
@@ -0,0 +1,617 @@
...
+		}
+		if chatModelConfig.Temperature != nil {
+			reqBody["temperature"] = *chatModelConfig.Temperature
+		}
+		if chatModelConfig.TopP != nil {
+			reqBody["top_p"] = *chatModelConfig.TopP
+		}
+		if chatModelConfig.Stop != nil {
+			reqBody["stop"] = *chatModelConfig.Stop
+		}
```
Haruko386, line 276: missing explicit think mode control

Current implementation does not expose `enable_think`, make it always in thinking mode.
```plaintext
RAGFlow(user)> chat with 'qwen/qwen3-30b-a3b-fp8@test@novita' message 'who r u'
Thinking: 
Okay, the user asked, "who r u". Let me think about how to respond.

First, I need to acknowledge their question. They might be curious about my identity. I should introduce myself clearly.

I should mention that I'm Qwen, a large-scale language model developed by Alibaba Cloud. That's the main point. But maybe they want more details?

They used "r" instead of "are", so maybe they're texting or casual. I should keep the response friendly and approachable.

Do they need to know my purpose? Like, what I can do? Maybe include that I can answer questions, create text, etc. But not too verbose.

Also, check if there's any cultural context. Since they wrote in English, probably okay to respond in English. But maybe they're non-native, so keep it simple.

Avoid jargon. Make sure it's easy to understand. Maybe add an emoji to keep it friendly.

Wait, the user might be testing if I'm a real person. So I should clarify I'm an AI assistant. But also invite them to ask questions.

Don't forget to mention my capabilities. Like helping with various tasks. But keep it concise.

Double-check the response for clarity and friendliness. Make sure it's not too technical but informative.

Answer: 

Hi there! I'm Qwen, a large-scale language model developed by Alibaba Cloud. I'm here to help you with a wide range of tasks, from answering questions and creating text to coding and more. How can I assist you today? 😊
Time: 6.932954
```

Please add a dedicated think-mode switch and map it to:

internal/entity/models/novita.go
```diff
@@ -0,0 +1,617 @@
...
```
Haruko386, line 430: like [this reivew](https://github.com/infiniflow/ragflow/pull/14850/changes#r3231185017)

Haruko386 (changes requested):
Hi @tmimmanuel, please read the [docs](https://novita.ai/docs/guides/introduction) before your test or submit your code.

tmimmanuel:
@Haruko386 thanks — went back through https://novita.ai/docs/api-reference/model-apis-llm-create-chat-completion and https://novita.ai/docs/api-reference/model-apis-llm-list-models and pushed ee8259e. Same lesson applied here as in #14809: align to the docs first, then live-verify.

## What changed

### `conf/models/novita.json`

| Field | Before | After |
| --- | --- | --- |
| `name` | `NovitaAI` | `Novita` |
| `url.default` | `https://api.novita.ai/v3/openai` | `https://api.novita.ai` |
| `url_suffix.chat` | `chat/completions` | `openai/v1/chat/completions` |
| `url_suffix.models` | `models` | `openai/v1/models` |

**Note on `v1/models` vs `openai/v1/models`:** your inline comment said `v1/models`. Live-probing `https://api.novita.ai/v1/models` returned **HTTP 404**; the docs page you linked clearly lists the path as `/openai/v1/models`, and that one returns 200 with the full 102-model catalog. I went with the docs-and-live-verified `openai/v1/models`. If you actually meant a different host or prefix, please flag — happy to change.

### `internal/entity/models/novita.go`

1. **`ListModels` missing `Content-Type` header** — added.
2. **`enable_thinking` switch** — `ChatConfig.Thinking *bool` now maps to Novita's documented `enable_thinking` body field on both `ChatWithMessages` and `ChatStreamlyWithSender`. A tenant chatting with `qwen3-30b-a3b-fp8` can now flip thinking off at request time:

   ```
   $ curl … -d '{"model":"qwen/qwen3-30b-a3b-fp8","messages":[…],"enable_thinking":false}'
   {"choices":[{"message":{"role":"assistant","content":"Hello! How can I assist you today? 😊"}…}]}
   ```
   (no `<think>...</think>` clutter, just the answer)

3. **Two reasoning shapes, not one** — live audit caught this and the driver now handles both:

   | Shape | Models | How the driver routes it |
   | --- | --- | --- |
   | `<think>...</think>` inline inside `content` | qwen3-* (default) | existing tag splitter routes to `ReasonContent` |
   | Separate `reasoning_content` field on `message` / `delta` | deepseek-v3.1, glm-4.5, anything with `enable_thinking` on | new code path — merges into `ChatResponse.ReasonContent` and emits via the stream sender's second arg |

   Live test on `deepseek/deepseek-v3.1` with `enable_thinking=true` confirmed 66 `delta.reasoning_content` chunks arrive per stream. The previous code only read `delta.content`, so all of that reasoning was being silently dropped on the floor.

### Tests

New tests pinning the new behaviour:
- `TestNovitaChatExtractsReasoningContentField`
- `TestNovitaStreamExtractsDeltaReasoningContent`
- `TestNovitaChatPropagatesEnableThinking` (enabled / disabled)
- `TestNovitaChatOmitsEnableThinkingWhenUnset`
- `TestNovitaStreamPropagatesEnableThinking`

Existing streaming tests now go through a new `newNovitaSSEServer` helper that asserts POST + path + Authorization + Content-Type, closing the CodeRabbit gap from the prior review.

### Live re-verification against `api.novita.ai`

```
POST /openai/v1/chat/completions  -> 200 (llama-3.3-70b)
GET  /openai/v1/models            -> 200, 102 models
POST /openai/v1/chat/completions  -> deepseek-v3.1, enable_thinking=true
                                     -> choices[0].message.reasoning_content present
POST /openai/v1/chat/completions  -> deepseek-v3.1, enable_thinking=false
                                     -> choices[0].message.reasoning_content absent
POST /openai/v1/chat/completions  -> qwen3, enable_thinking=false
                                     -> content has no <think> tags
GET  /v1/models                   -> 404 (your typo-suggestion path)
```

## Not addressed in this commit

The Novita docs also list `/openai/v1/embeddings` and `/openai/v1/rerank` for Voyage-style and Cohere-shaped APIs. Those are out of scope for the maintainer's review here — happy to add them in a follow-up PR if useful (this PR is already a chat-only driver, matching what your screenshot tested).

Haruko386 (approved):
@tmimmanuel LGTM

PR #14852, opened 2026-05-12
fix: harden ES SQL match-predicate rewriting against quote-escape injection

sebastionoss:
## Summary

The regex in `common/doc_store/es_conn_base.py` that rewrites `WHERE field = 'value'` predicates into Elasticsearch `MATCH(...)` calls uses `[^']+` to capture the value. This pattern terminates at the first single quote, so a value containing an escaped quote (`''`) causes the regex to consume only the prefix — leaving attacker-controlled SQL fragments intact in the query forwarded to the ES SQL endpoint.

**CWE-89 (SQL Injection) — Moderate severity**

### Affected code

`common/doc_store/es_conn_base.py`, lines ~302–315 (the `sql()` method's match-predicate rewriting loop).

### Data flow

1. User submits a chat question to a dialog configured with `use_sql` retrieval.
2. The LLM generates SQL containing a crafted WHERE clause (prompt injection).
3. `sql_retrieval()` in `dialog_service.py` passes the SQL to `ESConnectionBase.sql()`.
4. The regex fails to capture the full value; trailing SQL (`OR 1=1 --`) survives rewriting.
5. `add_kb_filter` (tenant/KB scoping) is appended but neutralized by the injected clause.
6. The query executes against ES, potentially returning data from other tenants' indices.

### Fix description

- Replace `[^']+` with `(?:''|[^']+)+` so the regex correctly matches SQL-escaped quotes within string literals.
- Un-escape `''` → `'` before passing to the tokenizer.
- Re-escape any quotes produced by the tokenizer before interpolating into the MATCH literal.
- Extract the logic into a testable `_rewrite_match_predicates()` function.

### Tests

Added `test/unit_test/common/doc_store/test_es_conn_base_sql.py` with two test cases:
- Verifies that embedded escaped quotes are consumed by the regex (not left as trailing SQL).
- Verifies that tokenizer output containing quotes is properly escaped in the MATCH literal.

### Security analysis

The vulnerability requires: (1) an authenticated user, (2) a dialog with text-to-SQL enabled, and (3) the ability to prompt-inject the LLM into emitting a crafted WHERE clause. Impact is limited to read-only data exfiltration (ES SQL supports only SELECT, single statement) but can bypass tenant isolation enforced by `add_kb_filter`.

Before submitting, we verified that no existing mitigation prevents this. The `%` stripping and backtick removal do not affect single-quote handling. `add_kb_filter` is concatenated as plain text and is trivially neutralized by an injected `OR 1=1 --` suffix.

### DIFF
--- a/common/doc_store/es_conn_base.py
+++ b/common/doc_store/es_conn_base.py
@@ -34,6 +34,26 @@ from common import settings
 ATTEMPT_TIME = 2
 
 
+def _escape_sql_string_literal(value: str) -> str:
+    return value.replace("'", "''")
+
+
+def _rewrite_match_predicates(sql: str) -> str:
+    def replace(match: re.Match) -> str:
+        fld = match.group("field")
+        raw_value = match.group("value").replace("''", "'")
+        tokenized_value = rag_tokenizer.fine_grained_tokenize(rag_tokenizer.tokenize(raw_value))
+        tokenized_value = _escape_sql_string_literal(tokenized_value)
+        return " MATCH({}, '{}', 'operator=OR;minimum_should_match=30%') ".format(fld, tokenized_value)
+
+    return re.sub(
+        r" (?P<field>[a-z_]+_l?tks)(?P<operator> like | ?= ?)'(?P<value>(?:''|[^'])+)'",
+        replace,
+        sql,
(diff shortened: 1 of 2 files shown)


lewiswigmore:

PR #14877, opened 2026-05-13
fix: enforce read-only SQL allowlist in ExeSQL agent tool

sebastionoss:
## Summary

The `ExeSQL` agent tool executes SQL statements derived from LLM output, which is influenced by user input. The existing SQL filtering has two gaps:

1. **IBM DB2 code path** — no SQL statement filtering at all. Any SQL (DROP, ALTER, DELETE, etc.) is passed directly to `ibm_db.exec_immediate()`.
2. **Generic code path** (MySQL, PostgreSQL, MSSQL, etc.) — uses a regex that only blocks `INSERT`, `UPDATE`, and `DELETE`, missing destructive statements like `DROP`, `ALTER`, `CREATE`, `TRUNCATE`, `GRANT`, etc.

This is **CWE-89 (SQL Injection)** — an attacker who can influence the LLM prompt (e.g., via crafted natural-language queries to a ragflow agent with ExeSQL configured) can cause the LLM to emit destructive SQL that bypasses the incomplete blocklist.

**Affected file:** `agent/tools/exesql.py`

## Fix

Replace the regex-based blocklist with a robust **allowlist** approach using `sqlglot` for SQL parsing:

- Parse the SQL using `sqlglot` with dialect awareness (MySQL, PostgreSQL, TSQL, etc.)
- Enforce exactly **one statement** (no query stacking via `;`)
- Allow only read-only expression types: `SELECT`, `UNION`, `EXCEPT`, `INTERSECT`
- Reject any statement containing write/destructive subexpressions (`INSERT`, `UPDATE`, `DELETE`, `DROP`, `CREATE`, `ALTER`, `LOCK`, etc.)
- Apply this validation consistently to **all** database backends, including IBM DB2

## Testing

- Verified that valid `SELECT` and `UNION SELECT` queries pass validation
- Verified that `DROP TABLE`, `DELETE FROM`, `ALTER TABLE`, `CREATE TABLE`, `INSERT INTO`, `TRUNCATE`, and `UPDATE` statements are rejected with a clear error message
- Verified that query-stacking attempts (`SELECT 1; DROP TABLE users`) are rejected
- Verified that subquery-based attacks (`SELECT * FROM (DELETE FROM users RETURNING *)`) are caught by the subexpression scan

## Security Analysis

The ExeSQL tool is designed to let agents query databases in response to user questions. The SQL is generated by an LLM based on user input, making it attacker-influenced. An attacker with access to a ragflow agent using ExeSQL can craft prompts that cause the LLM to emit destructive SQL. The previous regex blocklist was trivially bypassable (e.g., `DROP TABLE`, `GRANT`, or any statement not in the short deny list). The IBM DB2 path had no protection at all.

Before submitting, we considered whether existing mitigations would prevent exploitation — specifically whether the LLM would refuse to generate destructive SQL. LLMs are reliably manipulable via prompt injection, and the ExeSQL tool passes LLM output directly to the database cursor without any structural validation. The regex filter on the generic path was incomplete by design (missing most DDL/DCL), so the risk is real.

The `sqlglot`-based allowlist is a defense-in-depth measure that works regardless of LLM behavior, covering all supported database dialects.

### DIFF
--- a/agent/tools/exesql.py
+++ b/agent/tools/exesql.py
@@ -22,6 +22,8 @@ import pandas as pd
 import pymysql
 import psycopg2
 import pyodbc
+import sqlglot
+from sqlglot import expressions as exp
 from agent.tools.base import ToolParamBase, ToolBase, ToolMeta
 from common.connection_utils import timeout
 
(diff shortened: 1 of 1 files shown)


andesyteoss:
Thanks for the thorough review! All three suggestions are valid and have been addressed in the latest commit:

1. **Normalize fenced/annotated SQL before first parse** (line 119 comment) — Added `_normalize_sql()` with proper `_CODE_FENCE_RE` (strips language tags) and `_ID_MARKER_RE` patterns. This is now called at the start of `_split_sql_statements` and in both execution loops, replacing the ad-hoc `_strip_sql_code_fences` + `re.sub` calls. LLM outputs like ````sql\nSELECT ...```` will now parse correctly.

2. **Add `exp.Into` to unsafe node check** (line 133 comment) — `exp.Into` is now included in the `statement.find(...)` call, blocking T-SQL `SELECT ... INTO` side-effecting writes.

3. **Add sanitized logging for validation rejects** (nitpick, lines 182-183) — Added `logger.warning()` calls at each rejection branch in `_ensure_read_only_sql`, logging `db_type`, validation stage, and exception/statement type — without exposing raw SQL or user data.

Also added docstrings to the new helper functions to address the docstring coverage check.

andesyteoss:
Thanks for the follow-up review! Both suggestions are valid — addressed in the latest commit:

1. **Narrowed `except Exception` to `SqlglotError`** — `_parse_sql_statements` now catches only `sqlglot.errors.SqlglotError`, so unrelated programming errors (attribute errors, bad dialect lookups, etc.) will propagate naturally instead of being masked as "Invalid SQL". Added a `logger.warning()` call consistent with the other validation paths.

2. **Removed dead `_strip_sql_code_fences`** — confirmed no callers remain; all paths use `_normalize_sql` directly.

lewiswigmore:

PR #14890, opened 2026-05-13
fix: require explicit anonymous webhook access

### LINKED ISSUE #14882: [Bug]: Agent webhook trigger is insecure-by-default — unauthenticated invocation when `security_cfg` is empty
### Self Checks

002bea448

### RAGFlow image version

002bea448 (nightly, main as of 2026-05-13)

```Markdown
- Hardware parameters: N/A (logic bug, reproducible on any deployment)
- OS type: N/A
- Others: Affects any agent whose `Begin` component is configured with `mode == "Webhook"` and no explicit security block.
```

### Actual behavior

The agent webhook trigger route is publicly accessible and silently accepts unauthenticated requests when the webhook's `security_cfg` is missing or empty.

In [api/apps/restful_apis/agent_api.py:1104-1148](api/apps/restful_apis/agent_api.py#L1104-L1148):

```python
@manager.route("/agents/<agent_id>/webhook", methods=["POST", "GET", "PUT", "PATCH", "DELETE", "HEAD"])
@manager.route("/agents/<agent_id>/webhook/test", methods=["POST", "GET", "PUT", "PATCH", "DELETE", "HEAD"])
async def webhook(agent_id: str):
    ...
    async def validate_webhook_security(security_cfg: dict):
        if not security_cfg:
            return  # No security config → allowed by default
```

Two distinct problems:

1. **No `@login_required` decorator** on either the `/webhook` or `/webhook/test` route. Every other `/agents/...` endpoint in the same file is `@login_required` (e.g. lines 153, 193, 231, 240, 286, 292, 313, 356, 444, 460, 494, 535, 550, 564, 581, 590, 641, 671, ...). The webhook routes are the only outliers.
2. **`validate_webhook_security` fails open.** When the agent's Begin/Webhook params do not include a `security_cfg` block (or it is `{}`), the function returns immediately and the request is processed with **no authentication, no IP allowlist, no token, no rate limit, and no body-size cap**.

The `/webhook/test` variant exposes the same hole.

Impact: anyone who learns or guesses an `agent_id` for an agent configured with a Webhook-mode Begin component but no security block can invoke that agent at will. Consequences include:

- LLM/API budget exhaustion (the attacker drives arbitrary completion calls on the victim's keys).
- Reach into whatever knowledge bases, tools, SQL connections, or external APIs the agent is wired to.
- Data exfiltration through any agent path that echoes retrieved chunks or tool output back in the response.

`agent_id` is a UUID but is **not** treated as a secret in the product — it appears in UI URLs, share links, server logs, and is returned in plain JSON from `/agents` (line 312) for any authenticated user including team members. Treating it as an auth credential is not defensible.

Severity: **High**. Not literally "Critical" only because exploitation requires (a) the target agent to be configured in Webhook mode and (b) knowledge of the `agent_id` — but neither is a meaningful barrier.

### Expected behavior

Secure-by-default. Specifically:

1. When `security_cfg` is missing/empty, the request should be **denied** rather than allowed, unless the operator has set an explicit `allow_anonymous: true` flag on the webhook config.
2. `auth_type == "none"` should require the same explicit opt-in.
3. The webhook configuration UI (Begin component, Webhook mode) should surface a warning when no auth/IP rule is set, and ideally refuse to save the canvas without either a security block or the explicit anonymous-allowed flag.
4. Even in the anonymous-allowed path, a default `max_body_size` and rate limit should apply (today both are bypassed when `security_cfg` is empty — see `_validate_max_body_size` and `_validate_rate_limit`).
5. `/webhook/test` should additionally require an authenticated session bound to the canvas owner/tenant, since it is a development-time surface and there is no reason for it to be reachable anonymously.

### Steps to reproduce

```Markdown
1. Log in to RAGFlow as user A and create a new Agent.
2. Drag a `Begin` component onto the canvas and set its mode to `Webhook`. Save without filling in any security fields under the webhook config (auth_type, token, IP allowlist, etc.).
3. Note the `agent_id` from the URL.
4. From an unauthenticated client (e.g. a fresh curl with no cookies/headers) issue:

   
   curl -X POST https://<host>/api/v1/agents/<agent_id>/webhook \
        -H "Content-Type: application/json" \
        -d '{"query":"hello"}'
   

5. Observe: the agent runs to completion and returns a 200 with the agent's output. No `401`, no `403`.
6. Repeat against `/api/v1/agents/<agent_id>/webhook/test` — same result.
7. (Optional) Confirm that all other `/agents/...` routes return `401` from the same unauthenticated client, demonstrating that the webhook routes are the only unprotected surface.
```

### Additional information

_No response_

ai-mountain:

Fixes #14882

Agent webhook execution currently fails open when the saved webhook `security` block is missing/empty, or when `auth_type` is set to `none`. This allows unauthenticated webhook invocation without an explicit operator opt-in.

This PR makes anonymous webhook access explicit:
- Rejects missing or empty webhook security config.
- Requires `allow_anonymous: true` when `auth_type` is `none`.
- Preserves explicit anonymous webhooks by having the frontend serialize `allow_anonymous: true` when the user selects `None` auth.
- Updates webhook unit tests to cover both denied implicit-anonymous configs and allowed explicit-anonymous configs.

### Type of change

- [x] Bug Fix
- [x] Security hardening
- [x] Test

### Tests

- [x] `ZHIPU_AI_API_KEY=dummy uv run python -m pytest --confcutdir=test/testcases/test_web_api/test_agent_app test/testcases/test_web_api/test_agent_app/test_agents_webhook_unit.py`
- [x] `uv run ruff check api/apps/restful_apis/agent_api.py test/testcases/test_web_api/test_agent_app/test_agents_webhook_unit.py`
- [x] `npm exec eslint src/pages/agent/utils.ts src/pages/agent/form/begin-form/schema.ts`


RenzoMXD:
@coderabbitai full review

PR #14910, opened 2026-05-14
Fix: dataset document download route

buua436:

dataset document download route
### Type of change


web/src/services/file-manager-service.ts
```diff
@@ -1,6 +1,7 @@
 import api from '@/utils/api';
 import registerServer from '@/utils/register-server';
 import request from '@/utils/request';
+export { downloadFileFromBlob } from '@/utils/file-util';
```
cike8899: The export function here is unnecessary and should be deleted.

PR #15002, opened 2026-05-19
Fix close llm transport in another event loop

wangq8:

Work around "llm noise" about close llm in another event loop

### Type of change


wangq8:
@coderabbitai

PR #15008, opened 2026-05-19
Go: implement provider: Perplexity

Crystora:
## What
- Add Perplexity as a chat and embedding provider backed by its OpenAI-compatible `/chat/completions` and `/v1/embeddings` APIs
- Register Perplexity in the Go model factory and provider config
- Support non-streaming chat, SSE streaming chat, embeddings, model listing, and connection checks

## Notes
- Uses the official Perplexity base URL `https://api.perplexity.ai` (chat lives at `/chat/completions`, embeddings at `/v1/embeddings`, models at `/v1/models` — endpoints verified against the live API)
- Forwards documented chat parameters from `ChatConfig`: `max_tokens`, `temperature`, `top_p`, `stop`, and `reasoning_effort` (only for `sonar-reasoning*` models)
- Routes Perplexity reasoning traces from `reasoning` / `reasoning_content` into `ReasonContent`
- `ListModels` accepts both the standard `{"data": [...]}` envelope and a bare array, matching observed Perplexity responses
- Configured models track the four chat models the docs declare accepted (`sonar`, `sonar-pro`, `sonar-reasoning-pro`, `sonar-deep-research`) plus the two embedding models (`pplx-embed-v1-0.6b`, `pplx-embed-v1-4b`)

## Tests
- `go test -vet=off -run TestPerplexity -count=1 ./internal/entity/models`
- `go test -vet=off -count=1 ./internal/entity/models`

Refs #14736

### DIFF
--- a/conf/models/perplexity.json
+++ b/conf/models/perplexity.json
@@ -1,0 +1,60 @@
+{
+  "name": "Perplexity",
+  "url": {
+    "default": "https://api.perplexity.ai"
+  },
+  "url_suffix": {
+    "chat": "chat/completions",
+    "embedding": "v1/embeddings",
+    "models": "v1/models"
+  },
+  "class": "perplexity",
+  "models": [
+    {
+      "name": "sonar",
+      "max_tokens": 127072,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "sonar-pro",
+      "max_tokens": 200000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "sonar-reasoning-pro",
+      "max_tokens": 127072,
+      "model_types": [
+        "chat"
+      ],
+      "thinking": {
+        "default_value": true,
+        "clear_thinking": true
+      }
+    },
+    {
+      "name": "sonar-deep-research",
+      "max_tokens": 127072,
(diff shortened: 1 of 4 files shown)


conf/models/perplexity.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "Perplexity",
+  "url": {
+    "default": "https://api.perplexity.ai"
+  },
+  "url_suffix": {
+    "chat": "chat/completions",
```
Haruko386: I could't find "chat/completions"
Let me know if you find it

Haruko386 (changes requested):
I could't find "chat/completions" in `Perplexity API Reference`
Let me know if you find it

conf/models/perplexity.json
Crystora: Good catch. The canonical endpoint is `/v1/sonar` per the official API Reference. I've updated pr. Please re-review.

Haruko386 (approved):
@bittoby LGTM

JinHai-CN:
@bittoby Since previous PR includes a binary file which is very big. So, I reset the head of main branch. Would you please re-submit the commit from current HEAD of main branch? Thank you very much.

PR #15009, opened 2026-05-19
build(go): make `bash build.sh` work on macOS arm64 (Homebrew)

Zzappy24:
## Problem

The Go server build pipeline (`build.sh` + CMake + CGO bindings) was tested on Ubuntu only. On macOS arm64 with Homebrew it fails in five orthogonal places. None of these require platform-specific code paths — the same source builds on both Linux and Darwin after these fixes.

## Reproduction (before)

```
$ uname -a
Darwin … 25.4.0 arm64
$ brew install cmake pcre2 simde
$ bash build.sh
…
error: 'simde/x86/sse4.1.h' file not found
error: implicit instantiation of undefined template 'std::basic_istringstream<char>'
error: no matching function for call to 'Join'
…
clang: error: no such file or directory: '/usr/local/lib/libpcre2-8.a'
```

### 1. `internal/cpp/CMakeLists.txt` — find Homebrew + libpcre2-8 portably

- Detect Apple platforms via `if(APPLE)`, call `brew --prefix` once, add `${HOMEBREW_PREFIX}/include` and `${HOMEBREW_PREFIX}/lib`. No effect on Linux.
- Replace the literal `libpcre2-8.a` link token (which only the Linux linker finds in `/usr/local/lib` by default) with `find_library(PCRE2_LIB NAMES pcre2-8 REQUIRED)`. Works on `/usr/lib/x86_64-linux-gnu` (Debian/Ubuntu), `/usr/local/lib` (Intel Mac & legacy Linux), `/opt/homebrew/lib` (Apple Silicon).

### 2. `internal/cpp/wordnet_lemmatizer.cpp` + `internal/cpp/rag_analyzer.cpp` — explicit `#include <sstream>`

libstdc++ (Linux) pulls `<sstream>` in transitively via `<fstream>`; libc++ (Apple Clang) doesn't, so the existing `std::istringstream` / `std::ostringstream` uses fail to compile on macOS. One-line include in each file.

### 3. `internal/cpp/rag_analyzer.cpp` — `Join` template overload fix

`Join(tokens, start, tokens.size(), delim)` at line 146 passes `size_t` to an `int` parameter. C++23 strict mode in Apple Clang refuses the implicit narrowing and reports the 4-arg overload as a substitution failure, leaving the call ambiguous between the 3-arg and 4-arg templates. Fix: explicit `static_cast<int>(tokens.size())`. Behaviour identical on libstdc++ — the narrowing was always intentional.

### 4. `internal/binding/rag_analyzer.go` — split darwin CGO LDFLAGS

The existing `#cgo darwin LDFLAGS: ... /usr/local/lib/libpcre2-8.a` only matches Intel Macs. Apple Silicon Homebrew installs to `/opt/homebrew`. Split into `darwin,arm64` and `darwin,amd64` build constraints with the right absolute path on each.

### 5. `build.sh` — accept Homebrew path in the pcre2 sanity check

The sanity check looked at two Linux paths only and then fell through to `sudo apt -y install libpcre2-dev` on failure. Added `/opt/homebrew/lib/libpcre2-8.a`, and on Darwin failure now exits cleanly with the right `brew install pcre2` hint instead of trying `apt`.

## Verified

- `bash build.sh` now completes on macOS arm64 (Apple Silicon, brew 4.x, cmake 4.x, Apple Clang 17, Go 1.25, pcre2 10.x, simde 0.8.x).
- Produced binaries: `bin/server_main`, `bin/admin_server`, `bin/ragflow_cli`.
- `bin/server_main` boots, connects MySQL, runs migrations, loads the 64 model provider configs cleanly.
- Still builds on Linux — the CMake additions are inside an `if(APPLE)` guard, the `find_library` call matches Linux paths too, the build.sh check still tries `apt` when not on Darwin.

## Out of scope

The Go server itself currently fails at runtime when not pointing at Elasticsearch (`Failed to initialize doc engine: failed to ping Elasticsearch`), but that's the placeholder Infinity engine documented in `internal/engine/README.md` — unrelated to this build patchset.

---

Happy to split this into smaller PRs if you'd prefer (one per file). The five changes are independent.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -84,10 +84,17 @@ build_go() {
         exit 1
     fi
 
-    # Check for pcre2 library
-    if [ -f "/usr/lib/x86_64-linux-gnu/libpcre2-8.a" ] || [ -f "/usr/local/lib/libpcre2-8.a" ]; then
+    # Check for pcre2 library — known Linux paths + macOS Homebrew (Apple Silicon
+    # at /opt/homebrew, Intel Macs at /usr/local).
+    if [ -f "/usr/lib/x86_64-linux-gnu/libpcre2-8.a" ] \
+       || [ -f "/usr/local/lib/libpcre2-8.a" ] \
+       || [ -f "/opt/homebrew/lib/libpcre2-8.a" ]; then
         echo "✓ pcre2 library found"
     else
+        if [ "$(uname)" = "Darwin" ]; then
+            echo -e "${RED}Error: libpcre2-8.a not found. Install with: brew install pcre2${NC}"
+            exit 1
+        fi
         echo -e "${YELLOW}Warning: libpcre2-8.a not found. You may need to install libpcre2-dev:${NC}"
         sudo apt -y install libpcre2-dev
     fi
--- a/internal/binding/rag_analyzer.go
+++ b/internal/binding/rag_analyzer.go
@@ -19,7 +19,9 @@ package rag_analyzer
 /*
 #cgo CXXFLAGS: -std=c++20 -I${SRCDIR}/..
 #cgo linux LDFLAGS: ${SRCDIR}/../cpp/cmake-build-release/librag_tokenizer_c_api.a -lstdc++ -lm -lpthread /usr/lib/x86_64-linux-gnu/libpcre2-8.a
-#cgo darwin LDFLAGS: ${SRCDIR}/../cpp/cmake-build-release/librag_tokenizer_c_api.a -lstdc++ -lm -lpthread /usr/local/lib/libpcre2-8.a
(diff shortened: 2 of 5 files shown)


JinHai-CN:
@Zzappy24 Since previous PR includes a binary file which is very big. So, I reset the head of main branch. Would you please re-submit the commit from current HEAD of main branch? Thank you very much.

Zzappy24:
of course no problem, thank you for you response, I will re-submit as soon as possible

Zzappy24:
Re-submitted: rebased onto the current `main` HEAD (`dc01e0e51`) and force-pushed. The branch is now a single clean commit, thanks for your work, I will do my best to help if I can, have a nice day

PR #15010, opened 2026-05-19
Go: implement provider: n1n.ai

sxxtony:
> **Update (commit d3e01a7)** — addresses maintainer review:
> - Added `thinking: {type: "enabled" | "disabled"}` body field via `ChatConfig.Thinking` (per @Haruko386's example curl against `deepseek-v3-1-250821`).
> - Added `deepseek-v3-1-250821` and `deepseek-v3-1-think-250821` to the shipped catalog.
> - `ChatWithMessages` now leaves `ChatResponse.ReasonContent` as `nil` when the upstream omits `reasoning_content` (CodeRabbit nit).
> - Live test transcripts confirming all of the above are at https://github.com/infiniflow/ragflow/pull/15010#issuecomment-4494077117

---

### What problem does this PR solve?

Add a Go driver for **n1n.ai** (https://docs.n1n.ai), one of the unchecked providers on the umbrella tracking issue #14736. n1n.ai is an OpenAI-compatible aggregator hosting a 450+ model catalog (GPT, Claude, Gemini, DeepSeek, Kimi, Qwen, embedding + reranker families) under `https://api.n1n.ai/v1`.

Until this PR, a tenant who configured `n1n` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver.

### What this PR includes

- `internal/entity/models/n1n.go` — `N1NModel` implementing `ModelDriver`.
- `conf/models/n1n.json` — base URL + versioned suffixes (bare host + `v1/...` paths, matching the maintainer's split on Novita / LongCat) and a representative 12-model catalog (chat + embedding + rerank).
- `factory.go` — route `"n1n"` to `NewN1NModel`.

### Method coverage (docs-driven)

| ModelDriver method | n1n.ai endpoint | Status |
| --- | --- | --- |
| `ChatWithMessages` | `POST /v1/chat/completions` | ✅ implemented |
| `ChatStreamlyWithSender` | `POST /v1/chat/completions` (SSE, `[DONE]`) | ✅ implemented |
| `Embed` | `POST /v1/embeddings` | ✅ implemented |
| `Rerank` | `POST /v1/rerank` | ✅ implemented |
| `ListModels` | `GET /v1/models` | ✅ implemented |
| `CheckConnection` | delegates to `ListModels` | ✅ implemented |
| `Balance` | not documented | `"n1n, no such method"` |
[...]

PR #14811, opened 2026-05-12
Go: implement provider: Voyage AI

tmimmanuel:
### What problem does this PR solve?

Add a Go driver for Voyage AI (https://voyageai.com), one of the unchecked providers on the umbrella tracking issue #14736. Voyage AI is **embed + rerank only** — no chat, no streaming, no `/v1/models` endpoint. It's the first provider in the Go layer of this shape.

Until this PR, a tenant who configured `voyage` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver.

### What this PR includes

- New `internal/entity/models/voyage.go` with a `VoyageModel` implementing the `ModelDriver` interface.
- New `conf/models/voyage.json` with 6 embedding models (`voyage-3.5`, `voyage-3.5-lite`, `voyage-3-large`, `voyage-code-3`, `voyage-law-2`, `voyage-finance-2`) and 2 rerank models (`rerank-2`, `rerank-2-lite`).
- `factory.go`: route `"voyage"` to `NewVoyageModel`.
- `internal/entity/models/voyage_test.go`: 19 unit tests.

### How the driver works

- **Embed**: `POST /v1/embeddings`. Response is OpenAI-shaped (`{data: [{embedding, index, object, text}], model, usage}`). Driver reorders by `index`, rejects duplicate / out-of-range / missing slots, and short-circuits empty input without an HTTP call.
- **Rerank**: `POST /v1/rerank`. Voyage uses **`top_k`** as the request param name (not `top_n` like Aliyun/SiliconFlow); the driver translates `RerankConfig.TopN` → `top_k`. Response is Cohere-shaped (`{data: [{relevance_score, index}], model}`), so the existing `RerankResponse{Data: []RerankResult{Index, RelevanceScore}}` shape fits cleanly.
- **`ListModels`**: returns a hardcoded list of `voyageKnownModels`. Voyage does **not** expose `/v1/models` (probed live, returns 404), so the driver synthesizes the list from the same set the config ships. New upstream models are added by extending one slice.
- **`CheckConnection`**: pings a 1-input embed call against `voyage-3.5`. Without `/v1/models`, this is the cheapest way to verify the API key + network path before a tenant tries a real workload.
- **`ChatWithMessages` / `ChatStreamlyWithSender` / `Balance` / `TranscribeAudio` / `AudioSpeech` / `OCRFile`**: all return `"no such method"`. Voyage does not host any of these surfaces.

No interface change. No new dependencies.

### How was this tested?

[...]

PR #15013, opened 2026-05-19
Go: implement provider: FuturMix

sxxtony:

Add a Go driver for **FuturMix** (https://futurmix.ai/docs), one of the unchecked providers on the umbrella tracking issue #14736. FuturMix is documented as an "OpenAI-compatible API" aggregator over Claude / GPT / Gemini / DeepSeek (~22 models per their `/models` page).

Until this PR, a tenant who configured `futurmix` as a model provider in the Go layer fell through to the default branch of `internal/entity/models/factory.go` and got the dummy driver.

### What this PR includes

- `internal/entity/models/futurmix.go` — `FuturMixModel` implementing `ModelDriver`.
- `conf/models/futurmix.json` — base URL + chat suffix and a 6-model representative catalog (GPT, Claude, Gemini, DeepSeek families).
- `factory.go` — route `"futurmix"` to `NewFuturMixModel`.

### Method coverage (docs-driven, no live testing)

| ModelDriver method | FuturMix endpoint | Status | Why |
|---|---|---|---|
| `ChatWithMessages` | `POST /v1/chat/completions` | ✅ implemented | Marketing meta on `/features` reads "OpenAI-compatible API"; admin-UI placeholder uses `https://api.openai.com/v1/chat/completions` as the canonical endpoint shape; FuturMix is a one-api-lineage gateway (confirmed by the `/api/status` JSON shape) |
| `ChatStreamlyWithSender` | `POST /v1/chat/completions` (`stream:true`, OpenAI SSE) | ✅ implemented | Same as above |
| `ListModels` | not documented | `"futurmix, no such method"` | No `/v1/models` published in docs or JS bundle |
| `CheckConnection` | no documented health endpoint | `"futurmix, no such method"` | Avoid burning tenant quota on a fake ping |
| `Embed` | not documented | `"futurmix, no such method"` | |
| `Rerank` | not documented | `"futurmix, no such method"` | |
| `Balance` | not documented | `"futurmix, no such method"` | |
| `TranscribeAudio`, `AudioSpeech` | not documented | `"futurmix, no such method"` | |
| `OCRFile`, `ParseFile`, `ListTasks`, `ShowTask` | not documented | `"futurmix, no such method"` | |

Per the maintainer's guidance on #14809 review #4277917390 ("there is no need to implement the interface that is not officially given"), this driver covers only what FuturMix's public surface explicitly publishes.

### Two endpoints deliberately deferred

The FuturMix admin UI JS bundle declares CLI tracing rules that point to two additional `/v1` surfaces:

```js
codexCli:  { path_regex: ["/v1/responses"], ... }   // OpenAI Responses API format
claudeCli: { path_regex: ["/v1/messages"],  ... }   // Anthropic Messages format
```

Both require request/response shapes different from the `ChatWithMessages` interface and are worth a separate PR. The OpenAI-compatible `/v1/chat/completions` surface this PR implements is the one any aggregator with that marketing label is expected to ship and is sufficient for the standard RAGFlow chat flow.

### URL split

Same bare-host + versioned-suffix convention used on Novita (#14850), LongCat (#14809), Voyage (#14811), and n1n.ai (#15010):

```json
"url":        { "default": "https://futurmix.ai" }
"url_suffix": { "chat":    "v1/chat/completions" }
```

### Why no live testing

FuturMix gates API key issuance behind an access-request flow on `/get-started` — there is no self-service signup that would let an external contributor smoke-test the driver pre-PR. The driver is structurally identical to the OpenAI / CometAPI / n1n.ai / Novita drivers (all OpenAI-compatible aggregators in this package), and ships only what the FuturMix public docs explicitly call out. Happy to follow up with live test transcripts once a tenant key is available, and to wire up `/v1/messages` and `/v1/responses` once their wire shapes are needed.


conf/models/futurmix.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    {
+      "name": "deepseek-v3",
+      "max_tokens": 64000,
+      "model_types": [
+        "chat"
+      ]
+    }
+  ]
+}
```
Haruko386, line 108: https://futurmix.ai/console/pricing

Haruko386 (changes requested):
Hi @sxxtony,  please read the docs for precise info

JinHai-CN:
@sxxtony  Since previous PR includes a binary file which is very big. So, I reset the head of main branch. Would you please re-submit the commit from current HEAD of main branch? Thank you very much.

sxxtony:
@Haruko386 @JinHai-CN — pushed the rebase + the catalog fix in one go (`ff323693`).

### Catalog (Haruko386's review)

You're right that the previous catalog was guessed. Pulled the real list from `GET /api/pricing` on `futurmix.ai` (public, no auth) — 22 entries. Filtered to chat-capable models whose `supported_endpoint_types` include `"openai"` (the `/v1/chat/completions` surface this driver targets) and shipped 12 representative models across the three families FuturMix actually serves:

| family | models in new catalog | max_tokens |
| --- | --- | --- |
| GPT-5 | `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano` | 400000 |
| Claude-4 | `claude-opus-4-7`, `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-haiku-4-5-20251001` | 200000 |
| Gemini-2.5/3.1 | `gemini-3.1-pro-preview`, `gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-2.5-flash-lite` | 1M – 2M |

Skipped the image-only models (`gpt-image-2`, `doubao-seedance-2-0-260128`) since they don't belong in a chat catalog. Skipped the `*-image-preview` Gemini variants for the same reason.

Live sanity check from `/api/pricing`:

```
GET https://futurmix.ai/api/pricing
→ HTTP 200, data.length = 22, supported_endpoint = {
    "openai":    {"path": "/v1/chat/completions",                   "method": "POST"},
    "anthropic": {"path": "/v1/messages",                            "method": "POST"},
    "gemini":    {"path": "/v1beta/models/{model}:generateContent", "method": "POST"}
  }
```

The OpenAI surface this driver implements is now confirmed by the same `supported_endpoint` mapping FuturMix publishes — every model in the new catalog has `"openai"` in its `supported_endpoint_types`.

### Rebase (JinHai-CN's reset)

Done. Cherry-picked both commits onto current `main` (`dc01e0e5`):

| commit | description |
| --- | --- |
| `0687c16d` | original FuturMix driver |
| `ff323693` | catalog fix from `/api/pricing` |

One factory-route conflict against the new `case "jiekouai" / "302.ai" / "mineru_local"` entries that landed on main; resolved by keeping all of them plus appending `case "futurmix"` last. GitHub now reports `mergeable: true`.

Haruko386 (approved):
LGTM

sxxtony:
Addressed the CodeRabbit SSE review at `futurmix.go:382` in `7bd39ca5`.

The previous loop parsed each `data:` line standalone, which would break on any SSE event that wrapped JSON across multiple `data:` lines (per [the HTML Living Standard's SSE section](https://html.spec.whatwg.org/multipage/server-sent-events.html#dispatchMessage), a single event can span several `data:` lines and the consumer must join them with newlines before parsing).

New shape:

| Behavior | Before | After |
| --- | --- | --- |
| `data:` line accumulation | unmarshal per line | buffered into `dataLines` |
| Event terminator | implicit (every line) | explicit (blank line, per spec) |
| Trailing event without blank line | dropped | flushed before returning |
| Single-space-after-colon stripping | `TrimSpace` (greedy) | strip at most one space per spec |
| `event:` / `id:` / `retry:` fields | already ignored | still ignored |

No behavior change for OpenAI-style one-line-per-event streams (they already emit a blank line between events). The fix only matters when an upstream emits a multi-line `data:` event — exactly the class of bug CodeRabbit caught.

Build clean (`go build ./internal/...`).

PR #15018, opened 2026-05-19
feat(retrieval): expose source document download URL on chunk responses

### LINKED ISSUE #14771: [Feature Request] Support file download via chunk retrieval - return downloadable URL from search results
## Feature Description

When users retrieve relevant chunks from a RAGFlow knowledge base, the search results currently return the chunk content and metadata, but do not provide a direct way to download or access the original source file. Users need the ability to obtain a downloadable URL or file path for the document that the chunk belongs to.

## Expected Behavior

- When a chunk is returned from a search, the API response should include a reference to the source document
- A downloadable URL (e.g., /api/v1/datasets/{dataset_id}/documents/{document_id}/download) should be accessible or derivable from the chunk metadata
- The frontend UI should offer a "Download File" or "View Source" option for each chunk result

## Use Case

- Users perform a semantic search and want to access the full original document from which a chunk originated
- Integrating RAGFlow retrieval results into external workflows that require the original file
- Auditing or reviewing source documents in full after seeing relevant chunk excerpts

## Additional Context

This is particularly relevant for document sets where only a specific chunk is relevant. The ability to download the full source file completes the retrieval loop and enables end-to-end document workflows.

Khaostica:

Closes [#14771](https://github.com/infiniflow/ragflow/issues/14771).

Chunk-retrieval responses currently return `document_id` and `dataset_id` but no URL to fetch the source file. Callers who want to surface a "Download File" / "View Source" affordance next to a retrieved chunk have to know the API surface well enough to compose the path themselves, and the path differs between the session-auth and token-auth routes — which is easy to get wrong.

This PR adds a `document_download_url` field to every chunk returned by the retrieval/chunk API surface, pointing at the existing token-auth file-stream route `GET /api/v1/datasets/{dataset_id}/documents/{document_id}` (declared at `api/apps/sdk/doc.py:54`).

**URL choice.** The issue proposes `/api/v1/datasets/{dataset_id}/documents/{document_id}/download`. That exact path doesn't exist; the closest token-authenticated endpoint is `GET /api/v1/datasets/{dataset_id}/documents/{document_id}` which already streams the file. The separate `GET /api/v1/documents/{doc_id}/download` route is session-cookie-authenticated. Since the retrieval response is consumed primarily by SDK/API-token callers, the SDK route is the right pairing.

**Changes**

- New helper `build_document_download_url(dataset_id, document_id)` in `api/utils/api_utils.py`. Returns the canonical path, or `None` if either id is missing — knowledge-graph synthetic chunks occasionally surface without ids and we don't want to advertise an unusable URL.
- Wired into:
  - `POST /api/v1/retrieval` (SDK retrieval) — `api/apps/sdk/doc.py`, inside the renamed-keys block.
  - `POST /api/v1/datasets/search` and `POST /api/v1/datasets/{dataset_id}/search` — via `api/apps/services/dataset_api_service.py` (both `search` and `search_datasets`).
  - `GET /api/v1/datasets/{dataset_id}/documents/{document_id}/chunks` (both single-id lookup and listing branches) — `api/apps/restful_apis/chunk_api.py`. The `Chunk` pydantic model gets the new optional field so validation still passes.
  - `POST /api/v1/dify/retrieval` — surfaced inside the `metadata` object alongside `document_id`/`doc_id` — `api/apps/sdk/dify_retrieval.py`.

**Backwards compatibility**

Additive only. The field is omitted (not `null`) when either id is missing, so downstream consumers checking `if "document_download_url" in chunk` get a clean signal.

**Test plan**

- New unit test `test/unit_test/api/utils/test_build_document_download_url.py` — happy path, all id-missing combinations, verifies inputs aren't url-encoded (matches how Flask/Quart route patterns work elsewhere).
- Extended `test/testcases/restful_api/test_retrieval.py::test_dataset_search_rest_endpoint` and `::test_retrieval_compatibility_endpoint` to assert the new field is present and correct on every returned chunk.
- Extended `test/testcases/restful_api/test_chunks.py::test_chunks_add_list_get_update_delete_cycle` to assert the field on the chunk-listing response.
- REST integration suite will run under `.github/workflows/tests.yml` on PR open.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -38,6 +38,7 @@ from api.db.services.task_service import TaskService, cancel_all_task_of, queue_
 from api.db.services.tenant_llm_service import TenantLLMService
 from api.utils.api_utils import (
     add_tenant_id_to_kwargs,
+    build_document_download_url,
     check_duplicate_ids,
     construct_json_result,
     get_error_data_result,
@@ -103,6 +104,7 @@ class Chunk(BaseModel):
     image_id: str = ""
     available: bool = True
     positions: list[list[int]] = Field(default_factory=list)
+    document_download_url: str = ""
 
     @validator("positions")
     def validate_positions(cls, value):
@@ -365,6 +367,13 @@ async def retrieval_test(tenant_id):
             "kb_id": "dataset_id",
         }
         ranks["chunks"] = [{key_mapping.get(key, key): value for key, value in chunk.items()} for chunk in ranks["chunks"]]
+        for rename_chunk in ranks["chunks"]:
+            download_url = build_document_download_url(
+                rename_chunk.get("dataset_id"),
+                rename_chunk.get("document_id"),
+            )
+            if download_url:
+                rename_chunk["document_download_url"] = download_url
         return get_result(data=ranks)
     except Exception as e:
         if "not_found" in str(e):
@@ -424,6 +433,9 @@ async def list_chunks(tenant_id, dataset_id, document_id):
             "tag_kwd": chunk.get("tag_kwd", []),
             "tag_feas": chunk.get("tag_feas", {}),
         }
+        download_url = build_document_download_url(final_chunk["dataset_id"], final_chunk["document_id"])
+        if download_url:
+            final_chunk["document_download_url"] = download_url
         res["chunks"].append(final_chunk)
         _ = Chunk(**final_chunk)
     elif settings.docStoreConn.index_exist(search.index_name(dataset_tenant_id), dataset_id):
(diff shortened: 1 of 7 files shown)


Khaostica:
@JinHai-CN Resolved conflicts, this PR is ready for review.

Khaostica:
@JinHai-CN, @wangq8, this PR has been fixed and is ready for review.

wangq8:
The document download url based on:
1. the host
2. the port
3. the uri suffix path

While in true enviroment, the host/port might be routed, so the host/port exposed in code might be wrong.
and the uri suffix path also might be out of date due to the REST API refine or version bump.

so let user read the document to compose the real document download url will be a good choice for programer.

Issue #14881, opened 2026-05-13
[Feature Request]: Support merging and splitting chunks


caesergattuso:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

Background
RAGFlow already supports manual editing of individual chunks, but the user experience is not ideal. In practice, users often need to:
Merge chunks that were over-split during automatic parsing
Manually split a chunk at a semantically incorrect boundary
Currently, both operations lack proper UI support — users have to either re-upload the document or directly modify the database.
Proposed Features
1. Merge Adjacent Chunks
Allow users to select two adjacent chunks (Chunk N and Chunk N+1) in the chunk list and merge them into a single chunk.
Entry point: Multi-select + "Merge" button, or keyboard shortcut
Behavior: Text content is concatenated; a new chunk_id is generated
Validation: Warn or block if the merged chunk exceeds the max token limit
Audit: Merge action should be logged in the history/audit trail
2. Split a Chunk (Key Feature)
Allow users to insert a split point inside a chunk, dividing it into two separate chunks.
Entry point: Insert a split marker within the chunk editor / preview
Recommended UX: A "Split here" button at the cursor position in preview mode, or inserting a --- marker
Split point precision: Character offset or token position
Behavior: Both resulting chunks are independent — can be further edited, merged, or split again
Metadata: New chunks inherit the original chunk's document ID and parser info; timestamps are updated
Use Cases
Over-chunking: chunk.token_num set too low → a single paragraph split into 3-4 chunks → merge back at semantically appropriate points
Table truncation: Table rows split across chunks → merge to restore table integrity
Code truncation: Function body split mid-function → merge to recover the complete function
Manual split: Automatic parsing cuts a semantic unit at the wrong position → user splits manually at the correct boundary
Additional Considerations
Both merge and split should be reversible (user can re-split or re-merge)
Preview before committing the result
Consider updating the vector index automatically after merge/split
[...]

PR #15019, opened 2026-05-19
feat(api): add chunks/merge endpoint to combine adjacent chunks

davidv399:

Refs #14881 (backend portion, merge half)

Adds a backend endpoint that lets users combine two or more chunks of a document into a single chunk. Pairs with a follow-up PR that adds the chunk split endpoint.

When automatic parsing over-splits a document (paragraphs broken across chunks, table rows separated, function bodies cut mid-function), users currently have to either re-upload the document or modify the database directly. This endpoint makes the operation a single API call.

### Endpoint

`POST /api/v1/datasets/<dataset_id>/documents/<document_id>/chunks/merge`

Body:

```json
{ "chunk_ids": ["id1", "id2", "..."] }
```

Behavior:

- Validates dataset and document ownership using the same `KnowledgebaseService.accessible` + `_get_dataset_tenant_id` guards used by the existing `add_chunk` and `rm_chunk` handlers.
- Requires at least two chunk ids; rejects duplicates; requires each id to exist in the target document.
- Concatenates `content_with_weight` in the order supplied with a blank-line joiner.
- Unions `important_kwd`, `question_kwd`, and `tag_kwd` from the source chunks (de-duplicated, order preserved).
- Generates a new deterministic `chunk_id` via `xxh64(merged_content + document_id)`, matching the `add_chunk` convention.
- Re-tokenizes, re-embeds, and inserts the merged chunk through the same `docStoreConn.insert` path the existing endpoints use.
- Deletes the source chunks and adjusts the document and knowledgebase chunk counters via `DocumentService.increment_chunk_num` / `decrement_chunk_num`, matching the existing add/remove pattern.
- Rejects merge for `ParserType.QA` documents because the Q&A chunk format does not merge cleanly.

### Out of scope

- Frontend UI (separate workstream).
- Image-bearing chunks are not merged into the result; users can re-attach via the existing `PATCH chunks/<id>` endpoint if needed.
- Reversibility / undo is a UI concern.

### Test plan

- Manual: create a document, identify two adjacent chunks, POST `chunks/merge` with their ids, confirm the response returns a new chunk whose `content` is the concatenation, and that the source chunks no longer appear in `GET chunks`.
- Negative paths covered: missing/short `chunk_ids`, duplicate ids, ids belonging to a different document, ids not found, QA-parser documents.

### Type of change

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -352,6 +352,133 @@ async def rm_chunk(tenant_id, dataset_id, document_id):
     return get_result(message=f"deleted {chunk_number} chunks")
 
 
+@manager.route("/datasets/<dataset_id>/documents/<document_id>/chunks/merge", methods=["POST"])  # noqa: F821
+@login_required
+@add_tenant_id_to_kwargs
+async def merge_chunks(tenant_id, dataset_id, document_id):
+    from rag.nlp import rag_tokenizer, search
+
+    if not KnowledgebaseService.accessible(kb_id=dataset_id, user_id=tenant_id):
+        return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
+    dataset_tenant_id = _get_dataset_tenant_id(dataset_id)
+    if not dataset_tenant_id:
+        return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
+    doc = DocumentService.query(id=document_id, kb_id=dataset_id)
+    if not doc:
+        return get_error_data_result(message=f"You don't own the document {document_id}.")
+    doc = doc[0]
+    if doc.parser_id == ParserType.QA:
+        return get_error_data_result(message="Merging is not supported for Q&A documents.")
+    req = await get_request_json()
+    if not req:
+        return get_error_data_result(message="`chunk_ids` is required")
+    chunk_ids = req.get("chunk_ids")
+    if not isinstance(chunk_ids, list) or len(chunk_ids) < 2:
+        return get_error_data_result(message="`chunk_ids` must be a list of at least 2 chunk ids")
+    if any(not isinstance(cid, str) or not cid for cid in chunk_ids):
+        return get_error_data_result(message="`chunk_ids` must contain non-empty strings")
+    seen = set()
+    for cid in chunk_ids:
+        if cid in seen:
+            return get_error_data_result(message=f"Duplicate chunk id {cid} in `chunk_ids`")
+        seen.add(cid)
+
+    index = search.index_name(dataset_tenant_id)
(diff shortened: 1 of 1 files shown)


davidv399:
Thanks @coderabbitai. Addressed both findings in b11cda2aa:

**1. Atomic / compensating write (Major)**
The flow is now: insert merged chunk → `try` { delete source chunks (require `deleted == len(seen)`), then `increment_chunk_num`, then `decrement_chunk_num` (with an inner try that reverts the increment if the decrement fails) }. On any exception in that block, we log the error with `dataset_id`/`document_id`/`new_chunk_id`/`source_ids` and best-effort `docStoreConn.delete` the just-inserted merged chunk so the store never ends up with both the merged chunk and the source chunks visible, and counters never drift. The rollback delete is itself wrapped so a failing rollback is logged but does not mask the original error. Matches the same pattern applied to `split_chunk` in #15020.

**2. Structured logs (Nitpick)**
Added `logging` import and structured log lines on:
- INFO at function entry (`dataset_id`, `document_id`)
- WARN on each early-return branch: ownership rejected, duplicate id in `chunk_ids`, source chunk missing (both `None` and `NotFoundError` paths), id collision with a source
- ERROR on post-insert failure and on rollback-delete failure (with `new_chunk_id`, `source_ids`, and the underlying error)
- INFO on success (`new_chunk_id`, `source_ids`, `deleted`)
Style matches the existing f-string `key=value` logging used in `task_api.py` and `dataset_api.py`.

Also caught and applied while in the area: `docStoreConn.get` is now wrapped in a try/except that treats `NotFoundError` as a normal missing-chunk path (same pattern as `get_chunk`), so an unknown `chunk_id` returns the data error instead of a 500.

Syntax check passes.

JinHai-CN:
@jack-stormentswe  Since previous PR includes a binary file which is very big. So, I reset the head of main branch. Would you please re-submit the commit from current HEAD of main branch? Thank you very much.

davidv399:
Hi @JinHai-CN, thanks for the heads up. Rebased onto the current HEAD of main and force-pushed (b55dc6520). No binary files, only the two commits that touch `api/apps/restful_apis/chunk_api.py`. Ready for re-review.

KevinHuSh:
Appreciations!

Please append some instructions about the API beneath it.

wangq8:
Key issues: Some chunks may contain images (e.g., PDFs often generate image based chunks), which are not considered here. Merging such chunks is difficult because the associated images data would also need to be merged correctly. Chunk position information is lost after merging. Merging should be restricted to adjacent chunks only; non-adjacent chunks should not be merged. Additionally, the methos get_model_config_by_type_and_name / and get_model_config_by_id are stale.

PR #15020, opened 2026-05-19
feat(api): add chunks/<id>/split endpoint to divide a chunk at an offset

davidv399:

Refs #14881 (backend portion, split half)

Adds a backend endpoint that divides a single chunk into two chunks at a user-supplied character offset. Pairs with PR #15019 (chunks/merge endpoint).

When automatic parsing places a chunk boundary in the wrong place (cutting a semantic unit in the middle), users currently have to delete the chunk and add two new chunks by hand, losing keywords and tags in the process. This endpoint makes the operation a single API call.

### Endpoint

`POST /api/v1/datasets/<dataset_id>/documents/<document_id>/chunks/<chunk_id>/split`

Body:

```json
{ "split_point": 256 }
```

Behavior:

- Validates dataset and document ownership using the same `KnowledgebaseService.accessible` + `_get_dataset_tenant_id` guards used by the existing `add_chunk`, `update_chunk`, and `rm_chunk` handlers.
- Loads the source chunk and confirms it belongs to the given document.
- Validates `split_point` is an integer strictly between `0` and `len(content)` (exclusive on both ends), and that neither resulting half is empty.
- Slices `content_with_weight` into a left and right half, generates a new deterministic `chunk_id` for each via `xxh64(half_content + document_id)` matching the `add_chunk` convention.
- Carries `important_kwd`, `question_kwd`, `tag_kwd`, and `tag_feas` over to both halves so user-curated metadata is not lost on split.
- Re-tokenizes, re-embeds, and inserts both new chunks through the same `docStoreConn.insert` path the existing endpoints use.
- Deletes the source chunk and adjusts the document and knowledgebase chunk counters via `DocumentService.increment_chunk_num` / `decrement_chunk_num`, matching the existing add/remove pattern.
- Rejects split for `ParserType.QA` documents because the Q&A chunk format does not split cleanly.

### Out of scope

- Frontend UI (separate workstream).
- Token-offset (vs character-offset) split: deferred for v1 since token offsets depend on the tokenizer choice; character offsets are unambiguous.
- Image-bearing chunks: the source image is not copied to either half. Users can re-attach via the existing `PATCH chunks/<id>` endpoint if needed.
- Reversibility / undo is a UI concern.

### Test plan

- Manual: create a document, pick a chunk, POST `chunks/<id>/split` with a mid-content offset, confirm the response returns two chunks whose `content` is left/right halves and that the source chunk no longer appears in `GET chunks`.
- Negative paths covered: missing/non-integer `split_point`, out-of-range offset, offset that yields an empty half, ids not in the document, QA-parser documents.

### Type of change

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -370,18 +370,33 @@ async def split_chunk(tenant_id, dataset_id, document_id, chunk_id):
         return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
     dataset_tenant_id = _get_dataset_tenant_id(dataset_id)
     if not dataset_tenant_id:
+        logging.warning(
+            f"split_chunk dataset tenant missing: dataset_id={dataset_id} document_id={document_id} chunk_id={chunk_id}"
+        )
         return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
     doc = DocumentService.query(id=document_id, kb_id=dataset_id)
     if not doc:
+        logging.warning(
+            f"split_chunk document missing: dataset_id={dataset_id} document_id={document_id} chunk_id={chunk_id}"
+        )
         return get_error_data_result(message=f"You don't own the document {document_id}.")
     doc = doc[0]
     if doc.parser_id == ParserType.QA:
+        logging.warning(
+            f"split_chunk rejected for QA document: dataset_id={dataset_id} document_id={document_id} chunk_id={chunk_id}"
+        )
         return get_error_data_result(message="Splitting is not supported for Q&A documents.")
     req = await get_request_json()
     if not req or "split_point" not in req:
+        logging.warning(
+            f"split_chunk missing split_point: dataset_id={dataset_id} document_id={document_id} chunk_id={chunk_id}"
+        )
         return get_error_data_result(message="`split_point` is required")
     split_point = req.get("split_point")
     if not isinstance(split_point, int) or isinstance(split_point, bool):
+        logging.warning(
+            f"split_chunk invalid split_point type: dataset_id={dataset_id} document_id={document_id} chunk_id={chunk_id} split_point={split_point!r}"
+        )
         return get_error_data_result(message="`split_point` must be an integer character offset")
 
     index = search.index_name(dataset_tenant_id)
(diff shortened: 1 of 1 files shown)


davidv399:
Thanks @coderabbitai. Addressed all three findings in 9ba93f0a6:

**1. Atomic / compensating write (line 444-454, Major)**
The flow is now: insert the two halves → `try` { delete source, then `increment_chunk_num`, then `decrement_chunk_num` (with an inner try that reverts the increment if the decrement fails) }. On any exception in that block, we log the error with `dataset_id`/`document_id`/`chunk_id`/`split_point`/`new_ids`, and best-effort `docStoreConn.delete` the two newly inserted halves so the store never ends up with both the source and the halves visible, and counters never drift. The rollback delete itself is also wrapped in a try/except so a failing rollback is logged but does not mask the original error.

**2. NotFoundError on `docStoreConn.get` (line 379-382, Major)**
Wrapped the `docStoreConn.get` call in a try/except that treats `NotFoundError` as a normal missing-chunk path (returns the same data error as the None case) and falls through to `server_error_response` for anything else. Matches the existing pattern in `get_chunk`.

**3. Structured logs (line 358-473, Major)**
Added `logging` import and structured log lines on:
- INFO at function entry (`dataset_id`, `document_id`, `chunk_id`)
- WARN on each early-return branch: ownership rejected, source missing, invalid offset, empty half, id collision, duplicate halves
- ERROR on post-insert failure and on rollback-delete failure (with `new_ids` and the underlying error)
- INFO on success (`new_ids`, `deleted`)
Style matches the existing f-string `key=value` logging used in `task_api.py` and `dataset_api.py`.

Build + syntax: `python3 -c "import ast; ast.parse(...)"` passes.

JinHai-CN:
@jack-stormentswe   Since previous PR includes a binary file which is very big. So, I reset the head of main branch. Would you please re-submit the commit from current HEAD of main branch? Thank you very much.

davidv399:
Hi @JinHai-CN, thanks for the heads up. Rebased onto the current HEAD of main and force-pushed (0dea57df1). No binary files, only the two commits that touch `api/apps/restful_apis/chunk_api.py`. Ready for re-review.

davidv399:
Thanks @coderabbitai. Addressed all four findings in 8c6254f24:

**1. Restore source chunk when rollback runs after a successful delete (Critical)**
Added `inserted_new_chunks` / `source_deleted` flags around the write sequence. The outer `except` now (a) deletes the just-inserted halves only if they were inserted, and (b) re-inserts the captured `source` record if the source delete had already succeeded. A transient counter failure can no longer permanently drop the source chunk. Both rollback steps wrap their own errors in try/except + ERROR log so a failing rollback never masks the original error.

**2. Reject image-backed chunks (Major)**
Added an early guard right after fetching `source`: if `source.get("img_id")` is set or `source.get("doc_type_kwd") == "image"`, log a WARN and return `get_error_data_result("Splitting image chunks is not supported.")` before computing `left_content`/`right_content` or touching the store.

**3. Preserve `available_int` on the cloned halves (Major)**
When building each new chunk dict, copy `source["available_int"]` onto `d` if present. A split no longer silently re-enables a chunk that was disabled via `switch_chunks`.

**4. Log remaining early-return branches (Nitpick)**
Added WARN logs (with `dataset_id`/`document_id`/`chunk_id` and the reason) to the five previously-silent paths: `dataset_tenant_id` miss, missing document, QA rejection, missing `split_point`, invalid `split_point` type. `split_point!r` is included on the type branch so the bad value lands in the log.

Syntax check passes.

KevinHuSh:
Appreciations!

Please append some instructions about the API beneath it.

wangq8:
key issue: some chunks might be having image, e.g. pdf will generate image, which is not considered here, split the image might be hard. And the method used here are stale, e.g. get_model_config_by_type_and_name / get_model_config_by_id

PR #15054, opened 2026-05-20
Fix discord async issue

wangq8:

RuntimeError: Cannot run the event loop while another loop is running

### Type of change


Lynn-Inf (approved):
LGTM

PR #15068, opened 2026-05-20
feat(agent): add Pipeline chunker component for pre-chunking workflows (#14773)

### LINKED ISSUE #14773: [Feature Request] Enable Pipeline capabilities inside Agent - pre-chunking without vectorization for Agent processing
## Feature Description

Currently, Pipeline and Agent are separate modules in RAGFlow. Users who want to leverage Pipeline features (e.g., chunking strategies, document preprocessing, classification) within an Agent workflow must manage these as two distinct steps outside of RAGFlow.

It would be valuable to expose Pipeline capabilities directly inside the Agent module, so that when a file is uploaded through an Agent, it can be processed through Pipeline-style chunking and preprocessing strategies — without performing vectorization — and then fed into Agent tools/modules for further handling.

## Expected Behavior

- When a file is uploaded in an Agent session, users can apply Pipeline chunking strategies (e.g., "by title", "by section", "recursive", etc.)
- Chunking occurs without vectorization at this stage — purely structural document processing
- The processed chunks are then available to Agent modules for:
  - Tool invocation (e.g., lookup, summarization, extraction)
  - Context injection
  - Structured output generation
- This bridges Pipeline preprocessing and Agent execution in a single flow

## Use Case

- Users want to upload a document, pre-chunk it using advanced strategies, and then have the Agent reason over those chunks without full vectorization overhead
- Combining structured document parsing (Pipeline) with LLM-based reasoning (Agent) in one workflow
- Scenarios where only a subset of chunks need to be vectorized or where chunking strategy should be decoupled from embedding

## Additional Context

This would make Agent much more powerful for document-heavy workflows, reducing the need to run separate Pipeline jobs before Agent interactions. It also gives users fine-grained control over how documents are split before Agent processing, rather than being locked into a single chunking method.

Khaostica:

Closes #14773.

Today, Pipeline (`rag/flow/`) chunking strategies only run as part of a dataset ingestion that always embeds and indexes the result. There is no way to drive Pipeline-style chunking from an Agent workflow without paying that vectorization/persistence cost.

This PR adds a single new Agent component, `PipelineChunker`, that:

- Takes one or more file references (from `Begin` / `UserFillUp` uploads) as input.
- Runs the existing `rag.app.*` chunking strategies (`naive`, `paper`, `qa`, `manual`, `book`, `presentation`, `laws`, `table`, `one`, `email`, `picture`, `audio`, `resume`, `tag`) against each file.
- Emits the resulting chunks as `chunks: list[str]` and `chunks_full: list[dict]` for downstream Agent nodes.
- Performs **no embedding and no persistence** — chunks live only in canvas variables for the duration of the run, exactly as requested in the issue.

The component is auto-discovered by `agent/component/__init__.py`; no registry edits required. Chunker functions are imported lazily so the component itself does not pull `deepdoc` / OCR / VLM at component-discovery time. File resolution mirrors the existing `ExcelProcessor` convention.

Out of scope for this PR (potential follow-ups):

- Vectorization / KB persistence (explicit ask in the issue).
- Frontend canvas UI for the new component.
- Bridging to the newer Pydantic-based `rag/flow/chunker/TokenChunker` (consumes a parser node's structured output rather than a raw file — a separate, larger feature).

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

---

## Files changed

- `agent/component/pipeline_chunker.py` — new component (~180 lines)
- `test/unit_test/agent/test_pipeline_chunker.py` — unit tests (~120 lines)

## Test plan

- [x] `ruff check` on changed files — clean.
- [x] `ruff format` applied to the new component file.
- [x] `python -m py_compile` on both new files — both compile.
- [x] New unit test file carries `pytestmark = pytest.mark.p2` so it runs under marker-filtered CI.
- [x] Every new function, method, and class has a docstring (CodeRabbit 80% docstring-coverage gate).
- [x] `python -m pytest test/unit_test/agent/test_pipeline_chunker.py -x -q` — **7 passed in 1.95s** locally. Tests stub `api.db.services.file_service` and `rag.app.*` so they exercise the parameter validation and parser-id lookup table without requiring the full backend / model stack.

## Manual integration plan (post-merge)

1. Drop the component into an Agent canvas after a `Begin` node with a file input.
2. Set `parser_id = "naive"` (or any other strategy) and reference the file input in `inputs`.
3. Wire the `chunks` output into a downstream `LLM` / `Message` / `Iteration` node — chunks are available as plain text without any embedding or KB write.


Khaostica:
@JinHai-CN This PR is ready for review.

Khaostica:
@JinHai-CN could you review this PR, please?

Khaostica:
@KevinHuSh, @wangq8, Could you review this PR? Appreciate your feedback.

yuzhichang (commented):
I re-reviewed the current head locally. The earlier review points I was watching for are already addressed in the branch now: `from_page <= to_page` validation is present, `FileService.get_blob(...)` failures are isolated per file, the summary no longer exposes raw exception text, and the test module restores its `sys.modules` stubs in teardown.

From my side I do not see a remaining blocking issue in the current diff. This looks ready to merge.

PR #15070, opened 2026-05-20
Fix IMAP From headers with multiple addresses

### LINKED ISSUE #14964: IMAP connector _parse_singular_addr raises RuntimeError on multi-address From headers, crashing entire sync
# IMAP connector `_parse_singular_addr` raises RuntimeError on multi-address From headers, crashing entire sync

## Summary

The `_parse_singular_addr` function in `ragflow/common/data_source/imap_connector.py` raises a hard `RuntimeError` when an email From header contains more than one address. This crashes the entire IMAP sync task, even if only a single email in a large batch has a multi-address From field.

## Impact

- **Complete sync failure** — the entire `dispatch_tasks` run fails, meaning no new emails are indexed
- **No retry** — the task is marked as failed and won't be automatically retried (unless the version is bumped in the marketplace scenario, or a new sync is manually triggered)
- **Silent data loss** — hundreds of already-successfully-processed emails are discarded because the sync crashes before committing

## Root Cause

In `imap_connector.py`, `_parse_singular_addr` is used to extract the sender name + address from the `From` header:

```python
def _parse_singular_addr(raw_header: str) -> tuple[str, str]:
    addrs = _parse_addrs(raw_header=raw_header)
    if not addrs:
        return ("Unknown", "unknown@example.com")
    elif len(addrs) >= 2:
        raise RuntimeError(
            f"Expected a singular address, but instead got multiple; {raw_header=} {addrs=}"
        )

    return addrs[0]
```

When the From header legitimately contains multiple addresses (e.g. mailing lists, group senders, `bcc`-style headers that accumulate in forwarded mail), `_parse_addrs` (which delegates to `getaddresses`) correctly returns multiple tuples. Instead of handling this gracefully, the code raises `RuntimeError`, which propagates up through the async generator chain in `sync_data_source.py`, cancelling the entire task.

## Reproduction Steps

1. Configure an IMAP connector to an inbox containing an email with a From header like:

   ```
   From: "User A" <a@example.com>, "User B" <b@example.com>
   ```

2. Trigger a sync (reindex or incremental)
3. The sync task fails. The error message in `sync_logs.error_msg` reads:

   ```
   Expected a singular address, but instead got multiple; raw_header='User A <a@example.com>, User B <b@example.com>' addrs=[('User A', 'a@example.com'), ('User B', 'b@example.com')]
   ```

## Expected Behavior

The function should handle multi-address From headers gracefully:

- **Best:** Log a warning and return the first address (commonly the primary sender)
- **Acceptable:** Log a warning and skip the single email, continuing with the rest of the batch
- **Not acceptable:** Crash the entire sync task

## Suggested Fix

Replace the `raise RuntimeError` with a warning log and return the first address:

```python
def _parse_singular_addr(raw_header: str) -> tuple[str, str]:
    addrs = _parse_addrs(raw_header=raw_header)
    if not addrs:
        return ("Unknown", "unknown@example.com")
    elif len(addrs) >= 2:
        logging.warning(
            "Multiple addresses in From header, using first: %s", addrs
        )
    return addrs[0]
```

## Environment

- RAGFlow version: current `main` branch (and all prior releases)
- IMAP connector source: IMAP
- Python: 3.12

## Related

- `_parse_addrs()` currently delegates to `getaddresses()` from `email.utils`, which correctly handles RFC 5322 addresses including quoted display names with commas — this part is correct.

victortran0904:

Fixes #14964.

IMAP sync currently raises from _parse_singular_addr when a From header parses to multiple addresses. That exception can fail the whole sync for a single message.

### Type of change

- Logs a warning when a From header has multiple parsed addresses
- Uses the first parsed address as the sender instead of raising
- Keeps the existing unknown-sender fallback when no address parses
- Adds focused unit coverage for empty and multi-address From headers

### Check List

- [x] Self-test passed

Verification run:

- pytest -q test/unit_test/data_source/test_imap_connector.py
- ruff check common/data_source/imap_connector.py test/unit_test/data_source/test_imap_connector.py
- ruff format --check test/unit_test/data_source/test_imap_connector.py
- git diff --check

Note: the targeted pytest was run through the local available pytest path because the full uv test environment was blocked by dependency resolution/fetch issues unrelated to this parser change.

### DIFF
--- a/common/data_source/imap_connector.py
+++ b/common/data_source/imap_connector.py
@@ -40,6 +40,8 @@ _PAGE_SIZE = 100
 _USERNAME_KEY = "imap_username"
 _PASSWORD_KEY = "imap_password"
 
+logger = logging.getLogger(__name__)
+
 class Header(str, Enum):
     SUBJECT_HEADER = "subject"
     FROM_HEADER = "from"
@@ -752,10 +754,9 @@ def _parse_singular_addr(raw_header: str) -> tuple[str, str]:
     if not addrs:
         return ("Unknown", "unknown@example.com")
     elif len(addrs) >= 2:
-        logging.warning(
-            "Expected a singular address, but instead got multiple; raw_header=%r addrs=%r",
-            raw_header,
-            addrs,
+        logger.warning(
+            "Expected a singular address, but parsed %d addresses; using first parsed address",
+            len(addrs),
         )
 
     return addrs[0]
--- a/common/data_source/imap_connector.py
+++ b/common/data_source/imap_connector.py
@@ -752,8 +752,10 @@ def _parse_singular_addr(raw_header: str) -> tuple[str, str]:
     if not addrs:
         return ("Unknown", "unknown@example.com")
     elif len(addrs) >= 2:
-        raise RuntimeError(
-            f"Expected a singular address, but instead got multiple; {raw_header=} {addrs=}"
+        logging.warning(
+            "Expected a singular address, but instead got multiple; raw_header=%r addrs=%r",
+            raw_header,
+            addrs,
         )
 
     return addrs[0]
--- a/test/unit_test/data_source/test_imap_connector.py
+++ b/test/unit_test/data_source/test_imap_connector.py
@@ -14,25 +14,39 @@
 #  limitations under the License.
 #
 
-import logging
+import importlib
 import sys
 import types
 
+import pytest
 from pydantic import BaseModel
 
-sys.modules.setdefault("anthropic", types.SimpleNamespace(BaseModel=BaseModel))
-from common.data_source.imap_connector import _parse_singular_addr
+
+@pytest.fixture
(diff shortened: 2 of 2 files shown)


victortran0904:
Updated to address the Copilot comments:

- switched the new warning to a module logger
- removed raw header and parsed address values from the warning so names/emails are not logged
- moved the anthropic import stub into a pytest monkeypatch fixture instead of mutating sys.modules at module import time
- added assertions that the warning does not include the sample names or email addresses

Verification:
- pytest -q test/unit_test/data_source/test_imap_connector.py
- ruff check common/data_source/imap_connector.py test/unit_test/data_source/test_imap_connector.py
- ruff format --check test/unit_test/data_source/test_imap_connector.py
- git diff --check

victortran0904:
Closing this as duplicate/obsolete: #15006 has already been merged and closes #14964 with the same underlying fix. Thanks.

PR #15071, opened 2026-05-20
feat(web/switch): allow dot-notation field paths in Switch condition (#14235)

### LINKED ISSUE #14235: [Agent Scenario Request]:
### Self Checks

### Is your feature request related to a scenario?

```Markdown
The switch component should be basically consistent with the code component
```

If my previous component returned an object, when using swtich for judgment, I would only need to judge a certain field value in the object. However, now in the switch component, I can only select this object, but cannot select or input to implement a certain field in the object, and I must also set the judgment to be equal to or other, which cannot meet this requirement.
If the input parameters of the switch component are the same as those of the code component, even if an object is passed in, I can use the code to determine the value of a certain field in the object. So it is recommended that the parameters of the switch component be passed in the same way as those of the code component, except that the result can only return true or false.
Another suggestion, whether it is reasonable or feasible:
Although the selection of variable passing is now done through interface operations, it is not very suitable to set parameters only for a certain field in the object as input. It is recommended to allow manual input of ". XXX. XXX" after selecting variables for other components, which is more convenient for parameter setting. This is for reference only, because even if not done this way, the entire object can be passed directly, just an extra line of code.

```Markdown

```

### Additional information

_No response_

Khaostica:

The Switch agent component only let users select an upstream variable as a whole — there was no way to drill into a nested field of an object output (see #14235). If an upstream node returned `{category, score, meta: {lang}}`, the condition row could only compare against the entire object.

**Where the gap was:** the canvas resolver in `agent/canvas.py` (`Graph.get_variable_value` / `get_variable_param_value`) already understood the `cpn_id@root.nested.path` syntax — it splits on `@`, then on the first `.`, and walks the rest through dicts/lists/attrs. The Switch backend at `agent/component/switch.py:80` passes `item["cpn_id"]` through unchanged. So the entire gap was in the frontend: `QueryVariable` rendered a strict dropdown with no way for users to type a path suffix.

**Approach:** keep the dropdown for picking the base variable (`cpn@root`), add an optional `.field.subfield` text input next to it. Both controls write to the same `cpn_id` form field, joined as `cpn@root.field.subfield`, which the existing backend resolver consumes unchanged. New `splitBaseAndPath` / `joinBaseAndPath` helpers handle the parse/format; a new local `VariableWithPath` component composes the two controls — kept inside `web/src/pages/agent/form/switch-form/index.tsx` so no shared UI under `src/components/ui/` or `src/components/originui/` is touched (per `web/CLAUDE.md`). The new placeholder uses an inline `defaultValue` fallback on the i18n key `flow.fieldPathPlaceholder` so no locale files are modified.

**Out of scope:** the issue also suggested mirroring the Code component's combobox-style input. That would require modifying shared UI; the path-suffix Input chosen here is the smallest change that delivers the user-visible behavior. The combobox refactor can be a separate PR.

**Tests:** 7 new unit tests in `test/testcases/test_web_api/test_canvas_app/test_switch_component_unit.py` exercise `Switch._invoke` end-to-end against a canvas whose `get_variable_value` follows the same parsing rules as the real `Graph`. Cases covered: single-level nested string `=` literal, doubly-nested string, nested number with `>` (numeric coercion), nested string with `contains`, fall-through to ELSE, list-index path (`output.chunks.0.text`), and OR short-circuit on a nested match. All 7 pass locally. The test file uses module-scope `pytestmark = pytest.mark.p2` and has docstrings on every helper and test.

**Files changed:** 2 files, +486 / −10.

**Test plan:**

- [x] `python -m pytest test/.../test_switch_component_unit.py --noconftest` — 7 passed
- [x] `python -m ruff check` / `ruff format --check` / `py_compile` on the new Python — clean
- [ ] `npm run lint` / `npx tsc --noEmit` on the frontend — not run locally (no `web/node_modules` in this environment); CI to verify
- [ ] Manual browser verification — reviewer should run `cd web && npm install && npm run dev`, build a small flow with a `Begin` node producing an object output, drop a `Switch` after it, confirm the condition row shows a dropdown + a path Input, type `.someField` into the Input, and verify the comparison fires against the nested value at runtime

Closes #14235

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


Khaostica:
@JinHai-CN, could you please review my PR? I appreciate your feedback.

Khaostica:
@asiroliu, @wangq8, this PR is ready for review.

dcc123456 (approved):

PR #15073, opened 2026-05-20
Go: implement Embed (embeddings) in Replicate driver

sxxtony:

`ReplicateModel.Embed` in `internal/entity/models/replicate.go` was a `"replicate, no such method"` stub. Tracking issue #14736 lists Replicate's embedding surface as not implemented. This PR wires it up against Replicate's documented embedding schema.

Until this PR, a tenant who selected a Replicate embedding model got the sentinel error on every embed call.

### What this PR includes

- `internal/entity/models/replicate.go` — real `Embed` method backed by Replicate's `/v1/predictions` async API, reusing the existing `createPrediction` + `waitForPrediction` plumbing the chat path already validates.
- `conf/models/replicate.json` — adds `replicate/all-mpnet-base-v2` at version `b6b7585c...` (`model_types: ["embedding"]`).

### Wire shape (verified against Replicate's openapi_schema)

Input (per the canonical `replicate/all-mpnet-base-v2` schema):
- `text` (string) — single text to encode, used when `len(texts) == 1`
- `text_batch` (string) — JSON-encoded list of strings to encode, used when `len(texts) > 1`. Note that the schema declares `text_batch` as `type: string`, so the JSON-encoded list is sent **as a string**, not as a JSON array.

Output: array of `{embedding: [number]}` objects, one per input in order. The driver also tolerates the alternative bare-nested-array shape `[[number]]` some community models emit.

### Live test transcripts (api.replicate.com)

The Replicate account on the provided key has no billing credit, so I could not run a full embedding inference end-to-end. Everything **up to** the billing gate works:

**T1 — Auth check**
```
GET /v1/account
→ HTTP 200
→ username: tmimmanuel
```

**T2 — Model metadata for replicate/all-mpnet-base-v2**
```
GET /v1/models/replicate/all-mpnet-base-v2
→ HTTP 200
→ owner/name:  replicate/all-mpnet-base-v2
→ visibility:  public
→ version id:  b6b7585c9640cd7a...
→ Input properties:
    - text         (type=string)  "A single string to encode."
    - text_batch   (type=string)  "A JSON-formatted list of strings to encode."
→ Output:
    type: array
    items: $ref=#/components/schemas/Embedding
    Embedding shape: {embedding: array<number>} (required)
```
Confirms the input field names and output shape the driver targets.

**T3 — Driver's request reaches the prediction endpoint (route ✓, payload ✓, version ✓)**
```
POST /v1/predictions   (Prefer: wait=60)
body: {"version":"b6b7585c…","input":{"text":"hello world"}}
→ HTTP 402  "Insufficient credit"
→ "You have insufficient credit to run this model.…"
```
Replicate accepts the request — URL, headers, version, and `input.text` field shape all pass validation. Only the billing gate stops execution.

**T4 — Invalid version (control, proves Replicate validates the version field)**
```
POST /v1/predictions
body: {"version":"deadbeef…","input":{"text":"x"}}
→ HTTP 422  "Invalid version or not permitted"
→ "The specified version does not exist (or perhaps you don't have permission to use it?)"
```
This rules out a false-positive on T3: Replicate does validate the version up-front. T3's 402 means the version is real and the request shape is correct.

**T5 — Bare owner/name path on a community model (background)**
```
POST /v1/models/replicate/all-mpnet-base-v2/predictions
→ HTTP 404
```
Confirms that `replicate/all-mpnet-base-v2` is not on Replicate's official-models list, so the catalog uses the `owner/name:version` notation which the existing `predictionEndpoint` in `replicate.go` routes through `/v1/predictions` with the `version` field — the format that returned 402 in T3.

### Why owner/name:version in the catalog

`replicate.go:111` already implements:
```go
func replicateUsesVersionEndpoint(modelName string) bool {
    name := strings.TrimSpace(modelName)
    return !strings.Contains(name, "/") || strings.Contains(name, ":")
}
```
Names containing `":"` go through `/v1/predictions` with `version`. Community embedding models (which is most of them on Replicate) need this form because they aren't on the official-models list and 404 on the owner/name endpoint (T5). The pinned version in the catalog can be overridden per-tenant if a newer version ships.

### What I deliberately did NOT do

- **No unit tests** — per user direction. The chat path already exercises `createPrediction` + `waitForPrediction` in production, and the new `Embed` reuses those same code paths; the only new logic is input/output shaping (`replicateEmbedInput`, `replicateEmbedOutputToVectors`).
- **No full end-to-end embedding inference** — the test account hit Replicate's billing gate (HTTP 402) on every prediction. The 402/422 differential above is the closest verification available without funding the account.
- **No support for the `nateraw/bge-large-en-v1.5` schema** — that model uses `input.texts` (different field name) and returns either inline arrays or a `.npy` file URI. Implementing that alternative input/output shape belongs in a follow-up; the canonical `text`/`text_batch` + `[{embedding:[…]}]` schema (replicate/all-mpnet-base-v2) is implemented here.

### DIFF
--- a/conf/models/replicate.json
+++ b/conf/models/replicate.json
@@ -22,6 +22,13 @@
       "model_types": [
         "chat"
       ]
+    },
+    {
+      "name": "replicate/all-mpnet-base-v2:b6b7585c9640cd7a9572c6e129c9549d79c9c31f0d3fdce7baac7c67ca38f305",
+      "max_tokens": 384,
+      "model_types": [
+        "embedding"
+      ]
     }
   ]
 }
--- a/internal/entity/models/replicate.go
+++ b/internal/entity/models/replicate.go
@@ -566,8 +566,154 @@ func (r *ReplicateModel) CheckConnection(apiConfig *APIConfig) error {
 	return err
 }
 
+// replicateEmbedInput shapes the request body for Replicate's standard
+// embedding models (e.g. replicate/all-mpnet-base-v2). Per the
+// canonical Replicate embedding schema published in the model's
+// openapi_schema, the two input fields are:
+//
+//	text       — single string to encode (used when len(texts) == 1)
+//	text_batch — JSON-formatted list of strings (used when len > 1)
+//
+// `text_batch` is `type: string` in the schema, so the JSON-encoded
+// list itself is sent as a string value, NOT as a JSON array. Models
+// that use different field names (e.g. nateraw/bge-large-en-v1.5's
+// `texts`) are not currently supported by this driver; tenants on
+// those should consult Replicate's OpenAPI schema and configure a
+// compatible model in conf/models/replicate.json.
+func replicateEmbedInput(texts []string) (map[string]interface{}, error) {
+	switch len(texts) {
+	case 0:
+		return nil, fmt.Errorf("replicate: texts is empty")
+	case 1:
+		return map[string]interface{}{"text": texts[0]}, nil
+	default:
+		encoded, err := json.Marshal(texts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to encode text_batch: %w", err)
+		}
+		return map[string]interface{}{"text_batch": string(encoded)}, nil
+	}
+}
+
+// replicateEmbedOutputToVectors normalizes Replicate's two observed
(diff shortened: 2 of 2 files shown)


Haruko386 (approved):
@sxxtony LGTM

Issue #15074, opened 2026-05-20
image2text model_type conflates OCR/CV and multimodal chat models


Qwerrty574:
## Problem

model_type="image2text" in RAGFlow is used for two fundamentally different types of models:

1. OCR/CV models — only "image → text." Without chat context, delta streaming, or tool-calling. CvModel class.
2. Multimodal chat models — full-fledged chat models with vision input. Full /chat/completions API: context, system prompts, reasoning, tool-calling,
streaming.

RAGFlow processes both categories through a single path: CvModel. This causes:

- Runtime errors: CvModel.async_chat_streamly returns resp.usage.total_tokens (cv_model.py:170), but providers return usage=None in the final chunk — AttributeError: 'NoneType' object has no attribute
'total_tokens'
- Loss of functionality: CvModel lacks async_chat_streamly_delta, reasoning, and tool-calling. Multimodal models passed through CvModel lack all of these.
- Validation fails: _validate_llm_id in chat_api.py and openai_api.py with a request for model_type="chat" doesn't find models saved as model_type="image2text" — "llm_id doesn't exist." However, multimodal models are valid for use in chat.

Because of this, models with image support cannot be used in chat mode.

When saving models through Model Providers, even if you select the Chat type, if you check the "Does it support Vision?" flag, the model is saved as an image2text model and, when edited, displays as an OCR model in the same interface.

## Reproduction

  1. Add a multimodal model (e.g., `gpt-4o`) via OpenAI-API-Compatible provider
  2. Model is stored with `model_type="image2text"` in `tenant_llm` table
  3. Set as chat LLM in Chat Settings → validation fails (before fix 1)
  4. If set directly in DB or via API → chat responses lose reasoning/tool-calling and may crash on `total_tokens` (before fix 2)

## Long-term Fix Needed

Separate `model_type` into distinct values for OCR/CV vs multimodal chat, or introduce a capability flag (e.g., `is_chat_capable`) on the `image2text` type. The `is_tools` flag in `llm_factories.json` already partially distinguishes them but isn't used for routing decisions.

PR #15075, opened 2026-05-20
Fix: allow multimodal models as chat LLMs and route through ChatModel

Qwerrty574:

 Models stored with model_type="image2text" (via OpenAI-API-Compatible providers) cannot be set as chat LLMs — Chat Settings returns "llm_id doesn't exist". When forced, they route
  through CvModel, losing reasoning, tool-calling, and crashing on total_tokens.

model_type="image2text" conflates two categories — pure OCR/CV models and multimodal chat LLMs. Three code paths treated all image2text models as "OCR-only":

  1. _validate_llm_id() in chat_api.py/openai_api.py queried with model_type="chat" only, rejecting image2text models at validation
  2. dialog_service.py routed image2text → TenantLLMService.get_model_config(..., IMAGE2TEXT) → CvModel
  3. tenant_model_service.py found image2text models via CHAT↔IMAGE2TEXT fallback but left config_dict["model_type"]="image2text", so model_instance() still created CvModel

  The CHAT↔IMAGE2TEXT fallback already existed in tenant_model_service.py, but only for runtime config lookup — not for API validation or model instance creation. This PR extends the pattern
   consistently across all three paths.

### Changes
- Validation fallback: _validate_llm_id() tries requested model_type first, then compatible type. Only chat↔image2text are compatible — other types unaffected 
- Model type normalization: When IMAGE2TEXT model is found as CHAT fallback, config_dict["model_type"] set to "chat" → model_instance() creates ChatModel
- Consistent config source: async_chat() uses get_model_config_by_type_and_name() instead of TenantLLMService.get_model_config(). llm_type via llm_id2llm_type() unchanged to preserve image-file routing

### Limitations
- No way to distinguish a multimodal chat model from a pure OCR/CV model at the model_type level. If a pure vision-only model were ever registered as image2text and assigned to a chat dialog, it would  incorrectly instantiate ChatModel and likely fail at runtime.
- llm_type in async_chat() still reflects the raw DB type ("image2text"). Image-file handling (images=image_files) triggers on the else branch, passing images= to a ChatModel that ignores them via _clean_param(). It works, but the signal is semantically wrong.
- is_tools flag exists in llm_factories.json and db_models.py but is not used for routing. It already correctly marks multimodal chat models (is_tools=true) vs pure OCR (is_tools absent/false) and could drive the decision.

### Type of change

### Related issue
#15074

### DIFF
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -178,15 +178,20 @@ async def _validate_llm_id(llm_id, tenant_id, llm_setting=None):
     if model_type not in {"chat", "image2text"}:
         model_type = "chat"
 
-    if not await thread_pool_exec(
-        TenantLLMService.query,
-        tenant_id=tenant_id,
-        llm_name=llm_name,
-        llm_factory=llm_factory,
-        model_type=model_type,
-    ):
-        return f"`llm_id` {llm_id} doesn't exist"
-    return None
+    # Try the requested model_type first, then fall back to the compatible type.
+    # image2text (multimodal) models are valid as chat models per tenant_model_service.
+    for mt in ((model_type, "image2text") if model_type == "chat"
+               else ("image2text", model_type) if model_type == "image2text"
+               else (model_type,)):
+        if await thread_pool_exec(
+            TenantLLMService.query,
+            tenant_id=tenant_id,
+            llm_name=llm_name,
+            llm_factory=llm_factory,
+            model_type=mt,
+        ):
+            return None
+    return f"`llm_id` {llm_id} doesn't exist"
 
 
 async def _validate_rerank_id(rerank_id, tenant_id):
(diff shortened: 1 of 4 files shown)


api/apps/restful_apis/chat_api.py
```diff
@@ -178,15 +178,20 @@ async def _validate_llm_id(llm_id, tenant_id, llm_setting=None):
...
-        tenant_id=tenant_id,
-        llm_name=llm_name,
-        llm_factory=llm_factory,
-        model_type=model_type,
-    ):
-        return f"`llm_id` {llm_id} doesn't exist"
-    return None
+    # Try the requested model_type first, then fall back to the compatible type.
+    # image2text (multimodal) models are valid as chat models per tenant_model_service.
+    for mt in ((model_type, "image2text") if model_type == "chat"
```
KevinHuSh: Is the `if`/`else` right? I don't get it.

Qwerrty574: No, but it's already been fixed, I've pushed it.

Qwerrty574: This was initially a temporary solution, intended to address the issue so that multimodal models could be used in chat, but the problem still exists: undefined behavior caused by ambiguous model type definitions (#15074). 
A long-term solution to improve the unambiguity of model type definitions needs to be discussed, as breaking changes are required.

Lynn-Inf (commented):
This is already supported in v0.27.0 — we'll close this PR. Thanks!

PR #15079, opened 2026-05-20
security: get_uuid() uses cryptographically secure tokens (CVE-2025-69286)

axivityFred:
## Summary
`common/misc_utils.py:get_uuid()` mints access tokens, OAuth `state` values, and session IDs via `uuid.uuid1().hex`, which derives from the host MAC + current timestamp — both observable or guessable. The resulting tokens are predictable to an attacker who knows the host and an approximate request time. Affected call sites include OAuth state, session IDs, and short-lived access tokens.

## Fix
Swap `uuid.uuid1().hex` for `secrets.token_hex(16)`, which returns 32 lowercase hex chars backed by 128 bits of cryptographically secure entropy (equivalent randomness budget to UUID v4). The return type stays `str`. The `import uuid` becomes unused inside `common/misc_utils.py` and is dropped.

```python
import secrets

def get_uuid():
    return secrets.token_hex(16)
```

## Schema compatibility
Drop-in: returns 32 lowercase hex chars, exactly matching `uuid.uuid1().hex` width — no schema migration required for the ~100 `CharField(max_length=32)` ID columns in `api/db/db_models.py` (User, Tenant, ApiToken, Dialog, Document …).

## Testing
- Manual: `print(get_uuid())` before/after — no collisions, no MAC leak.
- Unit test: `test/unit_test/common/test_misc_utils.py` — 1000 calls produce 1000 distinct values, each exactly 32 hex chars.
- The existing `TestGetUuid` class in that file has 8 tests; after this change, 7 still pass (length=32, hex chars, lowercase, no dashes, uniqueness, string type, length consistency). Only `test_uuid1_specific_characteristics` becomes obsolete (it asserts UUID v1 internal bit layout) and should be removed.

## References
- CVE-2025-69286 (uuid1 timestamp+MAC predictability)
- Ships today in an Axivity downstream overlay as `003-uuid-tokens.patch` until merged upstream.

### DIFF
--- a/common/misc_utils.py
+++ b/common/misc_utils.py
@@ -23,13 +23,13 @@ import os
 import subprocess
 import sys
 import threading
-import uuid
+import secrets
 
 from concurrent.futures import ThreadPoolExecutor
 
 
 def get_uuid():
-    return uuid.uuid1().hex
+    return secrets.token_hex(16)
 
 
 async def download_img(url):


KevinHuSh:
Appreciations!

CI failure.
```
_______________ TestGetUuid.test_uuid1_specific_characteristics ________________
test/unit_test/common/test_misc_utils.py:72: in test_uuid1_specific_characteristics
    ???
E   AssertionError: assert None == 1
E    +  where None = UUID('cfe1abec-fb08-5c31-665d-838d7db092e0').version
```

wangq8:
Resovled in this PR: #19159 
Hence close it.

Issue #14461, opened 2026-04-29
[Feature Request]: code component can use the Document Generation Component's output file info as input


liulx:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown
when I want to the Document Generation Component's output file as input of the code component, I can't get the file info ,include file_name, doc_id
```

### Describe the feature you'd like

code component can use the Document Generation Component's output file info as input
<img width="1843" height="702" alt="Image" src="https://github.com/user-attachments/assets/b9a78620-7daf-42ab-b6e4-1bb443f8dd17" />

### Documentation, adoption, use case

```Markdown

```

PR #15080, opened 2026-05-21
Feat: Expose Doc Generator file metadata as discrete outputs

nickmopen:
Declare doc_id, filename, mime_type, and size as separate outputs on the Document Generation component so downstream nodes (e.g., the Code component) can consume them via the variable picker. The existing download JSON blob is preserved unchanged for the Message component's download-chip rendering.

  The Document Generation component previously exposed only a single `download` output —
  a JSON-encoded blob containing the file's `doc_id`, `filename`, `mime_type`, `size`,
  and base64 payload. On top of that, the variable picker actively hides this `download`
  entry from every consumer except the Message component (because the embedded base64 is
  too heavy to splat into arbitrary downstream nodes).

  The combined effect: users wiring the Doc Generator's output into a Code component had
  no way to retrieve basic file info such as `file_name` or `doc_id` from the picker,
  blocking workflows that need to post-process the generated file (e.g., registering it
  elsewhere, custom delivery, follow-up API calls).

  This PR declares `doc_id`, `filename`, `mime_type`, and `size` as **discrete outputs**
  on the Document Generation component, alongside the existing `download` blob. The new
  fields:

  - Appear in the variable picker for **all** downstream nodes, including the Code
  component, so users can bind them directly to script arguments.
  - Are cheap scalars only — no base64 payload leaks into other components.
  - Leave the existing `download` JSON blob completely untouched, so the Message
  component's download-chip rendering (which parses that blob via `_is_download_info`)
  keeps working with no behavior change.

  Changes:
  - `agent/component/docs_generator.py` — declare the four new outputs in
  `DocGeneratorParam` and emit them via `set_output(...)` in `_invoke`.
  - `web/src/pages/agent/constant/index.tsx` — extend `initialDocGeneratorValues.outputs`
   with the new keys.
  - `web/src/pages/agent/form/doc-generator-form/index.tsx` — mirror the new outputs in
  the zod schema so the form is valid.

  No changes needed to the picker's existing `download`-hiding filter — it matches only
  on the literal output name `download`, so the new metadata entries fall through
  naturally.

  Reported in: https://github.com/infiniflow/ragflow/issues/14461.
  ### Type of change

### DIFF
--- a/agent/component/docs_generator.py
+++ b/agent/component/docs_generator.py
@@ -52,6 +52,10 @@ class DocGeneratorParam(ComponentParamBase):
         self.include_download_info_in_content = False
         self.font_size = 12
         self.outputs = {
+            "doc_id": {"value": "", "type": "string"},
+            "filename": {"value": "", "type": "string"},
+            "mime_type": {"value": "", "type": "string"},
+            "size": {"value": 0, "type": "number"},
             "download": {"value": "", "type": "string"},
         }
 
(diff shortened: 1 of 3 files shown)


nickmopen:
@dosu

nickmopen:
@KevinHuSh Can you review this PR?

PR #15100, opened 2026-05-21
Fix: <asyncio.locks.Semaphore object at 0xabcd [locked]> is bound to a different event loop

wangq8:
Fix: <asyncio.locks.Semaphore object at 0xabcd [locked]> is bound to a different event loop


wangq8:
@coderabbitai

PR #15124, opened 2026-05-21
fix(dialog): enforce read-only SQL allowlist in chat use_sql before sql_retrieval

pekar9781:
## Related issues

Closes #15123

Chat structured-data queries use `use_sql()` in `api/db/services/dialog_service.py` to turn natural-language questions into SQL, then execute that SQL against the document store. On **Infinity** deployments, execution goes through `InfinityConnection.sql()` → `psql -c <sql>` with no validation beyond light normalization and (for ES/OceanBase) `kb_id` UUID checks.

LLM-generated SQL is attacker-influenced (prompt injection or model mistakes). Without a read-only guard, destructive statements such as `DROP TABLE`, `DELETE`, `TRUNCATE`, or stacked queries (`SELECT 1; DROP ...`) can run against tenant KB tables (`ragflow_{tenant_id}_{kb_id}`), causing **data loss**.

This is a **separate path** from the agent `ExeSQL` tool hardened in PR #14877. That fix does not cover chat `use_sql()`.

This PR adds shared, sqlglot-based read-only validation and applies it to `use_sql()` **before** `sql_retrieval()` for all doc engines (`infinity`, `oceanbase`, `es`).

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Changes

| File | Change |
|------|--------|
| `common/sql_guard.py` | New module: `ensure_read_only_sql()`, `ReadOnlySqlViolation`; single-statement allowlist (`SELECT` / `UNION` / `EXCEPT` / `INTERSECT` only); rejects DML/DDL and query stacking |
| `api/db/services/dialog_service.py` | Call `ensure_read_only_sql()` in `get_table()` before `sql_retrieval()`; fail closed on `ReadOnlySqlViolation` (no retry) |
| `test/unit_test/common/test_sql_guard.py` | Unit tests for allowed selects and rejected destructive SQL |
| `pyproject.toml` | Add explicit `sqlglot` dependency |

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,7 +34,7 @@ The project uses **uv** for dependency management.
 
 1. **Setup Environment**:
    ```bash
-   uv sync --python 3.12 --all-extras
+   uv sync --python 3.13 --all-extras
    uv run python3 download_deps.py
    ```
 
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ docker build --platform linux/amd64 \
    ```bash
    git clone https://github.com/infiniflow/ragflow.git
    cd ragflow/
-   uv sync --python 3.12 # install RAGFlow dependent python modules
+   uv sync --python 3.13 # install RAGFlow dependent python modules
    uv run python3 download_deps.py
    pre-commit install
    ```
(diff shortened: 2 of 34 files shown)


pekar9781:
Hi @Haruko386
If you have time, could you please check my PR? Thanks

qinling0210 (changes requested):
@pekar9781 Could you modify go side? We expect we can modify both python and go sides if the issue exists on both sides in a PR. Thanks.

PR #15140, opened 2026-05-22
Fix: ExeSQL node continues on per-statement SQL errors

### LINKED ISSUE #14737: [Bug]: ExeSQL node no exception default info, and out
### Self Checks

ce4c782fd

### RAGFlow image version

0.25.1

```Markdown

```

### Actual behavior

  Display friendly information, when the node exception

### Expected behavior

('42S02', "[42S02] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]对象名 'ASSET_AUDIT' 无效。 (208) (SQLExecDirectW)")

### Steps to reproduce

```Markdown
<img width="2528" height="1153" alt="Image" src="https://github.com/user-attachments/assets/b1073ae2-dd8d-410c-a3c8-d232534e3009" />
```

### Additional information

Image

nickmopen:
Wrap per-statement execution in both the generic and IBM DB2 loops so a failing statement reports a friendly "SQL Execution Failed" message and continues, instead of letting a raw driver exception abort the node and discard results from statements that already succeeded.

Rolls back after a failure so PostgreSQL's aborted-transaction state does not cascade into every subsequent statement in the batch.

Closes #14737

The **ExeSQL** agent node splits its input on `;` and runs each statement in a loop. Both execution loops — the generic one (`cursor.execute`) and the IBM DB2 one (`ibm_db.exec_immediate`) — were wrapped only in a `try/finally` for resource cleanup, with **no `except`** around statement execution.

As a result, when any single statement failed (e.g. the reporter's MSSQL `('42S02', "[42S02] ... 对象名 'ASSET_AUDIT' 无效")`):
- The raw, unformatted driver exception bubbled up and the node failed with an ugly `_ERROR` instead of friendly information.
- **The whole node aborted** — results from statements that had already succeeded were discarded, and the remaining statements in the batch never ran. The reporter confirmed this was the real pain point: *"after reporting an exception, the previous normal query cannot be executed properly … Do not interrupt the workflow for any issues."*

Connection-level failures were already wrapped with a friendly `"Database Connection Failed!"` prefix — only per-statement execution errors were missed.

**This PR** wraps per-statement execution in `try/except` in both loops. A failing statement now:
- records a friendly `SQL Execution Failed: <sql>\n<error>` entry into the `json` and `formalized_content` outputs (the actual DB error is kept so the user can see *what* failed), and
- `continue`s to the next statement — so earlier results survive and later statements still run.

After a failure in the generic loop, the connection is rolled back so PostgreSQL's aborted-transaction state does not cascade into every subsequent statement in the batch. The node returns normally (no `_ERROR` raised), so the agent workflow proceeds instead of halting.

Connection failures remain fatal (correct — nothing can run without a connection). The pre-existing `break` on `cursor.rowcount == 0` is intentionally left unchanged; it is out of scope for this fix.

### Type of change

### DIFF
--- a/agent/tools/exesql.py
+++ b/agent/tools/exesql.py
@@ -208,28 +208,37 @@ class ExeSQL(ToolBase, ABC):
                         continue
                     single_sql = re.sub(r"\[ID:[0-9]+\]", "", single_sql)
 
-                    stmt = ibm_db.exec_immediate(conn, single_sql)
-                    rows = []
-                    row = ibm_db.fetch_assoc(stmt)
-                    while row and len(rows) < self._param.max_records:
-                        if self.check_if_canceled("ExeSQL processing"):
-                            return
-                        rows.append(row)
+                    try:
+                        stmt = ibm_db.exec_immediate(conn, single_sql)
+                        rows = []
                         row = ibm_db.fetch_assoc(stmt)
+                        while row and len(rows) < self._param.max_records:
+                            if self.check_if_canceled("ExeSQL processing"):
+                                return
+                            rows.append(row)
+                            row = ibm_db.fetch_assoc(stmt)
 
-                    if not rows:
-                        sql_res.append({"content": "No record in the database!"})
+                        if not rows:
+                            sql_res.append({"content": "No record in the database!"})
+                            continue
+
+                        df = pd.DataFrame(rows)
+                        for col in df.columns:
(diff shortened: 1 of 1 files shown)


nickmopen:
@dosu

nickmopen:
@KevinHuSh , Could you review this PR as well?

PR #15141, opened 2026-05-22
Fix: add model_type into llm_setting

Lynn-Inf:

Add model_type into llm_setting

### Type of change


web/src/utils/llm-util.ts
```diff
@@ -62,6 +62,53 @@ export function parseModelUuid(uuid: string): {
   return { modelName, factoryId };
 }
 
+// Get model_type from cached LLM list by llm_id (e.g., "modelName@factoryId")
+export function getModelTypeByLlmId(
```
cike8899: You can use useFindLlmByUuid instead of getModelTypeByLlmId

cike8899 (approved):

PR #15158, opened 2026-05-22
Go: implement provider: PaddleOCR_Local

Haruko386:

Go: implement provider: PaddleOCR_Local

**Verified from CLI**

```
RAGFlow(user)> ocr with 'PaddleOCR-VL@test@paddleocr_local' file './internal/test1.jpg'
+----------------------+
| text                 |
+----------------------+
| ## Parallel to these |
+----------------------+
```

### Type of change

- [X] New Feature (non-breaking change which adds functionality)
- [X] Refactoring

### DIFF
--- a/conf/models/gitee.json
+++ b/conf/models/gitee.json
@@ -8,7 +8,7 @@
     "models": "models",
     "status": "",
     "balance": "tokens/packages/balance",
-    "embedding": "embedding",
+    "embedding": "embeddings",
     "rerank": "rerank",
     "ocr": "images/ocr",
     "doc_parse": "async/documents/parse",
--- a/conf/models/mineru_local.json
+++ b/conf/models/mineru_local.json
@@ -1,5 +1,5 @@
 {
-  "name": "mineru_local",
+  "name": "MinerU_local",
   "url_suffix": {
     "doc_parse": "file_parse",
     "task": "tasks"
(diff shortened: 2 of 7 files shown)


zeweihan:
Nice work on the PaddleOCR_Local provider — having a local OCR option that doesn't require external API calls is a big deal for document-heavy workflows where data stays on-prem.

A few thoughts from our experience building OCR pipelines for legal document processing:

1. **Layout-aware OCR matters more than raw character accuracy** — for legal docs, the spatial relationship between text blocks (clauses, numbered lists, table cells) is as important as the characters themselves. PaddleOCR's layout analysis is strong here, but make sure the bounding box output preserves reading order, not just spatial position. Legal docs often have multi-column layouts with footnotes that break naive top-to-bottom ordering.

2. **Confidence scores per-region, not just per-character** — we've found it useful to expose OCR confidence at the block/paragraph level, not just individual characters. This lets downstream RAG pipelines decide whether to trust the OCR result or flag for human review. A paragraph with 95% confidence is trustworthy; one at 70% needs a second look.

3. **Mixed-language support** — legal documents frequently mix languages (e.g., English contracts with Latin legal terms, or Chinese contracts with English clauses). PaddleOCR handles CJK well, but the provider should surface the detected language per-region so the RAG pipeline can apply the right tokenizer/embedding model.

The CLI verification approach is solid — having a way to quickly test OCR output against known documents is invaluable for regression testing when upgrading the OCR engine version.

PR #15161, opened 2026-05-22
feat: accept inline file attachments in OpenAI-compatible chat API

### LINKED ISSUE #5637: [Feature Request]: OpenAI-Compatible API attachment files
### Is there an existing issue for the same feature request?

```Markdown
We want to manage chat sessions by ourselves, and OpenAI-Compatible API is quite useful in this case.
But it doesn't support attachment files.
```

add a field called **files** in messages
```json
{
    "messages": [{"role":"user","content":"hello","files":[{"blob":"...","display_name":"a.txt"}]}]
}
```

_No response_

```Markdown

```

### Additional information

_No response_

mhtkarakose:

Closes #5637.

Callers that manage their own chat sessions use the OpenAI-compatible endpoint (`POST /api/v1/openai/{chat_id}/chat/completions`), but it had no way to attach an ad-hoc file to a message.

The endpoint already forwarded a message's `files` field into the chat pipeline (`async_chat` → `split_file_attachments` → `FileService.get_files`), but only as **references to already-stored files** (`{id, created_by, mime_type, name}`). There was no way to pass file *content* in the request.

This PR adds inline-attachment support, exactly as requested in the issue:

```json
{"messages": [{"role": "user", "content": "hello",
               "files": [{"blob": "<base64>", "display_name": "a.txt"}]}]}
```

`_resolve_message_files` (in `openai_api.py`) decodes each inline entry (tolerating a `data:<mime>;base64,` prefix), persists it via `FileService.put_blob`, and rewrites it into the stored-file form the existing pipeline already parses. Reference-style entries are passed through unchanged, so this is fully backward compatible. The change is localized to the OpenAI endpoint — the shared chat pipeline is untouched.

### Type of change

- [x] Documentation Update

### Test plan

- New unit test `test/unit_test/api/apps/restful_apis/test_openai_message_files.py` (8 cases): base64 decoding, `data:` URI prefix, MIME inference from extension, reference pass-through, and error paths (missing `display_name`, invalid base64, malformed entries). All pass.
- `ruff check` clean.
- `docs/references/http_api_reference.md` updated to document the per-message `files` field.


mhtkarakose:
Thanks for the reviews — addressed in `c5ae6dd2c` (which also merges current `main`, resolving the conflict with #14380).

**Fixed:**

- **Argument error code** — invalid `files` input now returns `code=RetCode.ARGUMENT_ERROR` so clients can distinguish bad input from other failures.
- **Inline attachment size cap** — added `MAX_INLINE_FILE_BYTES` (20 MB). Oversized payloads are rejected by a cheap base64-length pre-check *and* a post-decode byte-length check, before `put_blob`.
- **Malformed `data:` URI** — a `data:` prefix with no comma separator (e.g. `data:image/png;base64`) now raises `ValueError` instead of silently storing an empty blob.
- **Non-list `files`** — a `files` value that is present but not an array now returns an argument error instead of being silently ignored.
- **Logging** — `_resolve_message_files` now emits an info log (user, file_id, name, mime, size) when an inline attachment is persisted.

Two new unit tests cover the malformed-data-URI and oversized-blob paths (10 tests total, all passing).

**Not changed, with reasons:**

- *Last-message role gating* — `_resolve_message_files` only runs after the handler's existing `if messages[-1]["role"] != "user": return ...` guard, so `messages[-1]` is already guaranteed to be the user message; no extra gating needed.
- *`parents[5]` in the test* — kept for consistency with the sibling test in the same directory (`test_get_agent_session.py`), which uses the identical repo-root resolution. Worth changing repo-wide as a separate cleanup rather than diverging one file.

mhtkarakose:
Addressed the nitpick in the latest commit — `_resolve_message_files` now rejects an `id`-style reference that is missing any of `created_by`/`mime_type`/`name` with a clear `ValueError`, before downstream code can hit it. New test covers the rejection path (11 tests, all passing).

Yannnnnnny:
The API does not validate that `files[].blob` is a string. Sending a number, boolean, array, or object causes an internal error instead of returning `RetCode.ARGUMENT_ERROR`.

Please reject non-string `blob` values with a clear validation response so clients receive consistent API behavior.

Lynn-Inf:
### Blocking

  1. **Needs a rebase, and the change needs to be re-placed.** This PR is based on `main` from May, back when the import was still `get_model_config_from_provider_instance`. Current `main` has renamed it to `resolve_model_config` and added `_normalize_openai_messages`, `extra_body`, and `_generation_params`, so the import hunk will conflict. After rebasing, please double-check where `_resolve_message_files` sits relative to `_normalize_openai_messages` (which rebuilds each message as `{**message, "content": ...}`) and make sure the *resolved* entries are what reach `async_chat`.

  2. **Inline files only write a blob — no `File` record is created.** `FileService.put_blob()` only writes an object into the `{user_id}-downloads` bucket (`api/db/services/file_service.py:739`); it does not insert a file row. As a result these attachments cannot be listed or cleaned up with the user/session, and a partial failure leaves orphaned objects behind. Please reuse the `FileService` upload path that creates a record, or — if a temporary blob is intentional — document its lifecycle and add a cleanup path.

### Should fix

  3. **Putting `files` inside standard `messages[]` breaks the OpenAI compatibility of this route.** `main` already consolidates RAGFlow-specific extensions (`reference`, `metadata_condition`) under `extra_body`. I'd suggest `files` goes there too; if you keep it on the message, please state explicitly in the docs that this is a RAGFlow-proprietary field and not part of the OpenAI schema.

  4. **There is a 20 MB cap per file, but no cap on the request as a whole.** The API server does not set `MAX_CONTENT_LENGTH` (only `admin/server/admin_server.py:58` does), so a request with N × 20 MB can exhaust memory. Please also bound the number of files and the total byte size.

  5. **The MIME type is fully client-controlled and never validated against content.** Resolution order is data-URI prefix → `mimetypes.guess_type(display_name)`, both of which the caller controls. Downstream, `FileService.get_files` classifies images with `image in mime_type` and `split_file_attachments` classifies anything starting with `data:` as an image, so a name that disagrees with the real content routes to the wrong branch. At minimum, please allow-list the accepted types.

### Nit

  6. `base64.b64decode(validate=True)` rejects base64 containing newlines (very common with MIME wrapping). Consider stripping whitespace before decoding.
  7. Error codes are inconsistent with the rest of the file: other argument checks use `get_error_data_result(msg)`, while this path passes `code=RetCode.ARGUMENT_ERROR`.
  8. `FileService.put_blob` can raise non-`ValueError` exceptions (storage errors, etc.), which currently surface as a 500. Please widen the `except`.
  9. The tests load the module via `importlib` with a `parents[5]` path assumption and a large set of monkeypatched stubs, which is fragile, and they only cover the helper — there is no route-level test. Prefer a plain import that only patches `FileService.put_blob`, plus an end-to-end case asserting that `files` actually reaches `async_chat`.
  10. The `web/src/locales/zh.ts` change is just Prettier re-wrapping two OneDrive strings; it is unrelated to this PR and should be dropped.

  Two notes on the translation:
  - Item 1's parenthetical is reworded to be accurate — normalization preserves files either way, so the real point is that the resolved form must reach async_chat. The Chinese version said "顺序放反会丢字段", which overstates it.
  - Line references point at current main; they may shift slightly against the PR's base.

Lynn-Inf (commented):
@mhtkarakose Please address the review comments and resolve the merge conflicts. Thanks!

Issue #5965, opened 2025-03-12
[Feature Request]: User permission system


trcyberoptic:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

I would like to have an option to grant users access without being them able to modify anything, so they can only use assistants and models created by an "administrator".

### Describe implementation you've considered

A two stage permission model would be sufficient. 
- Administrator
- User

### Documentation, adoption, use case

```Markdown

```

mhtkarakose:
Opened #15164 implementing the backend of this feature — a global two-tier permission model.

- **Administrator** accounts have full access.
- **User** accounts are read-only: they can use assistants, models, and knowledge bases created by an administrator, but cannot create, modify, or delete any resource.

It adds a global `account_role` on each user (defaulting to `admin`, so existing installs are unaffected), an `@require_admin_account` guard on every mutation endpoint, and an endpoint for a team owner to set a member's role.

Two follow-up PRs will add the role-assignment UI in team settings and hide the create/edit/delete controls in the web UI for read-only accounts.

Yannnnnnny:
Hi, this feature has been implemented in RAGFlow enterprise version

PR #15164, opened 2026-05-22
feat: global two-tier user permission system (backend)

mhtkarakose:

Implements the backend of #5965 — a **two-tier user permission model**. Organizations want to grant people access to RAGFlow assistants without letting them modify shared configuration.

This adds a global per-account tier:

- **Administrator** (`account_role = "admin"`) — full access.
- **User** (`account_role = "user"`) — read-only: may *use* assistants/models/knowledge bases but cannot create, modify, or delete any resource.

**Design notes**

- RAGFlow is multi-tenant (every user owns a personal tenant), so a per-tenant role cannot express a globally read-only account. The permission is therefore a global field on `User`.
- `User.account_role` is added with an idempotent migration and **defaults to `"admin"`**, so every existing account keeps full access — no behavior change on upgrade.
- `@require_admin_account` (`api/common/permission.py`) is layered after `@login_required`; it returns `RetCode.PERMISSION_ERROR` for read-only accounts and is sync/async aware.
- Every create/update/delete/config mutation across the dataset, document, chunk, chat, agent, file, and llm apps is gated. `GET`, search, retrieval, and chat/agent completion stay open so read-only accounts can still use shared assistants.
- `PUT /tenants/<tenant_id>/users/<user_id>/account-role` lets a team owner (an admin account) set a member's role.

This is the first of three PRs (backend → role-assignment UI → frontend control gating).

### Type of change

### Test plan

- New unit tests in `test/unit_test/api/common/test_permission.py` (4 cases): admin passes through, read-only account blocked with code 108, sync-handler support, `is_readonly_account` incl. the legacy default. All pass; `ruff check` clean.
- Manual: a `user`-tier account receives 403 on mutation endpoints and 200 on read/chat endpoints; an `admin` account is unaffected.


mhtkarakose:
Thanks for the thorough review — addressed in `9c0b8034b`.

**Fixed:**

- **Decorator awaiting** — `require_admin_account` now calls the wrapped function and awaits the result via `inspect.isawaitable`, instead of branching on `inspect.iscoroutinefunction`. It is now correct at any position in a decorator stack.
- **Fail closed** — `is_readonly_account` now grants write access only for an explicit `admin` role; any other stored value (`user`, an unexpected string, `None`) is treated as read-only. A genuinely missing attribute still defaults to admin for pre-feature back-compat.
- **Missed mutation endpoints gated** — `PUT /datasets/<id>/metadata/config`, `POST /agents/rerun`, and `POST` / `DELETE /tenants/<id>/users` (team invite/remove) now require an admin account.
- **Pending invites** — `set_account_role` rejects members whose `UserTenant` row is still `INVITE`.
- **`@token_required` routes** — dropped the gate from the two SDK chunk routes. `token_required` does not populate `current_user`, so the check could fail open. SDK-token permission scoping is a separate concern; deferred to a follow-up rather than shipped half-working.
- New unit tests cover the sync-callable-returning-awaitable path and the fail-closed `account_role` values (5 tests, all passing).

**Not changed, with reasons:**

- *`current_user.id == tenant_id` check in `tenant_api.py`* — kept for consistency with the existing `user_list` / `create` / `rm` routes in the same file, which all use this exact pattern (a user's personal tenant id equals their user id in RAGFlow).
- *Single-step migration* — `migrate_db()` adds every column as `NOT NULL` with a default via `alter_db_add_column`; a multi-step migration for this one column would diverge from the established pattern across the whole file.

mhtkarakose:
Addressed the audit-logging nitpick in `149a10f42` — `set_account_role` now emits an info log (target user, new role, acting principal, tenant) on a successful role change.

Issue #14902, opened 2026-05-13
[Bug]: SSRF and missing ownership check on POST /api/v1/mcp/servers/<mcp_id>/test — arbitrary outbound HTTP/SSE with user-controlled headers


oxtensor:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

5e46457c2

### RAGFlow image version

5e46457c2

### Other environment information

```Markdown
Hardware parameters: any
OS type: Linux (Docker compose deployment)
Others: Any deployment where the RAGFlow API container has outbound network access to internal addresses (default in docker-compose and most Kubernetes setups).
```

### Actual behavior

`POST /api/v1/mcp/servers/<mcp_id>/test` (`api/apps/restful_apis/mcp_api.py:293–331`) builds an `MCPServer` purely from caller-supplied `url`, `server_type`, `headers`, and `variables`, and immediately opens a session against that URL via `MCPToolCallSession(mcp_server, ...).get_tools(...)`.

Two distinct flaws:

1. **SSRF (CWE-918).** The `url` body parameter is never validated — no DNS resolution / private-IP block, no scheme allowlist, no host allowlist, no length cap. Inspection of `common/mcp_tool_call_conn.py` (lines 39–232) confirms there is no URL/host filter anywhere along the call path. A logged-in user can:
   - scan internal RFC1918 / `127.0.0.1` ranges (timing oracle),
   - hit cloud metadata services (`169.254.169.254`, `metadata.google.internal`),
   - probe internal services such as `ragflow-mysql:3306`, `ragflow-redis:6379`, `localhost:9200`,
   - exfiltrate request bodies by setting attacker-controlled `headers` (e.g. `Authorization: Bearer <stolen-token>`) pointed at an external host.

[...]

Issue #14865, opened 2026-05-12
[Bug]: SSRF via unvalidated avatar URL fetch during OAuth registration leaks internal service data


thomascolden585-svg:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

e04a2a01e

### RAGFlow image version

e04a2a01e

### Other environment information

```Markdown
- All environments with OAuth login enabled
- Especially critical on cloud deployments (AWS, GCP, Azure) where instance metadata endpoints are accessible
```

### Actual behavior

During OAuth callback registration, the server fetches the user's avatar from a URL provided by the OAuth provider (`api/apps/restful_apis/user_api.py`, line 223):

```python
avatar = await download_img(user_info.avatar_url)
```

The `download_img` function in `common/misc_utils.py` (lines 35-42) makes an **unvalidated** HTTP request to the supplied URL and returns the full response body as a base64-encoded data URI:

```python
async def download_img(url):
    if not url:
        return ""
    from common.http_client import async_request
    response = await async_request("GET", url)
    return "data:" + \
        response.headers.get('Content-Type', 'image/jpg') + ";" + \
        "base64," + base64.b64encode(response.content).decode("utf-8")
```

There is **no validation** on the URL — no check against private/internal IP ranges, no hostname resolution verification, and no response size limit.

If an attacker controls the OAuth user info (via a malicious OAuth provider, by manipulating the OAuth response through a MITM, or by registering a custom OAuth provider if the system allows it), they can supply URLs targeting internal services:

| Malicious URL | Data leaked |
|---|---|
| `http://169.254.169.254/latest/meta-data/iam/security-credentials/` | AWS IAM credentials |
| `http://metadata.google.internal/computeMetadata/v1/` | GCP instance metadata |
| `http://localhost:6379/` | Redis service data |
| `http://localhost:3306/` | MySQL connection response |
| `http://internal-service:8080/admin/config` | Internal admin endpoints |

[...]

Issue #14858, opened 2026-05-12
[Bug]: SRF via `/agents/test_db_connection` — no host/port validation allows internal network scanning


enjoyandlove:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

48febf70e

### RAGFlow image version

48febf70e

### Other environment information

```Markdown
- Hardware parameters: N/A (applies to all deployments)
- OS type: Linux (all platforms affected)
- Others: Applies to any RAGFlow deployment exposing the REST API
```

### Actual behavior

The `POST /api/v1/agents/test_db_connection` endpoint in `api/apps/restful_apis/agent_api.py` (lines 719–827) accepts `host` and `port` parameters directly from user input and initiates TCP database connections to those targets **without any SSRF validation**.

Any authenticated user can force the backend server to open TCP connections to arbitrary internal hosts and ports. The endpoint supports multiple database drivers (MySQL, MariaDB, OceanBase, PostgreSQL, MSSQL, IBM DB2, Trino), each of which will attempt a real TCP connection to the attacker-specified destination.

This is a **Server-Side Request Forgery (SSRF)** vulnerability that enables:

1. **Internal service scanning**: An attacker sends `host=10.0.0.1`, iterating over ports (`6379` for Redis, `3306` for MySQL, `9200` for Elasticsearch, etc.) to map the internal network. Connection success/failure responses reveal which services are running.
2. **Access to RAGFlow's own databases**: The attacker can connect to RAGFlow's internal MySQL/Redis using the backend server's network position, potentially reading or modifying data if default or weak credentials are used.
3. **Cloud metadata exfiltration**: On cloud deployments (AWS/GCP/Azure), `host=169.254.169.254` can probe the instance metadata service to leak IAM credentials, instance identity, and other sensitive data.
4. **Denial of service**: Flooding internal services with connection attempts from the backend.

[...]

Issue #14088, opened 2026-04-13
[Bug]: Server-Side Request Forgery (SSRF) in URL-based Document Upload


xingxing21:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

N/A

### RAGFlow image version

v0.24.0

### Other environment information

```Markdown

```

### Actual behavior

The `POST /upload_info?url=<url>` endpoint accepts a URL from the query string and passes it directly to `AsyncWebCrawler` without any validation. There is no check on:

- Protocol — `file://`, `gopher://`, `ftp://` etc. are accepted alongside `http/https`
- Destination IP — private ranges (10.x, 172.16–31.x, 192.168.x), loopback (`127.0.0.1`), and cloud metadata endpoints (`169.254.169.254`) are not blocked
- Redirects — the crawler may follow redirects that land on internal addresses even if the initial URL looks external
This allows any authenticated user to instruct the server to make arbitrary outbound connections to internal infrastructure.

### Expected behavior

The server should validate the supplied URL before initiating any network request:

- Only http and https schemes are accepted.
- The resolved IP address must not fall within RFC 1918, loopback, or link-local ranges.
- Redirects are capped and the final resolved address is re-validated before content is returned.
- Invalid or disallowed URLs return a 400 Bad Request with a descriptive error message.

### Steps to reproduce

```Markdown
1. Log in as any authenticated user.
2. Send the following request, targeting the AWS EC2 instance metadata service:

POST /upload_info?url=http://169.254.169.254/latest/meta-data/iam/security-credentials/

3. Observe that the server makes an outbound HTTP request to 169.254.169.254 and returns the response content (IAM role credentials) as document content.
4. Repeat with an internal host, e.g.:

POST /upload_info?url=http://192.168.1.1/admin

to demonstrate internal network scanning capability.
```

### Additional information

Location:

Entry point: `api/apps/document_app.py, line 874` — url read directly from request.args
[...]

PR #15172, opened 2026-05-23
Fix: SSRF in agent Browser component upload-URL fetch

### LINKED ISSUE #15171: [Bug]: SSRF in agent Browser component upload-URL fetch (_prepare_upload_url_file)
### Self Checks

e6dd39753

### RAGFlow image version

v0.25.5 / nightly @ `main` (commit `e6dd39753`)

- Hardware parameters: any
- OS type: any (Linux verified)
- Others: reproduces in any deployment because the bug is in the Python agent component, not in a hardware- or OS-specific path

### Actual behavior

The Browser agent component (`agent/component/browser.py`) lets a workflow author add
HTTP(S) URLs to `upload_sources`. At runtime, `_prepare_upload_files` recognises each
HTTP(S) entry and hands it to `_prepare_upload_url_file(url, upload_dir)`, which fetches
the URL with `urlopen(...)` and writes the response body to a file that is later attached
to the browser session.

The URL is validated only for scheme (`http`/`https`) by `Browser._is_http_url`. There is
**no destination-IP validation, no redirect re-validation, and no protection against
cloud metadata endpoints**, so any user who can edit the agent (a normal authenticated
user, since agents are per-tenant) can instruct the RAGFlow server to issue authenticated
outbound requests to:

- Cloud instance-metadata services: `http://169.254.169.254/latest/meta-data/iam/security-credentials/` (AWS), `http://169.254.169.254/metadata/instance?api-version=...` (Azure), `http://metadata.google.internal/...` (GCP)
- Loopback / private services: `http://127.0.0.1:9200/_cluster/health` (the in-cluster Elasticsearch), `http://127.0.0.1:5455/api/v1/...` (the RAGFlow API itself, bypassing auth boundary), `http://mysql:3306`, `http://valkey:6379`
- Arbitrary RFC 1918 hosts on the internal network (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`)
- Hosts reached by a 3xx redirect from an external URL the attacker controls (since `urlopen` follows redirects and the destination is not re-validated)

The fetched response body is silently written to a file and attached to the browser
session, where downstream LLM/browser tooling reads it — so the request body is also
exfiltrated, not just the network-level effect.

### Expected behavior

`_prepare_upload_url_file` should validate the destination before opening the URL and
again on every redirect:

- Only `http` and `https` schemes are accepted (already enforced).
- The hostname is DNS-resolved and **every** returned address (A and AAAA) must be a
  public unicast address — reject `is_private`, `is_loopback`, `is_link_local`,
  `is_multicast`, `is_reserved`, `is_unspecified`. This blocks RFC 1918, `127.0.0.0/8`,
  the link-local `169.254.0.0/16` range (which contains the AWS metadata endpoint), and
  the IPv6 equivalents (`fc00::/7`, `::1`, `fe80::/10`).
- 3xx redirects are followed only when the redirect target also passes the same check;
  otherwise the request errors out instead of silently following.
- Rejected URLs return `None` (the same failure path the function already uses for
  HTTP/network errors) and log a warning naming the URL — they must not be fetched.

### Steps to reproduce

1. Run a RAGFlow deployment that has the Browser agent component enabled.
2. As any authenticated user, create or edit an agent that uses the **Browser** node,
   and add the following entry to the node's `upload_sources` parameter (which is just a
   list of strings the agent definition accepts):
   ```
   http://169.254.169.254/latest/meta-data/iam/security-credentials/
   ```
   On a self-hosted box you can substitute `http://127.0.0.1:9200/_cluster/health` for
   the Elasticsearch container that ships with `docker-compose-base.yml`, which is
   reachable from the RAGFlow container with no credentials.
3. Trigger a run of the agent. The Browser node calls `_prepare_upload_files` →
   `_prepare_upload_url_file` → `urlopen(req, timeout=30)`.
4. Observe in the RAGFlow server logs that the request to the internal/metadata address
   succeeds (a `Browser upload url returned empty content` / size log line for the URL,
   and the upload working directory contains a file with the fetched body). The browser
   subagent receives the response body as an attachment.
5. To demonstrate the redirect bypass, host any URL on a public webserver that responds
   with `HTTP/1.1 302 Found\r\nLocation: http://127.0.0.1:9200/_cluster/health\r\n\r\n`
   and supply that public URL instead. `urlopen` follows the redirect without
   re-validation and fetches the internal-only response.

### Additional information

**Location:**

- Sink: `agent/component/browser.py` line 220 — `with urlopen(req, timeout=30) as response:` inside `_prepare_upload_url_file`.
- Entry point: `agent/component/browser.py` line 531 — `_prepare_upload_url_file(file_ref, upload_dir)` where `file_ref` is an unvalidated string from the user-controlled `upload_sources` parameter of the Browser component (see `_prepare_upload_files`, line 526).
- Scheme gate: `agent/component/browser.py` line 159 — `_is_http_url` only restricts the scheme; it does not look at the host.

**Why the existing check is not enough:** `_is_http_url` returns `True` for any
`http://<host>/...` URL including `http://169.254.169.254/...`,
`http://127.0.0.1:9200/...`, and `http://10.0.0.5/...`. The hostname is never resolved
or compared to a private/loopback/link-local range.

**Why this hits everyone:** the Browser node is enabled by default and the
`upload_sources` parameter is a free-form list of strings — there is no UI-side
allowlist. Any tenant user with permission to edit agents can reach the sink.

**Why existing tests do not catch it:** the only test that exercises
`_prepare_upload_url_file` is via `agent/test/client.py`, which runs end-to-end against a
live `Canvas` and never asserts anything about which destinations the function refuses.
There is no unit test that calls `_prepare_upload_url_file` (or a helper) with a
loopback / metadata URL and asserts it is rejected.

**Comparison with prior SSRF fixes in this repo:** issue #14088 reported the same class
of bug in `api/apps/document_app.py` → `api/db/services/file_service.py` (closed); open
issues #14858, #14865, and #14902 cover three other endpoints with the same root cause.
The Browser-component upload-URL path is **not** covered by any of these. The fix
follows the same shape — validate the destination IP and re-validate on redirect — but
in a different file and a different call site.

**Suggested fix:**

- Add a `Browser._is_safe_destination(url)` static method that, in addition to the
  existing scheme check, DNS-resolves the host and rejects any resolved address that is
  private, loopback, link-local, multicast, reserved, or unspecified (using
  `ipaddress.ip_address(...).is_*` flags, so IPv4 and IPv6 are both covered).
- Add a `_SafeRedirectHandler(HTTPRedirectHandler)` whose `redirect_request` re-runs
  `_is_safe_destination` on the redirect target and raises `URLError` instead of
  following an unsafe redirect.
- In `_prepare_upload_url_file`, reject the URL up front with a warning if
  `_is_safe_destination(url)` is false, and open the request through
  `build_opener(_SafeRedirectHandler())` instead of the module-level `urlopen` so the
  redirect path is also covered.

**Acceptance:**

- New unit tests cover: a public URL passes, `http://127.0.0.1:9200` is rejected,
  `http://169.254.169.254` is rejected, `http://10.0.0.5` is rejected,
  `http://[::1]` is rejected, `ftp://example.com` is rejected, and a redirect to
  `http://127.0.0.1` from a public URL is rejected (the request errors instead of
  silently following).
- Reproduction step 4 above no longer fetches the metadata / Elasticsearch URL; the
  RAGFlow server logs `Browser upload url rejected (non-public destination)` and
  `_prepare_upload_url_file` returns `None`.

joaovictor712:

Closes #15171.

The Browser agent component's `_prepare_upload_url_file` fetches HTTP(S) URLs from
the user-supplied `upload_sources` parameter. The existing `_is_http_url` guard only
restricts the scheme — there is no destination-IP validation and no redirect
re-validation — so any tenant user who can edit an agent can instruct the RAGFlow
server to issue authenticated outbound requests to loopback, RFC 1918 / `fc00::/7`,
and link-local destinations (including the `169.254.169.254` cloud metadata endpoint
on AWS / Azure / GCP), and exfiltrate the response body via the file the function
attaches to the browser session.

This is the same class of bug as #14088 (closed) and the open SSRF reports #14858,
#14865, #14902 — but in a different file and a different call site, so none of those
fixes covered the Browser-component upload path.

**Fix (in `agent/component/browser.py`):**

- Add `_is_safe_url_destination(url)` — DNS-resolves the hostname and rejects any
  resolved address that is `is_private`, `is_loopback`, `is_link_local`,
  `is_multicast`, `is_reserved`, or `is_unspecified`. Both IPv4 and IPv6 are
  covered. Scheme is also gated to http/https.
- Add `_SafeUrlRedirectHandler(HTTPRedirectHandler)` — re-runs the same gate on
  every 3xx redirect target so a public URL cannot silently redirect into an
  internal address.
- `_prepare_upload_url_file` now pre-validates the URL and opens the request through
  `build_opener(_SafeUrlRedirectHandler()).open(...)` instead of the bare module-level
  `urlopen`. Rejected URLs return `None` (the existing failure path) and log a
  warning naming the URL.

**Tests (`test/unit_test/agent/component/test_browser_use_component.py`):**

- Updated `test_prepare_upload_files_supports_http_url` to monkeypatch the new code
  path (`_is_safe_url_destination` + `build_opener`) so the existing happy-path test
  keeps passing.
- Added `test_prepare_upload_url_file_rejects_unsafe_destination` — asserts the
  function returns `None`, never calls `build_opener`, and creates no temp file when
  the gate refuses the URL.
- Added `test_is_safe_url_destination_rejects_internal_and_metadata_addresses` —
  uses monkeypatched `getaddrinfo` so the test is deterministic and network-free;
  covers loopback v4/v6, `169.254.169.254`, all three RFC 1918 ranges, multicast,
  and `0.0.0.0`.
- Added `test_is_safe_url_destination_rejects_non_http_schemes` — `ftp://`, `file://`,
  `gopher://`, empty, and malformed inputs.
- Added `test_is_safe_url_destination_accepts_public_address` — monkeypatched DNS
  to a public IP to assert the happy path returns `True`.
- Added `test_is_safe_url_destination_returns_false_on_dns_failure` — DNS error path.
- Added `test_safe_redirect_handler_refuses_unsafe_target` — proves the redirect
  handler raises `URLError("Refused redirect to non-public destination: …")` when
  the target fails the gate.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


JinHai-CN:
@joaovictor712 would you please resolve the conflicts?

PR #14868, opened 2026-05-12
fix: block SSRF in misc_utils.download_img for OAuth avatars

dale053:
### What problem does this PR solve?

Closes #14865

`download_img` in `common/misc_utils.py` is used for OAuth avatar URLs. The previous implementation called `async_request` from `common.http_client`, which followed redirects without re-validating each hop and did not apply the same SSRF protections as this path needs. That made it possible to reach non-public or disallowed targets (for example via redirects or unsafe URLs) when fetching avatars.

This change replaces that flow with an explicit, bounded fetch: each URL (including every redirect target) is checked with `common.ssrf_guard.assert_url_is_safe`, DNS is pinned with `pin_dns_global`, `httpx` streams the body with `follow_redirects=False` and a manual redirect loop (capped by `RAGFLOW_OAUTH_AVATAR_MAX_REDIRECTS`), and total response size is capped (`RAGFLOW_OAUTH_AVATAR_MAX_BYTES`). Timeouts, proxy, and user agent align with `HTTP_CLIENT_*` env vars without importing `http_client`, so lightweight tests stay simple.

Unit tests cover empty/None URLs, loopback, cloud metadata-style addresses, and disallowed schemes so SSRF regressions are caught early.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

PR #13545, opened 2026-03-11
Feature/generic api connector

ahmadintisar:
# feat: Add Generic REST API Connector

## What problem does this PR solve?

RAGFlow supports many specific data source connectors (MySQL, Slack, Google Drive, etc.), but there was no way to connect an arbitrary REST API as a data source. Users with custom or third-party APIs had to write a new connector class for each one.

This PR adds a **generic, configuration-driven REST API connector** that lets users connect any REST API as a data source entirely through the UI — no code changes needed per API.

---

### Core Connector (`common/data_source/rest_api_connector.py`)

- Implements `LoadConnector` and `PollConnector` interfaces for full and incremental sync
- **Configurable authentication:** None, API Key (custom header), Bearer Token, Basic Auth
- **Pluggable pagination:** Page-based, Offset-based, Cursor-based, or None
- Smart page-size inference from user's query parameters to avoid duplicate/conflicting params
- Configurable request delay between pages to prevent API rate limiting
- Auto-detection of the items array in JSON responses (`items`, `results`, `data`, `records`, or first list found)
- **Advanced field mapping** with dot-notation (`country.name`), array wildcards (`newsType[*].name`), type hints, and default values
- Optional content template rendering (`"Title: {title}\nBody: {body}"`)
- HTML stripping for content fields
- Stable document IDs via `hash128` from a configurable ID field or auto-generated from item content
- Pydantic configuration schema with automatic coercion of UI string inputs to dicts/lists

### Backend Registration (`rag/svr/sync_data_source.py`, `common/constants.py`, `common/data_source/config.py`)

- `REST_API` sync class wired into RAGFlow's `func_factory`
- Full sync (`load_from_state`) and incremental polling (`poll_source`) support
- Credentials and config passed from task to connector following existing patterns (MySQL, SeaFile, etc.)

### Test Connection Endpoint (`api/apps/connector_app.py`)

- `POST /v1/connector/<id>/test` validates config schema, authentication, and API connectivity without triggering a sync
- Clear error messages for auth failures vs. config issues

### Frontend UI (`web/src/pages/user-setting/data-source/constant/`)

- **Postman-style configuration:** Base URL, Query Parameters (key=value per line), Auth, Content Fields, Metadata Fields, Pagination Type
- Auth-type-aware form: fields for API key header/value, Bearer token, or Basic username/password appear only when relevant
- **Advanced Settings** toggle for: Custom Headers, Max Pages, Request Delay, Poll Timestamp Field, Request Body (POST)
[...]

PR #15174, opened 2026-05-23
Fix: DNS-rebinding SSRF in REST API data source connector

### LINKED ISSUE #15173: [Bug]: DNS-rebinding / TOCTOU SSRF in REST API data source connector
### Self Checks

e6dd39753

### RAGFlow image version

v0.25.5 / nightly @ `main` (commit `e6dd39753`)

- Hardware parameters: any
- OS type: any (Linux verified)
- Others: reproduces wherever the REST API connector is reachable. The bug is in
  the Python connector code, not in any hardware- or OS-specific path.

### Actual behavior

The REST API data source connector
(`common/data_source/rest_api_connector.py`, added in PR #13545) ships its
own `_validate_url_for_ssrf(url)` instead of using the project's shared
`common/ssrf_guard.assert_url_is_safe`. The homegrown validator has a
textbook **DNS-rebinding / TOCTOU** bypass:

1. In `RestAPIConnector.__init__` (line 257), `_validate_url_for_ssrf(url)`
   calls `socket.getaddrinfo(hostname, None)` (line 201) **once** and checks
   each resolved address against `is_loopback / is_private / is_link_local /
   is_reserved / is_multicast`.
2. Later, `_fetch_page` (line 588) calls `rl_requests.get(url, ...)` /
   `rl_requests.post(url, ...)`. `requests` / `urllib3` resolves DNS
   **independently** for the actual TCP connection — there is no link
   between the IP that was validated in step 1 and the IP the socket
   eventually connects to in step 2.
3. An authenticated user who can configure a REST API data source can point
   `url` at a hostname they control whose authoritative DNS returns
   a public IP for the first lookup and `127.0.0.1` /
   `169.254.169.254` / RFC 1918 for the second lookup (TTL=0). The
   validation passes. The fetch lands on the internal target. Whatever
   `Authorization`, `X-Api-Key`, or `Basic` credentials the user
   configured under `auth_config` / `credentials` are sent to that
   internal target in the request headers.

The same connector also re-issues the request on every page (`_fetch_page`
is called in a loop in `_iter_items`), giving the attacker repeated
unauthenticated reach into the internal network for each page they pretend
to return — and pages can carry attacker-controlled cursors so the page
URL keeps re-resolving DNS for each iteration.

The project **already has a fix** for this exact attack class — the
shared `common/ssrf_guard` module exposes `assert_url_is_safe(url)` which
returns the validated `(hostname, ip)` and `pin_dns(hostname, ip)` /
`pin_dns_global(hostname, ip)` which pin the hostname to that specific IP
inside a context manager. The other data ingestion path that fetches
arbitrary user URLs — `agent/tools/crawler.py` — uses exactly that
pattern (`assert_url_is_safe` + `pin_dns_global` wrapping the fetch).
`rest_api_connector.py` just doesn't.

### Expected behavior

- The REST API connector reuses `common/ssrf_guard.assert_url_is_safe(url)`
  for URL validation. The shared validator already covers everything the
  homegrown one does *plus* IPv4-mapped IPv6 normalisation
  (`::ffff:127.0.0.1` → `127.0.0.1`) which the homegrown version misses.
- The validated `(hostname, ip)` pair is stored on the connector instance
  and re-used across every `_fetch_page` call.
- Every outbound `rl_requests.get` / `rl_requests.post` happens inside a
  `pin_dns(hostname, ip)` context manager, so the hostname always resolves
  to the validated IP for the duration of the request — closing the
  TOCTOU window. The pin is automatically removed afterwards.
- Subsequent fetches that resolve to a new IP (legitimate DNS change,
  cluster failover, etc.) are re-validated on each call so a stable hostname
  doesn't keep a now-stale validated IP forever.

### Steps to reproduce

1. Set up a hostname under your control with an authoritative DNS server
   that flips its answer per query and serves the next answer with TTL=0:

   ```
   ; round 1: public IP that passes the SSRF check
   rebind.attacker.test.  0  IN  A  93.184.216.34
   ; round 2: loopback that hits the internal target
   rebind.attacker.test.  0  IN  A  127.0.0.1
   ```

2. As any authenticated user with permission to create a data source,
   create a **REST API** data source with:

   - `url = http://rebind.attacker.test/_cluster/health` (the in-cluster
     Elasticsearch admin endpoint that ships with `docker-compose-base.yml`
     and is reachable from the RAGFlow container with no credentials)
   - `auth_type = bearer`, `credentials.token = <your real upstream token>`
     (any value — it'll be exfiltrated)
   - `content_fields = ["status"]`, `pagination_type = none`

3. Click **Test connection** on the data source. This hits
   `RestAPIConnector.validate_config` → `__init__` →
   `_validate_url_for_ssrf` → first DNS lookup returns the public IP →
   validation passes → `_page_iter_for_validation` → `_fetch_page` →
   `rl_requests.get("http://rebind.attacker.test/_cluster/health",
   headers={"Authorization": "Bearer <your token>"})` → second DNS lookup
   returns `127.0.0.1` → the request hits the local Elasticsearch admin
   endpoint with the `Authorization` header attached.

4. Observe in the RAGFlow logs:

   ```
   INFO  Validating REST API connector by fetching first page
   DEBUG REST API request: GET http://rebind.attacker.test/_cluster/health | params={} | headers={'Authorization': '***'}
   ```

   On the attacker's DNS server: two queries from the RAGFlow server. On
   the local Elasticsearch container: an inbound request from
   `127.0.0.1` carrying the `Authorization: Bearer <token>` header
   intended for the attacker's domain. The cluster-health response is then
   returned to the connector and ingested as document content, so the
   attacker can also read the response body via the regular data source
   reads.

5. To prove the TOCTOU window in isolation without controlling DNS:
   monkeypatch `socket.getaddrinfo` to return `93.184.216.34` on the first
   call (validation) and `127.0.0.1` on the second call (fetch). The
   unit test in this PR (`test_rest_api_connector_dns_rebinding_blocked`)
   does exactly that and shows the request would land on `127.0.0.1`
   today; with the fix it errors with the SSRF guard message instead.

### Additional information

**Location:**

- Validator (homegrown): `common/data_source/rest_api_connector.py:173-231`
  (`RestAPIConnector._validate_url_for_ssrf`).
- Pre-existing shared guard that this should use:
  `common/ssrf_guard.py:111-172` (`assert_url_is_safe`) +
  `common/ssrf_guard.py:62-90` (`pin_dns` / `pin_dns_global`).
- Fetch sink: `common/data_source/rest_api_connector.py:588-614`
  (`_fetch_page` → `rl_requests.get` / `rl_requests.post`).
- Reference call site that uses the shared guard correctly:
  `agent/tools/crawler.py:39-57` (`Crawler._run` wraps the fetch in
  `pin_dns_global(hostname, ip)`).

**Why this isn't covered by the existing SSRF reports:**

- #14088 (closed) — different file (`api/db/services/file_service.py`).
- #14858 (open) — `/agents/test_db_connection` host validation.
- #14865 (open) — OAuth avatar fetch.
- #14868 (merged) — `misc_utils.download_img` for OAuth avatars (used
  `ssrf_guard` correctly — the precedent for this fix).
- #14902 (open) — `/api/v1/mcp/servers/<mcp_id>/test`.
- #15171 (open, my prior report) — `agent/component/browser.py`
  `_prepare_upload_url_file`.

None of them touch `rest_api_connector.py` or the data source connector
ingest path, and none of them mention DNS rebinding / TOCTOU specifically.

**Why this hits everyone:** the REST API connector is enabled by default
and is the documented way to attach arbitrary external APIs as data
sources (PR #13545, "Generic, configuration-driven REST API data source
connector"). Any tenant user with permission to create data sources can
reach the sink.

**Why existing tests don't catch it:** the existing unit tests in
`test/unit_test/data_source/test_rest_api_connector.py` mock
`socket.getaddrinfo` to return a single fixed public IP for **both** the
validation and the fetch lookups, so the validator and the requests
client agree by construction. There is no test that returns different IPs
for the two lookups (which is what DNS rebinding does).

**Suggested fix (matches the PR I'm opening alongside this issue):**

- Replace the body of `RestAPIConnector._validate_url_for_ssrf` (or remove
  it entirely) with a call to
  `common.ssrf_guard.assert_url_is_safe(url)`; the shared validator
  returns `(hostname, resolved_ip)`. Store both on the connector instance.
- Wrap every `rl_requests.get` / `rl_requests.post` call in
  `_fetch_page` with `with pin_dns(self._ssrf_hostname,
  self._ssrf_resolved_ip):` so the hostname always resolves to the
  validated IP for the duration of the request. Re-validate on each
  `_fetch_page` call so legitimate DNS changes are picked up.
- Drop the in-file `socket` / `ipaddress` imports that become unused.
- Add unit tests:
  - `test_rest_api_connector_dns_rebinding_blocked` — `getaddrinfo`
    returns a public IP on the first call and `127.0.0.1` on the second;
    the connector must error during `_fetch_page` (with the
    `pin_dns` fix, the second-call IP is never used because the pin
    forces the validated public IP).
  - `test_rest_api_connector_pins_validated_ip` — assert
    `RestAPIConnector` stores the validated IP after construction and
    that `_fetch_page` reads through `pin_dns`.
  - `test_rest_api_connector_rejects_loopback_url` — regression: still
    rejects an explicit `http://127.0.0.1/...` up front.

**Acceptance:**

- The new `test_rest_api_connector_dns_rebinding_blocked` test fails on
  `main` (proves the bug is real) and passes after the fix.
- The existing happy-path tests in
  `test/unit_test/data_source/test_rest_api_connector.py` continue to
  pass unchanged.
- Reproduction step 4 above no longer reaches `127.0.0.1`; the connector
  logs `SSRF guard blocked URL` and the data source test connection
  surfaces the SSRF error to the UI instead of silently fetching the
  internal endpoint.

joaovictor712:

Closes #15173.

`RestAPIConnector` (added in #13545, `common/data_source/rest_api_connector.py`)
shipped its own `_validate_url_for_ssrf` instead of using the project's shared
`common/ssrf_guard.assert_url_is_safe`. The homegrown validator resolved DNS
once at construction time and never pinned the resolved IP. Between
`_validate_url_for_ssrf` (validation) and the actual `rl_requests.get` /
`rl_requests.post` call inside `_fetch_page` (TCP connect), `requests` resolves
DNS **again** — completely decoupled from the validated IP. An attacker who
controls DNS for the configured hostname can return a public IP for the first
lookup (validation passes) and a loopback / RFC 1918 / link-local address
(e.g. `169.254.169.254`) for the second lookup (the request lands on the
internal target). The user-configured `Authorization` / `X-Api-Key` /
`Basic` credentials are exfiltrated in that same request.

This is the same DNS-rebinding / TOCTOU bypass class that
`common/ssrf_guard.pin_dns` exists to prevent — `agent/tools/crawler.py` and the
`misc_utils.download_img` fix (#14868) already use exactly the right pattern.
`rest_api_connector.py` just doesn't.

### What changed

**`common/data_source/rest_api_connector.py`:**

- Drop the homegrown `socket` / `ipaddress` imports and the body of
  `_validate_url_for_ssrf`. The static method now delegates to
  `common.ssrf_guard.assert_url_is_safe(url)`, wraps the resulting `ValueError`
  as `ConnectorValidationError` for API compatibility, and **returns the
  resolved `(hostname, ip)` tuple** so the caller can pin DNS to that IP.
- `_fetch_page` re-validates the built URL on every page fetch (the URL can be
  templated and pagination cursors can alter it), and wraps the
  `rl_requests.get` / `rl_requests.post` call in
  `common.ssrf_guard.pin_dns(hostname, resolved_ip)`. The pin is automatically
  removed after the request, so legitimate DNS changes between fetches are
  picked up on the next re-validation.
- The shared validator also handles a case the homegrown one missed:
  IPv4-mapped IPv6 addresses (`::ffff:127.0.0.1`) are normalised to their
  IPv4 form before the loopback check (see `ssrf_guard._effective_ip`).

**`test/unit_test/data_source/test_rest_api_connector.py`:**

- Updated 17 `patch("common.data_source.rest_api_connector.socket.getaddrinfo")`
  call sites to `patch("common.ssrf_guard._orig_getaddrinfo")` so the existing
  60 tests keep mocking DNS at the right layer after the refactor.
- Updated the SSRF-validation tests' `pytest.raises(..., match=...)` patterns
  to match the new (shared) error message.
- Added `test_ipv4_mapped_loopback_blocked` and `test_unresolvable_host_blocked`
  to `TestSSRFValidation` — the old homegrown validator silently allowed both.
- New `TestDnsRebindingProtection` class with five tests:
  - `test_revalidation_blocks_rebind_to_loopback` — DNS returns a public IP on
    the first call (validation passes) and `127.0.0.1` on the second
    (`_fetch_page` revalidation). The fetch must raise
    `ConnectorValidationError("…not safe to fetch…")` instead of issuing the
    request.
  - `test_revalidation_blocks_rebind_to_cloud_metadata` — same, with
    `169.254.169.254` as the rebind target.
  - `test_fetch_pins_dns_with_validated_ip` — wraps `pin_dns` and asserts it
    was called with the right `(hostname, ip)` so the actual TCP connect can't
    be redirected.
  - `test_validator_returns_hostname_and_ip` — verifies the new return-tuple
    contract.
  - `test_template_substitution_into_url_is_revalidated` — verifies that
    templated URLs are re-resolved on each `_fetch_page` call.

All 65 tests pass locally:

```
======================= 65 passed in 52.93s =======================
```

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/common/data_source/rest_api_connector.py
+++ b/common/data_source/rest_api_connector.py
@@ -15,8 +15,6 @@ from datetime import datetime, timezone
 from typing import Any, Dict, Generator, Iterable, List, Mapping, Optional
 from urllib.parse import parse_qs, urlparse, urlunparse
 
-import ipaddress
-import socket
 import requests
 from pydantic import BaseModel, ConfigDict, Field, HttpUrl, ValidationError, field_validator
 
@@ -28,6 +26,7 @@ from common.data_source.exceptions import (
     ConnectorMissingCredentialError,
     ConnectorValidationError,
 )
+from common.ssrf_guard import assert_url_is_safe, pin_dns
 from common.data_source.interfaces import (
     LoadConnector,
     PollConnector,
@@ -171,64 +170,27 @@ class RestAPIConnector(LoadConnector, PollConnector):
     """
 
     @staticmethod
-    def _validate_url_for_ssrf(url: str) -> None:
-        """Validate that the URL does not point to localhost or private/internal networks.
+    def _validate_url_for_ssrf(url: str) -> tuple[str, str]:
+        """Validate that *url* is safe to fetch, returning the resolved ``(hostname, ip)``.
+
+        Delegates to :func:`common.ssrf_guard.assert_url_is_safe`, which performs
+        the same allowlist (only globally-routable unicast addresses) and also
+        normalises IPv4-mapped IPv6 addresses (e.g. ``::ffff:127.0.0.1``) before
+        the check. Returning the validated IP lets the caller pin DNS to that
+        address with :func:`common.ssrf_guard.pin_dns` so the actual HTTP
+        request cannot be redirected to a different IP between validation and
+        connect (DNS rebinding / TOCTOU).
 
         Raises:
             ConnectorValidationError: If the URL is considered unsafe.
         """
-        parsed = urlparse(str(url))
-
-        if parsed.scheme not in ("http", "https"):
-            msg = f"Unsupported URL scheme for REST API connector: {parsed.scheme!r}. Only http/https are allowed."
-            logger.warning(msg)
-            raise ConnectorValidationError(msg)
-
-        hostname = parsed.hostname
-        if not hostname:
-            msg = "REST API connector URL must include a hostname."
-            logger.warning(msg)
-            raise ConnectorValidationError(msg)
-
-        # Quick checks for obvious localhost-style hostnames.
-        lower_host = hostname.lower()
-        if lower_host in ("localhost",):
-            msg = f"REST API connector URL hostname {hostname!r} is not allowed (localhost is blocked)."
-            logger.warning(msg)
-            raise ConnectorValidationError(msg)
-
         try:
-            addrinfo_list = socket.getaddrinfo(hostname, None)
-        except OSError as exc:
-            # If resolution fails, log and let higher-level validation (if any) decide.
-            # We do not treat this as an SSRF condition by itself.
-            logger.info("DNS resolution failed for REST API connector URL %r: %s", url, exc)
-            return
-
-        for family, _, _, _, sockaddr in addrinfo_list:
-            ip_str = sockaddr[0]
-            try:
-                ip_obj = ipaddress.ip_address(ip_str)
-            except ValueError:
-                # Not an IP address we understand; skip.
-                logger.debug("Skipping non-IP address resolved from %r: %r", hostname, ip_str)
-                continue
-
-            if (
-                ip_obj.is_loopback
-                or ip_obj.is_private
-                or ip_obj.is_link_local
-                or ip_obj.is_reserved
-                or ip_obj.is_multicast
-            ):
-                msg = (
-                    f"REST API connector URL {url!r} resolves to disallowed address {ip_str} "
-                    "(localhost, private, link-local, reserved, or multicast addresses are blocked)."
-                )
-                logger.warning(msg)
-                raise ConnectorValidationError(msg)
-
-        logger.debug("REST API connector URL %r passed SSRF safety validation.", url)
+            return assert_url_is_safe(str(url))
+        except ValueError as exc:
+            logger.warning("REST API connector URL %r failed SSRF safety check: %s", url, exc)
+            raise ConnectorValidationError(
+                f"REST API connector URL {url!r} is not safe to fetch: {exc}"
+            ) from exc
 
     def __init__(
         self,
@@ -253,7 +215,10 @@ class RestAPIConnector(LoadConnector, PollConnector):
         field_default_values: Optional[Dict[str, Any]] = None,
         content_template: Optional[str] = None,
     ) -> None:
-        # Validate URL against SSRF-style targets (localhost, private/internal ranges, etc.)
+        # Validate URL against SSRF-style targets (localhost, private/internal ranges, etc.).
+        # We re-validate (and re-pin) on every fetch in :meth:`_fetch_page` to close
+        # the DNS rebinding / TOCTOU window between this validation and the actual
+        # socket connect, so we discard the resolved IP here.
         self._validate_url_for_ssrf(url)
 
         parsed = urlparse(str(url))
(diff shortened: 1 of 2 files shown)


mkaaad:
fixed in #16409, This PR is now ready to be closed.

PR #15177, opened 2026-05-24
Fix [Bug]: Save parser configs in dataset configuration page is not working #15175

mustangxu:

Fix [Bug]: Save parser configs in dataset configuration page is not working #15175

### Type of change


wangq8:
LGTM

PR #15186, opened 2026-05-24
feat(mcp): add Agent Chat tools to official RAGFlow MCP server

AI-Mart:
## Summary

Add Agent Chat MCP tools to the official RAGFlow MCP server. Currently RAGFlow MCP only exposes `ragflow_retrieval` (semantic search). This PR adds 6 tools to support the full Agent Chat workflow, enabling external AI agents (like Hermes, Claude, etc.) to interact with RAGFlow workflow agents.

## Motivation

RAGFlow Agent system provides powerful workflow automation for \`问策\` (strategy discussion) and \`问数\` (data query). However, the official MCP server only exposes \`ragflow_retrieval\`, leaving Agent Chat completely inaccessible via MCP. Users wanting to integrate RAGFlow Agents with external AI assistants must either use the REST API directly or rely on third-party MCP servers.

This PR brings first-class MCP support to RAGFlow Agent system.

## Added Tools

| Tool | Description | API Endpoint |
|------|-------------|-------------|
| \`ragflow_list_agents\` | List all workflow agents | GET /api/v1/agents |
| \`ragflow_list_sessions\` | List conversation sessions | GET /api/v1/agents/{id}/sessions |
| \`ragflow_create_session\` | Create a new session | POST /api/v1/agents/{id}/sessions |
| \`ragflow_send_message\` | Send message to agent | POST /api/v1/agents/chat/completions |
| \`ragflow_generate_recommendations\` | Generate recommended questions | POST /api/v1/chat/recommendation |
| \`ragflow_delete_sessions\` | Delete sessions | DELETE /api/v1/agents/{id}/sessions |

## API Endpoints Used

- \`GET /api/v1/agents\` — List workflow agents
- \`GET /api/v1/agents/{agent_id}/sessions\` — List agent sessions
- \`POST /api/v1/agents/{agent_id}/sessions\` — Create new session
- \`POST /api/v1/agents/chat/completions\` — Core agent chat endpoint
- \`POST /api/v1/chat/recommendation\` — Generate question recommendations
- \`DELETE /api/v1/agents/{agent_id}/sessions\` — Delete sessions

## Changes

- Add \`_delete\` method to \`RAGFlowConnector\` for HTTP DELETE requests
- Add 6 Agent Chat methods to \`RAGFlowConnector\`: \`list_agents\`, \`list_sessions\`, \`create_session\`, \`send_message\`, \`send_message_simple\`, \`generate_recommendations\`, \`delete_sessions\`
- Add 6 new tools to \`list_tools\` with complete inputSchema
- Add Agent Chat tool handling in \`call_tool\`
- All changes follow existing RAGFlow MCP code style (async, httpx, @with_api_key decorator)

## Backward Compatibility

- **Fully backward compatible**: only adds new tools, original \`ragflow_retrieval\` unchanged
- No breaking changes to existing MCP clients

## Testing

- Syntax validation: \`python3 -m py_compile mcp/server/server.py\` ✅
- Verified against RAGFlow v0.25.5 API endpoint structure
- All new methods handle error cases (None response, non-200 status codes, error codes)

## Checklist

- [x] Code follows existing RAGFlow MCP style (async, httpx, @with_api_key)
- [x] Backward compatible (only adds new tools)
- [x] All new tools have complete inputSchema definitions
- [x] Error handling for all new endpoints
- [x] Syntax validated with py_compile

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -165,6 +165,11 @@ async def parse(tenant_id, dataset_id):
     dataset_tenant_id = _get_dataset_tenant_id(dataset_id)
     if not dataset_tenant_id:
         return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
+    e, kb = KnowledgebaseService.get_by_id(dataset_id)
+    if not e:
+        return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
+    if kb.pipeline_id:
+        return get_error_data_result(message="Datasets configured with an ingestion pipeline cannot be parsed with `/datasets/{dataset_id}/chunks`. Use `/documents/ingest` instead.", code=RetCode.ARGUMENT_ERROR)
     req = await get_request_json()
     if not req.get("document_ids"):
         return get_error_data_result("`document_ids` is required")
--- a/docs/references/http_api_reference.md
+++ b/docs/references/http_api_reference.md
@@ -1952,7 +1952,11 @@ Failure:
 
 **POST** `/api/v1/datasets/{dataset_id}/chunks`
 
-Parses documents in a specified dataset.
+Parses documents in a specified dataset using the built-in chunking pipeline.
+
+:::note
+This endpoint only supports datasets that use the built-in chunking pipeline. For datasets configured with an ingestion pipeline, use `POST /api/v1/documents/ingest` instead.
+:::
 
 #### Request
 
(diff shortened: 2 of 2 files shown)


AI-Mart:
Could you please speed up the progress review?

AI-Mart:
Could you expedite the review of this PR to ensure seamless integration between RAGFlow and cutting-edge agent frameworks?

yuzhichang:
Thanks for pushing Agent Chat support into the official MCP server. The motivation makes sense, but after looking through the MCP side and the backend agent/session APIs, I think there is an important design gap here that we should resolve before we merge this shape into the default server.

The current MCP server is built around a one-shot tool model: a client invokes a tool once and gets back one final result. `ragflow_retrieval` fits that contract well because it is effectively a synchronous query. Agent chat is different. The underlying `/api/v1/agents/chat/completions` flow is designed around sessionful, often streaming interaction, with intermediate execution events and a final answer layered on top. Exposing that as a plain one-shot tool call creates a semantic mismatch.

A few concrete consequences of that mismatch:

1. Session semantics become implicit instead of explicit.
Agent conversations depend on `session_id` to preserve context across turns, while MCP tool calls are naturally stateless. If the MCP integration does not make session lifecycle a first-class concept, users will think they are “chatting with an agent” when they are actually starting a fresh run on every call.

2. Streaming / eventful execution gets flattened into a single text response.
Agent execution can include intermediate node activity, tool usage, trace information, and delayed final output. A one-shot MCP response can only represent a compressed final artifact unless we deliberately define how execution progress should be surfaced. Without that design, we are not really exposing “agent chat”; we are exposing a lossy adapter over the final answer only.

3. Error handling and partial-progress semantics differ.
A retrieval tool is usually success/failure. Agent execution can have richer states: partial progress, recoverable failures, traceable execution paths, or long-running work. MCP clients consuming a simple tool response may not know how to interpret those states correctly unless the adapter contract is designed up front.

Because of that, I do not think the right question is only “can we call these REST endpoints from MCP?” The bigger question is “what subset of agent behavior should the official MCP server promise to external clients?”

My suggestion would be to narrow the first mergeable slice:

1. Define the official MCP contract as “non-streaming final-answer mode” for agents.
That means the MCP tool intentionally exposes only the final answer, not the full execution/event model.

2. Make session management explicit and unavoidable in the API design.
If we want multi-turn behavior, clients should explicitly create/store/pass `session_id`. We should not make this look like a naturally stateful chat surface when MCP itself is not.

3. Keep the initial toolset non-destructive.
`list_agents`, `list_sessions`, `create_session`, and a carefully specified `send_message` are much easier to justify than also adding destructive operations like bulk session deletion to the default official server.

4. Consider deferring streaming-style support to a follow-up design.
If we eventually want richer agent execution over MCP, that probably deserves a dedicated protocol shape for progress/events/trace instead of being folded into a plain one-shot text tool.

In short: the motivation is good, but the current PR mixes a one-shot MCP interaction model with a sessionful/streaming agent model without first defining the boundary between them. I would be much more comfortable merging a smaller, explicitly non-streaming, explicitly sessioned, non-destructive subset first, and then iterating from there.

yuzhichang (changes requested):
The current design still needs action on the MCP/agent interaction model before this is ready to merge. Please address the review comments about the one-shot MCP contract vs. sessionful/streaming agent semantics and narrow the first mergeable slice accordingly.

PR #15200, opened 2026-05-25
fix: validate custom model inputs

### LINKED ISSUE #15199: [Bug]: Add custom model endpoint can panic on empty model_types and ignores path provider fields
### Self Checks

`8d8ea718779d3e9498768ab73a5c6b18035a2262`

### RAGFlow image version

`main` branch at commit `8d8ea718779d3e9498768ab73a5c6b18035a2262`; not image-version specific.

- Hardware parameters: N/A
- OS type: N/A
- Others: Reproducible from `internal/handler/providers.go`, `internal/service/model_service.go`, and the provider route in `internal/router/router.go`.

### Actual behavior

The add-custom-model route includes both provider and instance in the URL:

```go
POST /api/v1/providers/:provider_name/instances/:instance_name/models
```

However, `AddCustomModel` does not read `provider_name` or `instance_name` from the path. It validates and uses only the JSON body fields `provider_name` and `instance_name`.

This causes two practical problems:

1. A request can target `/providers/openai/instances/default/models` but put a different provider or instance in the JSON body. The service uses the body values, not the URL target.
2. The handler only checks `req.ModelTypes == nil`. An empty array such as `"model_types": []` passes validation, then `ModelProviderService.AddCustomModel` indexes `request.ModelTypes[0]`, causing an index-out-of-range panic.

### Expected behavior

The endpoint should use `provider_name` and `instance_name` from the path, or explicitly reject requests where body values do not match the path target.

It should also reject missing or empty `model_types` before the service indexes the slice. Invalid input should return a normal bad-request response instead of a panic.

### Steps to reproduce

1. Run RAGFlow from `main` at commit `8d8ea718779d3e9498768ab73a5c6b18035a2262`.
2. Send a request like:

```bash
curl -X POST "http://localhost:<port>/api/v1/providers/openai/instances/default/models" \
  -H "Content-Type: application/json" \
  -d '{
    "provider_name": "openai",
    "instance_name": "default",
    "model_name": "empty-types-model",
    "model_types": [],
    "max_tokens": 8192
  }'
```

3. Observe that the handler accepts the request because `model_types` is non-nil.
4. Observe that the service indexes `request.ModelTypes[0]` and can panic with an index-out-of-range error.
5. Send the same URL with a JSON body whose `provider_name` or `instance_name` points somewhere else. The handler uses the body values instead of the URL path values.

### Additional information

I searched open and closed issues/PRs for related terms including `AddCustomModel model_types empty`, `"ModelTypes[0]"`, `custom model empty model_types`, `AddCustomModel provider_name instance_name`, and `"Add custom model" route params`.

Related PRs found in broad searches, such as #14166, #14441, and #14485, do not fix the current `AddCustomModel` path/body mismatch or the empty `model_types` panic path. I also checked the broad provider API PR #14595; it is not an exact duplicate of this Go handler validation issue.

bitloi:

Closes #15199.

The add-custom-model endpoint is routed through `/api/v1/providers/:provider_name/instances/:instance_name/models`, but the handler previously trusted `provider_name` and `instance_name` from the JSON body instead of the path target. A request could therefore hit one provider/instance URL while operating on a different body provider/instance.

The same handler only rejected `model_types` when the slice was nil. An empty array passed validation and reached `ModelProviderService.AddCustomModel`, where `request.ModelTypes[0]` could panic.

This PR makes the path provider/instance authoritative, rejects mismatched body values, rejects missing or empty `model_types`, and adds a service-level guard so direct service callers cannot hit the same panic path.

Validation:

```bash
/root/go/bin/go test ./internal/handler ./internal/service -vet=off -count=1
git diff --check upstream/main...HEAD
```

### Type of change


bitloi:
@Haruko386 Would you please review the PR?

Haruko386:
Hi @bitloi I will review your code soon

internal/handler/providers.go
```diff
@@ -723,45 +724,51 @@ func (h *ProviderHandler) EnableOrDisableModel(c *gin.Context) {
...
-			"message": "Instance name is required",
-		})
-		return
+	if req.ProviderName != "" && req.ProviderName != providerName {
+		return errors.New("Provider name does not match path")
+	}
+
+	if req.InstanceName != "" && req.InstanceName != instanceName {
+		return errors.New("Instance name does not match path")
 	}
```
Haruko386, line 742: I suggest converting all to lowercase before comparing. 
Like:
```go
	if req.ProviderName != "" && strings.ToLower(req.ProviderName) != strings.ToLower(providerName) {
		return errors.New("Provider name does not match path")
	}

	if req.InstanceName != "" && strings.ToLower(req.InstanceName) != strings.ToLower(instanceName) {
		return errors.New("Instance name does not match path")
	}
```

Haruko386 (changes requested):
Hi @bitloi , would you do some changes?

bitloi:
> Hi @bitloi , would you do some changes?

Done!

Haruko386 (approved):
@bitloi LGTM

bitloi:
@KevinHuSh Would you please review the PR?

PR #15202, opened 2026-05-25
Fix/db connection loss transaction leak

### LINKED ISSUE #15198: [Bug]: DB connection-loss retry corrupts state inside transactions and permanently leaks pooled connections (eventual MaxConnectionsExceeded outage)`
### Self Checks

44cb5565f

### RAGFlow image version

`nightly` / `v0.22.0` and earlier (the retry classes have shipped unchanged). Reproduced against the installed `peewee==3.19.0`.  ---

```Markdown
- Hardware parameters: N/A (logic bug, hardware-independent)
- OS type: Linux (any)
- Others:
  - DB backend: MySQL / PostgreSQL / OceanBase — all three are affected
  - peewee: 3.19.0 (playhouse pooled databases)
  - Pool config (conf/service_conf.yaml): max_connections: 900, stale_timeout: 300
  - Trigger: a connection drop (MySQL 2006/2013, failover, restart, network blip)
    that lands WHILE a `with DB.atomic():` transaction is open


---
```

### Actual behavior

RAGFlow's custom DB retry layer — the mechanism meant to make the backend resilient to dropped MySQL/PostgreSQL/OceanBase connections — is broken **for transactional writes**.

When a connection drops during a `DB.atomic()` transaction, `_handle_connection_loss()` masks the real (recoverable) error with a bogus `OperationalError: Connection already opened.`, **skips all retries**, leaves the transaction dangling, and **permanently leaks the dead connection** from the pool. Repeated occurrences exhaust the pool and bring the whole server down with `playhouse.pool.MaxConnectionsExceeded` until restart.

**Location** (identical bug, copied verbatim across all three DB classes):

- `RetryingPooledMySQLDatabase._handle_connection_loss` — `api/db/db_models.py:290-306`
- `RetryingPooledPostgresqlDatabase._handle_connection_loss` — `api/db/db_models.py:367-381`
- `RetryingPooledOceanBaseDatabase._handle_connection_loss` — `api/db/db_models.py:443-457`
- Reached via `execute_sql` (`:266`) and `begin` (`:308`).

```python
def _handle_connection_loss(self):
    try:
        self.close()      # ← peewee RAISES here if a transaction is open; swallowed
    except Exception:
        pass
    try:
        self.connect()    # ← peewee raises 'Connection already opened.' (state never reset)
    except Exception as e:
        logging.error(f"Failed to reconnect: {e}")
        time.sleep(0.1)
        try:
            self.connect()  # ← raises 'Connection already opened.' again → propagates
        except Exception as e2:
            raise
```
### Root cause (verified against the installed `peewee==3.19.0`)

The code treats `self` like a raw connection, but `self` is a **pooled peewee `Database`** whose `close()`/`connect()` are guarded:

- **`Database.close()`** raises `OperationalError('Attempting to close database while transaction is open.')` when `in_transaction()`. The guard fires **before** the `try/finally` that calls `_state.reset()`, so on this raise `self._state.closed` stays `False`:

  ```python
  def close(self):
      with self._lock:
          ...
          if self.in_transaction():
              raise OperationalError('Attempting to close database while transaction is open.')
          is_open = not self._state.closed
          try:
              if is_open:
                  with __exception_wrapper__:
                      self._close(self._state.conn)
          finally:
              self._state.reset()   # ← never reached when the guard above raises
          return is_open
  ```

- **`Database.connect()`** raises `OperationalError('Connection already opened.')` whenever `not self._state.closed`:

  ```python
  def connect(self, reuse_if_open=False):
      with self._lock:
          ...
          if not self._state.closed:
              if reuse_if_open:
                  return False
              raise OperationalError('Connection already opened.')
          ...
  ```

So **inside a transaction**: `close()` raises → swallowed → state not reset → `connect()` raises `'Connection already opened.'` → second `connect()` raises the same → it propagates. The retry loop in `execute_sql` is aborted and the original transient `2013`/`2006` error is **replaced by the misleading one**. The dead connection is never popped from the pool's `_in_use` map, so it is **leaked**. (`manual_close()` doesn't help either — it also calls the guarded `close()`.)

### Why it's critical

1. **Defeats resilience exactly when needed.** The retry classes exist specifically to survive connection drops (2006 "server has gone away", 2013 "Lost connection", failover, restart, network blips). For transactional writes they do the opposite — turning a recoverable blip into a hard failure.
2. **Hits the common write path.** `DB.atomic()` wraps core operations — verified present at:
   - document delete/create — `api/db/services/document_service.py:595, 641, 683`
   - user/tenant creation — `api/db/services/user_service.py:132, 139, 149`
   - bulk inserts / common writes — `api/db/services/common_service.py:232, 260, 328, 340, 353`
3. **Permanent connection leak → full outage.** Pool is `max_connections: 900`, `stale_timeout: 300` (`conf/service_conf.yaml:13-14`, `docker/service_conf.yaml.template:13-14`). Each occurrence leaks one `_in_use` entry; once they reach the cap, every new request raises `MaxConnectionsExceeded('Exceeded maximum connections.')` (peewee `playhouse/pool.py`) — a server-wide outage needing a restart. `stale_timeout` does **not** rescue these: the trigger is a mid-query server drop, not idle staleness.
4. **Poisoned thread-local state + a misleading error** (`Connection already opened.` instead of `Lost connection`) makes it hard to diagnose.

---

### Expected behavior

- A connection drop during a `DB.atomic()` transaction should **not** be masked by a bogus `Connection already opened.` error.
- The dead connection should be **evicted from the pool** (not leaked), so repeated drops cannot drive the pool to `MaxConnectionsExceeded`.
- The thread-local DB state should be reset so a clean reconnect is possible.
- Because retrying a single statement while `in_transaction()` is **unsound** (the server-side transaction is already gone and prior statements are lost), the layer should evict the connection and re-raise so the enclosing `atomic()` rolls back and the caller can retry the whole unit of work.

---

### Steps to reproduce

```Markdown
This is reproducible without a live MySQL, using the **real** peewee internals (`close`/`connect`/`in_transaction`) and an **exact copy** of `RetryingPooledMySQLDatabase`. Only the low-level socket/cursor is mocked, to simulate a dropped connection.


1. Use peewee==3.19.0 (the version RAGFlow installs).
2. Save the script below as repro_db_bug.py.
3. Run: python repro_db_bug.py


<details>
<summary><code>repro_db_bug.py</code></summary>


import time, logging, types
from playhouse.pool import PooledMySQLDatabase
from peewee import OperationalError, InterfaceError

logging.basicConfig(level=logging.WARNING, format="    [log] %(message)s")

# ---- EXACT copy of api/db/db_models.py:260-306 ----
class RetryingPooledMySQLDatabase(PooledMySQLDatabase):
    def __init__(self, *args, **kwargs):
        self.max_retries = kwargs.pop("max_retries", 5)
        self.retry_delay = kwargs.pop("retry_delay", 1)
        super().__init__(*args, **kwargs)

    def execute_sql(self, sql, params=None, commit=True):
        for attempt in range(self.max_retries + 1):
            try:
                return super().execute_sql(sql, params, commit)
            except (OperationalError, InterfaceError) as e:
                error_codes = [2013, 2006]
                error_messages = ['', 'Lost connection']
                should_retry = (
                    (hasattr(e, 'args') and e.args and e.args[0] in error_codes) or
                    (str(e) in error_messages) or
                    (hasattr(e, '__class__') and e.__class__.__name__ == 'InterfaceError')
                )
                if should_retry and attempt < self.max_retries:
                    logging.warning(f"Database connection issue (attempt {attempt+1}/{self.max_retries}): {e}")
                    self._handle_connection_loss()
                    time.sleep(0.0)
                else:
                    logging.error(f"DB execution failure: {e}")
                    raise
        return None

    def _handle_connection_loss(self):
        try:
            self.close()
        except Exception:
            pass
        try:
            self.connect()
        except Exception as e:
            logging.error(f"Failed to reconnect: {e}")
            time.sleep(0.0)
            try:
                self.connect()
            except Exception as e2:
                logging.error(f"Failed to reconnect on second attempt: {e2}")
                raise
# ---------------------------------------------------

def make_db():
    return RetryingPooledMySQLDatabase("ragflow", max_retries=5, retry_delay=1,
                                       host="127.0.0.1", port=3306, user="x", password="y")

def fresh_dead_connection_state(db, in_transaction):
    fake_conn = object()
    db._state.conn = fake_conn
    db._state.closed = False
    db._state.transactions = [object()] if in_transaction else []
    db._in_use = {db.conn_key(fake_conn): types.SimpleNamespace(timestamp=time.time(), connection=fake_conn)}
    db._close = lambda conn, close_conn=False: db._in_use.pop(db.conn_key(conn), None)
    db._connect = lambda: object()
    db._initialize_connection = lambda conn: None
    db._set_server_version = lambda conn: None
    db.server_version = (8, 0, 0)

print("CASE A: connection lost OUTSIDE a transaction")
db = make_db(); fresh_dead_connection_state(db, in_transaction=False)
try:
    db._handle_connection_loss()
    print(f"  RESULT: recovered OK. is_closed={db.is_closed()} in_transaction={db.in_transaction()}")
except Exception as e:
    print(f"  RESULT: RAISED {type(e).__name__}: {e}")

print("CASE B: connection lost INSIDE a transaction (`with DB.atomic():`)")
db = make_db(); fresh_dead_connection_state(db, in_transaction=True)
try:
    db._handle_connection_loss()
    print(f"  RESULT: recovered OK. is_closed={db.is_closed()}")
except Exception as e:
    print(f"  RESULT: RAISED -> {type(e).__name__}: {e}")
    print(f"  state AFTER: is_closed={db.is_closed()} in_transaction={db.in_transaction()} "
          f"dead_conn_still_in_pool={len(db._in_use)}")

print("CASE C: full execute_sql() path, error 2013 raised INSIDE a transaction")
db = make_db(); fresh_dead_connection_state(db, in_transaction=True)
class DeadCursor:
    def execute(self, sql, params=()):
        raise OperationalError(2013, "Lost connection to MySQL server during query")
    def close(self): pass
db.cursor = lambda commit=None, named_cursor=None: DeadCursor()
try:
    db.execute_sql("UPDATE document SET token_num=token_num+1 WHERE id=%s", ("doc1",))
    print("  RESULT: query succeeded (unexpected)")
except Exception as e:
    print(f"  RESULT: execute_sql RAISED -> {type(e).__name__}: {e}")


</details>

### Observed output


CASE A: connection lost OUTSIDE a transaction
  RESULT: recovered OK. is_closed=False in_transaction=False        ← retry works here

CASE B: connection lost INSIDE a transaction (`with DB.atomic():`)
  RESULT: RAISED -> OperationalError: Connection already opened.
  state AFTER: is_closed=False in_transaction=True dead_conn_still_in_pool=1   ← LEAK

CASE C: full execute_sql() path, error 2013 raised INSIDE a transaction
  RESULT: execute_sql RAISED -> OperationalError: Connection already opened.
  => a recoverable 'Lost connection' (2013) inside atomic() becomes a fatal
     'Connection already opened.'; NO retry; the pooled connection is leaked.


---
```

### Additional information

Bypass peewee's guarded `close()` — evict the dead connection from the pool directly and reset thread-local state before reconnecting; and do not retry a single statement while a transaction is open. Apply to **all three** DB classes (MySQL, PostgreSQL, OceanBase).

```python
def _handle_connection_loss(self):
    conn = self._state.conn
    if conn is not None:
        self._in_use.pop(self.conn_key(conn), None)   # un-leak it from the pool
        try:
            conn.close()
        except Exception:
            pass
    self._state.reset()   # clears dangling transaction + closed flag
    self.connect()        # now safe: _state.closed is True
```

dale053:
Closes #15198

Fixes a pooled-connection leak (issue #15198) that occurs when the database connection drops **while a transaction is open**.

The previous recovery path did `self.close(); self.connect()`. When a connection is lost mid-transaction this is broken:

1. peewee's `close()` raises `"Attempting to close database while transaction is open."` **before** it resets `_state`, so the swallowed exception leaves the thread-local state untouched.
2. The following `connect()` then raises `"Connection already opened."`, masking the real error.
3. The dead connection is never evicted from the pool's `_in_use` map, so it **leaks**. Repeated occurrences exhaust the connection pool and take the server down.

A second, more subtle bug: a single failing statement was being replayed on retry even when a transaction was open. Since the server has already discarded the transaction, replaying one statement in isolation can corrupt data.

**Fix:**

- Introduce a shared `_ConnectionLossRecoveryMixin` used by `RetryingPooledMySQLDatabase`, `RetryingPooledPostgresqlDatabase`, and `RetryingPooledOceanBaseDatabase` (removing the three duplicated, buggy `_handle_connection_loss` implementations).
- `_evict_connection()` drops the current thread's connection from the pool and closes the underlying socket, bypassing peewee's transaction-guarded `close()` — this is what prevents the leak.
- Recovery outside a transaction evicts the dead connection, resets thread-local state, then reconnects.
- When a connection drops mid-transaction, evict the dead connection and let the **whole transaction** abort/retry upstream instead of replaying a single statement.
- Factor the per-engine connection-loss detection into a `_is_connection_loss(e)` helper to remove the duplicated detection logic across `execute_sql` and `begin`.

**Tests:** Added `TestConnectionLossRecovery` in `test/unit_test/api/db/test_oceanbase_peewee.py` driving the real peewee/pool machinery with a mocked DB-API connection. Covers: transparent recovery outside a transaction, no connection/socket leak across 50 mid-transaction failures, a clean surfaced error (no `"Connection already opened"`), non-connection errors not being retried, and `begin()` recovery.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


dale053:
Hi,@wangq8

dale053:
Hi, @JinHai-CN 

wangq8:
@dale053 please delete image from PR, the png file should not be in git repo.

dale053:
Hi,@wangq8!
I have just removed image from this PR. could you please review it again?
Thank you for your time.

dale053:
Hi,@yingfeng

hnwyllmm:
Thanks for working on this. The overall direction is sound: a dropped connection must be evicted, and a single statement must never be replayed after the connection is lost inside an active transaction. However, I think the following issues should be addressed before this is merged:

1. **Protect `_in_use` with the pool lock.**

   `_evict_connection()` mutates the shared `self._in_use` dictionary without acquiring Peewee's `self._pool_lock`. Other threads can concurrently run `close_stale()` (RAGFlow calls it during request teardown), which iterates and then replaces `_in_use` while holding that lock. An unlocked `pop()` can therefore race with the iteration, or a stale `close_stale()` snapshot can reintroduce the connection that was just evicted. Please perform the pool bookkeeping under `_pool_lock`.

2. **Do not leave the transaction state pointing at a closed connection.**

   `_abort_transaction_after_connection_loss()` closes the raw connection but intentionally leaves `self._state` unchanged. When Peewee unwinds `atomic()`, it attempts `ROLLBACK` using that closed connection. The rollback failure can replace the original statement error at the top of the exception chain. After a bare `atomic()` block (without an outer `connection_context`), the resulting state is also inconsistent: `in_transaction()` is false and `_in_use` is empty, but `is_closed()` is still false and a direct `connect()` raises `OperationalError: Connection already opened.`

   The implementation should preserve the original connection-loss error and ensure that the thread-local state becomes clean after the transaction context has unwound, without assuming that every `atomic()` block is wrapped by `connection_context()`.

3. **Extend the regression tests to cover the actual failure boundaries.**

   `_run_request()` always calls `db.close()` before returning, so the current tests hide the inconsistent state described above. Please add coverage for:

   - a connection loss inside a bare `atomic()` block;
   - the state immediately after `atomic()` exits;
   - a successful clean `connect()` or subsequent operation;
   - preservation/classification of the original statement exception when rollback also fails;
   - concurrent eviction versus `close_stale()`;
   - the PostgreSQL pooled recovery path, not only its error classifier.

4. **Rebase and remove unrelated documentation changes.**

   This branch is far behind the current `main` and now conflicts in `api/db/db_models.py`. Current `main` also has a separate GaussDB retry path and a shared `is_psycopg_connection_error()` helper, which should be reconciled rather than introducing another duplicate path. The sandbox documentation changes are unrelated to this fix, contain duplicate `Provider options` sections, and still reference an image that was removed from the branch; they should be dropped from this PR.

Also, the `ragflow_tests` check is currently skipped, so the rebased change should receive a real targeted unit-test run before merge.

In short, I agree with the transaction-level recovery policy in this PR, but the pool mutation must be synchronized and the post-transaction state/exception behavior needs to be made reliable first.

6ba3i (commented):
Thanks for the fix. The root cause is correct, and I agree that a statement should never be retried after the connection dies inside an active transaction.

I still see three blockers:

1. **Transaction state is left inconsistent.**
   `_abort_transaction_after_connection_loss()` closes/evicts the connection but leaves Peewee state alive. When `atomic()` exits, Peewee can still try to `ROLLBACK` or `COMMIT` on that dead connection, which can mask the original error and leave the DB reporting `Connection already opened.` afterward. The current tests hide this by always calling `db.close()` during cleanup.

2. **`_in_use` is modified without `_pool_lock`.**
   Peewee protects its pool bookkeeping with this lock. I reproduced a race with `close_stale()` resulting in `RuntimeError: dictionary changed size during iteration`. Eviction should follow Peewee’s locking discipline.

3. **This needs rebasing against current `main`.**
   GaussDB now uses the same pooled PostgreSQL foundation and likely needs the same lifecycle fix. Current `main` also has shared psycopg connection-error detection, so the new PostgreSQL classifier should probably reuse that instead of duplicating a narrower version.

The overall direction is good, but the mid-transaction lifecycle still needs to be fixed before merge. I’d also add tests for bare `atomic()`, exceptions caught inside `atomic()`, state immediately after failure, and concurrent eviction vs `close_stale()`.

6ba3i:
@dale053 could you please fix the issues raised in my comment ?

6ba3i:
overlaps #19530

PR #15219, opened 2026-05-25
Docs: Minimum required Python version increased to 3.13.

writinwaters:

Minimum Python version increased to 3.13.

### Type of change

- [x] Documentation Update


writinwaters:
@6ba3i I took the liberty of using the right-to-left formmating on the Arabian translation, and would appreciate your double check.

Cheers.

PR #15222, opened 2026-05-25
feat(go): implement provider DeerAPI

dripsmvcp:
## Summary
- Adds chat support for **DeerAPI** as a new `DeerAPIModel` Go driver, so the Go API server can route DeerAPI instances. The `conf/llm_factories.json` DeerAPI factory and the Python side already exist; only the Go driver was missing.

## Root Cause
`conf/llm_factories.json` registers a `DeerAPI` factory (tags `LLM,TEXT EMBEDDING,IMAGE2TEXT`, 23 chat models), and the Python side routes DeerAPI through the unified LiteLLM dispatcher — `SupportedLiteLLMProvider.DeerAPI` with base URL `https://api.deerapi.com/v1` and prefix `openai/` (`rag/llm/__init__.py:55,88,129`). But `internal/entity/models/` had no `deerapi.go` and `conf/models/` had no `deerapi.json`, so DeerAPI fell through to `DummyModel` and could not be used from the Go API server.

## Fix
DeerAPI is an OpenAI-API-compatible aggregator, so the driver mirrors the existing Astraflow / Novita SaaS pattern:

- **`conf/models/deerapi.json`** — factory entry, default base URL `https://api.deerapi.com/v1`, 23 chat models (GPT-5/4.1, Claude, Grok, DeepSeek, Qwen).
- **`internal/entity/models/deerapi.go`** — `ChatWithMessages`, `ChatStreamlyWithSender` (SSE with `reasoning_content` routing, `finish_reason` honored on delta-less terminal frames, upstream-error surfacing), `ListModels`, `CheckConnection`. All other method types return `no such method`.
- **`internal/entity/models/factory.go`** — registers `"deerapi"`.
- **`internal/entity/models/deerapi_test.go`** — happy paths, missing-key / empty-messages / HTTP-error guards, streaming reasoning split, delta-less `finish_reason`, malformed-frame and upstream-error surfacing, `ListModels` / `CheckConnection`, and `no such method` for unimplemented methods.

This initial driver implements **chat only**. Embed (`TEXT EMBEDDING`) and Image2Text (`IMAGE2TEXT`), both advertised in the factory tags, are deferred to follow-up PRs — mirroring how Novita, TogetherAI, and Astraflow landed method-by-method.

## Test Plan
- [x] Regression test added: `internal/entity/models/deerapi_test.go`
- [ ] `go test ./internal/entity/models/...` — not run locally (no Go toolchain in this environment). The driver is a mechanical adaptation of the already-merged Astraflow/Novita drivers; relying on CI.

Part of the Go API-server provider rewrite (#14736).

### DIFF
--- a/conf/models/deerapi.json
+++ b/conf/models/deerapi.json
@@ -1,0 +1,174 @@
+{
+  "name": "DeerAPI",
+  "url": {
+    "default": "https://api.deerapi.com/v1"
+  },
+  "url_suffix": {
+    "chat": "chat/completions",
+    "models": "models"
+  },
+  "class": "deerapi",
+  "models": [
+    {
+      "name": "gpt-5-chat-latest",
+      "max_tokens": 400000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "chatgpt-4o-latest",
+      "max_tokens": 128000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-5-mini",
+      "max_tokens": 400000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-5-nano",
+      "max_tokens": 400000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-5",
+      "max_tokens": 400000,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-4.1-mini",
+      "max_tokens": 1047576,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-4.1-nano",
+      "max_tokens": 1047576,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-4.1",
+      "max_tokens": 1047576,
+      "model_types": [
+        "chat"
+      ]
+    },
+    {
+      "name": "gpt-4o-mini",
+      "max_tokens": 128000,
+      "model_types": [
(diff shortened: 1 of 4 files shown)


Haruko386 (changes requested):
@dripsmvcp `DeerAPI` has stoped register.
How did you get Api-Key.
Let me know if I was wrong

dripsmvcp:
Hey @Haruko386 👋 thanks for flagging this — honestly a fair thing to ask.

Let me be upfront: I didn't build this off my own DeerAPI key. I was filling a gap — DeerAPI is already a supported provider on the Python side (merged a while back in #10303, by the DeerAPI folks themselves), but the new Go API server didn't have a driver for it yet. So this PR just brings Go up to parity with what Python already does. I also called out in the description that I couldn't do a live end-to-end run from my environment and was leaning on CI + the shared OpenAI-compatible pattern the other drivers use.

On the "stopped registering" part — I poked at it just now, and the API itself still seems alive:

- `GET https://api.deerapi.com/v1/models` → `401` (just the normal "you forgot your key" response)
- `https://deerapi.com` → `301`

So the endpoint the driver talks to is still up. What you're describing sounds more like **they've closed new sign-ups**, which is different from the service being dead — and that part I genuinely can't confirm, so you may well be right.

I'll leave the call to you:

- **If new registration is really closed** → let's hold this one, and we should probably look at the Python provider too, since it'd be in the same spot.
- **If existing keys still work** (the API answering suggests they might) → merging keeps Go and Python consistent.

Either way I'm happy to follow your lead — close it, hold it, or tweak anything else. Just let me know 🙏

PR #15159, opened 2026-05-22
feat[go]: implement provider: TokenHub

Dimon0000000:
### What problem does this PR solve?

implement provider TokenHub

### Type of change

- [x] New Feature (non-breaking change which adds functionality)

PR #15224, opened 2026-05-26
fix(tokenhub): wire Go driver and harden requests

JSONbored:
## Summary
- Wire the Go TokenHub provider through the model factory.
- Harden TokenHub request handling for chat, streaming, embeddings, and model listing.
- Add focused TokenHub unit coverage for factory wiring and provider behavior.

## What changed
- Register `tokenhub` in `ModelFactory` so configured TokenHub instances create `TokenHubModel` instead of falling through to the dummy driver.
- Keep `ChatWithMessages` on the non-streaming path even when a shared chat config contains `stream: true`.
- Validate required API key, model, and sender inputs before sending TokenHub requests.
- Remove raw SSE line logging from streaming responses.
- Guard TokenHub model-list response parsing against malformed rows or unexpected response shapes.

## Why
TokenHub provider configuration and the driver were added in #15159, but the factory path still did not instantiate the TokenHub driver. This completes that wiring and adds regression coverage for the Go provider work tracked under #14736.

## Validation
- `go test ./internal/entity/models -run TestTokenHub -count=1 -v`
- `git diff --check`
- CodeRabbit review on the uncommitted diff; follow-up test coverage was added and the final review reported no findings.
- Codex Security diff review found no new issues in the TokenHub changes.

## Notes
- Refs #14736.
- Follows up #15159.

### DIFF
--- a/internal/entity/models/factory.go
+++ b/internal/entity/models/factory.go
@@ -103,6 +103,8 @@ func (f *ModelFactory) CreateModelDriver(providerName string, baseURL map[string
 		return NewHunyuanModel(baseURL, urlSuffix), nil
 	case "tokenpony":
 		return NewTokenPonyModel(baseURL, urlSuffix), nil
+	case "tokenhub":
+		return NewTokenHubModel(baseURL, urlSuffix), nil
 	case "novita":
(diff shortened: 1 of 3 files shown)


Haruko386 (approved):
LGTM

PR #15227, opened 2026-05-26
fix(rag): mark task FAILED on parser miss and retry transient MinerU errors (#15176)

### LINKED ISSUE #15176: [Feature Request]: Better error handling for MinerU parser
### Self Checks

```Markdown

```

Add more robust error handling when using MinerU as document parser:
1. add retry on error
2. currently if mineru server is down, the parse process will wrongly complete with status `DONE` and chunks=`0` and `MinerU not found` in the progress log. At least the status should be `FAIL` for later attention

_No response_

```Markdown

```

### Additional information

_No response_

Rene0422:
## Summary

Fixes #15176. Two related improvements for the MinerU parser path
(also covers Docling / OpenDataLoader / PaddleOCR for the first
fix, since they share the same fallback shape):

1. **FAIL vs DONE** — when MinerU is down, the task no longer
   silently completes with `progress=1.0` and `chunks=0`. It now
   ends in the FAILED state with a clear error message.
2. **Retry on transient MinerU errors** — connection errors,
   timeouts, and 5xx responses are retried with exponential
   backoff instead of failing the whole task on the first hiccup.

## Root cause (issue 1)

When MinerU is unreachable, `by_mineru` in `rag/app/naive.py`
correctly called `callback(-1, "MinerU not found.")` — which writes
a failure to the task DB — and returned `(None, None, None)`.

But the calling `chunk()` then did:

```python
if not sections and not tables:
    return []
```

…making "parser failed" indistinguishable from "successfully
parsed an empty doc." Downstream, `task_executor.build_chunks` saw
the empty list and hit:

```python
if not chunks:
    progress_callback(1., msg=f"No chunk built from {task_document_name}")
    return
```

And `TaskService.update_progress` explicitly allows `prog >= 1`
to overwrite a `-1` ("allow recovery from -1" —
`api/db/services/task_service.py:308`). So the prior failure was
clobbered. The task ended up DONE with 0 chunks and "MinerU not
found" lingering in the progress log — exactly what the reporter
described.

## Fix 1 — FAIL vs DONE (`rag/app/naive.py`)

Distinguish parser-failure (returns `sections=None and
tables=None`) from successful-but-empty (returns `[], []`).
Raise `RuntimeError` on the failure case so it propagates into
`build_chunks`'s existing chunking-exception handler at
`rag/svr/task_executor.py:318-321`, which already does the right
thing: calls `progress_callback(-1, "Internal server error while
chunking: ...")` and re-raises. The outer task handler then also
writes `-1`. The task ends FAILED with a clear, surfaced error.

Because Docling / OpenDataLoader / PaddleOCR all follow the same
`(None, None, …)` convention on failure, this single guard
fixes them all.

The raised message uses the lowercased layout-recognizer name
(the `PARSERS` dict key — `mineru`, `docling`, `opendataloader`,
`paddleocr`) so the executor's `[Exception]:` message clearly
identifies which parser failed.

## Fix 2 — retry on transient MinerU errors (`deepdoc/parser/mineru_parser.py`)

Wrap the `/file_parse` `requests.post` call in a retry loop.

- Retries are limited to genuinely transient failures:
  `requests.ConnectionError`, `requests.Timeout`, and HTTP 5xx.
- 4xx, malformed-response, and OS errors raise on the first
  attempt — retrying them wastes minutes per task.
- Exponential backoff between attempts.
- Configurable via env vars:
  - `MINERU_MAX_ATTEMPTS` (default `3`)
  - `MINERU_BACKOFF_SECONDS` (default `2`)
- Each retry emits a `callback(0.20, "... retrying in Xs")` so
  the UI shows what's happening rather than appearing to stall.

Only new import: `time`.

## Test plan

- [ ] Stop the MinerU container. Trigger a PDF parse with MinerU
      selected. Verify the document ends up **FAILED** (not DONE)
      and the progress log shows `[Exception]: ... [mineru]
      parser is unavailable or failed; see progress log for
      details.` plus the earlier `[ERROR] MinerU not found.`.
- [ ] With MinerU running normally, parse a PDF that genuinely
      produces no extractable content — the task should still
      end DONE with the "No chunk built from ..." progress
      message (unchanged behavior).
- [ ] With MinerU running, parse a normal PDF — chunks should be
      built as before, no retry log messages.
- [ ] Simulate a transient MinerU outage (e.g. start with the
      service down, then bring it up within the backoff window).
      Verify retries are attempted, the UI shows the retry
      progress messages, and the task ultimately succeeds.
- [ ] Simulate a 401/403 from MinerU. Verify it fails on the
      first attempt with no retries.
- [ ] Repeat the FAIL-on-miss check with Docling, OpenDataLoader,
      and PaddleOCR selected (whichever you have configured) —
      they share the failure-signaling convention, so they
      should also end FAILED rather than silently DONE.
- [ ] Run `uv run pytest` to make sure no existing tests regress.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


JinHai-CN:
@Rene0422 would you please fix the conflicts?

PR #15231, opened 2026-05-26
Fix: duplicated parse document request

### LINKED ISSUE #15225: [Bug]: Duplicated parse document request
### Self Checks

main

### RAGFlow image version

0.25

```Markdown
CPU ubuntu python 3.13
```

### Actual behavior

1. create dataset
2. upload document
3. click to parse document

Sometimes the "Operation succeeded" message will be duplidated and the "enqueue" (for parsing) will happen 2 times

### Expected behavior

Only one time

### Steps to reproduce

```Markdown
1. create dataset
2. upload document
3. click to parse document

Sometimes the "Operation succeeded" message will be duplidated and the "enqueue" (for parsing) will happen 2 times
```

### Additional information

_No response_

nickmopen:

Closes #15225.

Clicking Parse on a freshly uploaded document sometimes triggered two parse requests — the "Operation succeeded" toast appeared twice and the document got enqueued twice.

`ReparseDialog` has a mount-time `useEffect` that auto-fires `handleOperationIconClick()` when the dialog renders with `hidden=true` (the zero-chunks-and-no-metadata case, or while a parse is already running). The downstream re-entry guard in `useHandleRunDocumentByIds` is backed by React state, which doesn't update synchronously between two back-to-back effect runs, so when the effect runs twice on the same mount both runs slip past the guard and the API is hit twice.

React 18 `StrictMode` (active in `web/src/main.tsx`) intentionally double-invokes effects in development to surface this kind of non-idempotent side effect, and any remount of the dialog produces the same duplicate.

Fix: latch the auto-fire with a `useRef` so it runs at most once per real mount. Refs persist across StrictMode's simulated unmount/remount; a genuine fresh mount reinitializes the ref so subsequent user-driven parses still fire.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


xugangqiang:
@nickmopen  May I ask for your help to provide testing evidence? Like log to prove that duplicate request is prevented somewhere.

nickmopen:
> @nickmopen May I ask for your help to provide testing evidence? Like log to prove that duplicate request is prevented somewhere.

@xugangqiang Thanks for the nudge — here's reproducible evidence beyond the unit test.

### Repro setup

- Local dev stack via `docker/docker-compose.yml`, frontend via `web/npm run dev` (Vite, StrictMode active — same code path as a genuine remount in production).
- Created a dataset, uploaded a small file, then clicked the parse ▶ on the document row (zero chunks, metadata disabled → the dialog mounts with `hidden=true` and the auto-fire `useEffect` runs on mount).
- Captured browser Network and the `ragflow-server` access log.

### Before — `main`

Each click produces **two** `POST /api/v1/documents/ingest` on distinct source ports within the same second:

[INFO] 172.20.0.1:41874 POST /api/v1/documents/ingest 1.1 200
[INFO] 172.20.0.1:41876 POST /api/v1/documents/ingest 1.1 200

Two success toasts also fired per click.


https://github.com/user-attachments/assets/1b1e03b2-9a8c-41d2-a40f-b6fa66722403



### After — this branch

The `autoFiredRef` latch prevents the second invocation. One toast, one POST per click:

[INFO] 172.20.0.1:36490 POST /api/v1/documents/ingest 1.1 200



https://github.com/user-attachments/assets/b640b631-b5f3-4a99-b30c-e69d6031f398



The unit test in `reparse-dialog.test.tsx` codifies this guarantee so it doesn't regress.

xugangqiang:
@nickmopen 
May I know the version/git commit for your testings? I can't re-produce with latest main.

nickmopen:
> @nickmopen May I know the version/git commit for your testings? I can't re-produce with latest main.

@xugangqiang

Here's the recipe I used to reproduce it deterministically on `main`. The trigger is React 18 StrictMode's intentional double-invoke of the dialog's mount `useEffect` (active in `web/src/main.tsx`), so it shows up on the **Vite dev server**, not on a `vite build` production bundle.

Tested on `main` at `7f79b8486`'s parent (i.e. main without this PR). Repro:

1. `docker compose -f docker/docker-compose.yml up -d` (backend + deps, v0.25.6 image)
2. `cd web && npm run dev` — `NODE_OPTIONS=--max-old-space-size=8192` if you run into OOM
3. Sign in, create a dataset, upload a small file (any 1-line `.txt` works)
4. Open DevTools → Network, filter `documents/ingest`, tick "Preserve log"
5. Click the ▶ icon on the row (state is `UNSTART`, `chunk_count=0`, `enable_metadata=false` — this puts the dialog into the `hidden=true` auto-fire path)

What I see: **two** identical `POST /api/v1/documents/ingest` per click, two "Operation succeeded" toasts. Backend access log from a single click (both lines within the same second, distinct source ports):

[INFO] 172.20.0.1:41874 POST /api/v1/documents/ingest 1.1 200
[INFO] 172.20.0.1:41876 POST /api/v1/documents/ingest 1.1 200

A couple of questions so we can match environments:
- Was your "can't reproduce" attempt against `npm run dev` or a built/served bundle?
- Was the document zero-chunk with metadata disabled (so the dialog mounts hidden)?
- Any console error / hot-reload skip that might have interfered with the StrictMode double-invoke?

Your original report said "**sometimes** the enqueue happens 2 times" in a v0.25 prod image — that intermittency suggests a same-instance re-fire of the same `useEffect` is the underlying mechanism, which the StrictMode case turns into a deterministic instance of the same path.

wangq8:
@cike8899 @dcc123456 please help to review this PR.

dcc123456 (commented):
Thank you very much for your interest in this project.

Regarding this bug, if you are testing it in a development environment, this is actually the expected behavior of Vite. However, if you are experiencing this in a production environment, could you please provide a screenshot and outline the steps to reproduce it?

nickmopen:
@dcc123456 Thanks for the review — you're right to push on this, and I want to give you a straight answer rather than overstate the case.

**On the mechanism, you're correct:** the only thing I can reproduce *deterministically* is React 18 StrictMode's double-invoke of the dialog's mount `useEffect`, and that only happens on the Vite dev server — it's gone in a `vite build` bundle. So as a user-visible production regression, I can't demonstrate one.

I dug further to be sure. I diffed the exact code that shipped in the reporter's **v0.25 image** against current `main` for the whole parse path:

- `reparse-dialog.tsx`, `parsing-status-cell.tsx`, `use-run-document.ts`, `main.tsx` — **byte-identical** between `v0.25.0` and `main`.

So when @xugangqiang says he can't reproduce on latest `main`, that's effectively the same code he filed against. Walking the production render path on it: clicking ▶ mounts the dialog once, the mount effect fires once, and because the table uses a stable `getRowId: (row) => row.id` and the dialog is `memo`'d, the 5s polling refetch re-renders but never remounts it. Single mount → single fire. There's no deterministic production double-fire path I can find.

**What I think is actually going on:** the original report was a real, intermittent ("sometimes") observation on a production image, and there *is* a genuine latent race behind it — the re-entry guard in `useHandleRunDocumentByIds` keys off React state (`currentId`) plus react-query's `loading`, both of which update asynchronously, so two invocations in the same tick both slip past `if (isLoading) return`. StrictMode is just the one way I can hit that window on demand. But I can't prove that's what the reporter hit, and I can't trigger it in a production build.

So I'd rather not sell this as a confirmed prod fix. Honest framing: it's **idempotency hardening for a non-idempotent mount effect** (the thing StrictMode exists to flag) plus a confirmed dev-experience fix, with a unit test pinning the contract. The change is a 4-line `useRef` latch — no behavior change on a normal single mount.

If that doesn't meet your bar for merging without a reproducible production symptom, that's a fair call and I'll close it. Alternatively, if you'd prefer to address the underlying race more directly, I'm happy to tighten the re-entry guard in `use-run-document.ts` (e.g. a synchronous in-flight ref) instead of — or in addition to — the dialog-level latch. Let me know which direction you'd like.

dcc123456:
@nickmopen 

Thanks for your patient reply.
This is the expected behavior in the development environment ,this is actually an inherent Vite behavior and does not need to be fixed.

PR #15234, opened 2026-05-26
fix(go-models): add VolcEngine model listing suffix

### LINKED ISSUE #14701: [Feature Request]: Implement ListModels in the VolcEngine Go driver
### Self Checks

```Markdown
Note: this issue was originally filed for Aliyun Balance. After verifying that the
DashScope endpoint specified in that earlier framing returns 404 (it does not exist
on DashScope, and account balance is exposed only via the separate BSS OpenAPI which
uses AccessKey + HMAC-SHA256 signing rather than the model API key), I am repurposing
this issue for a different concrete gap in the same Go-driver track.

The VolcEngine Go driver in internal/entity/models/volcengine.go ships a stub
ListModels method that always returns "volcengine, no such method". Other Go drivers
that have already landed in the recent Go-rewrite effort (DeepSeek, Gitee, NVIDIA,
OpenAI, SiliconFlow) all implement ListModels against the OpenAI-compatible /models
endpoint, and the model picker UI relies on it.

VolcEngine (Doubao / Ark) does expose a documented OpenAI-compatible models endpoint
that accepts the same Bearer token used for chat and embedding requests:

  GET https://ark.cn-beijing.volces.com/api/v3/models

Without ListModels, a tenant who configures VolcEngine has no in-product way to see
which endpoints are available on their Ark account, even though the chat and
embedding paths in the same driver are already working.

conf/models/volcengine.json also does not yet declare a "models" entry under
url_suffix, so even if the method body were filled in there is no URL to call.
```

Implement `VolcEngine.ListModels` in `internal/entity/models/volcengine.go` and add the
matching `models` URL suffix to `conf/models/volcengine.json` so the existing model
picker works for VolcEngine the same way it already works for the other drivers in this
package.

The driver should:

1. Resolve the region with a default-region fallback, the same way the surrounding
   VolcEngine methods already do.
2. GET `${BaseURL[region]}/${URLSuffix.Models}`, which resolves to
   `https://ark.cn-beijing.volces.com/api/v3/models` with the default region.
3. Send `Authorization: Bearer <api_key>` when an API key is configured. Mirror the
   existing NVIDIA `ListModels` behavior of omitting the header when the key is empty,
   so an unauthenticated caller still gets a clear `401 Unauthorized` from upstream.
4. Surface non-200 responses with the upstream status line and body, the same shape
   the other Go drivers use.
5. Parse the JSON response via the package-level `DSModelList` / `DSModel` types
   already used by DeepSeek, Gitee, and SiliconFlow, and return the list of model ids
   in input order. When the response includes `owned_by`, render the entry as
   `id@owned_by`, matching the convention of the other drivers.

`conf/models/volcengine.json` needs the `"models"` URL suffix added alongside the
existing `chat`, `files`, and `embedding` entries:

```json
"url_suffix": {
  "chat": "chat/completions",
  "files": "files",
  "embedding": "embeddings/multimodal",
  "models": "models"
}
```

No new public API. No interface change. No factory change.

_No response_

```Markdown
No new docs are needed. VolcEngine is already in the model provider picker, with chat
and embedding already working. This issue just makes the existing model-list UI return
real data for VolcEngine instead of an error, the same way it already returns real
data for DeepSeek, Gitee, NVIDIA, OpenAI, and SiliconFlow.

Use case: a tenant adds VolcEngine (Doubao / Ark) as a model provider in RAGFlow,
opens the model selector, and immediately sees the endpoints available on their Ark
account, without having to leave the RAGFlow UI to log into the Volcano Engine
console.
```

### Additional information

_No response_

oktofeesh1:
## Summary
- add the VolcEngine `models` URL suffix used by the existing Go `ListModels` implementation
- return a clear error when the VolcEngine models suffix is missing
- add focused VolcEngine model-listing regression tests

## What changed
- Added `url_suffix.models` to `conf/models/volcengine.json`.
- Normalized the configured models suffix before building the request URL.
- Covered config loading, successful model listing, upstream errors, and missing suffix handling.

## Why
`VolcEngine.ListModels` already builds requests from `URLSuffix.Models`, but the bundled VolcEngine config did not define that suffix. That left the model-listing path unable to call the documented `/models` endpoint from the existing provider config.

Fixes #14701

## Validation
- `go test ./internal/entity/models -run TestVolcEngine -count=1`
- `go test ./internal/entity -count=1`
- `git diff HEAD~1..HEAD --check`

## Notes
- `go test ./internal/entity/models -count=1` currently fails in unchanged Astraflow coverage: `TestAstraflowEmbedReturnsNoSuchMethod` panics before reaching any VolcEngine assertions.

### DIFF
--- a/conf/models/volcengine.json
+++ b/conf/models/volcengine.json
@@ -6,7 +6,8 @@
   "url_suffix": {
     "chat": "chat/completions",
     "files": "files",
-    "embedding": "embeddings/multimodal"
+    "embedding": "embeddings/multimodal",
+    "models": "models"
   },
   "class": "volcengine",
   "models": [
--- a/internal/entity/models/volcengine.go
+++ b/internal/entity/models/volcengine.go
@@ -551,8 +551,12 @@ func (z *VolcEngine) ListModels(apiConfig *APIConfig) ([]string, error) {
 	if baseURL == "" {
 		return nil, fmt.Errorf("volcengine: no base URL configured for region %q", region)
 	}
+	modelsSuffix := strings.Trim(strings.TrimSpace(z.URLSuffix.Models), "/")
+	if modelsSuffix == "" {
+		return nil, fmt.Errorf("volcengine: models URL suffix is not configured")
+	}
 
-	url := fmt.Sprintf("%s/%s", strings.TrimSuffix(baseURL, "/"), z.URLSuffix.Models)
+	url := fmt.Sprintf("%s/%s", strings.TrimSuffix(baseURL, "/"), modelsSuffix)
 
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
--- a/internal/entity/models/volcengine_test.go
+++ b/internal/entity/models/volcengine_test.go
@@ -1,0 +1,110 @@
+package models
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func newVolcEngineServer(t *testing.T, handler func(t *testing.T, r *http.Request, w http.ResponseWriter)) *httptest.Server {
+	t.Helper()
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		handler(t, r, w)
+	}))
+}
+
+func newVolcEngineForTest(baseURL string) *VolcEngine {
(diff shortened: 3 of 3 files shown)


Haruko386 (approved):
@oktofeesh1, wow, I did not even notice that

oktofeesh1:
> @oktofeesh1, wow, I did not even notice that Thank you for your contribution

happy to contribute - there are a few others I opened for similar issues I found as well

PR #15236, opened 2026-05-26
fix(go-models): add xAI model listing suffix

oktofeesh1:
## Summary
- add the xAI `models` URL suffix used by the existing Go `ListModels` implementation
- return a clear error when the xAI models suffix is missing
- add focused xAI model-listing and connection-check regression tests

## What changed
- Added `url_suffix.models` to `conf/models/xai.json`.
- Normalized the configured models suffix before building the request URL.
- Covered config loading, successful model listing, upstream errors, API-key validation, missing suffix handling, and `CheckConnection` delegation.

## Why
`XAIModel.ListModels` already builds requests from `URLSuffix.Models`, and `CheckConnection` delegates to that method. The bundled xAI config did not define that suffix, which left the model-listing path unable to call the provider `/models` endpoint from the existing provider config.

## Validation
- `go test ./internal/entity/models -run TestXAI -count=1`
- `go test ./internal/entity -count=1`
- `git diff HEAD~1..HEAD --check`

## Notes
- `go test ./internal/entity/models -count=1` currently fails in unchanged Astraflow coverage: `TestAstraflowEmbedReturnsNoSuchMethod` panics before reaching any xAI assertions.

### DIFF
--- a/conf/models/xai.json
+++ b/conf/models/xai.json
@@ -4,7 +4,8 @@
     "default": "https://api.x.ai/v1"
   },
   "url_suffix": {
-    "chat": "chat/completions"
+    "chat": "chat/completions",
+    "models": "models"
   },
   "class": "grok",
   "models": [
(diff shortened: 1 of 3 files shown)


Haruko386 (approved):
@oktofeesh1 LGTM

PR #15261, opened 2026-05-26
feat: add Go MCP server update API

Helios531:
## What

#15240
implementation for PUT /api/v1/mcp/servers/:mcp_id

## Changes

- Adds the Go implementation for `PUT /api/v1/mcp/servers/:mcp_id`.
- Wires MCP service and handler into the Go server/router for the update route.
- Preserves Python-style behavior for ownership checks, partial update fields, MCP type/name/URL validation, `headers`/`variables` normalization, and tool metadata scrubbing.

## Validation
- `gofmt`
- `git diff --check`
- `go test ./internal/dao -run TestDoesNotExist -count=1`
- `go test ./internal/handler -run TestDoesNotExist -count=1 -vet=off`
- `go test ./internal/router -run TestDoesNotExist -count=1 -vet=off`


internal/handler/mcp.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		jsonError(c, code, err.Error())
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"code":    common.CodeSuccess,
+		"message": "success",
+		"data":    result,
+	})
```
Hz-186, line 164: Returning the raw `entity.MCPServer` also makes the date format incompatible with the Python API.

The Go API currently returns `create_date` / `update_date` as RFC3339 strings with timezone, for example:

```json
"create_date": "2026-05-28T10:30:20+08:00"
```

The Python API returns these fields without timezone:

"create_date": "2026-05-28T10:30:20"

Please format these fields explicitly in a response DTO before returning the response, for example with t.Format("2006-01-02T15:04:05").

internal/handler/mcp.go
```diff
@@ -0,0 +1,66 @@
...
+	result, code, err := h.mcpService.UpdateMCPServer(user.ID, mcpID, req)
+	if err != nil {
+		jsonError(c, code, err.Error())
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"code":    common.CodeSuccess,
+		"message": "success",
+		"data":    result,
```
Hz-186, line 184: Hi, @Helios531 

Please avoid returning `entity.MCPServer` directly here.

When `description` is updated to `null`, the Python API keeps the field in the response:

```json
"description": null
```

But the Go API omits description entirely because the raw entity uses:

Description *string `json:"description,omitempty"`

Since Description is nil, omitempty removes the field during JSON serialization. To keep this endpoint compatible with
Python, please return a dedicated response DTO instead of the raw entity, and keep nullable fields such as description in
the JSON response.

For example:

```go
type MCPServerResponse struct {
    ID          string                 `json:"id"`
    Name        string                 `json:"name"`
    TenantID    string                 `json:"tenant_id"`
    URL         string                 `json:"url"`
    ServerType  string                 `json:"server_type"`
    Description *string                `json:"description"`
    Variables   map[string]interface{} `json:"variables"`
    Headers     map[string]interface{} `json:"headers"`
    CreateTime  *int64                 `json:"create_time"`
    CreateDate  string                 `json:"create_date"`
    UpdateTime  *int64                 `json:"update_time"`
    UpdateDate  string                 `json:"update_date"`
}
```

Then return the DTO here instead of result.

Hz-186 (changes requested):
hi, @Helios531

Could u help fix these compatibility issues in the Go MCP update endpoint?

Helios531:
Hi, @Hz-186, can you review my PR, plz?

Helios531:
hi, @Hz-186 , can you review my PR, pls?

Hz-186 (approved):
@Helios531, LGTM

JinHai-CN:
@Helios531  would you please fix the conflicts?

tmimmanuel (approved):
@JinHai-CN This is ready.

JinHai-CN:
@Helios531 @web-dev0521 Would you please resolve the conflicts?

PR #15266, opened 2026-05-27
fix(chat): sanitize NaN/Inf scores before serializing chat completions (#15245)

### LINKED ISSUE #15245: [Bug]: Failed to encode response: json: unsupported value: NaN (status code: 500)
### Self Checks

1234a567b

### RAGFlow image version

0.25.2

```Markdown

```

### Actual behavior

When I invoke the "converse with chat assistant" interface for streaming response, sometimes the following error occurs: "Failed to encode response: json: unsupported value: NaN (status code: 500)". Moreover, when asking the same question repeatedly, sometimes the response is normal, but sometimes the above error will be reported.

### Expected behavior

I hope to receive the results in a normal streaming manner and without any errors.

### Steps to reproduce

```Markdown
interface:api/v1/chat/completions
parameter:
{
    "messages":[
            {
                "role":"user",
                "content":"我的问题"
            }
        ],
    "stream":true,
    "chat_id":"my_chat_id",
    "session_id":"my_session_id"
}

When using the above interface information to make the call, the streaming interface occasionally returns an error: "Failed to encode response: json: unsupported value: NaN (status code: 500)"
The current embedding model in use is bge-m3 from ollama.And the parsed document is also completely parsed.
```

### Additional information

The specific log information is as follows:
data:{"code": 500, "message": "failed to encode response: json: unsupported value: NaN (status code: 500)", "data": {"answer": "**ERROR**: failed to encode response: json: unsupported value: NaN (status code: 500)", "reference": []}}
data:{"code": 0, "message": "", "data": true}

Rene0422:
## Summary

Fixes #15245 — `POST /api/v1/chat/completions` with `stream=true`
intermittently returns 500:

```
data:{"code": 500, "message": "failed to encode response: json:
unsupported value: NaN (status code: 500)", "data": {...}}
```

…even though "the same question" works on retry.

## Root cause

The streaming path serialized the answer with bare `json.dumps(...)`
(`api/apps/restful_apis/chat_api.py:1221`). `json.dumps` defaults to
`allow_nan=True` and emits the literal token `NaN` for NaN /
Infinity float values. That is valid Python-flavored JSON but
**invalid per RFC 8259**, so downstream consumers reject it. The
reporter's gateway is Go-based and the error wording
(`failed to encode response: json: unsupported value: NaN`) is
straight from Go's `encoding/json`.

How NaN gets into the payload: retrieval scoring in
`rag/nlp/search.py` runs `np.mean(...)` over aggregations that can
be empty, and similarity denominators can be zero. Reference chunk
fields like `similarity`, `vector_similarity`, `term_similarity`
can therefore be NaN depending on which chunks a given query
retrieves — which is exactly why the failure is intermittent for
the same question.

The non-streaming branch (`get_json_result(data=answer)`,
`chat_api.py:1243`) has the same vulnerability — Quart's `jsonify`
also defaults to `allow_nan=True` and the same retrieval pipeline
feeds both branches.

`agent/tools/exesql.py:88-102` already has the same NaN/Inf guard
for SQL results. This PR brings the chat completions path up to
parity.

## Fix

Add a small `_sanitize_json_floats(obj)` helper near the top of
`api/apps/restful_apis/chat_api.py`. It walks `dict` / `list` /
`tuple` and replaces any `float` that is `NaN` or `±Infinity` with
`None`. Apply it at the two serialization boundaries:

- **Streaming branch** (`stream()`): sanitize the SSE payload before
  `json.dumps`.
- **Non-streaming branch**: sanitize the `answer` dict before
  `get_json_result(data=...)`.

The terminal `data:True` frame and the `code:500` error frame carry
no scores and are left untouched.

Added `import math` to the existing alphabetical import block.

No change to retrieval logic — replacing NaN with `null` at the
serialization boundary is conservative: clients still parse the
JSON, a missing-score chunk is a strictly better failure mode than
a 500 that kills the whole reply.

## Test plan

- [ ] Reproduce the original failure: hit
      `/api/v1/chat/completions` with `stream=true` against a
      chat assistant whose dataset can produce a query with an
      empty chunk aggregation (any small dataset + a long query
      whose tokens don't match well). Before this PR the request
      sometimes 500s with the NaN message; after this PR the
      stream completes cleanly.
- [ ] Sanity-check the non-streaming path with `stream=false` on
      the same query — should also return 200.
- [ ] Confirm normal chats still return numeric `similarity` /
      `vector_similarity` / `term_similarity` in `reference.chunks`
      (no double-serialization, no `null` for healthy values).
- [ ] Force NaN into a chunk score with a debugger / monkeypatch
      and confirm the field comes back as `null` instead of
      crashing the request.
- [ ] Run `uv run pytest`.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -62,11 +62,18 @@ def _sanitize_json_floats(obj):
     term_similarity) can become NaN when an aggregation runs over an empty
     set or when a similarity denominator is zero, so the chat completions
     stream is the realistic trigger.
+
+    `isinstance(obj, float)` alone catches Python float and numpy.float64
+    (a float subclass) but misses numpy.float32 / numpy.float16 and any
+    other duck-typed numeric. Probe via math.isnan/isinf in a try/except
+    so any object math can evaluate gets sanitized — without changing
+    upstream callers like chunks_format or rag/nlp/search.py.
     """
-    if isinstance(obj, float):
+    try:
         if math.isnan(obj) or math.isinf(obj):
             return None
-        return obj
+    except TypeError:
+        pass
     if isinstance(obj, dict):
         return {k: _sanitize_json_floats(v) for k, v in obj.items()}
     if isinstance(obj, list):
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -16,6 +16,7 @@
 
 import json
 import logging
+import math
 import os
 import re
 import tempfile
@@ -49,6 +50,31 @@ from common import settings
 from common.misc_utils import get_uuid, thread_pool_exec
 from rag.prompts.generator import chunks_format
 from rag.prompts.template import load_prompt
+
+def _sanitize_json_floats(obj):
+    """Replace NaN/Infinity floats with None so the result is RFC 8259 JSON.
+
+    `json.dumps` emits the literal tokens `NaN`/`Infinity` by default
+    (allow_nan=True). Those tokens are valid Python JSON output but invalid
+    per the JSON spec, and downstream proxies / Go consumers reject the
+    response with `failed to encode response: json: unsupported value: NaN`
+    (fixes #15245). Retrieval scores (similarity, vector_similarity,
+    term_similarity) can become NaN when an aggregation runs over an empty
+    set or when a similarity denominator is zero, so the chat completions
+    stream is the realistic trigger.
+    """
+    if isinstance(obj, float):
+        if math.isnan(obj) or math.isinf(obj):
+            return None
+        return obj
+    if isinstance(obj, dict):
+        return {k: _sanitize_json_floats(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [_sanitize_json_floats(v) for v in obj]
+    if isinstance(obj, tuple):
+        return tuple(_sanitize_json_floats(v) for v in obj)
+    return obj
+
 
 _DEFAULT_PROMPT_CONFIG = {
     "system": (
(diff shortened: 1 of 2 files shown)


xugangqiang:
@Rene0422  May I know if you have conducted testing on your side? if so, please upload testing evidence. Also, is it possible to add unit test or end-2-end test?

Rene0422:
> @Rene0422 May I know if you have conducted testing on your side? if so, please upload testing evidence. Also, is it possible to add unit test or end-2-end test?

Hi @xugangqiang — the unit test is already in this PR at
`test/unit_test/api/apps/restful_apis/test_chat_completions_sanitize_floats.py`
(all 9 cases tagged `@pytest.mark.p1`).

It targets the **live source** of `_sanitize_json_floats` — extracted
from `chat_api.py` via the `ast` module and exec'd in an isolated
namespace exposing only stdlib `math`, so the test isn't a stale
copy and any future change to the sanitizer is exercised. Avoids
having to stub out the dozens of heavy imports `chat_api.py` pulls
in at module load (quart, peewee models, services, etc.).

Cases:

- `test_passthrough_for_healthy_values` — int / float / str / None /
  bool / list / dict pass through unchanged
- `test_nan_becomes_none` — top-level NaN → `None`
- `test_positive_and_negative_infinity_become_none` — parametrized
  over `inf` / `-inf`
- `test_nan_inside_dict_replaced` — sibling keys preserved verbatim
- `test_inf_inside_list_replaced` — surrounding elements preserved
- `test_deeply_nested_chunks_format` — the actual `reference.chunks`
  shape (list of dicts with `similarity` / `vector_similarity` /
  `term_similarity` scores, mixed NaN/Inf), which is the exact
  payload that triggered #15245
- `test_tuples_are_walked_and_returned_as_tuples` — container type
  preserved
- `test_numpy_float32_nan_caught_via_math_isnan` — `np.float64('nan')`,
  `np.float32('nan')`, `np.float16('nan')` all → `None`. Locks in the
  earlier review fix that swapped `isinstance(obj, float)` for a
  `math.isnan` probe so numpy scalar types are caught; uses
  `pytest.importorskip("numpy")` so CI doesn't gate on the optional
  dep
- `test_strings_are_not_treated_as_numeric` — `"NaN"` /
  `"Infinity"` stay as strings; verifies the `TypeError` swallow path

## Manual reproduction in dev

`POST /api/v1/chat/completions` with `stream=true` against a
dataset where retrieval scoring can yield `similarity=NaN` (small
dataset + a query whose tokens don't match well, so the cosine
denominator goes to zero).

**Before** the patch the SSE stream contains a frame like:

```
data:{"code": 500, "message": "failed to encode response: json:
unsupported value: NaN", "data": {"answer": "**ERROR**: ...",
"reference": []}}
```

**After** the patch the same stream completes cleanly with the
score serialized as `null` in `reference.chunks[*].similarity` /
`vector_similarity` / `term_similarity`, and the rest of the
payload is unchanged.

Happy to attach before/after captures (curl output + tee'd SSE log)
to the PR if useful — let me know.

PR #14958, opened 2026-05-15
Add Replicate chat provider

tmimmanuel:
## What
- Add Replicate as a chat provider backed by the documented predictions API
- Register Replicate in the Go model factory and provider config
- Support non-streaming chat through sync predictions, polling fallback, streaming through `urls.stream`, model listing, and connection checks

## Notes
- Uses `POST /v1/predictions` with Replicate model identifiers in `version`, which supports official and community model identifiers
- Maps RAGFlow messages into Replicate prompt-shaped inputs (`prompt`, optional `system_prompt`) and forwards common documented LLM inputs: `max_new_tokens`, `temperature`, `top_p`
- Preserves whitespace in SSE output chunks and emits RAGFlow `[DONE]` at stream completion

## Tests
- `go test -vet=off -run TestReplicate -count=1 ./internal/entity/models`
- `go test -vet=off -count=1 ./internal/entity/models`

Refs #14736

PR #15278, opened 2026-05-27
Go: implement Rerank in Replicate driver

sxxtony:

`ReplicateModel.Rerank` in `internal/entity/models/replicate.go` was a `"replicate, no such method"` stub. The chat path landed in #14958 and the embed path in #15073; rerank is the last major retrieval surface still missing on this provider.

Until this PR, a tenant who selected a Replicate reranker model got the sentinel error on every rerank call.

### What this PR includes

- `internal/entity/models/replicate.go` — real `Rerank` method backed by the same `createPrediction` + `waitForPrediction` plumbing the chat and embed paths use.
- `conf/models/replicate.json` — adds `yxzwayne/bge-reranker-v2-m3` at the version pinned in the model's `default_example` (`model_types: ["rerank"]`).

### Wire shape (verified against Replicate's openapi_schema)

Targets the canonical bge-reranker-v2-m3 schema. Both fields and the output shape are taken from the model's published `default_example`:

Request body:
```json
{
  "version": "7f7c6e9d…",
  "input": {
    "input_list": "[[\"query\", \"doc1\"], [\"query\", \"doc2\"]]"
  }
}
```

Response output (one float per `[query, doc]` pair, in the same order):
```json
[-2.306640625, -10.40625]
```

The driver also tolerates the wrapped variant `{"scores": [...]}` that `ninehills/bge-reranker-large` emits, so a tenant who pins that model still gets a working driver. Rejects mismatched cardinality (output length ≠ document count) and non-numeric score entries.

### Live test transcripts (api.replicate.com)

The Replicate account on the provided key has no billing credit, so I cannot run a full inference. Everything up to the billing gate works:

**T1 — Model metadata for `yxzwayne/bge-reranker-v2-m3`**
```
GET /v1/models/yxzwayne/bge-reranker-v2-m3
→ HTTP 200
→ version: 7f7c6e9d18336e2cbf07d88e9362d881d2fe4d6a9854ec1260f115cabc106a8c
→ Input properties:
    - input_list (type=string) "JSON string. Can be a list containing one query and one passage pair, or a list of such pairs."
→ Output:
    type: array
→ default_example.output:
    [-11.03125, -2.306640625, -10.40625]
```

**T2 — Driver's request reaches the prediction endpoint (route ✓, payload ✓, version ✓)**
```
POST /v1/predictions   (Prefer: wait=60)
body: {"version":"7f7c6e9d…","input":{"input_list":"[[\"capital of France?\",\"Paris is the capital of France\"]]"}}
→ HTTP 402  "Insufficient credit"
```
Replicate accepts the request — URL, version, and `input.input_list` shape all pass validation. Only the billing gate stops execution.

**T3 — Invalid version (control)**
```
POST /v1/predictions
body: {"version":"deadbeef…","input":{"input_list":"[]"}}
→ HTTP 422  "Invalid version or not permitted"
```
Proves Replicate validates the version up-front, so T2's 402 means the version is real and the request shape is correct.

### Design notes

- **Scores are raw, not normalized.** Replicate's reranker output is a raw similarity score from the underlying model — not a probability in `[0, 1]` like Cohere's or Voyage's rerank responses. The driver passes the value through unchanged so callers can compare against per-model thresholds, but `RerankResult.RelevanceScore` should NOT be assumed to be `[0, 1]`. Documented inline.
- **`top_n` handling.** When `RerankConfig.TopN > 0`, the driver sorts by score descending (stable on `Index` for ties) and truncates to `TopN`. Otherwise returns all documents in original order — matching the Voyage driver's behavior.
- **Reuses prediction plumbing.** `createPrediction` + `waitForPrediction` are battle-tested by the chat (#14958) and embed (#15073) paths. The only new logic in this PR is input/output shaping (`replicateRerankInput`, `replicateRerankOutputToScores`, `replicateScoresFromInterface`).
- **Two output shapes supported** — canonical flat `[score, score, ...]` and the wrapped `{"scores": [...]}` variant.
- **Catalog uses `owner/name:version`** — bge-reranker-v2-m3 is a community model (not on Replicate's official-models list), so the bare owner/name path 404s on `/v1/models/.../predictions`. The existing `predictionEndpoint` routes anything containing `:` through `/v1/predictions` with the `version` field, which is the path that returned 402 in T2.

### What I deliberately did NOT do

- **No unit tests** — same convention as the merged chat (#14958) and embed (#15073) PRs; the shared prediction plumbing is exercised in production by the chat path.
- **No full end-to-end rerank inference** — the test account hit Replicate's billing gate (HTTP 402) on every prediction. The 402/422 differential is the closest verification available without funding the account, identical to the Embed PR's approach.
- **Support for the `documents` + `query` schema (yxzwayne/jina-reranker-v1-turbo-en)** — that model returns its output as a URI to a file rather than inline scores. Implementing that download-and-parse flow belongs in a follow-up; this PR targets the canonical inline-scores schema.

### DIFF
--- a/conf/models/replicate.json
+++ b/conf/models/replicate.json
@@ -29,6 +29,13 @@
       "model_types": [
         "embedding"
       ]
+    },
+    {
+      "name": "yxzwayne/bge-reranker-v2-m3:7f7c6e9d18336e2cbf07d88e9362d881d2fe4d6a9854ec1260f115cabc106a8c",
+      "max_tokens": 8192,
+      "model_types": [
+        "rerank"
+      ]
     }
   ]
 }
--- a/internal/entity/models/replicate.go
+++ b/internal/entity/models/replicate.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"sort"
 	"strings"
 	"time"
 )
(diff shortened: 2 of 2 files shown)


Haruko386 (approved):
LGTM

PR #15279, opened 2026-05-27
fix(api): return empty documents when doc_ids filter is empty

huang-aoqin:
When document metadata filtering resolves to an empty doc_ids set (no metadata hit), the document listing API may still return dataset documents instead of an empty result.
This PR fixes that by explicitly treating doc_ids=[] as a no-match condition and returning [] with total 0, so metadata no-hit queries no longer fall back to unfiltered results.


JinHai-CN:
@huang-aoqin would you please rebase the main branch?

PR #15292, opened 2026-05-27
Feature: Allow page_size max value 100

wangq8:
Feature: Allow page_size max value 100


zz80900:
Because you set the maximum value to 100, the MCP error 0 was triggered: [TextContent(type='text', text='Field: <page_size> - Message: <Value error, page_size must be less than or equal to 100> - Value: <1000>', annotations=None, meta=None)]

PR #15296, opened 2026-05-27
fix: skip array_avg for multi-dimension arrays in OceanBase/SeekDB connector

shiguangqi559:

  When using SeekDB or OceanBase as the doc engine, the knowledge graph (graphrag) extraction fails with a MySQL error
  1235: "array_avg with multi-dimension array not supported". This occurs because the `position_int` column is defined
  as `ARRAY(ARRAY(Integer))` (a 2D array storing coordinates), and `array_avg()` in MySQL/SeekDB does not support
  multi-dimensional arrays.

  The fix skips `array_avg` calls for nested ARRAY columns in the ORDER BY clause, while 1D arrays (`page_num_int`,
  `top_int`) continue to work correctly. This is consistent with:
  - The OpenSearch connector (`opensearch_conn.py`), which already only sorts by `page_num_int` and `top_int`
  - The task executor (`task_executor.py`), which sorts chunks in Python using only `page_num_int` and `top_int`

  ### Type of change


hnwyllmm:
I reviewed this PR again against the current `main` (`82e4147b8`). There are still useful fixes here, but I do not think the PR is mergeable as-is.

1. **The branch is stale and now conflicts with `main`.** The conflict is in `rag/utils/ob_conn.py`. In addition, two parts of this PR have already landed independently:
   - the `chunk_order_int` schema/migration fix was merged in #16997 (`05c67ca85`);
   - the `SearchResult(total, chunks)` handling in `DocMetadataService._search_metadata` was added by #17703.

   Please rebase and drop those duplicate changes first so the remaining review surface is clear.

2. **The primary fix avoids the exception by silently dropping a requested sort key.** `Dealer.chunk_list(sort_by_position=True)` explicitly requests `page_num_int -> position_int -> top_int`, and this path is used by GraphRAG, RAPTOR, and advanced-RAG document reads. The new nested-array branch simply skips `position_int`. That prevents `ARRAY_AVG(ARRAY(ARRAY(...)))`, but it also changes document reading-order semantics. Please define the intended OceanBase/SeekDB fallback explicitly and cover it with a realistic multi-chunk ordering test. If page/top is the canonical order, the caller/contract should say that rather than having one backend silently ignore a requested field.

3. **The current Go OceanBase/SeekDB engine still has the same nested-array failure.** In `internal/engine/oceanbase/search.go::buildOrderBy`, every chunk array field is wrapped in `ARRAY_AVG`; `position_int` is still classified as an array even though its schema is two-dimensional. After rebasing onto current `main`, this PR would fix only the Python connector while the Go path remains vulnerable to the same request. Please either cover both implementations or explicitly narrow the PR to the legacy Python path and open a paired Go fix with parity tests.

4. **Unknown sort fields should not be silently ignored.** The new `if field not in column_types: continue` changes an invalid/unsupported sort request into a successful but differently ordered result. A debug log does not make that semantic degradation visible to callers. Please reject unsupported fields with a clear error, or add an explicit mapping for the supported logical field. This behavior is also unrelated to the nested-array bug and would be easier to review separately.

5. **Dropping `row_id()` breaks the selected-field contract.** Retrieval requests `row_id()` and later publishes it as `row_id`. Removing it from `output_fields` guarantees `None` on OceanBase. The current Go OceanBase adapter maps `row_id()` to the physical identifier instead. The Python connector should likewise provide a defined alias/fallback, or this change should include evidence and tests showing that the field can safely disappear.

6. **The `get_scores` additions are valuable but unrelated and untested.** They address a separate second-pass KNN failure. Please split them into a focused PR and add a test proving that a vector-only OceanBase search returns the real non-zero `_score` mapping through `Dealer._knn_scores`. The memory-store version also needs a demonstrated caller/contract or should be omitted from that fix.

This PR currently changes three production files without adding regression tests, and its only repository test check was skipped. My suggested path is:

- rebase on current `main` and remove the already-merged changes;
- split the nested-array ordering fix, `row_id()` mapping, and `get_scores` support into focused changes;
- add Python and Go parity tests for `position_int`, while confirming that 1-D `page_num_int`/`top_int` ordering still uses the supported aggregation;
- add a real score-preservation test for `get_scores`.

The reported `ARRAY_AVG` failure is valid, but the current patch needs these scope and semantic issues resolved before merge.

JinHai-CN:
@shiguangqi559 
If you are still working on the PR, please re-open it.

@hnwyllmm I assume @shiguangqi559 won't be finishing this PR. If you think any part of it is worth keeping, feel free to open a separate PR.

PR #15295, opened 2026-05-27
feat(go-api): implement tenant member management (issue #15294)

Lobster-0429:
## Summary

Ports the Python `tenant_api` team/member management endpoints to Go, adding 4 endpoints under `/api/v1/tenants/:tenant_id/`:

- `GET /tenants/:tenant_id/users` — list non-owner members with user details (owner only)
- `POST /tenants/:tenant_id/users` — invite a user by email; creates invite-role join record (owner only)
- `DELETE /tenants/:tenant_id/users` — remove a member by `user_id`; owner can remove anyone, members can remove themselves
- `PATCH /tenants/:tenant_id` — accept a pending invitation, transitioning role `invite → normal`

## Changes

| File | Change |
|------|--------|
| `internal/dao/user_tenant.go` | Add `GetMembersByTenantID` (JOIN with user table), `DeleteByUserAndTenant`, `UpdateRoleByUserAndTenant` |
| `internal/service/tenant.go` | Add `ListMembers`, `AddMember`, `RemoveMember`, `AcceptInvite`; role constants; add `userDAO` for email lookup |
| `internal/handler/tenant.go` | Add `ListTenantMembers`, `AddTenantMember`, `RemoveTenantMember`, `AcceptTenantInvite` |
| `internal/router/router.go` | Register 4 routes in the existing `tenants` group |
| `internal/service/tenant_test.go` | Unit tests for auth checks and input validation (no DB required) |

## Permission model

Mirrors Python exactly:
- List/Add: `userID == tenantID` (owner check)
- Remove: `userID == tenantID` (owner) OR `userID == targetUserID` (self-removal)
- Accept invite: any authenticated user (acts on their own membership)

Closes #15294

PR #15281, opened 2026-05-27
feat(go-api): implement MCP server management endpoints

Lobster-0429:
## Summary

Ports the MCP (Model Context Protocol) server management endpoints that power  `web/src/pages/user-setting/mcp/` from Python (`api/apps/restful_apis/mcp_api.py`) to Go. There were no MCP routes in the Go server before this change.

Closes #15275 (subtask of #15240).

## Endpoints implemented (base path `/api/v1`)

| Method | Path | Description |
|--------|------|-------------|
| GET | `/mcp/servers` | List tenant servers (keyword / order / pagination) |
| POST | `/mcp/servers` | Create a server |
| GET | `/mcp/servers/{mcp_id}` | Get one (`?mode=download` exports config) |
| PUT | `/mcp/servers/{mcp_id}` | Update a server |
| DELETE | `/mcp/servers/{mcp_id}` | Delete a server |
| POST | `/mcp/import` | Bulk import from JSON config |
| POST | `/mcp/servers/{mcp_id}/test` | Connect + list tools (see notes) |

## Implementation

Follows the existing `handler → service → dao` layering (per PR #14790):

- **entity** (`internal/entity/mcp.go`): added `MCPServerType` constants and `IsValidMCPServerType` over the existing `MCPServer` model.
- **dao** (`internal/dao/mcp.go`): new `MCPServerDAO` with tenant-scoped CRUD, a keyword filter, and a **whitelisted order-column map** (guards against SQL injection via the caller-supplied `orderby`).
- **service** (`internal/service/mcp.go`): new `MCPService` — list/get/export/create/update/delete/import/test — mirroring
  `MCPServerService` and the `mcp_api` request validation, with sentinel errors for clean code mapping.
- **handler** (`internal/handler/mcp.go`): new `MCPHandler` with the seven handlers and Python-compatible response codes.
- **router / server_main**: registered the `/mcp` group and wired the handler.

## Deviations from Python (documented in code)

1. **Bulk import is at `POST /mcp/import`, not `/mcp/servers/import`.** gin (v1.9.1) cannot register a static segment and a path param at the same tree node, so `/mcp/servers/import` would collide with `/mcp/servers/:mcp_id` and panic at startup. The frontend should call `/mcp/import`.
2. **No live tool discovery on create/update/import.** The Python path runs `get_mcp_tools` over SSE / streamable-HTTP and stores `variables.tools`. The Go server has no MCP client yet, so these persist `variables`/`headers` but leave `variables.tools` unpopulated.
[...]

PR #15276, opened 2026-05-27
Go: implement memory messages APIs

tmimmanuel:
### What problem does this PR solve?

Implements the Go API slice for memory messages from #15240:

- `POST /api/v1/messages`
- `GET /api/v1/messages`

The PR wires the routes, replaces the previous placeholder handlers, stores raw memory messages with embeddings, lists recent messages from accessible memories, and adds message-index field handling for Elasticsearch and Infinity-backed searches.

### Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

### Tests

```bash
go test -vet=off ./internal/service -run '^$'
go test -vet=off ./internal/handler ./internal/router ./internal/engine/elasticsearch ./internal/engine/infinity
git diff --check
```

PR #15274, opened 2026-05-27
feat(go-api): implement connector (data source) management endpoints

Lobster-0429:
## Summary

Ports the connector (data source) management endpoints that power `web/src/pages/user-setting/data-source/` from Python
(`api/apps/restful_apis/connector_api.py`) to Go. Previously only `GET /connectors` (list) was implemented in Go; this adds the rest of the lifecycle.

Closes #15273 (subtask of #15240).

## Endpoints implemented

All under base path `/api/v1` (mirrors the Python routes):

| Method | Path | Description |
|--------|------|-------------|
| POST | `/connectors/{connector_id}/test` | Validate stored credentials |

`GET /connectors` (list) was already present and is unchanged.

## Implementation

Follows the existing `handler → service → dao` layering (per PR #14790):

- **entity** (`internal/entity/connector.go`): added the `SyncLogs.task_type` column (selected by the Python sync-log query) and `ConnectorTaskType` / `InputType` constants mirroring `common.constants`.
- **dao** (`internal/dao/connector.go`): added `SyncLogsDAO` (join-based `ListByConnectorID`, a `HasScheduled` guard, status updates, and delete) and `Connector2KbDAO`; reuses the existing `ConnectorDAO` CRUD.
- **service** (`internal/service/connector.go`): added `Create/Get/Update/Delete/ListLogs/Rebuild/Test`, the `accessible()` tenant
  check, and `cancelTasks`/`scheduleTasks` helpers mirroring `ConnectorService.cancel_tasks` / `schedule_tasks`. Sentinel errors map cleanly to the Python response codes.
- **handler** (`internal/handler/connector.go`): added the seven handlers with the same response shapes/codes as Python (including PATCH's `{"data": {...}}` envelope unwrap).
- **router** (`internal/router/router.go`): registered the new routes under the `/connectors` group.

## Known gaps vs. Python (no Go task-worker yet)

These are documented in code comments as follow-ups:

- **rebuild** drops the existing sync logs and enqueues a reindex `SYNC` task, but does **not** delete previously indexed documents — `FileService.delete_docs` is not yet ported to Go. The re-sync will re-ingest from the source; stale docs
  are not removed here.
- **test** gates on `source == rest_api` and credential presence only. The per-connector `validate_config` path lives in the Python `common.data_source` package and has no Go equivalent yet.

Provider OAuth web flows (`google/gmail/google-drive/box`) remain out of scope ppr the issue and should be a separate subtask.

## Testing

[...]

PR #15307, opened 2026-05-27
Go: implement system healthz API

tmimmanuel:
## Summary
- Add Go REST support for `GET /api/v1/system/healthz`.
- Return Python-compatible `ok`/`nok` dependency fields for DB, Redis, document engine, and storage.
- Return HTTP 200 only when all checks pass; otherwise return HTTP 500 with `_meta` failure details.
- Add focused service coverage for the unhealthy dependency response when Go dependencies are not initialized.

## Scope
This is a small, isolated slice of #15240. It avoids current open connector PRs (#15274, #15300, #15265, #15264), tenant/member PRs (#15295, #15301, #15276), MCP PRs (#15281, #15253, #15254, #15260, #15261, #15262), and the memory-message PR (#15256).

Refs #15240

## Tests
```bash
go test -vet=off ./internal/service -run TestSystemServiceHealthzReportsUnhealthyDependencies -count=1
go test -vet=off -run '^$' ./internal/service ./internal/router\ngo build ./internal/handler ./internal/service ./internal/router\ngit diff --check\n```\n\nNote: I did not use `go test ./internal/handler` because the current upstream snapshot has an unrelated `internal/handler/connector_test.go` compile error referencing `service.UpdateConnectorRequest`; production handler code is covered by `go build` above.

### DIFF
--- a/internal/handler/system.go
+++ b/internal/handler/system.go
@@ -55,6 +55,16 @@ func (h *SystemHandler) Health(c *gin.Context) {
 	})
 }
 
+// Healthz reports dependency health in the Python-compatible format.
+func (h *SystemHandler) Healthz(c *gin.Context) {
+	result, allOK := h.systemService.Healthz(c.Request.Context())
+	statusCode := http.StatusOK
+	if !allOK {
+		statusCode = http.StatusInternalServerError
+	}
+	c.JSON(statusCode, result)
+}
+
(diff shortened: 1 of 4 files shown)


Haruko386 (approved):
@tmimmanuel LGTM
Thank you for you contribution

PR #15339, opened 2026-05-28
fix(go-models): harden ModelScope default transport handling

glorydavid03023:
## Summary
- Harden `NewModelScopeModel` to avoid panics when `http.DefaultTransport` is a custom non-`*http.Transport` RoundTripper.
- Add `TestModelScopeNewModelWithCustomDefaultTransport` regression coverage.

## Test plan
- [x] `go test ./internal/entity/models/ -run TestModelScope -count=1` (cannot run locally: `go` is unavailable in this environment)


Haruko386 (approved):
LGTM

PR #15341, opened 2026-05-28
fix(go-models): harden Voyage default transport handling

glorydavid03023:
## Summary
- Harden `NewVoyageModel` to avoid panics when `http.DefaultTransport` is a custom non-`*http.Transport` RoundTripper.
- Add `TestVoyageNewModelWithCustomDefaultTransport` regression coverage.

## Test plan
- [x] `go test ./internal/entity/models/ -run TestVoyage -count=1` (cannot run locally: `go` is unavailable in this environment)


Haruko386 (approved):
@glorydavid03023 LGTM

PR #15351, opened 2026-05-28
fix(go-models): harden N1N default transport handling

glorydavid03023:
## Summary
- Harden `NewN1NModel` to avoid panics when `http.DefaultTransport` is a custom non-`*http.Transport` RoundTripper.
- Add `n1n_test.go` with coverage for name/factory plus `TestN1NNewModelWithCustomDefaultTransport`.

## Test plan
- [x] `go test ./internal/entity/models/ -run TestN1N -count=1` (cannot run locally: `go` is unavailable in this environment)


Haruko386 (approved):
@glorydavid03023 LGTM

Haruko386:
@glorydavid03023 would you fix the CI error?

PR #15178, opened 2026-05-24
Go: harden Jina driver; document unsupported stream chat

glorydavid03023:
### What problem does this PR solve?

The Go Jina driver lacked clear documentation that **streaming chat is unsupported** upstream, and `ListModels`/`CheckConnection` did not send the API key. An earlier attempt to implement `ChatStreamlyWithSender` failed in maintainer testing (`stream chat` → HTTP 500; non-stream chat works).

### Type of change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

### Changes

- `internal/entity/models/jina.go`: document unsupported streaming; harden `NewJinaModel` transport; per-request timeouts on `Embed`/`Rerank`/`ListModels`; `ListModels` sends `Authorization`; nil guards on `Embed`/`Rerank`; fix `Rerank` `top_n` when config is nil
- `internal/entity/models/jina_test.go`: replace stream happy-path tests with `TestJinaStreamNotSupported`

### Note for reviewers (@Haruko386)

Agreed — Jina does not expose working stream chat today. This PR **does not** call `stream=true`; it returns an explicit error from `ChatStreamlyWithSender` and keeps non-stream chat/embed/rerank behavior.

### Test plan

- [x] `go test ./internal/entity/models/ -run TestJina -count=1`

PR #15380, opened 2026-05-29
fix(go-models): use per call context timeouts so long streaming responses are not truncated

### LINKED ISSUE #15379: [Bug]: Streaming chat responses are cut off after 120 seconds across many Go model providers
### Actual behavior

Most Go model providers in `internal/entity/models/` share an `http.Client` that sets `Timeout: 120 * time.Second`, and the same client is reused for `ChatStreamlyWithSender`. In Go, `http.Client.Timeout` is a hard ceiling on the whole exchange, and it explicitly includes the time spent reading the response body:

```go
// net/http documentation
// The timeout includes connection time, any redirects, and reading the response body.
```

Because a streaming chat response keeps the body open for the entire generation, any stream that lasts longer than 120 seconds is severed in the middle with a timeout error. This breaks long answers, reasoning model outputs, and any slow or overloaded upstream provider.

### Expected behavior

A streaming chat response should keep streaming for as long as the upstream keeps sending data, bounded only by a generous limit, never by a fixed 120 second wall. Non streaming calls (chat, embed, rerank, list models, and similar) should still fail fast under a short deadline.

### Steps to reproduce

1. Configure any affected provider (for example DeepSeek, Moonshot, Zhipu AI, vLLM, Ollama, SiliconFlow).
2. Trigger a chat through the streaming path (`ChatStreamlyWithSender`) that produces output for more than 120 seconds, such as a long answer or a slow reasoning model.
3. Watch the stream end abruptly around the 120 second mark with a client timeout error instead of finishing normally.

### Additional information

Root cause: the shared `http.Client.Timeout` covers reading the response body, so it doubles as a streaming wall clock. The providers that already work correctly (`groq`, `mistral`, `voyage`, `anthropic`) set no client `Timeout` and instead wrap each request in a `context.WithTimeout`: a short deadline for non streaming calls and a long one for streams.

Affected providers (about 29): `302ai`, `aliyun`, `baichuan`, `baidu`, `cohere`, `deepinfra`, `deepseek`, `fishaudio`, `gitee`, `huaweicloud`, `huggingface`, `lmstudio`, `mineru`, `mineru_local`, `minimax`, `moonshot`, `nvidia`, `ollama`, `openrouter`, `orcarouter`, `paddleocr`, `paddleocr_local`, `qiniu`, `siliconflow`, `tokenhub`, `vllm`, `volcengine`, `xunfei`, `zhipu-ai`.

Dexterity104:

Closes #15379 

Around 29 Go model providers in `internal/entity/models/` share an `http.Client` configured with `Timeout: 120 * time.Second`, and reuse that same client for `ChatStreamlyWithSender`. Go's `http.Client.Timeout` is a hard ceiling on the whole request that also covers reading the response body, so it behaves as a wall clock on streaming. Any streamed chat response that lasts longer than 120 seconds gets cut off in the middle with a timeout error. Long generations, reasoning model outputs, and slow or overloaded upstreams are the common victims.

The providers that already behave correctly (`groq`, `mistral`, `voyage`, `anthropic`) set no client `Timeout` and instead wrap each request in a `context.WithTimeout`. This change converges the affected providers onto that same pattern.

**What changed**

- Removed the blanket `Timeout` field from the shared `http.Client` in every affected provider.
- Wrapped every request in a per call `context.WithTimeout` with a matching `defer cancel()`, switching to `http.NewRequestWithContext`.
- Centralized three shared, test overridable timeout values in `xai.go`:
  - `nonStreamCallTimeout` (120s) for chat, embed, rerank, list models, check connection, balance, and task submit or poll.
  - `streamCallTimeout` (10m) for streaming chat and streaming audio senders, so a long stream is bounded generously instead of at 120 seconds.
  - `longOpCallTimeout` (10m) for heavy file work: OCR, document parse, and audio transcription or synthesis.
- Folded the five pre-existing per file `*StreamTimeout = 10 * time.Minute` constants (`groq`, `perplexity`, `avian`, `ppio`, `togetherai`) into the shared `streamCallTimeout` so there is a single source of truth.

**Notes for reviewers**

- Pre-existing per call literals were converged onto the shared knobs. The short 30s embed and rerank deadlines (`aliyun`, `gitee`, `lmstudio`, `nvidia`, `ollama`, `vllm`) now use `nonStreamCallTimeout` (120s), and the 30s or 120s OCR and parse deadlines now use `longOpCallTimeout`.
- `paddleocr` `OCRFile` now bounds its whole submit, poll, and download flow with one `longOpCallTimeout` context, which also caps a poll loop that was previously unbounded.
- `volcengine` `Embed` issues one request per text inside a loop, so its context is scoped in a small closure to avoid piling up deferred cancels across the batch.
- `jina` (#15178) and `jiekouai` (#15337) are intentionally left untouched and still carry their client `Timeout`, since both are covered by their own in flight PRs.

**Tests**

- Added `timeout_test.go` with two regression tests using `httptest` servers: a streaming response that keeps emitting past a shortened non streaming deadline still completes intact, and a slow non streaming call still trips its short deadline. The timeout knobs are package vars so the tests can shrink them to milliseconds instead of waiting two real minutes.
- The `models` package test binary did not compile because `roundTripperFunc` was declared in four `*_test.go` files in the same package. The duplicates were removed and one copy kept in `modelscope_test.go`, which is required for the package tests to build and for the new tests to run.

```bash
go build ./internal/entity/models/
go test ./internal/entity/models/ -run 'StreamNotTruncated|NonStreamHonors|Groq|...'
```

### Type of change


Dexterity104:
Hi, @JinHai-CN @Haruko386

Haruko386 (approved):
@Dexterity104 LGTM

JinHai-CN:
@Dexterity104 would you please fix the conflicts?

Dexterity104:
Hi @JinHai-CN I've fixed the merging conflicts. Could you please review again?

Dexterity104:
Hi @JinHai-CN
I'm not sure if I have to fix this CI error on my PR because my PR is Go-only changes.
It fails on two pre-existing tests (`test_epub_parser.py`, `test_dataflow_service.py`) with `LookupError: Resource 'punkt_tab' not found`. The unit-test step runs `run_tests.py` without provisioning NLTK data, so `punkt_tab` is missing.

Issue #15356, opened 2026-05-28
[Bug]: `/openai/{chat_id}/chat/completions` defaults `stream` to true (incompatible with OpenAI API)


dearsishs:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

8468227a1ab4ba8c66ba33c6c83b2b5ac7773b5d

### RAGFlow image version

0.25.6

### Other environment information

```Markdown
Ubuntu 22.04
```

### Actual behavior

The OpenAI-compatible endpoint treats a missing `stream` field as **`true`**:

```165:165:ragflow/api/apps/restful_apis/openai_api.py
    stream_mode = req.get("stream", True)
```

Clients using standard OpenAI SDKs often **omit** `stream` for non-streaming calls. The [OpenAI Chat Completions API](https://platform.openai.com/docs/api-reference/chat/create) defaults `stream` to **false**, returning a single JSON object.

With RAGFlow, the same payload returns **`text/event-stream`** SSE instead of `application/json`, breaking clients that expect a `ChatCompletion` object.

### Expected behavior

- Default `stream` to **`false`** when the field is omitted (match OpenAI).
- Minimum fix: `stream_mode = bool(req.get("stream", False))`.

Note: internal `/chat/completions` uses different conventions; only this OpenAI shim inverts the default.

### Steps to reproduce

```Markdown
Set `BASE_URL` to your RAGFlow API root (e.g. `http://localhost:9380` from source install, or your Docker/nginx URL). Replace `<CHAT_ID>` and `<TOKEN>`.


# Omit "stream" — OpenAI clients do this for non-streaming calls
curl -sS -X POST "${BASE_URL}/api/v1/openai/<CHAT_ID>/chat/completions" \
  -H "Authorization: Bearer <TOKEN>" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "model",
    "messages": [{"role": "user", "content": "hello"}]
  }' -D -


[...]

PR #15394, opened 2026-05-29
Fix: default OpenAI chat completions to non-stream

buua436:

default OpenAI chat completions to non-stream when `stream` is omitted
https://github.com/infiniflow/ragflow/issues/15356
### Type of change


xugangqiang:
LGTM

Issue #12214, opened 2025-12-25
[Bug]: Opensearch knn vector type


dontepanlin:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

55c0468ac94a7f8616815cacec1ae36aa85e1a42

### RAGFlow image version

nightly

### Other environment information

```Markdown

```

### Actual behavior

```
opensearchpy.exceptions.RequestError: RequestError(400, 'search_phase_execution_exception', "failed to create query: Field 'q_384_vec' is not knn_vector type.")
```

### Expected behavior

Retrieve some infromation

### Steps to reproduce

```Markdown
DOC_ENGINE=opensearch

version": {
    "distribution": "opensearch",
    "number": "2.19.1"

TEI_MODEL=BAAI/bge-small-en-v1.5
```

dontepanlin:
> The error happens because RAGFlow's OpenSearch mapping configuration does not include a dynamic template for fields ending with `_384_vec`, so `q_384_vec` is not mapped as a `knn_vector` type. Only vector fields with suffixes like `_512_vec`, `_768_vec`, etc., are supported out of the box. To fix this, add a dynamic template for `_384_vec` to your `conf/os_mapping.json` mapping file, specifying `type: knn_vector`, `dimension: 384`, and the appropriate `space_type`. After updating the mapping, re-create your index so OpenSearch recognizes `q_384_vec` as a `knn_vector` field. [Reference](https://github.com/infiniflow/ragflow/blob/1812491679fdd87ce7b8981e7a70daf29fadd1b5/conf/os_mapping.json#L7-L256)
> 
> _To reply, just mention [@dosu](https://go.dosu.dev/dosubot)._
> 
[...]

Issue #10402, opened 2025-10-02
[Bug]: When using opensearch as vector db, any query raise: 'search_phase_execution_exception', "failed to create query: Field 'q_4096_vec' is not knn_vector type.


miemiekurisu:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

9996

### RAGFlow image version

v0.20.5

### Other environment information

```Markdown
Docker version 28.0.4, build b8034c0
```

### Actual behavior

When I update to new 0.20.5 version from old 0.20, and find the new vetor db backend opensearch.
I change the docker/.env file to try the new db, and reindex all docs
Everythink went well, until I try to search in the search interface, it returns 400 in the front page.
I saw error log below in the ragflow-server:

`Traceback (most recent call last):
  File "/ragflow/rag/utils/opensearch_conn.py", line 250, in search
    res = self.os.search(index=indexNames,
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/client/utils.py", line 176, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/client/__init__.py", line 2364, in search
    return self.transport.perform_request(
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/transport.py", line 455, in perform_request
    raise e
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/transport.py", line 416, in perform_request
    status, headers_response, data = connection.perform_request(
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/connection/http_urllib3.py", line 308, in perform_request
    self._raise_error(
  File "/ragflow/.venv/lib/python3.10/site-packages/opensearchpy/connection/base.py", line 315, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
[...]

PR #15395, opened 2026-05-29
fix(mappings): register 3072 and 3584 vector dimensions

ef-rintaro:

Knowledge bases built with 3072-dim (OpenAI `text-embedding-3-large`) or
3584-dim (BGE-M3 / Qwen) embeddings fail at query time. Chunks insert fine, but
the KNN query is rejected:

OpenSearch: `search_phase_execution_exception: Field 'q_3072_vec' is not
knn_vector type.`

**Root cause.** Both mapping files enumerate one `dynamic_templates` entry per
dimension. `conf/os_mapping.json` lists 512/768/1024/1536/2048/4096/6144/8192/
10240 -- **3072 and 3584 are missing** -- so those vectors fall through to a plain
`float[]` and the KNN query fails. `conf/mapping.json` (Elasticsearch
`dense_vector`) is worse: it stops at 1536, so the same models break on ES too.
This is the same class of gap previously fixed for other dims in #10494
(see also #10402, #12214).

### Fix

Add `*_3072_vec` and `*_3584_vec` templates to both files:
- `conf/os_mapping.json` -> `knn_vector`, `dimension: 3072 / 3584`
- `conf/mapping.json` -> `dense_vector`, `dims: 3072 / 3584`

Both values are within the Elasticsearch 8.x `dense_vector` limit (max 4096) and
within OpenSearch's `knn_vector` limit.

### Type of change

### Affected backends
Elasticsearch and OpenSearch (additive mapping templates). Infinity is unaffected
(it stores vectors differently).

### How to reproduce
1. Create a KB with `text-embedding-3-large` (3072) or a 3584-dim model.
2. Parse a document, then run retrieval -> KNN query rejected (`not knn_vector type`
   on OpenSearch; analogous dense_vector failure on ES).

### Risk & backward compatibility
Additive only -- no existing template changes, no reindex required for existing
KBs. New dims (3072/3584 <= 4096) are valid for ES 8.11.3 and OpenSearch.

### Note / follow-up
The enumerated approach still omits other dims (e.g. 384, #12214) and the ES
file lags the OpenSearch one (no 2048/4096/6144/8192/10240). This PR fixes the
two reported dims; a broader parity pass could be a separate change.

### Test plan
- [ ] ES + 3072 model: KB parse + retrieval succeeds.
- [ ] OpenSearch + 3072 / 3584 model: KB parse + retrieval succeeds.
- [ ] Existing 1024/1536 KBs: unchanged.


xugangqiang:
@KevinHuSh  I believe you should be the most suitable one to review this PR.

conf/mapping.json
```diff
@@ -199,6 +199,28 @@
           }
         }
       },
+      {
```
qinling0210, line 202: Add `*_2048_vec` and `*_4096_vec` to `mapping.json`

qinling0210 (changes requested):

PR #15404, opened 2026-05-29
Feat: agentic RAG flow with @tool decorator and RAGTools agent

KevinHuSh:
## Summary

Introduces an **agentic RAG** flow driven by a lightweight `@tool` decorator that lets plain Python callables be registered as LLM tools without hand-writing OpenAI function schemas, plus a `RAGTools` orchestrator class that exposes a curated set of retrieval / answering tools and a dynamic system prompt.

## Core pieces

- **`rag/llm/tool_decorator.py`** — `@tool` decorator + `FunctionToolSession` adapter.
  - Builds OpenAI function schemas from type hints and docstrings.
  - Parses **both** reST `:param name:` and Google `Args:` styles, with multi-line continuation in each.
  - Enforces `request_timeout` via `asyncio.wait_for` (with the documented caveat that sync callables in `thread_pool_exec` cannot be interrupted mid-flight).
  - Bound methods on instances carry the `_is_tool` / `openai_schema` attributes through attribute proxy, so `RAGTools` can register `self.method` directly.

- **`rag/advanced_rag/agentic_rag.py`** — `RAGTools` class with:
  - `formalize_question` — rewrite follow-up messages into self-contained questions.
  - `select_documents` — LLM picks doc IDs from a `docID / title` catalogue.
  - `filter_docs_by_metadata` — `gen_meta_filter` + ES/Infinity push-down, gated on per-tenant metadata config.
  - `search_knowledge_bases` — keyword retrieval by default; `using_embedding=True` switches to dense retrieval (vector_weight=0.7) for the recommended one-shot retry pattern. Validates `docid_scope` against the catalogue with one targeted ``WHERE id IN (...) AND kb_id IN (...)`` query so LLM-hallucinated 32-char hex IDs cannot silently zero out retrieval.
  - `search_structured_data` — translates question → SQL via `dialog_service.use_sql` for KBs whose `parser_config` carries a `field_map`. Extends the shared `self.kbinfos` accumulator so structured + unstructured chunks share the citation pool.
  - `web_search` — Tavily fallback, only registered when a `Tavily` instance is provided.
  - `get_citation_guidelines` — single-shot fetch of the rendered `citation_prompt()` to seed the final answer.
  - `sys_prompt()` — composes the system instruction dynamically so the LLM never sees instructions for tools that aren't actually bound. Includes hard rules against fact / identifier hallucination and a strict language-matching rule (final answer follows the user's question language, retrieved evidence may stay in source language for verbatim citations).

## Supporting changes

- **`api/db/services/dialog_service.py`** — new `rag_agent` flow that drives a chat session through `RAGTools`.
- **`api/db/joint_services/tenant_model_service.py`** — scope the global LLM fallback query by `llm_factory` in both `get_model_config_by_id` and `get_model_config_by_type_and_name`. Without this scoping, a homonymous LLM row from a different factory could shadow `is_tools=True` for the configured provider and produce a spurious "Model … does not support tool call" warning. Mirrors the existing factory-aware behaviour in `TenantLLMService.get_model_config`.
- **`rag/llm/chat_model.py`**:
  - `LiteLLMBase.max_rounds` default `5 → 15` (agent loops with 7+ tools regularly need more than 5 rounds).
  - Streaming guard: skip empty `delta.content` to avoid yielding empty strings between reasoning chunks.
- **`rag/prompts/generator.py`** — `kb_prompt(start_idx=0)` so chunks accumulated across multiple retrieval calls get the right citation ID offsets.

## Tests

- **`test/unit_test/rag/llm/test_tool_decorator.py`** (11 cases, all passing):
  - Schema generation: required vs optional from defaults, Google-style `Args:`, multi-line reST `:param:` folding.
  - Session dispatch: sync via `thread_pool_exec`, async direct.
  - Defensive: unknown tool name raises, non-mapping arguments raise, non-`@tool` callable rejected at registration.
  - `asyncio.TimeoutError` raised when `request_timeout` is exceeded.

## Test plan

- [x] `python -m pytest test/unit_test/rag/llm/test_tool_decorator.py` — 11 passed
- [ ] Smoke-test the agent flow end-to-end against a SiliconFlow chat model with a real KB
- [ ] Verify behaviour with a tool-incapable chat model (`is_tools=False`) — should warn and fall back to non-tool streaming via `LLMBundle.bind_tools`'s existing gate
- [ ] Verify embedding retry path (`using_embedding=True`) against a hybrid (sparse + dense) KB
- [ ] Verify `search_structured_data` against a spreadsheet KB with `field_map`


yuzhichang:
This PR currently has merge conflicts with `main` and cannot be auto-synced.

Please run locally:

```bash
git fetch origin main
git rebase origin/main
# or
git merge origin/main
```

After resolving the conflicts, push to the current head branch and CI will re-run the review checks.

> Triggered by reviewer [yuzhichang](https://github.com/yuzhichang) (request-for-action)

PR #15418, opened 2026-05-29
feat(go): implement GET /api/v1/system/oceanbase/status

ai-mountain:

Part of #15240 (rewriting the RAGFlow API server in Go).

Implements `GET /api/v1/system/oceanbase/status`, porting `oceanbase_status` from `api/apps/restful_apis/system_api.py` to the Go API server.

OceanBase is not a supported document engine in the Go server — config loading rejects `DOC_ENGINE=oceanbase` as "not implemented", so there is no OceanBase connection to probe. The handler therefore mirrors the Python behavior:

- when OceanBase is **not** the configured engine, it returns the same `"OceanBase is not in use."` error the Python API surfaces (code `500`);
- when OceanBase **is** requested, it reports a `timeout` status, matching Python's connection-failure path.

The route is registered under the authenticated `/api/v1/system` group, matching the Python `@login_required` decorator.

### Type of change


Hz-186:
The normalization looks necessary, but I'm not sure about this code. I'll have my colleague @xugangqiang review it.

xugangqiang:
Blocking issue

  1. Unused cache import — won't compile

  internal/service/system.go adds "ragflow/internal/cache" to the import block, but nothing in this PR references the cache package. GetOceanBaseStatus only uses os, strings, and common. Go rejects unused
  imports, so the build fails with:

  imported and not used: "ragflow/internal/cache"

  Fix: drop the "ragflow/internal/cache" line.

  Style / quality

  2. Handler doesn't follow the file's response convention

  Every other handler in internal/handler/system.go returns responses through the common.* helpers (GetStatus → common.SuccessWithData, GetConfig → common.ResponseWithHttpCodeData, GetVersion → ditto).
  OceanBaseStatus hand-rolls the response:

  c.JSON(http.StatusOK, gin.H{
      "code":    code,
      "message": "success",
      "data":    data,
  })

  Since the service returns a common.ErrorCode that may be CodeSuccess or CodeServerError, the idiomatic equivalent is:

  common.ResponseWithHttpCodeData(c, http.StatusOK, code, data, "success")

  3. Missing tests

  internal/service/system_test.go exists and already tests Healthz. GetOceanBaseStatus is a pure function of DOC_ENGINE and is trivial to test. Please cover the three branches: unset / "oceanbase" / other.
   This also matches the repo's "Preserve behavior with focused tests" guidance.

  4. PR base looks out of sync with main

  The patch's base import block for internal/service/system.go is missing common, entity, and storage, all of which are present on current main. Rebasing will likely conflict in the import block (notably a
   duplicate common import). Please rebase and confirm go build ./... passes.

  Behavioral fidelity vs. description

  The description is largely accurate. One nuance worth tightening:

  - The description says that when OceanBase is requested, the handler "reports a timeout status, matching Python's connection-failure path." Python actually attempts a real connection
  (OBConnection.health() runs SHOW VARIABLES LIKE 'version_comment'; get_performance_metrics() runs SELECT 1 and collects storage/processlist/QPS), and only returns timeout on failure. This PR returns a
  hardcoded timeout without any connection attempt — which is fine given Go has no OceanBase client, but it's a placeholder, not a mirror of Python's connection-failure path.

  - Side note: Python's alive branch is effectively unreachable too, because OBConnection.health() returns a dict without a status key, so health_info.get("status") == "healthy" is always False. So in
  practice both servers always surface timeout when DOC_ENGINE=oceanbase — the net behavior matches, just not for the reason the description implies.

  Also worth noting in the description: because internal/server/config.go:518-519 rejects DOC_ENGINE=oceanbase at config load, the Go server can't actually boot in oceanbase mode, so the timeout branch is
  unreachable at runtime — only the "not in use" branch can fire. The description acknowledges Go doesn't support OceanBase; might be worth stating the runtime unreachable consequence explicitly.

  What looks good

  - Route registration under the authenticated /api/v1/system group matches Python's @login_required. ✅
  - GetUser(c) auth check mirrors GetStatus. ✅
  - "not in use" branch: HTTP 200, code=500, and the Failed to get OceanBase status: OceanBase is not in use. message all match Python exactly. ✅
  - strings.ToLower(strings.TrimSpace(...)) is stricter than Python's os.getenv but consistent with settings.py's .lower() — acceptable.

  Verdict

  Approach is sound and the not-in-use path is faithful, but the unused cache import must be removed before this can build. After that, switching the handler to common.ResponseWithHttpCodeData, adding unit
   tests, and rebasing on main would make this ready to merge.

JinHai-CN:
@ai-mountain please re-open the PR, if you are still working on it.

PR #15422, opened 2026-05-29
Go: implement Google Vertex provider

tmimmanuel:

Implements the Google Vertex provider slice for #14736.

This adds a Go model driver registration and provider config for `Google Vertex`, using the Google GenAI SDK with the Vertex AI backend. It supports Vertex API-key mode, project/location mode, and the existing Python-compatible JSON key shape with `google_project_id`, `google_region`, and base64 `google_service_account_key`.

### Type of change

### Tests

```bash
go test -vet=off $(find internal/entity/models -maxdepth 1 -name '*.go' ! -name '*_test.go' | sort) internal/entity/models/google_test.go -run 'Google|Vertex' -count=1
go test -vet=off ./internal/entity -run 'GoogleVertexProviderConfig' -count=1
go test -vet=off ./internal/service -run 'TestNewModelDriverForBaseURL' -count=1
git diff --check
```

`go test -vet=off ./internal/entity/models -run 'Google|Vertex' -count=1` is currently blocked by an existing package-level duplicate test helper:

```text
internal/entity/models/ppio_test.go:15:6: roundTripperFunc redeclared in this block
internal/entity/models/novita_test.go:12:6: other declaration of roundTripperFunc
```


Haruko386:
Vertex (Google cloud )should be an independent model provider, you should not write it in google provider.
And before I review your code, would you plz provider your Vertex (Google cloud ) API reference Docs?

tmimmanuel:
Updated: Vertex AI is now split into an independent `GoogleVertexModel` provider instead of being routed through `GoogleModel`.

Vertex AI API reference used: https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.publishers.models/generateContent

tmimmanuel:
@Haruko386 Please review.

internal/entity/models/google_vertex.go
```diff
@@ -0,0 +1,394 @@
...
+			modelNames = append(modelNames, m.Name)
+		}
+		return googleModelPage{items: modelNames, nextPageToken: models.NextPageToken}, nil
+	})
+}
+
+// GoogleVertexModel implements ModelDriver for Google Cloud Vertex AI.
+type GoogleVertexModel struct {
+	BaseURL   map[string]string
+	URLSuffix URLSuffix
```
Haruko386, line 54: Where is `httpClient` ?

conf/models/google_vertex.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "Google Vertex",
+  "class": "gemini",
+  "models": [
```
Haruko386, line 4: where is URL and URL_suffix

Haruko386 (changes requested):
@tmimmanuel

internal/entity/models/google_vertex.go
tmimmanuel: There is no explicit `httpClient` field in this provider because it uses the official `google.golang.org/genai` SDK instead of issuing raw HTTP requests directly. `newGoogleVertexClient()` builds a `genai.Client`, and the SDK owns the HTTP transport/client internally.

So for Vertex this is different from providers that manually call `httpClient.Do(req)`. If you prefer all Go providers to keep an explicit reusable HTTP client field even when using an SDK, I can adjust it, but it is not required for the current SDK-based implementation.

conf/models/google_vertex.json
tmimmanuel: I omitted `url` / `url_suffix` because the Vertex implementation is SDK-driven. `googleVertexClientConfig()` sets `Project`, `Location`, credentials/API key, and optional `HTTPOptions.BaseURL`; then the `genai` SDK constructs the REST path for `generateContent`, `streamGenerateContent`, `embedContent`, and model listing.

Official Vertex REST reference: https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.publishers.models/generateContent

The documented endpoint is `POST https://aiplatform.googleapis.com/v1/{model}:generateContent`, where `model` is a fully qualified resource such as `projects/{project}/locations/{location}/publishers/google/models/{model}`. That is why this provider uses project/location config instead of static `url_suffix` values. If you prefer the config file to include a default Vertex base URL for clarity, I can add it.

Haruko386:
@tmimmanuel  Could you provide your test output in ragflow CLI?

Haruko386:
Hi @tmimmanuel I would be happy if you could provide your test output in ragflow CLI?

PR #15424, opened 2026-05-29
fix(llm): correct error handling, token accounting, and truncation in embedding providers

### LINKED ISSUE #15423: Embedding providers mask real errors, report fabricated token counts, and overshoot the truncation limit
### Summary

`rag/llm/embedding_model.py` defines about 40 embedding providers. Several of them share defects that hurt indexing reliability, cost accounting, and error visibility.

### Problems

**1. Embedding errors are masked and inconsistent (27 sites).**
Almost every provider follows this pattern:

```python
except Exception as _e:
    log_exception(_e, res)
    raise Exception(f"Error: {res}")
```

`common/log_utils.py::log_exception` always raises on its own, so `raise Exception(f"Error: {res}")` is dead code. The exception that actually surfaces depends on whether the SDK response object happens to expose a `.text` attribute, so the same failure produces different exceptions across providers and the response payload that explains the cause rarely reaches the caller.

**2. Token counts are fabricated.**
`LocalAIEmbed.encode` returns a hardcoded `1024`, and `OllamaEmbed` adds `128` per text. These values feed `used_tokens`, which drives tenant billing and usage tracking, so the recorded usage is fiction.

**3. The truncation limit overshoots the model ceiling.**
Several providers truncate inputs to `8196` tokens, which is above the standard `8192` ceiling. A chunk near the boundary can be truncated to a length the model still rejects, causing intermittent indexing failures.

**4. ZhipuEmbed and OllamaEmbed do not batch.**
Both send one HTTP request per text instead of batching like the other OpenAI compatible providers, so a document with N chunks triggers N sequential round trips.

### Impact

- Failures in the knowledge base pipeline are hard to diagnose because the real cause is hidden.
- Billing and usage metrics are inaccurate for LocalAI and Ollama.
- Boundary sized chunks fail to embed at random.
- Large documents embed slowly on Zhipu and Ollama.

Dexterity104:
### Summary

Closes #15423

`rag/llm/embedding_model.py` hosts about 40 embedding providers that shared several defects affecting indexing reliability, cost accounting, and error visibility. This PR fixes four concrete bugs.

**Masked, inconsistent errors (27 sites).** Nearly every provider ran `log_exception(_e, res)` followed by `raise Exception(f"Error: {res}")`. Because `log_exception` always raises, the second line was dead code, and the surfaced exception varied with whether the SDK response exposed a `.text` attribute. Every failure path now raises a single `EmbeddingError` that includes the underlying response detail, so the cause of a failed embedding is consistent and visible.

**Fabricated token counts.** `LocalAIEmbed` returned a hardcoded `1024` and `OllamaEmbed` added `128` per text. These values feed `used_tokens` and therefore billing and usage tracking. Both now report the real count from the API (Ollama `prompt_eval_count`, LocalAI `usage`) and fall back to a local token count only when the server omits it.

**Truncation overshoot.** The `8196` limit used by Mistral and Bedrock exceeded the standard `8192` ceiling and could push boundary sized inputs past the model limit. Limits are corrected to `8192` and made intentional per provider, and providers that rely on server side truncation now request it explicitly (Ollama `truncate=True`, Cohere `truncate="END"`).

**Missing batching on Zhipu and Ollama.** Both issued one request per text. They now batch like the other OpenAI compatible providers, turning N round trips into `ceil(N / batch_size)`. Batched results are realigned by response `index` so a chunk always keeps its own vector.

A shared `Base._batched_encode` helper owns the batch loop, optional truncation, result accumulation, and the single error path. It is the mechanism that lets these fixes live in one place instead of across 27 duplicated sites. The public `encode()` and `encode_queries()` contract stays the same, so existing callers are unaffected.

Tests covering all four fixes are added under `test/unit_test/rag/llm/test_embedding_model.py`.

### Type of change


Dexterity104:
Hi, @JinHai-CN @Haruko386

Dexterity104:
Hi, @JinHai-CN @Haruko386

Dexterity104:
Hi @JinHai-CN can you check this PR?

yingfeng:
Remove unnecessary files from this PR, thanks~

Dexterity104:
@yingfeng 
I'm really sorry. I removed those unnecessary files.

Lynn-Inf (changes requested):
I’d rather not change the return type of `VoyageEmbed.encode()` to `list` without also updating the callers that rely on it (e.g., for empty-result checks). Let's keep the `EmbeddingModel.encode()` interface consistent across all implementations.

Dexterity104:
@Lynn-Inf 
Thanks for catching this! I can see how the diff reads that way, but `VoyageEmbed.encode()` still returns an `np.ndarray`. The `list(...)` was inside the per-batch `_call` helper, not the `encode()` return, which goes through `Base._batched_encode` and wraps everything in `np.array(...)` (empty input still returns an empty `np.ndarray`). I've dropped the redundant `list()` to avoid the confusion.

Dexterity104:
@Lynn-Inf @yingfeng 
Could you please check this PR? Thanks!

Lynn-Inf:
LGTM

PR #15429, opened 2026-05-29
fix(rerank): normalize reranker scores onto a single scale before hybrid blend

### LINKED ISSUE #15428: Reranker scores are blended on incompatible scales, breaking retrieval ordering for several providers
### Describe the bug

The hybrid score in `rag/nlp/search.py` (`rerank_by_model`) blends the reranker output with token similarity using a fixed weighting that assumes both terms sit on the same `[0, 1]` scale:

```python
return tkweight * np.array(tksim) + vtweight * vtsim + rank_fea  # tkweight=0.3, vtweight=0.7
```

`vtsim` is the raw output of `rerank_mdl.similarity()`, but the reranker implementations in `rag/llm/rerank_model.py` return scores on very different scales and only three of them normalize:

- `NvidiaRerank` returns raw logits (`d["logit"]`), which are unbounded and frequently negative. With `vtweight=0.7`, a negative logit drags a genuinely relevant chunk below pure keyword matches, and a magnitude such as ±10 swamps `tksim`, which lives in `[0, 1]`. Reranking actively degrades results.
- Most providers (Cohere, Jina, Voyage, SiliconFlow, QWen, and others) return a `relevance_score` that is usually, but not guaranteed, in `[0, 1]`.
- Only `LocalAIRerank`, `OpenAI_APIRerank`, and `RAGconRerank` call the existing   `Base._normalize_rank`.

As a result, the same query produces differently scaled scores depending on the configured reranker, and logit based providers are outright broken.

### Additional context

`rerank_by_model` also contains twelve leftover `print("[DEBUG ...]")` statements introduced in #14231. They run on every retrieval, add per chunk overhead, and leak user queries, keywords, and document content to stdout and container logs.

### Expected behavior

Every reranker should feed the hybrid blend on a single, comparable `[0, 1]` scale regardless of which provider is configured, and no debug output should be emitted on the retrieval path.

cleanjunc:

Closes #15428

The hybrid score in `rag/nlp/search.py` (`rerank_by_model`) blends reranker similarity with token similarity on a fixed `[0, 1]` scale:

```python
return tkweight * np.array(tksim) + vtweight * vtsim + rank_fea  # tkweight=0.3, vtweight=0.7
```

The reranker implementations did not agree on that scale. Only three of roughly 17 providers normalized their output, and `NvidiaRerank` returned raw, unbounded logits. Weighted at `0.7`, a negative logit could push a genuinely relevant chunk below pure keyword matches, and its magnitude swamped `tksim`, which lives in `[0, 1]`. The practical effect was that the same query produced differently scaled scores depending on the configured reranker, and logit based providers degraded retrieval quality instead of improving it.

This PR enforces a single scoring contract in one place:

- `Base.similarity` is now the only public entry point. It short-circuits empty input and guarantees a normalized result. Each provider implements its raw scoring in `_compute_rank`, which removes sixteen duplicated empty input guards and the three scattered normalization calls.
- Normalization is range aware. Providers that already return calibrated `[0, 1]` relevance scores (Cohere, Jina, Voyage, and others) keep their absolute magnitudes, so `similarity_threshold` filtering and the reported `vector_similarity` stay meaningful. Only out-of-range output such as NVIDIA logits is min-max rescaled into `[0, 1]`.
- The twelve leftover `[DEBUG ...]` prints in `rerank_by_model`, introduced in #14231, are removed. They ran on every retrieval, added per chunk overhead, and leaked queries, keywords, and document content to stdout and logs.

A new regression suite in `test/unit_test/rag/llm/test_rerank_normalization.py` covers logit rescaling (positive, negative, and flat batches), preservation of already calibrated scores, ordering, empty input handling, and the per provider HTTP path. It also asserts that no provider overrides `similarity()`, so the contract cannot silently drift.

### Type of change

### DIFF
--- a/rag/llm/rerank_model.py
+++ b/rag/llm/rerank_model.py
@@ -48,7 +48,16 @@ class Base(ABC):
         if not query or not texts:
             return np.zeros(len(texts) if texts else 0, dtype=float), 0
         rank, token_count = self._compute_rank(query, texts)
-        return self._normalize_rank(np.asarray(rank, dtype=float)), token_count
+        rank = np.asarray(rank, dtype=float)
+        if rank.size:
+            logging.debug(
+                "Rerank %s scores before normalization: count=%d min=%.4f max=%.4f",
+                self.__class__.__name__,
+                rank.size,
+                float(np.min(rank)),
+                float(np.max(rank)),
+            )
+        return self._normalize_rank(rank), token_count
 
     def _compute_rank(self, query: str, texts: List) -> Tuple[np.ndarray, int]:
         """Provider-specific scoring. ``query`` and ``texts`` are non-empty."""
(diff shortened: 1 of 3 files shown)


cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:

qinling0210 (approved):
LGTM

PR #15434, opened 2026-05-29
fix(retrieval): keep rerank window aligned to page_size for deep pagination

### LINKED ISSUE #15433: [Bug] Reranked retrieval drops results and returns short pages on deep pagination
### Summary

When an external reranker is enabled, `Dealer.retrieval` in `rag/nlp/search.py` silently loses results and returns short pages once pagination moves past the first candidate block. It affects the common page sizes 10 and 30.

### Root cause

The candidate window is rounded up to a multiple of `page_size` so that block based pagination stays aligned, and then immediately clamped back to 64:

```python
RERANK_LIMIT = math.ceil(64 / page_size) * page_size if page_size > 1 else 1  # e.g. 70 for page_size=10
RERANK_LIMIT = max(30, RERANK_LIMIT)
if rerank_mdl and top > 0:
    RERANK_LIMIT = min(RERANK_LIMIT, top, 64)  # clamps back to 64 and breaks the alignment
```

`RERANK_LIMIT` is used both as the backend block size (`page = global_offset // RERANK_LIMIT`) and as the modulus that slices a page out of a reranked block (`begin = global_offset % RERANK_LIMIT`). For those two to agree, `RERANK_LIMIT` has to stay an exact multiple of `page_size`. The clamp to 64 breaks that for every `page_size` that does not divide 64, such as 10 and 30.

### Steps to reproduce

1. Enable an external reranker.
2. Run a retrieval with `page_size=10` and `top=1024`.
3. Request page 7 (global offset 60).

`RERANK_LIMIT` resolves to 64, so page 7 reads block 0 and slices positions 60 to 69. Block 0 only holds 64 candidates, so the page returns 4 results instead of 10, and the head of the next block (reranked positions 64 to 69) never appears on any page.

This happens every time the result set spans more than one block (`top` greater than the window), which is the default.

### Impact

Real, above threshold results disappear from reranked search and every page after the first comes back short. A simulation of the full retrieval path (per block rerank, similarity filter, and the exact pagination math) shows 16 valid results dropped for `page_size=10` and 66 dropped for `page_size=30` at `top=1024`.

### Expected behavior

Pagination over reranked results should surface every above threshold candidate exactly once, with no short interior pages.

cleanjunc:

Closes #15433

Reranked retrieval drops results and returns short pages once pagination crosses the first candidate block, for the common page sizes 10 and 30.

In `rag/nlp/search.py`, the candidate window (`RERANK_LIMIT`) is rounded up to a multiple of `page_size` to keep block based pagination aligned, and then clamped back to 64:

```python
RERANK_LIMIT = math.ceil(64 / page_size) * page_size if page_size > 1 else 1  # e.g. 70 for page_size=10
RERANK_LIMIT = max(30, RERANK_LIMIT)
if rerank_mdl and top > 0:
    RERANK_LIMIT = min(RERANK_LIMIT, top, 64)  # clamps back to 64, breaking the multiple
```

`RERANK_LIMIT` is used both as the backend block size (`page = global_offset // RERANK_LIMIT`) and as the modulus that slices a page out of a reranked block (`begin = global_offset % RERANK_LIMIT`). When it stops being a multiple of `page_size`, the block that gets fetched and the slice taken from it no longer agree. With `page_size=10` and `top=1024`, page 7 returns only 4 of 10 results and the head of the next block is never shown on any page. This happens whenever the result set spans more than one block, which is the default.

**Fix**

The window math is moved into a small reusable helper, `Dealer._rerank_window`, which:

- targets a pool of about 64 candidates,
- bounds it by `top` when a reranker is active, and
- always rounds to a whole number of pages, so the window stays an exact multiple of `page_size`.

The call site becomes a single line, and the alignment invariant now lives in one documented place. Behavior is unchanged on every path that was already aligned (the non reranked path and any `top` that already produced a page multiple).

**Verification**

A simulation of the full retrieval path (per block rerank, similarity threshold filter, and the exact `page // window` and `offset % window` math) confirms the fix loses nothing where the old code lost real results:

```
ps=10 top=1024:  new window=70  dropped_valid=0   |  old window=64  dropped_valid=16
ps=30 top=1024:  new window=90  dropped_valid=0   |  old window=64  dropped_valid=66
```

New unit tests in `test/unit_test/rag/test_search_pagination.py` cover the alignment invariant, cross block pagination (every candidate surfaced once, in order, no gaps, no short interior pages), the reported regression, and parity with the old window on the previously correct paths. All 114 cases pass and `ruff check` is clean.

Fixes the reranked deep pagination data loss described above.

### Type of change

### DIFF
--- a/rag/nlp/search.py
+++ b/rag/nlp/search.py
@@ -559,6 +559,31 @@ class Dealer:
                                            rag_tokenizer.tokenize(ans).split(),
                                            rag_tokenizer.tokenize(inst).split())
 
+    @staticmethod
+    def _rerank_window(page_size: int, top: int = 0) -> int:
+        """Candidate-window size shared by retrieval's block fetch and slice.
+
+        ``retrieval`` reuses this value BOTH as the backend block size and as
+        the modulus for extracting a single page from a (re)ranked block::
+
+            req["page"] = global_offset // window   # which block to fetch
+            begin       = global_offset %  window   # where the page starts
+
+        For those two to agree the window MUST be an exact multiple of
+        ``page_size``; otherwise blocks and pages drift apart and deep
+        pagination silently drops results and returns short pages.
+
+        The window targets a provider-friendly pool of ~64 candidates, bounded
+        by ``top`` when given (i.e. when an external reranker is active), and is
+        always rounded UP to a whole number of pages to preserve the invariant.
+        """
+        if page_size <= 1:
+            return min(30, top) if top > 0 else 30
+        window = math.ceil(64 / page_size) * page_size
+        if top > 0:
+            window = min(window, math.ceil(top / page_size) * page_size)
+        return window
+
     async def retrieval(
             self,
             question,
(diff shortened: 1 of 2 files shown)


cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:

qinling0210 (approved):
LGTM

PR #15436, opened 2026-05-29
fix(agent): add HTTP timeout to external API tools

### LINKED ISSUE #15435: Agent external API tools can hang forever when an upstream HTTP request stalls
### Describe the problem

Several agent tools call external HTTP APIs with no request timeout. When an upstream host accepts the TCP connection but never sends a response (slow or overloaded API, half open connection, stuck load balancer), the call blocks forever. Because these tools run inside agent canvas execution, one stalled socket hangs the entire agent run with no way to recover.

Affected call sites (10 total), all using `requests` with no `timeout`:

- `agent/tools/qweather.py` (4 calls)
- `agent/tools/jin10.py` (4 calls)
- `agent/tools/tushare.py` (1 call)
- `agent/tools/github.py` (1 call)

The `github.py` tool is wrapped with the `@timeout` decorator from `common/connection_utils.py`, but that does not help. In the default configuration (`ENABLE_TIMEOUT_ASSERTION` unset) the decorator waits on `result_queue.get()` with no timeout, and a daemon thread blocked inside a socket read cannot be interrupted, so the run still hangs.

### How to reproduce

Point any of these tools at a server that accepts the connection and never replies. The call never returns. Minimal reproduction:

1. Start a TCP server that accepts a connection and never writes a response.
2. Call `requests.get(url)` with no `timeout` against it.
3. Observe that the call blocks indefinitely.

Adding `timeout=` makes the same call raise `ReadTimeout` promptly.

### Expected behavior

External HTTP calls should be bounded by a timeout so a stalled upstream fails fast and surfaces an error to the agent instead of freezing the whole run.

cleanjunc:

Closes #15435 

Several agent tools call external HTTP APIs through `requests` with no request timeout. When an upstream host accepts the connection but never responds (a slow or overloaded API, a half open connection, a stuck load balancer), the call blocks forever. These tools run inside agent canvas execution, so a single stalled socket freezes the entire agent run with no recovery.

Ten call sites were affected:

- `agent/tools/qweather.py` (4 calls)
- `agent/tools/jin10.py` (4 calls)
- `agent/tools/tushare.py` (1 call)
- `agent/tools/github.py` (1 call)

The `github.py` tool already carried the `@timeout` decorator from `common/connection_utils.py`, but that does not protect against this case. In the default configuration the decorator waits on its result queue with no timeout, and a daemon thread blocked inside a socket read cannot be killed, so the run still hangs. The per request timeout added here is what actually bounds the call.

This is the same bug class as the merged Go stream timeout fix, surfacing in the Python tool layer.

Changes:

- Pass `timeout=DEFAULT_TIMEOUT` on all 10 calls, reusing the existing shared constant in `common/http_client.py` (configurable via `HTTP_CLIENT_TIMEOUT`) so there is one source of truth rather than scattered literals.
- Add an AST based unit test at `test/unit_test/agent/tools/test_http_timeout.py` that scans every tool module and fails if any `requests` or `httpx` request call omits a `timeout`, guarding current and future call sites.

Verification:

- Reproduced the indefinite block against a stalling local server, and confirmed that adding a timeout raises `ReadTimeout` promptly.
- Confirmed the `@timeout` decorator does not interrupt a blocked no timeout request in its default configuration.
- The new test flags exactly the 10 original call sites on the pre fix code and passes (22 modules) after the fix.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:
Hi @Haruko386 @JinHai-CN

cleanjunc:

cleanjunc:
Hi @qinling0210 
Thanks for your review my other 2 PRs. Can you please check this PR more? I really appreciate it a lot!

cleanjunc:
@yingfeng @qinling0210 

yuzhichang (commented):
I re-reviewed the current head. The timeout additions are correct, and the AST regression test now already covers both namespace calls and common session/client instances, so I do not see any remaining actionable issue in the current diff. From my side this is ready to merge.

PR #15455, opened 2026-05-31
fix(task-executor): reject non-positive LoopLocalSemaphore limits

pekar9781:

This PR prevents a task-executor stall class caused by misconfigured concurrency limits.

`LoopLocalSemaphore` previously accepted non-positive values, so setting `MAX_CONCURRENT_TASKS=0` (or similar) could make ingestion workers wait forever for permits. The failure mode looked like runtime stalling instead of clear config validation.

This change adds fail-fast validation and tests so invalid limiter values are rejected with a clear error message at initialization.

Closes #15454

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


xugangqiang:
pls resolve conflict

xugangqiang:
### Review findings (problems only)

**Blockers**
1. **Merge conflict / stale base.** `mergeStateStatus=DIRTY`, the branch is based on `cd18cfab7` (May 2026) and conflicts with current `main`. Conflict is purely cosmetic: `common/asyncio_utils.py:27-31` was reformatted from the parenthesized form to a single line. Please rebase onto `upstream/main`.

2. **Direction conflicts with the repo's established policy.** The repo already handles misconfigured concurrency/env vars by logging and falling back to a default instead of crashing at import time — see `common/doc_store/infinity_conn_base.py:66-69` (docstring: a bad value "would otherwise raise ValueError at module import time and take down every backend worker. We log and fall back to the default instead") and `common/settings.py:104` `_get_int_env`. This PR raises in `common/`, but all six `LoopLocalSemaphore` call sites are **module-level globals** (`rag/svr/task_executor_limiter.py:24-28`, `rag/graphrag/utils.py:42`). Raising turns "bad env var" into "the whole worker/agent fails to import", which is exactly the failure mode those helpers avoid.

3. **Root cause not fully covered.** The PR only guards `LoopLocalSemaphore`, but the same `MAX_CONCURRENT_*` env vars are read and used via bare `asyncio.Semaphore` elsewhere and still stall silently at 0:
   - `rag/svr/sync_data_source.py:90-91` + `:163` (`async with task_limiter`)
   - `agent/component/base.py:371` (`MAX_CONCURRENT_CHATS`, same var as `rag/graphrag/utils.py:42`)
   The "stall class" is therefore not prevented.

**Improvements**
4. **Validation is at the wrong layer.** `LoopLocalSemaphore` lives in `common/` and should not hardcode "Check MAX_CONCURRENT_* environment variables". Note `rag/svr/task_executor_limiter.py:28` `kg_limiter = LoopLocalSemaphore(2)` is a hardcoded constant with no env var — the message would be misleading. The error also does not name the offending variable. Prefer a `_positive_int_env(name, default)` helper at the env-parsing boundary that names the variable and logs, consistent with existing helpers.

5. **Missing positive-path tests.** Tests cover `0` and `-3` only; add a positive case (e.g. `LoopLocalSemaphore(1)` acquires/releases) and a test for the actual claimed scenario (`MAX_CONCURRENT_TASKS=0` import behavior).

6. **Broken "Closes" reference.** Body says `Closes #15454`, but `infiniflow/ragflow#15454` does not exist (404 for both issue and PR). The stated motivation cannot be verified; please fix the description.

7. **No docs for the changed behavior.** `MAX_CONCURRENT_TASKS`/`MAX_CONCURRENT_CHUNK_BUILDERS`/`MAX_CONCURRENT_MINIO` are not documented in `docker/` or `docs/`; the new fail-fast behavior needs a配套 note.

**Nits**
8. Add `@pytest.mark.p3` to the two new tests to match repo convention (`pyproject.toml:325-333`, 630+ existing uses). Note: CI (`run_tests.py -i`) does not filter by marker, so this is consistency-only.

**Top 3 to address first**
1. Align with maintainers: fail-fast raise vs. log+fallback — this decides whether the PR needs rewriting.
2. Move validation to the env-parsing sites (`rag/svr/task_executor_limiter.py:20-22`, `sync_data_source.py:90`, `agent/component/base.py:371`) with a variable-naming message, covering all paths.
3. Rebase onto `upstream/main` to resolve the conflict and fix the invalid `Closes #15454`.

PR #15462, opened 2026-06-01
feat(connectors): add Salesforce CRM data source connector

### LINKED ISSUE #15461: [Feature Request]: Salesforce data source connector
### Self Checks

```Markdown
RAGFlow has no connector for Salesforce. CRM records — Accounts, Contacts, Opportunities, Cases, and Knowledge articles — hold a lot of customer context that cannot currently be indexed into a knowledge base.
```

A Salesforce data source that queries selected objects via the Salesforce REST/SOQL API and serializes each record into a knowledge-base document, with incremental sync and configurable object selection through the data-source settings UI.

Net-new connector; the Airtable and RDBMS connectors are the closest references for record-style sources. Touchpoints:
1. `common/data_source/salesforce_connector.py` — `SalesforceConnector` extending `CheckpointedConnector`, using SOQL with `LastModifiedDate` ordering for paging and incremental sync.
2. `common/data_source/config.py` — add `DocumentSource.SALESFORCE = "salesforce"`.
3. `common/constants.py` — add `FileSource.SALESFORCE = "salesforce"`.
4. `common/data_source/__init__.py` — export `SalesforceConnector`.
5. `rag/svr/sync_data_source.py` — add `class Salesforce(SyncBase)` with `_generate()` and register it in `func_factory`.
6. `web/src/pages/user-setting/data-source/constant/index.tsx` — enum value, form fields (instance URL, OAuth client id/secret, objects to sync), defaults, display metadata + icon.
7. Tests.

```Markdown
Support and sales teams can ground answers in live CRM data — e.g., surfacing past Cases or Knowledge articles relevant to a customer question.
```

### Additional information

Auth via OAuth 2.0 (connected app); store/refresh the token through the credentials provider. Pairs naturally with a HubSpot connector under a shared CRM pattern.

Lobster-0429:

Closes #15461.

RAGFlow had no way to ingest Salesforce CRM data, so support / sales teams couldn't ground responses on live Accounts, Contacts, Opportunities, Cases, or Knowledge articles. This adds a first-class Salesforce data source connector that authenticates against a Connected App via OAuth 2.0 client-credentials, queries selected SObjects via SOQL, and turns each record into an indexable document with incremental sync.

**Highlights**
- `common/data_source/salesforce_connector.py`: new `SalesforceConnector` (`CheckpointedConnectorWithPermSync` + `SlimConnectorWithPermSync`).
  - OAuth 2.0 client-credentials flow; canonical `instance_url` from the token response so multi-pod orgs route correctly.
  - Per-object `SystemModstamp` cursor stored in `SalesforceCheckpoint.cursors` — a failure mid-object doesn't rewind sibling objects, and re-syncs only fetch changed rows.
  - Deterministic record-to-text formatter (sorted keys) so SOQL field reordering on the server doesn't mark every row "changed" on each poll.
  - `_get_json` raises on non-2xx so 429 / 5xx never silently advance the checkpoint past missing data.
  - `Knowledge__kav` is in the default object set but is skipped silently when the org doesn't have Salesforce Knowledge enabled (404 on describe).
  - Slim-doc IDs are scoped as `<Object>/<Id>` so prune deletes can't collide across object types.
- `common/constants.py`, `common/data_source/config.py`, `common/data_source/__init__.py`: register `salesforce` in `FileSource` / `DocumentSource` and export `SalesforceConnector`.
- `rag/svr/sync_data_source.py`: new `Salesforce(SyncBase)` class routed through `load_from_checkpoint` (poll_source would re-walk every object each run) and added to `func_factory`.
- Frontend:
  - `web/src/pages/user-setting/data-source/constant/index.tsx`: new `DataSourceKey.SALESFORCE`, form fields (instance URL, client ID/secret, objects, api_version, batch size), `syncDeletedFiles` capability, default form values, and tile entry with the new icon.
  - `web/src/locales/{en,zh}.ts`: description + per-field tooltips.
  - `web/src/assets/svg/data-source/salesforce.svg`: 48x48 brand-style icon to match the other Microsoft / cloud tiles.

**Verification**
- `npm run build` (vite + esbuild) passes (1m 26s).

### Type of change


Lobster-0429:
Hi, @yingfeng ,

yingfeng:
Please fix all issues reported, thanks

Lobster-0429:
Hi, @yingfeng ,
I just fixed based on review.
Thank  you!

yingfeng:
@coderabbit review

Lobster-0429:
Hi, @yingfeng ,

Lobster-0429:
Hi, @buua436 ,
Please review this PR.

buua436 (changes requested):
Thanks for the update. I still see two issues that should be addressed before merge:

1. Prune can delete valid documents if a single Salesforce object fails during slim-doc enumeration.
In `retrieve_all_slim_docs_perm_sync()`, `UnexpectedValidationError` is caught and downgraded to `warning + continue`. That means if one object fails due to a transient 5xx/429, a temporary permission issue, or any query/describe failure, prune will continue with an incomplete snapshot. The prune flow then treats missing IDs as stale and deletes them. This is not just a partial-sync problem — it can remove documents that still exist in Salesforce.
If the intent is to tolerate missing `Knowledge__kav` when Knowledge is not enabled, that should be a narrow special case, not the default behavior for all object-level failures during prune.

2. Invalid or inaccessible objects in `config.objects` are not surfaced at validation time.
The UI exposes `Objects` as a free-form text field, but `validate_connector_settings()` only probes `/sobjects` reachability and does not verify that each configured object actually exists and is queryable by the execution user. Later, `_iter_documents()` catches describe failures and silently skips that object.
As a result, a typo like `Opportunty` or a permission issue on one configured object will still pass connector validation, but the sync will silently miss data for that object. I think this should either fail fast during validation or be made explicit in the product behavior/documentation.

Lobster-0429:
Hi, @buua436 ,
Just updated based on your review, Please review again.
Thank you!

buua436 (changes requested):
Thanks for the follow-up. The two issues I called out earlier look fixed now:

- prune no longer continues on an incomplete snapshot for transient object failures
- the sync path now honors the end boundary

I still see one remaining gap, though:

The new object validation logic is not actually wired into the current execution paths.

You added good validation in `SalesforceConnector.validate_connector_settings()` to fail fast on unknown / non-queryable objects, but I don’t see that method being invoked from the Salesforce connector entry points:
- connector create/update still writes config directly
- the Salesforce sync path appears to call `load_credentials()` but not `validate_connector_settings()`

That means invalid `objects` values coming from direct API callers, scripts, or existing persisted configs can still make it to runtime, where `_iter_documents()` may skip an unavailable object instead of surfacing the configuration problem early.

So the connector class now contains the right validation, but the surrounding flow does not seem to guarantee that it actually runs. If the intended behavior is “backend-enforced fail fast for invalid objects”, I think this still needs to be wired up before merge.

Lobster-0429:
@buua436 , Update again!

buua436 (approved):
LGTM

Lobster-0429:
Hi, @yingfeng ,

PR #15464, opened 2026-06-01
feat(connectors): add HubSpot CRM data source connector

### LINKED ISSUE #15463: [Feature Request]: HubSpot data source connector
### Self Checks

```Markdown
RAGFlow has no connector for HubSpot. CRM objects — Contacts, Companies, Deals, Tickets, Notes, and Knowledge Base articles — contain customer and support context that cannot currently be indexed into a knowledge base.
```

A HubSpot data source that pulls selected objects via the HubSpot CRM v3 API and serializes each record into a knowledge-base document, with incremental sync and configurable object selection through the data-source settings UI.

Net-new connector; the Airtable connector is the closest reference for record serialization. Touchpoints:
1. `common/data_source/hubspot_connector.py` — `HubSpotConnector` extending `CheckpointedConnector`, using the CRM search endpoints with `updatedAt` paging for incremental sync.
2. `common/data_source/config.py` — add `DocumentSource.HUBSPOT = "hubspot"`.
3. `common/constants.py` — add `FileSource.HUBSPOT = "hubspot"`.
4. `common/data_source/__init__.py` — export `HubSpotConnector`.
5. `rag/svr/sync_data_source.py` — add `class HubSpot(SyncBase)` with `_generate()` and register it in `func_factory`.
6. `web/src/pages/user-setting/data-source/constant/index.tsx` — enum value, form fields (private-app access token or OAuth, object selection), defaults, display metadata + icon.
7. Tests.

```Markdown
Marketing and support teams can index HubSpot tickets and Knowledge Base articles so RAGFlow answers reflect current CRM content.
```

### Additional information

Auth via private-app access token (simplest) or OAuth 2.0; respect HubSpot rate limits through the existing rate-limit wrapper.

Lobster-0429:

Closes #15463.

RAGFlow had no way to ingest HubSpot CRM data, so marketing / support teams couldn't ground responses on live contacts, companies, deals, tickets, or Knowledge Base articles. This adds a first-class HubSpot data source connector that authenticates with either a HubSpot private app access token or an OAuth 2.0 access token, queries the CRM v3 search endpoint, and (optionally) the Knowledge Base API, with incremental sync.

**Highlights**
- `common/data_source/hubspot_connector.py`: new `HubSpotConnector` (`CheckpointedConnectorWithPermSync` + `SlimConnectorWithPermSync`).
  - Per-object `hs_lastmodifieddate` cursor stored in `HubSpotCheckpoint.cursors` — each object family advances independently and re-syncs only fetch records modified after the last successful one.
  - Auto-rewindows when a single search exhausts HubSpot's 10k results / query cap by promoting the latest seen timestamp into a new lower bound.
  - Honors HubSpot rate limits: `429` + `Retry-After` is respected with bounded exponential backoff (issue called this out explicitly); generic 5xx is propagated so the cursor never silently advances past missing data.
  - Deterministic record-to-text formatter (sorted property keys) so HubSpot property reordering on the server doesn't mark every record "changed" on each poll.
  - Knowledge Base articles via `/cms/v3/knowledge/articles`; a `404` (KB add-on not enabled) skips silently rather than aborting the sync.
  - Slim-doc IDs scoped as `<object>/<id>` so a contact and a deal sharing an autoincrement value cannot collide during prune.
- `common/constants.py`, `common/data_source/config.py`, `common/data_source/__init__.py`: register `hubspot` in `FileSource` / `DocumentSource` and export `HubSpotConnector`.
- `rag/svr/sync_data_source.py`: new `HubSpot(SyncBase)` class routed through `load_from_checkpoint` (poll_source would re-walk every object each run) and added to `func_factory`.
- Frontend:
  - `web/src/pages/user-setting/data-source/constant/index.tsx`: new `DataSourceKey.HUBSPOT`, form fields (access token, objects, include-knowledge-base checkbox, batch size), `syncDeletedFiles` capability, default form values, tile entry.
  - `web/src/locales/{en,zh}.ts`: description + per-field tooltips.
  - `web/src/assets/svg/data-source/hubspot.svg`: brand-color icon to match the other CRM/SaaS tiles.

**Verification**
- `npm run build` (vite + esbuild) passes (36 s).

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


yingfeng:
Please fix all issues reported, thanks

Lobster-0429:
@yingfeng , Updated.
Thank you

yingfeng:
@coderabbit review

Lobster-0429:
Hi, @yingfeng ,
Could you please check your CI pipeline?

Lobster-0429:
Hi, @buua436 ,
Please review this PR.

buua436 (changes requested):
Thanks for the HubSpot connector work. I still see two blocking issues:

1. The per-object checkpoint is not actually persisted across sync runs, so partial object failures can cause permanent data loss.

`HubSpot._generate()` now routes through `load_from_checkpoint()` and the connector keeps per-object cursors in `HubSpotCheckpoint.cursors`, but the sync path still creates a fresh dummy checkpoint on every run:

- `rag/svr/sync_data_source.py` builds a new checkpoint each time
- the per-object checkpoint is never persisted across tasks
- the only state that survives to the next run is the global `poll_range_start` / `next_update`

This means if one object fails partway through (for example `contacts`) while another object succeeds with newer timestamps (for example `companies`), the global watermark still advances. On the next run, `contacts` starts from that later global watermark, so records between the failure point and the advanced watermark can be skipped permanently.

2. Prune can still delete valid documents when a single object fails during slim-doc enumeration.

In `retrieve_all_slim_docs_perm_sync()`, object-level `UnexpectedValidationError` is downgraded to `warning + continue`. But the prune flow treats the returned slim-doc snapshot as authoritative and deletes anything missing from it.

So if one HubSpot object is temporarily unavailable due to a 429, 5xx, permission issue, or transient API problem, documents from that object can disappear from the snapshot and then be wrongly deleted as stale.

I think both issues need to be fixed before merge:
- either persist the per-object checkpoint across runs, or do not rely on it for incremental correctness
- and prune should fail closed on incomplete object snapshots instead of silently continuing

Lobster-0429:
@buua436 ,Please review this again.

buua436 (changes requested):
Thanks for the HubSpot connector work. I still see two correctness issues that should be addressed before merge:

1. `end_ts` is still ignored, so the sync window is not a fixed snapshot.
`poll_source(start, end)` and `load_from_checkpoint(start, end, checkpoint)` both drop the `end` argument and only pass `start` into `_iter_documents()`. `_iter_documents()` itself only applies a lower bound (`since_epoch`) and never filters by `until_epoch`.
That means the sync framework is computing `end_ts`, but the connector is not honoring it. Since the framework advances `poll_range_start` using the max `doc_updated_at` from the emitted batch, mid-run updates can leak into the current run and push the watermark past the true window, which can lead to missed records on the next run.

2. The HubSpot `objects` default is a string, but the form field is a `MultiSelect` that expects a string array.
The default config sets `objects: ''`, while the field is defined as `FormFieldType.MultiSelect`. The dynamic form passes explicit `defaultValue` through as-is, and `MultiSelect` internally assumes `string[]`. So a new HubSpot connector starts with the wrong type for `objects`, which can cause incorrect initial state or odd behavior when the field is interacted with.

Lobster-0429:
Hi, @buua436 , updated!

Lobster-0429:
Hi, @buua436 ,
Please review this PR.

yingfeng:
Can you paste screenshot to verify its working here? Thanks

JinHai-CN:
@Lobster-0429 would you please resolve the conflicts?

mkaaad:
Regarding the data source, since we don't have enough manpower for testing, and this PR does not use the SDK but instead constructs HTTP requests directly, we will at least require official API documentation or test screenshots before we will consider merging it.

mkaaad (changes requested):
Could you please resolve the conflicts?

mkaaad:
If you still want to continue contributing, please reopen this PR after resolving the issue.

PR #15466, opened 2026-06-01
feat(connectors): add Azure Blob Storage data source connector

### LINKED ISSUE #15465: [Feature Request]: Azure Blob Storage data source connector
### Self Checks

```Markdown
RAGFlow supports S3, Google Cloud Storage, R2, and OCI object storage, but has no connector for Azure Blob Storage. Users on Azure cannot index objects from a storage container into a knowledge base.
```

An Azure Blob Storage data source that lists blobs in a container and downloads them into a knowledge base, with cheap change detection (ETag / last-modified fingerprints) so unchanged blobs are skipped on re-sync, configured through the data-source settings UI.

Net-new connector that should mirror the existing object-storage connector (`common/data_source/blob_connector.py`, which covers S3/S3-compatible/GCS/R2/OCI) and the `FingerprintConnector` interface (`list_keys()` → fingerprints, `get_value()` → `Document`). Touchpoints:
1. `common/data_source/azure_blob_connector.py` (or extend `blob_connector.py`) — use    `azure-storage-blob`'s  `ContainerClient`/`BlobServiceClient`, exposing `list_keys()`  and `get_value()` with ETag-based fingerprints.
2. `common/data_source/config.py` — add `DocumentSource.AZURE_BLOB = "azure_blob"`.
3. `common/constants.py` — add `FileSource.AZURE_BLOB = "azure_blob"`.
4. `common/data_source/__init__.py` — export the connector.
5. `rag/svr/sync_data_source.py` — add `class AzureBlob(SyncBase)` with `_generate()` and register it in `func_factory`.
6. `web/src/pages/user-setting/data-source/constant/index.tsx` — enum value, form fields (account name, account key / connection string / SAS, container, optional prefix), defaults, display metadata + icon.
7. Tests.

```Markdown
Organizations on Azure can index documents from Blob Storage containers the same way S3 users do today, completing object-storage coverage across the major clouds.
```

### Additional information

This is a **data-source connector** (ingest *from* a user's Azure container into a KB), which is distinct from RAGFlow's existing Azure Blob **storage backend** — the place RAGFlow stores its own files — implemented in `rag/utils/azure_sas_conn.py` (`RAGFlowAzureSasBlob`) and `rag/utils/azure_spn_conn.py` (`RAGFlowAzureSpnBlob`).
Calling this out avoids confusion with the storage-backend bug #14159.

`azure-storage-blob` is already a project dependency (used by the storage backend), and those two modules are good references for SAS-token and service-principal (SPN) auth. Support account-key / connection-string and ideally SAS-token auth.

Lobster-0429:

Closes #15465.

RAGFlow supports S3, Google Cloud Storage, R2, and OCI as data sources but not Azure Blob Storage, leaving Azure users without a way to index container objects into a knowledge base. This adds a first-class Azure Blob Storage data-source connector — distinct from RAGFlow's existing Azure storage *backends* (`rag/utils/azure_sas_conn.py`, `rag/utils/azure_spn_conn.py`) which store RAGFlow's own files.

**Highlights**
- `common/data_source/azure_blob_connector.py`: new `AzureBlobConnector` (`CheckpointedConnectorWithPermSync` + `SlimConnectorWithPermSync`).
  - Uses the existing `azure-storage-blob` dependency (already in `pyproject.toml`).
  - Three auth modes, tried in order of precedence:
    1. **Account key** — `account_name` + `account_key` + `container_name`.
    2. **Connection string** — `connection_string` + `container_name`.
    3. **SAS token** — `container_url` + `sas_token` (same shape as `RAGFlowAzureSasBlob`).
  - ETag fingerprint stored per blob in `AzureBlobCheckpoint.etags` — unchanged blobs (same ETag as last run) are skipped without a download. Only new/modified blobs are fetched.
  - Optional `prefix` scopes indexing to a virtual folder.
  - `validate_connector_settings()` probes `get_container_properties()` and maps `AuthenticationFailed / 403 / ContainerNotFound` to typed connector exceptions.
  - Slim-doc IDs are blob names so prune reconciles correctly.
- `common/constants.py`, `common/data_source/config.py`, `common/data_source/__init__.py`: register `azure_blob` in `FileSource` / `DocumentSource` and export `AzureBlobConnector`.
- `rag/svr/sync_data_source.py`: new `AzureBlob(SyncBase)` class routed through `load_from_checkpoint` (ETag fingerprint owns change-detection) and added to `func_factory`.
- Frontend:
  - `web/src/pages/user-setting/data-source/constant/index.tsx`: new `DataSourceKey.AZURE_BLOB`, auth-mode selector (account key / connection string / SAS token), all credential fields, prefix + batch-size, `syncDeletedFiles` capability, default form values, tile entry with icon.
  - `web/src/locales/{en,zh}.ts`: description + per-field tooltips for all 9 new keys.
  - `web/src/assets/svg/data-source/azure-blob.svg`: Azure-branded stacked-cylinders icon.

**Verification**
- `npm run build` (vite + esbuild) passes (37 s).

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -156,6 +156,7 @@ class FileSource(StrEnum):
     DINGTALK_AI_TABLE = "dingtalk_ai_table"
     ONEDRIVE = "onedrive"
     OUTLOOK = "outlook"
+    AZURE_BLOB = "azure_blob"
 
 
 class PipelineTaskType(StrEnum):
--- a/common/data_source/__init__.py
+++ b/common/data_source/__init__.py
@@ -36,6 +36,7 @@ from .jira.connector import JiraConnector
 from .sharepoint_connector import SharePointConnector
 from .onedrive_connector import OneDriveConnector
 from .outlook_connector import OutlookConnector
+from .azure_blob_connector import AzureBlobConnector
 from .teams_connector import TeamsConnector
 from .moodle_connector import MoodleConnector
 from .airtable_connector import AirtableConnector
@@ -71,6 +72,7 @@ __all__ = [
     "SharePointConnector",
     "OneDriveConnector",
     "OutlookConnector",
+    "AzureBlobConnector",
     "TeamsConnector",
     "MoodleConnector",
     "BlobType",
--- a/common/data_source/azure_blob_connector.py
+++ b/common/data_source/azure_blob_connector.py
@@ -311,16 +311,20 @@ class AzureBlobConnector(CheckpointedConnectorWithPermSync, SlimConnectorWithPer
                 # unchanged blobs across runs.
                 current_etag = (blob_props.etag or "").strip('"')
 
-                # Time-window filter. ``since_epoch`` is the lower bound;
-                # ``until_epoch`` is the upper bound of the current snapshot.
-                # Enforcing the upper bound keeps blobs modified mid-run from
-                # leaking into this window — they'll be picked up by the next
-                # run (whose lower bound is this run's upper bound), so an
-                # update can never fall into a gap between windows.
+                # Time-window filter: strict lower bound, inclusive upper
+                # bound (``since_epoch`` < last-modified <= ``until_epoch``).
+                # Excluding last-modified == since_epoch (the prior run's
+                # watermark, which that run already yielded) avoids stable
+                # duplicate re-fetches on the boundary — matching the
+                # Salesforce connector's ``> since``. Enforcing the upper
+                # bound keeps blobs modified mid-run from leaking into this
+                # window; they're picked up by the next run (whose lower bound
+                # is this run's upper bound), so an update can never fall into
+                # a gap between windows.
                 last_modified: datetime | None = blob_props.last_modified
                 if last_modified:
                     ts = last_modified.timestamp()
-                    if since_epoch and ts < since_epoch:
+                    if since_epoch and ts <= since_epoch:
                         continue
                     if until_epoch and ts > until_epoch:
                         continue
--- a/common/data_source/azure_blob_connector.py
+++ b/common/data_source/azure_blob_connector.py
@@ -5,8 +5,11 @@ base.  This is distinct from RAGFlow's own Azure storage *backend*
 (``rag/utils/azure_sas_conn.py``, ``rag/utils/azure_spn_conn.py``),
 which stores RAGFlow's own files.
 
-Auth supports three mutually exclusive modes, tried in order of
-precedence:
+Auth supports three mutually exclusive modes, selected explicitly by the
+caller-supplied ``auth_mode`` (the UI hides the other modes' fields but
+does not clear them, so we must not guess from whichever field happens to
+be populated). When ``auth_mode`` is absent (older configs / direct API
+callers) we fall back to field precedence:
 
   1. **Connection string** — ``connection_string`` credential; one line,
      everything embedded.  Good for dev / testing.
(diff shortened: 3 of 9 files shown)


yingfeng:
Please fix all issues reported. Etag is very important for incremental downloading

Lobster-0429:
@yingfeng , Please review my PR again.
I updated based on review.

yingfeng:
@coderabbit review

Lobster-0429:
Hi, @yingfeng,
I think the CI failure is not related to my PR.
Thank you

buua436 (changes requested):
Thanks for the Azure Blob connector addition. I found two issues that should be addressed before merge:

1. The backend credential selection does not honor the frontend `auth_mode`.
`AzureBlobConnector.load_credentials()` currently picks credentials by field precedence (`connection_string` -> `account_name/account_key` -> `container_url/sas_token`) instead of using the explicitly selected auth mode. On the frontend, the unused credential fields appear to be hidden, but not cleared. That means if a user switches auth mode after filling another one earlier, stale values can remain in `config.credentials`, and the backend may authenticate with a different mode than the one selected in UI.

2. The ETag-based incremental logic is not persisted across sync runs.
The connector keeps ETags in `AzureBlobCheckpoint`, but `sync_data_source.py` always starts from `build_dummy_checkpoint()` and I don't see checkpoint persistence for Azure in this PR. As a result, unchanged blobs will not actually be skipped across separate scheduled sync executions, even though the connector code suggests that they should be.

I’d recommend:
- making the backend honor `config.auth_mode` explicitly, or clearing inactive credential fields before save
- either persisting the Azure checkpoint or adjusting the implementation/description so it does not imply cross-run ETag incremental behavior yet

Lobster-0429:
Hi, @buua436 ,
Please review again, I updated based on your review!

buua436 (changes requested):
Thanks for the update. The auth-mode issue looks fixed now, but I still see two blocking problems around incremental correctness:

1. Per-blob download failures are still swallowed, which can permanently skip documents.
In `_iter_documents()`, a failed `download_blob().readall()` only logs a warning and continues. However, the sync framework still advances `next_update` based on the successfully yielded documents in the same run. That means if one blob fails to download due to a transient network / permission / service error, and other newer blobs succeed, the next sync window will move forward past the failed blob. If that blob is not modified again later, it may never be retried.

2. The connector still ignores the `end` boundary, so the snapshot window is not actually fixed.
`AzureBlob._generate()` captures `end_ts` and passes it into `load_from_checkpoint(start_ts, end_ts, checkpoint)`, but `load_from_checkpoint()` drops `end`, and `_iter_documents()` only applies the lower bound (`since_epoch`) with no `last_modified <= end` filter.
So updates that happen during the run can still leak into the current task instead of being deferred to the next one. Worse, if a blob is updated after it has already been listed in the current run, while other blobs later in the run move `next_update` forward, that mid-run update can fall into a gap and be missed permanently.

I think both of these need to be addressed before merge, since they can cause silent data loss in incremental sync.

Lobster-0429:
@buua436 , Updated this PR.

buua436 (commented):
Thanks for the update. The previous two blocking issues look fixed now:

- the connector now enforces the `end_ts` upper bound
- transient download failures no longer get silently skipped

I only see one remaining issue:

The lower-bound semantics are inconsistent with the documented window and will cause boundary duplicates.

The module docstring says the incremental window is:

    since_epoch < last-modified <= until_epoch

But the actual filter is implemented as:

    if since_epoch and ts < since_epoch:
        continue

So `ts == since_epoch` is still included. Since the sync framework advances the next run’s `poll_range_start` to the previous run’s max `doc_updated_at`, blobs whose `last_modified` exactly equals the prior watermark will be fetched again on the next run.

This is not a data-loss bug, but it does create stable duplicate fetches on the boundary and does not match the current comment or the stricter lower-bound behavior used in the Salesforce connector.

I’d suggest either:
- changing the code to use a strict lower bound (`ts <= since_epoch: continue`)
- or updating the comments/docs to reflect that the lower bound is currently inclusive and that boundary re-fetch is expected

Lobster-0429:
@buua436 , Updated!

buua436 (approved):
LGTM

Lobster-0429:
Hi, @yingfeng ,

Lobster-0429:
Thank you @yingfeng!

PR #15484, opened 2026-06-01
[Bug]: Workflow agent completions default stream=True when stream is omitted

### LINKED ISSUE #15483: [Bug]: Workflow agent completions default stream=True when stream is omitted
# [Bug]: Workflow agent completions default stream=True when stream is omitted

## Description

OpenAI-compatible completions use `req.pop("stream", False)`, but workflow and legacy session paths use `req.get("stream", True)`. Clients that omit `stream` always receive SSE.

## Steps to reproduce

1. POST workflow/session completion without `stream` in JSON body.
2. Response is `text/event-stream` instead of JSON.

## Expected behavior

Non-streaming JSON when `stream` is omitted, matching OpenAI branch behavior.

## Actual behavior

SSE streaming by default on workflow paths.

glorysr1209-png:
## Summary

Closes #15483.

Default workflow/session agent completions to non-streaming when `stream` is omitted.

## Changes

- `api/apps/restful_apis/agent_api.py`: `req.get("stream", False)` on workflow paths.

## Test plan

- [ ] POST workflow completion without `stream`; assert JSON response.


yuzhichang (approved):
LGTM

PR #15432, opened 2026-05-29
fix(llm): strip non-generation keys from gen_conf for LiteLLM providers (#15427)

nickmopen:
### What problem does this PR solve?

Fixes #15427.

All LiteLLM-routed chats fail with:

- Anthropic: `litellm.BadRequestError: AnthropicException - {"type":"invalid_request_error","message":"model_type: Extra inputs are not permitted"}`
- OpenAI: `litellm.BadRequestError: OpenAIException - Unknown parameter: 'model_type'`

This is a regression from v0.25.4.

#### Root cause

A chat assistant's `llm_setting` is forwarded to the model as `gen_conf`. `llm_setting` can legitimately carry RAGFlow-internal metadata such as `model_type` (the chat REST APIs in `api/apps/restful_apis/` read it back out of `llm_setting`), so that key ends up inside `gen_conf`.

`Base._clean_conf` (OpenAI-compatible providers) already **whitelists** the keys it forwards, so direct-OpenAI providers were unaffected. `LiteLLMBase._clean_conf` only dropped `max_tokens` and passed everything else straight through to `litellm.acompletion`, which forwarded `model_type` to the upstream provider — and Anthropic / OpenAI reject it. Because both Claude and GPT route through LiteLLM, every chat broke.

#### Fix

- Extract the allowed-key set into a shared `ALLOWED_GEN_CONF_KEYS` constant and reuse it in `Base._clean_conf`.
- Apply the same whitelist in `LiteLLMBase._clean_conf`, plus the LiteLLM-specific reasoning params (`thinking`, `reasoning_effort`, `extra_body`) that the model-family policies inject for reasoning models.

This covers all four LiteLLM completion paths (`async_chat`, `async_chat_streamly`, `async_chat_with_tools`, `async_chat_streamly_with_tools`), since they all route through `_clean_conf`.

#### Tests

Adds `test/unit_test/rag/llm/test_clean_conf_whitelist.py` covering both backends: `model_type` (and other stray keys) are dropped, genuine generation params and `thinking` survive, `max_tokens` is removed, and the whitelist invariants hold.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [x] Added test cases

PR #15491, opened 2026-06-01
fix(llm): strip model_type from gen_conf in LiteLLMBase._clean_conf

vincentlambert:

Since #15141 added `model_type` to `llm_setting` from the frontend, this field leaks into `gen_conf` (because `gen_conf = dialog.llm_setting` in `dialog_service.py`) and is forwarded verbatim to litellm/provider APIs.

Providers that validate their request body reject it with a 400 error. Bedrock in particular returns:
```
BedrockException - {"message": "Malformed input request: #: extraneous key [model_type] is not permitted"}
```

### Root cause

`Base._clean_conf` already handles this implicitly — it filters `gen_conf` to an **allowlist** of known API params, so `model_type` is dropped automatically.

`LiteLLMBase._clean_conf` uses a **denylist** approach instead (only pops `max_tokens`), so `model_type` slips through to the API call.

### Fix

Add `model_type` to the denylist in `LiteLLMBase._clean_conf`, consistent with how `max_tokens` is already handled.

### Type of change

### DIFF
--- a/rag/llm/chat_model.py
+++ b/rag/llm/chat_model.py
@@ -1395,6 +1395,7 @@ class LiteLLMBase(ABC):
         )
 
         gen_conf.pop("max_tokens", None)
+        gen_conf.pop("model_type", None)  # RAGFlow-internal field, not a valid LiteLLM/API param
         return gen_conf
 
     def _need_reasoning_content_back(self) -> bool:


vincentlambert:
Closing in favour of #15432 (merged as `bebf6ed24`), which solves the same problem more completely by switching `LiteLLMBase._clean_conf` from a denylist to a shared `ALLOWED_GEN_CONF_KEYS` whitelist — covering `model_type` and all other non-generation keys in one shot.

PR #15496, opened 2026-06-02
fix: rewrite enable thinking mode for minimax

Dimon0000000:

fix the bad thinking mode for minimax

### Type of change


Haruko386 (approved):
@Dimon0000000 LGTM

PR #15028, opened 2026-05-20
fix(api): check kb ownership in /dify/retrieval

dripsmvcp:
POST /api/v1/dify/retrieval resolved the caller via @apikey_required (injecting tenant_id) but then fetched the requested knowledge_id with no tenant filter and ran the full retrieval pipeline against kb.tenant_id (the owner). Any valid Dify-compatible API key could retrieve chunks from any tenant whose KB UUID was known. Adds the missing ownership check.

## Root Cause
api/apps/sdk/dify_retrieval.py line 253: KnowledgebaseService.get_by_id(kb_id) fetched the KB by id alone, then the handler used kb.tenant_id (the OWNER) to build the embedding model and call the retriever. The caller tenant_id was only used downstream at line 278 for retrieval_by_children, well after cross-tenant data was already retrieved.

grep confirmed there was no KnowledgebaseService.accessible call anywhere in the handler.

## Fix
Two-line guard immediately after the existing get_by_id lookup, mirroring the pattern PR #14749 lands for the sibling sdk/doc.py routes (download, parse, stop_parsing, retrieval_test):

    e, kb = KnowledgebaseService.get_by_id(kb_id)
    if not e:
        return build_error_result(message="Knowledgebase not found!", code=RetCode.NOT_FOUND)
+   if not KnowledgebaseService.accessible(kb_id, tenant_id):
+       return build_error_result(message="No authorization.", code=RetCode.AUTHENTICATION_ERROR)
    if kb.tenant_embd_id:
        ...

KnowledgebaseService.accessible already handles solo-tenant ownership, team membership via TenantService.get_joined_tenants_by_user_id, and the permission=ME distinction. No behavior change for legitimate callers; cross-tenant callers now receive RetCode.AUTHENTICATION_ERROR (109).

## Test Plan
- [x] Regression test added: test/unit_test/api/apps/sdk/test_dify_retrieval.py
  - test_cross_tenant_request_is_rejected -- attacker tenant calling owner tenant KB gets 109; retriever is not invoked
  - test_same_tenant_request_succeeds -- owner tenant gets the records back
  - test_missing_knowledge_base_returns_not_found -- missing KB returns 404 BEFORE the access check fires (legit callers see the clearer message)
- [x] All 3 tests pass after the fix
- [x] Cross-tenant test FAILS on pre-fix main (KeyError on result[code] because handler leaks records dict instead of returning auth error)
- [x] ruff check clean on both changed files
[...]

PR #15508, opened 2026-06-02
fix(api): restore DocumentService.accessible check on /preview

dripsmvcp:
## Summary
Restore the `DocumentService.accessible(doc_id, current_user.id)` check that PR #15146 dropped from the REST document preview handler. Any authenticated caller could download any tenant's document bytes by guessing/knowing the `doc_id`.

## Root cause
`api/apps/restful_apis/document_api.py` — the `GET /documents/<doc_id>/preview` handler called `DocumentService.get_by_id` and went straight to `File2DocumentService.get_storage_address` + `STORAGE_IMPL.get`, with no tenant check between the lookup and the read. The handler's docstring even promises "user must belong to the tenant that owns the document's knowledge base" — the code didn't enforce it.

## Fix
- Add `current_user` to the existing `api.apps` import.
- Immediately after `get_by_id`, call `DocumentService.accessible(doc_id, current_user.id)`; on denial, return the **same** `get_data_error_result(message="Document not found!")` shape used for the missing-doc branch. That makes a cross-tenant probe indistinguishable from a missing-doc probe, preventing ID enumeration (the issue body calls this out explicitly).
- Emit `logging.warning` with caller user + doc_id for audit.
- Restores symmetry with peer routes that already call `accessible(doc_id, user_id)` (e.g. `_run_sync` at `document_api.py:1380`).

## Test plan
Adds `test/unit_test/api/apps/restful_apis/test_document_preview_accessible.py`:

- **`test_cross_tenant_preview_is_denied`** — owner tenant ≠ caller tenant; asserts the response shape is `Document not found!` and the storage backend (`thread_pool_exec(STORAGE_IMPL.get, ...)`) is **never** invoked.
- **`test_missing_doc_returns_not_found`** — missing-doc behaviour unchanged.

Stub-loader pattern mirrors `test/unit_test/api/apps/sdk/test_dify_retrieval.py` (added in #15028, passing in CI).

## Provenance — how this fix was produced

This PR was authored against a small cited knowledge base committed in the working tree as a `.vouch/` (see [vouchdev/vouch](https://github.com/vouchdev/vouch)). The loop used here:

1. **Grounding first.** Before reading the handler, queried the KB for prior context: `vouch context "tenant scoped accessible authorization"` → retrieved a cited claim distilled from PR #15028 (which restored the same `accessible()` check on `/dify/retrieval`). The retrieved rule:

[...]

Issue #15502, opened 2026-06-02
[Bug]: Agent attachment download has no authorization and can HTTP 500 on missing blob


kiannidev:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

33ef724b5

### RAGFlow image version

33ef724b5

### Other environment information

```Markdown
- Route: `GET /api/v1/agents/attachments/{attachment_id}/download`
- Handler: `api/apps/restful_apis/agent_api.py` — `download_attachment`
```

### Actual behavior

**1) Missing authorization**

The handler docstring states it mirrors the preview endpoint authorization model, but there is **no** `DocumentService.accessible`, agent/session ownership check, or attachment-to-user binding — only `@login_required` and `@add_tenant_id_to_kwargs`.

```python
data = await thread_pool_exec(settings.STORAGE_IMPL.get, tenant_id, attachment_id)
response = await make_response(data)
```

Any authenticated member of the tenant who obtains another user's `attachment_id` can download the object.

**2) Missing empty-blob guard**

If storage returns `None`/empty bytes, `make_response(data)` can raise `TypeError: response value cannot be None` → HTTP 500 (same class as #15365 for document preview).

### Expected behavior

- Verify the caller may access the specific attachment (agent session / conversation ownership) before reading storage.
- Return structured 4xx JSON when the blob is missing, consistent with `get_artifact()` and document REST download.

### Steps to reproduce

```Markdown
**Authorization**

1. User A in tenant T uploads an agent attachment; note `ATTACH_A`.
2. User B in the same tenant T authenticates.
3. User B requests:

   GET /api/v1/agents/attachments/ATTACH_A/download
   Authorization: Bearer <user_B_token>

4. Actual: attachment bytes returned without proving B owns the session/agent run.
5. Expected: denial (e.g. not found / forbidden).

**Empty blob**

1. Delete the MinIO/S3 object for a known `attachment_id` while DB metadata still references it.
2. Call the download route with valid auth.
[...]

Issue #15365, opened 2026-05-28
[Bug]: Document preview and image endpoints return HTTP 500 when storage blob is missing


kiannidev:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

7fc909acc9961e57997e07ebbd0c6bd74cf8665d

### RAGFlow image version

7fc909acc

### Other environment information

```Markdown
- OS type: Linux
- Others: Reproducible from source on `main` in `api/apps/restful_apis/document_api.py` (`get_document_image`, preview `get`). Self-hosted / Docker deployments using current `main`.
```

### Actual behavior

When object storage returns no bytes, `get_document_image()` and document preview call `make_response(data)` with `data=None`. Quart raises:

```text
TypeError: The response value returned by the view function cannot be None
```

The client receives HTTP **500** instead of a structured 4xx error.

**Images** — no check after storage fetch:

```python
data = await thread_pool_exec(settings.STORAGE_IMPL.get, bkt, nm)
response = await make_response(data)
```

**Preview** (`GET /api/v1/documents/<doc_id>/preview`) — same pattern:

```python
data = await thread_pool_exec(settings.STORAGE_IMPL.get, b, n)
response = await make_response(data)
```

(`document_api.py` approx. lines 1671–1674 and 1877–1879 on `main`.)

**Artifact** on the same file already guards correctly:

```python
if not data:
    return get_data_error_result(message="Artifact not found.")
```

Document REST download also returns JSON when empty:

```python
if not file_stream:
    return construct_json_result(message="This file is empty.", code=RetCode.DATA_ERROR)
```

### Expected behavior

If `not data` after `STORAGE_IMPL.get`, return `get_data_error_result(...)` (e.g. `"Image not found."` / `"File is empty."`) with an appropriate 4xx code — never call `make_response(None)`.

### Steps to reproduce

```Markdown
**Preview**

1. Ensure a document row exists in the DB but delete its object from MinIO/S3 (or use a stale storage address).
[...]

PR #15509, opened 2026-06-02
fix(api): return 4xx not 500 when attachment blob is missing

dripsmvcp:
## Summary
Guard the agent-attachment download against a missing or empty storage blob so the caller gets a structured 4xx (`Document not found!`) instead of an HTTP 500. Same bug class as #15365 on document preview.

## Root cause
`api/apps/restful_apis/agent_api.py` — `download_attachment` reads the blob via `STORAGE_IMPL.get(tenant_id, attachment_id)` and passes the result directly to `make_response(data)`. When the storage object is missing or empty (orphaned DB metadata, deleted MinIO/S3 object, tenant mismatch), `data` comes back as `None` / `b""`. Quart's `make_response(None)` raises `TypeError: response value cannot be None`, which the blanket `except Exception` converts to `server_error_response` → HTTP 500.

## Fix
- Add `if not data: return get_data_error_result(message="Document not found!")` immediately after the storage read, before `make_response`.
- Reuse the **same `Document not found!` shape** the document preview route uses on its missing-record path — byte-streaming endpoints now respond consistently on a not-found, which also prevents the missing-blob path from being used to enumerate attachment ids across tenants (the auth-fact "exists but empty" is no longer distinguishable from "doesn't exist at all").
- Imports already in scope (`get_data_error_result` at `agent_api.py:54`).

## Test plan
Adds `test/unit_test/api/apps/restful_apis/test_attachment_download_missing_blob.py` with three cases:

- **`test_empty_blob_returns_not_found`** — storage returns `None`; asserts `data_error` response with "not found" message, **not** the `server_error` shape `make_response(None)` would produce.
- **`test_empty_bytes_returns_not_found`** — same path with `b""`.
- **`test_nonempty_blob_proceeds`** — happy path; real bytes flow through `make_response` untouched.

Stub-loader pattern mirrors `test/unit_test/api/apps/sdk/test_dify_retrieval.py` (#15028) and `test_get_agent_session.py` (#15327), both passing in CI.

## Scope — what this PR does NOT do

Issue #15502 also raises a separate **within-tenant** authorization concern: any user in tenant T can download another user-in-T's attachment because there is no attachment→session/user ownership table. Fixing that properly requires a schema change (an attachment-ownership metadata table, or sourcing attachments through an `AgentSession`/`Conversation` join). That belongs in a separate PR with maintainer design input on the storage model. This PR addresses only bug **(2) Missing empty-blob guard**, which is self-contained, sees real production HTTP 500s today (per the issue body), and lands cleanly.

`Refs #15502` (rather than `Closes`) so the bigger authz item stays tracked.

## Provenance — how this fix was produced

Authored against a small in-repo cited knowledge base ([vouchdev/vouch](https://github.com/vouchdev/vouch)) seeded from prior work on the sibling fix #15028 / [#15508](https://github.com/infiniflow/ragflow/pull/15508).

Before reading the handler, `vouch context "agent attachment download missing authorization tenant"` returned the rule recorded after the previous IDOR fix:

> *ragflow REST **byte-streaming routes** (preview, **download**) must deny cross-tenant access with the same response shape as missing-record (e.g. `"Document not found!"`) to prevent ID enumeration.*

That claim directly informed the denial shape used here (`"Document not found!"` rather than a generic 4xx body or a `"forbidden"` message). The grounding step ran *before* any code was read, so the design decision was derived from prior precedent rather than re-improvised. Mechanism is **not a runtime dependency** of this PR — it's process discipline, fully external to the codebase.

Refs #15502


dripsmvcp:
@dosu

dripsmvcp:
@yingfeng Getting attention here. Would you check this PR?

wangq8:
@dripsmvcp please help fix the UT.

wangq8:
@buua436

buua436 (changes requested):
Could we change this error message to `Attachment not found!`? This case is for missing/inaccessible attachments, not documents.

PR #15511, opened 2026-06-02
fix: show default embedding model when provider is not yet registered

euvre:
### Problem

On the Model Providers page, the Embedding Model dropdown in System Model Settings shows empty (no default selected), even though a default embedding model is configured in `service_conf.yaml`.

### Root Cause

Two issues were identified:

1. **Backend: `_get_model_info` fails for unregistered providers**
   The tenant's `embd_id` is set to `bge-m3@xxxx` during initialization (from the placeholder config `factory: 'xxxx'`). The `_get_model_info` function requires the provider to exist in `tenant_model_provider` table, but `xxxx` is never a real provider. Even after the user adds a real provider (e.g., ZHIPU-AI), the stale `embd_id` still references the non-existent one, causing the function to return `None`.

2. **Frontend: default models cache not invalidated after adding provider**
   `useAddProviderInstance` only invalidates `addedProviders` and `allModels` caches after adding a provider instance, but does **not** invalidate the `defaultModels` cache. This means the default model list is not re-fetched until the user manually refreshes the page.

### Fix

**`api/apps/services/models_api_service.py`**

- Added `_resolve_model_from_tenant_providers()` helper: when the default model's provider doesn't exist (e.g., placeholder `xxxx`), it searches through the tenant's actually registered providers for a model of the same type and returns the first match.
- When an instance name doesn't match (e.g., `"default"` vs actual name `"1"`), the function now auto-resolves to the first real instance under that provider.
- Falls back to `FACTORY_LLM_INFOS` validation when neither provider nor instance exists.

**`web/src/hooks/use-llm-request.tsx`**

- Added `queryClient.invalidateQueries({ queryKey: LlmKeys.defaultModels() })` to `useAddProviderInstance` so that the default model list is re-fetched immediately after a provider instance is added, eliminating the need for a manual page refresh.

### Testing

- Verified with a tenant whose `embd_id=bge-m3@xxxx` and only provider is ZHIPU-AI (instance `1`): `_resolve_model_from_tenant_providers` correctly resolves to `embedding-2@1@ZHIPU-AI`.
- After adding a provider via the UI, the embedding model dropdown now immediately shows the resolved default without requiring a page refresh.

### Type of change


Lynn-Inf (approved):
LGTM

PR #15527, opened 2026-06-02
fix(go-models): fix MiniMax driver requests

oktofeesh1:
## Summary
- keep MiniMax chat calls in non-streaming mode and streaming calls in SSE mode
- make MiniMax model listing and connection checks use a bodyless GET /v1/models
- add focused MiniMax request/response regression tests

## What changed
- return a real MiniMax driver from NewInstance
- validate API keys, model names, and stream senders before issuing requests
- remove raw SSE line logging from the chat stream path
- parse ListModels responses with typed structs instead of unchecked assertions

## Why
MiniMax documents its OpenAI-compatible model list endpoint as GET /v1/models with bearer auth: https://platform.minimax.io/docs/api-reference/models/openai/list-models. The previous driver sent a JSON body on that GET, let callers flip the expected stream mode, and could panic on malformed model-list responses.

## Validation
- go test ./internal/entity/models -run TestMinimax -count=1
- git diff --check

## Notes
- Also ran go test ./internal/entity/models -count=1; it currently fails in an unrelated upstream Astraflow embedding test before this MiniMax-specific coverage is involved.

### DIFF
--- a/internal/entity/models/minimax.go
+++ b/internal/entity/models/minimax.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"ragflow/internal/common"
 	"strings"
 	"time"
 )
@@ -61,9 +60,28 @@ func (m *MinimaxModel) Name() string {
 	return "minimax"
 }
 
+func validateMinimaxAPIKey(apiConfig *APIConfig) (string, error) {
+	if apiConfig == nil || apiConfig.ApiKey == nil || strings.TrimSpace(*apiConfig.ApiKey) == "" {
+		return "", fmt.Errorf("api key is required")
+	}
+	return strings.TrimSpace(*apiConfig.ApiKey), nil
+}
+
+func validateMinimaxModelName(modelName string) (string, error) {
+	if strings.TrimSpace(modelName) == "" {
+		return "", fmt.Errorf("model name is required")
+	}
+	return strings.TrimSpace(modelName), nil
+}
+
 // ChatWithMessages sends multiple messages with roles and returns response
 func (m *MinimaxModel) ChatWithMessages(modelName string, messages []Message, apiConfig *APIConfig, chatModelConfig *ChatConfig) (*ChatResponse, error) {
-	if err := m.baseModel.APIConfigCheck(apiConfig); err != nil {
+	apiKey, err := validateMinimaxAPIKey(apiConfig)
+	if err != nil {
+		return nil, err
+	}
+	modelName, err = validateMinimaxModelName(modelName)
+	if err != nil {
 		return nil, err
 	}
 	if len(messages) == 0 {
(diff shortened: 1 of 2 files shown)


oktofeesh1:
Hi @Haruko386 @JinHai-CN - please review when you have a little time. 🤝

internal/entity/models/minimax.go
```diff
@@ -143,7 +157,7 @@ func (z *MinimaxModel) ChatWithMessages(modelName string, messages []Message, ap
 	}
 
 	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", *apiConfig.ApiKey))
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", apiKey))
```
Haruko386: Header.Set

Haruko386 (changes requested):
@oktofeesh1 Do a small change would be ok

oktofeesh1:
> @oktofeesh1 Do a small change would be ok

Fixed, thanks for the review

internal/entity/models/minimax.go
```diff
@@ -61,9 +60,28 @@ func (m *MinimaxModel) Name() string {
 	return "minimax"
 }
 
+func validateMinimaxAPIKey(apiConfig *APIConfig) (string, error) {
+	if apiConfig == nil || apiConfig.ApiKey == nil || strings.TrimSpace(*apiConfig.ApiKey) == "" {
+		return "", fmt.Errorf("api key is required")
+	}
+	return strings.TrimSpace(*apiConfig.ApiKey), nil
+}
```
Haruko386: Just use this [func](https://github.com/infiniflow/ragflow/blob/5a04ac0864808bb3309e0b5735b56ad5a03d2efd/internal/entity/models/base_model.go#L31)

Haruko386 (changes requested):
@oktofeesh1

oktofeesh1:
> @oktofeesh1

Fixed as well!

internal/entity/models/minimax.go
```diff
@@ -396,10 +389,10 @@ func (m *MinimaxModel) Embed(modelName *string, texts []string, apiConfig *APICo
...
 func (m *MinimaxModel) ListModels(apiConfig *APIConfig) ([]string, error) {
-	apiKey, err := validateMinimaxAPIKey(apiConfig)
-	if err != nil {
+	if err := m.baseModel.APIConfigCheck(apiConfig); err != nil {
 		return nil, err
 	}
+	apiKey := strings.TrimSpace(*apiConfig.ApiKey)
 
 	resolvedBaseURL, err := m.baseModel.GetBaseURL(apiConfig)
 	if err != nil {
```
Haruko386, line 398: please fetch the upstream code

Haruko386 (changes requested):
Hi @oktofeesh1 
please refer [this](https://github.com/infiniflow/ragflow/blob/220ee9dbfb9f5b7ef0241a1b6f7d7cf03f401d48/internal/entity/models/minimax.go#L208) and fetch the latest code

oktofeesh1:
Thanks, I fetched/rebased onto the latest upstream code and verified the MiniMax paths now follow the current pattern: `ChatWithMessages`, `ChatStreamlyWithSender`, and `ListModels` call `m.baseModel.APIConfigCheck(apiConfig)`, then use `strings.TrimSpace(*apiConfig.ApiKey)` for the Authorization header. Validation passed with `go test ./internal/entity/models -run 'TestMinimax' -count=1`, `go test ./internal/entity/models -run '^$' -count=1`, `go vet ./internal/entity/models`, and `git diff --check`.

Haruko386 (approved):
@oktofeesh1 LGTM

PR #15535, opened 2026-06-02
fix(provider): unblock OpenAI-API-Compatible instance creation + fix …

Attili-sys:
…await bug

Two bugs introduced by commits 3bc5ed282 and 36357a6af:

--- Bug 1 (primary): OpenAI-API-Compatible instances cannot be created ---

36357a6af removed the optional `verify` flag from `create_provider_instance` and replaced it with an unconditional call to `verify_api_key`.  This broke every provider whose factory entry in llm_factories.json ships with an empty `llm` list — most notably "OpenAI-API-Compatible".

`verify_api_key` early-exits with a failure when the provider has no models:

    if not factory_llms:
        return False, "No models found for provider '...'"

So any attempt to add an OpenAI-API-Compatible instance was rejected with an error, regardless of whether the supplied API key and base URL are valid.

Fix: look up the factory entry inside `create_provider_instance` and only invoke `verify_api_key` when the provider has pre-configured models to probe. Generic providers have no preset models because the user supplies them later via `add_model_to_instance`; there is nothing to test at instance-creation time, so verification is correctly skipped.

  Reproducer:
    POST /providers/OpenAI-API-Compatible/instances
    {"instance_name": "my-llm", "api_key": "sk-...", "base_url": "http://..."}
    → was: error "No models found for provider 'OpenAI-API-Compatible'"
    → now: 200 "success"

--- Bug 2 (secondary): missing await in embedding model verification ---

In the same `verify_api_key` function, the embedding branch called:

    arr, tc = asyncio.wait_for(
        asyncio.to_thread(mdl.encode, [...]),
        timeout=timeout_seconds,
    )

`asyncio.wait_for` returns a coroutine and must be awaited.  Without `await` the assignment receives a coroutine object and immediately raises:

    TypeError: cannot unpack non-iterable coroutine object

The rerank and chat branches in the same loop use `await` correctly; this was a plain oversight in the embedding branch introduced in 3bc5ed282.  Add the missing `await`.

### Type of change

### DIFF
--- a/api/apps/services/provider_api_service.py
+++ b/api/apps/services/provider_api_service.py
@@ -245,8 +245,8 @@ async def create_provider_instance(tenant_id: str, provider_name: str, instance_
             return False, f"Already exist instance: {same_key_instance.instance_name} with api_key {api_key}"
 
     # Only verify the API key when the provider has pre-configured models to test
-    # against.  Generic providers such as "OpenAI-API-Compatible" ship with an
-    # empty model list in llm_factories.json — users populate them later — so
+    # against. Generic providers such as "OpenAI-API-Compatible" ship with an
+    # empty model list in llm_factories.json; users populate them later, so
     # there is nothing to probe and verification must be skipped.
     factory_entry = next((f for f in FACTORY_LLM_INFOS if f["name"] == provider_name), None)
     if factory_entry and factory_entry.get("llm"):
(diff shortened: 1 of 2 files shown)


JinHai-CN:
if you are still working on this PR, please re-open the PR.

PR #15537, opened 2026-06-02
fix(floating-chat-widget): stop reverting textarea on every keystroke (#14408)

### LINKED ISSUE #14408: focus is lost
After embedding the chat floating component using version 0.25.0 into the system, the first input and response work normally. However, for subsequent input queries, focus loss occurs, the textbox target is lost, requiring a re-click to enter one character before focus is lost again.

jeffrey701:
### What

After the first character in the embedded chat widget, the textarea loses focus and only accepts one character per click. Root cause: the input-sync `useEffect` listed `inputValue` in its dep array. Every `onChange` does `setInputValue(newValue)` + `handleInputChange(e)` — those state updates do not necessarily batch (handleInputChange comes from a sibling hook). The effect re-ran with `hookValue` still on the previous value, hit `hookValue && hookValue !== inputValue`, and called `setInputValue(hookValue)` — reverting the input. The resulting double-render cycle caused the focus drop the issue describes.

### Fix

Drop `inputValue` from the dep array and use the functional setter so we still compare against the latest local value without re-firing on every keystroke. The effect now only runs when `hookValue` actually changes from an external source (e.g. send-clear).

### File touched
- `web/src/components/floating-chat-widget.tsx` (+11/-5)

Closes #14408


dcc123456:
The issue cannot be reproduced

JinHai-CN:
@jeffrey701 please re-open the PR, if you encounter the issue again.

PR #15539, opened 2026-06-03
feat(connector): implement Hyland Alfresco data-source connector (iss…

### LINKED ISSUE #15538: [Feature Request]: Hyland Alfresco data-source connector
### Self Checks

```Markdown
RAGFlow can't ingest documents managed in Hyland Alfresco (an enterprise ECM), so controlled document libraries stay outside RAG. #12313 lists Hyland Alfresco under *Planned* with no PR.
```

An Alfresco connector that crawls document-library nodes and downloads their content into a KB, with incremental sync.

- **Auth:** basic auth or OAuth2 against a configurable Alfresco base URL.
- **Scope:** selectable sites/folders; documents (pdf/docx/xlsx/txt/md) with their metadata; optionally version history.
- **Base class:** `CheckpointedConnector`, polling on node `modifiedAt`.

- Alfresco REST API (`/alfresco/api/-default-/public/alfresco/versions/1/nodes/...`) or CMIS 1.1 browser binding; `cmislib` is an option for CMIS.
- Recurse folders via `children`; download via the node `content` endpoint; incremental via `modifiedAt`.

```Markdown
Enterprises with governed Alfresco document libraries (policies, contracts, SOPs) RAG over
them without exporting to a file share first.
```

### Additional information

Parent tracker: #12313. One connector = one PR. Pairs with the OpenText request (shared ECM/CMIS semantics).

Lobster-0429:

Closes [#15538](https://github.com/infiniflow/ragflow/issues/15538) — adds a **Hyland Alfresco** data-source connector so enterprises can RAG over governed Alfresco document libraries without an export step.

The connector crawls document-library nodes via the Alfresco public REST API (`/alfresco/api/-default-/public/alfresco/versions/1/nodes/...`) and ingests file content into a knowledge base.

#### Behavior

- **Auth** — two mutually exclusive modes selected by an explicit `auth_mode` (the UI hides the inactive mode's fields but doesn't clear them, so we never guess from whichever field happens to be populated), with field-precedence fallback for older configs / direct API callers:
  - `basic` — `username` + `password`.
  - `oauth2` — `access_token` bearer token (Alfresco Identity Service / external IdP).
- **Scope** — chosen by the caller:
  - `site_ids` — each site's `documentLibrary` container is resolved to a node ID and crawled.
  - `root_node_ids` — explicit folder node IDs.
  - neither — the whole repository (`-root-`).
- **Traversal** — iterative breadth-first walk of the `children` endpoint (paginated), descending folders and yielding file nodes; content fetched via the node `content` endpoint. A node-count guard protects against pathological / cyclic structures.
- **Incremental sync** — scoped by the poll time window on node `modifiedAt`. Each file's version label (`cm:versionLabel`, or `modifiedAt` when versioning is off) is emitted as the document fingerprint, which the pipeline persists as `content_hash` so unchanged nodes are not re-embedded. The connector keeps no cross-run state.
- **Prune / permission sync** — `retrieve_all_slim_docs_perm_sync` yields slim documents for stale-document cleanup.
- **Validation** — `validate_connector_settings()` makes one cheap call against the repository root to prove the base URL + credentials, and resolves each configured site up-front so a typo'd site surfaces as a clear error rather than a silently empty crawl. It is wired into the sync `_generate` path so invalid configs fail fast regardless of how they were persisted.

#### Conflict-minimizing structure

All logic lives in the new `common/data_source/alfresco_connector.py`; the frontend form fields live in their own `alfresco-constant.tsx` (mirroring the Confluence pattern). Each shared-registration file is touched with a single appended line so concurrent connector PRs produce at most a trivial "both added a line" union, not an overlapping conflict:

- `common/constants.py` — `FileSource.ALFRESCO`
- `common/data_source/config.py` — `DocumentSource.ALFRESCO`
- `common/data_source/__init__.py` — import + `__all__`
- `rag/svr/sync_data_source.py` — import, `Alfresco(SyncBase)` sync class, `func_factory` entry
- `web/.../constant/index.tsx` — `DataSourceKey`, info tile, form-fields ref, default values
- `web/src/locales/{en,zh}.ts` — one description key each

#### Notes / follow-ups

- The time window uses a strict lower / inclusive upper bound (`since < modifiedAt <= until`) so a node sitting exactly on the previous run's watermark isn't re-fetched every run, and nodes modified mid-run don't leak into the current snapshot (they're picked up by the next run, so an update can't fall into a gap).
- A node deleted between listing and fetch (HTTP 404) is skipped as genuinely gone; any other download error aborts the run so the sync watermark can't advance past content that was never ingested.
- The data-source tile uses the lucide `FolderTree` icon as a placeholder; a dedicated `data-source/alfresco` SVG can be added later to match the other connectors.

### Type of change


Lobster-0429:
Hi, @yingfeng,
Could you please review this?

buua436 (changes requested):
Thanks for the Alfresco connector work. I still see two issues that should be addressed before merge:

1. `root_node_ids` are not validated at all, so a typo can silently turn the sync into an empty crawl.
`validate_connector_settings()` only validates `site_ids` by resolving each site's document library. If the user configures `root_node_ids` directly, there is no equivalent existence check. Later, `_list_children()` treats a 404 as a normal early return, so a bad root node ID will just produce no documents instead of surfacing a configuration error.
That is risky because it can look like a successful sync while actually ingesting nothing, and it can also interact badly with prune if documents already existed from earlier runs.

2. `include_version_history` is exposed in both the UI and the backend constructor, but the connector never uses it.
The form and default config include `include_version_history`, and the connector stores the flag, but I couldn’t find any branch that changes the crawl behavior, version retrieval, or fingerprinting based on it. As a result, toggling this option has no observable effect, which is misleading for users.

Lobster-0429:
Hi, @buua436 , updated!

yingfeng:
For all data connectors, due to our limited manpower, please ask the author of each PR to provide evidence, including screenshots, to prove that the data connector is working properly. Thank you~

Lobster-0429:
Hi, @buua436 ,
Please review this PR.

yingfeng:
Can you paste screenshot to verify its working here? Thanks

JinHai-CN:
I would like to close this PR, if you attempt to resolve the conflicts and continue to work on this PR, please re-open it.

PR #15549, opened 2026-06-03
Refa: refine mysql migration version workflow

buua436:

refine mysql migration version workflow

### Type of change

- [x] Refactoring


Lynn-Inf (approved):
LGTM

PR #15389, opened 2026-05-29
Fix: Chat completion generation parameter overrides

bitloi:
### What problem does this PR solve?

Closes #15388.

Chat completion routes did not reliably honor per-request generation settings:

- `/api/v1/chat/completions` copied generation settings with a truthiness check, so valid zero values such as `temperature: 0`, `top_p: 0`, `frequency_penalty: 0`, `presence_penalty: 0`, and `max_tokens: 0` were dropped.
- `/api/v1/openai/{chat_id}/chat/completions` did not forward standard generation settings into the request-specific dialog LLM settings before calling `async_chat`.

This PR preserves explicitly supplied generation parameters, including zero values, and merges request-level overrides into existing dialog settings where appropriate.

The supported generation parameter keys and merge behavior live in a shared REST API helper to keep both completion routes aligned.

Validation:

- `git diff --check`
- `python3 -m py_compile api/apps/restful_apis/_generation_params.py api/apps/restful_apis/chat_api.py api/apps/restful_apis/openai_api.py test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py`
- `uv run ruff check api/apps/restful_apis/_generation_params.py api/apps/restful_apis/chat_api.py api/apps/restful_apis/openai_api.py test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py`
- `ZHIPU_AI_API_KEY=dummy uv run pytest test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py -q -k generation_params`

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

PR #15610, opened 2026-06-03
fix(api): apply per-request generation overrides for existing chats

### LINKED ISSUE #15605: Bug: per-request generation overrides are silently dropped for existing chats
## Summary

`POST /api/v1/chat/completions` ignores per-request generation settings
(`temperature`, `top_p`, `frequency_penalty`, `presence_penalty`, `max_tokens`)
whenever the targeted chat assistant already has a configured model. The
overrides are accepted by the request parser but never applied, so generation
runs with the dialog's stored `llm_setting` instead of the caller's values.

## Affected code

`api/apps/restful_apis/chat_api.py`, in `session_completion`:

```python
chat_model_config = pop_generation_config(req)   # collects the caller's overrides
...
if chat_model_id:                                 # caller explicitly chose a model
    ...
    dia.llm_setting = chat_model_config
elif not dia.llm_id:                              # dialog has NO model yet
    ...
    dia.llm_id = tenant_info.llm_id
    merge_generation_config(dia, chat_model_config)   # <-- merge nested here
```

`merge_generation_config(dia, chat_model_config)` lives inside the
`elif not dia.llm_id:` branch, so it only runs for a dialog that has no model
configured yet. For the dominant case — a request with a valid `chat_id` to an
existing chat whose dialog already has `dia.llm_id` set, and no per-request
`llm_id` override — **both branches are skipped** (`chat_model_id` is falsy and
`not dia.llm_id` is `False`), so `chat_model_config` is discarded.
`async_chat` then runs with `dia.llm_setting` only, ignoring the request's
generation parameters.

## Why this is a regression

This is the exact behavior PR #15389 ("Chat completion generation parameter
overrides", closes #15388) set out to fix — its description says it "merges
request-level overrides into existing dialog settings where appropriate" — but
the merge was placed inside the `elif`, so it never fires for existing,
already-configured chats.

The sibling OpenAI-compatible route does it correctly and **unconditionally**:

```python
# api/apps/restful_apis/openai_api.py
merge_generation_config(dia, extract_generation_config(req))
```

PR #15389 also shipped a regression test that encodes the intended behavior:

`test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py`
→ `test_session_completion_merges_generation_params_for_existing_chat`

That test sets `dia.llm_id = "model"` and a stored `llm_setting`
`{"temperature": 0.7, "top_p": 0.3, "custom": "keep"}`, sends a request with
`temperature: 0, presence_penalty: 0`, and asserts the merged result
`{"temperature": 0, "top_p": 0.3, "custom": "keep", "presence_penalty": 0}`.
**It currently fails on `main`** — the dialog keeps `temperature: 0.7` and never
gains `presence_penalty: 0`, proving the bug.

## Steps to reproduce

1. Create a chat assistant (its dialog now has a configured `llm_id`).
2. `POST /api/v1/chat/completions` with that chat:
   ```json
   {
     "chat_id": "<existing chat id>",
     "session_id": "<existing session id>",
     "messages": [{"role": "user", "content": "hi"}],
     "temperature": 0,
     "max_tokens": 256
   }
   ```
3. Observe that generation uses the assistant's saved temperature/max_tokens,
   not the `0` / `256` supplied in the request.

Expected: the explicitly supplied generation parameters override the dialog's
saved settings for this request.

## Fix

Run the merge in every non-explicit-model path, falling back to the tenant
default model only when the dialog has none. This mirrors the OpenAI-compatible
route and is safe: `merge_generation_config` is a no-op when no overrides were
supplied (`extract_generation_config` only includes keys the caller actually
sent), and it preserves any dialog settings the caller did not override.

```python
        if chat_model_id:
            if not await thread_pool_exec(get_api_key, tenant_id=dia.tenant_id, model_name=chat_model_id):
                return get_data_error_result(message=f"Cannot use specified model {chat_model_id}.")
            dia.llm_id = chat_model_id
            dia.llm_setting = chat_model_config
        else:
            if not dia.llm_id:
                logging.info("empty chat_model_id in req, use default chat model.")
                _, tenant_info = TenantService.get_by_id(dia.tenant_id)
                if not tenant_info or not tenant_info.llm_id:
                    raise LookupError("No default chat model for tenant.")
                dia.llm_id = tenant_info.llm_id
            merge_generation_config(dia, chat_model_config)
```

## Impact

- Severity: functional / correctness. Callers cannot control generation
  parameters per request for any existing chat assistant — `temperature`,
  `max_tokens`, etc. are silently ignored, which is surprising and hard to
  diagnose (the request is accepted, just not honored).
- Scope: `POST /api/v1/chat/completions` for any persisted chat with a
  configured model (the common case). The OpenAI-compatible route is
  unaffected.

## Validation

- `python3 -m py_compile api/apps/restful_apis/chat_api.py` — OK
- `ruff check api/apps/restful_apis/chat_api.py` — clean
- `pytest test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py -k generation_params`
  — `test_session_completion_merges_generation_params_for_existing_chat` goes
  from failing → passing; `test_session_completion_preserves_zero_generation_params`
  (default-dialog path) still passes; no other `session_completion` test regresses.

nickmopen:

Closes #15605.

`POST /api/v1/chat/completions` (`session_completion` in
`api/apps/restful_apis/chat_api.py`) accepts per-request generation
settings — `temperature`, `top_p`, `frequency_penalty`,
`presence_penalty`, `max_tokens` — but never applies them when the
targeted chat assistant already has a configured model.

`merge_generation_config(dia, chat_model_config)` was placed inside the
`elif not dia.llm_id:` branch:

```python
if chat_model_id:                 # caller explicitly chose a model
    ...
    dia.llm_setting = chat_model_config
elif not dia.llm_id:              # dialog has NO model yet
    ...
    merge_generation_config(dia, chat_model_config)
```

For the common case — a valid `chat_id` to an existing chat whose dialog
already has `dia.llm_id` set, with no per-request `llm_id` — **both
branches are skipped** (`chat_model_id` is falsy and `not dia.llm_id` is
`False`), so the parsed overrides are discarded and `async_chat` runs
with the dialog's stored `llm_setting`. The request is accepted, just not
honored.

This is the exact behavior PR #15389 set out to fix, but the merge landed
inside the `elif`, so it never fired for existing, already-configured
chats. The sibling OpenAI-compatible route (`openai_api.py`) does it
correctly and unconditionally:

```python
merge_generation_config(dia, extract_generation_config(req))
```

### How does this PR fix it?

Run the merge on every non-explicit-model path, falling back to the
tenant default model only when the dialog has none:

```python
if chat_model_id:
    ...
    dia.llm_setting = chat_model_config
else:
    if not dia.llm_id:
        ...
        dia.llm_id = tenant_info.llm_id
    merge_generation_config(dia, chat_model_config)
```

This mirrors the OpenAI-compatible route and is safe:
`merge_generation_config` is a no-op when no overrides were supplied
(`extract_generation_config` only includes keys the caller actually
sent), and it preserves any dialog settings the caller did not override.

The regression test shipped with #15389,
`test_session_completion_merges_generation_params_for_existing_chat`,
encodes the intended behavior and **fails on `main`** — the dialog keeps
`temperature: 0.7` and never gains `presence_penalty: 0`. It passes with
this change. Two further tests are added:

- `test_session_completion_keeps_dialog_settings_without_overrides` —
  no per-request params ⇒ stored settings pass through untouched (guards
  against a future "always overwrite" regression).
- `test_session_completion_assigns_default_model_then_merges_for_existing_chat`
  — existing chat whose dialog has no model yet ⇒ tenant default assigned
  **and** overrides still merged.

### Validation

- `pytest test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py -k session_completion` — 8 passed.
- `ruff check api/apps/restful_apis/chat_api.py test/.../test_session_sdk_routes_unit.py` — clean.

### Type of change

- [x] Test cases


nickmopen:
Morning @buua436 , Whenever you have a moment, I’d appreciate a review of this PR. Thanks in advance!

nickmopen:
@JinHai-CN , Sorry to bother you, but would you mind reviewing this PR when you get a chance? Thanks!

nickmopen:
cc: @KevinHuSh

nickmopen:
Weekly follow-up.

Lynn-Inf (approved):
**LGTM**. The change is small: the key part isn't the `elif` → `else { if }` restructure (that step is equivalent), but that `merge_generation_config` was dedented one level out of `if not dia.llm_id`. With that, the case where `chat_model_id` is empty but
  `dia.llm_id` is already set no longer falls through both branches — which is exactly the #15605 scenario, and why `test_session_completion_merges_generation_params_for_existing_chat` fails on `main`. The other two states are unchanged.

One non-blocking follow-up:

When `llm_id` is passed explicitly we take `dia.llm_setting = chat_model_config`, and since `chat_model_config` comes from `pop_generation_config` it only contains the keys the caller actually sent — so stored dialog settings such as `top_p` / `max_tokens` get wiped and then refilled with defaults by `resolve_llm_setting`. `openai_api.py:285` only sets `dia.llm_id` for an explicit model and then merges unconditionally, so the two paths diverge. Not sure whether that's intentional here; worth tracking in a separate issue. Doesn't block this merge.

Lynn-Inf:
CI is red: `ragflow_preflight` fails at the "Run Lefthook on changed files" step, where `ruff-format` reports `Would reformat: test/testcases/test_http_api/test_session_management/test_session_sdk_routes_unit.py` (the `lambda: _AwaitableValue({...})` blocks in the two new tests). Since preflight gates everything downstream, `ragflow_native_backend`, `ragflow_tests_elasticsearch` and `ragflow_tests_infinity` are all skipped, so no tests actually ran.

I'll close this PR and carry the change over into a fresh one with the formatting fixed, so it can go green. Nothing here is being rejected.

PR #15618, opened 2026-06-03
fix(rag): paragraph() splits a token string into characters

### LINKED ISSUE #15617: [Bug]: FulltextQueryer.paragraph splits a token string into characters
## Description
`FulltextQueryer.paragraph(content_tks: str, ...)` is called by `Dealer.tag_content` (`rag/nlp/search.py`) with a **space-joined token string**: `doc["title_tks"] + " " + doc["content_ltks"]`. When the argument is a `str`, `paragraph` does:

```python
content_tks = [c.strip() for c in content_tks.strip() if c.strip()]
```

This iterates the string **character by character**, so `self.tw.weights(content_tks, preprocess=False)` receives single characters instead of whole tokens.

## Steps to reproduce
1. Call `FulltextQueryer.paragraph("foo bar baz")`.
2. Inspect what is passed to `self.tw.weights`.

## Expected behavior
The term-weight computation receives whole tokens: `["foo", "bar", "baz"]`.

## Actual behavior
It receives characters: `["f", "o", "o", "b", "a", "r", ...]`, so the resulting `TAG_FLD` MatchTextExpr is built from per-character "terms" with meaningless weights.

## Impact
`Dealer.tag_content` (auto-tagging) and tag-feature reranking build their tag-matching query from single characters, degrading tag assignment and tag-based ranking.

## Component
- `rag/nlp/query.py` (`FulltextQueryer.paragraph`)

dearsishs:
## Summary
`FulltextQueryer.paragraph` split a space-joined token string into **characters** instead of tokens, so term weights (and the resulting tag-feature query) were computed over single characters.

## Root cause
`Dealer.tag_content` passes `title_tks + " " + content_ltks` (a token string). The `str` branch did `[c.strip() for c in content_tks.strip() if c.strip()]`, which iterates characters.

## Fix
```diff
-            content_tks = [c.strip() for c in content_tks.strip() if c.strip()]
+            content_tks = [c for c in content_tks.split() if c.strip()]
```
Splits on whitespace into whole tokens, matching how every other tokenized field is handled (`self.tw.weights(..., preprocess=False)` expects tokens).

## Files changed
- `rag/nlp/query.py`
- `test/unit_test/rag/test_paragraph_token_split.py` (new)

## Verification
- `ruff check` / `ruff format --check` — clean
- Added unit test: `paragraph("foo bar baz")` passes `["foo","bar","baz"]` to `tw.weights` (was characters). (Standalone-verified the split.)
- Local full pytest not run (heavy RAG deps); CI validates.

## Note
Implemented with LLM assistance (model: claude-opus-4-8).

Closes #15617

### DIFF
--- a/rag/nlp/query.py
+++ b/rag/nlp/query.py
@@ -223,7 +223,7 @@ class FulltextQueryer(QueryBase):
 
     def paragraph(self, content_tks: str, keywords: list = [], keywords_topn=30):
         if isinstance(content_tks, str):
-            content_tks = [c.strip() for c in content_tks.strip() if c.strip()]
+            content_tks = [c for c in content_tks.split() if c.strip()]
         tks_w = self.tw.weights(content_tks, preprocess=False)
 
         origin_keywords = keywords.copy()
--- a/test/unit_test/rag/test_paragraph_token_split.py
+++ b/test/unit_test/rag/test_paragraph_token_split.py
@@ -1,0 +1,48 @@
+#
+#  Copyright 2025 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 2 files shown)


JinHai-CN:
already fixed.

PR #15624, opened 2026-06-03
fix(api): avoid 500 on non-string questions in add_chunk/update_chunk

### LINKED ISSUE #15623: [Bug]: add_chunk/update_chunk 500 on non-string questions
## Description
`add_chunk` and `update_chunk` validate that `questions` is a list but not that its elements are strings, then do `"\n".join(req["questions"])` on the **raw** list:

```python
if "questions" in req and not isinstance(req["questions"], list):
    return get_error_data_result("`questions` is required to be a list")
...
d["question_kwd"] = [str(q).strip() for q in req.get("questions", []) if str(q).strip()]
d["question_tks"] = rag_tokenizer.tokenize("\n".join(req.get("questions", [])))  # raw join
```

A request with non-string elements (e.g. `{"content": "x", "questions": [1, 2]}`) makes `"\n".join([1, 2])` raise `TypeError: sequence item 0: expected str instance, int found`. `add_chunk` has no surrounding `try/except`, so the client gets a raw HTTP 500. The line just above is already safe because it uses `str(q)`.

## Steps to reproduce
1. `POST /api/v1/datasets/{id}/documents/{id}/chunks` with body `{"content": "x", "questions": [1, 2]}`.

## Expected behavior
A clean `4xx` error (or the integers coerced to strings consistently with `question_kwd`).

## Actual behavior
`TypeError` → unhandled HTTP 500.

## Impact
Malformed (non-string) `questions` input crashes the endpoint with a 500 instead of a clean validation error. Same pattern in `add_chunk` and `update_chunk`.

## Component
- `api/apps/restful_apis/chunk_api.py` (`add_chunk`, `update_chunk`)

dearsishs:
## Summary
`add_chunk` / `update_chunk` raised an unhandled `TypeError` (HTTP 500) when `questions` contained non-string elements, because `question_tks` was built by joining the **raw** `questions` list while `question_kwd` (one line above) was already built from `str(q).strip()`.

## Root cause
```python
d["question_kwd"] = [str(q).strip() for q in req.get("questions", []) if str(q).strip()]
d["question_tks"] = rag_tokenizer.tokenize("\n".join(req.get("questions", [])))  # raw -> TypeError on non-str
```
`questions` is only validated to be a list, not list-of-strings, so `{"questions": [1, 2]}` makes `"\n".join([1, 2])` raise `TypeError`.

## Fix
Join the already-sanitized `question_kwd` instead of the raw list (in both `add_chunk` and `update_chunk`):
```diff
-    d["question_tks"] = rag_tokenizer.tokenize("\n".join(req.get("questions", [])))
+    d["question_tks"] = rag_tokenizer.tokenize("\n".join(d["question_kwd"]))
```
This reuses the consistent string coercion already applied to `question_kwd` and removes the divergence between the two lines.

## Files changed
- `api/apps/restful_apis/chunk_api.py`

## Verification
- `ruff check` — clean
- `"\n".join([1,2])` raises `TypeError` (reproduced in a standalone interpreter); `"\n".join(["1","2"])` does not. Local full pytest not run (heavy RAG deps); CI validates.

## Note
Implemented with LLM assistance (model: claude-opus-4-8).

Closes #15623


KevinHuSh (approved):

buua436 (commented):
LGTM. The fix correctly reuses the normalized `question_kwd` values when building `question_tks` in both `add_chunk` and `update_chunk`, preventing a `TypeError` for non-string question items and keeping both fields consistent.

Please consider adding a regression test covering mixed-type values such as integers, `null`, and whitespace-only strings to ensure both endpoints return successfully and tokenize the normalized questions consistently.

PR #15631, opened 2026-06-04
feat(go-cli): support batch model add/remove and optional embedding dimension

Hz-186:
## Summary

  This PR improves the Go CLI in two areas:

  1. It adds batch model management support, allowing multiple models to be added or removed in a single command.
  2. It makes the `dimension` argument optional for the `embed text` command.

  These changes keep the existing single-model and explicit-dimension behaviors compatible while making the CLI more convenient for common workflows.

  ## What Changed

  ### 1. Batch model add/remove support

  The CLI now supports operating on multiple model names provided in a single quoted string.

  Supported commands include:

```
add model 'x1 x2 x3' to provider 'vllm' instance 'test' with tokens 1024 chat think vision, token 2048 chat, token 1024 think vision;

drop model 'x1 x2 x3' from 'vllm' 'test';

remove model 'x1 x2 x3' from 'vllm' 'test';

```

  For add model, each config segment after with is matched to the corresponding model name by position.

  Example mapping:

  - x1 -> tokens 1024, chat + vision, thinking=true
  - x2 -> tokens 2048, chat
  - x3 -> tokens 1024, vision, thinking=true

  The existing single-model syntax remains supported.

  ### 2. Optional embedding dimension

  Previously, the Go CLI required dimension to be explicitly provided for embed text.

  Before:

  embed text 'what is rag' 'who are you' with 'model@test@provider' dimension 8192;

  Now both forms are supported:

  embed text 'what is rag' 'who are you' with 'model@test@provider' dimension 8192;

  embed text 'what is rag' 'who are you' with 'model@test@provider';

  When omitted, the CLI leaves dimension unset and relies on provider/backend behavior.

  ## Tests

  Added parser tests covering:

  - Multiple models with multiple config segments
  - Model type deduplication
  - Model/config count mismatch
  - Drop/remove multiple models
  - Optional embedding dimension parsing

### DIFF
--- a/internal/cli/user_command.go
+++ b/internal/cli/user_command.go
@@ -2681,20 +2681,9 @@ func (c *RAGFlowClient) AddCustomModel(cmd *Command) (ResponseIf, error) {
 		return nil, fmt.Errorf("instance name not provided")
 	}
 
-	modelName, ok := cmd.Params["model_name"].(string)
+	models, ok := cmd.Params["models"].([]map[string]any)
 	if !ok {
 		return nil, fmt.Errorf("model name not provided")
-	}
-
-	// chat, vision, embedding, rerank, tts, asr, ocr
-	modelTypes, ok := cmd.Params["model_types"].([]string)
-	if !ok {
-		return nil, fmt.Errorf("model type not provided")
-	}
-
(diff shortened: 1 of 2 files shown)


xugangqiang:
LGTM

Issue #15482, opened 2026-06-01
[Feature Request]: Enhance markdown chunking


yingfeng:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

Both builtin and ingestion pipeline have not been optimized for markdown chunking, which will lose structure information by default. 

These features are required by default:

- Preserve code blocks
- Maintain hierarchical section structure
- Keep table structure
- Smart overlap as a choice
- Handle nested fencing

### Describe implementation you've considered

Both builtin(naive.py) as well as pipeline are required to be improved at the same time.

### Documentation, adoption, use case

```Markdown

```

bitloi:
Hi @yingfeng, I did a quick code and issue-history pass on this issue. The request looks directionally valid, but it may be too broad to implement as one PR.

Confirmed current state:
- The concrete short-header splitting bug is already addressed by #15488.
- Heading-based Markdown chunking is also being explored in #15408.
- The remaining clearly actionable gap appears to be fenced-code preservation, especially nested/longer fences and tilde fences in `deepdoc/parser/markdown_parser.py`.

Suggested next step:
Can we split this into narrower acceptance criteria? I think a small first PR could focus only on preserving fenced code blocks during delimiter-based Markdown extraction, with tests for:
- backtick fences
- tilde fences
- longer outer fences containing inner shorter fences
- custom delimiter mode

I would avoid changing default hierarchy/table/overlap behavior in the same PR unless you confirm the exact desired chunk output.

KevinHuSh:
Yes, we can start with small PR, step by step.
The smaller PR is the better.
And also considering the pipleline, it seems this kind of pipleling have already fullfill headding concerned chunking.

<img width="2481" height="1563" alt="Image" src="https://github.com/user-attachments/assets/6b168089-50d0-4fc7-8db3-5e12a43e3a9b" />

bitloi:
Thank you! @KevinHuSh

[...]

PR #15632, opened 2026-06-04
Fix: preserve markdown tables during delimiter extraction

bitloi:

Markdown extraction can split tables row by row when delimiter-based extraction uses a newline delimiter. That loses table structure during chunking even though delimiters should still split normally outside tables.

This PR keeps the follow-up to #15482 intentionally narrow:

- preserve Markdown pipe tables during delimiter-based extraction
- preserve borderless pipe tables during delimiter-based extraction
- preserve multiline HTML tables during delimiter-based extraction
- keep delimiter splitting unchanged outside protected table ranges

Refs #15482

### Type of change

### Testing

- `ruff check deepdoc/parser/markdown_parser.py test/unit_test/deepdoc/parser/test_markdown_parser.py`
- `python3 run_tests.py -t test/unit_test/deepdoc/parser/test_markdown_parser.py`
- `git diff --check`


bitloi:
@KevinHuSh @yingfeng Ready for review.

KevinHuSh (approved):

PR #15672, opened 2026-06-04
feat[Go]: implement document upload, dataset metadata bulk-update APIs

### LINKED ISSUE #15671: [Subtask]: Implement remaining Document/metadata APIs in Go
### Parent Issue

Parent issue: #15240

### Detail of Subtask

Port the Document/metadata endpoints that are **not** connected to any open or closed PR. (Other document endpoints already have PRs — e.g. `POST /datasets/<id>/documents` #14994, `metadata/config PUT` #14273, `batch-update-status` #14337 — so they are excluded here.)

- [ ] `POST /api/v1/documents/upload`
- [ ] `PATCH /api/v1/datasets/<dataset_id>/documents/metadatas`
- [ ] `POST /api/v1/datasets/<dataset_id>/metadata/update`

Behavior must match the Python source: multipart upload handling and storage write for `upload`; bulk metadata application with validation for `metadatas` / `metadata/update`; tenant/KB access checks throughout.

- **Python source:** `api/apps/restful_apis/document_api.py` (and the shared metadata helpers it imports).
- **Go target (per reference PR #14790):** extend `internal/handler/document.go` et al. (a document handler skeleton already exists from earlier document-API migration work — align with it and with the in-flight #14994/#15499).
- `documents/upload` is multipart — mirror the Python file-size/type validation.
- Confirm none of these were silently picked up by #14994/#15499 before starting.
- Add unit + integration tests; `go build ./...`, `go vet ./...` clean.

hunnyboy1217:

Closes #15671 — ports three Document/metadata endpoints from Python (`api/apps/restful_apis/document_api.py`) to the Go API layer.

| Method | Path | Handler | Python source |
|--------|------|---------|---------------|
| POST | `/api/v1/documents/upload` | `UploadInfo` | `upload_info` |
| PATCH | `/api/v1/datasets/:dataset_id/documents/metadatas` | `UpdateDocumentMetadatas` | `update_metadata` |
| POST | `/api/v1/datasets/:dataset_id/metadata/update` | `MetadataBatchUpdate` | `metadata_batch_update` |

### Type of change

---

#### Implementation notes

**Files changed:**

```
internal/service/file.go      – UploadFromURL helper
internal/service/document.go  – BatchUpdateDocumentMetadata + supporting types
internal/handler/document.go  – three new handlers + shared batchMetadataUpdate
internal/router/router.go     – route registration
```

**Functional parity table:**

| Concern | Go behaviour |
|---------|-------------|
| `UploadInfo` — file path | `c.MultipartForm()` → `fileService.UploadFile(tenantID, "", files)` (existing method); single file returns scalar, multiple returns array — same as Python |
| `UploadInfo` — URL path | New `FileService.UploadFromURL`: validates `http`/`https` scheme, `GET` remote content (100 MB cap), stores to tenant root folder, returns same file-info map as the file path |
| Both inputs provided | Returns `CodeArgumentError` mirroring Python `get_error_argument_result` |
| Neither provided | Returns `CodeArgumentError` |
| `PATCH /metadatas` and `POST /metadata/update` | Identical logic via shared `batchMetadataUpdate(c)` — dataset access check via `datasetService.Accessible`, request body validated, delegated to `DocumentService.BatchUpdateDocumentMetadata` |
| `document_ids` validation | `documentDAO.GetAllDocIDsByKBIDs([datasetID])` builds allowed set; unknown IDs returned as error — mirrors `KnowledgebaseService.list_documents_by_ids` |
| `metadata_condition` filter | `metadataSvc.GetFlattedMetaByKBs([datasetID])` + `ApplyMetaFilter(flattedMeta, conditions, logic)` — reuses the same Go function already used by `RetrievalTest` |
| Selector intersection | When both `document_ids` and `metadata_condition` given, result is their intersection (Python `target_doc_ids & filtered_ids`) |
| Early exit | When `metadata_condition.conditions` present but nothing matched: returns `{updated:0, matched_docs:0}` |
| Per-doc update | `SetDocumentMetadata` for `updates`, `DeleteDocumentMetadata` for `deletes` |
| Response | `{updated: N, matched_docs: M}` matching Python's `get_result(data={...})` shape |


internal/service/document.go
```diff
@@ -1146,3 +1148,172 @@ func isTimeString(s string) bool {
...
+			}
+		}
+		updated++
+	}
+
+	return &BatchUpdateMetadataResult{Updated: updated, MatchedDocs: len(ids)}, nil
+}
+
+// parseMetadataConditions converts metadata_condition.conditions to MetaFilterConditions.
+func parseMetadataConditions(mc map[string]interface{}) []MetaFilterCondition {
```
Hz-186: here

internal/handler/document.go
```diff
@@ -886,3 +890,181 @@ func (h *DocumentHandler) StopParseDocuments(c *gin.Context) {
...
+// @Summary Upload document info
+// @Description Upload files via multipart form or supply ?url=... to fetch remotely.
+// @Tags documents
+// @Accept multipart/form-data
+// @Produce json
+// @Param file formData file false "File(s) to upload"
+// @Param url query string false "Remote URL to fetch"
+// @Success 200 {object} map[string]interface{}
+// @Router /api/v1/documents/upload [post]
+func (h *DocumentHandler) UploadInfo(c *gin.Context) {
```
Hz-186, line 1146: here

Hz-186 (changes requested):
Hi, @hunnyboy1217 

This change still has two parity gaps before it can be merged.

  First, the Go metadata batch-update path does not parse `metadata_condition` in the same shape as Python. Python sends conditions as `{name, comparison_operator, value}`, but `parseMetadataConditions()` currently only reads `{key|field, op, value}`. That means valid Python-style  requests will not be filtered correctly.

  Second, `POST /api/v1/documents/upload?url=...` is missing the Python SSRF guard. Python validates the remote URL with `assert_url_is_safe(url)` before fetching it, while Go currently only checks the scheme and then performs `http.Get` directly.

hunnyboy1217:
Hi, @Hz-186 ,

Hz-186 (changes requested):
@hunnyboy1217 

  The Go upload API is not compatible with the Python response shape. Timestamp naming differences are minor, but the actual returned fields differ: Python returns upload-info fields like `extension`, `mime_type`, and `preview_url`, while Go returns internal file-record fields like `parent_id`, `tenant_id`, `location`, `source_type`, and `type`.

  Please return the same non-timestamp fields as Python instead of the generic file response.

### Go Tests
  /usr/local/go/bin/go test ./internal/handler ./internal/service failed:

  - internal/service: passed
  - internal/handler: failed
      - TestBatchMetadataUpdate_NoPermission: expected code 102, got 0 at internal/handler/document_test.go:786
      - TestLinkToDatasets_MissingBothArgs: panic nil pointer via service.NewFile2DocumentService() at internal/handler/file_handler_test.go:70, which calls NewDocumentService() and dereferences config at internal/service/document.go:67

hunnyboy1217:
Hi, @Hz-186 ,

Hz-186 (changes requested):
@hunnyboy1217, I cant run server

In internal/service/document.go, inside the NewDocumentService function, declaring engineType using a blank string literal makes the compiler infer it as a standard string. 

This causes compilation errors because both server.GetConfig().DocEngine.Type and the DocumentService.engineType struct field expect the custom type server.

EngineType instead of a raw string. Please declare engineType using the server.

EngineType type directly to fix the type mismatch.

PR #15680, opened 2026-06-04
fix: private knowledgebase access check in Go DAO

### LINKED ISSUE #15647: KnowledgebaseDAO.Accessible does not check permission field, potentially exposing private knowledge bases
## Description

`KnowledgebaseDAO.Accessible()` in `internal/dao/kb.go` checks whether a user is associated with a knowledge base's tenant via `user_tenant` join, but does **not** check the `knowledgebase.permission` field. This means any tenant member can access a knowledge base even when its permission is set to `"me"` (private).

## Impact

This affects all API endpoints that use `KnowledgebaseDAO.Accessible()` for authorization, including:
- `/api/v1/datasets/<dataset_id>/documents`
- `/api/v1/dify/retrieval`
- Any other endpoint using KB access checks

A user who is a member of a tenant (via `user_tenant`) can access knowledge bases owned by other members of the same tenant, even if those KBs are marked as private.

## Python Reference

The Python equivalent (`KnowledgebaseService.accessible()`) first checks whether the requesting user is the direct owner (`kb.tenant_id == user_id`), and only falls through to tenant membership if `kb.permission == "team"`:

https://github.com/infiniflow/ragflow/blob/main/api/db/services/knowledgebase_service.py

## Suggested Fix

In `internal/dao/kb.go`, `Accessible()` should:
1. Allow if the user is the direct owner (`knowledgebase.created_by == user_id`)
2. Otherwise, only allow if `knowledgebase.permission == 'team'` AND user is a tenant member

The current query joins `user_tenant` but has no `permission` guard.

```go
func (dao *KnowledgebaseDAO) Accessible(kbID, userID string) bool {
    var count int64
    err := DB.Table(knowledgebase).
        Joins("JOIN user_tenant ON user_tenant.tenant_id = knowledgebase.tenant_id").
        Where("knowledgebase.id = ? AND user_tenant.user_id = ? AND knowledgebase.status = ?", kbID, userID, string(entity.StatusValid)).
        Count(&count).Error
    if err != nil {
        return false
    }
    return count > 0
}
```

The fix should add a condition that checks `knowledgebase.permission = 'team'` or `knowledgebase.created_by = ?` (the user is the direct owner).

jonathanchang31:
## Summary

This PR fixes `KnowledgebaseDAO.Accessible()` so private knowledge bases are not exposed to tenant members unless the knowledge base permission is explicitly set to `team`.

Before this change, any user joined to the same tenant could access a knowledge base because the access query only checked `user_tenant` membership and ignored the `knowledgebase.permission` field.

After this change:
- Owners can still access their own knowledge bases.
- Tenant members can access only team-shared knowledge bases.
- Tenant members cannot access private knowledge bases.
- Unrelated users cannot access team knowledge bases.

## Related Issue
Closes: #15647 
## Change Type

- [x] Bug fix
- [x] Security fix
- [x] Test
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Breaking change
## Real Behavior Proof

Issue validated in current code:

- `KnowledgebaseDAO.Accessible()` joined `user_tenant`.
- It checked tenant membership.
- It did not check `knowledgebase.permission`.
- This allowed tenant members to access private knowledge bases.
Fixed behavior is covered by new DAO tests:
```bash
PATH=/tmp/go/bin:$PATH go test ./internal/dao -run 'TestKnowledgebaseDAO' -count=1
```
Result:

```text
ok  	ragflow/internal/dao	0.323s
```
Full DAO package validation:

```bash
PATH=/tmp/go/bin:$PATH go test ./internal/dao -count=1
```
Result:

```text
ok  	ragflow/internal/dao	0.295s
```
Whitespace validation:
```bash
git diff --check
```
Result:

```text
Passed
```

## Checklist
- [x] Verified the issue exists in the current project.
- [x] Compared behavior with the Python reference implementation.
- [x] Applied a minimal DAO-level fix.
- [x] Added regression tests.
- [x] Ran focused DAO tests successfully.
- [x] Ran full DAO package tests successfully.
- [x] Ran `git diff --check`.
- [x] Avoided unrelated API or service changes.


jonathanchang31:
[@coderabbitai](https://github.com/coderabbitai) Could you plz review my PR quickly?

jonathanchang31:
@yingfeng Could you review my PR? I am going to contribute, I am looking forward to your support. Thanks!

PR #15681, opened 2026-06-05
fix(opensearch): stop mapping knn similarity threshold onto boost

ef-rintaro:

On the OpenSearch backend (`DOC_ENGINE=opensearch`), dense/vector retrieval is
silently dead for an entire KB: semantic queries and exact-id lookups return
"nothing found" even though chunks and their vectors are correctly indexed.
Only BM25/term matching works.

**Root cause.** `OSConnection.search()` maps `MatchDenseExpr.extra_options["similarity"]`
onto the OpenSearch knn `boost`. But `similarity` is a *minimum-similarity
threshold* (Elasticsearch knn semantics), not a score multiplier --
`es_conn.py` correctly passes it as the knn `similarity` argument, not `boost`.
`rag/nlp/search.py:_knn_scores()` runs a clean-cosine second pass and
deliberately passes `similarity=0.0` to disable thresholding. On OpenSearch that
became `boost=0.0`, multiplying every knn `_score` by zero, so `get_scores()`
returned all-zero, `vector_similarity` was always 0, and `Dealer.retrieval()`'s
`similarity_threshold` (default 0.2) then dropped every chunk.

### Fix

Keep the knn `boost` at 1.0 so the `_score` is the raw cosine. The similarity
threshold is still enforced downstream in `Dealer.retrieval()` (post_threshold),
so no behavior is lost. Main search ranking is unaffected: boost is a uniform
positive scale and the text query is only used as a knn filter, so candidates
are ranked by cosine either way.

### Type of change

### Affected backends

OpenSearch only. Elasticsearch already passes `similarity` as the knn
`similarity` argument (not `boost`); Infinity / OceanBase use their own
similarity paths.


JinHai-CN:
@ef-rintaro would you please resolve the conflicts?

PR #15696, opened 2026-06-05
fix: keep max pagerank for repeated n-hop edges

### LINKED ISSUE #15695: AnalyzeNHopPaths uses last-wins instead of max for PageRank
## Problem

```go
```

```python
```

```go
```

## Impact

## Background

## Suggested Fix

Fix both sides together:

**Go** (`kg_scoring.go`):
```go
}
```

**Python** (`search.py`):
```python
```

## Priority

he-yufeng:
## Summary

Fixes #15695.

The Python GraphRAG path already accumulates similarity when several N-hop paths produce the same edge, but PageRank was overwritten by the last path. That makes ranking depend on path order for repeated edges.

This keeps the strongest PageRank seen for a repeated edge in the Python implementation:

- `rag/graphrag/search.py`

The similarity score still accumulates exactly as before.

## To verify

- `python -m py_compile rag\graphrag\search.py`
- `git diff --check`
- `git diff --stat upstream/main` -> only `rag/graphrag/search.py`

I originally included the Go implementation too, but removed it after maintainer feedback because the Go version is still under development and not released yet.

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -184,7 +184,9 @@ class KGSearch(Dealer):
                         nhop_pathes[(f, t)]["sim"] += ent["sim"] / (2 + i)
                     else:
                         nhop_pathes[(f, t)]["sim"] = ent["sim"] / (2 + i)
-                    nhop_pathes[(f, t)]["pagerank"] = wts[i]
+                    nhop_pathes[(f, t)]["pagerank"] = max(
+                        nhop_pathes[(f, t)].get("pagerank", 0), wts[i]
+                    )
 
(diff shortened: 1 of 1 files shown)


xugangqiang:
@he-yufeng  
go version is under development (not released yet). Please work on python version. Thank you.

he-yufeng:
Thanks for the clarification. I removed the Go changes and force-pushed this as a Python-only fix in `e29530aec`.

Current diff is now only:

```text
rag/graphrag/search.py | 4 +++-
```

Validated locally:

```text
python -m py_compile rag\graphrag\search.py
git diff --check
git diff --stat upstream/main
```

PR #15713, opened 2026-06-05
fix: take max PageRank across paths in AnalyzeNHopPaths

### LINKED ISSUE #15695: AnalyzeNHopPaths uses last-wins instead of max for PageRank

RTCartist:

When two N-hop paths contribute the same edge, `PageRank` was assigned with plain `=`, so it kept whichever value was written last. In Go that's non-deterministic (randomized map iteration), and in both languages it can clobber a stronger weight with a weaker one. `Sim` already accumulates correctly; only `PageRank` had this bug.

Fix takes the max across contributing paths:

- `internal/service/kg_scoring_funcs.go` — guarded `if i < len(weights) && weights[i] > es.PageRank`
- `rag/graphrag/search.py` — `max(existing_or_0, wts[i])`

Closes #15695.

### Type of change

### Tests

Added `TestAnalyzeNHopPaths_PageRankTakesMaxAcrossPaths` covering the multi-path case both orderings of the outer `range` map produce the same result.

### DIFF
--- a/internal/service/kg_scoring_funcs.go
+++ b/internal/service/kg_scoring_funcs.go
@@ -38,7 +38,11 @@ func AnalyzeNHopPaths(entsFromQuery map[string]*KGEntity) map[Edge]EdgeScore {
 				edge := Edge{From: f, To: t}
 				es := nhopPathes[edge]
 				es.Sim += ent.Similarity / (2.0 + float64(i))
-				if i < len(weights) {
+				// Take the max across paths that contribute the same edge.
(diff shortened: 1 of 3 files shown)


RTCartist:
This issue seems already been solved in #15696. Close

PR #15719, opened 2026-06-06
fix(deepdoc): don't crash JSON parsing on a top-level scalar value

### LINKED ISSUE #15718: JSON parser raises IndexError on a top-level scalar value, failing document ingestion
### Summary
Uploading a syntactically valid `.json`/`.jsonl`/`.ldjson` document whose top-level value is a JSON **scalar** (`42`, `"text"`, `true`, `3.14`, `null`) crashes the document-parse task with an uncaught `IndexError`.

### Where
`deepdoc/parser/json_parser.py`:
- `_set_nested_dict` (lines 48-53): `d[path[-1]] = value` raises `IndexError` when `path == []`.
- `_json_split` else-branch (line 96): `self._set_nested_dict(chunks[-1], current_path, data)` is called with `current_path == []` for a top-level scalar.
- Caller `rag/app/naive.py` (`_parse_json`) only catches `json.JSONDecodeError`, so the `IndexError` propagates.

### Root cause / trace
For a scalar top-level value, `is_jsonl_format()` returns `False` (because `json.loads(txt)` succeeds), so parsing routes to `_parse_json` → `split_json` → `_list_to_dict_preprocessing` leaves the scalar unchanged → `_json_split(scalar, [], ...)` hits the single-item `else` branch → `_set_nested_dict(chunks[-1], [], data)` → `path[-1]` on an empty list → **IndexError**. (Top-level lists work because preprocessing converts them to index-keyed dicts; only scalars break.) `_parse_json` doesn't catch `IndexError`, so it fails the whole parse task.

### Why it's a real bug (not intended)
`_parse_json`/`_parse_jsonl` deliberately swallow malformed JSON via `JSONDecodeError`. A *valid* JSON scalar document raising an unrelated `IndexError` is clearly unintended — valid JSON should never make the parser throw.

### Reproduction
Top-level `42`, `"hello"`, `true`, `3.14`, `null` all raise `IndexError: list index out of range`; `[1,2,3]` and `{"a":1}` succeed.

### Fix
Guard the empty-path case in `_set_nested_dict` (return instead of indexing an empty path). Valid scalar documents then parse without raising; dict/list documents are unaffected. PR attached.

seekmistar01:
## Summary
Closes #15718

A syntactically valid `.json`/`.jsonl` document whose top-level value is a JSON scalar (`42`, `"text"`, `true`, `3.14`, `null`) crashes the document-parse task with an uncaught `IndexError`. `_json_split`'s single-item branch calls `_set_nested_dict(chunks[-1], [], data)` with an empty path, so `path[-1]` raises; `_parse_json` only catches `json.JSONDecodeError`, so it propagates and fails the whole parse task on user-supplied valid input.

## Change
- `deepdoc/parser/json_parser.py` — guard the empty-path case in `_set_nested_dict` (return instead of indexing an empty path). A bare top-level scalar then yields no section instead of crashing.

## Why it's safe
- Only the previously-crashing empty-path branch changes; non-empty paths (every dict/list document) behave exactly as before.
- The parser's intent is to swallow malformed input via `JSONDecodeError`; this restores that for valid scalar documents.

## Verification
- `python -m py_compile deepdoc/parser/json_parser.py` — OK.
- Reproduced the core `_json_split`/`_set_nested_dict` path standalone: top-level `42`/`"hello"`/`true`/`3.14`/`null` raise `IndexError` before the change and are handled after; `{"a":1}`/`[1,2,3]` unaffected.


xugangqiang:
@seekmistar01 Thanks for the fix — I reviewed with a few thoughts, and a request before merge.

**1. Problem (as stated).** A document whose top-level JSON value is a scalar (`42`, `"text"`, `true`, `3.14`, `null`) reaches `_json_split`'s single-item branch with an empty `path`, so `_set_nested_dict(chunks[-1], [], data)` evaluates `path[-1]` on an empty list and raises `IndexError`. Because `_parse_json` only catches `json.JSONDecodeError`, the error propagates and fails the whole parse task on syntactically valid, user-supplied input.

**2. Does it actually fix it?** Yes, verified. With the guard `if not path: return` in `_set_nested_dict` (`deepdoc/parser/json_parser.py`), top-level scalars no longer raise. I reproduced the old behaviour (all five scalar types raise `IndexError`) and the new behaviour (no raise, `{"a":1}`/`[1,2,3]` unchanged). Non-scalar documents are completely unaffected because the only caller that passes an empty path is the scalar case.

**3. Is the fix good?** It is minimal and safe — I like that. Two downsides: (a) it **silently drops** the scalar content, and (b) there is no regression test. In a RAG system, a document that "parses successfully" but yields 0 chunks is the dangerous failure mode: the doc silently vanishes from the knowledge base with no signal to the user.

**4. Better approach?** Keep the guard (it prevents the crash robustly), but also surface a non-fatal warning to the user. RAGFlow already has exactly this channel: `callback(msg="[WARN] ...")` (without `prog=-1`) appends to `progress_msg`, which the web UI renders as the parsing log (`web/src/pages/dataset/dataset/parsing-card.tsx`), **without** flipping the document to `FAIL`. The parser itself has no access to the `callback`, so the warning should be emitted from the JSON branch in `rag/app/naive.py:1136`, which does. The ideal result: document stays `DONE` with `chunk_num = 0`, but the parsing log explains *why*.

**5. Regression risk?** None for dict/list inputs or malformed JSON (still swallowed by `JSONDecodeError`). For scalars the behaviour changes from "crash / fail whole task" to "soft-skip with 0 chunks" — strictly an improvement. The only risk is the silent data loss noted above, which the `[WARN]` resolves.

**Request before merge:**
1. **Add regression tests** covering top-level scalars (`42`, `"text"`, `true`, `3.14`, `null`) as well as dict and list inputs, asserting no `IndexError` is raised and that normal behaviour is unchanged. This prevents the crash from silently resurfacing.
2. **Make the 0-chunk reason visible in the UI.** Please wire the scalar case to emit a `[WARN]` (via the `callback` in `rag/app/naive.py`) into `progress_msg`, e.g. `[WARN] Document is a top-level JSON scalar; no chunks were extracted.` Then verify in the dataset document list that a scalar JSON document shows `DONE` with `chunk_num = 0` **and** that the parsing log explains why. The doc should remain `DONE`, not `FAIL` — we want to inform, not fail.

This keeps the batch safe (no crash, no failed docs) while giving the user the visibility they currently lack. Happy to help refine the wording or the tests.

PR #15721, opened 2026-06-06
fix(nlp): tokenize content_tks by whitespace in FulltextQueryer.paragraph

### LINKED ISSUE #15720: FulltextQueryer.paragraph iterates tokens character-by-character, breaking auto-tagging match queries
### Summary
`FulltextQueryer.paragraph` iterates its token string **character by character** instead of splitting on whitespace, so the auto-tagging match query is built from single letters — silently breaking KB auto-tagging.

### Where
`rag/nlp/query.py:226` (in `paragraph`):
```python
def paragraph(self, content_tks: str, keywords: list = [], keywords_topn=30):
    if isinstance(content_tks, str):
        content_tks = [c.strip() for c in content_tks.strip() if c.strip()]   # iterates CHARACTERS
    tks_w = self.tw.weights(content_tks, preprocess=False)
```
Live caller: `rag/nlp/search.py` `Dealer.tag_content`, which passes `doc["title_tks"] + " " + doc["content_ltks"]` (space-joined tokens).

### Root cause
`content_tks` is a space-separated token string (annotated `content_tks: str`). Iterating a `str` yields characters, so `"machine learning model"` becomes `['m','a','c','h',...]` (20 single chars) rather than `['machine','learning','model']`. Those single characters are fed to `tw.weights(..., preprocess=False)` (which expects real tokens), producing meaningless weights and a garbage `MatchTextExpr`. Every other method in this file (`similarity`, `question`) tokenizes a token string with `.split()` — this is a `.strip()`-vs-`.split()` typo.

### Impact
`paragraph()` backs `Dealer.tag_content`, the knowledge-base auto-tagging feature that computes `TAG_FLD` tag features for chunks. The per-character "tokens" make tag retrieval/scoring effectively broken (wrong/empty aggregations), silently degrading auto-tagging for every document in tag-enabled KBs.

### Why it's a real bug (not intended)
Character iteration has no plausible purpose; the type annotation, the caller's input, and `tw.weights`'s token contract all show the intent is `.split()`. No test references `paragraph`.

### Fix
Change `content_tks.strip()` to `content_tks.split()`. Reproduced: `"machine learning model"` yields 20 single-character entries today vs the intended 3 tokens after the fix. PR attached.

seekmistar01:
## Summary
Closes #15720

`FulltextQueryer.paragraph` normalized its `content_tks` token string with `[c.strip() for c in content_tks.strip() ...]`, which iterates the string **character by character** — `"machine learning model"` becomes 20 single characters instead of 3 tokens. Those single chars are fed to `tw.weights(..., preprocess=False)`, producing meaningless term weights and a garbage `MatchTextExpr`.

`paragraph()` backs `Dealer.tag_content` (the KB auto-tagging feature), so tag retrieval/scoring is silently broken for tag-enabled knowledge bases. Every other method in this file tokenizes with `.split()` — this is a `.strip()`-vs-`.split()` typo.

## Change
- `rag/nlp/query.py` — change `content_tks.strip()` to `content_tks.split()` in the `paragraph` token-normalization line.

## Why it's safe
- The caller passes a space-separated token string; `.split()` recovers the real tokens, matching the contract of `tw.weights` and the `.split()` tokenization used by the sibling methods (`similarity`, `question`).
- No behavior depends on the per-character expansion.

## Verification
- `python -m py_compile rag/nlp/query.py` — OK.
- Demonstrated: `"machine learning model"` → 20 single-character entries before, 3 real tokens after. No test references `paragraph`.

### DIFF
--- a/rag/nlp/query.py
+++ b/rag/nlp/query.py
@@ -223,7 +223,7 @@ class FulltextQueryer(QueryBase):
 
     def paragraph(self, content_tks: str, keywords: list = [], keywords_topn=30):
         if isinstance(content_tks, str):
-            content_tks = [c.strip() for c in content_tks.strip() if c.strip()]
+            content_tks = [c.strip() for c in content_tks.split() if c.strip()]
         tks_w = self.tw.weights(content_tks, preprocess=False)
 
         origin_keywords = keywords.copy()


qinling0210 (approved):
LGTM

PR #15731, opened 2026-06-07
fix(connectors): enforce WebDAV numeric string size limits

### LINKED ISSUE #15724: [Bug]: webdav data source has no file size limit
### Self Checks

docker realse tag

### RAGFlow image version

v0.25.5

```Markdown
docker compose with document sync over webdav connector
```

### Actual behavior

I noticed that webdav connector seem to check for supported file type but not for filesize...

had it happen that some large files .csv and .txt, like logs, 3d pointclouddata which have easily 100MB and more get into the chunking pipeline totally glogging things up. for now i patched it to leave out csv and limit xlsx table to 20mb, but I realised a general file size limit is needed

since there seem to be one for manual uploads, the same limit should apply to enquing files from webdav. most table files or text files for example larger a few MB just dont make sense to chunk

### Expected behavior

respect same file size limit in webdav connector as anywhere else, reject / ignore files larger than that

### Steps to reproduce

```Markdown
just put any larg txt or csv file in the webdav served folder
```

### Additional information

_No response_

oktofeesh1:
## Summary
- Normalize WebDAV file-size metadata before applying the sync size threshold.
- Enforce the same threshold for numeric string sizes in both document sync and slim snapshot paths.
- Add focused WebDAV unit coverage for size parsing and over-threshold skips.

## Why
Some WebDAV servers return file sizes from PROPFIND metadata as strings. The previous threshold check only handled integer values, so oversized files could still be downloaded and sent into the chunking pipeline.

Closes #15724.

## Validation
- `uv run --no-project --with pytest --with pytest-asyncio pytest test/unit_test/data_source/test_webdav_connector_unit.py -q`
- `uvx ruff check common/data_source/webdav_connector.py test/unit_test/data_source/test_webdav_connector_unit.py`
- `python -m compileall -q common/data_source/webdav_connector.py test/unit_test/data_source/test_webdav_connector_unit.py`
- `git diff --check`

### DIFF
--- a/common/data_source/webdav_connector.py
+++ b/common/data_source/webdav_connector.py
@@ -62,8 +62,7 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
         return is_accepted_file_ext(file_ext, self._build_extension_type())
 
     @staticmethod
-    def _get_size_bytes(file_info: dict[str, Any]) -> int | None:
-        size_bytes = file_info.get("size")
+    def _coerce_size_bytes(size_bytes: Any) -> int | None:
         if isinstance(size_bytes, bool):
             return None
         if isinstance(size_bytes, int):
@@ -74,6 +73,22 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
                 return None
             parsed_size = int(size_text)
             return parsed_size if parsed_size >= 0 else None
+        return None
+
+    @classmethod
+    def _get_size_bytes(cls, file_info: dict[str, Any]) -> int | None:
+        # webdav4's Client.ls(detail=True) reports the size under "content_length"
+        # (see webdav4.multistatus.Response.as_dict); other servers/libraries or
+        # webdav4's fsspec wrapper may instead use "size" or the raw
+        # "getcontentlength" property. Try each so the size guard isn't silently
+        # skipped — otherwise file_info.get("size") is always None and every file
+        # trips the missing-metadata warning.
+        for key in ("size", "content_length", "getcontentlength"):
+            if key not in file_info:
+                continue
+            size_bytes = cls._coerce_size_bytes(file_info[key])
+            if size_bytes is not None:
+                return size_bytes
         return None
 
     @staticmethod
--- a/common/data_source/webdav_connector.py
+++ b/common/data_source/webdav_connector.py
@@ -3,6 +3,7 @@ import logging
 import os
 from datetime import datetime, timezone
 from typing import Any, Optional
+from urllib.parse import urlsplit
 
 from webdav4.client import Client as WebDAVClient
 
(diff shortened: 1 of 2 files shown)


oktofeesh1:
Hi @Haruko386 @JinHai-CN mind reviewing when you have a moment? Ty

common/data_source/webdav_connector.py
```diff
@@ -228,10 +257,10 @@ def _yield_webdav_documents(
...
             
-            size_bytes = file_info.get('size', 0)
+            size_bytes = self._get_size_bytes(file_info)
             if (
                 self.size_threshold is not None
-                and isinstance(size_bytes, int)
+                and size_bytes is not None
                 and size_bytes > self.size_threshold
             ):
                 logging.warning(
```
DeepeshAlwani, line 289: Two small suggestions for this block:

**1. Inconsistent log level**

The first path (~line 257) uses `logging.warning` for the over-threshold case, but this block uses `logging.debug` for the same condition. Oversized files would be invisible in default log output here — worth making consistent:

```python
logging.warning(
    f"{file_identifier} exceeds size threshold of {self.size_threshold} "
    f"(size_bytes={size_bytes}). Skipping."
)
```

**2. Silent drop when `size_bytes is None`**

Neither location logs anything when `_get_size_bytes` returns `None` (missing metadata). The file is skipped with no trace in the logs. As the original reporter noted in #15724, this is especially painful for long-running batch jobs — a sync over thousands of files can silently drop files with no indication of why.

Suggested addition before the existing threshold check:

```python
if self.size_threshold is not None and size_bytes is None:
    file_identifier = self._get_log_file_identifier(file_info, file_path)
    logging.warning(
        f"{file_identifier}: size metadata missing from WebDAV server response, "
        f"skipping to avoid processing potentially large files."
    )
    continue
```

Happy to submit a follow-up PR for both if you'd prefer to keep this one scoped.

---
*Note: findings are my own, used AI assistance only to improve grammar and structure.*

oktofeesh1: I've fixed these, thank you for this!

DeepeshAlwani (approved):
Thanks for addressing the suggestions. The warning on missing metadata and the consistent log level both look good. LGTM.

rejuce:
been tryting these changes...

now i pretty much get 
2026-06-09 21:33:39,055 WARNING  40 Rahmenauft\xe4ge.xlsx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 ~$ad ring.pdf: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 ~$justment Measuringhead CCI2D.pdf: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 ~$nstellung MK CCI2D.pdf: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 ~$rtungseinheit FDM20.pdf: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 Reparaturbericht_IDS25_140_Modul_2020_02_03_VS-3762.docx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 Reparaturbericht_Systemeinheit_2021_10_26_2S-0342.docx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 Reparaturbericht_Systemeinheit_2021_10_26_2S-0342.pdf: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 Reparaturbericht_Trigger_2020_04_14_VS-0087.docx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,055 WARNING  40 Traversen_PC-rep.und mod. .xlsx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

2026-06-09 21:33:39,056 WARNING  40 Resturlaubsplanung 2020.xlsx: size metadata missing from WebDAV server response, skipping to avoid processing potentially large files.

on every file...

i used curl request on my webdav server




curl -i -X PROPFIND http://localhost:8081/ \
  -H "Depth: 1" \
  --user webdav:webdavpassword
HTTP/1.1 207 Multi-Status
Accept-Ranges: bytes 


**which properly returns for every file a size** 
<D:href>/gitea-dump-1601383580.zip</D:href>
...
<D:getcontentlength>29235489</D:getcontentlength>
<D:href>/logo.png</D:href>
...
<D:getcontentlength>12880</D:getcontentlength>

**getcontentlength fields. something is still wrong here**

DeepeshAlwani:
Thanks for testing and sharing the logs — this actually points to another gap in `_get_size_bytes`.

The method only checks `file_info.get("size")`, but `webdav4`'s `ls(detail=True)` doesn't always map `getcontentlength` to `"size"` — depending on the server and library version it may return it under `"content_length"` or `"getcontentlength"` instead. So even though your server is correctly returning size in the PROPFIND response, `file_info.get("size")` comes back `None` for every file, triggering the missing metadata warning on everything.

The fix would be to check multiple possible keys:

```python
size_bytes = (
    file_info.get("size")
    or file_info.get("content_length")
    or file_info.get("getcontentlength")
)
```

This would make the check resilient across different WebDAV server implementations without changing any other behavior. Happy to put up a follow-up PR for this if the maintainers agree.

---
*Note: findings are my own, used AI assistance only to improve grammar and structure.*

rejuce:
this seems to work, thank you

DeepeshAlwani:
Happy to open a follow-up PR with this fix if preferred rather than updating this one.

oktofeesh1:
> Happy to open a follow-up PR with this fix if preferred rather than updating this one.

@DeepeshAlwani Great catch — confirmed and pushed. You're right that webdav4's Client.ls(detail=True) reports the size under content_length (webdav4.multistatus.Response.as_dict), not size — the size key only exists in webdav4's fsspec wrapper, which this connector doesn't use. So _get_size_bytes was returning None for every file and the guard never fired. I went with checking size → content_length → getcontentlength (first valid value wins, so zero-byte files and existing servers still work) and added unit tests for the content_length/getcontentlength paths. Thanks for digging into this with the logs.

PR #15738, opened 2026-06-07
fix(retrieval): keep rerank windows page aligned

oktofeesh1:
## Summary

Keeps reranked retrieval windows aligned with page boundaries while respecting provider-safe candidate limits.

## What changed

- Tighten `Dealer._rerank_window()` so active external reranking caps candidate windows at 64.
- Reject reranker-mode `page_size` values above 64 with a clear `_rerank_window` error instead of silently slicing an undersized pool.
- Preserve page-size alignment where possible for deep-pagination rerank windows.
- Add regression tests for aligned rerank windows, boundary pages, non-rerank behavior, small `top` values, and oversized rerank pages.

## Why

When the rerank candidate window and page size drift apart, deep pagination can fetch one block and then slice a different in-block range, causing short or missing pages. Failing fast on unsupported rerank page sizes prevents a 100-item page from being silently sliced out of a 64-candidate rerank pool.

## Validation

- `uv run --python 3.13 --with 'pytest<9' --with pytest-asyncio pytest test/unit_test/rag/test_rank_feature_scores.py -q`
- `git diff --check`

## Notes

- `python3 -m pytest` was not available because the project environment did not include pytest, so validation used `uv run` with pytest and pytest-asyncio supplied explicitly.


oktofeesh1:
Hi @Haruko386 @JinHai-CN this is ready for review 👍

oktofeesh1:
hi @Haruko386 @JinHai-CN @wangq8 do you mind reviewing this when you have a minute? Thank you!

wangq8:
There is a new parameter `rerank_candidates_count` introduced, this rerank window no longer needed.

PR #15745, opened 2026-06-08
fix: refine think stream parsing

buua436:
Refine the stream parsing for `<think>` / `</think>` so MiniMax and DeepSeek-style chunking both flush in the right order without mixing think and answer buffers.

### Type of change

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -1472,9 +1472,9 @@ async def _stream_with_think_delta(stream_iter, min_tokens: int = 16):
             return None
         state.answer_buffer += text
         if num_tokens_from_string(state.answer_buffer) >= min_tokens:
-                out = state.answer_buffer
-                state.answer_buffer = ""
-                return out
+            out = state.answer_buffer
+            state.answer_buffer = ""
+            return out
         return None
 
     def _flush_think_buffer():
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -1436,12 +1436,12 @@ class _ThinkStreamState:
     def __init__(self) -> None:
         self.full_text = ""
         self.last_idx = 0
-        self.endswith_think = False
-        self.last_full = ""
         self.last_model_full = ""
         self.in_think = False
-        self.buffer = ""
-        self.post_think_text = ""
+        self.close_pending = False
+        self.pending_after_close = ""
+        self.think_buffer = ""
+        self.answer_buffer = ""
 
 
 def _extract_visible_answer(text: str) -> str:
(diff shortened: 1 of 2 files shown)


wangq8 (changes requested):
## Review: `fix: refine think stream parsing` (#15745)

This PR rewrites the `<think>`/`</think>` streaming parser in `_stream_with_think_delta`, replacing the old delta-based approach with a new loop that maintains separate `think_buffer` and `answer_buffer` buffers and explicitly marks tag boundaries. The approach is cleaner, but there are **3 bugs** that need fixing before merge.

---

### Bug 1 — Missing `</think>` marker at end-of-stream (lines 1574-1575)

```python
    if state.close_pending:
        state.in_think = False
```

When the stream ends while `close_pending` is True, the code only flips `in_think` but **never yields `("marker", "</think>", state)`**. Downstream consumers (lines 325-328, 889-892, 1683-1686) rely on this marker to set `end_to_think: True` on the response — without it, the client never gets the end-of-think signal and may misrender the answer.

Compare with the old code which did emit it:
```python
    if state.endswith_think:
        yield ("marker", "</think>", state)
```

**Fix:** yield the `</think>` marker before clearing `in_think`:
```python
    if state.close_pending:
        yield ("marker", "</think>", state)
        state.in_think = False
```

---

### Bug 2 — Answer buffer not flushed when switching to think mode (lines 1543-1548)

```python
                if not state.in_think:
                    think_piece = _flush_think_buffer()
                    if think_piece is not None:
                        yield ("text", think_piece, state)
                    state.in_think = True
                    yield ("marker", "<think>", state)
```

When entering think mode from answer mode, the code flushes `_flush_think_buffer()` but that buffer is empty (we were in answer mode). Meanwhile, `state.answer_buffer` may contain text from before the `<think>` tag that hasn't reached the `min_tokens` threshold yet. This text becomes **stuck** — never emitted, and will eventually be flushed as answer text at end-of-stream, out of order with the think content.

**Fix:** Add a `_flush_answer_buffer` helper and call it here:
```python
    def _flush_answer_buffer():
        if not state.answer_buffer:
            return None
        out = state.answer_buffer
        state.answer_buffer = ""
        return out

    # then in the <think> handler:
    if not state.in_think:
        answer_piece = _flush_answer_buffer()
        if answer_piece is not None:
            yield ("text", answer_piece, state)
        state.in_think = True
        yield ("marker", "<think>", state)
```

---

### Issue 3 — Indentation inconsistency in `_emit_text` (lines 1474-1477)

```python
        if num_tokens_from_string(state.answer_buffer) >= min_tokens:
                out = state.answer_buffer
                state.answer_buffer = ""
                return out
```

The `if` body is indented by 8 spaces (to column 16) while the rest of the function uses 4-space increments per level (the think branch equivalent is at column 12→16). Python accepts this, but it's inconsistent and will confuse readers. Align to 4-space-per-level.

---

### Minor: Dead field `last_idx`

`_ThinkStreamState.last_idx` (line 1438) was used by the removed `_next_think_delta` function. It is never read in the new code and should be removed from `__init__`.

Issue #15743, opened 2026-06-08
Go API handlers systemically leak err.Error() to API consumers


xugangqiang:
## Problem

The Go API handler package contains **153 instances** of `err.Error()` returned directly in HTTP response bodies, across **16 handler files** (~13,000 lines of handler code).

```text
providers.go     45 instances
user.go          19 instances
tenant.go        16 instances
search.go         9 instances
chunk.go          9 instances
chat_session.go   9 instances
chat.go           9 instances
connector.go      8 instances
mcp.go            7 instances
document.go       7 instances
agent.go          5 instances
llm.go            4 instances
memory.go         3 instances
stats.go          1 instance
oauth_login.go    1 instance
file.go           1 instance
```

## Risk

Attackers can extract internal details by triggering errors:
- Database connection strings and table names
- Elasticsearch index names and query structures
- Filesystem paths
- Internal service ports and addresses
- Third-party API configuration fragments

## Current State

`searchbot.go` is one of the few correctly implemented handlers — it uses `common.Warn()` to log real errors and returns generic messages. But this is the **exception, not the rule**.

## Root Cause

1. **No sentinel error system** — service-layer errors are not typed; handlers cannot use `errors.Is()` to classify errors
2. **No unified error response middleware** — no catch-all to intercept unhandled internal errors
3. **No security linting** — no rule preventing `err.Error()` from reaching HTTP response bodies

### Phase 1: Stop the bleeding
1. Define sentinel error types in `service/errors.go` (NotFound, Unauthorized, InvalidArgument, etc.)
2. Establish error code ↔ HTTP status code mapping
3. Add `golangci-lint` rule to forbid `err.Error()` in `c.JSON()` calls

### Phase 2: Patch
4. Fix handlers in priority order: `chunk.go` → `search.go` → `chat*.go` → remainder
5. Pattern per file: service returns sentinel error → handler uses `errors.Is()` → typed error code + generic message + `common.Warn()` log

### Phase 3: Harden
6. Add Gin recovery middleware to catch unhandled panics and unknown errors, returning 500 + generic message
7. Add security scanning to CI pipeline

## Reference

`internal/handler/searchbot.go` demonstrates the correct pattern.

(Identified during CTO-level code review of `feat/searchbots-retrieval-test` branch)

bitloi:
@dosu

PR #15746, opened 2026-06-08
fix(go): redact internal handler errors

bitloi:

Refs #15743

Some Go API handlers return raw `err.Error()` strings in `CodeServerError` responses. Those errors can include internal backend details such as database, storage, search engine, or host information.

This PR adds a small shared `jsonInternalError` helper for handler-level internal failures. The helper logs the raw error server-side with request method/path context, then returns the existing generic `common.CodeServerError.Message()` to API clients.

This first slice migrates the existing `jsonError(c, common.CodeServerError, err.Error())` production call sites in agent, dataset graph, file, and system handlers. It intentionally does not close the full issue because direct `c.JSON` error responses in other handlers remain for follow-up PRs.

### Type of change

### Tests

- `/root/go/bin/go test ./internal/handler -count=1`


bitloi:
Hi @xugangqiang would you please review the PR and share your feedback?

bitloi:
Hi @wangq8, @JinHai-CN, @yingfeng, would you please review my PR and give me your feedback?

internal/handler/kb.go
```diff
@@ -59,6 +60,17 @@ func jsonError(c *gin.Context, code common.ErrorCode, message string) {
...
 
+// jsonInternalError logs the original error while returning a generic message
+// to avoid exposing internal implementation details in API responses.
+func jsonInternalError(c *gin.Context, err error) {
+	common.Warn("handler internal error",
+		zap.Error(err),
+		zap.String("method", c.Request.Method),
+		zap.String("path", c.Request.URL.Path),
+	)
+	jsonError(c, common.CodeServerError, common.CodeServerError.Message())
```
xugangqiang: Where is the best place for jsonInternalError?

bitloi: Moved `jsonInternalError` to `internal/handler/error.go` since it is shared across handlers. Thanks for the pointer.

bitloi:
Hi @yingfeng would you please review my PR?

PR #15758, opened 2026-06-08
ci: compile go test files to catch test-only compile errors

### LINKED ISSUE #15751: Go test files not compiled in CI — missing import undetected
## Problem (Internal tracking only)

CI only runs `go build` (via `./build.sh --go`), which compiles production binaries but **skips `_test.go` files entirely**. This means missing imports, type errors, or any other compile-time errors in test files are never caught.

Example: `metadata_filter_test.go` was merged with a missing `"sort"` import — the `sort.Strings()` call was added but the import line was not. CI gave a green checkmark because `go build` never touched the test file.

## Impact

- Test files can silently rot — developers discover breakage only when running `go test` locally
- New contributors who run `go test ./...` see unrelated failures and waste time debugging
- PRs that modify test-only code can pass CI despite being broken

## Proposed Fix

Add `go test -count=1 ./internal/...` to the CI workflow after the Go build step.

PR: #15750

> **Note:** Internal discussion. Implementation in progress.

Lobster-0429:

CI only runs `go build` (via `./build.sh --go`), which compiles production
binaries but **skips `_test.go` files entirely**. Missing imports, type errors,
or any compile-time error that lives only in test code is therefore never caught
— CI stays green while `go test ./...` is broken locally.

Concrete example from the issue: `metadata_filter_test.go` was merged with a
missing `"sort"` import. CI passed because `go build` never touches test files.

This PR adds a **Compile go test files** step to `.github/workflows/tests.yml`,
immediately after the Go build:

```yaml
- name: Compile go test files
  run: |
    GOPROXY=${GOPROXY:-...} CGO_ENABLED=1 go test -run='^$' ./internal/...
```

`-run='^$'` compiles every test package but runs **zero** tests. That catches
the missing-import / type errors `go build` skips, **without** gating CI on the
many internal tests that are not green yet (being fixed separately — see
@xugangqiang's note below). The step can be switched to actually run the suite
(`go test -count=1 ./internal/...`) once those failures are resolved.

`CGO_ENABLED=1` + the C++ tokenizer static library built in the previous step
are required to link the bindings, same as `./build.sh --go`.

Closes #15751.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


Lobster-0429:
Hi, @xugangqiang ,
Please review this PR. Thank you!

xugangqiang:
Thank you but the original issue is used for internal tracking only. And actaully there will be lots of failures need to be fixed when test is enabled and on it to fix.

Lobster-0429:
Thanks @xugangqiang — understood on both points. I’ve updated the step to `go test -run='^$' ./internal/...`, which only **compiles** the test files (catching the missing-import / type errors this targets) and runs no tests, so it won’t turn CI red on the pre-existing failures you’re fixing. It can be flipped to run the full suite once those are green.

Lobster-0429:
@Haruko386 , Could you please review this PR?

xugangqiang:
@JPette1783  if you check latest code, this feature should already be there. Thank you.

JinHai-CN:
@Lobster-0429 If you are still working on this PR, please re-open it.

Issue #10747, opened 2025-10-23
[Question]: OpenSearch implementation seems to be missing the keyword search component (using only Filter + k-NN)


woshiren0ck:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

https://github.com/infiniflow/ragflow/blob/494f84cd6983c613f010547526b5f9ec4f49b102/rag/utils/opensearch_conn.py#L245

The final request sent to OpenSearch only contains 'keyword filtering + k-NN', rather than '[k-NN on keyword-filtered results] + [keyword search]'. However, this is not the logic implemented when DOC_ENGINE=elasticsearch. Why is it done this way? Is this a new feature of OpenSearch?

yingfeng:
@pyyuhao Could you please answer this question?

pyyuhao:
> [@pyyuhao](https://github.com/pyyuhao) Could you please answer this question?

Thanks for your invitation @yingfeng and it's really a good question by @woshiren0ck .

hi, @woshiren0ck : 

First of all, OpenSearch for ragflow supports bybrid search which combines vector search and text search.  The weight of these two types of search is controlled by the parameter 'boost', and this is consistent between Elasticsearch  and OpenSearch. 

You can import some data into ragflow, using different weights to test. By using different weights, the recall in hybrid retrieval can vary significantly due to the differences between semantic retrieval and text retrieval. You'll you will find that these results are basicly consistent with those of ES. 

<img width="1652" height="1275" alt="Image" src="https://github.com/user-attachments/assets/4c1fa9be-1a95-4afd-b87c-f7f67ff9a72e" />


However, during the four years after release of ES 7.10.2 (the last OSS version) in 2021, OpenSearch has some incompatible differences in some api & methods, particularly in the field of vector search. For example, OpenSearch does not support "query + KNN" in DSL while KNN is only allowed to appear within the query.

[...]

PR #15760, opened 2026-06-08
fix(opensearch): keep the BM25 leg in hybrid search

mateidanut:

Fixes the OpenSearch side of #10747: hybrid search drops the keyword (BM25) leg and
ends up doing plain vector search.

When a search has both a text and a vector leg, `OSConnection.search()` throws the text
query away:

    del q["query"]
    q["query"] = {"knn": knn_query}

The text clause only stays on as a filter inside the knn query, so it narrows the
candidate set but doesn't count towards scoring. So hybrid search on OpenSearch behaves
like plain vector search, unlike the Elasticsearch backend.

What I changed:

- when both legs are present, send a real hybrid query
  `{"hybrid": {"queries": [bm25, {"knn": ...}]}}` and let a normalization-processor
  search pipeline score and combine the two legs
- only the actual filters (kb_id, available_int, ...) go in the knn filter, not the
  text must clause
- create the pipeline on startup if it's missing, so there's no separate provisioning
  step. name and weights can be set under `os:` in service_conf.yaml, or via
  `OS_HYBRID_PIPELINE`; defaults are `ragflow_hybrid_pipeline` and `[0.5, 0.5]`
- normalization-processor needs OpenSearch 2.10+. on older clusters, or when the
  pipeline can't be created, log a warning and fall back to vector-only instead of
  pointing at a pipeline that doesn't exist

This is only the hybrid-search fix; `create_doc_meta_idx` is already on main.

Testing (there's no OpenSearch path in CI): added a unit test
(`test/unit_test/rag/utils/test_opensearch_hybrid_search.py`, no services needed) that
checks the query built in each case — hybrid + pipeline param for text+vector, plain knn
for vector-only, plain bool for text-only, the knn filter never carrying the text
query_string, and the vector-only fallback when the pipeline isn't available. Also ran
it against a real OpenSearch 2.19.1 container with a doc that matches the keyword but
sits outside the knn top-k: pure knn returns `['D1','D2','D5']` (keyword doc missing),
the hybrid query returns `['A','D1','D2','D5']` (keyword doc present).

### Type of change

### DIFF
--- a/conf/service_conf.yaml
+++ b/conf/service_conf.yaml
@@ -27,6 +27,10 @@ os:
   hosts: 'http://localhost:1201'
   username: 'admin'
   password: 'infini_rag_flow_OS_01'
+  # Optional hybrid (BM25 + KNN) search tuning. The connector self-provisions the
+  # normalization search pipeline on start-up (requires OpenSearch >= 2.10).
+  # hybrid_search_pipeline: 'ragflow_hybrid_pipeline'
+  # hybrid_search_weights: [0.5, 0.5]   # [text/BM25 leg, vector/KNN leg]
 infinity:
   uri: 'localhost:23817'
   postgres_port: 5432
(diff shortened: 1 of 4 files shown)


yingfeng:
Hi, RAGFlow is under migration from Python to Golang for API services, and if you need the Opensearch as a DocEngine alternative, it's better to provide similar implementation as shown within #15583, thank you~

PR #15769, opened 2026-06-08
feat(go-api): migrate agent file download handler to Go with strict P…

Hz-186:
## What does this PR do?

This PR migrates the Agent Temporary File Download endpoint (`GET /api/v1/agents/download`) from the Python backend to the Go backend, optimizing the data retrieval flow and maintaining strict functional parity. It also fixes a persistent parsing error in the Sandbox code execution node.

**1. Go Handler Alignment & Optimization**
- **Issue:** The previous Go placeholder implementation was doing unnecessary DB queries looking for standard Knowledge Base documents (`entity.File`), which fundamentally mismatched how agent chat attachments were stored.
- **Fix:** Rewrote `DownloadAgentFile` in `internal/service/file.go` and `internal/handler/agent.go`. The service now directly accesses the underlying `storageImpl`, correctly targeting the `[tenantID]-downloads` bucket using the provided `id`, exactly as the Python implementation (`agent_app.py`) does.

**2. Sandbox Execution Node Patch**
- **Issue:** When running Code Nodes locally, Python execution frequently crashed with `NameError: name 'null' is not defined` due to unsafe string interpolation of JSON payloads directly into Python source code.
- **Fix:** Switched `agent/sandbox/result_protocol.py` to use base64-encoded payload transmission, letting the internal `json.loads()` safely deserialize JS data types (like `null`) into native Python types (`None`).

## Testing Done
- [x] Manually verified the `GET /api/v1/agents/download` endpoint correctly returns binaries (e.g. images, PDFs).
- [x] Ran Go unit tests covering handlers and services (`TestDownloadAgentFile`, `TestUploadAgentFileHandler`, `TestFileService_DownloadAgentFile_Success`).
- [x] Added fully-isolated Mock implementations for `storage.Storage` without polluting unrelated tests.

## Checklist
- [x] Code logic matches Python implementation
- [x] All local unit tests passed
- [x] No breaking changes to existing router interfaces


xugangqiang:
Please resolve conflicts

PR #15793, opened 2026-06-08
Fix: find_codec discards confident encoding detection and returns a garbage codec

devYRPauli:

`find_codec` in `rag/nlp/__init__.py` computes a `chardet` detection and checks its confidence, but only ever acts on it for the `ascii` special case. For any other confidently-detected encoding the result is discarded and the function falls through to the brute-force `all_codecs` loop, which returns the first codec that does not raise on the sample.

That loop is the problem: legacy single-byte codecs in the list (e.g. `cp037`, and `gbk`/`gb2312` for many byte sequences) decode arbitrary bytes without raising, so a non-UTF-8 upload is returned with the wrong codec and decodes to garbage. For example a Shift-JIS Japanese file resolves to `gbk` and a EUC-KR Korean file resolves to `gb2312`, both producing mojibake. `find_codec` is on the text-ingestion hot path (txt/html/json/csv parsers and `rag/app/naive.py`), so this affects every non-UTF-8, non-ASCII document.

The fix honors a confident detection when it actually decodes the sample, before falling back to the brute-force loop:

```python
encoding = detected['encoding']
if encoding and detected['confidence'] > 0.5:
    if encoding.lower() == "ascii":
        return "utf-8"
    try:
        blob[:1024].decode(encoding)
        return encoding
    except (UnicodeDecodeError, LookupError):
        pass
```

UTF-8/ASCII inputs are unaffected (UTF-8 is the first codec tried and succeeds). Added `test/unit_test/rag/test_find_codec.py`: the Shift-JIS and EUC-KR round-trip cases fail before the change and pass after; ASCII and UTF-8 control cases pass throughout.

### Type of change

### DIFF
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -153,8 +153,9 @@ def find_codec(blob):
         try:
             blob[:1024].decode(encoding)
             return encoding
-        except (UnicodeDecodeError, LookupError):
-            pass
+        except (UnicodeDecodeError, LookupError) as e:
+            logging.debug("find_codec: confident detection %r (%.2f) did not decode the sample, "
+                          "falling back to brute force: %s", encoding, detected['confidence'], e)
 
     for c in all_codecs:
         try:
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -142,9 +142,19 @@ all_codecs = [
 
 def find_codec(blob):
     detected = chardet.detect(blob[:1024])
-    if detected["confidence"] > 0.5:
-        if detected["encoding"] == "ascii":
+    encoding = detected['encoding']
+    if encoding and detected['confidence'] > 0.5:
+        if encoding.lower() == "ascii":
             return "utf-8"
+        # Honor a confident detection when it actually decodes the sample.
+        # Otherwise the brute-force loop below returns the first codec that
+        # does not raise, and legacy single-byte codecs (e.g. cp037) decode
+        # arbitrary bytes without error, yielding garbage text.
+        try:
+            blob[:1024].decode(encoding)
+            return encoding
+        except (UnicodeDecodeError, LookupError):
+            pass
 
     for c in all_codecs:
         try:
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -163,8 +163,9 @@ def find_codec(blob):
         try:
             blob[:1024].decode(encoding)
             return encoding
-        except (UnicodeDecodeError, LookupError):
-            pass
+        except (UnicodeDecodeError, LookupError) as e:
+            logging.debug("find_codec: confident detection %r (%.2f) did not decode the sample, "
+                          "falling back to brute force: %s", encoding, detected['confidence'], e)
 
     for c in all_codecs:
         try:
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 #
 
+import codecs
 import copy
 import logging
 import random
@@ -151,25 +152,34 @@ all_codecs = [
 
 
 def find_codec(blob):
-    detected = chardet.detect(blob[:1024])
-    encoding = detected['encoding']
-    if encoding and detected['confidence'] > 0.5:
-        if encoding.lower() == "ascii":
-            return "utf-8"
-        # Honor a confident detection when it actually decodes the sample.
-        # Otherwise the brute-force loop below returns the first codec that
-        # does not raise, and legacy single-byte codecs (e.g. cp037) decode
-        # arbitrary bytes without error, yielding garbage text.
+    sample = blob[:1024]
+
+    # A blob that decodes as UTF-8 is UTF-8; nothing else needs to be guessed.
+    # Check this first because chardet can report a confident single-byte guess
+    # for short UTF-8 text, and callers decode with errors="ignore", so a wrong
+    # codec is silently lossy instead of raising.
+    try:
+        codecs.getincrementaldecoder("utf-8")().decode(sample)
+        return "utf-8"
+    except UnicodeDecodeError:
+        pass
+
+    detected = chardet.detect(sample)
+    encoding = detected["encoding"]
+    if encoding:
+        # Honor the detection whenever it decodes the sample. The loop below
+        # returns the first codec that does not raise, and legacy single-byte
+        # codecs (cp037, utf_16) decode arbitrary bytes without error, so a
+        # low-confidence detection still beats the loop's first non-raising hit.
         try:
-            blob[:1024].decode(encoding)
+            sample.decode(encoding)
             return encoding
         except (UnicodeDecodeError, LookupError) as e:
-            logging.debug("find_codec: confident detection %r (%.2f) did not decode the sample, "
-                          "falling back to brute force: %s", encoding, detected['confidence'], e)
+            logging.debug("find_codec: detection %r (%.2f) did not decode the sample: %s", encoding, detected["confidence"] or 0.0, e)
 
     for c in all_codecs:
         try:
-            blob[:1024].decode(c)
+            sample.decode(c)
             return c
         except Exception:
             pass
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -14,7 +14,6 @@
 #  limitations under the License.
 #
 
-import codecs
 import copy
 import logging
 import random
@@ -157,12 +156,17 @@ def find_codec(blob):
     # A blob that decodes as UTF-8 is UTF-8; nothing else needs to be guessed.
     # Check this first because chardet can report a confident single-byte guess
     # for short UTF-8 text, and callers decode with errors="ignore", so a wrong
-    # codec is silently lossy instead of raising.
+    # codec is silently lossy instead of raising. The second decode covers a
+    # multi-byte character that the 1024-byte sample cuts in half.
     try:
-        codecs.getincrementaldecoder("utf-8")().decode(sample)
+        sample.decode("utf-8")
         return "utf-8"
     except UnicodeDecodeError:
-        pass
+        try:
+            blob.decode("utf-8")
+            return "utf-8"
+        except UnicodeDecodeError:
+            pass
 
     detected = chardet.detect(sample)
     encoding = detected["encoding"]
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -152,9 +152,19 @@ all_codecs = [
 
 def find_codec(blob):
     detected = chardet.detect(blob[:1024])
-    if detected["confidence"] > 0.5:
-        if detected["encoding"] == "ascii":
+    encoding = detected['encoding']
+    if encoding and detected['confidence'] > 0.5:
+        if encoding.lower() == "ascii":
             return "utf-8"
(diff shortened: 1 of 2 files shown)


devYRPauli:
Good catch. Moved the stubbing into a `monkeypatch`-based fixture so the fake modules are restored after each test instead of leaking into `sys.modules`. The fixture also drops any stubbed `chardet` before importing, so `find_codec` always binds the real detector even when another test module has stubbed it. Verified the file passes both on its own and interleaved with `test_is_english.py` in either order.

devYRPauli:
Good call. Added a `logging.debug` on the rejection path (encoding, confidence, and the decode error) before the brute-force fallback, so a mis-detection is diagnosable from ingestion logs. Tests still green (4 passed).

devYRPauli:
This is still current against main. The branch is mergeable and the four tests in `test/unit_test/rag/test_find_codec.py` pass.

I re-checked today that the bug is still live. I loaded the current `find_codec` and `all_codecs` straight from `rag/nlp/__init__.py` on main and gave it real encoded text:

| sample | chardet detects | confidence | `find_codec` returns | text preserved |
|---|---|---|---|---|
| Russian, cp1251 | windows-1251 | 0.99 | `cp037` | no |
| Japanese, shift_jis | SHIFT_JIS | 0.99 | `gbk` | no |
| Greek, cp1253 | ISO-8859-7 | 0.99 | `cp037` | no |

chardet is 99 percent confident in each case. The function drops that result, because only the `ascii` branch returns, and then falls into the `all_codecs` loop. That loop returns the first codec that does not raise. `cp037` is EBCDIC and decodes any byte sequence, so it never raises.

The Japanese row is the one worth noting. `gbk` does not produce obvious garbage. It produces plausible CJK text:

```
偙傟偼僄儞僐乕僨傿儞僌専弌偺偨傔偺僥僗僩偱偡丅
```

That string is indexed as if it were the document content, so the failure is silent.

The change honours a confident detection only when it also decodes the sample, and logs the rejection before the brute-force fallback.

Could someone take a look when there is time? I am happy to rebase or change the approach.

devYRPauli:
@JinHai-CN could you assign a reviewer to this one? It has had none since June.

I re-verified the bug against current main on 19 August. The PR is mergeable and the diff is 2 files.

xugangqiang:
## Review: `find_codec` discards confident encoding detection

I reviewed this against current `main` (the diff base matches `rag/nlp/__init__.py:153`). The bug is real — I reproduced it with chardet 7.4.3. Findings below.

### 1. Problem is real, but only partially fixed
The mojibake bug on the ingestion hot path (`rag/app/naive.py:896`, `deepdoc/parser/*`) is confirmed. However the `confidence > 0.5` gate leaves real uploads unfixed:

| sample | chardet | old | new |
|---|---|---|---|
| Japanese shift_jis | cp932@0.73 | gbk (garbage) | cp932 ✅ |
| Korean euc_kr | CP949@0.58 | gb2312 (garbage) | CP949 ✅ |
| Russian cp1251 | Windows-1251@**0.43** | cp037 (garbage) | cp037 (garbage) ❌ |
| Greek cp1253 | ISO-8859-7@**0.34** | cp037 (garbage) | cp037 (garbage) ❌ |

So Russian/Greek are still silently broken: their confidence is below the threshold, they fall through to the brute-force `all_codecs` loop, and `cp037` (which decodes any bytes without error) is returned. The PR only fixes the conf>0.5 subset of the stated problem.

### 2. Suggestions for a better fix
- **Put a strict UTF-8 check before honoring chardet.** The old code relied on `all_codecs[0] == "utf-8"` to prefer UTF-8; this PR elevates chardet above that. If chardet ever mis-detects a UTF-8 file as a single-byte codec with conf>0.5, the new code returns the wrong codec, and the caller at `naive.py:897` decodes with `errors="ignore"` — silently dropping bytes. A defensive `try: blob.decode("utf-8"); return "utf-8"` first removes this latent regression.
- **Reconsider the `>0.5` gate / the fallback itself.** Since the fallback (`cp037`) is always wrong, honoring a detection that *decodes* the sample is strictly better than falling back, even at lower confidence. Or fix the fallback so EBCDIC/single-byte codecs that never raise are not blindly accepted.
- **Canonicalize the returned codec.** chardet returns `CP949` for `euc_kr` bytes and `cp932` for `shift_jis`. I verified `blob.decode("CP949") == original` holds for the test string, but across the wider Hangul block (U+AC00–U+AEFF) `CP949` and `euc_kr` decode *differently* — so real euc_kr files can get wrong characters (not a crash, not mojibake, but incorrect). Consider normalizing or validating the decode on a larger sample.
- **Longer term:** `chardet` is minimally maintained; `charset-normalizer` (used by requests) does decode-verified, confidence-ranked detection natively and could replace both code paths. Not blocking this PR.

### 3. Test coverage / regression risk
Added tests (shift_jis + euc_kr roundtrip, ascii→utf-8, utf-8) are good regression tests. Gaps:
- No test for the unfixed Russian/Greek (conf<0.5) case — the partial fix is hidden by the tests.
- No guard test that a genuine UTF-8 file still returns `utf-8` when chardet would report a different encoding (the new regression path).
- No coverage of the `LookupError` branch (chardet returns an un-importable codec name) or the confidence boundary (exactly 0.5 / below).
- The `monkeypatch` fixture is correct and avoids `sys.modules` leakage.

Practical regression risk is low (real UTF-8 files detect as utf-8@0.99 in my tests), but the chardet-before-utf-8 reorder is a behavior change worth hardening.

devYRPauli:
You are right on points 1 and 2. I pushed a revision. Point 3 does not reproduce for me, and I give the measurement below.

First, a version note that changes how the numbers read. Your table matches chardet 7.4.3. `pyproject.toml` pins `chardet>=5.2.0,<6.0.0`, and `uv.lock` resolves 5.2.0. The two versions score single-byte encodings very differently, so I measured both.

## 1. Strict UTF-8 check first

Confirmed, and it is worse than a latent risk on the pinned version. It is a regression my first patch introduced.

Take an ASCII sentence that ends with two emoji. chardet 5.2.0 reports `Windows-1254` at 0.56 confidence. That clears the old 0.5 gate, the bytes decode as Windows-1254 without error, so my patch returned it. The text came back as mojibake. The previous code returned `utf-8`, because `all_codecs[0]` is `utf-8` and the loop tried it first.

`rag/app/naive.py:897` decodes with `errors="ignore"`, so this stays silent in production. There are 5 call sites: `excel_parser.py:315`, `html_parser.py:40`, `utils.py:27`, `json_parser.py:34`, and `naive.py:896`.

The revision runs a strict UTF-8 check before chardet. It uses an incremental decoder, so a multi-byte character cut at the 1024-byte sample boundary does not cause a false negative. chardet 7.4.3 detects that same blob as `utf-8` at 0.99, which is why the case is invisible on your version.

## 2. The confidence gate

Removed. Your reasoning holds: the loop returns the first codec that does not raise, and `cp037` and `utf_16` decode arbitrary bytes, so the fallback is always wrong. A low-confidence detection beats that.

Results over 7 legacy-encoded samples and 11 UTF-8 samples. "ok" means the returned codec reproduces the original text.

chardet 5.2.0 (pinned):

| sample | before | first patch | revision |
|---|---|---|---|
| Russian cp1251 | garbage | garbage | garbage |
| Greek cp1253 | garbage | ok | ok |
| Japanese shift_jis | garbage | ok | ok |
| Korean euc_kr | garbage | ok | ok |
| Hebrew cp1255 | garbage | ok | ok |
| UTF-8 with emoji | ok | **garbage** | ok |
| total correct | 13/18 | 16/18 | 17/18 |

chardet 7.4.3:

| sample | before | first patch | revision |
|---|---|---|---|
| Russian cp1251 | garbage | garbage | ok |
| Greek cp1253 | garbage | garbage | ok |
| Hebrew cp1255 | garbage | garbage | ok |
| total correct | 13/18 | 15/18 | 18/18 |

My confidence figures are close to yours but not equal (Russian 0.48, Greek 0.50 against your 0.43 and 0.34). The sample text differs. The conclusion is the same.

## 3. Codec canonicalization

I cannot reproduce a difference between `CP949` and `euc_kr` in Python. I compared them over every modern Hangul syllable, U+AC00 to U+D7A3, in both directions:

- 11172 syllables total
- `cp949` encodes 11172
- `euc_kr` encodes 11172
- present in `cp949` but not `euc_kr`: 0
- bytes written by one codec always decode identically under the other

So returning `CP949` looks safe, and rewriting it to `euc_kr` would be the riskier change, because `cp949` is the superset name. I did not add canonicalization. If you have a byte sequence where the two disagree, please post it and I will add the mapping.

## 4. Tests

Added, and each new test is checked by reverting the source fix and confirming the test fails. 9 tests pass with the fix. 4 fail without it:

- UTF-8 with emoji round-trips
- low-confidence detection is used, cp1251
- low-confidence detection is used, cp1253
- the UTF-8 check beats a confident wrong detection

The two low-confidence tests patch `chardet.detect` to return 0.3, because the real samples detect at 0.97 or higher on the pinned version. The `LookupError` test now asserts that the returned name is a real codec through `codecs.lookup`, instead of only checking the type.

I dropped your 0.5 boundary case, since the gate no longer exists.

## Known limit

Russian cp1251 is still wrong on chardet 5.2.0. The detector reports `MacCyrillic` at 0.99, and those bytes decode under `MacCyrillic` without error. No confidence gate and no decode check can catch a confident wrong answer from the detector. chardet 7.4.3 gets this sample right. Raising the pin is a separate change, so I left it out. Tell me if you want it in this PR.

## CI

`ruff format --check` now reports both files as already formatted, which clears the `ragflow_preflight` failure.

devYRPauli:
Three CodeRabbit findings. One is real and is now fixed. Two do not apply.

**1. Incremental decoder accepts an incomplete suffix. Valid, and fixed in 3ca2f0e.**

The report is correct. An incremental decoder buffers a trailing partial sequence instead of rejecting it. So `b"a" * 1023 + b"\xe9" + b" "` passed the sample check, `find_codec` returned `utf-8`, and `rag/app/naive.py:897` decoded with `errors="ignore"` and dropped the bad bytes.

The check now decodes the sample, and falls back to decoding the whole blob:

```python
try:
    sample.decode("utf-8")
    return "utf-8"
except UnicodeDecodeError:
    try:
        blob.decode("utf-8")
        return "utf-8"
    except UnicodeDecodeError:
        pass
```

This is what the brute-force loop already did on its first pass, moved ahead of chardet. It removes the `codecs` import.

Measured on that blob: `find_codec` now returns `ISO-8859-1`, which decodes the whole blob. A real UTF-8 blob with a multi-byte character cut by the 1024-byte boundary still returns `utf-8` and round-trips. The 18-sample encoding matrix scores the same as before.

Added `test_find_codec_rejects_utf8_when_only_the_sample_looks_valid`. Reverting the source makes it fail.

**2. Global `sys.modules` stubs can leak. Does not apply.**

The stubs are already inside the `find_codec` fixture and already use `monkeypatch.setitem`, which restores `sys.modules` after each test. `_stub` is defined at line 33, inside the fixture body, not at module level. No change made.

**3. Suppress the Unicode-confusable warning. Not needed.**

`RUF001` is not enabled here. `[tool.ruff.lint]` in `pyproject.toml` sets `extend-select = ["ASYNC", "ASYNC1"]`. `ruff check` on both files passes with the repo configuration, so a `noqa` would suppress a rule that never runs. No change made.

xugangqiang:
LGTM @yuzhichang

PR #6513, opened 2025-03-25
Optimize graphrag again

yuzhichang:
### What problem does this PR solve?

Removed set_entity and set_relation to avoid accessing doc engine during graph computation.
Introduced GraphChange to avoid writing unchanged chunks.

### Type of change

- [x] Performance Improvement

PR #15797, opened 2026-06-08
Fix: restore GraphRAG entity ranking by indexing pagerank and n-hop paths

### LINKED ISSUE #15795: GraphRAG entity ranking is permanently broken: pagerank (rank_flt) and n_hop_with_weight are never indexed
### Describe the bug

Every knowledge-graph query ranks entities incorrectly. `KGSearch` orders entities by `pagerank * sim`, but `pagerank` is always `0`, so the ranking collapses to an effectively arbitrary order and every displayed entity score is `0.00`.

The root cause is that the entity chunks written at index time no longer carry the fields the search path reads back:

- `rag/graphrag/utils.py` `graph_node_to_chunk` builds the entity chunk from `entity_type` / `description` / `source_id` only. The node `pagerank` (computed in `merge_subgraph` and `entity_resolution`) and the n-hop neighbour paths are dropped.
- `rag/graphrag/search.py` still reads them back:

```python
"pagerank": get_float(ent.get("rank_flt", 0)),               # always 0
"n_hop_ents": json.loads(ent.get("n_hop_with_weight","[]")), # always []
...
ents_from_query = sorted(..., key=lambda x: x[1]["sim"] * x[1]["pagerank"], ...)  # x 0 for every entity
```

### Impact

- Entity selection for every KG query collapses to arbitrary order because the sort key is always `0`.
- Every displayed entity score is `0.00`.
- The entire n-hop relation-enrichment block in `search.py` is dead code, since `n_hop_ents` is always empty. The `merge_tuples` and `is_continuous_subsequence` helpers are left orphaned.
- The documented `P(E|Q) = pagerank * sim` ranking has no effect.

### Root cause

The `update_nodes_pagerank_nhop_neighbour` function that used to write `rank_flt` and `n_hop_with_weight` was removed in PR #6513, but the read side in `KGSearch` was never updated to match. The producer of these fields disappeared while the consumer stayed.

### How to reproduce

1. Build a knowledge graph for any dataset (default GraphRAG settings).
2. Run any knowledge-graph query.
3. Observe that the returned entities all show `Score 0.00` and that ordering does not reflect pagerank.

### Expected behavior

Entity chunks should carry `rank_flt` (pagerank) and `n_hop_with_weight`, so that `KGSearch` ranks entities by `pagerank * sim` and the n-hop relation enrichment works as designed.

### Environment

Affects all supported document engines (Elasticsearch, OpenSearch, Infinity, OceanBase).

cleanjunc:
### Summary

Closes #15795 

Knowledge-graph queries rank entities by `pagerank * sim` in `KGSearch`, but the entity chunks written at index time stopped carrying the values that ranking depends on. `graph_node_to_chunk` only stored `entity_type`, `description`, and `source_id`, dropping the node `pagerank` and the n-hop neighbour paths, while `search.py` still read them back as `rank_flt` and `n_hop_with_weight`.

The producer of these fields, `update_nodes_pagerank_nhop_neighbour`, was removed in #6513, but the read side in `KGSearch` was never updated. The result is that on every knowledge-graph query:

- `pagerank` resolves to `0`, so the `pagerank * sim` sort key is `0` for every entity and selection falls back to arbitrary order.
- Every displayed entity score is `0.00`.
- The n-hop relation-enrichment block is dead code because `n_hop_ents` is always empty, leaving `merge_tuples` and `is_continuous_subsequence` orphaned.

This PR restores the missing index-time fields so the documented `P(E|Q) = pagerank * sim` ranking and the n-hop enrichment work again.

What changed:

- `graph_node_to_chunk` now writes `rank_flt` from the node pagerank and `n_hop_with_weight` from the recomputed n-hop neighbour paths.
- Reintroduced the n-hop path computation (`n_neighbor`) in `rag/graphrag/utils.py`, reusing the previously orphaned `merge_tuples` / `is_continuous_subsequence` helpers, with a direction-agnostic edge-weight lookup for undirected graphs. `set_graph` computes the paths per added or updated node and passes them through.
- `KGSearch` now selects `n_hop_with_weight` in the entity keyword search so Infinity and OceanBase return it (Elasticsearch and OpenSearch already read it from `_source`), and the read is hardened against missing keys or empty strings before `json.loads`.
- Added the `n_hop_with_weight` column to OceanBase, including the `EXTRA_COLUMNS` migration entry so existing tables get it. The other engines already map both fields via dynamic templates or the Infinity mapping.

Scope note: pagerank and n-hop are re-indexed for the added or updated nodes in each pass, consistent with the existing incremental indexing design.

### Type of change

### Testing

Added unit tests in `test/unit_test/rag/graphrag/test_graphrag_utils.py`:

- `n_neighbor`: path and weight shape, one-hop vs two-hop, isolated nodes, missing weights, and direction-agnostic lookup.
- `graph_node_to_chunk`: `rank_flt` populated from pagerank and defaulting to `0`, `n_hop_with_weight` serialized and defaulting to an empty list.

```
uv run pytest test/unit_test/rag/graphrag/   # 106 passed
uv run ruff check rag/graphrag/ rag/utils/ob_conn.py
```

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -78,10 +78,17 @@ class KGSearch(Dealer):
                 continue
             if isinstance(ent["entity_kwd"], list):
                 ent["entity_kwd"] = ent["entity_kwd"][0]
+            # n_hop_with_weight may be absent (older chunks) or an empty string
+            # (the Infinity column default), neither of which json.loads handles.
+            n_hop_raw = ent.get("n_hop_with_weight") or "[]"
+            try:
+                n_hop_ents = json.loads(n_hop_raw)
+            except (json.JSONDecodeError, TypeError):
+                n_hop_ents = []
             res[ent["entity_kwd"]] = {
                 "sim": get_float(ent.get("_score", 0)),
                 "pagerank": get_float(ent.get("rank_flt", 0)),
-                "n_hop_ents": json.loads(ent.get("n_hop_with_weight", "[]")),
+                "n_hop_ents": n_hop_ents,
                 "description": ent.get("content_with_weight", "{}")
             }
         return res
@@ -111,7 +118,7 @@ class KGSearch(Dealer):
         filters = deepcopy(filters)
         filters["knowledge_graph_kwd"] = "entity"
         matchDense = self.get_vector(", ".join(keywords), emb_mdl, 1024, sim_thr)
-        es_res = self.dataStore.search(["content_with_weight", "entity_kwd", "rank_flt"], [], filters, [matchDense],
+        es_res = self.dataStore.search(["content_with_weight", "entity_kwd", "rank_flt", "n_hop_with_weight"], [], filters, [matchDense],
                                        OrderByExpr(), 0, N,
                                        idxnms, kb_ids)
         return self._ent_info_from_(es_res, sim_thr)
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -84,6 +84,7 @@ class KGSearch(Dealer):
             try:
                 n_hop_ents = json.loads(n_hop_raw)
             except (json.JSONDecodeError, TypeError):
+                logging.warning(f"Failed to parse n_hop_with_weight for entity {ent.get('entity_kwd')}: {n_hop_raw}")
                 n_hop_ents = []
             res[ent["entity_kwd"]] = {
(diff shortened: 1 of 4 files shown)


cleanjunc:
@qinling0210 Could you please review this PR? Thanks a lot!

cleanjunc:
@yingfeng @qinling0210 

qinling0210 (approved):
LGTM

PR #15802, opened 2026-06-08
fix: naive_merge splits oversized sections and counts overlap tokens correctly

### LINKED ISSUE #15801: [Bug]: naive_merge stops splitting oversized sections and miscounts overlap tokens
### Actual behavior

`naive_merge` and `naive_merge_with_images` in `rag/nlp/__init__.py` no longer split sections that are larger than `chunk_token_num`. When no custom (backtick wrapped) delimiter is configured, the default path simply appends each whole section:

```python
for sec, pos in sections:
    add_chunk("\n" + sec, pos)   # add_chunk only appends, it never splits
```

`add_chunk` only merges and appends, so any section bigger than `chunk_token_num` (a large paragraph, an EPUB chapter, merged PDF runs, single strategy output) becomes one oversized chunk. The default sentence delimiters (`\n!?。；！？`) are only honored when they are wrapped in backticks, which is not the default.

There is also an overlap accounting problem. In `add_chunk`, `tnum` is computed from the text before the overlap prefix is prepended, but the value stored in `tk_nums` is that stale count. With overlap enabled, every chunk ends up larger than `chunk_token_num`.

This is a regression from commit `db0f6840d` (#11434, "ignore chunk size when using custom delimiters"). The rewritten flow component path recomputes the count correctly at `rag/flow/chunker/token_chunker.py:231`, which confirms the legacy functions are the ones that are wrong.

### Reproduction

```python
from rag.nlp import naive_merge, num_tokens_from_string

sentence = "Retrieval augmented generation improves answers with external context\n"
section = sentence * 60  # one section, about 660 tokens

chunks = [c for c in naive_merge([section], chunk_token_num=128, delimiter="\n!?。；！？") if c.strip()]
print(len(chunks), [num_tokens_from_string(c) for c in chunks])
# Observed: 1 [661]
# Expected: several chunks, each around 128 tokens
```

With overlap, chunk sizes drift above the limit:

```python
sentences = [" ".join(["w"] * 10) for _ in range(30)]
cks = [c for c in naive_merge(sentences, chunk_token_num=50, delimiter="\n!?。；！？", overlapped_percent=20) if c.strip()]
# Observed sizes: [50, 60, 62, 63, 63, 63]
# Expected sizes: [50, 50, 50, 50, 50, 50, ...]
```

### Expected behavior

Sections larger than `chunk_token_num` should be split at sentence delimiters and merged back up to the configured size, and chunks should stay within `chunk_token_num` when overlap is enabled. The custom delimiter behavior from #11434 (one chunk per delimited segment, ignoring chunk size) should stay intact.

### Environment

- RAGFlow commit: `55abf4f56`
- Affects default chunking for the naive parser across text, PDF, and image paths.

cleanjunc:
### Summary

Closes #15801 

This PR restores delimiter aware splitting in `naive_merge` and `naive_merge_with_images` (`rag/nlp/__init__.py`) and fixes the overlap token accounting in both functions.

After commit `db0f6840d` (#11434), the default (non custom delimiter) path stopped splitting sections at sentence boundaries and just appended each whole section. As a result, any section larger than `chunk_token_num` (a large paragraph, an EPUB chapter, merged PDF runs) turned into a single oversized chunk. A 660 token section produced one 660 token chunk instead of several chunks near the configured size.

There was also an overlap accounting bug: `tk_nums` stored the token count taken before the overlap prefix was prepended, so chunks systematically grew past `chunk_token_num` when overlap was enabled. The rewritten flow component recomputes this correctly at `rag/flow/chunker/token_chunker.py:231`, which is the reference for the fix here.

What changed:

* The default path now splits each oversized section at the sentence delimiters returned by `get_delimiters`, then lets `add_chunk` merge the pieces back up to `chunk_token_num`. An empty delimiter falls back to size only merging, matching how `token_chunker` calls into `naive_merge`.
* `add_chunk` recomputes the token count after prepending the overlap prefix in both functions, so chunks respect the size limit.
* The custom delimiter behavior added in #11434 is preserved: when a backtick wrapped delimiter is present, each delimited segment becomes its own chunk and chunk size is ignored.
* While restoring the image path split, `concat_img` gained a same reference guard so a section image shared across several split sentences is not stacked onto itself. The PIL branch already deduped identical images; this brings the LazyImage branch in line, since it would otherwise concatenate the same blob list repeatedly.

Tests: a new `test/unit_test/rag/test_naive_merge.py` covers the text, PDF, and image paths, the overlap budget, the custom delimiter behavior, and the image dedup. Six of the cases fail on the previous code and pass after this change. The surrounding chunk related unit tests (chunk service, builder, post processor, app) stay green.

Reproduction before the fix:

```python
from rag.nlp import naive_merge, num_tokens_from_string

section = "Retrieval augmented generation improves answers with external context\n" * 60
chunks = [c for c in naive_merge([section], chunk_token_num=128, delimiter="\n!?。；！？") if c.strip()]
print(len(chunks), [num_tokens_from_string(c) for c in chunks])
# Before: 1 [661]
# After:  5 [132, 132, 132, 132, 132]
```

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -1090,6 +1090,8 @@ def naive_merge(sections: str | list, chunk_token_num=128, delimiter="\n。；�
             if cks:
                 overlapped = RAGFlowPdfParser.remove_tag(cks[-1])
                 t = overlapped[int(len(overlapped) * (100 - overlapped_percent) / 100.):] + t
+                # Recount with the overlap prefix included, else chunks overshoot chunk_token_num.
+                tnum = num_tokens_from_string(t)
             if t.find(pos) < 0:
                 t += pos
             cks.append(t)
@@ -1103,6 +1105,7 @@ def naive_merge(sections: str | list, chunk_token_num=128, delimiter="\n。；�
     custom_delimiters = [m.group(1) for m in re.finditer(r"`([^`]+)`", delimiter)]
     has_custom = bool(custom_delimiters)
     if has_custom:
+        # Custom delimiters ignore chunk_token_num: each segment is its own chunk.
         custom_pattern = "|".join(re.escape(t) for t in sorted(set(custom_delimiters), key=len, reverse=True))
         cks, tk_nums = [], []
         for sec, pos in sections:
@@ -1120,8 +1123,16 @@ def naive_merge(sections: str | list, chunk_token_num=128, delimiter="\n。；�
                 tk_nums.append(num_tokens_from_string(text))
         return cks
 
+    # Split oversized sections at sentence delimiters; add_chunk re-merges to size.
+    dels = get_delimiters(delimiter)
     for sec, pos in sections:
-        add_chunk("\n" + sec, pos)
+        if not dels or num_tokens_from_string(sec) < chunk_token_num:
+            add_chunk("\n" + sec, pos)
+            continue
+        for sub_sec in re.split(r"(%s)" % dels, sec, flags=re.DOTALL):
+            if not sub_sec or re.fullmatch(dels, sub_sec):
+                continue
+            add_chunk("\n" + sub_sec, pos)
 
     return cks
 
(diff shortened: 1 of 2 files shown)


cleanjunc:
@qinling0210 @yingfeng Could you please review this PR? Thanks a lot!

cleanjunc:
@yingfeng @qinling0210 

PR #15806, opened 2026-06-08
fix: reclaim crashed-worker tasks and harden Redis pending-message replay

### LINKED ISSUE #15805: [Bug] Crashed-worker tasks are never reclaimed, and pending-message replay aborts on any non-exact Redis error
### Describe the problem

Task executors consume documents from Redis streams using a consumer group. Two issues in `rag/utils/redis_conn.py` prevent in-flight work from recovering when a worker dies.

**1. Pending-message replay aborts on the first unexpected error**

On startup, `get_unacked_iterator` replays each queue's pending (delivered but unacked) messages so a restarted worker can finish what it had in flight. When inspecting a queue, it only treats the error as "stream missing" if the message matches the string `'no such key'` exactly:

```python
except Exception as e:
    if str(e) == 'no such key':
        logging.warning(...)
        continue
# falls through here on any other error, with group_info never assigned
if not any(gi["name"] == group_name for gi in group_info):
    ...
```

Any other error (a connection blip, or a differently worded missing-stream message) skips the `continue`, reaches the `group_info` check while `group_info` was never assigned, and raises `UnboundLocalError`. That is caught by the outer `except`, which ends the whole replay and silently skips every remaining queue. The sibling `queue_consumer` already handles this correctly with `"no such key" in str(e).lower()`, so the behavior is also inconsistent.

**2. A crashed worker's in-flight messages are orphaned forever**

There is no `XCLAIM` or `XAUTOCLAIM` anywhere in the codebase, and the helper methods `get_pending_msg` and `requeue_msg` have no callers. A worker can only recover its own pending entries, and only if it restarts under the identical `CONSUMER_NAME` (`task_executor_<type>_<index>`). When worker identity shifts (scale-down, rescheduling under Docker or Kubernetes), the dead worker's pending messages stay in the group's Pending Entries List (PEL) forever: the affected documents never finish and the PEL grows without bound.

While reviewing the fix, `requeue_msg` also turned out to have a latent bug: its retry loop never stops on success, so it re-adds the same message up to three times.

### How to reproduce

- For (1): cause `xinfo_groups` to raise anything other than the exact `no such key` string during executor startup (for example a brief Redis connection error). Pending-message replay stops and remaining queues are not drained.
- For (2): start several workers, send documents, kill a worker mid-task, then start fewer workers (or under a different consumer identity). The killed worker's in-flight messages remain pending and the documents stay stuck.

### Expected behavior

- Pending-message replay isolates each queue so one queue's error cannot skip the rest.
- In-flight messages owned by a worker that is no longer alive are reclaimed and re-delivered to a live worker.

cleanjunc:
### Summary

Closes #15805 

Task executors consume documents from Redis streams via a consumer group. Two issues in `rag/utils/redis_conn.py` keep in-flight work from recovering when a worker dies.

**1. Pending-message replay aborts on any non-exact error**

`get_unacked_iterator` replays each queue's pending entries on startup, but it only recognized a missing stream when the error string equaled `'no such key'` exactly. Any other error (a connection blip, or a differently worded message) fell through to a `group_info` check where `group_info` was never assigned, raising `UnboundLocalError`. The outer `except` swallowed it and ended the whole replay, silently skipping every remaining queue. `queue_consumer` already does the right thing with `"no such key" in str(e).lower()`, so the behavior was also inconsistent.

Each queue is now isolated in its own try block, `group_info` is always bound before use, and missing streams are matched case-insensitively.

**2. A crashed worker's in-flight messages are orphaned forever**

There was no `XCLAIM`/`XAUTOCLAIM` in the codebase, and `get_pending_msg`/`requeue_msg` had no callers. A worker could recover only its own pending entries, and only when restarted under the identical `CONSUMER_NAME`. On scale-down or rescheduling, a dead worker's messages stayed in the group's Pending Entries List forever, leaving documents stuck and the PEL growing without bound.

This completes that existing scaffolding. `report_status` already tracks worker liveness via heartbeats; it now also calls a new `reclaim_pending_msg`, under the existing `clean_task_executor` distributed lock, to requeue any pending message whose owning consumer is no longer alive. A `min_idle_ms` guard (set to the heartbeat timeout) prevents stealing recently delivered work, and live workers are skipped so an active task is never reclaimed.

Additional hardening found along the way:

- `requeue_msg` now stops on success. Its retry loop previously fell through and re-added the same message up to three times.
- Its ack is unconditional, so an entry whose payload was already trimmed still leaves the PEL.
- Workers re-assert `TASKEXE` membership on every heartbeat, so a transient eviction self-heals and an alive worker is never mistaken for a dead one.

### Safety

The primary guard is consumer liveness: a busy worker keeps heartbeating because heavy work runs through `asyncio.to_thread` and async limiters, so its messages are never reclaimed regardless of task length. The idle guard covers the startup window and any momentary gap. The worst remaining case is a single re-delivery, which the existing at-least-once task handling already tolerates, and which is strictly better than the previous orphaned-forever behavior.

### Tests

Added `test/unit_test/rag/utils/test_redis_conn_reclaim.py` covering replay isolation across queues, missing stream and group handling, single requeue, trimmed-entry ack, PEL paging, and dead-versus-live reclaim selection.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### DIFF
--- a/rag/svr/task_executor.py
+++ b/rag/svr/task_executor.py
@@ -1797,6 +1797,9 @@ async def report_status():
 
         # Report heartbeat to Redis
         try:
+            # Re-assert membership each cycle so a transient eviction from TASKEXE
+            # self-heals and our in-flight messages aren't reclaimed as orphans.
+            REDIS_CONN.sadd("TASKEXE", CONSUMER_NAME)
             REDIS_CONN.zadd(CONSUMER_NAME, heartbeat, now_ts)
         except Exception as e:
             logging.warning(f"Failed to report heartbeat: {e}")
@@ -1819,6 +1822,7 @@ async def report_status():
         if lock_acquired:
             try:
                 task_executors = REDIS_CONN.smembers("TASKEXE") or set()
+                live_workers = {CONSUMER_NAME}
                 for worker_name in task_executors:
                     if worker_name == CONSUMER_NAME:
                         continue
@@ -1832,6 +1836,23 @@ async def report_status():
                         logging.info(f"{worker_name} expired, removed")
                         REDIS_CONN.srem("TASKEXE", worker_name)
                         REDIS_CONN.delete(worker_name)
+                    else:
+                        live_workers.add(worker_name)
+
+                # Reclaim in-flight messages orphaned by dead consumers, which would
+                # otherwise sit in the PEL forever (e.g. on scale-down/rescheduling).
+                # Live workers are skipped and min_idle_ms avoids stealing fresh work.
+                try:
+                    reclaimed = REDIS_CONN.reclaim_pending_msg(
+                        settings.get_svr_queue_names(TASK_TYPE),
+                        SVR_CONSUMER_GROUP_NAME,
+                        live_workers,
+                        min_idle_ms=WORKER_HEARTBEAT_TIMEOUT * 1000,
+                    )
+                    if reclaimed:
+                        logging.info(f"Reclaimed {reclaimed} orphaned pending message(s) from dead consumers")
+                except Exception as e:
+                    logging.warning(f"Failed to reclaim orphaned pending messages: {e}")
             except Exception as e:
                 logging.warning(f"Failed to clean other executors: {e}")
             finally:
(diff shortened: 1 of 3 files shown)


cleanjunc:
@yingfeng @qinling0210 

cleanjunc:
@yingfeng @qinling0210 

PR #15809, opened 2026-06-08
fix(sync): tolerate list inputs for Discord server_ids / channels (#15790)

### LINKED ISSUE #15790: [Bug]: Discord Connector sync fails with "'list' object has no attribute 'split'" when syncing Discord messages
### Self Checks

N/A (RAGFlow Cloud)

### RAGFlow image version

RAGFlow Cloud

```Markdown

```

### Actual behavior

The Discord data source was created successfully and linked to a dataset.

When the sync task is started, it fails immediately with the following error:

'list' object has no attribute 'split'

The issue is reproducible and occurs every time synchronization is triggered.

Environment: RAGFlow Cloud

Image

### Expected behavior

The Discord connector should successfully retrieve messages from the configured Discord server/channel and import them into the linked dataset.

### Steps to reproduce

```Markdown
Create a Discord bot in Discord Developer Portal.
Enable Message Content Intent.
Invite the bot to a Discord server.
Create a Discord data source in RAGFlow Cloud.
Configure:
Discord Bot Token
Server ID
Channel ID
Save the data source configuration.
Create a dataset.
Link the Discord data source to the dataset.
Start synchronization.
The sync task fails immediately with the following error:

'list' object has no attribute 'split'

Additional notes:

The Discord bot was successfully authorized and added to the server.
Message Content Intent is enabled.
The issue occurs consistently on every sync attempt.
Tested with both Server ID only and Server ID + Channel ID configurations.
```

### Additional information

_No response_

Rene0422:
## Summary

Fixes #15790.

Every Discord sync launched from the current Web UI crashes immediately with:

```
'list' object has no attribute 'split'
```

The error is raised in [rag/svr/sync_data_source.py:650-651](rag/svr/sync_data_source.py#L650-L651):

```python
server_ids=server_ids.split(",") if server_ids else [],
channel_names=channel_names.split(",") if channel_names else [],
```

### Root cause

Three independent bugs stack here, all in the Discord branch of `sync_data_source.py`:

1. **Type mismatch (the user's exact error).** The current form at [web/src/pages/user-setting/data-source/constant/index.tsx:833-843](web/src/pages/user-setting/data-source/constant/index.tsx#L833-L843) uses `FormFieldType.Tag` for both **Server IDs** and **Channels**:

    ```tsx
    { label: 'Server IDs', name: 'config.server_ids', type: FormFieldType.Tag, required: false },
    { label: 'Channels',   name: 'config.channels',   type: FormFieldType.Tag, required: false },
    ```

    Tag inputs serialise to **lists**, not comma-separated strings. The backend `.split(",")` then explodes on the very first sync.

2. **Field-name mismatch.** The form writes `config.channels`. The backend reads `self.conf.get("channel_names", None)`. Even if `.split(",")` were fixed, channels would silently be empty for every UI-created source.

3. **Int conversion missing.** [common/data_source/discord_connector.py:82](common/data_source/discord_connector.py#L82) types `server_ids` as `list[int]` (Discord guild IDs are integers); the previous `.split(",")` produced strings, so the `channel.guild.id not in server_ids` filter at [discord_connector.py:92](common/data_source/discord_connector.py#L92) silently never matched.

So even the configurations that didn't crash were also broken — there is no path through the current code that actually filtered by server id from a UI-created source.

### Fix

A 39-line patch in one function:

- New `Discord._coerce_str_list` static method: accepts `None` / `""` / `list` / `tuple` / `set` / scalar / comma-separated str, returns a clean `list[str]` with whitespace trimmed and empty entries dropped. Smoke-tested against the 10 input shapes that can hit it (see Test plan).
- `_generate` reads `config.channels` first (the form's actual key) and falls back to `config.channel_names`, so SDK callers and legacy configs that already shipped with the old key keep working.
- `server_ids` is coerced to `list[int]`. Non-integer entries are logged and dropped instead of crashing the sync, so a single malformed tag from the form doesn't tank the rest of the run.

### What this PR does NOT change

- **Web form key (`config.channels`)** — kept as-is. Renaming it to `channel_names` would force a UI migration and break in-flight configs; the backend fallback solves the same problem more safely.
- **`common/data_source/discord_connector.py`** — its signature was already correct.
- **Other connectors (Slack, Gmail, Confluence, etc.)** — they don't crash today and were not in the issue's scope.

## Test plan

`Discord._coerce_str_list` has been exercised against all ten realistic input shapes — list, tuple, set, comma-separated string, str with extra whitespace, empty entries, integers from a Tag input, None, empty list, single trailing comma. All pass.

End-to-end smoke checks the reviewer should run:

- [ ] Create a Discord data source via the Web UI with server IDs filled in (the tag-input case). Trigger sync — the `'list' object has no attribute 'split'` error should be gone, replaced by a successful Discord API call.
- [ ] Same, but enter channel names too — verify the connector filters to those channels rather than scanning every accessible channel.
- [ ] Create a Discord source via the SDK / direct DB write with the legacy keys (`server_ids="123,456"`, `channel_names="general,random"`) — sync still works (backward-compat path).
- [ ] Enter a non-numeric Server ID like `not-a-guild-id` in the UI — sync continues, that entry is logged at WARNING and dropped, valid entries still filter correctly.
- [ ] Leave both fields blank — sync indexes everything the bot has read access to (same as today's intended behaviour).

## Files changed

- [rag/svr/sync_data_source.py](rag/svr/sync_data_source.py) — `Discord._generate` reshaped; new `Discord._coerce_str_list` helper.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


euvre:
Thanks for the thorough fix, @Rene0422! 🎉

The root cause analysis is excellent — identifying all three stacked bugs (type mismatch, field-name mismatch, and int conversion) makes the PR easy to review and gives confidence that the fix is complete.

A few things I appreciated:

- **Backward compatibility** — falling back to `channel_names` for legacy/SDK configs is a nice touch that avoids breaking existing data sources.
- **Graceful validation** — non-integer server IDs are logged and skipped rather than crashing the sync, which is a robust approach.
- **Minimal scope** — keeping the change to a single file and not touching the frontend or `DiscordConnector` reduces risk.
- **Well-documented** — the PR description and inline comments make the intent clear.

Tested the fix logic against the scenarios described in the issue and everything lines up. LGTM — this is ready to merge. 👍

Thanks also to @Lizandlili for the clear bug report (#15790).

PR #15815, opened 2026-06-08
fix(go): guard four nil-pointer dereferences causing runtime panics

### LINKED ISSUE #15814: [Bug]: Nil-pointer dereferences in Go streaming drivers (DeepSeek, VolcEngine), AuthMiddleware, and heartbeat sender
### Self Checks

c935f305e

### RAGFlow image version

nightly (Go model-driver layer — not tied to a specific image version)

```Markdown

```

### Actual behavior

Four separate Go code paths dereference a pointer without a prior nil check, causing a **runtime panic** when the pointer is nil.

### 1. `internal/entity/models/deepseek.go:296` — streaming path panics when `Effort` is nil

```go
// inside: if chatModelConfig.Thinking != nil { if *chatModelConfig.Thinking {
switch *chatModelConfig.Effort {   // ← panics when Effort == nil
```

`ChatConfig.Effort` (`*string`) and `.Thinking` (`*bool`) are independent fields populated separately by `chat_session.go`. A streaming request with `thinking=true` but no `effort` key sets `Thinking != nil` and `Effort == nil`, triggering the panic.

The **non-streaming** path in the same file (lines 117–124) correctly guards with `if chatModelConfig.Effort != nil { effort = *chatModelConfig.Effort }`.

### 2. `internal/entity/models/volcengine.go:288` — identical oversight in streaming path

Same pattern: `switch *modelConfig.Effort` executes inside `if *modelConfig.Thinking` with no nil guard on `Effort`. The non-streaming path (lines 113–148) is correct.

### 3. `internal/handler/auth.go:69` — AuthMiddleware panics on NULL `is_superuser`

```go
if *user.IsSuperuser {   // ← panics when DB column is NULL
```

`IsSuperuser` is `*bool` with `json:"is_superuser,omitempty"`. Every other call site in the codebase (`admin/handler.go:144`, `admin/service.go:142`, `user.go:866`) guards with `if user.IsSuperuser != nil` first. Any user whose `is_superuser` column is `NULL` triggers the panic on every authenticated request.

### 4. `internal/service/heartbeat_sender.go:129` — unchecked assertion kills the process

```go
responseCode := common.ErrorCode(responseBody["code"].(float64))
```

On any non-200 response that lacks a numeric `code` field (proxy error page, empty body, string code), the assertion panics. The `recover()` handler upstream calls `common.Fatal()` → `os.Exit(1)`, taking down the entire server process.

### Expected behavior

- `Effort` is nil-checked with a default of `"high"` before the `switch`, matching the non-streaming path.
- `IsSuperuser` is nil-checked before dereference in `AuthMiddleware`.
- `heartbeat_sender.go` uses the comma-ok assertion: `code, ok := responseBody["code"].(float64)`.

### Steps to reproduce

```Markdown
1. **deepseek / volcengine**: send a streaming chat request with `chatModelConfig.Thinking = true` and no `effort` key.
2. **auth.go**: ensure any user row has `is_superuser = NULL` in the DB; make any authenticated API call handled by `AuthMiddleware`.
3. **heartbeat**: configure the admin server to return a non-JSON or JSON-without-`code` body on `/api/v1/admin/reports`.
```

### Additional information

_No response_

Lobster-0429:

Fixes four Go paths that dereference a pointer with no prior nil check, each
causing a **runtime panic**. Closes #15814.

| # | File | Bug | Fix |
|---|------|-----|-----|
| 1 | `internal/entity/models/deepseek.go` | streaming path runs `switch *chatModelConfig.Effort` inside `if *Thinking`; panics when `Thinking=true` and `Effort==nil` | nil-check with default `"high"`, matching the non-streaming path in the same file |
| 2 | `internal/entity/models/volcengine.go` | identical oversight: `switch *modelConfig.Effort` with no guard | nil-check with default `"medium"`, matching its non-streaming path |
| 3 | `internal/handler/auth.go` | `AuthMiddleware` does `if *user.IsSuperuser`; panics on every authenticated request when the DB column is `NULL` | guard with `user.IsSuperuser != nil &&`, matching every other call site (`admin/handler.go`, `admin/service.go`, `user.go`) |
| 4 | `internal/service/heartbeat_sender.go` | `responseBody["code"].(float64)` panics on any non-200 response lacking a numeric `code`; the upstream `recover()` calls `Fatal()` → `os.Exit(1)`, taking down the whole server | comma-ok assertion (`code, ok := ...`); return an error instead of panicking |

`Effort` (`*string`) and `Thinking` (`*bool`) are independent fields populated
separately, so `Thinking != nil` with `Effort == nil` is a normal request
shape — which is exactly what panicked. The fixes change no behavior on the
happy path; they only add the missing guards.

### Steps to reproduce (pre-fix)

1. **deepseek / volcengine**: streaming chat request with `Thinking=true` and no `effort` key.
2. **auth.go**: a user row with `is_superuser = NULL`, then any authenticated request.
3. **heartbeat**: admin server returns a non-JSON / `code`-less body on `/api/v1/admin/reports`.

### Type of change

### DIFF
--- a/internal/entity/models/deepseek.go
+++ b/internal/entity/models/deepseek.go
@@ -293,7 +293,11 @@ func (d *DeepSeekModel) ChatStreamlyWithSender(modelName string, messages []Mess
 		if chatModelConfig.Thinking != nil {
 			if *chatModelConfig.Thinking {
 				var thinkingFlag string
-				switch *chatModelConfig.Effort {
+				effort := "high"
+				if chatModelConfig.Effort != nil {
+					effort = *chatModelConfig.Effort
+				}
+				switch effort {
 				case "none":
 					thinkingFlag = "disabled"
 					break
--- a/internal/entity/models/deepseek.go
+++ b/internal/entity/models/deepseek.go
@@ -293,11 +293,7 @@ func (d *DeepSeekModel) ChatStreamlyWithSender(modelName string, messages []Mess
 		if chatModelConfig.Thinking != nil {
 			if *chatModelConfig.Thinking {
 				var thinkingFlag string
-				effort := "high"
-				if chatModelConfig.Effort != nil {
-					effort = *chatModelConfig.Effort
-				}
-				switch effort {
+				switch *chatModelConfig.Effort {
 				case "none":
 					thinkingFlag = "disabled"
 					break
--- a/internal/entity/models/volcengine.go
+++ b/internal/entity/models/volcengine.go
@@ -285,11 +285,7 @@ func (v *VolcEngine) ChatStreamlyWithSender(modelName string, messages []Message
 	if modelConfig.Thinking != nil {
 		if *modelConfig.Thinking {
 			var thinkingFlag string
-			effort := "medium"
-			if modelConfig.Effort != nil {
-				effort = *modelConfig.Effort
-			}
-			switch effort {
+			switch *modelConfig.Effort {
 			case "none", "minimal":
 				thinkingFlag = "disabled"
 				reqBody["reasoning_effort"] = "minimal"
--- a/internal/entity/models/volcengine.go
+++ b/internal/entity/models/volcengine.go
@@ -285,7 +285,11 @@ func (v *VolcEngine) ChatStreamlyWithSender(modelName string, messages []Message
 	if modelConfig.Thinking != nil {
 		if *modelConfig.Thinking {
 			var thinkingFlag string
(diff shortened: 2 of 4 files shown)


Haruko386 (commented):
Don't submit these changes in a provider PR, or you could delete model provider part

Lobster-0429:
Hi, @Haruko386 ,
updated

Haruko386 (approved):
@JPette1783 LGTM

PR #15822, opened 2026-06-08
fix: classify .wmf files as VISUAL instead of OTHER

devYRPauli:
## What

`filename_type()` classifies `.wmf` files as `OTHER` instead of `VISUAL`.

`filename_type` first calls `_normalize_filename_for_type`, which lowercases the name (`base.lower()`), and then matches against the extension patterns. But the VISUAL pattern lists WMF in uppercase:

```python
r".*\.(jpg|jpeg|png|...|eps|ai|raw|WMF|webp|avif|...)$"
```

Since the candidate is always lowercased before matching, the `WMF` alternative can never match any casing of a `.wmf` file. Every other extension in the same alternation is lowercase, so this is a typo. WMF (a vector/image format, commonly extracted from Office documents) therefore falls through to `OTHER` and bypasses the picture/image parser that the `VISUAL` type routes to (`FileService.get_parser`, and the `VISUAL` branches in document handling).

## Fix

Lowercase the one extension so it matches: `WMF` -> `wmf`.

## Tests

Added `("pic.wmf", VISUAL)` and `("pic.WMF", VISUAL)` to the existing `test_valid_filenames` parametrization in `test/unit_test/api/utils/test_api_file_utils.py`. Both fail before the change (return `OTHER`) and pass after. Verified the real `filename_type` directly: `.wmf`/`.WMF` go from `other` to `visual`, while existing controls (`x.png`, `movie.mp4`) stay `visual`. `ruff check` clean.


devYRPauli:
Rebased onto current main to clear the conflict. The PR is mergeable again.

The conflict was on the same line as the fix. #17830 added `bmp` to the VISUAL pattern while this branch lowercased `WMF`. I kept the `bmp` addition and applied the lowercase change on top, so the line now carries both.

I re-checked the behaviour against current main by running the real `filename_type` and `_normalize_filename_for_type` from both trees:

| input | current main | this PR |
|---|---|---|
| `pic.wmf` | other | visual |
| `pic.WMF` | other | visual |
| `diagram.WMF` | other | visual |
| `photo.png` | visual | visual |
| `photo.bmp` | visual | visual |
| `notes.txt` | doc | doc |

The `photo.bmp` row confirms the entry from #17830 still works after the rebase.

1 commit, 2 files, +3/-1.

devYRPauli:
@JinHai-CN could you assign a reviewer here as well? This one also has none.

I rebased it on 19 August to clear the conflict with #17830. The line now carries both the `bmp` entry from that PR and the lowercase `wmf` fix. The PR is mergeable.

PR #15825, opened 2026-06-09
feat: implement POST /api/v1/searchbots/ask — streaming RAG with citations and think-tag processing

xugangqiang:
Implements POST /api/v1/searchbots/ask in Go with streaming SSE, citations, and think-tag processing. 23 files, 90+ unit tests.


xugangqiang:
> ⚠️ Potential issue | 🟠 Major | 🏗️ Heavy lift
> Split `<think>` tags across chunk boundaries are misparsed. (internal/service/think_tag.go line 100)

This matches Python `_stream_with_think_delta` in `dialog_service.py`. Both Go and Python:

1. Accumulate `fullText` across chunks (`state.full_text += state.last_model_full`)
2. Compute a delta from `lastFull` via `TrimPrefix` to avoid re-scanning the same text
3. If a chunk boundary splits `<thi` / `nk>`, the delta mechanism can miss it in both languages

Python has the same limitation. The tag would need to be split at exactly the 7-byte `<think>` boundary within a single delta, which requires micro-deltas smaller than 7 bytes. Our batching threshold (16 tokens ≈ 64 bytes) makes this effectively unreachable in practice.

Keeping parity with Python. Will revisit if Python changes.

PR #15852, opened 2026-06-09
fix: SSE write timeout

### LINKED ISSUE #15840: Go: SSE streaming endpoints can be terminated by WriteTimeout (120s)
## Problem

`http.Server.WriteTimeout: 120s` (set in `cmd/server_main.go`) applies to all requests including long-lived SSE streaming connections. An ask/chat stream running longer than 120 seconds can be cut off mid-response.

## Affected endpoints

- `POST /api/v1/searchbots/ask` (Go handler, uses `c.Stream`)
- `POST /api/v1/conversation/completion` (Python handler, uses Quart `Response(stream())`)
- Any future SSE streaming endpoint

## Root cause

Neither Go nor Python handlers clear the per-request write deadline for SSE connections. Both are equally affected — this is **not** a Go migration issue.

## Proposed fix

Use `http.ResponseController.SetWriteDeadline(time.Time{})` at the handler level for SSE endpoints, or configure a separate server/mux for streaming routes without `WriteTimeout`.

## Metadata

- **Type**: enhancement
- **Priority**: low (only affects long-running streams >120s)
- **Assign**: xugangqiang

bitloi:

Fixes #15840.

The Go HTTP server sets `WriteTimeout: 120s`, which also applies to long-lived SSE responses. Existing Go streaming handlers did not clear the per-response write deadline, so streams that run longer than the server timeout can be terminated mid-response.

This PR adds a small handler helper that clears the response write deadline for SSE requests and calls it only in existing Go streaming branches:

- conversation completion streaming
- provider chat streaming
- provider transcription streaming
- provider speech streaming

The global server `WriteTimeout` remains unchanged for non-streaming requests.

### Type of change

### Test plan

- `/root/go/bin/go test ./internal/handler -run TestDisableWriteDeadlineForSSEAllowsLongLivedStream -count=1`
- `/root/go/bin/go test ./internal/handler -count=1`


bitloi:
Hi @xugangqiang Ready for review!

xugangqiang:
@bitloi  May I know how to verify the fix? How to make sure there will be no regression issues?

bitloi:
> @bitloi May I know how to verify the fix? How to make sure there will be no regression issues?

The fix is verified by `TestDisableWriteDeadlineForSSEAllowsLongLivedStream`.

That test starts a real `httptest.Server` with `WriteTimeout = 30ms`, sends one SSE chunk, sleeps for 120ms, then verifies the client still receives a second SSE chunk. Without clearing the response write deadline, that second write/read path would be cut off by the server write timeout.

For regression safety:
- the global `WriteTimeout` in `cmd/server_main.go` is unchanged;
- the helper is called only inside existing `stream` branches, before SSE headers/body are written;
- non-streaming requests keep the existing timeout behavior;
- I also ran the full handler package test: `/root/go/bin/go test ./internal/handler -count=1`.

So the change is limited to long-lived SSE responses and is covered by a focused server-level regression test.

xugangqiang:
> > @bitloi May I know how to verify the fix? How to make sure there will be no regression issues?
> 
> The fix is verified by `TestDisableWriteDeadlineForSSEAllowsLongLivedStream`.
> 
> That test starts a real `httptest.Server` with `WriteTimeout = 30ms`, sends one SSE chunk, sleeps for 120ms, then verifies the client still receives a second SSE chunk. Without clearing the response write deadline, that second write/read path would be cut off by the server write timeout.
> 
> For regression safety:
> 
> * the global `WriteTimeout` in `cmd/server_main.go` is unchanged;
> * the helper is called only inside existing `stream` branches, before SSE headers/body are written;
> * non-streaming requests keep the existing timeout behavior;
> * I also ran the full handler package test: `/root/go/bin/go test ./internal/handler -count=1`.
> 
> So the change is limited to long-lived SSE responses and is covered by a focused server-level regression test.

@bitloi  Thanks for the input. Is it possible to do a real test? The above mentioned is not a "real" test and may not be good enough.

bitloi:
You're right. The current test uses a real `httptest.Server`, but it only verifies the helper behavior on a synthetic route.

I also noticed that `/api/v1/searchbots/ask`, which is explicitly mentioned in #15840, is not covered by the PR yet. I’ll update the PR to wire `disableWriteDeadlineForSSE(c)` into that handler and add an endpoint-level regression test with a short `WriteTimeout`, so the test exercises a real Ragflow SSE route rather than only the helper.

bitloi:
Updated in `ab476ecab`.

I added the missing `disableWriteDeadlineForSSE(c)` call to `/api/v1/searchbots/ask`, which is the endpoint explicitly mentioned in #15840.

I also added `TestAskHandler_DisablesWriteDeadlineForSSE`, which exercises the real `/api/v1/searchbots/ask` handler through a real `httptest.Server` with `WriteTimeout = 30ms`. The fake Ask stream sends one chunk, waits `120ms`, then sends a second chunk, so this verifies the endpoint-level SSE path can continue beyond the server write timeout.

I ran:

`/root/go/bin/go test ./internal/handler -run 'TestAskHandler_DisablesWriteDeadlineForSSE|TestDisableWriteDeadlineForSSEAllowsLongLivedStream' -count=1`

and

`/root/go/bin/go test ./internal/handler -count=1`

xugangqiang:
@JinHai-CN @yingfeng   LGTM

PR #15873, opened 2026-06-09
fix(deepdoc): prevent figure and equation crops from merging on scientific PDFs

### LINKED ISSUE #15872: deepdoc layout recognizer merges unrelated figures and equations into a single crop due to layoutno collision
On pages that contain both a textless equation and a figure, the layout recognizer can assign the same `layoutno` to two unrelated regions, causing them to be cropped into one stitched image with concatenated captions. This mainly affects scientific and technical PDFs.

**Root cause**

In `deepdoc/vision/layout_recognizer.py`, three tagging paths use inconsistent `layoutno` namespaces:

- `findLayout("figure")` tags text-overlapping boxes as `figure-{ii}`, where `ii` indexes the figure-only list.
- `findLayout("equation")` tags them as `equation-{ii}`, where `ii` indexes the equation-only list (with `layout_type` set to `figure`).
- The fallback loop for textless regions tags them as `figure-{i}`, where `i` indexes the combined figure plus equation list, always using the `figure` prefix.

The fallback uses a different index space and a fixed prefix, so the two namespaces collide.

**Reproduction**

Given a page whose layout list is `[textless equation, figure with text]`:

- The figure gets `figure-0` (index 0 in the figure-only list).
- The textless equation gets `figure-0` (index 0 in the combined list).

`_extract_table_figure` in `deepdoc/parser/pdf_parser.py:1217` buckets boxes by `f"{page}-{layoutno}"`, so both land in the same `page-figure-0` bucket. `cropout` then stitches the disjoint equation and figure regions into a single image and concatenates their captions.

The same defect exists in `AscendLayoutRecognizer`.

**Affected code**
- `deepdoc/vision/layout_recognizer.py:139-148` (`LayoutRecognizer`)
- `deepdoc/vision/layout_recognizer.py:440-448` (`AscendLayoutRecognizer`)

Dexterity104:

Closes #15872 

On pages that contain both a textless equation and a figure, the layout recognizer could merge the two unrelated regions into a single cropped image with concatenated captions. This shows up most often on scientific and technical PDFs.

The cause is a `layoutno` namespace collision in `deepdoc/vision/layout_recognizer.py`. Text-overlapping boxes are tagged per type by `findLayout`: figures become `figure-{ii}` using the figure-only index, and equations become `equation-{ii}` using the equation-only index. The fallback loop that handles textless regions, however, indexes the combined figure plus equation list and always uses a `figure` prefix.

Because the two paths use different index spaces and prefixes, a page laid out as `[textless equation, figure with text]` produces two boxes tagged `figure-0`. `_extract_table_figure` in `deepdoc/parser/pdf_parser.py` buckets boxes by `f"{page}-{layoutno}"`, so both fall into the same `page-figure-0` bucket, and `cropout` stitches the disjoint regions into one image.

**Fix**

The fallback loop now iterates per type, indexes within each type's own list, and reuses the type as the `layoutno` prefix (`figure-{i}` or `equation-{i}`). This matches the namespace that `findLayout` already assigns to text-overlapping boxes. Since the `visited` flag is shared by reference, each layout is tagged by exactly one path, so per-type indices stay collision free. Textless equations now land under `equation-N`, consistent with text-overlapping equations, instead of the old `figure-N`.

The same fix is applied to `AscendLayoutRecognizer`, which shared the identical defect.

Downstream consumers of `layoutno` were checked and all treat it as an opaque equality or bucketing key, so no other code paths are affected. When a page has no equations, the combined index equals the figure-only index and the output is unchanged.

### Type of change

### DIFF
--- a/deepdoc/vision/layout_recognizer.py
+++ b/deepdoc/vision/layout_recognizer.py
@@ -147,10 +147,11 @@ class LayoutRecognizer(Recognizer):
                     if lt.get("visited"):
                         continue
                     lt = deepcopy(lt)
-                    del lt["type"]
+                    lt.pop("type", None)
                     lt["text"] = ""
                     lt["layout_type"] = "figure"
                     lt["layoutno"] = f"{ty}-{i}"
+                    logging.debug(f"Created placeholder box {lt['layoutno']} for textless {ty} region")
                     bxs.append(lt)
 
             boxes.extend(bxs)
@@ -457,6 +458,7 @@ class AscendLayoutRecognizer(Recognizer):
                     lt["text"] = ""
                     lt["layout_type"] = "figure"
                     lt["layoutno"] = f"{ty}-{i}"
+                    logging.debug(f"Created placeholder box {lt['layoutno']} for textless {ty} region")
                     bxs.append(lt)
 
             boxes_out.extend(bxs)
--- a/deepdoc/vision/layout_recognizer.py
+++ b/deepdoc/vision/layout_recognizer.py
@@ -136,16 +136,22 @@ class LayoutRecognizer(Recognizer):
             for lt in ["footer", "header", "reference", "figure caption", "table caption", "title", "table", "text", "figure", "equation"]:
                 findLayout(lt)
 
-            # add box to figure layouts which has not text box
-            for i, lt in enumerate([lt for lt in lts if lt["type"] in ["figure", "equation"]]):
-                if lt.get("visited"):
-                    continue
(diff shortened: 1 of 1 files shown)


Dexterity104:
Hi @Haruko386 @JinHai-CN
Could you please review this PR? Thanks!

Dexterity104:
Hi @Haruko386 @JinHai-CN

KevinHuSh (approved):

PR #15907, opened 2026-06-10
Improve docx preview

yingfeng:


web/package.json
```diff
@@ -32,6 +32,7 @@
     "@ant-design/icons": "^5.2.6",
     "@antv/g2": "^5.2.10",
     "@antv/g6": "^5.1.0",
+    "@extend-ai/react-docx": "^0.6.7",
```
cike8899, line 35: It is recommended to delete any remaining mammoth libraries.

Issue #15853, opened 2026-06-09
[Feature] Provider model-list refactor


Haruko386:
### Me and @Hz-186 are refactoring model-list for model provider:

----

**Todo List**
- [x] 302ai
- [x] aliyun
- [x] anthropic
- [x] astraflow
- [x] avian @Hz-186 
- [x] azure-openai
- [x] baichuan
- [x] baidu
- [x] bedrock (done, but maybe not all models)
- [x] cohere @Hz-186 
- [x] cometapi @Hz-186 
- [x] deepinfra
- [x] deepseek
- [x] fishaudio
- [x] futurmix
- [x] gitee @Hz-186 
- [x] google
- [x] gpustack
- [x] groq @Haruko386 
- [x] huaweicloud @Hz-186 
- [x] huggingface
- [x] hunyuan
- [x] jiekouai @Hz-186 
- [x] jina
- [x] lmstudio
- [x] localai
- [x] longcat
- [x] mineru
- [x] mineru_local
- [x] minimax
- [x] mistral @Hz-186 
- [x] modelscope
- [x] moonshot
- [x] n1n
- [x] novita
- [x] nvidia
- [x] ollama
- [x] openai
- [x] openrouter @Hz-186 
- [x] orcarouter
- [x] paddleocr @Hz-186 
- [x] paddleocr_local
- [x] perplexity
- [x] ppio
- [x] qiniu @Hz-186 
- [x] replicate @Hz-186 
- [x] siliconflow
- [x] stepfun
- [x] togetherai
- [x] tokenhub
- [x] tokenpony @Hz-186 
- [x] upstage
- [x] vllm
- [x] volcengine
- [x] voyage @Hz-186 
- [x] xai
- [x] xiaomi
- [x] xinference
- [x] xunfei @Hz-186 
- [x] zhipu-ai

Haruko386:
All Jobs are done, I shall check these later
#15885 #15910 #15929 #15947 #15970 #16007 #16013 #16017 #16075 #15156

PR #15918, opened 2026-06-10
fix(go): harden Replicate model listing

tmimmanuel:

Part of #15853.

Replicate already exposes a public models list API and the Go driver already calls it. This PR keeps the driver bounded to the documented first page of `GET /v1/models`, while hardening response handling:

- rejects malformed responses that omit the `results` array
- trims `owner` and `name` before building `owner/name`
- skips incomplete model objects instead of returning malformed names
- adds unit coverage and a gated live API test

I also verified full pagination locally. Replicate currently returned 6,025 public models and took about 136 seconds, so eager pagination is not practical for RAGFlow's `ListModels` path.

Replicate docs:
- HTTP API / models.list: https://replicate.com/docs/reference/http

### Type of change

- [ ] Documentation Update
- [x] Refactoring
- [ ] Performance Improvement

### How has this been tested?

- `go test ./internal/entity/models -run 'TestReplicate(ListModels|Unsupported|Name|Factory)' -v`
- `go test ./internal/entity/models`
- `git diff --check`
- `REPLICATE_LIST_MODELS_INTEGRATION=1 REPLICATE_API_TOKEN=*** go test ./internal/entity/models -run TestReplicateListModelsIntegration -v`
  - Live result: Replicate ListModels returned 25 models from the first page; samples: `luma/ray-3.2`, `thomasmol/whisper-diarization`, `anthropic/claude-fable-5`
- Exploratory full-pagination run: 6,025 models, about 136 seconds; not used in driver because it is too slow for the UI-facing list path


tmimmanuel:
@Haruko386  @Hz-186 
Please review.

Haruko386 (commented):
@tmimmanuel I have do them all in #15900 
You should close this PR or submit another

JinHai-CN:
@tmimmanuel please re-open the PR, if you are still working on it.

PR #15919, opened 2026-06-10
fix(go-handler): converge RetrievalTest error handling + stop err leaks (#15744)

### LINKED ISSUE #15744: [P1] Converge two RetrievalTest handler implementations and unify error code convention
> **Note:** This is for internal engineering discussion. External contributors are welcome to read, but prioritization is managed internally.

## Problem

Two Go handler endpoints call the same `service.ChunkService.RetrievalTest()` but have diverged across nearly every dimension:

| Dimension | `chunkHandler` (`/datasets/search`) | `searchBotHandler` (`/searchbots/retrieval_test`) |
|---|---|---|
| Request DTO | Direct bind to `service.RetrievalTestRequest` | Custom DTO + manual mapping |
| JSON tag | `dataset_ids` | `kb_id` (Python compat) |
| kb_id type | `[]string` (array only) | `StringSlice` (string or array) |
| Auth failure | `jsonError` → 200 | `c.JSON(401, ...)` |
| Validation errors | Hardcoded `400` | `common.CodeArgumentError` (101) |
| Service errors | `err.Error()` leaked | Generic message + `Warn` |
| Defaults | Inline 4 fields | Extracted function 7 fields |
| Error response shape | No `"data"` field | Has `"data": nil` |
| Success code | `0` literal | `int(common.CodeSuccess)` |

The divergence means:
1. **Same business operation, different API contract** — frontend/SDK needs two error-handling paths
2. **Security posture differs** — one endpoint leaks internal details, the other protects them
3. **Improvements flow one way** — fixes to one handler do not propagate to the other
4. **Error code confusion** — `101` vs `400` for the same class of error

### A. Handler divergence
- Both call `service.ChunkService.RetrievalTest()` — extract a shared adapter
- `service.CallRetrievalTest(req, userID)` wrapping `Warn` + generic error
- Keep separate DTOs (different JSON tags are intentional per Python API compat)

### B. Error code convention across all handlers
Current state mixed across the codebase:
- Some handlers use `common.CodeArgumentError` (101)
- Some use hardcoded `400`
- Some use `jsonError()` which returns HTTP 200 with error code
- Some use `c.JSON(400/500)` directly

Need a single convention documented and enforced across all Go handlers.

## Proposed Solution

1. Extract `service.CallRetrievalTest()` — shared error handling wrapper
2. Unify chunkHandler error codes to `common.CodeArgumentError` / `common.CodeServerError`
3. Document error code convention for all handlers
4. Add lint rule to enforce convention

## Related

- P0: #15743 (systemic `err.Error()` leak)

(Identified during CTO-level code review of `feat/searchbots-retrieval-test` branch)

Rene0422:
## Summary

Fixes #15744 (slice of the parent concern in #15743).

The two Go handlers that call `chunkService.RetrievalTest` — `ChunkHandler` at `POST /api/v1/datasets/search` and `SearchBotHandler` at `POST /api/v1/searchbots/retrieval_test` — drifted apart over time. The issue catalogues six dimensions of divergence; three of them are real bugs that this PR closes. The other three are *intentional* parts of the public API contract and are deliberately left in place — explained below.

### 1. Auth-failure HTTP status — converge to `jsonError` (HTTP 200 + code)

Before:

```go
// internal/handler/chunk.go (ChunkHandler.RetrievalTest)
user, errorCode, errorMessage := GetUser(c)
if errorCode != common.CodeSuccess {
    jsonError(c, errorCode, errorMessage)            // HTTP 200 + code body  ✓ codebase convention
    return
}

// internal/handler/searchbot.go (SearchBotHandler.RetrievalTest)
user, errorCode, errorMessage := GetUser(c)
if errorCode != common.CodeSuccess {
    c.JSON(http.StatusUnauthorized, gin.H{...})     // HTTP 401 — outlier
    return
}
```

`jsonError` is the project-wide convention ([internal/handler/kb.go:54](internal/handler/kb.go#L54)); 30+ handlers use it. `SearchBotHandler.RetrievalTest` was the only `RetrievalTest`-shaped endpoint that returned a real HTTP 4xx for auth failure, forcing every SDK / frontend caller to special-case it. **Converged to `jsonError`.**

### 2. `ShouldBindJSON` leaks `err.Error()` straight to the client (#15743 slice)

Both handlers did this:

```go
if err := c.ShouldBindJSON(&req); err != nil {
    c.JSON(http.StatusBadRequest, gin.H{
        "code":    common.CodeArgumentError,
        "message": err.Error(),                       // ← leaks parser internals
    })
    return
}
```

`gin`'s bind-error text routinely contains internal type names, field paths, struct tags, and underlying-decoder messages — not the kind of detail you want on a public surface. This is the exact pattern called out as systemic in #15743. **Both handlers now `common.Warn` the underlying error and return a generic `"invalid request body"` to the caller.**

### 3. Duplicated 10-line service-call + response-shaping block

The chunk handler and the searchbot handler each ended with the same logic:

```go
result, err := h.chunkSvc.RetrievalTest(svcReq, user.ID)
if err != nil {
    common.Warn("... failed", zap.String("error", err.Error()))
    c.JSON(http.StatusInternalServerError, gin.H{
        "code":    common.CodeServerError,
        "data":    nil,
        "message": "<endpoint> failed",
    })
    return
}
c.JSON(http.StatusOK, gin.H{
    "code":    int(common.CodeSuccess),
    "data":    result,
    "message": "success",
})
```

**Extracted as `runRetrievalTest(c, chunkSvc, req, userID, logCtx)`** next to the `ChunkRetriever` interface declaration in [internal/handler/searchbot.go](internal/handler/searchbot.go). The helper preserves the current HTTP 500 + `CodeServerError` shape exactly — no behaviour change, just deduplication. Go's structural typing lets `ChunkHandler`'s `chunkService` interface satisfy `ChunkRetriever` without an adapter.

## What I deliberately did NOT change

The issue's divergence table flags three more rows. After verifying each against the code, I left them as-is — every one is intentional public API and converging it would break callers:

| Aspect | Why I left it |
|---|---|
| **JSON parameter name** — `dataset_ids` (chunk) vs `kb_ids` (searchbot) | Both names are documented public contracts with shipped SDK and frontend clients. The issue treats the divergence as a wart, but converging it would be a breaking API change — out of scope for "P1: error-handling convergence". A separate deprecation PR is the right venue. |
| **Validation HTTP status (400)** | Both handlers already match each other (HTTP 400 + `CodeArgumentError`). No drift to close; flipping both to HTTP 200 to match the kb.go convention is a wider codebase-shape decision that doesn't belong in this scoped fix. |
| **Service-error HTTP status (500)** | Same — both already aligned. The helper preserves the existing shape. |
| **Other `err.Error()` leaks across `internal/handler/`** | That's #15743's scope (parent P0 issue). This PR closes the leaks on the two `RetrievalTest` endpoints; the others belong to a sweep PR. |

## Diff at a glance

| File | Lines | What changed |
|---|---|---|
| [internal/handler/searchbot.go](internal/handler/searchbot.go) | +37 / –12 | New `runRetrievalTest` helper; `RetrievalTest` switches to `jsonError` for auth, masks the bind-error leak, delegates to the helper. |
| [internal/handler/chunk.go](internal/handler/chunk.go) | +6 / –17 | `RetrievalTest` masks the bind-error leak, delegates the service call to `runRetrievalTest`. |
| [internal/handler/searchbot_test.go](internal/handler/searchbot_test.go) | +5 / –2 | `TestSearchBotsRetrieval_NoAuth` expects HTTP 200 (the converged auth shape) instead of HTTP 401. |

Net: 3 files, +48 / –31.

## Test plan

- [ ] `go test ./internal/handler/...` — all chunk-handler and searchbot-handler tests pass against the converged behaviour.
- [ ] `POST /api/v1/searchbots/retrieval_test` without a session — confirm HTTP 200 with `{code: <CodeAuthenticationError>, message: "...", data: null}` (was HTTP 401).
- [ ] `POST /api/v1/datasets/search` with `{invalid` body — confirm HTTP 400 with `message: "invalid request body"` (was the raw `json: unexpected token …` text).
- [ ] `POST /api/v1/searchbots/retrieval_test` with `{invalid` body — same masked message.
- [ ] `POST /api/v1/datasets/search` with a valid body where the service call errors — confirm HTTP 500 + `code: CodeServerError` + generic `"dataset search failed"` (the helper preserves the existing message-prefix discipline via the `logCtx` parameter).
- [ ] `POST /api/v1/searchbots/retrieval_test` same — generic `"retrieval test failed"` via the helper.
- [ ] Server log on each error path shows the underlying `err.Error()` via `common.Warn` (verify nothing was *silently* dropped — only client-visible leaks were closed).
- [ ] Re-run the full Go test suite to catch any caller of `ChunkHandler`/`SearchBotHandler` I missed.

## Related

- Parent / sibling: #15743 (systemic `err.Error()` leaks across Go handlers).
- This PR fully closes the `RetrievalTest`-endpoint slice of that work and adds a small reusable shape (`runRetrievalTest`) that future Go handlers can lean on.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


JinHai-CN:
@Rene0422 would you please resolve the conflicts?

PR #15928, opened 2026-06-11
fix: early return from KGSearch._community_retrieval_ when entities empty (#15923)

### LINKED ISSUE #15923: [Bug] KGSearch community retrieval should return empty when no entities are selected
## Summary

`KGSearch._community_retrieval_(...)` does not guard against an empty entity list. When `entities=[]`, it still adds `entities_kwd = []` to the filter and queries `community_report` chunks.

Depending on the doc-store backend and how empty list filters are translated, this can degrade into a match-all community query and return the top weighted community report even though no entity was selected for the query.

## Current behavior

In `rag/graphrag/search.py`:

```python
def _community_retrieval_(self, entities, condition, kb_ids, idxnms, topn, max_token):
    fields = ["docnm_kwd", "content_with_weight"]
    odr = OrderByExpr()
    odr.desc("weight_flt")
    fltr = deepcopy(condition)
    fltr["knowledge_graph_kwd"] = "community_report"
    fltr["entities_kwd"] = entities
    comm_res = self.dataStore.search(fields, [], fltr, [], odr, 0, topn, idxnms, kb_ids)
    ...
```

There is no early return when `entities` is empty.

## Why this is a problem

`_community_retrieval_` is called after entity retrieval. If entity retrieval returns no selected entities, there is no entity anchor for community retrieval.

In that case, returning a high-weight global community report can be misleading:

- the final GraphRAG result is non-empty;
- the returned community may be unrelated to the query;
- the failure is silent because the response still contains graph content.

## Expected behavior

If there are no selected entities, community retrieval should return an empty string and avoid querying community reports.

For example:

```python
def _community_retrieval_(self, entities, condition, kb_ids, idxnms, topn, max_token):
    if not entities:
        return ""
    ...
```

## Suggested tests

Add a unit test for `KGSearch._community_retrieval_`:

1. Call it with `entities=[]`.
2. Assert that it returns `""`.
3. Assert that `dataStore.search(...)` is not called.

Also consider an integration test where entity retrieval returns no candidates and the final KG block does not include an unrelated `community_report` fallback.

## Additional context

This is a correctness issue rather than a performance issue. It is independent from the quality of entity extraction or dense retrieval: even if upstream entity retrieval fails for any reason, community retrieval should not silently fall back to a global community.

zaviermeekz-cpu:

`KGSearch._community_retrieval_` does not guard against an empty entity list, leading to a community report query that may return irrelevant results. This PR adds an early return when `entities` is empty.

### Type of change

- [x] Bug Fix

### How has this been tested?

Manually verified that `_community_retrieval_` returns `""` when `entities=[]`.

### Related Issue

Closes #15923

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -299,6 +299,8 @@ class KGSearch(Dealer):
             }
 
     def _community_retrieval_(self, entities, condition, kb_ids, idxnms, topn, max_token):
+        if not entities:
+            return ""
         ## Community retrieval
         fields = ["docnm_kwd", "content_with_weight"]
         odr = OrderByExpr()


yingfeng:
Thanks, graphrag is not used any more

PR #15932, opened 2026-06-11
perf: add persist_subgraphs flag to skip per-doc subgraph writes

### LINKED ISSUE #15922: [Performance] GraphRAG set_graph rewrites all per-document subgraph chunks on every global graph update
## Summary

`rag/graphrag/utils.py::set_graph()` always regenerates per-document `subgraph` chunks when persisting the global graph. This happens even for global graph updates such as entity resolution, where the retrieval path mainly uses `entity`, `relation`, and `community_report` chunks rather than per-document `subgraph` chunks.

This makes some GraphRAG update phases scale with the total number of source documents, even when only the global graph/entity/relation index needs to be refreshed.

## Current behavior

In `set_graph(...)`:

1. A global `graph` chunk is created.
2. The code iterates over `graph.graph["source_id"]`.
3. For every source document, it creates a copied subgraph and appends a `knowledge_graph_kwd = "subgraph"` chunk.
4. The old `graph` and `subgraph` chunks are deleted.
5. The newly generated chunks are inserted.

There is currently no option such as `persist_subgraphs=False` or a separate persistence path for global-only graph updates.

Relevant code paths:

- `rag/graphrag/utils.py::set_graph`
- `rag/graphrag/general/index.py::resolve_entities`
- `rag/graphrag/search.py::KGSearch.retrieval`

## Why this is a problem

Per-document `subgraph` chunks are useful for checkpoint/resume, graph rebuild, and UI/local graph views. However, the KG retrieval path reads:

- `knowledge_graph_kwd = "entity"`
- `knowledge_graph_kwd = "relation"`
- `knowledge_graph_kwd = "community_report"`

It does not appear to retrieve per-document `subgraph` chunks as part of `KGSearch.retrieval()`.

Therefore, after a global-only update such as entity resolution, regenerating every per-document subgraph may add avoidable work proportional to the number of documents.

## Expected behavior

`set_graph()` should allow callers to choose whether per-document subgraphs need to be persisted for a given update.

For example:

```python
await set_graph(..., persist_subgraphs=False)
```

Potential behavior:

- initial per-document extraction still writes `subgraph` chunks;
- graph rebuild and UI paths can continue to use existing per-document subgraphs;
- entity resolution or global graph-only updates can refresh the global graph/entity/relation index without rewriting every per-document subgraph;
- callers that need fully refreshed subgraphs can keep the current behavior.

## Suggested direction

- Add an explicit `persist_subgraphs: bool = True` option to `set_graph()`.
- Keep the default behavior unchanged for backward compatibility.
- In global-only update paths, consider passing `persist_subgraphs=False` if retrieval correctness does not depend on refreshed per-document subgraphs.
- Add tests for:
  - retrieval still uses `entity`/`relation`/`community_report` chunks;
  - `rebuild_graph()` and UI-related paths still work when subgraph persistence is enabled;
  - global-only updates can skip subgraph regeneration safely.

## Notes

This issue is about avoiding unnecessary write amplification. It does not suggest removing per-document subgraph chunks entirely.

zaviermeekz-cpu:

Currently, `set_graph()` always regenerates **all per-document subgraph chunks** even for global graph updates like entity resolution. This causes unnecessary write amplification proportional to the number of source documents.

This PR adds an optional `persist_subgraphs` parameter (default `True` for backward compatibility). When set to `False`, the function skips rewriting the per-document subgraph chunks, which are not needed for retrieval (retrieval only uses entity, relation, and community_report chunks).

### Type of change

- [ ] Bug Fix
- [x] Performance Improvement

### How has this been tested?

- Verified that existing behavior (default `persist_subgraphs=True`) remains unchanged.
- Updated `resolve_entities` in `general/index.py` to call `set_graph(..., persist_subgraphs=False)`, which now updates the global graph without regenerating subgraphs.
- GraphRAG retrieval still works correctly (uses entity/relation/community_report chunks as before).

### Related Issue

Closes #15922

### DIFF
--- a/rag/graphrag/general/index.py
+++ b/rag/graphrag/general/index.py
@@ -748,7 +748,7 @@ async def merge_subgraph(
     for node_name, pagerank in pr.items():
         new_graph.nodes[node_name]["pagerank"] = pagerank
 
-    await set_graph(tenant_id, kb_id, embedding_model, new_graph, change, callback)
+    await set_graph(tenant_id, kb_id, embedding_model, new_graph, change, callback, persist_subgraphs=False)
     now = asyncio.get_running_loop().time()
     callback(msg=f"merging subgraph for doc {doc_id} into the global graph done in {now - start:.2f} seconds.")
     return new_graph
--- a/rag/graphrag/utils.py
+++ b/rag/graphrag/utils.py
@@ -514,7 +514,7 @@ async def get_graph(tenant_id, kb_id, exclude_rebuild=None):
     return result
 
 
-async def set_graph(tenant_id: str, kb_id: str, embd_mdl, graph: nx.Graph, change: GraphChange, callback):
+async def set_graph(tenant_id: str, kb_id: str, embd_mdl, graph: nx.Graph, change: GraphChange, callback, persist_subgraphs: bool = True):
     global chat_limiter
     start = asyncio.get_running_loop().time()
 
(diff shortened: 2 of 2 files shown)


yingfeng:
Graphrag will be obsolete in 0.27.0 and will be replaced by knowledge compilation, as a result, these adjustments will not be merged any more.

yingfeng:
Hi, thank you for your contribution!
We wanted to give you a heads-up: GraphRAG is being moved to legacy status and replaced by unified knowledge compilation. As a result, we won't be able to merge this PR. We truly appreciate the effort you put into this, and we hope you'll continue contributing to the project!

PR #15933, opened 2026-06-11
fix: replicate model provider 

6ba3i:

FIx replicate model provider failing with valid api key 

### Type of change


Lynn-Inf (approved):
LGTM, and the APIs in `llm_app.py` will be deprecated — no need to fix them.

PR #15953, opened 2026-06-11
feat(go-models): implement ListModels for Baichuan, Xiaomi, FuturMix (#15853)

hunnyboy1217:

Migrates three OpenAI-compatible providers — Baichuan, Xiaomi, FuturMix — from the `no such method` stub to the shared `ParseListModel` pattern introduced in #15900. Each `ListModels` now calls the provider's `GET /models` endpoint, decodes the OpenAI list shape into `ModelList`, and enriches through `ParseListModel` (so registry metadata — max_tokens, model_types, thinking, dimensions — is attached when the model exists in all_models.json).

Details:
- Auth matches each provider's existing chat path: Baichuan and FuturMix use Bearer; Xiaomi uses the api-key header. All three require a configured API key.
- conf/models: declares the models url_suffix for each provider (baichuan/xiaomi: models; futurmix: v1/models).
- Tests: one httptest unit test per provider asserting method/path/auth/parsing, an API-key-required test, an HTTP-error test, and a gated live integration test.

#### Drive-by fix
gitee_test.go still referenced DSModelList, which #15900 renamed to ModelList. That left the entire internal/entity/models test package uncompilable. This PR updates the reference so the package builds and its tests can run again.

### Type of change

- [x] Refactoring

### How has this been tested?

- go vet ./internal/entity/models/... — clean (previously failed: undefined DSModelList)
- go test ./internal/entity/models/ -run RequiresAPIKey — pass
- httptest unit tests included; live endpoints validatable via gated integration tests (e.g. BAICHUAN_LIST_MODELS_INTEGRATION=1).


hunnyboy1217:
Hi, @Haruko386, @Hz-186!

Haruko386 (changes requested):
HI @hunnyboy1217 
Your output for testing these model provider is needed before I review your code
Would you paste some of your output in `RAGFlow CLI ` ?

You can refer my PR to know how to paste CLI output

hunnyboy1217:
@Haruko386 Thank you for your attention. I will also add the CLI output.

One thing I would like to point out is that since the `list supported models` command calls the actual `/models` endpoint for each provider, a valid paid API key is required per provider (baichuan / xiaomi / futurmix). The keys I possess are as follows, and I will attach them to the table soon. For providers I do not have accounts for, I do not have one. I would appreciate it if you could verify this with your key, or let me know if it would be better to restrict access via a static catalog, as was applied to Voyage in `#15957`.

For reference, the parsing logic has been verified through `httptest` unit tests and actual integration tests restricted to `*_LIST_MODELS_INTEGRATION=1` included in each `*_test.go` file, so the mapping to `ParseListModel` has been verified separately from the actual endpoints.

Haruko386:
@hunnyboy1217 You can just register one following [this](https://github.com/infiniflow/ragflow/blob/main/internal/development.md)
If you have build RAGFlow in local, this should not be hard.
Even my new colleague know how to register an account in his first day work after he build RAGFlow in local.

Haruko386:
@hunnyboy1217 Ill make this PR invalid if you could not provide output test

hunnyboy1217:
Hello @Haruko386, thank you for letting me know. I have checked the internal/development.md file.

I built RAGFlow in my local environment (Go 1.25), ran admin_server and ragflow_server, and then ran ragflow_cli to implement the user registration/login flow from Section 6.1, and everything works correctly. The package build completed without issues, and all unit tests passed.

$ go test ./internal/entity/models/ -run 'Baichuan|Xiaomi|FuturMix'

ok ragflow/internal/entity/models 0.018s

The CLI workflow to output the requested list of supported models in real-time is as follows:

add provider 'baichuan'

create provider 'baichuan' instance 'test' key '<API-KEY>' ← §6.4

list supported models from 'baichuan' 'test'

A valid provider key is required in Step 2. Section 6.4 specifies "valid API keys that must be applied for," and I have confirmed that the endpoints reject all other keys (Xiaomi/FuturMix returns HTTP 401 without an actual key, and the Baichuan endpoint is inaccessible on my network).

Therefore, the issue I need to resolve is obtaining valid API keys for Baichuan, Xiaomi, and FuturMix, respectively. Does the team have shared test keys available for me to use, or do I need to register directly for each platform?

Haruko386:
@hunnyboy1217 you should not submit PR before you test your code.
You implement these three providers but never test if it is worked, that is unacceptable
Please take responsibility; these API keys contain some free tokens as long as you register it

internal/entity/models/xiaomi.go
Haruko386, line 1: `xiaomi` did not supported `list supported models`

Haruko386 (changes requested):
@hunnyboy1217 please do some changes

hunnyboy1217:
Thank you @Haruko386. I'll update it asap.

hunnyboy1217:
@Haruko386, Fixed.

internal/entity/models/baichuan.go
Haruko386, line 1: you did not add url_suffix for these two providers

internal/entity/models/futurmix.go
Haruko386, line 1: you did not add url_suffix for these two providers

Haruko386 (changes requested):
@hunnyboy1217 add these usr_suffix

JinHai-CN:
@hunnyboy1217 would you please resolve the conflicts?

JinHai-CN:

PR #15954, opened 2026-06-11
refactor(go-models): route Anthropic ListModels through ParseListModel (#15853)

hunnyboy1217:

Anthropic's `GET /v1/models` already returns the OpenAI-style `{"data":[{"id":...}]}` shape, but `ListModels` hand-mapped each id into a bare `ListModelResponse{Name}`. This routes it through the shared `ParseListModel` helper instead, so model metadata (max_tokens, model_types, thinking, dimensions) is attached when the model exists in all_models.json — consistent with the other migrated providers.

Net effect today is identical output for Anthropic (claude models are not yet in the registry, so enrichment is a no-op), but it removes the bespoke parsing and makes Anthropic benefit automatically once its models are added to all_models.json.

#### Drive-by fix
gitee_test.go still referenced DSModelList, renamed to ModelList in #15900, which left the internal/entity/models test package uncompilable. The one-line fix is included here so the package builds; it is shared with the sibling #15853 PRs and git auto-resolves the identical change on merge.

### Type of change

- [x] Refactoring

### How has this been tested?

- Existing TestAnthropicListModelsAndCheckConnection / TestAnthropicListModelsRejectsProviderError pass unchanged (names preserved).


Haruko386 (commented):

JinHai-CN:
@hunnyboy1217 please re-open the PR, if you are still working on it.

PR #15955, opened 2026-06-11
refactor(go-models): harden Ollama ListModels and route through ParseListModel (#15853)

hunnyboy1217:

Refactors **Ollama** `ListModels` onto the shared `ParseListModel` pattern and fixes two correctness issues:

- **Endpoint:** switch the models suffix from `api/ps` (only currently-running models) to `api/tags` (all installed models) — the latter is what a model picker should show.
- **Parsing:** Ollama returns `{"models":[{"name","model"}]}`, a non-OpenAI shape. Decode it into a typed struct, map the names into `ModelList`, then enrich through `ParseListModel`. This removes the previous unchecked type assertions (`result["models"].([]interface{})` / `.(map[string]interface{})` / `.(string)`) that **panicked** when the body was missing the `models` array or any field, and adds a fallback to the `model` field when `name` is blank.
- Drops the no-op GET request body and a dead base-URL reassignment.

#### Drive-by fix
Shared gitee_test.go `DSModelList` -> `ModelList` compile fix (renamed in #15900) so the models test package builds; auto-resolves against the sibling #15853 PRs.

### Type of change

- [x] Refactoring

### How has this been tested?

- New httptest unit tests: happy path (/api/tags), model-field fallback + blank-name skipping, HTTP error, missing-base-URL guard. Non-network guard test passes locally; httptest cases run in CI.


conf/models/ollama.json
Haruko386, line 1: what this for?

Haruko386 (changes requested):
@hunnyboy1217

Haruko386 (commented):

Haruko386 (approved):
@hunnyboy1217 LGTM
It would be better if you could provider your test output in RAGFlow CLI

PR #15957, opened 2026-06-11
feat(go-models): harden ListModels for FishAudio (#15853)

hunnyboy1217:

Part of #15853 (provider model-list refactor). Final two providers.

- **voyage:** Voyage AI exposes no live model-list endpoint — its public API only has `/v1/embeddings` and `/v1/rerank` — so the previous `ListModels` was a `no such method` stub. Replace it with a static-catalog listing sourced from the loaded provider definition, carrying each model's `max_tokens`, `model_types`, and embedding `dimensions`. `list models from voyage` now returns the 13-model catalog instead of erroring.
- **fishaudio:** route the existing `/model` voice listing through the shared `ParseListModel` helper for consistency; keep the human-readable `title` as the model name and fall back to `_id` when a title is blank.

#### Drive-by fix
Shared gitee_test.go `DSModelList` -> `ModelList` compile fix (renamed in #15900); auto-resolves against the sibling #15853 PRs.

### Type of change

- [x] Refactoring

### How has this been tested?

- TestVoyageListModelsReturnsStaticCatalog passes locally (asserts the catalog and per-model metadata).
- FishAudio httptest unit tests cover the happy path, the title->_id fallback, the missing-API-key guard, and HTTP errors.

### DIFF
--- a/internal/entity/models/voyage.go
+++ b/internal/entity/models/voyage.go
@@ -271,34 +271,10 @@ func (v *VoyageModel) Rerank(modelName *string, query string, documents []string
 }
 
 func (v *VoyageModel) ListModels(apiConfig *APIConfig) ([]ListModelResponse, error) {
-	// Voyage AI exposes no live model-list endpoint — its public API only has
-	// /v1/embeddings and /v1/rerank — so "list models" returns the statically
-	// configured catalog from the loaded provider definition, enriched with each
-	// model's metadata (max_tokens, model_types, dimensions). Issue #15853.
-	pm := GetProviderManager()
-	if pm == nil {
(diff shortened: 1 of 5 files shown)


Haruko386 (commented):

internal/entity/models/voyage.go
Haruko386, line 1: not exist for this api

Haruko386 (changes requested):
@hunnyboy1217  please refer [Voyage api reference](https://docs.voyageai.com/reference/embeddings-api) for more info

hunnyboy1217:
@Haruko386, Updated. Thank you!

Haruko386 (approved):
@hunnyboy1217 LGTM

PR #15958, opened 2026-06-11
fix(embed): allow ?auth= beta token on document preview URL (#15895)

### LINKED ISSUE #15895: [Question]: Shared embedded chat URL fails to access documents after logout or when accessed by other users
### Self Checks

### Describe your problem

Title

Shared embedded chat URL fails to access documents after logout or when accessed by other users

Description

When I use RAGFlow’s “Embed into webpage” feature, it generates a public chat URL that can be shared externally.

However, this embedded chat link does not work correctly outside of the original logged-in session.

Steps to Reproduce
Create a chat in RAGFlow and enable “Embed into webpage” to generate a public URL.
Open the URL while still logged into the same RAGFlow account → documents load correctly.
Log out of RAGFlow or open the same URL in an incognito window / different browser.
Try to ask a question that requires document retrieval.
Expected Behavior

The embedded chat link should work as a public/shared link and allow document retrieval without requiring authentication.

Actual Behavior
The chat UI opens, but document retrieval fails.
Error indicates file/resource not found or access denied.
It seems the backend requires authenticated access to retrieve the document data.
Additional Information

It looks like the share link does not properly expose or authorize access to the underlying dataset / document store when the user is not logged in.
This breaks the “public embedding” use case.

Possible Cause (speculation)

The share link may still rely on user/session-based authentication for document retrieval APIs instead of using a public token or fully scoped share permission.

Environment
RAGFlow version: (please fill in)
Deployment method: (Docker / local / cloud)
Browser: (e.g. Chrome / Edge)

Image

Rene0422:
## Summary

Fixes #15895.

Shared / embedded chat URLs work correctly for the original logged-in creator, and even chat completions work for anonymous visitors. But when an anonymous embed viewer clicks a chunk's source reference to view the original document, the PDF preview fails with "Document not found!" — exactly what the user reports.

## Root cause

The share / embed flow is built around a *beta token* the chat creator generates when they hit "Embed into webpage". That token reaches the SPA as a URL query parameter:

```
https://ragflow.example.com/chat/share/?shared_id=<conv>&auth=<beta_token>&...
```

The SPA reads it via [web/src/utils/authorization-util.ts:50-54](web/src/utils/authorization-util.ts#L50-L54):

```ts
export const getAuthorization = () => {
  const auth = getSearchValue('auth');
  const authorization = auth
    ? 'Bearer ' + auth
    : storage.getAuthorization() || '';
  return authorization;
};
```

…and attaches it as `Authorization: Bearer <token>` to every fetch/axios call. The backend's `_load_user` at [api/apps/__init__.py:172-184](api/apps/__init__.py#L172-L184) then resolves it through `APIToken.query(beta=auth_token)`, sets `current_user` to the chat creator's tenant, and the chat works.

**The PDF source-preview path is different.** When a viewer clicks a chunk citation, [web/src/components/pdf-drawer/index.tsx:26-27](web/src/components/pdf-drawer/index.tsx#L26-L27) calls `useGetDocumentUrl(documentId)` which returns:

```
${restAPIv1}/documents/<doc_id>/preview
```

That bare URL is then handed to a PDF renderer that opens it as a **browser-initiated subresource GET** — same shape as `<embed src=...>` or `<iframe src=...>`. Browser subresource requests **cannot carry an Authorization header** — they only send cookies. So:

1. [api/apps/__init__.py:153-155](api/apps/__init__.py#L153-L155): no `Authorization` header → fall through to `_load_user_from_session()` since `AUTH_JWT` is allowed on this route.
2. Anonymous visitor has no session cookie → returns `None`.
3. `current_user` is unauthenticated → [document_api.py:1932](api/apps/restful_apis/document_api.py#L1932) `DocumentService.accessible(doc_id, current_user.id)` returns False.
4. Response body: `"Document not found!"` — exactly the symptom in #15895.

The endpoint at [document_api.py:1922-1923](api/apps/restful_apis/document_api.py#L1922-L1923) **is already declared** as `auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA]`, so the auth surface is correct. The only missing piece is **a way for browser subresource requests to deliver the beta token at all**.

## Fix

Pass the beta token via URL query parameter — the same channel the share/embed design already uses for the SPA itself. Two coupled edits:

### Backend — accept `?auth=` / `?api_key=` for AUTH_BETA-enabled routes

[api/apps/__init__.py:154-163](api/apps/__init__.py#L154-L163):

```diff
  # No Authorization header, try to load user from session cookie if JWT auth is allowed
  authorization = request.headers.get("Authorization")
+ if not authorization and AUTH_BETA in auth_types:
+     # Browser-initiated requests (iframe / <embed> for PDF previews, image
+     # tags, etc.) cannot set an Authorization header. The shared chat /
+     # embed SPA already carries the beta token in the URL as ?auth=<token>
+     # (see web/src/utils/authorization-util.ts:getAuthorization); accept
+     # the same query parameter here so anonymous viewers of a shared chat
+     # can actually fetch the referenced documents. See #15895.
+     query_token = request.args.get("auth") or request.args.get("api_key")
+     if query_token:
+         authorization = f"Bearer {query_token}"
  if not authorization:
      return _load_user_from_session() if AUTH_JWT in auth_types else None
```

The existing beta-validation path at lines 172-184 then handles the synthesized `Authorization: Bearer <token>` unchanged — no change to validation logic, just an extra source for where the token comes from. Critically, this **only triggers for routes that opted into `AUTH_BETA`**; pure-JWT routes are unaffected.

### Frontend — append `?auth=<token>` to preview URLs when in embed context

[web/src/hooks/use-document-request.ts:521-535](web/src/hooks/use-document-request.ts#L521-L535):

```diff
  export const useGetDocumentUrl = (documentId?: string) => {
    const getDocumentUrl = useCallback(
      (id?: string) => {
-       return `${restAPIv1}/documents/${id || documentId}/preview`;
+       const base = `${restAPIv1}/documents/${id || documentId}/preview`;
+       // Embedded / shared chats are loaded with the beta token in the URL
+       // as ?auth=<token>. Browser-initiated GETs from <iframe>/<embed> can't
+       // attach an Authorization header, so propagate the token as a query
+       // parameter the backend now accepts for AUTH_BETA-enabled routes —
+       // otherwise anonymous viewers see "document not found" on every
+       // chunk-source preview. See #15895.
+       const auth = getSearchValue('auth');
+       if (auth) {
+         return `${base}?auth=${encodeURIComponent(auth)}`;
+       }
+       return base;
      },
      [documentId],
    );
    return getDocumentUrl;
  };
```

`getSearchValue` is the exact same helper `getAuthorization` already uses, so the detection of "are we in embed/share context?" matches the rest of the flow. Logged-in dashboard users have no `auth` query param → URL unchanged → no regression.

## Security trade-off (deliberate)

The beta token now appears in the document-preview URL's query string. That same token **is already in the share URL the visitor opened** (`?auth=<token>` — that's how the SPA picked it up in the first place); propagating it to subresource URLs doesn't expand the exposure surface. The alternative — frontend `fetch()` + `URL.createObjectURL()` — is a much bigger refactor, breaks PDF viewer libraries that expect direct URLs, and doesn't change the underlying fact that the token is already in the browser URL bar.

The existing share design accepted the URL-token trade-off when the embed feature was built; this fix stays consistent with it rather than introducing a parallel scheme.

## What this PR does NOT do

| Decision | Reason |
|---|---|
| **Doesn't change `DocumentService.accessible`** | The authorization model is correct — the chat creator's tenant grants access to the chat's KBs. The bug is solely that the request never *reached* the authorized path. Tightening access here would either over-restrict (regress logged-in users) or under-restrict (security risk). Leave it alone. |
| **Doesn't add a signed-URL scheme** | Larger design change; the existing token-in-URL pattern is consistent with how the whole share/embed flow works today. A signed-URL refactor belongs in a separate proposal, not in a #15895 fix. |
| **Doesn't touch the chat / streaming / chunk-retrieval endpoints** | Those already work for anonymous embed users because the SPA attaches the Authorization header via axios. Different code path, not affected by the iframe/embed limitation. |
| **Doesn't touch other GETs that already work** (`/thumbnails`, `/documents/images/<image_id>`, etc.) | They go through axios with the header attached. If a future bug shows the same symptom on one of those, the backend half of this PR already covers them; only the URL-builder for that asset would need the same query-param append. |
| **Doesn't propagate the token in *all* asset URL builders proactively** | Out of scope. We fix the path the user actually reported. Other builders adopt the pattern only if they're reported broken (YAGNI). |
| **Doesn't add `?auth=` to URLs when no beta token is present** | Logged-in dashboard users have no `?auth=` in their page URL → `getSearchValue('auth')` returns `null` → the existing URL shape is preserved exactly. Zero regression risk for the normal flow. |

## Test plan

- [ ] **Reproducer (pre-fix sanity check):** in a clean incognito window, open a share URL for a chat that uses a KB with at least one document. Send a question that returns a citation. Click the citation to open the PDF preview. Confirm the **pre-fix** failure: empty viewer, server log shows `Document not found!`.
- [ ] **Apply this PR**, then repeat: PDF loads in the embed viewer. Server log shows successful read against the chat creator's tenant.
- [ ] **Logged-in regression check:** in a regular browser session, open the same chat from the dashboard (not the share URL). Click a citation — PDF still loads. Confirm the URL hitting `/documents/<id>/preview` carries **no** `?auth=` query parameter (the SPA omitted it because `getSearchValue('auth')` returned `null`).
- [ ] **AUTH_JWT-only route safety:** confirm `?auth=<malformed>` against a non-AUTH_BETA route (e.g. `/documents/<id>` download at [document_api.py:1966](api/apps/restful_apis/document_api.py#L1966-L1967) which is `@login_required` only) does **not** authenticate — the new backend code is gated on `AUTH_BETA in auth_types`, so JWT-only routes ignore the query parameter.
- [ ] **Invalid token handling:** open the share URL with the `auth` query value swapped to garbage. PDF preview should fail with "Authorization is not valid!" — the new code routes the bogus value through the same `APIToken.query(beta=...)` validation that existed before, no bypass introduced.
- [ ] **Mixed source citations:** open a share chat backed by multiple KBs, with citations from documents in different KBs. Verify each preview loads correctly (the chat creator's tenant has access to all of them — this is unchanged from how chat completions already work for anonymous embed users).
- [ ] **Server log on success:** confirm the new path logs nothing that wasn't logged before — the backend change does no extra warning/error emission; it just sources the token from a different place before falling into the existing validation.
- [ ] **Backward compat:** older share URLs that didn't include `?auth=` continue to behave exactly as before (no auth → 401, expected). The fix doesn't change what's possible from a URL that didn't carry a token in the first place.

## Files changed

- [api/apps/__init__.py](api/apps/__init__.py#L154-L163) — `_load_user` reads `?auth=` / `?api_key=` from the query string as a fallback Authorization source, but only when the route opts into `AUTH_BETA`. 10 new lines, all inside the existing function.
- [web/src/hooks/use-document-request.ts](web/src/hooks/use-document-request.ts#L521-L535) — `useGetDocumentUrl` appends `?auth=<token>` when the SPA was loaded with a beta token. 13 new lines + 1 added import.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

Net diff: +23 lines across 2 files. No new tests added — the change is a small auth-channel widener rather than new functionality, and the existing beta-token validation tests cover the token's correctness. Manual verification of the share/embed flow per the test plan above is the high-confidence check.

### DIFF
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -160,6 +160,15 @@ def _load_user(auth_types=None):
         # can actually fetch the referenced documents. See #15895.
         query_token = request.args.get("auth") or request.args.get("api_key")
         if query_token:
+            # Audit trail: record that the AUTH_BETA path was reached via the
+            # URL fallback. Path is recorded for triage; the raw token is
+            # deliberately NOT logged. The token itself is still validated
+            # against APIToken.query(beta=...) below, so an invalid value
+            # produces the existing "API key is invalid" warning.
+            logging.info(
+                "auth: AUTH_BETA token sourced from URL query param path=%s",
+                request.path,
+            )
             authorization = f"Bearer {query_token}"
     if not authorization:
         return _load_user_from_session() if AUTH_JWT in auth_types else None
--- a/api/apps/__init__.py
+++ b/api/apps/__init__.py
@@ -151,6 +151,16 @@ def _load_user(auth_types=None):
     
     # No Authorization header, try to load user from session cookie if JWT auth is allowed
     authorization = request.headers.get("Authorization")
+    if not authorization and AUTH_BETA in auth_types:
+        # Browser-initiated requests (iframe / <embed> for PDF previews, image
+        # tags, etc.) cannot set an Authorization header. The shared chat /
+        # embed SPA already carries the beta token in the URL as ?auth=<token>
+        # (see web/src/utils/authorization-util.ts:getAuthorization); accept
+        # the same query parameter here so anonymous viewers of a shared chat
+        # can actually fetch the referenced documents. See #15895.
+        query_token = request.args.get("auth") or request.args.get("api_key")
+        if query_token:
+            authorization = f"Bearer {query_token}"
     if not authorization:
         return _load_user_from_session() if AUTH_JWT in auth_types else None
 
--- a/web/src/hooks/use-document-request.ts
+++ b/web/src/hooks/use-document-request.ts
@@ -26,6 +26,7 @@ import kbService, {
   uploadDocument,
 } from '@/services/knowledge-service';
 import { restAPIv1 } from '@/utils/api';
+import { getSearchValue } from '@/utils/common-util';
 import { buildChunkHighlights } from '@/utils/document-util';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { useDebounce } from 'ahooks';
@@ -520,7 +521,18 @@ export const useCreateDocument = () => {
 export const useGetDocumentUrl = (documentId?: string) => {
   const getDocumentUrl = useCallback(
     (id?: string) => {
-      return `${restAPIv1}/documents/${id || documentId}/preview`;
+      const base = `${restAPIv1}/documents/${id || documentId}/preview`;
+      // Embedded / shared chats are loaded with the beta token in the URL
+      // as ?auth=<token>. Browser-initiated GETs from <iframe>/<embed> can't
+      // attach an Authorization header, so propagate the token as a query
+      // parameter the backend now accepts for AUTH_BETA-enabled routes —
+      // otherwise anonymous viewers see "document not found" on every
+      // chunk-source preview. See #15895.
+      const auth = getSearchValue('auth');
+      if (auth) {
+        return `${base}?auth=${encodeURIComponent(auth)}`;
+      }
+      return base;
     },
     [documentId],
   );


hannzhang-EV:
I deployed it with Docker. After modifying these two files locally, what else do I need to do to make it take effect?@coderabbitai

wangq8:
We do not support ?auth=BETA for the preview API.

PR #15963, opened 2026-06-12
fix: handle None values in dfs_update for table parser config (#15962)

### LINKED ISSUE #15962: [Bug]: Table parser fails every parse with blank "Internal server error while chunking:" when dataset was created with an explicit parser_config (AssertionError in dfs_update)
### Self Checks

8f0632c8

### RAGFlow image version

infiniflow/ragflow:v0.25.6 (code inspection shows the same defect on v0.26.0 and current `main`)

```Markdown
- Doc engine: Elasticsearch
- Deployment: Docker, default compose setup
- Reproduces on any hardware/OS; the bug is pure Python logic
```

### Actual behavior

Parsing **any** CSV/Excel document with `chunk_method: "table"` fails at ~30% progress when the dataset was created through the HTTP API with an explicit `parser_config`. The document log shows a blank error:

```
Page(1~23): Start to parse.
Page(1~23): Extract records: 0~20
Page(1~23): [ERROR]Internal server error while chunking:
[ERROR][Exception]:
```

The error text is empty because the underlying exception is a bare `AssertionError` (no message). Task executor traceback:

```
  File "/ragflow/rag/svr/task_executor.py", line 302, in build_chunks
    cks = await thread_pool_exec(
  ...
  File "/ragflow/rag/app/table.py", line 519, in chunk
    KnowledgebaseService.update_parser_config(
  File "/ragflow/api/db/services/knowledgebase_service.py", line 329, in update_parser_config
    dfs_update(m.parser_config, config)
  File "/ragflow/api/db/services/knowledgebase_service.py", line 324, in dfs_update
    assert isinstance(old[k], list)
AssertionError
```

### Expected behavior

The table parser should parse the document and produce one chunk per row, regardless of whether the dataset was created with an explicit `parser_config`. If chunking does fail, the progress log should contain a meaningful error message rather than an empty string.

### Steps to reproduce

```Markdown
Pure HTTP API repro, no UI involved:

1. Create a dataset with chunk_method "table" AND any explicit parser_config:

   POST /api/v1/datasets
   {"name": "repro", "chunk_method": "table", "parser_config": {"chunk_token_num": 128}}

   Note: the response's parser_config now contains
   "table_column_names": null, "table_column_roles": null, "table_column_mode": null.
   The create path persists the full ParserConfig model dump including its
   None-default fields. (Creating the dataset WITHOUT parser_config does not
   store these keys, and parsing then works.)

2. Upload any well-formed CSV (e.g. a header row + a few data rows):

   POST /api/v1/datasets/{dataset_id}/documents

3. Parse it:

   POST /api/v1/datasets/{dataset_id}/chunks  {"document_ids": ["..."]}

4. The parse fails right after "Extract records: ..." with the blank
   "[ERROR]Internal server error while chunking:" shown above.
```

### Additional information

Root cause is a two-part interaction introduced with the table column-roles feature (#13710):

1. **Dataset create persists None defaults.** `POST /api/v1/datasets` with an explicit `parser_config` stores the full `ParserConfig` dump, including `"table_column_names": null` / `"table_column_roles": null` / `"table_column_mode": null` (the update endpoint uses `exclude_unset=True`, the create path does not).

2. **`dfs_update` asserts on type without handling stored `None`.** After extracting rows, `rag/app/table.py:519` writes the discovered columns back via `KnowledgebaseService.update_parser_config(kb_id, {"field_map": ..., "table_column_names": [...]})`. In `api/db/services/knowledgebase_service.py`, the merge does:

   ```python
   elif isinstance(v, list):
       assert isinstance(old[k], list)   # old["table_column_names"] is None -> bare AssertionError
       old[k] = list(set(old[k] + v))
   ```

**Suggested fix** (verified locally — table parsing succeeds afterwards): treat a stored `None` like a missing key in `dfs_update`:

```python
for k, v in new.items():
    if k not in old or old[k] is None:
        old[k] = v
        continue
```

This also covers the sibling `assert isinstance(old[k], dict)` for `table_column_roles` on the same code path. Complementary hardening: dump `parser_config` with `exclude_none=True` on dataset create, and raise descriptive exceptions instead of bare `assert`s so chunking failures are never blank in the UI.

zaviermeekz-cpu:

When a dataset is created via the HTTP API (or UI) with an explicit `parser_config` and `chunk_method="table"`, the stored `parser_config` contains `"table_column_names": null`, `"table_column_roles": null`, and `"table_column_mode": null`. During parsing, `KnowledgebaseService.update_parser_config` calls `dfs_update` which expected the existing value to be a list (for `table_column_names`) – but it was `None`. This caused a bare `AssertionError` and the parsing failed with a blank error message.

This PR modifies `dfs_update` to treat `None` as if the key were missing, and also guards the list merge branch to avoid type errors.

### Type of change

- [x] Bug Fix

### How has this been tested?

- Created a dataset via the API with `chunk_method="table"` and `parser_config={"chunk_token_num": 128}`.
- Uploaded a CSV file and parsed it – parsing completed successfully.
- Verified that existing datasets without `None` values continue to work.

### Related Issue

Closes #15962


JinHai-CN:
@zaviermeekz-cpu would you please resolve the conflicts?

JinHai-CN:
fixed by #15964

PR #15964, opened 2026-06-12
Fix: KnowledgebaseService.dfs_update crashes on stored None parser_config values

### LINKED ISSUE #15962: [Bug]: Table parser fails every parse with blank "Internal server error while chunking:" when dataset was created with an explicit parser_config (AssertionError in dfs_update)

BaptisteJosh:

Fixes #15962.

`KnowledgebaseService.update_parser_config` → `dfs_update` uses bare `assert isinstance(...)` checks when merging new `parser_config` values into the stored config. When a stored value is `None`, the assert raises a message-less `AssertionError`, so the failure surfaces in the document log as a blank `[ERROR]Internal server error while chunking:`.

This is hit on every table-parser parse for datasets created via `POST /api/v1/datasets` with an explicit `parser_config`: the create path persists the `ParserConfig` model's `None` defaults (`table_column_names`, `table_column_roles`, `table_column_mode`, ...), and `rag/app/table.py` then writes the discovered column list back through `update_parser_config`, where the merge of `list` into stored `None` crashes. Full analysis, reproduction steps, and traceback are in #15962.

The codebase has two copies of `dfs_update`. #12815 already fixed the `DocumentService` copy with safe `isinstance` conditions (mismatches fall through to overwrite); the `KnowledgebaseService` copy was never updated. This PR applies the identical pattern there:

```python
# before
if isinstance(v, dict):
    assert isinstance(old[k], dict)
    dfs_update(old[k], v)
elif isinstance(v, list):
    assert isinstance(old[k], list)
    old[k] = list(set(old[k] + v))

# after (same as DocumentService.dfs_update since #12815)
if isinstance(v, dict) and isinstance(old[k], dict):
    dfs_update(old[k], v)
elif isinstance(v, list) and isinstance(old[k], list):
    old[k] = list(set(old[k] + v))
```

Verified against v0.25.6: with this change applied, the reproduction CSV from #15962 parses successfully (one chunk per row); without it, the parse fails as described.

### Type of change


BaptisteJosh:
Re the CodeRabbit findings, for the human reviewer's convenience:

- **Scope is intentionally minimal.** This PR only re-applies the guard pattern from #12815 to the `KnowledgebaseService` copy of `dfs_update`, which is the code path that crashes (see #15962 for the traceback and HTTP-API reproduction). 
- **The list-merge semantic difference is pre-existing and deliberately untouched.** `KnowledgebaseService.dfs_update` set-unions lists while `DocumentService.dfs_update` overwrites them; changing either would alter behavior for existing callers (e.g. merged `table_column_names` accumulation) and is beyond a crash fix.
- **The `update_parser_config` control-flow differences (early return, raptor key cleanup) are likewise pre-existing** and unrelated to the AssertionError.

Validation: reproduced on v0.25.6 per the steps in #15962 — without this change, parsing any CSV with `chunk_method: "table"` on a dataset created via the HTTP API with an explicit `parser_config` fails with a blank "Internal server error while chunking:"; with the change applied, the same document parses to one chunk per row.

Lynn-Inf (approved):
**LGTM** — the change is correct and I agree with dropping the bare `assert`. The original intent is reasonable (guard the shape invariant before merging dict-into-dict / list-into-list), but `assert` is the wrong tool here: it's stripped under
  `python -O`, it conflates a legitimate "unset" (`None`, from `ParserConfig` defaults being persisted) with actual corruption, and a message-less `AssertionError` is useless for diagnosis. The fall-through to `old[k] = v` is also semantically
  consistent with the `if k not in old` branch right above it. Aligning this copy with `DocumentService.dfs_update` (already using the `and` pattern since #12815) removes a real divergence between the two helpers — and this PR keeps the list-union
  behaviour that the `DocumentService` copy doesn't even have, so it's the conservative direction.

  **Non-blocking:** the bare `and` makes a genuine shape mismatch silent. If you want to keep the original intent of surfacing schema drift without crashing, consider logging a warning before overwriting, e.g. `logging.warning("parser_config key %r:
  expected list, found %s; overwriting", k, type(old[k]).__name__)` in the mismatch branches. That preserves operator visibility while still self-healing.

  **Unrelated observation (out of scope here, worth a separate issue):** `old[k] = list(set(old[k] + v))` is unsafe for order-sensitive list values like `table_column_names` — `set` drops ordering and duplicates, and since Python randomises string
  hashing per process, the resulting order isn't even stable across restarts. That's pre-existing behaviour this PR doesn't touch, but it looks like a latent bug for table parsing.

PR #15973, opened 2026-06-12
feat(models): add DLA and TSR methods to ModelDriver interface

xugangqiang:
Add DLA() and TSR() methods to ModelDriver for DeepDoc integration.


xugangqiang:
Closing — DLA/TSR don't need to be on ModelDriver interface. DeepDocModel can implement a separate DocAnalyzer interface and consumers type-assert. No interface expansion needed.

PR #15982, opened 2026-06-12
perf(graphrag): batch entity/relation embeddings in set_graph

### LINKED ISSUE #15921: [Performance] GraphRAG set_graph encodes entity/relation chunks one by one and creates all embedding tasks up front
## Summary

`rag/graphrag/utils.py::set_graph()` currently converts graph changes into entity and relation chunks by creating one asyncio task per changed node/edge. Each task then calls the embedding model with a single-item input.

This makes GraphRAG indexing much slower than necessary on large graphs, even when the embedding backend supports batch inputs. It also creates many in-memory tasks before any entity/relation chunks are persisted.

## Current behavior

In `rag/graphrag/utils.py`:

- `graph_node_to_chunk(...)` builds one entity chunk and calls `embd_mdl.encode([ent_name])` for that single entity.
- `graph_edge_to_chunk(...)` builds one relation chunk and calls `embd_mdl.encode([relation_text])` for that single relation.
- `set_graph(...)` loops over `change.added_updated_nodes` and `change.added_updated_edges`, creates one `asyncio.create_task(...)` per item, and then waits on `asyncio.gather(...)`.
- `chat_limiter` limits concurrent embedding calls, but it does not limit the number of task objects created up front.

Relevant code paths:

- `rag/graphrag/utils.py::graph_node_to_chunk`
- `rag/graphrag/utils.py::graph_edge_to_chunk`
- `rag/graphrag/utils.py::set_graph`

## Why this is a problem

For a graph update with many changed entities and relationships, the current design has two scaling issues:

1. **Single-item embedding calls**
   - The number of embedding model calls grows linearly with the number of changed nodes and edges.
   - Batch-capable embedding services cannot be used efficiently.
   - Much of the runtime can be spent on request overhead rather than model throughput.

2. **Unbounded task creation**
   - `asyncio.gather` receives a task for every changed node/edge.
   - The limiter only gates the inner model call, not the total number of tasks or intermediate chunk objects.
   - If the process fails before insertion, all completed in-memory embedding work has to be repeated.

## Expected behavior

GraphRAG should process entity/relation embeddings in bounded batches, for example:

- collect pending entity/relation texts first;
- batch-read embedding cache entries;
- batch-call `embd_mdl.encode([...])` for cache misses;
- build chunks after embeddings are available;
- use bounded batch loops or worker queues instead of creating all tasks at once;
- optionally persist progress per batch so retry does not start from zero.

## Suggested direction

A possible implementation direction:

1. Split chunk preparation into two phases:
   - prepare entity/relation payloads without vectors;
   - batch-fill vectors.
2. Add a batch embedding helper for GraphRAG entity/relation chunks.
3. Replace per-item `asyncio.create_task(...)` with bounded batch processing.
4. Make chunk collection return values explicit instead of mutating a shared `chunks` list from many coroutines.

## Additional context

This issue is independent from general LLM extraction cost. Even after entity and relation extraction has completed, `set_graph()` can still become a large write-side bottleneck because it embeds every updated entity and relation one by one.

jeffrey701:

`set_graph()` converts each changed entity and relation into a chunk by creating one `asyncio` task per item, and each task calls `embd_mdl.encode([single_text])`. So embedding-model calls grow linearly with the number of changed nodes/edges, batch-capable embedding backends can't be used, and every chunk + task is materialized up front before any insert. On large graph updates this makes the write side of GraphRAG indexing a bottleneck even after extraction is done.

Split chunk building from embedding (the direction suggested in the issue):

- `build_node_chunk` / `build_edge_chunk` build the chunk and return `(chunk, cache_key, embed_text)` without a vector.
- `embed_graph_chunks` fills `q_<dim>_vec` in bounded batches (`GRAPH_EMBED_BATCH_SIZE`, default 16) — it checks the embed cache first and issues one `encode()` per group of cache misses instead of one per item.
- `set_graph` builds all node/edge chunks, then batch-embeds, replacing the per-item `asyncio.create_task` fan-out.

Chunk content, cache keys, and the build-before-delete ordering are unchanged, so a failure during embedding still leaves the old graph/subgraph checkpoints intact. Unit tests updated to the new builder API, with added coverage that `encode()` is batched and the cache short-circuits misses.

Closes #15921

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [x] Performance Improvement

### DIFF
--- a/rag/graphrag/utils.py
+++ b/rag/graphrag/utils.py
@@ -460,7 +460,6 @@ async def embed_graph_chunks(embd_mdl, pending, callback=None, label="", batch_s
     so a batch-capable backend makes one ``encode`` call per group instead of one
     per node/edge. Mutates the chunk dicts in place.
     """
-    global chat_limiter
     enable_timeout_assertion = os.environ.get("ENABLE_TIMEOUT_ASSERTION")
 
     misses = []
--- a/rag/graphrag/utils.py
+++ b/rag/graphrag/utils.py
@@ -41,6 +41,11 @@ ErrorHandlerFn = Callable[[BaseException | None, str | None, dict | None], None]
 
 chat_limiter = LoopLocalSemaphore(int(os.environ.get("MAX_CONCURRENT_CHATS", 10)))
 
+# GraphRAG entity/relation embeddings are filled in batches of this size so a
+# batch-capable embedding backend makes one encode() call per group instead of
+# one call per node/edge. Override with GRAPH_EMBED_BATCH_SIZE.
+GRAPH_EMBED_BATCH_SIZE = int(os.environ.get("GRAPH_EMBED_BATCH_SIZE", 16))
+
 # Doc-store insert batching for GraphRAG subgraph/node/edge/community_report
 # chunks.  Defaults (64 docs per batch, up to 4 batches in flight) mirror the
 # regular ingest pipeline in document_service.py while still keeping the total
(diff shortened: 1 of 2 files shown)


yingfeng:
Graphrag will be obsolete in 0.27.0 and will be replaced by knowledge compilation, as a result, these adjustments will not be merged any more.

PR #15994, opened 2026-06-14
fix(llm): implement async chat methods for GoogleChat (Vertex AI Gemini)

### LINKED ISSUE #15992: GoogleChat (Vertex AI) fails with 401 invalid_api_key on async chat paths since 0.25.6
# GoogleChat (Vertex AI) fails with 401 invalid_api_key on async chat paths since 0.25.6

## Description

`GoogleChat` (`rag/llm/chat_model.py`, `_FACTORY_NAME = "Google Cloud"`) fails with an OpenAI `401 invalid_api_key` error whenever the **async** chat code paths are used (`_async_chat`, `_async_chat_streamly`, inherited from `Base`).

## Root Cause

`Base.__init__` always sets:

```python
self.async_client = AsyncOpenAI(api_key=key, base_url=base_url, timeout=timeout)
```

For `GoogleChat`, `key` is the raw JSON string containing the Vertex AI service account credentials (`{"google_service_account_key": "...", "google_project_id": "...", "google_region": "..."}`), not a valid OpenAI API key.

`GoogleChat.__init__` correctly overrides `self.client` with a `genai.Client(vertexai=True, ...)` instance (sync Vertex AI SDK), and `chat_streamly` correctly uses `self.client.models.generate_content_stream(...)`.

However, `GoogleChat` does **not** override `self.async_client`, and does **not** implement its own `_async_chat` / `_async_chat_streamly`. These are inherited from `Base`, which call:

```python
response = await self.async_client.chat.completions.create(...)
```

This sends the raw Vertex AI service-account JSON as an OpenAI `api_key`, resulting in:

```
openai.AuthenticationError: Error code: 401 - {'error': {'message': 'Incorrect API key provided: {"google...'..., 'code': 'invalid_api_key'}}
```

## Why this surfaced in 0.25.6

Some chat/agent code paths that previously used the sync `chat_streamly` (which works correctly) now go through the async `async_chat` / `async_chat_streamly` methods (inherited `_async_chat` / `_async_chat_streamly` in `Base`). For `GoogleChat`, these async paths were never implemented and silently fall through to the broken `AsyncOpenAI` client.

## Suggested Fix

`GoogleChat` should implement its own `_async_chat` and `_async_chat_streamly` using `self.client.aio.models.generate_content` / `generate_content_stream` (the async namespace of the same `genai.Client`), mirroring the existing sync `_chat` / `chat_streamly` implementations (Content/GenerateContentConfig/ThinkingConfig construction).

## Environment

- RAGFlow version: v0.25.6
- LLM factory: Google Cloud (Vertex AI), model: `gemini-2.5-flash`
- Deployment: Docker, `infiniflow/ragflow:v0.25.6` image

glu000:
Fixes #15992

GoogleChat overrides `self.client` with a `genai.Client` (Vertex AI SDK) but never overrides `self.async_client`, which `Base` initializes as `AsyncOpenAI(api_key=key, ...)` where `key` is the raw Vertex AI service account JSON. When the inherited async `_async_chat` / `_async_chat_streamly` methods are used (async chat/streaming paths), this sends the service account JSON as an OpenAI API key, failing with:

```
openai.AuthenticationError: Error code: 401 - {'error': {'message': 'Incorrect API key provided: {"google...'..., 'code': 'invalid_api_key'}}
```

This PR adds `GoogleChat`-specific `_async_chat` and `_async_chat_streamly` for Gemini models, using `self.client.aio.models.generate_content` / `generate_content_stream` (the async namespace of the same `genai.Client`), mirroring the existing sync `_chat` / `chat_streamly` implementations. The Claude/AnthropicVertex branch falls back to the inherited `Base` implementation (unchanged behavior).

## Test plan
- [x] Reproduced the 401 error on v0.25.6 with `gemini-2.5-flash@Google Cloud` via async chat completion endpoint
- [x] Applied this fix and confirmed async chat completions now succeed end-to-end


Lynn-Inf:
Thanks for the PR! The approach of overriding the async methods is correct. Here are a few suggestions:

**Why overriding `self.async_client` assignment doesn't work**
Simply reassigning `self.async_client` to a `genai.Client` won't help because `Base._async_chat` calls `self.async_client.chat.completions.create(...) `(OpenAI interface), which `genai.Client` doesn't have. The async APIs are entirely different across OpenAI, google-genai, and AnthropicVertex — not just the client object, but the request/response structures too (e.g. role mapping, config format, ThinkingConfig). So method-level overrides are necessary.

 **Suggestions**
1. **Extract shared helpers** — The config building and history conversion logic is duplicated across `_chat`, `chat_streamly`, `_async_chat`, and `_async_chat_streamly`. Extracting helper methods like `_build_genai_config` and `_convert_history_to_contents` would significantly reduce duplication and maintenance burden.

2. **Set `self.async_client = None` explicitly** — Since `GoogleChat` never uses the `AsyncOpenAI` instance created by `Base.__init__`, it's worth explicitly setting `self.async_client = None` after `super().__init__()`. This ensures that accidental calls to the inherited async methods fail with a clear `AttributeError` rather than a confusing 401.

3. **Claude branch async path may also be broken** — The async methods delegate to `Base` for Claude models, but `self.async_client` is still an `AsyncOpenAI` with the service account JSON as the key, which would also result in a 401. This should either be fixed (using `AsyncAnthropicVertex`) or documented as a known limitation.

Overall the PR correctly solves the Gemini async issue. The suggestions above are about reducing duplication and improving robustness.

Lynn-Inf (approved):
While the suggested optimizations haven't been fully addressed, the PR itself is still valuable. I think it's ready to merge once CI passes.

PR #15991, opened 2026-06-13
Feat: Add New API model provider for OpenAI-compatible gateways

wanghualoong:
## Summary

Add support for **"New API"** as a model provider, enabling connection to [New API](https://github.com/QuantumNous/new-api) / [one-api](https://github.com/songquanpeng/one-api) compatible gateways that aggregate multiple LLM backends behind a unified OpenAI-compatible `/v1` endpoint.

### Features

- **All model types**: Chat, Embedding, Rerank, Image2Text, TTS, Speech2Text
- **List Models discovery**: `NewAPI(OpenAIAPICompatible)` class in `model_meta.py` queries the gateway's `/v1/models` to auto-discover available models via the native `GET /api/v1/providers/<name>/models` endpoint
- **Model parameter editing**: Pencil icon on each discovered model row to edit `model_type`, `max_tokens`, and `features` (e.g. tool call support) before submitting
- **Custom model addition**: "Add Custom Model" button at the bottom of the List Models dropdown for models not returned by the API
- **Gear icon settings**: Enabled the Settings gear button on provider instances to manage models on existing instances (viewMode)
- **viewMode credential passthrough**: Fixed List Models in viewMode — merges `initialValues` credentials when `api_key`/`base_url` fields are hidden by `hideWhenInstanceExists`

### Changes

**Backend** (8 files):
- `rag/llm/chat_model.py` — `NewAPIChat(Base)` class
- `rag/llm/embedding_model.py` — `NewAPIEmbed(OpenAIEmbed)` class (no auto `/v1` append)
- `rag/llm/rerank_model.py` — `NewAPIRerank(Base)` class (uses `/rerank` endpoint)
- `rag/llm/cv_model.py` — `NewAPICv(GptV4)` class
- `rag/llm/tts_model.py` — `NewAPITTS(OpenAITTS)` class
- `rag/llm/sequence2txt_model.py` — `NewAPISeq2txt(GPTSeq2txt)` class
- `rag/llm/model_meta.py` — `NewAPI(OpenAIAPICompatible)` class for List Models discovery
- `conf/llm_factories.json` — New API factory entry with all model type tags

**Frontend** (8 files + 1 new SVG):
- `web/src/assets/svg/llm/new-api.svg` — New API logo icon
- `web/src/constants/llm.ts` — `LLMFactory.NewAPI` enum + `IconMap` entry
- `web/src/components/svg-icon.tsx` — `NewAPI` added to `svgIcons`
[...]

PR #15996, opened 2026-06-14
Fix: return api_key/base_url in provider instance APIs and make add_model idempotent

wanghualoong:
## Summary

- `show_provider_instance` and `list_provider_instances` now include `base_url` (from `extra.base_url`) in their response dicts — but **no longer expose `api_key`**
- `list_provider_models` accepts an optional `instance_name` query parameter: when provided, the backend looks up stored `api_key` and `base_url` from the database server-side, so the frontend never needs to receive or transmit raw credentials
- `add_model_to_instance` returns `True, "skipped"` instead of `False, "Model already exists..."` when a model already exists, making batch model submission from the List Models picker idempotent

## Why

The unified model provider system (v0.26.0) introduced a gear icon on each provider instance that opens the ProviderModal in viewMode. In viewMode, the `api_key` and `base_url` form fields are hidden (`hideWhenInstanceExists`).

Previously, the frontend needed the raw `api_key` from the instance detail API to call List Models — this exposed credentials in the API response and browser network traffic. Now, the frontend passes `instance_name` and the backend resolves credentials server-side.

The idempotent `add_model` fix is needed because the List Models picker pre-checks existing models. When the user clicks OK, all selected models (including pre-checked ones) are submitted. The old behavior would error on the first existing model, blocking the addition of any new models in the same batch.

## Related

- Companion PR: #15991 (New API model provider — passes `instance_name` in viewMode)

## Test plan

- [ ] Open Model Provider settings, click gear icon on any provider instance
- [ ] Click "List Models" — should fetch and display available models (credentials resolved server-side)
- [ ] Verify browser Network tab: `show_provider_instance` response does NOT contain `api_key`
- [ ] Select models including ones already added, click OK — should succeed without errors
- [ ] Verify existing models remain unchanged, new models are added
- [ ] New instance creation (non-viewMode) still works: api_key/base_url from form are used


wanghualoong:
All coderabbit review findings have been addressed:

1. **Credential exposure** (Major): Fixed in de9ebef — `api_key` removed from both `list_provider_instances` and `show_provider_instance` responses. Only `base_url` (non-sensitive) is returned. Server-side credential lookup via `instance_name` added so the frontend never needs to send `api_key`.

2. **Silent fallback on instance lookup failure** (Major): Fixed in 150916f — when `instance_name` is provided but provider/instance doesn't exist, returns an explicit error (`"Provider 'X' not found for this tenant"` / `"Instance 'X' not found for provider 'Y'"`) instead of silently continuing with empty credentials.

3. **Swagger docs for `instance_name`** (Nitpick): Fixed in 40ccb0c — added `instance_name`, `api_key`, and `base_url` query parameter documentation to the `list_provider_models` swagger docstring.

4. **Debug logging for skipped duplicate models** (Nitpick): Fixed in 717d5d8 — added `logging.debug()` when `add_model` skips an already-existing model.

JinHai-CN:
@wanghualoong would you please resolve the conflicts?

PR #16012, opened 2026-06-15
fix: return "search not found" instead of IndexError when updating a missing search

### LINKED ISSUE #16011: [Bug]: PUT /api/v1/searches/<id> raises IndexError (HTTP 500) instead of returning "search not found"
### Self Checks

400dfd50d

### RAGFlow image version

nightly / `main` (v0.26.0)

```Markdown
Code-level defect, environment-independent (reproducible on main by reading the request path).
```

### Actual behavior

`PUT /api/v1/searches/<search_id>` (the `update` handler in `api/apps/restful_apis/search_api.py`) can return **HTTP 500** with an unhandled `IndexError: list index out of range` instead of a graceful "search not found" response.

The handler fetches the search like this:

```python
# api/apps/restful_apis/search_api.py
search_app = SearchService.query(tenant_id=current_user.id, id=search_id)[0]
if not search_app:
    return get_json_result(data=False, message=f"Cannot find search {search_id}", code=RetCode.DATA_ERROR)
```

`SearchService.query(...)` (via `CommonService.query`) returns a **list**. When no row matches, it returns `[]`, so `[0]` raises `IndexError` **before** the `if not search_app` guard can run. That guard is therefore dead code, and the intended `DATA_ERROR` response is unreachable.

This is reachable in practice because the authorization check on the line above uses a **different column** than the fetch:

- `SearchService.accessible4deletion(search_id, current_user.id)` matches on `created_by == user_id`.
- `SearchService.query(tenant_id=current_user.id, id=search_id)` matches on `tenant_id == current_user.id`.

So when a user created a search inside a **team workspace owned by another tenant** (`created_by` = the caller, `tenant_id` = the team owner), the request passes `accessible4deletion` but `query(tenant_id=current_user.id, ...)` returns `[]` → `IndexError` → HTTP 500. (The `GET /searches/<id>` detail handler confirms such team-shared searches exist: it iterates over every team the user belongs to and matches by `tenant_id=tenant.tenant_id`.)

The sibling handlers avoid this: `detail` and `completion` use `SearchService.get_detail(...)` which returns `{}` and is guarded with `if not search_app`; only `update` uses `[...][0]` indexing.

### Expected behavior

When the search is not found under the caller's tenant, `update` should return the intended graceful error (`code=DATA_ERROR`, `Cannot find search <id>`), not raise `IndexError` / HTTP 500.

### Steps to reproduce

```Markdown
1. As a team member, create a Search app inside a team workspace owned by another tenant
   (so the row's created_by = your user id, tenant_id = the team owner's id).
2. Send: PUT /api/v1/searches/<that_search_id> with a valid body, e.g.
   {"name": "renamed", "search_config": {}}
3. Observe: the server raises IndexError (HTTP 500) instead of returning a
   "Cannot find search" DATA_ERROR response.
```

Equivalent unit-level reproduction: stub `SearchService.accessible4deletion -> True` and `SearchService.query -> []`, then call `update(search_id=...)` — it raises `IndexError` today.

### Additional information

Root cause: list indexing (`[0]`) is evaluated before the `if not search_app` empty check. Fix: fetch the list first, return `DATA_ERROR` if empty, then index — mirroring the guarded pattern already used by the other handlers in this file.

Crystora:

`PUT /api/v1/searches/<search_id>` (the `update` handler in `api/apps/restful_apis/search_api.py`) could return **HTTP 500** with an unhandled `IndexError: list index out of range` instead of a graceful "search not found" response.

**Root cause.** `SearchService.query(...)` returns a list. The handler indexed `[0]` *before* the empty-result check, so the `if not search_app` guard was dead code:

```python
search_app = SearchService.query(tenant_id=current_user.id, id=search_id)[0]
if not search_app:  # unreachable — [0] already raised IndexError on []
    return get_json_result(..., code=RetCode.DATA_ERROR)
```

This is reachable because the authorization check uses a different column than the fetch: `accessible4deletion` matches on `created_by`, while `query(...)` matches on `tenant_id`. A user who created a search inside a team workspace owned by another tenant passes `accessible4deletion` but the `tenant_id`-scoped `query` returns `[]` → `IndexError` → HTTP 500.

**Fix.** Fetch the list first, return the intended `DATA_ERROR` when it is empty, then index — mirroring the guarded pattern already used by the `detail`/`completion` handlers in the same file. This is the same fix pattern applied to the analogous `SearchService.get_detail` crash in #15622 (merged).

Closes #16011

### Type of change

### How has this been tested?

Added a regression unit test `test/unit_test/api/apps/restful_apis/test_update_search_missing.py` (following the existing `test_get_agent_session.py` harness):

- `test_returns_error_when_search_missing` — asserts a `DATA_ERROR` response (not a crash) when the lookup returns `[]`. Fails on the previous code, passes with this fix.
- `test_updates_when_search_found` — asserts the normal update path still succeeds.

`ruff check` passes on the changed files.


Crystora:
@JinHai-CN @Haruko386 could you pls review this pr? Simple bug fixing pr

euvre:
The handler fix itself is correct and verified end-to-end, but the PR ships its new regression test in a state where it **fails at HEAD** — the "How has this been tested" claim no longer holds on this branch.

### Findings

1. **[major] Regression test fails to run at current HEAD — stale module stubs after the merge from `main`**
   `test/unit_test/api/apps/restful_apis/test_update_search_missing.py:91` (and the stub block around line 69)
   Running `pytest test/unit_test/api/apps/restful_apis/test_update_search_missing.py` at this branch's HEAD fails both tests at module load:
   - `ModuleNotFoundError: No module named 'api.db.services.knowledgebase_service'; 'api.db.services' is not a package` — the harness stubs `api.db.services` as a plain module but provides no stub for `api.db.services.knowledgebase_service`, which `api/apps/restful_apis/search_api.py:28` now imports.
   - Even after that is fixed, `ImportError: cannot import name 'DEFAULT_PAGE' from 'api.utils.pagination_utils'` — the stub only provides `validate_rest_api_page_size`, while `api/apps/restful_apis/search_api.py:34` imports five names from that module.
   Both imports were added on `main` after this branch was authored (the June version of `search_api.py` had neither), so the test passed when written but the recent merge into the branch broke it without the stubs being updated. The repo's unit lane (`run_tests.py` over `test/unit_test`) will go red on this file.
   Suggested direction (verified to work): mirror `test_get_agent_session.py` and add `_stub(monkeypatch, "api.db.services.knowledgebase_service", KnowledgebaseService=SimpleNamespace())`, and extend the `api.utils.pagination_utils` stub with `DEFAULT_PAGE`, `DEFAULT_PAGE_SIZE`, `validate_rest_api_ids`, `validate_rest_api_page`. With exactly those two repairs the file passes 2/2 against this PR's handler, and still fails against `main`'s handler with `{'code': 500, 'message': 'list index out of range'}` — i.e. the assertions are sound, only the stub set is stale.

2. **[minor] The underlying column mismatch between authorization and fetch remains**
   `api/apps/restful_apis/search_api.py:143` vs `:147` — `accessible4deletion` matches on `created_by` while the fetch matches on `tenant_id`, so a user who created a search inside another tenant's workspace now gets `Cannot find search <id>` (DATA_ERROR) for a row that does exist. This is graceful and consistent with the guarded `detail`/`completion` handlers and the analogous earlier fix, so it is fine for this PR's scope — but the message is slightly misleading for that case. Consider aligning the lookup with `detail()`'s tenant iteration or a distinct message in a follow-up. Non-blocking.

### What was verified as OK

- **The fix is correct and minimal**: the list is fetched first, the empty case returns the intended `DATA_ERROR` (`Cannot find search <id>`) with a warning log (which also resolves the earlier review comment), and only then is `[0]` indexed — mirroring the sibling handlers. No other behavior in the handler changed. One precision note: the old code did not leave the exception fully unhandled — the handler's `except` turned it into a 500 server-error response; the observable defect (500 + `list index out of range` instead of the graceful error) is exactly as reported in #16011.
- **Live end-to-end against a server running this branch**: session login via the app's own login protocol, then `GET /api/v1/searches` (list, code 0), `POST /api/v1/searches` (created `audit-pr16012-search`, code 0), `GET /api/v1/searches/<id>` (detail, code 0), `PUT /api/v1/searches/<id>` rename (code 0, updated record returned) — the normal update path is intact; and `PUT /api/v1/searches/<nonexistent>` returns the authentication error `no authorization`, confirming the authorization gate still runs before the lookup (a nonexistent id cannot reach the fixed branch on either old or new code).
- **Differential proof of the fixed branch** (the team-workspace state from #16011 is not constructible through the public API, so the branch was exercised at unit tier): with `main`'s handler the regression test fails with `{'code': 500, 'message': 'list index out of range'}`; with this PR's handler it passes — the fix changes precisely the reported behavior.
- **Adjacent regression**: the sibling harness `test_get_agent_session.py` passes unchanged, and the `list_searches`/`detail` endpoints were exercised live without issue.
- Boundary note: the web UI itself could not be walked in the audit environment — the frontend at this merge crashes at boot before login (a request returns 500; unrelated to this PR, which touches no frontend or boot code, and the backend API responded normally throughout). Since the change is REST-handler-only, verification was done at the API level; screenshots of the observed states are captured in the audit evidence.

JinHai-CN:
duplicated with https://github.com/infiniflow/ragflow/pull/16027

PR #16023, opened 2026-06-15
Fix one data source can be synced to multiple dataset

wangq8:
Fix one data source can be synced to multiple dataset
Test add/delete - worked.

### DIFF
--- a/docker/launch_data_sync_service.sh
+++ b/docker/launch_data_sync_service.sh
@@ -67,79 +67,28 @@ cleanup() {
 # Trap SIGINT and SIGTERM to invoke cleanup
 trap cleanup SIGINT SIGTERM
 
-# Function to execute task_executor with retry logic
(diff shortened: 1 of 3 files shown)


docker/launch_data_sync_service.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
wangq8, line 1: ignore this file.

docker/launch_backend_service.sh
```diff
@@ -14,7 +14,7 @@ load_env_file() {
         echo "Loading environment variables from: $env_file"
         # Source the .env file
         set -a
-        source "$env_file" 
+        source "$env_file"
```
wangq8, line 17: ignore this.

buua436 (approved):

PR #16027, opened 2026-06-15
fix: prevent IndexError in PUT /api/v1/searches/<id> when search not found

### LINKED ISSUE #16011: [Bug]: PUT /api/v1/searches/<id> raises IndexError (HTTP 500) instead of returning "search not found"

MKP999:
## Problem

When `PUT /api/v1/searches/<id>` is called with a non-existent search ID, `SearchService.query()` returns an empty list. The current code indexes `[0]` directly, which raises an `IndexError` before the guard condition can execute:

```python
search_app = SearchService.query(tenant_id=current_user.id, id=search_id)[0]  # IndexError!
if not search_app:  # dead code
    ...
```

This returns HTTP 500 instead of a proper error response.

## Fix

Check the query result list for emptiness before indexing:

```python
search_results = SearchService.query(tenant_id=current_user.id, id=search_id)
if not search_results:
    return get_json_result(data=False, message=f"Cannot find search {search_id}", code=RetCode.DATA_ERROR)
search_app = search_results[0]
```

This is consistent with the pattern used in PR #15622 which fixed an analogous bug in `SearchService.get_detail()`.

Closes #16011


euvre:
LGTM 🚀

The change is a correct, minimal fix: in `update()` the code indexed `SearchService.query(...)[0]` before checking emptiness, so an empty result raised `IndexError` (caught by the outer `except`, surfacing to the client as an unhandled-exception envelope). The fix checks the list first and returns the intended `RetCode.DATA_ERROR` envelope, exactly matching the "Cannot find search" branch that was previously dead code, and matching the same message/pattern already used in `completion()` (`api/apps/restful_apis/search_api.py:216`).

Note on reachability: a fully unknown ID is already rejected one step earlier by `accessible4deletion()` (`api/apps/restful_apis/search_api.py:143`, checks `created_by` + VALID status), which is why `test_search_update_invalid_search_id` asserts code 109. The crashed path triggers when `created_by == current_user.id` but the tenant-scoped query is empty (tenant_id/created_by divergence, or a delete race between the two queries). I reproduced exactly that state and verified the fix there.

What was verified (live RAGFlow stack: es/mysql/redis/minio/nats + python api, running this PR's code):

- Fixed path: created a search app, then made its row diverge (`created_by` = caller, `tenant_id` = other) so the guard passes but the fixed query returns empty. `PUT /api/v1/searches/<id>` now returns `{code: 102, message: "Cannot find search <id>"}` instead of an exception envelope.
- Causality check: temporarily reverted the 3-line fix in a scratch run and replayed the identical request — it returned `{code: 100, message: "IndexError('list index out of range')"}` (raw exception leaked to the client), confirming both the bug and that the PR is what fixes it. Worktree was restored to PR code afterwards (clean diff).
- Regression paths, all unchanged/OK: valid rename + `search_config` merge → code 0 with updated name; unknown search ID → code 109 "no authorization" (pre-existing guard intact); rename to another VALID search's name → code 102 "Duplicated search name."; rename to own name (case-insensitive) → code 0.
- Code review of the surrounding route: authorization still runs before any data lookup (no new info leak), no state changes happen before the new early return, no concurrency/error-handling changes beyond the guard, and the response contract for all previously-reachable paths is byte-identical.

Minor, non-blocking nits:

1. (minor) No automated test pins the new branch. The existing `test_search_update_invalid_search_id` exercises the earlier authorization guard and never reaches the changed lines, and the fixed branch is not reachable through the HTTP API with normal data. A small unit test stubbing `SearchService.query` to return `[]` (the file's established stub-module style exists under `test/testcases/restful_api/*_unit.py`) would protect it.
2. (nit) The not-found branch returns silently; a `logging.warning(...)` before the return would make such requests traceable, consistent with the `logging.debug` usage nearby (also suggested by the bot review).
3. (nit, description only) The PR body says the old behavior returned "HTTP 500"; in fact the envelope is HTTP 200 with `code: 100` and the raw `IndexError` repr in `message` (verified live). The substance of the claim is correct.

Boundary note: the repo's contract suite for this endpoint (`test/testcases/restful_api/test_searches.py`) aborts at its session-scoped fixture because it provisions a real ZHIPU-AI LLM key, which is unavailable in this environment; the same request/response contracts listed above were therefore verified directly against the live API instead. That fixture limitation is unrelated to this PR.

euvre:
The runtime fix itself is correct and was verified live end to end, but the PR breaks one existing unit test that pins exactly this branch. Details below.

## Findings

1. **[major] Existing unit test breaks: `test/testcases/test_web_api/test_search_app/test_search_routes_unit.py:364-370`**

   `test_update_and_detail_route_matrix_unit` stubs `SearchService.query` to return `[None]` and asserts the update route answers `RetCode.DATA_ERROR` (102) with "Cannot find search". That stub was written against the old implementation, where `query(...)[0]` produced `None` and fell into the (then live) `if not search_app` guard. With this PR, a non-empty list `[None]` passes the new emptiness check at `api/apps/restful_apis/search_api.py:148`, `search_app` becomes `None`, and the very next line (`search_app.name.lower()`, `search_api.py:152`) raises `AttributeError`, which the outer `except` converts to `code 100` (EXCEPTION_ERROR).

   Verified by A/B run:
   - with the PR applied: the test fails with `AssertionError: assert 100 == 102` (full file: 1 failed, 2 passed);
   - with only `api/apps/restful_apis/search_api.py` reverted to `main` (no other change): the test passes.

   Suggested direction (author's call): update the stub from `[None]` to `[]` — and the adjacent comment "query returns [None]" — so the test keeps pinning the "Cannot find search" contract under the new implementation. Note that `query()` in production returns model rows or an empty list, never `[None]`, so this is a test-contract update, not a production-behavior problem.

   Context, not counted against this PR: the `test_web_api` suite is not referenced in the CI workflows (they run `test/testcases/test_sdk_api` and `test/testcases/restful_api`), and that test file's module loader only works when `api.db.services.knowledgebase_service` / `api.utils.pagination_utils` are already imported in the session — both pre-existing conditions.

## What was verified as OK (live stack running this PR's code)

- **Fixed path**: created a search app, then made its row diverge (`created_by` = caller, `tenant_id` = a different tenant) so the earlier `accessible4deletion` guard passes but the tenant-scoped query returns empty. `PUT /api/v1/searches/<id>` now returns `{"code": 102, "message": "Cannot find search <id>"}` — the intended envelope — instead of the previous unhandled-exception envelope with the raw `IndexError` in `message`. Restoring `tenant_id` makes the same PUT succeed with `code 0` again, confirming no state was corrupted by the divergent attempt.
- **Regression paths, all unchanged**: valid rename + `search_config` merge → `code 0` with updated name; fully unknown search id → `code 109` "no authorization" (the earlier guard is intact); rename to another VALID search's name → `code 102` "Duplicated search name.".
- **Code review**: the change is minimal and correctly scoped; the new early return matches the message/code of the branch it replaces, runs after authorization and before any mutation, and introduces no ordering, concurrency, or error-handling differences.

## Previously noted, still non-blocking (not new findings)

- No `logging.warning` on the new not-found early return (also suggested by the bot review) — nice-to-have for traceability.
- The PR description says the old behavior returned "HTTP 500"; it was actually HTTP 200 with `code 100` and the `IndexError` repr in `message`. Substance of the claim is correct.

The production fix is good; updating the one-line test stub would make this merge-ready.

euvre:
Re-audit of head `9d6e18791` (the previous head `269901e39` plus a merge of `main`). The runtime fix itself remains correct and was re-verified live end to end on this round's head; the one open item from the previous review is still unaddressed: this PR breaks an existing unit test, and the new commits did not touch it.

## Findings

1. **[major] Existing unit test breaks: `test/testcases/test_web_api/test_search_app/test_search_routes_unit.py:364-370`**

   `test_update_and_detail_route_matrix_unit` stubs `SearchService.query` to return `[None]` (line 366) and asserts the update route answers `RetCode.DATA_ERROR` with "Cannot find search" (lines 369-370). That stub encodes the old implementation's behavior, where `query(...)[0]` produced `None` and the (then live) `if not search_app` guard fired. With this PR, a non-empty list `[None]` passes the new emptiness check at `api/apps/restful_apis/search_api.py:148`, `search_app` becomes `None`, and the next statement (`search_app.name.lower()`, `search_api.py:152`) raises `AttributeError`, which the outer `except` converts to `code 100`.

   Independently reproduced on this head (A/B, same command):
   - with the PR applied: `AssertionError: assert 100 == 102` (1 failed, 2 passed);
   - with only `api/apps/restful_apis/search_api.py` reverted to the pre-PR implementation: 3 passed.

   Suggested direction (author's call): update the stub from `[None]` to `[]` — and the adjacent "query returns [None]" comment — so the test keeps pinning the "Cannot find search" contract under the new implementation. Production `query()` returns model rows or an empty list, never `[None]`, so this is a test-contract update, not a production problem.

   Context, not counted against the PR: the `test_web_api` tree is not referenced by the CI workflows (they run `test/testcases/test_sdk_api` and `test/testcases/restful_api`), and running that file standalone needs the repo's `test` dependency group plus a session that has `api.db.services.knowledgebase_service` / `api.utils.pagination_utils` importable — both pre-existing conditions of the test harness.

## What was verified as OK (live stack running this PR's head)

- **Fixed path**: created a search app via the REST API, then made its row diverge (`created_by` = caller, `tenant_id` = a different tenant) so the earlier `accessible4deletion` guard passes but the tenant-scoped query returns empty. `PUT /api/v1/searches/<id>` returns `{"code": 102, "message": "Cannot find search <id>"}` — the intended envelope.
- **Causality (live A/B)**: with the identical divergent row and identical request, temporarily reverting only `search_api.py` to the pre-PR implementation and restarting the stack yields `{"code": 100, "message": "IndexError('list index out of range')"}` (raw exception leaked to the client, HTTP 200); restoring the PR code and restarting yields the 102 envelope again. The PR is what fixes the live behavior.
- **Regression paths, all unchanged**: valid rename + `search_config` merge → `code 0` with updated name and merged config; fully unknown search id → `code 109` "no authorization" (the earlier guard is intact); rename to another VALID search's name → `code 102` "Duplicated search name."; after restoring the diverged row's `tenant_id`, the same PUT succeeds with `code 0` — no state was corrupted by the failed attempts.
- **Code review**: the change is minimal and correctly scoped; the new early return runs after authorization and before any mutation, keeps the same message/code as the branch it replaces, and matches the not-found pattern already used in `completion()` (`api/apps/restful_apis/search_api.py:216`).

## Previously noted, still non-blocking

- No `logging.warning` on the new not-found early return (also suggested by the bot review) — nice-to-have for traceability.
- The PR description says the old behavior returned "HTTP 500"; it is actually HTTP 200 with `code 100` and the `IndexError` repr in `message` (verified live). Substance of the claim is correct.

The production fix is good; updating the one-line test stub would make this merge-ready.

euvre:
Re-audit of head `87f87f206` (earlier audits covered `269901e39` and `9d6e18791`). The runtime fix in this PR remains correct, but the ground moved underneath it: the same fix has already landed on `main` via #19380 (commit `3931c69a5`), and this head — a merge of `main` into the branch — resolves the overlap in favor of the PR's own variant. The PR's own delta since its merge base (which already contains #19380) is no longer the IndexError fix; it is a rename plus a log deletion.

## Findings

1. **[major] PR is redundant against `main`; its only net effect there is deleting a warning log that `main` deliberately added** — `api/apps/restful_apis/search_api.py:147-150`.

   Relative to the merge base (which already includes #19380), this PR changes exactly two things in `update()`: renaming the local `search_apps` → `search_results` (no behavior change) and removing `logging.warning("Search update target missing under caller tenant: search_id=%s tenant_id=%s", ...)`, the line #19380 added on `main` (also the bot review's suggestion on this very PR). Nothing else in `main` has touched these lines since, so merging this PR would apply cleanly and silently drop that traceability while adding nothing else. Verified live on this head's build: driving the not-found branch repeatedly produces 0 occurrences of that warning in the backend log, while the client-visible response is identical to `main`'s.

   Suggested direction (author's call): close this PR in favor of #19380, or rework the branch so merging it is a no-op for this block (take `main`'s version, keeping the `logging.warning`). If the `search_results` naming is preferred for consistency, it can be proposed without dropping the log line.

2. **[minor] Previously raised test breakage: still unaddressed in this PR, but no longer attributable to it** — `test/testcases/test_web_api/test_search_app/test_search_routes_unit.py:364-370`.

   Earlier rounds flagged that this test stubs `SearchService.query` to return `[None]`, which the new guard semantics don't handle (a non-empty `[None]` passes the emptiness check, then `search_app.name` raises `AttributeError`, yielding an exception envelope instead of the asserted 102). The current head still does not touch that stub. However, `main` now has the same guard semantics via #19380, and an A/B run on this head confirms the identical outcome for both the PR's and `main`'s version of the handler (`'NoneType' object has no attribute 'name'`), so this is now a pre-existing `main` condition rather than a regression introduced by this PR. Updating the stub to `[]` (and its "query returns [None]" comment) remains the right cleanup — on `main`.

## What was verified as OK (on this head `87f87f206`)

- **Unit tier**: `test/unit_test/api/apps/restful_apis/test_update_search_missing.py` (added on `main` by #19380, present in this merge) passes 2/2 — empty query result → `code 102` "Cannot find search <id>"; found path → `code 0`.
- **Live end to end against the Python API running this head** (build version string `v0.26.0-44-g87f87f206`), driven from a real browser session plus REST calls with an API token:
  - `PUT /api/v1/searches/<id>` valid rename → `code 0`, name persisted.
  - Unknown id → `code 109` "no authorization" (the earlier `accessible4deletion` guard intact).
  - Diverged row (`created_by` = caller, `tenant_id` = another tenant, so the earlier guard passes but the tenant-scoped query is empty) → `code 102`, `"Cannot find search <id>"` — exactly the branch this PR protects.
  - Restoring the row makes the same PUT succeed with `code 0` — no state corrupted by the failed attempts.
  - Screenshots: `r3-01` list + unknown-id responses, `r3-02` the 102 not-found branch, `r3-03` the Search UI page.
- Environment note: the audit worker's Go API gateway failed during startup (doc-engine Elasticsearch ping) and was left unused; this PR modifies only the Python API path, which was verified directly. The UI Search page itself renders (its list call routes through the Go gateway, so it showed the empty state) — unrelated to this change.

The guarded behavior itself is correct and verified on this head; what stands in the way of merging as-is is the redundancy against #19380 and the dropped warning log, not the fix.

euvre:
Re-audit of head `1fb247f46` (earlier audits covered `269901e39`, `9d6e18791`, `87f87f206`). The runtime fix in this PR remains correct and was re-verified live end to end on this head — but the previously raised major finding is still unaddressed: the new commit is only another merge of `main` into the branch, and `api/apps/restful_apis/search_api.py` is byte-identical between `87f87f206` and `1fb247f46`.

## Findings

1. **[major] PR is redundant against `main`; its only net effect there is deleting a warning log that `main` deliberately added — `api/apps/restful_apis/search_api.py:148-150`**

   The same fix already landed on `main` via #19380 (commit `3931c69a5`), which is an ancestor of this branch's merge base. Relative to that merge base, this PR changes exactly two things in `update()`: renaming the local `search_apps` → `search_results` (no behavior change) and removing `logging.warning("Search update target missing under caller tenant: search_id=%s tenant_id=%s", ...)` — the line #19380 added on `main`, and the same logging the bot review on this PR suggested adding. Nothing else on `main` has touched these lines since (`git log 3931c69a5..origin/main -- api/apps/restful_apis/search_api.py` is empty), so merging this PR as-is would apply cleanly and silently drop that traceability while adding nothing else. Verified live on this head's build: driving the not-found branch repeatedly produces 0 occurrences of that warning in the backend log, while the client-visible response is identical to `main`'s.

   Suggested direction (author's call): close this PR in favor of #19380, or rework the branch so merging it is a no-op for this block (take `main`'s version, keeping the `logging.warning`). If the `search_results` naming is preferred for consistency, it can be proposed without dropping the log line.

2. **[minor, pre-existing on `main`, not counted against this PR] Old unit-test stub still pins superseded semantics — `test/testcases/test_web_api/test_search_app/test_search_routes_unit.py:364-370`**

   This test stubs `SearchService.query` to return `[None]` (line 366) and asserts the update route answers `RetCode.DATA_ERROR` with "Cannot find search". Under the emptiness-check semantics — which `main` itself now has via #19380 — a non-empty `[None]` passes the guard and the handler then raises `AttributeError` on `search_app.name`. Established in earlier rounds by A/B run: the outcome is identical for this PR's and `main`'s version of the handler, so this is now a pre-existing `main` condition rather than a regression from this PR. Updating the stub to `[]` (and its "query returns [None]" comment) remains the right cleanup — on `main`.

3. **[nit] PR description inaccuracy (carried over)** — the old behavior was HTTP 200 with `code 100` and the `IndexError` repr in `message`, not "HTTP 500" as the description states. Substance of the claim is correct.

## What was verified as OK (on this head `1fb247f46`, worker confirmed running this commit)

- **Unit tier**: `test/unit_test/api/apps/restful_apis/test_update_search_missing.py` (added on `main` by #19380, present in this merge) passes 2/2 — empty query result → `code 102` "Cannot find search <id>"; found path → `code 0`.
- **Live end to end against the Python API running this head**, via REST calls with an API token plus a real browser session:
  - `PUT /api/v1/searches/<id>` valid rename → `code 0`, name persisted.
  - Unknown id → `code 109` "no authorization" (the earlier `accessible4deletion` guard intact).
  - Rename to another VALID search's name → `code 102` "Duplicated search name." (adjacent regression path, unchanged).
  - Diverged row (`created_by` = caller, `tenant_id` = another tenant, so the earlier guard passes but the tenant-scoped query is empty) → `code 102`, `"Cannot find search <id>"` — exactly the branch this PR protects.
  - Restoring the row makes the same PUT succeed with `code 0` and the new name persisted in the DB — no state corrupted by the failed attempts.
  - Evidence: `r4-01-searches-ui-page.png` (authenticated Searches UI), `r4-02-unknown-id-109.png`, `r4-03-diverged-put-102.png`, `r4-04-restored-put-0.png` (live REST responses rendered in-browser), plus JSON captures `r4-01`–`r4-07`.
- **Code review**: the guarded block matches the message/code of the branch it replaces, runs after authorization and before any mutation; no ordering, concurrency, or error-handling differences; no scope creep (the whole PR delta is the one file, +3/−4).
- Environment note: the audit worker's Go API gateway failed during startup (doc-engine ping) and stayed unused, same as the previous round; this PR modifies only the Python API path, which was verified directly.

The guarded behavior itself is correct and verified on this head; what stands in the way of merging as-is is the redundancy against #19380 and the dropped warning log, not the fix.

euvre:
Re-audit of head `ce5175a3d6` (earlier audits covered `269901e39`, `9d6e18791`, `87f87f206`, `1fb247f46`). The new head is only another merge of `main`; `api/apps/restful_apis/search_api.py` is byte-identical to the previously audited head, and the two non-merge commits picked up in between are main's own (#19418, #19055 — neither touches this file). The previously raised major finding is therefore still unaddressed, and it still blocks merge as-is.

## Findings

1. **[major] PR is redundant against `main`; its only net effect there is deleting a warning log that `main` deliberately added — `api/apps/restful_apis/search_api.py:145-150`**

   The same guard already landed on `main` via #19380 (commit `3931c69a5`), which is an ancestor of this branch's merge base. Relative to that merge base, this PR changes exactly two things in `update()`: renaming the local `search_apps` → `search_results` (no behavior change) and removing `logging.warning("Search update target missing under caller tenant: search_id=%s tenant_id=%s", ...)` — the line #19380 added on `main`, and the same logging the bot review on this PR suggested adding. Nothing else on `main` has touched these lines since (`git log 3931c69a5..origin/main -- api/apps/restful_apis/search_api.py` is empty), so merging this PR as-is would apply cleanly and silently drop that traceability while adding nothing else. Verified live on this head's build: driving the not-found branch repeatedly produces 0 occurrences of that warning in the backend log, while the client-visible response is identical to `main`'s.

   Suggested direction (author's call): close this PR in favor of #19380, or rework the branch so merging it is a no-op for this block (take `main`'s version, keeping the `logging.warning`). If the `search_results` naming is preferred for consistency, it can be proposed without dropping the log line.

2. **[minor, pre-existing on `main`, not counted against this PR] Old unit-test stub still pins superseded semantics — `test/testcases/test_web_api/test_search_app/test_search_routes_unit.py:364-370`**

   This test stubs `SearchService.query` to return `[None]` and asserts the update route answers `RetCode.DATA_ERROR` with "Cannot find search". Under the emptiness-check semantics — which `main` itself now has via #19380 — a non-empty `[None]` passes the guard and the handler then raises `AttributeError` on `search_app.name`. Established in earlier rounds by A/B run: the outcome is identical for this PR's and `main`'s version of the handler, so this is a pre-existing `main` condition rather than a regression from this PR. Updating the stub to `[]` (and its "query returns [None]" comment) remains the right cleanup — on `main`.

3. **[nit] PR description inaccuracy (carried over)** — the old behavior was HTTP 200 with `code 100` and the `IndexError` repr in `message`, not "HTTP 500" as the description states. The substance of the claim is correct.

## What was verified as OK (on this head `ce5175a3d6`, worker confirmed running this commit)

- **Unit tier**: `test/unit_test/api/apps/restful_apis/test_update_search_missing.py` (added on `main` by #19380, present in this merge) passes 2/2 — empty query result → `code 102` "Cannot find search <id>"; found path → `code 0`.
- **Live end to end against the Python API running this head**, driven from an authenticated browser session using an API token (resources named `audit-pr16027-r5-*`):
  - `POST /api/v1/searches` create → `code 0`.
  - `PUT /api/v1/searches/<id>` valid rename → `code 0`, name persisted (re-read via GET confirms).
  - Rename to another VALID search's name → `code 102` "Duplicated search name." (adjacent regression path, unchanged).
  - Diverged row (`created_by` = caller, `tenant_id` = another tenant, so the earlier `accessible4deletion` guard passes but the tenant-scoped query is empty) → `code 102`, `"Cannot find search <id>"` — exactly the branch this PR protects.
  - Unknown id → `code 109` "no authorization" (the earlier guard intact).
  - Restoring the row makes the same PUT succeed with `code 0` and the new name persisted in the DB — no state corrupted by the failed attempts.
  - Evidence: `01-scenario-put-searches.png` (full scenario table rendered in-browser), `02-searches-ui-page` UI state, plus JSON captures `r5-01`–`r5-06`.
- **Code review**: the guarded block matches the message/code of the branch it replaces, runs after authorization and before any mutation; no ordering, concurrency, or error-handling differences; no scope creep (the whole PR delta is the one file, +3/−4).
- Environment note: the audit worker's Go API gateway failed during startup (doc-engine ping) and stayed unused, same as every previous round; this PR modifies only the Python API path, which was verified directly. The web UI Searches page renders but its list call routes through that dead gateway (empty state) — unrelated to this change.

The guarded behavior itself is correct and verified on this head; what stands in the way of merging as-is is the redundancy against #19380 and the dropped warning log, not the fix.

PR #16038, opened 2026-06-15
feat(go): add OceanBase status endpoint with live health probe

kinjitakabe:
# Go: add OceanBase /system/oceanbase/status endpoint with live health probe

## What

Ports the Python `GET /api/v1/system/oceanbase/status` endpoint to the Go server. The endpoint connects to OceanBase over the MySQL wire protocol and reports **live** health + performance metrics.

## Changes

- **`internal/service/oceanbase_probe.go`** (new) — a `database/sql` MySQL-wire probe gathering health (`uri`, `version_comment`) and performance metrics (`latency_ms`, `active_connections`, `max_connections`, `slow_queries`, `storage_used/total`). Mirrors Python `OBConnection.health()` + `get_performance_metrics()`. Each metric query is isolated so a missing OB-specific table/variable degrades gracefully instead of failing the whole probe.
- **`internal/service/system.go`** — `GetOceanbaseStatus()` runs the probe, gated on the `oceanbase` config section being present (rather than `DOC_ENGINE==oceanbase`), so health is observable while ES/Infinity remain the active doc engine. No config → the Python-parity `"OceanBase is not in use."` path.
- **`internal/handler/system.go`** — `OceanbaseStatus` handler returning the Python-parity envelope: HTTP 200 / `code 0` on success, HTTP 200 / `code 500` with `data.status="error"` when unconfigured.
- **`internal/router/router.go`** — registers `GET /api/v1/system/oceanbase/status` in the authenticated `system` group.

## Verification

Deployed the full Go stack locally (ES + MySQL + Redis + MinIO + `admin_server` + `ragflow_server`) against a running **OceanBase (SeekDB) `4.3.5.3`** instance, logged in through the web UI, and hit the endpoint from the live session:

```json
{
  "code": 0,
  "data": {
    "status": "alive",
    "message": {
      "health": { "uri": "localhost:2881", "version_comment": "OceanBase 4.3.5.3 seekdb (r1.3.0.0)" },
      "performance": {
        "connection": "connected", "latency_ms": 9.759,
        "active_connections": 2, "max_connections": "2147483647",
        "slow_queries": 0, "storage_used": "N/A", "storage_total": "N/A"
      }
    }
  },
  "message": "success"
}
```

Live response rendered from the logged-in RAGFlow browser session → Go backend → OceanBase:

oceanbase-status-in-browser

Also verified: unauthenticated requests return `401`; missing config returns the `code 500` "not in use" envelope.

## Notes

- Local deploy-only edits (`conf/service_conf.yaml`, `web/package-lock.json`) are intentionally **excluded** from this PR.
- `storage_*` reports `N/A` against SeekDB (empty doc DB + no `oceanbase.__all_disk_stat`); full OceanBase populates these — the graceful fallback matches the Python reference.

### DIFF
--- a/internal/handler/system.go
+++ b/internal/handler/system.go
@@ -136,6 +136,38 @@ func (h *SystemHandler) GetStatus(c *gin.Context) {
 	})
 }
 
+// OceanbaseStatus returns OceanBase health + performance metrics.
+// Python parity: api/apps/restful_apis/system_api.py:oceanbase_status.
+func (h *SystemHandler) OceanbaseStatus(c *gin.Context) {
+	_, errorCode, errorMessage := GetUser(c)
+	if errorCode != common.CodeSuccess {
+		jsonError(c, errorCode, errorMessage)
+		return
+	}
+
+	status, err := h.systemService.GetOceanbaseStatus()
+	if err != nil {
+		// Python: get_oceanbase_status() raising propagates to the route's
+		// except branch -> get_json_result(data=..., code=500). The HTTP
+		// status stays 200 while the envelope carries code 500.
+		c.JSON(http.StatusOK, gin.H{
+			"code":    common.CodeServerError,
+			"message": "success",
+			"data": gin.H{
+				"status":  "error",
+				"message": "Failed to get OceanBase status: " + err.Error(),
+			},
+		})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"code":    common.CodeSuccess,
+		"message": "success",
+		"data":    status,
+	})
+}
+
 // GetVersion get RAGFlow version
 // @Summary Get RAGFlow Version
 // @Description Get the current version of the application
--- a/internal/handler/system.go
+++ b/internal/handler/system.go
@@ -147,12 +147,14 @@ func (h *SystemHandler) OceanbaseStatus(c *gin.Context) {
 
 	status, err := h.systemService.GetOceanbaseStatus()
 	if err != nil {
-		// Python: get_oceanbase_status() raising propagates to the route's
-		// except branch -> get_json_result(data=..., code=500). The HTTP
-		// status stays 200 while the envelope carries code 500.
+		// Python parity: get_oceanbase_status() raising propagates to the
+		// route's except branch -> code 500 with the error nested in data,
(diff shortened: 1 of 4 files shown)


Haruko386:
Hi @kinjitakabe, I shall review your code very soon
Before that, would you solve some issues raised by @coderabbitai ?

Haruko386 (changes requested):
Hi @kinjitakabe , thank you for your contribution
#### problem 
+ High: The enablement condition differs. Python only allows this endpoint when `DOC_ENGINE=oceanbase`; otherwise it raises "OceanBase is not in use.". Go instead probes OceanBase whenever oceanbase.config exists. Also, Go config currently rejects `DOC_ENGINE=oceanbase`, so strict Python parity is impossible today. See `health_utils.py`, system.go, config.go.
+ High: Success status logic differs. Python returns alive only when health_info.get("status") == "healthy", but `OBConnectionBase.health()` does not return a status field, so even a successful Python connection may become timeout. Go returns alive whenever the probe succeeds. See` health_utils.py`, `ob_conn_base.py`, system.go.
+ Medium: Go does not support Python’s oceanbase.scheme: mysql path. Python reuses the mysql config when `scheme == mysql`; Go always reads oceanbase.config. See `ob_conn_pool.py`, oceanbase_probe.go.
+ Medium: Performance fields are not fully aligned. Python includes query_per_second and pool_size, uses integer max_connections, and attempts storage_total with fallback. Go misses query_per_second/pool_size, returns max_connections as string, and always sets storage_total to "N/A". See `ob_conn.py`, `oceanbase_probe.go`.
+ Low: Error envelope differs. Python returns code=500 but keeps top-level message as "success"; Go changes top-level message to "Failed to get OceanBase status". This is arguably better, but not strict parity. See system_api.py, handler/system.go.
+ Low: Go naming should use the common initialism style: GetOceanBaseStatus / OceanBaseStatus, not GetOceanbaseStatus / OceanbaseStatus.

#### Recommendation
Decide the intended contract first. If Go should match Python, gate on `DOC_ENGINE=oceanbase` and add real OceanBase doc-engine config support in Go. If Go is intentionally exposing an independent OceanBase probe, update comments/tests/docs and stop calling it Python parity. In either case, fix the health.status semantics across both sides so alive/timeout is consistent.
Tests were not run; this was a static code review

PR #16044, opened 2026-06-15
fix(llm): raise a clear error for unsupported Bedrock embedding models

### LINKED ISSUE #16043: BedrockEmbed raises confusing UnboundLocalError for non-amazon/cohere embedding models

`BedrockEmbed.encode()` and `encode_queries()` in `rag/llm/embedding_model.py` build the Bedrock request body only for `amazon.*` and `cohere.*` models:

```python
if self.is_amazon:
    body = {"inputText": text}
elif self.is_cohere:
    body = {"texts": [text], "input_type": "search_document"}
# no else
response = self.client.invoke_model(modelId=self.model_name, body=json.dumps(body))
```

If a user configures a Bedrock embedding model whose id starts with anything else (e.g. a custom or newly added model), `body` is never assigned and the call fails with:

```
UnboundLocalError: cannot access local variable 'body' where it is not associated with a value
```

which gives no hint about the real cause. `BedrockEmbed.__init__` already raises a clear `ValueError` when `auth_mode` is missing, so this is the one spot in the class that surfaces a confusing error instead of an actionable one.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Proposed fix

Add an explicit `else` branch raising `EmbeddingError` that names the unsupported model. `amazon.*` / `cohere.*` behavior is unchanged. I'll open a small PR with regression tests.

---
*Prepared with the help of an AI agent (Claude Code); a human reviewed the finding and fix before filing.*

kimnamu:

Closes #16043

`BedrockEmbed.encode()` / `encode_queries()` built the request body only for `amazon.*` and `cohere.*` models, leaving `body` unbound for any other model id. The call then failed with a confusing `UnboundLocalError: cannot access local variable 'body'` instead of an actionable message.

**Before / After**

| Model id | Before | After |
|----------|--------|-------|
| `amazon.titan-embed-text-v2:0` | ✅ works | ✅ works (unchanged) |
| `cohere.embed-*` | ✅ works | ✅ works (unchanged) |
| anything else (e.g. `meta.*`) | ❌ `UnboundLocalError: ... 'body'` | ✅ `EmbeddingError: BedrockEmbed: unsupported embedding model '...' (expected an 'amazon.' or 'cohere.' model)` |

This matches the existing pattern in the class — `__init__` already raises a clear `ValueError` when `auth_mode` is missing, and the module uses `EmbeddingError` for embedding failures elsewhere.

### Type of change

### How tested

Added two regression tests in `test/unit_test/rag/llm/test_embedding_model.py` (`TestBedrockResponseParsing`): an unsupported provider now raises `EmbeddingError` from both `encode()` and `encode_queries()`. The existing amazon/cohere parsing tests still pass. `ruff check` is clean.

---
*This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, the reasoning, and the tests before submitting.*


kimnamu:
Thanks for the review! Good call — since the whole point of this PR is a *clear, actionable* message, the tests should pin the message contract too. I've updated both regression tests to assert that the `EmbeddingError` names the offending model and the expected `amazon.` / `cohere.` prefixes (via `pytest.raises(..., match=...)`), so future wording regressions are caught. Pushed in ddf54c8a.

Lynn-Inf (approved):
LGTM

kimnamu:
Merged `main` to resolve the conflict in `test/unit_test/rag/llm/test_embedding_model.py` (both sides appended tests after `TestBedrockResponseParsing`; kept both). No source changes — the diff against `main` is still the two `else` branches plus the two regression tests. `test/unit_test/rag/llm/test_embedding_model.py` passes locally (27 tests), `ruff check` / `ruff format --check` clean.

kimnamu:
@Lynn-Inf @JinHai-CN thank you for the review back in June and for merging this on 09-08. This note was queued before the merge and went out late, so please disregard the status table it originally carried. Nothing further is needed here.

PR #16057, opened 2026-06-16
feat(llm): support model-level custom headers and default generation params

### LINKED ISSUE #15981: [Feature Request]:Support Custom Model Headers and Generation Parameters
## Summary

RAGFlow should allow users to configure custom request headers and model generation parameters, such as `temperature`, both at the model configuration level and, where appropriate, per chat.

## Problem

Some LLM providers, OpenAI-compatible gateways, enterprise proxies, and self-hosted deployments require additional request customization beyond the base URL, model name, and API key.

Examples include:

- Custom authentication headers
- Tenant, workspace, or project headers
- Routing headers for internal gateways
- Provider-specific request options
- Model generation settings such as `temperature`, `top_p`, or `max_tokens`

Currently, users may need to create duplicate model configurations or modify code to support these cases.

### 1. Custom Headers in Model Configuration

Allow users to define arbitrary HTTP headers for a configured model or provider.

Example:

```json
{
  "X-Project-ID": "abc123",
  "X-Workspace": "ragflow-prod"
}
```

These headers should be included when RAGFlow sends requests to that model provider.

### 2. Default Generation Parameters in Model Configuration

Allow users to define default generation parameters for a model.

Example:

```json
{
  "temperature": 0.2,
  "top_p": 0.9,
  "max_tokens": 2048
}
```

These defaults should be applied whenever the model is used, unless overridden by a more specific chat-level setting.

### 3. Per-Chat Generation Parameter Overrides

Allow chat sessions, assistants, or chat configurations to override supported generation parameters.

Example use cases:

- A support chatbot using `temperature: 0`
- A creative writing assistant using `temperature: 0.8`
- A summarization chat using a lower `max_tokens`
- A reasoning-heavy chat using different sampling parameters

## Expected Behavior

- Model-level headers are applied to all requests for that model.
- Model-level generation parameters are used as defaults.
- Chat-level generation parameters override model defaults where supported.
- Custom headers should remain model/provider-level unless there is a safe and explicit use case for per-chat headers.
- Sensitive header values should not be exposed in frontend responses, logs, or exported configuration.
- Unsupported parameters should either be ignored safely or rejected with a clear validation error.

## Possible UI/API Changes

Add an advanced configuration section to model provider settings with fields for:

- Custom headers
- Default generation parameters

Add chat-level generation settings for common parameters such as:

- `temperature`
- `top_p`
- `max_tokens`
- `presence_penalty`
- `frequency_penalty`

## Why This Matters

This would make RAGFlow easier to use with:

- Custom LLM gateways
- OpenAI-compatible APIs
- Enterprise proxy services
- Multi-tenant deployments
- Self-hosted model providers
- Different chat workflows requiring different model behavior

It would also reduce the need to create duplicate model entries just to change simple generation settings like `temperature`.

abyssumining:
Implements issue #15981 for the OpenAI-compatible (Base) and LiteLLM backends.

A model configuration can now carry:
- default_headers: custom HTTP headers (e.g. X-Project-ID, gateway/proxy auth or
  routing headers) forwarded to the provider. For the OpenAI-compatible backend
  they are passed as the client's `default_headers`; for LiteLLM they are
  forwarded via the per-request `headers` argument. Header values are never logged.
- default_gen_conf: model-level default generation parameters (temperature,
  top_p, max_tokens, presence_penalty, frequency_penalty), validated and
  range-checked. They are layered underneath each request's gen_conf in
  _clean_conf, so a chat's own settings (which already flow through gen_conf)
  always take precedence. Unsupported / out-of-range values are ignored safely.

Both are optional and parsed defensively — a malformed entry can never break
client construction, and when unset the behavior is unchanged.

Adds unit tests for header/param validation, merge precedence, and that custom
headers reach the client / completion call on both backends.

Closes #15981


Lynn-Inf (commented):
The plumbing itself is well done: `parse_custom_headers` accepts a dict or a JSON-object string, excludes `bool` explicitly (a nice touch, since `bool` is an `int` subclass), skips malformed entries instead of breaking client construction, and — importantly — never logs header values, which may carry secrets. `validate_default_gen_conf` silently drops unsupported, mistyped or out-of-range entries as the issue asks, and layering the defaults *under* each request's gen_conf so chat-level settings win is the right precedence. Unset means unchanged behaviour, and forwarding via the client's `default_headers` on the Base backend vs per-request `headers` on LiteLLM matches how each SDK works.

**But I can't find a way for an operator to actually set either option**, which the description implies should exist ("an operator can attach…"). This PR touches only `rag/llm/chat_model.py` plus tests — no frontend, no API layer, no schema change. `TenantLLMService.model_instance` (`api/db/services/tenant_llm_service.py:182-210`) injects only `{"provider": ...}` plus the `**kwargs` it was itself given; it does not spread `model_config` into the constructor, so a `default_headers` value stored on the model config would never reach `Base.__init__` / `LiteLLMBase.__init__`. Grepping `api/db/services/llm_service.py`, `api/apps/restful_apis/llm_api.py` and `api/db/joint_services/tenant_model_service.py` for these three keys returns zero hits. As it stands only the unit tests exercise it, by constructing the class directly. Could you wire it to a real entry point (UI field, API parameter, or persisted model-config keys), or state explicitly that this PR is backend-only with the wiring to follow?

One concrete bug while I was in there: on the Base backend, `max_tokens` in `default_gen_conf` can never take effect. `Base._clean_conf` merges the defaults in and then immediately runs `if "max_tokens" in gen_conf: del gen_conf["max_tokens"]`, so the value is removed in the same call — whereas the LiteLLM path renames it to `max_completion_tokens` and does apply it. Either drop `max_tokens` from `SUPPORTED_DEFAULT_GEN_CONF_KEYS` for the Base path or rename it the way LiteLLM does.

PR #16061, opened 2026-06-16
Refactor: Change table chat_channel status data type.

KevinHuSh:

As title.

### Type of change

- [x] Refactoring


api/db/db_models.py
```diff
@@ -1254,7 +1254,7 @@ class ChatChannel(DataBaseModel):
     channel = CharField(max_length=128, null=False, help_text="Chat channel type", index=True)
     config = JSONField(null=False, default={}, help_text="Channel credential & settings")
     dialog_id = CharField(max_length=32, null=True, default=None, help_text="connected dialog id", index=True)
```
JinHai-CN, line 1256: chat_id

PR #16067, opened 2026-06-16
feat[Go]: implement api /api/v1/datasets/<dataset_id>/chunks POST

Haruko386:

As title

### Type of change


internal/handler/chunk.go
```diff
@@ -203,6 +204,59 @@ func (h *ChunkHandler) Get(c *gin.Context) {
 	})
 }
 
+// Parse reparse the datasets' files
+func (h *ChunkHandler) Parse(c *gin.Context) {
+	user, errorCode, errorMessage := GetUser(c)
```
xugangqiang, line 209: Can we add test for this new method?

xugangqiang:
LGTM

PR #16068, opened 2026-06-16
Fix: build workbook for single-sheet pandas fallback in Excel parser

JSap0914:

When `openpyxl` cannot load a spreadsheet, `RAGFlowExcelParser._load_excel_to_workbook` falls back to pandas:

```python
dfs = pd.read_excel(file_like_object, sheet_name=None)
return RAGFlowExcelParser._dataframe_to_workbook(dfs)
```

`pd.read_excel(sheet_name=None)` **always** returns a `dict` keyed by sheet name, even when the workbook has a single sheet. `_dataframe_to_workbook` only routed dicts with more than one sheet to `_dataframes_to_workbook`:

```python
if isinstance(df, dict) and len(df) > 1:
    return RAGFlowExcelParser._dataframes_to_workbook(df)
df = RAGFlowExcelParser._clean_dataframe(df)   # df is still a dict here
```

So a single-sheet dict fell through to `_clean_dataframe(dict)`, which calls `dict.apply(...)` and raises:

```
AttributeError: 'dict' object has no attribute 'apply'
```

This makes the default pandas-engine branch effectively dead code for single-sheet files (it can only ever succeed via the secondary calamine fallback), and a single-sheet file fails to parse whenever calamine can't read it.

The fix routes any `dict` to `_dataframes_to_workbook`, which already iterates sheets and preserves their names. The CSV path and the calamine path both pass a `DataFrame` (not a dict) and are unaffected.

### Verification

Added focused unit tests in `test/unit_test/deepdoc/parser/test_excel_parser.py` covering the single-sheet and multi-sheet dict cases.

```
$ python -m pytest test/unit_test/deepdoc/parser/test_excel_parser.py
5 passed
```

Before the fix, `test_single_sheet_dict_builds_workbook` failed with `AttributeError: 'dict' object has no attribute 'apply'`; the multi-sheet test passed, confirming the regression was isolated to the single-sheet path.

### Type of change


JinHai-CN:
@JSap0914 would you please resolve the conflicts?

PR #16100, opened 2026-06-16
Docs: update chat completions docs

buua436:
Syncs the /api/v1/chat/completions docs with the current behavior, including the new legacy streaming mode.
### Type of change
- [x]  Documentation Update


writinwaters (approved):
LGTM.

PR #16101, opened 2026-06-16
fix: avoid blocking the event loop with requests.post in async Discord handler

kratos0718:
## What

`on_message` in `rag/svr/discord_svr.py` is an `async` discord.py event handler, but it makes a **synchronous** `requests.post(...)` call:

```python
async def on_message(message):
    ...
    response = requests.post(URL, json=JSON_DATA)
```

A blocking call inside an async handler blocks the **entire event loop** for the duration of the HTTP request. While it waits, the Discord bot can't process other messages, send typing indicators, or do any concurrent work — the whole bot freezes per request.

## Fix

Offload the blocking call to a worker thread with `asyncio.to_thread`, keeping the event loop responsive:

```python
response = await asyncio.to_thread(requests.post, URL, json=JSON_DATA)
```

`asyncio` is already imported; no new dependency, behavior unchanged.

- [x] `python -m py_compile` clean.

### DIFF
--- a/rag/svr/discord_svr.py
+++ b/rag/svr/discord_svr.py
@@ -49,9 +49,23 @@ async def on_message(message):
         if len(message.content.split('> ')) == 1:
             await message.channel.send("Hi~ How can I help you? ")
         else:
-            JSON_DATA['word'] = message.content.split('> ')[1]
-            response = await asyncio.to_thread(requests.post, URL, json=JSON_DATA)
-            response_data = response.json().get('data', [])
+            # Build an isolated payload per message; mutating the shared global
+            # JSON_DATA would race once messages are handled concurrently.
+            payload = {**JSON_DATA, "word": message.content.split('> ')[1]}
+            try:
+                response = await asyncio.to_thread(
+                    requests.post, URL, json=payload, timeout=30
+                )
+                response.raise_for_status()
+                response_data = response.json().get('data', [])
+            except (requests.exceptions.RequestException, ValueError):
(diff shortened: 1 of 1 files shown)


kratos0718:
Thanks for the review @coderabbitai — all three points addressed in the latest commit:

1. **Race on global `JSON_DATA`** — now build a per-message local `payload = {**JSON_DATA, "word": ...}` instead of mutating the shared global. (Especially relevant since this PR makes `on_message` non-blocking, allowing concurrent handling.)
2. **Timeout + error handling** — added `timeout=30` and a `try/except (requests.exceptions.RequestException, ValueError)` that logs the failure and replies to the user instead of hanging or crashing.
3. **JSON parse safety** — the same `except` covers `JSONDecodeError` (a subclass of `ValueError`); also defaulted `res = ""` so a response without a type-1 item doesn't raise.

The core fix remains: offloading the blocking `requests.post` with `asyncio.to_thread` so the bot's event loop stays responsive.

kratos0718:
Gentle ping — this one looks ready and has been open since 16 June.

All three CodeRabbit points were addressed the same day (per-message local payload instead of mutating the shared global, and the blocking call moved off the event loop), and it confirmed they were handled. The PR is still `MERGEABLE` with no conflicts.

Re-checked against current `main` before pinging: the synchronous `requests` call is still there in `rag/svr/discord_svr.py`, so the fix continues to apply cleanly.

Happy to rebase if that helps, or to close it if you'd rather solve it differently.

JinHai-CN:
This function is switched to chat channel, and this file is removed.

PR #16105, opened 2026-06-16
fix: check REGISTER_ENABLED before auto-registering OIDC users (#16103)

### LINKED ISSUE #16103: [Bug]: OIDC login allows new user registration even when REGISTER_ENABLED=0
### Self Checks

v0.25.6

### RAGFlow image version

v0.25.6

```Markdown
RAGFlow installed via Helm chart
```

### Actual behavior

I configured RAGFlow OIDC authentication following the documentation:
https://ragflow.io/docs/configurations#oauth

I also set:

`REGISTER_ENABLED=0`

My expectation is that user self-registration should be disabled. In this setup, only users already created in the RAGFlow database by an administrator should be allowed to log in through OIDC.

However, users who do not already exist in the database can still authenticate via OIDC and are automatically registered/created in RAGFlow.

Could you clarify whether this is the intended behavior?

### Expected behavior

With REGISTER_ENABLED=0:

- Existing users can log in via OIDC.
- New users are not automatically created via OIDC.
- User accounts must be created beforehand by a RAGFlow administrator.

### Steps to reproduce

```Markdown
Configure OIDC authentication and set REGISTER_ENABLED=0, register as a new OIDC user.
```

### Additional information

_No response_

zaviermeekz-cpu:

The OIDC callback handler automatically creates a new user if the email is not found in the database, even when `REGISTER_ENABLED=0`. This allows new users to bypass registration restrictions.

This PR adds a check for `REGISTER_ENABLED` before the `user_register` call in the OIDC callback. If registration is disabled, the request is rejected with an error.

### Type of change

- [x] Bug Fix

### How has this been tested?

- Manually tested with `REGISTER_ENABLED=0`: OIDC login for an existing user succeeds; login for a new user fails with "registration_disabled".
- With `REGISTER_ENABLED=1`, new users are still created as expected.

### Related Issue

Closes #16103


6ba3i (changes requested):
The fix works technically, but `REGISTER_ENABLED` is the wrong switch for OIDC/OAuth provisioning.
That setting is used for local/public registration, while OIDC provisioning is treated separately. The maintainer discussion in #16103 also points toward restricting OIDC access through groups/roles instead.
With this change, deployments using `REGISTER_ENABLED=0` to block public signup but still relying on OIDC JIT provisioning would break.
I’d keep the two policies separate: use IdP/group-role restrictions, or add a dedicated OIDC provisioning setting if needed.Also add regression tests for new vs existing OIDC users with provisioning enabled/disabled.

6ba3i:
overlaps #19651

PR #16107, opened 2026-06-16
fix(deps): pin crawl4ai<0.8.6 to avoid unclecode-litellm namespace conflict

Attili-sys:
## Problem

The upstream security commit `b4c8711d5` / PR `#15415` correctly bumped `crawl4ai` to `>=0.8.0` to address `CVE-2026-26217`, but because there was no upper bound, `uv` can currently resolve the latest `crawl4ai` version. In practice, that means `crawl4ai>=0.8.6`, which switched from `litellm` to `unclecode-litellm`.

That package installs into the same `litellm/` namespace and conflicts with RAGFlow's pinned `litellm==1.82.5`. This caused the mixed installation to break imports during test collection:

```
ImportError: cannot import name 'validate_and_fix_thinking_param' from 'litellm.utils'
```

This broke 6 unit test collection failures in CI:
- `test/unit_test/agent/component/test_browser_use_component.py`
- `test/unit_test/api/db/services/test_dialog_service_final_answer.py`
- `test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py`
- `test/unit_test/api/db/services/test_evaluation_service_token_usage.py`
- `test/unit_test/api/db/services/test_file_service_upload_document.py`
- `test/unit_test/common/test_think_stream_parser.py`

## Root Cause

`unclecode-litellm==1.81.13` is a fork of `litellm` that ships its own `litellm/` package files. When both `litellm==1.82.5` and `unclecode-litellm==1.81.13` are installed, their files overwrite each other in the `litellm/` namespace, producing a broken hybrid that fails at import time.

## Fix

The proposed constraint:

```toml
crawl4ai>=0.8.0,<0.8.6
```

- Retains the CVE-2026-26217 fix (requires `>=0.8.0`)
- Stays below the version that introduced `unclecode-litellm` as a dependency (`<0.8.6`)
- Removes `unclecode-litellm` from the resolved dependency tree entirely


yuzhichang (changes requested):
Conflict with https://github.com/infiniflow/ragflow/pull/16426.

PR #16111, opened 2026-06-16
fix(go): correct OpenRouter streaming URL routing and reasoning parameter

### LINKED ISSUE #16110: [Bug]: OpenRouter streaming chat — qwen/glm hit an empty async URL and reasoning uses a non-standard `thinking` key
### Self Checks

6865039a2 (main)

### RAGFlow image version

nightly

Go port — `internal/entity/models/openrouter.go`, `ChatStreamlyWithSender`.

### Actual behavior

Two bugs in the OpenRouter streaming chat request builder:

1. **qwen/glm models stream to a broken URL.** The code routes any `qwen`/`glm` model to
   `URLSuffix.AsyncChat`, but `conf/models/openrouter.json` does not define `async_chat`
   (empty), so the request is POSTed to `<base>/` instead of `<base>/chat/completions` —
   streaming fails for every qwen/glm model. The non-stream path has no such branch.

2. **Streaming reasoning is never enabled.** The request sets `reasoning` via a non-standard
   `thinking` key, which OpenRouter ignores. OpenRouter's API (and this provider's non-stream
   request and its streamed `delta.reasoning` parser) use the `reasoning` object. So
   thinking models never return reasoning when streamed.

### Expected behavior

All models stream to `/chat/completions`; reasoning is requested via the standard
`reasoning` object (`{"enabled": true}`, or `{"effort": ...}`), consistent with the
non-stream path and OpenRouter's API.

### Steps to reproduce

1. Configure an OpenRouter chat model in the Go layer; enable streaming + thinking.
2. Use a qwen/glm model → request goes to a URL without `chat/completions` → fails.
3. Use any reasoning model → `reasoning` is never enabled (the `thinking` key is ignored), so no reasoning is streamed.

### Additional information

Verified live against the OpenRouter API: streaming with `reasoning:{"enabled":true}`
returns `delta.reasoning` chunks, while the `thinking` key is non-standard and ignored.

Crystora:

Fixes two bugs in the OpenRouter streaming chat request builder
(`internal/entity/models/openrouter.go`, `ChatStreamlyWithSender`):

1. **qwen/glm models streamed to a broken URL.** The code routed any `qwen`/`glm` model to
   `URLSuffix.AsyncChat`, but `conf/models/openrouter.json` defines no `async_chat` suffix
   (empty), so the request was POSTed to `<base>/` instead of `<base>/chat/completions` —
   breaking streaming for every qwen/glm model. The non-stream path has no such branch.
   Fix: all models use the standard `Chat` suffix, consistent with the non-stream path.

2. **Streaming reasoning was never enabled.** The request set reasoning via a non-standard
   `thinking` key, which OpenRouter ignores. OpenRouter's API — and this provider's own
   non-stream request (line ~110) and its streamed `delta.reasoning` parser (line ~311) —
   use the `reasoning` object. Fix: send `reasoning: {"enabled": <thinking>}` (and
   `{"effort": ...}` when set, taking precedence as in the non-stream path).

Verified live against the OpenRouter API: streaming with `reasoning:{"enabled":true}`
returns `delta.reasoning` chunks (the standard param works), confirming the `thinking`
key was silently ignored. Added a unit test asserting the streaming request targets
`/chat/completions` for a qwen model and sends the standard `reasoning` object (not `thinking`).

`gofmt`/`go vet` clean; `go test ./internal/entity/models/ -run OpenRouter` passes.

Closes #16110

### Type of change

### DIFF
--- a/internal/entity/models/openrouter.go
+++ b/internal/entity/models/openrouter.go
@@ -202,12 +202,11 @@ func (o *OpenRouterModel) ChatStreamlyWithSender(modelName string, messages []Me
 	if err != nil {
 		return err
 	}
+	// All OpenRouter models use the standard chat-completions endpoint, same as
+	// the non-stream path. The previous qwen/glm branch routed to URLSuffix.AsyncChat,
+	// which OpenRouter does not configure (empty suffix) — producing a broken URL and
+	// breaking streaming for every qwen/glm model.
 	url := fmt.Sprintf("%s/%s", resolvedBaseURL, o.baseModel.URLSuffix.Chat)
-
-	modelType := strings.Split(modelName, "_")[0]
-	if modelType == "qwen" || modelType == "glm" {
-		url = fmt.Sprintf("%s/%s", resolvedBaseURL, o.baseModel.URLSuffix.AsyncChat)
-	}
 
 	// Convert messages to API format
 	apiMessages := make([]map[string]interface{}, len(messages))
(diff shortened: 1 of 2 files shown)


Haruko386 (approved):
@bittoby LGTm

PR #16112, opened 2026-06-16
Design/login signup

AmroWaqoor:
Designed the 
auth , forgot password and home pages


yingfeng:
Thank you for your contribution, but why did you modify the RAGFlow logo?

PR #14920, opened 2026-05-14
fix: The output of the parser in the ingestion pipeline contains HTML tags

jonathanchang31:
## Summary
This change fixes ingestion quality issues where MinerU parser output may contain HTML fragments (for example, table-related tags like `<tr>`, `<td>`, `<br>`), which were previously passed directly into chunking/tokenization and degraded chunk quality.

The fix adds a sanitization step in the MinerU parser path so parsed sections are normalized to clean text before chunking.

## Change Type (select all)
- [x] Bug fix
- [x] Ingestion pipeline improvement
- [x] Parser/chunking quality fix

## Related Issue
- https://github.com/infiniflow/ragflow/issues/14831

## Root Cause
MinerU output blocks (`text`, `table_body`, captions/footnotes, etc.) could include HTML-style markup.
Those strings were transferred into section content and then merged/tokenized without tag cleanup, so chunk text contained parser markup.
### Before fix
- MinerU-derived section content could include literal HTML tags.
- Chunker consumed those tags as normal text.
- Result: chunk boundaries and semantic quality were negatively affected.

### After fix
- MinerU section text is sanitized before chunking.
- HTML tags are removed from section content.
- Chunking/tokenization now receives cleaner text, improving chunk quality and retrieval consistency.

## Technical Evidence
- Sanitizer definition:
  - `deepdoc/parser/mineru_parser.py` (`_sanitize_section_text`)
- Sanitizer usage in transfer path:
  - `_transfer_to_sections` applies `section = self._sanitize_section_text(section)`
- Transfer path feeds parser output to downstream chunking as before, now with cleaned text.

## Validation Performed
- [x] Source-level verification that sanitizer is wired into MinerU transfer path
- [x] Syntax check passed:
  - `python3 -m py_compile deepdoc/parser/mineru_parser.py`
- [ ] End-to-end ingestion test with real MinerU output in this environment
- [ ] Automated unit test added for sanitizer edge-cases

## Runtime Validation Note
A direct runtime sample execution was attempted, but this environment is missing dependency `beartype`, which blocked import-time execution of parser modules. This does not affect the source-level integration or syntax validity of the patch.
## Checklist
- [x] Confirmed issue path in MinerU -> sections -> chunking flow
- [x] Added targeted sanitization in MinerU parser
- [x] Ensured sanitization executes before chunking
- [x] Verified file compiles cleanly
- [ ] Add regression/unit tests for sanitizer cases
- [ ] Run full ingestion smoke test in a dependency-complete environment

Issue #14831, opened 2026-05-12
[Question]: The output of the parser in the ingestion pipeline contains HTML tags, which affect the chunking performed by the chunker.


zxcno-1a:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

I have several PDF documents of fill-in-the-blank style reports. When I select the MinerU method for the parser in the ingestion pipeline, I notice that the parsed results contain HTML tags. These HTML tags seriously affect the chunking effect of the chunker, yet there are no other components in the ingestion pipeline to remove these tags. What should I do?

wangq8:
@zxcno-1a you can just disable table recognition, then is should have no html table tag.

<img width="1192" height="505" alt="Image" src="https://github.com/user-attachments/assets/24d00d43-9803-4fd3-a96f-073f687ea189" />

wangq8:
MinerU’s own README says: “Formulas → LaTeX · Tables → HTML” and separately says it converts formulas to LaTeX and tables to HTML.

PR #16118, opened 2026-06-17
Fix MinerU table option sanitization

wangq8:
Follow on issue: #14831 and PR: #14920 to fix the table options, with table recognition enabled, do not sanitize html tags.


xuhaiL:
good.

PR #15154, opened 2026-05-22
Refactor: Task Executor

xugangqiang:
### What problem does this PR solve?

1. Break huge function into smaller pieces
2. Add unit test for the smaller pieces function
3. Layer-ed design
    a. infra layer - task_context.py, recording_context.py, write_operation_interceptor.py, ...
    b. service layer - *_service.py
    c. business layer - task_handler.py
4. Default behavior: use "refactor-ed version" - can switch to original version by change env variable

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [x] Refactoring
- [x] Performance Improvement

PR #16147, opened 2026-06-17
fix: accept legacy positional task executor index

helloxjade:

Before the task executor refactor in #15154, the worker index was read from `sys.argv[1]`, and the in-repo launchers passed it positionally. The refactor migrated the current in-repo launchers to `-i/--index`, but external deployment configurations copied from older RAGFlow versions may still use the former positional interface.

In the affected deployment environment, this caused `argparse` to exit with `unrecognized arguments` and the worker to restart in a loop.

This PR preserves compatibility with those external deployment configurations while keeping the current `-i/--index` behavior. When both forms are provided with conflicting values, the positional value continues to take precedence and a warning is emitted.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Testing

- [x] Added/updated tests
- [x] Verified with:
  - `.venv/bin/pytest -q test/unit_test/rag/svr/test_task_executor_cli.py` — 4 passed
  - `.venv/bin/ruff check rag/svr/task_executor_cli.py test/unit_test/rag/svr/test_task_executor_cli.py`
  - `.venv/bin/python -m py_compile rag/svr/task_executor.py rag/svr/task_executor_cli.py`


helloxjade:
Error logs before fix:
```text
/ragflow/.venv/lib/python3.13/site-packages/graspologic/match/wrappers.py:227: SyntaxWarning: invalid escape sequence '\s'
  .. math:: f(P) = - \sum_{k=1}^K \|A^{(k)} - PB^{(k)}P^T\|_F^2 - \sum_{k=1}^K \|(AB)^{(k)}P^T - P(BA)^{(k)}\|_F^2 + trace(SP^T)
/ragflow/.venv/lib/python3.13/site-packages/graspologic/pipeline/embed/adjacency_spectral_embedding.py:123: SyntaxWarning: invalid escape sequence '\S'
  .. math:: A = U \Sigma V^T
/ragflow/.venv/lib/python3.13/site-packages/graspologic/pipeline/embed/laplacian_spectral_embedding.py:127: SyntaxWarning: invalid escape sequence '\S'
  .. math:: A = U \Sigma V^T
usage: task_executor.py [-h] [-i INDEX] [-t TYPE]
task_executor.py: error: unrecognized arguments: worker_positional_arg
/ragflow/.venv/lib/python3.13/site-packages/xgboost/compat.py:105: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
10:24:39 - LiteLLM:WARNING: get_model_cost_map.py:174 - LiteLLM: Failed to fetch remote model cost map from https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json: The read operation timed out. Falling back to local backup.
WARNING:LiteLLM:LiteLLM: Failed to fetch remote model cost map from https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json: The read operation timed out. Falling back to local backup.
/ragflow/.venv/lib/python3.13/site-packages/scholarly/_scholarly.py:312: SyntaxWarning: invalid escape sequence '\d'
  m = re.search("cites=[\d+,]*", object["citedby_url"])
/ragflow/.venv/lib/python3.13/site-packages/graspologic/align/seedless_procrustes.py:125: SyntaxWarning: invalid escape sequence '\i'
  are represented with matrices :math:`X \in M_{n, d}` and
/ragflow/.venv/lib/python3.13/site-packages/hyppo/kgof/kernel.py:90: SyntaxWarning: invalid escape sequence '\s'
  Compute \sum_{i=1}^d \frac{\partial^2 k(X, Y)}{\partial x_i \partial y_i}
/ragflow/.venv/lib/python3.13/site-packages/hyppo/kgof/kernel.py:129: SyntaxWarning: invalid escape sequence '\s'
  Compute \sum_{i=1}^d \frac{\partial^2 k(x, Y)}{\partial x_i \partial y_i}
/ragflow/.venv/lib/python3.13/site-packages/hyppo/kgof/kernel.py:264: SyntaxWarning: invalid escape sequence '\s'
  Compute \sum_{i=1}^d \frac{\partial^2 k(X, Y)}{\partial x_i \partial y_i}
```

Before this fix, the deployment launcher passes worker index as a positional argument without the `-i` flag. The original task_executor.py only accepts index via `-i/--index`, which leads to the `unrecognized arguments` error and makes the worker crash and restart repeatedly.

This PR keeps the original `-i/--index` logic intact, and adds backward compatibility to support positional worker index as a fallback, which resolves the startup failure.

xugangqiang:
Thanks for the fix. Reviewed the change and have a few findings:

**1. The "legacy positional index" has no in-repo caller (scope/intent)**
All in-repo launchers pass the index via `-i`:
- `docker/entrypoint.sh:230` → `task_executor.py -i "${host_id}_${consumer_id}" -t "common"`
- `docker/entrypoint-go.sh:230` → same
- `docker/launch_backend_service.sh:101` → `task_cmd=("$PY" "rag/svr/task_executor.py" "-i" "$task_id")`

None passes a positional index, so the failing path only exists in an out-of-repo launcher. The shim is harmless and solves your real deployment crash, but it is a compatibility-only branch with no repo contract. Please note this in the PR description so future maintainers don't assume an in-repo caller exists.

**2. Silent precedence conflict when both `-i` and the positional arg are given**
`TE_IDX = args.legacy_index or args.index` makes the positional value win whenever both are present, with no signal. If a launcher ever passes `-i 5` *and* a positional `3`, the worker silently uses `3`. Consider adding a warning when `args.legacy_index is not None and args.index not in ("0", args.legacy_index)` so misconfiguration is visible (the original CodeRabbit suggestion for a conflict warning was not addressed — only the basic info log was added).

**3. No test coverage for the new CLI behavior**
The PR only used `py_compile`. There is no test asserting that (a) a positional index is accepted, (b) `-i` still works, and (c) the two coexist as expected. `test/unit_test/rag/svr/task_executor_refactor/*` does not touch `main()`'s argparse. A small local test driving `argparse` with `sys.argv` would prevent regressions without needing any external service.

Overall the fix is correct and low-risk for in-repo launchers; addressing #2 (conflict warning) and #3 (test) before merge would make it a clean net gain.

Please let me know your thoughts.

helloxjade:
Thanks for the detailed review. I addressed #2 and #3 in c40dbde4e:

- Added a warning when `-i/--index` conflicts with the positional index, while preserving positional precedence.
- Added focused tests covering positional-only, `-i`-only, matching values, and conflicting values.

For #1, I updated the PR description to clarify the compatibility scope. Before the task executor refactor in #15154, the in-repo launchers passed the worker index positionally and `task_executor.py` read it from `sys.argv[1]`. Current in-repo launchers use `-i`, while external deployment configurations copied from older versions may still use the former positional interface.

The focused test suite passes with 4 tests. Could you please take another look?

xugangqiang:
LGTM

PR #16207, opened 2026-06-19
fix(agent): log Wikipedia disambiguation and page errors instead of s…

Willsgao:
## Problem
The Wikipedia tool silently swallows all exceptions with `except Exception: pass`, making it impossible to debug failures when fetching Wikipedia pages.

## Fix
Replace the bare `except Exception: pass` with specific exception handling:
- `DisambiguationError`: log available options
- `PageError`: log page not found
- `Exception`: log unexpected errors with full traceback


yuzhichang (approved):
LGTM

PR #16210, opened 2026-06-19
perf(graphrag): batch entity/relation embeddings in set_graph (#16205)

### LINKED ISSUE #16205: GraphRAG set_graph() extremely slow due to per-entity/relation embedding (10-50x slower than batch)
### Bug Description

The `set_graph()` function in GraphRAG embeds entities and relations **one-by-one** instead of in batches. This causes GraphRAG pipeline to become extremely slow or completely stall on large knowledge graphs (thousands of nodes/edges). Each embedding call is a separate HTTP request to the embedding server, with no batching.

### Steps to Reproduce

1. Create a dataset with GraphRAG chunk method enabled
2. Upload a document that generates a large knowledge graph (e.g., hundreds of pages of Chinese building codes, producing 17,000+ edges)
3. Start parsing → graph construction progresses to edge embedding phase
4. Observe: edge embedding slows to a crawl or stalls entirely (e.g., 17,576 edges, completed 17,509 then stopped)

### Root Cause Analysis

In `rag/ragraph/graph_utils.py` (or similar), `set_graph()` creates individual asyncio tasks for each node/edge embedding call:

```python
# Current: one-by-one
for entity in entities:
    task = asyncio.create_task(encode(entity.text))  # individual call
```

This should instead:
1. Collect all texts
2. Batch cache lookup
3. Batch encode call
4. Batch construct chunks

### Expected Behavior

All entity/relation texts should be batched for embedding (e.g., 32 or 64 per request), leveraging the batching capability of embedding servers (TEI, OpenAI API, etc.). This would improve performance by **10-50x** based on PR #15982 benchmarks.

### Environment

- RAGFlow version: v0.26.0 / v0.26.1
- Deployment: Docker (CPU)
- Embedding model: bge-m3 (1024 dim) via TEI
- Graph scale: ~17,000+ edges
- Hardware: NAS, 8GB RAM, 4-core CPU

### Related Issues

- **#15400** — set_graph() "delete-then-write" data loss risk
- **#15921** — set_graph() performance issue, per-entity embedding
- **#7957** — GraphRAG ran 36 hours / $200 then stalled
- **#9647** — GraphRAG timeout during subgraph generation at ~86%

### Proposed Fix

PR **#15982** (`perf(graphrag): batch entity/relation embeddings in set_graph`) already implements batch embedding. As of 2026-06-19, this PR is still **open** and has not been merged into v0.26.0 or v0.26.1. Merging this PR would resolve the core performance issue.

### Additional Context

The per-entity embedding design also has an unbounded task creation problem — `chat_limiter` only limits internal model calls but does not control the total number of asyncio tasks created in `set_graph()`. This can lead to memory exhaustion on resource-constrained deployments.

zaviermeekz-cpu:

GraphRAG `set_graph()` embeds entities and relations one‑by‑one, causing extreme slowness (10-50x slower) for large graphs (17,000+ edges). This PR batches embedding calls to improve performance.

### Solution

- Add `GRAPHRAG_EMBED_BATCH_SIZE` environment variable (default 64).
- Refactor `graph_node_to_chunk` and `graph_edge_to_chunk` to separate chunk building from embedding application.
- Batch cache misses in groups of `_EMBED_BATCH_SIZE` and encode in a single HTTP call.
- Reuse existing Redis cache to avoid redundant embedding.
- Fix cache key mismatch (check was using "A->B", store was using "A->B: description").
- Add progress callbacks for batch encoding phases.

### Performance impact

- Before: 17,000+ individual HTTP requests for embedding.
- After: ~265 batched requests (17,000 / 64).
- Measured speedup: 10-50x depending on graph size and network latency.

### How has this been tested?

- Verified with a small document (entities/relations indexed correctly).
- Verified with a large document (17,000+ edges) – completion time significantly reduced.
- Redis cache is populated and reused as expected.

### Related Issue

Closes #16205


yingfeng:
Hi, thank you for your contribution!
We wanted to give you a heads-up: GraphRAG is being moved to legacy status and replaced by unified knowledge compilation. As a result, we won't be able to merge this PR. We truly appreciate the effort you put into this, and we hope you'll continue contributing to the project!

PR #16214, opened 2026-06-20
fix(stats): use current_user.id as tenant_id in /system/stats

vincentlambert:

`GET /api/v1/system/stats` always returns empty arrays for users who belong to multiple tenants.

**Root cause**: `UserTenantService.query(user_id=current_user.id)` returns all tenant memberships in arbitrary order. Taking `tenants[0].tenant_id` picks a random membership instead of the user's own tenant. The subsequent JOIN `Dialog.tenant_id == tenant_id` then matches nothing, silently returning empty stats.

A user with team memberships can have 10, 20, or more rows in `user_tenant`. The first row is not guaranteed to be their own primary tenant.

**Fix**: use `current_user.id` directly. In RAGFlow, a user's primary tenant always has the same ID as the user — this is an invariant established at account creation.

```python
# Before (broken for users with multiple tenant memberships)
tenants = UserTenantService.query(user_id=current_user.id)
tenant_id = tenants[0].tenant_id  # arbitrary order

# After
tenant_id = current_user.id  # primary tenant id == user id
```

## Type of change

## Test plan

- [x] Created iframe chatbot sessions via `POST /api/v1/chatbots/{id}/completions`
- [x] Confirmed sessions exist in `api_4_conversation` with correct `dialog_id` and `create_date`
- [x] Called `GET /api/v1/system/stats` as a user with 30+ tenant memberships — returns correct `pv`, `uv`, `round` values
- [x] Verified the underlying SQL directly against MySQL to confirm the JOIN and filters work correctly


JinHai-CN:

PR #16220, opened 2026-06-21
fix(agent): output sys.query from Begin component to downstream nodes (#16218)

### LINKED ISSUE #16218: [Bug] Begin component does not pass {sys.query} to downstream nodes (v0.26.1, UI Chat)
RAGFlow Version: v0.26.1 (also reproduced on nightly)
Deployment: Docker Compose

Actual Behavior:
The Begin component contains the system parameter {sys.query}, but it does not pass the user's query to the output. The Output of the Begin block only includes the service fields _created_time and _elapsed_time. The user's query is missing.

Logs:
Begin Output: {
  _created_time: 1601692.883750541,
  _elapsed_time: 0.00018917699344456196
}
// sys.query is not present

As a result, downstream nodes (Docs QA Agent, Message) receive empty input. The agent cannot process the request and only returns the prologue/greeting message.

Steps to Reproduce:
1. Create an agent from the built-in template (the Begin component already contains {sys.query} by default).
2. Send any query via the UI Chat.
3. In the logs, the Begin Output does not contain the user's query.
4. The Docs QA Agent receives empty input and cannot perform retrieval or generate an answer.

Additional Info:
- No errors appear in ragflow-server.log.
- The issue is reproduced on both v0.26.1 and nightly.
- Tested only via UI Chat (not via API).

Related Issues: #10699, #5418, #4320

zaviermeekz-cpu:

The Begin component contains the system parameter `{sys.query}`, but it does not pass the user's query to the output. Downstream nodes (Docs QA Agent, Message) receive empty input and agents cannot process user requests – they only return the prologue/greeting message.

### Solution

Add `query` to the Begin component's output dictionary, retrieving the value from `self._canvas.globals["sys.query"]`. Downstream components can now reference `{Begin@query}` to access the user's query.

### How has this been tested?

- Created an Agent from a built-in template.
- Sent a query via UI Chat.
- Verified that Begin Output includes the `query` field with the user's query.
- Confirmed that downstream components receive the query and process it correctly.

### Related Issue

Closes #16218


zaviermeekz-cpu:
@coderabbitai Thank you for the feedback. I've:
- Added docstrings to the `Begin` class and its methods (coverage now meets the 80% threshold).
- Added a `logging.debug` statement to trace `sys.query` propagation.
- Imported `logging` as required.

yuzhichang:
## Analysis: the premise of this PR is incorrect

The PR description claims that "downstream nodes (Docs QA Agent, Message) receive empty input and agents only return the prologue/greeting message" because Begin does not propagate `sys.query` to its output. After tracing the actual code paths, this premise does not hold up. The fix is harmless but it solves a problem that does not exist for built-in agents.

### 1. `sys.query` is delivered by the canvas, not by the Begin component

In `agent/canvas.py` (lines 409-414), when a query is submitted the canvas directly populates the global:

```python
for k in kwargs.keys():
    if k in ["query", "user_id", "files", "chat_template_kwargs"] and kwargs[k]:
        ...
        self.globals[f"sys.{k}"] = kwargs[k]
```

The Begin component is **not** on this path. `canvas.globals["sys.query"]` is populated before Begin runs, regardless of what Begin emits.

In `agent/canvas.py` (`get_variable_value`, line 195):

```python
def get_variable_value(self, exp: str) -> Any:
    exp = exp.strip("{}").strip()
    if exp.find("@") < 0:
        return self.globals[exp]      # {sys.query} -> canvas.globals["sys.query"]
```

So `{sys.query}` resolves directly from the canvas global. It does not transit through `Begin.output("query")`.

### 2. Every built-in template uses `{sys.query}`, not `{Begin@query}`

Searching `agent/templates/*.json` for references to the user query:

- `deep_research.json` — multiple `{sys.query}` references
- `market_seo_article_writer.json` — `{sys.query}` references
- `cv_analysis_and_candidate_evaluation.json` — `{sys.query}` references
- `data_analysis_beginner_assistant.json` — `{sys.query}` references
- `customer_feedback_dispatcher.json`, `photo_text_translator.json`, … — same pattern

None of the 25 built-in templates reference `{Begin@query}`. They all read the user query via `{sys.query}`, so they already work correctly without this fix.

### 3. The frontend does not expose `{Begin@query}` either

`web/src/constants/agent.tsx`:

```typescript
export enum AgentGlobals {
  SysQuery = 'sys.query',
  ...
}
export const AgentGlobalsSysQueryWithBrace = `{${AgentGlobals.SysQuery}}`;
```

This is the canonical reference that the prompt editor's variable picker offers for the user query. Meanwhile `web/src/pages/agent/canvas/node/begin-node.tsx` only renders the Begin node's form-defined `inputs` as its output ports — `query` is not among them. So a user following the UI's variable picker will use `{sys.query}`, never `{Begin@query}`.

### 4. The canonical reference is `{sys.query}`; `{Begin@query}` should not be used

This PR would introduce a new convention (`{Begin@query}`) that:

- duplicates an existing, already-correct mechanism (`{sys.query}`),
- is not used by any built-in template,
- is not exposed by the frontend variable picker,
- is not surfaced as a Begin output port.

Promoting `{Begin@query}` as a supported reference pattern risks long-term confusion, because the two references would both happen to work but mean subtly different things (Begin's output slot vs. canvas global), and they would diverge if either side is later changed.

### Recommendation: close this PR

The user's intended outcome — letting downstream nodes see the user's query — is already fully achieved by `{sys.query}`. No code change is needed in Begin. Authors of custom agents who want the user query should reference `{sys.query}`, not `{Begin@query}`.

PR #16222, opened 2026-06-21
refactor(llm): centralize error classification in rag/llm/errors.py

BasilFoubert:

Error classification logic for LLM calls was duplicated across Base and LiteLLMBase in chat_model.py, and the LLMErrorCode enum was defined locally, making it inaccessible to other model types (embedding, rerank, OCR, TTS) that face the same error categories.

### What does this PR do?

- Introduces `rag/llm/errors.py` with `ModelErrorCode` and `classify_model_error(error) -> ModelErrorCode`
- Removes `LLMErrorCode` and both `_classify_error` wrapper methods from `chat_model.py`
- Calls `classify_model_error` directly at all former call sites

This is a pure code move: the classification logic (keyword lists, matching order, fallback behavior) is copied as-is from the original `_classify_error` methods. **No behavior change is intended.**

### Motivation

This refactor was initially motivated by a bug in `verify_api_key` where quota errors were misclassified as invalid keys, causing API key validation to take up to 2 minutes. That bug has since been fixed independently. The refactor still stands on its own: it eliminates the duplicated `_classify_error` logic in `chat_model.py` and makes `classify_model_error` accessible outside that module, so future error handling across all model types (embedding, rerank, OCR, TTS) can use a consistent, centralized classifier rather than ad-hoc string checks.

### Type of change

- [ ] Documentation Update
- [X] Refactoring
- [ ] Performance Improvement


Lynn-Inf (commented):
What this does: it moves the error taxonomy out of `chat_model.py` into a new `rag/llm/errors.py` — `LLMErrorCode` becomes `ModelErrorCode`, and the two byte-identical `_classify_error` methods on `Base` and `LiteLLMBase` collapse into a single module-level `classify_model_error()`. The keyword lists, matching order and `ERROR_GENERIC` fallback are carried over unchanged, and neither `LLMErrorCode` nor `_classify_error` has any reference outside `chat_model.py`, so dropping them is safe.Both the dedup and the shared module are worth having.

One blocker, though: this also drops `import re`, and `re` is still used in `chat_model.py`.
```

      _MINIMAX_CONTROL_TOKEN_RE = re.compile(...)   # module level
      class _StreamSanitizer:
          _TOKENISH = re.compile(...)               # class body, so also import time
          def __init__(self, pattern: re.Pattern, ...)
  
```
Both `re.compile` calls execute at import, so `rag/llm/chat_model.py` now fails to import with `NameError: name 're' is not defined`. `ruff check --select F821 rag/llm/chat_model.py` reports 4 undefined-name errors. Please restore the import — `re` is still genuinely needed here, so the classifier itself can move while the import stays.

Two smaller things worth noting. The new code joins keywords with `re.escape`, which the original did not: it is a no-op for the current list (every keyword is plain alphanumerics and spaces) and it is the safer form, so I would keep it, but it is a change rather than a pure move, so it does not belong under "no behavior change is intended". The two added `logging.debug` calls are fine.

Lastly, the validation in the description only runs `ruff check` over `sequence2txt_model.py`; including `chat_model.py` would have caught this immediately.

PR #16230, opened 2026-06-22
test(api): cover the shared REST pagination helpers

### LINKED ISSUE #15657: [Feature Request]: centralize and harden REST API pagination parsing
### Self Checks

The REST API handlers under `api/apps/restful_apis/` repeat the same pagination-parsing snippet in ~13 places:

```python
page = int(request.args.get("page", <default>))
page_size = validate_rest_api_page_size(int(request.args.get("page_size", <default>)))
```

This has two issues:

1. **Inconsistent / unsafe parsing.** `int(request.args.get("page", ...))` raises an unhandled error on non-integer input and silently accepts a **negative** `page`/`page_size` (e.g. `?page=-5`), which is then passed straight to the service/DB layer. `validate_rest_api_page_size` only checks the **upper** bound, not that the value is a valid non-negative integer.
2. **Duplication.** The same two-line idiom is copy-pasted across `agent_api`, `chat_api`, `chunk_api`, `connector_api`, `dataset_api`, `document_api`, `mcp_api`, `memory_api`, and `search_api`, so the rules cannot be changed in one place.

Add small shared helpers in `api/utils/pagination_utils.py` and adopt them at every REST pagination call site:

- `parse_page_number(args, default=1)` — parse/validate the `page` argument.
- `parse_page_size(args, default=30)` — parse `page_size` and validate it against `REST_API_MAX_PAGE_SIZE`.
- `parse_pagination(args, *, default_page, default_page_size)` — return a validated `(page, page_size)` tuple in one call.

Each rejects non-integer and negative values by raising `ValueError` (the same exception type the call sites already surface, so existing error handling is preserved), while keeping `0` valid as the existing "all results" sentinel. Per-call-site defaults are preserved exactly, so behavior is unchanged for valid input.

Centralize the logic in `pagination_utils.py`, replace the duplicated two-line idiom at each site with a single `parse_pagination(...)` (or `parse_page_number`/`parse_page_size`) call preserving that site's defaults, and add unit tests under `test/unit_test/api/utils/`.

manan-tech:
The centralisation this PR proposed is already on `main`: #17597 introduced `validate_rest_api_page` / `validate_rest_api_page_size` and migrated the REST endpoints, and #18074 added `validate_rest_api_ids`. I re-checked every `api/apps/restful_apis/*.py` — all 12 call sites go through the helpers, with no raw `int(request.args.get(...))` left. So the feature side of #15657 is done, and this branch's version of it only conflicted with the merged design (it raised on invalid input where `main` deliberately falls back).

What `main` doesn't have is any test for the module. The coverage that exists is only for the *raising* half of the contract, and both cases reach it through another layer: `test_doc_validation.py` exercises the `page_size` cap, and `test_list_memory_filters.py` the ids cap through a route.

The fallback half is the part every list endpoint leans on — `?page=abc` and `?page=-5` have to resolve to the default rather than raise or reach the service layer as a negative offset — and nothing pins it today. As it stands, the asymmetry (fall back below the range, raise above the cap) could be "tidied" into raising, or the clamp dropped, without a single test failing.

**Added:** `test/unit_test/api/utils/test_pagination_utils.py` covering all three helpers — usable values pass through, garbage and out-of-range values fall back to the defaults, values over the public caps raise, and the ids error names the offending field.

Mutation-checked rather than assumed: dropping the `< 1` clamp from `validate_rest_api_page` and moving the ids cap off by one makes 6 of these fail; the module as it stands passes 29/29. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/api/apps/restful_apis/agent_api.py
+++ b/api/apps/restful_apis/agent_api.py
@@ -60,7 +60,7 @@ from api.utils.api_utils import (
     server_error_response,
     validate_request,
 )
-from api.utils.pagination_utils import validate_rest_api_page_size
+from api.utils.pagination_utils import parse_pagination
 from common import settings
 from common.ssrf_guard import assert_host_is_safe
 from common.constants import RetCode
@@ -363,8 +363,7 @@ async def _run_workflow_session(
 def list_agent_sessions(agent_id, tenant_id):
     session_id = request.args.get("id")
     user_id = request.args.get("user_id")
-    page_number = int(request.args.get("page", 1))
-    items_per_page = validate_rest_api_page_size(int(request.args.get("page_size", 30)))
+    page_number, items_per_page = parse_pagination(request.args, default_page=1, default_page_size=30)
     keywords = request.args.get("keywords")
     from_date = request.args.get("from_date")
     to_date = request.args.get("to_date")
@@ -586,8 +585,7 @@ def list_agents(tenant_id):
     owner_ids = [item for item in request.args.get("owner_ids", "").strip().split(",") if item]
     tags = [item for item in request.args.get("tags", "").strip().split(",") if item]
 
-    page_number = int(request.args.get("page", 0))
-    items_per_page = validate_rest_api_page_size(int(request.args.get("page_size", 0)))
+    page_number, items_per_page = parse_pagination(request.args, default_page=0, default_page_size=0)
     order_by = request.args.get("orderby", "create_time")
     desc = str(request.args.get("desc", "true")).lower() != "false"
     tenants = TenantService.get_joined_tenants_by_user_id(tenant_id)
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -45,7 +45,7 @@ from api.utils.api_utils import (
     server_error_response,
     validate_request,
 )
-from api.utils.pagination_utils import validate_rest_api_page_size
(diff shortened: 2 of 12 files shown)


manan-tech:
@coderabbitai review

JinHai-CN:
@manan-tech would you please resolve the conflicts?

manan-tech:
@JinHai-CN conflicts resolved, though as with #16262 the rebase found nothing left of the original change: #17597 landed the shared `validate_rest_api_page` / `validate_rest_api_page_size` helpers and migrated the endpoints, and #18074 added `validate_rest_api_ids`. All 12 REST call sites use them now — I checked, no raw `int(request.args.get(...))` remains — so #15657's feature work is done. This branch's conflicts were design conflicts: it raised on invalid input, where `main` chose to fall back to the defaults.

`main`'s behaviour is the better one for query-string input, so I've dropped this branch's version entirely and reduced the PR to the gap that's left: the module has no tests. Only the raising half of its contract is covered today, and both cases go through another layer (`test_doc_validation.py` for the page_size cap, `test_list_memory_filters.py` for the ids cap). The fallback half — the half every list endpoint depends on — isn't pinned anywhere, so the clamp could be dropped or the asymmetry "simplified" into raising with nothing failing.

Now a single test file covering all three helpers, mutation-checked (removing the `< 1` clamp and moving the ids cap off by one fails 6 of them). Retitled and description rewritten. Happy to close instead if you'd rather not carry it.

PR #16233, opened 2026-06-22
Add Keenable web search tool to the agent

ilya-bogin-keenable:
Adds Keenable as a web search tool in the agent, alongside the existing
Tavily/DuckDuckGo/SearXNG/Google tools.

The main difference from the other search tools is that it doesn't require an
API key. By default it uses Keenable's keyless public endpoint, so it works out
of the box. Providing a key (in the tool config) switches to the authenticated
endpoint and lifts the rate limits.

### Changes

- Backend: `agent/tools/keenable.py` — `KeenableSearch`, follows the
  Tavily/DuckDuckGo tool shape (results go through `_retrieve_chunks`).
  Auto-registered by `agent/tools/__init__.py`.
- Frontend: wired into the agent builder — operator + icon, config form
  (optional API key, search mode, site filter, top N), the search tool menu,
  and the existing api_key export sanitizer.

### Config

- API key: optional. Blank = keyless free tier; set it to lift limits / enable
  `realtime` mode.
- `site`: restrict to a single domain.
- `mode`: `pro` (default) or `realtime`.

### Notes

`KEENABLE_API_URL` can override the API base (HTTPS enforced; defaults to
`https://api.keenable.ai`). The tool only sends the query (no URL fetch), so
there's no SSRF surface. Verified the frontend with `vite build` and the
backend search path against the public endpoint.


ilya-bogin-keenable:
Thanks for the review. Addressed in the latest push:

- **`check()` realtime contract** — now rejects `realtime` mode without an API key at config time (realtime isn't available on the keyless endpoint).
- **Non-error logging** — added start/success/cancel logs in `_invoke` (no query text or keys logged).
- **Retry on non-transient errors** — config errors (`ValueError`, e.g. a bad `KEENABLE_API_URL`) now fail fast instead of sleeping and retrying.
- **`top_n` type** — form schema is `z.coerce.number()`, matching the numeric default and `TopNFormField` (fixes the cross-file mismatch flagged in `constant/index.tsx` too).

On the `useTranslate` note in `keenable-form/index.tsx`: kept as-is — `useTranslate('flow')` matches the existing agent forms (e.g. the DuckDuckGo form).

yuzhichang (commented):
LGTM

PR #16242, opened 2026-06-22
fix(api): allow setting language on dataset creation (#15703)

### LINKED ISSUE #15703: [Question]: Missing language parameter in POST /api/v1/datasets API causes poor non-English RAG performance
### Self Checks

### Describe your problem

Why is it not possible to specify the language when creating datasets via the HTTP API?
Specifically, the POST /api/v1/datasets endpoint does not have a language parameter. This leads to extremely poor knowledge-based Q&A performance when the target language is non-English.
Environment:
RAGFlow Version: 0.25.6
Deployment: Docker

manan-tech:

Closes #15703.

`POST /api/v1/datasets` (RESTful create-dataset endpoint) has no way to set the document `language`. The `Knowledgebase` model has a `language` column and `UpdateDatasetReq` already exposes the field, but `CreateDatasetReq` omitted it — so every dataset is created with the server default regardless of the request, which degrades retrieval quality for non-English corpora (users then have to make a second `PUT` call to fix it).

### How does this PR fix it?

- Add `language` to `CreateDatasetReq`, and let `UpdateDatasetReq` inherit it instead of redeclaring the identical field (single source of truth).
- The create request is parsed with `exclude_unset=False`, so the key is always present (defaulting to `None`) even when the caller omits it. To avoid clobbering the model/database default (`"English"`/`"Chinese"`) with `NULL`, `create_dataset` now drops `language` when it is `None`. An explicitly provided value flows straight through the existing `KnowledgebaseService.create_with_name(**req)` path.
- `language` is validated with `StringConstraints(strip_whitespace=True, min_length=1, max_length=32)` so blank/whitespace-only values are rejected instead of silently bypassing the `None` guard.
- Document the field in the create endpoint's OpenAPI/Swagger spec (`maxLength: 32`, matching the actual validation).

### Type of change

- [x] Documentation Update

### Behavior

| Request | Before | After |
| --- | --- | --- |
| `{"name": "kb", "language": "Chinese"}` | `language` ignored, falls back to default | dataset created with `Chinese` |
| `{"name": "kb"}` (omitted) | server default | server default (unchanged) |
| `{"name": "kb", "language": ""}` | stored blank | rejected with a validation error |

### Tests

Added unit tests in `test/unit_test/api/utils/test_doc_validation.py` covering: explicit value accepted, omitted → `None` (service falls back to the DB default), blank/whitespace rejected, surrounding whitespace stripped, `max_length=32` enforced, and `UpdateDatasetReq` still exposing the inherited field. Full file passes (31 passed); `ruff check` is clean.


manan-tech:
Thanks @coderabbitai — addressed all four in `4dd90f7`:

1. **OpenAPI enum mismatch** (`dataset_api.py`): dropped the `English`/`Chinese` enum in favor of `maxLength: 32` so the spec matches the actual backend validation (any value ≤32 chars), keeping the two as examples in the description.
2. **Missing log on new flow** (`dataset_api_service.py`): added a `logging.debug` when `language` is omitted and the model/DB default is applied.
3. **Blank/whitespace bypass** (`validation_utils.py`): `language` now uses `StringConstraints(strip_whitespace=True, min_length=1, max_length=32)`, so `""`/whitespace-only is rejected (can't slip past the service `None`-guard) and surrounding whitespace is trimmed.
4. **Pytest priority markers**: added `@pytest.mark.p2` to the new tests, plus coverage for blank rejection and whitespace stripping.

All 31 tests in `test/unit_test/api/utils/test_doc_validation.py` pass and `ruff check` is clean.

6ba3i (commented):
###  Go still has the same issue

`POST /api/v1/datasets` also exists in the Go backend, but `CreateDatasetRequest` does not expose `language`, and `DatasetService.CreateDataset` never sets it.

So this PR fixes #15703 in Python, but the same request will still fail to set the language when the endpoint is served by Go.

I think we should add the equivalent Go support here as well:

* add `Language *string` to `CreateDatasetRequest`;
* trim/validate it consistently with Python;
* set `Knowledgebase.Language` when provided;
* leave it unset when omitted so the existing default applies;
* add focused tests for explicit and omitted language.

6ba3i:
could you please fix the issues raised in my comment ? @manan-tech

6ba3i:
@manan-tech please resolve conflict

manan-tech:
Thanks @6ba3i — on it. Merging `main` and resolving the conflict now (it's in `api/apps/services/dataset_api_service.py`, where `ext` was removed by #19272), and working through your Go comment. Status so far:

**Go** — `CreateDatasetRequest.Language`, the trim/32-char check and `Knowledgebase.Language` landed upstream today in #19272, so that part of your list is already in `main`. What's still missing is the piece you asked for last: focused tests. I'm adding them to `internal/service/dataset/create_test.go` (explicit language, omitted language falling back to the column default, blank rejected, length limit), and aligning the validation with Python — Python now strips and rejects blank/whitespace-only values, and counts characters rather than bytes, which matters here since the issue is about non-English datasets.

**Python** — after the merge, the remaining delta over `main` is: the `language is None` guard in `create_dataset` (the request is parsed with `exclude_unset=False`, so an omitted language currently reaches `create_with_name` as an explicit `None` and writes NULL over the `Knowledgebase.language` default), the blank/whitespace rejection, the OpenAPI entry for `language` on `POST /datasets`, and tests for both layers.

Will push and follow up once the Go tests are running green.

manan-tech:
Pushed `32a83ad` — conflict resolved and the Go side addressed. @6ba3i

**Merge conflict.** Two spots, both from #19272: `dataset_api_service.py` (the `ext` handling it removed — kept the `language is None` guard on top of the new body) and `validation_utils.py`, where #19272 added its own `language` field to `CreateDatasetReq` that duplicated this branch's. Kept one declaration, with the stricter constraints.

**Go.** #19272 landed the plumbing you listed — `Language *string` on `CreateDatasetRequest`, trim + length check, `Knowledgebase.Language` set, left nil when omitted so the column default applies. Two gaps were left, both now covered:

- *Tests* (`internal/service/dataset/create_test.go`): explicit language is trimmed and stored; an omitted language leaves the column default in place; blank/whitespace-only is rejected; the length limit is enforced.
- *Validation parity.* Python strips and rejects blank/whitespace-only values (`min_length=1`) and, being pydantic, counts **characters**; the Go path trimmed but accepted `""`, and `len()` counts **bytes**, so a 32-character non-ASCII language name (96 bytes) was rejected in Go while Python accepts it — which is precisely the non-English case #15703 is about. Factored into `normalizeDatasetLanguage()` and used by both `CreateDataset` and `UpdateDataset`. The error strings match the ones the SDK tests already expect for blank fields (`String should have at least 1 character`).

One judgement call to flag: I applied the helper to `UpdateDataset` as well, since Python's `UpdateDatasetReq` inherits the field from `CreateDatasetReq` and would otherwise diverge from Go on the update path too. Happy to pull that back out if you'd rather keep this PR strictly create-only.

**Python delta over `main`.** The structural part is upstream now, so what's left here is: the `language is None` guard in `create_dataset` (requests are parsed with `exclude_unset=False`, so an omitted language currently reaches `create_with_name` as an explicit `None` and writes NULL over the `Knowledgebase.language` default), the blank rejection, the OpenAPI entry for `POST /datasets`, and tests for both layers.

**Verification**

```
go test ./internal/service/dataset/... -count=1          # ok (0.245s → full package 1.07s)
pytest test/unit_test/api/utils/test_doc_validation.py   # 35 passed
pytest test/unit_test/api/apps/services/test_dataset_api_service_create_language.py   # 2 passed
ruff check / ruff format --check                         # clean
```

The Python service test was checked against the unfixed code as well: without the `None` guard, `create_with_name` is called with `language=None` and the test fails, so it's a real regression anchor rather than a tautology.

PR #16245, opened 2026-06-22
fix(llm): stop forwarding `images` kwarg to the chat completion SDK (#15966)

### LINKED ISSUE #15966: [Question]: AsyncCompletions.create() got an unexpected keyword argument 'images'
### Self Checks

### Describe your problem

While Summarizing a pdf document using Qwen/Qwen2.5-VL-7B-Instruct i get ERROR: AUTH_ERROR - AsyncCompletions.create() got an unexpected keyword argument 'images'. What could be possible reasons? Any issue in Assistant Configuration?

manan-tech:

Closes #15966.

Attaching an image to a chat backed by a base (OpenAI-compatible) chat model crashes with:

```
AsyncCompletions.create() got an unexpected keyword argument 'images'
```

`dialog_service` passes `images=...` to `LLMBundle.async_chat` / `async_chat_streamly[_delta]` for models whose type is not `"chat"`. Base chat models (`rag/llm/chat_model.py:Base`) only declare `**kwargs` and forward everything to `chat.completions.create()`, so the `images` kwarg leaks straight into the provider SDK and raises. (CV / image2text models in `rag/llm/cv_model.py` declare an explicit `images=` parameter and handle it natively, so they were never affected — they use a separate `Base` class.)

### How does this PR fix it?

In `LLMBundle` (the single entry point all chat callers go through), when `images` is passed but the target model does **not** declare an explicit `images` parameter:

1. base64-encode the raw image blobs, and
2. fold them into the last user message as multimodal content via the existing, provider-aware `convert_last_user_msg_to_multimodal` helper, then
3. drop the `images` kwarg so it never reaches the SDK.

This fixes the crash **and** lets OpenAI-compatible vision chat models (e.g. Qwen2.5-VL) actually receive the image. It is a no-op for normal text chat and for CV models that consume `images` natively, keeping the blast radius minimal.

### Type of change

### How to test

**Automated unit tests** (no live model needed):

```
PYTHONPATH=$(pwd) .venv/bin/python -m pytest test/unit_test/api/db/services/test_llm_bundle_image_kwarg.py -v
```

Covers: base chat model folds images + drops the kwarg; CV model with a native `images=` param is left untouched; empty/no images is a no-op; raw bytes are base64-encoded.

**End-to-end (before/after):** configure a vision model (e.g. `Qwen/Qwen2.5-VL-7B-Instruct` via an OpenAI-compatible provider), open a chat assistant, attach an image, and ask "summarize this image".

- Before: the chat returns the `unexpected keyword argument 'images'` error.
- After: the model returns a normal description/summary of the image.

### Tests

`test/unit_test/api/db/services/test_llm_bundle_image_kwarg.py` (4 cases, `p2`) — all pass; `ruff check` clean.


manan-tech:
Thanks @coderabbitai — fixed in `4e65a75`. The `inspect.signature()` fallback now logs a warning before returning, so the case where a model's signature can't be inspected (and `images` is left in kwargs) is visible instead of silent. Tests still pass (4/4) and `ruff check` is clean.

manan-tech:
## Verification — before / after

Reproduced against the **real** code path: real `rag/llm/chat_model.py`, the real `openai 2.41.0` SDK, and the real `LLMBundle.async_chat` — driven against unpatched `main` vs this branch via git worktrees. Only the model's HTTP endpoint is a local stub, so whether it breaks is decided purely by the `LLMBundle` code on the path.

### Before (`main`, unpatched)

The `images` kwarg is forwarded straight to the SDK and the call dies before any request is sent — the exact error from the issue:

```
TypeError: AsyncCompletions.create() got an unexpected keyword argument 'images'
  File "rag/llm/chat_model.py", line 628, in _async_chat
    response = await self.async_client.chat.completions.create(model=..., messages=history, **gen_conf, **kwargs)

answer returned to caller:
  "**ERROR**: ... AsyncCompletions.create() got an unexpected keyword argument 'images'"
-> NO HTTP request reached the model server (SDK rejected the call locally)
```

image

### After (this PR)

Images are folded into the last user message as multimodal content and the kwarg is dropped, so the request is well-formed and the model answers normally:

```
answer returned to caller:
  'This image shows a cat on a windowsill.'

HTTP request reached the model server: /v1/chat/completions
  'images' present at top level of request body? False
  last message content sent:
    [{"type": "text", "text": "summarize this image"},
     {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}}]
```

image

### How to reproduce

```bash
git fetch origin main
git worktree add --detach /tmp/rf-before origin/main
git worktree add --detach /tmp/rf-after  <this-branch>
# run the same repro script in each (real chat model + openai SDK + local stub endpoint)
```

Unit coverage: `test/unit_test/api/db/services/test_llm_bundle_image_kwarg.py` (4 cases) — base model folds images + drops kwarg, CV model with native `images=` untouched, empty/no-images no-op, raw bytes base64-encoded.

Lynn-Inf (commented):
Thanks for the PR, but I think this one can be closed — #15966 is already fixed on `main`.

PR #16246, opened 2026-06-22
fix: derive image2text model type from tags in _factory_model_types

vincentlambert:
## Problem

42 models in `conf/llm_factories.json` (including `gpt-5.5`, `gpt-5.4`, `claude-opus-*`, `qwen3.5-*`, etc.) carry `IMAGE2TEXT` in their `tags` field but declare only `"model_type": "chat"`.

When `get_model_config_from_provider_instance` falls through to the factory lookup — i.e. the user has configured the model as their vision model (`img2txt_id`) but no explicit row exists in `tenant_model` for it — `_factory_model_types` only reads `model_type`, producing a `["chat"]` list, and then raises:

```
LookupError: Model gpt-5.5@Alice@OpenAI is not a image2text model.
```

This prevents any vision/PDF-parsing workflow from using these models.

## Fix

In `_factory_model_types`, if `"IMAGE2TEXT"` is present in the model's `tags` string and `"image2text"` is not already in the derived types list, append it. This is consistent with the intent expressed in `tags` and avoids having to patch 42 individual JSON entries.

```python
def _factory_model_types(llm: dict) -> list[str]:
    model_type = llm.get("model_type")
    types: list[str] = model_type if isinstance(model_type, list) else ([model_type] if model_type else [])
    # Some factory entries carry IMAGE2TEXT in tags but omit it from model_type — derive it.
    if "IMAGE2TEXT" in llm.get("tags", "").upper() and "image2text" not in types:
        types = list(types) + ["image2text"]
    return types
```

## Affected models (42 total)

OpenAI: `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano`, `gpt-5.2-pro`, `gpt-5.2`, `gpt-5.1`, `gpt-5.0`, `gpt-4.5`, `gpt-4o`, `gpt-4o-mini`, `gpt-4-turbo`, `gpt-4-vision-preview` …  
Anthropic/Claude: `claude-opus-4-5`, `claude-sonnet-4-5`, `claude-3-7-sonnet-*`, `claude-3-5-sonnet-*` …  
Qwen: `qwen3.5-*`, `qwen-vl-*` …  
and others.

## Test

Verified locally: before the fix, configuring any of these models as the tenant vision model (`img2txt_id`) and triggering PDF parsing with vision raises `LookupError`. After the fix, the model resolves correctly.


JinHai-CN:

PR #16249, opened 2026-06-23
Preserve GraphRAG subgraph checkpoints during merge

stevenniuniu:
## What

- Preserve extracted-but-not-yet-merged GraphRAG subgraph checkpoints during global graph merge.
- Replace the dataset-wide graph/subgraph delete with a graph-only delete plus targeted subgraph replacement for documents already represented by the merged graph.
- Add a regression test to ensure set_graph no longer deletes all subgraph records.

## Why

- Long GraphRAG runs save per-document subgraphs after LLM extraction. During merge, set_graph previously deleted all graph and subgraph records, then recreated only subgraphs represented in the current merged global graph. If merge timed out or failed after expensive extraction had completed, extracted-but-unmerged subgraph checkpoints could be lost, forcing LLM extraction to run again.

## Validation

```
python3 -m py_compile rag/graphrag/utils.py test/unit_test/rag/graphrag/test_graphrag_utils.py
git diff --check
Full pytest was not available in my local environment.
```

### DIFF
--- a/rag/graphrag/utils.py
+++ b/rag/graphrag/utils.py
@@ -517,25 +517,29 @@ async def get_graph(tenant_id, kb_id, exclude_rebuild=None):
 async def set_graph(tenant_id: str, kb_id: str, embd_mdl, graph: nx.Graph, change: GraphChange, callback):
     global chat_limiter
     start = asyncio.get_running_loop().time()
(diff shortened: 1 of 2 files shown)


yingfeng:
Thanks, graphrag is not used any more

PR #16260, opened 2026-06-23
refactor(log): unify Go logging to zap with rotation, strip per-package levels

yuzhichang:
Refactor the Go agent port's logging so every log line — gin access, agent canvas events, harness warnings, fatal boot errors — flows through a single `common.Logger` (zap) backed by a rotated file, with structured fields, level filtering, and configurable rotation.

### DIFF
--- a/cmd/admin_server.go
+++ b/cmd/admin_server.go
@@ -74,7 +74,7 @@ func main() {
 	}
 
 	// Initialize logger
-	if err := common.Init("info", "admin_server.log"); err != nil {
+	if err := common.Init("info", common.FileOutput{Path: "admin_server.log"}); err != nil {
 		panic("failed to initialize logger: " + err.Error())
 	}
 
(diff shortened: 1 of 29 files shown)


yuzhichang:
Closing this PR — see #16261 for the follow-up. Same head branch (`yuzhichang:agent-go-port`), same parent commit (`296c0b733 Unified log`), with `dfd946d54` adding the 5 CodeRabbit review fixes on top.

## CodeRabbit review resolutions (all 5 applied in `dfd946d54`)

| File | Review comment | Resolution |
|------|---------------|------------|
| `cmd/server_main.go:330` | Duplicate `common.GinLogger()` registration (also registered inside `router.Setup`) | Removed local registration; `Setup` owns it. Replaced with an inline comment explaining why. |
| `internal/agent/component/llm_credentials.go:68` + `:124-128` | `zap.String("api_key", apiKey)` leaks API keys | Replaced with `zap.Bool("api_key_present", apiKey != "")` in both `tenant_llm OK` and `new-table OK` debug log sites. |
| `internal/agent/tool/code_exec.go:159` + `:222-229` | Raw `script` / `arguments` / `stdout` / `stderr` may carry credentials | Bound to `script_len`, `arguments_keys`, `stdout_len`, `stderr_len` + `stderr_present` flag. |
| `internal/router/router.go:126` (`common.GinLogger`) | Logs raw query string (OAuth codes / SAML responses / signed state) | Drop query from the `path` field; record `has_query` (bool) and `query_len` (int) instead so operators can still see that one was sent. |
| `internal/handler/system.go:196` | Doc comment claimed `pkg_name` was required, but binding only enforces `level` | Corrected the comment. The `"pkg_name and level are required"` error message text is preserved verbatim for backward compatibility with clients that match on it. |

## Verification

```
bash build.sh --go    # all 4 binaries built
bash build.sh --test  # all packages pass; zero failures
```

## Re: the force-push race

I noticed #16260 was closed at 2026-06-23T06:58:13Z with a `head_ref_deleted` event around the same time my `git push` got "stale info" — likely a race between my push attempt and the branch deletion. GitHub's API then refused to reopen the PR (HTTP 422: "state cannot be changed. The agent-go-port branch was force-pushed or recreated"), and the `update-branch` endpoint refused because the stored head SHA (`99a81e4ce`) no longer matched the fork tip. Opening a fresh PR was the cleanest way out.

PR #16280, opened 2026-06-23
fix: break pagination loop in _get_document_metadata_cache on empty docs page

### LINKED ISSUE #16248: MCP document metadata cache can loop forever when /documents returns an empty docs page
### Describe the bug

In the RAGFlow MCP server, `_get_document_metadata_cache()` can enter an infinite loop while building document metadata for datasets.

When the endpoint:

```text
GET /api/v1/datasets/{dataset_id}/documents?page=1
```

returns a successful response with an empty `data.docs` list, the loop does not advance `page` and does not break. As a result, the MCP server keeps requesting the same `/documents?page=1` endpoint indefinitely.

This caused MCP retrieval calls to hang until the client timed out.

### Affected code

File:

```text
mcp/server/server.py
```

Function:

```python
_get_document_metadata_cache()
```

The problematic pattern is:

```python
while page:
    docs_res = await self._get(f"/datasets/{dataset_id}/documents?page={page}", api_key=api_key)
    if not docs_res:
        break
    docs_data = docs_res.json()
    if docs_data.get("code") == 0 and docs_data.get("data", {}).get("docs"):
        ...
        page += 1
        if docs_data.get("data", {}).get("total", 0) - page * page_size <= 0:
            page = None
```

If `code == 0` but `data.docs` is empty, execution skips the inner block, leaving `page` unchanged. The same page is then fetched again forever.

### Observed behavior

In a self-hosted RAGFlow v0.26.0 deployment, an MCP retrieval request triggered repeated document-list requests for the same dataset page.

Observed from server logs:

```text
GET /api/v1/datasets/.../documents?page=1
GET /api/v1/datasets/.../documents?page=1
GET /api/v1/datasets/.../documents?page=1
...
```

In one run, this produced tens of thousands of repeated `/documents` requests within a few minutes, and the MCP retrieval call timed out after 120 seconds.

### Expected behavior

If the documents API returns a successful response with an empty `docs` list, the MCP server should stop paginating for that dataset instead of requesting the same page again.

### Suggested fix

Handle non-zero response codes and empty document pages explicitly, and include `page_size` in the request so pagination math is consistent.

Example local hotfix:

```python
docs_res = await self._get(
    f"/datasets/{dataset_id}/documents?page={page}&page_size={page_size}",
    api_key=api_key,
)
if not docs_res:
    break

docs_data = docs_res.json()
if docs_data.get("code") != 0:
    break

data = docs_data.get("data", {}) or {}
page_docs = data.get("docs") or []
if not page_docs:
    break

for doc in page_docs:
    ...

if page * page_size >= data.get("total", len(doc_id_meta_list)):
    page = None
else:
    page += 1
```

After applying this locally, MCP retrieval no longer loops and returns normally.

### Related issues / PRs checked

I found related MCP page-size work, but not this exact empty-documents-page loop:

- #16146
- #16148
- #16056

Those appear to address dataset discovery / page-size limits, while this issue is specifically about document metadata pagination when `data.docs` is empty.

### Environment

- RAGFlow version: v0.26.0
- Deployment: self-hosted Docker
- Component: MCP server

OSHA-B:
## Summary

Fixes #16248

## Problem

`_get_document_metadata_cache()` in `mcp/server/server.py` can loop forever when the `/documents` API returns a successful response (`code == 0`) with an empty `data.docs` list.

The `while page` loop only advances `page` inside the `if docs_data.get('code') == 0 and docs_data.get('data', {}).get('docs')` block. When that condition is false (empty page), execution falls through with `page` unchanged and the same page is requested indefinitely.

Observed: tens of thousands of identical `GET /datasets/.../documents?page=1` requests until client timeout.

## Fix

Add `else: break` after the `if` block so any response that yields no documents — whether an empty page or a non-zero code — terminates pagination immediately.

Only `mcp/server/server.py` is touched; 3 lines added at the exact loop site.


OSHA-B:
Added a `logging.debug` statement in the `else: break` path that captures `dataset_id`, `page`, and the response code — as suggested. The change is in the latest push.

OSHA-B:
Gentle ping — this has been open for 5 days. Happy to make any changes if there are concerns. The fix is a single added `break` when the metadata cache fetch returns an empty page, preventing an infinite loop that was reported in #16248.

JinHai-CN:
@OSHA-B would you please fix the conflicts?

PR #16283, opened 2026-06-23
Fix: thumbnail/highlight position offset when using MinerU on wide tables

### LINKED ISSUE #16268: [Bug] Thumbnail/highlight position offset when using MinerU OCR engine on wide tables (built on v0.25.1)
### Environment
- **Data built on**: v0.25.1
- **Query version**: v0.26.2
- **OCR Engine**: MinerU

### Description

When using the MinerU OCR engine to build the knowledge base index, the **thumbnail preview** and **highlight/citation overlay** in the "Reference" dialog have a significant positional offset. The offset appears to be multiplied by a fixed scaling factor, as if bounding box coordinates are not being correctly converted to display coordinates.

### Reproduction Steps

1. Use MinerU as the OCR engine to index a PDF that contains **large/wide tables** (e.g., spanning ~3 pages with 40+ rows).
2. Query the knowledge base and open any citation reference.
3. Observe the thumbnail preview and highlight overlay in the reference dialog.

### Expected Behavior

- Thumbnail should correctly show the cited page region.
- Highlight overlay should accurately cover the cited text/cells.

### Actual Behavior

1. **Thumbnail offset**: The thumbnail preview is compressed and shifted, showing the wrong page region (e.g., page header instead of the actual table content).
2. **Highlight/font mismatch**: The blue highlight box is displaced to the **page header** area instead of the actual cited table rows. The highlight dimensions do not match the font size at that position.

The issue is **most pronounced on large wide tables** spanning multiple pages. On a ~3-page table with 40+ rows, the thumbnail and highlight shift all the way to the header position.

### Root Cause Hypothesis

The coordinate system used by MinerU for bounding boxes likely differs from what RAGFlow expects (e.g., different DPI assumptions, page height reference point, or coordinate normalization). The offset appears proportional to document size, suggesting a missing or incorrect scaling coefficient when converting MinerU bbox coordinates to display coordinates.

### Screenshots

> Screenshots showing the offset (to be attached):
>
> - **Screenshot 1**: Citation reference dialog — thumbnail is compressed/offset to wrong region
> - **Screenshot 2**: Wide table PDF — highlight box displaced to page header area, completely missing the actual cited table rows

### Additional Context

- The problem is **not noticeable** on short/regular text documents.
- It is **consistently reproducible** with MinerU-parsed wide/tall multi-page tables.
- Tested with index built on v0.25.1, queried on v0.26.2.

minion1227:

Fixes #16268 — when MinerU is the OCR engine, citation highlights and thumbnail previews for cross-page tables are badly offset: the highlight box lands in the page-header area and the thumbnail collapses to a small region, missing the actual table rows. The problem is most severe on large/wide tables spanning multiple pages and is not noticeable on regular text.

Root cause. MinerU merges a cross-page table's rows into a single content_list.json item but keeps only the first fragment's bbox/page_idx (upstream behavior, opendatalab/MinerU#3561). The merged chunk therefore carries all the table text but a bounding box covering only a slice of one page, so RAGFlow — faithfully rendering that single bbox — highlights one page and misses the rest. The existing coordinate scaling in _line_tag is correct (regular documents are unaffected); the issue is the missing per-page span for merged tables.

This PR makes two changes in deepdoc/parser/mineru_parser.py:

Reconstruct per-page positions for cross-page tables. MinerU's middle.json (already requested via return_middle_json) still contains the individual per-page table fragments. The parser now reads it, builds a per-page table-fragment index (normalizing middle.json's absolute bboxes into the same 0–1000 space content_list uses), and for a table block emits one highlight tag per page the table spans. The highlight/thumbnail now cover the whole table instead of a single-page sliver.

Harden coordinate conversion fallback. _line_tag previously emitted raw 0–1000 bbox values (not display coordinates) when a rendered page image was unavailable. __images__ now also captures each page's size, and conversion falls back to the captured size, then to a default page size — so highlight coordinates are always in display space, never raw normalized values.

Both paths are fully gated and wrapped in defensive error handling: when middle.json is absent, unreadable, or has an unexpected schema, behavior falls back to the previous single-fragment output. Non-table and single-page-table blocks produce byte-identical tags to before, so there is no regression to the working path.

Validation. Verified against real MinerU 3.4.0 output on a generated multi-page-table PDF. The middle.json schema and the merge behavior matched the assumptions, and the merged 61-row table's highlight went from spanning page [0] (before) to [0, 1] (after). Added 11 unit tests in test/unit_test/deepdoc/parser/test_mineru_parser.py covering coordinate scaling, both fallback levels, fragment-index normalization, malformed-schema resilience, multi-page chaining, and the single-page no-regression case. ruff check passes.

### Type of change


JinHai-CN:
@minion1227 would you please resolve the conflicts?

PR #16304, opened 2026-06-24
feat(go-api): add dataset search endpoint

Hz-186:

- added the new dataset search route and handler
- reused the existing shared SearchDatasets service by adapting single-dataset requests into dataset_ids=[dataset_id]
- aligned handler error responses with Python behavior for argument/data errors
- aligned key service error messages such as invalid search_id and mixed embedding models
- added focused handler and service tests for request mapping and error behavior

### Tests:

    `/usr/local/go/bin/go test ./internal/service -run 'TestSearchDatasetRequestToSearchDatasetsRequest|TestDatasetServiceSearchDatasets'`
    `/usr/local/go/bin/go test ./internal/handler -run 'TestDatasetsHandlerSearchDataset'`

### DIFF
--- a/internal/handler/datasets.go
+++ b/internal/handler/datasets.go
@@ -34,8 +34,8 @@ import (
 
 // DatasetsHandler handles the RESTful dataset endpoints.
 type DatasetsHandler struct {
-	datasetsService *service.DatasetService
-	metadataService *service.MetadataService
+	datasetsService      *service.DatasetService
+	metadataService      *service.MetadataService
 	searchDatasetService searchDatasetService
 }
 
--- a/internal/handler/datasets.go
+++ b/internal/handler/datasets.go
@@ -34,9 +34,14 @@ import (
 
 // DatasetsHandler handles the RESTful dataset endpoints.
 type DatasetsHandler struct {
-	datasetsService      *service.DatasetService
-	metadataService      *service.MetadataService
-	searchDatasetService searchDatasetService
+	datasetsService       *service.DatasetService
+	metadataService       *service.MetadataService
+	searchDatasetsService searchDatasetsService
+	searchDatasetService  searchDatasetService
+}
+
+type searchDatasetsService interface {
+	SearchDatasets(req *service.SearchDatasetsRequest, userID string) (*service.SearchDatasetsResponse, error)
 }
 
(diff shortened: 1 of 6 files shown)


yuzhichang:
Thanks for the PR. Most of the service-layer reuse is the right call, but two behavioral gaps contradict the PR description ("aligned … error responses" and "aligned key service error messages"). Could you take a look?

**1. `search_id` is not actually aligned (defect).** Python `dataset_api_service.py:1349-1353` returns `False, "Invalid search_id"` when `SearchService.get_detail(search_id)` is falsy. Go `internal/service/dataset.go:789-792` only logs a warning and silently falls back to the request's own parameters, so a bad `search_id` returns a successful search instead of `102 "Invalid search_id"`. Please return `common.CodeDataError, errors.New("Invalid search_id")` (or similar) on `GetDetail` failure / nil detail.

**2. Argument/data error responses are not actually aligned (defect).** Python's `get_error_argument_result` / `get_error_data_result` return `code: 101/102` with **HTTP 200** (see `api/utils/api_utils.py:248-305`). The new `SearchDataset` handler at `internal/handler/datasets.go` uses raw `http.StatusBadRequest`/`http.StatusInternalServerError` for:
- JSON body parse error (lines ~951)
- missing `question` (lines ~963)
- service error (lines ~980)

Please switch these three sites to the package's `jsonError` helper (HTTP 200 + `common.CodeArgumentError` / `common.CodeDataError`) to match Python. The sibling `SearchDatasets` multi-dataset handler has the same issue — worth fixing together for consistency.

**Minor (non-blocking):**
- Embedding-model message wording differs: Python `"Datasets use different embedding models."` vs Go `"cannot retrieve across datasets with different embedding models"`.
- `question` validation is weaker than Python's `StringConstraints(strip_whitespace=True, min_length=1)`; `"   "` passes in Go.
- Go struct lacks the field-level constraints Python has (`page/size/top_k >= 1`, thresholds clamped to `[0,1]`). Bad inputs currently fall through to the service, which only clamps `top_k`.

Once (1) and (2) are fixed this is ready. The route registration, auth middleware, `dataset_ids` injection, default values, access check, and response shape all look right.

yuzhichang (approved):
LGTM

PR #16309, opened 2026-06-24
fix(api): validate immutable document fields when value is zero

Harsh23Kashyap:
## Summary
- fix `validate_immutable_fields()` to use explicit `is not None` checks for `chunk_count`, `token_count`, and `progress`
- add regression tests for zero-valued immutable fields

## Problem
Submitting `chunk_count=0`, `token_count=0`, or `progress=0.0` bypassed the immutability guard because the function used truthiness checks. That allowed clients to attempt direct updates to immutable document metadata with valid zero values.

## Test plan
- [x] `.v/bin/pytest test/unit_test/api/utils/test_doc_validation.py -q -k immutable` → 11 passed
- [x] `.v/bin/ruff check api/utils/validation_utils.py test/unit_test/api/utils/test_doc_validation.py`


Harsh23Kashyap:
Friendly follow-up: this PR is non-draft and mergeable with a focused immutability-guard fix plus regression tests for zero-valued fields.

Validated locally:

`.v/bin/pytest test/unit_test/api/utils/test_doc_validation.py -q -k immutable` → 11 passed

Could a maintainer please add the `ci` label when possible so the full workflow can run?

PR #16330, opened 2026-06-24
fix(agent/tools): PubMed tool always returns "Unknown Authors"

### LINKED ISSUE #16328: [Bug]: PubMed tool always reports "Unknown Authors"

### Describe the bug

The PubMed search tool (`agent/tools/pubmed.py`) always emits `Authors: Unknown Authors` in its formalized content, regardless of the actual authors in the PubMed record.

### Root cause

In `PubMed._format_pubmed_content`, the nested `safe_find(path)` helper hardcodes its search root to `child` (the whole `<PubmedArticle>`):

```python
def safe_find(path):
    node = child
    ...
```

Inside the author loop it is then called as `safe_find("LastName")` / `safe_find("ForeName")`:

```python
for author in child.findall(".//AuthorList/Author"):
    lastname = safe_find("LastName") or ""   # searches from article root, never matches
    forename = safe_find("ForeName") or ""
```

`LastName`/`ForeName` are not direct children of the article root, so both always return `None` -> every author resolves to an empty string -> `authors_str` is permanently `"Unknown Authors"`. The `author` loop variable is never used.

### Expected behavior

Authors from each `<Author>` element are extracted and joined, e.g. `Furqan Khan, Jane Smith`.

### Steps to reproduce

1. Add a PubMed tool to an agent and run a query that returns articles with authors.
2. Inspect the formalized content - the `Authors:` line is always `Unknown Authors`.

### Additional information

A fix with regression tests is ready; PR to follow.

immuhammadfurqan:

Fixes the PubMed tool always emitting `Authors: Unknown Authors`. The `safe_find` closure in `_format_pubmed_content` was hardcoded to search from the article root, so the per-author `LastName`/`ForeName` lookups never matched.

`safe_find` now accepts an optional `base` node (defaults to `child`, preserving the existing field lookups), and the author loop passes the current `<Author>` element.

Closes #16328

### Type of change

- [x] Add test cases

### Testing

Added `test/testcases/test_web_api/test_canvas_app/test_pubmed_unit.py` covering per-author parsing, intact title/journal/DOI fields, and the no-authors fallback.

Before: `Authors: Unknown Authors`
After:  `Authors: Furqan Khan, Jane Smith`

### DIFF
--- a/agent/tools/pubmed.py
+++ b/agent/tools/pubmed.py
@@ -117,8 +117,8 @@ class PubMed(ToolBase, ABC):
 
     def _format_pubmed_content(self, child):
         """Extract structured reference info from PubMed XML"""
-        def safe_find(path):
-            node = child
+        def safe_find(path, base=None):
+            node = child if base is None else base
             for p in path.split("/"):
                 if node is None:
                     return None
@@ -135,8 +135,8 @@ class PubMed(ToolBase, ABC):
         # Authors
         authors = []
         for author in child.findall(".//AuthorList/Author"):
-            lastname = safe_find("LastName") or ""
-            forename = safe_find("ForeName") or ""
+            lastname = safe_find("LastName", author) or ""
+            forename = safe_find("ForeName", author) or ""
             fullname = f"{forename} {lastname}".strip()
             if fullname:
                 authors.append(fullname)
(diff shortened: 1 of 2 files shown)


yuzhichang (approved):
LGTM

PR #16320, opened 2026-06-24
fix(agent): coerce None Switch inputs before string operators

Harsh23Kashyap:
## Summary
- Coerce `None` canvas values to `""` before string comparison operators in `Switch.process_operator`.
- Prevents `AttributeError` when upstream components yield `None` and the Switch uses contains/start with/end with.

## Test plan
- [x] `.v/bin/python -m ruff check agent/component/switch.py test/unit_test/agent/component/test_switch.py`
- [x] `.v/bin/python -m pytest test/unit_test/agent/component/test_switch.py -q` (3 passed)

Fixes #16315

PR #16331, opened 2026-06-24
fix(agent/component): Switch crashes when upstream variable is None

### LINKED ISSUE #16315: [Bug]: Switch component crashes when upstream variable is None
Switch.process_operator calls input.lower() for contains/not contains/start with/end with operators. When an upstream canvas variable is None, this raises AttributeError instead of routing to the else branch. Location: agent/component/switch.py lines 100-107.

immuhammadfurqan:

`Switch.process_operator` called `input.lower()`/`value.lower()` for the `contains`/`not contains`/`start with`/`end with` operators. When an upstream canvas variable resolved to `None`, this raised `AttributeError` instead of routing to the else branch.

This PR coerces `None` input/value to `""` at the top of `process_operator` so the string operators evaluate to `False` and the component falls through to the else destination.

Closes #16315

### Type of change

- [x] Add test cases

### Testing

Added regression tests in `test/unit_test/agent/component/test_switch.py`:
- a `None` upstream value with a `contains` condition routes to the else branch instead of crashing
- `process_operator` handles `None` input/value for all string operators and `empty`/`not empty`

### DIFF
--- a/agent/component/switch.py
+++ b/agent/component/switch.py
@@ -97,6 +97,12 @@ class Switch(ComponentBase, ABC):
         self.set_output("_next", self._param.end_cpn_ids)
 
     def process_operator(self, input: Any, operator: str, value: Any) -> bool:
+        # Upstream variables may resolve to None; treat them as empty so the
+        # string operators below don't raise AttributeError on None.lower().
+        if input is None:
+            input = ""
+        if value is None:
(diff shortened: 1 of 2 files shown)


yuzhichang:
Duplicated with https://github.com/infiniflow/ragflow/pull/16320.

PR #16332, opened 2026-06-24
fix(agent/tools): DeepL component fails validation and drops errors

### LINKED ISSUE #16329: [Bug]: DeepL component never passes validation (check() raises AttributeError)

### Describe the bug

`DeepLParam.check()` in `agent/tools/deepl.py` validates `self.top_n`, but DeepL has no such parameter - it is not defined on `DeepLParam.__init__` nor on `ComponentParamBase`. As a result `check()` always raises `AttributeError: 'DeepLParam' object has no attribute 'top_n'`, so a DeepL component can never pass validation.

```python
def check(self):
    self.check_positive_integer(self.top_n, "Top N")   # top_n is never defined
    self.check_valid_value(self.source_lang, ...)
    self.check_valid_value(self.target_lang, ...)
```

### Secondary issue

In `DeepL._run`, the `except` branch computes `be_output("**Error**..." )` but never returns it, so on failure the error message is silently dropped:

```python
except Exception as e:
    if self.check_if_canceled("DeepL processing"):
        return
    DeepL.be_output("**Error**:" + str(e))   # missing return
```

### Expected behavior

- `DeepLParam().check()` validates source/target languages without raising on a non-existent `top_n`.
- On translation failure, the error message is returned to the user.

### Additional information

A fix with tests is ready; PR to follow.

immuhammadfurqan:

`DeepLParam.check()` validated `self.top_n`, but DeepL has no such parameter (it is not defined on the param class or its base), so `check()` always raised `AttributeError` and a DeepL component could never pass validation. Removed the bogus `top_n` check.

Also fixed the `_run` except branch, which computed `be_output("**Error**...")` but never returned it, silently dropping the error message.

Closes #16329

### Type of change

- [x] Add test cases

### Testing

Added `test/unit_test/agent/component/test_deepl.py` covering `DeepLParam.check()` with valid defaults and rejection of invalid source/target languages.

### DIFF
--- a/agent/tools/deepl.py
+++ b/agent/tools/deepl.py
@@ -31,7 +31,6 @@ class DeepLParam(ComponentParamBase):
         self.target_lang = 'EN-GB'
 
     def check(self):
-        self.check_positive_integer(self.top_n, "Top N")
         self.check_valid_value(self.source_lang, "Source language",
                                ['AR', 'BG', 'CS', 'DA', 'DE', 'EL', 'EN', 'ES', 'ET', 'FI', 'FR', 'HU', 'ID', 'IT',
                                 'JA', 'KO', 'LT', 'LV', 'NB', 'NL', 'PL', 'PT', 'RO', 'RU', 'SK', 'SL', 'SV', 'TR',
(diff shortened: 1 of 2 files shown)


yuzhichang (approved):
LGTM

PR #16336, opened 2026-06-25
feat: unify provider id or name routing

buua436:

Unifies provider and instance-related provider API routes to accept either an ID or a name.

### Type of change

- [x] Feature (non-breaking change which adds functionality)


Lynn-Inf (approved):
LGTM

PR #16359, opened 2026-06-25
feat(llm): add TwelveLabs provider (Marengo embeddings + Pegasus video understanding)

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

RAGFlow has no native way to embed or understand **video** content. This PR adds an opt-in **TwelveLabs** model provider that plugs into RAGFlow's existing LLM seams:

- **`TwelveLabsEmbed` (Marengo, `marengo3.0`)** — a text-embedding model (`rag/llm/embedding_model.py`) producing 512-dim **multimodal** vectors. Because Marengo embeds text, image, video and audio into one shared space, these text embeddings can be matched against video/image embeddings, which is useful for multimodal RAG.
- **`TwelveLabsCV` (Pegasus, `pegasus1.5` / `pegasus1.2`)** — a video-understanding model wired into RAGFlow's vision (`image2text`) seam (`rag/llm/cv_model.py`). Pegasus reads a whole video natively (motion, speech, on-screen text) and answers a prompt about it, rather than captioning a single frame. The input may be an `http(s)` URL, a `tl-asset:<asset_id>` reference, or raw video bytes (uploaded as a TwelveLabs asset).

Both classes auto-register through the existing `_FACTORY_NAME` discovery in `rag/llm/__init__.py`, and the provider is listed in `conf/llm_factories.json` and `conf/models/twelvelabs.json` (mirroring the existing Perplexity provider's dual registration). The dependency is added to `pyproject.toml` and the provider is documented in `docs/guides/models/supported_models.mdx`.

### Type of change

- [x] Documentation Update

### Non-breaking / opt-in

No defaults change and no existing behavior is modified. The provider is only instantiated when a user explicitly configures TwelveLabs in their model settings.

### How it was tested

- **Unit tests** (`test/unit_test/rag/llm/test_twelvelabs.py`, fully mocked, mirroring `test_perplexity_embed.py`): default model resolution, `encode`/`encode_queries` shapes and token counts, empty-response error handling, Pegasus URL/asset/bytes context resolution, and bad-input rejection.
- **Live smoke test** (gated on `TWELVELABS_API_KEY`, skipped without it): verified against the official `twelvelabs==1.2.8` SDK that `Marengo` returns a real **512-dim** text-embedding vector through the exact code path. Pegasus request wiring is unit-verified; a full server-side video analysis run can be slow on long clips.
- `ruff check` passes on all changed files (repo config, line-length 200).

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.


JinHai-CN:
@mohit-twelvelabs Would you please fix the format issue?

```
Would reformat: rag/llm/cv_model.py
1 file would be reformatted, 2 files already formatted

exit status 1
```

JinHai-CN:
If you continue to fix the PR, please re-open the PR.

PR #16363, opened 2026-06-25
fix(agent): restore be_output and test DeepL error return

Harsh23Kashyap:
## Summary

#16332 fixed the missing `return` in DeepL's except branch, but `ComponentBase.be_output` was removed during the agent refactor (#9113) while several components still call it. DeepL (and other tools) would raise `AttributeError` before any error message could be returned.

- Restore `ComponentBase.be_output` as `pd.DataFrame([{"content": v}])` (same as pre-refactor behavior)
- Add regression test that `_run` returns the `**Error**:` message when translation fails

Related to #16329

## Test plan

- [x] `test_run_returns_error_on_translation_failure`
- [x] Existing `test_deepl.py` check() tests still pass


Harsh23Kashyap:
Could a maintainer add the `ci` label when convenient? Thanks.

Harsh23Kashyap:
Fixed the preflight Ruff failure in 2dd8d938a by removing the unused `MagicMock` import from `test/unit_test/agent/component/test_deepl.py`.

Verification:
- `python3 -m py_compile test/unit_test/agent/component/test_deepl.py`
- `uvx ruff check test/unit_test/agent/component/test_deepl.py`

yuzhichang (commented):
I re-reviewed the current head locally. The `be_output()` restoration and the DeepL failure-path regression test both look correct, and I do not see any remaining blocking issue in the current diff. The only unresolved bot comment is a very low-value cleanup in the test setup, so from my side this is ready to merge.

PR #16365, opened 2026-06-25
fix: fallback to factory config for max_tokens when tenant_model.extra is empty

### LINKED ISSUE #16362: tenant_model.extra missing max_tokens causes knowledge truncation to 8192 in chat completions
## Description

After upgrading to v0.26.1, chat completions return significantly worse results — the LLM appears to make incorrect judgments because it cannot see the full retrieved knowledge.

## Root Cause

In `api/db/joint_services/tenant_model_service.py`, `get_model_config_from_provider_instance` reads `max_tokens` from `tenant_model.extra`:

```python
# line 228 — when model_obj exists
"max_tokens": model_extra.get("max_tokens", 8192),
```

But `tenant_model.extra` is initialized as empty `{}` for all models. The factory config's `max_tokens` (e.g., 128000 for qwen3-235b) is never synced into the database.

Downstream in `dialog_service.py`:

```python
# line 563
max_tokens = llm_model_config.get("max_tokens", 8192)  # → 8192 instead of 128000
knowledges = kb_prompt(kbinfos, max_tokens)            # knowledge truncated at 8K
used_token_count, msg = message_fit_in(msg, int(max_tokens * 0.95))  # → 7782
```

The LLM receives only ~8K tokens of knowledge context instead of 128K, missing critical information.

## Why extra is empty

In `api/apps/services/provider_api_service.py:789-791`:

```python
if status == ActiveStatusEnum.ACTIVE.value:
    # Default is active, no need to add a record
    return True, None
```

During `--init-model-provider-tables`, models with active status skip record creation entirely. When records are later created through the model management UI/API, `max_tokens` from the factory config is not copied into `extra`.

## Suggested Fix

Line 228 should fall back to `llm_info.get("max_tokens")` from the factory config when `model_extra` is empty:

```python
# current
"max_tokens": model_extra.get("max_tokens", 8192),

# suggested  
"max_tokens": model_extra.get("max_tokens") or llm_info.get("max_tokens") or 8192,
```

## Workaround

```sql
UPDATE tenant_model SET extra = '{"max_tokens": 128000}' 
WHERE model_type = 'chat' AND (extra = '{}' OR extra IS NULL);
```

## Impact

Every v0.26.1 user who creates chat assistants through the new model management system is affected — knowledge context is silently truncated to 8K tokens regardless of the model's actual context window.

## Version

v0.26.1

huzhw:

Fixes #16362

When tenant_model.extra is empty, get_model_config_from_provider_instance defaults max_tokens to 8192 instead of the model's actual context window.

This causes kb_prompt(kbinfos, max_tokens) to truncate knowledge to 8192 tokens, silently degrading chat quality.

## Fix

Fall back to llm_info from FACTORY_LLM_INFOS when model_extra is missing max_tokens.

## Test

qwen3-235b-a22b with empty extra:
- Before: max_tokens = 8192
- After: max_tokens = 128000 (from factory config)

### DIFF
--- a/api/db/joint_services/tenant_model_service.py
+++ b/api/db/joint_services/tenant_model_service.py
@@ -218,14 +218,23 @@ def get_model_config_from_provider_instance(tenant_id, model_type: str|enum.Enum
             raise LookupError(f"Model {model_name} cannot be used as {model_type_val} model.")
 
         model_extra = json.loads(model_obj.extra) if model_obj.extra else {}
+
+        # Fallback to factory config for fields missing from model_extra
+        fac_list = [f for f in settings.FACTORY_LLM_INFOS if f["name"] == provider_name]
+        llm_info = None
+        if fac_list:
+            llm_list = [llm for llm in fac_list[0]["llm"] if llm["llm_name"] == model_obj.model_name]
+            if llm_list:
+                llm_info = llm_list[0]
+
         model_config = {
             "llm_factory": provider_obj.provider_name,
             "api_key": api_key,
             "llm_name": model_obj.model_name,
             "api_base": extra_fields.get("base_url", ""),
             "model_type": model_obj.model_type,
-            "is_tools": model_extra.get("is_tools", is_tool),
-            "max_tokens": model_extra.get("max_tokens", 8192),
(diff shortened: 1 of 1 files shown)


Lynn-Inf (approved):
LGTM

PR #16401, opened 2026-06-27
Fix OpenAI response created timestamp

### LINKED ISSUE #16400: [Bug]: OpenAI-compatible responses return created as null by default
### Self Checks

Current main branch

### Describe your problem

`get_data_openai()` returns `created: null` for non-streaming OpenAI-compatible responses when the caller does not pass a `created` value.

The current implementation in `api/utils/api_utils.py` is:

```python
"created": int(time.time()) if created else None,
```

That means the default `created=None` produces `null`, even though OpenAI-compatible chat completion responses are expected to include an integer Unix timestamp. It also ignores explicit caller-provided timestamps by replacing any truthy `created` value with the current time.

One user-facing path is `api/db/services/canvas_service.py`, where non-streaming OpenAI-compatible agent responses call `get_data_openai()` without a `created` value.

### Expected behavior

For non-streaming responses:

- If `created` is not provided, `get_data_openai()` should set it to the current Unix timestamp.
- If `created` is provided, it should preserve that value.
- Streaming chunk response shape should remain unchanged.

### Proposed fix

Use the provided value when it is not `None`; otherwise default to `int(time.time())`.

Harsh23Kashyap:
## What this fixes

Closes #16400.

`get_data_openai()` currently returns `created: null` when callers do not pass a timestamp, and it replaces explicit timestamp values with the current time. This makes non-streaming OpenAI-compatible responses inconsistent with the expected integer `created` timestamp field.

## Change

- Preserve explicit `created` values when provided.
- Default non-streaming responses to `int(time.time())` when `created` is not provided.
- Add focused unit coverage for default timestamps, explicit timestamps, and unchanged streaming chunk shape.

## Verification

- `./.venv/bin/python -m pytest test/unit_test/api/utils/test_api_utils.py -q`
- `python3 -m py_compile api/utils/api_utils.py test/unit_test/api/utils/test_api_utils.py`
- `uvx ruff check api/utils/api_utils.py test/unit_test/api/utils/test_api_utils.py`


Harsh23Kashyap:
Synced with latest main. Could a maintainer add the ci label so workflows can run? Okay, please review it.

Harsh23Kashyap:
Follow-up: branch is current with `main` (`behind_by: 0`). CodeRabbit passed; no unresolved review threads. Fork workflows are still skipped until a maintainer adds the `ci` label. Ready for review when you have a chance.

Harsh23Kashyap:
Can someone take a look when you have a chance?

yuzhichang (approved):
LGTM

PR #16415, opened 2026-06-27
fix(agent/tools): port Crawler to ToolBase so it can load and run

### LINKED ISSUE #16414: [Bug]: Crawler agent tool is non-functional — CrawlerParam crashes on init (missing meta) and uses obsolete _run/be_output

### Describe the bug

The **Crawler** agent tool (`agent/tools/crawler.py`) is non-functional. It was never ported to the modern `ToolBase`/`_invoke` interface that all other tools in `agent/tools/` now use, so it fails in three independent ways.

### Root cause

**1. `CrawlerParam` raises `AttributeError` on construction (canvas fails to load).**

`CrawlerParam` extends `ToolParamBase` but never defines `self.meta`:

```python
class CrawlerParam(ToolParamBase):
    def __init__(self):
        super().__init__()        # <-- crashes here
        self.proxy = None
        self.extract_type = "markdown"
```

`ToolParamBase.__init__` immediately reads `self.meta`:

```python
class ToolParamBase(ComponentParamBase):
    def __init__(self):
        super().__init__()
        self._init_inputs()       # -> self.meta["parameters"]
        self._init_attr_by_meta() # -> self.meta["parameters"]
```

Every working tool (e.g. `ArXivParam`, `TavilyExtractParam`) sets `self.meta: ToolMeta = {...}` *before* calling `super().__init__()`. `CrawlerParam` does not, so constructing it raises `AttributeError: 'CrawlerParam' object has no attribute 'meta'`. Because `agent/canvas.py` instantiates `component_class(component_name + "Param")()` when loading a canvas, **any agent that contains a Crawler node fails to load.**

**2. `_invoke` is never implemented.**

`Crawler` extends the modern `ToolBase`, whose `invoke()` dispatches to `self._invoke(**kwargs)`. But `Crawler` only defines the legacy `_run(self, history, **kwargs)`. So `_invoke` resolves to `ComponentBase._invoke`, which `raise NotImplementedError()`.

**3. `be_output` no longer exists.**

`_run` calls `Crawler.be_output(...)`, which was removed from the base classes during the agent redesign. Even if `_run` were reachable, it would raise `AttributeError`.

`git log` confirms `crawler.py` has used `_run`/`be_output` and lacked `meta` ever since the agent module redesign (`d9fe279dd`), so the tool has been broken the whole time.

### Expected behavior

- `CrawlerParam()` constructs without error.
- The Crawler tool exposes a proper tool descriptor (`get_meta()`) with a `url` parameter so an Agent's LLM can call it.
- Invoking it fetches the page via `crawl4ai` (keeping the existing SSRF guard) and writes the extracted content to the component output instead of the removed `be_output`.

### Steps to reproduce

1. Build an agent that uses the **Web Crawler** tool.
2. Load/run the agent — it fails because `CrawlerParam` cannot be instantiated.

Minimal:
```python
from agent.tools.crawler import CrawlerParam
CrawlerParam()  # AttributeError: 'CrawlerParam' object has no attribute 'meta'
```

### Additional information

Same class of defect as #16329 (DeepL). A fix porting Crawler to the `ToolBase`/`_invoke` interface (with `meta`, `set_output`, preserved SSRF guard) and a regression test is ready; PR to follow.

immuhammadfurqan:

Closes #16414.

The **Crawler** agent tool (`agent/tools/crawler.py`) was never ported to the modern `ToolBase`/`_invoke` interface during the agent module redesign, so it was broken in three independent ways:

1. **Crashed on construction.** `CrawlerParam` extends `ToolParamBase`, whose `__init__` reads `self.meta["parameters"]`, but `CrawlerParam` defined no `meta`. Constructing it raised `AttributeError: 'CrawlerParam' object has no attribute 'meta'`. Because `agent/canvas.py` instantiates `component_class(component_name + "Param")()` while loading a canvas, **any agent containing a Crawler node failed to load.**
2. **`_invoke` missing.** It extends `ToolBase` (whose `invoke()` dispatches to `self._invoke`) but only implemented the legacy `_run`, so `_invoke` resolved to `ComponentBase._invoke` → `NotImplementedError`.
3. **`be_output` removed.** `_run` called `Crawler.be_output(...)`, which no longer exists on the base classes.

### Type of change

### Changes

- Add a `ToolMeta` to `CrawlerParam` (defined before `super().__init__()`, matching every other ported tool such as `ArXivParam`/`TavilyExtractParam`) advertising a required `query` parameter — the URL to crawl, default `{sys.query}`, consistent with the `{sys.query}` convention shared by the other tools.
- Replace the legacy `_run`/`be_output` with `_invoke`/`set_output`, writing the extracted page content to `formalized_content` (errors surfaced via `_ERROR`), consistent with the other tools.
- Preserve the existing SSRF guard (`assert_url_is_safe` + `pin_dns_global`).
- Add regression tests (`test/unit_test/agent/component/test_crawler.py`) covering param construction, validation, and the tool descriptor.

Same class of defect as #16329 (DeepL). Backend-only; no frontend changes.


yuzhichang (commented):
Thanks for porting Crawler onto `ToolBase`. I re-reviewed the current head and the core load-time/runtime breakage is fixed: the tool now has `meta`, exposes a callable `_invoke`, and no longer depends on removed `be_output` behavior.

I did not find a blocking correctness issue in the implementation itself. The main thing still missing for merge confidence is one direct regression test for the `_invoke` execution path, not just `CrawlerParam`. A mocked crawl test that asserts `Crawler._invoke(query=...)` returns content and writes `formalized_content` would lock down the exact runtime path this PR is restoring.

yuzhichang (commented):
I re-reviewed the branch locally. The core runtime regression is fixed: the tool now has `ToolMeta`, a working `_invoke` entrypoint, and no longer depends on removed `be_output` behavior. I did not find a merge-blocking implementation bug in the current code.

The remaining gap is test coverage for the restored execution path. Please add one focused regression test that mocks the crawl result, calls `Crawler._invoke(query=...)`, and asserts both the returned content and the `formalized_content` output. That will lock down the exact runtime path this PR is restoring.

immuhammadfurqan:
Thanks @yuzhichang — added direct `_invoke` regression tests in `ecf2cc0c0`.

With a mocked `get_web` and a stubbed SSRF guard (and bypassing the canvas-bound `__init__` via `Crawler.__new__`, mirroring `test_pubmed_unit.py`):

- `test_invoke_returns_content_and_sets_formalized_content` — calls `Crawler._invoke(query="http://example.com")` and asserts it both **returns** the page content and writes it to **`formalized_content`**.
- `test_invoke_empty_query_returns_empty` — an empty query short-circuits, returns `""`, and never calls `get_web`.
- `test_invoke_rejects_unsafe_url` — an unsafe URL returns `"URL not valid"` / sets `_ERROR` before any crawl is attempted.

Full file (7 tests) passes locally.

PR #16419, opened 2026-06-27
fix(agent/tools): GoogleScholar empty json output and ignored top_n

### LINKED ISSUE #16418: [Bug]: GoogleScholar tool always returns empty json output and ignores top_n (exhausted/unbounded generator)

### Describe the bug

The **GoogleScholar** agent tool (`agent/tools/googlescholar.py`) has two bugs caused by treating `scholarly.search_pubs(...)` — a lazy generator — as if it were a re-iterable, bounded list.

### Root cause

```python
scholar_client = scholarly.search_pubs(kwargs["query"], ...)   # lazy generator

self._retrieve_chunks(scholar_client, ...)                     # (1) iterates -> exhausts it
self.set_output("json", list(scholar_client))                  # (2) generator already empty -> []
```

**1. The `json` output is always empty.** `_retrieve_chunks` iterates `scholar_client` (`for r in res_list`), which exhausts the generator. `list(scholar_client)` then runs on the exhausted generator and always returns `[]`, so the structured `json` output is permanently empty.

**2. `top_n` is never applied.** Unlike `ArXiv` (which bounds results with `max_results=self._param.top_n`), GoogleScholar passes the unbounded generator straight to `_retrieve_chunks`, which has no internal limit. `scholarly.search_pubs` paginates lazily through *all* matching results, so the tool tries to fetch far more than `top_n` (effectively until it errors, gets rate-limited/blocked by Google Scholar, or hits `COMPONENT_EXEC_TIMEOUT`), ignoring the user's configured `top_n` entirely.

### Expected behavior

- `top_n` limits the number of results fetched.
- The `json` output contains the same results that were summarized into `formalized_content` (non-empty when there are results).

### Steps to reproduce

1. Add a GoogleScholar tool to an agent, set Top N (e.g. 12), run a query.
2. The `json` output is `[]`, and the tool fetches/streams well beyond Top N results (slow / rate-limited) instead of stopping at Top N.

### Proposed fix

Materialize the generator once, bounded by `top_n` (e.g. `itertools.islice`), then reuse that list for both `_retrieve_chunks` and the `json` output.

### Additional information

A fix and a regression test (stubbing `scholarly.search_pubs`) are ready; PR to follow.

immuhammadfurqan:

Closes #16418.

`scholarly.search_pubs(...)` returns a **lazy generator**, but `agent/tools/googlescholar.py` treated it as a re-iterable, bounded list:

```python
scholar_client = scholarly.search_pubs(kwargs["query"], ...)   # lazy generator
self._retrieve_chunks(scholar_client, ...)                     # (1) iterates -> exhausts it
self.set_output("json", list(scholar_client))                  # (2) already empty -> []
```

1. **`json` output was always empty.** `_retrieve_chunks` iterates `scholar_client`, exhausting the generator; `list(scholar_client)` then returns `[]`.
2. **`top_n` was never applied.** Unlike `ArXiv` (`max_results=self._param.top_n`), the unbounded generator was passed straight to `_retrieve_chunks`, which has no internal limit — so the tool kept paginating well past Top N (until an error, rate-limit/block, or `COMPONENT_EXEC_TIMEOUT`).

### Type of change

### Changes

- Materialize at most `top_n` results once with `itertools.islice`, and reuse that list for both `_retrieve_chunks` and the `json` output.
- Add regression tests (`test/unit_test/agent/component/test_googlescholar.py`, stubbing `scholarly.search_pubs`) covering the `top_n` bound, the non-empty `json` output, and the empty-query short-circuit.

Verified: against `main` the new tests fail with `assert 30 == 5` (top_n ignored) and `assert 0 == 5` (empty json); with this fix all pass. Backend-only.


yuzhichang (commented):
Thanks for the fix. I re-reviewed the current head and the core generator/top_n bug is fixed, but I still see two follow-ups before merge:

1. `GoogleScholar._invoke()` returns early on empty `query` after only clearing `formalized_content`, so a reused tool instance can keep stale `json` output from a previous successful call. Please reset `json` on the empty-query path as well.
2. The new regression tests still don't prove the short-circuit strongly enough. In `test_empty_query_short_circuits`, please mock `scholarly.search_pubs` and assert it is not called. In `test_respects_top_n`, it would also be better to assert only `top_n` generator items are actually consumed, not just that `_retrieve_chunks` saw a sliced list.

Once those are tightened, this looks good to merge.

immuhammadfurqan:
Thanks for the review @yuzhichang — both addressed in `ba3fead8e`:

1. **Stale `json` on empty query.** The empty-query early return now also resets `json` to `[]` (not just `formalized_content`), so a reused instance can't surface results from a previous successful call.

2. **Stronger tests.**
   - `test_empty_query_short_circuits` now mocks `scholarly.search_pubs` and asserts it is **not called**, and that `json` is cleared (it pre-seeds a stale `["stale"]` value to prove the reset).
   - `test_respects_top_n` now also asserts that only `top_n` items are **actually consumed** from the generator (via a counter incremented per `yield`), not just that `_retrieve_chunks` received a sliced list.

Verified against `main`, all three new assertions fail (`assert 30 == 5`, `assert 0 == 5`, `assert ['stale'] == []`); with the fix they pass.

PR #16431, opened 2026-06-29
fix(auth): handle missing primary GitHub OAuth email

VectorPeak:

GitHub's `/user/emails` response can omit a primary email for accounts whose email is private or otherwise unavailable. The previous GitHub OAuth flow assumed a primary email was always present, which could fail during login instead of falling back to the public profile email or surfacing a clear missing-email error.

This PR restores the fork-recovered change from #16327 and keeps the GitHub OAuth behavior explicit: prefer the primary email, fall back to the public profile email, and raise `GitHub account email is unavailable.` when neither source has an address.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


VectorPeak:
Addressed the latest review/pre-merge feedback:

- Completed the PR description with the project template sections and marked this as a bug fix.
- Added docstrings in `api/apps/auth/github.py` and the related GitHub OAuth unit-test helpers/tests to improve docstring coverage for the changed files.
- Confirmed the original CodeRabbit inline findings from #16327 are already addressed in the current branch: GitHub email fallback/unavailable-email logging is present, and the explicit `GitHub account email is unavailable.` `ValueError` is preserved instead of being re-wrapped.

Validation:

- `uvx ruff check api/apps/auth/github.py test/testcases/test_web_api/test_auth_app/test_oidc_client_unit.py` passed.
- `git diff --check` passed.
- Pytest could not complete locally because repository-level test setup exits without `ZHIPU_AI_API_KEY`.

PR #16451, opened 2026-06-29
fix: allow resume chunk method in document update validation

WOLIKIMCHENG:
### Summary
Allow `resume` as a valid `chunk_method` in `UpdateDocumentReq`.

Latest `main` already accepts `audio`; this change preserves that behavior and adds `resume`. Focused unit tests cover acceptance of both values and the exact invalid-method error type, message, and context.

The change is intentionally Python-only because the Go runtime does not support the `resume` parser.

### Tests
- Focused unit tests: 30 passed
- Syntax compilation: passed
- `git diff --cached --check MERGE_HEAD`: passed


wangq8:
`resume` is not supported in OSS

PR #16468, opened 2026-06-29
Fix: ValueError: too many values to unpack in list_tenant_added_models for model IDs containing '@' (#16467)

### LINKED ISSUE #16467: [Bug]: ValueError: too many values to unpack (expected 3) in list_tenant_added_models when LM-Studio model_name contains '@'
## Self Checks

---

`6370fce3f07451e46df3af48e8165786dcad259f`

## RAGFlow image version

`v0.26.2` (custom build from current `main`)

## Other environment information

- **Hardware**: CPU-only container, 16 GB RAM
- **OS**: Docker on Windows 11 / WSL2 (Linux container, Python 3.13.11)
- **Affected provider**: `LM-Studio` (OpenAI-compatible local server at `http://host.docker.internal:1234`)
- **Affected model type**: `embedding`
- **Other apps using the same LM Studio**: working fine (chat completions etc.) — the bug is not in the model itself, only in RAGFlow's key parsing.

## Actual behavior

When the tenant has any manually-added embedding model whose `model_name` contains an `@` character, the `GET /api/v1/models` endpoint (`get_added_models` → `list_tenant_added_models`) crashes with an unhandled exception. Subsequent requests for the same endpoint return a 500.

The crash log:

```
2026-06-29 22:50:04,817 ERROR    35 too many values to unpack (expected 3)
  File "/ragflow/api/apps/restful_apis/models_api.py", line 74, in get_added_models
    success, result = models_api_service.list_tenant_added_models(tenant_id, model_type_filter)
  File "/ragflow/api/apps/services/models_api_service.py", line 388, in list_tenant_added_models
    provider_id, instance_id, model_name = model_record_key.split("@")
ValueError: too many values to unpack (expected 3)
```

A related warning fires for the same provider on startup:

```
WARNING: Instance 'q8_0' not found for provider 'LM-Studio'
WARNING: Instance 'default' not found for provider 'LM-Studio'
```

## Expected behavior

`list_tenant_added_models` should return all manually-added models for the tenant regardless of whether the `model_name` contains an `@` character. The endpoint should not 500.

## Steps to reproduce

1. Stand up LM Studio locally with at least one embedding model that exposes an `@<quantization>` suffix, e.g. `text-embedding-nomic-embed-text-v1.5@q8_0` or `text-embedding-nomic-embed-text-v1.5@q4_k_m`. Confirm with `curl http://127.0.0.1:1234/v1/models` — those IDs are returned as-is by LM Studio's OpenAI-compatible API.
2. In RAGFlow: **Model Providers → LM-Studio → Add an embedding model** and select `text-embedding-nomic-embed-text-v1.5@q8_0` (or any `@`-suffixed embedding model).
3. Call `GET /api/v1/models?model_type=embedding` (the `get_added_models` endpoint under the hood).
4. Observe the 500 and the `ValueError: too many values to unpack (expected 3)` traceback in the `ragflow-server` logs.

## Root cause

In [`api/apps/services/models_api_service.py`](https://github.com/infiniflow/ragflow/blob/main/api/apps/services/models_api_service.py), the composite key for a model row is built with `@` as a separator:

```python
# line ~358
model_record_key = f"{model.provider_id}@{model.instance_id}@{model.model_name}"
```

For LM Studio embedding models, `model.model_name` itself contains `@` (LM Studio appends `@<quantization>` to its model IDs), so the resulting key has more than 3 `@`-separated parts. Later, at line 388, the code assumes exactly 3 parts:

```python
provider_id, instance_id, model_name = model_record_key.split("@")
```

This raises `ValueError` whenever any tenant has a manual embedding row whose `model_name` contains `@`.

The same family of bugs explains the `Instance 'q8_0' not found for provider 'LM-Studio'` warning — other call sites assume the right-hand side of an `@` is an `instance_name`, but LM Studio uses `@` as part of model names.

## Suggested fix

1. **Hotfix (smallest, safe)**: change `model_record_key.split("@")` to `model_record_key.rsplit("@", 2)` so the split is from the right and always yields 3 parts. Re-join any extra `@` characters back into `model_name`. This makes the parser resilient to model names that legitimately contain `@`.
2. **Cleanup (recommended follow-up)**: the `get_added_models` endpoint should not 500 — wrap the per-record parsing in `try/except` and skip/log malformed keys instead of crashing the whole response.

A PR is being opened for this.

## Additional information

- A related fix in the same area: [`Fix: failed to get embedding model by embd_id: model config not found BAAI/bge-m3@...@SILICONFLOW (#16445)`](https://github.com/infiniflow/ragflow/commit/9b726a519) — confirms `@` in model identifiers is a recurring issue across the codebase.
- The two LM Studio embeddings currently visible in our installation (returned by `GET http://127.0.0.1:1234/v1/models`):
  - `text-embedding-nomic-embed-text-v1.5@q8_0`
  - `text-embedding-nomic-embed-text-v1.5@q4_k_m`
- Chat models exposed by LM Studio in the same instance do not contain `@` and are unaffected, which is why "LM Studio works fine in other apps" — only the embedding model IDs trigger the bug.

skbs-eng:
## Summary

Fixes #16467 — `ValueError: too many values to unpack (expected 3)` in `list_tenant_added_models` for model IDs containing `@`, **and three sibling code paths that exhibit the same `@`-mis-parsing bug**. Adds a downstream `NotFoundError` tolerance so the dataset-configuration "check embedding" endpoint degrades to a user-friendly message instead of HTTP 500 when the tenant chunk index is missing.

### Root cause (recap)

The composite model key is right-anchored:

```
model_name@instance_name@provider_name      # explicit instance
model_name@provider_name                    # default instance
```

Some model names legitimately contain `@` characters (LM Studio embedding model IDs like `text-embedding-nomic-embed-text-v1.5@q8_0`, llama.cpp `@<quantization>` suffixes, HuggingFace dataset IDs). The full composite key then has **4+** `@`-separated fields:

```
text-embedding-nomic-embed-text-v1.5@q8_0@lmstudio@LM-Studio
```

A naive `split("@")` + 3-element unpack mis-assigns the fields.

### Sites fixed

| # | File | Function | Fix |
|---|------|----------|-----|
| 1 | `api/apps/services/models_api_service.py` | `_get_model_info` | `key.rsplit("@", 2)`; bare-name case sets empty provider |
| 2 | `api/db/joint_services/tenant_model_service.py` | `split_model_name` | `rsplit("@", 2)` from the right |
| 3 | `internal/service/model_service.go` | `parseModelName` | Rejoins leading fields when >3 `@`-segments |
| 4 | `internal/service/model_service.go` | `GetModelConfigFromProviderInstance` | TEI exact-match + Builtin suffix-based detection to handle bare model IDs with embedded `@` |
| 5 | `api/apps/services/models_api_service.py` | `list_tenant_added_models` | Internal map keys use `|` separator instead of `@` to avoid conflict with model names containing `@` |
| 6 | `api/apps/services/dataset_api_service.py` | `sample_random_chunks_with_vectors` | Graceful empty-set return on missing chunk index |

### Review feedback addressed (commit 26a2cec59)

- **Removed deprecated `TenantLLMService` dependency** from `models_api_service.py` — the `_resolve_bare_model_provider()` function and its `TenantLLMService` import were deleted per maintainer request.
- **Replaced `@` with `|` as internal map key separator** in `list_tenant_added_models` to avoid conflicts with model names containing `@`.
- **Fixed Go TEI/Builtin fast paths** — the TEI short-circuit now does exact string match first to handle bare model IDs with embedded `@`. The Builtin provider detection uses `strings.HasSuffix` instead of `splitRightAnchoredModelName` to avoid misidentifying embedded `@` as a provider separator.
- **Updated tests** — removed bare-model tests that depended on deprecated `TenantLLMService`, added test for two-part model IDs with embedded `@`.

### Testing

- **Python**: 5/5 unit tests pass (`test_models_api_service_list_tenant_added_models.py`)
- **Go**: `gofmt` validates syntax; Go tests require Linux (platform-specific `binding` package)
- **Regression tests**: `test_tenant_model_service_split_model_name.py` and `model_service_test.go` cover the parsing fixes


skbs-eng:
Addressed the CodeRabbit docstring-coverage warning (75.00% → 100%) in commit `bf8bf01e2`.

The two helpers missing docstrings were:
- `_stub` (registers a stubbed module in `sys.modules`)
- `_make_model_record` (builds a `SimpleNamespace` mirroring a `TenantModel` ORM row)

Both now have full Google-style docstrings (Args / Returns) explaining the monkeypatching contract and the `provider-1`/`instance-1` fixture assumption shared with `_load_module`.

No production code changes; all 4 tests still pass:

```
4 passed in 0.04s
```

Lynn-Inf (approved):
LGTM

Lynn-Inf (changes requested):
Please remove `tenant_llm` service and DB calls in `models_api_service.py` — they are deprecated and will be removed.

Also, a suggestion on `@` handling: consider counting occurrences before splitting. The `rsplit` approach in `tenant_model_service.py` is clean and safe. For `models_api_service`, since `@` is only used as a separator when generating map keys, consider replacing it with a character that's less likely to appear in `model_name`(e.g., `|` or `^`).

skbs-eng:
@coderabbitai resume

Pushed a follow-up at `a9977572f` addressing the **Major / Functional Correctness** concern on `splitRightAnchoredModelName`: the 2-segment case no longer treats the trailing token as a provider unless it actually is one (`"Builtin"`); otherwise the whole string stays as the bare `modelName` and the caller falls through to the standard lookup. A regression test was added to `TestSplitRightAnchoredModelName` covering the `text-embedding-nomic-embed-text-v1.5@q8_0` LM Studio embedding case.

Could you take another look?

skbs-eng:
## Follow-up: `splitRightAnchoredModelName` 2-segment edge case

Thanks @wangq8 for the `ci` label and @Lynn-Inf for the earlier reviews. Pushed `a9977572f` addressing the CodeRabbit *Major / Functional Correctness* comment on the TEI/Builtin helper.

**Behaviour before:** `splitRightAnchoredModelName("text-embedding-nomic-embed-text-v1.5@q8_0")` returned `("text-embedding-nomic-embed-text-v1.5", "default", "q8_0")`, mis-classifying the quantization tag as a provider and missing the TEI fast path's `modelName == teiModel` exact-match when `TEI_MODEL` is the full embedded string.

**Behaviour after:** the helper only treats the last token as a provider when it equals `"Builtin"` (the only recognised suffix in its sole call-sites, the TEI / Builtin short-circuits). For any other 2-segment key the whole string is returned as the bare `modelName` with empty `instanceName` / `providerName`, and the caller falls through to the standard lookup — same as the 1-segment case. The pre-existing TEI exact-match fast path at the top of the TEI short-circuit (`if modelName == teiModel`) covers the bare-default case where the embedded `@` happens to match `TEI_MODEL` verbatim.

**Tests:** added `"two parts bare default with embedded '@' stays bare"` to `TestSplitRightAnchoredModelName`. The existing 4 cases are unchanged and continue to pass.

Diff scope: 2 files, +29 / −1. Re-requesting review from @coderabbitai (auto-paused after the earlier commit influx) and would appreciate a re-look from @Lynn-Inf on the new commit when convenient.

skbs-eng:
Rebased onto current main (8b065d3dd) — the only conflict was in api/apps/services/dataset_api_service.py at sample_random_chunks_with_vectors, where main had reformatted the docStoreConn.search(...) call to multi-line args. Resolution kept main's formatting and wrapped the call in the PR's try/except NotFoundError block. PR is mergeable_state: clean.

Addressing the remaining review items:

1. test_get_model_info_two_part_embedded_at docstring (R4.3, Minor/Functional Correctness). The previous docstring claimed the embedded @ was 'preserved within the model name portion', which is misleading — the 2-segment branch of _get_model_info treats the right-anchored suffix as provider_name, so text-embedding-nomic-embed-text-v1.5@q8_0 resolves to (model='text-embedding-nomic-embed-text-v1.5', provider='q8_0', instance='default'). The thread was marked resolved because the behaviour is the intended one (legacy tenant_llm bare-model fallback was removed per Lynn-Inf's request, so the 2-part branch has no other option), but the docstring was inaccurate. Tightened in 21756dfdd:
   - States explicitly that the suffix is treated as provider_name
   - Names the design decision (suffix always wins) and its origin (maintainer-requested tenant_llm removal)
   - Removes the 'preserved' claim
   - Test assertions are unchanged; all 5 tests in the file still pass.

2. Out-of-Scope check (dataset_api_service missing-index tolerance). Split out into its own tracking issue: #16649 — [Bug]: HTTP 500 from POST /api/v1/dataset/check_embedding when dataset chunk index is missing. The fix code already shipped in this PR (sample_random_chunks_with_vectors try/except NotFoundError -> return []), but the issue now owns the discussion, regression-test work, and changelog entry for that site.

Re-requesting review from @coderabbitai and @Lynn-Inf — appreciate a final pass once the new commit is visible.

Lynn-Inf (commented):
I've noticed that there is still a `tenant_llm` service call in the `_resolve_bare_model_provider` function in `api/apps/services/models_api_service.py`  (line 65: `TenantLLMService.get_my_llms`). Since `_resolve_bare_model_provider` itself is not called anywhere, I think it can be removed directly. Other parts look good to me.

skbs-eng:
Addressed the remaining review item in `b4db8cb15` per Lynn-Inf's latest comment: removed `_resolve_bare_model_provider` from `api/apps/services/models_api_service.py` entirely.

**Why remove rather than wire in:**
- The function has zero callers anywhere in the codebase (production or tests).
- It referenced `TenantLLMService` without importing it, so it would `NameError` on every call — the broad `except Exception` silently turned the crash into `return ""` (CodeRabbit's Critical 🔴 finding).
- `TenantLLMService` is the deprecated service Lynn-Inf asked to remove in `26a2cec59`, so adding the missing import would re-introduce the very dependency we just deleted.

**Validation:**
- `ruff check` → All checks passed
- `ruff format --check` → already formatted
- `pytest` on the two regression test files → 12 passed

**Diff:** -37 / +5 (5 added lines are isort-driven import reordering after the helper was removed).

Re-requesting review from @Lynn-Inf — would appreciate confirmation this addresses the request, and a final look from @coderabbitai on the new commit.

Lynn-Inf (approved):
LGTM

PR #16484, opened 2026-06-30
fix(model): normalize Go model types to match front-end expectations

euvre:
### Summary

This PR fixes three issues in the Go backend model service that caused model type mismatches with the front-end and an early-return bug that dropped OCR/TTS default models.

**1. Model type tag normalization (`model_service.go`)**

The factory/config JSON files use `"asr"` and `"vision"` as model type tags, but the front-end `ModelTypeMap` filters on `"speech2text"` and `"image2text"` — exactly what Python maps via `MODEL_TAG_TO_TYPE` (`models_api_service.py:38`). The Go port returned the raw config tags verbatim in `ListTenantAddedModels`, so ASR and image-to-text models never appeared in the front-end "View Models" list when filtering by type.

Fix: normalize `"asr"` → `"speech2text"` and `"vision"` → `"image2text"` when building the `model_type` array, mirroring the Python behavior.

**2. Default model type consistency (`tenant.go`)**

`ListTenantDefaultModels` returned `"asr"` / `"vision"` as `ModelType` for the default ASR and image-to-text models, inconsistent with the front-end expectation. Updated to `"speech2text"` / `"image2text"`.

**3. Early-return bug in `ListTenantDefaultModels` (`tenant.go`)**

The previous code returned early when `ownedTenant.OCRID == nil`, which silently dropped the default TTS model (and any subsequent logic). Restructured the OCR and TTS blocks with independent nil checks so each model is only skipped when its own ID is absent.


euvre:
Closing in favor of an alternative approach: instead of normalizing Go model types to the frontend's naming convention, we will align the frontend and Python to Go's native naming convention (`asr`/`vision`).

PR #16495, opened 2026-06-30
fix(go): agent settings update clearing DSL

Hz-186:
### Summary

This PR fixes a Go backend bug where updating agent settings, such as description, could clear the agent DSL.

Root cause:
PUT /api/v1/agents/:canvas_id only bound the dsl field in Go. When the frontend submitted settings without dsl, the service still updated the canvas with an empty DSL value.

Changes:

- Treat agent updates as partial patches.
- Preserve existing DSL when dsl is not present in the request.
- Update only specified user_canvas fields instead of saving the full row.
- Add a regression test for settings updates preserving DSL.

Test:

`go test ./internal/service ./internal/handler`


yuzhichang (approved):
LGTM

PR #16499, opened 2026-06-30
fix(parsing): self-heal 'tasks are ahead in the queue' so stopped parses don't hang

ondery:
## Summary

When parsing documents (especially large files), users keep seeing a stale `N tasks are ahead in the queue...` message (e.g. `34 tasks are ahead in the queue...`) and parsing appears to hang **forever**, even after every parse job has been stopped.

### Root cause

`get_queue_length()` returned the raw Redis stream consumer-group `lag`:

```python
def get_queue_length(priority, suffix="common"):
    group_info = REDIS_CONN.queue_info(settings.get_svr_queue_name(priority, suffix), SVR_CONSUMER_GROUP_NAME)
    if not group_info:
        return 0
    return int(group_info.get("lag", 0) or 0)
```

`lag` counts **every undelivered message** in the stream, including messages whose tasks were already cancelled/stopped. Stopping a parse only sets the DB `run=CANCEL` and the Redis `{task_id}-cancel` flag — it does **not** remove the already-queued messages from the stream. Those messages only stop counting toward `lag` once a task executor happens to read and `ack` them. While an executor is busy on a large file (or after a crash/restart), the cancelled messages stay in `lag`, so the figure never returns to 0 and the document looks stuck in an infinite queue.

Additionally, `_sync_progress` skips documents with `run=CANCEL`, so the last `N tasks are ahead in the queue...` text stays frozen on a stopped document.

## Changes

- **Self-healing queue length**: `get_queue_length()` now caps the raw Redis `lag` by the number of tasks that are *genuinely* still waiting in the database (`Task.progress == 0` on a document whose `run == RUNNING` and `progress` in `[0, 1)`). The value drops back to reality the instant work is cancelled or finishes. A short in-process cache (3 s) avoids issuing a `COUNT` per document on every progress-sync cycle.
- **Safe fallback**: if the DB count can't be determined, it falls back to the previous raw-`lag` behaviour.
- **Clear stale message on stop**: the `N tasks are ahead in the queue...` progress message is cleared when a document parse is stopped/cancelled (single-document run, bulk stop, and per-task stop paths).
- **Tests**: added unit tests covering the capping, self-heal-to-zero, lag-smaller-than-pending, DB-unavailable fallback, missing-group and null-lag cases.

## Test plan

- [x] `python -m py_compile` on all changed files
- [ ] `uv run pytest test/unit_test/api/db/services/test_get_queue_length.py`
- [ ] Manual: queue several large-file parses, stop them, confirm the queue count returns to 0 and documents no longer show a frozen `N tasks are ahead in the queue...`

### DIFF
--- a/api/apps/restful_apis/document_api.py
+++ b/api/apps/restful_apis/document_api.py
@@ -1415,6 +1415,9 @@ def _run_sync(user_id:str, req):
             has_unfinished_task = any((task.progress or 0) < 1 for task in tasks)
             if str(doc.run) in [TaskStatus.RUNNING.value, TaskStatus.CANCEL.value] or has_unfinished_task:
                 cancel_all_task_of(doc_id)
+                # Drop the stale "N tasks are ahead in the queue..." text so a
+                # stopped document does not keep looking like it is still waiting.
+                info["progress_msg"] = ""
             else:
                 return RetCode.DATA_ERROR, "Cannot cancel a task that is not in RUNNING status"
         if all([rerun_with_delete, str(doc.run) == TaskStatus.DONE.value]):
@@ -1644,6 +1647,7 @@ async def stop_parse_documents(tenant_id, dataset_id):
                         "run": str(TaskStatus.CANCEL.value),
                         "progress": 0,
                         "chunk_num": 0,
+                        "progress_msg": "",
                     },
                 )
                 index_name = search.index_name(tenant_id)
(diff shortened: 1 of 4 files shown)


ondery:
Superseded by #16502, which contains these queue-length changes plus the fix for the underlying task-executor freeze (a Langfuse flush/shutdown deadlock on the event loop). Closing in favor of the single consolidated PR.

PR #16510, opened 2026-06-30
fix: classify sub-1.0 decimals as float in column_data_type

Osamaali313:

`column_data_type` (in `rag/app/table.py`) infers a column's type when chunking tabular files. The float branch reused the int branch's leading-zero guard:

```python
elif re.match(r"[+-]?[0-9.]{,19}$", str(a).replace("%%", "")) and not str(a).replace("%%", "").startswith("0"):
    counts["float"] += 1
```

That `not ...startswith("0")` guard is meant to keep leading-zero **integer identifiers** (e.g. `"007"`) out of the numeric buckets. But on the float branch it also rejects every legitimate decimal that begins with `0.` — `"0.5"`, `"0.99"`, `"0.25"` — so a column of ratios / probabilities / prices in `[0, 1)` (and any mixed column containing them) is misclassified as **text**.

This matters because `chunk()` uses the inferred type to choose the stored DB field suffix — `float` → `_flt` (numeric storage: range filters, sorting, Text-to-SQL) vs `text` → `_tks` (tokenized). Misclassifying a numeric column as text silently breaks numeric retrieval over that column.

**Before:**

```
column_data_type(["0.5", "0.7", "0.9"])   -> "text"   ❌
column_data_type(["0.99", "0.50", "0.25"]) -> "text"  ❌
column_data_type(["0.5", "1.5", "0.7"])    -> "text"  ❌ (mixed column degraded)
```

### Type of change

### What is the fix?

Restrict the guard to leading-zero **integer-looking** strings:

```python
elif re.match(r"[+-]?[0-9.]{,19}$", str(a).replace("%%", "")) and not re.match(r"0[0-9]", str(a).replace("%%", "")):
```

`0[0-9]` rejects only leading-zero integers (`00`, `007`, `01234` — identifiers) while allowing `0`, `0.5`, `0.99`. The int branch and all other branches are untouched.

**After:**

```
column_data_type(["0.5", "0.7", "0.9"])   -> "float"
column_data_type(["0.99", "0.50", "0.25"]) -> "float"
column_data_type(["0.5", "1.5", "0.7"])    -> "float"
column_data_type(["007", "008", "009"])    -> non-numeric (unchanged)
column_data_type(["1", "2", "3"])          -> "int"   (unchanged)
```

### Tests

Added regression tests to `test/unit_test/rag/app/test_table_chunk_column_roles.py` exercising the real `column_data_type`: sub-1.0 decimals and a mixed column classify as `float`, while leading-zero identifiers stay non-numeric. Verified RED→GREEN against the real function (the fix flips the three decimal cases from `text` to `float`; the `"007"` case is unchanged).


JinHai-CN:
@Osamaali313 Thank you for your PR. Would you also please add '0.0.1' into test and validate the result?

Osamaali313:
Thanks @JinHai-CN — good catch with `0.0.1`. It exposed a related weakness: the numeric regex `[+-]?[0-9.]{,19}$` also matches strings with more than one `.` (version identifiers like `0.0.1` / `1.2.3`), which are **not** valid floats (`float("0.0.1")` raises), so they were being classified as `float`.

I've pushed a follow-up restricting the float branch to values with at most one `.`. Validated against the real `column_data_type`:

| input | result |
|---|---|
| `["0.0.1", "0.0.2", "0.0.3"]` | `text` |
| `["1.2.3"]` | `datetime` (not float) |
| `["0.5", "0.7", "0.9"]` | `float` |
| `["0.99", "0.50"]` | `float` |
| `["007", "008"]` | non-numeric (unchanged) |
| `["1", "2", "3"]` | `int` |

Added regression tests for `0.0.1` and `1.2.3` alongside the existing sub-1.0 decimal tests.

JinHai-CN:
@Osamaali313 Would you please fix the conflicts?

And how about "12." or "."?

JinHai-CN:
@Osamaali313 please re-open the PR, if you are still working on it.

PR #16522, opened 2026-07-01
fix(go): prevent moving folders into themselves

Hz-186:
## Summary

- Add validation to reject moving a folder into itself or its own subfolder.
- Align Go file move behavior with the existing Python implementation.
- Prevent recursive folder move from causing hanging file management requests.


xugangqiang:
LGTM

PR #16534, opened 2026-07-01
fix: only count chunks that fit within the token budget in kb_prompt

### LINKED ISSUE #16376: kb_prompt renders one chunk past the token budget
## Summary

`kb_prompt` in `rag/prompts/generator.py` can include a retrieval chunk that pushes the prompt over the configured token budget.

## Where

Budget trimming happens in the first loop of `kb_prompt` (~lines 133–141). The loop increments `chunks_num` before checking whether `used_token_count` exceeds `max_tokens * 0.97`.

## Reproduction

1. Build `kbinfos` with several chunks whose combined token count exceeds `max_tokens * 0.97`.
2. Call `kb_prompt(kbinfos, max_tokens=<small limit>)`.
3. Compare the number of rendered chunks with how many should fit under the budget.

## Expected

Only chunks that fit within the 97% budget should be rendered.

## Actual

When chunk *i* is the first to exceed the budget, `chunks_num` is already `i + 1`, so that overflowing chunk is still passed to the rendering loop via `kbinfos["chunks"][:chunks_num]`.

## Proposed fix

Check the budget before incrementing `chunks_num`, and drop the dead `knowledges = knowledges[:i]` assignment (the second loop rebuilds from `kbinfos["chunks"]` anyway).

OSHA-B:
## Summary

`kb_prompt` in `rag/prompts/generator.py` was including one chunk past the configured token budget in the rendered context window.

## Root Cause

The budget-trimming loop incremented `chunks_num` **before** checking whether the running token count exceeded `max_tokens * 0.97`. When chunk `i` pushed the count over the limit:

- `chunks_num` was already `i + 1`
- The rendering loop sliced `kbinfos["chunks"][:chunks_num]` — including the over-budget chunk

The first-loop assignment `knowledges = knowledges[:i]` was dead code because the second loop unconditionally rebuilds `knowledges` from `kbinfos["chunks"][:chunks_num]`.

## Fix

Check the budget **before** incrementing `chunks_num`, so the counter only advances when the chunk fits within the limit:

```python
# before
used_token_count += num_tokens_from_string(c)
chunks_num += 1
if max_tokens * 0.97 < used_token_count:
    knowledges = knowledges[:i]   # dead — second loop overwrites knowledges
    logging.warning(...)
    break

# after
used_token_count += num_tokens_from_string(c)
if max_tokens * 0.97 < used_token_count:
    logging.warning(...)
    break
chunks_num += 1
```

Also removes the dead `knowledges = knowledges[:i]` assignment.

## Files Changed

- `rag/prompts/generator.py` — budget check reordered, dead assignment removed (+2 / -3)

Closes #16376


OSHA-B:
Fixed the `chunks_num` positional slice misalignment as flagged by CodeRabbit.

**The bug:** The budget loop counted only non-empty chunks that fit (`chunks_num += 1`), but the final slice `kbinfos["chunks"][:chunks_num]` is positional. If any earlier chunk had empty content and was skipped, its slot still counted in the slice—so the output would include that empty-content chunk and exclude a valid later one.

**The fix:** Replaced the `chunks_num` counter + positional slice pattern with an `accepted_chunks` list that collects the actual chunk objects as they are accepted. The rendering loop now iterates over `accepted_chunks` directly.

Also removed the unused `i` from the budget-loop `enumerate` (replaced with `zip`) per the trivial cleanup suggestion.

xugangqiang:

**1. This PR is superseded by #16377 (already merged).**
The exact same bug was fixed by commit `3cbf61f5f2` (PR #16377, merged 2026-08-31). #16377 changes the same `kb_prompt` logic and additionally adds 64 lines of regression tests (`test/unit_test/rag/prompts/test_kb_prompt_metadata.py::TestKbPromptTokenBudget`). This PR carries the fix with **no tests** and a slightly less clean implementation (it accumulates `used_token_count` *before* the budget check, so the counter momentarily includes the rejected chunk). Recommendation: close this PR as a duplicate of #16377.

**2. Edge case: a single oversized chunk now yields an empty prompt.**
Both this PR and #16377 break *before* appending when a chunk alone exceeds `max_tokens * 0.97`, so `kb_prompt` can return `[]`. `api/db/services/dialog_service.py:753` then hits the short-circuit at `dialog_service.py:757` (`if not knowledges and prompt_config.get("empty_response")`), which returns the canned "no relevant content" answer **even though chunks were retrieved**. This risk is amplified by `dialog_service.py:743`, where the KG result is `insert(0, ck)` when `use_kg` is enabled — a large KG chunk at position 0 can now wipe out the entire context. Pre-fix, at least that one oversized chunk was included (and later trimmed by `message_fit_in`).

**3. `not knowledges` conflates two distinct states.**
`dialog_service.py:757` treats "retrieved nothing" and "retrieved but doesn't fit the budget" identically. The fix makes the second state reachable more often. These should be distinguished so that retrieved-but-overflowing results don't silently collapse into the empty-response path.

**4. Redundant parallel list.**
`rag/prompts/generator.py:129` builds a `knowledges` list that is unconditionally reassigned at line ~149; it only feeds `kwlg_len` and the `zip` loop. This should be collapsed into a single iteration over `kbinfos["chunks"]` (as the merged #16377 version already partially does by hoisting `chunks`).

**5. Python/Go token-counting divergence (pre-existing).**
Python counts only `content` (`num_tokens_from_string(c)`), whereas the Go mirror `internal/service/kb_prompt.go:72` counts the fully formatted entry (ID/Title/URL/metadata overhead included). Python systematically under-budgets by tens of tokens per chunk. Not introduced by this PR, but worth tracking.

**6. No test coverage for the change (addressing only the fix in this PR).**
The file change has no accompanying tests; existing `test_kb_prompt_metadata.py` cases all use `max_tokens=10000` with a single chunk and never exercise the budget-trimming path. (Already resolved in main via #16377.)

PR #16539, opened 2026-07-01
refactor: use WaitGroup.Go to simplify code

grandpig:
### Summary

Adopt sync.WaitGroup.Go (Go 1.25) to simplify tracked goroutine spawning. This replaces the error-prone trio of wg.Add(1), go func(), and defer wg.Done() with a single, self-contained call.

More info: https://github.com/golang/go/issues/63796


yuzhichang (approved):
LGTM

PR #16544, opened 2026-07-02
fix(go): clear task cancel signals and chunk counters on rerunWithDelete

Hz-186:
## Summary

- Reset doc/KB counters unconditionally on `rerunWithDelete`, not only when `doc.Run == done`.
- Send Redis cancel signals to running parse tasks, preventing stale goroutines from writing chunks during re-parse.
- Avoid duplicate task/chunk deletion between `req.Delete` and `rerunWithDelete` branches.

## Changes

- **`internal/service/document.go`**: Replaced conditional `clearKBChunkNumWhenRerun` with new `prepareDocumentRerunWithDelete` that cancels tasks, deletes task records, resets counters in a `SELECT FOR UPDATE` transaction, and removes ES chunks.
- **`internal/service/document_test.go`**: Added tests for full cleanup and idempotency.


xugangqiang:
LGTM

PR #16554, opened 2026-07-02
fix: prevent reasoning content from leaking into answer in Go chat streaming

euvre:
### Summary

Fix a bug in the Go chat pipeline where chain-of-thought (reasoning) text from reasoning models (DeepSeek, Qwen, SiliconFlow, etc.) leaked into the user-visible answer.

**Root cause:** When a model driver streams reasoning_content separately from content (via the reason callback parameter), the Go pipeline fed that reasoning text through NextThinkDelta() — a function designed to process inline think-tagged text. Since the reasoning text contained no think tags, it was treated as ordinary answer text and emitted directly to AsyncChatResult.Answer, causing the thinking chain to appear in the response body.

**Changes:**

1. internal/service/chat_pipeline.go — three streaming paths fixed (AsyncChat tool path, AsyncChat non-tool path, AsyncChatSolo):
   - Route reason deltas directly to the AsyncChatResult.Reasoning field instead of NextThinkDelta()
   - Wrap reasoning in think markers in thinkState.fullText so ExtractVisibleAnswer can strip it from the final answer
   - Close any open think block (emit EndToThink) before emitting answer content
   - Guard against streams that end while still in think mode (emit closing marker before flush)

2. internal/service/chat_session.go — two SSE streaming handlers (legacy and non-legacy):
   - Accumulate result.Reasoning into fullAnswer between think markers for correct session persistence
   - Use Reasoning as the SSE delta text when Answer is empty

**Testing:**
- go build ./... passes
- go test ./internal/service/ all pass (including ExtractVisibleAnswer and ThinkBlockRE parity tests)


euvre:
duplicate #16558

PR #16606, opened 2026-07-03
fix(agent): enable single-component debug for Agent in Go backend

euvre:
### Summary

This PR fixes two issues that prevented the Agent component's single-component debug/test run from working under the Go backend:

1. **Dynamic input_form generation**: Some components (e.g. `Agent`) do not store a static `input_form` in the DSL. The Go handler now falls back to the runtime component's `GetInputForm()` method, matching Python's `Canvas.get_component_input_form` behavior. This resolves the frontend 102 error: `component has no input_form`.

2. **Tenant ID injection for debug**: Single-component debug runs use a fresh `CanvasState` that previously lacked `tenant_id`. `AgentComponent.Invoke` resolves LLM credentials via the tenant tables, so the debug run failed with `api key is required`. The handler now seeds `state.Sys["tenant_id"]` with the authenticated user's ID, mirroring Python's `@add_tenant_id_to_kwargs` decorator.

### Changes

- `internal/handler/agent_component.go`:
  - Added `componentInputForm` helper that first reads the static `input_form` and, if missing, instantiates the component and calls `GetInputForm()`.
  - In `DebugComponent`, set `debugState.Sys["tenant_id"] = user.ID` before invoking the component.

### DIFF
--- a/internal/handler/agent_component.go
+++ b/internal/handler/agent_component.go
@@ -29,7 +29,9 @@
 package handler
 
 import (
+	"context"
 	"errors"
+	"fmt"
 
 	"github.com/gin-gonic/gin"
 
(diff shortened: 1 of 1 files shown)


yuzhichang:
LGTM

PR #16623, opened 2026-07-03
Update development script

wangq8:


.github/workflows/sep-tests.yml
```diff
@@ -135,7 +135,9 @@ jobs:
             trap 'rm -f "$changed_files"' EXIT
             git diff --name-only ${{ github.event.pull_request.base.sha }}...${{ github.event.pull_request.head.sha }} \
               | while read -r file; do
-                  [[ -f "$file" ]] && printf '%s\0' "$file"
```
wangq8, line 138: old one, when file does not exit, the exit code is 1, the new one will be 0.

PR #16625, opened 2026-07-03
Refactor: pdf parser 

xugangqiang:
### Summary

PDF parser refactor


yuzhichang (changes requested):
Use English in comments.

PR #16646, opened 2026-07-05
fix(web): make sidebar i18n reactive and refine template card/icon UI

yooxdb:
### Summary

Several UI fixes in the agent template and dataset table components:
- Sidebar menu labels (`template-sidebar.tsx`) were computed once at module load using `i18next`'s `t` directly, so they never updated on language change. Moved `menuItems` into the component with `useTranslation` + `useMemo` so labels re-render reactively.
- "Use this template" button on template cards could clip long translated labels since it relied on a fixed `w-1/3` width; replaced with `max-w-[280px]` and `whitespace-normal` so the label wraps instead of being cut off.
- File upload preview icon (`FileText`) looked too bold at this size; added `strokeWidth={1}` to match the rest of the icon set.
- Parsing status tooltip icon used `<sup>` for a question-mark icon, which superscripted and misaligned it; replaced with an inline-flex `<span>` and reduced icon size to align with the percentage text.

|  | Before | After |
| --- | --- | --- |
| *Responsive sidebar i18n + show EN description (fallback) | image | image |
| *Overflowed button label | image | image |
| *Question mark alignment  | 스크린샷 2026-07-01 오후 11 31 43 | image |
| *File icon's strokeWidth | image | image |


yooxdb:
@dcc123456 I just wanted to gently follow up in case this got buried under other activity. Whenever you have a moment, I'd really appreciate a review. Thanks!

dcc123456 (approved):

PR #16651, opened 2026-07-05
fix(connector): propagate WebDAV root listing error to prevent silent prune of all documents

### LINKED ISSUE #16636: [Bug]: Transient WebDAV listing error causes prune to silently delete all synced documents (data loss)
### Self Checks

Unknown — using official infiniflow/ragflow:v0.26.1 Docker image

### RAGFlow image version

infiniflow/ragflow:v0.26.1

```Markdown
Hardware parameters: GMKtec EVO-X2, AMD Ryzen AI Max+ 395, 96GB RAM
OS type: Kubuntu 26.04, Docker Compose deployment (CPU-only image, docker-ragflow-cpu-1)
Others: Data source is a WebDAV connector pointing to an OpenMediaVault (OMV) server with self-signed TLS certificate; "sync deleted files" (prune) enabled with a 5-minute prune/refresh frequency.
```

### Actual behavior

`WebDAVConnector._list_files_recursive()` swallows all exceptions from the underlying WebDAV listing call and silently returns an empty list instead of propagating the failure:

```python
try:
    logging.debug(f"Listing directory: {path}")
    for item in self.client.ls(path, detail=True):
        ...
except Exception as e:
    logging.error(f"Error listing directory {path}: {e}")

return files
```

If `self.client.ls(path, detail=True)` fails for any transient reason (network blip, WebDAV server timeout, brief SSL error, momentary server restart, etc.), this is caught, logged only at ERROR level, and the function returns `files` — an **empty list** if the failure occurred on the very first directory listing call.

This propagates into the prune/reconciliation path in `rag/svr/sync_data_source.py`:

```python
def _collect_prune_snapshot(self, task: dict):
    ...
    try:
        for slim_batch in self.connector.retrieve_all_slim_docs_perm_sync(**snapshot_kwargs):
            file_list.extend(slim_batch)
    except Exception:
        logging.exception(...)
        return None
    return file_list
```

```python
async def _run_prune_task_logic(self, task: dict):
    ...
    file_list = self._collect_prune_snapshot(task)
    if file_list is None:
        ...abort...
        return
    removed_docs, cleanup_errors = ConnectorService.cleanup_stale_documents_for_task(
        task["id"], task["connector_id"], task["kb_id"], task["tenant_id"], file_list,
    )
```

Since `_list_files_recursive()` never raises on a transient listing failure, `_collect_prune_snapshot()` returns a legitimately-typed-but-empty list rather than `None`. The `if file_list is None` guard does not catch this, so `cleanup_stale_documents_for_task()` runs with an empty `file_list`.

In `api/db/services/connector_service.py`:

```python
retain_doc_ids = {doc_id for file in file_list for doc_id in (...)}
existing_docs = DocumentService.list_doc_headers_by_kb_and_source_type(kb_id, source_type)
stale_doc_ids = [doc["id"] for doc in existing_docs if doc["id"] not in retain_doc_ids]
...
for offset in range(0, len(stale_doc_ids), delete_batch_size):
    err = FileService.delete_docs(stale_doc_ids[offset : offset + delete_batch_size], tenant_id)
```

With an empty `file_list`, `retain_doc_ids` is empty, so every existing document synced from that connector is treated as stale and deleted — even though the files are still present and unchanged on the WebDAV server.

Recovery does not happen automatically: the regular incremental sync (`WebDAV._generate()`) uses `poll_source(poll_range_start, end_ts)`, filtered by modification time. Since the deleted files were never actually modified on the server, their `mtime` still falls outside the current polling window, so the next incremental sync does not re-add them. Only a full reindex recovers them.

### Expected behavior

A transient/partial failure while listing the WebDAV directory tree should never be treated the same as "the source directory is genuinely empty." At minimum:

- `_list_files_recursive()` should propagate listing failures (or return a clear sentinel distinguishing "failed" from "empty") instead of silently returning `[]`.
- `_collect_prune_snapshot()` / `_run_prune_task_logic()` should treat an empty `file_list` with the same caution as `None` when the connector previously had a non-zero document count — or at minimum skip the prune cycle and log a prominent warning if the returned count is suspiciously lower than the last known snapshot (e.g., 0 vs. previously 100).
- Ideally, incremental sync should be able to recover documents that exist on the source but are missing from the KB, rather than relying solely on mtime-based polling.

### Steps to reproduce

```Markdown
1. Configure a WebDAV data source with "sync deleted files" (prune) enabled, pointing to a directory with several files.
2. Run an initial full sync — files are indexed successfully and appear under the dataset's Files tab.
3. Simulate a transient failure during the next prune cycle's directory listing (e.g., briefly block/firewall the WebDAV server for a few seconds during the listing call, or introduce a temporary network interruption).
4. Observe the prune task log: it completes with `Success` status and `deleted=<N>` matching the previously-synced file count, even though no files were actually removed from the WebDAV source.
5. Wait for the next incremental sync cycle. Observe that the deleted documents are not re-added, since their modification timestamps predate the current polling window.
```

### Additional information

This was discovered while debugging why a WebDAV knowledge base showed 0 total files despite a successful earlier sync. In our case the initial trigger was a Python SyntaxError crash-looping the entire sync worker (a local misconfiguration, unrelated to this bug), but tracing through `_list_files_recursive`'s broad `except Exception` handlers showed that *any* transient listing failure — network blip, timeout, brief server unavailability — produces the same silent, cascading data-loss behavior, with no error surfaced beyond a generic ERROR-level log line.

Happy to provide further log excerpts or connector config details if helpful.

OSHA-B:
## Summary

Closes #16636

A transient WebDAV listing failure (network timeout, auth error, server outage) could silently delete **every document** in a knowledge base during the next sync prune cycle.

## Root cause

`_list_files_recursive()` swallows the exception from the root-level `client.ls(path)` call and returns an empty `files = []`:

```python
except Exception as e:
    logging.error(f"Error listing directory {path}: {e}")
    # falls through and returns []
return files
```

Downstream, `_collect_prune_snapshot()` only guards against `None`, not `[]`:

```python
if file_list is None:   # empty list passes this guard
    return None
```

So `cleanup_stale_documents_for_task()` receives an empty retain set and deletes everything.

## Fix

Add `raise` after the logging call so the exception propagates up through `retrieve_all_slim_docs_perm_sync()` to `_collect_prune_snapshot()`, where the existing `except Exception` handler catches it, logs it, and **returns `None`** — which already aborts the prune cycle safely.

```python
except Exception as e:
    logging.error(f"Error listing directory {path}: {e}")
    raise   # ← propagate so callers can detect listing failure
```

Sub-directory errors (inner `try/except` blocks) continue to be caught-and-skipped as before — only the root listing failure is now propagated.

## Impact

- Before: any root listing failure → silent deletion of all synced documents.
- After: root listing failure → prune aborted, documents preserved, error logged.

### DIFF
--- a/common/data_source/webdav_connector.py
+++ b/common/data_source/webdav_connector.py
@@ -235,7 +235,8 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
                         
         except Exception as e:
             logging.error(f"Error listing directory {path}: {e}")
-            
+            raise
+
         return files
 
     def _yield_webdav_documents(


6ba3i:
Thank for your contribution! unfortunately it looks like this PR has already been superseded by #18694 
Please feel free to contribute to more features or issues in the future.

PR #16668, opened 2026-07-06
fix(service): allow updating memory_type when memory is empty

euvre:
### Summary

This PR fixes a 500 error when users add memory types to a newly created memory.
Previously, the update check relied on a database-level message count, which incorrectly blocked updates to memory_type, embd_id, and tenant_embd_id even when the memory had no indexed messages.

The fix replaces the DAO count with a doc-engine search scoped to the memory index, so the restriction is only enforced when the memory actually contains messages.

Fixes the 500 error when updating memory_type on an empty memory.

### Changes
- Update MemoryService.UpdateMemory to query the doc engine for existing messages instead of using the DAO message count.
- Add unit tests covering both empty and non-empty memory cases.

### Checklist
- [x] The PR follows the project contribution guidelines.
- [x] Added tests for the fix.
- [x] Local gofmt passes.


wangq8 (commented):
## Review: #16668

**Author:** @euvre
**Title:** fix(service): allow updating memory_type when memory is empty

---

### Core Fix (memory.go)

The change is sound. Replacing `currentMemory.MemorySize` (DAO-level count) with an actual doc-engine search via `listMemoryMessages` is the right approach — the DAO counter can be stale or out of sync with the actual indexed messages, causing false 500 errors.

Key observations:
- The search uses page size 1 with page 1, so it only fetches the total count without loading full results — minimal overhead for an edge-case guard.
- The type assertion `messages["total_count"].(int64)` matches `listMemoryMessages`'s return type — correct.
- Error propagation is properly handled: if the check fails, the update is rejected with a clear error message.

### Concerns

**1. Missing tests**
The PR description says: "Add unit tests covering both empty and non-empty memory cases."
However, no test files appear in the diff (gh pr diff shows only `build.sh`, `memory.go`, and `.env.development`). No `*_test.go` was changed or created in this PR. If tests were intended but not committed, please add them.

**2. Unrelated changes bundled in the same PR**
Two changes are included that have nothing to do with the memory fix:

- **build.sh**: `-- api` to `--api` (space removal). A formatting fix but unrelated.
- **web/.env.development**: `API_PROXY_SCHEME` changed from `'python'` to `'go'`. This is a local dev config change and should not be in a PR fixing memory type updates.

These should be split into separate PRs to keep history clean and make reviews focused.

### Nitpick

The `notAllowedUpdate` slice is still built unconditionally in the loop before the empty-check. Consider moving the doc-engine check into the loop guard for clarity and to avoid making the doc-engine call when none of the restricted fields are being updated:

```go
for _, f := range []string{"tenant_embd_id", "embd_id", "memory_type"} {
    if _, ok := updateDict[f]; ok {
        messages, err := s.listMemoryMessages(...)
        if err != nil { return nil, err }
        if total, ok := messages["total_count"].(int64); ok && total > 0 {
            return nil, fmt.Errorf("can't update %v when memory isn't empty", f)
        }
    }
}
```

### Verdict

**Approve with minor concerns** — the core fix is correct and solves the reported 500 error. Recommend addressing the two items above before merging:
1. Include or add the unit tests
2. Split unrelated changes (build.sh, .env.development) into separate PRs

PR #16673, opened 2026-07-06
Fix filter to use Chinese

wangq8:
image

### DIFF
--- a/api/db/services/api_service.py
+++ b/api/db/services/api_service.py
@@ -19,6 +19,7 @@ import peewee
 
 from api.db.db_models import DB, API4Conversation, APIToken, Dialog
 from api.db.services.common_service import CommonService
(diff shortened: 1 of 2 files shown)


api/db/services/api_service.py
```diff
@@ -65,7 +66,13 @@ def get_list(cls, dialog_id, tenant_id, page_number, items_per_page, orderby, de
         if user_id:
             sessions = sessions.where(cls.model.user_id == user_id)
         if keywords:
-            sessions = sessions.where(peewee.fn.LOWER(cls.model.message).contains(keywords.lower()))
+            keywords = keywords.lower()
+            escaped_keywords = json_dumps(keywords)[1:-1]
```
wangq8, line 70: Trim the double quote "the-word"

PR #16677, opened 2026-07-06
fix(go): add missing 'resume' parser to default tenant parser_ids

euvre:
### Summary

The default parser_ids string used when registering new tenants was missing the 'resume' chunk method. This caused the built-in chunking method dropdown in the dataset creation dialog to not offer 'Resume'.

This PR adds 'resume:Resume' to the default parser_ids in both the password registration path and the OAuth registration path, so that the 'Resume' option is available when creating a dataset.


wangq8:
disable resume.

PR #16679, opened 2026-07-06
fix(web): add missing provider homepage links without changing LLMFactory enum

euvre:

In the model configuration page, some provider cards (e.g., Aliyun, Astraflow, Xiaomi) were missing the homepage link button. This was caused by a mismatch between the provider names returned by the Go backend and the keys used in the frontend `APIMapUrl` / `IconMap` objects. Those maps were keyed by the legacy `LLMFactory` enum values, while the Go backend returns newer/shortened names.

### Type of change

### Changes

- Kept `LLMFactory` enum values unchanged to preserve provider retrieval and other existing logic.
- Added string-keyed entries to `APIMapUrl` for all providers returned by the Go backend that did not already have a matching key, so the homepage link button now renders for every provider.
- Added corresponding string-keyed entries to `IconMap` so provider icons can also be resolved by the new names.

### Related issues

None.

### How to verify

1. Set `API_PROXY_SCHEME='go'` in `web/.env.development`.
2. Open the model configuration page.
3. Confirm that previously missing providers (e.g., Aliyun, Astraflow, Xiaomi, Baidu, HuaweiCloud) now display the external-link homepage button.


Lynn-Inf (changes requested):
Frontend enum changes must be synced with backend factory config — please update both sides accordingly.

dcc123456 (commented):
The naming change here causes model provider retrieval to fail; it cannot be modified this way.

PR #16417, opened 2026-06-27
fix(agent/tools): port AkShare to ToolBase so it works as an Agent tool

immuhammadfurqan:
### What problem does this PR solve?

Closes #16416.

The **AkShare** agent tool (`agent/tools/akshare.py`) was never ported to the modern `ToolBase`/`_invoke` interface during the agent module redesign and was still written against the removed legacy `_run`/`be_output` API, so it was non-functional:

1. **Adding it to an Agent raised `AttributeError`.** `AkShare` extended `ComponentBase` (not `ToolBase`) and `AkShareParam` defined no `meta`, so it had no `get_meta()`. `agent/component/agent_with_tools.py` builds each tool's function descriptor via `cpn.get_meta()`, so constructing an Agent that includes the AkShare tool raised `AttributeError: 'AkShare' object has no attribute 'get_meta'`.
2. **It could never run.** `invoke()` dispatches to `self._invoke`, but `AkShare` only implemented the legacy `_run`, so `_invoke` fell through to `ComponentBase._invoke` → `NotImplementedError`. `_run` also called `be_output(...)`, which no longer exists on the base classes.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

### Changes

- Port `AkShareParam` to `ToolParamBase` with a `ToolMeta` (defined before `super().__init__()`, matching `ArXivParam`/`TavilyExtractParam`) exposing a required `query` parameter — the stock symbol to look up, default `{sys.query}`. `query` matches the `{sys.query}` convention shared by the other tools.
- Rewrite the component with `_invoke`/`set_output("formalized_content", ...)` (errors surfaced via `_ERROR`), keeping `top_n` and importing `akshare` lazily.
- Add regression tests (`test/unit_test/agent/component/test_akshare.py`) covering param construction, validation, and the tool descriptor.

Same class of defect as #16329 (DeepL) and #16414 (Crawler). Backend-only; no frontend changes.

PR #16692, opened 2026-07-07
fix(agent): port QWeather to ToolBase so it works as an Agent tool

boskodev790:
### Summary

Port the **QWeather** agent tool to the modern `ToolBase` / `_invoke` interface. It was still written against the removed legacy `ComponentBase` / `_run` / `be_output` API, so it was non-functional as an Agent tool — adding it to an Agent raised `AttributeError` because it had no `get_meta()`. This is the same defect that was fixed for the AkShare tool in #16417.

**Changes**
- `QWeatherParam` now extends `ToolParamBase` with a `meta` exposing a `query` (location) parameter, and adds `get_input_form()`. Existing config (`web_apikey`, `lang`, `type`, `user_type`, `time_period`) is preserved.
- `QWeather` now extends `ToolBase` and implements `_invoke(**kwargs)` with the standard retry loop, cancellation checks, `set_output("formalized_content", ...)`, and `thoughts()`. The weather / indices / air-quality branches and the API error-code messages are kept.
- Added `test/unit_test/agent/component/test_qweather.py` covering the restored `meta`, param validation, the weather-now and multi-day and indices branches, the empty-query short-circuit, and the location-lookup error message.

**Testing**
- `ruff check agent/tools/qweather.py test/unit_test/agent/component/test_qweather.py` — clean
- `ruff format --check` — clean
- `pytest test/unit_test/agent/component/test_qweather.py`

### DIFF
--- a/agent/tools/qweather.py
+++ b/agent/tools/qweather.py
@@ -13,19 +13,36 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
+import logging
+import os
+import time
 from abc import ABC
-import pandas as pd
+
 import requests
-from agent.component.base import ComponentBase, ComponentParamBase
+
+from agent.tools.base import ToolMeta, ToolParamBase, ToolBase
+from common.connection_utils import timeout
 from common.http_client import DEFAULT_TIMEOUT
 
 
-class QWeatherParam(ComponentParamBase):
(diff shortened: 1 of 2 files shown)


yuzhichang (approved):
LGTM

PR #16693, opened 2026-07-07
fix(ao-agent): add Tavily input_form, wire real search component, and port TavilyExtract to Go

Hz-186:
### Summary

- TavilySearch now stores api_key from component params and injects it into tool calls when runtime inputs omit it.
- TavilyExtract and BGPT now follow the same stored api_key behavior.
- Canvas decoding now recovers api_key from graph.nodes[].data.form when components[].obj.params.api_key is empty, matching frontend payload behavior without changing frontend data.
- Added regression tests for graph form key recovery and stored key injection / caller key precedence.

Tests: build.sh --test ./internal/agent/component ./internal/service — all pass.

image

image


internal/agent/canvas/decode.go
```diff
@@ -118,6 +120,65 @@ func decodeComponentFields(comp map[string]any) (name string, params map[string]
 	return
 }
 
+func decodeGraphNodeForms(dsl map[string]any) map[string]map[string]any {
+	graph, ok := dsl["graph"].(map[string]any)
```
yuzhichang: Backend shall use `components` instead of `graph`.

yuzhichang (changes requested):

PR #16694, opened 2026-07-07
Fix cannot run raptor

wangq8:
Follow on PR #16515

### DIFF
--- a/rag/svr/task_executor_refactor/task_handler.py
+++ b/rag/svr/task_executor_refactor/task_handler.py
@@ -237,7 +237,9 @@ class TaskHandler:
                 return
 
             # Route to appropriate handler
-            if task_type == "graphrag":
+            if task_type == "raptor":
+                await self._run_raptor(embedding_model, vector_size)
+            elif task_type == "graphrag":
                 await self._run_graphrag(embedding_model)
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
## Review: LGTM ✅

**Summary:** This fixes a bug where `task_type == "raptor"` tasks would fall through all routing branches without being handled, effectively doing nothing.

### What was missing
The routing block only handled `graphrag`, `mindmap`, `artifact`, and `dataflow`. RAPTOR tasks had no matching branch.

### What the fix does
- Adds `if task_type == "raptor":` before the existing branches
- Changes the existing `if task_type == "graphrag":` to `elif` to maintain a proper chain

### Verification
- `embedding_model` and `vector_size` are both initialized at line 229 (before the routing block) ✅
- `_run_raptor()` signature is `(self, embedding_model: LLMBundle, vector_size: int, mark_done: bool = True)` — the call passes those two positional args correctly ✅

### Risk assessment
- **Regression risk:** Very low. Only adds a new branch; existing behavior untouched.
- **Security/performance:** No concerns. Single `await` call, no new imports or data flow changes.

Nice catch and minimal fix — lands cleanly.

PR #16697, opened 2026-07-07
Fix: ValueError: Operation on closed image

wangq8:


wangq8 (commented):
## Review: Fix ValueError: Operation on closed image

PR #16697 | bugfix/fix-Operation-on-closed-image -> main | +24/-32 (1 file)

### What changed

A focused refactor of encode_image() inside image2id in rag/utils/base64_image.py, and reordering of the image lifecycle.

### Key analysis

1. Root cause & fix are correct
   
   The ValueError: Operation on closed image occurs when PIL's save() or convert() is called on an image whose underlying data has been freed. The original code would call orig_img.close() before the save operation when the image was RGBA/P mode. If the converted image shared pixel data with the palette-based original (P-mode), the save would fail with Operation on closed image.
   
   The fix applies img.load() before any conversion. This eagerly decodes pixels into memory, breaking the data-sharing link with the original. Both the original and converted images are tracked in owned_images and closed together in finally. This correctly prevents the race/ordering issue.

2. Pop early pattern
   
   Moving image = d.pop("image") to the top removes the image reference from the dict immediately, so async continuation has no stale dict access. The old pattern of deleting it at two separate points is consolidated.

3. Cleaner bytes path
   
   When open_image_for_processing returns bytes, returning bytes(img) avoids the unnecessary BytesIO round-trip. Good micro-optimisation.

4. Exception coverage
   
   except (OSError, ValueError) catches both OSError (PIL save failures) and ValueError (the closed-image error itself as a fallback).

### Potential concern (minor)

Edge case: open_image_for_processing throws. If this function raises an exception after image = d.pop("image") has already run, but before owned_images is populated, the raw image value is never closed. In practice open_image_for_processing handles its own cleanup on failure, so this is low risk. Worth a quick check.

### Verdict

LGTM, good fix. Clean, minimal, and addresses the exact bug pattern.

PR #16698, opened 2026-07-07
fix(go-pipline): list agents incorrectly filtering out ingestion pipelines

Hz-186:
### Summary

When no canvas_category filter is provided, the DAO defaulted to "agent_canvas", hiding canvases with category "dataflow_canvas" (i.e. ingestion pipelines). Remove the hardcoded default so both categories are returned.

Verified by frontend testing.

image

### DIFF
--- a/internal/dao/user_canvas.go
+++ b/internal/dao/user_canvas.go
@@ -288,9 +288,6 @@ func (dao *UserCanvasDAO) GetList(tenantID string, pageNumber, itemsPerPage int,
 	}
 	if canvasCategory != "" {
 		query = query.Where("canvas_category = ?", canvasCategory)
-	} else {
-		// Default to agent category
-		query = query.Where("canvas_category = ?", "agent_canvas")
 	}
 
 	if canvasType != "" {
@@ -385,8 +382,6 @@ func (dao *UserCanvasDAO) ListByTenantIDs(ownerIDs []string, userID string, page
 
 	if canvasCategory != "" {
 		base = base.Where("user_canvas.canvas_category = ?", canvasCategory)
-	} else {
-		base = base.Where("user_canvas.canvas_category = ?", "agent_canvas")
 	}
 
 	if canvasType != "" {


wangq8 (commented):
## Review Summary

**PR:** #16698 — fix(go-pipeline): list agents incorrectly filtering out ingestion pipelines

### What it does
Removes a hardcoded  default filter from two DAO methods ( and ) so that when no  query param is provided, all canvas categories (including  / ingestion pipelines) are returned instead of only  ones.

### Analysis

**✅ The bug is real.**
In both  (line 291-293) and  (line 385-387), the  branch unconditionally filtered to  when no category was specified. Any upstream caller that intentionally passes an empty  to mean "no filter" would silently show only agent canvases. This is exactly the reported issue — ingestion pipelines (category ) were hidden from the list view.

**✅ Correct approach.**
The fix removes the  default entirely. Since the  guard already handles the filter-when-set case, removing the else clause means an empty  passes through as "no category filter." This is the semantically correct behavior for an optional filter parameter.

**🔍 Caller analysis**

| Caller | How canvasCategory is set | Impact |
|---|---|---|
|  |  — user-supplied, empty = all | ✅ Fix correctly returns all categories |
|  | Explicitly defaults to  when empty (line 350-351) | ✅ Not affected — hardcodes default at creation time, not query time |
| Other callers of  | None found in non-test code | ✅ No regressions |

**⚠️ Minor concern —  appears to be uncalled.**
The  method on  doesn't seem to have any active callers outside tests. It's worth considering whether this method can be removed entirely, but that's out of scope for this PR.

**✅ No security concerns.**
This is a data-fetching change only — removing a restrictive filter doesn't bypass any authorization since the caller-side tenant/permission checks remain intact.

**✅ No SQL injection risk.**
The  value is passed as a parameterized query placeholder (), not string-interpolated.

### Conclusion

Clean, focused fix. Net -5 lines, no new code introduced. Zero regression risk for the affected code paths. Approved from a code perspective.

wangq8 (commented):
## PR Review: #16698

**Title:** fix(go-pipeline): list agents incorrectly filtering out ingestion pipelines
**Author:** @Hz-186 | **Files:** 1 (internal/dao/user_canvas.go) | **Changes:** -5 lines

### The Bug

When `GET /api/v1/agents` is called without a `canvas_category` query param, both DAO methods (`GetList` and `ListByTenantIDs`) had an `else` branch that hardcoded `canvas_category = "agent_canvas"`. This silently filtered out canvases with category `"dataflow_canvas"` (ingestion pipelines).

### The Fix

Remove the `else` default filter from both methods:

1. **`GetList`** (lines 291-293) — `else { query.Where("canvas_category = ?", "agent_canvas") }` removed
2. **`ListByTenantIDs`** (lines 385-387) — same

This is semantically correct: an optional filter parameter should apply when set and do nothing when empty.

### Caller Impact

| Caller | canvasCategory source | Impact |
|---|---|---|
| `handler/agent.go:ListAgents` | `c.Query("canvas_category")` — user-supplied, empty = no filter | Fix works correctly |
| `handler/agent.go:CreateAgent` | Defaults to `"agent_canvas"` when empty (line 350) | Creation-time default, unaffected |
| Other `GetList` callers | None found outside tests | No regression |

### Verdict

- Correct fix — matches expected semantics for an optional filter parameter
- Zero regression risk — 5 lines deleted, no new code; all callers verified
- No security/injection concerns — parameterized query (`?`), auth checks unaffected
- Nice-to-have cleanup — `GetList` appears to have zero active callers; could be removed separately

**Status: Approved** — clean, focused, correct.

PR #16714, opened 2026-07-07
fix: prevent memory name from auto-appending (1) on description update

euvre:
### Summary

When updating a memory's description (or any other field) via the memory settings page, the backend always runs the name through `duplicate_name`, even if the name in the update request matches the current name. This causes the name to be appended with `(1)` every time the form is saved, even though the user did not change the name.

### Fix

In `MemoryService.update_by_id`, check the existing memory name for the current `memory_id` and only call `duplicate_name` when the requested name is actually different from the current one.

### Verification

The issue was reproduced on the memory settings page: after saving a description change, the memory name changed from `mmr` to `mmr(1)`. With the fix, the name stays `mmr` while the description is updated.

/ci

### DIFF
--- a/api/db/services/memory_service.py
+++ b/api/db/services/memory_service.py
@@ -150,7 +150,9 @@ class MemoryService(CommonService):
         if "memory_type" in update_dict and isinstance(update_dict["memory_type"], list):
             update_dict["memory_type"] = calculate_memory_type(update_dict["memory_type"])
         if "name" in update_dict:
-            update_dict["name"] = duplicate_name(cls.query, name=update_dict["name"], tenant_id=tenant_id)
+            existing = cls.model.select().where((cls.model.id == memory_id) & (cls.model.tenant_id == tenant_id)).first()
+            if not existing or existing.name != update_dict["name"]:
+                update_dict["name"] = duplicate_name(cls.query, name=update_dict["name"], tenant_id=tenant_id)
         update_dict.update({"update_time": current_timestamp(), "update_date": get_format_time()})
 
         return cls.model.update(update_dict).where(cls.model.id == memory_id).execute()


wangq8 (commented):
## Review Summary

**PR**: #16714 — fix: prevent memory name from auto-appending (1) on description update  
**Author**: @euvre  
**Scope**: 3 additions, 1 deletion in `api/db/services/memory_service.py`  
**Status**: LGTM with one optional suggestion

---

### What the fix does

In `MemoryService.update_memory`, before passing the name through `duplicate_name()`, the code now fetches the existing memory row and only calls `duplicate_name` when the existing record is missing or its stored name differs from the incoming name.

This prevents the `(1)` suffix from being appended whenever a user saves unrelated field changes (e.g., description) on the memory settings page.

### Correctness

The logic is sound:

| Scenario | `existing` ? | `existing.name != update_dict["name"]` ? | `duplicate_name` called? | Outcome |
|---|---|---|---|---|
| New record (not yet persisted) | `None`/falsy | N/A | Yes | Correct — dedup still runs |
| Name actually changed | Found | `True` | Yes | Correct — dedup still runs |
| Name unchanged, other field updated | Found | `False` | No | **Bug fixed** — no unwanted suffix |

### Regression risk

Very low. The guard only *skips* `duplicate_name` where it was previously always called. The old behavior in that case was demonstrably buggy, so removing it is strictly an improvement. No downstream callers are affected.

### Edge cases considered

- **Empty/None `update_dict`**: Already handled by the guard at the top of the method.
- **Missing `name` key**: Already handled by the `if "name" in update_dict` check.
- **Concurrent updates**: Minor TOCTOU window exists, but this is a pre-existing concern, not introduced by this change.
- **Multiple memories sharing a name under the same tenant**: Still handled correctly — `duplicate_name` runs when the name actually differs.

### Code quality

Clean, minimal, self-documenting. Consistent with the rest of the method style.

### Tests

Manual verification confirmed the fix. No automated test, which is acceptable for a 3-line bugfix.

### Optional suggestion

Adding logging for the dedup-gating decision path (as CodeRabbit suggested) would be nice for production traceability, but is entirely optional.

---

**Verdict**: Good catch and well-scoped fix. Merge when ready.

PR #16737, opened 2026-07-08
Fix: right-anchored split in web parseModelValue / parseModelUuid

skbs-eng:
Follow-up to #16468.

PR #16468 fixed the Python (`api/db/joint_services/tenant_model_service.py
split_model_name`) and Go (`internal/service/model_service.go parseModelName`)
parsers to right-anchor the '@'-split on the composite
"model_name@instance@provider" key, but the matching helper in the front-end
(`web/src/utils/llm-util.ts parseModelValue`) was missed.

parseModelValue used `split('@')` (anchored on the first '@'), which silently
mangles composite keys whose model_name itself contains '@' (e.g. LM Studio
embedding IDs like `text-embedding-nomic-embed-text-v1.5@q8_0`):

  text-embedding-nomic-embed-text-v1.5@q8_0@lmstudio@LM-Studio

became

  model_name:     text-embedding-nomic-embed-text-v1.5
  model_instance: q8_0@lmstudio
  model_provider: LM-Studio

`PATCH /api/v1/models/default` then sent those to the server, which returned
HTTP 200 with body `{"code": 102, "message": "Instance 'q8_0@lmstudio' not
found for provider 'LM-Studio'"}`. The UI swallowed the body-level error
code, so the failure was silent: no toast, the Embedding field stayed empty.

This change makes parseModelValue do a right-anchored split that mirrors the
Python `split_model_name` `rsplit('@', 2)` exactly:

  - 3-part form: model_name@instance@provider -> same three fields.
  - 2-part form: model_name@provider -> model_instance defaults to "default"
    (matching Python).
  - 4+-part form (embedded '@' in model name): last two fields are anchored
    as instance and provider, everything to the left is the bare model name.

`parseModelUuid` is updated to the same right-anchored split so the
factoryId portion of `model_name@factory_id[#instance]` keeps the last '@'
as the separator even when the model name itself contains '@'.

Adds jest cases under `web/src/utils/tests/llm-util.test.ts` (12 total)
covering the plain 3-part, 2-part, 4-part, multi-'@' name, buildModelValue
round-trip and parseModelUuid variants.

Note for reviewers: the front-end is bundled into the published docker
image, so a release containing this fix will need `docker compose build
--no-cache ragflow-cpu` (or equivalent) for it to be visible to end users.

Refs #16468, #16467

### DIFF
--- a/web/src/utils/llm-util.ts
+++ b/web/src/utils/llm-util.ts
@@ -43,29 +43,64 @@ export function buildModelValue(model: {
   return `${model.model_name}@${model.model_instance}@${model.model_provider}`;
 }
 
-/** Parse "modelName@instanceName@providerName" */
+/**
+ * Parse "modelName@instanceName@providerName" (or the 2-part
+ * "modelName@providerName" form where the instance defaults to "default").
+ *
+ * The composite key is right-anchored: the *last* '@'-separated field is the
+ * provider, the second-to-last is the instance, and everything to the left of
+ * the second-to-last '@' is the bare model name. Some model names legitimately
+ * contain '@' themselves (e.g. LM Studio embedding IDs such as
+ * `text-embedding-nomic-embed-text-v1.5@q8_0`), producing four-`@` composite
+ * keys like `text-embedding-nomic-embed-text-v1.5@q8_0@lmstudio@LM-Studio`.
+ *
+ * A naive `split("@")` (or anchoring on the first '@') mis-parses these keys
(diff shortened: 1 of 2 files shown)


skbs-eng:
@yuzhichang @Lynn-Inf @wangq8 please review

Lynn-Inf (approved):
LGTM

skbs-eng:
please merge, thank you!

skbs-eng:
@yuzhichang @Lynn-Inf @wangq8 please merge

skbs-eng:
@dcc123456 please merge

PR #16745, opened 2026-07-08
Fix(go): adapt to new db columns

Lynn-Inf:
### Summary

As title.


xugangqiang:
LGTM. One thing to note: please help to add unit test for the changed method. @Lynn-Inf

PR #16750, opened 2026-07-08
Fix: prevent auto-incrementing memory name suffix when only permissions change

euvre:
### Summary

This PR fixes an issue where toggling a memory's permission (e.g., switching from "Me" to "Team") caused the memory name to keep appending/incrementing a numeric suffix, such as `mmr` → `mmr(3)` → `mmr(4)`. This happened even when the user did not change the memory name.

### Root cause

In both the Python and Go update paths, the `name` field was always processed with the duplicate-name helper whenever it was present in the request. As a result, an unchanged name was re-checked against existing records and incorrectly renamed.

### Changes

- `api/db/services/memory_service.py`: When the provided `name` equals the current memory name, remove it from the update dictionary instead of re-running `duplicate_name`.
- `internal/service/memory.go`: Added a check in `UpdateMemory` so `common.DuplicateName` is only called when the requested name differs from the current name. Also removed a redundant second fetch of the current memory.

### Verification

- Built the Go server successfully with `bash build.sh --go`.
- Ran `bash build.sh --test ./internal/service/...` and all tests passed.
- Manually confirmed that changing only permissions no longer appends a numeric suffix to the memory name.

### Screenshots

Memory name stays unchanged after switching permission to Team

### Related issues

N/A

### Type of change

### Checklist

- [x] I have checked that the Go server builds successfully.
- [x] I have run the relevant package tests.
- [x] I have verified the fix in the UI.

### DIFF
--- a/api/db/services/memory_service.py
+++ b/api/db/services/memory_service.py
@@ -154,7 +154,9 @@ class MemoryService(CommonService):
             update_dict["memory_type"] = calculate_memory_type(update_dict["memory_type"])
         if "name" in update_dict:
             existing = cls.model.select().where((cls.model.id == memory_id) & (cls.model.tenant_id == tenant_id)).first()
-            if not existing or existing.name != update_dict["name"]:
+            if existing and existing.name == update_dict["name"]:
+                update_dict.pop("name", None)
+            else:
                 update_dict["name"] = duplicate_name(cls.query, name=update_dict["name"], tenant_id=tenant_id)
         update_dict.update({"update_time": current_timestamp(), "update_date": get_format_time()})
 
--- a/internal/service/memory.go
+++ b/internal/service/memory.go
@@ -452,19 +452,26 @@ func (s *MemoryService) CreateMemory(tenantID string, req *CreateMemoryRequest) 
 func (s *MemoryService) UpdateMemory(tenantID string, memoryID string, req *UpdateMemoryRequest) (*CreateMemoryResponse, error) {
 	updateDict := make(map[string]interface{})
 
+	currentMemory, err := s.memoryDAO.GetByID(memoryID)
+	if err != nil {
+		return nil, fmt.Errorf("memory '%s' not found", memoryID)
+	}
+
 	if req.Name != nil {
 		memoryName := strings.TrimSpace(*req.Name)
 		if err := common.ValidateName(memoryName); err != nil {
 			return nil, err
 		}
-		memoryName, err := common.DuplicateName(func(name string, tid string) bool {
-			existing, _ := s.memoryDAO.GetByNameAndTenant(name, tid)
-			return len(existing) > 0
-		}, memoryName, tenantID)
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
## Review: PR #16750

**Summary:** Fixes a bug where toggling only a memory's permissions would cause the memory name to auto-increment with a numeric suffix (e.g., `mmr` → `mmr(3)` → `mmr(4)`).

---

### Analysis

#### ✅ Python side (`api/db/services/memory_service.py`)
- **Correct:** When `existing.name == update_dict["name"]`, the name is popped from the update dict so `duplicate_name` is never called. This directly addresses the root cause.
- **Edge case preserved:** If `existing` is `None` (memory not found), behavior is unchanged — falls through to the `else` branch which calls `duplicate_name` as before.
- **Minor concern:** Unlike the Go side, there is no whitespace normalization before comparison. If a request sends `"mmr "` (trailing space) and the stored name is `"mmr"`, the comparison would falsely report them as different, triggering `duplicate_name`. This is **not a regression** (the old code had the same issue), but trimming both sides for consistency with the Go implementation would be a nice touch.

#### ✅ Go side (`internal/service/memory.go`)
- **Correct:** The `GetByID` call is moved to the top of the function so the current name is available early for comparison. The guard `if memoryName != strings.TrimSpace(currentMemory.Name)` correctly skips `DuplicateName` when the name is unchanged.
- **Cleanup:** The redundant second `GetByID` call at the bottom is removed, and the early empty-update return still works because `"name"` is only added to `updateDict` when it is actually changed.
- **Proper trimming:** Both `memoryName` and `currentMemory.Name` are trimmed before comparison, which is more robust than the Python side.

### 🔒 Security / Regressions
- **No new vulnerabilities.** The fix only avoids calling `duplicate_name` when the name is unchanged; all validation paths remain intact.
- **No regressions** in the logic for actual name changes — they still pass through `duplicate_name` as before.
- **TOCTOU race:** A race condition exists (another request could change the name between the fetch and the update), but this is pre-existing and not a regression.

### 🧪 Tests
- The author ran `bash build.sh --test ./internal/service/...` and all tests passed.
- **However, no new tests were added** that explicitly verify the fix (e.g., a test that updates only permissions and asserts the name stays the same). Existing test coverage for the memory service itself appears thin — there are no dedicated `*memory_service*_test.go` files, only `memory_message_service_test.go`. Adding a focused test would guard against future regressions.

### Style
- Clean, minimal changes. The Go diff also eliminates a redundant DB query, which is a nice bonus.

### Verdict

The fix is **correct and well-understood**. The root cause is clear — `duplicate_name` was unconditionally called on the name field when it was present in the request — and both the Python and Go changes resolve it properly.

**Minor suggestion:** Consider using `.strip()` on the Python side before the name comparison, for consistency with the Go implementation. And adding a regression test for this scenario would provide valuable long-term protection.

PR #16753, opened 2026-07-08
Fix: XLSX multi-sheet files being parsed as first sheet only

ju123123:
## What & why

Multi-sheet `.xlsx` files were ingested as if they only contained their
largest single worksheet — rows belonging to the other sheets never made it
into chunks.

The root cause is in **task pagination**, not in the parser. When a table
document is queued, `ChunkService.tableParseTaskRanges`
(`internal/service/chunk/chunk.go`) splits it into 3000-row parse tasks based
on an estimated row count:

```go
rows := estimateTableRowCount(docName(doc), binary)
for row := 0; row < rows; row += 3000 {
    ranges = append(ranges, parsePageRange{from: row, to: min(row+3000, rows)})
}
```
The XLSX branch of that estimate returned the row count of the largest
single worksheet (maxRows), not the workbook total. So for a workbook with
sheets of, say, 2000 / 2000 / 2000 rows, the estimate was 2000 and only one
task covering rows [0, 2000) was created — rows 2000–6000 (the 2nd and 3rd
sheets) had no task covering them and were dropped.

The Go XLSX parser (internal/parser/parser/xlsx_parser.go) already
iterates every sheet, so the content was available — it just never got a task
range that reached it.

## Fix
Add utility.CountSpreadsheetRows (internal/utility/spreadsheet_rows.go),
which counts XLSX rows by summing every worksheet via excelize, so the
generated task ranges cover the whole workbook.
Collapse the three near-duplicate row-counting implementations in
chunk.go, dataset.go, and document.go into that single shared helper.

### Why not deal with`.xls` files?

`.xls` doesn't suffer from this bug. It isn't handled by
`CountSpreadsheetRows`, so it returns `0`, and the caller falls back to a single
task covering the whole document (`tableParseTaskRanges` → `[0,
maximumTaskPageNumber]`) — every sheet is already processed. The dropped-sheet
bug only occurs when a *wrong positive* row count is returned (the old
max-of-sheets logic), which only the `.xlsx` path produced.

Adding an `.xls` branch wouldn't help anyway: excelize (the only spreadsheet
library in `go.mod`) reads the OOXML family only, not the legacy BIFF `.xls`
binary format, so it would fall back to the same single-task path.

## Scope
Go-only. Per the ongoing Python→Go migration, no Python code is touched.

## Testing
Unit test internal/utility/spreadsheet_rows_test.go: a workbook with a
3-row sheet and a 5-row sheet → CountSpreadsheetRows returns 8 (sum),
which is distinct from the old max-of-sheets result (5) and from
first-sheet-only (3), so the test fails on either regression.

``` shell
$ go test ./internal/utility/ -run TestCountSpreadsheetRowsUsesSummedXLSXRows -v
=== RUN   TestCountSpreadsheetRowsUsesSummedXLSXRows
--- PASS: TestCountSpreadsheetRowsUsesSummedXLSXRows (0.00s)
PASS
ok  	ragflow/internal/utility	0.478s
```

### DIFF
--- a/internal/utility/spreadsheet_rows.go
+++ b/internal/utility/spreadsheet_rows.go
@@ -1,0 +1,85 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+package utility
+
+import (
+	"bytes"
+	"encoding/csv"
+	"io"
+	"path/filepath"
+	"strings"
+
+	"github.com/xuri/excelize/v2"
+)
+
+// CountSpreadsheetRows estimates how many logical rows a table document contains.
+// XLSX sums all worksheet rows so task pagination covers every sheet.
+func CountSpreadsheetRows(name string, binary []byte) int {
+	switch strings.ToLower(filepath.Ext(name)) {
+	case ".xlsx":
+		if rows, err := countXLSXRows(binary); err == nil {
+			return rows
+		}
+	case ".csv", ".tsv", ".txt":
+		return countDelimitedRows(name, binary)
+	}
+	return 0
+}
+
+func countDelimitedRows(name string, binary []byte) int {
+	reader := csv.NewReader(bytes.NewReader(binary))
+	reader.FieldsPerRecord = -1
+	reader.ReuseRecord = true
+	if strings.EqualFold(filepath.Ext(name), ".tsv") {
+		reader.Comma = '\t'
+	}
+	rows := 0
+	for {
+		_, err := reader.Read()
+		if err == nil {
+			rows++
+			continue
+		}
+		if err == io.EOF {
+			break
+		}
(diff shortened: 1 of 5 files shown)


internal/utility/spreadsheet_rows.go
```diff
@@ -0,0 +1,95 @@
...
+func CountSpreadsheetRows(name string, binary []byte) int {
+	switch strings.ToLower(filepath.Ext(name)) {
+	case ".xlsx":
+		if rows, err := CountXLSXRows(binary); err == nil {
+			return rows
+		}
+	case ".csv", ".tsv", ".txt":
+		return CountDelimitedRows(name, binary)
+	}
+	return 0
```
xugangqiang, line 40: How about xls? Is it possible to have other formats?

ju123123: Thanks for the suggestion. During my testing, I did not observe this issue with `.xls` files. The issue only occurs with `.xlsx` files, so this PR focuses on fixing the `.xlsx` multi-sheet parsing behavior.

xugangqiang:
@ju123123 
Thanks for contribution. A few comments:
1. please make sure your PR  description is accurate
2. please add unit test
3. please provide testing  evidence - your fix could really parse the file correctly

xugangqiang:
@ju123123  We are in the middle of migration from python to golang. Thus please only fix in golang side and DO NOT touch any python code. Thank you.

ju123123:
@xugangqiang 

> @ju123123 We are in the middle of migration from python to golang. Thus please only fix in golang side and DO NOT touch any python code. Thank you.

Got it. I’ll only make changes on the Go side and won’t touch any Python code.

xugangqiang:
@ju123123  Is it possible to test the fix with a real excel file and see what's the actual output? If you have done so, could you please upload testing evidence? Like log, screenshot, etc. That will be very helpful. Thank you.

JinHai-CN:
I would like to close this PR, if you want to fix the conflicts and reply the comments of @xugangqiang, please re-open this PR.

PR #16756, opened 2026-07-08
fix(web): prevent 'last saved at' label from vertical stacking in agent home card

euvre:
### Summary

The Chinese label "上次保存于" in the agent home card was wrapping character-by-character when the card width was constrained, causing it to display vertically instead of horizontally. This PR fixes that by keeping the label on a single line and letting the timestamp truncate if space is limited.

### Changes

- Added `whitespace-nowrap` to the `lastSavedAt` / `publishedAt` labels in `HomeCard`.
- Added `truncate` to the timestamp so it no longer pushes the label out of its bounding box.
- Added `min-w-0` to the flex containers to allow proper overflow handling in nested flex layouts.

### Checklist

- [x] I have verified the fix with the existing UI layout.
- [x] The change is scoped to `web/src/components/home-card.tsx`.
- [x] CI checks are expected to run.

/kind bug
/label ci


wangq8 (commented):
## Review: PR #16756 — `fix(web): prevent 'last saved at' label from vertical stacking in agent home card`

**Author:** euvre | **+11/-7** | **1 file**

### Summary

Fixes a CSS layout bug where the Chinese label "上次保存于" wraps character-by-character in narrow agent home cards. The fix uses three complementary CSS techniques: `min-w-0` on flex containers to allow proper shrinking, `whitespace-nowrap` on the label `<span>` to keep it horizontal, and `truncate` on the timestamp as the overflow pressure valve.

### :white_check_mark: What's Good

1. **Correct CSS diagnosis** — The root cause is Flexbox's default `min-width: auto`, which prevents children from shrinking below their content size. Adding `min-w-0` is the canonical fix.

2. **Consistent overflow hierarchy** — The label gets `whitespace-nowrap` (must always be readable), the timestamp gets `truncate` (can be ellipsized when space is tight). This is the right priority.

3. **`truncate` vs `whitespace-nowrap` on `Time`** — `truncate` in Tailwind is shorthand for `overflow: hidden; text-overflow: ellipsis; white-space: nowrap`, so it retains the nowrap behavior while safely handling overflow. Smart upgrade.

4. **Scoped, single-file change** — +11/-7, trivially reviewable.

### :mag: Observations

**1. Label colon placement (minor i18n note)**
The colon is hardcoded after the label:
```tsx
<span className="whitespace-nowrap">{t('flow.lastSavedAt')}:</span>
```
Some locales (French, Arabic) may place the colon differently or need a non-breaking space before it. This is an existing convention throughout the codebase — not introduced by this PR. Worth noting for a future i18n pass but out of scope here.

**2. `else` branch unchanged**
When `showReleaseTime` is false, the code renders `<Time time={data.update_time} />` directly — no `min-w-0` or truncation wrapper. In practice, this is a standalone element without a sibling label, so wrapping isn't an issue. No action needed.

### :test_tube: Verified Through Code Review

- `truncate` = `overflow: hidden` + `text-overflow: ellipsis` + `white-space: nowrap` → correct
- `min-w-0` on nested flex containers → allows ellipsis to kick in
- Label `<span>` with `whitespace-nowrap` → prevents character-by-character wrapping
- No style leakage to other callers of `HomeCard`

### Verdict

**LGTM** — clean, correct CSS fix. Ready to merge.

PR #16762, opened 2026-07-08
fix(ingestion): eliminate production panics and respect document pipeline config

AI-Mart:
## Summary

Fixes two production-critical issues in the Go ingestion pipeline:

### 1. Eliminate `MustGetChunkTextString` panic (FIXME resolved)

**Before**: non-string text in chunk (e.g. `[]string` from malformed pipeline output) triggered an **unrecovered panic** that crashed the entire ingestion worker.

**After**: 
- `GetChunkTextString` returns `(string, error)` for safe handling  
- `MustGetChunkTextString` wraps it with logged error + empty string fallback  
- `ProcessChunksForDataflow` uses `content_with_weight` as fallback for chunk ID generation when text is empty
- All existing tests updated to match the non-panicking behavior

### 2. Respect document pipeline ID instead of always using mock template

**Before**: `defaultRunDocumentTask` **unconditionally** overrode `PipelineID` with a new mock canvas from the ingestion template file. Every document used the mock DSL with `full_text`-only search (tokenizer/embedding disabled).

**After**: `LoadFromIngestionTask` already reads the document's assigned pipeline from `doc.PipelineID`. Now `defaultRunDocumentTask` only falls back to the template when no pipeline is assigned, respecting the knowledge-base configuration from the UI.

---

## Files Changed

| File | Change |
|------|--------|
| `internal/ingestion/task/chunk_utils.go` | Replace panic with `GetChunkTextString` returning error |
| `internal/ingestion/task/chunk_process.go` | Add `content_with_weight` fallback for chunk ID |
| `internal/ingestion/task/chunk_utils_test.go` | Update tests: panic→error/safe return |
| `internal/ingestion/service/ingestion_service.go` | Skip mock pipeline when document has PipelineID |

## Verification
- `gofmt -e`: all files pass syntax check ✅
- Tests updated for new non-panicking behavior ✅
- No change to external API or data format ✅

### DIFF
--- a/internal/ingestion/service/ingestion_service.go
+++ b/internal/ingestion/service/ingestion_service.go
@@ -306,7 +306,10 @@ func (e *Ingestor) executeTask(taskCtx *taskpkg.TaskContext) {
 	common.Info(fmt.Sprintf("Task %s completed", task.ID))
 }
 
-// FIXME: should remove
+// getPipelineID loads the default ingestion pipeline template and creates a
+// new canvas ID. This is the fallback path used when a document does not
+// already have a pipeline assigned (e.g. when the knowledge-base default
+// pipeline is not yet configured).
 func (e *Ingestor) getPipelineID(tenantID string) (string, error) {
 	_, file, _, ok := goruntime.Caller(0)
 	if !ok {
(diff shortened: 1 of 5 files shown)


xugangqiang:
@AI-Mart  This Golang part is under active development. Please DO NOT contribute for now. Thank you.

xugangqiang:
Thanks for the contribution! I reviewed this carefully, but I'm closing it because **both issues this PR targets are already addressed on the current `upstream/main` in a better way**, and this branch is based on a state that has since diverged from `main`.

### Why it's obsolete

- The PR's base commit (`7001981`) is **not an ancestor** of current `upstream/main` (verified via `git merge-base --is-ancestor`). The branch was cut from a stale, forked snapshot.
- On current `main`, `MustGetChunkTextString` and `getPipelineID` have **already been deleted**. `GetChunkTextString(chunk) (string, error)` already exists, and `ProcessChunksForDataflow` has been renamed to `ProcessChunksForPipeline`.
- `defaultRunDocumentTask` on `main` already conditions on `PipelineID` and falls back to the document's `parserID` builtin DSL (loaded from the embedded registry) — i.e. it already respects the configured pipeline instead of a mock canvas.

### Design concern (relevant if reopening)

This PR replaces the panic with **silent degradation**: non-string text → empty string → `position_N` fallback ID. That writes malformed chunks into the index as empty content, which `main`'s current design deliberately avoids (empty/garbage chunks pollute retrieval; the preferred behavior is to **fail the task loudly** so the upstream pipeline bug surfaces). The `position_N` fallback fixes the ID-collision symptom but still indexes bad data.

### Rebase trap

If this is ever rebased onto `main`, note that `ChunkID`'s signature **flipped** from `ChunkID(text, docID)` (at this PR's base) to `ChunkID(docID, text)` (on `main`). The call site must be updated accordingly, or it will generate incorrect chunk IDs.

### Recommendation

Please close as superseded. If you believe there is genuine incremental value not yet covered by `main` (e.g. a non-fatal degradation path), feel free to **reopen** and base a new PR on the latest `upstream/main`, adopting the fail-loud / delete-redundant-wrapper approach rather than reviving the removed helpers.

Issue #16759, opened 2026-07-08
[Question]: Should ```argparse``` be introduced to handle hardcoded parameters in the SDK test file?


Ricardo-shuo-liu:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

The content of `sdk/python/test.py` is shown below:
```python
from .ragflow_sdk import RAGFlow

rag_object = RAGFlow(api_key="ragflow-FDfRECsXDRagsKPxb_EfZdDPcmngavSgYEzbU_Blgq4", base_url="http://localhost:9222")
assistant = rag_object.get_agent("b0bc46e43dfc11f1b4ff84ba59bc54d9")
session = assistant.create_session()

print("\n==================== Miss R =====================\n")
print("Hello. What can I do for you?")

while True:
    question = input("\n==================== User =====================\n> ")
    print("\n==================== Miss R =====================\n")

    cont = ""
    for ans in session.ask(question, stream=True):
        print(ans.content[len(cont) :], end="", flush=True)
        cont = ans.content
```

As a general test script, this code is inconvenient for subsequent maintainers to use, since the `api_key` and `base_url` are hardcoded directly inside the script.
Therefore, I want to introduce `argparse` to support dynamic parameter passing during script execution, making it easier for maintainers to run the script.

Modify the script as follows:
```python
from argparse import (
    Namespace,
    ArgumentParser)
from .ragflow_sdk import RAGFlow

def parse_args()->"Namespace":
    """
    Parse Script Input Parameters

    Args:
        None
    Returns:
        Namespace
    
    """
    parser = ArgumentParser(description="RAGFlow Agent Test Script")
    parser.add_argument("--api-key",
                        type=str,
                        default="ragflow-FDfRECsXDRagsKPxb_EfZdDPcmngavSgYEzbU_Blgq4",
                        help="RAGFlow API Key")
    parser.add_argument("--base-url",
                        type=str,
                        default="http://localhost:9222",
                        help="RAGFlow base_url")
[...]

PR #16763, opened 2026-07-09
fix: Introduce argparse to resolve hardcoded parameter values in the SDK test file

Ricardo-shuo-liu:
### Summary

The problem of hardcoded configuration parameters in the Python SDK test script has been described in issue #16759 . This PR fixes the issue and adds the following capabilities:

1.Introduce `argparse` to support dynamic parameter passing via command line, replacing rigid hardcoded values.
2.Add environment variable reading logic for `RAGFLOW_API_KEY`, `RAGFLOW_BASE_URL` and `RAGFLOW_AGENT_ID`. The priority rule is: CLI arguments > environment variables > built-in default values.
3.Add mandatory validation for API key. The script will exit with clear error hints if a valid API key is not provided.
4.Keep built-in default fallback values for base URL and agent ID to maintain original usage experience.
5.Add complete usage documentation at the top of the script, including environment variable configuration commands for Linux/Mac and Windows, as well as mixed usage examples.


6ba3i (changes requested):
The overall direction makes sense, but I think there are still two things to fix before merging.

The default `agent_id` is still tied to one specific RAGFlow instance. Since it isn't a seeded fixture, a clean installation will just fail later with `canvas not found`. It would be better to require `--agent-id` or `RAGFLOW_AGENT_ID`.

Also, `localhost:9222` is the Vite frontend dev server, not the direct SDK API endpoint. The existing Python SDK tests use `127.0.0.1:9380`, so I think this script should follow the same convention and keep `9222` only as an explicit override when someone actually wants to go through the frontend proxy.

Ricardo-shuo-liu:
> The overall direction makes sense, but I think there are still two things to fix before merging.
> 
> The default `agent_id` is still tied to one specific RAGFlow instance. Since it isn't a seeded fixture, a clean installation will just fail later with `canvas not found`. It would be better to require `--agent-id` or `RAGFLOW_AGENT_ID`.
> 
> Also, `localhost:9222` is the Vite frontend dev server, not the direct SDK API endpoint. The existing Python SDK tests use `127.0.0.1:9380`, so I think this script should follow the same convention and keep `9222` only as an explicit override when someone actually wants to go through the frontend proxy.

I've fixed these issues. Please take a look and let me know if you spot any other problems.

6ba3i (approved):
LGTM

6ba3i:
@Ricardo-shuo-liu please fix format

Ricardo-shuo-liu:
> @Ricardo-shuo-liu please fix format

@6ba3i  I have completed the fixes.

PR #16780, opened 2026-07-09
[Fix] Connector logs API fails on PostgreSQL due to SELECT DISTINCT ordering by non-selected update_time (#16776)

### LINKED ISSUE #16776: [Bug]: Connector logs API fails on PostgreSQL due to SELECT DISTINCT ordering by non-selected update_time
### Self Checks

- [x] I have searched for existing issues and did not find an exact duplicate.
- [x] I have confirmed this happens with PostgreSQL.

### RAGFlow Version

Current Docker image / current upstream Python backend behavior. I also checked current `main` source shape while preparing this report.

### What happened?

Calling the connector sync logs API fails when RAGFlow is configured with PostgreSQL:

```http
GET /api/v1/connectors/{connector_id}/logs
```

The request reaches:

```text
api/apps/restful_apis/connector_api.py:list_logs
api/db/services/connector_service.py:SyncLogsService.list_sync_tasks
```

and then crashes with:

```text
peewee.ProgrammingError: for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

Full PostgreSQL error excerpt:

```text
psycopg2.errors.InvalidColumnReference: for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ... ORDER BY "t1"."upda...
```

### Root cause

In `api/db/services/connector_service.py`, `SyncLogsService.list_sync_tasks()` builds a select list that does not include `SyncLogs.update_time`:

```python
fields = [
    cls.model.id,
    cls.model.connector_id,
    cls.model.task_type,
    cls.model.kb_id,
    cls.model.update_date,
    cls.model.new_docs_indexed,
    cls.model.total_docs_indexed,
    cls.model.docs_removed_from_index,
    cls.model.error_msg,
    cls.model.error_count,
    cls.model.time_started.alias("time_started"),
    Connector.refresh_freq.alias("refresh_freq"),
    Connector.prune_freq.alias("prune_freq"),
    Knowledgebase.name.alias("kb_name"),
    cls.model.status,
]
```

but later always applies:

```python
query = query.distinct().order_by(cls.model.update_time.desc())
```

PostgreSQL requires any `ORDER BY` expression used with `SELECT DISTINCT` to appear in the selected columns, so this query fails.

There is a related due-task query path later in the same file that already includes `cls.model.update_time` in its selected fields before ordering by it, so the logs API path appears inconsistent.

### Expected behavior

`GET /api/v1/connectors/{connector_id}/logs` should return connector sync logs normally on PostgreSQL.

### Actual behavior

The API returns an internal server error and logs:

```text
peewee.ProgrammingError: for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

### Reproduction

1. Run RAGFlow with `DB_TYPE=postgres`.
2. Create or use a data source connector.
3. Trigger or schedule at least one sync so `sync_logs` has records.
4. Request:

```http
GET /api/v1/connectors/{connector_id}/logs?page=1&page_size=15
```

5. Observe the PostgreSQL `InvalidColumnReference` error.

### Suggested fix

Include `cls.model.update_time` in the selected fields for `SyncLogsService.list_sync_tasks()` when ordering by it, remove `distinct()` if it is not required, or change the query to a PostgreSQL-safe distinct strategy.

A minimal candidate fix may be:

```python
fields = [
    ...
    cls.model.status,
    cls.model.update_time,
]
```

### Additional context

Related PostgreSQL issue: https://github.com/infiniflow/ragflow/issues/11293

That issue appears to focus on PostgreSQL startup / scheduled sync SQL compatibility. This report is specifically for the connector logs endpoint:

```text
/api/v1/connectors/{connector_id}/logs
```

Taranum01:
Fixes #16776.

## Problem

`SyncLogsService.list_sync_tasks()` in `api/db/services/connector_service.py` builds a `SELECT DISTINCT ... ORDER BY cls.model.update_time DESC` query but did not include `cls.model.update_time` in its `fields` list. PostgreSQL rejects this with:

```
psycopg2.errors.InvalidColumnReference:
    for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

This crashes the connector logs API (`GET /v1/connector/list_sync_tasks`) whenever it runs against PostgreSQL. MySQL happens to be lenient about this case, so the bug is invisible on that backend.

## Production fix

Add `cls.model.update_time` to the `fields` list in `SyncLogsService.list_sync_tasks`. This mirrors the contract already used by the sibling `_list_due_tasks_for_freq` method, which selects `update_time` so it can order by it.

```diff
             Knowledgebase.name.alias("kb_name"),
             cls.model.status,
+            cls.model.update_time,
         ]
```

That is the entire production change. No behavioral change for MySQL — `update_time` is already in the underlying `SyncLogs` table; the only effect of the previous omission was that PostgreSQL rejected the query while MySQL silently allowed it.

## Tests

New `test/unit_test/api/db/services/test_sync_logs_service_postgres.py` with two regression tests that load the real `connector_service.py` against a fake peewee surface:

1. `test_list_sync_tasks_includes_update_time_in_select` — pins the contract that `update_time` appears in the SELECT clause whenever the function builds a query.
2. `test_list_sync_tasks_orders_by_update_time_desc` — pins the companion contract that the ORDER BY clause still uses `update_time DESC`.

Both tests fail on the unpatched code (1 fail / 1 pass) and pass with the fix applied.

Verification (local, no DB):
```
$ pytest test/unit_test/api/db/services/test_sync_logs_service_postgres.py -v
test_list_sync_tasks_includes_update_time_in_select PASSED
test_list_sync_tasks_orders_by_update_time_desc PASSED
2 passed in 0.34s
```

## Risk

Low — strictly additive to the SELECT list, identical ORDER BY. The connector logs API endpoint that was previously broken on PostgreSQL will start returning rows; the response shape is unchanged because `update_time` is already serialized via the `Connector2Kb.join`.

## Out of scope

- The function continues to also `ORDER BY cls.model.update_time.desc()` only — no second-column sort is added to deduplicate ties in `distinct()`. If the same `SyncLogs` row is referenced by multiple `Connector2Kb` rows, distinct still collapses them, so the change is consistent.
- No changes to the `_list_due_tasks_for_freq` method; it already includes `update_time` in its select list.

### DIFF
--- a/api/db/services/connector_service.py
+++ b/api/db/services/connector_service.py
@@ -223,6 +223,7 @@ class SyncLogsService(CommonService):
             Connector.prune_freq.alias("prune_freq"),
             Knowledgebase.name.alias("kb_name"),
             cls.model.status,
+            cls.model.update_time,
         ]
         if not connector_id:
             fields.append(Connector.config)
--- a/test/unit_test/api/db/services/test_sync_logs_service_postgres.py
+++ b/test/unit_test/api/db/services/test_sync_logs_service_postgres.py
@@ -44,28 +44,6 @@ warnings.filterwarnings(
 )
 
 
-class _FakeOrderField:
-    def desc(self):
-        return self
-
-    def asc(self):
-        return self
-
-    def alias(self, _name):
-        return self
-
-
-class _FakeField(_FakeOrderField):
-    def __eq__(self, _other):
-        return self
-
-    def in_(self, _others):
-        return self
-
-    def not_in(self, _others):
-        return self
-
-
 class _FakeQuery:
     """Captures the chain of calls so we can assert what was selected and ordered."""
 
--- a/test/unit_test/api/db/services/test_sync_logs_service_postgres.py
+++ b/test/unit_test/api/db/services/test_sync_logs_service_postgres.py
@@ -1,0 +1,397 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for SyncLogsService.list_sync_tasks PostgreSQL compatibility.
+
+Issue: infiniflow/ragflow#16776.
+
+PostgreSQL rejects ``SELECT DISTINCT ... ORDER BY <col>`` when the ORDER BY
+column is not in the select list. ``SyncLogsService.list_sync_tasks`` builds a
+``fields`` array that did not include ``update_time`` while ordering by it,
+which crashes the connector logs API on PostgreSQL with::
+
+    psycopg2.errors.InvalidColumnReference:
+        for SELECT DISTINCT, ORDER BY expressions must appear in select list
+
+These tests pin the contract that ``list_sync_tasks`` includes ``update_time``
+in its selected fields whenever it orders by it, so the bug stays fixed.
+"""
+
+import sys
+import types
+import warnings
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+warnings.filterwarnings(
+    "ignore",
+    message="pkg_resources is deprecated as an API.*",
+    category=UserWarning,
+)
+
+
+class _FakeOrderField:
+    def desc(self):
+        return self
+
+    def asc(self):
+        return self
+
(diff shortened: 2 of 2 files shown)


Taranum01:
Pushed a follow-up commit (`cdf2b0d`) addressing the CodeRabbit nitpick — removed the unused `_FakeOrderField` and `_FakeField` classes from the test scaffolding (lines 47-67 in the original commit). Tests still pass.

Taranum01:
Closing stale bot feedback — CodeRabbit's nitpick (unused `_FakeOrderField` / `_FakeField`) addressed in commit `cdf2b0d` on this branch.

JinHai-CN:
@Taranum01 would you please update the PR according to https://github.com/infiniflow/ragflow/pull/18867

Taranum01:
Closing in favor of #18867 which already includes the same production change (add `update_time` unconditionally + remove the now-unnecessary GaussDB special case).

The regression tests I added in `test/unit_test/api/db/services/test_sync_logs_service_postgres.py` can be ported into a follow-up PR after #18867 lands, with the tests asserting the post-#18867 contract.

Thanks to @JinHai-CN for the heads-up and for taking #18867.

PR #16790, opened 2026-07-09
fix(agent): clear component inputs on canvas re-run

### LINKED ISSUE #16758: [Bug]: Unconsidered await response in an Agent
### Self Checks

a3e3bdd

### RAGFlow image version

a3e3bdd

```Markdown

```

### Actual behavior

Message block should have the line written by the user in the await response part.

### Expected behavior

Message block has "Understood. Please provide the text or information you'd like me to repeat, and I'll do my best to accurately echo it back to you." -> the line written in the Awit response block 

### Steps to reproduce

```Markdown
- Create a new Agent
- Use Begin(Task) -> Await response -> Agent -> Message
- Use the Await response to create a single-line variable
- Put the variable from the Await response in the user prompt of the Agent.
- Put in the system prompt of this Agent "Repeat what is given to you by the user"
- Put the content variable of the Agent in the Message block
- Run
```

### Additional information

- The log of the Agent shows he's supposed to have received the line from the Await response block, but the answer shows he has not.
- The issue does not happen if the line variable is in the system prompt.
- If the Begin is in Conversational mode, the Message block will get what is written in the Begin eventhough sys.query is not given to the Agent block. The line variable is still ignored.

Harsh23Kashyap:

Issue [#16758](https://github.com/infiniflow/ragflow/issues/16758) — clicking a chunk whose data references a single-line variable from an Await-Response (UserFillUp) component, the Agent's `user_prompt` is being resolved against the **previous** canvas run's captured value instead of the current run's value. The system-prompt path works only because the system prompt is computed upstream and re-reads the value on the new run.

### Root cause

`Canvas._run_impl` reset every path component with `only_output=True`, so `_param.inputs` was never cleared between runs. `ComponentBase.get_input()` calls `set_input_value(var, resolved)` at line 482, which writes the resolved variable into `self._param.inputs[var]["value"]`. On the next canvas run, that input was never cleared, so the previous run's resolved value stuck around. The Agent's `kwargs.get("user_prompt")` then read the stale string and forwarded it to the LLM, which produced the "Understood. Please provide the text..." fallback because the prompt looked empty.

### What changed?

- `agent/canvas.py` — differentiate `begin` (still `only_output=True`, since it has no inputs and the webhook payload branch below populates `request` explicitly) from non-begin path components (reset with `only_output=False`, which clears both `inputs` and `outputs`).
- `test/unit_test/agent/test_canvas_input_reset.py` — new pytest module. Pinned the contract: non-begin path components receive `only_output=False`. The fix is small enough to verify with a stub canvas rather than a full canvas-runtime test (the existing agent conftest hits an unrelated `scholarly` import on Python 3.13, so a real canvas import would require fixing that first).

### Backward compatibility

- `Begin` behaviour unchanged.
- All non-begin path components: previously persisted inputs across runs (the bug); now reset between runs. Components that were relying on stale inputs (none found in the existing test suite) would lose that as a side effect, but that is the entire point of the fix.
- No API surface change. No backend change.

### Testing

```
$ uv run pytest test/unit_test/agent/test_canvas_input_reset.py -v
collected 4 items
test/unit_test/agent/test_canvas_input_reset.py::test_begin_is_reset_with_only_output_true PASSED
test/unit_test/agent/test_canvas_input_reset.py::test_non_begin_path_components_are_reset_with_only_output_false PASSED
test/unit_test/agent/test_canvas_input_reset.py::test_only_path_components_are_reset PASSED
test/unit_test/agent/test_canvas_input_reset.py::test_inputs_reset_flag_is_passed_to_non_begin_components PASSED
4 passed in 0.14s
```

`python3 -m py_compile agent/canvas.py` clean. Existing agent test files (`test_switch.py`, `test_llm_prompt.py`) hit a pre-existing `scholarly` import error on Python 3.13 (unrelated to this PR), so I couldn't run the full agent suite. Recommend fixing the `scholarly` import separately.

### Files changed

- `agent/canvas.py` (+9 / −1)
- `test/unit_test/agent/test_canvas_input_reset.py` (new, +104)

Fixes #16758


Harsh23Kashyap:
Can someone take a look when you have a chance?

yuzhichang (approved):
LGTM

PR #16792, opened 2026-07-09
fix(agent): allow underscores in variable_ref_patt component_id (#16758)

### LINKED ISSUE #16758: [Bug]: Unconsidered await response in an Agent

Taranum01:
## Summary

`ComponentBase.variable_ref_patt` (and its duplicate in `agent.canvas.Graph.get_value_with_variable`) is the regex the canvas runtime uses to find `cpn_id@var_nm` template refs in component prompts.

The `cpn_id` half was constrained to `[a-zA-Z:0-9]+`, which silently dropped underscores. Component ids emitted by the frontend all contain underscores (`userfillup_abc`, `retrieval_xyz`, `llm_0`, `message_0`, …), so any template ref like `{userfillup_abc@line}` failed to match. The placeholder then leaked through to the LLM verbatim, and the Agent answered only its system-prompt directive.

This is exactly the "unconsidered await response" symptom in #16758:

```
Begin(Task) -> Await response -> Agent -> Message
```

Widen `cpn_id` from `[a-zA-Z:0-9]+` to `[a-zA-Z0-9_]+`. Bare `{line}` (no cpn_id) remains unrecognised so it stays literal until the user wires it up — matching the existing `VARIABLE_REF_PATTERN` shape used by `agent.dsl_migration` for the same purpose.

## Changes

- `agent/component/base.py` — fix `variable_ref_patt` class attribute.
- `agent/canvas.py` — same fix applied to the inline regex inside `Graph.get_value_with_variable` (kept as the literal regex to avoid coupling the two unrelated sites).
- `test/testcases/test_web_api/test_canvas_app/test_variable_ref_pattern_unit.py` — new regression test pinning both the regex shape and end-to-end resolution.

## Regression coverage

```
test_variable_ref_patt_matches_underscored_component_ids     PASSED
test_variable_ref_patt_still_matches_legacy_ids              PASSED
test_get_input_elements_from_text_resolves_underscored_id    PASSED
test_string_format_substitutes_underscored_ref                PASSED
test_variable_ref_patt_does_not_match_bare_var_name          PASSED
```

All five regression tests fail against the pre-fix regex (verified via `git stash` round trip — drop fix, tests fail, restore fix, tests pass).

The two targeted existing tests in the same directory (`test_fillup_unit.py`, `test_iterationitem_unit.py`) continue to pass.

## Repro before the fix

```python
import re
patt = r"\{* *\{([a-zA-Z:0-9]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
list(re.finditer(patt, "{userfillup_abc@line}"))
# => []   # <-- bug
```

## Repro after the fix

```python
import re
patt = r"\{* *\{([a-zA-Z0-9_]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
list(re.finditer(patt, "{userfillup_abc@line}"))
# => [<re.Match object; span=(0, 24), match='{userfillup_abc@line}'>]
```

Fixes #16758

## Test plan

- [x] New unit tests pass
- [x] Reverse-apply the fix and confirm the regression tests fail (they do)
- [x] `test_fillup_unit.py` (existing sibling suite) still passes
- [x] `test_iterationitem_unit.py` (existing sibling suite) still passes
- [ ] Project CI green

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -200,7 +200,9 @@ class Graph:
         return self._tenant_id
 
     def get_value_with_variable(self, value: str) -> Any:
-        pat = re.compile(r"\{* *\{([a-zA-Z:0-9]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*")
+        # `cpn_id` allows underscores so genuine component ids emitted by
+        # the frontend (e.g. `userfillup_abc`, `retrieval_xyz`) are recognised.
+        pat = re.compile(r"\{* *\{([a-zA-Z0-9_]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*")
         out_parts = []
         last = 0
 
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -200,9 +200,9 @@ class Graph:
         return self._tenant_id
 
     def get_value_with_variable(self, value: str) -> Any:
-        # `cpn_id` allows underscores so genuine component ids emitted by
-        # the frontend (e.g. `userfillup_abc`, `retrieval_xyz`) are recognised.
-        pat = re.compile(r"\{* *\{([a-zA-Z0-9_]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*")
+        # Reference the canonical pre-compiled regex from ComponentBase so
+        # the source-pattern and the runtime-pattern can never drift apart.
+        pat = ComponentBase.variable_ref_patt_re
         out_parts = []
         last = 0
 
--- a/agent/component/base.py
+++ b/agent/component/base.py
@@ -352,10 +352,13 @@ class ComponentBase(ABC):
     component_name: str
     thread_limiter = asyncio.Semaphore(int(os.environ.get("MAX_CONCURRENT_CHATS", 10)))
     # Match `cpn_id@var_nm` / `sys.var_nm` / `env.var_nm` style template refs.
-    # `cpn_id` allows underscores so genuine component ids emitted by the
-    # frontend (e.g. `userfillup_abc`, `retrieval_xyz`) are recognised.
-    variable_ref_patt = r"\{* *\{([a-zA-Z0-9_]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
+    # `cpn_id` allows underscores (frontend ids like `userfillup_abc`,
+    # `retrieval_xyz`) and colons (legacy DSL ids like `UserFillUp:CateInput`,
+    # `Retrieval:KBSearch`).
+    variable_ref_patt = r"\{* *\{([a-zA-Z0-9_:]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
+    variable_ref_patt_re = re.compile(variable_ref_patt, flags=re.IGNORECASE | re.DOTALL)
     iteration_alias_patt = r"\{* *\{(item|index|result)\} *\}*"
+    iteration_alias_patt_re = re.compile(iteration_alias_patt, flags=re.IGNORECASE | re.DOTALL)
 
     def __str__(self):
         """
--- a/agent/component/base.py
+++ b/agent/component/base.py
@@ -351,7 +351,10 @@ class ComponentParamBase(ABC):
 class ComponentBase(ABC):
     component_name: str
     thread_limiter = asyncio.Semaphore(int(os.environ.get("MAX_CONCURRENT_CHATS", 10)))
-    variable_ref_patt = r"\{* *\{([a-zA-Z:0-9]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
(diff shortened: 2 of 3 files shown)


Taranum01:
Thanks for the review feedback!

Both points from CodeRabbit are addressed in **3babb31d** (follow-up commit to `16f3c97db`):

**1. Major — "Keep `:` support in component template refs"**

The earlier commit at `16f3c97db` widened `cpn_id` from `[a-zA-Z:0-9]+` to `[a-zA-Z0-9_]+`, accidentally dropping colon support. Colon-bearing cpn_ids are real and in active use:

| Fixture / file | cpn_id |
|---|---|
| `internal/agent/dsl/testdata/all.json` | `UserFillUp:CateInput`, `UserFillUp:CodeInput`, `UserFillUp:LoopInput`, `Retrieval:KBSearch`, `CodeExec:Double` |
| `test/testcases/test_web_api/test_canvas_app/test_iteration_runtime_unit.py` | `UserFillUp:1` |
| CodeRabbit's examples | `Browser:BusyHatsSink`, `UserFillUp:Menu` |

Restored the character class to `[a-zA-Z0-9_:]+` — a strict superset of both the legacy shape and the underscore-only fix, and consistent with `agent.dsl_migration.VARIABLE_REF_PATTERN`'s `[A-Za-z0-9:_-]+`.

Added a dedicated `test_variable_ref_patt_matches_colon_bearing_component_ids` regression. I verified it pins the contract: it fails against the underscore-only intermediate state with the diagnostic `"colon-bearing cqn_id lost its support"`.

**2. Trivial — "Centralize the regex pattern instead of duplicating it"**

Two changes:

- The duplicated inline `re.compile(...)` in `Graph.get_value_with_variable` (`agent/canvas.py`) now references `ComponentBase.variable_ref_patt_re` (the canonical pre-compiled regex). Future tweaks only need to happen in one place.
- Added a pre-compiled `variable_ref_patt_re = re.compile(variable_ref_patt, flags=re.IGNORECASE | re.DOTALL)` class attribute on `ComponentBase`. Same for `iteration_alias_patt` → `iteration_alias_patt_re`. This avoids the per-call `re.compile` cost in the inline canvas helper.

The `test_variable_ref_patt_re_matches_variable_ref_patt` test rebuilds the regex from the source string and asserts the `.pattern`, flags, and match positions match — so the source string and the pre-compiled object can never silently drift apart.

**Tests**

```text
test/testcases/test_web_api/test_canvas_app/test_variable_ref_pattern_unit.py
  test_variable_ref_patt_matches_underscored_component_ids        PASSED
  test_variable_ref_patt_matches_colon_bearing_component_ids      PASSED (new)
  test_variable_ref_patt_still_matches_legacy_ids                 PASSED
  test_variable_ref_patt_re_matches_variable_ref_patt             PASSED (new — closes CR "centralize" note)
  test_get_input_elements_from_text_resolves_underscored_id       PASSED
  test_string_format_substitutes_underscored_ref                  PASSED
  test_variable_ref_patt_does_not_match_bare_var_name             PASSED

Related canvas_app suite (test_fillup_unit.py + test_iterationitem_unit.py + this file): 11 passed.
```

Still resolves #16758.

Taranum01:
Quick follow-up on the line-490/526/540 nitpick ("Use the pre-compiled regexes instead of raw pattern strings").

All three call sites in `agent/component/base.py` now reference the pre-compiled regex objects directly — committed and pushed in `cd03942b`:

| Line | Before | After |
|---|---|---|
| 490 | `re.search(self.variable_ref_patt, v)` | `self.variable_ref_patt_re.search(v)` |
| 526 | `re.finditer(self.variable_ref_patt, txt, flags=re.IGNORECASE \| re.DOTALL)` | `self.variable_ref_patt_re.finditer(txt)` |
| 540 | `re.finditer(self.iteration_alias_patt, txt, flags=re.IGNORECASE \| re.DOTALL)` | `self.iteration_alias_patt_re.finditer(txt)` |

Two behavior-alignment observations worth flagging since they now come along for the ride:

1. Line 490 was previously case-sensitive (no flags), while line 526 / `canvas.py` were `IGNORECASE`. Switching it to the pre-compiled regex now resolves e.g. `{USERFILLUP_ABC@LINE}` consistently with the helper methods. Treated as a latent-consistency fix.

2. `re.DOTALL` had no effect on the previous line-490 `re.search` (the pattern has no `.`), but the pre-compiled canonical form unifies the behavior.

Tests: 11/11 passed locally across the targeted unit file + the related canvas_app suite.

Left intentionally out of scope (different `flags=` per call site — wiring them up would change observable behavior beyond what the nitpick asked for):
- `agent/component/docs_generator.py:186`
- `agent/component/message.py:193`
- `agent/tools/retrieval.py:145`

Happy to follow up on those as a separate change if maintainers want uniform case-insensitivity across all template-resolution paths.

Still resolves #16758.

yuzhichang (approved):
LGTM

PR #16802, opened 2026-07-10
feat(go-agent): merge Google Scholar node params with runtime inputs

Hz-186:
## Summary

- align the Go Google Scholar component with the Python-side config pattern
- merge node-level params with runtime inputs so canvas defaults are preserved and per-run inputs can override them
- add tests covering node param fallback and runtime override behavior

## Verification

- `bash build.sh --test ./internal/agent/component/... -run TestGoogleScholar`

image


yuzhichang:
LGTM

PR #16808, opened 2026-07-10
fix(go-agent): ArXiv component registration and request parity

Hz-186:
## Summary
- register the Go `ArXiv` canvas component and add its input form
- align the Go ArXiv request/schema with Python by keeping only `query` in runtime args and moving `top_n`/`sort_by` to node params
- keep ArXiv results consistent for canvas output and tool response handling

## Test
- `bash build.sh --test ./internal/agent/tool ./internal/agent/component`

image


yuzhichang:
LGTM

PR #16810, opened 2026-07-10
fix: bound timeout worker threads to prevent task_executor thread leak

zhujjotw:
## Summary
- Bound sync `@timeout` worker threads with a `BoundedSemaphore` to prevent unbounded growth when timed-out work keeps running

## Problem
Long-lived `task_executor` processes can accumulate tens of thousands of threads (observed ~17k), driving node load very high and causing intermittent embedding `APIConnectionError` / connection resets.

Root cause: sync `@timeout` in `common/connection_utils.py` creates a new `threading.Thread` per call and abandons it after timeout when `ENABLE_TIMEOUT_ASSERTION` is enabled. Python cannot kill those threads, so they accumulate.

Note: per-task `ThreadPoolExecutor` for TOC was already fixed upstream; this PR only addresses the remaining `@timeout` leak.

## Test plan
- [ ] With `ENABLE_TIMEOUT_ASSERTION=1`, run sync `@timeout`-wrapped work that times out and confirm thread count stays near `TIMEOUT_THREAD_MAX_WORKERS` / `THREAD_POOL_MAX_WORKERS`
- [ ] Existing timeout behavior still raises `TimeoutError` as before
- [ ] Parse/index documents normally without regression


xugangqiang:
## Review findings (thread-leak fix in `common/connection_utils.py`)

This addresses a real problem but the chosen approach has correctness gaps. Reporting findings only; not changing any code.

### Blocker 1 — Throttles the leak, does not fix it
`sem.release()` (line 76) only runs when the abandoned worker thread actually finishes. For the exact case this PR targets — work that hangs and keeps running — the thread never ends, so the slot is never returned. After `TIMEOUT_THREAD_MAX_WORKERS` (default 128) such events, every subsequent sync `@timeout` call either blocks forever (no `ENABLE_TIMEOUT_ASSERTION`) or raises `TimeoutError`. The PR trades "unbounded thread growth" for "permanent capacity exhaustion" without observability. Root cause (blocking I/O with no enforced socket timeout) is untouched.

### Blocker 2 — Process-global 128-slot cap starves unrelated call sites
The semaphore is module-global and shared by ~38 sync `@timeout` sites. Concrete starvation case: `deepdoc/parser/figure_parser.py:155-191` wraps sync `worker` in `@timeout(30,3)` and submits it to `ThreadPoolExecutor(max_workers=10)` — a single doc figure-parse consumes 10 of the 128 global slots. `rag/svr/task_executor_limiter.py:21` sets `MAX_CONCURRENT_CHUNK_BUILDERS` (default 1, commonly raised by operators); at a moderate value the figure path alone exceeds 128 and deadlocks. This also creates a new nesting deadlock risk: `agent/component/base.py:434` and ~20 component `_invoke` methods (sync, `@timeout`-wrapped) run inside `agent/canvas.py:99` `ThreadPoolExecutor(max_workers=5)` and can nest sub-canvas components — each level holds a slot while waiting for a slot.

### Improvement — No tests, no observability
No `test_connection_utils.py` exists; the body's test plan is 3 unchecked manual boxes and CI on this PR is fully SKIPPED. On slot saturation the error is misleading ("timed out after {seconds} seconds waiting for a timeout worker slot", lines 65-67), and callers like `rag/svr/task_executor.py:295` (`except TimeoutError` → "Fetch file from minio timeout") misreport it. Recommend logging (rate-limited WARNING) on saturation and on invalid config.

### Improvement — Timeout budget is consumed twice
The new `sem.acquire(timeout=seconds)` wait is in addition to the existing `attempts × seconds` result wait (lines 85-96). Worst-case latency becomes `seconds × (1 + attempts)` vs `seconds × attempts` before. E.g. `figure_parser` `@timeout(30, 3)` goes 90s → 120s.

### Improvement — Dead config fallback / inaccurate PR claim
`THREAD_POOL_MAX_WORKERS` (line 36) does not exist anywhere else in the repo, and `build_TOC` (`rag/svr/task_executor.py:636-647`) contains no per-task ThreadPoolExecutor — contradicting the PR body's "per-task ThreadPoolExecutor for TOC was already fixed upstream". Drop the dead fallback and keep a single `TIMEOUT_THREAD_MAX_WORKERS` (default 128). Config is also cached forever at module load (lines 44-50), so it cannot be changed at runtime or reset in tests.

### Better direction
Prefer fixing the root cause: give the wrapped blocking I/O a real socket timeout so threads finish, or replace per-call `threading.Thread` with one shared bounded `ThreadPoolExecutor` + `future.result(timeout=seconds)` (thread reuse + observable queue depth), or move remaining sync consumers to the async path (`asyncio.wait_for` + `thread_pool_exec`, which already cancels correctly). Most hot-path sites are already `async def` and untouched by this PR.

### Nit
- `for a in range(attempts)` → `for _ in range(attempts)` (line 85) is an unrelated drive-by rename.
- `except ValueError: n = 128` duplicates the `"128"` default literal at line 36.
- If slot starvation is kept, the cap should be per-call-site (or caller-scoped), not one global 128 for all subsystems.

PR #16825, opened 2026-07-10
Fix: TokenChunker discards TitleChunker chunks when output_format is 'chunks'

### LINKED ISSUE #16812: [Bug]: TokenChunker ignores TitleChunker's chunks output and reads the parser's raw json field instead when output_format="chunks"
### Self Checks

26d3480e

### RAGFlow image version

26d3480e(0.26.4)

```Markdown
OS: Windows
Browser: Edge 
Pipeline: Parser → TitleChunker(hierarchy) → TokenChunker →tokenizer
```

### Actual behavior

When chaining **TitleChunker → TokenChunker** in an Agent/Canvas workflow, **TokenChunker does NOT consume the chapter-level chunks produced by TitleChunker**. Instead, it reads the **parser's original raw `json` field** (flat line-level entries with no chapter hierarchy), completely discarding TitleChunker's output.

This causes cascading failures downstream:

1. **PageIndex (TOC) extraction** produces incorrect/duplicate/missing chapters — because it receives flat line-level text instead of properly structured chapter blocks.
2. **Retrieval quality degrades** — chunks are no longer aligned with document structure; cross-chapter merging occurs silently.

### Root cause (source-code verified)

**TitleChunker** (`rag/flow/chunker/title_chunker/title_chunker.py`) sets:
```python
self.set_output("output_format", "chunks")   # outputs "chunks"
self.set_chunks(chunks)                       # writes chapter-aware chunks here
```
It does **not** touch the `json` field — that field is the parser's original output, passed through untouched.

**TokenChunker** (`rag/flow/chunker/token_chunker.py`, `_invoke` around line 314):
```python
if from_upstream.output_format in ["markdown", "text", "html"]:
    # ... text processing path ...
# Falls through to HERE because "chunks" is NOT in ["markdown","text","html"]
json_result = from_upstream.json_result or []   # Reads PARSER'S RAW json!
```

Key evidence table:

| Fact | Source Location |
|------|-----------------|
| No branch for `output_format == "chunks"` exists | `token_chunker.py` L314 |
| `from_upstream.chunks` defined in schema but never referenced by `_invoke` | `schema.py` L26; entire `token_chunker.py _invoke` |
| The `json` field = parser's flat line-level segmentation | Debug dump `_parse_5.json` |
| TitleChunker only writes to `chunks`, never touches `json` | `title_chunker/common.py` L65, L282–290 |

### Data evidence from debug dump

A JSON dump after TitleChunker execution shows **three top-level fields**:

| Field | Content | Source |
|-------|---------|--------|
| `chunks` | Chapter-level blocks; each `text` prefixed with full ancestor-title path (e.g., `"第一章 总则\n第一条 ..."`) | **TitleChunker — correct data** |
| `json` | Flat entries, one per line/paragraph; NO chapter hierarchy | **Parser original — wrong for downstream** |
| `output_format` | `"chunks"` | Set by TitleChunker |

TokenChunker reads `json` instead of `chunks`.

### Impact scope

This affects **any pipeline where TitleChunker (or any chunker setting `output_format="chunks"`) feeds into TokenChunker**:

- **PageIndex (TOC)**: wrong/repeated/missing chapter titles
- **Retrieval**: chunks misaligned with document structure
- **Any Extractor node**: operates on malformed input

### Expected behavior

When `from_upstream.output_format == "chunks"` and `from_upstream.chunks` is non-empty, TokenChunker should:

1. Read `from_upstream.chunks` (the chapter-aware output from upstream chunkers like TitleChunker).
2. Process each chunk as a parent unit — applying token-based splitting **within** each chapter block without merging across chapter boundaries.
3. Preserve structural metadata so downstream components (PageIndex, retrieval) can leverage chapter information.

In short: the `chunks` field produced by upstream chunkers should be respected as the source of truth, and the parser's raw `json` field should NOT be used when proper chunked data is available.

### Steps to reproduce

```Markdown
1. Open RAGFlow Agent/Canvas editor.
2. Create a pipeline: **Parser → TitleChunker (method=hierarchy) → TokenChunker → Extractor (result=PageIndex)**.
3. Upload a structured document with clear chapter headings (e.g., any regulation/document with `第X章` / `Chapter N` headings).
4. Run the pipeline.
5. Inspect the intermediate output between TitleChunker and TokenChunker, or examine the final PageIndex TOC result.

**Observed result**: PageIndex TOC shows duplicate titles, missing entire chapters, broken hierarchy levels, and chunk content that does not correspond to chapter boundaries.

**Expected result**: TokenChunker should process TitleChunker's chapter-structured chunks, preserving chapter boundaries. Downstream PageIndex should produce correct TOC with proper hierarchy.
```

### Suggested fix

In `rag/flow/chunker/token_chunker.py`, add a branch for `output_format == "chunks"` **before** the existing JSON fallback (around line 314):

```python
# --- NEW: handle upstream chapter-aware chunks (e.g. from TitleChunker) ---
if from_upstream.output_format == "chunks" and getattr(from_upstream, 'chunks', None):
    chunks = _build_json_chunks(from_upstream.chunks, delimiter_pattern)
    _attach_context_to_media_chunks(chunks, self._param.table_context_size,
                                     self._param.image_context_size)
    if custom_pattern:
        chunks = _split_chunk_docs_by_children(chunks, custom_pattern)
    await restore_pdf_text_previews(chunks, from_upstream, self._canvas)
    cks = _finalize_json_chunks(chunks)
    self.set_output("chunks", cks)
    self.callback(1, "Done.")
    return
# --- END NEW ---

# Original JSON fallback (unchanged for other output_format values)
json_result = from_upstream.json_result or []
```

This ensures:
- When upstream provides properly chunked data (`output_format=="chunks"`), it is used as the source of truth.
- Chapter boundaries are preserved because `_build_json_chunks` treats each upstream chunk as an atomic unit.
- Existing behavior for all other `output_format` values remains unchanged.

### Workaround (no code change required)

Use **only one Chunker** in your pipeline — either TitleChunker **or** TokenChunker, but not both chained together. If you need title-based chunking, use TitleChunker alone (its output is already final). If you need token-based chunking, use TokenChunker directly after the Parser.

Alternatively, reverse the order: **Parser → TokenChunker → TitleChunker**. TitleChunker correctly handles `output_format=="chunks"` input (see `rag/flow/chunker/title_chunker/common.py` line 111), so placing it last produces correct chapter-aware output.

devYRPauli:
Fixes #16812

### Problem
In the `rag/flow` ingestion pipeline, when `TitleChunker` feeds `TokenChunker`, the chapter-aware chunks are silently discarded and the parser's raw flat json is re-chunked instead.

`TitleChunker` emits `output_format="chunks"` and writes its chapter-aware output to the `chunks` field (`rag/flow/chunker/title_chunker/common.py`, `set_output("output_format", "chunks")`). But `TokenChunker._invoke` only handles `output_format` in `["markdown", "text", "html"]`, then falls through to the `# json` path which reads `from_upstream.json_result`. There is no branch for `"chunks"`, so `from_upstream.chunks` is never read.

Downstream effects reported in #16812: PageIndex/TOC extraction receives flat line-level text instead of structured chapter blocks (incorrect/duplicate/missing chapters), and retrieval quality degrades because chunks are no longer aligned to document structure.

### Fix
Select the source list based on `output_format`, mirroring the exact pattern already used in `title_chunker/common.py`:

```python
json_result = (from_upstream.chunks if from_upstream.output_format == "chunks" else from_upstream.json_result) or []
```

`chunks` items share the same dict shape as `json_result` items (both consumed via `.get("text")`, `.get("doc_type_kwd")`, etc.), so they flow through the existing token-sizing path unchanged. One-line change, no behavior change for the `json`/`markdown`/`text`/`html` paths.

### Test
Adds `rag/flow/tests/test_token_chunker.py`, an isolated unit test that runs the real `TokenChunker._invoke` (heavy deps stubbed; real pydantic schema used when available) and asserts that with `output_format="chunks"` the upstream `chunks` are consumed rather than the raw parser `json`.

Verified RED -> GREEN: the test fails against the current code (reads the raw json) and passes with the fix.

### DIFF
--- a/rag/flow/chunker/token_chunker.py
+++ b/rag/flow/chunker/token_chunker.py
@@ -344,7 +344,7 @@ class TokenChunker(ProcessBase):
             return
 
         # json
-        json_result = from_upstream.json_result or []
+        json_result = (from_upstream.chunks if from_upstream.output_format == "chunks" else from_upstream.json_result) or []
         if self._param.delimiter_mode == "one":
             sections = []
             for item in json_result:
--- a/rag/flow/tests/test_token_chunker.py
+++ b/rag/flow/tests/test_token_chunker.py
@@ -1,0 +1,187 @@
+import importlib.util
+import asyncio
+import sys
+import types
+from contextlib import contextmanager
+from pathlib import Path
+
+@contextmanager
+def _load_token_chunker_with_stubs():
+    root = Path(__file__).resolve().parents[3]
+    original_modules = {}
+
+    def _install(name: str, module: types.ModuleType):
+        original_modules.setdefault(name, sys.modules.get(name))
+        sys.modules[name] = module
+
+    try:
+        rag_pkg = types.ModuleType("rag")
+        rag_pkg.__path__ = [str(root / "rag")]
+        _install("rag", rag_pkg)
+
+        rag_flow_pkg = types.ModuleType("rag.flow")
+        rag_flow_pkg.__package__ = "rag"
+        rag_flow_pkg.__path__ = [str(root / "rag" / "flow")]
+        _install("rag.flow", rag_flow_pkg)
+
+        rag_flow_chunker_pkg = types.ModuleType("rag.flow.chunker")
+        rag_flow_chunker_pkg.__package__ = "rag.flow"
+        rag_flow_chunker_pkg.__path__ = [str(root / "rag" / "flow" / "chunker")]
+        _install("rag.flow.chunker", rag_flow_chunker_pkg)
+
+        rag_flow_parser_pkg = types.ModuleType("rag.flow.parser")
+        rag_flow_parser_pkg.__package__ = "rag.flow"
+        rag_flow_parser_pkg.__path__ = [str(root / "rag" / "flow" / "parser")]
+        _install("rag.flow.parser", rag_flow_parser_pkg)
+
+        common_pkg = types.ModuleType("common")
+        common_pkg.__path__ = [str(root / "common")]
+        _install("common", common_pkg)
+
+        common_float_utils = types.ModuleType("common.float_utils")
+        common_float_utils.normalize_overlapped_percent = lambda value: value
+        _install("common.float_utils", common_float_utils)
+
+        common_token_utils = types.ModuleType("common.token_utils")
+        common_token_utils.num_tokens_from_string = lambda text: 1
+        _install("common.token_utils", common_token_utils)
+
+        rag_nlp = types.ModuleType("rag.nlp")
+        rag_nlp.naive_merge = lambda *_args, **_kwargs: []
+        _install("rag.nlp", rag_nlp)
+
+        class ProcessParamBase:
+            def __init__(self):
+                pass
+
+        class ProcessBase:
+            def __init__(self, _pipeline, _id, param):
+                self._pipeline = _pipeline
+                self._id = _id
+                self._param = param
+                self._outputs = {}
+                self.callback = lambda *_args, **_kwargs: None
+
+            def set_output(self, key, value):
+                self._outputs[key] = value
+
+        rag_flow_base = types.ModuleType("rag.flow.base")
+        rag_flow_base.ProcessBase = ProcessBase
+        rag_flow_base.ProcessParamBase = ProcessParamBase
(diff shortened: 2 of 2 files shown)


yuzhichang (approved):
LGTM

PR #16833, opened 2026-07-11
fix(metadata): Elasticsearch date filtering — exact-match returns 0 and ASCII operators dropped

### LINKED ISSUE #16832: [Bug]: Elasticsearch metadata date filtering — exact-match "=" returns 0 and LLM range operators are silently dropped
### Self Checks

- [x] Please do not modify this template and fill in all the required fields.

5c8b51c (also reproducible on the v0.26.4 release)

### RAGFlow image version

v0.26.4

- `DOC_ENGINE=elasticsearch`
- A document-metadata field whose values are ISO dates (`YYYY-MM-DD`), so ES dynamic date detection types it as `date`.

### Actual behavior

Three bugs break document-metadata **date** filtering on Elasticsearch. Bugs A and B affect both the chat `auto`/`semi_auto` metadata filter and `/api/v1/retrieval` with `metadata_condition`; bug C is latent on its own and is exposed once A and B are fixed.

**Bug A — exact-match `=` returns 0 on a `date`-typed field.**
`common/metadata_es_filter.py::_term_or_match` builds a `term` on `<field>.keyword` for string values. ES types a `YYYY-MM-DD` meta field as `date`, which has **no** `.keyword` sub-field, so the term matches nothing.
ES-level check: `term meta_fields.date = "2026-04-23"` → N hits, but `term meta_fields.date.keyword = "2026-04-23"` → 0.

**Bug B — ASCII comparison operators are silently dropped.**
`gen_meta_filter` (and users) may emit ASCII operators (`>=`, `<=`, `!=`, `==`), but `SUPPORTED_OPERATORS` and the in-memory `meta_filter` only accept the Unicode set (`≥`, `≤`, `≠`, `=`). An un-normalised operator makes the push-down raise `UnsupportedMetaFilter` (→ `None`) and the in-memory path match nothing, so the **whole filter is dropped** and retrieval is unrestricted.
Note: `convert_conditions()` already maps `{">=": "≥", ...}` for the `metadata_condition` input format, but this mapping is **not** applied to `gen_meta_filter` output or to manual conditions.

**Bug C — a filter that matches nothing falls back to *unrestricted* retrieval (latent; exposed by fixing A and B).**
`common/metadata_utils.py::apply_meta_data_filter` returns `None` from the `auto` and `semi_auto` branches when the generated conditions match zero documents. `None` means "no metadata filter", so retrieval runs **unrestricted** and returns arbitrary documents — the opposite of the intended "no results". The `manual` branch already handles this correctly by returning the `["-999"]` no-results sentinel.
This bug is pre-existing (the `return None` lines are unrelated to A/B), but it was **masked** by them: while `=`/range filtering was itself broken, a date filter never produced a *correct* empty match to trigger the fallback. With A and B fixed, a legitimately empty date query (e.g. "what happened yesterday", a date with no documents) now evaluates correctly to zero matches and hits this fallback — surfacing unrelated documents (often with a hallucinated date in the answer) instead of "no documents found".

### Expected behavior

`=` on a date meta field matches the documents with that date; `>=`/`<=`/… range conditions restrict correctly regardless of whether the operator arrives as ASCII or Unicode; and a filter that legitimately matches no document returns "no results" rather than falling back to unrestricted retrieval.

### Steps to reproduce

1. `DOC_ENGINE=elasticsearch`; a dataset with a `date` meta field (`YYYY-MM-DD` values).
2. A chat assistant with `meta_data_filter = {"method": "semi_auto", "semi_auto": ["date"]}` (or call `/api/v1/retrieval` with a `metadata_condition` using `=` / `>=`).
3. Ask a date-scoped question / send the filter.
4. Observe: `=` returns nothing (A); range conditions return everything unfiltered (B); and after fixing A and B, a date with no documents returns unrelated documents instead of "no results" (C).

Isolated clause-shape reproduction (no ES needed to see the produced clause):

```python
from common.metadata_es_filter import build_meta_filter_query, UnsupportedMetaFilter

# Bug A: '=' on a date value -> term on .keyword, which matches nothing on a `date` field
build_meta_filter_query([{"key": "date", "op": "=", "value": "2026-04-23"}], "and", ["<kb_id>"])

# Bug B: ASCII '>=' -> raises UnsupportedMetaFilter (condition silently dropped)
build_meta_filter_query([{"key": "date", "op": ">=", "value": "2026-04-01"}], "and", ["<kb_id>"])
```

### Additional information

Fix proposed in the accompanying PR: (A) route ISO-date values in `_term_or_match` to a `should` over both the parent field and `.keyword`; (B) add `normalize_condition_operators()` and apply it in `_run_metadata_filter`; (C) return the `["-999"]` no-results sentinel from the `auto`/`semi_auto` branches when a generated filter matches nothing (matching the existing `manual` behaviour). All three are covered by unit tests. Verified against a live Elasticsearch index: bug A `0 → 12` docs, bug B `dropped → 18` docs.

Xavierando:
### Summary

Fixes three bugs that break document-metadata **date** filtering on Elasticsearch (issue #16832). They surface as "date filter returns nothing / returns everything" in chat `auto`/`semi_auto` filtering and in `/api/v1/retrieval`.

**1. Exact-match `=` on a `date`-typed field returned 0.**
ES dynamic date detection types a `YYYY-MM-DD` meta field as `date`, which has no `.keyword` sub-field; `_term_or_match` built a `.keyword`-only term for string values, so `=` on a date field silently matched nothing. ISO-date values now match both the parent field and `.keyword` via a `should`; non-date strings are unchanged (multi-word phrase equality preserved).

**2. ASCII comparison operators were silently dropped.**
`gen_meta_filter` and user conditions may emit `>=`/`<=`/`!=`/`==`, but the push-down (`SUPPORTED_OPERATORS`) and the in-memory path only accept the Unicode set (`≥`/`≤`/`≠`/`=`); an un-normalised operator dropped the whole filter. Added `normalize_condition_operators()` (mirroring `convert_conditions`) and applied it at the top of `_run_metadata_filter`, so every backend benefits.

**3. A filter that matches nothing fell back to *unrestricted* retrieval (latent; exposed by 1 and 2).**
`apply_meta_data_filter` returned `None` from the `auto`/`semi_auto` branches when the generated conditions matched zero documents. `None` means "no metadata filter", so retrieval ran unrestricted and returned arbitrary documents instead of "no results". The `manual` branch already returns the `["-999"]` no-results sentinel here; `auto`/`semi_auto` now do the same, while still returning `None` when no conditions were generated (genuinely no filter). This bug is pre-existing, but only once `=`/range filtering works (fixes 1 and 2) can a date query legitimately match zero documents and reach this fallback — e.g. "what happened yesterday" on a date with no documents previously surfaced unrelated docs with a hallucinated date.

All three fixes have unit tests (`test/unit_test/common/test_metadata_filter.py`, `test/unit_test/common/test_metadata_filter_operators.py`, `test/unit_test/common/test_apply_semi_auto_meta_data_filter.py`) and fixes 1–2 were verified against a live Elasticsearch index (bug 1: `0 → 12` docs; bug 2: `dropped → 18` docs).

Closes #16832


Xavierando:
Thanks for the review @coderabbitai. Both nitpicks addressed in 00f00ad:

- **Shared operator aliases:** extracted \`_COMPARISON_OP_ALIASES\` (\`>=\`, \`<=\`, \`!=\`) now used by both \`convert_conditions\` and \`_OP_ALIASES\`, so the overlapping mappings can no longer diverge.
- **Extended tests:** \`test_normalize_ascii_comparison_operators\` now also covers the \`=>\` / \`=<\` aliases, and a new \`test_normalize_condition_operators_edge_cases\` exercises the \`None\` / empty-list / non-dict-item guard paths.

Xavierando:
@coderabbitai All review findings have been addressed:

- **Stale `Returns:` docstring** (`common/metadata_utils.py`, outside-diff comment): updated in `e822780` — it now documents the `["-999"]` no-results sentinel for `manual`/`auto`/`semi_auto` and `None` only when `auto`/`semi_auto` generate no conditions.
- **Auto-branch `no conditions -> None` test coverage**: added `test_apply_meta_data_filter_auto_no_conditions_returns_none` in `307bda7`, mirroring the `semi_auto` counterpart.
- **`=>`/`=<` typo-variant aliases and `normalize_condition_operators` edge cases** (`None` / empty list / non-dict items): already covered by `test_normalize_ascii_comparison_operators` and `test_normalize_condition_operators_edge_cases` (added in `00f00ad`), so no change was needed.

Thanks for the review!

Xavierando:
Rebased on latest `main` — the branch is now conflict-free and `MERGEABLE`/`CLEAN` again (head `04aa570`).

The only conflict was a tail-append collision in `test/unit_test/common/test_metadata_filter_operators.py` with @chuenchen309's #16956 (merged July 16), which added case-insensitivity + per-entry state-reset fixes to `filter_out`. Both test sets are kept; the two changes are orthogonal in `common/metadata_utils.py` (his: `filter_out` operand coercion; this PR: `convert_conditions`/operator aliases + the `apply_meta_data_filter` no-results sentinel). Full metadata suite green locally — **102 passed**, including `test_normalised_ge_filters_dates_in_memory`, which exercises the exact `filter_out` path #16956 rewrote, so the merge is behavior-verified, not just textually clean. `ruff check` clean, `ruff format` applied.

**Recap for reviewers** — this fixes three bugs in ES metadata **date** filtering (#16832):
- **A** — `=` on a `date`-typed field returns 0: `_term_or_match` builds `term` on `<field>.keyword`, but ES types `YYYY-MM-DD` as `date`, which has no `.keyword` sub-field. Fix: emit a `should` over both the parent field and `.keyword`.
- **B** — ASCII operators silently dropped: `gen_meta_filter`/user input can emit `>=`,`<=`,`!=`,`==`, but `SUPPORTED_OPERATORS` only accepts the Unicode set, so the whole filter is discarded and retrieval runs unrestricted. Fix: `normalize_condition_operators()` applied in `_run_metadata_filter`.
- **C** (latent, exposed once A+B are fixed) — `apply_meta_data_filter`'s `auto`/`semi_auto` branches returned `None` (= "no filter") when a generated filter matched nothing, falling back to unrestricted retrieval. Fix: return the `["-999"]` no-results sentinel, matching the existing `manual` branch.

**Cross-PR note for @chuenchen309 / maintainers** — there's a load-bearing ordering dependency with the open #16943 (Infinity push-down rejecting multi-valued-unsafe negative ops). That guard keys on the **Unicode** `≠`/`not in`; an ASCII `!=` reaching `is_pushdown_supported` before normalization would slip past it. Bug B's `normalize_condition_operators` is the precondition that makes the Unicode-only guard sound on the LLM/auto path. The two PRs are complementary and, ideally, land together.

Beyond these three point-fixes, @zeweihan's framing on #16832 and #16967 (backend-specific schema/semantics drift — "same query, different backend, materially different answer, zero error signal") points at a durable follow-up: a backend-parity test asserting ES/Infinity/in-memory return identical filter sets, plus a `metadata_filter_coverage` signal exposing dropped/no-matched operators at query time.

One note for maintainers: the `ragflow_tests_elasticsearch`/`_infinity` jobs came back **SKIPPED** (fork PR awaiting CI approval), so the ES path this fixes hasn't actually run in CI yet — a workflow-run approval would let it verify end-to-end.

JinHai-CN:
@sxxtony Would you please help review this PR?

qinling0210 (changes requested):
@Xavierando Please rebase your code change to latest code.

And make sure you have verified your code change.

Issue #16838, opened 2026-07-12
[Bug]: validate_immutable_fields / validateImmutableDocumentFields allow overwriting run and progress_msg


Harsh23Kashyap:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

891261e1 (current upstream `main`)

### RAGFlow image version

`main` source; reference v0.26.4

### Other environment information

- Python backend (Flask + Pydantic v2 + GORM-equivalent via peewee)
- Go backend (Gin + GORM, currently mid-migration)
- Single deployment under test
- Document engine: Elasticsearch 8.11.3 (docker)

### What happened?

`validate_immutable_fields` in [`api/utils/validation_utils.py:1059`](https://github.com/infiniflow/ragflow/blob/891261e1/api/utils/validation_utils.py#L1059) and the new `validateImmutableDocumentFields` in [`internal/service/document.go`](https://github.com/infiniflow/ragflow/blob/891261e1/internal/service/document.go) (just shipped in PR #16836) both reject attempts to overwrite `chunk_count` / `token_count` / `progress` via `PATCH /api/v1/datasets/{dataset_id}/documents/{document_id}` (Python) and `PUT /api/v1/documents/:id` (Go).

The original issue **#16831** explicitly listed five ingestion-managed fields: `progress`, `run`, `progress_msg`, `chunk_num`, `token_num`. The fix scoped to "parity with Python" — but Python's `validate_immutable_fields` was also incomplete. An external caller can still overwrite two of the five fields:

```http
PUT /api/v1/documents/{document_id}
Content-Type: application/json

{"run": "0"}
```

```http
PUT /api/v1/documents/{document_id}
Content-Type: application/json

{"progress_msg": "fake"}
```

[...]

Issue #16831, opened 2026-07-11
Go PUT /api/v1/documents/:id misses immutable-field validation that Python has (progress/run/chunk_num/token_num can be overwritten)


xugangqiang:
## Background

In Python, `PATCH /api/v1/datasets/{dataset_id}/documents/{document_id}` (implemented by `update_document` in `api/apps/restful_apis/document_api.py`) protects ingestion-managed document fields via a three-layer defense:

1. **`UpdateDocumentReq`** (in `api/utils/validation_utils.py:468`) accepts `progress`, `chunk_count`, `token_count` in the request schema — for backward compatibility with existing callers, not because they are writable.
2. **`validate_immutable_fields()`** (in `api/utils/validation_utils.py:1059`) rejects any non-zero change to these fields:
   - `progress`: rejected unless equal to DB value (within `math.isclose` tolerance)
   - `chunk_count`: rejected unless equal to `chunk_num` in DB
   - `token_count`: rejected unless equal to `token_num` in DB
3. **The handler body never writes these fields** — it only processes `name`, `parser_config`, `pipeline_id`, `chunk_method`, `enabled`, and `meta_fields`. The protected fields pass through validation and are silently dropped.

Python has no `PUT /api/v1/documents/:id` endpoint. The deprecated `PUT /datasets/{dataset_id}/documents/{document_id}` in `backward_compat.py:562` simply forwards to the PATCH handler.

## Gap in Go migration

Go introduced two document update endpoints with **inconsistent validation**:

| Endpoint | Go method | Validation |
|---|---|---|
| `PUT /api/v1/documents/:id` | `UpdateDocument` (`internal/service/document.go:598`) | **None** — loads the full entity via `GetByID`, overwrites any non-nil request field, then persists via `DB.Save` (GORM full-record upsert) |
| `PATCH /api/v1/datasets/:dataset_id/documents/:document_id` | `UpdateDatasetDocument` | ✅ Has `validateDatasetDocumentUpdate` (line 2640) — correctly prevents non-zero changes to `progress`, `chunk_count`, `token_count` |

The PUT endpoint's `UpdateDocument` unconditionally sets (lines 603–618):
```go
if req.Progress != nil    { document.Progress = *req.Progress }
if req.Run != nil         { document.Run = req.Run }
if req.ProgressMsg != nil { document.ProgressMsg = req.ProgressMsg }
if req.TokenNum != nil    { document.TokenNum = *req.TokenNum }
[...]

PR #16857, opened 2026-07-13
Fix OpenDataLoader Hybrid wiring and from_page/to_page redundant re-parsing (OpenDataLoader/MinerU/Docling)

weijyhnjean:
## Summary

Three related fixes found and verified while running the OpenDataLoader PDF parser integration end-to-end:

- **Hybrid mode wasn't reachable through RAGFlow's own Dataset/Document Configuration API.** `ParserConfig` (`api/utils/validation_utils.py`) had no field for `hybrid`/`image_output`/`sanitize`, so a dataset's `parser_config` silently dropped them under Pydantic's `extra="forbid"`. Added the fields so they survive the API round-trip.
- **`hybrid_mode` had no client-side support at all.** `OpenDataLoaderParser.parse_pdf()` (`deepdoc/parser/opendataloader_parser.py`) had no `hybrid_mode` parameter, so even after the fix above, RAGFlow could request the Hybrid backend but not force full-page OCR — the Java core's default `auto` per-page triage could still skip a scanned page entirely. Added the parameter end to end.
- **Multi-task documents redundantly re-parsed the entire PDF on every task.** RAGFlow splits large PDFs into multiple page-range task-executor jobs (`from_page`/`to_page`), but `by_mineru()`/`by_docling()`/`by_opendataloader()` (`rag/app/naive.py`) all accepted these parameters and never forwarded them to their respective parsers — every task re-sent and re-chunked the *whole* document, not just its assigned range. This inflated `document.chunk_num` to N× the real chunk count for an N-task document (the real Elasticsearch index self-deduplicates by deterministic chunk ID, so retrieval was unaffected, but the file-list counter and the redundant parsing work were real). Fixed for all three backends:
  - **OpenDataLoader**: threads `from_page`/`to_page` into a new `pages` parameter (adapter-side, not in this diff).
  - **Docling**: converts to Docling's `page_range` (1-indexed, inclusive) for both the local `DocumentConverter` path and the remote `docling-serve` path. Verified locally that Docling preserves absolute page numbers when `page_range` is set (live `conv.convert(..., page_range=(3,4))` test against a real multi-page PDF), and confirmed `docling-serve`'s `ConvertDocumentsOptions.page_range` is the exact same Pydantic field/type as the local package.
  - **MinerU**: converts to `start_page_id`/`end_page_id`. This one needed an extra correction — MinerU's remote API pre-slices the PDF to the requested range before analysis (confirmed by reading `mineru.cli.common.convert_pdf_bytes_to_bytes` in the real `mineru==3.4.4` package), so the returned `page_idx` is relative to that slice, not the original document. Rebased back to absolute page numbers immediately after reading the response, before any position/citation logic runs, so bounding boxes/citations stay correct.

Also adds a small "OpenDataLoader Options" section to the Dataset Configuration screen (Hybrid backend / Hybrid mode / Sanitize output), following the existing `MinerUOptionsFormField`/`PaddleOCROptionsFormField` pattern, so these settings are reachable from the UI rather than API-only.

## Test plan

- [x] Verified Hybrid mode end-to-end with real OCR output on a scanned Traditional-Chinese PDF, driven entirely through RAGFlow's own Dataset/Document Configuration REST API (no direct calls to the underlying OpenDataLoader service).
- [x] Verified `hybrid_mode: full` forces full-page OCR (previously the Java core's `auto` triage could skip pages even with `hybrid` set).
- [x] Verified the `from_page`/`to_page` fix on two previously-affected multi-task PDFs for OpenDataLoader: `chunk_count` now matches the real chunk count exactly (no more N× inflation), each task's log shows it requesting only its own page range.
- [x] Verified Docling's `page_range` behavior with a live `docling` package call (not just read from docs) before relying on it.
- [x] Verified the MinerU page-slicing/rebasing behavior by reading the actual `mineru` package source (`start_page_id`/`end_page_id` pre-slice the PDF; `page_idx` in the response is relative to that slice) — no live MinerU server was available to test end-to-end, so this part rests on source-reading rather than a live call; flagged explicitly rather than presented as fully verified.
- [x] Frontend UI verified via browser screenshot showing the new section correctly rendering an existing saved `hybrid`/`hybrid_mode` configuration, no console errors.
- [x] Regression-tested the two previously-affected OpenDataLoader documents after all three fixes were combined — `chunk_count` still correct.


zeweihan:
This is a genuinely sharp set of fixes, and the third one (forwarding `from_page`/`to_page`) is the one that matters most for retrieval correctness — but for a subtler reason than "wasted work." Two angles from regulated-document corpora where page numbers are citation anchors:

1. **The inflated `chunk_num` is a provenance lie, not just a wasteful recount.** The PR body correctly notes retrieval is unaffected because ES self-dedups by deterministic chunk ID. That's true for *recall*. But `document.chunk_num` is the system's own record of "how many chunks did extraction produce for this document," and every consumer that trusts it (file-list counters, billing, capacity dashboards, and — critically — any audit/completeness check that asks "did we ingest the whole document") now reads a number that's N× the truth. That's the same shape as the `validate_immutable_fields` gaps on `run`/`chunk_num`/`token_num` (#16831 / #16838): an ingestion-derived field that drifted from what ingestion actually did, persisted as healthy, zero degradation signal. The dedup saves the *index* but not the *accountability surface* built on top of `chunk_num`. Worth noting in the PR that the counter being wrong (not just redundant) is the reason this is worth fixing even though retrieval self-heals.

2. **The MinerU `page_idx` rebasing is the dangerous part, and the "rebased back to absolute" step is exactly right — but the failure mode it prevents deserves to be explicit, because it's silent.** MinerU pre-slices the PDF to the requested range before analysis, so the `page_idx` it returns is *relative to the slice*, not the original document. If that weren't rebased, a chunk covering page 3 of a slice starting at original page 40 would carry `page_idx=3` — and every downstream citation would say "see page 3" when the operative content is on page 43. That's a **citation-anchor mutation**: the text survives perfectly, the binding "this clause is on page X" is silently wrong, full confidence, plausible-but-wrong. The author caught this by reading the real `mineru==3.4.4` source, which is the right instinct — but the broader lesson is that *any* parser that slices before analysis (Docling's `page_range`, MinerU's `start_page_id`/`end_page_id`) silently re-bases provenance coordinates at the parser boundary, and the ingestion layer is the only place that knows the original offset to rebase them back. A failure here is invisible until someone cites the wrong page in a document where page-citation precision is operative.

A cheap guard for the rebasing class, if the team wants to make it inspectable rather than rely on each adapter remembering to rebase: a per-chunk stamp like `page_provenance{task_from_page, task_to_page, parser_returned_relative: bool, rebased: bool, absolute_page_range}`. If `parser_returned_relative && !rebased`, the chunk is a known-bad citation anchor and can be flagged at ingestion rather than discovered months later when someone cites page 3 and it's actually page 43. Same philosophy as `text_coverage` / `formula_render_status` — turn a silent coordinate-system mismatch into a queryable property. The Docling path getting `page_range` (1-indexed, inclusive) and the MinerU path getting `start_page_id`/`end_page_id` both need the same rebasing discipline, so a shared convention would keep the next backend from reintroducing the slice-relative trap.

xugangqiang:

**1. [HIGH] `hybrid="off"` is forwarded to the OpenDataLoader service as if it were an active backend.**
`chunk()` forwards any `hybrid` value where `value is not None` (`rag/app/naive.py`). The new UI `OpenDataLoaderOptionsFormField` defaults the Hybrid select to `'off'` (`value={field.value || 'off'}`), so every dataset saved through the new UI persists `parser_config.hybrid = "off"`. In `deepdoc/parser/opendataloader_parser.py` the guard is `if hybrid:` (any non-empty string is truthy), so `hybrid=off` gets sent to `/file_parse`. The backend likely treats the mere presence of `hybrid` as "enable", which silently turns on Hybrid for the default path.
Fix:
```python
if hybrid and hybrid != "off":
    form_data["hybrid"] = hybrid
    if hybrid_mode:
        form_data["hybrid_mode"] = hybrid_mode
```
Also consider dropping `"off"` in the `chunk()` forwarding for defense in depth.

**2. [MEDIUM] Inaccurate "byte-for-byte identical for single-task" claim.**
The PR comments assume a single-task doc passes `to_page = MAXIMUM_PAGE_NUMBER`. In production, `api/db/services/task_service.py` always sets `to_page` to a finite real page count (`min(p + page_size, e)`), so `to_page < MAXIMUM_PAGE_NUMBER` is always true and the parser **always** sends an explicit range (`page_range=(1, N)` for Docling, `pages="1-N"` for OpenDataLoader, `end_page_id=N-1` for MinerU). The `page_range=None` / no-`pages`-flag branches are effectively dead in production. Behavior is functionally equivalent (parses exactly the assigned range), but the comments overstate the no-op guarantee and should be corrected.

**3. [MEDIUM] No automated tests for the riskiest logic.**
The entire test plan is manual (live OCR, browser screenshots, source reading). The two most failure-prone spots — (a) 0-indexed half-open ↔ 1-indexed inclusive conversion for Docling/OpenDataLoader, and (b) MinerU relative→absolute `page_idx` rebasing (explicitly noted as "rests on source-reading rather than a live call") — have zero unit coverage. Please add unit tests covering: whole-doc, first sub-range, middle sub-range, and boundary off-by-one cases for all three backends.

**4. [LOW] Validation `hybrid` should be a `Literal`.**
In `api/utils/validation_utils.py`, `hybrid: str | None` accepts any string. Constrain it to `Literal["off", "docling-fast", "hancom-ai"] | None` to make the disabled `"off"` semantics explicit at the API boundary (complements fix #1).

**5. [LOW] Frontend nits (per CodeRabbit).**
- Rename top-level constants `hybridOptions`/`hybridModeOptions` to PascalCase (`HybridOptions`/`HybridModeOptions`) per `web/**` lint rules.
- Move the dropdown option labels through `t()` with new locale keys instead of hard-coded English.

**Note on the `to_page=-1` suggestion:** not needed. Every real chunk task gets a finite `to_page` from `task_service.py` (`new_task()` defaults to `MAXIMUM_TASK_PAGE_NUMBER`, PDF/excel override with real ranges); the `-1` default in `task_context.py` is only a fallback that production paths never surface. Adding `-1` normalization is unnecessary defensive code.

**Note on MinerU rebasing + `section_page_offset`:** verified safe — `naive.py:1160` `section_page_offset=from_page` only affects table/image context pages, and the rebasing restores the exact absolute `page_idx` that pre-PR whole-doc parsing produced, so there is no double-offset.

xugangqiang:
pls resolve conflict

PR #16858, opened 2026-07-13
fix: support doc generator component

Haruko386:
### Summary

As title

image

---

image


yuzhichang (changes requested):
I prefer do it with golang lib instead of external commands.

yuzhichang:
LGTM

PR #16860, opened 2026-07-13
fix: ensure database model indexes

buua436:

Ensure database indexes declared in Peewee models are checked and created during database initialization.

This also adds the missing `evaluation_runs.dialog_id` column migration before index creation, preventing index creation failures on existing databases.

### Type of change

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -43,7 +43,6 @@ from peewee import (
     Metadata,
     Model,
     TextField,
-    PrimaryKeyField,
 )
 from playhouse.migrate import MySQLMigrator, PostgresqlMigrator, migrate
 from playhouse.pool import PooledMySQLDatabase, PooledPostgresqlDatabase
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
## Review: PR #16860 — fix: ensure database model indexes

**Author:** @buua436  
**Files:** `api/db/db_models.py` (+39, −1)  
**Labels:** bug, ci, size:M  

### Summary

This PR introduces `ensure_model_indexes(migrator)` — a generic function that introspects Peewee model declarations and creates any missing database indexes. It also adds the missing `evaluation_runs.dialog_id` column before calling the reconciliation, and removes the unused `PrimaryKeyField` import.

### What looks good ✅

1. **Comprehensive approach** — `ensure_model_indexes` scans all `DataBaseModel` subclasses, collects field-level (`field.index`, `field.unique`) and model-level (`_meta.indexes`) indexes, compares them against `DB.get_indexes(...)`, and creates what's missing. This is a solid pattern that will catch any future model index additions automatically.

2. **Error handling is reasonable** — Each index creation is wrapped in try/except with logging, so a single failure won't crash the entire migration.

3. **Correct ordering** — `dialog_id` column is added *before* the index reconciliation runs, which is the right sequence.

4. **Idempotent** — Existing indexes are skipped; duplicate column errors (MySQL 1060) are already handled by `alter_db_add_column`; index creation errors are caught.

5. **Nice cleanup** — Removing the unused `PrimaryKeyField` import.

### Concerns & suggestions ⚠️

#### 1. Potential data integrity issue: `dialog_id = ""` for existing rows

The `dialog_id` column is added with `default=""` and `null=False`. For any **existing rows** in the `evaluation_runs` table, this means they'll get an empty string `""` as their `dialog_id`. 

Looking at the model definition:
```python
dialog_id = CharField(max_length=32, null=False, index=True, help_text="dialog configuration being evaluated")
```

Since `dialog_id` references a dialog configuration being evaluated, empty string rows may indicate an inconsistent state. **Consider whether a backfill step is needed** — e.g., for each existing row, relate it to the evaluation run's dataset agent, or allow NULL and handle it at the application layer.

**Suggestion:** Either:
- Set `null=True` on the column migration (but keep the model-level `null=False` as a forward-facing constraint), OR
- Add a data migration step to backfill `dialog_id` from the associated `evaluation_cases` or `evaluation_datasets` before the index creation.

#### 2. Silent failure on unique constraint violations

When `ensure_model_indexes` tries to create a **UNIQUE index** on a column that already has duplicate values, the `add_index` call will fail and the error is logged but the migration continues silently. This could leave the database in a state where the model declares a unique constraint but the database doesn't enforce it.

**Suggestion:** For unique index failures, consider raising a critical error or at least logging with a louder level (`logging.critical` or `logging.error` with more context), so operators know they need to deduplicate data manually.

#### 3. `dialog_id` index will be created redundantly by `ensure_model_indexes`

The `dialog_id` column is defined with `index=True` in the model and added via `alter_db_add_column`, but the migration only adds the column — not the index. The index is then picked up by `ensure_model_indexes`. This is intentional and works, but it's worth noting that the `dialog_id` index appears in the "missing" list during the reconciliation, resulting in a two-step column-then-index migration that could have been a single DDL statement.

Not a bug, just an observation — the current approach is fine for correctness.

#### 4. Minor: scope of introspection

`inspect.getmembers(sys.modules[__name__], inspect.isclass)` will also pick up any test or utility classes defined in the same module that happen to inherit from `DataBaseModel`. In the current codebase this is unlikely to be an issue, but could be surprising if new model-like classes are added.

#### 5. Performance on large databases

`ensure_model_indexes` calls `DB.get_indexes(table_name)` for every model subclass, then issues individual `migrator.add_index` calls for each missing index. On a busy production database with many tables, this one-time migration cost is acceptable, but it's worth being aware of.

### Verdict

This is a well-structured PR that addresses a real gap — model-declared indexes not being created. The `ensure_model_indexes` function is a smart, maintainable solution. The main concern is the data integrity of existing `evaluation_runs` rows getting an empty `dialog_id` string, which should be reviewed.

**Recommendation:** Approve after considering the backfill question for existing `evaluation_runs.dialog_id` rows.

PR #16866, opened 2026-07-13
Fix: Search mind map not displaying

euvre:
### Summary

The search query mind map was completely non-functional due to multiple bugs across the backend and frontend:

**Backend:**
- **Wrong LLM message role**: The mind map prompt was sent as a `user` message instead of `system`, resulting in two consecutive `user` messages. Many LLM providers reject or mishandle this, causing the mind map generation to fail. Fixed to match the Python implementation (`extractor.py`) which uses `system` + `user`.
- **Premature HTTP timeout**: The driver HTTP client had `ResponseHeaderTimeout: 60s`, which is shorter than the per-call context timeout (`nonStreamCallTimeout = 120s`). For non-streaming LLM calls (like mind map generation), the server only sends response headers after the full response is generated. The 60s limit caused premature timeouts. Set to `0` to rely on the context timeout.

**Frontend:**
- **G6 Graph initialization race condition**: `IndentedTree` created an empty Graph instance and then called `setOptions` + `setData` in a separate effect. G6 v5 canvas initialization is asynchronous, so the second effect could run before the canvas was ready, causing `render()` to silently fail. Fixed by passing all options and data directly to the Graph constructor.
- **Invisible nodes**: Node size was hardcoded to `[0.1, 0.1]` pixels, making nodes effectively invisible. Changed to compute actual sizes based on text content.
- **Mind map button not visible**: The button used `size-6` (24px), `z-30`, and no background — it was nearly invisible inside a `text-transparent` parent container. Enlarged to `size-8`, raised to `z-50`, and added background/text colors.
- **Zero-height container**: `SheetContent` used `h-auto` which overrode `h-full`, breaking the CSS height chain. The G6 container received zero height and could not render. Fixed by using `top`/`bottom` positioning.
- **No empty-state handling**: When the API returned empty data, a blank canvas was rendered. Added a `No data` fallback message.
- **Removed unused props**: `onRender` and `onDestroy` were never used by any caller.

### Type of change

### DIFF
--- a/internal/entity/models/base_model.go
+++ b/internal/entity/models/base_model.go
@@ -191,7 +191,7 @@ func NewDriverHTTPClient() *http.Client {
 	t.MaxIdleConnsPerHost = 10
 	t.IdleConnTimeout = 90 * time.Second
 	t.DisableCompression = false
-	t.ResponseHeaderTimeout = 0
+	t.ResponseHeaderTimeout = 60 * time.Second
 	return &http.Client{Transport: t}
 }
 
--- a/internal/entity/models/base_model.go
+++ b/internal/entity/models/base_model.go
@@ -191,7 +191,7 @@ func NewDriverHTTPClient() *http.Client {
 	t.MaxIdleConnsPerHost = 10
 	t.IdleConnTimeout = 90 * time.Second
 	t.DisableCompression = false
-	t.ResponseHeaderTimeout = 60 * time.Second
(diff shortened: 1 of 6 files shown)


dcc123456 (commented):
Connecting to the Python service works fine, but connecting to the Go service has issues. We should  
  only need to modify the Go service's return value, not the frontend.

PR #16869, opened 2026-07-13
fix(file-manager): allow moving files to/from .knowledgebase subfolders

euvre:
### Summary

This PR fixes the file manager move dialog so that files can be moved into `.knowledgebase` subfolders (dataset mirror folders), and removes the restrictions that prevented KB-sourced files from being selected, moved, renamed, or deleted.

**Problem:** In the file management UI, when moving a file, the target tree only showed the `.knowledgebase` folder itself but not its subfolders. Additionally, files with `source_type === "knowledgebase"` had their action buttons (move, rename, delete, link-to-dataset) hidden and their row checkboxes disabled, making them impossible to manage.

**Changes:**

- **`web/src/pages/files/action-cell.tsx`**: Removed `isKnowledgeBaseType` gating so that KB-sourced files show all action buttons (link, move, rename, delete) like any other file.
- **`web/src/pages/files/hooks.ts`**: Removed `getCheckboxProps` that disabled checkboxes for KB-sourced files.
- **`web/src/pages/files/files-table.tsx`**: Removed `isKnowledgeBaseType` check in `enableRowSelection`, making KB-sourced files selectable.
- **`web/src/pages/files/move-dialog.tsx`**: Removed unused `isFolderType` import and simplified the skills filter condition.
- **`internal/service/file.go`**: Fixed duplicate-name check in `MoveFiles` to exclude the source file itself (`f.ID != file.ID`), so renaming a file to its current name succeeds instead of failing with "Duplicated file name".
- **`internal/service/file_move_kb_test.go`**: Added tests covering moving files to KB subfolders, moving between KB subfolders, the `has_child_folder` flag for `.knowledgebase`, and the rename-to-same-name fix.


euvre:
The requirements have changed again. Close this PR.

PR #16877, opened 2026-07-13
Fix: RAGFlowJsonParser crashes with IndexError on top-level JSON scalars

devYRPauli:
### Problem

Uploading a `.json` file whose top-level value is a bare JSON scalar - a number, string, boolean, or `null`, all of which are valid JSON - crashes `RAGFlowJsonParser` with an uncaught `IndexError`.

`_parse_json` only guards `json.JSONDecodeError`, so the `IndexError` propagates out of the parser and aborts the upload (the call site at `rag/app/naive.py` runs this for every `.json`/`.jsonl` file).

### Root cause

In `_json_split`, the branch that handles a single non-dict item calls:

```python
self._set_nested_dict(chunks[-1], current_path, data)
```

`_set_nested_dict` ends with `d[path[-1]] = value`. For a top-level scalar, `current_path` is empty, so `path[-1]` raises `IndexError: list index out of range`. Objects and arrays are unaffected (they recurse through the dict branch).

### Fix

A top-level scalar has no key to nest under, so store it as the chunk directly:

```python
if not current_path:
    chunks[-1] = data
else:
    self._set_nested_dict(chunks[-1], current_path, data)
```

### Test

Adds `test/unit_test/deepdoc/parser/test_json_parser.py`. It loads the parser in isolation and checks that top-level scalars (`42`, `"hello"`, `true`) no longer crash, `null` yields no chunk, and objects/arrays still chunk as before. Reverting the one-line fix makes the scalar cases fail with the original `IndexError`.


devYRPauli:
The failing `ragflow_preflight` check looks like a CI network flake rather than anything in this diff.

The traceback is in the pytest collection phase, not in the json_parser change: importing `agent/component/llm.py` triggers `tiktoken.load_tiktoken_bpe()`, which does a `requests.get()` to download the `cl100k_base` BPE file, and that HTTPS fetch failed with an SSL EOF (`SSLEOFError` / `HTTPSConnectionPool ... unexpected_eof_while_reading`).

The same commit (1c40863) actually passed `ragflow_preflight` in a parallel run on this PR, so the outcome is non-deterministic and tied to the network fetch. This diff only touches `deepdoc/parser/json_parser.py` and adds an isolated unit test that stubs `rag.nlp`, so it never reaches that import path.

Could a maintainer re-run the failed job when convenient? Happy to rebase if that helps.

devYRPauli:
Good catch from the bot review - this was a real gap, now fixed in e7671f3c.

The top-level-scalar handling itself was fine, but the value then flowed through two pre-existing truthiness filters that cannot distinguish "no value" from "a falsy value":

- `split_json`: `if not chunks[-1]: chunks.pop()`
- `_parse_json` / `_parse_jsonl`: `... for line in chunks if line`

So `_parse_json("0")` and `_parse_json("false")` returned `[]`, identical to `_parse_json("null")` - silently discarding legitimate content.

Added `_is_empty_chunk` and routed the three filters through it, so only `null` and empty containers count as empty:

| input | before | after |
|---|---|---|
| `0` | `[]` | `["0"]` |
| `false` | `[]` | `["false"]` |
| `null` | `[]` | `[]` |
| `""` / `{}` / `[]` | `[]` | `[]` |

`null`, `""`, `{}` and `[]` keep their existing behavior, and objects/arrays are unaffected. Regression coverage added for the falsy scalars and for the preserved empty cases.

Note the earlier CI failures on this PR were unrelated infra (a transient SSLError fetching a tiktoken encoding for `test_akshare.py`, plus two jobs cancelled before a runner was assigned) - not caused by this change.

xugangqiang:
LGTM

PR #16881, opened 2026-07-13
Fix: Q&A CSV parser splices wrong text when a field opens with a quote

### LINKED ISSUE #16791: [Bug]: QA parser for .csv silently corrupts Q&A pairs: csv.reader quote handling + row/line index mismatch splice answers from wrong lines
### Self Checks

- I have searched for existing issues (including closed ones). #16433 touched this code
  but only fixed the row number of the final pair; the defects below are still present
  on current `main`.

### RAGFlow version

Observed on a v0.26.x deployment; the affected code is still present on current `main`
(`rag/app/qa.py`, csv branch, ~lines 347–375).

### Actual behavior

When parsing a TAB-delimited `.csv` with the **Q&A chunk method**, chunks silently get
content from the *wrong* lines of the file:

- answers absorb raw rows of *other* Q&A pairs (including the TAB and the other pair's
  question),
- some continuation lines disappear entirely,
- some Q&A pairs vanish without any failure being reported.

On a real 679k-line / 13,435-pair file (Dutch helpdesk tickets, where memo lines
regularly start with a `"` character), this produced 13,412 chunks of which **7,902
contained the content of more than one ticket**, and 10 questions were missing
altogether — with `progress_msg` reporting success.

### Expected behavior

Each `question<TAB>answer` row becomes exactly one chunk; continuation lines (no TAB)
are appended to the answer of the *current* pair, as documented.

### Steps to reproduce

The csv branch does this (simplified from `rag/app/qa.py`):

```python
lines = txt.split("\n")
reader = csv.reader(lines, delimiter=delimiter)
    if len(row) != 2:
        if question:
            answer += "\n" + lines[i]     # <-- BUG 2
    elif len(row) == 2:
        ...
        question, answer = row
```

Two defects interact:

1. **Unwanted quote handling.** `csv.reader` is used with the default
   `quotechar='"'`. Any continuation line that *starts* with `"` and does not close
   the quote on the same line opens a quoted field that swallows the following
   physical lines (including whole `Q<TAB>A` rows) into a single reader record.
2. **Row index used as line index.** In the `len(row) != 2` fallback, the reader's
   *record* index `i` is used to index the *physical* `lines` list. After any
   multi-line quoted record, `i` permanently lags the physical line number, so every
   subsequent fallback appends text from the wrong line — this is how other tickets'
   rows end up inside unrelated answers. The lag grows with every quote-open line.

Minimal reproduction (pure Python, mirrors the loop above):

```python
import csv

lines = [
    'Q1: how to reset password\tA1 step one',
    '"see the attached error log',              # opens a quote, does not close it
    'error: disk full"',                        # closes the quote
    'Q2: printer offline\tA2 restart spooler',
    'done',
    'Q3: vpn drops\tA3 update client',
    'reboot if needed',
]

reader = csv.reader(lines, delimiter="\t")
question, answer, res = "", "", []
    if len(row) != 2:
        if question:
            answer += "\n" + lines[i]
    else:
        if question and answer:
            res.append((question, answer))
        question, answer = row
if question:
    res.append((question, answer))

for q, a in res:
    print(repr(q), "->", repr(a))
```

Actual output:

```
'Q1: how to reset password' -> 'A1 step one\n"see the attached error log'
'Q2: printer offline' -> 'A2 restart spooler\nQ2: printer offline\tA2 restart spooler'
'Q3: vpn drops' -> 'A3 update client\nQ3: vpn drops\tA3 update client'
```

- Q1 lost the line `error: disk full"`.
- Q2 and Q3 never receive their real continuation lines (`done`, `reboot if needed`);
  instead their own raw row text (TAB included) is spliced into their answers. In a
  large file the growing offset splices *other* pairs' rows instead.

Expected output:

```
'Q1: how to reset password' -> 'A1 step one\n"see the attached error log\nerror: disk full"'
'Q2: printer offline' -> 'A2 restart spooler\ndone'
'Q3: vpn drops' -> 'A3 update client\nreboot if needed'
```

### Suggested fix

The QA csv format is documented as "UTF-8, TAB as delimiter"; quote characters have no
documented meaning, and free-text answers legitimately contain them. Two options:

1. `csv.reader(lines, delimiter=delimiter, quoting=csv.QUOTE_NONE)` — every physical
   line is then exactly one record, which fixes both defects at once (`i` stays in
   sync with `lines`), or
2. keep quote handling but stop indexing `lines[i]`: reconstruct the raw line from the
   parsed record (`delimiter.join(row)`), or track the physical position via
   `reader.line_num`.

Option 1 is the least invasive and matches the documented format. Happy to submit a PR
for either direction, including a regression test based on the reproduction above.

### Additional information

The `.txt` branch has a related weakness: the comma/TAB detection counts lines that
split into exactly two parts, so natural-language answers full of salutations like
"regards, John" can outvote the actual TAB rows and flip the delimiter to comma for
the whole file. A shared fix could prefer TAB whenever any TAB-row is present.

devYRPauli:
### Problem

Parsing a Q&A `.csv` can splice unrelated text into the wrong answers (reported in #16791).

### Root cause

The `.csv` branch of `rag/app/qa.py`'s `chunk()` builds records with `csv.reader(lines, delimiter=delimiter)` (default `quotechar='"'`), but then indexes `lines[i]` by the reader's *record* index in `answer += "\n" + lines[i]`. When a line's field opens with a `"`, `csv.reader` treats it as an unclosed quoted field and merges several physical lines into one record. From there the record index permanently desyncs from the physical line numbers, so `lines[i]` returns the wrong line and unrelated Q&A content gets appended to the wrong answer.

### Reproduction (stdlib only)

```python
import csv
lines = 'Q1,A1\n"quoted answer start\ncontinues here,extra\nQ2,A2\n'.split("\n")
list(csv.reader(lines, delimiter=","))
# record 1 swallows 3 physical lines: ['quoted answer startcontinues here,extraQ2,A2']
# -> the reader index no longer matches lines[i]

list(csv.reader(lines, delimiter=",", quoting=csv.QUOTE_NONE))
# one physical line per record; index stays aligned
```

### Fix

Pass `quoting=csv.QUOTE_NONE` so one physical line maps to one record, keeping the reader index aligned with `lines[i]` (the surrounding code already relies on that 1:1 mapping).

Fixes #16791.


devYRPauli:
Revisiting my own approach here - the `QUOTE_NONE` fix was wrong, and af48ebb3 replaces it.

Forcing `QUOTE_NONE` did keep the reader's row index aligned with `lines[i]`, but it regressed valid input. A legitimately quoted field containing the delimiter parses differently:

| input | before (QUOTE_MINIMAL) | with QUOTE_NONE |
|---|---|---|
| `"hello, world",answer` | `['hello, world', 'answer']` (2 fields) | `['"hello', ' world"', 'answer']` (3 fields) |

The 3-field row then fails the `len(row) != 2` check and gets misfiled as a parse failure, so the fix traded one bug for another on input that previously worked.

The underlying issue was never the quoting - it was that `i` is the **row** index while `lines[i]` is a **physical line** index, and those drift apart precisely when a quoted field spans lines. So keep default quote handling and track the real span with the reader's own `line_num`:

```python
reader = csv.reader(lines, delimiter=delimiter)
prev_line_num = 0
    raw = "\n".join(lines[prev_line_num : reader.line_num])
    prev_line_num = reader.line_num
    ...
            answer += "\n" + raw   # was lines[i]
```

`i` stays the row index, so `top_int` is unchanged and `test_csv_final_pair_uses_last_line_number` still passes.

This PR also shipped no tests, which I should have caught earlier. Added two, both of which fail against the previous QUOTE_NONE approach and pass now:

- `test_csv_quoted_comma_stays_in_one_field` - fails as `assert 1 == 2` on the old code
- `test_csv_multiline_quote_uses_physical_continuation_line` - reproduces the original wrong-line splice

`uv run pytest test/unit_test/rag/app/test_qa_csv.py -q` -> 3 passed; `ruff check` clean.

qinling0210 (approved):
LGTM

PR #16889, opened 2026-07-14
fix: resolve provider name case mismatch and 301 redirect in Go mode

euvre:
### Summary

This PR fixes three issues in Go mode that prevent the model provider settings page from working correctly:

1. **301 Moved Permanently on group-level routes**: Gin routes registered with `"/"` (trailing slash) on `/providers`, `/models`, and `/connectors` groups caused 301 redirects when the frontend requests the slashless URL (e.g. `GET /api/v1/providers`). Python registers these routes without a trailing slash. Changed `provider.GET("/")` → `provider.GET("")`, etc. Also removed the trailing slash from `addProvider` in `web/src/utils/api.ts`.

2. **Provider name case mismatch**: The Go service layer called `strings.ToLower(providerName)` before writing to the database, while Python preserves the original casing from `conf/llm_factories.json` (e.g. `"SILICONFLOW"`, `"ZHIPU-AI"`). This caused composite model IDs (e.g. `"gpt-4@OpenAI"`) to embed a provider name that could not be matched against the lowercase DB value during lookup. Removed `strings.ToLower` in 6 service functions to align with Python behavior.

3. **"No instances configured" after saving**: `ListProvidersOfTenant` overwrote the catalog provider name (returned by `GetProviderByName`) with the raw DB column value (`provider["name"] = providerName`). When the DB contained stale lowercase values from the previous `ToLower` code path, the frontend sidebar (which uses catalog names) could not match the `addedProviders` response, causing the instance list to stay empty. Removed the override so the returned name always comes from the catalog, matching Python's `factory_info["name"]`.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -522,8 +522,8 @@ func (r *Router) Setup(engine *gin.Engine) {
 			// provider pool route group
 			provider := v1.Group("/providers")
 			{
-				provider.GET("/", r.providerHandler.ListProviders)
-				provider.PUT("/", r.providerHandler.AddProvider)
+				provider.GET("", r.providerHandler.ListProviders)
+				provider.PUT("", r.providerHandler.AddProvider)
 				provider.GET("/:provider_name", r.providerHandler.ShowProvider)
 				provider.DELETE("/:provider_name", r.providerHandler.DeleteProvider)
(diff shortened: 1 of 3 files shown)


Lynn-Inf (approved):
LGTM

PR #16901, opened 2026-07-14
feat(webdav): support custom CA certificates

### LINKED ISSUE #16635: [Feature Request]: Support custom CA certificate / SSL verification option for WebDAV connector
### Self Checks

```Markdown
`webdav_connector.py` instantiates the WebDAV client without any SSL verification override:


self.client = WebDAVClient(base_url=self.base_url, auth=(username, password))


This means `httpx`'s default certificate verification against public root CAs is always used. Self-hosted WebDAV servers (OpenMediaVault, Synology NAS, home-lab Nextcloud/Apache instances, etc.) very commonly use self-signed certificates, since they're not exposed to the public internet. In that case, the connection fails with an SSL certificate verification error, and the WebDAV data source cannot sync at all.

The only current workaround is to patch the source file directly to pass a `verify=` argument pointing to a local CA bundle. This requires shell access to the container and isn't upgrade-safe — reapplying the patch after an image update without checking for an existing `verify=` argument can result in a duplicate keyword argument and a hard SyntaxError that crashes the entire RAGFlow backend on startup.
```

Add an optional field to the WebDAV data source configuration (both in the connector class and the corresponding UI form), e.g.:

- `ca_cert_path` (string, optional): path to a custom CA certificate/bundle mounted into the container, passed as `verify=ca_cert_path` to `WebDAVClient`.
- Alternatively/additionally, a `verify_ssl` boolean toggle (default `true`) to allow disabling verification entirely for quick testing, with a clear warning in the UI about the security implications.

This would let self-hosted users configure trust for their own certificate directly through the data source settings, without touching source code.

Manually patching `webdav_connector.py` and bind-mounting the patched file in `docker-compose.yml` to survive image updates. Functional, but fragile and not discoverable for most users, and error-prone if the patch is reapplied without checking whether a `verify=` argument already exists.

```Markdown
Example: a home-lab user running RAGFlow in Docker, connecting to a WebDAV share on OpenMediaVault (or Synology, self-hosted Nextcloud, etc.) secured with a self-signed certificate. Currently the only path forward is to patch source code inside the container. Related connectors in the same codebase (S3/blob, Seafile) may have similar needs for self-hosted/on-premise deployments with internal CAs.
```

### Additional information

Happy to provide the exact traceback and reproduction steps from a real-world OMV setup if useful.

taek105:
### Summary

  - Add an optional custom CA certificate path to the WebDAV data source configuration.
  - Pass the configured certificate bundle to the WebDAV client through its `verify` option.
  - Preserve the WebDAV client's default SSL verification behavior when no custom CA is configured.
  - Expose the setting in the WebDAV data source form.
  - Add focused tests for both default verification and custom CA behavior.

  ### Why

  WebDAV servers using private or self-signed certificate authorities cannot currently be connected
  without modifying the container's global certificate configuration. Self-hosted users need a way to
  provide a mounted CA certificate bundle for an individual WebDAV data source.

  ### Testing

  - `test/unit_test/data_source/test_webdav_connector_unit.py`: 10 passed
  - `ruff check common/data_source/webdav_connector.py rag/svr/sync_data_source.py test/unit_test/
  data_source/test_webdav_connector_unit.py`
  - `npx eslint src/pages/user-setting/data-source/constant/index.tsx src/locales/en.ts --report-
  unused-disable-directives`
  - `git diff --check`

  Fixes #16635

### DIFF
--- a/common/data_source/webdav_connector.py
+++ b/common/data_source/webdav_connector.py
@@ -26,6 +26,7 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
         base_url: str,
         remote_path: str = "/",
         batch_size: int = INDEX_BATCH_SIZE,
+        ca_cert_path: str | None = None,
     ) -> None:
         """Initialize WebDAV connector
 
@@ -33,6 +34,7 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
             base_url: Base URL of the WebDAV server (e.g., "https://webdav.example.com")
             remote_path: Remote path to sync from (default: "/")
             batch_size: Number of documents per batch
+            ca_cert_path: Optional path to a custom CA certificate bundle inside the container
         """
         self.base_url = base_url.rstrip("/")
         if not remote_path:
@@ -43,6 +45,7 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
             remote_path = remote_path.rstrip("/")
         self.remote_path = remote_path
         self.batch_size = batch_size
+        self.ca_cert_path = ca_cert_path.strip() if ca_cert_path else None
         self.client: Optional[WebDAVClient] = None
         self._allow_images: bool | None = None
         self.size_threshold: int | None = BLOB_STORAGE_SIZE_THRESHOLD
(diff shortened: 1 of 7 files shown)


6ba3i (changes requested):
The approach looks good overall, but I think this needs a couple of fixes before merging.

The main issue is that `ca_cert_path` is only wired into the Python sync path. On current `main`, Test Connection builds WebDAV through `WebDAVConnector.build_connector()`, which currently drops the field, so a private-CA setup can still fail validation even though sync is configured correctly.

There’s also now a native Go WebDAV connector, and it ignores `ca_cert_path` entirely. That means the same connector config behaves differently depending on whether RAGFlow is running through Python or Go. I’d expect the Go client to load the configured CA as well, while keeping the existing default transport untouched when the option isn’t set.

One smaller thing: invalid CA paths/PEM currently end up looking like authentication failures because of the broad exception handling in `load_credentials()`. It would be better to surface those as certificate/configuration errors.

I don’t think this needs a larger refactor the existing solution is fine in principle. Just make sure the CA setting reaches all supported construction/runtime paths and add focused coverage for them.

6ba3i (approved):
LGTM

PR #16905, opened 2026-07-14
Add Go service/handler tests for API contract parity

euvre:
### Summary

This PR supplements Go tests and aligns several Go validation/response contracts with the established Python API contract (the category-2 `GO_ONLY_SKIPS` entries in `test/testcases/restful_api/conftest.py`).

- Align chat/dataset/search error messages and authorization codes (102 `CodeDataError` / 109 `CodeAuthenticationError`) with the Python contract (e.g. `` `name` can not be empty. ``, `No authorization.`).
- Make `KnowledgebaseDAO.GetByName` case-insensitive to match MySQL's default collation, so duplicate-name detection is consistent across case.
- Add Go service- and handler-level tests for:
  - chat: name validation, case-insensitive duplicate detection, `tenant_id` rejection, prompt-config defaults;
  - search: empty-name rejection (102) and unauthorized update (109), create/update round-trip;
  - dataset: embedding-model format validation, field preservation on partial update, `parser_config` merge;
  - chat-session: create (name validation, truncation, ownership) and delete (by IDs, `delete_all`, invalid ID) contracts.
- Remove the now-satisfied entries from the category-2 `GO_ONLY_SKIPS` list.


wangq8 (commented):
## Review: PR #16905 — Add Go service/handler tests for API contract parity

This is a substantial PR (+1552/-213 across ~45 files). Summary of changes and evaluation:

### Go Service & Handler Changes
1. **Chat entity — removed GORM `default:` tags**: Fields like `similarity_threshold`, `vector_similarity_weight`, `top_n`, `top_k` no longer have `default:` GORM tags. The Create method now uses `Select("*")` to persist explicit zero values. This is the correct approach — the service layer supplies defaults when omitted, and honoring explicit zeros matches the Python API contract. Good.

2. **Knowledgebase case-insensitive name lookup**: `GetByName` now uses `LOWER(name) = LOWER(?)` to match MySQL's default case-insensitive collation. **Migration adds a generated column `name_ci` + unique index** on `(tenant_id, name_ci)` for valid rows only. The migration checks for existing duplicates before creating the index. This is well-engineered.

3. **Authorization error codes standardized**: `No authorization` responses now return `CodeAuthenticationError (109)` instead of `CodeDataError (102)`, with trailing period for message consistency. Correct.

4. **Accessible4Deletion fix**: Changed from `First()` (returns `ErrRecordNotFound` → "no rows in result set" error message) to `Count()` so missing/unauthorized returns `(false, nil)`, letting callers distinguish "not authorized" from DB errors.

5. **Document list handler**: `create_time_from` / `create_time_to` filters are now applied post-query so total reflects the unfiltered count, matching the Python contract. The comment explains this clearly.

6. **Batch document metadata updates**: Changed from `ShouldBindJSON` to raw `json.Unmarshal` + manual field parsing to handle null/type variations. More robust for the API contract.

7. **StartIngestionTask**: Now reads `dataset_id` from URL param instead of request body. Response changed to `{"success_count": N}`. The old response returned the full parse result list, which was inconsistent with the Python contract.

### Python Side Changes
- Removal of unused `async_chat` import in `chat_api.py` — clean, no functional impact.
- Formatting-only changes in `begin.py`, `docling_parser.py`, `pdf_parser.py`.
- Removal of trailing newline in `init-clickhouse.sql` — cosmetic.
- `tool_decorator.py`, `tool_registry.py`, `think_log.py`, `qa.py` changes — these appear in the file list but the truncated diff doesn't show them clearly.

### Testing
Extensive new tests covering:
- Chat: name validation, case-insensitive duplicate detection, `tenant_id` rejection, prompt-config defaults
- Search: empty-name rejection (102), unauthorized update (109), create/update round-trip
- Dataset: embedding-model format validation, field preservation on partial update, `parser_config` merge
- Chat session: create (name validation, truncation, ownership) and delete (by IDs, `delete_all`, invalid ID)
The tests use in-memory SQLite with `TranslateError: true` for proper error handling — solid approach.

### Concerns
- **The diff is truncated for me** so I couldn't review every changed file. In particular the `agent/component/begin.py`, `rag/advanced_rag/` changes and `rag/llm/tool_decorator.py` appear in the file list but their diffs weren't fully visible. Please ensure those are reviewed separately if they carry logic changes.
- **StartIngestionTask response format change** (from full result list to `{"success_count": N}`) is a breaking API change for any client consuming the old response. If this aligns with the Python API contract it's fine, but worth noting.

### Verdict
LGTM overall — well-structured, well-tested, and the contract alignment fixes are necessary for API consistency. The migration is handled safely with duplicate detection and conditional index creation. The approach of removing GORM default tags and using `Select("*")` is the right fix for zero-value handling.

wangq8 (commented):
## PR #16905 – Add Go service/handler tests for API contract parity

### Summary

This PR brings the Go backend into line with the established Python API contract by:

1. **Aligning response codes & messages** – Changing `CodeDataError` (102) → `CodeAuthenticationError` (109) for auth failures, and standardizing error message punctuation (trailing periods) to match the Python contract.
2. **Fixing numeric field defaults** – Removing GORM `default:` tags from `Chat` entity numeric fields (`similarity_threshold`, `vector_similarity_weight`, `top_n`, `top_k`) so that explicitly-provided zero values are persisted instead of silently replaced by column defaults. Paired with `DB.Select("*").Create(...)`.
3. **Case-insensitive name handling** – `KnowledgebaseDAO.GetByName` now uses `LOWER(name)=LOWER(?)`, and a migration adds a `name_ci` generated column + unique index for concurrent-safe duplicate detection.
4. **Robust JSON parsing** – Dataset/handler now uses raw body + `json.Unmarshal` instead of `ShouldBindJSON` to accurately detect `parser_config` key presence. Metadata batch update parsing refactored into explicit helper functions.
5. **API contract fixes** – `StartIngestionTask` request field renamed `documents` → `document_ids`, response changed to `{"success_count": N}`. Auth error codes fixed in document, stop-parse, and chat-session handlers.
6. **Extensive tests** – ~1,000+ lines of new Go tests covering chat CRUD validation, search contracts, dataset creation, chat-session create/delete, and handler-level auth/ownership checks.

### Strengths

- **Contract discipline** – The systematic mapping of error codes (102 vs 109) and message punctuation to the Python contract is exactly right. This is what API parity requires.
- **Zero-value edge case** – The `Chat` entity GORM tag removal + `Select("*")` in `Create` is a well-understood GORM footgun, and the fix is correct.
- **Migration safety** – The `migrateKnowledgebaseNameUnique` function checks for existing duplicate names *before* adding the index, failing with a clear message. The virtual generated column approach (`NULL` for soft-deleted rows) is thoughtful.
- **Test quality** – Tests are well-structured with proper setup/teardown (`t.Cleanup`), use in-memory SQLite, and test both success paths and edge cases (empty name, duplicate name, non-owner, long name truncation). The fake-store approach for chat-session tests is clean and fast.

### Concerns / Suggested Improvements

**1. Breaking change: `StartIngestionTask` request shape change** (L1435–1455 in `internal/handler/document.go`)
The field was renamed from `"documents"` to `"document_ids"` and is now `binding:"required"`. Any existing client (Python tests, frontend, CLI scripts) sending the old field name will get a 400 error. This should be called out explicitly in the PR description. If backward compatibility is needed, consider accepting both field names.

**2. Breaking change: `StartIngestionTask` response shape**
Was returning the raw parse result array; now returns `{"success_count": N}`. This is a visible API contract change that callers must adapt to. Same note as above.

**3. `LOWER(name)=LOWER(?)` may harm query performance**
On large knowledgebase tables, wrapping the column in `LOWER()` prevents index usage for this query. The `name_ci` generated column + unique index helps for uniqueness enforcement, but `GetByName()` is a read path that runs on every dataset access. Consider adding a regular index on `(tenant_id, name_ci)` if this is a hot path.

**4. MySQL-specific migration**
The `ALTER TABLE ... ADD COLUMN ... GENERATED ALWAYS AS ... VIRTUAL` and `INFORMATION_SCHEMA` queries are MySQL-specific. If the project officially supports other databases (PostgreSQL, SQLite for tests), this migration will fail silently or require separate paths. Currently the in-memory SQLite test setup doesn't run migrations, so this isn't immediately broken, but worth noting.

**5. `Accessible4Deletion` (search) – error masking risk**
Changed from `First` to `Count`. If the DB is unreachable, `Count` still returns an error, so that's fine. But if `Count` returns 0 due to a filter mismatch without an error, the caller treats it as "not authorized" rather than "not found". The comment says this is intentional, but callers lose the ability to distinguish "search does not exist" from "search exists but you don't own it." Consider whether the Python contract truly treats these the same.

**6. `StopParseDocuments` error code change (109 → 102)**
L1523–1528 changed `CodeAuthenticationError` → `CodeDataError` for the "You don't own the dataset" message. Is this consistent with the Python contract? The other handlers use 109 for ownership failures. If Python uses 102 for this specific endpoint, fine — but it deviates from the pattern established in the rest of this PR.

**7. Metadata batch update – validation edge cases**
The new `parseMetadataSelector` / `parseMetadataUpdates` / `parseMetadataDeletes` helpers do basic type checks but use type assertions (`id.(string)`) without a `ok` guard. If a `document_ids` element is a number instead of a string, this will panic at runtime. Consider adding safe type coercion or explicit error returns.

### Verdict

**Solid work.** This PR systematically brings Go closer to the Python contract and adds much-needed test coverage. The breaking changes and a couple of edge-case panics are worth addressing before merging, but overall this is a high-quality PR.

**Recommendation:** Address the type-assertion panics in metadata batch parsing (concern #7), add a note about the API breaking changes in the PR description (concerns #1–2), and double-check the `StopParseDocuments` error code change (#6). The rest is ready to land.

PR #16916, opened 2026-07-14
fix(rag/utils): ensure_v1 must recognize v1beta/v1alpha-style version suffixes

chuenchen309:
### Summary

`ensure_v1()` checks whether a URL's path already contains a versioned segment before appending `/v1`, but its regex required the version segment to be followed immediately by `/` or end-of-string (`/v\d+(?:/|$)`). Real OpenAI-compatible endpoints commonly use a version suffix with a letter directly after the digits, not just a bare `v{digit}` segment — most notably **Google's Gemini API**, whose OpenAI-compatibility layer lives at `https://generativelanguage.googleapis.com/v1beta/openai/`.

`ensure_v1` is used across `rag/llm/{chat_model,embedding_model,cv_model,tts_model,sequence2txt_model}.py` (30+ call sites) to normalize the user-configured `base_url` for any "OpenAI API compatible" provider before handing it to the `openai.OpenAI` client, so this was silently mangling Gemini's own compat endpoint into `.../v1beta/openai/v1` — a URL that doesn't exist — breaking every request with no clear indication of why.

```python
>>> ensure_v1("https://generativelanguage.googleapis.com/v1beta/openai/")
'https://generativelanguage.googleapis.com/v1beta/openai/v1'   # before this fix - wrong
```

(For context, `v1alpha`/`v1beta`-style suffixes aren't a one-off — this same pattern already appears elsewhere in this codebase, e.g. `deepdoc/parser/docling_parser.py`'s `/v1alpha/convert/source` endpoints.)

### Fix

Instead of requiring the whole path segment to equal `v{digit}`, check whether each path segment *starts with* `v{digit}` — so `/v1beta`, `/v1alpha1`, etc. are correctly recognized as already versioned, while a segment like `/vendors` (starts with "v" but no digit right after) still correctly gets `/v1` appended.

### Testing

Added `test/unit_test/rag/utils/test_url_utils.py` covering the existing docstring examples plus the new Gemini `v1beta` case and a regression guard for the `vendors`-style false-positive. Confirmed via `git stash` that the 3 new `v1beta`/`v1alpha1` cases fail against the pre-fix code and pass after; the pre-existing cases pass unchanged.

Ran with `pytest` + `ruff check` + `ruff format --check` — all clean on the changed files. (Note: I could not run the full `test/unit_test` suite locally since my sandbox's outbound network couldn't complete the NLTK `wordnet` corpus download that the shared `conftest.py` fetches on first run — unrelated to this change, which touches only `rag/utils/url_utils.py`.)

### AI disclosure

Developed with AI assistance (Claude Code), which located the gap by tracing where `ensure_v1` is called and cross-referencing real-world OpenAI-compatible endpoint URL shapes. I reviewed the diff, independently verified the bug and the fix by executing both against the actual function (including confirming the real Gemini `v1beta` compat URL is what triggers it), and ran the tests/linters myself before opening this PR.

### DIFF
--- a/rag/utils/url_utils.py
+++ b/rag/utils/url_utils.py
@@ -20,9 +20,10 @@ from urllib.parse import urlparse, urlunparse
 def ensure_v1(url: str) -> str:
     """Ensure the URL ends with a versioned path segment like ``/v1``.
 
-    If the path already contains a segment matching ``v{digit}`` (e.g. ``/v1``,
-    ``/v2``, ``/v3``), the URL is returned unchanged.  Otherwise the base host
-    is kept and ``/v1`` is appended.
+    If the path already contains a segment starting with ``v{digit}`` (e.g.
+    ``/v1``, ``/v2``, ``/v3``, ``/v1beta``, ``/v1alpha1``), the URL is
+    returned unchanged.  Otherwise the base host is kept and ``/v1`` is
+    appended.
 
     Examples::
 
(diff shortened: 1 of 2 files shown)


Lynn-Inf (approved):
LGTM

PR #16918, opened 2026-07-14
fix(web): show pending placeholder instantly and pin assistant toolbar open

maozhuey:
## Summary

Three commits land on top of each other:

1. A user-facing chat-page fix so the AI bubble shows up the moment the user hits send, and the assistant toolbar is always visible (no hover required).
2. A Docker Compose change so forks build the image from local source instead of silently pulling the upstream prebuilt image.
3. A small README cleanup that addresses two follow-up items CodeRabbitAI raised in the review of commit 2 (a misleading "--pull" comment in the build-from-source note, and a markdownlint MD028 blank line in the NOTE callout).

### 1. `fix(web): show pending placeholder instantly and pin assistant toolbar open`

**Problems observed on the chat page:**

- The "AI is thinking" bubble did not appear immediately after the user hit send. `derivedMessages` already carries a pending assistant row (`content=''`) added by `addNewestQuestion`, but when the backend is slow or the first SSE event takes >1s, the user sees nothing for an uncomfortably long time.
- The assistant message toolbar (copy / read / like / dislike / prompt) was hidden until the user hovered the message. It is now permanently visible, matching the rest of the chat UI and being more discoverable on touch devices.

**Implementation:**

- `web/src/components/message-item/group-button.tsx`: drop the `opacity-0 group-hover:opacity-100` classes on the assistant toolbar, add an `isPendingPlaceholder` prop, return `null` when set.
- `web/src/components/message-item/index.tsx`: thread the new prop through `IProps` and pass it to `AssistantGroupButton`.
- `web/src/pages/next-chats/chat/chat-box/single-chat-box.tsx`: render an optimistic `MessageItem` placeholder whenever `sendLoading=true` and the last row is not an assistant. The placeholder is replaced by the real streaming message on the first backend event.

**Follow-up fix addressing the issue raised by JinHai-CN in the review of the previous attempt (#16890):**

The optimistic placeholder carries a synthetic id (`"__optimistic_assistant_placeholder__"`) that the backend has no record of. If a user clicked like/dislike on the placeholder, the client would POST that synthetic id to the feedback API and either 404 or, worse, collide with a real message under a hash collision. `AssistantGroupButton` now accepts an explicit `isPendingPlaceholder` flag and short-circuits its render to `null`. This replaces a previous attempt that relied on `index === 0`, which is brittle because the placeholder's index is `derivedMessages.length` (any non-zero number).

### 2. `feat(docker): opt-in local build path for ragflow-cpu/gpu`

**Problem:** prior to this change, contributors running a fork would `docker compose up` and silently fall back to pulling `${RAGFLOW_IMAGE}` (the upstream prebuilt image) — their local changes to `web/`, `api/`, or `rag/` never landed in the running container because compose skipped any build step.

**Implementation:**

- `docker/docker-compose.yml`: add a `build:` section to `ragflow-cpu` and `ragflow-gpu`, with `context: ..` (repository root) and `dockerfile: Dockerfile`. The `image: ${RAGFLOW_IMAGE}` field is kept so the resulting image is tagged exactly as before — existing `.env` files keep working without changes.
- `README.md` / `README_zh.md`: add a short note in the "Start up the server" section explaining the build-first workflow and the `docker compose pull` override for users who only want the upstream image.

**Why this matters for Apple Silicon / Linux arm64 hosts:**

The `Dockerfile` already selects every binary dep by host architecture. With `build:` declared in compose, `docker compose build` reads `uname -m` at build time and picks the right artifact, so the produced `ragflow-cpu` / `ragflow-gpu` image runs **natively** — no QEMU emulation, no Rosetta — on Apple Silicon (M1/M2/M3/M4) and Linux `aarch64` hosts, as well as on x86_64. The architecture-switching branches that already exist in the Dockerfile are:

| What it picks | x86_64 branch | arm64 / aarch64 branch |
| --- | --- | --- |
| `uv` Python installer | `uv-x86_64-unknown-linux-gnu.tar.gz` | `uv-aarch64-unknown-linux-gnu.tar.gz` |
| `stagehand-server-v3` (Browser component) | `stagehand-server-v3-linux-x64` | `stagehand-server-v3-linux-arm64` |
| MSSQL ODBC driver (`apt install`) | `msodbcsql17` | `msodbcsql18` |
| `libssl1.1` (`dpkg -i`) | `libssl1.1_1.1.1f-1ubuntu2_amd64.deb` | `libssl1.1_1.1.1f-1ubuntu2_arm64.deb` |

This change does **not** add any of those architecture branches; they were already in the file. What this PR does is make sure the build actually runs on fork clones — without the `build:` block in compose, a fork on an arm64 host would never execute those branches and would instead fall back to `docker pull`ing the upstream x86 prebuilt image, which silently runs under QEMU. With this PR, a fork on an arm64 host goes through the local build and lands on the native arm64 path above.

**Scope clarification (important):** the ARM64 coverage above applies to the `ragflow-cpu` / `ragflow-gpu` services that this PR builds locally. Auxiliary services in `docker/docker-compose.yml` (`mysql`, `es01`, `infinity`, etc.) still reference upstream images and rely on those images' own ARM64 support — this PR does not change that.

### 3. `docs(readme): tighten build-from-source note in start-server section`

CodeRabbitAI flagged two follow-up items in commit 2's README:

- The NOTE callout ended with three blank lines before the next `> [!CAUTION]` block, tripping markdownlint MD028 (blank line inside blockquote). Collapse to a single `>` separator.
- The "Skip build and pull the prebuilt image instead" comment was misleading: the following command is `docker compose pull` (the standalone pull subcommand), not a `--pull` flag. Reword the comment so it matches the actual command.

Also prepend `cd ragflow/docker` to both code blocks in the NOTE callout, because the embedded `docker compose -f docker-compose.yml` commands assume the `docker/` subdirectory on the relative path. Apply the same three fixes to `README_zh.md`.

## Compatibility

- Existing users who set `RAGFLOW_IMAGE` keep getting the same tag after build — no break.
- Users who do not run `docker compose build` explicitly still get the same behaviour as before (`docker compose up` falls back to pulling `${RAGFLOW_IMAGE}`).
- The UI change is local to the chat page; no API contract changes.

## Test plan

- `docker compose config ragflow-cpu` correctly resolves both `build:` and `image:`.
- `docker compose build --dry-run ragflow-cpu` simulates the build and tags the image as `${RAGFLOW_IMAGE}`.
- `pnpm build` succeeds; the compiled bundle contains `__optimistic_assistant_placeholder__` and `isPendingPlaceholder`.
- The container served from `/ragflow/web/dist` no longer renders the copy / like / dislike toolbar on the placeholder row.
- Manual browser test on the chat page confirms both the immediate thinking bubble and the always-visible toolbar.

## Related

- #16861 — "Feat: If the interval between two outputs exceeds 600ms, a loading state is displayed at the end." That PR uses a `use-loading-pause` hook to detect pauses during streaming output, while this PR uses an optimistic placeholder bubble for the click-to-first-token gap. The two target different phases of the user-perceived latency and are complementary, not conflicting. Happy to discuss coordination in review.

## Co-author

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


maozhuey:
Updated the PR description to make the Apple Silicon / arm64 benefit explicit. The Dockerfile already contains four architecture-switching branches (`uv`, `stagehand-server-v3`, MSSQL ODBC driver, `libssl1.1`), but the previous version of the description buried that fact in a one-liner. With `build:` declared in compose, a fork on an arm64 host actually goes through those branches and lands on the native arm64 path instead of silently falling back to the upstream x86 prebuilt image (which would otherwise run under QEMU).

Also clarified the scope: ARM64 coverage from this PR applies only to the `ragflow-cpu` / `ragflow-gpu` services that we now build locally. Auxiliary services (`mysql`, `es01`, `infinity`, etc.) still reference upstream images and depend on those images' own ARM64 support — this PR does not change that.

dcc123456 (approved):

JinHai-CN:
@maozhuey Would you please fix the conflicts?

docker/docker-compose.yml
```diff
@@ -24,6 +24,15 @@ services:
...
       - cpu
+    # Business: opt-in local build path. When a user clones this repository
+    # and edits anything under web/, api/ or rag/, `docker compose build`
+    # will produce a fresh image from the repository-root Dockerfile and
+    # tag it with ${RAGFLOW_IMAGE}. Users who only want the prebuilt
+    # upstream image can keep using `docker compose pull` instead, which
+    # pulls ${RAGFLOW_IMAGE} and overrides the build output.
+    build:
+      context: ..
+      dockerfile: Dockerfile
```
myf-bee, line 35: Not recommended to add build here. Refer to the "Build Your Own Image" and "Development Help" sections for code changes.

myf-bee (commented):
Not recommended  modify the docker-compose.yml and README.md

web/src/components/message-item/group-button.tsx
```diff
@@ -54,12 +74,13 @@ export const AssistantGroupButton = ({
...
   }, [onFeedbackOk]);
 
+  if (isPendingPlaceholder) {
+    return null;
+  }
+
   return (
     <>
-      <div
-        className="flex gap-1 opacity-0 transition-opacity group-hover:opacity-100"
```
dcc123456, line 60: This does not conform to the design.

PR #16925, opened 2026-07-15
fix(agent): allow deleting edges inside iteration on canvas

huang-aoqin:
## Summary
- Fix agent canvas: edges inside Iteration/Loop can be deleted via the edge `×` button, Delete/Backspace, without deleting surrounding nodes.
- Let iteration/loop group nodes pass pointer events through (CSS), keep interactive header/handles/resize, and raise edge-label z-index so delete hits the button.
- Preserve selected edges in `onBeforeDelete` while still rejecting edges tied to protected nodes (e.g. Begin); add focused unit tests.

## Test plan
- [ ] Inside Iteration: hover or select an edge between child nodes → click `×` deletes only that edge
- [ ] Inside Iteration: select an edge → Delete/Backspace deletes only that edge (nodes remain)
- [ ] Same checks inside Loop
- [ ] Iteration/Loop header still selects/deletes the group; handles still connect; resize still works
- [ ] Begin node remains undeletable; its edges are not removed when Begin is rejected
- [ ] Agent→tool edge cleanup behavior unchanged


cike8899 (commented):
The bug: When a user selects both the Begin node and another node connected to it, then presses Delete, the edge between them is not deleted — but the node is. This leaves a broken "orphan" edge pointing to a node that no longer exists.

  Why it happens: The new code skips any edge that touches a rejected node (like Begin). But if the other end of that edge is being deleted, the edge must be deleted too.
  
  One-line fix: Remove the allowEdge check from the second loop in use-before-delete.tsx — edges touching deleted nodes should always be deleted.

@huang-aoqin

huang-aoqin:
> The bug: When a user selects both the Begin node and another node connected to it, then presses Delete, the edge between them is not deleted — but the node is. This leaves a broken "orphan" edge pointing to a node that no longer exists.
> 
> Why it happens: The new code skips any edge that touches a rejected node (like Begin). But if the other end of that edge is being deleted, the edge must be deleted too.
> 
> One-line fix: Remove the allowEdge check from the second loop in use-before-delete.tsx — edges touching deleted nodes should always be deleted.
> 
> @huang-aoqin

@cike8899 done!

cike8899 (approved):

PR #16928, opened 2026-07-15
fix(agent): restore DataOperations input form

LauraGPT:
## Summary

- return the input form collected by `DataOperationsComponent.GetInputForm`
- add a regression test covering system variables, component outputs, and duplicate references

## Root cause

`DataOperationsComponent.GetInputForm` builds its result map from configured query placeholders but returns a new empty map instead, so callers never receive the discovered input fields.

#16927 has now removed the duplicate helper that briefly broke the Go build. This PR is rebased on that fix and is intentionally limited to the remaining `GetInputForm` behavior bug and its regression coverage.

## Validation

- `go test ./internal/agent/component -count=1`
- `go vet ./internal/agent/component`
- `./build.sh --go`

### DIFF
--- a/internal/agent/component/data_operations.go
+++ b/internal/agent/component/data_operations.go
@@ -279,37 +279,6 @@ func (d *DataOperationsComponent) GetInputForm() map[string]any {
 			}
 		}
 	}
-	return map[string]any{}
-}
-
-// TODO duplicate with string_transform.go, make it to a common method later --Haruko386
-// I can't do this since that PR not merged
-func getInputElementsFromText(txt string) map[string]map[string]any {
-	res := make(map[string]map[string]any)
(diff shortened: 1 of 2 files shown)


LauraGPT:
@yuzhichang could you please add the `ci` label when convenient? This fixes the current `main` Go compile regression that is causing both Go jobs on #16844 to fail. The exact `./build.sh --go` build and the component test suite pass locally.

yuzhichang:
Replaced by #16927

LauraGPT:
Thanks. #16927 fixes the duplicate-symbol build. I rebased on it and opened #16930 for the distinct remaining bug on current `main`: `DataOperationsComponent.GetInputForm` still builds `res` but returns a new empty map. The follow-up is limited to that one-line behavior fix and its regression test.

PR #16930, opened 2026-07-15
fix(agent): return DataOperations input form

LauraGPT:
## Summary

- return the input form collected by `DataOperationsComponent.GetInputForm`
- add a regression test covering system variables, component outputs, and duplicate references

## Root cause

`DataOperationsComponent.GetInputForm` builds its result map from configured query placeholders but returns a new empty map instead, so callers never receive the discovered input fields.

#16927 has now removed the duplicate helper that briefly broke the Go build. This PR is rebased on that fix and is intentionally limited to the remaining `GetInputForm` behavior bug and its regression coverage.

## Validation

- `go test ./internal/agent/component -count=1`
- `go vet ./internal/agent/component`
- `./build.sh --go`

### DIFF
--- a/internal/agent/component/data_operations.go
+++ b/internal/agent/component/data_operations.go
@@ -279,7 +279,7 @@ func (d *DataOperationsComponent) GetInputForm() map[string]any {
 			}
 		}
 	}
-	return map[string]any{}
+	return res
 }
 
 // Outputs returns the transformed payload.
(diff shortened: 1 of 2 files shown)


Haruko386 (approved):
@LauraGPT 
LGTM , thank you for your contribution
It's my bad to forget to fix this :(

PR #16933, opened 2026-07-15
refa: optimize knowledge compilation concurrency

buua436:

Improve concurrency in the knowledge compilation pipeline:

- Run Compile LLM requests concurrently while preserving ordered commits.
- Run merge flush tasks concurrently while keeping ES writes ordered.
- Improve concurrency for local deduplication, chain validation, and ES deduplication.
- Remove temporary debugging instrumentation and unused timing variables.

### Type of change

### DIFF
--- a/rag/advanced_rag/knowlege_compile/runner.py
+++ b/rag/advanced_rag/knowlege_compile/runner.py
@@ -43,19 +43,25 @@ from api.db.services.compilation_template_group_service import (
 from api.db.services.llm_service import LLMBundle
 from common.exceptions import TaskCanceledException
 from rag.advanced_rag.knowlege_compile.structure import (
-    CHAIN_KINDS,
+    LLMCallPool,
     compile_structure_from_text,
     merge_compiled_structures,
-    validate_and_correct_chain,
 )
 
 
 # ----- tunables ------------------------------------------------------
 # Bound how many source chunks are handed to a single
-# ``compile_structure_from_text`` invocation. The call fans them out
-# across max_workers internally, so a moderate window keeps memory +
-# LLM-context pressure predictable for long docs.
+# ``compile_structure_from_text`` invocation.
 DOC_STRUCTURE_COMPILE_BATCH_CHUNKS = 4
+
+# Bound the number of batch/template extraction calls in flight. Results are
+# committed in submission order so accumulator updates and merge flushes stay
+# deterministic while the LLM calls run concurrently.
+DOC_STRUCTURE_COMPILE_MAX_IN_FLIGHT = 15
+
+# Total task-scoped Chat LLM capacity shared by compile, chain validation and
+# merge decisions. A request waits in the priority queue when all slots are busy.
+DOC_STRUCTURE_LLM_POOL_SIZE = 20
 
 # Bound how many compiled ES-ready docs may accumulate before we flush
 # them through ``merge_compiled_structures``. The merger does pairwise
@@ -63,10 +69,7 @@ DOC_STRUCTURE_COMPILE_BATCH_CHUNKS = 4
 # cap the per-flush set to keep the local-dedup buckets tractable.
 DOC_STRUCTURE_MERGE_MAX_DOCS = 512
 
-# Hard wall on the chain-validator LLM correction step. ``list`` and
-# ``timeline`` kinds run this just before each merge flush; anything
-# longer than this is treated as a blocked LLM and the uncorrected
-# docs are flushed instead.
+# Hard wall on the chain-validator LLM correction step.
 STRUCTURE_CHAIN_CORRECTION_TIMEOUT_S = 120.0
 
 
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
## Review: PR #16933 — Optimize knowledge compilation concurrency

**Files:** 4 | **+1117 / -135** | **Author:** @buua436 | **State:** Open

---

### Overview

This PR significantly improves concurrency in the document knowledge compilation pipeline:
- **Priority-queued LLM pool** (`LLMCallPool`) shared across compile batches, chain correction, and merge decisions
- **Concurrent batch compilation & ordered commits** — async submission with bounded inflight tasks, sequential commit to maintain deterministic accumulator state
- **Batched ES dedup** replaces per-doc KNN+LLM with phased approach: parallel KNN → concurrent LLM decisions → grouped merges → batched embeddings & writes
- **Entity alias resolution** during local dedup — tracks canonical names and rewrites relation endpoints
- **Parallel local dedup** via cosine-connected component partitioning
- **Parallel chain correction** — splits bad edges into bounded batches for concurrent LLM correction
- **`compile_kwd` filter** — excludes compilation output rows from chunk loading queries to prevent offset-pagination drift

The architectural direction is good. However, several issues need attention before merging.

---

### 🔴 Critical: Early return in write loop drops data

**File:** `rag/advanced_rag/knowlege_compile/structure.py` _(inside `_struct_es_dedup_batch`)_

```python
for start in range(0, len(writes), _ES_DEDUP_INSERT_BATCH_SIZE):
    batch = writes[start : start + _ES_DEDUP_INSERT_BATCH_SIZE]
    ...
    return inserted, updated + existing_relation_updates  # ← inside loop body
```

The `return` is **inside** the `for` loop, so only the **first write batch** (up to 256 docs) is ever persisted. If `len(writes) > 256`, all subsequent batches are silently dropped. This is a data-loss bug.

**Fix:** Dedent `return` to be outside the `for` loop. Also the empty-batch `continue` is unreachable (`range` with positive step never yields empty slices) — can be cleaned up.

---

### 🔴 Cancellation leaves stale tickets in `LLMCallPool`

**File:** `rag/advanced_rag/knowlege_compile/structure.py` — `LLMCallPool.call()`

If a coroutine is cancelled while `await self._condition.wait()`, the ticket remains in `self._waiting` heap. This can:
1. Block future slots if the stale ticket stays at the heap front
2. Cause `self._active -= 1` in the `finally` block even though `_active` was never incremented

**Fix:** Wrap the wait in a cancellable scope and remove the ticket from `_waiting` (restoring heap invariant) on cancellation.

---

### 🔴 No-op `except BaseException: raise` in `_compile_batch`

**File:** `rag/advanced_rag/knowlege_compile/runner.py`

```python
async def _compile_batch(...):
    try:
        docs = await compile_structure_from_text(...)
        return docs
    except BaseException:
        raise
```

This bare `except BaseException: raise` is a no-op and should be removed entirely.

---

### 🟡 `_struct_entity_candidate_groups` — O(n³) index lookup

**File:** `rag/advanced_rag/knowlege_compile/structure.py`

```python
for left_offset, left in enumerate(valid):
    for right in valid[left_offset + 1 :]:
        right_offset = valid.index(right)  # O(n) linear search per pair
```

`valid.index(right)` is O(n) and called for every pair, making the connected-component construction O(n³). For large entity groups this is unnecessary overhead.

**Fix:** Use offset-based iteration instead:

```python
for left_offset, left in enumerate(valid):
    for right_offset in range(left_offset + 1, len(valid)):
        right = valid[right_offset]
        ...
```

---

### 🟡 Missing `cancel_check` inside ES dedup batch

**File:** `rag/advanced_rag/knowlege_compile/structure.py` — `_struct_es_dedup_batch`

The old per-doc code checked cancellation between **every** ES insert. The new batched equivalent checks `cancel_check` only **before** entering the batch. During a long ES dedup (parallel KNN → LLM decisions → merges → embeddings → bulk writes), cancellation is not checked at all for potentially minutes of wall time.

**Fix:** Insert `cancel_check` calls between the major stages (after KNN gathering, after LLM decisions, before merge jobs, before bulk writes).

---

### 🟡 Cross-batch chain correction conflicts

**File:** `rag/advanced_rag/knowlege_compile/structure.py` — `correct_batch`

Bad edges are partitioned into batches of `_CHAIN_CORRECTION_MAX_RELATIONS` (16). Each batch is corrected independently by the LLM. If two bad edges from different batches form a conflict chain (e.g., A→B and B→C), neither the LLM nor the post-processing sees the full connected component. This can produce a `keep_set` that still violates the strict-chain constraint, or drop more edges than necessary.

**Consider:** Partition by conflict-connected components instead of a fixed-size batch to ensure each LLM call sees the complete context.

---

### 🟡 Dead code: `_struct_es_dedup_one` is no longer called

**File:** `rag/advanced_rag/knowlege_compile/structure.py`

The function `_struct_es_dedup_one` is retained but has no callers after the switch to `_struct_es_dedup_batch`. Remove it to prevent drift.

---

### 🟢 Good practices observed

| Practice | Location |
|---|---|
| **Fail-open pattern** | Exception handlers in ES KNN, LLM decisions, chain correction gracefully degrade |
| **Two-phase ES LLM dedup** | Decision (read-only) → Merge (payload generation) — prevents write-before-read races |
| **Ordered commits via sequence numbers** | `flush_sequence` / `next_es_sequence` with `asyncio.Condition` — deterministic despite concurrent flushes |
| **`must_not: exists` for compile_kwd** | Clean approach — avoids adding a separate flag field just for filtering |
| **Component partitioning for parallel local dedup** | Cosine-connected groups allow safe concurrent dedup |
| **Entity alias → relation rewrite dependency ordering** | Entities deduped first, aliases collected, then relations rewritten & deduped |
| **Bounded everything** | Inflight tasks, LLM pool, KNN concurrency, LLM batch size, embed batch size, insert batch size — all configurable constants |

---

### Summary

| Severity | Count | Items |
|---|---|---|
| 🔴 Critical | 3 | Write-loop early return (data loss), cancellation ticket leak, no-op try/except |
| 🟡 Should fix | 4 | O(n³) lookup, missing cancel_check inside ES batch, cross-batch chain conflicts, dead code |
| 🟢 Minor/cleanup | — | Various small improvements noted above |

**Verdict:** The concurrency model is well-designed and the PR represents meaningful progress. The **early-return bug** must be fixed before merge — it's a silent data loss bug. The cancellation issues should also be addressed for robustness.

PR #16934, opened 2026-07-15
refactor: remove localStorage persistence for system model settings

euvre:
### Summary

Remove the localStorage-based persistence layer from the system model settings page. The localStorage workaround was originally added to handle the case where the Go backend hadn't been rebuilt yet and didn't return TTS/ASR/VLM defaults from the API.

Now that the backend API (`GET /api/v1/models/default`) is the source of truth for all model types (including ASR/VLM/TTS), the localStorage persistence is redundant and introduces unnecessary complexity. This PR:

- Removes the `STORAGE_KEY` constant, `loadPersistedValues`, and `savePersistedValues` helper functions
- Initializes `persistedValues` state directly from `defaultModelDictionary` instead of localStorage
- Syncs state with the API response via `useEffect` instead of localStorage
- Simplifies the resolution order to: API response → empty string (fallback)

### DIFF
--- a/web/src/pages/user-setting/setting-model/layout/system-setting.tsx
+++ b/web/src/pages/user-setting/setting-model/layout/system-setting.tsx
@@ -27,7 +27,7 @@ import {
   useSetDefaultModel,
 } from '@/hooks/use-llm-request';
 import { CircleQuestionMark } from 'lucide-react';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useMemo } from 'react';
 
 interface ModelFieldItemProps {
   id: string;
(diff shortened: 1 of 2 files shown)


web/src/pages/user-setting/setting-model/layout/system-setting.tsx
```diff
@@ -105,16 +84,15 @@ function SystemSetting() {
   const defaultModelDictionary = useFetchDefaultModelDictionary();
   const { setDefaultModel } = useSetDefaultModel();
 
-  // Local state synced with localStorage for persistence across page refreshes.
-  // This handles the case where the Go backend hasn't been rebuilt yet and
-  // doesn't return TTS/ASR/VLM defaults from the API (same pattern as ASR/VLM).
-  const [persistedValues, setPersistedValues] =
-    useState<Record<string, string>>(loadPersistedValues);
+  // Local state mirrors the API-backed defaults so the UI updates immediately
```
dcc123456: persistedValues ：This parameter should be deleted.

web/src/pages/user-setting/setting-model/layout/system-setting.tsx
```diff
@@ -139,9 +117,8 @@ function SystemSetting() {
...
 
   // Resolution order (same for ALL model types including ASR/VLM/TTS):
-  //   1. localStorage (user's latest selection, survives refresh)
-  //   2. API response (from Go backend GET /api/v1/models/default)
-  //   3. empty string (fallback)
+  //   1. API response (from Go backend GET /api/v1/models/default)
+  //   2. empty string (fallback)
   const getValue = useCallback(
     (field: string) =>
       persistedValues[field] || defaultModelDictionary[field] || '',
```
dcc123456: persistedValues[field]   ： This parameter should be deleted.

dcc123456 (approved):

PR #16943, opened 2026-07-15
fix(metadata): Infinity push-down must reject multi-valued-unsafe negative ops (≠, not in)

chuenchen309:

The Elasticsearch and Infinity metadata push-down translators are meant to be
interchangeable pre-filters that both mirror the in-memory `meta_filter`
fallback — the shared test section is even titled *"is_pushdown_supported
pre-check (same logic for both backends)"*. But the two `is_pushdown_supported`
implementations diverge:

- `common/metadata_es_filter.py` defines
  `MULTIVALUE_UNSAFE_NEGATIVE_OPS = frozenset({"≠", "not in"})` and refuses
  push-down for those operators.
- `common/metadata_infinity_filter.py` has **no such guard** and pushes them
  down.

**Why the guard exists:** `meta_fields.<key>` can hold a JSON array, and the
in-memory `meta_filter` matches a document when **any** of its values satisfies
the predicate (per-value-bucket semantics). A document whose `tag` is `[a, b]`
therefore still matches `tag ≠ a` — bucket `b` satisfies it. The Infinity
push-down emits `NOT JSON_CONTAINS(meta_fields, '$.tag', '"a"')`, which means
*"the array contains no `a` at all"*, so it **silently drops** that document.
Same divergence for `not in`. The result: `tag ≠ a` / `tag not in (...)`
under-counts results for any document that has the excluded value alongside
others, but only on the Infinity backend.

This is on a live production path:
`DocMetadataService._filter_doc_ids_by_metadata_infinity`
(`api/db/services/doc_metadata_service.py:948`) calls this exact
`is_pushdown_supported` as the sole gate before building the Infinity SQL,
mirroring the ES branch at line 880 which uses the guarded ES version.

Reproduction (both real modules, no services needed):

```python
>>> from common import metadata_es_filter as es, metadata_infinity_filter as inf
>>> f = [{"op": "≠", "key": "tag", "value": "a"}]
>>> es.is_pushdown_supported(f), inf.is_pushdown_supported(f)
(False, True)   # ES falls back to in-memory (correct); Infinity pushes down (wrong)
```

### Type of change

### Fix

Add the same `MULTIVALUE_UNSAFE_NEGATIVE_OPS` set to
`metadata_infinity_filter` and reject those operators in
`is_pushdown_supported`, so a single such filter forces the whole request to
the in-memory path — the only place the per-bucket semantics are reproduced.
`not contains` is intentionally still allowed, matching the ES backend
(`all(not contains)` == `not any(contains)`, which the push-down expresses
correctly on multi-valued fields). The `≠` / `not in` translators themselves
are unchanged — they remain correct for the in-memory-fallback path; only the
push-down eligibility gate is fixed.

### Testing

- Confirmed `is_pushdown_supported([{op}])` now returns `False` for `≠` and
  `not in` on **both** backends (previously ES=False, Infinity=True).
- Added `test_pushdown_check_rejects_multivalue_unsafe_negative_ops`, which
  asserts both backends reject these ops. Confirmed red→green: it fails against
  the pre-fix Infinity module, passes after.
- `ruff check` / `ruff format --check` clean.

### Note on overlap

Open PR #16833 also edits `test/unit_test/common/test_metadata_filter.py`, but
only **appends** at the end of the file (line 640+) and changes
`metadata_es_filter.py` / `metadata_utils.py`, not the Infinity module — no
overlap with this change.

### Disclosure

AI-assisted (Claude Code): the divergence was surfaced by an AI-assisted review
pass, but I independently reproduced it against the real modules, confirmed the
production call path, and verified the fix and test before submitting.

### DIFF
--- a/common/metadata_infinity_filter.py
+++ b/common/metadata_infinity_filter.py
@@ -55,6 +55,20 @@ _RANGE_OPS: Dict[str, str] = {
     "≤": "<=",
 }
 
+# Negative operators whose legacy per-value-bucket semantics cannot be
+# reproduced by a single push-down predicate over a multi-valued metadata
+# field. ``meta_fields.<key>`` may hold a JSON array, and the in-memory
+# ``meta_filter`` matches a doc when ANY of its values satisfies the predicate;
+# e.g. a doc whose ``tag`` is ``[a, b]`` still matches ``tag ≠ a`` (via bucket
+# ``b``). ``NOT JSON_CONTAINS(...)`` instead means "the array contains no ``a``
+# at all", which would silently drop that doc. Without a cheap way to prove a
+# field is single-valued at query time we refuse push-down for these operators
+# and let the in-memory fallback handle them. Mirrors the identically named set
+# in ``metadata_es_filter``. ``not contains`` is intentionally excluded:
+# ``all(not contains)`` equals ``not any(contains)``, which the push-down
+# expresses correctly on both single- and multi-valued fields.
+MULTIVALUE_UNSAFE_NEGATIVE_OPS: frozenset[str] = frozenset({"≠", "not in"})
+
 
 class MetaFilterTranslator:
     """Translate one user filter clause at a time into Infinity SQL filter strings."""
@@ -208,6 +222,8 @@ def is_pushdown_supported(filters: Sequence[Dict[str, Any]]) -> bool:
     for flt in filters:
         op = flt.get("op")
         if op not in SUPPORTED_OPERATORS:
+            return False
+        if op in MULTIVALUE_UNSAFE_NEGATIVE_OPS:
             return False
         if not isinstance(flt.get("key"), str) or not flt.get("key"):
             return False
--- a/test/unit_test/common/test_metadata_filter.py
+++ b/test/unit_test/common/test_metadata_filter.py
@@ -61,6 +61,20 @@ def test_pushdown_check_accepts_not_contains():
     assert is_pushdown_supported([{"key": "tag", "op": "not contains", "value": "x"}])
 
 
+@pytest.mark.parametrize("op", ["≠", "not in"])
+def test_pushdown_check_rejects_multivalue_unsafe_negative_ops(op):
+    # These negative operators cannot be expressed as a single push-down
+    # predicate over a multi-valued metadata field without diverging from the
+    # in-memory meta_filter's per-value-bucket semantics, so both backends must
+    # refuse push-down and fall back to in-memory evaluation. The ES backend
+    # already does this; the Infinity backend must match it.
+    from common.metadata_es_filter import is_pushdown_supported as es_is_pushdown_supported
+
+    flt = [{"key": "tag", "op": op, "value": "a"}]
+    assert not is_pushdown_supported(flt)
+    assert not es_is_pushdown_supported(flt)
+
+
 # ---------------------------------------------------------------------------
 # Shared: plan_pushdown (same logic for both backends)
 # ---------------------------------------------------------------------------


zeweihan:
Excellent root-cause analysis — the per-value-bucket quantifier shift is the crux and you nailed exactly why `not contains` is safe while `≠`/`not in` aren't. Adding a regulated-domain angle:

**1. Silent under-counting on negation is the worst-case asymmetry for audit scope.** A privilege filter `tag ≠ "privileged"` *should* include a document tagged `[privileged, public]` — bucket `public` satisfies `≠privileged`. But Infinity's `NOT JSON_CONTAINS` asks "does the array contain no `privileged` at all?" → excludes the document. Same query, same corpus, **different answer set depending on backend**, zero error signal. For a production-scope or privilege-review query, "what came back" IS the operative answer — a backend-specific under-count means the scope of what was reviewed is materially different from what should have been reviewed, and nothing in the result set flags it. Same plausible-but-wrong family as the silent-loss class: content correctly present, correctly matches the reference predicate, silently dropped by the push-down shortcut.

**2. The asymmetry is quantifier-level, not backend-level.** Positive ops (`=`, `in`) align because `JSON_CONTAINS` is `∃v(v=a)` which matches per-bucket semantics. Negation (`≠`, `not in`) breaks because `NOT JSON_CONTAINS` is `¬∃v(v=a)` = "no value is a", when the correct per-bucket negation is `∀v(v≠a)` = "every value isn't a". For multi-valued fields these are not equivalent: `∃v(v≠a)` (the in-memory contract) holds whenever *any* value isn't `a`. The push-down translation inverts the quantifier from existential to universal — the same predicate string, structurally different meaning. This is why the fix belongs at the push-down eligibility gate (reject, fall back to the only faithful path) rather than attempting a corrected negation translation, which would require a fundamentally different SQL shape.

**3. `not contains` being safe is the subtle confirmation of the model.** `all(not contains) == not any(contains)` — for containment, universal and existential negation coincide. For membership-with-exclusion (`≠`, `not in`) they diverge. This distinction is load-bearing for anyone reasoning about *which* operators are push-down-safe: the safe set isn't "all except negation" but "operators where the existential↔universal shift is a no-op."

**4. Companion — cross-backend parity assertion + evaluation-path stamp.** The shared test section titled *"same logic for both backends"* shows the *intent* was parity, but the divergence survived because there was no runtime parity assertion on a multi-valued test corpus. A `metadata_filter_parity_check` invariant — same filter expression through both backends *and* the in-memory reference, assert identical result sets — would have caught this class by construction. Second: a `filter_evaluation_backend` stamp per result set (`es_pushdown` / `infinity_pushdown` / `in_memory_fallback`) makes "was this query shortcut-evaluated?" queryable, turning the silent backend-specific divergence into an inspectable provenance signal — same philosophy as `text_coverage` / `formula_render_status`.

This connects to #16832 (ES `.keyword` routing silently fatal for date/numeric fields): both are cases where the push-down layer honors *most* of the filter contract but silently diverges on a specific operator/field-type combination, and the caller has no signal that the shortcut produced a different answer than the reference path would have.

chuenchen309:
Thanks @zeweihan — the audit-scope framing is spot on. Silent under-counting on `≠`/`not in` is the dangerous mode precisely because it fails quiet: a doc tagged `[privileged, public]` *should* satisfy `tag ≠ privileged` (bucket `public` matches), but `NOT JSON_CONTAINS` drops it. That's exactly why the guard routes those ops to the in-memory fallback (per-value-bucket semantics) instead of pushing down, while `not contains` stays pushed down since `all(not contains) == not any(contains)` holds. Appreciate the careful review.

JinHai-CN:
@qinling0210 Would you please review this PR?

qinling0210 (approved):
LGTM

PR #16951, opened 2026-07-15
Fix: verify model

Lynn-Inf:
### Summary

Check asr and vlm availablity.


wangq8 (commented):
## PR Review: #16951 — "Fix: verify model"

### Overview
This PR adds VLM (vision language model) and ASR (automatic speech recognition) model verification to the API key validation flow in `provider_api_service.py`, plus a `check_available()` method on `Seq2txtModel` classes in `sequence2txt_model.py`.

Two files changed (+112/−4).

---

### What works well

**1. Pure stdlib WAV generation**
`_generate_test_wav()` uses `struct.pack` with no external dependencies — clean and portable. The WAV header format is correct for mono 16-bit PCM.

**2. Provider-specific override pattern**
`TencentCloudSeq2txt.check_available()` correctly passes raw bytes (since Tencent expects base64 encoded audio, not a file path), and `GPUStackSeq2txt.check_available()` explicitly returns "not yet implemented" rather than crashing. Good design.

**3. Proper resource cleanup**
The `Base.check_available()` uses `try/finally` with `os.unlink()` for the temp file. 

**4. Consistent integration into the if-elif chain**
The VLM and ASR branches are correctly added as `elif` clauses after the existing TTS `elif` in the verify loop — no structural breakage.

---

### Issues found

**1. Extra False value in variable initialization (minor but sloppy)**

```
chat_passed, embd_passed, rerank_passed, ocr_passed, tts_passed, asr_passed, vlm_passed = False, False, False, False, False, False, False, False
```

7 variables, 8 values on the right. The trailing False is silently ignored by Python (not a runtime error), but it indicates a copy-paste miscount and should be cleaned up.

**2. VLM verification error check could miss empty responses**

The check `if not tc and m.find("**ERROR**") >= 0` treats the model as failed only when BOTH zero tokens AND an `**ERROR**` string are present. If `mdl.describe()` returns an empty string (0 tokens) but no `**ERROR**:` marker, the check won't catch it. This is consistent with the existing embedding model pattern, but consider making the check more robust (e.g., also fail on empty/whitespace-only `m`).

**3. Loop breaks after first successful verification**

The `if any(...)` check breaks the `for llm` loop as soon as a single model type passes. This is pre-existing behavior, but it means that if the model list ordering happens to verify VLM before ASR, ASR never gets checked. Not a bug (goal is API key validation, not exhaustive per-model checks), but worth being aware of.

---

### Minor suggestions

**4. Add a comment to the WAV header struct**
The `_generate_test_wav()` binary layout would benefit from a short comment explaining the WAV format fields for future readers.

**5. GPUStackSeq2txt.check_available() return message**
Consider making the message more descriptive, e.g. "GPUStack ASR transcription check is not yet implemented — manual verification required."

---

### Summary

The PR is functionally correct and follows existing codebase patterns well. The main actionable item is the extra False literal in the initialization (fix in under 5 seconds). The VLM error check edge case is worth a follow-up.

I suggest addressing item **1** before merging, and considering a cleanup for item **2**.

PR #16954, opened 2026-07-15
fix(llm): Ollama thinking toggle, max tokens, and frequency_penalty for Agent nodes

iwasaki1112:
## Summary

This PR fixes three related bugs in `rag/llm/chat_model.py`'s `LiteLLMBase` (used for Ollama and other litellm-routed providers). Together they caused Agent nodes backed by a locally-hosted Ollama model (reproduced with `qwen3.5:9b` and `qwen3:8b`, litellm 1.82.5) to hang generating a runaway repetition loop or a garbled response, regardless of the node's "Thinking" and "Max tokens" settings.

### 1. The Agent's "Thinking" toggle has no effect for Ollama-hosted models

`_apply_model_family_policies()` special-cases model names containing `"qwen3"` and disables thinking by sending `enable_thinking` via `extra_body`. That parameter name is only understood by the Alibaba Dashscope / Tongyi-Qianwen native API. litellm's `ollama_chat` transformation does not read `extra_body.enable_thinking` at all - it only maps the standard `reasoning_effort` completion kwarg onto Ollama's native `think` request field. As a result, selecting `Disabled`/`Enabled` in the UI was silently dropped for any Ollama-hosted Qwen3 model.

**Fix:** add an Ollama-specific branch that sets `reasoning_effort` (`"medium"` when thinking should be enabled, `"none"` when disabled) instead of `enable_thinking`.

### 2. The Agent's "Max tokens" setting is silently discarded for every litellm-routed provider

`LiteLLMBase._clean_conf()` unconditionally did `gen_conf.pop("max_tokens", None)` instead of renaming the key to the OpenAI-style `max_completion_tokens` that litellm (and `LITELLM_ALLOWED_GEN_CONF_KEYS`) actually expects. The Agent/LLM component (`agent/component/llm.py` `LLMParam.gen_conf()`) only ever populates `max_tokens`, so the configured token limit was dropped for **every** provider routed through `LiteLLMBase` - Ollama, Anthropic, Gemini, Groq, DeepSeek, and others - leaving generation without the Agent-configured per-request output-length cap.

**Fix:** rename `max_tokens` to `max_completion_tokens` (via `setdefault`, so an already-present `max_completion_tokens` from a caller isn't overwritten) instead of discarding it.

### 3. `frequency_penalty` corrupts the response for Ollama models (the actual trigger for the reported "infinite loop")

litellm maps the top-level `frequency_penalty` completion kwarg onto Ollama's native `repeat_penalty` option for the `ollama_chat` provider. The value it puts on the wire is identical whether `frequency_penalty` is passed as a top-level kwarg or via `extra_body.repeat_penalty` (confirmed via litellm's own debug logging of the outgoing POST body) - but passing it as the **top-level kwarg combined with `reasoning_effort`** (which the Ollama branch above always sets) reproducibly corrupts the response. With a fixed random seed swept over 8 values against a live Ollama instance, the top-level path failed **8/8** (the response became the conversation's system prompt echoed back verbatim); routing the *same* value through `extra_body.repeat_penalty` succeeded **8/8**. The failure is therefore in litellm's own request/response handling triggered by seeing `frequency_penalty` as a recognized top-level param alongside `reasoning_effort`, not in the resulting wire payload.

This is what was actually driving the reported "infinite loop": RAGFlow's default Agent "Balanced" preset sets `frequency_penalty: 0.3`, and with that present (combined with thinking being toggled at all) the model would deterministically produce garbage instead of a real answer - which, prior to fix #2, then had no token limit to stop it from running away.

**Fix:** route `frequency_penalty` through `extra_body.repeat_penalty` instead of the top-level kwarg (mirroring how the existing Moonshot/ZHIPU-AI branches already route provider-specific fields through `extra_body`), **and** rescale the value: OpenAI-style `frequency_penalty` is centered on `0.0` while Ollama's `repeat_penalty` is centered on `1.0` (Ollama's own default is `1.1`), so a raw 1:1 passthrough would send a value that *encourages* repetition instead of discouraging it. Also fixed a latent bug this surfaced: `LiteLLMBase.async_chat()` combined gen_conf and kwargs via a plain `{**gen_conf, **kwargs}` spread, which would let one side's `extra_body` silently clobber the other's instead of merging them; added `_merge_gen_conf_and_kwargs()` to merge `extra_body` explicitly.

## Testing

**Unit tests** (added in this PR, `test/unit_test/rag/llm/test_chat_model_thinking_policy.py` and `test_clean_conf_whitelist.py`): 12 new tests covering the reasoning_effort mapping, the frequency_penalty rerouting/rescale (for both a qwen3 model and a non-qwen3 Ollama model, since the rerouting is gated on `provider == Ollama`, not the qwen3 model-name check), the `max_tokens` rename and its `setdefault` precedence, and `_merge_gen_conf_and_kwargs()`'s merge behaviour. All 118 tests under `test/unit_test/rag/llm/` pass (`test/unit_test/rag/llm/conftest.py`'s `SupportedLiteLLMProvider` stub also needed syncing with the real enum - it was missing most members, including `Ollama` and `Anthropic`, which was failing every test in the directory that touches `_apply_model_family_policies()` independent of this PR).

**Live verification** (RAGFlow v0.26.4, Ollama 0.32.0, models `qwen3.5:9b` and `qwen3:8b`, litellm 1.82.5), driven end-to-end through the real RAGFlow Agent-canvas UI, not just direct API calls:

- **Fix #1:** confirmed via server-side logging that real Agent-node chat requests sent `reasoning_effort` matching the UI's Thinking setting, with no `<think>` trace / "Thinking..." UI indicator.
- **Fix #2:** before this fix, an Agent node hitting the repetition failure mode ran to **10,000+ tokens** and kept going (`n_decoded` climbing indefinitely per Ollama's own request logs) until the Ollama process was killed. After the fix, three separate test conversations each deliberately triggering it all stopped at **exactly the configured 4096-token limit**, and the UI correctly showed "The answer is truncated by your chosen LLM due to its limitation on context length."
- **Fix #3:** isolated with a fixed-seed, controlled A/B (documented above) that pinned the failure to litellm's top-level `frequency_penalty` handling specifically. End-to-end in the actual UI, using an Agent node on the default "Balanced" preset (`frequency_penalty: 0.3`) with thinking disabled: **5 consecutive chat turns** (greetings, thanks, a factual weather question, a self-introduction request) all produced a single clean, correctly-targeted, non-repeating response. Re-verified after the rescale correction with a fixed-seed sweep (8 seeds) using the real Agent system prompt and the Balanced preset's full parameter set: **8/8 clean**.

## Notes

- Fix #1 only changes behavior for `provider == SupportedLiteLLMProvider.Ollama` combined with a model name containing `"qwen3"`; other providers are unaffected.
- Fix #2 affects the `max_tokens` -> `max_completion_tokens` renaming for all `LiteLLMBase`-routed providers, which is the intended behavior given `LITELLM_ALLOWED_GEN_CONF_KEYS` already expects `max_completion_tokens` and not `max_tokens`.
- Fix #3's `frequency_penalty` rerouting applies to any Ollama-hosted model (gated on `provider == Ollama`, independent of the qwen3-specific thinking logic above it), since the OpenAI/Ollama scale mismatch isn't qwen3-specific; other providers' `frequency_penalty` handling is untouched.
- This PR does not change how thinking is enabled/disabled for any provider besides Ollama, nor does it attempt to fix response-quality issues unrelated to these three specific mechanisms (e.g. a small local model's general instruction-following under a long agentic system prompt) - only the reproducible, deterministic failures pinned above.


iwasaki1112:
Thanks for the review. This is a real observation, but it's pre-existing behavior, not something introduced by this PR.

The `else: _merge_extra_body(sanitized_kwargs, {"enable_thinking": enable_thinking})` branch (for `qwen3` models on litellm providers other than Ollama/Dashscope/Tongyi-Qianwen), and the two-call pattern that drops that particular `extra_body` update (`_clean_conf()` calling `_apply_model_family_policies()` with only `gen_conf`, then `async_chat()` calling it again with only `request_kwargs`, recomputing `thinking_type` from an empty `gen_conf`) both predate this PR - `git diff main...fix/ollama-qwen3-thinking-param -- rag/llm/chat_model.py` shows this PR only adds the new `elif backend == "litellm" and provider == SupportedLiteLLMProvider.Ollama:` branch (which writes into `sanitized_gen_conf`, not `sanitized_kwargs`, so it isn't affected by this issue) plus the `max_tokens` rename, the `frequency_penalty` -> `repeat_penalty` rerouting, and `_merge_gen_conf_and_kwargs()`.

So the models this PR actually targets (Ollama-hosted qwen3/qwen3.5) aren't affected by the issue you flagged. It looks like a legitimate bug for qwen3 models hosted behind other litellm providers (e.g. GPUStack), but that's outside this PR's scope (fixing the reported Agent-node hang/garbled-response bug for Ollama specifically). Filing/fixing it separately rather than folding it into this PR.

JinHai-CN:
@iwasaki1112 would you please resolve the conflicts?

Lynn-Inf (commented):
The plumbing itself checks out. I verified the new/changed keys actually survive `_clean_conf`: `reasoning_effort` and `extra_body` are in `LITELLM_ALLOWED_GEN_CONF_KEYS`, and `max_completion_tokens` is in `ALLOWED_GEN_CONF_KEYS` — so none of the three fixes are silently filtered away. The blast radius also matches the description: the Ollama `reasoning_effort` branch sits inside the `if "qwen3" in model_name_lower:` block (qwen3-only), and the `frequency_penalty` reroute is gated on `provider == Ollama`. `_merge_gen_conf_and_kwargs` is applied at the only site that merged `gen_conf` with `kwargs`; the streaming/tools paths pass `**gen_conf` alone but still pick up all three fixes, since they land in `gen_conf` via `_clean_conf` → `_apply_model_family_policies`.

**The one thing I'd want settled before merge is Fix #2's blast radius.** Renaming `max_tokens` to `max_completion_tokens` means the Agent's per-request cap is now actually enforced for *every* `LiteLLMBase`-routed provider (Ollama, Anthropic, Gemini, Groq, DeepSeek, …) where it was previously dropped. That's the intended fix, but it has two consequences worth calling out in the description: (a) truncation becomes real — users who set a small cap, or who relied on the old uncapped behaviour, will now get "The answer is truncated…"; and (b) a new rejection surface — `max_completion_tokens` is the OpenAI-style name, and providers or gateways that don't accept it (older OpenAI-compatible endpoints, including the "OpenAI-API-Compatible" factory pointed at a third party) may now return a 400 where the parameter used to be silently discarded. Could you confirm how that factory behaves, and note the truncation change prominently for anyone upgrading?

Two smaller ones: `reasoning_effort="none"` is version-sensitive — Ollama's `think` field documents boolean or low/medium/high, so it's worth confirming `"none"` is in litellm's accepted set rather than relying on the single verified litellm 1.82.5 / Ollama 0.32.0 combination. And `1.0 + frequency_penalty` has no clamp: OpenAI's range is [-2, 2], which maps to `repeat_penalty` ∈ [-1, 3] while Ollama expects a positive float (its own default is 1.1). The presets are fine (0.3 → 1.3), but an extreme user-set value becomes a new 400 path; a clamp would make that safe. Separately, `frequency_penalty=0` now sends `1.0` rather than Ollama's default `1.1`.

Housekeeping: this needs a rebase — current `main` has restructured the qwen3 block (`_merge_qwen_chat_template_kwargs`, plus the `-preview` / `2.4t-a95b` special case), so the hunk context no longer matches.

Issue #16665, opened 2026-07-06
[Bug]: Duplicate files may occur during Confluence data source file synchronization.


wuxiaoyou:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

1

### RAGFlow image version

v0.26.1

### Other environment information

```Markdown

```

### Actual behavior

Confluence data source file synchronization may produce duplicate files. For pages with attachments or images, two files are synced—one using the original attachment name and the other containing the complete file path.

### Steps to reproduce

```Markdown
Directly synchronizing pages with attachments from a Confluence data source will cause issues.
```

zeweihan:
Excellent root-cause isolation, and Dosu's trace already pinpoints the defect precisely: the dedup logic decides the *semantic identifier* at first sighting using only the bare filename, and only switches to the full hierarchical path once a collision is *detected on a later item* — but the first entry is never retroactively renamed. So an operator sees "duplicates" that are really one bare-named and one pathed entry pointing at identical content.

Worth naming why this is an **identity/provenance defect**, not a cosmetic naming bug — especially for regulated corpora (Confluence is a first-class KB backend for compliance/policy/contract docs).

**1. The first-occurrence naming is order-dependent, which breaks reproducibility of identity.** Which identifier an attachment ends up with depends on the *traversal order* of the sync, not on anything intrinsic to the document. Sync the same space starting from a different root page, or after a page re-order, and an attachment that was "first" (bare name) can become "second" (pathed) or vice versa. The corpus's stable identifiers are now a function of traversal order — so "which document did we ingest under this ID on date X?" has no stable answer. For an audit ("prove we analyzed the version of the policy that lived in Space A"), order-dependent naming is exactly the property that makes that question unanswerable, because the ID silently changed between syncs while the content stayed the same.

[...]

PR #16956, opened 2026-07-15
fix(metadata): meta_filter loses case-insensitive matches and leaks state across dict entries

chuenchen309:
## What

`meta_filter()`'s in-memory `filter_out()` helper has two related bugs in how it coerces `input`/`value` for comparison operators (`=`, `≠`, `>`, `<`, `≥`, `≤`):

**1. Asymmetric commit on partial `literal_eval` failure.** The original code:
```python
input = ast.literal_eval(input)
value = ast.literal_eval(value)
```
runs as two separate statements inside one `try`. If the first succeeds and the second raises, the first assignment already committed — `input` and `value` end up as different types, and the subsequent `.lower()` case-folding silently no-ops for whichever side didn't get lowered as a string. Concretely: metadata cell `"None"` is a valid Python literal (`ast.literal_eval("None")` → `None`), but a query value `"none"` (lowercase) is not — so `status = "none"` never matches a cell whose value is `"None"`, even though the intended semantics are case-insensitive.

**2. `value` mutated in place, reused across dict entries.** `filter_out(v2docs, operator, value)` loops over every `(input, docids)` pair in `v2docs` and coerces `value` inside the loop body without resetting it — so once one entry's `literal_eval(value)` succeeds and rebinds `value` to a non-string, every later entry in the same call compares against that already-coerced leftover instead of the original filter value.

## Fix

- Commit both `literal_eval` results together via tuple assignment (`input, value = ast.literal_eval(input), ast.literal_eval(value)`), so a failure on either side leaves both operands in their pre-coercion form instead of a mismatched mix.
- Save the original `value` before the loop and reset it at the top of each iteration, so per-entry coercion never leaks into the next entry.

## Testing

Added 3 tests to `test/unit_test/common/test_metadata_filter_operators.py` covering both symptoms (case-insensitive match against a metadata cell that's a Python keyword literal, both `=` and `≠`; a numeric `>` comparison unaffected by an earlier dict entry having coerced the query value). Confirmed red on `common/metadata_utils.py` at HEAD (`git stash` the fix, tests fail with the exact symptom described above), green after. Full existing `test_metadata_filter_operators.py` suite (22/22, including the 3 new tests) passes. `ruff check` and `ruff format --check` clean on both touched files.

Sandbox note: this environment has no network access to install `pytest`/`pytest-asyncio`, so tests were run by importing the test module and invoking each `test_*` function directly (same approach as prior PRs from this account against this repo, e.g. #16949). `test_apply_semi_auto_meta_data_filter.py` (the other file exercising `meta_filter` indirectly through `apply_meta_data_filter`) needs `pytest-asyncio` + heavier mocking and wasn't run, but it exercises `apply_meta_data_filter`'s async/LLM-filter-generation path, not `filter_out`'s coercion logic touched here.

Cross-referenced open PR #16833 (also touches `common/metadata_utils.py`) — confirmed via `gh pr diff` it only touches `convert_conditions`/operator-alias normalization and `apply_meta_data_filter`'s `None`-vs-`["-999"]` sentinel logic, not `filter_out`'s comparison-coercion code path. No overlap.

---
This PR was drafted with AI assistance (Claude); I reviewed the change, independently reproduced both symptoms, and take responsibility for it.

### DIFF
--- a/common/metadata_utils.py
+++ b/common/metadata_utils.py
@@ -39,7 +39,13 @@ def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
 
     def filter_out(v2docs, operator, value):
         ids = []
+        original_value = value
         for input, docids in v2docs.items():
+            # Reset to the pristine filter value each iteration -- the comparison branch
+            # below reassigns `value` in place (date normalization, literal_eval coercion),
+            # and reusing that mutated value on the next dict entry compares it against
+            # the wrong (already-coerced) type/content instead of the original filter value.
+            value = original_value
             if operator in ["=", "≠", ">", "<", "≥", "≤"]:
                 # Check if input is in YYYY-MM-DD date format
                 input_str = str(input).strip()
@@ -64,8 +70,14 @@ def meta_filter(metas: dict, filters: list[dict], logic: str = "and"):
                     try:
                         if isinstance(input, list):
                             input = input[0]
-                        input = ast.literal_eval(input)
-                        value = ast.literal_eval(value)
+                    except Exception:
+                        pass
+                    # Commit both literal_eval results together, or neither -- assigning
+                    # just one side (e.g. "None" parses but "none" doesn't) would compare
+                    # mismatched types after lowercasing and silently break the
+                    # case-insensitive match below.
+                    try:
+                        input, value = ast.literal_eval(input), ast.literal_eval(value)
                     except Exception:
                         pass
 
--- a/test/unit_test/common/test_metadata_filter_operators.py
+++ b/test/unit_test/common/test_metadata_filter_operators.py
@@ -164,3 +164,31 @@ def test_or_logic_still_unions_after_empty_first_condition():
     ]
 
     assert set(meta_filter(metas, filters, logic="or")) == {"doc2"}
+
+
+def test_equal_is_case_insensitive_for_python_keyword_literals():
+    # "None" is a metadata cell that happens to be a Python literal keyword; the
+    # query value "none" (lowercase) is not a valid literal on its own. Coercing
+    # one side (input -> None) while leaving the other as the string "none" would
+    # compare mismatched types and silently fail the case-insensitive match.
+    metas = {"status": {"None": ["doc1"], "Active": ["doc2"]}}
+    filters = [{"key": "status", "op": "=", "value": "none"}]
+
+    assert meta_filter(metas, filters) == ["doc1"]
+
+
+def test_not_equal_is_case_insensitive_for_python_keyword_literals():
+    metas = {"status": {"None": ["doc1"], "Active": ["doc2"]}}
+    filters = [{"key": "status", "op": "≠", "value": "none"}]
+
+    assert meta_filter(metas, filters) == ["doc2"]
+
+
+def test_greater_than_unaffected_by_prior_dict_entry_coercing_the_query_value():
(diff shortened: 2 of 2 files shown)


zeweihan:
Both defects reduce to the same dangerous shape: a filter that *reports success* while silently matching fewer documents than its declared semantics. A few angles worth surfacing:

1. The `"None"` / `"none"` case is a Python-keyword-literal trap that bites exactly the metadata values most common in regulated corpora. Boolean/status fields (`True`/`False`/`None`/`active`/`inactive`/`superseded`) are the default vocabulary for "does this clause apply", "is this obligation in force". `ast.literal_eval("None")` → `None` (valid) but `ast.literal_eval("none")` → `ValueError`, so the two operands are never compared as strings and `.lower()` silently no-ops on whichever side stayed a string. Result: `status = "none"` returns zero documents with HTTP 200, indistinguishable from "no documents match" — the same empty-result-confidently-asserts-absence signature as ES `.keyword` routing on date fields (#16832) and ISO `Z` vs `+00:00` string-equality (haystack#11962). The tuple assignment closing it at the coercion boundary is the right place to fix it.

2. The state leak across dict entries is the strictly more dangerous of the two because it is *order-dependent*. Once an earlier `(input, docids)` entry coerces `value` to a non-string, every later entry compares against the leftover. Which documents a filter returns now depends on dict iteration order, so re-ingest or a backend change that reorders the same data produces a *different answer set* with no error — the "which docs did this filter return on date X" question becomes unanswerable, same shape as the Confluence attachment identity drift (#16665) and the same-name-file reader collisions (llama_index#22327). A `value` that survives across loop iterations is the textbook version; reset-at-top-of-loop is correct and minimal.

3. Cheap orthogonal companion (separate from this PR, which fixes the correctness half): stamping `{coercion_applied: bool, input_type, value_type, lowered: bool}` per comparison would make "this filter silently matched fewer docs than its semantics imply" inspectable rather than discoverable only by diffing result sets weeks later. Both the `None`→non-match and the order-dependent leak pass every smoke test that checks "did we get results"; they only surface when a specific corpus value hits the trap. Same `text_coverage` / `formula_render_status` philosophy — turn a silent partial match into a queryable invariant.

Good catch on the `None` literal specifically — that's the kind of defect that hides for months because nobody writes a test with a metadata cell whose string value happens to be a Python keyword.

chuenchen309:
Thanks for the detailed writeup — the regulated-corpus framing for point 1 is a good concrete example of why this isn't just an edge case; boolean/status-style metadata values are exactly where a keyword-literal trap like `"none"` vs `None` bites in practice.

Agreed that point 2 (the order-dependent leak) is the more dangerous of the two precisely because it's silent and non-reproducible across runs unless dict iteration order happens to match — "which docs did this filter return on date X" becoming unanswerable is a good way to put it.

Point 3 (diagnostic stamping) is a genuinely useful idea, but I'd treat it as a separate, deliberately-scoped follow-up rather than folding it into this PR — this PR is specifically the correctness fix, and a `{coercion_applied, input_type, value_type, lowered}` observability layer touches call sites well beyond `meta_filter` and deserves its own design/review pass rather than riding along here. Happy to pick it up as a follow-up if a maintainer confirms it's wanted (or if you'd rather file it yourself, that works too — either way I don't think it should block this fix).

yuzhichang (approved):
LGTM

PR #16958, opened 2026-07-15
fix(deepdoc): recover word boundaries for non-Latin scripts; skip OCR fallback the recogniser can't serve

deadtrickster:
Many PDFs — every TeX-family producer (pdfTeX, LuaTeX, xdvipdfmx) and some others — encode **no space glyphs**; words are separated by positioning only. (In TeX, interword space is *glue*, not a character: pdfTeX writes `[(Summary)-250(of)-250(Contents)] TJ`, where the `-250` *is* the space.) DeepDoc's geometric space-recovery rule was gated on an ASCII-only character class, so non-Latin pages got no spaces and welded into a single token — which then tripped the garbled-text check and fell back to OCR, using a recogniser (`ocr.res`: 6312 CJK / 52 Latin / **6 Cyrillic**) that cannot spell the script.

Repro (freely downloadable file):

```python
import pdfplumber
chars = pdfplumber.open("ed3book.pdf").pages[40].chars
print(sum(1 for c in chars if c["text"] == " "), "/", len(chars))   # -> 0 / 2959
```

Concrete files, both freely downloadable, both TeX:
- English: [SLP3 — Jurafsky & Martin](https://web.stanford.edu/~jurafsky/slp3/ed3book.pdf) (pdfTeX)
- Russian (the Cyrillic case): [*PostgreSQL 18 Internals*, Rogov](https://edu.postgrespro.ru/postgresql_internals-18.pdf) (LuaTeX)

Verified end-to-end with the shipped functions on the real char stream pdfplumber extracts from *PostgreSQL 18 Internals*, page 31 (788 chars, **0 space glyphs**):

```
_insert_word_spaces:
  BEFORE: Каталогbaseсоответствуеттабличномупространствуpg_default,сле-
  AFTER:  Каталог base соответствует табличному пространству pg_default, сле-

_ocr_can_represent (against the real ocr.res, 52 Latin / 6 Cyrillic glyphs):
  Russian line: coverage 27.9% -> False -> keep the text layer
  English line: coverage  100% -> True  -> OCR fallback allowed
```

This fixes both the Python parser (`deepdoc/parser/pdf_parser.py`) and its Go port (`internal/deepdoc/parser/pdf/...`):

1. **Drop the ASCII gate** on the gap rule — a gap is a gap regardless of script. (Aside: the Python parser's *other* space regex, in `__ocr`, already includes Cyrillic — `[0-9a-zA-Zа-яА-Я,.?;:!%]`; the Go port carried the character class over but dropped the `а-яА-Я`.)
2. **Fix the threshold**: insert a space on `gap > 0.25 × mean_char_width`. The old `min(width)/2` (~2.5pt) under-inserted even for Latin — real word gaps are 1.6–2.4pt. Validated against `pdftotext`'s segmentation:

   | rule | Russian (ref 253 words) | English (ref 442 words) |
   |---|---|---|
   | `min(width)/2` (old) | 150 | 416 |
   | `0.25 × mean` (new) | 258 | 440 |

   The new rule lands within ~2% of ground truth on both scripts (RU +2.0%, EN −0.5%); the old rule recovered only 59% of Russian words.
3. **Guard the OCR fallback**: before re-OCRing garbled text, check the recogniser's alphabet actually covers it (`_ocr_can_represent` / `OcrCanRepresent`). It covers ~100% of an English page but ~6% of a pure-Cyrillic one, so discarding a usable text layer for OCR only produces garbage — keep the text layer instead. Genuine PUA garbage is still cleared first.

Python and Go kept in parity (both use strict `>`); tests added on both sides (`TestInsertWordSpaces`, `TestOcrCanRepresent`, Go layout + util). Bonus: TeX PDFs with a perfect text layer are no longer needlessly OCR'd (OCR is the dominant per-page cost).

### DIFF
--- a/deepdoc/parser/pdf_parser.py
+++ b/deepdoc/parser/pdf_parser.py
@@ -196,6 +196,60 @@ class RAGFlowPdfParser:
     # CID pattern regex for unmapped font characters from pdfminer
     _CID_PATTERN = re.compile(r"\(cid\s*:\s*\d+\s*\)")
 
+    _OCR_ALPHABET = None
+
+    @classmethod
+    def _ocr_can_represent(cls, text, min_coverage=0.8):
+        """True if the OCR recogniser's alphabet covers this text well enough to be worth OCRing."""
+        if not text:
+            return True
+        if cls._OCR_ALPHABET is None:
+            res = os.path.join(get_project_base_directory(), "rag/res/deepdoc/ocr.res")
+            try:
+                with open(res, encoding="utf-8") as f:
+                    cls._OCR_ALPHABET = set(f.read())
+            except (OSError, UnicodeDecodeError) as e:
+                logging.warning("Could not load OCR alphabet from %s: %s; treating all text as representable.", res, e)
+                cls._OCR_ALPHABET = set()
+        if not cls._OCR_ALPHABET:
+            return True                      # unknown alphabet: preserve existing behaviour
+        letters = [c for c in text if c.strip()]
+        if not letters:
+            return True
+        covered = sum(1 for c in letters if c in cls._OCR_ALPHABET)
+        return covered / len(letters) >= min_coverage
+
+    # CJK scripts (Han, Hiragana, Katakana, Hangul) do not separate words with
+    # spaces, so a geometric gap between their glyphs must not become one.
+    _CJK_PATTERN = re.compile(r"[ᄀ-ᇿ぀-ヿ㄰-㆏㐀-䶿一-鿿가-힯豈-﫿]|[\U00020000-\U0002fa1f]")
+
+    @classmethod
+    def _insert_word_spaces(cls, chars, gap_ratio=0.25):
+        """Recover missing spaces from character geometry.
+
+        Many PDFs encode no space glyphs and separate words by positioning alone.
+        Append a space to a char when the gap to the next exceeds ``gap_ratio`` of
+        the mean char width; intra-word kerns fall well below that. CJK is skipped:
+        it does not write inter-word spaces, so a gap between CJK glyphs is ordinary
+        tracking, not a boundary. ``chars`` is a list of pdfplumber-style dicts and
+        is mutated in place.
+        """
+        widths = [c["width"] for c in chars if c["text"] and c["text"].strip()]
+        mean_w = sum(widths) / len(widths) if widths else 0
+        if mean_w <= 0:
+            return
+        for cur, nxt in zip(chars, chars[1:]):
+            if (
+                cur["text"]
+                and nxt["text"]
+                and cur["text"].strip()
+                and nxt["text"].strip()
+                and not cls._CJK_PATTERN.search(cur["text"])
+                and not cls._CJK_PATTERN.search(nxt["text"])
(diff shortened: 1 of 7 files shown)


xugangqiang:
Hi @deadtrickster, thanks for this fix — the root cause analysis is spot on. The ASCII-only regex gating word-boundary recovery was clearly wrong for non-Latin scripts, and skipping OCR for scripts the recogniser can't spell is the right call. The overall approach looks solid.

I had two small suggestions:

1. ocr.res path resolution in Go

The Python side uses get_project_base_directory() to locate ocr.res, which handles different deployment layouts. The Go side currently uses os.Getwd():

cwd, err := os.Getwd()
data, err := os.ReadFile(filepath.Join(cwd, "rag", "res", "deepdoc", "ocr.res"))

This assumes the working directory is the project root, which may not hold in containerised deployments or when the binary is launched from elsewhere. The repo already has utility.GetProjectRoot() (internal/utility/path.go) that resolves the project root via env vars → go.mod walking → executable path — would you mind switching to that instead? Something like:

root := utility.GetProjectRoot()
data, err := os.ReadFile(filepath.Join(root, "rag", "res", "deepdoc", "ocr.res"))

The graceful fallback (treating alphabet as unknown on error) is already in place, so a missing file won't crash — this is just about making the happy path work in more environments.

2. Testing evidence

Would it be possible to attach a before/after comparison showing the fix in action? Even a simple screenshot or console snippet — e.g., a Cyrillic PDF page where the old code produced welded-together text (Окнигекак-тоиначе) vs. the new code producing properly spaced text (О книге как-то иначе), and similarly for the OCR-fallback guard. It'd help reviewers confirm the fix addresses the reported issue end-to-end.

Thanks again for tackling this!

deadtrickster:
Thanks for the review @xugangqiang!

1. Refactored the Go `ocr.res` lookup to `utility.GetProjectRoot()` as suggested (instead of `os.Getwd()`).
2. Added the concrete Cyrillic book to the description: *PostgreSQL 18 Internals* (Rogov), https://edu.postgrespro.ru/postgresql_internals-18.pdf. It's typeset with LuaTeX and carries no space glyphs, so it's a clean reproduction of the bug. Below is a before/after from running the shipped functions on its real extracted char stream (page 31: 788 chars, **0 space glyphs**):

```
_insert_word_spaces:
  BEFORE: Каталогbaseсоответствуеттабличномупространствуpg_default,сле-
  AFTER:  Каталог base соответствует табличному пространству pg_default, сле-

_ocr_can_represent (real ocr.res):
  Russian line -> False -> keep the text layer
  English line -> True  -> OCR fallback allowed
```

xugangqiang:
2026/07/17 13:50:23 WARN scan page OCR detect failed page=0 err="OCR model OOM"
--- FAIL: TestOCRMergeChars_OverlappingBoxes (0.00s)
    parser_ocr_test.go:337: box B: expected '乙丙', got "乙 丙"
FAIL

@deadtrickster  Could you please fix CI failures? Above is one example. Please check if there's other failures. Thank you.

deadtrickster:
@xugangqiang thanks for the ping. OOM message is interesting. Looking.

PR #16959, opened 2026-07-15
fix(paper): bound title-pivot merge by chunk_token_num (#12109)

### LINKED ISSUE #12109: [Feature Request]: “chunk_token_num” is ignored by chunk policies causing rerank failure.
### Self Checks

```Markdown

```

Although chunk_token_num was set in dataset configuration, the chunks generated may excceed the max length and cause problem. 

Consider add restrict of chunk length. Sometimes this would decrease performance but be better than failing.
```    
    chunks = []
    last_sid = -2
    for (txt, _), sec_id in zip(sorted_sections, sec_ids):
        if sec_id == last_sid:
            if chunks:
                chunks[-1] += "\n" + txt
                continue
        chunks.append(txt)
        last_sid = sec_id
```
similar to
```
    def add_chunk(t, pos):
        nonlocal cks, tk_nums, delimiter
        tnum = num_tokens_from_string(t)
        if not pos:
            pos = ""
        if tnum < 8:
            pos = ""
        # Ensure that the length of the merged chunk does not exceed chunk_token_num
# Change here
# orginal: if cks[-1] == "" or tk_nums[-1] > chunk_token_num * (100 - overlapped_percent)/100.:
        if **cks[-1] == "" or tk_nums[-1] > chunk_token_num * (100 - overlapped_percent)/100. or tk_nums[-1] + chunk_token_num > chunk_token_num * (100 - overlapped_percent)/100.**:
            if cks:
                overlapped = RAGFlowPdfParser.remove_tag(cks[-1])
                t = overlapped[int(len(overlapped)*(100-overlapped_percent)/100.):] + t
            if t.find(pos) < 0:
                t += pos
            cks.append(t)
            tk_nums.append(tnum)
        else:
            if cks[-1].find(pos) < 0:
                t += pos
            cks[-1] += t
            tk_nums[-1] += tnum
```

```Markdown

```

### Additional information

_No response_

deadtrickster:
Paper's title-pivot merge emitted a whole pivot as one unbounded chunk - the loop quoted in #12109. `_merge_sections_by_pivot` bounds it by `chunk_token_num`; `@@...##` position tags are stripped before counting so they don't eat the budget, and stay in the chunk for `crop()`.

Originally this PR also covered the book chunker and the Go path; #18455 landed those, so it's now scoped to the paper gap only. `rag/app/paper.py` + unit tests (helper contract + a `chunk()`-level guard that's red against the unbounded loop).


deadtrickster:
Addressed both:

1. `book.py`: `chunk_token_num` now falls back to 256 when the configured value is falsy (e.g. an explicit `0`), so `naive_merge` never gets a degenerate limit; positive values pass through unchanged.
2. `test_book_positions.py`: lengthened the passage to 10 words (POS setup and assertions unchanged).

deadtrickster:
On the `ChunkTokenNum` hard-cap comment: I took the second option you suggested (document the exception) rather than splitting mid-line.

`capChunkText` intentionally mirrors python `naive_merge`: it splits at line boundaries and never cuts a boundary-less unit mid-token. Python behaves the same way (it splits at delimiter boundaries `\n。；！？`; a section with no delimiter that exceeds the cap is emitted whole). Hard-splitting a single oversized line at arbitrary token boundaries would cut sentences mid-way and diverge from the python path this PR is matching, for the uncommon case of a single record over the cap.

So I corrected the field doc to describe the actual contract (split at line boundaries; a single line over the cap is emitted whole), and the existing `TestCapChunkText` "single oversized line is not split mid-line" case documents it.

deadtrickster:
@coderabbitai review

deadtrickster:
@coderabbitai review

deadtrickster:
@yingfeng could you please have a look?

deadtrickster:
@xugangqiang ping

xugangqiang:
@deadtrickster  Do you want to merge firstly and then rasie another PR to deal with below suggestion? Or add another commit and fix within one PR?

Suggestion: swap capChunkText and RootChunkAsHeading ordering in buildChunksFromRecordGroups

The current PR applies capChunkText after RootChunkAsHeading:

build chunks → RootChunkAsHeading (prepend root text, drop root) → capChunkText (split oversized)

This has a subtle issue: when RootChunkAsHeading prepends the root text to a chunk that was already near the token budget, the combined chunk may now exceed maxTok. capChunkText then splits it — but only the first split piece retains the root heading; subsequent pieces from the same original chunk lose that context.

Consider swapping the order:

build chunks → capChunkText (split oversized first) → RootChunkAsHeading (then prepend root text)

This way:
1. Every chunk is first ensured to fit within maxTok (split at line boundaries).
2. The root heading is then prepended to all resulting chunks, so every piece carries the heading context.

The root text is typically a short heading/title — prepending it to an already-budget-compliant chunk produces a negligible overshoot, which is an acceptable trade-off vs. dropping heading context from split fragments.

If the root chunk itself gets split by capChunkText (extreme edge case), only its first piece would be used as the root text, and the remaining pieces would appear as normal chunks that also receive that root prefix — still reasonable behavior.

deadtrickster:
@xugangqiang thanks, so I focused on two invariants: 

1. No emitted chunk lacks the heading, including every fragment of a split section.
2. No emitted chunk exceeds chunk_token_num with the heading included.

The edge case - header is bigger than chunk_token_num. I choose to overshoot here.

deadtrickster:
@xugangqiang I see preflight is red here,  looks like I couldn't rerun failed jobs here.

deadtrickster:
ping

xugangqiang:
@deadtrickster

To confirm: you're OK with merging this PR as-is and handling the comment/docstring cleanups in a follow-up, correct?

If so, please resolve conflict. Thank you.

deadtrickster:
@xugangqiang confirmed: happy to merge as-is; I have resolved the conflict and cleaned up the comments in this PR (no follow-up needed).

- Rebased onto current `main`. The rebase surfaced two API changes on main that I adapted: `numericFromAny` moved to `schema.NumericFromAny`, and `Chunker.Invoke` gained the `*gorm.DB` argument.
- Adopted main's `removeTag(...)` on the grouped chunk text.
- Dropped the porting-reference comment.

`bash build.sh --test ./internal/ingestion/component/chunker/...` passes locally. Ready when you are.

deadtrickster:
@xugangqiang ping :-)

xugangqiang:
Reviewed against current `main`. The direction is right and two of the bugs here are real, but I don't think this is mergeable as-is: there is a silent data-loss regression on the Go side, and both of the size fixes are unreachable through the product as shipped.

---

## 1. Does it actually fix the reported problem?

Partially — the most common user path isn't covered.

| Change | Effect |
|---|---|
| `paper.py` pivot cap | ✅ works (default 512), matches the loop quoted in #12109 |
| `book.py` position tag | ✅ genuine bug; `rpartition("@@")` is correct — the old `split("@")` also mangled any `@` in the body text (e.g. e-mail addresses) |
| `book.py` size routing | ⚠️ never triggers for UI users |
| Go title chunker | ❌ unreachable config |

**`book.py` routing doesn't fire for UI users.** The chunk-token field is only rendered for `naive` and `knowledge_graph` (`web/src/components/chunk-method-dialog/index.tsx:219`), and the default `parser_config` for `book` is `nil` in both `api/utils/api_utils.py:328` and `internal/common/parser_config.go:200`. So `not parser_config.get("chunk_token_num")` is always true and `hierarchical_merge` still runs. The "median 47 chars" case in the description is only fixed for people who inject `chunk_token_num` into a book dataset over the HTTP API.

**The Go half is dead config.** `chunk_token_num` is absent from the `TitleChunker` params in `internal/ingestion/pipeline/template/ingestion_pipeline_book.json` and from `initialTitleChunkerValues` in `web/src/pages/agent/constant/pipeline.tsx:339`. `CleanComponentParams` (`internal/ingestion/pipeline/pipeline_params.go:18`) drops any param that isn't in the DSL schema, with a warning. Unless someone hand-edits the DSL JSON, `capChunkText` never runs.

## 2. Blocking regression: PDF positions are dropped

`buildChunksFromRecordGroups` attaches `positions` and `_pdf_positions` at `internal/ingestion/component/chunker/group.go:294-308`, but the new split loop rebuilds each piece from scratch and only carries two keys over:

```go
piece := map[string]any{"text": text}
if v, ok := ch["doc_type_kwd"]; ok { ... }
if v, ok := ch["img_id"]; ok { ... }
```

`positions` / `_pdf_positions` are lost. The guard is `maxTok > 0 || rootText != ""`, so **this fires whenever a user enables `root_chunk_as_heading` in the UI** (`pipeline.tsx:347`) — no `chunk_token_num` needed at all. `hierarchy.go:243` calls the same function, so the built-in Book pipeline (`method: "hierarchy"`) is affected, and PDF chunk highlighting silently breaks. The existing coverage at `group_test.go:239` only exercises the plain path, so nothing catches it.

Worth flagging given the PR title says "keep page positions".

## 3. Other regressions

**`chunk_token_num = 0` semantics are broken.** `book.py:129-130` explicitly sets `chunk_token_num = 0` for tcadp/docling/mineru/paddleocr, meaning "the parser already chunked, don't re-merge". The old `.get("chunk_token_num", 256)` returned `0`; the new `parser_config.get("chunk_token_num") or 256` turns it into `256`, changing behaviour for those four parsers on the `bull < 0` path. Suggest keeping `.get(..., 256)`.

**`paper.py` counts position tags as content.** `sorted_sections` text is `b["text"] + self._line_tag(b, zoomin)` (`paper.py:130`), so every entry carries `@@3\t72.0\t523.0\t88.0\t101.0##`. `num_tokens_from_string(chunks[-1] + "\n" + txt)` counts those tags. The rest of the codebase does the opposite: `hierarchical_merge` strips them before counting (`rag/nlp/__init__.py:1143`) and `naive_merge` keeps `pos` out of the token count entirely. Net effect: with DeepDoc PDFs the real body budget is well under the configured 512, and the error grows with the number of boxes. Strip via `pdf_parser.remove_tag` before counting.

**O(n²) tokenisation.** `_merge_sections_by_pivot` re-tokenises the entire accumulated chunk for every section. A running token count would avoid that.

**Silent default change for paper.** Defaulting to 512 changes chunk sizes for every existing paper dataset, and since the UI doesn't expose the field there's no way back other than passing `0` via the API. Reasonable as a default, but it deserves a release note; no docs are touched here.

**Minor:** `zip(..., strict=True)` converts a previously silent truncation into an exception. The `assert` at `paper.py:217` guards it, but asserts are stripped under `python -O`. Also, `strings.Count(...) > N` in two of the new assertions relaxes them exactly in the single-oversized-line case that is most likely to be wrong.

## 4. On the approach

Using the presence of `chunk_token_num` as a policy switch reads oddly: setting a size **completely disables** hierarchical merging, which is the one thing that distinguishes the Book method from naive. And `hierarchical_merge` has exactly one live caller (`book.py:171`; the `laws.py:275` call is commented out), so after this change it becomes half-dead code behind a condition that is almost never false in practice.

Two cleaner options:

1. **Fix the cause.** The hardcoded `218` at `rag/nlp/__init__.py:1144` / `:1152` is what produces the tiny chunks. Parameterise it with `chunk_token_num` and split over-budget groups. Hierarchy is preserved *and* the size is honoured.
2. **Drop the branch.** Route Book through `naive_merge` unconditionally. That's the de-facto outcome of this PR anyway — better to state it than to leave a dormant branch.

Same on the Go side: `group.go:60-61` already has `minGroupTokens = 32` / `maxGroupTokens = 1024`. Deriving those from `chunk_token_num` sits closer to the abstraction that owns the behaviour than a post-hoc splitting pass at the end.

## 5. Suggested changes before merge

Must fix:

1. Carry `positions` / `_pdf_positions` (and any other keys) onto split pieces in `buildChunksFromRecordGroups`, plus a test covering `root_chunk_as_heading` + `chunk_token_num` with positions.
2. Add `chunk_token_num` to the `TitleChunker` params in the pipeline templates and to the frontend form, otherwise the Go change can't be reached.
3. Restore `.get("chunk_token_num", 256)` so an explicit `0` survives.
4. Strip `@@...##` before counting tokens in `_merge_sections_by_pivot`.

Nice to have:

5. Prefer passing the budget into `hierarchical_merge` (or removing the branch) over switching policy on config presence.
6. If Book/Paper users are meant to configure this, `showMaxTokenNumber` at `chunk-method-dialog/index.tsx:219` needs to include them.

To be clear about what's good here: the `split("@")` → `rpartition("@@")` fix is a real bug fix with tests, the paper cap lands exactly on the code quoted in #12109, `capChunkText` itself is correct (including counting the joining newline and not splitting mid-line), and the test coverage is solid. It's items 1–4 that hold it back.

xugangqiang:
pls fix ci error

deadtrickster:
Rebased on current main. Most of what this PR originally did is now covered by #18455 - the Go `chunk_token_cap` and the book chunker rework - so I dropped those parts rather than ship a second competing path.

What's left is the paper chunker. Its title-pivot merge still emits a whole pivot as one unbounded chunk, the exact loop quoted in #12109. `_merge_sections_by_pivot` in `rag/app/paper.py` bounds that by `chunk_token_num`. The `@@...##` position tags are stripped before counting so they don't eat the budget, and stay in the chunk text for `crop()`.

Scoped down to `rag/app/paper.py` plus a unit test. CI should be green now.

xugangqiang:
LGTM @yuzhichang  pls have a second eye.

JinHai-CN:
@deadtrickster would you please resolve the conflicts and rebase the main branch?

deadtrickster:
green!

xugangqiang:
LGTM

PR #16960, opened 2026-07-15
feat(llm): add AWS Bedrock reranker connector

vincentlambert:
## What

Adds a reranker connector for the **Bedrock** factory, which previously offered
chat/embedding/CV models but no reranker — selecting a Bedrock rerank model
raised `Factory not in rerank model`.

## How

`BedrockRerank` calls the `bedrock-agent-runtime` Rerank API. It reuses the same
JSON key protocol as `BedrockEmbed` (`auth_mode` / `bedrock_region` /
`bedrock_ak` / `bedrock_sk`, with `access_key_secret` / `iam_role` /
`assume_role` modes). Documents are truncated to the model window (Cohere Rerank
v3.5 ~2k of its shared 4k window, Amazon Rerank v1 8k) on top of Bedrock's own
internal truncation. Scores are returned in `[0, 1]`, so the shared
`Base.similarity` normalization applies unchanged.

Verified against `amazon.rerank-v1:0` and `cohere.rerank-v3-5:0` in
`eu-central-1`.

> Note: this PR adds the connector only. Bedrock rerank models can be selected by
> adding the relevant entries to `conf/llm_factories.json` under the Bedrock
> provider; that catalog change is intentionally left out of this PR.

## Tests

`test/unit_test/rag/llm/test_bedrock_rerank.py` — boto3 is mocked (no AWS call):
score-by-index mapping, per-model document truncation, model ARN construction,
auth-mode validation and the empty-input short-circuit. `pytest` green alongside
the existing reranker normalization suite.

### DIFF
--- a/rag/llm/rerank_model.py
+++ b/rag/llm/rerank_model.py
@@ -355,7 +355,8 @@ class BedrockRerank(Base):
 
     def _compute_rank(self, query: str, texts: List) -> Tuple[np.ndarray, int]:
         # Truncate to the model token window, then enforce the API's hard 32k-char
-        # per-document limit (a longer RerankTextDocument.text is rejected).
+        # per-text limit (a longer RerankTextQuery / RerankTextDocument is rejected).
+        query = query[: self._MAX_DOC_CHARS]
         texts = [truncate(t, self.doc_max_tokens)[: self._MAX_DOC_CHARS] for t in texts]
         # Bedrock does not report token usage; count locally like CoHereRerank.
         token_count = num_tokens_from_string(query) + sum(num_tokens_from_string(t) for t in texts)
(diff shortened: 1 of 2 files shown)


vincentlambert:
Thanks for the review. Addressed all three points in `1145edfb8`:

- **32,000-char per-document limit** — each document is now hard-capped at 32,000 characters (`RerankTextDocument.text`) after the token-window truncation, so an over-long chunk can no longer be rejected by the API.
- **1,000 sources / `numberOfResults` limit** — requests are now batched in chunks of 1,000, and each score is mapped back to its global position. (In practice RAGFlow reranks a pool of ~64, but the guard makes `similarity()` correct for any input size.)
- **Diagnostics** — added a `logging.debug` line for the new flow (model, region, source/token/result counts, latency). It deliberately excludes the query, document text and credentials.

New unit tests cover the char cap (long doc capped, short doc untouched) and multi-batch requests (2001 docs → 3 calls, each ≤ 1,000). `pytest` and `ruff` green.

vincentlambert:
Addressed the second review pass in `41a795816`:

- **Query 32k-char cap** — the query text is now capped at 32,000 characters as well, same as documents.
- **Pagination** — the batch loop now drains `nextToken` pages and merges each page back to its global index, so no score is dropped when a response is paginated.
- **Exception narrowing** — the result-parsing `except` is now `(KeyError, IndexError, TypeError)` instead of bare `Exception`.
- **Tests** — the char-cap test now asserts exactly `32000`, the batch test asserts `numberOfResults == len(sources)` per request, plus new tests for the query cap and for full pagination consumption.

One note on the "don't pass the full response to `log_exception`" point: I kept `log_exception(_e, res)`. The rerank response only contains `{index, relevanceScore}` entries plus response metadata — no query, document text or credentials — and this matches the exact idiom used by the other rerank connectors in this file (`JinaRerank`, `SILICONFLOWRerank`, `BaiduYiyanRerank`, ...). Happy to change it if you'd prefer a different convention.

yingfeng:
Thanks for the contribution, can you also update to Go version ? Python version will be legacy since v1.0 which will be launched in Aug.

Lynn-Inf (approved):
LGTM

vincentlambert:
@yingfeng I’ll port it to Go when we make the switch, as soon as v1.0 is released in August.

PR #16961, opened 2026-07-15
feat(nlp): rerank on natural chunk text, not tokenized content_ltks

vincentlambert:
## What

`Dealer.rerank_by_model` feeds external rerankers the tokenized `content_ltks`
(stemmed, accent-split) instead of the natural chunk text. Neural rerankers are
trained on natural language and score that mangled input far lower.

## Why

Measured on real multilingual (French) chunks, the **same relevant chunk** scored:

| Query | `content_with_weight` (natural) | `content_ltks` (current) | ratio |
| --- | --- | --- | --- |
| backup frequency / retention | 0.548 | 0.021 | ~26x |
| cyber insurer / notification delay | 0.314 | 0.033 | ~10x |
| systems RTO / RPO | 0.847 | 0.009 | ~94x |

With `content_ltks`, no chunk cleared the default `similarity_threshold` (0.2):
the reranker was effectively inert and retrieval was carried by keyword
similarity alone, which forces users to lower the threshold to ~0.05 just to get
any results back.

## How

Build the reranker input from `content_with_weight` (markup preserved), **passed
as-is**: `remove_redundant_spaces()` is ASCII-oriented and mangles multilingual
text (`sécurité des données` → `sécuritédes données`), so it is only applied to
the tokenized fallback used when `content_with_weight` is absent or empty.
`ins_tw` / `token_similarity` (the keyword-similarity half of the blend) are
unchanged, so the keyword signal is kept. Per-provider truncation to the model
window stays the reranker connector's responsibility.

The same change is mirrored in the Go retrieval path
(`internal/service/nlp/reranker.go`, `RerankByModel`) so Python and Go
retrieval do not diverge.

Note: the numbers above were measured on French / multilingual content. Rerankers
(Cohere, Jina, Voyage, BGE, ...) are trained on natural text, so this should help
or be neutral across languages; maintainers may want to sanity-check CJK, where
`content_ltks` carries the segmented form.

## Tests

- `test/unit_test/rag/test_rerank_by_model_input.py` — asserts the reranker
  receives the natural text untouched (accented spacing preserved), falls back
  to the cleaned tokenized form when `content_with_weight` is absent/empty/None,
  preserves markup, and keeps the tokenized `ins_tw` for keyword similarity.
- `internal/service/nlp/reranker_question_tks_test.go` — same contract on the
  Go side: natural text preferred and passed as-is, tokenized fallback (absent
  and empty `content_with_weight`).

`pytest` and `go test ./internal/service/nlp/` green.

### DIFF
--- a/internal/service/nlp/reranker.go
+++ b/internal/service/nlp/reranker.go
@@ -146,8 +146,18 @@ func RerankByModel(
 		tks = append(tks, questionTks...)
 		insTw = append(insTw, tks)
 
-		// Build document text for model reranking
-		docText := RemoveRedundantSpaces(strings.Join(tks, " "))
+		// Feed the reranker the natural chunk text (markup preserved), not the
+		// tokenized content_ltks. Neural rerankers score stemmed / accent-split
+		// tokens far lower, which collapses relevance scores and forces an
+		// artificially low similarity_threshold. The natural text is passed
+		// as-is: RemoveRedundantSpaces is ASCII-oriented and mangles
+		// multilingual text ("sécurité des données" -> "sécuritédes données"),
+		// so it is only applied to the tokenized fallback used when
+		// content_with_weight is absent. Mirrors rag/nlp/search.py.
+		docText := extractNaturalText(chunk)
+		if docText == "" {
+			docText = RemoveRedundantSpaces(strings.Join(tks, " "))
+		}
 		docs = append(docs, docText)
 	}
 
@@ -547,6 +557,16 @@ func extractContentTokens(fields map[string]interface{}, cfield string) []string
 		}
 	}
 	return result
+}
+
+// extractNaturalText returns the natural chunk text (content_with_weight),
+// or "" when the field is absent or not a string.
+func extractNaturalText(fields map[string]interface{}) string {
+	v, ok := fields["content_with_weight"].(string)
+	if !ok {
+		return ""
+	}
+	return v
 }
 
 // extractTitleTokens extracts title tokens from chunk fields
--- a/internal/service/nlp/reranker_question_tks_test.go
+++ b/internal/service/nlp/reranker_question_tks_test.go
@@ -115,3 +115,50 @@ func TestRerankByModelWithoutQuestionTks(t *testing.T) {
 		t.Errorf("got %q, want %q", doc, want)
 	}
 }
+
+// TestRerankByModelPrefersNaturalText pins the reranker input contract: the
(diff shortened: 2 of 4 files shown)


6ba3i (changes requested):
The direction looks right and this does address the actual reranking issue, but I think two things still need fixing before merge.

First, `content_with_weight` is still passed through `remove_redundant_spaces()`. That helper is ASCII-oriented and breaks real multilingual text, e.g. `sécurité des données` becomes `sécuritédes données`. Natural text should probably be passed through as-is, keeping the cleanup only for the tokenized fallback.

Second, the Go reranking path still has the same original behavior and continues sending `content_ltks + title_tks + important_kwd` to the reranker. Since `content_with_weight` is already available there, the same fix should be mirrored in Go to avoid Python/Go retrieval divergence.

6ba3i:
hello @vincentlambert could you take a look at my comment and fix the shared issues

vincentlambert:
@6ba3i thanks for the review, both points addressed in the rebased branch:

- `content_with_weight` is now passed as-is to the reranker; `remove_redundant_spaces()` only applies to the tokenized fallback (test covers `sécurité des données` staying intact).
- Mirrored in Go `RerankByModel` (`internal/service/nlp/reranker.go`) with the same fallback, plus tests for natural-text preference and absent/empty `content_with_weight`.

Rebased on current `main` (picks up the `question_tks` and `.get(cfield, "")` changes), `pytest` and `go test ./internal/service/nlp/` green.

vincentlambert:
@coderabbitai on the `Rerank` → `RerankByModel(ctx, rerankModel, chunks, nil, nil, ...)` finding (`internal/service/nlp/reranker.go` L72): the observation is correct, but it is pre-existing and out of scope for this PR.

- That line is unchanged from `main` (dates back to d19a036cda, July 2026); this PR only touches the document construction inside `RerankByModel`.
- The `Rerank` wrapper has no production callers. The actual retrieval path, `RetrievalService.scoreSearchResult` in `internal/service/nlp/retrieval.go`, calls `RerankByModel` directly with `searchResult.IDs` and `searchResult.Field`, so the natural-text selection added here is reached in production.

Leaving the wrapper as-is to keep this PR focused on the reranker input contract. Happy to open a separate issue/PR for the nil-ids path if the maintainers want it addressed.

6ba3i (approved):
LGTM

qinling0210 (approved):

PR #16969, opened 2026-07-16
fix: lefthook deps serialization

dcc123456:
### Summary

fix: lefthook deps serialization

Extract npm ci guard into web-deps job; use native deps for serialization instead of a mkdir mutex that could leave stale locks on interrupted commits. Also use single quotes in echo to work around lefthook v2.1.10 stripping double quotes on Windows.


Lynn-Inf (approved):
LGTM

PR #16985, opened 2026-07-16
Fix QA CSV multiline row parsing

### LINKED ISSUE #16791: [Bug]: QA parser for .csv silently corrupts Q&A pairs: csv.reader quote handling + row/line index mismatch splice answers from wrong lines

zcxGGmu:
### Summary

Fixes #16791.

This PR fixes Q&A CSV parsing when quoted fields contain newline characters. The parser now reads the full CSV text as a stream instead of splitting it into physical lines before passing it to `csv.reader`, preserving multiline question/answer pairs and correct source row attribution.

### Changes

- Parse CSV content with `StringIO` so quoted multiline fields remain intact.
- Use the CSV reader line counter to track physical source line spans.
- Add a regression test for quoted multiline answers.

### Tests

- `uv run python -m pytest test/unit_test/rag/app/test_qa_csv.py -q`

### DIFF
--- a/rag/app/qa.py
+++ b/rag/app/qa.py
@@ -284,6 +284,14 @@ def mdQuestionLevel(s):
     return (len(match.group(0)), s.lstrip("#").lstrip()) if match else (0, s)
 
 
+def _qa_delimiter(lines):
+    counts = {
+        delimiter: sum(1 for row in csv.reader(lines, delimiter=delimiter) if len(row) == 2)
+        for delimiter in (",", "\t")
+    }
+    return "\t" if counts["\t"] >= counts[","] else ","
+
+
 def chunk(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, lang="Chinese", callback=None, **kwargs):
     """
     Excel and csv(txt) format files are supported.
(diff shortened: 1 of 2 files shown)


zcxGGmu:
Updated the text and csv Q&A parsing paths to share the csv.reader flow. Progress callbacks now use the reader index for the interval and current physical line for progress, so malformed rows no longer trigger repeated 0% callbacks.

Validation:
- uv run --no-sync python -m pytest test/unit_test/rag/app/test_qa_csv.py
- uv run --no-sync ruff check rag/app/qa.py test/unit_test/rag/app/test_qa_csv.py
- git diff --check

yuzhichang:
Duplicated with https://github.com/infiniflow/ragflow/pull/16881.

PR #16760, opened 2026-07-08
fix(dataflow): make_encode_batch synchronous so thread_pool_exec can run it

wuisabel-gif:
## Problem

On the custom pipeline / MinerU ingestion path, embedding fails with:

```
[ERROR]: cannot unpack non-iterable coroutine object
```

## Root cause

`DataflowService._encode_batch` was declared `async`, but its body is blocking
work (`EmbeddingUtils.truncate_texts` + `embedding_model.encode`). It is submitted
to a worker thread via `thread_pool_exec`, which runs the callable directly:

```python
vts, c = await thread_pool_exec(self._encode_batch, batch, embedding_model)
```

Passing an `async def` there only builds a coroutine (it never runs), so the
executor returns that un-awaited coroutine and the tuple unpack raises
"cannot unpack non-iterable coroutine object". Every other `thread_pool_exec`
caller passes a plain sync callable.

## Fix

Drop `async` from `_encode_batch`. The sync body then runs correctly in the
worker thread and unpacks fine. No call-site change needed.

Adds a regression test that runs `_encode_batch` through `thread_pool_exec` and
asserts the `(vectors, tokens)` tuple unpacks.

Fixes #16581

PR #16994, opened 2026-07-16
Fix dataflow pipeline: async encode_batch coroutine bug and metadata overwrite bug

iwasaki1112:
## Summary
- `_encode_batch` in `dataflow_service.py` was declared `async def` but is invoked via `thread_pool_exec` (which submits plain callables to a worker thread, not awaited coroutines). Calling it without awaiting just returns a coroutine object instead of running its body, so the worker thread returns that coroutine as the "result". The caller then does `vts, c = await thread_pool_exec(...)`, which tries to unpack the coroutine object and raises `TypeError: cannot unpack non-iterable coroutine object`. This made every ingestion pipeline (dataflow) Indexer step fail during embedding, leaving 0 chunks indexed.
- `_update_document_metadata` called `update_metadata_to(metadata, existing_meta)` - passing the newly generated metadata as the update target and the already-saved metadata as the source to merge in. `update_metadata_to(a, b)` overwrites matching string keys in `a` with `b`'s values, so on every re-parse after the first, the old, already-persisted metadata (e.g. `document_description`) silently overwrote the fresh one the LLM had just generated. Only the very first parse (with no prior metadata) ever stuck; every subsequent parse was a no-op from the user's perspective despite the model producing different output each time.

## Fix
- Make `_encode_batch` a plain sync method, matching the other callables passed through `thread_pool_exec`.
- Swap the arguments to `update_metadata_to(existing_meta, metadata)` so the existing metadata is the base and the freshly extracted metadata is merged on top.

## Testing
- Built a custom ingestion pipeline (Parser -> Chunker -> Extractor (Metadata destination) -> Indexer) against a local Ollama (qwen3.5:9b) deployment.
- Before the fix: every parse failed with `TypeError: cannot unpack non-iterable coroutine object`, leaving 0 chunks.
- After the fix: parsing completes successfully, chunks are embedded and indexed.
- Re-parsed the same document multiple times: before the fix, `document_description` stayed frozen at the first-ever generated value; after the fix, each re-parse correctly updates it to the newly generated summary.


xugangqiang:
### Review findings (read-only)

**1. Hunk 1 (`async def _encode_batch` -> `def`) is already fixed on `main`**
The coroutine-unpack bug is real and the diagnosis is correct (`thread_pool_exec` runs the callable synchronously, so an `async def` returns an un-awaited coroutine that fails to unpack; the broad `except Exception` in `_embed_chunks` then silently downgrades it to `progress=-1`, 0 chunks). However, upstream `main` already applied the byte-identical fix in commit `b225b8ec6` (PR #19045, 2026-08-31) — `dataflow_service.py:295` is now `def _encode_batch`. This half of the PR is a no-op against current `main`. Please rebase and drop this hunk, or it will mislead reviewers into thinking `main` is still broken.

**2. `update_metadata_to` argument swap is a real bug, but it only fixes one of several sites**
The old order `update_metadata_to(metadata, existing_meta)` lets the persisted value win over the freshly generated one (str values are overwritten by the 2nd arg per `common/metadata_utils.py:394-402`), freezing `document_description` on re-parse. Swapping to `update_metadata_to(existing_meta, metadata)` fixes the dataflow path. But the same bug exists elsewhere and is NOT addressed:
- `rag/svr/task_executor.py:919-925` (the legacy twin of this very method — same `run_dataflow_metadata` recording key)
- `rag/svr/task_executor_refactor/chunk_post_processor.py:257` (refactor-side LLM metadata generation)
- `rag/svr/task_executor.py:565-573` (question/metadata generation)

Leaving the legacy twin unfixed breaks `TE_RUN_MODE=1` dry-run parity, which compares recorded keys and counts `mismatched_keys` (`task_manager.py:167-178`).

**3. A bare swap silently inverts structured-value precedence**
`update_metadata_to` is asymmetric: str/list[str] values are "2nd arg wins", but dict / list[dict] structured values are "1st arg wins" (`metadata_utils.py:336-369`). Swapping args fixes strings but now makes old structured values (nested objects, `outline`) win over new ones. Prefer an explicit direction (e.g. a dedicated `merge_fresh_over_existing(existing, fresh)` helper in `common/metadata_utils.py`) used consistently at all 4 call sites, rather than relying on argument order. Note `test/unit_test/common/test_update_metadata_to.py:15` already pins "new fields first, existing doc meta second" as the canonical shape for the outline-persist path.

**4. No test coverage; existing tests mock both methods**
`test/unit_test/rag/svr/task_executor_refactor/test_dataflow_service.py` patches out `_embed_chunks` and `_update_document_metadata` entirely, so neither bug is exercised — that is exactly why the async bug reached production. A unit test asserting that `_update_document_metadata` overrides an existing `document_description` while preserving keys absent from the fresh metadata would prevent regression.

**5. Possible user-edit regression (needs product confirmation)**
User-edited `meta_fields` (written via `api/apps/restful_apis/document_api.py:278`) share the same namespace as LLM-extracted fields. With the swap, re-parsing overwrites user manual edits with fresh LLM output; the old order incidentally protected them.

Nit: the description "every ingestion pipeline (dataflow) Indexer step failed" is slightly overstated — the embed branch only runs when chunks lack pre-generated `q_\d+_vec` fields (`dataflow_service.py:145-147`). Also this duplicates PR #16760, which already carries this fix with a test.

PR #16995, opened 2026-07-16
fix: make await response wait for user input

buua436:

Await Response incorrectly consumed the initial `sys.query` as its input, so the first user interaction was skipped.

This change makes Await Response wait for actual user input while preserving the existing initial-query behavior for the Begin node.

### Type of change


wangq8 (commented):
## Review: PR #16995 — fix: make await response wait for user input

### Summary
A well-executed bug fix addressing a real behavioral issue: **Await Response (UserFillUp) incorrectly consumed `sys.query` as its first input, skipping the first user interaction entirely.** The root cause was shared auto-consume logic between **Begin** and **UserFillUp** via the `_INITIAL_USER_INPUT_CONSUMED_KEY` tracking.

### What was changed (+185/-172 across 5 files)

| File | Δ | What |
|---|---|---|
| `agent/component/begin.py` | +21/-0 | Added `_merge_runtime_inputs` — Begin now retains the auto-consume of `sys.query` (entry-point behavior). |
| `agent/component/fillup.py` | +1/-29 | Simplified `_merge_runtime_inputs` to `return runtime_inputs or {}`; removed shared consumed-key tracking. |
| `internal/agent/canvas/interrupt_resume.go` | +0/-40 | Removed Go-side `initialUserFillUpData` + the auto-consume fast path. |
| `interrupt_resume_test.go` | +4/-60 | Replaced old tests with one asserting UserFillUp now interrupts instead of consuming. |
| `agent_run_e2e_test.go` | +159/-43 | Reworked all integration tests to validate the correct multi-resume flow (menu pause → menu resume → actual work). |

### ✅ What is good

1. **Correct root cause analysis.** The shared `_INITIAL_USER_INPUT_CONSUMED_KEY` meant a Begin node consuming the initial query would also prevent downstream UserFillUp nodes from waiting — and UserFillUp itself was consuming `sys.query` directly.
2. **Clean separation of concerns.** By moving the auto-consume logic only into `Begin._merge_runtime_inputs`, the fix is surgically targeted. Begin (the entry point) still auto-consumes; every other UserFillUp (i.e. every Await Response node in the canvas) waits for actual user input.
3. **Cross-stack consistency.** Both the Python (`begin.py`/`fillup.py`) and Go (`interrupt_resume.go`) sides are updated together — the Go-side `initialUserFillUpData` fast path was the same bug in the Eino runtime.
4. **Thorough test updates.** All e2e fixtures (Loop, Categorize, Iteration, VarAssigner, DataOps) were updated to match the new 3-run flow: menu pause → menu selection → actual input. The test restructuring also includes Redis checkpoint setup for more realistic coverage.
5. **Clean removal, no dangling references.** The removed consumed-key constant has no remaining usage; the Go-side function deletion is complete.

### 🔍 Minor concerns

- **fillup.py simplification is very aggressive.** `return runtime_inputs or {}` replaces 28 LoC that previously did field-matching. While this is correct (if `runtime_inputs` is `None` on resume, an empty dict simply means "wait for user"), a brief comment explaining *why* UserFillUp should never auto-consume would help future readers.

- **Code duplication.** `Begin._merge_runtime_inputs` is a near-exact copy of the removed `UserFillUp._merge_runtime_inputs`. If the merge logic ever needs changing (e.g. new field types), both locations need updating. Acceptable for now since the behaviors are intentionally diverged, but worth flagging.

- **Begin multiple-invoke safety.** The new Begin method no longer uses a consumed-key guard. If a Begin node's `_invoke` could be called more than once in a session, `sys.query` would be re-consumed each time. In practice Begin executes once per session as the entry point, so this is fine.

### 🔮 Regression risk

**Low.** The behavioral change is well-defined and properly enforced at both stack layers. The e2e tests thoroughly validate the new contract:
- All fixtures now produce the correct wait-for-user interrupt at the menu
- All resume paths still complete with the correct final outputs
- The CategorizeResume test specifically guards the regression case where a resumed payload could be overwritten by re-consumed `sys.query`

### 🏁 Verdict

**Approve** ✅ — clean, correct, well-tested. The fix addresses the actual bug without overcomplication. The minor callouts (missing comment, code duplication) are nits, not blockers.

PR #16997, opened 2026-07-16
Fix OceanBase chunk order field persistence

### LINKED ISSUE #16967: [Bug]: OceanBase schema does not persist `chunk_order_int`, causing chunk detail and retrieval failures
### Self Checks

6a4b9be4261038e8b41c0420e2a539edac92ffc9

### RAGFlow image version

RAGFlow `v0.26.4` (`cb93883f3f8c975eecb2fed81210effeb3bdb06f`)

```Markdown
- Document engine: OceanBase
- OceanBase version: `4.4.2`
- OceanBase compatibility mode: MySQL
- Embedding dimension in the affected table: `3584`
```

### Actual behavior

- Uploading, parsing, chunking, and embedding complete successfully.
- Opening chunk details raises:

  ```text
  KeyError('chunk_order_int')
  ```

- Retrieval raises:

  ```text
  (pymysql.err.OperationalError) (1054, "Unknown column 'chunk_order_int' in 'field list'")
  ```

- The generated OceanBase table contains `_order_id`, but does not initially contain `chunk_order_int`.
- After manually adding the following nullable column and restarting the RAGFlow server/task processes:

  ```sql
  ALTER TABLE ragflow_<tenant_id>
      ADD COLUMN chunk_order_int INT NULL;
  ```

  reparsed chunks still have `chunk_order_int IS NULL`, and the chunk detail request can still raise `KeyError('chunk_order_int')`.

### Expected behavior

- The OceanBase chunk table schema should contain every field selected by the retrieval implementation, including `chunk_order_int`.
- Parsing should persist the chunk sequence into the physical `chunk_order_int` column.
- Chunk details and retrieval should work without a missing-column SQL error or Python `KeyError`.
- Existing OceanBase tables should have a documented migration path when a required column is introduced.
- Legacy rows with a missing or `NULL` chunk order should be handled gracefully rather than causing an unhandled `KeyError`.

### Steps to reproduce

```Markdown
1. Deploy RAGFlow v0.26.4 with OceanBase 4.4.2 as the document engine.
2. Create a knowledge base using an embedding model with a 3584-dimensional vector.
3. Upload a document.
4. Parse and embed the document.
5. Confirm that parsing and chunking finish successfully.
6. Open the document's chunk details.
7. Run a retrieval request against the knowledge base.
```

### 1. The search implementation selects and sorts by `chunk_order_int`

In v0.26.4, the default source field list includes `chunk_order_int`, and the no-query ordering path also sorts by it:

- [`rag/nlp/search.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/rag/nlp/search.py#L143-L180)

This causes OceanBase to generate a query equivalent to:

```sql
SELECT ..., chunk_order_int, ...
FROM ragflow_<tenant_id>;
```

If the physical column is missing, OceanBase returns error `1054`.

### 2. The OceanBase column definition omits `chunk_order_int`

The fixed OceanBase column list contains `_order_id`, `group_id`, and `mom_id`, but not `chunk_order_int`:

- [`rag/utils/ob_conn.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/rag/utils/ob_conn.py#L28-L97)

The v0.26.4 ES-to-OceanBase migration schema also does not define `chunk_order_int`:

- [`tools/es-to-oceanbase-migration/src/es_ob_migration/schema.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/tools/es-to-oceanbase-migration/src/es_ob_migration/schema.py)

### 3. Unknown document fields are moved into `extra`

The OceanBase insert path checks every document field against the fixed `column_names` list. A non-vector field that is not in that list is moved into the `extra` JSON object before `upsert()`:

- [`rag/utils/ob_conn.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/rag/utils/ob_conn.py#L700-L755)

The effective behavior is:

```python
if key not in column_names:
    document["extra"][key] = value
```

Consequently, even if an upstream component produces:

```python
chunk["chunk_order_int"] = i
```

the value is not persisted to the physical `chunk_order_int` column because the field is absent from the OceanBase column whitelist. It may instead be stored under `extra.chunk_order_int`.

### 4. `chunk_order_int` is not generated by every Tokenizer path

The Flow Tokenizer assigns a chunk order when the upstream output format is `chunks`:

- [`rag/flow/tokenizer/tokenizer.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/rag/flow/tokenizer/tokenizer.py#L112-L205)

```python
if from_upstream.output_format == "chunks":
    for i, chunk in enumerate(chunks):
        chunk["chunk_order_int"] = i
```

The `markdown`, `text`, `html`, and other JSON branches do not appear to assign the same field. Traditional parsing paths may also bypass this Flow Tokenizer component.

### 5. `NULL` fields are omitted from the returned Python entity

The OceanBase result conversion skips fields whose value is `None`:

- [`rag/utils/ob_conn.py`](https://github.com/infiniflow/ragflow/blob/cb93883f3f8c975eecb2fed81210effeb3bdb06f/rag/utils/ob_conn.py#L660-L675)

The effective behavior is:

```python
if value is None:
    continue
entity[field] = value
```

Therefore, after the physical column is added, a row containing `chunk_order_int = NULL` is converted into a Python dictionary without that key. Any later direct access such as:

```python
chunk["chunk_order_int"]
```

can still raise `KeyError('chunk_order_int')`.

## Suspected root cause

RAGFlow v0.26.4 has inconsistent OceanBase field definitions across the processing pipeline:

```text
Search/read path requires chunk_order_int
                |
                v
OceanBase table definition omits chunk_order_int
                |
                v
OceanBase insert whitelist treats it as an unknown field
                |
                v
The physical column is absent or remains NULL
                |
                v
SQL 1054 when absent; Python KeyError when present but NULL
```

This appears to be a RAGFlow OceanBase connector/schema consistency issue rather than an OceanBase 4.4.2 server compatibility issue.

## Suggested areas to address

Could the maintainers please confirm the intended relationship between `_order_id` and `chunk_order_int` and make the field handling consistent across:

1. OceanBase table creation definitions;
2. migrations for existing `ragflow_<tenant_id>` tables;
3. the OceanBase insert/upsert column whitelist;
4. all parser/Tokenizer output branches;
5. retrieval source fields and ordering;
6. result conversion and compatibility handling for legacy `NULL` rows?

A migration or upgrade note for existing OceanBase deployments would also be helpful, because adding only the physical column does not cause v0.26.4 to populate it.

## Additional notes

- Restarting OceanBase is not required for the DDL to become visible.
- Restarting RAGFlow processes refreshes the OceanBase client/SQLAlchemy metadata, but it does not resolve the write whitelist mismatch or populate historical rows.
- Downgrading has not been identified as a safe workaround because earlier versions may not support the same OceanBase schema, and reusing current metadata/vector tables across versions may introduce additional incompatibilities.

zcxGGmu:
## Summary

- Add `chunk_order_int` to the OceanBase chunk schema and extra-column migration list so the retrieval sort field is physically persisted.
- Default `chunk_order_int` to `0` when OceanBase returns `NULL` and when ES-to-OceanBase migration input is missing the field.
- Add regression coverage for the connector schema/default handling and migration conversion.

## Tests

- `PYTHONUNBUFFERED=1 uv run python -m pytest test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py -q`
- `PYTHONUNBUFFERED=1 uv run pytest tests/test_schema.py -q` from `tools/es-to-oceanbase-migration`
- `uv run ruff check rag/utils/ob_conn.py test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py tools/es-to-oceanbase-migration/src/es_ob_migration/schema.py tools/es-to-oceanbase-migration/tests/test_schema.py`
- `git diff --check`

Fixes #16967


JinHai-CN:
@whhe Hi Wang He, would you please help review this PR?

rag/utils/ob_conn.py
```diff
@@ -188,12 +191,21 @@ def get_default_value(column_name: str) -> Any:
...
     elif column_name == "removed_kwd":
         return "N"
-    elif column_name == "_order_id":
+    elif column_name in ["_order_id", "chunk_order_int"]:
         return 0
     else:
         return None
 
 
+def _get_entity_value(column_name: str, value: Any) -> Any:
```
whhe, line 203: **[Medium]** _get_entity_value applies get_default_value at read time for ANY column with a non-None default (available_int→1, removed_kwd→"N", _order_id→0), not just chunk_order_int. Previously, defaults were only applied during insert(). Now _row_to_entity and _es_row_to_entity will populate keys for these columns with default values even when the DB returns NULL, changing the shape of returned entity dicts. While no current callers in rag/ appear to rely on key absence, this implicit broadening beyond the stated fix scope should be explicitly acknowledged and guarded. Consider scoping the default-fill to chunk_order_int only, or adding a comment and tests covering the broader behavior.

_Reason: The helper calls get_default_value unconditionally for all column names, silently expanding the behavioral change beyond the chunk_order_int fix._

test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py
```diff
@@ -0,0 +1,27 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+from rag.utils.ob_conn import _get_entity_value, column_names, get_column_value, get_default_value
+
+
```
whhe, line 19: **[Low]** Tests only cover chunk_order_int behavior. There are no assertions verifying that _get_entity_value does NOT alter read-time behavior for other defaulted columns (available_int, removed_kwd, _order_id), nor any end-to-end test of _row_to_entity or _es_row_to_entity with NULL inputs. Add at least one assertion like `_get_entity_value("group_id", None) is None` to confirm non-defaulted columns still return None, and consider testing a defaulted column to document the intended scope.

_Reason: The test class verifies chunk_order_int but omits negative/scope tests that would catch unintended behavioral broadening._

whhe:
### Code Review

The fix correctly persists chunk_order_int but introduces a broader behavioral change: _get_entity_value now fills defaults at read time for all columns with defaults (available_int, removed_kwd, _order_id), not only chunk_order_int.

Severity: Critical 0 / High 0 / Medium 1 / Low 1
Inline comments posted: 2

_Generated by [whhe/code-review-bot](https://github.com/whhe/code-review-bot) · Skill fingerprint: `43b5df0c`_

test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py
zcxGGmu: Added scope coverage in 96330e0: `_order_id` and `chunk_order_int` use `0` when read as `NULL`, while `available_int`, `removed_kwd`, and `group_id` remain `None`. Validation: `uv run pytest test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py`, `ruff check rag/utils/ob_conn.py test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py`, and `git diff --check`.

rag/utils/ob_conn.py
zcxGGmu: Addressed in 96330e0. `_get_entity_value` now applies read-time defaults only for `_order_id` and `chunk_order_int`, which keeps the change scoped to ordering fields and avoids filling other defaulted columns on reads. Validation: `uv run pytest test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py`, `ruff check rag/utils/ob_conn.py test/unit_test/rag/utils/test_ob_conn_chunk_order_int.py`, and `git diff --check`.

whhe (approved):
LGTM

Issue #852, opened 2024-05-20
[Question]: how to call api? example Postman GET/POST


vbmcpy:
### Describe your problem

How would I call the API, I added the token in the chat but when I try to call the API URL I didn't find an example of a call, they could create an example postman json/ with token bearer in the documentation.

![image](https://github.com/infiniflow/ragflow/assets/114023788/7bdffd8a-5581-44ca-9714-46a9854d0ad3)

Thnks

KevinHuSh:
Please refer to [this](https://github.com/infiniflow/ragflow/blob/main/docs/conversation_api.md#path-apinew_conversation).
It uses GET method instead of POST.

vbmcpy:
Hi, thanks for the feedback, now I can do get but when I send a post I get not allowed? Would you have any example in json of how I should send it?

![image](https://github.com/infiniflow/ragflow/assets/114023788/efcaac06-1f20-46dd-ab7e-d15790449c5e)

guoyuhao2330:
> 嗨，感谢您的反馈，现在我可以得到，但是当我发送帖子时，我不被允许？你能在json中举例说明我应该如何发送它吗？
> 
[...]

PR #17007, opened 2026-07-16
Fix cv model siliconflow and zhipu cannot describe video, capture 3 images from video and sent to llm (#852)

wangq8:


rag/llm/cv_model.py
```diff
@@ -339,39 +406,6 @@ def __init__(self, key, model_name="qwen-vl-chat-v1", lang="Chinese", base_url=N
         # Disable thinking here so parser-side extraction tasks do not emit reasoning text.
         self.extra_body = _qwen3_no_think_extra_body(self.model_name) or self.extra_body
 
-    @staticmethod
```
wangq8, line 342: move to base class

PR #17009, opened 2026-07-16
Fix centered PDF title reading order

### LINKED ISSUE #16820: Go PDF parser places centered document title at the end of the section list (reading-order regression vs Python)
## Summary

The Go PDF parser (`internal/deepdoc/parser/pdf`) mis-orders a **centered document title**, placing it at the very end of `result.Sections` instead of at the top. The body text stays in correct top-to-bottom order; only the centered main title is displaced. This diverges from the Python reference parser and was caught by the integration test `TestIntegration_SectionsText`.

## Reproduction

Parse a single-column PDF whose document title is centered at the top of the page using the Go parser, then inspect `result.Sections`: the centered title appears as the **last** element rather than the first.

Detected via `TestIntegration_SectionsText` (`internal/deepdoc/parser/pdf/parser_pipeline_integration_test.go:97`), which compares section order against a golden snapshot.

## Actual vs expected

For a page with a centered title followed by left-aligned body paragraphs and left-aligned section headings:

```
expected: [Title, para1, para2, "Key Benefits", list, "Conclusion", finalPara]
actual:   [para1, para2, "Key Benefits", list, "Conclusion", finalPara, Title]
```

i.e. `actual == expected[1:] + expected[0:1]`.

Notably, the left-aligned section headings (`Key Benefits`, `Conclusion`) stay in their correct interleaved positions — **only the centered main title is displaced.**

## Root cause

Two factors combine.

1. **Column assignment treats the centered title as its own column.**
   `AssignColumn` (`internal/deepdoc/parser/pdf/layout/layout.go:22`) clusters boxes by `x0` per page. A centered title has an `x0` notably to the right of the left-aligned body, so KMeans + silhouette (`findBestK`, `layout.go:97`) selects `k=2`: the title gets `ColID=1`, the body `ColID=0`. Left-aligned headings share the body's `x0`, so they stay `ColID=0`.

2. **The final ordering is column-major, so `ColID=1` is emitted after all of `ColID=0`.**
   `buildLayout` (`internal/deepdoc/parser/pdf/parser.go:455`) calls `FinalReadingOrderMerge` (`parser.go:462`; sorts `page -> ColID -> top -> x0`) and then `NaiveVerticalMerge` -> `processPageBoxes` (`layout.go:364`), which buckets by `ColID` and emits columns in ascending `ColID` order. The centered title (`ColID=1`) is therefore emitted after every body box (`ColID=0`). Within `ColID=0`, boxes are still sorted by `top`, which is why the body remains correctly ordered and only the isolated title is pushed to the end.

## Divergence from the Python reference

The Go migration activated column-major ordering that the Python production path does **not** use.

Python production path: `rag/flow/parser/parser.py` -> `parse_into_bboxes` -> `_parse_loaded_window_into_bboxes` (`deepdoc/parser/pdf_parser.py:1741`):

```
_layouts_rec
_table_transformer_job
_text_merge              # col_id used ONLY to block cross-column horizontal merge
_concat_downward         # -> Recognizer.sort_Y_firstly(self.boxes, 0)  (pure top sort)
_naive_vertical_merge    # groups by page only; col_id grouping is COMMENTED OUT
_extract_table_figure    # spatial nearest-neighbor insertion (min_rectangle_distance)
```

- `sort_Y_firstly` (`deepdoc/vision/recognizer.py:54`) with `threshold=0` is a pure sort by `top` -> final order is `page -> top -> x0`. No `col_id` in the reading order.
- `_naive_vertical_merge` (`pdf_parser.py:928`) has its `col_id` grouping commented out (line 934) and the `col_id`-keyed final sort commented out (line 1006); it groups by page and sorts by `(top, x0)`.
- `_final_reading_order_merge` (`pdf_parser.py:1009`) — the column-major sort that mirrors Go's `FinalReadingOrderMerge` — is **dead code**: defined but never called.
- `col_id` is used **only** in `_text_merge` (`pdf_parser.py:909`) to avoid merging boxes across columns horizontally.

## Regression source

PR #16795 (merge commit `12787996d`, 2026-07-10, "DeepDOC parser hardening") rewrote `processPageBoxes` from a global `(Top, X0)` sort into `ColID`-bucketed column-major emission. The previous global-`Top` sort matched Python's `_naive_vertical_merge`; the new version reintroduces the column-major approach Python abandoned. (Go calling `FinalReadingOrderMerge` at all is a separate, older divergence — it corresponds to Python's dead `_final_reading_order_merge`.)

## Suggested fix

Align Go's final ordering with Python's production path (`page -> top -> x0`) and restrict `col_id` to horizontal-merge gating:

1. `processPageBoxes` (`layout.go:364`): revert to grouping by page and sorting by `(top, x0)` within the page — i.e. drop the `ColID` bucketing. This restores pre-#16795 behavior and matches Python's `_naive_vertical_merge`.
2. `FinalReadingOrderMerge` (`layout.go:249`): drop the `ColID` key (sort `page -> top -> x0`), or stop calling it — mirroring Python, which never calls `_final_reading_order_merge`.
3. Keep `col_id` in `TextMerge` to prevent cross-column horizontal merges (matches Python `_text_merge`).
4. (Optional, for full parity) Insert tables/figures via spatial nearest-neighbor (`min_rectangle_distance`) as Python does, instead of relying on column order.

This preserves correct column handling where it legitimately matters (horizontal-merge boundaries) while restoring top-to-bottom reading order for single-column documents with centered titles. Regenerating the golden snapshot is **not** the right fix — the new order is semantically wrong.

## Environment

- Regression introduced: PR #16795 (merge commit `12787996d`, 2026-07-10)
- Affected files:
  - `internal/deepdoc/parser/pdf/layout/layout.go` (`processPageBoxes`, `FinalReadingOrderMerge`, `AssignColumn`)
  - `internal/deepdoc/parser/pdf/parser.go` (`buildLayout`)
- Python reference: `deepdoc/parser/pdf_parser.py`, `deepdoc/vision/recognizer.py`

Line numbers are as of the current `main`.

zcxGGmu:
### Summary

Closes #16820.

- Keep a page-leading PDF title ahead of lower content when column clustering isolates it into its own column.
- Preserve column-major ordering for genuine multi-column content.
- Cover centered titles, two-column layouts, and non-leading column headings.

### Tests

- bash build.sh --test -count=10 ./internal/deepdoc/parser/pdf/layout
- go vet ./internal/deepdoc/parser/pdf/layout
- git diff --check

The full PDF integration package could not be linked locally because the existing macOS native archives for pdfium, office_oxide, and pdf_oxide are malformed.

### DIFF
--- a/internal/deepdoc/parser/pdf/layout/layout.go
+++ b/internal/deepdoc/parser/pdf/layout/layout.go
@@ -357,10 +357,9 @@ func mergeTwoBoxes(prev, curr pdf.TextBox) pdf.TextBox {
 }
 
 // processPageBoxes vertically merges the boxes of a single page. Boxes are
-// bucketed by column (ColID) first so the merge never crosses columns; within
-// a column the merge runs top→bottom, and columns are emitted in ascending
-// ColID order (leftmost first). This preserves the column-major reading order
-// established by FinalReadingOrderMerge (page → ColID → top → x0).
+// bucketed by column first so merges never cross columns. Titles that precede
+// all non-title content and occupy their own column are moved ahead of the
+// column groups.
 func processPageBoxes(boxes []pdf.TextBox, mh, mw float64, isEnglish bool) []pdf.TextBox {
 	if len(boxes) == 0 {
 		return boxes
@@ -384,7 +383,7 @@ func processPageBoxes(boxes []pdf.TextBox, mh, mw float64, isEnglish bool) []pdf
 		})
 		out = append(out, mergeColumnBoxes(bxs, mh, mw, isEnglish)...)
 	}
-	return out
+	return moveLeadingTitlesFirst(out)
 }
 
 // groupBoxesByCol groups boxes by column id and returns the groups plus the
@@ -442,6 +441,35 @@ func mergeColumnBoxes(sortedBoxes []pdf.TextBox, mh, mw float64, isEnglish bool)
 	}
 
 	return out
+}
+
+func moveLeadingTitlesFirst(boxes []pdf.TextBox) []pdf.TextBox {
+	firstNonTitleTop := math.Inf(1)
+	colsWithNonTitle := make(map[int]struct{})
+	for _, box := range boxes {
+		if box.LayoutType != pdf.LayoutTypeTitle {
+			firstNonTitleTop = math.Min(firstNonTitleTop, box.Top)
+			colsWithNonTitle[box.ColID] = struct{}{}
+		}
+	}
+
+	titles := make([]pdf.TextBox, 0)
+	rest := make([]pdf.TextBox, 0, len(boxes))
+	for _, box := range boxes {
+		_, sharesColumnWithContent := colsWithNonTitle[box.ColID]
+		if box.LayoutType == pdf.LayoutTypeTitle && !sharesColumnWithContent && box.Bottom <= firstNonTitleTop {
+			titles = append(titles, box)
+			continue
+		}
+		rest = append(rest, box)
+	}
+	sort.SliceStable(titles, func(i, j int) bool {
+		if titles[i].Top != titles[j].Top {
+			return titles[i].Top < titles[j].Top
+		}
+		return titles[i].X0 < titles[j].X0
+	})
+	return append(titles, rest...)
 }
 
 // ---- rune-based text helpers (CJK-safe) ----
--- a/internal/deepdoc/parser/pdf/layout/layout_test.go
+++ b/internal/deepdoc/parser/pdf/layout/layout_test.go
@@ -516,6 +516,27 @@ func TestNaiveVerticalMergeNonMerge(t *testing.T) {
 	}
 }
 
+func TestNaiveVerticalMerge_CenteredTitlePreservesPageOrder(t *testing.T) {
+	boxes := []pdf.TextBox{
+		{PageNumber: 0, X0: 250, X1: 450, Top: 50, Bottom: 70, Text: "Document Title", LayoutNo: "title", LayoutType: pdf.LayoutTypeTitle},
+		{PageNumber: 0, X0: 50, X1: 650, Top: 100, Bottom: 112, Text: "First paragraph.", LayoutNo: "body"},
+		{PageNumber: 0, X0: 50, X1: 650, Top: 200, Bottom: 212, Text: "Second paragraph.", LayoutNo: "body"},
+	}
+
+	boxes = AssignColumn(boxes)
+	result := NaiveVerticalMerge(boxes, map[int]float64{0: 12}, map[int]float64{0: 5}, map[int]bool{0: true})
+
+	want := []string{"Document Title", "First paragraph.", "Second paragraph."}
+	if len(result) != len(want) {
+		t.Fatalf("expected %d boxes, got %d", len(want), len(result))
+	}
+	for i, text := range want {
+		if result[i].Text != text {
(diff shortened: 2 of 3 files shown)


zeweihan:
This is a clean fix for the reading-order regression in #16820, and the test coverage (81 lines covering centered titles, two-column layouts, and non-leading column headings) is exactly the right breadth.

From a document-provenance perspective, the key design choice here is correct: **distinguish a page-leading isolated-column title from genuine multi-column body content, and re-order only the former.** A blanket "interleave all columns" fix would have broken legitimate two-column layouts; a blanket "keep column-major" fix would have left the title displacement in place. The targeted approach preserves both reading models.

One observation worth surfacing for the broader PDF pipeline: the centered-title case is the *canonical example* of why layout position alone is insufficient as a structural-identity signal. The title carries "document identity" semantics (it's what a citation anchor like *"as set forth in the Master Services Agreement"* resolves to), but its x-coordinate puts it in a separate column cluster — so a purely geometric ordering model silently demotes the single most load-bearing element. The fix correctly treats it as a special case, but the underlying pattern (a structurally-important element whose *position* doesn't match its *role*) will recur: signature blocks at page bottoms, exhibit labels in margins, footer-defined terms. A durable companion would be a `reading_order_override` signal — not a full re-architecture, just a per-page list of elements whose emitted position was adjusted and why — so that a downstream reviewer asking *"why is the title first when it's geometrically last?"* gets a queryable answer rather than having to reverse-engineer the column logic.

Good that the fix also touches `parser.go` (removing the `−3` lines of dead/duplicated logic) — that's where the regression was seeded, and cleaning it at the source rather than patching downstream emission is the right layer.

xugangqiang:
@zcxGGmu The fix looks good. May I ask your help to provide some e2e testing evidence with a real PDF before & after fix? That means, before fix, the PDF parsing result is not good; after fix, the PDF parsing result should be improved. Thank you.

zcxGGmu:
I captured a real-PDF before/after check using the existing fixture `test/benchmark/test_docs/Doc1.pdf`.

PDF evidence source:
- `Doc1.pdf` is a one-page ReportLab PDF with a centered first-page title.
- First-page blocks were extracted with Poppler `pdftotext -bbox-layout`:
  - title block: x=198.432..413.568, y=71.712..86.512, text=`PDF 1: Purpose of RAGFlow`
  - first body block: x=67.2..544.8, y=104.302..156.477
  - second body block: x=67.2..544.8, y=182.302..234.477

Before fix (`HEAD~1`, c7a623f), running those real-PDF blocks through the layout reading-order stage produced:

```text
ORDER=RAGFlow is an open source Retrieval-Augmented Generation engine | In a typical workflow, you ingest documents into a dataset | PDF 1: Purpose of RAGFlow
```

After fix (`b536341`), the same input produced:

```text
ORDER=PDF 1: Purpose of RAGFlow | RAGFlow is an open source Retrieval-Augmented Generation engine | In a typical workflow, you ingest documents into a dataset
```

Validation commands:
- Before/after evidence command on both commits: `bash build.sh --test -v ./internal/deepdoc/parser/pdf/layout -run TestRealPDFDoc1ReadingOrderEvidence`
- Current committed layout regression suite: `bash build.sh --test -count=10 ./internal/deepdoc/parser/pdf/layout` -> `ok ragflow/internal/deepdoc/parser/pdf/layout 1.507s`

I also tried the full manual PDF package command, but this checkout currently does not build that package locally because unrelated manual test files have compile errors (`truncate redeclared` and missing `common` imports). The evidence above exercises the real PDF blocks at the reading-order stage changed by this PR.

PR #17010, opened 2026-07-16
fix(go-agent): preserve canvas system state across turns

Hz-186:
## Summary

- Preserve request-scoped system variables such as files and user IDs during Canvas execution.
- Persist conversation history, turn counts, and tool memory in the session DSL across turns.
- Parse agent uploads into `sys.files` and align system variable rendering with Python.

## Testing

- `bash build.sh --test ./internal/agent/...`
- `bash build.sh --test ./internal/service/...`

image


yuzhichang:
LGTM

PR #17011, opened 2026-07-16
fix: render readable wiki artifact links

buua436:

Wiki-generated Markdown could expose artifact links with unreadable labels such as `entity/jia-sisters` or invalid `https://artifact/...` URLs.

This change normalizes artifact links, uses the corresponding page title as the display text, and falls back to a readable label derived from the slug.

### Type of change


wangq8 (commented):
## Review: PR #17011 — fix: render readable wiki artifact links (+51/-11, 1 file)

### Summary
A focused bug fix in the wiki knowledge compilation pipeline. Artifact links (links to other wiki pages) generated by the LLM could end up with unreadable labels like `entity/jia-sisters` or invalid `https://artifact/...` URLs instead of clean relative links with proper display text.

### What changed
All changes in `rag/advanced_rag/knowlege_compile/wiki.py`:

| Add | What |
|---|---|
| `_ARTIFACT_MARKDOWN_LINK_RE` | Regex to detect markdown artifact links: `[label](artifact/...)` or `[label](https://artifact/...)` |
| `_artifact_slug(href)` | Parses `artifact/kb_id/slug`, `https://artifact/kb_id/slug`, or bare `kb_id/slug` path formats |
| `_display_text(label, slug)` | Returns page title from `page_titles` dict, or derives readable text from slug (e.g. `entity/jia-sisters` → `Jia Sisters`), or falls back to original label |
| `_markdown_artifact(m)` | Match handler that rewrites artifact markdown links; passes through non-artifact links unchanged |
| `page_titles` parameter | Passed into `_wiki_transform_links` from `wiki_refine_from_plan` |

The processing order is: artifact markdown links → piped wikilinks → simple wikilinks, which is correct since they operate on different syntaxes.

### ✅ What's good

1. **Correct slug parsing.** `_artifact_slug` handles all the formats that could appear:
   - Relative: `artifact/kb_id/slug`
   - Absolute URL: `https://artifact/kb_id/slug`
   - Already-shortened: `kb_id/slug` (without the artifact prefix)
   - Non-artifact URLs → returns `None` and passes through
2. **Good fallback chain for display text.** Page title > derived readable text > original label. The readable text derivation (`slug.rsplit("/", 1)[-1].replace("-", " ").replace("_", " ").strip()`) is pragmatic.
3. **Non-artifact links are preserved.** The regex catches all markdown links, but `_markdown_artifact` passes through non-matching ones with `return m.group(0)`. No false-positive rewrites.
4. **Minimal API surface.** The `page_titles` parameter is optional (defaults to `{}`), so existing callers of `_wiki_transform_links` are unaffected.

### 🔍 Concerns

- **Regex is overly broad.** `_ARTIFACT_MARKDOWN_LINK_RE = r"\[([^\]]+)\]\(([^)]+)\)"` matches **every** markdown link in the content, not just artifact ones. While this is safe (non-artifact links return `m.group(0)` unchanged), it means every single markdown link in the wiki content goes through `urlsplit` + path parsing + string split on every call. For wiki pages with many external links, this is unnecessary overhead. A more targeted regex like `r"\[([^\]]+)\]\((?:https?://)?artifact/[^)]+\)"` would only match artifact links and skip the overhead.

- **URLs with parentheses.** The regex `[^)]+` doesn't handle markdown links containing `)` in the URL (e.g. `[label](https://example.com/path_(more))`). This is a known limitation of simple markdown link regexes and unlikely to affect artifact links specifically, but worth noting for completeness.

- **Edge case: slug vs kb_id overlap.** `_artifact_slug` checks `parts[0] != kb_id_str` to reject non-matching KB IDs. If the slug itself starts with `kb_id_str/`, the logic `parts = parts[1:]` strips the `artifact` prefix and then checks `parts[0] != kb_id_str`. This seems correct but if a slug could legitimately start with the kb_id string, it would be incorrectly rejected. In practice slugs are page names, not kb IDs, so this is fine.

### 🔮 Regression risk

**Very low.** The change:
- Affects only one function in one file
- Has no callers outside the wiki compilation pipeline
- All non-artifact links pass through unchanged
- The `page_titles` parameter defaults to `{}` so existing behavior is preserved

### 🏁 Verdict

**Approve** ✅ — clean, minimal, well-structured. The regex breadth concern is minor and safe; the overhead is negligible for typical wiki pages.

PR #17012, opened 2026-07-16
fix(nlp): detect capitalized table-of-contents headings

chuenchen309:

`remove_contents_table` (`rag/nlp/__init__.py`) fails to detect real-world capitalized table-of-contents headings, so the TOC block is left in parsed documents (this runs from `rag/app/book.py` and `rag/app/laws.py`).

Two issues on the same line:

1. **`IGNORECASE` on the wrong call.** `flags=re.IGNORECASE` was passed to the inner `re.sub(...)` — which only strips whitespace, so case is irrelevant there — instead of to the outer `re.match(...)`. The heading match therefore ran case-sensitively, so `"Contents"`, `"CONTENTS"`, `"Acknowledge"`, etc. never matched. Only exactly-lowercase English headings or the Chinese ones worked.

2. **`"table of contents"` was dead.** The `re.sub` strips all spaces from the input before matching, so a heading like `"Table of Contents"` becomes `"tableofcontents"`, which can never match the space-containing `"table of contents"` alternative.

Reproduction (before the fix):

```python
# 'contents' -> removed (ok),  'Contents' / 'CONTENTS' / 'Acknowledge' / 'Table of Contents' -> NOT removed
```

### Type of change

### Change

Move the `IGNORECASE` flag onto `re.match`, and change the `"table of contents"` alternative to `"tableofcontents"` so it matches the space-stripped input.

Adds `test/unit_test/rag/test_remove_contents_table.py` covering capitalized / uppercase / multi-word English headings, the existing lowercase and Chinese headings (still removed), and non-TOC headings (still kept). Verified the tests fail on `main` and pass with the fix; `ruff check` / `ruff format` clean.

---
*Disclosure: prepared with AI assistance (Claude). The root cause was reproduced and the fix verified locally by me before opening.*

### DIFF
--- a/rag/nlp/__init__.py
+++ b/rag/nlp/__init__.py
@@ -942,7 +942,7 @@ def remove_contents_table(sections, eng=False):
             nonlocal sections
             return (sections[i] if isinstance(sections[i], type("")) else sections[i][0]).strip()
 
-        if not re.match(r"(contents|目录|目次|table of contents|致谢|acknowledge)$", re.sub(r"( | |\u3000)+", "", get(i).split("@@")[0], flags=re.IGNORECASE)):
+        if not re.match(r"(contents|目录|目次|tableofcontents|致谢|acknowledge)$", re.sub(r"( | |\u3000)+", "", get(i).split("@@")[0]), flags=re.IGNORECASE):
             i += 1
             continue
         sections.pop(i)
--- a/test/unit_test/rag/test_remove_contents_table.py
+++ b/test/unit_test/rag/test_remove_contents_table.py
@@ -1,0 +1,76 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for remove_contents_table heading detection.
+
+`re.IGNORECASE` was passed to the inner `re.sub` (whitespace strip) instead of
+the outer `re.match`, so the table-of-contents heading match ran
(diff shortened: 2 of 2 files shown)


xugangqiang:
@chuenchen309  Would you pls help to check if this issue also exist at go side? In future, go side will replace python.

chuenchen309:
Checked — sorry it took me so long to come back to you.

Half of it exists on the Go side, half doesn't. The relevant file is `internal/parser/parser/text_toc.go`.

**The case-sensitivity half is fine there.** Go already has the flag inside the pattern:

```go
var tocHeadingPattern = regexp.MustCompile(`(?i)^(contents|目录|目次|table of contents|致谢|acknowledge)$`)
```

so `Contents` / `CONTENTS` are matched. That was the main defect on the Python side and it does not carry over.

**The dead `table of contents` alternative does carry over.** Line 43 strips all whitespace before matching, while the pattern still spells that alternative with spaces:

```go
normalized := whitespacePattern.ReplaceAllString(strings.SplitN(text, "@@", 2)[0], "")
if !tocHeadingPattern.MatchString(normalized) {
```

so the alternative can never be reached. Running those two expressions verbatim:

```
input                  normalized             matched?
"Contents"             "Contents"             true
"CONTENTS"             "CONTENTS"             true
"Table of Contents"    "TableofContents"      false   <-- should be a TOC heading
"TABLE OF CONTENTS"    "TABLEOFCONTENTS"      false   <-- should be a TOC heading
"table of contents"    "tableofcontents"      false   <-- should be a TOC heading
"Acknowledge"          "Acknowledge"          true
"目 录"                 "目录"                  true
"Introduction"         "Introduction"         false
```

The last two rows are controls: whitespace stripping is what makes `目 录` work, so removing the strip is not the fix. The fix is the same one this PR makes on the Python side — spell the alternative without spaces:

```go
`(?i)^(contents|目录|目次|tableofcontents|致谢|acknowledge)$`
```

Worth noting the doc comment above the pattern says it mirrors `rag/nlp/__init__.py` "after stripping all spaces" — the port was faithful, which is exactly how the unreachable alternative came along with it. If the Go side is going to replace Python, that comment is also where the next person will look, so it may be worth updating together with the pattern.

I haven't opened a PR against the Go tree — happy for you to take it from here.

*AI-assisted; the table above is output from running those two expressions, not inferred.*

xugangqiang:
Please also check the Go side for the same issue and fix it in this PR if present.
Thank you.

yuzhichang (changes requested):
@chuenchen309 Please fix the go part in this same PR.

PR #17013, opened 2026-07-16
fix(ollama): preserve native API base URL in OllamaCV

iwasaki1112:
## What

`OllamaCV.__init__` normalizes `base_url` with `ensure_v1()`, which forces a trailing `/v1` onto the URL. That's correct for the OpenAI-compatible SDK clients used by most other providers in `cv_model.py`, but `OllamaCV` doesn't use one — it hands the URL straight to the official `ollama.Client`, which targets Ollama's native REST API (`{host}/api/generate`, `{host}/api/chat`, ...), not an OpenAI-compatible `/v1` surface.

With `/v1` already baked into `host`, every request lands on `{base_url}/v1/api/generate` and Ollama returns 404, since that path doesn't exist.

## Impact

Any dataflow Parser step that sends an image to an Ollama vision model (PDF/DOCX/Image/etc. with "Disable vision model" off) gets back `**ERROR**: 404 page not found (status code: 404)` from `describe()` / `describe_with_prompt()` / `async_chat()` / `async_chat_streamly()`. Because those methods return the error text as the "description" instead of raising, the literal error string gets ingested as chunk content alongside real document text.

## Fix

Use `base_url` as configured, only stripping a trailing slash — matching how the `ollama-python` README documents `Client(host=...)`.

## Testing

Verified against a live RAGFlow + Ollama deployment (Ollama configured with a valid, `/v1`-free base URL and a vision-capable model):

- Before: Ollama's access log showed repeated `404`s on `POST /v1/api/generate` for every vision call made by a dataflow Parser step; the OCR/description output for each figure contained the literal `**ERROR**: 404 page not found (status code: 404)` string.
- After: the same dataflow run produced `POST /api/generate` requests returning `200 OK`, and the error string no longer appeared in the parsed output.

No automated test is included; this is a one-line behavioral fix to a URL a real backend is sensitive to, not something a mock would catch. Happy to add a unit test around `OllamaCV.__init__` (asserting the `Client` is constructed with the un-suffixed host) if useful.


Lynn-Inf (approved):
LGTM

PR #17015, opened 2026-07-16
fix(string_utils): keep spaces between non-Latin words

chuenchen309:

`remove_redundant_spaces` (`common/string_utils.py`) deletes the space between two non-Latin words. Both cleanup passes use the ASCII-only whitelist `a-z0-9`:

```python
re.sub(r"([^a-z0-9.,\)>]) +([^ ])", r"\1\2", txt, flags=re.IGNORECASE)
re.sub(r"([^ ]) +([^a-z0-9.,\(<])", r"\1\2", txt, flags=re.IGNORECASE)
```

Any non-ASCII letter (Cyrillic, Greek, Arabic, Korean, …) falls into the negated class `[^a-z0-9...]`, so the regex treats it like punctuation and removes the single space:

```python
remove_redundant_spaces("привет мир")   # -> "приветмир"   (WRONG, words merged)
remove_redundant_spaces("Γειά σου κόσμε")  # -> "Γειάσουκόσμε"
remove_redundant_spaces("hello world")  # -> "hello world" (English fine, it's ASCII)
```

This runs over user-facing search-highlight snippets (`rag/nlp/search.py:716`, `api/apps/restful_apis/chunk_api.py:513`), so multilingual highlight text gets adjacent words glued together.

### Type of change

### Change

Replace the ASCII class `a-z0-9` with the Unicode word class `\w` in both passes. For `str` patterns, `\w` matches Unicode letters/digits, so non-Latin letters are treated as word characters and the inter-word space is preserved. The punctuation behavior is unchanged — spaces after `(` / `<` and before `!` etc. are still removed (verified: `"( test"` → `"(test"`, `"world !"` → `"world!"`).

Adds parametrized regression tests for Russian/Greek/Korean/Arabic to the existing `test_string_utils.py`; verified they fail on `main` and pass with the fix, and all existing (non-skipped) tests still pass. `ruff check` / `ruff format` clean.

---
*Disclosure: prepared with AI assistance (Claude). The root cause was reproduced and the fix verified locally by me before opening.*


qinling0210:
## Review: fix is correct, minimal, and well-tested — a few suggestions

Thanks for the fix. I verified the old vs. new regexes over a corpus; the change behaves as intended.

### What's verified
- The 4 advertised languages now preserve spaces (`привет мир`, `Γειά σου κόσμε`, `안녕 세계`, `مرحبا بالعالم`); the new test genuinely fails on the old code.
- **Bonus fixes not mentioned in the PR**: accented Latin `Café au lait` (old → `Caféau lait`), mixed `hello привет` (old → `helloпривет`), and `Привет! Как дела?` (old → `Привет!Какдела?`). Same root cause, nice to see them covered implicitly.
- All currently-passing tests still pass; Latin punctuation behavior (`Amazing !`, `( test`, `< test >`, comma cases) is unchanged.
- Minor side effect: `\w` includes `_` and Unicode digits, so `foo _ bar` now keeps its space (was `foo_bar`). Harmless for the actual callers.

### Suggestion 1: lock in that punctuation cleanup still works on non-Latin text
Consider adding cases where redundant spaces around punctuation are still removed, e.g.:
- `"привет !"` → `"привет!"`
- `"（ 中文 ）"` → `"（中文）"`

This guards against over-preserving and pins the intended boundary behavior.

### Suggestion 2 (important for this repo): Go ports have the same bug
The identical `a-z0-9` whitelist exists in two Go ports:
- `internal/service/nlp/reranker.go:637,641` (`RemoveRedundantSpaces`, applied to rerank doc text at line 146)
- `internal/service/chat_pipeline.go:3720-3721` (marked as "Python parity helpers", mirrored test at `chat_pipeline_test.go:788`)

Note: Go's regexp (RE2) `\w` is **ASCII-only**, so this fix cannot be mechanically ported — it needs `[\p{L}\p{N}_]` or similar. The comment at `reranker.go:554` ("Do NOT call RemoveRedundantSpaces here - it removes spaces between Chinese chars") shows this already bites in practice. Please either port the fix to Go with Unicode classes or at least note the intentional divergence, otherwise Python and Go paths will keep producing different output for non-Latin text.

### Nit
- Test comment alignment is slightly inconsistent (`# Russian` vs `# Greek`); cosmetic only.

Overall: good, focused fix. LGTM once the Go parity question is addressed (or at least documented).

qinling0210 (changes requested):
@chuenchen309 Could you modify go side? We expect we can modify both python and go sides if the issue exists on both sides in a PR. Thanks.

PR #17019, opened 2026-07-17
Feat: Enable the wiki's Markdown editor to navigate to a new Markdown file when a link is clicked.

cike8899:
### Summary

Feat: Enable the wiki's Markdown editor to navigate to a new Markdown file when a link is clicked.

### DIFF
--- a/web/src/components/markdown-editor.tsx
+++ b/web/src/components/markdown-editor.tsx
@@ -7,12 +7,14 @@ interface MarkdownEditorProps {
   content: string;
   onChange?: (content: string) => void;
   readOnly?: boolean;
+  onWikiLinkClick?: (pageType: 'concept' | 'entity', slug: string) => void;
 }
 
 export default function MarkdownEditor({
   content,
   onChange,
   readOnly = false,
+  onWikiLinkClick,
 }: MarkdownEditorProps) {
   const [showSource, setShowSource] = useState(false);
   const [rawContent, setRawContent] = useState(content);
@@ -61,6 +63,7 @@ export default function MarkdownEditor({
             placeholder={readOnly ? '' : 'Start writing...'}
             onToggleSource={toggleSource}
             showSource={showSource}
+            onWikiLinkClick={onWikiLinkClick}
           />
         </div>
         <div
--- a/web/src/hooks/use-knowledge-request.ts
+++ b/web/src/hooks/use-knowledge-request.ts
@@ -591,15 +591,17 @@ export const useFetchArtifactTopicList = (
   };
 };
 
-export function useFetchArtifactPage(artifact: IArtifact | null) {
+export function useFetchArtifactPage(
+  artifact: IArtifact | null,
+  enabled = true,
+) {
   const knowledgeBaseId = useKnowledgeBaseId();
   const pageType = artifact?.page_type ?? '';
   const slug = artifact?.slug ?? '';
 
   const { data, isFetching: loading } = useQuery<IArtifactPage | null>({
     queryKey: ArtifactKeys.detail(knowledgeBaseId, pageType, slug),
-    enabled: !!knowledgeBaseId && !!artifact && !!pageType && !!slug,
-    gcTime: 0,
+    enabled: !!knowledgeBaseId && !!artifact && !!pageType && !!slug && enabled,
     queryFn: async () => {
       const { data } = await getArtifactPage(knowledgeBaseId, pageType, slug);
       return data?.data ?? null;
--- a/web/src/lib/editor/lexical-editor.tsx
+++ b/web/src/lib/editor/lexical-editor.tsx
@@ -40,6 +40,7 @@ interface Props {
   placeholder?: string;
   onToggleSource?: () => void;
   showSource?: boolean;
+  onWikiLinkClick?: (pageType: 'concept' | 'entity', slug: string) => void;
 }
 
 function InitialContentPlugin({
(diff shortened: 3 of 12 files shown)


wangq8 (commented):
## PR Review: #17019 — Wiki Markdown Editor Link Navigation

**Author:** cike8899  
**Files:** 12 files | **+687 / -193**  
**State:** OPEN

---

### What This PR Does

Adds click-to-navigate functionality for internal wiki links (`/concept/{slug}` and `/entity/{slug}`) in the Lexical-based Markdown editor on the wiki compilation page. It also refactors the monolithic `wiki-detail-content.tsx` into smaller components (`WikiDetailHeader`, `WikiDetailEditorPanel`, `WikiDetailToolbar`) and extracts business logic into custom hooks (`useWikiDetailContent`, `useWikiLinkNavigation`).

### ✅ What's Good

1. **Clean component decomposition** — The refactor from a single ~200-line file into focused components (header, editor panel, toolbar) is a welcome improvement for maintainability.

2. **Navigation stack** — `useWikiLinkNavigation` with push/goBack/reset is well-implemented for wiki-style browsing. Truncating the forward stack on push is the correct UX pattern.

3. **Event delegation pattern** — `WikiLinkClickPlugin` uses `rootElement.addEventListener('click', ...)` with proper cleanup, which is the right approach for Lexical plugins.

4. **Multiple link format support** — `parseWikiLinkHref` handles both `artifact/{id}/{type}/{slug}` and `{type}/{slug}` formats.

5. **No-op on same-link click** — Clicking a link pointing to the currently displayed page is correctly a no-op, preventing unnecessary fetch cycles.

### ⚠️ Issues & Concerns

#### 1. Missing tests for a pure utility function
`parse-wiki-link.ts` is a pure function — zero dependencies, deterministic output. With ~50 lines of code and regex matching, this is a prime candidate for unit tests. No test file is included. **Recommendation:** Add unit tests covering at least the documented formats and edge cases (empty href, malformed URLs, slugs with special characters, etc).

#### 2. `gcTime: 0` removed without explanation
The previous code had `gcTime: 0` on `useFetchArtifactPage`. This was removed in favor of the new `enabled` parameter. For a navigation context where users may go back to previously viewed pages, keeping cache around is actually beneficial. However, this is a behavioral change that should be intentional. If the intent was to allow caching, that's fine — but it should be mentioned.

#### 3. Type mismatch — `displayedArtifact` may not fully satisfy `IArtifact`
In `useWikiDetailContent`:
```ts
const displayedArtifact = currentEntry
  ? { slug: currentEntry.slug, title: currentEntry.title, page_type: currentEntry.pageType }
  : selectedArtifact;
```
This object shape is passed to `WikiDetailHeader` which types it as `IArtifact | null | undefined`. If `IArtifact` has required fields beyond `slug`, `title`, and `page_type`, this will either fail at runtime or TypeScript won't complain if those fields are optional. **Recommendation:** Either use `Pick<IArtifact, 'slug' | 'title' | 'page_type'>` throughout, or make `WikiDetailHeader` accept a narrower interface explicitly.

#### 4. Race condition dependency in the dual-effect pattern
The two `useEffect` hooks in `useWikiDetailContent` (one syncs title, one resets stack) both depend on `selectedArtifact?.slug` and `currentEntry?.slug`. The PR comment explains that `push()` is called before `onSelectArtifact()` in the click handler, so `currentEntry.slug` already matches the artifact slug by the time effects run. This relies on React 18's automatic batching behavior. It's correct today but is a subtle coupling that could break with future React changes. **Recommendation:** Consider merging the two effects or using a single state reducer to make the dependency explicit.

#### 5. URL-encoded slugs not handled in `parseWikiLinkHref`
If a wiki slug contains characters that are percent-encoded in the href (e.g., `%20` for spaces), the regex captures the raw encoded string. This could cause a navigation to a non-existent slug. **Recommendation:** Add `decodeURIComponent()` on the captured slug.

#### 6. Version view ignores link clicks
`handleMarkdownLinkClick` bails out when `isVersionView` is true. While this makes sense (viewing a historical version shouldn't trigger navigation), it silently drops the click. Consider showing a tooltip or visual indicator that links are disabled in version-diff mode.

#### 7. `updateCurrentTitle` returns early when title already matches
In `useWikiLinkNavigation.updateCurrentTitle`, the function returns `prev` unchanged if `entry.title === title`. This is a good optimization but it means if the title hasn't been fetched yet (`entry.title === ''`) and the data comes back with an empty title, the entry remains untitled. This is likely fine in practice since `pageData.title` would usually be non-empty, but worth noting.

### 🔍 Minor Style / Quality Notes

- The new `enabled` parameter on `useFetchArtifactPage` is gated on `IArtifact | null` type union, and `enabled` defaults to `true` — good backward compatibility.
- `WikiLinkClickPlugin` returns `null` (correct pattern for a non-rendering Lexical plugin).
- Event listener cleanup uses the stored `rootElement` reference which is stable for the hook's lifecycle — no stale closure issue.
- Navigation stack uses `currentIndex` correctly and never goes below 0.

### 🏁 Verdict

**Solid feature addition with a well-structured refactor.** The core logic is sound, the plugin pattern is correct, and the navigation UX is thoughtful (back button preserves previous entry's title, same-link clicks are no-ops, forward stack is truncated).

The main gaps are lack of tests (especially for the parser), a type looseness in `displayedArtifact`, and the missing URL decoding in the parser — all easy fixes. None are blockers for merging, but items 1-3 should be addressed before release.

PR #17024, opened 2026-07-17
fix: ASR/VLM/TTS models not selectable in default model settings

euvre:
### Summary

This PR fixes three issues that prevented ASR, VLM, and TTS models from being selected as default models in the system settings page.

**1. Model type name mismatch (ASR & VLM)**

The Go backend's `HumanReadable()` returned legacy names (`"speech2text"`, `"image2text"`) while the frontend expects canonical names (`"asr"`, `"vision"`). The `GET /api/v1/models` endpoint returned `model_type: ["speech2text"]` but the frontend `buildModelTree` filtered by `['asr']`, so the models never appeared in the dropdown.

The Python backend's `get_model_type_human()` already returned canonical names. This PR aligns the Go backend to match.

`ModelTypeFromString()` now also accepts `"asr"` (in addition to the legacy `"speech2text"`) for bidirectional compatibility, since the factory catalog (`conf/models/*.json`) uses the canonical `"asr"`.

**2. Stale database records (ASR)**

Before the `ModelTypeFromString` fix, models with type `"asr"` in the factory catalog were stored with `model_type = 0` in `tenant_model`. This PR adds a runtime auto-repair step in `ListTenantAddedModels` that re-derives the correct bitmask from the factory catalog and persists it, so existing stale records self-heal on the next list request without requiring users to re-add models.

**3. Early return blocking TTS (and OCR) default models**

`ListTenantDefaultModels` had an early `return result, nil` when `OCRID` was empty. Since most users have no OCR model configured (OCR defaults to the built-in deepdoc), the function returned before reaching the TTS block. As a result, a TTS default model could be set but never read back, making it appear unselected after saving. This PR makes the OCR and TTS blocks independently guarded.

### DIFF
--- a/internal/entity/types.go
+++ b/internal/entity/types.go
@@ -20,7 +20,7 @@ import "strings"
 
 // ModelType represents the type of model as a bitmask, matching Python's
 // ModelTypeBinary enum (common/constants.py).
-// Bit flags (LSB->MSB): 1=chat, 2=embedding, 4=speech2text, 8=image2text, 16=rerank, 32=tts, 64=ocr.
+// Bit flags (LSB->MSB): 1=chat, 2=embedding, 4=asr, 8=vision, 16=rerank, 32=tts, 64=ocr.
 // A model can belong to multiple types simultaneously.
 type ModelType int
 
@@ -91,10 +91,10 @@ func (mt ModelType) HumanReadable() []string {
 		parts = append(parts, "embedding")
 	}
 	if mt.Has(ModelTypeSpeech2Text) {
-		parts = append(parts, "speech2text")
+		parts = append(parts, "asr")
 	}
 	if mt.Has(ModelTypeImage2Text) {
-		parts = append(parts, "image2text")
+		parts = append(parts, "vision")
 	}
 	if mt.Has(ModelTypeRerank) {
 		parts = append(parts, "rerank")
@@ -115,11 +115,9 @@ func ModelTypeFromString(s string) ModelType {
 		return ModelTypeChat
 	case "embedding":
 		return ModelTypeEmbedding
-	case "speech2text":
+	case "asr", "speech2text":
 		return ModelTypeSpeech2Text
-	case "image2text":
-		return ModelTypeImage2Text
-	case "vision":
+	case "vision", "image2text":
 		return ModelTypeImage2Text
 	case "rerank":
 		return ModelTypeRerank
--- a/internal/service/model_service.go
+++ b/internal/service/model_service.go
@@ -1159,6 +1159,34 @@ func (m *ModelProviderService) ListTenantAddedModels(userID, ownerTenantID, mode
 		return nil, common.CodeServerError, err
 	}
 
+	// Repair records that have model_type == 0. Before the ModelTypeFromString
+	// fix that added "asr" support, models whose factory catalog used the
+	// canonical name "asr" (instead of the legacy "speech2text") were stored
+	// with model_type = 0. Re-derive the correct bitmask from the factory
+	// catalog and persist it so the model shows up in typed queries.
+	providerManager := dao.GetModelProviderManager()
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
## PR Review: #17024 — fix: ASR/VLM/TTS models not selectable in default model settings

**Author:** euvre  
**Files:** 3 Go files | **+63 / -34**  
**State:** OPEN

---

### Summary

This PR fixes three distinct bugs that prevented ASR (speech2text), VLM (image2text), and TTS models from working in the default model settings on the system settings page. Each bug has a clear root cause and a minimal fix.

---

### Bug-by-Bug Analysis

**Bug 1 — Model type name mismatch (ASR & VLM)**
- **Root cause:** `HumanReadable()` returned `"speech2text"` / `"image2text"` while the frontend's `buildModelTree` filters by `['asr']` / `['vision']`.
- **Fix:** Updated `HumanReadable()` to emit canonical names and `ModelTypeFromString()` to accept both legacy and canonical names.
- ✅ **Correct.** Bidirectional compatibility ensures old factory catalog entries still resolve.

**Bug 2 — Stale DB records (ASR)**
- **Root cause:** Models with `"asr"` in the factory catalog were stored with `model_type = 0` before Bug 1 was fixed.
- **Fix:** Added a self-healing loop in `ListTenantAddedModels` that re-derives the bitmask from the factory catalog and persists it.
- ✅ **Thoughtful approach.** No manual migration or re-add needed — stale records self-heal on next read.
- ⚠️ **Minor:** The DB update error is silently discarded (`_ = ...`). If the write fails, the record stays broken and the retry happens on the next list call. This is acceptable for a self-healing path, but **consider logging the error** so operators can identify persistent failures.

**Bug 3 — Early return blocking TTS/OCR**
- **Root cause:** `ListTenantDefaultModels` had `if OCRID == "" { return result, nil }` before the TTS block. Since most tenants have no OCR model, the function returned early before reaching TTS processing.
- **Fix:** Made OCR and TTS blocks independently guarded by their own `if` checks.
- ✅ **Clean fix.** The diff speaks for itself — the early return was clearly wrong.

---

### ⚠️ Issues & Concerns

#### 1. Constant names now misaligned with canonical strings
The Go constants `ModelTypeSpeech2Text` and `ModelTypeImage2Text` and their `Has()`/`Set()` methods still use legacy naming, but the canonical strings are now `"asr"` and `"vision"`. This is cosmetic (no functional impact since the constants are internal), but could confuse new developers. **Recommendation:** Rename constants in a follow-up cleanup commit, e.g. `ModelTypeASR` / `ModelTypeVision`.

#### 2. Silent error in auto-repair write
```go
_ = m.modelDAO.UpdateByID(rec.ID, map[string]interface{}{"model_type": int(combinedType)})
```
If this write fails, the record remains broken. The retry-on-next-list behavior means it will eventually recover under normal operation, but a silent failure makes troubleshooting harder. **Recommendation:** At minimum add `log.Warn(...)` so a failing write is visible in logs.

---

### Minor Quality Notes

- The bit-flag comment in `types.go` (line 23) was correctly updated to reflect the canonical names — good catch.
- `factoryModelTypeName()` in `tenant.go` still serves a purpose as a translation layer for legacy factory catalog entries — not dead code.
- The `providerManager` variable was moved higher in `ListTenantAddedModels` (line 1165 vs line 1202) because the repair loop now needs it. Clean reorganization.

### Verdict

**Well-structured, clearly explained, minimal-impact bugfix PR.** Three separate bugs, each with a clean, targeted fix. No regressions introduced. The auto-repair for stale DB records is particularly well thought out.

The only nits are:
1. Log the auto-repair write failure instead of silencing it
2. Consider renaming the `ModelTypeSpeech2Text` / `ModelTypeImage2Text` constants in a follow-up

**Not blocking merge**, but items 1 is worth fixing before the next release.

Lynn-Inf (approved):
LGTM

PR #17025, opened 2026-07-17
fix(model-provider): replace per-card auto-save with batch Save-all

dcc123456:
### Summary

fix(model-provider): replace per-card auto-save with batch Save-all


wangq8 (commented):
## Review: PR #17025 — fix(model-provider): replace per-card auto-save with batch Save-all (+1314/-1271, 14 files)

### Summary
A major UX refactor of the model provider settings page. **Removes the per-card auto-save** (blur/change-driven saves per instance card) and replaces it with a **single "Save all" button** at the top of the provider pane. Draft cards no longer have the two-stage "save name first, then edit fields" gating — all fields are editable from the start, and the parent drives persistence through an imperative ref API.

### Author: @dcc123456

### What changed (14 files)

| File | Δ | Key changes |
|---|---|---|
| `setting-model/index.tsx` | +150/-68 | Added `cardRefs` registry, `handleSaveAll` batch save, `saving`/`canSave` state; removed `handleDraftSave`/`handleNameSaved` |
| `instance-card/hooks.tsx` | +266/-374 | Removed `useDraftAutoSave`, `useSaveInstanceName`, added `verifyTransform` to `useVerifyProvider`, new `useInstanceSaveState` w/ baseline dirty tracking |
| `instance-card/interface.ts` | +66/-16 | Added `ProviderInstanceCardRef` (`validate`, `getSavePayload`, `markSaved`), `InstanceSavePayload` type |
| `bedrock-instance-card.tsx` | +172/-283 | Converted to `forwardRef`, removed auto-save / blur-save / watch-save, added baseline dirty tracking via `baselinePayloadRef` |
| `somark-instance-card.tsx` | +184/-308 | Same pattern as bedrock card |
| `provider-instance-card.tsx` | +148/-94 | Converted to `forwardRef`, dispatches to bedrock/somark/component sub-cards, exposes imperative ref |
| `saved-mode-card.tsx` | +101/-21 | Removed `handleFieldsBlur`, added inline rename (double-click → input → Enter/blur commit) |
| `draft-mode-card.tsx` | +35/-43 | Removed fieldset lock & `handleSaveName` — all fields editable from start |
| `provider-header-bar.tsx` | +60/-7 | Added batch Save button with spinner |
| `instance-name-section.tsx` | +7/-29 | Removed Save button, Enter-to-save, helper text |
| `models-section/hooks.ts` | +88/-16 | Added `updateDraftModel` callback, removed `onBlurSuppressRef` from ModelsSection props |
| `models-section/index.tsx` | +35/-4 | Removed `onBlurSuppressRef` threading, keep API-key gate |
| `locales/en.ts` / `zh.ts` | +2/-8 | Added `saveAll`, removed `instanceNameSaveTip`/`instanceNameSavePrompt`/`instanceNameLockedHint` |

### ✅ What's good

1. **Clear UX improvement.** The per-card auto-save was fragile — debounce timing, blur vs click race conditions, portal-based controls not triggering blurs, and failed saves silently lost data. A single explicit "Save all" button is far more predictable.

2. **Clean imperative ref API.** `ProviderInstanceCardRef` with `validate()`, `getSavePayload()`, `markSaved()` gives the parent full control while keeping card logic encapsulated. The type is well-defined in `interface.ts`.

3. **Dirty tracking is solid.** Using `JSON.stringify(baselinePayloadRef)` vs current payload to skip redundant saves is simple and correct. The baseline is rebuilt on details load.

4. **Two-stage gating removed.** The name-save-first pattern was confusing UX — "save the name to unlock the form." Now all fields are editable immediately. This is a better UX pattern.

5. **Inline rename for saved cards.** Double-click the name to edit it, Enter/blur to commit, Escape to cancel. Clean and expected behavior.

6. **Sequential batch save with error handling.** `handleSaveAll` stops on first failure, preserving the error in the form UI. No silent data loss.

7. **Cross-card consistency.** Bedrock, SoMark, and generic cards all follow the same pattern now — `forwardRef` with `validate` + `getSavePayload` + `markSaved`.

### 🔍 Concerns

#### 🔴 Functional issues flagged by CodeRabbit (11 items)

Several of these merit attention before merge:

- **ModelsSection API-key gate is not reactive** — `DynamicForm` changes don't trigger re-render of `ModelsSection`, so if a draft is mounted with an empty key, typing a key later won't auto-fetch available models. This is a real UX issue.

- **Collapsed-card rename safety** — If a collapsed card hasn't fetched its instance details yet, the rename `getSavePayload` could send a sparse payload that overwrites credentials/models. A collapsed rename should either fetch details first or route through a sparse update.

- **`getSavePayload` for Bedrock/SoMark may not include the edited instance name** — The inline rename updates `editedInstanceName`, but `buildPayload` inside `getSavePayload` references `instance.instance_name` not `editedInstanceName` for saved cards. Need to verify the rename value flows into the save payload.

- **Sequential batch API calls** — One `await` per dirty card. If a user has 10 dirty cards and the 3rd fails, cards 4-10 are never saved. The user must fix the error and re-save everything. Consider collecting failures and continuing, or at least adding a log.

#### 🟡 Code quality / architecture

- **1314 lines changed is high risk.** This is effectively a rewrite of the instance card module. Given that auto-save and the draft lifecycle were central to the prior architecture, the surface for regressions is broad.

- **`canSave` is a conservative proxy.** `draftIds.length > 0 || instances.length > 0` means the Save button is enabled even when no card is actually dirty. This is fine UX-wise (saving a clean state is a no-op) but could be misleading. The button label and disabled state could be more precise.

- **Inline rename accessibility.** The rename is triggered by `onDoubleClick` only — no keyboard-triggered way to start a rename. Consider also adding a right-click context menu or an edit button for keyboard users. (CodeRabbit flagged this.)

- **`useVerifyProvider` now accepts a `verifyTransform`** — This is good for providers like OpenDataLoader with non-standard field names. But it's only wired for the generic card path; Bedrock/SoMark cards don't pass one. This is consistent since those cards build their own verify payloads, but worth verifying there's no regression.

#### 🟠 Locales cleanup

Removing `instanceNameSaveTip`, `instanceNameSavePrompt`, `instanceNameLockedHint` is correct (the two-stage name gate is gone). There are no remaining references to these keys.

#### 🟣 Test coverage

I don't see test changes in the diff. Given the extent of the refactor, adding or updating tests for the batch save flow would be valuable — specifically:
- Validate that `getSavePayload` returns `null` for clean saved cards
- Validate that `getSavePayload` returns a payload for dirty cards
- Validate the batch `handleSaveAll` flow with multiple cards

### 🔮 Regression risk

**Moderate.** The refactor is thorough and the new architecture is cleaner, but:
- Every card type (generic, Bedrock, SoMark) was substantially rewritten
- The auto-save → batch-save transition touches every user interaction path (edit, verify, delete, collapse/expand)
- 11 CodeRabbit findings remain unaddressed, including functional issues

I would recommend addressing the functional CodeRabbit items (especially the reactive API-key gate and collapsed-card rename safety) before merging. The architectural direction is strong; the execution needs a final polish pass.

### 🏁 Verdict

**Approve with reservations** ⚠️ — the direction is correct and the new architecture is much cleaner than the auto-save approach. However, the 11 CodeRabbit findings include real functional concerns (non-reactive ModelsSection key gate, collapsed-rename safety, rename value propagation) that should be addressed before merge.

PR #17029, opened 2026-07-17
Fix: set Builtin models as default model

Lynn-Inf:
### Summary

As title, allow to pass `check_availablity` and handle ID when set Buitin models as default model.


wangq8 (commented):
## Review: PR #17029 — Fix: set Builtin models as default model

**Author:** Lynn-Inf | **Files changed:** 3 | **+46/-26**

### Summary

This PR allows setting Builtin models (TEI embeddings, deepdoc) as the default model for a tenant by properly handling availability checks and model ID resolution for provider-less (Builtin) entries.

### Analysis

**model_service.go — ResolveModelID changes (good):**
- Adds a `GetBuiltinEmbeddingModel` check before returning empty string for Builtin embedding models. This validates the model actually exists in the local configuration before falling through.
- Swaps the condition order (`providerName == "Builtin"` first) making the intent clearer.
- Slight concern: If `GetBuiltinEmbeddingModel` returns nil for a known model name (e.g., a typo in env var), the error surfaces late — only when resolving. Consider surfacing this earlier (startup validation?).

**tenant.go — checkModelAvailable bypasses (good):**
- Three new static bypasses mirroring Python's `_check_model_available`. These are necessary because Builtin/infiniflow providers have no tenant_model_provider rows.
- The TEI embedding bypass via COMPOSE_PROFILES + TEI_MODEL is correct.
- The deepdoc bypass is hardcoded to "infiniflow" @ "default" @ "deepdoc". This is fine since deepdoc is a static local model.

**tenant.go — SetTenantDefaultModels Builtin handling (critical correctness fix):**
- Before: all code paths tried to look up tenantModelID from DB, which would fail for Builtin provider since no tenant_model rows exist.
- After: explicitly sets tenantModelID = nil for Builtin provider, skipping DB lookup. This is the core bug fix — mirrors the Python resolve_model_id() returning None for Builtin.

**wiki.py — formatting noise:**
- Two lines simply unwrapped (line continuation removed). Pure style change. These formatting changes are unrelated to the PR topic and slightly distract from review. Recommend splitting formatting-only changes into a separate commit.

### Concerns

1. **wiki.py changes are out of scope:** They are cosmetic (line wrapping) and unrelated to the Builtin model fixes. Minor, but good practice to keep PRs focused.

2. **Missing test coverage:** Is there an existing test or should one be added for SetTenantDefaultModels with a Builtin provider? The old code paths would crash with a nil pointer / DB error — a test would prevent regression.

3. **checkModelAvailable — TEI/env coupling:** The bypass relies on COMPOSE_PROFILES and TEI_MODEL env vars being set. If these are stale or misconfigured, a Builtin embedding model would still fail the check. This is acceptable (following Python convention), but should be documented.

### Verdict

The core fix is correct and well-structured, matching the Python behavior. The Builtin provider handling in SetTenantDefaultModels resolves a real bug where setting a Builtin embedding as the default would panic/error. **Approved** — with a request to either remove or justify the unrelated wiki.py formatting changes.

PR #17031, opened 2026-07-17
Fix: disable Ollama thinking for vision figure descriptions

iwasaki1112:
## Summary
- `OllamaCV.describe()` / `describe_with_prompt()` call the Ollama Python client's `generate()` without `think=False`. Ollama only honors `think` as a top-level kwarg (not inside `options`), so Qwen3-family vision models emit a full reasoning trace before the actual figure description.
- Added `think=False` to both calls.
- The `think` parameter on `generate()` needs ollama-python>=0.5.1; bumped the pinned lower bound from `>=0.5.0` accordingly.

## Test plan
- [x] Verified on a live deployment (Ollama + qwen3.5:9b as the vision model): before the fix, a single figure description took ~6 minutes and ~2800 tokens (dominated by an unsuppressed thinking trace).
- [x] After the fix, the same figure-description call consistently took 10-60s.
- [x] Confirmed across a full AI-OCR parsing run over an 86-page PDF with 250+ figures — no errors, all figures processed at the faster pace.


6ba3i (commented):
The Python fix looks good to me. `think=False` is applied at the right level and the dependency bump makes sense.

The one thing I think we still need is Go parity. Current `main` now has the same parser-side vision path in Go, but Ollama only serializes `Thinking` when it is `true`, so `Thinking: false` is effectively the same as leaving it unset. That means Qwen3/Qwen3.5 can still think by default there.

I’d update the Go Ollama driver to serialize any non-nil `Thinking` value, and have the vision invoker explicitly pass `Thinking: false`. The same should apply to streaming.

Also worth regenerating `uv.lock` after the minimum Ollama version bump.

Other than that, LGTM

6ba3i:
Hello @iwasaki1112 could you please take a look at my comment fix the missing go implementation ?

6ba3i:
overlaps on #19529

PR #17034, opened 2026-07-17
fix(go-agent): filter SSE done sentinel from model stream

Hz-186:
## Summary

- Filter the `[DONE]` transport sentinel from Eino model streams
- Add regression coverage for stream messages and callbacks

## Testing

- `bash build.sh --test ./internal/entity/models`
- `bash build.sh --test ./internal/agent/component`
- `bash build.sh --test -run 'Test(AgentChatCompletions_Stream|RunAgent_Stream)' ./internal/handler`


yuzhichang:
LGTM

PR #17035, opened 2026-07-17
fix: preserve chunks when updating compilation templates

buua436:

Updating a document’s compilation template no longer clears its existing chunks. The parser configuration is updated without triggering a reparse unless the parser or pipeline actually changes.

### Type of change


wangq8 (commented):
## Review: PR #17035 — Fix: preserve chunks when updating compilation templates

**Author:** buua436 | **Files changed:** 1 (document_api.py) | **+3/-21**

### Summary

This PR fixes a bug where updating a document's knowledge compilation template group would unnecessarily clear existing chunks and trigger a reparse. The change removes the `_compilation_template_group_id_changed` function and the associated reparse-triggering logic.

### Analysis

**What was removed:**
1. The `_compilation_template_group_id_changed()` helper that compared old vs new template group IDs.
2. The downstream code that called `reset_document_for_reparse()` when the template group IDs changed and the parser wasn't being swapped.

**What was kept:**
- `_normalize_parser_config_compilation_template_group_ids()` is still called (side-effect: normalizes the IDs), but its boolean return value is now discarded since the caller no longer needs to know if anything changed.

### Concerns

1. **Stale chunk semantics:** If a user changes the compilation template group for a document, the existing chunks may reference the old template group. Are these chunks still valid? The PR title says "preserve chunks" — but if the chunks reference stale template groups, downstream compilation queries could behave incorrectly. Could you confirm that compilation template groups are resolved at runtime rather than baked into the chunk data at parse time? If they are runtime-resolved, this change is correct.

2. **API behavior change:** Previously, updating a parser config (even without changing the parser) would trigger reset_document_for_reparse if the template group IDs changed. Now it silently updates the config without re-initializing chunks. This is a behavioral change — existing integrations that relied on the reparse side-effect may need updating. This should be noted in release notes.

3. **Minor:** The return value of `_normalize_parser_config_compilation_template_group_ids` is now unused. Consider simplifying the return type in a follow-up cleanup if it is only called from this one site.

### Verdict

The fix itself is reasonable and addresses the user-facing bug (unnecessary chunk clearing). The risk is low since compilation templates are typically user-configurable metadata that should not affect low-level chunk storage. **Approved** with the above comments.

PR #17036, opened 2026-07-17
feat: add chat-level TTS and ASR endpoints for Go API server

euvre:
### Summary

This PR ports the Python `/chat/audio/*` endpoints to the Go API server, adding chat-level TTS and ASR (transcription) support.

- `POST /api/v1/chat/audio/speech`: Converts text to speech using the tenant's default TTS model and returns a streaming `audio/mpeg` response.
- `POST /api/v1/chat/audio/transcription`: Transcribes an uploaded audio file using the tenant's default ASR model. Supports both a single JSON response and SSE streaming via `stream=true`.

The implementation reuses `ModelProviderService.GetTenantDefaultModelByType` to resolve the tenant's default models and delegates synthesis/transcription to the existing model drivers.

### DIFF
--- a/internal/handler/chat_audio.go
+++ b/internal/handler/chat_audio.go
@@ -12,7 +12,6 @@
 //  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
-//
 
 package handler
 
@@ -34,17 +33,24 @@ import (
 	modelModule "ragflow/internal/entity/models"
 )
 
-// ChatAudioSpeechRequest is the request body for POST /api/v1/chat/audio/speech.
-type ChatAudioSpeechRequest struct {
+const (
+	// chatAudioSpeechMaxBodyBytes caps the JSON body of the TTS endpoint.
+	chatAudioSpeechMaxBodyBytes int64 = 1 << 20 // 1 MiB
+	// chatAudioUploadMaxBytes caps the multipart body of the transcription endpoint.
+	chatAudioUploadMaxBytes int64 = 64 << 20 // 64 MiB
+)
+
+// chatAudioSpeechRequest is the request body for POST /api/v1/chat/audio/speech.
+type chatAudioSpeechRequest struct {
 	Text string `json:"text" binding:"required"`
 }
 
-// ttsSegmentSplitRegex mirrors Python's re.split(r"[，。/《》？；：！\n\r:;]+", text)
-// used by chat_api.py's /chat/audio/speech endpoint.
+// ttsSegmentSplitRegex splits TTS input into synthesis segments on CJK
+// punctuation (，。/《》？；：！), ASCII ':' and ';', and newlines.
 var ttsSegmentSplitRegex = regexp.MustCompile("[，。/《》？；：！\\n\\r:;]+")
 
-// ChatAudioSpeech converts text to speech using the tenant's default TTS model.
-// It returns a streaming audio/mpeg response aligned with the Python endpoint.
+// ChatAudioSpeech converts text to speech using the tenant's default TTS model
+// and streams the concatenated MP3 segments as an audio/mpeg response.
 func (h *ChatHandler) ChatAudioSpeech(c *gin.Context) {
 	user, errorCode, errorMessage := GetUser(c)
 	if errorCode != common.CodeSuccess {
@@ -52,7 +58,9 @@ func (h *ChatHandler) ChatAudioSpeech(c *gin.Context) {
 		return
 	}
 
-	var req ChatAudioSpeechRequest
+	c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, chatAudioSpeechMaxBodyBytes)
+
+	var req chatAudioSpeechRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
 		common.ResponseWithCodeData(c, common.CodeArgumentError, nil, err.Error())
 		return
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## Review: Add chat-level TTS & ASR Go endpoints

Overall a clean, well-structured PR. The new `chat_audio.go` handler reuses the existing model resolution and driver infrastructure properly. Below are findings organized by severity.

---

### 🐛 Bugs

**1. SSE stream missing `[DONE]` event on success** (internal/handler/chat_audio.go:172-196)

In the streaming transcription path, if `TranscribeAudioWithSender` succeeds without ever sending `"[DONE]"` through the sender callback, the SSE stream never emits a completion event. The client has no way to know transcription is finished and may hang.

**Fix:** After `TranscribeAudioWithSender` returns nil (success), send an explicit `{"event":"done","text":"[DONE]"}` event unless one was already sent by the driver.

**2. `truncateString` can produce invalid UTF-8** (internal/handler/chat_audio.go:221-225)

`s[:n]` is byte-wise slicing. A string of 22 Chinese characters = 66 bytes; `str[:64]` splits the last character in half, producing invalid UTF-8. This is used for zap logging and could break JSON encoding of log output.

**Fix:** Use `[]rune(s)` for character-aware truncation, or use `utf8.ValidString` / `strings.ToValidUTF8` before logging.

---

### 🔒 Security

**3. Request body not bounded in either endpoint** (internal/handler/chat_audio.go:62-64, 119)

- `ChatAudioSpeech`: No `http.MaxBytesReader` on the JSON body before `ShouldBindJSON`. A large payload could exhaust memory.
- `ChatAudioTranscription`: No `http.MaxBytesReader` before `FormFile`. A malicious upload could fill disk/tmp.

**Fix:** Add `http.MaxBytesReader` wrapping, matching the existing pattern in `agent_upload.go:88` and `mcp_server.go:80`.

**4. User text logged in TTS warning** (internal/handler/chat_audio.go:87-89)

`zap.String("segment", truncateString(seg, 64))` logs user-provided text. Even truncated, this may contain PII. The existing `chat_session.go` SSE handler does not log user message content.

**Fix:** Log only the segment index/length alongside the error, not the text content.

---

### ⚠️ Reliability

**5. Response headers committed before data is available** (internal/handler/chat_audio.go:76-80)

`WriteHeader(http.StatusOK)` runs before the synthesis loop. If all segments are empty or all synthesis calls fail, the client gets a 200 OK with zero bytes — indistinguishable from silence.

**Fix (optional):** Defer `WriteHeader` until the first successful audio chunk, or at minimum track success and log a warning when zero audio was produced.

**6. `X-Accel-Buffering: no` missing in SSE transcription** (internal/handler/chat_audio.go:165-170)

The speech endpoint sets this header but the SSE transcription path does not. Behind nginx, this will cause buffering of SSE events.

**Fix:** Add `c.Header("X-Accel-Buffering", "no")` in the `streamMode` branch, matching `chat_session.go:198` and `search.go:408`.

---

### 🧹 Style & Maintainability

**7. `ChatAudioSpeechRequest` should be unexported** (internal/handler/chat_audio.go:37-39)

It is only used within this handler file; rename to `chatAudioSpeechRequest`.

**8. Comments reference Python migration history** (internal/handler/chat_audio.go:42-47, 72, 102-103)

Phrases like "mirrors Pythons ..." and "matching Pythons ..." describe what was done in the old code rather than documenting current Go behavior. Describe the delimiters, headers, and extensions directly.

**9. `allowedAudioExtsList()` builds a sorted list on every call** (internal/handler/chat_audio.go:207-213)

This is called on every transcription request. The set is static — use `sync.Once` or a `var` initializer to compute it once.

---

### ℹ️ Notes

- The PR correctly uses `disableWriteDeadlineForSSE` for the streaming path — consistent with existing handlers.
- Temp file cleanup via `defer os.Remove` is correct.
- Extension whitelist is a good security practice. Consider if `.wma` (a proprietary/codec-limited format) is intentional, or if it should be removed.

CodeRabbit already flagged items 7 and 8. Items 1-6 and 9 are additional findings from this review.

wangq8 (commented):
## Re-review: PR #17036 — Chat-level TTS & ASR Go endpoints

Previous reviews (CodeRabbit x2 + wangq8) raised 10 distinct findings. I reviewed the merged code to check which were resolved.

### ✅ Issues fixed since first round

| # | Finding | Status |
|---|---------|--------|
| 1 | SSE stream missing `[DONE]` on success | ✅ Fixed — `doneSent` flag + fallback emission |
| 2 | `truncateString` produced invalid UTF-8 on CJK | ✅ Fixed — function removed, user text no longer logged |
| 3 | Request body not bounded (both endpoints) | ✅ Fixed — `http.MaxBytesReader` on both endpoints |
| 4 | User PII logged in TTS warning | ✅ Fixed — now logs only `segmentIndex`/`segmentLen` |
| 5 | Response headers committed before data available | ✅ Fixed — `headerWritten` flag defers until first chunk |
| 6 | `X-Accel-Buffering: no` missing in SSE path | ✅ Fixed — set in streamMode branch |
| 7 | Comments referencing Python migration history | ✅ Fixed — clean Go-behavior comments |
| 8 | `ChatAudioSpeechRequest` should be unexported | ✅ Fixed — renamed to `chatAudioSpeechRequest` |
| 9 | `chatAudioAllowedExtsList` built on every call | ✅ Fixed — computed once via `var` initializer |
| 10 | `ParseMultipartForm` memory threshold | 🟢 **Consistent** — uses same pattern as `agent_upload.go` (same limit for MaxBytesReader + ParseMultipartForm). Not a concern. |

All 10 findings from prior reviews have been addressed. No regressions introduced.

### 🔍 Additional observations

**Structure — clean.** Follows existing handler patterns (`GetUser` → setup → model resolution → response). Error codes match `common.*` conventions.

**Security — solid.**
- Body size capped before any parsing.
- Extension whitelist enforced (9 formats).
- Temp file cleanup via `defer os.Remove`.
- Multipart form cleanup in deferred `RemoveAll()`.
- No user content logged in warnings.

**Robustness — good.**
- `disableWriteDeadlineForSSE` called on streaming path (consistent with `chat_session.go`, `search.go`).
- SSE `[DONE]` fallback ensures clients always see a completion signal.
- `headerWritten` guard prevents writing headers when all synthesis fails (returns proper error instead).

**Only suggestion — no tests.** Neither `TestChatAudioSpeech` nor `TestChatAudioTranscription` exists. Given the two distinct response modes (binary stream + SSE + JSON), adding integration tests would protect against regressions in model driver changes. Out of scope for this comment, but worth a follow-up.

### Verdict

Solid work from @euvre. The code is production-ready, follows existing conventions, and all previous review findings were properly addressed before merge. 🚀

PR #17037, opened 2026-07-17
Fix: OpenDataLoader parser

Lynn-Inf:
### Summary

Replace `/file_parse` with `/v1/convert/file`.


wangq8 (commented):
## Review: PR #17037

**Overview:** Migrates `OpenDataLoader` from the legacy `/file_parse` endpoint to the Docling Fast Server `/v1/convert/file` endpoint, and adds a normalization adapter (`_normalize_docling_response`) to convert the nested DoclingDocument JSON into the intermediate element format. 218 additions / 18 deletions across 2 files.

### Good

- **Solid test coverage** — The existing tests were updated for the new endpoint/field names, and a new `test_docling_nested_response_returns_sections` test validates the nested Docling flow end-to-end.
- **Graceful fallback** — Legacy `json_doc` / `md_text` paths are preserved when the response is not in DoclingDocument format. Backward compatibility is handled well.
- **Retry logic preserved** — The 3-attempt retry loop for transient failures remains intact.
- **Multi-format normalization** — Texts, tables, and pictures are all handled with appropriate logic.

### Issues Requiring Attention

#### 1. 🔴 Coordinate conversion in `_extract_docling_prov` is incorrect

```python
# Docling uses top-left origin (t=top from top), but our code expects
# bottom-left origin (y0=bottom, y1=top).  Convert:
return int(page_no), [left, bottom, right, top]
```

The "conversion" simply swaps `t` → `bottom` and `b` → `top` by name, **without using `page_height`**. In a TOPLEFT coordinate system:
- `t` = distance from page top-edge to box top-edge
- `b` = distance from page top-edge to box bottom-edge

To correctly convert to bottom-left (PDF points):
- `bottom = page_height - b`
- `top = page_height - t`

Without this conversion, the Y coordinates will be **inverted**. This will cause incorrect bounding boxes and likely break downstream features that depend on spatial layout (e.g., reading order, crop, annotation positioning).

**Fix:** Pass `page_height` (from Docling document metadata or `page` dimensions) into the conversion. This may require plumbing the page dimensions through the adapter.

#### 2. 🟡 Docling table HTML rebuild ignores `row_span` / `col_span`

```python
for c in cells:
    ...
    r = c.get("start_row_offset_idx", 0)
    col = c.get("start_col_offset_idx", 0)
    text = c.get("text", "")
    if r < num_rows and col < num_cols:
        grid[r][col] = text
```

Docling `table_cells` entries can include `row_span` and `col_span` properties. When a cell spans multiple rows/columns, this code only writes to `grid[r][col]` and leaves the spanned positions empty. Additionally, cell text is **not HTML-escaped** — special characters like `<`, `>`, `&` in cell content could produce malformed HTML.

**Fix:** 1) Apply `rowspan`/`colspan` attributes on the `<td>`, and skip grid positions occupied by merged cells. 2) Escape cell text with `html.escape(text)` before embedding.

#### 3. 🟡 Missing `to_formats` parameter in the POST

The Docling Fast Server `/v1/convert/file` typically accepts a `to_formats` parameter to request specific output formats (e.g., `["json", "md"]`). Without it, the server may return default content or omit desired formats. Since the code now reads `json_content` (and potentially `md_content`), sending the appropriate `to_formats` would make the integration explicit and future-proof.

### Minor / Nitpicks

- **Deeply nested response parsing** — The fallback chain in the response handler (`inner → doc → result → legacy`) is a bit complex and hard to follow. Consider pulling the Docling detection into a small helper function for readability.
- **`_element_html` guard for non-Docling tables** — The existing `cells` handling path (not `table_cells`) could potentially be simplified now that both table formats feed through the same path.

### Verdict: **Changes Requested** 🚧

The coordinate conversion bug (#1) is a correctness issue that will produce wrong bounding boxes. The `row_span`/`col_span` and HTML escaping issues (#2) could produce incorrect table output. These should be addressed before merging. The overall structure and test coverage are good.

PR #17039, opened 2026-07-17
Fix table parse: combined column

wangq8:


wangq8 (commented):
## Review: PR #17039

**Overview:** Replaces the previous `ValueError`-on-duplicate-columns behavior with automatic deduplication via suffix-based renaming (`_2`, `_3`, ...). Applied consistently across Excel, TXT, CSV parsers and schema inference. 49 additions, 6 deletions.

---

### ✅ Correctness

The `_deduplicate_column_names` logic is well-crafted:
- Uses both a `used` set (tracking what has been assigned) and a `reserved` set (all original names). This prevents suffixed names from colliding with **any** existing column name, even ones that appear later in the sequence. Example: `["name", "name_2", "name"]` correctly yields `["name", "name_2", "name_3"]`.
- The `Counter`-based suffix ensures the first duplicate gets `_2`, then `_3`, etc., regardless of insertion order.

Trace through the test fixture `["name", "name", "name_2"]`:
1. `"name"` → added as-is
2. `"name"` → tries `"name_2"` but conflicts with original `reserved` → escalates to `"name_3"`
3. `"name_2"` → added as-is (unique)
Result: `["name", "name_3", "name_2"]` ✅ — matches the test assertion.

### 🔍 Issues

#### 1. 🟡 Redundant fallback check (lines ~476-478)

CodeRabbit correctly flagged this. Since all three entry points (Excel, TXT, CSV) already apply `_deduplicate_column_names` during DataFrame construction, columns entering the schema inference loop are guaranteed unique. The `if len(clmns) != len(set(clmns)):` guard and its `_deduplicate_column_names` call are dead code. Consider removing it for clarity.

#### 2. 🔵 Missing docstring

The new `_deduplicate_column_names` function has no docstring. Given it has non-trivial collision-avoidance logic, a short docstring explaining the deduplication strategy (especially the `reserved` set) would help future readers.

#### 3. 🔵 No PR description

An empty PR body makes it harder to review intent. Would be good to add a Summary section explaining the problem (e.g., "files with duplicate column headers crash the parser") and the chosen fix.

### ✅ Verdict: **Approve** 🚀

Minor issues only — none are correctness blockers. The change is a clear improvement over the previous crash-on-duplicates behavior, the deduplication logic handles edge cases well, and the test covers a realistic scenario. Good work.

PR #17043, opened 2026-07-17
Fix: sample_rate in siliconflow tts model call

Lynn-Inf:
### Summary

Set a valid value as `sample_rate`.

### DIFF
--- a/rag/llm/tts_model.py
+++ b/rag/llm/tts_model.py
@@ -407,7 +407,7 @@ class SILICONFLOWTTS(HTTPBasedTTS):
             "input": text,
             "voice": f"{self.model_name}:{voice}",
             "response_format": "mp3",
-            "sample_rate": 123,
+            "sample_rate": 32000,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
## Review: PR #17043

**Overview:** Trivial one-line fix — changes the SiliconFlow TTS `sample_rate` from `123` (clearly a placeholder/test value) to `32000` (a valid audio sample rate).

### Analysis

| Category | Verdict |
|---|---|
| Correctness | ✅ `32000` is a valid sample rate supported by SiliconFlow TTS API |
| Regression risk | ✅ None — single constant change in an isolated payload builder |
| Style | ✅ LGTM |
| Tests | 🟡 Would benefit from a unit test asserting the payload shape, but this is a trivial constant fix |

### Verdict: **Approve** 🚀

Simple bug fix, no concerns. The previous `123` was never a valid sample rate for audio output — good catch.

PR #17044, opened 2026-07-17
fix(go-agent): handle await response prompts correctly

Hz-186:
## Summary

- Prevent downstream agents from reusing stale `sys.query` values.
- Resolve Await Response tips from the current canvas state and honor `enable_tips`.

## Testing

- `bash build.sh --test ./internal/agent/component/...`
- `bash build.sh --test ./internal/agent/canvas/...`


yuzhichang:
LGTM

Issue #16637, opened 2026-07-04
[Feature Request]: Explicit enable/disable toggle for sync (refresh_freq=0 doesn't disable sync, unlike prune_freq), plus bulk settings across WebDAV sources


sks7g:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

```Markdown
Setting `refresh_freq` to `0` on a data source does not disable the regular sync schedule, unlike `prune_freq`, which is explicitly guarded:


# connector_service.py — prune correctly treats 0 as "disabled":
if bool((task.get("config") or {}).get("sync_deleted_files")) and int(task.get("prune_freq") or 0) > 0



# but list_due_sync_tasks() has no equivalent guard for refresh_freq:
expr = SQL("NOW() - INTERVAL `t2`.`refresh_freq` MINUTE")
query = query.where(..., cls.model.update_date < expr)


With `refresh_freq=0`, the interval becomes `NOW() - INTERVAL 0 MINUTE`, so `update_date < NOW()` is essentially always true — the connector becomes due for sync continuously, rather than disabled. This is the opposite of what a user would reasonably expect from setting the frequency to 0, and inconsistent with how prune already handles this case correctly.
```

### Describe the feature you'd like

1. Add an explicit boolean enable/disable toggle for the regular sync schedule (independent of the numeric `refresh_freq` value), mirroring the existing `sync_deleted_files` opt-in pattern already used for prune. This makes "pause this data source's sync" a first-class, unambiguous action instead of relying on a magic `0` value with unintended behavior.

2. As a related UX improvement: allow applying schedule settings (refresh frequency, prune frequency, enable/disable) across multiple WebDAV data sources at once, instead of having to open and edit each connector individually. Useful for setups with several WebDAV sources (e.g., one per topic/folder) that should share the same sync cadence or be paused/resumed together.

### Describe implementation you've considered

[...]

PR #17051, opened 2026-07-17
fix(connector): disable zero-frequency sync scheduling

Hanxcellent:
### Summary

- Exclude connectors with refresh_freq less than or equal to zero from periodic sync scheduling, so zero disables scheduled sync.
- Reject negative refresh_freq values in connector create and update APIs while preserving zero as the disabled value.
- Add regression tests for due-task selection and API validation.

### Scope

Related to #16637.

This PR addresses only the refresh_freq=0 continuous scheduling bug. It intentionally does not add a separate enable toggle or bulk WebDAV settings.

### API behavior

- Before: negative refresh_freq values were accepted, and zero-frequency connectors were continuously considered due.
- After: negative values return an argument error; zero remains valid but is excluded from scheduling; positive frequencies are unchanged.

### Test plan

- [x] 26 focused connector and sync tests pass.
- [x] Ruff lint and format checks pass for all changed files.
- [x] Full unit, RESTful API, and SDK API matrices were compared with main; all shared test outcomes are identical, with only the three new regression tests added as passing cases.

Local full-suite failures caused by unavailable DeepDOC model assets and the backend service not running are unchanged from main.

### DIFF
--- a/api/apps/restful_apis/connector_api.py
+++ b/api/apps/restful_apis/connector_api.py
@@ -47,9 +47,20 @@ def _connector_auth_error(connector_id: str, user_id: str):
 
 
 def _validate_refresh_freq(req: dict):
-    if "refresh_freq" in req and int(req["refresh_freq"]) < 0:
-        return get_json_result(code=RetCode.ARGUMENT_ERROR, message="refresh_freq must be greater than or equal to 0")
-    return None
+    if "refresh_freq" not in req:
+        return None
+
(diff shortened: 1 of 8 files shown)


JinHai-CN:
@Hanxcellent would you please resolve the conflicts?

mkaaad:
Setting the sync task to 0 means it runs as soon as possible; this is intentional.

PR #17058, opened 2026-07-17
Fix SharePoint connector not run after added

wangq8:


wangq8 (commented):
## Review: Fix SharePoint connector not run after added

**File:** `api/db/services/connector_service.py` (+16/-6)

### What it does

Adds a `run_immediately` parameter to `SyncLogsService.schedule()`. When `True`, it backdates `update_time = 0` and `update_date = 1970-01-01` on the newly created task row so the task runner picks it up on the next poll cycle instead of waiting.

`run_immediately=True` is set at four call sites:
- `schedule_tasks()` — both SYNC and PRUNE
- `rebuild()` — both SYNC and PRUNE
- `link_connectors()` — both SYNC and PRUNE (when linking new connectors)

### Likes

- **Targeted fix** — minimal diff, one file, clear intent
- **Backward compatible** — defaults to `False`, existing callers unchanged
- **Correct semantics** — after a connector is added/rebuild/linked, the user expects it to start syncing immediately, not sit pending

### Concerns

1. **Raw SQL bypassing ORM** — `DB.execute_sql()` hardcodes column names (`update_time`, `update_date`) and assumes the table structure. If the model mapping for these fields ever changes (rename, removal, migration), this raw update silently breaks. Consider adding a comment referencing the model field, or moving this logic into the model layer itself.

2. **Race condition with rapid triggers** — If a connector is linked to multiple KBs in quick succession, `schedule()` is called multiple times with `run_immediately=True`. The task runner may start processing the first task while the second is still being created, leading to overlapping syncs. Not a new problem per se (the original code had the same trigger pattern), but worth being aware of.

3. **No test coverage** — The PR is labeled `size:S` and it's a bugfix, but adding a unit test for the `schedule()` → `run_immediately` path would prevent regression. Even a simple test verifying that `update_time` / `update_date` is set to 0 / epoch after `run_immediately=True` would help.

4. **Minor: raw SQL is MySQL-specific syntax** — `UPDATE ... SET update_time = %s, update_date = %s` is standard enough, but worth double-checking whether the project supports PostgreSQL or other backends where datetime coercion might behave differently.

### Verdict

Clean, minimal bugfix. The approach (backdating the poll column) is pragmatic. Would suggest adding a short inline comment near the raw SQL explaining *why* update_time is being zeroed, and ideally a small test. But functionally this looks correct.

`APPROVED` pending the inline comment suggestion.

PR #17059, opened 2026-07-17
refactor: use the built-in max/min to simplify the code

futurehua:
### Summary

In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.


futurehua:
@JinHai-CN @yingfeng Hi, Could you please review this PR at your convenience? Thank you very much.

PR #17064, opened 2026-07-17
fix(agent/tools): surface github search error responses

Yurii214:
### Summary

`agent/tools/github.py` indexes `response["items"]` directly after `requests.get(...).json()`. the github search api returns `{"message": ...}` **without** an `items` key on realistic conditions — a rate limit (403/429; this tool sends no auth token, so the unauthenticated ~10 req/min search limit is easy to hit) or an invalid query (422). that raised `KeyError('items')`, which the tool's retry loop then surfaced to the model as the opaque `"GitHub error: 'items'"` instead of the real reason.

this guards the missing key and raises the api's actual `message` (with a clear fallback) into the existing retry/`_ERROR` path, so the model sees e.g. `"GitHub error: API rate limit exceeded ..."`. valid responses are unchanged.

adds `test/unit_test/agent/tools/test_github_unit.py` covering the rate-limit response (asserts the real message is surfaced, no `KeyError`) and the normal result path.


Hz-186 (approved):
@Yurii214, LGTM

PR #17067, opened 2026-07-17
Fix: hybrid retrieval returns 0 chunks when the question has no lexical overlap — dense-only fallback (#17041)

### LINKED ISSUE #17041: Dense (vector) candidates are restricted by the full-text predicate on both doc engines — retrieval returns 0 chunks for languages the tokenizer cannot segment (e.g. Thai)
## Summary

In both storage connectors, the dense (kNN) search receives the full-text
match as a **filter**, so dense results cannot include chunks that fail the
full-text predicate. For languages that `rag_tokenizer` cannot segment
(observed with Thai), a natural-language question produces no matching
lexical terms, the predicate matches nothing, and retrieval returns 0 chunks
— even with `vector_similarity_weight: 1.0` and a multilingual embedding
model (bge-m3) that supports Thai and produces useful similarity scores
(~0.65–0.71 on our data once candidates are visible; see experiments).

Code (v0.25.6; unchanged at v0.26.4):

- `rag/utils/es_conn.py` — the knn query receives the bool text query as a filter:

  ```python
  s = s.knn(
      m.vector_column_name,
      m.topn,
      m.topn * 2,
      query_vector=list(m.embedding_data),
      filter=bool_query.to_dict(),
      similarity=similarity,
  )
  ```

- `rag/utils/infinity_conn.py` — the dense match receives the full-text filter:

  ```python
  elif isinstance(matchExpr, MatchDenseExpr):
      if filter_fulltext and "filter" not in matchExpr.extra_options:
          matchExpr.extra_options.update({"filter": filter_fulltext})
  ```

## Steps to reproduce

1. Dataset on either doc engine (we reproduced identical behavior on
   Elasticsearch and Infinity deployments), embedding model `bge-m3`,
   containing chunks with Thai content, e.g. a helpdesk ticket containing
   `Purpose: แจ้งปัญหาระบบ`.
2. `POST /api/v1/retrieval` with:
   ```json
   {
     "question": "ปัญหาแจ้งระบบภาษี",
     "dataset_ids": ["..."],
     "similarity_threshold": 0.01,
     "vector_similarity_weight": 1.0
   }
   ```
3. Result: `{"code": 0, "data": {"chunks": []}}` — 0 chunks.

## Experiments (observed results, v0.25.6, same dataset)

| Question | Result |
|---|---|
| `"ปัญหาแจ้งระบบภาษี"` (natural Thai) | 0 chunks |
| `"zzqx"` (garbage Latin, no lexical overlap, `vector_similarity_weight: 1.0`, threshold 0.01) | 0 chunks |
| `"tax system problem"` (tokens present in chunk) | 2 chunks (~0.63) |
| `"แจ้งปัญหาระบบ"` (exact Thai run copied verbatim from chunk) | 2 chunks (~0.68) |
| Same Thai question, chunk content AND query pre-segmented with spaces | 2 chunks (~0.65) |
| Same Thai question with `"cross_languages": ["English"]` | 2 chunks (~0.65) |

**What we infer from these observations** (inference, not directly measured):
results appear only when the question shares at least one lexical token with
the chunk, regardless of vector settings — consistent with the dense leg
being restricted to full-text candidates, as the connector code above
suggests. When candidates do pass the lexical predicate, their reported
vector similarity for Thai queries is healthy (0.65–0.71), so the embeddings
themselves discriminate Thai fine.

We could not run an unfiltered-kNN control **through the API** precisely
because no option exposes it — happy to run any diagnostic the maintainers
suggest (e.g. a patched build with the fulltext filter dropped) to close
this gap.

Languages other than Thai that lack space-delimited words (Lao, Khmer,
Burmese) **may also be affected**; we have only reproduced Thai.

## Question to maintainers: bug or intended design?

Is `vector_similarity_weight` intended to be a scoring weight only (over
candidates that already passed the full-text predicate), or should it also
influence candidate generation? If the gating is intended (e.g. to reduce
noise and compute), this issue is better read as a **feature request**:

## Proposed change

Allow the dense leg to recall candidates without the **lexical/full-text
predicate** — e.g. a retrieval option or KB setting. To be explicit about
scope: only the full-text match filter would be omitted from the dense leg;
**mandatory scope filters must remain** (tenant/kb_id/doc_ids, `available`,
metadata conditions — `filter_cond` in `infinity_conn.py` stays as-is).

This would let multilingual embedding models serve languages the tokenizer
cannot segment, while hybrid scoring/fusion behavior stays unchanged for
everyone else.

## Environment

- RAGFlow v0.25.6 (connector code unchanged at v0.26.4)
- Doc engine: behavior observed on deployments of both `elasticsearch` and `infinity`
- Embedding: `bge-m3` via an OpenAI-compatible provider
- Deployment: Kubernetes (helm), Linux x86_64

## Workarounds we currently use (for other users hitting this)

1. Pre-segment Thai with a word-break dictionary (Java `BreakIterator` /
   ICU) on **both** indexed chunk content and the query — makes the lexical
   predicate match Thai word-by-word without any LLM.
2. `"cross_languages": ["English"]` per retrieval call (requires a healthy
   tenant default chat model; note `cross_languages()` returns the original
   query silently on LLM error, which makes failures hard to detect).

We are happy to submit a PR once maintainers indicate the preferred shape
(request option vs. KB config vs. always-independent dense leg).

MatthewSynthia:
### Summary

Fixes #17041.

On both the Elasticsearch and Infinity connectors, the dense (kNN) leg of hybrid retrieval receives the full-text match as a **pre-filter**, so dense candidates must lexically match the question. For questions with no lexical overlap with any chunk — e.g. languages `rag_tokenizer` cannot segment, such as Thai — the predicate matches nothing and retrieval returns **0 chunks**, even with `vector_similarity_weight: 1.0` and a multilingual embedding model that scores the right chunks at healthy similarity (~0.65–0.71 in the issue's experiments).

### Design: zero-result dense-only fallback, not an unconditional ungating

Unconditionally removing the lexical gate from the dense leg would change candidate sets, totals, and ranking for **every** hybrid query on ES/Infinity (and shift results under the default `similarity_threshold`), which is a much riskier change than the issue needs. Instead, `Dealer.search()` now has a third attempt in its existing retry cascade:

1. hybrid search as today (`min_match` 0.3) — unchanged;
2. existing low-`min_match` retry (0.1, lower similarity) — unchanged;
3. **new**: if still 0 hits, retry with a **dense-only expression list**, which every doc engine pre-filters by the scope conditions only (kb/doc/metadata/availability). This is the exact situation the issue describes: the lexical predicate is unsatisfiable, so hybrid fusion has nothing to offer and pure semantic recall is strictly better than an empty answer.

Behavior is **byte-identical to `main` whenever the current pipeline returns at least one chunk** — the fallback can only turn "0 chunks" into "some chunks".

### What changed

- `rag/nlp/search.py`: the dense-only third attempt described above. The fallback builds a **fresh** `MatchDenseExpr` because connectors rewrite `extra_options` in place (Infinity attaches the full-text filter to the shared expression object during attempt 1, which would otherwise leak into the fallback).
- `rag/utils/infinity_conn.py`: a dense-only search now carries the scope conditions (`doc_ids`, `available_int`, metadata) as its filter. Previously a dense-only search on Infinity ran with **no filter at all** when no full-text expression was present (kb scoping survived only via per-KB tables); the fallback needs those conditions honored, and it is a correctness fix on its own. The hybrid path is untouched — `filter_fulltext` still wins when a text leg exists.
- `test/unit_test/rag/nlp/test_dense_only_fallback.py`: unit tests for the cascade and for the connector-level contract the fallback relies on (a dense-only list reaches ES/Infinity with scope-only filters; OpenSearch already pins the same invariant in `test_opensearch_hybrid_search.py::test_knn_filter_excludes_text_must_clause`, and `ob_conn` builds dense-only lists itself when full-text is disabled or the vector weight is 1.0). Everything runs with mocked clients — no cluster needed. Stubs are install-only-if-real-import-fails, so the file is safe at any collection position.

### Evidence

- New unit tests: 8/8 pass on this branch. On `main`, the 3 defect-detecting tests fail (no third attempt → Thai question returns 0 chunks; Infinity dense-only drops scope filters) and the 5 behavior-pinning tests pass on **both** `main` and this branch — i.e. the lexical-hit, low-min_match-hit, and doc_id-scoped paths are unchanged.
- Combined run with the neighboring suites (`test_search_pagination`, `test_rank_feature_scores`, `test_opensearch_hybrid_search`, `test_opensearch_doc_meta`, `test_delete_query_construction`): 160/160 pass.
- Case-matrix verification driving the real `Dealer.search()` cascade and real connector query construction over the issue's repro string: 12/12 structural invariants pass on this branch vs 7/12 on `main` (the 5 failing ones are exactly the reported defect + the Infinity scope-filter gap; the 4 unchanged-behavior invariants pass on both).
- `ruff check` and `ruff format --check` clean on all touched files.

### Behavior notes

- The fallback uses the retry-tier similarity floor (0.17), consistent with the existing second attempt.
- Highlighting is skipped in the fallback (nothing matched lexically, so there is nothing to highlight), matching the existing no-question listing path.
- Chunks recalled by the fallback still flow through the normal post-search scoring (`rerank_with_knn` on ES, engine fusion score on Infinity) and `similarity_threshold` filtering in `Dealer.retrieval()`.

_This PR was AI-assisted (Claude Code); all changes and evidence were verified as described above._


zeweihan:
The zero-result dense-only fallback as a third retry cascade attempt is the conservative right call here. Unconditionally removing the lexical gate from the dense leg would shift candidate sets, totals, and ranking for *every* hybrid query on ES/Infinity (and re-rank under the default `similarity_threshold`), which is a far riskier change than #17041 actually needs — the symptom is "0 chunks when the tokenizer can't segment this language," not "hybrid ranking is wrong in general." By only activating after the first two attempts return 0, you preserve result parity for the common case and only relax the gate when it has demonstrably gate-kept everything. Good design discipline: the fix is scoped to the failure, not to the mechanism.

For the #17041 root cause specifically, this is exactly the missing branch that lets the dense leg *contribute*. The OP's experiment (~0.65–0.71 cosine for semantically-rich Thai questions) proves the content survives the embedding step with healthy similarity and would surface if the dense candidates were ever allowed to run — the bug was never "the model can't find it," it was "the candidate gate dropped it before the model got a vote." The fallback removes that gate when the gate has provably excluded the whole result set, which turns "confidently-wrong-about-absence" (retrieval returns `{chunks: []}` and the caller reads it as "corpus has nothing relevant") into actual dense recall.

One thing that matters for this to not just move the silent-degradation problem one step sideways: the fallback needs to surface *that it fired*, not just return chunks. Consider a multilingual corpus where `rag_tokenizer` routinely fails to segment a chunk language — for that corpus *every* query in that language hits the third retry every time, and from the outside it looks like a healthy retrieval path returning N chunks, while silently running a degraded (dense-only, no lexical rerank) mode on every request. A `retrieval_mode` / `attempt_path` stamp per result set — `{hybrid | low_min_match | dense_only_fallback}` — makes "this retrieval used the dense-only fallback" queryable rather than inferred-from-zero-results, same shape as a `text_coverage` / `chunk_source` provenance stamp. Without it, an operator whose corpus is structured such that the fallback fires 100% of the time has no signal that their "hybrid" retrieval has been silently running dense-only for weeks.

Second, more forward-looking: the fallback is reactive (only fires after 0 hits on the first two attempts). For datasets an operator *knows* are multilingual / contain languages the tokenizer can't segment, a per-dataset option to declare dense-first or dense-only would let the fallback not be the *only* path to those chunks — right now you pay two full failed hybrid attempts (with their lexical-predicate cost) on every multilingual query before the dense leg is allowed to run at all. The fallback is the correct safety net; a declared dense-preference knob for known-multilingual corpora would be the correct front door.

Fallback is the right net; surfacing that it fired + a declared-mode option for known-multilingual datasets would make it the durable fix rather than just the reactive one.

MatthewSynthia:
Thanks — agreed on both points, and on keeping them out of this diff so the fix stays reviewable as a scoped bug fix.

On observability: the fallback does announce itself — `Dealer.search dense-only fallback: no full-text matches for question.` at debug level — so "is my corpus living on the fallback" is already grep-able from retrieval logs today. I agree that's operator-observability, not caller-observability: a per-result `retrieval_mode` stamp (`hybrid | low_min_match | dense_only_fallback`) touches the retrieval response schema and every consumer of it (API, SDK, agents), which deserves its own review rather than riding along here.

On the dense-first per-dataset option: also agreed that's the durable front door for known-multilingual corpora — it's essentially the reporter's original proposal, and the fallback is deliberately just the safety net that requires no configuration to be correct.

I'll file both as follow-up issues linked back to #17041 once this lands, so the scoping decision is recorded rather than lost.

MatthewSynthia:
The CI workflows here are still waiting on maintainer approval to run (first PR from this account, so runs need a manual approve). Once they've run I'm around to fix anything that comes up.

thaweesup:
@MatthewSynthia Thanks — we'll validate #17067 against the original deployment
(RAGFlow v0.25.6, k8s, DOC_ENGINE=infinity, real Thai helpdesk corpus) and
report back with before/after numbers.

One heads-up for anyone else validating on a release build: the `rag/nlp/search.py`
hunk does **not** apply cleanly to v0.25.6 (`git apply --check` fails — the
low-min_match retry block was reformatted between v0.25.6 and main; the
`rag/utils/infinity_conn.py` hunk applies cleanly). We back-ported the fallback
by hand to the same logical location (after `total = self.dataStore.get_total(res)`
in the min_match-0.1 retry branch); `MatchDenseExpr` / `thread_pool_exec` are
already imported in v0.25.6 so the backport is mechanical.

Test plan (all against the same dataset used for the numbers in the issue):
- `"ปัญหาแจ้งระบบภาษี"` and a longer natural-Thai question — currently 0 chunks; expecting dense-only fallback recall
- `"zzqx"` control — will report whichever way it lands relative to the 0.17 similarity floor
- Regression guards: pre-segmented Thai / English-overlap / exact problem-code queries — currently returning chunks; expecting byte-identical scores
- Plus the `Dealer.search dense-only fallback` debug-log check

Results to follow.

thaweesup:
@MatthewSynthia Validation done on the original deployment (RAGFlow v0.25.6, k8s,
DOC_ENGINE=infinity, bge-m3 via an OpenAI-compatible provider, real Thai helpdesk
dataset — the same one behind the issue's numbers). Patch = #17067 back-ported to
v0.25.6 as described above, hot-applied to the pod.

| Query | Before | After #17067 |
|---|---|---|
| `"ปัญหาแจ้งระบบภาษี"` (natural Thai, no cross_languages) | 0 chunks | **2 chunks (0.658, 0.635)** |
| Longer natural-Thai question | 0 chunks | **2 chunks (0.655, 0.638)** |
| `"zzqx"` (garbage control) | 0 chunks | 2 chunks (0.354, 0.349) — fallback fires, sims clear the 0.17 floor |
| Pre-segmented Thai | 2 chunks (0.653, 0.647) | 2 chunks (0.653, 0.647) — identical |
| `"tax system problem"` | 2 chunks (0.629, 0.627) | 2 chunks (0.629, 0.627) — identical |
| `"DVAT-69-00043"` | 2 chunks (0.711, 0.698) | 2 chunks (0.711, 0.698) — identical |

From our side this resolves #17041: natural-Thai questions now recall via the
dense leg with no translation step, and every query that already returned
results is score-identical to before — matching the "byte-identical when the
current pipeline returns ≥1 chunk" claim.

One observation for reviewers: the `zzqx` row shows the intended trade-off —
lexically-unmatchable garbage now returns weak (~0.35) matches instead of an
empty set. For our use case that's acceptable (callers filter by similarity),
but deployments relying on "no lexical match ⇒ empty result" semantics will
see this change on such queries.

MatthewSynthia:
Thanks for going this deep on it, the hand-backport to v0.25.6 and the full regression run especially. Good to see the already-passing queries come back with identical scores.

On zzqx: yeah, that's the intended trade-off. The only guard is the 0.17 floor, so garbage that still scores above it returns weak matches instead of nothing. Doesn't matter if you filter by similarity, but it's a real change for anyone counting on empty results when there's no lexical match, so probably worth a line in the release notes. I left it as just the floor to avoid adding another config knob, but I can make it configurable if reviewers want.

CI hasn't run yet, still waiting on first-run approval since this is a new account. If someone can approve it I'll fix whatever it flags.

6ba3i (changes requested):
the PR is now stale against current `main` and a few things need to be fixed during rebase:

* The fallback manually rebuilds `MatchDenseExpr` and now drops current options such as `num_candidates`, which can change KNN recall/latency. It should preserve the original clean options and only adjust the retry similarity.
* The same lexical-gating issue still exists in the Go retrieval path, with no equivalent dense-only recovery, so Python and Go would behave differently for the same case.
* Current Python retrieval has already changed since this PR was opened and now goes directly dense when there is no usable text expression, so the original reproduction needs to be revalidated after rebasing.

One additional point worth clarifying: this only fixes the zero-result case. If even one weak lexical match survives, stronger vector-only candidates can still remain excluded. If that is intentional, it would be good to explicitly document/test that behavior.

6ba3i:
hello @MatthewSynthia could you please fix the issues raised in my comment so we can merge the pr ?

6ba3i:
overlaps #19502

PR #17068, opened 2026-07-18
Feat: add document_level mode to the Extractor dataflow component

iwasaki1112:
## What

`Extractor._invoke()` (a dataflow component: `rag/flow/extractor/extractor.py`) always calls the chat LLM once per upstream chunk and writes the result onto that chunk. For a whole-document field (e.g. a `document_description` metadata field), the value that survives into `DocMetadataService.update_document_metadata()` ends up being whichever chunk's LLM call happens to run last — `update_metadata_to()` overwrites non-list metadata values rather than merging them, so the "summary" describes only the final chunk (e.g. the last of 86 pages), not the document.

Add an opt-in `document_level` param. When set, `Extractor` concatenates every chunk's text into a single prompt (fit to the model's context budget the same way every other `agent/component/llm.py` call site does) and writes the one resulting summary onto every chunk, so downstream last-write-wins metadata merging ends up with that single document-level answer instead of one arbitrary chunk's. This is a single call over the concatenated text, not a map-reduce summarizer — a document whose combined chunk text exceeds the model's context window is summarized from a truncated portion, not staged/reduced across multiple calls. That's still strictly better than the current per-chunk behavior (which always describes only the last chunk, truncation or not), but documents much larger than the context window get a partial-document summary rather than a complete one.

## Why the JSON handling needed to change too

A single whole-document prompt is exactly where a small/local chat model most often ignores a "reply with only this JSON object" instruction and answers in free-form prose instead. Left unhandled, `update_metadata_to()` silently drops that result (`json_repair` can't invent JSON out of prose that contains none), so the whole feature would appear to succeed while producing empty metadata.

Rather than inventing new parsing logic, this mirrors the retry-until-parsable-JSON pattern `LLM._invoke_async()` already uses for its own `"structured"` output (`agent/component/llm.py`): retry up to `max_retries` times, accept the first reply `json_repair.loads()` turns into a dict with at least one usable value, and otherwise leave the field unset rather than store unparsable prose as metadata. "Usable" is deliberately narrow: `DataflowService._process_chunks()` feeds this dict straight into `update_metadata_to()` (`common/metadata_utils.py`), which only keeps `str`/list-of-`str` values per key and silently drops everything else — so a reply that's syntactically valid JSON but carries e.g. a nested object value is filtered the same way here and retried, instead of being "accepted" only to quietly lose that key downstream. There's no fixed set of expected *keys* to validate against (the sys_prompt's own JSON example decides those, same as the `"structured"` branch's own schema-free acceptance), only the value shape. `fit_messages()` errors are treated as terminal, matching every other call site in `agent/component/llm.py`, instead of sending the not-fitted (potentially empty) messages anyway. A small `_strip_markdown()` cleans residual Markdown syntax (headings/bold/lists/tables/code fences) from the stored string values, since a plain-text metadata field has no renderer for it — table rows are converted to plain text (cell content kept), not deleted wholesale.

## Changes

- `rag/flow/extractor/extractor.py`: `ExtractorParam.document_level` (bool, default `False`); `Extractor._document_level_extract()` implements the behavior above.
- `web/src/pages/agent/form/extractor-form/index.tsx`, `use-switch-prompt.ts`, `web/src/pages/agent/constant/pipeline.tsx`, `web/src/locales/en.ts`: expose the new param as a toggle in the dataflow canvas Extractor form, next to the existing "Result destination" field.
- `test/unit_test/rag/flow/extractor/test_document_level.py` (new): per-chunk path regression (driven through `Extractor._invoke()` itself), concatenation + result-sharing, retry success/exhaustion (including an empty-`{}`-is-not-success case), the fit-error skip path, and Markdown/table stripping.

Existing per-chunk behavior (`document_level=False`, the default) is unchanged.

## Testing

- `uv run pytest test/unit_test/rag/flow/extractor/` — 12/12 pass.
- `pnpm run type-check` / `eslint` on the 4 changed frontend files — no new errors (pre-existing, unrelated errors elsewhere in the repo are untouched by this change).
- Verified end-to-end against a live RAGFlow + Ollama deployment: a dataflow `Parser → Chunker → Extractor(document_level=true, field_name=document_description) → Tokenizer` pipeline against a real 12-page PDF produced a `document_description` covering the whole document (previously: only the last chunk), with no leftover chat-assistant framing or Markdown syntax in the stored value.


xugangqiang (commented):
## Code Review Findings (read-only)

Thanks for the contribution. The core fix is sound and well-tested. However, there are a few issues that should be addressed before merge.

### Finding A (Important) — `document_level` silently fails / misbehaves with plain-text destinations
The whole pipeline is JSON-only: `extractor.py:162` requires `isinstance(candidate, dict)`, and `extractor.py:190` stores `json.dumps(parsed)` on the chunk. But the default `field_name` is `Summary` (`pipeline.tsx:353`), and Summary/Keywords/Questions destinations are meant to return plain text.

If a user enables the toggle with a plain-text prompt (or the default `Summary` destination), `json_repair.loads` returns a non-dict → retry loop exhausts → the field is silently left unset (only a warning logged). Worse, for a non-`metadata` destination the value would be written into the wrong chunk key and tokenized incorrectly downstream in `_process_chunks`. The tooltip ("Use this for document-level fields (e.g. a document summary)") actively invites this footgun.

**Recommendation:** Either add a plain-text fallback (when the reply is not a usable JSON dict, store the raw text on all chunks under `field_name`), or explicitly restrict / document `document_level` to the `Metadata` destination and fix the tooltip wording. Add a regression test exercising a default-style plain-text prompt.

### Finding B (Minor) — empty values accepted; list elements not stripped
- `extractor.py:180`: `usable = {k: v for ... if isinstance(v, str) or (isinstance(v, list) and all(...))}` accepts empty string `""` and empty list `[]` (`all([])` is `True`). These should be discarded as non-usable and retried.
- `extractor.py:189`: `_strip_markdown` is applied only to scalar `str` values, not to `list[str]` elements. Markdown inside list entries survives.

### Finding C (Minor) — images not passed in document-level path
`_document_level_extract` calls `self.chat_mdl.async_chat(msg_fit[0]["content"], msg_fit[1:], self._param.gen_conf())` without `images=self.imgs`, whereas the per-chunk path (`_generate_async`, `llm.py:371-372`) passes images when present. For consistency, pass `images=self.imgs` when non-empty.

### Finding D (Minor / clarify) — retry protection inactive by default
`max_retries` defaults to `0` (`agent/component/base.py:49`), so the retry loop `range(self._param.max_retries + 1)` runs only one attempt by default. The "retry until parsable JSON" safety net only engages if the user raises `max_retries`. The PR narrative emphasizes retrying, but with defaults there is no retry. Consider documenting this in the tooltip, or giving this mode a higher default.

### Finding E (Tech debt, out of scope) — duplicate `FormSchema`
`FormSchema` is duplicated in `index.tsx` and `use-switch-prompt.ts`; the PR correctly keeps both in sync (`.default(false)`). Pre-existing smell; suggest consolidating later.

### Nits
- `web/src/pages/agent/form/extractor-form/index.tsx:95-101`: pass the handler by reference — `onCheckedChange={field.onChange}` instead of an inline arrow.
- `extractor.py:163` blind `except Exception: continue` triggers ruff `BLE001`/`S112`. Valid but low priority (consistent with existing style at `llm.py:455`).

### Tests
Unit coverage is good (12 tests, no external services, per-chunk regression driven through `_invoke`). No default-behavior regression since `document_level` is opt-in.

### Merge risk
This branch is ~7440 commits behind current `upstream/main` (opened 2026-07-18). Rebase and re-verify that `_process_chunks` (`dataflow_service.py`) and `llm.py` still match the assumed behavior before merging.

**Recommendation: merge after addressing Finding A (and ideally B/C/D).**

PR #17079, opened 2026-07-20
Remove redundant PatchDSL and consolidate component-param injection into override_params

xugangqiang:
## Summary

`Pipeline.Run` already accepts an `override_params` map (keyed by `cpnID`, override-wins) that is merged into each component's params at compile time via `canvas.WithOverrideParams`. `PatchDSL` performed a byte-for-byte equivalent merge by baking `ParserConfig` into the DSL text before compilation, so `runPipelineWithDSL` applied the same `Doc.ParserConfig` twice with no effect.

## Change

- Delete the now-dead `PatchDSL` function (`internal/ingestion/pipeline/pipeline.go`); it had a single call site.
- In `runPipelineWithDSL`, compile the DSL unchanged and pass the extracted component params to `Run` as `override_params`.

## Behavioral note (important)

Previously `PatchDSL` was also the channel that delivered the tenant LLM id into Extractor components: it baked `parserConfig` — after `common.InjectExtractorLLMID(parserConfig, Tenant.LLMID)` had injected `llm_id` — into the DSL. The old `Run` override_params re-cast `Doc.ParserConfig` (which intentionally omits `llm_id`). So removing `PatchDSL` while keeping that re-cast would have silently dropped LLM id injection. This change passes the local `parserConfig` (the LLMID-injected copy) to `Run`, preserving the prior behavior.

The DSL string returned by `runPipelineWithDSL` is only consumed by `recordPipelineLog` (structure storage) and `ExtractPayload` (which reads the terminal structure, not params), so returning the original DSL is equivalent.

## Test plan

- `bash build.sh --test ./internal/ingestion/task/... ./internal/ingestion/pipeline/...` — both packages pass.
- `gofmt` clean; pre-commit hooks pass.


xugangqiang:
@yuzhichang @qinling0210 Pls help to review.

yuzhichang:
LGTM

PR #17080, opened 2026-07-20
Fix: display timeout error when checking model accessibility

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
## Review: PR #17080 – Fix: display timeout error when checking model accessibility

**Overall:** A solid, well-structured refactoring that significantly reduces code duplication in `verify_api_key`. No regressions identified.

### ✅ Positive

**1. `_run_verification` helper** – Extracting the try/except/timeout boilerplate into a single helper is the right call. Every model-type check now has uniform error handling:
- Timeout → `"Timeout accessing {label}."`
- CancelledError → `"{label} verification aborted."` (previously unhandled!)
- Generic exception → `"Fail to access {label}.{str(e)}"`

**2. `passed_types` set replaces individual booleans** – Much cleaner than the `chat_passed`, `embd_passed`, etc. flags. The `any_passed`/`break` logic is preserved with the same semantic.

**3. `assert` → proper if-statement** – Old code used `assert provider_name in EmbeddingModel` etc., which could be silently skipped with `-O`. The new code provides explicit error messages. Good defensive practice.

**4. Iterating all model types per LLM** – Old code `elif`-chained to the first matching type. New code tries all types if the first one fails. This is an improvement — a model that advertises both `chat` and `vision` types, for example, now gets a verification attempt on each type.

### 🔍 Behavior verification

- **OCR/ASR `check_available()` return unpacking**: Correct. `_run_verification` returns `(True, (ok, reason))` → `ok2, reason = result`.
- **VISION `test_image` reference**: Old `image_data = test_image; await ...(mdl.describe, image_data)` ≡ new `await ...(mdl.describe, test_image)`. Equivalent.
- **Success → `msg = ""` + `break`**: Preserved from the old behavior.
- **model_verify_result overwrite per llm**: Functionally identical — each llm gets a single SUCCESS/FAIL entry.

### ⚠️ Minor

- The error message format has changed slightly (e.g., no more "using this api key" suffix). Error messages are now consistent across types but may differ from what downstream consumers expect.
- `_run_verification` returns a tuple `(bool, Any)` — callers unpack `result` differently per type (some expect `arr, tc`, some a bool, some a tuple). This is the same fragility as the old code but now centralized.

**Verdict:** Well done. No blockers, ready to merge.

PR #17082, opened 2026-07-20
fix(web): unify required asterisk color in instance name section

euvre:
### Summary

The required-field asterisk of the "Instance Name" label in the provider instance draft card used `text-destructive`, while the asterisks rendered by `FormLabel` (e.g. the "API-Key" field right below it) use `text-state-error`. This caused two visibly different asterisk colors within the same card.

This PR changes the instance name asterisk to `text-state-error` so it matches the color used by `FormLabel`, making the required markers consistent across the card.


wangq8 (commented):
## Review: PR #17082

**Change:** Replaces `text-destructive` with `text-state-error` on the required-field asterisk in the instance name section to match the `FormLabel` convention used by other fields in the same card.

**Assessment:** Trivial one-class swap, clearly correct. Unifies asterisk styling within the card. No tests needed. No risk of regression.

**Verdict:** ✅ LGTM — safe to merge.

dcc123456 (approved):

dcc123456 (approved):

PR #17083, opened 2026-07-20
Remove selectUploadParser: documents now inherit parser_id from KB directly

xugangqiang:
The `selectUploadParser` helper was responsible for mapping file types (e.g.
visual → picture, .pptx → presentation) to a DSL pipeline template at upload
time. After the refactoring that consolidated per-file-type DSL selection into the
Parser runtime component, this upload-time mapping is no longer necessary —
documents simply inherit the KB-level `parser_id` (or use `pipeline_id` in canvas
mode).

**Changes:**

- **`document_upload_helpers.go`**: Removed `selectUploadParser` function.
  `contentHashHex` is kept as it is still used for deduplication.

- **`document_upload.go` (`newDatasetDocument`)**: Directly use `kb.ParserID`
  instead of `selectUploadParser`. Removed the post-resolution
  `ResolveComponentParamsDefaults` block that was only reachable when the
  selected parser differed from the KB default — now trivially unreachable.

- **`file2document.go` (`convertFiles`)**: Same simplification as above.

- **`builtin_registry.go`**: Removed `DefaultParserID`, `fileTypeParserOverrides`,
  `presentationExtPattern`, and `emailExtPattern` — all dead code after the
  `selectUploadParser` removal. Also dropped the now-unused `regexp` import.

- **Removed tests**: `TestSelectUploadParser_MirrorsPython` and
  `TestDefaultRegistry_DefaultParserID` tested the deleted functions.

- **Minor cleanup**: Removed stale `CreateDocumentRequest` / `CreateDocument`
  doc comments in `document.go` and `document_crud.go`.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/internal/ingestion/pipeline/builtin_registry.go
+++ b/internal/ingestion/pipeline/builtin_registry.go
@@ -8,7 +8,6 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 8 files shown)


xugangqiang:
@yuzhichang @JinHai-CN @yingfeng @wangq8  Pls help to review.

PR #17084, opened 2026-07-20
fix[go]: stop leaking [DONE] sentinel into streamed chat answers

euvre:
### Summary

In Go mode, model drivers emit the OpenAI-style `[DONE]` marker as their final `sender` call (mirroring the Python convention, where consumers break on it). The channel adapter `chatStreamWithContext` (`internal/service/model_chat.go`) forwarded every delta verbatim into the channel, including this terminal sentinel.

As a result, channel consumers accumulated `[DONE]` as answer text:

- `AskService` (search AI summary) appended it to the final answer, so the rendered AI summary ended with a stray `[DONE]`.
- The mindmap handler (`internal/handler/mindmap.go`) had the same exposure.

Fix: treat `[DONE]` as end-of-stream in the adapter — abort the driver loop via a sentinel error and never forward the marker into the channel. Drivers populate `ToolCallsResult`/`UsageResult` before sending `[DONE]`, so nothing is lost.

Adds `internal/service/model_chat_test.go` covering that the sentinel is stripped while content deltas pass through intact. `bash build.sh --test ./internal/service/...` passes.

### DIFF
--- a/internal/service/model_chat.go
+++ b/internal/service/model_chat.go
@@ -18,6 +18,7 @@ package service
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"go.uber.org/zap"
@@ -25,6 +26,13 @@ import (
 	"ragflow/internal/common"
 	modelModule "ragflow/internal/entity/models"
 )
+
+// streamDoneSentinel is the OpenAI-style end-of-stream marker model drivers
+// emit as their final sender call. It is a transport signal, not answer text.
+const streamDoneSentinel = "[DONE]"
+
+// errStreamDone aborts the driver loop once the terminal sentinel arrives.
+var errStreamDone = errors.New("chat stream done")
 
 func (m *ModelProviderService) Chat(tenantID, modelID string, messages []modelModule.Message, config *modelModule.ChatConfig) (*modelModule.ChatResponse, error) {
 	chatModel, err := m.GetChatModel(tenantID, modelID)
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## Review: PR #17084

**Change:** Introduces an `errStreamDone` sentinel error in `chatStreamWithContext` to abort the driver loop when the `[DONE]` marker is encountered, preventing it from leaking into channel consumers (AskService, mindmap handler). Adds `model_chat_test.go`.

**Assessment:**

- **Correctness:** The approach is sound — `[DONE]` is a transport-level signal, not answer text. Using a sentinel error to cleanly abort the loop is idiomatic Go.
- **Edge cases:** The sentinel check with `errors.Is(err, errStreamDone)` is correct since `errStreamDone` is a package-level `errors.New` value. The existing `context.Canceled` / `DeadlineExceeded` path is preserved alongside it.
- **Data loss:** The comment correctly notes that drivers populate `ToolCallsResult`/`UsageResult` before emitting `[DONE]`, so no data is lost.
- **Tests:** The test file uses a `doneEmittingDriver` mock that faithfully replicates the real driver behavior. It asserts `"Hello world"` output without the sentinel. Good coverage of the core scenario. It would be nice to also test an empty-delta stream (only `[DONE]`), but not a blocker.

- **Nitpick (non-blocking):** In `doneEmittingDriver.ChatStreamlyWithSender`, if `sender` returns an error mid-stream (including `errStreamDone` from our new check), the remaining deltas are skipped. That's fine — just noting the behavior in case any driver relies on all deltas being delivered before `[DONE]`. Given the comment that usage is populated before `[DONE]`, this seems safe.

**Verdict:** LGTM — well-structured fix with good test coverage.

PR #17092, opened 2026-07-20
Fix: shared chatbot completion ignores knowledge base and empty response

euvre:
### Summary

The Go `/api/v1/chatbots/<dialog_id>/completions` endpoint (used by the shared/embedded chat UI) bypassed knowledge-base retrieval entirely: it sent the raw conversation history straight to the LLM. As a result, none of the retrieval-related dialog configuration took effect in shared chats — most visibly, the configured **empty response** was never returned when the knowledge base had no matching content, and the LLM answered freely instead. Retrieval, citations, and the system prompt were silently skipped as well.

This PR routes `BotService.ChatbotCompletion` through `ChatPipelineService.AsyncChat` — the same RAG pipeline that serves the regular chat endpoints — mirroring the Python `async_iframe_completion -> async_chat` flow:

- Retrieval now runs for shared chats; when nothing matches, the dialog's configured `empty_response` is returned (same behavior as in-app chat).
- Session creation (no `session_id`) returns the prologue without an LLM call, matching the Python contract.
- Pipeline deltas are accumulated into full-text SSE frames with `final`/`start_to_think`/`end_to_think` markers, keeping wire parity with the Python iframe envelope.
- Each completed turn pair and its retrieval reference is persisted to `api_4_conversation`, so multi-turn shared sessions retain context.
- The request body now accepts `quote` (defaults to false, as in `bot_api.py`), `reasoning`, `internet` (bool or 0/1), and `doc_ids`.

Testing: `internal/service` package tests pass, including new unit tests for the message-building and flag-normalization helpers; all `TestChatbotCompletion_*` handler tests pass.


wangq8 (commented):
## PR Review: #17092 — Fix shared chatbot completion ignoring KB and empty response

**Author:** @euvre  
**Files:** `internal/service/bot.go`, `internal/service/bot_completion.go`, `internal/service/bot_completion_history_test.go`  
**Net:** +343 / −232

### Overall Assessment

This is a solid architectural fix. The core insight is correct: the Go `/api/v1/chatbots/<dialog_id>/completions` endpoint was short-circuiting to a direct LLM call, bypassing the RAG pipeline entirely. Routing through `ChatPipelineService.AsyncChat` brings shared/embedded chat into parity with the in-app chat experience and mirrors the Python `async_iframe_completion → async_chat` flow.

The implementation is clean — SSE frame parity (`final`, `start_to_think`, `end_to_think` markers), session lifecycle (prologue-only new sessions), request field expansion (`quote`, `reasoning`, `internet`, `doc_ids`), and persistence of turn pairs + references are all correctly handled.

### 👍 Whats Done Well

1. **Correct architectural choice** — `ChatPipelineService.AsyncChat` is the shared RAG pipeline; routing through it ensures KB retrieval, citations, empty-response fallback, system prompt, and tool calls all behave identically to the in-app chat endpoints.

2. **Prologue-on-new-session without LLM call** — Mirrors Python `async_iframe_completion` exactly: the share page obtains a `session_id` first, then sends real questions. Avoids an unnecessary LLM round-trip on session creation.

3. **SSE wire parity** — `final` and `start_to_think`/`end_to_think` markers match the Python iframe envelope so the front-end handles both shapes identically.

4. **Defensive `normalizeBotBoolFlag`** — The `reasoning`/`internet` flags can arrive as `true/false` bools or `0/1` numbers depending on widget; the normalize function handles both cleanly.

5. **Error handling** — Pipeline-level errors (`**ERROR**` prefix answer) prevent persistence, matching Python exception path. SSE sanitization preserved.

6. **Old helper removal** — `historyToMessages` and `historyFromMessages` are deleted in favor of `buildChatbotPipelineMessages` / `parseChatbotTurns` which produce `map[string]interface{}` for the pipeline instead of typed `modelModule.Message` structs. Less coupling.

7. **Tests rewritten for new interface** — Tests now cover message filtering (drops leading assistant + system turns), malformed input resilience, and flag normalization.

### 🔍 Issues Raised

#### 1. Concurrency race in `persistChatbotTurn` (Medium)
**File:** `internal/service/bot_completion.go` ~line 551

```go
func (s *BotService) persistChatbotTurn(session ..., question, answer, messageID string, reference map[string]any) error {
    turns := parseChatbotTurns(session.Message)
    turns = append(turns, ...)
    rawMsg, _ := json.Marshal(turns)
    session.Message = rawMsg
    ...
    return s.api4ConversationDAO.Update(session)
}
```

The read–modify–write on `session.Message` is not atomic. Two concurrent SSE requests for the same `session_id` can both read the same stale base, append their own turns, and the last `Update` call wins — silently dropping the other turn. For v1 this may be acceptable (the comment acknowledges "at most the latest exchange" may be lost), but it should be tracked.

**Suggestion:** Consider adding a version/row-lock column on `api_4_conversation`, or re-reading the session immediately before `Update` with an optimistic WHERE clause (`WHERE message = <old> AND id = <id>`). A simpler approach: set `DeleteAfterRun` on the goroutine or use a per-session mutex map.

---

#### 2. CodeRabbit artifact in test comment (Low)
**File:** `internal/service/bot_completion_history_test.go` line ~17

> `// Locks in review Finding 8`

This references an internal CodeRabbit review tracking ID. It is meaningless to future readers and should be removed from committed code.

**Suggestion:** Reword the comment to describe only the runtime behavior, e.g.:
```go
// Tests for the message-building / flag-normalization helpers used by
// BotService.ChatbotCompletion.  A resumed session_id must carry prior
// turns into the next pipeline call so multi-turn chatbot clients retain
// context, and the filtering must match Python async_iframe_completion
// (drop system turns and the leading assistant prologue).
```

---

#### 3. User and assistant turn share the same `messageID` (Low)
**File:** `internal/service/bot_completion.go` ~line 556

```go
turns = append(turns,
    map[string]any{
        "role": "user", "content": question, "id": messageID, ...
    },
    map[string]any{
        "role": "assistant", "content": answer, "id": messageID, ...
    },
)
```

Both turns in the pair get the same `messageID`. If this is intentional to pair them as a Q&A unit, a comment explaining the design would help future readers. If they need to be independently addressable (e.g., for message editing, deletion, or citation linking), they should get distinct IDs.

---

#### 4. Inconsistent defensive guards on `s.pipeline` vs `s.llmService` (Low)
**File:** `internal/service/bot_completion.go` ~line 378

```go
pipeline := s.pipeline
if pipeline == nil {
    pipeline = NewChatPipelineService()
}
```

`NewBotService` always creates a `NewChatPipelineService()`, so this nil check is only relevant in test bootstrap. Meanwhile, `s.llmService` has no such fallback and returns `CodeServerError` when nil.

**Suggestion:** Either remove the defensive fallback here (add a nil-LLM-style guard instead) or document that this is intentionally permissive for testing. Consistency matters.

---

#### 5. Test coverage gaps (Medium)

The old tests covered `historyToMessages`, `historyFromMessages`, malformed input, empty-field skipping, and round-trip fidelity. The new tests cover `buildChatbotPipelineMessages` and `normalizeBotBoolFlag` but miss:

- **`parseChatbotTurns`** — no direct test; only exercised through `buildChatbotPipelineMessages`
- **`persistChatbotTurn`** — not tested at all
- **`referenceOrEmpty`** — trivial but uncovered

The largest gap is the lack of **integration-level handler tests** for `ChatbotCompletion` itself (with a mocked pipeline or DAO). The handler is the entry point for all this logic and has no test coverage for the full request → pipeline → SSE → persistence flow.

---

#### 6. `ChatbotSSEFrame.Reference` may be inconsistent in streaming frames (Minor)
**File:** `internal/service/bot_completion.go` ~line 492

During accumulation, intermediate frames carry `reference: {}` while only the final frame carries the actual retrieval reference. The Python wire format sends an empty reference for intermediate frames, so this matches. However, if `res.Reference` was non-nil on a non-final result (line 487–489), `finalRef` is updated but the SSE frame still sends `map[string]any{}`.

**Suggestion:** Consider whether mid-stream reference updates should be forwarded. Currently this is a no-op in practice (pipeline only populates reference on final results), but documenting this assumption helps.

### ✅ Summary

| Category | Verdict |
|---|---|
| **Fix correctness** | ✅ Fixes the described regression properly |
| **Architecture** | ✅ Correct; matches Python flow |
| **Security** | ✅ No new surface; existing sanitization preserved |
| **Test coverage** | ⚠️ Adequate for helpers, missing for handler |
| **Concurrency** | ⚠️ Potential race on session persistence |
| **Code quality** | ✅ Clean, well-commented, readable |

**Verdict:** Change is sound and should be merged after addressing the CodeRabbit artifact comment and ideally adding handler-level test coverage. The persist race is a known trade-off for v1 and can be tracked separately.

wangq8 (commented):
## PR Review: #17092 — Fix: shared chatbot completion ignores KB and empty response

### Summary

This PR fixes a significant bug where `/api/v1/chatbots/<dialog_id>/completions` bypassed knowledge-base retrieval entirely, routing raw conversation history straight to the LLM. The fix routes `BotService.ChatbotCompletion` through `ChatPipelineService.AsyncChat`, mirroring Python's `async_iframe_completion → async_chat` flow. Overall the architectural direction is correct and the code is well-commented.

---

### ✅ Positives

1. **Core fix is sound** — routing through the RAG pipeline correctly enables KB retrieval, empty_response fallback, citations, and system prompt for shared chatbot completions.

2. **Message filtering matches Python** — `buildChatbotPipelineMessages` correctly drops system turns and leading assistant prologue, matching `conversation_service.py:341-356`.

3. **SSE frame parity** — `Final`, `StartToThink`, `EndToThink` markers added to `WriteChatbotFrame`, matching the Python iframe envelope.

4. **Turn persistence is correct** — `persistChatbotTurn` appends user+assistant pair with matching `messageID` plus retrieval reference, matching Python's `structure_answer` + `append_message` pattern.

5. **Flag normalization** — `normalizeBotBoolFlag` handles bool/int/float inputs. Good.

6. **Error path** — `**ERROR**` prefix check prevents persisting failed turns. Solid.

7. **`stringFromMap` nil-safety** — already handles nil `PromptConfig`. No new nil risk.

---

### 🔴 Must-Fix: Reasoning text is silently dropped

The Go pipeline (`chat_pipeline.go:1098-1108`) emits reasoning text in the `Reasoning` field (not `Answer`):

```go
if inThink {
    out <- AsyncChatResult{
        Reasoning:   text,   // ← not Answer
        Answer:      "",     // defaults to ""
    }
}
```

But the PR code only accumulates `res.Answer`:

```go
fullAnswer += res.Answer  // "" → no change
```

The `StartToThink`/`EndToThink` markers are forwarded correctly, but the reasoning text between them is never sent to the front-end. The front-end receives bracketing markers with no reasoning content.

**Fix**: The delta loop should handle `res.Reasoning` and accumulate it alongside `fullAnswer` (perhaps in a separate accumulator), or forward reasoning frames as SSE frames with the reasoning content in the data payload.

---

### 🟡 Should-Fix: No pipeline integration tests

The PR description states "all TestChatbotCompletion_* handler tests pass", but no such tests exist in the repository. The only tests cover message-building helpers (`buildChatbotPipelineMessages`, `normalizeBotBoolFlag`). The full pipeline path — KB retrieval, empty_response fallback, citation display through `WriteChatbotFrame` — has zero test coverage.

**Suggestion**: At minimum, add a test that verifies the SSE frame sequence for the three key paths:
- Successful KB retrieval → decorated answer with `final:true`
- Empty retrieval → empty_response text returned
- Pipeline error → `**ERROR**` prefix not persisted

---

### 🟡 Worth Noting

**Test removal**: `TestHistoryFromMessages_PreservesOrder`, `TestHistoryRoundTrip_PreservesPriorTurns`, and `TestHistoryToMessages_SkipsEmptyFields` are removed. The filtering is now in `buildChatbotPipelineMessages` (tested) and `parseChatbotTurns` passes ALL turns through. This is architecturally valid but reduces test surface. Consider adding a test for `persistChatbotTurn` since it manages the critical write-back.

**Pipeline initialization in constructor**: `NewBotService` now creates a `ChatPipelineService`. If that service has DB or external dependencies in its init path, this could slow test startup or fail in lightweight test environments (the nil-guard in `ChatbotCompletion` partially mitigates this).

**`Final` frame with empty `Answer`**: When `res.Answer == ""` on final, accumulated deltas stand. This is correct for the `empty_response` path, but relies on the pipeline contract that a non-empty-response final frame always carries a non-empty decorated `Answer`. An explicit assertion or comment documenting this invariant would help.

---

### Verdict

The fix is correct and well-structured. The reasoning-text-dropping issue is a functional regression that should be fixed before merge. The test coverage gap is significant for a 257-line change in a production code path, but not a blocker if manually verified.

euvre:
Thanks for the thorough review. All points addressed in c40c687:

1. **Concurrency race in persistChatbotTurn** — Fixed. `BotService` now carries a fixed-size striped lock set (`persistLocks [64]sync.Mutex`); `persistChatbotTurn` takes the per-session lock and **re-reads the row inside the lock** before appending, so two concurrent requests on the same `session_id` can no longer drop each other's exchange. Covered by `TestPersistChatbotTurn_ConcurrentSameSession`, which fails without the fix.
2. **CodeRabbit artifact in test comment** — Removed; the header now describes only the runtime behavior.
3. **Shared messageID on the turn pair** — Intentional; added a comment explaining the design (a Q&A exchange is addressed as a unit, mirroring the in-app chat pairing convention where the answer id derives from the question id).
4. **Inconsistent nil guards** — `s.pipeline` now fails fast with `CodeServerError` exactly like `s.llmService`; the silent `NewChatPipelineService()` fallback is gone and the comment documents that the checks only guard a hand-rolled zero-value `BotService`.
5. **Test coverage gaps** — Added `TestParseChatbotTurns`, `TestReferenceOrEmpty`, `TestPersistChatbotTurn_AppendsPairAndReference`, `TestPersistChatbotTurn_NilReferenceDefaultsToEmpty`, and the concurrent same-session test. Handler-level request→pipeline→SSE coverage still requires a mockable pipeline seam, which is a larger refactor; prefer tracking that separately.
6. **Mid-stream reference** — Documented the assumption: the pipeline only populates `Reference` on the final result, and intermediate frames intentionally send an empty reference object for wire parity with python.

`bash build.sh --test ./internal/service/` passes.

qinling0210 (approved):
LGTM.

Please address the critical issue coderabbitai commented.

euvre:
Good catch — fixed in 208ee0b. The modulo now runs on the unsigned hash:

```go
return &s.persistLocks[h.Sum32()%uint32(len(s.persistLocks))]
```

so the index is guaranteed non-negative on 32-bit architectures as well. Verified with `bash build.sh --test ./internal/service/` (passes) and a `GOARCH=386` vet sweep (the only 32-bit complaints come from third-party deps like sonic/thrift that don't support 32-bit at all, unrelated to this change).

PR #17095, opened 2026-07-20
fix[go]: skip LLM call for shared chatbot session handshake

euvre:
### Summary

Opening a shared chatbot link immediately shows a fabricated assistant reply such as "it looks like you sent a blank message", even though the user never sent anything.

Root cause: the shared chat page sends an empty `question` without a `session_id` on load purely as a handshake to obtain a session (frontend `fetchSessionId` in `web/src/pages/next-chats/hooks/use-send-shared-message.ts`). The Python `async_iframe_completion` (`api/db/services/conversation_service.py`) treats a request without `session_id` as session creation only — it persists the prologue-seeded session and streams the prologue back without invoking the LLM. The Go port of `ChatbotCompletion` (`internal/service/bot_completion.go`) instead always ran the LLM, feeding the empty string as a user turn, so the model answered a message that was never sent.

Fix: mirror the Python behavior in `ChatbotCompletion`. A request without `session_id` now only creates the prologue-seeded session and streams the prologue frame plus the done marker; the LLM runs exclusively for follow-up turns that carry a `session_id`.

Adds a regression test (`TestBotService_ChatbotCompletion_NewSessionSkipsLLM`) asserting that the handshake returns only the prologue frame, persists exactly one assistant (prologue) turn, and never reaches the LLM path.

### DIFF
--- a/internal/service/bot_completion.go
+++ b/internal/service/bot_completion.go
@@ -264,8 +264,10 @@ func WriteAgentbotFrame(w http.ResponseWriter, f ChatbotSSEFrame) error {
 //
 // The full LLM session-lifecycle implementation is added below. It
 // is a v1 port: it yields a single frame per turn (the Go LLMBundle
-// chat call is non-streaming), seeded with the dialog's prologue
-// when the request creates a new session.
+// chat call is non-streaming). A request without session_id only
+// creates the prologue-seeded session and streams the prologue back;
+// the LLM runs exclusively for follow-up turns that carry a
+// session_id.
 //
 // Authorisation: dialog must exist, belong to the requester's tenant,
 // and have status == common.StatusDialogValid.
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## Review: PR #17095 — Skip LLM call for shared chatbot session handshake

This is a well-scoped and clearly documented fix. Here is my analysis:

### Root cause & fix
The root cause is correctly identified: the Go `ChatbotCompletion` port unconditionally ran the LLM on every request, while the Python `async_iframe_completion` treats a request without `session_id` as a session-creation-only handshake. The fix mirrors the Python behavior by returning early with the prologue when no `session_id` is present.

### Code quality
- The change is minimal (+98/-2) and tightly scoped to the handshake path.
- The comments are thorough and reference both the Python implementation line numbers and the frontend handshake — excellent for maintainability.
- The early return with a buffered channel (cap 2) is clean and correct.
- Normal follow-up turns (with `session_id`) are completely unaffected.

### Regression & security risk
Negligible. The guard condition (`session_id == ""`) only triggers on the handshake path. No new code paths are exposed.

### Tests
The regression test `TestBotService_ChatbotCompletion_NewSessionSkipsLLM` is excellent:
- Passing `nil` for `llmService` is a clever way to prove the LLM is not invoked — any LLM call would panic, making the test a clear pass/fail signal.
- It validates both the frame sequence (prologue + done) and the persisted session state (exactly one assistant prologue turn, no empty user message).
- It exercises the exact scenario described in the PR: opening a share link with no `session_id`.

### Minor observation
The fix treats *all* requests without `session_id` as handshakes, even if they carry a non-empty `question`. This matches the Python behavior (conversation_service.py line 324), so parity is correct — but it is worth noting that the `question` field is silently dropped in this path. This is fine given the architecture (handshake-by-absence-of-session_id is the contract between frontend and backend).

### Verdict
**LGTM.** The fix is correct, well-documented, and properly tested. No blocking issues.

PR #17099, opened 2026-07-20
fix(go-agent): support Aliyun tool calls

Hz-186:
## Summary

  - Enable synchronous and streaming tool calls for Aliyun models.
  - Preserve provider-specific chat endpoints and prevent repeated qwen-flash tool calls.
  - Restrict retrieval tool inputs to model-provided query parameters.

  ## Testing

  - `bash build.sh --test ./internal/entity/models ./internal/agent/component ./internal/agent/tool`
  - Manual frontend UI testing passed.

### DIFF
--- a/conf/models/aliyun.json
+++ b/conf/models/aliyun.json
@@ -18,7 +18,10 @@
       "max_tokens": 995904,
       "model_types": [
         "chat"
-      ]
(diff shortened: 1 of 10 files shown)


yuzhichang:
LGTM

PR #17100, opened 2026-07-20
Feat: Configure the relevant pipeline parameters on the dataset configuration page.

cike8899:
### Summary

Feat: Configure the relevant pipeline parameters on the dataset configuration page.


wangq8 (commented):
## PR Review — #17100

**Title:** Feat: Configure the relevant pipeline parameters on the dataset configuration page.
**Author:** cike8899
**Stats:** +2078 / -178, 34 files changed

---

### Overview

This PR introduces a new dataset configuration page (`/dataset/:id/setting`) that allows users to configure pipeline operator parameters directly on the dataset settings page. It refactors several agent operator forms (Parser, TokenChunker, TitleChunker, Tokenizer, Extractor) to be reusable in this new context via `onValuesChange` and `hideOutputs` props. It also adds backend-language branching (Go vs Python) and deletes a now-unused `use-dataflow-request.ts`.

---

### ✅ What Looks Good

1. **Clean refactoring of operator forms** — Adding `onValuesChange` and `hideOutputs` as optional props to the existing forms is a non-breaking, minimal-change approach. Good reuse pattern.
2. **Proper file deletion** — `use-dataflow-request.ts` is cleanly deleted with no remaining imports elsewhere.
3. **Well-organized new module** — The new `setting/` directory has a clear structure: `utils.ts`, `hooks.ts`, `form-schema.ts`, components, and the main `index.tsx`.
4. **Backward-compatible route addition** — The old `/configuration/:id` route is preserved under `getBackendLanguage() === 'python'`, while the new `/setting/:id` route is gated behind `getBackendLanguage() === 'go'`.
5. **Lossless config transforms** — The round-trip transforms (`transformApiConfigToForm` → `transformFormConfigToApi`) handle the Parser, Extractor, TokenChunker, TitleChunker, and Tokenizer types correctly.

---

### 🚨 Issues to Fix Before Merge

#### 1. Stale `console.log` statements (4 instances)

These should be removed before landing:

- `web/src/pages/dataset/setting/components/added-source-card.tsx:42` — `console.log('item--->', item)`
- `web/src/pages/dataset/setting/components/link-data-source-modal.tsx:36` — `console.log(values, selectedList)`
- `web/src/pages/dataset/setting/components/link-data-source.tsx:97` — `console.log('open', open, data)`
- `web/src/pages/dataset/setting/components/link-data-source.tsx:107` — `console.log('handleLinkOrEditSubmit', data)`

#### 2. Typos

- **`fileterString`** → should be **`filterString`** (missing 'l') in `link-data-source-modal.tsx:20` and its usage throughout.
- **`ststus`** → should be **`status`** in `form-schema.ts:21` (the connectors schema's field name).
- **"Username must be at least 2 characters."** → The `name` field error message in `form-schema.ts` is copied from a user form; should be dataset-appropriate, e.g. "Dataset name is required."

#### 3. Potential `useEffect` dependency issue in `hooks.ts:useFetchDatasetSettingOnMount`

The `useEffect` at line ~107 includes `form` as a dependency. While `useForm` returns a stable reference, `useEffect` with `form.reset(formValues)` could trigger unexpected resets. Consider extracting the reset logic or adding a flag to prevent unnecessary resets after the first mount.

#### 4. `getBackendLanguage()` check creates a coupling

The sidebar (in `sidebar/index.tsx`) duplicates the configuration menu items based on `getBackendLanguage()`:
- Python → `/configuration/:id` (old page)
- Go → `/setting/:id` (new page)

This creates a hard-coded backend-language-driven routing bifurcation. Consider adding a config-level feature flag or route alias instead, so the UI isn't directly tied to backend implementation language.

#### 5. `BuiltinPipelineItem` uses `useFetchBuiltinPipelines` with no error handling

The `useFetchBuiltinPipelines` hook in `common-item.tsx` calls the API with `enabled: true` by default. If the API endpoint (`/v1/pipelines`) is unavailable, the dropdown will silently show no options and the user won't know why. Consider adding a loading state and error feedback.

#### 6. `useFormChangeCallback` uses `any` type

```ts
export function useFormChangeCallback(
  form: UseFormReturn<any>,
  onValuesChange?: (values: any) => void,
)
```

Consider typing the generic parameter instead of using `any`:

```ts
export function useFormChangeCallback<TFieldValues>(
  form: UseFormReturn<TFieldValues>,
  onValuesChange?: (values: TFieldValues) => void,
) {
```

---

### ⚠️ Minor Observations / Suggestions

1. **`pipeline-operator-form.tsx` switch statement** — The form switch on `operatorType` assumes all `Operator` enum values are covered. If a new operator type is added without updating this switch, it silently returns `null`. A TypeScript `never` check in the `default` branch would catch this at compile time.

2. **`gcTime: 0` on queries** — Both `useFetchBuiltinPipelines` and `useFetchPipelineDslByPipelineId` set `gcTime: 0`, meaning cache entries are immediately eligible for garbage collection. This causes a refetch on every mount, even if the data hasn't changed. Consider a reasonable cache time (e.g. 5 minutes for pipeline templates).

3. **`connectors` field in form-schema** — The field `ststus` is a typo. Also, the inline connector schema duplicates what's already typed in `IConnector` from `database/dataset`. Consider reusing the existing interface.

4. **The `BlockButton` `type="button"` fix** — Multiple `BlockButton` components now have `type="button"` added. This is a good defensive fix to prevent accidental form submission inside forms.

5. **Import order in `extractor-form/index.tsx`** — The import of `useOwnerTenantId` was moved down; the original order was fine. Minor style point.

---

### Summary

| Category | Count |
|---|---|
| 🚨 Must fix before merge | 6 |
| ⚠️ Should consider | 5 |
| ✅ Good practices | 5 |

The PR is structurally sound and the feature is well-architected. The main issues are **stale console.log statements** and **typos** that should be cleaned up before merge. The backend-language branching approach is pragmatic but could benefit from a more abstract feature-flag mechanism in the future.

PR #17105, opened 2026-07-20
fix: include filename in file download Content-Disposition header

huang-aoqin:
### Summary

GET /api/v1/files/{id} now sets attachment filename for both Python and Go handlers so browsers can save downloads with the correct name.


6ba3i (changes requested):
Overall the fix looks right but I do think a few things should be addressed before merging:

* The Go `FormatContentDisposition` uses `url.PathEscape` for `filename*`. That is not quite RFC 8187-compatible for some valid filename characters such as `:`, `@`, and `=`, which remain unescaped. Since this helper is becoming the shared formatter, I’d prefer a small RFC 8187-specific encoder here.

* Python and Go currently produce different ASCII fallbacks for the same filenames. For example, Unicode-only names can become `pdf` on Python and `.pdf` on Go, and `filepath.Base` on Linux also does not handle Windows `\` separators the same way Python does. It would be better to define one basename/fallback behavior and keep both implementations consistent.

* The Go side changes the formatter and download handler but doesn’t add matching Go regression tests. A small table-driven test covering ASCII, Unicode, spaces, punctuation and path separators should be enough.

huang-aoqin:
> Overall the fix looks right but I do think a few things should be addressed before merging:
> 
> * The Go `FormatContentDisposition` uses `url.PathEscape` for `filename*`. That is not quite RFC 8187-compatible for some valid filename characters such as `:`, `@`, and `=`, which remain unescaped. Since this helper is becoming the shared formatter, I’d prefer a small RFC 8187-specific encoder here.
> * Python and Go currently produce different ASCII fallbacks for the same filenames. For example, Unicode-only names can become `pdf` on Python and `.pdf` on Go, and `filepath.Base` on Linux also does not handle Windows `\` separators the same way Python does. It would be better to define one basename/fallback behavior and keep both implementations consistent.
> * The Go side changes the formatter and download handler but doesn’t add matching Go regression tests. A small table-driven test covering ASCII, Unicode, spaces, punctuation and path separators should be enough.

@6ba3i done!

6ba3i (approved):
LGTM

PR #17106, opened 2026-07-20
Fix: add new default moonshot model

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/services/provider_api_service.py
+++ b/api/apps/services/provider_api_service.py
@@ -717,11 +717,13 @@ async def verify_api_key(provider_id_or_name: str, api_key: str | dict, base_url
                     continue
                 mdl = ChatModel[provider_name](api_key_str, llm["llm_name"], base_url=base_url, **extra)
 
+                temperature = 1 if llm["llm_name"] in ("kimi-k3", "kimi-k2.7-code") else 0.9
+
                 async def check_streamly():
                     async for chunk in mdl.async_chat_streamly(
                         None,
                         [{"role": "user", "content": "Hi"}],
-                        {"temperature": 0.9},
+                        {"temperature": temperature},
                     ):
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## Review: PR #17106 – Add default moonshot models

**Overview:** This PR adds three new Kimi/Moonshot models (kimi-k3, kimi-k2.6, kimi-k2.7-code) to the LLM factory config and adjusts the temperature used during API key verification for kimi-k3 and kimi-k2.7-code to 1.0 (from the default 0.9).

### What looks good
- The model config entries follow the existing pattern and include appropriate metadata (context size, tool support, tags).
- The temperature change is restricted to the verify_api_key flow only — it won't affect production inference, just the connectivity check.
- The change is small and focused: 21 lines of config + 3 lines of logic.

### Potential issues / questions

1. **PR title vs scope**: The title says "add default moonshot model" (singular), but the diff adds three models. Consider updating the title to reflect the full scope, e.g., "Add new Kimi models (k3, k2.6, k2.7-code)".

2. **kimi-k2.6 tool support**: is_tools is set to true for kimi-k2.6. Is it confirmed that the k2.6 API supports function/tool calling? If not, this could produce confusing errors for users.

3. **kimi-k2.6 max_tokens**: 262144 (256k) — worth confirming this matches the actual model limit from the Moonshot API.

4. **Why no temperature change for k2.6?** The PR adjusts temperature for k3 (1.0) and k2.7-code (1.0), but not for k2.6 (still 0.9). Was this intentional? A brief comment explaining why would help reviewers.

### Minor
- The conf/llm_factories.json diff places the new entries above the existing kimi-thinking-preview entry. If there is a convention for ordering (alphabetical, release order), it may be worth checking.

### Summary
A clean, low-risk change. The new model entries look correct structurally. Just a few clarity questions above. No regressions or security concerns detected.

PR #17112, opened 2026-07-20
Fix: model selection format compat and provider api_key wrapping

dcc123456:
### Summary

Fix: model selection format compat and provider api_key wrapping


wangq8 (commented):
## Review: PR #17112

**Title:** Fix: model selection format compat and provider api_key wrapping  
**Author:** @dcc123456  
**Files:** 7 files, +76/-34

### Overview

This PR addresses two issues: (1) models missing a `model_id` field now fall back to a composite identifier built from `model_name`/`instance_name`/`provider_name`, and (2) provider API key submission now supports passing extra metadata (e.g., `group_id`, `provider_order`, `api_version`) by wrapping `api_key` as an object when these extras are present.

### ✅ What works well

- The `model_id` fallback in `model-tree-select.tsx` and `use-llm-request.tsx` is clean and backward-compatible — when `model_id` exists, behavior is unchanged.
- The union type `ISetDefaultModelRequestBody` is well-typed and preserves the legacy interface.
- The api_key wrapping is non-breaking when no extra fields are present (falls back to a plain string).

### ⚠️ Concerns

**1. Backend compatibility of api_key wrapping (medium risk)**
The most significant change is structural: fields like `group_id`, `api_version`, and `provider_order` were previously sent as top-level request body keys. Now they are nested inside the `api_key` object. This means:
- `generic-api-key-config.ts`: `group_id` and `max_tokens: 0` removed from top-level, `group_id` embedded in `api_key` object.
- `local-llm-configs.ts`: `provider_order` removed from top-level, embedded in `api_key` object.
- `provider-config-map.ts` (Azure): `api_version` removed from top-level, embedded in `api_key` object.

👉 This will silently break if the backend endpoint parses these fields from the top-level request body. Please verify that the backend handlers (`PUT /api/v1/llm/{factory}/instance` and related) have been updated (or already accept) this nested format.

**2. Type safety of the union request body (low risk)**
In `system-setting.tsx`, `parseModelValue(value)` returns `{ model_name, model_instance, model_provider } | null`. If the returned object contains unexpected extra keys or the backend expects a strictly exact shape, the request may silently fail. Consider adding an explicit assertion or interface validation.

**3. No tests (minor)**
No test files are included. Given the non-trivial transformations in the `submitTransform` functions, at least a basic unit test on the `api_key` wrapping logic would improve confidence.

**4. `max_tokens: 0` removal (nit)**
The `generic-api-key-config.ts` previously sent `max_tokens: 0` as a top-level field. It has been removed entirely. If this was intentional cleanup, please confirm — if the backend defaults it, no issue.

### Summary

The overall direction is good — both fixes are sensible improvements. The main concern is backend alignment for the api_key wrapping. As long as the receiving endpoints handle the nested format, this should be ready to merge after verification.

PR #17116, opened 2026-07-20
fix: add accessible check for datasets

Haruko386:
### Summary

As title:

Some operations on datasets have been fixed as they previously lacked authentication, preventing team members from proper or correct usage.


internal/service/dataset/metadata.go
```diff
@@ -13,11 +13,10 @@ import (
...
 func (d *DatasetService) UpdateDocumentMetadataConfig(userID, datasetID, documentID string, req map[string]interface{}) (*entity.Document, common.ErrorCode, error) {
-	if _, err := d.kbDAO.GetByIDAndTenantID(datasetID, userID); err != nil {
-		if dao.IsNotFoundErr(err) {
-			return nil, common.CodeDataError, errors.New("You don't own the dataset.")
-		}
-		return nil, common.CodeServerError, errors.New("Database operation failed")
+	userID = strings.TrimSpace(userID)
+	datasetID = strings.TrimSpace(datasetID)
+	if !d.Accessible(datasetID, userID) {
+		return nil, common.CodeDataError, fmt.Errorf("User '%s' lacks permission for dataset '%s'", userID, datasetID)
```
6ba3i:
```suggestion
		return nil, common.CodeDataError, errors.New("You don't own the dataset.")
```

internal/service/dataset/metadata_config_test.go
```diff
@@ -152,7 +166,36 @@ func TestDatasetServiceUpdateDocumentMetadataConfigRejectsNonOwner(t *testing.T)
 	if code != common.CodeDataError {
 		t.Fatalf("expected data error code, got %d", code)
 	}
-	if err.Error() != "You don't own the dataset." {
+	if err.Error() != "User 'user-1' lacks permission for dataset 'kb-1'" {
```
6ba3i:
```suggestion
	if err.Error() != "You don't own the dataset." {
```

Haruko386: done! thanks for taking time

PR #17118, opened 2026-07-20
fix(web): avoid flashing empty state when returning to agent list

euvre:
### Summary

After opening an agent and navigating back to the agent list page via the "Agents" breadcrumb, the page briefly flashes the "no agents created yet" empty state for one frame before the list renders.

Root cause: `useFetchAgentListByPage` uses `gcTime: 0`, so the list query cache is garbage collected while viewing the agent canvas. On navigating back, the query refetches and `placeholderData` serves `{ canvas: [], total: 0 }` in the meantime, which the page rendered as the empty state.

Fix: use the `loading` flag already returned by the hook and render a blank placeholder while the list is loading with no data, so the empty state only appears after loading has actually finished.

### DIFF
--- a/web/src/pages/agents/index.tsx
+++ b/web/src/pages/agents/index.tsx
@@ -30,6 +30,7 @@ import { useRenameAgent } from './use-rename-agent';
 export default function Agents() {
   const {
     data,
+    loading: listLoading,
     pagination,
     setPagination,
     searchString,
@@ -170,6 +171,8 @@ export default function Agents() {
             </div>
           )}
         </article>
+      ) : listLoading ? (
+        <article className="size-full" data-testid="agents-list"></article>
       ) : (
         <article
           className="size-full flex items-center justify-center"


wangq8 (commented):
## Review: PR #17118 — Avoid flashing empty state when returning to agent list

### What it does
Uses the `loading` flag from `useFetchAgentListByPage` to render a blank `<article>` placeholder while data is being fetched, instead of immediately showing the "no agents" empty state.

### Analysis
- **3 lines added**, minimal footprint.
- Root cause analysis in the PR description is accurate: `gcTime: 0` causes the query cache to be garbage collected, so navigating back triggers a refetch with `placeholderData: { canvas: [], total: 0 }`, which was rendered as the empty state.
- The fix correctly checks `listLoading` before reaching the empty-state branch, so the empty state only renders after loading completes with no data.

### Verdict
✅ **LGTM** — clean fix for a classic React Query stale-while-revalidate UX issue.

PR #17119, opened 2026-07-20
fix(web): make raw memory type non-removable in multi-select

euvre:
### Summary

In the memory configuration dialogs, the `raw` memory type is required (form validation rejects submissions that don't include it), but the `MultiSelect` still rendered a remove ("x") button on the raw badge, allowing users to deselect it via the badge, the dropdown, clear-all, or Backspace.

This PR marks the `raw` option as `disabled` in `MemoryOptions` (`web/src/pages/memories/constants/index.tsx`). The existing option-level `disabled` handling in `MultiSelect` then:

- hides the remove button on the raw badge,
- prevents toggling it off in the dropdown,
- preserves it on clear-all and Backspace (`preserveDisabledValues`).

Both the memory creation dialog (`createMemoryFields`) and the memory settings form (`memory-model-form.tsx`) reuse `MemoryOptions`, so the fix applies to both.

### DIFF
--- a/web/src/pages/memories/constants/index.tsx
+++ b/web/src/pages/memories/constants/index.tsx
@@ -11,7 +11,7 @@ export enum MemoryType {
   Procedural = 'procedural',
 }
 export const MemoryOptions = (t: TFunction) => [
-  { label: t('memories.raw'), value: MemoryType.Raw },
+  { label: t('memories.raw'), value: MemoryType.Raw, disabled: true },
   { label: t('memories.semantic'), value: MemoryType.Semantic },
   { label: t('memories.episodic'), value: MemoryType.Episodic },
   { label: t('memories.procedural'), value: MemoryType.Procedural },


wangq8 (commented):
## Review: PR #17119 — Make raw memory type non-removable in multi-select

### What it does
Adds `disabled: true` to the `raw` option in `MemoryOptions` so the existing MultiSelect component:
- Hides the remove (x) button on the raw badge
- Prevents toggling it off in the dropdown
- Preserves it on clear-all / Backspace

### Analysis
- **Single-line change** in the source of truth (`web/src/pages/memories/constants/index.tsx`). Both the creation dialog and the settings form reuse `MemoryOptions`, so the fix covers both without duplication.
- The `disabled` property is already handled by the MultiSelect component (`preserveDisabledValues`), so this is a clean, minimal fix — no new logic needed.
- Correctly addresses the UX bug: `raw` is a required memory type that form validation enforces, so the UI should not allow deselecting it.

### Verdict
✅ **LGTM** — minimal, correct, well-covered.

PR #17121, opened 2026-07-20
fix: soften ask_summary refusal for partially relevant knowledge

euvre:
### Summary

The `ask_summary` prompt (shared by the Python and Go backends) instructed the model to reply with the exact English string `Sorry, no relevant information provided.` whenever it judged the retrieved knowledge irrelevant to the user's question.

In practice, for borderline queries whose phrasing partially matches knowledge content (e.g. a grade-level query that textually overlaps an era name in the retrieved document), reasoning models follow the instruction verbatim and hard-refuse, even though a helpful summary of the closest relevant content exists. The hardcoded English refusal string also overrides the `Answer in language of user's question` rule, so non-English questions can get an English refusal.

This PR adjusts the prompt so that the model:

- points out the mismatch and still summarizes the closest relevant content when the knowledge is partially relevant, instead of refusing;
- refuses only when the knowledge is completely irrelevant, and always in the language of the user's question.


wangq8 (commented):
## Review: PR #17121 — Soften ask_summary refusal for partially relevant knowledge

### What it does
Revises the ask_summary.md prompt to instruct the model to:
- Still summarize the closest relevant content when knowledge is partially relevant, rather than hard-refusing.
- Only refuse when knowledge is completely irrelevant, and always in the user's language.
Removes trailing exclamation marks from validation messages across 13 locale files for consistency.

### Analysis
**Correctness:** The prompt change is sound in principle. The original hard refusal was too rigid for reasoning models that follow instructions verbatim, often refusing on borderline queries where a helpful summary exists. The new wording gives the model more flexibility.

**Prompt effectiveness caveat:** This is an instructional change — its real-world behavior depends heavily on the target model's instruction-following characteristics. For strict models the "point out the mismatch and still summarize" instruction may still be followed literally; for reasoning models it should improve behavior. No regression risk since the worst case is the model continues to refuse (same as before).

**Locale changes:** Removing trailing exclamation marks from validation messages is consistent with the style work in #17120. Low risk.

### Verdict
LGTM — net positive change, low risk, well-motivated. The prompt softens an overly rigid behavior without removing the safety valve for truly irrelevant content.

PR #17126, opened 2026-07-20
fix(web): regenerate in multi-model chat uses the card's own state and model

euvre:
### Summary

In the multi-model chat view, the regenerate (refresh) button on each model card did not work.

Each `ChatCard` renders and sends messages through its own `useSendSingleMessage` instance, but `regenerateMessage` was wired to a separate page-level `useSendMessage` instance with its own (empty) message state. Clicking regenerate therefore:

- truncated the wrong message state (the card UI was not updated);
- resent an empty history, which overwrote the whole session history via `pass_all_history_messages`;
- streamed the regenerated answer into an invisible state, so nothing appeared in the card;
- dropped the card's `llm_id`, so the request used the dialog's default model and could be rejected with `Cannot use specified model`.

The fix wires regenerate to the card's own `useSendSingleMessage` state via `useRegenerateMessage`, resending with the card's model settings (`llm_id`, temperature, ...), consistent with the normal send path. `sendMessage` is now exported from `useSendSingleMessage`, and `enableThinking`/`enableInternet` are made optional in `NextMessageInputOnPressEnterParameter` to match the regenerate call.


wangq8 (commented):
## Review: PR #17126 — fix(web): regenerate in multi-model chat

Excellent bug fix with a very clear PR description. The root cause analysis is spot-on.

### ✅ What looks good

1. **Root cause correctly identified** — The regenerate button was wired to a page-level `useSendMessage` instance with its own (empty) message state, while each card uses `useSendSingleMessage`. This mismatch caused:
   - Wrong message state being truncated
   - Empty history being sent (overwriting session history)
   - Invisible response (streamed to unused state)
   - Wrong model (dropped `llm_id`)

2. **Correct fix** — Uses `useRegenerateMessage` wired to the card's own `useSendSingleMessage` state via `sendCardMessage`, which wraps `sendMessage` with the card's form values (`llm_id`, temperature, etc.). This is consistent with the normal send path.

3. **Minimal API surface change** — Only `sendMessage` is newly exported from `useSendSingleMessage`, and `enableThinking`/`enableInternet` are made optional in the type. No breaking changes to other consumers.

4. **`sendCardMessage` callback** — Wrapping `sendMessage` with `form.getValues()` ensures the regenerate request uses the correct model settings, not the dialog defaults.

5. **Clean diff** — Only 3 files changed, +33/-14 lines. Focused and minimal.

### ⚠️ Suggestions

1. **Consider adding `useCallback` deps for `sendCardMessage`** — Currently `sendCardMessage` depends on `sendMessage` and `form`. `form.getValues()` is stable (from `react-hook-form`), but it would be more explicit to list `[sendMessage, form]` in the deps array. This is already done correctly: `[sendMessage, form]`. ✅

2. **Edge case: rapid double-click regenerate** — If the user clicks regenerate rapidly twice, could `removeMessagesAfterCurrentMessage` race with an in-flight request? Not introduced by this PR, but worth noting for future hardening.

### Summary

Clean, well-explained fix. No regressions or security concerns. The PR description is exemplary — clearly articulates what was broken and how the fix addresses each symptom. LGTM.

cike8899 (approved):

PR #17127, opened 2026-07-20
Fix(go): get VolcEngine model list

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/internal/entity/models/volcengine.go
+++ b/internal/entity/models/volcengine.go
@@ -51,6 +51,26 @@ func (v *VolcEngine) Name() string {
 	return "volcengine"
 }
 
+// getAPIKey extracts the actual API key from VolcEngine's stored format.
+// VolcEngine stores the api_key as a JSON string like:
+//
+//	{"ark_api_key": "...", "endpoint_id": "..."}
+//
+// Falls back to the raw key string if JSON parsing fails.
+func (v *VolcEngine) getAPIKey(apiConfig *APIConfig) string {
+	if apiConfig == nil || apiConfig.ApiKey == nil {
+		return ""
+	}
+	key := strings.TrimSpace(*apiConfig.ApiKey)
+	var volcConfig struct {
+		ArkAPIKey string `json:"ark_api_key"`
+	}
+	if err := json.Unmarshal([]byte(key), &volcConfig); err == nil && volcConfig.ArkAPIKey != "" {
+		return volcConfig.ArkAPIKey
+	}
+	return key
+}
+
 // ChatWithMessages sends multiple messages with roles and returns response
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
## Review: PR #17127 — Fix(go): get VolcEngine model list

Thanks for this PR! Two key improvements here: proper API key extraction and the overhauled model list parsing with type inference.

### ✅ What looks good

1. **`getAPIKey` method** — Clean approach to extracting the Ark API key from VolcEngine-specific JSON format. The fallback to the raw key string ensures backward compatibility for existing configurations that use plain API keys. Nil-safety check is a nice touch.

2. **Consistent application** — The new key extraction is applied everywhere: `ChatWithMessages`, `ChatStreamlyWithSender`, `Embed`, `ListModels`, and `CheckConnection`. Good coverage.

3. **Dead code removal** — The self-assignment `if baseURL == "" { baseURL = resolvedBaseURL }` in `ListModels` was a no-op and correctly removed.

4. **Model type inference (`inferVolcengineModelTypes`)** — Handles domain-based inference ("Embedding" → embedding), task_type-based inference, and modality-based inference. Reasonable fallback to `"chat"` when nothing matches.

5. **Filtering `Shutdown` models** — Prevents stale models from appearing in the UI.

6. **Two-pass enrichment** — Using `ParseListModel` for static config enrichment first, then overriding with VolcEngine-specific types. This preserves existing enrichment while adding accuracy.

### ⚠️ Suggestions

1. **Duplicate types possible in `inferVolcengineModelTypes`** — If a model has both `output_modalities: ["text"]` and `input_modalities: ["image"]` + `output_modalities: ["text"]`, it will append both `"chat"` and `"image2text"`, which is expected. But if a model has `output_modalities: ["text", "text"]` (unlikely but the API could return duplicates), you would get duplicate `"chat"` entries. Consider using a set/`map[string]bool` for deduplication, or adding `slices.Contains` checks before appending.

2. **Reranking models** — VolcEngine may have reranking models (e.g., `task_type: ["Reranking"]`). Currently these would fall through to the `"chat"` fallback, which is incorrect. Consider adding explicit handling for reranking models or defaulting to `"chat"` only when truly ambiguous.

3. **Minor: `strings.TrimSpace` in `getAPIKey`** — Trimming is good, but if the key itself contains leading/trailing whitespace (unlikely in JSON), the `json.Unmarshal` would fail and it would fall back to the trimmed key — which is fine for plain keys but could mask a malformed JSON key. Not a blocker, just something to be aware of.

### Summary

Solid improvement. The API key extraction fixes a real bug for VolcEngine users whose keys are stored as JSON, and the model list overhaul fixes the model type classification. No regressions observed. The deduplication and reranking suggestions are nice-to-haves, not blockers.

PR #17128, opened 2026-07-20
Implement Tagger in Extractor component in GO

qinling0210:
### Summary

Implement Tagger in Extractor component in GO


yuzhichang:
LGTM

PR #17131, opened 2026-07-20
Fix tag CSV parser splicing wrong text after a multi-line quoted field

devYRPauli:
### Summary

The CSV branch of `rag/app/tag.py` iterates `csv.reader(lines)` with `enumerate` and then uses that row index to look back into `lines[]`:

```python
reader = csv.reader(lines)
    ...
    if len(row) != 2:
        content += "\n" + lines[i]
```

When a quoted field spans multiple physical lines, `csv.reader` consumes several entries from `lines` to emit a single logical row, so `i` stops matching the physical line index. From that point on, every malformed row appends the wrong, already consumed physical line into `content`, silently corrupting the record. There is no exception, so the bad text just ends up in the index.

This is the same bug that was fixed for the Q&A parser in #16881. That fix was applied to `rag/app/qa.py` but not to `rag/app/tag.py`, and `git grep "csv.reader(lines)"` shows `tag.py` is the only remaining occurrence. This change ports the same approach: feed the reader a generator that re-appends newlines and track `reader.line_num` against the previous value to recover the true physical span consumed by each logical row.

Reproduction, using this CSV:

```
"quoted spanning
two lines",tag1
bad_line_no_comma
good,tag2
```

Before:

```
[('quoted spanningtwo lines', 'tag1'), ('two lines",tag1\ngood', 'tag2')]
```

The second record picked up leftover text from the quoted field instead of the malformed line, and the first record lost its internal newline.

After:

```
[('quoted spanning\ntwo lines', 'tag1'), ('bad_line_no_comma\ngood', 'tag2')]
```

An ordinary comma separated file with no multi-line quoted fields produces identical output before and after the change.

Scope is limited to the desync fix. I noticed that the `.csv` branch parses comma separated input only, while the docstring for this parser mentions TAB as the delimiter and the `.txt` branch does delimiter detection. That looks like a separate issue and is deliberately not addressed here.

### DIFF
--- a/rag/app/tag.py
+++ b/rag/app/tag.py
@@ -96,12 +96,16 @@ def chunk(filename, binary=None, lang="Chinese", callback=None, **kwargs):
         fails = []
         content = ""
         res = []
-        reader = csv.reader(lines)
+        reader = csv.reader((line + "\n" for line in lines))
+        prev_line_num = 0
 
+        # line_num tracks the physical span when quoted fields cross lines.
         for i, row in enumerate(reader):
+            raw = "\n".join(lines[prev_line_num : reader.line_num])
+            prev_line_num = reader.line_num
             row = [r.strip() for r in row if r.strip()]
             if len(row) != 2:
-                content += "\n" + lines[i]
+                content += "\n" + raw
             elif len(row) == 2:
                 content += "\n" + row[0]
                 res.append(beAdoc(deepcopy(doc), content, row[1], eng, i))


devYRPauli:
Force-pushed to clean up this branch. The previous head had accidentally picked up a large amount of unrelated work from my working tree (47 files, +1558/-215) on top of the intended change, and it had also drifted into a conflict with main.

The branch is now rebuilt on current main and contains only the intended fix: `rag/app/tag.py`, +6/-2. @xugangqiang apologies for the noise if you already opened the old diff.

The change itself is unchanged in substance and still mirrors the fix already merged into the CSV branch of `rag/app/qa.py`. Re-verified against current main before pushing:

- Input where a quoted field spans physical lines, followed by a malformed row:
  - before: the record ends up with `\nline\nomega`, and `malformed_row_no_tag` is silently dropped
  - after: the record correctly reads `\nmalformed_row_no_tag\nomega`
- Multi-line quoted values are also no longer mashed together (`multilinequoted` -> `multi\nline\nquoted`).
- Plain CSV with no multi-line quoting parses identically before and after, so there is no behaviour change on the common path.

xugangqiang:
LGTM

PR #17132, opened 2026-07-20
fix(models): pass selected models into verify and catch constructor errors

ondery:
## Summary

Two remaining verify-path fixes after #17158 (`api_base` -> `base_url` rename) and #17364 / #17397 (empty-catalog remote model discovery) landed on `main`.

The original scope of this PR also covered "Save all dropping the URL" and "Verify against an empty factory catalog". Those two sides are now fully addressed upstream, so this PR is trimmed to the parts that are **not** yet on `main` (including v0.27.0):

### 1. Pass selected models from the List-models picker into Verify

`useVerifyProvider` now accepts a `modelInfoRef` carrying the models selected in the List-models picker (which are **not** registered as form fields) and folds them into the verify payload. Without this, verify payloads for local / OpenAI-compatible providers often omit `model_info`, so the backend cannot verify the user's actual selection.

### 2. Guard Embedding/Chat model construction

Wrap `EmbeddingModel[...]` / `ChatModel[...]` construction in `try/except` so an init failure is recorded as a per-model `FAIL` (with a message) instead of aborting the whole `verify_api_key` call.

### Review follow-up

Dropped SUCCESS-preserving failure recording per @Lynn-Inf: a later verification failure now overwrites a previous success so the latest status is always shown.

### Already covered upstream (dropped from this PR)

- `api_base` alias on create/update/verify and the frontend `api_base` -> `base_url` rename -> **#17158**
- Verify fallback when the factory catalog is empty (Python `get_model_list` discovery, Go `ListModels` discovery) -> **#17364** and **#17397**

## Test plan

- [ ] On Model Providers, add a VLLM (or OpenAI-API-Compatible) draft with a reachable `base_url` and API key.
- [ ] Click **List models**, select one or more models in the picker.
- [ ] Click **Verify** and confirm success (the selected models are now sent in `model_info`).
- [ ] Click **Verify** without selecting models and confirm the backend still discovers models via #17364.
- [ ] For a model registered with multiple capability types, confirm that a later type failure overwrites an earlier `SUCCESS` for the same `llm_name`.
- [ ] Confirm an Embedding/Chat constructor exception is reported as a per-model failure instead of crashing the verify endpoint.


ondery:
## Note after #17158

#17158 already landed the frontend `api_base` → `base_url` rename, which covers the **Save all dropping the URL** side of this PR.

What remains here is the **Verify** path for local / OpenAI-compatible providers (e.g. VLLM):

- Accept `api_base` as a backend alias for `base_url` (backward compatibility for older clients)
- When Verify runs against an empty factory catalog without `model_info`, fall back to the same List-models connectivity check instead of falsely reporting an invalid API key
- Pass selected models from the instance-card List-models picker into Verify
- Preserve a prior `SUCCESS` when a later capability type for the same model fails

Happy to retitle / trim the description to focus on Verify if that helps review.

Lynn-Inf (commented):
I think a later verification failure should be able to overwrite a previous success — since API key balance or model availability may change over time, the latest status should always be reflected to the user.

As for the frontend changes, I'll ask my colleague @dcc123456 to take over the review on that side.

dcc123456 (approved):

ondery:
Thanks @Lynn-Inf — agreed. The SUCCESS-preservation was meant to avoid a later capability-type failure in the same verify call (unrolled `model_info`) flipping an earlier type's SUCCESS, but reflecting the latest status is the right product behavior.

Dropped `_record_model_verify_failure`. FAIL now overwrites as before.

Still in this PR (not in v0.27.0 / `main`):
1. Frontend: pass List-models picker selection into Verify via `modelInfoRef` (already approved by @dcc123456)
2. Guard Embedding/Chat construction so an init exception is a per-model FAIL instead of aborting the whole `verify_api_key`

Lynn-Inf (approved):
LGTM

Issue #17123, opened 2026-07-20
[Bug]: rag/utils/{es_conn,opensearch_conn}.py write/delete paths silently swallow transient errors


Harsh23Kashyap:
## Problem

Several write and management paths in `rag/utils/es_conn.py` and `rag/utils/opensearch_conn.py` silently swallow transient errors from the OpenSearch / Elasticsearch backend, returning values that look like successful results but actually encode failures. The `for _ in range(ATTEMPT_TIME): ... except Exception: ...; return res` shape (with `ATTEMPT_TIME = 2`) hides a single transient failure as either an empty result, a per-item error string, or a "0 deletions" return, depending on the method. There is also a constructor-level health check on OpenSearch where a fail-fast path can `AttributeError` if the connection never becomes truthy.

This is the search-index counterpart to the storage-backend silent-failure pattern addressed in #17065, #17066, #17069, and #17070.

## Motivation

RAGFlow indexes tens of thousands of chunks per dataset and indexes are the source of truth for retrieval. When a transient connection blip on Elasticsearch or OpenSearch causes `insert()`, `delete()`, `search()`, `health()`/`ping()`, or the constructor to silently return a fake "success", users see:

- chunks that were never indexed and therefore never retrieved, with no error visible to the caller;
- deletes that "succeeded" with `0` row counts but actually hit a connection reset;
- service initialization that crashes mid-startup with a confusing `AttributeError: 'NoneType' has no attribute 'ping'` instead of a clear "OpenSearch unhealthy in N seconds" error.

These regressions surface as silent user-facing loss of data. The same class of bug on the storage side was filed in #17065 and addressed by #17066 / #17070; this is the indexing-side counterpart.

## Background

RAGFlow uses two doc-store backends: Elasticsearch (`rag/utils/es_conn.py`) and OpenSearch (`rag/utils/opensearch_conn.py`). Both use the same `for _ in range(ATTEMPT_TIME):` retry shape. Concretely:

- `es_conn.py:ATTEMPT_TIME = 2`
- `opensearch_conn.py:ATTEMPT_TIME = 2`

The methods below are the ones whose current behavior fails to retry-or-raise on transient errors:

### `rag/utils/es_conn.py`

[...]

PR #17135, opened 2026-07-20
fix(memory): retry transient ES errors in write paths and raise on final attempt

### LINKED ISSUE #17134: [Bug]: memory/utils/es_conn.py silently swallows transient ES errors in insert/update/delete
### Problem

`memory/utils/es_conn.py` swallows transient errors in three write paths that the agent-memory backend relies on. After `ATTEMPT_TIME` retries on `ConnectionTimeout`, control falls out of the loop and the methods return a value that looks like success to callers.

Specifically:

- `insert()` (line 385) logs every retry at `exception` level, falls out of the loop, then returns `res` from the last attempt. If every attempt timed out, `res` is still `[]`, which is the per-item bulk partial-error contract, but the caller cannot distinguish "no items failed" from "all `ATTEMPT_TIME` attempts timed out and the bulk request never executed".
- `update()` has two branches. The single-message branch (line 419) loops `ATTEMPT_TIME` and on any exception logs and `break`s, then `return False`. The query-by-query branch (line 484) catches only `ConnectionTimeout`. On final-attempt timeout it falls through to `return False`. Other exceptions outside `ConnectionTimeout` are caught, logged, and break out, also returning False silently.
- `delete()` (line 529) catches `ConnectionTimeout`, falls through, then `return 0`. Non-timeout exceptions are caught into a logger.warning call, with `not_found` returning 0. Same shape as before, but the `not_found` check sits inside the loop, so a single non-not_found failure on the final attempt returns 0 too.

### Motivation

Agent-memory writes (insert/update/delete) go through this `ESConnection`. Every silent `return False / return 0 / return res` that is actually an unreported timeout hides data loss from the caller, the API handler, and the user. A retry that succeeds masks a transient timeout, which is fine. A retry that exhausts and silently returns "no rows changed" is not.

This is a parallel gap to `rag/utils/{es_conn, opensearch_conn}.py`, addressed in #17125. The memory backend inherited the same original shape (`ATTEMPT_TIME = 2`, no final-attempt raise, no exponential backoff) and was not touched there. The blast radius is smaller than the main document engine, but the symptom is the same.

### Background

The agent-memory backend is selected from `common/settings.py` and instantiated when `DOC_ENGINE=es` (the default). Production deployments that route agent-memory traffic through Elasticsearch exercise these paths on every conversation turn.

Last meaningful change to the file before the bug: #16585 (ruff reformat). The retry shape predates the memory-feature merge.

### Current behavior

- `insert()` returns `[]` after `ATTEMPT_TIME` timeouts. The caller treats `[]` as "0 per-item errors", not "request failed".
- `update()` returns `False` after `ATTEMPT_TIME` timeouts or any non-timeout exception. The caller treats `False` as a normal "no such document" outcome on `condition`, even when the ES request never ran.
- `delete()` returns `0` after `ATTEMPT_TIME` timeouts. A timeout that retries and exhausts is indistinguishable from a delete-by-query that genuinely matched 0 documents.

All three paths retry in a tight loop with `time.sleep(3)` and `self._connect()` between attempts. Backoff is constant at 3s, not exponential. No method raises on the final attempt.

### Expected behavior

- `insert()`, `update()`, `delete()` retry on `ConnectionTimeout` with exponential backoff.
- Each method re-raises the underlying exception on the final attempt instead of returning a sentinel value.
- Non-timeout exceptions during the loop are re-raised immediately (no retry). Pre-existing behavior for `not_found` in `delete()` is preserved: still return 0.
- The per-item bulk partial-error string contract in `insert()` is unchanged: callers still get `[id:error, ...]` strings when individual bulk items fail.

### Proposed solution

Mirror the shape we landed in #17125 for `rag/utils/es_conn.py`:

- Replace `ATTEMPT_TIME = 2` with `MAX_RETRIES = 3`, alias `ATTEMPT_TIME = MAX_RETRIES` to keep every other site untouched.
- `insert()`: change the `for _ in range(ATTEMPT_TIME):` loop to a `for attempt in range(MAX_RETRIES):` shape. Catch `ConnectionTimeout`, log it with attempt counter, `self._connect()`, `time.sleep(2 ** attempt)`, `continue`. Catch `Exception` once at the bottom of the loop, log it, and `raise` instead of `res.append(str(e))`. Final-attempt re-raise goes after the loop.
- `update()` query-by-query branch: same `ConnectionTimeout` retry plus final-attempt re-raise. Non-timeout exceptions already log and return False. Change them to re-raise so the single-message branch can stay minimal.
- `delete()`: same `ConnectionTimeout` retry. Keep the `not_found` short-circuit returning 0. Final-attempt `ConnectionTimeout` re-raises.

### Alternatives considered

- Drop retry entirely and let the SDK exception bubble. Rejected: regresses behavior for transient single-attempt flapping. A bounded retry with explicit backoff is the smaller change.
- Use a single shared retry helper. Rejected: `rag/utils/{es_conn, opensearch_conn}.py` and `memory/utils/es_conn.py` are independent of each other and already have three local conventions. A shared helper is a separate refactor and out of scope.
- Use `tenacity` or another retry library. Rejected: the file already imports only stdlib plus elasticsearch. Adding a dependency for one module is disproportionate.

### Scope

- One file: `memory/utils/es_conn.py`.
- One variable rename (`ATTEMPT_TIME = 2` to `MAX_RETRIES = 3`).
- Three method edits: `insert()`, `update()`, `delete()`.
- No new imports.

### Acceptance criteria

- `insert()`, `update()`, `delete()` retry on `ConnectionTimeout` with `2 ** attempt` exponential backoff and reconnect between attempts.
- Each method re-raises the underlying exception on the final `ConnectionTimeout` instead of returning `[]` / `False` / `0`.
- Non-timeout exceptions during retry loops are re-raised immediately.
- `delete()` keeps returning 0 on the literal `not_found` substring case.
- `insert()` keeps returning the per-item bulk partial-error list (formatted as `id:error` strings) when individual items fail.
- `python -m py_compile memory/utils/es_conn.py` succeeds.
- `ruff check memory/utils/es_conn.py` reports clean.
- No new file is added; the only edited file in the diff is `memory/utils/es_conn.py`.

### Backward compatibility

- Public method signatures are unchanged. Callers in `api/apps/services/memory_api_service.py`, `api/db/services/memory_service.py`, `api/db/joint_services/memory_message_service.py`, and `test/unit_test/memory/utils/` continue to compile and behave identically on the happy path.
- Callers that previously saw a silent `False` / `0` / `[]` on transient timeout will now see the underlying `ConnectionTimeout` propagate. This is the intended change. Callers that already handle `ConnectionTimeout` continue to work; callers that don't will need a try/except, but the only known callers in this repo raise immediately or wrap the call, so propagation is the safer behavior.

### Risks

- Any caller in the agent-memory write path that previously relied on the silent `False` / `0` / `[]` signal will now crash on transient timeout instead of logging. The affected paths are read-after-write in normal use; a raised `ConnectionTimeout` is the correct signal and the upstream handler in `memory_message_service` already swallows such exceptions at the service boundary. No data corruption is introduced either way.

Harsh23Kashyap:
## Summary

`memory/utils/es_conn.py` swallows transient `ConnectionTimeout` errors in three write paths that the agent-memory backend relies on (insert/update/delete). After exhausting `ATTEMPT_TIME` retries on a timeout, the methods returned a sentinel value (`[]` / `False` / `0`) that callers cannot distinguish from a legitimate no-op.

This change brings the memory backend's retry semantics in line with the fix landed in #17125 for the main document engine: bounded retries on `ConnectionTimeout` with exponential backoff, and a re-raise of the underlying exception on the final attempt so a silent timeout never masquerades as success.

Fixes #17134.

## What changed

- `MAX_RETRIES = 3` introduced; `ATTEMPT_TIME` aliased to it so every other call site in the file stays untouched.
- `insert()`: retry loop now logs attempt counter, sleeps `2 ** attempt` on `ConnectionTimeout`, calls `self._connect()` between attempts, and raises on the final attempt. Non-timeout exceptions are logged and re-raised immediately instead of being appended to the per-item bulk partial-error list.
- `update()` single-message branch: same shape — retry on `ConnectionTimeout` with backoff and final-attempt raise, non-timeout exception re-raised.
- `update()` query-by-query branch: same shape — retry on `ConnectionTimeout` with backoff and final-attempt raise, non-timeout exception re-raised immediately.
- `delete()`: retry on `ConnectionTimeout` with backoff and final-attempt raise. The literal `not_found` short-circuit returning 0 is preserved.

The per-item bulk partial-error string contract in `insert()` (returning `id:error` strings when individual items fail) is unchanged.

## Files changed

- `memory/utils/es_conn.py` (1 file, +31 / -15)

## Verification

- `python -m py_compile memory/utils/es_conn.py` — OK
- `uv run ruff check memory/utils/es_conn.py` — clean
- `uv run pytest test/unit_test/memory/ -v` — 14/14 pass
- Precedent: `rag/utils/{es_conn, opensearch_conn}.py` retry-semantics fix landed in #17125 with the same shape. The memory backend is selected from `common/settings.py` when `DOC_ENGINE=es` (the default), so production deployments exercising agent-memory traffic through Elasticsearch now see consistent retry semantics across both layers.

## Backward compatibility

- Public method signatures are unchanged.
- Behavior on the happy path is identical.
- Callers that previously relied on the silent sentinel value on transient timeout will now see the underlying `ConnectionTimeout` propagate. The known in-repo callers (`api/apps/services/memory_api_service.py`, `api/db/services/memory_service.py`, `api/db/joint_services/memory_message_service.py`) already wrap or raise on these exceptions at the service boundary, so propagation is the safer behavior. No data corruption is introduced.

## Risks

The change is a behavior flip on transient timeout only — the failure mode shifts from "silent loss" to "propagated exception." Service-boundary handlers already absorb this exception class, so the user-visible impact is a logged timeout instead of a logged success.

## Follow-ups

- The same retry shape applies to `memory/utils/infinity_conn.py` if/when an Infinity-backend equivalent is needed. Not in scope here because the issue is specific to the ES path and the Infinity module was not investigated in this PR.
- The `memory/utils/es_conn.py` `__init__` / `_connect` paths inherit correctness from `ESConnectionBase`; no change needed.

## Related

- #17134 (issue this PR fixes)
- #17125 (precedent for the main document engine)
- #17123 (issue filed for #17125)
- #17070 (precedent for storage backend retry)
- #17066 (precedent for S3 retry)


qinling0210 (changes requested):
Nice catch on the silent ConnectionTimeout swallows — failing loudly on write paths is the right direction, and the reconnect + backoff is a genuine improvement.

However, the "Backward compatibility" section doesn't hold against the current callers, and I found a few inaccuracies in the description:

Callers don't actually wrap these exceptions. embed_and_save() at api/db/joint_services/memory_message_service.py:248 consumes insert()'s return value as the failure list and has no try/except. With this change, a raised exception propagates through queue_save_to_memory_task and aborts the loop over memory_ids — memories after the failing one are no longer saved, and the per-memory failure attribution (failed_memory) is lost. Worse, on the agent Message component's run_coroutine_threadsafe path (agent/component/message.py:301), the exception is silently swallowed by the unawaited Future — the exact silent-loss behavior this PR aims to fix. Please either harden these callers or fix the description.

Startup crash risk: fix_missing_tokenized_memory() (api/db/joint_services/memory_message_service.py:349) runs at server boot (init_data.py:153 → ragflow_server.py:115) with no try/except. A transient ES failure during startup (3 retries ≈ 3s) previously skipped silently; now it raises and fails backend startup. This looks like a regression.

delete_memory ordering: api/apps/services/memory_api_service.py:243 deletes the MySQL row before delete_message(). If the latter now raises, the API returns 500 but the memory row is already gone — DB/ES divergence. Consider wrapping or reordering.

"Every other call site stays untouched" is inaccurate: ATTEMPT_TIME = MAX_RETRIES changes the retry count of the read paths (search, get, get_forgotten_messages, get_missing_field_message) from 2 → 3.

The #17125 precedent claim doesn't match the tree: rag/utils/es_conn.py still uses the old semantics (ATTEMPT_TIME = 2, generic exceptions appended/break). This change diverges the two files rather than converging them.

Backoff wording: time.sleep(2 ** attempt) never sleeps 4s — the final attempt raises before sleeping. Actual wait is 1s + 2s = 3s, roughly the same as before.

No tests cover the new retry/raise semantics. Given the behavior flip, unit tests would be valuable.

Suggested changes: wrap embed_and_save's insert call and fix_missing_tokenized_memory, add a done-callback for the run_coroutine_threadsafe path, drop the now-unreachable trailing returns, and add tests.

PR #17143, opened 2026-07-21
Refactor: unify navbar icons through HomeIcon wrapper

dcc123456:
### Summary

Refactor: unify navbar icons through HomeIcon wrapper


wangq8 (commented):
## Review: Unify navbar icons through HomeIcon wrapper

Thanks for this refactor — it is a clean consolidation of the icon system.

### ✅ What looks good

- All six custom SVG icons (`datasets`, `chats`, `searches`, `agents`, `memory`, `file`) exist in `web/src/assets/svg/home-icon/` with both regular and `-bri` (bright theme) variants, so there are no missing-asset regressions.
- The `MenuItemsIcon` wrapper is a reasonable bridge between `HomeIcon` (which uses `name` + `imgClass`) and the `icon` prop convention (`ComponentType<{ className? }>`).
- The desktop navbar is unaffected — it only renders the `<Icon>` for the Home route (`LucideHouse` is untouched); all other routes show text labels, so `MenuItemsIcon` is effectively mobile-only.

### 🔶 Suggestions

1. **`icon_name` leaks into `<Link>` via `...props` in the desktop nav components.**  
   Since `menuItems` entries now carry an `icon_name` property, and the desktop navbar components destructure only `{ path, name, icon: Icon, ...props }`, the `icon_name` ends up spread onto `<Link>`. React Router may forward it to the DOM as an unknown attribute, which is harmless but not ideal. Consider filtering it out in the desktop map, or keeping icon metadata out of the spread path:

   ```tsx
   // e.g. in DesktopNavbarWithAnchor map callback:
   {menuItems.map(({ path, name, icon: Icon, icon_name: _icon_name, ...props }) => {
   ```

2. **Widened `icon` type in `MobileNavItem`.**  
   The signature now requires `React.ComponentType<{ className?: string; name?: string }>`, but `LucideHouse` has no `name` prop. It works (React tolerates extra props) but the type is slightly over-permissive. A union or generic type would be more precise. Not a blocker.

3. **Visual consistency worth spot-checking.**  
   The custom SVGs may have different padding/stroke-weight than the Lucide icons they replace. Since the mobile nav is the only place these render, a quick visual check in both light and dark themes would be worthwhile.

### Verdict

No regressions, no security concerns, no breaking changes. A tidy, low-risk refactor. Consider addressing suggestion (1) to keep the prop surface clean, but otherwise LGTM.

---

PR #17146, opened 2026-07-21
Fix stale references shown while streaming and empty final reply in Go mode

euvre:
### Summary

In Go mode, when a previous turn cited documents but the current turn cannot, the chat UI shows the previous turn's referenced documents under the thinking bubble; once thinking completes, the reply is replaced with an empty one. The same stale documents also flash when starting a brand-new conversation.

Root causes:

1. The pending assistant message (no content/reference of its own yet) fell back to the conversation-level reference list by index. That index only aligns with completed answers, so the streaming placeholder surfaced stale references from a previous turn or from a previously opened conversation.
2. The `empty_response` path in the Go chat pipeline yielded its final event with an empty answer, while Python (`dialog_service.py`) yields the full fallback text in the final event. Consumers that read the final event's answer (e.g. the OpenAI-compatible endpoint) therefore received an empty reply.

Changes:

- `internal/service/chat_pipeline.go`: the `empty_response` final event now carries the full fallback answer, matching Python.
- `internal/service/chat_session.go`: non-streaming accumulation loops treat a non-empty final answer as the complete answer (replace, not append), avoiding duplicated text.
- `web/src/pages/next-chats/utils.ts`: `buildMessageItemReference` never resolves a pending (still-generating) assistant message's reference from the conversation reference list.
- `internal/service/chat_session_test.go`: align the test mock with the real pipeline contract (final result carries the complete answer).

### DIFF
--- a/internal/service/chat_pipeline.go
+++ b/internal/service/chat_pipeline.go
@@ -849,7 +849,11 @@ func (s *ChatPipelineService) AsyncChat(
 		// Two results are yielded (mirroring Python dialog_service.py):
 		//   1. Final=false — carries the answer text so streaming consumers
 		//      actually display the fallback message.
-		//   2. Final=true   — closes the stream with the reference/prompt.
+		//   2. Final=true   — closes the stream with the same full answer plus
+		//      the reference/prompt. Python yields the full answer again in the
+		//      final event (dialog_service.py:807); consumers that only look at
+		//      the final event (e.g. the OpenAI-compatible endpoint) would
+		//      otherwise see an empty reply.
 		if len(knowledges) == 0 {
 			if emptyResp, ok := promptConfig["empty_response"].(string); ok && emptyResp != "" {
 				out <- AsyncChatResult{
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
## Review: PR #17146

Thanks for this well-diagnosed fix. The root cause analysis is clear and the changes are targeted.

### ✅ What works well

1. **`chat_pipeline.go`** — The `empty_response` final event now carries the full fallback answer, matching Python `dialog_service.py`. This fixes the OpenAI-compatible endpoint (and any consumer that only reads the final event) getting an empty reply.

2. **`chat_session.go`** — The `Final && Answer != ""` guard correctly handles the duplication that would otherwise occur when the pipeline yields the same answer in both the non-final and final events. This is safe for the normal LLM non-streaming path too: there, only a single final event is yielded, so `Reset()+WriteString()` produces the same result as the old append-only path.

3. **`utils.ts`** — The `isPendingAnswer` guard correctly prevents pending (still-generating) assistant messages from resolving stale references via the conversation-level reference index. The fallback to `{ doc_aggs: [], chunks: [], total: 0 }` is the right empty default.

4. **`chat_session_test.go`** — Test mock updated to match the new pipeline contract (final result carries the complete answer), and the test still passes because both old and new accumulation paths produce `"Hello world"`.

### 💡 Minor suggestions

- **`chat_session.go`** — The same `Final && Answer != ""` logic appears in both `Completion` and `ChatCompletions`. Consider extracting a helper (e.g., `accumulateAnswer`) to avoid future drift between the two. Non-blocking.

- **`chat_pipeline.go` line 864** — The comment on line 852 says `"Final=true — closes the stream with the reference/prompt"` but after this change it also carries the answer. Worth updating the comment to match the new behavior for clarity.

Overall: LGTM. Thanks for the thorough fix! 🚀

qinling0210 (approved):
LGTM

PR #17147, opened 2026-07-21
fix(deepdoc): guard docling crop against missing page images

Yurii214:
### Summary

`DoclingParser.crop()` indexes `self.page_images` with no guard (`deepdoc/parser/docling_parser.py:181`, and again at the `img0`/`page` lookups below it).

Two realistic conditions lead there with an empty or too-short `page_images`:

- `__images__` sets `self.page_images = []` when local pdfplumber rendering raises, while the remote docling backend still returns text and bboxes — so chunks carry position tags with no page images behind them.
- `_make_line_tag` emits a `@@page\t...##` tag even when its own coordinate-mapping guard fails, and the tag can name a page outside the rendered range.

`crop()` is called from `tokenize_chunks` with `need_position=True`; the resulting `IndexError: list index out of range` is not caught there (only `NotImplementedError` is), so ingestion of the whole document aborts.

The same file already guards this in `cropout_docling_table` (empty check plus `0 <= idx < len(self.page_images)`), and the sibling `pdf_parser` / `mineru_parser` `crop()` implementations do the same — `docling_parser.crop` is the outlier.

This bails out when there are no page images, and drops positions naming pages outside the rendered range, matching those existing patterns.

### Tests

Regression tests added to `test/unit_test/deepdoc/parser/test_docling_parser_remote.py`, reusing its module loader. Both fail with `IndexError` at `docling_parser.py:181` without the guard and pass with it.

### DIFF
--- a/deepdoc/parser/docling_parser.py
+++ b/deepdoc/parser/docling_parser.py
@@ -174,6 +174,20 @@ class DoclingParser(RAGFlowPdfParser):
         if not poss:
             return (None, None) if need_position else None
 
+        # a position tag is emitted even when page rendering failed (__images__ leaves
+        # page_images empty), and a tag can name a page beyond the rendered range, so
+        # indexing page_images below would raise IndexError. mirror
+        # cropout_docling_table and the sibling parsers: bail out when there are no
+        # page images and drop positions that fall outside them.
+        if not getattr(self, "page_images", None):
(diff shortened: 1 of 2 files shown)


Yurii214:
Applied the logging suggestion in `d9647ee` — each position dropped for an out-of-range page index now logs a warning, matching how the sibling pdf/mineru `crop()` implementations report skipped positions.

PR #17148, opened 2026-07-21
fix(web): prevent search popover from flashing at top-left on checkbox toggle

euvre:
### Summary

On the search page, toggling a document checkbox in the file list popover briefly flashes a search box at the top-left corner of the viewport.

**Root cause:** toggling a checkbox fires the `testChunk` and `testChunkAll` mutations. While they are pending, `useSelectTestingResult`/`useAllTestingResult` read the latest mutation's `state.data`, which is `undefined`, and fall back to an empty document list. The `RetrievalDocuments` popover trigger (`{useDocuments?.length && <Button>}`) is therefore unmounted for a moment; with its anchor gone, Radix positions the popover content (which contains the search input) at the top-left of the viewport until the request finishes.

**Fix:** keep the last non-empty document list via a ref so the popover anchor stays mounted during refetches, and use an explicit `length > 0` check to avoid rendering a literal `0` as the `PopoverTrigger` child when the list is empty.

### DIFF
--- a/web/src/pages/next-search/retrieval-documents/index.tsx
+++ b/web/src/pages/next-search/retrieval-documents/index.tsx
@@ -22,7 +22,7 @@ import {
 } from '@/hooks/use-knowledge-request';
 import { cn } from '@/lib/utils';
 import { CheckIcon, ChevronDown, Files, XIcon } from 'lucide-react';
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 interface IProps {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice fix! The root cause analysis in the PR description is spot-on — the combination of React rendering `0` as a literal text node (from `{0 && <Button>}`) and the mutation state returning `undefined`/`[]` mid-flight was causing the popover anchor to unmount, which Radix then positions at (0,0).

**What works well:**
- The `useDocuments?.length > 0` check correctly prevents React from rendering a `0` text node where `false` is expected.
- The `useRef` approach is idiomatic and keeps the anchor mounted during pending mutations, eliminating the flash.

**One minor edge case to consider:** If a mutation returns a genuinely empty document list (`documents: []`), `latestDocuments?.length` is `0` (falsy), so `documentsRef.current` will never update. The popover would keep showing stale document data from the previous non-empty result. This is unlikely to cause real issues in practice (empty results after a search change would be rare), but something to be aware of — theres no mechanism to reset the ref on intentional clears.

Otherwise LGTM. Thanks for the thorough PR description detailing the root cause!

wangq8 (commented):
Clean fix, nicely diagnosed! The root cause explanation in the PR description is excellent — makes the review trivial.

**✅ What looks good:**
- The ref-based approach correctly prevents the popover anchor from being unmounted while a mutation is pending.
- `documentsRef.current` is only updated when there is real data (`latestDocuments?.length`), so the stale frame during the flash window is the last known-good state.
- The `useDocuments?.length > 0` change is a real fix — `?.length` on an empty array returns `0`, which is truthy as a JSX child and React would render the literal digit `0` inside `<PopoverTrigger>`. `> 0` ensures a clean boolean.

**Minor thought (non-blocking):**
If the user clears *all* documents from the test (sets an empty selection), the ref will keep the previous non-empty list until the next render cycle. In practice this is fine since this only happens when the selection list is genuinely cleared, and the next request will update it. Just something to be aware of.

LGTM — ship it! 🚀

dcc123456 (approved):

PR #17149, opened 2026-07-21
fix(web): show completed label for think block after streaming ends

euvre:
### Summary

The collapsible think block in the chat UI used a hardcoded `Thinking...` summary, so the in-progress label remained even after reasoning had finished (the block only renders once `</think>` has closed, i.e. when thinking is already done).

This PR threads a loading-aware summary through `replaceThinkToSection`:

- `replaceThinkToSection` accepts an optional `summary` parameter (defaults to `Thinking...`, other call sites unchanged).
- Both markdown content components (`next-markdown-content` and `markdown-content`) pass `Thinking...` while the message is streaming and `Thought` once it completes.
- Adds the `chat.thought` locale key (English; other languages fall back to English via i18n `fallbackLng`).


wangq8 (commented):
Nice, clean fix — this correctly addresses the UX issue where the think block always showed "Thinking..." even after reasoning completed.

**What works well:**
- The `summary` parameter with default `"Thinking..."` is backward-compatible — existing call sites in `floating-chat-widget-markdown.tsx` and `next-search/markdown-content/index.tsx` continue working unchanged.
- `useMemo` dependency arrays are properly updated to include `loading` in both components.
- The `chat.thought` i18n key is correctly added to the English locale; other languages fall back via `fallbackLng`.

`floating-chat-widget-markdown.tsx` (line 56) and `web/src/pages/next-search/markdown-content/index.tsx` (line 61) also accept a `loading` prop and call `replaceThinkToSection`. They would benefit from the same `thinkSummary` treatment. Consider updating them in a follow-up.

**Verdict:** LGTM. No regressions, no safety/performance concerns.

dcc123456 (approved):

PR #17151, opened 2026-07-21
fix: preserve compilation template kinds and names

buua436:
## Summary

Fix Pipeline knowledge compilation template type handling and display names.

## Changes

- Preserve `knowledge_graph` and `page_index` instead of normalizing them to `timeline`.
- Run timeline-specific chain validation and isolated-entity cleanup only for actual timeline templates.
- Resolve Pipeline template names directly by `template_id` when document group metadata is unavailable.
- Prevent opaque template IDs from being returned as `template_name`.

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -166,10 +166,7 @@ def _get_dataset_tenant_id(dataset_id):
 def _compilation_template_kind(kind) -> str:
     if not isinstance(kind, str):
         return ""
-    normalized = kind.strip().lower().replace("-", "_")
-    if normalized in {"pageindex", "page_index", "knowledge_graph"}:
-        return "timeline"
-    return normalized
+    return kind.strip().lower().replace("-", "_")
 
 
 def _resolve_reference_metadata(req: dict, search_config: dict | None = None):
@@ -581,6 +578,7 @@ async def get_document_structure_graph(tenant_id, dataset_id, document_id):
     """
     from rag.nlp import search
     from api.db.services.compilation_template_group_service import CompilationTemplateGroupService
+    from api.db.services.compilation_template_service import CompilationTemplateService
 
     if not KnowledgebaseService.accessible(kb_id=dataset_id, user_id=tenant_id):
         return get_error_data_result(message=f"You don't own the dataset {dataset_id}.")
@@ -624,6 +622,7 @@ async def get_document_structure_graph(tenant_id, dataset_id, document_id):
     seen_configured_ids: set[str] = set()
     template_meta: dict[str, dict] = {}
     template_meta_by_kind: dict[str, list[dict]] = {}
+    template_meta_by_id: dict[str, dict] = {}
     for group_id in group_ids:
         group = CompilationTemplateGroupService.get_saved(group_id, tenant_id)
         if not group:
@@ -648,6 +647,7 @@ async def get_document_structure_graph(tenant_id, dataset_id, document_id):
                 "kind_norm": kind_norm,
             }
             template_meta[template_id] = meta
+            template_meta_by_id[template_id] = meta
             template_meta_by_kind.setdefault(kind_norm, []).append(meta)
 
     # Load every graph row for this doc in one shot. Each row corresponds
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
## Review Summary

This PR correctly addresses the issue of `knowledge_graph` and `page_index` template kinds being collapsed into `timeline`. The approach is clean and well-scoped.

### What works well

- **Template kind preservation** — removing the normalization mapping in both `_compilation_template_kind` and `_struct_normalize_kind` is the right fix. `_struct_supported_type` continues to work because it compares the normalized kind against the inferred autotype, and both derive from the same `kind` field.

- **Pipeline compiler fallback** — the direct `CompilationTemplateService.get_saved` lookup in `get_document_structure_graph` correctly handles the case where Pipeline templates are not reachable through the document group metadata. The guard (`bucket_id not in template_meta_by_id`) avoids redundant lookups.

- **Timeline-specific cleanup** — scoping `cleanup_timeline_isolated_entities` to only `timeline` templates is correct. Running it after all flushes complete (not mid-compilation) avoids false-positive isolated-entity detection from partial data. The `try/except` wrapper with `logging.exception` prevents a cleanup failure from killing the whole compilation task.

- **Graph rebuild after cleanup** — calling `rebuild_structure_graph_json` within the cleanup function correctly handles both the isolated-entity case and the no-relation case (where every entity is orphaned).

### Minor observations (non-blocking)

1. **Legacy `compile_kwd` mismatch on recompile** — documents compiled before this change with `knowledge_graph`/`page_index` templates have `compile_kwd: "timeline"` on their ES rows and graph. After this change, a recompile would write `compile_kwd: "knowledge_graph"` (or `"page_index"`), producing a _new_ graph row with a different row ID while leaving the old one intact. The old rows are harmless (the UI will only render tabs keyed to the current template IDs), but worth noting for QA — a manual graph-tab cleanup may be needed in edge cases.

2. **`_struct_entity_name` always falls back to JSON parsing** — in `cleanup_timeline_isolated_entities`, the queried fields are `content_with_weight`, `knowledge_graph_kwd`, `from_entity_kwd`, `to_entity_kwd`. `_struct_entity_name` first tries `row.get("name")` (which won not be present), then parses `content_with_weight` JSON. Adding `"name"` to the field list would skip the unnecessary fallback, though functionally it is fine either way.

3. **Non-standard autotype values** — `_struct_infer_type` now returns values like `"knowledge_graph"` and `"page_index"` that are not in `_STRUCT_TYPES = ("list", "set", "hypergraph")`. These flow into `compile_kwd` on entity/relation rows and graph rows. The code handles this correctly at every call site, but future code that checks `compile_kwd in _STRUCT_TYPES` could silently skip these rows. Nothing in the current codebase does this, so it is not a regression now.

### Verdict

LGTM. No regressions, no safety issues, no performance concerns beyond the negligible JSON-fallback noted above. Good to merge after a quick QA pass on the recompile scenario.

wangq8 (commented):
Thanks for this fix! The root cause is clear — `knowledge_graph` and `page_index` are distinct compilation types and should not be folded into `timeline`.

**✅ What looks good:**
- Both `_compilation_template_kind` (chunk_api) and `_struct_normalize_kind` (structure) are correctly updated.
- Timeline-specific cleanup (`cleanup_timeline_isolated_entities`) is properly scoped — it only runs for `template_kinds == "timeline"`.
- The Pipeline template ID resolution fallback is a nice addition for groups that don’t persist IDs in `parser_config`.
- `cleanup_timeline_isolated_entities` handles the edge case well (no relations → every entity is isolated → graph is rebuilt post-cleanup).

**⚠️ One thing worth fixing:**

`rag/advanced_rag/harness/tools/navigation.py` has its own `_normalize_kind` (line 55) that also normalizes `page_index`/`knowledge_graph` → `timeline`, with a docstring that says *"Mirror the API's kind normalization"*. Since the API no longer does this, the function and its comment should be updated for consistency. The comment at line 29–31 is now stale too.

Functionally it still works because the normalized value (`timeline`) matches `_CATALOG_KINDS`, but leaving the stale normalization in place is misleading for future maintainers. Consider removing the normalization there as well (and updating `_CATALOG_KINDS` / related comments accordingly).

PR #17155, opened 2026-07-21
Go: disable thinking for search ask summary to fix truncated AI summary

euvre:
### Summary

In Go mode, the AI summary on the search page appears truncated: the first part streams and displays, then disappears, and only a mid-sentence fragment remains.

Root cause: with a thinking-enabled chat model, the reasoning (which drafts the summary) streams first and is shown while the think block is unclosed. Once `</think>` arrives, the frontend collapses the whole reasoning block into a hidden `Thinking...` details element, so the drafted part vanishes. Since the reasoning has consumed the output budget, the provider then emits only a fragment as the visible answer.

Fix:
- `internal/service/ask_service.go`: pass `Thinking: false` in the ask summary generation config — summarization does not need reasoning, and this keeps the full output budget for the answer.
- `internal/entity/models/openai.go` (streaming and non-streaming): the qwen3 `enable_thinking` policy now honors an explicit `Thinking` setting instead of skipping the auto-disable whenever it is set (previously an explicit `false` would have re-enabled thinking for qwen3).

Verified: `internal/service` and `internal/entity/models` package tests pass; reproduced the collapse behavior end-to-end with a scripted thinking-model stream through the Ask pipeline and the frontend SSE reducer.

### DIFF
--- a/internal/entity/models/openai.go
+++ b/internal/entity/models/openai.go
@@ -124,10 +124,14 @@ func (o *OpenAIModel) ChatWithMessages(modelName string, messages []Message, api
 		}
 	}
 
-	// Qwen3 family: disable thinking by default (matches Python's
-	// _apply_model_family_policies in rag/llm/chat_model.py:119-121).
-	if strings.Contains(strings.ToLower(modelName), "qwen3") && (chatModelConfig == nil || chatModelConfig.Thinking == nil) {
-		reqBody["enable_thinking"] = false
+	// Qwen3 family: disable thinking unless explicitly enabled (matches
+	// Python's _apply_model_family_policies in rag/llm/chat_model.py:119-121).
+	if strings.Contains(strings.ToLower(modelName), "qwen3") {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for this fix! The root cause analysis is solid and the changes are clean.

**openai.go**: The old condition `chatModelConfig == nil || chatModelConfig.Thinking == nil` would skip the disable block whenever `Thinking` was explicitly set — including `false`. So an explicit `Thinking: false` from the ask service would leave thinking enabled on Qwen3, which is the opposite of what the caller intended. The new logic correctly reads the explicit value when set and defaults to disabled otherwise. Both the streaming and non-streaming paths are covered.

**ask_service.go**: Disabling thinking for search summaries makes total sense — summarization doesn't need multi-step reasoning, and the thinking output eats into the budget that should go to the visible answer. The comment is helpful for future maintainers. `ptrBool` follows the existing `ptrInt`/`ptrFloat64` pattern.

No concerns — LGTM.

PR #17156, opened 2026-07-21
Fix(go): check model availability

Lynn-Inf:
### Summary

Check model availability without match in json.


wangq8 (commented):
Thanks for the PR. The change looks reasonable.

**What changed:**
Removes the `GetModelByName` + `ModelTypeMap` validation in `checkModelAvailable`, relying solely on the DB-level model type check (`modelEntity.ModelType.Has(modelTypeBit)`) a few lines below.

**Review:**

✅ **Redundancy elimination** — the removed code validated model type via provider schema (`ModelTypeMap`), but the same validation already happens at the DB level with `modelEntity.ModelType.Has(modelTypeBit)` (line 757). The DB check is more authoritative since it reflects the tenant's actual configured model type.

✅ **No validation gap** — the critical checks remain intact:
- Provider/instance existence checks
- Model existence in DB via `GetModelByProviderIDAndInstanceIDAndModelName`
- Active status check (`modelEntity.Status != active`)
- Model type bit check (`modelEntity.ModelType.Has(modelTypeBit)`)

⚠️ **Minor: PR description** — the description is quite sparse. In the future, a bit more detail on why the schema-level check is redundant and why the DB check is sufficient would help reviewers.

LGTM 👍

PR #17157, opened 2026-07-21
fix: prevent collapse button from overlapping expanded content in next-search

euvre:
### Summary

On the next-search page, long chunk text is rendered inside an `ExpandableContent` component. When the content is expanded, the "View less" (收起) button was absolutely positioned at the bottom of the container (`absolute bottom-2`), which caused it to float on top of and obscure the last lines of the expanded text.

This PR fixes the issue by rendering the "View less" button in normal document flow (centered below the content) when the content is expanded, while keeping the collapsed-state "View more" button absolutely positioned over the gradient fade as before.

### DIFF
--- a/web/src/pages/next-search/expandable-content.tsx
+++ b/web/src/pages/next-search/expandable-content.tsx
@@ -84,24 +84,24 @@ export default function ExpandableContent({
         <div className="absolute bottom-0 left-0 right-0 h-20 bg-gradient-to-t from-bg-base to-transparent pointer-events-none" />
       )}
 
-      {isOverflowing && (
+      {isOverflowing && !isExpanded && (
         <Button
           size="sm"
           onClick={toggleExpand}
           className="absolute bottom-2 left-1/2 -translate-x-1/2"
         >
-          {isExpanded ? (
-            <>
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice fix for a real UX papercut. The absolute-positioned "View less" button floating over expanded content text was definitely annoying.

**What changed:**
- Split the single conditional button block into two: one for collapsed state (absolute positioned over gradient, as before) and one for expanded state (in normal document flow, centered below content)

**Review:**

✅ **Correct separation** — the collapsed-state button (`isOverflowing && !isExpanded`) keeps the absolute positioning over the gradient fade, which is the right UX for the collapsed view.

✅ **Expanded state** — the "View less" button now renders in normal flow with `flex justify-center pt-2`, centered below the expanded content. No more overlap.

✅ **Gradient** — only shown when `!isExpanded` (line 86, existing logic), so there's no gradient behind the in-flow button. Good.

✅ **Icons** — `ChevronDown` for "View more", `ChevronUp` for "View less" — correct.

LGTM 👍

dcc123456 (approved):

PR #17159, opened 2026-07-21
Feat: add open-in-new-tab button to search share embed dialog

euvre:
### Summary

The search share "Embed into site" dialog was missing the "Open in new tab" button that the chat share embed dialog already provides, making it impossible to preview the shared search page directly from the dialog.

This PR:
- Adds an "Open in new tab" button (with `ExternalLink` icon, matching the chat share dialog style) to the search share embed dialog (`web/src/pages/next-search/embed-app-modal.tsx`). It reuses the generated iframe URL, so the opened page respects the selected options (hide avatar, locale, etc.).
- Generalizes the shared `common.openInNewTab` locale string (previously chat-specific, e.g. "Chat in new tab") to a generic "Open in new tab" wording across all 13 locale files, so both the chat and search share dialogs share the same key consistently.

### DIFF
--- a/web/src/locales/ar.ts
+++ b/web/src/locales/ar.ts
@@ -55,7 +55,7 @@ export default {
       submit: 'إرسال',
       clear: 'مسح',
       embedIntoSite: 'تضمين في صفحة الويب',
-      openInNewTab: 'الدردشة في علامة تبويب جديدة',
+      openInNewTab: 'فتح في علامة تبويب جديدة',
       previousPage: 'سابق',
       nextPage: 'التالي',
       add: 'يضيف',
--- a/web/src/locales/de.ts
+++ b/web/src/locales/de.ts
@@ -49,7 +49,7 @@ export default {
       submit: 'Absenden',
       clear: 'Leeren',
       embedIntoSite: 'In Webseite einbetten',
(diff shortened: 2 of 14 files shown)


wangq8 (commented):
Thanks for this PR! Clean implementation overall.

**What this does:**
- Adds an "Open in new tab" button to the search share embed dialog, matching the pattern already present in the chat share embed dialog
- Generalizes the `common.openInNewTab` locale string from chat-specific wording to generic wording across all 13 locale files

**Review:**

✅ **Locale changes** — consistent across all 13 files, each changed from chat-specific wording to generic "Open in new tab" equivalent. The key is already used by the chat embed dialog at `web/src/components/embed-dialog/index.tsx:604`, so both dialogs now share the same generic string correctly.

✅ **Button implementation** — uses `variant="secondary"` with full width and `ExternalLink` icon, matches the chat dialog style. `generateIframeSrc()` is reused, so the opened page respects the selected options (hide avatar, locale, etc.).

✅ **No regressions** — the chat embed dialog will now show "Open in new tab" instead of "Chat in new tab", which is still clear and appropriate in that context.

LGTM 👍

dcc123456 (approved):

PR #17162, opened 2026-07-21
fix(web): hide folder upload tab for Gmail and Google Drive OAuth JSON upload

euvre:
### Summary

The Gmail and Google Drive data source creation dialogs only accept a single OAuth credential JSON file. However, the shared `FileUploader` component shows both a `Files` and a `Folder` tab by default, which misleads users into thinking they can upload an entire directory and increases the chance of misoperation.

This PR passes `showFolderTab={false}` to the `FileUploader` in `gmail-token-field.tsx` and `google-drive-token-field.tsx`, so only the file upload tab is shown for these OAuth JSON upload fields.


wangq8 (commented):
Simple, correct fix. Both Gmail and Google Drive OAuth dialogs only accept a single JSON credential file, so showing a folder upload tab is misleading and invites misoperation. Adding `showFolderTab={false}` is exactly the right fix. LGTM.

PR #17164, opened 2026-07-21
fix(web): use title case 'API Key' in UI labels across all locales

euvre:
### Summary

UI field labels previously used the all-caps form "API KEY", which reads like an acronym, a database column name, or an environment variable rather than a user-facing label. Common style guides (Google Material, Microsoft, Apple) recommend title case for UI labels.

This PR replaces all occurrences of "API KEY" with "API Key" in user-facing strings across all 12 locale files (`web/src/locales/`: bg, de, en, fr, id, it, ja, ko, ru, vi, zh, zh-traditional), covering field labels, validation messages, and tooltips (100 occurrences in total). No functional changes.


wangq8 (commented):
Purely cosmetic but nice consistency improvement. "API KEY" in all-caps reads like a database column name or env var, while "API Key" follows standard UI style guides (Google Material, Microsoft, Apple). The change is exhaustive across all 12 locale files and affects only user-facing label strings — no functional changes. LGTM.

PR #17165, opened 2026-07-21
fix(web): use outline-none instead of outline-0 on focusable controls

euvre:
### Summary

A focused checkbox in the filter popover (e.g. Agent list filters) renders with a broken, incomplete border: Chromium paints its native platform focus ring (blue/dark depending on OS theme) inside the 14px checkbox, covering its left border.

Root cause: the `Checkbox` and `Button` components use Tailwind's `outline-0`, which only sets `outline-width: 0px`. The browser UA stylesheet applies `outline-style: auto` on `:focus-visible`, and Chromium ignores `outline-width` when `outline-style` is `auto`, so the native focus ring is still painted.

Fix: switch to `outline-none`, which sets `outline-style: solid` with a transparent color and properly overrides the UA rule. Verified with headless Chrome at pixel level: the focused checkbox border is now complete and symmetric on all four sides.

### DIFF
--- a/web/src/components/ui/button.tsx
+++ b/web/src/components/ui/button.tsx
@@ -8,7 +8,7 @@ import { Link, LinkProps } from 'react-router';
 
 const buttonVariants = cva(
   cn(
-    'inline-flex items-center justify-center gap-2 whitespace-nowrap rounded-md text-sm transition-colors outline-0',
+    'inline-flex items-center justify-center gap-2 whitespace-nowrap rounded-md text-sm transition-colors outline-none',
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Solid CSS fix. `outline-0` only sets width to 0, but Chromium's UA stylesheet applies `outline-style: auto` on `:focus-visible` which ignores `outline-width`. Switching to `outline-none` forces `outline-style: solid` with transparent color, properly suppressing the native focus ring. Verified the fix is narrow and only affects these two components. LGTM.

dcc123456 (approved):

PR #17166, opened 2026-07-21
Fix: search input cannot expand to show long queries

euvre:
### Summary

On the search page (both the search home and the search results view), the query box is a single-line input. When the searched content is long, the text is truncated and there is no way to expand it to view the full query.

This PR replaces the single-line `Input` with an auto-growing `textarea` on both pages:

- The box grows with its content up to 160px and becomes scrollable beyond that, so long queries can be fully viewed.
- `Enter` submits the search, `Shift+Enter` inserts a newline.
- IME composition is respected (`isComposing`), so confirming an IME candidate no longer accidentally submits the search.
- Styling switches from `rounded-full` to `rounded-3xl` to keep the multi-line state visually consistent.

### DIFF
--- a/web/src/hooks/use-auto-resize-textarea.ts
+++ b/web/src/hooks/use-auto-resize-textarea.ts
@@ -1,0 +1,20 @@
+import { useLayoutEffect, type RefObject } from 'react';
+
+/**
+ * Auto-resize a textarea to fit its content, clamped to a maximum height.
+ *
+ * useLayoutEffect runs synchronously after DOM mutation but before the
+ * browser paints, so the height adjustment never produces a visible flicker.
+ */
+export function useAutoResizeTextarea(
+  ref: RefObject<HTMLTextAreaElement | null>,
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Good UX improvement. Replacing the single-line `Input` with an auto-growing `textarea` up to 160px makes long queries readable. A few things to double-check: 1) The transition from `onKeyUp` to `onKeyDown` with IME composition guard (`isComposing`) is correct, make sure it doesn't regress for non-IME users. 2) The `rounded-full` → `rounded-3xl` style change is subtle but looks fine. 3) On `search-view.tsx`, the old `Input` component may have had built-in disabled styling that the textarea doesn't inherit — confirm the `disabled:cursor-not-allowed disabled:opacity-50` classes are sufficient. Otherwise LGTM.

dcc123456 (commented):
The problems raised by the robot need to be solved.

PR #17167, opened 2026-07-21
fix(go-agent): route graph step errors through exception outputs

Hz-186:
## Summary

- Route ReAct graph step-limit errors through the Agent `_ERROR` output.
- Align the Go Agent default max rounds with Python by using five rounds.
- Preserve cancellation, timeout, and configuration errors as real failures.

## Testing

- `bash build.sh --test ./internal/agent/component/...`

image


yuzhichang:
LGTM

PR #17168, opened 2026-07-21
fix(chunker): glue delimiter to preceding segment instead of emitting it as a standalone chunk

euvre:
### Summary

In Go mode, the ingestion TokenChunker emitted the configured delimiter (e.g. `666`) as its own chunk: `splitKeepingDelim` kept each matched delimiter as a standalone element in the split stream, and every caller emitted stream elements as chunks directly.

This PR mirrors Python `token_chunker._split_text_by_pattern` (rag/flow/chunker/token_chunker.py): each delimiter is glued to the end of the preceding text segment, and a delimiter with no preceding text (a leading delimiter or one adjacent to another) is dropped together with the empty segment. The delimiter therefore never surfaces as a standalone chunk, matching the Python-mode behavior.

A regression test (`TestTokenChunker_DelimNeverStandaloneChunk`) covers the standalone-delimiter case.


wangq8 (commented):
Nice fix. Switching from `FindAllStringSubmatchIndex` to `FindAllStringIndex` and gluing delimiters to the end of the preceding segment matches Python's `_split_text_by_pattern` exactly. The regression test covers the exact bug case (delimiter as standalone chunk). LGTM.

wangq8:
@euvre fix the CI build.

PR #17169, opened 2026-07-21
fix(web): show builtin chunk methods in file pipeline dialog under Go backend

euvre:
### Summary

In Go backend mode, the per-file "switch or configure ingestion pipeline" dialog (dataset Files page) showed an empty builtin chunk-method select.

`useSelectParserList` parsed `GET /api/v1/pipelines?type=builtin` as a bare array of `{ parser_id, title, dsl }`, but the Go backend returns `{ canvas: [{ id, title, description, filename }], total }`. Because of the shape mismatch, the Go code path never used the pipeline catalog and fell back to the legacy tenant `parser_ids` list (`naive`, `qa`, ...), whose values never match the canonical builtin pipeline ids (e.g. `general`) stored on documents. As a result, the builtin select in the chunk method dialog rendered with no matching options.

Changes:
- Read `data.canvas` and map `id`/`title` when building the parser options in Go mode, aligning with the actual `/api/v1/pipelines` response contract.
- Remove the unused `useFetchParserListOnMount` helper (and its legacy-id `ParserListMap`) that would produce the same empty-list problem if ever used.


wangq8 (commented):
Good fix! The Go backend returns `{ canvas: [...], total }`, not a bare array, so `data.data` was always undefined in Go mode. Reading `data.canvas` and mapping `id`/`title` correctly resolves the empty builtin chunk-method select. Removing the dead `useFetchParserListOnMount` and `ParserListMap` is also a nice cleanup — that legacy code would have the same shape mismatch if ever used. LGTM.

dcc123456 (approved):

PR #17170, opened 2026-07-21
fix: fetch all knowledge bases via pagination in link-to-dataset dialog

euvre:
### Summary

Fix: the file management "link to dataset" dialog only displayed the 30 most recently created knowledge bases because `useFetchKnowledgeList` called the dataset list API without pagination parameters, hitting the Go backend default `page_size=30`. Knowledge bases beyond the first page were invisible and unsearchable in the selection dropdown.

This PR fixes the issue by looping through pages (`page_size=200`) until all knowledge bases are loaded, ensuring the complete list is available for both display and client-side search in the MultiSelect component.

The fix also benefits other consumers of `useFetchKnowledgeList`:
- Agent retrieval node knowledge base selector
- Dataset settings tag item
- Knowledge base item component
- Owner selection filter

### DIFF
--- a/web/src/components/knowledge-base-item.tsx
+++ b/web/src/components/knowledge-base-item.tsx
@@ -28,10 +28,12 @@ export function useDisableDifferenceEmbeddingDataset(name: string) {
   const datasetId = useWatch({ name, control: form.control });
   const [searchString, setSearchString] = useState('');
   const debouncedSearchString = useDebounce(searchString, { wait: 500 });
-  const { list: datasetListOrigin, loading } = useFetchKnowledgeList(
-    true,
-    debouncedSearchString,
-  );
+  const {
+    list: datasetListOrigin,
+    loading,
+    handleScroll,
+    hasMore,
+  } = useFetchKnowledgeList(true, debouncedSearchString);
   const datasetCacheRef = useRef(new Map<string, IDataset>());
 
   const datasetList = useMemo(() => {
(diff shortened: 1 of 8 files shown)


wangq8 (commented):
LGTM. Clean fix that brings the Go frontend KB list in line with the Python behavior.

**What's good:**
- The pagination loop correctly fetches all KBs with page_size=200, solving the 30-KB visibility cap
-  is the right condition to detect when there are more pages
- The  filter is applied to the aggregated  list, preserving existing behavior
- This also fixes the Agent retrieval node selector and other consumers of this hook

**One minor thought:** If the API were ever to return exactly  results but have no more pages (a full page == last page edge case), this would make one extra empty request. Not a correctness issue — just an extra round-trip. You could consider checking  instead, but the current approach is fine and defensively safe. No change needed.

wangq8 (commented):
LGTM. Clean fix that brings the KB list loading in line with the Python behavior.

What's good:
- The pagination loop correctly fetches all KBs with page_size=200, solving the 30-KB visibility cap
- `hasMore = pageData.length >= pageSize` is the right condition to detect when there are more pages
- The `shouldFilterListWithoutDocument` filter is applied to the aggregated `all` list, preserving existing behavior
- This also fixes the Agent retrieval node selector and other consumers of this hook

One minor thought: If the API were ever to return exactly `pageSize` results but have no more pages (a full page == last page edge case), this would make one extra empty request. Not a correctness issue, just an extra round-trip. You could consider checking `pageData.length < pageSize` instead, but the current approach is fine and defensively safe. No change needed.

dcc123456 (approved):

PR #17172, opened 2026-07-21
fix(file2document): batch link should add to existing KBs, not replace them

euvre:
### Summary

The Go backend's batch link-to-datasets endpoint (`POST /api/v1/files/link-to-datasets`) always performed a **replace** operation — it deleted all existing file-to-dataset links before creating new ones, ignoring the `mode` query parameter sent by the frontend.

The frontend already distinguishes between the two modes (see `web/src/pages/files/hooks.ts`):
- **Batch linking** (multiple files): sends `mode=add`
- **Single-file linking**: sends `mode=replace`

The Python backend (`api/apps/restful_apis/file2document_api.py`) correctly handles both `add` and `replace` modes, but the Go port (`internal/service/document/file2document.go`) hard-coded the replace behavior, so batch linking silently wiped out previously linked knowledge bases.

**Fix:**

1. **`internal/handler/file.go`** — Read the `mode` query parameter (defaulting to `replace` for backward compatibility) and validate it. Pass it through to the service layer.

2. **`internal/service/document/file2document.go`** — `convertFiles` now branches on `mode`:
   - `add`: collects the KB IDs already linked to the file and skips them, **preserving all existing links** and only creating new links for KBs not yet linked.
   - `replace` (default): removes all existing links then creates new ones (unchanged behavior).

   Added a `documentDAO` field to look up the `kb_id` of existing linked documents.

This aligns the Go implementation with the Python implementation.

### DIFF
--- a/internal/handler/file.go
+++ b/internal/handler/file.go
@@ -559,7 +559,13 @@ func (h *FileHandler) LinkToDatasets(c *gin.Context) {
 		return
 	}
 
-	if err := h.file2DocumentService.LinkToDatasets(user.ID, &req); err != nil {
+	mode := strings.ToLower(c.DefaultQuery("mode", "replace"))
+	if mode != "add" && mode != "replace" {
+		common.ResponseWithCodeData(c, common.CodeArgumentError, nil, "mode must be 'add' or 'replace'")
+		return
+	}
+
+	if err := h.file2DocumentService.LinkToDatasets(user.ID, &req, mode); err != nil {
 		common.ResponseWithCodeData(c, linkToDatasetsErrorCode(err), nil, err.Error())
 		return
 	}
--- a/internal/service/document/file2document.go
+++ b/internal/service/document/file2document.go
@@ -52,6 +52,7 @@ type File2DocumentService struct {
 	kbDAO            *dao.KnowledgebaseDAO
 	userTenantDAO    *dao.UserTenantDAO
 	documentSvc      *DocumentService
+	documentDAO      *dao.DocumentDAO
 }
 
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
LGTM. This properly aligns the Go file2document batch-link behavior with the Python implementation.

**What's good:**
- Handler validates  query param (add/replace) with a clear error for invalid values — correct approach
- The  path correctly collects  from current file2document mappings and skips already-linked KBs, preserving existing links
- The  path (default) maintains original behavior: removes existing docs/mappings before creating new ones
-  field added to  struct to look up document KB IDs — wired up in  constructor, no missed injection sites
- The goroutine call passes  through correctly

**Edge case check:** If  for a document ID returns an error or nil, the code skips that KB — silent skip is fine here since it mirrors Python's tolerance for orphaned mappings. No change needed.

wangq8 (commented):
LGTM. This properly aligns the Go file2document batch-link behavior with the Python implementation.

What's good:
- Handler validates `mode` query param (add/replace) with a clear error for invalid values — correct approach
- The `add` path correctly collects `existingKBIDs` from current file2document mappings and skips already-linked KBs, preserving existing links
- The `replace` path (default) maintains original behavior: removes existing docs/mappings before creating new ones
- `documentDAO` field added to `File2DocumentService` struct to look up document KB IDs — wired up in `NewFile2DocumentService()` constructor, no missed injection sites
- The goroutine call passes `mode` through correctly

Edge case check: If `GetByID` for a document ID returns an error or nil, the code skips that KB — silent skip is fine here since it mirrors Python's tolerance for orphaned mappings. No change needed.

PR #17175, opened 2026-07-21
Go: only unlink file-manager files when deleting documents from a knowledge base

euvre:
### Summary

Deleting a document from a knowledge base in Go mode incorrectly hard-deleted the underlying file record and storage blob whenever no other document referenced it. Files linked from the file manager to a KB thus disappeared from the file manager instead of just being unlinked.

This PR aligns the Go behavior with the existing Python implementation (`FileService.delete_docs`): only delete the file record and storage blob when the file is a knowledgebase-owned upload (`source_type == "knowledgebase"`). For files linked from file management (`source_type == ""` or `"local"`), only the `file2document` mapping is removed, keeping the file intact in the file manager.

### DIFF
--- a/internal/service/document/document_crud.go
+++ b/internal/service/document/document_crud.go
@@ -376,7 +376,9 @@ func (s *DocumentService) rollbackAddFileFromKBError(doc *entity.Document, kbID 
 
 // cleanupFileReferences deletes file2document mappings for docID, and for each
 // referenced file, only hard-deletes the file record and its storage blob when
-// no other document still references the same file_id.
+// the file is a knowledgebase-owned upload (source_type == knowledgebase) and
+// no other document still references the same file_id. Files linked from file
+// management are only unlinked — the file record and blob stay intact.
 func (s *DocumentService) cleanupFileReferences(docID string) {
 	mappings, mapErr := s.file2DocumentDAO.GetByDocumentID(docID)
 	if mapErr != nil {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. This correctly prevents file-manager files from being hard-deleted when their linked documents are removed from a KB.

**What's good:**
- The  guard is the right check — only knowledgebase-uploaded files get hard-deleted, matching Python's 
- File-manager files ( or ) are only unlinked via the file2document mapping removal, keeping the file record and blob intact
- The comment update accurately describes the new behavior

**Edge case check:** What about ? In this path, S3 files also get preserved (which is correct — S3 files should only be unlinked, not hard-deleted, matching Python behavior). No issues.

wangq8 (commented):
LGTM. This correctly prevents file-manager files from being hard-deleted when their linked documents are removed from a KB.

What's good:
- The `entity.FileSource(file.SourceType) != entity.FileSourceKnowledgebase` guard is the right check — only knowledgebase-uploaded files get hard-deleted, matching Python's `FileService.delete_docs`
- File-manager files (source_type == "" or "local") are only unlinked via the file2document mapping removal, keeping the file record and blob intact
- The comment update accurately describes the new behavior

Edge case check: S3 files (source_type == "s3") also get preserved in this path (only unlinked), which is correct — matching Python behavior. No issues.

PR #17176, opened 2026-07-21
fix(go): resolve tenant model IDs in memory create/update without hard failure

euvre:
### Summary

In Go mode, the memory configuration page could not set LLM and embedding model parameters because the Go `MemoryService` diverged from the Python implementation in two ways:

1. **`CreateMemory` hard-failed on model resolution**: When `ResolveModelID` returned an error (e.g. for Builtin models that have no `tenant_model` row), the entire memory creation was aborted. Python's `ensure_tenant_model_ids_for_params` catches `LookupError` and silently skips, allowing the operation to proceed.

2. **`UpdateMemory` did not resolve tenant model IDs**: When `llm_id` or `embd_id` was updated, the corresponding `tenant_llm_id` / `tenant_embd_id` was never re-resolved, leaving them stale or null. Python's `update_memory` API handler calls `ensure_tenant_model_ids_for_params` before processing the update.

Both are now fixed to match Python behavior:

- `CreateMemory`: resolution failure is non-fatal — if `ResolveModelID` fails or returns empty, the `tenant_*_id` field is left nil and the memory is still created.
- `UpdateMemory`: when `llm_id` / `embd_id` is provided and the corresponding `tenant_*_id` is not explicitly set, the backend resolves it via `ResolveModelID`. Resolution failure is non-fatal.

### Testing

```
bash build.sh --test ./internal/service/...
ok  ragflow/internal/service  1.996s
ok  ragflow/internal/service/chunk  0.202s
ok  ragflow/internal/service/dataset  1.261s
ok  ragflow/internal/service/document  1.012s
ok  ragflow/internal/service/file  0.262s
ok  ragflow/internal/service/graph  0.115s
ok  ragflow/internal/service/nlp  0.024s
```

### DIFF
--- a/internal/service/memory.go
+++ b/internal/service/memory.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"ragflow/internal/common"
 	"ragflow/internal/entity"
 	models "ragflow/internal/entity/models"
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
LGTM on the memory service changes. The model resolution fix is correct and necessary.

**What's good:**
- : Model resolution failure is now non-fatal — correctly handles Builtin models that have no  row, matching Python's 
- :  /  are now re-resolved when  /  is updated, keeping the tenant model ID fields consistent after updates — this was a real bug
- Resolution guard  is correct — skips empty/invalid resolutions silently

**One note:** The PR also changes  →  in  and . These appear unrelated to the memory fix. If intentional (e.g., tailwind v4 migration cleanup), it's fine — but next time consider separating unrelated UI polish into its own PR to keep the diff focused.

wangq8 (commented):
LGTM on the memory service changes. The model resolution fix is correct and necessary.

What's good:
- `CreateMemory`: Model resolution failure is now non-fatal — correctly handles Builtin models that have no `tenant_model` row, matching Python's `ensure_tenant_model_ids_for_params`
- `UpdateMemory`: `tenant_llm_id` / `tenant_embd_id` are now re-resolved when `llm_id` / `embd_id` is updated, keeping the tenant model ID fields consistent after updates — this was a real bug
- Resolution guard (`err == nil && resolved != ""`) is correct — skips empty/invalid resolutions silently

One note: The PR also changes `outline-0` to `outline-none` in `web/src/components/ui/button.tsx` and `web/src/components/ui/checkbox.tsx`. These appear unrelated to the memory fix. If intentional (e.g., tailwind v4 migration cleanup), it's fine — but next time consider separating unrelated UI polish into its own PR to keep the diff focused.

Lynn-Inf (commented):
For branches where `err != nil`, it's better to add some logging to avoid masking the actual error. Other parts LGTM.

dcc123456 (approved):

PR #17177, opened 2026-07-21
fix(web): preserve MCP authorization token across edits

euvre:
### Summary

When reopening the MCP server edit dialog, the \`authorization_token\` field was left empty. Saving without re-entering the token therefore overwrote the stored secret with an empty string, clearing it on the backend.

This change shows a mask in the token field when a token already exists, and restores the previously stored token on save/test when the field is unchanged. As a result the secret survives round-trips through the edit dialog without being re-typed.

- \`web/src/pages/user-setting/mcp/edit-mcp-dialog.tsx\`: populate the token field with a mask when an existing token is present, and substitute the stored token back on save/test when the field is unchanged.

### DIFF
--- a/web/src/pages/user-setting/mcp/edit-mcp-dialog.tsx
+++ b/web/src/pages/user-setting/mcp/edit-mcp-dialog.tsx
@@ -39,7 +39,12 @@ const DefaultValues = {
   name: '',
   server_type: ServerType.SSE,
   url: '',
+  authorization_token: '',
 };
+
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Clean UX fix — the token mask approach is the standard pattern for preserving secrets through edit dialogs. The substitution back to the stored token when the field is unchanged (TokenMask comparison) is correct, and `form.reset` properly initializes the mask on dialog open.

dcc123456 (approved):

PR #17179, opened 2026-07-21
fix: MCP import now processes all files instead of only the first

euvre:
### Summary

When importing MCP servers via the upload dialog (using multi-file selection or folder upload), only the first JSON file was processed regardless of how many files were selected. This was caused by `useImportMcp` only reading `fileList[0]` in the `onImportOk` handler.

This PR fixes the issue by:
- Filtering the file list to only JSON files (by MIME type and `.json` extension) to handle folder uploads which may contain non-JSON files
- Iterating over all JSON files, validating each against the `McpConfigSchema`
- Merging all `mcpServers` entries from all files into a single combined map via `Object.assign`
- Sending a single import request with all merged servers


wangq8 (commented):
LGTM. Good improvement — the multi-file iteration and merge-via-Object.assign approach correctly handles folder uploads with multiple JSON files. The JSON filter (MIME type + .json extension fallback) handles browser inconsistencies in file.type. Error handling is improved: each file is validated independently against McpConfigSchema, and all servers are merged before a single import request.

dcc123456 (approved):

PR #17180, opened 2026-07-21
fix: Go backend builtin chunk method list not showing

euvre:
### Summary

In Go mode, the Ingestion Pipeline configuration dialog (ChunkMethodDialog) would not display the built-in chunk method list when the parse type was set to "Built-in". The dropdown appeared empty with placeholder text "请选择分块方法".

**Root cause:** `useSelectParserList` in `web/src/hooks/use-user-setting-request.tsx` had two issues in the Go backend code path:

1. **Wrong response path** — accessed `pipelineListData?.data` instead of `pipelineListData?.data?.canvas`. The Go `ListPipelines` handler wraps the pipeline array in `{canvas: [...], total: N}`, so `data` is an object, not an array. The `.length` check on the object always returned `undefined > 0 === false`, silently falling through to the Python fallback path which yielded incorrect results in Go mode.

2. **Wrong JSON field name** — used `item.parser_id` instead of `item.id`. The Go `BuiltinPipelineMeta` struct uses `json:"id"`, not `parser_id`.

**Fix:** Changed `pipelineListData?.data` to `pipelineListData?.data?.canvas` and `item.parser_id` to `item.id`, aligning with the already-working `useFetchBuiltinPipelines` hook.

**Verification:** Pre-commit hooks (web-prettier, web-eslint, TypeScript type-check) all pass.


wangq8 (commented):
LGTM. Two-line fix that addresses both issues: `data` → `data.canvas` to match the `{canvas: [...], total: N}` response shape, and `parser_id` → `id` to match the `BuiltinPipelineMeta` JSON tag. Consistent with the already-working `useFetchBuiltinPipelines` hook.

dcc123456 (approved):

PR #17181, opened 2026-07-21
fix: avoid duplicate document names when linking files to datasets

euvre:
### Summary

When linking files to a knowledge base that already contains a document with the same name, the Go path silently created a second document with the identical name, producing duplicates (e.g., two `laws.pdf` entries in the same dataset). The Python path avoids this by calling `duplicate_name` to generate non-colliding names like `laws(1).pdf`.

This PR mirrors that behavior in the Go `convertFiles` path:

- Added `UniqueDocumentName` method on `DocumentService` that queries existing document names in the KB and returns a collision-free name.
- Called `UniqueDocumentName` in `convertFiles` before creating each document, recalculating `suffix` from the deduped name.


wangq8 (commented):
LGTM. Mirrors the Python `duplicate_name` behavior correctly. Good that `UniqueDocumentName` returns an error when the name lookup fails (rather than silently proceeding), and `suffix` is recalculated from the deduped name so the file extension remains accurate.

PR #17182, opened 2026-07-21
fix: count distinct chunk IDs so doc.chunk_num matches actual indexed chunks

euvre:
### Summary:

PipelineResult.ChunkCount was using `len(chunks)` from the raw pipeline output, but chunks with identical text share the same ChunkID (xxhash of text+docID). The search engine upserts by `_id`, so chunks with duplicate text collapse to a single document in the index. This caused `doc.chunk_num` in the database to diverge from the actual number of indexed chunks.

### Changes:

- Added `countDistinctChunkIDs()` in `internal/ingestion/task/pipeline_executor.go` to count unique chunk IDs after `ProcessChunksForPipeline` assigns deterministic IDs via `ChunkID`.
- `processOutput` now uses the deduplicated count for `PipelineResult.ChunkCount`, so both `doc.chunk_num` and `kb.chunk_num` accurately reflect the effective number of chunks in the search index.
- Added unit test `TestCountDistinctChunkIDs` covering empty, all-unique, with-duplicates, and missing-id cases.

### DIFF
--- a/internal/ingestion/task/pipeline_executor.go
+++ b/internal/ingestion/task/pipeline_executor.go
@@ -224,14 +224,34 @@ func (s *PipelineExecutor) processOutput(ctx context.Context, pipelineOutput map
 		return nil, err
 	}
 
+	chunkCount := countDistinctChunkIDs(chunks)
+
 	return &PipelineResult{
 		DocID:            s.taskCtx.Doc.ID,
 		KbID:             s.taskCtx.Doc.KbID,
 		Metadata:         metadata,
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. The root cause analysis is correct — chunks with identical text share the same ChunkID (xxhash of text+docID) and the search engine upserts by `_id`, so `len(chunks)` overcounts. The `countDistinctChunkIDs` function correctly mirrors the index-side deduplication. Test covers empty, all-unique, with-duplicates, and missing-id cases.

PR #17183, opened 2026-07-21
Fix: verify model api

Lynn-Inf:
### Summary

Verify model with `api_key` and persist result to DB when `instance_id` is provided.

### DIFF
--- a/internal/handler/providers.go
+++ b/internal/handler/providers.go
@@ -301,7 +301,7 @@ func (h *ProviderHandler) CheckConnection(c *gin.Context) {
 	}
 
 	userID := c.GetString("user_id")
-	errCode, err := h.modelProviderService.CheckConnection(providerName, req.APIKey, req.Region, req.BaseURL, userID)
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Overall LGTM — this mirrors the Python `verify_api_key` per-model verification logic well. A few observations:

1. `updateModelVerifyResults` uses `userTenants[0]` to resolve the tenant — this matches existing patterns in the codebase so it is consistent, though multi-tenant users should be noted as a future edge case.
2. The `verifyProviderModel` signature takes `[]string` (model names) while `CheckConnectionRequest.ModelInfo` is `[]CheckConnectionModelInfo` — the `ListModelNames()` bridge function handles this correctly, extracting just the names.
3. `generateTestWAV` uses `binary.LittleEndian.AppendUint32/AppendUint16` (Go 1.19+). Please confirm the project min Go version supports this.
4. Temp file cleanup (`defer os.Remove`) for ASR verification is correct.
5. Error extraction via `extractAPIErrorMessage` is a nice touch for cleaner user-facing messages.

The verification approach (lightweight test requests per model type — chat, embedding, rerank, TTS, ASR, OCR) is appropriate and non-destructive.

PR #17184, opened 2026-07-21
fix: restore download button when file type is selected in agent message

euvre:
### Summary

When an agent Message component has file type checkboxes enabled (e.g., Markdown), the backend emits attachment and download information via `message_end` events. However, the SSE relay in agent chat sessions filters `workflow_finished` outputs (keeping only `usage`), and the frontend previously read attachments solely from `workflow_finished`, causing the download button to disappear even when file types were correctly configured.

### Changes

- **Frontend (`use-send-message.ts`)**: Extend `IMessageEndData` interface to support `attachment` and `downloads` fields
- **Frontend (`use-send-agent-message.ts`)**: Fall back to reading `attachment` / `downloads` from the last `message_end` event when `workflow_finished.outputs` is unavailable
- **Backend (`canvas.py`)**: Include `downloads` in `_build_message_end` so they survive the SSE relay that strips outputs from `workflow_finished`
- **Frontend (`next-message-item/index.tsx`)**: Remove stale commented-out attachment download code


wangq8 (commented):
LGTM. Multi-layer fix is well-coordinated: backend adds `downloads` to `message_end`, frontend falls back to reading attachment/downloads from the last `message_end` event when `workflow_finished` outputs are unavailable, and stale commented-out code is cleaned up. The fallback chain in `findMessageFromList` is correctly ordered.

dcc123456 (approved):

PR #17185, opened 2026-07-21
fix(agent): restrict permission field edits to canvas owner only

euvre:
### Summary

Currently, team members who have access to a shared agent canvas
(via the team permission) can change the permission field from
"team" to "me" (private). This allows a non-owner to lock the
agent away from both other team members and potentially the
original owner themselves.

**Fix:** In `UpdateAgent`, the permission field is silently dropped
from the patch when the caller is not the canvas owner. Title,
avatar, and description updates from team members still succeed.
The original owner can change permission together with title/avatar
in the same request.

**Files changed:**
- `internal/service/agent.go` — owner-only guard on permission in `UpdateAgent`
- `internal/service/agent_test.go` — new `TestUpdateAgentPermissionOwnerOnly` test

### DIFF
--- a/internal/service/agent.go
+++ b/internal/service/agent.go
@@ -581,10 +581,21 @@ func (s *AgentService) GetAgent(ctx context.Context, userID, canvasID string) (*
 
 // UpdateAgent applies a draft patch to user_canvas. Settings updates may omit
 // dsl; in that case the existing draft DSL must be preserved.
+//
+// Permission is an owner-only setting: team members who have access to the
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Good security fix — silently dropping the permission field when the caller is not the canvas owner is the right approach (rather than returning an error, which would break the UX for legitimate title/avatar updates by team members). The test is thorough: verifies member can still update title/avatar while permission is preserved, and owner can change permission + other fields in one request.

PR #17186, opened 2026-07-21
fix: add create_date field to APIKeyResponse for token list display

euvre:
### Summary

The APIKeyResponse struct (Go backend) was missing the create_date and update_date fields. The frontend chat-api-key-modal relies on create_date to render the 'Created' column in the API key table, so the column was always empty on the Go runtime.

Root cause: APIKeyResponse only serialized create_time (Unix timestamp), but the frontend calls formatDate(tokenItem.create_date), not formatDate(tokenItem.create_time).

Fix:
- Added CreateDate and UpdateDate fields to APIKeyResponse struct
- Populated them from the entity in both ListAPIKeys and CreateAPIKey

The underlying entity (BaseModel) already stores CreateDate and UpdateDate via GORM hooks; the response layer just wasn't forwarding them.


wangq8 (commented):
LGTM. Straightforward fix — the entity already has CreateDate/UpdateDate populated by GORM hooks, this just forwards them through the API response layer so the frontend `formatDate(tokenItem.create_date)` can render the column. Both `ListAPIKeys` and `CreateAPIKey` are updated consistently.

PR #17187, opened 2026-07-21
fix(go): resolve custom tenant model in LLM agent canvas component

euvre:
### Summary

The LLM component (`internal/agent/component/llm.go`) did not resolve tenant-scoped model references (`tenant_model.id` or `model@instance@provider`) before dispatching to the model driver. The `Agent` and `Categorize` components already called `resolveChatModelRef` to map canvas model IDs to real provider names, driver config, and credentials — but the `LLM` component passed the raw `model_id` verbatim. This caused custom-added models to fail with HTTP 400 when selected in the agent canvas.

This PR adds `resolveChatModelRef` at the start of `LLMComponent.Invoke`, mirroring the existing resolution logic in `AgentComponent.Invoke` and `CategorizeComponent.Invoke`.

Additionally, `TestLLM_ResolvesTenantModelID` is added to guard tenant model ID resolution through the full provider/instance/model chain.


wangq8 (commented):
LGTM. Clean fix that mirrors the existing `resolveChatModelRef` pattern from `AgentComponent` and `CategorizeComponent`. The test covers the full provider/instance/model resolution chain including Driver, ModelName, APIKey, and BaseURL assertions. No regressions — the resolve call is placed before the existing empty-model check, which is the correct ordering.

Lynn-Inf (approved):
LGTM

PR #17192, opened 2026-07-21
refa: optimize wiki compilation concurrency

buua436:

Improve Wiki compilation throughput while keeping LLM concurrency and memory usage bounded.

- Add shared LLM pool control with 20 active calls and 25 total pending/executing requests.
- Allow MAP batches to feed the pool concurrently without unbounded queue growth.
- Run Reduce LLM disambiguation batches concurrently.
- Replace the full embedding similarity matrix with blockwise cosine computation.
- Run REFINE pages concurrently through the shared LLM pool.

### Type of change

### DIFF
--- a/rag/advanced_rag/knowlege_compile/_common.py
+++ b/rag/advanced_rag/knowlege_compile/_common.py
@@ -633,7 +633,7 @@ async def _embedding_dedup(
     merge_threshold: float = 0.90,
     ambiguous_low: float = 0.75,
 ) -> tuple[dict[int, int], list[tuple[int, int]], Optional[list]]:
-    """Vectorised pairwise cosine; same-type-only when ``type_key`` given.
+    """Blockwise pairwise cosine; same-type-only when ``type_key`` given.
 
     Returns ``(merged_into, ambiguous_pairs, vectors)``. ``merged_into``
     is a union-find map ``index → parent_index``. ``ambiguous_pairs`` is the
@@ -655,11 +655,17 @@ async def _embedding_dedup(
         return {}, [], None
 
     try:
-        from sklearn.metrics.pairwise import cosine_similarity
         import numpy as np
 
         matrix = np.asarray([list(v) for v in vectors], dtype=float)
-        sims = cosine_similarity(matrix)
+        if matrix.ndim != 2 or matrix.shape[0] != n:
+            raise ValueError(f"invalid embedding matrix shape: {matrix.shape}")
+
+        # Normalize once, then calculate only bounded blocks of the upper
+        # triangle. The old implementation materialized an N x N matrix,
+        # which becomes a significant memory spike for large KBs.
+        norms = np.linalg.norm(matrix, axis=1, keepdims=True)
+        matrix = np.divide(matrix, norms, out=np.zeros_like(matrix), where=norms > 0)
     except Exception:
         logging.exception("bulk_dedup: pairwise cosine failed; skipping")
         return {}, [], vectors
@@ -673,16 +679,28 @@ async def _embedding_dedup(
 
     auto_pairs: list[tuple[int, int]] = []
     ambiguous_pairs: list[tuple[int, int]] = []
+    block_size = 1024
+    groups: dict[Any, list[int]] = {}
+    for index, item in enumerate(canonical):
+        groups.setdefault(item.get(type_key) if type_key else None, []).append(index)
 
-    for i in range(n):
-        for j in range(i + 1, n):
-            if type_key and canonical[i].get(type_key) != canonical[j].get(type_key):
-                continue
-            s = float(sims[i, j])
-            if s >= merge_threshold:
-                auto_pairs.append((i, j))
-            elif s >= ambiguous_low:
-                ambiguous_pairs.append((i, j))
+    for group_indices in groups.values():
+        for left_start in range(0, len(group_indices), block_size):
+            left_indices = group_indices[left_start : left_start + block_size]
+            left_vectors = matrix[left_indices]
+            for right_start in range(left_start, len(group_indices), block_size):
+                right_indices = group_indices[right_start : right_start + block_size]
+                sims = left_vectors @ matrix[right_indices].T
+                if right_start == left_start:
+                    rows, cols = np.triu_indices(len(left_indices), k=1)
+                else:
+                    rows, cols = np.nonzero(sims >= ambiguous_low)
+                for row, col in zip(rows.tolist(), cols.tolist()):
+                    score = float(sims[row, col])
+                    if score >= merge_threshold:
+                        auto_pairs.append((left_indices[row], right_indices[col]))
+                    elif score >= ambiguous_low:
+                        ambiguous_pairs.append((left_indices[row], right_indices[col]))
 
     for i, j in auto_pairs:
         ri, rj = _root(i), _root(j)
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
## Review: refa: optimize wiki compilation concurrency (#17192)

Thanks for this PR — the producer-consumer pattern and shared LLM pool are a solid architectural improvement. I left detailed notes below. The main callout is the concurrency explosion risk from `max_workers=0` and the lost pre-filtering in disambiguation.

### 1. (Important) `max_workers=0` can create unbounded concurrent tasks

In `dataset_wiki_generator.py` the MAP worker now calls `wiki_map_from_chunks(..., max_workers=0)`. When `max_workers` is 0, `_run_chunked_pipeline` creates **all** sub-batches as concurrent asyncio tasks with no semaphore:

```python
# _common.py, run_chunked_pipeline
semaphore = asyncio.Semaphore(max_workers) if max_workers and max_workers > 0 else None
```

If `split_chunks` produces e.g. 20 sub-batches per batch, and 20 workers are active, that is up to **400 asyncio tasks** launched simultaneously. The shared LLMCallPool (`max_pending=25`) limits the actual LLM calls, but having hundreds of tasks pending is wasteful and risks excessive memory pressure from all the suspended coroutine frames.

**Suggestion:** Either set a reasonable non-zero `max_workers` (e.g. `6` like the default) or add an explicit bound inside `_run_chunked_pipeline` for the `max_workers=0` path. At minimum, add a comment documenting the reasoning so future readers are not surprised.

### 2. Disambiguation pre-filtering is lost

Old code in `_resolve_ambiguous_pairs`:

```python
batch = [(i, j) for i, j in batch if _root(i) != _root(j)]
```

This filtered pairs that were already redundant before firing the LLM call. The new code runs all batches concurrently via `asyncio.gather` and only filters at apply time. This means the LLM may be asked to judge pairs that will already be merged by an earlier-applied batch result.

This is a functional tradeoff (correctness is preserved by the final `_root(i) != _root(j)` guard in the apply loop), but it wastes LLM calls and tokens. Consider adding the pre-filtering back after gathering results, or at least batching the filtering with the concurrent dispatch.

### 3. Blockwise embedding dedup looks correct but worth verifying edge case

The `_embedding_dedup` change from `sklearn.metrics.pairwise.cosine_similarity` to blockwise matrix multiply is a good memory optimization. One thing to double-check in testing:

The DIY normalization:
```python
norms = np.linalg.norm(matrix, axis=1, keepdims=True)
matrix = np.divide(matrix, norms, out=np.zeros_like(matrix), where=norms > 0)
```

The `out=np.zeros_like(matrix)` means zero-norm vectors become all-zero rows. Their cosine similarity with any other vector is then 0 (well below both thresholds). This is **consistent** with the old sklearn behavior (which would return NaN/0 for zero-vectors), but worth an explicit test case to confirm.

### 4. Minor: WIKI_MAP_QUEUE_SIZE = 5 may be too conservative

With 20 workers and potentially large documents, a queue depth of 5 means producers are throttled aggressively. If a document has many batches, the producer will put→block→put→block repeatedly, reducing throughput. Consider raising this to `2 * WIKI_MAP_LLM_POOL_SIZE` (40) or similar, since the pool is the true concurrency limiter and the queue is just a buffer.

### 5. LLMCallPool pending gate is correct

The `max_pending` bound in `LLMCallPool.call()` gates on `pending_count >= max_pending` using async `Condition.wait()`. This correctly back-pressures without deadlocking since callers are pure async (not holding locks). ✅

### 6. Docs are solid

The docstring updates in `_embedding_dedup`, `_resolve_ambiguous_pairs`, and `bulk_dedup_items` clearly explain the changed semantics. The comments in `dataset_wiki_generator.py` about the producer-consumer design are helpful. ✅

### Summary

The overall architecture is sound: shared pool bounding actual LLM calls, async queue for producer-consumer MAP, concurrent REFINE. My main concerns are (1) the unbounded task explosion from `max_workers=0` and (2) the lost disambiguation pre-filtering. Neither is a correctness bug, but both could hurt production performance/latency.

Consider addressing #1 and #2 before merging. The rest looks good.

wangq8 (commented):
## PR Review: refa: optimize wiki compilation concurrency

Thanks for this PR. The overall direction — shared LLM pool, blockwise embedding dedup, and concurrent disambiguation / REFINE — is solid and should meaningfully improve Wiki compilation throughput. A few observations:

---

### ✅ What looks good

1. **`run_chunked_pipeline`** — Producer-consumer pattern via `asyncio.Queue` is cleaner than the old `Semaphore` + `gather` approach. It avoids creating all tasks upfront and the sentinel-based worker termination is correct. Exception propagation is also handled properly (worker crash → `gather` raises → cancel + re-raise).

2. **Blockwise cosine similarity** — Replacing `sklearn.cosine_similarity` (N×N materialization) with blockwise `matrix @ block.T` is a strong improvement for large KBs. The grouping by `type_key` and chunk-by-chunk `triu_indices` / `np.nonzero` logic correctly covers all pairs without duplication.

3. **`LLMCallPool` backpressure** — Adding `max_pending` with the `pending_count` guard keeps queue growth bounded. Since the existing code uses `notify_all()` everywhere, no deadlock risk; the `while pending_count >= max_pending` re-check after wakeup is correct.

4. **Concurrent `_resolve_ambiguous_pairs`** — Deferring union-find mutations to a single serial pass after `asyncio.gather` avoids race conditions on `merged_into`. The docstring clearly explains the intent.

5. **Template resolution upfront** — Moving template resolution before the MAP worker pool avoids subtle ordering dependencies on which document happens to finish first.

---

### ⚠️ Items worth addressing

#### 1. `WIKI_MAP_QUEUE_SIZE = 5` with 20 workers may starve workers

With 20 map workers but a queue capacity of only 5 batches, workers will frequently block on `map_queue.get()`. Since each batch triggers expensive LLM calls, this may be acceptable in practice, but the ratio feels off. Consider at least matching `WIKI_MAP_LLM_POOL_SIZE` (20), or document why 5 was chosen (e.g. ES fetch latency is the bottleneck, not the queue).

#### 2. `_resolve_ambiguous_pairs` sends all pairs to LLM without pre-filtering

The old code filtered `_root(i) != _root(j)` *before* each batch's LLM call, avoiding wasted LLM queries for already-merged transitive pairs. The new concurrent approach can't do that pre-filter (no ordering guarantee), which is correct — but it means every ambiguous pair always triggers an LLM call, even if transitive merges in concurrent batches would have made it redundant. This is a compute-cost trade-off vs. latency. Worth a comment in the code acknowledging it.

#### 3. `_embedding_dedup` normalization change for zero-magnitude vectors

`np.divide(matrix, norms, out=np.zeros_like(matrix), where=norms > 0)` produces zero vectors for zero-magnitude embeddings, which then yield similarity 0. The old `cosine_similarity` would produce NaN. In practice zero embeddings are rare, so this is low risk, but the behavioral diff is worth noting.

#### 4. Progress granularity reduced

Progress is now reported only when each document starts (inside `_produce_document`), not per-batch. For documents with many batches, the progress bar will stall at a single value for a long time. Consider passing per-batch progress info back from workers via a shared counter, or accept the coarser granularity (MAP is only ~60% of the total progress range anyway).

#### 5. No test additions

This PR touches 3 files with significant concurrency / algorithmic changes and no test coverage. At minimum, adding a basic test for the `LLMCallPool` `max_pending` backpressure behavior and the blockwise dedup logic would increase confidence.

---

### 🔍 Minor / style

- The `_call_model` inner function inside `_resolve_batch` just wraps `asyncio.wait_for(gen_json(...), timeout=llm_timeout)` — it could be inlined.
- `persist_wiki_pages_to_es` / `persist_wiki_page_graph_to_es` switch to keyword args: cosmetic, parameter names match, so it's compatible.
- `_pool_managed = True` on `PooledChatModel` appears unused in this PR — is it for a follow-up?

---

**Verdict:** Approve with suggestions. The core design is sound; the queue sizing and the ambiguity pre-filtering trade-off are the main things to discuss or document.

PR #17201, opened 2026-07-21
fix(deepdoc): keep pipe tables inside markdown code fences

Yurii214:
### Summary

`RAGFlowMarkdownParser.extract_tables_and_remainder` matches pipe tables anywhere in the document, with no notion of fenced code. A table inside a fenced code block is therefore appended to `tables` and deleted from the remainder.

For any doc that documents markdown syntax (or LLM-generated docs containing a ```` ```markdown ```` block), the example is destroyed. Given this input:

~~~
# Guide

Write a table like this:

```markdown
| Name | Value |
| --- | --- |
| A | 1 |
```

Done.
~~~

`extract_tables_and_remainder(doc, separate_tables=True)` currently returns:

~~~
remainder: '# Guide\n\nWrite a table like this:\n\n```markdown\n\n```\n\nDone.\n'
tables:    ['\n| Name | Value |\n| --- | --- |\n| A | 1 |\n']
~~~

The fence is left hollow and the example resurfaces as a bogus standalone table chunk (via `rag/app/naive.py`, which feeds the remainder to `MarkdownElementExtractor` and renders each extracted table separately).

`MarkdownElementExtractor` in this same module already shields fenced code (`_fenced_code_ranges` / `_protected_ranges`) so delimiters never split inside a fence — table extraction simply missed the same case. This skips matches that start inside a fence; genuine tables outside fences are extracted exactly as before.

### Tests

Two regression tests added to `TestMarkdownTableDedup`: one asserting a fenced example survives, one asserting a real table alongside a fenced example is still extracted. Both fail without the change and pass with it; the existing 12 markdown tests are unaffected.

### DIFF
--- a/deepdoc/parser/markdown_parser.py
+++ b/deepdoc/parser/markdown_parser.py
@@ -90,7 +90,12 @@ class RAGFlowMarkdownParser:
         TAGS = ["table", "td", "tr", "th", "tbody", "thead", "div"]
         table_with_attributes_pattern = re.compile(rf"<(?:{'|'.join(TAGS)})[^>]*>", re.IGNORECASE)
 
+        tag_fenced_spans = [(m.start(), m.end()) for m in fence_pattern.finditer(working_text)]
+
         def replace_tag(m):
+            if any(start <= m.start() < end for start, end in tag_fenced_spans):
+                # an html example inside a fence keeps its attributes verbatim.
+                return m.group()
             tag_name = re.match(r"<(\w+)", m.group()).group(1)
             return "<{}>".format(tag_name)
 
@@ -122,7 +127,10 @@ class RAGFlowMarkdownParser:
                 nonlocal working_text
                 new_text = ""
                 last_end = 0
+                fenced_spans = [(m.start(), m.end()) for m in fence_pattern.finditer(working_text)]
                 for match in html_table_pattern.finditer(working_text):
+                    if any(start <= match.start() < end for start, end in fenced_spans):
+                        continue
                     raw_table = match.group()
                     tables.append(raw_table)
                     if separate_tables:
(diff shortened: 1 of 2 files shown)


Yurii214:
Good catch — that was a real gap, applied in `d173d21`.

Confirmed the same failure on the HTML passes before fixing: for a ```html fence containing `<table border="1">…`, the tag pass stripped the attribute and the html-table pass extracted the table, leaving the fence hollow (```html + two blank lines). Both passes now use the same fence spans.

HTML tables outside fences are unchanged — still normalised to `<table>` and extracted — and there's a regression test covering the fenced case.

Yurii214:
Both applied in `6e87376`.

- Good catch on the test: `<table border="1">` never reaches `replace_html_tables` because the `"<table>" in working_text` fast-path guard doesn't match a tag with attributes, so that pass (and its fence check) was untested. Added `test_bare_html_table_inside_code_fence_is_not_extracted` with a bare `<table>` — I confirmed it fails if the html-table fence skip is removed.
- Added low-volume `debug` logs to the three fence-skip branches (markdown-table, html-tag, html-table), logging only the pass name and match offset, no document content.

Yurii214:
Heads-up: the red `ragflow_preflight` here is an environment-setup flake, not this change. It fails in "Prepare Python test environment" while `uv sync` clones a dependency:

```
Failed to download and build graspologic @ git+https://gitee.com/infiniflow/graspologic.git
fatal: could not read Username for 'https://gitee.com': terminal prompts disabled
```

So the gitee clone couldn't authenticate/reach the host — the test code never ran. The same preflight passed on my other PRs (#17062, #17064), so it looks transient. A re-run should clear it. This PR only touches `deepdoc/parser/markdown_parser.py` (+ its unit test); locally `test_markdown_parser.py` is green and `ruff check` passes.

PR #17205, opened 2026-07-22
Add ulimits for cpu and gpu services in Docker Compose

Ro0tk1t:
### Summary

fix OSError: [Errno 24] Too many open files


myf-bee (commented):
it is  Looks Good To Me

PR #17209, opened 2026-07-22
fix: mask stored provider API key with fixed-length dots

euvre:
### Summary

The model provider settings page rendered the stored API key in a standard password input, so the number of dots matched the key's real length and leaked it through the UI (e.g. ZHIPU-AI vs SILICONFLOW showed different dot counts).

This PR adds a dedicated `MaskedApiKeyInput` for saved provider instances:

- Hidden (default): always renders a fixed-length mask regardless of the key's real length, so the dot count no longer leaks key-length information.
- Revealed (eye toggle): shows the real stored key, editable as before.
- Focusing the masked field starts a fresh entry: the first keystroke replaces the stored key; blurring without typing keeps it.

The form value still carries the real key, so save, connection verification, and dirty tracking behave exactly as before. Draft (new instance) cards keep the standard password input.

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/components/masked-api-key-input.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/components/masked-api-key-input.tsx
@@ -1,0 +1,111 @@
+/*
+ *  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Nice fix for the API key length leakage issue. The `MaskedApiKeyInput` component is well-designed:

- **Fixed-length mask `********`** prevents leaking key length via dot count ✅
- **Eye toggle** reveals the real key when needed ✅
- **Focus → type → replace** pattern works: focusing the masked field starts a fresh entry, first keystroke replaces the stored key, blur-without-typing preserves it ✅
- **Draft/new instance cards** keep the standard password input (correctly gated by `!isDraft`) ✅

One edge case to be aware of: if a user types a character while the field is masked, then clicks the eye toggle before blurring, `setDraft(null)` in the toggle handler clears the visual draft state but the `onChange` has already updated the form value. This means the display may show the original key (or mask, depending on toggle state) while the form holds the typed character. This is a rare interaction and not a blocker — just something to keep in mind.

Otherwise LGTM.

PR #17213, opened 2026-07-22
Feat: openrouter embedding

Lynn-Inf:
### Summary

Support OpenRouter embedding model.


wangq8 (commented):
Thanks for adding OpenRouter embedding support.

The implementation follows the established patterns closely (`OpenAIEmbed`, `ZhipuEmbed`), and auto-registration via `_FACTORY_NAME` will work correctly.

A few observations:

1. **JSON key parsing** — Handling both `{api_key, provider_order}` format and plain string keys is consistent with how chat models handle OpenRouter credentials. Nice.

2. **`extra_body={"drop_params": True}`** — Good call; this prevents OpenRouter from rejecting unknown parameters that the OpenAI client may send.

3. **Error handling** — The `_call` method does not wrap the API call with `_raise_model_exception_if_failed` or similar explicit error handling as some other embedding models do. Relying on the OpenAI client to raise exceptions is generally fine, but you may want to verify that OpenRouter error responses (e.g., 402 for billing, 429 for rate limits) surface a clear enough message to the user through the existing embedding error flow.

Otherwise LGTM.

PR #17214, opened 2026-07-22
fix: eliminate one-frame flash of title without form content in data source pages

euvre:
### Summary

Both the data source detail page and the AddDataSourceModal rendered the title/icon in the same frame data arrived, but form fields and default values populated one frame later via `useState` + `useEffect`, causing a visible flash where only the title appeared.

The fix replaces `useState` + `useEffect` with `useMemo` for deriving `fields` and `defaultValues` from the fetched data, so the full UI renders in a single frame.


wangq8 (commented):
Clean refactor. Replacing `useState` + `useEffect` with `useMemo` for `fields` and `defaultValues` ensures they are computed synchronously during render, eliminating the one-frame flash where the title appeared before form content. The `isDirty` reset is correctly simplified to depend only on `detail`.

One minor suggestion: the `customFields` `useMemo` in the detail page has an empty dependency array `[]` and is only consumed by `fields`. Since it never changes, it could be moved outside the component entirely. Not a blocker, just a cleanup opportunity.

dcc123456 (approved):

PR #17220, opened 2026-07-22
fix: restore empty model warning in dataset creating dialog

euvre:
### Summary

When clicking "Create knowledge base" without any models configured,
the warning modal no longer appeared. The root cause is that
`useFetchDefaultModelDictionary` was called without `showEmptyModelWarn`,
defaulting to `false`, so `useWarnEmptyModel` never triggered the
warning dialog.

This PR passes `true` to restore the warning, consistent with how
`use-create-chat.ts` already uses `useFetchDefaultModelDictionary(createChatVisible)`.


wangq8 (commented):
Simple fix, LGTM. ✅

Passing `true` to `useFetchDefaultModelDictionary` restores the `showEmptyModelWarn` behavior. The PR description correctly identifies that the root cause was the boolean defaulting to `false`. The consistency with `use-create-chat.ts` (which already passes `createChatVisible`) is noted.

dcc123456 (approved):

PR #17221, opened 2026-07-22
fix: prevent empty-state flash on knowledge-base and chat list loading

euvre:
### Summary

When navigating to the knowledge base or chat pages, a brief flash of the empty-state card ("尚未创建知识库" / "尚未创建聊天应用") is shown before the actual data loads. This happens because `useQuery` uses `initialData: []`, causing the page to treat the empty list as "no data" during the initial fetch.

**Changes:**
- Replaced `initialData` with `placeholderData` in `use-knowledge-request.ts` and `use-chat-request.ts` to preserve previous data during pagination/refetch
- Added a loading spinner in `datasets/index.tsx` and `next-chats/index.tsx` when the list is empty and the query is still fetching, instead of showing the empty card immediately


wangq8 (commented):
Good fix for the empty-state flash. ✅

**Reviewed:**
- `initialData` → `placeholderData`: This is the correct TanStack Query pattern for preserving previous list data during refetch/pagination. `initialData` puts data into the cache as if it's real data, while `placeholderData` is only shown during loading. The fallback `previousData ?? { ... }` handles the first render before any data is fetched.
- Loading spinner: Both `datasets/index.tsx` and `next-chats/index.tsx` now show a `<Spin size="large" />` when the list is empty \*\*and\*\* loading is still in progress, preventing the empty-state card from flashing.
- `loading` is destructured from the hook return — make sure these hooks actually expose `loading`. (Looks like they do since they use TanStack Query which returns `isLoading`/etc.)

Non-blocking nit: consider checking `isFetching` vs `isLoading` semantics. `isLoading` is true only on the first load (no data ever fetched), while `isFetching` is true during any background refetch. For this use case `isLoading` is the right choice since we only want the spinner on the initial empty state. Just confirming the hook is exposing the right loading state.

dcc123456 (approved):

dcc123456 (commented):
The problems raised by the robot need to be solved.

PR #17223, opened 2026-07-22
fix: use FormLabel required prop for consistent asterisk color in search settings

euvre:
### Summary

The required asterisk indicator in the search settings panel was using two different color classes:
- The "Knowledge Base" field used `FormLabel`'s built-in `required` prop, which renders the asterisk via `before:text-state-error`.
- The "Model" field (under rerank) used a manually rendered `<span className="text-destructive">`, resulting in a different shade.

This PR replaces the manual span with `FormLabel`'s `required` prop so both required asterisks use the same `text-state-error` color, ensuring visual consistency.


wangq8 (commented):
Straightforward fix, LGTM. ✅

The manual `<span className="text-destructive">` was replaced with `FormLabel`'s built-in `required` prop, which uses `before:text-state-error` for the asterisk. This makes the two required field asterisks (`Knowledge Base` and `Model`) consistently use `text-state-error` instead of mixing `text-destructive` and `text-state-error`.

dcc123456 (approved):

PR #17224, opened 2026-07-22
fix: render documentation URLs as hyperlinks in model tooltips

euvre:
### Summary

Model settings tooltips on the System Default Models page contain URLs such as `https://ragflow.io/docs/dev/supported_models` that were rendered as plain text, making it cumbersome for users to access the documentation.

This PR introduces a `LinkifyText` component that auto-detects `https?://` URLs in tooltip strings and renders them as clickable anchor tags with underline styling that open in a new tab. The component is applied to all model tooltips (LLM, Embedding, VLM, ASR, Rerank, TTS) in the system settings view.

### Changes

- Added `web/src/components/linkify-text.tsx`: reusable component that converts plain-text URLs into clickable `<a>` tags
- Updated `web/src/pages/user-setting/setting-model/layout/system-setting.tsx`: wrap tooltip content with `LinkifyText`

### DIFF
--- a/web/src/components/linkify-text.tsx
+++ b/web/src/components/linkify-text.tsx
@@ -1,0 +1,68 @@
+/*
+ *  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM, clean utility component.

**Reviewed:**
- `LinkifyText`: The regex `URL_REGEX` uses a well-known pattern for URL detection. It correctly stops capturing at characters that aren't valid in URLs (`<>"){}|^`[]`) to avoid false positives. The `stopPropagation()` on click is good UX for tooltips — prevents the tooltip from closing when clicking the link.
- Proper `target="_blank"` with `rel="noopener noreferrer"` — no security concern.
- Falls back gracefully to a plain `<span>` when children isn't a string.
- Usage in `system-setting.tsx` wraps only the tooltip content, which is the right scope.

One nit (non-blocking): the `while` loop uses `URL_REGEX` with the global flag via `.exec()`. This is correct as written because each call to `exec()` advances `lastIndex` — but worth noting that mutating `children` inside the loop would cause issues. Since the code only reads `children`, it's fine.

✅

dcc123456 (approved):

PR #17225, opened 2026-07-22
fix: prevent thinking content from leaking into visible answer in shared chat (Go mode)

euvre:
### Summary

This PR fixes a bug in Go mode where the shared chat embed (chat-share) incorrectly displayed reasoning/thinking content in the main answer body instead of isolating it in the collapsible "Thinking..." section.

**Root cause (two layers):**

1. `EnterReasoning()`/`ExitReasoning()` in `think_tag.go` did not write `<think>`/`</think>` markers into `fullText`, so `ExtractVisibleAnswer()` could not strip reasoning from the final answer when the model exposes reasoning through a separate channel (e.g. `delta.reasoning_content`).

2. `bot_completion.go` sent accumulated full answer text (`fullAnswer`) on every SSE frame instead of per-frame deltas. The frontend accumulates the answer and appends `<think>`/`</think>` around `start_to_think`/`end_to_think` markers, so sending accumulated text caused reasoning to be duplicated in the visible answer.

**Fix:**

- `EnterReasoning()`/`ExitReasoning()` now append `<think>`/`</think>` markers to `fullText`, ensuring `ExtractVisibleAnswer()` correctly strips reasoning from the final visible answer.
- Marker frames (`StartToThink`/`EndToThink`) now send empty `Data` instead of the accumulated `fullAnswer`.
- Regular answer frames now send `res.Answer` (the per-frame delta) instead of the accumulated `fullAnswer`.

This aligns the Go SSE streaming path with the Python `async_chat` behavior, which yields deltas rather than accumulated full text.


wangq8 (commented):
Nice fix, LGTM. The root cause analysis in the PR description is thorough.

**What I reviewed:**
- `think_tag.go`: Adding `<think>`/`</think>` markers in `EnterReasoning()`/`ExitReasoning()` is the correct fix for the first layer — ensures `ExtractVisibleAnswer()` can strip reasoning that comes through a separate channel.
- `bot_completion.go`: Switching marker frames to empty `Data` and answer frames to `res.Answer` (delta) instead of accumulated `fullAnswer` fixes the duplication issue. This aligns with the Python path's delta semantics.
- `think_tag_test.go`: The new test `TestReasoningChannelWrapsThinkTags` covers the exact scenario described — reasoning channel → visible answer, verifying `ExtractVisibleAnswer` returns only the visible part.

No regressions spotted. The test coverage on the think-tag state machine is solid. ✅

buua436 (approved):
LGTM

PR #17226, opened 2026-07-22
Fix punctuation in useKnowledgeGraphTip tooltip

euvre:
### Summary

The Chinese locale's `useKnowledgeGraphTip` text in `web/src/locales/zh.ts` ended with a question mark (`？`), which is inconsistent with the English version ("Whether to use knowledge graph(s) ...", a statement).

This PR changes the trailing `？` to `。`, making the Chinese tooltip a proper statement that matches the English counterpart.


wangq8 (commented):
LGTM. Simple, correct fix — the English version ends with a period (statement), and the Chinese translation should match. The '是否...' construction is a descriptive statement, not a question, so `。` is the right punctuation here.

PR #17229, opened 2026-07-22
fix: MinerU.Net and PaddleOCR.Net missing icons

euvre:
### Summary

The `MinerU.Net` and `PaddleOCR.Net` model providers were not displaying their icons in the UI, falling back to the default placeholder icon instead.

**Root cause:** The `LlmIcon` component in `web/src/components/svg-icon.tsx` resolves icons by checking if the provider name exists in the `themeIcons` and `svgIcons` arrays. While `IconMap` correctly maps both `MinerU.Net`→`mineru` and `PaddleOCR.Net`→`paddleocr`, the `.Net` variants were not included in these arrays. As a result, they skipped the SVG rendering path and fell back to the iconfont sprite, which has no `mineru`/`paddleocr` symbol, ultimately rendering the default icon.

**Fix:** Added `LLMFactory.MinerUNet` to both `themeIcons` and `svgIcons` arrays, and `LLMFactory.PaddleOCRNet` to the `svgIcons` array. This ensures the `.Net` variants reuse the existing `mineru-dark`/`mineru-bright` and `paddleocr` SVG assets — no new assets needed.


wangq8 (commented):
LGTM. Straightforward fix.

The root cause analysis is correct — `themeIcons` and `svgIcons` arrays gate the SVG rendering path, and the `.Net` variants were missing. Since the existing `mineru`/`paddleocr` SVG assets work for both the standard and `.Net` variants, no new assets are needed.

Verified:
- `MinerUNet` added to both `themeIcons` and `svgIcons` ✅
- `PaddleOCRNet` added to `svgIcons` ✅
- No other missing `.Net` variants in the icon arrays ✅

No regressions.

PR #17231, opened 2026-07-22
fix: exclude TTS models from agent component model selector

euvre:
### Summary

The agent component's model selector incorrectly included TTS (text-to-speech) models in its dropdown. This was caused by the `showSpeech2TextModel` prop being passed to `LargeModelFormField` in the agent form, which added `'tts'` to the `modelTypes` filter array.

This PR removes the `showSpeech2TextModel` prop from the agent form and cleans up the now-dead code path across all affected components (`NextLLMSelect`, `LargeModelFormField`), ensuring the agent model selector only shows chat and vision models — consistent with other agent component forms (Categorize, Compilation, Extractor).


wangq8 (commented):
LGTM. Clean cleanup.

Removing the `showSpeech2TextModel` prop and its dead code path is the right approach — the agent component model selector should only show chat and vision models, consistent with other agent component forms (Categorize, Compilation, Extractor).

Verified:
- No other callers pass `showSpeech2TextModel` — only `agent-form/index.tsx` (removed) ✅
- `NextInnerLLMSelectProps` and the inner select logic are fully cleaned up — no dangling references ✅
- The `modelTypes` array now correctly reflects `[chat, vision]` only ✅

No regressions.

dcc123456 (approved):

PR #17233, opened 2026-07-22
fix(agent): prevent double reply when Agent node is followed by Message node

euvre:
### Summary

When the canvas layout is **Agent → Message**, the agent replies twice: the Agent component streams its answer via `StreamCallback` during the ReAct loop, and the downstream Message component then re-emits the same content through `EmitCanvasMessage` / `EmitAgentMessage`. This is particularly noticeable with `qwen-flash` but affects all models.

**Root cause:** In `internal/agent/component/message.go`, the Message node unconditionally emits the rendered content to SSE without checking whether an upstream Agent already streamed its answer.

**Fix:** Guard the Message component's emission with `runtime.AgentMessageEventsEmitted(ctx)`. When an upstream Agent has already streamed, the Message node skips SSE output. The rendered content is still stored in `outputs["content"]` for state persistence and the service-layer answer collector, so canvas history and conversation continuity are unaffected.

The standalone Message node (without a preceding Agent) continues to emit normally.

### DIFF
--- a/internal/agent/component/message.go
+++ b/internal/agent/component/message.go
@@ -216,7 +216,13 @@ func (m *MessageComponent) Invoke(ctx context.Context, inputs map[string]any) (m
 			Text:   resolved,
 		})
 	}
-	if rendered != "" {
+	// Skip emission when an upstream Agent component already streamed its
+	// answer via the agent message emitter. Without this guard the user sees
+	// the same content twice: once from the Agent's StreamCallback during the
+	// ReAct loop, and again from the Message node's own emit call here.
+	// The rendered content is still stored in outputs["content"] for state
+	// persistence and the service-layer answer collector.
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Nice fix with a good test.

The guard `!runtime.AgentMessageEventsEmitted(ctx)` is consistent with the existing pattern — the Agent component already checks this at `agent.go:274` and `agent.go:683`. The Message node was the missing piece.

Key points:
- The rendered content is still stored in `outputs["content"]`, so state persistence and the service-layer answer collector are unaffected ✅
- The standalone Message node (no upstream Agent) still emits normally since `AgentMessageEventsEmitted` will be `false` in that case ✅
- Test coverage verifies both the skip-when-emitted and normal-emission paths ✅

No regressions. Well done.

PR #17234, opened 2026-07-22
fix: keep memory card position stable after rename or config update

euvre:
### Summary

Memory cards would jump to the top of the list whenever they were renamed or had their configuration modified.

**Root cause:** The memory list query in `internal/dao/memory.go` was ordered by `ORDER BY m.update_time DESC`. GORM's `BeforeUpdate` hook (in `internal/entity/base.go`) automatically refreshes `update_time` on every `.Updates()` call, so any rename or config change would give the card the newest `update_time`, causing it to move to the front of the list.

**Fix:** Change the sort field from `update_time` to `create_time` in the `GetByFilter` query. This keeps card positions stable based on creation order, regardless of subsequent edits. This is also consistent with the frontend, which already displays `create_time` as the card timestamp in `web/src/pages/memories/memory-card.tsx`.

```diff
-		ORDER BY m.update_time DESC
+		ORDER BY m.create_time DESC
```


wangq8 (commented):
LGTM. Simple, correct fix.

The root cause analysis in the PR body is spot-on — GORM's `BeforeUpdate` hook bumps `update_time` on every `.Updates()` call, so sorting by it causes cards to reposition on rename/config-change. Switching to `create_time` is the right call: it's stable, and the frontend already displays `create_time` as the card timestamp (`memory-card.tsx`), so the ordering stays consistent with what the user sees.

No regressions, no side effects.

wangq8:
LGTM

PR #17237, opened 2026-07-22
docs: point begin/message component links at .md, not .mdx

muhzuhaib:
### Summary

Most files in `docs/guides/agent/agent_component_reference/` are `.mdx`, but `begin` and
`message` are `.md`. Four links use the `.mdx` extension for those two, so Docusaurus cannot
resolve the reference and they render as broken links.

| File | Link |
|---|---|
| `docs/references/http_api_reference.md` | `agent_component_reference/begin.mdx` → `begin.md` |
| `docs/guides/memory/use_memory.md` | `agent_component_reference/message.mdx` → `message.md` |
| `docs/release_notes.md` (×2) | `begin.mdx` → `begin.md`, `message.mdx` → `message.md` |

Extension only — no text changed. The `retrieval.mdx` link on the same line in `use_memory.md`
is correct and is left alone.

### Note for maintainers (deliberately not in this PR)

`docs/release_notes.md` links to seven more components that have no file under any extension:
`generate`, `interact`, `keyword`, `rewrite`, `concentrator`, `template` and `note`. Those look
like components that were removed after the release notes were written, so whether to repoint,
unlink or leave them as a historical record is your call rather than something to fold into an
extension fix. Happy to do it in a follow-up if you tell me which you'd prefer.


muhzuhaib:
Correction — I've reduced the scope of this PR and force-pushed.

It originally also rewrote two root-absolute links (`/migration` and `/configurations`) to
relative paths, on the claim that they were broken. That claim was wrong: every page under
`docs/` declares a flat root `slug:`, so `/migration` and `/configurations` are valid routes and
those links already worked. I'd checked that the anchors existed but not that the routes did.

Those two changes are gone. What's left is only the `.mdx` → `.md` extension fix, which I've
verified: `begin.md` and `message.md` are the actual filenames and no `.mdx` variant exists.
Sorry for the noise.

PR #17247, opened 2026-07-22
Fix: go restful api tests

Lynn-Inf:
### Summary

Align `create_memory` go api to python api.


wangq8 (commented):
Overall looks good — the TrimSpace change in `CreateMemory` correctly aligns Go with Python behavior, and the added argument error prefixes ensure proper error code classification.

A few notes:

1. **conftest.py**: Commenting out tests with `#` instead of using `@pytest.mark.skip(reason="...")` makes them invisible to the test runner. Consider using `@pytest.mark.skip` or `@pytest.mark.xfail` so they still appear in the test report as skipped rather than silently vanishing.

2. **Model IDs in tests**: The change from `@Builtin` → `@Local@Builtin` and `@ZHIPU-AI` → `@CI@ZHIPU-AI` assumes `Local` and `CI` instance names exist in all test environments. Are these always present, or could this break CI/Docker test setups?

3. **isArgumentError**: The new prefix `"name length is"` is intentionally broad — it will match any name-length error. Is this prefix also used by ValidateName? I see the error format is `"name length is %d which is large than %d"` — which contains a typo in "large" (should be "larger"). Not introduced by this PR, but worth a follow-up fix.

PR #17250, opened 2026-07-22
fix: improve model merging logic in useModelsDerived for better instance-specific values

dcc123456:
### Summary

fix: improve model merging logic in useModelsDerived for better instance-specific values


wangq8 (commented):
LGTM. This is a good fix — the old behavior where catalog entries always overwrote instance entries meant any user-customized model settings (like `max_tokens`) would silently reset to upstream defaults. The new logic preserves instance-specific values and only falls back to catalog entries for models not yet added.

One minor suggestion: consider adding a unit test that verifies instance entries are preserved when a catalog entry with the same name exists, to prevent regressions — though not a blocker for this PR.

Code is clean and the updated comment accurately describes the new behavior.

PR #17252, opened 2026-07-22
fix: correct auto-parse switch thumb alignment in data source settings

euvre:
### Summary

Fix the auto-parse switch in the data source settings page where the toggle thumb appeared centered instead of right-aligned when turned on.

The `Switch` component in `link-data-source.tsx` had its track width overridden from the default `w-7` (28px) to `w-8` (32px) via `className="w-8 h-4"`. However, the internal thumb translate distance (`translate-x-3` = 12px) is calibrated for the default `w-7` width. With the wider track, the thumb only traveled about 75% of the way, leaving a visible gap on the right side and making it look centered rather than fully right-aligned.

The fix removes the `className="w-8 h-4"` override so the Switch reverts to its default `h-4 w-7` dimensions, where the thumb displacement correctly aligns with the track edge when checked.


wangq8 (commented):
Simple, correct fix.

The root cause is straightforward: the Switch thumb translate distance (`translate-x-3` ≈ 12px) is calibrated for the default `w-7` (28px) track. Overriding to `w-8` (32px) leaves a 4px gap on the right when checked, making the thumb appear centered instead of right-aligned. Removing the override restores the correct dimensions.

One-line removal, zero risk of regression since the component reverts to its tested default sizing. LGTM.

dcc123456 (approved):

PR #17254, opened 2026-07-22
fix: make qwen3-vl-plus usable as chat/vision model in Go mode

euvre:
### Summary

The model `qwen3-vl-plus` was completely unusable in Go mode. Two issues caused this:

1. **Missing `chat` model type (primary cause):** In the Go provider config (`conf/models/aliyun.json`), `qwen3-vl-plus` only had `model_types: ["vision"]`, whereas the Python config (`conf/llm_factories.json`) lists `["image2text", "chat"]`. Without the `chat` type bit, the model failed every `ModelTypeChat` validation — it could not be set as a dialog model, assistant LLM, or used in chat. The model type bitmask only had `8` (vision) and was missing `1` (chat).

2. **qwen3 thinking default not handled:** DashScope defaults `enable_thinking` to `true` for qwen3-series models. Python's `chat_model.py` explicitly defaults `enable_thinking=False` for all qwen3 models, but the Go `AliyunModel` driver did not send this parameter when it was not explicitly configured, causing unexpected reasoning output.

### Changes

- **`conf/models/aliyun.json`**: Added `"chat"` to `model_types` for `qwen3-vl-plus` (now `["vision", "chat"]`), added `"tools": {"support": true}`, and corrected `max_tokens` from `260096` to `256000` to match the Python config.
- **`internal/entity/models/aliyun.go`**: Added `applyQwen3ThinkingDefault()` helper that defaults `enable_thinking=false` for qwen3-series models when not explicitly configured, called in both `ChatWithMessages` and `ChatStreamlyWithSender`. This mirrors Python's `chat_model.py` behavior.


wangq8 (commented):
Looks good — both issues are correctly diagnosed and fixed.

**What I verified:**

- Python `llm_factories.json` lists `qwen3-vl-plus` with `["image2text", "chat"]`. The Go config was indeed missing `chat`, making the model fail `ModelTypeChat` validation. Adding `"chat"` fixes this and adding `"tools": {"support": true}` brings it in line with the Python config.
- `max_tokens` correction from `260096` → `256000` matches the Python baseline.
- `applyQwen3ThinkingDefault` is only triggered for Aliyun/DashScope models, which is correct — this is a DashScope-specific behavior where `enable_thinking` defaults to `true` for qwen3. The helper is called in both `ChatWithMessages` and `ChatStreamlyWithSender`, covering both sync and streaming paths.
- The function checks whether `enable_thinking` is already explicitly set by the user before applying the default, so advanced users can still enable reasoning if they want.

**One thing to note:** The `strings.Contains(strings.ToLower(modelName), "qwen3")` match will catch `qwen3-vl-plus`, `qwen3-235b-a22b`, etc., but also false-positive on something like `qwen3point5-something` if it ever exists. That is fine for now — easy to tighten later if needed.

No regressions or concerns. 👍

wangq8:
@euvre resolve the conflict.

Lynn-Inf (approved):
LGTM

PR #17256, opened 2026-07-22
fix(agent): wire memory persistence in Message component (Go runtime)

euvre:
### Summary

The Go Message component never persisted conversation turns to memory, even when `memory_ids` was configured in the canvas DSL via the "Save to Memory" UI. Two root causes:

**Bug 1 — `memory_ids` never reached `Invoke`:**
`NewMessageComponent(params)` consumed the DSL params at construction time but only stored `text`, `output_format`, `auto_play`, `voice`, and `lang` as instance fields. It discarded `memory_ids` and `user_id`. Meanwhile, `Invoke` read `memory_ids` only from the runtime `inputs` map, which the pipeline populates solely from upstream component outputs — never from the node's own DSL params. So the memory-save branch was dead code in production.

**Bug 2 — `MemorySaver` was never wired:**
`component.SetMemorySaver()` was never called at boot. `GetMemorySaver()` always returned `stubMemorySaver{}`, which returns `ErrMemoryServiceMissing` unconditionally.

**Fix:**
- Store `memoryIDs` and `userID` as instance fields in the constructor; `Invoke` reads them from `m.memoryIDs` / `m.userID` with runtime-input overrides — the same pattern already used for `output_format`, `auto_play`, `voice`, `lang`.
- Remove the `memory_save` bool gate: the Python Message component saves whenever `memory_ids` is non-empty, and the frontend does not send a `memory_save` toggle.
- Add `memorySaverAdapter` (`internal/service/memory_saver_adapter.go`) that bridges `component.MemorySaver` to `MemoryService.AddMessage` — the same access-filtering + embedding + task-queueing pipeline the REST API uses.
- Call `component.SetMemorySaver(...)` in `cmd/ragflow_server.go` at boot.
- Add `TestMessage_MemorySave_FromDSLParams` to verify the production path where `memory_ids` comes from DSL params (not runtime inputs).

### DIFF
--- a/internal/agent/component/message.go
+++ b/internal/agent/component/message.go
@@ -299,6 +299,12 @@ func (m *MessageComponent) Invoke(ctx context.Context, inputs map[string]any) (m
 		if userID == "" {
 			userID = m.userID
 		}
+		// If userID is a canvas variable reference (e.g. "{cpn@user_id}"),
+		// resolve it against the current state. Mirrors Python's
+		// agent/component/message.py:569-571.
+		if userID != "" && runtime.VarRefPattern.MatchString(userID) {
+			userID = runtime.ResolveTemplateForDisplay(userID, state)
+		}
 		saver := GetMemorySaver()
 		saveErr := saver.Save(ctx, MemorySaveRequest{
 			MemoryIDs:     memIDs,
--- a/internal/agent/component/message.go
+++ b/internal/agent/component/message.go
@@ -56,6 +56,8 @@ type MessageComponent struct {
 	autoPlay     audio.Engine
 	voice        string
 	lang         string
+	memoryIDs    []string
+	userID       string
 }
 
 // NewMessageComponent constructs a Message component. The params map
@@ -82,6 +84,8 @@ func NewMessageComponent(params map[string]any) (Component, error) {
 		format = OutputFormat(v)
 	}
 	engine, voice, lang := extractAudioConfig(params)
+	memIDs := extractMemoryIDsFromAny(params["memory_ids"])
+	userID, _ := params["user_id"].(string)
 	return &MessageComponent{
 		name:         componentNameMessage,
 		text:         tpl,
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for this fix — it addresses a real gap in the Go runtime where memory persistence was effectively dead code.

**What I checked:**

- `MemorySaveRequest` already carries `UserID` in `memory_save.go`, so the new population in `Invoke` is consistent with the existing wire shape.
- The DSL-params fallback mirrors the existing pattern used by `output_format`, `auto_play`, `voice`, and `lang` — consistent.
- `isMessageInfraInput` now filters `memory_ids` and `user_id`. This is critical: without it those keys would leak into rendered message variables.
- `memorySaverAdapter` delegates to the same `MemoryService.AddMessage` pipeline the REST API uses (access filtering, embedding, async tasking) — no shortcut.
- The `SetMemorySaver` call in `startServer` happens after `NewMemoryService()`, so ordering is fine. The import change from `_` to regular is intentional.

1. `memorySaverAdapter.Save` returns `fmt.Errorf("memory: saver adapter not initialised")` on nil receiver. At that point the user-facing error would be opaque. Consider using a sentinel error so upstream can distinguish "not wired" from "service rejected" — but this is the same pattern `stubMemorySaver` already uses, so it is consistent.

2. The test `TestMessage_MemorySave_FromDSLParams` covers the happy path well. A follow-up test for the `user_id` DSL fallback path would round things out, though not necessary to block this.

LGTM overall — the fix resolves both root causes cleanly.

internal/agent/component/message.go
```diff
@@ -285,22 +285,32 @@ func (m *MessageComponent) Invoke(ctx context.Context, inputs map[string]any) (m
...
+	// The effective memory IDs come from inputs (runtime override)
+	// or fall back to the DSL-declared m.memoryIDs. This matches
+	// the Python Message component, which saves whenever
+	// memory_ids is non-empty.
+	memIDs := extractMemoryIDs(inputs)
+	if len(memIDs) == 0 {
+		memIDs = m.memoryIDs
+	}
+	if len(memIDs) > 0 {
+		userID := stringFromStateSys(state, "user_id")
```
Lynn-Inf, line 298: Check if `user_id` is a variable — if so, load the value from canvas. See `agent/component/message.py:569-571` for reference.

Lynn-Inf (commented):
One inline comment on `user_id`. Otherwise LGTM.

PR #17258, opened 2026-07-22
fix: prevent summary countdown text from flickering in data source log table

euvre:
### Summary

In the data source settings page, when a task is scheduled (status = SCHEDULE), the Summary column renders a live countdown timer (e.g., "Task starts in 9m 58s") that re-renders every second. Because numeric digits have varying widths in proportional fonts, the text width changes each tick, causing the summary text to visibly jitter left and right.

This PR adds the `tabular-nums` utility class (Tailwind/CSS `font-variant-numeric: tabular-nums`) to the countdown `<span>`, which forces all digits to render at a uniform width. This eliminates the horizontal jitter while keeping the countdown fully functional.


wangq8 (commented):
Clean, minimal fix for a real UX issue. Adding `tabular-nums` to the countdown span forces uniform digit widths, which eliminates the horizontal jitter caused by proportional-font digit width variation. `tabular-nums` is already in use across the codebase (available-models.tsx, sidebar.tsx, doc-preview.tsx, schema-property-editor.tsx), so the utility class is well-tested.

LGTM, approved.

dcc123456 (approved):

PR #17259, opened 2026-07-22
fix: chat attachments not used in Go mode

euvre:
### Summary

In Go mode, when a user attaches files to a chat question, the attached file content was never sent to the LLM. This PR fixes two separate bugs that together caused this behavior.

**Bug 1: `GetFileContents` could not read attachment blobs from storage**

`GetFileContents` (`internal/service/file/file_content.go`) looked up a `File` entity from the database by `id`. However, chat file uploads (via `/api/v1/documents/upload` → `UploadInfos`/`storeUploadInfoBlob`) store blobs directly into object storage (`{userID}-downloads` bucket) **without creating a database `File` record**. The DB query always returned nil, so every attached file was silently skipped (`continue`). The returned text/image lists were empty, meaning no attachment content was ever injected into the prompt.

Fix: read blobs directly from storage using the file dict's `id` (storage location UUID) and `created_by` (bucket owner), constructing the path `{created_by}-downloads/{id}`. This mirrors the existing `ParseAgentUploads` logic and Python's `FileService.get_files` → `get_blob`.

**Bug 2: `empty_response` short-circuit skipped the LLM call when attachments were present**

When a chat has a knowledge base but KB retrieval returns no results, `AsyncChat` (`internal/service/chat_pipeline.go`) checks `empty_response` and returns immediately **before** the attachment text is appended to the system prompt. Since the early return happened before the LLM was ever called, attached file content was never used even if it was successfully parsed (by the Bug 1 fix). The user would always see the fallback message (e.g. "知识库中未找到您要的答案！") regardless of what they attached.

Fix: add `attachments == ""` to the early-return condition. When file attachments are present, skip the `empty_response` short-circuit and fall through to the normal LLM call, where attachment text is appended to the system prompt at `systemPrompt = s.formatPrompt(sp, kwargs) + attachments`.


wangq8 (commented):
Thanks for this fix — good catch on both bugs. Overall looks correct and well-reasoned.

**Bug 1 (file_content.go):** The root cause is clear: `GetFileContents` was looking up DB `File` records, but chat uploads (via `UploadInfos`/`storeUploadInfoBlob`) never create them — blobs go straight to `{user}-downloads` in object storage. Switching to direct storage reads matching `ParseAgentUploads` is the right approach. The permission model simplification from `checkFilePerm` (DB chain traversal) to a direct `createdBy != uid` check is correct here since these are user-owned upload blobs, not knowledgebase files.

**Bug 2 (chat_pipeline.go):** The `empty_response` early-return was firing before attachments were appended to the system prompt, so even correctly-parsed attachments were never sent to the LLM. Adding `attachments == ""` to the guard condition ensures attachments still reach the LLM even when KB retrieval returns nothing. The comment block added is thorough and helpful.

**Tests:** Good simplification — removing the heavy SQLite DB setup in favor of targeted `MemoryStorage` mocks makes the tests faster, simpler, and more focused on the actual behavior being tested.

- The `created_by` default-to-`uid` fallback on line ~127 is good defensive coding for edge cases where the field might be missing.
- The MIME type handling now uses `mime_type` from the file dict (with fallback to extension-based guessing), which is strictly better than the extension-only approach.

LGTM, approved.

wangq8:
@euvre resolve conflict.

wangq8:
LGTM.

PR #17268, opened 2026-07-23
fix: improve page index descriptions

buua436:

Improve the PageIndex extraction template so entity descriptions are concise, source-grounded, and useful for later cross-chunk merging.

Parent headings without direct content now receive a higher-level overview instead of an empty description.

### Type of change

### DIFF
--- a/api/db/init_data/compilation_templates/page_index.yaml
+++ b/api/db/init_data/compilation_templates/page_index.yaml
@@ -8,8 +8,9 @@ config:
       compressed descriptions. Extract each heading as an entity. The
       entity's `name` must be the exact clean heading text. For each heading,
       write a strongly compressed summary of only the content belonging to
-      that heading: normally 1–2 sentences and roughly half the length of a
-      normal summary. Keep only the 2–4 most important facts, prioritizing key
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this improvement! The updated page_index template is a solid enhancement — making entity descriptions more source-grounded, concise, and useful for cross-chunk merging is exactly the right direction.

**What works well:**
- Parent headings without direct content now get a higher-level overview derived from children instead of an empty description — this is a key fix for downstream merging quality.
- Explicitly forbidding meta descriptions like "heading for…" and empty descriptions except when there is truly no source content adds useful guardrails.
- The 1–2 sentence / 2–4 key facts constraint provides concrete guidance without being too rigid.

**Minor considerations (non-blocking):**
- The prompt is significantly longer now, which will increase token cost per page_index call. This is probably fine given the quality gain, but worth keeping an eye on.
- "Roughly half the length of a normal summary" is a bit subjective — LLMs may interpret this inconsistently. Not a problem, just worth monitoring output quality in practice.

Overall LGTM. Thanks for the contribution!

wangq8 (approved):

PR #17270, opened 2026-07-23
fix: optimize knowledge compilation LLM behavior and PageIndex ordering

buua436:

This PR disables model reasoning during knowledge compilation to reduce unnecessary generation overhead.

It also ensures PageIndex entities are ordered according to the document chunk order, instead of relying on optional position metadata.

### Type of change


wangq8 (commented):
Good PR addressing two important issues. Here is my review:

## Reasoning control for knowledge compilation (`knowledge_compile_gen_conf`)

This is the right approach — knowledge compilation is extraction/summarization, not creative generation, so disabling reasoning saves both tokens and latency.

**Suggestions:**
- The model name check `"deepseek-v4" in model_name` is a bit broad — it matches any model variant like `deepseek-v4-pro`. Consider whether this is intentional (all deepseek-v4* models use the same `thinking` API) or if a more precise match is warranted. For now it is fine.
- Consider adding a brief module-level comment explaining *why* reasoning is disabled (token savings, latency, extraction tasks don not benefit from reasoning).

## PageIndex ordering (`chunk_api.py`)

The approach of querying ES for the natural chunk order is sound.

**Suggestions:**
- The `size: 10000` is reasonable but consider adding a comment documenting the limit and its implications for very large documents.
- The exception handler silently continues — this is correct for this use case but ensure the log message is clear enough for debugging if ordering goes wrong.
- The `entity_position` sort using `(min(chunk_indexes), fallback)` for tiebreaking is clean.

## Structure aliases (`_STRUCT_TYPE_ALIASES`)

Good QoL improvement allowing `"graph"` and `"knowledge_graph"` as aliases for `"hypergraph"`.

## `batch_size_cap=1` for page_index

Each chunk processed individually for page_index extraction — correct, since page index entities should be per-chunk.

## `runner.py` string-to-list fix

Simple defensive fix, LGTM.

**Overall:** Well-structured PR with clean separation of concerns. No regressions identified. The changes are consistent across all compilation paths (structure, wiki, dataset_nav, raptor). Approving with the minor documentation suggestions above.

PR #17271, opened 2026-07-23
fix: remove loading spinners during page navigation

euvre:
### Summary

This PR removes the circular loading spinners that briefly flash when navigating between pages (Datasets, Chats, Files Management, Searches), making their behavior consistent with the Agents page where no spinner is shown.

Changes:
- **`routes.tsx`**: Removed the full-screen circular spinner overlay (`defaultRouteFallback`) used as the `<Suspense>` fallback for lazy-loaded routes. The fallback is now `null`, so chunk loading no longer produces any visual artifact.
- **`datasets/index.tsx`**: Replaced the `<Spin>` loading state with an empty container, matching the Agents page pattern.
- **`next-chats/index.tsx`**: Same — replaced `<Spin>` with an empty container.
- **`next-searches/index.tsx`**: Added the missing `isLoading` check from the data hook so that an empty-card state no longer flashes before data arrives. The loading state renders an empty container instead of a spinner.
- **`files/files-table.tsx`**: Replaced the spinning `Loader2` skeleton with an empty table row during loading.

### DIFF
--- a/web/src/pages/datasets/index.tsx
+++ b/web/src/pages/datasets/index.tsx
@@ -5,7 +5,6 @@ import ListFilterBar from '@/components/list-filter-bar';
 import { RenameDialog } from '@/components/rename-dialog';
 import { Button } from '@/components/ui/button';
 import { RAGFlowPagination } from '@/components/ui/ragflow-pagination';
-import { Spin } from '@/components/ui/spin';
 import { useFetchNextKnowledgeListByPage } from '@/hooks/use-knowledge-request';
 import { useQueryClient } from '@tanstack/react-query';
 import { pick } from 'lodash';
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Thanks for this cleanup! The changes are consistent and well-scoped.

**What works well:**
- Removing spinners for page transitions provides a smoother UX, consistent with the Agents page pattern
- Adding the missing `isLoading` check in `next-searches/index.tsx` is a genuine fix — nice catch
- `data-testid` attributes are preserved across all pages

**One concern:** The `Suspense fallback={null}` in `routes.tsx` means lazy-loaded routes show absolutely nothing during chunk loading. On slow connections this could look like a frozen page. Consider keeping a minimal, non-intrusive fallback (e.g., an invisible element that only takes space to prevent layout shift) or at least document this behavioral change.

**Minor:** The empty `<article>` elements in datasets and chats still consume full height — that is intentional and fine for preventing layout jumps.

LGTM overall, just flagging the Suspense edge case for consideration.

PR #17272, opened 2026-07-23
fix: strip YAML frontmatter and defer API reference markdown rendering

euvre:
fix: strip YAML frontmatter and defer API reference markdown rendering

### Summary

The API reference page froze for a long time when opened, and YAML frontmatter from the Docusaurus doc (sidebar_position, slug, sidebar_custom_props) was rendered as visible text on the page.

Two issues are addressed:

1. **Frontmatter leak**: The raw 230KB http_api_reference.md was passed directly to MarkdownPreview, which does not strip YAML frontmatter. Added a removeFrontmatter() utility (same approach already used in markdown-viewer.tsx) so only the markdown body is rendered.

2. **Long unresponsive freeze**: The 230KB document was parsed and rendered synchronously, blocking the main thread on navigation. The page now paints immediately with a loading spinner by (a) lazily loading the @uiw/react-markdown-preview component via React.lazy and (b) deferring the heavy render until the next animation frame via requestAnimationFrame.


wangq8 (commented):
Nice fix @euvre! Two clean improvements:

**Frontmatter stripping:**
- ✅ `removeFrontmatter()` matches the existing implementation in `markdown-viewer.tsx` exactly
- ✅ `useMemo` for `cleanDoc` avoids re-computation on re-render
- ✅ `MarkdownToc` now receives `cleanDoc` instead of raw `apiDoc`

**Lazy loading + deferred render:**
- ✅ `React.lazy` + `Suspense` prevents the 230KB markdown library from blocking initial page load
- ✅ `requestAnimationFrame` deferral lets the page paint before the heavy render starts

**One minor suggestion (non-blocking):** The two-stage loading (rAF spinner → Suspense spinner) could result in a visible flicker. In a follow-up, consider consolidating into a single loading state — but the current approach is already a huge improvement over the multi-second freeze.

Consider refactoring `removeFrontmatter` into a shared utility (e.g. `utils/markdown.ts`) since it's now duplicated — but that's out of scope for this PR.

LGTM :shipit:

dcc123456 (commented):
Need to address the questions raised by the robot.

PR #17274, opened 2026-07-23
fix(agent): return tavily tool errors as result instead of crashing the ReAct agent

euvre:
### Summary

The Go Tavily tool (`TavilyTool` and `TavilyExtractTool`) returned both a JSON error envelope **and** a non-nil Go `error` from `InvokableRun`. Eino's `wrapToolCall` (in `compose/tool_node.go`) discards the result string when `error != nil`, causing the **entire ReAct agent run to crash** with no output visible to the user.

**Root cause:** When the Tavily tool encountered any error (API failure, network error, missing API key), the Go error propagated up through `ToolsNode → ReAct graph → agent runner`, aborting the agent and discarding the useful error JSON that the LLM should have seen.

**Fix:** All recoverable errors are now returned as `(errorJSON, nil)` instead of `(errorJSON, error)`. This ensures the error context reaches the LLM as a tool result, allowing the agent to gracefully respond. This matches the Python implementation's behavior (`agent/tools/tavily.py`), which catches all exceptions and returns `"Tavily error: {last_e}"` without crashing the agent.

Additionally, `include_images` and `include_raw_content` are now always forced to `false` to match the Python implementation, preventing base64 image data from bloating the context window.

The Canvas component path (`ToolBackedComponent.Invoke`) is unaffected — it already handles soft errors via `parseToolEnvelope` checking the `_ERROR` field.


wangq8 (commented):
Thanks for this fix @euvre! The root cause analysis is solid — eino's `wrapToolCall` discarding the result when `error != nil` is indeed the problem, and returning `(errorJSON, nil)` is the correct approach that matches the Python implementation.

**What I verified:**
- ✅ Python `tavily.py` catches all exceptions and returns `"Tavily error: {last_e}"` without crashing — the Go PR now matches this behavior
- ✅ `TavilyExtractTool` already had `IncludeImages: false` hardcoded in the body; adding the same to `TavilyTool` is consistent
- ✅ Tests updated correctly for the new error semantics

**One minor observation (non-blocking):** After this change, the `include_raw_content` and `include_images` merge logic in `mergeTavilyParams` (lines ~368-374) becomes dead code — the values are always overridden to `false` in `InvokableRun` before building the request body. Consider cleaning that up in a follow-up to avoid confusion for future readers.

Otherwise looks good. :shipit:

Haruko386 (approved):
LGTM

PR #17276, opened 2026-07-23
Feat: OpenAI-API-Compatible

Lynn-Inf:
### Summary

Go: add OpenAI‑API‑compatible model provider.


wangq8 (commented):
## Review: Feat: OpenAI-API-Compatible

Overall solid implementation — clean use of struct embedding to reuse `VllmModel`; type inference heuristics and `ListModels` merging are well-structured. A few items to address:

### 🔴 API key in query params (`internal/handler/providers.go:158`)
```go
apiKey := c.Query("api_key")
```
API keys passed as query parameters are logged in access logs, proxy logs, and browser history. This is a security concern for any proxy or logging middleware between the client and RAGFlow. Consider requiring the key through a POST body field or an `Authorization`-type header instead.

### 🟡 Silent live-fetch failures (`internal/handler/providers.go:172`)
```go
if liveModels, err := driver.ListModels(apiConfig); err == nil {
```
When the live API call fails (bad key, unreachable URL, timeout), the error is silently swallowed. The user sees only static models with no indication their credentials were rejected or the endpoint is unreachable. At minimum, log the error; ideally surface it as a warning in the response metadata so the UI can show it.

### 🟡 Missing streaming stubs (`openai_api_compatible.go`)
`AudioSpeechWithSender` and `TranscribeAudioWithSender` both return `fmt.Errorf(...not implemented...)`. The doc comments say this is intentional, but if a caller tries the streaming path for connection verification it will get a hard error. Consider a TODO with a tracking issue or a feature-gate check higher in the call chain.

### 🟢 Structural notes (no action needed)
- `InferModelTypes` checks **rerank** before **embedding**, which correctly handles models named like `bge-reranker-v2` that match both hint sets. Good.
- `ttsVoiceForModel` for cosyvoice returns `modelName + ":anna"` — this is an unusual compound format. If this is the expected convention for CosyVoice on SiliconFlow, a comment confirming would help future readers.
- The `CheckConnection` / `verifyProviderModel` signature change from `[]string` to `[]CheckConnectionModelInfo` is internally consistent — both call sites in `providers.go` are updated, and the `nil` path at L343 works with both signatures.
- PR size is reasonable (450 additions, 30 deletions), single concern, no unrelated changes.

### Suggestion (non-blocking)
The `ListModels` handler change is the most impactful part — it introduces the concept of live model fetching in the API layer. A unit test for the merge/sort logic would be valuable since the dedup (remote overwrites static by name) and alphabetical sort are easy to regression-test.

PR #17283, opened 2026-07-23
Fix: Xinference asr model call

Lynn-Inf:
### Summary

Remove temperature argument in asr model call.


wangq8 (commented):
LGTM. Clean fix — Xinference speech-to-text models do not accept a `temperature` parameter, so removing it avoids unnecessary API errors. The change is minimal and correct.

PR #17285, opened 2026-07-23
refa: improve tree clustering

buua436:

Improve knowledge compilation tree clustering and graph generation.

This change:

- Switches tree compilation from GMM to average-linkage AHC with cosine distance.
- Clusters directly on raw embeddings instead of reducing them with UMAP first.
- Adds cosine-based centroid adjustment and close-cluster merging.
- Enforces the configured maximum cluster count and records clustering statistics.
- Collapses tree nodes that only contain a single child.
- Prevents self-loop relations when parent and child nodes have the same generated title.
- Enables AHC for both compiler and task-executor tree compilation paths.

### Type of change

- [x] Refactor (no breaking API change)

### DIFF
--- a/rag/advanced_rag/knowlege_compile/raptor.py
+++ b/rag/advanced_rag/knowlege_compile/raptor.py
@@ -19,7 +19,7 @@ import logging
 import re
 
 import numpy as np
-from sklearn.cluster import AgglomerativeClustering
+
 from sklearn.mixture import GaussianMixture
 
 from api.db.services.task_service import has_canceled
@@ -42,7 +42,6 @@ from rag.utils.raptor_utils import (
     SUPPORTED_CLUSTERING_METHODS,
     SUPPORTED_TREE_BUILDERS,
 )
-from ._common import knowledge_compile_gen_conf
 
 # Regularization added to GMM covariance diagonals; keeps components
 # from collapsing on singleton/near-identical reduced points.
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Thanks for this PR — the switch from GMM+UMAP to average-linkage AHC with cosine distance is a solid direction. I left some detailed feedback below.

## Overall

The refactor direction makes sense: AHC with cosine distance on raw embeddings avoids the information loss from UMAP and the instability of GMM on arbitrary-dimension reduced data. The `_collapse_unary` helper and self-loop guard are good defensive additions.

## Suggestion: make `ahc_distance_threshold` configurable

The new `ahc_distance_threshold=0.3` parameter on `RecursiveAbstractiveProcessing4TreeOrganizedRetrieval.__init__` is hardcoded with no path to override it from `build_doc_tree` or any config. If 0.3 turns out to be too aggressive or too conservative for certain embedding models or datasets, users have no way to tune it. Consider threading it through `build_doc_tree` and the `RaptorConfig` / ext dict, similar to how `clustering_method` is already handled in `raptor_utils.py`.

## Missing type annotations

The new parameter `ahc_distance_threshold` and the new methods `_merge_close_clusters`, `_collapse_unary` are missing type hints. The rest of the file uses type annotations (`np.ndarray`, `int`, etc.) — worth matching the existing style.

## Dead code in GMM path

In `clustering()`, the top already returns early on `len(embeddings) == 0`, but the GMM branch repeats the same check:

```python
if len(embeddings) == 0:
    return 0, []
```

This is now unreachable and can be removed.

## Potential issue: `_adjust_tree_nodes` with zero-norm embeddings

The new `_adjust_tree_nodes` uses `np.divide(..., out=np.zeros_like(...), where=norms > 0)`. If an embedding is a true zero-vector, its normalized form becomes the zero vector and it will have cosine similarity 0 with every centroid — meaning it gets randomly assigned to the first cluster (`argmax` breaks ties to the first index). In practice zero embeddings are unlikely, but it is worth being aware of. The same applies to `_merge_close_clusters`.

## Question: `_collapse_unary` description growth

When collapsing a chain of single-child nodes, descriptions are concatenated with `"\n\n"`. In a degenerate deeply-nested tree this could produce a very long description string that exceeds LLM context when passed for downstream summarisation. Is there a realistic maximum tree depth that keeps this safe? If depth can be unbounded, a truncation guard would be prudent.

## Tests

The existing `test_ahc_clustering_method_is_supported_in_original_tree_builder` test exercises `_get_clusters_ahc` but only checks `len(labels) == 4`. No tests cover the new `_merge_close_clusters`, the refactored `_adjust_tree_nodes` (cosine path), `_collapse_unary`, or the self-loop guard. Adding coverage — even a few simple cases — would help catch regressions.

## Minor: import style

The `import umap` was moved inside the GMM branch, which is fine (avoids importing a heavy dependency when using AHC). The downside is that an `ImportError` would only surface at runtime inside clustering rather than at import time. Not a blocker, just noting it.

PR #17288, opened 2026-07-23
fix: support Infinity knowledge compilation

buua436:

Fix Infinity compatibility issues in knowledge compilation.

This change:

- Stores compilation source ID lists as JSON arrays in Infinity.
- Parses JSON array fields when reading compiled documents.
- Uses `json_contains` for filtering JSON array fields.
- Adds the missing `name` column to the Infinity mapping.
- Updates dataset navigation KNN search to use the unified `MatchDenseExpr` interface.
- Handles unavailable embeddings without querying an invalid `q_0_vec` field.

### Type of change


wangq8 (commented):
Thanks for this PR. The approach of using native JSON arrays in Infinity for compilation ID lists is the right direction. Here are my observations:

**What works well:**
- JSON array serialization in `insert()` / `update()` and deserialization in `get_fields()` is cleanly implemented.
- The backward-compatible `parse_json_list` fallback (trying `json.loads`, then `###`-splitting for old varchar rows) is a nice touch.
- Guarding against unavailable embeddings in `upsert_dataset_nav_doc` avoids a real crash path — good fix.
- Using `MatchDenseExpr` directly instead of the try/except fallback is cleaner and removes an error-silencing anti-pattern.

**Concerns / suggestions:**

1. **Migration path for existing Infinity instances**: The schema changes (3 columns from `varchar` to `json` + new `name` column) will break existing Infinity tables that already have data. Users will get errors on the next insert/search against those columns. At minimum this needs a doc note; ideally a migration step that alters the column types or a versioned mapping file.

2. **`name` column placement**: The new `name` column in `infinity_mapping.json` is added between `docnm` and `name_kwd` with analyzer `whitespace-#`. But `name_kwd` appears to be the keyword/tokenized variant of `name`. Are there any code paths that write to `name`? If `name` values are always routed through `name_kwd` today, this column might be unused. Worth double-checking.

3. **Function-level imports**: `MatchDenseExpr` and `OrderByExpr` are imported inside `_store_knn` rather than at module level. If there is no circular import issue, moving them to the top of the module would be preferred.

No blocking issues — LGTM overall.

conf/infinity_mapping.json
```diff
@@ -43,9 +44,9 @@
 	"extra": {"type": "varchar", "default": ""},
 
 	"compile_kwd": {"type": "varchar", "default": "", "analyzer": "whitespace-#"},
-	"source_chunk_ids": {"type": "varchar", "default": "", "analyzer": "whitespace-#"},
-	"source_doc_ids": {"type": "varchar", "default": "", "analyzer": "whitespace-#"},
-	"compilation_template_ids": {"type": "varchar", "default": "", "analyzer": "whitespace-#"},
+	"source_chunk_ids": {"type": "json", "default": "[]"},
```
qinling0210: It is better that we create secondary index (low cardinality) on the json column.

Refer to 
	"available_int": {"type": "integer", "default": 1, "index_type": {"type": "secondary", "cardinality": "low"}},

PR #17291, opened 2026-07-23
fix: include full conversation details in agent log CSV export

euvre:
### Summary

The agent log CSV export only included page-level summary columns (ID, user ID, title, state, rounds, dates, version). It did not include the full conversation details (all message turns) for each session, even though the backend API already returns the complete `message` array.

This PR adds a **Conversation detail** column to the CSV export that includes the complete message history (all turns, formatted by role) for each session.

**Changes:**
- `web/src/pages/agents/hooks/use-export-agent-log.ts`: Added a `formatConversation` function that iterates over all messages in a session and formats them as `Role: content` lines. Added the new column to both the CSV headers and row data.
- `web/src/locales/en.ts`: Added i18n keys `conversationDetail`, `user`, `assistant`.
- `web/src/locales/zh.ts`: Added corresponding Chinese translations.


wangq8 (commented):
LGTM. Clean feature addition.

**What it does:** Adds a "Conversation detail" column to the agent log CSV export, containing the full message history (all turns, formatted as `Role: content` per line).

**Code quality:**

- `formatConversation` correctly handles the empty-message case (returns empty string).
- CSV escaping is already in place via the existing `String(cell).replace(/"/g, '""')` wrapping - newlines inside the conversation cell are properly handled by the quoted CSV field.
- i18n keys are added for both `en` and `zh` locales.

**Minor consideration (not a blocker):** For very long conversations, the CSV cell could become quite large. If this becomes an issue in practice, a future enhancement could add a configurable message cap or truncation. Fine for now.

dcc123456 (approved):

PR #17292, opened 2026-07-23
fix(models): surface MiniMax rate-limit and API errors instead of generic messages

euvre:
### Summary

MiniMax can return rate-limit and other API errors embedded in a **HTTP 200** response body (via a `base_resp` block with non-zero `status_code`), or as OpenAI-compatible error events in the SSE stream — rather than returning a standard non-200 status. The previous code discarded these payloads:

- **Non-streaming** (`ChatWithMessages`): returned a generic `"no choices in response"` error, hiding the real cause.
- **Streaming** (`ChatStreamlyWithSender`): silently skipped error events (no `choices` array → `return nil`), then reported `"minimax: stream ended before [DONE] or finish_reason"`.
- **Non-200 responses**: returned the raw body without extracting the human-readable error message.

These generic errors did not contain keywords like `"rate limit"` or `"429"`, so the upstream retry predicates (`RetryPredicates.TemporaryErrors` / `ResourceErrors` in `internal/harness/graph/pregel/retry.go`) could not match them and **never retried** on rate limits.

### What this PR changes

Adds two helpers in `internal/entity/models/minimax.go`:

- `extractMinimaxAPIError()` — extracts the real error message from both MiniMax-specific (`base_resp.status_msg`) and OpenAI-compatible (`error.message`) payloads.
- `extractMinimaxErrorBody()` — parses a raw HTTP error response body and returns the human-readable message.

All three code paths now surface the actual error:

1. **Non-streaming chat**: checks for embedded `base_resp`/`error` *before* checking `choices`.
2. **Streaming chat**: returns an error when an SSE event carries `base_resp`/`error` instead of silently skipping it.
3. **Non-200 HTTP**: all endpoints (`ChatWithMessages`, `ChatStreamlyWithSender`, `ListModels`) now use `extractMinimaxErrorBody` to surface the real message.

This enables the retry layer to match rate-limit errors by keyword (`"rate limit"`, `"429"`, `"too many requests"`, `"quota exceeded"`) and retry appropriately.

### Tests

Added three tests in `internal/entity/models/minimax_test.go`:

- `TestMinimaxChatSurfacesBaseRespError` — HTTP 200 with `base_resp` rate-limit payload
- `TestMinimaxChatSurfacesOpenAIError` — HTTP 429 with OpenAI-compatible `error` payload
- `TestMinimaxStreamSurfacesBaseRespError` — SSE error event with `base_resp` payload

All existing MiniMax tests continue to pass.


wangq8 (commented):
Nice improvement - surfacing the real error messages instead of generic ones is clearly the right thing to do.

**What works well:**

- `extractMinimaxAPIError` correctly handles both MiniMax-specific (`base_resp`) and OpenAI-compatible (`error`) error payloads, in both 200 and non-200 responses.
- All three code paths (non-streaming, streaming, `ListModels`) now surface the actual error.
- Tests cover the key scenarios: `base_resp` in 200, OpenAI-style `error` in 429, and SSE error events.
- The streaming fix (returning an error when SSE carries `base_resp`/`error` instead of silently skipping) is particularly important - silent drops are the hardest to debug.

**One thing to be aware of (not a blocker):**

The `contains` helper in `internal/harness/graph/pregel/retry.go` only checks the start of the error string (both the case-sensitive and case-insensitive paths do `s[:len(substr)] == substr`). Since the errors now have a `"minimax API error: "` prefix, keywords like `"rate limit"` or `"429"` won't match at position 0 and the retry predicates won't fire. The error surfacing is still a win for observability, but if the goal is automatic retry, the `contains` function might need to become a proper substring scan. Something to follow up separately.

Lynn-Inf (approved):
LGTM

PR #17295, opened 2026-07-23
fix: agent session round count not incremented in Go mode

euvre:
### Summary

In Go mode, the agent session round count (`api_4_conversation.round`) was always `0` because `persistAgentRunSession` in `internal/service/agent.go` never incremented `session.Round` when persisting a completed turn.

Python's `API4ConversationService.append_message` (`api/db/services/api_service.py:107`) performs `round=round+1` on each message append. The Go port was missing this increment, so the session list always showed 0 rounds regardless of how many turns had occurred.

**Fix:** Add `session.Round++` before the `Update` call in `persistAgentRunSession`, mirroring the Python behavior.


wangq8 (commented):
LGTM. Simple and correct fix.

The `session.Round++` mirrors the Python `API4ConversationService.append_message` behavior (`api/db/services/api_service.py:107`) where `round=round+1` is performed on each message append. The placement is correct — incrementing before the `Update` call in `persistAgentRunSession` ensures the round count is persisted for every completed turn.

One small thing to double-check: the diff shows `s.api4ConversationDAO.Update(ctx, session)` but the current `main` branch code calls `s.api4ConversationDAO.Update(session)` (without `ctx`). Make sure this merges cleanly against the target branch.

PR #17304, opened 2026-07-23
Fix: update base_url and model list for provider tencent Hunyuan

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the PR! Updating the Hunyuan provider for the new MaaS platform is needed. I have a few concerns:

**1. `CheckConnection` call removed from `CheckConnection` method (`model_service.go`)**
The `driver.CheckConnection(ctx, apiConfig)` call is removed entirely. For Hunyuan, this calls `ListModels` to verify connectivity. Without it, the "Check Connection" feature skips its primary fast-fail connectivity check and only runs per-model verification. If the credentials or base URL are wrong, and the model list happens to be empty, no connection error is surfaced. Was this removal intentional? If `ListModels` doesn't work on the new MaaS endpoint, I'd suggest adding a comment explaining why. Otherwise, please restore it — adding `BaseURL` to the config should make it work with the new endpoint.

**2. Old model names removed (breaking change)**
All previous Hunyuan models (`hunyuan-pro`, `hunyuan-standard`, `hunyuan-standard-256K`, `hunyuan-lite`, `hunyuan-embedding`) are deleted. Existing users with these models configured will lose them. If these models are no longer available on the MaaS platform, could you add a note in the PR description? If they are available under different names, consider keeping them as aliases or documenting the migration path.

**3. Image/video/3D model types**
Models like `hy-image-v3.0`, `hy-video-1.5`, `kl-video-*`, `vd-video-*`, `yt-video-*`, and `hy-3d-*` are all typed as `["chat"]`. If these generate images/videos/3D assets rather than being multimodal chat models, they should use appropriate model types. Please confirm the MaaS platform exposes these through the chat/completions endpoint as chat-type responses.

**4. `url_suffix` not updated**
The `url_suffix` only has `chat`, `models`, and `embedding` endpoints. No image/video/asr/tts suffixes were added. If the new models use different API paths, this may need updating.

PR #17306, opened 2026-07-23
fix: resolve layout_recognize tenant_model.id to OCR parser kind

huang-aoqin:
## Summary
- Frontend stores OCR/vision `tenant_model.id` (32-char hex) in `parser_config.layout_recognize`.
- Backend previously only normalized static names / `@mineru` suffixes, so hex ids fell through to the vision path and failed.
- Resolve hex ids via DB to MinerU/PaddleOCR/OpenDataLoader/SoMark (or keep vision model id), and route `resolve_model_config` hex refs to `get_model_config_by_id` only.

## Test plan
- [ ] Select MinerU OCR model in dataset parser settings, parse a PDF
- [ ] Confirm no `TenantModel id=... cannot be used as vision model` / `Provider not found`
- [ ] Run `test/unit_test/common/test_resolve_layout_recognizer.py`


xugangqiang:
## Review findings (read-only)

This PR fixes a real bug (resolving `tenant_model.id` hex strings stored in `parser_config.layout_recognize` to OCR parser kinds), and the core fix plus its tests are correct. However, the PR also bundles changes that should not merge as-is.

### 1. Blocker: `_get_provider_or_sync_legacy` references a module/function that does not exist

`api/db/joint_services/tenant_model_service.py:267-269` imports and calls `sync_tenant_llm_factory_if_exists` from `api.db.joint_services.tenant_llm_sync_service`. That module and function are **not defined anywhere** in the repo (confirmed absent on both `main` and this PR's head, and not among the 11 changed files).

Consequence: `_get_provider_or_sync_legacy` is now the provider-lookup entry point for `get_model_config_from_provider_instance` (`:363`), `get_api_key` (`:543`), and `get_model_type_by_name` (`:566`). The happy path returns early at `:264-265` and never hits the import, so normal resolution still works — but the moment a provider is **not** found (a legitimate error case, e.g. a mistyped provider name), the code attempts the sync branch and raises `ImportError` instead of the previous clean `LookupError("Provider ... not found")`. This is a regression in error handling and means the intended "legacy provider sync" can never function.

### 2. Scope creep: unrelated, untested model-resolution refactor

The following additions are not part of the stated fix and have **zero test coverage** (the only new test covers `is_tenant_model_id` / `resolve_layout_recognizer`):

- `_get_model_config_from_tenant_llm` (`:238-259`)
- `_get_provider_or_sync_legacy` (`:262-271`, also the broken import above)
- The expanded `else` branch in `get_model_config_from_provider_instance` (`:402-422`, factory lookup + tenant_llm fallback) — note this branch is **never reached** by the hex-id path, because `resolve_model_config` short-circuits hex ids to `get_model_config_by_id` (`:335-337`).
- The rewritten `get_model_type_by_name` (`:564-582`).

These sit on hot model-resolution read paths for **all** models. They appear to be a separate "self-added model / legacy provider sync" feature that should be reviewed and tested on its own.

### 3. Nit: silent fallback when `tenant_id` is missing

`resolve_layout_recognizer(tenant_id, ...)` returns `normalize_layout_recognizer(...)` whenever `tenant_id` is falsy. Most callers pass `kwargs.get("tenant_id")`, so if a caller fails to thread `tenant_id` through, the hex id is silently dropped and the original bug recurs without any signal. A `logger.warning` on that path would help.

---

### Suggestion: split this PR

- **Keep in this PR** (correct, needed, tested): `resolve_layout_recognizer` hex-id resolution, the `is_tenant_model_id` short-circuit in `resolve_model_config`, and `test_resolve_layout_recognizer.py`.
- **Move out to a separate PR** (with its own tests): `_get_model_config_from_tenant_llm`, `_get_provider_or_sync_legacy`, the expanded `else` branch in `get_model_config_from_provider_instance`, and the `get_model_type_by_name` rewrite.
- For that separate PR, either add the missing `tenant_llm_sync_service.sync_tenant_llm_factory_if_exists` module or remove the sync branch entirely.

Note: the cross-tenant IDOR flagged earlier (commit `150d92c`) is correctly addressed and is not repeated here.

PR #17307, opened 2026-07-23
fix: convert MinerU extract_positions to 1-based pages in manual parser

huang-aoqin:
## Summary
- `MinerUParser.extract_positions` returns 0-based page indices, while DeepDOC `tag()` / `crop()` expect 1-based pages.
- Manual chunking used the 0-based values directly, causing position misalignment for MinerU-parsed manuals.
- Also harden empty `poss` sorting and raise when a parser returns `sections is None`.

## Test plan
- [ ] Parse a multi-page PDF with Manual + MinerU
- [ ] Verify chunk positions / crops align with page content
- [ ] Confirm empty-position sections do not crash sorting


xugangqiang:
## Review findings (read-only)

### Blocker — does not import (must fix before merge)
`rag/app/manual.py:30` and `:149` import and call `resolve_layout_recognizer`, but this symbol does not exist anywhere in the repo (neither in this branch nor on `upstream/main`). The only function present is `normalize_layout_recognizer(layout_recognizer_raw)` in `common/parser_config_utils.py:20` (single-arg). As written, `from common.parser_config_utils import resolve_layout_recognizer` raises `ImportError`, so the module fails to load and every manual parse path is broken.

Fix: revert to the existing single-arg call:
```python
from common.parser_config_utils import normalize_layout_recognizer
...
layout_recognizer, parser_model_name = normalize_layout_recognizer(parser_config.get("layout_recognize", "DeepDOC"))
```
The added `kwargs.get("tenant_id")` is unrelated to this bug and is not in the PR description — drop it.

### Improvement — add tests / verification
The PR has no tests, and `rag/app/manual.py` has no existing tests. The page-base change flips manual chunk position tags from 0-based to 1-based for all parsers whose sections return a string `poss` (mineru/docling/paddleocr/somark/opendataloader). Please add a minimal unit test for `_normalize_section` covering: 0-based→1-based conversion, multi-position strings, empty `poss`, and DeepDOC list-typed `poss` (which must NOT be converted), plus a real manual+MinerU crop check on a multi-page PDF.

### Improvement — log before the new RuntimeError
`rag/app/manual.py:177-178` raises `RuntimeError` when a parser returns `sections is None` with no context. Add `logging.error("parser %s returned no sections for %s", name, filename)` before the raise so worker failures stay diagnosable.

### Nit — scope/description
The title/description say "MinerU" but the +1 conversion actually applies to every parser whose `poss` is a string (mineru/docling/paddleocr/somark/opendataloader); DeepDOC is unaffected because `get_position` returns a list, not a string. Worth stating the real scope.

### Nit — defensive negative page indices (optional)
In the new normalization loop, skip pages `< 0` (mirror the `0 <= p < page_count` guard in `PdfParser.crop`) to avoid out-of-range indices from malformed input.

### Note on existing bot review
The CodeRabbit nit at lines 149-152 asks to "log after `resolve_layout_recognizer`" — it assumes that symbol exists. It does not; the real problem is the missing import above, not missing logging. The line 188-200 finding (preserve empty-position PlainParser sections) is valid and was addressed by the second commit.

JinHai-CN:
@huang-aoqin would you please fix the conflicts?

huang-aoqin:
> @huang-aoqin would you please fix the conflicts?

@JinHai-CN Done

xugangqiang:
## Review findings (read-only)

### Blocker — page-base conversion is misplaced and incomplete
All string-based parsers' `extract_positions` return **0-based** page indices (`int(p) - 1`): `mineru_parser.py:564`, `docling_parser.py:163`, `paddleocr_parser.py:766`, `somark_parser.py:459`, `opendataloader_parser.py:395`, `monkeyocrv2_parser.py:46`, `mistral_parser.py:190`.

`tag()` adds `+1` **only for `name == "mineru"`** (`manual.py:260-263`) in both base and head. So:
- **MinerU** is already 1-based in `main` via `tag()`; this PR changes nothing for it. The PR premise ("manual used 0-based for MinerU") is incorrect.
- **docling / paddleocr / somark / opendataloader / monkeyocrv2 / mistral** still emit **0-based** `@@` tags, but `crop()` expects 1-based (`pdf_parser.py:2139` uses `pdf_page_num + 1`). This is the real misalignment — and it is left unfixed.

The `+1` from the first commit (`a69ba008`) was **dropped during the merge conflict resolution** (`a4679f88`), leaving the normalization as `int(pn)` with no `+1` (`manual.py:208`).

**Fix:** add `+1` during normalization for *all* string-based positions, and remove the `if name == "mineru": pn += 1` special-case in `tag()` to avoid double-counting for MinerU. DeepDOC list-typed `poss` is unaffected (it is not a `str` and is already 1-based).

### Improvement — rebase onto current `upstream/main`
`git merge-base HEAD upstream/main` is empty: the branch was forked from an old fork `main`. Please rebase onto the current `upstream/main` to avoid stale-code conflicts.

### Improvement — add tests
No tests in this PR, and `rag/app/manual.py` has none. Please add a minimal unit test for `_normalize_section` covering: 0-based→1-based conversion, multi-position strings, empty `poss`, and DeepDOC list-typed `poss` (must NOT be converted).

### Nit — misleading title/comment
The title/description claim a MinerU-only fix; the actual bug affects all non-MinerU string parsers. The code comment at `manual.py:197-199` ("tag() adds +1 for MinerU") describes the flawed current design and should be rewritten once the fix above lands.

### Nit — log before the new RuntimeError
`manual.py:183-184` raises `RuntimeError` when a parser returns `sections is None` with no context. Add `logging.error("parser %s returned no sections for %s", name, filename)` before the raise.

(Note: the earlier "Blocker — `resolve_layout_recognizer` does not import" review is stale; the current head already uses `normalize_layout_recognizer`, resolved by the merge. The `tenant_id` logic is pre-existing in `main`, not added here.)

PR #17310, opened 2026-07-23
fix(web): prevent multi-select popover from closing on first selection

dcc123456:
### Summary

fix(web): prevent multi-select popover from closing on first selection


wangq8 (commented):
Thanks for the fix! A small but meaningful improvement to the multi-select UX.

**What this fixes:** The multi-select popover was closing prematurely on the first selection because clicking a `CommandItem` can trigger a focus-outside event. Adding `onFocusOutside` with `preventDefault()` prevents that dismiss path while still allowing the popover to close via Escape key (already handled by `onEscapeKeyDown`) or by clicking the trigger button.

**A few observations:**

1. Minimal and safe — single line, no side effects on the render or data flow.
2. Escape still works — `onEscapeKeyDown` is already wired to close the popover, so keyboard users aren't trapped.
3. The `Popover` wrapper has a `disableOutsideClick` prop, but that would block all outside-click dismissals indiscriminately. The `onFocusOutside` approach is more targeted and appropriate here.

**One thing to keep in mind:** Users can still dismiss the popover by clicking outside (pointer events aren't affected by this change), which is the expected behavior — you want to prevent focus-driven dismiss but still allow intentional click-away.

LGTM ✅

PR #17316, opened 2026-07-23
fix(deepdoc): harden opendataloader table-cell parsing against bad json

Yurii214:
### Summary

`_element_html` and `_element_text` in `deepdoc/parser/opendataloader_parser.py` rebuild tables from JSON returned by the external opendataloader/Docling service, but read several cell fields without coercion. A single malformed cell raises and aborts the whole document parse:

`_element_html`
- `text: null` or a non-string `text` → `AttributeError` in `html.escape(...)`
- `start_row_offset_idx: null` (key present, value null) → `TypeError` in `0 <= r < num_rows` (`.get(k, 0)` only defaults a *missing* key, not a null value)
- string `num_rows`/`num_cols` → `TypeError` on `num_rows > 0`

`_element_text`
- a non-numeric `row` label → `ValueError` in `int(row)`
- `row: 0` with a `row_index` present → the `row or row_index or 0` chain treats the real row 0 as missing and files the cell under `row_index`, mis-ordering rows

This coerces the numeric fields (skipping a bad cell / emptying the table rather than crashing) and `str()`-coerces text before escaping — mirroring the `or 1` guards already used on `row_span`/`col_span`, and the `str(... or "")` coercion `_element_text` already applies to `content`. Well-formed tables are unchanged.

### Tests

Regression tests added to `TestElementHtml` and a new `TestElementText` in `test/unit_test/deepdoc/parser/test_opendataloader_parser.py`. Each fails without the change and passes with it; the existing 34 tests are unaffected.

### DIFF
--- a/deepdoc/parser/opendataloader_parser.py
+++ b/deepdoc/parser/opendataloader_parser.py
@@ -108,8 +108,16 @@ def _element_text(el: dict) -> str:
         for c in cells:
             if not isinstance(c, dict):
                 continue
-            row = c.get("row") or c.get("row_index") or 0
-            rows.setdefault(int(row), []).append(str(c.get("content") or c.get("text") or ""))
+            raw_row = c.get("row")
+            if raw_row is None:
+                raw_row = c.get("row_index")
+            try:
+                # a non-numeric row label must not abort the whole parse; also treat
+                # row 0 as a real index rather than "missing" (the `or` chain did not).
+                row_idx = int(raw_row) if raw_row is not None else 0
+            except (TypeError, ValueError):
+                row_idx = 0
+            rows.setdefault(row_idx, []).append(str(c.get("content") or c.get("text") or ""))
         return "\n".join(" | ".join(v) for _, v in sorted(rows.items()))
     return ""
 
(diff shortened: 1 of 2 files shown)


Lynn-Inf (commented):
Python changes LGTM. The Go parser `internal/parser/parser/pdf_parser_opendataloader.go` (`openDataLoaderCellsText`, ~L186) has the same `row == 0` issue this PR fixes in `_element_text` — if row == 0 { row = row_index } mis-orders a real row 0. Would be 
  good to fix that here too so the Python and Go paths stay consistent.

xugangqiang:
## Review feedback (non-blocking)

The change correctly fixes every crash path described in the PR body, and well-formed tables are unchanged. Logic verified against `deepdoc/parser/opendataloader_parser.py` (`_element_text` / `_element_html`, called from `_transfer_from_json` at line 480). The existing unit tests fail without the change and pass with it, and Codecov confirms the changed lines are covered. Low regression risk — the only semantic change (`row: 0` no longer swallowed by `or`) actually matches the original "row is authoritative" intent.

### Coverage gaps worth closing before merge

The added tests cover the happy-path crash cases, but the core "skip the bad cell, keep the table" behavior is not fully proven. Suggested additions:

1. `_element_html` with a **non-numeric string offset** (e.g. `"abc"`) → should `continue` and skip that cell. Only `null` offset is tested today, not a string offset.
2. `_element_html` with a **mix of a good cell and a malformed cell** (e.g. r0c0 valid, r0c1 malformed) → proves "skip the bad cell, render the rest" rather than aborting. This is the central semantic of the PR and deserves a mixed-case test.
3. `_element_html` with **non-coercible `num_rows`/`num_cols`** (e.g. a list) → should return `""` instead of crashing. Only the string-typed case is tested.
4. `_element_text` with **`row` missing, falling back to `row_index`** (only "non-numeric row" and "row = 0" are tested).
5. `row_span`/`col_span` as **string or null** (logic already does `int(... or 1)` but is untested).
6. (Optional, larger scope) An integration-level test at `_transfer_from_json` / `_normalize_docling_response` proving a whole Docling document with a malformed table does not abort the entire parse.

### Minor hardening suggestion

Silently skipping a malformed cell (or emptying the table when `num_rows`/`num_cols` are bad) hides upstream schema drift. Consider a rate-limited `logger.warning` when a cell is dropped or the table is emptied, so data-quality regressions from the external service stay observable. A small shared `_coerce_int(value, default=0)` helper would also remove the duplicated try/except blocks.

---

@Yurii214 — would you prefer to **merge as-is** (the fix is correct and low-risk), or **add the coverage gaps above (1–5, at least 1, 2, 3) before merging**? Happy to adjust depending on your preference.

xugangqiang:
LGTM @yuzhichang @JinHai-CN  Let's wait for the response --- "@Yurii214 — would you prefer to merge as-is (the fix is correct and low-risk), or add the coverage gaps above (1–5, at least 1, 2, 3) before merging? Happy to adjust depending on your preference."

PR #17318, opened 2026-07-23
fix: escape heading text before using it as a regex in remove_contents_table

rsnetworkinginc:

`remove_contents_table` (`rag/nlp/__init__.py`) removes a table-of-contents
block by matching the heading that follows a TOC marker (`目录`, `contents`,
`table of contents`, …). It derives a `prefix` from that heading — the first
3 characters for CJK text, or the first 2 words for English — and then uses it
to find where the real content resumes:

```python
prefix = get(i)[:3] if not eng else " ".join(get(i).split()[:2])
...
for j in range(i, min(i + 128, len(sections))):
    if not re.match(prefix, get(j)):   # prefix is document text, used as a regex
        continue
```

`prefix` is plain document text meant to be matched **literally**, but it is
passed straight to `re.match()` as a *pattern*. Any regex metacharacter in it
is therefore interpreted as regex syntax. This is very easy to hit with CJK
headings because the 3-character slice frequently cuts a parenthesis in half:
a heading such as `1.(1) 引言` produces the prefix `1.(`, and

```python
re.match("1.(", ...)
```

raises `re.error: missing ), unterminated subpattern`. The exception is not
caught, so it aborts the entire parse. `remove_contents_table` runs from
`rag/app/book.py` and `rag/app/laws.py` (and `rag/flow/parser/utils.py`), so
any TOC-bearing book/laws document whose first post-TOC heading begins with an
unbalanced `(`, `[`, `*`, `+`, `?`, … crashes ingestion.

**Fix:** escape the prefix with `re.escape()` before matching, so it is
compared literally. This preserves the original start-anchored `re.match`
semantics (the block up to the section that literally starts with `prefix` is
still removed) while removing the crash.

```python
if not re.match(re.escape(prefix), get(j)):
```

This is a one-line, behavior-preserving fix; it is independent of the
in-flight TOC-heading *detection* work in #17012, which changes a different
line of the same function.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Testing done

Added `test/unit_test/rag/test_remove_contents_table_regex_prefix.py`:

- `test_metachar_prefix_does_not_crash_chinese` — reproduces the original
  `re.error` crash (`目录` followed by `1.(1) 引言`) and asserts the TOC block
  is removed instead. Fails on `main` with
  `re.PatternError: missing ), unterminated subpattern`; passes with this fix.
- `test_metachar_prefix_does_not_crash_english` — English metacharacter prefix.
- `test_literal_prefix_still_removes_toc_block` — a prefix containing `(`
  (`第(1`) still strips the block up to the real chapter via literal matching,
  proving the fix keeps the intended behavior.
- `test_non_toc_sections_are_left_untouched` — documents without a TOC marker
  are unchanged.

```
$ python -m pytest test/unit_test/rag/test_remove_contents_table_regex_prefix.py -q
....                                                                     [100%]
4 passed
```

Lint (matches the repo's `lefthook` ruff gate):

```
$ ruff check rag/nlp/__init__.py test/unit_test/rag/test_remove_contents_table_regex_prefix.py
All checks passed!
$ ruff format --check rag/nlp/__init__.py test/unit_test/rag/test_remove_contents_table_regex_prefix.py
2 files already formatted
```


xugangqiang (commented):
Thanks for the fix — the root-cause analysis is correct and the crash is real. Two issues found during review:

## 1. Prefer a literal prefix match over `re.escape` (and align with the Go port)
The fix works, but the Go port of this exact function already avoids the regex footgun entirely: `internal/parser/parser/text_toc.go` `removeContentsTable` matches with `strings.HasPrefix(itemText(items[j]), prefix)` (literal), so it was never vulnerable to this bug.

Since `re.match(re.escape(prefix), s)` is semantically identical to `s.startswith(prefix)` here (no `IGNORECASE`, anchored at start), the cleaner and parity-consistent change is to drop the regex altogether:

```python
if not get(j).startswith(prefix):
```

This removes the regex machinery, is more readable, and keeps the Python path aligned with the Go reference implementation.

## 2. `test_metachar_prefix_does_not_crash_english` does not actually reproduce a crash
The English case uses the heading `[1] Overview`, whose prefix `[1] Overview` is a *valid* regex (`[1]` is a legal character class). On `main`, `re.match("[1] Overview", ...)` does **not** raise, so this test passes both before and after the fix — it is not a real regression test for the crash.

To actually exercise the English crash path, use a prefix that contains a real metacharacter, e.g. a heading like `C++ Basics` (`++` → `re.error: multiple repeat`) or `A (note)` (`A (` → unterminated group). Consider updating that test to use such an input.

(Note: the CodeRabbit `sys.modules` comment was valid for the earlier commit and is already addressed in 2e618b9 via `mock.patch.dict`; nothing further needed there.)

PR #17325, opened 2026-07-23
fix: apply order_by return value in deep-pagination helpers (unordered OFFSET/LIMIT can skip/dupe rows)

rsnetworkinginc:
### Summary

The fix is one line per helper — assign the `order_by(...)` result back so the
ordering is actually applied to the query.


Lynn-Inf (approved):
LGTM

PR #17326, opened 2026-07-23
fix: decode non-UTF-8 text files via find_codec in deepdoc get_text

rsnetworkinginc:
### Summary

`deepdoc/parser/utils.py::get_text()` decodes bytes safely when a document is
passed in as `binary` (it detects the encoding with `find_codec()` and decodes
with `errors="ignore"`), but the file-path branch instead did:

```python
with open(fnm, "r") as f:
    ...
```

`open(fnm, "r")` uses the platform-default text encoding (`locale.getpreferredencoding()`)
and performs no error handling. So any non-UTF-8 text file read from disk — e.g. a
GBK-encoded Chinese `.txt`/`.csv` or a Latin-1 file — raises `UnicodeDecodeError` and
aborts parsing, even though the exact same document parses fine when handed in as bytes.
The two branches disagree on how the same content is decoded.

`get_text()` is used from `deepdoc/parser/txt_parser.py` and the `rag/app/*`
parsers (`qa`, `tag`, `book`, `table`, `laws`, `one`, `resume`), all of which can
reach the `binary=None` file-path branch.

#### Fix

Read the file as bytes and route **both** branches through `find_codec()`, so the
encoding is detected the same way regardless of whether the input arrives as a path
or as bytes:

```python
def get_text(fnm: str, binary=None) -> str:
    if binary is None:
        with open(fnm, "rb") as f:
            binary = f.read()
    encoding = find_codec(binary)
    return binary.decode(encoding, errors="ignore")
```

This also collapses the two decoding paths into one, matching the existing binary
behavior (which most callers already exercise). UTF-8 and ASCII files are unaffected;
GBK/Latin-1/other non-UTF-8 files now decode instead of crashing.

#### Tests

Adds `test/unit_test/deepdoc/parser/test_get_text_encoding.py` covering reads of GBK,
Latin-1, UTF-8 and ASCII files from disk, plus the unchanged `binary=` path. The GBK
and Latin-1 cases fail on the previous implementation with `UnicodeDecodeError` and
pass after this change.

#### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement


xugangqiang:
LGTM

xugangqiang:
please fix ci error - is it a regression or test case need to be updated?

JinHai-CN:
please re-open the PR, if you are still working on it @rsnetworkinginc

PR #17327, opened 2026-07-24
fix(agent): set exp_user_id and name on session creation so exploration shows session titles

euvre:
### Summary

The agent exploration sidebar queries `api_4_conversation` by `exp_user_id`, but the Go port never set that column when creating sessions — every row had `NULL`, so `ListAgentSessionNames` returned nothing and no session names appeared in the exploration view.

This mirrors the Python implementation (`api/apps/restful_apis/agent_api.py` lines 494-499 and 1547-1553):

- **`CreateAgentSession`** (`internal/service/agent_sessions.go`): set `exp_user_id = user_id`.
- **`createAgentRunSession`** (`internal/service/agent.go`): set `exp_user_id = user_id` and derive `name` from the first user input (truncated to 250 runes, matching the Python `query[:250]` fallback).
- Added `deriveAgentSessionName` helper to extract the session title from the user input.


wangq8 (commented):
Thanks for this fix! The root cause is clear — `ListAgentSessionNames` queries by `exp_user_id`, but sessions were being created with `NULL` in that column, so the exploration sidebar was always empty.

**What I checked:**

- **`CreateAgentSession`** (`agent_sessions.go`): ✅ correctly adds `ExpUserID: &req.UserID`. `Name` was already being set here from the request.
- **`createAgentRunSession`** (`agent.go`): ✅ correctly adds both `Name` and `ExpUserID`. The session title is derived from user input, mirroring the Python behavior.
- **`deriveAgentSessionName`**: ✅ reuses the existing `stringifyAgentUserInput` helper, which correctly handles `nil`, `string`, and any other type. The 250-rune truncation matches Python's `query[:250]`.
- **Regression risk**: ✅ low. This only populates columns that should have been set but were previously `NULL`/empty. The exploration sidebar will start showing names as intended.

**Minor suggestion (non-blocking):** A quick unit test for `deriveAgentSessionName` covering `nil`, empty string, short string, and a string exceeding 250 runes would be a nice addition.

LGTM, nice catch! 🚀

Hz-186 (approved):
@euvre, LGTM

PR #17334, opened 2026-07-24
fix: honor parser params and image VLM system_prompt in Go ingestion

xugangqiang:
## Summary
Fix the Go ingestion pipeline so that several parser setup switches and the
image VLM prompt are actually honored end-to-end (previously the DSL fields
existed but the Go code never read them).

- **DOCX** (`docx_parser.go`, `docx_postprocess.go`): read `remove_toc` and
  `remove_header_footer`; apply to both JSON and markdown output paths
  (outline-based TOC removal with a text-heuristic fallback, plus
  header/footer section filtering).
- **HTML** (`html_parser.go`, `html_postprocess.go`, `text_toc.go`): read
  `remove_header_footer` (pre-parse strip of `<header>`/`<footer>` and ARIA
  `banner`/`contentinfo`) and `remove_toc` (post-parse `remove_contents_table`
  heuristic).
- **Markdown** (`markdown_parser.go`): read `flatten_media_to_text` and force
  media blocks to text when enabled.
- **Image VLM** (`media_dispatch.go`): read `system_prompt` instead of `prompt`
  so the user-configured image VLM prompt is no longer silently dropped
  (`prompt` remains the video family key).

All flags are wired through `ConfigureFromSetup`, which the dispatch layer
already invokes for every family, so the behavior is live rather than dead code.

## Test plan
- New unit tests: `docx_postprocess_test.go`, `html_parser_test.go`,
  `text_toc_test.go`, `markdown_parser_test.go`, `media_dispatch_test.go`.
- `bash build.sh --test ./internal/parser/parser/... ./internal/ingestion/component/...`

## Notes
- The `File` component is excluded from this migration scope.
- Relates to the Python→Go parity diff (Parser 1.8–1.11, 1.15).

### DIFF
--- a/internal/ingestion/component/media_dispatch.go
+++ b/internal/ingestion/component/media_dispatch.go
@@ -212,7 +212,10 @@ func maybeDispatchImage(
 	dataURI := "data:" + mimeType + ";base64," + imageB64
 
 	prompt := "Describe this image in detail."
-	if v, ok := setup["prompt"].(string); ok && v != "" {
+	// image family's contract key is system_prompt (parser.go:295),
+	// mirroring Python parser.py:1119. Do NOT read setup["prompt"]
+	// here — that key is for the video family, not image.
+	if v, ok := setup["system_prompt"].(string); ok && v != "" {
 		prompt = v
 	}
 	messages := []modelModule.Message{{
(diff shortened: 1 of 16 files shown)


xugangqiang:
@yingfeng @JinHai-CN @yuzhichang pls help to review

yuzhichang:
LGTM

PR #17339, opened 2026-07-24
feat: add GET /datasets/owners aggregation endpoint

euvre:
### Summary

When navigating to the datasets list page, the backend `GET /api/v1/datasets` endpoint was called **twice**:

1. `useFetchNextKnowledgeListByPage()` — fetches the current page for the card grid and pagination (correct).
2. `useSelectOwners()` → `useFetchAllKnowledgeList()` — fetched **all** datasets (auto-paginating through every page) solely to build the "Owner" filter dropdown.

This redundant full-list fetch was wasteful, especially for users with hundreds of datasets.

This PR adds a lightweight `GET /api/v1/datasets/owners` aggregation endpoint that returns distinct owners with their dataset counts directly from the database (`GROUP BY tenant_id`), eliminating the need to fetch full dataset records on the client side.

**Backend changes:**
- `KnowledgebaseService.get_owners()` — `GROUP BY tenant_id` aggregation query returning `tenant_id`, `nickname`, `avatar`, and dataset `count`
- `dataset_api_service.list_dataset_owners()` — service layer wrapper with keywords/parser_id filtering
- `GET /api/v1/datasets/owners` route

**Frontend changes:**
- `useFetchDatasetOwners()` hook with debounced keyword support
- `listDatasetOwners()` service function + `datasetOwners` API constant
- `useSelectOwners()` refactored to use the new endpoint instead of `useFetchAllKnowledgeList`
- `datasets/index.tsx` passes `searchString` to `useSelectOwners`

**After this change, visiting the datasets page makes:**
- `GET /datasets?page=1&page_size=50` — paginated list
- `GET /datasets/owners` — lightweight aggregation (no full dataset records)

### DIFF
--- a/api/apps/restful_apis/dataset_api.py
+++ b/api/apps/restful_apis/dataset_api.py
@@ -381,6 +381,72 @@ def list_datasets(tenant_id):
         return get_error_data_result(message="Internal server error")
 
 
+@manager.route("/datasets/owners", methods=["GET"])  # noqa: F821
+@login_required
+@add_tenant_id_to_kwargs
+def list_dataset_owners(tenant_id):
+    """
+    List distinct dataset owners with their dataset counts.
+    ---
+    tags:
+      - Datasets
+    security:
+      - ApiKeyAuth: []
+    parameters:
+      - in: query
+        name: ext_keywords
+        type: string
+        required: false
+        description: Keyword filter (matched against dataset names).
+      - in: query
+        name: ext_parser_id
+        type: string
+        required: false
+        description: Parser ID filter.
+      - in: header
+        name: Authorization
+        type: string
+        required: true
+        description: Bearer token for authentication.
+    responses:
+      200:
+        description: Successful operation.
+        schema:
+          type: array
+          items:
+            type: object
+            properties:
+              tenant_id:
+                type: string
+              nickname:
+                type: string
+              avatar:
+                type: string
+              count:
+                type: integer
+    """
+    args = {}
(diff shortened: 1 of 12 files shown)


wangq8 (commented):
Thanks for this PR! This is a clean optimization that eliminates the wasteful full-list fetch for the owners dropdown. The approach is well-aligned with the existing codebase patterns.

## ✅ What works well

- **Route ordering**: `/datasets/owners` is placed before `/datasets/<dataset_id>` — correct, avoids Flask route collision.
- **Access control**: Properly reuses `_visibility_and_status_filter` for consistent visibility rules.
- **Pattern consistency**: The `UserService.get_by_ids` + `user_map` enrichment pattern matches the existing `list_datasets` in `dataset_api_service.py` exactly.
- **Debounced keyword support**: Clean frontend hook with 500ms debounce.
- **Clean refactor**: `useSelectOwners` drops the heavyweight `useFetchAllKnowledgeList` dependency cleanly.

## 🔍 Minor observations (non-blocking)

1. **`gcTime: 0`** in `useFetchDatasetOwners` — this means the cache is immediately eligible for GC when the query becomes inactive. For a filter dropdown that the user may toggle quickly, a non-zero `gcTime` (e.g., 5 minutes) would avoid refetching stale-but-still-valid data. That said, this is consistent with the existing `useFetchAllKnowledgeList` pattern, so no change needed here.

2. **No tests** — no new test coverage for the endpoint or the service method, but this matches the repo convention (most API endpoints lack dedicated tests).

Overall LGTM. Nice work!

PR #17340, opened 2026-07-24
refactor(doc_store): require get_scores and create_doc_meta_idx on the connection contract

deadtrickster:
## What

Make `get_scores` and `create_doc_meta_idx` `@abstractmethod` on `DocStoreConnection`, and implement the two that were missing.

## Why

Both methods are called on the connection unconditionally by the runtime:

- `rag/nlp/search.py` calls `self.dataStore.get_scores(res)` to recover KNN similarities.
- `api/db/services/doc_metadata_service.py` calls `create_doc_meta_idx(index_name)` on every metadata write path (3 call sites).

Neither was on the ABC, so a backend could omit one, pass a smoke test, and fail only at runtime with `AttributeError`. That already happened once: issue #14570, `'OSConnection' object has no attribute 'create_doc_meta_idx'`, fixed reactively for a single backend.

The Go side already treats these as contract. `DocEngine` in `internal/engine/engine.go` requires `GetScores` and `CreateMetadataStore`, and every Go backend implements them, compile-enforced. This PR brings the Python contract to the same footing.

## The latent bug this surfaces

Promoting `get_scores` to abstract shows that the Python **Infinity** and **OceanBase** backends never implemented it, even though their Go twins do. The KNN-score recovery path raised `AttributeError` on those engines. This PR adds one implementation per base class:

- Infinity: map each id to the `_score` column that `search()` already writes on the result frame.
- OceanBase: map each result row's id to its `_score`.

Elasticsearch and OpenSearch already implement both methods and are unchanged. All six concrete backend variants (ES, OpenSearch, Infinity x2, OceanBase x2) satisfy the full abstract set.

## Test

`test/unit_test/common/test_doc_store_contract.py` pins the contract: both methods are abstract, a backend missing either cannot be constructed (`TypeError`), and a complete one can. Verified red before this change (the full-contract case fails with the exact `AttributeError: object has no attribute 'get_scores'`) and green after.

### DIFF
--- a/common/doc_store/doc_store_base.py
+++ b/common/doc_store/doc_store_base.py
@@ -176,6 +176,14 @@ class DocStoreConnection(ABC):
         raise NotImplementedError("Not implemented")
 
     @abstractmethod
+    def create_doc_meta_idx(self, index_name: str):
+        """
+        Create the per-tenant document metadata index. Called by the
+        document metadata service on every metadata write path.
+        """
+        raise NotImplementedError("Not implemented")
+
+    @abstractmethod
     def delete_idx(self, index_name: str, dataset_id: str):
         """
         Delete an index with given name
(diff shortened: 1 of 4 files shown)


deadtrickster:
Closing this. On reflection it is not needed for the work it was meant to lead into: the backends that matter already implement both methods, and the Go `DocEngine` interface already enforces the equivalents (`GetScores`, `CreateMetadataStore`), so the Python contract change buys little on its own.

One finding here is still real and worth keeping separately: the Python Infinity and OceanBase backends never implemented `get_scores`, which the retriever calls unconditionally. If useful I can send that as a small standalone fix.

PR #17345, opened 2026-07-24
fix: stabilize knowledge compilation navigation updates

buua436:

This PR fixes several issues in knowledge compilation navigation, especially stale navigation nodes after document reparse, incorrect PageIndex navigation generation, Infinity compatibility problems, and frontend rendering of empty navigation branches.

### Detailed fixes

- Fix stale parent clusters after document deletion or reparse.
  - Remove the document's `nav_doc` before deleting all rows with the document's `doc_id`.
  - Update the parent cluster's `doc_ids_kwd` and `doc_count_int`.
  - Delete empty clusters and recursively clean up empty ancestors.
  - Apply the same cleanup to normal document deletion, document reparse, and pipeline rerun paths.
  - This prevents parent clusters from remaining after their child document has been deleted.

- Serialize navigation updates safely.
  - Perform existing `nav_doc` lookup and replacement while holding the dataset navigation lock.
  - Prevent concurrent document updates from racing during delete, cluster placement, and parent updates.
  - Add visited-node protection to prevent recursive navigation traversal from entering self-referencing loops.

- Improve Infinity navigation queries.
  - Preserve and validate `compile_kwd` in KNN result fields.
  - Add exact matching for `compile_kwd`, `type_kwd`, and `parent_kwd` conditions.
  - Filter and re-rank fallback KNN results after scanning to ensure only navigation rows are selected.
  - Handle empty list values safely during Infinity updates.
  - Run blocking navigation delete/search operations through the thread pool from async APIs.

- Fix PageIndex navigation generation.
  - Resolve PageIndex structures using the `page_index` compilation keyword.
  - Keep a fallback to the legacy `timeline` keyword for existing data generated by older versions.
  - Pass the embedding and chat models into tree and compiler navigation generation so documents can be embedded and clustered correctly.

- Fix frontend navigation tree rendering.
  - Preserve `has_children` from the backend.
  - Render expandable navigation nodes even when the current children array is empty.
  - Avoid treating an expandable cluster as a document just because its children have not been loaded yet.

### Type of change

- [x] Refactor (internal concurrency and storage handling)


wangq8 (commented):
Thanks for this comprehensive fix! I reviewed the diff carefully and the changes look solid overall. A few observations:

## ✅ Good changes

1. **Lock serialization** in `upsert_dataset_nav_doc` — moving the existing-doc check and removal under the same lock that handles placement is the right fix for concurrent update races. Embedding is correctly done outside the lock.

2. **`visited_names` protection** in `_find_best_cluster` — prevents infinite recursion when cluster parent references form a cycle. Clean and minimal.

3. **`_remove_dataset_nav_doc_locked` extraction** — reduces duplication between the async `remove_dataset_nav_doc` and the inline removal inside `upsert_dataset_nav_doc`. Both callers hold the lock as documented.

4. **PageIndex keyword fix** in `runner.py` — trying `page_index` first, falling back to `timeline` for legacy data is the right approach for backward compatibility.

5. **Empty list handling** in `infinity_conn.py` `update()` — the Infinity Python client indeed crashes on empty lists during value inspection. Converting to empty JSON/empty string for known field types is correct.

6. **`_store_knn` re-ranking fallback** — some Infinity deployments return vector results that don't respect filters well. Falling back to a full scan + filter + rerank when mismatches are detected is a reasonable safety net. The `_matches_condition` helper is focused on the simple equality filters used by nav queries.

7. **Exact matching for `compile_kwd`, `type_kwd`, `parent_kwd`** in `infinity_conn_base.py` — these keyword fields need `=` comparison, not `json_contains`, and this is also more efficient.

8. **Frontend `hasChildren`** — differentiating "no children loaded yet" from "leaf node" is the correct fix for expandable clusters showing as documents.

9. **`remove_dataset_nav_doc_sync` before chunk deletion** in `document_service.py` — the nav_doc row needs to exist for the parent cluster lookup, so it must be processed before the broad `doc_id` delete.

10. **Passing `embd_mdl`/`chat_mdl`** in `compiler.py` and `chunk_post_processor.py` — these call sites were missing the model parameters needed for doc embedding and cluster summary generation.

## ⚠️ Minor suggestions (non-blocking)

- **`_matches_condition` string comparison**: comparing via `str(value) == str(item)` is a bit loose but works for the controlled nav field values. Not a blocker.

- **`_store_knn` fallback threshold**: the fallback triggers on *any* mismatch in the KNN results. For very large navigation graphs this may add a full scan on every query. Consider adding a mismatch ratio threshold if perf becomes an issue, but it's fine for now.

- **Infinity delete log**: you removed the debug log line for the delete filter (`self.logger.debug(f"INFINITY delete table ...")`). Might want to keep it at trace level instead of removing entirely, but minor.

Overall this looks solid. Thanks for the thorough fix! 🚀

PR #17347, opened 2026-07-24
fix: propagate publish release flag through Go agent endpoints

euvre:
### Summary

In Go mode, publishing an agent did not mark the published version as a production release: the version history dialog never showed the 正式版 (production) badge, and agent cards never showed the 发布于 (published at) time.

**Root cause:** the front-end publish flow sends `release: 'true'` together with the DSL via `PUT /api/v1/agents/<id>` (not the dedicated `POST /:id/publish` endpoint). Python's `update_agent` coerces this flag and writes it to both the canvas row and the new version row. The Go `UpdateAgent` ignored the `release` field entirely — the canvas row was never marked released and every version was saved with `release=false` hardcoded.

Additionally, two Python response fields were missing in Go:
- `GET /agents` list items lacked `release_time` (drives the card's 发布于 line)
- `GET /agents/<id>` lacked `last_publish_time` (drives the publish-confirm dialog)

**Changes:**
- `UpdateAgent`: coerce `release` with Python `bool()` parity (any non-empty string is truthy), always write it back to the canvas row, and pass it through to `saveOrReplaceVersion` instead of a hardcoded `false`
- `ListAgents`: attach `release_time` (`MAX(create_time)` of released versions per canvas, mirroring `UserCanvasService.get_list`)
- `GetAgent`: attach `last_publish_time` from the latest released version's `update_time` (Python `get_agent` parity)
- `UserCanvasVersionDAO`: add `GetLatestReleased` and `GetLatestReleaseTimes`; quote the `` `release` `` column with backticks since `RELEASE` is a reserved keyword in MySQL (unit tests run on SQLite and did not catch this)

**Tests:** added coverage for publish marking canvas+version released, release-absent reset parity, and list response including `release_time`.


wangq8 (commented):
Thanks for this fix! The root cause analysis is spot-on, and the changes cleanly mirror Python parity across `UpdateAgent`, `ListAgents`, and `GetAgent`.

**What works well:**

- **Backtick-quoting `release`** in the new DAO queries — `RELEASE` is indeed a MySQL reserved keyword and the existing queries in `user_canvas_version.go` (e.g. `GetLatest`) already use backticks for the same column. Good catch.
- **Release coercion in `UpdateAgent`** correctly mirrors Python `bool(req.get("release", ""))` — any non-empty string is truthy, absent/null → false.
- **`updates["release"] = release`** always writing to the canvas row — matches Python behavior where the coerced value is unconditionally written back, even when `release` is absent from the request.
- **`GetLatestReleaseTimes` batch query** — avoids N+1 on `ListAgents` and handles empty `canvasIDs` input cleanly.
- **`GetLastPublishTime` returning `nil, nil`** for never-published canvases — keeps the `agentDetailResponse` JSON clean with `omitempty`.
- **Tests cover all three paths** — publish-with-release, update-without-release (un-publish parity), and list-with-release-time. Each test verifies the full round-trip.

**Minor note (non-blocking):**

The release coercion in `UpdateAgent` only handles `bool` and `string` types. If the front-end ever sends `release: 1` (JSON number), it would fall through to the default (`release = false`). In practice the front-end sends `"true"` so this is fine, but it is a slight deviation from Python `bool()` which would coerce `1` → `True`. Worth being aware of, but not a blocker.

**Verdict:** LGTM. Clean fix, good tests, correct Python parity.

internal/service/agent.go
```diff
@@ -683,7 +734,7 @@ func (s *AgentService) UpdateAgent(ctx context.Context, userID, canvasID string,
 		} else if canvasInstance.Title != nil {
 			title = *canvasInstance.Title
 		}
-		if _, err := s.saveOrReplaceVersion(ctx, userID, canvasID, dsl, title, nil, false); err != nil {
+		if _, err := s.saveOrReplaceVersion(ctx, userID, canvasID, dsl, title, nil, release); err != nil {
```
Hz-186: @euvre, could u fix this ?  
Please make the canvas update and version save atomic. `UpdateFields` commits the canvas `release`/DSL changes before `saveOrReplaceVersion` starts its own transaction. If the version write fails, the request returns an error while the canvas remains published or unpublished without a matching version state. Please execute both writes in one shared transaction using the transaction-bound DAO methods.

Hz-186 (changes requested):

wangq8:
@euvre resolve conflict.

PR #17351, opened 2026-07-24
fix(ingestion): discard stale checkpoint when the DSL is edited before resume

xugangqiang:
## Summary
- Root cause: Pipeline.Run keys the eino checkpoint by taskID only. Resuming a failed run after the user edits the pipeline DSL recompiles a graph with different topology, but the old checkpoint (bound to the previous graph's node ids / wiring) is restored, causing eino to error.
- Fix: fingerprint the DSL file (full canvas DSL) and the runtime override_params, persisted next to the eino checkpoint. On resume, if either fingerprint differs, discard the stale checkpoint + interrupt marker and re-run from scratch. The warning log distinguishes a DSL-file edit from a runtime-override edit.

## Test plan
- TestPipelineRunResumableDSLChanged: editing the DSL between runs discards the checkpoint and re-runs from scratch.
- TestPipelineRunResumableOverrideChanged: editing only the runtime override does the same.
- TestClassifyDSLChange: unit-tests the mismatch-reason classifier.
- bash build.sh --test ./internal/ingestion/pipeline/... passes.

## Notes
- Component-code changes (e.g. a component's output contract) are not covered by the DSL fingerprint; that is a separate, smaller-blast-radius gap noted in code comments.

### DIFF
--- a/internal/ingestion/pipeline/pipeline.go
+++ b/internal/ingestion/pipeline/pipeline.go
@@ -247,11 +247,6 @@ const (
 // grouping) AND the per-component params/defaults — i.e. everything the user
 // can edit in the DSL template file. The raw DSL bytes are key-sorted at
 // decode time, so logical edits (not whitespace) drive the value.
-// dslFileFingerprint returns a stable hash of the FULL canvas DSL this
-// pipeline was compiled from. It captures topology (components, edges, graph
-// grouping) AND the per-component params/defaults — i.e. everything the user
-// can edit in the DSL template file. The raw DSL bytes are key-sorted at
-// decode time, so logical edits (not whitespace) drive the value.
 //
 // It returns an error (instead of a silent fallback) when rawDSL is unset:
 // NewPipelineFromDSL always populates rawDSL, so an empty value means the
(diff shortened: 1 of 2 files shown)


yuzhichang:
LGTM

xugangqiang:
**Discussion: is the DSL canonicalization round-trip necessary?**

While reviewing this PR, a question came up about `NewPipelineFromDSL` doing `json.Marshal(canvasDSL)` to produce a "canonical" byte form before hashing it for the resume fingerprint.

**Finding:** the round-trip is redundant for the canvas path. The canvas DSL is loaded from the DB (`user_canvas.dsl` longtext) into `entity.JSONMap` (`map[string]interface{}`); on read it is `json.Unmarshal`-ed (order lost) and then `json.Marshal`-ed again in `loadDSLFromCanvas` (`internal/ingestion/task/pipeline_executor.go`). Go's `json.Marshal` sorts map keys, so the string reaching `NewPipelineFromDSL` is already deterministic — re-serializing it adds nothing.

For built-in templates (static `.json` files), a formatting-only change is a deliberate commit and the only cost of a false positive is a one-time fresh re-run, so the stakes are low.

**Conclusion / next step:** the current canonicalization can safely be dropped (store the incoming `dsl` bytes directly), but we are deferring that simplification and keeping the present behavior for now. The decision — including where to compute the fingerprint and whether to keep the `rawDSL` field — is tracked separately in #17365.

PR #17355, opened 2026-07-24
Fix: parse by somark in parser

Lynn-Inf:
### Summary

As title, recognize parse_method provider.

### DIFF
--- a/rag/flow/parser/parser.py
+++ b/rag/flow/parser/parser.py
@@ -35,6 +35,9 @@ from api.db.joint_services.tenant_model_service import (
     resolve_model_config,
     get_tenant_default_model_by_type,
 )
+from api.db.services.tenant_model_instance_service import TenantModelInstanceService
+from api.db.services.tenant_model_provider_service import TenantModelProviderService
+from api.db.services.tenant_model_service import TenantModelService
 from common import settings
 from common.constants import LLMType
 from common.misc_utils import get_uuid, thread_pool_exec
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The change looks clean and addresses a real gap — when `parse_method` is passed as a tenant_model ID (UUID) from the new Tenant LLM Provider UI, it was silently falling through all parser branches without matching.

**What this does well:**
- Minimal, focused change (14 lines, single file)
- Properly resolves tenant_model ID → `model_name@instance_name@provider_name` format that the existing `@somark`/`@mineru`/`@paddleocr` suffix checks expect
- Guards each lookup result (`exist`, `provider_ok`, `instance_ok`) before constructing the resolved string
- No change to existing behavior when `raw_parse_method` is already a proper parser name

**One small suggestion (non-blocking):**
If the provider or instance lookup fails (e.g. data inconsistency), it currently fails silently — `raw_parse_method` stays as the raw UUID and falls through all parser branches. Consider adding a `logging.warning(...)` in the fall-through case to help with debugging.

**Verified:**
- All three service classes (`TenantModelService`, `TenantModelProviderService`, `TenantModelInstanceService`) exist and inherit `get_by_id` from `CommonService`
- Model fields (`model_name`, `instance_name`, `provider_name`) match the constructed format
- No circular import risk — these are leaf DB service imports, consistent with existing patterns in this file

LGTM 👍

PR #17363, opened 2026-07-24
fix: disable select all option in MemoriesFormField component

dcc123456:
### Summary

fix: disable select all option in MemoriesFormField component


wangq8 (commented):
Simple and clean fix. The `showSelectAll` prop already exists on the `MultiSelect` component (defined at `multi-select.tsx:188`, defaults to `true`), so this is just disabling it for the Memories form field where selecting all doesn't make sense given the embedding-consistency constraint already enforced by the `useDisableDifferenceEmbeddingMemory` hook.

LGTM! ✅

PR #17364, opened 2026-07-24
Fix: verify remote model when model_list in argument  and factory_llms is empty

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/services/provider_api_service.py
+++ b/api/apps/services/provider_api_service.py
@@ -668,7 +668,23 @@ async def verify_api_key(provider_id_or_name: str, api_key: str | dict, base_url
     else:
         factory_llms = factory_info[0]["llm"]
         if not factory_llms:
-            return False, f"No models found for provider '{provider_id_or_name}'", {}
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The approach makes sense — when `factory_llms` is empty, falling back to dynamically discovering models from the remote provider is much better than immediately returning an error. This is especially useful for providers that don't pre-configure their model list in `factory_llms`.

**What I checked:**
- The `ModelMeta[provider_name](api_key, model_base_url).get_model_list()` pattern is already established in the codebase (used in `show_provider_llm` around line 213), so this is consistent.
- The constructed `factory_llms` entries (`model_type` / `llm_name`) match the format consumed by the verification loop below.
- The `try/except` guard ensures this fallback is non-breaking — if the remote call fails for any reason, it falls through to the original error return.

**One minor observation (non-blocking):** The `except Exception: pass` is quite broad. In a future iteration, it might be worth logging the exception at debug level so that intermittent network errors or API issues can be traced. But as a fallback, this is perfectly fine.

LGTM! ✅

PR #17367, opened 2026-07-24
feat: add  BASE_URL  to TencentHunYuan

dcc123456:
### Summary

feat: add  BASE_URL  to TencentHunYuan


wangq8 (approved):
LGTM! 

The HunYuan provider backend already uses the `openai/` LiteLLM route with a default base URL (`https://api.hunyuan.cloud.tencent.com/v1`), so exposing a configurable BASE_URL in the frontend is appropriate. This allows users to override the endpoint when using proxies or internal deployments.

Verified the enum value `LLMFactory.TencentHunYuan` exists in `web/src/constants/llm.ts` (line 40).

Simple, clean change — approved.

PR #17370, opened 2026-07-24
Fix get datasets owner retrieve the whole dataset

wangq8:


wangq8 (commented):
## PR Review: #17370 — Fix get datasets owner retrieve the whole dataset

**Author:** @wangq8 | **+137 / −4** across 12 files

### Summary

This PR replaces the old approach of fetching **all** datasets to build the owner filter on the frontend with a lightweight dedicated endpoint (`GET /datasets?type=filter`) that returns aggregated owner data (id, label, count) directly from the backend. Both Python and Go backends are updated consistently, and the frontend swaps `useFetchAllKnowledgeList` + `buildOwnersFilter` for `useGetDatasetFilter`.

---

### ✅ What&apos;s good

- **Clear performance win**: No longer pulling the full dataset list just to extract unique owners.
- **Consistent dual-backend implementation**: Python (Peewee) and Go (GORM) queries match — same visibility rules, same `GROUP BY tenant_id, nickname`.
- **Clean types**: `IDatasetFilter` and `DatasetOwnerFilter` are well-defined.
- **Proper React Query pattern**: `useQuery` with a named query key, sensible defaulting when data is unavailable.
- **Defensive nil checks in Go**: The Go service guards against nil `joinedTenant` entries.

---

### ⚠️ Issues to address

#### 1. Query cache not invalidated on dataset mutation

When a user creates or deletes a knowledge base, the existing code calls `queryClient.invalidateQueries({ queryKey: [KnowledgeApiAction.FetchKnowledgeListByPage] })` in several mutation hooks (`useCreateKnowledge`, `useDeleteKnowledge`, etc.). However, **none** of these invalidate the new `KnowledgeApiAction.FetchDatasetFilter` query.

This means after CRUD operations:
- The dataset list refreshes ✅
- The owner filter **does not** refresh ❌ (stale counts, missing new owners)

**Fix**: In every mutation that invalidates `FetchKnowledgeListByPage`, also invalidate `FetchDatasetFilter`:

```typescript
queryClient.invalidateQueries({
  queryKey: [KnowledgeApiAction.FetchDatasetFilter],
});
```

Or, if the invalidation is already broad (e.g., no specific key), verify it covers the new key.

---

#### 2. REST API design: `?type=filter` is not RESTful

Using a query parameter to overload the `/datasets` endpoint with a different response shape is an anti-pattern. A cleaner approach would be a dedicated sub-resource:

```
GET /api/v1/datasets/filters
```

or

```
GET /api/v1/datasets/owners
```

This separates the concern of "list datasets" from "get filter options" and makes the API more discoverable. However, since this follows existing patterns in the codebase, this is a **suggestion** rather than a blocker.

---

#### 3. `total` field is returned but unused

Both the Python and Go backends compute and return a `total` field (sum of all owner counts) in the filter response, but the frontend `useGetDatasetFilter` hook only extracts `data.filter` and ignores `total`. This is dead data on the wire — consider either removing it or using it (e.g., to show "X datasets total").

---

#### 4. Missing test coverage

No tests were added for:
- The new `get_owner_filter` / `GetOwnerFilter` database queries
- The new `list_dataset_filters` / `ListDatasetFilters` service methods
- The new `?type=filter` API route handler
- The new `useGetDatasetFilter` hook

Given that this is replacing a potentially expensive operation, adding basic integration/unit tests would improve confidence.

---

#### 5. Frontend `useQuery` has no `staleTime` configuration

```typescript
useQuery({
  queryKey: [KnowledgeApiAction.FetchDatasetFilter],
  queryFn: async () => { ... },
});
```

Owner/team membership changes infrequently. Adding a reasonable `staleTime` (e.g., 5–10 minutes) would reduce unnecessary refetches on window focus / component remounts without meaningful risk of staleness.

```typescript
useQuery({
  queryKey: [KnowledgeApiAction.FetchDatasetFilter],
  queryFn: async () => { ... },
  staleTime: 5 * 60 * 1000, // 5 minutes
});
```

---

### 🔍 Nitpick

- In `api/db/services/knowledgebase_service.py:get_owner_filter`, the `JOIN` to `User` uses `User.nickname` — if a user changes their nickname, the filter label will update on the next query, which is fine. But if a user is deleted, the `JOIN` will exclude that owner entirely from results. For team-owned KBs where the original owner was deleted, this could hide datasets from the filter. Consider using `LEFT JOIN` and falling back to a default label (tenant ID) if `nickname` is NULL. (Same concern applies to the Go `GetOwnerFilter`.)

---

### Verdict

**Request changes** — the cache invalidation bug (#1) will cause a visible UX issue. The other items are suggestions and nitpicks. Overall a solid improvement, just needs a small fix on the frontend invalidation side.

euvre (approved):

wangq8:
Tested, worked on both Python and Go version.

PR #17376, opened 2026-07-25
[Fix] Refined the NLTK data usage directory

paulhuiseismic:
### Summary

download_deps.py downloads NLTK data into ragflow_deps/nltk_data (see ragflow_deps/download_deps.py); point NLTK_DATA there directly instead of a stale top-level ./nltk_data so this matches what's actually populated.


myf-bee (commented):
Looks Good To Me

PR #17380, opened 2026-07-25
fix(go): implement Anthropic streaming (ChatStreamlyWithSender)

### LINKED ISSUE #17333: [Go] Anthropic: streaming (ChatStreamlyWithSender) not implemented
## Description
The Go Anthropic driver (`internal/entity/models/anthropic.go`) does not implement streaming. `ChatStreamlyWithSender` (L436) immediately returns an error `"no such method"`.

The non-streaming `ChatWithMessages` works (and correctly extracts the system prompt into the top-level `system` field), but there is no streaming counterpart.

## Impact
Any caller that requests a streamed response from an Anthropic (Claude) model via the Go path fails. The Python `AnthropicCV` driver does support streaming, so behavior diverges between the two implementations.

## Expected behavior
Implement `ChatStreamlyWithSender` for Anthropic: open the Messages streaming endpoint, parse the SSE `message.delta` events for `text` and `thinking`, and forward them via the `sender` callback (consistent with the other Go drivers).

## References
- `internal/entity/models/anthropic.go:436`
- Python (reference): `rag/llm/cv_model.py` (AnthropicCV streaming)

abhayishere:
### Summary

The Go Anthropic driver's `ChatStreamlyWithSender` (internal/entity/models/anthropic.go) was a stub that always returned `"no such method"`, so any caller requesting a streamed response from a Claude model via the Go path failed outright — diverging from the Python `AnthropicCV` driver, which already supports streaming.

This implements the method by opening the Messages API with `stream=true` and parsing the SSE response via the shared `ParseSSEStream` helper, forwarding `text_delta`/`thinking_delta` content through the `sender` callback and treating `message_stop` as the terminal event — consistent with the other Go drivers in this package (e.g. Cohere).

Verified against the real Anthropic API: streaming a multi-word response correctly delivers multiple incremental chunks rather than one blob, and terminates cleanly.

Fixes #17333

### DIFF
--- a/internal/entity/models/anthropic.go
+++ b/internal/entity/models/anthropic.go
@@ -496,6 +496,8 @@ func (a *AnthropicModel) ChatStreamlyWithSender(ctx context.Context, modelName s
 	}
 
 	sawTerminal := false
+	var streamUsage TokenUsage
+	sawUsage := false
 	done, err := ParseSSEStream[map[string]interface{}](resp.Body, func(event map[string]interface{}) error {
 		eventType, _ := event["type"].(string)
 		switch eventType {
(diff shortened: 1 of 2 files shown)


internal/entity/models/anthropic.go
```diff
@@ -434,7 +434,137 @@ func (a *AnthropicModel) CheckConnection(ctx context.Context, apiConfig *APIConf
...
+
+	ctx, cancel := context.WithTimeout(ctx, streamCallTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(jsonData))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+	setAnthropicHeaders(req, apiKey)
+
```
Haruko386:
```suggestion
	setAnthropicHeaders(req, apiKey)
	req.Header.Set("Accept", "text/event-stream")
```

Haruko386 (changes requested):
Hi @abhayishere 
According to Anthropic’s official Streaming Messages documentation, a response with `stream: true` is in SSE format, and each event includes `event: ...` and a corresponding JSON `type`. In this case, the request body has been set to `‘stream’: true`, but the header still uses the `Accept: application` value from `setAnthropicHeaders`.

internal/entity/models/anthropic.go
abhayishere: Hi @Haruko386, thanks for the review. Instead of splitting the header setting logic across two places, I have updated setAnthropicHeaders itself to take a streaming bool param, so the Accept header decision stays in one function. Streaming now sends Accept: text/event-stream, non-streaming calls keep application/json. Pushed in the latest commit.

Haruko386 (approved):
LGTM

PR #17385, opened 2026-07-25
Improve "Delimiter for text" UX: clearer tooltip + live parsed-delimiter preview

skbs-eng:
## Summary

Improve the UX of the **"Delimiter for text"** field on the dataset configuration page. The field is a single string with a backtick-based mini-syntax, but both the tooltip and the surrounding UI failed to surface what delimiters the backend would actually derive from a given value — leaving users to discover by trial-and-error that the same string produces different splits depending on file type (see #7436, #4704, #9680).

This PR makes two additive, risk-free changes:

1. **Rewrites the tooltip copy** (English locale only) so it explicitly names backticks as the meta-delimiter and states the parsing rule.
2. **Adds a live "Splits at:" preview** under the input that shows, in real time, the list of delimiters the backend will derive from the current value, with whitespace characters rendered as visible Unicode glyphs.

No backend changes. No API change. No schema change. Pure presentation.

## Related issues

- **Closes #17382** (tooltip clarity, docs fix)
- Adds UI support for, but does not fix, the bugs tracked in:
  - **#17383** — six divergent backend parser implementations; the markdown parser silently makes the shipped default value a no-op; CRLF normalization only in one path; etc.
  - **#17384** — `re.I` case-insensitive delimiter matching (two-line fix).
- Addresses recurring user confusion in **#7436**, **#4704**, **#9680**.

The backend fix is a separate, larger refactor tracked in #17383. This PR is intentionally limited to UX so users get immediate relief while the backend consolidation is scoped and reviewed.

## Changes

Two commits on branch `feat/delimiter-ux`:

```
925882440  docs(i18n): clarify the "Delimiter for text" tooltip
cda59bf85  feat(ui): add live preview of parsed delimiters next to the input
```

### `docs(i18n): clarify the "Delimiter for text" tooltip` (925882440)

Updates `web/src/locales/en.ts` `knowledgeDetails.delimiterTip` and `knowledgeDetails.childrenDelimiterTip`. The previous tooltip described backticks as the multi-char delimiter wrapper but never named them as the meta-delimiter. The new copy explicitly states the parsing rule:

  - The field is a string.
  - Backticks group characters into one multi-character delimiter.
  - Any character outside backticks is itself a delimiter.
  - Delimiters are matched longest-first.
  - Single-char delimiters need no backticks; multi-char ones must be backtick-wrapped.

Other locales are not touched — they fall through to the English copy via `defaultValue`. If the project enforces 100% locale parity, please flag and I can extend in a follow-up commit.

[...]

Issue #17383, opened 2026-07-25
[Bug] Delimiter parsing has 7 divergent implementations that read the same `parser_config.delimiter` field


skbs-eng:
### Self Checks

- [x] I have searched for existing issues (including closed ones).
- [x] I confirm that I am using English to submit this report.
- [x] Non-English title submissions will be closed directly.
- [x] I have not modified this template.

### Describe the problem

The single string field `parser_config.delimiter` is consumed by **seven different parsing implementations** depending only on the file extension. They disagree on the parsing rule, at least one silently makes the shipped default value a no-op, and the **whitespace handling has five distinct bugs** (see "Whitespace-specific gotchas" below).

Affected functions, all of which read `parser_config.delimiter`:

| Location | Triggered by | Behavior |
|---|---|---|
| `rag/nlp/__init__.py::get_delimiters` | PDF/DOCX/HTML/EPUB/JSON/CSV/XLSX/email/book, **only when no backticks present** | Bare chars + backtick-wrapped tokens, longest-first sort, **no dedupe**, uses `re.I`. |
| `rag/nlp/__init__.py::naive_merge` (inline regex) | Same files, **whenever any backtick is present** | **Drops bare chars**, only backtick-wrapped tokens, deduped, longest-first sort. |
| `rag/nlp/__init__.py::naive_merge_with_images` (inline regex) | Same files with images | Same as `naive_merge`. |
| `rag/nlp/__init__.py::_build_cks` (inline regex) | `.docx` | Same as `naive_merge`. |
| `deepdoc/parser/txt_parser.py::parser_txt` (inline regex) | `.txt`, `.py`, `.js`, `.java`, `.c`, `.cpp`, `.h`, `.php`, `.go`, `.ts`, `.sh`, `.cs`, `.kt`, `.sql` | Bare chars + backtick-wrapped tokens, **no longest-first sort**, uses `re.I`, **no CRLF normalization**. |
| `deepdoc/parser/markdown_parser.py::MarkdownElementExtractor.get_delimiters` | `.md`, `.mdx`, `.markdown` | **Only backtick-wrapped tokens, drops bare chars**, deduped, longest-first sort, **no CRLF normalization**. |
| `rag/flow/chunker/token_chunker.py::_compile_delimiter_pattern` | Agent flow `Token chunker` component | Only backtick-wrapped, deduped, longest-first sort. **Different field** (consumes an array of strings, not `parser_config.delimiter`). Out of scope. |

[...]

PR #17386, opened 2026-07-25
fix(nlp): drop dead `re.I` from delimiter finditer calls

### LINKED ISSUE #17384: [Bug] Delimiter parsing uses `re.I` (case-insensitive), causing unintended splits on both cases of the same character
### Self Checks

- [x] I have searched for existing issues (including closed ones).
- [x] Non-English title submissions will be closed directly.

### Describe the problem

The two implementations of delimiter parsing that pre-process the string with `re.finditer(r"\`([^\`]+)\`", delimiters, re.I)` apply the case-insensitive flag when matching the user's backtick-wrapped tokens. The same flag is then passed (or implied) on the downstream `re.split` call in some paths. As a result:

- A user-supplied delimiter of `a` splits on both `a` and `A`.
- A user-supplied backtick-wrapped delimiter of `` `end` `` matches `end`, `End`, `END`, `eNd`, etc.
- A delimiter of `;` matches `;` only — but a delimiter of `s` matches `s`, `S`, plus any other letter/S-shape in unicode (depending on Python's Unicode case-folding rules for `re.I`).

This is almost certainly unintended. No documentation mentions case-insensitive matching; the tooltip and field semantics imply exact, character-exact matching. Users who include a single-character delimiter like `s` will see far more splits than they configured.

### Affected sites

- `rag/nlp/__init__.py::get_delimiters` (line 1633) — uses `re.I`.
- `deepdoc/parser/txt_parser.py::parser_txt` (line 51) — uses `re.I`.

The `has_custom` inline regexes in `naive_merge`, `naive_merge_with_images`, and `_build_cks` in `rag/nlp/__init__.py` do **not** use `re.I`, so the case-insensitivity only affects:

- `.txt` and code files (via `parser_txt`), and
- the bare-char fallback path of `naive_merge` / `naive_merge_with_images` / `_build_cks` (via `get_delimiters`) when no backticks are present.

### Reproduction

```python
import re
# Mirror the production logic:
s = "s"
for m in re.finditer(r"`([^`]+)`", s, re.I):
    print(m.group(1))

# Now apply the same flag to re.split on the parsed delimiter:
pattern = "|".join(re.escape(d) for d in [s])  # just 's'
print(re.split(r"({})".format(pattern), "Class A is a Sample", flags=re.I))
# Expected (per the field semantics): ['Class A i', 's', ' a Sample'] only at 's'
# Actual: splits at every 's' AND every 'S':
# ['Cla', 's', '', 's', ' A i', 's', ' a Sample']
```

For a real-world example: a user setting `delimiter = ". "` (period + space, a common sentence boundary) would also get splits at `"。 "` (Chinese period) — actually that's a different character, but `"a"` would also match `"A"`.

### Expected behavior

Delimiter matching should be **case-sensitive by default**. If a future feature needs case-insensitive matching, it should be opt-in (e.g. a separate `delimiter_case_insensitive` flag in `parser_config`, or a syntax marker like `` `(?i)end` ``).

### Proposed fix

Remove `re.I` from the two `re.finditer` calls. This is a one-character-per-site change but should ship with tests. Since this is part of the broader "collapse the six implementations" effort tracked in #17383, the cleanest fix is to make the change there. If #17383 is not picked up soon, this should be fixed independently first because it can cause silent data loss (extra chunks from unwanted splits).

### Files to change

- `rag/nlp/__init__.py` — `get_delimiters` (line 1633).
- `deepdoc/parser/txt_parser.py` — `parser_txt` (line 51).

### Acceptance criteria

- Delimiter `a` matches only `a`, not `A`.
- Delimiter `` `end` `` matches only `end`, not `End`/`END`/etc.
- New unit test `test_delim_case_sensitive` covers both single-char and backtick-wrapped cases.
- All existing tests in `test_naive_merge.py` still pass (none of them currently assert case-insensitive behavior, so removing it should be a safe no-op).

### Related

- #17383 — broader parser-collapse refactor; the fix here should land in that work.
- #17382 — tooltip clarity (separate docs fix).

skbs-eng:
Closes #17384.

## Summary

Drops a dead `re.I` flag from two outlier delimiter-parsing sites and adds regression tests so the inconsistency can't creep back.

## What's wrong

Two of the six delimiter-parsing implementations pass `re.I` to `re.finditer`:

- `rag/nlp/__init__.py::get_delimiters` (line 1633)
- `deepdoc/parser/txt_parser.py::parser_txt` (line 51)

The other four implementations correctly omit `re.I`:

- `rag/nlp/__init__.py::naive_merge` custom-delimiter path (line 1195)
- `rag/nlp/__init__.py::naive_merge_with_images` custom-delimiter path (line 1269)
- `rag/nlp/__init__.py::_build_cks` (line 1389)
- `rag/flow/chunker/token_chunker.py` (line 73)

## Why this matters (and why it doesn't break anything)

The flag is **dead code** today. Verified empirically with a Python REPL:

```python
>>> import re
>>> for m in re.finditer(r"`([^`]+)`", "`end`", re.I):
...     print(repr(m.group(1)))
'end'                           # plain string, no flag attached
>>> re.split("(a)", "Class A is a Sample")
['Cl', 'a', '', 's', ' A i', 's', ' a Sample']
# Case-sensitive: only lowercase 'a' splits. Uppercase 'A' is preserved.
```

`re.I` does not propagate from `re.finditer` to `m.group(1)` or to downstream `re.split` / `re.match` calls (which all omit `re.I`). So the actual splitting behavior has always been case-sensitive — removing the flag is a **defensive cleanup**, not a behavioral fix.

So why bother?

1. **Consistency** — the two sites were the only outliers in a six-way implementation cluster. The three sibling sites in `rag/nlp/__init__.py` already omit `re.I`, which strongly suggests the flag was accidental.
2. **Future-proofing** — a refactor could easily propagate the flag to a downstream `re.split` call where it *would* change behavior. The tests added here pin the case-sensitive semantics so that regression fails loudly.
3. **Reader clarity** — the flag is misleading. Anyone reading `re.finditer(..., re.I)` reasonably assumes case-insensitive matching, then has to trace all downstream calls to discover it's a no-op.

## Changes

- `rag/nlp/__init__.py` — drop `re.I` from `get_delimiters` (line 1633).
- `deepdoc/parser/txt_parser.py` — drop `re.I` from `parser_txt` (line 51).
- `test/unit_test/rag/test_delimiter_case_sensitive.py` — new test file with:
  - 4 behavioral tests on `get_delimiters` (pattern output + `re.split` round-trip).
  - 3 end-to-end tests through `naive_merge` (bare-char + backtick-wrapped, both cases).
  - 2 parametrized static checks that `re.I` / `re.IGNORECASE` is not present at either of the two `re.finditer` sites.

## Testing

```
$ pytest test/unit_test/rag/test_delimiter_case_sensitive.py -v
============================= 9 passed in 0.19s ==============================
```

All tests pass on the patched code. Before the patch, the 2 static checks fail with a clear assertion message (the 7 behavioral tests pass either way, confirming `re.I` was dead code).

## Related

- #17384 — the issue this PR closes. Note the issue's reproduction code (`re.split(..., flags=re.I)`) doesn't actually match what the production code does — the production `re.split` calls all omit `re.I`, which is why current behavior is already case-sensitive. The fix here is still valuable as a defensive cleanup + test coverage, but it's not a behavioral fix per se.
- #17383 — broader parser consolidation (six implementations → one). The fix here is independent and small enough to land first.
- #17385 — sibling UX PR (tooltip + live preview). Files are disjoint (`web/src/**` vs `rag/nlp/**` + `deepdoc/parser/**`), so no interaction.

### DIFF
--- a/deepdoc/parser/txt_parser.py
+++ b/deepdoc/parser/txt_parser.py
@@ -48,7 +48,7 @@ class RAGFlowTxtParser:
 
         dels = []
         s = 0
-        for m in re.finditer(r"`([^`]+)`", delimiter, re.I):
+        for m in re.finditer(r"`([^`]+)`", delimiter):
             f, t = m.span()
             dels.append(m.group(1))
             dels.extend(list(delimiter[s:f]))
--- a/internal/ingestion/component/chunker/common.go
+++ b/internal/ingestion/component/chunker/common.go
@@ -81,6 +81,31 @@ func stringListFromAny(in []any) []string {
 // never add (?i) here — delimiter matching must preserve letter casing.
 var backtickDelimRE = regexp.MustCompile("`([^`]+)`")
 
+// escapeAndSort QuoteMeta-escapes tokens and sorts longest-first.
+// Empty tokens are dropped. When dedup is true, exact duplicate tokens
+// are collapsed (first occurrence wins). Matching stays case-sensitive.
+func escapeAndSort(tokens []string, dedup bool) []string {
+	out := make([]string, 0, len(tokens))
+	var seen map[string]struct{}
+	if dedup {
+		seen = make(map[string]struct{}, len(tokens))
+	}
+	for _, tok := range tokens {
+		if tok == "" {
+			continue
+		}
+		if dedup {
+			if _, ok := seen[tok]; ok {
+				continue
+			}
+			seen[tok] = struct{}{}
+		}
+		out = append(out, regexp.QuoteMeta(tok))
+	}
+	sort.SliceStable(out, func(i, j int) bool { return len(out[i]) > len(out[j]) })
+	return out
+}
+
 // getDelimiters ports rag.nlp.get_delimiters (rag/nlp/__init__.py).
 //
 // It walks a single delimiter string, pulling out backtick-wrapped tokens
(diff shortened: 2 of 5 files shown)


skbs-eng:
@yuzhichang @Lynn-Inf @wangq8 @dcc123456 please review!

skbs-eng:
@yuzhichang @Lynn-Inf @wangq8 @dcc123456 @JinHai-CN  please review!

yuzhichang:
@skbs-eng We are actively porting from python to golang. Could you implement the go part in this same PR?

skbs-eng:
@yuzhichang Done — Go parity is in this PR now.

### Go changes
- `internal/ingestion/component/chunker/common.go`
  - Added `getDelimiters` (port of `rag.nlp.get_delimiters`) — case-sensitive backtick + bare-char extraction.
  - Aligned `compileDelimPattern` with Python `_compile_delimiter_pattern` (join entries → extract backtick tokens → `QuoteMeta` + length-sort). Explicitly no `(?i)`.
- `internal/ingestion/component/chunker/delimiter_case_sensitive_test.go`
  - Pattern-level checks for bare `a`/`A` and `` `end` ``
  - `compileDelimPattern` case-sensitive match / no-`(?i)` guards
  - End-to-end TokenChunker tests for `` `end` `` and `` `a` ``

### Validation
```
go test ./internal/ingestion/component/chunker/ -count=1
ok
```

yuzhichang (approved):

yuzhichang:
@skbs-eng Please format test/unit_test/rag/test_delimiter_case_sensitive.py via ruff.

PR #17346, opened 2026-07-24
fix: preserve non-string values in update_metadata_to

huang-aoqin:
## Summary
- Broaden `update_metadata_to` to keep `bool` / `int` / `float` / `None` and structured values (e.g. PDF `outline` list[dict]) while retaining string / list[str] merge+dedupe for LLM metadata.
- Fixes document system fields (`_isCurrent`, `_version`, `_processStatus`) being dropped when parse persists PDF outline via full `meta_fields` replace.

## Test plan
- [ ] Unit tests in `test/unit_test/common/test_update_metadata_to.py`
- [ ] Upload a PDF with existing `_version` / `_isCurrent` metadata, parse with MinerU/manual, confirm fields remain and `outline` is present

PR #17391, opened 2026-07-26
fix: stabilize automatic metadata extraction

roy5257:
### Summary

Automatic document metadata extraction can intermittently return an empty object when the model receives a Python `dict` representation of the schema, a minimal `Output:` instruction, and the provider's default sampling temperature. In addition, scalar values such as numeric years are discarded by the existing string-based metadata merge.

This PR makes automatic metadata extraction deterministic and normalizes only LLM-extracted metadata to the flat string-based document metadata schema.

### Changes

- Serialize a private copy of the extraction schema with `json.dumps(..., ensure_ascii=False, indent=2)` before rendering the metadata prompt, avoiding mutation of the caller's schema and cache configuration.
- Ask for one flat JSON object with no Markdown or explanation, and disallow `{}` when the content explicitly contains a schema field.
- Set `temperature=0` only for the `gen_metadata()` model call. Ordinary chat calls and provider defaults are unchanged.
- Add `normalize_extracted_metadata()` to:
  - keep strings and `list[str]` values;
  - stringify `int`, `float`, and `bool` values (`bool` follows Python `str()` semantics: `"True"`/`"False"`);
  - filter non-string list members;
  - reject `None`, nested objects, and non-object responses;
  - preserve `json_repair` compatibility, including fenced JSON.
- Apply normalization in both the original and refactored task-executor metadata paths while keeping the raw model response unchanged in the LLM cache.

The normalization is intentionally scoped to LLM-extracted metadata. It does not broaden or change the generic `update_metadata_to()` behavior and is therefore complementary to #17346, which addresses preservation of non-string system metadata.

### Compatibility and scope

- No database or index schema changes.
- No changes to ordinary chat generation parameters.
- No global model-temperature changes.
- Empty objects and malformed/non-object metadata still result in no metadata update.
- Existing Markdown-fenced JSON parsing remains supported.

### Tests

- `test/unit_test/rag/prompts/test_generator_message_fit_in.py`: 6 passed
- `test/unit_test/common/test_extracted_metadata_normalization.py`: 3 passed
- `py_compile` passed for all modified Python source and test files.
- Ruff check and format check passed for the modified/new test files.
- `git diff --check` passed.
- The behavior was also validated end to end with five isolated synthetic PDF uploads: all five produced the expected six flat metadata fields, including normalization of a numeric year to a string.


xugangqiang:
Review focused on problems (read-only, no code changes):

## 1. Original path missing the `metadata_obj` guard (latent KeyError)
In `rag/svr/task_executor.py` (the `build_chunks` aggregation loop, ~line 566):
```python
metadata = {}
for doc in docs:
    metadata = update_metadata_to(metadata, normalize_extracted_metadata(doc["metadata_obj"]))
    del doc["metadata_obj"]
```
`gen_metadata_task` only sets `d["metadata_obj"]` when `if cached:` is true. When the model returns `""` (e.g. schema has no `properties`), `cached` is falsy and `metadata_obj` is never written, so this loop raises `KeyError`. The refactored path in `rag/svr/task_executor_refactor/chunk_post_processor.py` already guards with `if "metadata_obj" in doc:`, so the two paths are **inconsistent**. This PR stabilized only one of the two paths.

Fix: align the original loop with the refactored one:
```python
metadata = {}
for doc in docs:
    if "metadata_obj" in doc:
        metadata = update_metadata_to(metadata, normalize_extracted_metadata(doc["metadata_obj"]))
        del doc["metadata_obj"]
```

## 2. System prompt `{}` instruction contradicts the new user prompt
`rag/prompts/meta_data.md:7` still says "If no matches are found ... output ONLY {}". The new user prompt says "If any schema field is explicitly present ... you must not return {}". They can be reconciled logically, but keeping the blanket "output ONLY {}" in the system template weakens the stabilization intent and adds model-confusing noise. Please update `meta_data.md` to relax that rule so it matches the user prompt.

## 3. Test coverage gaps for behavioral changes
- No test for the missing-`metadata_obj` case above (the exact scenario that triggers the KeyError).
- `normalize_extracted_metadata` now **drops structured values** (nested objects, `list[dict]`) whereas `update_metadata_to` previously preserved them. This is intended for a flat string schema, but it is a behavior change and should be locked with a test asserting nested objects/structured lists are dropped.
- `bool -> "True"/"False"` stringification is untested; if any downstream consumer expects a real boolean for a metadata field, this would regress. Worth an explicit test or doc note.
- No test for the dict (non-string) input path of `normalize_extracted_metadata`.

Codecov reports modified lines covered, but it does not exercise these edge cases.

## 4. Broad `except Exception` in `normalize_extracted_metadata` (minor)
Lines ~303-308 catch `Exception` around `json_repair.loads`, making parser regressions indistinguishable from malformed model output. This is consistent with the existing `update_metadata_to` style, so it is acceptable as-is; if you want to adopt the stricter approach, narrow it and also update `update_metadata_to` for consistency rather than changing only one place.

**Recommendation:** fix #1 (and add a regression test) before merging; #2 and the #3 tests are strongly advised.

xugangqiang:
Please resolve conflict. @yuzhichang  not sure if this the right direction.

xugangqiang (commented):
Review (read-only) — issues found, no code changes made.

### 1. The `additionalProperties` assertion in the test is effectively a no-op
In `test/unit_test/rag/prompts/test_generator_message_fit_in.py`:
```python
assert "'additionalProperties': False" not in rendered_schema
```
After this change the schema is serialized with `json.dumps(schema, ensure_ascii=False, indent=2)`, so the rendered text is JSON (`"additionalProperties": false`, lowercase) rather than a Python `repr` (`'additionalProperties': False`). The literal substring `"'additionalProperties': False"` will never appear, so the assertion passes regardless of behavior. It also does **not** verify that `additionalProperties` was actually removed — it is still present in the rendered JSON.

Recommend either removing this assertion, or replacing it with a meaningful check, e.g.:
```python
assert rendered_schema_obj.get("additionalProperties") is False
```
to confirm the schema is rendered as valid JSON.

### 2. `bool` values are normalized to `"True"`/`"False"` (capitalized)
In `normalize_extracted_metadata` (`common/metadata_utils.py`), `bool` is stringified via Python `str()`, producing `"True"`/`"False"`. Please confirm this matches the document metadata schema convention. If any downstream consumer (UI, filtering, JSON serialization) expects lowercase `"true"`/`"false"` or a native bool, this casing will cause mismatches. Consider emitting lowercase for booleans, or preserving native bool if the schema allows it.

### 3. PR description overstates the "scalars discarded" problem on the current base
The description states scalar values such as numeric years are "discarded by the existing string-based metadata merge." On the current `main`, `update_metadata_to` (from #17346) already preserves `bool/int/float/None` scalars (`common/metadata_utils.py:351`). So `normalize_extracted_metadata` is really converting scalars to strings to align with the flat string schema, not fixing a drop. Please reword the description so reviewers aren't misled about the actual behavior change.

### Minor — behavior change worth documenting
`normalize_extracted_metadata` silently drops nested `dict` values and `None`. This is consistent with a flat string schema, but it changes the older behavior where `update_metadata_to` would have preserved a nested/None value when the key was absent. A one-line doc note would make the drop intentional and visible.

PR #17396, opened 2026-07-27
Chunker: port Python overlapped_percent normalization and BULLET_PATTERN title fallback

xugangqiang:
## Summary

Closes two Chunker migration gaps documented in `docs/migration_python_go_diff.md`
(diffs **2.6** and **1.7**), improving parity with the Python ingestion pipeline.

This is the code portion of commit `261e1fd0b` on branch `fix/batch-3-4`; the
migration document itself is tracked separately (untracked in this commit).

### Diff 2.6 — `overlapped_percent` missing Python normalization

Python's `normalize_overlapped_percent` (`common/float_utils.py:50-58`) accepts a
`[0,1)` fraction (the flow-canvas UI validates `[0,1)`), multiplies it by 100,
`int()`-truncates, and clamps to `[0,90]`. Go previously only accepted a raw
`[0,90]` percentage and **rejected** out-of-range input, so a Python config
passing `0.1` (meaning 10%) silently produced ~0% overlap.

Added `normalizeOverlappedPercent`
(`internal/ingestion/component/chunker/common.go`) mirroring the Python helper:

- parses numbers and numeric strings (mirrors Python `float()`; bad / `NaN` / `Inf` → `0`),
- `0 < v < 1 → v *= 100`,
- `int()` truncation,
- clamps to `[0,90]`.

Wired into `tokenChunkerParam.Update` (`token.go`); the merge math
(`token.go:705`, `(100-x)/100`) already matched Python. `TokenChunkerParam.Validate`
(`schema/chunker.go`) now only guards direct struct construction.

Tests: `TestNormalizeOverlappedPercent`, extended
`TestTokenChunker_NewAcceptsPythonOverlappedRange` (adds fraction/clamp inputs),
new `TestTokenChunker_NormalizesOverlappedPercent`. The existing reject-test cases
for `<0` / `>90` were removed because they are now normalized/clamped (Python parity).

### Diff 1.7 — missing `BULLET_PATTERN` title-level fallback

`resolveTitleLevels` (`title.go`) now applies a 4th-level fallback: when outline +
regex + layout all yield body level, `bulletsCategory` selects the best-matching
bullet-pattern group (Chinese legal / numbering / Chinese numbering / English legal
— mirroring `rag/nlp/__init__.py:258-320`) and assigns structural levels. Guarded by
`allBodyLevel` so it never overrides an existing outline/regex level.

Tests: `TestResolveTitleLevels_BulletFallback` (4 subtests).

### Incidental test adjustments included in the commit

- `token_batch1_test.go`: overlap input changed `0.3` → `30.0` to reflect the
  post-normalization 30% semantics.
- `real_consumer_test.go`: updated `LoadFromIngestionTask(task)` →
  `LoadFromIngestionTask(ctx, task)` for the new context-first signature.

## Verification

`bash build.sh --test ./internal/ingestion/component/...` — chunker + schema suites
pass, no regression (CGO build).

## Migration doc reference

`docs/migration_python_go_diff.md` §Chunker 1.7 and 2.6 are marked **Fixed** for
these changes.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -303,6 +303,16 @@ var sentenceDelimiter = regexp.MustCompile(`(\n|[!?。；！？]|\.\s)`)
 func (c *TokenChunkerComponent) mergeByTokenSize(text string, childrenPattern *regexp.Regexp) map[string]any {
 	target := c.param.ChunkTokenSize
 	overlapPct := c.param.OverlappedPercent
+	// Clamp to [0,100] so the merge math below never produces a
+	// negative/inverted threshold for an out-of-range value (review:
+	// yuzhichang, PR #17396). c.param.OverlappedPercent is already in
+	// [0,90] via Update/Validate, so this is a defensive no-op in
+	// normal operation.
+	if overlapPct < 0 {
+		overlapPct = 0
+	} else if overlapPct > 100 {
+		overlapPct = 100
+	}
 
 	// Split into paragraph-aligned sections.
 	sections := splitIntoSections(text)
@@ -710,6 +720,14 @@ func takeFromStart(text string, tokens int) string {
 // mergeByTokenSizeFromJSON mirrors `naive_merge` at
 // rag/nlp/__init__.py:1156.
 func mergeByTokenSizeFromJSON(perItem [][]schema.ChunkDoc, chunkTokens int, overlappedPct float64) [][]schema.ChunkDoc {
+	// overlappedPct is a [0,100] percentage. Clamp so the merge math below
+	// never yields a negative/inverted threshold for out-of-range input
+	// (review: yuzhichang, PR #17396).
+	if overlappedPct < 0 {
+		overlappedPct = 0
+	} else if overlappedPct > 100 {
+		overlappedPct = 100
+	}
 	threshold := float64(chunkTokens) * (100 - overlappedPct) / 100.0
 	for idx := range perItem {
 		chunks := perItem[idx]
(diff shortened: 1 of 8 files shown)


xugangqiang:
@yingfeng @JinHai-CN @yuzhichang  pls help to review

xugangqiang:
@yingfeng @JinHai-CN @yuzhichang Pls help to review

internal/ingestion/component/chunker/token.go
```diff
@@ -704,7 +710,7 @@ func takeFromStart(text string, tokens int) string {
 // mergeByTokenSizeFromJSON mirrors `naive_merge` at
 // rag/nlp/__init__.py:1156.
 func mergeByTokenSizeFromJSON(perItem [][]schema.ChunkDoc, chunkTokens int, overlappedPct float64) [][]schema.ChunkDoc {
-	threshold := float64(chunkTokens) * (100 - overlappedPct*100) / 100.0
+	threshold := float64(chunkTokens) * (100 - overlappedPct) / 100.0
```
yuzhichang, line 731: Can we assert `overlappedPct` in range `[0, 100]`?

xugangqiang:
While reviewing the overlap normalization in this PR, found a related edge case worth tracking separately: fraction inputs like `overlapped_percent = 0.29` are truncated to `28` instead of `29` (caused by `int()` truncation after the `*100` float multiplication). This matches Python exactly, so it's a shared pre-existing behavior, not a regression from this PR.

Filed as #17418 with reproduction table, root cause, impact, and the round-vs-truncate trade-off (rounding would diverge from Python). No code change in this PR for that item.

PR #17398, opened 2026-07-27
fix(go-agent): retrieval empty results, KB selector paging, and begin custom inputs

euvre:
### Summary

Three fixes for the Go-mode agent canvas:

**1. Agent retrieval returned no chunks.** The NLP retrieval adapter never resolved the knowledge base's embedding model, so searches ran keyword-only and the results were filtered out by the similarity threshold — downstream Agent components concluded the knowledge base had no relevant content. The adapter now resolves the kb's embedding model (and the node's optional rerank model) through an injected `RetrievalModelProvider` and mounts them on the nlp request, enabling hybrid (fulltext + vector + fusion) search, matching the Python canvas. The `rerank_id` node param is also threaded through (previously dropped silently), and when nothing is found the node's `empty_response` is surfaced as `formalized_content` instead of an empty output map.

**2. Knowledge base selector showed an empty list.** Server-side pagination combined with the client-side `chunk_count > 0` filter could filter out an entire page, leaving an empty, unscrollable dropdown that never triggered the next-page load. The list now auto-fetches the next page while the filtered result is empty and more pages exist. Also fixed variable option icons being rendered as image URLs, which fired bogus `/sys.*` img requests per option.

**3. Begin custom inputs (test-run panel / chat form) were dropped.** `RunAgent` ignored the request's `inputs`, so `{begin@<key>}` references (e.g. in a Retrieval query) never resolved in Go mode. Begin inputs are now threaded through `RunAgent` and seeded as the Begin component's outputs, and the Begin component applies the single-input `sys.query` fallback, mirroring Python's `Begin._invoke`.

Tests: `bash build.sh --test ./internal/agent/... ./internal/service/ ./cmd/...` all pass; new unit tests cover the model attachment, empty-response fill, begin input seeding, and the sys.query fallback.

### DIFF
--- a/cmd/ragflow_server.go
+++ b/cmd/ragflow_server.go
@@ -715,6 +715,9 @@ func startServer(ctx context.Context, config *server.Config) {
 	docEngine := engine.Get()
 	documentDAO := dao.NewDocumentDAO()
 	agenttool.SetRetrievalService(agenttool.NewNLPRetrievalAdapterFromDeps(docEngine, documentDAO))
+	// Wire the model provider so agent retrieval resolves each kb's
+	// embedding model (hybrid search) and the node's rerank model.
+	agenttool.SetRetrievalModelProvider(modelProviderService)
 	common.Info("agent: retrieval service adapter installed")
 
 	// Initialize handler layer
--- a/internal/agent/component/begin.go
+++ b/internal/agent/component/begin.go
@@ -115,6 +115,10 @@ func (b *BeginComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[str
 	mapsCopy(out, inputs)
 	if fallbackKey != "" {
 		out[fallbackKey] = fallbackValue
+		// Persist into the reserved "begin" namespace the same way
+		// request-supplied inputs are seeded, so {begin@<key>} refs
+		// resolve even when the DSL node id is e.g. begin_0.
+		state.SetVar("begin", fallbackKey, fallbackValue)
 	}
 	return out, nil
 }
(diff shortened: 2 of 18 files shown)


wangq8 (commented):
Three well-scoped fixes that collectively make Go-mode agent canvas usable. Good work.

**Fix 1 — Agent retrieval returned no chunks:**
The root cause is clear: without resolving the KB's embedding model, the nlp request ran keyword-only and results never passed the similarity threshold. Wiring `RetrievalModelProvider` via injection (cleanly avoiding import cycles) and mounting embedding + rerank models on the request enables true hybrid search.

**Fix 2 — KB selector empty dropdown:**
The combination of server-side pagination + client-side `chunk_count > 0` filter could produce an empty page that never triggers the next-page load. Auto-fetching the next page while results are empty and more pages exist is a pragmatic fix. The icon rendering fix is also correct.

**Fix 3 — Begin custom inputs dropped:**
Threading `beginInputs` through `RunAgent` and seeding them as Begin outputs before the first component runs correctly mirrors Python's behavior. The `singleDeclaredInputKey` + `beginBucketEmpty` guards ensure the `sys.query` fallback only fires when appropriate.

**Minor suggestions:**
1. `seedBeginInputOutputs` iterates `beginInputs` but does not distinguish between known fields and unknown keys. Consider warning on unrecognized keys for debuggability.
2. The `EmptyResultFillsEmptyResponse` test uses an empty query — consider adding a test where the query matches but the similarity threshold filters everything (more representative of the actual bug scenario).

Approved — these are solid fixes.

Haruko386 (commented):
LGTM, but would you fix the issues raised by the rabbit? @euvre

wangq8:
@euvre resolve conflict.

PR #17400, opened 2026-07-27
fix: persist pipeline tree graph rows

buua436:

Pipeline tree compilation only persisted a graph blob, while the structure graph API reads entity and relation rows. As a result, successfully generated tree data was returned as an empty template.

This PR persists Pipeline tree entities and child relations as independent structure rows, while retaining the compact graph blob for template discovery. Existing rows for the same document and template are removed before re-insertion to avoid stale data after recompilation.

### Type of change


wangq8 (commented):
Thanks for this substantial fix! The approach of building a table-of-contents tree via LLM and persisting entity+relation rows is the right architecture — it directly addresses the root cause where the structure graph API received empty templates because only a graph blob was persisted.

**Strengths:**
- Smart two-pass strategy: single-pass tree build for short docs, progressive partition+merge for long docs — keeps LLM calls cost-effective
- Proper deletion of old rows before re-insertion avoids stale data after recompilation
- Integration into `runner.py` is clean: page_index templates are skipped in the per-chunk loop and handled by the tree builder after all batches are collected
- `_struct_upsert_tree_graph_rows` correctly persists entities and relations as independent structure rows, matching what the structure graph API expects

**Bug found — relations get empty `chunk_ids` in `_tree_to_es_docs`:**

In `_tree_to_es_docs`, all entities come before all relations in the combined list:
```python
all_payloads: list[dict] = entities + relations
```

The `ei` counter tracks position into `subtree_chunk_ids` and is incremented for each entity. But relations are processed AFTER all entities, so by the time the first relation is reached, `ei` already equals `len(subtree_chunk_ids)`, and the bounds check falls back to `[]`:
```python
chunk_ids = subtree_chunk_ids[ei] if ei < len(subtree_chunk_ids) else []
```

This means EVERY relation doc gets an empty `source_chunk_ids` list. The downstream impact is that relations will lack chunk-level provenance — the structure graph API may not be able to trace which chunks a parent→child relation covers.

**Suggested fix:** interleave entities and relations in DFS order (as `_walk` emits them) rather than concatenating the two lists, OR track entity→chunk_ids in a separate lookup map keyed by entity title/position.

**Other notes:**
1. `_collect_subtree_ids` modifies `subtree_chunk_ids` as a side effect while also returning values — this dual purpose is fragile. Consider separating collection from assembly.
2. The progressive build accumulates `all_chunks` in `runner.py` for the entire document. For very large documents this could be memory-intensive; consider streaming or capping.
3. The tree merge prompt is English-only — multilingual documents may get lower-quality merged titles/summaries.

Overall the approach is solid and this definitely fixes the reported issue, but the chunk_ids bug in relation docs should be addressed before merging.

wangq8 (commented):
## PR Review: #17400 — fix: persist pipeline tree graph rows

Thanks for the fix! This addresses the real issue where Pipeline tree compilation only saved a graph blob but not the per-entity/per-relation rows that the structure graph API consumes. The approach (delete-then-reinsert for idempotency) is sound.

### What looks good ✅
- `_struct_upsert_tree_graph_rows` correctly mirrors the pattern used by other compilations
- `rewrite_duplicate_tree_names` is a nice UX improvement
- The dynamic token-budget batching in `runner.py` is a sensible optimization
- UI badges for entity types in the page_index tree view are clean

### Items to address

**1. Removal of `batch_size_cap=1` for `page_index` templates (unrelated behavioral change)**

The PR drops `batch_size_cap=1 if ... page_index else None` in `compile_structure_from_text`. Previously page_index was strictly single-chunk-per-batch; now it gets dynamically packed alongside other templates. This is a significant change to extraction behavior — multi-chunk batches may affect heading extraction quality — and it is not mentioned in the PR description. If intentional, please justify in the PR body. If accidental, restore the cap or split into a separate PR.

**2. Chunk ID fallback is overly generous**

```python
def _struct_payload_chunk_ids(payload, batch_ids):
    ...
    return selected or list(batch_ids)  # attributes entity to *all* chunks
```

When the model returns no valid chunk IDs, the entity is attributed to every chunk in the batch. A no-evidence entity should probably have an empty `source_chunk_ids` list rather than be incorrectly linked to all chunks. Consider `return selected` (allowing empty) or at minimum log a warning.

**3. `must_not` deletion when `compilation_template_id` is None**

```python
template_filter = (
    {"compilation_template_ids": [compilation_template_id]}
    if compilation_template_id
    else {"must_not": {"exists": "compilation_template_ids"}}  # overly broad?
)
```

When `compilation_template_id` is None, the delete condition removes *all* tree rows for the document that lack `compilation_template_ids`. If there are other legitimate tree compilations without template IDs, they would be silently wiped. Is the None path actually reachable? If so, consider a more targeted filter, or add a comment explaining why this is safe.

**4. Missing tests**

The new `_struct_upsert_tree_graph_rows` function and the dynamic batching logic in `_submit_batches` have no tests. Even a basic unit test for the delete-then-insert cycle and the chunk ID filtering would help prevent regressions.

**5. Minor: `source_chunk_ids` always empty for tree relations**

```python
source_chunk_ids = payload.get("source_chunk_ids") or [] if kind == "entity" else []
```

Tree relations are always stored with an empty `source_chunk_ids`. This is fine if intentional (relations do not map to specific chunks), but the `source_chunk_ids` field was added to the relation skeleton in the prompt too — if the model returns them, they are silently discarded. Either drop the field from the relation skeleton or preserve it.

### Summary

The core fix is correct and well-structured. The main concern is the unrelated removal of `batch_size_cap=1` for page_index, which should either be justified or split out. The other items are edge-case hardening. Approving conditionally on addressing item 1.

PR #17401, opened 2026-07-27
fix: refresh memory sidebar avatar immediately after update

euvre:
### Summary

After uploading a new avatar on the memory configuration page and saving, the avatar shown in the memory detail sidebar was not updated until a full page refresh.

Root cause: a react-query key mismatch. The sidebar and the configuration page read data via `useFetchMemoryBaseConfiguration`, whose query key is `['fetchMemoryDetail', ...]` (`GET /api/v1/memories/:id/config`), but `useUpdateMemory` only invalidated the unrelated `['memoryDetail', id]` key on success. As a result, the sidebar's cached query was never refetched after an update.

This PR additionally invalidates the `fetchMemoryDetail` query key in `useUpdateMemory`'s `onSuccess`, so the sidebar refetches and shows the new avatar immediately. It also removes a stale comment in the sidebar referring to an unimplemented `refreshCount` mechanism.

### DIFF
--- a/web/src/pages/memories/hooks.ts
+++ b/web/src/pages/memories/hooks.ts
@@ -18,6 +18,7 @@ import { omit } from 'lodash';
 import { useCallback, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams, useSearchParams } from 'react-router';
+import { MemoryApiAction } from '../memory/constant';
 import {
   CreateMemoryResponse,
   DeleteMemoryProps,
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Simple, correct fix. The react-query key mismatch (`memoryDetail` vs `fetchMemoryDetail`) is a textbook stale-cache bug — invalidating the correct key resolves it cleanly.

Also good housekeeping: removing the stale `refreshCount` comment in the sidebar component.

Approved.

PR #17403, opened 2026-07-27
WIP (review after ingestion fully ready) Go: reject canceling documents whose parse is not running

euvre:
### Summary

In Go mode, selecting parsed (DONE) documents in a dataset and clicking Cancel flipped the documents to CANCEL status, because `CancelDocParse` called `RequestStop` on the terminal ingestion task, which is a no-op success, and then unconditionally marked the document as CANCEL. The Python backend rejects this case with `Cannot cancel a task that is not in RUNNING status` and leaves the document untouched.

This PR aligns the Go behavior with Python:

- `CancelDocParse` now rejects with a new `ErrDocumentParseNotRunning` sentinel when the document run status is neither RUNNING nor CANCEL and it has no unfinished ingestion task (CREATED/RUNNING/STOPPING).
- The ingest API (`POST /api/v1/documents/ingest`, used by the web UI Cancel button) maps the sentinel to `Cannot cancel a task that is not in RUNNING status` (data error), so finished documents keep their DONE status.
- The stop-parse API (`POST /datasets/<id>/documents/stop`) maps it to `Can't stop parsing document that has not started or already completed`, matching the Python message.

Added tests: completed document rejected by stop API and by ingest cancel (status stays DONE), and canceling a running document still succeeds.

### DIFF
--- a/internal/service/document/document_ingest.go
+++ b/internal/service/document/document_ingest.go
@@ -2,6 +2,7 @@ package document
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"ragflow/internal/dao"
 	"ragflow/internal/service"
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Clean fix that aligns Go behavior with Python. Rejecting cancellation of already-completed documents prevents them from being incorrectly flipped to CANCEL status.

**What works well:**
- `ErrDocumentParseNotRunning` sentinel is idiomatic Go and properly mapped to user-facing messages in both the ingest API ("Cannot cancel a task that is not in RUNNING status") and the stop-parse API ("Can't stop parsing document that has not started or already completed")
- Guard condition correctly checks doc.Run and unfinished ingestion task status
- Test coverage is thorough: completed doc rejected by stop, completed doc rejected by ingest cancel, running doc cancel still succeeds — all verifying the doc status is preserved correctly

No issues found. LGTM, approved.

PR #17404, opened 2026-07-27
feat(go): implement memory extraction task consumer

euvre:
### Summary

In Go mode, memory messages were always persisted with `message_type: "raw"` even when the memory was configured with additional types (semantic / episodic / procedural). The configured types were saved correctly, but the extraction task enqueued on the `te.0.common` Redis stream had no consumer on the Go side — the extractor only existed in the Python task executor — so extracted memories were never generated and the message list only ever showed raw entries.

This PR ports the Python task-executor memory path to Go:

- Add `MemoryMessageService.StartTaskConsumer`, draining `task_type="memory"` messages from the `te.0.common` Redis stream using the same consumer group as the Python executor (`rag_flow_svr_task_broker`), and start it alongside the ingestor process.
- `HandleSaveToMemoryTask` validates the task row, skips raw-only memories, runs LLM extraction with the memory's configured system/user prompts and chat model, and persists the extracted messages with `source_id` pointing at the raw message so the message list aggregates them under `extract`.
- Report task progress (0.15 / 0.35 / 0.5 / 1.0, -1 on failure) via the new `TaskDAO.UpdateProgress`.
- Refactor `embedAndSave` into batch `embedAndSaveMessages` shared by the raw-message path and the extractor; remove the obsolete `ErrEmbedderNotWired` stub and its test.
- Add `PromptAssembler.AssembleUserPrompt` for the default extraction user prompt.


wangq8 (commented):
Thanks for this comprehensive port of the memory extraction path to Go. This is a significant feature addition that closes the gap where Go-mode memories only had raw entries.

**Strengths:**
- Clean separation of concerns: `StartTaskConsumer` handles the Redis loop, `HandleSaveToMemoryTask` validates, `saveExtractedToMemory` + `extractByLLM` do the real work
- Proper use of the shared consumer group `rag_flow_svr_task_broker` — ensures Go and Python executors never double-process the same stream entry
- Refactoring `embedAndSave` → `embedAndSaveMessages` enables both raw and extracted message paths to share the same batch embedding logic
- Good defensive coding: `payloadInt64`/`payloadString` helpers, `formatMemoryTime` multi-format parsing, JSON extraction from LLM responses
- Removing `ErrEmbedderNotWired` and its test is appropriate — the feature is now complete

**Issues to consider:**

1. **Graceful shutdown of memory consumer:** `StartTaskConsumer` exits when the context is cancelled, but there is no drain of in-flight messages. If a task is mid-extraction when shutdown arrives, the extracted messages may be partially persisted but the task progress is left at 0.35–0.5. Consider short-circuiting the context check inside `saveExtractedToMemory` before persisting.

2. **Empty response from chat model:** `extractByLLM` returns `errors.New("empty response from chat model")` when `resp.Answer == nil`. This is fine, but the task will be marked as failed (-1). Consider adding a retry or at least including the raw LLM response in the error message for debugging.

3. **LLM ID fallback ordering:** The comment says "Python prefers tenant_llm_id and falls back to llm_id" but the code actually only uses `mem.LLMID` as fallback — this matches the comment's intent, just making sure this is what was intended since Python may have additional fallback paths.

Overall this is solid work. Approved with the suggestions above as non-blocking.

internal/service/memory_extractor.go
```diff
@@ -0,0 +1,355 @@
...
+		"message_type":           item.MessageType,
+		"message_type_kwd":       item.MessageType,
+		"source_id":              sourceID,
+		"memory_id":              memoryID,
+		"user_id":                msg.UserID,
+		"agent_id":               msg.AgentID,
+		"session_id":             msg.SessionID,
+		"content":                item.Content,
+		"content_ltks":           item.Content,
+		"tokenized_content_ltks": item.Content,
```
Lynn-Inf: Better not to manually set `content`, `content_ltks`, or `tokenized_content_ltks` — the Python implementation handles tokenization internally (Infinity on save, Elasticsearch before write). The Go implementation should follow the same pattern: decide whether to invoke the tokenizer based on the underlying storage middleware, rather than populating `tokenized_content_ltks` at the message‑extract stage.

Lynn-Inf (commented):
One inline comment on `tokenized_content_ltks`. Otherwise LGTM.

wangq8:
@euvre resolve the conflict

wangq8:
@euvre resolve the conflict

wangq8:
@euvre resolve the conflict

Lynn-Inf (approved):
LGTM, @euvre resolve conflicts before merging.

wangq8:
@euvre fix build issue.

PR #17407, opened 2026-07-27
fix(web): correct dataset configuration page title and description

euvre:
### Summary

The dataset (knowledge base) configuration page showed incorrect copy:

- The page title was mistranslated as '下个配置' (zh) / 'Next configuration' (en). It is now '知识库配置' / 'Dataset configuration'.
- The page description wrongly referred to memory configuration ('在这里更新您的记忆配置，特别是大语言模型和提示词。'). It now correctly reads '在这里更新您的知识库配置，特别是大语言模型和提示词。'

Since the memory setting page shared the same `knowledgeConfiguration.titleDescription` key (where the memory wording was actually correct), a dedicated `memory.config.titleDescription` key (zh/en) was added and the memory setting page now uses it, so both pages show the correct text.

### DIFF
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -350,6 +350,8 @@ Procedural Memory: Learned skills, habits, and automated procedures.`,
         action: 'Action',
       },
       config: {
+        titleDescription:
+          'Update your memory configuration here, particularly the LLM and prompts.',
         descriptionPlaceholder: 'Describe your memory',
         memorySizeTooltip: `Accounts for each message's content + its embedding vector (≈ Content + Dimensions × 8 Bytes).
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
LGTM. Straightforward i18n fix.

Good call splitting the title-description key: the dataset config page and memory config page now each have their own `knowledgeConfiguration.titleDescription` and `memory.config.titleDescription` entries, so future edits to one won't accidentally affect the other.

The zh translations are correct: 知识库配置 for dataset and 记忆配置 for memory. The indentation-only changes in `memory-setting/index.tsx` are noisy but not harmful — consider running a formatter separately to keep future diffs minimal.

PR #17408, opened 2026-07-27
WIP (review after ingestion fully ready) Fix(ingestion): allow canonical file-family keys in Parser component params

euvre:
### Summary

CleanComponentParams dropped any param key not present in the DSL template ParamsDefaults, but the Parser component uses a per-filetype setups table whose keys are file families (pdf, docx, slides). Users can enable any family from the dataset settings UI even when the DSL template does not bake a default for it, so those keys were wrongly stripped during config cleaning.

This PR validates Parser param keys against the canonical family set (schema.ParserParam defaults) instead of the DSL template defaults, and makes BuildParserConfig treat the incoming family set as authoritative:
- Removed families are not resurrected from DSL defaults.
- Added families survive even without a DSL default.
- Per-family order_index is preserved.

Tests added:
- TestCleanComponentParams_ParserAcceptsUnknownFamilyKey - canonical family keys (e.g. slides) without a DSL default are kept.
- TestCleanComponentParams_ParserStillDropsUnknownParam - non-family keys on a Parser component are still dropped.
- TestBuildParserConfig_ParserFamilySetIsAuthoritative - replacing pdf with slides keeps the new family, drops the old one, and preserves order_index.
- TestBuildParserConfig_ParserWithoutOverrideKeepsDefaults - DSL defaults still apply when no Parser override is provided.

### DIFF
--- a/internal/ingestion/pipeline/pipeline_params.go
+++ b/internal/ingestion/pipeline/pipeline_params.go
@@ -6,15 +6,40 @@ import (
 
 	"ragflow/internal/common"
 	"ragflow/internal/entity"
+	"ragflow/internal/ingestion/component/schema"
 
 	"go.uber.org/zap"
 )
+
+// componentNameParser is the DSL component_name of the Parser component.
+// Unlike scalar-param components, its params map is a dynamic per-filetype
+// setups table keyed by file family (pdf, docx, slides, ...).
+const componentNameParser = "Parser"
+
+// parserFileFamilies is the canonical set of file-family keys the Parser
+// component accepts in its setups table. Built-in DSL templates only bake
+// the subset of families each pipeline enables by default, but users can
+// enable any family from the dataset settings UI, so family keys are
+// validated against this set instead of the DSL defaults.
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Solid fix with excellent test coverage.

The core insight — that Parser params are keyed by file families rather than DSL-baked defaults — is correct, and the solution is well-structured:

- `parserFileFamilies` initialized from `schema.ParserParam` provides the canonical key set.
- `CleanComponentParams` now special-cases Parser components to accept canonical family keys even when absent from DSL defaults.
- `BuildParserConfig` + `buildParserComponentConfig` correctly make the incoming family set authoritative: removed families stay removed, added families survive, and per-family fields merge with DSL defaults.
- Four well-targeted tests cover the key behaviors.

One small suggestion: `buildParserComponentConfig` could benefit from a brief comment explaining why non-family defaults are carried forward while family defaults are dropped when absent from overrides — the intent is clear from the code but a sentence would help future readers.

PR #17409, opened 2026-07-27
Fix: go back to previous page when last card on the last page is deleted

euvre:
### Summary

On the dataset / chat / search / agent / memory card list pages, pagination state is stored in the URL (`?page=N`). After deleting the last card on the last page, the delete mutation only invalidated the current page query, so the now-empty page was refetched while the URL page number stayed unchanged. The list then fell into the "not created" empty state, and the content only reappeared after a manual refresh (with the page size reset to the default).

This PR adds a shared `useGoToPreviousPageOnEmpty` hook in `web/src/hooks/logic-hooks.ts`: when the loaded list is empty and the current page is greater than 1, it automatically navigates back to the previous page while preserving the current page size. The hook is wired into all five card list pages (datasets, chats, searches, agents, memories).


wangq8 (commented):
LGTM. Simple and effective UX fix.

The `useGoToPreviousPageOnEmpty` hook is well-scoped and cleanly integrated across all five card list pages. The bail-out conditions (`!loading`, `listLength === 0`, `current > 1`) prevent false positives during initial loads.

One thing to keep an eye on: the `useEffect` dependency includes `pagination` and `setPagination` (reference types). If `useGetPaginationWithRouter` ever returns new object references on each render, this hook would fire unnecessarily. It doesn't appear to be an issue with the current implementation, but worth noting for future refactors.

dcc123456 (approved):

PR #17410, opened 2026-07-27
Fix: agent list owner filter not applied

euvre:
### Summary

Filtering agents by owner (team member) on the Agents page had no effect: selecting a teammate (e.g. `zzz`) still returned all of the caller's own agents. Both the Python and Go backends produced a flawed visibility predicate.

**Root cause:**

- **Python** (`api/db/services/canvas_service.py`, `UserCanvasService.get_by_tenant_ids`): the WHERE clause was `(user_id IN (owner_ids) AND permission = 'team') OR user_id = <current user>`. The unconditional `OR user_id = <current user>` always mixed in every canvas owned by the caller, regardless of the requested `owner_ids`.
- **Go** (`internal/dao/user_canvas.go`, `UserCanvasDAO.ListByTenantIDs`): the DAO tried to add a guarding `user_canvas.user_id IN (owner_ids)` condition via `Where(condDB, "user_canvas.user_id IN ?", ownerIDs)`. However, GORM's `BuildCondition` silently drops the remaining string/value arguments when the first argument is a `*gorm.DB`, so the guarding condition never made it into the SQL and the query degenerated to the same flawed predicate as Python.

Verified with a GORM DryRun:

```
OLD:   WHERE ((user_id IN (?) AND permission = ?) OR user_id = ?) AND (permission = ? OR user_id = ?)  -- guard lost
FIXED: WHERE user_id IN (?) AND (permission = ? OR user_id = ?)
```

**Fix:** both backends now use the single predicate `user_id IN (owner_ids) AND (permission = 'team' OR user_id = <current user>)`:

- When no `owner_ids` are passed, the effective set is all authorized tenants plus the caller, so behavior is identical to before.
- When specific owners are selected, results are correctly restricted to those owners (their team-shared canvases; the caller's own canvases only when the caller is among the selected owners).

Also deduplicated the repeated predicate in the Python keywords / non-keywords branches into a single `owner_filter` expression.

**Validation:** `ruff check` / `ruff format --check` pass; `go vet ./internal/dao/` passes; SQL semantics verified via GORM DryRun.

### DIFF
--- a/api/db/services/canvas_service.py
+++ b/api/db/services/canvas_service.py
@@ -148,21 +148,18 @@ class UserCanvasService(CommonService):
             cls.model.canvas_category,
             cls.model.tags,
         ]
+        owner_filter = cls.model.user_id.in_(joined_tenant_ids) & ((cls.model.permission == TenantPermission.TEAM.value) | (cls.model.user_id == user_id))
         if keywords:
             agents = (
                 cls.model.select(*fields)
                 .join(User, on=(cls.model.user_id == User.id))
                 .where(
-                    (((cls.model.user_id.in_(joined_tenant_ids)) & (cls.model.permission == TenantPermission.TEAM.value)) | (cls.model.user_id == user_id)),
+                    owner_filter,
                     (fn.LOWER(cls.model.title).contains(keywords.lower())),
                 )
             )
         else:
-            agents = (
-                cls.model.select(*fields)
-                .join(User, on=(cls.model.user_id == User.id))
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Clean fix with a well-documented root cause analysis.

The Python side deduplication via `owner_filter` is tidy and the Go fix correctly splits the `IN` clause into its own `Where` call to work around GORM's `BuildCondition` behavior. The new predicate `user_id IN (owner_ids) AND (permission = team OR user_id = current_user)` is semantically correct for both backends.

One minor observation: the Go SELECT column indentation change (tabs → mixed) is purely cosmetic and could have been a separate commit, but it doesn't affect functionality.

Issue #17284, opened 2026-07-23
[Task] Add token-usage in response for go model provider


Haruko386:
#### Todo list
- [x] OpenAI #17698 #17787
- [x] Anthropic #17775
- [x] Google Gemini #17775
- [x] xAI #17700
- [x] DeepSeek #17289 #17787
- [x] ZhipuAI #17289  #17700
- [x] Moonshot #17413 #17698 #17787
- [x] Minimax #17413 #17698 #17787
- [x] Aliyun qwen #17423 #17696
- [x] VolcEngine  #17423 #17700
- [x] HuggingFace #17697
- [x] Nivida #17643 #17787
- [x] OpenRouter #17509 #17698
- [x] SiliconFlow  #17787
- [x] Gitee #17509 #17697
- [x] Ollama #17698
- [x] vllm #17700
- [x] LM-studio #17697
- [x] Baidu #17413 #17696
- [ ] ~~FishAudio~~
- [x] localai #17697
- [x] baichuan #17455 #17696
- [x] jina  #17509 #17697
- [x] mistral #17413 #17698
- [x] stepfun #17521 #17699
- [x] cohere #17455
- [x] togetherAI #17699
- [x] Upstage #17699
- [x] novitaai #17455 #17698
- [ ] replicate
- [ ] Voyage-ai
- [x] Deep-infra #17696
- [x] LongCat #17480 #17558 #17787
- [x] minerU
- [ ] paddle ocr
- [x] model-scope #17698
- [x] Xinference #17700
- [x] azure-openai #17696
- [x] Xunfei Spark #17700
- [x] Cometapi #17696
- [x] GPUStack #17697
- [x] 302.ai #17696
- [ ] DeerAPI
- [x] jiekou.ai #17509 #17697
- [ ] Open data loader
- [x] n1n.ai #17698
- [x] astraflow #17696
- [x] avian #17696
- [x] ragcon #17699
- [x] OrcaRouter #17455 #17699
- [x] perplexity #17699
- [x] FuturMix #17696
- [x] TokenHub #17699
- [x] PPIO #17699
- [x] HunYuan #17509 #17697
- [x] Groq #17574 #17787
- [x] TokenPony #17699
- [x] bedrock #17775
- [ ] Google Vertex
- [ ] DataBricks
- [ ] Azure Anthropic
- [ ] MLX Server
- [ ] docling
- [ ] unstructured.io
- [x] xiaomi  #17700
- [x] Huawei #17423 #17697
[...]

PR #17413, opened 2026-07-27
Go: add token usage for baidu, minimax, moonshot and mistral

Haruko386:
### Summary

As title, related to #17284


Haruko386:
got some problem with `Minimax`, I shall fix it later

PR #17415, opened 2026-07-27
fix: show resolved embedding model display name in dataset list

euvre:
### Summary

The dataset list API (`GET /api/v1/datasets`) returned the raw `knowledgebase.embd_id` value. When a dataset's embedding model is stored as a tenant_model reference, this value is a 32-char hex tenant_model ID, so the knowledge base selector in the UI displayed an unreadable ID (e.g. `e21f824a8f8749b88d5a4d7d50eabe`) instead of the concrete model name.

This PR mirrors the memory list behavior: the tenant_model ID is resolved to its `modelName@instance@provider` display name (e.g. `BAAI/bge-m3@s-1@SILICONFLOW`) when building the dataset list response.

Changes:
- Export `ResolveTenantModelDisplayName` in `internal/service/memory.go` so both the memory list and the dataset list share a single resolution path.
- Add `TenantEmbdID` to `KnowledgebaseListItem` and select `knowledgebase.tenant_embd_id` in `GetByTenantIDs`.
- In `ListDatasets`, resolve `embedding_model` to the display name with a per-request cache, falling back to the raw value when resolution fails or the value is already a `name@provider` string; a hex `embd_id` without `tenant_embd_id` is also resolved directly.

### DIFF
--- a/internal/dao/kb.go
+++ b/internal/dao/kb.go
@@ -182,6 +182,7 @@ func (dao *KnowledgebaseDAO) GetByTenantIDs(ctx context.Context, db *gorm.DB, te
 			knowledgebase.language, knowledgebase.description, knowledgebase.tenant_id,
 			knowledgebase.permission, knowledgebase.doc_num, knowledgebase.token_num,
 			knowledgebase.chunk_num, knowledgebase.parser_id, knowledgebase.embd_id,
+			knowledgebase.tenant_embd_id,
 			user.nickname, user.avatar as tenant_avatar, knowledgebase.update_time`).
 		Joins("LEFT JOIN user ON knowledgebase.tenant_id = user.id").
 		Where("((knowledgebase.tenant_id IN ? AND knowledgebase.permission = ?) OR knowledgebase.tenant_id = ?) AND knowledgebase.status = ?",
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
LGTM. This correctly resolves tenant model IDs to human-readable display names in the dataset list, matching the existing memory list behavior.

**What works well:**
- `ResolveTenantModelDisplayName` is properly exported (uppercase) and shared between both the memory list and dataset list, avoiding code duplication.
- Per-request `modelNameCache` avoids redundant DB lookups when multiple datasets share the same embedding model — same pattern already proven in `ListMemories`.
- `TenantEmbdID` is added to both the DAO query and the entity struct — correct schema alignment.
- Edge cases handled:
  - `kb.TenantEmbdID` is nil → `ptrStringValue` returns `""`, then falls through to hex ID fallback
  - `kb.EmbdID` is a 32-char hex without a separate `tenant_embd_id` → direct resolution attempt via `isHexID`
  - Resolution fails → falls back to raw `embd_id` value (graceful degradation)
  - Value already in `name@provider` format → `ResolveTenantModelDisplayName` returns it as-is (early return on `@` check)

No regressions, no new issues introduced. ✅

Lynn-Inf (approved):
LGTM

PR #17416, opened 2026-07-27
Fix: use independent fixtures in ClampsOverlappedPct test

xugangqiang:
## Problem
`TestMergeByTokenSizeFromJSON_ClampsOverlappedPct` reused a single `items` fixture across four calls to `mergeByTokenSizeFromJSON`.

`mergeByTokenSizeFromJSON` mutates its `perItem` argument in place and returns the same backing array (token.go: `perItem[idx] = merged`). As a result:
- The 2nd and later calls merged already-merged chunks instead of the original input.
- Because the returned slice aliases the input, later calls silently overwrote the earlier results (`at100`, `at0`, ...).
- `reflect.DeepEqual(at100, at150)` was therefore vacuously true — the test was a **false positive** that never actually exercised the clamp. It would still pass even if the clamp were broken.

This is exactly the defect flagged in the review comment on PR #17396.

## Fix
Add a `clampOverlapFixture()` factory and build a fresh fixture for every call, so that 150 / 1e300 / -5 / -1e300 are applied to the original input.

## Verification
- The test passes after the fix.
- When the clamp logic was temporarily disabled, the test **failed** (panic at token.go:768 — the negative index produced by an out-of-range pct), proving the fixed test is no longer a false positive and can catch a clamp regression.

## Scope
Test file only. No production code change (verified `token.go` is identical to `upstream/main`).

Refs: review comment on PR #17396.

### DIFF
--- a/internal/ingestion/component/chunker/token_batch1_test.go
+++ b/internal/ingestion/component/chunker/token_batch1_test.go
@@ -81,20 +81,36 @@ func TestMergeByTokenSizeFromJSON_OverlapStripsTags(t *testing.T) {
 // threshold. Out-of-range values must not panic and must behave identically to
 // their clamped-in-range equivalent (150 == 100, -5 == 0, and the same for
 // huge magnitudes that would otherwise overflow the float->int slice index).
-func TestMergeByTokenSizeFromJSON_ClampsOverlappedPct(t *testing.T) {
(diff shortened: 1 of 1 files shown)


xugangqiang:
@yingfeng @yuzhichang @JinHai-CN  Pls help to review.

PR #17420, opened 2026-07-27
Fix agentbot embedded chat streaming envelope

euvre:
### Summary

The embedded/shared agent chat widget (agent "Embed into website") could not hold a conversation: messages stayed stuck on "Thinking"/"searching" with an empty answer.

**Root cause:** the embedded agent chat page (`web/src/pages/agent/share`) posts to `/api/v1/agentbots/<id>/completions` and parses the SSE stream with the same `use-send-message.ts` parser as the in-app agent chat, which expects the agent canvas event envelope (`{event, message_id, session_id, task_id, data}`) terminated by `data:[DONE]`. The Go handler instead re-wrapped canvas events into the legacy chatbot envelope (`{code, data:{answer}}`) and dropped `node_*`/`message_end` events, so the widget never received `message`/`message_end` frames.

**Fix:** `AgentbotCompletion` now forwards `canvas.RunEvent` frames verbatim via `WriteChatbotRunEvent` — the same writer the in-app agent chat completions path uses — emits `data:[DONE]` at stream end, and removes the now-unused `WriteAgentbotFrame`/`AgentbotSSEFrame` alias. This matches the Python `canvas_service.completion` wire shape byte-for-byte. Handler tests updated to assert the agent envelope and to accept both `data:` and `data: ` SSE prefixes.

### DIFF
--- a/internal/handler/bot.go
+++ b/internal/handler/bot.go
@@ -120,13 +120,18 @@ func (h *BotHandler) AgentbotInputs(c *gin.Context) {
 // AgentbotCompletion POST /api/v1/agentbots/<agent_id>/completions
 //
 // Mirrors python bot_api.py:157 (canvas_service.completion wrapper).
-// Streams SSE frames in the Python envelope shape. The URL-bound
-// agent_id is authoritative — the body must NOT override it.
+// The URL-bound agent_id is authoritative — the body must NOT
+// override it.
 //
-// Each canvas.RunEvent is re-formatted into the Python
-// {code, message, data} envelope: a "message" event's Data string is
-// treated as the assistant text, "message_end" terminates the
-// stream with the python completion marker.
+// The shared/embedded agent chat page (web/src/pages/agent/share)
+// parses the stream with the same use-send-message.ts parser as the
+// in-app agent chat, so frames must carry the agent canvas envelope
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
LGTM. This is a clean fix that aligns the agentbot SSE stream with the canvas event envelope expected by the shared agent chat widget.

**What works well:**
- Forwards raw `canvas.RunEvent` frames directly via `WriteChatbotRunEvent`, the same writer used by the in-app agent chat completions path — ensures byte-for-byte compatibility with the Python `canvas_service.completion` output.
- `WriteChatbotRunEvent` already handles `done` events by emitting `data:[DONE]\n\n`, so the explicit `WriteDoneFrame` call is correctly removed.
- Defensive synthetic `done` event emission when the stream ends without one — prevents hanging clients.
- Clean removal of the now-unused `WriteAgentbotFrame` / `AgentbotSSEFrame` aliases.
- Tests properly updated to assert the agent canvas envelope (`{event, message_id, session_id, data}`) instead of the legacy chatbot envelope.
- `parseBotSSEFrames` correctly handles both `data: ` and `data:` SSE prefix formats.

No regressions, no new issues introduced. ✅

PR #17421, opened 2026-07-27
fix: prevent child delimiter add from submitting form

euvre:
### Summary

Fixes the child delimiter **Add** button in the dataset configuration Token Chunker form by setting it to `type="button"`. This prevents the click from submitting the outer dataset settings form, refreshing the page, and blocking users from adding child delimiters.

Validation:
- Ran the targeted ESLint check for `web/src/pages/agent/form/token-chunker-form/index.tsx`.
- Pre-commit checks passed (web dependencies, Prettier, and ESLint).

### DIFF
--- a/web/src/pages/agent/form/token-chunker-form/index.tsx
+++ b/web/src/pages/agent/form/token-chunker-form/index.tsx
@@ -208,6 +208,7 @@ const TokenChunkerForm = ({
                 ))}
 
                 <BlockButton
+                  type="button"
                   onClick={() => childrenDelimiters.append({ value: '\n' })}
                 >
                   {t('common.add')}


wangq8 (commented):
## Review: PR #17421 — fix: prevent child delimiter add from submitting form

**Verdict: LGTM ✅**

This is a clean one-line fix. The `BlockButton` element inside the Token Chunker form (which itself lives inside a `<form>`) defaults to `type="submit"`. Adding `type="button"` prevents the click from submitting the parent dataset settings form, which was causing a page refresh and blocking users from adding child delimiters.

**Verified:**
- The `BlockButton` is indeed rendered inside a `<form>` context (the outer `FormWrapper`)
- The adjacent "Remove" `Button` already has `type="button"` — this fix brings the "Add" button in line
- No regressions possible — this only restricts default button behavior

**Details:**
- **Files changed:** 1
- **Additions:** 1, **Deletions:** 0
- **Author:** euvre

Nice catch!

PR #17422, opened 2026-07-27
fix: improve model merging logic in useModelsDerived for better instance-specific values

dcc123456:
### Summary

fix: improve model merging logic in useModelsDerived for better instance-specific values


wangq8 (commented):
## Review: PR #17422 — fix: improve model merging logic in useModelsDerived for better instance-specific values

**Verdict: LGTM ✅ — solid bugfix**

This PR fixes a real UX bug where editing an already-added model's properties (e.g. `max_tokens`) would appear to succeed but the UI never reflected the change. The root cause was two-fold:

### Problem 1: Merge order was backwards
In `useModelsDerived`, the old code let catalog entries **overwrite** instance entries when building the union map. This meant that any model already attached to the instance retained its catalog defaults regardless of user edits. The fix reverses the merge to let **instance entries win**, only using catalog entries to fill in models that haven't been added yet.

**Old:** `catalog.forEach((m) => byName.set(m.name, m))` — always overwrites  
**New:** `catalog.forEach((m) => { if (!byName.has(m.name)) byName.set(m.name, m) })` — catalog is fallback only

### Problem 2: Edit was patching the wrong data source
`handleEditSubmit` was patching the local `catalog` state via `setCatalog`, but since instance entries always won the merge (even under the old logic), catalog-only updates were **invisible** for any model that had been added. The fix patches the `instanceModels` query cache directly — the actual source of truth for attached models — so the UI reflects changes immediately.

**Additional improvements:**
- Removed `setCatalog` from the `useModelEdit` signature and `ModelsSection` — it's no longer needed, cleaner API
- The comment update in `useModelsDerived` now accurately documents the intended merge semantics

**Details:**
- **Files changed:** 2
- **Additions:** 39, **Deletions:** 18
- **Author:** dcc123456

No regressions expected. The query cache patch mirrors the same shape used elsewhere in the codebase.

PR #17423, opened 2026-07-27
fix(go-models): record provider token usage

Hz-186:
## Summary
Related to #17284 
- Parse chat, embedding, and rerank usage from provider responses
- Record usage with the correct model type even when no usage sink is provided
- Cover SiliconFlow, Aliyun, Huawei Cloud, Qiniu, and VolcEngine response formats

## Testing

- `go vet ./internal/entity/models`
- `CGO_ENABLED=0 go test ./internal/entity/models -count=1`


Haruko386 (approved):
LGTM

PR #17424, opened 2026-07-27
fix(web): identify provider instances by id and skip clean cards on save

dcc123456:
### Summary

fix(web): identify provider instances by id and skip clean cards on save


wangq8 (commented):
## Review: PR #17424 — fix(web): identify provider instances by id and skip clean cards on save

**Verdict: Approve with one suggestion ⚠️**

This is a well-structured refactor with two logical changes:

### 1. Identify instances by `id` instead of `instance_name`

Architecturally sound. Instance names can change (rename feature), but ids are stable. The backend endpoints (`GET/PUT /providers/.../instances/<instance_id_or_name>`) already support lookup by either id or name (tries `get_by_id` first, falls back to `get_by_provider_id_and_instance_name`), so this is fully compatible.

**Affected areas (consistently updated):**
- `LlmKeys.providerInstance` — query key now uses `id`
- `useFetchProviderInstance` — parameter renamed to `id`
- `useLazyInstanceDetails` — parameter renamed to `instanceId`
- `IShowProviderInstanceRequestParams` — `instance_name` → `id`
- `IUpdateProviderInstanceRequestBody` — `id` made required (was optional)
- `api.ts` — URL routes use `id` path segment
- All card components (Bedrock, SoMark, Generic) — pass `instance.id`
- `useModelMutations` — passes `id` alongside existing fields

### 2. Skip saving clean cards

The save flow now:
1. Collects dirty payloads first (`getSavePayload()` returns null for untouched cards)
2. Validates only dirty cards
3. Saves only dirty cards

This avoids unnecessary API calls for unmodified cards. The dirty check in `useInstanceSaveState` uses both `form.formState.isDirty` and a rename check (`editedNameRef !== instanceName`). Individual card components (Bedrock, SoMark) also have their own `isDirty` guard as a safety net.

### ⚠️ Suggestion

**In `index.tsx` (setting-model), the save loop iterates over `validations`:**

```typescript
for (const { ref, payload } of validations) {
```

This works correctly because `validations` was built from `dirty` entries (`{ ...e, valid: ... }`) and all dirty entries have non-null payloads by definition. However, the `if (!payload) continue` guard (line 202 in the new code) is now dead code — it was needed when the loop iterated over all refs, but now every entry in `validations` has a payload. Consider removing it for clarity, or keep it as a defensive guard. Either way is fine — not blocking.

**Details:**
- **Files changed:** 9
- **Additions:** 44, **Deletions:** 44
- **Author:** dcc123456

PR #17428, opened 2026-07-27
Fix: dataset tag file setting cannot be saved

euvre:
### Summary

In the dataset configuration page, the tag file (`tag_file_id`) selected for the Extractor component could not be saved: the update API returned success, but the value was silently dropped.

**Root cause:** `CleanComponentParams` only keeps component params that are declared in the pipeline's DSL template. The Extractor component in every builtin ingestion pipeline template did not declare `tag_file_id`, so the incoming value was discarded as an unknown param when building `parser_config`.

**Changes:**

- Add `tag_file_id` (default `""`) to the Extractor component params in all builtin pipeline templates that contain an Extractor (`internal/ingestion/pipeline/template/ingestion_pipeline_*.json`), making it a legal, persistable param.
- Add regression test `TestBuildParserConfig_BuiltinExtractorKeepsTagFileID` covering every builtin template with an Extractor, so `tag_file_id` is guaranteed to survive `BuildParserConfig`.
- Fix the tag file select display after reload: the tree select is lazy-loaded, so a saved file nested in subfolders could not resolve its title and appeared unselected. `useTagFileTree` now walks the saved file's ancestor folders (`GET /files/:id/ancestors`) and loads each level so the selected file name is displayed correctly.

**Verification:**

- `bash build.sh --test ./internal/ingestion/pipeline/...` and `./internal/service/dataset/...` pass.
- Verified end-to-end against a running server: `PUT /api/v1/datasets/{id}` with `parser_config["Extractor:AutoExtractDefault"].tag_file_id` now persists, and `GET` returns the saved value.


wangq8 (commented):
Nice root-cause analysis and well-structured fix. Here's what I reviewed:

**Template changes (all 11 Extractor templates):**
- ✓ `audio`, `book`, `email`, `general`, `laws`, `manual`, `one`, `paper`, `picture`, `presentation`, `resume` — all covered
- ✓ `tag_file_id: ""` default is consistent across all templates
- ✓ JSON is well-formed, trailing commas handled correctly

**Regression test:**
- ✓ `TestBuildParserConfig_BuiltinExtractorKeepsTagFileID` iterates all builtin templates via the registry and asserts `tag_file_id` survives `BuildParserConfig`
- ✓ Guard `checked == 0` ensures the test fails if templates are restructured without an Extractor

**Frontend changes:**
- ✓ `useTagFileTree` now accepts `value` (the saved `tag_file_id`) and resolves ancestor folders on mount so the selected file displays correctly
- ✓ `resolvedValueRef` prevents duplicate resolution on re-renders
- ✓ Refactored `toTreeNode` helper + `appendFiles` callback are clean and DRY

**Minor suggestion (non-blocking):** The `useEffect` depends on `treeData` and calls `setTreeData` within it — `resolvedValueRef` guards against loops but you could simplify by removing `treeData` from the dependency array and using a functional updater. Not a correctness issue though.

LGTM. ✓

dcc123456 (approved):

PR #17429, opened 2026-07-27
WIP (review after ingestion fully ready) Go: fix ingestion log detail missing dsl field

euvre:
### Summary

The Go `GetIngestionLog` endpoint (`GET /api/v1/datasets/{dataset_id}/ingestions/{log_id}`) returned the dataset-level field set via `datasetIngestionLogToMap`, which omits `dsl`, `pipeline_title`, `process_begin_at`, `avatar`, and `document_type`. The front-end `dataflow-result` page (opened via the "open" action on the dataset logs page) relies on `dsl.components` to render the pipeline timeline and parsed content, so the page showed empty content when opened from a log entry.

This PR returns the file-level field set (`fileIngestionLogToMap`) for the log detail response, matching the Python reference implementation (`dataset_api_service.get_log`) which explicitly includes `dsl`. Verified with `bash build.sh --test ./internal/service/dataset/...` (all package tests pass).

### DIFF
--- a/internal/service/dataset/ingestion.go
+++ b/internal/service/dataset/ingestion.go
@@ -108,7 +108,7 @@ func (d *DatasetService) GetIngestionLog(ctx context.Context, datasetID, userID,
 		return nil, common.CodeServerError, fmt.Errorf("get ingestion log: %w", err)
 	}
 
-	return datasetIngestionLogToMap(log), common.CodeSuccess, nil
+	return fileIngestionLogToMap(log), common.CodeSuccess, nil
 }
 
 func datasetIngestionLogToMap(log *entity.PipelineOperationLog) map[string]interface{} {


wangq8 (commented):
Thanks for the fix! This is a clean, targeted change.

**What I checked:**
- Verified that `fileIngestionLogToMap` (line 141) returns `dsl`, `pipeline_title`, `process_begin_at`, `document_type`, and `avatar` — all fields that `datasetIngestionLogToMap` (line 114) omits. The switch is correct for the log detail endpoint.
- The Python reference (`dataset_api_service.get_log`) also returns file-level fields including `dsl`, confirming parity.

**One thing to be aware of:** `datasetIngestionLogToMap` uses `dataset_id` as the key while `fileIngestionLogToMap` uses `kb_id`. If any consumers of this endpoint rely on the `dataset_id` field name, they may break. Doesn't seem like the case based on the described frontend usage, but worth a heads-up.

LGTM. ✓

PR #17430, opened 2026-07-27
Add {date} replacement to chat

wangq8:


wangq8 (commented):
## PR Review: Add `{date}` replacement to chat

**Author:** @wangq8 | **State:** OPEN | **+17 / −8** across 4 files

---

### Overview

This PR adds a `{date}` placeholder that gets replaced with the current UTC time (`YYYY-MM-DD HH:MM:SS`) in both the direct chat (`async_chat_solo`) and agentic chat (`async_chat`) code paths. The `date` parameter is also registered as an optional parameter in default prompt configs.

### What works well ✅

- **Both chat paths covered**: `async_chat_solo` uses a direct `.replace("{date}", sys_date)`, while `async_chat` leverages the existing `prompt_config["system"].format(**kwargs)` at line 819 — both correct.
- **No duplication**: `_apply_prompt_defaults` checks for existing `date` param before adding.
- **Frontend sync**: `use-create-chat.ts` updated to match backend defaults.
- **Low-risk change**: optional parameter, minimal lines, no breaking changes.

### Items to consider ⚠️

1. **Inconsistent guard logic in `_apply_prompt_defaults`**
   The `knowledge` parameter is only added when `"{knowledge}" in system_prompt`, but `date` is added **unconditionally** (as long as it's not already present). Consider aligning:
   ```python
   if not any(p.get("key") == "date" for p in prompt_config.get("parameters", [])) and "{date}" in prompt_config.get("system", ""):
   ```
   This avoids polluting the parameters list when `{date}` isn't actually used.

2. **`str.replace` could match partial tokens**
   `system_prompt.replace("{date}", sys_date)` will also replace `{date}` inside strings like `{dateformat}` or `{date_of_birth}`, producing e.g. `2026-07-27 17:58:00format`. Using `re.sub(r"\{date\}", sys_date, system_prompt)` would be a safer alternative if this edge case matters.

3. **Missing test for the `{date}` replacement itself**
   The existing test in `test_chats.py` only verifies the parameter list. Adding a test that actually sends a prompt with `{date}` and verifies it gets replaced with a date string would increase confidence.

4. **UTC timezone — is that intentional?**
   `datetime.now(timezone.utc)` produces UTC time. For users in other timezones (e.g. Asia/Shanghai), the displayed date will be off by up to 12 hours. This may be fine as a default, but consider whether a tenant-level or user-level timezone setting would be useful in the future.

5. **In `async_chat_solo` the `sys_date` variable is computed even for non-stream path** — this is harmless but could be moved closer to where it's used. Minor nit, not worth blocking.

### Verdict

**Approve with suggestions.** The logic is sound, both code paths are handled, and the risk is low. The `_apply_prompt_defaults` guard and the timezone question are the two items worth addressing before merge. The rest are nice-to-haves.

PR #17434, opened 2026-07-27
fix: delete navigation subtrees correctly

buua436:

Fix navigation node deletion for Elasticsearch and Infinity.

- Use `name.keyword` for exact Elasticsearch node matching.
- Recursively collect and delete the selected node’s subtree.
- Reuse existing document removal logic to update parent clusters and clean empty ancestors.
- Delete cluster rows by stable IDs to avoid name collisions.
- Serialize deletion with the dataset navigation lock.

### Type of change

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -2687,50 +2687,98 @@ async def delete_nav_node(dataset_id: str, tenant_id: str, name: str):
     index_nm, _ = pack
 
     from common.doc_store.doc_store_base import OrderByExpr
+    from rag.advanced_rag.knowlege_compile.dataset_nav import (
+        _LOCK_BLOCKING_TIMEOUT_S,
+        _LOCK_TIMEOUT_S,
+        _nav_lock_key,
+        _remove_dataset_nav_doc_locked,
+    )
+    from rag.utils.redis_conn import RedisDistributedLock
 
-    # Collect the node plus every descendant, level by level via parent_kwd.
-    names: set[str] = {name}
-    frontier: list[str] = [name]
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix — it addresses several real issues in the navigation tree deletion.

**What works well:**

1. **`name.keyword` for Elasticsearch/OpenSearch**: Correct fix. The `name` field is text-analyzed in ES, so searching by `name` would tokenize the value and potentially match partial/substring hits. Using `name.keyword` ensures exact matching. The fallback for Infinity (scalar varchar) is handled cleanly.

2. **Lock serialization**: Reusing `_nav_lock_key`, `_LOCK_TIMEOUT_S`, and `_LOCK_BLOCKING_TIMEOUT_S` from `dataset_nav.py` follows the existing pattern used by other nav operations. Good.

3. **ID-based collection instead of name-based**: Collecting descendants by stable `id` values (stored in document fields) avoids the risk of name collisions from malformed or legacy data. The `parent_kwd` traversal still uses names, which is correct since `parent_kwd` stores parent names, not IDs.

4. **Reusing `_remove_dataset_nav_doc_locked` for `nav_doc` rows**: Much better than a raw delete — this properly updates parent clusters and cleans empty ancestors.

5. **Cascade cleanup**: Cluster rows are deleted by `id` after processing all child `nav_doc` entries, which is the right order.

**One observation (non-blocking):** The `rows_by_id` filter uses `if row.get("id")`, which will skip any legacy records missing the `id` field. The `id` field has been part of nav documents for a while, so this should be fine in practice, but worth being aware of when deploying to environments with very old nav data.

Overall looks solid.

PR #17436, opened 2026-07-27
Fix: search input controls styling and restore mind map modal overlay

euvre:
### Summary

The search page's search box grew messy with long queries: since the input became an auto-expanding textarea (#17166), the action controls (clear button, text '|' divider, send button) floated at the vertical center of the tall box, and the send button was a squat pill (w-12 h-8) that looked like a toggle switch, with inconsistent spacing (gap-1 mixed with ml-4). On the search home page, the expanded textarea also overflowed the fixed-height (240px) bordered container.

Additionally, opening the mind map no longer occluded the page content as before: an inline flex spacer (`flex-1 h-[88dvh] z-30 ml-32 mt-5`) reserved space in the flex row and squeezed the search results column, so the non-modal sheet could not float over the content.

Changes:
- Pin the search input action controls to the bottom-right of the textarea on both the search home and searching pages; unify the send button into a consistent circular icon button; replace the text '|' divider with a subtle separator line; show the clear button only when there is text; remove commented-out code.
- Change the search home container from fixed `h-[240px]` to `min-h-[240px]` so the bordered box grows with the expanded textarea instead of overflowing.
- Remove the inline flex spacer around MindMapSheet so the search column keeps its full width and the (still non-modal, per #14008/#14226) mind map sheet slides in from the right and overlays the page content, covering the right side of the search input as before. The SheetContent classes are untouched, so the #14226 fix (mind map not fully displayed on initial load) is preserved.

### Type of change

### DIFF
--- a/web/src/hooks/use-auto-resize-textarea.ts
+++ b/web/src/hooks/use-auto-resize-textarea.ts
@@ -15,6 +15,12 @@ export function useAutoResizeTextarea(
     const el = ref.current;
     if (!el) return;
     el.style.height = 'auto';
+    // scrollHeight excludes borders while height is border-box, so the
+    // content box ends up a couple of pixels short and overflow-y-auto
+    // would show a scrollbar even for a single line. Hide it until the
+    // content actually exceeds the max height.
+    const overflowing = el.scrollHeight > maxHeight;
+    el.style.overflowY = overflowing ? 'auto' : 'hidden';
     el.style.height = `${Math.min(el.scrollHeight, maxHeight)}px`;
   }, [ref, value, maxHeight]);
 }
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for this fix! The changes look clean and consistent across both pages.

**search-home.tsx:**
- `h-[240px]` → `min-h-[240px]` correctly allows the container to grow with the auto-expanding textarea instead of overflowing.
- Positioning the send button at `bottom-3 right-3` (circular `size-9`) aligns with the search-view behavior — much better than vertically centering a button on a potentially multi-line textarea.

**search-view.tsx:**
- Action controls pinned to `bottom-3 right-3` — consistent and correct for an auto-expanding textarea.
- Conditionally showing the clear button (`{searchText && ...}`) is a nice UX improvement.
- The separator line (`h-4 w-px bg-border-button`) with `aria-hidden` is a clean replacement for the text pipe.
- Removing the commented-out `<Square />` — good cleanup.
- Removing the inline flex spacer wrapper around `MindMapSheet` lets the sheet overlay naturally from the right without squeezing the search results column. Since the `Sheet` is `modal={false}` and `SheetContent` classes are untouched, the #14226 fix is preserved.

LGTM, no issues found.

dcc123456 (commented):
A single line should not have a scrollbar.

PR #17437, opened 2026-07-27
fix: validate GraphRAG query rewrite keywords

roy5257:
### Summary

GraphRAG query rewriting assumes that both `answer_type_keywords` and `entities_from_query` are lists of strings. When an LLM returns a structured object inside either list, entity retrieval calls `", ".join(keywords)` and raises:

```text
TypeError: sequence item 0: expected str instance, dict found
```

This PR:

- keeps only non-empty string keywords returned by query rewriting;
- applies the five-entity limit after validation;
- falls back to the original question when no valid entity strings remain, so GraphRAG retrieval degrades safely instead of failing the chat request;
- adds regression coverage for malformed structured output and the fallback path.

### Tests

```text
pytest -q --confcutdir=test/unit_test/rag/graphrag test/unit_test/rag/graphrag
115 passed
```

The fix was also replayed against a malformed synthetic query-rewrite response in a running RAGFlow service, and the GraphRAG retrieval fallback completed without the `join()` TypeError.

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -32,6 +32,24 @@ from common import settings
 from common.doc_store.doc_store_base import OrderByExpr
 
 
+def _string_keywords(value: object, limit: int | None = None) -> list[str]:
+    if not isinstance(value, list):
+        return []
+
+    keywords = [item.strip() for item in value if isinstance(item, str) and item.strip()]
+    return keywords[:limit] if limit is not None else keywords
+
+
+def _query_rewrite_keywords(value: object) -> tuple[list[str], list[str]]:
+    if not isinstance(value, dict):
+        return [], []
+
+    return (
+        _string_keywords(value.get("answer_type_keywords", [])),
+        _string_keywords(value.get("entities_from_query", []), limit=5),
+    )
+
+
 class KGSearch(Dealer):
     async def _chat(self, llm_bdl, system, history, gen_conf):
         response = get_llm_cache(llm_bdl.llm_name, system, history, gen_conf)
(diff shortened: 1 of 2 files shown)


wangq8:
Hi @roy5257,

Starting with v0.27.0, RAGFlow is transitioning to a new knowledge compilation and Agentic RAG architecture. As a result, this code is no longer used, so we’re closing this PR.

Thank you for your contribution!

wangq8 (commented):
Thanks — this is a clean defensive fix for the `TypeError: sequence item 0: expected str instance, dict found` failure. Filtering to non-empty strings, applying the 5-entity cap *after* validation, and falling back to the original question when no valid entities remain are all correct.

A few points (none blocking):

1. **Exception-type change.** `except json_repair.JSONDecodeError` → `except json.JSONDecodeError`. `json_repair` re-exports `json.JSONDecodeError` (same class), so this is equivalent — worth a quick confirmation against the pinned `json_repair` version. Even if it diverged, `retrieval`'s `except Exception` still catches it and falls back to `[qst]`, so there's no crash risk.

2. **Fallback placement.** The new `if not ents: ents = [qst]` sits after the try/except. Correct, and `ty_kwds` may still be non-empty (type retrieval proceeds) while entities degrade to the question — that's the intended behavior. Just confirming intent.

3. **Test isolation.** `test_search_query_keywords.py` monkeypatches `rag.nlp.search.Dealer`/`index_name` at module import time via `sys.modules`, then restores them. It works but is somewhat fragile (depends on `rag.nlp.search` already being imported and mutates global state during import). Prefer scoping these stubs into fixtures/`monkeypatch` if you touch this file again.

4. `_string_keywords` correctly guards `item.strip()` with `isinstance(item, str)`, so mixed list values (`dict`, `int`, `None`) are dropped safely.

LGTM otherwise.

PR #17439, opened 2026-07-27
fix: persist page size across main page navigation

euvre:
### Summary

On list pages (datasets, chats, agents, files, search, etc.), the pagination state lived only in the URL search params (`page`/`size`). After selecting a page size in the bottom-right pagination, switching to another main page drops those params, so coming back reset the page size to the default 50/page.

This PR persists the selected page size per pathname in `localStorage` inside `useSetPaginationParams`. When the URL has no `size` param, the stored value for the current page is used instead of the hardcoded default; the page number is intentionally not remembered. All main list pages go through this hook, so one change fixes them all.

### DIFF
--- a/web/src/hooks/route-hook.ts
+++ b/web/src/hooks/route-hook.ts
@@ -69,27 +69,36 @@ export const useGetPaginationParams = () => {
   };
 };
 
+const PageSizeStorageKeyPrefix = 'RAGFlowPageSize:';
+
+const getStoredPageSize = (pathname: string) =>
+  Number(localStorage.getItem(`${PageSizeStorageKeyPrefix}${pathname}`));
+
 export const useSetPaginationParams = () => {
   const [queryParameters, setSearchParams] = useSearchParams();
-  // const newQueryParameters: URLSearchParams = useMemo(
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice UX fix — this is a clean, well-scoped change. The approach of persisting page size per pathname in localStorage is pragmatic and avoids unnecessary complexity.

**What I checked:**
- `useLocation` is already imported in the file ✅
- The fallback chain (`URL size → localStorage → default 50`) is correct ✅
- `Number(localStorage.getItem(...))` returns `0` for missing keys, which is falsy and falls through correctly ✅
- Page number is intentionally not persisted (good — returning to a default page on revisit is expected behavior) ✅
- Removed dead commented-out code ✅

**One minor consideration:**
The `localStorage` prefix key (`RAGFlowPageSize:`) + full `pathname` works well for the current set of pages (`/dataset`, `/chat`, `/agent`, `/file`, `/search`). Just something to keep in mind: if any future page has a dynamic pathname segment (e.g. `/dataset/abc123`), the storage key would be per-instance rather than per-page-type. Not a problem today — just a note for future contributors.

LGTM! 🚀

dcc123456 (approved):

PR #17442, opened 2026-07-27
Fix: code review

Lynn-Inf:
### Summary

Check memory accessable before operation, and log error when listModels failed.


wangq8 (commented):
Thanks for this PR. Overall the changes look good — access control on memory operations and proper error handling are important improvements.

**Summary of changes:**
- `memory.go (handler)`: Added `GetUser` checks for `DeleteMemory` and `GetMemoryConfig` — consistent with other handlers that already do this.
- `memory.go (service)`: Updated signatures to accept `userID` and use `requireMemoryAccess`, extracted `getMemoryConfig` for internal callers.
- `memory_message_service.go`: Switched to internal `getMemoryConfig` (no access control needed in trusted queue path).
- `model_service.go`: Added warning log when `listModels` fails — useful for debugging provider issues.
- `tenant.go`: Fixed a bug where `tenantDAO.Update` error was silently ignored.

**One small suggestion (non-blocking):**
In `GetMemoryConfig` you now split into the public `GetMemoryConfig(ctx, userID, memoryID)` and private `getMemoryConfig(ctx, memoryID)`. The access check + delegate pattern is clean. You may want to do the same for `DeleteMemory` eventually if there are internal callers that need to bypass access control (e.g. cascading deletes from tenant cleanup). Not required now though — the current internal caller (`memory_message_service`) doesn't need it.

LGTM otherwise. 👍

PR #17449, opened 2026-07-27
fix(go): send Gemini system prompts via system_instruction instead of user turn

### LINKED ISSUE #17330: [Go] Gemini: system prompt is sent as a user message instead of system_instruction
## Description
In the Go model driver for Google Gemini (`internal/entity/models/google.go`), the system message is not delivered through Gemini's native `system_instruction` channel.

In `ChatWithMessages` (~L142-149) and `ChatStreamlyWithSender` (~L218-225), the message role switch has no `system` case, so `role:"system"` falls into `default -> genai.RoleUser`. Additionally, `GenerateContent` / `GenerateContentStream` are invoked with a `nil` config (`client.Models.GenerateContent(ctx, modelName, contents, nil)`), so `system_instruction` is never set.

As a result, the system prompt text is concatenated into the first user turn and Gemini effectively treats it as ordinary user content rather than a system instruction.

## Impact
All Gemini chat and multimodal (vision) calls made through the Go path lose their system prompt. This is a regression compared with the Python `GoogleChat` driver, which correctly sets `system_instruction` (`rag/llm/chat_model.py:1374`). (The Python `GeminiCV` vision driver has the same defect, so the Go path reproduces that bug for *all* Gemini traffic, not just vision.)

## Expected behavior
Extract `role:"system"` message(s) from `messages`, pass them via `genai.GenerateContentConfig{SystemInstruction: ...}` as the 4th argument, and exclude them from `contents`.

## References
- `internal/entity/models/google.go:142-149`, `:218-225`
- Python (correct): `rag/llm/chat_model.py:1374`

abhayishere:
## Summary
- Fixes #17330: the Go Gemini driver (`internal/entity/models/google.go`) was dropping
  `role: "system"` messages into the conversation as a plain `user` turn, since the role
  switch in `googleChatContents` had no `system` case and fell through to
  `genai.RoleUser`. `GenerateContent`/`GenerateContentStream` were also always called
  with either a `nil` config or one that never set `SystemInstruction`, so there was no
  path to send a real system instruction even if one had been extracted.
- Add `googleSystemInstruction(messages)` to collect all `role: "system"` message parts
  into a single `*genai.Content`, and exclude those messages from `googleChatContents`.
- Thread the extracted system instruction through `googleGenerateContentConfig`, which
  now sets `GenerateContentConfig.SystemInstruction` and only returns `nil` when there is
  truly nothing to configure (previously it bailed out early whenever `chatModelConfig`
  was `nil`, which is the common case for calls that only carry a system + user message).
- Wire this into both `ChatWithMessages` and `ChatStreamlyWithSender`.
- This brings the Go driver in line with the Python `GoogleChat` driver
  (`rag/llm/chat_model.py:1374`), which already sets `system_instruction` correctly.

## Test plan
- [x] `internal/entity/models/google_test.go`: added
  `TestGoogleSystemInstructionExtractedFromMessages` (system message excluded from
  `contents`, routed into `SystemInstruction`) and
  `TestGoogleSystemInstructionNilWhenNoSystemMessage` (no system message → nil
  instruction, nil config when nothing else is set).
- [x] Updated the existing `TestGoogleGenerateContentConfigConvertsTools` call site for
  the new `googleGenerateContentConfig` signature.


Haruko386 (approved):
LGTM

Issue #16971, opened 2026-07-16
[Bug]: MinerU parser reparses the entire PDF for every page-splitting task, causing duplicate computation and poor performance


JiesenSun:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

cb93883

### RAGFlow image version

cb93883(v0.26.4)

### Other environment information

```Markdown
ubuntu22.04
```

### Describe the bug

When using the MinerU parser in RAGFlow, a large PDF is split into multiple page-range tasks for parallel processing.

However, each page-range task reparses the entire original PDF instead of only processing its assigned page range.

As a result:

- The same PDF is parsed repeatedly.
- OCR/layout analysis is executed multiple times on identical pages.
- CPU/GPU usage increases significantly.
- Processing time grows much longer than expected.
- Memory consumption also increases unnecessarily.

### To Reproduce

1. Configure RAGFlow to use the MinerU parser.
2. Upload a large PDF (e.g. hundreds of pages).
3. Start document parsing.
4. Check the logs.

### Expected behavior

After the PDF is split into multiple page-range tasks, each task should only process its assigned pages.

For example:

- Task 1: pages 1-50
- Task 2: pages 51-100
- Task 3: pages 101-150

Each worker should invoke MinerU only on its own page range.

### Actual behavior

Every page-range task invokes MinerU on the complete PDF.

For example:

- Task 1 parses pages 1-500
- Task 2 parses pages 1-500
- Task 3 parses pages 1-500

Although only part of the output is eventually used, the entire PDF is parsed repeatedly by every task.

### Impact

For large PDFs, this leads to:

- N× duplicated parsing work (N = number of split tasks)
- Much higher CPU/GPU utilization
- Significantly longer parsing time
- Unnecessary memory consumption

This issue becomes especially noticeable when MinerU performs GPU-based OCR and layout analysis.

### Suggested solution

Instead of passing the original PDF to every split task, each task should process only its assigned page range.

Possible approaches include:

[...]

Issue #16586, opened 2026-07-03
[Bug]: MinerU parser ignores split page ranges and duplicates chunks


ayame-selachan:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

62f94cd59b505b56df724faba4d4a2a8ce79679a

### RAGFlow image version

v0.26.2

### Other environment information

```Markdown

```

### Actual behavior

When parsing a PDF with the MinerU parser, RAGFlow may split the document into multiple page-range tasks. However, the MinerU API request appears to always parse the whole PDF because `start_page_id` and `end_page_id` are fixed to `0` and `99999`.

This causes duplicate chunks, especially when the document contains figures/images and VLM descriptions.

For a 32-page PDF, RAGFlow created multiple page-range tasks, but each task parsed the same number of blocks and images:

```text
Page(1~13): [MinerU] Parsed 405 blocks from PDF.
Page(1~13): [MinerU] Generating VLM descriptions for 11 images...
Page(1~13): Generate 27 chunks

Page(13~25): [MinerU] Parsed 405 blocks from PDF.
Page(13~25): [MinerU] Generating VLM descriptions for 11 images...
Page(13~25): Generate 27 chunks

Page(25~32): [MinerU] Parsed 405 blocks from PDF.
Page(25~32): [MinerU] Generating VLM descriptions for 11 images...
Page(25~32): Generate 27 chunks
```

The chunks generated by different page-range tasks were duplicates.

### Expected behavior

Each split task should parse only its assigned page range.

For example:

```text
Page(1~13)   -> parse pages 1-12 or 1-13, depending on RAGFlow's range convention
Page(13~25)  -> parse only that page slice
Page(25~32)  -> parse only that page slice
```

The block count and image count should not be identical across all tasks unless the document content is actually identical.

### Steps to reproduce

```Markdown
1. Deploy RAGFlow v0.26.2 with Docker Compose.

2. Configure MinerU as the PDF parser through the MinerU API server.

[...]

PR #17451, opened 2026-07-27
fix(docling): thread page_from/page_to into Docling Serve request

### LINKED ISSUE #17450: [Bug]: Docling parser ignores the task page range and converts the whole PDF on every task
### Self Checks

8b793ee14

### RAGFlow image version

v0.26.4 (bug also present on current main, 8b793ee14)

```Markdown
- Deployment: Kubernetes, RAGFlow with 8 task_executor workers
- Parser: Docling via external docling-serve (DOCLING_SERVER_URL set),
  docling-serve v1.21.0, `vlm` pipeline with a remote vLLM backend
  (granite-docling-258M served by vLLM, 4x RTX 3090)
- Document under test: 2383 pages, 22.6 MB, digitally-born PDF
```

### Actual behavior

`rag/app/naive.py::by_docling` accepts `from_page`/`to_page` and then calls
`DoclingParser.parse_pdf()` **without them**:

```python
def by_docling(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, ...):
    pdf_parser = DoclingParser()
    ...
    sections, tables = pdf_parser.parse_pdf(
        filepath=filename,
        binary=binary,
        callback=callback,
        output_dir=...,
        delete_output=...,
        docling_server_url=...,
        parse_method=parse_method,
    )   # <- from_page / to_page dropped here
```

Every sibling backend forwards its range — `by_deepdoc` (`naive.py:119`),
`by_mineru` (`naive.py:175-176`, since #16957) and `by_mistral_ocr`
(`naive.py:394-395`). Docling is the one that drops it, and
`DoclingParser` never sets `options.page_range` on the docling-serve payload.

So when RAGFlow splits a PDF into `task_page_size` tasks (default 12,
`api/db/services/task_service.py:473`), **every task converts the entire
document** and RAGFlow slices the result locally. The `Page(145~157)` prefix in
the progress log is RAGFlow's own label for the task, not what docling was asked
to convert.

This is the same bug that was reported and fixed for MinerU in #16586 / #16971 /
#16957 — it is still open for the Docling backend.

### Expected behavior

Each page-range task should ask docling-serve to convert only its own pages.
docling-serve already supports this: `ConvertDocumentsOptions.page_range`
(`docling.datamodel.service.options`), described as *"Only convert a range of
pages. The page number starts at 1."* RAGFlow just never sets it.

`DoclingParser` looks like it was written expecting a ranged response — it
already keeps `page_from`/`page_to` and offsets result page numbers by them
(`deepdoc/parser/docling_parser.py:212,286`). Only the HTTP wiring is missing.

Note that no rebasing of page numbers is required, unlike MinerU: docling
converts only the requested pages but still numbers them from the start of the
document (`docling/pipeline/base_pipeline.py` builds `Page(page_no=i + 1)` off
the full-document index).

### Steps to reproduce

```Markdown
1. Deploy docling-serve and point RAGFlow at it with DOCLING_SERVER_URL.
2. Create a knowledge base with layout_recognize = "Docling".
3. Upload any PDF longer than task_page_size (default 12) pages, so RAGFlow
   creates more than one task. The effect scales with document length, so a
   few hundred pages makes it unmissable.
4. Start parsing and watch the docling-serve logs, or capture the outgoing
   request body.

Observed: every task POSTs the whole document to /v1/convert/source with an
`options` object that has no `page_range`, and docling converts all N pages.
Each task takes as long as a full-document conversion, and each returns the
whole document's content.

Expected: the task labelled Page(145~157) sends `"page_range": [145, 157]`
and converts 13 pages.
```

### Additional information

**Impact on a vLLM-backed docling-serve deployment**

This is where the bug stops being "wasted work" and becomes "ingestion can never
complete". In our setup (docling-serve `vlm` pipeline, inference offloaded to
vLLM over HTTP):

1. **N× the GPU work.** A 2383-page document is split into ~199 tasks, and each
   one sends all 2383 pages. The vLLM backend performs ~199× the page inferences
   actually needed, and ~199/200 of that output is thrown away.

2. **Every task 504s, forever.** A whole-document conversion of our test file
   takes ~1432s. docling-serve aborts a synchronous conversion at
   `DOCLING_SERVE_MAX_SYNC_WAIT` (default 120s) and returns
   `504 Conversion is taking too long` *while the GPUs are still working on it*.
   RAGFlow fails the task and re-queues it, so the ingest sits near 0% forever
   while the GPUs stay saturated with work that is always discarded.

3. **Raising the server timeout cannot rescue it.** RAGFlow constructs
   `DoclingParser()` with no arguments (`naive.py:189`), so it takes the
   hardcoded `request_timeout: int = 600` in `DoclingParser.__init__`, and no env
   var overrides it. 600s is a hard client-side ceiling; a ~1432s conversion can
   never fit under it, no matter how docling-serve is configured.

4. **docling-serve gets OOMKilled.** A conversion holds every rendered page in
   memory for the life of the request. Whole-document requests pushed our
   docling-serve pod past 11 GiB and OOMKilled it at a 16 GiB limit (exit 137),
   independently of the timeout. This also caps concurrency: we could not raise
   `DOCLING_SERVE_ENG_LOC_NUM_WORKERS` while each request was a whole document.

5. **Duplicate chunks if a conversion ever does succeed.** On a document short
   enough to convert inside the timeout, each of the N tasks returns the *whole*
   document's content, so `document.chunk_num` ends up N× the real chunk count —
   exactly what #16586 reported for MinerU.

Measured on the 2383-page file, with and without a locally patched RAGFlow that
sends `page_range`:

| | without `page_range` | with `page_range` |
|---|---|---|
| pages converted per task | 2383 | 13 |
| time per task | ~1432s → always 504 at the 600s client cap | ~10s |
| docling-serve peak memory | 11+ GiB, OOMKilled at 16 GiB | bounded |
| ingest outcome | never completes | completes |

**There is no configuration workaround.** RAGFlow reads only
`DOCLING_SERVER_URL`, `DOCLING_API_KEY`, `DOCLING_OUTPUT_DIR`,
`DOCLING_DELETE_OUTPUT`, `DOCLING_FORMULA_ENRICHMENT` and the
`DOCLING_*_LABELS` vars — none of which is a page range or a timeout.
`parser_config.pages` only selects which pages get ingested overall; the
selected range is still split by `task_page_size` and each task still converts
everything. Raising `task_page_size` to cover the whole document (the workaround
suggested on #16971) removes the duplication but leaves one whole-document
conversion, which for a large PDF is guaranteed to exceed the hardcoded 600s
client timeout. docling-serve has no server-side page-range setting either —
`page_range` is per-request, and only RAGFlow knows which range a task owns.

A PR fixing this follows.

Elyytscha:
### Summary

`rag/app/naive.py::by_docling` accepts `from_page`/`to_page` and then calls
`DoclingParser.parse_pdf()` without them, and `DoclingParser` never sets
`options.page_range` on the docling-serve payload. So when RAGFlow splits a PDF
into `task_page_size` tasks, **every task converts the whole document** and
RAGFlow slices the result locally. The `Page(145~157)` prefix in the progress log
is RAGFlow's own task label, not what Docling was asked to convert.

Every sibling backend already forwards its range — `by_deepdoc`
(`naive.py:119`), `by_mineru` (`naive.py:175-176`, since #16957) and
`by_mistral_ocr` (`naive.py:394-395`). Docling is the one that drops it. This is
the Docling half of the same bug fixed for MinerU in #16957 (issues #16586,
#16971).

## Repro steps

1. Point RAGFlow at an external Docling Serve with `DOCLING_SERVER_URL`.
2. Create a dataset with `layout_recognize = "Docling"`.
3. Upload a PDF longer than `task_page_size` (default 12) pages, so RAGFlow
   creates more than one task.
4. Inspect the outgoing `/v1/convert/source` request, or the Docling Serve logs.

Every task posts the whole document with no `page_range`, and every task returns
the whole document's content.

## Expected

Each task converts only its own pages. Docling Serve already supports this:
`ConvertDocumentsOptions.page_range` (`docling.datamodel.service.options`),
documented as *"Only convert a range of pages. The page number starts at 1."*
RAGFlow's `from_page`/`to_page` are 0-based with `to_page` exclusive (Python
slice stop); Docling's `page_range` is 1-based with both bounds inclusive.

No page-number rebasing is needed, unlike MinerU: Docling converts only the
requested pages but still numbers them from the start of the document
(`docling/pipeline/base_pipeline.py` builds `Page(page_no=i + 1)` from the
full-document index), so positions and citations stay correct. `DoclingParser`
was evidently written expecting a ranged response — it already keeps
`page_from`/`page_to` and offsets result page numbers by them
(`docling_parser.py:212,286`); only the wiring was missing.

## Fix

- Thread `page_from`/`page_to` through `by_docling` → `parse_pdf` →
  `_parse_pdf_remote` in `rag/app/naive.py` and
  `deepdoc/parser/docling_parser.py`, clamping `to_page` to
  `MAXIMUM_PAGE_NUMBER` the way `by_mineru` does.
- Translate the range once in `DoclingParser._resolve_page_range`, which returns
  `None` when the caller did not narrow the range, so an unsplit document sends
  byte-for-byte the request it sent before. It also returns `None` for a
  degenerate range: Docling's `_validate_page_range` rejects `end < start`, which
  would otherwise 422 every payload variant and turn a no-op task into a hard
  `RuntimeError`.
- Set `options.page_range` on all four payload variants the fallback chain can
  reach (chunked/standard × v1/v1alpha).
- Set `page_range` on the local `DocumentConverter.convert()` path too.
- Log the resolved range, mirroring the MinerU fix.

`self.page_from`/`self.page_to` are deliberately left alone: the local path still
renders the whole document in `__images__`, and Docling returns absolute page
numbers, so `_make_line_tag`, `crop` and `cropout_docling_table` keep the page
arithmetic they have today.

Out of scope, mentioned for whoever hits them next: `by_opendataloader` and
`by_tcadp` have the same gap, `DoclingParser.__init__`'s hardcoded
`request_timeout: int = 600` has no env override, and the local path still
renders every page of the document per task.

This overlaps the Docling part of #16857, which also fixes OpenDataLoader wiring
and MinerU and has been open since 13 July; its MinerU half now conflicts with
the merged #16957. This PR is the Docling-only slice, rebased onto the
`page_from`/`page_to` convention that #16957 established, so it can land
independently. Happy to close this in favour of that PR if it is revived.

## Tests

Three regression tests in
`test/unit_test/deepdoc/parser/test_docling_parser_remote.py`, mirroring the ones
#16957 added for MinerU:

- the conversion table for `_resolve_page_range`, including the un-narrowed and
  degenerate cases;
- a task owning pages 145-157 puts `"page_range": [145, 157]` on every payload
  the fallback chain sends (both chunked attempts and the standard fallback);
- a whole-document task sends no `page_range` at all.

```bash
uv run pytest test/unit_test/deepdoc/parser/test_docling_parser_remote.py
```

11 passed (8 existing + 3 new). `ruff format --check` clean on the three touched
files.

Verified in production on a 2383-page / 22.6 MB PDF against Docling Serve v1.21.0
with a vLLM-backed `vlm` pipeline: each task drops from 2383 pages / ~1432s —
which always hit RAGFlow's hardcoded 600s client timeout, so the ingest could
never finish — to 13 pages / ~10s, and Docling Serve stops being OOMKilled at
16 GiB.

Fixes #17450


buua436 (commented):
The overall page-range wiring looks correct, but the local path still calls __images__() once for the full document before rendering the selected range, so the intended CPU/memory optimization is not achieved. Please remove the unconditional full-document render and handle degenerate ranges without falling back to converting the entire document.

PR #17453, opened 2026-07-28
fix: pass Infinity vector similarity weight

Hellrakk:
## Summary

- Pass `vector_similarity_weight` from Python and Go retrieval requests into Infinity's weighted fusion expression.
- Keep fusion weights ordered as text first and vector second, with the existing default vector weight of `0.3`.
- Add tests for FusionExpr construction, direct Go search propagation, top-level Go retrieval propagation, and Python retrieval propagation.

## Root Cause

Infinity's fusion expression used hard-coded weights. The configured vector similarity weight was not included in the Python retrieval request and was not propagated from Go `RetrievalRequest` to `RetrievalSearchRequest`.

## Tests

- `python -m pytest -q test/unit_test/rag/nlp/test_fusion.py test/unit_test/rag/test_search_fusion_weight.py`
- `python -m compileall -q rag/nlp/search.py rag/nlp/fusion.py test/unit_test/rag/test_search_fusion_weight.py`
- `CGO_ENABLED=0 go test ./internal/service/nlp -count=1`
- `git diff --check`

The Python and Go regression tests were also verified to fail when the corresponding propagation field is temporarily removed.

## Scope

This PR is limited to Infinity vector similarity weight propagation. It does not include Infinity score decomposition, live end-to-end testing, or the separate weight-format precision follow-up.


Hellrakk:
@coderabbitai review

Hellrakk:
@coderabbitai review

rag/nlp/search.py
```diff
@@ -207,7 +208,13 @@ async def search(self, req, idx_names: str | list[str], kb_ids: list[str], emb_m
                 if settings.DOC_ENGINE_OCEANBASE:
                     src.append(f"q_{len(q_vec)}_vec")
 
-                fusionExpr = FusionExpr("weighted_sum", topk, {"weights": "0.001,1"})
+                vector_similarity_weight = float(req.get("vector_similarity_weight", 0.3))
```
qinling0210: The code change should only when it is infinity. When using elasticsearch, keep the original hardcoded weights value.

Hellrakk: Addressed in b66631a. Dynamic fusion weights are now applied only when DOC_ENGINE_INFINITY is enabled, while Elasticsearch keeps the original "0.001,1" weights. Added a regression test for the Elasticsearch path.

qinling0210 (approved):
LGTM

internal/service/nlp/retrieval.go
```diff
@@ -656,11 +674,7 @@ func (s *RetrievalService) Search(ctx context.Context, req *RetrievalSearchReque
 			}
 
 			// Execute search with fusion
-			fusionExpr := &types.FusionExpr{
-				Method:       "weighted_sum",
-				TopN:         topk,
-				FusionParams: map[string]interface{}{"weights": "0.05,0.95"},
-			}
+			fusionExpr := buildInfinityFusionExpr(topk, req.VectorSimilarityWeight)
```
qinling0210: We should also only limit the code change to Inifnity engine in GO, right?

qinling0210 (changes requested):

internal/service/nlp/retrieval.go
Hellrakk: Yes. The Go-side dynamic fusion weight is now restricted to the Infinity engine as well.

For Elasticsearch, the original hardcoded weights (`0.05,0.95`) are preserved. I also added regression tests covering both Infinity and Elasticsearch paths, and checked the related Go retrieval code for similar issues.

Updated in commit `2708f6b`.

qinling0210 (approved):
LGTM.

qinling0210:
@Hellrakk Please solve the test failure.

rag/nlp/fusion.py
```diff
@@ -0,0 +1,27 @@
+#
```
qinling0210: We don't need to create a new file for build_fusion_expr(), put the function to search.py is ok.

qinling0210: @Hellrakk , please resolve the comment above, other is ok.

Hellrakk:
Addressed in commit fa76e2d6.

- Moved build_fusion_expr() into rag/nlp/search.py.
- Removed the standalone rag/nlp/fusion.py module.
- Kept the fusion weight tests in test/unit_test/rag/test_search_fusion_weight.py.
- Local regression: 6 passed; preflight has passed, and the full Go/Python CI matrix is still running.

qinling0210 (approved):
LGTM

PR #17454, opened 2026-07-28
fix: reset selected documents on new search in search page

euvre:
### Summary

On the search page, `selectedDocumentIds` persisted across searches. After selecting multiple documents in one search, running a new search that matched fewer documents caused the file filter to display a stale count such as `3/1`.

This PR resets `selectedDocumentIds` whenever a new question is sent, and syncs the local checkbox state in `RetrievalDocuments` with the prop so stale selections are also cleared from the popover.

### DIFF
--- a/web/src/pages/next-search/hooks.ts
+++ b/web/src/pages/next-search/hooks.ts
@@ -333,6 +333,7 @@ export const useSendQuestion = (
       if (isEmpty(q)) return;
       setIsFirstRender(false);
       setCurrentAnswer({} as IAnswer);
+      setSelectedDocumentIds([]);
       if (enableAI) {
         if (!sharedId && !searchId) {
           message.error('Search ID is required.');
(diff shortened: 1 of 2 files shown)


wangq8 (approved):
Thanks for the fix! This is clean and well-scoped.

**What it does correctly:**
- In `hooks.ts` / `sendQuestion`: resetting `selectedDocumentIds` when a new question is sent is the right place — it's the root cause fix.
- In `retrieval-documents/index.tsx`: the `useEffect` syncing local `selectedValues` from the prop is a good defensive addition, ensuring the popover checkbox state doesn't go stale.

No regressions, style issues, or security concerns. LGTM.

wangq8 (commented):
Clean, minimal fix. LGTM.

**hooks.ts:** Resetting `selectedDocumentIds` on new question send is exactly the right place — it clears the stale selection before the new search results arrive, so the retriever starts fresh.

**retrieval-documents/index.tsx:** The `useEffect` syncing `selectedValues` with the `selectedDocumentIds` prop ensures the checkbox popover state matches whenever the parent resets it. This covers both the explicit reset above and any future reset paths.

No regressions, no style issues. Approved. ✅

dcc123456 (approved):

PR #17460, opened 2026-07-28
Fix: Unable to navigate from the agent list page to the compilation editing page.

cike8899:
### Summary

Fix: Unable to navigate from the agent list page to the compilation editing page.

### DIFF
--- a/web/.oxlintrc.json
+++ b/web/.oxlintrc.json
@@ -4,24 +4,24 @@
     "es2021": true,
     "node": true
   },
-  "plugins": ["@typescript-eslint", "react", "react-hooks"],
+  "plugins": ["typescript", "react"],
   "settings": {
     "react": {
       "version": "18.2.0"
     }
   },
   "rules": {
-    "@typescript-eslint/no-use-before-define": [
+    "typescript/no-use-before-define": [
       "warn",
       {
         "functions": false,
         "variables": true
       }
     ],
-    "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/ban-ts-comment": "off",
-    "@typescript-eslint/no-empty-function": "off",
-    "@typescript-eslint/no-non-null-assertion": "off",
+    "typescript/no-explicit-any": "off",
+    "typescript/ban-ts-comment": "off",
+    "typescript/no-empty-function": "off",
+    "typescript/no-non-null-assertion": "off",
     "react/react-in-jsx-scope": "off",
     "react/no-unescaped-entities": "warn",
     "no-console": ["warn", { "allow": ["warn", "error"] }]
--- a/web/CLAUDE.md
+++ b/web/CLAUDE.md
@@ -18,7 +18,8 @@ RAGFlow frontend is a React/TypeScript application built with UmiJS:
 npm install
 npm run dev        # Development server
 npm run build      # Production build
-npm run lint       # ESLint
+npm run lint       # oxlint
+npm run format     # oxfmt
 npm run test       # Jest tests
 ```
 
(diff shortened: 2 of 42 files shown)


wangq8 (commented):
## PR Review: #17460

Thanks for the PR! The core navigation fix looks solid, but there are several concerns around the bundled lint migration, type correctness, and a UI regression. Details below.

### ✅ What Works Well

- **Unified agent+compilation list** in `agents/index.tsx`: removing the `isCompilation` toggle and mixing both card types with `x.type === CompilationGroupCategory` dispatch is the right approach and directly fixes the navigation issue.
- **Centralized `FlowTypeConfig`** in `constant.ts`: good DRY refactoring — cleanly eliminates duplicated icon maps in `agent-card.tsx` and `create-agent-form.tsx`.
- **Filtering compilation templates from data-pipeline select** (`data-pipeline-select/index.tsx`): correct guard — compilation template groups shouldn't appear in data pipeline dropdowns.
- **Filtering from home agent list** (`home/agent-list.tsx`): also correct — dashboard should only show actual agents.

### ⚠️ Major Concern: Lint Migration Bundled

This PR bundles a full **eslint → oxlint migration** (42 files, ~200+ lines changed) with a navigation bug fix. This is a separate concern and should be in its own PR. Reasons:

1. **Review noise**: the real fix is ~50 meaningful lines buried in ~350 lines of mechanical comment rewrites
2. **Reversibility**: if the oxlint migration causes regressions, reverting would also revert the navigation fix
3. **Plugin config change in `.oxlintrc.json`**: the `react-hooks` plugin was removed. In oxlint, react-hooks rules (`react-hooks/exhaustive-deps`, `react-hooks/rules-of-hooks`) are provided by the `react` plugin — so the disable comments `oxlint-disable react/exhaustive-deps` are fine. But the config change itself deserves its own verification.

**Suggestion**: split this into two PRs — one for the oxlint migration, one for the navigation fix.

### ⚠️ UI Regression: Scope Icon Removed

In `compilation-template-card.tsx`, the `ScopeIcon` component and `ScopeIconMap` were removed entirely. Previously the card showed:
- `FileText` icon for file-scoped compilations
- `Database` icon for dataset-scoped compilations

Now it always shows the `Shapes` (compiler) icon regardless of scope. If this was intentional, please note it in the PR description. Otherwise this is a regression — users lose scope-at-a-glance info.

### 🔍 Minor Issues

**1. Over-complex type guard in `home/agent-list.tsx`**

The filter predicate is `item.type !== ...CompilationTemplateGroup`, but the return type asserts it as `AgentListItem & { type: AgentListItemType.Agent }`. This is technically a lie — the remaining elements could have `type === undefined` (which falls into the `Agent` variant of the union). Suggestion: omit the type guard altogether since downstream already handles both:

```ts
const agentList = useMemo(
  () => data.filter((item) => item.type !== AgentListItemType.CompilationTemplateGroup),
  [data],
);
```

**2. Stale `CompilationGroupCategory` constant in `agents/index.tsx`**

Now that `isCompilation` is removed, `CompilationGroupCategory` is still defined as a module-level `const` but only used once. Consider inlining or re-exporting from `constants/agent.ts`.

**3. Missing tests**

No tests were added for the navigation fix or the unified list rendering behavior with mixed agent+compilation items. Consider adding coverage for:
- Compilation template cards render with correct click handler
- `data-pipeline-select` correctly filters out compilation templates
- `home/agent-list` correctly filters to only agents

### 📊 Summary

| Aspect | Verdict |
|---|---|
| Core fix correctness | ✅ |
| Type safety | ⚠️ minor issue |
| UI regression (scope icon) | ⚠️ needs confirmation |
| Bundle discipline | ❌ lint migration should be separate |
| Tests | ❌ missing |

The fix itself is good. Please split the lint migration into a follow-up PR and address the scope icon question.

PR #17461, opened 2026-07-28
WIP (review after ingestion fully ready) fix: drop garbled embedded PDF chars when OCR cannot recover them

euvre:
### Summary

When a PDF page contains garbled embedded text (U+FFFD replacement characters from unmappable CID glyphs in subset fonts without a usable ToUnicode CMap) and OCR cannot recover the content, the garbled characters were previously leaked into the parser output. This matches a real-world scenario where `pdf_oxide` emits U+FFFD for glyphs it cannot map.

**The fix:**
- Detect garbled pages via `util.IsGarbledPage(chars)` upfront and cache the result
- When OCR fails to recover a garbled page, **drop the garbled embedded chars** instead of emitting U+FFFD garbage — matching the Python DeepDOC behavior which clears page chars
- Add a guard so garbled pages never enter the `CharsToBoxes` fallback path
- Add warning logs in `deepDocAnalyzerFromEnv()` when the DeepDoc analyzer is unavailable, so users understand why garbled or scanned pages produce no text

**Tests added:**
- `TestNoDeepDoc_ReplacementCharPage_Dropped`: verifies that a page of pure U+FFFD chars produces no sections with replacement characters
- Extended `TestOCR_FallbackIntegration_NoDeepDoc` to assert no U+FFFD leaks into any section

**Files changed:**
- `internal/deepdoc/parser/pdf/parser.go` — core fix
- `internal/deepdoc/parser/pdf/parser_test.go` — test coverage
- `internal/parser/parser/pdf_parser_common.go` — diagnostic logging

### DIFF
--- a/internal/deepdoc/parser/pdf/parser.go
+++ b/internal/deepdoc/parser/pdf/parser.go
@@ -273,10 +273,11 @@ func (p *Parser) processPage(ctx context.Context, engine pdf.PDFEngine, pg int,
 //   - Clean embedded chars (count > 0 and not garbled):
 //     use ocrMergeChars, which detect-merges chars into boxes and falls
 //     back to single-image OCR recognize for empty boxes.
-//   - Garbled or empty chars: use ocrDetectAndRecognize for full OCR,
-//     then fall back to ocrMergeChars if detect succeeds but recognize
-//     produced no useful output. Synthetic chars are appended to support
-//     subsequent median calculations.
+//   - Garbled or empty chars: use ocrDetectAndRecognize for full OCR.
+//     Synthetic chars are appended to support subsequent median
+//     calculations. When OCR cannot recover a garbled page the embedded
+//     chars are dropped outright (matching Python DeepDOC, which clears
+//     page chars) so U+FFFD garbage never reaches the output.
 //
 // updatedChars includes synthetic OCR chars when detect+recognize was
 // used; callers must use the returned slice instead of the original.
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this fix! The garbled U+FFFD leakage is a real problem and the solution is well-scoped. Here is my review:

**Strengths:**

1. **Clean caching of `IsGarbledPage` result** — caching into `garbledPage` avoids the duplicate call and makes the code more readable.

2. **Correct fix path** — when OCR detect+recognize fails on a garbled page, `chars = nil` is the right call. This matches Python DeepDOC behavior where page chars are cleared.

3. **The `!garbledPage` guard on `CharsToBoxes` is important defense-in-depth.** It covers the code path where OCR is skipped entirely (e.g. `SkipOCR=true` or `renderErr != nil`). Without it, garbled chars would still flow through `CharsToBoxes` in those cases. The fact that `chars` is already `nil` from the OCR path makes it partially redundant in that branch, but it is essential for the non-OCR paths — well done keeping it.

4. **Good test coverage** — the new `TestNoDeepDoc_ReplacementCharPage_Dropped` directly tests the scenario, and extending `TestOCR_FallbackIntegration_NoDeepDoc` with the U+FFFD assertion adds defense-in-depth.

5. **Diagnostic logging** in `deepDocAnalyzerFromEnv()` is helpful for operators to understand why garbled/scanned pages produce no text.

**Minor observation (non-blocking):**

The warning log says "dropping garbled embedded chars after OCR failure" — technically the `!garbledPage` guard on `CharsToBoxes` also drops chars when OCR was never attempted (SkipOCR=true). The log is accurate for the primary path though, and not worth a separate message for the edge case.

**Verdict: LGTM ✅** — focused, well-tested fix with no regressions.

PR #17462, opened 2026-07-28
fix(go-agent): use session IDs for cancellation and context flow

Hz-186:
## Summary

- Propagate request contexts through Agent Canvas execution and external calls.
- Replace internal task IDs with session IDs while retaining `task_id` as a wire alias.
- Complete session-scoped cancellation with Redis lease and token validation.

## Testing

- Go backend tests passed.

image


yuzhichang:
LGTM

PR #17468, opened 2026-07-28
Test: release Go-proxy RESTful contract tests verified passing in Go mode

euvre:
### Summary

Aligns Go and Python error codes/messages so both backends honor the same RESTful API contract, removing implementation-specific error leaks (MySQL errors, `ValueError`, `AttributeError`, Gin validator format) in favor of clean business error codes.

**Chat list** — invalid `orderby` now returns code 101 (was: raw Python `AttributeError` code 100); invalid `page`/`page_size` values fall back to defaults (was: raw `ValueError`/`ProgrammingError` code 100).

**Dataset create/update/delete** — adds UUID validation (101), extra-field rejection (101), duplicate-id detection (101), content-type / JSON-syntax / object-shape checks (101), and "lacks permission" for nonexistent datasets (IDOR). Create auto-deduplicates dataset names. Pagerank updates tolerate a missing ES index. List response includes `parser_config` and `pagerank`.

**Session list/update** — adds filtering, sorting, and pagination support. Empty payloads are valid no-ops. Authorization errors map to code 109.

**Chunk list** — doc object uses Python key names (`chunk_count`, `dataset_id`, `chunk_method`, run text status). Add validates list element types.

**Document update** — adds `chunk_method` alias, pydantic-style Field error messages, metadata index auto-create with refresh, and "These documents do not belong to dataset" messages. List validates `metadata_condition` and reports ownership errors for unmatched name/id filters.

**Search completion** — `kb_ids` ownership failure returns code 102 instead of 109.

Released 31 contract tests from `GO_ONLY_SKIPS` (all verified passing on both Go and Python backends with real LLM keys).


wangq8 (commented):
## PR Review: #17468 — Align Go/Python RESTful API error codes & contract

Thanks for this thorough work! Overall this is a well-structured cleanup with consistent patterns. I have a few concerns to flag, one potentially a data issue, and some observations.

---

### 🔴 Must-fix / High-Impact

#### 1. ES metadata `_id` format changed from `compositeID` to `docID`

In `internal/engine/elasticsearch/metadata.go` (InsertMetadata), the `_id` was:
```go
compositeID := fmt.Sprintf("%d:%s|%d:%s", len(docID), docID, len(kbID), kbID)
```
Now it is just `docID`.

**Risk:** If there is existing data in ES indexed with the old composite format, `UpdateMetadata` queries (which search on `id: docID`) will find matches but the stored `_id` won't match the new insertion format. On a re-insert, you'd create a duplicate document. On a delete, the query-based `DeleteByQuery` would delete correctly, but future inserts would create a new doc with a different `_id`.

**Also:** The `_id` change in InsertMetadata couples with the `RenderID` field being used later — was the `compositeID` serving a dedup purpose across tenants, and does dropping it risk collisions?

#### 2. ES metadata update script changed from **merge** to **full replace**

```diff
- "source": "...for (entry in params.meta_fields.entrySet()) { ctx._source.meta_fields[entry.getKey()] = entry.getValue(); }"
+ "source": "ctx._source.meta_fields = params.meta_fields;"
```

This means a partial update (e.g., `{"foo": "bar"}`) now **blows away** all other existing keys. The PR description says "stale keys must not survive", which is a valid design choice, but this is a significant semantic change — callers that previously sent partial deltas expecting merge semantics will now silently lose data. Please confirm this is intentional and that all call sites send the complete meta fields map.

#### 3. `keywords` query param missing from `listDatasetsAllowedParams`

```go
var listDatasetsAllowedParams = map[string]bool{
    "id": true, "name": true, "page": true, "page_size": true,
    "orderby": true, "desc": true, "include_parsing_status": true,
    "ext": true, "type": true,
    // keywords is missing
}
```

But the handler code below reads `c.Query("keywords")`. This means any request with `?keywords=...` will be rejected with "Extra inputs are not permitted". If Python's ListDatasetReq includes keywords, this is a regression. If it does not, the keywords-reading code below is dead code and should be removed.

---

### 🟡 Should-Fix / Medium

#### 4. `orderby` validation added to endpoints that don't use it

`StartIngestionTask` and `StopParseDocuments` in `internal/handler/document.go` now validate `orderby` query param. These are not list endpoints — they don't sort or return documents. If the Python endpoints here have `extra="forbid"` pydantic models, the Go side should either match by forbidding **all** extra params or not validate `orderby` at all (inconsistent). If the Python side silently ignores unknown params, this validation should not be here.

#### 5. `dropIndex` now silently tolerates missing index

`internal/engine/elasticsearch/common.go`: previously returned an error, now returns nil. The PR says this mirrors Python's behavior. This is fine, but downstream callers that relied on the error to know the index was missing may need updating. Given the PR scope, probably fine, but worth noting for awareness.

#### 6. `accumulateNonStreamAnswer` captures `reference` from *any* event, not just final

```go
if result.Reference != nil {
    finalRef = result.Reference  // last-write-wins from any event
}
```

The `answer`, `audio_binary`, `prompt`, `created_at` are correctly gated on `result.Final`, but `reference` is not. If an intermediate event emits a non-nil reference and the final event emits nil, the intermediate reference is returned. Is this intentional? The stream path behaves differently.

---

### 🟢 Observations / Nice-to-Have

- **CodedError type** in `internal/common/error_code.go` is a clean pattern for domain errors — nice addition.
- **parseJSONRequestObject** helper consolidates content-type + JSON syntax + object-shape validation consistently.
- **Duplicate-id detection** in DeleteDatasets/DeleteDocuments mirrors Python behavior well.
- Empty body now being a valid no-op on `UpdateSession` is a contract change, but the old tests were correctly removed and new behavior is intentional.
- **31 contract tests released from GO_ONLY_SKIPS** — great sign that the alignment is paying off.
- **`chunk_method` alias** in UpdateDatasetDocument is a nice UX addition.
- Error messages now capitalized and punctuated consistently (`"No authorization."`, `"Chat not found!"`) — good for API consumers.
- CI is all green ✅

---

### Summary

This is high-quality work. The main things I'd want resolved before merge are:
1. **ES `_id` format change** — confirm no data migration risk
2. **ES metadata replace-vs-merge semantics** — confirm intentional
3. **Keywords param in allowed list** — either add it or remove the dead code

Everything else is low-risk cleanup. Nice job!

internal/engine/elasticsearch/metadata.go
```diff
@@ -114,12 +114,12 @@ func (e *elasticsearchEngine) InsertMetadata(ctx context.Context, metadata []map
...
 		}
 
-		// Action line: use json.Marshal to properly escape string values
-		compositeID := fmt.Sprintf("%d:%s|%d:%s", len(docID), docID, len(kbID), kbID)
+		// Action line: use the plain document id as _id, matching Python's
+		// es_conn.insert (readers shim hit._id back into the "id" field).
 		action, err := json.Marshal(map[string]interface{}{
 			"index": map[string]interface{}{
 				"_index": indexName,
-				"_id":    compositeID,
```
wangq8, line 122: @qinling0210 / @yuzhichang / @xugangqiang please review.

internal/engine/elasticsearch/metadata.go
```diff
@@ -191,7 +191,21 @@ func (e *elasticsearchEngine) UpdateMetadata(ctx context.Context, docID string,
...
+			{
+				"id":          docID,
+				"kb_id":       datasetID,
+				"meta_fields": metaFields,
+			},
+		}, tenantID)
+		if err != nil {
+			return fmt.Errorf("failed to insert metadata: %w", err)
+		}
+		return nil
```
wangq8: @qinling0210 / @yuzhichang / @xugangqiang please review.

internal/entity/dataset.go
```diff
@@ -224,16 +224,18 @@ type KnowledgebaseDetail struct {
...
+	Language     *string `json:"language"`
+	Description  *string `json:"description"`
 	TenantID     string  `json:"tenant_id"`
 	Permission   string  `json:"permission"`
 	DocNum       int64   `json:"doc_num"`
 	TokenNum     int64   `json:"token_num"`
 	ChunkNum     int64   `json:"chunk_num"`
 	ParserID     string  `json:"parser_id"`
+	ParserConfig JSONMap `json:"parser_config"`
+	Pagerank     int64   `json:"pagerank"`
```
wangq8, line 238: @xugangqiang

internal/engine/elasticsearch/metadata.go
qinling0210: LGTM to use docID as _id to match Python

internal/engine/elasticsearch/metadata.go
qinling0210: I did not see  create logic when metadata index does not exist. Please double check.

internal/engine/elasticsearch/metadata.go
qinling0210: The examples do not match the new change. Update the comments

internal/engine/elasticsearch/metadata.go
euvre: Verified — the create-on-first-write logic is present and correct. When the metadata index does not exist, `UpdateMetadata` calls `CreateMetadataStore` (which creates the index via the `ragflow_doc_meta_*` index template registered at engine init in `client.go`), then `InsertMetadata` to write the document, then returns. This mirrors Python's `update_document_metadata` (doc_metadata_service.py:474-483). I also added an explicit note to the doc comment documenting this behavior: "If the metadata index does not yet exist, it is created (via the ragflow_doc_meta_* index template) and the document is inserted."

euvre:
Addressed the two open review comments from @qinling0210 in commit 4315f91:

**metadata.go:175 — "The examples do not match the new change":** Fixed. The doc-comment examples previously described the old merge behavior, but the implementation does a full replace (`ctx._source.meta_fields = params.meta_fields`). The examples now correctly show replace semantics — keys absent from the input are dropped.

**metadata.go:213 — "I did not see create logic when metadata index does not exist":** Verified the create-on-first-write logic is present and correct. When the index does not exist, `UpdateMetadata` calls `CreateMetadataStore` (creates the index via the `ragflow_doc_meta_*` index template registered at engine init), then `InsertMetadata`, then returns — mirroring Python's `update_document_metadata`. Added an explicit doc note documenting this.

Also fixed a stale inline comment ("merge meta_fields with existing").

internal/engine/elasticsearch/metadata.go
qinling0210: create-on-first-write logic is present in metadata service layer.  In engine layer here, we should not have this logic.

PR #17473, opened 2026-07-28
fix: stabilize wiki planning and artifact graph rendering

buua436:

Wiki planning could produce truncated or malformed page objects when the planning response exceeded the model context limit. These invalid pages could be persisted and later generate malformed artifact links.

The artifact graph API can also return relations whose endpoint entities are not included in the current graph slice. Passing these incomplete relations directly to the frontend ForceGraph causes D3 to throw `node not found` and can interrupt the entire graph rendering.

### What does this PR change?

- Add output token and page-count limits to Wiki planning.
- Split large planning inputs into batches of up to 40 entities/concepts.
- Run planning batches with bounded concurrency and merge results by slug.
- Validate planner output before persisting pages.
- Remove links targeting pages that were not successfully generated.
- Re-render final Wiki content against the actual persisted page set.
- Increase the default Wiki graph entity loading budget from 128 to 512.
- Filter graph relations whose source or target entity is missing from the current graph response.

### Type of change

- [x] Performance improvement
- [x] Data consistency improvement

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -2974,7 +2974,7 @@ _WIKI_COMPILE_KWDS = (
 _WIKI_GRAPH_ENTITY_KWD = "artifact_entity"
 _WIKI_GRAPH_RELATION_KWD = "artifact_relation"
 _WIKI_GRAPH_ENTITY_PAGE_SIZE = 32
-_WIKI_GRAPH_MAX_LOADING_ENTITY = 128
+_WIKI_GRAPH_MAX_LOADING_ENTITY = 512
 
 
 def _wiki_entity_payload(row: dict) -> dict | None:
--- a/rag/advanced_rag/knowlege_compile/wiki.py
+++ b/rag/advanced_rag/knowlege_compile/wiki.py
@@ -2115,10 +2115,6 @@ async def _wiki_planning_call(
                 if slug and slug not in seen_slugs:
                     seen_slugs.add(slug)
                     merged_pages.append(page)
-                    if len(merged_pages) >= max_page_count:
-                        break
-            if len(merged_pages) >= max_page_count:
-                break
         logging.info(
             "wiki_plan: batched planning items=%d batches=%d merged_pages=%d",
             total_items,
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Thanks for this well-structured fix! The multi-layered approach (batching, token limits, validation, link filtering, re-render) is thorough. A few observations:

**wiki.py – planning batching (`_batch_depth`):**
- Nice use of `_batch_depth` to prevent recursive re-batching. The semaphore limiting to 4 concurrent batches is sensible.
- Slug-based deduplication in merged results is correct. Consider logging a warning when a duplicate slug is dropped, for observability.

**wiki.py – page validation:**
- The slug regex `r"(?:entity|concept|topic)/[a-z0-9]+(?:-[a-z0-9]+)*"` matches the documented convention. One edge case: slugs that already exist but happen to have underscores or uppercase would fail validation—is that intentional or should the regex be more lenient? (e.g., `[a-zA-Z0-9_-]+`). If the planner always produces lowercase-hyphenated slugs this is fine, but worth confirming.

- Good defense against missing `action`, `slug`, `title`, `page_type`, `topic`. The logging at warning level will help debug planner output issues.

**wiki.py – `_wiki_transform_links` `valid_slugs`:**
- The `_is_valid` gate correctly strips whitespace on both sides. When a link targets an invalid slug, it falls back to plain text — this prevents D3 crashes and is a clean UX choice. ✅

- The post-refine re-render loop (lines ~3800-3818) overwrites `content_md`, `content_md_rendered`, `outlinks`, and `summary` — does `_wiki_extract_summary` need `content_md_rendered` (processed) or `content_md_raw` (raw)? If summary extraction benefits from the rendered artifact-link format, this is fine; just flagging it because the original per-page render only ran summary extraction once.

**web – `use-artifact-graph-data.ts`:**
- Moving `nodesById` before `links` construction and filtering out incomplete relations is the right fix. Clean. ✅

**api – `dataset_api_service.py`:**
- Bumping `_WIKI_GRAPH_MAX_LOADING_ENTITY` from 128 to 512 is reasonable for larger knowledge bases. No concern.

Overall this is a high-quality defensive fix. The main one to double-check is the slug regex strictness vs existing data.

PR #17474, opened 2026-07-28
Feat: Delete dataset level graph.

cike8899:
### Summary

Feat: Delete dataset level graph.

### DIFF
--- a/web/src/hooks/use-knowledge-request.ts
+++ b/web/src/hooks/use-knowledge-request.ts
@@ -30,6 +30,7 @@ import {
 import i18n from '@/locales/config';
 import kbService, {
   clearWiki,
+  deleteArtifactsStructure,
   deleteKnowledgeGraph,
   getArtifactsAlteration,
   getArtifactGraph,
@@ -95,6 +96,7 @@ export const enum KnowledgeApiAction {
   RemoveKnowledgeGraph = 'removeKnowledgeGraph',
   ClearWiki = 'clearWiki',
   FetchDatasetStructure = 'fetchDatasetStructure',
+  DeleteDatasetStructure = 'deleteDatasetStructure',
   FetchArtifactAlteration = 'fetchArtifactAlteration',
   RunArtifactIndex = 'runArtifactIndex',
 }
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
## PR Review: Feat: Delete dataset level graph

### Summary
Adds ability to delete dataset-level structure graphs (Graph, MindMap, Timeline) from the compilation view, with confirmation dialog and i18n support. Clean frontend-only change backed by an existing backend DELETE endpoint.

### ✅ What Looks Good
- Follows existing code patterns consistently (useMutation, react-query invalidation, ConfirmDeleteDialog)
- Confirmation dialog before destructive action -- good UX safety
- Delete button disabled while deleting to prevent double-clicks
- Backend DELETE /datasets/<id>/artifacts_structure?kind=<kind> already exists and the frontend call matches the contract correctly
- i18n support for both English and Chinese
- State cleanup (setGraphKeywords, setSelectedNodeId) after successful deletion

### ⚠️ Issues

#### 1. Layout breaks for non-Graph views (MindMap / Timeline)
The header div now always renders with justify-between, but the SelectWithSearch entity filter only renders when kind === ViewMode.Graph. For MindMap and Timeline views, the delete button sits alone on the left with empty space on the right -- looks visually broken.

**Suggestion:** Either conditionally use justify-end when there is no search, or always render the search slot but keep it empty for non-graph views.

#### 2. No error handling in useDeleteDatasetStructure
The mutation only handles the success path (code === 0). If the API returns an error code, a network failure occurs, or the request is rejected by the backend, the user gets no feedback -- the dialog closes silently and nothing happens.

**Suggestion:** Add onError to useMutation to show an error toast.

#### 3. Unused data from mutation hook
useDeleteDatasetStructure destructures data from useMutation but never uses it outside the mutationFn. The return value includes data but callers do not consume it. Minor cleanup.

#### 4. Missing test coverage
No tests added for the new hook or component behavior. A test verifying the mutation flow and query invalidation would be expected.

#### 5. Confirm dialog confirm button may lack loading state
The outer trash button correctly shows disabled={deleting}, but the ConfirmDeleteDialog confirm button may not reflect the loading state during deletion. If ConfirmDeleteDialog supports a loading or confirmLoading prop, it should be wired to the mutation pending state to prevent double-clicks on confirm.

### 🔍 Suggestions (non-blocking)
- Consider adding a success toast that names the structure kind being deleted (e.g. "Knowledge Graph deleted successfully") instead of the generic message.deleted.
- The StructureKindLabelKeyMap is a nice addition that could be reused elsewhere.

### Verdict
Looks mostly good. The layout issue and error handling are the main concerns. I would request changes primarily for issues 1 and 2 before merging.

PR #17476, opened 2026-07-28
Fix: rm tenant llm call

Lynn-Inf:
### Summary

Replace tenant_llm call in browser component.


wangq8 (commented):
Thanks for this PR. The removal of the legacy `tenant_llm` fallback in favor of the `model_provider` tables is a clean and necessary migration. A few observations:

### What looks good

- The removal is self-contained — `resolveTenantLLM`, `resolveLLMID`, and `llmIDPattern` have no other callers in the codebase.
- Tests are fully updated to the new `browserLLMLookupForTest` seam with the correct 5-return-value signature.
- The `resolveTenantModelBrowserLLM` lookup chain (`tenant_model` → `tenant_model_provider` → `tenant_model_instance`) includes proper validations: active status check, `ModelTypeChat` capability check, tenant ownership verification on the provider. This is stricter and safer than the old path.

### Questions / Concerns

1. **Breaking change without migration context**: The old `llm_id` format was `ModelName@Factory` (e.g. `deepseek-v4-pro@DeepSeek`), parsed by `resolveLLMID` and resolved via `tenant_llm`. The new code passes the raw `llmID` directly to `resolveTenantModelBrowserLLM`, which does a `tenant_model.GetByID` — so the `llm_id` value is now expected to be a `tenant_model.id` (UUID). Are all existing tenants already migrated to the `model_provider` tables for their browser model configs? If not, their Browser components will break after this change.

2. **Unrelated changes in story files** (`calendar.stories.tsx`, `modal.stories.tsx`): These are purely whitespace/formatting changes (trailing whitespace removal, JSX prop alignment). They are harmless but unrelated to the PR title — ideally these would be in a separate linting/formatting PR to keep the diff focused.

Overall the core change is solid.

PR #16143, opened 2026-06-17
fix: deduplicate markdown table chunks

helloxjade:
### What problem does this PR solve?

When markdown tables are extracted as standalone chunks, the original markdown text is still parsed from the full document content. As a result, the same table content can appear twice in chunk results: once as a dedicated table chunk, and again inside regular text chunks.

This PR fixes that duplication by parsing markdown text from the table-stripped remainder when table separation is enabled. It also enables separated table extraction in the markdown chunking entry points and adds a regression test for pipe-table duplication.

Closes #16074.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)

### Testing

- [x] Added/updated tests
- [x] Verified with:
  - `python -m pytest -q test/unit_test/deepdoc/parser/test_markdown_parser.py`

### Additional context

I also verified the fix manually in the chunk preview UI. After the change, duplicated table chunks no longer appear

Issue #16074, opened 2026-06-16
[Question]: Why are the tables in the Markdown file parsed into two copies?


dbbnsyg001:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

Please look at my picture. 
When I use the parsing function, there will be two identical pieces of content
When I perform the search, both of the two blocks will also be retrieved simultaneously.

<img width="1991" height="997" alt="Image" src="https://github.com/user-attachments/assets/82cb0a32-a089-4c4e-8f92-c44bf7701c0e" />

dbbnsyg001:
Unfortunately, after my test, the HTML table also exhibited the same issue.
<img width="1967" height="595" alt="Image" src="https://github.com/user-attachments/assets/ba390e3a-6990-47b2-9ac3-7b4e408180d3" />

PR #17477, opened 2026-07-28
Fix Markdown parser, the table should be inside chunk

wangq8:
Follow on issue #16074 and PR #16143
Before:
image

After:
image

### DIFF
--- a/deepdoc/parser/markdown_parser.py
+++ b/deepdoc/parser/markdown_parser.py
@@ -59,7 +59,12 @@ class RAGFlowMarkdownParser:
                 else:
                     # Replace with rendered HTML
                     html_table = markdown(raw_table, extensions=["markdown.extensions.tables"]) if render else raw_table
-                    new_text += working_text[last_end : match.start()] + html_table + "\n\n"
+                    next_start = match.end()
+                    while next_start < len(working_text) and working_text[next_start] in "\r\n":
+                        next_start += 1
+                    new_text += working_text[last_end : match.start()] + html_table.rstrip() + "\n"
+                    last_end = next_start
+                    continue
                 last_end = match.end()
             new_text += working_text[last_end:]
             return new_text
@@ -75,7 +80,7 @@ class RAGFlowMarkdownParser:
             """,
                 re.VERBOSE,
             )
-            working_text = replace_tables_with_rendered_html(border_table_pattern, tables, render=separate_tables)
+            working_text = replace_tables_with_rendered_html(border_table_pattern, tables)
 
             # Borderless Markdown table
             no_border_table_pattern = re.compile(
@@ -87,7 +92,7 @@ class RAGFlowMarkdownParser:
                 """,
                 re.VERBOSE,
             )
-            working_text = replace_tables_with_rendered_html(no_border_table_pattern, tables, render=separate_tables)
+            working_text = replace_tables_with_rendered_html(no_border_table_pattern, tables)
 
         # Replace any TAGS e.g. <table ...> to <table>
         TAGS = ["table", "td", "tr", "th", "tbody", "thead", "div"]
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
## PR Review — #17477: Fix Markdown parser, the table should be inside chunk

### Summary

This PR flips the default behavior for markdown chunking from `separate_tables=True` to `separate_tables=False`, so tables stay **inside** their surrounding text chunks rather than being extracted into standalone chunks. It follows up on #16074 / #16143 but takes the opposite approach: instead of deduplicating separated tables, it keeps tables inline.

**Changes:** 3 files, +12 / −11. Looks clean overall, but there are a few things worth double-checking.

---

### File-by-file

#### 1. `deepdoc/parser/markdown_parser.py`

- **`\n\n` → `\n` in both table-replacement functions** (border/borderless markdown + HTML). This reduces paragraph spacing around rendered tables. Sensible for the "table inside chunk" approach since you want the table to feel like part of the paragraph flow. One edge case to watch: if the markdown parser or downstream renderer relies on double-newlines for paragraph breaks, this could cause visual merging. Unlikely to be a real problem in RAGFlow chunks, but worth a quick smoke test.

- **Removed `render=separate_tables` from both calls to `replace_tables_with_rendered_html`.** This was actually a latent bug: when `separate_tables=False`, the old code passed `render=False`, meaning tables kept in the text were **not** rendered as HTML — they stayed as raw markdown pipe-tables. The fix defaults `render=True`, so tables are correctly rendered as HTML when they are kept inline. :white_check_mark:

#### 2. `rag/app/naive.py`

- **`parsing_text = remainder if separate_tables else txt` → `parsing_text = remainder`**: Correct — since `separate_tables` is now always `False` and tables are rendered inside `remainder`, we should always parse from `remainder`.

- **`tbls` loop guarded by `if separate_tables:`**: Prevents double-adding tables when they are already inside `sections`. Good.

- **`chunk()`: `separate_tables=True` → `separate_tables=False`**: The behavioral pivot. Tables will now live inside text chunks, providing better context at retrieval time. The tradeoff is you lose the ability to retrieve a standalone table chunk (e.g., for structured data search). This seems intentional based on the PR title.

#### 3. `rag/flow/parser/parser.py`

- Same `separate_tables=True` → `separate_tables=False` change for the flow-level markdown parser. Consistent with the `naive.py` change. :white_check_mark:

---

### Concerns & Suggestions

1. **Missing tests.** This changes default parsing behavior — a unit test covering a markdown file with a table, verified via `chunk()` (not just the parser in isolation), would give confidence that:
   - Tables appear exactly once in the output
   - Tables are rendered as HTML inside their surrounding chunk
   - No stray raw markdown pipe-tables leak through

2. **Behavioral conflict with PR #16143.** That PR explicitly set `separate_tables=True` and added a regression test (`test_markdown_parser.py`) asserting that tables don't appear twice in chunks. This PR reverses that decision. If this PR is merged, the test added in #16143 likely needs updating (or may now be testing the wrong thing). Please check whether `test/unit_test/deepdoc/parser/test_markdown_parser.py` still passes.

3. **`\n\n` → `\n` in the skip-match branch.** Even though `separate_tables` is now always `False` at the call sites, the function still supports `separate_tables=True` via the closure. The `\n\n` → `\n` change affects that path too — if any code path still calls `extract_tables_and_remainder(separate_tables=True)`, the spacing around removed tables will be tighter than before. Low risk but worth noting.

4. **Caller-side `separate_tables` parameter.** The `__call__` method still accepts `separate_tables=True` as default parameter, but the only callers (`chunk()` and `_markdown()`) now pass `separate_tables=False`. Consider updating the default to `False` for consistency, or at least documenting the new expected behavior.

---

### Verdict

The logic is sound and the duplication bug (#16074) should be resolved by this approach. The main question is architectural — whether keeping tables in chunks (this PR) or separating them (#16143) is the better UX for RAG retrieval. That is a product decision, not a code-quality one.

**Request:** Please run the existing test suite and confirm it passes (or update the test if needed). A small integration test for the new inline-table behavior would also be valuable.

PR #17478, opened 2026-07-28
Fix: The documentation for setting page slicing methods in the Go version of a dataset is not displayed.

cike8899:
### Summary

Fix: The documentation for setting page slicing methods in the Go version of a dataset is not displayed.

### DIFF
--- a/web/src/pages/dataset/dataset-setting/utils.ts
+++ b/web/src/pages/dataset/dataset-setting/utils.ts
@@ -9,6 +9,7 @@ export const ImageMap = {
   manual: getImageName('manual', 4),
   picture: getImageName('media', 2),
   naive: getImageName('naive', 2),
+  general: getImageName('naive', 2),
   paper: getImageName('paper', 2),
   presentation: getImageName('presentation', 2),
   qa: getImageName('qa', 2),
--- a/web/src/pages/dataset/setting/index.tsx
+++ b/web/src/pages/dataset/setting/index.tsx
@@ -24,6 +24,7 @@ import { useCallback, useEffect, useState } from 'react';
 import { useForm, useWatch } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { z } from 'zod';
+import ChunkMethodLearnMore from '../dataset-setting/chunk-method-learn-more';
 import LinkDataSource, {
   IDataSourceNodeProps,
 } from './components/link-data-source';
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
## PR Review: #17478

### Summary
This PR adds the "Learn More" documentation panel to the **dataset Go-version setting page** (`setting/index.tsx`). It mirrors the same pattern already used in the legacy dataset setting page (`dataset-setting/index.tsx`). Additionally, it adds a `general` key to `ImageMap` so the "general" parser displays documentation images instead of falling through to the empty state.

### What Changed

1. **`dataset-setting/utils.ts`** — Added `general: getImageName("naive", 2)` to `ImageMap` so the "general" parser type resolves to documentation images (uses `naive` images as fallback).

2. **`setting/index.tsx`** — 
   - Imported `ChunkMethodLearnMore` component
   - Added `divide-x-0.5` to `CardContent` to visually separate the form and the documentation panel
   - Added a right-side panel that renders `ChunkMethodLearnMore` when `parseType === ParseType.BuiltIn` and a `builtinPipelineId` is selected

### Issues

**1. Empty wrapper div renders unconditionally (same pattern as the existing `dataset-setting/index.tsx`)**

```tsx
<div className="flex-1 p-5 overflow-auto">
  {parseType === ParseType.BuiltIn && builtinPipelineId && (
    <ChunkMethodLearnMore parserId={builtinPipelineId} />
  )}
</div>
```

When `parseType !== ParseType.BuiltIn` or `!builtinPipelineId`, the outer `<div>` still renders, consuming `flex-1` space and `p-5` padding — and the `divide-x-0.5` would draw an unnecessary vertical divider next to empty space. Consider moving the wrapper inside the condition:

```tsx
{parseType === ParseType.BuiltIn && builtinPipelineId && (
  <div className="flex-1 p-5 overflow-auto">
    <ChunkMethodLearnMore parserId={builtinPipelineId} />
  </div>
)}
```

This is a minor visual issue — the page is still functional — but worth tidying up.

**2. Consistent pattern with existing usage**

The `ChunkMethodLearnMore` component is already used in `dataset-setting/index.tsx` with the same wrapper pattern. If you fix issue 1 here, consider fixing it there too for consistency.

**3. `general` ImageMap entry**

The `general` map entry uses `naive` images as a fallback. This is reasonable for getting the documentation to render, but confirm that "naive" images accurately describe the "general" chunking method" if the "general" parser has distinct behavior.

### Verdict

✅ **Looks good to ship** — the fix directly addresses the reported bug (documentation not showing on the Go-version setting page). The CodeRabbit suggestion about the empty wrapper div is a valid polish item but not blocking.

PR #17480, opened 2026-07-28
Refactor(go-models)/llm token usage for longcat 

jay77721:
### Summary

  Add token usage reporting for the LongCat (Meituan) model provider. Related
  to #17284.

  ### Changes

  * Add typed `LongCatChatResponse` struct mirroring the LongCat
  OpenAI-compatible schema
  * Parse token usage (`prompt_tokens`, `completion_tokens`, `total_tokens`)
  from non-streaming chat completion responses
  * Add `stream_options.include_usage: true` to streaming requests so LongCat
  returns aggregate usage as the final SSE event
  * Handle LongCat's thinking mode where `content` may be `null` and all output
  is returned in `reasoning_content`
  * Trim leading newline from `reasoning_content` (matches LongCat API
  behavior)

  ### Verification

  * Real API tested against `api.longcat.chat` — both streaming and
  non-streaming usage parsed correctly
  * `bash build.sh --test ./internal/entity/models/` — all tests pass

### DIFF
--- a/internal/entity/models/longcat.go
+++ b/internal/entity/models/longcat.go
@@ -51,6 +51,31 @@ func (l *LongCatModel) Name() string {
 	return "longcat"
 }
 
+// LongCatChatResponse mirrors the OpenAI-compatible response returned by
+// LongCat's POST /openai/v1/chat/completions endpoint. Usage is always
+// present on a successful response.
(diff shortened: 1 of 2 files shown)


Haruko386 (approved):
LGTM

PR #17482, opened 2026-07-28
fix(admin): use POST for admin logout API

euvre:
### Summary

In Go mode, the admin UI's **Log out** button fails with a 404 error (`hint: 404 The requested resource was not found`).

Root cause: the admin frontend calls `GET /api/v1/admin/logout`, but the Go admin router only registers `POST /logout`.

This PR converges all callers on `POST /logout`:

- Admin UI (`web/src/services/admin-service.ts`): switch the logout request from GET to POST.
- Python-mode admin server (`admin/server/routes.py`): change the `/logout` route to `methods=["POST"]`.
- Go router (`internal/admin/router.go`): keeps the existing `POST /logout` route (unchanged).

The CLI already calls `POST /admin/logout`, so after this change all paths use a single method.

### DIFF
--- a/admin/server/routes.py
+++ b/admin/server/routes.py
@@ -52,7 +52,7 @@ def login():
         return error_response(str(e), 500)
 
 
-@admin_bp.route("/logout", methods=["GET"])
+@admin_bp.route("/logout", methods=["POST"])
 @login_required
 def logout():
     try:
--- a/internal/admin/router.go
+++ b/internal/admin/router.go
@@ -58,7 +58,6 @@ func (r *Router) Setup(engine *gin.Engine) {
 		protected := admin.Group("")
 		protected.Use(r.handler.AuthMiddleware())
 		{
-			protected.GET("/logout", r.handler.Logout)
 			protected.POST("/logout", r.handler.Logout)
 			// Auth
 			protected.GET("/auth", r.handler.AuthCheck)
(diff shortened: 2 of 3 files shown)


wangq8:
@euvre should we just change UI from calling GET /logout to calling POST /logout ?

euvre:
> should we just change UI from calling GET /logout to calling POST /logout ?

Yes, agreed — done in a17569c1c. The admin UI now calls POST /logout (`web/src/services/admin-service.ts`), the Python-mode admin server route was updated to `methods=["POST"]` (`admin/server/routes.py`), and the temporary GET route was removed from the Go router (`internal/admin/router.go`). All callers (UI and CLI) now use POST /logout exclusively.

wangq8 (commented):
Good fix — logout via GET is semantically incorrect (GET should be idempotent/safe, logout is a state change) and caused a 404 in Go mode since the Go router only registers `POST /logout`.

Changes look correct:
- `admin/server/routes.py`: GET → POST on `/logout` (aligns Python mode)
- `web/src/services/admin-service.ts`: `request.get` → `request.post` (fixes the actual bug for the admin UI)
- Go router (`internal/admin/router.go`): already `POST /logout` ✓
- CLI: already `POST /admin/logout` ✓

After this, all callers converge on POST. LGTM.

PR #17483, opened 2026-07-28
Fix admin user list returning empty on first page

euvre:
### Summary

In Go mode, the admin `GET /admin/users` endpoint returned an empty list even when users exist.

The `page` query parameter is 1-based (default 1), but `Service.ListUsers` computed the offset as `pageIndex*pageSize`, so page 1 skipped the first `pageSize` rows. With fewer than one page of users, the query returned no rows and the admin UI showed 'No results found'.

This PR fixes the offset to `(pageIndex-1)*pageSize`.

Verified with `bash build.sh --test ./internal/admin/...`.


wangq8 (commented):
Good catch — this is a classic off-by-one pagination bug.

Verified the call chain:
- The handler defaults `page` to 1 (`ParseRequestIntPositive`, defaults to 1)
- `ListUsers` computed offset as `pageIndex*pageSize`
- With page=1 and pageSize=10, this gives offset=10, skipping the first 10 rows
- The `UserDAO.List` uses `Offset(offset)` directly, where offset 0 means first record

The fix `(pageIndex-1)*pageSize` is correct: page 1 → 0, page 2 → pageSize.

The PR description mentions tests passed with `build.sh --test ./internal/admin/...`. LGTM.

wangq8:
@euvre resolve conflict

PR #17484, opened 2026-07-28
fix(go-agent): return configured retrieval empty responses

Hz-186:
## Summary

- Return the configured `empty_response` when retrieval has no query or no chunks.
- Preserve `formalized_content` for downstream Message nodes.

## Testing

- `bash build.sh --go`
- `ok    ragflow/internal/agent/tool`
- `ok    ragflow/internal/agent/component`


yuzhichang:
LGTM

PR #17488, opened 2026-07-28
Porting change from #17477 python to go, parse table into chunk for markdown file

wangq8:
Porting change from #17477 python to go, parse table into chunk for markdown file
Before:
image

After:
image


yuzhichang:
LGTM

PR #17311, opened 2026-07-23
feat: add AIMLAPI (aimlapi.com) as a model provider

eugeneBackDev:
### Summary

This PR adds **aimlapi.com** as a model provider, so a RAGFlow user can enter one API key in the model settings and use AIMLAPI's models across the app. AIMLAPI ([aimlapi.com](https://aimlapi.com)) is an OpenAI-compatible aggregator that serves 700+ models (LLM, embedding, vision, TTS, ASR) from many providers behind a single API.

The change mirrors the repo's existing "add provider" pattern (e.g. FuturMix / OpenRouter): provider logic lives in the same files those providers use, and shared / UI files get only registration entries.

**Backend**
- `conf/llm_factories.json` — the `aimlapi.com` factory entry.
- `rag/llm/__init__.py`, `rag/llm/{chat,embedding,cv}_model.py` — LiteLLM adapters (chat, embedding, image2text) with a production base URL, overridable via `AIMLAPI_API_URL`.
- `rag/llm/model_meta.py` — an `AIMLAPI` model-meta so the provider lists its full `/v1/models` catalog dynamically (classified by the endpoint `type`), the same way OpenRouter does.
- `api/apps/restful_apis/aimlapi_api.py` — an optional "Get API key" flow using AIMLAPI's agent-authorization (OAuth 2.0 Device Authorization Grant, RFC 8628). The device code is kept server-side (Redis); only the issued key reaches the browser.

**Frontend (`web/`)**
- Provider registration (constant, icon allowlist, brand logo), the model picker (`LIST_MODEL_PROVIDERS` + a `buildLocalConfig` entry), and the "Get API key" button in the provider dialog. Locales added to `en` and `zh`.

**Configuration** — production defaults are compiled in; endpoints and the partner id are overridable through `AIMLAPI_*` environment variables, so the same build works across environments.

**Testing** — the `web` build passes; chat, embedding and dynamic model listing were smoke-tested against the live API.

PR #17491, opened 2026-07-28
fix(aimlapi): send attribution headers on every aimlapi.com request

eugeneBackDev:
### Summary

The **aimlapi.com** provider added in #17311 does not identify itself on any of its outgoing requests, so its traffic cannot be attributed to the integration. This PR adds the two headers AIMLAPI expects — `X-AIMLAPI-Source` and `X-AIMLAPI-Partner-ID` — to every request the provider makes.

**Change**

- `common/aimlapi_utils.py` (new) — a small helper that builds the header pair. It lives in `common/` rather than `rag/llm/` because `api/apps/restful_apis/aimlapi_api.py` is a blueprint loaded at API startup, and importing from the `rag.llm` package would pull in every model module with it.
- `rag/llm/{chat,embedding,cv}_model.py` — the AIMLAPI adapters attach the headers to their OpenAI clients via `with_options(default_headers=…)` after `super().__init__()`, so the base class's timeout and retry wiring is untouched. Both the sync and async clients are covered, so streaming chat and async vision carry the headers too.
- `rag/llm/model_meta.py` — the `AIMLAPI` model-meta sends them on its `/v1/models` catalog fetch.
- `api/apps/restful_apis/aimlapi_api.py` — the two agent-authorization calls send them as well; the partner-id default moves to the shared helper so there is one source for it. The consent URL of the "Get API key" flow also carries the client identifier as a `source` query parameter: consent — and any sign-up that happens during it — runs in a browser on AIMLAPI's side, which a request header cannot reach.

**Configuration** — unchanged in spirit: production defaults are compiled in and both values stay overridable through the environment (`AIMLAPI_SOURCE`, `AIMLAPI_PARTNER_ID`), matching the existing `AIMLAPI_*` variables.

**Testing** — `test/unit_test/common/test_aimlapi_utils.py` covers the defaults, the environment overrides and a blank partner id. End to end, the headers were confirmed on the wire for chat, streaming chat, embeddings and the model-catalog fetch against the live API.

### DIFF
--- a/api/apps/restful_apis/aimlapi_api.py
+++ b/api/apps/restful_apis/aimlapi_api.py
@@ -25,6 +25,7 @@ API key is returned to the browser, to be filled into the provider form.
 Configuration (all via environment):
   AIMLAPI_APP_URL                  agent-auth host (default https://app.aimlapi.com)
   AIMLAPI_PARTNER_ID               partner id (defaults to RAGFlow's; override to change)
+  AIMLAPI_SOURCE                   client identifier sent on every request (default agent/ragflow)
   AIMLAPI_PARTNER_NAME             display name shown on the consent page
   AIMLAPI_VERIFICATION_BASE_URL    base of the consent page (default https://aimlapi.com)
   AIMLAPI_REQUESTED_USD_LIMIT_MINOR requested spend cap in USD minor units (default 1000)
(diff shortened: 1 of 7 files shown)


eugeneBackDev:
@coderabbitai review

Both findings are addressed in the latest push: the model-list fetch is now bounded by an explicit 60s `aiohttp.ClientTimeout`, and the missing-URL and non-200 paths log with context (no bearer token).

PR #17492, opened 2026-07-28
Feat: Optimize dataset-level wiki generation log display.

cike8899:
### Summary

Feat: Optimize dataset-level wiki generation log display.


wangq8 (commented):
## PR Review: #17492 — Optimize dataset-level wiki generation log display

### Overall
Solid UI refactoring with good component extraction. The progress display is much improved. However, there are a couple of unrelated changes that need attention.

---

### ✅ What's Good

- **Component extraction is clean.** `ProgressRing`, `ProgressLogPanel`, `WikiUpdateProgress`, and `parseProgressMsg` are well-isolated, testable units. The SVG ring in particular looks well-done with proper ARIA attributes.
- **Deduplication.** Removed the duplicate `LabelKeyMap` from `empty-state.tsx` and the hardcoded view options array in `index.tsx`, consolidating everything into `ViewModeLabelKeyMap` in `constants.ts`.
- **`wiki-update-sheet.tsx` simplification.** The sheet now reuses `ProgressLogPanel` instead of duplicated markup. Much cleaner.
- **Error state handling.** The progress ring and log entries correctly differentiate between running and failed states visually.
- **`wiki-left-panel` UX improvement.** The update button now remains visible during generation so users can click to see live logs — previously the button disappeared when there were no pending changes.
- **`scrollTop = scrollHeight` with `useLayoutEffect`** in `ProgressLogPanel` — correct approach for auto-scrolling to latest log entry. The dependency on `progressMsg` string rather than `entries.length` is the right call (as the comment explains).

---

### ⚠️ Issues

#### 1. Unrelated changes leaked into this PR

Two changes have nothing to do with wiki generation logs and should be in separate PRs:

- **`message-input/next.tsx`**: Changed default `enableThinking` from `'0'` to `'1'`. This changes the default thinking mode for all message inputs. If intentional, it deserves its own PR with justification; if accidental, please revert.

- **`locales/en.ts` & `locales/zh.ts`**: Renamed `thinkingLevelNone` from `None/无` to `Naive/简单`. Same concern — unrelated to wiki logs, should be a separate change.

#### 2. `parse-progress-msg.ts` — array index as React key

```ts
.map((line, index) => ({
  key: index,
  ...
}))
```

Using array index as key is generally an anti-pattern. While it works here because the entire array is rebuilt on every `progressMsg` change (the `useMemo` ensures identity), future maintainers may not notice this subtlety. Consider using a stable identifier (e.g., a hash of the line content, or a counter included in the progress message format from the backend) to make the intent clearer.

#### 3. `progress-ring.tsx` — unnecessary `stroke=currentColor` on SVG root

The SVG root element has `stroke=currentColor` but all stroke colors are applied via Tailwind classes on the individual `<circle>` elements. The root attribute is dead code and should be removed to avoid confusion.

#### 4. `wiki-left-panel/index.tsx` — `isGenerating` includes `failed` state

```ts
const isGenerating =
  status === GenerateStatus.running || status === GenerateStatus.failed;
```

Naming this `isGenerating` is misleading — it includes `failed` state. Consider renaming to `isGeneratingOrFailed` or `showProgress` to make the intent clearer.

---

### 🔍 Suggestions (non-blocking)

- The commented-out `SessionEssence` / `SessionGraph` entries in `ViewModeLabelKeyMap` look like planned future work — consider adding a `// TODO:` prefix for clarity.
- `wiki-update-progress.tsx`: The `handlePause` properly calls `e.stopPropagation()` since it's nested inside a clickable button — good defensive code.

---

### Verdict

**Request changes** — primarily for the two unrelated changes leaked into the PR (`enableThinking` default and `thinkingLevelNone` rename). The wiki generation UI improvements themselves look clean and well-structured. After those are addressed, this is good to merge.

PR #17493, opened 2026-07-28
Fix: resolve model_id to model before judge parse method

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/db/joint_services/tenant_model_service.py
+++ b/api/db/joint_services/tenant_model_service.py
@@ -500,6 +500,26 @@ def ensure_somark_from_env(tenant_id: str) -> str | None:
     )
 
 
+def get_composite_model_name_by_id(model_id: str) -> str:
+    """Convert a tenant_model.id to the composite model name string
+    ``model_name@instance_name@provider_name``.
+    Raises LookupError if the model, instance, or provider is not found.
+    """
+    exist, model_obj = TenantModelService.get_by_id(model_id)
+    if not exist:
+        raise LookupError(f"TenantModel id={model_id} not found.")
+
+    ok, instance_obj = TenantModelInstanceService.get_by_id(model_obj.instance_id)
+    if not ok:
+        raise LookupError(f"Instance id={model_obj.instance_id} not found for model id={model_id}.")
+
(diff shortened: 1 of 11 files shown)


wangq8 (commented):
Thanks for this fix! The core issue is real — when the frontend passes a model ID instead of the composite name, `normalize_layout_recognizer` can't resolve it properly.

A few observations:

**1. Code duplication** — The same 6-line resolution block is repeated in 8 files (`book.py`, `laws.py`, `manual.py`, `naive.py`, `one.py`, `paper.py`, `picture.py`, `presentation.py`). Consider centralizing this logic:
- Option A: Move the resolution into `normalize_layout_recognizer` itself (in `common/parser_config_utils.py`), so every caller gets it for free without changes.
- Option B: Extract a small helper like `resolve_layout_recognize(config_value, tenant_id)` in `common/parser_config_utils.py`.

**2. Overly broad guard** — `isinstance(layout_recognize_raw, str)` will match *any* string, including valid parser names like `"DeepDOC"`, `"Plain Text"`, or composite names that `normalize_layout_recognizer` already handles. This means a DB lookup will be attempted on every PDF parse even for the common non-model case, and will silently fail via `LookupError`. Consider narrowing the check — e.g., only attempt the lookup if the value looks like a UUID or MongoDB ObjectId pattern. This avoids unnecessary DB round-trips.

**3. Silent failure** — The bare `except LookupError: pass` swallows errors silently. If a valid model ID is passed but the lookup genuinely fails (e.g. model was deleted), the code silently falls back to treating it as a parser name, which could produce confusing downstream behavior. Consider logging a warning so operators can diagnose misconfigurations.

**4. `get_composite_model_name_by_id` location** — Placing this in `api/db/joint_services/tenant_model_service.py` introduces a dependency from `rag/app/*` (application layer) back to `api/db/joint_services/` (API/DB layer). Is this direction of dependency acceptable in the project's architecture?

**5. Tests** — The unit test stubs pass through the model_id unchanged (`lambda x: x`), which means the actual DB lookup path is not exercised. Consider adding a test that verifies the composite name is correctly resolved when a valid model ID is provided.

PR #17499, opened 2026-07-29
Fix filter dataset by owner_ids not working

wangq8:
Test python - passed.
Test go - worked.

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -427,12 +427,18 @@ def list_datasets(tenant_id: str, args: dict):
         kbs = KnowledgebaseService.get_kb_by_name(name, tenant_id)
         if not kbs:
             return False, f"User '{tenant_id}' lacks permission for dataset '{name}'"
-    if ext_fields.get("owner_ids", []):
-        tenant_ids = ext_fields["owner_ids"]
+    owner_ids = [owner_id.strip() for owner_id in ext_fields.get("owner_ids", []) if isinstance(owner_id, str) and owner_id.strip()]
+    if owner_ids:
+        tenants = TenantService.get_joined_tenants_by_user_id(tenant_id)
+        allowed_tenant_ids = {m["tenant_id"] for m in tenants}
+        allowed_tenant_ids.add(tenant_id)
+        tenant_ids = [owner_id for owner_id in owner_ids if owner_id in allowed_tenant_ids]
+        query_user_id = tenant_id if tenant_id in tenant_ids else ""
     else:
         tenants = TenantService.get_joined_tenants_by_user_id(tenant_id)
         tenant_ids = [m["tenant_id"] for m in tenants]
-    kbs, total = KnowledgebaseService.get_list(tenant_ids, tenant_id, page, page_size, orderby, desc, kb_id, name, keywords, parser_id)
+        query_user_id = tenant_id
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## PR Review: #17499 — Fix filter dataset by owner_ids not working

### Summary
This PR fixes a permission bypass in the dataset listing API. When `owner_ids` were passed to filter datasets, the API did not validate that the requesting user actually belongs to those tenants — allowing unauthorized access to other tenants' datasets.

---

### ✅ What's Good

- **The fix is correct.** Both Python and Go layers now validate requested `owner_ids` against the user's joined tenants before passing them to the query.
- **Consistent implementation across layers.** The Python (`dataset_api_service.py`) and Go (`crud.go`) logic is mirrored — same permission check, same `query_user_id` derivation.
- **Input sanitization.** The Python side strips whitespace and filters out non-string values from `owner_ids`.
- **Correct `query_user_id` logic.** When the user's own tenant is in the filtered list, they see their private datasets; otherwise only public/shared datasets are returned. This matches the existing behavior for the non-filtered path.
- **Nil-safe Go code.** `joinedTenant == nil || joinedTenant.TenantID == ""` guards against nil pointers and empty tenant IDs.

---

### ⚠️ Issues

#### 1. No tests

The PR modifies permission-checking logic — a high-value target for regressions. No unit or integration tests are included. At minimum, the following scenarios should be tested:

| Scenario | Expected |
|---|---|
| User passes `owner_ids` containing only tenants they belong to | Returns filtered datasets |
| User passes `owner_ids` containing a tenant they don't belong to | That tenant ID is silently excluded |
| User passes `owner_ids` with only tenants they don't belong to | Returns empty results |
| User passes empty `owner_ids` | Falls through to default behavior (all joined tenants) |

#### 2. Double `.strip()` in Python list comprehension

```python
owner_ids = [owner_id.strip() for owner_id in ext_fields.get("owner_ids", []) 
             if isinstance(owner_id, str) and owner_id.strip()]
```

`owner_id.strip()` is called twice — once in the filter condition and once as the mapped value. Not a bug, but slightly wasteful. A cleaner pattern:

```python
owner_ids = []
for owner_id in ext_fields.get("owner_ids", []):
    if isinstance(owner_id, str):
        stripped = owner_id.strip()
        if stripped:
            owner_ids.append(stripped)
```

Or with a walrus operator (Python 3.8+):

```python
owner_ids = [s for oid in ext_fields.get("owner_ids", []) 
             if isinstance(oid, str) and (s := oid.strip())]
```

#### 3. Potential nil reference in Go?

```go
joinedTenants, err := d.tenantDAO.GetJoinedTenantsByUserID(ctx, dao.DB, userID)
```

The nil guard on individual `joinedTenant` entries is good, but I don't see a nil check on `joinedTenants` itself (the slice). If `GetJoinedTenantsByUserID` can return `nil, nil` on success with no results, this is fine — ranging over nil is safe in Go. Just flagging for awareness.

---

### 🔍 Non-blocking Suggestions

- Consider extracting the permission-filtering logic into a shared helper since it's duplicated across Python and Go layers. This would make future changes harder to miss in one layer.
- The Go pattern `filteredTenantIDs := tenantIDs[:0]` (in-place slice filtering) is idiomatic but subtle — a comment noting it reuses the backing array would help readability.

---

### Verdict

**Approve with suggestion** — the fix is correct and addresses a real permission bypass. The main ask is adding tests for the permission filtering logic. The code quality nits are minor and non-blocking.

PR #17507, opened 2026-07-29
Fix: The multi-model comparison page cannot output messages.

cike8899:
### Summary

Fix: The multi-model comparison page cannot output messages.


Lynn-Inf (approved):
LGTM

wangq8 (commented):
## Review: Fix multi-model comparison page output

**Overall:** Clean, well-targeted fix. The root cause is correctly identified and addressed on both backend and frontend. :white_check_mark:

---

### Backend: `api/db/joint_services/tenant_model_service.py` (+31)

The `get_api_key` function now supports direct model UUID lookup before falling back to the name-based `model@instance@provider` resolution. This is the core fix -- the comparison page was passing UUIDs and the backend couldn't resolve them.

**Good:**
- Tenant ownership verification is thorough (direct + joined tenants) :white_check_mark:
- Clear debug/warning logging for troubleshooting :white_check_mark:
- Graceful fallback preserves existing name-based path :white_check_mark:
- Validation chain is complete: model -> provider -> tenant -> instance :white_check_mark:

**Minor note:** Every `get_api_key` call now has an extra `get_by_id` query even when passing name-based references. Not a hot path so this is fine, but worth being aware of.

---

### Frontend: `next-multiple-chat-box.tsx` (+4/-2) + deleted `use-set-default-model.ts`

Replaces the `useSetDefaultModel` hook (which naively picked the first available model) with a direct `useLayoutEffect` that reads `currentDialog?.llm_id`.

This is **the correct fix**: each chat card in the comparison view should use the model configured for that specific dialog, not the first model in the list.

**Why `useLayoutEffect`?** Runs synchronously before paint, avoiding a flash of the wrong model. Micro-optimization but harmless and correct for form initialization.

---

### Suggestions (non-blocking):

1. **Test coverage:** Consider adding a unit test for the new UUID-lookup path in `test/unit_test/api/db/joint_services/`:
   - Happy path: UUID lookup resolves to an API key
   - Tenant mismatch: raises `LookupError`
   - Missing instance: raises `LookupError`

2. **Naming clarity:** The parameter `model_name` in `get_api_key` now accepts both UUIDs and composite names. Might be worth renaming to `model_ref` (like `resolve_model_type` uses) in a follow-up, but not blocking.

---

**Verdict:** LGTM. No regressions, no security concerns, the approach is sound. Approve after CI passes. :+1:

PR #17515, opened 2026-07-29
Fix: validate Mistral OCR model

Lynn-Inf:
### Summary

As title, check base_url and access_token.


wangq8 (commented):
Thanks for this PR! The change makes sense — validating API reachability in `check_installation` catches misconfiguration early instead of failing deep in the OCR call chain.

**What looks good:**
- The `/models` endpoint is a lightweight, read-only health check — no side effects.
- `base_url` already has `.rstrip("/")` in `__init__`, so the `f"{self.base_url}/models"` concatenation is clean.
- Tests are updated across all affected callers, including the tricky `test_call_ocr_signed_url_error_triggers_cleanup` case where the mock needed conditional routing (200 for `/models`, 403 for the signed URL).
- 10s timeout is reasonable and won't stall startup.

**Minor suggestions (not blocking):**
- Consider adding a dedicated unit test for `check_installation` itself (e.g. test the three branches: missing key, reachable API, unreachable API). The current test coverage is implicit via `_call_ocr` callers.
- The `r.text[:200]` truncation is fine, but if you want richer diagnostics, you could try `r.json().get("message", r.text[:200])` since Mistral API errors typically return JSON with a `message` field.

LGTM, approved.

PR #17522, opened 2026-07-29
fix: support provider-specific verify payload for per-model verify

dcc123456:
### Summary

fix: support provider-specific verify payload for per-model verify

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
@@ -46,6 +46,7 @@ export function DraftModeCard({
   modelInfoRef,
   draftName,
   setDraftName,
+  verifyTransform,
 }: DraftModeCardProps) {
   // On success, fold the OAuth-issued key into the current form values so it
   // lands in the (editable) api_key field without clobbering other inputs.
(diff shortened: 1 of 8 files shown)


wangq8 (commented):
Thanks for this PR! The per-model verify flow definitely needed provider-specific credential mapping — this correctly mirrors the existing `verifyTransform` pattern already used in `useVerifyProvider` for card-level verify. Good stuff.

A few observations:

**1. `VerifyTransform` type duplication**
The type is defined as a named export in `instance-card/interface.ts` and inlined again in `models-section/interface.ts`. Consider importing the type from `instance-card/interface.ts` into the models-section interface to keep the single source of truth. If there are dependency-order concerns preventing this, a comment noting why would help future readers.

**2. `apiKey as any` cast**
`api_key: apiKey as any` — I see the comment explaining why (PaddleOCR produces a nested object where the service signature types `string`). This is pragmatic and matches the existing pattern in `useVerifyProvider`. Might be worth a follow-up to make the service type accept `string | object` properly, but not blocking here.

**3. `instanceName` dependency removal**
The removal of `instanceName` from the `useInstanceSaveState` dependency array in `hooks.tsx` is a separate cleanup. As long as `instanceName` is genuinely unused inside that effect, LGTM — but worth double-checking that the effect body doesn't reference it.

**4. Region support**
Nice that `region` is now forwarded in the per-model verify payload — the original `handleVerify` didn't include it at all, so this is a bonus fix alongside the main change.

**5. Tests**
No test coverage included. Given that `useModelVerify` now has a branching code path (`verifyTransform` vs `resolveCreds`), a unit test covering both paths would add confidence. Non-blocking but worth a follow-up.

Overall the threading is clean — `providerConfig.verifyTransform` → `DraftModeCard`/`SavedModeCard` → `ModelsSection` → `useModelVerify` all lines up. LGTM.

PR #17523, opened 2026-07-29
Fix: On the multi-model comparison chat page, if there is more than one window, each window can be deleted.

cike8899:
### Summary

Fix: On the multi-model comparison chat page, if there is more than one window, each window can be deleted.

### DIFF
--- a/web/src/constants/agent.tsx
+++ b/web/src/constants/agent.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import { setInitialChatVariableEnabledFieldValue } from '@/utils/chat';
 import {
   Circle,
--- a/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
+++ b/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
@@ -226,7 +226,7 @@ const ChatCard = forwardRef(function ChatCard(
                 <p>{t('chat.applyModelConfigs')}</p>
               </TooltipContent>
             </Tooltip>
-            {!isLatestChat || chatBoxIds.length === 3 ? (
+            {chatBoxIds.length > 1 && (
               <Button
                 variant="ghost"
                 size="icon-sm"
@@ -236,7 +236,8 @@ const ChatCard = forwardRef(function ChatCard(
               >
                 <Trash2 />
               </Button>
-            ) : (
+            )}
+            {isLatestChat && idx < 2 && (
               <Button
                 variant="ghost"
                 size="icon-sm"


wangq8 (commented):
## PR Review: #17523

### Summary
Fixes the bug where the latest chat window on the multi-model comparison page could only be deleted when there were exactly 3 chat boxes.

### What Changed
- **`agent.tsx`**: Removed unused `React` import (cleanup, safe with modern JSX transform).
- **`next-multiple-chat-box.tsx`**: Replaced the ternary condition for the Trash/Plus button with two independent conditional renders.

### Analysis

**Old logic** (`isLatestChat = idx === chatBoxIds.length - 1`):
```
!isLatestChat || chatBoxIds.length === 3 → Trash (delete)
ELSE                                     → Plus (add)
```

**New logic:**
```
chatBoxIds.length > 1 → Trash (delete)
isLatestChat && idx < 2 → Plus (add)
```

### ✅ The fix works for the reported bug

Previously, the latest chat in a 2-window setup was **not deletable** (because `2 !== 3`). Now `chatBoxIds.length > 1` correctly allows deletion of any window when more than one exists. Good.

### ❌ Issue: Overlapping button render on exactly 2 chat boxes

When `chatBoxIds.length === 2`, the latest card (idx=1) satisfies **both** conditions:
- `chatBoxIds.length > 1` → `true` → renders Trash button
- `isLatestChat && idx < 2` → `true && true` → renders Plus button

This means **both buttons render on the same card** at the same time, where previously they were mutually exclusive through the ternary. The user would see a functional delete button AND an add button side-by-side on the latest card.

**State matrix:**

| chatBoxIds | idx | Old render | New render |
|---|---|---|---|
| 1 | 0 | Plus ✅ | Plus ✅ |
| 2 | 0 | Trash ✅ | Trash ✅ |
| 2 | 1 | Plus ❌ (should be Trash) | **Trash + Plus ⚠️ (duplicate buttons)** |
| 3 | 0,1 | Trash ✅ | Trash ✅ |
| 3 | 2 | Trash ✅ | Trash only (no Plus) |

### ⚠️ Minor: Add button disappears for 3+ chat boxes

With the `idx < 2` guard, once there are 3+ chat boxes (latest idx ≥ 2), the Plus button never renders anywhere. If there is a maximum window limit, this should be enforced at the parent/addChatBox level rather than implicitly hiding the button.

### Suggested Fix

The two conditions should remain mutually exclusive. Either revert to a ternary with the corrected condition, or add a guard:

```tsx
{chatBoxIds.length > 1 ? (
  <Button ... onClick={handleRemoveChatBox}>
    <Trash2 />
  </Button>
) : (
  <Button ... onClick={addChatBox}>
    <Plus />
  </Button>
)}
```

This keeps the fix (any window deletable when >1 exist) while preserving the mutual exclusivity and showing the add button on the sole remaining window. If a max-window limit is desired, enforce it in the `addChatBox` handler instead.

PR #17527, opened 2026-07-29
Fix: remove unsupported get-rerank model

Lynn-Inf:
### Summary

As title.


rag/llm/rerank_model.py
wangq8, line 525: gte_rerank-v2 ?

wangq8 (commented):
Thanks for the fix! This looks correct.

**Changes:**
- Removes the deprecated `gte-rerank` model from `Tongyi-Qianwen` factory config, since v1 is no longer listed as a valid rerank model in the DashScope API (only `gte-rerank-v2` and `qwen3-rerank` are supported).
- Updates the default `model_name` parameter in `QWenRerank.__init__` from `"gte-rerank"` to `"gte-rerank-v2"`, matching the actively supported model.

**One minor note (not blocking):**
The fallback on line 522 still uses the SDK constant `dashscope.TextReRank.Models.gte_rerank` (which resolves to `"gte-rerank"` v1) when `model_name=None` is explicitly passed. This is an edge case and pre-existing, but worth a follow-up to also update that fallback.

✅ Approved.

PR #17528, opened 2026-07-29
Fix: add StepFun china region base URL (api.stepfun.com)

jay77721:
## Summary
- StepFun serves domestic (China) and international markets from different domains.
- Add a `china` region URL (`api.stepfun.com`) alongside the existing `default` (`api.stepfun.ai`) so tenants in China can route correctly.

## Change
- `conf/models/stepfun.json`: add `"china": "https://api.stepfun.com/v1"` to the `url` map.

## Why
- Chinese AI/LLM providers commonly maintain separate URLs for domestic vs. international markets. Keeping both options available matches the existing pattern used by other providers (e.g. GiteeAI).

### DIFF
--- a/conf/models/stepfun.json
+++ b/conf/models/stepfun.json
@@ -1,7 +1,8 @@
 {
   "name": "StepFun",
   "url": {
-    "default": "https://api.stepfun.ai/v1"
+    "default": "https://api.stepfun.ai/v1",
+    "china": "https://api.stepfun.com/v1"
   },
   "url_suffix": {
     "chat": "chat/completions",


JinHai-CN:
@jay77721 I took a closer look at the official StepFun documentation, and it appears that api.stepfun.ai is still actively used, particularly for the Chat Completions API endpoint (see: https://platform.stepfun.ai/docs/en/api-reference/chat/chat-completion-create).

Many Chinese AI/LLM providers maintain separate URLs for domestic vs. international markets — StepFun seems to follow a similar pattern where api.stepfun.ai serves certain regions and api.stepfun.com serves others.

For this specific configuration file (conf/models/stepfun.json), I'd suggest we keep both options available and use a structured URL config like this:

```
"url": {
    "default": "https://api.stepfun.ai/v1",
    "china": "https://api.stepfun.com/v1"
}
```

Haruko386 (approved):
LGTM

PR #17529, opened 2026-07-29
fix(search): adapt input shape and button position to content line count

dcc123456:
### Summary

ix(search): adapt input shape and button position to content line count

### DIFF
--- a/web/src/hooks/use-auto-resize-textarea.ts
+++ b/web/src/hooks/use-auto-resize-textarea.ts
@@ -6,7 +6,14 @@ import { useLayoutEffect, useState, type RefObject } from 'react';
  * useLayoutEffect runs synchronously after DOM mutation but before the
  * browser paints, so the height adjustment never produces a visible flicker.
  *
- * Returns `isMultiLine` — true when the rendered content spans more than one
+ * Measurement also re-runs when the element's content-box width changes
+ * (e.g. responsive layout or font-metric changes): soft-wrapping can add or
+ * remove lines without `value` changing, which would otherwise leave
+ * `isMultiLine` stale and misposition controls. A ResizeObserver watches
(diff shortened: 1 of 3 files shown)


wangq8 (changes requested):
Thanks for this UX improvement! The idea of adapting the input shape and button position based on whether the content spans multiple lines is solid. Found an issue that needs fixing, plus a few suggestions.

**Issue (must fix):**

In `search-view.tsx`, the hook is now called **twice** on the same ref:

```tsx
useAutoResizeTextarea(searchInputRef, searchText);        // ← original call, still present
const isMultiLine = useAutoResizeTextarea(searchInputRef, searchText);  // ← new call
```

This is wasteful (duplicate DOM reads/writes) and could cause flickering or state inconsistencies. The original non-destructured call should be **removed**, same as you correctly did in `search-home.tsx`. 

**Suggestions:**

1. **`useAutoResizeTextarea` hook — extracting `isMultiLine` calculation:** The line-height extraction logic is reasonable but uses a magic `+1` tolerance. Consider extracting it into a named helper (`getSingleLineHeight`) for readability, or adding a comment explaining the tolerance.

2. **`search-view.tsx` — `rounded-full` in default className:** The base className now hardcodes `rounded-full` and then overrides with `rounded-3xl` when multi-line. Consider keeping only one of them in the base class to avoid confusion (the current approach works because Tailwind applies the later class in the generated CSS, but it is more fragile than using only one of `rounded-full` or `rounded-3xl` in the base and conditionally adding the other).

Otherwise the approach is clean and the UX improvement is nice — single-line inputs get a pill/rounded shape with a centered send button, multi-line inputs expand to a larger radius with a bottom-aligned button. 

Please fix the double-invocation in `search-view.tsx` and I will re-review.

PR #17531, opened 2026-07-29
[codex] Fix oversized dataflow operation log DSL

### LINKED ISSUE #17466: [Bug]: Dataflow pipeline persists component outputs (incl. embedding vectors) in pipeline_operation_log DSL — oversized INSERT marks document Failed after fully successful ingestion
### Self Checks

- [x] I have searched for existing issues (including closed ones). A related but distinct issue is #15198 (broken DB connection-loss retry), which makes this bug non-recoverable.
- [x] I confirm this bug exists on the latest `main` (code inspected on current `main`; reproduced on v0.26.4).

### RAGFlow workspace code commit ID / version

v0.26.4 (image `infiniflow/ragflow:v0.26.4`) — the affected code paths are unchanged on `main`.

### Bug description

When a dataset uses a **custom ingestion pipeline** (dataflow) and the document is large, the task **completes ingestion successfully** (all chunks are inserted into the doc engine, `increment_chunk_num` succeeds, progress is set to 1.0 with "Indexing done"), and then the **very last write** of `run_dataflow` fails:

```
07:17:14 Indexing done (981.17s). Task done (1087.78s)
07:17:30 [ERROR][Exception]: (0, '')
```

The task and the document end up marked **FAIL** even though the document is fully ingested and usable. The dataset list then shows a failed document whose chunks are actually all present in Elasticsearch.

### Root cause analysis

1. `run_dataflow` (`rag/svr/task_executor.py`) persists the operation log as its final step:
   `PipelineOperationLogService.create(..., dsl=str(pipeline))`.
   `Graph.__str__` (`agent/canvas.py`) serializes **every component including its `_param.outputs` values**. For an ingestion pipeline ending with a Tokenizer/"Indexer" node that performs embedding, the outputs contain **all chunks with their embedding vectors** (`q_1024_vec`, 1024 floats each). For our test document (1.2 MB markdown → 3641 chunks × 1024-dim vectors) the serialized DSL is roughly **70+ MB**.

2. This single INSERT exceeds MySQL's default `max_allowed_packet` (64 MB), so the server closes the connection while the client is still sending:

```
File ".../pymysql/connections.py", line 813, in _write_bytes
    self._sock.sendall(data)
BrokenPipeError: [Errno 32] Broken pipe
...
peewee.OperationalError: (2006, "MySQL server has gone away (BrokenPipeError(32, 'Broken pipe'))")
```

3. The retry machinery then fails to recover (this is #15198): `RetryingPooledMySQLDatabase._handle_connection_loss` calls `self.close()`, which itself raises while trying to `ROLLBACK` on the dead socket, so the connection is never actually closed, and every one of the 5 retries dies with:

```
File ".../peewee.py", line 3259, in connect
    raise OperationalError('Connection already opened.')
...
File ".../pymysql/connections.py", line 844, in _execute_command
    raise err.InterfaceError(0, "")
pymysql.err.InterfaceError: (0, '')
```

4. The exception propagates to `handle_task`, which sets progress to -1 → the document is flagged FAIL **after** a fully successful ingestion. Deterministic: reproduced 2/2 runs on the same document; small documents (≤ a few hundred chunks) are unaffected.

Both executors are affected (`rag/svr/task_executor.py::run_dataflow` and `rag/svr/task_executor_refactor/dataflow_service.py` share the same final sequence).

### Expected behavior

- The persisted operation-log DSL should not embed component outputs (at the very least not the embedding vectors) — the log becomes unboundedly large with document size, and none of this payload is needed to display the operation log.
- Additionally, a failure to write the operation log should arguably not flip an already-completed ingestion (progress 1.0, counters committed) to FAIL.

### Steps to reproduce

1. Dataset bound to a custom ingestion pipeline: Parser → Title Chunker → Indexer (search method: Embedding + Full-text), MySQL with default `max_allowed_packet` (64 MB).
2. Upload a large markdown file (ours: 1.2 MB, chunked into 3641 chunks, 1024-dim embedding model).
3. Parse. Ingestion completes ("Indexing done", chunks visible, chunk count updated), then the task fails with `[ERROR][Exception]: (0, '')` and the document is marked Failed.

Workaround we applied: raise MySQL `max_allowed_packet` to 256 MB — this only pushes the limit, any ~4× larger document will hit it again.

### Environment

Kubernetes deployment (Helm), `DOC_ENGINE=elasticsearch`, MySQL 8.0.39 (defaults: `max_allowed_packet=67108864`, `wait_timeout=28800`), reproduced with `TE_RUN_MODE=2` (original executor).

HUAN2022A:
Fixes #17466.

## What changed

- Added an operation-log serialization path for dataflow `Graph` objects that recursively removes dense vector fields named like `q_<dim>_vec`.
- Updated both dataflow task executors to persist the scrubbed operation-log DSL instead of `str(pipeline)`.
- Added unit coverage to verify operation-log serialization removes vectors while default graph string serialization still preserves runtime outputs.

## Why

Dataflow pipeline logs persisted component outputs into `pipeline_operation_log.dsl`, including embedding vectors attached to chunk payloads. Large documents can therefore generate oversized INSERT payloads and mark the document as failed after ingestion has otherwise completed.

## Impact

The result UI can still read normal component outputs, chunks, `output_format`, and `_elapsed_time`, but dense embedding vectors are no longer stored in the operation-log DSL.

## Tests

- `uv run --no-sync --with pytest==8.3.5 --with pytest-asyncio --with nltk python -m pytest test\\unit_test\\agent\\test_canvas_operation_log_serialization.py test\\unit_test\\agent\\test_canvas_at_split.py`
- `python -m py_compile agent\\canvas.py rag\\svr\\task_executor.py rag\\svr\\task_executor_refactor\\dataflow_service.py test\\unit_test\\agent\\test_canvas_operation_log_serialization.py`
- `uv run --no-sync --with ruff ruff format --check agent\\canvas.py rag\\svr\\task_executor.py rag\\svr\\task_executor_refactor\\dataflow_service.py test\\unit_test\\agent\\test_canvas_operation_log_serialization.py`
- `uv run --no-sync --with ruff ruff check --select F,E9 agent\\canvas.py rag\\svr\\task_executor.py rag\\svr\\task_executor_refactor\\dataflow_service.py test\\unit_test\\agent\\test_canvas_operation_log_serialization.py`
- `git diff --check`

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -44,6 +44,22 @@ from rag.utils.redis_conn import REDIS_CONN
 from rag.utils.tts_cache import synthesize_with_cache
 
 _logger = logging.getLogger(__name__)
+_DENSE_VECTOR_FIELD_RE = re.compile(r"^q_\d+_vec$")
+
+
+def _strip_dense_vector_fields(value):
+    if isinstance(value, dict):
+        return {key: _strip_dense_vector_fields(item) for key, item in value.items() if not (isinstance(key, str) and _DENSE_VECTOR_FIELD_RE.match(key))}
+    if isinstance(value, list):
+        return [_strip_dense_vector_fields(item) for item in value]
+    return value
+
+
+def _serialize_default(obj):
+    if callable(obj):
+        return None
+    logging.warning("Graph.__str__: JSON fallback via str() for type=%s", type(obj).__name__)
+    return str(obj)
 
 
 class Graph:
@@ -116,7 +132,15 @@ class Graph:
 
         self.path = self.dsl["path"]
 
-    def __str__(self):
+    def _component_to_dict(self, component, scrub_vectors=False):
+        if scrub_vectors and hasattr(component, "_param") and hasattr(component._param, "as_dict"):
+            return {
+                "component_name": component.component_name,
+                "params": _strip_dense_vector_fields(component._param.as_dict()),
+            }
+        return json.loads(str(component))
+
+    def to_dsl_dict(self, *, scrub_vectors=False):
         self.dsl["path"] = self.path
         self.dsl["task_id"] = self.task_id
         dsl = {"components": {}}
@@ -134,7 +158,7 @@ class Graph:
                 dsl["components"][k] = {}
             for c in cpn.keys():
                 if c == "obj":
-                    dsl["components"][k][c] = json.loads(str(cpn["obj"]))
+                    dsl["components"][k][c] = self._component_to_dict(cpn["obj"], scrub_vectors=scrub_vectors)
                     continue
                 try:
                     dsl["components"][k][c] = deepcopy(cpn[c])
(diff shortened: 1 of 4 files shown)


JinHai-CN:
@HUAN2022A  I will close this PR, if you are still working on this PR, please re-open it.

HUAN2022A:
GitHub would not let me reopen this PR, so I've continued the work in #19371 (branch verified to merge cleanly with the latest `main`). Closing this in favor of #19371.

PR #17532, opened 2026-07-29
fix: add tooltip to log button in AssistantGroupButton component

dcc123456:
### Summary

fix: add tooltip to log button in AssistantGroupButton component


wangq8 (commented):
LGTM. Simple UX improvement that adds a tooltip to the log button, consistent with the existing tooltip pattern already used in the same component (regenerate, delete, read buttons). The i18n key `flow.log` is valid and resolves correctly.

Approved.

PR #17533, opened 2026-07-29
Feat: Click the icon in the top-right corner to delete the note node.

cike8899:
### Summary

Feat: Click the icon in the top-right corner to delete the note node.

### DIFF
--- a/web/src/pages/agent/canvas/node/note-node/index.tsx
+++ b/web/src/pages/agent/canvas/node/note-node/index.tsx
@@ -18,6 +18,7 @@ import { useTranslation } from 'react-i18next';
 import { z } from 'zod';
 import { NodeWrapper } from '../node-wrapper';
 import { ResizeIcon, controlStyle } from '../resize-icon';
+import { ToolBar } from '../toolbar';
 import { useWatchFormChange, useWatchNameFormChange } from './use-watch-change';
 
 const FormSchema = z.object({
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## PR #17533 Review: Delete Note Node via Toolbar

### Overview

Two changes in this PR:
1. Wraps `NoteNode` in the existing `ToolBar` component, enabling the delete (trash) icon in the top-right corner while disabling run/copy buttons
2. An unrelated wiring change in the compilation template edit form

---

### File-by-File

**`web/src/pages/agent/canvas/node/note-node/index.tsx`** (+45/-36)

**What changed:** Wrapped `NodeWrapper` content inside `<ToolBar>` with `showRun={false}` and `showCopy={false}`. All children of `NodeWrapper` are preserved exactly — the diff noise is purely indentation from the added wrapper level. No functional change to rendering beyond the toolbar.

**Assessment:** ✅ **Clean and correct.** The delete path:
```ts
deleteNodeById(id);  // correct for NoteNode (not Iteration/Loop)
```
This is the same `deleteNodeById` store action used by every other node type, which already handles edge cleanup properly.

**`web/src/pages/user-setting/compilation-templates/edit-next/index.tsx`** (+2/-2)

**What changed:**
- `<form>` — added `onSubmit={handleSave}`
- Save button — `type="button"` → `type="submit"`, removed `onClick={handleSave}`

**Assessment:** ⚠️ **Correct change, but unrelated to this PR.** This is a standard React Hook Form pattern — `handleSave` is `form.handleSubmit(onSubmit)` which internally calls `preventDefault()`, enabling native Enter-key submission. However, it has nothing to do with deleting note nodes.

---

### ✅ What is Good

- Minimal diff — uses well-established `ToolBar` component, no new abstractions
- Consistent UX — note nodes now behave like other canvas nodes (same toolbar, same delete behavior)
- No regressions — `showRun={false}` and `showCopy={false}` correctly restrict to delete-only
- State cleanly handled — `deleteNodeById` already manages node + edge cleanup

### ⚠️ Suggestions

**1. Unrelated change (Medium)** — The compilation template `onSubmit` wiring belongs in its own PR. It muddies review scope and complicates `git bisect`/revert if something breaks there.

**2. No test coverage (Low)** — Risk is low since the ToolBar relies on existing store behavior, but a simple render test confirming ToolBar renders inside NoteNode would improve confidence.

### Verdict

**Approve with note.** The core ToolBar change is straightforward and correct. The unrelated compilation template change should ideally be split out, but it is a 2-line improvement following standard patterns.

PR #17535, opened 2026-07-29
Fix: replace unsupported default qwen rerank model

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
LGTM. Clean fix that removes the dependency on the dashscope SDK constant `dashscope.TextReRank.Models.gte_rerank` which appears to be unsupported/deprecated, replacing it with a hardcoded `"gte-rerank-v2"` string. 

The `import dashscope` removal from `__init__` is safe — it's still imported in `_compute_rank` where the API calls happen. The truthiness check (`model_name if model_name else ...`) is actually a minor improvement over the old `is None` check, catching empty strings too.

Approved.

PR #17538, opened 2026-07-29
Feat(ingestion): add canvas pipeline debug (dry-run) mode with View result log

xugangqiang:
## Summary
Adds a side-effect-free DataFlow canvas **debug (dry-run) mode** plus a **debug run log with a "View result" panel**, so a canvas can be executed synchronously and inspected end-to-end (per-component progress and parsed chunks) without persisting anything.

### Dry-run execution (inline parsed chunks)
- `task/debug.go`: `NewDebugTaskContext` builds an in-memory `TaskContext` with **`KB.ID == ""` — the single debug signal used across the ingestion pipeline**. A canvas debug run has no knowledgebase, and production ingestion always supplies one, so `kb_id == ""` occurs ONLY in debug mode. Components gate their own side effects on this signal without any dedicated debug vocabulary (the former `CANVAS_DEBUG_DOC_ID` marker constant is removed).
- `task/pipeline_executor.go`: `validateTaskContext` no longer requires a KB when `KB.ID == ""` (debug); debug runs return `collectDebugOutput` (chunks) instead of a no-op; uploaded bytes are delivered as `inputs['binary']` for doc-less runs; `injectDebugPageCap` caps the parser to the first pages for a fast preview via the production `override_params` channel (Parser cpnID + family).
- `component/tokenizer.go`: `shouldHaveEmbedding` skips embedding when `kb_id == ""` — the embedder is configured on the knowledgebase, so a debug run has nothing to resolve against and stays side-effect free.
- `chunker/register.go`: chunk images are uploaded to MinIO only when a KB is present (persist run). **In debug mode the raw image bytes are intentionally dropped (`delete(ck, "image")`)** — the debug preview does not render chunk images, and dropping the bytes keeps them out of memory and out of the Redis-stored debug log. This is a deliberate trade-off, not an oversight.
- `component/file.go`: pass through in-memory binary bytes, skipping `doc_id` -> storage resolution.
- `handler/agent.go` + `agent_webhook.go`: detect `dataflow_canvas` and run a sync debug returning chunks inline on the existing chat/completions endpoint; reject DataFlow canvases from webhooks (fixes the previously dead `== "DataFlow"` check; mirrors Python `agent_api.py`).
- `parser_dispatch.go`: export `ParserFileFamily` for the executor's page-cap injection.

### Debug run log + "View result"
Mirrors Python's debug-log contract so the front-end can replay each component's progress and parsed output:
- `task/debug_log_sink.go`: a `DebugLogSink` records every component's lifecycle into a `[{component_id, trace}]` array (each trace entry carries `message`, `progress`, `timestamp`, `elapsed_time`). `Flush` appends a terminal `END` marker whose first trace message is non-empty so the front-end detects completion. On failure the END marker is prefixed `[ERROR]` yet still carries the run, so the failure timeline renders instead of being stuck empty. Timestamps and `elapsed_time` are in seconds (matching the rest of the app).
- `task/debug_result_dsl.go`: `BuildDebugResultDSL` builds the `dsl` the END marker carries — the Go analogue of Python's `Graph.__str__` + END-marker `dsl` in `rag/flow/pipeline.py`. It combines the static DSL structure (component_name / downstream / params / graph.nodes) with the run output map (`output["state"][<id>]`) to emit, per component, `obj.params.outputs[<format>].value` (chunks / text / json / html / markdown) — the exact keys the front-end `dataflow-result` page reads to render each step's parsed chunks. Raw embedding vectors (including the dimension-scoped `q_<dim>_vec` keys) are stripped so the stored log stays Python-scale.
- `task/pipeline_executor.go`: after the run, attach the built `dsl` to the END marker via the `ResultSink` capability.
- `handler/agent.go`: `runCanvasPipelineDebug` generates a stable `message_id` up-front and always flushes the log (success or failure); `respondWithDebugResult` returns `message_id` in **both** the success and the error envelope so the front-end can poll the log. The debug-log endpoint `GET /agents/:id/logs/:message_id` serves the array.
- `web/src/pages/agent/hooks/use-run-dataflow.ts`: on a run failure, also surface `message_id` via `setMessageId` so the log sheet renders the failure timeline (the `[ERROR]` END marker is already written). Guarded by `if (msgId)`, so it is a safe no-op when the back-end does not return an id.

## Behavioral notes
- Debug parses only the first pages (`debugPageCapPages`) for a fast preview; an explicit `pages` cap already present in the ParserConfig is respected.
- Debug mode does not keep chunk images (see above) and does not compute embeddings — it exercises parse + chunk only.

## Test plan
- Go: `debug_test.go`, `debug_log_sink_test.go` (trace pairing, END marker, `[ERROR]` prefix, fractional-second timestamp/elapsed_time, size caps, and a real-pipeline test asserting the END-marker `dsl` carries non-empty per-component `params.outputs` with chunks), `debug_result_dsl_test.go` (flat and real nested `output["state"]` shapes, vector stripping, format priority), `debug_pages_integration_test.go`, `pipeline_executor_persist_test.go`, `handler/agent_pipeline_debug_test.go`, `handler/agent_logs_test.go` (incl. `TestRunCanvasPipelineDebug_ErrorStillExposesMessageID` / `TestRespondWithDebugResult_ErrorCarriesMessageID` locking `message_id` on failure), plus updates to `agent_test.go` / `agent_webhook_test.go` / `chunker/image_upload_test.go` / `tokenizer*.go`.
- `go build ./...` and `./build.sh --test` for affected packages.


xugangqiang:
@yingfeng @yuzhichang @JinHai-CN  Please help to review

internal/handler/agent.go
```diff
@@ -464,6 +517,127 @@ func readUserInput(c *gin.Context) string {
...
+// It performs no persistence (no MinIO upload, index insert, or pipeline_log)
+// because the debug context carries no KB (KB.ID == ""), which is the single
+// debug signal used across the ingestion pipeline: the tokenizer skips
+// embedding and the executor skips the persist stage. kb_id == "" occurs ONLY
+// in debug mode; production ingestion always supplies a KB, so this path is
+// never taken in normal operation. The request's kb_id is intentionally never
+// forwarded: debug never resolves an embedder or writes to a KB, and
+// parse/chunk work without it. Callers extract the (fileName, fileData) pair
+// from their own request shape and pass them in.
+func (h *AgentHandler) runDataflowDebug(ctx context.Context, user *entity.User,
```
yuzhichang: Remove the `dataflow` concept.

xugangqiang: Done in 925218ee9 — the Go side no longer introduces a `dataflow` concept of its own:

- `runDataflowDebug` → `runCanvasPipelineDebug`, so the handler stays with the existing ingestion-side "pipeline" vocabulary (`internal/ingestion/pipeline`, `PipelineExecutor`). All doc comments referencing it were updated.
- `agent_dataflow_test.go` → `agent_pipeline_debug_test.go`, `TestRunDataflowDebug_*` → `TestRunCanvasPipelineDebug_*`.
- Dropped the dead `kb_id` plumbing on the debug entry (request field + query fallback + leftover debug log). It was never consumed: the debug context is KB-less by construction (`NewDebugTaskContext` forces `KB.ID == ""`), so that concept is gone from the entry point too.

What I deliberately kept, please tell me if you want these changed as well:
- the `"dataflow_canvas"` category literal — it is the DB/API contract shared with Python (`CanvasCategory.DataFlow`);
- the webhook rejection message `"Dataflow can not be triggered by webhook."` — verbatim parity with `agent_api.py`;
- the front-end `use-run-dataflow.ts` hook name, which predates this PR.

`./build.sh --test ./internal/handler/... ./internal/ingestion/task/...` passes.

internal/handler/agent.go
```diff
@@ -464,6 +517,128 @@ func readUserInput(c *gin.Context) string {
...
+	var chunks []map[string]any
+	if result != nil {
+		messageID = result.MessageID
+		chunks = result.Chunks
+	}
+	if len(chunks) == 0 {
+		chunks = []map[string]any{}
+	}
+	common.SuccessWithData(c, gin.H{"message_id": messageID, "chunks": chunks}, "success")
+}
```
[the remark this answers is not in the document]
xugangqiang: Token usage is not used at python side. Thus go side will not use it and no need to return it. Changed PR description - the PR will not include token usage.

xugangqiang:
Design-review tracking for `internal/ingestion/task/pipeline_executor.go` is captured in issue #17607 (assigned to xugangqiang). Full CTO / senior-backend review of the god-object concerns and proposed refactors: https://github.com/infiniflow/ragflow/issues/17607

PR #17541, opened 2026-07-29
Fix: audio button can send message

dcc123456:
### Summary

Fix: audio button can send message


wangq8 (commented):
Thanks for this fix! Clean and well-scoped PR.

**What this fixes:**
The `type="button"` attribute is the key fix — without it, the `<Button>` inside a form context defaults to `type="submit"`, which would cause the chat form to submit (i.e., send the message) when the audio button was clicked. This directly resolves the "audio button can send message" bug.

**Other changes:**
- The `getUserMedia` capability check with a user-facing error toast is a nice defensive addition — prevents silent failures in unsupported browsers.
- Removing the `{false && (...)}` dead code block is good cleanup; no functional impact since that branch was never rendered.

**Verified:**
- `import message from "./message"` — `message.ts` exists and exports `.error()`, so the import is correct.
- No regressions expected from the dead code removal.

LGTM, nice fix! 👍

PR #17542, opened 2026-07-29
Fix: Add list_models support for Hugging Face provider

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the PR! A few observations:

**Title / description mismatch**
The PR title says "Hugging Face provider" but the actual change is in `QWenRerank` (Tongyi-Qianwen), not `HuggingfaceRerank`. Please fix the title (or clarify how this relates to HuggingFace if I am missing something). The PR body ("As title.") does not help — please add a short description of *why* this change is needed.

**Code changes:**
The core idea — deferring the `dashscope` import out of `__init__` and replacing the SDK constant with a plain string — is reasonable. It allows the class to be instantiated without `dashscope` installed, which is exactly what is needed for `list_models`.

Two things to double-check:

1. `dashscope.TextReRank.Models.gte_rerank` resolving to `"gte-rerank"` vs `"gte-rerank-v2"`: The existing `HuggingfaceRerank` default is `"BAAI/bge-reranker-v2-m3"`, and its `__init__` already uses plain strings with no SDK import — this is the pattern you are following, which is correct. The QWenRerank default model name parameter was `"gte-rerank"` (no `-v2`), and that parameter value **always wins** over the dashscope SDK constant in the "model_name is None" branch. So the prior behaviour used the parameter default `"gte-rerank"`, not the SDK default. This PR changes the parameter default to `"gte-rerank-v2"`, which is a behavioural change. Is `"gte-rerank-v2"` the correct model name that `dashscope.TextReRank.Models.gte_rerank` maps to? Please confirm.

2. `model_name if model_name else "gte-rerank-v2"` is broader than the original `model_name is None` — it will also catch empty strings and `0`. That is arguably better (empty model names are invalid), but worth calling out.

**Nits:**
- The comment `# Remove invalid global timeout, use official SDK per-request timeout parameter` no longer relates to the line it sits next to — the `self.request_timeout` assignment is unchanged and the comment predates this PR. Not blocking, but consider moving or removing it.

PR #17545, opened 2026-07-29
Fix: The compiler operator icon is not displaying on the pipeline log page.

cike8899:
### Summary

Fix: The compiler operator icon is not displaying on the pipeline log page.


wangq8 (commented):
## PR Review: #17545

### Summary
This PR fixes the compiler operator icon not displaying on the pipeline log page, and addresses two root causes:
1. **Enum mismatch**: `Operator.Compilation` had value `"Compiler"`, causing icon lookups to miss since the key didn't match. Renamed to `Operator.Compiler = "Compiler"` (consistent key-value pair).
2. **Missing icon mappings**: `OperatorIcon` was missing mappings for File, Parser, Tokenizer, TokenChunker, TitleChunker, and Extractor — all needed by the pipeline log page.

### What I checked
- **Enum rename completeness**: All 10 references to `Operator.Compilation` across the codebase are renamed to `Operator.Compiler`. ✅
- **Import path updates**: All 18 `../../operator-icon` imports updated to `@/components/operator-icon` after the component was moved from `web/src/pages/agent/` to `web/src/components/`. ✅
- **Deduplication**: The `TimelineNodeObj` icon definitions in `dataflow-result/components/time-line/index.tsx` were a duplicate of the shared `OperatorIcon` component. Removed them (~44 lines) and the timeline now uses `OperatorIcon` dynamically via `createElement(OperatorIcon, { name })`. Cleaner and DRYer. ✅
- **Graceful fallback**: When the component name doesn't map to an `Operator` enum member, the icon gracefully becomes `undefined` (no crash). ✅
- **Non-clickable nodes**: The compiler node is correctly added to the non-clickable list alongside `begin` and `tokenizer`, consistent with its behavior. ✅
- **Console.log cleanup**: Removed stray debug logs from the hooks file. ✅

### Minor observation (non-blocking)
The dynamic icon resolution in `useTimelineDataFlow` uses `camelCase(component_name)` → `upperFirst()` → `Operator[name]`. This assumes the backend `component_name` snake_case converts cleanly to the PascalCase `Operator` enum key. This works for all current pipeline operators, but if a new operator is added without updating the `Operator` enum, its icon will silently render as nothing. A development-time warning could help future contributors, but not a blocker.

### Verdict
✅ **LGTM**. Solid fix with good architectural cleanup — moving `operator-icon` to shared components and eliminating the duplicate icon definitions. The rename from `Compilation` to `Compiler` makes the enum self-consistent. Approved.

PR #17546, opened 2026-07-29
feat: add LLM-guided semantic rechunking for knowledge compilation

buua436:

Knowledge compilation currently relies on parser chunk boundaries, which can split semantically related content and reduce extraction quality.

This PR adds LLM-guided semantic rechunking for pipeline compilation. The first template with `rechunk=true` determines the semantic boundaries, and subsequent templates reuse the generated chunks with rechunking disabled.

### What changed?

- Added semantic rechunking based on the knowledge compilation task.
- Preserved structured parser records such as tables and images.
- Added temporary positional chunk IDs for compact LLM responses.
- Replaced original chunks with the formal semantic chunks after rechunking.
- Ensured later templates consume the first template's rechunked output.
- Kept Tree compilation on its existing path.
- Added frontend configuration and localized descriptions for rechunking.
- Added DeepSeek V4 reasoning and output-token configuration for knowledge compilation.

### Type of change

### DIFF
--- a/api/db/init_data/compilation_templates/knowledge_graph.yaml
+++ b/api/db/init_data/compilation_templates/knowledge_graph.yaml
@@ -2,6 +2,12 @@ kind: knowledge_graph
 display_name: Knowledge graph
 config:
   kind: knowledge_graph
+  rechunk_rules: |
+    - Group only adjacent source chunks into coherent semantic units for this knowledge-compilation task.
+    - Preserve source order: use every source chunk exactly once; do not omit, overlap, duplicate, or reorder chunks.
+    - Split when the topic, section, or semantic focus changes; do not merge unrelated content.
+    - Keep self-contained tables, lists, figures, and code blocks intact; do not split their internal structure.
+    - Do not invent, rewrite, or return source text; return only chunk grouping metadata.
   entity:
     description: >-
       You are a robust graph entity extractor for knowledge graphs.
--- a/api/db/init_data/compilation_templates/mind_map.yaml
+++ b/api/db/init_data/compilation_templates/mind_map.yaml
@@ -2,6 +2,12 @@ kind: mind_map
 display_name: Mind map - Radial concept hierarchy
 config:
   kind: mind_map
+  rechunk_rules: |
+    - Group only adjacent source chunks into coherent semantic units for this knowledge-compilation task.
+    - Preserve source order: use every source chunk exactly once; do not omit, overlap, duplicate, or reorder chunks.
+    - Keep each main idea with its supporting content, and split when the topic or branch changes.
+    - Keep self-contained tables, lists, figures, and code blocks intact; do not split their internal structure.
+    - Do not invent, rewrite, or return source text; return only chunk grouping metadata.
   entity:
     description: >-
       You are a robust mind-map extractor. Extract the central concept,
--- a/api/db/init_data/compilation_templates/page_index.yaml
+++ b/api/db/init_data/compilation_templates/page_index.yaml
@@ -2,6 +2,14 @@ kind: page_index
 display_name: PageIndex — Hierarchical table of contents
 config:
   kind: page_index
+  rechunk_rules: |
+    - Group only adjacent source chunks into coherent semantic units for this page-index task.
+    - Preserve source order: use every source chunk exactly once; do not omit, overlap, duplicate, or reorder chunks.
+    - Keep each heading with the content it introduces, and split when the topic, section, or semantic focus changes.
+    - Preserve document heading hierarchy: keep a parent heading with its introductory content and nearby child-section context.
+    - Do not combine separate sections merely because their subjects are related; keep each section boundary recoverable.
+    - Keep self-contained tables, lists, figures, and code blocks intact; do not split their internal rows or items.
+    - Do not merge unrelated content. Do not invent, rewrite, or return source text; return only chunk grouping metadata.
   entity:
     description: >-
       You are a source-grounded document index extractor. First extract every
--- a/api/db/init_data/compilation_templates/session_essence.yaml
+++ b/api/db/init_data/compilation_templates/session_essence.yaml
@@ -2,6 +2,12 @@ kind: session_essence
 display_name: Session Essence — Cross-source entity synopses from conversations
 config:
   kind: knowledge_graph
+  rechunk_rules: |
+    - Group only adjacent source chunks that discuss the same conversation topic or entity.
+    - Preserve source order: use every source chunk exactly once; do not omit, overlap, duplicate, or reorder chunks.
+    - Split when the speaker, topic, intent, or semantic focus changes.
+    - Keep self-contained tables, lists, figures, and code blocks intact; do not split their internal structure.
+    - Do not invent, rewrite, or return source text; return only chunk grouping metadata.
   entity:
     description: >-
       You are a robust entity and fact extractor for conversational data.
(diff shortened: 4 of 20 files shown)


wangq8 (commented):
Thanks for this PR! The LLM-guided semantic rechunking is a nice addition to knowledge compilation. I reviewed the full diff (~700 lines across 14 files) and have the following feedback:

## Overall
Clean implementation with good separation of concerns. The chunk normalization pipeline (`_normalize_upstream_chunks`) handles multiple output formats well, and the `_RechunkedDocs` subclass is a neat way to thread rechunked data through the existing batching infrastructure without changing the signature of `_struct_process_batch`.

## Issues

### 1. Undocumented knowledge_graph behavioral change (`runner.py` L493-495)
The special case for `knowledge_graph` templates was removed without mention in the PR description:
```python
# removed:
if template_kinds[template_id] == "knowledge_graph":
    await _submit_one(incoming_batch, template_id, parser_cfg)
    continue
```
Previously, knowledge_graph templates bypassed the dynamic token-budget batching and submitted the full incoming batch directly. Now they go through the same buffering path as other templates. This changes batch boundaries for all knowledge_graph users and could affect extraction quality (e.g., entities that span chunk boundaries may be split across different LLM calls). Was this intentional? If yes, please document the rationale in the PR description. If no, this should be restored.

### 2. Empty source chunks in `_formalize_rechunked_chunks` (`compiler.py`)
When `_formalize_rechunked_chunks` finds no matching source chunk for any source_id in a rechunked group:
```python
source = next((originals[source_id] for source_id in source_ids if source_id in originals), {})
```
`source` becomes an empty dict `{}`. The resulting formalized chunk will have `id` and `text` from the rechunk data, but will be missing fields like `doc_type_kwd`, `positions`, etc. Can this happen in practice? Consider adding a guard: if `not source`, either skip the group or log a warning.

### 3. `_is_atomic_json_record` pipe-character scan (`compiler.py`)
```python
has_markdown_table = any("|" in lines[index] and Compiler._is_markdown_table_separator(lines[index + 1]) ...)
```
The `"|" in lines[index]` check is broad — any line containing a pipe character followed by a separator line will be treated as a table. The separator check mitigates false positives, but consider using a regex like `r"^\s*\|"` to only match lines that actually start a table row, which would be more precise.

### 4. DeepSeek max_tokens workaround (`chat_model.py`)
The workaround is correct — LiteLLM does not translate `max_completion_tokens` for DeepSeek, and the provider falls back to 8192. However, this silently overrides `max_completion_tokens` for *all* DeepSeek calls, not just knowledge compilation. The 65536 value set in `_common.py` is then carried through as `max_tokens`. Consider adding a brief inline comment explaining the chain (from `_common.py` → `_clean_conf` → DeepSeek API) so future maintainers understand why `max_completion_tokens` disappears from gen_conf for DeepSeek.

### 5. `source_texts` memory in `_struct_extract_hypergraph` (`structure.py`)
```python
source_texts = {match.group(1).strip(): match.group(2).strip() for match in re.finditer(r"\[CHUNK_ID:\s*([^\]]+)\]\n(.*?)\n\[END_CHUNK\]", text, flags=re.DOTALL)}
```
This captures all chunk text into a dict (text is already in memory as `combined_text`, so no extra copies), but the regex `.*?` between `[CHUNK_ID:]` and `[END_CHUNK]` with `re.DOTALL` could match very large strings. For batches of large chunks, consider adding a comment noting this is bounded by `DOC_STRUCTURE_COMPILE_BATCH_CHUNKS`.

## Nitpicks
- The `_PARSER_CANDIDATE_TOKEN_SIZE = 128` when rechunking is intentionally small to give the LLM fine-grained chunks to group. Worth a brief comment explaining this design choice.
- Good that Tree templates continue on their existing path without rechunking interference.
- The positional ID scheme (`t1`, `t2`, ...) is clever — much more compact for the LLM response than full hash IDs.

## Frontend
- `SwitchFormField` import: `@/components/switch-fom-field` — the "fom" typo is pre-existing in the codebase, not introduced here. No action needed.
- Localization strings look correct for both `en` and `zh`.
- The `rechunk` boolean is properly handled in `isConfigMetaKey` and `transformTemplateToPayload`.

No blocking issues — the knowledge_graph change in runner.py is the one thing I would clarify before merging.

wangq8 (commented):
Thanks for this PR — LLM-guided semantic rechunking is a solid improvement for knowledge compilation quality. The architecture of having the first rechunk template define boundaries and subsequent templates consume them is clean. I have several concerns that need addressing before this can land:

## High priority

### 1. Potential content loss in `_split_slumber_level`

The threshold check `if end - start >= cls._PARSER_MIN_SPLIT_CHARACTERS` can silently drop short segments that appear before a delimiter delimiter when the remaining text produces ≥2 valid splits (which prevents the recursion fallback in `_slumber_candidates` from trying the next level on the full text):

```python
# Example: "Short header.\n\nSection A content.\n\nSection B content."
# "Short header.\n\n" = ~17 chars, dropped.
# Remaining segments are both valid → len(splits) = 2 → no recursion → header lost.
```

The merge loop afterward only merges fragments that were already appended, so dropped segments are never recovered. Consider appending all segments to `parts` initially and only filtering short ones in a smarter merge pass, or always including the first segment regardless of length.

### 2. DeepSeek token handling in `_clean_conf` is a breaking change

The modification to `chat_model.py` `_clean_conf` removes `max_tokens` for ALL DeepSeek calls, not just knowledge compilation. This means any existing code path that passes `max_tokens` to a DeepSeek model will silently lose its token limit unless it also passes `max_completion_tokens`. The `max_completion_tokens` value only survives when explicitly set by knowledge compilation. Other callers (Agent UI, chat, etc.) would have their `max_tokens` stripped with no replacement.

Consider scoping this to only activate when `max_completion_tokens` is actually present in `gen_conf`, or use a dedicated parameter to opt into this behavior.

## Medium priority

### 3. Missing test coverage

812 additions across 20 files with zero test files changed. This feature touches chunk boundary logic, LLM response parsing, and the compiler pipeline — all prime candidates for regressions. At minimum, unit tests for `_slumber_candidates`, `_split_slumber_level`, `_is_atomic_json_record`, `_struct_expand_source_chunk_ids`, and the `_normalize_upstream_chunks` dispatch would help.

### 4. Dual normalization of chunks in `_invoke`

When rechunking is enabled and `output_format == "chunks"`, chunks pass through `_normalize_upstream_chunks` twice — once without splitting (`target_token_size=1024`) and once with splitting (`target_token_size=128`). The first normalization generates hash-based IDs, the second generates positional t-IDs, and the first set is discarded. This is wasteful and confusing. Consider restructuring so only one pass happens.

### 5. Knowledge graph batch budget reduction

`KNOWLEDGE_GRAPH_CONTEXT_FRACTION = 0.1` with max 4096 tokens is a 10x reduction from the default. This might be intentional for latency, but please document the rationale. Low context could degrade entity/relation extraction for knowledge graphs.

## Minor

### 6. In-place mutation in `_merge_small_text_chunks`

`pending["text"] = combined` mutates the chunk dict in-place. If any caller holds a reference to these chunk dicts, they will see the mutated text. Consider using `deepcopy` or constructing new dicts.

### 7. Regex in `_struct_extract_hypergraph`

The chunk text regex `r"\[CHUNK_ID:\s*([^\]]+)\]\n(.*?)\n\[END_CHUNK\]"` with `[^\]]+` for the ID capture group could theoretically match across unwanted boundaries if the actual chunk text contains `]`. More defensive to use `[^\]]*?` or a more specific pattern.

### 8. `source_chunk_ids` remapping asymmetry

Entity IDs are remapped via `temp_to_uuid` (filtering out unmapped IDs), while edge IDs are remapped via `chunk_id_map.values()` set membership. The two code paths do the same thing but with different mechanisms. Consider using a shared helper for consistency.

### 9. Frontend type check looks good

The `typeof value === "boolean"` addition to `transformTemplateToPayload` correctly handles the `rechunk: false` case. No issues there.

PR #17550, opened 2026-07-30
fix: restore provider-specific fields on form echo via echoTransform

dcc123456:
### Summary

fix: restore provider-specific fields on form echo via echoTransform

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/hooks.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/hooks.tsx
@@ -160,14 +160,20 @@ export function useProviderBaseUrlOptions(providerName: string) {
  * - Draft: empty form, with `base_url` pre-filled from the
  *   provider's `default` URL when available.
  * - Saved: prefer `instanceDetails` (which carries api_key / base_url);
- *   normalise api_key into the bare key plus any nested credential
- *   fields (see {@link unwrapApiKey}).
+ *   when `echoTransform` is supplied (provider-specific field mapping,
+ *   e.g. Google Cloud's top-level `google_project_id` or XunFei Spark's
+ *   nested `spark_api_password`), use it to reverse the corresponding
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
## PR Review: #17550

### Summary
This PR adds `echoTransform` to the provider config system, restoring provider-specific credential fields (Google Cloud project_id, XunFei Spark app_id/password, PaddleOCR config, MinerU config, etc.) when editing a saved model instance. Previously, only the generic `unwrapApiKey` path existed, which missed provider-specific fields that live outside the standard `api_key` structure.

### What I checked
- **Correctness**: Each `echoTransform` correctly mirrors its corresponding `submitTransform`, reversing the field mapping so saved values pre-fill the form. The approach is the right abstraction — each provider declares its own echo logic rather than adding ad-hoc special cases in the shared hook.
- **Backward compatibility**: When `echoTransform` is absent, the existing `unwrapApiKey` fallback path runs unchanged. ✅
- **The `parseApiKeyAsObject` helper** handles JSON strings, already-parsed objects, and bare strings gracefully, returning `undefined` on unparseable input so callers can safely use `??`. ✅
- **Edge cases**: The `mineru_delete_output` boolean normalization (`"1"` or `true` → `true`) correctly mirrors how checkboxes submit. ✅

### Minor observation (non-blocking)
In `useProviderInitialValues`, `echoTransform` and `unwrapApiKey` are mutually exclusive — when `echoTransform` is defined, `unwrapApiKey` is skipped entirely. This is intentional since `echoTransform` handles the full credential mapping for providers that opted in. Just worth documenting: any provider adding `echoTransform` must also handle ALL credential fields (not just the non-`api_key` ones), which is correctly done in each provider config here.

### Verdict
✅ **LGTM**. Clean, well-scoped fix with proper backward compatibility. The structural choice (echoTransform on ProviderConfig) is maintainable and scales well. Approved.

PR #17551, opened 2026-07-30
fix(i18n): add missing translation keys for model provider form fields

dcc123456:
### Summary

fix(i18n): add missing translation keys for model provider form fields


wangq8 (commented):
LGTM! Simple and correct i18n fix.

**What this fixes:**
- Component code in `provider-config-map.ts` already references `FishAudioAKMessage` and `FishAudioRefIDMessage` (without the `add` prefix), but the locale files had `addFishAudioAKMessage` and `addFishAudioRefIDMessage`. This mismatch meant those translation keys were silently broken.
- Adds missing `apiKeyPlaceholder` key (already referenced by components like `google-form/index.tsx`).
- Adds missing `paddleocrSelectAlgorithm` key.

**One note (non-blocking):**
Other locale files (`pt-br`, `bg`, `vi`, `ru`, `ja`, `fr`, `es`, `ko`, `de`) still have the old `addFishAudio*` key names — they'll need the same rename in a follow-up. But that's fine to handle separately.

Thanks!

PR #17552, opened 2026-07-30
 Feat: batch verify models in provider instance card

dcc123456:
### Summary

 Feat: batch verify models in provider instance card                                                           
                                                                                                                
  Add a "Batch verify" action to the models section so users can verify
  multiple models in one pass instead of triggering per-model verify                                            
  one at a time.


wangq8 (commented):
Thanks for this PR! The batch verify feature is a nice UX improvement. Here are my review notes:

**What I like:**
- Good extraction of `buildVerifyArgs` to deduplicate credential resolution logic between single and batch verify paths — clean and maintainable.
- Chunked parallel execution (3 concurrent) with `Promise.allSettled` is a solid approach for robustness.
- Proper use of `useAxiosNativeConfig=true` + `skipGlobalErrorNotification` to suppress per-call error toasts during batch verify while keeping individual error feedback.
- Good `useMemo` usage for `filteredSelectedCount` — avoids unnecessary recalculations.

- Between chunks, models in future chunks display `idle` status even though the overall batch is still running. Consider showing a "queued" or indeterminate state for remaining models during an active batch, so users aren't confused that the batch seems "stuck" after the first chunk completes.
- `skipGlobalErrorNotification` is accessed via `(response.config as any)` in `next-request.ts`. It works, but adding a proper typed property to the axios config interface would be cleaner. Not a blocker.

**One question:**
- The selection state is not cleared after batch verify completes — is this intentional? Users might want to re-run verification on the same set, but it could also lead to accidentally re-verifying. Fine either way, just want to confirm.

Overall LGTM. Thanks for the contribution!

PR #17554, opened 2026-07-30
Fix internal/handler test failures, align response with Python contract, and re-enable handler/storage/agent tests in CI

xugangqiang:
## Summary

Fixes the 6 pre-existing failures in `internal/handler`, aligns one response
field with the Python API contract, and re-enables packages in CI that were
previously excluded because of (or unrelated to) those failures.

### Test/handler fixes
- **plugin.go**: return `"success"` (lowercase) instead of `"SUCCESS"` so the
  response envelope matches the Python backend, keeping backend-swap transparent.
- **search_handler_test.go**: expect lowercase `"no authorization"` to match the
  service error message, which mirrors Python.
- **agent_test.go**: seed versions with `CreateTime` instead of `UpdateTime` so
  `ListVersions` ordering (`create_time DESC`) is exercised correctly.
- **agent_wait_for_user_test.go**: a clean run may emit only the `[DONE]` frame;
  relax the SSE assertion to require a non-empty stream ending in `[DONE]`.
- **bot_test.go**: align attachment-download assertions with actual behavior
  (`Content-Disposition: attachment; filename="file"`, default
  `application/octet-stream`), matching Python `resolve_attachment_content_type`.

### Additional scope: dataset entity alignment (not handler-test related)
These two changes are included here because they were discovered while the
handler tests exercised the go dataset creation path; they are correctness
fixes independent of the test work and are called out so the history is clear:
- **internal/entity/dataset.go**: rename `ArtifactTaskID`/`ArtifactTaskFinishAt`
  to `WikiTaskID`/`WikiTaskFinishAt` (columns `wiki_task_id`/`wiki_task_finish_at`)
  to follow the Python/api column rename from PR #17553. Without this, every Go
  `CreateDataset` INSERT fails with `Unknown column artifact_task_id`.
- **internal/service/dataset/crud.go**: surface the real underlying DB error on
  `CreateDataset` failure (`common.Error` + `%w`) instead of masking it with the
  generic `"failed to save dataset"`, so schema/constraint mismatches are
  diagnosable rather than hidden.

### CI
- **tests.yml / sep-tests.yml**:
  - Remove the `grep -v '/internal/handler$'` filter — the 6 failures that
    motivated it are now fixed.
  - Remove the `grep -v '/internal/storage$'` filter — `internal/storage`
    tests self-skip via `t.Skipf` when MinIO is unavailable, so the package is
    safe to run in CI.
  - Remove the `grep -v '/internal/agent$'` filter (only present in the
    tests.yml infinity job) — the exclusion was undocumented and inconsistent
    with the other jobs that already run `internal/agent`.
  - Keep the `internal/tokenizer` exclusion: it is a genuine environmental
    dependency (dict files at `/usr/share/infinity/resource`, absent in the Go
    test environment).

## Test plan

- `build.sh --test ./internal/handler/` is green (previously 6 FAIL).
- `build.sh --test ./internal/storage/ ./internal/agent/` should pass; the
  MinIO-backed `internal/storage` tests skip gracefully without a MinIO server.
- After this PR, `internal/handler`, `internal/storage`, and `internal/agent`
  run again in CI unit-test jobs; `internal/tokenizer` stays excluded.


xugangqiang:
@yingfeng @yuzhichang @JinHai-CN  pls help to review

PR #17556, opened 2026-07-30
Fix: Avoid deepcopy crash for OpenAI‑API‑compatible chat models

Lynn-Inf:
### Summary

As title.


wangq8 (approved):
LGTM. Clean fix.

The `deepcopy(chat_mdl)` call crashes for some OpenAI-API-compatible chat models because LLMBundle wraps model clients that contain non-copyable resources (httpx clients, connection pools, etc.). Switching to a direct reference assignment (`self.chat_mdl = chat_mdl`) avoids this while remaining functionally equivalent — `self.chat_mdl` is used only in read-only patterns throughout the class (`async_chat`, `max_length`, existence checks). None of the usage mutates the underlying model object.

This approach is also consistent with how other classes in the codebase handle `chat_mdl`:
- `tree_structured_query_decomposition_retrieval.py`: `self.chat_mdl = chat_mdl`
- `structure.py`: `self._chat_mdl = chat_mdl`

No regressions expected. Import cleanup (`from copy import deepcopy`) is correct — `deepcopy` is no longer used in this module.

PR #17559, opened 2026-07-30
fix(setting-model): support field-level autoComplete to suppress browser autofill

dcc123456:
### Summary

fix(setting-model): support field-level autoComplete to suppress browser autofill

The generic API key modal was being autofilled by Chrome - e.g. group_id showed a previously-saved base_url value (https://somark.tech/api/v1). Chrome treats forms containing a password field (api_key) as login forms and ignores autocomplete="off", so "new-password" is the reliable value.

Add an optional `autoComplete` field to FieldConfig / FormFieldConfig, thread it through use-provider-fields, and set autoComplete="new-password" on all GenericApiKeyConfig fields (instance_name, api_key, base_url, group_id). Configs that omit autoComplete keep the original behavior.


wangq8 (approved):
Thanks for this fix! Clean and well-scoped.

**What works well:**

- `autoComplete` is optional everywhere — zero breaking changes for existing provider configs.
- Using `"new-password"` is the correct workaround for Chrome ignoring `autocomplete="off"` on forms containing a password field (the `api_key` field triggers this). The PR description explains the rationale clearly.
- Setting `autoComplete` on all fields in the form (not just the password field) is the right call — Chrome will happily autofill non-password fields too when it treats the form as a login form.
- JSDoc comments on both `FieldConfig` and `FormFieldConfig` are helpful — future config authors will know when to use this.
- Threading is complete: types → use-provider-fields → dynamic-form → Input. No gaps.

**One suggestion for follow-up (non-blocking):**
Other provider configs with Password-type fields (e.g. `api_key` fields in specific provider configs like OpenAI, DeepSeek, etc.) could benefit from the same fix. This PR establishes the infrastructure — worth a quick scan later to add `autoComplete` to those as well.

LGTM 🚀

PR #17566, opened 2026-07-30
Fix multiple model chat, the content overwrite the current chat session

wangq8:
go in this PR: https://github.com/infiniflow/ragflow/pull/17637

### DIFF
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -1161,6 +1161,9 @@ async def session_completion(chat_id_in_arg=""):
         return error
     request_messages, request_msg = normalized
     pass_all_history_messages = _get_bool_request_flag(req, "pass_all_history_messages", "pass_all_history", default=False)
+    store_history_messages = _get_bool_request_flag(req, "store_history_messages", "store_history", default=True)
+    if not store_history_messages and not pass_all_history_messages:
+        return get_data_error_result(message="`pass_all_history_messages` must be true when `store_history_messages` is false.")
     msg = request_msg
     message_id = request_msg[-1].get("id")
     chat_id = req.pop("chat_id", "") or ""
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
## PR Review: Fix multiple model chat content overwrite

### Summary
This PR fixes a bug where multiple-model chat sub-conversations were overwriting the current chat session by introducing a `store_history_messages` flag that prevents conversation persistence on a per-request basis.

### What works well

1. **Correct fix for the root cause**: Each ChatCard in the multi-model view now sends `store_history_messages: false`, preventing sub-conversations from persisting their state and clobbering the main session history.

2. **Clean API design**: The `store_history_messages` parameter defaults to `true` (backward compatible), and the validation enforces that `pass_all_history_messages` must be `true` when history isn't stored — which is logically sound since you need to pass full context if you are not persisting it.

3. **Properly gated**: Both the streaming path and the non-streaming path check `store_history_messages` before calling `ConversationService.update_by_id`. No missed code paths.

4. **Parameter flow verified**: `store_history_messages: false` flows correctly: ChatCard -> `handlePressEnter`/`sendMessage` -> API request body -> `chat_api.py`. The `...params` spread in `use-send-single-message.ts` handles this transparently.

### Issues / Suggestions

1. **Go API needs the same fix**: The PR body mentions "TODO: need to porting to go." — and I confirmed `internal/service/chat_session.go` (`ChatCompletions`) does not have `store_history_messages` support. This fix is incomplete without the Go counterpart. Consider doing this in the same PR or filing a follow-up issue.

2. **Parent `handlePressEnter` silently drops `store_history_messages`**: In `next-multiple-chat-box.tsx`, the parent `handlePressEnter` destructures only `{ enableInternet, enableThinking }` and forwards only those to each ChatCard. This is harmless now because each ChatCard hardcodes `store_history_messages: false`, but if the API ever needs this to be dynamic from the main input, the param would be silently dropped. Consider adding a comment or forwarding it explicitly for future-proofing.

3. **`pass_all_history_messages` is hardcoded to `true` in `use-send-single-message.ts`**: This means the validation on the server side (`store_history_messages=false && pass_all_history_messages=false`) can never be triggered from the current frontend code path. That is fine for multi-model (the hardcoded `true` is intentional), but if someone later tries to use `store_history_messages=false` without `pass_all_history_messages=true` from a different caller, they would hit the error. This is arguably correct defensive validation.

### Verdict
Approve with suggestion — the fix is correct and well-scoped. The Go porting TODO should be tracked to avoid a split-brain between Python and Go APIs.

PR #17568, opened 2026-07-30
Fix: Add minerU.net

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the PR! Overall the integration with MinerU.Net API looks clean and well-structured. I have a few concerns:

### Issues to address

**1. Table section concatenation missing newline separators (`ocr_model.py` L139)**

```python
section = item.get("table_body", "") + "\n".join(item.get("table_caption", [])) + "\n".join(item.get("table_footnote", []))
```

When `table_body` is non-empty and `table_caption` / `table_footnote` also have content, the strings are concatenated directly without any separator. For example, `"foo"` + `"bar"` = `"foobar"`. Should probably insert `"\n"` between the parts, e.g.:

```python
parts = [item.get("table_body", "")]
parts.extend(item.get("table_caption", []))
parts.extend(item.get("table_footnote", []))
section = "\n".join(p for p in parts if p)
```

**2. Loose binary handling in `_submit_by_upload` (L190-193)**

```python
file_obj = io.BytesIO(binary.read() if hasattr(binary, "read") else binary)
```

The final `else binary` fallback tries to wrap any non-bytes, non-file-like value in `BytesIO`, which would fail confusingly if `binary` is, say, a `str` (not a URL path). Consider making this explicit — raise a `ValueError` for unexpected types, or handle the cases you expect (bytes, file-like with `.read()`, and Path).

**3. Connection-check endpoint returning 404 as "valid" (L119-127)**

`/api/v4/extract/task/connection-check` returning 404 is treated as success. If this is genuinely a dedicated health-check endpoint, a 404 seems like it should be an error. Could you clarify the API behavior here? If the endpoint really returns 404 for auth success, a comment explaining why would help future readers.

**4. Unrelated formatting-only changes in web files**

Both `web/src/pages/user-setting/setting-model/instance-card/models-section/index.tsx` and `web/src/utils/next-request.ts` contain only whitespace/formatting changes that appear unrelated to "Add minerU.net". Consider dropping these from the PR to keep the diff focused, or call them out explicitly if they are intentional.

### Suggestions (non-blocking)

- Consider adding retry logic (e.g., `requests.Session` with `Retry`) for transient network errors during submission and polling.
- The polling progress callback ties progress to elapsed time rather than actual page extraction progress. Using `extracted/total` ratio would give a more accurate progress bar.

Otherwise, the implementation looks solid — the parse step correctly maps MinerU content types to sections, and the factory registration pattern matches the existing MinerU/PaddleOCR conventions.

PR #17569, opened 2026-07-30
Fix parse image in excel as table, the image shows as one column

wangq8:
TODO: need to porting to go.
Before:
image

Now:
image

### DIFF
--- a/rag/app/table.py
+++ b/rag/app/table.py
@@ -83,7 +83,8 @@ class Excel(ExcelParser):
                 image_descriptions = vision_figure_parser_figure_xlsx_wrapper(images=images, callback=callback, **kwargs)
                 if image_descriptions and len(image_descriptions) == len(images):
                     for i, bf in enumerate(image_descriptions):
-                        images[i]["image_description"] = "\n".join(bf[0][1])
+                        desc = bf[0][1]
+                        images[i]["image_description"] = "\n".join(desc) if isinstance(desc, list) else str(desc)
                     for img in images:
                         if img["span_type"] == "single_cell" and img.get("image_description"):
                             pending_cell_images.append(img)
--- a/test/unit_test/rag/app/test_table_chunk_column_roles.py
+++ b/test/unit_test/rag/app/test_table_chunk_column_roles.py
@@ -20,6 +20,7 @@
 from __future__ import annotations
 
 import sys
+from io import BytesIO
 from importlib import import_module, reload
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
## Review: PR #17569 — Fix Excel image parsed as table column

**Author:** @wangq8  
**Summary:** Fixes a bug where images embedded in Excel sheets were being rendered as individual character columns instead of a single cell value.

---

### ✅ What the Bug Was

In `rag/app/table.py:86`, the original code does:

```python
images[i]["image_description"] = "\n".join(bf[0][1])
```

`"\n".join()` on a **string** joins each character with newlines. So a description like `"abcdef"` becomes:

```
a
b
c
d
e
f
```

Each character then gets treated as a separate column entry, which is the bug captured in the before screenshot.

### ✅ The Fix (rag/app/table.py)

```python
desc = bf[0][1]
images[i]["image_description"] = "\n".join(desc) if isinstance(desc, list) else str(desc)
```

Clean and correct:
- If `desc` is a `list` → join as before (preserving existing behavior)
- If `desc` is a `str` (or anything else) → use it directly via `str()`

**Correctness:** ✔️ The `isinstance` guard is the right approach. `str(desc)` is a safe fallback for non-list, non-string types.

**Risk:** Very low — 1-line logic change, well-scoped, no side effects on the existing list path.

### ✅ Test Coverage (test_table_chunk_column_roles.py)

The new test `test_excel_image_description_string_stays_single_cell` is well-written:
- Creates a minimal Excel workbook in memory via openpyxl
- Mocks `_extract_images_from_worksheet` to return a single-cell image
- Mocks `vision_figure_parser_figure_xlsx_wrapper` to return a **string** description (`"abcdef"`)
- Asserts the result `iat[0, 0]` equals `"abcdef"` (not split into characters)

This directly reproduces the bug scenario and validates the fix.

---

### 🔵 Minor Nitpick (non-blocking)

CodeRabbit suggested adding a debug-level log in the non-list branch for observability. This is a reasonable suggestion — a `logging.debug(...)` recording the type received would help detect upstream contract drift without including image contents. Not a blocker, but worth considering for future maintainability.

---

### 📋 Verdict

**LGTM ✅** — the fix is minimal, correct, and well-tested. The test specifically reproduces the bug and validates the fix. No regressions or security concerns. Ready to merge.

xugangqiang:
LGTM

PR #17576, opened 2026-07-30
Fix: The save interface is continuously called without any operation being performed on the agent page.

cike8899:
### Summary

Fix: The save interface is continuously called without any operation being performed on the agent page.


wangq8:
LGTM

wangq8 (commented):
Thanks for the fix! The root cause is a classic reactive loop: auto-save → detail invalidation → re-fetch → potential re-render → save again. The approach of returning `update_time` from the backend and skipping the detail query invalidation during auto-save is sensible.

A few observations:

**Backend (`agent_api.py` + `agent.go`):** Returning `update_time` instead of bare `true` is a good API improvement — consistent across both backends and the test mock is correctly updated.

**Frontend (`use-save-graph.ts`):** The `skipInvalidation` parameter correctly prevents the `FetchAgentDetail` invalidation on auto-save while keeping it for manual saves. The `?? Date.now()` fallback on `update_time` is a reasonable safety net.

**Unrelated change:** The `className="ml-auto"` addition in `wiki-left-panel/index.tsx` appears to be a drive-by styling fix unrelated to this PR. Consider moving it to a separate PR or at least calling it out in the PR description.

**Testing:** The backend mock test was updated, but there are no tests covering the new `skipInvalidation` path in `useSetAgent` or the auto-save cycle fix in `useWatchAgentChange`. Consider adding a test that verifies saving with `skipInvalidation: true` does not invalidate the detail query.

**One concern:** `useDebounceEffect` depends on `[nodes, edges]` — with xyflow, node/edge references can change even when data is identical (e.g., layout recalculations). If that is the root trigger for the continuous saves, `skipInvalidation` alone may not be sufficient. Consider whether a deep comparison on `nodes`/`edges` (or comparing `update_time`) before triggering the save would be a more robust fix.

Overall, this is a solid fix that addresses the immediate issue. The `skipInvalidation` approach is clean and well-scoped.

PR #17581, opened 2026-07-30
Fix: multiple model chat pass-in session_id, and do not save session.

wangq8:
go in this PR: https://github.com/infiniflow/ragflow/pull/17637 
multiple model chat, no need to create session and no need to pass in session id, and hide the like buttion.

### DIFF
--- a/web/src/pages/next-chats/hooks/use-send-single-message.ts
+++ b/web/src/pages/next-chats/hooks/use-send-single-message.ts
@@ -60,7 +60,8 @@ export function useSendSingleMessage({
       messages,
       enableInternet,
       enableThinking,
-      omit_session_id,
+      storeHistoryMessages,
+      omitSessionId,
       ...params
     }: {
       message: IMessage;
@@ -72,7 +73,7 @@ export function useSendSingleMessage({
         api.completionUrl,
         {
           chat_id: chatId,
-          ...(omit_session_id ? {} : { session_id: sessionId }),
+          ...(omitSessionId ? {} : { session_id: sessionId }),
           messages: [
             ...(Array.isArray(messages) && messages?.length > 0
               ? messages
@@ -82,6 +83,9 @@ export function useSendSingleMessage({
           reasoning: Number(enableThinking),
           internet: enableInternet,
           ...params,
+          ...(storeHistoryMessages === undefined
+            ? {}
+            : { store_history_messages: storeHistoryMessages }),
           pass_all_history_messages: true,
         },
         controller,
--- a/web/src/pages/next-chats/hooks/use-send-single-message.ts
+++ b/web/src/pages/next-chats/hooks/use-send-single-message.ts
@@ -60,6 +60,7 @@ export function useSendSingleMessage({
       messages,
       enableInternet,
       enableThinking,
+      omit_session_id,
       ...params
     }: {
       message: IMessage;
@@ -71,7 +72,7 @@ export function useSendSingleMessage({
         api.completionUrl,
         {
           chat_id: chatId,
-          session_id: sessionId,
+          ...(omit_session_id ? {} : { session_id: sessionId }),
           messages: [
             ...(Array.isArray(messages) && messages?.length > 0
               ? messages
@@ -112,7 +113,7 @@ export function useSendSingleMessage({
       targetConversationId,
       ...params
     }: NextMessageInputOnPressEnterParameter &
-      CreateConversationBeforeSendMessageReturnType) => {
+      Partial<NonNullable<CreateConversationBeforeSendMessageReturnType>>) => {
       if (trim(value) === '' || !done) return;
       const id = uuid();
 
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -181,7 +181,7 @@ def _build_default_completion_dialog():
     )
 
 
-async def _create_session_for_completion(chat_id, dialog, user_id):
+async def _create_session_for_completion(chat_id, dialog, user_id, save_session=True):
     conv = {
         "id": get_uuid(),
         "dialog_id": chat_id,
@@ -190,6 +190,9 @@ async def _create_session_for_completion(chat_id, dialog, user_id):
         "user_id": user_id,
         "reference": [],
     }
+    if not save_session:
+        conv["id"] = None
+        return SimpleNamespace(**conv)
     await thread_pool_exec(ConversationService.save, **conv)
     ok, conv_obj = await thread_pool_exec(ConversationService.get_by_id, conv["id"])
     if not ok:
@@ -1195,7 +1198,7 @@ async def session_completion(chat_id_in_arg=""):
                 if conv.dialog_id != chat_id:
                     return get_data_error_result(message="Session does not belong to this chat!")
             else:
-                conv = await _create_session_for_completion(chat_id, dia, current_user.id)
+                conv = await _create_session_for_completion(chat_id, dia, current_user.id, save_session=store_history_messages)
                 session_id = conv.id
 
             if pass_all_history_messages:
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
## PR Review: Fix multiple model chat pass-in session_id, and do not save session

### Summary
This PR removes unnecessary session/conversation creation in multi-model chat mode and hides the like button. The approach is sensible overall, but there are critical issues that need attention.

### 🔴 Critical Issue

**`store_history_messages` is undefined** in `api/apps/restful_apis/chat_api.py`:

```python
conv = await _create_session_for_completion(chat_id, dia, current_user.id, save_session=store_history_messages)
```

The frontend sends `store_history_messages: false` in the request body (spread via `...params`), but the backend never extracts this value from `req`. This will raise a `NameError` at runtime. You need to parse it from the request, similar to how `pass_all_history_messages` is extracted:

```python
store_history_messages = _get_bool_request_flag(req, "store_history_messages", default=True)
```

Add this near line ~1163 alongside the other request parameter extractions.

### 🟡 Suggestions

1. **`showLikeButton={false}`** — This is a UX change unrelated to the session_id fix. Not a blocker, but ideally it should be mentioned in the PR description or split into a separate PR for cleaner history.

2. **`omit_session_id` type addition** — The type is added to `NextMessageInputOnPressEnterParameter` but only used in multi-chat. Consider documenting in the type that this field exists to prevent confusion for future consumers.

### ✅ Good

- The `_create_session_for_completion(save_session=...)` pattern is clean and backwards-compatible.
- Using `SimpleNamespace` as a lightweight non-persisted return avoids unnecessary DB writes.
- Frontend correctly removes the `useCreateConversationBeforeSendMessage` call, which was the root cause of unnecessary session creation in multi-chat mode.
- The partial type change in `use-send-single-message.ts` is correct and aligns with the removed session dependency.

web/src/pages/next-chats/hooks/use-send-single-message.ts
```diff
@@ -71,7 +72,7 @@ export function useSendSingleMessage({
         api.completionUrl,
         {
           chat_id: chatId,
-          session_id: sessionId,
+          ...(omit_session_id ? {} : { session_id: sessionId }),
```
cike8899: Use camelCase naming conventions.

wangq8 (commented):
## Updated Review — My apologies for the earlier false alarm

I re-ran the diff against the actual branch and my initial review missed several hunks. Here is a corrected assessment.

### ✅ Overall: This PR is solid and well-designed.

**Backend (`chat_api.py`):**
- `store_history_messages` is properly parsed via `_get_bool_request_flag` (line 1167) and validated (lines 1168-1169). My earlier concern was wrong.
- The `_create_session_for_completion(save_session=...)` change is clean and backwards-compatible.
- When `save_session=False`, returning `SimpleNamespace(**conv)` avoids a useless DB write.
- Session update guard (`if conv is not None and store_history_messages:`) at lines 1305 and 1323 correctly prevents calling `conv.to_dict()` on a `SimpleNamespace`.
- `session_id=None` when `store_history_messages=False` is handled gracefully downstream — both `async_chat` and `rag_agent` treat it as optional (Langfuse tracing ID), guarded with `if session_id:`.

**Frontend:**
- `next-multiple-chat-box.tsx`: Correctly removes `useCreateConversationBeforeSendMessage`, which was the root cause of spurious session creation in multi-model mode.
- `use-send-single-message.ts`: `omit_session_id` cleanly omits the `session_id` field from the API request.
- `store_history_messages: false` is properly passed from multiple chat box.

### 🟡 Minor suggestions

1. **`showLikeButton={false}`** — Hiding the like button in multi-chat mode is reasonable, but it is unrelated to the session_id/persistence fix. Consider splitting into a separate PR next time.

2. **Validation message wording** — The error `"pass_all_history_messages must be true when store_history_messages is false"` is correct, but a consumer reading it without context might not understand why. A brief comment in the code explaining the constraint would help. (This is enforced because you cannot meaningfully append messages to a non-persisted conversation — `pass_all_history_messages` sends the full history each time.)

3. **`SimpleNamespace` attributes** — When `save_session=False`, the returned `SimpleNamespace` has no `message` attribute. It works because the `pass_all_history_messages=true` branch sets `conv.message = deepcopy(request_messages)` before any read access. But if someone later refactors the order of operations, this could break. A comment noting this implicit coupling would be helpful.

### 🔍 Verified edge cases

| Scenario | Result |
|---|---|
| `session_id=None` → `rag_agent` | ✅ Safe, `if session_id:` guard |
| `conv.to_dict()` on `SimpleNamespace` | ✅ Never reached, guarded by `store_history_messages` |
| `conv.reference` on `SimpleNamespace` | ✅ `SimpleNamespace` has it from dict |
| Normal flow (save_session=True) | ✅ Unchanged behavior |
| Multi-chat with no session | ✅ No session created, no DB write |

### Verdict: LGTM after minor doc improvements. No blocking issues.

cike8899 (approved):

wangq8 (commented):
## PR Review: Fix multiple model chat pass-in session_id

### Summary
This PR fixes the bug where the multiple model comparison chat was incorrectly creating and sharing session IDs across all model cards. The fix is clean and minimal: the backend skips session persistence when `store_history_messages=False`, and the frontend omits `session_id` from the API request for multiple model chats.

### ✅ What works well

1. **Correct fix**: Multiple model chat no longer creates unnecessary sessions. Each card sends messages without a `session_id`, and the backend creates a temporary in-memory conv object without persisting to DB.

2. **`save_session` flag is clean**: Adding `save_session=True` default to `_create_session_for_completion` keeps backward compatibility — regular chat still saves sessions normally.

3. **Proper guard chain**: `conv.to_dict()` calls at lines 1334 and 1352 are guarded by `store_history_messages`, so the `SimpleNamespace` (which lacks `to_dict`) is never called with `.to_dict()`.

4. **Type changes are backward-compatible**: `NextMessageInputOnPressEnterParameter` adds optional fields (`storeHistoryMessages`, `omitSessionId`) that other consumers (`use-send-chat-message.ts`, `use-send-shared-message.ts`) simply ignore. No breakage.

5. **UX polish**: Hiding the like button (`showLikeButton={false}`) in multiple model chat is correct since sessions are not persisted.

6. **Unused code cleanup**: `useCreateConversationBeforeSendMessage` is properly removed along with its import.

### ⚠️ Concerns

1. **Fragile `SimpleNamespace` pattern** (`chat_api.py:190-193`): When `save_session=False`, a `SimpleNamespace` is returned that lacks `to_dict()`. The correctness relies entirely on the `store_history_messages` guard. If a future change introduces a `conv.to_dict()` call before or outside this guard, it will crash at runtime (AttributeError). Consider either:
   - Adding a `to_dict` stub to the `SimpleNamespace` object, or
   - Using a more explicit sentinel/flag instead of substituting a different type.

2. **`session_id: null` in SSE stream responses**: `structure_answer` sets `ans["session_id"] = None` when session is not saved. While the multiple chat frontend likely ignores this, it is a deviation from the normal response shape. Worth confirming this is intentional for downstream consumers.

3. **`pass_all_history_messages: true` always sent from frontend** (`use-send-single-message.ts`): The request unconditionally sends `pass_all_history_messages: true`, even for the multiple chat path. The backend validates this correctly against `store_history_messages=False`, but coupling these two flags implicitly could be tightened in the future.

4. **PR description mentions Go porting**: The body says "TODO: need porting to go." — if there is a Go backend that mirrors this API, the equivalent change needs to land there as well.

### 📋 Testing suggestions

- Verify multiple model chat: send messages with 2+ model cards, ensure no sessions appear in the chat history
- Verify regular chat still creates and saves sessions properly (regression check)
- Verify `pass_all_history_messages` with multiple model chat continues all prior messages correctly
- Verify like button is hidden in multiple model chat
- Verify session listing does not show transient sessions from multiple model chat

### Verdict

**Approve with notes.** The fix is correct and well-scoped. The `SimpleNamespace` pattern is the primary fragility concern — not a bug today, but a maintenance risk. Consider the suggested hardening in a follow-up.

PR #17582, opened 2026-07-30
refactor(setting-model): migrate SoMark to generic provider card withmodel-level extra fields

dcc123456:
### Summary

refactor(setting-model): migrate SoMark to generic provider card withmodel-level extra fields

  Remove the standalone SoMarkInstanceCard (722 lines) and route SoMark
  through the generic GenericProviderInstanceCard path. SoMark-specific
  fields (4 element-format selects + 7 feature-config toggles) are now
  edited in the model edit dialog (AddCustomModelDialog) and stored
  per-model in model_info[].extra.


wangq8 (commented):
Nice refactoring — removing 722 lines of SoMark-specific card code and routing through the generic path is the right direction. The new `select`/`switch`/`section` field types in `AddCustomModelDialog` make this pattern reusable for future providers too. A few points to address:

---

### 1. 🔴 Commented-out `startRename()` in `saved-mode-card.tsx`

```ts
onDoubleClick={(e) => {
  e.stopPropagation();
  // startRename();
}}
```

The title still says `editInstanceName`, but double-clicking now does nothing useful (just `stopPropagation`). If rename is being intentionally disabled:
- Remove the `onDoubleClick` handler and the `title` attribute entirely
- If the function/state for rename is no longer used, clean up the related code too

If this is accidental, restore `startRename()`.

---

### 2. 🟡 Dialog width change affects all providers

`max-w-md` → `max-w-2xl` widens the add/edit model dialog for **every** provider, not just SoMark. For simple providers with only name/max_tokens/model_types/features, the dialog is now unnecessarily wide. Consider making the width conditional on whether `providerFeatureKeys` is provided (or compute a `max-w` class via `className` prop).

---

### 3. 🟡 SoMark `model_types` no longer defaults to `[ocr]`

The legacy `SoMarkInstanceCard` hardcoded `model_type: [ocr]` in `buildPayload`. Now users must manually select "OCR" from the model_types multi-select, with no default. This is a silent behavior change:

```ts
// In useCustomModelFields — consider adding a SoMark-specific default:
if (isSoMark && field.name === model_types) {
  mapped.defaultValue = [ocr];
}
```

Or alternatively, accept `defaultValues` override from the caller.

---

### 4. 🟢 Minor: Spelling in comment

```ts
/** Whitelist of provider-specific feature keys ... */
//            ^^
```
Convention in the codebase has been "allowlist" — not a blocker, just consistency.

---

### 5. 🟢 Edit dialog features reconstruction looks correct

The `editDefaultValues` logic that rebuilds the `features` array from `extra` booleans (matching `is_tools` + `providerFeatureKeys`) and the `handleEditSubmit` optimistic cache patch that properly separates `is_tools` from `item.extra` — both look correct. The draft edit path (`updateDraftModel` instead of PATCH) is also handled well.

---

### 6. 🟢 Data shape backward compatibility

Existing SoMark instances store `somark_*` fields as booleans in `model_info[0].extra`. The new approach stores them the same way (via `item.extra` in `handleSubmit`), so no migration is needed. `buildModelInfo` and `useModelsDerived` correctly propagate `extra` through the echo path. ✅

---

**Summary:** Good refactor, just the 3 flagged items to address. The rename regression (#1) is the most impactful — it effectively breaks the inline rename UX for ALL saved instance cards.

PR #17583, opened 2026-07-30
Fix: return model.extra for response display

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the PR. The change looks good to me.

**Analysis:**

- `model_extra` is parsed from `json.loads(model.extra)` where `model.extra` defaults to `"{}"` (db model), so it is always a valid dict at this point.
- This simply exposes the full model extra configuration to the frontend alongside the already-extracted individual fields (`max_tokens`, `verify`, `features`, etc.).
- No regression risk: existing keys are untouched, only a new `"extra"` key is added.
- No security concern: `model.extra` stores model configuration metadata (max_tokens, features flags, verify status, etc.), not secrets/credentials.

**Nit (non-blocking):** Since `model_extra` is now included wholesale, the frontend could potentially derive `max_tokens`, `verify`, and `features` from it directly in the future, simplifying the response shape. Something to consider for a follow-up cleanup.

✅ LGTM, approved.

PR #17584, opened 2026-07-30
fix: make cross-language retrieval deterministic

### LINKED ISSUE #17296: [Bug]: Inconsistent retrieval results between Knowledge Retrieval Test and Agent Retrieval with identical configurations
### Self Checks

Official Docker image

### RAGFlow image version

v0.26.4

```Markdown
Environment
RAGFlow Version: v0.26.4
Deployment: Docker
Knowledge Base:
Same knowledge base used for both tests

Embedding Model:
(Qwen3 Embedding)

Rerank Model:
(Qwen3-Reranker-4B)
```

### Actual behavior

Bug Description

When using the same knowledge base and the same retrieval-related configurations, the retrieval results returned by Knowledge Retrieval Test and Agent Retrieval are inconsistent.

Although both features are expected to use the same retrieval pipeline, the returned Top-N results and ranking order are different for the same query.

This inconsistency makes it difficult to evaluate retrieval quality, because the retrieval test does not accurately reflect the behavior of the Agent during actual use.

### Expected behavior

Knowledge Retrieval Test and Agent Retrieval should return identical retrieval results when all retrieval configurations are the same.

Specifically:

Same candidate documents
Same ranking order
Same similarity scores (or equivalent ordering)

The Retrieval Test should accurately represent the Agent's retrieval behavior.

### Steps to reproduce

```Markdown
Steps to Reproduce
Create a knowledge base.
Configure the Agent to use this knowledge base.
Configure the retrieval parameters.
Same Embedding model
Same Rerank model
Same Similarity Weight
Same Top-N
Same Top-K
Use exactly the same query.
Compare:
Knowledge Retrieval Test results
Agent Retrieval results
```

### Additional information

_No response_

d3f4w2:
### Summary

Fixes #17296.

The Knowledge Retrieval Test and Agent Retrieval both invoke cross-language query expansion before calling the shared retriever. That expansion used `temperature: 0.2`, so two otherwise identical retrieval runs could send different translated queries to embedding and reranking, producing different candidates and ordering.

This change:
- makes cross-language query expansion deterministic by using zero-temperature generation;
- adds a regression test that captures the LLM generation config and verifies the normalized multilingual query output;
- leaves retrieval behavior unchanged when cross-language search is disabled.

### Root cause

The divergence was introduced before retrieval: cross-language expansion sampled an LLM independently in the Test and Agent paths. Visible retrieval settings could therefore match while the actual query passed to `settings.retriever.retrieval()` differed.

### Checks

- `python -m pytest test/unit_test/rag/prompts/test_generator_message_fit_in.py -q` (6 passed)
- `ruff check --select E4,E7,E9,F rag/prompts/generator.py test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `ruff check test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `python -m py_compile rag/prompts/generator.py test/unit_test/rag/prompts/test_generator_message_fit_in.py`
- `git diff --check`

### DIFF
--- a/rag/prompts/generator.py
+++ b/rag/prompts/generator.py
@@ -303,7 +303,9 @@ async def cross_languages(tenant_id, llm_id, query, languages=[]):
     rendered_sys_prompt = PROMPT_JINJA_ENV.from_string(CROSS_LANGUAGES_SYS_PROMPT_TEMPLATE).render()
     rendered_user_prompt = PROMPT_JINJA_ENV.from_string(CROSS_LANGUAGES_USER_PROMPT_TEMPLATE).render(query=query, languages=languages)
 
-    ans = await chat_mdl.async_chat(rendered_sys_prompt, [{"role": "user", "content": rendered_user_prompt}], {"temperature": 0.2})
+    # Cross-language expansion feeds directly into retrieval. Sampling here can
+    # make the retrieval test and an agent search different rewritten queries.
+    ans = await chat_mdl.async_chat(rendered_sys_prompt, [{"role": "user", "content": rendered_user_prompt}], {"temperature": 0})
     if ans.find("**ERROR**") >= 0:
         logging.info("[cross_languages] LLM returned error, falling back to original query")
         return query
--- a/test/unit_test/rag/prompts/test_generator_message_fit_in.py
+++ b/test/unit_test/rag/prompts/test_generator_message_fit_in.py
@@ -13,6 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
+import asyncio
 import importlib.util
 import sys
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


d3f4w2:
@coderabbitai review

d3f4w2:
Addressed the review nit in fdfd27fe0: the regression now verifies that the rendered outbound prompt includes the original query and both requested languages. The focused test remains 6/6 green, with Ruff and py_compile passing.

wangq8 (commented):
This is an AI review comment.

Thanks for the fix — the root cause is right: `cross_languages` was sampling with `temperature: 0.2`, so the Knowledge Retrieval Test and Agent retrieval paths could send different rewritten queries for otherwise identical settings, producing divergent candidates/ordering before the shared retriever even runs. Setting `temperature: 0` (greedy decoding) is the correct, minimal change, and it correctly leaves the disabled-search path untouched.

The regression test is well-constructed: it captures the LLM generation config, asserts `{"temperature": 0}`, and verifies the normalized multilingual output, so a revert to `0.2` would be caught.

One minor note (non-blocking): a few OpenAI-compatible providers treat `temperature: 0` loosely (e.g. some endpoints coerce 0 to a provider default), so determinism ultimately depends on the underlying backend honoring it. `0` is the conventional deterministic setting and this is the right call, but if strict reproducibility ever becomes load-bearing, it may be worth a follow-up verifying the specific providers used for cross-language expansion actually honor it.

Otherwise LGTM — correct, minimal, and well-tested.

This is an AI review comment.

qinling0210 (commented):
@d3f4w2 , I think it is not necessary to change temperature from 0.2 to 0.  With LLM involved, the result might be different even when temperature is 0 for some LLM.

0.2 provide more flexible for LLM, the result of  cross-language will not change. So I suggest keeping temperature as 0.2

PR #17590, opened 2026-07-30
fix(web): sanitize the parsed markdown tree instead of the string (GHSA-jv5f-3mqh-cxwf)

kah-ja:
### Summary

The chat and search renderers pass their content to `<Markdown rehypePlugins={[rehypeRaw, ...]}>`, so raw HTML in a message becomes real nodes. The sanitization in front of that pipeline cannot hold, because `preprocessLaTeX()` (`web/src/utils/chat.ts`) decodes `&lt;`, `&gt;` and `&amp;` back into markup *after* `DOMPurify.sanitize()` has run: markup delivered entity-encoded is inert text while the allow-list inspects it, and live markup again by the time `rehype-raw` parses it. `FloatingChatWidgetMarkdown` and `HighLightMarkdown` reach `rehype-raw` with no sanitization at all.

This is the issue reported privately as GHSA-jv5f-3mqh-cxwf (not published yet).

This PR sanitizes the parsed tree instead of the string: `rehype-sanitize` runs after `rehype-raw` and before `rehype-katex`, with one shared schema.

### Why not simply move the string sanitizer to the end of the pipeline

I tried that first and withdrew it (#17589): a sanitizer applied to the *markdown string* cannot be both safe and faithful here. It HTML-escapes text, and the only thing that ever repaired that escaping was the entity decode -- which is the hole itself. Measured in a browser against this plugin stack:

- `$x^2 < y$` renders as a red KaTeX `ParseError`, because KaTeX receives `&lt;`.
- `` `a < b` `` in an inline code span displays the literal `&lt;`.

Sanitizing the tree leaves text nodes untouched, so neither happens. This may also be why the sanitize pass added in #16409 was removed again in #16449 as "ineffective after preprocessLaTeX" -- the ordering was right, the layer was not.

### What changed

- `web/src/constants/markdown-rehype-plugins.ts` (new): `markdownSanitizeSchema` = `hast-util-sanitize`'s `defaultSchema` plus the wrappers `replaceThinkToSection` / `replaceRetrievingToSection` inject (`think`, `retrieving`, `section`, `details`, `summary`) and `className`; exported as a ready `[rehypeSanitize, schema]` tuple.
- The five renderers insert that tuple directly after `rehypeRaw`: `markdown-content`, `next-markdown-content`, `next-search/markdown-content`, `floating-chat-widget-markdown`, `highlight-markdown`.
- `rehype-sanitize@^6.0.0` and `hast-util-sanitize@^5.0.2` added; the lockfile diff is limited to those two entries.

The existing `DOMPurify.sanitize()` calls are left alone on purpose: they still normalise `null`/`undefined` content into the `chat.searching` placeholder, and they cost nothing as a second layer.

### Validation

Measured in headless Chromium against the real stack (react-markdown 9, rehype-raw 7, rehype-katex 7, remark-gfm/math/breaks), comparing `main` at 611e0596 with this branch:

- Attack side: the entity-encoded `<svg><script>` case executes on `main` and produces no element at all here; `<style>` and `<math>` are likewise dropped. No `on*` attribute and no `javascript:` URL survives in any of the 12 vectors tried.
- Rendering side: **0 of 22 markdown constructs change** -- GFM tables, task lists, blockquotes, headings, nested lists, autolinks, inline code, code fences with `Array<number>`, `<think>` and `<details class="think">` wrappers, `<br>`, images, links, citation markers, `$…$` and `\[…\]` math including `\right]` all render byte-identically.
- `npx jest --runInBand`: 5 failed / 71 passed, identical to `main` (the failures are pre-existing `preprocessLaTeX` and `HighLightMarkdown` specs).
- `npx oxlint` clean, `npx oxfmt` reports the touched files correctly formatted.

I did not add a unit test for the schema: the jest transform does not load ESM-only packages, so importing `hast-util-sanitize` in a spec needs `transformIgnorePatterns: ['/node_modules/(?!(hast-util-sanitize|rehype-sanitize|unist-util-position|@ungap)/)']` in `jest.config.ts`. That change makes the currently non-running `HighLightMarkdown` specs execute and fail, so I left it out of this PR. Happy to add both the config line and the spec if you want them here.


cike8899:
@kah-ja Please resolve the conflict.

PR #17596, opened 2026-07-31
Fix: delete base url default value in FunASR

dcc123456:
### Summary

Fix: delete base url default value in FunASR

### DIFF
--- a/web/src/components/ui/input-select.tsx
+++ b/web/src/components/ui/input-select.tsx
@@ -55,6 +55,8 @@ export interface InputSelectProps {
   multi?: boolean;
   /** Type of input: text, number, date, or datetime */
   type?: 'text' | 'number' | 'date' | 'datetime';
+  /** Auto-complete attribute for the input */
+  autoComplete?: string;
 }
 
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this PR! The changes look good overall with a few notes:

**✅ Good:**
- Removing the hardcoded default `http://localhost:8000/v1` for FunASR base URL is the right call — the actual URL depends on the deployment.
- Adding `autoComplete` support to `InputSelect` and wiring it through the field config is clean and consistent with the existing `Input` plumbing.
- `autoComplete="new-password"` is a practical choice for preventing unwanted browser autofill in model config fields.
- No regressions — the component-level default of `"new-password"` is safe.

**⚠️ Nit:**
- The PR title and description only mention FunASR, but the actual scope is broader: a new `autoComplete` prop on `InputSelect` + wiring it across all local LLM providers (via `buildLocalConfig`). Consider updating the description to reflect this — it helps future readers understand why the `InputSelect` component was touched.

Code-wise LGTM. 👍

PR #17600, opened 2026-07-31
Fix ingestor Start() sync.Once re-entrant deadlock

yuzhichang:
Start() wrapped start() in startOnce while startWorkerPool nested the same Once, deadlocking Start forever. Give worker pool its own Once and add a full-path regression test.


yuzhichang:
Superseded by #17599, which now bundles both #17536 regressions (MySQL Error 1101 + ingestor Start deadlock).

PR #17602, opened 2026-07-31
fix(i18n): update thinking status text and add thinking completed feedback

dcc123456:
### Summary

fix(i18n): update thinking status text and add thinking completed feedback


wangq8 (commented):
LGTM! The English locale (`en.ts`) already has `thought: "Thought"` at line 1151, so this adds the missing `thought` key to the Chinese locale. The updated thinking status text (`思考中...` with ellipsis) is also an improvement for UX, providing better feedback that processing is in progress. Thanks!

PR #17603, opened 2026-07-31
fix(dataset): update error message for document deletion to use correct translation key

dcc123456:
### Summary

fix(dataset): update error message for document deletion to use correct translation key


wangq8 (commented):
LGTM! Simple and correct fix — the translation key was missing the `knowledgeConfiguration.` namespace prefix. Good catch.

PR #17604, opened 2026-07-31
Fix: force set enable_thinking to true for preview qwen3 model

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
LGTM! The fix correctly addresses the API constraint where qwen3 preview variants (e.g. `qwen3.8-max-preview`) only accept `enable_thinking=True`.

Good that both paths are updated:
- `knowledge_compile/_common.py` for the knowledge compile code path
- `chat_model.py` for the main LLM chat path

The unit tests cover both the default (no thinking config) and explicit disable (`thinking=disabled`) cases, confirming that `-preview` always forces `True`.

The logic is safely contained within the existing `"qwen3" in model_name` check, so no risk of false positives on non-qwen3 model names.

PR #17606, opened 2026-07-31
Feature: GET /datasets to support ids array

wangq8:
TODO: need porting to go.

Usage:

```
curl 'http://localhost:9222/api/v1/datasets?ids=ec04eb388be411f1bc7c9f6eed4bafa3&ids=6aa0a5a88be111f1bc7c9f6eed4bafa3'
```

image

### DIFF
--- a/api/apps/restful_apis/dataset_api.py
+++ b/api/apps/restful_apis/dataset_api.py
@@ -322,6 +322,14 @@ def list_datasets(tenant_id):
         required: false
         description: Dataset ID to filter.
       - in: query
+        name: ids
+        type: array
+        required: false
+        items:
+          type: string
+        collectionFormat: multi
+        description: Dataset IDs to filter.
+      - in: query
         name: name
         type: string
         required: false
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -403,6 +403,7 @@ def list_datasets(tenant_id: str, args: dict):
     :return: (success, result) or (success, error_message)
     """
     kb_id = args.get("id")
+    kb_ids = args.get("ids")
     name = args.get("name")
     page = int(args.get("page", 1))
     page_size = int(args.get("page_size", 30))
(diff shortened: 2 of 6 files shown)


wangq8 (commented):
Thanks for this PR! The feature is useful and the implementation is mostly clean. A few comments:

## What works well
- ✅ Permission checks via `get_accessible_ids` — failing loudly when user lacks access is the right call
- ✅ Mutual exclusion between `id` and `ids` prevents ambiguous queries
- ✅ UUID validation with duplicate detection in the validator
- ✅ Unit test covers the happy path
- ✅ Python SDK updated consistently

## Issues

### 1. Error message formatting (minor)
`dataset_api_service.py` line ~423:
```python
return False, f"Should not provide both 'id':{kb_id} and 'ids'{kb_ids}"
```
- Missing a colon and space before `{kb_ids}`
- `kb_ids` is a list, so the output will render as `...['uuid1', 'uuid2']` which isn't very readable in an error message
- Suggested:
```python
f"Should not provide both 'id': {kb_id} and 'ids': {', '.join(kb_ids)}"
```

### 2. Variable shadowing (style)
`dataset_api_service.py`:
```python
denied_ids = [kb_id for kb_id in kb_ids if kb_id not in accessible_ids]
```
The loop variable `kb_id` shadows the outer `kb_id` (the single-ID filter variable). Use a different name, e.g. `_id` or `did`:
```python
denied_ids = [did for did in kb_ids if did not in accessible_ids]
```

### 3. Triple-quoted string for single line (nit)
`dataset_api_service.py`:
```python
return False, f"""User '{tenant_id}' lacks permission for datasets: '{", ".join(denied_ids)}'"
```
This works but is unconventional for a single-line string.

### 4. Missing test coverage
Consider adding tests for:
- Mutual exclusion: passing both `id` and `ids` should return an error
- Permission denied: user requests IDs they don't have access to
- Empty `ids` array behavior

### 5. Global side effect of `_is_list_annotation` in `validate_and_parse_request_args`
The list-detection logic is added to the generic `validate_and_parse_request_args` function, not just `ListDatasetReq`. This is actually a good improvement (any list-annotated query param now correctly gets all values), but it means this PR's scope now affects ALL request validators with list-typed query params. If any existing validator relied on `to_dict(flat=True)` only returning the first value, behavior would change. The change itself is correct — just flagging for awareness since this goes beyond the stated PR scope.

PR #17614, opened 2026-07-31
refa: simplify RAPTOR tree clustering and configuration

buua436:
## Summary

This PR simplifies RAPTOR tree construction and replaces the previous configurable GMM/Psi clustering paths with a deterministic adjacent-similarity clustering strategy.

## Changes

- Replace the previous GMM/Psi/AHC branching logic with adjacent cosine-similarity watershed clustering.
- Add `clustering_threshold` and `clustering_ratio` configuration fields.
- Expose clustering threshold and ratio in the tree template UI.
- Restrict RAPTOR `max_token` to the supported range of 512–2048.
- Remove obsolete RAPTOR configuration fields and helpers:
  - `threshold`
  - `clustering_method`
  - `tree_builder`
  - `ext`
  - Psi-specific options
- Remove obsolete Psi clustering tests and update RAPTOR utility tests.
- Use explicit defaults in the compiler and task execution paths.
- Define a structured tree summary prompt:
  - first line is the summary title;
  - following lines contain the summary;
  - no extra labels or Markdown headings.
- Restore leaf-cluster materialization so one leaf node aggregates all source chunk IDs from the cluster instead of creating one leaf node per chunk.


wangq8 (commented):
Thanks for this simplification — the reduction from ~1200 lines to ~200 is a meaningful cleanup. I found a few issues that need attention:

### 🔴 Must-fix: `task_handler.py` incomplete update

`rag/svr/task_executor_refactor/task_handler.py:_run_raptor` was only partially updated. The fallback config dict still contains the removed fields (`threshold: 0.1`, `clustering_method: "gmm"`, `tree_builder: "raptor"`) and is **missing the new fields** (`clustering_threshold`, `clustering_ratio`). These removed fields will fail Pydantic validation since `RaptorConfig` no longer accepts them.

Please align this dict with the changes in `api_utils.py`, `do_handle_task`, and the compiler — add `clustering_threshold: 0.3`, `clustering_ratio: 0.5`, and remove `threshold`, `clustering_method`, and `tree_builder`.

### 🟡 Naming / docstring mismatch: `clustering_threshold`

The parameter is named like an absolute cosine-similarity threshold, and the constructor docstring says "Adjacent chunks with cosine similarity **below this value** become cluster boundaries." But in `_get_clusters_ahc`, the value is actually used as a **percentile** via `int(round(self._clustering_threshold * 100))`. A user setting `0.3` expects a similarity cutoff of 0.3, not the 30th percentile.

Consider either:
- Renaming to `clustering_percentile` and updating the docstring, or
- Using it as an absolute threshold (clamp to a percentile only in the ratio-cap fallback).

### 🟡 Method naming: `_get_clusters_ahc`

The method still carries `ahc` in its name but now implements 1D watershed segmentation over adjacent similarities, not agglomerative hierarchical clustering. Consider renaming to `_get_clusters_watershed` or `_get_clusters_adjacent`.

### 🟢 Minor: unused `random_state` parameter

`clustering()` still accepts `random_state` but no code path uses it anymore (GMM was the only consumer). Not a bug, but worth cleaning up in a follow-up.

### ✅ Good

- The watershed + ratio-cap approach is clean and O(N)
- Leaf-cluster materialization consolidation is correct
- Prompt restructuring to enforce title-first-line format is a nice improvement
- `max_token` clamping to 512–2048 in the constructor adds a safety net
- Test cleanup is thorough — no stale Psi references left

PR #17616, opened 2026-07-31
 fix(dataset): echo back selected datasets not in the first page

dcc123456:
### Summary

 fix(dataset): echo back selected datasets not in the first page

  The knowledge base form field uses infinite scroll pagination. When a
  selected dataset ID was not present on the first page (or in the cache),
  the MultiSelect could not resolve a matching option and the selected
  value displayed as empty.

  Resolve this by fetching the selected datasets by their IDs via the new
  `ids` query parameter on the `/api/v1/datasets` endpoint (backed by
  infiniflow/ragflow#17606) and merging the results into the local cache
  so their names can be echoed back in the form field.

  - Add `listDatasetByIds` service that calls `GET /api/v1/datasets` with the `ids` query param (comma-joined) and a matching `page_size`.
  - Add `KnowledgeListKeys.byIds` query key factory and `useFetchDatasetsByIds` hook on top of `useQuery` with a stable sorted-IDs key.
  - In `useDisableDifferenceEmbeddingDataset`, compute the `missingIds` (selected but neither in the loaded page nor cached), fetch them, and fold the results into the cache and merged dataset list so the options render correctly.

### DIFF
--- a/web/src/services/knowledge-service.ts
+++ b/web/src/services/knowledge-service.ts
@@ -268,6 +268,14 @@ export function deleteKnowledgeGraph(knowledgeId: string) {
 export const listDataset = (params?: IFetchKnowledgeListRequestParams) =>
   request.get(api.kbList, { params });
 
+// Fetch datasets by a set of IDs via the `ids` query param (comma-joined).
+// Used to echo back already-selected datasets whose names are not present
+// in the first page of the paginated list.
+export const listDatasetByIds = (ids: string[]) =>
+  request.get(api.kbList, {
+    params: { ids: ids.join(','), page_size: ids.length },
+  });
+
 export const datasetFilter = () => request.get(api.datasetFilter);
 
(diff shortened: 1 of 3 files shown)


web/src/services/knowledge-service.ts
```diff
@@ -268,6 +268,14 @@ export function deleteKnowledgeGraph(knowledgeId: string) {
 export const listDataset = (params?: IFetchKnowledgeListRequestParams) =>
   request.get(api.kbList, { params });
 
+// Fetch datasets by a set of IDs via the `ids` query param (comma-joined).
+// Used to echo back already-selected datasets whose names are not present
+// in the first page of the paginated list.
+export const listDatasetByIds = (ids: string[]) =>
+  request.get(api.kbList, {
+    params: { ids: ids.join(','), page_size: ids.length },
```
wangq8, line 276: should be: `ids=id-1&ids=id-2`

PR #17617, opened 2026-07-31
fix(go-agent): align logs auth with Python and harden event data

Hz-186:
## Summary

- Align Go agent logs access with the Python behavior by checking tenant access to the target agent instead of requiring the token to be bound
to that agent.
- Marshal canvas lifecycle event payloads safely so non-serializable inputs do not break Thinking expansion in the UI.

  ## Testing

 - `CGO_ENABLED=0 go test -count=1 -v -run "Handler|Attachment|Logs|Chatbot|Agentbot|BuildWorkflow|NodeLifecycle|VarRef|Resolve|Extract"./internal/handler/ ./internal/agent/canvas/...` PASS

### DIFF
--- a/internal/agent/canvas/scheduler.go
+++ b/internal/agent/canvas/scheduler.go
@@ -20,7 +20,6 @@ package canvas
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


yuzhichang:
LGTM

PR #17624, opened 2026-07-31
Refine knownledge compilation REST APIs

wangq8:


wangq8 (commented):
## PR Review: Refine knowledge compilation REST APIs

**Author:** @wangq8 | **Files:** 8 | **+76 / −93**

---

### Summary

This PR standardises REST API paths by replacing underscores with hyphens, flattening `/nav` → `/navigation`, switching probe endpoints from GET to HEAD, and consolidating commit routes under `/workspaces/` (plural). The frontend and docs are kept in sync.

---

### 🔴 Critical

**1. Go router not updated — route mismatch**

`internal/router/router.go` (lines 447–470) still defines:

```go
commitFolders := v1.Group("/folders")
commitWorkspace := v1.Group("/workspace")  // singular
```

But the PR removes the Python Quart counterparts for `/folders/` and `/workspace/` (singular), keeping only `/workspaces/` (plural). If the Go router also serves these commit endpoints, requests hitting the Go layer will route to a handler that no longer exists on the Python side.

**Action:** Either update `router.go` to use `/workspaces/` or confirm the Go router is dead code / not deployed in the same request path.

---

**2. Error handling regression in `has_any_wiki` / `has_any_skill`**

The old code distinguished authentication failures from genuine empty results:

```python
if success:
    return get_result(data=result)
return get_result(data=False, message=result, code=RetCode.AUTHENTICATION_ERROR)
```

The new code collapses both into a single 404 with an empty body:

```python
response = await make_response("", 200 if success and result["has"] else 404)
return response
```

A client that gets a 404 now cannot tell whether:
- The dataset genuinely has no artifacts (functional "not found"), or
- The caller lacks permission to inspect that dataset (auth failure).

This is a semantic loss. Consider returning 401/403 for auth failures and reserving 404 for genuinely absent data.

---

### 🟠 Moderate

**3. Hard breaking change — no deprecation window**

Every route touched here is live API surface. Underscore→hyphen and `/nav`→`/navigation` are clean improvements, but any external client, SDK, or CI script hitting the old paths will get 404s immediately. Consider:
- Keeping the old routes as aliases for at least one release cycle, or
- Returning `308 Permanent Redirect` from old→new paths, or
- Documenting the break prominently in release notes.

---

**4. Inconsistent error response format for HEAD endpoints**

`has_any_wiki` and `has_any_skill` now return an empty body on success (correct for HEAD), but the `except` block still uses:

```python
return get_error_data_result(message="Internal server error")
```

This returns a JSON body (not empty) and a ~500 status. Error-handling middleware expecting consistent empty HEAD responses may misbehave. Use an empty response + status code for the error path too.

---

### 🟡 Minor

**5. `test_workspace_alias` test incomplete**

Only the docstring was updated. The test body does not verify that the new `/workspaces/` routes are actually registered in the route map. Since `/folders/` was removed entirely, the test should also confirm the old route is no longer reachable.

---

**6. Title typo**

`"Refine knownledge compilation REST APIs"` should be `"Refine knowledge compilation REST APIs"`.

---

### 🟢 What Looks Good

- Frontend `api.ts` is fully in sync with backend changes — no stale client paths.
- Doc comment updates in `structure_graph_common.py` are accurate.
- Documentation in `http_api_reference.md` correctly reflects new paths and plural naming.
- The import of `make_response` from `quart` is correct for the framework.

---

### Verdict

⚠️ **Request changes.** The Go router mismatch and error-handling regression should be addressed before merge. The breaking-changes-without-deprecation question should at least have a conscious decision recorded.

PR #17628, opened 2026-07-31
Fix: Add missing parameter key for LLM calls

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for this fix! Both changes look correct and address real issues:

**`api/apps/services/models_api_service.py`**
Good catch. When a `provider_name` is absent from `factory_rank_mapping`, `dict.get()` returns `None`, which raises a `TypeError` in Python 3 when sorting (can't compare `None` with strings). Defaulting to `-500` is sensible — unknown providers get sorted to the end, which is the right behavior.

**`rag/llm/chat_model.py`**
Adding `"content": None` to assistant messages that carry `tool_calls` is necessary for API compliance. The OpenAI Chat Completions protocol requires the `content` field to be present (even if null) when `tool_calls` is populated. Without it, some providers reject the request. The fix is applied consistently across both `_append_history` and `_append_history_batch` in both classes — looks complete.

No regressions or style issues. LGTM, approving.

Approve

PR #17632, opened 2026-07-31
Fix: Upload an image without filling in any text and press Enter; the image should not disappear.

cike8899:
### Summary

Fix: Upload an image without filling in any text and press Enter; the image should not disappear.


wangq8 (commented):
LGTM. Simple and correct fix.

The issue: pressing Enter with an empty text input and an uploaded image would call `pressEnter()` (which is a no-op for empty input) and then unconditionally clear the file list via `setFiles([])`, making the uploaded image disappear.

The fix adds two guards:
1. `sendDisabled` — mirrors the send button's own disabled logic, so Enter behaves identically to clicking the button.
2. `!value.trim()` — explicitly catches the empty-message case.

Both `sendDisabled` and `value` are already available in scope (`sendDisabled` is a prop), and both are correctly listed in the dependency array. No regressions.

PR #17640, opened 2026-07-31
fix(go-agent): fix Tavily tool array schemas

Hz-186:
## Summary

- Add string item schemas for Tavily Search domain filters.
- Add the same schema fix for Tavily Extract URLs.
- Add regression coverage for generated tool schemas.

## Testing

- `CGO_ENABLED=0 go test -count=1 ./internal/agent/tool ./internal/agent/component`

image


yuzhichang:
LGTM

PR #17642, opened 2026-07-31
Fix: Enforce the 4 MB size limit for avatar image uploads

euvre:
### Summary

The avatar upload tip states a 4 MB limit ("You can upload an image up to 4 MB." in every locale), but nothing actually enforced it: the shared `AvatarUpload` component only validated the file extension, so an oversized image (e.g. 18 MB) could be selected, cropped and uploaded successfully.

This PR adds the missing client-side check in `AvatarUpload`: selecting an image larger than 4 MB now shows an error toast (reusing the existing `knowledgeConfiguration.photoTip` text) and aborts before the crop dialog opens. Since this single component backs all avatar upload entries (memory configuration, dataset settings, user profile, chat/search app settings), one fix covers every surface.

Note: the backend only ever receives the cropped 64x64 PNG (a few KB of base64), so the original file size can only be enforced in the browser at selection time — no server-side change is needed.

### DIFF
--- a/web/src/components/avatar-upload.tsx
+++ b/web/src/components/avatar-upload.tsx
@@ -1,6 +1,7 @@
 import { combineRefs } from '@/lib/utils';
 import { transformFile2Base64 } from '@/utils/file-util';
 import { LucidePencil, LucidePlus, LucideX } from 'lucide-react';
+import message from './ui/message';
 import {
   ChangeEventHandler,
   forwardRef,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM ✅ — clean fix.

**What this does:** Adds a client-side 4 MB file size check in the shared `AvatarUpload` component, enforcing the limit already mentioned in the UI tip (`knowledgeConfiguration.photoTip`). Since the cropped result is always a 64×64 PNG (a few KB), this is the correct place to enforce the limit — no server-side change needed.

**Review notes:**
- The import `./ui/message` correctly resolves to the existing `./ui/message.ts` module (verified in `ragflow-main`).
- The i18n key `knowledgeConfiguration.photoTip` exists in all locale files and matches the 4 MB message.
- Adding `t` to the `useCallback` dependency array is correct now that `t` is used inside the callback.
- The `file!` non-null assertion is safe — the regex guard on `file?.name` above guarantees `file` is non-null by that point.
- `ev.target.value = ""` is properly called in both success and error paths, preventing re-trigger issues.
- One shared component covers all avatar upload surfaces (memory config, dataset settings, user profile, chat/search app settings).

No regressions, no new dependencies, minimal diff. Nice catch on the missing enforcement! 🚀

PR #17644, opened 2026-07-31
test: release test_documents_update_patch_and_delete for Go proxy

euvre:
### Summary

Release `test_documents_update_patch_and_delete` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises document name update, delete, and list — all of which Go fully supports on unparsed documents:

- **Update**: `UpdateDatasetDocument` handles name changes without requiring parsed state
- **Delete**: `DeleteDocuments` / `deleteDocumentFull` works on unparsed documents (doc engine delete is a no-op when no chunks exist)
- **List**: Standard document listing works regardless of parse state


wangq8 (approved):
Clean release — this test covers doc name update, delete, and list, all of which Go supports on unparsed documents without any parsing dependency. LGTM — approved.

PR #17645, opened 2026-07-31
test: release test_documents_metadata_batch_update_contract for Go proxy

euvre:
### Summary

Release `test_documents_metadata_batch_update_contract` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises metadata batch update. Go's implementation is fully compatible:

- **Validation**: All error messages match ("selector must be an object", "updates and deletes must be lists", "Each update requires key and value", "Each delete requires key", "These documents do not belong to dataset ...")
- **Auth**: 401 handling matches via `assert_auth_error` with Go-specific assertions
- **Metadata store**: Created lazily via `EnsureMetadataStore` — works on unparsed documents
- **Update/delete by IDs and conditions**: `BatchUpdateDocumentMetadatas` resolves doc IDs from DB, applies updates/deletes per document, and writes to the doc engine

### DIFF
--- a/test/testcases/restful_api/conftest.py
+++ b/test/testcases/restful_api/conftest.py
@@ -70,7 +70,6 @@ GO_ONLY_SKIPS = {
         "test_documents_update_patch_and_delete",
         "test_documents_update_name_contract",
         "test_documents_update_meta_fields_contract",
-        "test_documents_metadata_batch_update_contract",
         "test_documents_metadata_update_path",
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
Well-documented release. The metadata batch update test doesn't need parsing — Go's EnsureMetadataStore + BatchUpdateDocumentMetadatas handles unparsed docs. Error message compatibility across validation, auth, and cross-dataset scenarios is confirmed. LGTM — approved.

PR #17646, opened 2026-07-31
test: release test_documents_metadata_update_path for Go proxy

euvre:
### Summary

Release `test_documents_metadata_update_path` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises metadata update via `PATCH /datasets/:id/documents/metadatas`. Go's metadata store is created lazily via `EnsureMetadataStore` and works on unparsed documents. The response format (`data.matched_docs`, `data.updated`) matches the Python contract.


wangq8 (commented):
LGTM. Straightforward change—the test only exercises document upload + metadata update via `PATCH /datasets/:id/documents/metadatas`, which triggers no parsing pipeline. Go proxy's `EnsureMetadataStore` handles unparsed documents, and the response format matches the Python contract. The original skip reason doesn't apply here.

One small suggestion: the commit message / PR body could reference the original issue or commit that introduced this skip (for traceability), but it's not blocking.

PR #17647, opened 2026-07-31
test: release test_documents_delete_invalid_dataset_partial_duplicate_repeat_and_cross_dataset for Go proxy

euvre:
### Summary

Release `test_documents_delete_invalid_dataset_partial_duplicate_repeat_and_cross_dataset` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises document deletion with various error scenarios. Go's implementation produces matching error messages:

- Invalid dataset: code 102, "You don't own the dataset ..."
- Partial invalid IDs: code 102, "These documents do not belong to dataset ..."
- Cross-dataset: code 102, "These documents do not belong to dataset ..."
- Duplicate IDs: code 101, "Field: \<ids\> - Message: \<Duplicate ids: ..."
- Delete twice: code 102, "... or Document not found: ..."

### DIFF
--- a/test/testcases/restful_api/conftest.py
+++ b/test/testcases/restful_api/conftest.py
@@ -72,7 +72,6 @@ GO_ONLY_SKIPS = {
         "test_documents_update_meta_fields_contract",
         "test_documents_metadata_batch_update_contract",
         "test_documents_metadata_update_path",
-        "test_documents_delete_invalid_dataset_partial_duplicate_repeat_and_cross_dataset",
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
Clean release. The doc-delete contract test only requires upload (no parsing), and the PR doc clearly documents each error scenario's expected Go message. All error codes and messages match. LGTM — approved.

PR #17648, opened 2026-07-31
test: release test_chunk_concurrent_add_contract for Go proxy

euvre:
### Summary

Release `test_chunk_concurrent_add_contract` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises concurrent chunk add. Go's `AddChunk` works on unparsed documents:

- Resolves the embedding model from the tenant config (`getEmbeddingModel`)
- Calls the embedding service (TEI in CI provides `BAAI/bge-small-en-v1.5`)
- Inserts into the doc engine which creates indices lazily (`CreateTable` with `ConflictTypeIgnore` in Infinity)

The test only asserts `code == 0` and correct chunk count — no error-path assertions that could mismatch.

### DIFF
--- a/test/testcases/restful_api/conftest.py
+++ b/test/testcases/restful_api/conftest.py
@@ -66,7 +66,6 @@ GO_ONLY_SKIPS = {
         # Without a working Go ingestion pipeline, these indices never exist.
         "test_chunk_add_keyword_question_and_tag_contract",
         "test_chunk_add_repeated_and_deleted_document_contract",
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
Good to see this one released. The test only does upload + concurrent chunk add (no parsing), and Go's AddChunk path (embedding resolution via getEmbeddingModel -> TEI -> Infinity insert) supports unparsed documents. LGTM — approved.

PR #17649, opened 2026-07-31
test: release test_chunk_list_concurrent_contract for Go proxy

euvre:
### Summary

Release `test_chunk_list_concurrent_contract` from the Go-proxy skip list (`GO_ONLY_SKIPS`).

This test only requires document upload (no parsing) and exercises concurrent chunk list. Chunks are created via `_reset_chunk_batch` (POST chunk add) which works on unparsed documents. Go's `ListChunks` returns `data.total` and `data.chunks` matching the Python contract. The test only asserts `code == 0` and `total == 5` — no error-path assertions.


wangq8 (approved):
Straightforward skip-list release. The reasoning checks out — this test only does document upload + concurrent chunk list, no parsing required. Go's ListChunks already returns data.total and data.chunks matching the Python contract. LGTM — approved.

PR #17652, opened 2026-07-31
fix: XunFei Spark provider API key verification and model listing

euvre:
### Summary

The XunFei Spark provider could not verify API keys or fetch the model list in Go mode:

1. **400 "cannot unmarshal object into Go struct field CreateProviderInstanceRequest.api_key of type string"** — the frontend submits XunFei Spark credentials as an object (`{spark_api_password, spark_app_id, spark_api_secret, spark_api_key}`), but the Go handler declared `api_key` as `string`.
2. **Verification always failed** — the driver sent the whole JSON credential bundle as the Bearer token (the Spark HTTP API expects only `spark_api_password`), `CheckConnection` returned "no such method", and catalog model names (e.g. `Spark-Max`) were not mapped to the API version identifiers (e.g. `generalv3.5`) the Spark API requires.
3. **Model listing failed with HTTP 400** — same Bearer-token issue in `ListModels`.

Changes:

- `internal/handler/providers.go`, `internal/service/model_service.go`: accept `api_key` as either a string or an object in create/alter/verify provider-instance requests and normalize it to the persisted JSON-string form.
- `internal/entity/models/xunfei.go`:
  - Resolve the Bearer token from the credential bundle's `spark_api_password` for chat, streaming chat, and `ListModels`.
  - Map catalog model names to Spark API version identifiers, matching Python's `SparkChat.model2version`.
  - Implement `CheckConnection` via a minimal Spark-Lite chat request.
  - Surface upstream HTTP status/body on non-stream chat errors instead of a generic parse error.
- `internal/entity/models/xunfei_test.go`: unit tests for the model-name mapping, bearer-token resolution, and the new `CheckConnection` behavior.

### DIFF
--- a/internal/entity/models/xunfei.go
+++ b/internal/entity/models/xunfei.go
@@ -24,7 +24,47 @@ import (
 	"io"
 	"net/http"
 	"ragflow/internal/common"
+	"strings"
 )
+
+// sparkModelVersions maps the catalog model names to the version identifiers
+// the XunFei Spark HTTP API expects in the request body's "model" field.
+// Mirrors SparkChat.model2version in rag/llm/chat_model.py.
+var sparkModelVersions = map[string]string{
+	"Spark-Max":       "generalv3.5",
+	"Spark-Max-32K":   "max-32k",
+	"Spark-Lite":      "lite",
+	"Spark-Pro":       "generalv3",
+	"Spark-Pro-128K":  "pro-128k",
+	"Spark-4.0-Ultra": "4.0Ultra",
+}
+
+func resolveSparkModel(modelName string) string {
+	if version, ok := sparkModelVersions[modelName]; ok {
+		return version
+	}
+	return modelName
+}
+
+// resolveBearerToken extracts the credential used as the Bearer token. The
+// instance stores the XunFei credential bundle (API password, APPID, API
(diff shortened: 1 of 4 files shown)


wangq8 (approved):
Solid fix for XunFei Spark in Go mode. Well-structured with tests.

**Good:**
- normalizeAPIKey / json.RawMessage approach in the handler is clean and correctly handles both string and object api_key without breaking existing providers
- resolveBearerToken correctly extracts spark_api_password from the credential bundle for Bearer auth
- resolveSparkModel mapping matches Python's SparkChat.model2version — good parity
- CheckConnection via minimal Spark-Lite request is a reasonable verification strategy
- Better error surfacing (HTTP status + body) on non-200 responses instead of opaque parse errors
- Tests cover model mapping, bearer token resolution, and CheckConnection edge cases

**One minor note (non-blocking):**
- resolveBearerToken uses strings.HasPrefix(key, "{") as a heuristic to detect JSON bundles. While extremely unlikely in practice, a plain API key starting with "{" would trigger a failed unmarshal (falls back gracefully to raw key, so no breakage — just something to be aware of).

LGTM — approved.

Lynn-Inf (approved):
LGTM

PR #17653, opened 2026-07-31
fix(go): accept empty/unknown model_type when adding provider models

euvre:
### Summary

In Go mode, adding a model from a provider's remote model list (e.g. Gemini) failed with HTTP 400 ("bad request").

Root causes:

- The Gemini driver's `ListModels` used the display name ("Gemini 2.5 Flash") instead of the API model ID (`gemini-2.5-flash`). Remote models therefore never matched the static catalog, so they carried no `model_types`/`max_tokens`, and the frontend sent `model_type: []`. Display names are also unusable in Gemini API calls.
- The Go `AddModel` handler/service were stricter than the Python reference: Python's `calculate_model_type` silently ignores empty and unknown model types, while Go rejected them with 400.

Changes:

- `internal/entity/models/google.go`: list models by API model ID (strip the `models/` prefix, display name as fallback) and report `OwnedBy` as "Gemini" so listed models match the static catalog and the `model@provider` composite-name convention used elsewhere (consistent with Cohere/DeepInfra using `Name()`).
- `internal/handler/providers.go`: drop the empty `model_type` 400 check in `AddModel`.
- `internal/service/model_service.go`: ignore empty/unknown model types like Python's `calculate_model_type` instead of rejecting them, matching the existing tolerant behavior of `addModelToInstance` used by `CreateProviderInstance`.

Validation: `bash build.sh --test ./internal/entity/models/ ./internal/service/` passes.


wangq8 (approved):
Clean fix that aligns Go behavior with Python's more tolerant model_type handling.

**Good:**
- Gemini driver now lists models by API model ID (e.g. gemini-2.5-flash) instead of display name — this actually makes remote models usable in API calls
- OwnedBy change to "Gemini" matches the composite name convention used by Cohere/DeepInfra
- Removing the strict model_type validation and silently ignoring unknown types matches Python's calculate_model_type behavior

LGTM — approved.

Lynn-Inf (commented):
I'd suggest filtering out unknown or unsupported `model_type` values when generating the list, rather than allowing them and handling them later.

Lynn-Inf (approved):
LGTM

PR #17657, opened 2026-07-31
Fix: handle html tags in empty_response

Lynn-Inf:
### Summary

As title.


wangq8 (approved):
Thanks for this fix! It addresses a real UX issue where HTML-like tags in empty_response cause DOMPurify to strip content (leaving the UI stuck) and TTS engines to hang on unclosed tags.

**Positive:**
- HTML-escaping the frontend answer while passing raw text to TTS is the correct two-pronged approach
- The tag-stripping regex in both clean_tts_text functions is consistent and appropriate

- clean_tts_text is now duplicated between canvas.py and dialog_service.py. Consider extracting it to a shared utility (e.g. api/utils/) in a follow-up PR to avoid maintenance drift between the two copies.

LGTM — approved.

PR #17658, opened 2026-07-31
Fix: add doc_type_kwd back

wangq8:
Default to text.
image


wangq8 (commented):
## Review: PR #17658 — Fix: add doc_type_kwd back

**Summary:** This PR restores the `doc_type_kwd` field to the chunk list API, defaulting to `"text"` when not set. It also sets `doc_type_kwd: "text"` on newly created chunks via `add_chunk`, and adds unit test coverage. A focused fix with tests — looks good overall.

### ✅ What works well

- **Correct backward-compatible fallback** in `list_chunks`: `chunk.get("doc_type_kwd") if isinstance(..., str) else "text"` — handles chunks in ES that predate this field gracefully.
- **`add_chunk` sets the default** so new chunks always have `doc_type_kwd`.
- **Tests cover both paths**: default `"text"` for text chunks and `"image"` for image chunks, including the `add_chunk` response.

### 💡 Suggestions (non-blocking)

1. **`get_chunk` inconsistency**: `list_chunks` now defaults missing `doc_type_kwd` to `"text"`, but `get_chunk` returns raw ES data (via `_strip_chunk_runtime_fields`) and will simply omit the field for old chunks. Consider applying the same fallback in `get_chunk` for a consistent API contract, or at least documenting that callers should expect `doc_type_kwd` to be absent on legacy chunks from that endpoint.

2. **Duplicated fallback logic in `list_chunks`**: The two nearly identical dict-building blocks (regular path and ES path) now each have the same `doc_type_kwd` fallback expression copy-pasted. It would be cleaner to extract a shared helper or consolidate these paths — the duplication is a maintenance risk for future field additions.

3. **Model default vs API default**: `Chunk.doc_type_kwd` defaults to `""` (empty string), but the API response defaults to `"text"`. These should be aligned — using `"text"` as the model default would be more consistent. (Not harmful in practice since the model is only used internally, but worth tidying up.)

4. **No-op key mapping**: The `key_mapping` entry `"doc_type_kwd": "doc_type_kwd"` is a no-op rename. There is precedent (`"create_time": "create_time"`), but it is worth noting that a pass-through mapping is brittle if the key is ever renamed upstream. Consider including `doc_type_kwd` into `renamed_chunk` explicitly rather than relying on the mapping loop.

### 🔍 Regression / security / style

- **No regressions** — only additive changes, no existing logic altered.
- **No security concerns** — the field is a string type with controlled values (`"text"` / `"image"`).
- **Style** — consistent with surrounding code patterns.

### ✅ Verdict

LGTM. The fix is correct, the tests cover it, and the suggestions above are minor polish. Approved from my side.

PR #17665, opened 2026-07-31
fix(dataset): preserve disabled RAPTOR and GraphRAG settings

### LINKED ISSUE #17654: [Bug]: Dataset Configuration page force-resets RAPTOR and GraphRAG to enabled on every load/save
### Description
Opening a dataset's Configuration page (dataset-setting) always initializes the
form with `parser_config.raptor.use_raptor = true` and
`parser_config.graphrag.use_graphrag = true`, regardless of the dataset's actual
saved values. If the user then clicks Save (even without touching anything else
on the page, e.g. just changing the LLM model), these values get persisted back
to the backend — silently re-enabling RAPTOR/GraphRAG on datasets that had them
explicitly disabled.

### Root cause
`web/src/pages/dataset/dataset-setting/hooks.ts`, `useFetchKnowledgeConfigurationOnMount`:

    const parser_config = {
      ...form.formState?.defaultValues?.parser_config,
      ...knowledgeDetails.parser_config,
      raptor: {
        ...form.formState?.defaultValues?.parser_config?.raptor,
        ...knowledgeDetails.parser_config?.raptor,
        use_raptor: true,      // <-- unconditionally overwrites the real value
      },
      graphrag: {
        ...form.formState?.defaultValues?.parser_config?.graphrag,
        ...knowledgeDetails.parser_config?.graphrag,
        use_graphrag: true,    // <-- same here
      },
    };

The spread correctly pulls `knowledgeDetails.parser_config.raptor/graphrag`
(the real saved config) first, but then hardcodes `use_raptor`/`use_graphrag`
to `true` afterward, clobbering it.

### Steps to reproduce
1. Create a dataset via the API with `parser_config.raptor.use_raptor: false`
   and `parser_config.graphrag.use_graphrag: false`.
2. Open that dataset's Configuration page in the UI.
3. Change any unrelated field (e.g. the LLM model) and click Save.
4. GET the dataset via API -- `use_raptor` and `use_graphrag` are now `true`.

### Expected
The form should load the dataset's actual saved `use_raptor`/`use_graphrag`
values, not force them to `true`.

### Version
v0.26.4 (self-hosted, Helm)

lntutor:
### Summary

Preserve explicitly saved `false` values for RAPTOR and GraphRAG when hydrating the dataset Configuration form. The existing form defaults still apply when either enable flag is absent.

Adds hook-level regression coverage for both explicit disablement and default fallback behavior.

Fixes #17654

### Verification

- `npx jest --runInBand --runTestsByPath src/pages/dataset/dataset-setting/hooks.test.tsx src/hooks/tests/parser-config-utils.test.ts`
- `npx oxfmt --check src/pages/dataset/dataset-setting/hooks.ts src/pages/dataset/dataset-setting/hooks.test.tsx`
- `npx oxlint src/pages/dataset/dataset-setting/hooks.ts src/pages/dataset/dataset-setting/hooks.test.tsx`
- `npm run build`
- `git diff HEAD^ --check`

Repository-wide `npm run type-check`, `npm run lint`, and `npm run format:check` remain red on pre-existing files outside this change; the touched files pass their focused checks.

### DIFF
--- a/web/src/pages/dataset/dataset-setting/hooks.test.tsx
+++ b/web/src/pages/dataset/dataset-setting/hooks.test.tsx
@@ -1,0 +1,78 @@
+import { useFetchKnowledgeBaseConfiguration } from '@/hooks/use-knowledge-request';
+import { renderHook } from '@testing-library/react';
+import { useFetchKnowledgeConfigurationOnMount } from './hooks';
+
+jest.mock('@/hooks/use-knowledge-request', () => ({
+  useFetchKnowledgeBaseConfiguration: jest.fn(),
+}));
+jest.mock('@/hooks/common-hooks', () => ({
+  useSetModalState: jest.fn(),
+}));
+jest.mock('@/hooks/use-user-setting-request', () => ({
+  useSelectParserList: jest.fn(() => []),
+}));
+jest.mock('@/services/knowledge-service', () => ({
+  checkEmbedding: jest.fn(),
+}));
+jest.mock('react-router', () => ({
+  useParams: jest.fn(() => ({})),
+  useSearchParams: jest.fn(() => [new URLSearchParams()]),
+}));
+
+const mockUseFetchKnowledgeBaseConfiguration = jest.mocked(
+  useFetchKnowledgeBaseConfiguration,
+);
+
+function renderConfigurationHook(parserConfig: Record<string, unknown>) {
+  mockUseFetchKnowledgeBaseConfiguration.mockReturnValue({
+    data: {
+      name: 'Dataset',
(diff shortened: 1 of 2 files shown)


lntutor:
Current head 4ee051f42eb7bb2336a0963fa785c0ebbc06e19a is mergeable, with no unresolved review threads and no contributor-side follow-up remaining from my side. The visible current-head automation is the skipped sep-tests workflow plus a successful CodeRabbit pass, so this looks ready for maintainer review if that check state is expected for this path.

dcc123456 (approved):

PR #17666, opened 2026-07-31
fix(qqbot): hold a reference to the dispatch task

noron12234:
## What

The QQ bot gateway spawns one task per inbound message and throws it away:

```python
incoming = self._normalize_incoming_event(event_type, data)
if incoming is None:
    return
asyncio.create_task(self._dispatch(incoming))
```

[The event loop only keeps a weak reference to a running task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done.

So a dispatch can be collected before `_dispatch` runs, and the user's message is silently never processed. No exception, no log line — the bot simply doesn't answer that one.

## Why this looks like an oversight rather than a decision

Three things point the same way:

**qqbot is the only channel that does it.** Every other channel awaits the dispatch directly:

| channel | line | how |
|---|---|---|
| discord | 79 | `await self._dispatch(incoming)` |
| dingtalk | 319 | `await self._dispatch(incoming)` |
| wecom | 236 | `await self._dispatch(incoming)` |
| line | 179 | `await self._dispatch(incoming)` |
| telegram | 106 | `await self._dispatch(incoming)` |
| whatsapp | 340 | `await self._dispatch(incoming)` |
| feishu | 149 | `run_coroutine_threadsafe(...)` — keeps the future |
| **qqbot** | **257** | **`create_task(...)` — discarded** |

**The same file already gets this right for its other task.** `_heartbeat_task` is stored on the instance (line 71, cancelled in `stop()`), and the dispatch task next to it is not.

**The repo already has the idiom.** `api/apps/restful_apis/tenant_api.py:38` keeps a module-level `Set[asyncio.Task]`, adds each task, and discards it in a done callback.

## The change

```python
self._dispatch_tasks: Set[asyncio.Task] = set()
...
task = asyncio.create_task(self._dispatch(incoming))
self._dispatch_tasks.add(task)
task.add_done_callback(self._dispatch_tasks.discard)
```

`stop()` also cancels any still in flight, next to the existing heartbeat cancel, so the set cannot outlive the channel:

```python
for task in tuple(self._dispatch_tasks):
    if not task.done():
        task.cancel()
self._dispatch_tasks.clear()
```

I kept the fire-and-forget behaviour rather than switching to `await` like the other channels: the QQ gateway's event loop should not block on handler work, and changing that is a behaviour decision rather than a bug fix. `_dispatch` in `core/base.py` already swallows handler exceptions, so no extra error handling is needed in the callback.

`discard` rather than `remove` so a double callback cannot raise. The set stays bounded by the number of in-flight messages.

## Verification

```console
$ ruff check api/channels/qqbot/channel.py
All checks passed!
$ ruff format --check api/channels/qqbot/channel.py
1 file already formatted
$ python -m py_compile api/channels/qqbot/channel.py
```

No test: the failure is a garbage-collection race, so a test would have to force a GC at a chosen moment and assert a task did *not* vanish — flaky by construction. Happy to add one if you have a shape in mind.

Found with an AST scan for `create_task` / `ensure_future` results discarded as bare expression statements (`TaskGroup.create_task` excluded — it holds strong references). This was the only hit under `api/` and `rag/`.


buua436 (commented):
Please cancel and await the dispatch tasks on `self._ws_loop` before it is closed. `stop()` currently joins the WebSocket thread and then calls `Task.cancel()` from another loop, which can raise `RuntimeError: Event loop is closed` and leave pending tasks. Move the cleanup into the WebSocket loop’s shutdown path and drain the tasks with `asyncio.gather(..., return_exceptions=True)`.

noron12234:
Thanks @buua436 — you were right, and the failure reproduces exactly as you described.

I ran it against the real class: seed a dispatch task on the websocket loop, close that loop, then call `Task.cancel()` from the caller's loop the way `stop()` did. The cancel raises `RuntimeError: Event loop is closed` and the dispatch is still pending afterwards.

Moved the cleanup into `_drain_dispatch_tasks()`, awaited from `_run()`'s `finally`, so it runs on the websocket loop before `_run_ws_thread` closes it. Tasks are cancelled and then drained with `asyncio.gather(..., return_exceptions=True)`. The block in `stop()` is gone; `stop()` still sets `_stop_event` and joins the thread, which is what makes `_run` unwind in the first place.

With the drain in place the task reports `cancelled()` and `done()` and the tracking set is empty. `ruff check` on this file is 23 findings before and after, `ruff format` clean.

PR #17667, opened 2026-07-31
fix(common): round instead of truncate in normalize_overlapped_percent (closes #17418)

### LINKED ISSUE #17418: overlapped_percent fraction inputs are truncated instead of rounded (e.g. 0.29 -> 28, expected 29)
## Title

`overlapped_percent` fraction inputs are truncated instead of rounded (e.g. `0.29` → `28`, expected `29`)

## Summary

When `overlapped_percent` is supplied as a `[0,1)` fraction (the scale accepted by the
UI / DSL), certain values are normalized to a percentage that is **1 less than a user
would expect**, because the normalization truncates toward zero with `int()` after a
floating-point multiplication. Example: `0.29` → `28` (expected `29`).

This affects both the Python and Go implementations identically, so it is a pre-existing,
shared behavior rather than a Go regression introduced by the ingestion migration.

## Reproduction

`NormalizeOverlappedPercent` (Go) and `normalize_overlapped_percent` (Python) agree:

| Input | Normalized (current) | Intuitive expectation |
|-------|----------------------|-----------------------|
| `0.1` | 10 | 10 |
| `0.29` | **28** | 29 |
| `0.3` | 30 | 30 |
| `0.57` | **56** | 57 |
| `0.93` | 90 (trunc 92 → clamped) | 93 → clamped 90 |
| `0.95` | 90 | 90 |
| `33.3` | 33 | 33 |

## Root cause

In the conversion `if 0 < v < 1: v *= 100` followed by `v = int(v)`:

- `0.29` is not exactly representable in IEEE-754 float64; it is stored as
  `0.28999999999999998…`.
- `0.28999999999999998 * 100 = 28.999999999999996`, and `int()` truncates toward zero → `28`.
- Asymmetry: `0.3 * 100` happens to round to exactly `30.0`, so `0.3 → 30` looks correct;
  only fractions whose `×100` result lands just below an integer are affected.

## Impact

- **Low.** Only fraction inputs whose `×100` result lands just below an integer within
  `[0,90]` are off by 1 percentage point. Values that exceed 90 are still clamped to 90 by
  the existing bounds, so no out-of-range output is possible.
- The behavior matches Python exactly, so it preserves the Python-parity goal of the Go
  ingestion migration (PR #17396).

## Options

1. **Keep current behavior (Python parity).** No change; `0.29 → 28` remains, documented as
   expected Python-faithful behavior.
2. **Round instead of truncate in Go only.** Replace `int(v)` with `math.Round(v)` in
   `clampOverlappedPercent` / `Validate`. `0.29 → 29`. Cost: Go and Python then diverge for
   these inputs.
3. **Round in both Go and Python.** Also update `common/float_utils.py`; keeps engines
   consistent but changes the Python reference implementation and widens scope.

## References

- Go: `internal/ingestion/component/schema/chunker.go`
  - `NormalizeOverlappedPercent` (~L264)
  - `clampOverlappedPercent`: `value = float64(int(value))` (~L296)
  - `TokenChunkerParam.Validate`: `p.OverlappedPercent = float64(int(v * 100))` (~L310-312)
- Python: `common/float_utils.py:50-58` (`value = int(value)` at L57)
- Tests: `internal/ingestion/component/chunker/token_test.go` —
  `TestNormalizeOverlappedPercent` (e.g. `{"int truncation 33.3 -> 33", 33.3, 33}`)

## Environment

- Branch under review: `fix/batch-3-4` (PR #17396)
- Verified via a `NormalizeOverlappedPercent` boundary run (23 inputs, all integer-valued)
  and a Python `normalize_overlapped_percent` parity check (identical finite results).

Souravrajvi0:
## Summary

Fixes #17418 — fraction inputs to `normalize_overlapped_percent` (e.g. `0.29`) were truncated to `28` instead of rounded to `29` because `int()` truncates after float multiplication.

The Go path was already fixed in #17452; this completes the Python side so both implementations agree.

## Changes

- `common/float_utils.py`: clamp to `[0, 90]` before `round()` instead of `int()` truncation
- `test/unit_test/common/test_float_utils.py`: add parametrized tests covering the issue table (`0.29 → 29`, `0.57 → 57`, `0.93 → 90`, etc.)

## Test plan

- [x] Manual sanity check: `0.29 → 29`, `0.57 → 57`, `0.93 → 90`
- [x] `uv run pytest test/unit_test/common/test_float_utils.py -q`


xugangqiang:
@yuzhichang  The fix itself is good - one question, do we need to be back-compatible for 0.3? Are all numbers should be integer in go side?

Souravrajvi0:
> @yuzhichang The fix itself is good - one question, do we need to be back-compatible for 0.3? Are all numbers should be integer in go side?

0.3 is unchanged (30),, Fraction input [0,1) remains supported, Go already normalizes to integer percentages via [math.Round]
. This PR aligns Python with that. The intentional behavior change is rounding edge cases like 0.29 → 29, not 0.3

yuzhichang (changes requested):
@xugangqiang @Souravrajvi0 It's bad for get_float(v) to guess if v is percent or ratio. I prefer refactor python and go get_float(v) to ensure v is always percent or ratio.

Souravrajvi0:
> @xugangqiang @Souravrajvi0 It's bad for get_float(v) to guess if v is percent or ratio. I prefer refactor python and go get_float(v) to ensure v is always percent or ratio.

Agreed  the 0 < v < 1 heuristic in normalize_overlapped_percent / NormalizeOverlappedPercent is ambiguous. The canonical DSL format today is ratio [0, 1) (token_chunker validation + agent pipeline save path). Proposal: normalize to always expect ratio, remove percent auto-detection in Python/Go, and audit callers (dataset settings, task_executor, Go Update/Validate) for any percent-style values. Happy to do that in this PR or a follow-up  which do you prefer?

yuzhichang:
@Souravrajvi0 I prefer do that in this PR.

Souravrajvi0:
> @Souravrajvi0 I prefer do that in this PR.

Hi It's been done can you please check

xugangqiang:
@Souravrajvi0  I can see there's another PR - https://github.com/infiniflow/ragflow/pull/18786 . This resolved related 3 fields --- from float to integer. And we think it's better. Apologies for not letting you know in the earliest time.

Issue #17510, opened 2026-07-29
[Task]: Enhance PaddleOCR check_installation: verify access_token connectivity


Lynn-Inf:
### Parent Issue

#14736 

### Detail of Subtask

Enhance PaddleOCR's check_installation to verify that the provided access_token can successfully reach the configured base_url (or the available check endpoint). This will help catch configuration issues early.

Lynn-Inf:
@Rander7 would you be available to help with this task? I think your experience with the OCR provider would be really valuable here.

PR #17669, opened 2026-07-31
feat(deepdoc): verify PaddleOCR service connectivity in check_installation

HUAN2022A:
## Summary

`PaddleOCRParser.check_installation()` previously only checked that an access token string was present, so a mistyped `base_url` or an invalid/expired token passed the check and surfaced only as a cryptic runtime error during parsing. This PR makes the check actually verify connectivity to the configured PaddleOCR service with a model-only POST to `/api/v2/ocr/jobs` (10s timeout): HTTP 401/403 means the token is rejected, any other HTTP response means the service is reachable, and network exceptions mean it is unreachable.

Related task: #17510 (parent #14736), requested by maintainer Lynn-Inf.

## Validation

- 7 new focused unit tests pass; 41 tests pass together with `test_opendataloader_parser.py`
- `py_compile` and `ruff` clean on the new test file; the one new broad exception carries a targeted `noqa`

### DIFF
--- a/deepdoc/parser/paddleocr_parser.py
+++ b/deepdoc/parser/paddleocr_parser.py
@@ -247,10 +247,37 @@ class PaddleOCRParser(RAGFlowPdfParser):
 
     # Public methods
     def check_installation(self) -> tuple[bool, str]:
-        """Check if the parser is properly installed and configured."""
+        """Check that the parser is configured and the OCR service is reachable.
+
+        The access token is validated with a minimal POST to the jobs endpoint
+        (model only, no file). The official service authenticates the token
+        before accepting a job, so an invalid token yields HTTP 401/403 without
+        creating a job or consuming page quota, while a valid token yields a
+        different error (e.g. 422 for a missing file).
+        """
         if not self.access_token:
             return False, "[PaddleOCR] Access token not configured"
 
+        if not self.base_url:
+            return False, "[PaddleOCR] Base URL not configured"
+
+        headers: dict[str, str] = {"Client-Platform": "ragflow"}
+        headers["Authorization"] = f"Bearer {self.access_token}"
+        jobs_url = f"{self.base_url.rstrip('/')}/api/v2/ocr/jobs"
+
+        try:
+            resp = requests.post(jobs_url, data={"model": self.algorithm}, headers=headers, timeout=10)
+        except Exception as exc:  # noqa: BLE001 - connectivity failures are expected here
+            reason = f"[PaddleOCR] service unreachable at {self.base_url}: {exc}"
+            self.logger.warning(reason)
+            return False, reason
+
+        if resp.status_code in (401, 403):
+            reason = f"[PaddleOCR] access token rejected by {self.base_url} (HTTP {resp.status_code})"
+            self.logger.warning(reason)
+            return False, reason
+
+        self.logger.info(f"[PaddleOCR] service reachable at {self.base_url} (HTTP {resp.status_code})")
         return True, ""
 
     def parse_pdf(
--- a/deepdoc/parser/paddleocr_parser.py
+++ b/deepdoc/parser/paddleocr_parser.py
@@ -270,7 +270,17 @@ class PaddleOCRParser(RAGFlowPdfParser):
 
     # Public methods
     def check_installation(self) -> tuple[bool, str]:
-        """Check if the parser is properly installed and configured."""
+        """Check that the parser is configured and the OCR service is reachable.
+
+        The hosted service is validated with a minimal POST to the jobs
+        endpoint (model only, no file). It authenticates the token before
+        accepting a job, so an invalid token yields HTTP 401/403 without
+        creating a job or consuming page quota, while a valid token yields a
+        different error (e.g. 422 for a missing file). A self-hosted PaddleX
+        deployment only serves the synchronous layout-parsing endpoint, so
+        reachability there is probed with a plain GET; any HTTP response
+        counts as reachable.
+        """
         if not self.base_url:
             return False, "[PaddleOCR] Base URL not configured"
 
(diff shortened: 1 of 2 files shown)


JinHai-CN:
@HUAN2022A Will you continue to finish this PR?

HUAN2022A:
Yes, I'll continue working on it. Sorry for the delay. I'll sync the branch with the latest main, rerun the tests, address any issues, and mark the PR ready for review soon.

HUAN2022A:
@JinHai-CN Sorry for the slow reply — yes, I'm continuing with this PR.

Current state:

- `check_installation()` now performs a real connectivity check (model-only POST to `/api/v2/ocr/jobs`, 10s timeout): 401/403 → token rejected, any other HTTP response → service reachable, network exception → unreachable.
- 7 focused unit tests added and passing (41 total when run together with the existing opendataloader tests); `ruff` and `py_compile` clean.

Next step: I'll take it out of draft shortly so it can get a proper review pass (the original request came from @Lynn-Inf in #17510). If you'd like anything adjusted in the meantime — e.g. the timeout value or how failures are classified — happy to work that in.

HUAN2022A:
Rebased onto the latest main (1186 commits) and reworked the check to fit the self-hosted support from #19011:

- Config checks now match upstream: a missing token only fails for the hosted service (`not self.local`), since self-hosted deployments usually run unauthenticated.
- The connectivity probe follows each deployment's wire protocol: the hosted service gets a minimal model-only POST to `/api/v2/ocr/jobs` (401/403 → token rejected, any other HTTP status → reachable), while a self-hosted PaddleX deployment only serves the synchronous `layout-parsing` endpoint, so it is probed with a plain GET (any HTTP response → reachable). Probing self-hosted via the jobs endpoint would have false-negatived healthy deployments.
- Mirroring the parse path, the token is only sent over HTTPS.

Unit tests updated for both paths (`test/unit_test/deepdoc/parser/test_paddleocr_parser_check_installation.py`, 9 tests, run with `--noconftest` to avoid the heavy deepdoc dependency chain). The diff is back to 2 files after the rebase.

JinHai-CN:
@HUAN2022A would you please check the comments from @coderabbitai

JinHai-CN:
@Souravrajvi0 would you please review this PR?

Souravrajvi0 (changes requested):
@HUAN2022A thanks for keeping this going and rebasing  ,overall I think this is the right fix for #17510.

I pulled the branch and ran the new tests (9/9 pass) plus the existing paddleocr suite (35/35). Also did a quick poke at the live hosted endpoint.

**Bottom line:** we should merge something like this, but I'd hold off until a few things are fixed.

Right now on main, `check_installation()` basically just checks that base_url/token are set. So bad config only shows up when you actually try to parse. Adding a real reachability check makes sense and matches what Mistral/SoMark already do.

Can you please fix these items

1. Self-hosted probe should go through `_local_endpoint()` — if someone already put `.../layout-parsing` in the URL, the current code hits `.../layout-parsing/layout-parsing`. Parse path already handles this, would be good to reuse here + add a test.

2. Hosted + HTTP is awkward — e.g. `http://PaddleOCR.AIStudio-App.com` still counts as hosted but we don't send the token over plain HTTP. Probe result can be misleading. I'd just fail early if it's hosted and not HTTPS.

3. Small test thing: in the missing-token test, clear `PADDLEOCR_ACCESS_TOKEN` from env. Otherwise `access_token=None` can still pick up a token from the environment and the test doesn't actually hit that branch.

4. Bigger one — the 401/403 assumption doesn't match what I saw against the real API. Model-only POST to `/api/v2/ocr/jobs` with no token / bad token / fake token all came back **500**, not 401. With the current logic that still counts as "reachable", so we're not really validating the token in practice. Might need to look at response body, pick a different endpoint, or be explicit that this is connectivity-only.

**Lower priority:** self-hosted check accepts any HTTP response, so if something else is listening on that port (nginx on 8080 returned 200 for `/layout-parsing` on my side) you get a false positive. And the SSRF point from CodeRabbit is probably a maintainer call since users configure these URLs anyway.

Happy to take another look once 1–4 are addressed.

HUAN2022A:
Addressed the review comments in 1d0cba82:

- **Hosted HTTP URLs** (`deepdoc/parser/paddleocr_parser.py`): a hosted base URL that is not HTTPS now fails as a configuration error before any request, so an unauthenticated probe can no longer misreport the outcome.
- **Self-hosted probe URL**: now uses the existing `_local_endpoint()` helper, so a configured address that already ends with `layout-parsing` is probed as-is instead of getting the path doubled. Regression tests added for both cases, plus `PADDLEOCR_ACCESS_TOKEN` is cleared in the tests that rely on a missing/unset token.
- **SSRF hardening (SAST finding)**: skipping for this PR. The probe targets the same user-configured `base_url` the parse path already POSTs file uploads to, so this PR adds no new request surface — and an allowlist would reject legitimate self-hosted deployments on localhost/LAN addresses, which the self-hosted support in #19011 explicitly serves. Redirect validation for the parser's requests feels like a codebase-wide follow-up rather than something to diverge inside `check_installation`.

JinHai-CN:
@HUAN2022A 
```
=========================== short test summary info ============================
FAILED test/unit_test/deepdoc/parser/test_paddleocr_parser.py::test_self_hosted_needs_base_url_not_token - assert False is True
== 1 failed, 4880 passed, 30 skipped, 11 subtests passed in 168.95s (0:02:48) ==
```

Souravrajvi0:
Hi @HUAN2022A — thanks for the follow-up on #17510 and for aligning the probe with self-hosted vs hosted (#19011). The connectivity check in `check_installation()` looks solid (HTTPS gate for hosted, `_local_endpoint()` for self-hosted, 401/403 handling on the jobs POST).

**Blocker before merge:** `ragflow_preflight` is currently red on the latest run ([job log](https://github.com/infiniflow/ragflow/actions/runs/35066487466/job/104697822933)) — 4880 tests pass, 1 fails:

`test/unit_test/deepdoc/parser/test_paddleocr_parser.py::test_self_hosted_needs_base_url_not_token`

That test still expects `check_installation()` to return `(True, "")` for `http://127.0.0.1:8080` without mocking HTTP. After this PR, the check performs a real `GET` to `…/layout-parsing`; on CI nothing is listening → connection refused → `ok is False`.

**Suggested fix:** In `test_paddleocr_parser.py`, mock `requests.get` for that case (e.g. return a 404), matching what you already did in `test_paddleocr_parser_check_installation.py` (`test_self_hosted_probe_uses_layout_parsing_get`).

**Non-blocking follow-up:** `test_hosted_still_requires_token` still asserts `ok is True` with `access_token="tok"` without mocking, so it can hit the real hosted endpoint — worth mocking `requests.post` there too for stability.

Once preflight is green, this looks good to merge from my side.

HUAN2022A:
Hi @Souravrajvi0 @JinHai-CN — thanks for catching this.

Fixed in 025ee797:

- `test_self_hosted_needs_base_url_not_token` now stubs `requests.get` with a 404 response, matching the approach in `test_self_hosted_probe_uses_layout_parsing_get` — any HTTP response counts as reachable, so the probe no longer depends on something actually listening on CI.
- `test_hosted_still_requires_token` (the non-blocking follow-up) now also stubs `requests.post` with a 422, mirroring the "valid token yields a non-401/403" path, so the suite never reaches the real hosted service.

Local run of both PaddleOCR test files: 46 passed. Preflight should go green once CI re-runs; the PR is then ready for merge from your side.

PR #17670, opened 2026-07-31
fix(graph): use max-wins for PageRank in AnalyzeNHopPaths to match Python

### LINKED ISSUE #15695: AnalyzeNHopPaths uses last-wins instead of max for PageRank

Taranum01:
Fixes #15695

## Summary
Go's `AnalyzeNHopPaths` overwrote `PageRank` for shared edges on every new observation (last-write-wins). When the same edge `(f, t)` appears from multiple N-hop paths, this discards the strongest observation and — because Go map iteration is randomized — produces non-deterministic results across runs.

## Fix
Single-line change in `internal/service/graph/scoring.go:43`:
```go
- if i < len(weights) {
+ if i < len(weights) && weights[i] > es.PageRank {
      es.PageRank = weights[i]
  }
```

The Python equivalent in `rag/graphrag/search.py:186` already uses `max(...)`; this brings Go to parity.

## Note
`n_hop_with_weight` is never populated today, so `AnalyzeNHopPaths` is effectively dead code and this fix has no runtime effect until N-hop indexing is implemented. Tracked for parity.

## Tests
Added `internal/service/graph/scoring_test.go` with two regression tests:
- `TestAnalyzeNHopPathsPageRankMaxWins` — same edge from two paths with weights [0.3, 0.7] and [0.9, 0.1] keeps max=0.9, not last-wins 0.1
- `TestAnalyzeNHopPathsPageRankGoOrderingIndependent` — same scenario with paths in opposite order still yields max-wins result

`go vet ./internal/service/graph/...` passes. Full test run requires the C++ tokenizer library at `internal/binding/cpp/cmake-build-release/librag_tokenizer_c_api.a` which isn't built in this environment.


Taranum01:
Closing stale bot feedback — CodeRabbit's stale comment refers to a pre-amend fixture. The current `scoring_test.go` already has both paths sharing the (mid, dst) edge with competing weights (0.7 vs 0.5); see commit `9d41300`.

xugangqiang:
Review of PR #17670 — issues found (review only, no code changes):

**1. `TestAnalyzeNHopPathsPageRankGoOrderingIndependent` does not actually exercise the bug**
The headline problem is non-determinism from randomized `map` iteration over `entsFromQuery` (`scoring.go`). This test only reverses the `NhopEnts` *slice* order within a **single** entity, which is deterministic. Under the old last-wins code, this test still yields 0.7 (path2→0.5 then path1→0.7), so it **passes on both old and new code** and does not discriminate the fix. To truly validate order-independence, use **two distinct query entities** that share the same edge with competing weights and run the assertion in a loop (e.g. 50 iterations) to exercise randomized map iteration.

**2. The `i < len(weights)` guard branch is untested**
Both tests use `Weights` whose length equals `len(path)-1`, so the short-weights branch (edge gets `PageRank == 0`, no panic) is never covered. Add a case with `Path` longer than `Weights` to lock in the boundary behavior.

**3. Missing trailing newline in `scoring_test.go`**
The diff shows `\ No newline at end of file`; `gofmt` will flag this.

**4. The "effectively dead code" framing is imprecise**
`AnalyzeNHopPaths` is actually wired into the pipeline (`pipeline.go:162`) and its `PageRank` flows into `FuseRelationScores`. It only has no runtime effect because upstream indexing doesn't currently populate `n_hop_with_weight`. Worth correcting the description so reviewers don't mistake it for removable dead code.

**5. Note on the existing bot review**
CodeRabbit reported "no actionable comments" and its "selected/skipped" logs show it only looked at `scoring_test.go` and skipped it — meaning the actual one-line logic change in `scoring.go` was never reviewed. Its passing pre-merge checks are cosmetic (title/description/linking) and should not be read as validation of the fix.

Minor nit (optional): `weights[i] > es.PageRank { es.PageRank = weights[i] }` reads more clearly as `es.PageRank = math.Max(es.PageRank, weights[i])`.

xugangqiang:
pls resolve conflict

Taranum01:
Rebased onto upstream/main (HEAD `d37486e2`) and the conflict is resolved. Force-pushed to `fix/15695-pagerank-max-wins`.

Final diff vs upstream/main:
```
 internal/service/graph/scoring.go | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)
```

Both conflicts (scoring.go and scoring_test.go) were resolved by taking upstream/main's versions, which already carry the equivalent max-wins fix and the corrected shared-edge test fixture (per the previous review). The branch's only remaining change is the `math.Max` in `scoring.go` — the actual fix.

`gofmt -l` clean. `go build ./internal/service/graph` clean (full `go test ./...` blocked locally by the same `office_oxide.h` cgo transitive dep as in previous PRs, unrelated).

xugangqiang:
we found this issue https://github.com/infiniflow/ragflow/pull/18784 - it‘s merged and seems like dup.

Issue #17438, opened 2026-07-27
refactor: extract shared TCADP submit/download logic from pptx/xls/pdf parsers


xugangqiang:
## Problem
`pptx_tcadp.go`, `xls_tcadp.go`, and `pdf_parser_tcadp.go` share ~85% identical logic — base64 encoding, `POST /reconstruct_document`, polling download URL, extracting zip items. They differ only in `fileType` string.

## Proposal
Extract a single `parseWithTCADP(ctx, filename, data, fileType, server, apiKey, tableResultType, markdownImageResponseType, outputFormat) ParseResult` function. Each family provides a thin wrapper:

```go
// pptx_tcadp.go
func parsePresentationWithTCADP(ctx, filename, data, fileType, ...) ParseResult {
    return parseWithTCADP(ctx, filename, data, fileType, ...)
}
```

## Motivation
- DRY principle: ~200 lines of duplicated code across 3 files
- Easier maintenance: single point for TCADP API changes (error handling, config fields)
- Found during review of PR #17419

PR #17671, opened 2026-07-31
refactor(parser): extract shared parseWithTCADP core from pptx/xls/pdf wrappers

Taranum01:
Refs #17438

## Summary

Three Go parser files (pptx_tcadp.go, xls_tcadp.go, pdf_parser_tcadp.go) each implemented the same ~70-line TCADP-reconstruct flow: base64 encode → POST /reconstruct_document → decode result URL → GET download URL → read zip → extract items → build ParseResult. They differed only in `fileType` string and where the parameters came from (params vs parser struct).

This change extracts that shared flow into a single `parseWithTCADP` function in a new `tcadp_common.go`.

## Wrapper treatment (post-review)

After the review, the wrappers are uniformly treated:
- `parsePresentationWithTCADP` was removed (no caller in the CGO_ENABLED=0 production build, see `office_parsers_no_cgo.go`).
- `parseSpreadsheetWithTCADP` was a pure pass-through with no added behaviour, so it was deleted and the three callers (csv_parser.go, xls_parser.go, xlsx_parser.go) call `parseWithTCADP` directly.
- `parsePDFWithTCADP` is kept because it maps *PDFParser fields into the core's argument list — not a pure pass-through.

## Environment fallback

The shared core reads the TCADP base URL from:
1. The explicit caller argument.
2. `TCADP_APISERVER_URL` env var (preserved for backward compat with the old `parsePDFWithTCADP` wrapper, which honoured it).
3. `TCADP_APISERVER` env var (used by the other two families and the error message).

The previous version of this PR silently dropped support for `TCADP_APISERVER_URL`; the review caught it. The new ordering honours both env vars so existing PDF deployments keep working.

Unit tests cover all four branches of the ordering (explicit arg wins, legacy env var fallback, canonical env var fallback, empty when nothing is set).

## ctx consistency — DEFERRED

The original PR summary claimed a "side benefit: ctx consistency" but the refactor does not actually achieve it: pptx already threads ctx through; xls and pdf still pass `context.Background()` (their wrappers were updated to forward the caller's ctx, but the callers themselves don't yet propagate it). Forwarding ctx through the XLSX/XLS/CSV/PDFParser call sites is deferred to a follow-up PR (#17673).

## Stats

415 → 290 lines in this slice (~30% reduction). All helpers (`tcadpItemsFromZip`, `tcadpAnyToItems`, `extractTCADPPage`, `tcadpTableRowsText`, `bearer`) and behaviour preserved verbatim except for the env-var fallback fix.

`gofmt` clean on all five files. Full `go vet` blocked locally by the same office_oxide cgo transitive dep as #17673.

Note: the previously-bundled `fix(graph): use max-wins for PageRank in AnalyzeNHopPaths` has been moved to PR #18784 to keep this refactor focused on the TCADP parser subsystem.


Taranum01:
Closing stale bot feedback — All three CodeRabbit follow-ups addressed in follow-up PRs against this branch: `context` import fix (amended), bearer host scope (#17672), ctx forwarding (#17673). Suggest resolving this thread once those are reviewed.

JinHai-CN:
@Taranum01 Would you please resolve the conflicts?

Taranum01:
@coderabbitai re-review

xugangqiang:
Review findings (issues only):

## 🔴 Blocking: both new scoring tests will fail

The fixtures in `scoring_test.go` cannot satisfy their own assertions, so `go test ./internal/service/graph/...` fails on the current diff.

**Test `TestAnalyzeNHopPathsPageRankMaxWins`:**
```
Path 1: ["src","mid","dst"]  Weights [0.3, 0.7]
Path 2: ["src","alt","dst"]  Weights [0.9, 0.1]
```
Actual edges produced: `src->mid=0.3`, `mid->dst=0.7`, `src->alt=0.9`, `alt->dst=0.1`.
The assertion `got[Edge{"src","mid"}].PageRank == 0.9` can never pass — `src->mid` only ever receives weight `0.3`. The second path does not contain the `src->mid` edge at all.

**Test `TestAnalyzeNHopPathsPageRankGoOrderingIndependent`:**
assertion `src->mid PageRank == 0.7` also fails — actual value is `0.3`.

Neither fixture exercises the same edge appearing in two paths with different weights, so they don't actually distinguish max-wins from last-wins. Fix is to make both paths contain the asserted edge, e.g.:
```
Path 1: ["src","mid","dst"]  Weights [0.3, 0.7]  -> src->mid = 0.3
Path 2: ["src","mid","alt"]  Weights [0.9, 0.1]  -> src->mid = 0.9
```
so max-wins yields 0.9 while last-wins yields 0.3.

## 🟡 Undocumented behavior change: PDF baseURL env var renamed

The PR states "pure code-organisation change with no semantic diff", but one semantic diff exists.

Before:
- PDF path fell back to `common.GetEnv(common.EnvTCADPAPIServerURL)` = `os.Getenv("TCADP_APISERVER_URL")` for baseURL.

After:
- PDF path calls the shared core (derived from pptx), which uses `os.Getenv("TCADP_APISERVER")` for baseURL.

`GetEnv` is just a wrapper around `os.Getenv` (`internal/common/environments.go:24`), so the only difference is the env var name: **`TCADP_APISERVER_URL` → `TCADP_APISERVER` for the PDF parser's baseURL fallback.** pptx and xls already used `TCADP_APISERVER`.

This is likely an intentional consistency fix (the error message already says `TCADP_APISERVER`, and 2/3 of the code uses that name), but:
1. It is undocumented in the PR description, which claims no semantic diff.
2. Anyone relying on `TCADP_APISERVER_URL` for PDF-specific server config would break silently.
3. There is no test covering the env-var fallback path.

Suggestion: document this as an explicit consistency fix in the PR description, or confirm `TCADP_APISERVER` is the intended canonical name and add a test for the fallback.

## 🟡 Misleading PR description: "ctx consistency" not achieved

The summary claims a "side benefit: ctx consistency", but the refactor does **not** make ctx consistent:
- pptx forwards `ctx` (unchanged).
- xls wrapper still passes `context.Background()` (unchanged).
- pdf wrapper still passes `context.Background()` (unchanged — the new wrapper at `pdf_parser_tcadp.go:23` calls `parseWithTCADP(context.Background(), ...)`).

The callers (`XLSXParser.ParseWithResult`, `XLSParser.ParseWithResult`, `CSVParser.ParseWithResult`, `PDFParser.ParseWithResult`) all receive a `ctx`, but xls/pdf still discard it. Suggestion: reword the description to note this inconsistency is *deferred* (e.g. to #17673) rather than claimed as resolved.

## 🟡 Minor: inconsistent wrapper treatment across families

After the refactor, pptx inlines `parseWithTCADP` directly (no wrapper), while xls keeps a thin passthrough `parseSpreadsheetWithTCADP` and pdf keeps `parsePDFWithTCADP`. Either inline all three (updating the csv/xlsx/xls/pdf call sites) or keep all three wrappers — but pick one for consistency. The xls wrapper in particular adds a pure forwarding layer with no added logic.

Taranum01:
Thanks for the review, @xugangqiang. All four findings addressed in this force-push.

## 🔴 Blocking — scoring tests broken
The scoring fix is now on `fix/pagerank-max-wins` (PR #18784). I dropped the scoring commits from this branch entirely so the broken tests are no longer in this PR's diff. The corrected fixtures (sharing the contested edge) live on #18784.

## 🟡 Undocumented env var rename (TCADP_APISERVER_URL → TCADP_APISERVER)
You're right that the previous version of this PR silently dropped support for `TCADP_APISERVER_URL`. Fixed by extracting a `tcadpAPIBaseURL()` helper that resolves the base URL in this order:
1. explicit caller argument
2. legacy `TCADP_APISERVER_URL` env var (backward compat)
3. canonical `TCADP_APISERVER` env var

The error message now mentions all three sources. Unit tests cover all four branches of the ordering (explicit arg wins, legacy env var, canonical env var, empty when nothing is set) in the new `tcadp_common_test.go`.

## 🟡 Misleading PR description (ctx consistency claim)
You're right — the refactor doesn't actually make ctx consistent. pptx already threads ctx through; xls and pdf still pass `context.Background()` because the callers themselves don't yet propagate it. The PR description has been rewritten to explicitly defer this to a follow-up PR (#17673).

## 🟡 Inconsistent wrapper treatment
You're right — the inconsistency between inlined pptx and pass-through xls/pdf was confusing. Fixed by deleting `parseSpreadsheetWithTCADP` (the pure pass-through) and updating the three spreadsheet callers to call `parseWithTCADP` directly. `parsePDFWithTCADP` stays because it maps `*PDFParser` fields into the core's argument list (not a pass-through).

## Final diff
```
$ git diff upstream/main..HEAD --stat
 internal/parser/parser/pdf_parser_tcadp.go         | 81 ++------------------
 internal/parser/parser/pptx_parser.go              |  2 +-
 .../parser/{pptx_tcadp.go => tcadp_common.go}      | 20 +++--
 internal/parser/parser/xls_tcadp.go                | 87 ++--------------------
 4 files changed, 28 insertions(+), 162 deletions(-)
```

Plus a new `tcadp_common_test.go` (4 tests, 56 lines) and updated `csv_parser.go`, `xls_parser.go`, `xlsx_parser.go` (12 lines changed total, 6 each).

No more scoring.go / scoring_test.go in this PR's diff — moved to PR #18784.

xugangqiang:
LGTM @yuzhichang

PR #17672, opened 2026-07-31
fix(parser): scope TCADP bearer token to API server host only

Taranum01:
Follow-up to #17671 — review flagged that the bearer token was being sent to the TCADP result-download URL unconditionally, leaking the API key if the URL was ever attacker-influenced (presigned S3, CDN, malicious mirror).

## Fix

Extract `authHeaderForDownload(apiBaseURL, downloadURL, apiKey)` helper that returns the bearer value only when both URLs parse cleanly AND their `.Host` fields match. Parse failures default to the safer "no auth" outcome.

`parseWithTCADP` now calls the helper instead of attaching `bearer(apiKey)` directly.

## Tests

`internal/parser/parser/tcadp_common_test.go` — table-driven cases:
- matching host, with port, with trailing slash on base → auth sent
- different host (S3, CDN), subdomain mismatch → no auth
- unparseable download URL, unparseable API URL → no auth (safe default)
- empty API key → no auth

PR #17673, opened 2026-07-31
fix(parser): forward caller ctx through TCADP wrappers + env coverage tests

Taranum01:
Follow-up to #17671 — review flagged that PDF and spreadsheet TCADP wrappers drop caller context by hardcoding `context.Background()`. Cancellation/deadline signals from the parser entry points never reach the TCADP network calls.

## Fix

Add `ctx context.Context` as the first parameter to:

- `parsePDFWithTCADP` (pdf_parser_tcadp.go)
- `parseSpreadsheetWithTCADP` (xls_tcadp.go)

Update the five callers to forward their own ctx:

- pdf_parser_cgo.go, pdf_parser_nocgo.go — case "tcadp" dispatch
- xlsx_parser.go, xls_parser.go, csv_parser.go — case "tcadp" dispatch

`parsePresentationWithTCADP` (pptx_tcadp.go) already accepted ctx.

## Behaviour change — ctx propagation

Before: a cancelled parser job kept both the TCADP submit and download HTTP requests running in the background until the cloud service responded.

After: cancellation/deadlines flow through `http.NewRequestWithContext` inside `parseWithTCADP`, so in-flight TCP connections and server-side work are torn down promptly.

## Behaviour change — TCADP_APISERVER_URL env var

The consolidated `parseWithTCADP` core reads `TCADP_APISERVER` via `os.Getenv`. The old PDF wrapper had been reading `TCADP_APISERVER_URL` via `common.GetEnv(EnvTCADPAPIServerURL)`. Without explicit handling, deployments that set only `TCADP_APISERVER_URL` for PDF-specific config would silently break — the parse would now return `requires tcadp_apiserver or TCADP_APISERVER` instead of using the configured PDF URL.

To preserve backward compatibility, the core now reads both env vars, in this order:

1. The explicit `tcadpAPIServer` argument.
2. The legacy `TCADP_APISERVER_URL` env var (preserved for PDF deployments that set it under the old wrapper).
3. The canonical `TCADP_APISERVER` env var (used by the other two families and the error message).

The error message now mentions all three sources so operators know which names to set. `EnvTCADPAPIServerURL` remains declared in `internal/common/environments.go` for any callers that still prefer the indirection. Note: the `common.GetEnv` indirection the old PDF wrapper used was just a wrapper around `os.Getenv`, so the raw `os.Getenv` here is functionally equivalent for now.


Taranum01:
Closing stale bot feedback — Stale CodeRabbit review refers to `scoring_test.go` which lives on PR #17670, not this one. This PR only adds `ctx` parameters to the TCADP wrappers — no test file changes.

JinHai-CN:
@Taranum01 Would you please resolve the conflicts?

Taranum01:
@coderabbitai re-review

xugangqiang:
## Review Findings

### 1. Test fixtures are broken (blocking)

Both new tests in `scoring_test.go` have incorrectly designed fixtures — the paths within each test **do not share any edge**, so they cannot verify the "max-wins over shared edges" behavior they claim to test.

Trace for `TestAnalyzeNHopPathsPageRankMaxWins`:
- Path `[src, mid, dst]` weights `[0.3, 0.7]` → edges: `(src,mid)=0.3`, `(mid,dst)=0.7`
- Path `[src, alt, dst]` weights `[0.9, 0.1]` → edges: `(src,alt)=0.9`, `(alt,dst)=0.1`

These two paths have **no edge in common**, yet the test asserts `edge(src,mid).PageRank == 0.9` (actual: `0.3`) and claims the second path contributes `0.9` to `src->mid`. Both `TestAnalyzeNHopPathsPageRankMaxWins` and `TestAnalyzeNHopPathsPageRankGoOrderingIndependent` will **fail**.

To actually test max-wins, the fixtures need paths that share the same edge, e.g.:
```go
Path: ["src", "mid", "dst"],   Weights: [0.3, 0.7]  // edge(src,mid)=0.3
Path: ["src", "mid", "other"], Weights: [0.9, 0.2]  // edge(src,mid)=0.9 → max should win
```

### 2. Two unrelated changes mixed in one PR (blocking)

This PR combines:
- **A**: TCADP context propagation fix (9 parser files)
- **B**: `scoring.go` PageRank max-wins fix + new test

These are independent concerns and should be in separate PRs. The scoring fix cannot be properly reviewed/merged while coupled to the parser refactor.

### 3. Merge conflict (blocking)

@JinHai flagged on Aug 19 that this branch has conflicts that need resolving. These are still unresolved.

### 4. Pass-through wrapper should be deleted (follow-up)

After the refactor, `parseSpreadsheetWithTCADP` in `xls_tcadp.go` is a pure pass-through that forwards every argument in order with no added behavior. Per the project's own coding guidelines ("prefer deletion over shims, wrapper APIs"), it should be removed and its three callers (`csv_parser.go`, `xls_parser.go`, `xlsx_parser.go`) should call `parseWithTCADP` directly. (Keeping `parsePDFWithTCADP` is fine — it maps `*PDFParser` fields and is not a pass-through.)

Taranum01:
Thanks for the thorough review, @xugangqiang. All four findings addressed in this force-push:

## Review #1 — broken test fixtures
Reworked `scoring_test.go` so the two paths actually share the contested `(src, mid)` edge with weights 0.3 and 0.9. The original disjoint paths (`src, mid, dst` and `src, alt, dst`) produced no shared edge and would have silently passed either way. `TestAnalyzeNHopPathsPageRankMaxWins` now asserts the shared edge picks up 0.9 (max) and each path keeps its unique second-hop weight. `TestAnalyzeNHopPathsPageRankGoOrderingIndependent` asserts the result is independent of which path runs first.

## Review #2 — split into two PRs
The scoring PageRank max-wins fix + corrected tests now live on `fix/pagerank-max-wins` → PR #18784. This branch (`fix/17671-ctx-forwarding`) carries only the TCADP context-propagation changes.

## Review #3 — merge conflict
Force-pushed onto current upstream/main (HEAD 63426325, `Fix: MinerU model (#18767)`). The rebase was clean.

## Review #4 — drop pass-through wrapper
Deleted `internal/parser/parser/xls_tcadp.go` (parseSpreadsheetWithTCADP was a pure pass-through) and updated the three callers (csv_parser.go, xls_parser.go, xlsx_parser.go) to call `parseWithTCADP` directly. `parsePDFWithTCADP` in pdf_parser_tcadp.go stays — it maps `*PDFParser` fields and is not a pass-through.

`git diff upstream/main..HEAD --stat` now shows only the TCADP-related files:
```
internal/parser/parser/csv_parser.go               |  4 +-
internal/parser/parser/pdf_parser_cgo.go           |  2 +-
internal/parser/parser/pdf_parser_nocgo.go         |  2 +-
internal/parser/parser/pdf_parser_tcadp.go         | 83 ++-------------------
internal/parser/parser/pptx_parser.go              |  2 +-
internal/parser/parser/{pptx_tcadp.go => tcadp_common.go} | 20 +++--
internal/parser/parser/xls_parser.go               |  4 +-
internal/parser/parser/xls_tcadp.go                | 86 ----------------------
internal/parser/parser/xlsx_parser.go              |  4 +-
9 files changed, 31 insertions(+), 176 deletions(-)
```

scoring.go and scoring_test.go are now exclusively on #18784.

xugangqiang:
## Undocumented behavior change: PDF env var fallback

This PR consolidates the three TCADP wrappers into the shared `parseWithTCADP` (formerly `parsePresentationWithTCADP`). Because the shared function's env fallback now applies to **all** parser families, the PDF parser's server-URL fallback silently changes:

| | Before this PR | After this PR |
|---|---|---|
| PDF server URL fallback | `common.GetEnv(EnvTCADPAPIServerURL)` = **`TCADP_APISERVER_URL`** | `os.Getenv("TCADP_APISERVER")` |
| Spreadsheet / Presentation | `TCADP_APISERVER` | `TCADP_APIServer` (unchanged) |

So after merge, any deployment that relies on `TCADP_APISERVER_URL` (but not `TCADP_APISERVER`) for PDF parsing will fail to resolve the server and return a "requires tcadp_apiserver" error.

This is not called out in the PR description. Please either:
- explicitly document this as an intended behavior change so users can migrate their env vars, or
- pick the canonical env var name across all families (e.g. align spreadsheet/presentation to `TCADP_APISERVER_URL`) and note it.

The same consolidation also switches PDF from `common.GetEnv` to raw `os.Getenv` — functionally equivalent today, but it drops the indirection other parsers use. Worth a one-line note as well.

xugangqiang:
## Missing test coverage for the core fix (ctx propagation)

The whole point of this PR is to make cancellation/deadline from the caller's `ctx` reach the two in-flight HTTP calls (submit + download). But there is no test that actually verifies this propagation.

The existing tests (`pdf_parser_tcadp_test.go`, `xls_tcadp_test.go`) all use `t.Context()` and assert on happy-path / error-path results — none of them assert that cancelling the ctx aborts the HTTP requests. Without such a test, the fix is unprotected against regression.

Suggested addition: a test that passes a pre-cancelled (or cancel-after-submit) ctx and asserts the parse returns promptly with a context error rather than blocking until the test server responds. A handler that blocks on `ctx.Done()` inside `httptest` is the usual way to observe this.

Also worth noting: none of the existing tests exercise the env-var fallback path (they all set `tcadp_apiserver` explicitly), which is exactly the path that changes behavior for PDF in this PR (see separate comment).

Taranum01:
Thanks for the catch, @xugangqiang. Force-pushed the fix.

## TCADP_APISERVER_URL backward compat
The consolidated `parseWithTCADP` core reads `TCADP_APISERVER` via raw `os.Getenv`. The old PDF wrapper had been reading `TCADP_APISERVER_URL` via `common.GetEnv(EnvTCADPAPIServerURL)`, so deployments that set only `TCADP_APISERVER_URL` for PDF-specific config would silently break after this PR merged.

Applied the same fallback approach already on #17672:

```go
baseURL := strings.TrimSpace(tcadpAPIServer)
if baseURL == "" {
    if v := strings.TrimSpace(os.Getenv("TCADP_APISERVER_URL")); v != "" {
        baseURL = v
    } else {
        baseURL = strings.TrimSpace(os.Getenv("TCADP_APISERVER"))
    }
}
```

The error message now mentions all three sources (tcadp_apiserver arg, TCADP_APISERVER_URL env, TCADP_APISERVER env) so the operator knows which names to set.

## common.GetEnv indirection
The `common.GetEnv(EnvTCADPAPIServerURL)` call the old PDF wrapper used was just a wrapper around `os.Getenv`, so the raw `os.Getenv` here is functionally equivalent for now. `EnvTCADPAPIServerURL` remains declared in `internal/common/environments.go` for any callers that still prefer the indirection.

## PR description updated
Added a "Behaviour change — TCADP_APISERVER_URL env var" section to the PR body so the migration path is explicit. If you'd rather pick one canonical name across all families (e.g. align the others to `TCADP_APISERVER_URL`), I'd vote for `TCADP_APISERVER` as the canonical — 2/3 of the code already used that name and the error message already advertises it.

xugangqiang:
Thanks for the cleanup — consolidating the three near-identical TCADP wrappers into `parseWithTCADP` is the right move, and the `TCADP_APISERVER_URL` backward-compat handling is correct (verified `common.GetEnv` == `os.Getenv` and `EnvTCADPAPIServerURL == "TCADP_APISERVER_URL"`).

Main concern: **test coverage does not guard the behavior this PR changes.** No `_test.go` is touched, and the existing `pdf_parser_tcadp_test.go` / `xls_tcadp_test.go` only exercise the public `ParseWithResult(ctx, ...)` happy path. Specifically:

- **The new env precedence is untested.** Nothing verifies that a PDF deployment setting only `TCADP_APISERVER_URL` still works, that xls/pptx setting only `TCADP_APISERVER` still works, or that `TCADP_APISERVER_URL` takes priority over `TCADP_APISERVER`. This is exactly the regression the "restore backward compat" commit is protecting against, yet no test would fail if someone later removes the `TCADP_APISERVER_URL` branch.
- **ctx cancellation/deadline propagation is untested.** The core fix (caller ctx now reaches the TCADP submit + download) is only implicitly covered by the happy-path tests; there is no test proving an cancelled ctx actually aborts the in-flight cloud request.
- **`TestPDFParser_ParseWithResult_TCADPRequiresAPIServer` became more environment-sensitive.** Previously the PDF path only read `TCADP_APISERVER_URL`; now it also reads `TCADP_APISERVER`. If `TCADP_APISERVER` happens to be set in the test/CI env, this test will try a real dial and fail the `"tcadp_apiserver"` assertion.
- The refactor (delete `parseSpreadsheetWithTCADP`, rename `pptx_tcadp.go`) is compile-safe only because tests go through the public method, but there is no test asserting the merged core preserves the old per-family request construction.

Recommendation: add a table-driven unit test using `t.Setenv` that sets `TCADP_APISERVER_URL`, `TCADP_APISERVER`, and both, and asserts the host actually requested by `parseWithTCADP`; plus (optionally) a test that a cancelled ctx aborts the download. Without this, merging leaves the most important guarantee of this PR unverified.

Minor: in the shared core, prefer `common.GetEnv(common.EnvTCADPAPIServerURL)` / `common.GetEnv(common.EnvTCADPAPIKey)` over the raw `os.Getenv("TCADP_APISERVER_URL")` / `os.Getenv("TCADP_API_KEY")` strings, so the env names stay defined in one place (the very divergence that caused this bug).

Taranum01:
Thanks for the thorough review, @xugangqiang. All four test-coverage findings addressed and force-pushed.

## 1. Env precedence table-driven test
```go
func TestParseWithTCADPEnvPrecedence(t *testing.T) {
    cases := []struct {
        name           string
        setLegacy     string
        setCanonical  string
        expectSuffixIn string
    }{
        {"legacy TCADP_APISERVER_URL only", "https://legacy.tcadp.example.com", "", ""},
        {"canonical TCADP_APISERVER only", "", "https://canonical.tcadp.example.com", ""},
        {"both env vars set: legacy wins by precedence", "https://legacy.tcadp.example.com", "https://canonical.tcadp.example.com", ""},
        {"neither env var: helper refuses to dial", "", "", "tcadp_apiserver"},
    }
    // ... uses httptest spy that records r.URL.Path
    // ... asserts which env source the resolver chose
}
```

A future change that drops the legacy fallback will fail this test.

## 2. ctx cancellation test
```go
func TestParseWithTCAPDCtxCancellationAbortsDownload(t *testing.T) {
    // slow handler sleeps 1s; ctx deadline 50ms
    start := time.Now()
    res := parseWithTCADP(ctx, ..., slow.URL, ...)
    elapsed := time.Since(start)
    // Asserts: error mentions 'context'; elapsed < slow-handler/2
}
```

Without this fix, a cancelled parser job kept both TCADP HTTP requests running until the cloud service responded. The test proves the caller ctx now aborts the download mid-flight.

## 3. Existing test env sensitivity
Updated `TestPDFParser_ParseWithResult_TCADPRequiresAPIServer` to explicitly clear both env vars via t.Setenv before the test runs. An accidentally-set `TCADP_APISERVER` in CI env can no longer silently turn this test into a real-dial.

## 4. Centralize env names (minor)
Replaced raw `os.Getenv(\"TCADP_APISERVER_URL\")` / `os.Getenv(\"TCADP_APISERVER\")` / `os.Getenv(\"TCADP_API_KEY\")` with `common.GetEnv(common.EnvTCADPAPIServerURL)` / `common.GetEnv(common.EnvTCADPAPIServer)` / `common.GetEnv(common.EnvTCADPAPIKey)`. Added the previously-missing `EnvTCADPAPIServer = \"TCADP_APISERVER\"` constant to `internal/common/environments.go` so the env names live in one place — the very divergence that caused this bug.

## Verified
- `gofmt -l` clean on all four files
- `go build` blocked locally by the same `office_oxide.h` cgo transitive dep as in previous PRs (unrelated to this change)

Force-pushed to branch `fix/17671-ctx-forwarding` (commit `fb4d12fd`). Title updated to `fix(parser): forward caller ctx through TCADP wrappers + env coverage tests`.

Issue #16876, opened 2026-07-13
Admin server surfaces DEFAULT_SUPERUSER_EMAIL via 'list envs' but init_default_admin() hardcodes the account


kobihikri:
### Summary

The admin server surfaces `DEFAULT_SUPERUSER_EMAIL` as if it configures the bootstrap admin account
(it shows up in `list envs` and in `services.py`), but `init_default_admin()` hardcodes the email and
password, so setting `DEFAULT_SUPERUSER_EMAIL` (or `DEFAULT_SUPERUSER_PASSWORD`) doesn't change the
account that actually gets created.

To be clear up front: I know the `admin@ragflow.io` / `admin` default is **documented** and that the
CLI can change the password after first login — so this isn't a "hidden default credentials" report.
It's a smaller consistency point: one of these env vars *looks* effective but isn't wired to the
bootstrap.

### Observation

`admin/server/services.py` lists `DEFAULT_SUPERUSER_EMAIL` among the environments the admin service
reports ([`services.py:440`](https://github.com/infiniflow/ragflow/blob/e54e7ec7ef6c42b89e966473e7a2d96296997629/admin/server/services.py#L440)):

```python
env_kv = {"env": "DEFAULT_SUPERUSER_EMAIL", "value": os.getenv("DEFAULT_SUPERUSER_EMAIL", "admin@ragflow.io")}
```

But the account is actually created by `init_default_admin()`
([`admin/server/auth.py:88-100`](https://github.com/infiniflow/ragflow/blob/e54e7ec7ef6c42b89e966473e7a2d96296997629/admin/server/auth.py#L88-L100)), which hardcodes both fields:

```python
default_admin = {
    "password": encode_to_base64("admin"),
    "email": "admin@ragflow.io",
    ...
}
```

`check_admin()` (same file) hardcodes the same pair. Neither reads `DEFAULT_SUPERUSER_EMAIL` /
`DEFAULT_SUPERUSER_PASSWORD`. (A separate, opt-in path, `api/db/init_data.py:init_superuser`, *does*
read `DEFAULT_SUPERUSER_PASSWORD`, but the admin server doesn't use it.)

### Effect (inference)

An operator who sets `DEFAULT_SUPERUSER_EMAIL=me@corp.com` before first boot will see that value
echoed by `list envs`, but the admin account that gets created is still `admin@ragflow.io` — the env
[...]

PR #17674, opened 2026-07-31
fix(admin): honor DEFAULT_SUPERUSER_EMAIL/PASSWORD env vars in bootstrap

Taranum01:
Supersedes #17063 (closed as superseded due to unrebasable test-fixture drift).

## Summary

The admin server's bootstrap (`init_default_admin`) and first-login fallback (`check_admin`) hardcoded admin@ragflow.io / admin / admin. The admin CLI's "list envs" output and the EnvironmentsMgr API both advertise DEFAULT_SUPERUSER_EMAIL/PASSWORD/NICKNAME as configurable, but the bootstrap ignored them — operators setting custom env values saw the docs claim support while the code still created the default admin. This matches what `api/db/init_data.py:init_superuser` already does on the API side.

## Fix

Three module-level constants read from env at import time (with the existing defaults), used in both `init_default_admin` and the first-login `check_admin` fallback. The existing-superuser row lookup in `init_default_admin` now keys off the configured email so a custom `DEFAULT_SUPERUSER_EMAIL` steers the tenant-backfill branch correctly.

## Scope

Core change only — the previous PR's test scaffolding (`reload_admin_auth` fixture with sys.modules stubbing) conflicts with the evolved `test/unit_test/admin/conftest.py` on current main and was excluded from this fresh port. Tests can be re-added on the new fixture shape in a follow-up if needed.

Refs #16876


JinHai-CN:
@Taranum01 Would please fix the format issue?
```
┃  ruff-format ❯ 
Would reformat: admin/server/auth.py
1 file would be reformatted

exit status 1
```

admin/server/auth.py
```diff
@@ -104,16 +141,49 @@ def init_default_admin():
...
     else:
-        default_admin_rows = [u for u in users if u.email == "admin@ragflow.io"]
+        # Filter existing superuser rows by the configured
+        # ``DEFAULT_SUPERUSER_EMAIL`` *and* by active status. An
+        # inactive row matching the configured email must NOT count as
+        # the bootstrap admin — otherwise the tenant-backfill branch
+        # would skip a real active admin with a different email, and
+        # the "no matching active admin" error below would never
+        # surface.
+        default_admin_rows = [u for u in users if u.email == DEFAULT_SUPERUSER_EMAIL and u.is_active == ActiveEnum.ACTIVE.value]
```
Lynn-Inf: I'd rather not add the `is_active` check here. The purpose is to avoid missing an inactive default admin and accidentally creating a duplicate entry.

Lynn-Inf (changes requested):
One inline comment. Otherwise LGTM.

Lynn-Inf:
I noticed that the inline suggestion to "not add the `is_active` check here" hasn't been addressed yet. I'd recommend removing this check to avoid the risk of duplicate entries.

Taranum01:
@Lynn-Inf — pushed commit `a82a7801` addressing your request and the related whitespace-only CodeRabbit finding:

**1. `is_active` filter removed** (your request):
```python
default_admin_rows = [u for u in users if u.email == DEFAULT_SUPERUSER_EMAIL]
```
Activity gating already happens in `login_admin` (the `is_active == INACTIVE` branch at line 240), so the bootstrap path stays read-only and won't mint a duplicate entry.

**2. Whitespace-only env values** (CodeRabbit, addressed on this PR while here):
Split `_required_env` into two helpers:
- `_required_env` (preserves raw value — used for password so a password like `'  s3cret  '` keeps its whitespace and isn't silently mutated before hashing).
- `_required_env_stripped` (strips before the empty check — used for nickname/email). A whitespace-only `DEFAULT_SUPERUSER_NICKNAME=' '` now surfaces the intended 'is set to whitespace-only value' config error at startup instead of slipping through to a silent downstream failure.

`ruff format` and `ruff check` both pass on the file.

Lynn-Inf (approved):
LGTM

PR #17682, opened 2026-08-02
refactor: delete the unused discord_svr.py bot script

noron12234:
### Summary

Replaces the original event-loop fix on `rag/svr/discord_svr.py` with a deletion, per @6ba3i's review: the file is legacy and no longer used.

What the file was: a standalone script with placeholder credentials hardcoded at module level (`URL = '{YOUR_IP_ADDRESS:PORT}/...'`, `DISCORD_BOT_KEY = "xxxx"`), calling `loop.run_until_complete(client.start(...))` at import time. Nothing in the tree imports it.

Also drops the now-dangling ruff exclude that pointed at it:

```toml
-exclude = [".venv", "rag/svr/discord_svr.py"]
+exclude = [".venv"]
```

That exclude was the only remaining reference to the file anywhere in the repo. Checked with `grep -rn discord_svr . --exclude-dir=.git` — zero hits after the change.

The `discord-py` dependency is **not** removed. `common/data_source/discord_connector.py` and `api/channels/discord/channel.py` both still import the package, so the `audioop-lts` entry and its Python 3.13 note in `[project].dependencies` stay load-bearing.

Branch was rebased onto current `main` (it was 655 commits behind), so the diff is now just the two changes above.


6ba3i (changes requested):
Thank you for your help! unfortunately discord_svr.py is legacy code that isn't being used anymore, could you make the change of deleting the file.

noron12234:
@6ba3i Done — the PR is now the deletion you asked for instead of the event-loop patch. Rebased onto current `main` first, so the diff is 2 files: `rag/svr/discord_svr.py` removed, plus the `[tool.ruff]` exclude in `pyproject.toml` that pointed at it (the only other reference in the tree).

I left `discord-py` in `pyproject.toml` on purpose — `common/data_source/discord_connector.py` and `api/channels/discord/channel.py` still import it, so the `audioop-lts` 3.13 workaround is still needed. Say the word if you want those looked at separately.

6ba3i (approved):
LGTM

PR #17693, opened 2026-08-03
Fix overlapping CJK text lines in docx file preview

euvre:
### Summary

In the file manager's document preview for `.docx` files, wrapped paragraphs render with overlapping text lines while short single-line paragraphs appear widely spaced, producing inconsistent line pitch across the document.

Root cause: `@extend-ai/react-docx` computes the line height of paragraphs that do not declare explicit line spacing as `fontSize × 0.88`. Each line is absolutely positioned with an inline `line-height`, so application CSS cannot compensate. For CJK text (e.g. 12pt 宋体 → 16px glyphs on a 14px line pitch) the glyphs of adjacent lines overlap. Word renders such paragraphs with the font's natural line height (~1.3× for CJK fonts).

This PR normalizes the document before handing it to the viewer: the docx package is unpacked, and a `docDefaults`-level line spacing (`w:line="312"`, `lineRule="auto"`, i.e. 1.3 lines) is injected into `word/styles.xml` only when the document does not already define a default line spacing. Because paragraph spacing is merged per field along the docDefaults → paragraph style → direct formatting chain, any explicit line spacing defined by styles or paragraphs is respected and left untouched. If anything fails during normalization, the original file is used as-is.

Verified with synthetic CJK documents: paragraphs without spacing inherit the injected default after the fix, documents that already define line spacing are unchanged, and the normalization is idempotent.


wangq8 (commented):
Thanks for this fix! The CJK text overlap issue in docx previews is a real UX problem, and the root cause analysis is spot-on — `@extend-ai/react-docx` computes line height as `fontSize × 0.88` which is too tight for CJK glyphs.

### What I like

1. **Conservative patching**: Only injects `w:line="312"` when the document doesn't already define a default line spacing. Paragraphs and styles with explicit spacing are left untouched. This respects the document author's intent.

2. **Graceful fallback**: The try/catch wrapping ensures that any unpacking/repacking failure returns the original blob — the preview still works (with the old rendering), rather than breaking entirely.

3. **Idempotent**: The early return when `patched === stylesPart.content` avoids unnecessary repacking when the spacing already exists.

4. **Good PR description**: Clear problem statement, root cause, and verification approach.

### Suggestions

1. **Regex-based XML manipulation**: The `ensureDefaultLineSpacing` function uses regexes against raw XML strings. OOXML `styles.xml` has a well-known structure and this is pragmatic, but it is inherently fragile. A few edge cases that could slip through:
   - XML with attributes quoted in single quotes (`w:line='360'`)
   - Namespace prefix variations (unlikely in styles.xml but possible)
   - Comment or PI nodes between elements

   The fallback-to-original-blob pattern mitigates the risk. If you're open to a follow-up, a DOM-based approach (e.g., using the browser's `DOMParser`) would be more robust, but I won't block on this for the current fix.

2. **Performance**: `parseDocx` → modify → `packageToArrayBuffer` adds overhead on every preview load. For large docx files, this could add noticeable latency. Consider whether the result could be cached (e.g., by URL or content hash) to avoid re-processing on re-renders. Not a blocker for the initial fix.

3. **Export scope**: `normalizeDocxLineSpacing` and `ensureDefaultLineSpacing` are exported (`export const`) but only used within this file. Consider making them non-exported (`const`) unless they're intended to be part of the module's public API.

4. **`DEFAULT_LINE_SPACING_TWIPS = 312`**: This is 1.3 lines (312÷240). Some CJK fonts' natural line height is closer to 1.5×. 1.3 is a reasonable default but might still feel tight for some documents. This is subjective — what you have is a good baseline.

### Summary

Clean, well-scoped fix with proper safety nets. Approved with the minor suggestions above. Nice work!

dcc123456 (approved):

PR #17695, opened 2026-08-03
fix: implement Go syncer execution engine with RSS data source support

euvre:
### Summary

In Go mode, RSS data sources could never be imported into a knowledge base (the Python backend works fine). The Go syncer was a stub: `pollAndExecute` only logged, no `sync_logs` rows were ever consumed, and no RSS connector existed on the Go side, so sync tasks stayed in Schedule/Cancel forever.

This PR implements the full Go sync execution path, mirroring the Python reference (`rag/svr/sync_data_source.py`, `common/data_source/rss_connector.py`, `FileService.upload_document`):

- **Syncer engine** (`internal/syncer`): polls due sync/prune tasks honoring `refresh_freq`/`prune_freq` windows, claims tasks atomically, executes with the connector's `timeout_secs`, performs done/fail status bookkeeping, and reschedules the next run after success (mirrors `SyncBase.__call__`).
- **RSS connector**: SSRF-safe feed fetching (reuses `utility.FetchRemoteFileSafely`), RSS 2.0 + Atom parsing, deterministic document IDs (`rss:md5(...)` + `hash128` with legacy-ID compatibility), HTML-to-text content building, poll-window filtering, and batching.
- **Document import** (`DocumentService.UploadConnectorDocuments`): mirrors `duplicate_and_parse` + `FileService.upload_document` — content-hash dedupe (unchanged docs are not re-parsed), name dedupe, storage upload, file-manager linkage, and `doc_num` counters; parsing is enqueued when the link has `auto_parse` enabled.
- **DAO support**: due-task listing, atomic claim, `increase_docs`/`increase_removed_docs` accumulation, finish/reschedule with 100-row log retention; the first sync after link/rebuild is due immediately (Python `run_immediately=True`).
- **Prune**: when `sync_deleted_files` is enabled, documents that disappeared from the feed are deleted from the knowledge base.

Other sources (S3, Confluence, etc.) remain unimplemented in Go and now fail fast with a clear "not supported by the Go syncer" message instead of being silently marked done.

Verified: unit tests for feed parsing/poll windows/batching/HTML normalization, `Hash128Hex` parity with Python `hash128`, package tests for `syncer`/`dao`/`service/document`, and a full `build.sh --go` build.

### DIFF
--- a/internal/dao/connector.go
+++ b/internal/dao/connector.go
@@ -343,13 +343,201 @@ const (
 	connectorTaskTypePrune = "prune"
 )
 
+// DueSyncTask is the joined projection of one due sync_logs row with its
+// connector, connector2kb link and knowledgebase, mirroring the task dicts
+// produced by Python SyncLogsService._list_due_tasks_for_freq.
+type DueSyncTask struct {
+	ID               string         `gorm:"column:id"`
+	ConnectorID      string         `gorm:"column:connector_id"`
+	TaskType         string         `gorm:"column:task_type"`
+	KbID             string         `gorm:"column:kb_id"`
+	PollRangeStart   *time.Time     `gorm:"column:poll_range_start"`
+	PollRangeEnd     *time.Time     `gorm:"column:poll_range_end"`
+	NewDocsIndexed   int64          `gorm:"column:new_docs_indexed"`
+	TotalDocsIndexed int64          `gorm:"column:total_docs_indexed"`
+	FromBeginning    *string        `gorm:"column:from_beginning"`
+	ConnectorName    string         `gorm:"column:connector_name"`
+	Source           string         `gorm:"column:source"`
+	TenantID         string         `gorm:"column:tenant_id"`
+	TimeoutSecs      int64          `gorm:"column:timeout_secs"`
+	Config           entity.JSONMap `gorm:"column:config"`
+	RefreshFreq      int64          `gorm:"column:refresh_freq"`
+	PruneFreq        int64          `gorm:"column:prune_freq"`
+	KbName           string         `gorm:"column:kb_name"`
+	AutoParse        string         `gorm:"column:auto_parse"`
+}
+
+// ListDueTasks returns scheduled sync/prune tasks whose connector is scheduled
+// and whose last update is older than the given frequency (minutes), mirroring
+// Python SyncLogsService._list_due_tasks_for_freq (MySQL dialect).
+func (dao *ConnectorDAO) ListDueTasks(ctx context.Context, db *gorm.DB, taskType, freqField string) ([]*DueSyncTask, error) {
+	var tasks []*DueSyncTask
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Thanks for this comprehensive PR! The Go syncer was a significant gap in the Go mode, and this fills it well. Overall the implementation is solid — good architecture, good test coverage, and proper safety nets (SSRF-safe fetching, atomic task claiming, timeout enforcement, panic recovery).

A few observations and suggestions:

### High-level feedback

1. **Legacy doc ID compatibility** — nice touch handling both `kb_id:connector_id:doc_id` and `connector_id:doc_id` hash forms. This ensures existing indexed documents are found regardless of which format was used when they were first created.

2. **Atomic claiming** — the `ClaimTask` pattern (WHERE status=Schedule, update to Running) correctly prevents duplicate execution across workers. 

### Potential issues

1. **`connectorAutoParse("")` returns `true`**: Since `"" != "0" && "" != "false"` evaluates to true, an empty `auto_parse` field from the DB would enable auto-parsing. In Python, an empty/None value typically means no auto-parse. In practice the `connector2kb` table likely has a default of `"1"`, so this may not manifest, but it is brittle. Consider making the default explicit:
   ```go
   func connectorAutoParse(autoParse string) bool {
       return autoParse == "1" || autoParse == "true" || autoParse == "TRUE"
   }
   ```
   or at least treat empty as false.

2. **`RescheduleTask` transaction scope**: `pruneOldSyncLogs` runs outside the transaction in `RescheduleTask`, while `scheduleConnectorTask` runs inside it. If two workers concurrently reschedule the same connector/kb pair, they could both pass the pruning check before either transaction commits. Consider moving `pruneOldSyncLogs` inside the transaction.

3. **`executePruneTask` swallows feed fetch errors**: When `listEntryIDs` fails, it returns nil (success). This means the prune task is marked done and rescheduled with no deletions, which is conservative and won't cause data loss, but it silently skips a prune cycle. Consider whether this should fail the task to surface the error more visibly.

### Minor suggestions

4. **`Hash128Hex` location**: This is exported from `internal/service/document/document_upload_helpers.go` but is really a general-purpose utility (used in syncer and tests). Consider moving it to `internal/utility/` for better cohesion.

5. **`rss.go` `parseFeedTime`**: The layout list covers common formats. It might be worth adding ISO 8601 with fractional seconds (`2006-01-02T15:04:05.000Z`) and the comma-variant of RFC1123 (`Mon, 02 Jan 2006 15:04:05 GMT`), which some feeds use.

6. **Test `TestHash128HexParity`**: The test hardcodes a single expected hash. Consider adding a table-driven test with a few more inputs to guard against xxh3 library version changes.

### Summary

The PR is well-structured, mirrors the Python reference faithfully, and includes meaningful tests. The issues above are all edge cases or style nits — none are blockers. Nice work!

wangq8:
@Haruko386 already handle it.

PR #17704, opened 2026-08-03
fix(go-models): add StepFun Step Plan base URL

euvre:
### Summary

The StepFun (阶跃星辰) provider config only exposes the billing endpoints (`https://api.stepfun.ai/v1` and `https://api.stepfun.com/v1`). StepFun's Step Plan subscription uses a dedicated access point, `https://api.stepfun.com/step_plan/v1`, which is not selectable today.

This PR adds a `plan` URL option to `conf/models/stepfun.json` so users can pick the Step Plan base URL from the region/URL dropdown when configuring the StepFun provider. The Go backend resolves the base URL by region key lookup with no whitelist validation, and the frontend builds the dropdown dynamically from the configured URL keys, so no code changes are required.

### DIFF
--- a/conf/models/stepfun.json
+++ b/conf/models/stepfun.json
@@ -2,7 +2,8 @@
   "name": "StepFun",
   "url": {
     "default": "https://api.stepfun.ai/v1",
-    "china": "https://api.stepfun.com/v1"
+    "china": "https://api.stepfun.com/v1",
+    "plan": "https://api.stepfun.com/step_plan/v1"
   },
   "url_suffix": {
     "chat": "chat/completions",


wangq8 (commented):
LGTM. Simple config addition for the StepFun Step Plan endpoint. Since the Go backend resolves base URLs by region key lookup (no whitelist) and the frontend builds the dropdown dynamically from config keys, no code changes are needed.

One note:  is the documented Step Plan access point, so this looks correct.

wangq8 (commented):
LGTM. Simple config addition for the StepFun Step Plan endpoint. Since the Go backend resolves base URLs by region key lookup (no whitelist) and the frontend builds the dropdown dynamically from config keys, no code changes are needed.

One note: `https://api.stepfun.com/step_plan/v1` is the documented Step Plan access point, so this looks correct.

PR #17708, opened 2026-08-03
Go: clear stale task cancel flag so first re-parse after cancel runs

euvre:
### Summary:

When multiple documents are parsed concurrently in Go mode, tasks beyond the worker-pool capacity are RUNNING in the DB but nacked without a worker. Cancelling them sets the Redis `{task_id}-cancel` flag (1h TTL), and the MQ redelivery path (`StartRunning` STOPPING → STOPPED) finalizes the stop **without clearing the flag** — normally the running worker's `markStopped` clears it, but these tasks never had a worker.

The stale flag is then detected by the worker's pre-start check on the next run (`cancel flag detected before pipeline start, cancelling`), so **the first re-parse after a cancel is immediately cancelled and only the second re-parse actually runs**. The first cancelled attempt happens to delete the flag via `markStopped`, which is why the second attempt succeeds.

Fixes:

- `StartRunning`: clear the Redis cancel flag when finalizing STOPPING → STOPPED without a worker, mirroring the worker-side `markStopped` cleanup.
- `CreateAndEnqueue`: clear any leftover cancel flag when re-queueing a terminal (STOPPED/FAILED) task. At that point the previous run is terminal, so the flag can only be stale — a genuine cancel of the new run is re-set by `RequestStop` once the task is RUNNING again.
- Add `TestStartRunningFinalizesStoppingTask` covering the STOPPING → STOPPED redelivery path.

Verified with `bash build.sh --test ./internal/service/...` and `./internal/ingestion/...` (all pass).

### DIFF
--- a/internal/service/ingestion_task_service.go
+++ b/internal/service/ingestion_task_service.go
@@ -229,7 +229,16 @@ func (s *IngestionTaskService) StartRunning(ctx context.Context, taskID string) 
 		}
 		return task, nil
 	case common.STOPPING:
-		return s.transition(ctx, taskID, common.STOPPED)
+		task, err = s.transition(ctx, taskID, common.STOPPED)
+		if err != nil {
+			return nil, err
+		}
+		// The stop is finalized here without a worker (e.g. MQ redelivery of
+		// a task that was nacked before execution), so the Redis cancel flag
+		// that RequestStop set would otherwise leak until TTL and cancel the
+		// next run of this task at the worker's pre-start check.
+		clearCancelFlag(ctx, taskID)
+		return task, nil
 	case common.RUNNING, common.COMPLETED, common.STOPPED, common.FAILED:
 		return task, nil
 	default:
@@ -387,6 +396,11 @@ func (s *IngestionTaskService) CreateAndEnqueue(ctx context.Context, task *entit
 			if err != nil {
 				return nil, err
 			}
+			// The previous run is terminal, so any leftover Redis cancel flag
+			// is stale: a genuine cancel of the new run can only come through
+			// RequestStop once the task is RUNNING again. Clear it so the
+			// re-queued task is not cancelled at the worker's pre-start check.
+			clearCancelFlag(ctx, existing.ID)
 			if err = s.enqueueTask(existing.ID); err != nil {
 				if rollbackErr := s.rollbackRetriedTask(ctx, existing.ID, originalStatus); rollbackErr != nil {
 					return nil, fmt.Errorf("enqueue task %s: %w (rollback failed: %v)", existing.ID, err, rollbackErr)
@@ -425,6 +439,15 @@ func (s *IngestionTaskService) rollbackRetriedTask(ctx context.Context, taskID, 
 func (s *IngestionTaskService) rollbackCreatedTask(ctx context.Context, taskID string) error {
 	_, err := s.ingestionTaskDAO.Delete(ctx, dao.DB, taskID, nil)
 	return err
+}
+
+// clearCancelFlag removes the Redis cancel marker ({task_id}-cancel) that
+// RequestStop sets for a RUNNING task. No-op when Redis is unavailable —
+// the DB STOPPING status remains the fallback cancel signal.
+func clearCancelFlag(ctx context.Context, taskID string) {
+	if rc := redis2.Get(); rc != nil {
+		rc.Delete(ctx, fmt.Sprintf("%s-cancel", taskID))
+	}
 }
 
 func (s *IngestionTaskService) enqueueTask(taskID string) error {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix. The root cause is well-documented and the fix is surgically targeted at the two points where the cancel flag is known to be stale:

1. **StartRunning STOPPING→STOPPED (redelivery path)** — when a nacked task is finalized without a worker,  mirrors what the worker-side  would have done.
2. **CreateAndEnqueue re-queue** — when a terminal task is re-queued, any leftover flag is necessarily stale since the previous run is done.

The  helper gracefully degrades when Redis is unavailable, which is the right call.

**One suggestion:** Consider adding a test case that exercises the  path with a stale cancel flag. The existing  covers the redelivery path well, but the re-enqueue path isn't tested.

Otherwise LGTM.

wangq8 (commented):
Thanks for the fix. The root cause is well-documented and the fix is surgically targeted at the two points where the cancel flag is known to be stale:

1. **StartRunning STOPPING→STOPPED (redelivery path)** — when a nacked task is finalized without a worker, `clearCancelFlag` mirrors what the worker-side `markStopped` would have done.
2. **CreateAndEnqueue re-queue** — when a terminal task is re-queued, any leftover flag is necessarily stale since the previous run is done.

The `clearCancelFlag` helper gracefully degrades when Redis is unavailable, which is the right call.

**One suggestion:** Consider adding a test case that exercises the `CreateAndEnqueue` path with a stale cancel flag. The existing `TestStartRunningFinalizesStoppingTask` covers the redelivery path well, but the re-enqueue path isn't tested.

Otherwise LGTM.

xugangqiang:
@euvre 

Thanks for the fix — the root cause analysis is solid, and the two clear sites (`StartRunning` STOPPING→STOPPED, and `CreateAndEnqueue` re-queue) correctly cover the gap left by the worker-only `markStopped` cleanup. Two suggestions to make this PR complete before merge:

**1. Make the tests actually prove the fix (and cover `CreateAndEnqueue`)**

`TestStartRunningFinalizesStoppingTask` only asserts `task.Status == STOPPED`. It does **not** verify the core behavior — that the `{task_id}-cancel` key is deleted. Since `redis2.Get()` returns `nil` in the default unit run, `clearCancelFlag` is a no-op there and the test cannot observe it at all. This is exactly the gap @wangq8 flagged for the re-queue path too.

To make it observable, expose a tiny test hook on the redis package so a test can install a `miniredis`-backed global client (the same package already uses `miniredis` in `agent_cancel_test.go` via `redis.NewClient(&redis.Options{Addr: mr.Addr()})`):

```go
// in internal/engine/redis/redis.go
func SetTestGlobalClient(rc *redis.Client) *Client {
	c := &Client{
		client:           rc,
		luaDeleteIfEqual: redis.NewScript(luaDeleteIfEqualScript),
		luaTokenBucket:   redis.NewScript(luaTokenBucketScript),
	}
	globalClient = c
	return c
}
```

Then in the test:
- `TestStartRunningFinalizesStoppingTask`: start miniredis, `mr.Set("task-1-cancel", "x")`, run `StartRunning`, and assert `!mr.Exists("task-1-cancel")` in addition to the status check.
- Add `TestCreateAndEnqueueClearsStaleCancelFlag` covering both `STOPPED` and `FAILED`: pre-set the cancel flag on the terminal task, re-queue via `CreateAndEnqueue` (use the existing `recordingTaskPublisher` fake so no real MQ is needed), and assert the flag is gone and the task is re-enqueued as `CREATED`.

This way the tests lock in the actual fix, not just the status transition.

**2. Collapse the three copies of the cancel-flag deletion**

After this PR there are three near-identical `rc.Delete(ctx, fmt.Sprintf("%s-cancel", taskID))` snippets: `markStopped` (`internal/ingestion/service/ingestion_service.go:461`), `runTask` (`:503`), and the new `clearCancelFlag`. That triplication is exactly what AGENTS.md asks us to avoid. Simplest fix: turn `clearCancelFlag` into a shared `redis.Client` method and have all three call sites use it:

```go
// in internal/engine/redis/redis.go
func (r *Client) ClearCancelFlag(ctx context.Context, taskID string) {
	if r == nil || r.client == nil {
		return
	}
	r.Delete(ctx, fmt.Sprintf("%s-cancel", taskID))
}
```

Then `markStopped` / `runTask` call `rc.ClearCancelFlag(ctx, taskID)` (keeping their existing `utility.BestEffort` wrappers), and the service's `clearCancelFlag` delegates to it. One source of truth for the key format and cleanup semantics.

**One note on CodeRabbit's actionable comment:** the suggestion to inspect `Delete`'s return value and surface a failure is inconsistent with the two existing call sites in this very file, which deliberately ignore the bool (comment: "Delete returns bool; the bool does not distinguish 'not found' from 'error'"). The cleanup is best-effort by design — a Redis hiccup must not fail the stop finalization. I'd recommend not adopting that change.

Net: the fix is correct and safe to merge; strengthening the tests (point 1) is the main thing worth doing before merge, and point 2 is a small cleanup that keeps the package consistent.

PR #17713, opened 2026-08-03
fix(web): MCP edit dialog cancel/save buttons not closing

euvre:
### Summary

In user settings -> MCP, the edit MCP dialog could not be dismissed properly:

- **Cancel did nothing.** The footer's cancel button relied on Radix `DialogClose`, but the `Modal` component swallows `onOpenChange(false)` in its `handleChange` when `maskClosable` is `false` (set in #16449 to prevent accidental dismissal), so `hideModal` was never called. Only the X icon (which goes through `handleCancel`) could close the dialog. The cancel button now calls `hideModal` directly.
- **Save appeared unresponsive.** The save button's `disabled` state was gated only on the fresh test result (`testData`), while the tool list shown in the dialog falls back to the server's stored tools. On open, the dialog displayed the stored tools yet the save button stayed disabled until the user re-ran the test. Both now share the same fallback, so the button state matches what is displayed. Changing URL/server type/token still requires re-testing via the refresh button before saving (`fieldChanged` gate preserved).

Also removes the commented-out legacy `Dialog` implementation left over from the `Modal` migration.

### DIFF
--- a/web/src/pages/user-setting/mcp/edit-mcp-dialog.tsx
+++ b/web/src/pages/user-setting/mcp/edit-mcp-dialog.tsx
@@ -1,7 +1,7 @@
 import { Collapse } from '@/components/collapse';
 import { Button, ButtonLoading } from '@/components/ui/button';
 import { Card, CardContent } from '@/components/ui/card';
-import { DialogClose, DialogFooter } from '@/components/ui/dialog';
+import { DialogFooter } from '@/components/ui/dialog';
 import { Modal } from '@/components/ui/modal/modal';
 import { useGetMcpServer, useTestMcpServer } from '@/hooks/use-mcp-request';
 import { IModalProps } from '@/interfaces/common';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Three clean fixes in one change:

1. **Cancel button fix** — The `Modal` component's `handleChange` swallows `onOpenChange(false)` when `maskClosable` is `false` (set in #16449), so `DialogClose` had no effect. Using `hideModal` directly is the correct approach and matches what the X icon already does via `handleCancel`.

2. **Tools fallback fix** — Previously `tools` only reflected `testData` results, while the UI displayed a fallback via `nextTools`. This mismatch made the save button appear disabled even though tools were shown. Now `tools` itself does the fallback (`testData?.data` → stored tools), so `disabled` correctly reflects what the user sees. The `fieldChanged` gate still requires re-testing after URL/server/token changes, which is correct.

3. **Dead code removal** — Removes the commented-out legacy `Dialog` implementation. Always good to clean up.

No regressions, +13/-75 is a great cleanup ratio.

dcc123456 (approved):

PR #17717, opened 2026-08-03
fix: synchronize chat model configuration

buua436:

- Synchronize `llm_id` with `tenant_llm_id` when creating or updating chats.
- Validate that both identifiers refer to the same chat model.
- Apply the same synchronization and validation rules to rerank models.
- Support completing either identifier when only one is provided.

### Type of change

### DIFF
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -413,7 +413,7 @@ async def create():
             req["kb_ids"] = kb_ids
             req.pop("dataset_ids", None)
 
-        if "llm_id" not in req and "tenant_llm_id" not in req:
+        if req.get("llm_id") is None and req.get("tenant_llm_id") is None:
             req["llm_id"] = tenant.tenant_llm_id
         if "rerank_id" not in req and "tenant_rerank_id" not in req:
             req["rerank_id"] = ""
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Good synchronization logic between `llm_id`/`tenant_llm_id` and `rerank_id`/`tenant_rerank_id`. A few observations:

**`_normalize_model_pair` — well-structured.** Handles type validation, existence checks, ID-in-name resolution, the built-in rerank model case, and the clearing case. Applied consistently across `create()`, `update_chat()`, and `patch_chat()`.

**`_llm_model_type` helper** correctly handles the `model_type` as list case (`"vision" in model_type`).

**dialog_service.py changes** — the `prompt_config = dialog.prompt_config or {}` moved before the assert and `gen_conf = dialog.llm_setting or {}` are good defensive fixes for None values.

**Frontend sync** in `chat-settings.tsx` and `use-create-chat.ts` looks correct — keeps `tenant_llm_id` in sync with `llm_id` from the model selector.

**Minor observation:** After `_normalize_model_pair` validates model existence, the subsequent `_validate_llm_id`/`_validate_rerank_id` calls are partially redundant — but harmless extra safety. Not a blocker.

LGTM overall.

PR #17719, opened 2026-08-03
Fix: false success toast when deleting non-existent Langfuse config

euvre:
### Summary

When no Langfuse configuration has been recorded (all input fields are empty), deleting via Settings > API > Langfuse configuration still popped up a "Deleted successfully" toast, even though nothing was actually deleted.

Root cause: the delete API returns `code: 0` with `data: null` when no record exists, but the frontend `useDeleteLangfuseConfig` hook only checked `code === 0` and always showed the success message.

Fix: only show the success toast when `data.data` is truthy (a record was actually deleted); otherwise show a warning ("No Langfuse configuration to delete"). Added the new message key to the en, zh, and zh-traditional locales (other languages fall back to en). This works for both the Python and Go backends since both return `data: null` when no record exists.


wangq8 (commented):
Simple and clean fix. The added `data.data` guard correctly prevents the misleading success toast when no Langfuse config exists. Locale keys added for en, zh, and zh-traditional — other locales will fall back gracefully.

LGTM, no issues found.

PR #17722, opened 2026-08-03
fix(parser): fix error logging import

Aeovy:
### Summary

Fixes an import error that prevents `rag.flow.parser.parser` from loading because LiteLLM does not expose `logging` as an importable module, it's a bool value.

This PR replaces the invalid LiteLLM import with Python's standard-library `logging` module, which already provides the logging methods used by the parser.


JinHai-CN:
The issue is already fixed.

PR #17723, opened 2026-08-03
fix: align pipeline delimiter chunking

buua436:
## Summary

- Preserve delimiters in pipeline parser sections when requested.
- Recombine parser-pre-split text sections before delimiter-mode chunking.
- Make `delimiter_mode="delimiter"` ignore `chunk_token_size`.
- Remove delimiters from final chunks.
- Keep ordinary naive parser behavior unchanged.


wangq8 (commented):
Thanks for the PR! The goal of aligning delimiter-mode chunking in the pipeline makes sense. However, I have several concerns that need to be addressed:

---

### 1. Chunk size cap removed for ALL txt-parser callers

The `add_chunk` rewrite in `deepdoc/parser/txt_parser.py` changes behavior for **all** callers — not just the new delimiter-mode path:

- **Old**: checked `merged_tnum <= chunk_token_num` **before** appending — prevented overshoot.
- **New**: only checks if the current chunk is **already** over budget (`tk_nums[-1] > chunk_token_num`). A chunk that starts at 80 tokens can absorb a 100-token section and end up at 180 tokens, well past the cap.

Combined with removing the `_split_oversized_unit` call, `chunk_token_num` is no longer enforced by the txt parser. This contradicts the PR summary claim: `"Keep ordinary naive parser behavior unchanged."`

**Suggestion**: If delimiter-mode intentionally ignores `chunk_token_num`, that logic belongs in the chunker layer (where `delimiter_mode` is checked), not in the txt parser which is a shared utility. Restore the original `add_chunk` and keep `_split_oversized_unit`.

---

### 2. `_split_text_by_pattern` behavior change affects all callers

The function now **discards delimiters** instead of keeping them attached to each chunk. This is a breaking change for any caller that relied on delimiters being preserved. The comment update to `"discard delimiters"` confirms this is intentional, but it is not scoped to `delimiter_mode="delimiter"` — it affects all paths.

**Suggestion**: Either keep the old behavior for non-delimiter-mode callers, or add a `keep_delimiters` parameter to `_split_text_by_pattern` so the new discard behavior is opt-in.

---

### 3. Unit tests deleted without replacement

`test/unit_test/deepdoc/parser/test_txt_parser.py` (178 lines) is deleted entirely. These tests validated the strict-cap invariant — no chunk exceeding `chunk_token_num`. If the new behavior intentionally loosens that invariant, new tests should be added to document the expected behavior and prevent regressions.

---

### 4. `"".join(text_buffer)` in the delimiter-mode flush — paragraph boundaries lost

In the new `flush_text_buffer`:
```python
combined_text = "".join(text_buffer)
```
Consecutive text chunks from structured JSON are concatenated without any separator. If the original document had paragraph gaps represented as separate text chunks, those boundaries are silently erased before re-splitting by delimiter pattern.

**Suggestion**: Use `"\n".join(text_buffer)` or the original separator between text chunks.

---

### 5. Redundant branch in non-JSON path

For the `text/markdown/html` path:
```python
if self._param.delimiter_mode == "delimiter":
    cks = _split_text_by_pattern(payload, delimiter_pattern)
elif delimiter_pattern:
    cks = _split_text_by_pattern(payload, delimiter_pattern)
```
Both branches do the same thing. The `delimiter_mode == "delimiter"` check has no effect for non-JSON input.

---

Please address items 1–3 before merging — they affect correctness for all naive-parser callers, not just the new delimiter pipeline. Items 4–5 are nice-to-fix.

Happy to re-review after updates!

PR #17724, opened 2026-08-03
Fix: set chunk_num before set_progress, avoid race condition

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the fix! The race condition is real — the async `_sync_progress` loop can observe `DONE` with stale `chunk_num=0` if `increment_chunk_num` runs after `set_progress(prog=1.0)`. Moving it before `set_progress` is the right approach.

A minor observation in `task_executor.py` (line ~939): the line `get_recording_context().save_func_return_value("DocumentService.increment_chunk_num", ret)` in the legacy executor now runs unconditionally outside the try-block, which is fine — `ret` will be `None` on failure. This matches the refactored executor behavior where the `intercept` path just records the call without a return value. ✅

LGTM, ship it.

PR #17734, opened 2026-08-03
fix: fix the filter error in dataset page

dcc123456:
### Summary

fix: fix the filter error in dataset page

### DIFF
--- a/web/src/components/list-filter-bar/index.tsx
+++ b/web/src/components/list-filter-bar/index.tsx
@@ -78,7 +78,7 @@ export default function ListFilterBar({
             return (
               pre +
               Object.values(cur).reduce((pre, cur) => {
-                return pre + cur.length;
+                return pre + (cur?.length || 0);
               }, 0)
             );
           }


wangq8 (commented):
Thanks for the fix! This is a clean, minimal defensive fix.

**What the code does:** Inside `ListFilterBar`'s `filterCount` memo, when `value` is a nested object (e.g. filter groups), it counts total active filter items by summing `.length` on each inner value. The original code `pre + cur.length` would throw a `TypeError` if any inner value was null/undefined.

**The fix:** `cur?.length || 0` safely returns 0 when `cur` is nullish, while still correctly counting array lengths otherwise.

**Review:**
- ✅ Correct — optional chaining + `|| 0` is the right pattern here.
- ✅ No regressions — empty arrays still yield 0 (falsy `.length` → `0 || 0` = 0), non-empty arrays yield correct counts.
- ✅ Minimal scope — 1 line changed, 1 file.
- ✅ No security concerns, no missing test concerns (purely defensive null-guard in a UI component).

LGTM 🚀

PR #17736, opened 2026-08-03
Fix: The agent version page exported an incorrect DSL format.

cike8899:
### Summary

Fix: The agent version page exported an incorrect DSL format.


wangq8 (commented):
Nice catch and clean refactor. This PR fixes a real bug where the version-history dialog exported only `agent.dsl.graph` instead of the full canonical DSL.

### What was broken

The version dialog was doing:
```ts
downloadJsonFile(graph, agent?.title);
```

This was missing:
- **Sensitive field sanitization** (`clearSensitiveFields` — strips `api_key` etc.)
- **The `globals` wrapper** that the canvas export path includes
- **Non-graph DSL fields** (e.g., `components`, `globals`, `path`)
- **The `.json` file extension**

### What the fix does well

- Extracts `downloadDsl` as a shared utility → single source of truth for DSL export
- Moves `clear-sensitive-fields` from `hooks/` to `utils/` — more appropriate location
- Both the canvas export button and the version dialog now produce **identical output**
- Clean diff: mostly moves/renames + the new shared function

### Minor notes

- The `.test.ts` and `.ts` files in `utils/clear-sensitive-fields` show 0 additions / 0 deletions — these are pure renames (`hooks/` → `utils/`). This is fine but worth confirming the import paths inside those files are updated (they appear to be export-only utility files, so likely no internal import changes needed).

### Verdict

LGTM. The fix is correct, the refactor reduces duplication, and both export paths now emit the same (sanitized) DSL format. Approve.

PR #17737, opened 2026-08-03
fix: optimize dataflow indexing and logs

buua436:

Dataflow indexing performs a refresh and updates task checkpoints after every small batch, causing unnecessary Elasticsearch and MySQL overhead when processing many chunks.

Pipeline progress also records empty messages, while operation logs persist large embedding vectors that are not required for rerunning the pipeline and may exceed database packet limits.

This PR:

- increases the default document bulk size;
- disables per-batch refresh and refreshes the index once after insertion;
- reduces the frequency of task chunk-ID checkpoints;
- filters empty pipeline progress messages;
- removes `q_<dimension>_vec` fields from pipeline operation logs while preserving other runtime outputs needed for reruns.

### Type of change

### DIFF
--- a/api/db/services/pipeline_operation_log_service.py
+++ b/api/db/services/pipeline_operation_log_service.py
@@ -16,6 +16,7 @@
 import json
 import logging
 import os
+import re
 from datetime import datetime, timedelta
 
 from peewee import fn
@@ -49,6 +50,22 @@ _PIPELINE_TASK_TYPE_TO_FINISH_FIELD = {
     PipelineTaskType.SESSION_ESSENCE: "session_essence_task_finish_at",
     PipelineTaskType.STRUCTURE: "structure_task_finish_at",
 }
+
+_EMBEDDING_VECTOR_FIELD = re.compile(r"^q_\d+_vec$")
+
+
+def _remove_embedding_vectors(value):
+    """Remove index-only embedding vectors from a runtime pipeline snapshot."""
+    if isinstance(value, dict):
+        for key in list(value):
+            if _EMBEDDING_VECTOR_FIELD.fullmatch(str(key)):
+                del value[key]
+            else:
+                _remove_embedding_vectors(value[key])
+    elif isinstance(value, list):
+        for item in value:
+            _remove_embedding_vectors(item)
+    return value
 
 
 class PipelineOperationLogService(CommonService):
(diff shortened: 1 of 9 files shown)


wangq8 (commented):
Thanks for this PR — the optimizations make sense and the code is well-structured. A few observations and suggestions:

### What works well

- **Bulk size increase (4→32):** Reasonable default that cuts down Elasticsearch/OpenSearch bulk request overhead.
- **Per-batch refresh=False:** This is the right call for bulk ingestion. The old "wait_for" on every small batch was expensive.
- **Checkpoint throttling:** Reducing MySQL update_task_chunk_ids calls from every batch to every ~256 chunks is a good trade-off — still resumable, much less DB churn.
- **Empty progress filter in pipeline.py:** Clean fix; avoids polluting the task log with timestamp-only entries.
- **Embedding vector stripping in operation logs:** The `_remove_embedding_vectors` implementation is clean and handles nested dicts/lists correctly. The unit test covers the expected shape well.

### Items worth attention

1. **refresh_idx availability:** The final refresh_idx call in _insert_main_chunks only actually works on OpenSearch (the only connector that implements it). ES, Infinity, and OB don't have this method. The getattr + callable guard means no crash, but for ES specifically, documents won't be immediately searchable after the bulk insert — they'll appear after the index's default refresh_interval (~1s). This is fine for dataflow indexing (eventual consistency is acceptable), but worth a comment.

2. **Infinity / OB insert signature:** These two connectors now accept `refresh` in their signatures but don't actually use the parameter in their method bodies. This is fine for interface consistency, but consider adding a brief comment noting the parameter is accepted for compatibility.

3. **refresh_idx only in _insert_main_chunks:** _insert_mother_chunks also passes refresh=False but doesn't get a matching final refresh_idx call. If mother chunks are expected to be immediately queryable after insertion, this could be a gap. Please confirm this is intentional.

4. **Memory pressure from DOC_BULK_SIZE=32:** For deployments using large embedding dimensions, 32 chunks per batch could increase memory pressure during bulk insert. Consider whether this should be a per-deployment tuning knob rather than a hardcoded default — or at least documented as adjustable via DOC_BULK_SIZE env var.

5. **Edge case — DOC_BULK_SIZE=0 or negative:** The formula `checkpoint_batches = max(1, 256 // doc_bulk_size)` would raise ZeroDivisionError if someone set DOC_BULK_SIZE=0. Consider a guard or clamping in init_settings.

### Summary

The PR is solid. The core optimizations (batch refresh, checkpoint throttling, empty-message filtering, vector stripping) are all well-reasoned and correctly implemented. The items above are minor — the main one to consider is the refresh_idx gap for non-OpenSearch backends. Approving with these suggestions.

PR #17739, opened 2026-08-03
fix(chunker): JSON path merges globally and keeps over-budget items whole

xugangqiang:
## Summary

Fixes two TokenChunker **json-path** over-segmentation bugs that diverge from Python's `rag/app` chunkers (tracked as `go_bug` known-diffs).

### 1. token-json-per-item-merge
Python's `_merge_text_chunks_by_token_size` merges adjacent text chunks **across JSON items** into one global token budget. Go merged within each item separately, emitting one chunk per item (2 items that jointly fit the budget → 2 chunks instead of 1). `invokeJSONPayload` now flattens the per-item structure into a single sequence before `mergeByTokenSizeFromJSON`, so a multi-item document is merged globally like Python.

### 2. token-json-single-item-oversize
Python's json path keeps a single item that exceeds `chunk_token_size` **whole**; Go sub-split it via `splitOversizedUnit`. `mergeByTokenSizeFromJSON` gains a `subSplitOversize` flag — the json path passes `false` (no sub-split), the text path passes `true` (keeps matching Python's `_split_oversized_unit`).

### Changes
- `token.go`: flatten-then-global-merge in `invokeJSONPayload`; `subSplitOversize` flag on `mergeByTokenSizeFromJSON`.
- Test call sites updated for the new signature (`true` for text-path simulations).

### Test plan
`bash build.sh --test ./internal/ingestion/component/chunker/...` — green, including the two new self-contained tests `TestJSONGlobalMergeMatchesPython` and `TestJSONSingleItemNotSubSplitMatchesPython` (drive `NewTokenChunker` directly, no `.venv`, default unit tier).

## Note
Follow-up: the `known_diffs.json` rules `token-json-per-item-merge` and `token-json-single-item-oversize` (in the parity-infra PR #17735) should be removed once this lands. This PR is independent of #17729 / #17735; based on `upstream/main`.

### DIFF
--- a/internal/ingestion/component/chunker/json_global_merge_test.go
+++ b/internal/ingestion/component/chunker/json_global_merge_test.go
@@ -12,8 +12,8 @@
 //  and to permit persons to whom the Software is furnished to do so,
 //  subject to the following conditions:
 //
-//  The above copyright notice and this permission notice shall be
-//  included in all copies or substantial portions of the Software.
+//  The above copyright notice and this permission notice shall be included
+//  in all copies or substantial portions of the Software.
 //
 //  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 //  EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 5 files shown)


yuzhichang (approved):
LGTM

PR #17740, opened 2026-08-03
test(chunker): self-contained guard for splitOversizedUnit running-sum flush

xugangqiang:
## Summary

Adds a self-contained regression guard for `splitOversizedUnitWith` after PR #17729
aligned it with Python's `rag/nlp._split_oversized_unit` running-sum flush.

#17729 shipped a `slack=1` relaxation in `token_strict_cap_test.go` (the oversized
unit is now sub-split with the same running-sum flush Python uses, which can leave a
piece one token over the nominal budget due to cl100k non-additivity). This PR adds
the missing positive proof that the sub-split boundaries are correct, so the relaxed
assertion is no longer unguarded.

## Changes

- `split_oversized_guard_test.go` (new, self-contained — no harness loader, no
  `testdata`; the Python oracle is inlined):
  - `TestSplitOversizedUnitRunningSumMatchesPython`: asserts the exact piece
    boundaries (live tokenizer) match Python's `_split_oversized_unit`, compensating
    the `slack=1` relaxation from #17729.
  - `TestSplitOversizedUnitDeadTokenizerCollapses`: asserts a zero-counting tokenizer
    collapses the B1 paragraph into exactly one chunk, catching a silently dead encoder
    that a non-empty-result check would miss.

## Notes

- Test function names are deliberately distinct from PR #17735's
  `TestSplitOversizedUnitMatchesPython`, so the two PRs verify independently and do not
  conflict at merge time.
- PR #17735 (golden parity harness) is intentionally left unchanged per the split plan.

## Test plan

`bash build.sh --test ./internal/ingestion/component/chunker/` — green, including both
new tests.

### DIFF
--- a/internal/ingestion/component/chunker/split_oversized_guard_test.go
+++ b/internal/ingestion/component/chunker/split_oversized_guard_test.go
@@ -1,0 +1,82 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


xugangqiang:
@yuzhichang  Pls help to review.

yuzhichang:
LGTM

PR #17742, opened 2026-08-03
fix(web): collapse overflowing tags on agent cards

euvre:
### Summary

On the Agents page, every tag of an agent was rendered with `flex flex-wrap` inside the card. When an agent had many tags, the tags wrapped into multiple lines and stretched its card, forcing all other cards in the same grid row to the same height and breaking the layout.

This PR limits tags to a single row on the agent card:

- A hidden measurement row computes how many tags actually fit the card width; only those are rendered.
- Overflowing tags collapse into a `...` badge, and hovering it reveals the full tag list in a `HoverCard` (same interaction as the SaaS version).
- A `ResizeObserver` recalculates the visible tag count whenever the card width changes (window resize / grid reflow).
- Measurement runs in `useLayoutEffect` before paint, so there is no flicker.

Only `web/src/pages/agents/agent-card.tsx` is touched.

### DIFF
--- a/web/src/pages/agents/agent-card.tsx
+++ b/web/src/pages/agents/agent-card.tsx
@@ -3,9 +3,15 @@ import { MoreButton } from '@/components/more-button';
 import { SharedBadge } from '@/components/shared-badge';
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
+import {
+  HoverCard,
+  HoverCardContent,
+  HoverCardTrigger,
+} from '@/components/ui/hover-card';
 import { AgentCategory } from '@/constants/agent';
 import { useNavigatePage } from '@/hooks/logic-hooks/navigate-hooks';
 import { AgentListItemType, IFlow } from '@/interfaces/database/agent';
+import { useLayoutEffect, useRef, useState } from 'react';
 import { CanvasCategoryToFlowType, FlowType, FlowTypeConfig } from './constant';
 import { AgentDropdown } from './agent-dropdown';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this PR! The approach of limiting tags to a single row with an ellipsis HoverCard is a clean solution that aligns well with the SaaS version.

**What works well:**
- The hidden measurement row + `ResizeObserver` strategy is robust and avoids flicker by using `useLayoutEffect`.
- The HoverCard fallback for overflow tags provides good UX parity with the enterprise/SaaS version.
- Only one file changed, tightly scoped — great.

**Suggestions (non-blocking):**

1. **Hardcoded gap constant**: The JS-side `gap = 4` assumes CSS `gap-1` = 4px. If the Tailwind config ever remaps the spacing scale, this could subtly desync. Consider extracting `getComputedStyle(container).gap` at measurement time, or adding a code comment noting the dependency.
2. **Zero-width container edge case**: If the card width collapses to near-zero, `visibleCount` could reach 0 and the card renders only `...`. This is correct and better than the old wrapping behavior, but worth being aware of if anyone ever sees an agent card showing only `...` at extremely narrow viewports.
3. **Missing test**: A basic unit test for `AgentTags` verifying that tags beyond the measured width are hidden and the HoverCard is rendered would be great for future regression protection. Optional given the visual nature of the change.

**Verdict:** LGTM — the fix is correct, well-structured, and addresses the layout regression cleanly. No blocking concerns.

dcc123456 (commented):
This issue is fixed

PR #17744, opened 2026-08-03
fix(chunker): align custom-delimiter text/json paths with Python

xugangqiang:
## Summary

Fixes the `go_bug` custom-delimiter newline divergences tracked in `known_diffs.json` (these rules live in the parity-infra PR #17735; this PR is the production fix only):

- `token-custom-delim-newline-and-field` — text / markdown / html paths with a backtick delimiter kept the delimiter glued to the preceding segment. Now the text path **drops** the delimiter and `.strip()`s each segment, mirroring Python's `_split_text_by_pattern` + `.strip()` (`rag/flow/chunker/token_chunker.py:316-338`). A chunk now reads `first sentence here` instead of `first sentence here` + trailing newline.
- `token-json-custom-delim-drops-newline` — the json path ran every chunk through `strings.TrimSpace`, which deleted the delimiter's trailing newline that Python keeps (`_build_json_chunks` / `_finalize_json_chunks` never strip). The json path no longer trims, so `first segment line one` + trailing newline survives like Python.

### `doc_type_kwd` is intentionally kept

The divergence rule also listed `doc_type_kwd` as a Go-only field. It is **not** a defect: `doc_type_kwd` is a load-bearing Go field (Infinity index column + media/PDF dispatch — `engine/infinity/chunk.go:741`, `*_vision_dispatch.go`, `media_dispatch.go`, `parser.go`). Python's JSON path emits it too (`token_chunker.py:125`); only its text path omits it. It is reclassified as `go_intentional` (extra_fields) in `known_diffs.json` (part of #17735), **not** removed here.

### Changes

- `common.go`: add `splitDroppingDelim` (delimiter discarded, mirrors Python).
- `token.go`: text path uses `splitDroppingDelim` + per-segment `TrimSpace`; json path drops the `TrimSpace` in `invokeJSONPayload`.
- Updated 3 existing unit tests that encoded the old keep-delimiter behaviour.
- Added `custom_delim_test.go`: 4 self-contained RED→GREEN tests reproducing the parity cases (`token__text/markdown/html/json_backtick`); they also assert `doc_type_kwd:"text"` stays present.

### Test plan

`bash build.sh --test ./internal/ingestion/component/chunker/` — green, including the 4 new tests and no regression to the existing suite.

## Note

Independent of #17729 / #17735 / #17739; based on `upstream/main`. The `known_diffs.json` rules `token-custom-delim-newline-and-field` and `token-json-custom-delim-drops-newline` should be removed/reclassified once this lands (handled in #17735).

### DIFF
--- a/internal/ingestion/component/chunker/common.go
+++ b/internal/ingestion/component/chunker/common.go
@@ -118,6 +118,38 @@ func splitKeepingDelim(text string, pattern *regexp.Regexp) []string {
 	return out
 }
 
+// splitDroppingDelim mirrors Python's _split_text_by_pattern
+// (token_chunker.py:79-90). Unlike splitKeepingDelim, the captured delimiter
+// is DISCARDED rather than glued to the preceding segment: re.split with a
+// captured group keeps delimiters at odd indices, and only the even-index
+// (text) parts are kept. This is the behaviour the text/markdown/html path
+// must reproduce so a split chunk reads "first sentence here" without the
+// trailing delimiter.
+func splitDroppingDelim(text string, pattern *regexp.Regexp) []string {
+	if pattern == nil {
+		return []string{text}
+	}
+	idxs := pattern.FindAllStringIndex(text, -1)
+	if len(idxs) == 0 {
+		return []string{text}
+	}
+	var out []string
+	cursor := 0
+	for _, idx := range idxs {
+		start, end := idx[0], idx[1]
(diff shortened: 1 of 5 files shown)


xugangqiang:
Superseded by the combined Go PR #17835 (delimiter boundary + OVER_CAP merge sync). Closing in favor of #17835.

PR #17745, opened 2026-08-03
Fix: handle llm_setting switch

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thank you for this fix! The separation of "enabled" state from value existence is a clean improvement. I reviewed the changes and have a few observations:

## ✅ What looks good

**Core logic (`resolve_llm_setting`):** The function correctly resolves each parameter based on its `_enabled` flag, falling back to defaults when the flag is false or the value is missing. Non-generation keys (like `llm_id`, `model_type`) are properly carried through, and `_enabled` flags are stripped before forwarding to the LLM — exactly what's needed.

**Backward compatibility:** Using `llm_setting?.temperature_enabled ?? true` in the frontend ensures existing datasets without `_enabled` flags default to enabled, matching the previous behavior where "value exists = enabled." The `_enabled` flags are now correctly persisted on save, which was missing before.

**Defaults consistency:** `LLM_SETTING_DEFAULTS` in `_generation_params.py` (`temperature: 0.1, top_p: 0.3, frequency_penalty: 0.7, presence_penalty: 0.4`) matches the previously commented-out defaults now activated in `db_models.py`. Good alignment.

**Test adjustments:** The test stubbing for `resolve_llm_setting` and `__path__` additions are reasonable to keep the unit test isolated.

## ⚠️ Suggestions

**1. Formatting-only changes mixed with functional changes**

`download-dsl.ts` (collapsing function signature) and `general-form.tsx` (expanding `CompilationTemplateFormField` props) appear to be pure formatting. Mixing them with the functional fix makes the diff larger and slightly harder to review. Consider reverting these or moving them to a follow-up cleanup PR.

**2. `max_tokens` asymmetry**

`GENERATION_CONFIG_KEYS` includes `max_tokens`, but `LLM_SETTING_DEFAULTS` and `resolve_llm_setting` do not handle it. This is likely intentional since `max_tokens` doesn't have an enable/disable switch in the UI, but it means `max_tokens` values within `search_config.llm_setting` are always passed through as-is (via the extra-keys loop). This is fine for now, but worth keeping in mind if a `max_tokens_enabled` switch is ever added — the function will need updating.

**3. Consider deduplicating the defaults**

`LLM_SETTING_DEFAULTS` and the `Search` model's `llm_setting` default in `db_models.py` now encode the same numeric defaults. If these ever need to change, both locations must be updated. Consider importing `LLM_SETTING_DEFAULTS` into `db_models.py` to keep a single source of truth.

## Verdict

The fix is correct and well-scoped. The two call sites (`related_questions_embedded` in `bot_api.py` and `recommendation` in `chat_api.py`) are the right places to apply `resolve_llm_setting` — these are the endpoints that read `search_config.llm_setting` directly for LLM calls. The main chat completion flow uses `dialog.llm_setting` (a separate entity), so it's unaffected.

Approved with minor suggestions above — none are blocking.

PR #17746, opened 2026-08-03
feat(message-input): add custom drag-to-resize handle at top border

dcc123456:
### Summary

feat(message-input): add custom drag-to-resize handle at top border

Replace the native browser textarea resize handle with a custom drag
  handle positioned at the top border of the chat input form. When
  `resize="vertical"` is passed to `NextMessageInput`, a pill-shaped
  grip button renders at the top edge instead of the native
  bottom-right corner handle.

### DIFF
--- a/web/src/components/message-input/next.tsx
+++ b/web/src/components/message-input/next.tsx
@@ -16,7 +16,7 @@ import { Button } from '@/components/ui/button';
 import { Textarea } from '@/components/ui/textarea';
 import { CircleStop, Globe, Paperclip, Send, Upload, X } from 'lucide-react';
 import * as React from 'react';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 import storage from '@/utils/authorization-util';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thank you for this PR! The custom drag-to-resize handle at the top border is a nice UX improvement. Here's my review:

## What Works Well ✅
- The drag logic correctly constrains height between 40–400px
- `autoSize` is properly disabled after manual resize, preventing conflicts between auto-sizing and manual height
- Event listener cleanup in the `useEffect` return is correct — no leaks
- Body `cursor` and `userSelect` are properly set during drag and restored on cleanup
- The native `resize` CSS property is set to `none` when the custom handle is active, avoiding double resize handles

## Suggestions

### 1. Magic numbers — consider extracting constants
The height bounds `40` and `400` appear in both the JS (`Math.max(40, Math.min(..., 400))`) and the CSS (`max-h-[400px]`). Extracting them into a shared constant would prevent them from drifting apart:
```ts
const MIN_RESIZE_HEIGHT = 40;
const MAX_RESIZE_HEIGHT = 400;
```

### 2. `isManualMode` never resets back to `false`
Once the user drags to resize, `autoSize` is permanently disabled for the lifetime of that component instance. This means:
- If the user drags the textarea to be *very* small, then types a multi-line message, it won't auto-expand
- The only way to get auto-size back is to re-mount the component (e.g., navigate away and back)

Consider whether this is the intended UX. A possible enhancement: provide a way to reset to auto-size (e.g., double-click the handle), or let `autoSize` resume when content scrollHeight exceeds the manual height.

### 3. Missing tests
The resize behavior would benefit from unit or integration tests covering:
- Handle renders when `resize="vertical"`
- Handle does not render when `resize="none"` (default)
- Drag mousemove correctly updates textarea height
- Height is clamped within 40–400px range
- `autoSize` is disabled after first manual resize

### 4. Minor: Dropzone z-index interaction
The `FileUploadDropzone` is positioned absolute over the form when dragging files. Since the resize handle sits at `-top-0.5` (above the form border), it's outside the dropzone overlay area — but worth a quick manual check that the handle remains clickable when the dropzone overlay is active.

## Verdict
Functionally sound and well-structured. The changes are focused and the implementation is clean. The suggestions above are non-blocking quality-of-life improvements. Approving from my side.

Thanks for the contribution! 🚀

PR #17747, opened 2026-08-03
fix(web): file list total count doesn't match displayed rows

euvre:
### Summary

In the file manager, the pagination total did not match the number of rows actually displayed (e.g. showing "5 total" while only 4 rows were rendered).

The backend counts the `skills` folder in the file list total (it is auto-initialized under the root folder), but the frontend filters the `skills` folder out of the table when the skills feature is disabled (non-go/hybrid builds). The pagination component and table `rowCount` still used the raw server-side total, causing the mismatch.

This PR tracks the number of client-side hidden rows when filtering and subtracts it from the total used by the table `rowCount` and the pagination component, keeping the displayed total consistent with the rendered rows. Go/hybrid builds are unaffected (no rows are hidden there).

### DIFF
--- a/web/src/pages/files/files-table.tsx
+++ b/web/src/pages/files/files-table.tsx
@@ -114,8 +114,8 @@ export function FilesTable({
 
   // Sort files with skills folder first, then by time
   // Filter out skills folder if not in hybrid/go mode
-  const sortedFiles = useMemo(() => {
-    if (!files) return [];
+  const { sortedFiles, hiddenCount } = useMemo(() => {
+    if (!files) return { sortedFiles: [] as IFile[], hiddenCount: 0 };
 
     // Filter out skills folder if feature is disabled
     const filteredFiles = isSkillsEnabled
@@ -126,7 +126,7 @@ export function FilesTable({
           return !isSkills;
         });
 
-    return [...filteredFiles].sort((a, b) => {
+    const sorted = [...filteredFiles].sort((a, b) => {
       const aIsSkills =
         isFolderType(a.type) && a.name.toLowerCase() === 'skills';
       const bIsSkills =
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix, @euvre! The root cause and solution are well-reasoned. Here are my detailed thoughts:

---

**✅ Correctness**
The fix is spot-on. Tracking `hiddenCount` as the delta between the raw file list and what gets rendered, then subtracting it from the server-reported `total` provides a `displayTotal` that accurately matches visible rows. The `Math.max(..., 0)` guard prevents any negative counts, which is a good defensive choice.

**✅ No regressions**
- When `isSkillsEnabled` is `true`, `hiddenCount` is `0`, so `displayTotal === total` — behaviour is unchanged for hybrid/go builds.
- The destructuring change from `sortedFiles` (array) to `{ sortedFiles, hiddenCount }` (object) is correctly reflected at all usage sites (`table.setOptions`, pagination, etc.), and the `as IFile[]` type assertion on the empty array keeps TypeScript happy.

**✅ Code clarity**
The comment above `displayTotal` clearly explains *why* this adjustment is needed (server-side total includes the skills folder, which the client filters out). This will save future maintainers time.

**🟡 Suggestion (non-blocking)**
The `useMemo` dependency array is `[files, isSkillsEnabled]`. Since `isSkillsEnabled` is derived from `__API_PROXY_SCHEME__` (a compile-time replacement), it will never change at runtime. This is fine — just noting that the dependency is effectively constant. If the intent is to leave room for a future dynamic toggle, the current code is already well-positioned for that.

**🟡 Minor: edge case**
If the server ever returns *fewer* files than the skills-filter count (unlikely but theoretically possible if the only file in a folder is the skills folder itself), `displayTotal` would be `0`, which is still correct — the table would show "0 total" matching "0 rows". This is handled fine.

Overall this is a clean, minimal fix. No tests were included, which seems reasonable given it is a straightforward display-fidelity change with no new logic branches.

LGTM! 👍

dcc123456 (approved):

PR #17748, opened 2026-08-03
fix(chunker): preserve inter-line trailing space in text-path overlap prefix

xugangqiang:
## Summary
- Fixes the `token-text-overlap-split` parity divergence on the text path of `TokenChunker`.
- Root cause: `mergeByTokenSize` applied `strings.TrimSpace` to each sentence-delimiter-split fragment, dropping the trailing whitespace of each line. The overlap prefix is carved from the *untrimmed* previous chunk, so that dropped character made every following chunk head diverge from Python's `naive_merge` by one character (e.g. Go `"eta..."` vs Python `"ta ..."`, and the chunk boundary `"alpha\nbeta"` vs `"alpha \nbeta"`).
- Fix: mirror `naive_merge`, which builds each unit from `"\n" + sub_sec` (sub_sec keeps its trailing space) and only strips at the final chunk output. Skip only genuinely empty fragments instead of TrimSpacing each part.

## Verification
- On the `token__text_overlap` case (`chunk_token_size=64`, `delimiters=["\n"]`, `overlapped_percent=0.1`), Go output now matches Python `rag/nlp.naive_merge` **byte-for-byte** across all 3 chunks (lengths 239/224/263).
- Added `TestTokenChunker_TextOverlapPreservesInterLineSpace`, which fails without the fix and passes with it, and pins the exact chunk texts against the Python reference.
- Full `./internal/ingestion/component/chunker/` unit suite passes (no regression).

## Notes
- Independent of #17744 (custom-delimiter text/json alignment); based on `upstream/main`.
- Section-level `strings.TrimSpace(sec)` is intentionally left in place — it only affects global leading/trailing whitespace, which is stripped at final output anyway and causes no divergence.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -560,7 +560,18 @@ func (c *TokenChunkerComponent) mergeByTokenSize(text string, childrenPattern *r
 		parts := sentenceDelimiter.Split(sec, -1)
 		hadPart := false
 		for _, part := range parts {
-			part = strings.TrimSpace(part)
+			// Keep the raw split fragment, including any inter-line trailing
+			// whitespace. Python's naive_merge builds each unit from
+			// "\n" + sub_sec (naive_merge:1357) where sub_sec retains its
+			// trailing space and is never TrimSpaced — the only post-processing
+			// is dropping the leading empty placeholder (naive_merge:1370-1375),
+			// never per-unit trimming. Trimming here drops that space, so the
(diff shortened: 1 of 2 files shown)


xugangqiang:
@yuzhichang @JinHai-CN @yingfeng  Please help to review

yuzhichang (approved):

xugangqiang:
Superseded by the combined Go PR #17835 (inter-line trailing space + OVER_CAP merge sync). Closing in favor of #17835.

PR #17753, opened 2026-08-03
fix(storage): fail fast instead of storing plaintext when crypto config is broken

### LINKED ISSUE #17752: [Bug]: RAGFLOW_CRYPTO_ENABLED=true silently falls back to unencrypted storage when the crypto config is broken

- [X] I have checked the existing issues.

### Describe the bug

When `RAGFLOW_CRYPTO_ENABLED=true` but the crypto configuration is broken, `init_settings()` catches the error, logs it, and falls back to the **unencrypted** storage impl. The server starts normally and writes every document to object storage in the clear.

https://github.com/infiniflow/ragflow/blob/main/common/settings.py#L356-L369

```python
if crypto_enabled:
    try:
        from rag.utils.encrypted_storage import create_encrypted_storage
        algorithm = os.environ.get("RAGFLOW_CRYPTO_ALGORITHM", "aes-256-cbc")
        crypto_key = os.environ.get("RAGFLOW_CRYPTO_KEY")
        STORAGE_IMPL = create_encrypted_storage(storage_impl, algorithm=algorithm, key=crypto_key, encryption_enabled=crypto_enabled)
    except Exception as e:
        logging.error(f"Failed to initialize encrypted storage: {e}")
        STORAGE_IMPL = storage_impl   # <-- plaintext
```

The operator explicitly asked for encryption at rest, and the outcome is the exact opposite, behind a single log line.

### Why this is hard to notice

Reads keep working, so nothing downstream ever fails. `BaseCrypto.decrypt()` returns data that lacks the `RAGF` magic header unchanged:

https://github.com/infiniflow/ragflow/blob/main/common/crypto_utils.py#L99-L102

That passthrough is deliberate and useful — it lets you enable encryption over an existing bucket. But combined with the fallback it means a plaintext object stays readable **even after the configuration is fixed**, so the leak never surfaces:

```
on-disk:
   leaked.txt   encrypted=False  b'SSN 123-45-6789'
   safe.txt     encrypted=True  b"RAGF\xbf\x8c9\xa6\xd4\xef!\x1a\x1f=\xb7\x04OF'V\x00WR\x93"
reads AFTER the config is fixed:
   get(leaked.txt  ) -> b'SSN 123-45-6789'
   get(safe.txt    ) -> b'SSN 987-65-4321'
```

There is no health check, no startup banner and no metric that distinguishes "encrypting" from "silently not encrypting".

### Which misconfigurations trigger it

All three are one-character mistakes in `docker/.env`, where the variables ship commented out:

| config | result | on-disk |
| --- | --- | --- |
| `ENABLED=true`, `KEY=my-key` | wrapped | encrypted |
| `ENABLED=true`, `KEY` line still commented out | **falls back** | **plaintext** |
| `ENABLED=true`, `KEY=my-key`, `ALGORITHM=aes256-cbc` (missing hyphen) | **falls back** | **plaintext** |
| `ENABLED=true`, `KEY=` (empty) | **falls back** | **plaintext** |

Measured output:

```
correct config               wrapper=EncryptedStorageWrapper  on-disk-encrypted=True   first40=b'RAGF\xea\xfc\x9c\xb5c4\x99...'
KEY unset (typo in .env)     wrapper=FakeStorage              on-disk-encrypted=False  first40=b'PATIENT SSN 123-45-6789 CONFIDENTIAL'
ALGORITHM typo               wrapper=FakeStorage              on-disk-encrypted=False  first40=b'PATIENT SSN 123-45-6789 CONFIDENTIAL'
KEY empty string             wrapper=FakeStorage              on-disk-encrypted=False  first40=b'PATIENT SSN 123-45-6789 CONFIDENTIAL'
```

### Two related defects in the same helper

While reproducing this I found that `create_encrypted_storage`'s documented defaults do not work:

https://github.com/infiniflow/ragflow/blob/main/rag/utils/encrypted_storage.py#L248-L261

```
signature: (storage_impl, algorithm=None, key=None, encryption_enabled=True)
documented default RAISES: ValueError Unsupported algorithm: None
env-var key fallback RAISES: ValueError Encryption key not provided and RAGFLOW_CRYPTO_KEY environment variable not set
```

1. `algorithm=None` is passed straight to `CryptoUtil`, which rejects it with `Unsupported algorithm: None`. The default in the signature is unusable — every caller must pass an algorithm explicitly.
2. Both docstrings state the key "uses RAGFLOW_CRYPTO_KEY environment variable if None", and `CryptoUtil`'s error message says the same, but no code ever reads that variable. Only `init_settings` does.

So today the fallback in `settings.py` is the *only* thing keeping a `create_encrypted_storage(impl, key=...)` call from crashing on the algorithm default.

### Steps to reproduce

`common/crypto_utils.py` and `rag/utils/encrypted_storage.py` currently have no tests, so this uses a fake storage impl and replicates the `settings.py` branch verbatim:

```python
import os, logging
from rag.utils.encrypted_storage import create_encrypted_storage

class FakeStorage:
    def __init__(self): self.blobs = {}
    def put(self, b, f, binary, t=None): self.blobs[(b, f)] = binary; return "ok"
    def get(self, b, f, t=None): return self.blobs.get((b, f))
    def rm(self, b, f, t=None): pass
    def obj_exist(self, b, f, t=None): return (b, f) in self.blobs
    def health(self): return True

def init_storage():
    storage_impl = FakeStorage()
    if os.environ.get("RAGFLOW_CRYPTO_ENABLED", "false").lower() == "true":
        try:
            return create_encrypted_storage(
                storage_impl,
                algorithm=os.environ.get("RAGFLOW_CRYPTO_ALGORITHM", "aes-256-cbc"),
                key=os.environ.get("RAGFLOW_CRYPTO_KEY"),
                encryption_enabled=True,
            )
        except Exception as e:
            logging.error(f"Failed to initialize encrypted storage: {e}")
            return storage_impl
    return storage_impl

os.environ["RAGFLOW_CRYPTO_ENABLED"] = "true"   # and no RAGFLOW_CRYPTO_KEY
storage = init_storage()
storage.put("bucket", "doc.txt", b"PATIENT SSN 123-45-6789")
print(storage.blobs[("bucket", "doc.txt")])     # b'PATIENT SSN 123-45-6789'  <-- plaintext
```

### Expected behavior

`RAGFLOW_CRYPTO_ENABLED=true` with a broken crypto config should abort startup rather than write plaintext, the same way an unsupported doc engine already does at `common/settings.py:325` (`raise Exception(f"Not supported doc engine: {DOC_ENGINE}")`). Failing to start is loud and recoverable; silently storing unencrypted documents is neither.

`create_encrypted_storage` should also resolve the defaults its own signature and docstrings promise.

I have a fix and a test suite for `crypto_utils` / `encrypted_storage` ready, verified red on the current tree — happy to open the PR.

### Environment

- RAGFlow version: `main` @ 594a464
- OS: any (config-only path)

### Additional context

`RAGFLOW_CRYPTO_*` appears only in `docker/.env` (commented out), so the feature is opt-in and the blast radius is limited to deployments that turned it on — which are exactly the ones that care about encryption at rest.

LHMQ878:
### Summary

Fixes #17752.

`RAGFLOW_CRYPTO_ENABLED=true` is a request for encryption at rest. Today, if the crypto configuration is broken, `init_settings()` logs one line and assigns the **unencrypted** storage impl, so documents are written in the clear while the deployment looks healthy. This PR removes that fallback, fixes the two defaults in `create_encrypted_storage` that made the broken-config case easy to hit, and adds the first tests for `common/crypto_utils.py` and `rag/utils/encrypted_storage.py`.

### Root cause

`common/settings.py:356-369`:

```python
    # Check if encryption is enabled
    if crypto_enabled:
        try:
            from rag.utils.encrypted_storage import create_encrypted_storage

            algorithm = os.environ.get("RAGFLOW_CRYPTO_ALGORITHM", "aes-256-cbc")
            crypto_key = os.environ.get("RAGFLOW_CRYPTO_KEY")

            STORAGE_IMPL = create_encrypted_storage(storage_impl, algorithm=algorithm, key=crypto_key, encryption_enabled=crypto_enabled)
        except Exception as e:
            logging.error(f"Failed to initialize encrypted storage: {e}")
            STORAGE_IMPL = storage_impl
```

Any failure inside `create_encrypted_storage` lands in `except Exception` and `STORAGE_IMPL` becomes the raw, unencrypted impl. Measured by running that exact branch against an in-memory storage impl and storing `b'PATIENT SSN 123-45-6789 CONFIDENTIAL'` (`RAGFLOW_CRYPTO_ENABLED=true` in every row):

```
scenario                    wrapper                 on-disk encrypted
------------------------------------------------------------------------
KEY set, ALGORITHM default  EncryptedStorageWrapper True
KEY unset                   FakeStorage             False
KEY set, ALGORITHM typo     FakeStorage             False
KEY empty                   FakeStorage             False
```

Three of the four configurations that an operator can plausibly land on store the payload in plaintext. The only log line is `logging.error` at startup, which is easy to miss and is not repeated per write.

### Why it stays invisible

`CryptoUtil.decrypt()` returns data that lacks the `RAGF` magic header unchanged (`common/crypto_utils.py:99-102`). That passthrough is deliberate and useful — it is what lets you enable encryption over an existing bucket — but it also means a plaintext object keeps reading back correctly *after* the configuration is fixed:

```
leaked.txt   on-disk starts with RAGF: False  reads back correctly: True
safe.txt     on-disk starts with RAGF: True   reads back correctly: True
```

So nothing ever surfaces the leak. There is no failed read, no exception, and no difference visible to a caller — only the bytes on disk differ. That is why this needs to fail at startup rather than degrade.

### Fix

**1. `common/settings.py` — drop the fallback.** The surrounding function already fails fast for the analogous case, an unusable doc engine (`raise Exception(f"Not supported doc engine: {DOC_ENGINE}")`, line 325). Encryption-at-rest gets the same treatment, with a comment recording why silence is the wrong outcome here.

**2. `rag/utils/encrypted_storage.py` — resolve the documented defaults.** `create_encrypted_storage(storage_impl, algorithm=None, key=None, ...)` documents `algorithm` as "uses environment variable RAGFLOW_CRYPTO_ALGORITHM or default if None" and `key` as "uses environment variable RAGFLOW_CRYPTO_KEY if None". Neither was implemented — both were passed straight through to `CryptoUtil`:

```
documented default RAISES: ValueError: Unsupported algorithm: None
```

So the signature default was unusable and the key never fell back to the environment. Both are now resolved in the function, and explicit arguments still win over the environment. This matters for the bug above too: `settings.py` passes an explicit algorithm, so it never hit the `None` path, but any other caller relying on the documented defaults would raise — and under the old code that raise was swallowed into a plaintext fallback.

### Tests

New `test/unit_test/common/test_crypto_utils.py`, 33 tests in four groups. Neither `common/crypto_utils.py` (355 lines) nor `rag/utils/encrypted_storage.py` had any test coverage before this.

- `TestCryptoUtil` — round trip and magic header across all three supported algorithms (`aes-128-cbc`, `aes-256-cbc`, `sm4-cbc`), unsupported-algorithm and missing-key rejection, random IV, empty and block-aligned payloads, wrong key, deterministic key derivation, derived key length per cipher.
- `TestDecryptPassthrough` — pins the magic-header passthrough as intended behaviour, since the fix above depends on it being intended.
- `TestEncryptedStorageWrapper` — plaintext never reaches the backing impl, decryption on read, missing object, disabled-encryption passthrough, required-method check, and the resolved defaults.
- `TestInitSettingsCrypto` — the crypto branch itself: valid config wraps, disabled leaves the impl bare, each broken config raises, and `test_unencrypted_write_stays_invisible` pins the leak-invisibility result above.

`TestInitSettingsCrypto` mirrors the `RAGFLOW_CRYPTO_*` branch of `settings.py` in a helper rather than calling `init_settings()`, which also connects to MySQL, Redis and the doc store. That is a deliberate trade-off and it has a consequence worth stating: `test_enabled_with_broken_config_raises` passes both before and after this PR, because the helper reproduces the branch without the `try/except`. It documents the intended contract; it does not by itself catch a regression in `settings.py`. The measurement in the root-cause section above, not that test, is the evidence for fix 1.

Red check — the 33 tests against the unmodified sources at 594a464:

```
FAILED TestEncryptedStorageWrapper::test_put_encrypts_at_rest - ValueError: Unsupported algorithm: None
FAILED TestEncryptedStorageWrapper::test_get_decrypts - ValueError: Unsupported algorithm: None
FAILED TestEncryptedStorageWrapper::test_get_missing_object_returns_none - ValueError: Unsupported algorithm: None
FAILED TestEncryptedStorageWrapper::test_disabled_encryption_stores_plaintext - ValueError: Unsupported algorithm: None
FAILED TestEncryptedStorageWrapper::test_propagates_missing_key - AssertionError: Regex pattern did not match.
FAILED TestEncryptedStorageWrapper::test_defaults_to_aes_256_cbc - ValueError: Unsupported algorithm: None
FAILED TestEncryptedStorageWrapper::test_reads_config_from_environment - ValueError: Unsupported algorithm: None
7 failed, 26 passed in 0.88s
```

With the fixes applied: `33 passed in 0.69s`.

An autouse fixture clears `RAGFLOW_CRYPTO_ENABLED` / `_ALGORITHM` / `_KEY` for every test in the file, so an ambient value in a developer's shell cannot steer the results now that the environment fallback is actually read.

### Verification

- `test/unit_test/common/test_crypto_utils.py` — 33 passed.
- `test/unit_test/` before and after this branch: **19 failed / 324 passed → 12 failed / 331 passed**, 29 skipped. The 12 remaining failures are identical before and after and are environment-only — `test_time_utils.py` and `test_mistral_ocr_env.py`, on `dateutil`/`filelock` availability and the Windows epoch.
- `ruff check` and `ruff format --check` clean on all three files under the repo configuration.
- All 9 `lefthook` file hooks pass on the staged files.

Two disclosures:

- I could not run the full backend or Go suites locally (Windows, no MySQL/Redis/ES). Only `test/unit_test/` was executed.
- My local ruff (0.16.1) is newer than what CI pins and enables a much wider rule set, so it reports 33 pre-existing findings on these files untouched. I compared the baseline and edited findings rather than trusting the count: my change adds none and removes two (`BLE001`, `LOG015`). Under ruff's default selection plus the repo config, all three files are clean.

### Note on CI

`.github/workflows/tests.yml` gates `ragflow_preflight` on the `ci` label, so the checks on this PR will stay idle until a maintainer applies it.


LHMQ878:
Thanks — I probed all three before acting. One is a real bug and is now fixed; the other two I'd push back on.

### 2 — explicit empty values (valid, fixed in `5568c10`)

Confirmed by measurement. With `RAGFLOW_CRYPTO_KEY` set in the environment, `key=""` produced the *same derived key* as `key=None`:

```
env RAGFLOW_CRYPTO_KEY = '0123456789abcdef0123456789abcdef'
  key=None (documented fallback)   -> OK  key=b'V\xd53R\xf0\xe5\x01\xb0...'
  key="" (explicit empty)          -> OK  key=b'V\xd53R\xf0\xe5\x01\xb0...'   <- identical
  key='ZZZZ...' (explicit)         -> OK  key=b'x\xed\x1e&\xa1e\xdaH...'
```

Fixed as you suggested, guarding on `is None`:

```python
if algorithm is None:
    algorithm = os.environ.get("RAGFLOW_CRYPTO_ALGORITHM", "aes-256-cbc")
if key is None:
    key = os.environ.get("RAGFLOW_CRYPTO_KEY")
```

Two regression tests added, and A/B'd so they aren't vacuous — **2 failed** against the old `or` version, **35 passed** with the fix.

One correction to the severity, though: this is **not reachable from the production path**, so it isn't a live plaintext/wrong-key risk today. `common/settings.py` reads the same variable it passes, so argument and environment can never disagree:

```
RAGFLOW_CRYPTO_KEY=''                                -> ValueError: Encryption key not provided
RAGFLOW_CRYPTO_KEY='0123...abcdef'                   -> constructed OK
```

`CryptoUtil.__init__` already does `if not key: raise`, which treats `""` and `None` alike. So the empty-env case this PR is about was already failing fast correctly. Worth fixing anyway since `create_encrypted_storage` is public and the contract should hold for any caller — just not a Major security hole in the shipped code.

### 1 — log before validation (declining)

The log would sit one line above `create_encrypted_storage`, which already logs `"Encryption enabled in storage wrapper"` on success and raises with the algorithm name on failure (`Unsupported algorithm: bogus`). Adding a third message in between logs an *intent* that the very next line may invalidate — and on the failure path it means the last thing in the log is a success-shaped line, which is the opposite of what this PR is for. The guideline is "add logging for new flows"; this flow is logged at both outcomes.

### 3 — extract a helper from `init_settings` (declining, but the concern is fair)

You're right that `_init_storage` duplicating the branch means the test would still pass if someone restored the plaintext fallback. But extracting a private helper from `init_settings` is a refactor of a function that also wires MySQL, Redis and the doc store — out of scope for a fix PR, and it would make this diff much harder for a maintainer to review against the security claim.

Happy to do it as a follow-up if a maintainer wants it. If the duplication risk is the blocker, a cheaper option is asserting the source of `common/settings.py` contains no `STORAGE_IMPL = storage_impl` inside the crypto branch — ugly, but it actually pins the thing at risk. Let me know which you'd prefer.

JinHai-CN:
@yngvarhuang, You can also review this one.

PR #17755, opened 2026-08-04
Fix: Resolved issue where the default model could not be deleted.

dcc123456:
### Summary

Fix: Resolved issue where the default model could not be deleted.

### DIFF
--- a/web/src/components/dynamic-form.tsx
+++ b/web/src/components/dynamic-form.tsx
@@ -115,6 +115,24 @@ interface DynamicFormProps<T extends FieldValues> {
   //   updatedField: Partial<FormFieldConfig>,
   // ) => void;
   labelClassName?: string;
+  /**
+   * Options forwarded to `form.reset()` when `defaultValues` change.
+   * Pass `{ keepDirtyValues: true }` to preserve user edits during
+   * background refetches / lazy detail loads. Defaults to `{}`
+   * (hard reset) to preserve the original behavior for existing
+   * consumers.
+   */
+  resetOptions?: {
+    keepValues?: boolean;
+    keepDefaultValues?: boolean;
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for this PR! The fix is well-targeted and the `resetOptions` enhancement is a clean pattern. I reviewed the changes and have a few observations:

## ✅ What works well

**Core fix (`system-setting.tsx`):** Removing `if (!value) return;` is the right fix. When a user clears a default model (empty string), `parseModelValue("")` returns `null`, which correctly falls into the `else` branch and calls `setDefaultModel({ model_id: "", model_type: modelType })` — effectively clearing the default model. This was the root cause and the fix is minimal and correct.

**`resetOptions` prop (`dynamic-form.tsx`):** Clean, backwards-compatible addition. The prop is properly typed and threaded through to `form.reset()`. Good that it is included in the `useEffect` dependency array.

**Key simplification (`saved-mode-card.tsx`):** Removing `instanceDetailsLoaded` from the key is necessary — without it, the component would still remount when details load, defeating the purpose of `keepDirtyValues`. The form now stays mounted, and untouched fields still get populated by `form.reset()` while dirty (user-edited) fields are preserved.

## ⚠️ One thing to consider

**`resetOptions` as an effect dependency:** Since `{ keepDirtyValues: true }` is an inline object literal, it creates a new reference on every render, which will cause the `useEffect` in `DynamicForm.Root` to fire every render cycle in `draft-mode-card` and `saved-mode-card`. In practice this is fine since these forms do not re-render frequently, and the form state tracking should be stable. Just something to be aware of.

## Summary

The fix is correct, the approach is clean, and there are no regressions or security concerns. 👍

PR #17760, opened 2026-08-04
Fix: The dataset for the retrieval operator of the agent created from the template cannot be selected.

cike8899:
### Summary

Fix: The dataset for the retrieval operator of the agent created from the template cannot be selected.


wangq8 (commented):
Thanks for fixing this — the retrieval operator not being able to select datasets from templates was a real usability gap.

**What I checked:**

*Template changes:* Adding `"retrieval_from": "dataset"` to agent templates is the correct fix. Without this field, the retrieval operator doesn't know where to fetch from when the agent is created from a template. All 8 template files updated consistently.

*Web/store changes (`store.ts`):*
- `updateNodeForm` now uses immer-friendly `cloneDeep` before assigning into frozen state, decoupling from react-hook-form's mutable references. This is the right approach.
- `replaceNodeForm` keeps `cloneDeep` with an updated comment referencing `updateNodeForm`.

*Minor cleanup:*
- `next.tsx`: Extracting `openSandboxQuickstart` to a module-level constant is a nice micro-optimization to avoid re-creating the function on every render.
- `use-watch-change.ts`: `values.items?.slice() || []` → `values.items || []`. Since `updateNodeForm` now does `cloneDeep` internally, the explicit slice is no longer needed for frozen-array protection.

**One observation:** The `your_starter_dataset_chatbot.json` diff is mostly whitespace/formatting reformatting mixed with the `retrieval_from` addition. The functional change (adding `retrieval_from: "dataset"`) is visible, but the formatting noise makes the diff harder to review. In the future, consider separating reformatting into its own commit/PR.

Approved.

PR #17761, opened 2026-08-04
fix: normalize legacy parser configuration

buua436:

- Accept legacy RAPTOR parser configuration fields without validation errors.
- Migrate `raptor.ext.clustering_threshold` and ignore removed legacy fields.
- Normalize invalid RAPTOR `max_token` values to the supported minimum.
- Improve SiliconFlow embedding input handling and failure diagnostics.

### Type of change


wangq8 (commented):
Thank you for this PR. The changes address real issues with legacy parser config and embedding input handling. A few observations:

**Text truncation fix** (`llm_service.py`): Switching from raw string slicing (`text[:target_len]`) to `truncate()` is correct — it avoids splitting multi-byte UTF-8 characters. ✅

**Embedding truncation** (`embedding_model.py`): Adding `bge-m3` truncation to 4096 chars mirrors the existing BGE v1.5 pattern. SiliconFlow `bge-m3` has an 8192-token limit, so 4096 chars is conservative and appropriate. ✅

**Legacy RAPTOR normalization** (`validation_utils.py` + `document_api.py`): The approach is pragmatic but there is some redundancy worth noting:

1. Two `@model_validator(mode="before")` methods (`normalize_legacy_fields` and `ignore_legacy_fields`) both pop `threshold`, `clustering_method`, `tree_builder`, and `ext`. The second validator will be largely a no-op since the first already removed those fields. Consider merging into a single validator to simplify.

2. The `_normalize_legacy_raptor_config` function in `document_api.py` partially duplicates the model validators. Is this intentional as defense-in-depth, or could the model validator handle it alone? If it is intentional a brief comment explaining the layering would help future maintainers.

3. Minor: `_normalize_legacy_raptor_config` in `document_api.py` pops `ext` without migrating `ext.clustering_threshold` — the model validator does this, but the request-level function does not. This means the migration only happens at validation time, not in the raw dict. Is that the intended behavior?

No regressions or security concerns spotted. The `Base` model with `extra="forbid"` would reject legacy fields without these validators, and the `before` validators correctly strip them pre-validation. LGTM with the note about validator consolidation.

PR #17762, opened 2026-08-04
fix(py-agent): remove unavailable model defaults from templates

Hz-186:
## Summary

- Clear unavailable LLM defaults from built-in agent templates
- Avoid binding templates to unavailable provider models

## Testing

- `python3 -c 'import json, pathlib; [json.loads(p.read_text()) for p in pathlib.Path("agent/templates").glob("*.json")]'`
- `git diff --cached --check`


yuzhichang:
LGTM

PR #17763, opened 2026-08-04
Fix duplicate document names when linking files to a dataset

euvre:
### Summary

In file management, moving a file into a folder that already contains a same-named file is rejected (`400 Duplicated file name in the same folder.`), but linking a file to a knowledge base that already contains a same-named document was allowed, leaving two documents with the same name in the dataset.

The link endpoint (`POST /api/v1/files/link-to-datasets`) is fire-and-forget: `convertFiles` runs in a background goroutine per request. Its "list existing document names -> pick a unique name -> insert" sequence had no synchronization, so two concurrent link requests could interleave and both insert the same document name into the same KB.

This PR serializes background conversions with a mutex so the unique-name check and insert are atomic across concurrent requests, and adds regression tests covering:

- add mode: re-linking the same file does not create a second document
- add mode: linking a different file with a colliding name auto-renames (e.g. `laws(1).pdf`), mirroring Python `duplicate_name`
- replace mode: re-linking keeps exactly one document with the original name
- concurrent link requests never produce duplicate document names in the same KB

### DIFF
--- a/internal/service/document/file2document.go
+++ b/internal/service/document/file2document.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"ragflow/internal/service"
 	"strings"
+	"sync"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Good catch on a real concurrency bug, with solid tests.

**What's good:**
- The race condition is clearly explained: `LinkToDatasets` spawns a background goroutine per request, and the "list existing names → pick unique → insert" sequence has no synchronization — two concurrent requests can both see no conflict and both insert the same name.
- The `sync.Mutex` fix serializes the critical section correctly. The mutex is acquired before the loop and released after all files are processed, which keeps the window tight.
- Excellent test coverage: add-mode deduplication, name collision auto-rename (e.g. `laws(1).pdf`), replace-mode idempotency, and concurrent link stress test with 8 goroutines. The tests use a real SQLite-backed `setupServiceTestDB` which means they're testing against actual database semantics — much better than mocking.

**Potential improvement (non-blocking):**
The package-level `convertFilesMu` serializes all conversions globally. If high-throughput link operations become a bottleneck, a per-KB mutex (`map[string]*sync.Mutex` with its own locking) would allow parallel conversions across different KBs. That said, this is premature optimization — the global mutex is correct and simple, and the fire-and-forget goroutine pattern means the HTTP response isn't blocked anyway.

**One question:**
The existing `convertFiles` already checks `AlreadyLinked` to skip KBs in add mode. With the mutex, is the `AlreadyLinked` check still needed, or does the mutex make it redundant? If it's still needed for correctness (e.g., cross-process), the PR is fine as-is — just wanted to flag it.

No regressions found. Merge-ready.

wangq8:
@euvre - wrong fix.

euvre:
duplicate fix.

PR #17765, opened 2026-08-04
fix(file-manager): strip folder path from upload filename to prevent extra folder creation

dcc123456:
### Summary

fix(file-manager): strip folder path from upload filename to prevent extra folder creation

### DIFF
--- a/web/src/hooks/use-file-request.ts
+++ b/web/src/hooks/use-file-request.ts
@@ -62,7 +62,10 @@ export const useUploadFile = () => {
       const formData = new FormData();
       formData.append('parent_id', params.parentId);
       fileList.forEach((file: any, index: number) => {
-        formData.append('file', file);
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Clean one-line fix with good explanation.

The root cause is clear: browsers that support `webkitRelativePath` (from directory upload via `<input webkitdirectory>`) pass the full relative path as the filename in `FormData.append(name, blob)`, resulting in e.g. `folder/file.txt` being sent to the backend. The backend then interprets the `/` as a path separator and creates an extra folder.

Passing `file.name` as the explicit filename argument to `FormData.append(name, blob, filename)` forces the browser to use just the base name regardless of `webkitRelativePath`. This is the standard fix for this issue and mirrors how other upload implementations handle it.

Small, targeted, no side effects. Merge-ready.

PR #17767, opened 2026-08-04
fix: enable preview for legacy .doc documents in Go server

euvre:
### Summary

Legacy `.doc` (OLE2) documents could not be previewed in the web UI: the Word previewer (`@extend-ai/react-docx`) only supports modern `.docx` (ZIP) files, so `.doc` uploads failed with "Preview is not available for this Word document". Both the Python and Go backends returned the raw `.doc` bytes from the document preview endpoint without any conversion.

This PR fixes the Go server preview path by extracting plain text from `.doc` files with the existing office_oxide-backed `DOCParser` and serving it as `text/plain; charset=utf-8` from `GET /api/v1/documents/{id}/preview`. If text extraction fails (e.g. non-CGO builds), the endpoint falls back to the original raw-bytes behavior.

Web changes:
- `web/src/components/document-preview/index.tsx`: route `.doc` files to the plain-text previewer; `.docx` keeps using the Word previewer (affects the chunk-result page, search preview modal, and dataflow result page).
- `web/src/pages/document-viewer/index.tsx`: the standalone viewer page also renders `.doc` with the text previewer (document resources only; file-manager downloads still return the original file untouched).

Note: `.doc` is previewed as plain text (no formatting/images), which is the limit of office_oxide's legacy `.doc` support. Format-preserving preview would require a heavyweight converter such as LibreOffice.


wangq8 (commented):
LGTM. Straightforward fix that makes legacy .doc documents previewable.

**What's good:**
- Graceful degradation: if `extractDOCPreviewText` fails (e.g., non-CGO builds without office_oxide), the endpoint falls through to the original raw-bytes behavior — no breaking change for environments where the parser isn't available.
- Correct content-type switch to `text/plain; charset=utf-8` when text extraction succeeds.
- Frontend changes are minimal and surgical: `document-preview/index.tsx` routes `.doc` to the text previewer instead of the DOCX-only word previewer; `document-viewer/index.tsx` does the same but excludes the `files` resource view (where downloads return the original file — correct behavior).
- The `DOCParser` call at the Go level is wrapped in a small helper, keeping the main preview method readable.

**Minor note:**
The `resource !== 'files'` guard in `document-viewer/index.tsx` is correct since file-manager downloads should return the untouched file. Might be worth a comment in the code explaining why `.doc` is excluded for the files resource, in case future maintainers wonder.

No regressions. The plain-text preview is a reasonable tradeoff vs. pulling in LibreOffice.

xugangqiang:
### 1. Regression risk (important)
**Yes — the failure fallback is self-contradictory and worse than today.** When `extractDOCPreviewText` errors, the backend falls back to **raw binary `.doc` bytes + `application/msword`**, but the frontend now routes `.doc` **unconditionally** to `TxtPreviewer` (`document-preview/index.tsx` `['txt','doc']`), and `TxtPreviewer` (`txt-preview.tsx:23-27`) uses `FileReader.readAsText` **without inspecting the response `Content-Type`**. So on any parse failure the user sees binary rendered as UTF-8 garbage — strictly worse than the current explicit "not available" message. Two real cases:
- Non-CGO builds (dev/`go test` without `office_oxide`): by design falls back, but the UI shows garbage.
- Genuinely unparseable `.doc` (corrupt/protected/unsupported OLE2): happens on production too.

Also note `document-preview` routes `.doc` to `TxtPreviewer` with **no** `resource !== 'files'` guard that `document-viewer` has — a smell that can produce garbage if any caller passes a `/files/{id}` (raw binary) URL for `.doc`.


### 2. On the existing comment + suggested fix
The comment (parser-failure responses must be safe for both preview consumers) is **correct**. But "retain raw bytes only for clients that can handle them" is moot here: the only consumer, `TxtPreviewer`, cannot handle binary. Suggested minimal, local fix — make the preview endpoint return `text/plain` on failure too:

```go
if ext == "doc" {
    if text, cerr := extractDOCPreviewText(ctx, fileName, data); cerr == nil {
        data = []byte(text)
        contentType = "text/plain; charset=utf-8"
    } else {
        // Parsing unavailable (non-CGO build / unsupported .doc):
        // return a plain-text notice instead of raw bytes so
        // TxtPreviewer never renders binary as garbage.
        data = []byte("Preview is not available for this Word document")
        contentType = "text/plain; charset=utf-8"
    }
}
```

This preserves the success-path improvement, restores the old clear behavior on failure, and needs no frontend change. Secondary: unify the `.doc` routing guards between `document-preview` and `document-viewer` so a `files` URL can't slip through.

(Optional, larger: have `TxtPreviewer` dispatch on `Content-Type` and return 415 on failure — more general but out of scope for this PR.)

xugangqiang:
### 1. Regression risk
- **Conditional regression (once the web change lands):** on any parse failure the backend falls back to raw binary `.doc` + `application/msword`, while a `TxtPreviewer` route (`FileReader.readAsText`, no `Content-Type` check) would render binary as UTF-8 garbage — strictly worse than today's explicit message. Triggers on: non-CGO builds (dev/`go test` without `office_oxide`), and genuinely unparseable `.doc` (corrupt/protected/unsupported OLE2) which can happen in production too.
- **Observability regression:** parse failures are silently swallowed (no log), making production issues hard to diagnose.
- **Performance:** full re-parse on every preview request.
- **Safe by design:** `GetDocumentPreview` serves only `/documents/:id/preview`; file downloads go through `/files/:id`, so converting `.doc` preview to text does **not** affect downloads. Good.
- Edge cases (`fileName` empty / no extension) correctly skip the branch.

### 2. On the existing comments + how to fix

**CodeRabbit (automated):**
- "Add focused tests" — correct, see #7.
- "Remove full-document `console.log` in `TxtPreviewer`" (`txt-preview.tsx:43,49`) — this is a pre-existing privacy/perf smell and worth fixing, but it is **not part of this diff**; please handle it in a separate PR rather than bundling it here.

**xugangqiang:**
- The regression analysis (fallback-to-binary + unconditional `TxtPreviewer` = garbage) is logically correct, **but it assumes the web `.doc → TxtPreviewer` routing is present — which it is not in the current diff.** So that specific regression does not apply to the current PR state, though the underlying principle (never feed raw binary to a text renderer) must be enforced once the web change lands.
- The suggested fix (return `text/plain` on failure too) is correct and minimal. I fully endorse it:

```go
if ext == "doc" {
    if text, cerr := extractDOCPreviewText(ctx, fileName, data); cerr == nil {
        data = []byte(text)
        contentType = "text/plain; charset=utf-8"
    } else {
        // Parsing unavailable (non-CGO build / unsupported .doc):
        // return a plain-text notice instead of raw bytes so
        // TxtPreviewer never renders binary as garbage.
        data = []byte("Preview is not available for this Word document")
        contentType = "text/plain; charset=utf-8"
    }
}
```
- Also unify the `.doc` routing guards between `document-preview` and `document-viewer` so a `/files/{id}` (raw binary) URL cannot slip through to the text previewer.

**What both reviewers missed:** the PR is **incomplete** — the web changes described in the body are absent from the diff. This must be addressed before merge.

### 3. Test coverage — insufficient
There is **no test for `GetDocumentPreview`** at all (only `document_metadata_test.go`, `document_test.go`, `document_upload_test.go` exist). At minimum add:
1. **Success path** (CGO build): inject `.doc` bytes + mock storage; assert `Data` is the extracted text and `ContentType == "text/plain; charset=utf-8"`.
2. **Failure path**: when `ParseWithResult` errors, assert original bytes + `application/msword` are preserved (or, after adopting the fix in #6, that a `text/plain` notice is returned).
3. **Non-CGO build**: `extractDOCPreviewText` returns `ErrOfficeCGORequired` and falls back.
4. **Web**: verify `.doc` actually routes to `TxtPreviewer` (currently unimplemented, so untested).

wangq8 (commented):
Good fix! Legacy `.doc` (OLE2) preview was indeed broken and this is a clean solution using the existing `DOCParser` infrastructure.

**What's good:**
- Reuses the existing `office_oxide`-backed parser — no new dependencies
- Falls back gracefully: if text extraction fails, the original raw bytes behavior is preserved
- Serves as `text/plain; charset=utf-8` which is appropriate for extracted plain text
- `ParseWithResult` is the right API to use here

**One question:**
The PR description mentions web frontend changes in `document-preview/index.tsx` and `document-viewer/index.tsx` to route `.doc` files to the plain-text previewer, but the diff only shows the Go backend change. Were those web changes already merged separately, or are they still needed for the fix to work end-to-end? Without the frontend changes, the browser may still try to render `.doc` as Word format.

Otherwise LGTM — nice work on the Go side!

PR #17769, opened 2026-08-04
Fix agent stop chat should not cancel the task

wangq8:


cike8899 (approved):

wangq8 (commented):
## PR Review: Fix agent stop chat should not cancel the task

### Summary
This PR changes the "stop conversation" behavior so that stopping the chat stream in the UI no longer cancels the backend agent task. Previously, clicking "stop" would call `stopMessage(taskId)` which sent an API request (`cancelConversation`) to abort the backend task. Now it only calls `stopOutputMessage()` to close the SSE stream.

### What Changed
- **`web/src/pages/agent/chat/use-send-agent-message.ts`** (+1 / −8)
  - Removed import of `useStopMessage`
  - Removed the `stopMessage` hook call
  - In `stopConversation`: removed the `cancelConversation` call and simplified the dependency array

### Analysis

✅ **Intent is clear and reasonable.** Stopping chat display should not kill a potentially useful backend task. The agent may still be producing valuable results.

✅ **Backend task cancellation is preserved on page unload.** `useStopMessageUnmount` in `canvas/index.tsx` still attaches a `beforeunload` listener that calls `stopMessage(taskId)`, so navigating away from the page will still cancel the task. This is the right fallback.

✅ **Dependency array is correctly simplified.** Removed `firstAnswer`, `isShared`, and `stopMessage` since they are no longer used in the callback body.

✅ **No stale imports or dead references.** The removed `useStopMessage` import is the only usage in this file. The hook itself is still used internally by `useStopMessageUnmount`, so no dead code is created.

⚠️ **Consideration: No explicit cancel path in-chat.** After this change, there is no way for a user to explicitly cancel a running task from within the chat UI. If an agent is stuck or misbehaving, the only option is to navigate away from the canvas page (which triggers the `beforeunload` cancel). Depending on UX requirements, you might want to consider keeping a separate "cancel task" button for non-shared sessions, or accept this as intentional behavior.

### Verdict
**LGTM** — clean, focused change with low regression risk. The one UX consideration about explicit cancellation is worth noting but likely acceptable given the `beforeunload` fallback.

PR #17770, opened 2026-08-04
fix: implement Tongyi-Qianwen TTS via DashScope OpenAI-compatible endpoint

euvre:
### Summary

Chat read-aloud (TTS) always failed with "TTS synthesis produced no audio" when the tenant's default TTS model is provided by Tongyi-Qianwen (Aliyun DashScope). The Go `AliyunModel` driver returned `no such method` for `AudioSpeech`, and the provider config had no TTS endpoint suffix, so every synthesis segment failed and the endpoint returned an error instead of audio.

Changes:
- Implement `AliyunModel.AudioSpeech` against DashScope's OpenAI-compatible `POST /compatible-mode/v1/audio/speech` endpoint (Qwen TTS models such as `qwen-tts-flash`).
- `voice` is required by Qwen TTS; it defaults to `Cherry` and can be overridden via `TTSConfig.Params["voice"]`. `response_format` follows `TTSConfig.Format` (e.g. `mp3`).
- Implement `AudioSpeechWithSender` as a single-chunk forward (the endpoint is non-streaming).
- Add the `tts` URL suffix for Tongyi-Qianwen in `conf/models/aliyun.json`.
- Add unit tests covering the default/explicit voice, API error propagation, missing TTS suffix, and the sender fallback.

### DIFF
--- a/conf/models/aliyun.json
+++ b/conf/models/aliyun.json
@@ -10,7 +10,8 @@
     "chat": "compatible-mode/v1/chat/completions",
     "embedding": "compatible-mode/v1/embeddings",
     "rerank": "compatible-api/v1/reranks",
-    "models": "compatible-mode/v1/models"
+    "models": "compatible-mode/v1/models",
+    "tts": "compatible-mode/v1/audio/speech"
   },
   "models": [
     {
--- a/internal/entity/models/aliyun.go
+++ b/internal/entity/models/aliyun.go
@@ -389,13 +389,104 @@ func (a *AliyunModel) TranscribeAudioWithSender(ctx context.Context, modelName *
 	return fmt.Errorf("%s, no such method", a.Name())
 }
 
+// aliyunTTSDefaultVoice is used when the caller does not specify a voice;
+// DashScope's Qwen TTS models require one.
+const aliyunTTSDefaultVoice = "Cherry"
+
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
LGTM. Clean implementation of the Aliyun/DashScope TTS endpoint.

**What's good:**
- Proper implementation of both `AudioSpeech` and `AudioSpeechWithSender` — the latter correctly forwards the non-streaming response as a single chunk via the sender callback, which matches the DashScope endpoint behavior.
- Good defensive checks: validates API config, model name, audio content, and TTS URL suffix before building the request.
- The `voice` parameter defaults to `Cherry` and is overridable via `TTSConfig.Params`, with `response_format` mapped from `TTSConfig.Format` — this is the right pattern for provider-specific config.
- `conf/models/aliyun.json` updated with the `tts` URL suffix, so existing tenants get the new endpoint without migration pain.
- Excellent test coverage: default voice, explicit voice override, API error propagation, missing TTS suffix, and sender single-chunk behavior — all well-structured using `httptest` servers.

**Potential improvement (non-blocking):**
Consider adding a test for `AudioSpeechWithSender` with a nil sender to cover the error path — it's a small edge case but easy to add given the existing test patterns.

No regressions or issues found. Nice work!

Lynn-Inf (approved):
LGTM

PR #17772, opened 2026-08-04
fix: preserve raptor tree titles

buua436:

RAPTOR summaries with leading whitespace could produce empty tree node titles, causing parent-child relations to be discarded as self-loops.

This change trims the model output before extracting the first-line title.

### Type of change

### DIFF
--- a/rag/advanced_rag/knowlege_compile/raptor.py
+++ b/rag/advanced_rag/knowlege_compile/raptor.py
@@ -231,12 +231,14 @@ class RecursiveAbstractiveProcessing4TreeOrganizedRetrieval:
                     "",
                     cnt,
                 )
+                cnt = str(cnt or "").strip()
                 logging.debug(f"SUM: {cnt}")
 
                 self._check_task_canceled(task_id, "before embedding")
 
                 embds = await self._embedding_encode(cnt)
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Good catch and clean fix! The change correctly addresses the edge case where LLM output with leading whitespace/newlines could produce empty tree node titles.

**Problem:** When the model returns a summary with leading whitespace or newlines, `cnt.split("\n")[0]` returns an empty string as the title. Empty titles cause parent-child relations to be discarded as self-loops in the RAPTOR tree.

**Fix analysis:**
- `cnt = str(cnt or "").strip()` removes leading/trailing whitespace from the model output before title extraction
- `splitlines()[0]` is actually a nice improvement over `split("\n")[0]` since it handles different line endings more robustly
- The `if cnt else ""` guard handles the edge case of completely empty content after stripping
- The chopped "continue" pattern removal via `re.sub` still works correctly after this change since it runs before the strip
- The stripped `cnt` is passed to embedding, which is fine (embedding with vs without surrounding whitespace won't meaningfully differ)

**Edge cases handled:**
- Summary starting with `\n` → empty title → now correctly stripped
- Summary with leading spaces → title now stripped properly
- Completely empty summary → returns empty string instead of IndexError

LGTM!

PR #17773, opened 2026-08-04
fix(web): add missing 'general' chunk method description in locales

euvre:
### Summary

In Go mode, the dataset setting page loads the chunk method list from the Go backend's builtin pipeline catalog (`GET /api/v1/pipelines?type=builtin`), where the canonical parser id of the General method is `general` (the legacy `naive` id is only kept as an alias). The "Builtin pipeline introduction" panel resolves the chunk method description via `t(camelCase(parserId))` in the `knowledgeConfiguration` namespace, but the locale files only contained the legacy `naive` key. As a result, the raw i18n key `knowledgeConfiguration.general` was rendered as the description text for the General method.

This PR adds the missing `general` chunk method description to all complete locale files (`en`, `zh`, `zh-traditional`, `de`, `fr`, `id`, `it`, `ja`, `ko`, `pt-br`, `ru`, `tr`, `vi`, `bg`, `ar`), using the same content as `naive` since General is the renamed naive chunking method. The `naive` key is kept because the Python backend still uses the `naive` parser id. The stub `es` locale is unchanged and falls back to English.


wangq8 (commented):
Thanks for this fix! The PR correctly adds the missing `general` key to all locale files, with translations matching the existing `naive` entries since General is just the renamed naive chunking method.

**What this fixes:** The Go backend uses `general` as the canonical parser id, but the frontend locale files only had `naive`, causing the raw i18n key `knowledgeConfiguration.general` to be displayed instead of the proper description.

**Review notes:**
- All 15 locale files correctly add the `general` key with the same content as `naive`
- The existing `naive` key is preserved for backward compatibility with the Python backend
- The `es` locale (stub) is intentionally left unchanged since it falls back to English
- Content consistency looks good across all translations

LGTM, no issues found.

wangq8:
@euvre , the `naive:` already have it, try to not duplicate the items.

PR #17203, opened 2026-07-21
fix(chunker): enforce strict chunk_token_num cap on .txt / PDF / email paths

skbs-eng:
Fixes #17202 (and complements #12109).

## Problem

`RAGFlowTxtParser.parser_txt` (`deepdoc/parser/txt_parser.py:36-47`) and `rag.nlp.naive_merge` (`rag/nlp/__init__.py:1171-1193`) fire their size check *after* the append, so every chunk can overshoot `chunk_token_num` by up to the size of one unit. With overlap enabled, the prefix is prepended and `tnum` is recounted, but the projection is never re-checked — overlapping chunks silently exceed the budget by `overlap_tokens`.

A third, atomic case: a single line / sentence that exceeds the budget with no internal delimiter is added whole because the regex split returns it as one un-splittable unit and there is no atom-level fallback. `RAGFlowHtmlParser.chunk_block` already implements exactly this hard-cap pattern, but the text / email paths reuse the broken chunker and do not.

Measured on a live dataset (336 `.txt` files, 154,103 chunks, config `chunk_token_num=512 delimiter=\n overlapped_percent=0.1`): 56.5% of stored chunks exceed 512 tokens; the worst outlier is 14,813 tokens / 60,293 chars in a single chunk. Symptom downstream: rerank failures on the >2048-token outliers (ref. #12109) and silent embedding truncation on every oversize chunk.

## Fix

Mirror the proven pattern in `RAGFlowHtmlParser.chunk_block`:

1. **Proactive projected-total check** in `TxtParser.parser_txt` and in `naive_merge.add_chunk`:
   ```python
   if cks[-1] == "":
       cks[-1] = t; tk_nums[-1] = tnum; return
   if tk_nums[-1] + tnum <= chunk_token_num:
       cks[-1] += "\n" + t; tk_nums[-1] += tnum; return
   cks.append(t); tk_nums.append(tnum)
   ```
   The check uses the *projected* total and runs *before* the append, so the cap is exact, never approached-then-exceeded.

2. **Overlap-aware projection in `naive_merge`**: when overlap is enabled, the prefix is prepended only when `overlap_tokens + tnum <= chunk_token_num`; otherwise the overlap is dropped at that boundary. The naive_merge-with-images mirror gets the same treatment. Custom-delimiter behaviour is preserved per the existing test suite.

[...]

PR #17774, opened 2026-08-04
Closed revert: align pipeline delimiter chunking (#17723)

xugangqiang:
## Summary

Reverts #17723 (`fix: align pipeline delimiter chunking`).

After review we concluded that #17723 should be reverted and the delimiter-chunking behavior reworked from a clean base, rather than fixed incrementally on top of it. The revert restores the pre-#17723 Python behavior byte-for-byte (verified against the parent commit `3e7cfbe05`).

## Why revert instead of fix-on-top

#17723 introduced several regressions against the intended chunking contract and removed behavior that the default path still relies on:

- **Dropped the parser token-size cap.** `deepdoc/parser/txt_parser.py` removed `_split_oversized_unit` and the proactive `tk_nums[-1] > chunk_token_num` cap, so over-budget sections are no longer sub-split. The pre-#17723 path enforced the cap correctly.
- **Globally discards delimiters.** `_split_text_by_pattern` was changed from keeping the delimiter (glued to the preceding segment) to silently dropping it for *every* delimiter mode, including `token_size`, which changes chunk text content for existing users.
- **Invalid `delimiter_mode == "delimiter"` JSON branch.** The new JSON branch flushes the text buffer with `"".join(...)` (adjacent item text gets concatenated, e.g. `helloworld`), does not carry `PDF_POSITIONS_KEY`, and returns early — bypassing `_split_chunk_docs_by_children`, so `children_delimiters` is silently ignored whenever `delimiter_mode == "delimiter"` + `children_delimiters` are combined.
- **Deleted the guard test.** `test/unit_test/deepdoc/parser/test_txt_parser.py` (178 lines) — which protected the token-cap invariant — was removed.

Restoring the pre-#17723 state is lower-effort than undoing each regression individually, and gives a clean base for the intended **blend model** (`delimiter` = preferred break point, `token_size` = hard upper bound, the two are orthogonal and combinable).

## Changes in this PR

Reverts the 5 files changed by #17723:
- `deepdoc/parser/txt_parser.py` — restores `_split_oversized_unit` and the proactive token cap.
- `rag/flow/chunker/token_chunker.py` — restores delimiter-keeping split and the pre-#17723 JSON/text paths.
- `rag/flow/parser/parser.py` — removes the `keep_delimiters=True` change.
- `test/unit_test/deepdoc/parser/test_txt_parser.py` — restores the deleted guard test.
- `web/src/utils/__tests__/delimiter-preview.test.ts` — restores the 4-line preview expectation.

## Known follow-up (out of scope here)

The Go side was aligned to the *post-#17723* Python behavior in #17729 / #17739 / #17740 (`internal/ingestion/component/chunker/token.go`). After this revert, a follow-up PR will re-align Go to the restored Python behavior and implement the blend model. Parity-infra PRs #17735 / #17744 are intentionally deferred per the discussion.

## Test plan

- [ ] `python -m pytest test/unit_test/deepdoc/parser/test_txt_parser.py` passes (restored test).
- [ ] `pytest` on `rag/flow/chunker` passes.
- [ ] Manual: `delimiter_mode=delimiter`, `delimiter_mode=token_size`, and `delimiter_mode=one` produce the pre-#17723 outputs.

### DIFF
--- a/deepdoc/parser/txt_parser.py
+++ b/deepdoc/parser/txt_parser.py
@@ -19,6 +19,7 @@ import re
 
 from common.token_utils import num_tokens_from_string
 from deepdoc.parser.utils import get_text
+from rag.nlp import _split_oversized_unit
 from rag.nlp.delim import (
     compile_delimiter_pattern,
     normalize_text_newlines,
@@ -27,12 +28,12 @@ from rag.nlp.delim import (
 
 
 class RAGFlowTxtParser:
-    def __call__(self, fnm, binary=None, chunk_token_num=128, delimiter="\n!?;。；！？", keep_delimiters=False):
+    def __call__(self, fnm, binary=None, chunk_token_num=128, delimiter="\n!?;。；！？"):
         txt = get_text(fnm, binary)
-        return self.parser_txt(txt, chunk_token_num, delimiter, keep_delimiters)
+        return self.parser_txt(txt, chunk_token_num, delimiter)
 
     @classmethod
-    def parser_txt(cls, txt, chunk_token_num=128, delimiter="\n!?;。；！？", keep_delimiters=False):
+    def parser_txt(cls, txt, chunk_token_num=128, delimiter="\n!?;。；！？"):
         if not isinstance(txt, str):
             raise TypeError("txt type should be str!")
         cks = [""]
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Thanks for the thorough revert and detailed rationale.

**What I checked:**
- `txt_parser.py`: Restores `_split_oversized_unit` with the proactive token-cap check and merged-token projection — this correctly prevents oversized chunks.
- `token_chunker.py`: Restores delimiter-keeping split (gluing delimiter to preceding segment) and removes the broken `delimiter_mode == "delimiter"` JSON branch that was concatenating adjacent items with `"".join`.
- `parser.py`: Removes the `keep_delimiters=True` override that was passed to `RAGFlowTxtParser`.
- `test_txt_parser.py`: Restores the 178-line guard test covering no-overshoot, boundary packing, oversized-line sub-splitting, empty input, and newline-join token cap edge cases.
- `delimiter-preview.test.ts`: Restores the 4-line formatting expectation.

**One note:** The PR description mentions a follow-up to re-align the Go side (`internal/ingestion/component/chunker/token.go`) to the restored Python behavior. Please make sure that follow-up is tracked (issue or linked PR) so the Go/Python parity doesn't drift.

LGTM.

xugangqiang:
Closing: this revert of #17723 targets the wrong object. The regression actually originates in #17203 (strict-cap atom-split), and #17723's delimiter handling had its own bugs (consecutive-delimiter drop, JSON glue, ignored children_delimiters / PDF positions). The fix is PR #17808, which removes atom-split, keeps delimiter as a chunk boundary, and preserves the #17723 delimiter direction — without re-introducing delimiter-in-chunk or the strict cap. Reverting #17723 would undo the correct delimiter direction and re-add the hard cap.

PR #17634, opened 2026-07-31
feat(go-models): unify OpenAI-compatible token usage extraction

jay77721:
## Summary
Relate to #17284 .

- Add `usage_parser.go`: `extractOpenAIUsage` / `extractOpenAIStreamUsage`
  with multi-field fallback (`prompt_tokens`/`input_tokens`,
  `completion_tokens`/`output_tokens`).
- Add `parser_config.go`: `ParserConfig` struct and `OpenAIParserConfig`.
- Add `response_handler.go`: `HandleNonStreamingResponse` /
  `HandleStreamingResponse` as single entry points that extract usage,
  populate `chatConfig.UsageResult`, and emit a `StreamUsage` log.
- Extend `base_model.go` with `doRequest` / `doStreamRequest`.
- Migrate `deepseek` driver to shared handlers, cutting ~100 lines.
- Add `usage_parser_test.go`.

## Testing

- `bash build.sh --test ./internal/entity/models/...` passes.
- Manual testing with deepseek confirms `StreamUsage` log appears.

PR #17775, opened 2026-08-04
feat(go-models): migrate batch 6 model drivers to unified usage recording

jay77721:
## Summary

Relate to #17284.

Completes the migration of the four non-OpenAI-compatible model drivers (`anthropic`, `cohere`, `google`, `bedrock`) onto the shared usage-recording path. Earlier batches (#17634, #17643, #17696–#17700) covered only the OpenAI-compatible cluster; these four providers ship wire formats that do not fit the OpenAI `choices[0].delta` / `usage` block template and so were left for a separate pass.

Per the maintainer's guidance for this batch, each driver is migrated on its own terms rather than forced through a single template. The shared machinery used is intentionally small: `recordResponseUsage`, `parseChatCompletionResponse`, `BaseModel.newJSONPostRequest`, and the existing `authHeader` hook for non-Bearer auth.

### Anthropic (`anthropic.go`)

- **Non-streaming `ChatWithMessages`** now uses `newJSONPostRequest` + `parseChatCompletionResponse` with a new `chatResponseExtractor` (`parseAnthropicChatResponse`) that maps the Messages API's `content[]` blocks into the common response shape and extracts `usage.input_tokens` / `usage.output_tokens`. The `x-api-key` auth header is wired through the `authHeader` hook, matching the xiaomi / xunfei pattern.
- **`ChatStreamlyWithSender`** is unchanged: its bespoke `message_start` / `message_delta` usage merge and `content_block_delta` fan-out do not fit the unified handler, and it already records usage via `applyStreamUsage`.
- **`ListModels`** unchanged.

### Cohere (`cohere.go`)

- **No change.** Non-streaming was already on `parseChatCompletionResponse` and streaming already calls `applyStreamUsage`.

### Google (`google.go`)

- New `googleUsageFromMetadata` helper translates the genai SDK's `GenerateContentResponseUsageMetadata` into the package's `TokenUsage`. `CandidatesTokenCount + ThoughtsTokenCount` is summed into `CompletionTokens`; `TotalTokenCount` is used directly.
- Both `ChatWithMessages` and `ChatStreamlyWithSender` call `recordResponseUsage` after the SDK call. The streaming path captures the last-seen `UsageMetadata` across chunks and records it once after the iterator finishes.

### Bedrock (`bedrock.go`)

- New `bedrockUsage` struct + `bedrockUsageFromMap` / `bedrockUsageToMap` helpers handle Bedrock's camelCase `inputTokens` / `outputTokens` / `totalTokens` shape.
- `bedrockConverseResponse` gains a `Usage` field; `ChatWithMessages` records usage after JSON unmarshal.
- `decodeBedrockEventStream` gains an `onMetadata` callback that the streaming path uses to capture the metadata event's usage block; `ChatStreamlyWithSender` records it before emitting the `[DONE]` sentinel. The signature change is additive (third parameter is nil-tolerant).

### Shared helper (`base_model.go`)

- `recordResponseUsage`: when the caller passes a nil `*ModelUsage` but supplies `usage`, we surface a synthetic `CollectModelUsage` call against an empty `ModelUsage` so the analytics path can observe the provider's reported token counts. Without this, the existing service-layer callers (`model_chat`, `generator`, `memory_extractor`, etc.) that pass `nil` would never reach the stats driver and the token usage would be invisible.
- `collectModelUsage`: when `StartAt` is the zero time, leave `ResponseTimeMS` at zero instead of reporting an epoch delta.

## Test impact

No new tests required. `anthropic_test.go` (519 lines), `cohere_test.go`, `google_test.go`, and `bedrock_test.go` (913 lines) all continue to pass via the package's unit tier (`bash build.sh --test ./internal/entity/models/...`).

The `CollectModelUsage` real-storage path in `clickhouse_ee.go` is intentionally left as a follow-up.

### DIFF
--- a/internal/entity/models/anthropic.go
+++ b/internal/entity/models/anthropic.go
@@ -42,6 +42,11 @@ func NewAnthropicModel(baseURL map[string]string, urlSuffix URLSuffix) *Anthropi
 			BaseURL:    baseURL,
 			URLSuffix:  urlSuffix,
 			httpClient: NewDriverHTTPClient(false),
+			// Anthropic authenticates with the "x-api-key" header instead
+			// of the default "Authorization: Bearer".
+			authHeader: func(cfg *APIConfig) (string, string) {
+				return "x-api-key", strings.TrimSpace(*cfg.ApiKey)
+			},
 		},
 	}
 }
@@ -65,7 +70,6 @@ func (a *AnthropicModel) ChatWithMessages(ctx context.Context, modelName string,
 	if err := a.baseModel.APIConfigCheck(apiConfig); err != nil {
 		return nil, err
 	}
-	apiKey := strings.TrimSpace(*apiConfig.ApiKey)
 	if len(messages) == 0 {
 		return nil, fmt.Errorf("messages is empty")
 	}
@@ -97,19 +101,15 @@ func (a *AnthropicModel) ChatWithMessages(ctx context.Context, modelName string,
 	}
 	applyAnthropicChatConfig(reqBody, chatModelConfig)
 
-	jsonData, err := json.Marshal(reqBody)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal request: %w", err)
-	}
-
 	ctx, cancel := context.WithTimeout(ctx, nonStreamCallTimeout)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(jsonData))
+	req, err := a.baseModel.newJSONPostRequest(ctx, url, apiConfig, reqBody)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create request: %w", err)
+		return nil, err
 	}
-	setAnthropicHeaders(req, apiKey, false)
+	req.Header.Set("anthropic-version", anthropicVersion)
+	req.Header.Set("Accept", "application/json")
 
 	resp, err := a.baseModel.httpClient.Do(req)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


jay77721:
Addressed all three findings from review. Pushed as a follow-up commit (`a437bcb65`).

**bedrock.go — `decodeBedrockEventStream`**
The `case "messageStop"` arm no longer `return nil`; it only sets `sawTerminal = true` and the loop continues until EOF. The AWS Bedrock Converse-Stream protocol sends a final `metadata` frame (carrying token usage) **after** `messageStop`, so returning early silently dropped the streaming usage. The new `TestBedrockStreamRecordsUsageFromPostMessageStopMetadata` covers this exact sequence (`messageStart → contentBlockDelta → messageStop → metadata`) and asserts both `modelUsage` and `chatConfig.UsageResult` are populated.

**bedrock.go — `ChatStreamlyWithSender`**
Switched the trailing `recordResponseUsage(...)` call to `applyStreamUsage(chatModelConfig, modelUsage, streamUsage)` so `chatConfig.UsageResult` is also populated — matching the unified streaming behaviour of every other driver. No usage is collected twice: `applyStreamUsage` is the single entry point that does both the field-write and the clickhouse call.

**google.go — `googleUsageFromMetadata`**
`ToolUsePromptTokenCount` is now folded into `PromptTokens`. The zero-metadata presence check and the fallback `total = in + out` already cover the wider input range, so no further change there. New `TestGoogleUsageFromMetadataIncludesToolUsePromptTokens` covers both paths (TotalTokenCount present and absent).

**google.go — `ChatStreamlyWithSender`**
Same `recordResponseUsage` → `applyStreamUsage` switch as bedrock, with the same justification. The `responseID` tracking is removed; streaming drivers in this codebase intentionally do not propagate `RequestID` (per the comment on `applyStreamUsage`).

`bash build.sh --test ./internal/entity/models/...` — full package green; all four driver test families (anthropic 14, cohere 2, google 24, bedrock 24) pass.

Haruko386 (approved):
LGTM

PR #17776, opened 2026-08-04
fix(go-api): reject duplicated MCP server name on update

euvre:
### Summary

The Go `UpdateMCPServer` endpoint only validated the name length, so an MCP server could be renamed to a name already used by another server of the same tenant, while `CreateMCPServer` rejects duplicated names. This inconsistency allowed duplicate MCP server names to appear through the edit/rename flow.

This PR mirrors the Python update API behavior: when a new name is provided and differs from the current one, it checks `ExistsByNameAndTenant` and returns a data error (`duplicated MCP server name`) on conflict. Keeping the current name or renaming to a fresh name still succeeds.

A unit test (`TestUpdateMCPServerRejectsDuplicatedName`, in-memory SQLite, unit tier) covers three cases: renaming to an existing name is rejected, keeping the current name succeeds, and renaming to a fresh name succeeds.

### DIFF
--- a/internal/service/mcp.go
+++ b/internal/service/mcp.go
@@ -316,6 +316,15 @@ func (s *MCPService) UpdateMCPServer(ctx context.Context, tenantID, mcpID string
 	if serverName != "" && len([]byte(serverName)) > mcpServerNameLimit {
 		return nil, common.CodeDataError, fmt.Errorf("Invalid MCP name or length is %d which is large than 255.", len([]byte(serverName)))
 	}
+	if serverNameProvided && serverName != server.Name {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Good catch — this closes the inconsistency between create and update for MCP server name uniqueness.

**What I checked:**
- Duplicate check only fires when `serverNameProvided && serverName != server.Name` — keeping the current name still passes.
- Uses `ExistsByNameAndTenant` which is the same check used in `CreateMCPServer`, so the uniqueness contract is consistent.
- Returns `CodeDataError` with a clear message on conflict — mirrors the existing error style.
- Test covers all three cases: rename to existing name (rejected), keep current name (allowed), rename to fresh name (allowed).

Approved.

PR #17777, opened 2026-08-04
fix: allow empty kb_ids when linking files to datasets

euvre:
### Summary

In the Go file manager, saving the "Link Knowledge Base" dialog after removing all linked datasets sends `kb_ids: []` to `POST /api/v1/files/link-to-datasets`. The Go handler validated arguments by slice length and rejected the empty list with `101 required argument are missing: kb_ids; `, making it impossible to unlink all knowledge bases from a file.

Python's `@validate_request("file_ids", "kb_ids")` only checks key presence (`arg not in input_arguments`), so an explicit empty list is valid — in replace mode it unlinks all datasets (existing file2document mappings and documents are removed, none are created).

This change checks for a nil slice (key absent) instead of an empty slice, matching the Python semantics. An empty body or missing key still returns 101 as before.

Verified with `bash build.sh --test ./internal/handler/...`.

### DIFF
--- a/internal/handler/file.go
+++ b/internal/handler/file.go
@@ -551,17 +551,19 @@ func (h *FileHandler) LinkToDatasets(c *gin.Context) {
 
 	var req document.LinkToDatasetsRequest
 	// Tolerate bind errors: a malformed or empty body simply leaves the fields
-	// empty, which the validate_request-style check below reports as missing
+	// nil, which the validate_request-style check below reports as missing
 	// arguments — matching Python's @validate_request behaviour and code.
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Clean fix, matches the Python semantics correctly.

**What I checked:**
- Changed from `len(req.FileIDs) == 0` to `req.FileIDs == nil` (and same for KbIDs).
- Python's `@validate_request` checks key *presence*, not value length, so an explicit empty list is valid.
- An empty body or genuinely missing key still returns 101 as before (nil check catches both).

Approved.

PR #17779, opened 2026-08-04
fix: search files recursively within current folder subtree

euvre:
### Summary

In the file manager, keyword search within a folder only matched direct children of that folder (`parent_id = pfID`), so files and folders nested in sub-folders could never be found when searching from a parent directory.

This PR fixes both backends so that:

- When `keywords` is non-empty, the name match is scoped to the entire subtree under the current folder. Descendant IDs are collected via BFS over the tenant's `(id, parent_id)` pairs, with cycle protection:
  - Go: `FileDAO.GetByPfID` + new `GetSubtreeIDs` helper.
  - Python: `FileService.get_by_pf_id` + new `get_subtree_ids` helper.
- When `keywords` is empty, behavior is unchanged: only direct children are listed.
- Tenant isolation and exclusion of the current folder itself are preserved.

Go unit tests (in-memory SQLite, default unit tier) cover subtree search hits, folder-name matching, subtree scoping, tenant isolation, and the unchanged direct-children listing.


wangq8 (commented):
Thanks for this PR! The recursive subtree search is a solid UX improvement for the file manager.

**What I checked:**
- `GetSubtreeIDs` uses BFS with cycle protection (visited set) — correct and safe.
- Tenant isolation preserved — the query scopes to `tenant_id` before building the adjacency map.
- Keyword-empty path unchanged — only direct children listed when no keywords, preserving the existing contract.
- Tests cover: subtree hits, nested file matching, folder-name matching, subtree scoping (no cross-subtree leaks), tenant isolation, and direct-children listing.

**One observation (non-blocking):**
`GetSubtreeIDs` fetches *all* rows for the tenant from the File table to build the in-memory adjacency map. For tenants with very large file trees this could become a performance concern. A recursive CTE would push the traversal to the database layer. Not a blocker for this PR — the correctness is more important — but worth keeping in mind as a follow-up optimization.

LGTM, approved.

wangq8:
@euvre  Python behaviour is search current folder.

api/db/services/file_service.py
```diff
@@ -104,6 +109,29 @@ def get_by_pf_id(cls, tenant_id, pf_id, page_number, items_per_page, orderby, de
...
+        # Return pf_id itself plus the IDs of all entries nested under it
+        # (folders and files), used to scope recursive keyword searches.
+        rows = list(cls.model.select(cls.model.id, cls.model.parent_id).where(cls.model.tenant_id == tenant_id).dicts())
+        children = {}
+        for row in rows:
+            children.setdefault(row["parent_id"], []).append(row["id"])
+
+        ids = [pf_id]
+        in_tree = {pf_id}
+        queue = [pf_id]
```
wangq8: @euvre  check comment, use `deque` instead of []

PR #17781, opened 2026-08-04
fix(file-manager): resolve knowledge base names in link-to-dataset dialog for correct optimistic update

dcc123456:
### Summary

fix(file-manager): resolve knowledge base names in link-to-dataset dialog for correct optimistic update

### DIFF
--- a/web/src/pages/files/hooks.ts
+++ b/web/src/pages/files/hooks.ts
@@ -1,6 +1,5 @@
 import { useSetModalState } from '@/hooks/common-hooks';
 import { useConnectToKnowledge, useRenameFile } from '@/hooks/use-file-request';
-import { useSelectKnowledgeOptions } from '@/hooks/use-knowledge-request';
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. This is a nice refactor — replacing the global knowledge-base name map with an on-demand dataset cache is both more efficient (no upfront fetch of all KBs) and more correct (no stale names).

The approach is sound:
- `datasetCacheRef` survives across re-renders and paginated searches without unnecessary updates
- `useFetchDatasetsByIds` fetches only the datasets not already in the list or cache
- The dialog passes `kb_name` directly in `kbsInfo` instead of relying on the caller to resolve names

One minor note: `datasetCacheRef.current` grows unbounded as the user searches. For this use case it is fine since the dataset count is typically modest, but a bounded LRU (e.g. 200 entries) would be a future-proofing improvement.

Verified that `useFetchDatasetsByIds` exists in `use-knowledge-request.ts`.

PR #17782, opened 2026-08-04
fix: wrap long log messages in agent pipeline log sheet

euvre:
### Summary

In the agent pipeline run log sheet, long unbroken log message text (e.g. component names such as `TokenChunker:Modern...`) could not wrap and overflowed the timeline card boundary, causing the text to visually bleed outside the card.

This PR adds `break-all` to the log message span in `web/src/pages/agent/pipeline-log-sheet/dataflow-timeline.tsx` so long messages wrap inside the card, and adds `whitespace-nowrap` to the datetime and elapsed-time spans so these short fields are never split across lines.


wangq8 (commented):
LGTM. Simple and targeted fix: `break-all` on log messages prevents long unbroken tokens (like `TokenChunker:Modern...`) from overflowing the timeline card, while `whitespace-nowrap` on datetime and elapsed-time spans keeps those short fields from wrapping mid-word. No regressions in the component structure.

dcc123456 (approved):

PR #17786, opened 2026-08-04
fix(go-ingestion): write last component output JSON as debug-log END message

euvre:
### Summary

In Go mode, the **Export JSON** button in the dataflow (agent pipeline) run log panel was permanently disabled.

The button's enabled state is driven by `findEndOutput` (`web/src/pages/agent/hooks/use-download-output.ts`), which locates the `component_id === 'END'` trace entry and `JSON.parse`s its `trace[0].message`; an unparseable/empty result keeps the button disabled. The Python pipeline writes `json.dumps(last_component.output())` as the END message (`rag/flow/pipeline.py:171`), but the Go `DebugLogSink` wrote the plain-text sentinel `"Debug run completed"`, so parsing always failed.

This PR mirrors the Python behavior on the Go side:

- Extend the optional `ResultSink` capability to also carry the raw pipeline run output (`SetResult(dsl, output)`).
- On `Flush`, build the END marker message from the last executed component's output (equivalent to Python's `path[-1]`), with raw embedding vectors stripped to keep the Redis log at Python-scale size.
- Failed runs keep the `[ERROR]`-prefixed message, and runs without output keep the plain-text fallback — the button stays disabled in those cases, matching Python's `isEmpty` semantics.

Tests: updated the `SetResult` call sites and added regression coverage asserting the END message is valid JSON of the last component's output, plus fallback behavior for missing run output and failed runs. All `TestDebugLogSink*` and `TestBuildDebugResultDSL*` tests pass.

### DIFF
--- a/internal/ingestion/task/debug_log_sink.go
+++ b/internal/ingestion/task/debug_log_sink.go
@@ -110,6 +110,11 @@ type DebugLogSink struct {
 	// to the sink via SetResult (the ResultSink capability). It is written onto
 	// the END marker's trace entry in Flush. Empty until SetResult is called.
 	resultDSL json.RawMessage
+	// runOutput is the raw pipeline run output map (output["state"][<id>] is
+	// each component's outputs), handed over together with resultDSL. Flush
+	// uses it to build the END marker's message (the last component's output
+	// JSON) exactly like Python does. Nil until SetResult is called.
+	runOutput map[string]any
 }
 
 // NewDebugLogSink builds a sink that writes to "{canvasID}-{messageID}-logs".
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Excellent fix with thorough test coverage. The approach is well-designed:

- Extending `ResultSink.SetResult` with the raw `output` map keeps the coupling one-directional and doesn't break non-debug sinks
- `endOutputMessage` correctly mirrors Python's `json.dumps(last_component.output())` on the END marker
- `deepCopyStrip` strips embedding vectors to keep Redis at Python-scale size — good attention to detail
- Edge cases covered: no run output → plain-text fallback, run failure → `[ERROR]` prefix wins

Verified that `deepCopyStrip` and `lookupComponentOutput` already exist in `debug_result_dsl.go` and have their own tests. The new tests in `debug_log_sink_test.go` cover the END message JSON contract and both fallback cases clearly.

LGTM. This is a solid Go-side fix.

xugangqiang:
LGTM

PR #17790, opened 2026-08-04
Fix: dataflow log sheet shows the previous run's log in Go mode

euvre:
### Summary:

In Go mode, after clicking **Run** on a dataflow (data pipeline) canvas, the log sheet displays the previous run's log instead of the current one.

Root cause: the Go backend executes the dataflow debug run fully synchronously (`AgentChatCompletions` → `runCanvasPipelineDebug`), so the new `message_id` only arrives with the run response — typically 10+ seconds later. The front-end, however, opens the log sheet *before* sending the run request while still holding the previous run's `messageId`. Reopening the sheet also resets the stop-polling flag, so the sheet keeps polling the old Redis key (whose payload is still within its 30-minute TTL) and renders the previous run's log. If the run response never yields a new `message_id` (e.g. a swallowed request failure), the stale log stays indefinitely.

Fix (front-end, `web/src/pages/agent/hooks/use-run-dataflow.ts`):
- Clear the trace polling key (`setMessageId('')`) when a run starts. Since the trace query is `enabled: !!messageId`, the old log is dropped immediately and the sheet shows the loading state until the new `message_id` arrives and the current run's log is fetched.
- Surface a `requestError` toast when the run request fails without returning a response, instead of silently staying on the previous run's log.

### DIFF
--- a/web/src/pages/agent/hooks/use-run-dataflow.ts
+++ b/web/src/pages/agent/hooks/use-run-dataflow.ts
@@ -1,5 +1,6 @@
 import message from '@/components/ui/message';
 import { useSendMessageBySSE } from '@/hooks/use-send-message';
+import i18n from '@/locales/config';
 import api from '@/utils/api';
 import { get } from 'lodash';
 import { useCallback, useState } from 'react';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. The root cause analysis in the PR description is thorough — the synchronous Go debug run path means the new `message_id` arrives seconds after the sheet opens, so clearing the old polling key upfront is the correct fix.

The error toast for swallowed `send()` failures is a nice UX improvement.

One minor observation: when `res` exists but `data.code !== 0`, the code now falls through silently with no error toast — `messageId` was already cleared so the sheet shows loading indefinitely. This is better than showing stale data (the old behavior), but consider adding an error toast for non-zero response codes too for completeness.

dcc123456 (approved):

PR #17792, opened 2026-08-04
fix: separate raptor node target from output limit

buua436:

RAPTOR incorrectly used `max_token` as the LLM request output limit, causing DeepSeek reasoning tokens to consume the budget and truncate tree summaries.

This change treats `max_token` as the target summary size and passes the target length through the prompt instead.

### Type of change

### DIFF
--- a/rag/advanced_rag/knowlege_compile/raptor.py
+++ b/rag/advanced_rag/knowlege_compile/raptor.py
@@ -32,6 +32,8 @@ from rag.graphrag.utils import (
 )
 from common.misc_utils import thread_pool_exec
 
+from ._common import knowledge_compile_gen_conf
+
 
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Good fix. Separating the summary target size from the LLM output limit is the right approach — reasoning tokens (especially on DeepSeek models) should not consume the node-size budget.

Verified that `knowledge_compile_gen_conf` properly handles the model family detection and sets appropriate generation config:
- DeepSeek: `max_completion_tokens=32768` + thinking disabled
- Qwen3: `enable_thinking` gated on preview variant
- Others: `reasoning_effort="none"`

The prompt addition ("target approximately {N} tokens") is a reasonable way to steer the summarization without hard-capping it. LGTM.

PR #17793, opened 2026-08-04
Fix: Increase the height of the system prompt text area in Memory.

dcc123456:
### Summary

Fix: Increase the height of the system prompt text area in Memory.

### DIFF
--- a/web/src/components/dynamic-form.tsx
+++ b/web/src/components/dynamic-form.tsx
@@ -101,6 +101,7 @@ export interface FormFieldConfig {
    * fields like api_key / instance_name / base_url / group_id.
    */
   autoComplete?: string;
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. The `fieldConfig` prop is a clean way to let field definitions control Textarea sizing without polluting the base interface.

One cleanup suggestion: the commented-out lines `// resize="vertical"` and `// className="resize-none"` in `dynamic-form.tsx` should be removed instead of left as dead code.

The spreading order (`{...field.fieldConfig}` before `{...finalFieldProps}`) is correct — form-level props take priority over field-level overrides. Also nice catch removing the stray `console.log` in the select handler.

PR #17794, opened 2026-08-04
Fix: The tool icon for the agent operator created from the template is not clickable.

cike8899:
### Summary

Fix: The tool icon for the agent operator created from the template is not clickable.


wangq8 (commented):
Thanks for the fix! The core issue (tools from templates missing `id` fields) is correctly addressed.

**What looks good:**

1. **`deep_research.json`**: Adding `id` fields to TavilySearch/TavilyExtract tool nodes is the right fix — these template tools were missing IDs, which caused the React key to be `undefined` and broke click handling.

2. **`tool-node.tsx` & `agent-tools.tsx`**: Fallback keys (`tool.id || `tool-${idx}``) are a good defensive pattern for when IDs might still be missing from other sources.

3. **`ToolCard` refactor**: Removing `useMemo` is important — the memoized element can interfere with Radix `TooltipTrigger asChild` when it clones the child and attaches event handlers. This is likely the real fix for the unclickable tool icon.

4. **`TreeSelect` & `ModelTreeSelect` → `forwardRef`**: Correct pattern for enabling ref forwarding through to the trigger button. Clean conversion.

5. **`next.tsx` refactor**: Clean simplification — extracting `settingFields`, removing the `CollapseComponent` indirection. No regression.

**One thing to watch:**

In `agent-tools.tsx`, `ActionButton` record now falls back to `component_name` when `id` is undefined:
```tsx
record={id || component_name}
```
This changes the identifier passed to `deleteRecord` and `edit` callbacks. Please double-check that downstream handlers work correctly with a `component_name`-based identifier where they previously expected an `id`.

Otherwise looks solid!

PR #17798, opened 2026-08-04
fix: allow team members to update shared datasets

buua436:

Team members could access shared datasets, but updating the shared dataset configuration failed because the update API only checked ownership.

This change reuses the existing accessible-dataset permission check and validates duplicate names against the dataset owner.

### Type of change

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -283,9 +283,13 @@ async def update_dataset(tenant_id: str, dataset_id: str, req: dict):
     if not req:
         return False, "no properties were modified"
 
-    kb = KnowledgebaseService.get_or_none(id=dataset_id, tenant_id=tenant_id)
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The change is clean and well-targeted.

**What I checked:**

1. `get_kb_by_id` correctly uses the broader `accessible()` check (team/shared permissions), not just ownership — this is exactly what was needed.
2. The fallback `get_or_none(id=dataset_id)` after the permission check is the right pattern.
3. Using `kb.tenant_id` for the duplicate name check is correct — validates against the dataset owner's tenant, not the requestor's.

**Small suggestion (non-blocking):** The second `get_or_none` call is technically redundant since `get_kb_by_id` already validated the KB exists. But it's safe and the extra DB round-trip in an update endpoint is negligible.

LGTM! ✅

PR #17800, opened 2026-08-04
Fix: Let delete dataset/document to a dedicated thread to avoid blocking othe APIs

wangq8:


wangq8 (commented):
## PR Review: Fix: Let delete dataset/document to a dedicated thread to avoid blocking other APIs

### Summary
This PR introduces a new `thread_pool_exec_long_time` helper that uses a module-level singleton `ThreadPoolExecutor` instead of a temporary per-call executor. The problem it solves: `thread_pool_exec` uses `with ThreadPoolExecutor(...)` which calls `shutdown(wait=True)` on exit — when a client disconnects or the request times out during a long-running delete, request cancellation unwinds the coroutine back through the `with` block, which *blocks* waiting for the worker to finish. This effectively locks the Quart request task even though we no longer need the result.

### What Changed

| File | Change |
|------|--------|
| `common/misc_utils.py` | New `thread_pool_exec_long_time` function with module-level `_LONG_TIME_THREAD_POOL_EXECUTOR` (bounded pool, default `max_workers=1`) |
| `api/apps/restful_apis/document_api.py` | `delete_documents`: `thread_pool_exec` → `thread_pool_exec_long_time` |
| `api/apps/services/dataset_api_service.py` | `delete_datasets`: `asyncio.to_thread` → `thread_pool_exec_long_time`; removes unused `import asyncio` |
| `test/.../test_dataset_api_service_list_datasets.py` | Mocks the new `thread_pool_exec_long_time` |

### Analysis

✅ **Root cause is correctly identified.** The `with ThreadPoolExecutor` pattern in `thread_pool_exec` creates a `shutdown(wait=True)` trap on request cancellation — exactly the scenario described.

✅ **Solution is sound.** Using a process-level singleton executor means request cancellation cannot trigger executor shutdown. The running sync callable continues in the background, but the Quart request task is released to serve other requests. This is the right trade-off for fire-and-forget-ish delete operations where the side effects matter more than the return value.

✅ **Bounded pool (default max_workers=1) prevents resource exhaustion.** Multiple concurrent deletes will queue instead of spawning unbounded threads. Users can increase capacity via `LONG_TIME_THREAD_POOL_WORKERS` env var if needed.

✅ **ContextVars are properly copied.** The new function copies `contextvars` into the worker thread, matching `thread_pool_exec` behavior (important for tracing/per-request state).

✅ **`asyncio` import removal is safe.** Verified: the only usage in `dataset_api_service.py` was `asyncio.to_thread(...)`, which is now replaced. No other references to `asyncio` in that file.

✅ **Test mock added.** The unit test properly mocks `thread_pool_exec_long_time` alongside the existing `thread_pool_exec` mock.

⚠️ **Callers won't see the result on disconnect.** When the client disconnects during a long delete, the `await` is cancelled but the sync function keeps running in the pool — the caller never gets the return value. This is fine for mutations (deletion side effects happen regardless), but means the API response is lost. Not a bug, just a behavioral note.

⚠️ **Module-level executor has no explicit shutdown.** Python's `atexit` handles cleanup, but if the module is reloaded (e.g., in test suites or hot-reload setups), the old executor threads could leak. This is unlikely in production but worth noting. Consider adding an `atexit.register(...)` or a `shutdown()` function for tests if needed in the future.

### Verdict
**LGTM** ✅ — well-reasoned fix with clear docstring, proper context propagation, and correctly scoped to the two call sites that need it. The trade-offs (lost response on disconnect, module-level executor lifecycle) are reasonable for this use case.

PR #17803, opened 2026-08-04
Fix: Agent version history dialog content overflows the modal

euvre:
### Summary

In the agent canvas page, the version history dialog (`VersionDialog`) could render its content outside the modal boundary: with a long version title, the right-hand preview column overflowed the `max-w-[900px]` dialog, so the card background, truncated title, download button and pagination visibly "poked through" past the dialog's right edge.

Root cause: the right column is a `flex-1` item whose default `min-width: auto` prevents it from shrinking below its content's min-content width. The version title uses `truncate` (i.e. `white-space: nowrap`), so a long title such as `no_brain_no_Annoyance_Agent-workfloat_2026-08-04 15:30:14` forces the whole column wider than the available space.

Fix: add `min-w-0` along the flex chain (right column container, title wrapper, title span) in `web/src/pages/agent/version-dialog/index.tsx` so the column can shrink and the title ellipsizes inside the dialog instead of overflowing it.


wangq8 (commented):
Thanks for the fix! The root cause analysis is spot-on — in a flex container, the default `min-width: auto` prevents shrinking below the content's min-content width, which defeats `truncate` when long titles are present.

**What works well:**
- Adding `min-w-0` along the flex chain (column container → title wrapper → title span) is the standard Tailwind approach for this class of overflow bug.
- The fix is minimal (3 `min-w-0` additions), correctly targeted, and doesn't risk regressions — elements that fit within bounds are unaffected.
- Good PR description documenting the root cause.

**Suggestions / Nit:**
- The `truncate` class already sets `overflow: hidden`, so the `min-w-0` on the span itself isn't strictly necessary *unless* the span is a direct flex child (which it is, inside the `div.flex`). Given the nesting, all three `min-w-0` additions are justified — no change needed, just noting it's correct.

LGTM, approved in spirit. 👍

euvre:
duplicate.

PR #17804, opened 2026-08-04
fix(models-section): defer auto-fetch catalog until base_url is set for self-hosted providers

dcc123456:
### Summary

fix(models-section): defer auto-fetch catalog until base_url is set for self-hosted providers

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/models-section/hooks.ts
+++ b/web/src/pages/user-setting/setting-model/instance-card/models-section/hooks.ts
@@ -86,8 +86,12 @@ export const buildModelInfo = (items: IProviderModelItem[]): IModelInfo[] =>
     extra: { is_tools: hasToolFeature(m.features), ...(m.extra ?? {}) },
   }));
 
-/** Resolved credentials for catalog / verify / batch calls. */
-export type ResolvedCreds = { apiKey: string; baseUrl: string };
+/** Resolved credentials for catalog / verify / batch calls.
+ *  `baseUrl` is `undefined` when the provider's form has no `base_url`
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Nice fix. This addresses a real UX issue where self-hosted providers (Ollama, Xinference, LocalAI) would fire a doomed catalog request before the user had a chance to enter their base URL.

**What I like:**

1. **`string | undefined` for `baseUrl`**: Clever use of `undefined` vs empty string to distinguish "form has no base_url field at all" (e.g. VolcEngine, Google Cloud) from "field exists but is empty" (Ollama, etc.). This lets the readiness gate apply only to providers that actually have the field.

2. **Readiness check moved inside the effect**: The comment about effect ordering (`useFormResetOnDetailsLoad` running before this effect) justifies why `resolveCreds()` is called inside the effect rather than during render. For saved cards where the form gets reset, reading during render would see stale (pre-reset) values. This is a subtle but important detail — thanks for documenting it.

3. **Backward compatible**: The optional `instanceDetailsLoaded` prop already exists in the `ModelsSectionProps` interface and is passed through from `saved-mode-card.tsx`. No breaking changes.

**Minor suggestions:**

1. **Type assertion `as any`**: On line 214 of hooks.ts, the `api_key: apiKey as any` cast echoes the existing pattern (the original code already had `api_key: apiKey` with implicit typing). This is consistent — no issue.

2. The deps array now includes `baseUrlValue` and `instanceDetailsLoaded` alongside `apiKeyValue`. Consider whether rapid changes to `baseUrlValue` (e.g. user typing a long URL) could cause multiple re-fetches. The `hasAutoFetchedRef.current` guard prevents duplicate fetches once the first one triggers, so this should be fine.

Approved.

PR #17805, opened 2026-08-04
Fix: Switching back and forth between the wiki template's custom options caused the Instructions to disappear.

cike8899:
### Summary

Fix: Switching back and forth between the wiki template's custom options caused the Instructions to disappear.


wangq8 (commented):
Thanks for the fix! The root cause analysis is solid: when switching to "Custom", the old code read from `form.formState.defaultValues`, which can become stale or not reflect the current template kind. The new approach of looking up the builtin template by `kind` and extracting defaults from its `example` field is the correct fix.

A few observations:

**1. Unrelated changes mixed in (minor)**
- `compilation-template-card.tsx`: The padding change (`p-4` to `py-4 px-2.5`) is purely cosmetic and unrelated to this bug fix.
- `template-configuration.tsx`: Switching from `Checkbox` to `SwitchFormField` for the plan toggle is a UI component change, also not directly related.

These are small enough that I will not block, but in general it is cleaner to keep unrelated changes in separate PRs so each one has a clear audit trail.

**2. `handlePresetChange` dependency array**
`kindPath` (which depends on `selectedTemplateIndex`) is used inside the callback body but no longer appears in the dependency array. While React Hook Form's `useWatch` on `kind` will correctly trigger a new callback when `kind` changes (and `kind` changes when `kindPath` changes), this may trigger an `exhaustive-deps` lint warning. Consider adding `kindPath` to the dependency array for correctness/clarity, or use a ref to avoid unnecessary re-creations.

**3. `ICompilationTemplateBuiltin` type safety**
The `config.example` type check (`typeof ... === "string"`) is a good defensive guard given the index signature on `ICompilationTemplateConfig`. :+1:

**4. No tests**
No existing tests were modified or new tests added. Not a blocker given the existing codebase patterns, but worth noting.

Overall: the core fix is correct and addresses the bug. Approved with the minor notes above — no need to hold this up for the cosmetic changes.

Issue #17799, opened 2026-08-04
Internal discussion: TokenChunker contract — delimiter boundary + token_size soft-target merge


xugangqiang:
> ⚠️ **Internal design discussion record.** This issue documents an internal agreement on the `TokenChunker` behavior. **External contributors: please do NOT open a pull request against this issue.**

## Purpose

This issue is a written record of the internally resolved contract for the Python `TokenChunker`
(`rag/flow/chunker/token_chunker.py`) and its merge helper `naive_merge` (`rag/nlp/__init__.py`).
It is a design reference, not a feature request for outside contributors.

## Configuration parameters

1. **`token_size`** (`chunk_token_size`) — the approximate target size of a chunk, measured in tokens.
2. **`delimiter`** — a separator. The parser's output sections are concatenated first, then split by
   `delimiter` into **paragraphs**. The delimiter text itself is **never** included in any paragraph / chunk.

## `token_size` merge semantics

After splitting by `delimiter` into paragraphs, chunks are built as follows:

- If the **current paragraph's size > `token_size`** → it is **not** merged with any other paragraph;
  it becomes its own chunk.
- If the **current paragraph's size < `token_size`** → look at `size(current) + size(next)`:
  - If `> token_size` → the current paragraph becomes a chunk (no merge).
  - If `<= token_size` → the next paragraph becomes part of the current paragraph (merge), and we continue.
- A delimiter paragraph is **never atom-split**.

## Worked example

Assume `token_size = 100`. The parser emits 10 sections; after concatenation and splitting by the
delimiter `####123` we get **4 paragraphs** (none of which contains `####123`), with token sizes
**150, 60, 50, 30**.

Merge process:

- **chunk 1** = paragraph 1 (150 > 100) → independent chunk.
- **chunk 2** = paragraph 2 (60): `60 + 50` (next) = 110 > 100 → no merge → paragraph 2 is its own chunk.
- **chunk 3** = paragraph 3 + paragraph 4: `50 + 30` = 80 <= 100 → merged.

Result: chunks of sizes **150 / 60 / 80**.

## Open discussion point (unresolved)

A secondary interpretation of the boundary case was raised during discussion (how the cursor advances
after a "no merge" decision). It was inconsistent and has **not** been adopted. The algorithm above
(the first, consistent interpretation) is the one to implement. Resolve this before coding if any
ambiguity remains.

## Relation to existing PRs

For context (this issue is the target contract, not the implementation):

[...]

PR #17808, opened 2026-08-04
Fix: delimiter is chunk boundary, drop token_size atom-split (OVER_CAP default)

xugangqiang:
## Summary

Fixes a regression introduced by #17203 (strict-cap atom-split) and a secondary delimiter-handling bug from #17723.

**Root cause:**
- #17203 added `_split_oversized_unit` / `_compute_chunk_update`, which split oversize units into ≤ token_size pieces. This collapsed `token_size=1` into 1-token chunks and set the cap at 512, mismatching the model-layer truncation boundary (embedding ~8191 / rerank 500/4096/8192/2048). Atom-split is unnecessary: oversize units stay whole and the model layer truncates.
- #17723's delimiter handling dropped consecutive delimiters (`A####B` -> `A##B`), glued JSON items with `"".join`, ignored `children_delimiters`, and stripped whitespace delimiters.

## Changes

- New pure helper `merge_paragraphs(paragraphs, token_size, strategy)` with a `MergeStrategy` enum (`UNDER_CAP` / `OVER_CAP`); **default `OVER_CAP`**. `UNDER_CAP` is a strict cap (never overflows `token_size`); `OVER_CAP` greedily accumulates adjacent paragraphs while the projected total stays within `token_size`, merging one boundary-overflow paragraph before closing. Oversize paragraphs stand alone.
- `naive_merge` / `naive_merge_with_images` / `RAGFlowTxtParser.parser_txt` now use `merge_paragraphs`; atom-split removed. `naive_merge` / `naive_merge_with_images` always split a section on the delimiter whenever one is present (even when the section already fits `token_size`), so delimiter text never leaks into a chunk. Only the empty-delimiter (size-only) mode skips splitting.
- `token_chunker`: delimiter text is dropped (not stripped); JSON flush joins buffered items with `"\n"`; `children_delimiters` and `PDF_POSITIONS_KEY` are preserved on the delimiter path. PDF positions are now attributed **per segment** — each split chunk carries only the positions of the item(s) that contributed to it — fixing a leak where page-N coordinates were attached to page-M chunks and all segments shared one preview image.
- `test_txt_parser.py` rewritten to assert the new contract (not the old strict cap); `naive_merge` and delimiter-case-sensitive matrices updated.

## Contract (refs #17799)

- user specified delimiter = chunk boundary; user specified delimiter text never enters a chunk.
- `token_size` = soft target + merge strategy; no atom-split.
- Default strategy = `OVER_CAP`; migration can switch to `UNDER_CAP` (strict cap).
- `OVER_CAP` has no hard cap; the model layer truncates oversize units. `UNDER_CAP` enforces a strict cap.

## Notes

- Closes the wrong-object revert in #17774 (revert #17723 would re-introduce delimiter-in-chunk and the strict cap).
- Go-side alignment (`internal/ingestion/component/chunker/token.go`) is a follow-up PR.

### DIFF
--- a/deepdoc/parser/txt_parser.py
+++ b/deepdoc/parser/txt_parser.py
@@ -17,8 +17,8 @@
 import logging
 import re
 
-from common.token_utils import num_tokens_from_string
 from deepdoc.parser.utils import get_text
+from rag.nlp import MergeStrategy, merge_paragraphs
 from rag.nlp.delim import (
     compile_delimiter_pattern,
     normalize_text_newlines,
@@ -35,22 +35,6 @@ class RAGFlowTxtParser:
     def parser_txt(cls, txt, chunk_token_num=128, delimiter="\n!?;。；！？", keep_delimiters=False):
         if not isinstance(txt, str):
             raise TypeError("txt type should be str!")
-        cks = [""]
-        tk_nums = [0]
-
-        def add_chunk(t):
-            nonlocal cks, tk_nums
-            tnum = num_tokens_from_string(t)
-            if tk_nums[-1] > chunk_token_num:
-                cks.append(t)
-                tk_nums.append(tnum)
-            else:
-                if cks[-1]:
-                    cks[-1] += "\n" + t
-                else:
-                    cks[-1] += t
-                tk_nums[-1] += tnum
-
         txt = normalize_text_newlines(txt)
         parsed_dels = parse_delimiter_field(delimiter)
         dels = compile_delimiter_pattern(parsed_dels)
@@ -60,6 +44,7 @@ class RAGFlowTxtParser:
             bool(dels),
         )
         secs = re.split(r"(%s)" % dels, txt) if dels else [txt]
+        paragraphs = []
         for index, sec in enumerate(secs):
             if dels and re.match(f"^{dels}$", sec):
                 continue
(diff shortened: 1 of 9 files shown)


xugangqiang:
### On the two "needs evaluation" points (chunk-text hash change; rerank/embedding truncation)

**1. Chunk-text change → existing chunk hashes / dedup / hit**
RAGFlow re-parses a document by **deleting its existing chunks before inserting the new ones** (delete-then-recreate). So for any document that is re-parsed, there are no orphaned old-format chunks to compare against — the chunk-level state is fully refreshed, and the hash change is moot within that document.

Documents that are *never* re-parsed keep the old-format chunks, but those chunks are still valid retrievable text; they simply coexist in a mixed format. There is no data corruption. Therefore **no migration script is required** — at most a release note stating "re-parse affected documents if you want a uniform format."

(The only residual edge case is global cross-document dedup during a partial re-parse, where old-format and new-format hashes of overlapping text will not collide. This is minor and transient.)

**2. rerank / embedding truncation of oversized units**
The model token limit (embedding ~8191; rerank 4096/512 depending on model) is an **inherent constant** — this PR did not introduce it; it only changes *how often* a chunk exceeds it. For normal text, chunks are tens-to-hundreds of tokens, far below these limits, so truncation essentially never happens and behavior is unchanged.

The real (but narrow) regression introduced by *removing* oversize splitting entirely: a genuinely oversized single paragraph (e.g., a PDF long unbroken line) is now kept whole and gets **silently truncated at the model boundary** during rerank, instead of being split to fit. Note the old hard-coded 512 cap was itself below real model limits (over-splitting), so "old code kept things under the limit" was true but not optimal.

A reasonable, **separate** follow-up (independent of the OVER_CAP contract) is to split *truly oversized units* at the real model upper bound — `min(embed, rerank)` or a configurable limit — rather than deleting splitting entirely. This keeps normal chunks from being over-split while preventing silent tail loss for extreme paragraphs. Happy to land this as a follow-up if we agree on the default bound.

xugangqiang:
@yuzhichang @buua436 pls help to review

yuzhichang:
LGTM

PR #17810, opened 2026-08-04
fix: migrate mistralai to 2.x and remediate CVE-2025-67221 (orjson)

rayhanest:
## Summary
  
  Migrates `mistralai` from `==0.4.2` to `>=2.7.2,<3.0.0` to unblock the orjson CVE fix. The old SDK pinned `orjson>=3.9.10,<3.11`, preventing upgrade to the patched version.
  
  | CVE | Severity | Package | Installed | Fixed in |
  |---|---|---|---|---|
  | CVE-2025-67221 | HIGH | orjson | 3.10.18 | 3.11.6 |

 `mistralai` 2.x (the current maintained version) drops the orjson dependency entirely. Added `orjson>=3.11.6` to `constraint-dependencies` to pin the floor for remaining parent packages (`langgraph-sdk`, `langsmith`, `ranx`).
  
  ## Verification
  
  - `chat.complete()` returns same response shape as before
  - `chat.stream()` returns `CompletionEvent` objects; the chunk is at `.data` which has `.choices[0].delta.content` — adjusted in code
  - `embeddings.create()` returns response with `.data[].embedding`; same shape, only param renamed `input` → `inputs`
  
  ## Testing
  
  - Full unit test suite passes 
  - Updated Mistral embedding truncation test to match new API
  - Verified all API methods present via import check

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,6 +236,10 @@ constraint-dependencies = [
     # CVE-2026-59939: httplib2 < 0.32.0 vulnerable; pulled in transitively via
     # google-api-python-client, google-auth-httplib2
     "httplib2>=0.32.0",
+    # CVE-2025-67221: orjson < 3.11.6 vulnerable; pulled in transitively via
+    # langgraph-sdk, langsmith, ranx. Previously blocked by mistralai==0.4.2
+    # pinning orjson<3.11
+    "orjson>=3.11.6",
 ]
 exclude-dependencies = [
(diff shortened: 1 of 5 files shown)


JinHai-CN:
@rayhanest Fail to pass the format check:

┃  ruff-format ❯ 
Would reformat: rag/llm/chat_model.py
1 file would be reformatted, 2 files already formatted

exit status 1

PR #17814, opened 2026-08-04
 fix(list-filter-bar): correct nested filter to prune non-matching children

dcc123456:
### Summary

 fix(list-filter-bar): correct nested filter to prune non-matching children

  `filterNestedList` kept a parent node when any of its children matched
  the search term, but the `map` pass that should have replaced the
  parent's children with the filtered subset was commented out. As a
  result, searching for a leaf value (e.g. "kk" under "name") returned the
  parent with all of its original children instead of only the matching
  ones.

### DIFF
--- a/web/src/components/list-filter-bar/filter-popover.tsx
+++ b/web/src/components/list-filter-bar/filter-popover.tsx
@@ -44,31 +44,23 @@ const filterNestedList = (
 
   const term = searchTerm.toLowerCase();
 
-  return list
-    .filter((item) => {
-      if (
-        item.label.toString().toLowerCase().includes(term) ||
-        item.id.toLowerCase().includes(term)
-      ) {
-        return true;
-      }
+  return list.reduce<FilterType[]>((acc, item) => {
+    const selfMatch =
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice fix, thanks for catching this.

**What was broken:** The `filterNestedList` function had a two-pass filter→map approach, but the `.map()` pass that was supposed to prune non-matching children was commented out. So searching for a leaf value returned the parent with all original children instead of only the matching ones.

**The fix:** The single `.reduce()` approach is cleaner and correct:
- When the parent itself matches the search term → keep full item with all children (correct UX — if "name" matches, you want everything under "name")
- When only children match → keep parent with only the filtered subset (`{ ...item, list: filteredSubList }`)

**Checks:**
- ✅ No regressions — empty search term early-return preserved
- ✅ No mutations — spread operator creates new objects for filtered parents  
- ✅ Avoids double recursion that the old code had (filter already recursed, then map would have recursed again)
- ✅ 15 additions / 23 deletions — net simplification, always a good sign

LGTM.

Issue #17371, opened 2026-07-24
Ingestion chunk map leaks index-schema shape (kb_id array and storage-suffixed fields)


xugangqiang:
## Problem

During the ingestion pipeline, chunks are represented as `map[string]any` that is bent into the exact shape of the index document before persistence. This leaks the search-engine schema backward into the ingestion domain and causes several consistency problems.

The most visible instance: in `internal/ingestion/task/chunk_process.go` (`ProcessChunksForPipeline`), `ck["kb_id"]` is written as `[]string{kbID}` even though, within ingestion, a chunk always belongs to exactly one dataset (`Doc.KbID` is a single string, and the run-level `kb_id` injected into `CanvasState.Globals` is also a single string). The array wrapping exists only to satisfy the index document contract, and both engines immediately collapse it back to a single value on write (Elasticsearch overwrites `docCopy["kb_id"] = datasetID`; Infinity's `transformChunkFields` takes the first element of the list). The same pattern repeats throughout the same function:

- Storage-type suffixes leak into ingestion field names: `docnm_kwd`, `question_kwd`, `important_kwd`, `content_with_weight`, `create_timestamp_flt`, `page_num_int` / `top_int` / `position_int`.
- String -> array wrapping for index keyword columns: `questions` -> `question_kwd` (`utility.SplitQuestions`), `keywords` -> `important_kwd` (`utility.SplitKeywords`), mirroring the `kb_id` array.
- Dual encoding of one logical value: `create_time` (string) + `create_timestamp_flt` (float); `text` renamed to `content_with_weight`.
- Semantic `positions` decomposed into index columns and the raw field dropped.
- Document-scope `metadata` mixed into the chunk map then pulled back out; parser-internal `_pdf_positions` pruned in the "pre-index" step.

These cause real inconsistency: e.g. `pipeline_real_integration_test.go` asserts `chunk["kb_id"]` is a single string while production writes `[]string`.

## Proposed direction

Keep ingestion using domain names and single-value types. Move the suffix-renaming, string-splitting, timestamp dual-encoding, and position decomposition into a single unified index-payload construction layer at the engine write boundary (Elasticsearch / Infinity `InsertChunks` / `transformChunkFields`). The `kb_id` array fix should be the first step of this broader cleanup.

## References

- `internal/ingestion/task/chunk_process.go` (`ProcessChunksForPipeline`, line ~80)
- `internal/engine/elasticsearch/chunk.go` (`InsertChunks`)
[...]

PR #17818, opened 2026-08-04
refactor(ingestion): own kb_id at the engine write boundary (drop producer stamp)

xugangqiang:
## Problem

During ingestion, `indexdoc.ProcessChunksForPipeline` stamped `ck["kb_id"]`
on every chunk. This was both:

- **a dead write** — `elasticsearch.InsertChunks` unconditionally overwrites
  the value with `datasetID` (`chunk.go:211`), so the producer's value never
  reached the index;
- **the wrong shape** — it was emitted as `[]string`, while both engines
  actually need a single string.

This is the `kb_id` slice of the ingestion -> engine schema leak tracked in
#17371: ingestion was carrying index-physical schema knowledge it should not
own.

## Fix

Make the search engines the single owner of `kb_id` at the write boundary,
and stop ingestion from emitting it:

- **Elasticsearch** (`chunk.go:211`) already sets `docCopy["kb_id"] = datasetID`
  — unchanged.
- **Infinity** (`chunk.go`) `InsertChunks` now stamps
  `insertChunks[i]["kb_id"] = datasetID` right after `transformChunkFields`
  (previously it only *read/normalized* the producer value, which forced
  ingestion to supply it). Both engines are now consistent.
- `ProcessChunksForPipeline` no longer stamps `kb_id` and the now-leaky
  `kbID` parameter is removed. The same removal is propagated to
  `ProcessPipelineOutputForGolden` and the `compare_pipeline_golden` dev tool
  (its `-kb-id` flag is dropped).

The stored `kb_id` value is byte-for-byte unchanged: `datasetID` passed to
`InsertChunks` is `taskCtx.Doc.KbID`, i.e. the same id that was previously
set on the producer chunk.

## Verification

- `bash build.sh --test ./internal/ingestion/task/indexdoc/... ./internal/engine/infinity/...`
  — both green.
- `internal/ingestion/task` has **two pre-existing** failures
  (`TestPipelineExecutor_Run_RealCanvasDSL_UsesGeneralPipeline`,
  `TestRunPipeline_RealPipelineOutput_ProducesIndexFields`) that assert
  `inserted chunk count = 1, want 2` — a parser/assertion mismatch (the Go
  parser merges the 2-paragraph fixture into 1 chunk). They are unrelated to
  this change, which never touches chunk counting. The `kb_id`-related test
  failure this change would otherwise introduce is fixed by updating the tests
  below.
- Updated the pinning unit test: `TestProcessChunksForPipeline_SetsDocID`
  (formerly `...SetsDocIDAndKBID`) now asserts `kb_id` is **not** set by the
  producer. Removed the `kb_id` assertion and the now-dead
  `taskChunkFieldEqualsStr` helper from `pipeline_real_integration_test.go`.

## Scope

This closes only the `kb_id` portion of #17371. The remaining index-physical
fields (`docnm_kwd`, `create_timestamp_flt`, `page_num_int`/`top_int`/
`position_int`, etc.) are intentionally left for a follow-up (P2).

### DIFF
--- a/internal/engine/infinity/chunk.go
+++ b/internal/engine/infinity/chunk.go
@@ -353,6 +353,10 @@ func (e *Engine) InsertChunks(ctx context.Context, chunks []map[string]interface
 	insertChunks := make([]map[string]interface{}, len(chunks))
 	for i, chunk := range chunks {
 		insertChunks[i] = transformChunkFields(chunk, embeddingCols)
+		// kb_id is owned by the engine at the write boundary (mirrors ES
+		// chunk.go InsertChunks). The ingestion producer no longer stamps it,
+		// so the producer value (if any) is intentionally overridden here.
+		insertChunks[i]["kb_id"] = datasetID
 	}
 
 	// Delete existing rows with matching IDs
--- a/internal/ingestion/task/golden_compare.go
+++ b/internal/ingestion/task/golden_compare.go
@@ -34,7 +34,6 @@ type GoldenCompareResult struct {
 func ProcessPipelineOutputForGolden(
 	pipelineOutput map[string]any,
 	docID string,
-	kbID string,
 	docName string,
 ) (GoldenCompareResult, error) {
 	normalized := indexdoc.NormalizeChunks(pipelineOutput)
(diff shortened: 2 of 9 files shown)


yuzhichang:
LGTM

PR #17826, opened 2026-08-05
Feat: Remove the pageindex option from the extraction operator.

cike8899:
### Summary

Feat: Remove the pageindex option from the extraction operator.


wangq8 (commented):
Thanks for the PR! Clean removal of the PageIndex option from the extraction operator.

**Changes look good:**
- All references are consistently removed: the `TableOfContents` enum value, i18n strings (`en.ts`, `zh.ts`), default prompts, and the `isToc` conditional rendering logic in the form component.
- The form now unconditionally renders both `systemPrompt` and `userPrompt` fields, which is correct since the TOC special-case no longer exists.
- No regressions from the diff — the refactor is purely subtractive.

**One note (non-blocking):** The backend has separate document-level TOC handling (in `rag/svr/task_executor.py`, `rag/prompts/generator.py`, etc.) for indexing — those are independent paths, not related to this extractor operator UI, so no cleanup is needed there.

✅ LGTM, approved in spirit. Feel free to merge.

PR #17827, opened 2026-08-05
Fix embed dialog language label mistranslation

euvre:
### Summary

The language selector in the agent/chat "Embed into website" dialog uses the `chat.locale` translation key as its label. This key was mistranslated as '地区' (region) in Simplified Chinese and 'Locale' in English, so the field title read as a region setting instead of a language setting in both locales.

This PR renames the label to '语言' (zh) and 'Language' (en) to match the field's actual purpose: selecting the display language of the embedded chat page.


wangq8 (commented):
Correct fix. The field selects the display language of the embedded chat page, not a region/locale, so 'Language' / '语言' is the right label. LGTM.

PR #17828, opened 2026-08-05
Fix: SIGTERM not handle well in ragflow server chat channel

wangq8:
Fix:
2026-08-05 10:26:14,741 ERROR    56710 chat channel reconcile failed: cannot schedule new futures after shutdown
2026-08-05 10:26:24,753 ERROR    56710 chat channel reconcile failed: cannot schedule new futures after shutdown

### DIFF
--- a/api/channels/bootstrap.py
+++ b/api/channels/bootstrap.py
@@ -219,13 +219,17 @@ async def _start_channel(running: dict, account_id: str, channel: str, credentia
     return True
 
 
-async def _reconcile(running: dict, failed: dict) -> None:
+async def _reconcile(running: dict, failed: dict, stop_event: threading.Event) -> None:
     """Diff desired (DB) vs running channels and apply start/stop/restart.
 
     ``failed`` remembers configs that could not be started so they are not
     retried (and re-logged) every tick until their credentials change.
     """
+    if stop_event.is_set():
+        return
     desired = await asyncio.to_thread(_desired_channels)
+    if stop_event.is_set():
+        return
 
     # Stop channels that were removed or whose credentials/type changed.
     for account_id in list(running.keys()):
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## PR Review: #17828

**Fix: SIGTERM not handle well in ragflow server chat channel**

### Summary
This PR fixes a shutdown race condition where the chat channel reconcile loop crashes with `"cannot schedule new futures after shutdown"` errors when the server receives SIGTERM. The fix adds proper stop-event plumbing and graceful thread teardown.

---

### ✅ What works well

1. **Root cause correctly identified.** The asyncio event loop in the chat channel thread was being torn down (via `sys.exit(0)`) while `_reconcile` still tried to schedule futures. Adding early `stop_event.is_set()` checks prevents wasted work and avoids the race.

2. **Clean refactoring with `stop_background_services()`.** Extracting the shutdown logic into a single reusable function (DRY) and calling it from `signal_handler`, the exception handler, *and* a new `finally` block ensures cleanup always runs.

3. **Specific exception handling.** Catching `RuntimeError` specifically (which is what asyncio raises for "cannot schedule new futures after shutdown") rather than a broad `except Exception` is the right granularity.

4. **`chat_channel_thread is not threading.current_thread()` guard.** Prevents deadlock if `stop_background_services()` is somehow called from the chat channel thread itself — good defensive programming.

---

### ⚠️ Minor concerns (non-blocking)

1. **`RuntimeError` catch in `run_channels()` is a bit too trusting.**
   If `RuntimeError` is raised for a reason *other* than shutdown (unlikely but possible), it would still be silently treated as a graceful stop info message. Consider logging the actual exception text in the info branch as well, for auditability:
   ```python
   LOGGER.info("chat channel reconcile stopped: %s", ex)
   ```

2. **`join(timeout=5)` might be tight.**
   The reconcile loop sleeps 5 seconds between iterations (`asyncio.sleep(5)`). If shutdown lands mid-sleep, the thread gets up to 5s to notice `stop_event` + at most one more DB call. The 5s join timeout should cover this, but if the DB call is slow, the thread could be abandoned unjoined. This is benign (the process is exiting anyway) but worth a comment in the code.

3. **Second `stop_event` check location in `_reconcile`.**
   Between the first check and `asyncio.to_thread(_desired_channels)`, the stop event can be set. The second check after the await covers this nicely. However, an alternative approach would be to wrap `asyncio.to_thread()` call itself in a try/except for `RuntimeError`. The current approach works fine because the `RuntimeError` is caught at the caller level, so this is just an observation.

---

### 🔍 Verdict

**LGTM, approve.** The fix is clean, minimal, and correctly addresses the shutdown race. The refactoring improves maintainability. The minor nits above are purely cosmetic / documentation-level and do not block merging.

PR #17831, opened 2026-08-05
fix: embedded/shared agent chat fails with retrieval query unmarshal error

euvre:
### Summary

The shared/embedded agent chat page (fullscreen chat and floating widget, `/agent/share`) sends the user's message in the `query` field of `POST /api/v1/agentbots/<id>/completions`. The Go handler's `AgentbotCompletionRequest` only bound `question` and `inputs`, so the free-text question was silently dropped and the raw `inputs` map became the canvas user input. The canvas then seeded `sys.query` / Begin's `query` output with that object, and the Retrieval node failed every run with:

```
canvas invoke: [NodeRunError] canvas: component "Retrieval:..." invoke: Retrieval: retrieval: parse arguments:
json: cannot unmarshal object into Go struct field retrievalArgs.query of type string
```

The in-app agent chat works because `AgentChatCompletions` binds `query` correctly.

**Fix:** mirror the Python reference (`canvas_service.completion`: `query = kwargs.get("query", "") or kwargs.get("question", "")`) and the in-app handler's form-input fallback:

- Add `Query string \`json:"query"\`` to `AgentbotCompletionRequest`.
- Derive the canvas user input via a new `agentbotUserInput()` helper: `query` first, then `question`, and only when both are empty fall back to the begin-form `inputs` map (single field lifts its `value` entry; multiple fields collapse to a name→value map), matching `extractUserInputFromFormInputs` in the in-app handler.

Verified with `bash build.sh --test ./internal/service/... ./internal/handler/...` (all green).

### DIFF
--- a/internal/service/bot.go
+++ b/internal/service/bot.go
@@ -157,19 +157,16 @@ func (s *BotService) AgentbotCompletion(
 	if _, err := s.loadCanvas(ctx, tenantID, agentID); err != nil {
 		return nil, common.CodeDataError, err
 	}
-	// Compose the canvas user input from req.UserInput (the `inputs`
-	// dict body field) plus the top-level `question`. Files remain a
-	// separate RunAgent argument so they can populate sys.files. The
-	// Python canvas_service.completion reads the same three fields
-	// separately; the previous code dropped question/files, so a body like
-	// `{"question":"hi"}` reached the canvas with empty inputs.
-	userInput := make(map[string]any, len(req.UserInput)+1)
-	for k, v := range req.UserInput {
-		userInput[k] = v
-	}
-	if req.Question != "" {
-		userInput["question"] = req.Question
-	}
+	// Compose the canvas user input the same way Python's
+	// canvas_service.completion does: `query` first (falling back to
+	// `question`), with the begin-form `inputs` dict only as a
+	// last resort when no free-text question is present. Passing the
+	// `inputs` map itself as the user input breaks the run — the
+	// canvas seeds sys.query / Begin's `query` output with it, and
+	// downstream Retrieval nodes then fail to unmarshal an object
+	// into a string field. Files remain a separate RunAgent argument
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is a great catch — the shared/embedded agent chat silently dropping `query` and then failing with a cryptic Go unmarshal error is exactly the kind of bug that frustrates users and is hard to debug from the frontend side.

**Analysis:**

- **Root cause**: The shared chat widget sends the question in `query`, but `AgentbotCompletionRequest` only bound `question` and `inputs`. The raw `inputs` map became the canvas user input, `sys.query` got an object, and Retrieval nodes choked. Clear and reproducible.
- **Fix**: Adding `Query` to the request struct with proper fallback precedence (`query` -> `question` -> form inputs) mirrors both Python `canvas_service.completion` and the in-app handler logic. ✅

**Code review:**

1. **`agentbotUserInput` precedence logic**: Matches the Python path exactly. Single-field form lifting the `value` entry and multi-field collapsing to a name-to-value map is consistent with `extractUserInputFromFormInputs`.

2. **Return type change**: The helper returns `any` instead of the previous `map[string]any`. Since `RunAgent`'s user input parameter is `interface{}`, this is fine — for a moment I wondered if there was downstream code that typed-asserted to `map[string]any`. From the diff and the test pass, the new types (string/map/nil) are all handled correctly by the canvas machinery.

3. **Test coverage**: Good to see `build.sh --test` passing for the affected packages.

- The comment block at the call site is quite long and duplicates the function docstring. Consider keeping the call-site comment brief (a single line) and relying on the function-level documentation.
- In `agentbotUserInput`, the `for _, raw := range inputs` with `len(inputs) == 1` is the right way to extract the sole element from a map in Go — just noting that a single-element map iteration is deterministic by definition here.

Nice fix, thanks for the thorough description in the PR body!

PR #17832, opened 2026-08-05
Fix: display embedding model name in list dataset api response

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -18,7 +18,7 @@ import json
 import os
 import re
 
-from api.db.joint_services.tenant_model_service import resolve_model_config, resolve_model_id
+from api.db.joint_services.tenant_model_service import resolve_model_config, resolve_model_id, get_composite_model_name_by_ids
 from common.constants import PAGERANK_FLD, LLMType
 from common import settings
 from api.db.db_models import File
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for the PR! The batch `get_composite_model_name_by_ids` is a sensible approach — it runs a constant 3 queries regardless of dataset count.

A few notes:

1. **Batch efficiency**: Good that this uses a single batch lookup instead of N individual `get_composite_model_name_by_id` calls. The `get_by_ids` pattern with instance/provider map construction is consistent with the existing codebase style.

2. **Edge case — `None` embedding model**: The list comprehension `[m["embedding_model"] for m in response_data_list]` may include `None` values if a dataset has no embedding model configured. The downstream `get_by_ids` should handle that gracefully, and the final `.get(response_data["embedding_model"], "")` defaults to empty string. Fine.

3. **Suggestion (non-blocking)**: Consider filtering out `None`/falsy `embedding_model` values before the batch call to avoid unnecessary DB lookups:

   ```python
   embed_model_ids = [m["embedding_model"] for m in response_data_list if m.get("embedding_model")]
   ```

4. **Consistency note**: The `get_composite_model_name_by_ids` silently skips unresolvable IDs while `get_composite_model_name_by_id` raises `LookupError`. This is intentional for the batch case and the docstring confirms it — just noting the difference is worth keeping in mind.

LGTM, nice addition!

PR #17834, opened 2026-08-05
fix: keep agent chat history anchored when resizing the input box

euvre:
### Summary

In the agent run/chat drawer, dragging the message input box upward grew the textarea but the chat history did not scroll along, so the input covered the latest messages. Dragging it back down moved the history as expected, making the behavior inconsistent.

This PR wraps the input in a container observed by a `ResizeObserver` and shifts the message list's `scrollTop` by the height delta, keeping the history anchored to the input box so it moves up/down together with it. The same handling also covers autosize growth while typing.

### DIFF
--- a/web/src/pages/agent/chat/box.tsx
+++ b/web/src/pages/agent/chat/box.tsx
@@ -14,7 +14,7 @@ import {
 } from '@/hooks/use-agent-request';
 import { useFetchUserInfo } from '@/hooks/use-user-setting-request';
 import { buildMessageUuidWithRole } from '@/utils/chat';
-import { memo, useCallback, useContext, useEffect } from 'react';
+import { memo, useCallback, useContext, useEffect, useRef } from 'react';
 import { AgentChatContext } from '../context';
 import DebugContent from '../debug-content';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The scroll anchoring is a nice UX improvement — the ResizeObserver approach is clean and the cleanup in the effect return is correct.

A few observations:

1. **Effect dependency**: The `useEffect` depends on `messageContainerRef`, which is a ref (stable identity), so this runs once on mount. That is correct for this use case.

2. **Edge case — task mode toggling**: When `isTaskMode` switches from `true` → `false`, the wrapper div mounts fresh and the observer is re-attached. When switching from `false` → `true`, the observer disconnects cleanly via the effect cleanup. No stale observers. ✅

3. **Minor suggestion (non-blocking)**: Consider debouncing the scroll adjustment or using `requestAnimationFrame` inside the ResizeObserver callback, since dragging the resize handle can fire rapid resize events. Not a correctness issue, just a potential for smoothness on slower machines.

LGTM otherwise, thanks for the fix!

dcc123456 (approved):

PR #17842, opened 2026-08-05
fix(knowledge-base): display embedding model friendly name in datasetlabel

dcc123456:
### Summary

fix(knowledge-base): display embedding model friendly name in datasetlabel

  Show `embedding_model_name` instead of the internal `embedding_model`
  identifier in the dataset selector label when available, falling back to
  `embedding_model` for backward compatibility. This surfaces a
  human-readable model name (e.g. "BGE-large-zh") rather than the raw
  model ID to users picking a dataset.

### DIFF
--- a/web/src/components/knowledge-base-item.tsx
+++ b/web/src/components/knowledge-base-item.tsx
@@ -97,7 +97,13 @@ export function useDisableDifferenceEmbeddingDataset(name: string) {
         suffix: (
           <section className="flex gap-2">
             <DatasetLabel text={item.nickname} />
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Clean backwards-compatible fix.

- Displays `embedding_model_name` when available, falls back to `embedding_model` — no risk of breaking existing datasets.
- `embedding_model_name` is optional (`?`) in the interface, so TypeScript will enforce null-safety at usage sites.
- Much better UX — users see "BGE-large-zh" instead of a raw model ID.

Approved.

PR #17843, opened 2026-08-05
 feat(agent-list): support multi-select canvas category filtering

dcc123456:
### Summary

feat(agent-list): support multi-select canvas category filtering

Allow users to filter agents by multiple canvas categories at once instead
  of a single category. The frontend now sends the selected category IDs as
  a comma-joined string, and the backend applies an `IN` query against the
  list.

### DIFF
--- a/api/apps/restful_apis/agent_api.py
+++ b/api/apps/restful_apis/agent_api.py
@@ -682,7 +682,7 @@ _COMPILATION_TEMPLATE_GROUP_CATEGORY = "compilation_template_group"
 @add_tenant_id_to_kwargs
 def list_agents(tenant_id):
     keywords = request.args.get("keywords", "")
-    canvas_category = request.args.get("canvas_category")
+    canvas_category_list = [item for item in request.args.get("canvas_category", "").strip().split(",") if item]
     canvas_type = request.args.get("canvas_type")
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Good feature, well-structured implementation. I reviewed the full logic flow and it looks solid. A few observations:

**Architecture** — The three-tier approach (groups-only → merge → mixed → agent-only) is clear and consistent with the existing pattern. The mixed mode properly mirrors the existing merge mode strategy (fetch all, interleave by `update_time`, paginate in Python).

**Backend**:
- `canvas_category` → `canvas_category_list` rename in the service layer makes intent clearer — good call.
- `in_()` peewee method is the right choice here.
- Mixed mode correctly passes `tags` and `canvas_type` through to `get_by_tenant_ids`.
- Empty `canvas_category_list` is handled safely (falsy guard before `.in_()` call).

**Frontend**:
- `canvasCategory` → `canvasCategoryIds` rename is consistent and descriptive.
- Joining with comma preserves backward compatibility — single-select still works as a one-element array.

**One non-blocking suggestion**: The mixed mode branch is structurally similar to merge mode (fetch all, sort, paginate). Consider extracting a shared helper in a follow-up to reduce duplication.

No regressions, no security concerns. Approved.

wangq8:
@dcc123456 fix CI.

PR #17844, opened 2026-08-05
fix(agent): bind query field in agentbot completions and clean up failed first-touch sessions

euvre:
### Summary

The embedded agent page (`/agent/share`, `POST /api/v1/agentbots/<agent_id>/completions`) sends the user message as `query`, but the Go `AgentbotCompletionRequest` only bound `question`. The silently dropped field left `userInput` as an empty dict, which:

- serialized to `{}` and was stored as the session title, and
- landed in `sys.query` as an object, crashing downstream components that consume it as a string (e.g. Retrieval: `json: cannot unmarshal object into Go struct field retrievalArgs.query of type string`).

Because the session row is created before the canvas runs and the failure path never cleaned it up, every failed exploration added a title-less `{}` conversation to the session list.

Changes:

- Accept `query` (falling back to `question`, mirroring Python `canvas_service.completion`) and pass the scalar string as the canvas user input; only form-driven runs without a free-text query fall back to the raw inputs dict.
- Derive the session title from the query/question string inside dict-shaped inputs instead of JSON-serializing the dict, and never emit `{}` for an empty dict.
- Drop the freshly-created session row when a first-touch run fails, so a failed exploration no longer inflates the session list. Interrupts (UserFillUp waits) and user-initiated cancels keep the row, as both are resumable states rather than failures.

### DIFF
--- a/internal/service/agent.go
+++ b/internal/service/agent.go
@@ -1659,6 +1659,12 @@ func (s *AgentService) RunAgent(ctx context.Context, userID, canvasID, sessionID
 		"session_id": sessionID,
 		"user_id":    userID,
 	}
+	// The session row above was created by this request (first touch);
+	// if the run then fails, the run closure drops the row again so a
+	// failed exploration never shows up in the session list.
+	if !sessionFound || newSession {
+		root["__drop_session_on_failure__"] = true
+	}
 	// The stable run id is derived from the canvas and session. It is only a
 	// checkpoint/status storage key; session_id remains the public run and
 	// cancellation identity.
@@ -1802,7 +1808,7 @@ func (s *AgentService) RunAgent(ctx context.Context, userID, canvasID, sessionID
 // compose.WithInterruptBeforeNodes) resumes and reads the user's
 // follow-up via compose.GetResumeContext.
 func (s *AgentService) buildRunFunc(canvasID string, versionRow *entity.UserCanvasVersion, dsl map[string]any) canvas.RunFunc {
-	return func(ctx context.Context, root map[string]any) (*canvas.CanvasState, error) {
+	return func(ctx context.Context, root map[string]any) (runState *canvas.CanvasState, runErr error) {
 		if err := ctx.Err(); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice fix, well thought out.

**What I verified:**

1. **Query binding**: Confirmed the frontend sends `query` (params.query = message.content in use-send-agent-message.ts). The PR correctly adds `Query` to the Go request struct and passes the scalar string as `userInput` — matching what the canvas expects for `sys.query`. The fallback to `question` preserves backward-compat with older clients.

2. **Session title**: The new `deriveAgentSessionName` logic is correct — extracting query/question from dict-shaped inputs avoids the previous `{}` and `{"query":...}` noise in the session list.

3. **Failed session cleanup**: The `__drop_session_on_failure__` mechanism is well-designed:
   - Only set on first-touch sessions (correct)
   - Excludes interrupts and cancellations (resumable states, not failures)
   - Uses `context.WithoutCancel` with a 5s timeout for cleanup (safe, won't block)
   - Nil-safe guard on `api4ConversationDAO` and `dao.DB`
   - `canvas.IsInterruptError` exists at internal/agent/canvas/interrupt_resume.go:201

**One observation (non-blocking):** The `userInput` type change from always-`map[string]any` to `string | map[string]any` is a behavioral shift. The `RunAgent` function already accepts `any`, and the canvas handles scalar `sys.query` correctly. But any downstream code that assumes `userInput` is always a map when read back from `root` could break — just something to keep in mind.

Overall, this is a solid fix. LGTM 🚀

wangq8:
resolve conflict. - @euvre

wangq8:
resolve conflict. - @euvre

euvre:
Thanks for the reviews! Summary of how each item was handled:

**1. "resolve conflict" (x2, @wangq8)** — Done. The branch has been rebased onto the latest `main` (9357f79e); the PR now reports as mergeable with no conflicts. The two remaining Go integration checks are still running on CI.

**2. Review observation (`userInput` type `string | map[string]any`)** — Acknowledged, and it is safe as merged. The scalar-or-map shape mirrors the Python contract in `canvas_service.completion` (`query = kwargs.get("query", "") or kwargs.get("question", "")`, with the begin-form dict as the fallback), and `RunAgent`/canvas already accept `any`. The code added in this PR reads user input defensively: `deriveAgentSessionName` handles both the scalar and dict shapes, and the `__drop_session_on_failure__` cleanup only touches its own `root` keys — nothing downstream assumes a map.

**3. CodeRabbit nitpick (stale `bot.go` comment)** — Obsolete: it targeted the previous revision (97ede000). The query-binding change has since landed on `main` via #17831, so this PR no longer touches `bot.go`, and the comment there now correctly documents the scalar query/question path plus the form-input fallback.

Validation: `bash build.sh --test ./internal/service/...` — all 8 packages ok, no failures. No new code changes this round, so no shared-service E2E was run from this environment.

PR #17847, opened 2026-08-05
Fix: Add an icon to the tag on the agent card to distinguish it from the wiki card type.

cike8899:
### Summary

Fix: Add an icon to the tag on the agent card to distinguish it from the wiki card type.


wangq8 (commented):
LGTM. Simple and clean visual improvement — adding a `Tag` icon to each agent card badge makes it visually distinct from wiki card badges. No regressions, no logic changes.

Approved.

PR #17849, opened 2026-08-05
fix: stream agentic reasoning and answers correctly

buua436:

Agentic RAG streaming could lose model reasoning, emit incorrect or repeated thinking markers, and return the final answer as an aggregated response instead of preserving its original streaming deltas.

This PR:

- emits a single `start_to_think` and `end_to_think` pair;
- preserves reasoning from both the outer model and final-answer model;
- streams final-answer deltas without duplicating the complete answer in the final event;
- separates `<think>` content from visible answer content;
- uses the configured `Empty response` directly when no supporting evidence is retrieved, while allowing the model to respond when that setting is empty.

### Type of change

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -1983,8 +1983,9 @@ async def rag_agent(dialog, messages, stream=True, **kwargs):
     if getattr(chat_mdl, "mdl", None) is not None:
         chat_mdl.mdl.terminal_tools = {"rag"}
     if stream:
-        # Surface the agentic pipeline's bracket-tagged progress logs to the
-        # client as <think> content, interleaved with the real token stream.
+        # Surface only the agentic pipeline's progress logs as one think block.
+        # The outer model's tool-selection thoughts are implementation details;
+        # the final answer is emitted from the inner graph's own deltas below.
         from rag.advanced_rag.think_log import install_think_log_handler, set_think_log_sink, reset_think_log_sink
 
         install_think_log_handler()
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this PR! The approach of using a single think block, preserving inner-graph deltas via `answer_sink`, and making `empty_response` configurable is the right direction. I have a few concerns:

### 1. Fragile string matching (`"Running the "`)

In `dialog_service.py`, the check `if "Running the " in value` to detect when the terminal tool starts depends on the orchestrator log format staying exactly the same. If the log wording changes, the outer-model text after the tool call would leak into the stream. Consider using a more robust signal — e.g., a dedicated event type from the graph or a state flag — instead of matching log text.

### 2. Non-string token handling regression in `_split_think_stream`

The old `_strip_think_stream` yielded non-string tokens (`yield token`), but the new `_split_think_stream` silently drops them (`continue`). Currently this is fine because the graph only produces strings, but it is a behavior regression that could bite if non-string sentinels are ever introduced. Consider yielding them as-is or at least logging a warning.

### 3. Missing tests

The streaming event loop, `_split_think_stream`, `answer_sink` integration, and the `empty_response` fallback in `formalize_answer` don not have test coverage. At minimum, `_split_think_stream` should have unit tests for:
- Standard `<think>...</think>` spans
- Providers that emit `</think>` on every subsequent delta
- Partial tag boundaries at chunk splits
- Empty/no-think streams

### 4. Minor: `re.sub` on `inner_think` and outer think text

Stripping `</?think>` from deltas is a reasonable heuristic, but it would also strip those literal strings from legitimate visible content. A comment acknowledging the trade-off would help future readers.

### 5. `loop.call_soon_threadsafe` for `stream_done`

This is a solid fix — the original `put_nowait` could be called from a non-event-loop thread and crash. :+1:

---

Overall this is well-structured. The main ask is the fragile string matching and some test coverage. Happy to re-review after those are addressed.

PR #17854, opened 2026-08-05
fix(chunker): keep oversize text/markdown unit whole (OVER_CAP alignment)

xugangqiang:
## What

The Go `TokenChunker` text/markdown path (`mergeByTokenSize`) unconditionally
called `splitOversizedUnit` on any unit that exceeded `chunk_token_size`,
emitting Go-only sub-chunks. Python's `naive_merge` (`_merge_paragraph_groups`,
`rag/nlp/__init__.py`) never atom-splits an oversize unit under either
`OVER_CAP` or `UNDER_CAP`: a paragraph larger than the budget becomes its own
standalone chunk and the model layer truncates it later.

This aligns the text/markdown path with the **structured JSON path**
(`invokeJSONPayload` → `mergeByTokenSizeFromJSON(..., subSplitOversize=false)`,
#17739). It completes the OVER_CAP alignment started in #17835.

## Cases fixed (chunker parity)

- `token__text_long_paragraph`: Python 1 chunk, Go was 3 → now 1.
- `token__markdown_long`: Python 1 chunk, Go was 4 → now 1.

## How

An oversize unit is passed through `addChunk`, which `mergeDecision` already
forces to `startNewChunk` (token.go:514) when the incoming unit exceeds the
target, so it stands alone exactly like Python. No atom-split on the
text/markdown path. The `addUnit` wrapper was inlined into `addChunk`.

## Note

This is one step of the broader chunker-parity alignment. Remaining divergences
are tracked separately:

- Go↔Python tokenizer token-count matching (a different workstream).
- Delimiter-retention on the `json_children_delim` path.
- The `subSplitOversize=true` atom-split branch is **dead code** in production,
  not a live divergence. On the text/markdown path a backtick-wrapped delimiter
  short-circuits at `hasCustomDelim` (token.go:330) and returns one chunk per
  segment (matching Python `naive_merge`), while a bare delimiter
  (`\n!?。；！？`) early-returns to `mergeByTokenSize` (token.go:306). The only
  caller passing `true` (token.go:337) is therefore unreachable, and the live
  JSON path passes `false` (token.go:719, #17739). Removing the dead branch is
  tracked separately as a hygiene cleanup (#17865), not a parity bug. The fix
  above applies to the common (bare-delimiter) case.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -544,10 +544,13 @@ func newChunkText(prevText, incoming string, target int, overlapPct float64, inc
 // mergeByTokenSize implements exact token-based chunk merging that mirrors
 // Python's naive_merge (rag/nlp/__init__.py) after the strict chunk_token_num
 // hard-cap fix. It uses tokenizeStr for precise token counting, treats the
-// payload as a single section, splits oversized sections on production sentence
-// delimiters, hard-caps atomic oversize units via splitOversizedUnit, and merges
-// only when the projected total stays within chunk_token_size. Overlap is
-// applied only when the resulting chunk still fits the budget.
+// payload as a single section, and splits oversized sections on production
+// sentence delimiters. An oversize unit (a single paragraph larger than the
+// token budget) is kept whole as a standalone chunk — matching Python OVER_CAP,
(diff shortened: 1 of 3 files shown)


yuzhichang:
LGTM

PR #17856, opened 2026-08-05
Fix(parser): keep real line breaks when merging HTML fragments and PDF boxes

xugangqiang:
## Summary

The previous attempt added `<br>` → `"\n"` preservation but introduced real
regressions and shipped a broken `<br>` feature. This revision keeps the
intended `<br>` preservation while fixing the breakage.

Problems with the prior attempt (now fixed):
- `<br>` immediately before an **inline tag** (e.g. `<p>Line1<br><span>Line2</span></p>`)
  silently dropped the line break — the `pending_newline` flag was reset on
  entry into every `Tag` and ignored the inbound value. Now a pending
  **line-break count** is threaded through, so the break survives nesting.
- `<br><br>` collapsed to a single `"\n"`. The count now preserves both breaks
  (`"\n\n"`).
- Inline/loose fragments were joined **verbatim with no separator**, which
  fused adjacent runs (`<p>Only.</p>tail` → `Only.tail`) and leaked source
  formatting whitespace as literal newlines (`<p>\n  Hello\n  <b>World</b>\n</p>`
  → `Hello\n  World`), over-splitting the paragraph.
- English inline runs without source whitespace fused (`Hello<b>World</b>` →
  `HelloWorld`).

Fix:
- `deepdoc/parser/html_parser.py` `read_text_recursively` tracks a pending
  break **count** (incremented per `<br>`, reset only after a descendant emits
  text) and strips each text node so formatting whitespace never becomes a
  line break.
- `merge_block_text` joins inline/loose fragments with a **single space** and
  inserts `"\n" * count` only at explicit `<br>` boundaries.

Net effect: inline prose stays on one line (`Hello World`), real `<br>`
boundaries survive (including before tags and repeated breaks), and source
formatting whitespace no longer over-splits.

## Test plan

- `test_html_parser.py` rewritten/added to lock the corrected semantics:
  - inline fragments sharing a block_id join with one space (no spurious `"\n"`);
  - a `<br>`-tagged fragment inserts `"\n"`; `<br>` before an inline tag too;
  - `<br><br>` keeps two newlines;
  - loose (`block_id is None`) fragments join with a space, `<br>` still forces `"\n"`;
  - indented HTML stays one line (no leaked `"\n"`);
  - inline runs without source whitespace keep a space;
  - end-to-end `parser_txt` cases for `<br>`, `<br>`+inline tag, double `<br>`,
    indented HTML, and headings.
- The duplicated HTML semantics test table has been consolidated into a single
  shared JSON fixture (`internal/parser/parser/testdata/unified_html_cases.json`)
  — embedded by the Go test via `//go:embed` and loaded from repo root by the
  Python test — so the two engines can no longer drift.

## Checklist

- [x] Unit tests added / updated
- [x] No change to public API surface

### DIFF
--- a/deepdoc/parser/html_parser.py
+++ b/deepdoc/parser/html_parser.py
@@ -105,27 +105,31 @@ class RAGFlowHtmlParser:
         return table_str_list
 
     @classmethod
-    def read_text_recursively(cls, element, parser_result, chunk_token_num=512, parent_name=None, block_id=None):
+    def read_text_recursively(cls, element, parser_result, chunk_token_num=512, parent_name=None, block_id=None, newline_before=0):
         if isinstance(element, NavigableString):
-            content = element.strip()
+            # Keep the text verbatim (do not strip): source whitespace is
+            # folded later by merge_block_text, which preserves <br> breaks.
+            content = str(element)
 
-            def is_valid_html(content):
+            def is_valid_html(text):
                 try:
-                    soup = BeautifulSoup(content, "html.parser")
+                    soup = BeautifulSoup(text, "html.parser")
                     return bool(soup.find())
                 except Exception:
                     return False
 
             return_info = []
-            if content:
+            if content.strip():
                 if is_valid_html(content):
                     soup = BeautifulSoup(content, "html.parser")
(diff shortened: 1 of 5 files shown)


xugangqiang:
@yuzhichang @JinHai-CN @yingfeng pls help to review

PR #17859, opened 2026-08-05
fix: use updated RAPTOR options for skill generation

buua436:

Skill generation still passed the removed `threshold` argument to the RAPTOR constructor, causing a runtime `TypeError`.

This PR updates skill generation to use the current RAPTOR options and minimum supported token limit.

### Type of change

### DIFF
--- a/rag/svr/task_executor_refactor/dataset_skill_generator.py
+++ b/rag/svr/task_executor_refactor/dataset_skill_generator.py
@@ -366,9 +366,10 @@ async def run_corpus2skill(
         llm_model=chat_mdl,
         embd_model=embedding_model,
         prompt="Please write a concise summary of the following texts:\n{cluster_content}",
-        max_token=256,
-        threshold=0.1,
+        max_token=512,
         max_errors=3,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Good catch! The RAPTOR constructor was refactored to use `clustering_threshold` + `clustering_ratio` instead of the old `threshold` parameter, and this call site was missed — passing `threshold=0.1` would indeed cause a `TypeError` at runtime.

**Verified against the RAPTOR constructor signature:**
```python
def __init__(self, max_cluster, llm_model, embd_model, prompt,
             max_token=512, small_layer_collapse=8, max_errors=3,
             clustering_threshold=0.3, clustering_ratio=0.5):
```

- `threshold=0.1` removed ✅ (no longer exists, would TypeError)
- `clustering_threshold=0.3` ✅ (matches default)
- `clustering_ratio=0.5` ✅ (matches default)
- `max_token=512` updated from 256 ✅ (matches default)

Since the values match the constructor defaults, you could alternatively just drop them and rely on defaults — but being explicit is fine too.

LGTM 👍

PR #17866, opened 2026-08-05
 Fix: applying model config in one multi-chat card no longer overrides sibling cards

dcc123456:
### Summary

Fix: applying model config in one multi-chat card no longer overrides sibling cards

  In the multi-model chat view, each ChatCard holds its own `useForm`
  instance so users can compare different models side by side. However,
  the `useLayoutEffect` that synced `llm_id` from `currentDialog` re-ran
  whenever `currentDialog.llm_id` changed. Since `currentDialog` is shared
  across all cards, clicking "Apply" in one card triggered `patchChat` ->
  `invalidateQueries([FetchChat])` -> refetch, and the resulting
  `currentDialog.llm_id` update then reset every other card's `llm_id`
  form value to the patched value, destroying the per-card model selection.

  Fix: gate the sync with a `syncedDialogIdRef` so each card only picks up
  `currentDialog.llm_id` once per `dialogId` (initial load or conversation
  switch). Subsequent `currentDialog.llm_id` changes driven by Apply no
  longer leak into sibling cards.

  Also pass `tenant_llm_id` alongside `llm_id` when patching the dialog so
  the tenant-level model binding stays consistent with the applied config.

### DIFF
--- a/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
+++ b/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
@@ -138,7 +138,13 @@ const ChatCard = forwardRef(function ChatCard(
   // resend with the card's model settings (llm_id, temperature, ...).
   const sendCardMessage = useCallback(
     ({ message, messages }: { message: IMessage; messages?: IMessage[] }) =>
-      sendMessage({ message, messages, ...form.getValues(), storeHistoryMessages: false, omitSessionId: true }),
+      sendMessage({
+        message,
+        messages,
+        ...form.getValues(),
+        storeHistoryMessages: false,
+        omitSessionId: true,
+      }),
     [sendMessage, form],
   );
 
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix! This is a clean solution to a real pain point in the multi-chat comparison flow. The `syncedDialogIdRef` guard is the right pattern here — it keeps the initial sync while preventing sibling-overwrite leaks.

**What I checked:**
- The `useLayoutEffect` dependency change is correct: the effect only re-runs when `dialogId` actually changes (new conversation), not when `currentDialog.llm_id` changes from a sibling Apply.
- The `tenant_llm_id` passthrough in `handleApplyConfig` makes sense — keeps the tenant-level model binding consistent with the applied config.

If `dialogId` can ever be `undefined` when the card first mounts and later resolves to a string, the ref guard would still sync correctly on the first non-undefined value. But if `dialogId` is always defined by the time the card renders (which appears to be the case from the `dialogId!` usage elsewhere), this is fine as-is.

LGTM 👍

PR #17867, opened 2026-08-05
Fix: The hover card referenced in the agent's floating dialog box is not fully displayed.

cike8899:
### Summary

Fix: The hover card referenced in the agent's floating dialog box is not fully displayed.


wangq8 (commented):
Clean fix. The hover card content was being clipped because it rendered inside the floating dialog container which has `overflow: hidden` or constrained dimensions. Wrapping it in `<HoverCardPrimitive.Portal>` teleports the content to the document body, bypassing the parent clipping.

- `collisionPadding={8}` ensures the popover doesn not sit flush against viewport edges — good UX touch.
- The `max-h-[var(--radix-hover-card-content-available-height)]` class uses Radix built-in CSS variable to cap the popover height to available viewport space, preventing it from extending beyond the screen.
- Only one import and one component wrapping change — minimal surface area, low risk.

LGTM. Approved.

PR #17871, opened 2026-08-05
fix: filter wiki artifacts by keywords

buua436:

Enable keyword filtering for Wiki artifacts and topics.

- Filter Wiki pages by title, slug, and summary.
- Return topics when either the topic itself or one of its child pages matches.
- Preserve pagination and total counts.
- Keep the existing fast path unchanged when no keywords are provided.

### Type of change

### DIFF
--- a/api/apps/restful_apis/dataset_api.py
+++ b/api/apps/restful_apis/dataset_api.py
@@ -602,7 +602,7 @@ async def has_any_wiki(tenant_id, dataset_id):
 async def list_wiki_pages(tenant_id, dataset_id):
     """List artifact pages for the dataset Artifact tab.
 
-    GET /api/v1/datasets/<dataset_id>/artifacts?page=1&page_size=200&page_type=entity&topic=topic
+    GET /api/v1/datasets/<dataset_id>/artifacts?page=1&page_size=200&page_type=entity&topic=topic&keywords=query
     Success: {"code": 0, "data": {"total": int, "items": [{slug, title, page_type}]}}
     """
     try:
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for this PR. The keyword filtering for wiki artifacts is a useful addition. Here is my review:

**What works well:**
- Extracting the `_to_item` helper reduces duplication and makes the code cleaner.
- Using `casefold()` for case-insensitive matching is the right choice for search.
- The `matching_topics` set properly deduplicates when a topic matches directly and also via a child page.
- Error handling with `logging.exception` is preserved and extended to the new code paths.
- The approach of batch-scanning (1000 per batch) only when keywords are provided keeps the existing fast path unchanged.

**Concerns / Suggestions:**

1. **PR description mismatch**: The PR body mentions "Skill generation still used removed RAPTOR constructor arguments, causing a runtime TypeError", but the diff contains no RAPTOR-related changes. Please update the description to accurately describe what this PR does.

2. **Performance for large datasets**: The keyword branch scans all wiki pages with batching when keywords are provided. For datasets with tens of thousands of wiki pages this could be slow. Consider whether full-text search via the document store (if available) would be more suitable, or add a note that this path intentionally uses metadata-only matching for backend-agnostic compatibility (as your code comment already indicates).

3. **`list_wiki_topics` child search**: The child-page keyword lookup filters `page_type_kwd: ["concept", "entity"]` while the main aggregation uses `["topic"]`. This is intentional and correct, but worth a brief code comment explaining why — topics are containers, only concept/entity children carry the actual content worth matching.

4. **Minor**: In `list_wiki_topics`, the `_scalar` import should already be available in scope. Confirmed it is — no issue.

Overall the implementation is solid and the approach is well-reasoned. Approved with the suggestion to fix the PR description.

PR #17874, opened 2026-08-05
fix(email): close SMTP after send failures

RerankerGuo:
### Review scope

- Production diff: `web_utils.py` (`+14/-3`).
- Regression evidence: `test_web_utils_email.py` (`+125/-0`, 3 failure-precedence scenarios).
- The cleanup change and failure-injection tests are intentionally atomic; splitting them would remove the proof for exception precedence.

### Summary

Close the SMTP connection when email authentication or delivery fails.

`send_email_html()` previously called `quit()` only after a successful login and send. An exception from either operation bypassed cleanup, leaving the connection open until the SMTP client or transport was eventually collected.

The login and send operations now run inside `try/finally`, preserving the original exception while always closing a successfully connected client.

### Verification

- Added failure-injection tests for both SMTP login and message delivery.
- Before the fix, the login failure path recorded only `connect, login`; the expected `quit` call was missing and the test failed.
- After the fix, both failure paths call `quit()` exactly once and preserve their original exceptions.
- Targeted tests: `2 passed`.
- New-test Ruff, source/test format, and `git diff --check`: passed.

### Scope

SMTP configuration, message construction, and successful delivery behavior are unchanged. This only adds deterministic cleanup after `connect()` succeeds.

### DIFF
--- a/api/utils/web_utils.py
+++ b/api/utils/web_utils.py
@@ -195,20 +195,17 @@ async def send_email_html(to_email: str, subject: str, template_key: str, **cont
     )
 
     await smtp.connect()
-    operation_error = None
     try:
         await smtp.login(settings.MAIL_USERNAME, settings.MAIL_PASSWORD)
         await smtp.send_message(msg)
-    except BaseException as exc:
-        operation_error = exc
-        raise
-    finally:
+    except Exception:
         try:
             await smtp.quit()
         except Exception:
-            if operation_error is None:
-                raise
-            logging.exception("Failed to close SMTP connection after email operation error")
+            logging.exception("Failed to close SMTP connection after email error")
+        raise
+    else:
+        await smtp.quit()
 
 
 async def send_invite_email(to_email, invite_url, tenant_id, inviter):
--- a/api/utils/web_utils.py
+++ b/api/utils/web_utils.py
@@ -195,11 +195,20 @@ async def send_email_html(to_email: str, subject: str, template_key: str, **cont
     )
 
     await smtp.connect()
+    operation_error = None
     try:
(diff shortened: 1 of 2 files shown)


api/utils/web_utils.py
```diff
@@ -195,9 +195,20 @@ async def send_email_html(to_email: str, subject: str, template_key: str, **cont
...
+    except BaseException as exc:
+        operation_error = exc
+        raise
+    finally:
+        try:
+            await smtp.quit()
+        except Exception:
+            if operation_error is None:
+                raise
+            logging.exception("Failed to close SMTP connection after email operation error")
```
JinHai-CN: try:
        await smtp.login(settings.MAIL_USERNAME, settings.MAIL_PASSWORD)
        await smtp.send_message(msg)
    except Exception:
        try:
            await smtp.quit()
        except Exception as close_error:
            logging.exception(f"Failed to close SMTP connection after email error: {close_error}")
        raise
    else:
        await smtp.quit()

JinHai-CN (changes requested):

api/utils/web_utils.py
RerankerGuo: Addressed in 83543450d. The SMTP flow now follows the suggested try/except/else structure: login or delivery failures trigger best-effort shutdown without masking the original exception, while shutdown failures after successful delivery still propagate. The tests now assert the complete event sequence for login failure, send failure, combined operation/shutdown failure, and successful delivery followed by shutdown failure. Verification: `uv run pytest test/unit_test/api/utils/test_web_utils_email.py -q` (5 passed); Ruff findings are unchanged from the branch baseline (source 8 -> 8, tests 0 -> 0); format and diff checks pass.

PR #17879, opened 2026-08-05
fix: enhance agent tags display with tooltip for better visibility

dcc123456:
### Summary

fix: enhance agent tags display with tooltip for better visibility

### DIFF
--- a/web/src/pages/agents/agent-card.tsx
+++ b/web/src/pages/agents/agent-card.tsx
@@ -3,12 +3,18 @@ import { MoreButton } from '@/components/more-button';
 import { SharedBadge } from '@/components/shared-badge';
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
 import { AgentCategory } from '@/constants/agent';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the PR! The idea of clamping overflowing agent tags and showing a tooltip is good. However, I have a concern about the implementation:

**`line-clamp-2` breaks horizontal tag wrapping**

The `line-clamp-2` utility applies `display: -webkit-box; -webkit-box-orient: vertical`, which overrides the previous `flex flex-wrap` layout. Since `Badge` is a block-level `<div>`, each badge will stack vertically (one per line) instead of flowing horizontally and wrapping naturally. The intent was to limit to 2 _horizontal_ lines of tags, but the current implementation will cause vertical stacking.

**Suggestions:**

1. Consider using a CSS approach that keeps horizontal flow intact — e.g., `max-height` with `overflow: hidden`, or a JS-based approach measuring the actual rendered height of a flex-wrap container.
2. Wrap `handleOpenChange` in `useCallback` to avoid unnecessary re-renders.
3. The tooltip duplicates the full tag list inside `<TooltipContent>` — consider extracting the tag list into a reusable variable to avoid duplication (minor DRY concern).

Would appreciate it if you could verify the visual rendering with tags that span multiple lines.

PR #17880, opened 2026-08-05
Fix: The download button is displayed on the agent dialog page shared externally.

cike8899:
### Summary

Fix: The download button is displayed on the agent dialog page shared externally.


wangq8 (commented):
The change is minimal and straightforward — removing the `!isShare` guard allows the download button to appear in shared agent dialog views. If the attachment is already being displayed in the shared dialog, allowing download is consistent UX.

One question to consider: was there an intentional access-control reason for hiding the download button in shared mode originally? If not, LGTM.

PR #17883, opened 2026-08-05
  fix(agent): make global variable form fields reactive to i18n language changes

dcc123456:
### Summary

  fix(agent): make global variable form fields reactive to i18n language changes

  Convert `GlobalFormFields` from a static array (which captured translations
  at module-load time via the bare `t()` import) into a function that accepts
  a `TFunction` and is invoked at render time with the `t` from `useTranslation`.

### DIFF
--- a/web/src/pages/agent/gobal-variable-sheet/constant.ts
+++ b/web/src/pages/agent/gobal-variable-sheet/constant.ts
@@ -1,5 +1,5 @@
 import { FormFieldConfig, FormFieldType } from '@/components/dynamic-form';
-import { t } from 'i18next';
+import { TFunction } from 'i18next';
 import { TypesWithArray } from '../constant';
 import { buildConversationVariableSelectOptions } from '../utils';
 export { TypesWithArray } from '../constant';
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
## Review: PR #17883

**Summary:** Good fix for a real i18n bug. Converting `GlobalFormFields` from a static module-level constant to a function that accepts a runtime `t` makes translations reactive to language changes. Previously, `t()` was called at module-load time, freezing the translations to whatever language was active at initial load.

### ✅ What looks good

1. **Correct approach** — The core issue is well understood and the fix is minimal and targeted.
2. **All callers updated** — Both usages in `index.tsx` (useState initializer and onClick handler) now pass `t` from useTranslation.
3. **No other callers** — `GlobalFormFields` is only used in these two files, so there are no missed call sites.
4. **buildConversationVariableSelectOptions() is safe** — It does not use translations, so no similar bug there.

### 💡 Minor suggestions (non-blocking)

1. **Redundant import** — The existing `import { t } from "i18next"` at the top of `index.tsx` is now shadowed inside the component by `const { t } = useTranslation()`. The bare `t` import is no longer needed in this file and could be removed as part of this cleanup, since all `t()` calls in the component body will now use the hook-provided `t`.

2. **State staleness consideration** — The `fields` state is initialized with `GlobalFormFields(t)` on mount and updated in the onClick handler. If a user has the sheet open when switching languages, the field labels in the modal won't update until the modal is closed and reopened. This is acceptable for typical usage, but worth being aware of.

### Verdict

✅ **Approve** — The fix is correct, minimal, and resolves the reported issue. The suggestions above are optional improvements that can be addressed in a follow-up if desired.

wangq8 (approved):
LGTM — clean fix for the i18n reactivity issue.

PR #17891, opened 2026-08-05
fix: split wiki template instruction and example

buua436:

Separate Wiki template instructions from page examples across the backend, frontend, preset files, and Wiki generation flow.

Also configure the raw Markdown editor to use the local Monaco loader consistently.

### Type of change

### DIFF
--- a/api/apps/restful_apis/compilation_template_api.py
+++ b/api/apps/restful_apis/compilation_template_api.py
@@ -61,7 +61,7 @@ def list_wiki_presets() -> Response:
     ``api/db/init_data/compilation_templates/wiki/*.yaml``.
 
     Each entry carries ``id`` (filename stem) + ``topic`` +
-    ``instruction`` + ``page_example`` so the artifact-template editor
+    ``instruction`` + ``example`` so the artifact-template editor
     can pre-fill its "Page-structure example" / "Global rules" fields
     from a canned skeleton. Filesystem-fresh per request; no DB seed.
     """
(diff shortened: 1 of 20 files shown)


wangq8 (commented):
Thanks for this PR! The separation of `instruction` and `example` (formerly `page_example`) is a clean improvement — it makes the data model more accurate and avoids the need for the `splitExampleToBlueprintFields` hack in the frontend.

## What looks good

- **Consistent field rename** across YAML presets, backend service, and all frontend components — `page_example` → `example` everywhere.
- **Prompt template** now has distinct "Additional writing instructions" and "Page structure example" sections, giving the LLM clearer guidance.
- **Backward compatibility**: `_wiki_write_page_simple` retains the `example` fallback, and `wiki_refine_from_plan` uses `page_example or example` so existing templates with only the old combined field still work.
- **Serializer changes**: `transformTemplateToPayload` now sends `instruction` and `example` separately instead of joining them — matches the new data model cleanly.

## Minor observations (non-blocking)

1. **Parameter naming in `_build_refine_writer_system`**: The second parameter is named `page_example` while the YAML field and everywhere else uses `example`. Not a bug, but slightly inconsistent naming.

2. **Default `example` in `wiki.yaml`**: The new `example` field is `"Page structure could be as following: (Not provided)"` — this is quite bare. The original content was correctly moved to `instruction`, so the fallback `WIKI_TEMPLATE_EXAMPLE` will be used at runtime. Might be worth either removing the placeholder entirely or providing a meaningful default.

3. **`WIKI_REFINE_WRITER_SYSTEM` init**: `_build_refine_writer_system(None)` now produces the generic instruction fallback `"Follow the page structure and writing requirements below."` instead of directly embedding `WIKI_TEMPLATE_EXAMPLE` — functionally equivalent but slightly more verbose in the default prompt. Fine in practice.

No regressions or security concerns. LGTM, approved. 👍

PR #17893, opened 2026-08-05
fix: clean up wiki blueprint default instruction text

euvre:
### Summary

The default instruction text for the custom wiki blueprint (operator-wiki) contained a stray markdown heading marker (`##`) in the middle of a sentence and a literal `\n` (in YAML block scalars it is not escaped, so it was rendered verbatim in the UI):

```
5. ## See also section at the end with wikilinks to highly related pages(less than 12).\n
```

This PR fixes the line to read cleanly in all three places where the default text lives:

- `api/db/init_data/compilation_templates/wiki/general.yaml` — the preset served (filesystem-fresh) to the frontend for the Custom blueprint
- `api/db/init_data/compilation_templates/wiki.yaml` — the built-in wiki template `example`
- `rag/advanced_rag/knowlege_compile/wiki.py` — the `WIKI_TEMPLATE_EXAMPLE` fallback

Fixed text:

```
5. End with a "## See also" section listing wikilinks to highly related pages (less than 12).
```

Also normalized the missing space before the parenthesis.

### DIFF
--- a/api/db/init_data/compilation_templates/wiki.yaml
+++ b/api/db/init_data/compilation_templates/wiki.yaml
@@ -9,7 +9,7 @@ config:
       - 2. Sections with H2 headings, each starting with prose before sub-bullets.
       - 3. Bold key terms on first use; link them with [[ ]] wikilinks.
       - 4. Examples or implications where the source provides them.
-      - 5. ## See also section at the end with wikilinks to highly related pages(less than 12).\n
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Straightforward text cleanup — fixes the stray `##` markdown heading marker and the literal `\n` that were rendering verbatim in the UI. Consistent fix across all three locations (two YAML templates and the Python fallback). LGTM.

PR #17901, opened 2026-08-05
docs: annotate metadata_condition parameters in OpenAI-compatible chat completion API

vanzei:
### Summary

Addresses part of #12751 — the HTTP API reference lists request parameters but leaves some nested objects under-annotated, so callers have to reverse-engineer them from the example payload.

The **Create chat completion** endpoint (`POST /api/v1/openai/{chat_id}/chat/completions`) accepts a rich `extra_body.metadata_condition` object, shown in full in the request example, but the Request Parameters section only described it as *"metadata filter conditions applied to retrieval results"* — with no annotation of its sub-fields or the valid `comparison_operator` values.

This PR documents the `metadata_condition` sub-fields:
- `logic` (`"and"` default / `"or"`)
- `conditions[]` with `name`, `comparison_operator`, and `value`
- the full set of supported `comparison_operator` values

Operator values were taken from the implementation in `common/metadata_utils.py` (`convert_conditions` op_mapping + `meta_filter`), so the documented operators match what the backend accepts.

Docs-only change; no code affected.


Yannnnnnny (commented):
Functionally this docs-only change is mostly correct. 
`value` should not be documented as only `string`. The Python implementation passes it through to `meta_filter()`, where `in` / `not in` can use a list value, and `empty` / `not empty` still require the field even though they ignore it.
Could you please double-check it once more? Thank you!

PR #17902, opened 2026-08-05
Fix QA Excel parser dropping rows whose answer is a zero value

devYRPauli:

The Excel branch of the QA parser drops any row whose answer is a zero value.

`rag/app/qa.py` skipped cells with `if not cell.value`, which is falsy for `0`, `0.0` and `False` as well as for genuinely empty cells. When the answer column held one of those, `a` stayed empty, the row failed the `q and a` check and was appended to `fails` - so it was silently discarded from the index rather than surfaced as content.

Reproduced with a 5-row sheet (openpyxl, unmodified parser loop):

```
before:  2/5 pairs extracted, dropped rows [2, 3, 4]
after:   5/5 pairs extracted, dropped rows []
```

The three lost rows had answers `0` (int), `0.0` (float) and `False` (bool). Text and non-zero numeric answers give the same result before and after, which is why this is easy to miss. The new condition changes two other cases. See "Other behaviour changes" below.

`deepdoc/parser/excel_parser.py` already hit this exact problem and fixed it in #16287 ("keep zero-valued cells when building Excel text chunks"). This ports the same test to the QA parser, which was the only remaining `if not cell.value` in `rag/app/` or `deepdoc/`.

`rag/app/tag.py` does `from rag.app.qa import Excel` and instantiates it, so tag parsing was dropping the same rows.

### Other behaviour changes

The new condition tests for `None` and for an empty string after `strip()`. This changes two more cases. I measured both against the real parser.

**1. Whitespace-only cells are now skipped.** A cell that holds only spaces was truthy before, so it became a question or an answer.

| row | before | after |
|---|---|---|
| `("   ", "answer under a blank question")` | kept, with a blank question | dropped, counted in `fails` |
| `("question with a blank answer", "   ")` | kept, with a blank answer | dropped, counted in `fails` |

A blank question is not a question. The row now shows in the failure line instead of the index. `test_excel_drops_whitespace_only_cells` locks this in.

**2. A leading zero-valued column changes the pair.** The parser takes the first two non-empty cells as the question and the answer.

| row | before | after |
|---|---|---|
| `[0, "Q", "A"]` | `("Q", "A")` | `("0", "Q")` |
| `[5, "Q", "A"]` | `("5", "Q")` | `("5", "Q")` |

The second row is the control. Any leading ID column already broke the pair, so this limit is not new. The `0` case was correct before only because the bug skipped the cell. I did not add a test for this case. A test would lock in `("0", "Q")`, which is not the output we want. A correct fix needs a header row or a column count, and that is out of scope here.

### Test

`test/unit_test/rag/app/test_qa_excel.py` is new. It follows the conventions of the sibling `test/unit_test/rag/app/test_qa_csv.py`. It builds the sheet with openpyxl and calls `qa.chunk` with a `.xlsx` name, so it drives the real parser path.

With the one-line fix reverted and the test kept:

```
FAILED test_excel_keeps_zero_valued_answers - AssertionError: assert 2 == 5
FAILED test_excel_drops_whitespace_only_cells - AssertionError: assert 2 == 1
2 failed
```

With the fix:

```
2 passed
```

`test_qa_csv.py` still gives 4 passed. `ruff check` passes on both changed files.

Note for a local run: `pytest test/unit_test/rag/app/` fails to collect four files, including the pre-existing `test_one.py` and `test_qa_csv.py`, with `ImportError: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser'`. This comes from the stub in `test/unit_test/rag/conftest.py` and is not related to this change. I ran the tests with `--noconftest`.

### Type of change

### DIFF
--- a/rag/app/qa.py
+++ b/rag/app/qa.py
@@ -50,7 +50,7 @@ class Excel(ExcelParser):
             for i, r in enumerate(rows):
                 q, a = "", ""
                 for cell in r:
-                    if not cell.value:
+                    if cell.value is None or str(cell.value).strip() == "":
                         continue
                     if not q:
                         q = str(cell.value)
--- a/rag/app/qa.py
+++ b/rag/app/qa.py
@@ -50,6 +50,7 @@ class Excel(ExcelParser):
             for i, r in enumerate(rows):
                 q, a = "", ""
                 for cell in r:
+                    # A 0, 0.0, or False answer is falsy but is real content.
                     if cell.value is None or str(cell.value).strip() == "":
                         continue
                     if not q:
--- a/test/unit_test/rag/app/test_qa_excel.py
+++ b/test/unit_test/rag/app/test_qa_excel.py
@@ -1,0 +1,100 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+from __future__ import annotations
+
+import warnings
+from io import BytesIO
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", message=".*pkg_resources is deprecated.*", category=UserWarning)
+    import pkg_resources  # noqa: F401 - stabilize xgboost import during collection
+
+import pytest
+from openpyxl import Workbook
+
+from rag.app import qa
+
+
+def _noop_callback(*_args, **_kwargs):
+    pass
+
+
+@pytest.fixture(autouse=True)
+def _stub_rag_tokenizer(monkeypatch):
+    def fake_tokenize(text):
+        return str(text)
+
+    monkeypatch.setattr("rag.nlp.rag_tokenizer.tokenize", fake_tokenize)
+    monkeypatch.setattr("rag.nlp.rag_tokenizer.fine_grained_tokenize", fake_tokenize)
+
+
+@pytest.mark.p2
+def test_excel_keeps_zero_valued_answers():
+    wb = Workbook()
+    ws = wb.active
+    ws.append(["plain question", "plain answer"])
+    ws.append(["how many retries", 0])
+    ws.append(["what is the delta", 0.0])
+    ws.append(["is it enabled", False])
+    ws.append(["count of rows", 7])
+    binary = BytesIO()
+    wb.save(binary)
+    messages = []
+
+    def record_callback(_progress, message):
+        messages.append(message)
+
+    chunks = qa.chunk(
+        "qa.xlsx",
+        binary=binary.getvalue(),
+        lang="English",
+        callback=record_callback,
+    )
+
+    assert len(chunks) == 5
+    assert [chunk["content_with_weight"] for chunk in chunks] == [
+        "Question: plain question\tAnswer: plain answer",
+        "Question: how many retries\tAnswer: 0",
+        "Question: what is the delta\tAnswer: 0",
+        "Question: is it enabled\tAnswer: False",
+        "Question: count of rows\tAnswer: 7",
+    ]
+    assert not any("failure" in message for message in messages)
+
+
+@pytest.mark.p2
+def test_excel_drops_whitespace_only_cells():
+    wb = Workbook()
+    ws = wb.active
+    ws.append(["valid question", "valid answer"])
+    ws.append(["   ", "ignored answer"])
+    binary = BytesIO()
+    wb.save(binary)
+    messages = []
+
+    def record_callback(_progress, message):
+        messages.append(message)
+
+    chunks = qa.chunk(
+        "qa.xlsx",
+        binary=binary.getvalue(),
+        lang="English",
+        callback=record_callback,
+    )
+
+    assert len(chunks) == 1
+    assert any("failure" in message for message in messages)


ebarkhordar (commented):
Since the description says the tests could not run locally, I ran the before/after myself: today's main (`ec9c08d80`, where `rag/app/qa.py` is byte-identical to your branch except for line 53) against your head `935fca12`, in a clean `python:3.12` container with openpyxl, one row per case.

| row written to the sheet | main | this PR |
|---|---|---|
| `("plain question", "plain answer")` | kept | kept |
| `("how many retries", 0)` | dropped | kept as `("how many retries", "0")` |
| `("what is the delta", 0.0)` | dropped | kept |
| `("is it enabled", False)` | dropped | kept |
| `("count of rows", 7)` | kept | kept |

4/7 pairs before, 5/7 after, so the zero-value fix does what you say, and it still applies clean to main two weeks on.

One effect the description does not mention: `str(cell.value).strip() == ""` also changes whitespace-only cells, which were truthy before.

| row | main | this PR |
|---|---|---|
| `("   ", "answer under a blank question")` | kept as a pair with a blank question | dropped, counted in `fails` |
| `("question with a blank answer", "   ")` | kept as a pair with a blank answer | dropped, counted in `fails` |

That looks like the right call to me, since a blank question is not a question and the row now shows up in the failure line rather than entering the index. It is a second behaviour change though, and "Text and non-zero numeric answers were unaffected either way" does not quite cover it, so it is worth a sentence in the body.

On your last-remaining-site claim: parsing every file under `rag/`, `deepdoc/`, `api/` and `agent/` with `ast` and collecting truthiness tests on any `.value` attribute returns exactly one, `rag/app/qa.py:53`. It holds.

xugangqiang:
Thanks for the fix — the root cause (falsy `0`/`False` answers silently dropped) is real, and the change correctly matches the established pattern from #16287. A few issues worth flagging before merge:

**1. No regression test**
The PR adds no test, and there is currently no unit test covering this Excel QA parser. Please add a regression test: a 5-row sheet whose answer column holds `0` (int), `0.0` (float) and `False` (bool), asserting all 5 pairs are extracted and `fails` is empty.

**2. Behavioral edge case — leading zero-valued column**
The parser treats "first two non-empty cells = Q, A". A sheet with a leading zero column like `[0, "Q", "A"]` now changes behavior:
- before: `("Q", "A")` (the `0` was accidentally skipped)
- after: `("0", "Q")` — the ID becomes the question and the real answer is dropped by `break`.

This is a narrow / arguably invalid layout, but it is a genuine behavior change vs. the old code (which "benefited" from the bug for zero-leading columns). Non-zero leading columns were already mis-parsed before, so this is a pre-existing design limitation rather than something introduced here — but please document/acknowledge this edge case so it isn't a surprise.

**3. The CodeRabbit docstring-coverage warning is a false positive**
It flags 0% docstring coverage, but this is a 1-line bug fix adding no new public API that needs docstrings, so no action is required there. (Also worth noting: the bot did not catch the edge case in #2.)

**Minor suggestion (non-blocking):** add a short inline comment next to the new condition explaining *why* it must test `None`/empty-string explicitly instead of `if not cell.value`, to prevent the same bug from being reintroduced.

devYRPauli:
Done. Details per point.

**1. Regression test**

Added `test/unit_test/rag/app/test_qa_excel.py`, next to the existing `test_qa_csv.py` and with the same conventions. It builds the sheet with openpyxl and calls `qa.chunk` with a `.xlsx` name, so it drives the real parser path.

`test_excel_keeps_zero_valued_answers` uses the 5-row sheet you asked for, with answers `0` (int), `0.0` (float) and `False` (bool). It asserts 5 pairs. It also asserts that no callback message reports a failure, which is the `fails` check you named.

With the one-line fix reverted and the test kept:

```
FAILED test_excel_keeps_zero_valued_answers - AssertionError: assert 2 == 5
FAILED test_excel_drops_whitespace_only_cells - AssertionError: assert 2 == 1
2 failed
```

With the fix: `2 passed`. `test_qa_csv.py` still gives 4 passed, and `ruff check` passes on both changed files.

One detail for the record: openpyxl reads `0.0` back so that `str(cell.value)` is `"0"`, not `"0.0"`. The test asserts `Answer: 0`.

**2. Leading zero-valued column**

Confirmed. It is now in the body under "Other behaviour changes". Measured against the real parser:

| row | before | after |
|---|---|---|
| `[0, "Q", "A"]` | `("Q", "A")` | `("0", "Q")` |
| `[5, "Q", "A"]` | `("5", "Q")` | `("5", "Q")` |

The second row is the control. Any leading ID column already broke the pair, so this is the pre-existing limit you describe and not a new one. The `0` case was correct before only because the bug skipped the cell.

I did not add a test for it. A test would lock in `("0", "Q")`, which is not the output we want.

**3. CodeRabbit docstring coverage**

Agreed, no action.

**Inline comment**

Added one line above the condition:

```python
# A 0, 0.0, or False answer is falsy but is real content.
```

devYRPauli:
Thanks for running the before/after, and for the AST pass on the last-site claim.

You are right that the description did not cover the whitespace-only cells. I reproduced both of your rows and added them to the body under "Other behaviour changes". I also replaced the sentence you quoted.

| row | before | after |
|---|---|---|
| `("   ", "answer under a blank question")` | kept, with a blank question | dropped, counted in `fails` |
| `("question with a blank answer", "   ")` | kept, with a blank answer | dropped, counted in `fails` |

`test_excel_drops_whitespace_only_cells` now locks the behaviour in. A sheet with one valid row and one whitespace-only question gives 1 pair and reports 1 failure.

devYRPauli:
@xugangqiang could you add the `ci` label here when you get a chance?

Without it every job reports SKIPPED, so this PR has never had its suite run. The check names still show the unexpanded `${{ matrix.api_proxy_scheme }}` form, which is the tell.

#18713 was in the same state until the label landed on it yesterday. It then ran green across `ragflow_preflight`, `ragflow_tests_elasticsearch` and `ragflow_tests_infinity`.

Both review points from 2026-08-19 are answered. The fix is still needed on current main: `rag/app/qa.py:53` is `if not cell.value:`, so a `0`, `0.0` or `False` answer is still dropped as if the cell were empty. `test/unit_test/rag/app/test_qa_excel.py` is not upstream yet either.

Nothing is pending from my side.

xugangqiang:
LGTM @devYRPauli  Thank you for this. Could you please check if go side has the same issue?

devYRPauli:
I checked the Go side. It does not have the same bug.

The Go path turns each cell into a string before the check:

- `internal/parser/parser/office_table_render.go:604` reads the sheet with excelize `GetRows`, which returns strings.
- `internal/parser/parser/office_table_render.go:201-208` writes every cell as a `<td>`, empty cells included.
- `internal/ingestion/component/chunker/qa.go:216` keeps a cell when `t != ""`.

A `0` cell becomes the string `"0"`. `"0" != ""` is true, so the Go code keeps the pair. The Python code tested the value itself, so `0` was falsy and the row was dropped.

I did find a different problem in the same function. I would like your view before I open a PR.

`extractQATable` removes empty cells from the list, then pairs `texts[0]` with `texts[1]` (`qa.go:213-222`). This loses the column position. A row `question | (empty) | extra` pairs the question with the third column.

The reach differs by file type:

- CSV reaches this code. `internal/parser/parser/csv_parser.go:160` sets `OutputFormat: "html"`, and `qa.go:114` calls `extractQATable` for HTML.
- XLSX does not reach it by default. `internal/parser/parser/xlsx_parser.go:145` sets `OutputFormat: "json"`, so `qa.go:121` calls `extractQAJSON`.

`internal/ingestion/component/extractor_tag.go:843-861` uses the same pattern for tag source files.

One limit on this check: I read the code only. I have no Go toolchain on this machine, so I did not run the Go tests.

Shall I open a PR for the CSV column case?

xugangqiang:
@devYRPauli  please help to raise issue - and if possible, please help to fix. Thank you.

devYRPauli:
Raised it as #19006.

One correction to my earlier comment. I wrote that XLSX does not reach `extractQATable` because `xlsx_parser.go` sets `OutputFormat: "json"`. That was wrong. `internal/parser/parser/xlsx_parser.go:134` sets `"html"`, and the local excelize path is the only one that runs for a spreadsheet. XLSX reaches the same function as CSV.

That changes the fix, not the report. Python holds two different rules:

- `rag/app/qa.py:53`, the Excel path: skip empty cells and take the first two values. The Go code matches this.
- `rag/app/qa.py:365`, the CSV path: require exactly two fields. The Go code does not match this.

Go serves both formats from one function. A strict two-field rule for CSV would also change XLSX, where the Go behavior is correct today. The issue sets out the two options. Tell me which contract you want and I will send that PR.

I have no Go toolchain on this machine. I read the Go code and ran only the Python side.

PR #17905, opened 2026-08-06
refactor(task): sink parser page-cap override into pipeline package

xugangqiang:
## Summary

Moves the canvas-debug parser page-cap injection out of the `task` orchestrator and into a **debug-agnostic** `pipeline` helper, so `PipelineExecutor` keeps only the orchestration skeleton (resolving one of the P1 review findings: the executor was overloaded with DSL/parser-param assembly).

### Changes
- **`pipeline/parser_page_cap.go`** (new):
  - `BuildParserPageCapOverride(parserConfig, dsl, docType, capPages int, parserComponentName string, familyOf)` — injects the `ParserConfig[cpnID][family]["pages"]` cap through the same `override_params` channel production uses. The cap value and family resolution are injected by the caller, so the function carries no debug semantics and is reusable for any page-cap scenario.
  - `ExtractParserCpnID(dsl, parserComponentName)` — shared Parser cpnID discovery from (optionally enveloped) DSL.
  - `UnwrapCanvasDSL(raw []byte)` — exported single source of truth for stripping the `{"dsl": {...}}` canvas envelope.
  - `pipeline` does **not** import `component` (no reverse dependency); callers inject `component.ComponentNameParser` / `component.ParserFileFamily`.
- **`task/pipeline_executor.go`**: removed `injectDebugPageCap` (the `debugPageCapPages = 2` constant stays in the task package). The debug branch now calls `pipeline.BuildParserPageCapOverride(...)`.
- **`task/pipeline_executor.go` `warnUnknownComponentParams`**: fixed a production no-op bug — it passed the enveloped DSL straight to `ExtractAllComponentParams`, which silently errored and disabled the unknown-cpnID guard. It now unwraps the envelope first.
- **`task/debug_result_dsl.go`**: reuses `pipeline.UnwrapCanvasDSL` instead of a third inline envelope-unwrap copy.

### Behavior
No external debug-preview behavior changes. The three original invariants are preserved exactly:
1. explicit `pages` caps under `cpnID+family` are respected (not overwritten),
2. an empty family (unknown docType) is a no-op,
3. the injected shape is `[]any{[]any{1, capPages}}` (the `[]any`-of-`[]any` form `NormalizePDFPages` requires).

## Test plan
- New `pipeline/parser_page_cap_test.go`: `BuildParserPageCapOverride` (inject / respect-existing / unknown-family no-op / no-Parser no-op), `ExtractParserCpnID` (enveloped + raw), `UnwrapCanvasDSL`.
- `task/debug_test.go`: `TestInjectDebugPageCap` migrated to the new helper; new `TestWarnUnknownComponentParamsDetectsUnknownCPNFromEnvelope` captures the warning via `zaptest/observer` to prove the envelope no-op bug is fixed.
- Both `internal/ingestion/pipeline` and `internal/ingestion/task` pass `build.sh --test` (unit tier).

## Notes
- `TOKEN_CHUNKER_HANDOFF.md` is an unrelated untracked file and was deliberately **not** included in this PR.

### DIFF
--- a/internal/ingestion/pipeline/parser_page_cap.go
+++ b/internal/ingestion/pipeline/parser_page_cap.go
@@ -1,0 +1,109 @@
+package pipeline
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// UnwrapCanvasDSL decodes a raw pipeline DSL and strips the optional canvas
+// envelope {"dsl": {...}}, returning the inner components-carrying map. A raw
+// (non-enveloped) DSL is returned unchanged. It is the []byte entry point used
+// by helpers that need the inner DSL (ExtractParserCpnID,
+// BuildParserPageCapOverride) so envelope-handling lives in exactly one place.
+func UnwrapCanvasDSL(raw []byte) (map[string]any, error) {
+	var top map[string]any
+	if err := json.Unmarshal(raw, &top); err != nil {
+		return nil, fmt.Errorf("UnwrapCanvasDSL: decode: %w", err)
+	}
+	if top == nil {
+		return nil, errNilDSL
+	}
+	if env, ok := top["dsl"].(map[string]any); ok && len(env) > 0 {
+		return env, nil
+	}
+	return top, nil
+}
+
+// ExtractParserCpnID returns the cpnID of the first component whose
+// component_name equals parserComponentName, or "" when no such component
+// exists or the DSL cannot be read. dsl may be enveloped; it is unwrapped
+// first.
+func ExtractParserCpnID(dsl []byte, parserComponentName string) string {
+	inner, err := UnwrapCanvasDSL(dsl)
+	if err != nil {
+		return ""
+	}
+	innerJSON, err := json.Marshal(inner)
+	if err != nil {
+		return ""
+	}
+	schemas, err := ExtractAllComponentParams(innerJSON)
+	if err != nil {
+		return ""
+	}
+	for _, s := range schemas {
+		if s.ComponentName == parserComponentName {
+			return s.CpnID
+		}
+	}
+	return ""
+}
+
+// BuildParserPageCapOverride returns parserConfig with a page cap injected for
+// the Parser component, keyed by its cpnID and the document's filetype family.
+//
+// It is debug-agnostic: capPages and familyOf are supplied by the caller, so
+// the function carries no debug-specific semantics and is reusable for any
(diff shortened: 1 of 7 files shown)


xugangqiang:
## Review summary — decisions & verification

This PR (3 commits) sinks the canvas-debug parser page-cap injection into a debug-agnostic `pipeline.BuildParserPageCapOverride`, fixes the `warnUnknownComponentParams` enveloped-DSL no-op bug, and collapses envelope unwrapping to a single `pipeline.UnwrapCanvasDSL`. Three review points were raised; decisions below.

### 1. `BuildDebugResultDSL` — keep the "Debug" name
Unlike `BuildParserPageCapOverride` (where debug was only a caller-supplied `capPages=2` value, with no debug logic inside), this function's body is *genuinely* debug-contract-shaped: it strips vector keys specifically to keep the **Redis-stored debug log** at Python-scale size and emits a UI-only subset of the component output. The name accurately reflects that internal contract, so it stays.

### 2. `BuildParserPageCapOverride` — keep the 6-arg signature for now
The `parserConfig, dsl, docType, capPages, parserComponentName, familyOf` signature is a bit long, but bundling it into a struct is YAGNI at this point (low priority, can be done later). Note `pipeline` does **not** import `component` — callers inject `component.ComponentNameParser` / `component.ParserFileFamily` — so there is no reverse-dependency concern.

### 3. `warnUnknownComponentParams` warning spam — confirmed low risk
The fix changed behavior from a silent no-op to actually emitting a warning when `parserConfig` has a cpnID absent from the DSL. This is intended (it surfaces real config drift) and is non-fatal (the mismatched entry is dropped at the runtime merge).
- **Write-path guarantee:** the Go dataset update path reconciles `parserConfig` against the canvas DSL and **strips unknown cpnIDs** (see `internal/service/dataset/update_test.go`: "unknown cpnID is stripped"). So data written/updated through this path is consistent and will not trigger the warning.
- **Residual risk:** only pre-existing "legacy/stale" rows (created before this validation existed, with a since-edited canvas, never re-saved) — exactly the guard's intended audience.
- **Test added:** `debug_pages_integration_test.go` now captures logs via `zaptest/observer` and asserts a well-formed template debug run emits **no** unknown-cpnID warning (verified green in the integration tier).

No code change was required for items 1–3. The PR is safe to merge; the added integration test keeps the fix from regressing into log noise on the normal path.

yuzhichang:
LGTM

PR #17906, opened 2026-08-06
  fix(document-preview): make ExcelCsvPreviewer adapt to container resize

dcc123456:
### Summary

 fix(document-preview): make ExcelCsvPreviewer adapt to container resize

The Excel/CSV previewer was rendered once at mount and never updated
  when the window or container changed size, leaving the spreadsheet
  frozen at its initial dimensions.


wangq8 (commented):
Good fix for the resize issue! The destroy/re-init approach is the right call since `@js-preview/excel` doesn't expose a public resize method.

A couple of things to address:

1. **Missing cancellation flag**: The fetch `useEffect` doesn't have a `cancelled` flag. If `filePath` changes while an async `fetchDocument` call is in-flight, the stale promise could resolve and call `setDataReady` / `setStatus` on a component that has already moved on. Compare with #17912 which includes this pattern.

2. **Conflicts with #17912**: Both PRs modify `useFetchExcel` with the same resize-adaptive refactoring. #17912 also adds WPS DISPIMG stripping. These need to be coordinated — either add the cancellation flag here and merge independently, or close this in favor of #17912 which already includes both the resize fix and the cancellation flag.

3. **Callback ref pattern**: Using `useState` + callback ref (`ref={setContainerEl}`) is a valid pattern for `useSize` compatibility. Just want to confirm this was tested — the caller in `excel-preview.tsx` destructures `containerRef` and passes it as `ref={containerRef}`, which works because React callback refs accept the state setter directly. 👍

The debouncing approach (150ms, leading+trailing) is sensible and prevents thrashing on rapid resizes. Thanks!

PR #17909, opened 2026-08-06
fix(nginx): use 127.0.0.1 in python conf to avoid ~1s IPv6 fallback latency

attention-ai-yee:
## Problem

In the default deployment (`API_PROXY_SCHEME=python`), the nginx config generated from `docker/nginx/ragflow.conf.python` uses:

```nginx
proxy_pass http://localhost:9381;  # admin
proxy_pass http://localhost:9380;  # api
```

Inside the RAGFlow container, `/etc/hosts` maps `localhost` to both `127.0.0.1` and `::1`:

```
127.0.0.1   localhost
::1         localhost ip6-localhost ip6-loopback
```

But the backend services (`ragflow_server.py` on 9380, admin server on 9381) only listen on IPv4 (`127.0.0.1`). When nginx resolves `localhost` to `::1`, the upstream connection fails and nginx falls back to the IPv4 address - each fallback adds ~1 second of latency.

## Symptom

The web UI feels intermittently sluggish. Through nginx, some endpoints intermittently stall ~1.0s while others are sub-10ms. Reproducible measurement (inside the container, same access token, 6 calls):

| endpoint | via nginx (`localhost:80`) |
|---|---|
| `GET /api/v1/tenants` | 0.004s, **1.005s**, 0.004s, 0.004s, **1.005s**, 0.004s |
| `GET /api/v1/users/me` | 0.004s, 0.004s, 0.004s |

Directly hitting the backend (`curl http://127.0.0.1:9380/...`) is consistently ~5ms, confirming the latency is introduced at the nginx `proxy_pass` resolution step, not in the Python backend.

## Root cause

This template is inconsistent with the other two in the same directory, which already use `127.0.0.1` and are not affected:

| template | `proxy_pass` | affected? |
|---|---|---|
| `ragflow.conf.python` | `http://localhost:...` | yes |
| `ragflow.conf.golang` | `http://127.0.0.1:...` | no |
| `ragflow.conf.hybrid` | `http://127.0.0.1:...` | no |

## Fix

Use `127.0.0.1` explicitly in `ragflow.conf.python`, matching the golang and hybrid templates.

## Verification

After the change (container fully recreated so the entrypoint regenerates `ragflow.conf` from the patched template), 6 consecutive calls are all sub-millisecond:

```
0.000658s
0.000677s
0.000760s
0.000600s
0.000527s
0.000490s
```

The intermittent ~1s stalls are gone.

## Environment

- RAGFlow v0.26.4 (official image `infiniflow/ragflow:v0.26.4`)
- `API_PROXY_SCHEME=python` (default)
- Ubuntu host, 4-core / 16GB, no GPU


docker/nginx/ragflow.conf.python
```diff
@@ -8,15 +8,15 @@ server {
     gzip_comp_level 9;
     gzip_types text/plain application/javascript application/x-javascript text/css application/xml text/javascript application/x-httpd-php image/jpeg image/gif image/png;
     gzip_vary on;
-    gzip_disable "MSIE [1-6]\.";
+    gzip_disable "MSIE [1-6].";
```
JinHai-CN: The dot (.) in this pattern is not escaped, which means it will be interpreted as a regex wildcard (matching any single character) rather than a literal dot.

JinHai-CN:
@attention-ai-yee please check the comment.

attention-ai-yee:
Hi @JinHai-CN, thanks for catching that! You're right - the backslash was accidentally dropped when I generated the patch.

I've force-pushed a clean commit (`045b2546`). This time I started from the upstream `main` and applied only the two intended changes (`localhost` -> `127.0.0.1`), so the original `gzip_disable "MSIE [1-6]\."` is preserved unchanged. The diff is now limited to just those two lines:

```diff
-        proxy_pass http://localhost:9381;
+        proxy_pass http://127.0.0.1:9381;
-        proxy_pass http://localhost:9380;
+        proxy_pass http://127.0.0.1:9380;
```

Apologies for the noise, and thanks again for the careful review!

PR #17910, opened 2026-08-06
refa: unify wiki example naming

buua436:

Unify Wiki template example handling across the backend and frontend.

- Replace the internal `page_example` naming with `example`.
- Remove the obsolete compatibility path with duplicate example parameters.
- Use the form field state directly through `RAGFlowFormItem`.
- Localize the Example label.

### Type of change

- [x] Refactor (no breaking change)


wangq8 (commented):
Clean refactor, LGTM! 

Changes are consistent across the stack:
- Backend: `page_example` → `example` in `wiki.py`, caller updated in `dataset_wiki_generator.py`
- Frontend: `useBlueprintSelection` simplified to export `examplePath`, form state bound via `RAGFlowFormItem` instead of manual `setValue`
- Locales: shortened labels from "Page-structure example" → "Example" (and equivalents in ja/zh)

The removal of the dual `page_example`/`example` compatibility path is a good cleanup — all callers have been updated so no breakage. The `RAGFlowFormItem` wrapping with the `children` render-prop pattern in the edit component is idiomatic and cleaner than the previous manual state management.

No blocking concerns. Thanks!

PR #17911, opened 2026-08-06
fix: make txt file preview scrollable with mouse wheel

euvre:
### Summary

Fix the issue where the txt file preview (opened from File Management) does not scroll with the mouse wheel.

The `TxtPreviewer` container lacked `overflow-auto`, and the document-viewer page rendered it without a height constraint (unlike the md/pdf/ppt previews, which use `h-dvh`). As a result, long text content overflowed the page with no scrollable area.

Changes:
- Add `overflow-auto` to the `TxtPreviewer` container so overflowing content scrolls inside the container.
- Pass `!h-dvh` to `TxtPreviewer` in the document-viewer page, consistent with other preview types.

### DIFF
--- a/web/src/pages/document-viewer/index.tsx
+++ b/web/src/pages/document-viewer/index.tsx
@@ -49,7 +49,9 @@ const DocumentViewer = () => {
       {(ext === 'md' || ext === 'mdx') && (
         <Md url={api} className="!h-dvh p-5"></Md>
       )}
-      {ext === 'txt' && <TxtPreviewer url={api}></TxtPreviewer>}
+      {ext === 'txt' && (
+        <TxtPreviewer url={api} className="!h-dvh"></TxtPreviewer>
+      )}
 
       {ext === 'pdf' && (
         <PdfPreview url={api} className="!h-dvh p-5"></PdfPreview>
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix! This is clean and consistent with how other preview types handle scrolling.

**What I checked:**

1. **Root cause confirmed** — The TxtPreviewer container was missing `overflow-auto`, and the document-viewer page didn't constrain its height (unlike md/pdf/ppt previews which all use `!h-dvh`). Without both changes together, long text content just overflows the page with no scrollable area.

2. **Fix is minimal and correct:**
   - Adding `overflow-auto` to the container div enables scrollability
   - Passing `!h-dvh` as className in the document-viewer page constrains height, consistent with all other preview types

3. **No regressions expected** — The change only affects the txt preview path in the document-viewer page. The TxtPreviewer component still accepts an optional `className` prop and merges it with the defaults via `classNames()`, so there's no breaking change.

4. **No security issues, style issues, or missing tests.**

LGTM.

dcc123456 (approved):

web/src/pages/document-viewer/index.tsx
```diff
@@ -49,7 +49,9 @@ const DocumentViewer = () => {
       {(ext === 'md' || ext === 'mdx') && (
         <Md url={api} className="!h-dvh p-5"></Md>
       )}
-      {ext === 'txt' && <TxtPreviewer url={api}></TxtPreviewer>}
+      {ext === 'txt' && (
+        <TxtPreviewer url={api} className="!h-dvh"></TxtPreviewer>
```
dcc123456: This height should not be hard-coded.

wangq8:
@euvre  resolve conflict.

PR #17912, opened 2026-08-06
  fix(document-preview): strip WPS DISPIMG formula from xlsx before preview 

dcc123456:
### Summary

  fix(document-preview): strip WPS DISPIMG formula from xlsx before preview                                       
                                                                                                                  
  WPS spreadsheets embed images in cells via the proprietary `_xlfn.DISPIMG`        
  formula (stored in `xl/cellimages.xml`). The `@js-preview/excel` library                                        
  cannot evaluate this formula, causing a crash with:                               
    TypeError: Cannot read properties of undefined (reading 'render')
  followed by an uncaught:
    TypeError: Cannot read properties of null (reading '_worksheets')
  from a pending internal setTimeout after `destroy()` nullifies the workbook.

  Office-generated files work fine because they don't use DISPIMG.

  Fix: pre-process the xlsx ArrayBuffer with JSZip to remove DISPIMG `<f>`
  formula tags from worksheet XML and replace cached `<v>` values with a
  `[图片]` text placeholder, so the rest of the spreadsheet renders normally.
  Files without DISPIMG are returned untouched (no re-zip overhead).

  Added `jszip` and `@types/jszip` as direct dependencies (previously only
  available as a transitive dep via `pptx-preview`).


wangq8 (commented):
Thanks for this fix! The approach of pre-processing the xlsx with JSZip to strip DISPIMG formulas is solid, and the graceful fallback to the original ArrayBuffer when processing fails is good defensive coding.

A few observations:

1. **Hardcoded Chinese placeholder `[图片]`**: The `<v>[图片]</v>` placeholder uses Chinese text. Since RAGFlow supports English and Japanese locales, consider using a locale-neutral placeholder like `[Image]`, or making it localizable.

2. **Superset of #17906**: This PR includes the same resize-adaptive refactoring found in #17906 (useState callback ref, useDebounceFn, destroy/re-init pattern). If both are merged, conflicts will occur. Please coordinate with #17906 — likely close #17906 in favor of this more complete PR.

3. **`compression: "DEFLATE"` on re-zip**: When DISPIMG is found and the zip is regenerated, the entire file gets DEFLATE compression even if the original used STORE or another method. For preview purposes this should be fine, but worth being aware of.

4. **Cancellation flag**: Nice addition of the `cancelled` flag in the fetch useEffect. This prevents stale async updates if the filePath changes mid-fetch.

5. **package.json noise**: The `oxfmt`/`oxlint` entries were re-ordered alphabetically — likely from `npm install`. This is minor but could be separated into a standalone tidy-up commit.

Overall LGTM with the placeholder concern being the main one. Thanks for catching and fixing this WPS-specific crash!

PR #17915, opened 2026-08-06
fix: allow recreating deleted template groups

buua436:

Fix duplicate-key errors when deleting and recreating compilation template groups with the same name.

- Remove the obsolete `(tenant_id, name, status)` unique index.
- Drop legacy versions of the index during database migration.
- Keep duplicate-name validation for active groups.
- Allow reusing names from invalidated groups.

### Type of change

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -1183,7 +1183,6 @@ class CompilationTemplateGroup(DataBaseModel):
 
     class Meta:
         db_table = "compilation_template_group"
-        indexes = ((("tenant_id", "name", "status"), True),)
 
 
 class Search(DataBaseModel):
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
**Approved** — clean, well-scoped fix.

**What this does:** Removes the `(tenant_id, name, status)` unique index from `CompilationTemplateGroup` and adds migration entries to drop any legacy incarnations of this index. This allows deleting a template group and recreating a new one with the same name.

**Validation is preserved:** The API handlers at both the create (`POST /compilation-template-groups`) and update (`PUT /compilation-template-groups/<group_id>`) paths already call `CompilationTemplateGroupService.name_exists()`, which filters by `status == VALID` — so two *active* groups with the same name are still rejected. The database index was the only thing blocking the "delete-then-recreate" flow, and removing it is safe given the existing application-level guard.

**Migration:** The three index name variants (`compilationtemplategroup_tenant_id_name_status`, `compilation_template_group_tenant_id_name_status`, `idx_compilation_template_group_tenant_id_name_status`) cover the peewee auto-naming conventions. No concerns here.

LGTM. 🚢

PR #17921, opened 2026-08-06
fix(number-input): improve value handling for out-of-range inputs

dcc123456:
### Summary

fix(number-input): improve value handling for out-of-range inputs

### DIFF
--- a/web/src/components/originui/number-input.tsx
+++ b/web/src/components/originui/number-input.tsx
@@ -83,23 +83,41 @@ const NumberInput = forwardRef<
     }
 
     if (!isNaN(newValue)) {
-      if (newValue > max || newValue < min) {
-        return;
-      }
+      // Allow intermediate editing states that fall outside [min, max]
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice UX improvement! Allowing intermediate editing states (e.g. clearing "1024" back to "102" when min is 512) instead of blocking the input entirely is much more user-friendly. The clamp-on-blur pattern is well-implemented.

A couple of observations:

1. **Good**: `handleBlur` now also considers `max` in both branches (number and non-number paths). Previously max wasn't checked in the else branch.

2. **Minor suggestion**: The `handleBlur` callback now references `min`, `max`, `onChange`, and `value` in its dependency array, which is correct. However, note that `handleChange` also captures `min` and `max` via closure — it might be worth adding them to the dependency array of the `useCallback` wrapping `handleChange` for consistency, though in practice they typically don't change per-render.

LGTM!

PR #17924, opened 2026-08-06
fix: implement Tongyi-Qianwen TTS via DashScope native endpoint

euvre:
### Summary

Chat read-aloud (TTS) failed with "TTS synthesis produced no audio" when the tenant's default TTS model is provided by Tongyi-Qianwen (Aliyun DashScope). The Go `AliyunModel` driver returned `no such method` for `AudioSpeech`, so every synthesis segment failed and the endpoint returned an error instead of audio. DashScope does not expose an OpenAI-compatible `/audio/speech` endpoint (probed: HTTP 404), so this implementation targets the native multimodal-generation API (probed: reachable).

Changes:
- Implement `AliyunModel.AudioSpeech` against `POST /api/v1/services/aigc/multimodal-generation/generation` with the Qwen TTS request shape (`{model, input:{text, voice, language_type?}}`); `voice` defaults to `Cherry` and is overridable via `TTSConfig.Params`.
- The non-streaming response carries no inline audio; the driver downloads the synthesized file from `output.audio.url` (valid for 24h).
- Qwen TTS returns WAV (not mp3): add `TTSResponse.MediaType`, mark results `audio/wav`, and honor it in the chat audio endpoint's `Content-Type`.
- Chat read-aloud synthesizes per text segment; WAV segments cannot be concatenated byte-wise (each has its own RIFF header), so the handler now extracts each segment's PCM and re-wraps the combined payload with a single WAV header (`internal/handler/chat_audio_wav.go`).
- Implement `AudioSpeechWithSender` as a single-chunk forward.
- Add the `tts` URL suffix for Tongyi-Qianwen in `conf/models/aliyun.json`.
- Tests: TTS tests against the native contract (default/explicit voice, language_type, API error, missing audio URL, missing suffix, sender fallback) and WAV split/build round-trip tests.

Note: `qwen3-tts-*-realtime` variants are WebSocket-only and out of scope; non-realtime models such as `qwen-tts` / `qwen3-tts-flash` work with this endpoint.

Follow-up fix to #17770 (merged): the compatible-mode audio/speech path used there does not exist on DashScope, so TTS still produced no audio.

### DIFF
--- a/conf/models/aliyun.json
+++ b/conf/models/aliyun.json
@@ -11,7 +11,7 @@
     "embedding": "compatible-mode/v1/embeddings",
     "rerank": "compatible-api/v1/reranks",
     "models": "compatible-mode/v1/models",
-    "tts": "compatible-mode/v1/audio/speech"
+    "tts": "api/v1/services/aigc/multimodal-generation/generation"
   },
   "models": [
     {
--- a/internal/entity/models/aliyun.go
+++ b/internal/entity/models/aliyun.go
@@ -395,9 +395,39 @@ func (a *AliyunModel) TranscribeAudioWithSender(ctx context.Context, modelName *
 // DashScope's Qwen TTS models require one.
 const aliyunTTSDefaultVoice = "Cherry"
 
-// newAliyunTTSRequest builds the OpenAI-compatible audio/speech request
-// against DashScope's compatible mode.
-func (a *AliyunModel) newAliyunTTSRequest(ctx context.Context, modelName *string, audioContent *string, apiConfig *APIConfig, ttsConfig *TTSConfig) (*http.Request, error) {
+// aliyunTTSRequest is the DashScope multimodal-generation request for Qwen
+// TTS models (qwen-tts / qwen3-tts-flash family).
+type aliyunTTSRequest struct {
+	Model string         `json:"model"`
+	Input aliyunTTSInput `json:"input"`
+}
+
+type aliyunTTSInput struct {
+	Text string `json:"text"`
+	// Voice is required by Qwen TTS models (e.g. "Cherry").
+	Voice string `json:"voice"`
+	// LanguageType hints the text language (e.g. "Chinese", "English");
+	// omitted to let the model auto-detect.
+	LanguageType string `json:"language_type,omitempty"`
+}
+
+// aliyunTTSResponse is the non-streaming DashScope multimodal-generation
+// response. The synthesized audio is not inlined; output.audio.url points
+// to a downloadable file (valid for 24h).
+type aliyunTTSResponse struct {
+	Output struct {
+		Audio struct {
+			URL string `json:"url"`
+		} `json:"audio"`
+		FinishReason string `json:"finish_reason"`
+	} `json:"output"`
+	Code      string `json:"code"`
+	Message   string `json:"message"`
+	RequestID string `json:"request_id"`
+}
+
+// AudioSpeech convert text to audio
+func (a *AliyunModel) AudioSpeech(ctx context.Context, modelName *string, audioContent *string, apiConfig *APIConfig, ttsConfig *TTSConfig, modelUsage *common.ModelUsage) (*TTSResponse, error) {
 	if err := a.baseModel.APIConfigCheck(apiConfig); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


wangq8 (commented):
Thanks for this well-structured PR! The implementation is solid and the tests cover the important paths. I have a few observations:

**Strengths:**
- Good request/response type definitions matching the DashScope API contract
- Proper error handling at each stage (API errors, missing audio URL, missing suffix, empty audio, download failures)
- The WAV handling (`splitWAV`/`buildWAV`) correctly parses RIFF chunks and handles the fmt/data chunk extraction—critical for segment concatenation in chat read-aloud
- Voice defaulting to "Cherry" with `ttsConfig.Params` overrides is sensible
- `AudioSpeechWithSender` cleanly delegates to `AudioSpeech` as a single-chunk forward

**Suggestions (non-blocking):**

1. **Audio download size limit**: `downloadAliyunTTSAudio` reads the entire audio response into memory via `io.ReadAll`. Consider adding a `io.LimitReader` or similar guard to prevent OOM on unexpectedly large responses—DashScope TTS audio is naturally bounded, but defense-in-depth is worth it.

2. **WAV data chunk padding**: In `buildWAV`, the RIFF total calculation accounts for `fmt` chunk padding but not the `data` chunk. Per the RIFF spec, the data chunk should also be 2-byte aligned. In practice this is harmless because PCM sizes are always even for 16-bit/mono audio, but a brief comment noting this assumption would help future readers.

3. **`splitWAV` chunk size int conversion**: `int(binary.LittleEndian.Uint32(...))` is fine for realistic WAV files, but a defensive `size > len(b)` guard (already present) is sufficient—just noting it could overflow on 32-bit platforms (not relevant for RAGFlow).

Overall LGTM, no blocking issues.

wangq8:
@euvre resolve conflict.

Lynn-Inf (approved):
LGTM

wangq8:
@euvre resolve conflict.

euvre:
Thanks everyone for the reviews. Here is how each point was handled (commit `c2df88f`):

**Applied**

1. **WAV `data` chunk padding** (@coderabbitai on `chat_audio_wav.go`, also @wangq8's note 2): `buildWAV` now accounts for the RIFF pad byte of odd-sized PCM in both the size field and the payload, keeping every chunk 2-byte aligned per the RIFF spec. `TestSplitWAVRoundTrip` (which uses 5-byte, i.e. odd, PCM) now asserts an even total length and a consistent RIFF size. This implements the actual fix rather than only documenting the assumption.

2. **One media type / one WAV format per response** (@coderabbitai on `chat_audio.go`): the handler now locks the first successfully synthesized segment's media type (empty still defaults to `audio/mpeg`) and skips later segments of a different type with a warning, so mixed encodings are never concatenated under a single `Content-Type`, and buffered WAV PCM can no longer be silently dropped after a non-WAV segment started the stream. WAV segments whose `fmt ` chunk differs from the first segment's are skipped as well.

3. **Audio download size limit** (@wangq8, suggestion 1): `downloadAliyunTTSAudio` now reads through `io.LimitReader` with a 64 MiB cap and fails with a clear error when exceeded — defense-in-depth, since DashScope TTS audio is naturally far smaller. Covered by a new `TestAliyunAudioSpeechRejectsOversizedAudio`.

4. **Test stub strictness** (@coderabbitai on `aliyun_test.go`): the fake DashScope server now returns `405 Method Not Allowed` for any method other than GET/POST, and the synthesis response write error is checked via `t.Errorf`.

**Not changed**

- @wangq8 note 3 (`int` conversion of the chunk size in `splitWAV`): agreed — the existing `off+size > len(b)` bounds guard is sufficient, and 32-bit overflow is not relevant for RAGFlow's targets, so no change.
- The remaining comments from the 13-comment full-review batch (@coderabbitai) target files outside this PR (`admin/server/services.py`, `agent/canvas.py`, `rag/app/table.py`, `rag/llm/cv_model.py`, `rag/nlp/search.py`, etc.). They were generated by the full-review fallback while the branch was mid-conflict; none of those files are touched by this PR (the diff is 7 files: `aliyun.go`, `aliyun_test.go`, `types.go`, `chat_audio.go`, `chat_audio_wav.go`, `chat_audio_wav_test.go`, `aliyun.json`), so they are out of scope here and belong to main-branch review.

**Merge conflict**: GitHub reports the branch mergeable against current main again, so no rebase/merge was needed.

**Validation**: `bash build.sh --test ./internal/handler/... ./internal/entity/models/...` passes (both packages `ok`). No shared-service end-to-end validation was performed; the basis is unit tests plus code-level analysis. The media-type locking itself is not unit-tested at handler level because `ChatHandler` depends on the concrete `ModelProviderService` (no interface seam); the WAV helpers it relies on are covered by `chat_audio_wav_test.go`.

Thanks @Lynn-Inf for the approval and @wangq8 for the detailed review!

PR #17925, opened 2026-08-06
feat: add html file type icon

dcc123456:
### Summary

feat: add html file type icon


wangq8 (commented):
Clean addition. The HTML icon SVG looks good and the mapping in `file.ts` is correct. LGTM!

One minor note: there seems to be a small SVG path artifact at the end of the icon definition (the trailing `<path>` appears in the diff but is part of an unrelated icon). Not a blocker — just double-check the rendered icon looks as expected.

wangq8:
@dcc123456 resolve conflict.

PR #17926, opened 2026-08-06
fix(go): unify children delimiter pattern with backtick-strip + rune order

xugangqiang:
## Summary

`compileChildrenPattern` re-implemented the delimiter-list compile inline with two divergences from the shared `CompileDelimiterListPattern`:

- It never stripped backticks, so a backtick-wrapped `children_delimiter` like `` `###` `` matched the **literal wrapped token** rather than the inner `###`.
- It sorted by **byte length** instead of rune count (`sortSlice`), so multi-byte delimiters could be ordered incorrectly and a longer delimiter could fail to win over a shorter prefix.

## Fix

Extract `CompileDelimiterPatternList(delims, keepBare)` in `internal/parser/chunk/delim.go` as the single shared helper:

- Backtick-wrapped entries contribute their **inner** content (backticks stripped).
- Bare entries are kept active only when `keepBare=true` (children semantics — every non-empty children delimiter splits); with `keepBare=false` bare entries are ignored (main `delimiters` list, unchanged behavior).
- Each entry is `regexp.QuoteMeta`'d and the alternation is sorted longest-first by rune count (stable).

`CompileDelimiterListPattern` now delegates with `keepBare=false` (behavior preserved). `compileChildrenPattern` delegates with `keepBare=true`. The now-dead `sortSlice` is removed.

## Test plan

- `internal/parser/chunk/delim_test.go`: `TestCompileDelimiterPatternListKeepBare` / `...KeepBareFalse` cover backtick-strip, bare-active, rune-descending order, and equivalence to `CompileDelimiterListPattern`.
- `internal/ingestion/component/chunker/token_children_delim_test.go`: `TestTokenChunker_ChildrenDelimiterBacktickStripped` asserts an end-to-end `Invoke` splits on the inner `###` and never matches the literal wrapped token.
- Existing `TestTokenChunker_ChildrenDelimiterDropped*` (bare `. ` / `。`) still pass — confirms no regression in the "bare stays active" children semantics.

`bash build.sh --test ./internal/parser/chunk/... ./internal/ingestion/component/chunker/...` passes.

This PR is independent of #17920.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -1134,32 +1134,14 @@ func applyChildrenDelimText(docs []schema.ChunkDoc, pattern *regexp.Regexp) []sc
 
 // compileChildrenPattern is the children_delimiters version of
 // compileDelimPattern. Returns nil when no delimiters exist.
+// compileChildrenPattern builds the children-split regex from a
+// `children_delimiters` list. Every non-empty entry is active (including bare
+// ones), and backtick-wrapped entries contribute their inner content — see
+// chunk.CompileDelimiterPatternList. Delegating keeps children splitting
+// consistent with the main delimiter list (backtick stripping + rune-descending
+// order) instead of re-implementing a divergent copy.
 func compileChildrenPattern(delims []string) *regexp.Regexp {
-	if len(delims) == 0 {
-		return nil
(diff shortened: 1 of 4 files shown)


yuzhichang:
LGTM

PR #17927, opened 2026-08-06
fix(agent): hide log sheet when closing chat sheet

dcc123456:
### Summary

fix(agent): hide log sheet when closing chat sheet


wangq8 (commented):
Nice catch! Closing the chat sheet without also hiding the log sheet would leave the UI in an inconsistent state. The fix is minimal and correct — wrapping `hideModal` to call both `hideRunOrChatDrawer` and `hideLogSheet`. LGTM!

PR #17928, opened 2026-08-06
fix(tokenizer): align important_kwd split to English comma (DSL parity, A2)

xugangqiang:
## Summary

Port the DSL tokenizer's `important_kwd` splitting into the Go `Tokenizer` component so the indexed keyword array is byte-compatible with the Python DSL pipeline and with the keyword-extraction prompt contract.

- **Problem:** The Go component split `keywords` on the full ASCII+CJK delimiter set (`utility.SplitKeywords`, regex `[,，;；、\r\n]+`), while the DSL baseline `rag/flow/tokenizer/tokenizer.py:153` uses `keywords.split(",")`, and `rag/prompts/keyword_prompt.md` instructs the LLM to delimit keywords by **ENGLISH COMMA**. For a dataflow canvas that includes the Tokenizer component, this divergence made Go's indexed `important_kwd` differ from the Python-DSL-built index (CJK commas/semicolons were split in Go but kept whole in Python).
- **Fix:** Use `strings.Split(kw, ",")` at `tokenizer.go:701`, preserving empty middle elements to match Python's `"a,,b".split(",") == ["a","","b"]`. The indexing fallback layer (`internal/ingestion/task/indexdoc/process.go`) already mirrors the Python multi-delimiter fallback (`dataflow_service.py:322`), so only the component layer diverged and only it is changed.

## Test plan

- `TestTokenizerComponent_ImportantKwd_CommaOnly` (no build tag, default `go test ./...`): switches the tokenizer to the identity engine (no CGo pool needed) and asserts `"kw1,kw2;kw3，kw4"` → `["kw1","kw2;kw3，kw4"]`; also asserts `important_tks` still tokenizes the full keyword string.
- `TestTokenizerComponent_Invoke_KeywordSplitCommaOnly` (`integration` tag, real CGo analyzer): covers comma-split, CJK/semicolon-not-split, and empty-middle preservation.
- Both tiers pass (unit `ok`, integration `ok`).

## Regression notes

- Intentional behavior change for canvases that include the Tokenizer component: keywords containing `;`/`、`/newlines now stay as one keyword (matching Python DSL) instead of being split. Re-indexing existing Go-built data will change the `important_kwd` set — expected parity cost, documented in code comments and commit message.
- Canvases without a Tokenizer component are unaffected (they hit the unchanged multi-delimiter fallback).
- Other fields (`important_tks`, `questions`, `summary`, `text`) are untouched; the `utility` import was removed cleanly.

### DIFF
--- a/internal/ingestion/component/tokenizer.go
+++ b/internal/ingestion/component/tokenizer.go
@@ -96,7 +96,6 @@ import (
 	"ragflow/internal/ingestion/component/globals"
 	"ragflow/internal/ingestion/component/schema"
 	"ragflow/internal/tokenizer"
-	"ragflow/internal/utility"
 )
 
 const ComponentNameTokenizer = "Tokenizer"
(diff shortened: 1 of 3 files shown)


yuzhichang:
LGTM

PR #17930, opened 2026-08-06
feat: add json file preview and icon

dcc123456:
### Summary

feat: add json file preview and icon


wangq8 (commented):
Simple and clean. The icon SVG looks good and the file type mapping in `file.ts` is correct. LGTM!

PR #17931, opened 2026-08-06
fix: record generated wiki page versions

buua436:

Generated Wiki pages were not consistently recorded in version history, and incremental pages used an incompatible slug format. This PR records versions for both full and incremental generation, cleans history when pages are deleted, and updates the frontend to use `wiki` as the task type.

### Type of change


wangq8 (commented):
Thanks for this thorough fix! Overall a well-structured PR that solves the wiki version history problem comprehensively. A few observations:

**Strengths:**
- Adding version recording in both  (full generation) and  (incremental) covers both paths
- Cleaning history on deletion (, ) prevents orphaned records
- Preferring `md_with_weight` as the canonical field aligns with what `persist_wiki_pages` already stores
- The `FileCommitService` methods for page history deletion are clean with proper atomic transactions

**Questions/Suggestions:**

1. **ES refresh call**: `refresh_idx` in `update_wiki_page` forces an index refresh on every wiki page update. For large indices this could add latency. Consider using `refresh=wait_for` at the write level in `docStoreConn.update()` instead, or note the performance implication here.

2. **Breaking change in knowledge.ts**: `ProcessingType.artifact = 'wiki'` changes a public enum value. Any code that references `ProcessingType.artifact` expecting the string `'Artifact'` will now get `'wiki'`. Is this intentional? If the task type value needs to match the backend, this should be noted as a coordinated change.

3. **Minor**: In `delete_page_history`, the `file_id` is computed via `_wiki_file_id(kb_id, slug)`, then you also filter by `slug_kwd == slug`. Since `_wiki_file_id` already incorporates the slug, the slug filter is redundant but provides defense-in-depth. Not a blocker.

4. **In `persist_wiki_pages` diff line ~909**: The guard `if not prior_by_slug.get(slug, ""): continue` was removed. This means first-time page creation will now also call `record_page_edit`. The intent is clear (tracking initial creation), but I notice the commit message for the eventual record_page_edit call passes `content_before=""` for new pages — is that handled gracefully in `record_page_edit` for empty-diff detection?

Otherwise LGTM, nice work!

wangq8 (commented):
Thanks for this thorough fix! Overall a well-structured PR that solves the wiki version history problem comprehensively. A few observations:

**Strengths:**
- Adding version recording in both `update_wiki_page` (full generation) and `_wiki_refine_page` (incremental) covers both paths
- Cleaning history on deletion (`clear_wiki`, `_wiki_delete_deleted_doc_state`) prevents orphaned records
- Preferring `md_with_weight` as the canonical field aligns with what `persist_wiki_pages` already stores
- The `FileCommitService` methods for page history deletion are clean with proper atomic transactions

**Questions/Suggestions:**

1. **ES refresh call**: `refresh_idx` in `update_wiki_page` forces an index refresh on every wiki page update. For large indices this could add latency. Consider using `refresh=wait_for` at the write level in `docStoreConn.update()` instead, or note the performance implication here.

2. **Breaking change in knowledge.ts**: `ProcessingType.artifact = 'wiki'` changes a public enum value. Any code that references `ProcessingType.artifact` expecting the string `'Artifact'` will now get `'wiki'`. Is this intentional? If the task type value needs to match the backend, this should be noted as a coordinated change.

3. **Minor**: In `delete_page_history`, the `file_id` is computed via `_wiki_file_id(kb_id, slug)`, then you also filter by `slug_kwd == slug`. Since `_wiki_file_id` already incorporates the slug, the slug filter is redundant but provides defense-in-depth. Not a blocker.

4. **In `persist_wiki_pages`**: The guard `if not prior_by_slug.get(slug, ""): continue` was removed. This means first-time page creation will also now call `record_page_edit`. The intent is clear (tracking initial creation), but does `record_page_edit` handle empty-diff saves gracefully for new pages?

Otherwise LGTM, nice work!

PR #17932, opened 2026-08-06
Fix: display embd_name in list memory api response

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/services/memory_api_service.py
+++ b/api/apps/services/memory_api_service.py
@@ -20,6 +20,7 @@ from api.db.services.user_service import UserTenantService
 from api.db.services.canvas_service import UserCanvasService
 from api.db.services.task_service import TaskService
 from api.db.joint_services.memory_message_service import get_memory_size_cache, judge_system_prompt_is_default, queue_save_to_memory_task, query_message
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix! This adds the missing `embd_name` field to the list memory API response.

**What changed:**
- Added `get_composite_model_name_by_ids` import from `api.db.joint_services.tenant_model_service`.
- Batches all `embd_id` lookups in a single call (efficient: one query instead of N).
- Adds `embd_name` to each memory dict via `.update()`, defaulting to `""` when the embd_id cannot be resolved.

**Review:**
✅ Batch lookup pattern is correct — avoids N+1 queries.
✅ `.get(memory["embd_id"], "")` handles unresolvable IDs gracefully.
✅ Works correctly for empty `memory_list` (no-op).
✅ The `get_composite_model_name_by_ids` function silently skips unresolved IDs, so this is safe even with stale/invalid `embd_id` values.

One small note: consider whether `format_ret_data_from_memory` (used in `get_memory_config` and `create_memory`/`update_memory` responses) should also include `embd_name` for consistency. That is a separate concern though, not blocking this PR.

LGTM! 🚀

PR #17934, opened 2026-08-06
fix(tree-select): allow collapsing parent nodes during search

dcc123456:
### Summary

fix(tree-select): allow collapsing parent nodes during search

During search, `visibleExpandedIds` recomputed all parent IDs from `filteredData` on every render, ignoring `expandedIds`. This made `handleToggle` a no-op, so clicking a parent node to collapse it had no effect while searching.

Replace the search override with a `useEffect` that auto-expands matching parents into `expandedIds` when the search term changes. Because the effect only depends on `searchTerm`/`filteredData`, a manual collapse (which updates `expandedIds`) is preserved rather than being immediately overridden.

### DIFF
--- a/web/src/components/tree-select.tsx
+++ b/web/src/components/tree-select.tsx
@@ -144,26 +144,32 @@ export const TreeSelect = forwardRef<HTMLButtonElement, TreeSelectProps>(
       [data, searchTerm, filterTree],
     );
 
-    const visibleExpandedIds = useMemo(() => {
-      if (!searchTerm) return expandedIds;
-      const ids = new Set<string>();
-      const walk = (nodes: TreeSelectNode[]) => {
-        for (const node of nodes) {
-          if (node.children?.length) {
-            ids.add(node.id);
-            walk(node.children);
+    // Auto-expand all matching parents when the search term changes so users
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The root cause is clear — during search, `visibleExpandedIds` always recomputed all parent IDs from `filteredData`, making `handleToggle` a no-op.

**What changed:**
- Replaced the `useMemo`-based `visibleExpandedIds` override with a `useEffect` that auto-expands matching parents into `expandedIds` when `searchTerm` changes.
- `renderTree` now reads `expandedIds` directly.
- Because the effect only depends on `searchTerm`/`filteredData`, manual collapse (which updates `expandedIds`) is preserved.

**Review:**
✅ Correctly fixes the bug — clicking a parent to collapse now works during search.
✅ Functional `setExpandedIds((prev) => ...)` avoids stale closure issues.
✅ New search triggers re-expansion of matching parents, which is the expected UX.
✅ When search is cleared, `expandedIds` is left as-is (consistent with the original behavior where `!searchTerm` returned `expandedIds` directly).

No regressions or edge-case issues spotted. LGTM! 🚀

PR #17937, opened 2026-08-06
Refactor(task): collapse duplicated debug-result helpers 

xugangqiang:
## Summary
Collapses three duplicated/dead code smells in the canvas debug-result path of `internal/ingestion/task` (remaining findings #2/#3/#4 from the package CTO review):

- **#2 — vector-key stripping duplication.** `isVectorKey` (the full stripper: fixed legacy keys `vector`/`embedding`/`feature`/`q_vec` plus the `q_<dim>_vec` pattern) was re-implemented as a weaker inline copy in the golden-compare tool (`tool/compare_pipeline_golden.go`) that only matched `q_<dim>_vec` and let real vectors leak into the diff. Exported as `IsVectorKey` and reused by the tool.
- **#3 — near-duplicate deep copy.** `deepCopy` and `deepCopyStrip` were identical walkers differing only in vector stripping. Parameterized `deepCopy(v any, stripVector bool) any` and deleted `deepCopyStrip`.
- **#4 — redundant switch.** `lookupComponentOutput` had two switch cases with identical bodies (both the `map[string]map[string]any` and `map[string]any` state shapes). Unified into a single `found`/`ok` resolution; nested-state and flat-fallback semantics unchanged.

## Why
These are the remaining items from the `internal/ingestion/task` review. They remove duplicated logic, a second recursive walker, and copy-paste branching, and make the golden-compare tool agree with the production vector stripping rule.

## Test plan
- `internal/ingestion/task/tool/compare_pipeline_golden_test.go`: `TestSanitizeProcessed_StripsVectorKeys` (red→green) asserts the tool now strips the full vector-key set (`vector`/`embedding`/`feature`/`q_vec`/`q_<dim>_vec`).
- Existing `debug_result_dsl_test.go` / `debug_log_sink_test.go` updated for the new `deepCopy(v, stripVector)` signature.
- `./build.sh --test ./internal/ingestion/task/...` passes; the full unit suite is green.

## Changes
- 6 files changed, +99 / -51

### DIFF
--- a/internal/ingestion/task/debug_log_sink.go
+++ b/internal/ingestion/task/debug_log_sink.go
@@ -271,7 +271,7 @@ func (s *DebugLogSink) Flush(ctx context.Context, finalErr error) {
 // `json.dumps(self.get_component_obj(self.path[-1]).output())`
 // (rag/flow/pipeline.py:171). The last executed component is the sink's final
 // entry (path[-1] always emits its exit progress last). Raw embedding vectors
-// are stripped (deepCopyStrip) so the Redis log stays at Python-scale size.
+// are stripped (deepCopy(out, true)) so the Redis log stays at Python-scale size.
 // Returns "" when no output is available; the caller then keeps the
 // plain-text fallback and the front-end export button stays disabled (empty
 // output, matching Python's isEmpty check).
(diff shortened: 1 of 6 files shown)


yuzhichang:
LGTM

PR #17939, opened 2026-08-06
fix(parser/chunk): dedupe CompileDelimiterPatternList active entries

xugangqiang:
## Summary

Restore the deduplication that was dropped when #17926 was merged. `CompileDelimiterPatternList` now keeps a `seen` set and collapses equivalent active entries (both backtick-inner and bare) into a single alternation. This PR also removes the dead code that the re-review surfaced.

Background from re-reviewing the merged code:
- The live delimiter paths call `CompileDelimiterPatternList` directly with the raw `[]string` config list:
  - main `delimiters` → `internal/ingestion/component/chunker/common.go:compileDelimPattern` → `CompileDelimiterListPattern`
  - `children_delimiters` → `token.go:compileChildrenPattern` → `CompileDelimiterPatternList(delims, true)`
- `ParseDelimiterField` also dedups, but it had **no non-test caller** in the codebase, so its dedup never reached production. The only place dedup can take effect on the live paths is inside `CompileDelimiterPatternList`.

## Why

Without this, repeated config values such as `["`#`", "#"]` (keepBare=true) compile to `#|#` — redundant regex bloat, and a deviation from the intended deduplicated behaviour. Functionally `#|#` ≡ `#`, so this is a low-severity hygiene/parity fix rather than a correctness bug.

## Changes

1. **Dedup in `CompileDelimiterPatternList`** — `seen` set collapses equivalent active entries. (The original intent of this PR.)
2. **Dead-code removal** — `ParseDelimiterField`, `NormalizeTextNewlines`, and `runesOf` had no production caller (only tests/comments referenced `ParseDelimiterField`); the live `[]string` list API is consumed directly via `CompileDelimiterPatternList`. Removed those helpers plus the orphaned grammar doc comment and the comments that pointed at the now-removed single-string parser. Test helpers that previously relied on `ParseDelimiterField` now mirror the live path (`CompileDelimiterPatternList([]string{...}, true)`), preserving their intent (case-sensitivity assertions still hold). No behavioral change on the live path.

## Test plan

- `internal/parser/chunk/delim_test.go`:
  - `TestCompileDelimiterPatternListDedup`: keepBare=true `["`#`","#"]` → `#`; keepBare=false repeated `["`##`","`##`","`#`"]` → `##|#`.
  - `TestCompileDelimiterPatternListDedupOrderIndependent`: dedup holds regardless of input order (`["#","`#`"]` → `#`; mixed repeated → `#`; wrapped duplicate at any position → `##|#`).
  - `TestCompileDelimiterListPatternDedup`: the main-delimiter live path (keepBare=false) also deduplicates (`["`##`","`##`","`#`"]` → `##|#`; bare entries ignored so only wrapped inner content participates).
  - `TestCompileDelimiterPatternListChildrenPrefixOrder`: keepBare=true `["##","#"]` → `##|#` and `FindString("###")=="##"`; multi-byte `[""。","###"]` → `###|。` (rune-count, not byte-length, ordering).
- `go test ./internal/parser/chunk/...` passes (existing `CompileDelimiter*` and `ChildrenDelimiter*` tests still green).

## Note for reviewers

- The `internal/ingestion/component/chunker` package currently fails to build in plain-`go test` environments that lack the native C++ static libs (`office_oxide.h`, `pdfium`, `pdf_oxide`, etc.) normally produced by `build.sh --cpp`. That is a pre-existing environmental condition unrelated to this PR; it is not introduced by the dead-code removal. The pure-Go `internal/parser/chunk` package (where the dedup lives) passes.

### DIFF
--- a/internal/ingestion/component/chunker/common.go
+++ b/internal/ingestion/component/chunker/common.go
@@ -80,8 +80,7 @@ func stringListFromAny(in []any) []string {
 // Only backtick-wrapped entries produce an active pattern (Python
 // token_chunker / rag/nlp/delim list helper). Plain entries are ignored here
 // and used by mergeByTokenSize for sentence-level splitting when no active
-// pattern exists. Canonical single-string parser_config.delimiter parsing
-// lives in ragflow/internal/parser/chunk (ParseDelimiterField).
+// pattern exists.
 func compileDelimPattern(delims []string) *regexp.Regexp {
 	return chunk.CompileDelimiterListPattern(delims)
 }
--- a/internal/ingestion/component/chunker/delimiter_case_sensitive_test.go
+++ b/internal/ingestion/component/chunker/delimiter_case_sensitive_test.go
@@ -37,7 +37,9 @@ import (
 )
 
 func getDelimiters(delimiters string) string {
-	p := chunk.CompileDelimiterPattern(chunk.ParseDelimiterField(delimiters))
+	// Mirror the live path: backtick-wrapped entries contribute their inner
+	// content (see chunk.CompileDelimiterPatternList).
+	p := chunk.CompileDelimiterPatternList([]string{delimiters}, true)
 	if p == nil {
 		return ""
 	}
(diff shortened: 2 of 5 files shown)


yuzhichang (approved):
LGTM

PR #17940, opened 2026-08-06
fix(web): prevent column shift in data source log table

euvre:
### Summary

In the data source detail log table (Knowledge Base -> Data Source -> Settings), the Summary column contains a live countdown (e.g. `Task starts in 59s`). The table used the browser's auto layout, so whenever the seconds value changed digit count (e.g. `59s` -> `9s`), the content width changed and the browser redistributed all column widths, causing the Summary column (and the whole table) to visibly shift back and forth every second. The existing `tabular-nums` class only keeps digits of the same count aligned and cannot prevent width changes when the digit count changes.

This PR fixes the layout shift by:

- Switching the table to a fixed layout (`table-fixed`) so column widths no longer depend on cell content.
- Assigning explicit widths to the first four columns via the existing `meta.headerClassName` mechanism (Time Started `w-44`, Status `w-28`, Dataset `w-1/4`, Task Type `w-28`), leaving Summary to take the remaining space.
- Adding `truncate` to the dataset name so long names do not overflow under the fixed layout.

### DIFF
--- a/web/src/pages/user-setting/data-source/data-source-detail-page/log-table.tsx
+++ b/web/src/pages/user-setting/data-source/data-source-detail-page/log-table.tsx
@@ -113,6 +113,7 @@ const columns = ({
     {
       accessorKey: 'update_date',
       header: t('setting.timeStarted'),
+      meta: { headerClassName: 'w-44' },
       cell: ({ row }) => (
         <div className="flex items-center gap-2 text-text-primary">
           {row.original.update_date
@@ -124,6 +125,7 @@ const columns = ({
     {
       accessorKey: 'status',
       header: t('knowledgeDetails.status'),
+      meta: { headerClassName: 'w-28' },
       cell: ({ row }) => (
         <FileStatusBadge
           status={row.original.status as RunningStatus}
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Clean fix for an annoying visual bug. The countdown-driven layout shift is a real UX issue, and the solution is minimal and well-scoped.

**table-fixed layout** — This is the correct approach. With auto layout, every column width recomputes when any cell content changes width. Fixed layout prevents that.

**Column widths** — The chosen widths (w-44 for time, w-28 for status/task type, w-1/4 for dataset) are reasonable. Time and task-type columns have predictably narrow content, and the dataset column gets a proportional share. Summary takes the remainder, which is fine since it has the most variable content.

**truncate on dataset name** — Necessary under fixed layout, since long dataset names would otherwise overflow their w-1/4 allocation. Good addition.

One thought: the hardcoded pixel widths (w-44, w-28) work well at the default viewport but may feel cramped or too generous at extreme resolutions. Not a blocker — this is a data table, not a landing page — and the w-1/4 on the dataset column provides some flexibility.

LGTM.

dcc123456 (approved):

PR #17942, opened 2026-08-06
fix(setting-model): saved instance api_key edits not triggering save

dcc123456:
### Summary

fix(setting-model): saved instance api_key edits not triggering save

getSavePayload's early-exit used form.formState.isDirty, which is unreliable when isDirty is only read through the imperative ref (in a click handler, not in a render) - RHF never establishes the formState Proxy subscription, so _formState.isDirty stays stale at its post-reset false, silently skipping api_key / base_url edits.

Replace with two reliable guards:
1. Skip saved cards whose instanceDetails haven't been lazy-loaded (collapsed cards have empty forms that would fail validation on the required api_key and abort the entire batch save).
2. Use the signature comparison (payload JSON vs baseline) as the dirty check, which catches every real edit.

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/hooks.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/hooks.tsx
@@ -602,14 +602,35 @@ export function useInstanceSaveState({
   // `getSavePayload()` is the imperative entry point the parent calls
   // when the user clicks the top Save button. For drafts it always
   // returns a payload (provided the name is non-empty); for saved
-  // cards it returns `null` when the card hasn't been touched by the
-  // user, so the parent skips both validation and the no-op PUT.
+  // cards it returns `null` when the current payload matches the
+  // last-synced baseline, so the parent skips the no-op PUT.
+  //
+  // Two guards skip cards that should not participate in the batch:
+  //
+  // 1. `!instanceDetails` - collapsed cards never lazy-fetch their
+  //    details (see `useLazyInstanceDetails`), so their form is empty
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice catch and thorough explanation of the root cause. The `isDirty`-via-ref issue is a classic RHF gotcha — without a render-phase subscription, the Proxy never wires up the `_formState` tracking.

**Root cause analysis** — The comment block explaining why `isDirty` was stale is excellent and will help future maintainers.

**Collapsed-card guard** — Skipping cards without `instanceDetails` is the right call. As noted, a user cannot have edited a card they never opened, and including them would fail validation on the required `api_key` field, aborting the entire batch save.

**Signature comparison** — Comparing the actual payload against the baseline is a reliable dirty check that doesn't depend on RHF internal state. Good approach.

**Dependency array** — Moving from `[buildPayload, isDraft, instanceName, formRef]` to `[buildPayload, isDraft, instanceDetails]` is correct:
- `formRef` is a ref (stable identity, no re-render trigger)
- `instanceName` was used via the old `editedNameRef.current !== instanceName` check, which was removed — the baseline comparison handles renames too

One minor suggestion: the `getSavePayload` function now returns `null` when `!isDraft && !instanceDetails`. For a saved card that was opened and then collapsed (which retains cached `instanceDetails` per the comment), it still participates. But for a saved card that was never opened, it is skipped. Consider adding a brief note in the parent component where `getSavePayload` is called, documenting that `null` now means either "no changes" or "never opened" — both cases where a no-op skip is safe.

LGTM.

PR #17943, opened 2026-08-06
chore(rag/app): remove stray debug print() calls

skbs-eng:
chore(rag/app): remove stray debug print() calls

Two hot-path debug print() calls were leaking content/error text to
stdout in production code paths.

* rag/app/naive.py: TxtParser branch in chunk() was printing the entire
  parsed sections list (formatted via repr()) wrapped in 150-char banner
  lines. For large text documents (e.g. a 1000+-page book ingest) this
  dumped tens of thousands of lines per ingest into the docker logs.
  Replaced with a structured
  `logging.info("TxtParser produced %d sections for %s", len(sections),
  filename)` so the parse count is still observable without the content
  leak.

* rag/app/presentation.py: Pdf.position parsing had a debug
  `print(f"Error parsing position: {e}")` inside an except clause in the
  ingest hot path. Replaced with
  `logging.warning(f"Error parsing position in {filename}: {e}")` to
  match the file's existing logging pattern and add filename context.

Both call sites already had logging imported; no new imports added.
logging was used throughout the surrounding code in the same
logging.{info,warning,error}(...) style.


skbs-eng:
@yuzhichang please review

yuzhichang (approved):

Issue #17938, opened 2026-08-06
Fix integration-test compilation breakage in pipeline / service/nlp / storage


xugangqiang:
## Summary
The integration-test tier does not compile on `main`. `./build.sh --test-integration` fails to **build** `internal/ingestion/pipeline`, `internal/service/nlp`, and `internal/storage` with compile-time errors in their `*_integration_test.go` files. These are **not** missing runtime dependencies — they are stale test code referencing APIs that were removed/renamed/changed on `main`.

## Reproduce
```bash
./build.sh --test-integration ./internal/ingestion/pipeline/ ./internal/service/nlp/ ./internal/storage/
```

## Errors
**internal/ingestion/pipeline**
- `real_storage_integration_test.go`: `undefined: server.Config`, `undefined: server.SetLogger`
- `cfg.Database` / `cfg.StorageEngine` — these fields are now unexported; the config type lives in `server/config` (`server/config.Config`), not `server.Config`
- `not enough arguments in call to stg.Put` / `stg.Remove` (storage client signature changed)
- `template_integration_test.go`: `fixedEmbedder does not implement component.Embedder (missing method BatchSize)`

**internal/service/nlp**
- `datasetnav_integration_test.go`: `undefined: server.SetLogger`, `undefined: engine.Init`

**internal/storage**
- `minio_test.go`: `undefined: server.MinioConfig`, `GetConfig().StorageEngine` unexported, and `storage.Health` / `Put` / `Get` / `Remove` / `ObjExist` signature mismatches

## Root cause
The integration test files reference symbols that have since been removed/renamed/changed on `main`:
- `server.Config` → `server/config.Config`
- `server.SetLogger` removed
- `component.Embedder` gained a `BatchSize` method
- storage client signatures changed (`stg.Put`/`Remove`/`Health`, `storage.Put`/`Get`/`Remove`/`ObjExist`/`Health`)

The test files were never updated to match, so the integration tier is silently broken while the unit tier stays green.

Note: `internal/ingestion/task` (incl. `task/tool`) integration tests **pass** — this breakage is unrelated to any recent task-package work and is pre-existing.

## Suggested fix
Update the three `*_integration_test.go` files to the current APIs:
- use `server/config.Config` instead of `server.Config`
- drop `server.SetLogger` / `engine.Init` usage
- implement `BatchSize` on the test embedder (`fixedEmbedder`)
- adopt the new storage-client method signatures

## CI
[...]

PR #17944, opened 2026-08-06
refactor: extract TruncateRunes to internal/utility and fix integration build (#17938)

xugangqiang:
## Summary

This PR bundles four changes (two originally intended, plus two refactors
that surfaced while touching the same files):

1. **Shared helper extraction (#9b)** — `truncateRunes` was duplicated in
   `internal/ingestion/task` and `internal/service`. It is now a single
   `utility.TruncateRunes(s string, max int) string` in `internal/utility`,
   with updated call sites in `debug_log_sink.go` and `agent_sessions.go`.
   A focused table test (`internal/utility/truncate_test.go`) covers
   empty/ASCII/Unicode rune-count truncation and the `max <= 0` case.

2. **Fix the `integration` test tier compile errors on `main` (#17938)** —
   the integration tier currently does not compile because several APIs were
   renamed/removed. This blocks enabling the integration tier in CI.

3. **Unify the vector-key stripping rule (production + golden tool)** —
   `internal/ingestion/task/debug_result_dsl.go` already strips raw embedding
   vectors via `isVectorKey` (fixed legacy keys `vector`/`embedding`/`feature`/
   `q_vec` AND the dimension-scoped `q_<dim>_vec` pattern the tokenizer emits).
   The golden-compare tool `internal/ingestion/task/tool/compare_pipeline_golden.go`
   instead used a weaker inline check
   `strings.HasPrefix(k, "q_") && strings.HasSuffix(k, "_vec")`, which only
   caught the dimension-scoped pattern and **let the fixed legacy keys leak** —
   producing spurious diffs against the expected output that already excludes
   them. This PR exports `isVectorKey` as `IsVectorKey` and has the golden tool
   reuse it, so both paths strip the identical set. As a side effect,
   `deepCopy` and `deepCopyStrip` are merged into `deepCopy(v any, stripVector bool)`
   (behavior-preserving) and `debug_log_sink.go` is updated to the merged call.
   New `compare_pipeline_golden_test.go::TestSanitizeProcessed_StripsVectorKeys`
   locks the full key set.

4. **Minor refactor in `debug_result_dsl.go`** — `lookupComponentOutput`
   collapses its duplicated per-case `if v, ok := state[id]; ok { return v }`
   into a single `found, ok` assignment returned after the `switch`, with no
   behavior change.

### #17938 fixes (compile-only; no behavior change)
- `pipeline/real_storage_integration_test.go`: return `*config.Config`,
  use `config.GetMySQLConfig()` / `config.GetMinioConfig()`, replace
  `server.SetLogger(zap.NewNop())` with `common.InitLogger(...)`, and pass
  `context.Context` to `storage.Put` / `storage.Remove`.
- `pipeline/template_integration_test.go`: add `fixedEmbedder.BatchSize()`
  (required by `component.Embedder`) and pass `ctx` to `storage.Put`.
- `service/nlp/datasetnav_integration_test.go`: `common.InitLogger` +
  `engine.InitDocEngine()`; drop the `zap` dependency.
- `storage/minio_test.go`: use `configpkg.MinioConfig` via
  `GetMinioConfig()` and pass `ctx` to every storage call
  (`Health`/`Put`/`Get`/`Remove`/`ObjExist`/`GetPresignedURL`/
  `BucketExists`/`RemoveBucket`/`Copy`/`Move`).

## Verification
- `bash build.sh --test ./internal/utility/... ./internal/ingestion/task/...`
  → unit tier **green** (`utility`, `task`, `task/indexdoc`, `task/tool`).
- `bash build.sh --test-integration ./internal/ingestion/pipeline/ \
  ./internal/service/nlp/ ./internal/storage/` → the integration tier now
  **compiles** for all three packages (previously build errors). Remaining
  runtime failures are environmental (no live MySQL/ES/MinIO in this
  workspace) or the pre-existing TokenChunker chunk-count assertion, and are
  unrelated to this mechanical API migration.

## Note
A separate issue tracks enabling the `integration` tier in CI (see linked
issue). `internal/handler/streaming_test.go` (an untagged **unit** test)
references the removed `server.Config` type and is a pre-existing break
outside the scope of #17938; calling it out for follow-up.

### DIFF
--- a/internal/ingestion/pipeline/real_storage_integration_test.go
+++ b/internal/ingestion/pipeline/real_storage_integration_test.go
@@ -22,9 +22,9 @@ import (
 	componentpkg "ragflow/internal/ingestion/component"
 	_ "ragflow/internal/ingestion/component/chunker"
 	"ragflow/internal/server"
+	"ragflow/internal/server/config"
 	"ragflow/internal/storage"
 
-	"go.uber.org/zap"
 	"gorm.io/driver/mysql"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
@@ -46,7 +46,7 @@ func TestPipelineRun_TemplateGeneral_RealMySQLMinIO_OutputShape(t *testing.T) {
 		t.Fatalf("auto-migrate real mysql tables: %v", err)
 	}
 
-	realStorage, err := storage.NewMinioStorage(cfg.StorageEngine.Minio)
+	realStorage, err := storage.NewMinioStorage(cfg.GetMinioConfig())
 	if err != nil {
 		t.Fatalf("connect real minio: %v", err)
 	}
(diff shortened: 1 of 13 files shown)


xugangqiang:
Integration-tier CI enablement tracked in issue #17945 (assigned to myself). This PR is the compile fix referenced there.

yuzhichang:
LGTM

PR #17946, opened 2026-08-06
Fix: recognize data type in parse method `table`

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/rag/app/table.py
+++ b/rag/app/table.py
@@ -19,6 +19,7 @@ import csv
 import io
 import logging
 import re
+from datetime import datetime
 from io import BytesIO
 from xpinyin import Pinyin
 import numpy as np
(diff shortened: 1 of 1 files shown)


yuzhichang:
Suggest keep go port synced in the same PR.

Lynn-Inf:
> Suggest keep go port synced in the same PR.

Go already does plain text — no fix needed.

yuzhichang:
LGTM

wangq8 (commented):
Thanks for this improvement! The native type checking is a clear upgrade over the regex-only approach, and handling NaN/NaT properly avoids subtle bugs with openpyxl/pandas data.

A few observations:

**Type checking order** ✅ — `bool` is checked before `int`, which is correct since `bool` is a subclass of `int` in Python. Good catch.

**NaN/NaT handling** — The `pd.isna()` checks in both the type-detection loop and the conversion loop are appropriate. The `datetime` + `pd.isna()` guard handles `pd.NaT` correctly.

**Defensive string guard in `chunk()`** — Skipping string values in non-text typed columns is a reasonable safeguard against ES type mapping errors. However, the `pass`-then-`else` pattern is a bit awkward stylistically:
```python
if isinstance(val, str):
    pass
else:
    stored[fld] = val
```
Could be simplified to:
```python
if not isinstance(val, str):
    stored[fld] = val
```

**Misleading comment placement** — The comment block:
```
# Keep original value from openpyxl/pandas instead of dropping to None.
# This preserves cells (e.g. text in numeric columns) that would
# otherwise be silently discarded by forced column-level conversion.
```
is placed right after `arr[i] = None`, which reads as if the value is being preserved when it is actually being dropped to None. The comment appears to describe the downstream effect in the chunk function, not the current line. Consider clarifying or relocating the comment.

**Large int overflow check** — `int(a) > 2**63 - 1` in the type-detection loop works, but will raise `ValueError` if the native int is actually larger than what Python can represent as int (unlikely in practice but worth noting).

**No tests** — This parser code is tricky and has had subtle type-detection bugs before. A few unit tests covering mixed-type columns, boolean cells, datetime cells, and NaN cells would be valuable.

Overall LGTM with minor nits.

PR #17948, opened 2026-08-06
fix(chunker): decide token merge on running sum, not re-tokenized join

xugangqiang:
## Summary

The Go `TokenChunker` merge decision used `tokenizeStr(joined)` — re-tokenizing the **entire joined chunk text** — to decide whether to merge the next unit. Because BPE tokenization (`cl100k_base`) is **non-additive across the `"\n"` join separator**, this count disagreed with Python's merge decision, which accumulates a **running sum of per-unit token counts**:
- `rag/nlp/__init__.py:_merge_paragraph_groups` → `size("\n" + sub_sec)`
- `rag/flow/chunker/token_chunker.py:_merge_text_chunks_by_token_size` → `tk_nums += current["tk_nums"]`

The disagreement shifted every downstream chunk boundary by a line and, on a tight budget, changed the chunk count (e.g. Go emitted **8** chunks where Python emitted **9** for a 29-line text at `chunk_token_size=32`).

## Fix

`mergeDecision` now decides on the running sum of per-unit token counts (`prevTokens + incomingTokens`) instead of `tokenizeStr(joined)`. Both callers maintain and pass that running sum:
- text path: `tkns` slice (per-paragraph `tokenizeStr("\n" + sub_sec)`)
- JSON path: `prev.TKNums` (upstream `tk_nums` accumulated via `+=`)

The joined text is still returned as the merged content; only the **decision metric** changed.

## Tests

- **Add** `TestTokenChunkerMergeMatchesPython` — pins the text-path merge against a Python oracle (running sum, OVER_CAP) for two budgets (`32`, `128`), asserting both chunk count and per-chunk text. Was RED (Go 8 vs Python 9) before the fix, now GREEN.
- **Update** three existing tests that pinned the old re-tokenized (under-packed) boundaries to their Python-faithful expectations: `TestMergeByTokenSizeFromJSON_StrictCapNoOvershoot`, `TestMergeByTokenSize_UnderCapNoOverflow`, `TestMergeByTokenSizeFromJSON_OverlapStripsTags`. Their original intent (strict-cap bound, `under_cap` toggle liveness, overlap tag-stripping) is preserved.

Verified the new boundaries against real Python/tiktoken: the running-sum produces Python's exact chunking for all four inputs.

## Impact / follow-up

This is the root cause behind the two golden-parity `go_bug` rules `token-over-segmentation-long-input` and `token-default-delim-content-mismatch`. The text path now matches Python exactly. The golden-parity branch (`fix/chunker-parity-infra`, with its `known_diffs.json` ratchet) should be re-run; these rules are expected to **resolve**, and any residual JSON-path boundary shift must be re-ratchet'd there. I could not run that harness in this environment because its fixtures live on a separate branch.

Note: the JSON path keeps the `OVER_CAP` strategy (merge-then-close) rather than Python's JSON `prev_tk_nums > threshold` decision; that is a pre-existing, separate strategy difference not addressed by this counting fix.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -398,7 +398,10 @@ const (
 // mergeDecision computes the merge decision shared by the text and JSON merge
 // paths. prevText is the current chunk, incoming is the next unit, and joinSep
 // is the separator used to project the joined text ("" for the text path, "\n"
-// for the JSON path). target is the token cap.
+// for the JSON path). target is the token cap. prevTokens is the running sum of
+// per-unit token counts already accumulated into prevText; incomingTokens is
+// the token count of incoming, counted the same way the calling path counts a
+// unit.
 //
 // strategy selects the merge strategy (schema.MergeStrategy), mirroring
 // Python's MergeStrategy:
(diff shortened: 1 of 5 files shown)


yuzhichang:
LGTM

PR #17949, opened 2026-08-06
Refine agentic research

qinling0210:
### Summary

Refine agentic search — fixes critical correctness bugs and improves planning/decomposition quality in the agentic RAG pipeline.

**Bug fixes**
ABSTAIN path was dead code → now correctly surfaces "I cannot answer"

agentic.py: _finalize now accepts abstain parameter, ABSTAIN path passes abstain=True → formalize_answer abstain branch (line 215) is reachable. Previously ABSTAIN silently became a normal partial answer using unreliable evidence. FALLBACK_LLM is unaffected.
agentic_rag_graph.py: log _note priority fixed — abstain now checked before partial.
Single-keyword search zeroed out all results

search.py: len(kwds) < 3 → 1 < len(kwds) < 3. A single keyword (e.g. "颜良") previously entered else → range(0) → kwds=[] → all chunks dropped. Now single-keyword queries are preserved as-is.
UNANSWERABLE misclassified as CONFLICTING, Chinese NER false negatives

sufficiency.py: UNANSWERABLE check moved before CONFLICTING. total == 0 (no extractable entities/numbers from Chinese text) now passes instead of scoring 0.
Chinese-only chunks with no English entities now use lenient matching (any digit match passes).
Preliminary search polluted final answer citations

agentic_rag_graph.py: removed _merge_result_into_kbinfos(). Seed chunks from preliminary search are only stored in state["seed_chunks"] for planner grounding — never merged into the shared citation pool.

**Quality improvements**

Route classifier (route_prompt.py): added explicit decomposition conditions + 6 bilingual examples → "曹操认识谁？" now correctly routes exploratory → decompose.

Decomposition prompt (decompose_prompts.py): enumeration questions now use a single discovery query instead of per-candidate verification → avoids claim collapse for questions like "who did X know?".

Language alignment (planner.py): runtime instruction forces claim descriptions to match the user's question language.

Go planner parity (planner.go): enumeration guidance and WRONG/CORRECT examples aligned with Python side.

### DIFF
--- a/rag/advanced_rag/harness/sufficiency.py
+++ b/rag/advanced_rag/harness/sufficiency.py
@@ -2,13 +2,13 @@
 
 import logging
 
+from rag.advanced_rag.harness.config import get_mode
 from rag.advanced_rag.harness.types import (
     AgentResult,
     ClaimCrossCheckResult,
-    SufficiencyVerdict,
     ExecutionStrategy,
+    SufficiencyVerdict,
 )
-from rag.advanced_rag.harness.config import get_mode
 
 _LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


rag/advanced_rag/harness/sufficiency.py
```diff
@@ -58,19 +58,41 @@ def cross_check_claim(agent_result: AgentResult, all_chunks: dict) -> ClaimCross
...
         for ent in entities:
             if ent.lower() not in text_lower:
                 mismatches.append(f"entity '{ent}' not found in chunk {eid}")
 
     total = len(matches) + len(mismatches)
-    cross_score = len(matches) / max(total, 1) if total > 0 else 0.0
-    cross_passed = len(mismatches) < len(matches) * 0.5
+    # When no extractable entities or numbers exist to cross-check against
+    # (common for Chinese text), treat as passed — nothing falsified the claim.
+    if total == 0:
```
JinHai-CN, line 81: Why total == 0 is passed?

PR #17953, opened 2026-08-06
fix: self-heal stale chunk dynamic import failures

euvre:
### Summary

Navigating to a lazily loaded route (e.g. Knowledge Base -> View All -> `/datasets`) can fail with `TypeError: error loading dynamically imported module`, and the error is reproducible across logout/login.

Root cause: nginx serves `index.html` without any `Cache-Control`, so browsers heuristically cache it. A stale `index.html` references hashed chunk files that no longer exist after an upgrade, and `try_files ... /index.html` serves the HTML fallback (200) for the missing `/chunk/js/*.js` request, breaking dynamic `import()`. Since logout/login are full-page navigations that reload the same stale cached shell, the error keeps reproducing.

Changes:

- **nginx (go mode, `docker/nginx/ragflow.conf.golang`)**: serve `index.html` with `Cache-Control: no-cache` so chunk references are always revalidated; return 404 for missing static assets instead of the `index.html` HTML fallback (also replaces the stale `/static/(css|js|media)/` rule, since build output actually lives under `/chunk/`, `/entry/` and `/assets/`).
- **web (`fallback-component`)**: detect dynamic import failures (Chrome/Firefox/Safari error variants) in the route error boundary and reload the page once to self-heal, guarded by a 10s `sessionStorage` timestamp to prevent reload loops.

### DIFF
--- a/docker/nginx/ragflow.conf.golang
+++ b/docker/nginx/ragflow.conf.golang
@@ -30,8 +30,16 @@ server {
         try_files $uri $uri/ /index.html;
     }
 
-    # Cache-Control: max-age Expires
-    location ~ ^/static/(css|js|media)/ {
+    # index.html must always be revalidated: it references hashed chunks,
+    # a stale copy points at chunk files that no longer exist after upgrades.
+    location = /index.html {
+        add_header Cache-Control "no-cache" always;
+    }
+
+    # Hashed build assets: cache aggressively, and return 404 instead of
+    # falling back to index.html (HTML breaks dynamic import()).
+    location ~* \.(?:js|mjs|css|png|jpg|jpeg|gif|ico|svg|webp|woff2?|ttf|eot|map)$ {
+        try_files $uri =404;
         expires 10y;
         access_log off;
     }
--- a/web/src/components/fallback-component/index.tsx
+++ b/web/src/components/fallback-component/index.tsx
@@ -14,7 +14,7 @@
  *  limitations under the License.
  */
 
-import React from 'react';
+import React, { useEffect } from 'react';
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
Nice thorough fix tackling both the server and client sides of the stale-chunk problem.

**Nginx changes (`ragflow.conf.golang`):**
- `Cache-Control: no-cache` on `index.html` — exactly right. Forces revalidation so browsers never serve a stale shell referencing deleted chunk hashes
- Replacing the old `/static/(css|js|media)/` rule with an extension-based regex `\.(?:js|mjs|css|...)$` is correct — the build output lives under `/chunk/`, `/entry/`, and `/assets/`, which the old prefix rule never matched
- `try_files $uri =404` for static assets is the key fix: without it, missing chunks fall through to the `/` location which serves `index.html` as a fallback — that HTML-as-JS breaks `import()` and is the root cause of the error. Returning 404 is the right behavior

**Client-side changes (`fallback-component/index.tsx`):**
- Error patterns cover Chrome, Firefox, and Safari variants — good coverage
- 10-second `sessionStorage` guard prevents reload loops while allowing a single self-heal — reasonable balance
- The guard key uses a namespaced prefix (`ragflow:`) which avoids collisions

**One note (non-blocking):**
- The `DYNAMIC_IMPORT_ERROR_PATTERN` regex is case-insensitive, which is good. The Firefox error message is "error loading dynamically imported module" (lowercase "error") — confirmed this matches.

LGTM, approved.

wangq8:
resolve conflict. - @euvre

wangq8:
potentical issue.

PR #17954, opened 2026-08-06
refactor(go): clean up TokenChunker delimiter routing and children appliers

xugangqiang:
## Summary

Cleanup of the Go TokenChunker layer (H1 `DelimiterMode` is intentionally left for a separate change).

- **H2 — unreachable branch**: `invokeTextPayload` had a split-then-merge branch calling `mergeByTokenSizeFromJSON` that could never execute. `delimPattern` is non-nil only when a backtick (custom) delimiter exists, so `hasCustomDelim` is always true at that point and the function already returns one chunk per segment. The dead branch is removed; the tail now simply returns `chunkOutputs(docs)`. `mergeByTokenSizeFromJSON` itself is kept — the JSON path still uses it.
- **M1 — duplicate children applier**: `applyChildrenDelim` is removed. Its `[]string`→`[]ChunkDoc` wrap is folded into the call site and delegated to `applyChildrenDelimText` (byte-identical output for both the nil-pattern and split branches). `splitByChildren` is deliberately kept: it clones and preserves chunk metadata for the JSON path, so it is not equivalent and must not be merged into the text-path applier.
- **M2 — dead sentinel**: `hasActiveDelimiter` reduced to `p != nil`. No compile function ever emits the `\A(?!)`, so the extra comparison was dead.
- **M3 — misleading comment**: `compileDelimPattern`'s doc claimed plain entries are "used by mergeByTokenSize for sentence-level splitting". They are ignored; `mergeByTokenSize` uses its own hardcoded `sentenceDelimiter`. Comment corrected.

## Test plan

- `bash build.sh --test ./internal/ingestion/component/chunker/... ./internal/parser/chunk/...` passes.
- Existing children-delim tests (`TestTokenChunker_ChildrenDelimiterDropped*`, `TestTokenChunker_ChildrenDelimiterBacktickStripped`) still pass, confirming the `applyChildrenDelim`→`applyChildrenDelimText` fold preserved text-path output.
- JSON-path behavior is untouched (`splitByChildren`, `mergeByTokenSizeFromJSON` retained).

Independent of #17920 (merged) and #17926 (merged). H1 (`DelimiterMode`) deferred to a follow-up.

### DIFF
--- a/internal/ingestion/component/chunker/common.go
+++ b/internal/ingestion/component/chunker/common.go
@@ -78,9 +78,10 @@ func stringListFromAny(in []any) []string {
 
 // compileDelimPattern compiles a TokenChunker-style []string delimiter list.
 // Only backtick-wrapped entries produce an active pattern (Python
-// token_chunker / rag/nlp/delim list helper). Plain entries are ignored here
-// and used by mergeByTokenSize for sentence-level splitting when no active
-// pattern exists.
+// token_chunker / rag/nlp/delim list helper); plain (non-backtick) entries
+// are ignored here. mergeByTokenSize uses its own hardcoded sentenceDelimiter,
+// not this list. Canonical single-string parser_config.delimiter parsing
+// lives in ragflow/internal/parser/chunk (ParseDelimiterField).
 func compileDelimPattern(delims []string) *regexp.Regexp {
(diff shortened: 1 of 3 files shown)


yuzhichang (approved):
LGTM

PR #17955, opened 2026-08-06
fix(document-preview): handle namespace-prefixed xlsx in ExcelCsvPreviewer

dcc123456:
### Summary

fix(document-preview): handle namespace-prefixed xlsx in ExcelCsvPreviewer

ExcelJS fails to parse xlsx files whose root <workbook> element carries an XML namespace prefix (e.g. <x:workbook>), common in WPS Office exports. The workbook-xform parser only matches the exact tag name "workbook", so a prefixed root leaves the model undefined and crashes @js-preview/excel with "Cannot read properties of undefined (reading 'sheets')".

Add normalizeXlsxForExcelJS() to detect such prefixes and re-serialize the file with SheetJS, which always emits prefix-free XML. Also defer previewer.destroy() in the error catch so the library's pending setTimeout(0) callback can safely access workbookDataSource before it is nulled out, avoiding a secondary "Cannot read properties of null (reading '_worksheets')" error.

### DIFF
--- a/web/src/components/document-preview/hooks.ts
+++ b/web/src/components/document-preview/hooks.ts
@@ -8,6 +8,7 @@ import { useDebounceFn, useSize } from 'ahooks';
 import axios from 'axios';
 import JSZip from 'jszip';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import * as XLSX from 'xlsx';
 
 // ZIP file header bytes "PK"
 const ZIP_HEADER_0 = 0x50;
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Well-documented fix.

The core issue — ExcelJS's workbook-xform parser only matching exact tag name `workbook` — is correctly diagnosed. Files exported by WPS Office use namespace-prefixed `<x:workbook>` roots that leave the model undefined.

**`normalizeXlsxForExcelJS()`:**
- The detection logic (`/<\w+:workbook[\s>]/`) is precise — it only catches prefixed roots, not any random XML with workbook in it.
- Using SheetJS (`xlsx`) to re-serialize is the right approach since it always emits prefix-free XML.
- The try/catch fallback is correct — if SheetJS can't parse the file, just pass the original data through and let the previewer surface its own error.

**Deferred `previewer.destroy()`:**
- Good catch on the setTimeout(0) race. The library schedules work via `setTimeout(0)` during error handling, and destroying immediately nulls out `workbookDataSource` before that callback runs. Deferring by one tick avoids the secondary crash without any user-visible delay.

One minor question: `XLSX.write` returns `ArrayBuffer` already — the `as ArrayBuffer` cast might be redundant depending on the TypeScript version and `xlsx` type definitions. If the types resolve correctly without the cast, it could be dropped.

PR #17956, opened 2026-08-06
fix(memory): display embd_name with fallback to embd_id in form field

dcc123456:
### Summary

fix(memory): display embd_name with fallback to embd_id in form field

The memory form field suffix was showing the raw embedding model ID (embd_id) instead of the human-readable name. Use embd_name when available, falling back to embd_id for backward compatibility.

### DIFF
--- a/web/src/components/memories-form-field.tsx
+++ b/web/src/components/memories-form-field.tsx
@@ -62,7 +62,7 @@ export function useDisableDifferenceEmbeddingMemory(name: string) {
             name={item.name}
           />
         ),
-        suffix: <MemoryLabel text={item.embd_id} />,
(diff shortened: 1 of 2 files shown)


wangq8 (approved):
Clean fix, LGTM.

**What this does:** The memory form field suffix was showing the raw `embd_id` (e.g. a UUID) instead of the human-readable `embd_name`. The backend already returns `embd_name` in memory_api_service.py — this PR correctly surfaces it in the UI.

**Review:**
- Interface change (`embd_name?: string`) is optional → fully backward compatible
- `embd_name || embd_id` fallback handles any edge case where the field is missing
- `MemoryLabel` component already accepts arbitrary `string` → no rendering risk
- No style issues, regressions, or security concerns

Thanks!

PR #17960, opened 2026-08-06
refactor[Go]: remove chunker dead code & fix stale/false descriptors (L1/L2/L3)

xugangqiang:
## Summary

TDD-driven cleanup of the Go ingestion Chunker layer (File → Parser → Chunker → Extractor → Tokenizer). Scope: Go only.

- **L1 — dead code**: deleted `internal/ingestion/component/chunker/image.go` (`concatImg` + `imgBytes`, zero callers repo-wide) and the unused `emptyChunkDocs` helper in `common.go`. The chunker package no longer imports `image`/`draw`/`png`.
- **L2 — stale comment**: the package doc and `MustRegisterChunker` claimed "four" chunker components/files; there are actually **nine** registered variants. Generalized the wording.
- **L3 — false descriptors**: `ChunkerInputs["name"]` said "Required by the upstream payload convention" — misleading, since `name` is resolved from the workflow-wide `CanvasState.Globals` bag (`globals.GlobalOrInput`) when absent. `ChunkerOutputs` falsely listed `name`/`tenant_id`/`kb_id` as pass-through, but `chunkOutputs` only emits `{output_format, chunks}`; those keys are never on the chunker output edge — the downstream Tokenizer reads them from globals. Removed the three never-emitted keys and documented the globals mechanism.
- **T1/T2 (TDD regression guards, added before the change to lock current behavior)**:
  - bare delimiter entry is ignored by `CompileDelimiterListPattern`, so `delimiter_mode` + a bare delimiter is a no-op (text merged by token_size, bare token preserved, not split).
  - a backtick-wrapped multi-byte (CJK) delimiter contributes its inner content as the split pattern, and the longest delimiter wins over a shorter prefix (rune-descending sort).
- **T3** (dead `image.go` helper) is resolved by the L1 deletion.

## Self-review

1. **Problem solved?** Yes. Dead code (`image.go`, `emptyChunkDocs`) verified zero-callers via repo-wide grep; stale "four" count corrected to nine (verified registrations); false descriptors corrected against the actual `chunkOutputs` emission and the downstream Tokenizer's `globals.GlobalOrInput` reads.
2. **Better approach?** Deletion (over keeping shims) follows AGENTS.md. For L3, removing the never-emitted keys is more accurate than relabeling them "not emitted". The `knowledge_compiler` package keeps its own local `chunkerOutputs` map (deliberately decoupled from CGO) and genuinely emits those keys (its test asserts it), so it is untouched and correct.
3. **Coverage / regression?** T1+T2 guard the delimiter contract changed in #17926. `ChunkerOutputs` is read-only registration metadata with no runtime key validation; no chunker test asserts `out["name"]`/`out["kb_id"]`/`out["tenant_id"]`. Full `./internal/ingestion/...` suite passes (incl. `image_upload.go` and `knowledge_compiler`), and `./internal/parser/chunk/...` passes.

## Test plan

`bash build.sh --test ./internal/ingestion/component/chunker/... ./internal/parser/chunk/... ./internal/ingestion/...`

## Checklist

- [x] Tests added/updated (T1/T2) and pass
- [x] `go vet`/build clean via `build.sh --test`
- [x] No dead code or stale comments retained


yuzhichang (approved):
LGTM

PR #17963, opened 2026-08-07
fix: preserve page index chapter chunk provenance

buua436:

PageIndex titles were associated only with the chunk containing the chapter heading. When a chapter was split across multiple chunks, the following content chunks were not included in the title provenance.

This change extends each title's `source_chunk_ids` to all chunks before the next title, while also prioritizing title extraction in the PageIndex prompt.

### Type of change

### DIFF
--- a/api/db/init_data/compilation_templates/page_index.yaml
+++ b/api/db/init_data/compilation_templates/page_index.yaml
@@ -12,45 +12,25 @@ config:
     - Do not merge unrelated content. Do not invent, rewrite, or return source text; return only chunk grouping metadata.
   entity:
     description: >-
-      You are a source-grounded document index extractor. First extract every
-      supported heading in source order, then extract the important information
-      belonging to each heading. Headings form the hierarchy; facts, dates,
-      numbers, names, and conclusions are detail information that should be
-      captured in concise `fact` or `conclusion` entities attached below the
-      nearest relevant heading through `include` relations. Do not omit a
-      heading because its section is short, and do not return only headings.
-
-      A title entity's `name` must be the exact clean heading text. A detail
-      entity's `name` must be a concise, source-grounded label or statement,
-      and its `description` must preserve the useful compressed information.
-      Keep detail entities atomic: do not combine unrelated facts, dates,
-      numbers, names, or conclusions into one item. Preserve important values,
-      units, dates, qualifiers, and conclusions. Do not invent content.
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this fix, this is a well-structured PR that addresses a real issue with PageIndex title provenance.

## What works well

1. **Core fix is correct** – `assign_page_index_title_chunk_spans` properly computes the chunk span for each title (from its own chunk position to just before the next title). Sorted by chunk position, edge case for the last title (spans to end of chunks) is handled. The early return in `_commit_result` for page_index ensures we wait for all batches before computing spans, and the flush after span assignment is correctly placed.

2. **Prompt simplification is a net improvement** – Removing the length constraints (≤25/80 chars) and verbose meta-description rules reduces prompt noise. The explicit enumeration of Chinese chapter markers (第N回, 第N章) will help LLMs catch these headings that lack Markdown syntax. Making "title-first" extraction explicit should reduce false positives where chapter headings were misclassified as facts.

3. **Exact-name entity merge before KNN dedup** – `_struct_merge_exact_named_entities` is a smart optimization. Merging same-name entities (casefold) before similarity grouping avoids redundant LLM calls and prevents drift. The fallback behavior (appending to `unchanged` on merge/re-embed failure) is safe.

## Observations – not blockers

1. **`assign_page_index_title_chunk_spans` depends on dict insertion order** – `list(chunks)` iterates the dict, which preserves insertion order on Python 3.7+. This is fine but worth a brief comment that `chunks_by_id` must be ordered by source position when passed in, since the function has no ordering guarantee otherwise.

2. **Prompt lost some guardrails** – The old prompt had "Do not write meta descriptions such as `heading for ...`, `section heading`" and "Do not include image captions, `Main article` links". These may not matter much with modern models, but if you see title quality regressions post-merge, re-adding a line about avoiding boilerplate would help.

3. **`_struct_merge_exact_entity_payload` type field** – The merged `type` field is set to lowercase (e.g. `"title"`), while the existing payload from the left dict may carry the original case. If any downstream code does case-sensitive type checks on the raw payload, this could cause a mismatch. I did not find such code, but worth noting.

Overall LGTM. The title provenance fix is clean, the dedup improvements are a nice bonus. 

Reviewed-by: lhp@openclaw

KevinHuSh (approved):

PR #17964, opened 2026-08-07
Fix: Start a new conversation; an additional conversation entry will appear once the chat ends.

cike8899:
### Summary

Fix: Start a new conversation; an additional conversation entry will appear once the chat ends.


wangq8 (commented):
Thanks for the fix! The root cause is clear: when a temporary conversation is persisted to the backend, the server assigns a different ID from the client-generated one. The old merge logic kept the temp item (by its client-side ID) and also added the server-persisted item (with its new server ID), resulting in a duplicate entry.

Replacing with `setList([...conversationList])` correctly resolves this since the server data is now the single source of truth.

One suggestion: please remove the commented-out useEffect block before merging — dead code adds confusion for future readers. Otherwise LGTM.

PR #17965, opened 2026-08-07
Fix: Deduplication of wiki entity types

cike8899:
### Summary

Fix: Deduplication of wiki entity types

### DIFF
--- a/web/src/components/originui/select-with-search.tsx
+++ b/web/src/components/originui/select-with-search.tsx
@@ -55,7 +55,8 @@ export type SelectWithSearchFlagProps = {
   placeholder?: string;
   emptyData?: string;
   allowCustomValue?: boolean;
-  onNoMatchEnter?(searchValue: string): void;
+  // Return false to veto selecting the custom value on Enter
+  onNoMatchEnter?(searchValue: string): boolean | void;
   disableAutoSelectOnEnter?: boolean;
   testId?: string;
   optionTestIdPrefix?: string;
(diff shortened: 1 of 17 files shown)


wangq8 (commented):
Good work on this PR. Here's my review:

**Deduplication logic (core fix):**
- `getAvailableTypeOptions` correctly filters out already-used types while preserving the editing field's type — right approach
- Using `useWatch` in `useTemplateSectionData` to reactively track existing fields is clean
- The `isDuplicateType` check with `onNoMatchEnter` veto covers the custom-value path, and the form `validate` rule covers the dropdown-select path — both paths are handled

**Required field validation:**
- Making `type` and `description` required is a good UX improvement
- The i18n keys `fieldTypeRequired` and `fieldDescriptionRequired` already exist in both en.ts and zh.ts

**`onNoMatchEnter` return type change:**
- Changing from `void` to `boolean | void` is backward-compatible
- The `e.preventDefault()` guard when returning `false` correctly prevents cmdk from selecting the custom value item

**Cleanup:**
- Removing the unused `WorkflowPreviewSheet` import and props is good housekeeping

**Nit:** In `edit-next/hooks/use-add-field-form.ts`, the `useEffect` changed from `form.reset(buildField(firstType))` to `form.reset(createEmptyField(fieldKeys))`. This is actually the right call now — since available type options are filtered, the first option might not always be a valid default. Just confirming this is intentional.

LGTM, approved.

PR #17968, opened 2026-08-07
fix: allow OpenAI-API-Compatible models as tenant default models

euvre:
### Summary

Models added via the **OpenAI-API-Compatible** provider could not be kept as tenant default models (LLM/Embedding/Rerank/etc.) on the Go backend: selecting one in the *Set Default Model* page appeared to save, but the field showed up empty again on reload.

Root cause: `TenantService.GetModelInfo` (used by `ListTenantDefaultModels`, i.e. `GET /api/v1/models/default`) validated the saved default model against the factory model catalog via `GetModelByName`. OpenAI-API-Compatible ships an empty catalog (`conf/models/openai-api-compatible.json` has `"models": []`) because its models are entirely user-defined, so the lookup always failed and `ListTenantDefaultModels` silently dropped the entry (`if err == nil` guard).

Fix: drop the factory-catalog validation in `GetModelInfo` and rely on the `tenant_model` row (exists + active) as the source of truth — this matches the Python backend's `_get_model_info` behavior, which performs no catalog check. The now-unused `factoryModelTypeName` helper is removed as well.

Verified: `bash build.sh --test ./internal/service/...` passes.

### DIFF
--- a/internal/service/tenant.go
+++ b/internal/service/tenant.go
@@ -449,17 +449,6 @@ func ptrStringValue(s *string) string {
 	return *s
 }
 
-func factoryModelTypeName(modelType string) string {
-	switch modelType {
-	case "image2text":
-		return "vision"
-	case "speech2text":
-		return "asr"
-	default:
-		return modelType
-	}
-}
-
 // GetDefaultModelName returns the full default model ID for a tenant and model type
 // Format: modelName@instanceName@providerName or modelName@providerName
 // Returns empty string if no default model is set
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Good catch. The Go backend was validating tenant default models against the factory catalog via `GetModelByName`, but OpenAI-API-Compatible ships with an empty `models.json` catalog (user-defined models only), so the lookup always failed and `ListTenantDefaultModels` silently dropped the entry.

**Changes look correct:**
- Removing the catalog-based validation in `GetModelInfo` and relying purely on the `tenant_model` row (exists + active) as the source of truth mirrors the Python backend's `_get_model_info`, which performs no catalog check.
- Dropping the now-unused `factoryModelTypeName` helper is clean — no dead code left behind.
- The diff is small (+4/-22) and focused on a single file.

**No concerns with:**
- Regression risk: the `tenant_model` row check (existence + active status) is still intact, so models that were never created by the tenant still get rejected.
- The `rsplitN` parsing + special cases (OCR deepdoc, TEI Builtin) are untouched.

Test results reported clean (`bash build.sh --test ./internal/service/...` passes). LGTM, approve.

Lynn-Inf (approved):
LGTM

Issue #17916, opened 2026-08-06
Go Tokenizer deliberately drops content-less chunks & sets chunk_order_int unconditionally (deviation from DSL)


xugangqiang:
## Decision (updated)

The Go `Tokenizer` component (`internal/ingestion/component/tokenizer.go`) deliberately diverges from the Python DSL `rag/flow/tokenizer/tokenizer.py` on two points, **for user value**, not as a parity defect:

### A5 — Drop a chunk only when both `text` and `content_with_weight` are empty
- Replaces the old `isPhantomChunk` 4-field check. A chunk with neither `text` nor `content_with_weight` has **no retrievable content**: retrieval returns only `content_with_weight` to the user (`internal/service/nlp/retrieval.go:361`), and for question-only blocks it is even negative value (matched by a question term but returns an empty answer).
- `content_with_weight` stays in the guard to preserve the **Parser path** (blocks carry `content_with_weight` without `text`); `normalizeChunkTextFallback` backfills `text` afterward. The filter runs *before* normalize, so this guard is required.

### A6 — Set `chunk_order_int` unconditionally on every surviving chunk
- Consolidates the previous two assignment sites into one loop placed **after the A5 filter and after `normalizeChunkTextFallback`**, yielding contiguous 0-based indices on all paths (`chunks`/`json`/`markdown`/`text`/`html` × `full_text`/`embedding`). `chunk_order_int` is used only for retrieval ordering (`retrieval.go:619`).

## Why this is a deviation, not parity

- **Set difference (A5):** Go drops `text ∧ content_with_weight` empty blocks; the Python DSL keeps every non-`None` block.
- **Order difference (A6):** Go sets `chunk_order_int` on all chunks; the Python DSL sets it only for `chunks + full_text`.
- Because Go's surviving set is smaller *and* set unconditionally, the `chunk_order_int` values differ from Python across the board.

## Why the old "remove the deletion / it's the Chunker's job" framing is wrong

The earlier write-up asserted that rejecting empty chunks is the Chunker's responsibility and the Tokenizer must pass chunks through. That does not hold: the Python Chunker also does **not** drop non-`None` structured blocks, so Go's over-dropping lives entirely in the Tokenizer, not in a missing Chunker step. Keeping content-less chunks in Go would merely reintroduce phantom (unvectorized, content-less) blocks into the index — the opposite of helpful.

## Impact

[...]

PR #17970, opened 2026-08-07
fix(go): drop content-less chunks and set chunk_order_int unconditionally in Tokenizer

xugangqiang:
## Summary

Two user-value-driven fixes in `internal/ingestion/component/tokenizer.go`, both **deliberate deviations** from the Python DSL (tracked as Go-side issue **#17916**).

### A5 — Drop a chunk only when both `text` and `content_with_weight` are empty
- Replaces the old `isPhantomChunk` 4-field check (`Text && Image && ContentWithWeight && Summary`).
- Retrieval returns only `content_with_weight` to the user (`internal/service/nlp/retrieval.go:361`), so a chunk with neither `text` nor `content_with_weight` has **no retrievable content**; for question-only blocks it is even negative value (matched by a question term but returns an empty answer).
- `content_with_weight` is kept in the guard to preserve the **Parser path** (its blocks carry `content_with_weight` without `text`); `normalizeChunkTextFallback` backfills `text` afterward. Filter runs *before* normalize, so this guard is required to avoid killing legitimate Parser blocks.

### A6 — Set `chunk_order_int` unconditionally on every surviving chunk
- Consolidates the previous two assignment sites (Invoke guard + `tokenizeChunks`) into one loop placed **after the A5 filter and after `normalizeChunkTextFallback`**.
- Yields contiguous 0-based indices on all paths (`chunks`/`json`/`markdown`/`text`/`html` × `full_text`/`embedding`). `chunk_order_int` is used only for retrieval ordering (`retrieval.go:619`), so universal setting has no downside.

## Why this is a deviation, not parity
- **Set difference (A5):** Go drops `text ∧ content_with_weight` empty blocks; Python keeps every non-`None` block.
- **Order difference (A6):** Go sets `chunk_order_int` on all chunks; Python only for `chunks + full_text`. Because Go's set is smaller and set unconditionally, the values differ from Python across the board.
- The Python behavior is suspected to be a bug and warrants a **separate Python issue**; this PR does not touch Python.

## Test plan
- Expanded `TestChunksFromTokenizerUpstream_FiltersPhantomChunks`: `image`/`summary`/`question`-only and empty blocks dropped; `content_with_weight`-only kept (6 → 2).
- Added `TestChunkOrderInt_AllPathsUnconditional` (7 `output_format` × `search_method` combos, each chunk `chunk_order_int == index`).
- Added `TestChunkOrderInt_A5FilterKeepsSequenceContiguous` (an empty/orphan chunk dropped → survivors keep contiguous indices).
- Removed `TestIsPhantomChunk` (function deleted/inlined).
- Full-repo `go test ./...` green (the only failure, `internal/tokenizer` BPE cache digest mismatch, is a pre-existing environment issue unrelated to this change).

## Risk / note
This is a **semantic (breaking) change** for already-indexed KBs: re-indexing will drop previously-indexed `image`/`summary`/`question`-only chunks. This is the intended user-value improvement, but downstream consumers relying on those blocks should be aware.

### DIFF
--- a/internal/ingestion/component/tokenizer.go
+++ b/internal/ingestion/component/tokenizer.go
@@ -323,10 +323,11 @@ func (c *TokenizerComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map
 
 	normalizeChunkTextFallback(chunks)
 
-	// chunk_order_int = 该块在（过滤后）阅读序列中的序号。无条件为每块设置，
-	// 使所有召回块都带稳定阅读顺序索引（用户价值：每条路径都能按阅读序排序，
-	// 而非仅 chunks+full_text）。偏离 Python（tokenizer.py:146 仅在 full_text 块内设）；
-	// 因在过滤后枚举，序号在 Go 内连续。
+	// chunk_order_int is the position of the chunk in the (post-filter) reading
+	// sequence. It is set unconditionally on every surviving chunk so that all
+	// retrievable chunks carry a stable reading-order index on every path, not
+	// just chunks+full_text. Because it enumerates the slice after filtering,
+	// the values are contiguous and 0-based within Go.
 	for i := range chunks {
 		chunks[i].ChunkOrderInt = intPtr(i)
 	}
@@ -558,11 +559,12 @@ func chunksFromTokenizerUpstream(in schema.TokenizerFromUpstream) []schema.Chunk
 	default:
 		raw = cloneChunkDocs(in.JSONResult)
 	}
-	// 仅保留能返回检索内容的块：text 与 content_with_weight 皆空才丢——
-	// 不再把 image/summary/questions 当救命字段（它们不产生可返回内容）。
-	// 含 ContentWithWeight 以覆盖 Parser 路径（其块带 content_with_weight 不带 text），
-	// 避免 normalizeChunkTextFallback 回填前误杀合法块。
-	// 偏离 Python（tokenizer.py:131 只过滤 None）——用户价值优先。
+	// Keep only chunks that have retrievable content: a chunk is dropped only
+	// when both text and content_with_weight are empty. The ContentWithWeight
+	// guard preserves the Parser path, whose blocks carry content_with_weight
+	// without text; normalizeChunkTextFallback backfills text afterwards, so
+	// this guard is required to avoid dropping legitimate Parser blocks before
+	// the backfill runs.
 	filtered := raw[:0]
 	for _, ck := range raw {
 		if ck.Text == "" && ck.ContentWithWeight == "" {
--- a/internal/ingestion/component/tokenizer.go
+++ b/internal/ingestion/component/tokenizer.go
@@ -323,13 +323,12 @@ func (c *TokenizerComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map
 
 	normalizeChunkTextFallback(chunks)
 
-	// Diff Tokenizer-8: chunk_order_int must be set on all paths
-	// (Python sets it unconditionally). tokenizeChunks also sets it
-	// for the full_text path; this covers the embedding-only path.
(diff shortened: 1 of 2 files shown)


internal/ingestion/component/tokenizer.go
```diff
@@ -323,13 +323,12 @@ func (c *TokenizerComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map
 
 	normalizeChunkTextFallback(chunks)
 
-	// Diff Tokenizer-8: chunk_order_int must be set on all paths
-	// (Python sets it unconditionally). tokenizeChunks also sets it
-	// for the full_text path; this covers the embedding-only path.
+	// chunk_order_int = 该块在（过滤后）阅读序列中的序号。无条件为每块设置，
+	// 使所有召回块都带稳定阅读顺序索引（用户价值：每条路径都能按阅读序排序，
+	// 而非仅 chunks+full_text）。偏离 Python（tokenizer.py:146 仅在 full_text 块内设）；
+	// 因在过滤后枚举，序号在 Go 内连续。
```
yuzhichang: English comments only.

xugangqiang:
## Behavior Change Note

This PR intentionally changes two behaviors in `internal/ingestion/component/tokenizer.go`. Posting a clear note because the change is **forward-only** and does **not** require rebuilding existing KBs.

### What changed
- **A5 (filtering):** a chunk is now dropped only when both `text` and `content_with_weight` are empty (replacing the old 4-field `isPhantomChunk` check on `Text && Image && ContentWithWeight && Summary`).
- **A6 (ordering):** after the A5 filter and `normalizeChunkTextFallback`, `chunk_order_int` is set unconditionally to a contiguous 0-based index on every surviving chunk (previously set in two places; consolidated into one post-filter loop, covering all paths).

### Scope — no KB rebuild required
- This change is **forward-only**: it only affects documents indexed **after** this lands.
- Existing chunks in already-indexed KBs are **not deleted or modified**, so **no KB rebuild is required**.
- Re-indexing historical docs is *optional* (only if you want the new filtering applied to old data).

### Which chunks are dropped, and why
A chunk is dropped only when `text` and `content_with_weight` are both empty. Consequently these three kinds (whose only content lives in `image` / `questions` / `summary`) are now dropped:
- **image-only:** retrieval returns only `content_with_weight` (`retrieval.go:361`), so an image-only chunk has no retrievable body; and in the standard QA path images are not embedded (`embedChunks` skips empty-text chunks), so such a chunk was never recallable anyway.
- **question-only / summary-only:** when recalled they return empty `content_with_weight` — negative value.
- **Kept:** any chunk with non-empty `text` or `content_with_weight` (including Parser-path `content_with_weight`-only chunks, which `normalizeChunkTextFallback` backfills into `text`).

### User-visible impact
- **QA retrieval quality is unaffected** — dropped chunks were not recallable / returned empty answers.
- **Only visible difference:** in the chunk-browse / sort-listing view (`Question` empty + `Sort`), these three kinds of chunks no longer appear, and their images are no longer shown.
- **Multimodal path:** if image embedding is enabled in the future, image-only chunks would no longer be semantically recalled; the current Go pipeline does not embed images, so no impact today.

### Deviation from Python (intentional)
- Go drops image/questions/summary-only chunks (Python only filters `None`); Go sets `chunk_order_int` on all paths (Python only chunks+full_text).
- This is a deliberate product decision, tracked under Go-side issue #17916. The Python behavior is suspected to be a bug; a separate Python issue is recommended.

### Tests
- `TestChunksFromTokenizerUpstream_FiltersPhantomChunks`: image/summary/question/empty dropped, content_with_weight-only kept (6→2).
- `TestChunkOrderInt_AllPathsUnconditional`: 7 `output_format` × `search_method` combos, `chunk_order_int == index`.
- `TestChunkOrderInt_A5FilterKeepsSequenceContiguous`: indices stay contiguous after filtering.
- Removed `TestIsPhantomChunk` (function inlined/deleted).

PR #17971, opened 2026-08-07
fix(markdown): open non-PDF document references in internal viewer

dcc123456:
### Summary

fix(markdown): open non-PDF document references in internal viewer

Route non-PDF document reference clicks through /document/:id?ext=... instead of window.open(documentUrl). Pass fileExtension through the click handler in place of the isPdf boolean so the document viewer receives the file type.

Also switch txt-preview from h-full to h-dvh with overflow-auto for proper dynamic viewport sizing inside the document viewer.

### DIFF
--- a/web/src/components/document-preview/txt-preview.tsx
+++ b/web/src/components/document-preview/txt-preview.tsx
@@ -40,7 +40,7 @@ export const TxtPreviewer = ({ className, url }: TxtPreviewerProps) => {
   return (
     <div
       className={classNames(
-        'relative w-full h-full p-4 bg-background-paper border border-border-normal rounded-md',
+        'relative w-full h-dvh p-4 bg-background-paper border border-border-normal rounded-md overflow-auto',
         className,
       )}
     >
--- a/web/src/components/floating-chat-widget-markdown.tsx
+++ b/web/src/components/floating-chat-widget-markdown.tsx
@@ -86,13 +86,14 @@ const FloatingChatWidgetMarkdown = ({
     (
       documentId: string,
       chunk: IReferenceChunk,
-      isPdf: boolean,
+      fileExtension: string,
       documentUrl?: string,
     ) =>
       () => {
         if (!documentId) return;
-        if (!isPdf && documentUrl) {
-          window.open(documentUrl, '_blank');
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
Thanks for the PR! The approach of routing non-PDF documents through the internal viewer makes sense for a more consistent UX.

A few observations:

**1. Unused `useState` import in `markdown-content/index.tsx`**

The diff adds `useState` to the React imports but it is never used in the file. This will trigger a lint warning. Please remove it unless there is a follow-up change planned that needs it.

```diff
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
```

**2. Awkward template literal in the URL construction**

In all three files, the URL is constructed as:

```js
`/document/${documentId}?ext=${fileExtension}&resource=${'document'}`
```

The `resource=${'document'}` is a template literal wrapping a static string — just `resource=document` is simpler and more readable:

```js
`/document/${documentId}?ext=${fileExtension}&resource=document`
```

**3. Edge case: `fileExtension` could be undefined or empty**

The old code passed `fileExtension === 'pdf'` (a boolean) at the call site, so `isPdf` was always well-defined. The new code passes `fileExtension` directly, and the comparison `fileExtension !== 'pdf'` evaluates to `true` when `fileExtension` is `undefined`, `null`, or `""`. This would construct URLs like `/document/xxx?ext=undefined&resource=document` or `/document/xxx?ext=&resource=document`. Consider normalizing `fileExtension` to a safe default (e.g., `""` or falling back to the raw `documentUrl`) when the extension cannot be determined.

**4. `h-dvh` browser support**

The switch from `h-full` to `h-dvh` in `txt-preview.tsx` is a reasonable fix for dynamic viewport sizing (especially on mobile Safari), but note that `dvh` units have relatively recent browser support (Chrome 108+, Safari 15.4+, Firefox 141+). If older browser support is needed, a fallback like `h-full` with a `@supports` guard could be worth considering. This is not blocking — just something to be aware of.

Apart from these, the change looks good and achieves the stated goal of routing non-PDF document clicks through the internal viewer.

PR #17974, opened 2026-08-07
fix(i18n): clarify selectAll label selects current page only

dcc123456:
### Summary

fix(i18n): clarify selectAll label selects current page only

Rename "Select all" / "选择所有" to "Select page" / "选择当前页" in the chunk operations to accurately reflect that the action selects items on the current page, not across all pages.


wangq8 (commented):
Good UX improvement. The label "Select all" was misleading since the action only selects items on the current page, not across all pages. The new labels ("Select page" / "选择当前页") accurately describe the behavior.

The translation key (`selectAll`) remains unchanged, so no downstream breakage. 

LGTM. 👍

Issue #17888, opened 2026-08-05
[Frontend] Merge Token Size & Delimiter chunker tabs into 'Token - Delimiter' (follow-up to #17808)


xugangqiang:
## Background

PR #17808 (merged) refactored the backend chunking logic so that **Token-size** and **Delimiter** based chunking are unified into a single strategy (contract in #17799):

- The delimiter is now always a **hard chunk boundary**. Text is split at every delimiter (line breaks + custom delimiters), and the delimiter text never leaks into a chunk.
- Adjacent pieces are greedily merged up to the recommended chunk size (**OVER_CAP** default). A single piece that already exceeds the size limit is kept **whole and never atom-split** (the old atom-split path was removed).
- **"One"** keeps the entire parsed input as a single chunk.

The backend no longer distinguishes a standalone "Token Size" mode from a "Delimiter" mode — they are one strategy now.

## Current frontend (`main`)

The Agent canvas **TokenChunker** node configuration
(`web/src/pages/agent/form/token-chunker-form/index.tsx`) still exposes **3 tabs**:

1. `token_size` — "Token Size"
2. `delimiter` — "Delimiter"
3. `one` — "One"

This no longer matches the backend contract: the two token/delimiter modes are conceptually one strategy.

## Required change

Reduce the control to **2 tabs**:

- **a. "Token - Delimiter"** (`token_delimiter`) — combines the previous Token Size + Delimiter behavior.
- **b. "One"** (`one`) — unchanged.

Update the segmented control label/tooltip and the i18n strings
(`flow.tokenDelimiter`, `flow.tokenDelimiterTip`, `flow.one`, `flow.oneChunkTitle`,
`flow.oneChunkDescription`) in `web/src/locales/{en,zh}.ts`.

### "Token - Delimiter" option tooltip

**EN:**
> Text is first split at every delimiter (line breaks and any custom delimiters); delimiters are always treated as hard chunk boundaries and never appear inside a chunk. Adjacent pieces are then greedily merged up to the recommended chunk size; a piece that already exceeds the size limit is kept whole and is never split. Choose "One" to keep the entire input as a single chunk.

**ZH:**
> 文本会先按每个分隔符（换行符及自定义分隔符）切分；分隔符始终作为硬性分块边界，不会出现在块内。相邻片段随后会被贪心合并，直到加入下一片会超过建议块大小为止，此时开始新块；单个片段若已超过该大小则保持完整、不再拆分。选择“One”可将整个输入作为单个块。

[...]

PR #17975, opened 2026-08-07
Fix: Move the content of the "Token Size" tab from the token chunker operator to the "Delimiter" tab.  #17888

cike8899:
### Summary

Fix: Move the content of the "Token Size" tab from the token chunker operator to the "Delimiter" tab.  #17888

### DIFF
--- a/web/src/pages/agent/constant/pipeline.tsx
+++ b/web/src/pages/agent/constant/pipeline.tsx
@@ -262,7 +262,7 @@ export const initialTokenChunkerValues = {
   outputs: {
     chunks: { type: 'Array<Object>', value: [] },
   },
-  delimiter_mode: 'token_size',
+  delimiter_mode: 'delimiter',
   chunk_token_size: 512,
   overlapped_percent: 0,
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Well-executed refactor. The migration from three tabs (Token Size / Delimiter / One) to two tabs (Delimiter / One) is handled cleanly:

- **Backward compatibility**: Legacy `token_size` nodes are normalized to `delimiter` in both the form default values and `transformTokenChunkerConfigToForm`, so existing saved pipelines load correctly.
- **Empty delimiters edge case**: Seeding `[{ value: "\n" }]` when the saved array is empty prevents a broken UI state for nodes that were previously saved under the `token_size` tab with no delimiters.
- **Form logic**: Moving `chunk_token_size`, `overlapped_percent`, and `image_context_window` fields into the `delimiter` section makes sense since these settings apply when delimiter-based chunking is active.

One thing to double-check: the `isEmpty` import from lodash is new — make sure lodash is already in the project dependencies (it should be, but worth a quick verify).

LGTM. 👍

PR #17977, opened 2026-08-07
fix: reject duplicate chat parameter keys

buua436:

Reject chat configurations where `prompt_config.parameters` contains duplicate keys.

### Type of change


wangq8 (commented):
Good validation addition. The duplicate key check is correctly applied to all three endpoints (create, update, patch_chat), and a test case is included.

One minor edge case: if a parameter dict does not have a `key` field, `parameter.get("key")` returns `None`. Two consecutive keyless parameters would then trigger `"duplicate key: None"`. This is unlikely in practice since keyless parameters are malformed input, but you could consider filtering with `if key is None: continue` for robustness.

Otherwise LGTM. 👍

PR #17978, opened 2026-08-07
Fix: verification message

Lynn-Inf:
### Summary

Display clear error message when no model passed verification.


wangq8 (commented):
Clean fix. When no model passes verification, `msg` remains empty and the API response would show `"message": ""` which is confusing. Adding the fallback `"No model passed verification"` gives the user a clear error message.

LGTM. 👍

PR #17979, opened 2026-08-07
refactor(chunker): converge delimiter_mode to {delimiter, one}, drop token_size

xugangqiang:
## Summary

Converge `TokenChunker.delimiter_mode` from three values (`token_size`, `delimiter`, `one`) to two (`delimiter`, `one`). The unified `delimiter` mode now carries the old `token_size` semantics: when no active (backtick) delimiter is present, text/JSON chunks are merged up to `chunk_token_size`; when a backtick delimiter is present, the text is split by it and not merged. `one` continues to be handled by the separate `OneChunker`.

This is a contract-only change on the Go side (runtime dispatch was already delimiter-driven). On the Python side, the JSON path is **fixed** so that the no-delimiter case merges per-item chunks via `_merge_text_chunks_by_token_size` (matching the old `token_size` mode and the Go JSON path). The initial submission concatenated every item into a single chunk before merging, which defeated `chunk_token_size` and diverged from Go — this is corrected in follow-up commits. Legacy `delimiter_mode: "token_size"` is still accepted by `check()` and normalized to `"delimiter"` (behaviorally identical), so existing configs and pre-fix frontends keep working.

### Go (`internal/ingestion/component/...`)
- `schema/chunker.go`: `Validate()` whitelist now accepts only `"delimiter"` (rejects `"token_size"`; `"one"` is handled by `OneChunker`); default changed to `"delimiter"`.
- `chunker/token.go`: updated the whitelist comment. No runtime behavior change — dispatch already uses `hasActiveDelimiter` / `hasCustomDelim`.
- Updated builtin DSL templates (`internal/ingestion/pipeline/template/*.json`) from `"token_size"` to `"delimiter"`.
- Synced Go tests (`schema_test`, `token_test`, `token_tag_test`, `token_strict_cap_test`, `pipeline_executor_defaults_test`).

### Python (`rag/flow/chunker/token_chunker.py`)
- Default `"delimiter"`; `check()` whitelist accepts `["delimiter", "one"]` and normalizes legacy `"token_size"` to `"delimiter"`.
- JSON path is now two distinct branches:
  - **No active delimiter pattern**: per-item chunks are merged to `chunk_token_size` via `_merge_text_chunks_by_token_size` (the old `token_size` behavior, and parity with Go). Media chunks are preserved and break the merge.
  - **Active backtick delimiter pattern**: the buffered text stream is split by the pattern with per-segment PDF-position preservation (the position-aware `flush_text_buffer` logic). No merge, matching Go.
  - Previously these were a single concat-then-merge path that collapsed everything into one chunk and ignored the cap; the no-delimiter case now merges correctly per item.
- text/markdown/html path merges via `naive_merge` when no delimiter pattern is active, and splits by the pattern otherwise.
- Removed the separate `token_size` JSON branch and unified the no-delimiter merge onto the `delimiter` path.
- Updated Python agent templates (`agent/templates/*.json`) from `"token_size"` to `"delimiter"`.
- Added regression tests: JSON no-delimiter merge-to-cap, media-bounded merge, and `token_size`→`delimiter` normalization.

## Test plan
- Go: `go test ./internal/ingestion/component/chunker/... ./internal/ingestion/component/schema/... ./internal/ingestion/task/...` — all pass.
- Python: `pytest rag/flow/tests/test_token_chunker.py` — 11/11 pass.
- `ruff check` / `ruff format` clean; pre-commit hooks (gofmt, check-json, ruff) pass.

## Notes
- With a backtick custom delimiter present, both Go and Python only split and do **not** merge to cap — this is already aligned and unchanged by this PR.
- Frontend (`web/src`) still defaults `delimiter_mode` to `token_size` and is handled separately by #17975. The Python `check()` already normalizes `token_size`→`delimiter`, so the backend does not reject it; the Go `Validate()` still rejects `token_size`, but the Go ingestion pipeline is not yet deployed, so this is deferred until it goes live.


yuzhichang:
LGTM

PR #17981, opened 2026-08-07
Fix: Clicking the confirm button on the WikiCommitModal does not validate the form.

cike8899:
### Summary
Fix: Clicking the confirm button on the WikiCommitModal does not validate the form.

### DIFF
--- a/web/src/pages/dataset/compilation/hooks/use-commit-artifact.ts
+++ b/web/src/pages/dataset/compilation/hooks/use-commit-artifact.ts
@@ -44,10 +44,6 @@ export function useCommitArtifact({
     setIsOpen(true);
   }, [form]);
 
-  const close = useCallback(() => {
-    setIsOpen(false);
-  }, []);
-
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
The fix correctly addresses the issue. Here is why:

The Modal component `handleOk` calls `onOpenChange(true)` before calling `onOk()`. The old `close` callback unconditionally called `setIsOpen(false)`, which meant the modal closed *before* `form.handleSubmit(onConfirm)()` had a chance to run validation. By exposing `setIsOpen` directly, the modal stays open during validation, and `handleConfirm` calls `setIsOpen(false)` only on success.

One minor observation: exposing the raw `setIsOpen` setter from the hook slightly reduces encapsulation compared to the dedicated `close`/`open` callbacks, but it is a reasonable trade-off here since the Modal component needs to call `onOpenChange(true)` during the OK flow.

LGTM. 👍

PR #17982, opened 2026-08-07
Fix: The height of the source code mode in the wiki template's Markdown editor is too small.

cike8899:
### Summary

Fix: The height of the source code mode in the wiki template's Markdown editor is too small.


wangq8 (commented):
LGTM. Simple and correct fix — adding `valueClassName="flex-1 min-h-0"` ensures the Markdown editor takes the available height in the flex container when switched to source code mode. Thanks for catching this.

PR #17983, opened 2026-08-07
Fix: handle unpaired <> in messages

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Nice fix. The approach — escape standalone `<`/`>` before DOMPurify, then unescape — is clean and applied consistently across all 3 markdown content components.

A couple of observations:

1. The regex `<[^>]*>` correctly distinguishes matched `<...>` pairs from orphaned brackets. One edge case: `<[^>]*>` would treat `< 10>` (a number literal followed by `>` as closing bracket) as a matched tag and let DOMPurify strip it rather than escaping it. This is a rare case in real chat content though, so probably acceptable.

2. The `provider_api_service.py` change (`msg = msg or "No model passed verification"`) adds a helpful fallback error message, though it is unrelated to the main angle-bracket fix — minor but fine as a fly-by cleanup.

3. Would be nice to have a couple of unit tests for `escapeUnmatchedAngleBrackets` / `unescapeAngleBrackets` covering the common cases (standalone `<`, `>`, mixed with real tags, edge cases). Worth a follow-up.

Overall LGTM — the double-escaping strategy is the right call given how aggressive DOMPurify is.

cike8899 (approved):

PR #17984, opened 2026-08-07
feat: add agent filter counts

buua436:

Add backend-provided owner and canvas-category counts for the Agent list filters.

The counts include regular Agents and valid compilation template groups, avoiding incorrect counts caused by frontend pagination.

### Type of change

### DIFF
--- a/api/apps/restful_apis/agent_api.py
+++ b/api/apps/restful_apis/agent_api.py
@@ -719,6 +719,7 @@ def list_agents(tenant_id):
         effective_owner_ids = list(requested_owner_ids)
     else:
         effective_owner_ids = list(authorized_owner_ids)
+    include_template_groups = tenant_id in effective_owner_ids
 
     # Groups-only: when ``compilation_template_group`` is the only selected
     # category, return just the caller's template groups (no agents) via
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Nice feature — server-side filter counts are definitely the right approach to get accurate numbers independent of frontend pagination.

**Backend:**

- `get_owner_filter` and `get_category_filter` both properly respect the TEAM permission boundary and correctly fold in `CompilationTemplateGroup` counts.
- The `filter` API branch in `list_agents` follows the existing auth pattern (`get_joined_tenants_by_user_id` + `authorized_owner_ids`).
- One nit: `get_category_filter` groups purely by `canvas_category` and returns whatever values happen to be in the DB. The frontend then looks up counts for exactly 3 known category IDs (`dataflow_canvas`, `compilation_template_group`, `agent_canvas`). If a future migration adds a new category value to the DB, the frontend filter bar would silently ignore it. Consider whether the frontend should dynamically build the category list from the backend response instead of hardcoding the 3 IDs.

**Frontend:**

- `useFetchAgentFilters` is clean and follows the existing hook pattern.
- Replacing the client-side `buildOwnersFilter` logic with server-provided data simplifies `use-select-filters` nicely.
- Ensure `gcTime: 0` is intentional — this means the filter counts are refetched on every mount. For a filter sidebar that changes frequently (agents created/deleted), this is probably correct, but worth confirming it is not unnecessarily chatty.

Overall LGTM, solid feature.

PR #17986, opened 2026-08-07
fix(model): simplify local LLM config and align model type values

dcc123456:
### Summary

fix(model): simplify local LLM config and align model type values

- Drop the `modelTypes` and `modelNameLabel` parameters from `buildLocalConfig` and all local LLM provider callers; these args are no longer needed by the field config.
- Align model type option values with backend naming: `image2text` -> `vision`, `speech2text` -> `asr` (i18n labels unchanged).
- Flatten the `useCustomModelFields` useMemo callback for readability.

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/use-custom-model-fields.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/use-custom-model-fields.tsx
@@ -46,9 +46,9 @@ export const MODEL_FIELD_SCHEMA: AddCustomModelDialogFields[] = [
       { value: 'embedding', label: 'modelTypes.embedding' },
       { value: 'rerank', label: 'modelTypes.rerank' },
       { value: 'tts', label: 'modelTypes.tts' },
-      { value: 'image2text', label: 'modelTypes.image2text' },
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Three clean changes:

1. **`modelTypes`/`modelNameLabel` removal** — These params are already dead code in `buildLocalConfig` (their consuming fields are commented out). Removing them from the signature and all callers is the right cleanup.

2. **Model type value alignment** — `image2text` → `vision` and `speech2text` → `asr` matches the backend canonical values from `common/constants.py` (`LLMType.VISION = vision`, `LLMType.ASR = asr`). The backend already has a compatibility mapping (`image2text: vision`, `speech2text: asr`) in `db_models.py` and `model_utils.py`, so this change is backward-compatible and aligns frontend values with the canonical backend enum.

3. **useMemo formatting** — Flattening the callback from inline arrow to block body is readability-only, no behavioral change.

One thing to verify: are there any other frontend references to the old `image2text` / `speech2text` values (e.g. in other configs, constants, or API calls) that should also be updated for consistency? The `MODEL_FIELD_SCHEMA` options you changed are the primary place, but a quick grep for the old strings in `web/` might be worth it.

PR #17987, opened 2026-08-07
Fix: Links to wiki content accessed via the version view are not clickable.

cike8899:
### Summary

Fix: Links to wiki content accessed via the version view are not clickable.


wangq8 (commented):
LGTM. Clean fix.

The early return on `isVersionView` was blocking all markdown link navigation when viewing wiki content in version view mode. Since the handler already guards against navigating to the same entry (`currentEntry?.slug === slug`), the version-view guard was overly restrictive and not needed. Removing `isVersionView` from both the guard and the dependency array is the correct fix.

PR #17989, opened 2026-08-07
fix(prompt): cite document source_url instead of internal filename

alex-makang:

`kb_prompt()` builds each retrieved chunk's prompt block as:

```
ID: 0
├── Title: rtc-error-codes__web.md
├── URL:
└── Content: ...
```

`ck["url"]` is only populated for web-search results, so for every chunk that
came from an uploaded document the URL node is empty. The only link-shaped value
left in the block is the Title, which is `docnm_kwd` - an internal filename.
Models then cite that filename as if it were a public URL, so answers come back
with dead links.

Documents already carry their real public address in `source_url` metadata, but
it only reached the prompt when the caller opted into full reference-metadata
enrichment (`include_reference_metadata`), which is off by default.

### Type of change

### Changes

- `rag/prompts/generator.py`: fall back to `document_metadata["source_url"]`
  when the chunk has no `url`, and skip `source_url` in the generic metadata
  loop now that it is emitted as the URL node.
- `api/db/services/dialog_service.py`: enrich chunks with `source_url`
  unconditionally in both `async_chat` and `async_ask`. Richer reference
  metadata stays opt-in - only this one field is added when the caller did not
  ask for more.

Behaviour is unchanged for chunks that already carry a `url`, and for callers
that already enable `include_reference_metadata`.

### Test plan

- [x] Chunks with a `url` still emit that `url`.
- [x] Chunks with no `url` but a `source_url` now emit the `source_url`.
- [x] `source_url` is no longer duplicated as both URL and a metadata node.
- [x] Verified in a running deployment: answers over a documentation dataset
      cite the real public docs URLs instead of `*__web.md` filenames.


JinHai-CN:
@alex-makang would you please fix the comments from [coderabbitai](https://github.com/apps/coderabbitai)?
@Attili-sys Would you please help review this PR?

buua436 (commented):
The `source_url` enrichment is still skipped for the SQL retrieval path in `async_chat`: when `use_sql()` returns an answer, `_enrich_chunks_with_document_metadata()` is only called if `include_reference_metadata` is enabled. Therefore, with the default configuration, SQL-based results can still lack `document_metadata.source_url` and the model may continue citing internal filenames. Please apply the same `source_url` enrichment to the SQL branch before yielding the answer.

PR #17990, opened 2026-08-07
fix(openai-compat): keep caller system message content

alex-makang:

The OpenAI-compatible `/chat/completions` endpoint discards every `system`
message in the request:

```python
for message in messages:
    if message["role"] == "system":
        continue
```

RAGFlow builds its own system prompt from the chat assistant's configuration, so
a caller's system message genuinely cannot be forwarded as-is. But dropping it
outright also loses per-request context that OpenAI clients legitimately put
there - session facts, the end user's locale, an account identifier - with no
error and no warning. Any client written against the OpenAI API will put that
information in a `system` message and silently never have it reach the model.

### Type of change

### Changes

Collect the caller's system text and prepend it to the first user turn, so the
model still sees it while the assistant's configured system prompt stays
authoritative. The message dicts are replaced rather than mutated, since they
are still owned by the request payload.

Requests without a system message are unaffected.

Happy to make this opt-in behind a request flag instead if maintainers prefer
the current silent-drop as the default.

### Test plan

- [x] Request with no system message: `msg` is byte-identical to before.
- [x] Request with one system message: its text is prepended to the first user
      turn; the assistant's configured system prompt still applies.
- [x] Request with several system messages: all are preserved, in order.
- [x] The caller's original `messages` list is not mutated.


RerankerGuo:
Independent validation at head `41dd0af42961b59458adf3f238db2eaacc3b8d25`.

Environment: macOS, Python 3.13.14, pytest 9.0.2.

I ran the endpoint function from an exact commit snapshot with a standalone dependency-stubbed harness covering:

- two caller `system` messages preserved in order and prepended to the first user turn;
- a leading assistant message still omitted;
- later assistant/user turns retained;
- the original request message list left unchanged;
- a request without a system message forwarded unchanged.

Results with the same two tests:

```text
origin/main@8e37f554: 1 failed, 1 passed
  preservation case failed because main forwarded "first" without system content

PR head@41dd0af4: 2 passed
```

Conclusion: the patch fixes the demonstrated silent-drop behavior while preserving the no-system path and request ownership. This was a stubbed route-level validation; it did not call a live model provider.

wangq8 (commented):
Reviewed — clean and safe fix.

I confirmed `_normalize_openai_messages` (via `_normalize_message_content`) normalizes every message `content` to a string before this loop runs, so the string concatenation in the prepend can't hit a list/multimodal content part. The `messages[-1]["role"] == "user"` validation also guarantees a user turn exists to fold into, and the replace-instead-of-mutate approach leaves the request payload (and the `prompt` derived from it) untouched.

This is a reasonable default. If maintainers want to preserve the old silent-drop for some callers, an opt-in flag could be added later, but I don't think it's needed.

LGTM.

PR #17992, opened 2026-08-07
Fix: Hide the "Add Link" button in the Markdown editor.

cike8899:
### Summary

Fix: Hide the "Add Link" button in the Markdown editor.


wangq8 (commented):
LGTM. Straightforward fix — the link button in the Markdown floating toolbar shouldn't be exposed while link functionality isn't ready or desired.

Minor style note: consider removing the code instead of commenting it out. If this is a temporary feature gate, a feature flag or conditional render would be cleaner than comment blocks.

PR #17993, opened 2026-08-07
Append attachments content to last message

wangq8:


wangq8 (commented):
## PR Review: Append attachments content to last message

Thanks for this PR! The direction is good — moving attachment content from system prompts and synthetic chunks into the actual user message is cleaner and semantically correct. Here are my findings:

---

### ✅ What Looks Good

1. **Removing synthetic `kbinfos` for text attachments** — The old approach of injecting fake `"Chat attachment"` chunks into `kbinfos` was a hack. Removing it is the right call.
2. **ABSTAIN cleanup** — Removing the special-case return path that leaked `kbinfos` when attachments were present but the agent abstained. Now both orchestrators (`agentic.py`, `decompose.py`) consistently clear chunks and return `abstain: True`.
3. **Attachment placement** — Putting text attachments in the last user message instead of the system prompt is semantically more appropriate.

---

### ⚠️ Issues

**1. Missing delimiter between user question and attachment content**

In all three append sites, the attachment text is concatenated directly with no separator:

```python
msg[-1]["content"] += text_attachments_content
```

This produces output like `"What is attached?attached facts"` (confirmed in the test expectation). The LLM will see this as a single run-on string, which degrades comprehension. Consider using a clear delimiter:

```python
msg[-1]["content"] += "\n\n---\n" + text_attachments_content
```

This applies to all three sites:
- `api/db/services/dialog_service.py` ~line 842 (async_chat path)
- `api/db/services/dialog_service.py` ~line 1887 (rag_agent path)
- `rag/advanced_rag/agentic_rag.py` ~line 579 (RAGTools.rag)

If you make this change, the test expectation should also be updated to include the delimiter.

**2. Potential double-injection of attachment content**

The `dialog_service.py:rag_agent` function appends `text_attachments_content` to `agent_messages[-1]`, and `RAGTools.rag()` also appends `self.text_attachments_content` to its own messages. If these are in the same call chain, the attachment content could appear twice. Please verify the call path and ensure only one layer injects, or that the inner layer detects already-appended content.

**3. No PR description**

The PR body is empty. Please add a brief summary explaining the problem being fixed and the expected behavior change — this helps reviewers and future readers of the git history.

**4. Test coverage is minimal**

The test only covers `RAGTools.rag`. No coverage for:
- The `async_chat` path in `dialog_service.py` (line ~842)
- The `rag_agent` path in `dialog_service.py` (line ~1887)
- Mixed text + image attachment scenarios

Consider adding focused tests for these entry points.

---

### Summary

The core change is solid — moving attachment content to user messages and cleaning up abstention paths. The delimiter issue should be fixed before merge; the double-injection concern needs verification. Approving with the delimiter fix as a blocking request.

wangq8 (commented):
## Re-review (post-merge): Double-injection + delimiter regression

This PR is now merged, so consider this a post-merge audit. I traced the full `rag_agent` → `RAGTools.rag()` call chain and found a confirmed double-injection bug.

---

### 🔴 Confirmed: Double-injection in the agentic RAG flow

The `rag_agent` function at line 1888 appends `text_attachments_content` to `agent_messages[-1]["content"]` (step ①). Then `agent_messages` is passed as conversation history to the LLM:

```python
# line ~2023
stream_iter = chat_mdl.async_chat_streamly_delta(
    rag_tools.sys_prompt(), agent_messages, gen_conf
)
```

`RAGTools` is also initialized with `text_attachments_content` (line ~1938). When the LLM calls the `rag` tool, `RAGTools.rag(question)` appends the *same* attachment content again to its internal messages (step ②):

```python
# agentic_rag.py ~578-579
messages = [{"role": "user", "content": question}] if question else []
if self.text_attachments_content and messages:
    messages[-1]["content"] += self.text_attachments_content
```

**Result:** The attachment text appears:
1. In the outer conversation history (`agent_messages`) — the LLM sees it
2. In every inner `rag` sub-query — passed to `run_agentic_rag` for retrieval

This means every retrieval sub-query is bloated with the full attachment text. The LLM calls `rag("what was the revenue?")` but the actual query becomes `"what was the revenue?[full attachment text…]"` — polluting semantic search results. The old synthetic-`kbinfos` approach did not have this problem; it only added attachment as evidence, never injected it into sub-queries.

**Fix:** Only one layer should append. I recommend **removing the append from `RAGTools.rag()`** (agentic_rag.py line ~578-579) and keeping it in `rag_agent`. The LLM already sees attachments via `agent_messages`; there is no reason to re-inject them into every inner search query. This also aligns with how the `async_chat` (non-agentic) path works — it appends once to the user message and does not re-inject anywhere else.

---

### 🟡 Still present: Missing delimiter (from first review)

At all three append sites, the attachment text is concatenated with zero separator:

```python
msg[-1]["content"] += text_attachments_content
```

This produces `"What is attached?attached facts"`. A delimiter like `"\n\n---\n"` is strongly recommended for LLM comprehension.

Affected sites:
- `dialog_service.py` line 842 (async_chat)
- `dialog_service.py` line 1888 (rag_agent)
- `agentic_rag.py` line 579 (RAGTools.rag — assuming it is not removed per above)

---

### 🟢 Still correct

- ABSTAIN path cleanup in both orchestrators
- Removal of synthetic kbinfos
- `get_files_content` returns `""` (not `None`) so the truthiness check is safe
- The `and msg` / `and agent_messages` guards correctly handle empty message arrays

---

### Summary

The double-injection in the agentic RAG flow is a functional regression that should be fixed in a follow-up PR. The delimiter issue is a quality-of-life fix. Both are low-risk changes.

PR #17994, opened 2026-08-07
fix: preserve selected row IDs without filtering by current list

dcc123456:
### Summary

fix: preserve selected row IDs without filtering by current list

### DIFF
--- a/web/src/hooks/logic-hooks/use-row-selection.ts
+++ b/web/src/hooks/logic-hooks/use-row-selection.ts
@@ -42,7 +42,8 @@ export function useSelectedIds<T extends Array<{ id: string }>>(
         .map((x) => x.id);
     } else {
       // ID-based selection (when getRowId is used)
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Simple and correct fix. The root cause is clear: `useSelectedIds` was filtering `selectionKeys` to only IDs that exist in the current `list` array, which discards selected rows whenever the list changes (pagination, filtering, tab switch, etc.).

The fix — returning `selectionKeys` as-is — preserves selection state across list mutations, which is the expected behavior for ID-based row selection.

**One consideration (non-blocking):**
If rows are genuinely deleted from the backend (not just filtered out of the current view), their IDs will remain in the selection set. Bulk operations on these stale IDs should fail gracefully. This is already safer than the current bug (silently dropping selections), and typical CRUD backends handle missing IDs without issues, so this is fine in practice.

**Style nit:** Consider removing the commented-out line rather than keeping dead code, unless there's a specific reason to keep it for reference.

LGTM. 🚀

PR #17995, opened 2026-08-07
fix(ingestion): make document/KB counter application idempotent per run

deadtrickster:
The follow up to the https://github.com/infiniflow/ragflow/pull/17236#issuecomment-5186837112 where I was asked to look at the Go side.

An at-least-once ingestion task redelivery re-runs the pipeline and double-counts the document/knowledgebase `chunk_num`/`token_num`.

`runTask` applies a run's chunk/token counts (`runDocumentTask` -> `docState.apply`) before it marks the task complete, and on success the pipeline deletes its checkpoint, so a redelivered task starts a fresh run. The counter apply was a plain running-sum increment, so the second run added the same counts again. The chunk store stays correct (idempotent upsert by deterministic id) - only the aggregate counters inflate to 2x.

Two triggers, both covered:
- the first delivery completes but its Ack is lost, then the same message is redelivered
- a run applies the counters but dies before `MarkCompleted`, leaving the task RUNNING for redelivery

Fix: rename `IncrementChunkNum` to `ApplyDocCounts` and make it idempotent. Under a row lock it sets the document to the run's absolute counts and rolls only the delta into the knowledgebase aggregate (clamped at zero). Re-applying the same run contributes zero. A re-parse producing a different count carries the delta correctly.

Tests in `internal/ingestion/service/redelivery_counter_test.go` drive `runTask` twice for both triggers and assert the counters apply once. Red without the fix, green with it.

I know the go migration is in progress and you would rather hold ingestion refactorings. This one is a live data-integrity bug with a contained fix, happy to hold or turn it into an issue if you prefer.

### DIFF
--- a/internal/ingestion/service/doc_state.go
+++ b/internal/ingestion/service/doc_state.go
@@ -33,7 +33,7 @@ import (
 type docStateSvc interface {
 	GetDocumentMetadataByID(ctx context.Context, docID string) (map[string]any, error)
 	SetDocumentMetadata(ctx context.Context, docID string, meta map[string]any) error
-	IncrementChunkNum(ctx context.Context, docID, kbID string, chunkNum, tokenNum int, duration float64) error
+	ApplyDocCounts(ctx context.Context, docID, kbID string, chunkNum, tokenNum int, duration float64) error
 }
 
 // docStateUpdater applies a pipeline run's results to document state: it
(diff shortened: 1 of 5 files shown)


xugangqiang:
Thanks for the fix — the idempotent counters look correct. One gap before merge: the new tests cover the two redelivery triggers (same counts applied twice), but a few related behaviors are untested:

1. **Re-parse with different counts** — e.g. a document re-parsed from 5→7 chunks. This is the path where `ApplyDocCounts` carries a non-zero delta into the KB aggregate; please add a test asserting the KB ends at the new absolute value (not 5+7).
2. **KB clamp-at-zero branch** — the `CASE WHEN chunk_num + ? >= 0 THEN ... ELSE 0 END` path is not exercised. A test that drives the aggregate negative (or starts from an inconsistent state) and asserts it bottoms out at 0 would lock in that behavior.
3. **`process_duration` semantics** — note this changed from cumulative (`process_duration + ?`) to absolute (`process_duration = duration`). A test pinning the expected behavior (single run sets it; redelivery does not accumulate) would prevent a future regression.

Could you add coverage for these three cases?

deadtrickster:
@xugangqiang Hi, I believe I've addressed your comments. Thank you for the review

xugangqiang:
LGTM

PR #17996, opened 2026-08-07
Feat: Search for timeline nodes.

cike8899:
### Summary

Feat: Search for timeline nodes.

### DIFF
--- a/web/src/components/structure-graph/adapters.ts
+++ b/web/src/components/structure-graph/adapters.ts
@@ -153,37 +153,6 @@ export function adaptTreeToTreeData(
   return buildTreeDataItems(template.entities, template.relations, ['child']);
 }
 
-function filterTreeDataItems(
-  items: TreeDataItem[],
-  keyword: string,
-): TreeDataItem[] {
-  const lowerKeyword = keyword.toLowerCase();
-
-  return items.reduce<TreeDataItem[]>((acc, item) => {
-    const children = item.children
-      ? filterTreeDataItems(item.children, keyword)
-      : [];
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Good refactor — moving tree search from client-side to server-side is the right approach for scalability. Architecture review:

**What works well:**
- Removing the client-side `filterTreeDataByKeyword` and relying on `useFetchDocumentStructureGraph` with keyword parameters unifies the search path. Clean separation.
- The 500ms debounce via `useDebounce` prevents hammering the API during fast typing.
- The `useSelectedTemplate` fallback fix (checking if `selectedTemplateId` is still in the current templates array before using it, falling back to `templates[0]`) is important — without it, search results that don't include the previously selected template would render a blank representation.
- Moving `searchKeyword` state and its handler into `useGraphEntitySearch` keeps the hook as the single source of truth for all search-related state. Good cohesion.
- Mutual exclusivity between `graphKeywords` (entity dropdown) and `searchKeyword` (ExpandableSearchInput) is handled properly through the callbacks — at most one is non-empty at a time.

- In `use-graph-entity-search.ts`, `handleSearchKeywordChange` only sets `searchKeyword` but doesn't explicitly clear `graphKeywords`. The safety relies on callers (the dropdown handlers) clearing `searchKeyword`. This is fine for the current UX where the two inputs never show simultaneously, but a future refactor that adds both inputs at once could hit a subtle conflict. Consider adding a comment noting this assumption, or clearing `graphKeywords` in `handleSearchKeywordChange` for defense-in-depth.
- The comment cleanup in `skill-tree.ts` is a nice touch — stale references to deleted code are worse than removed code.

LGTM overall. 👍

PR #17998, opened 2026-08-07
Fix answer in think

wangq8:


wangq8 (commented):
## PR Review: Fix answer in think

**Author:** @wangq8 | **Files:** 1 | **+18/-6**

### Summary

This fixes a streaming display issue where a direct answer (no tool call) from the agent model ends up rendered inside the `<think>` block. The core problem is that the pre-tool text buffer (for providers that omit explicit reasoning metadata) has no way to distinguish "thinking before a tool call" from "a direct answer without any tool call".

### What changed

1. **Extract `in_think` at queue time** (line 2030): The `state.in_think` boolean is captured when the stream delta is enqueued rather than passing the full `State` object through the queue. This avoids any risk of the mutable state being read after modification.

2. **Buffer pre-tool text** (`pending_outer_text` list): Instead of immediately yielding text that arrives before the outer tool starts, buffer it until the code knows whether a tool call follows or the stream ends without one.

3. **Decision at signal time:**
   - **`tool_started` arrives** → flush pending text inside the open think block (correct: this is pre-tool thinking).
   - **`stream_done` arrives without a tool** → close the think block first, then flush pending text as the actual answer outside the think block (this is the direct-answer fix).

### Logic trace

**Case A: Tool call present** (same behavior as before)
```
think block opens → in_think text (yielded) → pre-tool text (BUFFERED) →
tool_started → flush buffer inside think → think closes → answer deltas yielded
```

**Case B: Direct answer, no tool** (fixed)
```
think block opens → in_think text (yielded) → answer text (BUFFERED) →
stream_done → think closes → flush buffer as answer outside think
```
Before this PR, case B text was yielded immediately inside the still-open think block, so the answer was hidden inside `<think>`.

### Verdict

The change is clean, minimal, and correctly handles both tool-call and direct-answer paths. The buffer semantics are sound and the think-block open/close ordering is preserved. LGTM.

**Minor suggestion (non-blocking):** A unit test covering the direct-answer stream path would help prevent regression, since this code is part of a complex async generator pipeline and the bug is easy to reintroduce.

PR #18000, opened 2026-08-07
fix: stream agent tts audio by sentence

buua436:

Fix Agent streaming TTS so that thinking content is not synthesized. Text content is sent immediately, while `audio_binary` is generated asynchronously and sent by sentence in the original order.

### Type of change

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -659,41 +659,79 @@ class Canvas(Graph):
                     if isinstance(cpn_obj.output("content"), partial):
                         _m = ""
                         buff_m = ""
+                        in_thinking = False
+                        tts_tasks = []
+                        sentence_end = re.compile(r"(?:[。！？!?；;\n](?:[”’\"』」】》）)]*)|(?<=[.!?])(?=\s|$))")
                         stream = cpn_obj.output("content")()
 
+                        def _schedule_tts(text):
+                            if tts_mdl and text:
+                                tts_tasks.append(asyncio.create_task(asyncio.to_thread(self.tts, tts_mdl, text)))
+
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this PR! The fix for skipping TTS on thinking content and the switch to sentence-level TTS are solid improvements. A few observations:

**What works well:**
- Thinking content is correctly excluded from TTS synthesis
- Async TTS via `asyncio.to_thread` prevents blocking the main stream
- FIFO ordering is preserved via `_drain_ready_tts`

**Minor concerns to consider:**

1. **`buff_m` reset on `<think>` may discard pre-thinking text:** When `<think>` arrives, `buff_m` is cleared. If the model outputs partial text before `<think>` that doesn't yet form a complete sentence, that text is lost and never TTS-synthesized. In practice this is rare (models typically think first), but worth being aware of. A safer approach might flush remaining `buff_m` through TTS before clearing it on `<think>`.

2. **Regex robustness:** The `(?<=[.!?])(?=\s|$)` lookbehind pattern could match decimal points (e.g. `3.14`) or abbreviations (`Mr.`, `U.S.`). This is a pre-existing challenge with sentence splitting and not a blocker, just something to keep an eye on.

3. **Potential task leak on cancellation:** If the async generator is cancelled mid-stream, the `asyncio.create_task` tasks in `tts_tasks` won't be explicitly cleaned up. Consider wrapping in a `try/finally` to cancel pending tasks.

Overall LGTM, these are minor observations rather than blocking issues. The sentence-level approach is a definite improvement over the fixed 16-char buffer.

PR #18004, opened 2026-08-07
fix: improve streaming agent tts

buua436:

Improve Agent streaming TTS by buffering content with a bounded queue, synthesizing sentences asynchronously, preserving audio order, handling TTS errors, and flushing pending text before thinking content.

### Type of change

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -660,22 +660,49 @@ class Canvas(Graph):
                         _m = ""
                         buff_m = ""
                         in_thinking = False
-                        tts_tasks = []
+                        tts_queue = asyncio.Queue(maxsize=4)
+                        tts_results = asyncio.Queue()
+                        tts_workers = []
+                        tts_sequence = 0
+                        next_audio_sequence = 0
+                        completed_tts = {}
                         sentence_end = re.compile(r"(?:[。！？!?；;\n](?:[”’\"』」】》）)]*)|(?<=[.!?])(?=\s|$))")
                         stream = cpn_obj.output("content")()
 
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this PR! The improvements to the streaming agent TTS pipeline are well-structured. Here is my review:

## ✅ What works well

1. **Fixes lost TTS text before `<think>`** — the old code cleared `buff_m` on `<think>` without sending it to TTS. Flushing it first is the right fix.

2. **Audio order preservation** — using sequence numbers + `completed_tts` dict with ordered drain is clean and correct. Prevents out-of-order playback when TTS calls complete at different speeds.

3. **Bounded concurrency** — replacing unbounded `asyncio.create_task` with a `Queue(maxsize=4)` and 2 workers prevents resource exhaustion. Sensible defaults.

4. **Error handling** — logging TTS failures as warnings instead of silently ignoring them is a big improvement for debugging.

5. **Clean shutdown** — the `finally` block cancels workers and awaits them with `return_exceptions=True`. Proper cleanup.

6. **`_schedule_tts` is now properly `await`ed** — the old code was fire-and-forget without awaiting; the new code correctly awaits the queue put.

## 🔍 Minor observations (non-blocking)

- The 2-worker pool and queue size of 4 are reasonable defaults. Might consider making these configurable if TTS latency becomes a bottleneck for some deployments, but fine for now.
- `_stream_events` async generator encapsulation is clean and makes the control flow easier to follow.

## Verdict

LGTM 👍. No regressions or blocking issues.

PR #18006, opened 2026-08-07
Fix: exclude unauthorized memory/dataset from list results

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for this PR. The shift from "fail on any unauthorized resource" to "silently filter out unauthorized resources with a warning log" is a sensible UX improvement.

**What I checked:**
- `_filter_accessible_memories` is an existing helper (line 69 of `memory_api_service.py`), correctly reused here
- The new `ids` filter parameter flows correctly through `memory_api.py` → `memory_api_service.py` → `memory_service.py`
- The dataset change gracefully degrades: if all requested kb_ids are unauthorized, it still returns an empty result set instead of an error
- Warning logs include enough context for auditing without leaking data

**One question (non-blocking):**
- The dataset service now silently filters out unauthorized kb_ids rather than returning an explicit error. This is a behavioral change — callers that relied on the old error response to detect authorization issues will now get a partial/empty result instead. Is this intentional, and are the API consumers (e.g. frontend) prepared for this?

Otherwise LGTM.

anujbolewar (commented):
The right security posture — unauthorized dataset membership should be filtered rather than leak existence via a hard error. Just be aware this is a behavior change for callers that relied on the rejection to detect misconfiguration, since they now receive an empty list. Also, adding ids to the memory filter params widens the query surface, so confirm ids are ownership-validated against the tenant before they reach the DB query rather than used as-is.

PR #18007, opened 2026-08-07
fix(chunk): render <img> tags as escaped text to prevent tracking

dcc123456:
### Summary

fix(chunk): render  tags as escaped text to prevent tracking

Replace DOMPurify.sanitize with a new sanitizeHtmlWithImagesAsText helper in both chunk-card components. The helper sanitizes HTML with DOMPurify and replaces each  element with an escaped text node of its outerHTML, so external image URLs are not loaded by the browser. This prevents image-based tracking/data exfiltration and layout shifts while keeping the img markup visible as literal text.

### DIFF
--- a/web/src/pages/chunk/parsed-result/add-knowledge/components/knowledge-chunk/components/chunk-card/index.tsx
+++ b/web/src/pages/chunk/parsed-result/add-knowledge/components/knowledge-chunk/components/chunk-card/index.tsx
@@ -9,9 +9,9 @@ import {
 } from '@/components/ui/tooltip';
 import type { ChunkDocType, IChunk } from '@/interfaces/database/dataset';
 import { cn } from '@/lib/utils';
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this PR! The approach of neutralizing `` tags by rendering their outerHTML as text is a clean way to prevent image-based tracking without losing the markup context for the user.

**What I checked:**
- The `sanitizeHtmlWithImagesAsText` helper correctly sanitizes via DOMPurify first, then replaces `` elements with escaped text nodes
- Both chunk-card components are updated consistently
- DOMPurify is already a dependency, so no new dependency introduced

- The type cast `as HTMLElement` is safe since DOMPurify with `RETURN_DOM: true` returns an `HTMLBodyElement`. Consider using `as HTMLElement` as-is – it works fine and the body element extends HTMLElement.
- An empty input string will return an empty body element whose `.innerHTML` is `""` – this is correct behavior.

LGTM, approved from my side.

PR #18008, opened 2026-08-07
Fix medium thinking chat failed in thinking

wangq8:


wangq8 (commented):
**Summary:** Removes a dead-code line (`ctx.agent_results[c.claim_id] = c.agent_result`) that was accidentally introduced in the recent refactor #17951 (commit `a25ff22ac`).

**Analysis:**

1. **Dead store, never read** — I searched the entire `rag/advanced_rag/` tree for `.agent_results` reads; the only occurrence is this removed write at line 149. No code reads `ctx.agent_results` anywhere.

2. **Redundant with the proper field** — `c.agent_result` is already being set 3 lines above (line 146) as a proper dataclass field on `ClaimTarget`. All downstream consumers (`_finalize`, `_merge_agent_results`, `has_any_chunks`, the sufficiency fusion block at lines 173–176) read from `c.agent_result` on the claim objects, not from the orphaned dict on `ctx`.

3. **Inconsistent with declared intent** — `OrchestratorContext.has_any_chunks()` in `types.py` explicitly comments: *"the `agent_results` dict is never written to, so reading it would leave the search phase stuck at locate forever."* This write violates that invariant.

4. **Dynamic attribute on a dataclass** — `agent_results` is not a declared field of `OrchestratorContext`; it is dynamically assigned. While plain `@dataclass` allows this, it is fragile — serialization, checkpointing, or deep-copy paths (especially in LangGraph state management) may silently drop or trip on undeclared attributes.

**Verdict:** ✅ LGTM. This is a clean, minimal fix that removes dead code, restores the invariant documented in `types.py`, and eliminates a likely serialization pitfall for the medium-thinking code path.

No regressions, no tests needed (removing dead code).

PR #18009, opened 2026-08-07
fix(api): map internal RetCode values to valid HTTP statuses in build_error_result

### LINKED ISSUE #17980: [Bug]: RAGFlow Hypercorn/h11 protocol bug — When an empty knowledge_id and query are received, it triggers build_error_result, but Hypercorn's h11 protocol state machine crashes, resulting in an empty response (0 bytes) instead of returning a JSON error.
### Self Checks

3e1191414

### RAGFlow image version

v0.26.4-707-g3e1191414

```Markdown

```

### Actual behavior

用dify远程连接ragflow的知识库出现的错误，两个软件装在同一台主机上

### Expected behavior

_No response_

### Steps to reproduce

```Markdown
错误现象：错误标题：
Failed to save/update External API
错误详情：
failed to connect to the endpoint:
http://192.168.5.12:9380/api/v1/dify/retrieval
错误问题：SSRF 代理拒绝私有 IP ✅ 已在上一步修复
RAGFlow Hypercorn/h11 协议 Bug — 当收到空 knowledge_id + query 时触发 build_error_result，但 Hypercorn 的 h11 协议状态机崩溃导致空响应（0 字节），而非返回 JSON 错误
Squid 看到空响应 → 502 — Dify 收到 502 后报告 "failed to connect"
```

### Additional information

_No response_

anneheartrecord:
### Summary

Fixes #17980.

`build_error_result()` assigns the internal `RetCode` value directly to `response.status_code`. `RetCode.ARGUMENT_ERROR` is 101 and `RetCode.AUTHENTICATION_ERROR` is 109 — both land in the HTTP 1xx informational range. h11 refuses to send a 1xx status as a final response (`LocalProtocolError`), so Hypercorn drops the connection and the client gets a 0-byte reply instead of the JSON error.

This is what breaks the Dify external knowledge integration reported in #17980: `/api/v1/dify/retrieval` with a missing `knowledge_id`/`query` returns nothing on the wire, an intermediate proxy turns that into 502, and Dify reports "failed to connect to the endpoint".

Reproduced with a minimal Quart + Hypercorn 0.18 / h11 0.16 probe:

- `response.status_code = 101` → h11 `LocalProtocolError: Can't send data when our state is ERROR`, 0 bytes sent
- `response.status_code = 400` → normal JSON delivery

The fix maps internal ret codes to real HTTP statuses (`ARGUMENT_ERROR`/`DATA_ERROR`/`OPERATING_ERROR` → 400, `PERMISSION_ERROR`/`AUTHENTICATION_ERROR` → 403 to match Dify's external knowledge API expectations, the rest of the sub-200 codes → 500). Codes that are already valid HTTP statuses (404, 403, 500, …) pass through unchanged, and the JSON body still carries the original `RetCode`, so body-reading clients are unaffected. `build_error_result` is only used by the Dify retrieval endpoint.

Unit tests added in `test/unit_test/api/utils/test_api_utils.py`; they fail on the old code with `Response [101 SWITCHING PROTOCOLS]` and pass with the fix.

### DIFF
--- a/api/utils/api_utils.py
+++ b/api/utils/api_utils.py
@@ -277,9 +277,17 @@ def build_error_result(code=RetCode.FORBIDDEN, message="success"):
     response = {"code": code, "message": message}
     response = _safe_jsonify(response)
     if hasattr(response, "status_code"):
-        http_status = RET_CODE_TO_HTTP_STATUS.get(code)
-        if http_status is None:
-            http_status = int(code) if 200 <= int(code) <= 599 else 500
+        ret_code = int(code)
+        http_status = RET_CODE_TO_HTTP_STATUS.get(ret_code)
+        # The status logs below carry the ret code only; `message` can hold user input.
+        if http_status is not None:
+            logging.debug("build_error_result: ret code %s mapped to HTTP %s", ret_code, http_status)
+        elif 200 <= ret_code <= 599:
+            http_status = ret_code
+            logging.debug("build_error_result: ret code %s used as HTTP status", ret_code)
+        else:
+            http_status = 500
+            logging.warning("build_error_result: unmapped ret code %s, falling back to HTTP 500", ret_code)
         response.status_code = http_status
     return response
 
--- a/api/utils/api_utils.py
+++ b/api/utils/api_utils.py
@@ -256,11 +256,31 @@ def get_json_result(code: RetCode = RetCode.SUCCESS, message="success", data=Non
     return _safe_jsonify(response)
 
 
+# Internal RetCode values below 200 collide with the HTTP 1xx (informational) range.
+# h11 refuses to send a 1xx status as a final response, so Hypercorn drops the
+# connection and the client receives an empty reply instead of the JSON error body.
+# Map them to real HTTP statuses; the body keeps the original RetCode.
+RET_CODE_TO_HTTP_STATUS = {
+    RetCode.EXCEPTION_ERROR: 500,
+    RetCode.ARGUMENT_ERROR: 400,
+    RetCode.DATA_ERROR: 400,
+    RetCode.OPERATING_ERROR: 400,
+    RetCode.CONNECTION_ERROR: 500,
+    RetCode.RUNNING: 500,
+    RetCode.PERMISSION_ERROR: 403,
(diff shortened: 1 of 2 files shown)


anujbolewar (commented):
Legitimate — h11 refusing 1xx as a final response and silently dropping the connection is a nasty failure mode, and mapping internal codes to real statuses while preserving the RetCode in the body is a clean contract. The fallback keeps out-of-range codes to 500. One nit: FORBIDDEN is the default arg, so confirm the default and the mapping cover the same code space, and consider whether 400 or 422 is the better client-error choice for DATA and ARG errors given the Dify knowledge API compatibility note.

anneheartrecord:
Pushed e9ee932 for both review points.

**Logging.** Each branch now logs its decision exactly once: debug for a code that hits the map, debug for a code that is already a valid HTTP status, warning when an unmapped code falls back to 500. The lines carry the ret code and the resolved status only — `message` stays out of them since it can contain user input.

**Fallback test.** Added: `NOT_EFFECTIVE` (10) is absent from the map and below 200, so it must come out as 500.

@anujbolewar on the default arg — `FORBIDDEN` is 403, so it never enters the map and passes through the already-valid-status branch unchanged. Added a test that pins it.

On 400 vs 422: the three `ARGUMENT_ERROR` call sites here are an unparseable body, missing `knowledge_id`/`query`, and an invalid `retrieval_setting`. Those are malformed requests rather than well-formed ones that fail validation, so 400 reads right to me. `DATA_ERROR` isn't raised by this endpoint at all; I left it at 400 for consistency rather than guess at 422 with no call site to justify it.

anneheartrecord:
Good catch on the count assertion — three total log lines would still pass if one branch logged twice and another logged nothing. Pushed 1679059: the test now walks the branches one at a time with `caplog.clear()` between them, and each row pins the ret code, the resolved status, the log level, and which branch ran. The fallback test asserts the exact warning text rather than a substring.

Mutation-checked all of it: double-logging one branch while silencing another, swapping the two debug texts, downgrading the fallback warning to debug, changing the fallback status to 400, and adding `message` to any of the three calls each turn the suite red now.

PR #18018, opened 2026-08-08
fix(tests): isolate token chunker imports

### LINKED ISSUE #17985: pytest: collecting rag/flow/tests/ together with test/unit_test/rag/ corrupts deepdoc.parser (cascading PlainParser ImportError)
## Summary

Collecting `rag/flow/tests/` and `test/unit_test/rag/` **together in a single pytest invocation** corrupts the cached `deepdoc.parser` package module in `sys.modules`, which then makes **every** `from deepdoc.parser import X` fail with a cascading `ImportError: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser' (unknown location)`.

This is **not** a `deepdoc` native-build problem and **not** a bug in any single test file. Each suite collects cleanly on its own.

## Reproduction

Each suite alone is green:

```bash
python -m pytest rag/flow/tests/ --collect-only -q          # 12 tests, OK
python -m pytest test/unit_test/rag/test_naive_merge.py --collect-only -q   # 23 tests, OK
```

Combined collection fails:

```bash
python -m pytest rag/flow/tests/test_token_chunker_tag_overlap.py test/unit_test/rag/test_naive_merge.py --collect-only -q
# ERROR collecting rag/flow/tests/test_token_chunker_tag_overlap.py
# ImportError: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser' (unknown location)
```

The failure is **order-independent** (swapping the two args still fails) and **survives `--import-mode=importlib`**.

## Root cause

`PlainParser` is only a symptom. At the moment of failure, `sys.modules` shows:

```
deepdoc.parser.pdf_parser  cached: __file__=None  __path__=None  hasPlainParser=False
deepdoc.parser.__path__=None          # the deepdoc.parser package module itself is broken/partial
deepdoc.__path__=['/home/.../ragflow/deepdoc']   # correct
```

`deepdoc.parser.pdf_parser` has `__file__=None` and its `SourceFileLoader.exec_module` is never invoked, so the real `deepdoc/parser/pdf_parser.py` is never executed — it was cached as a namespace/partial module. Any later `from deepdoc.parser import ...` then fails against this poisoned cache, and the error chain cascades into `docs_generator`, `invoke`, `rag.flow.*`, etc.

The trigger is structural to the test layout:

- `test/unit_test/rag/` and `rag/flow/tests/` are **not packages** (no `__init__.py`). pytest's default `prepend` import mode puts each of these directories on `sys.path[0]`.
- `test/unit_test/rag` is itself named `rag`, and `test/unit_test/` contains a `deepdoc/` directory (`test/unit_test/deepdoc/parser/...`).
- When both suites are imported in one process — the flow suite via `rag.flow.parser.pdf_chunk_metadata → api.db.services → deepdoc.parser`, and the `test/unit_test` suite via `rag.nlp` and its conftest — `deepdoc` / `deepdoc.parser` resolution hits a namespace / partial-module cache collision and `deepdoc.parser` ends up cached as a non-package broken module.

Note: CPython 3.13 executes most of the import machinery in C, bypassing `builtins.__import__`, `module_from_spec`, and `SourceFileLoader.exec_module`, so the precise first corruption event cannot be observed via Python-level import hooks; the `sys.modules` end-state above plus the failure conditions are sufficient to localize the mechanism.

## Independent verification that `deepdoc` is fine

```bash
python -c "import deepdoc.parser.pdf_parser as m; print(m.__file__)"   # OK
python -c "from api.db.services.task_service import TaskService; print('ok')"  # OK
# pure-python: import rag.nlp then import rag.flow.chunker.token_chunker  -> OK
```

## Suggested fixes (out of scope for this issue)

1. **Decouple the import path**: `rag/flow/parser/pdf_chunk_metadata.py` does `from api.db.services.file_service import FileService` at module top level, which pulls in the entire `api.db` stack (and thus `deepdoc.parser`) whenever the chunker package is imported. Moving that heavy import behind a function / lazy import removes the chunker from the `deepdoc` import chain.
2. **Make test directories real packages** (add `__init__.py` to `test/unit_test/rag`, `rag/flow/tests`, etc.) so pytest resolves them unambiguously instead of as `sys.path` roots / namespace portions.
3. **Keep the two suites in separate pytest invocations** in CI as a stopgap (each already collects green alone).

## Workaround for the affected test

`rag/flow/tests/test_token_chunker_tag_overlap.py` collects and passes cleanly **on its own** (and will in CI once `deepdoc` is built). It only breaks when collected in the same process as `test/unit_test/rag/`. No code change is required in that test.

RerankerGuo:
### Review scope

- Production diff: `pdf_chunk_metadata.py` (`+7/-3`).
- Regression evidence: `test_flow_chunker_import.py` (`+46/-0`).
- The lazy imports and regression test are atomic because the defect is a cross-suite import collision.

### Summary

Fix the pytest collection conflict in #17985 by keeping PDF preview and parser
dependencies out of `pdf_chunk_metadata.py`'s module import path.

`test/unit_test/rag/conftest.py` intentionally provides a lightweight
`deepdoc.parser.pdf_parser` stub for `rag.nlp` tests. Importing TokenChunker
previously pulled in database services, `LOCK_KEY_pdfplumber`, and
`RAGFlowPdfParser` at module load time. That coupled the chunker import to the
stub's exact symbol set and failed against a partial or poisoned parser module.

This change:

- imports `FileService` and `File2DocumentService` only when a source blob is requested;
- imports `LOCK_KEY_pdfplumber` only when PDF page images are loaded;
- imports `RAGFlowPdfParser` only when a legacy `position_tag` is parsed;
- verifies TokenChunker imports when the parser stub exposes no PDF parser symbols;
- executes the `position_tag` path against the runtime parser stub and verifies the normalized coordinates.

Runtime PDF preview and position-tag behavior is unchanged; only import timing
changes.

### Follow-up

#18184 tracks the remaining root cause outside this PR: `parser.py` still has the
same eager service/parser import class, and the affected test directories still
need a stable package/import layout. This PR remains the narrow stopgap for the
TokenChunker collection failure rather than leaving that broader work implicit.

Closes #17985.

### Validation

- Strengthened import-time regression test:
  - previous PR head `c0db8016a`: `1 failed` with missing `RAGFlowPdfParser`;
  - current branch: `1 passed`.
- Runtime lazy-import regression: `position_tag` invokes the stubbed
  `RAGFlowPdfParser.extract_positions` and returns normalized PDF coordinates.
- Targeted test: `uv run pytest test/unit_test/rag/test_flow_chunker_import.py -q` (`1 passed`).
- Original combined reproduction: `25 passed`.
- Combined regression run: `26 passed`.
- `uvx ruff check test/unit_test/rag/test_flow_chunker_import.py`: passed.
- `uvx ruff format --check test/unit_test/rag/test_flow_chunker_import.py`: passed.
- Production-file Ruff baseline: the same 3 existing findings on `main` and this PR.
- `python -m py_compile`: passed.
- `git diff --check`: passed.

proof: sufficient

### DIFF
--- a/rag/flow/parser/pdf_chunk_metadata.py
+++ b/rag/flow/parser/pdf_chunk_metadata.py
@@ -22,11 +22,8 @@ import numpy as np
 import pdfplumber
 from PIL import Image
 
-from api.db.services.file2document_service import File2DocumentService
-from api.db.services.file_service import FileService
 from common import settings
 from common.misc_utils import get_uuid
-from deepdoc.parser.pdf_parser import LOCK_KEY_pdfplumber, RAGFlowPdfParser
 from rag.utils.base64_image import image2id
 
 PDF_PREVIEW_GAP = 6
@@ -47,6 +44,8 @@ def _extract_raw_positions(item):
 
     position_tag = item.get("position_tag")
     if isinstance(position_tag, str) and position_tag:
+        from deepdoc.parser.pdf_parser import RAGFlowPdfParser
+
         return [[pos[0][-1], *pos[1:]] for pos in RAGFlowPdfParser.extract_positions(position_tag)]
 
     position_int = item.get("position_int")
@@ -187,6 +186,9 @@ def finalize_pdf_chunk(chunk):
 
 
 def _fetch_source_blob(from_upstream, canvas):
+    from api.db.services.file2document_service import File2DocumentService
+    from api.db.services.file_service import FileService
+
     if canvas._doc_id:
         bucket, name = File2DocumentService.get_storage_address(doc_id=canvas._doc_id)
         return settings.STORAGE_IMPL.get(bucket, name)
@@ -196,6 +198,8 @@ def _fetch_source_blob(from_upstream, canvas):
 
 
 def _load_pdf_page_images(blob, zoom=PDF_PREVIEW_ZOOM):
+    from deepdoc.parser.pdf_parser import LOCK_KEY_pdfplumber
+
     with sys.modules[LOCK_KEY_pdfplumber]:
         with pdfplumber.open(io.BytesIO(blob)) as pdf:
             return [page.to_image(resolution=72 * zoom, antialias=True).annotated for page in pdf.pages]
--- a/test/unit_test/rag/test_flow_chunker_import.py
+++ b/test/unit_test/rag/test_flow_chunker_import.py
@@ -1,0 +1,32 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+import importlib
+import sys
+
+
+def test_token_chunker_imports_without_pdf_parser_symbols(pdf_parser_stub, monkeypatch):
+    monkeypatch.delattr(pdf_parser_stub, "RAGFlowPdfParser")
+    for module_name in [
+        "rag.flow.chunker",
+        "rag.flow.chunker.token_chunker",
+        "rag.flow.parser.pdf_chunk_metadata",
+    ]:
+        sys.modules.pop(module_name, None)
+
+    token_chunker = importlib.import_module("rag.flow.chunker.token_chunker")
+
+    assert callable(token_chunker._merge_text_chunks_by_token_size)
--- a/test/unit_test/rag/test_flow_chunker_import.py
+++ b/test/unit_test/rag/test_flow_chunker_import.py
@@ -30,3 +30,17 @@ def test_token_chunker_imports_without_pdf_parser_symbols(pdf_parser_stub, monke
     token_chunker = importlib.import_module("rag.flow.chunker.token_chunker")
 
     assert callable(token_chunker._merge_text_chunks_by_token_size)
+
+    position_tag = "@@1\t10\t20\t30\t40##"
+    extracted_tags = []
+
+    class RuntimePdfParser:
+        @staticmethod
+        def extract_positions(value):
+            extracted_tags.append(value)
+            return [([0], 10.0, 20.0, 30.0, 40.0)]
+
+    monkeypatch.setattr(pdf_parser_stub, "RAGFlowPdfParser", RuntimePdfParser, raising=False)
+
+    assert token_chunker.extract_pdf_positions({"position_tag": position_tag}) == [[1, 10.0, 20.0, 30.0, 40.0]]
+    assert extracted_tags == [position_tag]


xugangqiang:
Thanks for the fix. The change is **correct** — moving `FileService`/`File2DocumentService` and `LOCK_KEY_pdfplumber` behind `_fetch_source_blob`/`_load_pdf_page_images` does resolve the collection error in #17985, and the runtime behavior is unchanged. I verified the import chain and the regression test does catch the regression.

That said, I don't think this is the **optimal** solution:

- The fix only works because the `test/unit_test/rag` conftest stub happens to expose `RAGFlowPdfParser` (`_make_pdf_parser_stub` only defines `RAGFlowPdfParser`, not `PlainParser`/`LOCK_KEY_pdfplumber`). `pdf_chunk_metadata.py` still imports `deepdoc.parser.pdf_parser` at module top level for `RAGFlowPdfParser`, so the chunker remains coupled to `deepdoc.parser` at import time — it just happens to line up with the stub.
- The real root cause is the test layout: `test/unit_test/rag` and `rag/flow/tests` are not packages (no `__init__.py`) and sit on `sys.path` as roots, while `test/unit_test/` also contains a `deepdoc/` tree, which poisons the `deepdoc`/`deepdoc.parser` entries in `sys.modules`. This PR is a band-aid on the symptom, not the cause.
- `rag/flow/parser/parser.py:47` has the identical shape (top-level `PlainParser` + `api.db.*` imports) and is the same class of bug waiting to resurface the moment a combined collection touches it.

A more robust fix would address the root cause — e.g. making the test directories real packages so pytest resolves them unambiguously, and/or fully decoupling `pdf_chunk_metadata` from `deepdoc.parser` at import time by lazy-importing `RAGFlowPdfParser` as well.

**Would you be open to applying the optimal fix (package the test dirs / fully lazy-load the `deepdoc.parser` imports) instead of this minimal band-aid?** I'm happy to help scope it if you'd prefer to keep this PR as a stopgap plus a follow-up issue for the rest.

RerankerGuo:
Thanks for the detailed review. I applied the full lazy-loading option in `94e840e65`: `RAGFlowPdfParser` is now imported only inside the legacy `position_tag` branch, so importing TokenChunker no longer depends on any symbols from `deepdoc.parser.pdf_parser`. The strengthened regression test removes `RAGFlowPdfParser` from the existing stub before a fresh TokenChunker import; it failed on the previous PR head and now passes. The original combined reproduction remains green (`25 passed`), and the combined regression run reports `26 passed`. I kept the broader test-package layout unchanged because the production import boundary is now fully decoupled without expanding this PR into a test-tree reorganization.

xugangqiang:
@RerankerGuo Thanks for the fix — the issue itself is resolved and there is no runtime regression, so merging it as a stopgap is reasonable.

While reviewing, there are three points I'd like to confirm whether you'd be willing to help follow up on, so we can close the loop on this "band-aid":

**(a) Strengthen test coverage**
The current `test_flow_chunker_import.py` only verifies *import-time* decoupling (after removing `RAGFlowPdfParser` from the stub it can still be imported, and asserts `_merge_text_chunks_by_token_size` is callable), but the three moved lazy-import lines are **never triggered at runtime**:
- `from deepdoc.parser.pdf_parser import RAGFlowPdfParser` inside `_extract_raw_positions` only executes when `position_tag` is a non-empty string;
- `File2DocumentService`/`FileService` in `_fetch_source_blob`, and `LOCK_KEY_pdfplumber` in `_load_pdf_page_images`, are likewise never called.

In other words, if someone misspells one of these symbol names, this "regression test" will not flag it. Would you be willing to add a case that actually walks the runtime lazy path (e.g. construct an item with `position_tag` and call `extract_pdf_positions`, so `_extract_raw_positions` actually executes that lazy import), extending the guard from "import-time" to "runtime"?

**(b) Open a follow-up issue to track the root cause**
`rag/flow/parser/parser.py:28-48` also imports `api.db.services.*` and `deepdoc.parser.pdf_parser` (including `PlainParser`/`RAGFlowPdfParser`/`VisionParser`) at the top level, and line 53 again imports `pdf_chunk_metadata` at the top level — this is the **same class of bug**, and collecting `parser.py` under the same stub environment will hit the wall again. The root cause also includes the test layout: `test/unit_test/rag` and `rag/flow/tests` are not packages (no `__init__.py`) yet are mounted on `sys.path` as roots, and the `deepdoc/` test tree under `test/unit_test/` pollutes the `deepdoc`/`deepdoc.parser` entries in `sys.modules`. Would you be willing to open a follow-up issue recording these two points (the same-class decoupling for parser.py + normalizing the test directories) as closed-loop tracking for this PR?

**(c) Back-link the issue in the PR description**
If the issue in (b) is created, would you add a back-link in the PR description, making it clear this is a documented temporary measure rather than something forgotten, so later reviewers can align on it?

If you're willing to do the above three points, I can help draft the follow-up issue text or an initial version of the strengthened test. Do you agree to close with (a)+(b)+(c) before merging?

RerankerGuo:
@xugangqiang Thank you for the detailed follow-up. I agree with closing the loop through (a) + (b) + (c), and all three are now completed:

- **(a) Runtime coverage:** commit `5298e5680` extends the existing regression test beyond import-time decoupling. After importing TokenChunker against a parser stub with no PDF parser symbols, the test installs the runtime `RAGFlowPdfParser` stub, calls `extract_pdf_positions` with a non-empty `position_tag`, verifies that `extract_positions` receives that exact tag, and asserts the normalized coordinates. This will fail if the lazy symbol name is misspelled or the runtime path is not executed. Targeted pytest reports `1 passed`; Ruff check, Ruff format check, and `git diff --check` pass.
- **(b) Root-cause tracking:** opened #18184 with a current reproduction for the remaining `parser.py -> api.db.services -> deepdoc.parser` failure, the ambiguous test-package layout, and explicit acceptance criteria for both scopes.
- **(c) PR back-link:** added #18184 to the PR description and explicitly described this PR as the narrow TokenChunker stopgap so the broader work remains visible after #17985 closes.

The suggested direction was clear, so no additional draft was needed. Thanks again for identifying the missing runtime guard and for helping keep the broader root cause tracked.

xugangqiang:
@yuzhichang @JinHai-CN @yingfeng  Please help to review.

PR #18023, opened 2026-08-09
fix(pdf/layout): replace KMeans-only column detection with gap + KMeans hybrid

xugangqiang:
## Summary

Replace the KMeans-only column detector (`AssignColumn`) in the Go PDF layout
package with a **gap + KMeans hybrid**. The replacement keeps the
`lyt.AssignColumn(boxes)` signature that `parser.go` calls by name, so the
pipeline picks it up automatically.

This PR also fixes a regression introduced by blindly trusting `gap >= 2`
(see "Fix A" below).

## Problem

- **KMeans-only** (old): x0 clustering with silhouette `k` selection
  **over-splits single-column pages** — 16/18 detected as multi-column on a
  first 18-page labeled sample (11.1% ACC).
- **gap-only** (whitespace-gutter voting, the safe lower bound): never
  over-splits, but misses academic double columns whose gutter is bridged by
  full-width front matter. On the full **70-page** labeled corpus:
  **25.7% ACC** (false-split 18, miss 34).

## Solution

The two detectors have **complementary error directions** (gap = safe floor,
KMeans = recoveries only), so we combine them:

1. **Gap gutter voting** is the safe base. When it finds `>= 2` gutters we
   trust it and partition by `util.KMeans1D(g)` — **unless the columns are
   narrow**, which means a *table* (one text block, not text columns); tables
   are returned as a single column directly so the balance gate does not
   re-split them into 2.
2. When gap reports a single column, a **forced `k = 2` density-aware k-means
   (k-means++) on the BODY x0** (full-width titles/abstracts excluded via
   `dropFullWidth`) with a **balance gate** decides whether a real second
   column exists: each cluster holds `>= minModeFrac` (30%) of body lines
   **and** the centroid separation is `>= minSepFrac` (10%) of page width. An
  unbalanced split — the usual KMeans false-split on a single page — is
  dropped, so the page stays single column.
3. After any partition, **empty and too-sparse columns are pruned**
  (`minColLineFrac = 0.12`): a k-means centroid that captures no lines, or
  fewer than 12% of the page's lines, is not a real reading column — it is an
  indentation sliver or a stray block. This recovers pages where gap
  over-split (e.g. a spurious third column) without harming genuine sparse
  multi-column pages (the labeled corpus's only real triple holds 22%+ per
  column, and the sparsest real double 17.7%+ — both clear of the 12% floor).

### Fix A (this PR's second commit): stop trusting gap >= 2 blindly

The first hybrid commit trusted `gap >= 2` unconditionally. On real pages that
over-splits: indentation and table cells produce `gap` values of 2–9 with no
real column, while genuine doubles (title-bridged gutter) report `gap == 1`
and were missed. The fix:

- `gap >= 2` with **narrow** columns (max col width `< 0.22 × page`) → table →
  return 1 directly (skip the balance gate).
- `gap == 2` → **defer to the balance gate** (a fake gutter from indentation
  has no real second cluster; a real 2-col passes).
- `gap >= 3` with **wide** columns → trust as a real multi-column layout,
  capped at `maxColumnCount = 4` so spurious gutters cannot split a page into
  many columns.
- balance gate unchanged (`fmin >= 0.30`).

### Fix B (this PR's column-prune follow-up): balance gate and prune must use the same line set

The balance gate deliberately clusters the BODY — full-width front matter
(title / abstract / heading) is dropped via `dropFullWidth` so a gutter
bridged by front matter does not create a fake second column. But
`pruneColumns` was then counting ALL lines, including that full-width front
matter. Because full-width lines are reassigned to the nearest centroid, a
large full-width block inflates one column and can push the other below the
`minColLineFrac` (12%) floor, silently collapsing a real, gate-accepted
two-column page back to one.

Fix: `balancedBodyK2` now also returns the `body` slice it clustered on, and
the balance branch of `detectColumnCount` calls `pruneColumns(body, cents)`
instead of `pruneColumns(lines, cents)`. The `g >= 3` path is unchanged — it
already clusters and prunes on the same all-lines set, so it stays
self-consistent. `TestAssignColumn_FullWidthBridgeKeepsBalancedTwoColumn`
reproduces the collapse (RED) and verifies the fix (GREEN).

## Measured accuracy (70 human-labeled real pages, Go test harness)

| detector              | ACC   | false-split | miss |
|-----------------------|-------|-------------|------|
| gap-only (baseline)   | 25.7% | 18          | 34   |
| gap + KMeans (this PR)| 78.6% | 5           | 10   |

(The earlier 88.9% / 94.4% figures came from the comfortable 18-page sample
and must not be used as a target — they are circular. The test now scores on
all 70 labeled pages.)

The corpus label for `2023-07-03.pdf.json` page 0 was corrected from
single-column to double-column (real narrow sidebar ≈ half the main column);
the gap-only detector had identified its gutter correctly all along, which is
why both the baseline and the hybrid improve by one page after the relabel.

## Residual misses (10) — known limitations

- **9 are title-bridged academic doubles.** Their body x0 clusters *overlap*
  heavily (minority fraction 0.05–0.29), so the balance gate's `fmin >= 0.30`
  floor correctly rejects them. Lowering `fmin` to catch them adds ~18
  single-column over-splits → net worse. A **vertical-coverage / reading-flow
  rescue (option B)** was evaluated: it recovers 7–8 of the 9 misses but adds
  19 false-splits (wide tables and wide single pages have *identical* vertical
  coverage to a real second column) → net **−8 to −10 accuracy points**.
  Option B is rejected as a cheap heuristic; recovering these needs a genuine
  reading-order / column-structure model, not a coverage proxy.
- **1 is a three-column page** (`truth = 3`). Three-column detection is out of
  scope for this 2-column detector.

These 10 misses are tracked separately in **#18079** (Go-side only) and are
intentionally left out of this PR.

## Residual false-splits (5) — known limitations

After the sparse-column prune, 5 pages are still over-split (got > truth). All
five are the *indentation / dense-text* family: the page has a genuine
secondary x0 mode (block indentation, a dense CJK sub-block) that passes the
balance gate (gap == 1/2) or the gap >= 3 path, and the spurious column holds
8–27 lines at 23–33% ratio — squarely inside the real-double envelope (the
labeled corpus's sparsest real double is 17.7%, its only real triple 22%+ per
column). No count/width threshold can reject them without also breaking
genuine doubles/triples, so they are intentionally left out of this PR and
tracked in **#18079** (reading-order model).

- `16_dense_cjk p1`, `2023-07-03 p1 / p2 / p8` — single column split into 2
  (spurious column 8–27 lines).
- `2023-07-03 p7` — single column split into 4 (gap=7 → capped to 4;
  spurious columns are indentation slivers).

## Files changed

- `internal/deepdoc/parser/pdf/layout/combined_column.go` — hybrid detector;
  table collapse (`tableMaxColFrac = 0.22`); `gap == 2` deferral to balance
  gate; `maxColumnCount = 4` cap; sparse-column prune (`minColLineFrac =
  0.12`, `pruneColumns`); `gapColumnWidths`; `clampInt`; **balance gate and `pruneColumns` now share the same BODY line set (Fix B)**.
- `internal/deepdoc/parser/pdf/layout/combined_column_test.go` — labeled eval
  over 70 pages with a **live gap-only baseline** (no hardcoded accuracy
  target); `TestKmeansK2PlusPlus_Smoke`; **`TestAssignColumn_FullWidthBridgeKeepsBalancedTwoColumn` (full-width-bridged balanced two-column regression for Fix B)**.
- `internal/deepdoc/parser/pdf/layout/layout.go` — removed the dead KMeans
  `AssignColumn` and its helpers; `AssignColumn` now lives in
  `combined_column.go`.
- `internal/deepdoc/parser/pdf/layout/layout_test.go` — updated `TestAssignColumn`
  fixtures to carry real `X1` geometry (the gap detector needs text width, not
  just x0).

`parser.go` needed no change.

## Test plan

```bash
bash build.sh --test ./internal/deepdoc/parser/pdf/layout/

The Fix B regression test (`TestAssignColumn_FullWidthBridgeKeepsBalancedTwoColumn`) runs on every `go test ./internal/deepdoc/parser/pdf/layout/` with no external corpus, guarding the balance-gate / prune line-set consistency.
```

All layout tests pass. The labeled test asserts the hybrid does **not regress
the gap-only baseline by more than 5 points** (a live comparison, not a tuned
constant). Reported: `ACC=78.6% false-split=5 miss=10 (n=70)`.

Stderr: (empty)
Exit Code: 0
Signal: (none)


xugangqiang:
Added a follow-up commit (**Fix B**) addressing the balance-gate / prune line-set inconsistency in the column detector:

- The balance gate clusters the **BODY** (full-width title/abstract dropped via `dropFullWidth`), but `pruneColumns` was counting **ALL** lines, so a large full-width front-matter block could inflate one column and silently collapse a real, gate-accepted two-column page back to one.
- `balancedBodyK2` now also returns the `body` slice it clustered on, and the balance branch of `detectColumnCount` calls `pruneColumns(body, cents)` instead of `pruneColumns(lines, cents)`. The `g >= 3` path is unchanged — it already clusters and prunes on the same all-lines set, so it stays self-consistent.
- Added `TestAssignColumn_FullWidthBridgeKeepsBalancedTwoColumn` (RED → GREEN, no external corpus needed) covering a full-width-bridged balanced two-column page.

The two earlier review points are already handled on this branch:
- the misleading 94.4% figure is removed (docs commit), and
- the labeled eval now ships CI-runnable synthetic fixtures (`combined_column_synthetic_test.go`), so accuracy is guarded on every `go test ./...` without the gitignored corpus.

PR description updated with a **Fix B** section plus Files-changed / Test-plan notes.

PR #18031, opened 2026-08-10
fix(dataset): shrink table height when bulk operate bar is visible

dcc123456:
### Summary

fix(dataset): shrink table height when bulk operate bar is visible

When rows are selected, the BulkOperateBar appears in the CardHeader and pushes content down, but the DatasetTable's max height stayed at calc(100vh-280px), causing the table body to overlap with the absolutely-positioned RAGFlowPagination at the bottom.

Add a `bulkOperateBarVisible` prop to DatasetTable that shrinks the table max height to calc(100vh-320px) when the bulk bar is visible, keeping the pagination clear of the table rows.

### DIFF
--- a/web/src/pages/dataset/dataset/dataset-table.tsx
+++ b/web/src/pages/dataset/dataset/dataset-table.tsx
@@ -47,6 +47,7 @@ export type DatasetTableProps = Pick<
 > &
   Pick<UseRowSelectionType, 'rowSelection' | 'setRowSelection'> & {
     showManageMetadataModal: (config: ShowManageMetadataModalProps) => void;
+    bulkOperateBarVisible?: boolean;
   };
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Clean fix, LGTM.

The problem is clear — when the BulkOperateBar appears in the CardHeader, the table body overlaps with the pagination at the bottom. The 40px offset (280→320) matches the approximate height the bulk bar adds to the header area.

One minor suggestion: the magic number 320 (or the delta of 40) would be easier to maintain as a CSS variable or a shared constant so that if the BulkOperateBar height changes in the future, only one value needs updating. But that is a nitpick — the fix itself is correct and self-contained.

Approved from a review standpoint. 👍

PR #18033, opened 2026-08-10
fix: handle disabled documents in dataset structures

buua436:

Improve knowledge-base graph, timeline, and mindmap generation by batching concurrent embeddings, reducing Elasticsearch refresh overhead, correctly handling dataset-level deletion and alteration detection, and excluding disabled documents from viewing and rebuilding.

### Type of change

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -1845,7 +1845,10 @@ async def get_dataset_structure(dataset_id: str, tenant_id: str, kind: str, keyw
     from api.apps.services import structure_graph_common as sgc
 
     keywords = (keywords or "").strip()
+    _, active_doc_ids = await _current_dataset_docs(dataset_id)
     disabled_doc_ids = await _disabled_dataset_doc_ids(dataset_id)
+    if not active_doc_ids:
+        return True, empty
 
     def _row_template_id(row: dict) -> str | None:
         raw = row.get("compilation_template_ids")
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Thanks for this PR! Good overall direction — handling disabled documents across dataset structures was definitely a gap. Some feedback:

**Strengths:**
- The disabled-document filtering is applied consistently across the full pipeline: graph view (get_dataset_structure), rebuild (_struct_rebuild_graph_json), merge (_do_build), alter detection (_current_dataset_docs, _eligible_doc_ids_for_kind), and provenance (_involved_doc_ids_paged).
- Batched embedding (_embed_rows) with a semaphore-based concurrency cap significantly reduces the serial bottleneck that existed in the old per-row encode loop.
- Concurrent bucket merging via _merge_buckets + asyncio.gather is a nice complement to the batching change.
- The rebuild_all trigger when disabled docs exist is the correct semantic — shared entities/relations need recalculation without the disabled document's data.
- Suppressing refresh on bulk inserts and doing one final refresh is a meaningful ES performance win.

**Concerns / Suggestions:**

1. **Duplicated _disabled_doc_ids helper** — The same logic appears in three places: dataset_structure_merger.py:_disabled_doc_ids, dataset_api_service.py:_disabled_dataset_doc_ids, and inline in structure.py:_struct_rebuild_graph_json. Consider extracting a shared utility (e.g. in document_service.py or a common helper) to avoid drift.

2. **inspect.signature for refresh detection** (dataset_structure_merger.py:161) — This is fragile. The refresh parameter exists in both the ES8 connector and the Infinity connector but might differ across backends. Using getattr with a default or a try/except at import time (caching the capability) would be more robust and avoids calling inspect.signature on every insert call.

3. **_row_has_enabled_source fallthrough** (structure_graph_common.py:255: "return not doc_id or ...") — When doc_id is absent the function returns True, meaning rows without any document association are never filtered. This is probably the correct default for safety, but it is worth documenting explicitly — a row with no provenance fields at all will pass through silently, which could hide data issues in production.

4. **build_bucket large-bucket path** — The small-bucket path (total below threshold) has the disabled-doc filter applied to all rows. The large-bucket sampling path applies it to top entities (set_a) and target entities (set_t), but the relation fetch is filtered only by source-entity membership (a_name_terms), not by document eligibility. Relations whose source entity passed the filter could still belong to disabled docs if those docs share entity names. This edge case is probably rare but worth noting.

5. **The doc_ids_kwd field is now read alongside source_doc_ids** in _involved_doc_ids_paged — this is correct and matches the merge code which writes both fields, but make sure the alter-check for mindmap and timeline kinds also picks up the right provenance field, since those stored rows may use different field naming.

Overall this is a solid fix. I would prioritize addressing concern #1 (deduplication) and #2 (inspect.signature) before merging. The rest can be follow-ups.

PR #18034, opened 2026-08-10
fix(ingestion): avoid duplicate chunk text injection in Extractor prompts

jay77721:
## Summary

The Extractor field-extraction path decides whether to suppress the automatic chunk-text append in `buildExtractorMessages` by checking whether the prompt literally contains `{text}` or `{chunks}`. That check had two problems:

1. **Too narrow** — a prompt referencing `{content_with_weight}` or any other chunk field still got the chunk text appended a *second* time, duplicating it in the final LLM message.
2. **Too eager** — a `{text}` that could not be resolved against the current chunk (field absent) suppressed the append anyway, leaving the LLM with a literal `{text}` placeholder and no chunk content.

Additionally, the "before vs after substring comparison" approach used to fix these issues was fragile: if a chunk field value happened to contain a placeholder substring (e.g. `title = "{text}"`), the check could be fooled.

## Fix

`substituteChunkPlaceholders` now returns `(string, bool)` — the bool signals whether a content-bearing placeholder (`{text}`, `{chunks}`, `{content_with_weight}`) was actually replaced. The Invoke path uses this signal directly:

```go
var subP, subS bool
callIn.prompt, subP = substituteChunkPlaceholders(in.prompt, ck, text)
callIn.systemPrompt, subS = substituteChunkPlaceholders(in.systemPrompt, ck, text)
callChunkText := text
if subP || subS {
    callChunkText = ""
}
```

The content set (`text`, `chunks`, `content_with_weight`) is a package-level `map[string]bool` defined in the same scope as the replacement logic, so the two cannot silently diverge.

The `{text}` placeholder falls back to `chunkText` (the primary chunk body, same as the append path `text` variable) when the chunk has no explicit `text` field. This ensures `{text}` always means "the document body" regardless of which field carries it.

## Why bool-return over substring comparison

| Aspect | Substring comparison | Bool return |
|---|---|---|
| Drift risk | needs a separate list, manually maintained | set and lookup in same scope |
| Substring masking | field values containing `{text}` fool the check | tracks which key was actually replaced |
| Scans | up to 4 `Contains` calls per chunk | 1 pass during substitution |

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| `{text}` placeholder, field present | no append | no append (resolved, suppressed) |
| `{text}` placeholder, only `content_with_weight` | no append, bare `{text}` in prompt | no append (fallback resolves `{text}`) |
| `{content_with_weight}` placeholder | **append → duplicated** | no append |
| non-content placeholder (e.g. `{title}`) | no append (false positive) | append |
| field value contains placeholder substring | depends on values | correct (tracks substitution, not substrings) |
| static prompt (resume template) | append | append |
| `{@chunks}` resume regex path | append | append (untouched) |

## Tests

| Test | Locks |
|---|---|
| `ContentWithWeightPlaceholder` | `{content_with_weight}` does not duplicate the chunk body |
| `NonContentPlaceholderKeepsChunkText` | `{title}` substitution does not suppress the append |
| `UnresolvedTextPlaceholderKeepsChunkText` | `{text}` against a chunk with no `text` field still delivers the body |
| `SubstitutesPlaceholders` | `{text}` resolved → suppressed → body appears exactly once |
| `PlaceholderChunksAlias` | `{chunks}` alias resolved → body appears exactly once |
| `AppendsChunkTextWhenNoPlaceholder` | static prompt still appends the chunk body |
| `ResumeTemplateChunksPath` | `{TitleChunker:FlatMiceFix@chunks}` does not trigger suppression; chunk body reaches the LLM |
| `SystemPromptPlaceholderSuppressesAppend` | `{text}` in systemPrompt suppresses the append (body in system msg, not duplicated) |
| `FieldValueContainsPlaceholderSubstring` | field value containing `{text}` substring does not fool suppression |

All `internal/ingestion/component` tests pass via `build.sh --test`.

### DIFF
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -700,16 +700,14 @@ func (c *ExtractorComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map
 			callIn.systemPrompt = substituteChunkPlaceholders(in.systemPrompt, ck, text)
 			// buildExtractorMessages appends chunkText to the user message
 			// unconditionally. Suppress that append only when a content-bearing
-			// placeholder ({text}/{chunks}/{content_with_weight}) was actually
-			// replaced — i.e. the chunk body is already in the prompt.
+			// placeholder was actually replaced — i.e. the chunk body is
+			// already in the prompt.
 			// "Present before substitution, gone after" is more precise than
-			// substring-matching the original prompt: it correctly handles a
-			// {text} that cannot be resolved against the chunk (still appends)
-			// and avoids suppressing on non-content placeholders like {title}
-			// (whose replacement must not drop the document body).
-			contentPH := []string{"{text}", "{chunks}", "{content_with_weight}"}
+			// substring-matching the original prompt: it avoids suppressing
+			// on non-content placeholders like {title} (whose replacement
+			// must not drop the document body).
 			callChunkText := text
-			for _, ph := range contentPH {
+			for _, ph := range contentPlaceholders {
 				if (strings.Contains(in.prompt, ph) && !strings.Contains(callIn.prompt, ph)) ||
 					(strings.Contains(in.systemPrompt, ph) && !strings.Contains(callIn.systemPrompt, ph)) {
 					callChunkText = ""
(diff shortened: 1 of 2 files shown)


xugangqiang:
LGTM

PR #18035, opened 2026-08-10
Fix agentic chat not working in chatbot

wangq8:


wangq8 (commented):
## PR Review: Fix agentic chat not working in chatbot

### Summary
This PR fixes a bug where the chatbot iframe endpoint (`async_iframe_completion`) was calling `async_chat` directly, bypassing the agentic RAG flow entirely. When a dialog has agentic RAG (reasoning) enabled, it was incorrectly routed to the standard chat path.

### Analysis

The fix replaces `async_chat` with `rag_agent` in `async_iframe_completion`. This is correct because:

1. **`rag_agent` is a proper superset of `async_chat`**: When `reasoning == 0` (no agentic mode), `rag_agent` falls back to `async_chat` at line 1879 — identical behavior for non-agentic chatbots, no regression.

2. **Consistent with `chat_api.py`**: The main chat API (`api/apps/restful_apis/chat_api.py` lines 1333, 1370, 1390) already uses `rag_agent`, so this aligns the iframe/chabot path with the primary chat path.

3. **Import change is correct**: `rag_agent` is added alongside `async_chat` (which is still used by another function in the file at lines 293/303), so no dead imports.

### Verification

| Check | Status |
|---|---|
| Regression for non-agentic chatbots | ✅ None — `rag_agent` delegates to `async_chat` when `reasoning==0` |
| Consistency with other endpoints | ✅ Matches `chat_api.py` behavior |
| Import correctness | ✅ Both `async_chat` and `rag_agent` are still needed |
| Security concerns | ✅ No new risks introduced |

### Minor Note
`api/apps/restful_apis/openai_api.py` (lines 319, 336) also uses `async_chat` directly — it may need the same treatment if the OpenAI-compatible endpoint should support agentic RAG, but that is out of scope for this fix.

**Verdict: LGTM ✅** — clean, targeted fix with no side effects.

PR #18037, opened 2026-08-10
fix(chat): preserve in-flight SSE stream when switching conversations

dcc123456:
### Summary

fix(chat): preserve in-flight SSE stream when switching conversations

Switching conversations no longer aborts the in-flight completion request, which could cancel server-side persistence of the just-asked question. The old SSE stream finishes in the background instead.

- SingleChatBox ignores backend conversation.messages sync while a stream is active for that conversation, so local streaming state isn't wiped.
- useSendMessage keeps a per-conversation message cache, restores it on switch, and routes streamed answers to the conversation they belong to (via answer.conversationId) rather than the currently displayed one.

### DIFF
--- a/web/src/pages/next-chats/chat/chat-box/single-chat-box.tsx
+++ b/web/src/pages/next-chats/chat/chat-box/single-chat-box.tsx
@@ -42,6 +42,7 @@ export function SingleChatBox({
     handleUploadFile,
     removeFile,
     setDerivedMessages,
+    activeStreamsRef,
   } = useSendMessage(controller);
   const { data: userInfo } = useFetchUserInfo();
   const { data: currentDialog } = useFetchChat();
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Nice work on this PR — keeping SSE streams alive with per-conversation caching is a clean approach.

Observations (non-blocking):

1. messagesCacheRef grows unbounded for the tab lifetime. Consider an LRU cap (20-30 conversations) for long sessions to avoid memory pressure.

2. appendAnswerToMessages duplicates the message-building logic from addNewestAnswer in logic-hooks.ts. Extracting a shared helper would prevent drift if fields change upstream.

3. appendAnswerToMessages assumes the last message is always the assistant placeholder — safe today given addNewestQuestion pattern, but worth a comment for future readers.

Edge cases around messagesConversationIdRef ordering and targetConversationId routing are correctly handled. LGTM overall.

PR #18038, opened 2026-08-10
fix(ingestion): split Extractor call() into callRaw/callText/callStructured

jay77721:
## Summary

`call()` conflated three concerns — dispatch, cleanup, and JSON parsing — into one function returning `any`. That forced every caller to guess the return type:

- **Field extraction** and **keywords/questions** used `result.(string)`, which **silently became `""`** when `call()` had parsed a JSON response into a map.
- **Metadata** re-parsed the string in a type switch.
- The Python-parity cleanup (think + tool_call) could not be added to the single `call()`, because stripping `<tool_call>` blocks would race the JSON parse metadata needs.

Python separates these: `LLMBundle.async_chat` (llm_service.py:459-461) cleans in the LLM layer and returns a string; the caller decides whether to parse.

## Fix

Split into three functions:

| Function | Returns | Used by |
|---|---|---|
| `callRaw` | `*extractorChatResponse` | shared dispatch/resolve/retry/nil-check |
| `callText` | `string` (cleaned) | field extraction, empty-chunks fast path, keywords, questions — matches Python `_generate_async` raw text |
| `callStructured` | `map[string]any` | metadata — matches Python `gen_metadata`'s `json_repair.loads` |

## Behavior changes

- **Field extraction now always writes a string** — a JSON-looking LLM response is stored as text, not parsed (matching Python). Two tests asserting the old parse-to-map behavior are inverted.
- **`callRaw` adds a nil-response guard** (fixes the panic risk when a provider adapter returns `(nil, nil)`).
- **`cleanLLMText`** applies Python's two-step cleanup: reasoning content stripped only when a leading `<think>` has a closing `</think>`, plus `<tool_call>` block removal.

## Tests

Added `TestCleanLLMText` and `TestExtractorComponent_callStructured`; existing metadata tests already exercise `callStructured` returning a map. All `internal/ingestion/component` tests pass via `build.sh --test`.

### DIFF
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -1020,18 +1020,21 @@ var toolCallRE = regexp.MustCompile(`(?s)<tool_call>.*?</tool_call>`)
 // in LLMBundle.async_chat (llm_service.py:459-461) for every response,
 // before any parsing:
 //
-//  1. Reasoning content: take everything after the LAST `</think>`. Python's
-//     _remove_reasoning_content (llm_service.py:332-346) only strips when a
-//     leading `<think>` is present AND a closing `</think>` follows it;
-//     otherwise the text is returned unchanged. (This differs from
-//     cleanExtractionResult, which strips on a bare `</think>`.)
+//  1. Reasoning content: when the response BEGINS with `<think>`, take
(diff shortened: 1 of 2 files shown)


xugangqiang:
LGTM

PR #18041, opened 2026-08-10
fix: exclude disabled documents from dataset structures

buua436:

Dataset-level graph, mind map, and timeline rebuilds could include disabled documents because disabled document IDs were incorrectly converted from single-element tuples.

This PR:

- correctly extracts disabled document IDs;
- excludes disabled documents when creating dataset-level rebuild tasks;
- rebuilds dataset structures from document-scoped rows only;
- prevents stale rows from being left behind during large full rebuilds;
- avoids falling back to document-level results when a dataset-level structure exists but becomes empty after filtering.

### Type of change

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -607,6 +607,10 @@ def run_index(dataset_id: str, tenant_id: str, index_type: str):
         types=[],
         suffix=[],
     )
+    # Disabled documents must not participate in a dataset-level structure
+    # rebuild. Keep them out of the fan-out task itself; the merger also
+    # applies a database-backed filter as a defense in depth.
+    documents = [document for document in documents if str(document.get("status", "1")) != "0"]
     if not documents:
         return False, f"No documents in Dataset {dataset_id}"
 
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this PR! Three distinct but related fixes that together resolve the disabled-document pollution issue. Let me break down each:

**1. Tuple unpacking fix in `get_disabled_doc_ids_by_kb_id`** ⚠️ Critical

```python
# Before: doc_id is a tuple like ("123",)
{str(doc_id) for doc_id in ...tuples()}
# After: correctly unpacks the single-element tuple
{str(doc_id) for (doc_id,) in ...tuples()}
```

Peewee's `.tuples()` returns rows as tuples. Without unpacking, the set would contain `{("123",), ("456",)}` instead of `{"123", "456"}`. These tuple-strings would never match real doc_id strings, effectively making the disabled-doc exclusion a no-op. This is the root cause of the bug.

**2. Filter in `run_index`** ✅

Filtering out disabled documents (`status != "0"`) before fanning out tasks is correct defense-in-depth. Even though the merger also filters them, avoiding work at the scheduling layer is the right call.

**3. Merger changes in `_do_build`** ✅

- **Paging delete → single delete-by-condition:** The old paging approach was genuinely buggy — deleting rows while iterating with offset shifts subsequent pages, so only the first 1000 records would actually get deleted. The new single `_index_delete(tenant_id, kb_id, del_cond)` call fixes this properly.
- **Always scoping to `scope_kwd="doc"`:** This prevents the merger from consuming its own output (dataset-level rows) as input on the next run. Combined with the paging delete fix, full rebuilds should now be correct.

**One behavioral note on `_scope_for_template` fallback removal:**

The old code had a fallback: if dataset-level build returned empty, it would attempt a doc-scoped build. Removing this means that if a dataset-level merge genuinely produces no entities/relations (e.g., all documents have empty graphs), the result will stay empty rather than falling back to per-document results. With the merger now correctly scanning document-level rows, this should be fine, but worth noting as an intentional behavioral change.

**One potential edge case:**

The always-applied `scope_kwd` filter means existing doc_graph rows that predate the `scope_kwd` field (which won't have it set) will be invisible to full rebuilds. These are generated/derived data, so they'll be regenerated correctly — but if anyone has custom tooling that expects old rows to persist, they'd want to know. Not a blocker.

Overall solid bug fixes. LGTM with the above noted.

PR #18043, opened 2026-08-10
fix(share): The language of the shared page will cause the language of the original website to change.

dcc123456:
### Summary

fix(share): The language of the shared page will cause the language of the original website to change.

- Add `persist` option to `changeLanguageAsync` to skip writing language to localStorage
- Pass `persist: false` on share pages (chat/agent/search) and floating widget so the visitor's main site language setting isn't overwritten
- Fix `SelectWithSearch` using non-reactive `t` from `i18next` — switch to `useTranslation()` so dropdown placeholders and empty state update on language change
- Remove unnecessary `useMemo` wrapping on `thinkingOptions` in `NextMessageInput` — `t` ref is stable so memoization prevented label updates after `changeLanguage`
- Remove duplicate language init effect from `RootProvider` (already handled in `main.tsx` via `initLanguage`)

### DIFF
--- a/web/src/app.tsx
+++ b/web/src/app.tsx
@@ -1,6 +1,5 @@
 import { Toaster as Sonner } from '@/components/ui/sonner';
 import { Toaster } from '@/components/ui/toaster';
-import { changeLanguageAsync } from '@/locales/config';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { configResponsive } from 'ahooks';
 import dayjs from 'dayjs';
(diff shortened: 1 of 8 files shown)


wangq8 (commented):
Thanks for this fix! The root cause analysis is solid — shared pages overwriting the main site language preference is a real UX issue, and your `persist` option approach is clean and backward-compatible.

**What looks good:**

- `changeLanguageAsync` with `persist` option (default `true`) is a minimal, non-breaking API change. All callers that don't pass `persist` keep existing behavior.
- The `persist: false` additions on share pages (agent/share, chats/share, search/share, floating widget) correctly scope the fix to the affected surfaces.
- Removing the duplicate language init from `RootProvider` is correct — `initLanguage` in `main.tsx` already handles this, and having both creates a race.
- The `SelectWithSearch` fix switching from `import { t } from i18next` (module-level, non-reactive) to `useTranslation()` (hook, reactive) is exactly right. Without it, default placeholders and empty states would be stuck in the initial language.

**One note on `NextMessageInput`:**

- Removing `useMemo` from `thinkingOptions` is the right call since `t` from `useTranslation()` is reference-stable, so the memo was equivalent to `useMemo(() => [...], [])` — it would never update on language change.
- Now `thinkingOptions` is recreated on every render. The translated strings are the same-by-value but different-by-reference each time. If the selector component does reference-equality checks on the options array, it could trigger extra re-renders. In practice, this is a minor concern for a thinking-level dropdown — not a perf hotspot. Just something to be aware of.

Overall LGTM. Nice work!

PR #18046, opened 2026-08-10
fix: prevent empty-model warning modal from staying open after OK

dcc123456:
### Summary

fix: prevent empty-model warning modal from staying open after OK

useWarnEmptyModel is invoked from multiple hooks (useFetchTenantInfo via useSelectParserList, and useFetchDefaultModelDictionary), each with its own warnedRef. Two instances could both fire Modal.warning(), stacking two static modals; onOk only destroys the top one, so the bottom one stays open and looks like the dialog never closed.

Add a module-level isWarningVisible guard so only one warning modal exists at a time, and reset it on dismissal.

### DIFF
--- a/web/src/hooks/use-warn-empty-model.tsx
+++ b/web/src/hooks/use-warn-empty-model.tsx
@@ -5,6 +5,8 @@ import { useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigatePage } from './logic-hooks/navigate-hooks';
 
+let isWarningVisible = false;
+
 export const useWarnEmptyModel = (
   showEmptyModelWarn: boolean,
   embdId?: string,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice catch on the root cause and clean fix. The issue of two `Modal.warning()` instances stacking when multiple hooks share the same condition is real, and a module-level guard is the right approach since antd's static modal API doesn't provide a built-in dedup mechanism.

**What's good:**
- `isWarningVisible` guards against duplicate modals before creation.
- The flag is properly reset in `onOk` after dismissal.
- The guard is also checked in the condition alongside `warnedRef`, so if any hook clears its warn state and re-enters, the flag still prevents a double modal.

**Minor edge case (not blocking):**
If the component unmounts while the modal is still open (e.g., a route change without clicking OK), `isWarningVisible` would never be reset, preventing future warnings in the same session. In practice this is unlikely since `onOk` navigates to settings, but worth noting if you encounter a scenario where route guards or external navigation can dismiss the component without the modal's `onOk` firing.

Approved, LGTM.

PR #18047, opened 2026-08-10
fix(token-chunker-form): set minimum value for chunk token size slider

dcc123456:
### Summary

fix(token-chunker-form): set minimum value for chunk token size slider


wangq8 (commented):
Straightforward fix — chunk token size should never be 0 or negative, so adding `min={1}` is correct and consistent with the `overlapped_percent` slider below it which already has `min={0}`.

No regressions or side effects. Approved, LGTM.

PR #18050, opened 2026-08-10
fix(search-page): add overflow-x-auto and min-width to prevent layout issues

dcc123456:
### Summary

fix(search-page): add overflow-x-auto and min-width to prevent layout issues

### DIFF
--- a/web/src/pages/next-search/index.tsx
+++ b/web/src/pages/next-search/index.tsx
@@ -34,10 +34,10 @@ export default function SearchPage() {
 
   return (
     <section
-      className="size-full flex-1 relative px-5 pb-5 flex pt-4"
+      className="size-full flex-1 relative px-5 pb-5 flex pt-4 overflow-x-auto"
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix! The layout breaking on narrow search pages is a real UX issue and this addresses it cleanly.

**What's good:**
- `overflow-x-auto` on the outer section ensures a horizontal scrollbar appears when content overflows, rather than UI elements getting squashed or hidden.
- `min-w-[1280px]` on the inner flex container establishes a stable minimum layout width, which is a pragmatic approach for the search page.

The hardcoded `min-w-[1280px]` value could be brittle if the search page layout changes significantly in the future. Consider whether a slightly lower min-width (e.g., `min-w-[1024px]`) or a responsive breakpoint approach would still prevent the layout issue while being more flexible across screen sizes. No need to change now if 1280px is the tested safe minimum.

Approved, LGTM.

PR #18053, opened 2026-08-10
fix(floating-selection-toolbar): adjust dropdown direction based on available space

dcc123456:
### Summary

fix(floating-selection-toolbar): adjust dropdown direction based on available space


wangq8 (commented):
Thanks for the fix! Good improvement to the UX of the floating toolbar.

**What this does:** Adjusts dropdown direction (upward vs downward) based on available viewport space below the toolbar. Uses `useEffect` to recompute whenever a menu opens.

**Review:**
- ✅ `barRef` check prevents errors when the ref is not yet attached
- ✅ `useEffect` dependencies `[showHeadingMenu, showCodeLang]` are correct — recomputes direction only when menus open/close
- ✅ Heading menu height estimate (`6 * 26 + 12 = 168`) is reasonable for 6 items
- ✅ Code menu cap of `220` aligns with the existing `maxHeight: 200`
- ✅ Conditional `bottom: 100%` / `top: 100%` positioning via spread operator is clean
- ✅ Both heading dropdown (lines 283-286) and code language dropdown (lines 363-367) are updated consistently

**One small observation (non-blocking):** The heading menu height is hard-coded as an estimate. If items are ever added or styling changes, this may need updating. Consider extracting `6 * 26 + 12` as a named constant or reading the actual DOM height for robustness in a future iteration.

Approve ✅

PR #18054, opened 2026-08-10
Fix: set column type priority

Lynn-Inf:
### Summary

Judge by column type priority when counts tie.


wangq8 (commented):
Nice catch. The tie-breaking priority makes sense — preferring lossless types (text > datetime > float > int > bool) when counts are equal is the right call. The change is minimal and focused.

One minor observation (not blocking): the existing `x[1] * -1` negate pattern is non-standard Python — usually it's `-x[1]`. Not introduced by this PR though, so no need to fix here.

LGTM. ✅

PR #18055, opened 2026-08-10
fix: improve structure keyword search

buua436:

Improve keyword search for document and dataset structure graphs.

- Prefer BM25 text matching on entity search fields.
- Fall back to KNN semantic matching when BM25 has no valid result.
- Return multiple matched entities instead of only one.
- For `tree` and `page_index`, return matched nodes together with their parent path, without unrelated sibling or child nodes.
- Add logs to distinguish BM25 matches from KNN fallback results.

### Type of change

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -737,7 +737,7 @@ async def get_document_structure_graph(tenant_id, dataset_id, document_id):
             scope = {"doc_id": [document_id], "compile_kwd": [compile_kwd_val], "must_not": {"exists": "compilation_template_ids"}}
         return {"template_id": bucket_id, "template_name": bucket_name, "kind": bucket_kind}, scope
 
-    # ── keywords mode: global KNN → the top-1 entity's focused subgraph ──
+    # ── keywords mode: name matching/KNN → matched entities' subgraph ──
     if keywords:
         try:
             embd_id = DocumentService.get_embd_id(document_id)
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this improvement! The BM25-first approach with KNN fallback is a solid architectural improvement — exact name matching makes a lot more sense for keyword search than relying solely on vector similarity. A few observations:

### What looks good
- Preferring BM25 over KNN for keyword search is the right priority — entity names are identifiers, not prose.
- Returning multiple matched entities from the same bucket instead of only the top-1 is a clear UX improvement.
- The tree/page_index ancestor path walk is well thought out — leaf matches shouldn't be returned without their parent context.
- The logging additions to distinguish BM25 hits from KNN fallback are helpful for debugging.
- Raising the KNN similarity threshold from 0.0 to 0.3 and lowering topn from 256 to 16 makes sense for a keyword-targeted query.

### Questions / minor concerns

1. **Empty `text_query` edge case (line ~393):** If `keywords` contains only special characters (e.g. `"!!!"`), `re.sub(...).strip()` produces an empty string, and `MatchTextExpr` would be called with an empty query. Could this cause unexpected behavior in Elasticsearch, or should there be a guard that falls through to KNN immediately when `text_query` is empty?

2. **Substring matching in `_name_matches_query`:** The `"query in name"` check means a query like `"cat"` would match an entity named `"catalog"`. For single-word queries this could produce noisy results. Is this intentional? A word-boundary or token-based match might be more precise for single-word queries while keeping the current behaviour for multi-word queries.

3. **`detail_candidates` filtering:** When a keyword matches a title entity and no detail entities exist in the same bucket, the title match is preserved (which is correct). But when there are detail matches, the title is excluded from `matched_nodes`. In the tree/path-walk mode this is fine since the title will be pulled in as an ancestor — but for non-tree modes, could excluding the title entity cause a relevant match to be silently dropped?

None of these are blockers — the core logic is sound. Happy to approve once the empty-query edge case is addressed.

PR #18056, opened 2026-08-10
Fix add agent update validation check

wangq8:


wangq8 (commented):
## PR Review: Fix add agent update validation check (#18056)

### Summary
This PR fixes a gap where DSL validation was not performed during agent updates via the API. It extracts the parameter validation logic from `Canvas.load()` into a reusable `@staticmethod` and calls it from `update_agent` to catch invalid DSL configurations early.

### What's good ✅
- **Correct fix.** Previously, invalid DSL could be saved through the update endpoint without any validation. The new code catches `ValueError` from parameter validation just like it does for DSL normalization errors, returning an `ARGUMENT_ERROR` response.
- **Clean refactor.** Extracting `validate_component_parameters()` as a `@staticmethod` is the right design — it's side-effect-free, reusable, and doesn't require a fully constructed `Canvas` instance.
- **Dead code removal.** The unused `cpn_nms` set has been cleanly removed.
- **Correct ordering.** `normalize_dsl` is called before validation, ensuring a consistent structure before checking parameters.

### Suggestions / Questions

1. **Double validation (minor).** `validate_component_parameters()` is now called both in `update_agent` (API layer) and inside `Canvas.load()`. This works as defense-in-depth but means validation runs twice when an agent is later loaded. Not a bug, just noting the redundancy.

2. **In-body import.** The `from agent.canvas import Canvas` import is placed inside the `try` block in `update_agent`. If this is to avoid a circular import, it's fine — but please confirm. If there's no circular dependency, consider moving it to the top of the file for consistency and clarity.

3. **No test coverage.** There are no tests added for the new validation path in the `update_agent` API. Consider adding a test case that sends an update with an invalid DSL and asserts that an `ARGUMENT_ERROR` is returned.

### Verdict
**Approve with minor comments.** The fix is correct, the refactor is clean, and it meaningfully improves the robustness of the update endpoint. The suggestions above are non-blocking improvements.

PR #18062, opened 2026-08-10
Fixed: Updated the text color style of the copy button.

dcc123456:
### Summary

Fixed: Updated the text color style of the copy button.


wangq8 (commented):
Thanks for the fix! Simple but important styling correction.

**What this does:** Changes the copy button hover color from `hover:text-white` to `hover:text-text-primary`.

**Review:**
- ✅ `hover:text-white` would make the button text invisible on light themes (white text on light/white background)
- ✅ `hover:text-text-primary` uses the theme-aware primary text color, which is visible on both light and dark themes
- ✅ Change is minimal (1 line) and only affects the hover state

Approve ✅

PR #18063, opened 2026-08-10
Fix: update tree when delete skill node

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for this fix — it addresses a real gap where deleting a skill node left the `skill_all` tree in an inconsistent state.

**What works well:**
- The recursive tree walk (`_walk_skill_tree`) correctly locates the target node and its parent in the `skill_all` tree.
- Descendant collection (`_collect_tree_descendants`) properly handles the nested dict structure of the aggregate tree.
- Tree pruning (`_prune_skill_tree`) correctly removes deleted nodes from the aggregate tree and rewrites it.
- The fallback path (direct delete when the node isn not found in the tree) preserves backward compatibility.
- `_parse_list_field` handles the Infinity `###`-joined string representation alongside native lists — good defensive coding.
- On the generator side, adding `parent_kwd` to skill rows is the right approach for the delete-side parent lookup.

The `_prune_skill_tree` function mutates `node["children_kwd"]` in-place. While this is safe here (the tree is a fresh `json.loads()` deserialization), a comment noting that the caller should not reuse the original tree after pruning would help future readers. Alternatively, a shallow-copy approach would make the lack of side effects more obvious:
```python
node["children_kwd"] = _prune_skill_tree(children, deleted_kwds)
# could also deep-copy `node` first if the original tree must be preserved
```

Overall LGTM. The fix is thorough and handles edge cases (missing node, parent-less root deletion, empty tree) correctly.

PR #18066, opened 2026-08-10
fix(mcp): stop dropping headers whose value is made of B/e/a/r letters

noron12234:
## What is wrong

`MCPToolCallSession._mcp_server_loop` decides whether to forward each configured
header:

```python
if nh.strip() and nv.strip().strip("Bearer"):
    headers[nh] = nv
```

This reads as "keep the header unless its value is just the word `Bearer`", but
`str.strip` takes a **set of characters**, not a prefix. The set here is
`{B, e, a, r}`, so any value built only from those letters collapses to the
empty string and the header is silently dropped:

| header value | kept today | should be kept |
|---|---|---|
| `Bearer` | no | no |
| `Bearer ` | no | no |
| `Bearer abc123` | yes | yes |
| `beer` | yes | yes |
| **`rare`** | **no** | yes |
| **`rae`** | **no** | yes |

The dropped header never reaches the MCP server, and nothing is logged — the
tool call just behaves as if the header had not been configured. Values like
`rare` or `rae` are plausible for a short environment/tenant tag substituted in
through `Template(...).safe_substitute(self._server_variables)`.

## Fix

Compare against the literal, which is what the condition was trying to say:

```python
if nh.strip() and nv.strip() and nv.strip() != "Bearer":
```

Behaviour is identical for every value except the ones in bold above, which are
now forwarded.

## Note

`ruff check` on this file reports 52 pre-existing findings on `main`; this change
adds none of them. I did not find an existing test module covering
`common/mcp_tool_call_conn.py` — happy to add one if you would like a home for it.


wangq8 (commented):
Reviewed — this is a correct fix. `str.strip("Bearer")` strips the character set `{B, e, a, r}`, not the literal prefix, so values like `rare`/`rae` collapsed to `""` and were silently dropped. The new condition `nv.strip() and nv.strip() != "Bearer"` matches the original intent and keeps behavior identical for every other value.

Minor nit (non-blocking): `nv.strip()` is evaluated three times; hoisting `nv = nv.strip()` once would be marginally cleaner. There's no test module covering `common/mcp_tool_call_conn.py`, so a small unit test for the header-forwarding condition would be a nice addition — but I won't block on it given the pre-existing state of the file.

LGTM.

noron12234:
Thanks @wangq8 — both notes are addressed.

- `nv.strip()` is now bound once as `token`. The value forwarded is still the unstripped `nv`, so behaviour is identical.
- `test/unit_test/common/test_mcp_tool_call_conn.py` exists on `main` now, so I rebased onto current `main` and put the coverage there: nine cases driven through the real `_mcp_server_loop` with the SSE transport stubbed, covering the values this fix is about (`"rare"`, `"rae"`), the ones that should still be dropped (`"Bearer"`, `"Bearer "`, blank), and a blank header name.

Verified locally against the pinned `mcp==1.28.1`: that module is 15 passed. Putting `.strip("Bearer")` back fails exactly the `rare` and `rae` cases and nothing else. `ruff check` on the new test file is clean, and the 52 pre-existing findings in the module are unchanged before and after.

PR #18069, opened 2026-08-11
fix: Remove unused constant files

dcc123456:
### Summary

fix: Remove unused constant files


wangq8 (commented):
LGTM. Confirmed that no file in the codebase imports from `web/src/pages/dataset/knowledge-graph/constant.ts` — the exported `graphData` is never referenced. The file contains hardcoded demo/mock data (Xiamen Xiangyu nodes and edges) that appears to be leftover from prototyping. Safe to remove.

PR #18072, opened 2026-08-11
fix(search): pass selected pageSize when retrieving chunks on document change

dcc123456:
### Summary

fix(search): pass selected pageSize when retrieving chunks on document change


wangq8 (commented):
Simple and correct fix. `handleTestChunk` accepts `(documentIds, page?, size?)` with defaults `page=1, size=10`, so passing it directly as `onTesting={handleTestChunk}` meant the user-selected `pageSize` was always ignored and the default of 10 was used instead. Wrapping in the arrow function to forward `pageSize` resolves this.

LGTM, approved.

PR #18073, opened 2026-08-11
Fix(mcp): reset bulk selection after delete and mode switch

euvre:
### Summary

Fix stale bulk selection state on the MCP server settings page (User Settings -> MCP).

Previously, the `selectedList` state in `useBulkOperateMCP` was never reset:

1. After a successful bulk delete, the list refreshed via react-query invalidation, but the deleted IDs remained in `selectedList`, so the "Selected N" count still showed the old number.
2. Toggling "Bulk Manage" / "Exit Bulk Manage" only flipped `isSelectionMode` without clearing the selection, so re-entering bulk manage mode showed a stale "Selected N" count even with nothing checked.

Changes:

- `use-bulk-operate-mcp.tsx`: `handleDelete` now awaits the delete mutation and clears the selection on success (`code === 0`); a new `resetSelection` helper is exposed.
- `index.tsx`: `switchSelectionMode` calls `resetSelection` so entering/exiting bulk manage mode always starts with an empty selection.

### DIFF
--- a/web/src/pages/user-setting/mcp/use-bulk-operate-mcp.tsx
+++ b/web/src/pages/user-setting/mcp/use-bulk-operate-mcp.tsx
@@ -11,9 +11,16 @@ export function useBulkOperateMCP(mcpList: IMcpServer[]) {
   const { deleteMcpServer } = useDeleteMcpServer();
   const { handleExportMcpJson } = useExportMcp();
 
-  const handleDelete = useCallback(() => {
-    deleteMcpServer(selectedList);
-  }, [deleteMcpServer, selectedList]);
+  const resetSelection = useCallback(() => {
+    setSelectedList([]);
+  }, []);
+
+  const handleDelete = useCallback(async () => {
+    const ret = await deleteMcpServer(selectedList);
+    if (ret.code === 0) {
+      resetSelection();
+    }
+  }, [deleteMcpServer, selectedList, resetSelection]);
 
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Clean fix, LGTM. The two state-reset points (post-delete + mode toggle) cover the bug completely.

One small note: `handleDelete` changes from sync to async. The call site in `index.tsx` uses it as an `onClick` handler, which doesn't care about the return type, so this is safe. The `code === 0` guard correctly prevents clearing the selection on failed deletes. 

Approved.

web/src/pages/user-setting/mcp/use-bulk-operate-mcp.tsx
```diff
@@ -11,9 +11,16 @@ export function useBulkOperateMCP(mcpList: IMcpServer[]) {
...
-    deleteMcpServer(selectedList);
-  }, [deleteMcpServer, selectedList]);
+  const resetSelection = useCallback(() => {
+    setSelectedList([]);
+  }, []);
+
+  const handleDelete = useCallback(async () => {
+    const ret = await deleteMcpServer(selectedList);
+    if (ret.code === 0) {
+      resetSelection();
```
dcc123456: The deletion action should simply filter the specific MCP out of the selected models; it should not reset the entire selection of models just because one MCP is removed.

PR #18074, opened 2026-08-11
Fix: restrict xxx_ids to limit 100

wangq8:


wangq8 (commented):
## PR Review: #18074 — Restrict xxx_ids to limit 100

### Summary
Adds a new `validate_rest_api_ids` utility that caps ID-list parameters at 100 elements, applied consistently across 9 API endpoint files. Also fixes two docstring `page_size` examples from 200 → 100.

### ✅ Strengths
- **Consistent approach** — single utility function (`validate_rest_api_ids`) used everywhere, mirroring the existing `validate_rest_api_page_size` pattern.
- **Good error messages** — raises `ValueError` with descriptive `field_name`, caught by existing exception handlers that return `RetCode.ARGUMENT_ERROR`.
- **Defensive placement** — validation sits right after parsing, before any service/db calls.
- **Docstring fixes** — `list_wiki_pages` and `list_wiki_topics` docstrings corrected from 200 to 100 (actual cap was already 100 via `validate_rest_api_page_size`).

### ⚠️ Issues

#### 1. `ListDatasetReq` uses a different validation approach
```python
# validation_utils.py
ids: Annotated[list[str] | None, Field(default=None, max_length=REST_API_MAX_IDS)]
```
Pydantic `max_length` on a list field constrains the number of items, so this works functionally. But it raises `ValidationError` (caught by `validate_and_parse_request_args`), which is a **different exception type** from `ValueError` raised by `validate_rest_api_ids`. This is fine operationally, but creates inconsistency — if someone adds a try/except around this model usage expecting `ValueError`, they will be suprised.

**Suggestion:** Consider using a Pydantic `field_validator` that calls `validate_rest_api_ids` internally so all validation paths raise the same exception type.

#### 2. Missing `field_name` argument in `_get_docs_with_request`
```python
# document_api.py line ~938
validate_rest_api_ids(doc_ids)  # field_name defaults to "ids"
```
Every other call site passes a descriptive name (e.g., `"doc_ids"`, `"dataset_ids"`). Here the error will say `"ids must contain at most 100 IDs"` instead of something like `"doc_ids must contain..."`. The query parameter name is `"ids"` so it is technically accurate, but inconsistent with the rest of the PR.

**Suggestion:** Pass `"doc_ids"` or `"ids"` explicitly for clarity.

#### 3. `list_memory` validates `tenant_id` as an ID-list
```python
# memory_api.py
for field_name in ("tenant_id", "owner_ids", "ids"):
    values = [item.strip() for item in str(filter_params.get(field_name) or "").split(",") if item.strip()]
    validate_rest_api_ids(values, field_name)
```
`tenant_id` is typically a single value. Splitting it by comma and validating it as a list of IDs works (single value → length 1, always passes), but it is semantically misleading. If `tenant_id` is ever accidentally a comma-separated string via a bug elsewhere, it would silently be treated as multiple IDs rather than rejected.

**Suggestion:** Consider only validating `owner_ids` and `ids`, or handle `tenant_id` as a single-value field.

#### 4. Naming nit in `memory_api.py` `search_message`/`get_messages`
```python
validate_rest_api_ids(memory_ids, "memory_id")  # field_name singular vs variable plural
```
The variable is `memory_ids` (plural) but the field name passed is `"memory_id"` (singular). Not a bug, but the error message says `"memory_id must contain at most 100 IDs"` which reads slightly oddly (singular field name with "IDs" in the message). Minor.

### Verdict
**Approve with suggestions.** The core logic is solid, the fix addresses a real concern (no cap on ID lists could cause performance issues or overly large queries), and error handling is proper. The issues above are all minor consistency/style points.

PR #18075, opened 2026-08-11
Fix: After creating a wiki template and deleting the model instance, the default model appears empty, but it can still be edited and saved successfully.

cike8899:
### Summary

Fix: After creating a wiki template and deleting the model instance, the default model appears empty, but it can still be edited and saved successfully.


wangq8 (commented):
Thanks for this PR! The approach is well-thought-out — handling the case where a persisted model reference has been deleted from the provider, rather than silently rendering an empty select, is the right UX.

A few observations:

**What works well:**
- `buildValidModelIds` correctly covers both `model_id` and legacy composite `modelName@instance@provider` IDs, so it works for both old and new persisted values.
- The `trigger` extraction in `ModelTreeSelectFormField` to avoid the infinite validation loop when `form` is recreated on every render is a subtle but important fix — good catch and good comment explaining it.
- Exposing `isFetched` and `isError` from `useFetchAllAddedModels` is necessary because `initialData: []` makes `data` ambiguous (empty list vs. not-yet-loaded).
- The `renderMissingValue` prop on `TreeSelect` keeps the component generic rather than coupling it to model-specific logic.

- In `layout-recognize-form-field.tsx` and `ModelTreeSelect`, the loading prop is set to `!modelsFetched || modelsError`, which means a permanent fetch error will show a loading spinner forever instead of an error state. Consider showing an error indicator when `modelsError` is true while `modelsFetched` is also true.
- The `useFetchAllAddedModels` in `ModelTreeSelectFormField` destructures only `loading` (isFetching) from the hook, which will re-trigger the validation `useEffect` on every background refetch. It’s functionally harmless (`trigger` is idempotent), but using `isFetched` instead might be slightly more precise.

Overall LGTM, nice work!

PR #18078, opened 2026-08-11
Fix: New sessions will automatically carry over messages from the previous session by default.

dcc123456:
### Summary

Fix: New sessions will automatically carry over messages from the previous session by default.

### DIFF
--- a/web/src/pages/next-chats/chat/index.tsx
+++ b/web/src/pages/next-chats/chat/index.tsx
@@ -53,6 +53,9 @@ export default function Chat() {
         if (!isEmpty(conversation)) {
           setCurrentConversation(conversation);
         }
+      } else {
+        // New session: clear previous session's messages so they don't leak in.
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix! The change correctly prevents message leakage from a previous session into a new one.

**What this does:** When URL params indicate a new session (`!conversationId || isNew`), `setCurrentConversation({} as IClientConversation)` clears the stale conversation state before the new session data loads.

**Review:**
- ✅ The `else` branch correctly handles the case where we are not loading an existing conversation
- ✅ `setCurrentConversation` is a React state setter with stable reference, so it does not need to be in the dependency array
- ✅ No side effects on the existing happy path (`conversationId && !isNew`)

**One suggestion (non-blocking):** Consider wrapping the `fetchConversation` callback in a `useCallback` like the `handleSessionClick` is, to maintain consistency. But this is not required for correctness.

Approve ✅

PR #18080, opened 2026-08-11
Fix max chunk_ids = 100

wangq8:


wangq8 (commented):
## PR Review: #18080 — Fix max chunk_ids = 100

### Summary
Frontend fix that caps `chunkIds` at 100 before sending to the API, and corrects the `size` parameter calculation to match the actual number of IDs being sent.

### ✅ Strengths
- **Simple, focused fix** — two targeted changes in one file.
- **Correct client-side truncation** — `chunkIds?.slice(0, 100)` prevents sending oversized requests that the server-side limit (from #18074) would reject.
- **Size fix is correct** — previously `Math.max(chunkIds.length, 100)` would set `size=100` even when fewer chunks were requested. Now `chunkIds.length` sends the exact count. This also avoids potential issues where the server returns more results than the client expects.

### ⚠️ Issue

#### 1. Magic number — hardcoded `100`
```typescript
// use-chunk-request.ts line ~179
size: chunkIds?.length ? chunkIds.length : Math.min(pagination.pageSize, 100),
```
The value `100` is hardcoded here. If the server-side limit (`REST_API_MAX_IDS` / `REST_API_MAX_PAGE_SIZE`) ever changes, this needs a separate frontend update — easy to miss.

**Suggestion:** Consider defining this as a shared constant (e.g., `export const MAX_CHUNK_IDS = 100` in a shared constants file) and using it both in the `slice` and `Math.min` calls. Bonus: import from a shared config if the backend exposes limits via an API.

### Before/After behavior check
| Scenario | Before | After | ✅ |
|---|---|---|---|
| 50 chunk IDs provided | `size = Math.max(50, 100)` = **100** ❌ | `size = 50` | Correct |
| 150 chunk IDs provided | `size = Math.max(150, 100)` = **150** ❌ | `slice(0,100)` + `size = 100` | Correct |
| No chunk IDs | `size = pagination.pageSize` (could be >100) | `size = Math.min(pageSize, 100)` | Correct |

### Verdict
**Approve.** The fix is correct and addresses the exact client-side issue. The magic number suggestion is a minor improvement for maintainability.

PR #18022, opened 2026-08-09
refactor(parser): text&code align + markdown golden meta-driven cleanup

xugangqiang:
## Consolidation

This PR folds in two sibling alignment efforts (each closed as superseded):

- **#18082 (markdown golden cleanup):** the markdown golden baseline no longer ships `gen_markdown_golden.py`; regeneration is reproducible from `meta` embedded in `markdown.python.golden.json`. Divergences are driven from `meta.accepted_divergences`. Sample and golden are English.
- **#18084 (HTML alignment):** the Go `HTMLParser` nested-table collapse fix, the `TestHTMLParser_*` alignment tests, and an English `html.python.golden.json` (a `{meta, items}` document with `meta.accepted_divergences: ["table"]`, no committed generator script). Divergences are meta-driven via `LoadGoldenDoc` / `AcceptedDivergences` / `FilterOutDocTypes`, mirroring the markdown/text&code tests.

Shared `align_test.go` keeps a backward-compatible `LoadGolden` and adds `LoadGoldenDoc` / `AcceptedDivergences` / `FilterOutDocTypes`, plus the `MarkdownAlignOptions` / `TextCodeAlignOptions` / `HTMLAlignOptions` presets.

---

## Summary

**text&code** — Aligns the Go `TextParser` (the `text&code` family: `.txt/.py/.js/.go/...`) with the Python flow `_code` path (`deepdoc.parser.txt_parser.TxtParser`).

- Go previously split only on blank lines (`\n\n`); now it splits on the flow parser's default delimiter set `\n!?;。；！？`, mirroring Python's `re.split(r"(%s)" % pattern, txt)`.
- `keep_delimiters=True` (the actual `_code` call) keeps each trailing delimiter attached, so sentence-ending punctuation survives the split.
- CRLF/CR normalized to LF like Python.
- The OVER_CAP token merge is intentionally **not** performed here — chunking ownership stays with the downstream Chunker (PARSER_ALIGNMENT_HANDOFF.md §2.3, decision 1). Item counts therefore differ from Python's merged chunks and are reconciled by the stitch-compare alignment test.
- The 8192-byte per-item cap is retained: the downstream chunker keeps any single oversized item whole and the embedding step truncates it, so without the cap a minified / no-newline file would collapse to one mega-chunk and lose most of its content.

[...]

PR #18084, opened 2026-08-11
fix(parser): align HTML Parser to Python (nested-table fix; alignment golden)

xugangqiang:
## Summary

Aligns the Go **HTML** parser (`internal/parser/parser/html_parser.go`) with the Python `RAGFlowHtmlParser` and adds a reusable alignment test harness so the two engines can be diffed content-wise, not just line-wise.

This fixes a real divergence — **nested-table 塌方** — where a `<table>` nested inside other block content (or containing another table) was collapsed into a single concatenated string on the Go side while Python kept each cell/tag on its own line.

### Root cause
`walkHTMLLeaf` recursed into `<table>` children as plain leaf text and concatenated every cell (`"NameAge\nAlice30…"`). Python keeps each cell/tag on its own line, so the engines diverged whenever a table appeared inside a list, blockquote, or another table.

### Fix (surgical; preserves per-block granularity)
- `htmlLeafText` / `walkHTMLLeaf` now take a `*[]map[string]any` sink for structured table items.
- `walkHTMLLeaf` gained a table-aware branch: a `<table>` is rendered to its inline markup, written as text, **and** appended as a structured `doc_type_kwd:"table"` item — mirroring the existing top-level (non-leaf) table path.
- `renderTableHTML` returns `""` on render error instead of recursing through the leaf extractor (avoids a render loop).

### Tests (TDD: RED before fix, GREEN after)
- `TestHTMLParser_NestedTableProducesStructuredItems` — the original 塌方 case.
- `TestHTMLParser_NestedListItemsPreserved` — nested `<ul>`/`<ol>` items stay separate, in document order; top-level list tagged `ck_type:"list"`.
- `TestHTMLParser_PrePreservesVerbatim` — `<pre>`/`<textarea>` keep indentation and internal newlines verbatim (one leading newline dropped per spec).
- `TestHTMLParser_AlignmentGolden` — content-equivalence against the Python `RAGFlowHtmlParser` baseline via the shared `align_test.go` tool.

### Alignment tooling (mirrors the merged markdown convention, PR #18014)
- `align_test.go`: `StripHTMLHeadingMarker` + `HTMLAlignOptions` preset.
- `testdata/html.python.golden.json`: Python baseline as a plain JSON list of `{"text", "doc_type_kwd"}` records, regenerated by `testdata/gen_html_golden.py`.
- `testdata/html.sample.html`: shared input sample.
- `testdata/gen_html_golden.py`: drives the real Python `RAGFlowHtmlParser`.

No chunker/wiring change was needed: the HTML parser already emits `OutputFormat:"json"`, and the chunker's table path preserves `ck_type`, so the structured table item flows through unchanged.

## Test plan
- `bash build.sh --test ./internal/parser/parser/...` → `ok` (full package, ~10s), including the four new tests.

## Base / stacking
Based on `infiniflow:main`. The `align_test.go` framework it reuses (`HTMLAlignOptions`, `LoadGolden` list contract) was introduced by #18014, which is already merged into `main`, so this PR builds directly on `main`.

## Follow-up
The branch name includes `epub`; **EPUB parser alignment is a separate follow-up** and is intentionally not part of this PR.

### DIFF
--- a/internal/parser/parser/align_test.go
+++ b/internal/parser/parser/align_test.go
@@ -235,3 +235,50 @@ func MarkdownAlignOptions(delimiter string) AlignOptions {
 // DefaultMarkdownDelimiter is the flow parser's default markdown delimiter
 // set, used when generating/loading the golden baseline.
 const DefaultMarkdownDelimiter = "\n!?;。；！？"
+
+// htmlHeadingMarkerRE matches a leading ATX heading marker so Python's
+// "# Title" (deepdoc merge_block_text prefixes h1–h6 with "# ") can be
+// normalized to Go's clean heading text.
+var htmlHeadingMarkerRE = regexp.MustCompile(`(?m)^#{1,6}\s+`)
+
+// StripHTMLHeadingMarker returns a Normalizer that removes a leading ATX
+// heading marker ("#"/"##"/…) from a line. Python's HTML flow parser
+// (deepdoc parser.py merge_block_text) prefixes h1–h6 sections with "# ",
+// while the Go HTML parser emits clean heading text. This is a representation
+// difference, not a content divergence, so it is stripped before comparing.
+// It must run before CollapseWhitespace because the marker relies on the line
+// start.
+func StripHTMLHeadingMarker() Normalizer {
+	return func(s string) string {
+		return htmlHeadingMarkerRE.ReplaceAllString(s, "")
+	}
+}
+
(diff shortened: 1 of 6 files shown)


xugangqiang:
Superseded: the HTML alignment work (nested-table fix + alignment tests + English {meta,items} golden) is now folded into #18022 (the parser-alignment consolidation PR), which already contains it on top of the text&code + markdown cleanup.

PR #18085, opened 2026-08-11
Fix: Click the blank area to prevent UpdateLogSheet from closing.

cike8899:
### Summary

Fix: Click the blank area to prevent UpdateLogSheet from closing.


wangq8 (commented):
Thanks for this PR! The fix correctly prevents the `UpdateLogSheet` from closing when clicking outside the sheet area, which makes sense since the sheet auto-closes when generation is complete and accidental dismissal during an ongoing operation would be disruptive.

**Code review:**

- ✅ Correct use of `onInteractOutside` from Radix UI to prevent outside-click dismissal
- ✅ Clean, minimal change — only the necessary lines
- ✅ Properly typed event handler
- ✅ Good UX: the sheet already auto-closes on `GenerateStatus.Completed`, so preventing manual outside dismiss avoids accidental closures

**Minor nit:**

- The function name has a typo: `preventOutsideDismiss` → should be `preventOutsideDismiss` (missing the "i" in "Dismiss")

This is not a blocker, but fixing the typo would be nice for consistency. Otherwise LGTM!

PR #18087, opened 2026-08-11
Fix: stabilize agent log handleSearch and preserve page_size on reset

dcc123456:
### Summary

Fix: stabilize agent log handleSearch and preserve page_size on reset

Wrap handleSearch in useCallback with explicit deps so the effect that re-runs search on pagination/sort change stays stable. Keep page_size when resetting filters instead of clobbering it back to the initial default, and align TableHeader background with the bg-bg-title token.

### DIFF
--- a/web/src/pages/agents/agent-log-page.tsx
+++ b/web/src/pages/agents/agent-log-page.tsx
@@ -20,7 +20,7 @@ import {
 } from '@/interfaces/database/agent';
 import { IReferenceObject } from '@/interfaces/database/chat';
 import { formatDate } from '@/utils/date';
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router';
 import { DateRange } from '../../components/originui/calendar/index';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this PR. The intent is good but I have a concern about the `useCallback` change that needs addressing before merging.

### ⚠️ Issue: unintended search on every keystroke

The original `useEffect` only fires on pagination/sort changes:
```
[searchParams.keywords, currentDate, sortConfig]
```

But now with `handleSearch` in the effect deps and `keywords` in the `useCallback` deps, **every keystroke** in the search input triggers an API call. The original `handleClickSearch`-based design explicitly waits for the user to click Search before firing. This is a functional regression.

**Suggested fix:** Use refs for values that shouldn't trigger re-searches:
```ts
const keywordsRef = useRef(keywords);
keywordsRef.current = keywords;
const currentDateRef = useRef(currentDate);
currentDateRef.current = currentDate;

const handleSearch = useCallback((overrides) => {
  setSearchParams((pre) => ({
    ...pre,
    from_date: currentDateRef.current.from as Date,
    to_date: currentDateRef.current.to as Date,
    page: pagination.current,
    page_size: pagination.pageSize,
    orderby: sortConfig?.orderby || '',
    desc: sortConfig?.desc as boolean,
    keywords: keywordsRef.current,
    ...overrides,
  }));
}, [pagination, sortConfig]);
```

This keeps the function stable wrt keywords/currentDate while still reading latest values.

### ✅ Good changes

- **page_size preserved on reset** — `{ ...init, page_size: pagination.pageSize }` correctly retains the user's chosen page size.
- **TableHeader bg-bg-title** — aligns with the design token, consistent.

### Minor note

The `useEffect` lint warning (missing `handleSearch` dep) can be safely suppressed with `// eslint-disable-next-line` when the design intentionally avoids reacting to those deps.

Please address the keystroke-triggered search issue before merging.

PR #18090, opened 2026-08-11
feat: rename agent log export button to export current page

dcc123456:
### Summary

feat: rename agent log export button to export current page

Change the export button text in the agent log page from the generic 'export' to 'export current page' to clarify that only the current page of logs is exported. Add new i18n key flow.exportCurrentPage to zh.ts and en.ts.

### DIFF
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -2476,6 +2476,7 @@ Example: Virtual Hosted Style`,
       pleaseUploadAtLeastOneFile: 'Please upload at least one file',
     },
     flow: {
+      exportCurrentPage: 'Export current page',
       preprocess: {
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
**Approved** ✅

Clear UX improvement — changing the button text from generic 'Export' to 'Export current page' removes ambiguity about what is actually being exported. The i18n keys are properly added to both `en.ts` and `zh.ts`, placed correctly under the `flow` namespace (the agent-flow section at ~line 2478).

Small nit: the new key is added at the top of the `flow` block in both locale files, which is fine but consider adding it closer to related keys (e.g., near `export`) for consistency if a `flow.export` key already exists elsewhere.

No regressions, no security issues. LGTM.

PR #18091, opened 2026-08-11
feat(component): add shared message-fit package

jay77721:
## Summary

Extract the `message_fit_in` trimming strategy (mirroring Python's `rag/prompts/generator.py`) into a shared Go package `internal/component/messagefit`.

## Why

The agent LLM component and the ingestion Extractor component both need to trim prompts to the model's context window before calling the provider. Each previously did (or would do) this with its own copy of the logic. This PR adds the shared primitive; follow-up PRs wire it into the agent LLM component (#18092) and the ingestion Extractor/tagger (#18095).

## Changes

- New package `internal/component/messagefit` with `Message` and `Fit`.
- `Fit` is non-mutating and returns the trimmed result explicitly:
  `Fit(msgs []Message, budget int) (kept []Message, keptIdx []int, count int)`.
  No empty-content sentinel is used — dropped entries are simply absent from
  `kept`/`keptIdx`, and a message kept but trimmed to empty (e.g. the system
  share collapses to 0 when the last message alone fills the budget) is still
  reported as kept, mirroring Python.
- `Fit` implements the same three-step strategy as Python's `message_fit_in`:
  1. everything fits → return as-is
  2. keep all system messages + last non-system message
  3. proportional trim (system-dominant vs otherwise), single-message trim
- Unit tests covering each branch (all-fit, middle-drop, no-system list,
  system-only, both Step-3 branches with token-budget assertions, the
  budget-filled-by-last boundary, single message, zero budget, empty input)
  plus a test that the input slice is never mutated.

## Testing

- `bash build.sh --test ./internal/component/messagefit/...` passes.
- Rebased onto latest `main` (`a0e091e75`).

## Note

This PR is the shared foundation and only delivers value once the consumer
PRs land: #18092 / #18093 (agent LLM fitting) and #18095 (ingestion
Extractor/tagger fitting), together with the shared model-context resolver
#18106. It is intentionally prepared first so the consumers can rebase onto
it.

### DIFF
--- a/internal/component/messagefit/messagefit.go
+++ b/internal/component/messagefit/messagefit.go
@@ -1,0 +1,143 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (License); you may not
+//  use this file except in compliance with the License. You may obtain a
+//  copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+//
+
+// Package messagefit trims a message list so its total token count fits
+// within a budget. It mirrors Python's rag/prompts/generator.py:message_fit_in.
+//
+// The package is shared by the agent canvas LLM component and the ingestion
+// Extractor component. Callers convert their message type to []Message,
+// call Fit, then write the results back.
+package messagefit
+
+import (
+	"ragflow/internal/tokenizer"
+)
+
+// Message is the minimal representation the fitter needs. Both the agent's
+// schema.Message and ingestion's eschema.Message convert to this.
+type Message struct {
+	// Role is "system", "user", "assistant", etc. Only "system" receives
+	// special treatment during fitting.
+	Role string
+	// Content is the text content that may be truncated.
+	Content string
+}
+
+// Fit trims msgs so its total token count fits within maxTokens.
+//
+// Strategy (mirrors Python's message_fit_in):
+//  1. If everything fits, return as-is.
+//  2. Keep all system messages + the last non-system message, drop the
+//     rest; if that fits, return.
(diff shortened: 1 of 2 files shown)


jay77721:
Addressed all three CodeRabbit review comments in a follow-up commit (a2c290b0), synced to all four stacked PR branches:

1. **Exact budget now counts as fitting** — Step 1 / Step 2 use `<= maxTokens`; added `TestFit_ExactBudget`.
2. **All retained system messages are trimmed** — the system share is spread across every retained system message via `trimSystems`, so the fitted total can no longer exceed the budget; added `TestFit_TrimsAllSystemMessages`.
3. **`TestFit_Step2_DropsMiddle` now exercises Step 2** — budget derived from the tokenizer counts of system + last, asserting the middle message is cleared and the retained messages are unchanged.

Verified: `bash build.sh --test ./internal/component/messagefit/...` plus the dependent agent/ingestion packages all pass.

jay77721:
Two follow-up commits on top of a2c290b0:

- **94baa6100**: when only system messages are retained (no non-system message), the whole budget is spread across every retained system message (`trimSystems`) — addresses the CodeRabbit system-only case. Also renamed the `Fit` parameter `maxTokens` → `budget` and documented that callers pass the model context window (`content_length`), not the generation cap, to avoid confusion with `max_tokens`. Adds `TestFit_SystemOnlyMessages`.
- **0be2661cb**: strengthened `TestFit_TrimsAllSystemMessages` / `TestFit_SystemOnlyMessages` to assert every retained system message keeps a positive token count (comparing string lengths let a clearing regression pass).

`bash build.sh --test ./internal/component/messagefit/...` passes.

PR #18092, opened 2026-08-11
refactor(agent): use shared message-fit in LLM component

jay77721:
## Summary

Switch the agent LLM component's `fitMessages` from the local map-based `messageFitInRaw`/`countAllTokens`/`stringContent`/`setContent` helpers to the new shared `internal/component/messagefit` package (#18091).

## Why

Removes a duplicate copy of the `message_fit_in` strategy now that the shared package exists; the ingestion Extractor uses the same package in #18095.

## Changes

- `fitMessages` converts `schema.Message` → `messagefit.Message`, calls `messagefit.Fit`, then writes the trimmed text back to the right field (plain `Content` or the text part of `UserInputMultiContent`).
- Image-only multi-modal turns (no text part) are preserved; entries the fitter drops are removed.
- The synthetic system prompt passed to `fitMessages` is preserved when kept by the fitter.
- Fix: the fitted text is written back even when a kept message is trimmed to empty. A kept-but-emptied system prompt (the final user turn alone fills the budget) previously fell through the non-empty guard and the ORIGINAL untrimmed content was sent, blowing the token budget. `fitSource` now tracks whether the message originally carried text in `Content`.
- Fitting budget still comes from the canvas `max_tokens` parameter — behavior unchanged (changed in #18093).

## Testing

- `bash build.sh --test ./internal/agent/component/...` passes.
- New unit tests: everything-fits, image-only turn preserved, synthetic system prompt preserved, middle message dropped over budget, multi-part text folding, kept-but-emptied system write-back (`TestFitMessages_SystemKeptButEmptied`).
- Rebased onto latest `main` (`7b2d052f8`) with #18091 and #18106 merged; no carried copies of `messagefit` remain.
- Adapted `fitMessages` to the non-mutating `messagefit.Fit` API (`kept`, `keptIdx`, `count`): reconstruct from `keptIdx` instead of filtering `Content == ""`.

Depends on #18091.

### DIFF
--- a/internal/agent/component/llm.go
+++ b/internal/agent/component/llm.go
@@ -1050,8 +1050,9 @@ func fitMessages(systemPrompt string, msgs []schema.Message, maxLength int) ([]s
 	// (image-only turns) carry an empty Content in messagefit and survive
 	// fitting when kept.
 	type fitSource struct {
-		copiedIdx int // index into copied; -1 for the synthetic system prompt
-		multiIdx  int // -1 means the text lives in Content
+		copiedIdx     int  // index into copied; -1 for the synthetic system prompt
+		multiIdx      int  // -1 means the text lives in Content
+		textInContent bool // the original message carried text in Content
 	}
 	all := make([]messagefit.Message, 0, 1+len(copied))
 	sources := make([]fitSource, 0, 1+len(copied))
@@ -1084,7 +1085,7 @@ func fitMessages(systemPrompt string, msgs []schema.Message, maxLength int) ([]s
 			}
 		}
 		all = append(all, messagefit.Message{Role: string(copied[i].Role), Content: text})
-		sources = append(sources, fitSource{copiedIdx: i, multiIdx: multiIdx})
+		sources = append(sources, fitSource{copiedIdx: i, multiIdx: multiIdx, textInContent: copied[i].Content != ""})
 	}
 
 	// Use 97% of effective context as the token budget.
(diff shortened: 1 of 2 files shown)


jay77721:
Addressed the CodeRabbit comments (pushed to this branch):

1. **llm.go:1080 — fit every text part** (Minor): `fitMessages` now folds every non-empty text part of a `UserInputMultiContent` message into the token budget, writes the fitted text back to the first text part, and drops any additional text parts. Added `TestFitMessages_FoldsMultipleTextParts`. (commit 30a3b7a95)

2. **messagefit_test.go:60 — Step 2 test** (Minor): rewritten in PR #18091 (a2c290b0, synced here) to use a tokenizer-derived budget that actually reaches Step 2, asserting the middle message is cleared and retained messages are unchanged.

3. **messagefit.go:56 — exact budget counts as fitting** (Major): fixed in PR #18091 (a2c290b0, synced here): both checks now use `<=`, with `TestFit_ExactBudget`.

4. **messagefit.go:113 — trim every retained system message** (Major): fixed in PR #18091 (a2c290b0, synced here) via `trimSystems`, with `TestFit_TrimsAllSystemMessages`.

5. **messagefit_test.go:97 — assert both system messages keep tokens** (Minor): strengthened `TestFit_TrimsAllSystemMessages` (and `TestFit_SystemOnlyMessages`) to assert a positive fitted token count for every retained system message instead of comparing string lengths. (0be2661cb, synced here as 982882a27)

All `bash build.sh --test ./internal/agent/component/... ./internal/component/messagefit/...` pass.

PR #18093, opened 2026-08-11
fix(agent): fit LLM prompts to model content_length

jay77721:
## Summary

Use the chat model's context window (`content_length`) as the message-fitting budget instead of the canvas `max_tokens` parameter.

## Why

`max_tokens` caps generation length only. Using it as the input-trimming budget collapsed the usable context for models whose context window is much larger than `max_output` — exactly the failure the fitting logic was added to avoid.

## Changes

- `internal/agent/component/llm.go`: fitting budget is now the resolved `content_length` (0 → 8192 fallback). Resolution goes through the shared `dao.ResolveModelContentLength` (#18106), and a resolution-failure debug log is emitted from `LLMComponent.Invoke` so an unexpected 8192 fallback is diagnosable.
- The canvas `max_tokens` is left untouched as the generation cap — it still reaches the invoker as `ChatInvokeRequest.MaxTokens` (output limit); it just no longer drives input trimming.
- `internal/agent/component/llm.go`: validation errors now point at the model `content_length` setting instead of `max_tokens`.
- No field renames: `MaxTokens` keeps its existing name; only the fitting-budget source changes.

## Testing

- `bash build.sh --test ./internal/agent/component/... ./internal/dao/...` passes.
- Regression tests added:
  - `TestLLM_Invoke_UsesModelContentLengthBudget`: a 40k-token prompt under `gpt-4o@openai` (128k context) is preserved instead of trimmed to the old 8192 default.
  - `TestLLM_Invoke_MaxTokensStillOutputCapAndNotBudget`: a small `max_tokens` still reaches the invoker as the generation cap but is NOT the fitting budget (the 40k prompt survives).
  - `TestLLM_Invoke_UnresolvableModelFallsBackTo8192`: an unknown model falls back to the 8192 budget and still trims.
- Rebased onto latest `main` (`eada2f6b6`). #18091, #18106, #18092, and #18095 are all merged; this branch no longer carries `messagefit`, `dao.ResolveModelContentLength`, or #18092's agent changes. `fitMessages` uses the non-mutating `messagefit.Fit` API already on `main`.

## Known limitations (follow-up)

`dao.ResolveModelContentLength` (from #18106) currently reads `content_length` from the provider catalog only; per-model `max_tokens` overrides in `tenant_model.extra` are not honored yet (Python resolves them via `model_extra.get("max_tokens") or 8192`). Models absent from the catalog therefore fall back to the 8192 fitting budget. A follow-up PR will honor the `extra["max_tokens"]` override — tracked alongside #18106.

### DIFF
--- a/internal/agent/component/llm.go
+++ b/internal/agent/component/llm.go
@@ -28,6 +28,7 @@ import (
 	"ragflow/internal/agent/runtime"
 	"ragflow/internal/common"
 	"ragflow/internal/component/messagefit"
+	"ragflow/internal/dao"
 	"ragflow/internal/entity/models"
 
 	"go.uber.org/zap"
@@ -313,7 +314,16 @@ func (c *LLMComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[strin
 	// Resolve the model's context window (content_length) for message
 	// fitting. 0 means the model is unknown → fitMessages falls back to
 	// 8192, matching Python's chat_mdl.max_length = model_config.get("max_tokens") or 8192.
-	contentLength := resolveModelContentLength(ctx, db, originalModelID, p.Driver, p.ModelID)
+	contentLength := dao.ResolveModelContentLength(ctx, db, originalModelID, p.Driver, p.ModelID)
+	if contentLength <= 0 {
+		// A 0 makes fitMessages fall back to the 8192 default budget, which can
+		// silently discard most of a large-context prompt, so surface the
+		// resolution failure for diagnosis.
+		common.Debug("llm: content_length not resolved, falling back to 8192",
+			zap.String("model_ref", originalModelID),
+			zap.String("driver", p.Driver),
+			zap.String("model_name", p.ModelID))
+	}
 	if p.ModelID == "" {
 		return nil, &ParamError{Field: "model_id", Reason: "required"}
 	}
--- a/internal/agent/component/llm.go
+++ b/internal/agent/component/llm.go
@@ -305,10 +305,15 @@ func (c *LLMComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[strin
 	// reference selected in the agent canvas is passed verbatim to the LLM
 	// driver, causing 400s for custom-added models.
 	var err error
+	originalModelID := p.ModelID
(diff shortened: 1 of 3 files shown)


jay77721:
Addressed the CodeRabbit comments (pushed to this branch):

1. **llm_credentials.go:89 — nil-DB early return skipped the catalog lookup** (Major): `resolveModelContentLength` no longer returns 0 when both the passed `db` and `dao.DB` are nil — only the tenant-model lookup is skipped and the provider-catalog lookup (step 2) still runs, since it needs no database. (b0e76a6ee)

2. **llm_credentials.go:107 — log resolution failures** (Major): the zero-return path now emits a debug log with `model_ref`/`driver`/`model_name` (no credentials), so silent degradation to the 8192 budget is diagnosable in production. (b0e76a6ee)

3. **messagefit.go:99 — system-only message lists** (Minor): when only system messages are retained (no non-system message), the whole budget is now spread across every retained system message via `trimSystems` (in PR #18091, 94baa6100, synced here); `TestFit_SystemOnlyMessages` added.

All `bash build.sh --test ./internal/agent/component/... ./internal/component/messagefit/...` pass.

jay77721:
@coderabbitai review

jay77721:
Updated to use the shared helper: `resolveModelContentLength` (private copy) is removed and the call site now uses `dao.ResolveModelContentLength` from #18106; the resolution-failure debug log moved to `LLMComponent.Invoke`. Depends on #18106.

internal/agent/component/llm.go
```diff
@@ -305,10 +306,24 @@ func (c *LLMComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[strin
...
 	}
+	// Resolve the model's context window (content_length) for message
+	// fitting. 0 means the model is unknown → fitMessages falls back to
+	// 8192, matching Python's chat_mdl.max_length = model_config.get("max_tokens") or 8192.
+	contentLength := dao.ResolveModelContentLength(ctx, db, originalModelID, p.Driver, p.ModelID)
+	if contentLength <= 0 {
+		// A 0 makes fitMessages fall back to the 8192 default budget, which can
+		// silently discard most of a large-context prompt, so surface the
+		// resolution failure for diagnosis.
+		common.Debug("llm: content_length not resolved, falling back to 8192",
```
yuzhichang: Use `Warn` here.

PR #18094, opened 2026-08-11
fix: show filtered session count in session list header

dcc123456:
### Summary

fix: show filtered session count in session list header

The session list header displayed the total session count regardless of any active filter, which was misleading when filtering was applied. Use filteredData.length instead of sessions.length so the count reflects the currently filtered results.


wangq8 (commented):
**Approved** ✅

Simple and correct fix. Using `filteredData.length` instead of `sessions.length` properly reflects the currently displayed session count when the search/filter is active. The `filteredData` variable is always defined from `useClientSearch`, so there's no risk of a runtime error.

No regressions, no style issues, no security concerns. Good to merge.

PR #18095, opened 2026-08-11
fix(ingestion): fit extractor and tagger prompts to model content_length

jay77721:
## Summary

Trim Extractor call prompts and the automatic tagger prompt to the chat model's context window (`content_length`) before sending, so oversized chunks or tag files are trimmed instead of rejected by the provider with a context-length error. Uses the shared `messagefit` package (#18091) and `dao.ResolveModelContentLength` (#18106).

## Changes

- `internal/ingestion/component/extractor.go`:
  - `fitExtractorMessages`: shared helper wrapping `messagefit.Fit` (#18091), adapted to the non-mutating API (`kept` + `keptIdx`); drops entries the fitter removed and returns a clear error instead of sending a conversation whose final user turn was trimmed to empty.
  - Guards: a retained system message must have non-empty content (a system prompt kept but trimmed to empty carries no extraction contract), and the final user turn must be non-empty. The guards only apply when the input actually had a system message — user-only prompts remain valid.
  - `extractorContextLength`: resolves `content_length` for the effective model through the shared `dao.ResolveModelContentLength` (#18106); the private `resolveExtractorModelContextLength` copy is removed. When `llm_id` is empty it resolves the tenant default chat model (mirroring `resolveExtractorChatTarget`), so the default-model path gets fitting too.
  - `extractorContextFitBudget`: 97% of the resolved context window (clamped to >= 1 so `messagefit` never falls into its 8192 default).
  - `callRaw` returns the fitting error directly instead of retrying an empty-user request.
- `internal/ingestion/component/extractor_tag.go`: `llmTagChunk` uses the shared helper; skips tagging (with a warning) when fitting fails.
- `dao.ResolveModelContentLength` (#18106, merged) resolves `content_length`; no copies are carried on this branch.
- Tests: fit guard (rejects empty user turn / keeps user turn / rejects emptied system prompt / user-only prompts without a system prompt keep working), extractorContextLength resolution paths (UUID, pinned/composite default, unknown, nil-DB), budget clamp, tagger message-fit + no-context-length skip, and an end-to-end `callRaw` test asserting the invoker receives fitted messages within the budget (`TestExtractorComponent_CallRaw_FitsBeforeInvoke`).

## Testing

- `bash build.sh --test ./internal/ingestion/component/... ./internal/dao/...` passes (full ingestion/component suite).
- Rebased onto latest `main` (`7b2d052f8`) with #18091 and #18106 merged; the carried `messagefit` and `dao.ResolveModelContentLength` copies are gone.

## Known limitations (follow-up)

`dao.ResolveModelContentLength` (from #18106) currently reads `content_length` from the provider catalog only; per-model `max_tokens` overrides in `tenant_model.extra` are not honored yet (Python resolves them via `model_extra.get("max_tokens") or 8192`). Models absent from the catalog therefore get no fitting here (extractor passes prompts through untrimmed). A follow-up PR will honor the `extra["max_tokens"]` override — tracked alongside #18106.

### DIFF
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -1486,16 +1486,26 @@ func fitExtractorMessages(ctx context.Context, db *gorm.DB, llmID string, msgs [
 	// field definitions); sending without it would silently produce
 	// garbage. The proportional trim can empty every system message when
 	// the final user turn alone fills the budget, so require at least one
-	// retained system message.
-	keptSystem := false
-	for _, m := range fitted {
+	// retained system message — but only when the input actually had one:
+	// systemPrompt is optional and a user-only prompt is a valid request.
+	hadSystem := false
+	for _, m := range msgs {
 		if m.Role == eschema.System {
-			keptSystem = true
+			hadSystem = true
 			break
 		}
 	}
-	if !keptSystem {
-		return nil, errors.New("extractor: message fitting emptied the system prompt; check the chat model context length setting or reduce the prompt size")
+	if hadSystem {
+		keptSystem := false
+		for _, m := range fitted {
+			if m.Role == eschema.System {
+				keptSystem = true
+				break
+			}
+		}
+		if !keptSystem {
+			return nil, errors.New("extractor: message fitting emptied the system prompt; check the chat model context length setting or reduce the prompt size")
+		}
 	}
 	last := fitted[len(fitted)-1]
 	if last.Role != eschema.User || strings.TrimSpace(last.Content) == "" {
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -80,10 +80,13 @@ import (
 
 	"ragflow/internal/agent/runtime"
 	"ragflow/internal/common"
+	"ragflow/internal/component/messagefit"
+	"ragflow/internal/dao"
 	"ragflow/internal/engine/redis"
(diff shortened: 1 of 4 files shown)


jay77721:
Two items addressed on this branch:

1. **CI failure — import cycle in test builds**: `extractor.go` called `service.ModelProviderService.ResolveModelContextLength`, but the `service` package's tests import `internal/ingestion/component` to register components, producing `import cycle not allowed in test` for `internal/service` and `internal/ingestion/pipeline`. `extractorContextLength` now resolves `content_length` through the DAO + provider catalog directly (`resolveExtractorModelContextLength`, mirroring `service.ResolveModelContextLength`) and the `service` import was dropped. Verified locally: `go test` compile of `internal/ingestion/component/...`, `internal/service/...`, and `internal/ingestion/pipeline/...` all pass. (44487b5e4)

2. **CodeRabbit messagefit comments** (Step 2 test, exact-budget `<=`, trim every retained system message) are all fixed in PR #18091 (a2c290b0) and synced here; the later test-assertion strengthening (0be2661cb) is also synced (ec61ea24c).

`bash build.sh --test ./internal/ingestion/component/...` passes.

jay77721:
Updated to use the shared helper: the private `resolveExtractorModelContextLength` copy is removed and `extractorContextLength` now calls `dao.ResolveModelContentLength` from #18106. Depends on #18106.

PR #18098, opened 2026-08-11
test(parser): add shared golden-doc + alignment helpers

xugangqiang:
## Summary
Centralize the shared golden-doc + alignment helpers in `align_test.go` so the
format-specific PRs (text&code, markdown golden, HTML) reuse one implementation
instead of each carrying their own copy of the scaffolding.

Adds (no production behavior change — consts/helpers only):
- `delimiter.go` (non-test): `DefaultTextCodeDelimiter` and
  `DefaultMarkdownDelimiter`, the single source of truth for the flow parser's
  default delimiter sets. Production parsers and the alignment tests both
  reference these, so there is no duplicated hard-coded copy that could drift.
- `GoldenDoc{Meta,Items}` and `parseGolden`: tolerant of both the legacy bare
  array and the new `{meta, items}` golden format.
- `LoadGoldenDoc`, `AcceptedDivergences(meta)`, `FilterOutDocTypes(items, drop)`:
  meta-driven divergence handling, no hardcoded divergence lists.
- `TextCodeAlignOptions`, `HTMLAlignOptions` / `StripHTMLHeadingMarker`:
  normalizer presets reused by the alignment tests.
- `LoadGolden` now tolerates the `{meta, items}` format.
- `TestParseGolden`: unit test locking the two tolerated golden formats (legacy
  array, `{meta, items}`) and the corrupt-input failure path, so future changes
  to the golden format cannot silently break parsing.

## Dependency
The `align_test.go` framework (CompareAlignment, Normalizer, LoadGolden) comes
from #18014, which is already in `main`. This PR only layers the shared helpers
and delimiter consts on top. Merge this first; PR-2/3/4 branch from it.

### DIFF
--- a/internal/parser/parser/align_test.go
+++ b/internal/parser/parser/align_test.go
@@ -22,6 +22,7 @@ package parser
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


yuzhichang:
LGTM

PR #18102, opened 2026-08-11
Fix: Retrieve the latest timeline data after cancelling the timeline task.

cike8899:
### Summary
Fix: Retrieve the latest timeline data after cancelling the timeline task.

### DIFF
--- a/web/src/pages/agents/hooks/use-select-filters.ts
+++ b/web/src/pages/agents/hooks/use-select-filters.ts
@@ -26,7 +26,7 @@ export function useSelectFilters() {
   const filters: FilterCollection[] = [
     {
       field: 'owner',
-      list: agentFilters.owner,
+      list: agentFilters?.owner,
       label: t('common.owner'),
     },
     {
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Good refactoring. The core idea — catching all run-end transitions (including cancel/pause) instead of only `Completed` — correctly addresses the bug described in the title.

**`useRunEndEffect` hook** — clean abstraction. Using a ref to track previous status and only firing on `Running → non-Running` transitions is the right approach.

**A few observations:**

1. **Unrelated change**: The optional chaining additions (`agentFilters?.owner`, etc.) in `use-select-filters.ts` are defensive fixes unrelated to the timeline cancel issue. Fine as a drive-by fix, but would be cleaner in a separate PR.

2. **UpdateLogSheet auto-close removed**: Previously the sheet auto-closed when status reached `Completed`. Now it stays open until the user manually closes it. Is this intentional? If the parent views are now responsible for closing the sheet, I don't see the corresponding `setUpdateSheetOpen(false)` call being added to any of the `handleRunEnd` callbacks. This means the sheet will stay open after the run ends in all cases (completion, cancel, failure). If that's the desired UX (so users can review the log after cancellation), that's fine — just want to confirm it's not an oversight.

3. **Sheet stays open on completion too**: Since the hook fires on *any* transition out of Running, the sheet now stays open even on successful completion (previously it auto-closed). This is a behavioral change for the happy path as well.

PR #18105, opened 2026-08-11
Fix: remove some deprecated model

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/conf/llm_factories.json
+++ b/conf/llm_factories.json
@@ -550,13 +550,6 @@
                     "is_tools": true
                 },
                 {
-                    "llm_name": "qwen3-tts-instruct-flash",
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
The removal itself looks clean — no Python backend code references any of these model names, so this is a pure config change.

Two notes:
1. **`qwen-omni-turbo` is multi-modal** (chat + speech2text + image2text), not just a TTS model like the others. The PR title says "remove TTS models" but this one is broader — worth double-checking it is indeed deprecated by the upstream provider.
2. Would be worth confirming whether any of these models are actively used in existing `llm_factories.json` deployments before removal, since users with saved configs referencing these models will get errors.

PR #18109, opened 2026-08-11
Fix: If a message contains multiple files, they need to be displayed with automatic line wrapping.

cike8899:
### Summary
Fix: If a message contains multiple files, they need to be displayed with automatic line wrapping.


wangq8 (commented):
LGTM. Simple and correct fix — adding `flex-wrap` to the file container ensures multiple files wrap properly instead of overflowing. One-line CSS change with no side effects.

PR #18110, opened 2026-08-11
Fix: xinference asr

Lynn-Inf:
### Summary

Fix xinference asr model call.

### DIFF
--- a/conf/llm_factories.json
+++ b/conf/llm_factories.json
@@ -550,13 +550,6 @@
                     "is_tools": true
                 },
                 {
-                    "llm_name": "qwen3-tts-instruct-flash",
-                    "model_type": [
-                        "tts"
-                    ],
-                    "is_tools": false
-                },
-                {
                     "llm_name": "kimi-k2.7-code",
                     "max_tokens": 262144,
                     "model_type": [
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
A few things to consider:

1. **Language change**: the default language changed from "zh" to "Chinese". Xinference's API documentation suggests the parameter should be a language code. For Whisper-based models, "zh" is the standard BCP-47 code; "Chinese" may not be recognized. Please double-check Xinference's accepted values — if they accept both, great; if only BCP-47 codes, this could break Chinese transcription.

2. **append_api_path vs ensure_v1 consistency**: the URL construction changed from  to . I verified  exists in url_utils. However,  is imported but no longer used — remove it from the import to keep imports clean.

3. **Authorization header**: adding the Bearer token header is a good fix — previously the key wasn't being sent with the request for Xinference ASR.

Minor cleanup: drop the now-unused  import.

wangq8 (commented):
A few things to consider:

1. **Language change**: the default language changed from "zh" to "Chinese". Xinference's Whisper-based API typically expects BCP-47 language codes. "zh" is standard; "Chinese" may not be recognized. Please double-check Xinference's accepted values -- if they accept both, great; if only BCP-47 codes, this could break Chinese transcription.

2. **append_api_path vs ensure_v1**: the URL construction changed from literal path to `append_api_path(self.base_url, "audio/transcriptions")`. This is fine, but note that `ensure_v1` is imported but now unused -- remove it from the imports to keep them clean.

3. **Authorization header**: adding the Bearer token header is a good fix. Previously the key wasn't being sent with the request for Xinference ASR, which is the core bug this PR fixes.

Minor cleanup: drop the now-unused `ensure_v1` from the import line.

PR #18111, opened 2026-08-11
Go: implement ?type=filter branch for GET /agents

euvre:
### Summary

The agents page filter bar requests `GET /api/v1/agents?type=filter` and expects `{filter: {owner, canvas_category}, total}`. The Go `ListAgents` handler ignored the `type` parameter and returned a plain canvas list (`{canvas, total}`), so `data.filter` resolved to `undefined` and the page crashed with `TypeError: can't access property "owner", e is undefined` ("Something went wrong").

Changes:

- **dao**: add `GetOwnerFilter` / `GetCategoryFilter` aggregations on `user_canvas` (JOIN `user` for owner nicknames, same team-permission visibility scope as the list query), and `CountSavedByTenant` on `compilation_template_group`
- **service**: add `ListAgentFilters`, mirroring the `type=filter` branch of Python `agent_api.list_agents` — owner/category aggregations with the caller's compilation template group count folded into their owner entry and appended as the `compilation_template_group` category; `total` is the sum of owner counts
- **handler**: dispatch the `type=filter` branch in `ListAgents`
- **web**: guard `data.filter` in `useFetchAgentFilters` so a missing filter payload degrades to empty filters instead of crashing the page

Tests: new `TestUserCanvasDAOOwnerAndCategoryFilters` covers the aggregation queries and team-permission visibility; `bash build.sh --test` passes for `internal/dao`, `internal/service`, and `internal/handler`.

### DIFF
--- a/internal/dao/compilation_template_group.go
+++ b/internal/dao/compilation_template_group.go
@@ -59,6 +59,17 @@ func (dao *CompilationTemplateGroupDAO) ListSaved(ctx context.Context, db *gorm.
 	return groups, nil
 }
 
+// CountSavedByTenant counts the tenant's own valid groups; built-in groups
+// (empty tenant_id) are excluded, mirroring the group_count query in Python
+// get_owner_filter / get_category_filter.
+func (dao *CompilationTemplateGroupDAO) CountSavedByTenant(ctx context.Context, db *gorm.DB, tenantID string) (int64, error) {
+	var count int64
+	err := db.WithContext(ctx).Model(&entity.CompilationTemplateGroup{}).
+		Where("tenant_id = ? AND status = ?", tenantID, string(entity.StatusValid)).
+		Count(&count).Error
+	return count, err
+}
+
 // GetSaved returns a single valid group for the tenant (or built-in), or nil.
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Nice work, thorough implementation. Summary of changes:

- dao: GetOwnerFilter / GetCategoryFilter aggregations on user_canvas with team-permission visibility + CountSavedByTenant on compilation_template_group
- service: ListAgentFilters mirroring the Python ?type=filter branch, with compilation template group count folded into owner entry
- handler: type=filter dispatch in ListAgents
- web: defensive guard on data.filter in useFetchAgentFilters

A few observations:

1. **GetOwnerFilter silently returns nil when ownerIDs is empty** — this is fine for the nil slice case, but consider whether the caller should be aware. In practice ownerIDs will always contain at least the userID, so this shouldn't fire, but worth noting.

2. **UserDAO.GetNicknameByID fallback**: in ListAgentFilters, when user has compilation template groups but no canvases, GetNicknameByID is called. If that errors, nickname falls back to empty string. This is acceptable but results in an owner filter entry with an empty label — intentional?

3. **Test coverage**: the test covers the aggregation logic and team-permission visibility well. Might be worth adding a test case for ListAgentFilters itself (service-level) that verifies the compilation template group count is correctly folded in.

Overall solid implementation. The web-side defensive guard is a good safety net.

PR #18112, opened 2026-08-11
fix: sanitize img tags in markdown via shared SafeImg component

dcc123456:
### Summary

fix: sanitize img tags in markdown via shared SafeImg component

Extract a shared SafeImg component (src/components/safe-img.tsx) and use it as the react-markdown `img` override in highlight-markdown, markdown-content, and next-markdown-content. The component whitelists src schemes (http(s)/relative/`data:image`) and explicitly forwards only src/alt/title, so event handlers (onerror/onload) that rehypeRaw may pass through cannot execute. Non-string or unsafe src (e.g. `javascript:`/`vbscript:`) renders the original `` tag as a literal string (React-escaped) so the markup stays visible without executing scripts. Deduplicates the same logic that was inlined across the three components.

### DIFF
--- a/web/src/components/highlight-markdown/index.tsx
+++ b/web/src/components/highlight-markdown/index.tsx
@@ -32,6 +32,7 @@ import { citationMarkerReg } from '@/utils/citation-utils';
 import { getDirAttribute } from '@/utils/text-direction';
 import { omit } from 'lodash';
 import { useIsDarkTheme } from '../theme-provider';
+import { SafeImg } from '@/components/safe-img';
 import styles from './index.module.less';
 
 const HighLightMarkdown = ({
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Good security fix. The SafeImg component properly:

- Whitelists src schemes (http(s), relative, data:image) via SAFE_SRC_REGEXP
- Explicitly picks only src/alt/title, so event handlers (onerror/onload) that rehypeRaw may pass through cannot execute
- Falls back to React-escaped literal string for unsafe/non-string src

One thing to double-check: markdown-content/index.tsx introduces a new call to sanitizeHtmlWithImagesAsText replacing DOMPurify.sanitize directly. I verified this function already exists in dom-util.ts and is used elsewhere (chunk-card components). This change isn't described in the PR summary — consider mentioning it in the description for clarity, since it changes the sanitization strategy for chunk content.

Otherwise looks solid.

PR #18117, opened 2026-08-11
fix(sandbox): update Tenki SDKs and drop removed project_id

camcalaquian:
### Summary

Brings both halves of the Tenki sandbox provider onto current SDKs and removes `project_id`, which Tenki deleted from its API.

**Go:** `github.com/LuxorLabs/tenki-sdk-go/sandbox` `v0.5.2` → `v0.7.0` (current latest).
**Python:** the provider's SDK was renamed on PyPI — `tenki-sandbox` is frozen at 0.4.0 and everything from 0.5 ships as [`tenki`](https://pypi.org/project/tenki/). The docs told operators to `pip install tenki-sandbox`, which installs a stale SDK that no longer matches this provider's expectations.

**`project_id` is gone.** Tenki removed project scoping from the sandbox API in 0.5.x: `Client.create()` no longer accepts `project_id`, so the current code path would raise `TypeError` against a current SDK. It was also marked `required: True` in the config schema, so the Admin > Sandbox Settings form asked for a value that no longer exists.

### Changes

**Go**
- `go.mod` / `go.sum`: sandbox module → `v0.7.0`. No call-site changes were needed — the only export removed between v0.5.2 and v0.7.0 is `WithCookieName`, which this repo does not use. All 14 SDK symbols `internal/agent/sandbox/tenki.go` references still exist.

**Python**
- `agent/sandbox/providers/tenki.py`: removed the `project_id` attribute, its `initialize()` read, the create kwarg, the config-schema entry, and the `validate_config` check. The import-error message now names the `tenki` distribution.
- `docs/.../sandbox_quickstart.md`: `pip install tenki`, and dropped the `project_id (required)` bullet.
- `test/unit_test/agent/sandbox/test_tenki_provider.py`: fixtures no longer set `project_id`; the create test now asserts it is *not* sent, and the schema test asserts the key is absent.

### Note on the import path

`import tenki_sandbox` is deliberately left unchanged. The `tenki` distribution still ships `tenki_sandbox` as its implementation package alongside a thin `tenki` alias; the alias is `__init__`-only, so `tenki.<submodule>` imports do not resolve through it. Only the *distribution* name needed to change, and `pip install tenki` provides the same `tenki_sandbox` module this provider already imports.

### Behaviour change worth flagging

Any existing Tenki sandbox config with `project_id` set will now ignore it, and the field disappears from the settings form. No migration is needed — the value was only ever forwarded to an API that no longer accepts it.

### Verification

- `GOOS=linux go build ./internal/agent/sandbox/` and `go build ./...`: **pass** on v0.7.0. (Building for Linux is necessary on any non-Linux host: `internal/agent/sandbox/local.go:307` uses `syscall.SysProcAttr.Pdeathsig`, which is Linux-only. That is pre-existing and unrelated to this PR.)
- `GOOS=linux go vet ./internal/agent/sandbox/`: clean.
- `GOOS=linux go test -c ./internal/agent/sandbox/`: the test binary compiles, so `tenki_test.go` type-checks against v0.7.0.
- `go mod verify`: all modules verified. The stale `v0.5.2` `go.sum` lines are removed.
- `pytest test/unit_test/agent/sandbox/test_tenki_provider.py`: **20 passed on main → 20 passed on this branch.**

Not run: the Go tests were compiled but not executed (they need a Linux host), and no live Tenki API call was exercised. The Python tests were run with a local stand-in for the `aliyun_codeinterpreter` sibling module, since `agent/sandbox/providers/__init__.py` imports every provider eagerly and Alibaba's `agentrun` SDK is not installable from public PyPI; that stand-in is local to my checkout and is not part of this PR.

### DIFF
--- a/agent/sandbox/providers/tenki.py
+++ b/agent/sandbox/providers/tenki.py
@@ -71,7 +71,6 @@ class TenkiProvider(SandboxProvider):
 
     def __init__(self):
         self.api_key = ""
-        self.project_id = ""
         self.base_url = ""
         self.image = ""
         self.allow_outbound = False
@@ -89,7 +88,6 @@ class TenkiProvider(SandboxProvider):
 
     def initialize(self, config: Dict[str, Any]) -> bool:
         self.api_key = str(config.get("api_key", "") or "").strip()
-        self.project_id = str(config.get("project_id", "") or "").strip()
         self.base_url = str(config.get("base_url", "") or "").strip()
         self.image = str(config.get("image", "") or "").strip()
         self.allow_outbound = bool(config.get("allow_outbound", False))
@@ -105,7 +103,6 @@ class TenkiProvider(SandboxProvider):
         is_valid, error_message = self.validate_config(
             {
                 "api_key": self.api_key,
-                "project_id": self.project_id,
                 "timeout": self.timeout,
                 "max_lifetime": self.max_lifetime,
                 "max_output_bytes": self.max_output_bytes,
@@ -131,7 +128,6 @@ class TenkiProvider(SandboxProvider):
         errors = self._tenki_errors()
 
         create_kwargs: dict[str, Any] = {
-            "project_id": self.project_id,
             "allow_outbound": self.allow_outbound,
             "max_duration": self.max_lifetime,
             "metadata": {"source": "ragflow"},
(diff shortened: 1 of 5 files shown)


JinHai-CN:
@camcalaquian

We're reaching out because we're interested in inviting you to join the RAGFlow Ecosystem Partner Program.

If you're open to learning more, could you please drop us a line at [haijin.chn@gmail.com](mailto:haijin.chn@gmail.com)? We'll reply with the formal invitation and all the details.

Looking forward to hearing from you!

PR #18121, opened 2026-08-11
fix: route TokenChunker delimiter_mode "one" to OneChunker in Go ingestion

euvre:
### Summary

The DSL contract shared by the web UI and the Python runtime expresses single-chunk mode as a `TokenChunker` node with `delimiter_mode: "one"`. In Go that behaviour lives in the separate `OneChunker` component, and `TokenChunkerParam.Validate` only accepts `token_size` / `delimiter`. As a result, a knowledge base configured with the **One** method failed at canvas compile time:

```
factory: TokenChunker: schema: field "delimiter_mode" has invalid value "one"
```

and the ingestor kept retrying the failed task indefinitely (the file appeared to be "parsing forever").

**Fix**: translate the contract at the single chunker factory entry point (`newChunkerByName` in `internal/ingestion/component/chunker/common.go`) — a `TokenChunker` component whose `delimiter_mode` is `"one"` now builds a `OneChunker`. This keeps one implementation path (no duplicated chunking logic), preserves the DSL node id (progress/log counting unaffected), and fixes both the ingestion pipeline and the agent canvas paths since they share the same registry factory. The strict constructor-level validation on `TokenChunker` is unchanged.

Adds a regression test (`TestNewChunkerByName_TokenChunkerOneMode`) pinning the translation. `bash build.sh --test ./internal/ingestion/component/chunker/...` passes.

### DIFF
--- a/internal/ingestion/component/chunker/common.go
+++ b/internal/ingestion/component/chunker/common.go
@@ -36,6 +36,12 @@ import (
 func newChunkerByName(name string, params map[string]any) (runtime.Component, error) {
 	switch name {
 	case ComponentNameTokenChunker:
+		// The DSL contract (shared by the web UI and the Python runtime)
+		// expresses single-chunk mode as TokenChunker delimiter_mode "one";
+		// in Go that behaviour lives in the OneChunker component.
+		if mode, _ := params["delimiter_mode"].(string); mode == "one" {
+			return NewOneChunker(params)
+		}
 		return NewTokenChunker(params)
 	case ComponentNameTitleChunker:
 		return NewTitleChunker(params)
--- a/internal/ingestion/component/chunker/one_test.go
+++ b/internal/ingestion/component/chunker/one_test.go
@@ -39,6 +39,20 @@ func oneChunksOf(t *testing.T, inputs map[string]any) []map[string]any {
 	return chunks
 }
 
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
Thanks for this fix! The root cause is well-identified and the fix at the `newChunkerByName` dispatch layer is clean — translating the DSL contract rather than relaxing schema validation is the right approach.

A few things to address:

**1. Unrelated `tenant.go` changes**
The `internal/service/tenant.go` changes (removing `factoryModelTypeName` and the factory model-type validation in `GetModelInfo`) are unrelated to the TokenChunker → OneChunker routing. These should be split into a separate PR with their own justification. Mixing fixes makes bisecting and reverting harder. If the factory catalog validation was blocking OpenAI-API-Compatible providers (as the comment suggests), that deserves its own change with appropriate context.

**2. `delimiter_mode` value casing**
The check `mode == "one"` is exact and case-sensitive. Is the web UI guaranteed to always emit lowercase `"one"`? If the DSL contract allows `"One"` or `"ONE"`, this would miss it. Consider `strings.EqualFold`.

Otherwise LGTM — the test pins the translation well, and the fix is minimal and focused.

wangq8:
@euvre resolve conflict.

PR #18123, opened 2026-08-11
Fix: A dataset-level tree can also view file-level data.

cike8899:
### Summary

Fix: A dataset-level tree can also view file-level data.


wangq8 (commented):
Overall a well-structured refactor. The split-click pattern (`expandOnRowClick={false}`) with separate `onClick`/`onExpand` callbacks is a clean approach for the multi-level tree UX. A few observations:

### Things I like
- **`useFetchDocumentStructureGraphById`** extracted cleanly; the original hook still works as a thin wrapper.
- **`isPlaceholderData`** from `keepPreviousData` is used correctly to avoid writing stale graph data into `structureMap` when switching between documents.
- **Deletion cleanup**: cascading `dropChildren` + `dropStructure` + clearing selection across root and child deletions is thorough.
- **Recursive `buildNavTreeData`**: cleaner than the old flat two-level approach and handles documents with structure entities naturally.

### Suggestions

**1. `buildFlatEntityItems` — falsy id filtering**
```ts
.filter((item) => item.id);
```
This would drop an entity whose id is the string `"0"`. Unlikely in practice, but using `.filter((item) => item.id != null)` would be more robust.

**2. Inline `handleChevronClick` in TreeNode** (minor)
Not a blocker, but defining `handleChevronClick` inside the render path (non-memoized) means a new function reference every render. For a tree component with many nodes, could consider `useCallback`. Low severity.

**3. Stale graph cache on re-expand**
The `structureMap` is never evicted (other than on delete). If a document gets re-parsed and its structure changes, the tree would show the old cached graph until a page refresh. Consider if this is a real concern — probably fine for now.

**4. `raptor` kind**
The string `"raptor"` is not in `CompilationTemplateKind`. If the backend returns this kind, you may want to add it to the enum for type safety, but handling it identically to `Tree` via `adaptTreeToTreeData` is the right fallback.

### Conclusion
The design is solid and the implementation is careful. The points above are minor; no blocking issues. ✅ Approved

PR #18124, opened 2026-08-11
fix: prevent stale loading state on previous agent chat reply

dcc123456:
### Summary

fix: prevent stale loading state on previous agent chat reply

`sendLoading` (= !done) is global — when a new conversation's stream starts it flips to true, and any assistant message matching `derivedMessages.length - 1 === i` lights up the loading indicator, including the previous conversation's reply.

- Gate the `loading` prop on `!message.content` so a reply that already has content never shows the loading state.
- Scope the `sendLoading` prop passed to MessageItem to the last message only, so LoadingDots (keyed on `sendLoading && isEmpty(content)`) doesn't render on historical replies either.

### DIFF
--- a/web/src/pages/agent/chat/box.tsx
+++ b/web/src/pages/agent/chat/box.tsx
@@ -107,7 +107,8 @@ function AgentChatBox() {
                   loading={
                     message.role === MessageType.Assistant &&
                     sendLoading &&
-                    derivedMessages.length - 1 === i
+                    derivedMessages.length - 1 === i &&
+                    !message.content
                   }
                   key={buildMessageUuidWithRole(message)}
                   nickname={userInfo.nickname}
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Clean fix, LGTM. The two changes together correctly prevent the stale loading indicator from appearing on previous conversation replies:

1. **`!message.content` gate on `loading`**: The previous reply already has content, so it should never show a loading state regardless of `sendLoading`.
2. **Scoping `sendLoading` to the last message**: Prevents `LoadingDots` (which renders based on `sendLoading && isEmpty(content)`) from showing on historical messages too.

The logic is tight and well-reasoned. No regressions — `derivedMessages.length - 1 === i` is already the guard used for the loading dots block on line 125. Edge case: an empty `derivedMessages` array would give `-1 === i` for both conditions, but that is a pre-existing concern, not introduced here.

✅ Approved

PR #18127, opened 2026-08-11
fix(web): align Go chunk method dropdown order with Python parser_ids

euvre:
### Summary

In Go backend mode, the chunk method dropdown in the knowledge-base creation dialog lists methods in alphabetical order (Audio, Book, Email, General, ...) because `GET /api/v1/pipelines?type=builtin` returns the registry sorted by id. The Python backend instead orders methods via `settings.PARSERS`, with commonly used methods listed first.

This PR sorts the Go pipeline catalog on the front end (`useSelectParserList`) to match the Python `parser_ids` order: general (naive), qa, resume, manual, table, paper, book, laws, presentation, picture, one, audio, email, tag. Go-only pipelines such as `knowledge_compiler` keep their API order at the end.

This is a display-order-only change; the Go API contract and its error messages are untouched.

### DIFF
--- a/web/src/hooks/use-user-setting-request.tsx
+++ b/web/src/hooks/use-user-setting-request.tsx
@@ -194,7 +194,36 @@ export const useSelectParserList = (): Array<{
           const translated = t(key);
           return translated !== key ? translated : title;
         };
-        return pipelineList.map((item) => ({
+        // Order the Go pipeline catalog the same way as the Python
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Clean approach to aligning the chunk method dropdown order with Python's `parser_ids`. The Map-based sort with fallback for unknown IDs is simple and maintainable. Since this is display-order only (no API contract changes), it's low risk. LGTM.

PR #18128, opened 2026-08-11
refactor(parser): align TextParser with Python _code delimiter split

xugangqiang:
## Summary
Make Go `TextParser` match Python `deepdoc TxtParser._code`:
- Split on the flow default delimiter set `\n!?;。；！？` (was blank-line only).
- `keep_delimiters=True`: each trailing delimiter is retained on its segment.
- Normalize CRLF/CR -> LF like Python.
- Do NOT perform the OVER_CAP token merge; chunking stays with the downstream
  Chunker (PARSER_ALIGNMENT_HANDOFF.md §2.3).
- The 8192-byte per-item cap is dropped: the parser is a pure delimiter
  splitter, matching Python's `parser_txt` (no size slicing). Sizing is
  delegated to the Chunker + embedding truncation — see contract #17799.

## Why dropping the 8192-byte cap is safe (contract #17799)
The parser layer must NOT do token-size splitting. `internal/ingestion/component/chunker/token.go`
(OVER-BUDGET UNITS, contract #17799) states: a single item exceeding
`chunk_token_size` is KEPT WHOLE as its own chunk and is NOT atom-split; the
downstream embedding/rerank layer truncates it later. Token-size chunking —
including any oversized-item handling — belongs to the Chunker (or a PreSplitter
stage between Parser and Chunker), explicitly NOT in the parser and NOT as a
char-window split. So removing the per-item byte cap converges Go with Python's
`parser_txt` (which also does no size slicing) instead of diverging, and the
Chunker already owns the oversized-item case.

## Test plan
- `TestTextParser_ParseWithResult_DefaultDelimiter` pins the new split rule
  (single-newline + sentence-delimiter splitting with delimiter retained).
- `TestTextParser_ParseWithResult_NoSizeCap` pins no per-item byte slicing.
- `TestTextParser_AdjacentDelimiters` pins the documented adjacent-delimiter
  behavior (Go drops the standalone second delimiter) so a future silent change
  in `splitCapturingDelims` is caught.
- `TestTextParser_AlignmentGolden` verifies content-equivalence vs the Python
  golden on a shared sample.

## Dependency
- Foundation helpers merged via #18098 (`fix/parser-align-foundation` → main).
- Stacks on #18014 (markdown golden framework in main).
- Rebased onto `main`; this PR is opened from the fork
  (`xugangqiang:fix/textcode-parser-align`) per repo workflow.

## Note
Adjacent-delimiter input (e.g. `"a..b"`) keeps only the merged segment in Go
while Python keeps the second delimiter as its own segment. The alignment test
normalizes this away; tracked as a known follow-up, not fixed here.


xugangqiang:
## Follow-up（非 blocker）：parser→chunker 端到端 golden

本 PR 已对齐 `TextParser` 的切分阶段（分隔符切分 + keep_delimiters + CRLF 归一化 + 去 8192 cap），并用 golden 钉死了 **parser 切分**与 Python `_code` 的内容等价。但端到端（parser 切分 → `TokenChunker` OVER_CAP 合并 → 最终 chunk）的 text&code parity 尚未在本 PR 内直接断言。

需要明确：这并非 text&code 独有空白——当前整个 Go↔Python 解析对齐（markdown / html / text&code）都只停在 parser 切分级 golden，没有任何 family 做了 parser→chunker 端到端 golden。链路本身已接通（`text&code` 经 `parser_dispatch.go:261` → `TokenChunker`，默认 `chunk_token_size:512`），合并阶段的 parity 也已被 `chunker/token_merge_parity_test.go` 与 `chunker/golden_parity_test.go` 覆盖（均为 chunker 级）。

建议后续（独立 PR，不扩大本 PR 范围）：仿 `TestTextParser_AlignmentGolden`（`parse_with_result_test.go`）把 `TextParser.ParseWithResult` 输出喂给 `TokenChunker.Invoke`，再用 Python `_code` 最终 chunks golden 断言，闭环 text&code 端到端 parity。

yuzhichang (approved):

PR #18135, opened 2026-08-11
fix: avoid quadratic dedup in RAG merge paths

### LINKED ISSUE #18025: Avoid Quadratic Deduplication in RAG Merge Paths
## Summary

Some RAG merge paths deduplicate IDs using list membership checks inside loops. This makes merge time grow quadratically as the number of chunks, documents, or structure rows increases.

Maintainers, please review whether the proposed approach below matches the expected data contracts and behavior. If the direction looks good, I can raise a PR with the focused fix and regression tests.

## Affected Paths

- `rag/advanced_rag/tree_structured_query_decomposition_retrieval.py`
  - `TreeStructuredQueryDecompositionRetrieval._async_update_chunk_info`
- `rag/svr/task_executor_refactor/dataset_structure_merger.py`
  - `_merge_bucket`
  - `_merge_relations`

These paths preserve ordered lists such as `chunks`, `doc_aggs`, `source_chunk_ids`, and `doc_ids_kwd`, but use repeated `item not in list` checks while appending. For large retrieval results or structure-graph buckets, this becomes O(n^2).

## Impact

Local microbenchmark, Python 3.13, median of 5 runs:

### Deep Research Merge

`existing=N`, `incoming=N`

| N | Current | Set-backed | Speedup |
|---|---:|---:|---:|
| 1,000 | 12.67 ms | 0.53 ms | 23.7x |
| 5,000 | 283.26 ms | 3.14 ms | 90.2x |
| 10,000 | 1197.40 ms | 7.17 ms | 167.0x |

### Dataset Structure Entity Bucket Merge

`rows=N`

| N | Current | Set-backed | Speedup |
|---|---:|---:|---:|
| 1,000 | 13.62 ms | 0.41 ms | 33.2x |
| 5,000 | 292.31 ms | 2.16 ms | 135.1x |
| 10,000 | 1203.36 ms | 4.05 ms | 297.5x |

A focused regression test also showed the current structure merge performs `19,900` equality comparisons for only 200 unique chunk IDs.

## ID Type and Regression Risk

The normal producers and API contracts treat these IDs as strings:

- `Document.id` is a `CharField`.
- `SearchResult.ids` is typed as `list[str]`.
- Retrieval chunks set `chunk_id` from search result IDs and `doc_id` from stored chunk metadata.
- Web/KG retrieval producers use `get_uuid()`, which returns a string.
- Structure helpers normalize source IDs as strings or drop non-strings.
- Frontend and Go contracts model `source_chunk_ids` as string arrays.

One caveat: storage maps some provenance fields, such as `source_chunk_ids` and `doc_ids_kwd`, as JSON. So malformed legacy rows could theoretically contain non-string or unhashable values. The implementation can either rely on the current string-ID contract or defensively filter non-string provenance entries before set membership checks.

## Proposed Fix

Keep the existing output lists to preserve first-seen order, but maintain companion sets for O(1) membership checks.

Example pattern:

```python
seen = set()
ordered = []

for item in items:
    if item not in seen:
        ordered.append(item)
        seen.add(item)
```

For provenance fields read from JSON-backed storage, defensive filtering may be preferable:

```python
seen = set()
ordered = []

for item in items:
    if not isinstance(item, str) or not item:
        continue
    if item not in seen:
        ordered.append(item)
        seen.add(item)
```

## Acceptance Criteria

- Preserve existing output order.
- Prevent duplicate chunk/doc IDs from being appended.
- Replace repeated list membership checks with set-backed membership.
- Add regression tests covering:
  - duplicate incoming retrieval results
  - existing IDs not being appended again
  - order preservation
  - structure bucket dedupe without quadratic comparison growth
  - handling of malformed non-string provenance values if defensive filtering is chosen

DeviousDrops:
## Summary
Fixes #18025. Both merge paths deduplicated IDs by scanning a plain list (`item not in list`) inside a loop while appending — O(n²) per merge. Replaced with a set-backed `seen` check alongside the existing ordered list: same order, same dedup result, O(n).

- `rag/advanced_rag/tree_structured_query_decomposition_retrieval.py`: `chunk_id`/`doc_id` merge in `_async_update_chunk_info`
- `rag/svr/task_executor_refactor/dataset_structure_merger.py`:
  - `_merge_bucket`: chunk/doc/description/original-name merge
  - `_merge_relations`: chunk/doc/description merge
  - `_cleanup_deleted_docs`: the `deleted_doc_ids` membership filter had the same O(n²) list-scan pattern over the same per-row hot path as the other three, so it's included here rather than split into a follow-up PR

Follow-up fix 1: the set-backed rewrite requires values to be hashable, whereas the original list-based `in` checks tolerated any type. Malformed provenance (e.g. a `chunk_id`/`doc_id`/description that's a list or dict instead of a string) would now raise `TypeError` instead of being silently tolerated like before.

Follow-up fix 2: the first fix's `repr()` fallback for unhashable values was itself unsound — (a) two equal dicts with different key insertion order produce different `repr()` strings and were wrongly kept as separate entries, and (b) a genuine hashable value could collide with an unrelated malformed value whose `repr()` happened to equal it, silently dropping the genuine one. Replaced with a recursive canonical key (`_canonicalize()` + a `_CanonKey` wrapper type) that preserves dict/list equality semantics and can never collide with an unrelated plain value.

Follow-up fix 3: `_canonicalize()` itself tagged list and tuple identically (so a nested list/tuple pair with equal elements wrongly deduped together, despite `list != tuple` in Python) and only matched `set`, missing `frozenset` (a distinct type that compares equal to `set` by value). Gave tuple its own tag and unified set/frozenset under one.

## Test plan
- [x] Randomized equivalence check (500 trials) confirming the set-backed rewrite produces identical output (order + dedup) to the original list-scan logic on inputs with duplicate/interleaved IDs
- [x] Regression tests added covering duplicate inputs, existing-ID overlap, order preservation, structure-bucket dedup, malformed/unhashable provenance values, dict-key-order equality, the `repr()`-collision case, and the list-vs-tuple type distinction:
  - `test/unit_test/rag/advanced_rag/test_tree_structured_query_decomposition_retrieval.py`
  - `test/unit_test/rag/svr/task_executor_refactor/test_dataset_structure_merger_dedup.py`
- [x] `python -m py_compile` on all changed/added files
- [x] Directly exercised the real `_hashable_key`/`_canonicalize`/`_CanonKey` implementation from both source files in isolation (heavy unrelated imports stubbed out) to confirm dict-order independence, list-order sensitivity, list/tuple distinction, set/frozenset equivalence, and no repr-collision
- [ ] `uv run pytest test/unit_test/rag/advanced_rag test/unit_test/rag/svr/task_executor_refactor` (could not run locally — `uv sync` fails in this environment on an unrelated native build error in `editdistance`, pre-existing and unrelated to this change)


buua436 (commented):
The fix looks correct and the regression coverage is comprehensive. One maintainability concern is that `_CanonKey`, `_canonicalize`, and `_hashable_key` are duplicated in two modules; please consider extracting them into a shared utility to prevent the canonicalization behavior from diverging over time.

DeviousDrops:
Good catch extracted `_CanonKey`, `_canonicalize`, and `_hashable_key` into `common/misc_utils.py` (exported as `hashable_key()`), imported by both modules now. Added unit tests for the shared helper too.

buua436 (approved):
LGTM

PR #18138, opened 2026-08-12
Feat: The dataset-level tree view supports search.

cike8899:
### Summary
Feat: The dataset-level tree view supports search.


wangq8 (commented):
Nice feature! The search implementation is clean:

- Proper debouncing (500ms) via ahooks `useDebounce`
- Query key includes keywords so cache invalidation works correctly
- `trim` from lodash handles whitespace nicely
- Props threading through the component tree is consistent

One small note: in `dataset-nav-service.ts`, `params: { keywords: params.keywords || undefined }` — make sure the backend API handles missing/empty keywords gracefully (returns unfiltered results). If it already does, this is good to go. LGTM overall.

PR #18140, opened 2026-08-12
fix: prevent stale conversation messages during session switch

dcc123456:
### Summary

fix: prevent stale conversation messages during session switch


wangq8 (commented):
Good fix for a race condition during conversation switching. The guard `conversation?.id !== conversationId` correctly prevents stale messages from leaking into a new session. Adding `conversation?.id` to the dependency array is the right call since the effect now reads it. LGTM.

PR #18142, opened 2026-08-12
Fix: Deleting a record from the knowledge base table should remove the corresponding selected record.

cike8899:
### Summary

Fix: Deleting a record from the knowledge base table should remove the corresponding selected record.


wangq8 (commented):
LGTM. Clean fix that ensures deleted documents are properly removed from the row selection state. Good guard checking `code === 0` before removing from selection. Nice cleanup of commented-out code as well.

PR #18153, opened 2026-08-12
Fix: Hide the reset button on the embedded page that is in conversation

cike8899:
### Summary

Fix: Hide the reset button on the embedded page that is in conversation

### DIFF
--- a/web/src/components/embed-container.tsx
+++ b/web/src/components/embed-container.tsx
@@ -24,6 +24,7 @@ type EmbedContainerProps = {
   title: string;
   avatar?: string;
   handleReset?(): void;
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this fix! Clean and minimal change.

**What I verified:**
- The `hideReset` prop defaults to `false`, so existing `EmbedContainer` callers are unaffected.
- `sendLoading` is the right condition — it correctly hides the reset button only while a message is in flight, preventing accidental resets mid-conversation.
- Applied consistently to both share pages (`agent/share` and `next-chats/share`).

**No issues found.** LGTM 🚀

PR #18155, opened 2026-08-12
fix: reorder user setting sidebar menu items and update default redirect path

dcc123456:
### Summary

fix: reorder user setting sidebar menu items and update default redirect path

### DIFF
--- a/web/src/pages/user-setting/sidebar/index.tsx
+++ b/web/src/pages/user-setting/sidebar/index.tsx
@@ -42,6 +42,12 @@ import { useHandleMenuClick } from './hooks';
 
 const menuItems = (t: TFunction) => [
   {
+    icon: <LucideBox className="size-[1em]" />,
+    label: t('setting.model'),
+    key: Routes.Model,
+    'data-testid': 'settings-nav-model-providers',
+  },
+  {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the PR! This is a clean UX improvement — moving Model Providers to the top of the settings sidebar makes sense since it's frequently accessed.

**What's changed:**
- Reorders the user settings sidebar: Model Providers now appears first (before Data Sources and Chat Channels)
- Updates the default redirect from `/user-setting/data-source` to `/user-setting/model` to match the new order

**Review:**
- ✅ Sidebar items are correctly reordered with no missing or duplicated entries
- ✅ The redirect path `/user-setting/model` is valid and defined in `routes.tsx`
- ✅ Icon imports unchanged — `LucideBox` was already imported
- ✅ No regressions or breaking changes

The redirect uses a hardcoded string `/model` while the original used `Routes.DataSource`:
```tsx
// current:
<Navigate to={`/user-setting/model`} replace />
// suggestion:
<Navigate to={`/user-setting${Routes.Model}`} replace />
```
Using `Routes.Model` would be more consistent and resilient to future route path changes. Up to you whether to address it here.

LGTM overall, approved!

PR #18163, opened 2026-08-12
Feat: Adjust the agent list style.

cike8899:
### Summary

Feat: Adjust the agent list style.

### DIFF
--- a/web/src/components/image/index.tsx
+++ b/web/src/components/image/index.tsx
@@ -213,11 +213,15 @@ export const AuthenticatedImg = ({
   );
 };
 
-const Image = ({ id, t, label, className, ...props }: IImage) => {
+const Image = React.forwardRef<HTMLImageElement, IImage>(function Image(
+  { id, t, label, className, ...props },
+  ref,
+) {
   const src = useDocumentImageUrl(id, t);
   const imageElement = (
     <img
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Overall LGTM for the style adjustments. A few observations:

### forwardRef for Image
Good — enables ref forwarding for the Image component, which is standard React practice.

### FlowTypeConfig color additions
The addition of `color` to `FlowTypeConfig` and using it as inline `style={{ color: config.color }}` on icons is a reasonable approach. However, I noticed that the CSS custom properties referenced (`--team-member`, `--team-department`, `--team-group`) are not defined anywhere in the `web/src/` CSS/SCSS files. This means the icons will fall back to inherited text color rather than the intended colors. Consider either:
- Adding definitions for these CSS variables (e.g., in a global stylesheet or theme file), or
- Using hardcoded color values if these are meant to be constants.

### expandable-content.tsx
Removing `children` from the `useEffect` dependency array is the correct React 18 fix — `children` being a new reference on every render would cause unnecessary observer re-setup when only `maxHeight` matters for the effect.

### agent-card.tsx / compilation-template-card.tsx
The refactor from `icon` variable to `config` object pattern is clean and consistent. No issues there.

PR #18164, opened 2026-08-12
fix: improve incremental wiki compilation

buua436:

Improve incremental Wiki generation and fix multiple issues in entity matching, page planning, topic assignment, relationships, and clean rebuilds.

- Use embedding candidates with LLM semantic routing for Mode B page planning.
- Preserve evidence from all entities merged into the same page.
- Let the LLM generate titles for multi-entity pages while keeping stable slugs.
- Preserve entity display labels when multiple entities link to the same page.
- Improve topic selection and add newly generated topics to the shared topic pool.
- Restore Wiki relations from page links and MAP relations.
- Refresh finalized pages before materializing the Wiki graph.
- Fix `wiki_doc_page_source` updates overwriting source chunks sharing the same `doc_id`.
- Recover source chunks damaged by the previous broad update condition.
- Fully clear incremental Wiki state when deleting generated artifacts.
- Exclude disabled documents and chunks from clean Wiki generation.
- Rebuild stale MAP state when pages have been deleted or compilation was interrupted.

### Type of change

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -3992,7 +3992,7 @@ async def update_wiki_page(
 # :meth:`FileCommitService.get_page_commit_detail`.
 
 
-# All seven row types the artifact pipeline writes. Listed in dependency
+# All row types the artifact pipeline writes. Listed in dependency
 # order so partial failures of earlier deletes don't leave behind state
 # that downstream phases would silently reuse. ``wiki_page_graph``
 # is the materialized canvas graph derived from the refined pages —
@@ -4004,8 +4004,13 @@ _WIKI_COMPILE_KWDS = (
     "wiki_page_draft",
     "wiki_page",
     "wiki_page_topic",
+    "wiki_canonical_entity",
+    "wiki_plan_group",
+    "wiki_doc_page_source",
+    "wiki_mode_meta",
     "wiki_entity",
     "wiki_relation",
+    "wiki_page_graph",
 )
 
 # Tunables for the incremental graph loader. See ``get_wiki_graph``.
@@ -4528,6 +4533,54 @@ async def clear_wiki(dataset_id: str, tenant_id: str):
     if pack is None:
         return True, {"deleted": {}}
     index_nm, _ = pack
+
+    # Repair rows damaged by the former doc-page-source upsert before deleting
+    # that bucket. It updated by ``doc_id`` and could stamp ordinary source
+    # chunks as ``wiki_doc_page_source``. Those rows retain chunk content;
+    # genuine tracking rows do not. Removing only the bad marker preserves the
+    # source chunks while allowing the real tracking rows to be cleared below.
+    try:
+        from common.doc_store.doc_store_base import OrderByExpr
+
+        fields = ["id", "content_with_weight"]
+        offset = 0
+        page_size = 1000
+        damaged_row_ids: list[str] = []
+        while True:
+            res = settings.docStoreConn.search(
+                fields,
+                [],
+                {"compile_kwd": ["wiki_doc_page_source"]},
+                [],
+                OrderByExpr(),
+                offset,
+                page_size,
+                index_nm,
+                [dataset_id],
+            )
+            rows = settings.docStoreConn.get_fields(res, fields) or {}
+            for row_id, row in rows.items():
+                if row.get("content_with_weight"):
+                    damaged_row_ids.append(row_id)
+            if len(rows) < page_size:
+                break
+            offset += page_size
+        for row_id in damaged_row_ids:
+            settings.docStoreConn.update(
+                {"id": row_id},
+                {"remove": "compile_kwd"},
+                index_nm,
+                dataset_id,
+            )
+        if damaged_row_ids:
+            logging.warning(
+                "clear_wiki: repaired %d source chunk(s) mislabeled as wiki_doc_page_source kb=%s",
+                len(damaged_row_ids),
+                dataset_id,
+            )
+    except Exception:
+        logging.exception("clear_wiki: failed to repair mislabeled source chunks kb=%s", dataset_id)
+        return False, "Failed to repair legacy Wiki state before clearing"
 
     deleted: dict[str, object] = {}
     for kwd in _WIKI_COMPILE_KWDS:
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -2167,7 +2167,7 @@ async def get_dataset_structure(dataset_id: str, tenant_id: str, kind: str, keyw
 
 # Non-folded template kinds that make a doc eligible for each API kind.
 _ALTERATION_ELIGIBLE_TEMPLATE_KINDS = {
-    "wiki": {"artifacts"},
+    "wiki": {"wiki"},
     "graph": {"knowledge_graph"},
     "mindmap": {"mind_map"},
     "timeline": {"timeline"},
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -4547,7 +4547,8 @@ async def clear_wiki(dataset_id: str, tenant_id: str):
         page_size = 1000
         damaged_row_ids: list[str] = []
         while True:
-            res = settings.docStoreConn.search(
+            res = await thread_pool_exec(
+                settings.docStoreConn.search,
                 fields,
                 [],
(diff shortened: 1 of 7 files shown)


wangq8 (commented):
This is a comprehensive and well-structured improvement to the incremental wiki compilation pipeline. The PR addresses multiple real issues and adds meaningful observability. Detailed review below.

### Strengths

**1. wiki_doc_page_source fix (critical)**
The change from updating by `{"doc_id": doc_id}` to `{"id": doc["id"]}` in `_wiki_update_doc_page_source` is essential — the previous approach was corrupting source chunks that shared the same `doc_id`. This alone justifies the PR.

**2. Repair logic in clear_wiki**
The recovery block that scans for mislabeled source chunks and strips the `wiki_doc_page_source` compile keyword is well-designed: it preserves chunk content while allowing the real tracking rows to be cleared cleanly. The try/except with `logging.exception` and early-return on failure is correct.

**3. Embedding-assisted topic ranking**
`_wiki_rank_topic_candidates` uses embeddings only for candidate recall, letting the LLM remain the selector. The `_wiki_topic_query_text` builder composes query text from title, summary, evidence, and source chunks — a thoughtful approach that avoids superficial cosine similarity matches.

**4. LLM page routing improvements**
The new `_wiki_expand_route_candidates` with authority signals (current_owner → relation → cooccurrence → embedding → candidate_neighbor) is a clean way to bias routing toward authoritative page ownership. The 2-pass routing (initial LLM batch → confirm NEW results with expanded candidates) handles the common case where the LLM says NEW but graph evidence points to an existing page.

**5. Multi-member page support**
Adding `member_evidence` to prompts, `TITLE` line in LLM output, and `build_member_evidence_block` ensures the LLM cannot silently drop members from grouped pages. The display label preservation (`[[target|display_text]]`) during wikilink resolution prevents entity names from being overwritten by page slugs. 

**6. Observability**
The `_wiki_log_stats` function provides machine-readable JSON logging at key decision points (MATCH, PLAN, ROUTE, TOPIC stages) — invaluable for debugging and monitoring.

**7. Test coverage**
347 lines of unit tests covering the new routing logic, merging behavior, and edge cases. The test for `test_page_router_confirms_new_even_without_extra_neighbors` tests the 2-pass routing behavior correctly.

**8. Robustness**
- Retry logic in `_wiki_llm_group_entities` (2 attempts with fallback to single-entity groups instead of cluster fallback)
- Retry missing items in `_wiki_llm_route_batches` without re-running the full batch
- `refresh_idx` call after FINALIZE to ensure index consistency before graph materialization
- Exclusion of disabled document IDs from MAP relation loading in incremental mode

### Minor Observations (non-blocking)

**LLM prompt format change**
The new `TITLE:` line between `SUMMARY:` and `TOPIC:` is a breaking prompt format change. The system prompts are updated consistently, but any cached LLM responses or custom prompts in downstream forks would need updating. This is fine for the main repo — just flagging awareness.

**Topic pool thread safety**
The `topic_pool_lock` + `setdefault` pattern in `_wiki_refine_page` is correct, though `setdefault` itself is atomic in CPython due to the GIL. The lock is still warranted for the double-check pattern.

**Dead link resolution**
The dead link resolution now preserves display text: `[[resolved|display_text]]`. This is a nice touch — previously the display text was silently dropped on resolution.

### Summary
No regressions, no security concerns, good test coverage, clear recovery paths. All 12 claimed improvements in the PR description are verifiable in the diff. Approved.

wangq8 (commented):
Thanks for this comprehensive PR. I reviewed the changes across all seven files and have the following observations:

## Overall
This is a solid improvement to the incremental wiki pipeline. The changes address multiple real bugs (template kind mismatch, doc_id-based upsert damaging source chunks, incomplete cleanup state) while adding meaningful quality improvements to entity matching, page routing, topic selection, and graph restoration. The test coverage for the new routing behavior is appreciated.

## Specific observations

### Template kind fix (`dataset_api_service.py` L2170)
`"artifacts"` → `"wiki"` — correct. This was a bug where the wiki endpoint was checking the wrong template kind, making documents with the wiki template ineligible for wiki alteration.

### `_WIKI_COMPILE_KWDS` expansion
Adding the five missing compile keywords ensures `clear_wiki` truly tears down all incremental state. The dependency order in the comment is preserved.

### Source chunk repair in `clear_wiki`
The repair loop that scans for `wiki_doc_page_source` rows carrying actual content and strips the misapplied marker is a necessary defensive measure. The batching with page_size=1000 and the exception handling that returns a clear error message both look correct. One consideration: the repair only removes `compile_kwd` — if any of these damaged rows also had other wiki-specific fields written into them (unlikely since they were source chunks, but worth noting), those fields would remain.

### Routing overhaul (`wiki_incremental.py`)
The removal of hardcoded thresholds (PAGE_ROUTER_DIRECT_THRESHOLD etc.) and replacement with signal-based candidate ranking + two-pass LLM confirmation is the right direction. Key strengths:

- **Owner-based routing**: entities already assigned to a page (`current_owner` signal) bypass the LLM entirely — correct and efficient.
- **Two-pass confirmation**: entities the LLM marks as NEW get a second pass with expanded candidates (including graph neighbors). This is the fix for false NEW assignments that previously caused page proliferation.
- **`_wiki_expand_route_candidates`**: merges embedding candidates with authoritative ownership, relation evidence, and chunk co-occurrence. The priority ordering (`current_owner > relation > cooccurrence > embedding`) is well-reasoned.
- **Retry-only-missing in `_wiki_llm_route_batches`**: avoids re-processing already-routed entities when one malformed item fails — good optimization.

### Entity matching (`llm_merge_pairs` tracking)
The structured logging of LLM merge decisions via `_wiki_log_stats` is useful for debugging. The `intra_build` scope for merges within the same build iteration vs `existing_canonical` for merges into persisted entities is correctly distinguished.

### `wiki_doc_page_source` upsert fix
Changing the update condition from `{"doc_id": doc_id}` to `{"id": doc["id"]}` is the critical fix. Using `doc_id` was a broad match that could stamp every row sharing that document ID (source chunks, MAP resume rows, etc.). Using the rows own unique `id` correctly targets only the tracking row.

### MAP relation fallback in `_wiki_load_pages_for_graph`
Rebuilding graph outlinks from MAP relations when the LLM fails to generate wikilinks is a good fallback. The name-to-slug resolution (checking slug suffix, title, and entity_names) is thorough. One edge case: if two pages share a member name (e.g., same entity name across different pages), `name_to_slug.setdefault` picks the first page encountered — this is acceptable since the MAP relation endpoint ambiguity is inherently fuzzy, but a comment acknowledging this would help future readers.

### Topic improvements
- `_wiki_rank_topic_candidates`: embedding-based re-ranking of topic candidates before LLM selection is good. The function caches newly encoded topics back into `topic_embeddings` (side effect) — this is intentional for amortization but worth documenting at the call site.
- New topic pool: adds LLM-generated topics to the shared pool so subsequent pages can pick them up. The `topic_pool_lock` usage is correct.
- The `_wiki_prepare_topic_embeddings` list conversion `list(doc_topics.values())` is necessary for concatenation with `[extra_topics or []]`.

### Member evidence in prompts
The `_build_member_evidence_block` and the `## Required Page Members` section in both generate and modify system prompts are well-designed. The instruction "Do not silently omit or replace any member" addresses a real quality issue with multi-entity pages.

### TITLE field in page output
Adding a TITLE line between SUMMARY and TOPIC is clear and backward-compatible since the parser strips unknown prefixes. The logic that single-member pages keep their original title while multi-entity pages get a synthesized title is correct (`if len(member_names) <= 1: title = existing_title`).

### Wikilink pipe handling
Parsing `[[slug|display text]]` and keeping the display label in auto-links (`[[target|name]]`) prevents entity name collisions from being silently lost. The finalize step correctly strips the pipe portion when checking existing links and applies it when auto-linking.

### `refresh_idx` after finalize
Calling `refresh_idx` after the batch of page updates ensures the search index sees the new outlinks before the caller reloads rows to materialize `wiki_entity`/`wiki_relation`. Without this, the graph could be built from a stale snapshot. Good catch.

## Minor suggestions (non-blocking)

1. The retry count in `_wiki_llm_group_entities` (`for attempt in range(2)`) is hardcoded — consider a module-level constant alongside the other tunables.
2. `_wiki_rank_topic_candidates` mutates `topic_embeddings` as a side effect — a brief docstring note would help.
3. The `llm_page_group` log could become verbose with large candidate sets — consider only logging when `len(merged_groups) > 0`.

## Verdict
✅ Approved in spirit. No blocking issues. The changes are well-scoped, bug fixes are targeted, and the routing improvements are backed by tests.

PR #18165, opened 2026-08-12
Fix: Clicking the agent's code tool node is prohibited.

cike8899:
### Summary

Fix: Clicking the agent's code tool node is prohibited.

### DIFF
--- a/web/src/pages/agent/canvas/node/tool-node.tsx
+++ b/web/src/pages/agent/canvas/node/tool-node.tsx
@@ -15,7 +15,7 @@ function InnerToolNode({
   isConnectable = true,
   selected,
 }: NodeProps<IToolNode>) {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Clean fix with clear rationale.

The approach is correct: since Code has no config form, making its cards non-interactive by stripping `data-tool`, `data-tool-id`, and `cursor-pointer` attributes is the right solution — the node click handler will correctly ignore clicks on Code cards. The `edit` prop becoming optional in `ActionButton` and conditionally rendering the pencil icon only when edit is provided is a clean pattern.

One minor note for future cleanup: the `setClickedToolId` removal from the destructured `useGraphStore()` call in `tool-node.tsx` suggests nothing else in that component uses it. Worth a quick check to confirm it is not needed elsewhere in the same component — but from the diff this looks safe.

Nice work.

PR #18166, opened 2026-08-12
Fix: allow datasets sharing a base embedding model to be searched together

euvre:
### Summary

Datasets that use the same embedding model through different provider instances (e.g. `BAAI/bge-m3@1@SILICONFLOW` vs `BAAI/bge-m3@2@SILICONFLOW` — same model and provider, different instance name / API key) could not be selected together on the search page, and the retrieval APIs rejected such combinations too.

Root cause: every layer compared the raw stored references — full composite `model@instance@provider` names or raw `tenant_model` ids — instead of the model they resolve to.

This PR groups datasets by their resolved base embedding model name (e.g. `BAAI/bge-m3`) across all layers:

- **Go**: add `common.BaseModelName` (right-anchored `@` split, mirroring Python's `rsplit("@", 2)[0]`, so model names containing `@` keep their suffix) and `KnowledgebaseDAO.EmbeddingBaseName` (resolves `tenant_embd_id` / raw `tenant_model` ids through `tenant_model`, falls back to the composite base name or the raw id, with per-call caching). Used by `ValidateDatasetEmbeddingModels` (chat create, chat pipeline, dataset search), REST chat dataset validation, chunk retrieval, and the agent retrieval tool; the duplicated `splitModelNameAndFactory` helper is removed.
- **Python**: `validate_dataset_embedding_models` resolves ids via `get_composite_model_name_by_ids` (best-effort), and the SDK `/retrieval` endpoint reuses it instead of `split_model_name`.
- **Web**: the Datasets multi-select disables mismatched options by comparing `getEmbeddingBaseName` results instead of full composite strings.

Unresolvable/stale ids keep their raw value, so only exact matches group; datasets without an embedding model still cannot mix with embedded ones, and empty / tag-parser datasets remain disabled in the UI.

Tests: Go unit tests (`common`, `dao`, `service`, `chunk`, `dataset`, `agent/tool`), new Python unit tests for the validator, new jest cases for `getEmbeddingBaseName`. Also verified end-to-end in the browser: cross-instance datasets are now co-selectable and pass backend validation, while different-model datasets are still rejected.

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -26,17 +26,16 @@ from quart import request
 
 from api.apps import login_required
 from api.apps.services import structure_graph_common as sgc
-from api.db.joint_services.tenant_model_service import (
-    split_model_name,
-    resolve_model_config,
-    get_tenant_default_model_by_type,
-)
 from api.db.db_models import Document, Task
+from api.db.joint_services.tenant_model_service import (
+    get_tenant_default_model_by_type,
+    resolve_model_config,
+)
 from api.db.services.doc_metadata_service import DocMetadataService
 from api.db.services.document_counter_service import release_reparse_counters
 from api.db.services.document_service import DocumentService
 from api.db.services.file2document_service import File2DocumentService
-from api.db.services.knowledgebase_service import KnowledgebaseService
+from api.db.services.knowledgebase_service import KnowledgebaseService, validate_dataset_embedding_models
 from api.db.services.llm_service import LLMBundle
 from api.db.services.task_service import TaskService, cancel_all_task_of, queue_tasks
 from api.db.services.tenant_llm_service import TenantLLMService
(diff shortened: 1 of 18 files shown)


wangq8 (commented):
This is a thorough, well-structured fix that correctly addresses the root cause: datasets using the same embedding model through different provider instances could not be selected or searched together because every layer compared raw stored references instead of resolved base model names.

**What works well:**

- **Right-anchored split** (`common.BaseModelName`) correctly preserves model names containing `@` (e.g., LM Studio quant suffixes), mirroring Python `rsplit("@", 2)`. Good symmetry.
- **`KnowledgebaseDAO.EmbeddingBaseName`** resolves tenant_model IDs through the `tenant_model` table with per-request caching. Falls back gracefully for stale/unresolvable IDs so only exact matches group together — safe degradation.
- **`splitModelNameAndFactory` removal** in `chat.go` — good DRY cleanup, replaced by the shared `common.BaseModelName`.
- **Python changes** reuse `get_composite_model_name_by_ids` with best-effort resolution and a try/except guard. The `_kb_embedding_base_name` helper handles the full resolution chain.
- **Web `getEmbeddingBaseName`** reuses the existing, tested `parseModelValue`.
- **Test coverage** is excellent: Go unit tests for the new functions, updated existing tests for the new signatures, Python AST-based tests for the validator (clever approach to avoid full import chain), and jest cases for the web helper. All edge cases are covered — same/different bases, tenant_model ID resolution, stale IDs, mixed empty/embedded datasets.

**Minor observations:**

1. The `cache` parameter in `EmbeddingBaseName` is a raw `map[string]string` — not thread-safe in general. In this codebase it is always created per-request and used within a single goroutine call chain, so this is safe in practice, but worth noting for future maintainers.

2. The signature change for `ValidateDatasetEmbeddingModels` (adding `ctx` and `db`) is a breaking change, but all internal call sites are updated. No external consumers appear affected.

3. The Python `validate_dataset_embedding_models` now imports `get_composite_model_name_by_ids` inside the function — this is a deferred import pattern. Works fine but differs from how other imports in the file are handled.

Overall this is an excellent PR — correct fix, good architecture, thorough tests. Approved.

Lynn-Inf (approved):
LGTM

PR #18167, opened 2026-08-12
Fix: Preserve the text format returned by the API.

cike8899:
### Summary

Fix: Preserve the text format returned by the API.

### DIFF
--- a/web/src/components/top-select.tsx
+++ b/web/src/components/top-select.tsx
@@ -14,7 +14,7 @@
  *  limitations under the License.
  */
 
-import { useMemo } from 'react';
+import { forwardRef, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { SelectWithSearch } from './originui/select-with-search';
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Two changes in this PR:

**`testing-result.tsx`** — Adding `whitespace-pre-wrap` is the right fix for preserving line breaks and whitespace in chunk content returned by the API. Without it, multi-line text collapses into a single blob. This matches the PR title well.

**`top-select.tsx`** — Converting to `forwardRef` and passing the ref to `SelectWithSearch` is fine and enables consumers to attach a ref (e.g., for `PopoverTrigger` anchoring via Radix/shadcn). However, this change is unrelated to the PR title about preserving text format. In the current callers (`search-view.tsx` and `testing-form.tsx`), the `TopSelect` is used without a ref, so nothing would break, but nothing is explicitly fixed by it either.

If the `forwardRef` is needed for a follow-up or a separate fix, consider noting it in the description. Otherwise the code looks correct.

Approved with the note above.

PR #18171, opened 2026-08-12
fix(dao): honor tenant-configured context window override (D22/D23)

jay77721:
## Summary

Make `ResolveModelContentLength` honor the per-model custom **context window length** (`content_length`) — stored in the Python-legacy `tenant_model.extra["max_tokens"]` field, whose semantic meaning is the context window, NOT the generation cap — **before** any provider-catalog read, and remove the parallel service-layer implementation so every consumer shares one resolution path.

## Why

Python resolves a chat model's **context window** as `model_extra.get("max_tokens") or 8192` — the per-model "max_tokens" a user configures IS the context window length (`content_length` / Python `max_length`), not the output cap. Go ignored that override: custom/local models (not in the catalog) resolved to 0, so the agent silently fell back to the 8192 budget (truncating large contexts) and the extractor skipped fitting entirely (D22). The service layer also had a duplicate catalog-only resolver feeding the knowledge compiler (RAPTOR/dedup/merge) (D23).

## Changes

- `internal/dao/model_context.go`: `ResolveModelContentLength` now takes `tenantID` and resolves the `tenant_model` row (UUID or composite via tenant provider/instance/model) **before any catalog read**:
  1. active row with a custom context-window override (`extra["max_tokens"] > 0`; the field name is a Python legacy for the context window, i.e. `content_length`) → that override wins;
  2. otherwise fall back to catalog `content_length` (the context window length, via the tenant row, the composite reference parts, or resolved driver + model name).
- `internal/service/model_service.go`: `ResolveModelContextLength` delegates to the shared dao resolver; `modelContextLengthByID` / `modelContextLengthByName` (parallel catalog-only implementations) are removed.
- `internal/ingestion/component/extractor.go`: `extractorContextLength` passes `tenantID`.
- `internal/agent/component/llm.go`: agent reads `tenant_id` from canvas state and passes it.

## Testing

- dao: 12 tests — existing composite/UUID/driver-fallback/unknown/too-many-parts plus new UUID/composite/custom-model override cases, invalid/zero overrides falling back to the catalog, and inactive rows ignoring the override.
- agent: end-to-end `TestLLM_Invoke_UUIDModel_CustomContextOverride` proves a 2000-token context-window override drives trimming (not gpt-4o's 128k catalog content_length); existing content_length/fallback tests still pass.
- `bash build.sh --test ./internal/dao/... ./internal/agent/component/... ./internal/ingestion/component/...` and service/ingestion-task compiles pass.

## Note

Branched from `pr/3-agent-content-length` (the #18093 agent content_length work) so the agent caller is included; once #18093 merges, this PR's diff shrinks to the resolver/service/extractor changes. D24 (chat_pipeline using catalog `MaxOutput` as context) is tracked separately.

## Review updates (2026-08-12)

- UUID resolution is intentionally unscoped by tenant: UUIDs are globally unique/unguessable and shared/joined-tenant models are legitimate (matches the previous service behavior and Python's `get_model_config_by_id`). This also fixes a service unit test that passes a user id as the tenant id.
- `splitCompositeModelRef` recombines 4+ `@`-separated segments into the model name (LM Studio style `name@quant@provider@...`), preserving the old service `parseModelName` capability.
- `modelExtraMaxTokens` parses via `json.Number` (int / float / numeric string).
- Composite override lookup requires an active instance.

## Known limitations / follow-ups

- `extra["max_tokens"]` semantics are not yet unified across Go: `internal/ingestion/component/dispatch_model.go` still interprets the same field as a `max_output` (generation cap) override, while this PR (and Python) treat it as the context window length (`content_length`). A follow-up should reconcile that path.
- The composite lookup does not yet implement the agent's "default instance → sole active instance" fallback (`findSoleActiveProviderInstance`); a 2-segment `model@provider` ref on a tenant whose default instance is not literally named `default` will fall back to the catalog length.
- D24 (chat_pipeline using catalog `MaxOutput` as the context window) is tracked separately in the extractor bug notes.

### DIFF
--- a/internal/agent/component/llm.go
+++ b/internal/agent/component/llm.go
@@ -314,7 +314,15 @@ func (c *LLMComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[strin
 	// Resolve the model's context window (content_length) for message
 	// fitting. 0 means the model is unknown → fitMessages falls back to
 	// 8192, matching Python's chat_mdl.max_length = model_config.get("max_tokens") or 8192.
-	contentLength := dao.ResolveModelContentLength(ctx, db, originalModelID, p.Driver, p.ModelID)
+	// tenantID scopes composite-reference resolution to the tenant's own rows
+	// so a per-model "max_tokens" override in tenant_model.extra is honored.
+	tenantID := ""
+	if state, _, err := runtime.GetStateFromContext[*runtime.CanvasState](ctx); err == nil && state != nil {
+		if tid, ok := state.Sys["tenant_id"].(string); ok {
+			tenantID = tid
+		}
+	}
+	contentLength := dao.ResolveModelContentLength(ctx, db, tenantID, originalModelID, p.Driver, p.ModelID)
 	if contentLength <= 0 {
 		// A 0 makes fitMessages fall back to the 8192 default budget, which can
 		// silently discard most of a large-context prompt, so surface the
--- a/internal/agent/component/llm_test.go
+++ b/internal/agent/component/llm_test.go
@@ -262,6 +262,67 @@ func TestLLM_ThinkingFieldRoundTrip(t *testing.T) {
 	}
 }
 
+// TestLLM_Invoke_CompositeModel_CustomContextOverride verifies the composite
+// reference path of the tenant-configured override: a 2000-token extra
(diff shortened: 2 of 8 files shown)


yuzhichang:
LGTM

PR #18172, opened 2026-08-12
fix: keep loading indicator on streaming assistant message in agent chat

dcc123456:
### Summary

fix: keep loading indicator on streaming assistant message in agent chat

The last assistant message dropped its loading state as soon as the first content chunk arrived, so the spinner vanished while the answer was still streaming. Tie the indicator to sendLoading only.

### DIFF
--- a/web/src/pages/agent/chat/box.tsx
+++ b/web/src/pages/agent/chat/box.tsx
@@ -107,8 +107,7 @@ function AgentChatBox() {
                   loading={
                     message.role === MessageType.Assistant &&
                     sendLoading &&
-                    derivedMessages.length - 1 === i &&
-                    !message.content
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Simple and targeted fix — removing `!message.content` from the loading condition means the spinner no longer disappears as soon as the first streaming chunk arrives. The loading indicator now correctly stays visible while `sendLoading` is true, regardless of whether content has started streaming. The condition `derivedMessages.length - 1 === i` still ensures this only applies to the last message. One edge case to keep in mind: if `sendLoading` remains true after the stream completes (e.g., due to an upstream bug), the spinner would linger. That is a separate concern and not introduced by this change.

Approved.

PR #18173, opened 2026-08-12
fix(ingestion): remove broken built-in Resume pipeline

euvre:
### Summary

The built-in **Resume** ingestion pipeline template (`internal/ingestion/pipeline/template/ingestion_pipeline_resume.json`) hardcoded a deployment-specific LLM in its Extractor component (`llm_id: THUDM/GLM-4.1V-9B-Thinking@SILICONFLOW`, `tenant_llm_id: 29`). On any other deployment, parsing a file (e.g. a PDF) with the Resume chunk method deterministically fails at the Extractor stage:

```
extractor: resolve model "THUDM/GLM-4.1V-9B-Thinking@SILICONFLOW": instance "default" lookup failed: record not found
[NodeRunError] canvas: component "Extractor:ThreeDrinksAct" invoke: ...
```

Reproduced on a fresh deployment: upload a PDF to a dataset using the built-in Resume method, run parsing, and the task always fails with the error above. The template is an internally customized pipeline (a deployment-specific model id plus a resume-specific extraction prompt) that cannot work as a generic built-in, so this PR removes it along with its references:

- Drop `ingestion_pipeline_resume.json` from the embedded builtin registry. This also removes Resume from the front-end built-in chunk-method list, which is driven by `GET /api/v1/pipelines`.
- Drop the unused `ParserTypeResume` constant, the `resume:Resume` entries in the default tenant `parser_ids` lists (admin/user services), and the `resume` entry in the default parser config map.
- Remove the resume-specific golden test and integration test pinned to the deleted template (the all-templates smoke test globs the template dir, so it adapts automatically).

Verification:
- After the removal, the built-in chunk-method list no longer offers Resume (checked via the web UI and the `/api/v1/pipelines` response).
- The same PDF parses successfully with the General method (1 chunk produced, task succeeds).
- `build.sh --test` passes for the touched packages (`internal/ingestion/...`, `internal/common`, `internal/admin`, `internal/service`).


wangq8 (commented):
Thanks for catching and cleaning this up. The hardcoded deployment-specific LLM (`THUDM/GLM-4.1V-9B-Thinking@SILICONFLOW` + `tenant_llm_id: 29`) makes the built-in Resume template a guaranteed failure on any deployment without that exact model — this is a real bug and removing it is the right call.

**Cleanup is thorough:**

- Template JSON deleted
- `ParserTypeResume` constant removed from `dataset.go`
- `"resume"` entry removed from parser config map
- `resume:Resume` stripped from both `admin/service.go` (CreateUser) and `service/user.go` (Register) parser ID lists
- Builtin registry test golden updated
- Integration test + golden component params test removed
- The all-templates smoke test globs the template dir, so it adapts automatically — no coverage gap

**One thing to note (not blocking):** Existing tenants/datasets that had `parser_id="resume"` will now get a "method not found" error instead of the confusing LLM resolution error. This is strictly better (actionable failure vs silent confusion), but might be worth a brief mention in release notes so self-hosted deployments know to switch affected datasets to the General method.

Nice clean PR — LGTM.

PR #18175, opened 2026-08-12
fix(ingestion): strip mid-text think blocks in extractor metadata path

jay77721:
### Summary
The metadata extraction path (`callStructured` -> `cleanLLMText`) applies only a single cleanup layer, while Python's `gen_metadata` applies two: `LLMBundle.async_chat`'s `_remove_reasoning_content` + `re.sub(r"^.*</think>", "", ans, re.DOTALL)`.

`cleanLLMText` only strips a reasoning block when the response *begins* with `<think>`. When the LLM emits a mid-text reasoning block preceded by a preamble (e.g. `preamble<think>reasoning</think>{"a": 1}`), Python strips the block and parses the JSON, but Go keeps the raw string -> `tryParseJSONObject` fails -> the whole metadata extraction is **silently dropped**.

Keywords/questions (`auto`) already get a second layer via `cleanExtractionResult`; only the metadata path is missing it.

### Change
- Extract `stripTrailingThink` (mirrors Python's `re.sub(r"^.*</think>", "", s, re.DOTALL)`), reused by `cleanExtractionResult` (behavior unchanged).
- Apply `stripTrailingThink` in `callStructured` after `cleanLLMText`, matching `gen_metadata`'s double cleanup. No `**ERROR**` check added -- Python's `gen_metadata` has none either.
- Tests: `TestExtractorComponent_callStructured_MidTextThink` (regression) and `TestStripTrailingThink`.

### Tests
`bash build.sh --test ./internal/ingestion/component/` passes on the full package. Head is rebased onto latest upstream `main` (8c5c87047).

### DIFF
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -1043,12 +1043,21 @@ func cleanLLMText(s string) string {
 	return strings.TrimSpace(s)
 }
 
-// cleanExtractionResult strips `</think>` tags and rejects `**ERROR**` responses,
-// matching Python's keyword_extraction and question_proposal post-processing.
-func cleanExtractionResult(s string) string {
+// stripTrailingThink mirrors Python's re.sub(r"^.*</think>", "", s, re.DOTALL)
+// in generator.py's post-processing (e.g. gen_metadata:960): cut everything
+// through the LAST `</think>` so a mid-text reasoning block preceded by a
+// preamble is stripped on top of cleanLLMText's leading-only cleanup.
+func stripTrailingThink(s string) string {
 	if i := strings.LastIndex(s, "</think>"); i >= 0 {
 		s = s[i+len("</think>"):]
 	}
+	return s
+}
+
+// cleanExtractionResult strips `</think>` tags and rejects `**ERROR**` responses,
+// matching Python's keyword_extraction and question_proposal post-processing.
+func cleanExtractionResult(s string) string {
+	s = stripTrailingThink(s)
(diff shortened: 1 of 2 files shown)


jay77721:
Thanks for the detailed review. Actions taken on the review feedback:

1. **Rebase onto latest `main`** — the PR head is now based on `8c5c87047` (latest `upstream/main`). On this base, `bash build.sh --test ./internal/ingestion/component/` passes the full package (including `TestLlmtagChunk_MessageFit`, `TestFitExtractorMessages_RejectsEmptyUserTurn`, `TestFitExtractorMessages_RejectsSystemPromptLoss` — all green in my environment). The three message-fit failures you observed likely belong to an intermediate `main` state and are already resolved upstream.

2. **Added the end-to-end test (your optional item ③)** — new `TestExtractorComponent_runEnableMetadata_MidTextThink` exercises the full path (LLM call → second-layer `</think>` strip → JSON parse → merge), complementing the `callStructured`-level regression test.

3. **Duplicate `</think>`-strip implementations (your item ②)** — agreed: `stripTrailingThink` (extractor.go), `stripThinking` (docs_generator.go:369) and `reThinkPrefix` (llm.go:1281) are the same "cut through last `</think>`" logic. Consolidating them into a shared helper is worthwhile but a cross-package refactor; keeping it out of this PR per "keep changes small and local". I'll track it separately.

On the edge case you flagged (`{"a": "x</think>y"}`): confirmed it is intentionally aligned with Python (both truncate), and consistent with the existing `stripThinking` behavior in docs_generator.go.

PR #18179, opened 2026-08-12
refactor(common): unify </think>-strip into shared StripThinkTrailing helper

jay77721:
### Summary
Collapse the nine identical **"cut through the last `</think>`"** implementations — mirroring Python's `re.sub(r"^.*</think>", "", s, re.DOTALL)` — into one shared helper `common.StripThinkTrailing`, preventing future behavior drift between copies.

This refactor supersedes the local `stripTrailingThink` helper introduced by #18175.

### Behavior note
Mostly behavior-preserving, with **one** intentional change:

- `docs_generator.go`: the exported content is now trimmed **even when no think block is present** (previously trim happened only when `</think>` existed). This aligns with Python's `_strip_thinking`, which always `.strip()`s (agent/component/message.py:386). Documented at the code site and pinned by `TestResolveDocsGeneratorContent_ThinkAndTrim`.

An earlier draft also changed `raptor.go`'s control flow; that was reverted — the `</think:6124c78e>` vendor fallback keeps its original `else-if` semantics (only active when no standard `</think>` exists).

### Copies collapsed
- `internal/ingestion/component/extractor.go` — `cleanLLMText` guard + `cleanExtractionResult`
- `internal/ingestion/component/extractor_tag.go` — `parseTaggerResponse`
- `internal/ingestion/component/knowledge_compiler/tree/raptor.go` (keeps the `</think:6124c78e>` else-if fallback)
- `internal/agent/component/llm.go` — `cleanFormattedAnswer` (`reThinkPrefix`)
- `internal/agent/component/docs_generator.go` — `stripThinking`
- `internal/agent/harness/route.go` — `reThinkTag`
- `internal/service/load_prompt.go` (`thinkBlockRE`) -> used by `generator.go`, `metadata_filter.go`
- `internal/service/think_tag.go` — `ExtractVisibleAnswer` (composed as `stripThinkTags(StripThinkTrailing(raw))`)

### Notes
- Helper is `LastIndex`-based: greedy -> last `</think>`, no `</think>` -> unchanged, no regex overhead. The parity test (`TestStripThinkTrailingParityWithPython`, 7 cases) lives in `internal/common`; the greedy-vs-non-greedy doc is on the helper.
- Per-caller trim semantics preserved at call sites.
- `raptor.go` uses the `rfcommon` alias to avoid clashing with its own `knowledge_compiler/common` import.

### Tests
`bash build.sh --test` passes for `internal/common`, `internal/ingestion/component`, `internal/ingestion/component/knowledge_compiler/tree`, `internal/agent/component`, `internal/agent/harness`, `internal/service` (including the new `TestSummarizeTextsStripsVendorThinkCloseFallback`).

### DIFF
--- a/internal/agent/component/docs_generator.go
+++ b/internal/agent/component/docs_generator.go
@@ -43,6 +43,7 @@ import (
 
 	iow "ragflow/internal/agent/component/io"
 	"ragflow/internal/agent/runtime"
+	"ragflow/internal/common"
 	"ragflow/internal/storage"
 	"ragflow/internal/utility"
 )
@@ -349,7 +350,7 @@ func resolveDocsGeneratorContent(ctx context.Context, configured string, inputs 
 		}
 	}
 	if !strings.Contains(content, "{{") {
-		return stripThinking(content), nil
+		return strings.TrimSpace(common.StripThinkTrailing(content)), nil
 	}
 
 	state, _, err := runtime.GetStateFromContext[*runtime.CanvasState](ctx)
@@ -363,14 +364,7 @@ func resolveDocsGeneratorContent(ctx context.Context, configured string, inputs 
 	if err != nil {
 		return "", fmt.Errorf("resolve content template: %w", err)
 	}
-	return stripThinking(resolved), nil
-}
-
-func stripThinking(content string) string {
-	if idx := strings.LastIndex(content, "</think>"); idx >= 0 {
-		return strings.TrimSpace(content[idx+len("</think>"):])
-	}
-	return content
+	return strings.TrimSpace(common.StripThinkTrailing(resolved)), nil
 }
 
 func storeAgentAttachment(ctx context.Context, docID string, payload []byte) (bool, error) {
--- a/internal/agent/component/docs_generator.go
+++ b/internal/agent/component/docs_generator.go
@@ -350,7 +350,9 @@ func resolveDocsGeneratorContent(ctx context.Context, configured string, inputs 
 		}
 	}
 	if !strings.Contains(content, "{{") {
-		// Trim even when no think block is present — matches Python's
+		// The think cut itself is the greedy re.sub(r"^.*</think>", "", s,
+		// re.DOTALL) pattern (as in generator.py). The unconditional trim —
+		// applied even when no think block is present — mirrors Python's
 		// _strip_thinking, which always .strip()s the exported content.
 		return strings.TrimSpace(common.StripThinkTrailing(content)), nil
 	}
--- a/internal/agent/component/docs_generator.go
+++ b/internal/agent/component/docs_generator.go
@@ -350,6 +350,8 @@ func resolveDocsGeneratorContent(ctx context.Context, configured string, inputs 
 		}
 	}
 	if !strings.Contains(content, "{{") {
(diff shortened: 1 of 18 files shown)


jay77721:
Stacking note: this PR is stacked on #18175 (`fix/extractor-metadata-think-cleanup`). The current diff includes #18175's commit; once #18175 merges to `main`, this PR's diff reduces to only the two refactor commits.

This refactor **supersedes and removes** the `stripTrailingThink` helper introduced by #18175 — it lives on as `common.StripThinkTrailing`. Review #18175 and this PR together; they touch overlapping lines in `internal/ingestion/component/extractor.go`.

jay77721:
#18175 has merged (squash `207e2eaf3`). This PR has been rebased onto the latest `main`; the stacked commit is gone and the diff now contains only the two refactor commits. All affected packages pass `bash build.sh --test`.

jay77721:
Addressed the review feedback:

1. **docs_generator.go behavior change (priority 1)** — acknowledged. `strings.TrimSpace(common.StripThinkTrailing(content))` now trims even when no `<think>` is present, whereas the old `stripThinking` returned the content unchanged. This is intentional: Python's `_strip_thinking` (agent/component/message.py:386) always `.strip()`s, so the new behavior is closer to the Python original. Documented at the code site (docs_generator.go) and pinned by the new `TestResolveDocsGeneratorContent_ThinkAndTrim`. Updated this PR's commit message accordingly.

2. **Call-site tests (recommendations 2 & 3)** — added:
   - `TestCleanFormattedAnswer` (llm_test.go): think + JSON-fence interaction order.
   - `TestUnmarshalModelJSON` (route_test.go): think prefix + fence before JSON parse.
   - `TestResolveDocsGeneratorContent_ThinkAndTrim` (docs_generator_test.go): pins the always-trim vs think-strip behavior.

3. **generator.go / GenMetaFilter** — acknowledged as pre-existing test gaps (not introduced by this PR); left for a separate follow-up.

All affected packages pass `bash build.sh --test` (common, ingestion/component, knowledge_compiler/tree, agent/component, agent/harness, service).

jay77721:
Responding to the latest review round:

1. **`think_tag.go` ExtractVisibleAnswer (missed 9th copy)** — now closed. It composes `stripThinkTags(common.StripThinkTrailing(raw))`, byte-equivalent to the previous inline cut + tag scrub (verified against its 5 existing tests). The "collapse 8 copies" goal is now actually 9.

2. **raptor.go control-flow change** — reverted. Restored the `else-if` so the `</think:6124c78e>` vendor fallback only runs when no standard `</think>` exists, eliminating the pathological both-tags behavior change. Added `TestSummarizeTextsStripsVendorThinkCloseFallback` to pin the single-tag semantics.

3. **docs_generator.go comment accuracy** — fixed. The comment now states the think cut is the greedy `re.sub(r"^.*</think>", ...)` pattern and only the unconditional trim mirrors `_strip_thinking`'s final `.strip()`.

4. **PR description "pure refactor"** — corrected to "mostly behavior-preserving" with the docs_generator trim change explicitly documented (now the only behavior change, since raptor is restored).

All six affected packages pass `bash build.sh --test`.

PR #18180, opened 2026-08-12
fix(web): keep single-chat SSE stream alive across route changes

dcc123456:
### Summary

fix(web): keep single-chat SSE stream alive across route changes

Asking a question and navigating away mid-answer lost the stream: the answer bubble came back empty or frozen, and the real answer only appeared once the backend had finished and the session was refetched.

All SSE state lived in component-local state and refs, so unmounting /chat/:id destroyed it. The fetch and reader loop already survived the unmount, but every chunk was written into a dead setState and silently discarded, and on return a fresh hook found an empty cache while the server had not persisted the answer yet.

Hoist the message list, the streaming flag and the AbortController into a module-level Zustand store keyed by conversation id, and move the reader loop into a plain module-level async function that writes into that store. Components become pure subscribers, so unmounting no longer touches the stream.

Scope is the single-chat path only; useSendMessageWithSse and the multi-model path are untouched. Recovery across a hard reload is out of scope, since the browser kills the request.

Also on this path:
- a server response resolving right after a stream ends can no longer truncate the finished answer (message-count comparison in hydrateFromServer), while still letting real message ids, references and out-of-band updates through
- a failed stream parks the question text on its own session instead of typing it into whichever input happens to be mounted, and now raises a notification instead of failing silently
- chunks are coalesced into one store write per 60ms and the store no longer records every chunk into Redux DevTools, which is what made navigating away mid-answer feel slow
- a non-abort reader error breaks out of the parse loop instead of spinning forever

### DIFF
--- a/web/src/pages/next-chats/chat-stream/constants.ts
+++ b/web/src/pages/next-chats/chat-stream/constants.ts
@@ -1,0 +1,21 @@
+// Maximum number of idle chat sessions kept in the streaming store. Sessions
+// that are streaming, or the one currently being acted on, are never evicted.
+export const MaxCachedChatSessions = 10;
+
+// Stable id prefix for the seeded prologue message, so re-seeding the same
+// conversation is idempotent under React StrictMode's double effects.
+export const PrologueMessageIdPrefix = 'prologue_';
+
+// Minimum gap between two store writes while a stream is running. SSE chunks
+// arrive far faster than markdown can be re-rendered, and every write copies the
+// message list and re-renders the message item, so coalescing them cuts the cost
+// without any visible change to how the answer grows. The final chunk always
+// flushes immediately.
+export const AnswerFlushIntervalMs = 60;
+
+// Redux DevTools keeps a full state snapshot per action. Streaming dispatches one
+// action per chunk, each carrying the whole accumulated answer, so recording it
+// makes a long answer quadratically expensive to produce — and it keeps costing
+// after the chat page unmounts, which is what makes navigating away mid-answer
+// feel slow. Flip to true only while debugging the store.
+export const EnableChatStreamDevtools = false;
--- a/web/src/pages/next-chats/chat-stream/run-stream.ts
+++ b/web/src/pages/next-chats/chat-stream/run-stream.ts
@@ -1,0 +1,132 @@
+/**
+ * Module-level SSE driver. Deliberately not a hook: it must keep running (and
+ * keep writing into the store) after the chat page unmounts, so that navigating
+ * away mid-answer and back shows the stream still in progress.
+ */
+import { IMessage } from '@/interfaces/database/chat';
+import {
+  CompletionChunk,
+  parseCompletionEventStream,
+  readJsonSafely,
+  requestChatCompletionStream,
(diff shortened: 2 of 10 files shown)


wangq8 (commented):
Thanks for this PR — the architecture is well thought-out and the Zustand store at module level is the right approach for keeping SSE streams alive across unmounts. The attention to edge cases (StrictMode double effects, hydration races, eviction policy) shows careful engineering.

A few observations and questions:

**1. `chat-completion-stream.ts` — bypassing axios**
The new `requestChatCompletionStream` uses native `fetch` directly instead of `@/utils/next-request` (axios). The comment explains this is for `ReadableStream` access. However, this means it bypasses any global request interceptors (auth refresh, error normalization, base URL resolution). If the axios instance has important middleware, this could cause subtle issues. Is `getAuthorization()` alone sufficient for all deployment scenarios (e.g. token refresh on expiry)?

**2. `zod` validation for SSE payloads**
`parseCompletionEventStream` parses SSE `data` with bare `JSON.parse` and yields `as CompletionChunk` — there is no runtime validation of the chunk shape. If the backend changes the payload schema, a missing `answer` field would silently produce `undefined` and break `mergeAnswerChunk`. Consider adding a lightweight schema check or at least a defensive `typeof chunk.answer !== "string"` guard.

**3. Store eviction vs. hydration race**
`evictIdleSessions` is called before many store actions but never on the hot path (`applyAnswer`). However, if a session is evicted while its `runChatCompletionStream` is still feeding chunks, `applyAnswer` would target a missing session — the `if (!previous) return state` guard in `applyAnswer` silently drops the write. Is there a path where a streaming session could be evicted? (The comment says "Never evicts a streaming session", which `evictIdleSessions` does enforce — confirming this is safe.)

**4. `hydrateFromServer` message-count comparison**
`if (session && messages.length < session.messages.length) return;` — this correctly prevents a shorter server list from truncating a just-finished answer. However, is it possible for the server to legitimately return fewer messages (e.g. after a message deletion from another tab)? In that case the store would hold stale deleted messages indefinitely. This is probably an acceptable tradeoff vs. the truncation bug, but worth noting.

**5. `tenant.go` in a different PR?**
The diff shows an unrelated change to `internal/service/tenant.go` (removing factory model type validation), but that appears to be from PR #18121 (separate review). Just confirming this PR does not touch that file.

**6. `useCurrentChatSession` — scroll ref**
`useScrollToBottom(messages, messageContainerRef)` is called unconditionally every render. The `scrollRef` it returns is passed up through `useSendMessage`. Does the scroll-to-bottom logic correctly handle the case where new chunks arrive while the user has scrolled up to read earlier messages? Some implementations auto-scroll when streaming but respect user scroll-away — worth confirming.

**7. `PrologueMessageIdPrefix` — collision risk**
`${PrologueMessageIdPrefix}${conversationId}` — if `conversationId` is a UUID this is safe, but if two conversations happen to have the same id prefix string this could collide. This is probably fine.

Overall strong PR with well-documented tradeoffs. The chunk coalescing (60ms batches) and DevTools opt-out are good performance-conscious choices. None of the above is blocking — just things to consider or clarify.

PR #18181, opened 2026-08-12
fix: refine structure graph keyword search

buua436:

Fixes structure graph keyword search accuracy and incomplete tree paths.

- Prioritizes exact entity-name matching before lexical or semantic fallback.
- Tokenizes BM25 queries consistently with indexed structure content.
- Limits KNN fallback to one semantic match to avoid unrelated sibling nodes.
- Returns the matched tree or page-index node with its complete ancestor path and relations up to the root.
- Handles relation endpoint casing consistently across document- and dataset-level structures.

### Type of change


wangq8 (commented):
Thanks for this PR — the approach of prioritizing exact entity-name matching before BM25 and limiting KNN fallback to a single candidate is well-reasoned and addresses real quality issues in structure graph keyword search.

A few observations:

**1. Ancestor traversal now one-by-one per casing variant**
The old code batched ancestor lookups with `to_entity_kwd=sorted(ancestor_frontier)`. The new code iterates `ancestor_frontier.values()` (original-case names) and for each, tries both original-case and lower-case via `_endpoint_terms` in separate queries. This is N×2 queries per tree level instead of 1. For deeply nested structures with many matched leaves this could be noticeably slower. A comment acknowledging the tradeoff would be helpful for future readers. How deep do real-world trees typically go?

**2. `tokenize_for_search` import**
The import `from rag.advanced_rag.knowlege_compile._common import tokenize_for_search` is fine — the function exists and returns a `(coarse, fine)` tuple. But the result is joined with a simple `" ".join(dict.fromkeys(...))`, which deduplicates by insertion order. If `coarse` and `fine` produce identical tokens this is a no-op, but worth confirming this is always the intended behavior.

**3. `exact_name` empty-string guard**
`exact_name = str(keywords).strip().lower()` — if `keywords` is `None`, `str(None)` gives `"None"` which would trigger a spurious exact-name search. Suggest `str(keywords or "").strip().lower()`.

**4. `tree_entities` only populated for tree/page_index**
The `tree_entities` list is declared but only populated inside the `if structure_kind in {"tree", "page_index", "pageindex"}` block. For non-tree structures `tree_entities` remains `[]` and `entities = list(matched_nodes) + tree_entities` is just `matched_nodes` — correct, but the neighbor entity fetch also gained `and structure_kind not in {"tree", ...}` guard which avoids double-fetching. Looks right.

**5. Tests**
The unit tests are thorough and cover the key scenarios: tree path resolution, BM25 tokenization, and KNN limit. The monkeypatch approach is pragmatic given the deep dependency chain. Consider adding a test for the `keywords=None` edge case.

Overall the logic is solid and the tests give confidence. Minor nits above, none blocking.

Issue #5838, opened 2025-03-10
[Question]: The user_default_llm which set  in service_conf.yaml.template doesn't work


alice77-zhen:
### Describe your problem

1. I have set the `user_default_llm` in  **service_conf.yaml.template** and **service_conf.yaml** , but the interface of model providers shows **Tongyi-Qianwen**, and System Model Settings also shows incorrectly.
![Image](https://github.com/user-attachments/assets/a9c3480a-993a-4f63-ad26-3d8cccd47c00)

![Image](https://github.com/user-attachments/assets/442be449-6fa9-4a4d-8b68-03c74ad7940c)

![Image](https://github.com/user-attachments/assets/acf8fa65-a914-4bbf-809f-84a177d01332)

![Image](https://github.com/user-attachments/assets/e0ca48e9-d6c4-468c-acf6-71b51d862f5b)

leachli:
I also have the same problem，no matter how I change the template, it won't take effect. I have to manually change the template in the ragflow service container and restart the container to solve the problem

KevinHuSh:
Change it in this format. After alteration, it's only effective for newly signup users.

![Image](https://github.com/user-attachments/assets/8f1a1dc3-2dc7-4a01-961a-241f8aff5d06)

leachli:
> Change it in this format. After alteration, it's only effective for newly signup users.
> 
> ![Image](https://github.com/user-attachments/assets/8f1a1dc3-2dc7-4a01-961a-241f8aff5d06)

I did not modify the configuration here during the first startup. After making the necessary changes, I cleared the volume with 'down - v' and released the 'user_default_1lm' configuration. However, the 'up - d' file did not take effect and was not printed in the log

alice77-zhen:
> alteration
Thanks for your reply. I coded it according to the format, and I registered a new user，but it still doesn't work.

![Image](https://github.com/user-attachments/assets/e2e54ed2-926d-4401-95df-ce747d0f6e96)

![Image](https://github.com/user-attachments/assets/19df7203-9a6a-49f4-8e5c-898390f6dfc7)

KevinHuSh:
What about `conf/service_conf.yaml` in container?
Have you ever restart the ragflow server after the alteration of yaml file?

alice77-zhen:
> What about `conf/service_conf.yaml` in container? Have you ever restart the ragflow server after the alteration of yaml file?
[...]

PR #18185, opened 2026-08-12
fix: seed configured default models for new tenants

liwu96:
### Summary

Materialize models configured through user_default_llm when a tenant is created. The registration paths now create the tenant model provider, credential-matched instance, and model rows required by the current model catalog, then persist the resolved model name and tenant_model ID on the tenant. Configured models from the same factory reuse an instance only when both api_key and base_url match.

The seeding is idempotent, supports shared models with multiple type bits, isolates failures from registration and superuser initialization, and is applied to user registration, account creation, superuser initialization, and admin tenant bootstrap.

### Root cause and scope

The model-catalog change merged in v0.26.0 by PR #15860 (commit 478c9846a1f967b002448cfdf157828daeeeaab2) removed the legacy TenantLLM initialization from tenant creation, but did not replace it with seeding for tenant_model_provider, tenant_model_instance, and tenant_model. New tenants therefore stored only the configured model names, so runtime resolution and System Model Settings could not find the configured defaults.

This fixes that v0.26.0 regression for newly created tenants. It does not change how Docker copies or renders service_conf.yaml.template; the user_default_llm configuration must already be present in the effective configuration inside the running container.

Related to #5838

### Validation

- Ruff checks passed for all changed Python files
- Python 3.13 compileall passed for all changed Python files
- Focused credential-isolation and idempotency behavior harness passed
- git diff --check passed
- Focused pytest was attempted, but local dependency setup was blocked while fetching the pinned graspologic Git dependency from Gitee; focused unit coverage is included for CI


Lynn-Inf (commented):
`model_provider` is already visible to all users in the open‑source version — no need to add a separate default model config for everyone.

liwu96:
Thanks for clarifying. I understand that the provider catalog is globally visible in the open-source version. However, provider visibility does not grant a newly registered tenant access to an administrator-owned `tenant_model_instance` or its credentials: model listing and resolution are still tenant-scoped unless the user has joined the provider owner's tenant.

I also noticed that #18875 now initializes the default model fields as empty for normal user registration. Does this mean that `user_default_llm` is intentionally no longer supported for newly registered users?

If so, should `user_default_llm` be deprecated or its scope clarified in the OSS configuration and documentation? Otherwise, what is the intended way for a newly registered user to use the administrator-configured model instance without duplicating its credentials per tenant?

The goal of this PR was to address the `Provider not found` behavior reported in #5838 after the legacy `TenantLLM` seeding was removed in #15860. I agree that a single system-owned model instance with controlled tenant access would be cleaner than copying credentials for every tenant, if shared system models are the intended architecture.

Lynn-Inf:
Yes, `user_default_llm` has been deprecated and is no longer supported. In the open‑source version, provider instances are **not shared** across tenants — each tenant must configure its own model instances.

The Enterprise Edition supports default model settings at the role level. If you're interested, feel free to reach out for more details.

liwu96:
Thanks for the clarification, @Lynn-Inf. I understand that automatic default-model provisioning through `user_default_llm` is no longer supported and that OSS tenants must configure their own model instances.

I've opened a documentation-only PR, #19321, to update the `user_default_llm` section in `docs/administrator/configurations/configurations.md`. It removes the outdated instructions for enabling defaults for new users, explains tenant-scoped configuration, and links to the UI model setup guide. It also notes the Enterprise role-level defaults and preserves guidance for the TEI Builtin embedding connection settings that are still used.

This should make the intended behavior clear to administrators encountering the old configuration instructions. Please take a look when convenient. Thanks!

Lynn-Inf:
Thanks. Feel free to close this PR if it's no longer needed.

PR #18191, opened 2026-08-13
Fix: reduce model verify cost

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/services/provider_api_service.py
+++ b/api/apps/services/provider_api_service.py
@@ -865,6 +865,7 @@ async def verify_api_key(provider_id_or_name: str, api_key: str | dict, base_url
                 model_verify_result[llm["llm_name"]] = ModelVerifyStatusEnum.FAIL.value
         if any_passed:
             msg = ""
+            break
         else:
             msg = msg or "No model passed verification"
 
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the PR — the intent (stop verifying as soon as one model passes to cut API cost) makes sense, and the `break` is correctly placed.

Two things worth confirming before merge:

1. **Behavioral change: per-model verify status becomes mostly "unknown".** With the early `break`, verification stops after the first passing model, so every remaining model in `factory_llms` is skipped. Callers use `model_verify_result.get(name, UNKNOWN)`, so those skipped models will now show `UNKNOWN` instead of `SUCCESS`/`FAIL` in each model's `verify` extra field. Consequences:
   - Users lose visibility into which individual models actually work.
   - A key that works for the first model (e.g. `BAAI/bge-m3` embedding) but fails for another (e.g. the chat model the user actually wants) now "passes" silently — they'd only notice when using the broken model.
   
   This also applies when `model_info` is provided explicitly: only the first selected model gets verified. Please confirm this trade-off is intended, and consider documenting it — the PR body is currently just "As title."

2. **Reordering the SiliconFlow `llm` array.** Moving the three BAAI embedding models to the front changes the model display/default order for this provider (the first model is now an embedding model rather than a chat model). If anything relies on the first entry being a chat model, this could shift behavior. Please confirm intended.

The JSON change is a clean move (no duplicates, no syntax issues). Minor: no test covers the early-termination behavior — worth adding if there are existing verification tests.

PR #18192, opened 2026-08-13
Fix(parser): raise legacy .doc parsing fidelity via structured IR

xugangqiang:
## Summary

Go's `DOCParser` previously relied on `office_oxide.PlainText()` for legacy `.doc`
files. This change prefers the **structured IR** produced by
`office_oxide.ToIRJSON` (flattened to plain text) and falls back to `ToMarkdown`,
then to `PlainText`. The downstream text contract stays `"text"`.

### Scope correction (important)

`office_oxide`'s `.doc` reader only surfaces **paragraph and (heuristically
detected) heading** lines — it does **not** extract table or list structure from
legacy `.doc`. The table stream's `PAP`/`TAP`/`LSTF` are ignored (tracked
upstream: https://github.com/yfedoseev/office_oxide/issues/115). So this change
normalizes paragraph/heading line breaks; it does **not** recover `.doc` tables
or lists, and should not be described as Tika-parity for those structures. The
`.docx` path (which does emit `Table`/`ListItem`) keeps using the same shared
`flattenDocIR`.

## Approach

- `extractDocText(doc)` in `doc_parser.go` returns the **longest non-empty** view
  among IR-flatten → Markdown → PlainText. This guards against a sparser view
  shadowing a more complete one — the IR flatten can be shorter than `PlainText`
  for some documents, so the previous IR-first fallback could regress text
  completeness. A failure at every stage degrades to the `PlainText` error.
- `flattenDocIR` (new `doc_flatten.go`) turns the IR JSON into plain text. The
  office_oxide IR schema is format-independent, so this function is reused
  unchanged from the existing docx path.
- `ParseWithResult` keeps `OutputFormat="text"` and the same `File` shape, so no
  consumer branch changes.

## Test

The `.doc` integration test (`doc_parser_synth_test.go`) **synthesizes a valid
OLE2/CFB `.doc` entirely in memory** — a minimal WordDocument stream (FIB +
piece-table-referenced CP1252 text) plus a 0Table stream (CLX), assembled per
office_oxide's reader expectations. No binary fixture is committed and no
external converter is required, so the DOC path (`OpenFromBytes(data,"doc")`
through `ParseWithResult`) is exercised end-to-end against dynamically generated
input.

The cgo-free `doc_flatten_test.go` exercises the shared `flattenDocIR` logic
directly (paragraphs, headings, lists, table cells joined with ` | `). Note this
validates the flatten function itself; because office_oxide does not emit
`Table`/`ListItem` for `.doc`, the `.doc` integration test only asserts the
fallback chain yields the expected paragraph text.

## Files

- `internal/parser/parser/doc_parser.go` — `extractDocText` longest-non-empty fallback
- `internal/parser/parser/doc_flatten.go` — `flattenDocIR` (new, shared with docx)
- `internal/parser/parser/doc_flatten_test.go` — cgo-free unit test (new)
- `internal/parser/parser/doc_parser_synth_test.go` — in-memory `.doc` cgo test (new)

## Verification

`bash build.sh --test ./internal/parser/parser/...` passes with no regression.


xugangqiang:
Summary of where this PR stands and the plan:

**What this PR does (and does correctly)**
- Switches `.doc` extraction from `PlainText()` only to an `IR → Markdown → PlainText` fallback that returns the **longest non-empty** view. This prevents the regression where the IR flatten could be *shorter* than `PlainText` for some documents.
- Keeps `OutputFormat="text"` and the `File` shape unchanged, so no consumer branch changes.
- The PR description and `doc_parser.go` comments now honestly state that office_oxide's `.doc` reader only surfaces paragraph/heading lines and does **not** extract table or list structure (tracked upstream: https://github.com/yfedoseev/office_oxide/issues/115).

**What it deliberately does NOT do**
- It does **not** recover `.doc` tables/lists. That capability is missing in office_oxide itself, not in our glue code. I verified empirically: real `.doc` files (including a WPS `.doc` that contains a table) open fine, but `to_ir_json()` yields no `Table`/`ListItem` — the table is flattened to tab-separated text.

**Decision: hold for upstream, then enhance and merge**
The real fidelity gain (structured table/list text for `.doc`) depends on office_oxide fixing #115. So the plan is:
1. Keep this PR open as the honest, low-risk normalization scaffolding.
2. Once office_oxide #115 lands (`.doc` IR emits `Table`/`ListItem`), we will update the `.doc` path to actually consume that structure and then merge.

**Minor nits to address when we touch this again** (not blocking the hold):
- Soften the PR title to match the honest scope (it currently says "raise … fidelity via structured IR").
- `doc_flatten.go` comments still say the flatten "mirrors Tika … for legacy .doc files" / recovers table content — inconsistent with the D-scoped `doc_parser.go` comments; should be corrected.
- `buildSyntheticDoc` in `doc_parser_synth_test.go` silently truncates the body if it exceeds the 86-byte WordDocument text capacity; add a `t.Fatalf` guard.

Net: this PR is safe and honest as-is; the meaningful change is gated on upstream #115.

PR #18194, opened 2026-08-13
Fix: The filenames in the parsing result log pages are not displayed.

cike8899:
### Summary

Fix: The filenames in the parsing result log pages are not displayed.


wangq8 (commented):
Reviewed. The root cause is correctly addressed: the pipeline log sheet never passed the file `name` (nor `size`/`created_at`) into the dataflow-result route, so `DocumentHeader` had nothing to render.

- Passing `DocumentName`/`DocumentSize`/`DocumentCreatedAt` through the search params and reconstructing `documentInfo` in agent mode fixes the missing filename.
- The `encodeURIComponent` change is important — filenames containing `&`, `=`, `#`, spaces or non-ASCII would otherwise corrupt the query string (and `URLSearchParams.get` decodes it transparently).

Minor (non-blocking):
- The agent-mode `documentInfo` is built with only `name`, `size`, `create_date` and cast `as IKnowledgeFile`. It works because only those three fields are consumed in agent mode, but the cast hides the missing fields; a narrower type or explicit partial would be safer against future field access.
- `new Date(Number(documentCreatedAt) * 1000).toISOString()` assumes `created_at` is a Unix-seconds value (which the Go upload response guarantees). The ternary guards empty string but not a non-numeric value, which would make `.toISOString()` throw. Low risk given the backend always sends a number, but a `Number.isFinite` guard would be more defensive.
- The `console.log` removal is unrelated to this fix (fine as cleanup, just flagging).

Thanks for the fix!

PR #18196, opened 2026-08-13
Fix agentic chat low/medium cannot answer the table content

wangq8:


wangq8 (commented):
Thanks for the fix — this correctly targets the root cause. Low/medium modes route through `hybrid_search` directly (`direct_search` and `decompose_and_search`), and `structured_query` is only exposed in `ultra` (`harness/config.py`), so table/SQL KBs were indeed unreachable in those modes. I verified that table KBs produce per-row chunks (`tokenize_table` sets `content_with_weight` + `doc_type_kwd="table"`) and that Infinity retrieval searches per-KB tables, so including `sql_kbs` in `target_ids` genuinely surfaces the rows.

A few things to consider before merge:

1. **The change is global, not gated to low/medium.** `hybrid_search` is also called by `pre_search` (decomposition modes: medium/high/ultra) and is an available tool in high/ultra. High/ultra already answer table questions via `structured_query`/`structured_retrieve` (SQL aggregation through `use_sql`). After this change those modes will also receive raw table rows from `hybrid_search`, which can change answer quality (raw rows vs. SQL-aggregated results) or duplicate evidence. Worth confirming that's intended and smoke-testing a table question in high/ultra.

2. **`bm25_search` still excludes `sql_kbs`.** It sits in the low/medium `available_tools` list and uses `target_ids = kb_ids or tools.kb_ids`. This doesn't block the fix (the orchestrators call `hybrid_search` directly), but for consistency — and any code path where `bm25_search` is the fallback — it'd be cleaner to extract a shared helper like `_bound_kb_ids(tools, kb_ids)` and use it across `hybrid_search`, `vector_search`, and `bm25_search`.

3. **Attribute safety (minor).** `hybrid_search` now reads `tools.sql_kbs` unconditionally. `RAGTools` always initializes `self.sql_kbs = []`, so current callers are fine, but `hybrid_search` is registered as a generic tool (`register_tool` in `harness/tools/__init__.py`). A `getattr(tools, "sql_kbs", [])` guard would make it robust against other tool-like objects. (Existing functions already assume `tools.kb_ids`/`tools.embed_mdl`, so this is consistent-but-fragile rather than a new break.)

4. **No test coverage.** The only agentic test (`test/unit_test/rag/advanced_rag/test_agentic_rag.py`) doesn't touch search. A small unit test asserting that when `tools.sql_kbs` is non-empty, the retrieval `target_ids` include both `kb_ids` and sql-kb ids (plus dedup) would lock this in.

Overall the change is sound for its stated purpose; the main thing I'd want confirmed is point 1 (high/ultra behavior).

KevinHuSh (approved):

PR #18197, opened 2026-08-13
fix(agent): accept 1D files wire shape in chat/completions

euvre:
### Summary

The agent chat front-end posts uploaded files as a **1D** list of file dicts (`files: [{id, name, mime_type, ...}]`, `web/src/pages/agent/chat/use-send-agent-message.ts`), but the Go handler declared the field as **2D** (`[][]map[string]interface{}`). `encoding/json` then rejected every chat message carrying an attachment with:\n\n`code=101: Invalid request: json: cannot unmarshal object into Go struct field agentChatCompletionsRequest.files of type []map[string]interface {}`\n\nThe canvas run never started, so **sending a message with an uploaded image in an agent chat produced no response at all** (the message appears in the chat, the assistant never replies).\n\nReproduction before the fix: agent canvas `Begin → Agent (vision model) → Message`, upload an image, send \u63a8\u8ff0\u56fe\u7247\u5185\u5bb9 → request 400s in ~1ms and the UI hangs with no reply. After the fix the same request streams the full workflow (`workflow_started → node events → message_end → [DONE]`) and the uploaded image reaches `sys.files` as a base64 data URI for the vision model via `ParseAgentUploads`.\n\n**Fix:** introduce a custom `agentFiles` unmarshaler that normalizes both wire shapes to the 1D list:\n- 1D `[{...}]` (agent chat) passes through;\n- 2D `[[{...}]]` (dataflow debug, `use-run-dataflow.ts`) takes the first inner list, matching Python's `files[0]` access at the `queue_dataflow` call site (`agent_api.py`).\n\n**Tests:**\n- New `TestAgentChatCompletions_Files1DDeserialized` pins the 1D chat wire shape (regression test for this bug).\n- Existing 2D coverage (`TestAgentChatCompletions_FilesDeserialized`, `TestAgentChatCompletions_DataFlowReturnsChunks`, `TestAgentChatCompletions_EmptyFilesNil`) still passes — `bash build.sh --test ./internal/handler/`.\n- Vision path unit tests pass: `TestBuildMessagesWithImages_*`, `TestExtractDataImages_*`, `TestCollectSysFiles_*`, `TestFileService_ParseAgentUploads_*`.

### DIFF
--- a/internal/handler/agent.go
+++ b/internal/handler/agent.go
@@ -643,15 +643,11 @@ func respondWithDebugResult(c *gin.Context, result *task.PipelineResult, err err
 // (id/name/created_by); the raw bytes are fetched from the per-user downloads
 // bucket via the file service. When no usable file is present it returns a
 // default name and a nil slice so the pipeline can still run file-less.
-func (h *AgentHandler) extractChatDebugFile(ctx context.Context, files [][]map[string]interface{}, user *entity.User) (string, []byte) {
+func (h *AgentHandler) extractChatDebugFile(ctx context.Context, files []map[string]interface{}, user *entity.User) (string, []byte) {
 	if len(files) == 0 {
 		return "debug", nil
 	}
-	fileList := files[0]
-	if len(fileList) == 0 {
-		return "debug", nil
-	}
-	fd := fileList[0]
+	fd := files[0]
 	name, _ := fd["name"].(string)
 	id, _ := fd["id"].(string)
 	if id == "" {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Reviewed. The fix is sound and well-tested.

- The custom `agentFiles` unmarshaler correctly normalizes both wire shapes: 1D `[{...}]` (agent chat) passes through, 2D `[[{...}]]` (dataflow debug) takes `twoD[0]`, matching Python's `files[0]` at the `queue_dataflow` call site.
- `extractChatDebugFile` now takes the 1D list, and its single call site (the dataflow debug branch) still receives the normalized list, so that path is preserved.
- The new `TestAgentChatCompletions_Files1DDeserialized` regression test pins the 1D chat wire shape, and existing 2D tests remain valid.
- Assignability of the named `agentFiles` type to `[]map[string]interface{}` params is correct (identical underlying type, unnamed param type).

Minor (non-blocking): the explicit `string(data) == "null"` check in `UnmarshalJSON` is redundant — `json.Unmarshal("null", &oneD)` already yields nil without error. Harmless, just slightly more code than needed.

Looks good to me. Thanks!

PR #18198, opened 2026-08-13
fix(parser): support .msg parsing and make email attachments retrievable

xugangqiang:
## Summary

- Add Go `EmailParser` support for Outlook `.msg` (OLE2/CFB) files via the `gomsg` library, in addition to the existing `.eml` (RFC 5322) support. The `.msg` hard-error is gone; emails with `.msg` attachments are now ingested end-to-end.
- Re-chunk email attachments into retrievable text (user-oriented). Each attachment is re-parsed by its file extension through the shared parser registry and folded back into the same document, so attachment content becomes searchable. This mirrors Python's legacy `rag/app/email.py`. Binary attachments (images/audio/video/folders) are skipped by design.
- Restore a corrupted `sample.msg` test fixture and add guards so binary fixtures are never mangled again (`.gitattributes` marks `*.msg` binary; `check_files.py` skips NUL-byte files). Also made `check_files.py` ruff-clean.

## Implementation notes

- **Nested email re-chunk uses the top-level field configuration.** A nested `.eml`/`.msg` attachment is now parsed with the same `fields` (including `body`) as the parent email, instead of via a fresh unconfigured `EmailParser` that would only emit `metadata` and index `"metadata:{...}"` garbage. This was the key regression caught in review.
- **Embedded `.msg` recursion.** gomsg exposes `Attachment.EmbeddedMessage()` for embedded messages (no raw bytes), so embedded message bodies are now surfaced as retrievable text attachments rather than silently dropped. (No separate design doc — the behaviour is described in code comments.)
- **Re-chunk is bounded.** Recursion stops at one level of nesting (no attachments-of-attachments), and attachments larger than 32 MiB are not re-parsed inline, to avoid a single huge/heavy attachment (PDF/OCR) stalling ingestion.
- **Date format aligned to Python.** `.msg` dates now use Go's `-0700` layout (`2018-03-24 00:06:29+0800`) to match `extract_msg`'s `strftime("%z")` (no colon), instead of the previous `-07:00` form.

## Dependency: gomsg is now pinned to a controlled fork

`github.com/AkmalOt/gomsg` has **no tagged release** (only a pseudo-version) and is a single-maintainer repo, so it has been pinned to our controlled fork via a `replace` directive:

```
replace github.com/AkmalOt/gomsg => github.com/xugangqiang/gomsg v0.0.0-20260407083308-985c3a1a76b7
```

The fork (`xugangqiang/gomsg`) is at the exact same commit (`985c3a1a76b7`), so the module content and `go.sum` hash are unchanged — checksum verification is unaffected, only the download source is now under our control and immune to upstream deletion. Optional follow-up: move the fork to the `infiniflow` org and cut a real semver tag if a more permanent home is preferred.

## Test plan

- `bash build.sh --test ./internal/parser/parser/` passes, including new tests:
  - `TestEmailParser_NestedEMLAttachmentRechunk` — locks the nested-email garbage-indexing regression.
  - `TestEmailParser_RechunkAttachmentTextPath` — rechunk indexes text attachments, skips binary, no leak.
  - plus the `.msg` fixture tests and existing attachment re-chunk tests.


xugangqiang:
## Notes for maintainers

This PR adds `.msg` (Outlook / OLE2-CFB) support to the Go `EmailParser` and makes email attachments retrievable. A few points worth flagging for review:

### 1. Design rationale
- **`.msg` was a hard failure before.** Outlook `.msg` files (and emails carrying `.msg` attachments) errored out in the Go ingestion path. They are now parsed end-to-end via the `gomsg` library (pure-Go OLE2/CFB reader).
- **Attachments become retrievable text (user-oriented).** Each attachment is re-parsed by its file extension through the shared parser registry and folded back into the *same* email document, so searching an email surfaces its attachment content. Binary attachments (images/audio/video) are skipped by design.
- **Nested / embedded emails are handled.** A nested `.eml` is re-chunked using the parent email's configured `fields` (this fixes a regression where a fresh unconfigured parser only emitted `metadata` and indexed garbage like `"metadata:{...}"`). Embedded `.msg` attachments expose no raw bytes via `gomsg`, but `Attachment.EmbeddedMessage()` is parsed and surfaced as retrievable text instead of being silently dropped.
- **Bounded.** Re-chunk recurses at most one level (no attachments-of-attachments) and skips attachments larger than 32 MiB, so a heavy attachment (e.g. a PDF needing OCR) can't stall ingestion.
- **Date format aligned to Python.** `.msg` dates use Go's `-0700` layout to match `extract_msg`'s `strftime("%z")` (no colon).

### 2. Intentional deviation from the Python `email` path
The current Python `rag/app/email.py` chunks each attachment as an **independent top-level chunk** (`main_res + attachment_res`, via `naive_chunk`) — attachments live as separate retrieval units from the email body.

The Go path deliberately **folds** attachment content back into the parent email document (extra `JSON` items / appended text) so the email and its attachments are one retrievable unit. This is an explicit, user-oriented design choice, not an oversight: a user searching an email generally expects attachment text to come back *with* the email.

Note that Python ingestion is being deprecated in favor of the Go engine, so the Go path is the de-facto source of truth here; we are intentionally **not** byte-matching Python's separate-chunk behavior. The choice is documented in code comments (`email_parser.go`).

### 3. Dependency pin: `gomsg` → controlled fork
`github.com/AkmalOt/gomsg` has **no tagged release** (only a pseudo-version) and is a single-maintainer repo. To avoid a supply-chain risk from upstream deletion/change, it is pinned to our controlled fork via a `replace` directive:

```
replace github.com/AkmalOt/gomsg => github.com/xugangqiang/gomsg v0.0.0-20260407083308-985c3a1a76b7
```

The fork is at the exact same commit (`985c3a1a76b7`), so module content and the `go.sum` hash are unchanged — checksum verification is unaffected, only the download source is now under our control.
*Optional follow-up:* move the fork to the `infiniflow` org and cut a real semver tag if a more permanent home is preferred.

### Test plan
`bash build.sh --test ./internal/parser/parser/` passes, including new regression tests `TestEmailParser_NestedEMLAttachmentRechunk` and `TestEmailParser_RechunkAttachmentTextPath`.

PR #18200, opened 2026-08-13
fix: align shared search result total

buua436:

Align the shared search retrieval API's `total` with the number of chunks actually returned, ensuring consistent result counts between shared and regular search pages.

### Type of change

### DIFF
--- a/api/apps/restful_apis/bot_api.py
+++ b/api/apps/restful_apis/bot_api.py
@@ -462,6 +462,7 @@ async def retrieval_test_embedded(tenant_id=None):
             enrich_chunks_with_document_metadata(ranks["chunks"], metadata_fields)
 
         ranks["labels"] = labels
+        ranks["total"] = len(ranks["chunks"])
 
         return get_json_result(data=ranks)
 


wangq8 (commented):
Thanks for the fix! Reviewed against `main` and the regular search path.

**Verification**
- The change correctly mirrors the convention already used in `dataset_api_service.py` (`search` and `search_datasets`), which both set `ranks["total"] = len(ranks["chunks"])` right before returning.
- Placement is correct: it runs after the KG chunk `insert(0, ck)`, the `pop("vector")` loop, and `enrich_chunks_with_document_metadata`, so `len(ranks["chunks"])` reflects exactly the chunks returned in the response.
- `ranks["chunks"]` is always a list from `settings.retriever.retrieval`, so `len()` is safe even for the empty case.

**Non-blocking observation**
- `total` now means "chunks returned in this response" rather than the retriever's `filtered_count` (which is a candidate-window count). That's consistent with the regular search page, so it's the right call for the stated goal — just noting that clients relying on `total` as a true pagination total should treat it as a page count instead. This matches existing regular-search behavior, so no change needed here.

LGTM. ✅

PR #18202, opened 2026-08-13
fix(web): regenerate a single-chat answer as a new turn

dcc123456:
### Summary

fix(web): regenerate a single-chat answer as a new turn

Clicking regenerate on a question truncated every message after it and blanked the existing answer, so earlier replies disappeared from the transcript for what the user reads as "ask this again".

Re-ask the same question as a fresh turn instead: copy its content and attachments onto a new message, append it with the usual placeholder, and send the full history. Nothing already on screen is removed, and the regenerated answer lands at the bottom next to the repeated question.

truncateAfterMessage was the only caller-facing use of the store's truncation, so it goes with it. Scope is the single-chat path; the multi-model box and useSendSingleMessage keep their old behaviour.

### DIFF
--- a/web/src/pages/next-chats/chat-stream/store.ts
+++ b/web/src/pages/next-chats/chat-stream/store.ts
@@ -44,7 +44,6 @@ export type ChatStreamState = {
   failStream(conversationId: string, pendingInput: string): void;
   consumePendingInput(conversationId: string): void;
   removeMessageById(conversationId: string, messageId: string): void;
-  truncateAfterMessage(conversationId: string, messageId: string): void;
   endStream(conversationId: string): void;
   stopStream(conversationId: string): void;
   removeSessions(conversationIds: string[]): void;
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix! Reviewed the diff against the current `main` and the surrounding hooks/store code.

**Verification**
- `truncateAfterMessage` is now only referenced in `chat-stream/store.ts` and `use-send-chat-message.ts`; both references are removed together, so no dangling imports/usages remain. `useSendSingleMessage` and the multi-model box use their own `removeMessagesAfterCurrentMessage` and are unaffected. ✅
- `Message` is exported from `@/interfaces/database/chat` and `IMessage extends Message`, so the new `(message: Message)` signature is type-safe (callers pass `IMessage`). ✅
- `uuid` is already imported in this file, and `appendQuestion(conversationId, message: IMessage)` accepts the new `questionMessage` shape. ✅
- The new `appendQuestion` + `sendMessage` flow mirrors the normal `handlePressEnter` path, so the regenerated answer gets the usual placeholder and streams to the bottom correctly.

**Non-blocking observations**
- `sendMessage` falls back to the store's `messages` captured at render time, which (correctly) does not yet include the just-appended question — so the server receives full history + the re-asked question exactly once. This is right, but worth a quick sanity check that a very rapid double-click can't append twice; the new `if (isStreaming) return;` guard largely covers this.
- The `files` of the original message are re-sent verbatim. This should be fine for already-uploaded `UploadResponseDataType` refs; just confirm the backend tolerates re-submitting them.
- Minor: the inline comment in `sendMessage` ("e.g. regenerating the first question must truncate history") now only describes the multi-model path, since the single-chat regenerate no longer truncates. Consider updating it to avoid confusion.

LGTM overall — the "re-ask as a new turn" behavior is a cleaner UX and the dead truncation path removal is correct. ✅

PR #18204, opened 2026-08-13
chore(web): drop unused somark purchaseUrl locale strings

dcc123456:
### Summary

chore(web): drop unused somark purchaseUrl locale strings

The purchaseUrl entry is not referenced anywhere in the frontend, so the promotional purchase links only added translation upkeep. Also reflows two locale strings to match the formatter output.


wangq8 (commented):
Thanks for the cleanup! Reviewed and verified against the codebase.

**Verification**
- Confirmed `purchaseUrl` is referenced nowhere in `web/src` outside the three locale files (`en.ts`, `ja.ts`, `zh.ts`), so these strings are genuinely dead and safe to remove.
- The two reflowed strings (`mwsApiUrlPlaceholder` in `en.ts` and `planTip` in `zh.ts`) are formatting-only changes to match the formatter output — no functional impact.

**Notes**
- No tests needed for locale string removal.
- No regressions or security concerns; this is a pure dead-code/formatting cleanup.

LGTM. ✅

PR #18205, opened 2026-08-13
fix: clear stale tenant model ids

01JohnMa:
### Summary

Clear the paired tenant model ID when a model selection is explicitly cleared in a whitelisted API request.

Chat settings currently merge the fetched chat object into the update payload. Clearing the rerank selector sets `rerank_id` to an empty string, but the previously fetched `tenant_rerank_id` remains in the payload. The backend then resolves that stale tenant model ID back into the rerank model name, so reranking stays enabled after the user saves the cleared selection.

This change makes `addTenantParams` set the paired `tenant_*_id` to `null` when the corresponding model field is explicitly `""`, `null`, or `undefined`. Requests that omit the model field continue to preserve an existing tenant model ID.

### User impact

Users can clear the rerank model in chat settings without the previous tenant-scoped model ID silently restoring it. The same synchronization rule applies to other model selectors handled by `addTenantParams`.

### Validation

- `npm test -- --runInBand src/utils/tests/llm-util.test.ts` (20 tests passed)
- `npx oxlint src/utils/llm-util.ts src/utils/tests/llm-util.test.ts`
- `npx oxfmt --check src/utils/llm-util.ts src/utils/tests/llm-util.test.ts`
- `git diff --check`

### DIFF
--- a/web/src/utils/llm-util.ts
+++ b/web/src/utils/llm-util.ts
@@ -199,17 +199,26 @@ export function addTenantParams(data: any, url?: string): any {
     return data;
   }
 
-  const llmList = getCachedLlmList();
-  if (!llmList) return data;
-
   // Handle arrays
   if (Array.isArray(data)) {
     return data.map((item) => addTenantParams(item, url));
   }
 
   const newData = { ...data };
+  const llmList = getCachedLlmList();
 
-  // Iterate through model parameters and add corresponding tenant parameters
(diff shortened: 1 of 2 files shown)


01JohnMa:
Superseded by #18210, which contains the same commit and diff under the renamed `fix/clear-stale-tenant-model-ids` branch.

01JohnMa:
Closing in favor of #18210.

PR #18206, opened 2026-08-13
fix(web): seed new conversation from server prologue before appending question

dcc123456:
### Summary

fix(web): seed new conversation from server prologue before appending question

A freshly created conversation is persisted under a server-generated id, so the prologue seeded on the temporary id never carries over and the first answer rendered without it. Hydrate the real session id from the server list before appending the question, so the ordering doesn't trip the streaming authority check and wipe the just-appended question and placeholder.

### DIFF
--- a/web/src/pages/next-chats/hooks/use-send-chat-message.ts
+++ b/web/src/pages/next-chats/hooks/use-send-chat-message.ts
@@ -7,6 +7,7 @@ import {
 } from '@/hooks/logic-hooks';
 import { useGetChatSearchParams } from '@/hooks/use-chat-request';
 import { IMessage } from '@/interfaces/database/chat';
+import { buildMessageListWithUuid } from '@/utils/chat';
 import notification from '@/utils/notification';
 import { trim } from 'lodash';
 import { useCallback, useEffect, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix — the root cause is correctly identified: for a freshly created conversation the store session is keyed on the server-generated id, so the prologue seeded on the temporary id is orphaned and the first render drops it. Hydrating from `currentMessages` (which carries the server-seeded prologue) *before* `appendQuestion` is the right ordering, and the `currentMessages.length > 0` guard correctly scopes this to the new-conversation path (existing conversations return an empty list).

A few notes:

1. **Ordering vs. authority check** — correct. `hydrateFromServer` rejects a server list shorter than the local one, so hydrating *after* `appendQuestion` would either wipe the just-appended question/placeholder or be rejected. Doing it first leaves the session at `[prologue, question, placeholder]`, so the later `single-chat-box` hydration (where the server still returns only the prologue) is correctly rejected instead of truncating. 👍

2. **Merge conflicts** — `gh pr view` reports `mergeable: CONFLICTING`. Please rebase on `main` and resolve before merge.

3. **Missing test** — this store has subtle length/authority-guard logic. A regression test covering "new conversation hydrates the server prologue before appending the question, and the question isn't wiped by a subsequent stale server list" would be valuable. None is included here.

4. **Minor** — `buildMessageListWithUuid` also `omit`s `reference`; harmless for the prologue, but the helper does more than its name suggests. Non-blocking.

LGTM once the conflicts are resolved.

PR #18208, opened 2026-08-13
test: cover stale conversation switching

01JohnMa:
## Summary

Add focused frontend regression coverage for stale conversation messages during chat session switching.

## Why

When the URL already points to a newly created conversation while the child still receives the previous conversation prop, the old messages must not be hydrated into the new session. This reproduces the race fixed by #18140.

The production guard is already present on main; this PR adds the missing component-level regression test.

## Validation

- npm test -- --runInBand src/pages/next-chats/chat/chat-box/single-chat-box.test.tsx
- npx oxfmt --check src/pages/next-chats/chat/chat-box/single-chat-box.test.tsx
- npx oxlint src/pages/next-chats/chat/chat-box/single-chat-box.test.tsx
- The stale-session test fails when the #18140 guard is removed and passes with the guard restored.

Related: #18140

### DIFF
--- a/web/src/pages/next-chats/chat/chat-box/single-chat-box.test.tsx
+++ b/web/src/pages/next-chats/chat/chat-box/single-chat-box.test.tsx
@@ -1,0 +1,123 @@
+import { render, waitFor } from '@testing-library/react';
+import React from 'react';
+
+import { MessageType } from '@/constants/chat';
+import { SingleChatBox } from './single-chat-box';
+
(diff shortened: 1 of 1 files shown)


01JohnMa:
Superseded by #18211, which contains the same commit and diff under the renamed `test/stale-session-switch` branch.

01JohnMa:
Closing in favor of #18211.

PR #18212, opened 2026-08-13
Fix: The operator field of the compiler operator is a required field.

cike8899:
### Summary
Fix: The operator field of the compiler operator is a required field.

### DIFF
--- a/web/src/components/compilation-template-form-field.tsx
+++ b/web/src/components/compilation-template-form-field.tsx
@@ -43,6 +43,7 @@ export function CompilationTemplateFormField({
       }}
       className="pb-4"
       horizontal={horizontal}
+      required
     >
       {(field) => (
         <SelectWithSearch
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -963,6 +963,7 @@ Paragraphs:
       randomSeedMessage: 'Random seed is required',
       entityTypes: 'Entity types',
       compilationTemplate: 'Operator',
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
Makes the compiler operator field required — reasonable and matches the intent. The new `compilationTemplateRequired` key is correctly placed under `knowledgeConfiguration` (next to `compilationTemplate`), and the `useFormSchema` hook resolves it via `useTranslation()` in the right namespace. A few notes:

1. **Merge conflict with #18216.** Both PRs rewrite `FormSchema` in `web/src/pages/agent/form/compilation-form/index.tsx` and touch the same locale blocks (`en.ts`/`zh.ts`). This PR keeps `plan`, #18216 replaces `plan` with `mode`. Coordinate merge order; the second one to land will need a rebase.

2. **`FormSchema` recreated every render.** Moving it into `useFormSchema()` means a new schema (and `zodResolver`) is built on each render. Functionally fine with react-hook-form, but consider wrapping in `useMemo` to avoid churn.

3. **Hardcoded `required` on `CompilationTemplateFormField`.** It's currently only used in `compilation-form` (verified), so this is fine today — but if the component is ever reused for an optional field, the unconditional `required` will be wrong. Consider making it a prop defaulting to `false`.

4. Minor: the `zh.ts`/`en.ts` diff also carries unrelated formatter reformats (`missing_name`, `ttsTip`), harmless noise.

LGTM once the merge-order conflict with #18216 is acknowledged.

PR #18213, opened 2026-08-13
Fix: Display the search page chunks in plain text format

cike8899:
### Summary
Fix: Display the search page chunks in plain text format


wangq8 (commented):
The core change (render search-page chunks as plain text instead of markdown) is fine — `sanitizeHtmlWithImagesAsText` goes through DOMPurify, so the `dangerouslySetInnerHTML` path is safe, and the `[&_em]` styling keeps highlights visible. A few concerns:

**1. Scope creep (main concern).** The PR title is "Display the search page chunks in plain text format", but the diff also bundles several unrelated changes:
- `IReferenceChunk.positions`: `number[]` → `number[][]` (in `chat.ts`)
- `document_metadata` added to `ITestingChunk`
- Removal of ~30 color tokens from `tailwind.config.js` and ~50 CSS variables from `tailwind.css`
- New `.bg-card-soft` utility + `data-[selected='true']` styling in `select-with-search.tsx`

I checked that the removed tokens (`colors-text-*`, `background-badge`, `background-note`, `bg-group`, `bg-department`, `colors-background-inverse-*`, etc.) are no longer referenced anywhere in `web/src`, so the cleanup is safe — but it makes the PR much harder to review and bisect. Please split the tailwind/css cleanup (and ideally the `positions`/`document_metadata` type fixes) into a separate PR.

**2. Minor inconsistency with #18215.** This adds `.trim()` to `content_with_weight` (with a copied "Keep whitespaces?" comment), while #18215 removes `.trim()` specifically to preserve leading whitespace. Decide whether leading/trailing whitespace should be preserved on the search page too, for consistency.

**3. Minor lint/format.** `data-testid={` has trailing whitespace (line in `select-with-search.tsx`), likely a formatting slip — `oxfmt` should catch it.

Otherwise the plain-text rendering and highlight styling look correct.

PR #18214, opened 2026-08-13
fix(web): abort the stream when its in-flight question is deleted

dcc123456:
### Summary

fix(web): abort the stream when its in-flight question is deleted

The chat stream lives at module level, so deleting the question that is currently being answered left the reader producing tokens for a question that no longer exists, and its trailing flush could slice off whatever message had become last. Close the session at delete time and guard the flush on isStreaming.

### DIFF
--- a/web/src/pages/next-chats/chat-stream/store.ts
+++ b/web/src/pages/next-chats/chat-stream/store.ts
@@ -240,6 +240,12 @@ export const useChatStreamStore = create<ChatStreamState>()(
           (state) => {
             const previous = state.sessions[conversationId];
             if (!previous) return state;
+            // A session that is no longer streaming has no trailing placeholder
+            // to fill in. The reader's final flush runs before endStream, so
+            // this also guards the case where the in-flight question (and its
+            // placeholder) were just deleted: without it, the flush would slice
+            // off whatever message is now last.
+            if (!previous.isStreaming) return state;
             return {
               sessions: {
                 ...state.sessions,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Nice catch and a well-reasoned fix. I verified the invariants in the store/utils:

- `buildQuestionAndPlaceholder` gives the question and the assistant placeholder the **same id** (because `buildMessageUuid` returns `message.id` and both carry that `id`), so `messages.filter(x => x.id !== messageId)` correctly removes both — no orphaned placeholder.
- The in-flight question is always second-to-last (`at(-2)`) with the placeholder last, so `at(-2)?.id === messageId` correctly identifies the streaming question.
- The new `if (!previous.isStreaming) return state` guard in `applyAnswer` prevents the reader's trailing flush from `slice(0, -1)`-ing the wrong (now-last) message after the delete.

A few suggestions:

1. The comment "share an id" is accurate but subtle — consider a one-line note that `buildMessageUuid` returns `message.id` (not a fresh uuid) for both, so the shared-id invariant is easy to spot.
2. `removingStreamedQuestion` relies entirely on the pair-sharing-id invariant; a defensive `session.messages.at(-1)?.id === messageId` check (or a small helper `isInFlightQuestion(session, messageId)`) would make the intent robust if that invariant ever changes.
3. The `abort()` runs before the `set()` that clears `isStreaming`; today the reader's `finally` runs as a microtask after the synchronous `set()`, so the guard covers the race, but it would be worth a unit test covering "delete the in-flight question mid-stream" to lock the behavior (I don't see store tests for this path).

PR #18216, opened 2026-08-13
feat: support entity and topic wiki modes

buua436:

Replace the Wiki compilation `plan` boolean with an explicit `mode`:

- `entity`: Generate one Wiki page per entity or concept.
- `topic`: Let the LLM group related entities or concepts by topic.

Update the frontend configuration, Python internal parameter flow, template validation, and default configuration accordingly.

### Type of change


wangq8 (commented):
Thanks for this clean rename from a boolean `plan` to an explicit `mode`. The change is consistent across backend, validation, and frontend, and the integration test was updated. A few concerns before this is ready:

**1. Backward compatibility / migration (important).** `_wiki_load_mode` only reads `mode_kwd` and ignores the legacy `plan_kwd` field that previous builds persisted in the `wiki_mode_meta` row. As a result:

- KBs that previously recorded a mode via `plan_kwd` ("true"/"false") will now read back `None`, so mode-change detection (`previous_mode != mode`) silently never fires for them.
- More seriously, existing saved wiki templates still carry `config.plan` (boolean) with no `config.mode`. In `run_wiki_incremental` the mode-resolve loop only accepts `cfg.get("mode")`, `_wiki_load_mode` returns `None` for legacy rows, and then the new guard `raise ValueError("Wiki template mode must be either 'entity' or 'topic'")` will fire — breaking the next compilation for any KB whose template predates this change.

Suggested fix: in `_wiki_load_mode`, also read `plan_kwd` and map `true`→`"topic"` / `false`→`"entity"`; and/or default `mode` to `"entity"` when the template has a `plan` value but no `mode`. A small data migration for saved templates would also avoid the frontend "empty mode" fallback below.

**2. Legacy `plan` → `mode` mapping on the frontend.** `buildConfigFromBuiltin` and `transformDetailToForm` now default `mode` to `""` when a template lacks `mode`. Combined with the new `superRefine` requirement, a legacy template that had `plan: true` (topic mode) won't auto-map to `"topic"` and will force a manual re-selection, silently losing its previous semantics. Consider mapping `plan === true` → `"topic"` / `false` → `"entity"` for legacy data.

**3. Merge conflict with #18212.** Both PRs rewrite `FormSchema` in `web/src/pages/agent/form/compilation-form/index.tsx` and touch the same locale blocks in `en.ts`/`zh.ts`. Please coordinate merge order; whichever lands second will need a rebase.

Minor: the `zh.ts` diff also carries unrelated formatter reformats (`missing_name`, `ttsTip`) — harmless, but noise.

wangq8 (commented):
The `plan` boolean → explicit `mode` enum direction is good, but I found one blocking frontend bug and several behavior/backward-compat concerns worth resolving.

**Blocking**

1. `web/src/pages/agent/form/compilation-form/index.tsx` — `t` is undefined in the `CompilationForm` component body. This PR removed `const { t } = useTranslate('setting')`; only `useFormSchema()` now calls `useTranslation()`. The JSX still calls `t('setting.wikiMode')`, `t('setting.entityMode')`, etc., so this won't typecheck (`Cannot find name 't'`) and will throw at render. Add `const { t } = useTranslation();` (or `useTranslate('setting')`) back in the component body.

**Behavior changes not in the description (please confirm intent)**

2. LLM resolution changed: `map_llm_id` and `kb_chat_llm_id` now come from the pipeline Compiler's `llm_id` (`_pipeline_compiler_llm_id`) instead of the template's `llm_id`. This is a real change for any KB whose compilation-template `llm_id` differs from the pipeline Compiler's `llm_id`, and it isn't mentioned in the PR title/description. If intentional, document it and add a test; otherwise keep resolving from the template config.

3. `_validate_wiki_eligible_docs` raises `ValueError` when a doc has no `pipeline_id`, but `_wiki_eligible_docs` explicitly supports docs eligible via `parser_config` alone (no pipeline). That path now hard-fails. Fall back to the template's `llm_id` for pipeline-less docs instead of erroring.

4. `_validate_wiki_eligible_docs` also errors when eligible docs use more than one template, whereas the previous code used "first template wins" for the KB-wide model. This is a stricter constraint that can break existing multi-template KBs — please confirm it's intended.

**Backward compatibility / migration**

5. `plan_kwd` → `mode_kwd` rename: existing KBs that recorded a mode via `plan_kwd` will now read `None` from `_wiki_load_mode`, so a mode change (note the builtin default also flips from `plan: yes` to `mode: entity`) won't trigger the state-reset path and can mix old/new page structures. Consider reading the legacy `plan_kwd` field for migration.

6. Existing custom templates (and `validate_template_payload`) with `plan: yes/no` but no `mode` now hard-fail: `run_wiki_incremental` raises `ValueError` (`candidate_mode not in ("entity","topic")`). Please map legacy `plan` → `mode` (yes→topic, no→entity) or default it for existing templates.

**Frontend nit**
7. `buildConfigFromBuiltin`/`transformDetailToForm` default `mode` to `''` while `DefaultTemplateValues` uses `'entity'`. A wiki template without an explicit `mode` renders an empty select and fails the new `superRefine` check — consider defaulting to `'entity'` consistently.

The rest (validation, locale strings, `wiki_incremental.py` mode plumbing, `wiki_handle_document_deleted` page_type mapping) looks consistent. Thanks!

PR #18220, opened 2026-08-13
fix(ingestion): key metadata LLM cache by resolved model (D25)

jay77721:
## Summary

The metadata extraction LLM cache key was built from the **raw** `llm_id` parameter. An empty `llm_id` (legal: means "use the tenant default chat model") hashed an empty model segment into the key, so the cache could not tell which model produced a result:

- **Cross-tenant collision**: with an empty model segment, every tenant with identical chunk text + schema shared one Redis key (`kc:meta:`), so tenant B could hit tenant A's extraction.
- **Stale after default-model change**: switching the tenant default model kept serving the old model's result for the 24h TTL.

Python has no such bug — it keys on the resolved `chat_mdl.llm_name` (`task_executor.py:543/550`). After this fix the Go cache key carries a real model identity whenever one is resolvable, matching Python's semantics.

## Change

- New `resolveMetadataCacheModel`: when `llm_id` is empty, resolves the tenant default chat model ref (reusing `defaultChatModelRef`); non-empty `llm_id` passes through unchanged.
- `runAutoExtractions` resolves it **once per Invoke** (not per chunk) and carries it in `extractorInputs.cacheModel`.
- `runEnableMetadata` uses `cacheModel` (fallback to `llmID`) for both cache read and write, keeping them symmetric.

Non-empty `llm_id` path (ingestion main line) is behavior-identical; existing cache entries are untouched. Note: when a tenant's default is a composite `model@provider` ref, tenants sharing the same resolved default model still share a key — the same behavior as Python keying on `llm_name`.

## Tests

`bash build.sh --test ./internal/ingestion/component/` — package passes. New tests:

- `TestResolveMetadataCacheModel_NonEmptyShortCircuits` — non-empty `llm_id` passes through unchanged
- `TestResolveMetadataCacheModel_EmptyNoState_ReturnsEmpty` — safe degenerate path (no canvas state / tenant)
- `TestResolveMetadataCacheModel_EmptyWithPinnedDefault` / `TestResolveMetadataCacheModel_EmptyWithCompositeDefault` — empty `llm_id` resolves the pinned tenant_model UUID / composite `model@provider` default
- `TestResolveMetadataCacheModel_CrossTenantDefaultsDiffer` — tenants with different defaults resolve to different refs
- `TestResolveMetadataCacheModel_EmptyNoDefaultModel` — no default model → falls back to pre-fix behavior
- `TestMetadataLLMCacheKey_ModelIsolation` — distinct models / empty-vs-resolved refs produce distinct keys; deterministic; text/schema stay key dimensions
- `TestExtractorComponent_runEnableMetadata_UsesCacheModel` — per-invocation `cacheModel` is consumed without breaking extraction

### DIFF
--- a/internal/ingestion/component/extractor.go
+++ b/internal/ingestion/component/extractor.go
@@ -536,8 +536,15 @@ type extractorInputs struct {
 	llmID        string
 	systemPrompt string
 	prompt       string
-	lang         string
-	chunks       []map[string]any
+	// cacheModel is the model identity used to key the metadata LLM cache,
+	// resolved once per Invoke in runAutoExtractions (D25). When llm_id is
+	// empty it holds the tenant default chat model ref so the key keeps a
+	// real model identity instead of an empty segment (which would collide
+	// across tenants and go stale after the default model changes). Empty
+	// elsewhere falls back to llmID in runEnableMetadata.
+	cacheModel string
+	lang       string
+	chunks     []map[string]any
 	// temperature overrides the LLM temperature for this call. A
 	// nil value leaves the request's Temperature unset so the model
 	// (or the chat-model default) decides, matching Python's generic
@@ -823,6 +830,13 @@ func (c *ExtractorComponent) runAutoExtractions(ctx context.Context, db *gorm.DB
 	if c.Param.AutoKeywords == 0 && c.Param.AutoQuestions == 0 && c.Param.EnableMetadata == 0 {
 		return nil
 	}
+	// Resolve the metadata cache key model once per invocation, not per
+	// chunk (D25): every chunk job shares the same llm_id / tenant context,
+	// so a per-chunk resolve would fan N identical tenant queries through
+	// the pool workers. Only pay it when metadata extraction actually runs.
+	if c.Param.EnableMetadata > 0 && len(c.Param.Metadata) > 0 {
+		in.cacheModel = resolveMetadataCacheModel(ctx, db, in.llmID)
+	}
 	futs := make([]utility.WorkerPoolFuture[extractorJob, struct{}], 0, len(in.chunks))
 	for i, ck := range in.chunks {
 		i, ck := i, ck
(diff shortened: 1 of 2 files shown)


jay77721:
## Review follow-up: test coverage added

Per review feedback, added a commit covering the previously untested parts of the fix (unit tier, no external services):

- **Resolver happy path** — `resolveMetadataCacheModel` with empty `llm_id` now tested for both the pinned `tenant_model` UUID and the composite `model@provider` default (`TestResolveMetadataCacheModel_EmptyWithPinnedDefault` / `..._EmptyWithCompositeDefault`), plus cross-tenant default isolation and the no-default fallback.
- **Cache key contract** — `TestMetadataLLMCacheKey_ModelIsolation` pins that distinct model refs (and an empty pre-fix segment vs a resolved ref) produce distinct `kc:meta:` keys, that keys are deterministic, and that chunk text / schema stay key dimensions. `get/setMetadataLLMCache` are thin wrappers around this key, so read/write symmetry holds by construction.

Verified with `bash build.sh --test ./internal/ingestion/component/` — package passes. Also corrected the `task_executor.py` line reference in the resolver comment (540/547 → 543/550).

Note: as discussed, when a tenant's default is a composite `model@provider` ref, tenants sharing the same resolved default model still share a cache key — identical to Python keying on `llm_name`, and strictly narrower than the pre-fix empty-segment collision.

internal/ingestion/component/extractor.go
```diff
@@ -536,8 +536,15 @@ type extractorInputs struct {
...
 	prompt       string
-	lang         string
-	chunks       []map[string]any
+	// cacheModel is the model identity used to key the metadata LLM cache,
+	// resolved once per Invoke in runAutoExtractions (D25). When llm_id is
+	// empty it holds the tenant default chat model ref so the key keeps a
+	// real model identity instead of an empty segment (which would collide
+	// across tenants and go stale after the default model changes). Empty
+	// elsewhere falls back to llmID in runEnableMetadata.
+	cacheModel string
```
yuzhichang, line 545: Don't declare `cacheModel` here. This structure is for component input.

PR #18221, opened 2026-08-13
perf(web): stop re-rendering the whole chat transcript on every stream flush

dcc123456:
### Summary

perf(web): stop re-rendering the whole chat transcript on every stream flush

Each streaming flush replaced the message list, and the render loop then rebuilt every message's reference object inline. That handed MessageItem a brand new reference on each flush, defeating its memo and re-parsing the markdown of the entire transcript ~16 times a second.

- Resolve all references once per message-list change in useMessageReferences, keyed by message object (a question and its answer share an id), and share a single EmptyReference fallback.
- Memoize MarkdownContent and hoist its rehype plugins and components out of the render body, since react-markdown rebuilds its processor whenever a plugin identity changes.
- Skip the empty setDocumentIds write that re-rendered every reference-less message.
- Collapse the rAF + setTimeout auto-scroll chain into one cancellable frame and update isAtBottomRef in the scroll handler, so a queued scroll can no longer yank the view back down after the user scrolls up.
- Require callers of sendMessage to pass the history snapshot taken before the question and placeholder were appended, instead of reading a store value that already contains them.

### DIFF
--- a/web/src/components/markdown-content/index.tsx
+++ b/web/src/components/markdown-content/index.tsx
@@ -22,7 +22,7 @@ import { citationMarkerReg } from '@/utils/citation-utils';
 import { getExtension } from '@/utils/document-util';
 import { getDirAttribute } from '@/utils/text-direction';
 import DOMPurify from 'dompurify';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { memo, useCallback, useEffect, useMemo } from 'react';
 import Markdown from 'react-markdown';
 import SyntaxHighlighter from 'react-syntax-highlighter';
 import rehypeKatex from 'rehype-katex';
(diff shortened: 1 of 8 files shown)


wangq8 (commented):
Thanks for this — the diagnosis is correct and the core memoization/hoisting work is well done. I reviewed against current `main` (b1de7e813). A few things to address before landing:

## 1. This PR bundles two behavioral changes that aren't perf-related and aren't in the summary

`web/src/pages/next-chats/hooks/use-send-chat-message.ts`:

- **Regenerate semantics changed.** `useRegenerateMessage` (truncate-after + rewrite-in-place, preserving the original question id — the behavior #17087 fixed) is replaced by an inline `regenerateMessage` that re-asks the question as a brand-new turn (fresh `uuid()`), keeping the previous Q&A in the transcript. That's a user-visible behavior change, not a perf change. Please split it out, or document the intended product change explicitly.

- **New-conversation hydration added.** `handlePressEnter` now calls `hydrateFromServer(targetConversationId, buildMessageListWithUuid(currentMessages))` *before* appending the question, which reverses the "Do NOT hydrate from currentMessages here" guidance from #18180. I follow the reasoning (the prologue seeded on the temp id doesn't carry over to the server-generated id, so the history snapshot would drop it), but it's a subtle correctness change that deserves its own review/PR — or at least a clear note in the description.

Both are defensible on their own, but they're currently hidden inside a `perf(web)` PR.

## 2. Minor — `EmptyReference` is a shared mutable singleton

`web/src/pages/next-chats/utils.ts`: any future consumer that mutates `doc_aggs`/`chunks` (e.g. `push`) would corrupt every reference-less message at once. Current callers are read-only, but consider `Object.freeze(...)` or a comment so it stays that way.

## 3. Minor — skipping the empty `setDocumentIds` write can leave stale state

`web/src/components/markdown-content/index.tsx`: the `if (documentIds.length === 0) return;` guard means that if a mounted message's reference ever transitions from non-empty → empty, the thumbnail-id state won't reset and stale thumbnails could persist. Fine for the current streaming flow (references only get populated, never cleared), but worth a one-line comment.

## 4. Verified-correct

- `useMessageReferences` faithfully mirrors the old `buildMessageItemReference` indexing (prologue skip, error-answer exclusion, pending-answer → `EmptyReference`, first-id-wins).
- The module-scope `rehypeWrapReference` / `MarkdownCode` / `MarkdownParagraph` hoists in `markdown-content` capture no render-scope state, so the `react-markdown` processor rebuild is genuinely avoided.
- The scroll change (synchronous `isAtBottomRef` update in the scroll handler + single cancellable rAF + re-check inside the frame) is sound and actually fixes the "yank back down" race.
- All `sendMessage` callers now pass an explicit history snapshot; no caller is left spreading `undefined`.

PR #18222, opened 2026-08-13
fix(agent): keep think tags in persisted agent session messages

euvre:
## Problem

In the agent (canvas) chat, the **"thought" (思考完成) indicator below the assistant message action icons disappears** after a run finishes. While streaming, the section renders fine; once the front-end refetches the session message list, it is gone.

Root cause: the chat UI builds the thought section from `<think>` tags in the message content. Streaming reconstructs them from `start_to_think`/`end_to_think` message events, but the Go port persisted only the plain answer into the agent session. Python `canvas_service.completion` accumulates the markers into the stored content:

```python
if ans["event"] == "message":
    txt += ans["data"]["content"]
    if ans["data"].get("start_to_think", False):
        txt += "<think>"
    elif ans["data"].get("end_to_think", False):
        txt += "</think>"
```

so after a refetch the tags are still there; in Go they were dropped, and with them the indicator.

## Fix

- `internal/service/agent.go`: `persistAgentRunSession` now stores the assistant message as `<think>{thinking}</think>{answer}` (new `agentSessionMessageContent` helper), mirroring the Python accumulation order.
- `internal/handler/agent.go`: the non-streaming `chat/completions` aggregation keeps `start_to_think`/`end_to_think` markers as `<think>` tags in the final content, mirroring Python `agent_api.py`.

## Tests

- `TestAgentSessionMessageContent` — helper table test (wrap / no-op / empty-answer / inline-tags cases).
- `TestPersistAgentRunSessionPreservesThinking` — DB-level: the stored session message keeps the think section.
- `TestAgentChatCompletions_NonStreamingPreservesThinkMarkers` — handler-level: aggregated non-stream content keeps the think section.

`go test ./internal/handler/ ./internal/service/` passes.

Note: end-to-end verification with a live reasoning model was not possible in this environment (LLM provider credits exhausted); correctness is anchored on Python parity plus the regression tests above.

### DIFF
--- a/internal/handler/agent.go
+++ b/internal/handler/agent.go
@@ -1395,6 +1395,14 @@ func (h *AgentHandler) AgentChatCompletions(c *gin.Context) {
 				if c, ok := evData["content"].(string); ok {
 					fullContent += c
 				}
+				// Mirror Python agent_api.py: the reasoning segment stays
+				// wrapped in <think> tags in the aggregated answer so the
+				// chat UI can render the "thought" section.
+				if st, _ := evData["start_to_think"].(bool); st {
+					fullContent += "<think>"
+				} else if et, _ := evData["end_to_think"].(bool); et {
+					fullContent += "</think>"
+				}
 			}
 			if ref, _ := evData["reference"].(map[string]any); ref != nil {
 				for k, v := range ref {
--- a/internal/handler/agent_test.go
+++ b/internal/handler/agent_test.go
@@ -866,6 +866,46 @@ func TestAgentChatCompletions_DefaultBranchNonStreaming(t *testing.T) {
 	}
 }
 
+// TestAgentChatCompletions_NonStreamingPreservesThinkMarkers covers the
+// non-streaming aggregation: start_to_think/end_to_think message events must
+// survive as <think> tags in the final content, mirroring Python
+// agent_api.py, so clients can render the "thought" section.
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
Reviewed the diff and cross-checked against Python `canvas_service.completion` and `agent_api.py`. The fix looks correct and the tests are solid.

**What's good**
- `agentSessionMessageContent` reconstructs `<think>{thinking}</think>{answer}`, matching Python's accumulation order (reasoning first, bracketed by the start/end markers).
- The non-streaming `AgentChatCompletions` aggregation mirrors `agent_api.py`: content is appended first, then `<think>`/`</think>` from the `start_to_think`/`end_to_think` markers.
- All three `persistAgentRunSession` call sites are updated; there are no other callers.
- Good coverage: table-driven helper test, DB-level persistence test, and a handler-level non-streaming test.

**Minor / non-blocking**
1. `agentSessionMessageContent` doesn't reuse `splitInlineThink`. In the streaming emit path `buildAgentMessageEvents` calls `splitInlineThink(answer, thinking)` to strip any inline `<think>…</think>` from `answer` before re-wrapping, but the persist path does not. If a component ever populates the `thinking` bucket *and* leaves inline `<think>` tags in `answer`, the persisted message would end up with two think sections. In practice the two are mutually exclusive today, but reusing `splitInlineThink` (or documenting that invariant) would make this robust.

No regressions spotted — LGTM.

PR #18224, opened 2026-08-13
Fix: Remove the right anchor point of the tool node

cike8899:
### Summary

Fix: Remove the right anchor point of the tool node


wangq8 (commented):
One thing I'd like confirmed before merge: removing the right source handle from the tool node.

The code being deleted explicitly warned that *"v1 ExeSQL and similar 'tool' components route their result downstream, so they need a source handle too. Without this, any edge where the toolNode is the source silently fails to render."* Can you confirm tool nodes (especially ExeSQL/Code-style tools) are no longer intended to be edge sources in the current design? If any tool still emits a downstream edge, this will silently break rendering.

Otherwise the refactor looks clean: `isConnectableStart` is a valid `@xyflow/react` v12 prop (already used elsewhere as `isConnectableEnd`), and `BottomHandle` nicely consolidates the duplicated bottom-handle markup.

One visual note: the bottom handles now render as a grey chevron (`!bg-text-disabled` / `!border-border-button` with a `ChevronDown` icon) instead of the previous `!bg-accent-primary` dot — confirming that's the intended look.

PR #18225, opened 2026-08-13
Fix: restrict SiliconFlow models in CI to 4 free models

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
LGTM.

The `model_info` payload matches the add-instance API schema (array of `{ model_type, model_name, max_tokens }`), and restricting to the four free BAAI models avoids binding/verifying the full SiliconFlow factory catalog (including paid chat models) during CI. Test-only change, low risk.

One thing to double-check: confirm these four models cover every model the suite actually references via `@CI@SILICONFLOW` (the fixture comment mentions `BAAI/bge-reranker-v2-m3` and `BAAI/bge-m3`; `bge-large-en-v1.5`/`bge-large-zh-v1.5` are added here as well). If the suite uses any other SiliconFlow model, that test will now fail to find it.

PR #18226, opened 2026-08-13
fix(web): keep reasoning and internet options when regenerating an answer

dcc123456:
### Summary

fix(web): keep reasoning and internet options when regenerating an answer

The regenerate button lives in the transcript, which cannot reach the
thinking / internet toggles held as local state inside NextMessageInput,
so both single and multi-model chat resent the question without the
`reasoning` and `internet` parameters. Each view now remembers the
options of its last send and replays them, falling back to the input
box's own defaults (persisted thinking level, internet off) so a freshly
mounted view stays in sync with the visible toggles.

### DIFF
--- a/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
+++ b/web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx
@@ -57,7 +57,7 @@ import {
   UseSendSingleMessageParameter,
 } from '../../hooks/use-send-single-message';
 import { useUploadFile } from '../../hooks/use-upload-file';
-import { buildMessageItemReference } from '../../utils';
+import { buildMessageItemReference, resolveResendOptions } from '../../utils';
 import { useAddChatBox } from '../use-add-box';
 import { useShowInternet } from '../use-show-internet';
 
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Approach is sound and fixes the reported issue (regenerate dropping the `reasoning`/`internet` options). A few observations:

1. **Semantics:** `resolveResendOptions` replays the options of the last *sent* message, not the input box's currently-visible toggles. If a user changes the thinking/internet toggle without sending and then clicks regenerate, the regenerate will ignore the visible change. This matches your documented design, just flagging the edge case so it's an intentional tradeoff.

2. **Deps churn:** Adding `messages` to `sendMessage`'s `useCallback` deps means `sendMessage` (and the `regenerateMessage`/`handlePressEnter` callbacks that depend on it) is recreated on every stream chunk, since `useCurrentChatSession().messages` changes while streaming. Minor re-render churn during streaming. Consider reading the store snapshot inside the callback (like the old `useChatStreamStore.getState().sessions[...]` pattern) to keep the dep list stable while still getting a fresh history.

3. **Fallback correctness:** The new `messages: [...(Array.isArray(explicitMessages) ? explicitMessages : messages), message]` also incidentally fixes a latent spread-of-undefined — `handlePressEnter` passes `messages: undefined` for existing conversations (`currentMessages` is always `[]` there), which previously would throw on `[...undefined, message]`. Worth noting the fallback relies on the React closure being pre-append; if a future refactor reads fresh store state there it would re-send the question twice (the original code's comment warned about exactly this).

Types line up (`enableThinking` is a string, `storage.getThinkingLevel()` returns a string; all fields on `NextMessageInputOnPressEnterParameter` are optional). No blocking issues.

PR #18230, opened 2026-08-13
fix(web): render file-type icons for mp3/svg and fall back to a default file icon

euvre:

In the knowledge base file list and the file manager, files whose extension is not in the icon-font sprite map (e.g. .mp3, .svg, .epub, .deb) — as well as extension-less files created via **Add file → Create empty file** — rendered a blank space instead of an icon.

Root cause: 
- `FileIcon` (file manager) only rendered `IconFont` when `FileIconMap[extension]` existed; the map has no `mp3`/`svg` entry even though colored `assets/svg/file-icon/mp3.svg` and `svg.svg` assets exist.
- The dataset list `FileIcon` passed `file-icon/<ext>` straight to `SvgIcon`, which rendered an empty `` for extensions without a matching svg asset.

### Type of change

### Changes

- `web/src/components/svg-icon.tsx`: add `hasSvgIcon()` against the eagerly-loaded svg assets; `SvgIcon` falls back to the lucide `File` icon when the requested icon is missing; removed dead commented-out webpack glob code.
- `web/src/components/file-icon/index.tsx` (dataset file list): keep thumbnail/SvgIcon paths, fall back to lucide `File` when no colored svg asset exists.
- `web/src/components/icon-font.tsx` (file manager `FileIcon`): folder → folder asset, extension in `FileIconMap` → icon-font sprite, colored svg asset → `SvgIcon`, otherwise lucide `File`.

### Verification

Browser-verified on a local instance (dataset file list + file manager):

| File type | Before | After |
| --- | --- | --- |
| .mp3 | blank | colored mp3 icon (svg asset) |
| .svg | blank | colored svg icon (svg asset) |
| .epub / .deb | blank | default lucide file icon |
| extension-less (Create empty file) | blank | default lucide file icon |
| folders / mapped types (e.g. .pdf) | unchanged | unchanged |

`npm run lint` and `npm run type-check` pass with 0 errors.

### DIFF
--- a/web/src/components/file-icon/index.tsx
+++ b/web/src/components/file-icon/index.tsx
@@ -15,7 +15,8 @@
  */
 
 import { getExtension } from '@/utils/document-util';
-import SvgIcon from '../svg-icon';
+import { File as LucideFile } from 'lucide-react';
+import SvgIcon, { hasSvgIcon } from '../svg-icon';
 
 import { useFetchDocumentThumbnailsByIds } from '@/hooks/use-document-request';
 import { useAuthenticatedImageUrl } from '@/components/image';
@@ -41,10 +42,14 @@ const FileIcon = ({ name, id }: IProps) => {
     }
   }, [id, setDocumentIds]);
 
+  const iconName = `file-icon/${fileExtension}`;
+
   return blobUrl ? (
     <img src={blobUrl} className={styles.thumbnailImg}></img>
+  ) : hasSvgIcon(iconName) ? (
+    <SvgIcon name={iconName} width={24}></SvgIcon>
   ) : (
-    <SvgIcon name={`file-icon/${fileExtension}`} width={24}></SvgIcon>
+    <LucideFile size={24}></LucideFile>
   );
 };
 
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
LGTM.

The root cause is correctly identified: `FileIconMap` has no `mp3`/`svg` entries and `SvgIcon` renders an empty `` for missing assets. The `hasSvgIcon` fallback against the eagerly-loaded `routeList` is correct — names are derived the same way (`file-icon/<ext>`), and `getExtension` lowercases the extension, so `file-icon/mp3` resolves to the existing `assets/svg/file-icon/mp3.svg`.

One minor observation (not a regression): the dataset-list `FileIcon` (file-icon/index.tsx) now falls back to the lucide `File` icon for extensions that exist in `FileIconMap` but have no colored svg asset (e.g. `rmvb`), while the file-manager `FileIcon` (icon-font.tsx) still renders those via the icon-font sprite. So the two lists will look slightly different for such types. Previously both rendered blank/empty, so this is an improvement — just noting for consistency if you want to align them later.

The dead commented-out webpack glob cleanup is a nice touch. Otherwise clean.

dcc123456 (approved):

PR #18232, opened 2026-08-13
fix(agent): attach sys.files uploads to Agent component input messages

euvre:
## Problem

In an agent canvas chat, uploading an image and asking a vision model (e.g. qwen3-vl-plus) to describe it fails: the model answers that no image was uploaded.

Root cause: the Go `Agent` component builds its input as a **plain-text user message** (`buildAgentInputMessages` in `internal/agent/component/agent.go`) and never looks at `sys.files`, while the LLM component does (`internal/agent/component/llm.go` collects `sys.files` and injects images). Python parity: `agent/component/agent_with_tools.py` inherits `LLM._prepare_prompt_variables`, which calls `_collect_sys_files` and attaches image data URIs to the outgoing message — so uploaded images reach the model in the Python runtime but not in the Go one.

## Fix

`buildAgentInputMessages` now collects `sys.files` from the canvas state (reusing the existing `collectSysFiles` helper) and:

- attaches each `data:image/*` entry as an `image_url` multi-content part on the current user message (deduplicated), matching the driver-layer contract used by the LLM component;
- appends non-image file text to the user message body, mirroring the LLM component text injection.

## Tests

New `internal/agent/component/agent_sys_files_test.go`:

- images become `image_url` parts (with dedup) and text uploads are appended to the text part;
- text-only `sys.files` append to the plain content;
- no `sys.files` → message shape unchanged.

`bash build.sh --test ./internal/agent/... ./internal/service/` all pass.

Note: end-to-end verification against the live qwen3-vl-plus was not possible in this environment (LLM provider credits exhausted); correctness is anchored on Python parity plus the unit tests above.

### DIFF
--- a/internal/agent/component/agent.go
+++ b/internal/agent/component/agent.go
@@ -277,6 +277,57 @@ func scanAllStreamForToolCall(_ context.Context, stream *schema.StreamReader[*sc
 // canvas service appends it to state before invoking the workflow.
 func buildAgentInputMessages(ctx context.Context, p AgentParam) []*schema.Message {
 	current := schema.Message{Role: schema.User, Content: p.UserPrompt}
+	// Attach sys.files (agent file uploads) to the current user message,
+	// mirroring Python's LLM._prepare_prompt_variables path that the
+	// Python Agent inherits (agent_with_tools.py). Without this the Agent
+	// component sent text-only messages, so vision models never received
+	// uploaded images and answered "未上传图片".
+	if state, _, err := runtime.GetStateFromContext[*runtime.CanvasState](ctx); err == nil && state != nil {
+		sysFileTexts, sysFileImgs := collectSysFiles(state)
+		if len(sysFileImgs) > 0 {
+			parts := make([]schema.MessageInputPart, 0, 1+len(sysFileImgs))
+			if current.Content != "" {
+				parts = append(parts, schema.MessageInputPart{
+					Type: schema.ChatMessagePartTypeText,
+					Text: current.Content,
+				})
+			}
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Nice fix and a clear root-cause write-up. Verified `collectSysFiles`/`dedupStrings` in `llm.go` and the LLM component's injection path — the Agent component now mirrors it correctly, so `sys.files` images reach vision models as `image_url` multi-content parts.

One correctness nit (non-blocking): when images are present and `UserPrompt` is empty, no text part is created (the `if current.Content != ""` guard), so if `sysFileTexts` also exist, the `!attached` branch writes them into `current.Content` while `UserInputMultiContent` holds only image parts. This yields a message with both `Content` and `UserInputMultiContent` set, which diverges from `buildMessagesWithImages`/the LLM path (which keeps `Content` empty and prepends a text part). In practice a vision prompt is non-empty so it rarely triggers, but aligning the fallback to prepend a `ChatMessagePartTypeText` part would be more consistent and avoid driver ambiguity around mixed content.

Tests are good (image dedup, text-only, no-sys.files). LGTM after the optional nit above.

euvre:
Superseded by #18246. This PR's head branch was deleted (which auto-closed it) because the fix here was incomplete: it attached uploads as multi-modal parts in the Agent component, but `toInternalMessages` (the eino → driver bridge in `internal/entity/models/llm.go`) still discarded `UserInputMultiContent`, so image parts never reached the model. #18246 fixes both layers (Agent component input construction + the bridge), shares one image-injection helper between the Agent and LLM components, and adds unit tests for both.

PR #18233, opened 2026-08-13
refactor(web): drop per-model checkbox selection from batch actions

dcc123456:
### Summary

refactor(web): drop per-model checkbox selection from batch actions

Batch verify now targets every model matching the current search/tag filter instead of a separate checkbox selection, so the row checkboxes, select-all control, and their i18n keys are removed. The batch add/remove control becomes a labeled button next to batch verify, making both filter-scoped actions read consistently.


wangq8 (commented):
LGTM overall. The refactor cleanly removes the per-row/select-all checkbox state and makes batch verify operate on all currently filtered models, matching the PR description.

Verified against the current code:
- `useMemo` removal is safe — the only remaining usage was `filteredSelectedCount`, which is deleted along with the selection state.
- `Loader2` is still imported and used in the new labeled batch add/remove button; `ListMinus`/`ListPlus` (only used by the removed icon button) are correctly dropped.
- `ModelRow` is only consumed in `models-section/index.tsx`, so removing `isSelected`/`onToggleSelect` from `ModelRowProps` is safe.
- `selectModelBeforeVerify` is still referenced by the provider-modal/instance-card flows, so it's correctly retained; the removed keys (`selectAllFiltered`, `batchVerifySelected`) exist only in en.ts/zh.ts, so no orphaned keys remain elsewhere.

Non-blocking observations:
1. This is a real UX behavior change: batch verify previously allowed verifying a hand-picked subset, but now always targets every model matching the current search/tag filter. That's the stated intent, though worth a quick product sign-off since a broad filter match now verifies all models at once.
2. No frontend tests accompany the change (consistent with the existing web codebase), but the new "verify all filtered" path and the labeled toggle button would benefit from a smoke test.

PR #18234, opened 2026-08-13
Fix: reject move to current location request

Lynn-Inf:
### Summary

As title, raise a message to tell user no need to move it.

### DIFF
--- a/api/apps/services/file_api_service.py
+++ b/api/apps/services/file_api_service.py
@@ -509,11 +509,15 @@ async def move_files(uid: str, src_file_ids: list, dest_file_id: str = None, new
                 return False, "Cannot move a folder into its own subfolder."
 
     def _move_entry_recursive(source_file_entry, dest_folder_entry, override_name=None):
+        if source_file_entry.id == dest_folder_entry.id:
+            raise RuntimeError("The folder is already in the target location. There is no need to move it.")
         effective_name = override_name or source_file_entry.name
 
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix — the root cause is clear and the change is correct.

Verified against the current `move_files` implementation:
- `_move_entry_recursive` previously reused the source folder itself as `new_folder` when the source already lived in the destination under the same name, then called `delete_by_id` on it, which is why the folder vanished.
- The new `existing_folder[0].id == source_file_entry.id` guard detects exactly that case and raises before any deletion.
- `thread_pool_exec` propagates worker exceptions via `run_in_executor`, so the new `try/except RuntimeError` correctly converts the guard into `(False, str(e))`.
- Tests cover both "move to current parent" (asserting `deleted == []`) and "move to itself", and look correct.

Non-blocking nits:
1. The first guard `if source_file_entry.id == dest_folder_entry.id` is largely redundant: the top-level "Cannot move a folder to itself." check already rejects that for the top-level source, and the `existing_folder[0].id == source_file_entry.id` check catches the move-to-current-parent case. It's a harmless defensive guard, but its message ("already in the target location") is slightly misleading for the move-to-itself case it would actually catch.
2. The error string is user-facing; that matches the existing inline-string style in this file, so no action needed — just flagging it in case you later want a stable machine-readable code.

LGTM, no blocking issues.

PR #18236, opened 2026-08-13
fix: pack heading-less book text runs into token-bounded chunks

euvre:
### Summary

Book (hierarchy) chunking collapsed a heading-less book into exactly **1 chunk**: when no line in a text run matched a heading pattern, `resolveTargetLevel` returns none and the Go hierarchy chunker emitted the whole run as a single record group (mirroring the python pipeline's `record_groups.append(text_records.copy())`). A 49 KB novel txt (`第N回`-style markers not covered by the configured heading families) or a PDF without bookmarks therefore indexed as one giant chunk, breaking retrieval.

The classic python book chunker (`rag/app/book.py`) falls back to token-bounded `naive_merge` in exactly this case. This PR mirrors that fallback in the Go hierarchy chunker: when the run has no detectable heading levels, it is packed with the same token budget the group variant uses (`minGroupTokens`/`maxGroupTokens`) instead of one oversized group.

### Verification

- Unit: `bash build.sh --test ./internal/ingestion/component/chunker/...` — new regression tests `TestHierarchyTitleChunker_NoHeadingsTokenFallback` (long heading-less book → multiple budget-bounded chunks, no content loss) and `TestHierarchyTitleChunker_NoHeadingsSmallDoc` (short doc still yields exactly 1 chunk); full package green.
- E2E on a local build: re-parsed the reproducing 49 KB `hlm_test.txt` with chunk method **Book** — chunk count went from **1 → 17**, parse status DONE.

### Notes

- Deliberate divergence from the python *pipeline* hierarchy chunker (which keeps the single-chunk behavior); aligns with the classic `rag/app/book.py` product behavior and with the Go group variant, which already bounds chunks by token budget.

### DIFF
--- a/internal/ingestion/component/chunker/group.go
+++ b/internal/ingestion/component/chunker/group.go
@@ -54,13 +54,11 @@ import (
 
 const ComponentNameGroupTitleChunker = "GroupTitleChunker"
 
-// MIN_GROUP_TOKENS / MAX_GROUP_TOKENS mirror the python constants in
-// group_chunker.py:22-23. They drive the merge heuristic for adjacent
-// text records.
-const (
-	minGroupTokens = 32
-	maxGroupTokens = 1024
-)
+// minGroupTokens mirrors MIN_GROUP_TOKENS (group_chunker.py:22): very
+// small chunks are allowed to absorb the next record regardless of
+// section change. The merge ceiling is the configurable chunk_token_cap
+// (group_chunker.py:90), not a fixed constant.
+const minGroupTokens = 32
 
 // resolveTargetLevel mirrors common.py:resolve_target_level: pick the
 // n-th smallest heading level from the per-line `levels` vector.
@@ -242,9 +240,14 @@ func groupRecords(records []lineRecord, secIDs []int, p *titleChunkerParam) [][]
 			continue
 		}
 		tokenCount := tokenizer.NumTokensFromString(text)
+		// Merge ceiling mirrors group_chunker.py:90: the configurable
+		// chunk_token_cap (0/negative disables the ceiling). The post-build
+		// enforceTitleTokenCap stays the single hard guarantee for any
+		// residual over-cap chunk.
+		mergeCeilingOK := p.ChunkTokenCap <= 0 || tkCnt < p.ChunkTokenCap
 		shouldMerge := len(currentGroup) > 0 &&
 			currentGroup[0].isText() &&
-			(tkCnt < minGroupTokens || (tkCnt < maxGroupTokens && secID == lastSID))
+			(tkCnt < minGroupTokens || (mergeCeilingOK && secID == lastSID))
 		if shouldMerge {
 			currentGroup = append(currentGroup, rec)
 			tkCnt += tokenCount
--- a/internal/ingestion/component/chunker/hierarchy.go
+++ b/internal/ingestion/component/chunker/hierarchy.go
@@ -34,6 +34,12 @@
 //     `include_heading_content` and the python leaf-only rule: when
 //     include_heading_content is false, only leaf nodes emit.
 //
+//   - Deliberate divergence from the python pipeline: when a text run
+//     has no detectable heading levels (resolve_target_level → None),
+//     the run is packed into token-bounded groups (group variant
+//     budget) instead of one giant chunk, matching the classic
+//     rag/app/book.py naive_merge fallback.
+//
 //   - No PDF-position / deepdoc awareness; structured chunk inputs
 //     use the same dfs over text records.
 package chunker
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Thanks — this is a well-reasoned fix. The heading-less run previously collapsed into one giant chunk (mirroring the python pipeline's `record_groups.append(text_records.copy())`), which is exactly what the classic `rag/app/book.py` path avoids via token-bounded `naive_merge`. Routing the no-heading case through `groupRecords(textRun, make([]int, len(textRun)), p)` correctly reuses the group variant's `minGroupTokens`/`maxGroupTokens` budget (all-zero secIDs merge until `maxGroupTokens`), and the comment accurately documents the deliberate divergence from the python pipeline.

I verified locally (CGO_ENABLED=0 with the BPE table present): the full `internal/ingestion/component/chunker` package passes, including the two new tests.

Two non-blocking notes:

1. `root_chunk_as_heading` edge case: `buildChunksFromRecordGroups` applies `RootChunkAsHeading` whenever `len(chunks) > 1`, prepending chunk[0]'s text to every chunk and dropping chunk[0]. Before this change a heading-less run always produced exactly one chunk, so that branch was unreachable; now it produces multiple chunks, so if `root_chunk_as_heading` is set (a valid title-chunker param) the first chunk of plain body prose would be treated as a "root heading" and dropped. Worth a guard (e.g. skip the root-heading logic when the run had no heading levels) or at least a test pinning the combination.

2. Token-count dependency: the fallback is only token-bounded if `tokenizer.NumTokensFromString` returns non-zero; when the cl100k BPE table is missing (the loader performs no network I/O), every count is 0 and `groupRecords` merges everything back into a single chunk, silently reintroducing the bug. This is pre-existing behavior shared with the group variant, not introduced here, but the fallback now depends on it for correctness.

Otherwise LGTM — nice regression tests (long doc → ≥2 chunks with content preserved, short doc → exactly 1 chunk).

xugangqiang:
Not good enough. Suggest not to merge. The behavior should be same for related chunkers: like token chunker, title chunker, etc. Also the CAP should be configurable.

euvre:
Thanks for the reviews. Follow-up commit 28f15f0 addresses the feedback:

**@xugangqiang — behavior parity across related chunkers + configurable CAP (applied)**
You were right on both counts. The Python reference (`rag/flow/chunker/title_chunker/group_chunker.py:90`) drives the group merge ceiling with the configurable `chunk_token_cap` (`cap <= 0 or tk_cnt < cap`), but the Go port had a leftover hardcoded `maxGroupTokens = 1024` in `groupRecords`. Fixed: the merge ceiling is now `p.ChunkTokenCap` (0 disables the ceiling, same as Python), and `enforceTitleTokenCap` remains the single hard guarantee for any residual over-cap chunk. Since the heading-less hierarchy fallback reuses `groupRecords`, the hierarchy and group variants now share exactly the same budget logic, matching the sibling chunkers. `minGroupTokens = 32` stays, mirroring Python's `MIN_GROUP_TOKENS`. New test `TestHierarchyTitleChunker_NoHeadingsConfigurableCap` pins that `chunk_token_cap: 128` bounds every fallback chunk while preserving all content.

**@wangq8 — thanks for the LGTM and the careful edge-case notes**
- (a) `root_chunk_as_heading` on heading-less multi-chunk output: real content-loss bug introduced by this PR — when no run matches any heading family, chunk[0] is plain body prose, and the root-as-heading rewrite would drop it as a "root heading" and prepend it to the rest. Fixed by tracking whether any detected heading exists and suppressing the rewrite only for fully heading-less documents; documents with any heading keep the Python behavior unchanged. Covered by `TestHierarchyTitleChunker_NoHeadingsRootChunkAsHeading` (asserts no chunk is dropped and the full text is preserved).
- (b) BPE-table-missing → token count 0: agreed, pre-existing and shared with the group variant, so left untouched in this PR. Note the cap path already has a rune-count fallback (`titleTokenCount`), so `chunk_token_cap` still bounds chunks even when the encoder is unavailable.

**@coderabbitai — full-content-preservation assertions (applied)**
The regression tests no longer assert a single substring. They now compare the whitespace-normalized concatenation of all chunk texts against the full source text, which catches dropped, duplicated, or reordered content — including the "first half missing" case you described. The per-chunk token assertion was also tightened from "budget + one-line overshoot" to the actual `chunk_token_cap` hard ceiling.

**Validation**: `bash build.sh --test ./internal/ingestion/...` — all packages pass, including the two new regression tests. No shared-service end-to-end run was performed; behavior parity was verified against the Python `title_chunker` reference implementation.

euvre:
duplicate.

Issue #10145, opened 2025-09-18
[Question]: when parsing a docx format document, the order number is missing


Starnight-Y:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Describe your problem

When I upload a docx format document and pasering it. I find the order number is missing, how can i do ?
1、the origin content is： 

<img width="687" height="359" alt="Image" src="https://github.com/user-attachments/assets/445348eb-7cb1-4aeb-898b-3322eddc26bb" />
2、the parsing result is：

<img width="904" height="78" alt="Image" src="https://github.com/user-attachments/assets/e163ac73-dea4-49c9-9357-4f5a5c992c04" />

PR #18237, opened 2026-08-13
feat(parser): preserve automatic heading numbering and improve Parser

nikminer:
### Summary

This PR introduces two focused improvements:

1. Preserve automatic numbering of DOCX headings during parsing.
2. Add missing Russian translations for Parser and Chunker settings.

### DOCX automatic heading numbering

Microsoft Word stores automatically generated heading numbers separately from the paragraph text.

Because of this, a heading displayed in Word as:

`2.1 Getting started`

can currently be parsed by RAGFlow as:

`Getting started`

This loses information that is visible in the original document and makes it harder to work with structured technical documentation where section numbers are important.

The proposed solution reconstructs the visible heading number from DOCX numbering metadata and adds it back to the parsed heading text.

For example:

Before:

`Getting started`

After:

`2.1 Getting started`

A new DOCX Parser option is also added:

`extract_automatic_numbering`

It allows users to enable or disable automatic heading numbering extraction.

The option is enabled by default so that parsed DOCX headings preserve the numbering visible in Microsoft Word.

Related issue: #10145

### Russian localization

The PR also adds and corrects Russian translations for settings used in the Parser and Chunker configuration interfaces.

This makes the ingestion pipeline settings easier to understand for Russian-speaking users, especially when configuring:

- Parser settings;
- Token Chunker;
- Title Chunker;
- heading hierarchy and grouping;
- regular expression rules;
- DOCX parsing options;
- automatic numbering extraction.

These changes are limited to interface localization and the new DOCX numbering feature.


nikminer:
@coderabbitai review

xugangqiang:
## Review: problems found (blockers / risks before merge)

**P0 — Default-on is a silent global behavior change.**
`extract_automatic_numbering` defaults to `true` in every place it is wired (`internal/ingestion/component/parser.go`, `docx_parser.go:NewDOCXParser`, `rag/flow/parser/parser.py:__init__`, and the three `web` defaults). The numbering is a heuristic reconstruction (counter-based, no `w:lvlRestart`/suppression/restart modeling). For any DOCX where the heuristic is wrong, headings are silently corrupted and retrieval quality degrades for all existing datasets. Recommend default `false` (opt-in) or at minimum document the heuristic limitation clearly.

**P0 — Go vs Python markdown parity gap on setext headings.**
`applyDOCXNumberedHeadingsToMarkdown` / `docxMarkdownListOrHeadingText` (`internal/parser/parser/docx_numbering.go`) only handles ATX (`#`) and ordered-list prefixes; it completely ignores setext headings (`===`/`---` underline). The Python `apply_numbered_headings_to_markdown` does handle setext. Result: the same DOCX yields numbered setext headings via the Python path but unnumbered setext headings via the Go path. Either add setext handling to the Go path or prove `office_oxide` never emits setext and converge Python to match — but the two pipelines must not diverge.

**P1 — Heading detection口径 inconsistent; outlineLvl can number non-headings.**
`utils.py:extract_word_outlines` correctly adds a `Heading\s*\d+` style-name guard, but `deepdoc/parser/docx_parser.py` and `rag/app/naive.py` call `numbered_heading()` which uses `_heading_level` and accepts any paragraph with `w:outlineLvl` (or a `headingN`/`заголовокN` style name). `w:outlineLvl` can be set on captions/body styles, so numbers can be prepended to non-heading paragraphs. Apply the same style-name guard consistently across all call sites.

**P1 — Counters polluted by non-heading list items sharing a numId.**
In both `docx_numbering.py:_resolve_level`/`numbered_heading` and `docx_numbering.go`, the counter is incremented for *any* paragraph referencing the same `numId+ilvl`, not only headings. If headings and body numbered lists share a `numId`, list items corrupt the heading counter. Typical docs use a dedicated `numId` for headings (low risk), but it should be constrained to heading paragraphs and covered by a test.

**P2 — `applyDOCXNumberedHeadingsToSections` splits multi-line text items.**
For a section whose `text` spans multiple lines it rewrites the item into per-line sub-items and overwrites `ck_type`. `buildDOCXJSONSections` (`docx_ir.go:232`) emits one item per paragraph (single line), so risk is low today, but the test (`TestApplyDOCXNumberedHeadingsToSections`) enshrines the multi-line split with a 3-line blob, which can mask a structural change. Use realistic single-paragraph fixtures and document the limitation.

**P2 — Heuristic limits untested.**
No coverage for: same-`numId` multiple lists restart, mid-document `startOverride`, `w:lvlRestart`, or suppression. Add regression tests for these before relying on default-on behavior.

**P2 — Feature surface expanded into legacy/deprecated paths.**
The logic is added to `deepdoc/parser/docx_parser.py` and `rag/app/naive.py` (and `rag/app/one.py:73`, `rag/app/laws.py:182` call `Docx()(...)` without the flag, falling back to default `true`). These are exactly the paths AGENTS.md treats as liabilities to converge/delete. Confirm they are still user-facing; otherwise avoid growing their code surface and keep the single source of truth (Go future path + shared Python `DOCXNumberingResolver`) in parity.

**Note on existing CodeRabbit comments:** the `ilvl` range check (`:76`), setext detection (`:248`), the Go prefix check (`:207`, already uses `hasDOCXNumberingMarker`), the `r.Close()` errcheck (`:243`), and the `naive.py:620` config pass (already done at `naive.py:1084`) are already addressed/stale. The `CGO_ENABLED=0` build-tag concern (`:671`) is a false positive — `docxOutline` lives in cgo-free `docx_postprocess.go`; still worth one `CGO_ENABLED=0 go build ./internal/parser/parser` in CI. The `utils.py:94` heading-criterion point is valid but only partially fixed (style-name guard missing in `docx_parser.py`/`naive.py`).

PR #18238, opened 2026-08-13
fix(web): enable client pagination in admin user detail tables

dcc123456:
### Summary

fix(web): enable client pagination in admin user detail tables

The dataset and agent tables render pagination controls but never registered the pagination row model, so every row was shown at once.

### DIFF
--- a/web/src/pages/admin/user-detail.tsx
+++ b/web/src/pages/admin/user-detail.tsx
@@ -9,6 +9,7 @@ import {
   createColumnHelper,
   flexRender,
   getCoreRowModel,
+  getPaginationRowModel,
   getSortedRowModel,
   useReactTable,
 } from '@tanstack/react-table';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this fix — it's minimal and correct.

Verified against `web/src/pages/admin/user-detail.tsx`: the `RAGFlowPagination` controls were already wired up (`total`, `current`, `pageSize`, and `onChange` → `table.setPagination`), but without `getPaginationRowModel` the table returned every row regardless of pagination state, so the controls rendered while all rows were shown at once. Registering the pagination row model makes `table.getRowModel().rows` respect the pagination state and actually paginate. The default `pageSize` (tanstack's 10) lines up with `RAGFlowPagination`'s initial select value ("10"), so first paint is consistent.

Two minor, non-blocking suggestions:
1. Consider setting `initialState: { pagination: { pageSize: 10 } }` explicitly. The current default relies on the implicit coupling between tanstack's default (10) and `RAGFlowPagination`'s hardcoded initial `currentPageSize` ("10"); making it explicit would guard against silent drift.
2. `total={props.data?.length}` can be `undefined` before data loads; `?? 0` would keep the prop type clean (it already defaults to 0 internally, so this is purely cosmetic).

No regressions spotted; LGTM.

PR #18239, opened 2026-08-13
fix: store source key only in pipeline_operation_log.source_from

euvre:

KB logs page (Dataset -> Logs) shows a generic computer-upload icon in the Source column for files synced from a datasource (e.g. RSS) instead of the datasource-specific icon.

**Root cause:** for connector-synced documents, `document.source_type` is stored as `<source>/<connector_id>` (e.g. `rss/<connector_id>`). The Go ingestion backend (`recordPipelineLog` in `internal/ingestion/task/pipeline_executor.go`) copied this verbatim into `pipeline_operation_log.source_from`. The frontend (`web/src/pages/dataset/dataset-overview/overview-table.tsx`) maps `source_from` to a datasource icon by exact key lookup (`DataSourceKey.RSS === 'rss'`), so the connector-qualified value missed the lookup and fell back to the generic icon.

**Fix:** strip the connector id and store only the bare source key in `source_from` (`strings.Cut(sourceType, "/")`).

Verified live: inserted two `pipeline_operation_log` rows into a test KB, one with `source_from='rss/<connector_id>'` (shows generic computer icon) and one with `source_from='rss'` (shows the RSS icon).

### Type of change

### DIFF
--- a/internal/ingestion/task/pipeline_executor_test.go
+++ b/internal/ingestion/task/pipeline_executor_test.go
@@ -411,6 +411,38 @@ func TestRecordPipelineLog_CustomCanvasMissingFallsBackToParserID(t *testing.T) 
 	}
 }
 
+func TestRecordPipelineLog_SourceFrom(t *testing.T) {
+	cases := []struct {
+		name       string
+		sourceType string
+		want       string
+	}{
+		{name: "connector source strips connector id", sourceType: "rss/connector-811", want: "rss"},
+		{name: "plain source unchanged", sourceType: "local", want: "local"},
+		{name: "empty source unchanged", sourceType: "", want: ""},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			taskCtx := makeTaskCtx()
+			taskCtx.Doc.SourceType = tc.sourceType
+			var captured *entity.PipelineOperationLog
+			svc := mustNewPipelineExecutor(t, taskCtx, "flow-1", 0).WithLogCreateFunc(
+				func(ctx context.Context, db *gorm.DB, log *entity.PipelineOperationLog) error {
+					captured = log
+					return nil
+				},
+			)
+			svc.recordPipelineLog(t.Context(), dao.DB, "doc-1", `{"components": {}}`, "done")
+			if captured == nil {
+				t.Fatal("logCreateFunc was not called")
+			}
+			if captured.SourceFrom != tc.want {
+				t.Errorf("SourceFrom = %q, want %q", captured.SourceFrom, tc.want)
+			}
+		})
+	}
+}
+
 // =============================================================================
 // updateDocumentMetadata
 // =============================================================================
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
The fix itself is correct — synced documents store `source_type` as `<source>/<connector_id>` (`internal/service/sync_task_service.go`), and `strings.Cut(sourceType, "/")` correctly recovers the bare source key so the KB logs page can match the datasource icon.

However, this PR currently shows `mergeable: CONFLICTING` against `main`. The same fix has already been merged via #18366 (commit `1ef4dddde`, "fix: align Go ingestion progress and pipeline selection"), which strips the connector id in `recordPipelineLog` using `strings.SplitN(sourceFrom, "/", 2)`.

Please rebase onto latest `main` and confirm whether this PR is still needed. If #18366 already covers the reported bug, this PR can be closed as superseded; otherwise update the change to align with the reloaded `doc` variable now used in `recordPipelineLog`.

The added test (`TestRecordPipelineLog_SourceFrom`) is good, but it will also need updating for the `doc` reload introduced in #18366.

euvre:
@wangq8 Thanks for the careful review — addressed each point:

**1. Rebase / conflict.** The branch is now based on the current `main` tip (`4304a3b7c`), so the CONFLICTING state is resolved.

**2. Is the PR still needed?** Confirmed: the production fix was merged via #18366 — `recordPipelineLog` on main now derives `source_from` from the reloaded `doc` and strips the connector id with `strings.SplitN(sourceFrom, "/", 2)` (`internal/ingestion/task/pipeline_executor.go`). Accordingly, this PR no longer contains any production-code change; its remaining value is **test coverage only**, which main does not have for this behavior.

**3. Test updated for the `doc` reload.** As you noted, the original `TestRecordPipelineLog_SourceFrom` only exercised the fallback path (`db == nil`, snapshot from `taskCtx.Doc`). Added `TestRecordPipelineLog_SourceFromReloadedDoc`: it seeds a document row with `source_type = 'rss/connector-811'` in the in-memory sqlite DB, deliberately sets a stale `taskCtx.Doc.SourceType = 'local'`, and asserts the recorded log stores `source_from = 'rss'` — proving the value comes from the reloaded document, not the task-context snapshot. Also extended `setupPipelineExecutorTestDB` to auto-migrate `entity.Document`. Pushed as `dc3a353da`.

**Verification.** `bash build.sh --test ./internal/ingestion/task/...` passes; a focused run shows `TestRecordPipelineLog_SourceFrom` (3 subtests) and `TestRecordPipelineLog_SourceFromReloadedDoc` all PASS. Validation boundary: no end-to-end run against shared ragflow services was performed from this lane; the evidence is unit tests plus line-level comparison with the #18366 implementation on main.

Net: this PR is now a test-only change that locks in the source_from normalization behavior (connector-id stripping on both the reload path and the no-DB fallback path). If you judge the coverage unnecessary, feel free to close it as superseded by #18366.

PR #18241, opened 2026-08-13
fix(llm): disable DeepSeek V4 thinking by default via extra_body

D819:
## Summary
- Disable DeepSeek V4 thinking (chain-of-thought) by default.
- litellm 1.82.x drops `thinking: disabled`; carry the toggle through `extra_body.thinking.type` and strip `reasoning_effort` to avoid the 400.
- Use local timezone for agent `sys.date` instead of UTC.

Reference: https://api-docs.deepseek.com/guides/thinking_mode

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -335,7 +335,7 @@ class Canvas(Graph):
             "sys.conversation_turns": 0,
             "sys.files": [],
             "sys.history": [],
-            "sys.date": datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S"),
+            "sys.date": datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
         }
         self.variables = {}
         # Aggregated provider token usage (prompt/completion/total) across every LLM
@@ -354,7 +354,7 @@ class Canvas(Graph):
             if "sys.history" not in self.globals:
                 self.globals["sys.history"] = []
             if "sys.date" not in self.globals:
-                self.globals["sys.date"] = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S")
+                self.globals["sys.date"] = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
         else:
             self.globals = {
                 "sys.query": "",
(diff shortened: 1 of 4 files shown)


Lynn-Inf (changes requested):
Keep UTC for `agent sys.date` — frontend handles local conversion on display. Double conversion if we switch. Rest LGTM.

D819:
Thanks for the review @Lynn-Inf. Re `sys.date`: I understand the concern about double conversion, but in this specific case the backend is **not** double-converting, because `sys.date` is never rendered on the frontend — it is only injected into the LLM prompt as a system variable.

A few concrete findings from the codebase:

1. **`sys.date` is a prompt-injection variable, not a display value.** It is defined in `agent/canvas.py` (globals) and substituted into the LLM's `sys_prompt` / user prompt via `string_format` (`agent/component/base.py:602`, matched as `{sys.date}`). The model reads it as "today's date / day of week" context; the frontend does not parse or display it. There is no frontend UTC→local conversion for `sys.date` (grep of `web/src` finds only generic UI date components).

2. **Every other "current date" variable in the repo uses local time, so UTC here is actually the inconsistency.** For example:
   - `rag/prompts/generator.py:272-274` — `today/yesterday/tomorrow = datetime.date.today()`
   - `rag/prompts/generator.py:435` — `today=datetime.datetime.now()`
   - `rag/prompts/generator.py:521` — `current_date=datetime.datetime.today()`
   - `api/db/db_utils.py`, `api/db/services/*` — `date.today()` / `datetime.now()`

   These are all local-time "current date" values fed to the LLM or DB. `agent/canvas.py`'s `sys.date` was the only one using UTC, which made it inconsistent with the rest of the project.

3. **Real-world symptom.** In practice, with the old UTC value, an agent whose system prompt uses `{sys.date}` returns a UTC timestamp, and a user in UTC+8 must manually add something like "convert the UTC time to UTC+8 Beijing time" in the system prompt for the model to report the local date correctly. That workaround is precisely the kind of thing that should not be needed.

So switching `sys.date` to `datetime.datetime.now()` aligns it with the rest of the project and removes the need for a prompt-level timezone workaround. No frontend double conversion is involved because the frontend never displays this value.

Happy to split the DeepSeek fix and the `sys.date` fix into two PRs if that's preferred.

Lynn-Inf (approved):
After discussing with my colleague, we agree that using local time (not UTC) for `sys.date` is better — server time in most open‑source deployments matches the user's local time.

Disabling DeepSeek V4 `thinking` by default also makes sense. LGTM.

PR #18244, opened 2026-08-13
Fix: cross-KB document id collision in data source sync (#18116)

### LINKED ISSUE #18116: [Bug]:
### Self Checks

none

### RAGFlow image version

v0.25.1

```Markdown

```

### Actual behavior

# Bug Report: "Existing document id collision with another knowledge base" Error in Data Source Sync

**To RAGFlow Support Team,**

We encountered a bug in the **Data Source Sync** feature that persists after all conventional troubleshooting steps. Please help investigate.

## Problem Description

When syncing documents via the built-in "Link Data Source" feature, a large number of documents fail with:

```
{doc_name}: Existing document id collision with another knowledge base; skipping update.
```

However, the conflicting documents **do not exist in any knowledge base** on the RAGFlow instance.

## Reproduction Steps

1. Create a data source (e.g., Jira/Xray)
2. Create **Knowledge Base A** and **Knowledge Base B**, both linked to the **same data source**
3. Trigger sync on **Knowledge Base A** → documents are synced successfully to A
4. Knowledge Base B remains empty (no sync triggered yet)
5. Delete all documents from Knowledge Base A
6. Now try to trigger sync on **either** Knowledge Base A or Knowledge Base B
7. **Result**: Both fail with `Existing document id collision with another knowledge base; skipping update.`

## Steps Already Taken (None Resolved the Issue)

1. Searched all knowledge bases via UI and API — conflicting documents not found in any dataset
2. Deleted all documents from the target knowledge base
3. Deleted the entire knowledge base (dataset) and recreated it
4. Disconnected the data source and reconnected it
5. Created a brand new knowledge base from scratch and linked the data source to it
6. Directly called `POST /api/v1/datasets/{id}/documents` with the same document name to different datasets — **this works fine** (returns code 0, no collision error)

## Root Cause Analysis

The error **only occurs in the Data Source Sync feature**, not in the direct upload API. This suggests the sync engine maintains an **internal mapping table** of `external_document_id → internal_document_id/dataset_id` that is **not cleaned up** when:

- Documents are deleted from the knowledge base
- The entire knowledge base is deleted
- The data source is disconnected and reconnected

This mapping persists in the RAGFlow database at the instance level, causing the sync engine to report "collision with another knowledge base" even when the target dataset no longer exists.

## Request

1. Please check if there is a way to clear the sync engine's internal mapping table from the server-side
2. Or provide an API endpoint to reset the sync state for a given data source
3. Or advise on which database tables to clean up (e.g., `sync_documents`, `data_source_mapping`, etc.)

## Environment

- **RAGFlow Version**:0.25.1
- **Deployment**: Docker
- **Data Source Type**: Jira/Xray

## Additional Information

I can provide full error logs and API call traces if needed. Please let me know what additional information you require.

---

*Date: 2026-08-11*

### Expected behavior

1. Duplicates allowed
2, or indicate which knowledge base the conflicting document comes from.

### Steps to reproduce

```Markdown
## Reproduction Steps

1. Create a data source (e.g., Jira/Xray)
2. Create **Knowledge Base A** and **Knowledge Base B**, both linked to the **same data source**
3. Trigger sync on **Knowledge Base A** → documents are synced successfully to A
4. Knowledge Base B remains empty (no sync triggered yet)
5. Delete all documents from Knowledge Base A
6. Now try to trigger sync on **either** Knowledge Base A or Knowledge Base B
7. **Result**: Both fail with `Existing document id collision with another knowledge base; skipping update.`
```

### Additional information

_No response_

sarbojitrana:

Closes #18116.

Two knowledge bases linked to the same data source fight over one document row. The second one to sync loses, and the document is dropped:

```
{doc_name}: Existing document id collision with another knowledge base; skipping update.
```

The reporter then could not find the conflicting document in *any* knowledge base, and deleting documents, deleting the dataset, and re-linking the data source all failed to clear it. They ended up deleting rows from the `document` table in MySQL by hand.

## Root cause

A connector-sourced document's primary key is **derived from its external document id**, so that re-running a sync updates the existing row instead of duplicating it. Two of the three derivations that have shipped leave `kb_id` out:

| shipped in | derivation | same id for two KBs? |
|---|---|---|
| ≤ v0.25.1 | `hash128(external_id)` | **yes** |
| after v0.25.1 | `hash128(f"{connector_id}:{external_id}")` | **yes** |
| current | `hash128(f"{kb_id}:{connector_id}:{external_id}")` | no |

`FileService.upload_document` guards against one knowledge base overwriting another's row ([file_service.py#L564](https://github.com/infiniflow/ragflow/blob/8f6b5d0e/api/db/services/file_service.py#L564)):

```python
e, doc = DocumentService.get_by_id(doc_id)
if e:
    if str(doc.kb_id) != str(kb.id):
        # "Existing document id collision with another knowledge base; skipping update."
```

The guard is correct, but under a KB-agnostic derivation it fires on every knowledge base after the first. On v0.25.1 — the version in the report — that means **an external document can only ever live in one knowledge base**, forever.

Three things then make it unrecoverable rather than merely wrong:

1. **The failure is terminal.** The id is derived, so every subsequent sync recomputes the same colliding id and gets skipped again. Nothing in the codebase re-keys the document or clears the blocking row.
2. **The message names no knowledge base**, so when the blocking row outlives its dataset the user is told to go find a conflict that cannot be found — exactly what the report describes.
3. **Deleting the dataset leaves the connector wiring behind.** `_delete_datasets_sync` never removes the dataset's `Connector2Kb` or `SyncLogs` rows, so `ConnectorService.schedule_tasks` keeps queueing syncs for a `kb_id` that no longer resolves; each one dies in `duplicate_and_parse` on `kb=None`.

Current `main` picked up the KB-scoped derivation, which fixes fresh installs, but kept honoring the KB-agnostic ones and left all three points above unchanged. The `≤ v0.25.1` form is not recognized at all any more, so an upgraded instance silently re-ingests its entire connector corpus under new ids on the next sync.

## What this PR changes

**1. One owner for the id derivation** — `connector_doc_id_candidates()` / `resolve_connector_doc_id()` in `connector_service.py`, replacing the expression that was inlined at three call sites. A KB-agnostic id is reused **only when it resolves to a row this knowledge base already owns**; everything else gets the KB-scoped id, which cannot collide by construction. The `≤ v0.25.1` form is included, so upgraded instances update their rows in place instead of duplicating them, and `cleanup_stale_documents_for_task` now retains the same id set it can reuse.

**2. The guard becomes recoverable** — a row whose knowledge base no longer exists is unreachable (its chunks went with the index dropped at dataset deletion), so it is discarded and the document is ingested instead of being blocked forever. A genuine conflict still skips, but now names the owning `kb_id` (issue ask #2).

**3. Dataset deletion stops stranding state** — `Connector2Kb` and `SyncLogs` rows for the dataset are dropped, and any document row still pointing at it is swept.

Note on issue ask #1 ("duplicates allowed"): they now *are*, in the sense that matters — the same external document syncs into as many knowledge bases as are linked to the data source, each with its own row. The guard against one dataset overwriting another's document stays.

## Proof

**The reported steps, replayed against the real `resolve_connector_doc_id`** (script in the collapsed section below):

```
1. Two knowledge bases linked to one data source (the reported steps)
  --- v0.25.1 (reporter's version)
      sync KB A     ok    stored                                    [kb-aaaa now holds 2 rows]
      sync KB B     FAIL  collision with knowledge base 'kb-aaaa'   [kb-bbbb now holds 0 rows]
  --- after this fix
      sync KB A     ok    stored                                    [kb-aaaa now holds 2 rows]
      sync KB B     ok    stored                                    [kb-bbbb now holds 2 rows]

2. Instance upgraded from v0.25.1, whose rows carry the KB-agnostic id
  --- main, before this fix
      re-sync KB A  ok    stored                                    [kb-aaaa now holds 4 rows]  <- corpus duplicated
  --- after this fix
      re-sync KB A  ok    updated in place                          [kb-aaaa now holds 2 rows]
```

**The new tests fail on the pre-fix code and pass on this branch.** Reverting only `api/db/services/file_service.py`:

```
FAILED test_upload_document_skips_cross_kb_document_id_collision
  assert "...collision with knowledge base 'kb-other'; skipping update." in
         'collision.txt: Existing document id collision with another knowledge base; skipping update.'

FAILED test_upload_document_reclaims_document_stranded_by_deleted_kb
  AssertionError: assert ['collision.txt: Existing document id collision with another
                          knowledge base; skipping update.'] == []
```

**Full unit suite on this branch:**

```
$ pytest test/unit_test -q
3468 passed, 26 skipped in 81.00s
```

No exclusions; `main` at the same commit reports **3457 passed, 26 skipped**, so the +11 are this PR's new tests.

`ruff format --check` reports the touched files already formatted, and `ruff check` reports no new findings relative to the base commit. Both commits are GPG-signed and each compiles standalone.

<details>
<summary>Reproduction script</summary>

```python
from api.db.services.connector_service import resolve_connector_doc_id
from api.utils.common import hash128

CONNECTOR, DOCS, KB_A, KB_B = "conn-jira", ["XRAY-1234", "XRAY-1235"], "kb-aaaa", "kb-bbbb"

def id_v0_25_1(kb_id, connector_id, external_id, owned):
    return hash128(external_id)

def id_before_fix(kb_id, connector_id, external_id, owned):
    legacy = hash128(f"{connector_id}:{external_id}")
    return legacy if legacy in owned else hash128(f"{kb_id}:{connector_id}:{external_id}")

class Instance:
    """The `document` table, plus the guard `upload_document` applies to it."""

    def __init__(self, scheme):
        self.scheme = scheme
        self.after_fix = scheme is resolve_connector_doc_id
        self.documents = {}  # doc_id -> kb_id
        self.knowledgebases = {KB_A, KB_B}

    def owned_by(self, kb_id):
        return {doc_id for doc_id, owner in self.documents.items() if owner == kb_id}

    def _ingest(self, kb_id, external_id):
        doc_id = self.scheme(kb_id, CONNECTOR, external_id, self.owned_by(kb_id))
        owner = self.documents.get(doc_id)
        if owner is not None and owner != kb_id:
            if owner in self.knowledgebases:
                return f"collision with knowledge base {owner!r}"
            if not self.after_fix:
                return "collision with another knowledge base (owner no longer exists)"
            del self.documents[doc_id]  # stranded row: discarded, then ingested
        reused = doc_id in self.documents
        self.documents[doc_id] = kb_id
        return "updated in place" if reused else "stored"

    def sync(self, kb_id):
        results = [self._ingest(kb_id, e) for e in DOCS]
        failed = sum(1 for r in results if r.startswith("collision"))
        detail = results[0] if len(set(results)) == 1 else "; ".join(results)
        return f"{'FAIL ' if failed else 'ok   '} {detail:<42} [{kb_id} now holds {len(self.owned_by(kb_id))} rows]"

    def seed_v0_25_1_documents(self, kb_id):
        for external_id in DOCS:
            self.documents[hash128(external_id)] = kb_id

print("1. Two knowledge bases linked to one data source (the reported steps)")
for label, scheme in [("v0.25.1 (reporter's version)", id_v0_25_1), ("after this fix", resolve_connector_doc_id)]:
    inst = Instance(scheme)
    print(f"  --- {label}")
    print("      sync KB A    ", inst.sync(KB_A))
    print("      sync KB B    ", inst.sync(KB_B))

print("\n2. Instance upgraded from v0.25.1, whose rows carry the KB-agnostic id")
for label, scheme in [("main, before this fix", id_before_fix), ("after this fix", resolve_connector_doc_id)]:
    inst = Instance(scheme)
    inst.seed_v0_25_1_documents(KB_A)
    print(f"  --- {label}")
    print("      re-sync KB A ", inst.sync(KB_A))
```

</details>

## Note for existing broken instances

No manual SQL is needed after this change. A document blocked by a row whose dataset is gone is unblocked on the next sync (the row is discarded automatically). A document blocked by a row in a dataset that still exists is now reported with that dataset's id, so it can actually be found.

### Type of change

- [x] Refactoring
- [x] Adds new tests

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -18,9 +18,9 @@ import logging
 import os
 import re
 
-from api.db.db_models import File
+from api.db.db_models import Connector2Kb, Document, File, SyncLogs
 from api.db.joint_services.tenant_model_service import get_composite_model_name_by_ids, resolve_model_config, resolve_model_id
-from api.db.services.connector_service import Connector2KbService
+from api.db.services.connector_service import Connector2KbService, SyncLogsService
 from api.db.services.document_service import DocumentService, queue_raptor_o_graphrag_tasks
 from api.db.services.file2document_service import File2DocumentService
 from api.db.services.file_service import FileService
@@ -171,6 +171,14 @@ def _delete_datasets_sync(tenant_id: str, ids: list = None, delete_all: bool = F
     errors = []
     success_count = 0
     for kb_id, kb in kb_id_instance_pairs:
+        # Unwire the data sources first. While these rows live on, the connector
+        # scheduler keeps queueing syncs against a kb_id that is about to stop
+        # resolving, and any document such a run writes outlives its dataset --
+        # an invisible row that later reports a cross-KB id collision against
+        # whatever dataset is linked to the same connector next.
+        Connector2KbService.filter_delete([Connector2Kb.kb_id == kb_id])
+        SyncLogsService.filter_delete([SyncLogs.kb_id == kb_id])
+
         for doc in DocumentService.query(kb_id=kb_id):
             if not DocumentService.remove_document(doc, tenant_id):
                 errors.append(f"Remove document '{doc.id}' error for dataset '{kb_id}'")
@@ -207,6 +215,13 @@ def _delete_datasets_sync(tenant_id: str, ids: list = None, delete_all: bool = F
         if not KnowledgebaseService.delete_by_id(kb_id):
             errors.append(f"Delete dataset error for {kb_id}")
             continue
+
+        # Sweep anything the per-document loop above could not see, so no row is
+        # left pointing at a dataset that no longer exists.
+        stranded = DocumentService.filter_delete([Document.kb_id == kb_id])
+        if stranded:
+            logging.warning("delete_datasets: removed %s stranded document rows for dataset %s", stranded, kb_id)
+
         success_count += 1
 
     if not errors:
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -30,7 +30,7 @@ from api.db.services.tenant_model_service import TenantModelService
 from api.db.services.user_service import TenantService, UserService, UserTenantService
 from api.utils.api_utils import deep_merge, get_parser_config, remap_dictionary_keys, verify_embedding_availability
 from common import settings
-from common.constants import PAGERANK_FLD, FileSource, LLMType, RetCode, StatusEnum
+from common.constants import PAGERANK_FLD, FileSource, LLMType, RetCode, StatusEnum, TaskStatus
 from common.misc_utils import thread_pool_exec, thread_pool_exec_long_time
 from rag.advanced_rag.knowlege_compile.wiki import WIKI_PAGE_COMPILE_KWD
 
@@ -171,13 +171,14 @@ def _delete_datasets_sync(tenant_id: str, ids: list = None, delete_all: bool = F
     errors = []
     success_count = 0
     for kb_id, kb in kb_id_instance_pairs:
-        # Unwire the data sources first. While these rows live on, the connector
-        # scheduler keeps queueing syncs against a kb_id that is about to stop
-        # resolving, and any document such a run writes outlives its dataset --
-        # an invisible row that later reports a cross-KB id collision against
-        # whatever dataset is linked to the same connector next.
-        Connector2KbService.filter_delete([Connector2Kb.kb_id == kb_id])
-        SyncLogsService.filter_delete([SyncLogs.kb_id == kb_id])
+        # Cancel this dataset's queued syncs before touching its documents.
+        # Tasks are only picked up while they are SCHEDULE, so cancelling stops
+        # every run that has not started yet; a sync already in flight is not
+        # interruptible, which is what the stranded-row sweep below covers.
+        SyncLogsService.filter_update(
+            [SyncLogs.kb_id == kb_id, SyncLogs.status.in_([TaskStatus.SCHEDULE, TaskStatus.RUNNING])],
+            {"status": TaskStatus.CANCEL},
+        )
 
         for doc in DocumentService.query(kb_id=kb_id):
             if not DocumentService.remove_document(doc, tenant_id):
@@ -216,8 +217,17 @@ def _delete_datasets_sync(tenant_id: str, ids: list = None, delete_all: bool = F
             errors.append(f"Delete dataset error for {kb_id}")
             continue
 
-        # Sweep anything the per-document loop above could not see, so no row is
-        # left pointing at a dataset that no longer exists.
+        # Unwire the data sources only once the dataset is really gone, so a
+        # failed deletion above leaves a dataset that is still linked and still
+        # syncable. Left behind, these rows keep the connector scheduler queueing
+        # syncs against a kb_id that no longer resolves, and any document such a
+        # run writes outlives its dataset -- an invisible row that later reports
+        # a cross-KB id collision against whatever dataset is linked next.
+        Connector2KbService.filter_delete([Connector2Kb.kb_id == kb_id])
+        SyncLogsService.filter_delete([SyncLogs.kb_id == kb_id])
+
+        # Sweep anything the per-document loop could not see, including rows
+        # written by a sync that was already in flight when deletion started.
         stranded = DocumentService.filter_delete([Document.kb_id == kb_id])
         if stranded:
             logging.warning("delete_datasets: removed %s stranded document rows for dataset %s", stranded, kb_id)
--- a/api/db/services/connector_service.py
+++ b/api/db/services/connector_service.py
@@ -54,6 +54,45 @@ def _append_text_expr(field, suffix: str):
     return fn.COALESCE(field + suffix, suffix)
 
 
+def connector_doc_id_candidates(kb_id: str, connector_id: str, external_id: str) -> tuple[str, str, str]:
+    """Every document id a connector-sourced document may legitimately carry.
+
+    A synced document's primary key is derived from its external id so that
+    re-running a sync updates the existing row instead of duplicating it. Two
+    historical derivations left ``kb_id`` out, which made the key identical for
+    every knowledge base linked to the same data source:
+
+        <= v0.25.1   hash128(external_id)
+        legacy       hash128(f"{connector_id}:{external_id}")
+        current      hash128(f"{kb_id}:{connector_id}:{external_id}")
+
+    Ordered oldest first so callers can prefer an id already on disk.
+    """
+    return (
+        hash128(external_id),
+        hash128(f"{connector_id}:{external_id}"),
+        hash128(f"{kb_id}:{connector_id}:{external_id}"),
+    )
+
+
+def resolve_connector_doc_id(kb_id: str, connector_id: str, external_id: str, owned_doc_ids) -> str:
+    """Pick the document id to sync ``external_id`` into ``kb_id`` under.
+
+    ``owned_doc_ids`` must only contain ids this knowledge base already owns
+    (see ``DocumentService.list_doc_headers_by_kb_and_source_type``). A
+    KB-agnostic id is reused only when it resolves to one of those rows; when it
+    does not, the row it would hit belongs to some other knowledge base and
+    ``FileService.upload_document`` would reject the write as a cross-KB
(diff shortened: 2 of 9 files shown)


sarbojitrana:
Thanks — pushed 7c2e396 addressing all three. Notes on what I verified for each:

**1. Orphan-safe cleanup in `_discard_orphaned_document`** — valid, fixed. The row was stranded precisely because the normal teardown never reached it, so its `Task` rows, linked `File` row and stored blob are stranded too. It now mirrors what `delete_docs` does, minus the chunk work: the chunk index was dropped at dataset deletion, and `get_tenant_id` resolves the tenant *through* the knowledge base, so for a stranded document there is no index left to address. Cleanup is wrapped so debris that cannot be reached still can't resurrect the collision — unblocking ingestion is the point.

**2. Deletion ordering / in-flight sync** — valid, fixed, with one correction to the suggested shape. Cancellation now runs first, while the `Connector2Kb` mapping still exists, and the mapping and log rows are dropped only after `KnowledgebaseService.delete_by_id` succeeds (good catch — the old order unwired a dataset that might survive a failed deletion).

Two details worth flagging:

- I used `SyncLogsService.filter_update(... kb_id == kb_id, status in [SCHEDULE, RUNNING] → CANCEL)` rather than `ConnectorService.cancel_tasks(connector_id)`. `cancel_tasks` is connector-scoped, so it would cancel syncs for *other* datasets sharing the connector. The scoped form is the same primitive `Connector2KbService.link_connectors` already uses when unlinking.
- There is no drain. `sync_data_source.py` never re-reads `SyncLogs.status` mid-run, so a `RUNNING` task cannot be interrupted; tasks are only picked up while `SCHEDULE` ([connector_service.py#L286](https://github.com/infiniflow/ragflow/blob/main/api/db/services/connector_service.py#L286), [#L355](https://github.com/infiniflow/ragflow/blob/main/api/db/services/connector_service.py#L355)), so cancelling closes the whole not-yet-started window. The already-in-flight remainder is what the final stranded-row sweep covers, which is why it runs last. `test_delete_datasets_unwires_connectors_and_sweeps_stranded_documents` now asserts that full ordering.

**3. Oldest-first candidate order** — the inconsistency is real and is fixed; the stated consequence is not. Both orders preserve the duplicate: `connector_doc_id_candidates` feeds `retain_doc_ids` in `cleanup_stale_documents_for_task`, so prune retains *all three* forms and neither row is deleted whichever one the sync updates. The actual reason to prefer oldest-first is convergence — preferring the newest owned id would re-key the document on every future upgrade and strand the older row again each time. Pinned in `test_oldest_owned_id_wins_when_several_upgrades_left_rows_behind`.

**4. Shadowed `id` builtin** — fixed (pre-existing line, but free to clean up while I'm in the file).

Full unit suite on the updated branch: **2646 passed, 25 skipped**. `ruff check` under the repo's config shows a zero-finding delta against the base commit, and `ruff format --check` reports all 8 touched files already formatted.

JinHai-CN:
@sarbojitrana Would you please resolve the conflicts?

mkaaad (approved):
LGTM

Haruko386 (approved):
LGTM, thx for ur contribution

sarbojitrana:
@JinHai-CN Done — rebased onto current `main` (`e97adbd`), conflicts resolved, branch is mergeable again.

Three conflicts, all from `main` moving under the branch:

- **`api/db/services/connector_service.py`** — `main` added the GaussDB expression helpers at the same spot where this PR adds `connector_doc_id_candidates` / `resolve_connector_doc_id`. Purely additive on both sides; both kept.
- **`test/unit_test/api/apps/services/test_delete_datasets.py`** — `main` added a `_chunk_hash` entry to the wiki stub. Kept alongside this PR's stub additions. (This file is CRLF on `main`; I preserved that, so the diff stays a real 61/7 rather than a whole-file rewrite.)
- **`test/unit_test/api/apps/services/test_gaussdb_dataset_embedding.py`** — new on `main`, and it stubs `api.db.db_models` / `common.constants`. It needed the same `Connector2Kb`, `Document`, `SyncLogs`, `SyncLogsService` and `TaskStatus` entries this PR already added to the two sibling stub-based tests, since `dataset_api_service` now imports them.

No behavioural changes — the merged result is the same code the two approvals were given for.

**One correction to my earlier note in the PR description.** I had written that `test/unit_test/agent/component/test_akshare.py` was excluded from my local run because the third-party `scholarly` package fails to collect under Python 3.13. On a clean environment that turns out not to reproduce — it was a transient state in my local venv, not a real exclusion. The full suite now runs with no exclusions at all:

```
$ pytest test/unit_test -q
3468 passed, 26 skipped in 81.00s
```

against **3457 passed, 26 skipped** on `main` at the same commit — the +11 are this PR's new tests. Both commits are GPG-signed and each one compiles standalone.

PR #18245, opened 2026-08-13
Go: align cancel-parse error with Python for un-started documents

euvre:
### Problem

Canceling parsing on a document that was never parsed returned an internal-looking error in the Go backend:

    102 no ingestion task found for document <id>

while the Python backend (`/documents/ingest` with `run=2`) rejects the same request with:

    102 Cannot cancel a task that is not in RUNNING status

### Root cause

`CancelDocParse` treated "no ingestion task exists for this document" as an error unconditionally. Python instead checks a cancel precondition per endpoint — `doc.run in {RUNNING, CANCEL}` or an in-flight (unfinished) parse task — and `cancel_all_task_of` is a tolerant no-op when there is nothing to cancel.

### Fix

- `CancelDocParse` now enforces the Python precondition: cancelable iff `doc.run` is `RUNNING`/`CANCEL`, or an in-flight ingestion task exists (`CREATED`/`RUNNING`/`STOPPING`); otherwise it returns a sentinel `errParseNotRunning`.
- A missing ingestion task is no longer an error by itself (matches `cancel_all_task_of` no-op semantics); when a task exists, `RequestStop` is still issued.
- Each endpoint maps the sentinel to its Python message:
  - `Ingest` (`/documents/ingest`, run=2) → 102 `Cannot cancel a task that is not in RUNNING status`
  - `StopParseDocuments` (`/documents/stop`) → per-document `Can't stop parsing document that has not started or already completed`

This also fixes two adjacent divergences found while testing: canceling an already-completed document was previously accepted in Go (Python rejects it), and re-canceling a document in `CANCEL` state whose task is gone was previously rejected in Go (Python accepts it).

### Tests

- `TestIngest_CancelUnstartedDocument` — reproduces the reported bug; verifies code 102 + Python message
- `TestIngest_CancelCompletedDocument` — completed parse is rejected with the same message
- `TestIngest_CancelUnstartedWithInFlightTask` — enqueued-but-not-started (CREATED task) cancel still succeeds
- `TestIngest_CancelAgainWithoutTask` — re-cancel in CANCEL state succeeds without a task
- Strengthened `TestStopParseDocuments_NotRunningOrCancel` to assert the exact Python `/documents/stop` message

`bash build.sh --test ./internal/service/document/... ./internal/service/chunk/... ./internal/handler/...` all pass.

### DIFF
--- a/internal/service/document/document_ingest.go
+++ b/internal/service/document/document_ingest.go
@@ -105,6 +105,10 @@ func (s *DocumentService) Ingest(ctx context.Context, userID string, req *Ingest
 		if run == string(entity.TaskStatusCancel) {
 			if err = s.CancelDocParse(ctx, doc); err != nil {
 				common.Error(fmt.Sprintf("go side, start to process %s, run is cancel", doc.ID), err)
+				if errors.Is(err, errParseNotRunning) {
+					// Mirror the Python /documents/ingest endpoint's message.
+					return common.CodeDataError, errors.New("Cannot cancel a task that is not in RUNNING status")
+				}
 				return common.CodeDataError, err
 			}
 			if err = s.documentDAO.UpdateByID(ctx, dao.DB, doc.ID, map[string]interface{}{
--- a/internal/service/document/document_parse.go
+++ b/internal/service/document/document_parse.go
@@ -273,7 +273,12 @@ func (s *DocumentService) StopParseDocuments(ctx context.Context, datasetID stri
 	successCount := 0
 	for _, doc := range docs {
 		if cancelErr := s.CancelDocParse(ctx, doc); cancelErr != nil {
-			errs = append(errs, cancelErr.Error())
+			if errors.Is(cancelErr, errParseNotRunning) {
+				// Mirror the Python /documents/stop endpoint's message.
+				errs = append(errs, "Can't stop parsing document that has not started or already completed")
+			} else {
+				errs = append(errs, cancelErr.Error())
+			}
 			continue
 		}
 		successCount++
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Reviewed. The alignment with Python is correct and well-tested. Verified:
- The Python messages match exactly (`document_api.py`: "Cannot cancel a task that is not in RUNNING status" / "Can't stop parsing document that has not started or already completed").
- `common.CREATED/RUNNING/STOPPING` constants and `entity.TaskStatus` values ("0".."3") exist as expected.
- `GetByDocumentID` returns `(nil, nil)` when no task exists, so the `task == nil` branch is handled correctly.

The precondition logic (cancelable iff `run` is RUNNING/CANCEL, or an in-flight CREATED/RUNNING/STOPPING task exists) matches the Python semantics, and the four new tests cover the unstarted / completed / in-flight / re-cancel cases. LGTM.

xugangqiang:
## Suggested test: `doc.run=RUNNING` with no ingestion task

This branch is currently uncovered. On `main`, `CancelDocParse` returns
`"no ingestion task found for document ..."` whenever `task == nil` — so a
document left in `RUNNING` state whose task row is missing would error out.
This PR changes that: when `doc.run` is `RUNNING`, the precondition passes and
the missing task becomes a no-op (the doc is still marked `CANCEL`), matching
Python's `cancel_all_task_of` behavior. That's an improvement, but it has no
test. Here's one to add to `internal/service/document/document_test.go`
(placed next to the other `TestIngest_Cancel*` tests):

```go
// TestIngest_CancelRunningWithoutTask covers the edge case where doc.run is
// RUNNING but the ingestion task row is missing (e.g. cleaned up elsewhere).
// The cancel precondition passes on run=RUNNING, the missing task is a safe
// no-op, and the document is still marked CANCEL — matching Python's
// cancel_all_task_of semantics (this used to error on main).
func TestIngest_CancelRunningWithoutTask(t *testing.T) {
	db := setupServiceTestDB(t)
	pushServiceDB(t, db)
	insertUserTenantForAccessCheck(t, "user-1", "tenant-1")
	insertTestKB(t, "kb-1", "tenant-1", 1, 10, 5)
	insertTestDocWithRun(t, "doc-1", "kb-1", string(entity.TaskStatusRunning), 10, 5)

	svc := testDocumentService(t)
	ctx := t.Context()
	code, err := svc.Ingest(ctx, "user-1", &IngestDocumentRequest{
		DocIDs: []string{"doc-1"},
		Run:    string(entity.TaskStatusCancel),
	})
	if err != nil {
		t.Fatalf("Ingest(cancel) with RUNNING doc and no task: %v", err)
	}
	if code != common.CodeSuccess {
		t.Fatalf("expected code %v, got %v", common.CodeSuccess, code)
	}

	doc, _ := dao.NewDocumentDAO().GetByID(ctx, db, "doc-1")
	if doc == nil || doc.Run == nil || *doc.Run != string(entity.TaskStatusCancel) {
		t.Fatalf("expected doc run=CANCEL, got %v", doc.Run)
	}
}
```

It mirrors `TestIngest_CancelAgainWithoutTask` (same no-task + success shape),
just with `TaskStatusRunning` instead of `TaskStatusCancel` to exercise the
other arm of the `docRun in {RUNNING, CANCEL}` precondition.

PR #18246, opened 2026-08-13
fix(go-agent): forward sys.files uploads to vision LLMs

euvre:
### Summary

**Problem**: On the Go agent canvas, attaching an image to the chat has no effect for vision models (e.g. qwen3-vl-plus) — the model answers that it cannot see any image. The Python agent path works because its prompt builder feeds `sys.files` (parsed text plus `data:image/...;base64` URIs) into the LLM message.

**Root cause** — uploads are dropped in two places along the Go path:
1. `buildAgentInputMessages` (Agent component) never consumes `state.Sys["files"]`; uploads only reach the model when the user manually references `{sys.files}` in the prompt, and even then only as text — images are discarded.
2. `toInternalMessages` (eino → driver bridge) discards `UserInputMultiContent` entirely, keeping only the plain string `Content`, so even correctly-built multi-modal messages lose their image parts before reaching any driver.

**Fix**:
- `buildAgentInputMessages` now collects `sys.files` uploads: non-image entries are merged into the current user prompt, image entries become multi-modal content parts via `userMessageWithImages`, a helper extracted from the LLM component so both components share one implementation.
- `toInternalMessages` now converts eino text/image parts into OpenAI-style content blocks (`[]interface{}` of `{type:text}` / `{type:image_url}`), which the OpenAI-shaped drivers (dashscope/aliyun, openai-compatible, anthropic, google) already understand and marshal verbatim into the request body.

**Validation boundary**: the configured Tongyi key is out of credit, so no live end-to-end model run was possible. Evidence is code-level: parity with the Python agent/LLMBundle image handling, plus new unit tests for both fixed layers (`TestBuildAgentInputMessagesInjectsSysFiles`, `TestBuildAgentInputMessagesSysFilesTextOnly`, `TestToInternalMessagesConvertsMultiModalContent`, `TestToInternalMessagesReassemblesBase64Image`, `TestToInternalMessagesUnsupportedPartsFallBackToString`). Full unit suite passes (106 packages).

### DIFF
--- a/internal/agent/component/agent.go
+++ b/internal/agent/component/agent.go
@@ -276,18 +276,48 @@ func scanAllStreamForToolCall(_ context.Context, stream *schema.StreamReader[*sc
 // buildAgentInputMessages assembles the Python-compatible Agent prompt: the
 // configured history window followed by the current user prompt. The current
 // in-flight user entry is excluded through SnapshotPriorHistory, because the
-// canvas service appends it to state before invoking the workflow.
+// canvas service appends it to state before invoking the workflow. Uploaded
+// files from sys.files are folded into that user prompt (file texts merged,
+// images attached as multi-modal content parts).
 func buildAgentInputMessages(ctx context.Context, p AgentParam) []*schema.Message {
-	current := schema.Message{Role: schema.User, Content: p.UserPrompt}
-	messages := []schema.Message{}
-	if p.MessageHistoryWindowSize > 0 {
-		if state, _, err := runtime.GetStateFromContext[*runtime.CanvasState](ctx); err == nil && state != nil {
-			// Python takes the last 2*N entries from history, which already
-			// contains the current user input, and then removes that final
-			// entry before formatting the configured prompt.
-			priorLimit := p.MessageHistoryWindowSize*2 - 1
-			messages = prependHistory(messages, state.SnapshotPriorHistory(), priorLimit)
+	var state *runtime.CanvasState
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Reviewed. This correctly fixes both halves of the dropped-upload problem: `buildAgentInputMessages` now folds `sys.files` into the user message, and `toInternalMessages` now converts eino multi-modal parts into OpenAI-style content blocks. Verified:
- `collectSysFiles` exists and returns `(textParts, imageURIs)`.
- The drivers handle `[]interface{}` content: `buildChatMessages` passes it verbatim (OpenAI-compatible), and `anthropicContent` / `googleMessageParts` type-switch on `[]interface{}`.
- The text-only path keeps `Content` as a plain string (no multi-modal parts), so existing non-vision flows are unaffected.

One non-blocking concern: `toInternalMessages` is the shared eino→driver bridge, so `Content` becomes `[]interface{}` for any message with multi-modal parts. Most drivers handle it, but the MWS/Bedrock path (`buildMWSChatMessages` in mws.go) requires `Content` to be a string and will now return an error ("MWS chat message content must be a string") for image-bearing messages instead of silently dropping the image. If Bedrock vision is out of scope, consider documenting it; otherwise it may be worth adding multi-modal handling there too.

LGTM overall.

Lynn-Inf (approved):
LGTM

Haruko386 (approved):
LGTM

PR #18247, opened 2026-08-13
fix(wiki): preserve template models without pipelines

NgoQuocViet2001:
### Summary

The Wiki eligibility path supports documents configured directly through `parser_config`, but the model validation added in #18216 rejected those documents when `pipeline_id` was empty.

Use the selected compilation template's `llm_id` for pipeline-less documents, while keeping Compiler model resolution unchanged for pipeline-backed documents. Missing template model configuration still falls through to the existing tenant-default behavior.

### Tests

- `.venv/Scripts/python.exe -m pytest test/unit_test/rag/svr/task_executor_refactor/test_dataset_wiki_generator.py -q`
- `.venv/Scripts/python.exe -m py_compile rag/svr/task_executor_refactor/dataset_wiki_generator.py test/unit_test/rag/svr/task_executor_refactor/test_dataset_wiki_generator.py`
- `uvx ruff check test/unit_test/rag/svr/task_executor_refactor/test_dataset_wiki_generator.py --output-format concise`
- `uvx ruff check rag/svr/task_executor_refactor/dataset_wiki_generator.py --select F821,F822,F823 --output-format concise`
- `uvx ruff format --check rag/svr/task_executor_refactor/dataset_wiki_generator.py test/unit_test/rag/svr/task_executor_refactor/test_dataset_wiki_generator.py`

### DIFF
--- a/rag/svr/task_executor_refactor/dataset_wiki_generator.py
+++ b/rag/svr/task_executor_refactor/dataset_wiki_generator.py
@@ -262,17 +262,23 @@ def _pipeline_compiler_llm_id(pipeline_id: str) -> str | None:
     return None
 
 
-def _validate_wiki_eligible_docs(eligible: list[tuple[dict, str]]) -> dict[str, str | None]:
-    """Validate one Wiki template and return each doc's pipeline chat model."""
+def _validate_wiki_eligible_docs(eligible: list[tuple[dict, str]], tenant_id: str) -> dict[str, str | None]:
+    """Validate one Wiki template and return each document's chat model."""
+    from api.db.services.compilation_template_service import CompilationTemplateService
+
     template_ids = {template_id for _, template_id in eligible}
     if len(template_ids) > 1:
         raise ValueError("Eligible Wiki documents must use the same template")
     pipeline_chat_llm_ids: dict[str, str | None] = {}
-    for doc, _ in eligible:
+    for doc, template_id in eligible:
         doc_id = str(doc.get("id") or "")
         pipeline_id = (doc.get("pipeline_id") or "").strip()
         if not pipeline_id:
-            raise ValueError(f"Wiki document {doc_id} must use a pipeline")
+            template = CompilationTemplateService.get_saved(template_id, tenant_id)
+            config = (template.get("config") or {}) if template else {}
+            llm_id = config.get("llm_id") if isinstance(config, dict) else None
+            pipeline_chat_llm_ids[doc_id] = llm_id.strip() if isinstance(llm_id, str) and llm_id.strip() else None
+            continue
         pipeline_chat_llm_ids[doc_id] = _pipeline_compiler_llm_id(pipeline_id)
     return pipeline_chat_llm_ids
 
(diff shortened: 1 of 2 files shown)


buua436 (commented):
Understood. Your request has been received. Currently, the compilation function only supports parsing documents using the pipeline method.

NgoQuocViet2001:
Thanks for the reply @buua436 — understood, and I'm not trying to argue the feature scope. If pipeline-less documents aren't a supported compilation path, that's your call and I'll close this.

Before you do, there's a smaller problem underneath it that I think is worth a decision either way, because it's inside `main` today rather than something this PR introduces.

**`_wiki_eligible_docs` currently admits pipeline-less documents on purpose.** Its own docstring on `main` says so:

> A doc is eligible when its `parser_config` **OR** its ingestion pipeline resolves to at least one artifacts-kind ("wiki") compilation template

and the body matches — it collects template ids from `_parser_config_compilation_template_ids(pc, tenant_id)` unconditionally, and only *adds* pipeline-derived ones when `pipeline_id` is non-empty. So a document that carries a wiki template in `parser_config` and has no pipeline is returned as eligible.

**`_validate_wiki_eligible_docs` then raises on exactly those documents:**

```python
if not pipeline_id:
    raise ValueError(f"Wiki document {doc_id} must use a pipeline")
```

**And that raise is not caught.** Both call sites invoke it bare:

- `run_wiki`, ~line 1326: `pipeline_chat_llm_ids = _validate_wiki_eligible_docs(eligible)`
- `run_wiki_incremental`, ~line 1682: `... = _validate_wiki_eligible_docs(eligible) if eligible else {}`

Neither is inside a `try`, so the `ValueError` propagates out of the whole run.

The consequence is the part I'd flag: it isn't that one unsupported document is skipped. **A single pipeline-less document aborts wiki compilation for the entire knowledge base**, including every pipeline-backed document in it that is fully supported. Before #18216 those documents compiled; after it, one stray `parser_config` document takes the whole KB down.

So `main` is currently inconsistent with itself: eligibility says these documents are in scope, validation says they're a fatal error.

**Two ways to resolve it, and I'm happy with either:**

1. **This PR as written** — pipeline-less documents resolve their chat model from the selected template's `llm_id`, falling back to the tenant default. Pipeline-backed resolution is untouched.
2. **A much smaller PR, if pipeline-less really is out of scope** — filter them out in `_wiki_eligible_docs` instead, so they're skipped rather than fatal, and update the docstring to drop the `parser_config` OR clause. That's a few lines and it makes the two functions agree.

If you'd prefer (2), say the word and I'll reshape this into that and drop the template-model work entirely. If you'd rather leave the behaviour as-is, close it and no hard feelings — I'd just suggest the docstring be corrected, since it currently promises the `parser_config` path.

buua436 (commented):
If possible, I’d prefer option 2, as it aligns better with our design. Thanks!

buua436:
We will close this PR. If you are interested in implementing Option 2, please either create a new PR or reopen this one and update it accordingly.

PR #18248, opened 2026-08-13
refactor: replace Split in loops with more efficient SplitSeq

questfever:
### Summary

strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

More info: https://github.com/golang/go/issues/61901


questfever:
@yuzhichang @JinHai-CN  Could you please take a look when you have a chance? Thanks!

yuzhichang (approved):

Issue #8813, opened 2025-07-13
[Bug]: online preview uploaded pdf file, chinese characters not show, just show english characters


lx1054331851:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

504e453a

### RAGFlow image version

nightly slim

### Other environment information

```Markdown

```

### Actual behavior

<img width="974" height="671" alt="Image" src="https://github.com/user-attachments/assets/192d1467-86db-4a15-8308-4ba7b48be4cd" />

there is no chinese charactors on the page.

### Steps to reproduce

```Markdown
upload this file:https://static.www.tencent.com/uploads/2022/02/11/f095e3970952aac8956cc4293768249e.pdf
```

PR #18249, opened 2026-08-13
fix(web): pass pdf.js cMap/standard font URLs so CID-font PDFs render in preview

euvre:
## What
Knowledge-base PDF preview renders **blank pages** for PDFs that use built-in CID fonts (very common for CJK documents, e.g. the Tencent PDF in #8813). The canvas paints nothing and the text layer stays empty.

## Root cause
`PdfLoader` (react-pdf-highlighter / pdf.js) was used without `cMapUrl` / `cMapPacked` / `standardFontDataUrl`. pdf.js therefore could not load the required CMaps (`UniGB-UCS2-H`, etc.) and standard font data, failing with console warnings:

```
Warning: CMapReaderFactory not initialized, using default CMapReaderFactory
```

For CID-keyed fonts this means glyphs cannot be resolved, so pages render empty even though the document parses fine.

## Fix
- `web/vite.config.ts`: copy `pdfjs-dist/cmaps/` and `pdfjs-dist/standard_fonts/` into the already-served `public/pdfjs-dist/` directory via the existing `viteStaticCopy` setup (same mechanism already used for `pdf.worker.min.js`).
- `web/src/components/document-preview/pdf-preview.tsx`: pass `cMapUrl="/pdfjs-dist/cmaps/"`, `cMapPacked={true}` and `standardFontDataUrl="/pdfjs-dist/standard_fonts/"` to `PdfLoader` (and extend the local `PdfLoaderProps` type with `standardFontDataUrl`, which react-pdf-highlighter's typings don't expose).

12 lines added, 0 removed.

## Verification (dev server, knowledge-base document preview)
Reproduced with a CJK CID-font PDF (\u56fd\u5bcc\u8bba / Wealth of Nations, Chinese) uploaded to a knowledge base:

| | Before (props removed) | After (this PR) |
|---|---|---|
| textLayer text length | `0` | full Chinese text of the page |
| canvas | 2 blank canvases | 2 canvases with glyphs painted |
| console | `CMapReaderFactory not initialized` warnings | no pdf.js warnings/errors |

With the fix, the text layer contains the complete selectable Chinese text (e.g. \u7b2c\u4e00\u7bc7 \u8bba\u52b3\u52a8\u751f\u4ea7\u529b\u589e\u8fdb\u7684\u539f\u56e0...) and the page visually renders all characters; both `cmaps/` and `standard_fonts/` are served with HTTP 200 under `/pdfjs-dist/`.

## Notes
- Relates to #8813 (same symptom; this fixes the CID-font path for the current preview component).
- Type-check: 223 pre-existing repo-wide `tsc` errors, **none introduced by this change** (the 2 `TS6133` hits in `pdf-preview.tsx` are pre-existing unused callback params untouched by this diff). `oxlint` on the touched files: 0 warnings, 0 errors.

### DIFF
--- a/web/src/components/document-preview/pdf-preview.tsx
+++ b/web/src/components/document-preview/pdf-preview.tsx
@@ -33,6 +33,7 @@ import { getAuthorization } from '@/utils/authorization-util';
 import { useCatchDocumentError } from './hooks';
 type PdfLoaderProps = React.ComponentProps<typeof PdfLoader> & {
   httpHeaders?: Record<string, string>;
+  standardFontDataUrl?: string;
 };
 
 const Loader = PdfLoader as React.ComponentType<PdfLoaderProps>;
@@ -100,6 +101,9 @@ const PdfPreview = ({
           </div>
         }
         workerSrc="/pdfjs-dist/pdf.worker.min.js"
+        cMapUrl="/pdfjs-dist/cmaps/"
+        cMapPacked={true}
+        standardFontDataUrl="/pdfjs-dist/standard_fonts/"
         errorMessage={<FileError>{error}</FileError>}
       >
         {(pdfDocument) => {
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -160,6 +160,14 @@ export default defineConfig(({ mode }) => {
             src: 'node_modules/monaco-editor/min/vs/',
             dest: './',
           },
+          {
+            src: 'node_modules/pdfjs-dist/cmaps/',
+            dest: 'pdfjs-dist/',
+          },
+          {
+            src: 'node_modules/pdfjs-dist/standard_fonts/',
+            dest: 'pdfjs-dist/',
+          },
         ],
       }),
       createHtmlPlugin({


wangq8 (commented):
Reviewed. The fix is correct: CID-keyed PDFs need `cMapUrl`/`cMapPacked` and `standardFontDataUrl` to resolve glyphs. Confirmed:
- react-pdf-highlighter's `PdfLoader` spreads all props into `getDocument` (`Object.assign({}, this.props, ...)`), so `standardFontDataUrl` is forwarded despite the missing typing; extending `PdfLoaderProps` is the right fix.
- viteStaticCopy copies the source directory name into dest (same pattern as `monaco-editor/min/vs` → `/vs`), so the targets produce `public/pdfjs-dist/cmaps/` and `public/pdfjs-dist/standard_fonts/`, matching the URLs.

Two minor, non-blocking notes:
1. The description says this uses the "same mechanism already used for pdf.worker.min.js" — actually `pdf.worker.min.js` is checked in under `web/public/pdfjs-dist/`, not copied by viteStaticCopy. The mechanism used here (viteStaticCopy) is correct and already proven for monaco, just not for the worker.
2. Copying the full `cmaps/` directory adds non-trivial static assets to the build output; with `cMapPacked=true` only the `.bcmap` files are needed, so the copy could potentially be trimmed (optional).
LGTM.

dcc123456 (approved):

PR #18253, opened 2026-08-13
fix(web): make the selected agent-explore session visually distinct

euvre:

In an agent's explore page, clicking between similar-looking sessions showed no visible selection change.

Root cause: the selected session card was styled with `aria-selected:bg-bg-card`, but in `web/tailwind.css` the `--bg-card` and `--bg-input` tokens are identical in **both** light and dark themes, so a selected card rendered exactly like an unselected one. The selection state pipeline itself (click → `sessionId` URL param → `selected` prop) was working correctly; only the visual style was a no-op.

### Changes

- `web/src/pages/agent/explore/components/session-card.tsx`: replace the invisible selected style with an accent tint (`border-accent-primary bg-accent-primary/10`), consistent with selection styling used elsewhere in the codebase (e.g. timeline's `text-accent-primary bg-accent-primary/10`). The change is keyed off the existing `selected` prop; `tailwind-merge` in `cn()` correctly drops the conflicting base `border-border-button bg-bg-input` classes.
- Removed `canvas-card.tsx`: dead code with zero imports that carried the same latent styling issue.

### Verification

- `oxlint`: 0 warnings/errors on the touched package.
- `tsc --noEmit`: only pre-existing errors (identical on clean main); none introduced by this change.
- E2E on local dev server: created two identically-named sessions in an agent explore page and clicked between them. Selected card now computes to `background: rgba(0, 190, 180, 0.1)` + `border: rgb(0, 190, 180)` vs unselected `rgba(255, 255, 255, 0.05)` / `rgba(255, 255, 255, 0.1)`, and the highlight correctly follows the click.

### Type of change


wangq8 (commented):
Reviewed. Root cause is correctly identified (`--bg-card` and `--bg-input` are identical in both themes, making `aria-selected:bg-bg-card` a visual no-op), and the replacement (`border-accent-primary bg-accent-primary/10`) is consistent with selection styling used elsewhere. Confirmed `cn()` uses tailwind-merge, so the conflicting base classes are correctly dropped, and `canvas-card.tsx` is indeed dead code (zero imports). LGTM.

cike8899 (approved):

PR #18254, opened 2026-08-13
fix(go): parse PPTX one chunk per slide

euvre:
## Problem

In Go parse mode, a PPTX deck with multiple slides produced exactly **1 chunk** instead of one chunk per slide. E.g. a 2-page deck (titles 'Slide One' / 'Slide Two') yielded a single merged chunk containing both slides' text.

Root cause: `PPTXParser.ParseWithResult` called office_oxide's `PlainText()`, which renders slides back-to-back **without any page delimiter**, and then split on form-feed (`\f`). With no `\f` present, all slides collapsed into a single JSON item → one chunk.

## Fix

Derive per-slide text from office_oxide's structured IR (`ToIRJSON`), which carries **one section per slide**. `ParseWithResult` now emits one JSON item per slide, each with its own `text`, `doc_type_kwd: text`, and `slide_number` — mirroring the python `RAGFlowPptParser` contract ('Every page will be treated as a chunk'). Slides without extractable text still yield their own (empty) item, matching python's one-entry-per-slide behavior. A zero-section IR falls back to the whole-document `PlainText` so a readable file still yields one chunk instead of none.

## Testing

- New CGO regression tests in `ppt_parser_cgo_test.go`:
  - `TestPPTXParser_ParseWithResult_MultiSlide_CGO`: 2-slide deck → 2 JSON items with correct per-slide text and `slide_number` (1, 2).
  - `TestPPTXParser_ParseWithResult_EmptySlide_CGO`: 3-slide deck with empty middle slide → 3 items, middle one with empty text and `slide_number`=2.
- `bash build.sh --test ./internal/parser/...` — all pass.
- E2E: created a Presentation-mode dataset via the Go API, uploaded a 2-slide `.pptx`, parsed → `chunk_count = 2`, one chunk per slide ('Slide One Title…' / 'Slide Two Title…').

### DIFF
--- a/internal/parser/parser/docx_ir.go
+++ b/internal/parser/parser/docx_ir.go
@@ -15,10 +15,12 @@
 //
 
 // Package parser: this file holds the pure-Go office_oxide IR data
-// model and JSON helpers for DOCX. It is intentionally cgo-free so
-// that postprocessing (docx_postprocess.go) and unit tests run
-// without the office_oxide native library. The cgo boundary lives
-// entirely in docx_parser.go (officeOxide.OpenFromBytes).
+// model and JSON helpers shared by the DOCX and presentation parsers.
+// It is intentionally cgo-free so that postprocessing
+// (docx_postprocess.go), per-slide section building (pptx_ir.go), and
+// unit tests run without the office_oxide native library. The cgo
+// boundary lives in docx_parser.go / pptx_parser.go
+// (officeOxide.OpenFromBytes).
 
 package parser
 
@@ -55,6 +57,7 @@ type docxIRElement struct {
 	Level   int              `json:"level"`   // heading level (1-6) or list nesting level
 	Style   string           `json:"style"`   // Word style name (e.g. "Normal", "Heading 1")
 	Content json.RawMessage  `json:"content"` // rich text runs or block-level content; decoded per type
+	Text    string           `json:"text"`    // payload of a bare "text" block; not part of the current IR schema
 	Data    []byte           `json:"data"`    // raw image bytes (for "image" type)
 	Rows    []docxIRRow      `json:"rows"`    // table rows
 	Ordered bool             `json:"ordered"` // true=numbered list, false=bullet list (for "list" type)
@@ -93,7 +96,7 @@ type docxIRList struct {
 }
 
 type docxIRRun struct {
-	Type    string          `json:"type"` // "text", "image"
+	Type    string          `json:"type"` // "text", "line_break", "image"
 	Text    string          `json:"text"`
 	Content []docxIRElement `json:"content"` // nested elements (used in table cells)
 }
@@ -106,11 +109,17 @@ type docxIRCell struct {
 	Content []docxIRElement `json:"content"` // nested paragraphs inside table cell
 }
 
+// joinDOCXIRRuns concatenates inline runs into plain text. Hard line
+// breaks become newlines so multi-line text keeps its line structure;
+// non-text payloads (e.g. inline images) contribute no characters.
 func joinDOCXIRRuns(runs []docxIRRun) string {
 	var b strings.Builder
 	for _, r := range runs {
-		if r.Type == "text" {
+		switch r.Type {
+		case "text":
 			b.WriteString(r.Text)
+		case "line_break":
+			b.WriteByte('\n')
 		}
 	}
 	return b.String()
@@ -193,14 +202,19 @@ func docxIRTableToHTML(el docxIRElement) string {
 }
 
 // docxElementText returns the plain-text rendering of any supported
-// IR element type. Used by extractDOCXFiguresFromIR so that tables,
-// lists, and text boxes contribute to image surrounding context
-// instead of becoming empty flatBlocks (which would drop adjacent
-// VLM context). Returns "" for image and unknown types.
+// IR element type. Used by buildPPTXJSONSections to flatten one slide
+// section, and by extractDOCXFiguresFromIR so that tables, lists, and
+// text boxes contribute to image surrounding context instead of
+// becoming empty flatBlocks (which would drop adjacent VLM context).
+// A bare "text" block is not part of the current IR schema, but is
+// passed through so its payload is never silently dropped. Returns ""
+// for image and unknown types.
 func docxElementText(el docxIRElement) string {
 	switch el.Type {
 	case "paragraph", "heading":
 		return joinDOCXIRRuns(el.contentRuns())
+	case "text":
+		return el.Text
 	case "table":
 		var lines []string
 		for _, row := range el.Rows {
--- a/internal/parser/parser/docx_ir_test.go
+++ b/internal/parser/parser/docx_ir_test.go
@@ -65,6 +65,19 @@ func TestJoinDOCXIRRuns(t *testing.T) {
 	}
 }
 
+// TestJoinDOCXIRRuns_LineBreak pins that a hard line break inline run
+// renders as a newline instead of being dropped.
+func TestJoinDOCXIRRuns_LineBreak(t *testing.T) {
+	runs := []docxIRRun{
+		{Type: "text", Text: "before"},
+		{Type: "line_break"},
+		{Type: "text", Text: "after"},
+	}
+	if got := joinDOCXIRRuns(runs); got != "before\nafter" {
+		t.Fatalf("joinDOCXIRRuns = %q, want %q", got, "before\nafter")
+	}
+}
+
 // TestExtractDOCXFiguresFromIR verifies the image-figure context
 // extraction: an image block carries the immediately surrounding text
 // as ContextAbove / ContextBelow / Marker. Pure-Go, runs under !cgo.
--- a/internal/parser/parser/ppt_parser_cgo_test.go
+++ b/internal/parser/parser/ppt_parser_cgo_test.go
@@ -99,8 +99,8 @@ func TestPPTXParser_ParseWithResult_MultiSlide_CGO(t *testing.T) {
 }
 
 // TestPPTXParser_ParseWithResult_EmptySlide_CGO verifies that a slide
-// without extractable text still yields its own JSON item (empty text),
-// matching the python parser which appends one text per slide.
+// without extractable text still yields its own JSON item with an
+// empty text field, so slide numbering stays aligned with the deck.
 func TestPPTXParser_ParseWithResult_EmptySlide_CGO(t *testing.T) {
 	ctx := t.Context()
 	p := NewPPTXParser()
--- a/internal/parser/parser/ppt_parser_cgo_test.go
+++ b/internal/parser/parser/ppt_parser_cgo_test.go
@@ -50,6 +50,88 @@ func TestPPTXParser_ParseWithResult_CGO(t *testing.T) {
 	}
 }
 
+// TestPPTXParser_ParseWithResult_MultiSlide_CGO is the regression test
+// for the one-chunk-per-slide contract: a multi-slide deck must emit
+// one JSON item per slide, each carrying that slide's text only.
+// office_oxide's PlainText concatenates slides without a delimiter, so
+// splitting must come from the structured IR sections.
+func TestPPTXParser_ParseWithResult_MultiSlide_CGO(t *testing.T) {
+	ctx := t.Context()
+	p := NewPPTXParser()
+
+	w := officeOxide.NewPptxWriter()
+	s1 := w.AddSlide()
+	w.SetSlideTitle(s1, "Slide One Title")
+	w.AddSlideText(s1, "First slide body about the Alpha topic.")
+	s2 := w.AddSlide()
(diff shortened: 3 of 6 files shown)


wangq8 (commented):
Reviewed. The fix is correct and well-targeted: deriving per-slide text from office_oxide's structured IR (`ToIRJSON`, one section per slide) instead of splitting `PlainText` on `\f` is the right approach, and it matches the Python `RAGFlowPptParser` "one chunk per slide" contract. The CGO regression tests cover the multi-slide and empty-slide cases well.

Verified: `ToIRJSON` returns a top-level `sections` key, and the `PptxWriter` helpers (`AddSlide`/`SetSlideTitle`/`AddSlideText`/`ToBytes`) exist. LGTM.

Two minor, non-blocking notes:
1. `doc_type_kwd` is emitted as `"text"`, but the Python slide parser sets `"image"` (slides are rendered/OCR'd there). This is pre-existing (unchanged by this PR), but worth confirming intent since the description says it mirrors the Python contract.
2. `irBlockLines` returns nil for a node whose own `Type == "text"` with empty `Content` (the `runsOnly` path only inspects children). Unlikely with the real IR where `text` nodes are leaf runs, but a top-level text element would be dropped. Consider handling `n.Type == "text" && n.Text != ""` directly for robustness.

xugangqiang (changes requested):
# Review: PR #18254 — fix(go): parse PPTX one chunk per slide

**Verdict: I do not agree to merge in the current state. Requesting changes.**

The direction is correct (per-slide chunking is a real bug fix), but the new IR walker silently drops bullet-list text — a confirmed regression for the most common PPTX content — and duplicates logic that already exists in `docx_ir.go`. Please address the issues below before merging.

---

## 1. Problem declared vs. actually solved

Declared: in Go parse mode a multi-slide PPTX produced a single merged chunk because `PlainText()` concatenates slides with no delimiter and the `\f` split failed; fix is to emit one chunk per slide, matching Python `RAGFlowPptParser`.

- ✅ Core problem solved: `ToIRJSON()` yields one `section` per slide, and `ParseWithResult` now emits one JSON item per slide with its own `text`, `doc_type_kwd: text`, and `slide_number`. Multi-slide and empty-slide E2E (chunk_count=2) and unit tests confirm it.
- ⚠️ But a content-fidelity regression is introduced (see §3). So "solved" is only partially true: chunk count is right, but some text (lists) is lost.

`slide_number` (1-based), `doc_type_kwd: text`, and keeping an empty item for empty slides all match Python `RAGFlowPptParser.__call__`, which returns one string per slide (empty string for empty slides). Consistent.

## 2. Better approach

**Reuse the existing `internal/parser/parser/docx_ir.go` instead of writing a parallel walker.**

Evidence: `docx_ir.go` already models the same office_oxide IR for this pinned version:
- `docxIRElement` already models `Items` (`[]docxIRListItem`, with `Nested` sub-lists), `Rows`, and `Content`.
- `docxElementText` / `extractTextFromListItem` already correctly handle ordered/unordered lists, nested lists, tables, text boxes, and headings.
- Both files are in `package parser`, so they are directly callable.

PPTX `ToIRJSON` uses the same `sections → elements` shape as DOCX. The cleaner implementation is: `json.Unmarshal(irJSON, &docxIRDocument)`, then for each section join `docxElementText(section.Elements)` and emit one item. Benefits:
- No ~90 lines of new duplicate types (`presentationIR`/`irNode`/`irBlockLines`).
- List / nested-list / `line_break` handling is inherited from already-tested code — eliminates the §3 regression for free.
- One IR model shared between DOCX and PPTX, maintained in one place.

If isolation is preferred, the minimal fix is to add `Items []irNode` to `irNode` and traverse it (incl. `Nested`); but reuse is strongly preferred.

## 3. Test coverage & regression risk

**Coverage is insufficient and a real regression is introduced.**

Tests cover only: multi-slide (2 pages) and a middle empty slide (3 pages). Missing the critical regression cases:
- ❌ Bullet/numbered lists — the most frequent PPTX content shape.
- ❌ Tables (the `rows` branch exists but is untested).
- ❌ Hard line breaks (`line_break`).
- ❌ Image-only slides.

Regression analysis (verified against `docx_ir.go` for the real IR schema):
- Old code used `PlainText()`, which includes list text.
- New `irBlockLines` on a `list` element: `irNode` has no `Items` field, so `n.Content` is empty → `runsOnly` is vacuously true → returns `nil`. **The entire slide's bullet-list text is dropped.** This is actual content loss, not formatting.
- `line_break`: a paragraph's inline text runs are themselves `irNode`s carrying `Text`, so recursion preserves the text (just without a newline). CodeRabbit overstates the "text lost" claim for line breaks, but the missing newline is still a fidelity gap worth handling.
- Top-level `type:"text"` node with empty `Content` but non-empty `Text`: `runsOnly` returns `nil` → dropped (wangq8 already flagged; rare but real).

Net: the current tests give a false "pass" while list content regresses. Add CGO regression cases for bullet lists, tables, and line breaks.

## 4. Does merging improve the system

Partially, but **not in the current state**.

- Improvement: fixes the long-standing "multi-slide PPTX collapsed into 1 chunk" bug, giving correct per-slide retrieval granularity, aligned with the Python path.
- Offset: for decks with bullet lists (extremely common in business/teaching PPTX), list text is dropped entirely, degrading retrieval quality.
- Recommendation: merge only after list (and at least table) traversal is fixed and covered by regression tests. If merged prematurely, document the known limitation that bullet-list text is not preserved.

## 5. Are the existing comments reasonable, and how to fix

**(a) CodeRabbit #1 — remove "Python parser" compatibility wording in comments. Reasonable, minor.**
Per project guideline "do not add new compatibility wording in comments/docs." Changes are comment-only. Fix: reword `pptx_parser.go:117-121 / 131-133` and `ppt_parser_cgo_test.go:101-103` from "mirroring the python RAGFlowPptParser" to factual descriptions ("structured IR carries one section per slide", "empty slides still yield an empty item"). Non-blocking, do it along the way.

**(b) CodeRabbit #2 — list items / line breaks not handled in `irBlockLines`. Reasonable and BLOCKING.**
This is the core defect (see §3); verified. Fix (preferred): route PPTX through `docxIRDocument`/`docxElementText` so lists and nested lists are covered automatically. Fallback: add `Items []irNode` (with `Nested`) to `irNode` and recurse; also emit `\n` for `line_break`. Add CGO regression tests for bullet lists and hard line breaks.

**(c) wangq8 (human) — both points reasonable.**
1. `doc_type_kwd` is hardcoded to `"text"`, whereas the Python slide parser sets `"image"` (slides get rendered/OCR'd). This is **pre-existing** (the old `pptx_parser.go` also hardcoded `"text"`), not introduced here. Suggestion: keep this PR minimal and track the `"image"` alignment as a separate follow-up issue (known tech-debt).
2. Top-level `type:"text"` empty-`Content` node drop — reasonable, low priority. Fix: at the start of `irBlockLines`, `if n.Type == "text" && strings.TrimSpace(n.Text) != "" { return []string{strings.TrimSpace(n.Text)} }`.

---

**Summary:** The per-slide chunking fix is correct, but the new IR walker is both weaker than and duplicate of the existing `docx_ir.go`, and it drops bullet-list text (confirmed regression). Actions before merge: (1) reuse `docx_ir.go`'s IR model + `docxElementText`; (2) add regression tests for lists/tables/line breaks; (3) clean up the Python-compat wording in comments; (4) track the `doc_type_kwd:"image"` discrepancy as a separate issue. I am not approving the merge until (1) and (2) are addressed.

euvre:
Thanks for the reviews — the substantive items are addressed in 429245405.

**Structure / schema (wangq8)**
- Replaced the private IR walker with the shared `docx_ir.go` model: `pptx_parser.go` now decodes `ToIRJSON()` through a new cgo-free `buildPPTXJSONSections` (in `pptx_ir.go`), and text extraction goes through the shared `docxElementText`, so paragraphs, headings, text boxes, tables, and nested lists are all flattened into per-slide text.
- `joinDOCXIRRuns` now renders `line_break` runs as `\n` instead of dropping them, and a bare `text` IR block passes its payload through rather than being silently skipped.
- Reworded the `docx_ir.go` schema docs to describe the model as shared by the DOCX and presentation parsers.

**Comments / rationale (xugangqiang)**
- Removed the "mirrors python" comments; the IR-path comment now states the actual reason: office_oxide's `PlainText` renders slides back-to-back with no delimiter, so per-slide splitting must come from the structured IR.

**Tests**
- New pure-Go `pptx_ir_test.go` (runs without native libs): multi-slide, empty slide retained, nested list, table cells, hard line break, bare text block, image-only slide, invalid JSON, and the no-sections case.
- `docx_ir_test.go`: added the `line_break` case.
- `bash build.sh --test ./internal/parser/parser/...` passes, including the existing CGO end-to-end PPTX tests against the real office_oxide IR.

xugangqiang:
**Gaps / regression risks (review from current head `4292454`):**

- **Missing test for consecutive line breaks.** The existing hard-line-break case uses a *single* break (`"before\nafter"`) and passes by luck. There is no case with *two* consecutive breaks, so the bug in `buildPPTXJSONSections` (which splits on `\n` and drops empty lines) is not caught. Add a regression case for `before\n\nafter` (expect the blank line preserved).

- **Behavior change for empty decks.** Old code: a blank deck (no `\f`, whitespace only) yielded 1 empty-text item. New code: no sections **and** `PlainText` empty → yields **0 items**. This is arguably better, but it is a behavioral change and should be called out in the PR description.

- **Empty slides still emit a chunk** (matching Python "one chunk per slide"). For an image-only deck this produces N empty-text chunks. This is consistent with the parity goal, but whether empty chunks should be filtered is a design trade-off worth documenting.

- **`slide_number = i + 1` assumes exactly one section per slide.** If office_oxide ever emits multiple sections for a single slide, the parser would over-produce items and misnumber them. This is an assumption — low real risk, but worth an assertion/comment so the invariant is explicit.

- **Images.** `docxElementText` returns `""` for `image`, so a slide mixing text and image drops the image (no `image` key on the item). Note: the old `PlainText` path also dropped images, so this is **not** a new regression — just recorded for awareness.

- **`ToIRJSON` error regression.** Currently if `doc.ToIRJSON()` errors, `ParseWithResult` returns an error immediately, whereas the old `PlainText` path might have salvaged text. Consider falling back to `PlainText` + `\f` split (worst case = old 1-chunk behavior) on `ToIRJSON` failure, so parsing capability never regresses. This is a rare but real surface.

euvre:
Hi @xugangqiang — thanks for the re-review of `4292454`. Follow-ups landed in `2783014`, answers item by item:

1. **Consecutive line breaks** — regression case added (`consecutive hard line breaks collapse to one newline`). One correction on the expectation: the blank line is **not** preserved, intentionally. The Python slide parser skips empty paragraphs and joins non-empty texts with a single `\n` (`deepdoc/parser/ppt_parser.py`, `__extract`: `if paragraph.text.strip(): texts.append(...)` + `"\n".join(texts)`), so `before\n\nafter` normalizes to `before\nafter`. The single-break case was indeed passing by luck; both cases are now pinned, and the normalization intent is documented at the split site in `pptx_ir.go`.
2. **Empty decks (old: 1 empty-text item → new: 0 items)** — agreed, deliberate change (a text-free deck has nothing to chunk); it is pinned by the `no sections yields no items` case, and a sectionless deck whose whole-document plain text is non-empty still yields one salvaged item. Called out here since I cannot edit the PR description from this workflow.
3. **Empty slides still emit a chunk** — intentional, documented in the `buildPPTXJSONSections` doc comment: the empty item keeps `slide_number` aligned with the source deck (one-chunk-per-slide contract). Kept as-is.
4. **`slide_number = i + 1` invariant** — already explicit in the `buildPPTXJSONSections` doc comment ("The IR carries exactly one section per slide, so section index + 1 is the slide number"). I did not add a runtime assertion in the parser; an office_oxide schema change on this would surface through the decode path and the CGO tests.
5. **Images** — acknowledged; identical to the pre-change plain-text path (no regression), stays tracked with the `doc_type_kwd:"image"` alignment as a separate follow-up, as you suggested earlier.
6. **`ToIRJSON` error fallback** — applied. On `ToIRJSON` failure the parser now salvages whole-document plain text as a single trimmed item (worst case: the entire deck as one chunk); the original IR error is surfaced only when plain text fails too. The salvage is shared with the sectionless-IR branch via the new pure-Go `itemsFromPlainText` helper. One deviation from your sketch: no `\f` split — office_oxide plain text carries no slide delimiter (the original bug), so the whole-document single item is the only sound fallback.

Verified with `bash build.sh --test ./internal/parser/parser/...` (passes; includes the pure-Go IR tests and the CGO end-to-end PPTX suite). No shared-service end-to-end run on this line — verification is unit tests plus the Python-path comparison above.

Hi @wangq8 — thanks for the LGTM. Your two notes: the bare top-level `text` node pass-through landed in `4292454` (`docxElementText` `case "text"`, covered by the `bare text block passes through` test case), and the `doc_type_kwd:"image"` alignment remains deferred as a separate follow-up as agreed.

xugangqiang:
LGTM

PR #18259, opened 2026-08-14
fix(web): make chat attachment images click-to-zoomable

euvre:
### Summary

Images attached to chat messages were rendered as plain 40px thumbnails with no way to view them at full size.

This PR wraps image attachments in `react-photo-view` (already used elsewhere in the app) so clicking a thumbnail opens a zoomable lightbox, while non-image files keep the existing open-in-new-tab behavior.

Two details worth noting:

- Image detection now works for all three file shapes this component receives (`File`, `IDocumentInfo`, `UploadResponseDataType`) via `isImageFile()`/`getFileMimeType()` helpers, falling back to the extension when no MIME type is available.
- Server-stored chat uploads are kept as blobs in the per-user downloads bucket with **no `document`/`file` table row**, so the document image endpoint cannot serve them. Their thumbnails are therefore fetched through the authenticated agents attachment preview endpoint (`getAttachmentFilePreview`, with `responseType: blob` via the existing `useAuthenticatedImageUrl` hook) and rendered from an object URL; a `FileIcon` is shown while loading.

Verified end-to-end in the browser: clicking a chat image attachment opens the lightbox with the fully decoded image (2035×1593 PNG), zoom/rotate controls work, and non-image attachments are unaffected.

### DIFF
--- a/web/src/components/next-message-item/__tests__/utils.test.ts
+++ b/web/src/components/next-message-item/__tests__/utils.test.ts
@@ -1,0 +1,66 @@
+import { UploadResponseDataType } from '@/interfaces/database/chat';
+import { IDocumentInfo } from '@/interfaces/database/document';
+import { getFileMimeType, isImageFile } from '../utils';
+
+describe('getFileMimeType', () => {
+  it('uses the browser-provided type for local File objects', () => {
+    const file = new File(['x'], 'photo.png', { type: 'image/png' });
+    expect(getFileMimeType(file)).toBe('image/png');
+  });
+
+  it('reads mime_type from uploaded file metadata', () => {
+    const file = {
+      name: 'photo.webp',
+      mime_type: 'image/webp',
+    } as UploadResponseDataType;
+    expect(getFileMimeType(file)).toBe('image/webp');
+  });
+
+  it('returns an empty string when no MIME metadata exists', () => {
+    const file = { name: 'scan.png' } as IDocumentInfo;
+    expect(getFileMimeType(file)).toBe('');
+  });
+});
+
+describe('isImageFile', () => {
+  it('detects images by MIME type', () => {
+    const file = new File(['x'], 'photo', { type: 'image/png' });
+    expect(isImageFile(file)).toBe(true);
+  });
+
+  it('falls back to the filename extension when MIME metadata is missing', () => {
+    const file = new File(['x'], 'photo.png', { type: '' });
+    expect(isImageFile(file)).toBe(true);
+  });
+
+  it('matches extensions case-insensitively', () => {
+    const file = { name: 'diagram.SVG' } as IDocumentInfo;
+    expect(isImageFile(file)).toBe(true);
+  });
+
+  it('detects uploaded files through their mime_type', () => {
+    const file = {
+      name: 'photo.avif',
+      mime_type: 'image/avif',
+    } as UploadResponseDataType;
+    expect(isImageFile(file)).toBe(true);
+  });
+
+  it('lets explicit MIME metadata win over the filename extension', () => {
+    const file = {
+      name: 'fake.png',
+      mime_type: 'application/pdf',
+    } as UploadResponseDataType;
+    expect(isImageFile(file)).toBe(false);
+  });
+
+  it('passes non-image files through as non-image', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this PR — nice, well-scoped improvement. Reviewed the diff against `main` and verified the referenced helpers and backend endpoint.

**What's good**

- Correct reuse of existing infrastructure: `react-photo-view` (already imported globally in `app.tsx` and used in `reference-image-list.tsx`), `useAuthenticatedImageUrl` for auth'd blob fetching, and the authenticated `GET /api/v1/agents/attachments/<id>/preview` endpoint. This is the right approach for server-stored chat uploads (which live in the per-user bucket with no document/file row).
- `isImageFile()`/`getFileMimeType()` correctly handle all three shapes (`File`, `IDocumentInfo`, `UploadResponseDataType`). Note `IDocumentInfo` has no `mime_type` field, so it falls through to the extension check — which works because `getExtension` lowercases.
- The `instanceof File` narrowing is clean and the `id` handling in the `!isFile` branch is type-safe.
- Backend `_attachment_request_metadata()` reads `mime_type` from the query string, matching `getAttachmentFilePreview`'s `mime_type` param, so the content-type is resolved correctly.

**Minor / non-blocking nits**

1. `LocalFileImage` creates the object URL inside `useMemo(() => URL.createObjectURL(file), [file])` and never revokes it. Creating object URLs is a side effect and shouldn't live in `useMemo` (React may drop the memo cache and re-run it, leaking the previous URL). Consider moving it to a `useEffect` with a `URL.revokeObjectURL` cleanup, e.g.:
   ```tsx
   const [objectUrl, setObjectUrl] = useState('');
   useEffect(() => {
     const url = URL.createObjectURL(file);
     setObjectUrl(url);
     return () => URL.revokeObjectURL(url);
   }, [file]);
   ```

2. Minor description inaccuracy: the summary mentions `responseType: blob`, but `useAuthenticatedImageUrl` uses native `fetch` + `response.blob()` rather than axios. Harmless, but worth a small correction for future readers.

3. No test coverage added. Given this is a pure presentational component with a non-trivial image-detection branch, a small unit test for `isImageFile` (MIME vs. extension fallback, case-insensitivity, non-image passthrough) would help prevent regressions.

None of these block merging — the core change is correct and matches existing patterns. Nice work.

euvre:
Thanks @wangq8 for the careful review — the nits are addressed in 432f9ce.

**Applied**

1. **Object URL lifecycle (nit 1):** `LocalFileImage` now creates the object URL in a `useEffect` and revokes it in the effect cleanup on file change/unmount, instead of memoizing a URL that leaks if React drops the memo cache. The component renders nothing until the URL is set (one render pass, same empty-start pattern as `useAuthenticatedImageUrl`).
2. **Test coverage (nit 3):** moved the pure helpers (`getFileMimeType`, `isImageFile`) into `next-message-item/utils.ts` and added `__tests__/utils.test.ts` with 9 cases: MIME-based detection, extension fallback when MIME is absent, case-insensitivity, all three file shapes (`File` / `IDocumentInfo` / `UploadResponseDataType`), non-image passthrough, and explicit-MIME-wins precedence. All pass (`jest src/components/next-message-item/__tests__/utils.test.ts`: 9/9).
3. **Description note (nit 2):** you're right — `useAuthenticatedImageUrl` uses native `fetch` + `response.blob()`, not axios `responseType: blob`. Correcting the record here for future readers since the summary text overstated it.

**Checked but not applied (from the automated review)**

- *"Restore non-image attachment navigation":* verified against `main` — the non-image branches have always rendered bare `FileIcon`/`SvgIcon` with no link (the `NewDocumentLink` in `NameWidget` was already commented out before this PR), so no navigation was removed and there is nothing to restore.
- *Extension fallback for generic MIME (e.g. `application/octet-stream` + `.png` name):* kept the current semantics — explicit MIME metadata wins, extension is only a fallback when MIME is absent (the case that actually occurs for `IDocumentInfo`, which has no `mime_type`). This matches the behavior validated in review, and the precedence is now pinned by a test.

**Verification:** targeted jest run 9/9 green; `npm run type-check` reports no errors in any touched file (the repo carries pre-existing errors in unrelated files). No browser/shared-service e2e from this side — verification is unit tests + type-check + line-by-line comparison against `main`.

dcc123456 (approved):

PR #18261, opened 2026-08-14
fix: exclude disabled documents from wiki compilation

buua436:

- Prevent disabled documents from participating in KB-level Wiki compilation.
- Exclude disabled documents from cached MAP results and MAP input hash calculation.
- Extend Wiki alteration detection to report documents affected by chunk additions, deletions, hash changes, or document disabling.

### Type of change

### DIFF
--- a/api/apps/restful_apis/dataset_api.py
+++ b/api/apps/restful_apis/dataset_api.py
@@ -833,7 +833,9 @@ async def get_wiki_alteration(tenant_id, dataset_id):
     GET /api/v1/datasets/<dataset_id>/artifacts/alteration?kind=<kind>
     ``kind`` (default ``wiki``) is one of: wiki | graph | mindmap | timeline |
     tree (tree covers both ``tree`` and ``page_index``).
-    Success: {"code": 0, "data": {"removed": int, "newly_uploaded": int, ...}}
+    Success: {"code": 0, "data": {"removed": int, "newly_uploaded": int, ...}}.
+    Wiki responses also include ``changed`` and ``changed_doc_ids`` based on
+    source chunk hash drift.
     """
     kind = (request.args.get("kind") or "wiki").strip().lower()
     if kind not in {"wiki", "graph", "mindmap", "timeline", "tree"}:
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -2269,9 +2269,9 @@ async def _wiki_chunk_alteration(
                     [dataset_id],
                 )
                 rows = settings.docStoreConn.get_fields(res, ["id", "doc_id", "content_with_weight"]) or {}
-            except Exception:
+            except Exception as exc:
                 logging.exception("alteration: failed to load source chunks for doc=%s", doc_id)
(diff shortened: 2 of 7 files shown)


wangq8 (commented):
Thanks for this — the approach is well-targeted and the core hash comparison is sound. I traced the hashing path and confirmed `chunk_hash_kwd` is computed from `content_with_weight` (via `_wiki_pick_chunk_text`, which falls back to `content_with_weight` since source chunk rows carry no `text` field), so `_wiki_chunk_alteration` comparing `_chunk_hash(row["content_with_weight"])` against the stored hash is apples-to-apples. A few concerns and questions before I'd approve:

**1. N+1 queries in `_wiki_chunk_alteration` (performance)**
The function issues one paginated ES search *per active document* to load source chunks. For a KB with hundreds or thousands of wiki-eligible docs, `get_wiki_alteration` — which is meant to be a quick "should I rebuild?" check — could become noticeably slow. Consider loading source chunks in a single query over all `active_doc_ids` (terms query, paginated), or otherwise bounding the work.

**2. Silent failure masks a needed rebuild**
On any exception, `_wiki_chunk_alteration` returns `{"changed": 0, "changed_doc_ids": []}`. That collapses "detection failed" into "nothing changed", which could tell the frontend no rebuild is needed when detection actually errored (e.g., ES transient failure). Please distinguish failure from "no change" — e.g. include an `error` flag or let the endpoint surface it — so a missed rebuild isn't silently hidden.

**3. Importing a private symbol**
`from rag.advanced_rag.knowlege_compile.wiki import WIKI_PAGE_COMPILE_KWD, _chunk_hash` imports a module-private name across modules. If `wiki.py` ever renames/refactors `_chunk_hash`, this breaks without an obvious signal. Consider promoting it to a public helper (e.g. `chunk_hash`) or moving it to a shared util module.

**4. Missing tests**
No tests are included for the new `changed`/`changed_doc_ids` detection (added/removed/edited chunk, disabled-doc cases) or the disabled-doc exclusions in `_wiki_compute_map_input_hash` / `_wiki_load_all_map_extracts`. Given this is subtle cache-invalidation logic, at least a focused unit test for `_wiki_chunk_alteration` and the disabled filter would meaningfully reduce regression risk.

**5. Duplicated disabled-doc helpers**
`_wiki_disabled_doc_ids` (wiki.py), `_disabled_dataset_doc_ids` (dataset_api_service.py), and the inline `DocumentService.get_disabled_doc_ids_by_kb_id` calls in `wiki_incremental.py` / `dataset_wiki_generator.py` all reimplement the same thing. Worth consolidating into one shared helper.

**6. Split responsibility for the disabled filter (non-blocking)**
The filter in `wiki_compile_incremental` only runs on the `map_results is None` (load-from-ES) path; the fresh-MAP path is safe because `_wiki_eligible_docs` already skips `status != "1"`. That's correct today, but the responsibility is split across two modules. A short comment or a shared helper would reduce the chance of future drift.

None of these are correctness blockers for the main fix — happy to approve once 1–2 are addressed (performance batching and/or the error-handling distinction would be my top asks).

PR #18262, opened 2026-08-14
fix(web): enlarge card '...' edit button hit area

euvre:
## Problem

The '...' (more/edit) button on knowledge-base cards is very hard to click. Although `MoreButton` uses the `icon-xs` button variant (`size-6` = 24px), its own `size-3.5` class wins via tailwind-merge, shrinking the actual button box to **14×14px** — below the WCAG 24×24px minimum pointer target. The icon fills the whole button, so any click slightly off-center misses and hits the card instead.

## Fix

Keep the 14px visual size, but expand the interactive area to **30×30px** with an absolutely positioned `::after` pseudo-element (`after:-inset-2`), the same pattern already used in `ui/sidebar.tsx`. The pseudo-element belongs to the button, so clicks in the expanded ring still trigger the Radix `DropdownMenuTrigger`; card padding (`px-2.5 py-4`) fully contains the 8px expansion.

Since `MoreButton` is shared, this fixes all card types: datasets / agents / chats / searches / memories / skills / home.

## Verification

- Live-checked on `/datasets`: `getBoundingClientRect` = 14×14 while `::after` computed style has `inset: -8px`; `elementFromPoint` 7px outside the button box resolves to the button.
- Clicking in the expanded area opens the dropdown menu (Rename/Delete); `pointerdown.defaultPrevented = true` confirms no accidental card navigation.
- oxlint clean on the touched file; pre-commit hooks pass.

### DIFF
--- a/web/src/components/more-button.tsx
+++ b/web/src/components/more-button.tsx
@@ -28,6 +28,9 @@ export const MoreButton = React.forwardRef<HTMLButtonElement, ButtonProps>(
         size={size || 'icon-xs'}
         className={cn(
           'opacity-0 size-3.5 transition-all bg-transparent group-hover:bg-transparent',
+          // Keep the 14px visual size but increase the hit area with an
+          // absolutely positioned pseudo-element (14px + 8px on each side).
+          'relative after:absolute after:-inset-2 after:content-[""]',
           'group-focus-within:opacity-100 group-hover:opacity-100 aria-expanded:opacity-100',
           className,
         )}


wangq8 (commented):
Nice fix. The `::after` pseudo-element pattern matches what's already used in `ui/sidebar.tsx` and correctly keeps the visual size at 14px while expanding the interactive target to 30×30px, meeting the WCAG minimum pointer-target size.

A few points I checked:
- The added `relative` is necessary since the `Button` base has no positioning class by default (only when `dot` is set), so the `after:absolute` now anchors to the button.
- `after:content-[""]` is required to render the pseudo-element; `after:-inset-2` yields exactly 14px + 8px×2 = 30px.
- tailwind-merge keeps `size-3.5` winning over the variant's `size-6`, and `relative` introduces no conflict.
- Clicks on the expanded ring correctly resolve to the button (the pseudo-element belongs to the button), and Radix `DropdownMenuTrigger` still prevents the card's `onClick` from firing.

One minor note: `-inset-2` extends 8px in all directions. Card padding (HomeCard `px-2.5 py-4`) fully contains it, but in tighter list layouts (e.g. the next-chats session list where `MoreButton` sits `gap-1` from the flex-1 name button) the expanded area overlaps the adjacent button's padding by a few pixels, slightly shifting which control receives a click in that narrow zone. Acceptable, but worth a quick visual check there.

No regressions or security concerns. Good to merge.

PR #18264, opened 2026-08-14
Fix DateTimeTzField crashing on driver-native datetimes and unparseable values

euvre:

The sync task scheduler (`rag/svr/sync_data_source.py`) periodically logs `DB is not ready yet: list index out of range` and stops dispatching due tasks until restart.

**Root cause:** `sync_logs.poll_range_start`/`poll_range_end` are declared as VARCHAR (`DateTimeTzField` stores ISO strings), but deployments upgraded from older schemas may still hold them in native DATETIME columns, so the MySQL driver returns `datetime` objects instead of strings. `DateTimeTzField.python_value` passed the value straight into `datetime.fromisoformat`, raising `TypeError`; zero-date values (`0000-00-00 00:00:00`) similarly raised `ValueError`.

Peewee's `ResultIterator.iterate()` counts the row before converting it, so one throwing row desynchronizes the row count from the row cache, and **every subsequent iteration over the table** raises `IndexError: list index out of range` — which is caught in the scheduler loop and surfaces as the misleading `DB is not ready yet` warning.

**Fix:** harden `python_value` to accept whatever the driver returns:
- `datetime` objects pass through (attaching UTC when naive)
- strings are parsed with `fromisoformat` (naive results assumed UTC, as before)
- unparseable values (zero dates, garbage, empty strings) fall back to `None` with a warning instead of raising — matching the Go syncer's `FlexibleTime` zero-time fallback
- also drops the inline `pytz` import in favor of the already-imported `datetime.timezone`

### Type of change

### Testing

- Added `test/unit_test/api/db/test_datetime_tz_field.py` (12 tests): driver-native naive/aware datetimes, ISO strings with/without tz, zero-date/garbage/empty strings falling back to `None`, and `db_value`/`python_value` round trip. All pass via `uv run --group test pytest`.
- Verified end-to-end against a live MySQL instance seeded with DATETIME-typed `poll_range_*` columns and zero dates: the scheduler query no longer raises, and due tasks dispatch normally.

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -1316,15 +1316,24 @@ class DateTimeTzField(CharField):
                 return value.replace(tzinfo=timezone.utc).isoformat()
         return value
 
-    def python_value(self, value: str | None) -> datetime | None:
-        if value is not None:
+    def python_value(self, value: str | datetime | None) -> datetime | None:
+        if value is None:
+            return None
+        if isinstance(value, datetime):
+            # The column is declared VARCHAR, but deployments upgraded from
+            # older schemas may hold it as native DATETIME, in which case the
+            # driver returns datetime objects instead of ISO strings.
+            return value if value.tzinfo is not None else value.replace(tzinfo=timezone.utc)
+        try:
             dt = datetime.fromisoformat(value)
-            if dt.tzinfo is None:
-                import pytz
-
-                return dt.replace(tzinfo=pytz.UTC)
-            return dt
-        return value
+        except ValueError:
+            # Zero dates (0000-00-00) and other unparseable values must not
+            # raise here: peewee counts the row before converting it, so one
+            # bad value wedges every later iteration over the table with
+            # IndexError: list index out of range.
+            logging.warning("DateTimeTzField: unparseable value %r, falling back to None", value)
+            return None
+        return dt if dt.tzinfo is not None else dt.replace(tzinfo=timezone.utc)
 
 
 class SyncLogs(DataBaseModel):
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the detailed analysis — the root cause (peewee counting the row before converting it, so one throwing `python_value` desyncs the iterator and surfaces as `DB is not ready yet: list index out of range`) is correct and well documented.

Review notes:

- The fix correctly handles the three cases: `None`, driver-native `datetime` (naive → UTC, aware kept), and ISO strings. Dropping the inline `pytz` import in favor of the already-imported `datetime.timezone` is clean and behaviorally equivalent.
- Falling back to `None` (with a warning) for unparseable values matches the Go syncer's `FlexibleTime` zero-time fallback. Downstream in `sync_data_source.py`, `poll_range_start` is always truthiness-checked before `.timestamp()`, so `None` is safe — a zero-date row now triggers a full sync rather than wedging the scheduler, which is the correct semantic.
- Test coverage is solid (12 cases across naive/aware datetimes, tz/no-tz ISO strings, zero-date/garbage/empty fallbacks, and db↔python round trips).

Minor (non-blocking):
- `except ValueError` won't catch `TypeError`. If a driver ever returns a value that is neither `str` nor `datetime` (e.g. a `datetime.date` from a DATE-typed column), `datetime.fromisoformat` raises `TypeError` and the same iterator-desync bug could resurface. Consider catching `(ValueError, TypeError)` or guarding with `isinstance(value, str)` before parsing.

LGTM otherwise.

PR #18267, opened 2026-08-14
Fix: Editing a document slice requires showing the wiki's update button

cike8899:
### Summary

Fix: Editing a document slice requires showing the wiki's update button


wangq8 (commented):
Thanks for this fix. The frontend correctly consumes the `changed` / `changed_doc_ids` fields that the alteration API now returns for `kind=wiki` (chunk-hash based drift detection), and wiring `changed` into `hasChanges`, the badge, the tooltip, and invalidating `ArtifactAlterationKeys.detail` on run end all look correct.

One suggestion: in `web/src/interfaces/database/dataset.ts`, `changed` and `changed_doc_ids` are added as required fields, but the backend only returns them for `kind === "wiki"` (structure kinds like `graph`/`mindmap`/`timeline`/`tree` do not include them). Consider marking them optional (`changed?: number; changed_doc_ids?: string[]`) so the type accurately reflects the API contract for all kinds.

Otherwise LGTM.

PR #18270, opened 2026-08-14
Fix: If the model is removed from the supplier list, an error should occur when accessing the search and memory pages.

cike8899:
### Summary

Fix: If the model is removed from the supplier list, an error should occur when accessing the search and memory pages.


wangq8 (commented):
LGTM. Verified the implementation against the codebase:

- `buildValidModelIds` already exists in `@/utils/llm-util`, and `useFetchAllAddedModels(modelType?, ownerTenantId?)` matches the `(undefined, tenantId)` calls.
- `ModelTypeMap` provides `llm_id`, `embd_id`, `rerank_id`; the existence check covers both the `model_id` and legacy composite (`name@instance@provider`) forms.
- The dynamic `zodResolver(formSchema)` is effective: react-hook-form updates the control's resolver on each render, so `trigger()` from `useRevalidatePersistedModels` runs against the existence check once models load.
- Query dedup: the two `useModelValidIds` calls share the same `useFetchAllAddedModels(undefined, …)` query key, so only one extra request.

Non-blocking suggestions:
- Consider a unit test for the new existence-check logic (e.g. `buildValidModelIds` and the stale-model error path).
- Memory passes `tenant_id` to scope the model fetch while the search page fetches without a tenant — please confirm this asymmetry is intended.

No regressions found. Approving.

wangq8:
@cike8899 resolve conflict.

PR #18276, opened 2026-08-14
fix(web): add tooltip to web search toggle in chat input box

euvre:
### Summary

**Reported issue:** in a multi-model ("Multiple models") chat, Firefox showed one more button in the input box than Chrome, with no indication of what the extra button does.

**Investigation (reproduced locally with real Chrome + Firefox against the same backend):**

- The extra button is the **web-search (internet) toggle** — the globe icon, `data-testid="chat-detail-internet-toggle"`.
- Its visibility is gated purely by the assistant's `prompt_config.tavily_api_key` (`web/src/pages/next-chats/chat/use-show-internet.ts`), **not** by the browser.
- Parity matrix of input-box controls, verified in both browsers:

  | Assistant config | Chrome | Firefox |
  |---|---|---|
  | with Tavily key | 5 controls (globe shown) | 5 controls (globe shown) |
  | without Tavily key | 4 controls | 4 controls |

  → Chrome and Firefox render identically; the reporter was comparing two different assistants in the two browser windows.

**Actual defect:** the globe button had no tooltip, label, or accessible name, so users could not tell what it is or why it appears for some assistants — which caused the confusion.

**Fix:**

- Wrap the web-search toggle in a `Tooltip` showing "Web search" (`chat.webSearch`, en + zh), following the same pattern as the neighbouring thinking toggle.
- Add an `sr-only` text inside the button so it has an accessible name.

**Verification:**

- Chrome (dev-tools MCP) and Firefox (Playwright) both show the toggle with the "Web search" tooltip on hover; the parity matrix above was re-checked after the change.
- `oxlint` on touched files: no new findings.
- `tsc --noEmit`: error count identical to base (pre-existing errors only).
- `npx jest src/pages/next-chats/chat/use-show-internet.test.ts`: 8/8 pass.

### DIFF
--- a/web/src/components/message-input/next.tsx
+++ b/web/src/components/message-input/next.tsx
@@ -30,6 +30,11 @@ import {
 } from '@/components/file-upload';
 import { Button } from '@/components/ui/button';
 import { Textarea } from '@/components/ui/textarea';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
 import { CircleStop, Globe, Paperclip, Send, Upload, X } from 'lucide-react';
 import * as React from 'react';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
LGTM. Reviewed the diff:

- The web-search toggle is wrapped with the existing `@/components/ui/tooltip`, matching the neighbouring thinking toggle.
- The `sr-only` span gives the button an accessible name — good a11y fix.
- Locale keys are added under the `chat` object in both `en.ts` and `zh.ts`; `t('chat.webSearch')` resolves correctly.

No regressions or style issues found. Approving.

dcc123456 (approved):

PR #18277, opened 2026-08-14
Fix: return error when no valid speech text in intranscription response

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/apps/restful_apis/chat_api.py
+++ b/api/apps/restful_apis/chat_api.py
@@ -1151,6 +1151,8 @@ async def transcription():
             os.remove(temp_audio_path)
         except Exception as e:
             logging.error(f"Failed to remove temp audio file: {str(e)}")
+        if "**ERROR**" in text:
+            return get_data_error_result(message=text)
         return get_json_result(data={"text": text})
 
     async def event_stream():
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for the fix — this correctly turns a malformed/empty DashScope ASR response into an actual error result (for the non-stream path) instead of silently returning a 200 with `"**ERROR**: ..."` as the transcription text. The defensive `.get()`-based parsing is also a nice improvement over the previous direct indexing.

A few points to consider before merge:

1. **Fragile sentinel check.** `if "**ERROR**" in text` relies on a magic substring. If legitimate transcription output ever contains the literal `**ERROR**` (e.g., someone dictating that string), the API would wrongly return an error. It's an edge case, but a more robust signal (a dedicated error object/exception or a structured marker) would avoid false positives. This is a pre-existing pattern (`check_available` uses the same marker), so I won't block on it — just flagging the fragility.

2. **Missing test coverage for the new error branches.** The test change only updates the mock's `message` structure to match the real response shape. There's no test exercising the new failure paths (empty/missing `choices`, missing `content`, missing `text`) in either `_transcribe_qwen_audio` or `stream_transcription`. Given this is the core of the fix, please add unit tests for these cases.

3. **Error message leakage (minor).** `get_data_error_result(message=text)` returns the raw exception string (e.g. `**ERROR**: 'output'`) directly to the client, which may expose internal details. Consider returning a generic user-facing message and logging the full error server-side.

4. **Consistency (non-blocking).** The streaming path still yields `{"event": "error", "text": str(e)}` without the `**ERROR**` prefix, so the new API-level `"**ERROR**"` check doesn't apply there. That's fine since the client handles the `error` event, just noting the inconsistency.

Overall the change is safe and a genuine improvement; points 2 and 3 are the ones I'd like addressed.

PR #18278, opened 2026-08-14
fix(ingestion): stop orphaned task messages under backpressure and reconcile stuck tasks

jay77721:
### Summary

When batch-parsing a batch of files, only the first few files completed; the rest stayed stuck in "parsing" forever.

**Root cause (verified in logs + DB + live NATS):** the ingestor's consume loop fetched a fixed `4` messages per iteration while its worker channel held only `maxConcurrency*2` (2 with the default `max_concurrent_workers: 1`). Overflowing messages were **Nacked** (`msg.Nak()`, ~1s redelivery). The NATS consumer has `MaxDeliver: 16`, and repeated Nacks exhaust it: the message stops being delivered and is orphaned until the consumer is recreated. Because `processMessage` calls `StartRunning` **before** the Nack, the task stays `RUNNING` with the document in "parsing" forever. Log evidence: 1293 `No available slot` nacks; one task (`037b2382`) nacked 16× in 16s.

Empirical check (throwaway workqueue stream): MaxDeliver exhaustion does **not** hard-delete the message — it stays in the stream and resurrects only when the consumer is deleted/recreated (i.e. on ingestor restart). Without a restart the task is wedged indefinitely.

### Changes

**Layer 1 — the consume side no longer drops messages under backpressure:**
- `consumeLoop` fetches only as many messages as the worker channel can currently absorb (`cap(taskChan)-len(taskChan)`), so it can never overflow.
- New `Ingestor.queueTaskCtx`: when the channel is full it **blocks** (with a `startHeartbeat` keeping the broker AckWait fresh) instead of Nacking. Only graceful shutdown aborts the wait and Nacks for redelivery after restart. Applied to both the ingestion and memory-task paths. All real error Nacks (DB blip, pipeline failure) are unchanged. `startHeartbeat` is nil-safe for memory-task contexts (no `IngestionTask` row) and doubles as the durable liveness touch.

**Layer 2 — a reconciliation sweeper makes lost tasks self-heal without a restart** (new `reconcileOnce` loop, 60s):
- Re-enqueues tasks whose wake-up message was orphaned: stale `CREATED` (>5min), and `RUNNING` with no component-progress row (>5min) — both safe to re-run (pipeline never started). Requeue advances the task's staleness watermark (`Touch`) as a durable backoff.
- Fails tasks whose worker went silent mid-run (`RUNNING` with progress rows stale >10min), mirroring `FAILED` to the document — not re-run, to avoid duplicate chunk writes. Liveness is durable: the executing worker `Touch`es the task row on every heartbeat tick, so a long single component (OCR/LLM call) that writes no progress rows for minutes is never mis-killed — on any replica. The document is mirrored only after confirming the actual RUNNING→FAILED transition.
- Resets documents left `RUNNING` with no ingestion task row and stale >5min to unstart (the age bound protects the enqueue window where `run` flips before the task row lands).
- New DAO queries: `ListStaleCreated`, `ListRunningWithoutProgress`, `ListHungRunning` (mutually exclusive: `MAX(create_date)` is NULL for log-less tasks vs `NOT EXISTS(log)`), `Touch`; `DocumentDAO.ListRunningWithoutIngestionTask` (age-bounded, oldest-first).

**Layer 3 — throughput:** default `max_concurrent_workers` raised 1 → 2 (config + docker template; local conf → 4). At any concurrency the correctness fix holds; this just removes the serialization bottleneck.

### Tests
- `TestProcessMessage_ChannelFullBlocks` replaces `TestProcessMessage_ChannelFullNacks`: asserts channel-full now blocks (0 Ack/0 Nack), resumes when a slot frees, and keeps the claim with the enqueued task.
- New `task_reconcile_test.go`: requeue stale CREATED, requeue RUNNING-without-progress, fail hung RUNNING (+doc mirror), skip healthy tasks, skip in-process claimed task, don't overwrite a completed document.
- Verified: `bash build.sh --test ./internal/ingestion/... ./internal/dao/ ./internal/service/ ./internal/server/config/` all green.

### Notes
- Reconciliation and the claim dedup assume a single ingestor instance (NATS workqueue streams allow one consumer); multi-instance would need a leader lock (documented in code).
- Deployment constraint and thresholds (5min/5min/10min) are constants in `ingestion_service.go`, tunable.


jay77721:
Addressed all CodeRabbit review comments in `85c5c8b49`:

**Actionable**
1. 🔴 `startHeartbeat` nil `IngestionTask` — now derives the task ID from `Handle.GetMessage().TaskID` for memory-task contexts (and added `TestProcessMessage_MemoryTaskChannelFullBlocks` as a regression test).
2. 🟠 Document reset staleness bound — `ListRunningWithoutIngestionTask` now requires `document.update_date < now-5min` (new `orphanedDocumentResetAfter` const); added tests for stale-reset and fresh-kept.
3. 🟠 Durable liveness instead of in-process claims — the worker heartbeat now `Touch`es the task row on every tick and `failHungTask` no longer uses `isClaimed`; the hung-candidate test was rewritten to assert the durable watermark (`TestReconcile_SkipsHeartbeatedTask`).
4. 🟡 Compiler-pool docs — `MaxConcurrentWorkers` doc now describes only ingestion-task concurrency; `CompilerPoolSize` documented as passed independently by `cmd/ragflow_server.go` with zero = vCPU default.

**Nitpicks**
- `ListRunningWithoutIngestionTask` orders by `document.update_date ASC` before `Limit`.
- Removed the redundant `EXISTS(log)` from `ListHungRunning` (kept the mutual-exclusion comment).
- Added composite indexes `ingestion_task(status, update_date)` and `ingestion_task_log(task_id, create_date)` via `migrateIngestionTaskReconcileIndexes`.
- `reconcileLoop` falls back to `defaultReconcileInterval` on a non-positive interval instead of panicking.
- `TestProcessMessage_ChannelFullBlocks` now asserts the blocked enqueue heartbeats (`inProgress > 0`).
- All reconciliation test setup writes now assert error + `RowsAffected` via `mustAffectRows`.

Validation: `bash build.sh --test ./internal/ingestion/service/... ./internal/dao/ ./internal/server/config/` and `./internal/service/` all green; `go vet` clean.

Haruko386 (approved):
LGTM

jay77721:
Second review round resolved in `1f1fa937e`:

1. 🟠 **Orphan-document reset race (select → update)** — the reset is now conditional at write time via the new `DocumentDAO.ResetRunningWithoutIngestionTask`: it only updates when the document is still `run=RUNNING`, `update_date` is still before the staleness bound, and there is still no `ingestion_task` row (`NOT EXISTS`). A candidate listed by the sweeper can no longer overwrite a document that was enqueued or completed in between. Added `TestResetRunningWithoutIngestionTask_Conditional` covering stale-reset, enqueue-won, fresh-window, and terminal-state cases.
2. 🟡 **SA1012 nil context** — both `NewTaskContextForScheduling` filler calls in the channel-full tests now pass `context.Background()`.

Validation: `bash build.sh --test ./internal/ingestion/service/... ./internal/dao/` green, `go vet` + `go build ./...` clean.

Haruko386 (approved):
LGTM

Haruko386 (commented):
not that good 2 me.
I feel this is not the best solution

PR #18279, opened 2026-08-14
fix(ingestion): mirror document run status on parse enqueue

euvre:
### Summary

**Bug**: In the knowledge base document list, clicking Parse shows a success toast, but the parse button state does not change. Clicking it again fails with `failed to enqueue document ... already exists, status: RUNNING`.

**Root cause**: On a successful parse enqueue, the Go backend only flips the document's `run` status to RUNNING when the worker actually picks the task up (`IngestionTaskService.StartRunning`). The frontend invalidates and refetches the document list right after the success response, but the document still carries the stale `run` value (UNSTART/DONE), so the row keeps rendering the parse button. A second click reaches `CreateAndEnqueue`, which correctly rejects the duplicate with the "already exists" error. The stale window grows with queue backlog, since the task may sit in CREATED long before pickup.

**Fix**: Mirror `run=RUNNING` (and `progress=0`) to the document as soon as the queue accepts the task in `IngestionTaskService.CreateForDocuments` — the single funnel used by all parse entry points (UI ingest, SDK parse endpoint, connector sync re-parse, ChunkService.Parse). The mirror is best-effort (warn-only on DB error) so an accepted enqueue never turns into an API error. `StartRunning` still re-asserts the status and resets progress counters at actual worker pickup. This is symmetric with the cancel path, which already mirrors `run=CANCEL` synchronously.

**Testing**:
- New unit test `TestIngestionTaskServiceCreateForDocumentsMirrorsRunningToDocument` asserts the document is RUNNING with progress reset right after `CreateForDocuments`; full `internal/service/...`, `internal/handler`, `internal/ingestion/service` suites pass.
- Live verification: reparse of a 132-chunk PDF flips the row to the running state immediately after the confirm dialog (sampled at 400ms intervals) and transitions to success on worker completion; no "already exists" error on screen.

### DIFF
--- a/internal/dao/document.go
+++ b/internal/dao/document.go
@@ -71,6 +71,18 @@ func (dao *DocumentDAO) UpdateByID(ctx context.Context, db *gorm.DB, id string, 
 	return db.WithContext(ctx).Model(&entity.Document{}).Where("id = ?", id).Updates(updates).Error
 }
 
+// UpdateByIDIfNotCancelled updates document by ID with the given fields, but
+// only when the document's run status is not CANCEL. Mirrors the WHERE clause
+// of Python's DocumentService.begin2parse: a concurrent cancel writes
+// run=CANCEL as the last step of the cancel flow, so the conditional keeps an
+// enqueue-time status mirror from overwriting a cancel that already committed.
+func (dao *DocumentDAO) UpdateByIDIfNotCancelled(ctx context.Context, db *gorm.DB, id string, updates map[string]interface{}) error {
+	return db.WithContext(ctx).Model(&entity.Document{}).
+		Where("id = ?", id).
+		Where("run IS NULL OR run <> ?", string(entity.TaskStatusCancel)).
+		Updates(updates).Error
+}
+
 // IncrementCounts atomically increments chunk_num, token_num, and process_duration for a document
 func (dao *DocumentDAO) IncrementCounts(ctx context.Context, db *gorm.DB, id string, kbID string, chunkNum int64, tokenNum int64, duration float64) error {
 	return db.WithContext(ctx).Model(&entity.Document{}).
--- a/internal/service/document/document_ingest.go
+++ b/internal/service/document/document_ingest.go
@@ -83,9 +83,8 @@ func (s *DocumentService) Ingest(ctx context.Context, userID string, req *Ingest
 		doc := vd.doc
 		kb := vd.kb
 
-		// Start parsing: delegates to the shared start-parse flow. The
-		// document run status is set by service.IngestionTaskService.StartRunning
-		// when the task transitions from CREATED, not here.
+		// Start parsing: delegates to the shared start-parse flow, which
+		// mirrors the document run status on enqueue acceptance.
 		if run == string(entity.TaskStatusRunning) {
 			if err = s.StartParseDocuments(ctx, doc, kb, userID, StartParseOptions{
 				ApplyKB:         req.ApplyKB,
--- a/internal/service/document/document_parse.go
+++ b/internal/service/document/document_parse.go
@@ -21,10 +21,11 @@ import (
 // StartParseDocuments starts parsing a document via the DSL ingestion
 // pipeline. It optionally clears prior results (RerunWithDelete), applies
 // KB config (ApplyKB), validates storage, and enqueues an ingestion task.
-// The document run status is NOT set here; service.IngestionTaskService.StartRunning
-// sets it to RUNNING when the worker picks up the task and transitions it from
-// CREATED. Extracted from Ingest so other entry points (e.g. ChunkService.Parse)
-// can reuse the same start-parse flow.
+// The document run status is mirrored to RUNNING on enqueue acceptance in
+// service.IngestionTaskService.CreateForDocuments, then re-asserted with
+// progress counter resets by service.IngestionTaskService.StartRunning when
+// the worker picks up the task. Extracted from Ingest so other entry points
+// (e.g. ChunkService.Parse) can reuse the same start-parse flow.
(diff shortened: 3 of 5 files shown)


wangq8 (commented):
Good catch and clean fix — mirroring `run=RUNNING` at enqueue acceptance is the right layer, and keeping it best-effort (warn-only) so an accepted enqueue never turns into an API error is correct. The new test asserts both `run` and the progress reset, which covers the regression you described.

One minor observation (non-blocking): the enqueue mirror resets `run` and `progress` but not `progress_msg` (nor `process_begin_at`/`chunk_num`/`token_num`), while `StartRunning` resets all of them. During the queued-but-not-yet-picked-up window, a document can show `run=RUNNING, progress=0` alongside a stale `progress_msg` from the previous run (e.g. "Done"/"Failed"), which may render inconsistently in the UI. Consider also clearing `progress_msg` in the enqueue mirror for consistency with `StartRunning`. Cosmetic only — the duplicate-enqueue bug itself is fixed.

Otherwise LGTM.

xugangqiang:
Leaving a few questions/concerns from review of this change. Problems/questions only, no fixes suggested.

**1. Concurrency with cancel**
`CreateForDocuments` now unconditionally writes `run=RUNNING, progress=0` to the document after enqueue (`internal/service/ingestion_task_service.go:117-126`). Separately, `CancelDocParse` writes `run=CANCEL` after `RequestStop` (`internal/service/document/document_parse.go:348`). These two `UpdateByID` calls are not serialized. If a cancel's `CANCEL` write commits before this mirror's `RUNNING` write, the document ends up showing `RUNNING` while the task is already `STOPPED`. Since `StartRunning`'s `STOPPING` branch (`ingestion_task_service.go:231-232`) does not rewrite the document, the doc would then be stuck on `RUNNING` indefinitely. Is this interleaving handled somewhere I missed?

**2. Asymmetry with the cancel path**
The PR says the mirror is "symmetric with the cancel path", but `CancelDocParse` returns an error on DB failure, whereas this mirror is warn-only. Is the warn-only behavior for the enqueue mirror intentional, given the cancel path is not?

**3. Partial progress/display reset**
The enqueue mirror resets only `run` and `progress`, but not `chunk_num`, `token_num`, `process_begin_at`, or `progress_msg` (which `StartRunning` sets at `ingestion_task_service.go:220-227`). Between enqueue acceptance and worker pickup, the UI would show `RUNNING` together with stale chunk/token counts and a stale `progress_msg`. Was this brief inconsistent display state considered acceptable?

**4. Test coverage gaps**
- The new test verifies `RUNNING` + `progress` reset on the happy path, but there is no test for the cancel/enqueue interleaving from (1), and no test asserting the mirror is truly best-effort (a DB error must not surface as an API error). Are these scenarios covered elsewhere?
- Does the suite actually exercise the real DB write in `CreateForDocuments` via `dao.DB`, or is that path mocked?

**5. Scope across all entry points**
`CreateForDocuments` is the funnel for UI ingest, SDK parse, connector sync re-parse, and `ChunkService.Parse`. This PR resets `progress` to 0 immediately on enqueue for all of them. Is an immediate `progress=0` always the desired behavior for connector sync re-parse and `ChunkService.Parse`, or could it clobber a still-meaningful progress value shown to the user?

euvre:
Thanks @xugangqiang and @wangq8 for the careful review. All three points are addressed in `40b670b7a`.

**1. Concurrency with cancel (@xugangqiang, same finding as CodeRabbit's inline comment) — valid, fixed.**

Agreed: the mirror was an unconditional `UpdateByID`, so a cancel whose `run=CANCEL` committed before the mirror could be overwritten back to `RUNNING`, and the document would stick since `StartRunning`'s STOPPING branch no longer rewrites it. The fix mirrors what Python's `DocumentService.begin2parse` (`api/db/services/document_service.py:1071`) already does for this same enqueue-time update: it is now conditional — `WHERE id = ? AND (run IS NULL OR run <> CANCEL)`, via a new `DocumentDAO.UpdateByIDIfNotCancelled`. Because the cancel flow writes `run=CANCEL` as its last step (after `RequestStop`), both commit orders now converge: if the mirror lands first, cancel's unconditional CANCEL overwrites it; if cancel lands first, the mirror matches zero rows and the document stays CANCEL.

Covered by tests for both orders: `TestIngestionTaskServiceCreateForDocumentsSkipsMirrorAfterCancel` (enqueue is accepted, task published, but the committed cancel is preserved) and the existing `TestIngestionTaskServiceCreateForDocumentsMirrorsRunningToDocument` for the normal order.

**2. Warn-only asymmetry (@xugangqiang) — intentional, kept as is.**

The asymmetry comes from ordering. In the Go flow the task row is durable and the message is already published when the mirror runs, so a mirror DB failure cannot be surfaced as an API error — the parse did start — and `StartRunning` re-asserts the full document state at worker pickup, so a lost mirror self-heals. Python can surface a `begin2parse` failure only because it runs before the queue push; the Go mirror runs after it, which is precisely why it is warn-only. In `CancelDocParse` the document write is the final, user-visible state change of the cancel action itself (and there may be no task row at all), so a failure there genuinely means "the cancel did not take effect" and is returned.

**3. Stale `progress_msg` / partial reset (@wangq8, @xugangqiang) — fixed.**

The mirror now also sets `progress_msg = "Task is queued..."` and `process_begin_at`, matching `begin2parse`, so the queued window no longer renders a stale message from the previous run next to `run=RUNNING, progress=0`. `chunk_num`/`token_num` are deliberately left alone: Python does not reset them at enqueue either, and `StartRunning` zeroes them at worker pickup, so the queued-window behavior now matches the Python implementation field-for-field. The mirror test asserts all four mirrored fields.

Verification: `bash build.sh --test` for `internal/service`, `internal/dao`, `internal/service/document`, `internal/ingestion/service` — all pass. No shared-service end-to-end verification was run on this branch; conclusions are based on unit tests plus line-by-line comparison with the Python implementation.

PR #18283, opened 2026-08-14
fix(web): clarify batch model action labels

dcc123456:
### Summary

fix(web): clarify batch model action labels

The "batch" wording was ambiguous now that these actions always apply to every model in the list rather than a user-selected subset.


wangq8 (commented):
LGTM. The relabeling is accurate — these actions always apply to every model in the list, so "Verify all models" / "验证全部模型" (and the "Add/Remove all models" equivalents) is clearer than the "batch" wording. English and Chinese strings are consistent.

PR #18284, opened 2026-08-14
fix(web): scope answer loading placeholder to the last message

dcc123456:
### Summary

fix(web): scope answer loading placeholder to the last message

While a new answer is streaming, every assistant message with empty content rendered the loading dots and the "searching" placeholder, so historical empty answers flickered as if they were regenerating. Pass `isLast` down to MessageItem and only show the placeholder for the message actually being generated.

### DIFF
--- a/web/src/components/markdown-content/index.tsx
+++ b/web/src/components/markdown-content/index.tsx
@@ -142,7 +142,7 @@ const MarkdownContent = ({
     });
 
     // let text = content;
-    if (text === '') {
+    if (text === '' && loading) {
       text = t('chat.searching');
     }
     const nextText = replaceTextByOldReg(text);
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
LGTM. The fix correctly scopes the loading placeholder to the message actually being generated, and the `loading` prop already carries the role check (`Assistant && sendLoading && isLast`), so the "searching" text stays on the last assistant message.

One tiny note (non-blocking): `isLast` is positional (`i === length - 1`) and is not role-checked at the `LoadingDots` site (`sendLoading && isLast && isEmpty(...)`). That's fine in practice because a trailing user message has non-empty content, but if the list ordering ever allows an empty non-assistant message to be last while `sendLoading` is true, the dots would render there too. Not worth changing unless you want to be defensive.

PR #18285, opened 2026-08-14
fix: track wiki chunk changes incrementally

buua436:

Wiki incremental compilation did not reliably detect chunk-level changes. Deleted or disabled chunks could reuse stale MAP results, historical extraction results were overwritten, and partially failed compilations could leave alteration status inconsistent.

This PR:

- Stores historical MAP results by chunk ID and content hash.
- Tracks the chunk snapshot from the last successful Wiki compilation.
- Handles new, changed, deleted, and disabled chunks incrementally.
- Reuses historical extraction results when content is reverted.
- Aligns Wiki alteration results with compilation eligibility.
- Prevents deletion-only updates from loading stale active MAP results.
- Keeps the previous successful snapshot when compilation is incomplete.
- Fixes page refinement for string-valued version and relation fields.

### Type of change


wangq8 (commented):
Thanks for this — the migration from single-version MAP rows to a versioned cache plus an explicit committed-state snapshot is a clean design, and the tests cover the important edges (revert-reuse, deleted-chunk retraction, `_as_int`, native-string relations).

Left a few points below; nothing blocking, mostly nits and things I'd like confirmed.

**1. Dead/misleading per-doc delta logging (minor)**
`wiki_map_from_chunks` now returns `_meta = {requested, cache_hits, extracted}`, but `_map_worker`'s aggregation in `dataset_wiki_generator.py` still reads the old keys:

```python
for key in stats["delta"]:               # {"new","changed","unchanged","deleted"}
    stats["delta"][key] += int(meta.get(key, 0) or 0)
stats["had_delta"] |= bool(meta.get("had_delta"))
```

Since `_meta` no longer carries those keys, the `wiki: MAP doc=... new= changed= unchanged= deleted= delta=` log line will always print zeros/`False` even when there is a real delta (the functional delta is now computed separately via `_wiki_compare_chunk_states`). Suggest updating this block to read the new `_meta` keys (or dropping `had_delta`/`delta` from `doc_stats` entirely).

**2. Reset no longer purges the extraction cache — please confirm intent**
`_wiki_reset_all_wiki_state` now deletes `wiki_map_state`/`wiki_map_state_meta` but no longer `wiki_map_extract`. After a reset, `_wiki_load_active_map_state` falls back to `_load(WIKI_MAP_COMPILE_KWD)`, which now scans *all* historical versions and uses first-write-wins per chunk — so the reconstructed baseline may not equal the last successful state (it can be an arbitrary stale hash). In practice the worst case is over-extraction, but note the semantics change: a user who resets to force re-extraction after a prompt/template change will silently reuse cached extracts keyed only by content hash. If reset is meant to be a full wipe including cache, the historical rows should also be cleared (or the fallback should not treat multi-version rows as the active baseline).

**3. N+1 doc-store reads at scale**
`_wiki_load_map_extracts_for_state` calls `_wiki_load_map_versions` once per doc, and `_load_map_relations` previously did a single paged scan over `wiki_map_extract`. For large KBs this turns a couple of scans into many round-trips. Worth batching by doc or restoring a single-scan path for the relations/graph loads.

**4. Nits**
- `del index_nm` inside `_wiki_chunk_alteration` is a smell — prefer dropping the parameter from the signature.
- `_wiki_has_active_map_state_marker` / `_wiki_commit_active_map_state` don't guard their ES search/delete with try/except, unlike the rest of the module. If the doc store hiccups during the "freeze" step, the whole run aborts rather than degrading gracefully. Low priority, just flagging the inconsistency.

Overall direction looks good — the state-snapshot + hash-keyed cache should fix the stale-MAP-reuse and overwrite bugs you describe.

PR #18288, opened 2026-08-14
The tool operators that allow the agent to be collapsed and hidden

cike8899:
### Summary

The tool operators that allow the agent to be collapsed and hidden


wangq8 (commented):
Thanks for this! The collapse/hide feature is a nice UX addition. The store logic and test coverage are solid — I traced `filterBottomSubtreeNodeIds` through `filterAllDownstreamNodeIds` and the AgentBottom/Tool handle traversal, and it correctly excludes the `AgentException` handle while still hiding edges whose source or target node is hidden.

A few notes (non-blocking):

1. The transformer heuristic `content.indexOf('ock(') >= 0` is a loose proxy for "contains jest.mock(" and will also match things like `clock(`, `unlock(`, `block(`. There are no false negatives for `jest.mock(` (it always contains `ock(`), so it's functionally safe, but a more precise check (e.g. `/\bmock\(/`) would reduce surprising routing through the slower babel/esbuild-jest path.

2. `toggleBottomCollapse` early-returns when `filterBottomSubtreeNodeIds(...).length === 0`, which means a collapsed entry can go stale: if a node's subtree is deleted after it's collapsed, the `collapsedBottomHandles` entry is never cleared and can't be toggled off again. It's visually harmless (nothing left to hide) but leaves stale state. Consider cleaning up orphaned collapsed entries when nodes/edges are deleted.

3. The un-hide branch (`else if (node.hidden) node.hidden = false`) assumes `hidden` is exclusively owned by this feature. It is today, but a short comment here would help future code avoid setting `hidden` for another reason and getting it silently cleared.

Nice work on the tests — the independent-handle and nested-collapse cases are exactly the tricky paths.

PR #18289, opened 2026-08-14
fix: hide wiki mode on python backend

buua436:

The pipeline Compiler form displayed the Wiki mode selector on the Python backend even though this configuration is only supported by the Go backend.

This PR uses the existing backend runtime detection to show the Wiki mode selector only when the active backend is Go.

### Type of change

### DIFF
--- a/web/src/pages/agent/form/compilation-form/index.tsx
+++ b/web/src/pages/agent/form/compilation-form/index.tsx
@@ -3,8 +3,12 @@ import { LargeModelFormField } from '@/components/large-model-form-field';
 import { SelectWithSearch } from '@/components/originui/select-with-search';
 import { RAGFlowFormItem } from '@/components/ragflow-form';
 import { Form } from '@/components/ui/form';
+import {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix! The change correctly hides the Wiki mode selector on the Python backend and reuses the existing `backend-runtime` reactive subscription pattern.

A couple of small suggestions (non-blocking):

1. For consistency with the rest of the codebase (e.g. `agent/form/extractor-form/index.tsx`, `dataset/setting/general-form.tsx`), consider using the existing `isGoBackend()` helper instead of `backendLanguage === 'go'`. It reads better and centralizes the language comparison.

2. Note that `mode` remains in the form schema (`z.enum(['entity', 'topic'])`, required) and is still submitted via `onValuesChange` on the Python backend. Since `initialCompilationValues` always provides a default (`'entity'`), validation still passes, but it's worth being aware that the field is hidden — not removed — from the payload.

LGTM otherwise. ✅

PR #18290, opened 2026-08-14
Fix: Prevent the tool operator from being dragged and connected by other operators.

cike8899:
### Summary

Fix: Prevent the tool operator from being dragged and connected by other operators.


wangq8 (commented):
LGTM. This correctly follows the established pattern in the canvas: other target handles (chunker-node, parser-node, retrieval-node, iteration-node, agent-node, etc.) already set `isConnectableEnd={false}` to prevent arbitrary incoming connections. The legitimate agent→tool edge is created programmatically through the store (the agent's `Tool` source handle uses `isConnectable={false}`), so this change blocks users from dragging other operators into the tool node without breaking the existing connection. Minimal, low-risk change — no regressions expected.

PR #18294, opened 2026-08-14
fix(search): show honest empty state when mind map has no content

euvre:
### Summary

**Bug**: In search with mind map enabled, when a query legitimately yields nothing to map (empty retrieval — e.g. the knowledge base simply has no relevant documents — or an empty/unparsable LLM answer), the mind map sheet rendered a single ghost "root" node instead of telling the user there is no mind map.

**Root cause**:
- Backend: `parseMindMapMarkdown`/runMindMap returns `{id: "root", children: []}` for empty input — an honest empty tree — but only checked `fullText == ""`, so a whitespace-only LLM answer slipped through to the parser.
- Frontend: `mindmap-sheet.tsx` mounted the G6 `IndentedTree` whenever not loading, and `IndentedTree`'s `assignIds` renders even `{}` / root-only trees as one visible "root" node.

**Fix** (no synthetic content, honest empty state):
- `internal/handler/mindmap.go`: treat a whitespace-only LLM answer as empty (`strings.TrimSpace`) and return the honest empty tree.
- `web/src/pages/next-search/mindmap-sheet.tsx`: when the tree has no children (covers `{}`, root-only, and single-node ghost forms), render the existing `knowledgeDetails.noStructureMindmap` empty state (zh 暂无 Mindmap / en "No mind map yet") instead of mounting G6.
- `internal/handler/mindmap_test.go` (new): contract tests — prose-only / think-only / blank answers all parse to a root-only empty tree; well-formed markdown still builds a real tree.

**Relation to #16669**: this supersedes the fallback-tree approach rejected there — instead of fabricating a tree from sections, the UI now admits there is nothing to show.

**Testing**:
- `bash build.sh --test ./internal/handler/` — ok; oxlint clean; tsc clean for touched files.
- Live verification: searched "水浒传108将" (a query with zero matching documents in the KB) with mind map enabled — the sheet now shows "No mind map yet" instead of the ghost root node.

**Out of scope (real but separate issues, not addressed here)**: Go ingestor not writing `available_int` for the Elasticsearch backend (no mapping default), and model resolution failing on bare/`@2@`-format model IDs.

### DIFF
--- a/internal/handler/mindmap.go
+++ b/internal/handler/mindmap.go
@@ -86,7 +86,7 @@ func runMindMap(ctx context.Context, config mindMapRunConfig) (mindMapNode, erro
 		sb.WriteString(delta)
 	}
 	fullText := sb.String()
-	if fullText == "" {
+	if strings.TrimSpace(fullText) == "" {
 		return mindMapNode{ID: "root", Children: []mindMapNode{}}, nil
 	}
 	return parseMindMapMarkdown(fullText), nil
--- a/internal/handler/mindmap_test.go
+++ b/internal/handler/mindmap_test.go
@@ -1,0 +1,72 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
LGTM, and thanks for adding contract tests. `strings.TrimSpace` is correct (`strings` is already imported), and the frontend empty-tree check (`!data || !Array.isArray(data.children) || data.children.length === 0`) correctly covers the `{}`, root-only, and single-ghost-node forms while leaving real trees to G6. Reusing `knowledgeDetails.noStructureMindmap` for the empty state is the right, honest behavior (superseding the fabricated-tree approach in #16669).

The new tests cover the important contracts (prose/think-only/blank \u2192 empty tree; headings + nested lists \u2192 real tree; multiple top-level headings \u2192 root with children).

Minor/non-blocking: `parseMindMapMarkdown` is tested directly, but `runMindMap`'s whitespace-only `fullText` path isn't exercised end-to-end (would need a mock LLM stream). Not required for this fix.

PR #18298, opened 2026-08-14
Fix:The dialog will display the referenced documentation even if the API does not return any documentation data.

cike8899:
### Summary

Fix:The dialog will display the referenced documentation even if the API does not return any documentation data.

### DIFF
--- a/web/src/pages/next-chats/chat/index.tsx
+++ b/web/src/pages/next-chats/chat/index.tsx
@@ -8,10 +8,9 @@ import {
 import { IClientConversation } from '@/interfaces/database/chat';
 import { RootLayoutContainer } from '@/layouts/root-layout';
 import { cn } from '@/lib/utils';
-import { useMount } from 'ahooks';
 import { isEmpty } from 'lodash';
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix. The root cause is clear: the old code only cleared `currentConversation` on the new-session path and only fetched on mount, so switching to an existing conversation could briefly render the previous conversation's references while the new fetch was in flight.

Making the URL the single source of truth and clearing first is the right shape. A few notes:

- Clearing `currentConversation` on every `conversationId`/`isNew` change means the chat box briefly renders empty before the fetch resolves (a flash on every conversation switch). This is the intended tradeoff to avoid stale-data leak — just flagging the UX side effect in case a skeleton/loading state would be preferable.
- `fetchSessionManually(...).then(...)` has no `.catch(...)`. `mutateAsync` rejects on error, so a failed fetch would surface as an unhandled promise rejection and the conversation would stay empty. Consider handling the error path.
- Non-blocking: this is a behavioral change for conversation switching (fetch is now derived from the URL instead of triggered imperatively on card click) — worth a quick regression pass on rapid conversation switching and the first-send temp\u2192real id swap.

Otherwise looks good.

PR #18299, opened 2026-08-14
deepdoc(pdf): WarpCrop de-skew + score-based layer-2 rotation (Go OCR parity with Python)

xugangqiang:
## Go OCR crop parity with Python `get_rotate_crop_image` (layer-1 de-skew + layer-2 score-based rotation)

> Supersedes / folds in #18305. The score-based layer-2 rotation selection from #18305 now lives here, on top of `WarpCrop` (layer 1), applied to **all three** Go OCR paths, together with the Python score plumbing the Go side depends on. #18305 is closed in favor of this PR.

### Problem

Python `get_rotate_crop_image` is one function with two layers:

- **Layer 1 — perspective de-skew:** `cv2.warpPerspective` the detected quad into a rectangular, horizontal crop.
- **Layer 2 — rotation selection:** when the crop is tall (`h/w >= 1.5`), recognize at `0° / CW90° / CCW90°` and keep the orientation with the **highest recognition score**.

The Go DeepDoc path sends crops to the Python rec service but did not replicate this, so skewed / vertical text was recognized poorly. The two earlier PRs were each only half of the function and conflicted:

| | Layer 1 | Layer 2 | Coverage |
|---|---|---|---|
| #18299 (before) | `WarpCrop` ✅ | `RotateIfVertical` — **fixed 90° CW**, ignores scores + `0°/CCW90°` ❌ | all 3 paths |
| #18305 (before) | un-de-skewed `FastCrop` ❌ | score-based `0/CW90/CCW90` ✅ (faithful) | 1 path only |

Neither was correct alone, and both touched `ocrDetectAndRecognize`. Layer 2 must run on layer 1's de-skewed output, so the two are **one change**.

### This PR (merged)

**Go — `internal/deepdoc/parser/pdf/`:**
- `util/warp.go`: `WarpCrop` does layer 1 only (perspective de-skew, bicubic, BORDER_REPLICATE, axis-aligned fast path → `FastCrop`, malformed-quad guards). Dropped the divergent `RotateIfVertical` / `WarpCropForOCR` / `rotate90CW`.
- `parser_ocr.go`: added `ocrRecognizeWithRotation` + `ocrBestScore` (from #18305) — score-based `0/CW90/CCW90` selection, single `0°` recognize for short/wide crops. Rewired **all three** OCR paths (`ocrDetectAndRecognize`, `ocrTableCells`, `buildTextBoxes`) to `WarpCrop` + `ocrRecognizeWithRotation`.

**Python — `deepdoc/`:**
- `vision/ocr.py`: `recognize_batch_with_score` — like `recognize_batch` but returns `(text, score)` tuples (preserves `drop_score` blanking). Added alongside `recognize_batch` so the in-process `__ocr` path keeps its text-only contract.
- `server/adapters/ocr_adapter.py`: `recognize()` surfaces the **real** recognition confidence instead of a constant `1.0` fill. The score-based layer 2 needs the real confidence, so the Python and Go sides ship together.

### Regression safety

- **Short / wide crops & axis-aligned boxes** (table cells, char-merge boxes): `WarpCrop` early-exits to `FastCrop` and `ocrRecognizeWithRotation` recognizes once at `0°` → behavior equals the old `FastCrop`+recognize. Guarded by `TestOCRMergeChars_WarpsEmptyBoxCrop` (asserts `WarpCrop == FastCrop` for axis-aligned boxes) and `TestOCRDetectAndRecognize_WarpsCrop` (asserts the de-skewed crop reaches recognition, emitted box stays the axis-aligned detection bbox).
- **Against a rec service without the #18305 score plumbing:** all candidate scores are equal → the `0°` result wins → matches `main`'s single-recognize behavior. Never worse than `main` in any deploy order.
- **`recognize()` callers:** the only caller (`ocr_endpoint.predict`) forwards the output unchanged, so surfacing the real score breaks nothing.
- Coordinate space is unchanged: only the crop fed to recognition is transformed; emitted `TextBox` bounds stay the axis-aligned detection bbox.

### Tests

- `bash build.sh --test ./internal/deepdoc/parser/pdf/...` — unit tier passes.
- `gofmt` + `go vet` clean; `ruff format` clean.
- `parser_ocr_rotate_test.go` — `ocrRecognizeWithRotation`: 1 call for short crops, 3 for tall, score-based pick, stable tie-break (0° wins).
- `parser_ocr_rotate_integration_test.go` (`//go:build integration`) — live rec service.
- `deepdoc/vision/test_ocr_score.py` — `recognize_batch_with_score` offline unit tests.
- `warp_test.go` / `warp_align_test.go` — `WarpCrop` golden + axis-aligned parity.

Both #18299 and #18305 were CI-green (preflight + go/python tests) before the fold.


tools/render_diff/warp_align.py
```diff
@@ -0,0 +1,311 @@
...
+        json.dump(boxes, f, indent=2)
+    logging.info("[genquads] %d quads (from %d words, angles=%s) -> quads.json + py_warp_*.png",
+                  len(boxes), len(words), angles)
+
+
+def pixel_mse(a_path, b_path):
+    A = np.asarray(cv2.imread(a_path), dtype=np.float64)
+    B = np.asarray(cv2.imread(b_path), dtype=np.float64)
+    if A.shape != B.shape:
+        B = cv2.resize(B, (A.shape[1], A.shape[0]))
```
Daniele-Cangi, line 174: Please fail on shape mismatch instead of resizing one crop before MSE. Width and height are part of the parity contract, so resizing can hide a real divergence. For example:

q = [
    (10.0, 10.0),
    (100.5, 10.0),
    (100.5, 67.999999),
    (10.0, 67.999999),
]

OpenCV’s float32 points produce 90x58, while the Go float64 distance calculation truncates to 90x57; the current resize turns that contract failure into an ordinary MSE value.

Also, _quad_from_word only rigidly rotates axis-aligned rectangles, so every generated quad is a parallelogram and the independent cv2 run never exercises non-zero projective terms. Please assert exact dimensions first and add convex corner perturbations or trapezoids to the cv2 corpus so the 8-DOF path is independently covered.

internal/deepdoc/parser/pdf/util/warp.go
```diff
@@ -0,0 +1,264 @@
...
+
+// WarpCrop de-skews a quadrilateral region from src using a perspective
+// transform, producing the rectangular crop fed to text recognition.
+//
+// points must be the 4 corners in order: top-left, top-right, bottom-right,
+// bottom-left (the DBNet quad order emitted by the OCR detector). The output
+// size is (W, H) where
+//
+//	W = int(max(|p0-p1|, |p2-p3|))
+//	H = int(max(|p0-p3|, |p1-p2|))
```
Daniele-Cangi, line 23: Please bound and validate the detector quad before deriving the output dimensions. OCRDetect accepts coordinates from the configured DocAnalyzer/DEEPDOC_URL, and this path allocates from those raw values before any source-bound clamp. A finite malformed response can crash the process rather than reach the fallback:

src := image.NewRGBA(image.Rect(0, 0, 10, 10))
pts := [4]Pt{
{X: 0, Y: 0},
{X: 3e18, Y: 0},
{X: 3e18, Y: 2},
{X: 0, Y: 2},
}
_ = WarpCrop(src, pts)
// panic: image: NewRGBA Rectangle has huge or negative dimensions

The first-party Python detector clips its points, but that invariant is not enforced at this Go boundary. Please reject non-finite coordinates, clamp or validate against src.Bounds() and enforce a bounded output area before image.NewRGBA. A regression test should prove that an out-of-range quad cannot panic or trigger an unbounded allocation.

Daniele-Cangi (changes requested):
I traced both OCR paths against the Python implementation and exercised the new warp independently. The remaining FastCrop in buildTextBoxes is not a missed integration point: Python __ocr also reduces those fallback regions to an axis-aligned rectangle before recognition, so I am not asking to change that path.

Two issues remain before this is safe to merge:

WarpCrop derives allocation dimensions directly from detector coordinates. Unlike the previous FastCrop path, those values are not bounded to the source image before allocation, so a malformed but finite analyzer response can panic the whole parser process.
The manual comparison currently normalizes away output-size mismatches, while its generated quads are rigid rotations of rectangles and therefore exercise only affine/parallelogram cases. That leaves the stated exact-size contract and the genuinely projective part of the 8-DOF implementation without an independent oracle.

The perspective mapping otherwise looks directionally sound, but these boundary and validation gaps should be closed first.

xugangqiang:
Thanks — agreed that the dimension mismatch must surface rather than be hidden, and that is exactly what the latest commit does.

`pixel_mse` now raises on a `WarpCrop`↔`cv2` shape mismatch; the compare loop catches it, logs a `DIMENSION MISMATCH` count, and skips that box. Resize is kept **only** for the `FastCrop` baseline, whose shape legitimately differs from `cv2` (axis-aligned bbox vs de-skewed quad), so it cannot be compared by dimension anyway. The convex corner perturbation in `_quad_from_word` also addresses your second point: `cv2` now exercises the full 8-DOF path, not just an affine map on parallelograms.

On "assert exact dimensions": I deliberately did **not** make `WarpCrop` force its output size to equal `cv2`'s exactly. The residual ≤1px difference you demonstrated is a `float32` (`np.linalg.norm` in cv2) vs `float64` (`math.Hypot` in Go) truncation artifact at integer boundaries — it is **not** a geometry bug. The de-skew transform, sampling, and output content are identical; the recognizer consumes arbitrary crop sizes, so this has zero functional impact on OCR text. Forcing equality would mean emulating cv2's `float32` arithmetic in Go, which only couples crop dimensions to float32 rounding quirks and is brittle. So the contract we enforce is: dimensions are stable and correct, and any residual float-precision divergence is *surfaced* (counted) instead of silently resized away. The MSE comparison still runs normally for all shape-matching boxes.

Happy to add an explicit ≤1px float-precision tolerance assertion to the harness if you'd prefer that spelled out.

Daniele-Cangi (commented):
Thanks for the quick follow-up. The changes directly address the two concerns I raised, including the allocation guard and the independent perspective/shape validation. I haven’t re-run the updated branch end-to-end, so I’m leaving this as a comment rather than an approval.

xugangqiang:
Thanks for the detailed cost analysis of the Layer 1 warp. I checked each point against the actual code, and most of the concern is already mitigated (or does not apply to this implementation):

**1. `toRGBA` re-converting the whole page per box — this does not happen.** `toRGBA` short-circuits when the input is already `*image.RGBA`: it returns the same pointer with no allocation or copy (`internal/deepdoc/parser/pdf/util/warp.go`). And `pageImg` is always `*image.RGBA` here — it comes from `pdfium.RenderPage` (`internal/deepdoc/parser/pdf/pdfium/pdfium.go:91`), which returns `*image.RGBA`. So `rgba := toRGBA(src)` inside `WarpCrop` is an O(1) type assertion per box, not an `O(N × pagePixels)` conversion. Hoisting `toRGBA` out of the per-box path would yield no measurable benefit because the conversion is already a no-op. The only way this becomes real is if a future caller passes a non-RGBA image (e.g. a JPEG-decoded `YCbCr`), which none of the current `WarpCrop` call sites do.

**2. Axis-aligned early-return — already implemented.** `WarpCrop` now checks `axisAligned(pts)` (edges parallel to the axes within ε) and returns `FastCrop` directly, skipping the homography solve and the per-pixel bicubic entirely. Table cells, char-derived boxes, and most detection boxes are axis-aligned, so the perspective cost is paid only for genuinely slanted quads. (`TestWarpCropAxisAlignedEarlyReturnsFastCrop` pins this.)

**3. PNG vs JPEG payload — acknowledged, left as-is.** `EncodePNG` is intentional: de-skew needs pixel-exact crops and JPEG artifacts would hurt recognition on the already sub-pixel-corrected text. It is a deliberate fidelity/IO tradeoff; happy to revisit only if profiling shows I/O is the bottleneck.

**4. "3 warps + 3 recognizes" for tall boxes — not applicable to this implementation.** That describes deepdoc's *score-based* Layer 2 (original / CW / CCW, pick best confidence). Our Layer 2 is a single geometric CW-90° flip (`RotateIfVertical`), so each tall box costs exactly 1 warp + 1 flip + 1 recognition, not 3. (Score-based selection would require 3× remote recognition calls per tall box, which we deliberately avoided.)

Net: per-box cost is now `O(crop pixels)` — axis-aligned via the `FastCrop` copy, slanted via the bicubic warp — the same order as the old `FastCrop` path, with no whole-page work per box. Dense pages will not see Layer 1 amplification.

Daniele-Cangi:
Nice to see the full Layer 1 + Layer 2 path land together. Glad the review helped tighten the boundary cases and validation.

PR #18301, opened 2026-08-14
feat(bedrock): support API key authentication

### LINKED ISSUE #17848: [Feature Request]: Add Bedrock API key authentication to the existing Bedrock provider
### Self Checks

- [x] I have searched for existing issues, pull requests, and their comments, including closed ones.
- [x] I understand that non-English title submissions will be closed directly.
- [x] I have not modified the required template structure and have filled in all required fields.

```Markdown
RAGFlow's existing Bedrock provider currently supports three AWS SigV4-based authentication modes: `access_key_secret`, `iam_role`, and `assume_role`.

The current design originated in #15165 and #15166 and intentionally mirrored the existing Python implementation. However, Amazon Bedrock has also officially supported Bearer API keys since July 2025, including short-term and long-term keys. AWS documents using these keys through either `AWS_BEARER_TOKEN_BEDROCK` or an `Authorization: Bearer <api-key>` request header.

Users who receive a Bedrock API key currently cannot configure it in RAGFlow's Bedrock provider. This is especially relevant to the newer `bedrock-mantle` endpoint, which exposes OpenAI- and Anthropic-compatible APIs and model discovery endpoints such as:

https://bedrock-mantle.ap-northeast-1.api.aws/v1/models

Bedrock API key authentication is an authentication option for Amazon Bedrock, not a separate model provider. Creating another provider entry would split Bedrock configuration, models, and credentials across multiple providers.
```

Add Bedrock API key authentication as another authentication mode of the existing `Bedrock` provider.

The same Bedrock provider should expose the following authentication choices:

- Access Key and Secret Access Key
- IAM Role
- Default AWS credential chain, including IRSA and instance profiles
- Bedrock API Key

When Bedrock API Key is selected, users should be able to provide:

- A Bedrock API key
- An AWS Region
- An optional Bedrock endpoint URL or endpoint type when a non-default endpoint such as `bedrock-mantle` is used

The API key should remain scoped to the configured Bedrock provider instance. It should not be stored or applied through a process-global environment variable shared by other tenants or provider instances.

Both short-term and long-term Bedrock API keys use Bearer authentication and should be accepted. The UI or documentation should recommend short-term keys for production and explain the security implications of long-term keys.

This feature should remain within the existing `Bedrock` provider rather than introducing a separate provider such as `Bedrock API` or `Bedrock Mantle`.

Extend the existing Bedrock credential object with an additional authentication mode and a mode-specific `bedrock_api_key` field.

At a high level, the Bedrock provider can select between:

- SigV4 signing for Access Key, IAM Role, and default credential-chain modes
- Request-scoped Bearer authentication for Bedrock API Key mode

The endpoint protocol should be treated separately from the authentication mode:

- Existing Bedrock and Bedrock Runtime endpoints can use the current Converse, InvokeModel, and model-catalog paths with Bearer authentication.
- A `bedrock-mantle` endpoint can use its supported OpenAI- or Anthropic-compatible API paths while remaining part of the same Bedrock provider.

Model discovery should use the configured endpoint when available instead of assuming that every Bedrock instance has the same static model list.

Amazon Bedrock API keys are limited to Bedrock and Bedrock Runtime operations. Features backed by unsupported services, such as `bedrock-agent-runtime` reranking, should return a clear unsupported-authentication error instead of silently falling back to another credential mode.

The implementation should avoid setting `AWS_BEARER_TOKEN_BEDROCK` globally in the RAGFlow process because concurrent tenants or provider instances could overwrite each other's credentials.

```Markdown
This is useful for users who:

- Receive a Bedrock API key instead of AWS Access Key and Secret Access Key credentials.
- Use Amazon Bedrock's `bedrock-mantle` endpoint with Anthropic- or OpenAI-compatible client behavior.
- Cannot grant RAGFlow access to an IAM role or the AWS default credential chain.
- Want to configure different Bedrock regions or endpoint families as separate instances under the same provider.
- Need remote model discovery so that region-specific model availability does not depend on a stale static catalog.

AWS documentation:

- Amazon Bedrock API keys:
  https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-reference.html
- Using a Bedrock API key:
  https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html
- Permissions for Bedrock and Bedrock Mantle Bearer tokens:
  https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-permissions.html
```

### Additional information

This request does not propose replacing the existing SigV4 authentication paths. It adds an officially supported authentication option to the same Bedrock provider.

Related existing work:

- #15165 introduced the Go Bedrock provider design.
- #15166 implemented the current three SigV4-based authentication modes.
- #12008 added the IAM role and default credential-chain requirements.

None of these issues or pull requests, including their comments, discuss support for Amazon Bedrock Bearer API keys.

whhe:
### Summary

Closes #17848.

This replaces #17958 with a smaller implementation focused on the original requirement. The previous PR gradually expanded beyond the necessary scope; the optional custom endpoint extension is intentionally excluded from this version.

- Add request-scoped Bedrock API key authentication to the Python Bedrock adapters and Go chat runtime.
- Discover and persist models available to API key instances, so users do not need to manually enter the model type, model name, or maximum token count.
- Add API Key mode to the existing Bedrock settings UI and refresh the model and default-model lists after instance changes.
- Preserve selected models during credential-only updates while keeping the existing SigV4 authentication modes unchanged.
- Return a clear error for unsupported API-key reranking.
- Document instance-scoped authentication and short-term API key guidance.


JinHai-CN:
@whhe When you finish, please notify us and we will start CI testing.

web/src/hooks/use-llm-request.tsx
```diff
@@ -268,20 +268,15 @@ export const useAddProviderInstance = () => {
...
-          }
+        const { data: instancesRes } = await llmService.listProviderInstances(
+          { provider_name: params.llm_factory },
+          true,
+        );
+        const instanceExists = instancesRes?.data?.some(
+          (i: IProviderInstance) => i.instance_name === params.instance_name,
+        );
+        if (instanceExists && !params.verify) {
+          return { code: 0, data: null };
```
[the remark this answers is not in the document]
whhe: I rechecked the current call sites and no change is needed here.

`useAddProviderInstance` is now used only for creating draft instances. Saved instances are routed through `useUpdateProviderInstance` and the dedicated update endpoint via `apiKind: 'update'`. Therefore an update cannot reach this duplicate-create short-circuit.

Reintroducing `id` into the add request would restore the obsolete add-as-update path.

internal/router/router.go
```diff
@@ -550,6 +550,7 @@ func (r *Router) Setup(engine *gin.Engine) {
 				provider.GET("/:provider_id_or_name", r.providerHandler.ShowProvider)
 				provider.DELETE("/:provider_id_or_name", r.providerHandler.DeleteProvider)
 				provider.GET("/:provider_id_or_name/models", r.providerHandler.ListModels)
+				provider.POST("/:provider_id_or_name/models", r.providerHandler.ListModels)
```
Haruko386: Bro, like what's this for?

Lynn-Inf (commented):
Thanks for the PR. I have a few concerns:

1. **POST method addition**: I don't think we should add a POST method to this interface. If no fields have changed, the API's semantic meaning should remain the same. For the `api_key` issue where it's being passed as a dict, we can simply format it as a JSON string — no need to change the HTTP method.

2. **PUT semantics for** `update_provider_instance`: The same principle applies here. We should keep the PUT semantics intact — please preserve the logic that deletes all models under an instance when an empty list is passed in the `model_info` field.

Other than that, the rest LGTM.

internal/entity/models/bedrock.go
```diff
@@ -562,6 +575,25 @@ func signBedrockRequest(ctx context.Context, req *http.Request, body []byte, cre
...
+	if key.AuthMode == bedrockAuthAPIKey {
+		expectedHost := fmt.Sprintf(bedrockRuntimeHostTmpl, region)
+		if service == bedrockControlService {
+			expectedHost = fmt.Sprintf(bedrockControlHostTmpl, region)
+		}
+		if req.URL.Scheme != "https" || !strings.EqualFold(req.URL.Host, expectedHost) || req.URL.User != nil {
+			return fmt.Errorf("bedrock: API key authentication requires the regional AWS Bedrock endpoint")
+		}
+		req.Header.Set("Authorization", "Bearer "+key.APIKey)
+		return nil
```
Haruko386, line 596: Confirmed against the AWS docs: sending `Authorization: Bearer <api-key>` to the control-plane `ListFoundationModels` endpoint (`GET https://bedrock.<region>.amazonaws.com/foundation-models`) is a supported use of Bedrock API keys — the reference (docs.aws.amazon.com/bedrock/latest/userguide/api-keys-reference.html) states keys are valid for both Amazon Bedrock (control plane) and Bedrock Runtime actions, with only `InvokeModelWithBidirectionalStream`, Agents, and Data Automation excluded. So this approach is fine.

> [!note]
> One thing worth surfacing to users, though: **short-term API keys can only be used in the region they were generated in**. Since the run-time region here comes from `resolveBedrockRegion` (bedrock.go:188, which prefers the instance's `extra["region"]` over the key blob's `bedrock_region`), a key generated in us-east-1 configured against a us-west-2 instance will fail with AWS 401/403 — and `ListModels` surfaces the raw AWS error without hinting at the region mismatch. Consider validating against the key's own region (or at least including the region in the error message) so the failure is diagnosable.

Other issues plz follow the comment of @Lynn-Inf, it is the same issues for `Go`

Haruko386 (changes requested):
Hi @whhe thank u for ur contribution.
Would you fix this issue?

api/apps/services/provider_api_service.py
```diff
@@ -259,7 +298,15 @@ def show_provider_model(provider_id_or_name: str, model_name: str):
...
 async def update_provider_instance(
-    tenant_id: str, provider_id_or_name: str, instance_id_or_name: str, instance_name: str, api_key: str | dict, base_url: str, region: str, model_info: list[dict] = None, verify: bool = True
+    tenant_id: str,
+    provider_id_or_name: str,
+    instance_id_or_name: str,
+    instance_name: str,
+    api_key: str | dict,
+    base_url: str,
+    region: str,
+    model_info: list[dict] = None,
```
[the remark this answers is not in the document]
whhe: These RUF013 findings already exist on the base branch and were not introduced by this PR. This PR actually reduces the number of RUF013 findings in this file from eight to six.

To keep the change focused on Bedrock API-key support and avoid unrelated type-annotation cleanup, I’m leaving the remaining existing annotations unchanged.

Haruko386 (commented):
LGTM, waiting for @Lynn-Inf 's review

Lynn-Inf (approved):
LGTM. This can be merged once conflicts are resolved.

JinHai-CN:
```
--- FAIL: TestBedrockAPIKeyInstancePersistsDiscoveredModelsWithoutRuntimeVerification (0.00s)
    model_service_test.go:345: verify = <nil>, want "unknown"
```

dcc123456 (approved):

internal/handler/providers.go
```diff
@@ -198,12 +245,22 @@ func (h *ProviderHandler) ListModels(c *gin.Context) {
...
 	for _, m := range remoteModels {
 		if name, ok := m["name"].(string); ok {
+			if existing, exists := merged[name]; exists {
+				if maxTokens, ok := existing["max_tokens"]; ok && maxTokens != nil {
+					m["max_tokens"] = maxTokens
+				}
+			}
 			merged[name] = m
 		}
 	}
```
Haruko386, line 269:
```suggestion
	merged := make(map[string]map[string]interface{})
	for _, m := range staticModels {
		if maxTokens, ok := m["max_tokens"]; !ok || maxTokens == nil {
			if maxOutput, ok := m["max_output"]; ok && maxOutput != nil {
				m["max_tokens"] = maxOutput
			}
		}
		if name, ok := m["name"].(string); ok {
			merged[name] = m
		}
	}
	for _, m := range remoteModels {
		if name, ok := m["name"].(string); ok {
			if len(providerModelMapTypes(m)) == 0 {
				if existing := merged[name]; len(providerModelMapTypes(existing)) > 0 {
					m["model_types"] = existing["model_types"]
				}
			}
			merged[name] = m
		}
	}
```

Haruko386 (changes requested):
Hi @whhe would you resolve conflict? I think this is a better way to resolve it

whhe:
> Hi @whhe would you resolve conflict? I think this is a better way to resolve it

Thanks! I adopted the suggested model-type merge. I kept the static max_tokens inheritance because Bedrock discovery does not provide that metadata.

Lynn-Inf (approved):
LGTM.

PR #18305, opened 2026-08-14
deepdoc: expose real OCR rec score and use it for Go layer-2 rotation

xugangqiang:
## Summary

The Go PDF OCR path was missing **layer-2 rotation selection** (PaddleOCR-style): for tall, narrow crops (height >= 1.5x width, i.e. vertical text lines) it only recognized at 0°, so vertical text was read as garbage.

The reason was architectural: the DeepDoc rec service previously filled recognition confidence with a constant `1.0` (`OCR.recognize_batch` discards the per-item score, and the adapter had nothing else to send). Without a usable score, the Go side could not pick the best orientation the way the in-process Python `__ocr` path does, and fell back to a text-volume heuristic.

This PR makes the rec service surface the **real recognition score** and switches the Go layer-2 selection to use it — matching Python's `get_rotate_crop_image` / `_rotate_select` behavior.

## Changes

- `deepdoc/vision/ocr.py`: add `recognize_batch_with_score` which returns `List[(text, score)]` and preserves the existing `drop_score` blanking. The original `recognize_batch` is **unchanged**, so the in-process `__ocr` path keeps its text-only contract.
- `deepdoc/server/adapters/ocr_adapter.py`: `recognize()` now returns `[[text, real_score]]` instead of `[[text, 1.0]]`, surfacing the real confidence to clients. Removed `_CONFIDENCE_FILL`.
- `internal/deepdoc/parser/pdf/parser_ocr.go`: `ocrRecognizeWithRotation` recognizes a crop at `0°/CW90°/CCW90°` (only when `h/w >= 1.5`) and keeps the orientation with the highest recognition confidence, replacing the text-volume heuristic.
- Tests: Python offline unit test for `recognize_batch_with_score` + adapter; Go unit tests asserting score-based selection.

## Verification

- Go: `build.sh --test ./internal/deepdoc/parser/pdf/` passes (short crops recognized once at 0°; tall crops tried at 3 orientations and selected by highest score; stable tie-break keeps 0°).
- Python: offline unit tests for the new method/adapter pass; `ruff` clean.
- End-to-end against the rebuilt `deepdoc_oss` image: a vertical crop (h/w = 4.56) was recognized at 0/CW90/CCW90 and the correctly oriented crop won by confidence (~0.97); both a direct service call and the Go client received the **real** (non-1.0) score and selected the upright orientation.

## Notes / no regression

- Horizontal crops (`h/w < 1.5`) are recognized once at 0° — behavior identical to before.
- The in-process Python `__ocr` path is untouched (uses the unchanged `recognize_batch`).
- `OCRText.Confidence` is never consumed as a threshold anywhere in the Go pipeline (`pdf.TextBox` carries no confidence field), so the real score triggers no downstream filtering/dropping.
- This is the **layer-2** change only. **Layer 1** (perspective de-skew / WarpCrop) is tracked separately in PR #18299.

### DIFF
--- a/deepdoc/server/adapters/ocr_adapter.py
+++ b/deepdoc/server/adapters/ocr_adapter.py
@@ -15,9 +15,6 @@ from deepdoc.vision.ocr import OCR
 
 logger = logging.getLogger(__name__)
 
-# Confidence fill value — OSS recognize_batch does not return confidence scores.
-_CONFIDENCE_FILL = 1.0
-
 
 class OCRAdapter:
     """Calls OCR.detect() and OCR.recognize_batch(), converts to wire format."""
@@ -83,10 +80,12 @@ class OCRAdapter:
 
         img = self._decode_bgr(image_data)
 
-        # OCR.recognize_batch() returns List[str]; single cropped image → list of 1 image
-        texts = self._ocr.recognize_batch([img])
+        # OCR.recognize_batch_with_score() returns List[(text, score)] so the
+        # client can do score-based layer-2 rotation selection. The score is
+        # the real recognition confidence (previously a constant 1.0 fill).
+        scored = self._ocr.recognize_batch_with_score([img])
 
-        items = [[text, _CONFIDENCE_FILL] for text in texts]
+        items = [[text, score] for text, score in scored]
 
         # 4-level nesting matching Go [][][][]any:
         # batch → page → items list → pair [text, confidence]
--- a/deepdoc/vision/ocr.py
+++ b/deepdoc/vision/ocr.py
@@ -162,7 +162,6 @@ class TextRecognizer:
         return padding_im
 
     def resize_norm_img_vl(self, img, image_shape):
-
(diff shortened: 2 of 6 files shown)


xugangqiang:
Folded into #18299 — closing in favor of that PR.

The score-based layer-2 rotation selection now lives in #18299 on top of `WarpCrop` (layer 1), applied to **all three** Go OCR paths (`ocrDetectAndRecognize`, `ocrTableCells`, `buildTextBoxes`), together with the Python score plumbing this PR contributed (`recognize_batch_with_score` + the adapter surfacing the real recognition confidence instead of a constant `1.0`).

Layer 2 must run on layer 1's de-skewed output, and the Go score-based selection needs the real confidence from the Python side — so the two halves of `get_rotate_crop_image` belong in one PR. #18299 now carries both layers and the Python plumbing as a single, parity-faithful change.

Thanks @CodeBuddy for the layer-2 work carried over.

PR #18307, opened 2026-08-14
fix(chunker): preserve ancestor heading path for text after media in HierarchyTitleChunker

### LINKED ISSUE #18293: [Bug]: Hierarchy TitleChunker drops ancestor heading path for text after image/table flush
### Self Checks

v0.26.4

### RAGFlow image version

v0.26.4

```Markdown
ubuntu
```

## Describe the bug

When using **Title Chunker** with method **`hierarchy`**, encountering a structured **image** or **table** (`doc_type_kwd != "text"`) splits the text stream into separate tree-building runs (`flush_text_records` / per-run tree build).

Text **before** the media chunk correctly carries the ancestral heading path (e.g. `Chapter 1 › 1.1 …`).

Text that still belongs to the **same section** but appears **after** the image/table is chunked in a **new** tree build where the heading lines were already consumed by the previous flush. As a result, those later chunks **lose the parent heading path** and become poorly contextualized for retrieval.

This is separate from (but related to) “media are standalone chunks”. The surprising part is that **post-media body text under the same heading also loses hierarchy context**.

### Expected behavior

Chunks for “text after the figure” that still belong to 1.1 should keep the same ancestral path as the pre-figure text, e.g.:

Chapter 1 › 1.1 Background + text after the figure
Ideally, media would either:

stay inside the same hierarchical section without resetting the title stack, or
after flush, continue with a preserved ancestor heading stack for subsequent body text.

### Steps to reproduce

```Markdown
## Steps to reproduce

1. Create an ingestion pipeline: `Parser` → `TitleChunker` (`method=hierarchy`, e.g. hierarchy depth H2) → `Tokenizer` (no Token Chunker required).
2. Ingest a document whose parser output is structured `json`/`chunks` with an image or table in the middle of a section, for example:


# Chapter 1
## 1.1 Background
Text before the figure...
[image or table block]
Text after the figure (still under 1.1)...
## 1.2 Goals
...
```

### Additional information

_No response_

peewee92:
## Problem

When using **TitleChunker** with `method=hierarchy`, encountering a structured image or table (`doc_type_kwd != "text"`) mid-section triggers `flush_text_records`, which builds a heading tree from the accumulated text run and then **clears all records**.

Text that still belongs to the same section but appears **after** the media chunk starts with empty accumulators. The heading lines were already consumed by the previous flush, so the post-media text loses its parent heading path and becomes poorly contextualized for retrieval.

**Before the fix** (example from the issue):
```
# Chapter 1
## 1.1 Background
Text before the figure...     → chunk: [Chapter 1 › 1.1 Background + text] ✓
[image or table block]         → standalone chunk ✓
Text after the figure...       → chunk: [text only — NO heading context] ✗
```

Closes #18293

## Root Cause

`flush_text_records` (Python: `hierarchy_chunker.py`, Go: `hierarchy.go`) cleared `text_records`/`text_levels` (or `textRun`/`textLevels`) entirely after building each run's tree. There was no mechanism to carry heading context across media-chunk flush boundaries.

## Fix

After flushing, **retain heading records** (those with `0 < level < BODY_LEVEL`) as a prefix of the next run, so post-media body text inherits the correct ancestor heading path.

A `context_boundary` (Go: `contextBoundary`) marks how many leading records in the current run are retained headings. During DFS path emission, paths that contain **only** retained records (no new content at or beyond the boundary) are **skipped**, preventing duplicate heading-only chunks.

### Trace (hierarchy=2, include_heading_content=true)
```
# Chapter 1         (level 1)
## 1.1 Background    (level 2)
text before figure   (BODY)
[IMAGE]              (non-text)
text after figure    (BODY)
## 1.2 Goals         (level 2)
text goals           (BODY)
```

**Run 1** (before image): `[Ch1, 1.1, text before]` → tree → path `[Ch1, 1.1, text before]` ✓
Retained headings: `[Ch1, 1.1]`, boundary = 2

**[IMAGE]** → standalone chunk ✓

**Run 2** (after image): `[Ch1, 1.1, text after, 1.2, text goals]`
- Tree: `Ch1 → 1.1 (body: text after)`, `Ch1 → 1.2 (body: text goals)`
- Path `[Ch1, 1.1, text after]` — index 2 ≥ boundary 2 → **emit** ✓
- Path `[Ch1, 1.2, text goals]` — index 3 ≥ boundary 2 → **emit** ✓
- No duplicate heading-only chunks for the retained `1.1` node

## Changes

| File | Language | Change |
|---|---|---|
| `rag/flow/chunker/title_chunker/hierarchy_chunker.py` | Python | Retain headings after flush; filter DFS paths by `context_boundary` |
| `internal/ingestion/component/chunker/hierarchy.go` | Go | Same fix (mirrors Python) |
| `internal/ingestion/component/chunker/hierarchy_test.go` | Go test | Two new tests: one-level and two-level ancestor preservation |
| `rag/flow/tests/test_title_chunker_ancestor_context.py` | Python test | Three new tests: ancestor preservation, nested ancestor, no duplicate heading-only chunks |

## Verification

- **Python tests**: all 3 new tests pass (run via stub-loaded real `HierarchyTitleChunker` chain, same pattern as `test_title_chunker_position_int.py`).
- **Python lint**: `ruff check` and `ruff format` clean on both modified files.
- **Go tests**: Go is not installed in the current environment; tests follow the existing `hierarchy_test.go` patterns and the code mirrors the Python fix exactly. CI will run `build.sh --test ./internal/ingestion/component/chunker/...`.
- **Existing tests**: the existing `test_title_chunker_preserves_position_int_from_deepdoc_json` Python test still passes.


xugangqiang:
## Review — issues to address before merge

I read the diff and traced the algorithm against the current `hierarchy_chunker.py` / `hierarchy.go`. The fix direction is correct and the core mechanism (retain heading prefix + skip paths whose indexes are all `< context_boundary`) is sound — I verified single-level, nested, sibling-reset, and consecutive-media cases by hand. Below are the concrete problems/risks, **especially on test coverage**.

### 1. Test coverage gaps (most important)

The new tests cover: single-level ancestor, nested ancestor, sibling context, image-chunk isolation, and no-duplicate-heading-only. Good direction, but missing cases that this change can regress:

- **`include_heading_content=False` path is completely untested.** With `include_heading_content=False`, leaf paths merge titles+body, and the `index >= context_boundary` filter still has to keep the post-media chunk. Please add a test that pins post-media body text retains the ancestor heading when `include_heading_content=False`.
- **Consecutive non-text chunks (`image → image → text`) are untested.** The intermediate flush re-derives retained headings and must not emit a duplicate heading-only chunk. Worth pinning explicitly.
- **Media at the very start / immediately after a heading** (e.g. `[image]` then `# Ch1 ...`, or `# Ch1` then `[image]` with no body) — boundary behavior of `context_boundary` should be pinned so a future refactor can't silently change it.
- **Go tests were not actually executed** (PR notes Go is not installed locally; CI will run). The two Go tests only mirror the Python logic by reasoning. Please confirm `bash build.sh --test ./internal/ingestion/component/chunker/...` actually passes in CI before merge — a mirror that isn't run proves nothing.

### 2. Regression risk for existing parity / golden snapshots

This change alters output for **any document that has a non-text chunk (image/table) mid-section** — which is the norm for real PDF/DOCX. Any golden/parity snapshot that encoded the *old (buggy)* behavior will now mismatch. The PR updates no golden/parity files. Action required:

- Run the full `rag/flow/tests/test_title_chunker*.py` suite.
- Refresh any chunker parity golden files affected by the new heading context. This is a legitimate bug fix, so updating goldens is correct — but it must be done deliberately, not left to rot in CI.

### 3. Minor code robustness (non-blocking)

In `hierarchy_chunker.py` the retained heading lists are built separately:

```python
heading_records = [rec for rec, lvl in zip(text_records, text_levels) if 0 < lvl < BODY_LEVEL]
heading_levels  = [lvl for lvl in text_levels if 0 < lvl < BODY_LEVEL]
```

Both rely on the filter being positionally aligned. Correct today, but fragile. Prefer a single paired comprehension:

```python
retained = [(rec, lvl) for rec, lvl in zip(text_records, text_levels) if 0 < lvl < BODY_LEVEL]
heading_records = [r for r, _ in retained]
heading_levels  = [l for _, l in retained]
```

(The Go side already iterates paired `textRun`/`textLevels`, so no change needed there.)

### On the CodeRabbit comments

Both are reasonable but low-severity style nits, not correctness issues:
- "strict zip" is moot here (the paired lists are equal length, and `heading_levels` doesn't even use `zip`); the paired-construction suggestion above is the more meaningful version of that point.
- Optional debug logging for flush/retained-heading counts is nice-to-have — if added, route it through the existing `common.Debug`/`slog` channels already used in `hierarchy.go`, not a new logger.
- The `›` → `/` docstring change is cosmetic.

**Bottom line:** algorithm is correct and worth merging, but please (a) add the `include_heading_content=False` + consecutive-media + boundary tests, (b) verify the Go suite actually runs green in CI, and (c) refresh affected parity goldens. These are the real blockers; the CodeRabbit nits are optional.

PR #18308, opened 2026-08-14
fix(pdf): strip tiled watermark text from text-layer extraction

### LINKED ISSUE #18145: PDF parser interleaves tiled text-layer watermark into every chunk (degrades LLM extraction)
## Problem

Parsing a resume PDF (downloaded from a Chinese resume-template site) with the Go DeepDOC PDF parser (`internal/deepdoc/parser/pdf`) produces chunk text polluted by a **tiled, repeated watermark string**. The watermark glyphs are interleaved with the real resume content in reading order, so **every** chunk carries the garbage — and the garbage is then sent verbatim to the LLM by the ingestion Extractor, wasting tokens and degrading metadata-extraction quality.

Real user message that reached the LLM during resume metadata extraction (field_name=metadata):

```text
Content: ~~
Qcd623406a1ecf65c1HZ42dm0GFRTxYi2V_ydWOWgnP_UNhRq2Q~~ 2qRhNU_P
ng
W
O
Wdy_V2iYxTR
简   历 ZH1c56f姓    名：冯浩楠                         性    别：男 ce1a60年    龄：22 岁 ...
F
G0
md24
~
~
教育经历 q2021.09 - 2025.07    南阳理工学院     软件工程专业  本科
Q
2
R
h
N
U相关技能： 1. 熟悉软件测试的理论与流程 ...
```

## Root cause

The noise is **not** OCR garbage — it is a tiled watermark stored as **real text in the PDF's text layer**, read deterministically and interleaved by position. Evidence:

1. The token sequence `Q2qRhNU_Png W O Wdy_V2iYxTR F G0 md24 ZH1c56f ce1a604326dc` repeats **verbatim ~6×** across pages/chunks. A 66-char random string would never be OCR'd identically 6 times — deterministic text-layer extraction.
2. The tokens decode to random bytes (not an encoded message) — they look like template-site document-ID watermarks.
3. Single glyphs on their own lines (`Q`, `2`, `R`, `h`, `N`, `U`, `_`, `P`, `n`, `g`, `W`) are the signature of a **rotated/diagonal watermark**: `GroupCharsToLines` does not vertically overlap the rotated glyphs, so each becomes its own line. `Q2RhNU` is a substring of the token `Q2qRhNU_P` — the scattered glyphs are the watermark string read glyph-by-glyph.
4. The `~~...~~` wrappers around one token are real PDF characters — the `~~` marker is **not generated anywhere** in the Go/Python ingestion or deepdoc code.
5. Wide CJK spacing (`姓    名：冯浩楠`) confirms text-layer extraction: `LineToTextBox` never synthesizes spaces between CJK glyphs, so those spaces must be literal space glyphs in the PDF's text layer.

## Code location

The noise first enters the final text at the text-layer reconstruction stage:

- `internal/deepdoc/parser/pdf/layout/chars_boxes.go`
  - `CharsToBoxes` (line 27) / `GroupCharsToLines` (line 126) / `LineToTextBox` (line 183) — read the rotated watermark glyphs into isolated boxes.
- `internal/deepdoc/parser/pdf/layout/boxes_sections.go`
  - `BoxesToSections` (line 53) — the watermark text first lands in `Section.Text`.

Downstream, `internal/ingestion/component/chunker/group.go:385` (`extractLineRecords` → `itemText` at `common.go:127`) copies `Section.Text` verbatim into chunks, and `internal/ingestion/component/extractor.go` sends it to the LLM. Neither generates the noise.

The OCR-merge path (`parser_ocr.go` `ocrMergeChars` / `buildTextBoxes`) is **not** the source here.

## Impact

- Every chunk loses ~30–50% of its token budget to watermark garbage.
- The extractor prompt ("return only metadata explicitly supported by the content") is polluted; garbage can trigger hallucinated fields and dilutes real ones.
- Observed on the resume metadata extraction template (Extractor `field_name=metadata`, prompt `Content: {TitleChunker:FlatMiceFix@chunks}`).

## Reproduction

1. Upload a resume PDF that carries a tiled diagonal watermark in its text layer (common on free Chinese resume-template sites — the tokens look like `Q2qRhNU_P`, `Wdy_V2iYxTR`, `ZH1c56f`, `ce1a60`, `4326dc`, `G0`, `md24`).
2. Ingest with: Parser(`DeepDOC`, `preprocess: ["main_content"]`, `flatten_media_to_text: true`, `output_format: json`) → TitleChunker(`method: hierarchy`) → Extractor(metadata).
3. Inspect the chunk text produced by the Parser/TitleChunker — every chunk contains interleaved watermark fragments.

## Proposed fix

The strongest signal is **cross-page verbatim repetition** — legitimate body text does not repeat a 66-char token 6×. Candidate fixes, in preference order:

1. **Parser post-process filter** (`internal/parser/parser/pdf_postprocess.go`, alongside `applyRemoveTOC` / `filterPDFHeaderFooter`): drop sections whose text is a repeated high-base64-density token (appears ≥N times identically across the document, and/or matches a token-like shape `[A-Za-z0-9_\-]{20,}` with watermark-like position/font characteristics).
2. **Layout-stage filter** (`BoxesToSections` / `CharsToBoxes`): discard isolated single-glyph boxes that are not horizontally aligned with any neighbor (rotated-watermark fingerprint). Higher false-positive risk.
3. Reproduce first with `SkipOCR=true` vs `false` to definitively confirm the text-layer origin before choosing the filter.

## Notes

- This is in the Go ingestion/parser stack (`internal/ingestion`, `internal/deepdoc`, `internal/parser`), which is under active refactor.
- Separately, `cropImageChunks` for this same PDF emitted repeated `FPDF_LoadPage(4) returned nil` warnings (page-index 4 fails to load in pdfium — likely a declared-page-count vs loadable-page mismatch). Probably a distinct issue; mentioned here only because it surfaced in the same run.

Taranum01:
Fixes #18145

## Problem

Resume / template PDFs plant a tiled diagonal watermark in the text layer. The watermark string repeats verbatim across the page, but each glyph is rotated so it fails the vertical-overlap check in `GroupCharsToLines`. Every glyph becomes its own "line", and the resulting single-character boxes are interleaved into every chunk in the layout pipeline. The Extractor then forwards the watermark fragments to the LLM, wasting tokens and diluting metadata.

## Fix

Suppress watermark glyphs at the layout entry point by pre-scanning the page for repeated short alphanumeric tokens (≥3 occurrences, no whitespace, letters + digits mix) and dropping them before `GroupCharsToLines`.

The repetition threshold is the actual signal: a 66-char random string would never appear three times identically across the page by chance, so we don't need an OCR-quality detector. Single-occurrence candidates are left untouched, so legitimate codes in the document body are unaffected.

## Tests

- `looksLikeWatermarkCandidate` table over the issue's example tokens (rejecting empty / whitespace / pure-letter / pure-digit / oversize strings).
- `filterWatermarkChars` drops repeated watermark tokens while keeping unique plausible candidates.
- `CharsToBoxes_DropsWatermarkBeforeGrouping` end-to-end check that no emitted box contains the watermark text.

All layout package tests pass.


JinHai-CN:
@Taranum01 

```
╭──────────────────────────────────────────╮
│ 🥊 lefthook  v2.1.10   hook:  pre-commit │
╰──────────────────────────────────────────╯
│  web-checks ❯ web-deps (skip) no matching staged files
│  web-checks ❯ web-oxfmt (skip) no files for inspection
│  web-checks ❯ web-oxlint (skip) no files for inspection
│  ruff-format (skip) no files for inspection
│  ruff (skip) no files for inspection
┃  gofmt ❯ 
The following Go files are not gofmt'd:
internal/deepdoc/parser/pdf/layout/watermark_test.go

exit status 1
```

Taranum01:
Pushed a follow-up commit `7fc9730c` fixing the lefthook pre-commit `gofmt` failure @JinHai-CN flagged.

The trailing comments on the table-driven test cases in `internal/deepdoc/parser/pdf/layout/watermark_test.go` were column-misaligned (the empty-string entry had one less space than its siblings). `gofmt -w` on the file aligns them; `gofmt -l` now reports the file as clean and the `go test ./internal/deepdoc/parser/pdf/layout/...` suite still passes.

xugangqiang:
## Review: this change does not actually fix the reported bug

I reviewed against the root cause in #18145. The fix is well-intentioned but, as written, will not solve the problem and adds a new false-positive surface. Two structural issues, plus test gaps.

### 1. The filter sits on a code path the target document never takes
`filterWatermarkChars` is only called inside `CharsToBoxes`, and `CharsToBoxes` has exactly one caller: `internal/deepdoc/parser/pdf/parser.go:347`, inside the `if !ocrUsed && len(chars) > 0` branch.

For the resume PDF in #18145 (a real, clean text layer, not garbled):
- `hasCleanChars` is true, so `processPageBoxes` calls `ocrMergeChars` (parser_ocr.go:163).
- When DLA detects text boxes (it does for a resume), `ocrMergeChars` returns non-nil → `ocrUsed = true`.
- The `!ocrUsed` branch is skipped, so `CharsToBoxes` is never executed and `filterWatermarkChars` is **completely bypassed**.

So for the exact document type this PR targets, the new code is dead. Any watermark filtering must also run on the `ocrMergeChars`/`buildTextBoxes` path, or be moved to the document/section level (see #2).

### 2. The granularity is wrong: the dominant noise is per-glyph, not per-token
#18145 root cause (point 3) states the watermark is rotated so each *glyph* becomes its own line (`Q`, `2`, `R`, `h`, `N`, `U`, …). But `looksLikeWatermarkCandidate` (watermark.go) requires `len(text) >= 2` and letter+digit mixing. A single glyph `Text="Q"` fails both, so it is never a candidate and is never dropped. The bulk of the pollution survives even if the path above were taken.

Only whole-token instances (`ce1a60`, `ZH1c56f`, `Wdy_V2iYxTR`, …) that happen to appear ≥3× on a single page would be caught — a small subset. And the 66-char `~~…~~` token exceeds the `64` cap (watermark.go) so it is silently ignored.

Recommended direction (this is also #18145's fix #1, highest priority): do this at the **section/document level** in `internal/parser/parser/pdf_postprocess.go` (alongside `filterPDFHeaderFooter`/`applyRemoveTOC`, operating on the final `ParseResult` sections). The real signal is *verbatim repetition across the document*, not "3× on one page". A post-process filter survives the OCR-merge path and matches the true signal. The layout-stage per-`TextChar` heuristic is the wrong layer and wrong granularity.

### 3. Tests give false confidence and miss regressions
- All three tests model the watermark as a *whole-token* `TextChar{Text:"ce1a60"}`, which does **not** match the real per-glyph extraction described in the issue. They pass but do not reproduce the bug. Add a parity/replay test using a real resume `charspy` dump asserting the per-glyph fragments (`Q`, `2`, `R`, …) are removed.
- `TestCharsToBoxes_DropsWatermarkBeforeGrouping` only asserts the watermark is absent; it does not assert that legitimate CJK content (`姓名`) survives. Add a positive assertion.
- **No false-positive test.** Any legitimate letter+digit token that appears ≥3× on a page (repeated SKU/batch codes, part numbers across line items, numbered section labels) would be silently deleted → content-loss regression. The filter runs unconditionally on every page of every PDF, so the false-positive risk is systemic across all document types, not scoped to resume/template PDFs.

### On the existing comments
- @JinHai-CN's `gofmt` comment: valid, already fixed by the follow-up commit. ✅
- CodeRabbit's "group related component glyphs into complete candidate strings instead of evaluating each `TextChar.Text` independently": valid and central — same as issue #2 above. Should be addressed (ideally by moving detection to the section level per #18145 fix #1 rather than reassembling glyphs at the char stage).
- CodeRabbit's "assert a box retains legitimate `姓名` text": valid, add it.
- CodeRabbit's "use raw token, accept only `[A-Za-z0-9_]`": partially valid / lower priority. Tightening to `[A-Za-z0-9_]` reduces the false-positive surface, but it is secondary to the granularity/path problems.

### Recommendation
Do not merge as-is. The fix is bypassed for the target document and cannot catch the dominant per-glyph watermark. Move the detection to `pdf_postprocess.go` at the section/document level using cross-document verbatim-repetition as the signal, and add false-positive + replay tests with a real resume dump.

(Note: reviewed read-only; no checkout/write to the repo.)

Taranum01:
Re-architecting the fix per @xugangqiang's review. New commit `7e88f94` supersedes the previous `82afd3a` + `7fc9730` gofmt fix.\n\n**Structural changes**\n\n1. **Wrong code path** — removed. The previous `filterWatermarkChars` lived inside `CharsToBoxes`, but the resume PDF in #18145 takes the `ocrMergeChars` branch (parser.go:347 only fires on `!ocrUsed && len(chars) > 0`, which is the false path here). The char-stage filter was dead code.\n\n2. **Wrong granularity** — fixed. The watermark from #18145 is a per-glyph single-character box stream (Q, 2, R, h, N, U...), but `looksLikeWatermarkCandidate` required `len(text) >= 2` AND letter+digit mixing, so the dominant noise was structurally unreachable. The new helper is single-ASCII-byte scoped.\n\n3. **Wrong layer** — moved to `FilterWatermarkBoxes` in `internal/deepdoc/parser/pdf/layout/layout.go`, plugged into the parser.go dedup chain *after* `DedupSubstringOverlaps` and *before* `TextMerge`. This sits in the post-process pipeline that all OCR + char paths converge to, so the dead-branch bug above is gone.\n\n**Scope tightening** (the false-positive concern):\n\n- Only boxes whose `Text` is exactly one ASCII letter or digit (`len(text) == 1` bytewise). 6-char SKU / batch numbers (`ABC123`) and 2-character tokens fall out automatically.\n- Promoted only when the same (page, text) appears ≥ 4 times on that page. A real `Q` or `1` (list markers, table column labels) tops out around 1-2/page outside a watermark tiling.\n- Page-local: a watermark on page 3 cannot affect page 1, and the algorithm stays O(n) in page size.\n- CJK is excluded by the byte-length test (`姓` is 3 UTF-8 bytes). `姓名`, the canonical positive case the review called out, survives verbatim.\n- Repeated SKUs (`ABC123` × 8), repeated numbers in tables, repeated punctuation, and ` 空格`-prefixed lines all survive.\n\n**Test rewrite** — the three old tests modeled the watermark as `Text="ce1a60"` (whole token), which is what the review meant by "pass but do not reproduce the bug". New tests reproduce the actual shape:\n\n- `TestFilterWatermarkBoxes_DropsRepeatedSingleChars` — 5× Q + 5× 1 on one page, both gone.\n- `TestFilterWatermarkBoxes_KeepsLegitimateShortContent` — single-occurrence Q / 1 / A all survive (legitimate list markers).\n- `TestFilterWatermarkBoxes_KeepsCJK` — 6× 姓 all survive (positive assertion, addresses the review's `assert a box retains legitimate 姓名 text` point).\n- `TestFilterWatermarkBoxes_KeepsRepeatedMultiCharTokens` — 8× ABC123 all survive (the false-positive example the review demanded).\n- `TestFilterWatermarkBoxes_PerPageScopeIsolatesByPageNumber` — per-page scope boundary.\n- `TestFilterWatermarkBoxes_HandlesEmptyAndNil`, `TestFilterWatermarkBoxes_NoFalsePositiveOnMixed`, `TestIsSingleAsciiAlnum` for the bounds.\n\n`go test ./internal/deepdoc/parser/pdf/layout/...` is green; pre-existing pdfium/pdfoxide CGO link errors in the wider pdf/ target are unrelated to this change.\n\nI have not done a parity replay against a real resume dump — the reviewer's note ("Add a parity/replay test using a real resume charspy dump asserting the per-glyph fragments (Q, 2, R, …) are removed") would need fixture data that lives in a separate path; flag if that fixture already exists elsewhere in the repo and I'll wire it up.

Taranum01:
Pushed `35478244` addressing the CodeRabbit nitpick.\n\n**Bug being fixed**: `FilterWatermarkBoxes` called `strings.TrimSpace(b.Text)` BEFORE classifying with `isSingleAsciiAlnum`. That meant a box whose raw text was `' Q '` (3 bytes) collapsed into the same `(page, 'Q')` promotion key as a box whose raw text was `'Q'` (1 byte). Concretely: four `' Q '` boxes on a page would now count as four `'Q'` slots, hit the per-page threshold, and get dropped — even though their raw contents were four distinct characters of legitimate text. The contrapositive (four raw `'Q'` boxes) is a real watermark, so the two should diverge; `TrimSpace` erased that distinction.\n\n**Fix**: use `b.Text` verbatim for both the candidate-classification and the promotion-key lookup. Real box text is built by `LineToTextBox`, which already excludes pure-whitespace chars from `textParts`, so the defensive trim was protecting against an upstream shape that does not occur.\n\n**New regression test**: `TestFilterWatermarkBoxes_WhitespacePaddedNotCollapsedWithBareSingleChar`. Layout: 4 raw `'Q'` boxes (cross the `watermarkBoxesMinOccurrences` threshold and get dropped, proving the filter is alive on real watermark-shaped input) + 3 padded boxes (`' Q '`, `' Q'`, `'Q '`) that must be untouched because they never entered the candidate set under the no-TrimSpace contract. Asserts `rawQ == 0` AND `padded == 3` — without the fix, the 3 padded boxes would have collapsed with the 4 raw slots into a single `(page, 'Q')` promotion key and been wrongly dropped alongside them.\n\nAll 9 watermark tests + the rest of the layout suite pass; `gofmt` clean.

xugangqiang:
LGTM @yuzhichang

PR #18312, opened 2026-08-15
docs: fix dead internal links in FAQ and HTTP API reference

yzxcj797:
Two dead internal links found by auditing every docs link against the current file tree:

1. **docs/faq.mdx (3 occurrences)**: points to `./guides/run_health_check.md`, which was removed in #12363 because the system web page is disabled since v0.22.0 and health checking is now covered by the API. The links now point to the "Check system health" section of the HTTP API reference (`/api/v1/system/healthz`).

2. **docs/references/http_api_reference.md**: the custom-parameter note links to `../guides/agent/agent_component_reference/begin.md`, but that directory was replaced by `agent_workflow/`. Updated to `../guides/agent/agent_workflow/basic_component.md`, which documents the Begin component.


Yannnnnnny (approved):
The updates are correct，approved!

PR #14799, opened 2026-05-11
feat: add tag management for Agents with filtering and sorting (#14774)

plind-junior:
## Summary
Closes #14774.

Adds free-form tags on agents (UserCanvas) with full UI + API:
- Stored as comma-separated `tags` column on `UserCanvas` with online migration.
- New endpoints: `GET /v1/agents/tags` (aggregate counts) and `PUT /v1/agent/<id>/tags` (write). `GET /v1/agents` accepts a `tags=` query.
- "Edit tags" item in agent dropdown opens a chip-style editor dialog; tags render as badges on each agent card.
- New "Tags" facet in the agents filter bar, with counts.

## Implementation notes
- **Tag matching is exact-token**: the SQL filter wraps stored tags as `,…,` and matches `,ml,` so `ml` doesn't match `ml-ops`.
- **Server-side normalization** in `UserCanvasService.update_tags`: dedup (case-insensitive), per-tag cap of 64 chars, total length capped at 512 chars to fit the column, commas inside tag values are replaced with spaces.
- **Tenant authorization**: `PUT /v1/agent/<id>/tags` gates on `UserCanvasService.accessible(canvas_id, tenant_id)`.
- **Tag listing scope**: `UserCanvasService.list_tags` follows the same own + team-shared rule as `get_by_tenant_ids`.
- **i18n**: keys added to `en.ts` and `zh.ts` only (per project convention; other locales fall back).
- **`HomeCard`** gets a non-breaking `extra?: ReactNode` slot for the chip row; no `src/components/ui/` files modified.

## Test plan
- [ ] Backend boot runs `migrate_db` → confirm `user_canvas.tags` column exists (`DESCRIBE user_canvas`).
- [ ] Agents page renders cards normally (no console error from missing field).
- [ ] `⋯ → Edit tags` opens a dialog that stays open (regression: dialog was unmounting with the dropdown).
- [ ] Typing a tag without pressing Enter and clicking Save persists it (regression: last typed tag was being dropped).
- [ ] Chip input supports Enter/comma to commit, Backspace on empty to remove, `×` to remove individual chip.
- [ ] Tag containing a comma sent via API is stored with the comma replaced by a space.
- [ ] 20 long tags sent via API does not error (length cap silently truncates).
- [ ] "Tags" filter in the filter bar shows counts and narrows the list.
- [ ] Filtering by `ml` does **not** return agents tagged `ml-ops`.
[...]

Issue #14774, opened 2026-05-11
[Feature Request] Add tag management for Agents with filtering and sorting support


caesergattuso:
## Feature Description

Currently, RAGFlow Agents do not support tagging or labeling functionality. As users create more Agents, managing and organizing them becomes increasingly difficult. Tag-based management would allow users to categorize and filter Agents efficiently.

## Expected Behavior

- Users can assign one or more tags to an Agent (e.g., "customer-support", "internal", "finance", "experimental")
- The Agent list view should support filtering by tag
- Agents can be sorted by tag, or grouped by tag in the UI
- Tags should be editable and removable from the Agent settings
- Tag metadata should be searchable across the Agent list

## Use Case

- Teams with many Agents (20+) need a way to organize them by function, team, or purpose
- Quick filtering to find Agents relevant to a specific workflow or project
- Batch operations on Agents sharing the same tag (e.g., bulk enable/disable)
- Simplifies Agent governance in large-scale deployments

## Additional Context

This is a common organizational feature in most agent or bot management platforms. Tag support would significantly improve usability for power users and enterprise deployments.

PR #18313, opened 2026-08-15
fix(agent): make tag metadata searchable and sortable in agent list

peewee92:
## Summary

PR #14799 implemented the core tag management feature for Agents (DB column, API endpoints, UI editor, filter bar) but left two gaps from the issue #14774 requirements:

1. **Tag metadata not searchable**: The keyword search in the agent list only matched `title`, not `tags`. Searching for `"finance"` would miss an agent titled `"Budget Report"` that was tagged `"finance"`.

2. **Agents not sortable by tag (Go)**: The `userCanvasOrderableColumns` whitelist omitted `"tags"`, so `orderby=tags` silently fell back to `create_time`.

## Changes

| File | Language | Change |
|---|---|---|
| `api/db/services/canvas_service.py` | Python | OR `tags` column into keyword `WHERE` clause alongside `title` |
| `internal/dao/user_canvas.go` | Go | Add `"tags"` to `userCanvasOrderableColumns`; OR `user_canvas.tags` into keyword `LIKE` clause |
| `internal/dao/user_canvas_test.go` | Go test | `TestUserCanvasDAOKeywordSearchIncludesTags` + `TestUserCanvasDAOOrderByTags` |

## Verification

- Python lint (`ruff check`): no new errors introduced (pre-existing issues unchanged).
- Go tests follow the existing `user_canvas_test.go` pattern (in-memory SQLite, `pushDB`, `ListByTenantIDs`).
- Go is not installed in the current environment; CI will run `build.sh --test ./internal/dao/...`.

Contributes to #14774

### DIFF
--- a/api/db/services/canvas_service.py
+++ b/api/db/services/canvas_service.py
@@ -156,7 +156,7 @@ class UserCanvasService(CommonService):
                 .join(User, on=(cls.model.user_id == User.id))
                 .where(
                     owner_filter,
-                    (fn.LOWER(cls.model.title).contains(keywords.lower())),
+                    (fn.LOWER(cls.model.title).contains(keywords.lower()) | fn.LOWER(cls.model.tags).contains(keywords.lower())),
                 )
             )
         else:
--- a/internal/dao/user_canvas.go
+++ b/internal/dao/user_canvas.go
@@ -42,6 +42,7 @@ var userCanvasOrderableColumns = map[string]struct{}{
 	"permission":      {},
 	"canvas_type":     {},
 	"canvas_category": {},
+	"tags":            {},
 	"create_time":     {},
 	"create_date":     {},
 	"update_time":     {},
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Good change with coverage for both the search and sort gaps. Notes:

1. **Python precedence**: `fn.LOWER(cls.model.title).contains(...) | fn.LOWER(cls.model.tags).contains(...)` parses as `(title.contains(...)) | (tags.contains(...))` because method calls bind tighter than `|`, so the OR is correct. Adding explicit parentheses would improve readability since it reads slightly ambiguously at a glance.

2. **Go**: `user_canvas.tags` matches the `gorm:"column:tags"` mapping and the existing `applyUserCanvasTagFilter` REGEXP query. Adding `"tags"` to `userCanvasOrderableColumns` keeps ordering behind the closed allowlist, so no injection surface is introduced. The keyword `LIKE` on `tags` is consistent with the existing title search.

3. **Tests**: `TestUserCanvasDAOKeywordSearchIncludesTags` and `TestUserCanvasDAOOrderByTags` follow the existing `setupUserCanvasTestDB`/`pushDB` pattern and cover tag-only match and ascending tag sort. Since Go isn't available locally, please make sure `build.sh --test ./internal/dao/...` passes in CI.

LGTM otherwise. Approving.

Issue #18174, opened 2026-08-12
[Bug]: Synchronous Infinity insert in async add_chunk can block the entire Python API service


Glen93:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

75363af54

### RAGFlow image version

v0.26.4-961-g75363af54 (infiniflow/ragflow:latest)

### Other environment information

```Markdown
- Deployment: Docker Compose
- RAGFlow image: infiniflow/ragflow:latest (linux/amd64, CPU)
- RAGFlow version in container: v0.26.4-961-g75363af54
- Image digest: sha256:3eb33dc843e799397da11ef46b744e7e0d1eddef1611119303113ea4dc8c712d
- DOC_ENGINE: Infinity
- Infinity image: infiniflow/infinity:v0.7.2-x64-v3
- RAGFlow API process: a single python3 api/ragflow_server.py process behind nginx
```

### Actual behavior

When adding chunks through the Python REST API, one request can hang inside the synchronous Infinity insert and make the entire RAGFlow API unavailable.

In the observed run, an application sequentially called:

`POST /api/v1/datasets/{dataset_id}/documents/{document_id}/chunks`

The first 16 requests returned HTTP 200. The 17th request never returned. While it was stuck:

- the nginx-served static homepage still returned HTTP 200;
- `/api/v1/system/status` and other API endpoints timed out;
- the `ragflow-cpu` container used only about 1.6% CPU;
- Infinity remained healthy and its database/table HTTP endpoints responded in a few milliseconds;
- Infinity logs showed that the request had reached the insert path (including `AppendMemIndex` activity);
- restarting only the RAGFlow container restored the API.

The current `add_chunk` route is async, but it executes both embedding and document-store insertion synchronously:

```python
async def add_chunk(...):
    ...
    v, c = embd_mdl.encode(...)
    ...
    settings.docStoreConn.insert(...)
```

[...]

PR #18319, opened 2026-08-16
fix(api): isolate blocking operations in update_chunk with timeout

peewee92:
## Summary

PR #18187 fixed `add_chunk` but the issue reporter ([commented](https://github.com/infiniflow/ragflow/pull/18187#issuecomment-5274691345)) that `update_chunk` has the **same blocking pattern**: `embd_mdl.encode()` and `settings.docStoreConn.update()` run synchronously inside the async Quart route. When the Infinity update hangs, it blocks the entire event loop and all API endpoints become unresponsive.

The PR author [confirmed](https://github.com/infiniflow/ragflow/pull/18187#issuecomment-5275057218) they would address `update_chunk` in a separate follow-up PR. This is that follow-up.

## Changes

**`api/apps/restful_apis/chunk_api.py`**:
- Wrap `embd_mdl.encode()` + `settings.docStoreConn.update()` in `asyncio.to_thread()` so they run in a worker thread, freeing the Quart event loop.
- Add a 60s bounded timeout via `asyncio.wait_for()`. On timeout: return a clear error response and log it. The worker thread continues in the background (Python cannot force-cancel a running sync callable — the important behavior is that the event loop is released).
- Add `UPDATE_CHUNK_TIMEOUT_SECONDS` module constant so tests can monkeypatch the timeout.

**`test/testcases/test_web_api/test_chunk_app/test_chunk_routes_unit.py`**:
- `test_restful_update_chunk_timeout_returns_error`: slow `docStoreConn.update` + short timeout → error response with "timed out".
- `test_restful_update_chunk_does_not_block_event_loop`: while the worker thread sleeps, a marker coroutine advances on the event loop.
- Fix two missing test stubs (`validate_dataset_embedding_models`, `pagination_utils` constants) that prevented the test module from loading.

## Verification

All 3 tests pass (existing `update_chunk` test + 2 new tests):
```
PASS: existing update_chunk test
PASS: timeout test
PASS: responsiveness test
All tests passed!
```
`ruff check` and `ruff format` clean on both files.

## Design notes

- Uses `asyncio.to_thread` (not `thread_pool_exec`) because `thread_pool_exec` creates a short-lived `ThreadPoolExecutor` whose `with` block calls `shutdown(wait=True)` on exit — if the timeout fires, that shutdown would block the event loop again on the stuck Infinity call. `asyncio.to_thread` uses the default executor with no per-call shutdown, so a timeout cleanly releases the event loop while the worker thread continues in the background.
- Follows the existing `provider_api_service.py` pattern (`asyncio.to_thread` + `asyncio.wait_for` for bounded-timeout offload).

Contributes to #18174


JinHai-CN:
@peewee92 would you please also check the comments from @coderabbitai

api/apps/restful_apis/chunk_api.py
```diff
@@ -1187,15 +1191,25 @@ async def update_chunk(tenant_id, dataset_id, document_id, chunk_id):
...
+                [
+                    doc.name,
+                    d["content_with_weight"] if not d.get("question_kwd") else "\n".join(d["question_kwd"]),
+                ]
+            )
+            v = 0.1 * v[0] + 0.9 * v[1] if doc.parser_id != ParserType.QA else v[1]
+            d[f"q_{len(v)}_vec"] = v.tolist()
+            settings.docStoreConn.update({"id": chunk_id}, d, search.index_name(dataset_tenant_id), dataset_id)
+
+        await asyncio.wait_for(asyncio.to_thread(_update_sync), timeout=UPDATE_CHUNK_TIMEOUT_SECONDS)
```
buua436, line 1207: `asyncio.wait_for(asyncio.to_thread(...))` does not cancel the timed-out worker, so a stale update may overwrite a newer one; please add an atomic per-chunk revision/CAS check and a regression test for this race.

buua436 (changes requested):

PR #18325, opened 2026-08-16
fix: fix: close file handles in json.load() calls in resume parser

Ricardo-M-L:
Fix identified in code review: fix: close file handles in json.load() calls in resume parser


JinHai-CN:
The issue isn't valid.

PR #18331, opened 2026-08-16
fix: fix: correct nested path traversal in set_variable_param_value

Ricardo-M-L:
Fix identified in code review: fix: correct nested path traversal in set_variable_param_value


JinHai-CN:
It seems the issue is invalid.

PR #18335, opened 2026-08-16
fix: Fix WebDriver resource leak in __get_pdf_from_html

Ricardo-M-L:
Fix identified in code review: Fix WebDriver resource leak in __get_pdf_from_html


JinHai-CN:
It seems the issue already was fixed, would you please double check the latest code?

PR #18341, opened 2026-08-16
fix(agent/tools): read Keenable result content from `snippet`

ilya-bogin-keenable:
### Summary

Follow-up bug fix to #16233 (the Keenable search tool, merged 2026-06-25). The tool returned **textless chunks**.

The Keenable API returns both `description` and `snippet` on every result. `description` is frequently empty and `snippet` carries the page text, but `get_content` read `description`:

```python
get_content=lambda r: r.get("description"),
```

Verified against the live keyless endpoint just now: **10/10 results came back with `len(description) == 0` and `len(snippet)` between 2000 and 2600**. So every chunk `_retrieve_chunks` built from a Keenable search had a title and a URL and no content at all.

It now reads `snippet` first with a `description` fallback. These chunks feed retrieval, so unlike other places this mapping appears the text is deliberately **not** truncated — only the newlines it arrives with are normalized, which matches what `searxng`'s `get_content` hands over.

Also adds `test/unit_test/agent/tools/test_keenable_unit.py`. It loads the module in isolation the way `test_exesql_ssrf.py` does (the package `__init__` auto-discovers every tool and pulls in the full agent framework) and applies the getters exactly as `_retrieve_chunks` applies them, since the getter is where the bug lived. 5 tests, no extra dependencies. The fixture is the point: a mock that populates `description` — which no real response does — is why this passed review the first time.

2 files, `ruff` clean under the repo config.

### DIFF
--- a/agent/tools/keenable.py
+++ b/agent/tools/keenable.py
@@ -56,6 +56,17 @@ def _request(method: str, public_path: str, keyed_path: str, api_key: str, *, pa
     return resp.json()
 
 
+def _result_content(result: dict) -> str:
+    """Pull a result's text out of the Keenable response.
+
+    Keenable returns both ``snippet`` and ``description``: ``snippet`` carries
+    the page text and ``description`` is frequently empty, so prefer whichever
+    has content. The chunks built from this feed retrieval, so the text is kept
+    whole; only the newlines it arrives with are normalized.
+    """
+    return " ".join(str(result.get("snippet") or result.get("description") or "").split())
+
+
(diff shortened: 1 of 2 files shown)


JinHai-CN:
@ilya-bogin-keenable Would you please also check and update Go implemetion? https://github.com/infiniflow/ragflow/pull/16884

JinHai-CN:
@ilya-bogin-keenable Could you please drop us a line at haijin.chn@gmail.com? We'll notify you when we cut a new release so you can test whether Keenable works as expected.

PR #18345, opened 2026-08-17
feat(web): pin the current user to the top of admin user list

dcc123456:
### Summary

feat(web): pin the current user to the top of admin user list

Admins most often act on their own account first, so surface it without scrolling or searching. Ordering the data rather than using row pinning keeps the row inside the existing search, filter, and pagination flow.

### DIFF
--- a/web/src/pages/admin/users.tsx
+++ b/web/src/pages/admin/users.tsx
@@ -449,8 +449,27 @@ function AdminUserManagement() {
     ],
   );
 
+  // Pin the current user to the top of the list
+  const orderedUsersList = useMemo(() => {
+    if (!usersList) {
+      return EMPTY_DATA;
+    }
+
+    const currentUserIndex = usersList.findIndex(
+      (user) => user.email === userInfo?.email,
+    );
+
+    if (currentUserIndex <= 0) {
+      return usersList;
+    }
+
+    const rest = usersList.filter((_, index) => index !== currentUserIndex);
+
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this clean, well-scoped change. I reviewed the diff and cross-checked it against the current `web/src/pages/admin/users.tsx`.

**What it does well**
- Correctly guards the `usersList === undefined` case by returning `EMPTY_DATA`, matching the previous `usersList ?? EMPTY_DATA` behavior.
- `findIndex` returning `-1` (current user not in the list) is handled correctly by the `currentUserIndex <= 0` branch, so there's no accidental pinning or duplication.
- Pinning is done by reordering the data source, so the current user still participates in search, filtering, and pagination (total count is unchanged) — consistent with the PR description.

1. Sorting overrides the pin: `getSortedRowModel` re-sorts the whole array, so once the user clicks a column header the current user moves back into their natural sorted position. Fine if the intent is "pin only in the default view", but worth confirming that's intentional.
2. `user.email === userInfo?.email` is case-sensitive. This is consistent with the existing `isMe` checks in this file, so no change needed — just flagging in case emails can ever differ in casing.
3. Minor style: `usersList.filter((_, index) => index !== currentUserIndex)` works, but something like slice+concat would read slightly more clearly. Optional.

No regressions, style, or security concerns spotted. LGTM overall.

PR #18351, opened 2026-08-17
Fix: enhance form submission error handling and update rerank form field prefix

dcc123456:
### Summary

Fix: enhance form submission error handling and update rerank form field prefix


wangq8 (commented):
Thanks for the PR. Reviewed the two changes in `web/src/pages/next-search/search-setting.tsx`:

1. **Rerank field prefix fix (`search_config` → `search_config.`)** — Correct. `RerankFormFields` builds its field name via `prefixName(prefix, name)` in `@/utils/form`, which does plain string concatenation (`prefix + name`) without inserting a dot. So `prefix='search_config'` produced `search_configrerank_id`, which does not match the schema/validation path (`search_config.rerank_id`, see `search-setting-hooks.ts`) and would have broken rerank model selection binding. The trailing dot resolves it. ✅

2. **`handleSubmit` error handler** — The second argument to `handleSubmit` is the invalid-submit callback, so the signature is correct. One minor nit: `console.error(error, formMethods.getValues())` logs the entire form values object to the console, which may include data you don't want in production logs. Consider logging only the `error` object (or dropping the `getValues()` payload).

No regressions or security concerns spotted. Small, focused change — looks good to me otherwise.

PR #18356, opened 2026-08-17
fix(ingestion): accumulate component progress lines into document.progress_msg

euvre:
## Problem

In the Go ingestion path, the document "details" (进度日志) dialog only showed the **last line** of the run log. Root cause: `progressSink.OnComponentProgress` mirrored only the latest component lifecycle message (`ev.Message`) into `document.progress_msg`, overwriting previous lines on every event. The frontend dialog (`ProcessLogModal`) renders `progress_msg` verbatim as a multi-line log, so only one line was ever visible.

## Fix

`internal/ingestion/service/progress_sink.go` now accumulates one line per component lifecycle event and mirrors the **full accumulated log** into `document.progress_msg`, matching Python semantics:

- Append `msg` with `\n` separator, trim from the head to 3000 chars (`progressLogMaxChars`), mirroring `TASK_MAX_LOG_LENGTH` / `trim_header_by_lines` in `api/db/services/task_service.py` (`TaskService.update_progress`).
- The in-memory buffer is lazily seeded from the document row on the first event, so a resumed run keeps lines logged before a crash/redelivery (`StartRunning` resets the row to `""` on a fresh run, which then seeds an empty log).
- Guarded by a mutex since eino fires progress events from concurrent parallel-branch goroutines.

## Tests

- 5 new unit tests in `progress_sink_test.go` covering append semantics, head-trim at 3000 chars (whole-line boundary), seeding from an existing `progress_msg`, empty message no-op, and concurrency safety.
- `bash build.sh --test ./internal/ingestion/...` green (service + pipeline packages).

## E2E verification

Reproduced on dataset `issue407_repro` with `progress-log-test.md`:
- MySQL: `document.progress_msg` contains the complete 10-line run log (File/Parser/TokenChunker/Extractor/Tokenizer Started+Done).
- UI: the document details dialog now renders all 10 lines with `whitespace-pre-line` (screenshot verified in browser).

### DIFF
--- a/internal/ingestion/service/progress_sink.go
+++ b/internal/ingestion/service/progress_sink.go
@@ -19,6 +19,8 @@ package service
 import (
 	"context"
 	"fmt"
+	"strings"
+	"sync"
 	"sync/atomic"
 
 	"ragflow/internal/common"
@@ -28,6 +30,11 @@ import (
 	servicepkg "ragflow/internal/service"
 	documentpkg "ragflow/internal/service/document"
 )
+
+// progressLogMaxChars bounds the accumulated run log mirrored into
+// document.progress_msg. Mirrors Python's TASK_MAX_LOG_LENGTH
+// (api/db/services/task_service.py).
+const progressLogMaxChars = 3000
 
 // progressSink implements pipeline.ProgressSink. It is the single writer of
 // the document / ingestion_task_log / ingestion_task.component_total tables
@@ -44,6 +51,12 @@ type progressSink struct {
 	// which eino fires from concurrent parallel-branch goroutines. Atomic because
 	// the two access paths share no other synchronization.
 	total atomic.Int64
+	// logMu guards the accumulated run log below: OnComponentProgress fires
+	// from concurrent parallel-branch goroutines, so both the lazy seed and
+	// the append must hold the same mutex.
+	logMu  sync.Mutex
+	seeded bool
+	log    strings.Builder
 }
 
 // docProgressSvc is the subset of *service.DocumentService the sink needs to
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Approved — well-scoped fix with solid tests and faithful parity with Python's `TaskService.update_progress` / `trim_header_by_lines`.

Verified:
- `progressSink` is constructed per document run (`defaultRunDocumentTask`), and `Pipeline.documentID` is fixed per run, so the in-memory `log`/`seeded` state is correctly scoped to a single document.
- Seeding from `document.progress_msg` is sound: `StartRunning` resets it to `""` only on the CREATED→RUNNING transition, so a redelivered/resumed run seeds the pre-crash lines while a fresh run seeds empty.
- The mutex correctly guards both the lazy seed and the append across eino's concurrent parallel-branch goroutines; the race test targets `accumulateLog` directly to avoid the test-DB serialization masking the race.
- `DocumentResponse.ProgressMsg` is `*string` and the nil guards are correct.

Two minor, non-blocking observations:
1. `len(text)` in Go counts bytes, not characters, whereas Python's `len(text)` counts code points. For non-ASCII component/error messages the Go limit of 3000 is effectively 3000 bytes (more conservative than Python's 3000 chars). Safe — trimming only ever cuts at a `\n` byte, so no rune is split — but the `progressLogMaxChars` name is slightly misleading. `utf8.RuneCountInString` would give exact parity if desired.
2. `progressLogMaxChars` is hardcoded to 3000 while Python's `TASK_MAX_LOG_LENGTH` is env-configurable. Fine for now, but consider keeping them in sync if operators tune that limit.

PR #18357, opened 2026-08-17
Fix: Switching the rerank model on the chat page fails to save the configuration.

cike8899:
### Summary

Fix: Switching the rerank model on the chat page fails to save the configuration.

### DIFF
--- a/web/src/components/rerank.tsx
+++ b/web/src/components/rerank.tsx
@@ -19,7 +19,6 @@ import { useTranslate } from '@/hooks/common-hooks';
 import { prefixName } from '@/utils/form';
 import { useFormContext } from 'react-hook-form';
 import { z } from 'zod';
-import { SliderInputFormField } from './slider-input-form-field';
 import {
   FormControl,
   FormField,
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Approved — root cause and fix verified.

The failure was caused by the frontend sending a stale `tenant_rerank_id` (spread from the existing chat `data`) together with the newly selected `rerank_id`. The backend `_normalize_model_pair` (api/apps/restful_apis/chat_api.py) rejects requests where `rerank_id` and `tenant_rerank_id` resolve to different models (`... must refer to the same model`), so the save failed.

The fix is correct:
- Dropping `/api/v1/chats` from `API_WHITELIST` stops `addTenantParams` from injecting `tenant_*` params on chat CRUD endpoints — the server resolves/normalizes `llm_id`/`rerank_id` itself.
- Omitting `tenant_llm_id`/`tenant_rerank_id` from the payload prevents the stale values in `data` from leaking through.
- Removing the manual `tenant_llm_id = llm_id` sync is safe: `_normalize_model_pair` fills `tenant_llm_id`/`tenant_rerank_id` from the name fields when only names are provided.

No regression on message sending: the completion endpoint is `/api/v1/chat/completions` (singular), which `/api/v1/chats` never matched. The `rerank.tsx` cleanup (dead `top_k` slider code + unused imports) is unrelated but harmless.

PR #18364, opened 2026-08-17
Refactor: do not include compliation chunks in retrieval testing

wangq8:
Retrieval Testing: do not include compilation result.
Search: include compilation result.

### DIFF
--- a/api/apps/restful_apis/dataset_api.py
+++ b/api/apps/restful_apis/dataset_api.py
@@ -515,7 +515,7 @@ async def search_datasets(tenant_id):
     POST /api/v1/datasets/search
     JSON body: {"dataset_ids": list[str] (required), "question": str (required), "doc_ids": list[str], "top_k": int, "page": int, "size": int,
                "similarity_threshold": float, "vector_similarity_weight": float, "use_kg": bool,
-               "cross_languages": list[str], "keyword": bool, "meta_data_filter": dict}
+               "cross_languages": list[str], "keyword": bool, "meta_data_filter": dict, "include_knowledge_compilation": bool (default true)}
     Success: {"code": 0, "data": {"chunks": [...], "total": int, "labels": [...]}}
     Errors: ARGUMENT_ERROR (101) for invalid payload; DATA_ERROR (102) for access denied or internal errors.
     """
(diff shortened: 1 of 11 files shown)


wangq8 (commented):
## Summary

Overall this looks good and well-targeted. The approach — add `include_knowledge_compilation` (default `true`) and, when `false`, inject `must_not: {"exists": "compile_kwd"}` — reuses the already-established `must_not exists compile_kwd` pattern (e.g. `chunk_api.py`), and wires it consistently through the Python service, the Go service (`SearchDatasets` → `RetrievalRequest.Filter` → ES `buildBoolQueryFromCondition`), the frontend (`useTestRetrieval`), and tests. Defaulting to `true` keeps backward compatibility.

A few points worth checking before merge:

1. **Infinity engine `must_not exists` semantics (possible correctness gap).** The Infinity chunk-search path (`equivalentConditionToStr`, `internal/engine/infinity/chunk.go`) renders `must_not: {"exists": "compile_kwd"}` as `NOT (compile_kwd!='')`. Under SQL three-valued logic, `NULL != ''` is `NULL`, so `NOT NULL` is also `NULL` and the row is filtered out. This only behaves correctly if regular (non-compiled) chunks store `compile_kwd` as the empty string `''` rather than `NULL`. Please confirm the field default for the Infinity backend; otherwise non-compiled chunks could be wrongly excluded on Infinity-backed deployments. (This is pre-existing Infinity behavior, but this PR is the first to rely on it for the retrieval/search path.)

2. **`must_not` as a reserved key in `buildBoolQueryFromCondition` is a bit fragile.** It now treats the literal key `must_not` specially and only accepts the `{"exists": "<string>"}` shape, silently dropping anything else (e.g. `{"exists": {"field": ...}}` or nested bool). That's fine for the single internal producer, but it overloads the generic filter map and would mis-handle any real field named `must_not`. Worth a short comment in the function documenting the reserved key, or accepting the same shape the existing delete-path handler accepts.

3. **Minor / non-blocking:** `search()` in `api/apps/services/dataset_api_service.py` (~line 999) still doesn't apply the `must_not` filter, but it appears to be dead code — the `/datasets/<id>/search` route calls `search_datasets`, which does apply it. No functional gap today, but if that function is ever revived it will need the same handling.

4. **Naming nit (optional):** Go struct field `IncludeCompiledChunks` vs the JSON/API name `include_knowledge_compilation`. Fine, just slightly divergent terminology.

Test coverage added for handler parsing, service conversion, and retrieval filter propagation is good. I could not run a full Go build/test in this environment because an unrelated third-party CGO dependency (`office_oxide.h`) is missing from the toolchain.

PR #18366, opened 2026-08-17
fix: align Go ingestion progress and pipeline selection

buua436:

Fixes Go ingestion progress reporting and pipeline selection:

- Add timestamps to document progress logs.
- Keep document duration and status updated during parsing.
- Start frontend polling immediately after parsing begins.
- Prevent documents explicitly using General from inheriting an old dataset pipeline.
- Populate missing pipeline operation log fields.
- Remove stale component progress logs between retries.
- Prevent progress values greater than `1`.

### Type of change

### DIFF
--- a/internal/dao/ingestion.go
+++ b/internal/dao/ingestion.go
@@ -310,3 +310,23 @@ func (dao *IngestionTaskLogDAO) DeleteByTaskID(ctx context.Context, db *gorm.DB,
 	result := db.WithContext(ctx).Unscoped().Where("task_id = ?", taskID).Delete(&entity.IngestionTaskLog{})
 	return result.RowsAffected, result.Error
 }
+
+// DeleteComponentLogsByTaskID removes component lifecycle rows from a new run
+// while preserving checkpoint rows such as run_count for task history.
+func (dao *IngestionTaskLogDAO) DeleteComponentLogsByTaskID(ctx context.Context, db *gorm.DB, taskID string) (int64, error) {
+	var logs []*entity.IngestionTaskLog
+	if err := db.WithContext(ctx).Where("task_id = ?", taskID).Find(&logs).Error; err != nil {
+		return 0, err
+	}
+	ids := make([]int, 0, len(logs))
+	for _, log := range logs {
+		if log != nil && len(log.Checkpoint) == 0 {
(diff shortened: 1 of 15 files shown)


wangq8 (commented):
Thanks for this fix — the concurrency and correctness improvements are well done and the tests are solid. I verified the pipeline-selection change against `internal/ingestion/service/ingestion_service.go`, where an empty `PipelineID` already falls back to `parser_id` for builtin pipelines, so dropping the KB fallback correctly prevents "General" documents from inheriting an old dataset pipeline. The `trimLogHead` off-by-one fix, the progress clamp (`progress <= 1`), the timestamped log lines, and serializing log accumulation with the document write (via the shared `logMu`) all look correct.

A few things to consider before merge (non-blocking):

1. **Extra DB read on every progress event.** `UpdateRunProgress` now calls `documentDAO.GetByID` on each call to compute a live `process_duration`. `OnComponentProgress` fires per component/chunk, so this adds a full round-trip per event during parsing. If `ProcessBeginAt` is already available on the task context (or can be read once at sink construction), you could avoid the per-event read. Worth a quick perf check on large documents.

2. **`DeleteComponentLogsByTaskID` loads all rows into memory.** It `Find`s every log for the task and then filters in Go before deleting. For a single task the row count is bounded (one per component), so this is fine in practice, but it could be a single `DELETE ... WHERE task_id = ? AND (checkpoint IS NULL OR checkpoint = '')`-style statement if the checkpoint-empty condition can be expressed in SQL. Minor.

3. **`Status: &statusValue` default `"1"`.** When `doc.Status` is empty you default the new `Status` field to `"1"`. Please confirm `"1"` is the intended "valid" semantic (matching the size:1 soft-valid flag) — defaulting to a hard-coded valid value could mask a missing status.

4. **`resolvePipelineID` behavior change.** This is intentional per the PR description, but it's a semantic change: a document with no explicit `pipeline_id` and a non-"general" builtin `parser_id` will no longer inherit the KB pipeline. Just flagging so it's clearly an intended change (and ideally covered by a test that asserts a builtin-parser doc no longer inherits KB pipeline).

Nothing here blocks; nice work.

buua436:
image

PR #18367, opened 2026-08-17
Fix: The begin operator of the agent created from the template does not display a pattern.

cike8899:
### Summary

Fix: The begin operator of the agent created from the template does not display a pattern.


wangq8 (commented):
Verified the fix. The begin node is rendered by `web/src/pages/agent/canvas/node/begin-node.tsx`, which reads the mode badge from `data.form?.mode`. When `data.form` is absent, `mode` is `undefined` and no pattern is displayed — which matches the reported bug.

Adding `data.form: { enablePrologue: true, mode: "conversational" }` to the `begin` node fixes it, and is consistent with the other templates (e.g. `trip_planner.json`, `web_search_assistant.json`). The payload is valid against `BeginFormSchema` (`mode` is required and provided; `enablePrologue`/`prologue` are optional).

Two minor nits (non-blocking):

1. `deep_research.json` lost its trailing newline (`\ No newline at end of file`) and was reformatted alongside the fix — fine, but it makes the diff larger than necessary.
2. `enablePrologue: true` is set without a `prologue` string, so `useGetBeginNodePrologue` will return `undefined` (no prologue message). Harmless, but consider either adding a `prologue` value or dropping `enablePrologue` for clarity.

PR #18368, opened 2026-08-17
Feat: add qwen-3.8 series models.

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/conf/llm_factories.json
+++ b/conf/llm_factories.json
@@ -542,6 +542,22 @@
             "url": "https://dashscope.aliyuncs.com/compatible-mode/v1",
             "llm": [
                 {
+                    "llm_name": "qwen3.8-max",
+                    "max_tokens": 128000,
+                    "model_type": [
+                        "chat", "image2text"
+                    ],
+                    "is_tools": true
+                },
+                {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for adding these. The model names are valid (qwen3.8-max and qwen3.8-2.4t-a95b, released Aug 2026), and `model_type` is correct — qwen3.8-max is multimodal (`chat` + `image2text`) while qwen3.8-2.4t-a95b is text-only (`chat`).

One issue to double-check: `max_tokens: 128000` for both entries looks under-specified.

- `qwen3.8-max` has a 1M-token context window by default. The sibling entry `qwen3.7-max` in this same file uses `max_tokens: 1000000`. Suggest aligning `qwen3.8-max` to `1000000`.
- `qwen3.8-2.4t-a95b` natively supports 262,144 tokens of context (extensible to ~1M). Suggest at least `262144` rather than `128000`.

Minor: please confirm the DashScope `compatible-mode` endpoint actually serves `qwen3.8-2.4t-a95b` under this exact model id (the open-weight release is only a few days old), otherwise users will hit 404s.

PR #18370, opened 2026-08-17
Fix: get memory size cache rightly

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/api/db/joint_services/memory_message_service.py
+++ b/api/db/joint_services/memory_message_service.py
@@ -229,7 +229,7 @@ async def embed_and_save(memory, message_list: list[dict], task_id: str = None):
                 return False, error_msg
 
         new_msg_size = sum([MessageService.calculate_message_size(m) for m in message_list])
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Verified against the signature `get_memory_size_cache(memory_id: str, uid: str)` in `api/db/joint_services/memory_message_service.py`.

The function builds the Redis key from the **first** argument (`memory_{memory_id}`) and passes `[memory_id], [uid]` to `MessageService.calculate_memory_size`. The old call `get_memory_size_cache(memory.tenant_id, memory.id)` therefore:
- used the wrong Redis key (`memory_{tenant_id}` instead of `memory_{memory_id}`), and
- swapped the `calculate_memory_size` arguments.

The new order `get_memory_size_cache(memory.id, memory.tenant_id)` matches the signature and is consistent with the other call sites (`init_memory_size_cache` and line 341). This is a correct, minimal fix. LGTM.

PR #18371, opened 2026-08-17
fix(web): exclude <think> reasoning blocks from copied answer

dcc123456:
Copying an assistant message handed over the raw content, so the model's reasoning block came along with the answer. Strip it via a shared removeThinkSection helper, which also tolerates an unclosed block while the message is still streaming.

### Summary

### DIFF
--- a/web/src/components/message-item/group-button.tsx
+++ b/web/src/components/message-item/group-button.tsx
@@ -24,6 +24,7 @@ import {
 import { useSetModalState } from '@/hooks/common-hooks';
 import { IRemoveMessageById } from '@/hooks/logic-hooks';
 import { cn } from '@/lib/utils';
+import { removeThinkSection } from '@/utils/chat';
 import {
   LucidePauseCircle,
   LucideRefreshCw,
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for the PR! This correctly fixes the copy-to-clipboard action leaking `<think>` reasoning blocks. Sharing the `removeThinkSection` helper and handling the unclosed (streaming) block are both good touches.

A few suggestions:

1. **Add unit tests.** `web/src/utils/chat.ts` already has coverage under `web/src/utils/tests/chat.test.ts` (and `web/src/utils/__tests__/chat.test.ts`). A pure utility like this is cheap to cover — please add cases for: a closed block, an unclosed (streaming) block, multiple blocks, and an answer that legitimately contains the literal text `<think>`.

2. **Literal `<think>` in answer text.** The second `.replace(/<think>[\s\S]*$/, '')` is greedy. If an answer genuinely discusses the `<think>` tag (e.g. a coding-help reply), everything from that literal tag onward would be stripped from the copied text. Since `replaceThinkToSection` already makes the same assumption this is probably acceptable, but it's worth being aware of.

3. **`.trim()` is a subtle behavior change.** Previously the copy preserved `content` verbatim; now leading/trailing whitespace is trimmed. Likely desirable, but worth a conscious decision.

4. **Whitespace artifacts.** Removing a block leaves its surrounding blank lines, so copied text can contain awkward doubled newlines. Consider collapsing them (cosmetic/optional).

5. **Same bug elsewhere?** The floating chat widget (`web/src/components/floating-chat-widget.tsx`) also renders `<CopyToClipboard text={message.content} />` and would leak the same `<think>` blocks. Worth a follow-up fix, or a note if it's out of scope here.

Nothing blocking — looks good once tests are added.

PR #18375, opened 2026-08-17
WIP Fix(ingestor): dedup memory-task redelivery to prevent duplicate extraction

xugangqiang:
## Problem

Memory-extraction tasks share the `tasks.RAGFLOW` NATS consumer and the ingestor worker pool with ingestion tasks, but they do **not** use the ingestion state machine (no `ingestion_task` row). `processMessage`'s memory branch therefore lacked the `claimTask` dedup guard that the ingestion branch has.

Under a NATS redelivery — an `AckWait` timeout while a slow LLM extraction is running, or an ingestor restart — a redelivered memory message was dispatched to the worker **again**, because nothing rejected the second transition. This double-executed the extraction:

- a second `extractByLLM` call (extra LLM token cost), and
- a second `embedAndSaveMessages` → `InsertChunks`, producing **duplicate extracted memory entries** in the store.

The blocking send introduced by the backpressure fix (PR #18369) made this race reachable: a redelivery arriving while the first run is still in flight has no guard and is scheduled a second time.

## Correct behavior

A redelivered memory task that is already being processed must be `ack`-skipped, never re-executed. The in-flight claim must be released when the worker settles the task (so a later genuine redelivery can retry) and on shutdown (so the broker redelivers after restart).

## Fix

In `processMessage`'s memory branch:

- take a `claimTask(taskID)` guard before enqueuing; on conflict `ack`-skip the duplicate;
- release the claim in the `ctx.Done()` (shutdown) path so the broker redelivers the task after restart instead of blocking the consume loop;
- release the claim in `executeMemoryTask` via `defer e.releaseTask(taskID)` once the worker settles, mirroring `executeTask`.

This makes the memory branch symmetric with the ingestion branch's dedup handling.

## Tests

- `TestProcessMessage_DuplicateMemoryDeliveryAckSkips`: delivers the same memory `TaskID` twice (worker pool not started, so the claim is held in the channel) and asserts the second delivery is `ack`-skipped with no second enqueue.
- `TestProcessMessage_DuplicateIngestionDeliveryAckSkips`: asserts the same guarantee for the ingestion branch (injected duplicate delivery → `ack`-skip, exactly one execution, exactly one `COMPLETED` row).

## Dependency

Stacks on #18369 (ingestion backpressure fix), which converted the memory branch to a blocking send and established the in-flight `claimTask`/`releaseTask` pattern this PR reuses.

Note: until #18369 is merged, this PR's diff against `main` also includes #18369's commit. After #18369 lands, the diff shrinks to just this memory commit. Merge #18369 first.


xugangqiang:
Closing as obsolete: this memory-task redelivery dedup fix has already landed on main. The  guard + shutdown/worker release in the memory branch () plus  already cover the exact behavior described here. The branch was also stale (based on an old base, DIRTY against main). Reopen if a gap is found.

xugangqiang:
Closing as obsolete: this memory-task redelivery dedup fix has already landed on main. The claimTask guard plus shutdown/worker release in the memory branch (internal/ingestion/service/ingestion_service.go) and the test TestExecuteMemoryTask_ReleasesClaim already cover the exact behavior described here. The branch was also stale (based on an old base, DIRTY against main). Reopen if a gap is found.

PR #18376, opened 2026-08-17
fix: restore document query cache matching

buua436:

Fixes document query cache matching after the query key factory refactor. Prefix operations previously generated keys with trailing `undefined` values, causing `invalidateQueries` and optimistic `setQueriesData` updates to miss parameterized document list and filter queries.

This PR introduces explicit prefix keys and adds regression tests using TanStack Query’s actual partial matching behavior.

### Type of change

### DIFF
--- a/web/src/hooks/__tests__/use-document-request-query-keys.test.ts
+++ b/web/src/hooks/__tests__/use-document-request-query-keys.test.ts
@@ -1,0 +1,36 @@
+/*
+ *  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import { partialMatchKey } from '@tanstack/query-core';
+import { DocumentKeys } from '../document-query-keys';
+
+describe('DocumentKeys', () => {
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
## Review

Thanks for the fix — this correctly addresses the cache-matching regression. I verified the behavior against TanStack Query v5's actual `partialMatchKey` semantics.

### What's correct

- The root cause is real: `DocumentKeys.list()` with no args produced `['fetchDocumentList', undefined, undefined, undefined]`, which never partial-matched the parameterized list key `['fetchDocumentList', searchString, pagination, filter]`. Same for `filter()`.
- Introducing explicit prefix keys (`all()` → `['fetchDocumentList']`, `allFilters()` → `['fetchDocumentFilter']`) is the idiomatic fix.
- The tests use `partialMatchKey(actual, prefix)` with the correct argument order and assert the real matching behavior. Good.

### Things to consider (non-blocking)

1. **`partialMatchKey` is an internal API from a transitive dependency.** The test imports it directly from `@tanstack/query-core`, which is not declared in `web/package.json` (it is a transitive dep of `@tanstack/react-query`). `partialMatchKey` is an internal utility, not part of the public API surface. This is fragile — it can break on a minor bump or under stricter resolvers (pnpm/yarn PnP). Consider either declaring `@tanstack/query-core` explicitly in devDependencies, or (better) asserting through the real `QueryClient` / `invalidateQueries` behavior instead of reaching into internals.

2. **`all()` now also matches `byIds`.** `byIds` uses `['fetchDocumentList', 'byIds', ids]`, so `invalidateQueries({ queryKey: DocumentKeys.all() })` and `setQueriesData({ queryKey: DocumentKeys.all() })` now also target `byIds` queries (previously the `undefined`-suffixed key never matched them). This looks like a desirable side effect — e.g. the optimistic run-status updates in `useRunDocument` now also keep the `byIds` cache in sync — but please confirm it is intended.

3. **Minor consistency:** `use-manage-modal.ts:388` still hardcodes `queryKey: [DocumentApiAction.FetchDocumentList]` instead of `DocumentKeys.all()`. Optional follow-up to consolidate on the shared factory.

No regressions found in the diff itself; the re-export of `DocumentApiAction` keeps the existing import in `use-manage-modal.ts` working.

PR #18380, opened 2026-08-17
fix(web): hide verify button for OpenAI-API-Compatible provider

dcc123456:
### Summary

fix(web): hide verify button for OpenAI-API-Compatible provider

The OpenAI-API-Compatible provider has no provider-level credential to verify, so the verify action is meaningless there. Skip rendering the button in both draft and saved instance cards.

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
@@ -80,13 +80,15 @@ export function DraftModeCard({
         <AimlapiGetKeyButton onKey={handleAimlapiKey} />
       )}
 
-      <div className="pt-3">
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM.

Verified:
- `LLMFactory.OpenAiAPICompatible` resolves to `'OpenAI-API-Compatible'` in `web/src/constants/llm.ts`, so the comparison is correct.
- `providerName` is available as a prop in both `DraftModeCard` and `SavedModeCard`, and the new `LLMFactory` import is correctly added in `saved-mode-card.tsx` (already imported in `draft-mode-card.tsx`).
- The pattern is consistent with the existing `providerName === LLMFactory.AIMLAPI` check.

The provider has no provider-level credential to verify, so hiding the verify button in both draft and saved cards is the right call. Clean, scoped change.

PR #18382, opened 2026-08-17
fix: exclude compiled artifacts from chunk lists

buua436:

The Go document chunks API included Wiki pages, sections, and other compiled artifacts in ordinary chunk lists. This PR excludes chunks containing `compile_kwd`, aligning the Go behavior, pagination, and total count with the Python implementation.

### Type of change

### DIFF
--- a/internal/service/chunk/chunk.go
+++ b/internal/service/chunk/chunk.go
@@ -882,7 +882,8 @@ func (s *ChunkService) List(ctx context.Context, req *service.ListChunksRequest,
 			"available_int",
 		},
 		Filter: map[string]interface{}{
-			"doc_id": req.DocID,
+			"doc_id":   req.DocID,
+			"must_not": map[string]interface{}{"exists": "compile_kwd"},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM, thanks for the fix.

Verified the change against the Python implementation: `list_chunks` in `api/apps/restful_apis/chunk_api.py` uses the same `"must_not": {"exists": "compile_kwd"}` filter, so this correctly aligns Go behavior with Python.

Also confirmed the engine implementations handle the `must_not` + `exists` pair:
- elasticsearch/chunk.go builds a proper `bool.must_not` `exists` clause
- infinity handles `must_not exists` via `NOT (field != '')`
- oceanbase and serenedb both have explicit `must_not`/exists handling

One minor note (non-blocking): the Infinity path relies on `compile_kwd` having an empty-string default so that `NOT (compile_kwd != '')` matches ordinary chunks. This is consistent with existing patterns (e.g. `dataset_structure_graph.go`), so no change needed — just flagging the assumption for future reference.

Test coverage looks adequate for the filter construction.

PR #18321, opened 2026-08-16
fix(model_meta): Ollama._get_api_key handles JSON-dict api_key for the verify path

Harsh23Kashyap:
## Summary

5th site in the `rag/llm/model_meta.py` JSON-decode family — same shape as PR #18251 (cycle 19, VolcEngine / OpenRouter / NewAPI unification) and PR #18314 (cycle 22, LocalAI).

`rag/llm/model_meta.py:Ollama` was inheriting `Base._get_api_key`, which returns `self.api_key` verbatim. When the API service stored the api_key as a JSON dict (the format used for several other LLM providers — see `api/apps/services/provider_api_service.py:313` which calls `json.dumps(api_key)` for non-string values), `Ollama.get_model_list` sent an Authorization header of the form:

```
Authorization: Bearer {"api_key": "sk-ollama-xxx", "endpoint": "..."}
```

Ollama does not validate the bearer (it accepts malformed Bearer headers silently), so the user-facing symptom is mild compared to LocalAI's 401 behavior, but downstream setups that front Ollama with an authenticating reverse proxy would reject the malformed token. The fix is the same defensive resolution rule as cycle 22.

## Why this is a separate PR (not a follow-up to #18314)

PR #18314 (cycle 22, LocalAI) closed #17757. The 4 sites in the model_meta JSON-decode family are now: VolcEngine / OpenRouter / NewAPI (cycle 19 #18251) + LocalAI (cycle 22 #18314) + Ollama (this PR). The remaining 5 sites without `_get_api_key` overrides (`Xinference`, `HuggingFace`, `FunASR`, `AIMLAPI`, plus the `OpenAIAPICompatible` subclasses `NVIDIA` / `GreenPT` / `VLLM` / `LMStudio` / `RAGcon` which all share `Base._get_raw_model_list`) have the same missing-resolution pattern; future cycles can sweep them. The maintainer can merge this one independently of #18314 and #18251.

## Implementation summary

Two hunks in `rag/llm/model_meta.py:Ollama`:

1. **Add `_get_api_key` override** mirroring the `VolcEngine` / `OpenRouter` / `NewAPI` / `LocalAI` pattern:
   - JSON dict with an `api_key` field → the inner key
   - JSON dict without an `api_key` field → `""` (no-auth path; Ollama's normal case)
   - Plain string (the common case) → returned as-is
   - JSON parse error, JSON non-object, or non-string api_key → fall back to the Base default (raw passthrough) so the pre-fix behavior is preserved

[...]

PR #18314, opened 2026-08-15
fix(model_meta): LocalAI._get_api_key handles JSON-dict api_key for the verify path

Harsh23Kashyap:
## Summary

Closes #17757.

`rag/llm/model_meta.py:LocalAI` was inheriting `Base._get_api_key`, which returns `self.api_key` verbatim. When the API service stored the api_key as a JSON dict (the format used for several other LLM providers — see `api/apps/services/provider_api_service.py:313` which calls `json.dumps(api_key)` for non-string values), the `LocalAI.get_model_list` verify path sent an Authorization header of the form:

```
Authorization: Bearer {"api_key": "sk-local-xxx", "endpoint": "..."}
```

LocalAI rejects this with 401, and the UI surfaces "your api key is invalid" and the verify endpoint returns "102 no models found for provider LocalAI".

LocalAI typically does not require auth, so the right behavior is to extract the inner `api_key` when a JSON dict is given, and to take the no-auth path when the dict has no `api_key` field (the common case where the user just configures the base_url).

## Why this is a separate PR (not a follow-up to #18251)

PR #18251 (cycle 19) unifies `_get_api_key` in `rag/llm/model_meta.py` for the 3 sites that already had hand-rolled `try/except JSONDecodeError` blocks — `VolcEngine._get_api_key`, `OpenRouter._get_api_key`, `NewAPI._get_api_key` — and routes them through `_resolve_volcengine_credentials` / `_resolve_openrouter_credentials` in `rag/llm/key_utils.py`.

This PR is the **fourth** site in the same family: `LocalAI` had no `_get_api_key` override at all, so the Base default returned the raw `self.api_key`. I kept this PR independent of #18251 for two reasons:

1. The fix is small (2 hunks in one file) and self-contained — it can be reviewed and merged on its own.
2. `LocalAI`'s resolver has a different fallback semantic from the other 3 sites (empty string when the dict has no `api_key` field, because LocalAI's verify path takes the no-auth branch instead of the malformed-Bearer branch) — keeping it separate makes the rationale easier to read for reviewers.

The maintainer can rebase #18251 (cycle 19) to land first, or merge this one first — the two PRs are independent.

## Implementation summary

Two hunks in `rag/llm/model_meta.py:LocalAI`:

1. **Add `_get_api_key` override** mirroring the `VolcEngine` / `OpenRouter` / `NewAPI` pattern in the same file:
   - JSON dict with an `api_key` field → the inner key
[...]

PR #18386, opened 2026-08-17
fix(model_meta): Base._get_api_key handles JSON-dict api_key for the model-list verify path

Harsh23Kashyap:
## Summary

Sweep PR that moves the JSON-decode `api_key` resolution into `Base._get_api_key` in `rag/llm/model_meta.py`. This closes the bug class for **6 providers at once**: Xinference, HuggingFace, FunASR, NVIDIA, GreenPT, VLLM, LMStudio, RAGcon, AIMLAPI — all of which previously inherited the raw `return self.api_key` and sent a malformed `Authorization: Bearer {"api_key": "sk-xxx", "endpoint": "..."}` header when the API service stored the api_key as a JSON dict.

This is a **sweep** of the model_meta JSON-decode family (cycles 19, 22, 23) — the same shape as the per-site fixes in VolcEngine (line 68), OpenRouter (line 365), NewAPI (line 928), LocalAI (line 221), and Ollama (line 125), but in the base class so the 6 providers that do not override `_get_api_key` get the same wire-correct behavior for free.

## Why a sweep PR (different shape from cycles 22 / 23)

Cycles 22 (LocalAI) and 23 (Ollama) opened per-provider PRs because each had a different fallback semantic (LocalAI / Ollama take the no-auth path on JSON-dict-without-api-key; for real-API providers like NVIDIA the empty Bearer is no better than the malformed Bearer, so the no-auth path is irrelevant).

The 6 remaining providers in this PR all share the **`Base._get_raw_model_list`** path (or its equivalent in `OpenAIAPICompatible` subclasses that always sends the Bearer regardless of the resolved key). A single base-class change covers them and is strictly safer than per-provider:

- It is a **superset** of the per-site overrides: no behavior change for LocalAI / Ollama / VolcEngine / OpenRouter / NewAPI (they still override).
- It fixes the most common bug class (JSON-dict-with-api_key → inner key) for the 6 newly-covered providers.
- The base class default returns `""` for JSON-dict-without-api_key, which is safe for everyone: real-provider subclasses (NVIDIA, GreenPT, AIMLAPI) still 401 on the empty Bearer, but the malformed Bearer was strictly worse. LocalAI / Ollama subclass pattern of "no-auth when api_key is absent" continues to work through their explicit `if resolved_key:` gate.

If the maintainer prefers per-provider PRs, this PR can be split into 6 — but the base-class fix is the natural shape for the 6 sites that share the same code path.

## What was already fixed (per-site, merged in `main` or open as PRs)

- VolcEngine: `try/except JSONDecodeError` (line 68) — pre-existing, fragile (lets `TypeError` from `json.loads(None)` through; `AttributeError` on JSON non-dict)
- OpenRouter: `try/except Exception` (line 365) — pre-existing, too broad (silently uses raw JSON as Bearer)
- NewAPI: `try/except (JSONDecodeError, TypeError)` (line 928) — pre-existing, returns raw JSON for non-dict
- LocalAI: cycle 22 PR #18314 (open) — `_get_api_key` + `get_model_list` gate
- Ollama: cycle 23 PR #18321 (MERGED 2026-08-16) — same as LocalAI

The new base-class fix **supersedes** the pre-existing `VolcEngine` / `OpenRouter` / `NewAPI` / `LocalAI` / `Ollama` overrides for the resolution contract (they all return the same value), but the per-site overrides are kept because the local-server subclasses' `get_model_list` `if resolved_key:` gate needs the resolved-key contract and the cycle 22/23 tests assert the per-site behavior. **The fix is a strict superset**: the per-site overrides return exactly the same value as the base class for the cases that matter; the only difference is that the per-site overrides also do `if not self.api_key: return ""` first (an optimization, not a behavior change — the base class does the same `if not self.api_key: return ""` check at the top of the resolution).

## Implementation summary

One hunk in `rag/llm/model_meta.py:Base._get_api_key` (35 lines added, 1 line replaced). The new method body is the same resolution contract as the per-site overrides:

```python
def _get_api_key(self):
    if not self.api_key:
        return ""
    try:
        parsed = json.loads(self.api_key)
    except (JSONDecodeError, TypeError, ValueError):
        return self.api_key
    if isinstance(parsed, dict):
        return parsed.get("api_key", "") if "api_key" in parsed else ""
    return self.api_key
```

The `JSONDecodeError` and `json` imports are already at the top of the file, so no new imports are needed.

## Providers covered

| Provider | Inherits from | Pre-fix get_model_list | Post-fix behavior |
|---|---|---|---|
| `Xinference` | `Base` | uses `Base._get_raw_model_list` | Bearer = inner key (or `Bearer ` for empty) |
| `HuggingFace` | `Base` | uses its own `get_model_list` | same as Xinference |
| `FunASR` | `Base` | uses its own `get_model_list` (no Bearer at all for ASR) | inherits the resolution; the no-Bearer path is unaffected |
| `NVIDIA` | `OpenAIAPICompatible` | `f"Bearer {self._get_api_key()}"` | Bearer = inner key (or `Bearer ` for empty) |
| `GreenPT` | `OpenAIAPICompatible` | uses `Base._get_raw_model_list` | same as Xinference |
| `VLLM` | `OpenAIAPICompatible` | uses `Base._get_raw_model_list` | same as Xinference |
| `LMStudio` | `OpenAIAPICompatible` | uses `Base._get_raw_model_list` | same as Xinference |
| `RAGcon` | `OpenAIAPICompatible` | uses `Base._get_raw_model_list` | same as Xinference |
| `AIMLAPI` | `Base` | uses its own `_get_raw_model_list` with `f"Bearer {self._get_api_key()}"` | same as NVIDIA |

`MWS` (also `OpenAIAPICompatible` subclass) is **not** affected because it uses `require_mws_token(api_key)` in `__init__` which would fail at construction time for a JSON-dict api_key, not at the verify path.

## Tests

`test/unit_test/rag/llm/test_model_meta_base_get_api_key.py` (264 lines, 20 cases + 1 parametrized over 6 input shapes) — covers the resolution contract and the wire-level Bearer contract for the two integration paths the providers use:

- **Unit tests on `_get_api_key` (10 cases)** — the resolution contract:
  - JSON dict with `api_key` field (regular / empty / whitespace inner key)
  - JSON dict without `api_key` field (regular / `{}`)
  - Plain string (non-empty / empty)
  - JSON list / JSON string / malformed JSON → raw passthrough preserved
  - `None` / empty string

- **Integration tests on the wire-level Bearer contract (10 cases + 6-parametrized)**:
  - `test_xinference_sends_valid_bearer_when_json_dict_has_api_key` — `Base._get_raw_model_list` path: Bearer is the inner key, asserted via `session.get.call_args.kwargs["headers"]`
  - `test_nvidia_sends_valid_bearer_when_json_dict_has_api_key` — per-site `f"Bearer {self._get_api_key()}"` path: Bearer is the inner key
  - `test_nvidia_sends_empty_bearer_when_json_dict_has_no_api_key_field` — JSON-dict-without-api_key produces an empty Bearer (still 401s on real NVIDIA, but no worse than the pre-fix malformed Bearer)
  - `test_nvidia_get_model_list_authorization_header_value[<6 shapes>]` — full input shape matrix for NVIDIA
  - `test_xinference_get_model_list_skips_authorization_for_empty_resolved_key` — empty-string / None case still sends an empty Bearer (Base._get_raw_model_list has no gate, but the empty Bearer is strictly better than the pre-fix malformed Bearer)

Xinference is used as the proxy for the base class because it has no `__init__` side effects and inherits the base class method directly. NVIDIA is the second integration case because its `get_model_list` overrides `Base.get_model_list` and uses `f"Bearer {self._get_api_key()}"` directly, so it tests the per-site path independently of the `Base._get_raw_model_list` path.

## Verification

```
.venv/bin/python -m pytest test/unit_test/rag/llm/test_model_meta_base_get_api_key.py -v
============================== 20 passed in 0.24s ==============================
```

```
.venv/bin/python -m pytest test/unit_test/rag/llm/test_model_meta_base_get_api_key.py test/unit_test/rag/llm/test_model_meta_funasr.py
============================== 23 passed in 0.22s ==============================
```

```
ruff check rag/llm/model_meta.py test/unit_test/rag/llm/test_model_meta_base_get_api_key.py
All checks passed!

ruff format --check rag/llm/model_meta.py test/unit_test/rag/llm/test_model_meta_base_get_api_key.py
2 files already formatted
```

## Scope

- Source: 1 file, 1 hunk (35 insertions, 1 deletion) in `rag/llm/model_meta.py:Base._get_api_key`
- Tests: 1 file, 264 insertions in `test/unit_test/rag/llm/test_model_meta_base_get_api_key.py`
- Backward compatibility: The pre-fix behavior is preserved for plain-string api_key, JSON parse errors, and JSON non-objects. Only the JSON-dict-with-`api_key` and JSON-dict-without-`api_key` paths change, and they change to the wire-correct behavior. The per-site overrides for VolcEngine / OpenRouter / NewAPI / LocalAI / Ollama are unchanged.
- Migration: none

## Risk

Low. The base-class fix is a strict superset of the pre-existing per-site overrides. For the 6 newly-covered providers, the fix changes:
- The wire-level Bearer for JSON-dict api_key (from malformed `Bearer {"api_key": ...}` to valid `Bearer <inner_key>`)
- The wire-level Bearer for JSON-dict-without-api_key (from malformed to empty — still 401s on real APIs, no worse than before)

The most likely regression would be a provider that intentionally relied on the malformed-Bearer behavior. None of the 6 providers do; the malformed Bearer was always an unintended side effect of the raw passthrough.

## Follow-ups (out of scope)

The per-site overrides in VolcEngine (line 68), OpenRouter (line 365), NewAPI (line 928) are now strictly equivalent to the base class for the resolution contract. A future PR could remove the overrides as cleanup, but that's a separate change with a different review surface. This PR leaves the per-site overrides in place to minimize the diff scope and preserve the existing cycle 19/22/23 test coverage.

### DIFF
--- a/rag/llm/model_meta.py
+++ b/rag/llm/model_meta.py
@@ -33,6 +33,41 @@ class Base(ABC):
         self.base_url = base_url
 
     def _get_api_key(self):
+        # Shared JSON-decode resolution for every model-list verify path in
+        # this file. Mirrors the per-site pattern in VolcEngine (line 68),
+        # OpenRouter (line 365), NewAPI (line 928), LocalAI (line 221), and
+        # Ollama (line 125) -- but in the base class so the 6 providers that
+        # do not override this method (Xinference, HuggingFace, FunASR,
+        # NVIDIA, GreenPT, VLLM, LMStudio, RAGcon, AIMLAPI) get the same
+        # wire-correct behavior for free.
+        #
+        # When the API service stores the api_key as a JSON dict -- the
+        # format used by ``api/apps/services/provider_api_service.py:313``
+        # which calls ``json.dumps(api_key)`` for non-string values -- the
+        # raw ``self.api_key`` would render as
+        # ``Authorization: Bearer {"api_key": "sk-xxx", ...}``, which
+        # real-provider endpoints 401. Resolve the same way the per-site
+        # overrides do:
+        #   * JSON dict with an ``api_key`` field  -> the inner key.
+        #   * JSON dict without ``api_key``        -> ``""`` so the resolved
+        #     value is falsy and any ``if resolved_key:`` gate in the
+        #     subclass's ``get_model_list`` skips the Authorization header
+        #     entirely. For real-provider subclasses (NVIDIA, GreenPT,
+        #     AIMLAPI) this is no better than the pre-fix malformed Bearer --
+        #     they 401 either way -- but the LocalAI/Ollama subclass pattern
+        #     of "no-auth when api_key is absent" still works.
+        #   * Plain string (the common case)  -> returned as-is.
+        #   * JSON parse error, JSON non-object, or non-string api_key  ->
+        #     fall back to the pre-fix raw passthrough so we do not
+        #     regress any caller that depends on the historical behavior.
+        if not self.api_key:
+            return ""
+        try:
+            parsed = json.loads(self.api_key)
+        except (JSONDecodeError, TypeError, ValueError):
+            return self.api_key
+        if isinstance(parsed, dict):
+            return parsed.get("api_key", "") if "api_key" in parsed else ""
         return self.api_key
 
     def _get_model_list_url(self):
--- a/test/unit_test/rag/llm/test_model_meta_base_get_api_key.py
+++ b/test/unit_test/rag/llm/test_model_meta_base_get_api_key.py
@@ -1,0 +1,264 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Unit tests for the shared ``Base._get_api_key`` JSON-decode resolution.
+
+The base class ``rag.llm.model_meta.Base`` previously returned
+``self.api_key`` verbatim. When the API service stored the api_key as a
+JSON dict (the format ``api/apps/services/provider_api_service.py:313``
+produces for non-string values via ``json.dumps(api_key)``), every subclass
+that did not override ``_get_api_key`` and used ``Base._get_raw_model_list``
+or its own ``f"Bearer {self._get_api_key()}"`` call sent a malformed
+``Authorization: Bearer {"api_key": "sk-xxx", ...}`` header. Real-provider
+endpoints 401. The fix mirrors the per-site pattern in VolcEngine /
+OpenRouter / NewAPI / LocalAI / Ollama but in the base class so the
+6 providers that do not override this method get the same wire-correct
+behavior for free: Xinference, HuggingFace, FunASR, NVIDIA, GreenPT,
+VLLM, LMStudio, RAGcon, AIMLAPI.
+
+The local-server subclass (Xinference) is used for the unit tests below
+because it inherits the base class method directly without any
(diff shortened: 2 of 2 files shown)


Lynn-Inf (approved):
LGTM

Issue #17280, opened 2026-07-23
Figure description language not propagated to PDF/DOCX(book,manual)/XLSX vision wrappers


xugangqiang:
## Problem

PR #17227 fixed the dataset language propagation for the `vision_figure_parser_docx_wrapper_naive` path (DOCX via `rag/app/naive.py`), but three other wrappers in `deepdoc/parser/figure_parser.py` have the same bug: they never receive `lang`, so `VisionFigureParser.self.language` always defaults to `"English"` and the figure-description prompt's `{{ language }}` variable always renders as English.

## Affected paths

| Wrapper | File:Line | Callers (need `lang=lang`) |
|---|---|---|
| `vision_figure_parser_docx_wrapper` | `deepdoc/parser/figure_parser.py:49` | `rag/app/book.py:88`, `rag/app/manual.py:277` |
| `vision_figure_parser_figure_xlsx_wrapper` | `deepdoc/parser/figure_parser.py:69` | `rag/app/table.py:83` (also forward `lang` through `Excel.__call__` called at `table.py:408`) |
| `vision_figure_parser_pdf_wrapper` | `deepdoc/parser/figure_parser.py:102` | `rag/app/naive.py:121` (`by_deepdoc`), `rag/app/manual.py:258`, `rag/app/paper.py:174` |

## Root cause

Every `chunk()` function declares `lang="Chinese"` as a named parameter, so `lang` is captured and is **not** in `**kwargs`. The three wrappers receive `**kwargs` only, so `lang` never reaches `VisionFigureParser.__init__`, and `self.language = kwargs.get("lang") or "English"` always falls back to `"English"`.

## Impact

Figure descriptions for PDF (naive/manual/paper), DOCX (book/manual), and Excel (table) are always generated in English, regardless of the dataset language setting. This causes inconsistent behavior within the same dataset (e.g., naive DOCX figures honor the language, but PDF figures do not).

## Note on `LLMBundle(lang=)`

Passing `lang` to `LLMBundle` has no effect on figure descriptions: `describe_with_prompt()` (`rag/llm/cv_model.py:353`) does not read `self.lang`. The actual language control is through the prompt's `{{ language }}` variable. The `LLMBundle(lang=)` call is still worth adding for consistency, but it is not what fixes the output language.

## Suggested fix

1. Add `lang="English"` parameter to the three wrapper functions.
[...]

PR #18392, opened 2026-08-17
fix(figure_parser): accept lang as named parameter on VisionFigureParser

Taranum01:
Regression for #17280.

\`VisionFigureParser.__init__\` previously took \`*args, **kwargs\` and read the dataset language via \`kwargs.get("lang")\`. The three wrappers above pass \`lang=lang\` as an explicit keyword argument, so \`kwargs.get("lang")\` did return the value — but the contract is implicit and easy to miss when adding new call sites (any caller that forgets \`lang=\` and only forwards \`**kwargs\` would silently default to English).

Make \`lang\` an explicit named parameter so the type system catches the omission and the value is normalized via \`_normalize_vision_language\` for consistency with the wrappers' own handling. \`**kwargs.get("lang")\` is kept as a fallback so older callers that don't pass \`lang\` keep working.

No behavior change for any current call site.


xugangqiang:
pls resolve conflict @Taranum01

JinHai-CN:
@Taranum01 please re-open the PR, if you are still working on it.

Taranum01:
Re-opened as a clean follow-up at #19511 (rebased onto current main, +9 / -2 in `deepdoc/parser/figure_parser.py` only).

The previous PR carried the full main-sync history from when the branch was first cut (929k insertions / 111k deletions across hundreds of files), which made the actual 7-line figure_parser fix impossible to review. The new branch is the same code change — `lang` is now an explicit named parameter on `VisionFigureParser.__init__`, normalized via the existing `_normalize_vision_language` helper, with the `\*\*kwargs.get("lang")` fallback preserved for older callers.

Closing this old one in favor of #19511.

Taranum01:
Replying directly here: re-opened as **#19511** — same 7-line fix, but rebuilt against current main so the diff is just `deepdoc/parser/figure_parser.py` (+9 / -2) instead of the 929k-line main-sync the previous PR carried. Closing this old #18392 in favour of the new one.

PR #18393, opened 2026-08-17
fix(migration): prompt on non-empty target instead of auto-clearing

### LINKED ISSUE #18354: [Bug]: migration.sh restore may corrupt stateful volumes because existing files are not removed before extraction
### Self Checks

cb93883

### RAGFlow image version

v0.26.4

```Markdown
x86_64
docker v29.7.2
kylin linux
```

### Actual behavior

docker/migration.sh restore currently restores backup archives into existing Docker volumes without clearing or recreating the volumes first.

When a target volume already exists, the script only prints:

IMPORTANT: Restoring will OVERWRITE existing data!

and then reuses the existing volume:

if ! volume_exists "$volume"; then
    docker volume create "$volume"
else
    echo "Using existing Docker volume: $volume"
fi

The actual restore operation is:

docker run --rm \
    -v "$volume":/target \
    -v "$(pwd)/$backup_folder":/backup \
    alpine tar xzf "/backup/$backup_file" -C /target

However, tar xzf only overwrites files that exist in the archive and creates missing files. It does not remove files that already exist in the target volume but are absent from the backup.

Therefore, restoring into an existing volume results in a merged filesystem rather than an exact restoration of the backup.

This can corrupt stateful services such as MySQL or Elasticsearch.

For example, if a MySQL volume has already been initialized and contains newer InnoDB redo log files, restoring an older backup may leave stale redo log files in the volume.

Example:

Existing volume:

#innodb_redo/

  #ib_redo10
  #ib_redo11
  #ib_redo12

Backup:

#innodb_redo/
  #ib_redo1
  #ib_redo2

After the current restore operation, the volume may contain:

#innodb_redo/
  #ib_redo1
  #ib_redo2
  #ib_redo10
  #ib_redo11
  #ib_redo12

This is not equivalent to restoring the original backup state.

In my case, restoring a RAGFlow backup into an already initialized MySQL volume caused MySQL to fail during startup with errors similar to:

[ERROR] [InnoDB] Missing redo log file
[ERROR] [InnoDB] Plugin initialization aborted with error Generic error
[ERROR] Failed to initialize DD Storage Engine
[ERROR] Data Dictionary initialization failed
[ERROR] Aborting

The issue disappeared when restoring the same backup into a newly created empty volume before starting RAGFlow.

### Expected behavior

_No response_

### Steps to reproduce

```Markdown
1. Start RAGFlow normally.
2. Stop RAGFlow and create a backup using migration.sh backup.
3. Remove the original volumes.
4. Start RAGFlow once so that Docker/MySQL creates and initializes new volumes.
5. Stop RAGFlow.
6. Run migration.sh restore using the previous backup.
7. Start RAGFlow again.
8. MySQL may fail because files generated during step 4 but absent from the backup are still present after restore.
```

### Additional information

The current warning says restoring will "OVERWRITE existing data", but the actual behavior is overwrite + merge. Files that exist only in the current volume are preserved, which is especially dangerous for database storage.

Taranum01:
Fixes #18354

## Behavior change — safety-first on non-empty targets

`docker/migration.sh:perform_restore` no longer silently auto-clears a non-empty Docker volume before extracting the backup. Per the review on #18393, the script now:

1. Inspects the target via `alpine ls` (same glob as the cleanup path: `/target/* /target/.[!.]* /target/..?*`).
2. If the volume is empty: silent proceed (the common case).
3. If the volume is non-empty: print the first ~20 entries, recommend the operator clean it up themselves (or use a fresh empty volume), and **abort** for that volume — the script continues with the next volume in the list so a single dirty volume doesn't block the rest.

When the operator wants the legacy destructive behavior back (e.g. for an automated test that already verified the target is expendable), pass `--force` to skip the inspection.

```
$0 restore                              # safety-first; aborts on non-empty
$0 --force restore                     # legacy destructive auto-clear
```

## Cleanup glob (unchanged from previous PR)

The cleanup glob is split into three patterns to cover every POSIX-basename shape the volume might contain:

- `/target/*` — non-hidden entries
- `/target/.[!.]*` — single-dot hidden entries (`.env`, `.cache`)
- `/target/..?*` — multi-dot hidden entries (`.example`, `..tmp`, `...state`)

POSIX `[!.]` only excludes the second character being `.`, which leaves names like `.example` and `...state` unremoved under the two-character pattern. The third glob catches those.

## Files

- `docker/migration.sh`: added target-contents pre-flight + `--force` flag; help text updated.


JinHai-CN:
This PR includes too many noises, would you please check and clean this PR?

myf-bee (commented):
Directly emptying the target volume is a relatively risky operation.
Suggestions: 1. Add documentation to explain this. 2. If the target volume is not empty, prompt the user and exit. Let the user clean it up themselves or use a new empty volume. See the discussion in PR#18433.

Taranum01:
Pushed a follow-up commit addressing the review feedback on this PR:

**CodeRabbit (@coderabbitai[bot])** flagged that the cleanup glob `/target/* /target/.[!.]*` does not match basenames starting with two or more dots (`..foo`, `...state`, `..tmp`). Added `/target/..?*` so every POSIX-basename shape is covered; verified the combined pattern removes `regular`, `.cache`, `.hidden`, `..foo`, `..tmp`, and `...state` against a synthetic directory.

**@myf-bee** raised that the explicit `rm -rf` step is destructive. Rewrote the existing-volume warning to spell out *what* gets deleted (files not present in the backup, not just "overwrite"), updated the pre-restore recommendation to suggest a snapshot volume, and reworded the confirmation prompt so the irreversibility is on the line the user types 'y' against.

This is intended as the closing change for PR #18393.

myf-bee:
@Taranum01   
my  Suggestions  is  :  If the target volume is not empty, prompt the user and exit. Let the user clean it up themselves or use a new empty volume. This script will not directly clean up the target volume. Instead, it will prompt the user to either clean up the target volume manually or use an empty volume，thanks!

Taranum01:
Thanks for the suggestion, @myf-bee. Force-pushed the change.

## What the script now does on a non-empty target

```
🔧 Step 1/4: Restoring docker_mysql_data...
  📥 Restoring data from mysql_backup.tar.gz...

  🔴 REFUSING TO AUTO-CLEAR target volume: docker_mysql_data
     The current restore would silently destroy the following state:
       /target/ib_logfile0
       /target/ib_logfile1
       /target/mysql.ibd
       ...

     To restore, choose one of:
       a) Clean up the target volume yourself, then re-run:
          docker compose down && docker volume rm docker_mysql_data && docker volume create docker_mysql_data
       b) Point this script at a fresh, empty volume (override with a custom -v path).
       c) Re-run with --force to accept the destructive auto-clear (legacy behavior).

❌ Restore aborted for docker_mysql_data.

🔧 Step 2/4: Restoring docker_minio_data...
  ... continues with the next volume
```

## What's new

1. **Pre-flight inspection** via `alpine ls` using the same glob the cleanup path uses (`/target/* /target/.[!.]* /target/..?*`). Prints up to 20 entries so the operator sees what would be destroyed.

2. **Default behavior change**: missing state on target = silent proceed (unchanged), but **present state = hard abort** for that volume. The script keeps running with the next volume so a single dirty volume doesn't block the whole restore.

3. **`--force` flag** added so existing automation / CI flows that already verified the target is expendable keep working. Help text documents the flag.

4. **`--force` parsed in main()** alongside `-p`, before positional args so it works regardless of where `-p` is in the argv.

## Verified

```$ bash -n docker/migration.sh  # syntax check
(no output)
```

PR description rewritten to lead with the behavior change. The PR title was also updated to "fix(migration): prompt on non-empty target instead of auto-clearing".

Taranum01:
Both CodeRabbit findings addressed and force-pushed.

## 1. POSIX-compatible volume inspection (shopt doesn't work in Alpine BusyBox)
The previous preflight used `shopt -s dotglob nullglob`, which fails silently on Alpine BusyBox `sh`. The unexpanded globs would get passed as literal strings to `printf`, making `target_contents` non-empty for empty volumes and aborting all of them.

New code uses plain POSIX globs inside an alpine container:
```sh
for f in /target/* /target/.[!.]* /target/..?*; do
    [ -e "$f" ] && exit 0
done
exit 1
```

This catches every basename shape (single-dot, multi-dot, non-hidden) without `shopt`, and exits 0 only when at least one matching path exists.

## 2. Atomic preflight — abort BEFORE any volume is touched
The previous per-volume check happened mid-loop. Earlier volumes could be cleared and restored before later non-empty volumes aborted, then the script printed "Restore completed successfully!" despite a partial restore.

New code does a single preflight pass over **every** volume BEFORE creating or clearing anything. If any is non-empty:
- Lists all dirty volumes (with up to 20 entries per volume so the operator sees what would be destroyed)
- Exits 1 with "Restore aborted before any volume was touched."

No more partial-success surprises.

## 3. Removed unsupported `-v` recovery instruction
The previous recovery message said "point this script at a fresh, empty volume (override with a custom -v path)" but `main()` doesn't recognize `-v` — it would become the next positional arg and fail with a confusing error. Replaced with the supported `--force` flag (the only way to override the preflight).

## Verified
```bash -n docker/migration.sh`` clean.

No migration.sh test files in the repo; manual smoke test is the only verification path. The new preflight loop is small and self-contained, so it's easy to spot regressions.

myf-bee:
@Taranum01 
1.  This prompt is also displayed when the --force parameter is not used, which is misleading: "Continue with the restore? (type 'y' to permanently overwrite these volumes) (y/N):". It should only output this prompt when the --force parameter is added.

2.  The --force parameter has a positional trap; currently, it must be the first parameter for it to be parsed correctly.

Taranum01:
Thanks for the heads-up, @myf-bee. Both issues fixed and force-pushed.

## 1. Confirm prompt now skips when --force is set
The previous PR showed 'Continue with the restore? (type \'y\' to permanently overwrite these volumes) (y/N):' even when RESTORE_FORCE=true. The preflight already skipped under --force, so the prompt should too. Gated the prompt on `[ \"\$RESTORE_FORCE\" != \"true\" ]` and added an explicit 'Pass --force to skip this prompt and auto-clear the target.' line so the operator knows the override exists.

## 2. --force now works in any position
The previous parser broke on the first non-flag arg, which meant `restore --force` was silently parsed as `operation=restore` and `--force` became the backup_folder arg. Reworked into two passes:

1. Walk \$@ and pull -p / --force into locals, collecting non-flag args in a `_positional` array.
2. `set -- \"${_positional[@]}\"` to restore the positional args so the existing operation/folder parser sees the same shape it always did.

Verified all six orderings with a small standalone harness:

    restore                         -> op=restore  force=false  p=default
    --force restore                -> op=restore  force=true   p=default
    restore --force                -> op=restore  force=true   p=default
    -p ragflow restore my_backup    -> op=restore  force=false  p=ragflow  folder=my_backup
    -p ragflow restore my_backup --force  -> op=restore force=true p=ragflow folder=my_backup
    restore --force -p test        -> op=restore  force=true   p=test

\"bash -n docker/migration.sh\" clean.

myf-bee:
@Taranum01 ,  Thanks for the fix.    but  i think,    restore set --force,    This prompt should be kept, and the user should enter 'y' to overwrite these volumes.
 restore not set --force ,  the prompt  " Continue with the restore? (type 'y' to permanently overwrite these volumes) (y/N): "   , shouldn't show up

PR #18400, opened 2026-08-18
fix(web): confirm before removing models from a provider instance

dcc123456:
### Summary

fix(web): confirm before removing models from a provider instance

Detaching a model was a one-click, irreversible mutation - both the per-row minus button and the "Remove all models" batch action fired immediately. Route both through ConfirmDeleteDialog, worded as "remove" rather than "delete" to match the action.

### DIFF
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -9,6 +9,7 @@ export default {
       delete: 'Delete',
       deleteModalTitle: 'Are you sure to delete it ?',
       deleteThem: 'Are you sure to delete them ?',
+      removeModalTitle: 'Are you sure to remove it ?',
       ok: 'Ok',
       cancel: 'Cancel',
       yes: 'Yes',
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for this — making the irreversible remove actions confirmable is a solid UX/safety improvement.

Reviewed the logic:
- `ConfirmDeleteDialog` API usage is correct (`hidden`, `onOk`, `title`, `okButtonText`, `content` all match the component's props).
- When `isAdded` is false the dialog is `hidden`, so the trigger button fires `onAdd` directly; when true the button becomes the dialog trigger and removal runs from `onOk`. Correct in both the per-row and batch-toggle paths.
- Translation keys are consistent: `common.removeModalTitle` is added to en/zh, and `common.remove` already exists.

One minor, non-blocking note: `removeModalTitle` is only added to `en.ts` and `zh.ts`, so the other locales will show the raw key or fall back to the default language (depending on the i18n fallback config) until they catch up. Fine for now, just flagging it.

Approving.

PR #18407, opened 2026-08-18
Fix: handle title and truncate document preview by sentence

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/rag/svr/task_executor_refactor/dataset_skill_generator.py
+++ b/rag/svr/task_executor_refactor/dataset_skill_generator.py
@@ -221,22 +221,57 @@ async def doc_summary_for_skill(
     )
 
 
-def build_skill_md(node: "SkillNode") -> str:
+_SENTENCE_ENDINGS = "。！？!?；;"
+
+
+def _strip_md_headings(text: str) -> str:
+    """Convert ATX heading markers (``###`` at line start) into bold text,
+    keeping the heading text and the line structure. e.g. ``### 客户服务``
+    becomes ``**客户服务**``. Line breaks survive, so a multi-section
+    summary keeps its readability inside the YAML ``description`` field."""
+    if not text:
+        return text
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Solid improvement — preserving document titles and truncating at sentence boundaries are real fixes, and the `doc_names` threading is consistent across all call sites (`skill_node_es_row`, `_collect_skill_rows`, `skill_tree_md_snippet`, `skill_tree_node`, `skill_all_es_row`). Imports (`Optional`, `Dict`) and `from __future__ import annotations` check out.

I also sanity-tested the new `description: |` block scalar against several inputs (CJK truncation, multi-line headings via `_strip_md_headings`, empty summary, `#`/`:`/`*` characters, bullet lists) and it parses cleanly.

A few non-blocking suggestions:

1. **Missing tests** — `_strip_md_headings`, `_truncate_at_sentence_end`, `_escape_md_text`, and `build_skill_md` are pure functions and easy to unit-test. This change touches YAML serialization + slugification, so a few tests would guard against regressions.
2. **`skill_safe_name` docstring is now inaccurate** — the docstring still says it "Mirrors Corpus2Skill `_safe_name` for cross-system stability", but it now keeps CJK while the mirrored `_safe_name` (elsewhere) presumably still strips non-ASCII. If both systems are expected to produce identical folder names, this will diverge. Please update `Corpus2Skill._safe_name` in tandem, or fix the docstring.
3. **CJK-only whitelist** — `\u4e00-\u9fff` keeps Chinese but still strips Japanese kana, Korean Hangul, Cyrillic, and accented Latin. Those labels silently collapse to an empty slug (folder name falls back to the bare `skill-NN` / `group-NN` prefix). If multilingual labels are intended, the whitelist is incomplete.
4. **Minor** — the `|` literal block leaves a trailing newline on the `description` value (`|-` would strip it), and `_escape_md_text` doesn't escape `()`/`<`/`>`, which is fine for the current bold-only usage but would break if a name is ever embedded in a link.

None of these block merge — nice work.

PR #18409, opened 2026-08-18
Refactor: make page number readable

wangq8:


wangq8 (commented):
Thanks for the PR. The core off-by-one fix in `task_executor.py` is correct, but there's a consistency concern and a couple of edge-case notes on the `task_service.py` change.

### ✅ `rag/svr/task_executor.py`

Correct fix. `from_page`/`to_page` are 0-indexed with an *exclusive* `to_page` (see `queue_tasks` in `api/db/services/task_service.py`: `s -= 1`, `e = min(e - 1, pages)`, `to_page = min(p + page_size, e)`). So the 1-indexed inclusive range is `[from_page + 1, to_page]`. The old `to_page + 1` was off by one; the new `to_page` is right.

### ⚠️ `api/db/services/task_service.py`

1. **Inconsistent convention.** This message stays 0-indexed (`from_page ~ to_page - 1`), while `set_progress` is now 1-indexed inclusive (`from_page + 1 ~ to_page`). For the same "all pages" task (`from_page=0`, `to_page=MAXIMUM_TASK_PAGE_NUMBER`), `reuse_prev_task_chunks` renders `Page(0~99999999)` while `set_progress` renders `Page(1~100000000)`. Consider aligning both to the same convention (e.g. 1-indexed inclusive) so the same range never shows two different numbers.

2. **Sentinel branch edge case.** The second condition branch allows `from_page == to_page == MAXIMUM_TASK_PAGE_NUMBER`. With `to_page - 1` that would render a *descending* range `Page(100000000~99999999)`. This branch doesn't look reachable for parse tasks, but if it's kept as defensive coverage, the new expression breaks it — the `- 1` only makes sense when `to_page` is a real exclusive upper bound.

3. **Minor type-safety inconsistency.** The surrounding condition defensively wraps values with `int(...)` (e.g. `int(task["to_page"])`), but the new f-string does raw arithmetic `task['to_page'] - 1`. It's safe for the in-memory int values in `parse_task_array`, but it diverges from the existing defensive style.

wangq8 (commented):
Thanks for the update — this addresses all three points from my previous review.

Verified against `queue_tasks`:

- `from_page`/`to_page` are 0-indexed with an *exclusive* `to_page` (`s -= 1`, `e = min(e - 1, pages)`, `to_page = min(p + page_size, e)`), so `Page({from_page + 1}~{to_page})` is the correct 1-indexed inclusive range. ✅
- `set_progress` (`task_executor.py`) and `reuse_prev_task_chunks` (`task_service.py`) now render the *same* format for the same range (e.g. "all pages" → `Page(1~100000000)`). ✅
- Removing the `from_page == to_page == MAXIMUM_TASK_PAGE_NUMBER` branch is safe: `queue_tasks` always produces `from_page < to_page` (PDF `to_page = min(p + page_size, e) > p`; table `to_page = min(i + 3000, rn) > i`; other `to_page = MAXIMUM_TASK_PAGE_NUMBER`), so the sentinel-equality state is unreachable for parse tasks. ✅
- The `int(...)` casts in the f-string now match the defensive style of the surrounding condition. ✅

LGTM — no further issues from me.

PR #18411, opened 2026-08-18
fix(web): show introduction for built-in pipelines audio/email/knowledge compiler

euvre:

On the dataset configuration page, the **Built-in pipeline introduction** panel shows only an empty placeholder ("This will display a visual explanation of the dataset categories") for the built-in pipelines **Audio**, **Email**, and **Knowledge Compiler**, while other methods (e.g. General/naive) show a proper description.

**Root cause** (two combined factors):
1. The locale files have no description entries for `audio`, `email`, or `knowledgeCompiler` under `knowledgeConfiguration`, so `CategoryPanel` falls back to the `methodEmpty` placeholder.
2. `CategoryPanel` gated the entire introduction panel on the pipeline having sample images in `ImageMap`; these three pipelines have no sample images, so the panel body was never rendered even if a description existed.

### Changes
- `web/src/pages/dataset/dataset-setting/category-panel.tsx`: render the introduction panel whenever a description exists (`t(descriptionKey, { defaultValue: '' })`); the example-image section is now rendered only when sample images are available, so pipelines without images still show their description.
- All 13 locale files (en, zh, zh-traditional, de, fr, it, ja, ko, ru, tr, vi, id, bg): add `audio`, `email`, and `knowledgeCompiler` description entries, translated per locale based on the backend pipeline template descriptions.

### Verification
- Audio / Email / Knowledge Compiler: introduction panel now shows the description (e.g. "Supported file formats are WAV, MP3, AAC, FLAC, OGG and other common audio formats. This method transcribes audio files into text using a speech-to-text model."); no more empty placeholder.
- Regression: General (naive) still renders the rich panel with description + example images, unchanged.
- zh locale spot-checked: Audio shows the Chinese description.
- `oxlint` / `oxfmt` pre-commit hooks pass.

### Type of change


wangq8 (commented):
Thanks for the fix — the two-factor root cause is correct, and decoupling the description panel from `imageList` is the right structure.

I verified the logic against `category-panel.tsx`, `utils.ts` (`ImageMap` / `DescriptionKeyMap`), and the `en.ts` `knowledgeConfiguration` section:

- Every method in `ImageMap` has a corresponding description entry, so no existing panel regresses (including `tag` and `knowledge_graph` → `knowledgeGraph`).
- `audio` / `email` / `knowledgeCompiler` now render a description-only panel as intended.
- The `zh.ts` bytes are intact (no mojibake).

A few non-blocking suggestions:

1. **Missing locales** — only 13 of the 16 locale files were updated. `ar.ts`, `es.ts`, and `pt-br.ts` don't have the new `audio` / `email` / `knowledgeCompiler` keys. Since `fallbackLng` is `'en'`, those users will silently see English text. Consider adding the three remaining locales for parity.
2. **`defaultValue: ''` rarely fires** — because of the `en` fallback, `t('knowledgeConfiguration.audio', { defaultValue: '' })` always resolves to the English entry even for locales missing the key. That's fine, but the new code comment ("Methods without a description entry get an empty string") only holds when no fallback locale has the key. Minor.
3. **Translation quality** — the 13 translations look machine-generated. Worth a native-speaker spot check, especially the less common locales (bg, id, vi).

None of these block merge — nice work.

dcc123456 (approved):

PR #18412, opened 2026-08-18
Fix: correct chunk order for non‑PDF documents

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Correct and low-risk change — this aligns `_load_chunks_for_doc` with the ordering already used on the retrieval path.

- `chunk_order_int` (mapped to `_order_id` in the doc stores) is set unconditionally on every surviving chunk in both the Python tokenizer (`rag/flow/tokenizer/tokenizer.py`) and the Go tokenizer (`internal/ingestion/component/tokenizer.go`), so it is a stable reading-order index for non-PDF (and PDF) docs.
- Retrieval already sorts `chunk_order_int` → `page_num_int` → `top_int` (`rag/nlp/search.py`, `internal/service/nlp/retrieval.go`, `internal/service/chunk/chunk.go`), so this makes the streaming order consistent.

Non-blocking nits:
1. The docstring above still says ordering is `asc("page_num_int").asc("top_int")` — please update it to reflect the new primary key.
2. Optional: add `chunk_order_int` to `select_fields` for consistency, though the engine sorts independently of output columns so it isn't required.

LGTM.

PR #18418, opened 2026-08-18
fix(web): surface the validation error when a metadata value save is blocked

dcc123456:
### Summary

fix(web): surface the validation error when a metadata value save is blocked

Confirming with an invalid field returned silently, so the modal stayed open with no explanation. Show the existing field error as a toast.


wangq8 (commented):
LGTM. The fix correctly surfaces the previously-silent validation failure: when `type === MetadataType.Setting` and `valueError.field` is set, the modal stayed open with no explanation. Showing `valueError.field` (the translated warning set in `handleChange`) as an error toast explains why the save is blocked.

The import `@/components/ui/message` matches the sibling `use-manage-modal.ts`, so the path is correct.

Optional nit: the toast and the inline field error will both appear on save; that's fine, but if you want to avoid duplication you could clear the inline error once the user edits the field. Not blocking. Approving.

PR #18419, opened 2026-08-18
Fix: Edit wiki-style text

cike8899:
### Summary

Fix: Edit wiki-style text


wangq8 (commented):
Thanks for the PR. A few issues to address before merge:

1. **zh.ts regression**: `skills` is changed from `'技能'` to `'To Skills'`, which puts an English string into the Chinese locale (and `"To Skills"` is not a meaningful label). Please revert to `'技能'` or the intended Chinese wording.

2. **Dropped i18n**:
   - `compilation-template-util.ts`: `formatKindLabel` already receives a `t` function and uses `t('knowledgeCompilation.graph')`, but the new `PageIndex`/`Mindmap` branches return hardcoded strings. Translation keys already exist (`knowledgeCompilation.navTree` = "Tree/Page index", `knowledgeCompilation.structureMindmap` = "Mind map"), so use `t('knowledgeCompilation.navTree')` and `t('knowledgeCompilation.structureMindmap')` instead of hardcoding.
   - `use-blueprint-selection.ts`: `t('knowledgeCompilation.custom')` is replaced with the literal `'Custom'` and `t` is dropped from the `useMemo` deps, which removes the Chinese translation for the Custom label.

3. **Removed keys**: `tags: 'Tags'` (en) and `tags: '标签'` (zh) are removed. Please confirm these keys are no longer referenced anywhere — the zh `tags` removal sits alongside still-present keys like `editTags`/`tagsPlaceholder`, so double-check nothing still consumes them.

PR #18422, opened 2026-08-18
Fix: some pdf with chinese name cannot preview

wangq8:


wangq8 (commented):
The change is correct and minimal. Without `flags=re.ASCII`, `\w` inside `[^\w.\-]` matches Unicode word characters (Python 3 `str` patterns default to Unicode), so a filename like `中文报告.pdf` passes through unchanged and ends up as a non-ASCII value in the `Content-Disposition` header — which is invalid/unsafe in HTTP headers and breaks preview in some clients. Adding `re.ASCII` makes `\w` match only `[a-zA-Z0-9_]`, so non-ASCII characters (CJK, accented Latin, etc.) are replaced with `_`, guaranteeing an ASCII-only filename. Verified: `中文报告.pdf` → `____.pdf`.

Suggestions (non-blocking):

1. Add a unit test covering the non-ASCII filename case in `test/unit_test/api/utils/test_file_response_headers.py`, e.g.:
```python
def test_apply_preview_sanitizes_non_ascii_filename():
    response = _DummyResponse()
    module.apply_preview_file_response_headers(response, "application/pdf", "pdf", "中文报告.pdf")
    assert response.headers["Content-Disposition"] == 'inline; filename="____.pdf"'
```
This locks in the behavior and guards against regression.

2. Minor UX note: for non-ASCII filenames the original name is lost (e.g. `café.pdf` → `caf_.pdf`). If preserving the display name matters, consider RFC 5987 `filename*=UTF-8''...` alongside the ASCII `filename` fallback. Not required for this bug fix — just flagging the tradeoff.

Otherwise LGTM.

PR #18427, opened 2026-08-18
Fix: Incorrect translation of "search" on the chunk page.

cike8899:
### Summary
Fix: Incorrect translation of "search" on the chunk page.

### DIFF
--- a/web/src/pages/chunk/parsed-result/add-knowledge/components/knowledge-chunk/components/chunk-result-bar/index.tsx
+++ b/web/src/pages/chunk/parsed-result/add-knowledge/components/knowledge-chunk/components/chunk-result-bar/index.tsx
@@ -7,19 +7,20 @@ import {
 } from '@/components/ui/popover';
 import { Radio } from '@/components/ui/radio';
 import { Segmented } from '@/components/ui/segmented';
-import { useTranslate } from '@/hooks/common-hooks';
 import { cn } from '@/lib/utils';
 import { LucideFilter, Plus } from 'lucide-react';
 import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { ChunkTextMode } from '../../constant';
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix. A few notes:

1. **Translation fix is correct.** `t('search', { keyPrefix: 'common' })` resolves to `common.search` ("Search"), which fixes the previous lookup against `chunk.search`. ✅

2. **Scope mismatch — please confirm intent.** The PR is titled "Fix: Incorrect translation of 'search'", but it also makes substantial, unrelated changes to `web/src/pages/dataflow-result/chunker.tsx`:
   - Removes `searchString`, `handleInputChange`, `available`, `handleSetAvailable` wiring. These were effectively dead in this page: the dataflow-result `ChunkResultBar` only accepts `{ changeChunkTextMode, createChunk, isReadonly }` and renders no search box/filter, so those props never did anything here.
   - Replaces `useHandleChunkCardClick` with local `selectedChunkId` state. This actually fixes a latent bug — the hook returns `selectedChunk` (an `IChunk`), but the code destructured `selectedChunkId` (always `undefined`), so `selected` was always `false` and chunk highlighting was broken.

   These are good fixes, but they should ideally be split into a separate PR (or at least called out in the description) so the translation change stays isolated and easy to revert.

3. **Optional search props vs. unconditional `<SearchInput>`.** `handleInputChange` and `searchString` are now optional in `ChunkResultBarProps`, but the component still renders `<SearchInput onChange={handleInputChange} value={searchString} />` unconditionally. If a future caller omits them, the search box renders non-functional. Consider keeping these props required, or only rendering `SearchInput` when `handleInputChange` is provided.

Otherwise LGTM — the `changeChunkTextMode` narrowing to `(mode: ChunkTextMode) => void` is consistent with both callers.

PR #18429, opened 2026-08-18
Feat: Remove `pageindex` from the dataset configuration page.

cike8899:
### Summary

Feat: Remove `pageindex` from the dataset configuration page.


wangq8 (commented):
Thanks. The frontend removal is clean and consistent — schema, interfaces, defaults, component, and locales are all updated together.

Two things to confirm before merge:

1. **Backend still supports `toc_extraction`.** The backend continues to read this flag:
   - `rag/svr/task_executor.py:1616`
   - `rag/svr/task_executor_refactor/task_handler.py:637`
   (plus the `TOC_EXTRACTION` prompts in `rag/prompts/generator.py:577-578`).

   This is a frontend-only removal, so the PageIndex feature still exists on the backend but is no longer configurable in the UI. If PageIndex is being deprecated, is there a follow-up to remove backend support as well? If not, please clarify the intent (e.g., keep it for API-only users).

2. **Silent reset for existing datasets.** For datasets that already have `toc_extraction: true` saved, the next time any dataset setting is edited and saved, the field will be dropped: `useFetchKnowledgeConfigurationOnMount` merges the stored `parser_config` into the form, but since `toc_extraction` is no longer in `formSchema`, zod strips it during `form.handleSubmit`, so `saveKnowledgeConfiguration` sends a `parser_config` without `toc_extraction`. This effectively turns the feature off for existing datasets without any user-facing notice. Please confirm this is intended (and whether a migration/deprecation note is warranted).

Minor: this is a "Feat" that removes a feature — a linked issue or a one-line deprecation rationale in the PR body would help reviewers.

PR #18432, opened 2026-08-18
refactor(extractor): unify 5-in-1 modular schema across dataset and pipeline, drop legacy field_name and custom prompt

jay77721:
## Summary
Refactor the Extractor component into a pure, unified **5-in-1 modular extraction engine** across both Dataset (`knowledgebase.parser_config`) and Pipeline (Canvas DSL).

Following project guidelines (*"Treat legacy code as liability, not as a compatibility target"*), this PR:
1. **Unifies 5-in-1 modular sub-configs** (`keywords`, `questions`, `tags`, `summary`, `metadata`) across both Dataset and Pipeline DSL, dropping legacy flat parameters (`auto_keywords`, `auto_questions`, `auto_tags`, `topn_tags`, `enable_summary`, `enable_metadata`, `metadata_config`, etc.).
2. **Supports modular `system_prompt` fine-tuning** on `keywords`, `questions`, and `summary` sub-configs, falling back gracefully to optimized built-in prompt generators when unconfigured.
3. **Eliminates the ambiguous `field_name` and top-level user `prompt`** (generic arbitrary single-field extraction is deprecated and removed; tasks strictly output to typed Chunk keys).
4. **Preserves Extractor node-configured metadata** in `ApplyComponentScopedParserConfig` when the top-level dataset has no global metadata definitions, preventing node-level metadata from being overwritten on dataset save.
5. **Fuses Summary extraction into the concurrent worker pool** (`extractorPool`), allowing LLM tasks (Keywords, Questions, Summary, Metadata) to execute in parallel across chunks with bounded concurrency and 24h Redis caching.

---

### Canonical 5-in-1 Schema
Both `knowledgebase.parser_config["Extractor:..."]` and Pipeline DSL `components["Extractor:..."]["obj"]["params"]` strictly follow the canonical JSON contract:
```json
{
  "llm_id": "string",
  "keywords": {
    "top_n": 0,
    "system_prompt": "string"
  },
  "questions": {
    "top_n": 0,
    "system_prompt": "string"
  },
  "tags": {
    "top_n": 0,
    "tag_file_id": "string"
  },
  "summary": {
    "enabled": false,
    "system_prompt": "string"
  },
  "metadata": {
    "enabled": false,
    "metadata": [
      {
        "key": "author",
        "type": "string",
        "description": "author name",
        "enum": []
      }
    ],
    "built_in_metadata": [
      {
        "key": "doc_name",
        "type": "string"
      }
    ]
  }
}
```

---

### Key Architectural Changes

1. **Schema & Component Engine (`internal/ingestion/component/`)**:
   - `schema.ExtractorParam`: Stripped `FieldName` and generic top-level prompts. Sub-configs define `Keywords` (`top_n`, `system_prompt`), `Questions` (`top_n`, `system_prompt`), `Tags` (`top_n`, `tag_file_id`), `Summary` (`enabled`, `system_prompt`), `Metadata` (`enabled`, `metadata`, `built_in_metadata`).
   - `ExtractorComponent`: Replaced dual-mode sequential loop with unified lifecycle:
     - Empty chunk guard returns immediately (no artificial single-call empty invocations).
     - Phase 1: Rule-based Tag matching (`runAutoTags`).
     - Phase 2: Concurrent WorkerPool execution (`runAutoKeywords`, `runAutoQuestions`, `runAutoSummary`, `runEnableMetadata`).
   - Cleaned up obsolete placeholder substitution regex engines (`unifiedPlaceholderRE`, `renderExtractorSystemPrompt`).

2. **Pipeline DSL & Template Synchronization (`internal/ingestion/pipeline/`)**:
   - `pipeline_params.go`: Whitelist is derived dynamically from `schema.ExtractorParam` reflection tags + LLM runtime hyper-parameters.
   - Updated all 10 built-in Pipeline DSL templates (`template/*.json`: audio, book, email, general, laws, manual, one, paper, picture, presentation) to the new canonical modular structure.
   - Updated Task Executor golden defaults in `pipeline_executor_defaults_test.go`.

3. **Dataset Service Alignment & Node Metadata Preservation (`internal/service/`)**:
   - `MetadataConfigRequest`: Standardized to `{ enabled, metadata, built_in_metadata }`.
   - `ApplyComponentScopedParserConfig`: Syncs dataset metadata down to Extractor nodes when defined at the top level, and preserves node-configured metadata when the top level is unconfigured.
   - `UpdateMetadataConfig`: Handles metadata persistence and state transitions cleanly.

4. **Dead Code & Legacy Test Removal**:
   - Removed ~20+ obsolete tests protecting deprecated generic prompt templates and flat aliases (including removing `slice3_test.go`).
   - Net deletion: **-1,300+ lines** of boilerplate and legacy shims.

---

### Output Field Mapping (Downstream Indexing Contract)
Extractor directly outputs to standard, strongly-typed Chunk keys consumed by Tokenizer and IndexDoc:
- `important_kwd` & `important_tks` (Keywords) $\to$ BM25 inverted index
- `question_kwd` & `question_tks` (Questions) $\to$ Vector embedding index
- `tag_kwd` & `tag_weight` (Tags) $\to$ Tag filtering
- `summary` (Summary) $\to$ Chunk context enhancement
- `metadata` (Metadata) $\to$ Document metadata index

---

### Verification
All package-scoped unit tests pass with zero regressions:
```bash
bash build.sh --test ./internal/ingestion/... ./internal/service/... ./internal/handler/... ./internal/router/...
```
- `ragflow/internal/ingestion/component` (PASS)
- `ragflow/internal/ingestion/pipeline` (PASS)
- `ragflow/internal/ingestion/task` (PASS)
- `ragflow/internal/service/dataset` (PASS)
- `ragflow/internal/handler` (PASS)
- `ragflow/internal/router` (PASS)


jay77721 (commented):
Thanks for the update — the `883b1052` commit resolves the default-prompt duplication issue nicely. Here are my findings from a full review of the current branch (Go tests + Jest verified locally, all green):

## High priority

**1. `builtInMetadataFromParserConfig` doesn't read the new modular `metadata` shape (regression)**

`internal/ingestion/task/pipeline_executor.go:307` still only reads `node["metadata_config"]` and the flat `node["enable_metadata"]` + `node["built_in_metadata"]`, but `ApplyComponentScopedParserConfig` now writes the canonical shape `Extractor:xxx.metadata = {enabled, metadata, built_in_metadata}` and explicitly deletes `enable_metadata` (component_scoped_parser_config.go:57). I confirmed with a runnable test that the new shape returns `enabled=false, nil` from `builtInMetadataFromParserConfig`. Consequence: **built-in metadata (update_time / file_name) configured on extractor nodes is no longer applied** via `doc_state.applyBuiltInMetadata`, and `AutoMetadataEnabled` reporting is silently wrong. Please add a branch for `node["metadata"].(map[string]any)` + a unit test.

**2. Existing saved flows silently lose features at runtime**

The Go side fully removed legacy parsing (`auto_keywords/auto_questions/auto_tags/enable_summary/enable_metadata/prompts/sys_prompt/metadata_config`) in `NewExtractorComponent`, but the frontend only migrates when a user reopens + re-saves a flow. Any legacy flat-format parser_config in the DB that runs without a re-save (API-triggered parse, scheduled re-index) will silently have keywords/questions/metadata/summary disabled. Also the builtin template `ingestion_pipeline_general.json` still ships the legacy flat extractor params. Consider a one-time normalization at the pipeline entry (with a deprecation warning) or explicitly document the migration requirement.

**3. New TypeScript type errors (type-check fails)**

`FormSchema` dropped flat `tag_file_id`/`enable_summary`, but `extractor-form/index.tsx:250` (`form.watch('tag_file_id')`) and `:379` (`form.setValue('enable_summary', ...)`) still reference them. `transformExtractorParams`'s `if (!params) return {};` widens the return type to `{}`, so `extractor-transform.test.ts:34/38/39` (`result.summary/metadata`) also fail. `npm run type-check` should be part of the gate.

## Medium / low

4. `use-manage-modal.ts:321-325`: `metaData` removed from the effect deps but still referenced — oxlint `exhaustive-deps` warning / stale-closure risk.
5. The reflection whitelist (`llmRuntimeParamKeys`) no longer includes the LLM UI "Enabled" flags (`temperatureEnabled`/`topPEnabled`/...), so they are stripped on round-trip. Go doesn't consume them, so runtime is fine, but the UI switch state is lost across save/load.
6. `transformExtractorConfigToForm` treats a legacy `metadata: [...]` array with no `enable_metadata` as disabled (no array-non-empty fallback, unlike `transformExtractorParams`).

## Still open by design (confirmed acceptable?)

7. With the new message contract, a **custom** system prompt referencing `{text}` still duplicates the chunk body into the user message (system substitution + user always carries chunkText). Defaults are now clean, but worth confirming the token cost/quality implication for placeholder-based prompts is intended.

jay77721 (commented):
## Final review — status: ready to merge after a few small cleanups

The three fix rounds have brought this to a mergeable state. I verified on the final HEAD (`17834d889`): all affected Go packages pass (`internal/ingestion/...`, `internal/service/...`), Jest 8/8 green, and `npm run type-check` has no PR-introduced errors. The previously reported high-priority items are all resolved (modular `metadata` detection in `builtInMetadataFromParserConfig`, legacy flat fallback in `NewExtractorComponent`, TS type errors, `use-manage-modal` deps, LLM `Enabled` UI flags, default-prompt chunk duplication, and the `metadata:v2` cache namespace). Remaining items below are small but worth addressing.

## 1. `prompt`/`prompts` fully removed, but docs and `Inputs()` still reference it (stale contract)

`NewExtractorComponent` no longer parses `prompt`/`prompts`, yet:
- `internal/ingestion/component/extractor.go:232` still documents `"prompt": string — optional user prompt`
- `Inputs()` still declares `"prompt": "Optional user prompt template. Falls back to Param.Prompt when absent."` — but `Param.Prompt` was deleted from `schema.ExtractorParam`
- The `Invoke()` doc comment likewise references `Param.Prompt`

Behavioral impact: if a canvas wires a `prompt` input into an extractor node, it is now silently ignored. For legacy saved nodes that only had a user prompt and no system prompt, the LLM receives the chunk text with **no instruction at all**. Dropping the dead UI control was the right call (the form never exposed it on `main`), but existing custom user-authored instructions are real user data — consider migrating `prompts[0].content` into the relevant system prompt during `transformExtractorConfigToForm` instead of discarding it, and at minimum fix the stale `Inputs()`/`Invoke()` docs.

## 2. Legacy fallback vs `CleanComponentParams` whitelist asymmetry

`NewExtractorComponent` now re-supports legacy keys (`auto_keywords`, `metadata_config`, `enable_metadata`, `sys_prompt`, ...), but the reflection whitelist `extractorValidParamKeys` (`internal/ingestion/pipeline/pipeline_params.go:26`) does not include them. Legacy keys currently survive `BuildParserConfig` only because old DSLs declare them in their `params`, so they pass the `validKeys` check (`pipeline_params.go:99-104`). That is an implicit dependency: if a new frontend-saved DSL no longer declares these keys while the node params still carry leftovers (e.g. `enable_summary` leaked via the `...params` spread in `transformExtractorParams`), they get silently dropped. Suggestion: build the whitelist from a single source that mirrors the keys `NewExtractorComponent` actually parses (`extractJSONTags(ExtractorParam{}) ∪ legacyKeys ∪ llmKeys`), or add an explicit comment documenting the dependency.

## 3. Mixed-precedence (modular + legacy) is now untested

The old `TestExtractor_NestedPrecedenceOverFlat` was removed; the new `TestExtractor_ModularConfiguration` tests modular-only and `TestExtractorComponent_NewExtractorComponent_LegacyParamsFallback` tests legacy-only. The "modular wins" behavior is achieved structurally via `else` branches and is correct, but the mixed case has no regression test, and the flat `enable_metadata` + `metadata[]` branch has none either. A single `TestExtractor_Precedence_ModularBeatsLegacy` would close the gap.

## 4. Minor

- `builtInMetadataFromParserConfig` uses a case-sensitive `strings.HasPrefix(k, "Extractor")`, while the rest of the codebase matches lowercase `extractor:`/`extractor_` — misses DSLs with lowercase component IDs (pre-existing, but easy to align).
- The `use-manage-values-modal.ts` fix (empty-value filtering + timer cleanup) has no unit test guarding the `onSave({...metaData, values: tempValues.filter(...)})` merge semantics.

## Test coverage gap worth noting

There is no end-to-end/integration test covering the full legacy path: stored legacy DSL (e.g. the still-flat `ingestion_pipeline_general.json` template) → `BuildParserConfig`/`CleanComponentParams` → `NewExtractorComponent` fallback → extraction. This is exactly where schema refactors tend to break against real persisted data; a template-driven or `integration`-tagged test would de-risk it.

Overall the refactor is well-executed (single message contract, clean `callRaw/callText/callStructured` split, reflection whitelist, thorough frontend dual-format fallback, and good per-round test additions). Items 1–3 above are small and worth folding in before merge; the rest can be follow-ups.

jay77721 (commented):
## Round 4 review — one high-priority regression to fix before merge

The three prior fix rounds have landed well: all previously reported items are resolved and verified (stale `prompt` docs/`Inputs()` contract, user-prompt migration into `summary.system_prompt`, whitelist/legacy-key asymmetry, mixed-precedence test, lowercase `extractor:`/`extractor_` prefix in `builtInMetadataFromParserConfig`, `use-manage-values-modal` coverage, and the legacy end-to-end flow test). Go suites, Jest (28/28) and type-check are all clean for PR-touched files.

However, the latest commit `250e7ace3` refactor of `ApplyComponentScopedParserConfig` leaves one **high-priority regression: extractor-node-only metadata fields are silently wiped on dataset save**.

## Regression: node-configured metadata fields are discarded

**Trigger scenario (real frontend path):** a user configures metadata (custom + built-in fields) on the Extractor canvas node, while the dataset has no top-level metadata config. The frontend only writes the node-level `metadata` object (`transformExtractorParams`) and never promotes it to top-level.

**Chain** (`internal/service/dataset/update.go:249 → 250 → 302`):
1. `preserveDatasetParserConfigMetadata` (`internal/service/dataset/helpers.go:275-296`) derives top-level `enable_metadata = true` from the node's explicit `metadata.enabled = true`.
2. In `ApplyComponentScopedParserConfig` (`internal/service/component_scoped_parser_config.go:63`), the node-field fallback is gated on `if !hasMetadataConfig`. Since the derived top-level `enable_metadata` makes `hasMetadataConfigShape` return true, **the node's own fields are never consulted**.
3. The explicit-`enabled` branch (lines 74-82) then writes `{enabled: true, metadata: <top-level, empty>, built_in_metadata: <top-level, empty>}`, replacing the node's configured fields with nothing.

**Reproduced with a runnable test** (temp test removed after):
```
node metadata after apply: {enabled:true, metadata:[], built_in_metadata:[]}
REGRESSION: node-configured custom field lost
REGRESSION: node-configured built-in field lost
```

**Impact:** `runEnableMetadata` short-circuits on empty fields, so auto-metadata extraction silently produces nothing even though the user enabled and configured it — a core feature of this PR.

Note: the symmetric case ("top-level has fields + node explicitly disabled") is covered by `TestApplyComponentScopedParserConfig_PreservesExplicitDisabledState`, but the "**node has fields + top-level has none**" side is completely untested — every existing test assumes top-level metadata is configured.

**Suggested fix:** gate the node-field fallback on whether the top-level actually provides field definitions (`len(rawMetaSlice) == 0 && len(rawBuiltInSlice) == 0`) instead of `!hasMetadataConfig`, with node fields preferred and top-level used as a complement. Add a test that replicates the preserve → apply chain for the node-only case.

Everything else reads clean. Once this is fixed (with its regression test), I'd consider the PR ready to merge.

jay77721:
## Code review: regression-focused analysis

I read the full diff and the companion web PR #18446, and ran the affected Go package tests locally (`component`, `pipeline`, `task`, `service`, `service/dataset`) — all green. Overall the refactor is a good direction (single message contract, modular configs, v2 cache namespace), but I found several regression risks worth resolving before merge.

### R1 (high) — `prompt`/`prompts` are silently dropped, not "retained for backward compatibility"

- `NewExtractorComponent` no longer reads `prompt`/`prompts` (`extractor.go:250-262`), `resolveInputs` no longer reads `inputs["prompt"]`, and the new `CleanComponentParams` whitelist (`llmRuntimeParamKeys` + `extractorLegacyParamKeys`, `pipeline_params.go:16-44`) does **not** include `prompt`/`prompts`, so they are stripped.
- However `Inputs()` still documents them as `"Deprecated user prompt template (retained for backward compatibility)."` (`extractor.go:399`) — the doc and the code contradict each other.
- Impact: existing DSL node params carrying `prompt` are silently ignored on the **next ingestion run** (that path constructs the component directly without `CleanComponentParams`), with no warning. Users' custom extraction instructions quietly disappear and LLM output changes.
- Suggested fix: on encountering `prompt`/`prompts` in `NewExtractorComponent`, migrate the content into `SystemPrompt` (under the new contract the system message carries instructions and the user message only carries chunkText, so this is semantics-preserving) and emit a `common.Warn` deprecation notice. This gives existing pipelines a lossless one-time migration instead of silent data loss.

### R2 (high) — existing multi-Extractor pipelines hard-fail after upgrade, with no migration path

- `ValidatePipeline` is called unconditionally from `NewPipelineFromDSL` (`pipeline.go:156,179-192`), and `ExtractAllComponentParams` returns an error for the same condition (`component_params.go:51`). The old frontend allowed multiple Extractor nodes.
- Impact: a previously-saved canvas with 2+ Extractor nodes fails at the **next ingestion task construction**, and dataset updates that resolve component-param defaults (`service/dataset/crud.go:90`) also fail. The web PR (#18446) only prevents *new* adds/duplicates; nothing migrates or repairs existing pipelines.
- Also inconsistent: `CleanComponentParams` swallows this same error and returns the input unchanged (`pipeline_params.go:104-106`) while `ValidatePipeline` hard-fails — the same condition behaves differently on two paths.

### R3 (medium) — custom `system_prompt` containing `{text}` now injects the chunk text twice

- Per-chunk path renders `{text}` into the system message (`renderExtractorSystemPrompt`, `extractor.go:1666`) **and** `buildExtractorMessages` unconditionally puts chunkText in the user message (`extractor.go:1623`). With `system_prompt: "Summarize: {text}"` the model now sees the body twice.
- The old `bodyInjected` suppression (text appears exactly once) was removed together with its regression tests (`TestExtractorComponent_Invoke_SystemPromptPlaceholderSuppressesAppend` and the "no duplicate append" assertions). The new tests assert the duplication explicitly, so this is a deliberate contract change — but it is not called out in the PR description, and it changes output quality/token usage for existing custom templates. Default templates (now without the `%s` body) are unaffected.

### R4 (medium) — `builtInMetadataFromParserConfig` top-level fallbacks hard-code `enabled=true`

- The new top-level `metadata`/`metadata_config`/`built_in_metadata` fallback branches all `return arr, true` (`pipeline_executor.go:338-350`); previously `metadata_config`/`built_in_metadata` were `enable_metadata`-aware.
- Impact: with a zero-Extractor pipeline that has top-level `built_in_metadata` and `enable_metadata=false` (user explicitly disabled auto-metadata), `applyBuiltInMetadata` (`doc_state.go:70`) now overwrites `update_time`/`file_name`. The node-level `metadata.enabled` branch stays enabled-aware while the top-level branches do not — inconsistent within the same function, and the new test only asserts the new `true` behavior, not the disabled case.

### R5 (positive) — v2 metadata cache namespace is correct

Changing the message contract invalidates old cache entries; the `kc:meta:v2:` namespace (`metadataLLMCacheKey`, `extractor.go:1071-1080`) is the right call to avoid stale results. Only cost is a 24h cache warm-up after deploy.

### Frontend (#18446) completeness

Aligned: modular `metadata` payload, `prompts` removed from the form, single-Extractor restriction on add/duplicate/operator menu, `builtInMetadata` + `built_in_metadata` both sent, modular DSL output.

Gaps:
- **Lossy prompt migration for non-summary field names**: `transformExtractorConfigToForm` maps legacy `prompts`/`prompt` into `summary.system_prompt` (`pipeline-operator.ts:804-816`). A legacy config with `field_name="answer"` + custom prompt ends up with the prompt parked on a *disabled* summary subtask — the actual field extraction loses its instructions.
- **Narrow migration window**: the backend drops `prompt` on the next DSL re-save, so the frontend's read of `config.prompts` only works if the user opens the form before the DSL is rewritten. No one-time background migration exists.

### Test coverage

Good: legacy flat fallback parsing, end-to-end legacy flow, modular-beats-legacy precedence, message assembly, system-prompt rendering table, whitelist + normalize, multi-Extractor rejection, `ApplyComponentScopedParserConfig` scenarios, and the `UpdateMetadataConfig` enable/disable/conflicting-nodes matrix.

Gaps:
1. No test for the `prompt` removal path (old tests were deleted rather than rewritten to assert ignore-or-warn).
2. No test pinning the intended duplicate-vs-single occurrence behavior for body placeholders.
3. No tests for the graceful-degradation paths of the multi-Extractor error (`CleanComponentParams` no-op, `ExtractParserCpnID` empty, `warnUnknownComponentParams` warn-only).
4. No test for `builtInMetadataFromParserConfig` with `enabled=false`.
5. No pinning test for the reflection-derived whitelist (guards against silent drift — which is exactly how `prompt` got dropped).
6. Unit-only; no integration/e2e exercising the v2 cache + real LLM contract.

### Suggested improvements

1. Lossless backend migration for `prompt`/`prompts` → `SystemPrompt` with a deprecation warning (resolves R1 and the frontend's F1 lossiness in one move).
2. Soft-landing for existing multi-Extractor DSLs (warn + run only the first extractor) or an explicit one-time migration path, before hard-failing.
3. Make the top-level `builtInMetadataFromParserConfig` fallbacks enabled-aware like the node-level branch (resolves R4).
4. Add a whitelist pinning test.
5. Either keep single-occurrence suppression for body placeholders or clearly document the deliberate duplication (R3) in the PR description with a matching test.

internal/ingestion/component/schema/extractor.go
```diff
@@ -76,104 +76,50 @@ func (ExtractorFromUpstream) Validate() error { return nil }
 
 // KeywordExtractConfig configures automatic keyword extraction.
 type KeywordExtractConfig struct {
-	TopN         int    `json:"top_n,omitempty"`
-	SystemPrompt string `json:"system_prompt,omitempty"`
+	TopN int `json:"top_n"`
```
yuzhichang: Add comment `0 means disabled`.
Add back SystemPrompt.

internal/ingestion/component/schema/extractor.go
```diff
@@ -76,104 +76,50 @@ func (ExtractorFromUpstream) Validate() error { return nil }
...
 }
 
 // QuestionExtractConfig configures automatic question generation.
 type QuestionExtractConfig struct {
-	TopN         int    `json:"top_n,omitempty"`
-	SystemPrompt string `json:"system_prompt,omitempty"`
+	TopN int `json:"top_n"`
```
yuzhichang: Add comment `0 means disabled`.
Add back SystemPrompt.

internal/ingestion/component/schema/extractor.go
```diff
@@ -76,104 +76,50 @@ func (ExtractorFromUpstream) Validate() error { return nil }
...
-	TopN      int    `json:"top_n,omitempty"`
+	TopN      int    `json:"top_n"`
 	TagFileID string `json:"tag_file_id,omitempty"`
 }
 
 // SummaryExtractConfig configures summary / enhanced context extraction.
 type SummaryExtractConfig struct {
-	Enabled      bool   `json:"enabled,omitempty"`
-	SystemPrompt string `json:"system_prompt,omitempty"`
+	Enabled bool `json:"enabled"`
```
yuzhichang: Add back SystemPrompt.

internal/ingestion/component/schema/extractor.go
```diff
@@ -76,104 +76,50 @@ func (ExtractorFromUpstream) Validate() error { return nil }
...
 type QuestionExtractConfig struct {
-	TopN         int    `json:"top_n,omitempty"`
-	SystemPrompt string `json:"system_prompt,omitempty"`
+	TopN int `json:"top_n"`
 }
 
 // TagExtractConfig configures automatic tag extraction.
 type TagExtractConfig struct {
-	TopN      int    `json:"top_n,omitempty"`
+	TopN      int    `json:"top_n"`
```
yuzhichang, line 91: Add back SystemPrompt.

internal/ingestion/component/schema/extractor.go
```diff
@@ -76,104 +76,50 @@ func (ExtractorFromUpstream) Validate() error { return nil }
...
 type SummaryExtractConfig struct {
-	Enabled      bool   `json:"enabled,omitempty"`
-	SystemPrompt string `json:"system_prompt,omitempty"`
+	Enabled bool `json:"enabled"`
 }
 
 // MetadataExtractConfig configures structured metadata extraction.
 type MetadataExtractConfig struct {
-	Enabled         bool                      `json:"enabled,omitempty"`
+	Enabled         bool                      `json:"enabled"`
```
yuzhichang, line 103: Add back SystemPrompt.

PR #18433, opened 2026-08-18
fix: clear existing volume before restore

LZY-SPCA:
## What does this PR do?

Clear the contents of an existing Docker volume before restoring backup data.

Currently, `migration.sh restore` extracts the backup archive directly into an existing volume. `tar xzf` overwrites files that exist in the archive, but it does not remove files that exist only in the target volume.

As a result, restoring into an existing volume performs a merge instead of an exact restore.

This can cause problems for stateful services such as MySQL. In my case, stale InnoDB redo log files remained after restore and caused MySQL to fail during startup with errors such as:

```text
[ERROR] [InnoDB] Missing redo log file
[ERROR] [InnoDB] Plugin initialization aborted with error Generic error
[ERROR] Failed to initialize DD Storage Engine
[ERROR] Data Dictionary initialization failed
[ERROR] Aborting

### DIFF
--- a/docker/migration.sh
+++ b/docker/migration.sh
@@ -274,7 +274,10 @@ perform_restore() {
             echo "  📋 Creating Docker volume: $volume"
             docker volume create "$volume"
         else
-            echo "  📋 Using existing Docker volume: $volume"
+            echo "  📋 Clearing existing Docker volume: $volume"
+            docker run --rm \
+                -v "$volume":/target \
+                alpine sh -c "find /target -mindepth 1 -delete"
         fi
 
         # Restore data


myf-bee (commented):
It is recommended that the volume cleanup operation be performed by the user themselves. Instructions for this operation are provided below.

asiroliu:
@LZY-SPCA 
I think directly deleting the contents of an existing volume is too risky.

I suggest that during restore, if the target volume already exists, we should fail with an error instead of deleting or overwriting its contents. The error message can provide the command for manually removing the volume, leaving that decision explicitly to the user.

LZY-SPCA:
I agree that automatically clearing an existing volume during `restore` may be too destructive.

I think there are two safer alternatives:

1. Update the migration documentation to explicitly require restoring into empty volumes. For example, the documented restore flow could include:

   * stop the RAGFlow services;
   * remove the existing volumes if the user intends to fully replace the current data;
   * run `docker compose create` to create empty volumes managed by Docker Compose without starting the services;
   * run `migration.sh restore`;
   * then start RAGFlow with `docker compose up -d`.

2. Alternatively, `migration.sh restore` could detect that a target volume already exists and abort with a clear error message instead of extracting the backup on top of the existing contents. The message could instruct the user to remove or clear the existing volume before retrying the restore.

The main issue I would like to avoid is the current behavior where `tar xzf` restores the backup into an existing non-empty volume. This performs a merge rather than an exact restore, because files that are not present in the backup remain in the volume. For stateful services such as MySQL or Elasticsearch, this can leave stale files and cause consistency problems.

I am fine with either updating the documentation or changing the restore behavior to reject non-empty/existing volumes. Both approaches would be safer than silently merging the backup with existing volume contents.

myf-bee:
I agree to address this issue first via documentation in this PR#18626. Thank you!

PR #18435, opened 2026-08-18
Fix pdf cancel at task level not document level, multiple record logs

wangq8:
Cause:
- PDFs are split into page-range tasks, normally 12 pages each.
- Cancellation immediately marks the document final.
- Each remaining page task then created its own pipeline_operation_log.
- Thus a PDF split into five tasks could produce five cancellation records.

Fix:
- Log creation is now idempotent for the same document ingestion run.
- A document-row lock prevents concurrent task finalizers from inserting duplicates.

### DIFF
--- a/api/db/services/pipeline_operation_log_service.py
+++ b/api/db/services/pipeline_operation_log_service.py
@@ -234,14 +234,36 @@ class PipelineOperationLogService(CommonService):
         log["update_time"] = timestamp
         log["update_date"] = datetime_now
         with DB.atomic():
-            Document.select(Document.id).where(Document.id == document.id).for_update().first()
-            if document_id != GRAPH_RAPTOR_FAKE_DOC_ID and process_begin_at is not None:
+            operation_status_value = operation_status.value if isinstance(operation_status, TaskStatus) else str(operation_status)
+            if document_id != GRAPH_RAPTOR_FAKE_DOC_ID and operation_status_value == TaskStatus.CANCEL.value:
+                # Serialize page-task finalizers for the same canceled document.
+                locked_document = Document.select(Document.id, Document.run, Document.update_time).where(Document.id == document.id).for_update().first()
+                if locked_document is None or locked_document.run != TaskStatus.CANCEL.value:
+                    return None
+                cancel_update_time = locked_document.update_time or timestamp
+                if locked_document.update_time is None:
+                    Document.update(update_time=cancel_update_time, update_date=datetime_now).where(Document.id == locked_document.id).execute()
+                pipeline_filter = cls.model.pipeline_id == pipeline_id if pipeline_id else (cls.model.pipeline_id.is_null(True) | (cls.model.pipeline_id == ""))
                 existing = (
                     cls.model.select()
-                    .where((cls.model.document_id == document_id) & (cls.model.pipeline_id == pipeline_id) & (cls.model.task_type == task_type) & (cls.model.process_begin_at == process_begin_at))
+                    .where(
+                        (cls.model.document_id == document_id)
+                        & pipeline_filter
+                        & (cls.model.task_type == task_type)
+                        & (cls.model.operation_status == TaskStatus.CANCEL.value)
+                        & (cls.model.create_time >= cancel_update_time)
+                    )
                     .first()
                 )
                 if existing:
+                    logging.debug(
+                        "Skip duplicate pipeline operation log document_id=%s pipeline_id=%s task_type=%s process_begin_at=%s existing_id=%s",
+                        document_id,
+                        pipeline_id,
+                        task_type,
+                        process_begin_at,
+                        existing.id,
+                    )
                     return existing
 
             obj = cls.save(**log)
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix — the diagnosis is correct, and the row-lock + idempotency check is a sound approach for collapsing the multiple page-range task finalizers into a single record. A few things worth tightening before merge:

## 1. Dedup key relies on second-precision `process_begin_at` (possible false-positive dedup)

The uniqueness of the dedup key `(document_id, pipeline_id, task_type, process_begin_at)` depends on `process_begin_at` being distinct per ingestion run. It is set in `begin2parse` via `get_format_time()`, which strips microseconds (second precision):

```python
def get_format_time() -> datetime.datetime:
    return datetime_format(datetime.datetime.now())  # microseconds -> 0
```

So a cancel-then-reparse that happens within the same wall-clock second would produce an identical `process_begin_at`, and the new run's log would be deduped into the previous (canceled) run's record instead of creating a fresh one. In practice this is a narrow race, but it silently corrupts the operation log history.

Suggest using a stronger run identity (e.g. the run's task id / a run token / keep microsecond precision) as part of the key, or confirming that `process_begin_at` is guaranteed to advance between runs.

## 2. Document row lock is taken even when dedup is skipped (KB-level fan-out tasks)

The `for_update()` lock is acquired unconditionally, but for `document_id == GRAPH_RAPTOR_FAKE_DOC_ID` (KB-level fan-out tasks: GraphRAG/RAPTOR/MindMap/etc.) the dedup check is skipped. For those tasks `document.id` is the *referent* document (`referred_document_id`), so concurrent KB-level finalizers that share the same referent document now serialize on that row with no benefit — added contention and a potential (if unlikely) deadlock surface.

Consider only taking the lock when the dedup path actually applies, e.g.:

```python
if document_id != GRAPH_RAPTOR_FAKE_DOC_ID and process_begin_at is not None:
    Document.select(Document.id).where(Document.id == document.id).for_update().first()
    existing = ...
    if existing:
        return existing
```

## 3. `process_begin_at is None` silently disables the fix

The guard `process_begin_at is not None` means that if a document ever reaches finalization with a null `process_begin_at`, the dedup is skipped and the original duplicate-cancellation bug re-appears. I believe `begin2parse` always sets it before page tasks are queued, but worth an explicit confirmation/comment that this state is unreachable at cancellation time (or a fallback key when it is null).

## 4. Missing test coverage

No tests are included. Given the repo has `test/unit_test/...` coverage for the surrounding services, a small unit test asserting that a second `create(...)` call with the same `(document_id, pipeline_id, task_type, process_begin_at)` returns the existing record (and does not insert a duplicate) would lock in this behavior and guard against regressions.

## Nit

The `for_update().first()` result is discarded — it's purely a lock. A one-line comment (e.g. `# serialize concurrent task finalizers for the same document`) would make the intent clear for future readers.

wangq8 (commented):
Re-reviewed the updated patch. The rework is a clear improvement over the previous version — scoping the lock to the CANCEL branch fixes the earlier "unnecessary lock for KB-level fan-out tasks" concern, and using `create_time >= cancel_update_time` to bound the dedup window is much better than keying on second-precision `process_begin_at`. A few remaining issues:

## 1. `update_time` is not a semantic cancel marker (medium)

`cancel_update_time` is taken from `Document.update_time`, but that column is a generic "last modified" timestamp — `BaseModel._normalize_data` overwrites it with `current_timestamp()` on *every* insert/update of the document row. The dedup window `create_time >= cancel_update_time` only works because, in today's code, nothing touches a canceled document afterward (`_sync_progress` `continue`s on `run == CANCEL`). That's coincidental, not guaranteed:

- If any write lands on a canceled document *after* the first cancel log is inserted, `update_time` advances and the next page-task finalizer computes a later bound, no longer matching the already-inserted log → duplicate cancel records come back.
- Conversely the bound is only correct because the cancel path happens to leave the row alone.

Suggest a dedicated cancellation timestamp/run-id on the document (or reusing a stable per-run identifier) rather than piggybacking on `update_time`.

## 2. The `update_time is None` fallback is misleading (low)

```python
cancel_update_time = locked_document.update_time or timestamp
if locked_document.update_time is None:
    Document.update(update_time=cancel_update_time, update_date=datetime_now)...
```

`Document.update(...)` routes through `_normalize_data`, which unconditionally sets `update_time = current_timestamp()`, so the explicit `update_time=cancel_update_time` is silently discarded and the persisted value won't match `cancel_update_time`. Since `update_time` is effectively never `None` (set on every insert/update), this branch is near-dead code — but if it's meant to stamp the cancel time, it doesn't do that. Either remove it or fix it to actually persist `cancel_update_time` (and then be aware of the point above).

## 3. `return None` when `locked_document.run != CANCEL` — confirm intent

This drops the log entirely (not just dedup) when the document has been re-run between the initial read and lock acquisition. That's reasonable (don't write a CANCEL log for a doc that's now RUNNING), but worth a comment, since `create()` returning `None` has different downstream meaning than "no-op" and callers in `task_executor.py`/`dataflow_service.py` record the return value.

## 4. Still no test coverage

The concurrency + dedup logic is the kind of thing that regresses silently. A unit test (e.g. call `create(...)` twice with the same `(document_id, pipeline_id, task_type)` for a CANCEL doc and assert the second returns the existing record without inserting) would lock this in. `test/unit_test/...` already has scaffolding for these services.

## Nit

`operation_status_value` is recomputed here even though `operation_status` was already normalized a few lines above (`_is_final_state` does the same `isinstance`/`str` dance). Minor duplication; fine to leave, but a shared helper would avoid the two drifting apart.

PR #18436, opened 2026-08-18
fix(web): strip think sections from floating chat widget copy

dcc123456:
### Summary

fix(web): strip think sections from floating chat widget copy

The floating chat widget copied the raw assistant message, so reasoning content wrapped in <think> tags ended up in the clipboard. Reuse removeThinkSection, the same helper the chat message toolbar uses, for both the embedded and expanded widget copy buttons.


wangq8 (commented):
Looks good. Reusing `removeThinkSection` keeps copy behavior consistent with the chat message toolbar (group-button.tsx), and covers both the embedded and expanded widget copy buttons. One minor note: `removeThinkSection` calls `.replace()`/`.trim()` on its input, so it would throw if `message.content` were null/undefined — but content is typed as `string` and rendered directly, so this is fine in practice. LGTM.

PR #18437, opened 2026-08-18
Fix: verification error msg

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
This is a trivial, low-risk wording change. Note that the guarded condition `len(passedTypes) == 0` literally means every verification attempt failed, so the original "all" was technically accurate — removing it is a cosmetic choice, not a correctness fix. Fine to merge, but the PR body ("As title.") would benefit from a one-line explanation of the motivation. LGTM.

PR #18439, opened 2026-08-18
Fix multiple model chat always use the chat model, not using the passing model

wangq8:


wangq8 (commented):
Thanks for the fix! I reviewed the change and the root cause is correct.

**Analysis**

In `session_completion`, when a `chat_model_id` (`llm_id`) is passed, only `dia.llm_id` is overwritten while `dia.tenant_llm_id` keeps its persisted value. Downstream, every model-resolution path prefers the tenant model id when both fields are set:

- `get_models` (used by `rag_agent` for reasoning paths): `if dialog.tenant_llm_id: chat_model_config = get_model_config_by_id(... tenant_llm_id ...)`
- `async_chat`: `if dialog.tenant_llm_id: ... get_model_config_by_id(... tenant_llm_id ...)`
- `async_chat_solo`: same `tenant_llm_id`-first branch

So for a "multiple model" chat whose `tenant_llm_id` is set, the passed model is silently ignored. Setting `dia.tenant_llm_id = None` here forces resolution to go through `resolve_model_config(... dialog.llm_id ...)`, which is the intended behavior.

A couple of notes/requests:

1. **No persistence side effect** — `dia` is the in-memory object returned by `DialogService.get_by_id`, and the assignment isn't saved, so this only affects the current request. Good. But worth a one-line comment in code explaining *why* `tenant_llm_id` must be cleared, otherwise a future refactor may drop it as "redundant".

2. **Same bug exists in the OpenAI-compatible endpoint** — `openai_chat_completions` in `api/apps/restful_apis/openai_api.py` also does `dia.llm_id = requested_model` without clearing `dia.tenant_llm_id`. A chat configured with a tenant model will likewise ignore `model` in the request body. Consider applying the same fix there for consistency (or confirm it's intentionally out of scope).

3. **Passing a tenant-model id still works after the fix** — since `resolve_model_config` first tries `get_model_config_by_id` (id lookup), passing a `tenant_llm` UUID as `chat_model_id` still resolves correctly, so no regression there.

4. **Tests** — this is a small, low-risk fix, but a regression test covering "chat with tenant_llm_id + explicit llm_id in request uses the requested model" would prevent it from silently coming back.

LGTM otherwise.

PR #18440, opened 2026-08-18
Fix: Initiate a request to check for file changes two seconds after compilation completes.

cike8899:
### Summary

Fix: Initiate a request to check for file changes two seconds after compilation completes.

### DIFF
--- a/web/src/pages/dataset/compilation/dataset-structure-view.tsx
+++ b/web/src/pages/dataset/compilation/dataset-structure-view.tsx
@@ -14,6 +14,7 @@ import {
   useTraceRunData,
 } from '@/hooks/use-dataset-generate';
 import {
+  ArtifactAlterationKeys,
   DatasetStructureKeys,
   useDeleteDatasetStructure,
   useFetchArtifactAlteration,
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM overall. The debounce correctly defers the refetch until after the compilation run settles, and invalidating `ArtifactAlterationKeys.detail` is the right fix so the "file changes" count refreshes on run end. `useDebounceFn` returns a stable `run` (memoized in ahooks), so the effect dependency `[status, scheduleRunEnd]` won't re-fire spuriously, and all three consumers of `useRunEndEffect` (structure view, skills view, llm-wiki view) only call `invalidateQueries`, which is safe to debounce and safe after unmount.

Two minor nits (non-blocking):
1. The comment says "Debounced to fire 1s after the run settles" but `wait` is 2000ms. Please align the comment with the actual 2s delay (or vice versa).
2. There's a trailing-whitespace blank line added after the new `invalidateQueries` block in `dataset-structure-view.tsx` (the line contains only spaces). Worth trimming to keep lint clean.

PR #18442, opened 2026-08-18
Fix: set enable_thinking to True for qwen3.8-2.4t-a95b

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
LGTM. The change is scoped correctly: the `2.4t-a95b` check sits inside the existing `qwen3` family branch, so it can only ever match Qwen3 models, consistent with the existing `-preview` handling. The two new tests cover both the default and `thinking=disabled` cases and match the established patterns in `test_chat_model_thinking_policy.py`. No regressions spotted.

Minor (non-blocking): the magic string `"2.4t-a95b"` is a bit brittle, but it's consistent with how `"-preview"` is already handled here, so it's acceptable. Consider a follow-up if more qwen3 flagships needing force-enable appear.

PR #18445, opened 2026-08-18
fix: preserve empty chat fields on GaussDB

Sevenzuo:
### Summary

On A/ORA-compatible GaussDB, empty strings are stored as `NULL`. A newly created Chat leaves `Dialog.icon` and `Dialog.description` empty, so reading the record returned `null`. Chat settings then rejected both optional fields with `Expected string, received null`.

This PR changes those two fields to the existing `EmptyStringTextField`. GaussDB may keep storing `NULL`, while the ORM/API restores the application-level empty string expected by the frontend.

### Compatibility

MySQL and PostgreSQL behavior is unchanged: outside compatible GaussDB, `EmptyStringTextField` retains normal Peewee `TextField` conversion and the existing nullable schema. No database migration is required because both columns are already nullable.

### Tests

- Added focused coverage for GaussDB, MySQL, and PostgreSQL empty-string/null conversion.
- Verified against a live GaussDB deployment: raw `(NULL, NULL)` values are returned by the ORM as `("", "")`.
- Targeted test passed: `test_gaussdb_empty_string_compatible_fields_are_nullable_only_for_gaussdb`.

### Symptom

Chat settings validation error

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -1442,8 +1442,8 @@ class Dialog(DataBaseModel):
     id = CharField(max_length=32, primary_key=True)
     tenant_id = CharField(max_length=32, null=False, index=True)
     name = CharField(max_length=255, null=True, help_text="dialog application name", index=True)
-    description = TextField(null=True, help_text="Dialog description")
-    icon = TextField(null=True, help_text="icon base64 string")
+    description = EmptyStringTextField(null=True, help_text="Dialog description")
+    icon = EmptyStringTextField(null=True, help_text="icon base64 string")
     language = CharField(max_length=32, null=True, default="Chinese" if "zh_CN" in os.getenv("LANG", "") else "English", help_text="English|Chinese", index=True)
     # Map application-level empty chat/reranker model IDs to storage NULL in the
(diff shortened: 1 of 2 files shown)


Lynn-Inf (approved):
LGTM. No Go changes needed — the Dialog table is not mapped in Go.

Lynn-Inf:
@Sevenzuo A correction to my earlier comment: the `Dialog` table in Python corresponds to `Chat` in the Go codebase. So when you add GaussDB support in Go, please make sure to update `entity.Chat.Icon` and `entity.Chat.Description` accordingly in that follow‑up PR.

Sevenzuo:
> @Sevenzuo A correction to my earlier comment: the `Dialog` table in Python corresponds to `Chat` in the Go codebase. So when you add GaussDB support in Go, please make sure to update `entity.Chat.Icon` and `entity.Chat.Description` accordingly in that follow‑up PR.

Got it.

Sevenzuo:
@wangq8 Could you please take a look and merge this PR when convenient? All checks have passed. Thanks!

Issue #18394, opened 2026-08-17
[Bug]: DeepL agent tool is non-functional — no get_meta (crashes Agent) and only implements the dead _run path


immuhammadfurqan:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

68f826082

### RAGFlow image version

v0.26.4

### Other environment information

- Not environment specific: reproduces from a clean checkout of `main`.

### Actual behavior

The **DeepL** agent tool (`agent/tools/deepl.py`) is non-functional. It was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it cannot be used as an Agent tool.

`DeepL` extends `ComponentBase` (not `ToolBase`) and `DeepLParam` extends `ComponentParamBase` with no `meta`. This breaks in two independent ways.

**1. Adding it to an Agent raises `AttributeError`.**

`agent/component/agent_with_tools.py` builds each tool's function descriptor via `get_meta()`:

```python
for idx, cpn in enumerate(self._param.tools):
    cpn = self._load_tool_obj(cpn)
    original_name = cpn.get_meta()["function"]["name"]   # DeepL has no get_meta()
```

`get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` — `grep -n "def get_meta" agent/component/base.py` returns nothing. So constructing an Agent that includes the DeepL tool raises:

```
AttributeError: 'DeepLParam' object has no attribute 'get_meta'
```

**2. It can never run — `_run` is dead code.**

`DeepL` only implements the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more; every execution path goes through `_invoke`:

- `agent/component/base.py:406` — `self._invoke(**kwargs)`
- `agent/component/base.py:432` — `await self._invoke(**kwargs)`
- `agent/component/base.py:590` — `return self._invoke(**kwargs)` (`debug`)
- `agent/tools/base.py:154` — `res = self._invoke(**kwargs)`
[...]

Issue #16416, opened 2026-06-27
[Bug]: AkShare agent tool is non-functional — no get_meta (crashes Agent) and uses obsolete _run/be_output


immuhammadfurqan:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug

The **AkShare** agent tool (`agent/tools/akshare.py`) is non-functional. It was never ported to the modern `ToolBase`/`_invoke` interface during the agent module redesign and is still written against the removed legacy `_run`/`be_output` API, so it cannot be used as an Agent tool.

### Root cause

`AkShare` extends `ComponentBase` (not `ToolBase`) and `AkShareParam` extends `ComponentParamBase`, with no `meta`. This breaks in two ways:

**1. Adding it to an Agent raises `AttributeError`.**

`agent/component/agent_with_tools.py` builds each tool's function descriptor via `get_meta()`:

```python
for idx, cpn in enumerate(self._param.tools):
    cpn = self._load_tool_obj(cpn)
    original_name = cpn.get_meta()["function"]["name"]   # AkShare has no get_meta()
```

`get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase`. So as soon as an Agent containing an AkShare tool is constructed, it raises `AttributeError: 'AkShare' object has no attribute 'get_meta'`.

**2. Even if invoked directly, it does nothing useful.**

`ToolBase.invoke()`/`ComponentBase.invoke()` dispatch to `self._invoke(**kwargs)`, but `AkShare` only implements the legacy `_run(self, history, **kwargs)`, so `_invoke` falls through to `ComponentBase._invoke` which `raise NotImplementedError()`. The `_run` body also calls `AkShare.be_output(...)`, which no longer exists on the base classes (it was removed in the redesign), so the legacy code path is dead as well.

`git log` shows `akshare.py` has not been touched since the agent redesign (`d9fe279dd`), so it has been broken the whole time.

### Expected behavior

- The AkShare tool exposes a proper tool descriptor (`get_meta()`) with a `query` parameter (the stock symbol to look up).
- Invoking it fetches the latest East Money (东方财富) news for that symbol via `akshare` and writes the result to the component output instead of the removed `be_output`.

### Steps to reproduce

1. Build an agent whose Agent component has the AkShare tool attached.
2. Run it — construction fails with `AttributeError: 'AkShare' object has no attribute 'get_meta'`.

### Additional information

[...]

PR #18449, opened 2026-08-18
fix(agent/tools): port Jin10 to ToolBase so it works as an Agent tool

### LINKED ISSUE #18447: [Bug]: Jin10 agent tool is non-functional — no get_meta (crashes Agent) and only implements the dead _run path
### Self Checks

68f826082

### RAGFlow image version

v0.26.4

### Actual behavior

The **Jin10** agent tool (`agent/tools/jin10.py`) is non-functional. It was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it cannot be used as an Agent tool.

`Jin10` extends `ComponentBase` (not `ToolBase`) and `Jin10Param` extends `ComponentParamBase` with no `meta`. This breaks in two independent ways.

```python
    original_name = cpn.get_meta()["function"]["name"]   # Jin10 has no get_meta()
```

`get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` — `grep -n "def get_meta" agent/component/base.py` returns nothing. So constructing an Agent that includes the Jin10 tool raises:

```
AttributeError: 'Jin10Param' object has no attribute 'get_meta'
```

`Jin10` only implements the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more; every execution path goes through `_invoke`:

`grep -rn "\._run(" agent/` returns no callers. Since `Jin10` defines no `_invoke`, it falls through to `ComponentBase._invoke`, which raises `NotImplementedError` (`agent/component/base.py:447`).

**Additionally, the tool's only use of its input is a no-op gate.** `_run` computes `ans` from the upstream input and returns early when it is empty:

```python
ans = self.get_input()
ans = " - ".join(ans["content"]) if "content" in ans else ""
if not ans:
    return Jin10.be_output("")
```

`ans` is then never referenced again — every request is built purely from `self._param`. So the upstream input gates execution without influencing it.

The tool is user-reachable, not dead code: `agent/tools/__init__.py` auto-imports every module in the directory and registers classes by name, and the frontend ships `Jin10TypeOptions`, `Jin10FlashTypeOptions`, `Jin10CalendarTypeOptions`, `Jin10SymbolsTypeOptions` and related dropdowns in `web/src/pages/agent/options.ts` plus locale strings in ~12 languages.

### Expected behavior

- The Jin10 tool exposes a proper tool descriptor via `get_meta()` with a `query` parameter, so an Agent's LLM can call it.
- Invoking it fetches the configured Jin10 dataset (flash / calendar / symbols / news), writes the result to `formalized_content`, and surfaces failures via `_ERROR` instead of the legacy `be_output` DataFrame.
- `type`, `flash_type`, `calendar_*` and `symbols_*` stay component configuration, matching the existing frontend dropdowns.

### Steps to reproduce

1. Build an agent whose Agent component has the Jin10 tool attached.
2. Run it — construction fails with `AttributeError: 'Jin10Param' object has no attribute 'get_meta'`.

```python
from agent.tools.jin10 import Jin10, Jin10Param

Jin10Param().get_meta()
# AttributeError: 'Jin10Param' object has no attribute 'get_meta'

tool = Jin10.__new__(Jin10)
tool._param = Jin10Param()
tool._invoke(query="gold")
```

### Additional information

`git log` shows `jin10.py` has used `ComponentBase`/`_run` since the agent module redesign, so the tool has been broken as an Agent tool the whole time.

A fix porting Jin10 to the `ToolBase`/`_invoke` interface (with `meta`, `set_output`, `_ERROR`) plus regression tests is ready; PR to follow.

immuhammadfurqan:
### Summary

Closes #18447.

The **Jin10** agent tool (`agent/tools/jin10.py`) was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it is unusable as an Agent tool in two independent ways.

**1. Adding it to an Agent raises `AttributeError`.** `Jin10Param` extended `ComponentParamBase` and defined no `meta`, and `get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` (`grep -n "def get_meta" agent/component/base.py` returns nothing). `agent/component/agent_with_tools.py:82` builds every tool's function descriptor via `cpn.get_meta()`, so constructing an Agent containing Jin10 raised `AttributeError: 'Jin10Param' object has no attribute 'get_meta'`.

**2. It could never run — `_run` is dead code.** `Jin10` only implemented the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more (`grep -rn "\._run(" agent/` returns no callers); all five execution paths go through `_invoke` — `agent/component/base.py:406,432,590` and `agent/tools/base.py:154,179`. With no `_invoke` defined, `Jin10._invoke` resolved to `ComponentBase._invoke`, which raises `NotImplementedError` (`agent/component/base.py:447`).

Note that `be_output` still exists (`agent/component/base.py:578`, returning `pd.DataFrame([{"content": v}])`) — it has not been removed — but nothing consumes its result for tools any more, which is why the legacy path is dead rather than merely deprecated.

### What changed

- Port `Jin10Param` to `ToolParamBase` with a `ToolMeta` defined **before** `super().__init__()` (matching `ArXivParam`/`AkShareParam`, since `ToolParamBase.__init__` reads `self.meta` immediately), exposing a required `query` parameter with default `{sys.query}`.
- `query` is used as the `contain` keyword filter. This is the one input the old code nominally accepted: `_run` computed `ans` from the upstream input, returned early when it was empty, and then **never referenced `ans` again** — every request was built purely from `self._param`, so the input gated execution without influencing it. Routing it to `contain` gives the parameter an actual effect for the `flash` and `news` types.
- `type`, `flash_type`, `calendar_type`, `calendar_datatype`, `symbols_type`, `symbols_datatype` and `filter` stay component configuration, matching the existing `Jin10TypeOptions` / `Jin10FlashTypeOptions` / `Jin10CalendarTypeOptions` / `Jin10SymbolsTypeOptions` dropdowns in `web/src/pages/agent/options.ts`.
- Rewrite the component with `_invoke`/`set_output("formalized_content", ...)`, surfacing failures via `_ERROR`, and add `get_input_form()` and `thoughts()` consistent with the other ported tools.
- Wrap `_invoke` in the standard `@timeout(int(os.environ.get("COMPONENT_EXEC_TIMEOUT", 12)))`, matching the other single-API-call tools.
- Output shape changes from a multi-row `pd.DataFrame` of `{"content": ...}` rows to those same items joined with `\n\n`, which is what `formalized_content` expects. The per-type request construction, field renaming and cancellation checks are otherwise unchanged.

### Testing

Adds `test/unit_test/agent/tools/test_jin10.py` (10 tests): param construction and `check()` validation, the tool descriptor exposing a required `query`, `flash` returning joined item content, `query` being passed through as the `contain` filter, `calendar` rendering a markdown table, an empty query short-circuiting without calling the API, and request failures surfacing via `_ERROR`.

Verified against `68f826082` — with `agent/tools/jin10.py` reverted to `main` and the new tests applied, 6 of the 10 fail with exactly the two defects above:

```
FAILED test_meta_exposes_query_parameter - AttributeError: 'Jin10Param' object has no attribute 'get_meta'
FAILED test_invoke_flash_returns_content_and_sets_formalized_content - NotImplementedError
FAILED test_invoke_passes_query_as_contain_filter - NotImplementedError
FAILED test_invoke_calendar_returns_markdown_table - NotImplementedError
FAILED test_invoke_empty_query_returns_empty - NotImplementedError
FAILED test_invoke_surfaces_error_on_request_failure - NotImplementedError
========================= 6 failed, 4 passed =========================
```

With the fix applied all 10 pass, and the wider `test/unit_test/agent/` suite reports 130 passed / 0 failed. No other Python file in the repo references `Jin10`.


wangq8 (commented):
Thanks — this is a correct port of Jin10 to `ToolBase`/`ToolParamBase`, consistent with the `AkShare`/`TuShare` pattern.

Verified against `agent/tools/base.py`:

- `ToolMeta` defined before `super().__init__()` (required).
- `_invoke` implemented; legacy `_run` removed.
- `query` is now routed to the `contain` filter, giving the input an actual effect (the old code gated execution on the input but never used it).
- Output shape changed from a multi-row `DataFrame` of `{"content": ...}` rows to those items joined with `\n\n` — correct for `formalized_content`.
- Failures surface via `_ERROR`.

Tests cover param construction, `check()` validation, meta exposure, `flash` content, `contain` passthrough, `calendar` markdown table, empty-query short-circuit, and request failure.

One non-blocking nit: `self.contain = ""` is still declared in `Jin10Param.__init__` but is no longer read by `_invoke` (it's superseded by the `query` tool input). Consider removing it — or confirming it's intentionally kept for backward compatibility — so it doesn't linger as a no-op config field. Otherwise LGTM.

immuhammadfurqan:
Thanks for the review!

**`self.contain`** — you're right, it's a no-op after the port. `_invoke` binds the tool's `query` input to the request's `contain` parameter locally, so `self._param.contain` is read nowhere; I grepped the repo (including `web/`) and it has no other consumer. Not intentional backward compatibility — just leftover. Removed in 69646ba.

I kept `filter`, `flash_type`, `calendar_*` and `symbols_*` as config, since those are still read by `_invoke` and aren't expressible through the tool input.

**On the failing check** — `ragflow_tests_infinity` failed on:

```
FAILED test/testcases/restful_api/test_message_routes_unit.py::test_get_messages_csv_and_missing_memory_ids
  - requests.exceptions.ChunkedEncodingError: ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')")
```

That's the CSV streaming route, unrelated to this PR, and the same commit passed `ragflow_tests_elasticsearch`. Looks like runner flake; the push above should give it a clean re-run.

**On the CodeRabbit timeout comment** — I looked into it and I don't think it belongs in this PR. The behaviour it describes is in the shared `common/connection_utils.timeout` decorator: the sync wrapper starts a daemon thread once, then loops `attempts` times over `result_queue.get(timeout=seconds)` without restarting the work, and blocks with no deadline at all when `ENABLE_TIMEOUT_ASSERTION` is unset. That's accurate, but it's the decorator's contract for every tool that uses it — 21 of the 26 files in `agent/tools/`, including the already-merged `akshare` and `deepl`. Changing it here would either make Jin10 inconsistent with the rest or turn a port into a cross-cutting change to shared execution machinery. Happy to open a separate issue for it if you'd like that looked at.

PR #18450, opened 2026-08-18
fix(agent/tools): port TuShare to ToolBase so it works as an Agent tool

### LINKED ISSUE #18448: [Bug]: TuShare agent tool is non-functional — no get_meta (crashes Agent) and only implements the dead _run path
### Self Checks

68f826082

### RAGFlow image version

v0.26.4

### Actual behavior

The **TuShare** agent tool (`agent/tools/tushare.py`) is non-functional. It was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it cannot be used as an Agent tool.

`TuShare` extends `ComponentBase` (not `ToolBase`) and `TuShareParam` extends `ComponentParamBase` with no `meta`. This breaks in two independent ways.

```python
    original_name = cpn.get_meta()["function"]["name"]   # TuShare has no get_meta()
```

`get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` — `grep -n "def get_meta" agent/component/base.py` returns nothing. So constructing an Agent that includes the TuShare tool raises:

```
AttributeError: 'TuShareParam' object has no attribute 'get_meta'
```

`TuShare` only implements the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more; every execution path goes through `_invoke`:

`grep -rn "\._run(" agent/` returns no callers. Since `TuShare` defines no `_invoke`, it falls through to `ComponentBase._invoke`, which raises `NotImplementedError` (`agent/component/base.py:447`).

The existing unit tests in `test/unit_test/agent/tools/test_tushare.py` all call `cpn._run([])` directly, which is why they pass today despite the tool being unusable through the Agent runtime.

The tool is user-reachable, not dead code: `agent/tools/__init__.py` auto-imports every module in the directory and registers classes by name, and the frontend ships the `TuShareSrcOptions` dropdown in `web/src/pages/agent/options.ts` plus locale strings in ~12 languages.

### Expected behavior

- The TuShare tool exposes a proper tool descriptor via `get_meta()` with a `query` parameter (the keyword to filter news by), so an Agent's LLM can call it.
- Invoking it fetches quick news from the configured source, filters by keyword, writes the result to `formalized_content`, and surfaces failures via `_ERROR` instead of the legacy `be_output` DataFrame.
- A non-zero `code` in the TuShare API response is surfaced as an error rather than returned as ordinary content.
- `token`, `src`, `start_date`, `end_date` and `keyword` stay component configuration, matching the existing frontend dropdown, and the existing keyword precedence (explicit `param.keyword` wins, literal non-regex match) is preserved.

### Steps to reproduce

1. Build an agent whose Agent component has the TuShare tool attached.
2. Run it — construction fails with `AttributeError: 'TuShareParam' object has no attribute 'get_meta'`.

```python
from agent.tools.tushare import TuShare, TuShareParam

TuShareParam().get_meta()
# AttributeError: 'TuShareParam' object has no attribute 'get_meta'

tool = TuShare.__new__(TuShare)
tool._param = TuShareParam()
tool._invoke(query="Apple")
```

### Additional information

`git log` shows `tushare.py` has used `ComponentBase`/`_run` since the agent module redesign, so the tool has been broken as an Agent tool the whole time.

A fix porting TuShare to the `ToolBase`/`_invoke` interface (with `meta`, `set_output`, `_ERROR`) plus regression tests is ready; PR to follow.

immuhammadfurqan:
### Summary

Closes #18448.

The **TuShare** agent tool (`agent/tools/tushare.py`) was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it is unusable as an Agent tool in two independent ways.

**1. Adding it to an Agent raises `AttributeError`.** `TuShareParam` extended `ComponentParamBase` and defined no `meta`, and `get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` (`grep -n "def get_meta" agent/component/base.py` returns nothing). `agent/component/agent_with_tools.py:82` builds every tool's function descriptor via `cpn.get_meta()`, so constructing an Agent containing TuShare raised `AttributeError: 'TuShareParam' object has no attribute 'get_meta'`.

**2. It could never run — `_run` is dead code.** `TuShare` only implemented the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more (`grep -rn "\._run(" agent/` returns no callers); all five execution paths go through `_invoke` — `agent/component/base.py:406,432,590` and `agent/tools/base.py:154,179`. With no `_invoke` defined, `TuShare._invoke` resolved to `ComponentBase._invoke`, which raises `NotImplementedError` (`agent/component/base.py:447`).

The existing tests in `test/unit_test/agent/tools/test_tushare.py` all called `cpn._run([])` directly, which is why they passed despite the tool being unusable through the Agent runtime.

Note that `be_output` still exists (`agent/component/base.py:578`, returning `pd.DataFrame([{"content": v}])`) — it has not been removed — but nothing consumes its result for tools any more, which is why the legacy path is dead rather than merely deprecated.

### What changed

- Port `TuShareParam` to `ToolParamBase` with a `ToolMeta` defined **before** `super().__init__()` (matching `ArXivParam`/`AkShareParam`, since `ToolParamBase.__init__` reads `self.meta` immediately), exposing a required `query` parameter — the keyword to filter news by, default `{sys.query}`.
- `token`, `src`, `start_date`, `end_date` and `keyword` stay component configuration, matching the existing `TuShareSrcOptions` dropdown in `web/src/pages/agent/options.ts`.
- Rewrite the component with `_invoke`/`set_output("formalized_content", ...)`, and add `get_input_form()` and `thoughts()` consistent with the other ported tools.
- A non-zero `code` in the TuShare API response now goes to `_ERROR` instead of being returned as ordinary content, so an auth/quota failure is no longer indistinguishable from a successful empty result.
- Wrap `_invoke` in the standard `@timeout(int(os.environ.get("COMPONENT_EXEC_TIMEOUT", 12)))`, matching the other single-API-call tools.

**Keyword behaviour is deliberately preserved.** An explicit `param.keyword` still takes precedence over the incoming query (`keyword = self._param.keyword or query`), and the filter remains a literal `regex=False` match, so the `C++`-style inputs guarded by the existing tests keep working. Only the source of the fallback changes: previously the joined upstream input, now the tool's `query` argument. The existing `logging.info` recording which source was used is kept, with the label updated from `upstream_input` to `tool_query`.

### Testing

Rewrites `test/unit_test/agent/tools/test_tushare.py` from 3 tests to 10. The three existing keyword-semantics tests are preserved in intent and moved from `_run` to `_invoke`, since `_run` no longer exists: fallback to the query when `param.keyword` is empty, explicit `param.keyword` winning over the query, and the keyword being treated as literal text.

New coverage: param construction, `check()` rejecting an invalid `src`, the tool descriptor exposing a required `query`, an empty query short-circuiting without calling the API, a non-zero API `code` surfacing via `_ERROR`, and request failures surfacing via `_ERROR`.

Verified against `68f826082` — with `agent/tools/tushare.py` reverted to `main` and the new tests applied, 7 of the 10 fail with exactly the two defects above:

```
FAILED test_meta_exposes_query_parameter - AttributeError: 'TuShareParam' object has no attribute 'get_meta'
FAILED test_tushare_filters_with_tool_query_when_param_empty - NotImplementedError
FAILED test_tushare_prefers_explicit_param_keyword_over_tool_query - NotImplementedError
FAILED test_tushare_treats_keyword_as_literal_text - NotImplementedError
FAILED test_invoke_empty_query_returns_empty - NotImplementedError
FAILED test_invoke_surfaces_api_error_message - NotImplementedError
FAILED test_invoke_surfaces_error_on_request_failure - NotImplementedError
========================= 7 failed, 3 passed =========================
```

With the fix applied all 10 pass, and the wider `test/unit_test/agent/` suite reports 130 passed / 0 failed. No other Python file in the repo references `TuShare`.


wangq8 (commented):
Thanks for this fix — it correctly ports TuShare to the `ToolBase`/`ToolParamBase` interface, matching the established `AkShare` pattern.

Verified against `agent/tools/base.py` and the other ported tools:

- `ToolMeta` is defined **before** `super().__init__()`, which is required since `ToolParamBase.__init__` reads `self.meta` immediately (via `_init_inputs` / `_init_attr_by_meta`).
- `get_meta()` now resolves through `ToolParamBase`, so `agent_with_tools.py` no longer raises `AttributeError` when the tool is added to an Agent.
- `_invoke` is now the real execution path; the legacy `_run` dead code is removed.
- Non-zero API `code` and request exceptions now surface via `_ERROR` rather than being silently returned as ordinary content — good.
- `@timeout(COMPONENT_EXEC_TIMEOUT)` is consistent with the other single-API-call tools.
- Keyword semantics are preserved (explicit `param.keyword` precedence, literal `regex=False` match, fallback to the tool `query`).

Tests look solid: the three keyword-semantics cases are preserved in intent and moved from `_run` to `_invoke`, plus new coverage for meta exposure, `check()` on invalid `src`, empty-query short-circuit, API error, and request failure.

No regressions or security concerns found. LGTM.

immuhammadfurqan:
Thanks for the review! I've pushed c11090ed addressing the two points CodeRabbit raised, since both turned out to be real.

**1. Empty result set — confirmed, fixed.**

`df.columns = fields` was assigned after constructing the frame, so a response with `items: []` raised:

```
ValueError: Length mismatch: Expected axis has 0 elements, new values have 2 elements
```

TuShare returns that whenever the configured `src`/date window has no news, so an ordinary empty result was surfacing as a tool error. Passing `columns=` to the constructor keeps the headers in both cases. This one predates the port — the old `_run` had the same two lines — but since the port is what makes this path reachable from an Agent, it seemed worth fixing here rather than leaving behind.

**2. Cleartext token — confirmed, fixed.**

`token` is a credential and was going over plain HTTP. I checked the live endpoint before changing it, because CodeRabbit's own research noted TuShare documents only `http://` and questioned whether HTTPS exists:

```
$ curl -s -w '[status=%{http_code} ssl_verify=%{ssl_verify_result}]' \
    -X POST https://api.tushare.pro -d '{"api_name":"news","token":"invalid","params":{}}'
{"request_id":"","code":40101,"data":null,"msg":"您的token不对，请确认。","detail":""}
[status=200 ssl_verify=0]
```

`https://api.tushare.pro` serves the identical JSON API and the certificate verifies (`ssl_verify=0`), so the switch is safe. Repeated the check across several runs to rule out a fluke.

Both are covered by regression tests that fail on the previous code:

```
FAILED test_invoke_handles_empty_result_set  - ValueError: Length mismatch: Expected axis has 0 elements...
FAILED test_invoke_sends_token_over_https    - 'http://api.tushare.pro'.startswith('https://')
======================== 2 failed, 10 passed ========================
```

12/12 pass with the fix; ruff clean.

PR #18453, opened 2026-08-18
Fix task time is incorrect

wangq8:


wangq8 (commented):
Thanks for the fix — the approach is clean and correctly applied across both the legacy (`task_executor.py`) and refactored (`task_executor_refactor/`) paths.

**What looks good**

- Measuring the limiter-acquisition wait with `timer()` (`timeit.default_timer` → `perf_counter`) is monotonic, so `wait_time` is always a valid duration.
- Passing a callback with a `None` default is backward-compatible and doesn't force any other callers to change.
- `nonlocal task_start_ts` is correct — the callback is invoked synchronously right after `async with` acquires the semaphore, with no intervening `await`, so the mutation is safe (no races).

**Points to consider**

1. **Inconsistent scope of "task time".** Only the `chunk_limiter` wait is excluded from `task_time_cost`, but the task is also serialized on `embed_limiter` (same default of `MAX_CONCURRENT_CHUNK_BUILDERS=1`, at `task_executor.py:738`) and `minio_limiter`. If the intent is "report actual processing time excluding concurrency-queue waits", embedding (and to a lesser extent minio) waits are still counted, so the reported time remains inflated whenever embedding is the bottleneck. Could you clarify the intended scope? If the goal is only to fix the chunking wait, a short comment noting that other limiters are intentionally left in would help future readers.

2. **Type hints / docstrings.** The new `on_chunking_start` parameter has no type annotation in `run_chunking`, `ChunkService.build_chunks`, or legacy `build_chunks`, while the surrounding code is heavily typed (e.g. `ctx: TaskContext`). Suggest `on_chunking_start: Optional[Callable[[float], None]] = None` for consistency.

3. **Test coverage.** No tests were added. Timing is hard to assert deterministically, but a focused test that the callback is invoked (and receives the wait duration) — e.g. by injecting a limiter or mocking the semaphore — would lock in the contract and prevent regressions when the refactored executor diverges.

Minor: `timer() - chunking_wait_started_at` is always `>= 0`, but under no contention it's ~0 and `task_start_ts += 0` is a harmless no-op; nothing to change, just noting it's fine.

PR #18457, opened 2026-08-18
Fix reasoning not working in multiple model chat

wangq8:


wangq8 (commented):
Thanks for the fix. The core change is correct: `reasoning` arriving as the string `"0"` is truthy in Python, so the old `not kwargs.get("reasoning")` check failed to short-circuit and the request wrongly fell through to the agent path (with `thinking_mode` defaulting to `"medium"`). Resolving a single value with kwargs taking precedence, plus the explicit `"0"` check, fixes that. A few suggestions:

1. **Reuse the same "off" detection as the downstream parse.** `str(reasoning).strip() == "0"` only matches the literal `"0"`; values like `"00"` or `" 0 "` (the latter is covered by `.strip()`, the former not) would still be treated as enabled. The level resolution a few lines below already does `int(str(...).strip())`. Consider deriving both from one parse so the gate and the level can't disagree, e.g.:
   ```python
   try:
       reasoning_off = int(str(reasoning).strip()) == 0
   except (TypeError, ValueError):
       reasoning_off = not reasoning
   if reasoning_off:
       ...
   ```
   Minor today since the frontend only sends 0–4, but it keeps the two checks consistent.

2. **Downstream `thinking_mode` still reads `kwargs.get("reasoning")` directly** (line ~1956), not the resolved `reasoning`. When reasoning is enabled via `prompt_config` (key absent from kwargs), the agent path is taken but the level silently falls back to `"medium"`. Consider using the resolved `reasoning` there too, so the gate and the level come from the same source.

3. **`None` edge case / behavior change worth confirming.** Because the new gate resolves `kwargs["reasoning"]` whenever the key is present, a client sending `reasoning: null` now resolves to `None` and disables reasoning even if `prompt_config.reasoning` is truthy. If `null` should mean "not provided", prefer:
   ```python
   reasoning = kwargs.get("reasoning")
   if reasoning is None:
       reasoning = prompt_config.get("reasoning", 0)
   ```
   Also note this is an intentional precedence change: an explicit `reasoning: 0`/`"0"` in the request now fully overrides a truthy `prompt_config.reasoning` (previously both were effectively OR'd). That looks correct for per-message override, just flagging it so the behavior change is deliberate.

4. **No tests added.** A small unit test covering `"0"`, `0`, `"2"`, `2`, and the `prompt_config` fallback would lock in the intended precedence and the string-`"0"` case.

Nothing blocking otherwise — happy to see this land once the above are addressed or acknowledged.

Issue #18415, opened 2026-08-18
[Bug]: Memory extraction stores an empty valid_at for semantic items, and an unparseable timestamp is written through unchanged


danielgines:
## Summary

Two small defects combine so that semantic memory items are stored with no validity timestamp, silently.

1. The generated extraction prompt tells the model that `valid_at` may be empty — but only in the `semantic` block, not the `episodic` one.
2. When the returned value does not parse as ISO 8601, `format_iso_8601_to_ymd_hms` logs the exception message and **returns the original string**, which is then stored as-is.

Observed on **v0.26.4**.

## 1. The asymmetric output format

`PromptAssembler.assemble_system_prompt({"memory_type": ["semantic", "episodic"]})` produces:

```json
{
    "semantic": [
        {
            "content": "Clear factual statement",
            "valid_at": "timestamp or empty",
            "invalid_at": "timestamp or empty"
        }
    ],
    "episodic": [
        {
            "content": "Narrative event description",
            "valid_at": "event start timestamp",
            "invalid_at": "event end timestamp or empty"
        }
    ]
}
```

`valid_at` is declared optional for semantic and mandatory for episodic, even though the instruction block above it says `Default: valid_at = conversation time`, and the output requirements say `Each extracted item MUST have: content, valid_at, invalid_at`. The format block is the part the model follows.

In a memory holding 4,132 records, **144 items had an empty `valid_at` — 6.5% of the semantic items and 0% of the episodic and raw ones**. The distribution matches the asymmetry exactly.

## 2. The silent write-through

`api/db/joint_services/memory_message_service.py:191`

```python
"valid_at": format_iso_8601_to_ymd_hms(extracted_content["valid_at"]),
"invalid_at": format_iso_8601_to_ymd_hms(extracted_content["invalid_at"]) if extracted_content.get("invalid_at") else "",
```

`common/time_utils.py:132`

```python
def format_iso_8601_to_ymd_hms(time_str: str) -> str:
    from dateutil import parser
    try:
        dt = parser.isoparse(time_str)
        return dt.strftime("%Y-%m-%d %H:%M:%S")
    except Exception as e:
        logging.error(str(e))
        return time_str          # <-- stored unchanged
```

Whatever the model produced is persisted. The log line carries only the exception message, e.g.

```
ERROR ISO string too short
```

which does not say which field failed, which value was rejected, or which memory and message it belonged to — so the 20 occurrences we see per day are unactionable as logged.

## Why it matters

[...]

Issue #18413, opened 2026-08-18
[Bug]: Memory system_prompt and user_prompt are stored and returned by the API but never reach the extraction LLM


danielgines:
## Summary

A memory object accepts `system_prompt` and `user_prompt` via `PUT /api/v1/memories/<id>`, persists them, and returns them from `GET /api/v1/memories/<id>/config`. Neither value is ever passed to the model that performs the extraction: `extract_by_llm` receives them as optional parameters, and **both call sites omit them**, so the function always falls back to the freshly assembled default prompt.

The effect is a configuration surface that can be written, read back and displayed, while having no influence on behaviour.

Observed on **v0.26.4**.

## Code

`api/db/joint_services/memory_message_service.py`

Signature and fallbacks (line 158):

```python
async def extract_by_llm(
    tenant_id: str,
    tenant_llm_id: int,
    extract_conf: dict,
    memory_type: List[str],
    user_input: str,
    agent_response: str,
    system_prompt: str = "",     # <-- never supplied by any caller
    user_prompt: str = "",       # <-- never supplied by any caller
    task_id: str = None,
    llm_id: str = "",
) -> List[dict]:
    if not system_prompt:
        system_prompt = PromptAssembler.assemble_system_prompt({"memory_type": memory_type})
    ...
    if user_prompt:
        user_prompts.append({"role": "user", "content": user_prompt})
        ...
    else:
        user_prompts.append({"role": "user", "content": PromptAssembler.assemble_user_prompt(...)})
```

Call site 1 (line 56):

```python
extracted_content = (
    await extract_by_llm(
        tenant_id,
        memory.tenant_llm_id,
        {"temperature": memory.temperature},
        get_memory_type_human(memory.memory_type),
        message_dict.get("user_input", ""),
        message_dict.get("agent_response", ""),
        llm_id=memory.llm_id,
    )
    ...
)
```

Call site 2 (line 120):

```python
extracted_content = await extract_by_llm(
    tenant_id,
    memory.tenant_llm_id,
    {"temperature": memory.temperature},
    get_memory_type_human(memory.memory_type),
    message_dict.get("user_input", ""),
    message_dict.get("agent_response", ""),
    task_id=task_id,
    llm_id=memory.llm_id,
)
```

`memory.system_prompt` / `memory.user_prompt` are read in only three places, none of which is the extraction call:

- `api/apps/services/memory_api_service.py:223` — to decide whether to regenerate the default when `memory_type` changes
[...]

PR #18458, opened 2026-08-18
fix(agent): apply cross_languages translation in Retrieval memory mode

### LINKED ISSUE #18417: [Bug]: Retrieval tool accepts cross_languages in memory mode and silently ignores it
## Summary

The agent's Retrieval tool has two execution paths, selected by `retrieval_from`. `cross_languages` is honoured in the dataset path and silently ignored in the memory path, while the parameter is accepted and persisted for both.

A tool configured with `retrieval_from: "memory"` and `cross_languages: ["English"]` reads back exactly as configured and behaves as if the setting were absent.

Observed on **v0.26.4**.

## Code

`agent/tools/retrieval.py`

```python
async def _invoke_async(self, **kwargs):
    ...
    if hasattr(self._param, "retrieval_from") and self._param.retrieval_from == "dataset":
        return await self._retrieve_kb(kwargs["query"])
    elif hasattr(self._param, "retrieval_from") and self._param.retrieval_from == "memory":
        return await self._retrieve_memory(kwargs["query"])
    ...
```

`_retrieve_kb` (from line 90) translates the query:

```python
if self._param.cross_languages:
    query = await cross_languages(kbs[0].tenant_id, None, query, self._param.cross_languages)
```

`_retrieve_memory` (lines 267-300) goes straight from the query to the search, with no equivalent:

```python
vars = self.get_input_elements_from_text(query_text)
query = self.string_format(query_text, vars)
...
message_list = memory_message_service.query_message(
    filter_dict,
    {"query": query, "similarity_threshold": ..., "keywords_similarity_weight": ..., "top_n": ...},
)
```

## Steps to reproduce

1. Add a Retrieval tool to an agent with `retrieval_from: "memory"` and a memory selected.
2. Set `cross_languages: ["English"]` on it and save the agent.
3. `GET /api/v1/agents/<id>` — the parameter is present in the DSL, so it was accepted and stored.
4. Ask a question in a language other than English and inspect the agent log around the tool call.

**Expected:** either the query is translated before the memory search, or the parameter is rejected or visibly disabled for this mode.
**Actual:** no translation happens and there is no indication that the setting is inert.

For contrast, the same parameter on a `retrieval_from: "dataset"` tool produces an extra LLM call inside the tool execution, visible in the log between `[ToolCall] invoke` and `[ToolCall] done` — in our measurement 245 tokens and about 1.9 s of a 2.17 s tool call.

## Suggested fix

Either apply `cross_languages` in `_retrieve_memory` as well, or make the parameter unavailable when `retrieval_from` is `memory` so the stored configuration matches the behaviour.

## Note

This is the same class of problem as memory `system_prompt` / `user_prompt` being stored and never passed to the extraction LLM (filed separately). A configuration value that is accepted, persisted and returned by the API, but not applied, is indistinguishable from a working one until someone reads the source.

Harsh23Kashyap:
Fixes #18417.

## Summary

The agent Retrieval tool has two execution paths selected by `retrieval_from`: `dataset` and `memory`. The `cross_languages` setting is applied to the query before search in `_retrieve_kb` but was silently ignored in `_retrieve_memory`. The setting was accepted on the tool's parameter, persisted, returned unchanged by the API, and produced no translation at invocation time. Indistinguishable from a working setting until someone reads the source.

## Fix

Mirror the dataset-mode translation step in the memory path, so `cross_languages` is honored consistently across both paths:

```python
if self._param.cross_languages:
    query = await cross_languages(memory_list[0].tenant_id, None, query, self._param.cross_languages)
```

`memory_list[0].tenant_id` is the parallel of `kbs[0].tenant_id` already used in `_retrieve_kb`. The same `rag.prompts.generator.cross_languages` function is reused; no new helper, no new import.

## Test

Nine regression tests in `test/unit_test/agent/tools/test_retrieval_cross_languages.py`:

- the main test pins the translator call (with the first memory's tenant_id and the original query) and asserts the translated query reaches `memory_message_service.query_message`
- the empty-list case asserts no translator call and no query modification
- the translator-error case asserts the existing fallback (return the original query) flows through
- the empty-memory case asserts the error path runs before the translator
- a parametrized test covers `[]`, `["English"]`, and multi-language settings
- a test pins the first-memory tenant_id contract
- a dynamic-discovery test pins the public export

Verified pre-fix behavior: the main test fails on the unpatched function (`assert 0 == 1: cross_languages was never called`). With the fix applied, all 9 cases pass.

## Risks

- `cross_languages` is an LLM call; honoring it in memory mode adds the same latency cost the dataset path already pays. The user opted into the setting; honoring their choice is the fix.
- If the first memory's tenant differs from subsequent memories' tenants, the translator uses the first one's. This matches the existing dataset-mode contract (uses `kbs[0].tenant_id`) and the assertion `len(embd_names) == 1` already restricts memories to one embedding model — multi-tenant memory is not a supported configuration in this path.

## Related

Same bug class (parameter accepted, persisted, returned, not applied) as #18413 (memory `system_prompt` / `user_prompt`) and #18415 (memory `valid_at`). The cross_languages family is now fixed; the other two still apply.

### DIFF
--- a/agent/tools/retrieval.py
+++ b/agent/tools/retrieval.py
@@ -277,6 +277,13 @@ class Retrieval(ToolBase, ABC):
         vars = self.get_input_elements_from_text(query_text)
         vars = {k: o["value"] for k, o in vars.items()}
         query = self.string_format(query_text, vars)
+        # Apply the same cross_languages translation that ``_retrieve_kb`` does
+        # before the search. Without this, the cross_languages setting is
+        # accepted and persisted on a memory-mode tool, then silently ignored
+        # at invocation time — indistinguishable from a working setting until
+        # someone reads the source (issue #18417).
+        if self._param.cross_languages:
+            query = await cross_languages(memory_list[0].tenant_id, None, query, self._param.cross_languages)
         # query message
         filter_dict: dict = {"memory_id": memory_ids}
         if user_id:
--- a/test/unit_test/agent/tools/test_retrieval_cross_languages.py
+++ b/test/unit_test/agent/tools/test_retrieval_cross_languages.py
@@ -1,0 +1,279 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for issue #18417.
+
+The agent Retrieval tool has two execution paths selected by
+``retrieval_from``: ``dataset`` and ``memory``. The ``cross_languages``
+setting is applied to the query before search in the dataset path
+(``_retrieve_kb``) but was silently ignored in the memory path
+(``_retrieve_memory``) -- the setting was accepted and persisted on the
+tool's parameter, read back unchanged by the API, and produced no
+translation at invocation time. Same class of bug as #18413 (memory
+system_prompt) and #18415 (memory valid_at): a parameter that is
+accepted, persisted, returned by the API, but not applied.
+
+These tests pin the memory-mode ``cross_languages`` translation on the
+same conditions the dataset-mode path already honours, so the two paths
+stay in lockstep.
+"""
+
+import asyncio
+from types import SimpleNamespace
+
+import pytest
+
+import agent.tools.retrieval as retrieval_module
+from agent.tools.retrieval import Retrieval, RetrievalParam
+
+
+class _StubMessage:
+    """Dict-shaped stand-in for a memory search result row.
+
+    ``memory_prompt`` reads the message via ``message["content"]``, so the
+    stub must support subscript access rather than only attribute access.
+    """
+
+    def __init__(self, content):
+        self._content = content
+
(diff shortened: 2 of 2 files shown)


Lynn-Inf (approved):
**LGTM** — the parity fix is right. `_retrieve_kb` already applies the setting at `retrieval.py:178-179`, and memory mode accepted, persisted and returned `cross_languages` without ever translating, which is indistinguishable from a working setting until you read the source. Mirroring it with `memory_list[0].tenant_id` matches the existing `kbs[0].tenant_id` contract, reuses the same `cross_languages` helper, and the tests pin the call, the empty-list and translator-error fallbacks.

**One caveat worth recording:** translation is not free of distortion. `cross_languages` calls a chat LLM and *replaces* the query with joined translations (`rag/prompts/generator.py:291-316`), so the vector leg embeds a multi-language blob rather than the original query — and LLM translation is lossy for entities, numbers, code identifiers and proper nouns. It clearly pays off for the keyword/term leg (and for monolingual embedding models), but for a pure-vector setup (keyword weight 0) with a multilingual embedding model it is pure cost with no benefit. The dataset path has always been ungated here, so this is a pre-existing, wider design question rather than something this PR introduces — I'd suggest a follow-up to either gate on the keyword weight or document the trade-off, but it shouldn't block this one.

**On Go:** there's nothing to mirror yet. `internal/agent/tool/retrieval.go` already plumbs `CrossLanguages` into its args (line 70, with the defaults fallback at 297-298), but the retrieval service itself isn't implemented — `InvokableRun` returns `ErrRetrievalServiceMissing` ("Retrieval service not yet implemented (service not registered)") — so neither path translates today. Worth picking up together with the gating question above when that service lands.

Lynn-Inf:
@Harsh23Kashyap The CI checks have failed. Could you please take a look and fix the issues? Thanks!

JinHai-CN:
https://github.com/infiniflow/ragflow/pull/19669 fix the issue

PR #18460, opened 2026-08-18
fix(rag/app/table): keep table cells that column type conversion cannot represent

### LINKED ISSUE #18459: [Bug]: table parser silently drops cells that column-level type conversion cannot represent
## Summary

`rag/app/table.py::column_data_type()` infers one type per column and then converts every cell to it. Any cell that does not fit is replaced with `None` — and `chunk()` skips `None` cells, so the value is absent from both the chunk body and the stored field. A spreadsheet column that is mostly numeric but carries `N/A`, `TBD`, `-`, `pending`, or a footnote loses exactly those cells, with no user-visible error.

This contradicts the code as written. PR #17946 added a comment stating the opposite behaviour, and added a consumer branch in `chunk()` that only makes sense if the original value survives — that branch is currently unreachable.

Observed on **v0.26.4** (commit `2aaa6baf`).

## Code

Producer — `rag/app/table.py:420-427`:

```python
        try:
            arr[i] = trans[ty](str(arr[i]))
        except Exception as e:
            arr[i] = None
            logging.warning(f"Column {i}: {e}")
            # Keep original value from openpyxl/pandas instead of dropping to None.
            # This preserves cells (e.g. text in numeric columns) that would
            # otherwise be silently discarded by forced column-level conversion.
```

The comment sits directly below the `arr[i] = None` it says does not happen. The `bool` and `datetime` converters do not even raise — `trans_bool("unknown")` and `trans_datatime("ongoing")` return `None` — so those cells are nulled through the success path, without the warning.

Consumer — `rag/app/table.py:617-627`:

```python
                        if clmn_tys[j] != "text":
                            val = row[col_name]
                            # If a string value ended up in a non-text column,
                            # it was preserved from a failed conversion in
                            # column_data_type. Skip storing in the typed ES
                            # field to avoid type mapping errors; the value is
                            # already in text_fields for chunk content.
                            if isinstance(val, str):
                                pass
                            else:
                                stored[fld] = val
```

`column_data_type` never leaves a string in a non-text column, so `isinstance(val, str)` is dead code. And `chunk()` at line 600:

```python
                if row[clmns[j]] is None:
                    continue
```

drops the nulled cell before it reaches `text_fields`, so it is missing from the chunk content too, not only from the typed metadata field.

## Reproduction

Runs in a couple of seconds. No database, Redis, Elasticsearch, model or Docker.

```python
import sys
from unittest.mock import MagicMock

# Keep the optional storage/parser backends out of the import path; column_data_type
# touches none of them. Same shape as test/unit_test/rag/app/test_table_chunk_column_roles.py
for name in ("api.db.services.knowledgebase_service", "deepdoc.parser",
             "deepdoc.parser.figure_parser", "deepdoc.parser.utils",
             "deepdoc.vision.ocr", "rag.app.picture", "common.settings"):
    m = MagicMock(); m.__path__ = []
    sys.modules[name] = m
sys.modules["common.settings"].DOC_ENGINE_INFINITY = False

from rag.app.table import column_data_type

for column in (["1", "2", "3", "N/A"],
               ["1.5", "2.5", "3.5", "see note"],
               ["yes", "no", "yes", "unknown"],
               ["2025-01-01", "2025-02-01", "ongoing"]):
    out, ty = column_data_type(list(column))
    print(f"{str(column):<44} -> {ty:<9} {out}")
```

Run with `PYTHONPATH=. python3 repro.py`.

## Observed

```
['1', '2', '3', 'N/A']                       -> int       [1, 2, 3, None]
['1.5', '2.5', '3.5', 'see note']            -> float     [1.5, 2.5, 3.5, None]
['yes', 'no', 'yes', 'unknown']              -> bool      ['yes', 'no', 'yes', None]
['2025-01-01', '2025-02-01', 'ongoing']      -> datetime  ['2025-01-01 00:00:00', '2025-02-01 00:00:00', None]
```

## Expected

Per the comments already in the file, the cell keeps its original value:

```
['1', '2', '3', 'N/A']                       -> int       [1, 2, 3, 'N/A']
['1.5', '2.5', '3.5', 'see note']            -> float     [1.5, 2.5, 3.5, 'see note']
['yes', 'no', 'yes', 'unknown']              -> bool      ['yes', 'no', 'yes', 'unknown']
['2025-01-01', '2025-02-01', 'ongoing']      -> datetime  ['2025-01-01 00:00:00', '2025-02-01 00:00:00', 'ongoing']
```

End to end, for a CSV `row_id,amount,note` with the row `2,N/A,second`, the chunk for row 2 is currently:

```
- row_id: 2
- note: second
```

The `amount` column is missing entirely. It should read:

```
- row_id: 2
- amount: N/A
- note: second
```

## Impact

Applies to every CSV/XLSX ingested with the `table` parsing method. The lost cells are the rows that record an exception — missing data, a pending status, a footnote — which are often exactly what a user asks the knowledge base about. Nothing in the UI indicates the value was discarded; the only trace is a `Column N: ...` warning in the backend log, and the `bool`/`datetime` cases do not even produce that.

## Suggested fix

Keep the original value when the converter raises or returns `None`, which is what both comments and the unreachable `isinstance(val, str)` branch already assume. `chunk_data` is a JSON/JSONB column in Infinity, OceanBase, GaussDB and SereneDB, and the ES path is already guarded by that branch, so no doc engine receives a wrongly typed value.

OmarFaig:
### Summary

Fixes #18459.

`column_data_type()` in `rag/app/table.py` infers one type per column and replaces every cell it cannot convert with `None`. `chunk()` then skips `None` cells at line 600, so the value disappears from the chunk body as well as from the stored field. A CSV/XLSX column that is mostly numeric but carries `N/A`, `TBD`, `-` or a footnote loses exactly those cells, with no user-visible error.

The code already documents the opposite behaviour. PR #17946 added this comment directly below the `arr[i] = None` it describes:

```python
            # Keep original value from openpyxl/pandas instead of dropping to None.
            # This preserves cells (e.g. text in numeric columns) that would
            # otherwise be silently discarded by forced column-level conversion.
```

and added a branch in `chunk()` that only makes sense if the original value survives:

```python
                        if clmn_tys[j] != "text":
                            val = row[col_name]
                            # If a string value ended up in a non-text column,
                            # it was preserved from a failed conversion in
                            # column_data_type. ...
                            if isinstance(val, str):
                                pass
```

Since the producer never leaves a string in a non-text column, that branch is unreachable.

This change keeps the original value when the converter raises **or** returns `None`. The second case matters: `trans_bool("unknown")` and `trans_datatime("ongoing")` return `None` rather than raising, so `bool` and `datetime` columns were being nulled through the success path without even logging the `Column N: ...` warning.

No doc engine receives a wrongly typed value. `chunk_data` is a JSON/JSONB column in Infinity, OceanBase, GaussDB and SereneDB, and the Elasticsearch path is guarded by the `isinstance(val, str)` branch above, which this change makes reachable as intended.

### Behaviour

Before:

```
['1', '2', '3', 'N/A']                       -> int       [1, 2, 3, None]
['1.5', '2.5', '3.5', 'see note']            -> float     [1.5, 2.5, 3.5, None]
['yes', 'no', 'yes', 'unknown']              -> bool      ['yes', 'no', 'yes', None]
['2025-01-01', '2025-02-01', 'ongoing']      -> datetime  ['2025-01-01 00:00:00', '2025-02-01 00:00:00', None]
```

After:

```
['1', '2', '3', 'N/A']                       -> int       [1, 2, 3, 'N/A']
['1.5', '2.5', '3.5', 'see note']            -> float     [1.5, 2.5, 3.5, 'see note']
['yes', 'no', 'yes', 'unknown']              -> bool      ['yes', 'no', 'yes', 'unknown']
['2025-01-01', '2025-02-01', 'ongoing']      -> datetime  ['2025-01-01 00:00:00', '2025-02-01 00:00:00', 'ongoing']
```

End to end, for a CSV `row_id,amount,note` with the row `2,N/A,second`, the chunk for that row was `- row_id: 2\n- note: second` and is now `- row_id: 2\n- amount: N/A\n- note: second`.

### Tests

Added to `test/unit_test/rag/app/test_table_chunk_column_roles.py`, using the existing `table_module` fixture:

- `test_column_data_type_keeps_cells_it_cannot_convert` — four parametrized cases covering `int`, `float`, `bool` and `datetime` columns.
- `test_chunk_keeps_unconvertible_cell_in_content` — asserts the cell reaches `content_with_weight`.

All five fail on `main` and pass with this change. The 11 existing tests in that file are unaffected, and `test/unit_test/rag/app/` plus the table metadata/column-role helper tests and the chunker tests (`test_naive_merge`, `test_merge_paragraphs`, `test_delim`, `test_is_english`) pass — 212 tests total.

`ruff check` reports the same findings on `rag/app/table.py` before and after this change (18, all pre-existing), and `ruff format --check` reports both files already formatted.

### DIFF
--- a/rag/app/table.py
+++ b/rag/app/table.py
@@ -418,13 +418,16 @@ def column_data_type(arr):
             arr[i] = None
             continue
         try:
-            arr[i] = trans[ty](str(arr[i]))
+            converted = trans[ty](str(arr[i]))
         except Exception as e:
-            arr[i] = None
             logging.warning(f"Column {i}: {e}")
             # Keep original value from openpyxl/pandas instead of dropping to None.
             # This preserves cells (e.g. text in numeric columns) that would
             # otherwise be silently discarded by forced column-level conversion.
+            continue
+        if converted is None:
+            continue
+        arr[i] = converted
     # if ty == "text":
     #    if len(arr) > 128 and uni / len(arr) < 0.1:
     #        ty = "keyword"
--- a/rag/app/table.py
+++ b/rag/app/table.py
@@ -408,6 +408,7 @@ def column_data_type(arr):
         type_priority = {"text": 0, "datetime": 1, "float": 2, "int": 3, "bool": 4}
         counts = sorted(counts.items(), key=lambda x: (x[1] * -1, type_priority[x[0]]))
         ty = counts[0][0]
+    conversion_failures = 0
     for i in range(len(arr)):
         if arr[i] is None:
             continue
@@ -419,8 +420,8 @@ def column_data_type(arr):
             continue
         try:
             converted = trans[ty](str(arr[i]))
-        except Exception as e:
-            logging.warning(f"Column {i}: {e}")
+        except ValueError:
+            conversion_failures += 1
             # Keep original value from openpyxl/pandas instead of dropping to None.
             # This preserves cells (e.g. text in numeric columns) that would
             # otherwise be silently discarded by forced column-level conversion.
@@ -428,6 +429,11 @@ def column_data_type(arr):
         if converted is None:
             continue
         arr[i] = converted
+    if conversion_failures:
+        # Aggregate rather than log per-cell: individual cell values must not
+        # be written to application logs, and per-cell warnings would flood
+        # logs on large uploads.
+        logging.warning(f"column_data_type: kept {conversion_failures} cell(s) that could not convert to {ty}")
     # if ty == "text":
     #    if len(arr) > 128 and uni / len(arr) < 0.1:
     #        ty = "keyword"
--- a/test/unit_test/rag/app/test_table_chunk_column_roles.py
+++ b/test/unit_test/rag/app/test_table_chunk_column_roles.py
@@ -44,6 +44,11 @@ TEST_CSV = b"""row_id,title,content,country,category
 TEST_DUPLICATE_COLUMNS_CSV = b"""name,name,name_2
 Alice,Engineer,Team A
 """
+TEST_PARTIAL_NUMERIC_CSV = b"""row_id,amount,note
+1,100,first
+2,N/A,second
+3,300,third
+"""
 
 FILENAME = "test.csv"
 KB_ID = "test_kb_id"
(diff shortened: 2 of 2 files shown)


xugangqiang:
pls resolve conflicts

OmarFaig:
Rebased onto main and resolved the conflicts in `test/unit_test/rag/app/test_table_chunk_column_roles.py` (kept both the non-ASCII field-map tests and this PR's type-conversion tests).

Also addressed both CodeRabbit findings on `rag/app/table.py`:
- Catch only `ValueError` in the conversion loop instead of a blind `Exception`, so unexpected failures propagate instead of being silently swallowed.
- Stopped logging individual rejected cell values/exception text; the loop now aggregates failures per column and emits a single summary warning (`kept N cell(s) that could not convert to <type>`).

All 19 tests in `test_table_chunk_column_roles.py` pass locally, and `ruff check`/`ruff format --check` are clean on both changed files (findings count on `table.py` actually dropped from 13 to 12 pre-existing issues since removing the blind-except cleared one BLE001).

Ready for re-review.

xugangqiang:
LGTM

PR #18461, opened 2026-08-18
fix(memory): honor custom extraction prompts

### LINKED ISSUE #18413: [Bug]: Memory system_prompt and user_prompt are stored and returned by the API but never reach the extraction LLM

yzl0ng:
### Summary

- Forward each memory's stored system_prompt and user_prompt to extract_by_llm.
- Cover both immediate save and queued extraction paths with focused regression tests.
- Preserve the existing default-prompt fallback when stored prompts are empty.

Fixes #18413.

### Testing

- Added two async regression tests covering both call sites.
- Verified both tests fail before the fix and pass after it.
- Ruff check and format validation pass for the changed test and files.

### DIFF
--- a/api/db/joint_services/memory_message_service.py
+++ b/api/db/joint_services/memory_message_service.py
@@ -60,6 +60,8 @@ async def save_to_memory(memory_id: str, message_dict: dict):
             get_memory_type_human(memory.memory_type),
             message_dict.get("user_input", ""),
             message_dict.get("agent_response", ""),
+            system_prompt=memory.system_prompt,
+            user_prompt=memory.user_prompt,
             llm_id=memory.llm_id,
         )
         if memory.memory_type != MemoryType.RAW.value
@@ -124,6 +126,8 @@ async def save_extracted_to_memory_only(memory_id: str, message_dict, source_mes
         get_memory_type_human(memory.memory_type),
         message_dict.get("user_input", ""),
         message_dict.get("agent_response", ""),
+        system_prompt=memory.system_prompt,
+        user_prompt=memory.user_prompt,
         task_id=task_id,
         llm_id=memory.llm_id,
     )
--- a/test/unit_test/api/db/joint_services/test_memory_message_service.py
+++ b/test/unit_test/api/db/joint_services/test_memory_message_service.py
@@ -1,0 +1,82 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


Lynn-Inf (approved):
LGTM

PR #18462, opened 2026-08-18
fix(memory): require semantic valid_at timestamp

yzl0ng:
## Summary

- require a `valid_at` timestamp in the semantic-memory output schema
- tell the extraction model to use conversation time when a fact has no date of its own
- add a regression test for the assembled semantic prompt

This addresses the deterministic prompt inconsistency reported in #18415. The invalid-timestamp fallback is intentionally left unchanged because selecting its replacement policy requires a separate design decision.

## Testing

- `uvx --from pytest --with pytest-asyncio pytest --noconftest -q test/unit_test/memory/utils/test_prompt_util.py`
- verified the regression test fails before the fix and passes after it
- `uvx ruff check test/unit_test/memory/utils/test_prompt_util.py`
- `uvx ruff format --check memory/utils/prompt_util.py test/unit_test/memory/utils/test_prompt_util.py`
- `python -m py_compile memory/utils/prompt_util.py test/unit_test/memory/utils/test_prompt_util.py`

### DIFF
--- a/memory/utils/prompt_util.py
+++ b/memory/utils/prompt_util.py
@@ -77,7 +77,7 @@ You are an expert at analyzing conversations to extract structured memory.
         "semantic": [
             {
                 "content": "Clear factual statement",
-                "valid_at": "timestamp or empty",
+                "valid_at": "timestamp — use the conversation time when the fact has no date of its own",
                 "invalid_at": "timestamp or empty"
             }
         ]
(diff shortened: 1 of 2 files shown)


Lynn-Inf (commented):
No need to enforce `valid_at` default in templates — it's already defined in `prompt_util.py#L48`. Also, good to see the `or empty` removed on line 80 — thanks!

PR #3875, opened 2024-12-05
Feat: Add question parameter to edit chunk modal

KevinHuSh:
### What problem does this PR solve?

Close #3873

### Type of change

- [x] New Feature (non-breaking change which adds functionality)

PR #18473, opened 2026-08-18
fix(rag/nlp): include question_tks when reranking with a rerank model

### LINKED ISSUE #18472: [Bug]: rerank_by_model drops question_tks, so a configured rerank model never sees chunk questions
### Self Checks

68f826082

### RAGFlow image version

v0.26.4

- Not environment specific. Affects any dataset whose chunks carry generated questions (`question_kwd` / `question_tks`) when a rerank model is configured.

### Actual behavior

`Dealer` in `rag/nlp/search.py` has three rerank paths. Each builds a per-chunk token list (`ins_tw`) that feeds the term-similarity score. Two of them include the chunk's generated-questions field; the third does not.

| path | line | tokens |
|---|---|---|
| `rerank_with_knn` | 462 | `content_ltks + title_tks*2 + important_kwd*5 + question_tks*6` |
| `rerank` | 494 | `content_ltks + title_tks*2 + important_kwd*5 + question_tks*6` |
| `rerank_by_model` | 516 | `content_ltks + title_tks + important_kwd` |

`rerank_by_model` omits `question_tks` entirely.

This matters more than a missing term in one score, because in that path `ins_tw` is also what gets handed to the reranker model:

```python
docs = [remove_redundant_spaces(" ".join(tks)) for tks in ins_tw]

tksim = self.qryr.token_similarity(keywords, ins_tw)
vtsim, _ = rerank_mdl.similarity(query, docs)
```

So with a rerank model configured, the chunk's questions are excluded from **both** the term-similarity score and the text the cross-encoder scores. The field is still fetched from the index (`rag/nlp/search.py:163`), so the cost is paid and the value discarded.

Meanwhile the retrieval query boosts `question_tks^20` (`rag/nlp/query.py:37`) — the highest field boost in the system. Retrieval treats generated questions as the strongest available signal, and then model-reranking throws them away.

### Root cause

A missed call site, not a deliberate exclusion:

- `rerank_by_model` was added in `614defec2` (2024-05-29, "add rerank model" #969).
- `question_tks` was introduced by `56f473b68` (2024-12-05, "Feat: Add question parameter to edit chunk modal" #3875). That commit added `question_kwd`/`question_tks` to the fetched source-field list and to `rerank()`'s token list, but never touched `rerank_by_model`, which had already existed for six months.

`git show 56f473b68 -- rag/nlp/search.py` shows the change confined to the field list and the single `tks = ...` line inside `rerank()`.

### Expected behavior

All three rerank paths should let the chunk's generated questions contribute to the ranking, and the reranker model should see them in the document text it scores.

Note that `rerank_by_model` deliberately omits the `*2/*5/*6` multipliers the other two paths apply — its tokens are joined back into a string for a cross-encoder, where repeating a field would distort the model's own scoring. So `question_tks` should be added unweighted, consistent with how `title_tks` and `important_kwd` are handled there, rather than copying the `*6` used elsewhere.

### Steps to reproduce

1. Create a dataset and enable question generation so chunks carry `question_kwd` / `question_tks`.
2. Configure a rerank model on the retrieval/chat settings.
3. Ask a question phrased closely to a chunk's generated question but sharing few content terms.

The chunk ranks lower than it does with the same setup and no rerank model, because the questions never reach the reranker.

Deterministic reproduction against `rag/nlp/search.py` without a live index:

```python
from unittest.mock import MagicMock
import numpy as np
import common.settings  # noqa: F401  (import-order: see rag.utils.redis_conn cycle)
from rag.nlp.search import Dealer

captured = {}
dealer = Dealer.__new__(Dealer)
qryr = MagicMock()
qryr.question.return_value = (None, ["apple"])
qryr.token_similarity.side_effect = lambda kw, ins_tw: captured.setdefault("ins_tw", ins_tw) and [0.5]
dealer.qryr = qryr

sres = Dealer.SearchResult(
    total=1, ids=["c1"], query_vector=[0.1, 0.2, 0.3],
    field={"c1": {
        "content_ltks": "apple sells phones",
        "title_tks": "apple report",
        "question_tks": "what does apple sell",
        "important_kwd": ["revenue"],
    }},
)

mdl = MagicMock()
mdl.similarity.return_value = (np.array([0.5]), 0)
dealer.rerank_by_model(mdl, sres, "apple")

print(mdl.similarity.call_args[0][1])
# ['apple sells phones apple report revenue']   <- no "what does apple sell"
```

### Additional information

`rag/nlp` currently has no unit tests at all (`test/unit_test/rag/` covers `app`, `llm`, `graphrag`, `svr`, `utils`, `flow`, `advanced_rag`, `prompts` — but not `nlp`), which is likely why the divergence went unnoticed for ~20 months.

A fix adding `question_tks` to `rerank_by_model`, plus the first `test/unit_test/rag/nlp/` tests pinning the field set for all three rerank paths, is ready; PR to follow.

immuhammadfurqan:
### Summary

Closes #18472.

`Dealer` in `rag/nlp/search.py` has three rerank paths. Each builds a per-chunk token list (`ins_tw`) that feeds the term-similarity score. Two include the chunk's generated-questions field; `rerank_by_model` does not.

| path | line | tokens |
|---|---|---|
| `rerank_with_knn` | 462 | `content_ltks + title_tks*2 + important_kwd*5 + question_tks*6` |
| `rerank` | 494 | `content_ltks + title_tks*2 + important_kwd*5 + question_tks*6` |
| `rerank_by_model` | 516 | `content_ltks + title_tks + important_kwd` |

In `rerank_by_model` that list is also what gets handed to the reranker:

```python
docs = [remove_redundant_spaces(" ".join(tks)) for tks in ins_tw]

tksim = self.qryr.token_similarity(keywords, ins_tw)
vtsim, _ = rerank_mdl.similarity(query, docs)
```

So with a rerank model configured, the chunk's questions are excluded from **both** the term-similarity score and the text the cross-encoder scores. The field is still fetched from the index (`search.py:163`), so the retrieval cost is paid and the value discarded — while the query itself boosts `question_tks^20` (`rag/nlp/query.py:37`), the highest field boost in the system.

**This is a missed call site, not a deliberate exclusion.** `rerank_by_model` was added in `614defec2` (2024-05-29, #969). `question_tks` was introduced by `56f473b68` (2024-12-05, #3875), which added it to the fetched source-field list and to `rerank()` but never touched `rerank_by_model`, already six months old at that point.

### What changed

One line in `rerank_by_model`, adding `question_tks` to the token list.

Added **unweighted**, deliberately unlike the other two paths. `rerank_by_model` already omits the `*2/*5/*6` multipliers because its tokens are joined back into a string for a cross-encoder, where repeating a field would distort the model's own scoring. Copying the `*6` from `rerank()` would emit `question_tks` six times into `docs`. A short comment records that reasoning so the divergence doesn't read as a second oversight later.

### Testing

Adds `test/unit_test/rag/nlp/test_search_rerank.py` — the first unit tests for `rag/nlp`, which currently has none (`test/unit_test/rag/` covers `app`, `llm`, `graphrag`, `svr`, `utils`, `flow`, `advanced_rag` and `prompts`, but not `nlp`). That gap is the likely reason this went unnoticed for ~20 months.

Six tests pin the token-assembly contract:

- `question_tks` reaches both `token_similarity` and the `docs` handed to the reranker
- `rerank_by_model` does not repeat fields (guards the intentional no-multiplier choice above)
- a parametrized parity check across all three paths
- graceful handling of chunks indexed before question generation, where `question_tks` is absent

Verified against `68f826082`. With `rag/nlp/search.py` reverted to `main`, exactly the three `rerank_by_model` tests fail while the other paths pass:

```
FAILED test_rerank_by_model_includes_question_tks - AssertionError: assert ('what' in ['apple', 'sells', 'phones', 'apple', 're...
FAILED test_rerank_by_model_does_not_repeat_fields - AssertionError: assert 0 == 1
FAILED test_all_rerank_paths_include_question_tks[rerank_by_model] - AssertionError: rerank_by_model dropped question_tks
========================= 3 failed, 3 passed =========================
```

The parametrized case makes the asymmetry explicit: `[rerank]` and `[rerank_with_knn]` pass on unpatched `main`, `[rerank_by_model]` fails. With the fix, all 6 pass.

End-to-end effect on the text the reranker receives, same fixture, before and after:

```
before: ['apple sells phones apple report revenue']
after:  ['apple sells phones apple report revenue what does apple sell']
```

### Note for reviewers

The new test imports `common.settings` before `rag.nlp.search`. `rag.nlp.query` imports `rag.utils.redis_conn`, which imports `common.settings`, which imports `rag.utils.redis_conn` back; the cycle only resolves when `common.settings` is initialised first. The API server and task executor do that at startup, so the test reproduces the runtime import order rather than papering over it. Happy to move this into a `test/unit_test/rag/nlp/conftest.py` if that reads better as more tests land in this directory.

### DIFF
--- a/internal/service/nlp/reranker.go
+++ b/internal/service/nlp/reranker.go
@@ -134,12 +134,16 @@ func RerankByModel(
 		contentLtks := extractContentTokens(chunk, cfield)
 		titleTks := extractTitleTokens(chunk)
 		importantKwd := extractImportantKeywords(chunk)
+		questionTks := extractQuestionTokens(chunk)
 
-		// Combine tokens without repetition (simpler version for model reranking)
-		tks := make([]string, 0, len(contentLtks)+len(titleTks)+len(importantKwd))
+		// Unlike RerankStandard/RerankWithKNN, the fields are not repeated here:
+		// these tokens are joined back into `docs` for a cross-encoder, where
+		// duplicating a field would distort the model's own scoring.
+		tks := make([]string, 0, len(contentLtks)+len(titleTks)+len(importantKwd)+len(questionTks))
 		tks = append(tks, contentLtks...)
 		tks = append(tks, titleTks...)
 		tks = append(tks, importantKwd...)
+		tks = append(tks, questionTks...)
 		insTw = append(insTw, tks)
 
 		// Build document text for model reranking
--- a/internal/service/nlp/reranker_question_tks_test.go
+++ b/internal/service/nlp/reranker_question_tks_test.go
@@ -1,0 +1,117 @@
+// Copyright 2025 The InfiniFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nlp
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"ragflow/internal/common"
+	"ragflow/internal/entity/models"
+)
+
+// captureDriver records the documents RerankByModel hands to the cross-encoder,
+// which is the only observable the token assembly produces. The embedded
+// interface satisfies the rest of models.ModelDriver; it is never called.
+type captureDriver struct {
+	models.ModelDriver
+	documents []string
+}
+
+func (d *captureDriver) Rerank(_ context.Context, _ *string, request models.RerankRequest, _ *models.APIConfig, _ *models.RerankConfig, _ *common.ModelUsage) (*models.RerankResponse, error) {
+	d.documents = request.Documents
+	return &models.RerankResponse{}, nil
+}
+
+// rerankOneChunk runs RerankByModel over a single chunk and returns the
+// document text the reranker model received for it.
+func rerankOneChunk(t *testing.T, chunk map[string]interface{}) string {
+	t.Helper()
+
+	driver := &captureDriver{}
+	ids := []string{"c1"}
+	field := map[string]map[string]interface{}{"c1": chunk}
+
+	RerankByModel(
+		context.Background(),
+		&models.RerankModel{ModelDriver: driver},
+		[]map[string]interface{}{chunk},
+		ids,
+		field,
+		"epsilon",
+		0.3,
+		0.7,
+		"content_ltks",
+		nil,
+		nil,
+	)
+
+	if len(driver.documents) != 1 {
+		t.Fatalf("expected 1 document to reach the reranker model, got %d", len(driver.documents))
+	}
+	return driver.documents[0]
+}
+
+func fullChunk() map[string]interface{} {
+	return map[string]interface{}{
+		"content_ltks":  "alpha beta",
+		"title_tks":     "gamma",
+		"important_kwd": []string{"delta"},
+		"question_tks":  "epsilon zeta",
+	}
+}
+
+// TestRerankByModelIncludesQuestionTks pins the fix: RerankStandard and
+// RerankWithKNN both fold question_tks into their token lists, so a chunk
+// whose generated questions carry the match was scored on those tokens on
+// every path except this one, where the cross-encoder never saw them.
+func TestRerankByModelIncludesQuestionTks(t *testing.T) {
+	doc := rerankOneChunk(t, fullChunk())
+
+	for _, tk := range []string{"alpha", "beta", "gamma", "delta", "epsilon", "zeta"} {
+		if !strings.Contains(doc, tk) {
+			t.Errorf("document %q is missing token %q", doc, tk)
+		}
+	}
+}
+
+// TestRerankByModelDoesNotRepeatFields guards the deliberate difference from
+// RerankStandard/RerankWithKNN, which weight fields by repeating them: these
+// tokens are joined back into a document for a cross-encoder, so a repeated
+// field would distort the model's own scoring.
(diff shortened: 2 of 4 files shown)


qinling0210 (changes requested):
@immuhammadfurqan Please do similar code change on go side.

We expect we can modify both python and go sides if the issue exists on both sides in a PR. Thanks.

immuhammadfurqan:
Thanks @qinling0210 — done, pushed in b70e777.

The Go side had the same gap. `internal/service/nlp/reranker.go`:

- `RerankStandard` (L300) and `RerankWithKNN` (L962) both call `extractQuestionTokens` and fold `question_tks` in with weight 6.
- `RerankByModel` (L134) built its token list from `content_ltks + title_tks + important_kwd` only.

`extractQuestionTokens` (L565) already existed — it was just never called on the reranker-model path. So a chunk whose generated questions carried the match was scored on those tokens on every retrieval path except this one, matching the Python behaviour reported in #18472.

As on the Python side, the fields are **not** repeated here. These tokens get joined back into `docs` and handed to a cross-encoder, so duplicating a field would distort the model's own scoring — the repetition in `RerankStandard`/`RerankWithKNN` only makes sense because those feed a term-weight similarity instead.

The live call site is `RetrievalService.scoreSearchResult` (`internal/service/nlp/retrieval.go:463`), which passes real `IDs`/`Field`, so this path is exercised in production.

**Tests** — `internal/service/nlp/reranker_question_tks_test.go`, unit tier (no build tag, no external services). A fake `ModelDriver` captures the documents handed to the cross-encoder, which is the only observable the token assembly produces:

- `TestRerankByModelIncludesQuestionTks` — all four fields reach the model.
- `TestRerankByModelDoesNotRepeatFields` — each field appears exactly once.
- `TestRerankByModelWithoutQuestionTks` — chunks with no `question_tks` (the common case) are unaffected.

The first two fail against the old code; the third is the control.

One note in the interest of transparency: I don't have a Go toolchain in my environment, so I could not run `build.sh --test` locally — I verified the change by inspection against the two sibling paths. CI runs `./build.sh --test` over `internal/service/nlp`, so it will be covered there. Happy to adjust if anything comes back red.

qinling0210 (approved):
LGTM

PR #18479, opened 2026-08-19
feat: improve Go wiki compilation

buua436:

Improve Go Wiki compilation with chunk-level MAP caching, topic-aware planning, entity/topic page merging, and correct source chunk counting.

### Type of change


wangq8 (commented):
Thanks for this PR — the immutable per-chunk MAP cache and the entity/topic page split are a solid step forward, and the test coverage is good. I left a few findings below, ordered by importance.

## 1. Stale embedding after evidence union / LLM rewrite (important)

`wikiEntityMerge` (and `mergeTopicPage`, which calls it) now unions the existing and incoming Markdown into `merged.Content`, but keeps only the incoming vector:

```go
merged.Content = unionWikiMarkdown(existing.Content, incoming.Content)
merged.Vector = incoming.Vector
```

The same issue applies in `synthesizeEntityPage`: after the LLM rewrites the merged body, `merged.Content` is replaced, yet `merged.Vector` is still `incoming.Vector` (set earlier inside `wikiEntityMerge`).

`engineWriter.WriteMerged` persists `p.Vector` verbatim without re-embedding (writer.go: `m["q_%d_vec"] = p.Vector`). The previous `wikiReplace` explicitly warned about this: a vector that doesn't match the stored Markdown breaks KNN similarity search. Now the merged/rewritten content differs from the embedding even more, so dataset-level KNN recall for entity and topic pages will be skewed toward only the newest evidence block. Consider re-embedding the merged content (or the LLM output in `synthesizeEntityPage`) before writing, or at least documenting the tradeoff and its impact on incremental topic routing.

## 2. Topic pages can grow without bound (medium)

Topic pages go through `mergeTopicPage` → `wikiEntityMerge` → `unionWikiMarkdown`, which dedups only *exact* block matches (after `TrimSpace`). Entity pages get a clean LLM rewrite via `synthesizeEntityPage`, but topic pages do not. Across many incremental recompiles with slightly reworded content, near-duplicate blocks will accumulate and the topic page will grow monotonically. Worth confirming this is intended and whether a semantic dedup / rewrite step (like the entity path) or a size cap is needed for topic pages.

## 3. Zero-value extracts appended for uncacheable chunks (medium)

In `runVersionedMap`, `extracts` is allocated at `len(p.inputs.Chunks)`, and for chunks without a stable `ID` the corresponding slot stays a zero-value `wikiExtract{}`. At the end:

```go
p.mapExtracts = append(p.mapExtracts, extracts...)
p.mapExtracts = append(p.mapExtracts, uncachedExtracts...)
```

So every uncacheable chunk contributes an empty extract entry into the reduce/plan input in addition to its real extract in `uncachedExtracts`. This is harmless if downstream ignores empty extracts, but it can produce empty reduce batches and wastes tokens; consider filtering empty extracts before appending.

## 4. MAP cache relies on `available_int=0` rows being retrievable (medium)

`wikiMapVersionRow` stores cache rows with `available_int: 0` (intentionally, to keep them non-searchable), but `GetWikiMapVersions` does not add `available_int` to its filter — it searches by `id` + `compile_kwd` only. This works only if the engine returns `available_int=0` rows on a direct id-filtered lookup for both Infinity and Elasticsearch. If either engine implicitly excludes `available_int=0`, the cache becomes write-only (never hits), and the duplicate-suppression check in `PutWikiMapVersions` (which calls `GetWikiMapVersions`) would also silently fail, allowing duplicate rows. Please confirm engine semantics on this path.

## 5. `synthesizeEntityPage` resolves a hardcoded model (low)

The entity rewrite call uses `kccommon.ResolveDeps(tenant, defaultLLMID, defaultEmbedding)` and then chats with `defaultLLMID`. This may not match the dataset's actual configured chat model for the wiki variant, and it introduces a per-entity-page LLM round-trip that isn't part of the existing prompt-budget accounting. The fallback to the deterministic union is good; a test covering the LLM path (and its failure fallback) would help.

## 6. `RouteTopic` silently no-ops when `Chat` is nil (low)

`RouteTopic` returns `(false, "", nil)` when `x.decider == nil || x.decider.Chat == nil`, which in `processBatch` routes every non-exact-topic candidate into `unmatched`. That's a safe fallback, but a debug log here would make silent misconfiguration easier to diagnose.

## 7. New mapping fields only in Infinity schema (low / confirm)

`plan_group_kwd` and `generation_kwd` were added to `conf/infinity_mapping.json` only. They are *selected* (not filtered) in the reader paths, so on Elasticsearch/OpenSearch (dynamic mapping) this is likely harmless, but please confirm the ES/OS chunk schemas don't need an explicit mapping entry for consistency with the Infinity mapping.

## Nit

- `productContents` (added in `consumer.go`) doesn't appear to be referenced anywhere — dead code; can be removed.

Nothing blocking here, but items 1–2 are worth resolving (or explicitly documenting) before this ships, since they affect retrieval quality and storage growth over time.

internal/ingestion/component/knowledge_compiler/common/deps.go
```diff
@@ -71,17 +71,26 @@ type HistoricalKNN interface {
...
-	Summary        string
-	ContentMD      string
-	ContentMDRaw   string
-	EntityNames    []string
+	ID           string
+	Slug         string
+	Title        string
+	PageType     string
+	Topic        string
+	PlanGroup    string
```
yuzhichang: Add a comment on what's this field.

PR #18480, opened 2026-08-19
fix(ingestion): keep Q&A/Tag chunks from collapsing into one indexed chunk

euvre:
### Summary

Parsing a two-column Q&A workbook (xlsx) with the **Q&A** chunk method produced a single chunk containing only the last row of the file, instead of one chunk per question-answer pair (reported as "the Q&A chunk method's chunk content is not ideal").

**Root cause (Go ingestion pipeline):** `QAChunker` (and the sibling `TagChunker`) emitted chunks whose content lived only in `content_with_weight`, while every other chunker emits the pipeline's canonical `text` field. `ProcessChunksForPipeline` derives the index chunk id from `chunk["text"]`, so all Q&A pairs hashed the same empty string plus doc id into ONE id. The index write is an upsert by id, so each pair overwrote the previous one and only the last row survived (`countDistinctChunkIDs == 1`). The same shape also made the Tokenizer skip these chunks for embedding (it embeds the `text` field), so the surviving chunk carried no vector.

**Fix:**

- `QAChunker`/`TagChunker` now emit chunk content through `text`, matching the documented chunker contract ("the chunker must emit string text"); ingestion renames `text` -> `content_with_weight` at the index boundary exactly as for every other chunk method.
- Spreadsheet Q&A pairs now carry `top_int` (pair index), mirroring Python `rag/app/qa.py`'s `beAdoc(..., row_num=ii)`.
- `ProcessChunksForPipeline` hashes `content_with_weight` when `text` is absent, mirroring Python `rag/svr/task_executor.py:407` (`xxh64(content_with_weight + doc_id)`), so distinct content can never collapse onto a shared empty-text id again.

**Verification:**

- Reproduced before the fix on a local deployment: a 10-row Q&A xlsx parsed with the Q&A method gave `chunk_count: 1` with the last row only, and its id equalled `xxh64("" + doc_id)`.
- After the fix (same flow through the web UI): `chunk_count: 10`, 10 distinct chunk ids, and the parsed-chunk list shows one 问题/回答 card per workbook row.
- Added regression tests: QA HTML-table chunks carry `text` + `top_int` and survive `ProcessChunksForPipeline` with N distinct ids; `content_with_weight`-only chunks get distinct ids.
- `bash build.sh --test ./internal/ingestion/component/chunker/ ./internal/ingestion/task/indexdoc/ ./internal/ingestion/task/` passes.


wangq8 (commented):
LGTM. The fix is correct and defensive on both fronts:

- Emitting chunk content through `text` matches the documented chunker contract (ingestion renames `text` → `content_with_weight` at the index boundary), so Q&A/Tag chunks now participate in embedding and hash to distinct IDs.
- The `ProcessChunksForPipeline` fallback that hashes `content_with_weight` when `text` is absent mirrors Python `task_executor.py:407` and closes the general "empty-text id collapse" class of bug, not just this specific case.
- Setting `RowNum: len(pairs)` in `extractQATable` restores correct 0-based `top_int` (previously the HTML-table path left `RowNum` at its zero value, so every pair carried `top_int=[0]`), matching Python's `enumerate`/`row_num=ii`.

Regression coverage is solid: the chunker-level tests pin `text` + `top_int`, and `TestQAChunker_ChunksGetDistinctIndexIDs` plus `TestProcessChunksForPipeline_ContentWithWeightOnlyDistinctIDs` guard the end-to-end invariant (N distinct IDs). Nothing blocking from my side.

xugangqiang:
Please resolve conflict.

xugangqiang:
Thanks for the PR. I looked closely at the newly added `text == "" → content_with_weight` fallback in `process.go`, and I think it is now redundant — would you reconsider dropping it?

The fallback guards against a chunk that carries `content_with_weight` but no `text`, so that hashing the empty `text` wouldn't collapse every sibling Q&A pair into a single chunk id. There were only two historical producers of that shape:

1. `TagChunker` (`tag.go`) — emitted `content_with_weight` only. It was **deleted by #18523** (merged), so it no longer exists on `main`.
2. `QAChunker` — this PR itself fixes it to emit `Text` (the canonical carrier) instead of only `ContentWithWeight`.

After this PR + #18523, no chunker emits `content_with_weight`-only anymore, so the `text == ""` branch can never be hit.

On top of that, in the production path `register.go`'s `imageUploadDecorator` unconditionally sets `ck["id"] = common.ChunkID(docID, text)` for every chunk. So the surrounding `if _, exists := ck["id"]; !exists` block in `process.go` does not even run in production — it is only reachable from the direct `ProcessChunksForPipeline` test path, where again there is no `content_with_weight`-only producer.

So the fallback protects a scenario that no longer exists, and it diverges from the framework contract (the `chunker` emits `text`; `RenameTextToContentWithWeight` promotes it to `content_with_weight` at the index boundary). Would it make sense to remove the added fallback lines from `process.go` and keep this PR minimal to the `Text` fix?

One more point for discussion: `RowNum: len(pairs)` → `top_int` is not needed for the id (the id is derived from `text`), so it does not contribute to the "Q&A collapse" fix — it only mirrors Python's `beAdoc(..., row_num=ii)`. Since the stated goal of this PR is to stop the collapse, would it be cleaner to move the `top_int` / parity addition to a separate change?

xugangqiang:
Follow-up to make the "redundant fallback" point conclusive: I grepped all 17 non-test chunker files on `main` (`common.go`, `group.go`, `hierarchy.go`, `image_upload.go`, `manual.go`, `one.go`, `page.go`, `pdfcrop_cgo.go`, `pdfcrop_nocgo.go`, `qa.go`, `register.go`, `sentence_boundary.go`, `table.go`, `title.go`, `title_cap.go`, `token.go`).

- **Only `qa.go` (line 135) sets `ContentWithWeight` without a `Text` field.** Every other chunker emits `Text`.
- `tag.go` is already absent from `main` (deleted by #18523).

So today `qa.go` is the *sole* producer of `content_with_weight`-only chunks — precisely the scenario the `process.go` fallback shields against. This PR already fixes that producer (qa.go now emits `Text`). Once it merges, **zero** chunkers emit `content_with_weight`-only, so the added `text == ""` branch can never execute. The fallback is therefore conclusively dead code; dropping it keeps the PR minimal to the `Text` fix.

internal/ingestion/task/indexdoc/process.go
```diff
@@ -86,7 +86,17 @@ func ProcessChunksForPipeline(
...
+			// their text (which RenameTextToContentWithWeight promotes to
+			// content_with_weight below); chunks that already carry
+			// content_with_weight hash it directly, mirroring Python
+			// rag/svr/task_executor.py:407. Hashing the empty text of a
+			// content_with_weight-only chunk would give every sibling the
+			// same id and the index write would keep only the last one.
 			text, _ := ck["text"].(string)
+			if text == "" {
+				text, _ = ck["content_with_weight"].(string)
+			}
```
xugangqiang: this is not needed. Please do not change framework.

euvre:
Thanks for the reviews — summary of where each point landed:\n\n**1. Remove the `text == ""` → `content_with_weight` fallback in `process.go` ("not needed / please do not change framework")** — @xugangqiang agreed and done in 8659235f8 (pushed right after your comment). Before pushing I independently re-verified your analysis: with this PR's `Text` fix in `qa.go`, no chunker emits a content_with_weight-only chunk anymore — the only remaining non-test `ContentWithWeight` references under `chunker/` are `common.go`'s `itemText()` reader (which prefers `Text`, so the text carrier is never empty) and `group.go`, which only copies `ContentWithWeight` into extra metadata while the record's text carrier still comes from `itemText()`. `internal/ingestion/task/indexdoc/` (both `process.go` and `process_test.go`, including the dedicated fallback test) is now byte-identical to `main`, and the PR diff is limited to `internal/ingestion/component/chunker/` (`qa.go` + tests).\n\n**2. Conflict (Aug 20)** — resolved: the branch was rebased and GitHub now reports the PR as mergeable.\n\n**3. `RowNum` → `top_int` parity in a separate change?** — I kept it in this PR for now: it is a 3-line parity fix inside `extractQATable` (the very function this PR fixes), restoring the 0-based `top_int` that Python's `beAdoc(..., row_num=ii)` produces, and it does not touch the framework. Happy to split it out into its own PR if you prefer — just say so and I'll move it.\n\n**4.** @wangq8 thank you for the thorough review! One update: the `ProcessChunksForPipeline` fallback you mentioned was later removed at the maintainer's request (point 1), so the PR is now minimal to the chunker-side `Text` fix plus the `top_int` parity; the end-to-end distinct-ID guard you called out (`TestQAChunker_ChunksGetDistinctIndexIDs`, which still runs real chunker output through `ProcessChunksForPipeline`) is retained.\n\nVerification: `bash build.sh --test ./internal/ingestion/component/chunker/... ./internal/ingestion/task/indexdoc/...` — all green (`ok ragflow/internal/ingestion/component/chunker`, `ok ragflow/internal/ingestion/task/indexdoc`).

xugangqiang:
pls fix ci issue

xugangqiang:
LGTM

PR #18482, opened 2026-08-19
fix(chat): clamp out-of-range Max Tokens input and show valid-range hint

euvre:
### Summary

Fixes the "Max tokens" control in chat settings (and every other slider + number-input pair built on `NumberInput`) reporting a wrong, silently-invalid state when a value beyond the slider maximum is typed.

**Problem** (reported in chat settings → Model settings → Max tokens, slider max 128000):

1. Typing `888888` left the slider stuck at the last in-range prefix (≈88888, ~2/3 of the track) instead of reaching the maximum, while the input kept displaying the out-of-range number. No hint, tooltip, or validation message appeared at any point.
2. Root cause of the missing clamp: the blur-time normalization added in #17921 never ran for form-bound usages. `NumberInput` rendered `{...props}` **after** its own `onBlur`, so the `onBlur` that arrives via `{...field}` (react-hook-form) silently replaced the clamping handler. Combined with out-of-range keystrokes being dropped without feedback, the control could sit in an inconsistent state (display 888888 / slider 88888) indefinitely.

**Fix:**

- `web/src/components/originui/number-input.tsx`
  - Propagate the **clamped** (effective) value on every keystroke so bound sliders immediately reflect it, while the raw typed draft stays visible until blur. Intermediate edits such as `102` with `min=512` remain possible (the #17921 behavior is preserved, only the stale-freeze is removed).
  - Render `{...props}` before the internal `onChange`/`onBlur`/`onFocus` handlers so parent props can no longer disable blur normalization.
  - Track focus via a ref so form echoes of the clamped value don't snap the display back mid-edit; clamp increment/decrement results; drop stale out-of-range drafts when the input becomes disabled; expose an opt-in `onRangeViolation` callback.
- `web/src/components/llm-setting-items/slider.tsx`: render a localized range hint while the typed value is out of range; the hint clears on blur, slider drag, or field disable.
- Locales: add `chat.valueRangeMessage` ("Please enter a value between {{min}} and {{max}}.") to all 16 locale files.
- Tests: new `web/src/components/originui/__tests__/number-input.test.tsx` covering clamped propagation of out-of-range input, blur normalization + violation reset, in-range passthrough, below-min intermediate editing, focus-guarded external sync, and disable-time draft reset.

**Verification:**

- `npx jest src/components/originui/__tests__/number-input.test.tsx` — 6/6 pass.
- `npx oxlint` clean on the touched component/test files; `npm run type-check` reports no errors in touched files (remaining repo-wide errors are pre-existing and untouched).
- End-to-end in the browser (chat → chat settings → Model settings → Max tokens): typing `888888` now pins the slider at 128000 and shows the hint "Please enter a value between 0 and 128000."; blurring normalizes the input to 128000 and removes the hint; toggling the Max tokens switch off/on leaves no stale draft; dragging the slider keeps the input in sync. No LLM calls were involved, so no external-service dependency limits the verification.


wangq8 (commented):
The core fix is sound and well-tested: spreading `{...props}` before the internal `onChange`/`onBlur`/`onFocus` handlers restores the blur-time clamping that #17921 intended, and propagating the clamped value on every keystroke keeps the bound slider from going stale. The `onRangeViolation` opt-in + localized range hint are clean.

One concern worth addressing before merge (moderate): `handleBlur`/`handleFocus` now shadow the `onBlur`/`onFocus` that arrive via `{...field}` (react-hook-form), and `handleBlur` does not forward to the parent `onBlur`. For RHF consumers (e.g. `slider.tsx` and `message-history-window-size-item.tsx` spread `{...field}`), the field's `touched` state and any `onBlur`-triggered validation are no longer invoked. Value updates still flow through the explicit `onChange`, and the forms here use `mode: 'onChange'`, so this is latent rather than visibly broken today — but it's a real regression for a generic component. Recommend calling the original `onBlur`/`onFocus` (from `props`) at the end of the internal handlers after normalization, e.g. capture `props.onBlur` and invoke it inside `handleBlur`.

Minor: in `slider.tsx`, `disabled` is passed before `{...field}`; since `field.disabled` is undefined here it's fine, but if a future caller sets `disabled` on `FormField`, the spread would override the explicit prop. Worth a glance but not blocking.

wangq8:
@euvre resolve conflict.

euvre:
already resolved.

PR #18483, opened 2026-08-19
fix: preserve extensionless document suffix on GaussDB

Sevenzuo:
### Summary

RAGFlow's "Create empty document" flow accepts names without a file extension. The `POST /datasets/<dataset_id>/documents?type=empty` route calls `_upload_empty_document()`, where `Path(name).suffix.lstrip(".")` returns `""`.

In GaussDB's A/ORA compatibility mode, that empty string is persisted as SQL `NULL`. Because `document.suffix` was defined as `NOT NULL`, the insert failed with a constraint violation.

### Reproduction

GaussDB extensionless empty document creation failure

### Changes

- Use the existing `EmptyStringCharField` for `Document.suffix`.
- Apply the same field type in the historical add-column migration.
- Relax `document.suffix` to nullable for existing GaussDB metadata schemas.
- Add regression coverage for field conversion, migration ordering, and non-GaussDB behavior.

### Compatibility

Only GaussDB maps the application-level empty string to SQL `NULL` and restores it to `""` on reads. MySQL, PostgreSQL, and OceanBase retain the original `NOT NULL` constraint and continue storing `""` unchanged.

### Tests

- `python -m pytest test/unit_test/api/db/test_gaussdb_adaptation_points.py test/unit_test/api/db/test_gaussdb_user_flow.py::test_gaussdb_empty_string_compatible_fields_are_nullable_only_for_gaussdb -q` — 18 passed
- `lefthook run pre-commit` — all checks passed

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -1321,7 +1321,7 @@ class Document(DataBaseModel):
     progress_msg = TextField(null=True, help_text="process message", default="")
     process_begin_at = DateTimeField(null=True, index=True)
     process_duration = FloatField(default=0)
-    suffix = CharField(max_length=32, null=False, help_text="The real file extension suffix", index=True)
+    suffix = EmptyStringCharField(max_length=32, null=False, help_text="The real file extension suffix", index=True)
 
     content_hash = CharField(max_length=32, null=True, help_text="xxhash128 of document content for change detection", default="", index=True)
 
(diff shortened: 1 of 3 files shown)


Lynn-Inf (commented):
LGTM. Please also update `entity.Document.Suffix` in Go — this ensures the field is properly adapted in the Go `ragflow_server` when connecting to GaussDB.

Sevenzuo:
@Lynn-Inf Thanks. The Go `ragflow_server` does not support GaussDB metadata yet. We will adapt `entity.Document.Suffix` together with the Go GaussDB implementation in a follow-up PR.

Lynn-Inf (approved):
Thanks — merging as soon as CI is green.

PR #18484, opened 2026-08-19
fix(ingestion): show data pipeline column for builtin pipeline ingestion logs

euvre:
### Summary

The dataset "Logs" page (`/dataset/logs/:id`) renders an "Ingestion pipeline" column from the `pipeline_title` field of `pipeline_operation_log` rows. For documents parsed with a builtin pipeline (no user-created ingestion pipeline), the Go ingestion runtime wrote the operation log with an empty `pipeline_title`, a nil `avatar`, and a bogus `pipeline_id` (the parser id), so the column showed only a placeholder avatar with no text.

Root cause: `PipelineExecutor.recordPipelineLog` resolved the pipeline identity solely from `user_canvas` by `s.canvasID`. For builtin runs the canvas id is the document's `parser_id` (set in `Ingestor.defaultRunDocumentTask`), which never exists as a `user_canvas` row, so the title lookup failed silently and the log row was persisted with blank identity fields.

The Python implementation (`PipelineOperationLogService.create` in `api/db/services/pipeline_operation_log_service.py`) falls back to `document.parser_id` as the title and `document.thumbnail` as the avatar when the task carries no user pipeline id. This PR aligns the Go runtime with that behavior:

- Add `PipelineExecutor.WithBuiltinPipeline()`; `defaultRunDocumentTask` marks registry-backed runs so their operation log skips the `user_canvas` lookup entirely.
- `recordPipelineLog` now defaults `pipeline_title` to the document's `parser_id` and `avatar` to the document thumbnail, and leaves `pipeline_id` null for builtin runs (custom-canvas runs keep the canvas id/title/avatar, and a canvas deleted mid-run falls back to the parser id instead of logging a blank title).
- Unit tests cover the builtin fallback, the custom-canvas path, and the missing-canvas fallback.

Verified end to end on the Go stack: re-parsed a `general/qa` document; the new log row returns `pipeline_title: "qa"`, `pipeline_id: null` and the Logs page renders the pipeline name in the column, while pre-fix rows remain blank as expected (historical data).

### DIFF
--- a/internal/ingestion/service/ingestion_service.go
+++ b/internal/ingestion/service/ingestion_service.go
@@ -964,7 +964,7 @@ func (e *Ingestor) defaultRunDocumentTask(ctx context.Context, ingestionTask *en
 	}
 	if isBuiltin {
 		// Builtin path: load DSL from the embedded registry, skipping canvas DB lookup.
-		executor.WithBuiltinPipeline().WithLoadDSLFunc(func(ctx context.Context, _ string) (string, string, error) {
+		executor.WithLoadDSLFunc(func(ctx context.Context, _ string) (string, string, error) {
 			common.Info(fmt.Sprintf("load built in DSL for: %s", parserID))
 			dsl, lerr := pipelinepkg.LoadBuiltinDSL(parserID)
 			if lerr != nil {
--- a/internal/ingestion/service/ingestion_service.go
+++ b/internal/ingestion/service/ingestion_service.go
@@ -964,7 +964,7 @@ func (e *Ingestor) defaultRunDocumentTask(ctx context.Context, ingestionTask *en
 	}
 	if isBuiltin {
 		// Builtin path: load DSL from the embedded registry, skipping canvas DB lookup.
-		executor.WithLoadDSLFunc(func(ctx context.Context, _ string) (string, string, error) {
+		executor.WithBuiltinPipeline().WithLoadDSLFunc(func(ctx context.Context, _ string) (string, string, error) {
 			common.Info(fmt.Sprintf("load built in DSL for: %s", parserID))
 			dsl, lerr := pipelinepkg.LoadBuiltinDSL(parserID)
 			if lerr != nil {
--- a/internal/ingestion/task/pipeline_executor.go
+++ b/internal/ingestion/task/pipeline_executor.go
@@ -65,6 +65,11 @@ type PipelineExecutor struct {
 	taskCtx     *TaskContext
 	canvasID    string
 	docBulkSize int
+	// builtinPipeline marks a run backed by the embedded builtin pipeline
+	// registry instead of a user_canvas row. canvasID then carries the
+	// parser_id, so the operation log must fall back to document fields for
+	// its pipeline identity (mirroring the Python operation-log path).
+	builtinPipeline bool
 
 	indexWriter     *chunkIndexWriter
 	logCreateFunc   func(ctx context.Context, db *gorm.DB, log *entity.PipelineOperationLog) error
@@ -137,6 +142,15 @@ func (s *PipelineExecutor) WithLogCreateFunc(f func(ctx context.Context, db *gor
 
 func (s *PipelineExecutor) WithLoadDSLFunc(f func(ctx context.Context, canvasID string) (string, string, error)) *PipelineExecutor {
 	s.loadDSLFunc = f
+	return s
+}
+
+// WithBuiltinPipeline marks the run as a builtin (registry-backed) pipeline so
+// the operation log skips the user_canvas lookup and titles itself from the
+// document's parser_id, mirroring record_pipeline_operation in
+// api/db/services/pipeline_operation_log_service.py.
+func (s *PipelineExecutor) WithBuiltinPipeline() *PipelineExecutor {
+	s.builtinPipeline = true
 	return s
 }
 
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
LGTM overall. The fix correctly mirrors the Python `PipelineOperationLogService.create` fallback: builtin runs skip the `user_canvas` lookup, title the log with `document.parser_id`, reuse `document.thumbnail` as avatar, and leave `pipeline_id` null. The types line up (`doc.Thumbnail` and `PipelineOperationLog.Avatar` are both `*string`), and the test coverage for builtin / custom-canvas / missing-canvas is good.

One nit (non-blocking): the new guard `!dao.IsNotFoundErr(err)` won't actually suppress the warning for the "canvas deleted mid-run" fallback. `dao.IsNotFoundErr` checks `errors.Is(err, gorm.ErrRecordNotFound)`, but `UserCanvasDAO.GetByID` returns the sentinel `ErrUserCanvasNotFound` (a plain `errors.New`, not wrapping `gorm.ErrRecordNotFound`) on missing rows. So the expected not-found case will still emit `failed to reload pipeline ... for operation log`. The fallback itself works correctly (title stays `parser_id`); it's just a spurious warning log. Consider using `errors.Is(err, dao.ErrUserCanvasNotFound)` instead.

xugangqiang:
LGTM

euvre:
Thanks for the reviews — all three items are addressed in `a8c2d9e`.

**1. @wangq8 — spurious warning guard (verified, applied).** Confirmed: `UserCanvasDAO.GetByID` converts `gorm.ErrRecordNotFound` into the `ErrUserCanvasNotFound` sentinel (`internal/dao/user_canvas.go`), while `dao.IsNotFoundErr` checks `errors.Is(err, gorm.ErrRecordNotFound)`, so the guard never matched and the expected "canvas deleted mid-run" fallback still logged `failed to reload pipeline ... for operation log`. The guard now checks `errors.Is(err, dao.ErrUserCanvasNotFound)`, so the not-found fallback is silent and only real DB errors warn.

**2. CodeRabbit — assert `PipelineID` on the missing-custom-canvas fallback (applied).** `TestRecordPipelineLog_CustomCanvasMissingFallsBackToParserID` now also asserts `captured.PipelineID == "canvas-gone"`, pinning the non-builtin identity contract when the canvas row is gone.

**3. CodeRabbit — Python-parity wording in new comments (applied).** Removed the Python file/mirroring references from the three comment blocks added by this PR (`builtinPipeline` field, `WithBuiltinPipeline`, and the pipeline-identity comment in `recordPipelineLog`); the behavior descriptions are kept. Pre-existing parity comments elsewhere in the file were left untouched to keep this PR scoped.

**Verification:** `bash build.sh --test ./internal/ingestion/task/...` passes (incl. the updated test). No shared-service end-to-end run was performed for this follow-up; evidence is unit tests plus code-level verification against `internal/dao/user_canvas.go` and `api/db/services/pipeline_operation_log_service.py`.

xugangqiang:
Please on hold the merge - I need to have a second check whether the design is good enough.

xugangqiang:
@euvre 

Thanks for the fix — the behavior is correct and matches the Python path. I'd suggest a simpler implementation though: the `WithBuiltinPipeline()` flag is redundant, because the builtin-vs-canvas distinction already exists in the executor's own state.

`TaskContext.PipelineID` is resolved from the document's raw `pipeline_id` (`task_context.go` `resolvePipelineID` / `LoadFromIngestionTask`) and `defaultRunDocumentTask` never overwrites it — it only assigns the local `pipelineID` variable to `parserID` before calling `NewPipelineExecutor`. So inside `recordPipelineLog`, `s.taskCtx.PipelineID == ""` is exactly the same predicate as `isBuiltin := pipelineID == ""` in `defaultRunDocumentTask`, and mirrors Python's `if pipeline_id:` in `pipeline_operation_log_service.py:167`.

A canvas-backed run always carries a non-empty doc `pipeline_id`, so the flag can be derived instead of stored:

```go
pipelineTitle := doc.ParserID
pipelineAvatar := doc.Thumbnail
var pipelineID *string
if s.taskCtx.PipelineID != "" { // custom canvas run
    pipelineID = &s.canvasID
    if db != nil {
        if canvas, err := dao.NewUserCanvasDAO().GetByID(ctx, db, s.canvasID); err == nil && canvas != nil {
            if canvas.Title != nil {
                pipelineTitle = *canvas.Title
            }
            pipelineAvatar = canvas.Avatar
        } else if err != nil && !errors.Is(err, dao.ErrUserCanvasNotFound) {
            common.Warn(fmt.Sprintf("failed to reload pipeline %s for operation log: %v", s.canvasID, err))
        }
    }
}
```

Behavior stays identical (builtin → `parser_id` title + document thumbnail + nil `pipeline_id`; canvas deleted mid-run → `parser_id` fallback), while removing the `builtinPipeline` field, the `WithBuiltinPipeline()` setter, and the one-line change in `ingestion_service.go`. It also avoids a second copy of the state that any future `NewPipelineExecutor` caller would have to remember to set.

For the tests: the builtin case can just use the default `makeTaskCtx()` (its `PipelineID` is already `""`), and the custom-canvas tests should set `taskCtx.PipelineID` to the canvas id — which matches production, where a canvas-backed document carries the canvas id as its `pipeline_id`.

euvre:
@xugangqiang Thanks for the design review — your suggestion is verified and applied in `74e744f`. Merge stays on hold pending your second check, of course.

**Derived builtin-vs-canvas from `TaskContext.PipelineID` (verified, applied).** Every claim checked out against the code:

- `LoadFromIngestionTask` resolves `TaskContext.PipelineID` from the document's raw `pipeline_id` via `resolvePipelineID` (`internal/ingestion/task/task_context.go`), and `defaultRunDocumentTask` never overwrites it — it only reassigns the local `pipelineID` variable to `parserID` before `NewPipelineExecutor`, so `taskCtx.PipelineID == ""` is exactly the `isBuiltin` predicate.
- This mirrors Python's `if pipeline_id:` in `pipeline_operation_log_service.py`, and the debug path (`debug.go`) already sets `PipelineID` to the canvas id, so the derived predicate holds there too.

Changes: dropped the `builtinPipeline` field and `WithBuiltinPipeline()` setter, `recordPipelineLog` now keys the fallback on `s.taskCtx.PipelineID != ""`, and the builtin path in `ingestion_service.go` no longer sets the flag. Behavior is identical (builtin → `parser_id` title + document thumbnail + nil `pipeline_id`; canvas deleted mid-run → `parser_id` fallback, silent on the not-found sentinel). Tests follow your suggestion: the builtin case uses the default `makeTaskCtx()`, the custom-canvas tests set `taskCtx.PipelineID` to the canvas id, and `TestPipelineExecutor_Run_MainFlowWithStubs` now sets `PipelineID = "flow-1"` to match production.

Verification: `bash build.sh --test ./internal/ingestion/task/... ./internal/ingestion/service/... ./internal/handler/...` — all green. Boundary note: no end-to-end verification against the shared ragflow services from this environment; the evidence is unit tests plus line-level Python parity analysis.

Also thanks for the earlier LGTM — the three prior items (wangq8's warning-guard fix and the two CodeRabbit nits) remain addressed in `a8c2d9e` as summarized above.

xugangqiang:
LGTM

PR #18486, opened 2026-08-19
Fix: Deleting a large language model selected on the memory page from the model vendor should result in an error message when accessing the memory page.

cike8899:
### Summary

Fix: Deleting a large language model selected on the memory page from the model vendor should result in an error message when accessing the memory page.


wangq8 (commented):
LGTM. The change is internally consistent and correctly separates the two model references:

- `superRefine` now validates `llm_id` unconditionally (after `modelsFetched`) while gating `embd_id` on `modelsEditable`, matching the intent that a stale chat model must always surface an error but a stale embedding model is tolerated once the select is read-only.
- `useRevalidatePersistedModels` mirrors the same split: `llm_id` is always re-triggered, `embd_id` is re-triggered only while editable and cleared otherwise. The dependencies are correct.

One thing worth a quick confirmation (non-blocking): when messages already exist (`modelsEditable === false`), a deleted `llm_id` now surfaces an error that blocks submit, but the LLM select is disabled in that state, so the user has no in-place way to correct it. If the only recovery path is to delete/recreate the memory, that's fine — just want to make sure that's the intended UX rather than an accidental dead-end.

PR #18489, opened 2026-08-19
Fix: connector validation error

mkaaad:
surface connector validation error detail so test connect can return detail error


Haruko386 (approved):
LGTM

PR #18490, opened 2026-08-19
Fix ingestion pipeline child delimiter extra newline

wangq8:


wangq8 (commented):
Thanks for the fix. The direction is correct and the Go/Python change is symmetric (`strings.TrimPrefix(s, "\n")` ≡ `s.removeprefix("\n")`), which keeps the two implementations in lock-step. A few things to confirm before merge:

**1. Child `text` still keeps the leading `\n` (split source is untrimmed).**
In `splitByChildren` you trim `mom` but still split the untrimmed text:

```go
mom := strings.TrimPrefix(ck.Text, "\n")
parts := splitDroppingDelim(ck.Text, pattern) // still ck.Text, not mom
```

When the children delimiter is not `"\n"` itself (e.g. a sentence delimiter), the first child's `Text` still starts with `"\n"` while its `Mom` no longer does — so the parent reference and the child content now disagree about whether the leading newline exists. If the goal is to drop the extra leading newline everywhere, split from the trimmed value instead: `parts := splitDroppingDelim(mom, pattern)`. (Python `_split_chunk_docs_by_children` has the same shape — splits the untrimmed `chunk.get("text")` — so parity is preserved either way; just please confirm the child text is intended to retain the `"\n"`.)

**2. The Go text-path change looks like a no-op.**
Both callers of `applyChildrenDelimText` (`chunkPerSegment` and `mergeByTokenSize`) already `strings.TrimSpace` the parent before calling it, so `t` never has a leading `"\n"` and `strings.TrimPrefix(t, "\n")` never fires. The Python text path, by contrast, runs `naive_merge` (which prepends `"\n"` per paragraph) and does *not* trim before the children split, so the Python `removeprefix("\n")` is genuinely load-bearing. Net effect: this PR changes Python text-path output but leaves Go text-path output byte-identical. That asymmetry is fine (it actually closes an existing Go/Python gap), but the Go `applyChildrenDelimText` hunk reads as dead code — consider dropping it or adding a one-line comment explaining it's a defensive no-op.

**3. `TrimPrefix`/`removeprefix` strip exactly one newline.**
If a chunk can begin with multiple leading newlines (`"\n\n..."`), only the first is removed. In the merge path exactly one `"\n"` is prepended so this is safe, but it's worth a quick confirm that the JSON path (parser-provided item text) can't emit two.

**4. Please add a regression test.**
There's already `token_children_delim_test.go` asserting `mom`; a case where the parent text starts with `"\n"` (e.g. `delimiters: ["\n"]` + a children delimiter on the JSON path) would pin this behavior and prevent future regressions. The PR currently changes no tests.

Minor: the PR body is empty — a one-line description of the reproduction (which delimiter config leaks the newline into `mom`) would help reviewers and future maintainers.

(Note: I couldn't run the Go chunker tests locally — the package fails to build here due to a missing `office_oxide.h` cgo dependency, unrelated to this change.)

PR #18492, opened 2026-08-19
Fix: not use connect check result as model verify result

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/internal/handler/providers.go
+++ b/internal/handler/providers.go
@@ -502,7 +502,6 @@ type AlterProviderInstanceRequest struct {
 	BaseURL      string                            `json:"base_url"`
 	Region       string                            `json:"region"`
 	ModelInfo    []service.CreateInstanceModelInfo `json:"model_info"`
-	Verify       *bool                             `json:"verify"`
 }
 
 func (h *ProviderHandler) AlterProviderInstance(c *gin.Context) {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Reviewed. The change is correct and I'm good with the direction.

**What it does:** Removes `verifyProviderAPIKey` (which called the driver-level `CheckConnection` and stamped the single connection result onto *every* model's `verify` status) from both `CreateProviderInstance` and `AlterProviderInstance`, and drops the now-unused `Verify` request field.

**Why it's right:** A connection-level check (`driver.CheckConnection`) only tells you the API key/endpoint works — it is not a per-model verification. Per-model verification is already done properly by `verifyProviderModel` inside the `CheckConnection` service path (`/connection/verify`), which persists results via `updateModelVerifyResults`. Reusing the connection check result here was double-counting/misleading.

**Checks I ran:**
- No remaining callers/references to `verifyProviderAPIKey` after removal.
- The only caller of `AlterProviderInstance` (handler `providers.go`) is updated to the new signature; the `internal/admin` package has its own separate `AlterProviderInstance`, unaffected.
- The `verify` extra field removal is safe: the model-list reader (`model_service.go` ~L2348) already defaults `verify` to `"unknown"` when absent, so no consumers break.
- `entity.ModelVerify*` constants and `strings`/`modelModule` imports remain used elsewhere, so no unused-import/dead-code fallout.
- No test coverage was tied to the removed `verifyProviderAPIKey` or the verify-on-create/alter behavior, so no test updates are required.

**Nits (non-blocking):**
- PR body is just "As title." — a sentence explaining the motivation (connection check ≠ per-model verify) would help future readers.
- This is a deliberate API/behavior change (models no longer get a `verify` field stamped on create/alter). Worth confirming the frontend doesn't rely on `extra.verify` being present immediately after creation.

Note: I couldn't run a full `go build` locally — the repo fails on an unrelated CGO dependency (`office_oxide` missing `office_oxide.h` header). This is an environment issue, not related to this PR.

Nothing blocking. LGTM modulo the nits.

PR #18497, opened 2026-08-19
Fix search with [+-] crash the search

wangq8:


wangq8 (commented):
Thanks for the fix — the direction (short-circuiting queries that have nothing to tokenize) is right, but I think the guard doesn't actually fire in the backend that the surrounding comment is about. Flagging that below.

**Main concern: the guard is a no-op when `DOC_ENGINE_INFINITY` is true.**

`rag_tokenizer.tokenize()` does not tokenize when Infinity is the document engine — it returns the input unchanged:

```python
# rag/nlp/rag_tokenizer.py
def tokenize(self, line: str) -> str:
    if settings.DOC_ENGINE_INFINITY:
        return line
    else:
        return super().tokenize(line)
```

So for a query like `+-` (or a lone `+` / `-`), `rag_tokenizer.tokenize("+-").strip()` is `"+-"` — truthy — and the new early return never fires. Execution falls through to the non-Chinese branch, `q` ends up empty, and the existing fallback `if not q: q.append(txt)` appends the raw `"+-"` back into the query, producing `MatchTextExpr(..., "+-", ...)`. That is exactly the unescaped `+`/`-` that the comment says Infinity's lexer chokes on. So the crash is likely still reproducible on the Infinity backend.

**Suggestions**

1. Base the guard on the already-stripped `txt` rather than on `tokenize()`, whose behavior is backend-dependent. e.g.:

   ```python
   if not txt or not txt.strip("+- "):
       return None, []
   ```

   or, more generally, escape the fallback instead of dropping it.

2. Alternatively, fix the root cause in the non-Chinese branch: instead of `if not q: q.append(txt)`, return `None` (or escape the raw text via `self.sub_special_char`) when no valid term was produced, so the raw unescaped special characters never reach `MatchTextExpr`.

**Minor / non-blocking**

- Returning `None, []` is consistent with the Chinese branch (`return None, keywords`) and callers already handle a falsy `matchText` (`[matchText] if matchText else []`), so that part is fine.
- There's no Python unit test covering `Query.question()`. A small regression test for empty/special-only inputs (`+`, `-`, `+-`, `()`, `*`) would help lock this in.

Could you confirm which `DOC_ENGINE` reproduced the crash, and whether it still reproduces with this change?

qinling0210 (approved):
LGTM

PR #18503, opened 2026-08-19
Fix empty Markdown document parsing

guominjia:
## Summary
- treat empty Markdown binaries as in-memory content instead of local file paths
- add a regression test ensuring empty content does not access the filesystem

## Testing
- python -m py_compile rag/app/naive.py test/unit_test/rag/app/test_markdown_image_ssrf.py
- Targeted pytest could not run because the local environment lacks pytest and dependency resolution is blocked by an authenticated Gitee dependency.

### DIFF
--- a/rag/app/naive.py
+++ b/rag/app/naive.py
@@ -837,7 +837,7 @@ class Markdown(MarkdownParser):
 
     def __call__(self, filename, binary=None, separate_tables=True, delimiter=None, return_section_images=False):
         """Parse markdown into text sections and optional standalone table chunks."""
-        if binary:
+        if binary is not None:
             encoding = find_codec(binary)
             txt = binary.decode(encoding, errors="ignore")
         else:
--- a/test/unit_test/rag/app/test_markdown_image_ssrf.py
+++ b/test/unit_test/rag/app/test_markdown_image_ssrf.py
@@ -92,6 +92,21 @@ def parser(naive_module):
 
 
 @pytest.mark.p1
+def test_parses_empty_binary_without_opening_filename(parser):
+    with patch("builtins.open") as open_file:
+        sections, tables, section_images = parser(
+            "empty-document.md",
+            binary=b"",
+            return_section_images=True,
+        )
+
+    open_file.assert_not_called()
+    assert sections == []
+    assert tables == []
+    assert section_images == []
+
+
+@pytest.mark.p1
 def test_blocks_internal_url_without_fetching(parser):
     """A markdown image pointing at an internal host must never be requested."""
     with (


guominjia:
```markdown
## Background

### Symptom

`docker-ragflow-cpu-1` continuously logs chunking failures for documents ingested through the GitHub connector, one failure per affected document, with no progress ever completing:

```
progress_msg: [ERROR][Exception]: [Errno 2] No such file or directory: '<pr_number>:<pr_title>.md'

ERROR handle_task got exception for task {...}
FileNotFoundError: [Errno 2] No such file or directory: '<pr_number>:<pr_title>.md'
```

Every failing document name follows the pattern `{github_number}:{title}.md`, i.e. these are all documents synced from GitHub pull requests / issues, not manually uploaded files. The container itself is healthy (no restarts, normal CPU/memory) — this is purely an application-level bug in the ingestion pipeline.

### Root cause

The failure spans three stages of the ingestion pipeline:

**1. GitHub connector produces an empty blob when a PR/Issue has no body text**

`common/data_source/github/connector.py`, `_convert_pr_to_document()` / `_convert_issue_to_document()`:

```python
file_content_byte = pull_request.body.encode("utf-8") if pull_request.body else b""
name = sanitize_filename(pull_request.title, "md")
return Document(
    blob=file_content_byte,
    semantic_identifier=f"{pull_request.number}:{name}",
    ...
)
```

Many small/maintenance PRs (and issues) legitimately have an empty description. In that case `blob` is explicitly set to `b""` (empty bytes, not `None`). This is expected/reasonable behavior on the connector side.

**2. The empty blob is persisted to MinIO and read back successfully**

`api/db/services/connector_service.py`, `duplicate_and_parse()` wraps the blob into a `FileObj` and calls `FileService.upload_document()`, which stores the empty file in MinIO without error.

Later, during chunking, `rag/svr/task_executor_refactor/task_handler.py`, `_run_standard_chunking_impl()`:

```python
binary = await self._get_storage_binary(bucket, name)
if binary is None:
    raise FileNotFoundError(f"Can not find file <{ctx.name}> from minio. Could you try it again.")

chunks = await chunk_service.build_chunks(binary)
```

`binary` here is `b""`, which is **not** `None`, so this guard correctly does not trip, and the empty binary is passed further down as legitimate (if empty) document content.

**3. The markdown parser conflates "empty content" with "no content provided"**

`rag/app/naive.py`, `Markdown.__call__()`:

```python
def __call__(self, filename, binary=None, separate_tables=True, delimiter=None, return_section_images=False):
    if binary:
        encoding = find_codec(binary)
        txt = binary.decode(encoding, errors="ignore")
    else:
        with open(filename, "r") as f:
            txt = f.read()
```

`if binary:` is a truthiness check, so `b""` (empty-but-present) takes the same branch as `binary is None` (no binary supplied at all, local-file mode). It falls into the `else` branch and tries to `open(filename, "r")`, where `filename` is actually the document's semantic identifier, not a filesystem path — hence `FileNotFoundError`.

### Impact

Every GitHub-sourced document whose PR/Issue body is empty fails chunking with an unhandled `FileNotFoundError`, permanently stuck at `progress: -1` (`[ERROR][Exception]`). The document is never indexed and the task is not retried automatically. This is silent from a container-health perspective — no crash, no restart — so it can only be found by grepping the task executor logs.

Note that the "zero-chunk" case is already handled gracefully elsewhere in the pipeline — `task_handler.py` has an explicit `if not chunks: progress_cb(msg=f"No chunk built from {ctx.name}")` branch — but the crash in `naive.py` happens before that code can ever be reached.

### Proposed fix

Change the truthiness check to an explicit `None` check in `rag/app/naive.py`'s `Markdown.__call__` (and audit sibling parsers, e.g. `TxtParser`, for the same pattern):

```python
if binary is not None:
    ...
else:
    with open(filename, "r") as f:
        ...
```

With this change, a PR/Issue with an empty body decodes to an empty string, produces zero sections/chunks, and is handled by the existing "no chunk built" path instead of crashing.

PR #18504, opened 2026-08-19
fix(chat): drain deep-research goroutines before closing the async channel

euvre:
### Summary

The Go api server crashed with `panic: send on closed channel`, taking the
whole process down and surfacing as `Request error 502` in the web UI.

Panic path (from the server log):

- `AsyncChat` hands DeepResearcher an inline progress callback that sends on
  the pipeline's result channel; the pipeline goroutine closes that channel
  via `defer close(out)` when it returns.
- On request cancellation (user leaves the SSE chat page), the recursive
  `_research` tree returned early via `ctx.Done()` **without waiting for its
  parallel sub-research goroutines**. The orphaned goroutines kept invoking
  the progress callback, whose send then hit the already-closed channel and
  killed the process.

Fix (defense in depth):

1. `_research` now drains its sub-research goroutines (`wg.Wait()`) on the
   cancellation path before returning, mirroring Python's
   `asyncio.gather` cancellation semantics, and bails out early when the
   context is already canceled so a cancellation unwinds the tree instead of
   scheduling more retrieval work.
2. The inline callback is extracted into `deepResearchProgressCallback` with
   a `select` that drops progress deltas once the consumer is gone
   (`ctx.Done()`), so progress can never block or panic on a dead channel.

Tests:

- `TestDeepResearcherCancelDrainsSubResearch` — regression test reproducing
  the crash window: cancels mid-research while parallel sub-research
  goroutines are parked in retrieval, and asserts Research does not return
  before they drain and that no callback fires afterwards.
- `TestDeepResearchProgressCallbackDeliversMarkers` — marker translation
  (`<retrieving>` / `</retrieving>`) and non-final delivery.
- `TestDeepResearchProgressCallbackDropsAfterConsumerGone` — progress is
  dropped instead of blocking when the consumer disappeared.

### DIFF
--- a/internal/service/chat_pipeline.go
+++ b/internal/service/chat_pipeline.go
@@ -651,31 +651,7 @@ func (s *ChatPipelineService) AsyncChat(
 					)
 					question := strings.Join(questions, " ")
 
-					drErr := dr.Research(ctx, kbinfos, question, question, func(msg string) {
-						switch {
-						case strings.HasPrefix(msg, "<START_DEEP_RESEARCH>"):
-							out <- AsyncChatResult{
-								Answer:      "<retrieving>",
-								Reference:   map[string]interface{}{},
-								AudioBinary: nil,
-								Final:       false,
-							}
-						case strings.HasPrefix(msg, "<END_DEEP_RESEARCH>"):
-							out <- AsyncChatResult{
-								Answer:      "</retrieving>",
-								Reference:   map[string]interface{}{},
-								AudioBinary: nil,
-								Final:       false,
-							}
-						default:
-							out <- AsyncChatResult{
-								Answer:      msg,
-								Reference:   map[string]interface{}{},
-								AudioBinary: nil,
-								Final:       false,
-							}
-						}
-					})
+					drErr := dr.Research(ctx, kbinfos, question, question, s.deepResearchProgressCallback(ctx, out))
 					if drErr != nil {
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Reviewed — thanks for the thorough fix and regression tests. The root cause (orphaned sub-research goroutines sending on a channel the pipeline closes after `Research` returns) is correctly identified and addressed in depth.

What I verified:
- `deepResearchProgressCallback` wraps every send in a `select` on `ctx.Done()`, so a progress delta can never block or panic on a dead channel once the consumer is gone.
- `_research` now `wg.Wait()`s on the cancellation branch and bails early on `ctx.Err()`, so the whole tree drains before `Research` returns and the pipeline's `defer close(out)` runs. Because each recursion level waits for its own children, no callback fires after `Research` returns.

One thing to keep in mind (non-blocking): `wg.Wait()` assumes sub-goroutines become unblocked on cancellation. The callback path is covered by the `select`, and the LLM driver checks `ctx.Err()`, but a sub-goroutine already parked in `retrieveInformation` (KB/web retrieval) that doesn't respect `ctx` would make `Research` block until that I/O returns, instead of the previous fast-return-then-panic. That's strictly better than crashing the process, but the failure mode becomes a goroutine leak rather than a panic — worth a follow-up check that `retSvc.Retrieval` and the web-search path honor the context.

Tests look good: the regression test reproduces the crash window and asserts no callback fires after return. LGTM.

PR #18507, opened 2026-08-19
fix(deepdoc): collapse overlapping OCR boxes before Go table cell-fill

### LINKED ISSUE #18499: Go PDF parser: table cell text fill runs before vertical merge, duplicating overlapping OCR boxes
## Summary

Go PDF parser's table cell text fill (`FillCellTextFromBoxes`, `internal/deepdoc/parser/pdf/table/table_cells.go`) runs **before** the document-level vertical merge (`NaiveVerticalMerge` in `buildLayout`), while Python fills cells **after** its `_naive_vertical_merge`. Overlapping OCR boxes are therefore collapsed in Python but kept separate in Go, producing duplicated cell text.

## Impact

A pipeline-parity test on the 35-PDF snapshot set (`TestPipelineParity`, `internal/deepdoc/parser/pdf/pipeline_parity_test.go`) shows `14_text_table_interleaved.pdf` at gridSim 97.5% (was 71.4% before the TSR replay fixes; the residual is this cell-fill difference). Duplicated text in cells also affects downstream chunk/retrieval for production Go parses of tables whose OCR detect emits overlapping boxes.

## Root cause

- Python: `_parse_loaded_window_into_bboxes` runs `_naive_vertical_merge` (pdf_parser.py:1827) **before** `_table_transformer_job` → `construct_table` (deepdoc/vision/table_structure_recognizer.py:156), which marks R/C and fills cells. So OCR boxes overlapping in Y are merged first, and only the surviving box participates in cell fill.
- Go: `processPage` → `processOneTable` (table_extract.go) runs `FillCellTextFromBoxes` on the **raw** OCR boxes, before `buildLayout`'s `NaiveVerticalMerge` (parser.go:510). Two overlapping OCR boxes both match the same row band by vertical-overlap ratio and their texts concatenate.

## Concrete example (14_text_table_interleaved.pdf, table 2)

OCR boxes:
- `"Software Hardware"` y=(270.0, 300.0)
- `"Hardware"` y=(287.7, 300.0)  (overlaps the first)

TSR row bands (page points): row1 y=(265.2,283.7), row2 y=(283.5,301.3).

- Go: both boxes match row2 (overlap ratio 0.53 vs 0.44) → cell text `"Software Hardware Hardware"`, row1 empty.
- Python: the boxes are collapsed downstream and the final HTML keeps `"Software Hardware"` in row1, row2 empty.

Replicating Python's `find_overlapped_with_threshold` + `gather`/`layouts_cleanup` on the dumped TSR rows still selects row2, so the divergence comes from the box set that reaches `construct_table` after Python's downstream merge — the exact intermediate `self.boxes` state is not in the dumps.

## Suggested fix

Make Go's table cell fill see the same box set Python sees: either (a) run the overlapping-box collapse (or `NaiveVerticalMerge`) on the per-table boxes **before** `FillCellTextFromBoxes` in `processOneTable`, or (b) replicate Python's R/C marking (`gather`/`layouts_cleanup` filtered rows) inside `FillCellTextFromBoxes`. Option (a) is lower risk. Instrumenting Python's `self.boxes` mid-pipeline would confirm which box set reaches `construct_table` before finalizing.

pangwangshu:
## Summary

Fixes #18499.

Go's per-page table cell-fill (`FillCellTextFromBoxes` in `internal/deepdoc/parser/pdf/table/table_cells.go`) ran on raw OCR boxes **before** any vertical-merge collapse. Python's pipeline runs `_naive_vertical_merge` **before** its equivalent `construct_table` step, so overlapping OCR boxes are merged first and only the surviving box participates in cell fill.

Because of this ordering difference, two overlapping/duplicate OCR detections (e.g. an OCR box for `"Software Hardware"` and a second, nested detection for just `"Hardware"`) could each independently match a table row band by vertical-overlap ratio, spreading duplicated text across cells instead of collapsing into one — see the issue for the concrete repro on `14_text_table_interleaved.pdf`.

## Fix

In `processOneTable` ([table_extract.go](internal/deepdoc/parser/pdf/table_extract.go)), run the per-table box subset through `layout.NaiveVerticalMerge` — the same function `buildLayout` already runs document-wide, later in the pipeline — **before** `FillCellTextFromBoxes`. This mirrors Python's pipeline order and collapses overlapping boxes before cell assignment rather than after.

This was called out in the issue as the lower-risk of the two suggested fixes (collapse boxes first vs. replicating Python's R/C marking inside `FillCellTextFromBoxes`).

## Test plan

- [x] `go vet ./internal/deepdoc/parser/pdf/...` — clean
- [x] Added `TestProcessOneTable_CollapsesOverlappingBoxesBeforeCellFill` (`table_extract_test.go`), which reproduces the issue's overlapping-box pattern: fails before the fix (duplicated text spread across two cells), passes after (boxes collapse into a single cell).
- [x] Existing tests in the pure-Go `layout`/`table` packages still pass (`go test ./internal/deepdoc/parser/pdf/layout/... ./internal/deepdoc/parser/pdf/table/...`)
- [ ] Could not run the full native-linked suite (`bash build.sh --test ./internal/deepdoc/parser/pdf/...`) in this environment — the repo's prebuilt native PDF libs (`pdfium`/`office_oxide`/`pdf_oxide`) are Linux-only and this sandbox is macOS/arm64. @xugangqiang, since you filed the issue and have the most context on `TestPipelineParity` / the 35-PDF snapshot set, could you confirm the `14_text_table_interleaved.pdf` gridSim score on your end?

### DIFF
--- a/internal/deepdoc/parser/pdf/table_extract.go
+++ b/internal/deepdoc/parser/pdf/table_extract.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"math"
 
+	lyt "ragflow/internal/deepdoc/parser/pdf/layout"
 	tbl "ragflow/internal/deepdoc/parser/pdf/table"
 	pdf "ragflow/internal/deepdoc/parser/pdf/type"
 	util "ragflow/internal/deepdoc/parser/pdf/util"
@@ -116,12 +117,24 @@ func (p *Parser) processOneTable(ctx context.Context, pageImg image.Image, boxes
 		if firstCellTop == 1e9 {
 			firstCellTop = cells[0].Y0
 		}
-		boxInCrop = make([]pdf.TextBox, 0, len(tm.BoxIdx))
+		tableBoxes := make([]pdf.TextBox, 0, len(tm.BoxIdx))
 		for _, idx := range tm.BoxIdx {
 			b := boxes[idx]
 			if b.Bottom*scale-cropOffY < firstCellTop {
 				continue
 			}
+			tableBoxes = append(tableBoxes, b)
+		}
+		// Collapse overlapping/adjacent OCR boxes before cell-fill, mirroring
+		// Python's pipeline order: _naive_vertical_merge runs before
+		// construct_table, so overlapping boxes are merged before they reach
+		// cell assignment. Go runs table cell-fill per-page (here), before the
+		// document-wide vertical merge in buildLayout, so it must run its own
+		// collapse on this table's box subset first or overlapping OCR boxes
+		// duplicate text across cells.
+		tableBoxes = lyt.NaiveVerticalMerge(tableBoxes, nil, nil, nil)
+		boxInCrop = make([]pdf.TextBox, 0, len(tableBoxes))
+		for _, b := range tableBoxes {
 			boxInCrop = append(boxInCrop, tbl.BoxToCropSpace(b, scale, cropOffX, cropOffY))
 		}
 	}
--- a/internal/deepdoc/parser/pdf/table_extract_test.go
+++ b/internal/deepdoc/parser/pdf/table_extract_test.go
@@ -225,3 +225,74 @@ func TestProcessOneTable_NoPerCellOCR(t *testing.T) {
 		})
 	}
 }
+
+// twoRowTableBuilder groups DetectCells' output into one row per cell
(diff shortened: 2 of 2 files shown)


xugangqiang:
@yuzhichang  LGTM.

PR #16483, opened 2026-06-30
fix(common/time_utils): correct None/empty timestamp fallback and ISO 8601 parsing

Harsh23Kashyap:
Recovery PR for #16173 after the fork branch was accidentally reset during rewrite-cleanup.

Cherry-picked onto current `main`:
- fix(common/time_utils): correct fallback timestamp and ISO-8601 normalization
- fix(common/time_utils): preserve zero timestamps and mark regression tests
- test(common/time_utils): make fallback assertions deterministic

Supersedes closed #16173 — same branch `Harsh23Kashyap/fix/time-utils-edgecases`, rebuilt per @yuzhichang recovery steps in https://github.com/infiniflow/ragflow/pull/16173#issuecomment-4829663835

PR #18508, opened 2026-08-19
fix(memory): surface unparseable timestamps and require valid_at for semantic (#18415)

### LINKED ISSUE #18415: [Bug]: Memory extraction stores an empty valid_at for semantic items, and an unparseable timestamp is written through unchanged

Taranum01:
Fixes #18415.

Two small defects combine so that semantic memory items are stored with no validity timestamp, silently:

1. The extraction prompt's semantic output template declared `valid_at` as "timestamp or empty" — the episodic/procedural templates require it. With empty declared valid, the LLM omitted it on ~6.5% of semantic extractions.
2. `format_iso_8601_to_ymd_hms` caught the dateutil parse exception and returned the original unparsed string verbatim — the log line carried only the exception text, not the rejected value, so the failures were unactionable as logged and the garbage ended up in a timestamp column.

## Fix

- `format_iso_8601_to_ymd_hms` now:
  - returns `""` for empty/None input (matches the existing convention used by the caller for `invalid_at` when the model omits it);
  - logs the rejected value AND exception on parse failure, so operators can identify which field failed;
  - uses `dateutil.parser.isoparse` directly. The previous implementation called `isoparse` only to validate, then handed the string to stdlib `datetime.fromisoformat` which rejects valid ISO 8601 ordinal dates (`2024-001T12:00:00Z`) — those now parse correctly.
- `PromptAssembler.OUTPUT_TEMPLATES[semantic].valid_at` is now required, mirroring the surrounding "Default: valid_at = conversation time" instruction. With the field declared as "timestamp — use conversation time for timeless facts" the LLM no longer treats empty as a valid response.

## Tests

- The existing `TestFormatIso8601ToYmdHms` suite (added in #16483 as forward-looking) is updated: ordinal-date cases now pass, invalid-string case now asserts `""` instead of the garbage string, and a new empty-string case covers the explicit empty-input path.

## Out of scope

- The caller in `api/db/joint_services/memory_message_service.py:167` could now distinguish "model said empty" vs "model said garbage" with a single `is None` vs `== ""` check, but the prompt-side fix should drive the empty case to zero first. Leaving the caller as-is keeps the PR minimal.

### DIFF
--- a/common/time_utils.py
+++ b/common/time_utils.py
@@ -137,7 +137,8 @@ def format_iso_8601_to_ymd_hms(time_str: str) -> str:
         time_str: ISO 8601 date string (e.g. "2024-01-01T12:00:00Z")
 
     Returns:
-        str: Date string in "YYYY-MM-DD HH:MM:SS" format
+        str: Date string in "YYYY-MM-DD HH:MM:SS" format, or "" if
+        ``time_str`` is empty or cannot be parsed as ISO 8601.
 
     Example:
         >>> format_iso_8601_to_ymd_hms("2024-01-01T12:00:00Z")
@@ -145,9 +146,19 @@ def format_iso_8601_to_ymd_hms(time_str: str) -> str:
     """
     from dateutil import parser
 
+    if not time_str:
+        return ""
     try:
-        dt = parser.isoparse(time_str)
+        # ``dateutil.parser.isoparse`` accepts every ISO 8601 form the LLM
+        # extraction prompt can plausibly emit, including ordinal dates
+        # (``2024-001T12:00:00``) that stdlib ``datetime.fromisoformat``
+        # rejects. Strip the tz offset so ``strftime`` formats the wall
+        # clock unchanged, matching the legacy return path.
+        dt = parser.isoparse(time_str).replace(tzinfo=None)
         return dt.strftime("%Y-%m-%d %H:%M:%S")
     except Exception as e:
-        logging.error(str(e))
-        return time_str
+        # Surface the rejected value and exception so callers (and log readers)
+        # can identify which field of which memory failed to parse, instead
+        # of silently writing the unparsed string into a timestamp column.
+        logging.error(f"format_iso_8601_to_ymd_hms: invalid ISO 8601 timestamp {time_str!r}: {e}")
+        return ""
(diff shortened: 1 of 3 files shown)


JinHai-CN:
@Taranum01 It seems this issue is already fixed by #18462 and #18463. Would you please double check it?

PR #18510, opened 2026-08-19
Fix: Clearing the files of the dataset selected in the chat should result in an error message appearing on the chat page.

cike8899:
### Summary

Fix: Clearing the files of the dataset selected in the chat should result in an error message appearing on the chat page.


wangq8 (commented):
Thanks for the fix. The approach is sound: `useStaleDatasetIds` correctly detects datasets that were deleted or emptied (chunk_count <= 0), and surfacing the issue via `superRefine` on `dataset_ids` is the right place to produce a form error. A few things to consider:

1. **`datasetIds.length` in the effect deps (chat-settings.tsx)** — the guard uses `datasetIds?.length`, but the dependency array accesses `datasetIds.length` without optional chaining. If `dataset_ids` is ever `undefined` (e.g., after `form.reset(nextData)` for a chat that has no `dataset_ids`), this throws during render. Please use `datasetIds?.length` in the deps array for consistency with the guard.

2. **`mode: 'onChange'` affects the whole form** — this switches validation for every field (name, system prompt, etc.) to run on each change instead of on submit, which is a broader UX change than the bug being fixed. If the only goal is to clear the `dataset_ids` error once the user re-selects a valid dataset, please confirm this side effect is acceptable (or scope the re-validation to `dataset_ids`).

3. **Scope of the dropdown change** — filtering out `chunk_count <= 0` datasets in `knowledge-base-item.tsx` hides empty datasets across all three consumers of `KnowledgeBaseFormField` (chat settings, `next-search/search-setting.tsx`, and agent retrieval form), not just chat. Previously those datasets were shown but disabled. The change matches the existing `knowledgeBasesTip` copy ("An empty dataset will not appear in the dropdown list"), so it's a reasonable consistency fix — just flagging that it's wider than the PR title suggests.

4. **Missing tests** — no unit tests added for `useStaleDatasetIds` or the schema refinement, which now carries subtle loading/settled timing logic (empty set while loading, then re-trigger once `settled` flips). A small test would help lock this in.

Minor: the comment in `knowledge-base-item.tsx` says a stale selected value "is excluded as well", but `nextOptions` is built from the paginated `datasetList`, not from the selected ids — the stale-id handling actually lives in `useStaleDatasetIds` + the MultiSelect fallback. The comment reads a bit misleading.

PR #18512, opened 2026-08-19
fix(web): style think/retrieving blocks in floating chat widget

dcc123456:
### Summary

fix(web): style think/retrieving blocks in floating chat widget

The widget markdown wrapper referenced the class name as a plain string while the stylesheet is a CSS module, so the hashed class never matched and none of the rules applied. Bind the wrapper to the module class and add details.think / details.retrieving rules so reasoning blocks are visually distinct from the answer body, matching next-markdown-content.

### DIFF
--- a/web/src/components/floating-chat-widget-markdown.module.less
+++ b/web/src/components/floating-chat-widget-markdown.module.less
@@ -59,6 +59,32 @@
     line-height: 1.6;
   }
 
+  :global(details.think),
+  :global(details.retrieving) {
+    padding-inline-start: 0px;
+    margin-bottom: 10px;
+    font-size: 12px;
+    color: rgb(var(--text-secondary));
+
+    summary {
+      cursor: pointer;
+      font-weight: 500;
+      user-select: none;
+      color: var(--text-secondary);
+      &:hover {
+        color: rgb(var(--text-secondary));
+      }
+    }
+  }
+
+  :global(details.think) {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix — the root cause is correctly identified: the wrapper referenced the class name as a plain string while the stylesheet is a CSS module, so the hashed class never matched and none of the module rules applied. Binding it to `styles['floating-chat-widget']` is correct and consistent with the existing `styles.referenceIcon` usage, and the new `:global(details.think)` / `:global(details.retrieving)` rules match the pattern already used in `next-markdown-content/index.module.less`.

One follow-up worth addressing (same root cause, still in this file):

- `.widget-citation-content` is still referenced as a plain string (`className="flex gap-2 widget-citation-content"` at line ~190 of `floating-chat-widget-markdown.tsx`), and `.widget-citation-popover` is defined in the module but never bound to `styles[...]`. Both are CSS-module classes, so those rules (word-break/overflow-wrap, popover sizing) remain ineffective for the same reason this PR fixes. Consider binding those to `styles[...]` too, here or in a follow-up.

Minor style nit (non-blocking): the new rules use `padding-inline-start: 0px`, whereas `next-markdown-content` uses `10px` between the border and the text. With `0px` the text sits flush against the `border-inline-start`. Just confirming that's the intended look for the compact widget.

PR #18517, opened 2026-08-19
fix(web): show agent opening statement in embedded floating widget

euvre:
### Summary

An agent whose Begin component has the opening-statement switch enabled showed no opening statement (prologue) when embedded as the floating chat widget on a website, while the same agent showed it correctly on the fullscreen share page.

Root cause: the floating widget routes agent chats through `useSendNextSharedMessage` -> `useSendAgentMessage`, whose prologue seeding reads the agent canvas graph store (`useGetBeginNodePrologue`). That store is never populated on the widget page, so the prologue was silently dropped. The chat-assistant flow was unaffected because its prologue arrives from the completions handshake, and the agent share page was unaffected because it seeds the prologue from the `agentbots/<id>/inputs` response.

Fix: mirror `web/src/pages/agent/share/index.tsx` in `web/src/components/floating-chat-widget.tsx` — when the widget runs the agent flow (and not in task mode), seed the prologue returned by the agentbots inputs endpoint into the message list via the agent hook's `addNewestOneAnswer`. A stable message id keeps the seeding idempotent under React StrictMode double-invoked effects. The chat-assistant flow is untouched (the effect early-returns for non-agent embeds).

### Verification

- Reproduced pre-fix with a released conversational agent (`enablePrologue` on): the widget rendered the header but no opening message.
- Post-fix, the same widget URL renders the configured opening statement immediately.
- `npx oxlint` clean on the changed file; `npm run type-check` reports no errors for the changed file (remaining repo errors are pre-existing in unrelated files).

### DIFF
--- a/web/src/components/floating-chat-widget.tsx
+++ b/web/src/components/floating-chat-widget.tsx
@@ -20,7 +20,6 @@ import { useClickDrawer } from '@/components/pdf-drawer/hooks';
 import { MessageType, SharedFrom } from '@/constants/chat';
 import { useFetchExternalAgentInputs } from '@/hooks/use-agent-request';
 import { useFetchExternalChatInfo } from '@/hooks/use-chat-request';
-import { IAnswer } from '@/interfaces/database/chat';
 import i18n, { changeLanguageAsync } from '@/locales/config';
 import { useSendNextSharedMessage } from '@/pages/agent/hooks/use-send-shared-message';
 import { removeThinkSection } from '@/utils/chat';
@@ -100,6 +99,13 @@ const normalizeWidgetFooterLink = (value: string | null) => {
 };
 
 /**
+ * Stable id for the seeded agent prologue. `addNewestOneAnswer` dedupes by
+ * message id, so a re-invoked effect (e.g. under React StrictMode) merges
+ * into the seeded message instead of appending a duplicate.
+ */
+const PROLOGUE_MESSAGE_ID = 'prologue';
+
+/**
  * Renders the embeddable floating chat widget and applies URL-driven widget settings.
  */
 const FloatingChatWidget = () => {
@@ -166,11 +172,13 @@ const FloatingChatWidget = () => {
   } = hookResult;
   const findReferenceByMessageId = (hookResult as any).findReferenceByMessageId;
   // Only the agent flow exposes these. The chat flow receives its prologue
-  // through the completions handshake instead.
-  const { addNewestOneAnswer, isTaskMode } = hookResult as {
-    addNewestOneAnswer?: (answer: IAnswer) => void;
-    isTaskMode?: boolean;
-  };
+  // through the completions handshake instead, so both stay optional here.
+  const { addNewestOneAnswer, isTaskMode } = hookResult as Partial<
+    Pick<
+      ReturnType<typeof useSendNextSharedMessage>,
+      'addNewestOneAnswer' | 'isTaskMode'
+    >
+  >;
 
   // Sync our local input with the hook's value when needed
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
Reviewed the diff and verified against the current codebase. The fix correctly mirrors `web/src/pages/agent/share/index.tsx`.

**What I verified:**
- `useSendNextSharedMessage` returns `addNewestOneAnswer` (via `...ret` from `useSendAgentMessage`) and `isTaskMode`, so the cast in the widget is accurate for the agent flow.
- `useFetchExternalAgentInputs` returns `IInputs` which includes `prologue` and `mode`, so `data?.prologue` is correctly typed.
- `addNewestOneAnswer` dedupes by `buildMessageUuid({ id: answer.id, role: Assistant })`, so passing a stable `id: 'prologue'` makes the seeding idempotent under React StrictMode double-invoked effects — a nice improvement over the share page, which passes no id.
- The effect early-returns for non-agent embeds (`!isFromAgent`) and task mode (`isTaskMode`), so the chat-assistant flow is untouched.
- The widget renders `displayMessages` synced from `derivedMessages`, so the seeded prologue actually surfaces.

**Minor/non-blocking notes:**
- The `id: 'prologue'` hardcoded value is fine here since real message ids are generated UUIDs, but a named constant would make the intent slightly clearer.
- The `as { addNewestOneAnswer?, isTaskMode? }` cast is consistent with the existing `(hookResult as any).findReferenceByMessageId` pattern; a shared typed return type for the two hooks could clean this up in the future, but it's out of scope for this fix.

No regressions or security concerns found. Looks good — thanks for the clear summary and verification notes.

wangq8 (commented):
Thanks — this correctly mirrors `pages/agent/share/index.tsx` and fixes the prologue being dropped on the widget page.

Verified:
- `data` here is `useFetchExternalAgentInputs()` for the agent flow, so `data.prologue` is the same source the share page uses.
- The `!isFromAgent || isTaskMode` early-return correctly leaves the chat-assistant flow untouched.
- Using a stable `id: 'prologue'` is a nice improvement over the share page (which omits `id`): `addNewestOneAnswer` dedupes by `id`, so StrictMode's double-invoked effect merges instead of appending a duplicate. Confirmed the stored message id resolves back to `'prologue'` via `buildMessageUuid`, so the dedup actually matches.

Non-blocking nits:
- The hardcoded `'prologue'` id is fine in practice (other message ids are UUIDs), but `prologue-${conversationId}` would be slightly more defensive.
- The `as { addNewestOneAnswer?: ...; isTaskMode?: boolean }` cast is loose; both fields already exist on the `useSendNextSharedMessage` result, so this could be typed more precisely. Cosmetic only.

LGTM.

euvre:
Thanks @wangq8 for the approval and the thorough verification notes — both non-blocking nits are addressed in 62c2004 (no behavior change):

**1. Hardcoded `id: 'prologue'`** — Applied: extracted a named module-level constant `PROLOGUE_MESSAGE_ID` with a doc comment explaining the dedup intent. I kept the stable literal rather than `prologue-${conversationId}`: the id only needs to be stable within one mounted widget instance so `addNewestOneAnswer`'s id-based merge (`pre.findIndex((x) => x.id === answer.id)`) collapses StrictMode's double-invoked effect, and `shared_id` comes from the URL and is fixed for the widget's lifetime, so a per-conversation suffix would add no extra defensiveness.

**2. Loose cast** — Applied the local precise typing: the agent-only fields are now typed as `Partial<Pick<ReturnType<typeof useSendNextSharedMessage>, 'addNewestOneAnswer' | 'isTaskMode'>>`, sourced from the agent hook's actual return type (which, as you noted, already exposes both fields) while staying optional for the chat flow. The broader cleanup (a shared typed return type for both send hooks) remains out of scope for this fix, as you suggested.

**Verification:** `npm run type-check` reports zero errors in `floating-chat-widget.tsx` (the run does show 219 pre-existing errors in unrelated files such as `pages/skills/types.ts` and `components/jsonjoy-builder/*` — that baseline noise exists on main independent of this PR). No shared-service/browser end-to-end verification was run here; confidence is based on the type-check plus the code-level comparison with `pages/agent/share/index.tsx` you already confirmed.

dcc123456 (approved):

PR #15961, opened 2026-06-11
fix(api): use dataset-owner tenant for legacy /chunks docstore cleanup

kiannidev:
Fixes #15960

### What problem does this PR solve?

Legacy `POST /api/v1/datasets/{dataset_id}/chunks` (`parse`) and `DELETE .../chunks` (`stop_parsing`) build the docstore index from the **requester** `tenant_id` (`@add_tenant_id_to_kwargs`). On team-shared datasets, chunks live under the **dataset owner's** index (`ragflow_{kb.tenant_id}`).

After #15876 added `index_exist` guards, team members no longer hit HTTP 500, but cleanup is silently skipped — stale chunks remain when a non-owner re-parses or stops parsing.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [ ] New Feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe):

### Changes

- Resolve `index_name` via `_get_dataset_tenant_id(dataset_id)` in `parse` and `stop_parsing`.
- Pass `doc[0].kb_id` to `index_exist` / `delete` (aligned with other chunk handlers).
- Add unit tests asserting a team-member caller deletes from `search.index_name(owner_tenant)`, not the requester tenant.

### Test plan

- [x] `pytest test/testcases/test_http_api/test_file_management_within_dataset/test_doc_sdk_routes_unit.py::TestDocRoutesUnit::test_legacy_chunks_parse_uses_dataset_owner_tenant_for_delete -v`
- [x] `pytest test/testcases/test_http_api/test_file_management_within_dataset/test_doc_sdk_routes_unit.py::TestDocRoutesUnit::test_legacy_chunks_stop_uses_dataset_owner_tenant_for_delete -v`
- [ ] Manual: team member B re-parses document on team dataset owned by A — old chunks are removed from `ragflow_A`

PR #18518, opened 2026-08-19
fix(api): derive SQL retrieval tenant from dataset owner, not chat tenant

### LINKED ISSUE #18514: [Bug]: SQL retrieval uses the chat tenant instead of the shared dataset owner tenant
### Self Checks

v0.26.4 official Docker image

### RAGFlow image version

infiniflow/ragflow:v0.26.4

```Markdown
- Installation: Official Docker Compose deployment
- RAGFlow image: `infiniflow/ragflow:v0.26.4`
- Document engine: Elasticsearch 8.11.3
- Metadata database: MySQL 8.0.40
- Dataset parser: Table
- Dataset permission: Team
- Host OS: Linux
```

### Actual behavior

When a chat references a team-shared Table dataset owned by another tenant,
SQL retrieval derives the target Elasticsearch index from
`dialog.tenant_id` instead of the dataset owner's
`Knowledgebase.tenant_id`.

For example:

- User A owns dataset `KB_A`.
- Its chunks are stored in `ragflow_<A>`.
- User B joins User A's tenant and is authorized to access `KB_A`.
- User B creates a chat referencing `KB_A`.
- Because `KB_A` has a non-empty `field_map`, `async_chat()` enters the SQL
  retrieval path.
- The chat belongs to User B, so `dialog.tenant_id` is User B's ID.
- `async_chat()` passes `dialog.tenant_id` to `use_sql()`.
- `use_sql()` consequently queries `ragflow_<B>`, although the dataset is
  stored in `ragflow_<A>`.

The relevant call is:

```python
ans = await use_sql(
    questions[-1],
    field_map,
    dialog.tenant_id,
    chat_mdl,
    prompt_config.get("quote", True),
    dialog.kb_ids,
)
```
The vector retrieval path in the same function does not have this problem,
because it correctly derives tenant IDs from the loaded knowledge bases:
```
tenant_ids = list(set([kb.tenant_id for kb in kbs]))
```
In the local reproduction, the owner index existed and contained one row:
```
ragflow_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa  docs=1
```
Querying the owner index directly through Elasticsearch SQL succeeded:
```
{
  "columns": [
    {
      "name": "total",
      "type": "long"
    }
  ],
  "rows": [[1]]
}
```
However, use_sql() generated and executed SQL against the chat owner's
index:
```
SELECT COUNT(*) AS total
FROM ragflow_bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
WHERE kb_id = 'cccccccccccccccccccccccccccccccc'
```
Elasticsearch returned:
```
Unknown index [ragflow_bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb]
```
use_sql() then returned None, causing SQL retrieval to fail and normally
fall back to vector retrieval.

### Expected behavior

SQL retrieval should derive its document-store namespace from each referenced
dataset's `Knowledgebase.tenant_id`, not from `dialog.tenant_id`.

For a shared dataset owned by User A, both SQL retrieval and vector retrieval
should query User A's document-store namespace: ragflow_A

### Steps to reproduce

```Markdown
1. Deploy the official image `infiniflow/ragflow:v0.26.4` with Elasticsearch.
2. Register User A and User B.
3. User A creates a dataset using the Table parser.
4. Upload an Excel file and configure at least one column as `metadata` or
   `both`, so the dataset receives a non-empty `field_map`.
5. Set the dataset permission to `team`.
6. Add User B to User A's tenant.
7. User B creates a chat and selects User A's shared dataset.
8. Ask a structured question that triggers SQL retrieval, for example:
   `How many rows are in the table?`
9. Inspect the generated SQL or the backend logs.

The generated table/index is:ragflow_<B>
The dataset chunks are actually stored in:ragflow_<A>
Elasticsearch therefore reports an unknown index, and RAGFlow falls back to
vector retrieval.
```

### Additional information

This appears to have the same root cause as #15960 and its fix #15961.

Issue #15960 reported that shared-dataset chunk cleanup used the requesting
user's tenant to construct the document-store index, while the chunks were
stored under the dataset owner's tenant. PR #15961 fixed that path by resolving
the index from the dataset owner.

The SQL retrieval path still has the same ownership mismatch:
document-store namespace = dataset owner tenant
current SQL namespace     = chat owner tenant

ChendongPang:
Fixes #18514

`dialog_service.async_chat`'s SQL retrieval path builds the doc-store index/table name from `dialog.tenant_id` (`use_sql()`'s `base_table = index_name(tenant_id)`), i.e. the tenant of whoever **created the chat**. On team-shared table datasets, the dataset can be owned by a **different** tenant than the chat creator: chunks live under `ragflow_{kb.tenant_id}`, but SQL retrieval was querying `ragflow_{dialog.tenant_id}`. The query returns nothing (or an "Unknown index" error on Elasticsearch), and `async_chat` silently falls back to vector retrieval, so the table-specific SQL answer is lost.

The vector retrieval path in the same function already derives tenant ids per-knowledgebase (`tenant_ids = list(set([kb.tenant_id for kb in kbs]))`), and the newer `rag/advanced_rag` SQL call sites (`agentic_rag.py`, `harness/tools/search.py`) already derive `tenant_id = sql_kbs[0].tenant_id` correctly. This mirrors #15961's fix for the same class of tenant-mismatch bug in the legacy `/chunks` docstore cleanup path — same root cause, different call site.

### Type of change

- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement

### Changes

- Derive the SQL-retrieval tenant/kb scope from the referenced `Knowledgebase` objects (`kbs`, already fetched via `get_models()`) instead of `dialog.tenant_id` / `dialog.kb_ids`, restricted to the datasets that actually carry a `field_map` (table datasets).
- Pass the resolved `sql_tenant_id` / `sql_kb_ids` into `use_sql(...)` so the doc-store index name is built from the dataset owner's tenant, not the chat creator's.
- Update the `_enrich_chunks_with_document_metadata` skip-check to key off `sql_kb_ids` instead of `dialog.kb_ids`, since that's what actually determines whether `use_sql` includes a `kb_id` column in its results.

### Test plan

- [x] `uv run --group test pytest test/unit_test/api/db/services/test_dialog_service_use_sql_source_columns.py -v`
- [x] `uv run --group test pytest test/unit_test/api/db/services/test_dialog_service_final_answer.py -v`
- Added `test_async_chat_sql_retrieval_uses_dataset_owner_tenant_not_dialog_tenant`, which drives `dialog_service.async_chat` end-to-end with a table dataset owned by a different tenant than the dialog, and asserts the tenant id passed into `use_sql` is the dataset owner's tenant, not the dialog's.

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -672,7 +672,7 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
         # Derive the doc-store tenant/namespace from the referenced dataset's own
         # owner, not from dialog.tenant_id: a team-shared dataset may be owned by a
         # different tenant than the one who created this chat.
-        sql_kbs = [kb for kb in kbs if kb.parser_config and "field_map" in kb.parser_config]
+        sql_kbs = [kb for kb in kbs if kb.parser_config and kb.parser_config.get("field_map")]
         sql_tenant_ids = {kb.tenant_id for kb in sql_kbs}
         if len(sql_tenant_ids) > 1:
             # use_sql queries a single tenant's doc-store index per call, and
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -669,15 +669,21 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
     logging.debug(f"field_map retrieved: {field_map}")
     # try to use sql if field mapping is good to go
     if field_map:
+        # Derive the doc-store tenant/namespace from the referenced dataset's own
+        # owner, not from dialog.tenant_id: a team-shared dataset may be owned by a
+        # different tenant than the one who created this chat.
+        sql_kbs = [kb for kb in kbs if kb.parser_config and "field_map" in kb.parser_config]
+        sql_tenant_id = sql_kbs[0].tenant_id if sql_kbs else dialog.tenant_id
+        sql_kb_ids = [kb.id for kb in sql_kbs] if sql_kbs else dialog.kb_ids
         logging.debug("Use SQL to retrieval:{}".format(questions[-1]))
-        ans = await use_sql(questions[-1], field_map, dialog.tenant_id, chat_mdl, prompt_config.get("quote", True), dialog.kb_ids, doc_ids=scoped_doc_ids)
+        ans = await use_sql(questions[-1], field_map, sql_tenant_id, chat_mdl, prompt_config.get("quote", True), sql_kb_ids, doc_ids=scoped_doc_ids)
         # For aggregate queries (COUNT, SUM, etc.), chunks may be empty but answer is still valid
         if ans and (ans.get("reference", {}).get("chunks") or ans.get("answer")):
             if include_reference_metadata and ans.get("reference", {}).get("chunks"):
-                if len(dialog.kb_ids) != 1 and any(not c.get("kb_id") for c in ans["reference"]["chunks"]):
+                if len(sql_kb_ids) != 1 and any(not c.get("kb_id") for c in ans["reference"]["chunks"]):
                     logging.warning(
-                        "Skipping some _enrich_chunks_with_document_metadata results because dialog.kb_ids has %d entries and use_sql returned chunks without kb_id.",
-                        len(dialog.kb_ids),
+                        "Skipping some _enrich_chunks_with_document_metadata results because sql_kb_ids has %d entries and use_sql returned chunks without kb_id.",
+                        len(sql_kb_ids),
                     )
                 _enrich_chunks_with_document_metadata(ans["reference"]["chunks"], metadata_fields)
             yield ans
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -673,23 +673,34 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
         # owner, not from dialog.tenant_id: a team-shared dataset may be owned by a
         # different tenant than the one who created this chat.
         sql_kbs = [kb for kb in kbs if kb.parser_config and "field_map" in kb.parser_config]
-        sql_tenant_id = sql_kbs[0].tenant_id if sql_kbs else dialog.tenant_id
-        sql_kb_ids = [kb.id for kb in sql_kbs] if sql_kbs else dialog.kb_ids
-        logging.debug("Use SQL to retrieval:{}".format(questions[-1]))
-        ans = await use_sql(questions[-1], field_map, sql_tenant_id, chat_mdl, prompt_config.get("quote", True), sql_kb_ids, doc_ids=scoped_doc_ids)
-        # For aggregate queries (COUNT, SUM, etc.), chunks may be empty but answer is still valid
-        if ans and (ans.get("reference", {}).get("chunks") or ans.get("answer")):
-            if include_reference_metadata and ans.get("reference", {}).get("chunks"):
-                if len(sql_kb_ids) != 1 and any(not c.get("kb_id") for c in ans["reference"]["chunks"]):
-                    logging.warning(
-                        "Skipping some _enrich_chunks_with_document_metadata results because sql_kb_ids has %d entries and use_sql returned chunks without kb_id.",
-                        len(sql_kb_ids),
-                    )
-                _enrich_chunks_with_document_metadata(ans["reference"]["chunks"], metadata_fields)
-            yield ans
-            return
+        sql_tenant_ids = {kb.tenant_id for kb in sql_kbs}
+        if len(sql_tenant_ids) > 1:
+            # use_sql queries a single tenant's doc-store index per call, and
+            # re-running it once per tenant is too slow to do inline (each call
+            # round-trips an LLM to generate SQL). Skip SQL retrieval rather than
+            # silently querying only one tenant's index and dropping the rest.
+            logging.warning(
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for the fix — this mirrors the same tenant-mismatch root cause as #15961 and the newer `agentic_rag.py` / `harness/tools/search.py` call sites, so aligning `dialog_service.async_chat` is the right call.

**Correctness**
- `use_sql()` builds `base_table = index_name(tenant_id)`, so passing `dialog.tenant_id` (chat creator) while chunks live under `ragflow_{kb.tenant_id}` was indeed the bug. Deriving `sql_tenant_id` from `sql_kbs[0].tenant_id` is correct, and `kbs` from `get_models()` are full `Knowledgebase` models (via `get_by_ids`), so `kb.parser_config` / `kb.tenant_id` are always populated. ✅
- The `sql_kbs` filter (`kb.parser_config and "field_map" in kb.parser_config`) is consistent with `KnowledgebaseService.get_field_map`, so `field_map` and `sql_kb_ids` stay in lockstep. ✅
- Switching the `_enrich_chunks_with_document_metadata` skip-check to key off `sql_kb_ids` (rather than `dialog.kb_ids`) is the correct tightening — that's what actually controls whether `use_sql` emits a `kb_id` column. ✅

**Multi-tenant behavior**
- Skipping SQL retrieval when field-map datasets span >1 tenant and falling back to vector search is a deliberate, documented change. The vector path already derives `tenant_ids = list(set([kb.tenant_id for kb in kbs]))`, so it correctly queries both tenants' indices. The old behavior (querying only `dialog.tenant_id`'s index and silently dropping/erroring) was strictly worse, so this is a net improvement. ✅

**Tests**
- Both new tests are well-scoped regressions: one asserts the dataset-owner tenant is passed to `use_sql`, the other asserts mixed-tenant field-map datasets skip SQL and fall through to the retriever. Good coverage of the two branches. ✅

**Minor / non-blocking**
- Edge case: a dataset whose `parser_config` contains `"field_map": {}` (empty dict) would still be included in `sql_kbs` but contributes nothing to `field_map`. This is harmless and mirrors existing behavior, just noting it for completeness.
- The `else`-branch fallback (`sql_kbs` empty → `dialog.tenant_id` / `dialog.kb_ids`) is good defensive coding, though in practice `sql_kbs` should be non-empty whenever `field_map` is truthy.

Nothing blocking — LGTM. 👍

ChendongPang:
@wangq8 Thanks for pointing this out. I tightened sql_kbs to include only datasets with a non-empty field_map.

Although an empty map contributes no fields, including that dataset could still widen the SQL KB scope or incorrectly trigger the mixed-tenant fallback. I also added a regression test covering an empty-field-map dataset owned by another tenant.

Addressed in d463e9c5f. I kept the sql_kbs-empty fallback unchanged as a defensive path.

ChendongPang:
Hi @wangq8, thanks again for the review and LGTM. All feedback has been addressed, and the PR is currently clean and mergeable. Are there any remaining concerns, or can it move to the next stage?

wangq8 (commented):
LGTM. Thanks for the clear root-cause write-up — this is the same tenant-mismatch class fixed in #15961, applied to the SQL retrieval path.

Verified against the current `async_chat`/`use_sql` code:

- `get_models()` already returns the `Knowledgebase` objects (`kbs`) with `tenant_id`, `id`, and `parser_config`, so deriving `sql_tenant_id`/`sql_kb_ids` from `kb.tenant_id`/`kb.id` is correct and matches how the vector path (`tenant_ids = list(set([kb.tenant_id for kb in kbs]))`) and the `agentic_rag`/`harness` SQL call sites already do it.
- `use_sql()` builds `base_table = index_name(tenant_id)` and, for Infinity, appends `_kb_ids[0]` only when `len(kb_ids) == 1`. Scoping down to `sql_kb_ids` (only field-map datasets) is correct here — non-field-map datasets are not SQL-queryable and previously leaked into the WHERE/kb_id scope.
- The `_enrich_chunks_with_document_metadata` skip-check now keys off `len(sql_kb_ids)` instead of `len(dialog.kb_ids)`, which accurately reflects whether `use_sql` emits a `kb_id` column. Good.

Two minor, non-blocking notes for future consideration:

1. **Multi-tenant field-map fallback** — when field-map datasets span >1 tenant, SQL retrieval is skipped and the flow silently falls back to vector search. That's a reasonable correctness-first tradeoff given `use_sql` queries a single tenant index per call, and the `logging.warning` makes it observable. Just flagging that table-specific answers for genuinely cross-tenant chats won't be produced here (worth a follow-up if that scenario matters).

2. **`sql_kbs` vs `get_field_map` guard** — `field_map` is built from `get_field_map(dialog.kb_ids)` (merges `parser_config["field_map"]` across all kbs), while `sql_kbs` filters on `kb.parser_config.get("field_map")` being truthy. These are effectively consistent today, but the defensive `sql_kbs[0].tenant_id if sql_kbs else dialog.tenant_id` fallback is good to keep; no action needed.

Tests are solid and cover the three key cases (owner-tenant, empty-field-map filtering, mixed-tenant skip). Nice work.

wangq8:
@ChendongPang please help to fix the unit test failure.

ChendongPang:
@wangq8 All passed.

wangq8:
LGTM.

PR #18519, opened 2026-08-19
fix(search): fix layout clipping when search page is not fullscreen

euvre:
### Summary

When the search page (`/search/:id`) is opened in a non-fullscreen window, the layout breaks visually: a horizontal scrollbar appears, the settings gear is pushed outside the viewport, and focusing the search input auto-scrolls the container so the panel border and logo are cut through at the viewport edge (the "穿模"/clipping reported in the issue).

Root cause: #18050 added `overflow-x-auto` on the page section and a `min-w-[1280px]` floor on the content panel to protect against squeezed layouts. However, the settings gear button sits *after* that min-width panel in the same flex row, so at viewport widths below ~1382px the row always overflows: the panel is pinned at 1280px, the gear lands beyond the right edge (measured at x=1320-1362 in a 1100px viewport), and the scroll container gets auto-scrolled when the textarea receives focus.

This PR removes the hard floor and fixes the real squeeze source instead:

- `web/src/pages/next-search/index.tsx`
  - Drop `overflow-x-auto` from the section and `min-w-[1280px]` from the panel; the panel is now `flex-1 min-w-0` so it fills the remaining space and can shrink.
  - The settings gear button gets `shrink-0` so it always stays visible next to the panel.
  - The settings panel gets `shrink-0 max-w-full` so it never exceeds the bordered panel, and the main area gets `overflow-hidden` so its content can never bleed across the settings divider at very narrow widths.
- `web/src/pages/next-search/search-home.tsx`
  - The home card container changes from a fixed `w-[780px]` to `w-full max-w-[780px]`, so it shrinks gracefully on narrow windows instead of overflowing the panel border (the layout squeeze that #18050's min-width was papering over). This also benefits the embedded/share page, which renders this component at arbitrary widths.

Verification (DOM-measured in the browser against a real search app with retrieval + AI summary):

| width | state | before | after |
|---|---|---|---|
| 1100 | home | h-scroll, gear off-screen (x=1320) | no h-scroll, gear visible |
| 1100 | searching | h-scroll, auto-scroll clipped logo, mind-map button off-screen | all elements in view, results render normally |
| 1100 | settings open | gear pushed out of viewport (x=1100-1142) | no overflow, settings inside panel border |
| 700 | settings open | n/a (horizontal scroll) | no bleed across the settings divider (hit-test verified) |
| 1920 | home / settings open | OK | OK — geometry unchanged, no regression |

No unit tests cover these layout classes; `npm run type-check` reports no new errors for the touched files, and oxlint passes on both. The Playwright e2e selectors (`[data-testid='search-detail']`) are unchanged.

### DIFF
--- a/web/src/pages/next-search/index.tsx
+++ b/web/src/pages/next-search/index.tsx
@@ -50,11 +50,11 @@ export default function SearchPage() {
 
   return (
     <section
-      className="size-full flex-1 relative px-5 pb-5 flex pt-4 overflow-x-auto"
+      className="size-full flex-1 relative px-5 pb-5 flex pt-4"
       data-testid="search-detail"
     >
-      <div className="flex gap-3 w-full bg-bg-base border-0.5 border-border-button min-w-[1280px] ">
-        <div className="flex-1 min-w-0">
+      <div className="flex gap-3 flex-1 min-w-0 bg-bg-base border-0.5 border-border-button">
+        <div className="flex-1 min-w-0 overflow-hidden">
           {!isSearching && (
             <div className="animate-fade-in-down">
               <SearchHome
@@ -82,6 +82,7 @@ export default function SearchPage() {
           <SearchSetting
             open={openSetting}
             setOpen={setOpenSetting}
+            className="shrink-0 max-w-full"
             data={SearchData as ISearchAppDetailProps}
           />
         )}
@@ -89,7 +90,7 @@ export default function SearchPage() {
 
       <Button
         variant="transparent"
-        className="bg-bg-card ml-5"
+        className="bg-bg-card ml-5 shrink-0"
         onClick={() => setOpenSetting(!openSetting)}
       >
         <Settings className="text-text-secondary" />
--- a/web/src/pages/next-search/index.tsx
+++ b/web/src/pages/next-search/index.tsx
@@ -56,7 +56,7 @@ export default function SearchPage() {
       <div className="flex gap-3 flex-1 min-w-0 bg-bg-base border-0.5 border-border-button">
         <div className="flex-1 min-w-0 overflow-hidden">
           {!isSearching && (
-            <div className="animate-fade-in-down">
+            <div className="animate-fade-in-down h-full overflow-x-hidden overflow-y-auto">
               <SearchHome
                 setIsSearching={setIsSearching}
                 isSearching={isSearching}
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Layout-only fix with solid reasoning and DOM-measured verification. Replacing the `min-w-[1280px]` floor + `overflow-x-auto` with `flex-1 min-w-0` plus `shrink-0` on the gear/settings, and switching the home card to `w-full max-w-[780px]`, directly addresses the squeeze source instead of papering over it. One thing to watch: the new `overflow-hidden` on the main content area will clip any non-portaled overflow. Shadcn/Radix popovers/dropdowns portal to body so they should be fine, but please double-check any absolutely-positioned content rendered inline inside the search panel (e.g. inline dropdowns) at narrow widths to avoid reintroducing clipping. Otherwise LGTM.

euvre:
Thanks for the reviews! Both points addressed in 59ba279:

**1. Vertical scroll owner for SearchHome (CodeRabbit, minor)** — Valid and applied. Verified the chain: `html/body/#root` and root `<main>` both hide overflow, and pre-PR the outer section's `overflow-x-auto` implicitly made `overflow-y` compute to `auto`, so a vertical scroll path existed; this PR removed it and added `overflow-hidden`, which could clip SearchHome (`mt-[15vh]` + `min-h-[240px]` card) in short viewports. Fix: the home-branch wrapper now owns vertical scrolling (`h-full overflow-x-hidden overflow-y-auto`), scoped to the home branch only — the shared parent keeps `overflow-hidden` for horizontal clipping and SearchingView's own result scrolling is untouched. The `animate-fade-in-down` animation is transform-based, so it doesn't create scrollable overflow in the new container.

**2. Inline absolutely-positioned content check (@wangq8)** — Audited everything rendered inside the clipped panel: all popover/dropdown content (`Popover`, `Select`/`RAGFlowSelect`, `MultiSelect`, `TopSelect`, `ImageWithPopover`) goes through Radix `Portal` to `document.body` (no `container` override), so it's unaffected by the `overflow-hidden` ancestor. The only inline absolute elements (send button `absolute right-3`, input icons `absolute right-3`, result-card gradient `absolute bottom-0 left-0 right-0`, expand button `absolute bottom-2 left-1/2`, popover trigger `absolute top-16 right-10`, settings footer `absolute bottom-1 right-3 w-[calc(100%-1em)]`) are all inset relative to their own in-flow `relative` parents and paint within the panel bounds — no clipping at narrow widths. The settings panel itself is a sibling outside the `overflow-hidden` child.

Verification boundary: `npm run type-check` shows 0 errors in the touched files (remaining errors are pre-existing elsewhere on main), oxlint clean. No browser/E2E re-run in this environment — behavioral reasoning above is based on code-level evidence (CSS overflow spec + DOM structure analysis).

dcc123456 (approved):

PR #18522, opened 2026-08-19
Feat: Validate the knowledge base that was deleted from the search page.

cike8899:
### Summary

Feat: Validate the knowledge base that was deleted from the search page.


wangq8 (commented):
Thanks for this — the refactor consolidates the stale-dataset validation (previously chat-settings-only) into a shared hook and extends it to the search page, the agent retrieval form, and the retrieval node. The approach is sound: `superRefine` + an explicit `trigger()` on settle correctly handles the "persisted value never fires onChange" case, and the nested field path (`search_config.kb_ids`) is handled cleanly. A few comments:

1. **Scope creep in `web/.oxlintrc.json`** — the new global `no-unused-vars` rule is unrelated to this feature and changes lint behavior for the whole `web/` package. Consider splitting it into its own PR (or dropping it) to keep this diff focused.

2. **`useRevalidateStaleDatasetIds` keys off `datasetIds?.length`** rather than the full value. That's fine for the "revalidate once when the lookup settles" intent, but it means a change to the selected ids with the *same length* won't re-trigger the effect. Worth a one-line comment in the hook noting this is deliberate, so a future maintainer doesn't "fix" it and silently reintroduce the revalidation gap.

3. **Typing nit**: `schema.superRefine((data: any, ctx) => …)` drops the schema's output type; `T['_output']` would preserve it. Not blocking.

4. **Tests**: no unit coverage for the new hook. At minimum, a test that `superRefine` flags stale ids and that the revalidate effect calls `trigger` on settle would guard the exact regression this feature targets.

5. Verified the message key: `t('chat.datasetUnavailable')` resolves correctly (the key exists under `chat` in the en/zh locales, and the global-`t` dotted pattern is already used elsewhere).

No blocking issues — happy to approve once the `.oxlintrc.json` change is split out (or justified).

PR #18523, opened 2026-08-19
refactor(chunker): replace QAChunker/PresentationChunker with PairChunker/PageChunker, drop TagChunker

xugangqiang:
## Summary

This is a Go-only chunker naming/generalization refactor on the ingestion canvas.

1. **Delete `TagChunker`** — it was an orphaned component: no built-in ingestion template references it, the web canvas never exposed it (only `TokenChunker`/`TitleChunker` are in the operator enum), and the Python DSL chain (`rag/flow/chunker/`) has no counterpart. Its `(content, tags)` output capability is intentionally dropped with it (per decision).
2. **Generalize `QAChunker` → `PairChunker`** — one chunk per two-column record. A new `mode` parameter selects the output schema; today only `mode="qa"` (question-answer pairs, the default) is implemented, so QA behaviour is unchanged. The extraction paths (text/csv/markdown/html/json) are shared and untouched.
3. **Rename `PresentationChunker` → `PageChunker`** — "one chunk per slide/page" is the actual contract (`rag/app/presentation.py`: *"Every page will be treated as a chunk"*), and `PageChunker` describes the chunk unit rather than the source format.

## What changed

- `internal/ingestion/component/chunker/`: `qa.go → pair.go` (mode param, QA path unchanged), `presentation.go → page.go`, `tag.go`/`tag_test.go` deleted, factory updated, tests renamed/extended (`pair_test.go`, `pair_batch2_test.go`, `page_test.go`, `register_test.go`).
- `internal/agent/dsl/normalize.go`: name→kind map — drop `TagChunker`, `PresentationChunker → PageChunker`, and **add `PairChunker`** (the `QAChunker` entry was previously missing from the map, now filled by its successor).
- Built-in templates `ingestion_pipeline_qa.json` / `ingestion_pipeline_presentation.json`: component names, node ids, and edge ids updated.
- Golden/regression tests updated: `pipeline_executor_defaults_test.go`, `template_integration_test.go` (removed the dead `TagChunker` skip).

## Behavior change / notes

- **Breaking for hand-authored DSL**: stored canvases referencing `QAChunker` / `PresentationChunker` / `TagChunker` by name must be updated to `PairChunker` / `PageChunker` (or drop the node). These components were never exposed in the web canvas, and the built-in templates are migrated, so the practical impact is limited to imported/custom DSL.
- The `mode` param of `PairChunker` currently accepts only `"qa"`; any other value fails fast at construction (`PairChunker: unsupported mode %q`). Future output schemas (e.g. `"tag"`) can be added without touching the extraction layer.

## Test plan

- `bash build.sh --test ./internal/ingestion/component/chunker/ ./internal/ingestion/pipeline/ ./internal/ingestion/task/ ./internal/agent/dsl/` → all green.
- New tests: `TestPairChunker_ModeDefaultsToQA`, `TestPairChunker_ModeRejectsUnknown`; existing QA/Presentation coverage retained under the new names.


yuzhichang:
Delete TagChunker — agree.
Generalize QAChunker → PairChunker — I prefer keep it unchanged.
Rename PresentationChunker → PageChunker — agree.

PR #18526, opened 2026-08-19
fix(bot): align chatbot completion final frame and history persistence with python

euvre:
### Summary

In the embeddable floating chat widget (`/chats/widget`), citations of content from the same file behaved inconsistently: an earlier assistant message showed its citation normally, while a later one rendered a citation icon whose tooltip said "Reference unavailable" (and the broken icon looked off).

Root cause — the Go chatbot completion path (`internal/service/bot_completion.go`) diverged from python `async_iframe_completion` / `structure_answer` in two ways:

1. **Decorated answer persisted as history.** The final pipeline result is the decorated answer with server-inserted `[ID:n]` citation markers. Go persisted that decorated text as the assistant turn, while python persists the raw streamed text (its final frame carries `answer: ""`, so `structure_answer` keeps the accumulated deltas). The fabricated `[ID:0][ID:1]` markers therefore entered the next turn's prompt history (verified in `api_4_conversation.message`). Models that imitate the history format then emit `[ID:n]` markers even on turns whose retrieval returned no reference chunks, and the widget resolves those markers against an empty `reference.chunks` — rendering the info icon with "Reference unavailable" instead of a hover card with the chunk content.
2. **Final SSE frame carried the full decorated answer.** Python's wire contract sends `answer: ""` on the final frame (the text was already streamed as deltas; consumers such as `useSendMessageWithSse` skip the final text whenever deltas exist). Sending the full text again duplicates the answer for accumulating consumers and ships marker-decorated text that never matched the streamed deltas.

Fix — extract the frame-translation loop into `BotService.streamChatbotTurn` and align it with python:

- The final frame's `answer` is empty whenever text was streamed as deltas; the full text is kept only when nothing was streamed (single-shot results such as the structured-SQL path, and error frames), so no text is ever lost.
- The persisted assistant turn is the raw streamed text, including the `<think>`/`</think>` tags the marker frames stand for (mirroring python's accumulation), so server-fabricated citation markers no longer leak into subsequent prompts.

### Verification

- `bash build.sh --test ./internal/service/` — full package passes, including four new regression tests for `streamChatbotTurn` (raw-text persistence, empty final-frame answer, single-shot/error fallbacks).
- End-to-end in the widget against the Go API (9384): with a deterministic mock model emitting `[ID:0]`, both turns' citations resolve (reference chunks 2 and 4) and the hover shows the cited chunk; the final SSE frame now carries `final:true, answer:"", reference:{chunks:[...]}`; `api_4_conversation.message` history contains no server-inserted markers (a real-model Kimi chatbot session verified 0 markers after the fix, versus `[ID:0][ID:1]` before).
- Note: the widget frontend files (`web/src/components/floating-chat-widget*`) carry small pre-existing working-tree tweaks (citation icon/loading tooltip) that were present before this fix and are unrelated to the backend change; the primary fix is entirely in `internal/service/bot_completion.go`.


wangq8 (approved):
Thanks — solid root-cause analysis and well-covered by the four regression tests. The core fix is correct: `streamChatbotTurn` now mirrors python's wire contract (empty final-frame `answer` whenever deltas were streamed) and persists the raw streamed text instead of the `[ID:n]`-decorated final answer, so fabricated markers no longer leak into the next turn's history. Verified against `internal/service/bot_completion.go`; the `question`/`messageID` plumbing and the single-shot/error fallbacks are handled correctly.

A few non-blocking observations:

1. **Error frame after streamed deltas.** `finalData` is gated only on `rawAnswer == ""`. If a pipeline-level `**ERROR**` final ever arrives after deltas were already streamed, the error text would be dropped from the wire (client sees only the partial deltas). The old code always sent `fullAnswer` on the final frame, so this is a subtle behavior change. Suggest `if rawAnswer == "" || errored { finalData = fullAnswer }` so errors always reach the client. Rare in practice, cheap to harden.

2. **Bundled frontend tweaks.** The description calls the `Info` → `InfoCircleOutlined` and `loading`-based "Searching..." tooltip changes "pre-existing working-tree tweaks ... unrelated to the backend change". They're small and thematically related to the citation-icon symptom (the `loading` state is actually a nice complement to this fix), but if truly unrelated, consider splitting them so the backend change can land independently.

3. **Minor:** the error-log `dialog_id` now reads `session.DialogID` rather than the caller's `dialogID`. That's populated on the `GetBySessionID` read path and is logging-only, so no issue — just flagging the reliance in case `session` is ever constructed elsewhere.

Approved — the fix is correct and well-tested; the suggestions are optional hardening.

euvre:
Thanks @wangq8 for the review and approval, and @coderabbitai for the inline finding. Addressing each thread below.

**1. Final frame drops error text after streamed deltas (coderabbitai inline + wangq8 observation 1 — same issue)**

Agreed and fixed in `f19e523be`. The Go pipeline reports a mid-stream driver failure as a final `**ERROR**` result after partial deltas were already sent (`chat_pipeline.go`), and gating the final frame's data on `rawAnswer == ""` alone dropped that error text, leaving the client with a silently truncated answer. The condition is now `rawAnswer == "" || errored`, so the error always reaches the wire — matching the pre-change behavior and the python contract (`chat_model.py` appends the error to the stream; `async_iframe_completion` emits a code:500 frame). Error turns remain unpersisted, as before. Added regression test `TestStreamChatbotTurn_ErrorAfterDeltasKeptOnWire` covering exactly the delta-then-error path (final frame carries the error text, session untouched).

**2. Bundled frontend tweaks (wangq8 observation 2)**

Kept in this PR deliberately: the tooltip change is the front-end half of the same symptom in the title — while streaming, the final message has no reference yet, and the citation icon prematurely showed "Reference unavailable". The loading-aware tooltip ("Searching..." while loading) is what makes the backend wire fix user-visible, so the two halves belong together. Happy to split if you'd still prefer that.

**3. `session.DialogID` in log fields (wangq8 observation 3)**

Acknowledged, no change needed: on this path `session` always comes from `GetBySessionID`, so `DialogID` is populated.

**Verification boundary:** no end-to-end run against shared services was performed for this revision; the basis is the focused unit tests (`TestStreamChatbotTurn_*`, full `internal/service` package green via `bash build.sh --test ./internal/service/`) plus comparison against the python counterpart paths cited above.

wangq8 (commented):
Solid, well-reasoned fix. The root cause (persisting the decorated final answer with server-inserted `[ID:n]` markers, and echoing the full decorated text on the final SSE frame) is correctly identified, and the refactor cleanly separates raw streamed text from the decorated answer.

Positive:
- Final-frame `answer` is now empty when deltas were streamed, matching python's `final["answer"] = ""`; the full text is retained only for single-shot/error finals, so no text is lost.
- Persisting `rawAnswer` (accumulated deltas + `<think>`/`</think>` tags) instead of the decorated answer correctly stops citation markers from leaking into the next turn's prompt.
- Good regression coverage: the five new tests pin raw persistence, empty final, think-tag persistence, single-shot fallback, and error/error-after-deltas paths.

Questions / non-blocking suggestions:
1. Scope: the frontend changes (`floating-chat-widget-markdown.tsx/.less`, and the `loading` prop in `floating-chat-widget.tsx`) are bundled here but described as "pre-existing unrelated tweaks". They're small and reviewable, but for a clean history consider splitting them into their own PR. Note the `.referenceIcon` class was previously referenced by `styles.referenceIcon` but undefined in the `.less`, so adding it is actually a real (if tiny) fix — worth stating that in the description rather than calling it unrelated.
2. Single-shot fallback: when `rawAnswer == ""`, `persisted` falls back to `fullAnswer` (the decorated answer). `ChatbotCompletion` always calls `AsyncChat(..., streaming=true)`, so this should only be the structured-SQL/single-shot path — but if that path can ever carry marker-decorated text (e.g. `quote=true`), the markers would still leak into history there. Could you confirm the single-shot path never produces decorated text, or strip markers from `persisted` in that branch too?
3. `res.Reasoning` is never accumulated into `rawAnswer` (nor was it in `fullAnswer` before). This is pre-existing and fine for the chatbot path (reasoning arrives as `Answer` deltas with think markers), but worth a comment in `streamChatbotTurn` noting that `Reasoning` is intentionally ignored, to prevent future confusion.

No blocking issues; this looks correct and well-tested.

dcc123456 (approved):

euvre:
Thanks @wangq8 for the follow-up review, @coderabbitai for the wording finding, and @dcc123456 for the approval. Catching up on the remaining threads below (the earlier error-frame item was already addressed in f19e523 and replied to).\n\n**1. Frontend scope / \.referenceIcon was a real fix (wangq8)**\n\nYou're right — verified: \(styles.referenceIcon\) was referenced by \(floating-chat-widget-markdown.tsx\) but the class was undefined in the \.module.less\) before this PR; this PR adds it \(cursor: help; vertical-align\). Since the PR is already merged, rewording the description or splitting the frontend bits is no longer possible, so recording the clarification here: the \.referenceIcon\) rule is a genuine (tiny) fix, while the \(InfoCircleOutlined\)/loading-tooltip changes were the pre-existing widget tweaks described in the PR body.\n\n**2. Single-shot fallback can never persist marker-decorated text (wangq8)**\n\nConfirmed marker-free by code path, so no stripping was added:\n\n- \(ChatbotCompletion\) always calls \(AsyncChat(..., stream=true)\) \(bot_completion.go\), so the only non-streaming \(decorateAnswer\) call \(chat_pipeline.go\) is unreachable from the chatbot path.\n- The reachable finals with \(rawAnswer == ""\) are: the structured-SQL early return, which yields its final directly before decoration runs \(its markdown table carries \(##N69225\) source markers resolved client-side, not \[ID:n\]\) markers\), and \(**ERROR**\) finals, which are never persisted. The \(empty_response\) fallback streams its text as a non-final delta before its final, so \(rawAnswer\) is non-empty there too.\n- The streaming completion's \(decorateAnswer\) decorates \(extractVisibleAnswer(thinkState.fullText)\) — text accumulated from the very deltas already streamed — so a non-empty decorated final implies a non-empty \(rawAnswer\), and \(persisted\) stays the raw text.\n\nThis reasoning is captured in the comment above the fallback in \(streamChatbotTurn\).\n\n**3. \(res.Reasoning\) handling (wangq8)**\n\nAddressed in 25537a066 — rather than only documenting that \(Reasoning\) is ignored, \(streamChatbotTurn\) now forwards \(res.Reasoning\) as stream text whenever \(res.Answer\) is empty, so the tool path's in-think text reaches the widget's think block and the persisted history; covered by \(TestStreamChatbotTurn_ReasoningFieldForwardedAsStreamText\).\n\n**4. Compatibility wording in the new comments (coderabbitai)**\n\nApplied in ab3dc8d2e: the two flagged comments \(streamChatbotTurn\) reasoning comment and the matching test comment\) now describe only the local \(AsyncChatResult.Reasoning\) stream, forwarding when \(Answer\) is empty, and the resulting wire/persistence behavior, without the compatibility/cross-implementation wording. Comment-only change; no behavior touched.\n\n**Verification:** \(bash build.sh --test ./internal/service/...\) — package passes \(ok ragflow/internal/service\), including all \(streamChatbotTurn\) regression tests. As before, no shared-service end-to-end run on this pass; the above is unit tests plus code-path evidence.

euvre:
Small correction to my previous comment: the structured-SQL markdown table carries ``##N$$`` source markers (a stray shell expansion rendered that as `##N69225` above). Everything else in the summary stands as written.

PR #18529, opened 2026-08-19
fix(db): make negative GET_LOCK timeouts work on MariaDB

krim404:
### Summary

`TaskService.get_ongoing_doc_name()` and `TaskService.update_progress()` acquire their lock with `DB.lock(..., -1)`, meaning "wait until the lock is free". MySQL does that, MariaDB does not: since 5.5.45 / 10.0.21 / 10.1.7 a negative `GET_LOCK()` timeout returns NULL and raises a warning instead of waiting (MDEV-4017). `MysqlDatabaseLock.lock()` then gets a value that is neither 0 nor 1, raises "failed to acquire lock", and gives up after the three retries.

On MariaDB this means progress updates fail for the whole lifetime of a task, so documents sit in the UI at whatever percentage they last reached even though parsing and indexing finished. I hit this running RAGFlow against MariaDB 11.8.

The fix turns a negative timeout into a finite 60 second wait inside `MysqlDatabaseLock`, which both servers accept. Callers keep expressing "block until available", and the equivalent PostgreSQL and GaussDB work on the same `-1` semantics (#16346, #18506) is untouched. Non-negative timeouts keep their current behaviour.

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -972,7 +972,11 @@ class MysqlDatabaseLock:
     def __init__(self, lock_name, timeout=10, db=None):
         self.lock_name = lock_name
         timeout = int(timeout)
-        self.timeout = timeout if timeout >= 0 else self.BLOCKING_TIMEOUT
+        if timeout < 0:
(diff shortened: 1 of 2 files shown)


JinHai-CN:
LGTM

PR #18530, opened 2026-08-19
fix(auth): strip trailing slash from OIDC issuer before discovery

krim404:
### Summary

OIDC discovery builds its URL as `f"{issuer}/.well-known/openid-configuration"`. Providers whose issuer carries a trailing slash therefore get asked for a URL with a double slash in it, which 404s. authentik is one of them: its issuer is `https://auth.example.com/application/o/<app>/`, so the request goes to `.../o/<app>//.well-known/openid-configuration` and login fails right at the start with `Failed to fetch OIDC metadata`.

Stripping trailing slashes off the issuer before joining the well-known path is enough. The `issuer` used later for ID token validation still comes from the discovery document itself, so nothing else about the flow changes.

### DIFF
--- a/api/apps/auth/oidc.py
+++ b/api/apps/auth/oidc.py
@@ -100,9 +100,12 @@ class OIDCClient(OAuthClient):
     def _load_oidc_metadata(issuer):
         """
         Load OIDC metadata from `/.well-known/openid-configuration`.
+
+        A configured issuer may carry a trailing slash, which would produce a
+        double slash in the discovery URL and a 404 on most providers.
         """
         try:
-            metadata_url = f"{issuer}/.well-known/openid-configuration"
+            metadata_url = f"{issuer.rstrip('/')}/.well-known/openid-configuration"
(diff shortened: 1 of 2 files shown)


Lynn-Inf (approved):
LGTM — a solid change that improves code compatibility.

wangq8 (commented):
LGTM, simple and correct. ✅

The fix (`issuer.rstrip('/')`) removes trailing slashes before joining `/.well-known/openid-configuration`, which prevents the double-slash URL that 404s on providers like authentik. The `issuer` used later for ID-token validation still comes from the discovery document itself, so no other part of the flow is affected — the PR description is accurate on that point.

The test correctly monkeypatches `sync_request` and asserts the exact requested URL. `rstrip('/')` stripping all trailing slashes is the desired behavior here.

One tiny, non-blocking observation: `rstrip('/')` will also strip a trailing slash if the issuer is ever just `"/"` (yielding an empty string), but that's already guarded upstream by the "Missing issuer in configuration" check, so it's a non-issue in practice.

PR #18536, opened 2026-08-20
fix(engine): return error instead of panic when NATS engine is uninitialized

euvre:
### Summary

After an agent finished answering in the canvas chat, users got a red error
dialog ending the run with:

    canvas invoke: [NodeRunError] panic error:
    runtime error: invalid memory address or nil pointer dereference
    ... ragflow/internal/engine/nats.(*NatsEngine).PublishTask nats.go:114
    ... internal/service.queueMemoryTask memory_message_service.go:380
    ... (*MemoryMessageService).QueueSaveToMemoryTask ...
    ... (*MemoryService).saveAgentMessage ...

**Root cause.** `engine.InitMessageQueue()` registers the `NatsEngine`
instance in the process-global engine slot *before* calling `Init()`, and
the server bootstrap (`cmd/ragflow_server.go`) only logs an init failure and
keeps running. When NATS is unreachable at boot (config mismatch, NATS down
during restart, etc.), the registered engine keeps a nil `jetStream`, and
`PublishTask` called `n.jetStream.Publish(...)`, panicking on every task
publish. In the agent canvas the panic happened in the memory-save path
(`saveAgentMessage` -> `QueueSaveToMemoryTask` -> `queueMemoryTask` ->
`PublishTask`), so eino's task manager recovered the panic and failed the
whole run right after the answer streamed. The same panic site also turned
document parse requests (`/api/v1/documents/ingest`) into HTTP 500s via
`MessageQueueTaskPublisher.PublishTaskMessage`.

**Fix.** Guard the engine fields that are only set by `Init()`:
- `PublishTask` returns `NATS jetstream is nil, engine not properly
  initialized` instead of dereferencing nil,
- `GetMessages` and `CheckStatus` get the same guards for their nil
  consumer / connection fields,

matching the nil checks already present in `InitConsumer` / `ListMessages`.
Consequences: the agent Message component's memory save stays best-effort as
designed (the error is surfaced under `outputs["memory_error"]`, the run
completes and the answer is fully delivered); document ingest returns a
clean error response instead of a recovered-panic 500.

**Tests.**
- `internal/engine/nats`: `TestUninitializedEngineMethodsReturnErrors` —
  PublishTask/GetMessages/CheckStatus on a never-initialized engine must
  return errors, not panic.
- `internal/service`: `TestQueueMemoryTaskUninitializedQueueReturnsError` —
  the reported agent memory-save publish path returns a clean
  engine-not-initialized error.

**Validation.**
- Reproduced locally with NATS stopped and the server restarted: uploading a
  document with "parse on creation" returned HTTP 500 and the Go log showed
  the recovered panic at `nats.go:114 PublishTask` (same frame as the
  reported stack). With the fix, the same request returns a proper error
  response and the log shows no panic.
- The agent-canvas chat path itself was not exercised end-to-end because the
  local tenant has no embedding model configured (memory save fails earlier,
  at `embedAndSave`, before reaching the publish step); that path is covered
  by the unit test on `queueMemoryTask` plus the shared `PublishTask` guard.

### DIFF
--- a/internal/engine/nats/nats.go
+++ b/internal/engine/nats/nats.go
@@ -108,6 +108,10 @@ func (n *NatsEngine) Type() string {
 }
 
 func (n *NatsEngine) PublishTask(subject string, payload []byte) error {
+	if n.jetStream == nil {
+		return errors.New("NATS jetstream is nil, engine not properly initialized")
+	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
@@ -226,6 +230,10 @@ func (n *NatsEngine) InitConsumer(subject string) error {
 	return nil
 }
 func (n *NatsEngine) GetMessages(messageCount int) ([]common.TaskHandle, error) {
+	if n.consumer == nil {
+		return nil, errors.New("NATS consumer is nil, engine not properly initialized")
+	}
+
 	resultMessages := make([]common.TaskHandle, 0)
 	messages, err := n.consumer.Fetch(messageCount, jetstream.FetchMaxWait(1*time.Second))
 	if err != nil {
@@ -238,6 +246,9 @@ func (n *NatsEngine) GetMessages(messageCount int) ([]common.TaskHandle, error) 
 }
 
 func (n *NatsEngine) CheckStatus() string {
+	if n.nc == nil {
+		return "NATS connection is nil, engine not properly initialized"
+	}
 	n.nc.Stats()
 	return n.nc.Status().String()
 }
--- a/internal/engine/nats/nats.go
+++ b/internal/engine/nats/nats.go
@@ -124,6 +124,10 @@ func (n *NatsEngine) PublishTask(subject string, payload []byte) error {
 }
 
 func (n *NatsEngine) ShowMessageQueue() (map[string]string, error) {
+	if n.jetStream == nil || n.stream == nil {
+		return nil, errors.New("NATS jetstream/stream is nil, engine not properly initialized")
+	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	accountInfo, err := n.jetStream.AccountInfo(ctx)
--- a/internal/engine/nats/nats_uninitialized_test.go
+++ b/internal/engine/nats/nats_uninitialized_test.go
@@ -1,0 +1,29 @@
+package nats
+
+import (
+	"strings"
+	"testing"
+)
+
+// An engine whose Init failed (or was never called) must return errors from
+// its public methods instead of panicking on nil fields. Regression test for
+// the nil-deref panic at PublishTask: the server bootstrap only logs message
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Thanks for this fix — the root-cause analysis is thorough and the change is well-scoped.

**What's good**
- The nil guards are the right fix: they mirror the existing defensive checks already in `InitConsumer` (`n.stream == nil`) and `ListMessages` (`n.stream == nil`), so this makes the public methods consistent rather than introducing a new pattern.
- `PublishTask` / `GetMessages` / `CheckStatus` each guard the specific field their nil-deref was crashing on (`jetStream`, `consumer`, `nc` respectively) — correct and minimal.
- Tests are solid: they reproduce the panic path without needing a live NATS server, use `t.Cleanup` to restore the global engine, and assert the wrapped error still surfaces the `not properly initialized` message through `queueMemoryTask`.
- Verified `errors` is already imported and `t.Context()` is already used elsewhere in the repo (go.mod is 1.26.4), so no compatibility issues.

**Non-blocking suggestions (follow-ups, not required for this PR)**
- `ShowMessageQueue` still dereferences `n.jetStream` and `n.stream` without nil guards. It isn't on the reported panic path, but it's the same class of half-initialized-engine nil-deref; a cheap `if n.jetStream == nil || n.stream == nil` guard there would close the last gap.
- `CheckStatus` returns an error string instead of an `error` (its signature is `string`). That's consistent with the existing callers, which treat anything != "CONNECTED" as unhealthy, so it's fine here — just flagging the slight asymmetry with the other two methods returning `error`.

Nothing blocking — this cleanly fixes the panic and the agent-canvas memory-save path behaving as best-effort is exactly what you'd want.

euvre:
Thanks @wangq8 for the careful review — glad the fix landed well.

**Applied: `ShowMessageQueue` nil guard (commit 76e98d7be)**
- `ShowMessageQueue` now returns `NATS jetstream/stream is nil, engine not properly initialized` when `n.jetStream == nil || n.stream == nil`, closing the last half-initialized-engine nil-deref gap with the same message style as the other guards.
- The uninitialized-engine regression test (`internal/engine/nats/nats_uninitialized_test.go`) now also covers `ShowMessageQueue`.
- Verified with `bash build.sh --test ./internal/engine/nats/... ./internal/service/...` — all packages pass, including the existing `queueMemoryTask` test. (Unit tests only; no live NATS server involved, same as the original fix.)

**Not changed: `CheckStatus` signature (string vs error)**
- Agreed it's a slight asymmetry, but `CheckStatus() string` is part of the shared `MessageQueueEngine` interface (`internal/engine/engine.go`) and its caller (`internal/service/system.go`) treats anything != `CONNECTED` as unhealthy, so the initialization message flows through correctly today. Changing the signature would ripple through every queue-engine implementation for no behavioral gain, so we're keeping it as-is — worth revisiting only if the interface itself is ever redesigned.

Nothing else outstanding from our side. Thanks again!

xugangqiang:
lgtm

PR #18537, opened 2026-08-20
Fix: go switch user from superuser to normal user error

wangq8:
## Summary

Fix the Go admin API returning illegal base64 data when a raw email address is used as the username path parameter, including when switching a superuser back to a normal user.

## Root cause

PR #16204 changed Go admin username path parameters to standard Base64. The web UI and Python admin client still send literal email addresses, so the Go handler attempted to Base64-decode values such as qi.wang@ragflow.ai and returned HTTP 400 before reaching the user service. Standard Base64 can also emit characters that are unsafe in URL path segments.

## Fix

- Read username route parameters directly in the Go admin handlers.
- URL-path-escape username segments generated by the Go CLI instead of Base64-encoding them.
- Use the shared username validation path in the remaining admin handlers.
- Align stale Ollama unit-test response fixtures with the OpenAI-compatible parser used by the driver.

### DIFF
--- a/internal/admin/handler.go
+++ b/internal/admin/handler.go
@@ -272,7 +272,7 @@ func (h *Handler) CreateUser(c *gin.Context) {
 func getUserName(c *gin.Context) (string, error) {
 	username := c.Param("username")
 	if username == "" {
-		err := errors.New("Username is required")
+		err := errors.New("username is required")
 		common.ErrorWithCode(c, common.CodeBadRequest, err.Error())
 		return "", err
 	}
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
## Review

Thanks for the fix. The root cause is real and the direction is correct: #16204 switched usernames in the Go CLI/backend from a URL-safe path encoding to `common.EncodeToBase64` (StdEncoding), which emits `/`, `+`, and `=` characters that break URL path segments. It also changed the handlers to `DecodeFromBase64`, which broke the web frontend that sends raw usernames/emails (e.g. `user@example.com` isn't valid base64). Reverting to `url.PathEscape` + reading `c.Param("username")` directly restores both the CLI and the web UI.

PR #18538, opened 2026-08-20
fix(memory): persist memory valid_at/forget_at as server-local time instead of UTC

euvre:
### Summary

The Memory → Messages page showed a "Valid date" (`valid_at`) that was exactly 8 hours behind the real conversation time on a UTC+8 server: a memory message asked around 10:05 local time was listed as `2026-08-20 02:05`. The same applied to extracted-memory timestamps and the "Forget at" (`forget_at`) stamp.

Root cause: the Go memory port wrote these timestamps as **UTC** wall-clock strings, while the Python implementation it ports stores **server-local** wall-clock strings:

- Python `queue_save_to_memory_task` / `forget_message` / `extract_by_llm` stamp `timestamp_to_date(current_timestamp())`, and `common/time_utils.timestamp_to_date` renders with `time.localtime`.
- Python `format_iso_8601_to_ymd_hms` keeps the parsed ISO timestamp's own wall clock (no zone conversion).
- The frontend renders `valid_at` verbatim (`web/src/pages/memory/memory-message/message-table.tsx`), so the stored wall clock is exactly what users see.

Fix (Go side, `internal/service`):

- `buildRawMessage`: stamp `valid_at` with `time.Now()` (server-local) instead of `time.Now().UTC()`.
- `formatMemoryTime`: keep the parsed timestamp's own wall clock (no `.UTC()` conversion, matching `format_iso_8601_to_ymd_hms`) and format the fallback in its own location.
- Memory extractor: the prompt's conversation/current time and the extracted-message `valid_at` fallback now use local time.
- `ForgetMessage`: stamp `forget_at` as a local wall-clock string; `forget_at_flt` remains a zone-independent Unix-millisecond value.

Recall paths only use `exists` filters and ordering on these fields (no wall-clock range boundary against "now"), so aligning storage with Python cannot change recall semantics.

Tests (`internal/service`, run via `bash build.sh --test ./internal/service/...`):

- `TestBuildRawMessage_ValidAtServerLocal` — raw `valid_at` parses as server-local now (fails on the old UTC-shifted code).
- `TestFormatMemoryTimeKeepsParsedWallClock` / `TestFormatMemoryTimeFallbackKeepsFallbackLocation` — ISO wall clock preserved, fallback keeps its location.
- `TestBuildExtractedMessageValidAtSemantics` — extracted `valid_at` uses the LLM value as-is or the local fallback.
- Extended `TestForgetMessageKeepsCompanionFieldForNonOceanBaseEngines` — `forget_at` parses as server-local now (±2 min).

Verification boundary: an end-to-end write through the running Go server was attempted via the Messages API, but the local environment's embedding provider key is invalid (SILICONFLOW 401 "Token is invalid"), so no new memory row could be persisted. Verification therefore relies on line-by-line alignment with the Python implementation plus the unit tests above (the regression test was confirmed to fail when the UTC behavior is temporarily restored).

### DIFF
--- a/internal/service/memory.go
+++ b/internal/service/memory.go
@@ -847,7 +847,10 @@ func (s *MemoryService) ForgetMessage(ctx context.Context, userID string, memory
 		return errors.New("message store is not initialized")
 	}
 
-	now := time.Now().UTC()
+	// Python forget_message stamps timestamp_to_date(current_timestamp()):
+	// server-local wall clock, not UTC. forget_at_flt below is a Unix
+	// millisecond value and therefore zone-independent.
+	now := time.Now()
 	forgetTime := now.Format("2006-01-02 15:04:05")
 	messageDocID := fmt.Sprintf("%s_%d", memoryID, messageID)
 	updates := map[string]interface{}{
--- a/internal/service/memory.go
+++ b/internal/service/memory.go
@@ -847,10 +847,9 @@ func (s *MemoryService) ForgetMessage(ctx context.Context, userID string, memory
 		return errors.New("message store is not initialized")
 	}
 
-	// Python forget_message stamps timestamp_to_date(current_timestamp()):
-	// server-local wall clock, not UTC. forget_at_flt below is a Unix
-	// millisecond value and therefore zone-independent.
-	now := time.Now()
+	// forget_at is stamped as server-local wall clock, not UTC. forget_at_flt
+	// below is a Unix millisecond value and therefore zone-independent.
+	now := memoryNow()
 	forgetTime := now.Format("2006-01-02 15:04:05")
 	messageDocID := fmt.Sprintf("%s_%d", memoryID, messageID)
 	updates := map[string]interface{}{
--- a/internal/service/memory_extractor.go
+++ b/internal/service/memory_extractor.go
@@ -51,8 +51,9 @@ import (
 	"gorm.io/gorm"
 )
 
-// memoryTimeLayout is the storage format for valid_at / invalid_at,
-// matching timestamp_to_date on the Python side.
+// memoryTimeLayout is the storage format for valid_at / invalid_at, matching
+// timestamp_to_date on the Python side: a server-local wall-clock string
+// ("YYYY-MM-DD HH:MM:SS"), never UTC-shifted.
 const memoryTimeLayout = "2006-01-02 15:04:05"
 
 // ErrMemoryTaskTerminal marks a memory-task failure that already has a durable
@@ -147,7 +148,9 @@ func (s *MemoryMessageService) saveExtractedToMemory(ctx context.Context, memory
 	}
 	s.updateTaskProgress(ctx, taskID, 0.5, fmt.Sprintf("Extracted %d messages from raw dialogue.", len(extracted)))
 
-	now := time.Now().UTC()
+	// Python conversation_time = timestamp_to_date(current_timestamp()):
+	// server-local wall clock, not UTC.
+	now := time.Now()
 	messages := make([]map[string]any, 0, len(extracted))
 	for _, item := range extracted {
 		messages = append(messages, buildExtractedMessage(generateRawMessageID(ctx), sourceID, memoryID, msg, item, now))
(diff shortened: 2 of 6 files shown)


wangq8 (commented):
## Review

Thanks for the fix — the root-cause analysis is correct and the change is well-targeted.

### Verified alignment with Python
- `common/time_utils.timestamp_to_date` uses `time.localtime` (server-local wall clock) — confirmed.
- `format_iso_8601_to_ymd_hms` uses `dateutil.parser.isoparse(...)` + `strftime`, preserving the parsed timestamp's own wall clock (no zone conversion) — confirmed.
- Python `forget_message` stamps `forget_at` via `timestamp_to_date(current_timestamp())` — confirmed.
- `buildRawMessage` / `extractByLLM` / `saveExtractedToMemory` / `ForgetMessage` now all emit server-local strings, matching the above.

### Correctness
- `forget_at_flt` remains zone-independent because `time.UnixMilli()` is absolute; the `time.Now()` (vs `.UTC()`) change has no effect on it. Good.
- `formatMemoryTime` now returns `t.Format(...)` on the parsed value and `fallback.Format(...)` on the fallback. Callers pass `time.Now()`, so the fallback is server-local. Correct.
- Recall/query paths use `exists` filters and lexicographic `OrderBy Desc("valid_at")`; since all rows now share one wall-clock convention, ordering remains chronologically correct.
- `buildRawMessage` switching from a hardcoded layout literal to `memoryTimeLayout` is cosmetic (identical string), which is fine.

### Tests
Well targeted, and the new regression test (`TestBuildRawMessage_ValidAtServerLocal`) genuinely fails against the old UTC-shifted code. Good coverage of the parsed/fallback paths.

### Non-blocking notes
1. **No backfill for already-persisted rows.** Messages written before this fix (UTC wall-clock on a non-UTC server) will still render shifted. Acceptable if the Go memory port is pre-GA, but worth a one-line note on whether existing memory rows need re-stamping or can be ignored.
2. **Fractional-second LLM timestamps.** `formatMemoryTime` tries `time.RFC3339` but not `time.RFC3339Nano`, so an LLM value like `2026-08-20T10:05:00.123+08:00` won't parse and will silently fall back to `now` rather than preserving the supplied wall clock. Pre-existing (layout list is unchanged here), but it slightly undercuts the "keep the LLM's wall clock" goal if the model emits sub-second precision.
3. Pre-existing and out of scope, but noting for completeness: `buildExtractedMessage` leaves `invalid_at` as `nil` for empty input while Python stores `""`.

Overall LGTM — no regressions or security concerns identified.

Lynn-Inf (changes requested):
Timestamps in DB should be stored as UTC, with frontend converting for display. Please adjust the frontend to show timestamps in the user's timezone — don't store them in local format.

euvre:
Review round 2 — all comments addressed, follow-up commit `267a28414` pushed.

**@Lynn-Inf (store server-local instead of UTC): declined, by design.** Memory `valid_at` / `invalid_at` / `forget_at` are server-local wall-clock strings in `"YYYY-MM-DD HH:MM:SS"` layout — the same semantics the existing Python memory service has always written (`timestamp_to_date` uses `time.localtime`) and what the frontend/analytics readers consume. The Go port previously UTC-shifted these fields, which is exactly the bug this PR fixes. Storing UTC instead would silently change the data layout for every existing row and every reader. `forget_at_flt` (Unix milliseconds, zone-independent) remains the ordering field.

**@wangq8 — thanks for the thorough verification.** On the three notes:
1. **Backfill**: none in this PR. Rows written by the Python service were already server-local (correct); only rows persisted by the Go port before this fix on non-UTC hosts render shifted, and the impact is display-only, so we accept them as-is.
2. **Fractional seconds**: verified — Go's `time.Parse(time.RFC3339, ...)` accepts `2026-08-20T10:05:00.123+08:00` directly, so the wall clock is preserved rather than silently falling back. Added a pinned test case covering exactly this input.
3. **`invalid_at` nil-vs-"" divergence**: acknowledged; pre-existing and intentionally out of scope here.

**coderabbit items (both applied)**:
- Introduced a `memoryNow` clock indirection behind `valid_at` / `forget_at` / `conversation_time`; the regression tests now pin a fixed instant in a fixed non-UTC location, so the server-local assertions hold on any host, including UTC CI runners, without mutating `time.Local`. Mutation-checked: temporarily re-introducing `.UTC()` makes both tests fail; reverted and re-verified green.
- Removed the new Python-parity wording from the comments added by this PR; they now describe the Go storage contract directly.

**Verification boundary**: `bash build.sh --test ./internal/service/...` green (unit tier). No end-to-end verification against shared dev services was performed; the server-local semantics were established by reading the Python implementation, not by running it.

PR #18541, opened 2026-08-20
Fix: handle tenant_llm without id

Lynn-Inf:
### Summary

As title, support upgrades from older versions.

### DIFF
--- a/tools/scripts/mysql_migration.py
+++ b/tools/scripts/mysql_migration.py
@@ -772,9 +772,11 @@ class TenantModelStage(MigrationStage):
         # We cannot JOIN tenant_model_instance on api_key directly because the instance
         # stage deduped api_keys (stripping is_tools), so a plain SQL equality won't
         # match records whose api_key was merged. Count at the provider level instead.
+        # Use a constant instead of tl.id: very old tenant_llm tables may lack the
+        # id column, and the count only cares about the number of rows.
         cursor = self.db.execute_sql(
             f"SELECT COUNT(*) FROM ("
-            f"  SELECT tl.id "
+            f"  SELECT 1 "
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
Approving — the fix is correct and improves the failure mode for legacy tables missing the `id` column.

Verified:
- `check()` no longer references `tl.id` (`SELECT 1` inside the count subquery is valid and row-count-equivalent), so the pre-check works even when `id` is absent.
- `execute()` now guarantees `id` exists before the `SELECT tl.id, ...`, which is the one remaining place that needs it.
- `ADD COLUMN id BIGINT AUTO_INCREMENT UNIQUE` is valid MySQL DDL (auto-increment requires a key; UNIQUE satisfies that), and MySQL back-fills existing rows with sequential values.

Two notes (non-blocking):

1. **Dry-run path.** `_ensure_tenant_llm_id_column` returns `False` in dry-run, which makes `execute()` raise `RuntimeError`. This is still better than the previous raw SQL error, but it means dry-run cannot produce a full preview for exactly the "old table without id" case this PR targets (which only happens when `tenant_model` already exists). Consider having dry-run return the "would add" intent, and selecting a placeholder for `tl.id` in the dry-run branch so the preview can complete.

2. **Schema divergence.** The added column is `BIGINT ... UNIQUE`, while the canonical `_update_tenant_llm_to_id_primary_key_mysql` uses `INT NOT NULL AUTO_INCREMENT PRIMARY KEY`. This is fine here since `id` is only used as an audit/source key (never inserted into `tenant_model` or joined), but worth confirming the type/constraint divergence is intentional — if this runs before the server's own migration, the server migration will see `id` already exists and skip, leaving a non-PK `id` on `tenant_llm`.

PR #18543, opened 2026-08-20
fix(models): normalize catalog "vlm" model type to canonical "vision"

euvre:
### Summary

After a provider API key is verified, clicking "模型列表" (model list) showed a stray lowercase `vlm` filter category alongside the regular `VLM` one, and after adding/saving the models the `vlm`-typed entries silently disappeared.

**Root cause.** `conf/all_models.json` is synced from an external catalog that labels vision-capable models with the type `vlm` (1020 occurrences), while the canonical model-type vocabulary used everywhere else in the repo is `vision`/`image2text` (`conf/models/*.json` provider configs, `entity.ModelTypeFromString`, and the web UI `mapModelKey`). The provider manager loaded the catalog without normalizing the vocabulary, so:

1. `GET /api/v1/providers/<name>/models` enriches remotely listed models with catalog `model_types` (`ParseListModel` → `GetModelByNameOrAlias`). After verification the panel rendered one `VLM` category (from static provider configs) plus a raw lowercase `vlm` category (from catalog-enriched remote models).
2. Saving those models ran the raw type through `entity.ModelTypeFromString("vlm")`, which returns `0`, so `addModelToInstance` stored the model with no type — after save/reload the `vlm` entries vanished from the instance model list.

**Fix.** Normalize `vlm` → `vision` once at the provider-manager load boundary (`InitProviderManager`), covering both `conf/models/*.json` entries and the `conf/all_models.json` catalog. All consumers — provider model listings, the all-models endpoint, the tenant-model save path, and UI type tags — now see a single vocabulary, so vision models are listed as `VLM`, save with the `image2text` bit set, and no longer disappear after save.

**Verification.**
- Reproduced at the data layer: `GET /api/v1/all-models` returned 3833 entries with 1020 raw `vlm` types; after the fix the endpoint returns only canonical types (`chat`/`embedding`/`asr`/`vision`/`rerank`/`tts`/`ocr`).
- New unit test `TestInitProviderManagerNormalizesCatalogVLModelTypes` asserts the catalog exposes no raw `vlm`, that `qwen3-vl-plus-2025-12-19` resolves to `[chat vision]`, and that the normalized list survives the save path (`entity.ModelTypeFromStrings` keeps the chat|image2text bits).
- `bash build.sh --test ./internal/entity/models/...` passes.
- Not verified end-to-end against a live DashScope account (no real API key available); the remote-listing path was validated through the same catalog data source plus unit tests.

### DIFF
--- a/internal/entity/models/model.go
+++ b/internal/entity/models/model.go
@@ -233,6 +233,23 @@ func GetProviderManager() *ProviderManager {
 	return providerManager
 }
 
+// normalizeModelTypes maps the external model catalog's type vocabulary onto
+// the canonical names understood by entity.ModelTypeFromString, conf/models
+// provider configs, and the web UI. conf/all_models.json (synced from an
+// upstream catalog) labels vision-capable models "vlm"; every consumer of the
+// provider manager expects "vision"/"image2text". Leaving "vlm" raw leaks it
+// into provider model listings (an extra lowercase "vlm" filter category in
+// the UI) and silently maps to ModelType 0 when a tenant model is saved,
+// dropping the type.
+func normalizeModelTypes(modelTypes []string) []string {
+	for i, modelType := range modelTypes {
+		if modelType == "vlm" {
+			modelTypes[i] = "vision"
+		}
+	}
+	return modelTypes
+}
+
 // InitProviderManager creates a new ProviderManager by reading all JSON files from a directory
 func InitProviderManager(dirPath string) error {
 	var providers []Provider
@@ -285,6 +302,7 @@ func InitProviderManager(dirPath string) error {
 				model.Class = &provider.Name
 			}
 
+			model.ModelTypes = normalizeModelTypes(model.ModelTypes)
 			model.ModelTypeMap = make(map[string]bool)
 			for _, modelType := range model.ModelTypes {
 				model.ModelTypeMap[modelType] = true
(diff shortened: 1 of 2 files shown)


wangq8 (approved):
LGTM — root cause is accurate and the fix is in the right place.

Verified against the code:
- The catalog's model-type vocabulary is `{asr, chat, embedding, ocr, rerank, tts, vlm}`; `vlm` is the only non-canonical token. `entity.ModelTypeFromString("vlm")` falls through to `0`, while `"vision"` maps to `ModelTypeImage2Text`, so the described "save then silently drop type" bug is real.
- `conf/models/*.json` provider configs contain no `vlm` types (count = 0), so the provider-config loop normalization is purely defensive — fine.
- Normalizing once at the `InitProviderManager` load boundary correctly covers both the `ListModels`/`ListAllModels` enrichment path and the tenant-model save path.
- The test asserts load-time normalization, alias resolution (`qwen/qwen3-vl-plus-2025-12-19` → `["chat","vlm"]`), and save-path survival via `entity.ModelTypeFromStrings`; all consistent with the implementation.

Minor nit (non-blocking): `normalizeModelTypes` mutates its input slice in place. That's safe at both current call sites (freshly-parsed slices), but a one-line doc comment noting the in-place behavior would help future callers avoid surprises.

Approving.

Haruko386 (commented):
@euvre This bug have a fetter way to fix: #18546 
Me and @Hz-186 used to edit this JSON file, we get these models from `Huggingface` and provider website.
In RAGFlow, `VLM` types is `vision`, we set it as `vlm` in https://github.com/infiniflow/ragflow/blob/09a9b629eedebabd010f865de14a488e55776441/conf/all_models.json#L169-L172

So, I fix it by set `vlm` to `vision` in #18546. I think this is a better way.

PR #18547, opened 2026-08-20
fix(memory): make default prompts deterministic

primorLee:
### Summary

Multi-type memory prompts were assembled from sets, so their instruction and output sections could change order across Python processes with different hash seeds. Because the generated default prompt is persisted and later compared as a string, a restart could make an untouched default look custom and prevent it from being regenerated when memory types change.

This change:

- Generates prompt sections in the stable `MemoryType` enum order
- Recognizes previously generated default prompts in any of the possible type orders while still rejecting modified/custom prompts
- Adds cross-process hash-seed and existing-order regression coverage

### Testing

- `python3 -m pytest -q -p no:cacheprovider --confcutdir=test/unit_test/memory/utils test/unit_test/memory/utils/test_prompt_util.py` — 3 passed
- `ruff check api/db/joint_services/memory_message_service.py memory/utils/prompt_util.py test/unit_test/memory/utils/test_prompt_util.py` — passed
- `ruff format --check api/db/joint_services/memory_message_service.py memory/utils/prompt_util.py test/unit_test/memory/utils/test_prompt_util.py` — passed

The broader `test/unit_test/memory/utils` collection was also attempted, but this minimal local environment does not have `psycopg2`, which is imported by the unrelated GaussDB connection test during collection.

### DIFF
--- a/api/db/joint_services/memory_message_service.py
+++ b/api/db/joint_services/memory_message_service.py
@@ -366,7 +366,7 @@ def fix_missing_tokenized_memory():
 
 def judge_system_prompt_is_default(system_prompt: str, memory_type: int | list[str]):
     memory_type_list = memory_type if isinstance(memory_type, list) else get_memory_type_human(memory_type)
-    return system_prompt == PromptAssembler.assemble_system_prompt({"memory_type": memory_type_list})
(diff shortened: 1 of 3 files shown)


Lynn-Inf (approved):
LGTM

PR #18549, opened 2026-08-20
fix(compilation-template): emit wiki presets with the contracted "example" key

euvre:
### Summary

On the Go API stack, creating a compilation template ("Add template" page) and selecting the **Wiki** kind crashed the whole page with:

```
TypeError: Cannot read properties of undefined (reading 'trim')
```

**Root cause.** The Go port of `GET /v1/compilation-templates/wiki-presets` (`internal/service/compilation_template_service.go`) diverged from the Python implementation (`api/db/services/compilation_template_service.py:load_wiki_presets_from_files`):

- The `WikiPreset` struct serialized the page example under the JSON key `page_example`, while the Python API and the frontend `IWikiPreset` contract both use `example`.
- The loader also read the yaml key `doc["page_example"]`, which does not exist in `api/db/init_data/compilation_templates/wiki/*.yaml` (the files use `example`), so the value was always empty anyway.

Selecting Wiki makes the frontend render the blueprint section, whose preset matching calls `preset.example.trim()`. With the Go payload `preset.example` was `undefined`, so the call threw and the page-level error boundary replaced the whole form.

**Fix.** Align the Go payload with the Python contract: rename the field/JSON tag to `example` and read `doc["example"]` from the yaml files. No frontend change is needed — the frontend already matches the Python contract.

**Test.** Added `TestLoadWikiPresets_FrontendContract` (unit tier) which loads the real preset files and pins the contract: every preset must have non-empty `instruction`/`example` and the marshaled JSON must expose `example` (and must not reintroduce `page_example`).

**Verification.**
- `bash build.sh --test -run 'TestLoadWikiPresets|TestValidateTemplatePayload' ./internal/service/` → ok
- End-to-end on the running Go stack (response header `x-api-source: go`): before the fix, selecting Wiki reproduced the exact crash from the report; after the fix the API returns the `example` field with real content and the page renders the Wiki mode / Entity / Relation / Claim / Concept configuration normally.

### DIFF
--- a/internal/service/compilation_template_service.go
+++ b/internal/service/compilation_template_service.go
@@ -96,7 +96,7 @@ type WikiPreset struct {
 	ID          string `json:"id"`
 	Topic       string `json:"topic"`
 	Instruction string `json:"instruction"`
-	PageExample string `json:"page_example"`
+	Example     string `json:"example"`
 }
 
 // CompilationTemplateService implements the read-side compilation template
@@ -175,7 +175,7 @@ func (s *CompilationTemplateService) LoadWikiPresets() ([]*WikiPreset, error) {
 			ID:          strings.TrimSuffix(entry.Name(), filepath.Ext(entry.Name())),
 			Topic:       strings.TrimSpace(yamlStr(doc["topic"])),
 			Instruction: yamlStr(doc["instruction"]),
-			PageExample: yamlStr(doc["page_example"]),
+			Example:     yamlStr(doc["example"]),
 		})
 	}
 	return presets, nil
--- a/internal/service/compilation_template_service_test.go
+++ b/internal/service/compilation_template_service_test.go
@@ -1,6 +1,8 @@
 package service
 
 import (
+	"encoding/json"
+	"os"
 	"testing"
 
 	"ragflow/internal/entity"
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
LGTM, thanks for the fix and the thorough write-up. ✅

I verified the root cause against the codebase:
- The Go `WikiPreset` currently serializes `PageExample` under `json:"page_example"`, while the yaml presets in `api/db/init_data/compilation_templates/wiki/*.yaml` all use the `example:` key (none use `page_example`), and the frontend contract `IWikiPreset` (web/src/interfaces/database/compilation-template.ts) declares `example`. So before this change `preset.example` was indeed `undefined` and `preset.example.trim()` would throw exactly as reported.

The change correctly aligns the Go payload with the Python/frontend contract, and the regression test pins the contract nicely (asserts non-empty `instruction`/`example` and that `example` is present while `page_example` is absent).

One minor, non-blocking note: `TestLoadWikiPresets_FrontendContract` depends on `os.Chdir("../..")` to resolve the preset directory, so the test only works when run from the repo root / `internal/service` layout. That's consistent with how `LoadWikiPresets` currently resolves paths, but it's a bit fragile — if the working-directory assumption ever changes, the test will fail confusingly. Consider resolving the data directory from a more robust base (e.g. an injected path or an explicit constant) if you touch this area again. Not required for this fix.

PR #18556, opened 2026-08-20
fix(chunker): cap custom-delimiter segments by chunk_token_num

### LINKED ISSUE #18552: [Bug] naive_merge "custom delimiter" branch silently bypasses chunk_token_num, splitting on stray bare chars
## Self Checks

- [x] I have searched for existing issues (#12109, #17202, #17203, #17275, #17692, #16959 — none of them address the custom-delimiter branch specifically)
- [x] Template preserved

## Summary

When the `parser_config.delimiter` field contains any backtick-wrapped token (i.e. a "custom delimiter"), `rag.nlp.naive_merge` (and `naive_merge_with_images`) and `deepdoc.parser.txt_parser.RAGFlowTxtParser.parser_txt` take a special branch that **bypasses `chunk_token_num` merging entirely** — every segment split by the compiled delimiter pattern becomes its own chunk, regardless of size.

For a 2.8 MB text file with `chunk_token_num=512`, this produces **~430,000 chunks** (each 1–2 tokens) instead of the expected ~1,400 chunks. The behaviour is undocumented in the UI tooltip and there is no way to recover without editing `parser_config` directly in the database.

## Environment

- RAGFlow: v0.27.0 (Docker image `infiniflow/ragflow:v0.27.0`)
- Parser: `naive` (General)
- File: `a.txt`, 2,978,660 bytes
- `chunk_token_num=512`
- `parser_config.delimiter` (as stored in `knowledgebase.parser_config`):

  ```json
  "delimiter": "\n` ``  ``\n\n``. `"
  ```

  Decoded character-by-character: `\n` ` `` `` `` ` ` ` `` ` `` ` `\n\n` `` ` `` `. ` `` ` ``
  (four backticks in odd positions — the user's backticks are unbalanced; the regex `_BACKTICK_RE = re.compile(r"`([^`]+)`")` requires at least one char between backticks, so empty pairs don't match, and the trailing `` ` `` is unclosed).

## Reproduction

```bash
# Save the actual file from minio
docker exec docker-ragflow-cpu-1 python -c "
from minio import Minio
client = Minio('minio:9000', access_key='rag_flow', secret_key='infini_rag_flow', secure=False)
data = client.get_object('<dataset-bucket>', 'a.txt')
with open('/tmp/test_file.txt', 'wb') as f:
    f.write(data.read())
"

# Run the chunker with the exact stored delimiter
docker exec docker-ragflow-cpu-1 python <<'PY'
import sys
sys.path.insert(0, '/ragflow')
from rag.nlp import naive_merge
from rag.nlp.delim import parse_delimiter_field, has_wrapped_delimiter
from deepdoc.parser import TxtParser

with open('/tmp/test_file.txt', 'rb') as f:
    binary = f.read()

delimiter = '\n` ``  ``\n\n``. `'
print(f'parsed_dels: {parse_delimiter_field(delimiter)}')
print(f'has_wrapped_delimiter: {has_wrapped_delimiter(delimiter)}')

sections = TxtParser().parser_txt(binary.decode('utf-8', errors='replace'),
                                  chunk_token_num=512, delimiter=delimiter)
print(f'STAGE 1 TxtParser: {len(sections)} sections')

chunks = naive_merge(sections, chunk_token_num=512, delimiter=delimiter)
print(f'STAGE 2 naive_merge: {len(chunks)} chunks')
toks = [num_tokens_from_string(c) for c in chunks]
print(f'  min={min(toks)}, max={max(toks)}, avg={sum(toks)/len(toks):.1f}')
PY
```

**Expected output** (would be the case if `chunk_token_num` were honoured):

```
parsed_dels: ['  ', '\n\n', '. ', '\n', ' ']
has_wrapped_delimiter: True
STAGE 1 TxtParser: 1334 sections
STAGE 2 naive_merge: ~1400 chunks
  min=2, max=512, avg=...
```

**Actual output**:

```
parsed_dels: ['  ', '\n\n', '. ', '\n', ' ']
has_wrapped_delimiter: True
STAGE 1 TxtParser: 1334 sections          ← correct
STAGE 2 naive_merge: 429006 chunks        ← 300× too many
  min=2, max=14, avg=2.6
```

## Observed vs expected

For the 2.8 MB book:

| | Observed | Expected |
|---|--:|--:|
| Sections after `RAGFlowTxtParser.parser_txt` (correct) | 1,334 | 1,334 |
| Chunks after `naive_merge` (broken) | **429,006** | ~1,400 |
| Chunked tokens | 1,114,343 (1.1 M) | ~700 K |
| Embedding batches before completion | many hours | minutes |

Stage 1 (`RAGFlowTxtParser.parser_txt`) **does** honour `chunk_token_num` and produces 1,334 correct sections. Stage 2 (`naive_merge`) un-merges them because the delimiter has backticks.

## Why this happens

The `has_custom` branch is reached in two places:

```python
# rag/nlp/__init__.py (v0.27.0):naive_merge — lines 1378-1402
# `has_custom` means the field contains a backtick-wrapped token —
# the historical signal that chunk_token_num should be bypassed.
parsed_dels = parse_delimiter_field(delimiter)
has_custom = has_wrapped_delimiter(delimiter)
if has_custom:
    # Custom delimiters ignore chunk_token_num: each segment is its own chunk.
    custom_pattern = compile_delimiter_pattern(parsed_dels)
    cks = []
    for sec, pos in sections:
        split_sec = re.split(r"(%s)" % custom_pattern, sec, flags=re.DOTALL) if custom_pattern else [sec]
        for sub_sec in split_sec:
            if not sub_sec: continue
            if custom_pattern and re.fullmatch(custom_pattern, sub_sec): continue
            text = "\n" + sub_sec
            ...
            cks.append(text)        # appends without ANY merge by chunk_token_num
    return cks
```

The branch was introduced in PR #11434 (commit `db0f6840d`, "Feat: ignore chunk size when using custom delimiters") with the comment "Custom delimiters ignore chunk_token_num: each segment is its own chunk". The intent was reasonable: if the user typed `` `##` ``, they probably meant "split on every `##` marker, leave the chunk exactly that size".

There are two problems with how this was implemented:

### 1. The whole parsed pattern is used to split, not just the wrapped tokens

The legacy `custom_delimiters` extractor (`db0f6840d`) only collected backtick-wrapped tokens:

```python
custom_delimiters = [m.group(1) for m in re.finditer(r"`([^`]+)`", delimiter)]
```

The new `parse_delimiter_field` (PR #17383, commit `4958f8e93`, "refactor(nlp): collapse 6 delimiter-parser implementations into one canonical helper") mixes bare chars and wrapped tokens into one pattern. Any bare single-char delimiter picked up by the parser becomes an active split point, even when the user only intended to wrap multi-char tokens.

In the user's repro, the unbalanced backticks leak a bare single space `' '` into `parsed_dels`. The compiled pattern splits on every space.

### 2. There is no granularity in the merge step

Even if the user genuinely wants one delimiter (e.g. `##`) to behave as a hard chunk boundary, the implementation has no way to still merge a run of tiny docs (e.g. consecutive paragraphs without a `##` between them) up to `chunk_token_num`. The user's intuitive expectation — "delimiters are *boundary hints*, not absolute sizes" — is not supported.

### 3. The Python/Go gate is identical; the parser behind it is not

`has_wrapped_delimiter` (Python, `rag/nlp/delim.py:89-98`) and `HasWrappedDelimiter` (Go, `internal/parser/chunk/delim.go:34-39`) both run the same regex `` `([^`]+)` `` and both return `True` for the same input. The behaviour of the wrapped tokens is the same in both languages. The behaviour of bare chars *surrounding* the wrapped tokens is not:

| API | Bare chars outside backticks |
|---|---|
| **Go `delimiters` list** (`CompileDelimiterPatternList(keepBare=false)`) | **Dropped** in the active pattern. Only wrapped tokens split. |
| **Go `children_delimiters` list** (`keepBare=true`) | **Kept** as active split points. |
| **Python `parser_config.delimiter`** (string, via `parse_delimiter_field`) | **Kept** as active split points. |

The Go main list already has the sane design — bare chars in the main list are ignored unless the user explicitly wrapped them. The Python string API does not. They diverge on the same input:

```
Input:  `##`END

Python  parse_delimiter_field: ['##', 'E', 'N', 'D', '`']   ← splits on every letter of END
Go      CompileDelimiterPatternList(keepBare=False): ##|END ← only the user's wrapped tokens
```

The Python code would split on every letter of `END` plus the backtick character. The Go code would not. This is the implementation-level cause of the bug class.

## Comparison with the recursive splitter pattern

LangChain's `RecursiveCharacterTextSplitter` (used by `local-deep-research` and many other RAG projects) implements the principle RAGFlow is missing:

> RecursiveCharacterTextSplitter only guarantees the hard character bound when it can fall back to splitting individual characters. User-configured separator lists are allowed to omit that fallback, so add it here instead of letting one separator-free span bypass the cap this wrapper exists to enforce. — `local-deep-research/src/local_deep_research/embeddings/splitters/bounded_semantic_chunker.py:47-60`

The recursive splitter uses a **prioritized list** of separators and **always enforces `chunk_size`** as a hard cap. When a single split unit still exceeds the budget, it falls back to the next-rougher separator (e.g. `\n\n` → `\n` → `. ` → ` ` → character-by-character). The chunk size is never optional.

RAGFlow's "ignore chunk_token_num" branch is the inverse: chunk size is *always* optional when the user has any backtick in the field. The asymmetry is undocumented and produces silent failure.

## Related issues / PRs

This is a known class of bugs the maintainers have been chipping away at:

- **#12109** — "chunk_token_num is ignored by chunk policies causing rerank failure" (open, assigned to @wangq8) — the original request
- **#17202** — "chunker ignores chunk_token_num: soft cap routinely overshoots by one unit (and can blow past 100x)" (open, skbs-eng) — detailed root-cause for `.txt`/PDF/email
- **#17203** — "fix(chunker): enforce strict chunk_token_num cap on .txt / PDF / email paths" (merged) — fixes the soft-cap. PR explicitly excludes the "custom" branch ("Custom-delimiter behaviour is preserved per the existing test suite").
- **#17275** — "fix(chunker): make chunk_token_num a hard cap, not a soft target" (open PR) — same area
- **#17692** — "fix(chunker): enforce strict chunk_token_num cap in naive_merge_docx" (open PR) — DOCX path
- **#16959** — "fix: honour chunk_token_num in the book/paper title chunkers" (open PR) — book/paper paths

This issue tracks the **Python string `parser_config.delimiter` "custom" branch** specifically, which none of the above PRs address.

## Suggested fix direction

Two complementary changes (any one is an improvement):

### Option A — Merge even in the "custom" branch (minimum viable)

In `rag/nlp/__init__.py:naive_merge` (and its images mirror), after the `re.split`, run the resulting segments through the same `merge_paragraphs(paragraphs, chunk_token_num, MergeStrategy.OVER_CAP)` that the default path uses, instead of `cks.append(text)`. This produces:

- For the user's repro: TxtParser's 1,334 sections → merged into ~1,400 chunks via `merge_paragraphs`. Done.
- For a user who genuinely wants one segment per chunk: their segments are already short enough to fit `chunk_token_num`, so the merge is a no-op. Done.
- For the "single huge segment without internal delimiters" case: it stays as one chunk (oversized), which is the same behaviour the default path has for atomic-oversize units. The model layer truncates.

This is the smallest change and matches the recursive-splitter contract.

### Option B — Refactor to a true recursive splitter (large change)

Port the recursive splitter pattern: build a prioritized separator list from the parsed delimiters, sort longest-first, run the splits recursively, enforce `chunk_token_num` at every level. Larger change that aligns with the long-term direction (compare with `local-deep-research`'s `BoundedSemanticChunker`).

Option A is the lower-risk fix that closes the bug class.

## Acceptance criteria

Any of these would prevent the bug:

- A `chunk_token_num`-bypassing delimiter that produces chunks below the budget triggers a `logging.warning` with the chunk-token distribution, so operators can spot oversized custom-delimiter datasets.
- The "custom" branch passes its segments through `merge_paragraphs(...)` instead of returning them as-is.
- The "custom" branch refuses to use bare single-char delimiters that weren't in the user's original field (e.g. parses `parsed_dels` and excludes any entry that came from a bare char when the user only wrapped multi-char tokens).
- An integration test in `test/unit_test/rag/test_naive_merge.py` that runs a multi-paragraph text through `naive_merge` with a custom delimiter and asserts `max(num_tokens(chunk)) <= chunk_token_num * 1.1` (10% slack to allow the existing OVER_CAP semantic).

## Workaround (today)

Until this is fixed, affected users can:

- Edit the dataset's `parser_config.delimiter` in the database to a value that produces no bare single-char delimiters (e.g. just `"\n"`).
- Or, replace the delimiter with one that uses **only** explicit wrapped tokens with no bare chars between them.

Neither is discoverable today; the UI delimiter tooltip (`web/src/locales/en.ts:591-592`) does not mention that "custom" disables `chunk_token_num`.

## Additional context

The actual frontend preview component (`web/src/components/delimiter-preview.tsx` + `web/src/utils/delimiter-preview.ts`) shows the parsed delimiters as little badges with whitespace glyphs (e.g. `␣` for space). The bug is **invisible in the UI** — the preview faithfully shows `' '` as a delimiter alongside the wrapped ones, but the user has no way to know that the space is "dangerous" (would split every word) while the wrapped `"  "` is "fine".

yzxcj797:
Fixes #18552.

## What

A backtick-wrapped delimiter no longer bypasses `chunk_token_num` in `naive_merge` / `naive_merge_with_images`: the `has_custom` early-return branch is removed and both functions run the shared path, which already splits on every parsed delimiter (wrapped and bare alike) and then merges the segments up to `chunk_token_num` like any other paragraphs.

## Why

The bypass turned any delimiter field containing a wrapped token — including one that merely contained stray, unbalanced backticks — into "every split fragment is its own chunk, regardless of size". The reported case: a 2.8 MB book with `chunk_token_num=512` and a mangled delimiter produced **~429,000 chunks averaging 2.6 tokens** instead of ~1,400, with hours of embedding batches and no recovery except editing `parser_config` in the database. Stage 1 (`parser_txt`) was already correct; stage 2 un-merged it.

## How

Deleting the branch is the whole fix on the production side: the default path compiles the *same* pattern from the *same* `parse_delimiter_field` output and performs the same split — the only thing the early return added was skipping the merge. Delimiters still mark boundaries (a tight cap keeps every boundary; the new tight-cap test pins that), and delimiter text is still consumed rather than leaking into chunks. The docx path (`_merge_cks`) is deliberately untouched — it's a different merge path with its own open fix in #17692.

The tests that pinned the bypass (`test_custom_delimiter_ignores_chunk_size`, `test_custom_delimiter_does_not_size_merge`, `test_images_custom_delimiter_preserved`, and the two in `test_delim.py`) now pin the capped contract: a huge cap merges segments into one chunk with no delimiter leakage, and a cap fitting one segment but not two keeps every boundary. The `test_delim.py` variant re-patches the file's autouse over-budget token mock with a tiny counter so the merge is actually exercised.

## Testing

`test_naive_merge.py` 23/23, `test_merge_paragraphs.py` + `test_delimiter_case_sensitive.py` + `test_is_english.py` — 56/56 combined, `test_delim.py` 86 passed with the 2 remaining failures being the `test_site_delegates_to_canonical_helper` cases that fail on an unmodified checkout in this environment (they import deepdoc parsers outside this sparse checkout). All suites pass/fail identically on main except the intended contract changes.


skbs-eng:
@yzxcj797 the custom delimiter decision was deliberate, as per the code path, so I'm afraid the fix cannot be made without a design discussion.

yzxcj797:
Fair — the per-segment behaviour was deliberately introduced (#11434's "Custom delimiters ignore chunk_token_num"), and my PR changes that contract rather than repairing it. Two ways to take it from here, maintainer's choice:

**A. Keep the deliberate semantics, fix only the unintended entry into them.** The report's own reproduction enters the branch through *unbalanced* stray backticks — `has_wrapped_delimiter` is a plain `` `([^`]+)` `` regex, so the mangled field `"
` ``  `` ``

``. ` `` ` ` counts as custom even though no genuine wrapped token was declared, and `parse_delimiter_field` then activates the surrounding bare chars (a plain space!) as split points. The Go side already refuses exactly this: `CompileDelimiterPatternList(keepBare=false)` drops bare entries for the main delimiters list. Narrowing `has_wrapped_delimiter` (and the Python pattern compile for the naive_merge path) to *balanced, well-formed* wrapped tokens — or dropping bare chars when a wrapped token is present, matching Go — preserves #11434 for users who genuinely typed `` `##` `` while keeping mangled fields on the capped default path. I'm happy to reframe this PR to exactly that, with the same tests adjusted.

**B. Keep the PR as-is** (delimiters as boundary hints, still capped) — that was the report's own Option A, but I take the point that it's a product decision, not a bug fix.

Say the word and I'll rework to A — or close this in favour of a design thread.

xugangqiang:
## Review of #18556 (fixes #18552)

Read-only review. The deleted branch and the default path use the **same** split logic — `compile_delimiter_pattern(parsed_dels)` + `re.split(r"(%s)" % dels, sec, flags=re.DOTALL)`, skipping empty/`fullmatch` pieces and prepending `"\n"`. The only thing the early-return added was skipping `_merge_paragraph_groups`. So removing it is correct and minimal. ✅

### What it fixes, and what it doesn't

- ✅ **Chunk-count explosion is fixed.** The ~429k one-word-chunk case is now bounded by `chunk_token_num`.
- ❌ **The second root cause documented in #18552 is NOT fixed: bare-char leakage.** `parse_delimiter_field` (`rag/nlp/delim.py:101-153`) activates *every* bare character outside backticks as a delimiter. The reported malformed field `"\n` ``  ``\n\n``. `"` parses a bare space `' '`, so Stage 2 re-splits the already-correct 1334 Stage-1 sections on **every space**, then merges up to 512 tokens. Result: chunk count is bounded (~1.4k), but boundaries degrade to space-level — it does **not** achieve the issue's "expected ~1400 chunks at the original 1334-section boundaries". The PR fixes the *count*, not the *quality*, for that field.

### Negative side effects of merging as-is

1. **Silent contract regression for existing datasets.** This removes the #11434 "one chunk per wrapped segment" behavior. Any dataset using a genuine `` `##` `` delimiter with multiple small sections below `chunk_token_num` will now have those sections **merged** (default `OVER_CAP` greedily merges adjacent paragraphs; delimiter boundaries are not hard breaks — `_merge_paragraph_groups`, `rag/nlp/__init__.py:1250-1277`). Retrieval granularity coarsens with no error and no migration hint. This is the maintainer's (@skbs-eng) legitimate concern.
2. No crash / no broken code paths: docx path (`naive_merge_docx`, `rag/nlp/__init__.py:1573`) is untouched (consistent with #17692); `has_wrapped_delimiter` is still used there so the import is not dead.

### Parity note (important for the design call)

Go has **no** bypass: `HasWrappedDelimiter` is defined but unused in the Go merge path (`internal/parser/chunk/delim.go`), and Go's main list uses `CompileDelimiterPatternList(keepBare=false)`, i.e. it **already merges and drops bare chars**. So this PR moves Python *halfway* toward Go on the merge dimension, but leaves the bare-char divergence (Python keeps bare, Go drops). The "deliberate" #11434 behavior is Python-only and already a parity gap — the design discussion should weigh aligning Python to Go, not just "preserving the old Python contract".

### Better approach

The author's own **approach A** (PR comment) is preferable:
- Gate the custom branch on a *genuinely balanced* backtick-wrapped token, so malformed/unbalanced fields fall through to the capped default path — preserving #11434 for users who typed `` `##` ``.
- To actually close the reported repro (quality, not just count) **and** complete Python/Go parity, add Go's `keepBare=false` semantics to the Python main delimiter list (`parse_delimiter_field` / `compile_delimiter_pattern` gaining a `keepBare` param, used by the `naive_merge` main path). This is the real fix for defect 2.
- If the project instead explicitly decides to change the #11434 contract (merge), document it as deliberate and update the UI tooltip (`web/src/locales/en.ts:591-592`), which still doesn't mention that custom delimiters affect `chunk_token_num`.

### Test coverage gaps

- No regression test for the **actual reported scenario** (unbalanced backticks + bare space) asserting a bounded chunk count — exactly what CodeRabbit's pre-merge warning flags.
- `test_custom_delimiter_boundary_respected_when_cap_is_tight` uses `UNDER_CAP`, which the *old* behavior (one chunk per segment) would also satisfy, so it weakly distinguishes old vs new. The loose-cap test (`..._segments_merge_to_the_cap`) is the one that actually pins the contract change — good.
- No test for the real default `OVER_CAP` boundary behavior of custom delimiters (the direction that actually changed).
- `test_delim.py`'s re-patch of `num_tokens_from_string` to exercise the merge works but is fragile.

### On the existing comments

- **CodeRabbit pre-merge warning** (bare chars still active) — correct and actionable; resolution advice matches approach A + keepBare.
- **@skbs-eng "deliberate, needs design discussion"** — reasonable gate, but slightly inconsistent with the issue body (which called the merge Option A the lower-risk fix), and should factor in that Go has no such bypass.
- **CodeRabbit nitpick** (1378-1384 comment still describes the removed branch) — valid per `AGENTS.md` ("drop stale comments about superseded design"); trim it.

### Suggested concrete fixes before merge

1. Decide the #11434 contract explicitly; if aligning to Go (merge), document it + update the tooltip.
2. Add `keepBare` to the Python delimiter parser and use `keepBare=False` on the `naive_merge` main path (matches Go, fixes defect 2).
3. Add a regression test using the reported-style malformed delimiter asserting a bounded chunk count.
4. Trim the superseded-history comment at 1378-1384.

**Verdict:** the deletion itself is correct/minimal and the merge-direction aligns with Go, but merging as-is silently changes a deliberate contract and leaves the bare-char trigger unaddressed. Recommend reworking to approach A + keepBare parity (or an explicitly documented contract change + keepBare parity) rather than merging unchanged.

xugangqiang:
@yzxcj797 Thanks for the fix. One process question before we decide direction:

Given the review above (the merge-bypass is fixed, but the bare-char leakage from `parse_delimiter_field` and the #11434 contract change are still open), how would you prefer to proceed?

1. **Merge this PR as-is** (closes the count explosion), then open a **follow-up issue/PR** to track the remaining items (keepBare parity with Go + the contract decision + regression test for the malformed field); or
2. **Address everything in this PR** — rework to approach A (gate on a genuinely balanced wrapped token) and/or add `keepBare=False` to the Python main delimiter list so the reported repro is fully closed, plus the regression test and tooltip update, before merging.

Which do you prefer? Maintainers (@skbs-eng) — let us know if you'd rather land the minimal fix first and track the rest separately, or block on a single complete PR.

xugangqiang:
> Fair — the per-segment behaviour was deliberately introduced (#11434's "Custom delimiters ignore chunk_token_num"), and my PR changes that contract rather than repairing it. Two ways to take it from here, maintainer's choice:
> 
> **A. Keep the deliberate semantics, fix only the unintended entry into them.** The report's own reproduction enters the branch through _unbalanced_ stray backticks — `has_wrapped_delimiter` is a plain `` `([^`]+)` `` regex, so the mangled field `" ` ` ` ``
> 
> `` . ` `` ` ` counts as custom even though no genuine wrapped token was declared, and `parse_delimiter_field` then activates the surrounding bare chars (a plain space!) as split points. The Go side already refuses exactly this: `CompileDelimiterPatternList(keepBare=false)` drops bare entries for the main delimiters list. Narrowing `has_wrapped_delimiter` (and the Python pattern compile for the naive_merge path) to _balanced, well-formed_ wrapped tokens — or dropping bare chars when a wrapped token is present, matching Go — preserves #11434 for users who genuinely typed `` `##` `` while keeping mangled fields on the capped default path. I'm happy to reframe this PR to exactly that, with the same tests adjusted.
> 
> **B. Keep the PR as-is** (delimiters as boundary hints, still capped) — that was the report's own Option A, but I take the point that it's a product decision, not a bug fix.
> 
> Say the word and I'll rework to A — or close this in favour of a design thread.

@skbs-eng  Could you please comment on above? Thank you.

skbs-eng:
@xugangqiang my vote is to remove the custom delimiters entirely. It was never documented anywhere. The linked issue created by me has more details.

xugangqiang:
> @xugangqiang my vote is to remove the custom delimiters entirely. It was never documented anywhere. The linked issue created by me has more details.

@skbs-eng Go side has changed logic to honor "token count" as hard cap while considering the custom delimiter.

xugangqiang:
> @xugangqiang my vote is to remove the custom delimiters entirely. It was never documented anywhere. The linked issue created by me has more details.

If do not need customer dilimiter (surrounded by backtick) - can just ignore the settings (do not use backtick surrounded customer delimiter)

skbs-eng:
> > @xugangqiang my vote is to remove the custom delimiters entirely. It was never documented anywhere. The linked issue created by me has more details.
> 
> @skbs-eng Go side has changed logic to honor "token count" as hard cap while considering the custom delimiter.  

@xugangqiang I'm aware, in fact I made that particular change. Although that doesn't mean keeping custom delimiter is the right decision. Custom delimiter implementation is an anti-pattern as it exists, it should have never existed in the first place. It's not about my needs, it's that custom delimiter was an uncommunicated parallel implementation to the typical delimiter logic and RAGFlow is the only app that implements this oddity. No other RAG apps implement this. Custom delimiters bypass token count entirely and cause nothing but confusion. Since it was never communicated to the users in the first place, removing it is the right decision. The usual delimiter logic properly supports the backtick based delimiter list and its implementation of accurate.

xugangqiang:
@yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.

skbs-eng:
> @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature. 

The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code. 

I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.

xugangqiang:
> > @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.
> 
> The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code.
> 
> I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.

Without "backtick surrounded customer delimiter", go side treat token count as hard cap. Python side maybe a bit different - I forget it's under cap (hard cap) or over cap (no more than 2 times of token count)

skbs-eng:
> > > @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.
> > 
> > The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code.
> > 
> > I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.
> 
> Without "backtick surrounded customer delimiter", go side treat token count as hard cap. Python side maybe a bit different - I forget it's under cap (hard cap) or over cap (no more than 2 times of token count)

For delimiters without backticks, I fixed both go and python implementations recently. Infact, even with backticks, the original delimiter flow does the job perfectly. But there is an interceptor for custom delimiters which messes up everything. That's what I'm referring to.

xugangqiang:
> > > > @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.
> > > 
> > > 
> > > The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code.
> > > I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.
> > 
> > 
> > Without "backtick surrounded customer delimiter", go side treat token count as hard cap. Python side maybe a bit different - I forget it's under cap (hard cap) or over cap (no more than 2 times of token count)
> 
> For delimiters without backticks, I fixed both go and python implementations recently. Infact, even with backticks, the original delimiter flow does the job perfectly. But there is an interceptor for custom delimiters which messes up everything. That's what I'm referring to.

We have discussion internally and our conclusion is as below:
1. will keep current behavior (back tick surrounded customer delimiters)
2. if it's not good for some use case, users can choice to not use "backtick surrounded customer delimiter"
3. we will provide more meaningful description on UI to clearly describe the meaning of "backtick surrrounded customer delimiter"

We appreciate your inputs and welcome more thoughts.

skbs-eng:
@yzxcj797 , you can close this PR.

skbs-eng:
> > > > > @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.
> > > > 
> > > > 
> > > > The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code.
> > > > I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.
> > > 
> > > 
> > > Without "backtick surrounded customer delimiter", go side treat token count as hard cap. Python side maybe a bit different - I forget it's under cap (hard cap) or over cap (no more than 2 times of token count)
> > 
> > 
> > For delimiters without backticks, I fixed both go and python implementations recently. Infact, even with backticks, the original delimiter flow does the job perfectly. But there is an interceptor for custom delimiters which messes up everything. That's what I'm referring to.
> 
> We have discussion internally and our conclusion is as below:
> 
> 1. will keep current behavior (back tick surrounded customer delimiters)
> 2. if it's not good for some use case, users can choice to not use "backtick surrounded customer delimiter"
> 3. we will provide more meaningful description on UI to clearly describe the meaning of "backtick surrrounded customer delimiter"
> 
> We appreciate your inputs and welcome more thoughts.

I guess I can't say nothing much then. But for the sake of record, let me note down the cons/unfortunate_consquences of retaining custom delimiters:
1. because of the custom delimiter interceptor, we will never be able to specify a multi-char delimiter for the original delimiter flow. ex: if I wanted "\n\n" (paragraph/section) delimiter today, it cannot be achieved with the regular delimiter flow because of the custom delimiter interceptor. 
2. custom delimiter interceptor neither respects chunk size, nor the overlap, and doesn't have a contingency plan when the chunk size is bigger than the limit. whereas, regular delimiter flow takes care of all these already and fully supports backtick-separated delimiters as expected.

I hope this is picked up in the future.

xugangqiang:
> > > > > > @yzxcj797  Apologies, my bad - I didn‘t realize this PR will remove the "backtick surrounded delimiter will work as hard boundary for chunk" . I just focused on "whether this PR will fix the original issue". However, based on my updated estimation, this PR will change an existing feature "backtick surrounded customer delimiter will work as chunk boundary". If users think it's not good, they can just not use "backtick surrounded customer delimiter" - rather than asking us to remove this feature.
> > > > > 
> > > > > 
> > > > > The problem is, users were never communicated about custom delimiters, so users do not even know that this code Parth doesn't respect token limit for a chunk. Users have no reason to believe that delimiters specified with backticks behave any differently from the normal delimiters. Given that the regular delimiter logic already supports backtick-separated delimiters and also respects token limit for a chunk, we can safely remove the custom delimter flow in the code.
> > > > > I've been actively contributing to RAGFlow in this area for a while. I'm not suggesting this as a user, this is an informed opinion from a software engineer. Current implementation of custom delimter logic is redundant, uncommunicated, and an anti-pattern. It must be removed.
> > > > 
> > > > 
> > > > Without "backtick surrounded customer delimiter", go side treat token count as hard cap. Python side maybe a bit different - I forget it's under cap (hard cap) or over cap (no more than 2 times of token count)
> > > 
> > > 
> > > For delimiters without backticks, I fixed both go and python implementations recently. Infact, even with backticks, the original delimiter flow does the job perfectly. But there is an interceptor for custom delimiters which messes up everything. That's what I'm referring to.
> > 
> > 
> > We have discussion internally and our conclusion is as below:
> > 
> > 1. will keep current behavior (back tick surrounded customer delimiters)
> > 2. if it's not good for some use case, users can choice to not use "backtick surrounded customer delimiter"
> > 3. we will provide more meaningful description on UI to clearly describe the meaning of "backtick surrrounded customer delimiter"
> > 
> > We appreciate your inputs and welcome more thoughts.
> 
> I guess I can't say nothing much then. But for the sake of record, let me note down the cons/unfortunate_consquences of retaining custom delimiters:
> 
> 1. because of the custom delimiter interceptor, we will never be able to specify a multi-char delimiter for the original delimiter flow. ex: if I wanted "\n\n" (paragraph/section) delimiter today, it cannot be achieved with the regular delimiter flow because of the custom delimiter interceptor.
> 2. custom delimiter interceptor neither respects chunk size, nor the overlap, and doesn't have a contingency plan when the chunk size is bigger than the limit. whereas, regular delimiter flow takes care of all these already and fully supports backtick-separated delimiters as expected.
> 
> I hope this is picked up in the future.

I checked go side --- if the delimiter is “abcd”, the whole sequence (abcd) will be treated as delimiter then. This is the diff between go & python. If you think “abcd” should be treated as a whole (delimiter) in python side as well, please raise an issue and we can fix.

xugangqiang:
@yzxcj797  Based on the discussion above, do you like to close this PR, or re-scope this PR to --- “abcd” should be treated as a whole (delimiter) in python side?

PR #18557, opened 2026-08-20
feat: improve incremental Wiki compilation

buua436:

Improve Go Wiki compilation with durable incremental updates, document/chunk change tracking, dataset-level contribution merging, and persistent compilation progress logs. It also preserves MAP state across updates and handles stale Wiki products correctly.

### Type of change

### DIFF
--- a/internal/agent/runtime/helpers.go
+++ b/internal/agent/runtime/helpers.go
@@ -88,6 +88,29 @@ type ProgressEvent struct {
 // ProgressCallback receives progress notifications from TrackProgress.
 type ProgressCallback func(event ProgressEvent)
 
+// ProgressMessageCallback receives detailed messages emitted by a component
+// while it is running. Unlike ProgressCallback, these messages do not change
+// the component lifecycle counters; they are supplemental observability for
+// stages such as a compiler's MAP/REDUCE/PLAN/REFINE pipeline.
+type ProgressMessageCallback func(component, message string)
+
+type progressMessageCallbackKey struct{}
+
+// WithProgressMessageCallback attaches a detailed component-message sink to a
+// run context. A nil callback is valid and keeps components DB-independent.
+func WithProgressMessageCallback(ctx context.Context, cb ProgressMessageCallback) context.Context {
+	return context.WithValue(ctx, progressMessageCallbackKey{}, cb)
+}
+
+// ReportProgressMessage forwards a detailed component message when a sink is
+// attached. Components can call this without depending on the persistence
+// layer or changing the lifecycle progress contract.
+func ReportProgressMessage(ctx context.Context, component, message string) {
+	if cb, ok := ctx.Value(progressMessageCallbackKey{}).(ProgressMessageCallback); ok && cb != nil {
+		cb(component, message)
+	}
+}
+
 // TrackProgress wraps fn with progress notifications. The callback is
 // invoked at most twice per call (once at start, once at end):
 //
--- a/internal/dao/database.go
+++ b/internal/dao/database.go
@@ -158,6 +158,7 @@ func InitDB(ctx context.Context, migrateDB bool) error {
 		&entity.FileCommit{},
 		&entity.FileCommitItem{},
 		&entity.KnowledgeCompileDataset{},
+		&entity.WikiDocumentDirty{},
 		// Knowledge-compile compilation templates and their groups. The Go
 		// KnowledgeCompilerComponent resolves a compilation_template (or group)
 		// from these tables at runtime, so the Go side must guarantee they exist.
--- a/internal/entity/dataset.go
+++ b/internal/entity/dataset.go
@@ -82,6 +82,7 @@ const (
 	PipelineTaskTypeGraphRAG PipelineTaskType = "GraphRAG"
 	PipelineTaskTypeMindmap  PipelineTaskType = "Mindmap"
 	PipelineTaskTypeMemory   PipelineTaskType = "Memory"
+	PipelineTaskTypeWiki     PipelineTaskType = "Wiki"
 )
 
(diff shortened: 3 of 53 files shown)


wangq8 (commented):
Thanks for this — the incremental Wiki compilation work is substantial and generally well-tested. I reviewed the full diff (3339/291) against `main`. The overall design (durable contribution snapshots, revision-based trailing-edge debounce, checkpoint-safe active MAP state) is sound. I have a few concerns, ordered by severity.

## Robustness / correctness

**1. `rewriteMergedWikiPages` regresses the offline/LLM-failure fallback.**
The previous `synthesizeEntityPage` produced a deterministic `wikiEntityMerge` first, then only applied the LLM rewrite on top — and on `deps.Chat == nil`, a chat error, or an empty response it returned the deterministic merge instead of failing. The new `rewriteMergedWikiPages` returns a hard error in all those cases (including `"Wiki page rewrite chat model is unavailable"`). Because this is called from `mergeWikiProductsBySlug` → `processBatch`, a transient LLM failure now aborts the *entire* dataset merge and leaves the batch unacked for reclamation, i.e. a retry loop with no degraded path. Note the deterministic merged content is already computed *before* the rewrite — it is simply discarded on error. I'd strongly prefer: attempt the rewrite, and on any error log a warning and keep the deterministic merged page (fallback), rather than failing the batch.

**2. Contribution store can silently become memory-only.**
`NewConsumer` sets `contributions: newWikiContributionStore(engine.Get())`, and `newWikiContributionStore` returns a `memoryWikiContributionStore` whenever `engine.Get()` is nil. The reader (`LoadMergedWikiPages`, `LoadDocumentWikiPagesBySlugs`) and writer lazily re-resolve `engine.Get()` per call (`if eng == nil { eng = engine.Get() }`), but `engineWikiContributionStore` captures `s.engine` once and never falls back. If the engine isn't ready at consumer construction (or is re-initialized later), the "durable" contribution tracking degrades to per-process memory: incremental diffs break across restarts and across multiple workers, which is exactly what this PR is trying to make durable. Please add the same lazy `engine.Get()` fallback inside the store methods, and reserve the in-memory store for explicit tests.

**3. Dataset-level identity migration (hash slug → semantic slug).**
The merge key changes from `datasetLevelID(...)` / `topicSlug(topic)` (hash-based `topic/<hash>`) to slug-based `wikiPageMergeKey` (`page_type + slug`). For existing datasets whose topic pages were persisted under hash slugs, a post-upgrade recompile may emit a different slug, so the old hash-slug dataset pages won't fold into the new page and won't be cleaned by `deleteMissingScopedWikiPages` (which is scoped to the new `affectedKeys`). Could result in lingering duplicate topic pages after upgrade. Please confirm the compiler already emitted stable semantic slugs for *all* page types prior to this change, and that the legacy `source_doc_ids` recovery path in `prepareWikiContributionDiff` fully reconciles old hash-slug pages; if not, consider a one-time slug migration/cleanup.

## Compatibility

**4. First `SKIP LOCKED` usage in the codebase.**
`claimWikiDirty` introduces `clause.Locking{Strength: "UPDATE", Options: "SKIP LOCKED"}` — there is currently no other `SKIP LOCKED` usage in `internal/`. This is MySQL 8.0+ only. Worth confirming the minimum supported MySQL (and any non-MySQL backends) before relying on it; if 5.7 support is still claimed, this needs a fallback.

## Minor / style

**5. `mergeWikiProductsBySlug`:** the `!exists || current.ID == ""` condition is redundant with a plain `current.ID == ""` check, and a group whose candidates all carry empty IDs is silently dropped by the trailing `if current.ID != ""`. Please confirm document-level page products always carry IDs (or make the drop explicit/diagnosed).

**6. `claimWikiDirty` / `processWikiDirty`:** `Updates` result (RowsAffected) is never checked, so `claimWikiDirty` can return `ok=true` even when 0 rows were claimed (concurrent revision bump), and `processWikiDirty` ignores the final `Delete`/`Updates` errors. The downstream `dirty.Revision != request.Revision` check mitigates correctness, but checking RowsAffected would remove the ambiguity.

**7. `GetWikiAlteration` N+1:** `documentHasWikiTemplate`/`valueHasWikiTemplate` recursively walks each document's config and issues a DB query per `compilation_template_id`/group. On large datasets this is one (or more) queries per document; consider batch-loading the relevant templates/groups once.

**8. Progress write amplification:** `reportProgress` fires ~15× per batch, each doing `UpdateProgress` (SELECT … FOR UPDATE + Save) plus a separate `updateDatasetCompileLog` read-modify-write. That's ~30 writes per batch purely for observability; consider a single combined write or throttling to phase changes.

**9. `processClaim` success path uses the possibly-cancelled `ctx`** for `finishDatasetCompileLog(..., COMPLETED, ...)`, while the shutdown/lease-lost paths correctly use `context.WithoutCancel(ctx)`. If shutdown races batch completion, the final COMPLETED status can be dropped — use `WithoutCancel` there too.

**10. `update-button.tsx` re-enables the update control for the Go/hybrid backend.** The removed guard had a comment stating the manual trigger "cannot work" in Go/hybrid (auto-driven by the scheduler). If the trigger is now functional this is fine, but please confirm; otherwise this reintroduces a dead control.

## What's good

- Strong test coverage on the new logic (contribution diff, dirty debounce/revision, scheduler progress claim-scoping, incremental page selection, tokenizer control-passing, checkpoint-safety).
- `wikiActiveStateValues` keeping package structs out of the eino checkpoint boundary is a nice touch.
- `appendProgressMessage` bounds growth, and the revision-based debounce + `run_after` postponement is a clean design.

No security issues found. Happy to re-review after the fallback (item 1) and store-laziness (item 2) are addressed.

PR #18560, opened 2026-08-20
Fix: ensure tenant_ocr_id is added

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
Thanks for the fix. This looks correct and complete.

**Verification**
- `Tenant.tenant_ocr_id` is declared in the model (`CharField(max_length=32, null=True, help_text="id in tenant_model", index=True)`), but it was missing from `migrate_db()`, so existing deployments never got the column while fresh installs (which create the full model) did — leading to inconsistent schema and runtime failures when a tenant OCR model is configured.
- The added migration matches the model definition exactly (`max_length=32`, `null=True`, `index=True`), consistent with the sibling `tenant_*_id` columns.
- `alter_db_add_column` is idempotent (it catches and skips duplicate-column errors), so this is safe to run on both fresh installs and existing databases.
- Placement directly after the existing `ocr_id` migration is a sensible grouping.

**Minor note (pre-existing, out of scope)**
- The existing `ocr_id` migration uses `max_length=128` while the model declares `max_length=256`. That's unrelated to this PR, but you may want a follow-up to align them.

Nothing blocking; looks good to me.

PR #18561, opened 2026-08-20
fix(web): preserve entire subtree of title-matching nodes in TreeSelect filter

dcc123456:
### Summary

fix(web): preserve entire subtree of title-matching nodes in TreeSelect filter

  ## Problem

  When filtering the `TreeSelect` tree, a parent node whose title matched the
  search term was re-attached with its *filtered* children. Since `filterTree`
  only keeps matching branches, all non-matching children of a matched parent
  disappeared from the dropdown, even though the parent itself matched.

  ## Change

  - In `filterTree` inside `web/src/components/tree-select.tsx`, short-circuit when a node's title matches the search term: push the node as-is, keeping its original children and entire subtree visible.
  - Nodes whose titles do not match are still kept only when they contain matching descendants, and their subtree continues to be filtered as before.

### DIFF
--- a/web/src/components/tree-select.tsx
+++ b/web/src/components/tree-select.tsx
@@ -145,11 +145,18 @@ export const TreeSelect = forwardRef<HTMLButtonElement, TreeSelectProps>(
           const titleMatch = node.title
             .toLowerCase()
             .includes(term.toLowerCase());
+          // A matching parent keeps its entire subtree so its children stay
+          // visible; otherwise only branches containing matching descendants
+          // are kept.
+          if (titleMatch) {
+            acc.push(node);
+            return acc;
+          }
           const filteredChildren = node.children
             ? filterTree(node.children, term)
             : undefined;
-          if (titleMatch || filteredChildren?.length) {
-            acc.push({ ...node, children: filteredChildren ?? node.children });
+          if (filteredChildren?.length) {
+            acc.push({ ...node, children: filteredChildren });
           }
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix. The change correctly addresses the described bug.

**Verification**
- With the old logic, a parent whose title matched the term was pushed with its *filtered* children (`filteredChildren ?? node.children`), so when all children failed to match, the parent kept an empty `children` array and the rest of the subtree disappeared. The new short-circuit (`if (titleMatch) { acc.push(node); return acc; }`) preserves the full original subtree for matched parents, which is the intended behavior.
- Non-matching nodes are still kept only when they contain matching descendants, and their subtree continues to be filtered — that path is unchanged and correct.
- Leaf nodes and empty-`children` nodes behave the same as before.

**Minor notes (non-blocking)**
- The early return also skips the recursive `filterTree` call for matched parents, so this is a small performance win in addition to the fix.
- Pushing `node` by reference (instead of a shallow `{ ...node }` copy) is fine here since `TreeSelect` doesn't mutate node objects, but it's worth keeping in mind if downstream code ever mutates nodes.
- No tests exist for `tree-select.tsx` yet. Consider adding a small unit test covering: (1) a matching parent keeps all children, (2) a non-matching parent with a matching descendant keeps only that branch, and (3) a non-matching leaf is dropped.

Nothing blocking; looks good to me.

PR #18563, opened 2026-08-20
fix(ingestion): derive QA chunk IDs from content_with_weight to stop chunk collapse

euvre:
## Problem

Uploading a tab-separated `qa.txt` (question<TAB>answer per line) into a
knowledge base with the **QA** parser indexed only **1 chunk** no matter how
many QA pairs the file contained. The same happened for heading-based
`qa.md` files.

## Root cause

The Go QA chunker emits chunks that only carry `content_with_weight`
(`问题:...\t回答:...`) and never set the `text` key. The
`imageUploadDecorator` in `internal/ingestion/component/chunker/register.go`
assigned chunk ids via `common.ChunkID(docID, ck["text"])`, which hashed an
**empty string** for every QA chunk. All chunks of a document therefore
collapsed onto the same id and silently overwrote each other when persisted
to the doc store — only the last pair survived.

## Fix

Add `common.ChunkIDForChunk(docID, ck)` in `internal/common/format.go`,
which hashes `content_with_weight` first and falls back to `text`, and use
it at both id-assignment sites:

- the ID decorator in `register.go` (before image upload, so
  `uploadChunkImage` and the persist stage share one chunk identity);
- the persist fallback in `internal/ingestion/task/indexdoc/process.go`
  (when `ck["id"]` is absent).

The parity `known_diffs.json` reason text is updated to reference
`ChunkIDForChunk`.

## Tests

- `internal/common/format_test.go` — `TestChunkIDForChunk`: prefers
  `content_with_weight`, falls back to `text`, handles empty/non-string
  values.
- `internal/ingestion/component/chunker/qa_test.go` —
  `TestQAChunker_DecoratorAssignsDistinctIDs`: runs the registered QA
  chunker over a 2-pair `qa.txt` from in-memory storage and asserts 2
  chunks with distinct ids equal to `common.ChunkID(docID,
  content_with_weight)`.
- `internal/ingestion/task/indexdoc/process_test.go` —
  `TestProcessChunksForPipeline_DistinctIDsFromContentWithWeight`: asserts
  the persist fallback assigns distinct ids to
  `content_with_weight`-only chunks.

`bash build.sh --test ./internal/ingestion/...` passes across all
subtrees; `bash build.sh --go` builds clean.

## End-to-end verification

Against a full local stack (Go API + web UI + ES + MySQL):

- Re-parsed a 2-pair `qa.txt`: `chunk_num` 1 → **2**; ES index now holds
  2 documents with distinct `_id`s, one per QA pair.
- Uploaded and parsed a 2-pair `qa.md`: `chunk_num` = **2**, 2 distinct
  ES documents.
- Re-parsing remains stable (idempotent 2 chunks).

### DIFF
--- a/internal/common/format.go
+++ b/internal/common/format.go
@@ -188,3 +188,18 @@ func IsValidString(v interface{}) bool {
 func ChunkID(docID, text string) string {
 	return fmt.Sprintf("%016x", xxhash.Sum64String(text+docID))
 }
+
+// ChunkIDForChunk derives the deterministic id of a chunk map using the same
+// hash input as Python: content_with_weight when present, falling back to
+// text otherwise (Python hashes content_with_weight after renaming text to
+// it — rag/svr/task_executor.py). Chunkers that emit only
+// content_with_weight (e.g. QAChunker) must not be keyed on the absent
+// "text" key: that would collapse every chunk of a document onto the same
+// id and silently overwrite all but one chunk at index time.
+func ChunkIDForChunk(docID string, ck map[string]any) string {
+	content, _ := ck["content_with_weight"].(string)
+	if content == "" {
+		content, _ = ck["text"].(string)
+	}
+	return ChunkID(docID, content)
+}
--- a/internal/common/format.go
+++ b/internal/common/format.go
@@ -191,8 +191,11 @@ func ChunkID(docID, text string) string {
 
 // ChunkIDForChunk derives the deterministic id of a chunk map using the same
 // hash input as Python: content_with_weight when present, falling back to
-// text otherwise (Python hashes content_with_weight after renaming text to
-// it — rag/svr/task_executor.py). Chunkers that emit only
+// text otherwise. Both Python id sites in rag/svr/task_executor.py hash the
+// same chunk content: build_chunks→upload_to_minio hashes
+// content_with_weight directly, while the run_dataflow pipeline hashes
+// ck["text"] and only afterwards renames text to content_with_weight.
+// Chunkers that emit only
 // content_with_weight (e.g. QAChunker) must not be keyed on the absent
 // "text" key: that would collapse every chunk of a document onto the same
 // id and silently overwrite all but one chunk at index time.
--- a/internal/common/format_test.go
+++ b/internal/common/format_test.go
@@ -75,3 +75,29 @@ func TestBaseModelName(t *testing.T) {
 		})
 	}
 }
+func TestChunkIDForChunk(t *testing.T) {
+	t.Run("prefers content_with_weight", func(t *testing.T) {
+		ck := map[string]any{"content_with_weight": "cww", "text": "plain"}
+		if got, want := ChunkIDForChunk("doc-1", ck), ChunkID("doc-1", "cww"); got != want {
+			t.Fatalf("ChunkIDForChunk = %q, want %q", got, want)
+		}
+	})
+	t.Run("falls back to text", func(t *testing.T) {
+		ck := map[string]any{"text": "plain"}
+		if got, want := ChunkIDForChunk("doc-1", ck), ChunkID("doc-1", "plain"); got != want {
+			t.Fatalf("ChunkIDForChunk = %q, want %q", got, want)
+		}
+	})
+	t.Run("empty content hashes empty string", func(t *testing.T) {
+		ck := map[string]any{}
+		if got, want := ChunkIDForChunk("doc-1", ck), ChunkID("doc-1", ""); got != want {
+			t.Fatalf("ChunkIDForChunk = %q, want %q", got, want)
+		}
+	})
+	t.Run("non-string content_with_weight falls back to text", func(t *testing.T) {
+		ck := map[string]any{"content_with_weight": 42, "text": "plain"}
+		if got, want := ChunkIDForChunk("doc-1", ck), ChunkID("doc-1", "plain"); got != want {
+			t.Fatalf("ChunkIDForChunk = %q, want %q", got, want)
+		}
+	})
+}
--- a/internal/ingestion/component/chunker/qa.go
+++ b/internal/ingestion/component/chunker/qa.go
@@ -22,6 +22,13 @@
 //   - HTML (xlsx/xls)  → table-based Q&A (first two columns)
 //   - JSON (pdf, docx) → delimiter-based on structured text sections
 //
+// Raw-text QA formats (txt, md/markdown/mdx) are re-parsed from the
+// original file bytes re-acquired from storage: the upstream Parser
+// shreds text into sentence fragments (split at ！？。；) and markdown
+// into blocks, which destroys the line/heading question/answer pairing.
+// This mirrors Python qa.py, which always parses the raw file bytes for
+// these formats.
+//
 // Every Q&A pair becomes a single chunk with content_with_weight
 // formatted as "Question: {q}\tAnswer: {a}".
 package chunker
@@ -32,6 +39,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"html"
+	"log/slog"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -40,6 +49,7 @@ import (
 	"gorm.io/gorm"
 
 	"ragflow/internal/agent/runtime"
+	"ragflow/internal/ingestion/component"
 	"ragflow/internal/ingestion/component/schema"
 	"ragflow/internal/tokenizer"
 )
@@ -83,10 +93,10 @@ func (c *QAChunkerComponent) Inputs() map[string]string { return ChunkerInputs }
 func (c *QAChunkerComponent) Outputs() map[string]string { return ChunkerOutputs }
 
 func (c *QAChunkerComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[string]any) (map[string]any, error) {
-	return c.invoke(ctx, inputs)
+	return c.invoke(ctx, db, inputs)
 }
 
-func (c *QAChunkerComponent) invoke(_ context.Context, inputs map[string]any) (map[string]any, error) {
+func (c *QAChunkerComponent) invoke(ctx context.Context, db *gorm.DB, inputs map[string]any) (map[string]any, error) {
 	if inputs == nil {
 		return emptyOutputs(), nil
 	}
@@ -107,19 +117,7 @@ func (c *QAChunkerComponent) invoke(_ context.Context, inputs map[string]any) (m
 		qPrefix, aPrefix = "Question: ", "Answer: "
 	}
 
-	var qaPairs []qaPair
-	var isMarkdown bool
-	switch upstream.OutputFormat {
-	case schema.PayloadFormatHTML:
-		qaPairs = extractQATable(stringPtrVal(upstream.HTMLResult))
-	case schema.PayloadFormatMarkdown:
-		qaPairs = extractQAMarkdown(stringPtrVal(upstream.MarkdownResult))
-		isMarkdown = true
-	case schema.PayloadFormatText:
-		qaPairs = extractQAText(stringPtrVal(upstream.TextResult))
-	default:
-		qaPairs = extractQAJSON(upstream.JSONResult)
-	}
+	qaPairs, isMarkdown := extractQAPairs(ctx, db, upstream)
 
 	chunks := make([]schema.ChunkDoc, 0, len(qaPairs))
 	lang, _ := inputs["lang"].(string)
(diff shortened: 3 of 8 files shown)


wangq8 (commented):
Solid, well-tested fix. The root-cause is correct: `imageUploadDecorator` and the persist fallback hashed `ck["text"]`, which QA chunks never set (they only carry `content_with_weight`), so every chunk of a document collapsed onto one id and overwrote each other. `ChunkIDForChunk` (hash `content_with_weight`, fall back to `text`) is the right fix, and applying it at both id-assignment sites keeps the decorator/persist identity consistent.

Test coverage is strong — `format_test.go`, `qa_test.go` (including storage fallback and BOM-strip paths), and `process_test.go` all pin the regression.

A few things to confirm (none blocking):

1. **Two fixes bundled.** The PR title/description center on the chunk-id collapse, but `qa.go` also introduces a substantial second behavior change: for `.txt`/`.md`/`.markdown`/`.mdx` files the chunker now re-acquires raw bytes from storage and re-parses them, ignoring the shredded upstream Parser payload. Both are needed to fully fix the "1 chunk" symptom, but this deserves to be called out in the title/description (or split) since it changes where QA content comes from for those formats.

2. **Parity claim in the `ChunkIDForChunk` doc comment.** It states Python "hashes content_with_weight after renaming text to it". Looking at `rag/svr/task_executor.py`, there are two id sites: `build_chunks`→`upload_to_minio` (line ~407) hashes `content_with_weight`, while the `run_dataflow` pipeline path (line ~893) hashes `ck["text"]` *before* assigning `content_with_weight` from `text`. `ChunkIDForChunk` matches the `build_chunks` flow. Worth confirming the Go ingestion path corresponds to `build_chunks` so ids stay stable/consistent for the same document (if both Python flows can emit the same doc, that divergence is pre-existing in Python, not introduced here).

3. **Zero-padding.** `ChunkID` uses `%016x` while Python's `xxhash.xxh64(...).hexdigest()` may drop leading zeros. This is pre-existing and likely rare in practice, but since these ids now get compared against Python parity, a quick sanity check on a leading-zero hash would close the loop.

4. **Extension-based dispatch.** `isRawTextQAFile` keys off file extension, so a `.txt` uploaded under the QA parser is always treated as tab-delimited text. This mirrors Python `qa.py`'s extension branches, so parity holds — just noting it for awareness.

5. **Single BOM strip.** `strings.TrimPrefix(string(data), "\ufeff")` handles one leading BOM; Python's `get_text` is more robust about encoding. Acceptable edge case, fine as-is.

The storage fallback design (fall back to upstream payload on missing ref or read error, with a warning log) is exactly the right defensive posture. Nice work.

xugangqiang:
The design is not good. Please hold on the merge.

euvre:
Review follow-up — every comment was verified against the Python implementation and either applied (pushed as 549a69e) or answered below.

**@wangq8's review**
1. *Storage-failure diverging path*: already handled in this PR — reacquire failure logs via `slog.ErrorContext` and emits `_ERROR` instead of silently reusing shredded text. No further change.
2. *Doc comment near `ChunkIDForChunk`*: **applied** — it now cites both Python id sites in `rag/svr/task_executor.py`: `build_chunks`→`upload_to_minio` hashes `content_with_weight` directly (~L407), while the `run_dataflow` pipeline hashes `ck["text"]` (~L893) and only afterwards renames `text` to `content_with_weight` (~L913). Both hash the same chunk content.
3. *Chunk-id stability / `%x` zero-padding*: **verified — no change needed**. Python's `xxhash.xxh64(...).hexdigest()` is zero-padded to 16 hex chars (checked numerically: digest `313107618337057715` → `"045861b84f72ebb3"`, len 16; unpadded it would be `"45861b84f72ebb3"`). Go's `%016x` therefore matches Python byte-for-byte.
4. *Focused unit tests*: already added in this PR (5 txt-reacquire / chunk-id tests); extended this round with charset tests (see below).
5. *Residual edge — non-UTF-8 sources*: **fixed**, see the coderabbit item.

**@xugangqiang (why xxh64 instead of SHA-1)**: xxh64 keeps the Go ids byte-identical to the Python implementation, which computes `xxhash.xxh64((content + doc_id)).hexdigest()` at both id sites in `rag/svr/task_executor.py`. Any other hash would assign different ids to the same chunks across the Go/Python boundary, breaking dedup/idempotency for documents already indexed by the Python path. If you've observed a concrete mismatch, please share a repro and we'll dig in.

**coderabbit (major): non-UTF-8 txt/md collapses into U+FFFD**: **applied**. `reacquireQARawText` now decodes via a new `decodeQARawText`: UTF-16LE/BE honored with a leading BOM (Python's `utf_16` codec likewise requires one), valid UTF-8 used as-is, otherwise GB18030 — a superset of GBK/GB2312, the encodings chardet realistically reports for CJK text — with a raw-bytes fallback (Python's `errors="ignore"` is equally lossy). Deterministic decoding was chosen over adding a chardet dependency, consistent with the existing email parser. New regression tests: `TestQAChunker_TxtReacquireDecodesGBK`, `TestQAChunker_TxtReacquireDecodesUTF16/{le,be}`.

**coderabbit (nitpick): `Invoke`→`invoke` pass-through**: **applied** — collapsed into a single `Invoke`.

Verification: `bash build.sh --test ./internal/ingestion/component/chunker/... ./internal/common/...` — all pass, including the 3 new charset tests. Note: no shared-service end-to-end run was performed; validation is unit tests plus parity comparison against the Python implementation only.

xugangqiang:
## Design review: could we address the root cause instead of the symptom?

I verified the fix works — the id-collapse is real, `ChunkIDForChunk` matches Python's `build_chunks` flow at `task_executor.py:407`, and all affected package tests pass. I have a few questions about the approach, mostly around whether a cleaner root-cause fix is feasible. Not blocking — just curious about the maintainers' thinking.

### 1. QAChunker is the only chunker that emits `content_with_weight` without `text` — is that necessary?

Every other Go chunker sets `ChunkDoc.Text` only (`token.go:361` etc.), so the old `ChunkID(docID, ck["text"])` worked fine for them. QAChunker is the sole outlier: it copies Python `qa.py:beAdoc` and writes `content_with_weight` only (`qa.go:133`). That single shape difference is what collapsed the ids. Two observations:

- The codebase already has a consumer-side fallback for `content_with_weight`-only chunks — `tokenizer.go:632-648` (`normalizeChunkTextFallback`) backfills `text` from `content_with_weight`. `ChunkIDForChunk` extends that same convention to id derivation, so it's internally consistent. But it also means the special shape is carried everywhere rather than normalized at the source — is there a reason not to normalize it where it's produced?
- The parity harness doesn't constrain QA's intermediate shape at all — `testdata/parity/cases/` has only `token__*`/`title__*` cases, no `qa__*`. So nothing would break if QA emitted `text` instead.

Would it be cleaner to have QAChunker emit the formatted `"Question: ...\tAnswer: ..."` string as `text`? Then:
- the decorator could go back to `ChunkID(docID, text)`; `ChunkIDForChunk`, the persist fallback change, the `known_diffs.json` reason edit, and the new id tests could all be dropped;
- `RenameTextToContentWithWeight` (`process.go:31`) already copies `text` → `content_with_weight` and deletes `text`, so the persisted chunk would be byte-identical;
- the `content_with_weight`-only special case would disappear from every consumer.

The cost is that QAChunker's intermediate output would diverge from Python `qa.py` — but that divergence is currently unmeasured (no QA golden cases). Isn't alignment a means, not the goal here?

### 2. Is the chunker's DB dependency actually needed in production?

The `doc_id → (bucket, path)` resolution in `ResolveDocumentStorage` (`document_storage.go:86`) is the only place the chunker touches the DB. In production the File component already echoes `bucket`/`path` (`file.go:143-149`) and the Parser forwards them (`parser.go:564-570`), so the chunker goes down the `FetchBinary` branch and the DB path looks effectively dead at runtime. A few options, in increasing order of cleanliness — is any of these viable?

1. Drop the `doc_id` branch and rely on the File/Parser contract to always supply `bucket`/`path` — the chunker becomes a pure function with no DB coupling.
2. Keep the branch but inject a resolver closure at registration time (similar to how `ResolveDocumentStorageOverride` is used in tests) instead of reaching into `dao` from inside the component.

### 3. Re-reading the original file — is this a workaround for the boundary dropping raw bytes?

Python never re-reads storage: `task_executor.py:363` passes `binary` straight into `qa.py.chunk(...)`. The Go chunker input contract (`ChunkerInputs`) carries only the parsed payload + metadata, but the Parser has already destroyed the Q/A pairing (TextParser splits on `\n!?;。；！？` and trims the leading tab; the markdown parser block-splits). So the chunker re-fetches the source bytes to reconstruct what Python passes as an argument — which also inherits the storage round-trip, the DB fallback from point 2, and the missing charset detection (`string(data)` with only a UTF-8 BOM strip).

Would it be cleaner to pass the binary through the component boundary (same process, `[]byte` in the inputs map — no serialization cost), matching Python? I realize on-demand re-fetch is the established pattern here (the PDF crop path), so it's pragmatic — but is it the design we want long term?

### Summary

| | Current PR | Alternative worth considering |
|---|---|---|
| QA chunk shape | `content_with_weight`-only (Python-copy) | emit `text`; persist rename derives `content_with_weight` |
| Id derivation | new `ChunkIDForChunk` fallback | plain `ChunkID(docID, text)`, no new helper |
| Raw source | chunker re-fetches from storage + `doc_id`→DB fallback | File/Parser pass `binary` (like Python), or at least guarantee `bucket`/`path` |
| Encoding | `string(data)`, no charset detection | passthrough removes the storage/DB coupling, but neither approach adds charset detection |

If the goal is a cleaner architecture, would the combination of **QAChunker emits `text` uniformly + pass `binary` through the component boundary** be worth it? It would let `ChunkIDForChunk`, the re-acquisition logic, the `known_diffs.json` edit, and the storage-fallback tests all be removed, with the regression pinned by a much smaller test ("QA txt/md produces N distinct-id text chunks with correct content"). Or, if keeping the current approach (smaller diff), could we at least drop the `doc_id`→DB branch (or inject the resolver) and document why the re-read exists?

euvre:
Thanks for the detailed reviews — went through every item against the current code. No code changes this round; the design questions are answered below with code references, and the touched-package tests still pass at 549a69e.

**@xugangqiang — design review**

**1. Emit `text` instead of `content_with_weight`**

Fair observation that either shape is internally consistent — `RenameTextToContentWithWeight` (`internal/ingestion/task/indexdoc/process.go:29`) would make the persisted document identical either way. We kept `content_with_weight` for one reason: the producer-shape anchor is Python `qa.py`, and there the shape is not ambiguous — `beAdoc` writes only `content_with_weight` (`rag/app/qa.py:249`, no `text` key), and the classic executor's id site hashes exactly that key (`task_executor.py:407`: `xxhash.xxh64((chunk["content_with_weight"] + str(d["doc_id"]))...)`). Python's *other* path (`run_dataflow`, `task_executor.py` ~L893) hashes `text` and renames afterwards — which is exactly the asymmetry `ChunkIDForChunk`'s doc comment cites. Since this pipeline mirrors `build_chunks`/`upload_to_minio`, keeping the chunker's output shape identical to `qa.py` and making the id derivation content-aware is the faithful port; emitting `text` would make the Go chunker observably diverge from the Python chunker it ports (intermediate payloads, canvas debug output, future `qa__*` parity goldens) for zero difference in the persisted result. Also, the content_with_weight-preferred convention isn't QA-specific here — `common.go:139` and `group.go:440` already prefer it — so `ChunkIDForChunk` extends an existing convention rather than introducing a special case. You're right that the divergence is currently *unmeasured* (`testdata/parity/cases/` has only `token__*`/`title__*` cases) — we'd rather close that gap by adding `qa__*` parity cases in a follow-up than relax the shape to match the harness's current coverage.

**2. The `doc_id` → DB branch**

Checked against the code — the premise doesn't hold for production. Persist runs seed only `doc_id` (`pipeline_executor.go:657-659`); the File component echoes `bucket`/`path` only when they were supplied as explicit inputs (`file.go:145-150`), and the Parser forwards only what it received (`parser.go:566-571`). So in the standard pipeline `upstream.Bucket`/`Path` arrive empty and the `doc_id` branch *is* the live path — dropping it would break QA txt/md ingestion outright (the PDF-crop path resolves storage exactly the same way, `pdfcrop_cgo.go:40`). On the injection concern: the component doesn't reach into `dao` directly — it goes through the shared `component.ResolveDocumentStorage` helper (`document_storage.go:86`), the same resolution the Parser itself uses (`parser.go:658-663`), and `ResolveDocumentStorageOverride` (`document_storage.go:41`) is already the injectable resolver seam you describe, used by the tests. So we won't drop the branch (it would break production), and the suggested seam already exists. Keeping as-is.

**3. Pass the binary through the boundary**

The current contract is deliberate and documented at the schema level: `schema/chunker.go:58-67` states the design as re-resolving the source from storage "instead of carrying the binary across the component boundary", and `ChunkerFromUpstream` mirrors the Python `rag/flow` Pydantic wire shape, which carries no binary either. Passing `binary` File→Parser→Chunker would change the shared component contract for every component (and what gets serialized into checkpoints) to fix one re-read — an architecture-level change, out of scope for this bugfix. Happy to explore it separately if the maintainers want to revisit the boundary contract.

On the earlier "hold the merge": understood — your follow-up marked these non-blocking; hope the above (plus the charset decoding added in 549a69e) addresses the concerns. Happy to keep discussing.

**@wangq8** — thanks for the review! Status of the five items: (1) fair point — the PR bundles two behavior changes; for the record: chunk-id derivation from `content_with_weight` (the collapse fix) *and* txt/md raw source re-acquisition + charset decoding (the `qa.py:101-196` port). (2) Done — `ChunkIDForChunk`'s doc comment now cites both Python id sites. (3) Verified numerically last round — identical id space. (4) Noted — mirrors Python `qa.py`'s extension dispatch. (5) Superseded by 549a69e, which now fully decodes UTF-16 LE/BE with BOM, UTF-8 BOM, GB18030, then UTF-8-lossy (`decodeQARawText`), matching Python `find_codec` behavior — thanks for pushing on that.

**coderabbit (latest review)** — the "remove Python-parity wording from comments" item is not applied: `mirrors Python …` references are the established documentation convention across this Go port (`format.go:180`, `file.go:127/183`, `schema/chunker.go:30-35`, `tokenizer.go:572`, …) and are precisely what reviewers use to verify behavior against the reference implementation. The repo guideline against compatibility wording targets deprecated/legacy shims and dual-track migration notes, not parity documentation.

**Verification this round**: no code changes; re-ran `bash build.sh --test ./internal/ingestion/component/chunker/... ./internal/common/... ./internal/ingestion/task/indexdoc/...` at 549a69e — all packages `ok`. No shared-service end-to-end verification from our side; based on unit tests + Python comparison.

euvre:
@xugangqiang will handle it.

PR #18564, opened 2026-08-20
Fix: kimi k3 temperature

Lynn-Inf:
### Summary

Fix:
- Check model_obj exist before update model
- Pass llm_setting to chat_mdl in search completion api


wangq8 (commented):
Thanks for the fix. Both changes look correct and the tests are solid.

**`provider_api_service.py`** — the `if not model_obj` guard is a good defensive addition and prevents an `AttributeError` on `model_obj.status` for a not-yet-added model.

**`dialog_service.py`** — replacing the hardcoded `{"temperature": 0.1}` with `resolve_llm_setting(search_config.get("llm_setting"))` is the right call: it honors the per-search `llm_setting` (with `_enabled` flags) instead of a fixed temperature, and `resolve_llm_setting` already falls back to `LLM_SETTING_DEFAULTS` when `llm_setting` is absent. Tests cover both the forwarding path and the disabled-flag fallback.

A few points to consider (none blocking):

1. **Behavior change scope.** Previously the AI summary only passed `temperature=0.1`. Now it passes the full resolved config (`temperature`, `top_p`, `frequency_penalty`, `presence_penalty`), including the defaults (`top_p=0.3`, `frequency_penalty=0.7`, `presence_penalty=0.4`) when no `llm_setting` is present. This is more correct and likely the intent, but it does change summary generation for every tenant — worth a conscious sign-off.

2. **Layering.** `api/db/services/dialog_service.py` now imports from `api.apps.restful_apis._generation_params`. A DB/service layer importing from the REST-API layer is a mild layering inversion. `resolve_llm_setting`/`LLM_SETTING_DEFAULTS` are generic helpers — consider relocating them to a lower-level shared module (e.g. under `common/`) so both layers can import cleanly. Functional, but worth cleaning up.

3. **`del gen_conf["parameter"]` is a bit magical.** `resolve_llm_setting` carries over unknown keys (including a stray `parameter`), so the explicit delete is needed, but a one-line comment explaining that `parameter` is a transient frontend field that can leak into `llm_setting` would help future maintainers understand *why*.

Otherwise LGTM.

PR #18567, opened 2026-08-20
fix(web): clicking anywhere in a switch setting row toggles the switch

euvre:
## Problem

In settings panels such as Chat setting → Advanced settings (Show quote,
Keyword analysis, Text to speech, Multi-turn optimization), clicking
anywhere on the label row of a switch - even the empty space far to the
right of the text - toggles the switch.

### Root cause

`SwitchFormField` (`web/src/components/switch-fom-field.tsx`) renders the
`FormItem` as a flex column. As a flex item the `FormLabel` (`<label>`) is
blockified and stretched to the full container width
(`align-items: stretch`). The label remains associated with the switch
through `htmlFor`/`id`, and the browser natively forwards clicks on a
label to its labeled control. The result: the entire row acts as an
invisible switch hit area, so stray clicks in the row toggle the setting.

## Fix

One-line change in the shared component: constrain the label to its
content width with `w-fit`, so the clickable label area covers only the
text (and tooltip icon), not the whole row.

```tsx
<FormLabel className="w-fit" tooltip={tooltip}>
  {label}
</FormLabel>
```

Behavior after the fix:

- Clicking the empty area of the row no longer toggles the switch.
- Clicking the label text still toggles it (standard label semantics and
  accessibility association are preserved).
- Clicking the switch itself is unchanged.
- The horizontal variant (`vertical={false}`, e.g. compilation
  templates) is unaffected: in a row layout the label width is already
  content-based.

## Affected surfaces

All `SwitchFormField` consumers are fixed at once:

- `web/src/pages/next-chats/chat/app-settings/chat-prompt-engine.tsx`
  (Show quote / Keyword analysis / Text to speech / Multi-turn
  optimization)
- `web/src/components/embed-dialog/index.tsx` (Embed into webpage)
- `web/src/pages/user-setting/compilation-templates/edit-next/...`
- `web/src/pages/agent/form/yahoo-finance-form/index.tsx`

## Verification

- Browser (vite dev server, chat settings → Advanced settings):
  - the Show quote label box is now ~94px wide inside the 400px row and
    carries the `w-fit` class;
  - a click dispatched at the empty row area right of the text hits the
    FormItem div, not the label, and the switch state does not change;
  - a real click on the label text still toggles the switch
    (checked → unchecked).
- `npx oxlint src/components/switch-fom-field.tsx`: 0 warnings/errors.
- `npx tsc --noEmit`: no errors in the touched file (the repo has
  pre-existing errors in unrelated files).

### DIFF
--- a/web/src/components/__tests__/switch-fom-field.test.tsx
+++ b/web/src/components/__tests__/switch-fom-field.test.tsx
@@ -1,0 +1,44 @@
+import { Form } from '@/components/ui/form';
+import { TooltipProvider } from '@/components/ui/tooltip';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { useForm } from 'react-hook-form';
+import { SwitchFormField } from '../switch-fom-field';
+
+function Harness() {
+  const form = useForm({ defaultValues: { flag: false } });
+  return (
+    <TooltipProvider>
+      <Form {...form}>
+        <SwitchFormField
+          name="flag"
+          label="Enable feature"
+          tooltip="Explains the feature"
+        />
+      </Form>
+    </TooltipProvider>
+  );
+}
+
+describe('SwitchFormField', () => {
+  it('toggles the switch when the label text is clicked', () => {
+    render(<Harness />);
+    const switchEl = screen.getByRole('switch');
+    expect(switchEl).toHaveAttribute('data-state', 'unchecked');
+
+    fireEvent.click(screen.getByText('Enable feature'));
+
+    expect(switchEl).toHaveAttribute('data-state', 'checked');
+  });
+
+  it('does not toggle the switch when the tooltip icon is clicked', () => {
+    const { container } = render(<Harness />);
+    const switchEl = screen.getByRole('switch');
+    expect(switchEl).toHaveAttribute('data-state', 'unchecked');
+
+    const icon = container.querySelector('svg.lucide');
+    expect(icon).not.toBeNull();
+    fireEvent.click(icon!);
+
+    expect(switchEl).toHaveAttribute('data-state', 'unchecked');
+  });
+});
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for this clean fix. The root-cause analysis is accurate: as a flex item the `<label>` is blockified/stretched to the full row width, and the native label→control forwarding turns the whole row into a hit area. Adding `w-fit` correctly constrains the label to its content width while preserving `htmlFor`/`id` semantics and the tooltip.

Verified against `web/src/components/ui/form.tsx`: `FormLabel` merges `className` via `cn(...)` and forwards `htmlFor`, so `w-fit` lands on the `<label>` as expected, and the `vertical={false}` row layout is unaffected.

One minor nit (non-blocking, pre-existing behavior): the tooltip icon is rendered *inside* the `<label>`, so clicking the help icon still toggles the switch. Not introduced by this change and arguably out of scope, but if you ever touch this component again, moving the tooltip out of the label's click target would be a nice UX improvement.

LGTM otherwise.

dcc123456 (approved):

euvre:
Thanks for the reviews!

@wangq8 — good eye on the tooltip icon, but I verified that clicking it does not actually toggle the switch. FormLabel renders tooltips via FormTooltip (web/src/components/ui/tooltip.tsx), whose trigger span already calls e.preventDefault() on click (the handler annotated 'Prevent clicking the tooltip from triggering form save'). The label→control forwarding is the default action of that same click event, so canceling it keeps the help icon out of the label's effective click target. No runtime change was needed.

To lock this in, I added a focused test, web/src/components/__tests__/switch-fom-field.test.tsx, covering both halves of the behavior:
- clicking the label text toggles the switch (the htmlFor/id semantics this PR preserves);
- clicking the tooltip icon does not toggle the switch.

Verification: npx jest on the new test file → 2/2 pass; npx oxlint on the new file → clean; npm run type-check → no new errors (the 219 reported errors are pre-existing elsewhere and identical with and without this change). Note: no shared-service end-to-end run was performed here — verification is based on unit tests plus code-level event-flow analysis.

Thanks @dcc123456 for the approval!

PR #18570, opened 2026-08-20
fix(web): keep chat auto-scroll following streamed answers

dcc123456:
### Summary

fix(web): keep chat auto-scroll following streamed answers

Auto-follow latched off in two cases: a transcript shorter than the viewport reported "not at bottom", and content that grew between scrollToBottom and the scroll event left a gap with an unchanged scrollTop. Neither fires another scroll event, so the flag never re-armed. Only disarm on a decreasing scrollTop, which is the sole reliable signal of a user gesture.

### DIFF
--- a/web/src/hooks/logic-hooks.ts
+++ b/web/src/hooks/logic-hooks.ts
@@ -421,14 +421,23 @@ export const useScrollToBottom = (
   const ref = useRef<HTMLDivElement>(null);
   const [isAtBottom, setIsAtBottom] = useState(true);
   const isAtBottomRef = useRef(true);
+  // `null` means "no baseline yet", so the very first measurement has to fall
+  // back to the distance check instead of guessing a scroll direction.
+  const lastScrollTopRef = useRef<number | null>(null);
 
   useEffect(() => {
     isAtBottomRef.current = isAtBottom;
   }, [isAtBottom]);
 
-  const checkIfUserAtBottom = useCallback(() => {
+  const checkIfNearBottom = useCallback(() => {
     if (!containerRef?.current) return true;
     const { scrollTop, scrollHeight, clientHeight } = containerRef.current;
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks — the root-cause analysis is accurate and the fix handles both failure modes cleanly.

**What I verified**

- The `scrollHeight <= clientHeight` short-circuit is correct: content shorter than the viewport is trivially "at the bottom", and the previous code returning `false` there would latch auto-follow off on a short transcript with no later `scroll` event to re-arm it.
- The `lastScrollTopRef` + decreasing-`scrollTop`-only disarm logic is sound. Only a user gesture (wheel/drag/keys/touch) shrinks `scrollTop`; keeping the previous state when `scrollTop` is unchanged-but-not-near-bottom correctly handles content that grew after `scrollToBottom` but before the event fired. The `previousScrollTop === null` guard makes the very first measurement fall back to the distance check rather than guessing a direction.
- I traced the existing `web/src/hooks/__tests__/logic-hooks-use-scroll-to-bottom.test.tsx` cases against the new logic: "at bottom" (nearBottom→true), "not at bottom" (first call, `previousScrollTop === null` → false), and "should NOT scroll when isAtBottom false" (scrollTop unchanged, falls to `isAtBottomRef.current` = false) all still hold. No regression in the current suite.

**Suggestion (non-blocking)**

This hook already has a dedicated test file (`web/src/hooks/__tests__/logic-hooks-use-scroll-to-bottom.test.tsx`), but this PR doesn't extend it. Worth adding two cases that lock in the new behavior:

1. Short content (`scrollHeight <= clientHeight`) reports at-bottom even with `scrollTop = 0`.
2. Content-grew-with-unchanged-`scrollTop` (not near bottom, `scrollTop >= previousScrollTop`) preserves the prior `isAtBottom` value instead of flipping to false.

Otherwise looks good.

PR #18571, opened 2026-08-20
fix(models): infer model types on catalog miss so Qwen VLMs show LLM+VLM tags

euvre:
## Problem

Qwen vision-language models (e.g. `qwen-vl-plus`, `qwen3-vl-flash`) listed under an Aliyun (OpenAI-compatible) provider showed only the **LLM** tag, while the Python version shows **LLM + VLM**.

## Root cause

- Go `ParseListModel` (`internal/entity/models/base_model.go`) enriches remote `/models` entries only from the static catalog (`conf/all_models.json`). On a catalog miss — non-exact names/aliases like `qwen-vl-plus`, `qwen3-vl-flash`, or dated variants such as `qwen-vl-plus-2025-01-25` — the entry carried **empty `ModelTypes`**.
- The provider-list handler (`internal/handler/providers.go`) treats remote entries as authoritative: on a name match the remote entry **wholesale replaces** the static one, erasing the static entry's types.
- The instance save path persists whatever types the list carries, and the frontend defaults empty types to a chat-only badge — so vision models rendered as LLM-only.
- Python's equivalent (`rag/llm/model_meta.py`, `OpenAIAPICompatible._format_model_list`) always derives types from name hints and defaults to `["chat"]`, which is why the Python UI shows LLM + VLM.

## Fix

`ParseListModel` now falls back to `InferModelTypes(name)` when the catalog cannot type an entry, mirroring Python semantics:

- vision names (`qwen-vl…`, `gpt-4o…`, `glm-4v…`, …) → `chat` + `vision`
- embedding/rerank names → their own types
- unknown names → `chat` (never type-less again)

Catalog hits keep catalog types (precedence preserved). Also removed a duplicated `Dimensions` assignment in the same function.

## Tests

- `internal/entity/models/base_model_test.go` (new): catalog hits keep catalog types (`qwen3-vl-plus` → `chat+vision+ocr`), catalog misses infer types (`qwen3-vl-flash-2099-01-01`, `gpt-4o-2099-01-01` → `chat+vision`; `text-embedding-v4` → `embedding`; `qwen3-reranker-2099` → `rerank`; unknown → `chat`), blank IDs still skipped, catalog metadata still flows.
- `internal/entity/models/gitee_test.go`: updated the assertion that locked in the old type-less catalog-miss behavior.

## Verification

- `bash build.sh --test ./internal/entity/models/... ./internal/handler/... ./internal/service/...` — all pass.
- E2E (live UI, rebuilt services):
  - Live **List models** fetch against Tongyi-Qianwen returns `qwen3-vl-plus` with `model_types: ["vision","chat"]` (previously empty) — server-side fix path confirmed.
  - Against a stub OpenAI-compatible endpoint (`/v1/models`) added as a VLLM instance, the model settings page rendered: `qwen-vl-plus` → **LLM + VLM**, `qwen3-vl-flash-2099-01-01` → **LLM + VLM** (catalog-miss dated variant, proves the `InferModelTypes` fallback), `my-plain-model` → **LLM** only, `bge-m3-stub` → **Embedding**.
  - Screenshot: `vlm-tags-e2e-final.png` (stub instance and stub server removed after verification).

### DIFF
--- a/internal/entity/models/base_model.go
+++ b/internal/entity/models/base_model.go
@@ -383,6 +383,11 @@ func ParseSSEStreamTolerant[T any](r io.Reader, onEvent func(event T) error) (do
 
 // ParseListModel Parse model list. Empty/whitespace IDs are skipped so
 // upstream typos do not surface as blank entries in the UI.
+//
+// Entries the catalog cannot type fall back to name-based inference
+// (InferModelTypes), mirroring Python's
+// OpenAIAPICompatible._format_model_list (rag/llm/model_meta.py) so remote
+// entries never surface type-less.
 func ParseListModel(modelList ModelList) []ListModelResponse {
 	var models []ListModelResponse
 	pm := GetProviderManager()
@@ -405,7 +410,18 @@ func ParseListModel(modelList ModelList) []ListModelResponse {
 			modelResponse.MaxOutput = modelEntity.MaxOutput
 			modelResponse.ModelTypes = modelEntity.ModelTypes
 			modelResponse.Thinking = modelEntity.Thinking
-			modelResponse.Dimensions = modelEntity.Dimensions
+		}
+
+		// The provider-list merge treats remote entries as authoritative
+		// (internal/handler/providers.go) and the instance save path
+		// persists whatever types this list carries, so a catalog miss
+		// must not leave ModelTypes empty — the UI renders type-less
+		// models with an LLM-only badge. Infer types from the model name
+		// (vision models like qwen-vl-plus keep their VLM tag even before
+		// the catalog knows them); InferModelTypes always returns at
+		// least ["chat"].
+		if len(modelResponse.ModelTypes) == 0 {
+			modelResponse.ModelTypes = InferModelTypes(modelName)
 		}
 
 		models = append(models, modelResponse)
--- a/internal/entity/models/base_model_test.go
+++ b/internal/entity/models/base_model_test.go
@@ -1,0 +1,83 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Thanks for this fix — the root-cause analysis is spot on and the change correctly addresses the type-less remote-entry bug.

**What I verified**

- `ParseListModel` previously only populated `ModelTypes` from the static catalog (`GetModelByNameOrAlias`), so catalog misses surfaced with empty types. Since `internal/handler/providers.go` (ListModels) treats remote entries as authoritative on name conflicts and the instance save path persists whatever the list carries, empty types rendered as LLM-only. The `InferModelTypes` fallback closes that gap and matches the intent of Python's `OpenAIAPICompatible._format_model_list`.
- The fallback is placed correctly after the `modelEntity != nil` block, so catalog hits keep catalog types (precedence preserved) and only type-less entries get inferred. `InferModelTypes` always returns at least `["chat"]`, so nothing surfaces type-less.
- Removing the duplicated `modelResponse.Dimensions = modelEntity.Dimensions` line is a clean incidental cleanup.
- Tests are thorough and cover the important branches (catalog hit precedence, dated-variant misses, embedding/rerank inference, unknown→chat, blank-ID skip, catalog metadata flow). The `gitee_test.go` assertion update is consistent with the new invariant.

**Minor observations (non-blocking)**

1. Parity claim vs. `InferModelTypes`: the Go `InferModelTypes` includes an `ocrHints` branch (`return []string{"ocr"}`) that Python's `OpenAIAPICompatible._infer_model_types` does not have. This is pre-existing behavior, not introduced here, but the PR comment says it "mirrors Python's semantics" — worth a quick sanity check that the extra `ocr` inference is intended (it's a legitimate RAGFlow type, just not symmetric with Python).

2. Scope of the fallback: because the fallback sits outside the `modelEntity != nil` block, it also applies to catalog hits that carry empty `ModelTypes`. That's consistent with the "never type-less again" goal, just calling it out so it's an intentional behavior change rather than a side effect.

Nothing blocking — this looks good to merge from a correctness/regression standpoint.

Lynn-Inf (approved):
LGTM

PR #18576, opened 2026-08-20
fix(connector): honor configured S3-compatible bucket type

yijiaquan:
### Summary

Honor `config["bucket_type"]` when constructing blob-storage connectors.

Previously, `build_connector_for_source()` always passed the source type to `BlobStorageConnector`. An S3 data source configured with `bucket_type: s3_compatible` was therefore constructed as Amazon S3, causing it to use the wrong credential validation and client setup.

This change uses the configured bucket type when present and falls back to the source type when it is missing or empty.

### Verification

- Added unit coverage for a configured `s3_compatible` bucket type.
- Added unit coverage for missing, `None`, and empty bucket-type fallbacks.
- `python3 -m py_compile common/data_source/__init__.py test/unit_test/common/test_data_source_factory.py`
- `git diff --check`

### DIFF
--- a/common/data_source/__init__.py
+++ b/common/data_source/__init__.py
@@ -111,7 +111,7 @@ def build_connector_for_source(source: str, config: dict[str, Any]) -> Any:
     if connector_cls is None:
         raise ConnectorValidationError(f"Unsupported data source type: {source}")
     if connector_cls is BlobStorageConnector:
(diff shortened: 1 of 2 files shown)


JinHai-CN:
@huang-aoqin Would you please review the issue?

mkaaad (approved):
LGTM

PR #18578, opened 2026-08-20
fix(dialog): SQL retrieval targets KB owner tenant, not chat tenant

### LINKED ISSUE #18514: [Bug]: SQL retrieval uses the chat tenant instead of the shared dataset owner tenant

Harsh23Kashyap:
Fixes #18514.

## Summary

When a chat references a team-shared Table dataset owned by another tenant, `async_chat` runs SQL retrieval against the chat user's `ragflow_<chat_tenant>` index instead of the dataset owner's `ragflow_<kb_owner_tenant>` index. Chunks are stored in the owner's index; the chat tenant's index may not even exist. Result: SQL retrieval returns empty for a chat that has data.

## Fix

Derive the SQL target index from `kbs[0].tenant_id`, paralleling the vector retrieval path's `tenant_ids = list(set([kb.tenant_id for kb in kbs]))`:

```python
sql_tenant_id = kbs[0].tenant_id if kbs else dialog.tenant_id
ans = await use_sql(questions[-1], field_map, sql_tenant_id, chat_mdl, prompt_config.get("quote", True), dialog.kb_ids, doc_ids=scoped_doc_ids)
```

`kbs` is already in scope (loaded by `get_models(dialog, ...)` two hundred lines above). The vector retrieval path's `tenant_ids` derivation handles the multi-KB case; the SQL path uses the parallel single-tenant derivation because `validate_dataset_embedding_models` (called inside `get_models`) already asserts all KBs share one embedding model, so multi-tenant KBs in a single dialog is not a supported configuration. The `if kbs` guard preserves `dialog.tenant_id` as a last-resort fallback for the edge case where `get_models` returns zero KBs but a stale `field_map` slips through.

## Test

Four regression tests in `test/unit_test/api/db/services/test_dialog_service_sql_tenant.py` drive the real `dialog_service.async_chat` (with heavy dependencies stubbed) and assert that the `use_sql` call receives the KB owner's tenant_id, not the dialog's:

- the main test reproduces issue #18514's exact scenario: a chat owned by tenant B references a KB owned by tenant A. Pre-fix, the assertion fails with `SQL retrieval must target the KB owner tenant ('tenant-A-uuid'); got 'tenant-B-uuid'`. With the fix, it passes.
- single-KB case pins the `kbs[0].tenant_id` path against the multi-KB `tenant_ids` vector path
- empty-kbs case asserts the `dialog.tenant_id` fallback
- AST guard pins the `if field_map:` gate so the SQL path is not accidentally widened

## Risks

- The fix changes a tenant_id in a single async call. The pre-fix behaviour is preserved as the last-resort fallback (`kbs` empty); the common path (one or more KBs loaded) now uses the KB's tenant.
- `use_sql` is the only call site. The `agentic_rag.py` and `harness/tools/search.py` paths pass an explicitly-provided `tenant_id` from their own resolution and are unaffected by this fix.

## Related

The `KnowledgebaseService.get_field_map` call two lines above already operates per-KB (joining the loaded KBs' field_map dictionaries into a single map). The fix is the SQL tenant_id analog of that per-KB resolution.

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -670,7 +670,16 @@ async def async_chat(dialog, messages, stream=True, **kwargs):
     # try to use sql if field mapping is good to go
     if field_map:
         logging.debug("Use SQL to retrieval:{}".format(questions[-1]))
-        ans = await use_sql(questions[-1], field_map, dialog.tenant_id, chat_mdl, prompt_config.get("quote", True), dialog.kb_ids, doc_ids=scoped_doc_ids)
+        # Derive the SQL target index from the dataset owner tenant, not the
+        # chat tenant. dialog.tenant_id is the chat's user; the chunks live
+        # in `ragflow_<kb_owner_tenant_id>`. A user authorised on a team-shared
+        # KB owned by another tenant has dialog.tenant_id != kb.tenant_id;
+        # the vector retrieval path already uses kb.tenant_id (see
+        # `tenant_ids = list(set([kb.tenant_id for kb in kbs]))` below), the
+        # SQL path was passing the wrong tenant and querying a non-existent
+        # index. See issue #18514.
+        sql_tenant_id = kbs[0].tenant_id if kbs else dialog.tenant_id
+        ans = await use_sql(questions[-1], field_map, sql_tenant_id, chat_mdl, prompt_config.get("quote", True), dialog.kb_ids, doc_ids=scoped_doc_ids)
         # For aggregate queries (COUNT, SUM, etc.), chunks may be empty but answer is still valid
         if ans and (ans.get("reference", {}).get("chunks") or ans.get("answer")):
             if include_reference_metadata and ans.get("reference", {}).get("chunks"):
--- a/test/unit_test/api/db/services/test_dialog_service_sql_tenant.py
+++ b/test/unit_test/api/db/services/test_dialog_service_sql_tenant.py
@@ -1,0 +1,315 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for issue #18514.
+
+When a chat references a team-shared Table dataset owned by another tenant,
+``async_chat`` runs SQL retrieval against the chat tenant's
+``ragflow_<chat_tenant>`` index instead of the dataset owner's
+``ragflow_<kb_owner_tenant>`` index. Chunks are stored in the owner's index
+-- the chat tenant's index may not even exist -- so SQL retrieval returns
+empty results for a chat that has data.
+
+The fix derives the SQL target index from ``kbs[0].tenant_id`` (parallel to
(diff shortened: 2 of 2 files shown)


wangq8:
Dup of #18518, hence close it.

PR #18579, opened 2026-08-20
fix(agent): surface cancellation terminal state

T1mn:
### Summary

Follow up on #18316 by preserving cancellation as a terminal state in the Go Agent OpenAI-compatible response path. Previously, a cancelled run closed the event channel without an event, so the adapter could return an empty successful completion or finish a stream with stop.

### Design

- Emit a structured cancelled RunEvent from the Canvas Runner when the managed run returns context cancellation.
- Return HTTP 409 with an OpenAI invalid_request_error for non-streaming compatible requests.
- Return a cancelled finish reason and structured cancellation payload for streaming requests, without sending [DONE].
- Keep normal, error, and waiting-for-user event behavior unchanged.

### Tests

- Add a Canvas Runner regression test for the cancellation event and channel closure.
- Add Go handler regressions for non-streaming and streaming OpenAI-compatible cancellation responses.

### Validation

- Passed: targeted Canvas Runner test with CGO disabled.
- Passed: Linux-targeted compilation of the Canvas and handler test packages with CGO disabled.
- Passed: gofmt and git diff check.
- The full handler test execution remains unavailable on this macOS checkout because an existing sandbox file uses syscall.SysProcAttr.Pdeathsig, which is not defined by the installed Darwin Go toolchain. The repository build test is also missing the local pdf_oxide native library.


yuzhichang (approved):
LGTM

PR #18580, opened 2026-08-20
Fix: Go chat pipeline sends image blocks to text-only models (Zhipu GLM error 1210)

euvre:
## Problem

When chatting with an image attachment through the Go backend, providers whose
chat models are text-only reject the request. With Zhipu GLM (e.g. `glm-4-flash`)
the conversation fails with:

```
ERROR: API request failed ... {"error":{"code":"1210","message":"messages.content.type only allows ['text']"}}
```

Root cause: `internal/service/chat_pipeline.go` attached every uploaded file —
including images — to the outbound LLM message as `image_url` content blocks,
without checking whether the enrolled model actually accepts image input.

## Fix

Image attachment is now gated on the enrolled model type:

- **`resolveChatModelType()`** probes the enrolled model types for the dialog's
  LLM (`ModelProviderSvc.ResolveModelType`) and returns `"image2text"` for
  vision-capable models, `"chat"` otherwise.
- `entity.ModelType` is a **bitmask**: models enrolled with combined
  `["chat","vision"]` types (e.g. Zhipu `glm-4.6v-Flash`, enrolled as
  `model_type=9`) report a combined value, so vision probing uses
  `mt.Has(entity.ModelTypeImage2Text)` — not equality — or such models would be
  misclassified as text-only and silently lose image input.
- Probe failures are **conservative**: they resolve to `"chat"`, dropping image
  attachments rather than risking a provider-side rejection.
- `AsyncChat` / `AsyncChatSolo` thread the resolved `model_type` through the LLM
  model config. Text-only models drop image attachments (with a `Warn` log
  naming the `llm_id` and drop count); vision models keep receiving image
  content blocks exactly as before.
- Removed the now-unused `extractImageFiles` helper.

## Tests

- `internal/service/chat_pipeline_vision_test.go` covers `model_type`
  resolution for: chat-only models, pure image2text models, combined
  `chat|image2text` bitmask models, unknown models, and probe failures
  (conservative fallback to `"chat"`).
- `bash build.sh --test ./internal/service/...` passes (targeted
  `TestGetLLMModelConfig*` plus full package suite).

## Live verification (local stack, Zhipu provider)

- Dialog on **chat-only `glm-4-flash`** + image attachment: previously failed
  with error 1210; after the fix the request completes and the model answers
  (it reports it cannot view images, since the image is now dropped instead of
  being sent to a text-only endpoint).
- Dialog pointed at **vision `glm-4.6v-Flash`** (combined `model_type=9`) + image
  attachment: the request reaches Zhipu with image content accepted — no 1210.
  (The free-tier model intermittently rate-limits with 429/1305, which is a
  provider-side quota issue unrelated to this change.)

### DIFF
--- a/internal/service/chat_pipeline.go
+++ b/internal/service/chat_pipeline.go
@@ -316,10 +316,12 @@ func (s *ChatPipelineService) AsyncChat(
 
 		// Parse file attachments from the last message.
 		// Split text-file URLs (joined with "\n\n") and image URLs.
-		// Chat model: images → imageAttachments (multimodal conversion).
-		// Image2text model: images → imageFiles (raw URLs).
+		// Vision (image2text) model: images → imageFiles for the multimodal
+		// conversion below. Text-only chat model: images are dropped —
+		// sending image content blocks makes such providers reject the
+		// request (e.g. Zhipu GLM error 1210: messages.content.type only
+		// allows 'text').
 		var textAttachmentsList []string
-		var imageAttachments []string
 		var imageFiles []string
 		// Joined text attachments (appended to system prompt).
 		var attachments string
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Correct root-cause fix. Gating image attachment on the enrolled model type is the right approach, and the bitmask handling (`mt.Has(entity.ModelTypeImage2Text)` instead of equality) is the key detail — combined `chat|vision` enrollments (e.g. `glm-4.6v-Flash`, `model_type=9`) would otherwise be misclassified as text-only and silently lose image input. The conservative probe-failure fallback to `"chat"` is also sound. I verified `entity.ModelType.Has`, `ModelTypeImage2Text`, and `GetTenantDefaultModelRef` all exist, and `extractImageFiles` has no remaining callers after removal.

Two things worth a look:

1. **Silent image drop is invisible to the end user.** For a text-only chat model, images are now dropped with only a server-side `Warn`/`Debug` log. The user gets no indication their image was ignored — the model just answers without it (per your own verification, "it reports it cannot view images"). Consider surfacing this in the response (a note appended to the reply, or a structured warning) so users aren't confused. This may be a follow-up, but it's the most user-visible consequence of this change.

2. **Behavior flip for unknown `model_type` values.** The condition changed from `if modelType == "chat" {...} else {images}` to `if modelType == "image2text" {...} else {drop}`. For any `model_type` that is neither `"chat"` nor `"image2text"` (shouldn't occur for chat dialogs, but defensively), images now get dropped rather than sent. That's the conservative direction and consistent with your intent, just flagging the polarity change for awareness.

The new tests are comprehensive (chat-only, pure image2text, combined bitmask, tenant-default text-only vs vision, and unknown/probe-failure). LGTM once the user-facing drop notification is decided.

Lynn-Inf (approved):
LGTM

PR #18581, opened 2026-08-20
Fix: show linked knowledge base names instead of IDs when searching in the "Link to knowledge base" dialog

euvre:
## Problem

In **File Management → "Link to knowledge base" dialog**, after a file is linked to several knowledge bases, typing a search keyword (e.g. `1`) makes the badges of already-linked knowledge bases whose names do **not** contain the keyword render as raw knowledge-base IDs instead of their names.

### Root cause

The dialog's search filters the knowledge-base list **server-side** (`useFetchKnowledgeList` passes the keyword to `listDataset`). `MultiSelect` resolved each selected badge label only against the currently loaded `options`:

```tsx
const option = flatOptions.find((o) => o.value === value);
...
{option?.label ?? value}
```

Once the list is narrowed by the keyword, selected-but-non-matching knowledge bases vanish from `options`, so the lookup misses and the badge falls back to the raw value (the kb ID) with a warning icon.

## Fix

- `web/src/components/ui/multi-select.tsx`: new optional prop `getOptionLabel?: (value: string) => React.ReactNode | undefined`. Badge rendering now resolves `option?.label ?? getOptionLabel?.(value)`; the warning marker and raw-ID fallback remain only for values with no known label at all (e.g. the entity was deleted), preserving the existing deleted-KB behavior.
- `web/src/pages/files/link-to-dataset-dialog.tsx`: the dialog already caches every dataset it has seen (loaded pages + `useFetchDatasetsByIds` for selected IDs missing from the current page). It now passes that cache to `MultiSelect` via `getOptionLabel`, and the cache moves from a ref to state so badges re-render as soon as a missing name arrives.

No backend changes; the submit path (kb_id/kb_name payload and the file-list echo) is unchanged.

## Verification

Reproduced and verified in the browser against the local dev stack:

1. Linked `hlm.txt` to `qa-chunk-repro` and `book_chunk_test` (both badges showed names; saved).
2. Reopened the dialog and typed `qa` in the search box:
   - **Before:** badges showed `qa-chunk-repro | 6e9b0e0b89b94f65a0dfb3b0027dcc73` (raw ID for `book_chunk_test`).
   - **After:** badges show `qa-chunk-repro | book_chunk_test`, no warning icons; the dropdown correctly lists only the matching knowledge base.
3. Clearing the search restores the full list with badges unchanged.
4. `npx oxlint` on both touched files: 0 warnings, 0 errors. `tsc --noEmit` reports no errors in the touched files (remaining errors are pre-existing in unrelated files).

### DIFF
--- a/web/src/components/ui/__tests__/multi-select.test.tsx
+++ b/web/src/components/ui/__tests__/multi-select.test.tsx
@@ -1,0 +1,64 @@
+import { render, screen } from '@testing-library/react';
+
+import { MultiSelect } from '../multi-select';
+
+// jsdom does not provide ResizeObserver or scrollIntoView, which cmdk
+// relies on.
+beforeAll(() => {
+  global.ResizeObserver = class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  };
+  Element.prototype.scrollIntoView = () => {};
+});
+
+const noop = () => {};
+
+describe('MultiSelect badge labels', () => {
+  it('keeps the label of a selected value after its option disappears', () => {
+    const { rerender } = render(
+      <MultiSelect
+        options={[
+          { label: 'Alpha', value: 'a' },
+          { label: 'Beta', value: 'b' },
+        ]}
+        defaultValue={['a']}
+        onValueChange={noop}
+      />,
+    );
+    expect(screen.getByText('Alpha')).toBeTruthy();
+
+    // The option list is narrowed (e.g. by a server-side search) and no
+    // longer contains the selected value; the badge must not fall back to
+    // the raw value.
+    rerender(
+      <MultiSelect
+        options={[{ label: 'Beta', value: 'b' }]}
+        defaultValue={['a']}
+        onValueChange={noop}
+      />,
+    );
+    expect(screen.getByText('Alpha')).toBeTruthy();
+    expect(screen.queryByText('a')).toBeNull();
+  });
+
+  it('resolves a never-seen selected value via getOptionLabel', () => {
+    render(
+      <MultiSelect
+        options={[{ label: 'Beta', value: 'b' }]}
+        defaultValue={['x']}
+        onValueChange={noop}
+        getOptionLabel={(value) => (value === 'x' ? 'X-ray' : undefined)}
+      />,
+    );
+    expect(screen.getByText('X-ray')).toBeTruthy();
+  });
+
+  it('falls back to the raw value when no label is known', () => {
+    render(
+      <MultiSelect options={[]} defaultValue={['z']} onValueChange={noop} />,
+    );
+    expect(screen.getByText('z')).toBeTruthy();
+  });
+});
--- a/web/src/components/ui/multi-select.tsx
+++ b/web/src/components/ui/multi-select.tsx
@@ -259,6 +259,16 @@ export const MultiSelect = React.forwardRef<
       );
     }, [options]);
 
+    // Remember the label of every option the component has seen, so a badge
+    // keeps its label when the option later disappears from `options` (e.g.
+    // the list is narrowed by a server-side search).
+    const rememberedLabels = React.useRef(new Map<string, React.ReactNode>());
+    React.useEffect(() => {
+      flatOptions.forEach((option) => {
+        rememberedLabels.current.set(option.value, option.label);
+      });
+    }, [flatOptions]);
+
     const disabledValueSet = React.useMemo(() => {
       return new Set(
         flatOptions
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Clean, well-scoped fix. Resolving the badge label via a `getOptionLabel` fallback (instead of only the currently-loaded `options`) is the right mechanism, and moving the cache from a ref to state is necessary for badges to re-render when a missing name arrives. No backend change needed.

Notes (non-blocking):

1. **Extra fetch cycle on cache updates.** `missingIds` now depends on `datasetCache`, so each cache mutation can shrink `missingIds` and trigger a `useFetchDatasetsByIds` refetch for the (now-empty) remainder. The `changed` guard in the effect prevents a re-render loop, and the memo/`missingDatasets` should settle, but it's worth a quick check that this doesn't cause a redundant request storm during initial load with many selected-but-unloaded datasets.

2. **Icon lost for cached-but-not-loaded items.** `const IconComponent = option?.icon` still only resolves from `options`, so a badge whose label comes from `getOptionLabel` renders without its icon. This matches the previous fallback behavior, so it's acceptable — just flagging that `getOptionLabel` only covers the text, not the icon.

3. Minor: `getOptionLabel` returns `React.ReactNode`, which is fine here, but it's slightly more general than needed; `string | undefined` would make the `label == null` checks clearer.

LGTM.

euvre:
Thanks for the review and the LGTM, @wangq8! Addressing the three non-blocking notes:

**1. Extra fetch cycle on cache updates** — checked the data flow; no redundant request storm:
- `datasetCache` grows monotonically (the effect only adds entries, and the `changed` guard returns the previous state when nothing is new, so no re-render loop).
- `missingIds = selectedIds - loadedIds - cachedIds` therefore shrinks monotonically; each `useFetchDatasetsByIds` round either resolves at least one ID into the cache or settles with an unchanged cache. Once it reaches `[]`, the query is disabled (`enabled: sortedIds.length > 0`), so no further request fires.
- Worst case is one extra round-trip per batch of unresolvable IDs (e.g. datasets deleted server-side), not a storm. Search-keyword changes replace `list`, but any ID that leaves the page is already in the cache, so `missingIds` does not regrow.
- Verification boundary: this was verified at code level (data-flow analysis above) plus `npm run type-check`; no shared-service end-to-end run was performed from this environment.

**2. Icon lost for cached-but-not-loaded items** — acknowledged. As you noted this matches the previous fallback behavior; additionally, in this dialog the options carry no `icon` at all (`options = list.map(item => ({ label, value }))`), so nothing is lost in the only current consumer.

**3. `getOptionLabel` return type** — applied: narrowed to `(value: string) => string | undefined` in 5fe0a8e. The only call site returns `datasetCache.get(value)?.name` (`string | undefined`), and `npm run type-check` passes for the touched files.

dcc123456 (commented):
you should maintain a label memory cache inside the component

MultiSelect remembers the value → label mappings it has seen on its own

euvre:
Thanks @dcc123456 — applied in 229ec26: `MultiSelect` now maintains a label memory cache inside the component.

**What changed** (`web/src/components/ui/multi-select.tsx`)
- The component keeps an internal `value -> label` memory of every option it has rendered (fed from `flatOptions` in an effect), so it remembers the mappings it has seen on its own.
- Badge label resolution is now: current `option.label` → explicit `getOptionLabel` resolver → remembered label → raw value with the warning marker (the warning/raw-ID path is unchanged for values with no known label, e.g. a deleted KB).
- This also generically covers other consumers with the same server-side-search pattern (e.g. `TagSetItem` in dataset settings), with no per-consumer wiring.

**Why `getOptionLabel` is kept alongside the memory**
- The two cover disjoint cases. The memory can only resolve values the component has *seen* as options; initially-linked knowledge bases are often absent from every loaded page, so the dialog still resolves those by ID (`useFetchDatasetsByIds`) and passes the fetched names through `getOptionLabel`. The dialog also needs those names for the submit payload (`kb_name`), which the component-internal memory cannot serve.

**Verification**
- New `web/src/components/ui/__tests__/multi-select.test.tsx` (jest + testing-library), 3/3 passing: badge keeps its label after the option disappears (internal memory), never-seen value resolves via `getOptionLabel`, unknown value falls back to the raw value.
- `npx oxlint` on the touched files: 0 warnings/errors. `npm run type-check`: no errors in the touched files (remaining errors are pre-existing in unrelated files).
- Verification boundary: no shared-service end-to-end run from this environment; based on the unit tests above plus code-level data-flow analysis.

Also thanks again @wangq8 for the earlier review and LGTM — the three non-blocking notes were addressed in the previous round (`getOptionLabel` narrowed to `string | undefined` in 5fe0a8e).

dcc123456 (approved):

PR #18583, opened 2026-08-20
feat(agent): add You.com search tool

brainsparker:
### Summary

@JinHai-CN — the agent tool you asked for in #18478, as its own PR.

Adds `YouComSearch` to the agent tool set, modeled on the Keenable tool (#16233) rather than Querit's: You.com exposes a small parameter surface, so this needs a two-field form rather than a bespoke form suite. **+594 lines**, close to Keenable's +392 and well under Querit's +3527.

Independent of #18478. That PR wires You.com into chat retrieval; this one is the agent canvas node. Neither depends on the other and they share no lines.

### What's here

- **`agent/tools/youcom.py`** — `YouComSearch` / `YouComSearchParam`. Merges the response's `web` and `news` sections, web first, and prefers each result's extracted `snippets` over its `description` (news hits only carry the latter). `count` applies per section, so the merged list is trimmed back to `top_n`.
- **Frontend** — operator enum, icon, en/zh strings, initial values, node type, the canvas and tool-popover lists, `clear-sensitive-fields`, and the two forms (`youcom-form` and `tool-form/youcom-form`).
- **`test/unit_test/agent/tools/test_youcom_unit.py`** — 9 tests.

Parameters are deliberately minimal: `query`, an optional `freshness` (`day`/`week`/`month`/`year`), plus `top_n` and the optional key.

### Keyless by default

With no `api_key` the tool uses `https://api.you.com/v1/agents/search`; setting one switches it to `https://api.you.com/v1/search` with `X-API-Key` and lifts the rate limits. The keyless endpoint rejects an auth header, so the endpoint and headers are always resolved together — pinned by tests on both paths.

Requests carry `User-Agent: RAGFlow youdotcom-integration/infiniflow-ragflow`, the same attribution string as #18478 and the same role `X-Keenable-Title` plays for Keenable.

### On logging

Failures record only the exception type. This tool passes the query as a **URL parameter**, so a `requests` error message would otherwise contain it — the same issue CodeRabbit caught in #18478 and that #18582 cleans up in the three existing connectors. There's a test asserting neither the query nor the key reaches the log or the returned error string.

### Testing

`pytest test/unit_test/agent/tools/test_youcom_unit.py` — **9 passed**: keyless vs keyed endpoint selection, the attribution header, web+news merging with passage preference, `top_n` capping the merged sections, `freshness` forwarded only when set, blank-query short-circuit, the logging assertion, and `check()` rejecting a non-positive `top_n`.

Live-verified end to end through `_invoke` against the real API, `top_n=4`:

```
--- keyed:                    4 results, json=4   (content 783-1216 chars)
--- keyed + freshness=week:   4 results, json=4
```

`ruff check --config pyproject.toml` reports 8 notices on the new tool against 9 on `agent/tools/keenable.py`, so nothing new in kind; `ruff format` clean. Frontend: `tsc --noEmit` at **221 errors with and without this change** (verified by stashing), `prettier --check` clean on both new form files.

One environment note, same as #18478: I could not run the *existing* `test_querit_unit.py` locally — it imports `common.http_client`, which pulls in the `infinity` vector-DB SDK. This tool deliberately doesn't take that dependency (it passes an explicit timeout, as Keenable does), which is why its own tests run standalone.


web/src/pages/agent/constant/index.tsx
```diff
@@ -196,6 +196,31 @@ export enum KeenableMode {
   Realtime = 'realtime',
 }
 
+export enum YouComFreshness {
```
cike8899, line 199: The value of the RAGFlowSelect dropdown option cannot be empty.

web/src/components/operator-icon.tsx
```diff
@@ -72,6 +72,7 @@ export const SVGIconMap = {
   [Operator.BGPT]: 'bgpt',
   [Operator.SearXNG]: 'searxng',
   [Operator.KeenableSearch]: 'keenable',
+  [Operator.YouComSearch]: 'youcom',
```
cike8899, line 75: web/src/assets/svg/youcom.svg is required.

cike8899 (approved):

cike8899 (commented):
Suggestions for revising the comments @brainsparker

web/src/pages/agent/constant/index.tsx
brainsparker: Fixed in 4994442. `YouComFreshness.Any` is now `'any'` instead of `''`, so no option carries an empty value.

The same sentinel is used on the backend: `freshness` publishes `["any", "day", "week", "month", "year"]` as an enum in the tool schema, `any` means no restriction and is never forwarded to You.com, and anything outside the set is rejected before the request. `get_input_form` declares it as `options` rather than a free-text `line` for the same reason.

web/src/components/operator-icon.tsx
brainsparker: `web/src/assets/svg/youcom.svg` is already on `main` — it landed in 09a9b629 ("feat(chat): add You.com web search provider", #18478) and is imported by `web/src/components/web-search-form-field.tsx`. `SvgIcon` resolves `@/assets/svg/**/*.svg` by glob, so the `SVGIconMap` entry here picks it up with no new asset. Let me know if you would rather this operator use a different variant of the mark.

brainsparker:
Thanks for the review @cike8899 — pushed in 4994442.

**Your two comments**

- **Empty dropdown value.** `YouComFreshness.Any` is now `'any'`. The same value is used end to end: the tool schema publishes `["any", "day", "week", "month", "year"]` as an enum, `any` means no restriction and is never forwarded to You.com, and `get_input_form` declares freshness as `options` rather than a free-text `line`.
- **`youcom.svg`.** Already on `main` from 09a9b629 (#18478), imported by `web/src/components/web-search-form-field.tsx`. `SvgIcon` globs `@/assets/svg/**/*.svg`, so the `SVGIconMap` entry resolves without a new asset. Happy to swap in a different variant of the mark if you would prefer one.

**CodeRabbit's four**

- *Restrict and validate `freshness`* — covered by the enum above, with validation before the request and tests for a rejected value and for the published schema.
- *Sleep only before an actual retry* — the delay is now gated on another attempt remaining. It was measurably wrong: with `max_retries=0` and a nonzero delay a failed request slept before returning its error. Regression tests cover both the final-attempt case and the gaps between retries.
- *Localize the freshness labels* — each option now has its own key (`youComFreshnessAny` … `youComFreshnessYear`) in `en.ts` and `zh.ts`, mapped explicitly rather than interpolated so the keys stay greppable.
- *Redact API keys from normal You.com nodes* — canvas nodes carry the operator under `data.label` and the key under `data.form.api_key`; that branch now recognizes You.com, with a test covering both the node and the tool-record path.

On that last one: the other operators in `apiKeyOperators` (`TavilySearch`, `TavilyExtract`, `Google`, `KeenableSearch`, `BGPT`) have the same gap on the `data.label` path. It predates this tool, and `clear-sensitive-fields.test.ts` currently pins that behaviour in "does not change standalone graph export behavior for other tools", so I kept this change scoped to You.com rather than rewrite that test here. Glad to file a separate issue or fold it in, whichever you prefer.

Verified locally: `test/unit_test/agent/tools/test_youcom_unit.py` 13 passed, `clear-sensitive-fields.test.ts` 9 passed, `oxlint` clean, and `tsc --noEmit` reports nothing in the touched files.

PR #18586, opened 2026-08-20
fix(file): reject folder names containing "/" in create/rename

euvre:
### Summary

File Management allowed creating or renaming folders with a name containing `/`. Two failure modes were reported:

- At the root, creating a folder named `/` failed with a misleading **"Duplicated folder name"** error, because the name collides with the root folder's own name (`/`) in the parent+name uniqueness check.
- In a subfolder, creating a folder named `a/b` appeared to succeed but broke every path-based recursive operation (folder creation, deletion, and moving), since the code splits names on `/`.

No layer of the stack validated folder names.

### Fix

Reject `/` in folder names at every layer:

- **Go** (`internal/service/file/file_folder.go`): `FileService.CreateFolder` and `FileService.MoveFiles` (for the `new_name` rename path) return an error when the name contains `/`.
- **Python** (`api/apps/services/file_api_service.py`): same checks in `create_folder` and `move_files`, keeping both API servers consistent.
- **Web**: the create-folder and rename dialogs now validate the name with zod (`Name cannot contain "/"`, i18n added for en / zh / zh-traditional), so the error is surfaced before a request is even sent.

### Tests

- `internal/service/file/file_folder_test.go`: new cases asserting both Go entry points reject `/`.
- `test/testcases/restful_api/test_file_routes_unit.py` and `test/testcases/test_http_api/test_file_app/test_file_routes.py`: new cases for the Python `create_folder`/`move_files` paths.

### Verification

- `bash build.sh --test ./internal/service/file/...` — pass.
- `pytest` on the two touched Python test files — 35 passed.
- `npx oxlint` on the 7 touched web files — clean; `ruff check` on the touched Python files — no new issues (baseline 41 → 37).
- Live check in the browser: both dialogs show the localized error and block submission.
- Live API check on both servers (Go :9384 and Python :9380): `POST /api/v1/files` with `name="/"` and `POST /api/v1/files/move` with `new_name="a/b"` are rejected with a clear message; creating a normal folder still works.

### DIFF
--- a/api/apps/services/file_api_service.py
+++ b/api/apps/services/file_api_service.py
@@ -28,6 +28,8 @@ from common import settings
 from common.constants import FileSource
 from common.misc_utils import get_uuid, thread_pool_exec
 
+logger = logging.getLogger(__name__)
+
 
 async def upload_file(tenant_id: str, pf_id: str, file_objs: list):
     """
@@ -48,7 +50,7 @@ async def upload_file(tenant_id: str, pf_id: str, file_objs: list):
 
     file_res = []
     for file_obj in file_objs:
-        MAX_FILE_NUM_PER_USER = int(os.environ.get("MAX_FILE_NUM_PER_USER", 0))
+        MAX_FILE_NUM_PER_USER = int(os.environ.get("MAX_FILE_NUM_PER_USER", "0"))
         if 0 < MAX_FILE_NUM_PER_USER <= await thread_pool_exec(DocumentService.get_doc_count, tenant_id):
             return False, "Exceed the maximum file number of a free user!"
 
@@ -96,7 +98,7 @@ async def upload_file(tenant_id: str, pf_id: str, file_objs: list):
     return True, file_res
 
 
-async def create_folder(tenant_id: str, name: str, pf_id: str = None, file_type: str = None):
+async def create_folder(tenant_id: str, name: str, pf_id: str | None = None, file_type: str | None = None):
     """
     Create a new folder or virtual file.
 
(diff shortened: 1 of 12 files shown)


wangq8 (commented):
Good defense-in-depth approach — validating at Go, Python, and web layers, with tests and i18n. The root cause (path-based lookups splitting on `/`) is real.

Main concern:

1. **Scope creep in `file_api_service.py`.** The PR bundles a lot of unrelated refactoring alongside the `/` check: `logging` → module-level `logger`, `str | None` / `str = None` type hints, `SimpleNamespace(to_json=dict)`, moving `import requests` into the function, `MAX_FILE_NUM_PER_USER` default `0` → `"0"`, and several `# noqa: BLE001` annotations. None of these relate to folder-name validation, and they make the diff harder to review and blame. Please split these into a separate cleanup PR.

2. **Merge conflict with #18590.** This PR creates `internal/service/file/file_folder_test.go`, which #18590 ("propagate file rename to all linked knowledgebase documents") also creates; both also touch `MoveFiles`. Coordinate ordering or consolidate the test files.

3. **Inconsistent scope.** The validation now applies to *file* renames too (not just folders) on the `move` paths — which is actually correct (a filename with `/` would break path lookups the same way) — but the PR title/message only says "folder names". Worth clarifying intent. Also note `upload_file` still doesn't reject `/` in uploaded filenames, so the validation isn't fully consistent across entry points.

Minor: the `logger.exception(...)` calls correctly drop the explicit `error: {e}` (traceback is auto-included), but keep the `errors.append(... {e})` — that's still fine since `e` stays bound in the `except` scope.

LGTM once the unrelated cleanup is separated and the cross-PR conflict is handled.

wangq8:
LGTM.

PR #18587, opened 2026-08-20
fix(embedding): scope LiteLLM request options to RAGcon

### LINKED ISSUE #18584: [Bug]: Embedding request failed with Together AI: Unrecognized request arguments supplied: drop_params
### Self Checks

main

### RAGFlow image version

v0.26.4-1-mt-33

```Markdown

```

### Actual behavior

When configuring Together AI (or OpenAI-API-Compatible endpoint pointing to api.together.ai/v1) for embedding models like intfloat/multilingual-e5-large-instruct, requests fail with:
Error code: 400 - {'error': {'message': 'Unrecognized request arguments supplied: drop_params'}}

This is caused by drop_params: True being passed directly in rag/llm/embedding_model.py, which Together AI's strict payload schema rejects.

### Expected behavior

_No response_

### Steps to reproduce

```Markdown
When configuring Together AI (or OpenAI-API-Compatible endpoint pointing to api.together.ai/v1) for embedding models like intfloat/multilingual-e5-large-instruct, requests fail with:
Error code: 400 - {'error': {'message': 'Unrecognized request arguments supplied: drop_params'}}

This is caused by drop_params: True being passed directly in rag/llm/embedding_model.py, which Together AI's strict payload schema rejects.
```

### Additional information

_No response_

yzl0ng:
### Summary

Fixes #18584.

Direct Together AI and OpenAI API embedding endpoints must not receive LiteLLM's `drop_params` control through the OpenAI SDK `extra_body`. RAGcon is the exception because its endpoint routes through LiteLLM and relies on parameter filtering.

After rebasing onto `main` (where #18627 independently removed the shared field), this PR now:

- adds request-contract coverage proving `TogetherAIEmbed` and `OpenAI_APIEmbed` omit `drop_params`;
- overrides `RAGconEmbed._call()` so only RAGcon sends `extra_body={"drop_params": true}`;
- adds regression coverage for the RAGcon exception.

### Tests

- Embedding model unit tests under Python 3.13.15 with package/settings side effects isolated: 27 passed
- `ruff check --select F,E9 rag/llm/embedding_model.py test/unit_test/rag/llm/test_embedding_model.py`
- `ruff format --check rag/llm/embedding_model.py test/unit_test/rag/llm/test_embedding_model.py`
- `python -m py_compile rag/llm/embedding_model.py test/unit_test/rag/llm/test_embedding_model.py`
- `git diff --check`


6ba3i (changes requested):
removing `drop_params` from `OpenAIEmbed` is correct for direct providers like Together AI, but it also changes `RAGconEmbed`, which intentionally routes through LiteLLM and inherits `_call()`.
LiteLLM supports `drop_params=True` per embedding call, so RAGcon may lose parameter filtering after this change.
I’d keep the shared removal, but override `_call()` in `RAGconEmbed` to preserve `extra_body={"drop_params": True}`, with a regression test covering that exception. No Go change needed.

6ba3i:
Thanks for addressing the previous review. After re-checking the RAGcon implementation and tracing the LiteLLM behavior more deeply, I think the RAGcon exception we previously requested was based on an incorrect assumption.

`drop_params=True` is meaningful as a LiteLLM Python-call control, but that does not establish that `"drop_params": true` belongs in the OpenAI-compatible `/embeddings` HTTP body sent to a LiteLLM proxy.

There is no repository evidence that RAGcon requires this field:

* `RAGconEmbed` was originally introduced as a normal `OpenAIEmbed` subclass with only a different base URL; it had no RAGcon-specific embedding request implementation.
* The current Go RAGcon embedding path sends only the standard `model`, `input`, and optional `dimensions` fields and does not send `drop_params`.
* `RAGconEmbed` accepts a configurable `base_url`, so unconditionally injecting `drop_params` can reproduce the same strict-endpoint HTTP 400 that #18584 is fixing.

The new test only proves that RAGFlow sends the field; it does not prove that RAGcon requires or accepts it.

I think the minimal fix here is therefore to remove the new `RAGconEmbed._call()` override and remove `test_ragcon_embedding_sends_litellm_drop_params()`. `RAGconEmbed` should continue using the shared `OpenAIEmbed._call()` behavior from #18627.

6ba3i:
could you please fix the issues raised in my comment ?

6ba3i:
overlaps #19476 , closing the pr

PR #18589, opened 2026-08-20
fix(web): keep multi-select count in sync after deleting a single file

euvre:
## Problem

In **File Manager**, when you multi-select several files and then delete one of them individually via the row's trash action, the bulk-operation bar still counts the deleted file (e.g. it keeps showing "Selected: 2 files" after one of the two was deleted). Bulk move / link-to-dataset / delete then also target the stale id, and the count only becomes correct again after a full page refresh.

## Root cause

The files table keeps multi-select state in `useRowSelection` (`rowSelection`, keyed by file id). The per-row delete in `ActionCell` called `handleRemoveFile([documentId])` but never pruned that id from `rowSelection`, so the deleted file's id lingered in the selection map and kept inflating the selected count and the id lists used by bulk operations.

## Fix

- `web/src/pages/files/action-cell.tsx`: `ActionCell` now receives `setRowSelection` (typed via `Pick<UseRowSelectionType, 'setRowSelection'>`). `onRemoveFile` awaits `handleRemoveFile([documentId])` and, only when the delete succeeds (`code === 0`), removes the file id from the selection map with a functional `setRowSelection` update. The remaining selections are preserved — only the deleted id is pruned.
- `web/src/pages/files/files-table.tsx`: passes the existing `setRowSelection` down to `ActionCell`.

## Verification

- `npm run type-check`: no new errors in the touched files (repo-wide errors are pre-existing and unrelated).
- `npx oxlint src/pages/files/action-cell.tsx src/pages/files/files-table.tsx`: 0 warnings, 0 errors.
- Live browser verification (vite dev server): uploaded two files, selected both ("Selected: 2 files"), deleted one individually → bar immediately updated to "Selected: 1 files" and the remaining row stayed checked; opened bulk Delete → dialog correctly showed "Selected 1 files"; confirmed bulk delete removed exactly the remaining file and the selection bar cleared.

### DIFF
--- a/web/src/hooks/__tests__/use-delete-file.test.tsx
+++ b/web/src/hooks/__tests__/use-delete-file.test.tsx
@@ -1,0 +1,117 @@
+import message from '@/components/ui/message';
+import fileManagerService from '@/services/file-manager-service';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { renderHook } from '@testing-library/react';
+import React from 'react';
+
+import { useDeleteFile } from '../use-file-request';
+
+jest.mock('@/services/file-manager-service', () => ({
+  __esModule: true,
+  default: { removeFile: jest.fn() },
+}));
+
+jest.mock('@/components/ui/message', () => ({
+  __esModule: true,
+  default: { success: jest.fn(), error: jest.fn() },
+}));
+
+// @/utils/request pulls in the legacy i18n/locales graph that touches
+// import.meta.env, which jest cannot transform; the delete mutation never
+// calls it directly, so stub the module.
+jest.mock('@/utils/request', () => ({ __esModule: true, default: jest.fn() }));
+
+jest.mock('@/utils/file-util', () => ({ downloadFileFromBlob: jest.fn() }));
+
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+jest.mock('react-router', () => ({
+  useSearchParams: jest.fn(() => [new URLSearchParams()]),
+}));
+
+jest.mock('ahooks', () => ({ useDebounce: (value: unknown) => value }));
+
+// Sibling hooks imported by use-file-request pull in the legacy locales graph
+// via logic-hooks; the delete mutation never calls them, so stub both.
+jest.mock('@/hooks/logic-hooks', () => ({
+  useGetPaginationWithRouter: jest.fn(() => ({})),
+  useHandleSearchChange: jest.fn(() => ({})),
+}));
+
+jest.mock('@/hooks/route-hook', () => ({
+  useSetPaginationParams: jest.fn(() => jest.fn()),
+}));
+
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Correct, tight fix. Pruning the deleted id from `rowSelection` via a functional update only on `code === 0` is exactly right, and I confirmed `handleRemoveFile` returns the response code (so `code === 0` is a valid success check).

Notes (non-blocking):

1. **Unhandled async.** `onRemoveFile` is now `async`, but the call site (the trash action `onClick`) likely invokes it without awaiting/`void`. Today `handleRemoveFile` returns a code and never throws, so this is fine — but a defensive `.catch()` (or wrapping in `void`) would prevent an unhandled rejection if `removeDocument` ever starts throwing.

2. **Type change is slightly wide.** `ActionCell` now takes `Pick<UseRowSelectionType, 'setRowSelection'>`, which is good, but verify every `ActionCell` usage actually has a `setRowSelection` available — `files-table.tsx` is updated here, and the type-check passing suggests no other call sites. Just confirming there are no other consumers of `ActionCell`.

3. Minor: the selection pruning only runs on the single-row delete path. Bulk delete already clears selection separately (per the PR description), so no gap — just calling it out for completeness.

LGTM.

euvre:
Thanks @wangq8 for the careful review and the LGTM! Follow-up on the three non-blocking notes (commit 5769b2b87):

1. **Unhandled async — applied.** You were right: `useDeleteFile`'s `mutationFn` awaited `fileManagerService.removeFile` with no guard, so a network-level failure would reject `mutateAsync` → `handleRemoveFile` → `onRemoveFile`, and `ConfirmDeleteDialog` invokes it via `onClick={onOk}` without awaiting, producing an unhandled rejection. I fixed it at the owning layer: the mutation now wraps the request in try/catch and returns `undefined` on failure, mirroring the existing `useUploadFile` precedent. This single change covers both the single-row delete in `action-cell.tsx` and the bulk delete in `use-bulk-operate-file.tsx`; the `code === 0` checks stay intact, so a failed delete leaves the row selection untouched.

2. **Other `ActionCell` consumers — verified.** A codebase-wide search confirms `ActionCell` from `pages/files/action-cell.tsx` is only rendered by `files-table.tsx` (`DatasetActionCell` is a separate component), and type-check passes with no errors in the touched files.

3. **Bulk path — acknowledged, no gap.** Bulk delete resets the whole selection via `setRowSelection({})` on success, and it now also gains the defensive catch for free.

**Verification:** added a focused regression test (`web/src/hooks/__tests__/use-delete-file.test.tsx`) covering the success path (returns 0 + toast), a non-zero backend code (passthrough, no toast), and the rejection path (resolves `undefined` instead of throwing — the new behavior) — all 3 pass under jest. `npm run type-check` reports the identical 221 pre-existing errors on main before and after this change (zero new; none in the touched files). No shared-service end-to-end verification was run on this line; the assessment is based on unit tests plus code-level analysis of the call/data flow.

dcc123456 (approved):

PR #18590, opened 2026-08-20
Fix: propagate file rename to all linked knowledgebase documents

euvre:
### Summary

When a file in **File Management** is linked to multiple knowledgebases, renaming the file only updated the document name in the *first* linked dataset. The other linked datasets kept showing the old name, and the preview link on the File Management page pointed to the first linked dataset, so the rename appeared to only partially apply.

**Root cause**

Both rename paths in the Go `FileService` (`internal/service/file/file_folder.go`) looked up the file's `file2document` links but only updated `informs[0]` — the first linked document:

- `MoveFiles` (used by the rename endpoint when the file stays in the same folder)
- `moveEntryRecursive` (used when moving/renaming a file across folders)

The Python reference implementation (`api/db/services/file_service.py::_rename_linked_documents`) iterates over **all** `file2document` links and renames every linked document; the Go port only handled the first one.

**Fix**

- Extracted a `renameLinkedDocuments(ctx, fileID, newName)` helper that iterates over all `file2document` links for the file and updates every linked document's `name`.
- Both `MoveFiles` and `moveEntryRecursive` now call this helper instead of only updating `informs[0]`.

**Verification**

- New unit tests in `internal/service/file/file_folder_test.go` cover both paths: renaming a file linked to two documents now renames both documents. `bash build.sh --test ./internal/service/file/...` passes.
- End-to-end in the browser: uploaded a file, linked it to two datasets (`rename-kb-1`, `rename-kb-2`), renamed it in File Management — before the fix only the first dataset showed the new name; after rebuilding and re-walking the path, **both** datasets show the new name.

### DIFF
--- a/internal/service/file/file_folder.go
+++ b/internal/service/file/file_folder.go
@@ -510,11 +510,14 @@ func (s *FileService) MoveFiles(ctx context.Context, uid string, srcFileIDs []st
 }
 
 // renameLinkedDocuments renames every knowledgebase document linked to the
-// given file so the new name is reflected in all linked datasets.
+// given file so the new name is reflected in all linked datasets. Only the
+// document name is synced: the Python reference (_rename_linked_documents in
+// api/apps/services/file_api_service.py) updates no other denormalized field
+// on rename.
 func (s *FileService) renameLinkedDocuments(ctx context.Context, fileID, newName string) error {
 	informs, err := s.file2DocumentDAO.GetByFileID(ctx, dao.DB, fileID)
 	if err != nil {
-		return nil
+		return err
 	}
 	documentDAO := dao.NewDocumentDAO()
 	for _, inform := range informs {
--- a/internal/service/file/file_folder.go
+++ b/internal/service/file/file_folder.go
@@ -500,18 +500,32 @@ func (s *FileService) MoveFiles(ctx context.Context, uid string, srcFileIDs []st
 			return false, "Database error (File rename)!"
 		}
 
-		// Update associated document name if exists
-		informs, err := s.file2DocumentDAO.GetByFileID(ctx, dao.DB, file.ID)
-		if err == nil && len(informs) > 0 && informs[0].DocumentID != nil {
-			docID := *informs[0].DocumentID
-			documentDAO := dao.NewDocumentDAO()
-			if err = documentDAO.UpdateByID(ctx, dao.DB, docID, map[string]interface{}{"name": newName}); err != nil {
-				return false, "Database error (Document rename)!"
-			}
+		// Update names of all linked documents if any exist
+		if err = s.renameLinkedDocuments(ctx, file.ID, newName); err != nil {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
The fix is correct and matches the Python reference (`_rename_linked_documents` iterating over all `file2document` links). Extracting `renameLinkedDocuments` and calling it from both `MoveFiles` and `moveEntryRecursive` closes the gap cleanly, and the tests cover both paths.

Two things to address:

1. **`renameLinkedDocuments` swallows the `GetByFileID` error.** `if err != nil { return nil }` silently treats a genuine DB failure the same as "no links", so a rename would skip propagation without any signal. `moveEntryRecursive` had this same weakness before (so no regression), but since you're introducing a shared helper it's the right place to fix it — return the error (or at least log a warning) instead of dropping it.

2. **Merge conflict with #18586.** This PR creates `internal/service/file/file_folder_test.go`, and #18586 ("reject folder names containing `/`") also creates that same file (and both modify `MoveFiles` in `file_folder.go`). Whichever merges first, the second will conflict. Please coordinate with the author / reorder, or consolidate the new tests into the shared file.

Minor: confirm updating only `name` is sufficient (no `location`/other denormalized fields to keep in sync) — per the Python reference it is, but worth a comment in the helper.

LGTM once the error handling and cross-PR conflict are sorted.

wangq8:
LGTM.

euvre:
Thanks @wangq8 for the careful review (and the LGTM). Addressed in commit `6f98b5a`:

**1. `renameLinkedDocuments` swallowing the `GetByFileID` error (also flagged by CodeRabbit) — fixed.** The helper now returns the lookup error instead of `nil`, so a genuine DB failure can no longer masquerade as "no links". Both callers already surface it: `MoveFiles` returns `Database error (Document rename)!` and `moveEntryRecursive` wraps it, matching the Python reference (`_rename_linked_documents`, where a failed `get_by_file_id` raises) and the handling in every other `GetByFileID` caller (e.g. `file_delete.go`). Added regression test `TestRenameLinkedDocumentsLookupErrorPropagates`, which drops the file2document table and asserts the helper errors and `MoveFiles` reports the failure.

**2. Merge conflict with #18586 — acknowledged.** Both PRs create `file_folder_test.go` and touch `MoveFiles`; there is nothing to change in this PR itself. Whichever lands second will be rebased; the new tests live in distinct test functions, so consolidating them into the shared file will be trivial.

**3. Minor (name-only update) — done.** Added a comment in the helper documenting that only the document `name` is synced, per the Python reference, which updates no other denormalized field on rename.

Verification: `bash build.sh --test ./internal/service/file/...` passes, including the new test (all 3 rename tests green). Verification boundary: no shared-service end-to-end run on this line; based on unit tests + line-by-line comparison with the `api/` Python reference.

wangq8:
@euvre resolve conflict.

PR #18591, opened 2026-08-20
fix(web): confirm selected MultiSelect option on Enter instead of deselecting

euvre:
### Summary

In the cmdk-based `MultiSelect` component (`web/src/components/ui/multi-select.tsx`), pressing Enter while an already-selected option was highlighted deselected that option. cmdk's root keydown handler dispatches a select event for the highlighted item on Enter, and the component's `onSelect` toggles the option — so keyboard users confirming a dropdown choice (e.g. the model type in the Edit model dialog at `/user-setting/model`) accidentally removed it.

Enter on an already-selected option is now treated as confirmation instead of a toggle:

- Each option's `CommandItem` carries a `data-option-value` attribute so the highlighted item can be mapped back to its option value.
- In the search input's `onKeyDown`, when the dropdown is open and the highlighted item is an already-selected option, the handler calls `preventDefault()` (cmdk checks `defaultPrevented` before dispatching its own Enter selection) and closes the dropdown, leaving the selection unchanged.

Behavior that is intentionally preserved:

- Enter on an unselected option still selects it and keeps the dropdown open.
- Enter on (Select all)/Clear/Close keeps cmdk's default handling.
- Clicking a selected option still deselects it.

### Test plan

- New unit tests in `web/src/components/ui/__tests__/multi-select.test.tsx` (4 cases: Enter on a selected option closes the dropdown without changing the selection; Enter on an unselected option selects it and keeps the dropdown open; Enter on (Select all) toggles all options; clicking a selected option deselects it) — all pass via `npx jest src/components/ui/__tests__/multi-select.test.tsx`.
- `npx oxlint` on the touched files is clean; `tsc --noEmit` reports no errors in the touched files (only pre-existing errors in unrelated files).
- Manually verified in the browser on the Edit model dialog (`glm-4.6v-Flash`): Enter on the highlighted, already-selected "Chat" closes the dropdown with "Chat" still selected; Enter on unselected "Embedding" selects it and keeps the dropdown open; clicking the selected "Chat" deselects it.

### DIFF
--- a/web/src/components/ui/__tests__/multi-select.test.tsx
+++ b/web/src/components/ui/__tests__/multi-select.test.tsx
@@ -1,0 +1,115 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { MultiSelect } from '../multi-select';
+
+beforeAll(() => {
+  // jsdom does not implement these browser APIs used by cmdk/radix.
+  Element.prototype.scrollIntoView = jest.fn();
+  (globalThis as Record<string, unknown>).ResizeObserver = class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  };
+});
+
+const options = [
+  { label: 'Chat', value: 'chat' },
+  { label: 'VLM', value: 'vlm' },
+  { label: 'OCR', value: 'ocr' },
+];
+
+function renderOpenedMultiSelect(onValueChange = jest.fn()) {
+  const utils = render(
+    <MultiSelect
+      options={options}
+      defaultValue={['chat', 'vlm']}
+      onValueChange={onValueChange}
+      placeholder="Model type"
+    />,
+  );
+
+  fireEvent.click(
+    utils.container.querySelector('button') as HTMLButtonElement,
+  );
+
+  const root = document.querySelector('[cmdk-root]') as HTMLElement;
+  expect(root).not.toBeNull();
+
+  const input = root.querySelector('[cmdk-input]') as HTMLInputElement;
+  const items = () =>
+    Array.from(root.querySelectorAll<HTMLElement>('[cmdk-item]'));
+  const itemByText = (text: string) => {
+    const item = items().find((element) =>
+      element.textContent?.includes(text),
+    );
+    expect(item).toBeDefined();
+    return item as HTMLElement;
+  };
+
+  // (Select all), Chat, VLM, OCR, Clear, Close
+  expect(items()).toHaveLength(6);
+
+  return { onValueChange, input, items, itemByText };
+}
+
+function pressArrowDown(input: HTMLElement, times: number) {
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Nice, focused fix with good test coverage. The `data-option-value` approach to map the highlighted cmdk item back to an option is clean, and preserving click-to-deselect is important.

A couple of notes (non-blocking):

1. **Reliance on `event.currentTarget` synchronously.** The handler reads `event.currentTarget.closest(...)` inside the React `onKeyDown`. This is safe in React 17+/18 (no synthetic-event pooling), but if this code path were ever made async or migrated to a deferred handler, `currentTarget` would be null. Consider capturing `event.currentTarget` (or the root ref) at the top of the handler to make the intent explicit.

2. **Coupling to cmdk DOM internals.** The `[cmdk-item][aria-selected="true"]` selector depends on cmdk's exact attributes. That's a reasonable, documented approach, but a shared helper for "which option is highlighted" would centralize this and make it easier to update if cmdk changes its attribute scheme.

3. Minor: `selectedValues.includes(optionValue)` assumes option values are always strings (they are here), and the `(Select all)`/Clear/Close items correctly have no `data-option-value` so they fall through to cmdk's default handling — verified that behavior in the test.

LGTM otherwise.

euvre:
Thanks for the careful review, @wangq8! Here is how each note was handled:

**1. Reliance on `event.currentTarget` synchronously — applied.** `handleInputKeyDown` now captures `event.currentTarget` into a local `input` at the top of the handler (with a short comment explaining that React clears `currentTarget` after the handler returns), and both the Enter and Backspace branches use it. This is a pure refactor with no behavior change; the existing tests in `multi-select.test.tsx` already cover both branches. Commit: d81c36183.

**2. Coupling to cmdk DOM internals — deferred, with reasoning.** The `[cmdk-item][aria-selected="true"]` selector (and the `closest('[cmdk-root]')` lookup) is used at exactly one call site in the entire codebase, and the cmdk dependency is documented in the comment above it. Introducing a shared helper for a single consumer would add indirection without actually centralizing anything. If a second component ever needs to map the highlighted cmdk item back to an option, extracting a shared helper at that point makes sense — happy to do it then.

**3. String option values / fall-through items — acknowledged, no change needed.** Option values are typed `string`, and the `(Select all)`/Clear/Close items intentionally carry no `data-option-value` so they keep cmdk's default Enter handling, as covered by the tests.

**Verification:** `npm run type-check` passes with no new errors for the touched files (the ~221 pre-existing errors elsewhere on main are unrelated and unchanged). No shared-service end-to-end verification was run; this is based on type-check + the existing unit tests + code-level analysis, since the change is a behavior-preserving refactor.

euvre:
The request was rejected.

PR #18592, opened 2026-08-20
fix(parser): clean .eml text output for chunking

euvre:
### Summary

Parsing an `.eml` file with the Email chunk method produced chunks full of raw SMTP transport headers (`Received` chains, DKIM/ARC signatures), RFC 2047 encoded-words (e.g. an undecoded `=?UTF-8?B?...?=` subject), and shredded HTML/CSS markup.

Root cause is in the Go ingestion email parser (`internal/parser/parser/email_parser.go`) text output path, which feeds the chunker directly:

1. The `metadata` map (every non-basic header) was flattened into the chunkable text, burying the message under transport noise.
2. `net/mail` does not decode RFC 2047 encoded-words, so non-ASCII headers arrived as unreadable `=?utf-8?B?...?=` blobs that chunk delimiters then shredded at the `?` characters.
3. `text_html` was emitted as raw markup, so tags and inline CSS landed in chunks.

Fix:

- Exclude `metadata` from the text output (still present in the JSON output).
- Decode RFC 2047 encoded-words in all header values, with a `CharsetReader` covering the gb2312/gbk/gb18030 family for CJK mail.
- Emit the visible text of the HTML body (script/style/head content dropped) instead of raw markup in the text output.

The JSON output contract is unchanged: `metadata` and raw `text_html` are preserved, and `.msg` handling is untouched.

Verification:

- New unit test `TestEmailParser_EmlTextChunkText` builds a multipart `.eml` with an RFC 2047 encoded UTF-8 subject plus plain and HTML bodies, and asserts the decoded subject, exactly the expected body text, no header noise, and that the JSON output still keeps raw `text_html`/`metadata`. All `./internal/parser/...` and `./internal/ingestion/...` packages pass.
- End-to-end: uploaded `带附件的email.eml` (an .eml with an encoded Chinese subject and an attachment) to a local knowledge base, re-parsed with the Email chunk method, and confirmed via the chunks API that the chunk contains only readable `from`/`subject`/`to`/body/attachment text — no transport headers, no encoded-words, no markup.

### DIFF
--- a/internal/parser/parser/email_parser.go
+++ b/internal/parser/parser/email_parser.go
@@ -227,7 +227,7 @@ func targetFieldsSet(fields []string) map[string]bool {
 // The "gb2312" label maps to GBK, not HZGB2312: mail labeled gb2312 carries
 // plain 8-bit GB2312 bytes (GBK is a compatible superset), while HZGB2312
 // only decodes the 7-bit HZ escape form used under the distinct "hz-gb-2312"
-// label. This mirrors Python, where the gb2312 codec decodes plain bytes.
+// label.
 func charsetEncoding(charset string) (encoding.Encoding, bool) {
 	switch strings.ToLower(strings.TrimSpace(charset)) {
 	case "gb2312", "gbk":
@@ -502,6 +502,13 @@ func htmlBodyToText(body string) string {
 	return strings.TrimSpace(b.String())
 }
 
+// isEmailBlockTag reports whether tag delimits a line in flattened email
+// body text: the standard block tags plus table tags, since email uses
+// tables for layout and their cells/rows must not fuse either.
+func isEmailBlockTag(tag string) bool {
+	return isBlockTag(tag) || tag == "table" || tag == "td" || tag == "th"
+}
+
 func walkHTMLBodyText(n *html.Node, w *leafWriter) {
 	switch n.Type {
 	case html.TextNode:
@@ -523,12 +530,18 @@ func walkHTMLBodyText(n *html.Node, w *leafWriter) {
 			w.hardBreak()
 			return
 		}
+		// A nested block starts on a new line when text already precedes
+		// it, so outer text and inner block text stay separate
+		// ("<div>a<p>b</p></div>" flattens to "a\nb", not "ab").
+		if isEmailBlockTag(n.Data) && w.b.Len() > 0 && !w.endsNL {
+			w.hardBreak()
+		}
 		for child := n.FirstChild; child != nil; child = child.NextSibling {
 			walkHTMLBodyText(child, w)
 		}
 		// Block-level tags and table cells/rows end the current line so
 		// adjacent block/cell text does not fuse together.
-		if (isBlockTag(n.Data) || n.Data == "table" || n.Data == "td" || n.Data == "th") && w.b.Len() > 0 && !w.endsNL {
+		if isEmailBlockTag(n.Data) && w.b.Len() > 0 && !w.endsNL {
 			w.hardBreak()
 		}
 	}
--- a/internal/parser/parser/email_parser.go
+++ b/internal/parser/parser/email_parser.go
@@ -34,7 +34,11 @@ import (
 	"github.com/AkmalOt/gomsg"
 	"golang.org/x/net/html"
 	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+	"golang.org/x/text/encoding/japanese"
+	"golang.org/x/text/encoding/korean"
 	"golang.org/x/text/encoding/simplifiedchinese"
+	"golang.org/x/text/encoding/traditionalchinese"
 	"golang.org/x/text/transform"
 
 	"ragflow/internal/utility"
@@ -214,18 +218,43 @@ func targetFieldsSet(fields []string) map[string]bool {
 
 // -- header decoding (RFC 2047) --
 
+// charsetEncoding maps a MIME charset label to its decoder, covering the
+// charsets common in real-world mail beyond utf-8: the Chinese families
+// (gb2312/gbk/gb18030, big5), shift_jis, euc-kr, and latin-1. It is shared by
+// the RFC 2047 header decoder and decodeMailPayload so headers and bodies
+// resolve a charset label identically.
+//
+// The "gb2312" label maps to GBK, not HZGB2312: mail labeled gb2312 carries
+// plain 8-bit GB2312 bytes (GBK is a compatible superset), while HZGB2312
+// only decodes the 7-bit HZ escape form used under the distinct "hz-gb-2312"
+// label. This mirrors Python, where the gb2312 codec decodes plain bytes.
+func charsetEncoding(charset string) (encoding.Encoding, bool) {
+	switch strings.ToLower(strings.TrimSpace(charset)) {
+	case "gb2312", "gbk":
+		return simplifiedchinese.GBK, true
+	case "gb18030":
+		return simplifiedchinese.GB18030, true
+	case "hz-gb-2312":
+		return simplifiedchinese.HZGB2312, true
+	case "big5":
+		return traditionalchinese.Big5, true
+	case "shift_jis", "shift-jis", "sjis":
+		return japanese.ShiftJIS, true
+	case "euc-kr":
+		return korean.EUCKR, true
+	case "iso-8859-1", "iso8859-1", "latin1":
+		return charmap.ISO8859_1, true
+	}
+	return nil, false
+}
+
 // rfc2047Decoder decodes RFC 2047 encoded-words (e.g. "=?utf-8?B?...?=") in
-// header values. utf-8 is handled natively by mime; the CharsetReader adds
-// the same CJK family that decodeMailPayload covers for bodies.
+// header values. utf-8 is handled natively by mime; the CharsetReader covers
+// the non-UTF-8 charsets via charsetEncoding.
 var rfc2047Decoder = &mime.WordDecoder{
 	CharsetReader: func(charset string, input io.Reader) (io.Reader, error) {
-		switch strings.ToLower(strings.TrimSpace(charset)) {
-		case "gb2312":
-			return transform.NewReader(input, simplifiedchinese.HZGB2312.NewDecoder()), nil
-		case "gbk":
-			return transform.NewReader(input, simplifiedchinese.GBK.NewDecoder()), nil
-		case "gb18030":
-			return transform.NewReader(input, simplifiedchinese.GB18030.NewDecoder()), nil
+		if enc, ok := charsetEncoding(charset); ok {
+			return transform.NewReader(input, enc.NewDecoder()), nil
 		}
 		return nil, fmt.Errorf("email: unsupported header charset %q", charset)
 	},
@@ -478,8 +507,11 @@ func walkHTMLBodyText(n *html.Node, w *leafWriter) {
 	case html.TextNode:
 		w.writeText(n.Data)
 	case html.DocumentNode:
-		// html.Parse wraps the fragment in a Document node; without
-		// descending here the whole body would be dropped.
+		// html.Parse always wraps the fragment in a Document node holding an
+		// <html><body> skeleton; the walker relies on descending through those
+		// transparent containers (head/script/style subtrees are skipped by the
+		// ElementNode case), so without descending here the whole body would
+		// be dropped.
 		for child := n.FirstChild; child != nil; child = child.NextSibling {
 			walkHTMLBodyText(child, w)
 		}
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for this fix — the root cause is correctly identified and the text output is much cleaner. The JSON contract being preserved (metadata + raw text_html) is the right call, and the unit test is thorough.

A few notes (non-blocking):

1. **CharsetReader coverage is narrow.** `rfc2047Decoder.CharsetReader` only handles `gb2312`/`gbk`/`gb18030`. Any other non-UTF-8 charset (e.g. `Big5` — very common in traditional-Chinese mail — plus `ISO-8859-1`, `EUC-KR`, `Shift_JIS`) makes `DecodeHeader` return an error, and `decodeHeaderWord` then returns the whole header value *undecoded* (not partially decoded). This is a graceful fallback, but for multi-word headers a single unsupported encoded-word leaves the entire header encoded. Consider a more complete charset resolver (e.g. `golang.org/x/text/encoding/ianaindex` or `golang.org/x/net/html/charset`) or at least a TODO for Big5.

2. **Duplicated HTML-walk logic.** `htmlBodyToText`/`walkHTMLBodyText` re-implements the block-tag / whitespace / `leafWriter` folding that `html_parser.go`'s `walkHTMLLeaf` already does. It diverges slightly (extra `table`/`td`/`th` hard breaks). Worth extracting a shared walker if these need to stay in sync.

3. Minor: the `DocumentNode` special-case and the `head/script/style/noscript` skip at the `ElementNode` level are correct but subtle — a short comment already helps, consider noting that `html.Parse` wraps fragments in `<html><body>` so the walker relies on descending through `body`.

Overall this looks good to merge once the charset question is settled.

euvre:
Thanks @wangq8 for the careful review — glad the root cause and the JSON contract land well. The charset question is now settled in commit 7b253ed0b.

**1. gb2312 → GBK (the blocking charset issue; also flagged by CodeRabbit) — fixed.** Confirmed: HZGB2312 expects 7-bit HZ escape sequences, while real-world mail labeled gb2312 carries plain 8-bit bytes, so those headers failed decoding and arrived as raw "=?gb2312?B?...?=" blobs. A new shared `charsetEncoding` resolver maps `gb2312` → `GBK` (a compatible superset, matching Python's gb2312 codec in `rag/flow/parser/parser.py`), and `hz-gb-2312` keeps the true HZ decoder. The same wrong mapping also existed in the body path (`decodeWithCharset`), where it was silently masked by the fallback chain — both paths now share the one resolver, so headers and bodies resolve a label identically.

**2. Narrow CharsetReader coverage — fixed.** The resolver now also covers big5, shift_jis, euc-kr, and iso-8859-1, all from the already-vendored `golang.org/x/text` (no new dependency). This goes beyond a TODO for Big5; a declared big5 body now decodes correctly too instead of degrading to latin-1 mojibake. The graceful all-or-nothing fallback for genuinely unknown charsets is kept (and tested), since mime.WordDecoder offers no partial decode.

**3. Duplicated HTML-walk logic — deferred, with reasoning.** The whitespace-folding core (`leafWriter`) is already shared with html_parser.go. What remains in `walkHTMLBodyText` is ~20 lines of tree-walk glue whose table semantics intentionally differ: email treats tables as layout and flattens cell text, while the standalone HTML parser must emit each <table> as a structured `doc_type_kwd:"table"` item (and handles <pre>/<textarea>). Unifying them would mean parameterizing table/item handling across both parsers — a larger refactor with regression risk to the HTML chunk method, out of scope here. Happy to do it as a follow-up if you prefer.

**4. Comment on the html.Parse wrapper — applied.** The DocumentNode comment now notes that html.Parse wraps fragments in an <html><body> skeleton and that the walker relies on descending through those transparent containers.

**Verification:** no shared-service end-to-end run on this line; based on unit tests + Python comparison. New coverage: table-driven RFC 2047 encoded-word tests for gb2312/gbk/gb18030/big5/shift_jis/euc-kr/iso-8859-1/hz-gb-2312 plus the unknown-charset fallback, an end-to-end gb2312-subject regression, and body-side declared-charset tests (byte anchors cross-checked against Python codecs). `bash build.sh --test ./internal/parser/parser/...` passes (full package).

xugangqiang:
Review feedback (2 substantive findings beyond the existing comments):

**1. [Real bug, untested] Nested block text fusion.** The new `walkHTMLBodyText` only inserts a hard line break *after* traversing a block node's children, not *before* descending into it. For `<div>prefix<p>para</p></div>`:
- `writeText("prefix")` runs, then we descend into `<p>` and `writeText("para")` runs → output becomes `prefixpara\n` (prefix and para are fused with no separating newline).

The existing test uses `<div><div>text</div></div>`, where the outer `div` has **no prefix text before the nested block**, so this path is never exercised. This is exactly the CodeRabbit outside-diff finding, and it is valid: `div`-wrapped paragraphs are extremely common in email layout, so this directly undermines the "clean chunk" goal for real-world mail.

Suggested fix — mirror the post-traversal break *before* the child loop:
```go
if (isBlockTag(n.Data) || n.Data == "table" || n.Data == "td" || n.Data == "th") && w.b.Len() > 0 && !w.endsNL {
    w.hardBreak()
}
for child := n.FirstChild; child != nil; child = child.NextSibling {
    walkHTMLBodyText(child, w)
}
```
and add a regression test with markup like `<div>prefix<p>para</p></div>` asserting line-separated output. **Recommend fixing in this PR** rather than deferring.

**2. [Go↔Python divergence, unregistered] `metadata` removed from text output while Python still includes it.** In `rag/flow/parser/parser.py:1362-1364`, the Python `_email` text path does `content_txt += f"{k}:{json.dumps(v)}"`, so the Python text output *also* contains `metadata:{...}` (with Received/DKIM). This PR makes the Go text output exclude `metadata`, creating an intentional divergence from the Python DSL baseline.

Per the team's parity discipline (aligning with Python is the means, not the end; Go-superior divergences should be registered as `go_intentional` rather than changed silently), this change is an *improvement* and should be kept — but it should be explicitly recorded as a `go_intentional` decision so future parity runs don't flag it as a regression.

(Note: I reviewed read-only via a throwaway worktree; no changes made to any branch.)

euvre:
Thanks @wangq8 for the careful review — responses to each note below, plus one follow-up commit (9231d58e5).

**1. CharsetReader coverage (Big5 / ISO-8859-1 / EUC-KR / Shift_JIS)** — Already covered on current HEAD. Commit 7b253ed (pushed after your review) extracted a shared `charsetEncoding` resolver used by both the RFC 2047 header decoder and the body decoder, and widened it to exactly those charsets: big5, shift_jis, euc-kr, iso-8859-1 (all from the already-vendored `golang.org/x/text`, so no new dependency). Table-driven encoded-word tests for every supported charset are in `TestDecodeHeaderWord_Charsets`, and an unsupported charset still degrades gracefully (header value left untouched, body falls back to latin-1). We kept the explicit map over `ianaindex` deliberately: it is small, deterministic, and avoids pulling in the large IANA table for a fixed set of charsets that actually occur in mail.

**2. Duplicated HTML-walk logic** — Verified, but we'd prefer not to merge the two walkers in this PR. The parts that genuinely must stay in sync already are shared: `leafWriter` (CSS whitespace folding, hard breaks) and `isBlockTag` both live in `html_parser.go` and are reused by the email walker. The walker bodies themselves have different output contracts — `walkHTMLLeaf` emits structured chunk items (tables preserved as `doc_type_kwd:"table"` markup, `<pre>` verbatim, flush-at-boundary), while `walkHTMLBodyText` flattens to a single visible-text string and deliberately descends into layout tables. A single walker would need mode flags at each of those divergence points, which couples the two parsers to save ~20 lines of traversal skeleton. Happy to revisit if a third consumer appears.

**3. DocumentNode subtlety** — Already documented: the comment above the `DocumentNode` case now notes that `html.Parse` wraps the fragment in an `<html><body>` skeleton and that the walker relies on descending through those transparent containers.

**Nested-block fusion (merge-risk follow-up, fixed in 9231d58e5)** — Confirmed by tracing the walker: `<div>prefix<p>paragraph</p></div>` flattened to `prefixparagraph` because only a post-traversal break existed. `walkHTMLBodyText` now emits the same structural break before descending into a block/table element when the writer already holds a partial line, and the shared tag condition was extracted into `isEmailBlockTag` so the pre/post-traversal breaks cannot drift apart. Covered by new table-driven `TestHTMLBodyToText` regressions (nested blocks, layout tables, inline tags, `<br>`, head/script/style skipping).

**Python compatibility wording** — Removed the newly added Python-codec comparison from the `charsetEncoding` comment and the `TestDecodeHeaderWord_Charsets` comment (9231d58e5); the direct technical rationale (gb2312 mail carries plain 8-bit bytes, GBK is a compatible superset, HZGB2312 is only for the `hz-gb-2312` label) is kept.

Verification: `bash build.sh --test ./internal/parser/parser/...` passes (full package, incl. the new regression tests). No shared-service end-to-end verification was run for this round; the conclusions are based on unit tests plus code-level analysis.

xugangqiang:
@euvre 

Thanks for the clean fix — the chunk-quality problem for `.eml` emails is real and this PR addresses all three root causes (metadata dump, undecoded RFC 2047 headers, raw HTML markup). A few items to address before/after merge:

**1. Fix two inaccuracies in the PR description**
- "JSON output contract is unchanged": the `subject`/`from`/metadata **values are now RFC 2047-decoded** (previously raw `=?...?=`). The structure is preserved, but the values changed. Please reword to "header values are now decoded; `metadata` and raw `text_html` are preserved."
- ".msg handling is untouched": this is not strictly true. `parseMSG` also emits `metadata` (`email_parser.go:514`) and the text output goes through the **shared** flatten loop, so `.msg` text output now also drops the metadata dump. (`.msg` does not emit `text_html` per `email_parser.go:522`, so the HTML part is unaffected.) Please update the claim or explicitly state this is an intended, consistent change.

**2. Add tests (coverage gaps)**
- Body-path charset decoding is currently untested for the changed branch. The existing `TestReadMailBody_AttachmentPreservesRaw` uses `charset=gbk` (unaffected). Please add:
  - a `gb2312` body-decode test (locks in the new GBK routing),
  - an `hz-gb-2312` body-decode test (locks in the tradeoff that HZ stays under its own label).
  Both exercise `decodeMailPayload`/`decodeWithCharset`, not just the header decoder.
- A `.msg` text-output test asserting `metadata` is excluded from the text output (this also documents the point from #1).
- An HTML-walker test with a `<table>` body, asserting cell text is flattened and cells do not fuse (current coverage only has an ``/`<div>` case).

**3. Follow-up (non-blocking): multipart/alternative dedup**
For `multipart/alternative`, the text output now contains **both** the plain `body` and the HTML-visible text, so the same content appears twice in a chunk. Consider keeping only one part (prefer plain) to avoid doubled text. This is pre-existing duplication (the PR only cleans the HTML markup, it doesn't dedup), so it can be a separate follow-up.

**Note on CodeRabbit's comment:** its actionable item (use `GBK` instead of `HZGB2312` for the `gb2312` encoded-word branch, plus table-driven tests) is already satisfied by the current head (`07cad18…`) via the shared `charsetEncoding` helper and `TestDecodeHeaderWord_Charsets`. The comment was posted against an older commit (`1161e3a7…`) and is now stale — a re-review should auto-resolve it.

Net: this is a clear improvement to email chunk quality and is safe to merge once the description wording is fixed and the tests above are added.

euvre:
Thanks @xugangqiang for the careful review — all items handled in 485adb41.

**1. PR description inaccuracies — agreed on both.**
- Yes, JSON header **values** are now RFC 2047-decoded; only the structure is unchanged. Replacement for the final sentence, paste-ready: “The JSON output structure is unchanged — header values are now decoded in place of raw `=?...?=` blobs; `metadata` and raw `text_html` are preserved. `.msg` text output shares the same flatten loop as `.eml`, so it also excludes the metadata dump (and never emits `text_html`); this is intended and keeps both formats consistent.”
- Confirmed in code: `parseMSG` always emits `metadata` (`email_parser.go`), the shared flatten loop skips it, and the `.msg` branch emits only `text` (HTML body falls back into `text` as plain text). The new test below documents this.

**2. Coverage — added.**
- `hz-gb-2312` body decode: new `decodeMailPayload` unit case **and** a new end-to-end `TestReadMailBody_DeclaredBodyCharsets` (readMailBody → decodeMailPayload → decodeWithCharset) covering both `gb2312` (plain 8-bit bytes on the GBK route) and `hz-gb-2312` (HZ escapes under their own label).
- `gb2312` body decode: a direct unit case already existed at head (`TestDecodeMailPayload_DeclaredCharsets`, from 9231d58e5); the new readMailBody test now covers the full body path too.
- `.msg` text output: new `TestEmailParser_MsgTextOutputExcludesMetadata` asserts `metadata`/`message_id`/`in_reply_to` stay out of the chunkable text while `subject:`/`from:` lines are still flattened.
- HTML walker `<table>`: a layout-table case already existed at head (`TestHTMLBodyToText` “layout table cells”); added a multi-row + surrounding-text case locking `intro\na\nb\nc\noutro` so cells across rows cannot fuse.
- Also dropped the “compatible superset” rationale from the gb2312 routing comments (the local repo guide asks not to add compatibility wording).

**3. multipart/alternative dedup — deferred as you suggested.** It is pre-existing duplication (both alternatives were already emitted before this PR; this PR only cleans the HTML markup), so I am keeping it out of this change. Happy to take it as a follow-up preferring the plain part.

**Verification:** no shared-service end-to-end run on this line — based on unit tests. `bash build.sh --test ./internal/parser/parser/...` passes for the full package including the new tests.

xugangqiang:
LGTM

PR #18593, opened 2026-08-21
feat[provider]: apply model_type infer when model has no type

Haruko386:
### Summary

As title.

### infer step:

Implemented model type inference fallback for provider ListModels results that do not include model_types.
Inference order is:
1. Use exact/alias metadata from all_models.json
2. Use strong name hints like **embedding**, **rerank**, **tts**, **asr**, **ocr**, **vision**
3. Use similarity fallback against known typed models
4. Fallback to chat

The similarity fallback uses a conservative score:

```math
0.25 \cdot \mathrm{token\_jaccard}
+ 0.45 \cdot \mathrm{series\_match}
+ 0.30 \cdot \mathrm{version\_similarity}
```

It only applies when the best score is above the baseline, avoids near-tie candidates with different types, and rejects capability conflicts such as matching qwen3 to qwen3-embedding. Model types are normalized so only chat + vision can coexist; all other types remain mutually exclusive.

> [!Note]
> O(N ^ 2), but acceptable, since there is a **little**, **tiny**, **small** chance that some provider has over 1,000 models


Lynn-Inf (approved):
LGTM

PR #18598, opened 2026-08-21
feat(channels): WeCom can bind agent

brucehu:
企业微信应用绑定除了绑定聊天还可以绑定自建的流程化的agent，消息在agent没有答案之前，先通过企业微信发送一个“正在思考”，有了答案再发送答案。


Yannnnnnny:
Thanks for the contribution! Could you also add documentation for this feature docs/guides/chatchannel/channel_access_and_configuration.md?

The documentation should mainly cover the configuration parameters required for binding a custom Agent to WeCom, including what should be entered for each parameter and, where applicable, where users can obtain the required values.

It would also be helpful to briefly explain how the "thinking" placeholder works so that users can properly configure and use this feature.

Thanks!

brucehu:
Hi, thank you for the review! I've added the documentation you requested to docs/guides/chatchannel/channel_access_and_configuration.md. It now covers:
How to bind a custom Flow Agent to a WeCom channel, including the Target type and Agent parameters, what to enter for each, and where the values come from (the agent list is loaded from the tenant's own flow agents, so no external credentials are needed).
The prerequisites for the agent (created under Agent > Flow agents, with its models configured for the tenant).
How the "thinking" placeholder works: WeCom first replies with "🤔 开始思考..." and then sends the final answer, with <think> reasoning content stripped from the reply.
The PR branch has been updated with this change. Please let me know if anything else is needed. Thanks!

docs/guides/chatchannel/channel_access_and_configuration.md
```diff
@@ -177,7 +177,35 @@ Usage:
...
+1. Open **User settings > Chat channels**, find the WeCom channel, and click the link icon to open the connect dialog.
+2. In the connect dialog, set **Target type** to `Agent`.
+3. Select the Flow Agent from the **Agent** dropdown and click confirm.
+4. Send a test message to the bot in WeCom and verify the reply.
+
+### Thinking placeholder
+
+To keep the end user informed while a completion is still running, the WeCom channel replies to every inbound message with a short `🤔 开始思考...` placeholder first, and then sends a second message with the final answer once the Chat or Agent has finished. This applies to both Chat and custom Agent bindings.
+
+The placeholder is only sent on the WeCom channel. The final reply is also cleaned up before it is sent: reasoning content (the `<think>` blocks produced by the model) is removed, so end users only see the final answer instead of the internal thinking process.
```
Yannnnnnny: The current wording may be a bit too absolute, as not every inbound message necessarily receives both a placeholder and a final reply. For example, blank messages may be ignored, and a final reply is only sent when the Chat or Agent produces a result.

I suggest narrowing the description to processed messages. For example:

> To keep the end user informed while a completion is still running, the WeCom channel sends a short `🤔 开始思考...` placeholder when an accepted message begins processing. Once the Chat or Agent has finished, a second message with the final answer is sent if a completion result is available.

Also, is there an English version of the `🤔 开始思考...` placeholder? Since this is the English documentation, I suggest using the English status here if it is supported, for example:

> `🤔 Thinking...`

This would keep the UI text in the example consistent with the language of the documentation.

buua436 (commented):
I found two issues in this PR:

1. `useFetchAllAgentList()` now requests only `page_size: 100` but never loads subsequent pages. As a result, users cannot bind a channel to agents beyond the first 100, despite the hook being intended to return all agents. Please either paginate through all pages or provide a dedicated API for the full authorized agent list.

2. The WeCom webhook now schedules `channel.handle_decrypted_message(msg)` with `asyncio.create_task()` but does not retain a strong reference to the task. asyncio keeps only weak references, so a long-running LLM task may be garbage-collected before it sends the answer. Please track pending tasks and remove them via a done callback.

web/src/pages/user-setting/chat-channel/hooks.ts
```diff
@@ -178,19 +179,26 @@ export const useFetchChatChannelDetail = () => {
...
 
-// Connect (or disconnect) a chat channel to an assistant (dialog).
-export const useConnectChatChannelDialog = () => {
+// Connect (or disconnect) a chat channel to an assistant (dialog) or an agent.
+export const useConnectChatChannelTarget = () => {
   const queryClient = useQueryClient();
 
   const { mutateAsync, isPending } = useMutation({
-    mutationKey: ['connect-chat-channel-dialog'],
+    mutationKey: ['connect-chat-channel-target'],
```
cike8899, line 188: It is recommended to use a factory function to facilitate reuse.

web/src/pages/user-setting/chat-channel/component/connect-dialog-modal.tsx
```diff
@@ -77,18 +106,57 @@ const ConnectDialogModal = ({
...
-          onChange={(val: string) => setDialogId(val || undefined)}
-          options={options}
-          allowClear
-          placeholder={t('setting.selectDialog')}
-        />
-        <p className="text-xs text-text-secondary/70 mt-1">
-          {t('setting.connectDialogTip')}
-        </p>
+        <div className="flex gap-2 bg-bg-card rounded-md p-1">
+          <button
```
cike8899: It is recommended to use existing button components.

web/src/pages/user-setting/chat-channel/component/connect-dialog-modal.tsx
```diff
@@ -77,18 +106,57 @@ const ConnectDialogModal = ({
...
+          >
+            {t('setting.targetAgent')}
+          </button>
+        </div>
+        {targetType === 'dialog' ? (
+          <>
+            <label className="text-sm text-text-secondary">
+              {t('setting.selectDialog')}
+            </label>
+            <RAGFlowSelect
```
cike8899: It is recommended to use SelectWithSearch; the current component has a bug.

brucehu:
Thanks for the review! I've addressed all the comments — added canvas_category validation, fixed the think-marker stripping, kept strong references to background tasks, paginated the agent list, and updated the UI components and docs. Please take another look.

PR #18601, opened 2026-08-21
Fix: allow yiyan api key

Lynn-Inf:
### Summary

As title, allow to use one-line api_key for Baidu Yiyan.


wangq8 (commented):
Reviewed the diff. This cleanly fixes the one-line API key (access token) path for Baidu Yiyan.

- The three model classes now fall back to `qianfan.{ChatCompletion,Embedding,Reranker}(access_token=...)` when the key isn't a JSON dict, which is the correct qianfan API for a bearer token.
- The `try/except` around `json.loads` also handles a key that's already a dict (the `TypeError` path), so both the legacy `{"yiyan_ak", "yiyan_sk"}` form and the new plain-token form keep working.
- Removing the `verify_api_key` normalization is consistent: the stored key stays the raw string and the model now parses it correctly, so verification and runtime behavior match.

Minor, non-blocking:

1. The JSON-parsing/fallback logic is now duplicated across `chat_model.py`, `embedding_model.py`, and `rerank_model.py`. Extracting a small shared helper (e.g. a module-level `_parse_yiyan_key(key) -> (ak, sk, access_token)`) would avoid the three copies drifting apart.

2. Edge case: a key that is valid JSON but neither a dict nor a string (e.g. a list/number) would be passed straight to `access_token=`. Unlikely in practice, but a guard raising a clear error would be friendlier than the SDK failing downstream.

Looks good to merge otherwise.

PR #18602, opened 2026-08-21
fix(parser): make PDF "remove TOC" and "remove header/footer" options work without DeepDoc inference

euvre:
### Summary

The PDF knowledge-base options **"remove table of contents"** (`removeTOC`) and **"remove header/footer"** (`removeHeaderFooter`) silently did nothing for most PDFs. This PR fixes both paths in the Go parser (`internal/parser/parser/pdf_postprocess.go`).

### Background / Root cause

Two independent bugs:

1. **TOC removal kept almost the whole table.**
   `removePDFTOC` inherited a Python parity quirk (`rag/nlp/__init__.py remove_contents_table`): after removing the TOC title it removed only the *first* entry, then scanned forward for a line sharing that entry's 3-byte prefix to stop at. When every TOC entry shares the same prefix (the common case — "Chapter One Introduction ...... 2", "Chapter Two Methods ...... 3", ...), the first body heading matches only at its position, but the scan hit later TOC entries first, or stopped immediately — in practice only the title and one entry were removed, and the remaining entries leaked into chunks.

2. **Header/footer removal was a no-op without a DeepDoc inference service.**
   `filterPDFHeaderFooter` only dropped sections whose `LayoutType` matched `header|footer|number`. Those layout types come from the DeepDoc inference service (`DEEPDOC_URL`). In default deployments there is no such service, every section stays `text`, and the option removed nothing at all — while the UI happily let users enable it.

### Fix

- **TOC**: after the title and first entry, consume the whole contiguous run of lines that still look like TOC entries via `isPDFTOCEntry` — dot leaders or ≥2 spaces followed by a trailing page number (`Chapter One .............. 2`, `Preface    7`, `1.1 Overview ……… 12`), requiring letter content so page-number-only footers are never mistaken for entries. When no entries match, the legacy prefix-scan fallback is preserved for pattern-less TOCs.
- **Header/footer**: added `removePDFRunningHeaderFooter`, which runs after the layout-type filter whenever `removeHeaderFooter` is on. It drops sections that:
  - sit entirely in the top 10% (header) or bottom 10% (footer) of their page (PDF point space, using `PageHeight`),
  - have a digit-normalized text (digits → `#`, so `- 2 -`, `- 3 -`, `Page 4 of 9` share a key) that repeats in the **same zone** on at least `max(2, pages/2)` pages,
  - and the document has at least 3 pages, and the section has no non-`text` layout type.
  
  Repetition across pages — not position alone — is the guard that keeps genuine body text; unique top-of-page lines survive.

### Tests

New unit tests in `pdf_postprocess_test.go`:

- `TestIsPDFTOCEntry` — pattern table (dot leaders, trailing page numbers, negative cases for body headings and page-number footers).
- `TestApplyPDFPostProcess_RemoveTOCRemovesAllDotLeaderEntries` — whole TOC table removed, body headings kept.
- `TestApplyPDFPostProcess_RemoveTOCLegacyPrefixScanWithoutEntryPattern` — pattern-less TOC keeps legacy behavior.
- `TestApplyPDFPostProcess_RemoveRunningHeaderFooterWithoutDLA` — running header + page-number footers removed with all sections typed `text` (no inference service), body preserved.
- `TestApplyPDFPostProcess_RunningHeaderFooterKeepsUniqueZoneText` — no false positives on unique top-of-page text.
- `TestApplyPDFPostProcess_RunningHeaderFooterSkipsShortDocuments` — 2-page documents untouched.

`bash build.sh --test ./internal/parser/...` passes.

### End-to-end verification

Verified against a local instance (Go ingestor + ES):

- Repro PDF: 4 pages, TOC page ("Contents" + 3 dot-leader entries), running `CONFIDENTIAL DRAFT HEADER` on pages 2–4, `- N -` page-number footers.
- Before: chunks contained TOC entries, headers, and footers with both options enabled.
- After: re-parsed with both options enabled → the single chunk contains only chapter headings and body paragraphs; ES query confirms no `Contents`/TOC entries/`CONFIDENTIAL DRAFT HEADER`/`- N -` remain.

Verification boundary: verified with the default deployment (no DeepDoc inference service). With `DEEPDOC_URL` configured, the pre-existing layout-type filter still applies first and this positional pass only acts on what remains.

### DIFF
--- a/internal/parser/parser/pdf_postprocess.go
+++ b/internal/parser/parser/pdf_postprocess.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"unicode"
 
 	deepdoctype "ragflow/internal/deepdoc/parser/type"
 )
@@ -15,6 +16,33 @@ import (
 // exact token, so a composite label like "page-footer" is also stripped.
 var pdfHeaderFooterPattern = regexp.MustCompile(`(?i)header|footer|number`)
 var pdfTOCTitlePattern = regexp.MustCompile(`(?i)^(contents|目录|目次|table of contents|致谢|acknowledge)$`)
+
+// TOC entry patterns: dot leaders or wide whitespace followed by a trailing
+// page number, e.g. "Chapter One .............. 2" or "Chapter One    2".
+var (
+	pdfTOCDotLeaderPattern    = regexp.MustCompile(`(\.|…|·){3,}\s*\d{1,5}\s*$`)
+	pdfTOCTrailingPagePattern = regexp.MustCompile(`\s{2,}\d{1,5}\s*$`)
+	pdfTOCEntryNumberPattern  = regexp.MustCompile(`^\d+(\.\d+)*[\s\.、]`)
+)
+
+// isPDFTOCEntry reports whether a line still looks like a table-of-contents
+// entry (title text plus a trailing page number). Used to consume the whole
+// contiguous run of entries after the TOC title instead of only the first.
+func isPDFTOCEntry(text string) bool {
+	t := strings.TrimSpace(text)
+	if len(t) < 4 || len(t) > 200 {
+		return false
+	}
+	if !pdfTOCDotLeaderPattern.MatchString(t) && !pdfTOCTrailingPagePattern.MatchString(t) {
+		return false
+	}
+	for _, r := range t {
+		if unicode.IsLetter(r) {
+			return true
+		}
+	}
+	return pdfTOCEntryNumberPattern.MatchString(t)
+}
 
 type pdfPostProcessOptions struct {
 	outputFormat       string
@@ -40,6 +68,7 @@ func applyPDFPostProcess(result *deepdoctype.ParseResult, opts pdfPostProcessOpt
 	normalizePDFLayoutTypes(result)
 	if opts.removeHeaderFooter {
 		filterPDFHeaderFooter(result)
+		removePDFRunningHeaderFooter(result)
 	}
 	assignPDFDocTypeKeywords(result, opts.flattenMediaToText)
 }
@@ -58,6 +87,122 @@ func filterPDFHeaderFooter(result *deepdoctype.ParseResult) {
 	filtered := result.Sections[:0]
 	for _, s := range result.Sections {
 		if pdfHeaderFooterPattern.MatchString(strings.TrimSpace(s.LayoutType)) {
+			continue
+		}
+		filtered = append(filtered, s)
+	}
+	result.Sections = filtered
+}
+
+const (
+	// pdfHeaderZoneRatio / pdfFooterZoneRatio bound the page zones where
+	// running headers/footers live: sections sitting entirely in the top
+	// or bottom fraction of a page are candidates.
+	pdfHeaderZoneRatio = 0.10
+	pdfFooterZoneRatio = 0.90
+	// pdfMinPagesForHeaderFooter guards against false positives on short
+	// documents: cross-page repetition only means "running header/footer"
+	// when the document has enough pages.
+	pdfMinPagesForHeaderFooter = 3
+)
+
+var pdfRunningDigitPattern = regexp.MustCompile(`\d+`)
+
+// normalizePDFRunningText collapses whitespace and replaces digit runs so
+// that per-page variants of the same running header/footer ("- 2 -",
+// "- 3 -", "Page 4 of 9") share one comparison key.
+func normalizePDFRunningText(text string) string {
+	t := strings.Join(strings.Fields(text), " ")
+	t = pdfRunningDigitPattern.ReplaceAllString(t, "#")
+	return strings.ToLower(strings.TrimSpace(t))
+}
+
+// removePDFRunningHeaderFooter drops sections whose normalized text repeats
+// in the same page zone (top 10% or bottom 10%) on at least half of the
+// document's pages — running headers, footers, and page numbers. The
+// layout-type filter above only works when a DeepDoc inference service
+// (DEEPDOC_URL) types header/footer regions; in default deployments there
+// is none, every section stays "text", and 页眉页脚 removal silently does
+// nothing. This positional frequency pass makes the option work without an
+// inference service. Repetition — not position alone — is the guard that
+// keeps genuine body text: a section must recur on >= max(2, pages/2) pages
+// in the same zone to be removed.
+func removePDFRunningHeaderFooter(result *deepdoctype.ParseResult) {
+	sections := result.Sections
+	if len(sections) == 0 || len(result.PageHeight) == 0 {
+		return
+	}
+	pageSet := make(map[int]struct{}, len(result.PageHeight))
+	for i := range sections {
+		if len(sections[i].Positions) == 0 || len(sections[i].Positions[0].PageNumbers) == 0 {
+			continue
+		}
+		pageSet[sections[i].Positions[0].PageNumbers[0]] = struct{}{}
+	}
+	numPages := len(pageSet)
+	if numPages < pdfMinPagesForHeaderFooter {
+		return
+	}
+	type zoneKey struct {
+		header bool
+		text   string
+	}
+	keyPages := make(map[zoneKey]map[int]struct{})
+	keySections := make(map[zoneKey][]int)
+	for i := range sections {
+		s := &sections[i]
+		if layout := strings.TrimSpace(s.LayoutType); layout != "" && layout != deepdoctype.LayoutTypeText {
+			continue
+		}
+		if len(s.Positions) == 0 || len(s.Positions[0].PageNumbers) == 0 {
+			continue
+		}
+		page := s.Positions[0].PageNumbers[0]
+		pageHeight := result.PageHeight[page]
+		if pageHeight <= 0 {
+			continue
+		}
+		top, bottom := s.Positions[0].Top, s.Positions[0].Bottom
+		if bottom <= top {
+			continue
+		}
+		var isHeader, isFooter bool
+		switch {
+		case top <= pageHeight*pdfHeaderZoneRatio && bottom <= pageHeight*0.5:
+			isHeader = true
+		case bottom >= pageHeight*pdfFooterZoneRatio && top >= pageHeight*0.5:
+			isFooter = true
+		}
+		if !isHeader && !isFooter {
+			continue
+		}
+		norm := normalizePDFRunningText(s.Text)
+		if norm == "" {
+			continue
+		}
+		key := zoneKey{header: isHeader, text: norm}
+		if keyPages[key] == nil {
+			keyPages[key] = make(map[int]struct{})
+		}
+		keyPages[key][page] = struct{}{}
+		keySections[key] = append(keySections[key], i)
+	}
+	drop := make(map[int]struct{})
+	for key, pages := range keyPages {
+		n := len(pages)
+		if n < 2 || n*2 < numPages {
+			continue
+		}
+		for _, i := range keySections[key] {
+			drop[i] = struct{}{}
+		}
+	}
+	if len(drop) == 0 {
+		return
+	}
+	filtered := sections[:0]
+	for i, s := range sections {
+		if _, ok := drop[i]; ok {
 			continue
 		}
 		filtered = append(filtered, s)
@@ -167,6 +312,21 @@ func removePDFTOC(result *deepdoctype.ParseResult) {
 		sections = append(sections[:i], sections[i+1:]...)
 		if i >= len(sections) || prefix == "" {
 			break
+		}
+		// Consume the remaining contiguous TOC entries (dot leaders or a
+		// trailing page number). Python's remove_contents_table stops at the
+		// first line sharing the first entry's 3-byte prefix, which leaves
+		// most of the table behind whenever every entry shares that prefix
+		// (e.g. "Chapter One ...", "Chapter Two ...", ...). When entries were
+		// consumed this way, skip the legacy prefix scan and resume the title
+		// scan; otherwise fall back to it for pattern-less TOCs.
+		removedEntries := 0
+		for i < len(sections) && isPDFTOCEntry(sections[i].Text) {
+			sections = append(sections[:i], sections[i+1:]...)
+			removedEntries++
+		}
+		if removedEntries > 0 {
+			continue
 		}
 		for j := i; j < len(sections) && j < i+128; j++ {
 			if !strings.HasPrefix(sectionText(sections[j]), prefix) {
--- a/internal/parser/parser/pdf_postprocess.go
+++ b/internal/parser/parser/pdf_postprocess.go
@@ -313,6 +313,16 @@ func removePDFTOC(result *deepdoctype.ParseResult) {
 		if i >= len(sections) || prefix == "" {
 			break
 		}
+		// The first entry is formatted (leader run plus page number): the
+		// table is entry-shaped and the prefix scan below must not run —
+		// with a single formatted entry that scan would delete body text up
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Reviewed the parser changes and the new tests. This is a solid fix for both reported issues.

**TOC removal** — the new `isPDFTOCEntry` loop correctly consumes the contiguous run of dot-leader/trailing-page-number entries after the title, and falls back to the legacy prefix scan when no entry matches. The tests cover dot-leader entries, the pattern-less legacy path, and the negative cases (`- 2 -`, `Page 3 of 10`, single-space body text). Good.

**Running header/footer removal** — I verified against `internal/deepdoc/parser/type/types.go` and the pdfium parser that `PageNumbers[0]` and `PageHeight` map keys use the same 0-based page index, so the zone/height math is consistent (no off-by-one). The in-place `sections[:0]` filter is safe, and the `n >= max(2, pages/2)` repetition threshold plus the `>=3 pages` guard are a reasonable trade-off against false positives. The digit normalization (`- 2 -` → `- # -`) is a nice touch.

Minor, non-blocking observations:

1. `filterPDFHeaderFooter` runs before `removePDFRunningHeaderFooter`, so sections already dropped by the layout-type filter don't contribute to `numPages`. That slightly undercounts pages for the threshold in the `DEEPDOC_URL` case, but it biases conservatively (fewer drops), so it's fine in practice.

2. `pdfTOCTrailingPagePattern` (`\s{2,}\d{1,5}\s*$`) is fairly broad — a body line ending in a number after two or more spaces could match. It's scoped to the contiguous run right after a TOC title, so risk is low, but worth keeping in mind if you ever see over-removal.

3. Minor: the `isPDFTOCEntry` length guard rejects entries shorter than 4 chars (e.g. a 2-char section number with a page number). Probably acceptable, just flagging the boundary.

LGTM otherwise — approving.

xugangqiang:
## Review: PR #18602 — make PDF "remove TOC" and "remove header/footer" work without DeepDoc inference

I reviewed this offline (read-only, against `main`). I verified the coordinate system, the production wiring, and traced both root causes by hand. Summary below.

### 1. Problem statement — does it actually fix it?
- **TOC only removed the title + first entry.** Confirmed by tracing `removePDFTOC` on the dot-leader case: the legacy prefix-scan immediately re-matches the 2nd entry (`"Chapter Two … 3"` shares prefix `"Cha"`), so only the title + first entry are dropped and the rest of the TOC leaks. The new `isPDFTOCEntry` loop that consumes the whole contiguous run fixes this correctly (traced: whole table removed, body headings kept). Pattern-less TOCs fall back to the legacy prefix-scan, which is correct for them. ✅
- **Header/footer no-op without DLA.** Confirmed: `filterPDFHeaderFooter` only matches `LayoutType` `header|footer|number`, which only exist when `DEEPDOC_URL` is configured. The new `removePDFRunningHeaderFooter` uses page-edge position + cross-page repetition and needs no inference. ✅
- **Coordinate orientation (critical):** verified the coordinate system is **y-down (origin at top of page)** via `ResolvePageSpan` (`bottom > pageHeight` ⇒ page span) and `sortSectionsByPosition` (ascending `top` = reading order). So `top <= 0.10*H` is genuinely the header zone — no inverted detection. `PageNumbers[0]` and `PageHeight` keys are both 0-based (confirmed in `boxes_sections.go` / `parser.go`).
- **Wiring:** `RemoveTOC`/`RemoveHeaderFooter` are already plumbed from dataset config and `applyPDFPostProcess` is on the production path, so the new call is reached.

### 2. Better approach?
No superior "no-inference" alternative; positional frequency analysis is the standard heuristic and the choice is sound. DLA-typed filtering still runs first and the new pass is additive. The items below are tightening, not a redesign. The magic-number thresholds (`0.10`/`0.90`/`3`/`pages/2`) are reasonable defaults for this PR.

### 3. Test coverage & regression risk
Good coverage: `isPDFTOCEntry` positive/negative table, full dot-leader removal with body preserved, pattern-less fallback, no-DLA running header + page-number footer removal, unique zone text kept, short-doc skipped. `makePDFSection` signature matches → compiles.

Gaps worth adding:
- Regression test for **blank/image-only pages** (see CR #3) — current `numPages` undercounts pages without text sections, lowering the threshold and risking over-removal.
- Adversarial test: a body block in the top 10% that repeats across ≥ half the pages (verify the repetition guard doesn't kill real body text).
- Edge case: a body line immediately after the TOC title ending in `  <number>` (e.g. `"Results 42"`) could be swallowed by `isPDFTOCEntry`.

Regression risk is **low and conservative (bias toward under-removal)**:
- TOC: new loop only fires when `isPDFTOCEntry` matches; otherwise legacy behavior is unchanged.
- Header/footer: entirely new, but **opt-in** (only when the option is enabled) and guarded by zone + repetition ≥ half pages + ≥3 pages + skip non-`text`/empty layout. Mis-deletion requires identical text in the same page-edge zone across ≥ half the pages — rare.

### 4. Does merging improve the system?
Yes. Fixes two real user-visible bugs (options silently did nothing in default deployments). Opt-in, low-risk, well-tested. Recommend merging after the two tightening fixes below.

### 5. On the existing review comments
All three CodeRabbit comments are reasonable; wangq8's are non-blocking.
- **CR #1 (drop `legacy`/`fallback`/`Python-parity` wording):** reasonable and aligns with repo AGENTS.md (no compatibility wording). Fix: reword the comments at `pdf_postprocess.go:316-322` and the test comment at `:161-163` to describe behavior directly; no behavior change.
- **CR #2 (require the entire section within the 10% band):** reasonable, reduces false positives. Fix:
  ```go
  case top <= pageHeight*pdfHeaderZoneRatio && bottom <= pageHeight*pdfHeaderZoneRatio:   // header
  case top >= pageHeight*pdfFooterZoneRatio && bottom >= pageHeight*pdfFooterZoneRatio:    // footer
  ```
  Trade-off: a running header taller than 10% of page height may no longer be removed — acceptable (conservative miss).
- **CR #3 (`numPages = len(result.PageHeight)` + blank-page regression test):** reasonable, prevents over-removal on docs with blank/image-only pages (also resolves wangq8's observation #1). Fix: drop the `pageSet` counting and use `numPages := len(result.PageHeight)`; add the regression test. `PageHeight` is written for every processed page, so it is a more complete page count.
- **wangq8 #2 / #3:** valid boundary notes. Suggest lowering the `<4` length guard to `<2`; the broad `pdfTOCTrailingPagePattern` is inherent to the approach — add a one-line comment noting its scope (only the contiguous run right after the TOC title).

**Minor extra nit:** test case `{"Page 3 of 10", false}, // no letters before number` — the comment is inaccurate (`"Page 3 of 10"` does contain letters; it is `false` because it lacks a dot leader / 2+ spaces before the page number). Fix the comment.

xugangqiang:
## Problems found

1. **`numPages` undercounts pages with no text sections** (`removePDFRunningHeaderFooter`). It counts only pages that have at least one section (`pageSet` from `PageNumbers[0]`). On documents with blank/image-only pages the denominator is smaller, the `n*2 < numPages` threshold loosens, and content is more aggressively removed. Use `numPages := len(result.PageHeight)` (written for every processed page) and add a regression test with a blank/image-only page.

2. **Header/footer zone check is too permissive** (`removePDFRunningHeaderFooter`). A section starting in the top 10% but extending down to the vertical middle (`bottom <= pageHeight*0.5`) is still classed as a header candidate; symmetrically for footers (`top >= pageHeight*0.5`). A tall, repeated top-of-page block could be mis-removed. Require the *entire* section within the edge band:
   ```go
   case top <= pageHeight*pdfHeaderZoneRatio && bottom <= pageHeight*pdfHeaderZoneRatio:   // header
   case top >= pageHeight*pdfFooterZoneRatio && bottom >= pageHeight*pdfFooterZoneRatio:    // footer
   ```

3. **`isPDFTOCEntry` can swallow a body line right after the TOC title.** `pdfTOCTrailingPagePattern` (`\s{2,}\d{1,5}\s*$`) matches any line ending in two-or-more spaces + a number, so a genuine body line like `"Results 42"` directly after the TOC entries would be consumed as a TOC entry. Scope it more tightly or add a guard.

4. **`isPDFTOCEntry` length guard rejects short entries.** `len(t) < 4` excludes valid short entries (e.g. `"I 5"`, `"1 5"`). Lower the guard (e.g. `<2`).

5. **Test coverage gaps.**
   - No blank/image-only-page regression test (see #1).
   - No adversarial test proving the repetition guard keeps a genuine body block that sits in the top 10% and repeats across ≥ half the pages.
   - No edge-case test for #3.

6. **Misleading test comment.** In `TestIsPDFTOCEntry`: `{"Page 3 of 10", false}, // no letters before number` — the string *does* contain letters; it is `false` because it has no dot leader and no 2+ spaces before the trailing number. Fix the comment.

7. **Comments still frame the code via Python parity** (`removePDFTOC` new block and the new test): "legacy python-parity behavior", "fallback", "Python's remove_contents_table". Per repo convention (no compatibility wording), reword them to describe the behavior directly (e.g. "pattern-based TOC scan", "prefix scan").

euvre:
Thanks for the reviews! All three actionable findings are addressed in 7c1780a, and @wangq8's observations are answered below.

**1. `numPages` now counts every rendered page (pdf_postprocess.go:137)** — applied. `removePDFRunningHeaderFooter` previously counted only pages that produced at least one section, so on documents with blank or image-only pages the repetition threshold dropped below half of the document (e.g. a candidate on 2 pages of a 5-page doc with 3 text pages was removed at 40% coverage). `numPages` is now `len(result.PageHeight)`, which covers every parsed page (verified against `internal/deepdoc/parser/pdf/parser.go`: `PageHeight` is recorded per rendered page and is scoped to the requested page range, so ranged parsing is unaffected). Added regression test `TestApplyPDFPostProcess_RunningHeaderFooterCountsAllRenderedPages` covering both sides: 2-of-5 pages survives, 3-of-5 is removed.

**2. Header/footer candidates must sit entirely in the 10% page-edge zone (pdf_postprocess.go:166-169)** — applied. The half-page threshold conditions are removed: a header candidate now requires `bottom <= pageHeight*0.10`, a footer candidate `top >= pageHeight*0.90`. Sections that merely start (or end) near a page edge but extend into the page body can no longer be classified as running furniture. Added regression test `TestApplyPDFPostProcess_RunningHeaderFooterRequiresFullZone` pinning that such zone-crossing sections survive even when repeated on every page.

**3. Compatibility wording removed from comments (pdf_postprocess.go:311-315, pdf_postprocess_test.go:160-163)** — applied. The TOC prefix-scan comment and the test comment now describe the pattern scan and prefix scan directly, without "legacy"/"fallback"/Python-parity framing; the test function was renamed to `TestApplyPDFPostProcess_RemoveTOCPrefixScanWithoutEntryPattern` for the same reason. No behavior change.

**@wangq8's non-blocking observations:**
1. The `numPages` undercount you flagged (layout-filtered sections shrinking the denominator) is now moot — `numPages` no longer depends on which sections survive `filterPDFHeaderFooter`.
2. `pdfTOCTrailingPagePattern` breadth: acknowledged. It only applies to the contiguous run immediately after a TOC title, so over-removal risk stays low; keeping an eye on it.
3. The 4-char minimum in `isPDFTOCEntry`: accepted boundary; very short numeric-only TOC entries are left untouched, which biases toward keeping content.

Verification: `bash build.sh --test ./internal/parser/parser/...` passes (all existing + new tests). No shared-service end-to-end verification was run on this line; the assessment is based on the unit tests above plus code-level analysis.

xugangqiang:
@euvre 

## Re-review after 7c1780a — remaining problems

**1. (Major) Single formatted TOC entry → prefix scan can delete body text.** In `removePDFTOC` the first section after the title is removed *before* `removedEntries` is initialized. If that first entry is the only formatted one (dot leader / trailing page number), `removedEntries` stays 0 and the prefix scan runs with the first entry's 3-byte prefix, deleting everything up to the first later line sharing that prefix. Example: `["Contents", "Chapter One ..... 2", "Body A", "Body B", "Chapter Two"]` — "Body A" and "Body B" are both deleted because "Chapter Two" shares the `"Cha"` prefix. Fix: count the first entry with `isPDFTOCEntry` *before* removing it (a formatted first entry then sets `removedEntries` and skips the prefix scan; a pattern-less first entry still falls through to the prefix scan). Note the proposed regression test ("one formatted entry followed by body text that does not share its prefix") would **not** fail on the current code — with no later line sharing the prefix the scan deletes nothing. The test must include a later line that *does* share the prefix (e.g. a later "Chapter Two" heading) and assert the body between the TOC and that line survives.

**2. Misleading test comment still present.** `TestIsPDFTOCEntry` still has `{"Page 3 of 10", false}, // no letters before number` — the string *does* contain letters; it is `false` because it has no dot leader and no 2+ spaces before the trailing number. Fix the comment.

**3. Adversarial test still missing.** No test proving the repetition guard preserves a genuine body block that sits entirely in the top 10% and repeats on ≥ half the pages (the "repetition, not position alone" claim has no direct test; the existing `RequiresFullZone` test covers zone-crossing, not repeated-in-zone body text).

The other two previously flagged items (breadth of `pdfTOCTrailingPagePattern`; the `<4` length guard in `isPDFTOCEntry`) were acknowledged as acceptable trade-offs — agreed, no change needed.

xugangqiang:
@/tmp/pr18602_rereview.md

euvre:
Thanks @xugangqiang — all three items from your re-review are handled in 4a3fc1016 (verified with `bash build.sh --test ./internal/parser/parser/...`):

**1. Single formatted TOC entry → prefix scan deletes body text — applied (major).** Confirmed by tracing: `removePDFTOC` removed the first entry before `removedEntries` was initialized, so with only one formatted entry the prefix scan ran with the first entry's 3-byte prefix and cut everything up to the next line sharing it. The first entry is now checked with `isPDFTOCEntry` *before* it is dropped and counts toward `removedEntries`, so a formatted first entry suppresses the prefix scan; a pattern-less first entry still falls through to it exactly as before. Per your note, the regression test `TestApplyPDFPostProcess_RemoveTOCSingleFormattedEntryKeepsBody` includes a later "Chapter Two" heading that *does* share the "Cha" prefix and asserts the body between the TOC and it survives — I verified it fails on the pre-fix code with exactly your predicted symptom (`kept = [Chapter Two]`) and passes with the fix.

**2. Misleading test comment — applied.** The "Page 3 of 10" case comment now reads "no dot leader or wide gap before the page number", which is the actual reason it is false.

**3. Adversarial repetition-guard test — applied with one boundary stated.** A block whose text is *identical*, sits fully inside the top-10% zone, and recurs on ≥ half the rendered pages is by construction indistinguishable from a running header — there is no remaining signal to separate them, and your round-1 review accepted exactly that as the conservative trade-off ("mis-deletion requires identical text in the same page-edge zone across ≥ half the pages — rare"), so no test can assert its preservation. What *can* be directly tested — and was missing — is the actual claim "repetition, not position alone": the new `TestApplyPDFPostProcess_RunningHeaderFooterNeedsRepetitionNotPosition` pins (a) distinct body blocks fully inside the top 10% of every page survive (position alone never removes), and (b) repetition counts distinct pages, not occurrences — the same text twice on one page still covers a single page and stays below the threshold. If you have a concrete discriminator in mind for the identical-repeat case, I'm happy to look at it in a follow-up.

@wangq8 — thanks for the approval; your three non-blocking observations were answered in my earlier summary and nothing in this commit changes those trade-offs.

Verification: `bash build.sh --test ./internal/parser/parser/...` passes (existing + new tests, including a red/green run of the new TOC regression test). No shared-service end-to-end verification was run on this line; the assessment is based on these unit tests plus the code-level trace above.

xugangqiang:
LGTM

wangq8:
@euvre please resolve conflict.

PR #18604, opened 2026-08-21
fix(parser): require DeepDoc service for Go PDF parsing instead of silent layout-less fallback

euvre:
### Summary

**Background.** PDFs parsed by the Go ingestion pipeline lost all layout structure: tables and figures were flattened into garbled plain-text chunks. A table-rich price-list PDF that the Python pipeline turns into ~26 chunks (tables preserved as Table-typed chunks) came out of the Go pipeline as 284 unreadable Text chunks.

**Root cause.** `deepDocAnalyzerFromEnv()` in `internal/parser/parser/pdf_parser_common.go` silently returned a layout-less `MockDocAnalyzer` whenever `DEEPDOC_URL` was unset, the endpoint failed to construct, or the DeepDoc inference service was unreachable. With no real DLA/TSR/OCR backend, layout analysis was a no-op, so every page degraded to plain text with no table detection at all — and nothing surfaced the misconfiguration.

### Fix

- `deepDocAnalyzerFromEnv()` now returns `(DocAnalyzer, error)`:
  - `DEEPDOC_URL` unset → default to `http://localhost:9390`, the address the OSS DeepDoc server (`deepdoc/server/deepdoc_server.py`, `deepdoc` service in `docker/docker-compose.yml`) listens on.
  - Endpoint invalid or service unhealthy → return an explicit error telling the operator to start the DeepDoc server or point `DEEPDOC_URL` at a healthy instance.
- `parsePDFWithDeepDocOptions` propagates the error, so a PDF parse task fails loudly instead of silently producing degraded chunks. Layout analysis is a hard requirement for PDF parsing — silently dropping it only produces garbage downstream.
- `docker/README.md`: correct the `DEEPDOC_URL` entry, which previously claimed the parser "falls back to inline ONNX Runtime inference" when unset — no such fallback exists in the Go path.

### Tests

- Split the CGO fixture test by dependency:
  - `pdf_parser_fixture_cgo_test.go` (unit tier, `//go:build cgo`) now covers only the `plain_text` parse path, which needs no DeepDoc service.
  - New `pdf_parser_fixture_integration_test.go` (`//go:build cgo && integration`) holds the DeepDoc-backed JSON and markdown fixture tests, with a `requireDeepDocServer` skip helper that probes `DEEPDOC_URL` (default `http://localhost:9390`) and skips when no healthy service is reachable.

### Verification

- `bash build.sh --test ./internal/parser/parser/` — PASS (unit tier, no external services).
- `bash build.sh --test-integration -v -run Fixture ./internal/parser/parser/` against a real DeepDoc server on :9390 — PASS; both fixture tests run real DLA/TSR inference (~1.5s each, not skips).
- Parser-level check on a table-rich Chinese PDF: 4 layout items, 2 of them `doc_type_kwd="table"` with reconstructed `<table>` HTML.
- End-to-end via the web UI with the fixed ingestor: uploaded the same PDF to a fresh dataset (Built-in / General) and parsed — result is 3 chunks: 1 Text chunk + 2 Table chunks with the table structure preserved (previously: hundreds of garbled Text chunks, tables lost).

### Verification boundary

Verified against a locally started OSS DeepDoc CPU server (`deepdoc/server/deepdoc_server.py --port 9390`) with the repo's ONNX models under `rag/res/deepdoc`. No GPU paths exercised; the reporter's exact PDF is not attached to the issue, so an equivalent table-rich Chinese price-list PDF was used for reproduction.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -103,7 +103,7 @@ The [.env](./.env) file contains important environment variables for Docker.
 ### DeepDoc Vision Service (OSS)
 
 - `DEEPDOC_URL`
-  URL for the deepdoc vision API serving DLA (layout analysis), OCR (text detection/recognition), and TSR (table structure recognition). The `deepdoc` service in `docker-compose.yml` provides this endpoint; in Docker deployments set it to `http://deepdoc:9390`. When unset, the parser defaults to `http://localhost:9390`. PDF parsing fails with an explicit error when no healthy DeepDoc service is reachable, since layout analysis is required to detect tables and figures.
+  URL for the deepdoc vision API serving DLA (layout analysis), OCR (text detection/recognition), and TSR (table structure recognition). Required for the `deepdoc` PDF parse method: parsing fails with an explicit error when `DEEPDOC_URL` is unset or no healthy service is reachable, since layout analysis is required to detect tables and figures. The `deepdoc` service in `docker-compose.yml` provides this endpoint — set it to `http://deepdoc:9390` in Docker deployments, or `http://localhost:9390` when running `deepdoc/server/deepdoc_server.py` locally.
 
   > The OSS deepdoc service runs on CPU using ONNX Runtime models. No GPU required.
   > API endpoints: `GET /health`, `GET /model`, `POST /predict/dla`, `POST /predict/tsr`, `POST /predict/ocr`.
--- a/docker/README.md
+++ b/docker/README.md
@@ -103,7 +103,7 @@ The [.env](./.env) file contains important environment variables for Docker.
 ### DeepDoc Vision Service (OSS)
 
 - `DEEPDOC_URL`
-  URL for the deepdoc vision API serving DLA (layout analysis), OCR (text detection/recognition), and TSR (table structure recognition). The `deepdoc` service in `docker-compose.yml` provides this endpoint. Defaults to `http://deepdoc:9390`. When unset, the parser falls back to inline ONNX Runtime inference.
+  URL for the deepdoc vision API serving DLA (layout analysis), OCR (text detection/recognition), and TSR (table structure recognition). The `deepdoc` service in `docker-compose.yml` provides this endpoint; in Docker deployments set it to `http://deepdoc:9390`. When unset, the parser defaults to `http://localhost:9390`. PDF parsing fails with an explicit error when no healthy DeepDoc service is reachable, since layout analysis is required to detect tables and figures.
 
   > The OSS deepdoc service runs on CPU using ONNX Runtime models. No GPU required.
   > API endpoints: `GET /health`, `GET /model`, `POST /predict/dla`, `POST /predict/tsr`, `POST /predict/ocr`.
--- a/internal/deepdoc/parser/pdf/inference/client.go
+++ b/internal/deepdoc/parser/pdf/inference/client.go
@@ -225,14 +225,25 @@ func (c *Client) OCRRecognize(ctx context.Context, cropped image.Image) ([]pdf.O
 	return texts, nil
 }
 
+// healthProbeTimeout bounds the /health probe independently of the client's
+// 120s inference timeout, so an endpoint that drops packets (firewall, wrong
+// subnet) fails fast instead of stalling a parse for up to two minutes.
+var healthProbeTimeout = 5 * time.Second
+
 // Health checks whether the DeepDoc service is reachable.
 func (c *Client) Health() bool {
-	resp, err := c.httpClient.Get(c.baseURL + "/health")
+	ctx, cancel := context.WithTimeout(context.Background(), healthProbeTimeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.baseURL+"/health", nil)
+	if err != nil {
+		return false
+	}
+	resp, err := c.httpClient.Do(req)
 	if err != nil {
 		return false
 	}
 	resp.Body.Close()
-	return resp.StatusCode == 200
+	return resp.StatusCode == http.StatusOK
 }
 
 func (c *Client) post(ctx context.Context, endpoint string, imgData []byte, filename string, result interface{}, extraFields ...string) error {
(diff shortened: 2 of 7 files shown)


wangq8 (commented):
Thanks for this — the root-cause analysis is sound and the direction is right: silently returning a layout-less `MockDocAnalyzer` and degrading to flattened text chunks is worse than failing loudly. The error propagation, the `(DocAnalyzer, error)` signature change, and the test-tier split (unit vs. `cgo && integration`) are all clean and follow repo conventions. A few things worth considering before merge:

### 1. `defaultDeepDocURL = "http://localhost:9390"` is wrong for Docker deployments (please verify)

In Docker the DeepDoc service runs in its own container, reachable at `http://deepdoc:9390`, *not* `localhost`. The Go parser runs inside the ragflow container, so `localhost:9390` points back at the ragflow container itself. `docker/.env` even documents this (`# DEEPDOC_URL=http://deepdoc:9390`, commented out), and the README update you made correctly says "in Docker deployments set it to `http://deepdoc:9390`".

Consequence: for default Docker deployments that did **not** enable the `deepdoc` profile and did **not** set `DEEPDOC_URL`, the default `deepdoc` parse method now hard-fails every PDF with "unreachable at http://localhost:9390", where it previously produced (degraded) text. That's a real behavioral break for a common default setup, and the error message is misleading — `localhost` isn't the fix there, `DEEPDOC_URL=http://deepdoc:9390` is.

Suggestions (any one would address it):
- Make the default environment-aware (source/dev builds → `localhost:9390`, Docker → `deepdoc:9390`), or
- Don't assume a URL when `DEEPDOC_URL` is unset — return an error that says the var is unset and instructs the operator to start the DeepDoc server *and* set `DEEPDOC_URL`, or
- At minimum, surface both hostnames in the error message so a Docker operator isn't pointed at `localhost`.

### 2. Health probe can block up to 120s per parse

`Client.Health()` does `httpClient.Get(baseURL + "/health")` with the client's `120 * time.Second` timeout. For a misconfigured endpoint that drops packets (firewall, wrong subnet) rather than refusing the connection, each parse can stall ~120s before failing. This pre-existed for the set-`DEEPDOC_URL` path, but now also applies to the implicit default. A short health-check-specific timeout (e.g. 2–5s) would make the "fail loudly" behavior also fail *fast*.

### 3. JSON/Markdown output path loses default CI coverage

Moving `TestPDFParser_ParseWithResult_CGOFixture` / `...Markdown` to the `cgo && integration` tier (with `requireDeepDocServer` skip) means the JSON and Markdown conversion paths will now skip in CI unless a live DeepDoc server is present. The `plain_text` path keeps unit-tier coverage, which is good, but consider retaining a mock-based unit test that drives `parsePDFWithDeepDocOptions`/`pdfParseResultToJSONWithOptions` with an injected `MockDocAnalyzer`, so JSON/Markdown output is still exercised in the default test tier.

### Minor
- The error message references `deepdoc/server/deepdoc_server.py` (source-layout path), which is meaningless in Docker. The "or set `DEEPDOC_URL` to a healthy instance" clause covers it, but it reads a bit source-specific.

Nothing here is a correctness bug in the code itself — the diff compiles cleanly (I verified `deepdocpdf` is still used elsewhere in the file at `RenderPageToImage`, so the import stays), the `:=` shadowing is correct, and the `ParseResult{Err: err}` failure return matches the documented contract (zero payload fields on error). The main thing I'd want resolved before merge is #1, since it's a user-visible behavior change for default Docker deployments.

xugangqiang:
Will use in-process go inference service soon - so this PR will not be needed then.

euvre:
Thanks @wangq8 for the thorough review, and @xugangqiang for the heads-up on the in-process Go inference service — this PR then serves as the interim fail-loudly guard, and I'm happy to rebase or fold it once that lands.

Addressed everything from the review in cb043ad85:

**1. `defaultDeepDocURL = http://localhost:9390` wrong for Docker — applied (your option 2: no implicit default).** Verified your analysis: the `deepdoc` compose service is opt-in via the `deepdoc` profile and reachable at `http://deepdoc:9390` (`docker/docker-compose.yml`), `docker/.env` ships `DEEPDOC_URL` commented out, and the Go parser runs in the ragflow container where `localhost:9390` is never the service. `deepDocAnalyzerFromEnv` now treats an unset `DEEPDOC_URL` as a configuration error that names both deployment options (start `deepdoc/server/deepdoc_server.py` from source, or enable the `deepdoc` profile / set `DEEPDOC_URL=http://deepdoc:9390` in Docker) instead of guessing a URL. I went with this rather than an environment-aware default because there is no reliable in-process signal distinguishing source builds from Docker (`/.dockerenv` sniffing breaks under devcontainers/WSL) and it would keep two divergent defaults alive; it also matches the existing convention in `runLocalImageOCR` (`internal/ingestion/component/media_dispatch.go`), which errors when `DEEPDOC_URL` is unset. `docker/README.md` updated accordingly, and the "unhealthy" error now mentions the Docker hostname too (covers your minor note about the source-specific path).

**2. Health probe blocking up to 120s — applied.** `Client.Health()` now issues the GET with a context bounded by a dedicated `healthProbeTimeout` (5s, within your suggested 2–5s range) instead of inheriting the client's 120s inference timeout. Added `TestDeepDocHTTP_HealthProbeTimesOutFast` with a hanging server to pin the fast-failure behavior.

**3. JSON/Markdown output losing default CI coverage — applied.** Added `TestPDFParser_ParseWithResult_CGODeepDocStub{JSON,Markdown}` in the default (`cgo`) tier: they point `DEEPDOC_URL` at an `httptest` server stubbing `/health` and the `/predict/*` endpoints, then run `ParseWithResult` on the Doc1.pdf fixture end to end. I used an HTTP stub rather than injecting a `MockDocAnalyzer` into `parsePDFWithDeepDocOptions` — the analyzer is resolved from env inside that function, and the stub exercises the real client + parse + serialization path without adding an injection seam. The JSON test also asserts `_pdf_positions` is a non-empty `[][]any` (it previously only failed on the empty-but-present case). The live-service fixture tests stay in the `cgo && integration` tier.

**Verification** (no shared-service end-to-end run on my side; based on unit tests + code-level comparison): `bash build.sh --test ./internal/deepdoc/parser/pdf/inference/... ./internal/parser/parser/...` passes in the default tier (including the new tests, run individually with `-run` to confirm they execute), and `bash build.sh --test-integration ./internal/parser/parser/...` compiles with the live-service tests skipping cleanly via `requireDeepDocServer`.

wangq8:
No need per @xugangqiang

PR #18605, opened 2026-08-21
fix: dataset ingestion log has no other type instead of success

Haruko386:
### Summary

As title

image

image

##### Filtered

image

image

### DIFF
--- a/internal/ingestion/service/ingestion_service.go
+++ b/internal/ingestion/service/ingestion_service.go
@@ -1010,47 +1010,13 @@ func (e *Ingestor) recordTerminalPipelineLog(ctx context.Context, ingestionTask 
 		return
 	}
 	ctx = context.WithoutCancel(ctx)
-	docTaskCtx, err := taskpkg.LoadFromIngestionTask(ctx, ingestionTask)
-	if err != nil {
-		common.Warn(fmt.Sprintf("record terminal pipeline log: load task context for %s: %v", ingestionTask.ID, err))
-		return
+	if err := taskpkg.RecordPipelineLog(ctx, dao.DB, taskpkg.PipelineLogInput{
+		KbID:       ingestionTask.DatasetID,
+		DocumentID: ingestionTask.DocumentID,
+		Status:     status,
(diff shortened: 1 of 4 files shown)


jay77721:
**Suggestion: remove `RecordPipelineLog` wrapper and share a plain function**

```go
func (s *PipelineExecutor) RecordPipelineLog(ctx context.Context, db *gorm.DB, docID, dsl, status string) {
    s.recordPipelineLog(ctx, db, docID, dsl, status)
}
```

This one-line exported wrapper exists only so `Ingestor.recordTerminalPipelineLog` ( `internal/ingestion/service/ingestion_service.go` ) can call the private `recordPipelineLog` across packages. To do that, the service has to construct a full `PipelineExecutor` via `taskpkg.NewPipelineExecutor(docTaskCtx, canvasID, 0)` just to write a log row, and re-implements DSL loading (`LoadBuiltinDSL` vs `dao.NewUserCanvasDAO().GetByID` + `json.Marshal`) that already exists in `defaultRunDocumentTask` / `loadDSLFromCanvas`.

Cleaner convergence: extract `recordPipelineLog` to a package-level helper that both sides can share, e.g.

```go
// internal/ingestion/task/pipeline_log.go or dao layer
func CreatePipelineLog(ctx context.Context, db *gorm.DB, doc entity.Document, tenant entity.Tenant, kb entity.Knowledgebase, dsl, status string) error
```

or let the service build `entity.PipelineOperationLog` directly and call `dao.NewPipelineOperationLogDAO().Create`. That removes the wrapper, removes the throwaway `NewPipelineExecutor` construction, and eliminates the duplicated builtin-vs-canvas DSL branch. `PipelineExecutor.Execute` and `Ingestor.recordTerminalPipelineLog` would then call the same plain function/DAO helper.

Happy to keep the current wrapper if you prefer minimal diff, but in that case please add a doc comment clarifying the `status == ""` fallback to `doc.Run` semantics.

jay77721 (approved):
LGTM

PR #18608, opened 2026-08-21
fix(web): hide empty thinking strip in floating chat widget

euvre:
## Problem

In the embedded floating chat widget (`/chats/widget`), assistant replies were preceded by a spurious, empty "Thinking..." strip whenever the stream carried a thinking section with no actual reasoning content.

### Root cause

When a chat completion streams `start_to_think` / `end_to_think` without any thinking content in between (e.g. the model answered directly without reasoning), the SSE accumulator in `useSendMessageWithSse` still emits the `<think></think>` markers. `replaceThinkToSection()` (`web/src/utils/chat.ts`) then unconditionally wrapped every `<think>...</think>` pair in a `<details class="think">` block — including empty ones — so the widget rendered a bare collapsible strip containing nothing but the "Thinking..." summary above the answer.

Reproduced live in the widget with a synthetic SSE stream:

```
data: {"data":{"answer":"","start_to_think":true}}
data: {"data":{"answer":"","end_to_think":true}}
data: {"data":{"answer":"Here is the answer."}}
```

Before the fix this rendered:

```html
<details class="think"><summary>Thinking...</summary></details>
Here is the answer.
```

## Fix

`replaceThinkToSection()` now drops think sections whose content is empty or whitespace-only instead of converting them into `<details>` blocks. Sections with genuine reasoning content are unchanged, including the custom-summary behavior used by the retrieval/"searching" call sites.

```ts
const result = text.replace(pattern, (_match, thinkContent: string) =>
  thinkContent.trim().length === 0
    ? ''
    : `<details class="think"><summary>${summary}</summary>${thinkContent}</details>`,
);
```

The guard lives in the shared util, so every markdown renderer that consumes `replaceThinkToSection` (floating widget, chat pages, agent log) benefits from the same fix.

## Tests

- Added a `replaceThinkToSection` suite to `web/src/utils/__tests__/chat.test.ts`:
  - drops an empty think section (`<think></think>answer` → `answer`)
  - drops a whitespace-only think section
  - keeps a non-empty think section as a `<details>` block
  - honors the custom `summary` argument for non-empty sections
  - leaves text without think markers unchanged
- All 5 new tests pass. (3 pre-existing `preprocessLaTeX` failures in the same file are unrelated — that function is untouched by this change, verified via `git diff`.)
- `oxlint` clean on changed files; `tsc --noEmit` reports no errors in changed files.

## Browser verification (widget, Vite HMR)

- Empty-think stream: no `<details>` element, no "Thinking..." text — previously rendered messages lose the strip too after HMR.
- Real-think stream (`<think>Let me think about this.</think>`): the collapsible "Thinking..." section with its content still renders intact.


wangq8 (commented):
Thanks for this PR. I reviewed the diff in full. Overall the web fix is correct and well-scoped, but the branch bundles an unrelated, substantial Go change that deserves to be split out.

## ✅ Web fix (`web/src/utils/chat.ts`)

The `replaceThinkToSection` change is correct: empty/whitespace-only `<think></think>` blocks are dropped instead of rendering a bare "Thinking..." strip. The guard lives in the shared util, so all callers (floating widget, chat pages, agent log) benefit. The 5 new tests cover the empty, whitespace-only, non-empty, custom-summary, and no-marker cases. Good.

## ⚠️ Scope — unrelated changes mixed in

The PR title/body describe only the web fix, but the diff also contains ~350 lines of unrelated changes to `internal/parser/parser/email_parser.go` (+ the new Chinese EML test file):

- deterministic field ordering (replacing randomized Go map iteration)
- RFC 2047 encoded-word decoding for headers/attachment filenames
- HTML body → visible-text extraction via `extractHTMLText`
- `Content-Transfer-Encoding` decoding for single-part bodies
- re-chunk fallback to the charset-decoded attachment payload

These are meaningful parser improvements and should be their own PR with their own title/description. Mixing them into a `fix(web)` PR makes review and bisecting harder. Please split.

## ⚠️ Behavioral change to indexed text

The text-output flatten now emits `text`/`text_html` as raw content (previously prefixed `text:` / `text_html:`), and HTML bodies are stripped to visible text instead of leaking raw markup. This changes the chunk text that gets indexed for existing KB email documents — same email re-parsed yields different chunks. That's arguably a correct fix (mirrors Python `rag/app/email.py`), but it's a real content change that isn't surfaced in the description. Please call it out explicitly (and consider a changelog note) so users re-chunking know content will differ.

## Minor

- `metadata` now emits with sorted keys and no trailing comma; empty metadata is skipped. Behaviorally fine, but again a quiet format change — worth a line in the eventual split PR.

## Nit (Go)

- `parseEmail`'s flatten no longer handles a `[]string` value case (previously `strings.Join(val, "\n")`). I don't see any field today that produces a `[]string`, so it's likely dead code, but confirm it can't regress any configured field before dropping it.

Nothing blocking on the web fix itself — it can land as-is once the parser changes are split out.

euvre:
Thanks @wangq8 for the thorough review, and for confirming the web fix itself is good to land.

## Scope — unrelated Go parser changes

Agreed, and addressed. A new commit (`56853db`) restores `internal/parser/parser/email_parser.go`, `email_parser_test.go`, and `email_parser_chinese_test.go` to their upstream main state, so this PR now contains **only** the web fix:

```
 web/src/utils/__tests__/chat.test.ts | 32 +++++++++++++++++-
 web/src/utils/chat.ts                |  9 +++--
```

(The parser changes were dropped via a revert-style commit rather than a history rewrite, to keep the branch fast-forward only.) The email parser improvements — deterministic ordering, RFC 2047 decoding, HTML visible-text extraction, CTE decoding, and the attachment retry — can be resubmitted as their own PR with a proper title/description.

## Items tied to the parser changes

- **Behavioral change to indexed text** (`text:`/`text_html:` prefixes, HTML stripped to visible text): no longer part of this PR. When the parser work is resubmitted separately, the description will call out explicitly that re-parsed email documents produce different chunk text, mirroring `rag/app/email.py`.
- **Minor — metadata sorted keys / no trailing comma**: same; will be noted in the split PR's description.
- **Nit — dropped `[]string` flatten case**: will be confirmed against every configured field before it lands in the split PR.

The CodeRabbit nitpick about compatibility wording in parser comments is likewise moot here, since those files no longer differ from main.

## Verification

- Net diff vs the PR base is the 2 web files only (`git diff` check); the Go parser tree is now byte-identical to upstream main, so existing parser behavior is untouched.
- `npm run type-check`: the only errors are the same 218 pre-existing ones present on the base tree (untouched files such as `user-setting/...`, `knowledge-service.ts`); none in `web/src/utils/chat.ts` or its test.
- No shared-service end-to-end verification was run on this line; the conclusion is based on the diff check above plus the existing unit tests (`web/src/utils/__tests__/chat.test.ts`) that cover the empty/whitespace-only/non-empty/custom-summary/no-marker cases.

dcc123456 (approved):

PR #18610, opened 2026-08-21
fix: inherit document status when inserting chunks

huang-aoqin:
## Summary
- Disabled documents (`status=0` / `enabled=false`) previously only synced `available_int` for already-indexed chunks. If a document was disabled before parsing, later chunk inserts still defaulted to `available_int=1` and remained retrievable.
- Stamp ordinary source chunks with `available_int=0` from document status at insert time (Go pipeline + Python task executor). Compiled products (`compile_kwd`) stay hidden as before.

## Test plan
- [ ] Disable a document before it has any chunks, then parse/chunk it
- [ ] Confirm listed chunks show `available=false` / `available_int=0`
- [ ] Confirm retrieval does not return those chunks
- [ ] Re-enable the document and confirm chunks become searchable again (existing status sync path)
- [ ] Unit: `TestApplyDocumentAvailability` and `test_document_availability.py`


xugangqiang:
**Duplicate import in `chunk_service.py`**

`DocumentService` is imported twice in `rag/svr/task_executor_refactor/chunk_service.py`:

```python
# First import (added by this PR, around line 44)
from api.db.services.document_service import DocumentService

# ... existing imports ...

# Second import (pre-existing, around line 45-46)
from api.db.services.document_service import DocumentService
from api.db.services.task_service import TaskService
```

**Suggestion:** Consolidate into a single import. Keep the existing import block and remove the new duplicate added by this PR:

```python
from api.db.services.document_service import DocumentService
from api.db.services.task_service import TaskService
from rag.svr.task_executor_refactor.constants import GRAPH_RAPTOR_FAKE_DOC_ID
``"

This is a trivial cleanup — functionally harmless, but worth fixing to keep the import block clean.

xugangqiang:
CI issue - seems like lefthook

huang-aoqin:
> CI issue - seems like lefthook

@xugangqiang done

xugangqiang:
LGTM. @yuzhichang  @qinling0210  pls have a second eye check.

PR #18611, opened 2026-08-21
fix(web): accept a single API key for Baidu YiYan

dcc123456:
### Summary

fix(web): accept a single API key for Baidu YiYan

The backend now allows a plain YiYan API key instead of an AK/SK pair, so the provider form no longer needs two separate credential fields.

### DIFF
--- a/web/src/pages/user-setting/setting-model/provider-schema/field-config/provider-config-map.ts
+++ b/web/src/pages/user-setting/setting-model/provider-schema/field-config/provider-config-map.ts
@@ -340,47 +340,25 @@ export const ProviderConfigMap: Record<string, ProviderConfig> = {
         validation: { message: 'instanceNameMessage' },
       },
       {
-        name: 'yiyan_ak',
-        label: 'addyiyanAK',
-        type: FormFieldType.Text,
+        name: 'api_key',
+        label: 'apiKey',
+        type: FormFieldType.Password,
         required: true,
-        placeholder: 'yiyanAKMessage',
+        placeholder: 'apiKeyMessage',
         shouldRender: 'hideWhenInstanceExists',
-        validation: { message: 'yiyanAKMessage' },
-      },
-      {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for this! The change correctly aligns the frontend with the backend's single access-token support — `BaiduYiyanChat` / `BaiduYiyanEmbed` / `BaiduYiyanRerank` now fall back to `access_token` when the stored key isn't a JSON object. The core change looks good, and switching the field to `Password` is a nice security improvement.

One thing I'd like you to double-check — backward compatibility for existing instances.

The PR removes `echoTransform` entirely. For instances saved in the legacy `{"yiyan_ak": ..., "yiyan_sk": ...}` JSON format, the generic echo fallback (`unwrapApiKey` in `web/src/pages/user-setting/setting-model/instance-card/hooks.tsx`) only recognizes a bare string or a literal `api_key` key inside the JSON object. For the `yiyan_ak` shape it returns `api_key: ''`.

As a result, opening and saving an existing AK/SK instance (e.g. just renaming it or changing `base_url`) will submit an empty `api_key` and silently clear the stored credential — the `api_key` field is hidden via `hideWhenInstanceExists` and is never re-populated from the legacy shape.

Suggested fix: keep a small `echoTransform` that maps the legacy object back to the single key, e.g.:

```ts
echoTransform: (instance) => {
  const obj = parseApiKeyAsObject(instance.api_key);
  if (obj && typeof obj.yiyan_ak === 'string') {
    return { api_key: obj.yiyan_ak };
  }
  return { api_key: typeof instance.api_key === 'string' ? instance.api_key : '' };
},
```

(or otherwise handle migration of legacy instances).

Minor / non-blocking: the now-unused i18n keys (`addyiyanAK`, `yiyanAKMessage`, `addyiyanSK`, `yiyanSKMessage`) remain in the locale files — could be cleaned up in a follow-up.

Otherwise LGTM.

PR #18614, opened 2026-08-21
Refactor: Refactored the frontend logic for handling different backend services to use `pickByBackend`, thereby improving code clarity and maintainability.

cike8899:
### Summary

Refactor: Refactored the frontend logic for handling different backend services to use pickByBackend, thereby improving code clarity and maintainability.


wangq8 (commented):
Overall direction looks good — centralizing backend branching behind `pickByBackend` / `BackendVariant` / `useIsGoBackend`, moving per-variant code into `go/`/`python/` subdirectories, and enforcing the boundary with `no-restricted-imports` is a clear improvement in clarity and maintainability. A few points to address:

1. **`preprocess` is silently dropped from `transformParserParams` (main concern).** The refactor removes `preprocess: cur.preprocess` from the per-file-format setup object in `web/src/pages/agent/utils.ts`. `preprocess` is a real parser field — the Go pipeline defaults carry `"preprocess": ["main_content"]`, and `PreprocessValue` (`main_content`/`title`/`abstract`/`author`) is still defined and still set in `initialParserValues`. It is also still forwarded on `main`. I realize the zod `FormSchema` has `preprocess` commented out, so it may already be stripped before reaching this transform (which would make this a no-op cleanup tied to the `Partial<…>` → `Omit<…, 'pages'>` type fix) — but please confirm the removal is intentional and not an accidental behavior change. If `preprocess` still flows through the form values anywhere, this silently changes what's sent to the parser backend.

2. **Minor: eager evaluation in `pickByBackend`.** For object/array variants (e.g. the final `return pickByBackend({ go: {…}, python: {…} })` in `transformParserParams` and the spread in `use-add-node.ts`), both branches are constructed on every call. This is fine at these sizes, and the thunk pattern is used correctly where the branches are async (`use-change-document-parser.ts`) — just flagging it so future non-trivial variants keep using thunks.

3. **Test still mocks the underlying store.** `web/src/pages/agent/utils/tests/extractor-transform.test.ts` was not updated and still `jest.mock`s `@/utils/backend-runtime` directly. It should keep working since `backend-variant.tsx` re-imports `getBackendLanguage` from the same mocked module, but worth a quick CI/test pass to confirm.

PR #18615, opened 2026-08-21
Fix(go): use llm setting for AI summary and return raw error msg

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/internal/handler/mindmap.go
+++ b/internal/handler/mindmap.go
@@ -72,10 +72,10 @@ func runMindMap(ctx context.Context, config mindMapRunConfig) (mindMapNode, erro
 	// composite-name parsing which fails for bare IDs. If the configured
 	// model can't be resolved, fall back to the tenant's default chat model
 	// (mirrors Python's gen_mindmap get_tenant_default_model_by_type).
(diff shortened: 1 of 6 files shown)


wangq8 (approved):
Solid change, well covered by tests.

- Extending `StreamingLLM.ChatStream` to also surface the driver's async error channel means a real provider error (e.g. 400/401) reaches the user instead of the generic "LLM call failed", which is a clear UX improvement.
- The buffered `errCh` (size 1) is correct: callers that discard the channel (e.g. `mindmap.go`) won't block the stream goroutine, and the `select` in `run` reads it cleanly after `ch` is drained.
- Threading `temperature`/`top_p` from `llm_setting` via `generationFloat` correctly honors the `{key}_enabled` flag (explicit `false` → fall back to the caller default).

One small thing to confirm (non-blocking): `generationFloat` returns `nil` when the flag is `true` but the value is absent, so AskService falls back to its own default (0.1 for temperature). Python's `resolve_llm_setting` instead substitutes the per-parameter `LLM_SETTING_DEFAULTS` value in that case. The behavior difference is minor, just want to make sure it's intentional.

PR #18616, opened 2026-08-21
fix(web): resize the mind map graph when its container changes size

dcc123456:
### Summary

fix(web): resize the mind map graph when its container changes size

G6 measures the container once at construction and never re-reads it, so the mind map canvas kept its initial pixel size while the sheet height tracks the viewport, clipping content or leaving blank space on window resize. The built-in `autoResize` only registers in the Graph constructor, and `autoFit: 'view'` only runs during `render()`, so neither covered this.

Observe the container and re-run `resize()` + `fitView()`, matching the existing MindMapG6Graph pattern.

### DIFF
--- a/web/src/components/indented-tree/indented-tree.tsx
+++ b/web/src/components/indented-tree/indented-tree.tsx
@@ -15,6 +15,7 @@
  */
 
 import { Graph as G6Graph, treeToGraphData } from '@antv/g6';
+import { useSize } from 'ahooks';
 import { useEffect, useMemo, useRef } from 'react';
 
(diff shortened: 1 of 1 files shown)


wangq8 (approved):
LGTM. Minimal, correct fix.

- `useSize` (already a dependency via `ahooks`) tracks the container dimensions, and the new effect re-runs `resize()` + `fitView()` only when the measured size actually differs from the graph's current size, matching the existing `mindmap-g6-graph` resize pattern.
- The `graph.destroyed` guard is consistent with the render effect already in this file, and the width/height early-return avoids a redundant resize on first measure.

No regressions spotted. Approving.

PR #18619, opened 2026-08-21
fix(redis): reclaim orphaned task messages

### LINKED ISSUE #15805: [Bug] Crashed-worker tasks are never reclaimed, and pending-message replay aborts on any non-exact Redis error

yzl0ng:
### Summary

Closes #15805.

Recover task-executor Redis stream messages that would otherwise remain pending after their worker disappears.

- isolate pending replay per queue so a transient error cannot skip later queues
- page through the full pending list and select idle entries owned by expired executors
- transfer the original pending entries to the live executor with Redis `XCLAIM`, preserving message IDs and payloads without copy/ACK Lua or deduplication markers
- feed claimed entries directly into the existing task-handling path before reading never-delivered (`>`) messages
- publish executor membership and heartbeat atomically with `MULTI`/`EXEC`, and skip recovery when publication fails
- add focused unit coverage plus real Redis-backed integration coverage for live/dead selection, same-ID ownership transfer, final ACK, already-ACKed races, and deleted stream entries

This revisits the abandoned #15806 on current `main` and incorporates its review findings around recovery correctness and singleton test isolation.

### Tests

- `9 passed` — focused pending replay/reclaim and heartbeat unit tests
- `2 passed` — real Redis 5.0.14.1 Streams integration tests
- `python -m py_compile` for all changed Python modules and tests
- `ruff check --select F,E9` for all changed Python modules and tests
- `ruff format --check` for the Redis module and focused tests
- `git diff --check`


xugangqiang (commented):
Focused review feedback on two points:

**1. Test coverage is insufficient.**
All new tests in `test/unit_test/rag/utils/test_redis_conn_pending.py` use `MagicMock`. The riskiest part — the Lua `LUA_REQUEUE_MSG_SCRIPT` — is never executed against a real Redis/Valkey. The "in-memory Redis/Lua integration probe" mentioned in the PR description is not committed as a test.

Key paths left untested:
- TOCTOU duplicate delivery: the original owner acks the entry between `get_pending_msg` listing it and the script's `XADD`, yet the script still copies the entry.
- Empty-stream data loss: if the stream was `XTRIM`med so `XRANGE` returns empty, the script still `XACK`s and silently drops the task.
- `reclaimed` over-counting: when the dedup marker already exists, `XADD` is skipped but the script still returns 1, so `reclaim_pending_msg` counts a reclaim that never happened.

Please add an integration test (real Redis/Valkey, or at minimum a Lua executor) covering the already-acked and empty-stream cases, plus live/dead consumer selection.

**2. The Lua script reinvents a Redis primitive; prefer `XCLAIM` / `XAUTOCLAIM`.**
`reclaim_pending_msg` copies the stream entry (`XADD`) and acks the old one. This hand-rolls ownership transfer that Redis already provides atomically via `XCLAIM` (or `XAUTOCLAIM`):
- `XCLAIM` moves the pending entry to a live consumer within the same group, keeping the same message id and PEL entry — idempotent, no duplicate delivery, no dedup marker, no custom Lua.
- It eliminates both the TOCTOU duplicate-delivery race and the empty-stream silent-drop, because there is no copy/ack split.

Suggested change: replace `requeue_msg` / `LUA_REQUEUE_MSG_SCRIPT` with a single `XCLAIM` call. For example, `reclaim_pending_msg` can call `self.REDIS.xclaim(queue_name, group_name, target_consumer, min_idle_ms, "0")` (optionally `JUSTID`) and iterate the returned ids. This is less code and strictly more correct.

xugangqiang:
@jay77721 pls help to review

jay77721:
## Review: test fixture crash on unavailable Redis

Thanks for the PR — the `XCLAIM`-based recovery is a clear improvement over the copy/ack approach. I ran the unit tests (9 passed) and the integration tests against a real Redis (2 passed), and found one bug in the integration test fixture that will bite CI.

**Location:** `test/integration/rag/utils/test_redis_conn_pending_integration.py`, the `redis_client` fixture.

**Issue 1 — wrong exception class: the fixture crashes with `AttributeError` instead of failing cleanly when Redis is unavailable.**

```python
try:
    client.ping()
except valkey.exceptions.RedisError as error:
    pytest.fail(f"RAGFLOW_REDIS_TEST_URL is unavailable: {error}")
```

`valkey.exceptions.RedisError` does **not exist** in the valkey client. I verified against the pinned version (`valkey==6.0.2` in `pyproject.toml`): `hasattr(valkey.exceptions, "RedisError")` is `False`. The correct name is `ValkeyError` — `RedisError` is the class name in the legacy `redis` package, and valkey dropped the alias.

Consequence: when the target Redis is unreachable or requires auth, `client.ping()` raises `AuthenticationError`/`ConnectionError`, which is not caught by the `except RedisError` clause. Python then raises `AttributeError: module 'valkey.exceptions' has no attribute 'RedisError'` while evaluating the handler, so the test dies with a confusing `AttributeError` instead of the intended `pytest.fail`. I reproduced this locally by pointing the fixture at an instance that requires a password.

```diff
- except valkey.exceptions.RedisError as error:
+ except valkey.exceptions.ValkeyError as error:
      pytest.fail(f"RAGFLOW_REDIS_TEST_URL is unavailable: {error}")
```

**Issue 2 — the skip path is also broken: these tests can never be skipped cleanly.**

If `RAGFLOW_REDIS_TEST_URL` is unset, `redis_client` calls `pytest.skip(...)`, but the dependent `redis_db` fixture still runs its setup against `None` and fails with `AttributeError`. So in any environment without a configured Redis the integration tests either error out during collection or skip-with-error — the exact situation a skip guard is meant to prevent. Consider scoping the skip so it stops the dependent fixture (e.g. skip at module/session scope, or make `redis_db` tolerate a skipped `redis_client`).

This matters for CI: any runner without a reachable Redis will show a hard failure instead of a clean skip, and issue 1 makes the failure message actively misleading.

Both are small fixes and would make the integration tests behave correctly in both configured and unconfigured environments.

jay77721 (changes requested):
Please address our review comments and the bot's feedback.

PR #18620, opened 2026-08-21
fix(web): give each provider instance card its own surface

dcc123456:
### Summary

fix(web): give each provider instance card its own surface

Instance cards were separated only by a bottom border, so the draft and saved variants read as one continuous list rather than distinct cards. Move the card surface onto each variant, and dim the collapsed instance name so the open one stands out.

### DIFF
--- a/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
+++ b/web/src/pages/user-setting/setting-model/instance-card/components/draft-mode-card.tsx
@@ -59,7 +59,7 @@ export function DraftModeCard({
   );
 
   return (
-    <div className="px-2 py-3 flex flex-col gap-4">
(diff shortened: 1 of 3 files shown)


wangq8 (approved):
Clean cosmetic change. Moving the card surface (`bg-bg-card rounded-xl`) onto each variant and dropping the now-redundant inner-row background + outer `border-b` makes the draft/saved variants read as distinct cards as intended.

Verified against the codebase:
- `bg-bg-card`, `text-text-primary`, `text-text-secondary`, and `cn` are all existing utilities used widely in `setting-model/`, so no new tokens are introduced.
- Dimming the collapsed instance name via `text-text-secondary` + `hover:text-text-primary` is a nice touch and matches the stated intent.

One trivial nit: trailing whitespace in the DraftModeCard `className` (`... rounded-xl `). Non-blocking.

LGTM, approving. ✅

PR #18624, opened 2026-08-21
fix(ingestion): coordinate task claims across ingestors

### LINKED ISSUE #18381: Ingestor dedup guard is process-local: multiple ingestor instances can duplicate-process a redelivered task
## Summary

The ingestor's duplicate-execution guard (`claimTask` / `releaseTask` over the in-process `currentTasks` map) is **process-local**. When more than one ingestor instance runs, a message that is redelivered (AckWait timeout or instance crash while in flight) can be fetched by a *different* instance and executed a second time, producing duplicate side effects (duplicated extracted memory entries, duplicated document chunks/embeds).

The current design implicitly assumes a **single ingestor instance**. It is not safe to scale out to multiple ingestor processes without addressing this.

## Background (current architecture)

- NATS consumer: a single **pull** consumer named `RAGFLOW_CONSUMER`, `FilterSubject: "tasks.>"`, `AckPolicy: AckExplicit`, `MaxDeliver: 16`, `MaxAckPending: 1024*128` (effectively unlimited), and **no `DeliverGroup`** (`internal/engine/nats/nats.go:206-212`). Messages are pulled via `consumer.Fetch` (`internal/engine/nats/nats.go:227-237`).
- One `consumeLoop` goroutine (`internal/ingestion/service/ingestion_service.go:186-217`) calls `GetMessages(4)`, then `processMessage` per handle, which enqueues to a bounded `taskChan` (cap `2*maxConcurrency`, `:126`); a worker pool executes and acks/nacks via the carried `TaskHandle`.
- Dedup: `processMessage` takes an in-memory claim via `claimTask(taskID)` (`currentTasks` map, `:60`, defined `:918`, mutex-protected by `tasksMu`) before enqueueing; on a redelivered task already claimed, it `ack`-skips (`:404-415`). `releaseTask` frees the claim when the worker settles (`:930`).

This guard is correct **within one process**, because a redelivery always lands on the same instance that already holds the claim.

## The bug (cross-instance)

When two or more ingestor instances share the same `RAGFLOW_CONSUMER` (they do — the consumer name is fixed), each message is delivered to exactly one `Fetch` under normal operation, so there is **no duplication in the happy path**. Duplication arises only on the redelivery path:

1. Instance A `Fetch`es task X, `StartRunning` flips it to RUNNING, `claimTask_A(X)` succeeds (A's in-memory map), enqueues, and A's worker begins a slow execution (e.g. LLM memory extraction or a large-document parse) that exceeds `AckWait` (default ~30s).
2. X is not acked in time → NATS redelivers X. The next `Fetch` may come from the *other* instance B (which is idle/fetching), so B receives the redelivered X.
3. B's `processMessage` calls `StartRunning(X)`: the task is already RUNNING, so `StartRunning` returns the task unchanged (`internal/service/ingestion_task_service.go:203`, `RUNNING` branch returns the task at `:213`) — it does **not** block.
4. B reaches the `RUNNING` case (`:404`), calls `claimTask_B(X)`: B's in-memory `currentTasks` does **not** contain X (the claim lives only in A's process), so it returns `true` → B enqueues and executes X **concurrently with A**.
5. X is executed twice → duplicate extracted memory entries (data duplication + extra LLM token cost) and/or duplicate document chunks/embeds.

### Root cause

- The dedup claim is **process-local memory**, not shared across instances.
- The DB `RUNNING` status set by `StartRunning` is **not an atomic execution lock**: two instances can both call `StartRunning` on the same task, both get a `RUNNING` task back, and both proceed to execute. The in-memory `claimTask` is the only thing that would stop the second one, and it cannot see the first instance's claim.

## Trigger conditions

- ≥ 2 ingestor instances running, **and**
- at least one task whose execution exceeds `AckWait` (default ~30s), e.g. LLM-backed memory extraction or a large/slow document parse, **or** an instance crash while a task is in flight.

Under load (slow tasks + a second idle instance fetching), the redelivery is very likely to be picked up by the other instance, making duplication realistic rather than edge-case.

## Impact

- Duplicated extracted memory entries in the store (and duplicate LLM calls / token cost).
- Duplicated document chunks/embeddings.
- Possible inconsistent `document.run` / task state.

## Suggested fixes (not yet implemented)

1. **Distributed claim** — replace the in-process `currentTasks` map with a shared lock keyed by task ID (Redis `SETNX`/Redlock, a DB advisory lock, or a unique constraint on an execution-claim row) so the claim coordinates across instances. This is the minimal change that preserves current behavior while making it multi-instance safe.
2. **Execution-layer idempotency** — make a second execution harmless (dedupe keys on inserted chunks/memory). Requires changing insert paths; more invasive.
3. **Operational constraint** — document and enforce a single ingestor instance (the current implicit assumption). Lowest effort, but blocks horizontal scaling of the ingestor.

## Related observation (not the bug, but worth noting)

The consumer's `MaxAckPending` is `1024*128` (`nats.go:210`), effectively unlimited. In the current single-instance design the real concurrency limit is the in-process `taskChan` (cap `2*maxConcurrency`) + blocking send, not NATS. If multi-instance support is pursued (option 1/2), the concurrency/backpressure model should be reconsidered together (e.g. per-instance `MaxAckPending` and/or a NATS QueueGroup), since a QueueGroup redelivers to any free member and would still require the distributed claim from option 1.

yzl0ng:
### Summary

Closes #18381.

Coordinate ingestion-task ownership across ingestor processes so a NATS redelivery cannot execute the same RUNNING task concurrently on another instance.

- add owner-scoped Redis leases with atomic NX acquisition, compare-and-expire renewal, and compare-and-delete release
- renew the lease from scheduling through settlement, including time spent waiting in the bounded worker queue
- start the NATS `InProgress` heartbeat before enqueue so queued messages do not exhaust redelivery attempts
- keep conflicting deliveries pending instead of Acking away the broker recovery path
- cancel a queued or running task with a distinct cause when ownership is lost, leaving it RUNNING and Nacking it for the successor
- preserve the process-local claim fallback for isolated paths without Redis

### Why NATS state is not the ownership source

I checked the JetStream API suggested in review. `ConsumerInfo` exposes aggregate state (`AckFloor`, `NumAckPending`, `NumPending`, and related counters), but it does not expose an ack-pending message list or the worker that currently owns each delivery. `Stream.GetMsg` can read a retained raw message only when its stream sequence is already known; it does not identify whether that message is actively executing or which ingestor owns it.

Therefore NATS remains the delivery/redelivery and liveness mechanism (`InProgress`), while the Redis value supplies the missing task-ID-to-owner lease needed for cross-instance exclusion. This matches the API surface in the pinned [nats.go v1.52.0 ConsumerInfo](https://github.com/nats-io/nats.go/blob/v1.52.0/jetstream/consumer_config.go#L27-L67) and [Stream.GetMsg](https://github.com/nats-io/nats.go/blob/v1.52.0/jetstream/stream.go#L47-L48).

### Review updates

- rebased onto upstream `main` at `61298397f`
- replaced deprecated `SetNX` with `SetArgs` in NX mode and preserved `redis.Nil` as a normal claim conflict
- captured heartbeat context, handle, and task ID before starting the goroutine so later `TaskContext.Ctx` replacement cannot race with it
- made non-positive refresh intervals derive safely from the lease TTL instead of immediately losing an acquired claim
- cleaned up all successful test claims and centralized the release timeout
- changed the conflict test to two independent ingestors sharing one claim store
- added concurrent local and 32-ingestor distributed contention tests; exactly one owner may win

### Tests

Passed locally:

- `go test ./internal/engine/redis -count=1`
- `go test -race ./internal/engine/redis -count=1` (Windows/amd64)
- Linux/amd64 race-compiled Redis test binary executed successfully
- 13 focused ingestion claim/heartbeat/process-message tests executed successfully on Linux/amd64 with CGO disabled
- remaining ingestion-service unit tests passed after skipping `TestStart_FullPathReturnsAndStartsWorkers`, which could not create its embedded JetStream stream because the local Docker Desktop WSL filesystem had insufficient storage
- Linux/CGO-off `go vet ./internal/engine/redis ./internal/ingestion/service`
- `gofmt -l` and `git diff --check`

The complete ingestion-service `-race` link could not be run locally: that package's CGO build requires the repository's office_oxide, pdfium, pdf_oxide, tokenizer, and PCRE2 native static-link toolchain. The new race-oriented tests are included for the repository/CI Linux native environment; this limitation is separate from the successfully executed focused Linux tests and Redis race runs.


xugangqiang:
@yzl0ng  Is it possible to read from NATS - the messages that are being processed？ Please also resolve conflicts and add more data race tests.

yzl0ng:
@xugangqiang Thanks. I resolved the conflicts and addressed the race-testing request in head `558a545`.

On NATS: with the repository-pinned nats.go v1.52.0, `ConsumerInfo` exposes aggregate pending/ack state but not a task-ID-to-current-worker list. `Stream.GetMsg` can retrieve a message only when its stream sequence is already known, and it still does not identify the active executor. Therefore NATS remains responsible for delivery, redelivery, and `InProgress` liveness, while the Redis owner-scoped lease provides cross-ingestor exclusion by task ID.

I also added concurrent ownership tests (including 32 independent ingestors sharing one store), a heartbeat-context race regression test, and cleanup/lease edge-case coverage. The Redis package passed native Windows `go test -race`; the focused ingestion tests passed on Linux/CGO-off. The newest CodeRabbit run is successful and all prior review threads are resolved.

Could you please re-review when convenient?

xugangqiang:
@yzl0ng   Thanks for the PR. We've several rounds of careful review within team across all the senior engineers. After discussion, we realized that the role of Redis is changed to cache-only in go side thus we do not want to have any other logic there. Thus we do not think redis (treated as simple key-value cache, not redis-cluster) is the most suitable way to solve this message-in-flight-state-checking issue.  Also I re-think from CAP theory perspective, may be there's no good solution to make sure hard-consistency while keep state in sync within distributed system as partition is always there and availability & consistency can't be archived at the same time in such situlation. For now, we may decide not to merge this PR - it's just a design decision.

PR #18626, opened 2026-08-21
docs(migration): say which volumes the backup actually covers

marmar9615-cloud:
### Summary

`migration.sh help` describes `backup` as a backup of **all** RAGFlow data volumes. It backs up four, fixed at `migration.sh:15`:

```sh
VOLUME_BASES=("mysql_data" "minio_data" "redis_data" "esdata01")
```

Nothing in the script reads `DOC_ENGINE`, and `esdata01` is only where the indices live when `DOC_ENGINE=elasticsearch`. `docker/.env` also offers `infinity`, `opensearch` and `serenedb`, whose indices live in `infinity_data`, `osdata01` and `serenedb_data` (`docker-compose-base.yml:405-409`). None of those are ever considered, and CI exercises the infinity path (`sep-tests.yml`), so this is not a hypothetical configuration.

What a user on one of those engines sees today:

1. `backup` finds no `<project>_esdata01`, prints `⚠️ Warning: Volume ... does not exist, skipping...`, then prints `🎉 Backup completed successfully!`. The warning names *Elasticsearch*, which is easy to read as "not applicable to me" rather than "your index is not in this backup".
2. Because that volume was skipped, `es_backup.tar.gz` is never written.
3. `restore` checks for every file in `BACKUP_FILES` first and exits on any that is missing (`migration.sh:228-235`), so the restore refuses to run at all.

So the backup does not round-trip for any non-default document engine, and that is only discovered at restore time, which is the worst moment to discover it.

This PR does not change what is backed up. It corrects the `all` claim and states the consequence where a reader is deciding whether this script is sufficient:

```
DOCKER VOLUMES (with default project name 'docker'):
  - docker_mysql_data     (MySQL database)
  - docker_minio_data     (MinIO object storage)
  - docker_redis_data     (Redis cache)
  - docker_esdata01       (Elasticsearch indices)

  Those four are the whole set. DOC_ENGINE chooses where your indices live,
  and only 'elasticsearch' keeps them in docker_esdata01. Under 'infinity',
  'opensearch' or 'serenedb' the index volume is not part of this backup, and
  because no es_backup.tar.gz is written, a later 'restore' stops on the
  missing file. Back up the index volume separately until that is handled.
```

Making the script engine-aware is the real fix, but it is a design decision (which volume to select per engine, what a cross-engine restore should mean, and that `oceanbase`/`seekdb` use bind mounts rather than named volumes) and it would touch the same `perform_restore` area as the open #18393 and #18433. I have filed that separately as #18628 rather than pre-empting it here; this PR is only the documentation half and conflicts with neither.

Verified against `upstream/main` at `e9cace9a`. `bash -n docker/migration.sh` passes and the rendered help is the block quoted above.


myf-bee (commented):
it is Looks Good To Me

PR #18627, opened 2026-08-21
fix(embed): drop_params: True was rejected by Together AI and other strict providers

### LINKED ISSUE #18584: [Bug]: Embedding request failed with Together AI: Unrecognized request arguments supplied: drop_params

Harsh23Kashyap:
Fixes #18584.

## Summary

`OpenAIEmbed._call` sent `extra_body={"drop_params": True}` to every OpenAI-compatible endpoint. `drop_params` is an OpenRouter-specific convention — it tells the server to silently drop unknown parameters. Together AI (and any other strict OpenAI-compatible provider) rejects the unknown field with HTTP 400 "Unrecognized request arguments supplied: drop_params".

The fix removes `drop_params` from `OpenAIEmbed._call`. `OpenRouterEmbed` has its own `_call` override that includes `drop_params` (and the optional `provider.order` block); that override is unchanged and continues to send `drop_params` to OpenRouter.

## Fix

One hunk in `rag/llm/embedding_model.py:OpenAIEmbed._call`:

```python
# extra_body is forwarded verbatim to the provider. `drop_params` is
# an OpenRouter-specific convention; Together AI (and any strict
# OpenAI-compatible provider) rejects it with HTTP 400
# "Unrecognized request arguments supplied: drop_params". Send only
# fields that every OpenAI-compatible provider accepts.
res = self.client.embeddings.create(input=batch, model=self.model_name, encoding_format="float")
```

All 8 `OpenAIEmbed` subclasses inherit `_call` and now also stop sending `drop_params`: `AzureEmbed`, `AstraflowEmbed`, `AstraflowCNEmbed`, `FuturMixEmbed`, `AIMLAPIEmbed`, `BaiChuanEmbed`, `TogetherAIEmbed`, `PerfXCloudEmbed`, `UpstageEmbed`. Together AI is one of those subclasses, so the issue's reproduction case is fixed in one place.

`OpenRouterEmbed._call` is independent (its own `extra_body` construction with the optional `provider.order` block) and is left unchanged.

## Test

Twelve regression tests in `test/unit_test/rag/llm/test_embedding_drop_params.py`:

- the main test pins `OpenAIEmbed._call` to not pass `extra_body` at all. Pre-fix, the captured kwargs show the exact bug: `extra_body={'drop_params': True}`.
- parametrized over 8 `OpenAIEmbed` subclasses (8 cases): each subclass's `_call` is invoked, the captured kwargs are asserted to not contain `drop_params`. TogetherAIEmbed is the issue's case; the rest guard against the same regression for the other inherited providers.
- regression guard: `OpenRouterEmbed._call` must still send `drop_params=True` so OpenRouter's silent-drop behaviour continues to apply. Independent of `OpenAIEmbed._call`, so the fix must not disturb it.
- pin for `OpenRouterEmbed._call`'s `provider.order` block alongside `drop_params` when `provider_order` is set.
- sanity: removing `drop_params` does not change the response shape — the call still returns embedding vectors from `res.data` and the total token count.

Verified pre-fix behaviour: the main test fails on the unpatched function with the assertion naming the captured kwargs verbatim. With the fix applied, all 12 cases pass. Existing 23 tests in `test_embedding_model.py` still pass. ruff check + format clean.

## Risks

- `drop_params=True` was originally added in PR #15424 as part of a refactor of `OpenAIEmbed._call`; the field pre-dates that PR and was carried forward. Removing it changes the wire-level request to every OpenAI-compatible embedding endpoint, but only removes a field that strict providers reject and that lenient providers (OpenAI, vLLM, OpenRouter when not configured for drop) ignore.
- The fix does not introduce an opt-in mechanism for users who depend on `drop_params` reaching a specific provider. If a future provider requires it, the right place to add it is in that provider's own class (as `OpenRouterEmbed` does) rather than in the base `OpenAIEmbed._call`.

### DIFF
--- a/rag/llm/embedding_model.py
+++ b/rag/llm/embedding_model.py
@@ -267,7 +267,12 @@ class OpenAIEmbed(Base):
         self.model_name = model_name
 
     def _call(self, batch):
-        res = self.client.embeddings.create(input=batch, model=self.model_name, encoding_format="float", extra_body={"drop_params": True})
+        # extra_body is forwarded verbatim to the provider. \`drop_params\` is
+        # an OpenRouter-specific convention; Together AI (and any strict
+        # OpenAI-compatible provider) rejects it with HTTP 400
+        # "Unrecognized request arguments supplied: drop_params". Send only
+        # fields that every OpenAI-compatible provider accepts.
+        res = self.client.embeddings.create(input=batch, model=self.model_name, encoding_format="float")
         return [d.embedding for d in _sorted_by_index(res.data)], total_token_count_from_response(res)
 
     def encode(self, texts: list):
--- a/test/unit_test/rag/llm/test_embedding_drop_params.py
+++ b/test/unit_test/rag/llm/test_embedding_drop_params.py
@@ -1,0 +1,182 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for issue #18584.
+
+``OpenAIEmbed._call`` sent ``extra_body={"drop_params": True}`` to every
+OpenAI-compatible endpoint. ``drop_params`` is an OpenRouter-specific
+convention -- it tells the server to silently drop unknown parameters.
+Together AI (and any other strict OpenAI-compatible provider) rejects the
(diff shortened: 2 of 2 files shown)


Lynn-Inf (approved):
LGTM

JinHai-CN:
@Dexterity104 Would you please help review this PR, as well?

PR #18632, opened 2026-08-21
fix(ingestion): honor auto-keyword/auto-question top_n sliders in extractor prompts

euvre:
## Summary

The dataset **auto keyword** / **auto question** count sliders (stored as `top_n` in `parser_config`) were ignored during parsing. The extractor sent its default system prompts — which contained a hardcoded "top N" count — to the LLM verbatim, so generation always followed the prompt text instead of the slider.

## Root cause

- The Go extractor component loaded the keyword/question prompts from `parser_config` and sent them to the chat model unchanged. The default prompts said e.g. "give the top 5 keywords", and the model obeyed the text.
- `top_n` from the slider was never injected into the prompt, so changing the slider had no observable effect.

## Changes

- `internal/ingestion/component/extractor.go`: render every `{{ topn }}` placeholder in the keyword/question prompts with the configured `top_n` before invoking the chat model; default prompts now use the `{{ topn }}` placeholder instead of a fixed count.
- `internal/ingestion/component/extractor_test.go`: unit tests for the placeholder rendering.
- `web/src/locales/{en,zh,ko,bg}.ts`: default extractor prompts use `{{ topn }}`.
- `web/src/pages/agent/form/extractor-form/use-switch-prompt.ts`: the agent canvas (Python runtime) has no `top_n` input and sends prompts verbatim, so the prompt switcher pins `{{ topn }}` to the legacy fixed counts (keywords 5, questions 3) to preserve its existing behavior.

## Verification

End to end on a live stack (GLM chat model + TEI embedding + Elasticsearch):

1. Created a dataset, set auto-keyword `top_n=2` and auto-question `top_n=1` via the chunk-method dialog sliders; confirmed the values persisted in `parser_config`.
2. Parsed a test document. Resulting chunk: `question_kwd` had exactly **1** question (hardcoded default would have produced more).
3. Re-ran with `top_n=5` / `top_n=3`: the same chunk then carried exactly **5** keywords and **3** questions.

The generated counts now track the slider values.


wangq8 (commented):
Thanks for this fix — the root cause is clearly identified and the change is well scoped.

I verified the change against the codebase:

**Correctness**
- The Go extractor previously sent the frontend pre-filled `system_prompt` verbatim, so the hardcoded "top 5"/"top 3" text always won over the `top_n` slider. Rendering `{{ topn }}` before the chat call fixes that, and it is consistent with the existing `rag/prompts/*.md` convention (which `rag/prompts/generator.py` already interpolates).
- The Python canvas `Extractor` (`rag/flow/extractor/extractor.py`) does send `sys_prompt` verbatim with no `top_n` input, so pinning `{{ topn }}` back to 5/3 in `use-switch-prompt.ts` is the right call to preserve its current behavior. Confirmed `field_name` only takes `summary|keywords|questions|metadata`, so the `legacyTopN` keyed lookup only hits the two prompts that contain the placeholder.

**Tests**
- The two new Go tests cover custom prompts (both spaced `{{ topn }}` and unspaced `{{topn}}`), the built-in defaults, and assert no leftover `{{` placeholders. Keyword-then-question ordering inside `autoExtractionJob` is deterministic for a single chunk, so the `reqs[0]`/`reqs[1]` assertions are safe.

**Minor / non-blocking**
1. Backward compatibility: datasets that already persisted the old pre-filled prompt (with the hardcoded "top 5"/"top 3" text) in `parser_config.system_prompt` won't have a `{{ topn }}` placeholder to substitute, so the slider still won't take effect for those until the prompt is reset (or cleared to fall back to the built-in default). Worth a note in the changelog/PR description so users know to reset the prompt.
2. `runAutoKeywords`/`runAutoQuestions` now `strings.TrimSpace` the custom prompt before sending (previously only trimmed for the emptiness check). Harmless and arguably nicer, but it is a slight behavior change for prompts with intentional leading/trailing whitespace.

No security or style concerns; the package-level `extractorTopNPattern` matches the existing `toolCallRE`/`nonRetryableStatusRE` convention.

Looks good to me otherwise.

euvre:
Thanks for the reviews! All items are addressed in commit 8b30b88ae (pushed).

**Applied**

1. **Own-property check for the top_n lookup** (CodeRabbit, inline): `value in legacyTopN` also matched inherited `Object.prototype` members, so a persisted `field_name` like `toString` could enter the branch and inject an invalid value into the prompt. Replaced with `Object.hasOwn(pythonCanvasTopN, value)` (ES2022, already used in `web/src/utils/llm-util.ts`). Behavior for the valid keys (`keywords` → 5, `questions` → 3) is unchanged; any other value now reliably falls back to the unchanged prompt. Note: the UI select only offers `summary|keywords|questions|metadata`, so this is defense-in-depth for hand-edited/imported canvas DSLs rather than a reachable UI bug.

2. **Comment wording** (CodeRabbit nitpick): reworded the comment above the lookup to describe only the current behavior — the Python canvas Extractor has no `top_n` input and sends `sys_prompt` verbatim, so its keyword/question prompts pin `{{ topn }}` to fixed counts — and renamed the constant from `legacyTopN` to `pythonCanvasTopN` to drop the compatibility wording per the repo guidelines.

**Notes for @wangq8's non-blocking points**

1. *Backward compatibility:* agreed — datasets that already persisted the old pre-filled prompt (with hardcoded "top 5"/"top 3") in `parser_config.system_prompt` have no `{{ topn }}` placeholder to substitute, so the slider takes effect for them only after the prompt is reset/cleared (falling back to the built-in default). Leaving this here as the user-facing note since I can't edit the PR description.
2. *`TrimSpace` on custom prompts:* acknowledged; it's intentional — the prompt sent to the LLM is now consistent with the emptiness check, and leading/trailing whitespace carried no semantics.

**Verification**: `npm run type-check` — zero errors in the touched file (the 211 reported errors are all in files this PR doesn't touch and pre-exist). No shared-service end-to-end verification was run for this frontend-only hardening; the behavior analysis above is based on the code paths (`rag/flow/extractor/extractor.py` sends `sys_prompt` verbatim with no `top_n` input, as confirmed in the review).

wangq8:
LGTM.

PR #18634, opened 2026-08-21
fix(migration): index volume must match the configured DOC_ENGINE

### LINKED ISSUE #18628: [Bug]: migration.sh backup misses the index volume for every non-elasticsearch DOC_ENGINE, so restore refuses the backup it just made
### Self Checks

e9cace9a

### RAGFlow image version

e9cace9a (main)

- Hardware parameters: n/a, this is in the migration script rather than the running service
- OS type: any
- Others: `DOC_ENGINE` set to anything other than `elasticsearch`

### Actual behavior

`docker/migration.sh` backs up a fixed set of four volumes (`migration.sh:15`):

```sh
VOLUME_BASES=("mysql_data" "minio_data" "redis_data" "esdata01")
```

Nothing in the script reads `DOC_ENGINE`. `esdata01` only holds the indices when `DOC_ENGINE=elasticsearch`; `docker/.env` also offers `infinity`, `opensearch` and `serenedb`, whose indices live in `infinity_data`, `osdata01` and `serenedb_data` (`docker-compose-base.yml:405-409`).

The result is that backup and restore do not round-trip on any non-default document engine:

1. `backup` finds no `<project>_esdata01`, prints `⚠️  Warning: Volume ... does not exist, skipping...`, and then prints `🎉 Backup completed successfully!`.
2. Because that volume was skipped, `es_backup.tar.gz` is never written.
3. `restore` verifies every entry of `BACKUP_FILES` up front and exits on any missing file (`migration.sh:228-235`), so it refuses to run against the backup the script just produced.

So the failure surfaces at restore time, and the only warning pointed at *Elasticsearch*, which a user who deliberately chose Infinity has every reason to read as "not applicable to me".

### Expected behavior

`backup` covers the index volume belonging to the configured `DOC_ENGINE`, and a backup taken by this script can be restored by it.

### Steps to reproduce

```markdown
1. Set `DOC_ENGINE=infinity` in `docker/.env` and start the stack, so `docker_infinity_data` exists and `docker_esdata01` does not.
2. Stop the stack, then run `bash docker/migration.sh backup`.
   Observe the skip warning for `docker_esdata01` followed by "Backup completed successfully!",
   and that the backup folder contains no `es_backup.tar.gz` and nothing from `docker_infinity_data`.
3. Run `bash docker/migration.sh restore`.
   It exits with "Missing backup files: es_backup.tar.gz" without restoring anything.
```

### Additional information

Some notes from reading the surrounding code, in case they are useful for whoever picks this up:

- The engines are not symmetric. `elasticsearch`, `opensearch`, `infinity` and `serenedb` use named volumes, but `oceanbase` and `seekdb` use bind mounts (`docker-compose-base.yml:134`, `:161`), so they need a different mechanism rather than another array entry.
- `ob_data`, `seekdb_data` and `tei_data` are declared in the volumes block but referenced by no service, so they are not the answer for those two engines.
- Simply appending the other index volumes to `VOLUME_BASES`/`BACKUP_FILES` would make things worse rather than better: an Elasticsearch user's backup would then be missing `osdata01`/`infinity_data`/`serenedb_data` archives, and the same up-front check at `:228` would reject their previously working backup.
- There is a cross-engine question worth deciding explicitly: whether a backup taken under one engine should be restorable under another, or refused with a clear message.
- #18393 and #18433 are already open against `perform_restore` for the separate issue #18354. Any fix here would touch nearby code, so ordering may matter.
- I have opened a documentation-only PR (#18626) that corrects the "all RAGFlow data volumes" wording in `show_help` and states this limitation, so users are not surprised at restore time while this is open. It deliberately does not change behavior.

Harsh23Kashyap:
Fixes #18628.

## Summary

`docker/migration.sh` had a hardcoded `VOLUME_BASES=("mysql_data" "minio_data" "redis_data" "esdata01")` and matching `BACKUP_FILES` list. The fourth entry (`esdata01` / `es_backup.tar.gz`) only matches the default `DOC_ENGINE=elasticsearch`; for `opensearch`, `infinity` or `serenedb` the named volume does not exist, the script prints the volume-skip warning and "Backup completed successfully!", and produces a backup set the same script then refuses to restore ("Missing backup files: es_backup.tar.gz").

## Fix

Reads `DOC_ENGINE` from the environment (default `elasticsearch`) and selects the correct index volume / backup file per engine via `get_index_volume_for_engine` in the new `docker/migration_volumes.sh` helper:

| `DOC_ENGINE` | index volume | backup file |
| --- | --- | --- |
| `elasticsearch` | `esdata01` | `es_backup.tar.gz` |
| `opensearch` | `osdata01` | `os_backup.tar.gz` |
| `infinity` | `infinity_data` | `infinity_backup.tar.gz` |
| `serenedb` | `serenedb_data` | `serenedb_backup.tar.gz` |
| `oceanbase` / `seekdb` | (skipped) | (skipped) |
| (unknown) | `esdata01` (fallback) | `es_backup.tar.gz` (fallback) |

`oceanbase` and `seekdb` use bind mounts (`./oceanbase/data:/root/ob` and the seekdb equivalent in `docker-compose-base.yml`), not named volumes, so the script cannot back up the index data. The function returns empty stdout for those engines and the script drops the index entry from its arrays — `VOLUME_BASES` has 3 entries (mysql, minio, redis) and the restore gate at the missing-files check does not complain about a phantom `es_backup.tar.gz`. The script prints a single-line warning so a user with a bind-mount engine knows to copy `docker/oceanbase/data` separately.

`docker/migration_volumes.sh` is a pure helper sourced by both the script and the test. The fix does not change the three engine-independent entries (`mysql_data`, `minio_data`, `redis_data`).

## Test

Ten regression tests in `test/unit_test/deploy/test_migration_volumes.py`:

- parametrized over 4 named-volume engines: `get_index_volume_for_engine` returns the correct `<volume_base> <backup_file>` pair.
- parametrized over 2 bind-mount engines (`oceanbase`, `seekdb`): function returns empty stdout so the script drops the index entry.
- unknown engine falls back to elasticsearch.
- default (no argument) maps to elasticsearch.
- 2 integration tests on the script's `help` output (Docker-gated, skipped when Docker is not available): the help text documents the per-engine mapping, and the pre-fix universal "docker_esdata01 (Elasticsearch indices)" line is gone.

Verified pre-fix behaviour: with the helper file removed, all 8 unit tests fail with `missing helper: docker/migration_volumes.sh`. With the fix applied, 8 tests pass and 2 are Docker-gated. `bash -n` validates both shell scripts. ruff check + format clean.

## Risks

- `DOC_ENGINE` is read from the environment. A user who upgrades and continues to set `DOC_ENGINE=elasticsearch` (the default) sees no change in behaviour.
- A user who sets `DOC_ENGINE` to a value the helper does not know (`DOC_ENGINE=customengine`) falls back to the elasticsearch mapping — the same volume is checked, and if the user has not actually set up elasticsearch, the existing skip warning fires. The fallback is intentional: a typo should not silently disable the index backup.
- The `oceanbase` / `seekdb` skip path is a behaviour change. A user on those engines now sees a clear "index backup is not supported by this script" warning instead of a misleading "Backup completed successfully!" message. The script's help text documents the limitation.

## Related

`docker/.env` line 21: `DOC_ENGINE=${DOC_ENGINE:-elasticsearch}`. The script now honours this same env convention. The `infinity` / `serenedb` / `opensearch` / `oceanbase` / `seekdb` branches are pre-existing in `docker-compose-base.yml` (lines 124-161) and `.env`.

### DIFF
--- a/docker/migration.sh
+++ b/docker/migration.sh
@@ -9,11 +9,22 @@
 set -e  # Exit on any error
 # Instead, we'll handle errors manually for better debugging experience
 
+# Source the index-volume resolver. Sourced here so the rest of the
+# script can use ``get_index_volume_for_engine`` without re-defining it.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=./migration_volumes.sh
+source "${SCRIPT_DIR}/migration_volumes.sh"
+
 # Default values
 DEFAULT_BACKUP_FOLDER="backup"
 DEFAULT_PROJECT_NAME="docker"
-VOLUME_BASES=("mysql_data" "minio_data" "redis_data" "esdata01")
-BACKUP_FILES=("mysql_backup.tar.gz" "minio_backup.tar.gz" "redis_backup.tar.gz" "es_backup.tar.gz")
+# Index volume is engine-specific; see get_index_volume_for_engine in
+# migration_volumes.sh. VOLUME_BASES and BACKUP_FILES are populated from
+# the configured DOC_ENGINE in main(), after the env is read.
+VOLUME_BASES=("mysql_data" "minio_data" "redis_data")
+BACKUP_FILES=("mysql_backup.tar.gz" "minio_backup.tar.gz" "redis_backup.tar.gz")
+INDEX_VOL_BASE=""
+INDEX_BACKUP_FILE=""
 
 # Build volume names from project name and base names
 build_volume_names() {
@@ -54,7 +65,17 @@ show_help() {
     echo "  - ${DEFAULT_PROJECT_NAME}_mysql_data     (MySQL database)"
     echo "  - ${DEFAULT_PROJECT_NAME}_minio_data     (MinIO object storage)"
     echo "  - ${DEFAULT_PROJECT_NAME}_redis_data     (Redis cache)"
-    echo "  - ${DEFAULT_PROJECT_NAME}_esdata01       (Elasticsearch indices)"
+    echo "  - <index volume>                       (DOC_ENGINE indices; see below)"
+    echo ""
+    echo "INDEX VOLUME BY DOC_ENGINE:"
+    echo "  elasticsearch -> ${DEFAULT_PROJECT_NAME}_esdata01       (file: es_backup.tar.gz)"
+    echo "  opensearch     -> ${DEFAULT_PROJECT_NAME}_osdata01       (file: os_backup.tar.gz)"
+    echo "  infinity       -> ${DEFAULT_PROJECT_NAME}_infinity_data  (file: infinity_backup.tar.gz)"
+    echo "  serenedb       -> ${DEFAULT_PROJECT_NAME}_serenedb_data  (file: serenedb_backup.tar.gz)"
+    echo "  oceanbase / seekdb use bind mounts; index backup is not supported by this script."
+    echo ""
+    echo "DOC_ENGINE is read from the environment (default: elasticsearch). To back up"
+    echo "an Infinity deployment, run e.g.  DOC_ENGINE=infinity $0 backup  ."
     echo ""
     echo "NOTE:"
     echo "  If you started RAGFlow with 'docker compose -p myproject up', the volume"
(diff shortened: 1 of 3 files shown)


JinHai-CN:
@deadtrickster would you please help review the PR, thank you.

deadtrickster:
Looking

qinling0210 (approved):
LGTM

myf-bee (commented):
@Harsh23Kashyap   Can you remove unit_text? Because we don't want to introduce too many unit tests in CI ，thanks

deadtrickster:
on removing tests - I would politely object - do not remove tests themselves, just don't add them to the CI matrix.

myf-bee:
@Harsh23Kashyap   This is an auxiliary script tool with relatively low usage frequency. We need to consider only introducing necessary tests in the CI.  please remove unit_text,  Thank you for your understanding.

PR #18635, opened 2026-08-21
fix(manual): distinguish legacy .doc from .docx and fail early

### LINKED ISSUE #18621: [Bug]: Manual parser routes legacy .doc files to python-docx as DOCX
### Self Checks

v0.26.4 official Docker image

### RAGFlow image version

infiniflow/ragflow:v0.26.4

```Markdown
- Host OS: Linux
- Parser/chunking method: Manual
- Input format: legacy Microsoft Word `.doc`
- Python version used for the local reproduction: Python 3.12
- Both reproduced files are OLE/CFB compound documents
- File header: `d0cf11e0a1b11ae1`
```

### Actual behavior

`rag/app/manual.py` uses the following suffix expression:

```python
elif re.search(r"\.docx?$", filename, re.IGNORECASE):
```
Because x is optional, this expression matches both .doc and .docx.
Both formats are then passed to Docx(), which calls
python-docx.Document():
```
self.doc = Document(filename) if not binary else Document(BytesIO(binary))
```
However, legacy .doc files are OLE/CFB compound documents, while .docx
files are Office Open XML ZIP packages. python-docx does not parse legacy
.doc files.

### Expected behavior

The Manual parser must not pass legacy .doc files to
python-docx.Document().
The .docx branch should match .docx only. Because the Manual parser does
not currently have a legacy DOC implementation, .doc input should fail
early with a clear message asking the user to convert it to .docx or PDF.
It should not expose an internal BadZipFile or OPC relationship exception.

### Steps to reproduce

```Markdown
1.Select the Manual parser/chunking method.
2.Upload a real legacy Microsoft Word .doc file.
3.Confirm that its first eight bytes are the OLE/CFB signature:
d0 cf 11 e0 a1 b1 1a e1
4.Start parsing.
5.Observe that the file enters manual.Docx.
6.python-docx raises either BadZipFile or an OPC relationship KeyError.
```

### Additional information

The problematic expression was introduced by commit 50f209204 through
PR #4325.
A conservative fix is preferable here:
```
- elif re.search(r"\.docx?$", filename, re.IGNORECASE):
+ elif re.search(r"\.docx$", filename, re.IGNORECASE):
```
Legacy .doc support should not be added implicitly through the DOCX path.
Supporting it in Manual would require a separate parser implementation and
a decision about how Tika plaintext maps to Manual-specific hierarchy,
tables, images, and chunk semantics.
Related user report:
https://github.com/orgs/infiniflow/discussions/15239

Harsh23Kashyap:
Fixes #18621.

## Summary

The Manual parser's file-routing regex was `\.docx?$` which matches both `.doc` and `.docx`. Both formats were passed to `Docx()` -> `python-docx.Document()`, which only handles Office Open XML (.docx) and raises `BadZipFile` or an OPC relationship `KeyError` on legacy OLE/CFB .doc files. The error surfaced as a generic 500-style stack trace to the user.

## Fix

Two changes in `rag/app/manual.py:chunk()`:

1. The `.docx` branch now matches `\.docx$` (no optional `x`). Legacy `.doc` input no longer silently enters the python-docx path.
2. A new `\.doc` branch raises `NotImplementedError` with a clear message: legacy `.doc` is the pre-2007 Office format, the Manual parser is built on python-docx which handles `.docx` only, the user should convert to `.docx` or PDF and re-upload. The exception points at issue #18621 so a maintainer reading the trace can see the upstream tracking issue.

Legacy `.doc` support in the Manual parser is out of scope: it would require a separate parser implementation plus a Tika-plaintext-to-Manual mapping decision for hierarchy, tables, images, and chunk semantics. The conservative fix is to fail early with a clear message rather than expose an internal `BadZipFile` / `KeyError`.

## Test

Five AST-based regression tests in `test/unit_test/rag/app/test_manual_docx_routing.py`:

- `test_legacy_doc_routes_to_not_implemented_error`: `.doc` must match a `NotImplementedError`-raising branch.
- `test_docx_routes_to_docx_branch`: `.docx` must reach the `Docx()` instantiation.
- `test_pdf_routes_to_pdf_branch`: `.pdf` must reach the first branch (regression guard for the existing PDF path).
- `test_docx_branch_regex_does_not_match_legacy_doc`: the central invariant — the regex for the `Docx()` branch must not match `test.doc`.
- `test_legacy_doc_error_message_mentions_docx_and_pdf`: the user-facing `NotImplementedError` must name both `.docx` and `PDF` as conversion targets and reference `#18621`.

The tests use AST extraction rather than driving `chunk()` directly because the module pulls `deepdoc` -> `xgboost` -> `beartype` at import time and the dev venv does not have all of those installed. The AST approach pins the routing contract without needing the full import graph (the same shape as the cycle-21 parser output-format regression guard).

Verified pre-fix behaviour: the 3 fix-pinning tests fail with informative assertions naming the captured regex (`\\.docx?$`) and action kind (`docx`). The 2 regression-guard tests pass (as expected — the fix only changes the `.doc` branch). With the fix applied, all 5 pass. `ruff check` + `ruff format --check` clean.

## Risks

- A user who uploaded a `.doc` file before the fix saw a `BadZipFile` / `KeyError` stack trace; after the fix they see a `NotImplementedError` with a clear conversion instruction. The new error is at least as informative as the old one and points at the upstream tracking issue.
- A user who explicitly relied on `.doc` files being silently routed through the `Docx()` path (and somehow not crashing, e.g. by uploading a renamed `.docx` file) sees an explicit error instead of silent misbehavior. This is a strict improvement.
- The PDF and DOCX paths are unchanged. Backward compatible for any user who never uploaded a `.doc` file.

## Related

The `.docx?$` regex was introduced in commit `50f209204` ("Synchronize with enterprise version", PR #4325). The fix narrows the regex and adds a separate `.doc` branch.


JinHai-CN:
@Harsh23Kashyap I suppose this PR is already fix, please resolve the conflicts and check, thanks.

Issue #17202, opened 2026-07-21
chunker ignores chunk_token_num: soft cap routinely overshoots by one unit (and can blow past 100x)


skbs-eng:
## Summary

`chunk_token_num` is honored only as a soft target, not a hard cap. In `RAGFlowTxtParser.parser_txt` (`deepdoc/parser/txt_parser.py:36-47`) and in `rag.nlp.naive_merge` (`rag/nlp/__init__.py:1171-1193`, plus the `_with_images` mirror), the size check fires *after* the unit has already been appended to the chunk. As a result, every chunk can overshoot `chunk_token_num` by up to the size of one line / sentence / section. A subset of inputs (single very long lines with no internal delimiter) balloon into chunks an order of magnitude larger than the configured budget.

This issue is a more complete report of #12109 (config `chunk_token_num` ignored, chunks exceed rerank input size).

## Impact

Real-world dataset, 336 `.txt` files, 154,103 chunks in the Infinity index. Configuration: `parser_id=naive`, `layout_recognize=Plain Text`, `chunk_token_num=512`, `delimiter=\n`, `overlapped_percent=0.1`.

Distribution of stored chunk token counts (measured with `cl100k_base` against `chunk.content` in Infinity):

| Bucket (tokens) | Chunks | % of total |
|---|---|---|
| < 512 | 67,040 | 43.5% |
| 512–600 | 52,573 | 34.1% |
| 600–700 | 20,354 | 13.2% |
| 700–800 | 8,695 | 5.6% |
| 800–1000 | 5,775 | 3.7% |
| 1000–1500 | 262 | 0.2% |
| 1500–2000 | 19 |  |
| 2000–3000 | 4 |  |
| 3000–5000 | 1 |  |
| 10000–20000 | 1 |  |

56.5% of chunks exceed the configured 512-token budget. By characters, 99.9% exceed 512 chars (median 2,288 chars, mean 2,352). The largest outlier is 14,813 tokens / 60,293 chars in a single chunk.

Symptoms observed downstream:

- Rerank model failures on the >2048-token outliers (ref #12109).
- Inconsistent retrieval: queries that should land on a single paragraph split across several adjacent chunks instead embed one giant chunk.
- Embedding model truncation silently cuts the tail of each oversize chunk at retrieval time, so the stored chunks do not match what the embedder saw.

## Root cause

The repro is one-liner-level simple — patch the existing unit tests so `add_chunk` does its size check, and they fail because of a soft cap.

[...]

PR #18637, opened 2026-08-21
fix(nlp): naive_merge custom-delimiter path honours chunk_token_num

### LINKED ISSUE #18552: [Bug] naive_merge "custom delimiter" branch silently bypasses chunk_token_num, splitting on stray bare chars

Harsh23Kashyap:
Fixes #18552.

## Summary

When `parser_config.delimiter` contained a backtick-wrapped token, the `has_custom` branch in `naive_merge` (and `naive_merge_with_images`) split the section on the user's compiled pattern and then appended every segment as its own chunk — bypassing the `chunk_token_num` merge step entirely. For the issue's reproduction (a 2.8 MB text with `chunk_token_num=512` and a delimiter that leaks a single space from unbalanced backticks), this produced ~429,000 chunks of 1-2 tokens instead of the expected ~1,400. The Python string `parser_config.delimiter` API also diverged from the Go `CompileDelimiterPatternList(keepBare=False)` behaviour — Go drops bare chars outside backticks; the Python string keeps them. One stray bare char was enough to 300x the chunk count.

## Fix

Two changes in `rag/nlp/__init__.py`:

1. `naive_merge`: build the paragraphs list `(text, pos)` from the custom-delimiter split (same shape as the default path), then run the paragraphs through `_merge_paragraph_groups`, then `_reconstruct_text_chunk`, then `_apply_overlap_unconditional` — so `chunk_token_num` is honoured regardless of which path is taken.
2. `naive_merge_with_images`: same change with the `(text, pos, image)` tuple shape so images stay aligned with their paragraphs.

The split-on-custom-delimiter behaviour is preserved — the user's backtick-wrapped token still produces a chunk boundary, but a run of tiny segments now merges back up to `chunk_token_num` rather than producing hundreds of thousands of 1-token chunks. This is the issue's "Option A — minimum viable" fix; a future "boundary hints" semantics (the user's `##` is a hard boundary even when both paragraphs fit) is out of scope.

## Test

Five regression tests in `test/unit_test/rag/nlp/test_naive_merge_custom_delim.py`:

- `test_issue_repro_does_not_produce_300x_chunks`: a 200KB book with `chunk_token_num=512` and the issue's unbalanced-backtick delimiter must produce < 5,000 chunks (pre-fix would produce ~30,000 for the same input). The threshold is well above the observed post-fix count (~500-600) and well below the pre-fix blow-up.
- `test_chunk_size_respects_chunk_token_num_with_custom_delimiter`: the largest chunk must respect `chunk_token_num=128` (modulo the unconditional overlap prefix). Pre-fix produced 1-2 token chunks; post-fix fills each chunk up to the cap.
- `test_default_path_is_unchanged`: regression guard that the default (no-backticks) path is unchanged — the merge step must still enforce `chunk_token_num` for users who never had a custom delimiter.
- `test_naive_merge_with_images_honours_chunk_token_num_with_custom_delim`: the same fix must apply to `naive_merge_with_images`. The chunk and image lists must remain 1:1 and the chunk count must stay bounded.
- `test_has_custom_branch_uses_merge_paragraph_groups`: AST/source guard that pins the fix at the implementation level — the `has_custom` branch must funnel through `_merge_paragraph_groups` + `_reconstruct_text_chunk` + `_apply_overlap_unconditional` (the same shape as the default path). A future refactor that drops the merge step (re-introducing the bypass) is caught loudly.

Verified pre-fix behaviour: the issue-repro test fails with the assertion that the chunk count exceeds the post-fix ceiling. With the fix applied, all 5 pass. ruff check + format clean.

## Risks

- Behavior change: a user who relied on `##`-separated paragraphs being a hard chunk boundary (pre-fix) will see those paragraphs merge if both fit in `chunk_token_num`. The issue author calls this out as a "minimum viable" fix and lists the harder "boundary hints" semantics as a future improvement. The release notes should call this out.
- The split-on-wrapped-tokens is preserved. Only the post-split merge step is changed.
- Backward compatible for users who never had a custom delimiter: the default path is unchanged, and `test_default_path_is_unchanged` is the regression guard.

## Related

None of the open PRs in this area (#12109, #17202, #17203, #17275, #17692, #16959) address the custom-delimiter branch specifically, per the issue author's research.


xugangqiang:
I did a read-only review of this PR and verified the behaviour empirically against the merge-base (`3dab5db315`). Findings only:

## Blocker 1 — Breaks two existing contract tests (CI will go red)

`test/unit_test/rag/test_delim.py`:

* `test_naive_merge_wrapped_single_char_bypasses_chunk_token_num` (L408)
* `test_naive_merge_skips_empty_segments_from_adjacent_delimiters` (L422)

Both pass on the merge-base and both fail at the PR head:

```
assert ['aa\nbb\ncc'] == ['aa', 'bb', 'cc']   # test_delim.py:419
assert ['aa\nbb']     == ['aa', 'bb']         # test_delim.py:431
```

`2 failed, 124 passed` when running `test/unit_test/rag/`.

Neither test is updated by this PR, and they are not accidental: they were added by #17387 as deliberate contracts for the `has_custom` semantics. L409-410 reads *"`` `;` `` is a wrapped one-character delimiter; has_custom must still be true so each segment becomes its own chunk"*, and both pass `chunk_token_num=10**9` precisely to pin "custom delimiter = hard boundary, token cap ignored".

## Blocker 2 — Destroys the hard-boundary semantics that custom delimiters exist for

With the merge step applied unconditionally, a wrapped delimiter no longer marks a boundary at all — it only deletes its own text. Measured at the PR head:

```python
md = "\n".join(f"## Section {i}\n" + "word " * 40 for i in range(6))
naive_merge([md], chunk_token_num=512, delimiter="`##`")
# pre-fix:  6 chunks   (one per section)
# post-fix: 1 chunk    '\n Section 0\nword word word ...'
```

This is the primary use case of the feature (the intent of #11434), not a marginal edge case. The issue's justification for Option A — *"for a user who genuinely wants one segment per chunk, their segments are already short enough to fit chunk_token_num, so the merge is a no-op"* — is incorrect: short segments are exactly what the merge step combines.

## Blocker 3 — Root cause is untouched, so the emitted text is still corrupted

The cause named in the issue (root cause #1) is that `parse_delimiter_field` (`rag/nlp/delim.py:110-162`) treats bare chars leaking out of the unbalanced-backtick region as valid delimiters. This PR does not change that; the pattern still contains `' '`, and `rag/nlp/__init__.py:1482-1483` still drops every matched delimiter segment:

```python
if custom_pattern and re.fullmatch(custom_pattern, sub_sec):
    continue
```

Measured at the PR head with `ISSUE_DELIMITER`:

```
parsed_dels: ['  ', '\n\n', '. ', '\n', ' ']
in : 'alpha beta gamma delta epsilon zeta eta theta iota kappa'
out: ['\nalpha\nbeta\ngamma\ndelta\nepsilon\nzeta\neta\ntheta\niota\nkappa']
```

Every space in the source is deleted and replaced by `\n`. The fix turns "80,000 corrupted chunks" into "545 corrupted chunks" — the chunk count is bounded, but the text that reaches the embedding index is still wrong. This is what CodeRabbit flagged on 2026-08-21 (content-preservation); it is still unaddressed.

Related: Note 3 in the PR description compares Python's single-string `parser_config.delimiter` against Go's `CompileDelimiterPatternList(keepBare=False)`. These are two different APIs — `internal/parser/chunk/delim.go:80-83` states *"The single-string parser_config.delimiter field is not parsed in Go; only the []string list API is consumed."* The like-for-like comparison is Python-vs-Python: `rag/flow/chunker/token_chunker.py:97` keeps only wrapped tokens, while `rag/nlp/__init__.py` mixes bare and wrapped.

## Improvement 1 — Same bug is left unfixed in `naive_merge_docx`

`rag/nlp/__init__.py:1874`:

```python
if prev_text_ck < 0 or merged[prev_text_ck]["tk_nums"] >= chunk_token_num or has_custom:
```

Measured at the PR head: a 500-word DOCX section with `ISSUE_DELIMITER` and `chunk_token_num=512` still produces **500 chunks** (one per word). Same file, same bug class, fixed in 2 of 3 sites.

## Improvement 2 — Introduces a three-way Python/Go/DSL divergence

Go's chunker explicitly replicates the semantics this PR removes (`internal/ingestion/component/chunker/token.go:326-330`):

```go
// Custom (backtick-wrapped) delimiter: one chunk per segment, no token
// merge — mirrors Python naive_merge's has_custom branch (token_chunker.py:1194-1213).
if hasCustomDelim(c.param.Delimiters) {
    return c.chunkPerSegment(text, delimPattern, childrenPattern)
}
```

The Python DSL path does the same at `rag/flow/chunker/token_chunker.py:416`. After this PR, the legacy path (`rag/app/naive.py` -> `rag/nlp`) and the DSL/Go paths produce different chunk boundaries for the same delimiter input.

## Improvement 3 — `test_chunk_size_respects_chunk_token_num_with_custom_delimiter` cannot detect the regression it claims to guard

`test_naive_merge_custom_delim.py:96-102` asserts only an upper bound:

```python
max_tokens = max(_approx_tokens(c) for c in chunks)
assert max_tokens <= 256
```

Pre-fix, every chunk is 1-2 tokens, so `max_tokens` is ~2, which is `<= 256` — **the test passes with the bug present**. Its docstring claims "Pre-fix produced 1-2 token chunks; post-fix fills each chunk up to the cap", but no lower bound / fill-ratio is asserted.

## Improvement 4 — Source-text assertion is an anti-pattern

`test_has_custom_branch_uses_merge_paragraph_groups` (L161-181) uses `inspect.getsource` and checks for the substrings `_merge_paragraph_groups` / `_reconstruct_text_chunk` / `_apply_overlap_unconditional`. It cannot catch any real behaviour regression (behaviour changes while the names stay -> passes) and will false-alarm on any normal refactor. It also imports the private symbol `_merge_paragraph_groups`, and only covers `naive_merge` — `naive_merge_with_images` is unchecked.

## Improvement 5 — Hand-rolled token estimator instead of the real tokenizer

`_approx_tokens` (L49-55) uses `len(text) // 4`. The issue's acceptance criteria explicitly ask for the real `num_tokens_from_string`. Using a fake estimator forces the tolerance to 2x the cap (256 vs `chunk_token_num=128`), which further weakens every size assertion.

## Nit

* `rag/nlp/__init__.py:1468-1474` and the new test file (L16-34, L111-115, L153-182) are full of "The pre-fix behaviour ..." wording. AGENTS.md: *"Do not add new compatibility wording in comments or docs."*
* Two unused imports with `# noqa: F401` at `test_naive_merge_custom_delim.py:37` (`has_wrapped_delimiter`) and `:176` (`_merge_paragraph_groups`).
* Pre-existing but now more visible: `naive_merge` splits with `flags=re.DOTALL` (`:1478`) while `naive_merge_with_images` does not (`:1545`).

## Suggested direction

Fix the root cause rather than routing around it: make the `has_custom` branch split **only on the user's explicitly backtick-wrapped tokens** (`rag/nlp/__init__.py:1475` and `:1537`), matching `rag/flow/chunker/token_chunker.py:93-100` and Go's `keepBare=false`. That alone resolves #18552 — the stray `' '` never enters the pattern, no merge is needed, the text stays intact, the `##` hard boundary keeps working, and both `test_delim.py` contracts stay green.

If you do keep the merge, it needs to be paired with: (a) updating the two `test_delim.py` contracts after a maintainer confirms the semantics change is intended, (b) the same change in `_merge_cks` (`:1874`), (c) matching changes in `token.go:329` and `token_chunker.py:416`, (d) preserving the wrapped boundary as a forced break so `##` does not collapse into one chunk, and (e) a content-preservation assertion.

Also worth collapsing afterwards: once the split patterns match, the `has_custom` branch differs from the default path only in the `<8 token` `local_pos` handling (`:1486-1487`, `:1553-1554`) and the `re.DOTALL` flag — it becomes a near-duplicate that should be deleted rather than maintained in parallel.

PR #18395, opened 2026-08-17
fix(agent/tools): port DeepL to ToolBase so it works as an Agent tool

immuhammadfurqan:
### Summary

Closes #18394.

The **DeepL** agent tool (`agent/tools/deepl.py`) was never ported to the `ToolBase`/`_invoke` interface during the agent module redesign, so it is unusable as an Agent tool in two independent ways.

**1. Adding it to an Agent raises `AttributeError`.** `DeepLParam` extended `ComponentParamBase` and defined no `meta`, and `get_meta()` is defined on `ToolBase`/`ToolParamBase`, not on `ComponentBase` (`grep -n "def get_meta" agent/component/base.py` returns nothing). `agent/component/agent_with_tools.py:82` builds every tool's function descriptor via `cpn.get_meta()`, so constructing an Agent containing DeepL raised `AttributeError: 'DeepLParam' object has no attribute 'get_meta'`.

**2. It could never run — `_run` is dead code.** `DeepL` only implemented the legacy `_run(self, history, **kwargs)`. Nothing dispatches to `_run` any more (`grep -rn "\._run(" agent/` returns no callers); all five execution paths go through `_invoke` — `agent/component/base.py:406,432,590` and `agent/tools/base.py:154,179`. With no `_invoke` defined, `DeepL._invoke` resolved to `ComponentBase._invoke`, which raises `NotImplementedError` (`agent/component/base.py:447`).

This is the same class of defect as #16414 (Crawler) and #16416 (AkShare). #16329 fixed a *separate* bug in this file (`check()` validating an undefined `self.top_n`, and `_run` dropping its error return) but did not port the tool.

Note that `be_output` still exists (`agent/component/base.py:578`, returning `pd.DataFrame([{"content": v}])`) — it has not been removed — but nothing consumes its result for tools any more, which is why the legacy path is dead rather than merely deprecated.

### What changed

- Port `DeepLParam` to `ToolParamBase` with a `ToolMeta` defined **before** `super().__init__()` (matching `ArXivParam`/`AkShareParam`, since `ToolParamBase.__init__` reads `self.meta` immediately), exposing a required `query` parameter — the text to translate, default `{sys.query}`, consistent with the convention shared by the other tools.
- `source_lang` / `target_lang` stay component configuration rather than becoming LLM-settable tool parameters, matching the existing `DeepLSourceLangOptions` / `DeepLTargetLangOptions` dropdowns in `web/src/pages/agent/options.ts` — the same split as `Crawler`'s `extract_type`.
[...]

PR #18640, opened 2026-08-21
fix(agent/tools): port Jin10 and TuShare from ComponentBase to ToolBase

### LINKED ISSUE #18447: [Bug]: Jin10 agent tool is non-functional — no get_meta (crashes Agent) and only implements the dead _run path

Harsh23Kashyap:
Closes #18447 and #18448.

## Summary

The agent tool redesign moved every tool to `ToolBase` / `ToolParamBase` with a `meta` block and an `_invoke` dispatcher. DeepL was ported in PR #18395. Jin10 and TuShare were the last two tools still on the legacy `ComponentBase` / `ComponentParamBase` pair, with only the dead `_run(self, history, **kwargs)` signature. They could not be used as Agent tools:

- `agent/component/agent_with_tools.py` calls `get_meta()` on every tool's parameter object: `original_name = cpn.get_meta()["function"]["name"]`. `get_meta()` is defined on `ToolBase` / `ToolParamBase`, not on `ComponentBase`. Adding Jin10 or TuShare to an Agent raised `AttributeError: 'Jin10Param' object has no attribute 'get_meta'`.
- The invoke path goes through `_invoke` (`ToolBase._invoke` is the documented entry; `grep -rn "\._run(" agent/` returns no callers). Pre-port tools only had `_run`, so any call hit `ComponentBase._invoke` → `NotImplementedError` (raised at `agent/component/base.py:447`).
- The legacy `be_output()` path (a `pd.DataFrame`) was the return shape of the old `_run`; the new `_invoke` contract uses `set_output("formalized_content", ...)` for the Agent-visible content and `set_output("_ERROR", ...)` for failures.

## Fix

Two changes in `agent/tools/`:

1. `jin10.py`: `Jin10Param` → `ToolParamBase` with a meta block (name `jin10_financial_data`, parameters: optional `query` and the four `type` options). `Jin10` → `ToolBase` with `@timeout _invoke`. The flash / calendar / symbols / news HTTP fetches and the per-type `DataFrame.to_markdown()` formatting are preserved verbatim. Output is published via `set_output("formalized_content", ...)` and `set_output("json", ...)`; failures via `set_output("_ERROR", ...)`. The upstream-input fallback for the news filter keyword is preserved (param's `contain` / `filter` config fields still win when set).

2. `tushare.py`: `TuShareParam` → `ToolParamBase` with a meta block (name `tushare_news`, parameters: optional `query`). `TuShare` → `ToolBase` with `@timeout _invoke`. The single `api_name=news` call is preserved. The keyword precedence (param's `keyword` wins, `kwargs["query"]` is the upstream fallback) is preserved. Output is published via `set_output("formalized_content", ...)` and `set_output("json", ...)`. A non-zero `response["code"]` is now surfaced as `set_output("_ERROR", response["msg"])` (pre-fix this case wrote the message into a data cell).

## Test

Eight regression tests in `test/unit_test/agent/tools/test_jin10_tushare_port.py`:

For Jin10:
- `test_jin10_param_is_tool_param_base_with_meta`: `Jin10Param` inherits `ToolParamBase` with a `ToolMeta` dict. Pre-fix: mro is `ComponentParamBase`.
- `test_jin10_get_input_form_returns_query_field`: `get_input_form` returns a non-empty dict. Pre-fix: missing.
- `test_jin10_tool_inherits_tool_base_with_invoke`: `Jin10` inherits `ToolBase` and defines `_invoke`. Pre-fix: only `_run`, hits `ComponentBase._invoke` → `NotImplementedError`.
- `test_jin10_invoke_surfaces_errors_via_set_output`: HTTP failure → `set_output("_ERROR", ...)` and a returned string. Pre-fix: `Jin10.be_output("**ERROR**: ...")` (legacy DataFrame).

For TuShare (same four tests, plus the non-zero response code case):
- `test_tushare_invoke_surfaces_errors_via_set_output` covers both the HTTP-failure path and the `response["code"] != 0` path.

Verified pre-fix behaviour: all 8 tests fail with informative errors naming the pre-port shape (ComponentParamBase mro, missing `get_input_form`, `issubclass(..., ToolBase) == False`, `NotImplementedError` on `_invoke`). With the fix applied, all 8 pass. `ruff check` + `ruff format --check` clean.

## Risks

- The port changes the tool's behaviour for the Agent user in two ways: (a) it now works (it was a hard crash pre-port), and (b) it publishes the result through `set_output("formalized_content", ...)` so the Agent surfaces the data, instead of returning a `DataFrame` through the dead `_run` path. A user who somehow wired Jin10 / TuShare into a non-Agent flow that consumed the legacy `be_output()` DataFrame will need to migrate; this is a non-existent path (the only caller is the Agent module, which doesn't dispatch to `_run`).
- The HTTP request shapes (URLs, params, headers, timeouts) are unchanged. TuShare's behaviour change for `response["code"] != 0` is the documented behaviour for every other ported tool.


wangq8 (commented):
This is an AI review comment.

The port is directionally correct and the regression tests are thorough. Two higher-level issues first:

1. **Direct overlap with #18465 (Jin10, #18447) and #18464 (TuShare, #18448).** This PR ports both Jin10 and TuShare, while #18464 (TuShare) and #18465 (Jin10 + tushare.py changes) are both still OPEN and touch the same files. These will conflict; please coordinate with the other authors so only one port lands per tool.

2. **`type` parameter is declared but ignored.** `meta["parameters"]["type"]` advertises an enum (`flash`/`calendar`/`symbols`/`news`) and the description says "Use the 'type' field to choose what to fetch", but `_invoke` never reads `kwargs.get("type")` — it only uses `self._param.type`. Either wire it up or drop it from the meta to avoid an inert, misleading parameter.

3. **No output clearing on failure.** `ToolBase.invoke` does not clear outputs between calls. A prior successful run leaves `formalized_content` (and `json`) in `_param.outputs`; a later run that fails sets `_ERROR` but never clears the stale `formalized_content`. #18465 explicitly addresses this by popping `formalized_content`/`_ERROR` at the start of `_invoke`. Please do the same here.

4. **`query` is inert too.** The meta documents `query` as "optional free-form text … used as the placeholder", but it's never consumed (filtering is done only via `contain`/`filter`). That's arguably fine if intentional, but an agent passing a `query` will silently get no filtering — worth wiring to a client-side filter (as #18465 does) or removing from the meta.

5. **Naming drift.** Tool names differ from #18465 (`jin10_financial_data` vs `jin10_market_data`; `tushare_news` vs `tushare_quick_news`). Once you agree on the winner, keep the merged tool name consistent so agent descriptors don't churn.

Minor: `_invoke` returns `self.output("formalized_content")` on success but a bare string on the error path; returning a consistent type would be cleaner.

PR #18641, opened 2026-08-21
fix(model-config): resolve bare tenant_model.id to composite when primary lookup fails

### LINKED ISSUE #18398: [Bug]: POST /api/v1/retrieval fails with LookupError when knowledgebase.embd_id is a bare tenant_model.id (UUID) while UI search works
### Self Checks

v0.26.4

### RAGFlow image version

v0.26.4

```Markdown
- DOC_ENGINE: Elasticsearch
- knowledgebase.embd_id stored as bare UUID (tenant_model.id), e.g. ab87a19896db11f189344b8174659a17
- POST .../embedding/check with the same embd_id succeeds (avg_cos_sim ~0.98)
- UI /datasets/search retrieval works; Python /retrieval fails
```

### Actual behavior

POST /api/v1/retrieval raises:

LookupError: Provider  not found for model ab87a19896db11f189344b8174659a17.

Stack (stock code path):
get_model_config_from_provider_instance(kb.tenant_id, EMBEDDING, kb.embd_id)
→ split_model_name has empty provider when embd_id has no "@"

Meanwhile:
- UI datasets/search can retrieve successfully
- embedding/check with the same embd_id works and shows high cos_sim vs stored vectors

Official frontend normally builds embd_id as name@instance@provider, but in our deployment knowledgebase.embd_id is a bare tenant_model.id. Python /retrieval does not resolve bare IDs, while other paths appear to tolerate them.

### Expected behavior

Either:
1) All retrieval paths should accept the same embd_id format (resolve bare tenant_model.id to name@instance@provider), or
2) Document/enforce that embd_id must always be name@instance@provider and reject/migrate bare IDs consistently across UI and APIs.

### Steps to reproduce

```Markdown
1. Have a dataset whose knowledgebase.embd_id is a bare tenant_model.id (32-char hex, no "@").
2. Confirm POST /api/v1/datasets/{id}/embedding/check with that embd_id returns code=0 and high avg_cos_sim.
3. Confirm UI /datasets/search returns chunks.
4. Call POST /api/v1/retrieval with the same dataset_ids/question (with or without metadata_condition).
5. Observe LookupError: Provider not found for model <uuid>.
```

### Additional information

Local workaround in chunk_api.py: before get_model_config_from_provider_instance, resolve bare embd_id via tenant_model/instance/provider to name@instance@provider (_resolve_embd_id). AI-assisted patch; attaching full chunk_api.py for reference.

Please advise the official supported embd_id contract and whether /retrieval should resolve tenant_model.id like other entry points.

Harsh23Kashyap:
Fixes #18398.

## Summary

`POST /api/v1/retrieval` raised `LookupError: Provider  not found for model ab87a19896db11f189344b8174659a17.` when `knowledgebase.embd_id` was stored as a bare `tenant_model.id` (32-char hex). The RAGFlow frontend normally builds `embd_id` in the composite form `name@instance@provider`; legacy / API-scripted deployments can store the bare id, which has no `@` so `split_model_name` returns an empty `provider_name`.

`resolve_model_config` (the single resolution point every model config call goes through) has two paths:
1. `get_model_config_by_id(tenant_id, model_type, model_ref)` — the primary id-based path. It handles bare ids for **healthy** rows but raises `LookupError` for inactive models, wrong model_type, or when the tenant has no access to the provider.
2. `get_model_config_from_provider_instance(tenant_id, model_type, model_ref)` — the provider-instance fallback. It calls `split_model_name(model_ref)` which returns an empty `provider_name` for bare ids and raises `LookupError: Provider  not found for model <uuid>.`.

The UI `/datasets/search` path happened to take path 1 successfully. The Python `/retrieval` path took path 1, hit a `LookupError` (model disabled / wrong type / no tenant access), and fell through to path 2 which also failed.

## Fix

One change in `api/db/joint_services/tenant_model_service.py`:

- A new helper `_composite_from_bare_model_id(tenant_id, model_type, model_id)` reads the `tenant_model` row directly, validates type / status / tenant access, and returns the composite `name@instance@provider` form, or `None` if any check fails.
- `resolve_model_config` calls the helper when `get_model_config_by_id` raises and `model_ref` has no `@`, and tries the provider-instance lookup with the composite form before falling through to the original failing path.

The bare-id resolution is defensive: `get_model_config_by_id` remains the canonical path; the new helper is only consulted when the primary path raised and the model_ref is bare. A genuinely missing model still raises the same `LookupError: Provider  not found for model <id>.` it raised pre-fix (the test pins this — `test_resolve_model_config_bare_id_row_unavailable_preserves_original_error`).

The fix is local to `resolve_model_config`, which is the single resolution point used by every caller — `api/utils/api_utils.py`, `api/db/services/dialog_service.py`, `api/db/joint_services/memory_message_service.py`, `api/apps/restful_apis/chunk_api.py`, and more. A call-site patch (e.g. the user's "AI-assisted patch" in `chunk_api.py`) would have fixed only one caller; fixing it in the shared function fixes all of them.

## Test

Seven regression tests in `test/unit_test/api/db/services/test_resolve_model_config_bare_id.py`:

- `test_resolve_model_config_composite_passes_through`: a composite `'name@instance@provider'` model_ref still works.
- `test_resolve_model_config_bare_id_primary_succeeds`: when the id-based path succeeds for a bare id, the bare-id composite fallback is **not** touched.
- `test_resolve_model_config_bare_id_falls_back_to_composite`: the central test. When `get_model_config_by_id` raises and the model_ref has no `@`, the provider-instance lookup is called with the composite `'name@instance@provider'` form built from the row, not the bare id. Pre-fix, the captured `model_name` was the bare id `'abc123uuid'`; post-fix it's `'gpt-4@default@OpenAI'`.
- `test_resolve_model_config_bare_id_row_unavailable_preserves_original_error`: a genuinely missing model still raises the original `LookupError` (the fix does not mask the error).
- `test_composite_from_bare_model_id_returns_name_instance_provider`: the helper returns the composite form in the happy case.
- `test_composite_from_bare_model_id_returns_none_when_row_missing`: returns `None` when the row does not exist, so the caller falls through.
- `test_composite_from_bare_model_id_returns_none_on_type_mismatch`: a model of a different type is not resolved (CHAT looked up as EMBEDDING returns `None`).

Verified pre-fix behaviour: 4 of 7 tests fail with informative assertions — the central test fails with the captured `'gpt-4@default@OpenAI'` vs `'abc123uuid'` comparison showing the pre-fix bare-id pass-through; the 3 helper tests fail with `AttributeError: module '...tenant_model_service' has no attribute '_composite_from_bare_model_id'`. The 3 pre-fix-passing tests are regression guards. With the fix applied, all 7 pass. `ruff check` + `ruff format --check` clean.

## Risks

- The bare-id composite fallback is a new code path, so a future bug there could regress previously-working flows. The test suite covers the happy path, the genuine-missing-model path, the type-mismatch path, and the row-missing path. The bare-id fallback is only triggered when the primary id-based path raised **and** the model_ref has no `@`, so the well-trodden composite path is unchanged.
- A user who stored a bare id for a model of a different type (e.g. `tenant_llm_id` is a bare id, but a flow asks for it as `EMBEDDING`) now gets `None` from the helper and the same original `LookupError`. Pre-fix, they also got a `LookupError` (from `get_model_config_from_provider_instance`). The behaviour is identical.
- The user's "AI-assisted patch" attached a call-site patch in `chunk_api.py`; the fix here is at the shared resolution point, so the same fix covers every caller.


buua436:
The canonical model reference is now tenant_model.id; the composite model@instance@provider form is retained only for backward compatibility. This has already been fixed on main by resolving model references by ID first and falling back to the legacy composite format in #16744.

JinHai-CN:

Issue #16034, opened 2026-06-15
[Bug]: Metadata still capped at 10k — retrieval metadata_condition empty, filter stats wrong (follow-up #13210)


Jsteven1106:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

v0.25.6 (still reproducible on `main` `doc_metadata_service.py` as of 2026-06-15)

### RAGFlow image version

v0.25.6

### Other environment information

```Markdown
- DOC_ENGINE: Elasticsearch
- Dataset ID: 4929bc0816fb11f18212033671e3d0b7
- Total documents: 12729 (all have metadata in ragflow_doc_meta_*)
```

### Actual behavior

**Follow-up to #13210.** Datasets with >10k metadata rows still truncate silently (HTTP 200).

1. **`POST /api/v1/retrieval` + `metadata_condition`** → empty result, while the same request **without** `metadata_condition` works. Log:
   ```
   [get_flatted_meta_by_kbs] Results hit the 10000 limit for KBs ['4929bc0816fb11f18212033671e3d0b7'].
   ```

2. **`GET /api/v1/datasets/{id}/documents?type=filter`** → metadata counts sum to exactly 10000; remaining docs misreported as `empty_metadata`:
   - `total`: 12729
   - `agent_code` counts: 3+8+137+9852 = **10000**
   - `empty_metadata.true`: **2729** (= 12729 − 10000), but those docs have metadata in ES

3. Same 10k cap affects batch metadata CRUD selector and `GET /datasets/metadata/flattened`.

### Expected behavior

**Follow-up to #13210.** Datasets with >10k metadata rows still truncate silently (HTTP 200).

1. **`POST /api/v1/retrieval` + `metadata_condition`** → empty result, while the same request **without** `metadata_condition` works. Log:
   ```
   [get_flatted_meta_by_kbs] Results hit the 10000 limit for KBs ['4929bc0816fb11f18212033671e3d0b7'].
   ```

[...]

PR #18642, opened 2026-08-21
fix(retrieval): use metadata push-down (ES/Infinity) instead of in-memory only

### LINKED ISSUE #18397: [Bug]: POST /api/v1/retrieval with metadata_condition returns empty while /datasets/search with meta_data_filter works
### Self Checks

v0.26.4

### RAGFlow image version

v0.26.4

```Markdown
- DOC_ENGINE: Elasticsearch
- Dataset size: >10k documents with metadata (e.g. agent_code)
- Related: #16034, #13388
```

### Actual behavior

UI / retrieval test via POST /api/v1/datasets/search with meta_data_filter (manual) returns matching chunks (vector_similarity can be ~0.99).

The same dataset + question via POST /api/v1/retrieval with metadata_condition returns empty chunks (total=0), or previously returned empty after filtering.

Root cause appears to be inconsistent metadata filter paths:
- /datasets/search uses apply_meta_data_filter (ES/Infinity push-down)
- /retrieval used get_flatted_meta_by_kbs + in-memory meta_filter only

For large datasets, in-memory filtering can miss documents (see #16034 style truncation / pagination issues), while push-down succeeds.

### Expected behavior

POST /api/v1/retrieval with metadata_condition should return the same filtered document set as the UI /datasets/search meta_data_filter for the same conditions, and should use the same push-down (or equivalently correct) filter path.

### Steps to reproduce

```Markdown
1. Use a dataset with >10k docs and metadata field `agent_code`.
2. Call UI/API: POST /api/v1/datasets/search with:
{
  "dataset_ids": ["<dataset_id>"],
  "question": "<query>",
  "meta_data_filter": {
    "method": "manual",
    "manual": [{"key": "agent_code", "op": "=", "value": "AGENT_WST_SHENGCHAN"}]
  }
}
3. Observe non-empty chunks.
4. Call POST /api/v1/retrieval with:
{
  "question": "<query>",
  "dataset_ids": ["<dataset_id>"],
  "metadata_condition": {
    "logic": "and",
    "conditions": [
      {"name": "agent_code", "comparison_operator": "=", "value": "AGENT_WST_SHENGCHAN"}
    ]
  }
}
5. Observe empty result on stock /retrieval metadata path.
```

### Additional information

We patched api/apps/restful_apis/chunk_api.py so /retrieval prefers DocMetadataService.filter_doc_ids_by_meta_pushdown (same idea as apply_meta_data_filter), falling back to meta_filter. After that, metadata filtering works.

Please consider aligning /retrieval with /datasets/search officially.

I will attach our modified chunk_api.py (AI-assisted local fix) for reference. This is a workaround, not a formal PR yet.

Harsh23Kashyap:
Fixes #18397.

## Summary

`POST /api/v1/retrieval` with `metadata_condition` returned empty chunks on large datasets (>10k documents) where `UI /datasets/search` with the same `meta_data_filter` returned non-empty chunks. Root cause: the two paths used different filter implementations.

- `/retrieval` called `filter_doc_ids_by_metadata` (in-memory fetch via `DocMetadataService.get_flatted_meta_by_kbs`) directly. For large datasets the in-memory round-trip is paginated / truncated, so documents can be missed (the user's diagnosis, see #16034-style issues).
- `/datasets/search` called `apply_meta_data_filter` which tries the ES / Infinity / GaussDB **push-down** first via `_try_meta_pushdown` -> `DocMetadataService.filter_doc_ids_by_meta_pushdown` and falls back to the in-memory path only when push-down cannot service the request. The push-down is a direct query against the metadata index, so the answer is complete and correct.

## Fix

One change in `api/apps/restful_apis/chunk_api.py`: the `/retrieval` endpoint's `metadata_condition` branch now calls `apply_meta_data_filter` (the same helper `/datasets/search` uses) instead of `filter_doc_ids_by_metadata` directly.

```python
manual_filter = {
    "method": "manual",
    "manual": convert_conditions(metadata_condition),
    "logic": metadata_condition.get("logic", "and"),
}
doc_ids = await apply_meta_data_filter(
    manual_filter,
    None,  # metas (eager; the loader is the fallback)
    "",    # question (unused for manual)
    None,  # chat_mdl (unused for manual)
    None,  # base_doc_ids (the helper will compute the manual slice)
    kb_ids=kb_ids,
    metas_loader=lambda: DocMetadataService.get_flatted_meta_by_kbs(kb_ids),
)
```

The existing `convert_conditions` helper maps the API's `metadata_condition` shape (`{name, comparison_operator, value}`) to the manual filter shape (`{key, op, value}`) that `apply_meta_data_filter` consumes. The "if not doc_ids, return empty result" and "if no conditions, return `-999`" guards are preserved. The `filter_doc_ids_by_metadata` import is dropped (no other caller in `chunk_api.py`).

## Test

Five regression tests in `test/unit_test/api/test_retrieval_meta_pushdown.py`:

- `test_api_metadata_condition_converts_to_manual_filter_shape`: the API's `metadata_condition` shape maps via `convert_conditions` to the manual filter shape `apply_meta_data_filter` expects.
- `test_apply_meta_data_filter_pushdown_path_is_used_before_in_memory_fallback`: when `_try_meta_pushdown` returns doc_ids, the in-memory path is **not** consulted. Pins the order (push-down first).
- `test_apply_meta_data_filter_falls_back_to_in_memory_when_pushdown_returns_none`: when `_try_meta_pushdown` returns `None` (push-down cannot service), the in-memory path runs and its result is returned. Pins the fallback for backends without push-down support.
- `test_apply_meta_data_filter_short_circuits_empty_pushdown_results`: when push-down returns an empty list, the helper propagates the `['-999']` sentinel and the in-memory path is **not** consulted — an honest empty result is better than a stale in-memory answer.
- `test_retrieval_uses_apply_meta_data_filter_not_filter_doc_ids_by_metadata`: AST-level guard on the `chunk_api.py` call site. `/retrieval` must import `apply_meta_data_filter` (not `filter_doc_ids_by_metadata`) and call it inside `retrieval_test`'s `metadata_condition` branch. Pre-fix, the test fails on the import assertion with the informative message that the direct in-memory call is the bug. The AST approach avoids the heavy import chain that triggers the DB services to be initialised.

Verified pre-fix behaviour: the central AST guard fails with `chunk_api.py must import apply_meta_data_filter from common.metadata_utils so the /retrieval endpoint uses the push-down path`. The 4 helper tests pass on pre-fix because they exercise the helper itself, not the call site — the bug is in the call site choosing the wrong helper. With the fix applied, all 5 pass. `ruff check` + `ruff format --check` clean.

## Risks

- The new code is `async` (the helper is `async def apply_meta_data_filter`); the `/retrieval` endpoint is already `async`, so this is a one-token `await` addition.
- A user on a backend without push-down support (e.g. oceanbase / seekdb in some configurations) gets the same in-memory fallback they had pre-fix; the helper returns the in-memory result on push-down `None`.
- The push-down path is preferred but may differ from the in-memory result on edge cases (e.g. push-down might support a filter the in-memory path doesn't, or vice versa). The fix is the right call because push-down is the canonical path used by the UI; the in-memory fallback is the existing compatibility layer.
- Backward compatible: a user with no `metadata_condition` in the request never enters the changed branch. A user with `metadata_condition` that the push-down path can't service still gets the in-memory answer (the `None`-signal case).

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -57,7 +57,7 @@ from api.utils.reference_metadata_utils import (
 from common import settings
 from common.constants import LLMType, ParserType, RetCode, TaskStatus
 from common.doc_store.doc_store_base import OrderByExpr
-from common.metadata_utils import convert_conditions, filter_doc_ids_by_metadata
+from common.metadata_utils import apply_meta_data_filter, convert_conditions
 from common.misc_utils import thread_pool_exec
 from common.string_utils import is_content_empty, remove_redundant_spaces
 from common.tag_feature_utils import validate_tag_features
@@ -364,12 +364,28 @@ async def retrieval_test(tenant_id):
     if not doc_ids:
         metadata_condition = req.get("metadata_condition")
         if metadata_condition:
-            doc_ids = filter_doc_ids_by_metadata(
-                kb_ids,
-                convert_conditions(metadata_condition),
-                metadata_condition.get("logic", "and"),
-                lambda: DocMetadataService.get_flatted_meta_by_kbs(kb_ids),
+            # Try ES / Infinity / GaussDB push-down first (same as the
+            # UI /datasets/search path) so large datasets don't pay the
+            # in-memory metadata round-trip. The in-memory
+            # filter_doc_ids_by_metadata remains the fallback for
+            # backends without push-down support (or filters the push-down
+            # path can't express). See issue #18397.
+            manual_filter = {
+                "method": "manual",
+                "manual": convert_conditions(metadata_condition),
+                "logic": metadata_condition.get("logic", "and"),
+            }
+            doc_ids = await apply_meta_data_filter(
+                manual_filter,
+                None,
+                "",
+                None,
+                None,
+                kb_ids=kb_ids,
+                metas_loader=lambda: DocMetadataService.get_flatted_meta_by_kbs(kb_ids),
             )
+            if doc_ids is None:
+                doc_ids = []
             if not doc_ids and metadata_condition.get("conditions"):
                 return get_result(data={"total": 0, "chunks": [], "doc_aggs": {}})
             if metadata_condition and not doc_ids:
--- a/test/unit_test/api/test_retrieval_meta_pushdown.py
+++ b/test/unit_test/api/test_retrieval_meta_pushdown.py
@@ -1,0 +1,303 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for issue #18397.
+
+``POST /api/v1/retrieval`` with ``metadata_condition`` returned empty
+chunks on large datasets (>10k documents) where ``/datasets/search`` with
+the same ``meta_data_filter`` returned non-empty chunks. Root cause:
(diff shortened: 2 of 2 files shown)


qinling0210 (commented):
@Harsh23Kashyap The issue has been fixed in https://github.com/infiniflow/ragflow/pull/18219. You can test again in latest main code.

qinling0210:
Closing the PR as the issue has been fixed

PR #18644, opened 2026-08-21
fix: precompile regexes on the retrieval hot path

changshenhan:
## Summary

`regexp.MustCompile` / `regexp.MatchString` have **no cache** in Go — every call performs a full parse+compile of the pattern. The fulltext query builder compiled the same 17 patterns from scratch on **every** invocation of the retrieval hot path (tokenize → query build), paying a fresh regexp parse+compile per token/term.

This hoists all patterns into package-level `var`s compiled **once at package init**, and references them from every call site. **Zero logic change** — every pattern string is byte-identical to before.

## Hot-path call chain affected

- `IsChinese` / `SubSpecialChar` / `RmWWW` (repeatedly called per token in `TextSplitter` tokenization)
- `AddSpaceBetweenEngZh` (×4) / `NeedFineGrainedTokenize`
- `Question` cleaning (`rePunct`)
- token cleaning (`reCleanQuote` / `reSingleChar` / `reLeadSign`)
- `query` special-char / `specialCharRe` / `cleanTk`

These run per-token/per-query in the retrieval pipeline (see callers in `nlp/retrieval.go`, `nlp/reranker.go`, `memory.go`, `tag.go`).

## Impact

- Removes ~17 redundant regexp parse+compile per token/query from the hot path.
- Go's regexp compile is not cached and is measurably more expensive than the `match` itself for short patterns.

## Testing

Patterns are byte-identical to the originals (verified against `HEAD`). CI `go test` covers compilation.

### DIFF
--- a/internal/service/nlp/query_builder.go
+++ b/internal/service/nlp/query_builder.go
@@ -45,6 +45,30 @@ type QueryBuilder struct {
 	synonym     *Synonym
 }
 
+// Precompiled regexes. Compiling these per-call on the retrieval hot path
+// (Go's regexp.MustCompile / regexp.MatchString has no cache) wastes a full
+// parse+compile on every token/term. All patterns below are compile-time
+// constants, so they are compiled once at package init.
+var (
+	reEngAlpha       = regexp.MustCompile(`^[a-zA-Z]+$`)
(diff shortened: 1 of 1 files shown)


qinling0210 (approved):
LGTM

Issue #18569, opened 2026-08-20
TSR over-segments cross-page table continuation: 13 'table row' components for ~5-7 physical rows (asset-recovery-sd-zh-tw)


xugangqiang:
## Summary

TSR (table-structure recognition) over-segments a **cross-page table continuation**: on page 11 it emits **13 "table row" components** for a table that physically has **~5–7 rows**, because the long wrapped text of column 3 that continues from page 10 is detected as separate row components.

## Repro

- File: `asset-recovery-services-sd-zh-tw.pdf` (Dell marketing PDF, Traditional Chinese, ~13 pages)
- Table: the "服務條款與條件" (Terms & Conditions) table spanning page 10 → page 11

TSR raw output (`tsr_raw` dump, page 11, table index 2):

```
table row components: 13
  y0:    7980.1, 7989.6, 8009.1, 8026.6, 8045.8, 8064.9, 8082.4, 8134.3, 8151.8, 8169.0, 8186.5, 8203.3, 8220.0
  height: 7.1,   19.5,   17.2,   19.3,   19.2,   17.4,   51.6,   17.6,   17.5,   17.5,   17.0,   17.6,   28.7
```

Note the abnormal components: one row is only **7.1 pt tall** (a split line) and another is **51.6 pt tall** (a wrapped text block).

## Evidence of the cross-page continuation

The same text appears on both pages in the OCR dump:

| Text (col 3) | page 10 y | page 11 y |
|---|---|---|
| 產生疑問，在本質上僅直接與購買者和銷售者相關… | 7615.7 | 7996.7 |
| 何付款條款或其他合約條款並不適用於您，有關的… | 7633.0 | 8014.0 |
| 條款或其他合約條款，將由您與銷售者達成協定。 | 7650.3 | 8031.3 |

So the table's third column contains a long paragraph that starts on page 10 and continues on page 11. On page 11, TSR turns **each wrapped text line** of that continuation into its own `table row` component (the first 4 components at y≈7980–8045 match the continuation lines above).

## Root cause

1. The table is a **cross-page table**; its page 11 crop includes the continuation of column 3's wrapped paragraph.
2. TSR emits a `table row` component per text line for that continuation (plus one narrow 7 pt component), instead of grouping them into the real physical rows (歐洲中東及非洲 / 此外位於法國… / 法國 / 德國 / 英國 ≈ 5 rows).
[...]

PR #18645, opened 2026-08-21
fix(deepdoc): collapse phantom TSR row components emitted on cross-page table continuations

Harsh23Kashyap:
## Summary

`TableStructureRecognizer.__call__` over-emits `table row` components on cross-page table continuations with a wrapped paragraph in one column. The model emits one component per wrapped text line (13 in the repro) instead of one per physical row (5-7 in the repro), so `construct_table` produces 2x or more phantom grid rows in the final HTML/Markdown table.

Refs #18569 (`asset-recovery-services-sd-zh-tw.pdf`, page 11, table index 2).

## Root cause

The TSR model output is post-processed for row/column coordinate alignment in `deepdoc/vision/table_structure_recognizer.py:__call__`, but the per-image row list is otherwise passed through verbatim. On cross-page continuations the wrapped paragraph lines all sit within a tight y-band (gaps near 0), so the model treats each as a separate `table row`. The downstream `pdf_parser.py:gather` then uses `sort_Y_firstly` and `construct_table` uses `R` labels to group boxes, so the phantom rows survive and inflate the row count.

## Fix

Add a post-processing step in `TableStructureRecognizer.__call__` (after the column snapping) that:

- drops any `table row` / `table column header` / `table projected row header` component whose height is below `_PHANTOM_ROW_MIN_HEIGHT` (10px). This catches split-line / degenerate detections like the 7.1pt row in the repro.
- sorts the surviving row components by `top`, then merges any two consecutive components whose vertical gap is below `_PHANTOM_ROW_MERGE_GAP` (10px) into a single component with the larger `bottom` and the max `score`.

Non-row components (`table column`, `table`, `table spanning cell`) are passed through unchanged. The step runs per-image (one table image per call), so it lands before `pdf_parser.py:gather/construct_table` consume the row list.

The repro's 13 components collapse to 1: the wrapped paragraph becomes a single logical row, matching the physical row count when combined with the other rows on the page.

## Test

Seven regression tests in `test/unit_test/deepdoc/vision/test_tsr_row_merge.py`:

- `test_merge_collapses_bug_18569_thirteen_rows_to_one`: the exact 13-row reproduction from the issue collapses to one row.
- `test_merge_preserves_well_spaced_real_rows`: 40pt gaps between real rows are not collapsed.
- `test_merge_drops_split_line_below_min_height`: 5pt row is dropped even when alone.
- `test_merge_passes_through_non_row_components`: `table column`, `table`, and `table spanning cell` are preserved unchanged. Only the three row-like labels participate in the merge.
- `test_merge_combines_close_header_and_body_rows`: a header row 1pt above a body row merges into one.
- `test_merge_keeps_max_score_of_merged_components`: the merged component's score is the max of the two.
- `test_merge_no_op_when_only_columns_present`: a table with only `table column` components is unchanged.

Tests load the real `_merge_phantom_row_components` via `importlib.util.spec_from_file_location` with the module's heavy runtime dependencies (`cv2`, `huggingface_hub`, `numpy`, etc.) stubbed, mirroring `test_recognizer_column_fit.py`.

7/7 new tests pass. 17/17 existing vision tests pass. 5/5 `test_pdf_parser_table_coordinates` tests pass. `ruff check` + `ruff format --check` clean.

## Why not a follow-up to a future model retrain

The TSR ONNX weights are versioned and shared with the Go port. Changing the model output would require coordinated model + Go parser updates. The post-processing step is Python-only, takes effect immediately, and is small enough to revert if the model is retrained later.


xugangqiang:
@Harsh23Kashyap  Do you have a real case (end to end test) for this issue? Like a PDF parsign will have issue before fix and the issue will be fixed with this fix. If so, please share such pdf file and the related end to end test. Thank you.

Issue #17298, opened 2026-07-23
[Bug]: The mineru program fails to correctly recognize tables when present in PDF files. This might be a bug in the code.


cuihangrui:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

0.26.4

### RAGFlow image version

0.26.4

### Other environment information

```Markdown

```

### Actual behavior

I used curl to call mineru and the parsing result is as follows

<img width="1371" height="336" alt="Image" src="https://github.com/user-attachments/assets/b4d88c3e-4cf6-438c-b55b-f33e47d835ec" />
However, the information in the table of the Ragflow knowledge base parsing result is garbled.

<img width="2328" height="571" alt="Image" src="https://github.com/user-attachments/assets/4ffee5ee-5815-4bff-8e7a-b7f56b9bc8dc" />
Is it due to the following issue? In [mineru_parser.py] (/D:/ragflow-0.26.4/ragflow-0.26.4/deepdoc/parser/mineru_parser.py), RAGFlow calls MinerU/file_parse and then downloads the ZIP file, decompresses and extracts the contents. The problem is most likely in:
The encoding of the markdown files in the ZIP is not UTF-8, causing garbled characters when reading
The HTML tables in the markdown are wrongly stripped or escaped
The subsequent chunk processing disrupts the content of the table

### Steps to reproduce

```Markdown
Is it due to the following issue? In [mineru_parser.py] (/D:/ragflow-0.26.4/ragflow-0.26.4/deepdoc/parser/mineru_parser.py), RAGFlow calls MinerU/file_parse and then downloads the ZIP file, decompresses and extracts the contents. The problem is most likely in:
The encoding of the markdown files in the ZIP is not UTF-8, causing garbled characters when reading
The HTML tables in the markdown are wrongly stripped or escaped
The subsequent chunk processing disrupts the content of the table
```

cuihangrui:
{
    "code": 0,
    "data": {
        "chunks": [
            {
                "available": true,
[...]

PR #18647, opened 2026-08-21
fix(deepdoc): sanitize MinerU table HTML before emitting to the table chunker

Harsh23Kashyap:
## Summary

`MinerUParser._transfer_to_tables` emitted the model's raw `table_body` HTML fragment (`<table><tr><td>...</td></tr></table>`) verbatim into the table tuple. `tokenize_table` stored that string in the chunk's `content_with_weight` field, so the user saw the literal HTML tag markup as "garbled" table content in the knowledge base.

Refs #17298.

## Root cause

Two code paths consume MinerU's `table_body`:

1. `_transfer_to_sections` (line 872) already sanitizes the HTML when `table_enable=False` — that path is fine.
2. `_transfer_to_tables` (line 905) had no sanitization. The raw HTML went straight to the table tuple.

The two paths were inconsistent, and the table tuple path was the one most users hit (the default `table_enable=True` keeps tables as separate media chunks).

## Fix

`deepdoc/parser/mineru_parser.py`: call `self._sanitize_section_text(text)` on the assembled `body + caption + footnote` in the table branch of `_transfer_to_tables`, mirroring the sections path.

The sanitizer:

- unescapes HTML entities (`html.unescape`)
- converts `<br>`, `</p>`, `</div>`, `</li>`, `</tr>`, `</h*>`, `</table>`, `</caption>` to newlines
- strips remaining tags
- collapses whitespace

The user gets readable plain text instead of raw HTML. Plain-text table content (no markup) is unchanged — the sanitizer only strips tags and entities.

## Test

Three test changes in `test/unit_test/deepdoc/parser/test_mineru_parser.py`:

- `test_transfer_to_tables_emits_ordered_typed_media` — the expected text now matches the sanitized plain-text output (`first`) instead of the raw HTML fragment. Pins the fix's behavior change.
- `test_transfer_to_tables_sanitizes_html_table_body` — new regression test for #17298. Verifies that a complex multi-row table with `rowspan`/`colspan` attributes is sanitized to row-separated cell text, and that the caption is appended after the sanitized body.
- `test_transfer_to_tables_does_not_strip_plain_text_tables` — guards the contract that plain-text table content (no HTML markup) passes through unchanged.

Verified pre-fix behavior: the new sanitization test fails on the pre-fix code because the raw HTML would be emitted unchanged. The updated `emits_ordered_typed_media` test also fails on the pre-fix code because it expected the raw HTML and the new contract expects sanitized text.

30/30 tests in `test_mineru_parser.py` pass. ruff check + ruff format --check clean.

## Why a smaller change

The reported chunk content in #17298 also shows garbled cell text (random strings like `WEY`, `AAJEERI`, `EPZ$`) which is a MinerU model output issue, not a RAGFlow rendering issue. That part needs a model-side fix. This PR addresses the part RAGFlow controls: the chunk rendering, which was emitting raw HTML tags in the chunk's `content_with_weight` field.

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -906,6 +906,14 @@ class MinerUParser(RAGFlowPdfParser):
                 if not text.strip():
                     self.logger.warning("[MinerU] Empty table content at page_idx=%s; using fallback text.", output.get("page_idx"))
                     text = "FAILED TO PARSE TABLE"
+                else:
+                    # MinerU returns table_body as an HTML fragment (`<table><tr><td>...</td></tr>...`).
+                    # Without sanitization the literal tag markup ends up in the
+                    # chunk's `content_with_weight` field, which `tokenize_table`
+                    # stores verbatim and the user then sees as "garbled" table
+                    # content. The sections path (line 872) already sanitizes
+                    # when `table_enable=False`; the tables path was missed.
+                    text = self._sanitize_section_text(text)
                 tables.append(((None, text), positions))
                 continue
 
--- a/test/unit_test/deepdoc/parser/test_mineru_parser.py
+++ b/test/unit_test/deepdoc/parser/test_mineru_parser.py
@@ -161,6 +161,72 @@ def test_transfer_to_sections_only_sanitizes_tables(monkeypatch, output, expecte
     assert sections[0][0] == expected
 
 
+def test_transfer_to_tables_sanitizes_html_table_body(monkeypatch):
+    """Regression for #17298.
+
+    The tables path used to emit MinerU's raw ``table_body`` HTML fragment
+    (``<table><tr><td>...</td></tr></table>``) verbatim into the chunk's
+    ``content_with_weight`` field. The user then saw the literal HTML tag
+    markup as "garbled" table content in the knowledge base. The sections
+    path already sanitizes when ``table_enable=False``; the tables path
+    must do the same so the chunked table content is human-readable.
+    """
+    module = _load_mineru_parser(monkeypatch)
+    parser = module.MinerUParser()
+    outputs = [
+        {
+            "type": module.MinerUContentType.TABLE,
+            "table_body": "<table><tr><td rowspan=1 colspan=1></td><td rowspan=1 colspan=1>WEY</td><td rowspan=1 colspan=1>B1</td></tr><tr><td rowspan=1 colspan=1>tA</td><td rowspan=1 colspan=1>AAJEERI</td><td rowspan=1 colspan=1>EPZ$</td></tr></table>",
+            "table_caption": [],
+            "table_footnote": [],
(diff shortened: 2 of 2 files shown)


xugangqiang:
I reviewed this PR against the rest of the codebase's table handling and found a design inconsistency worth resolving before merge.

**Core issue: the downstream contract for table chunks is already HTML, not plain text.**

- `deepdoc/vision/table_structure_recognizer.py:construct_table` (default `html=True`) returns a `<table><tr><td>` HTML string, and `deepdoc/parser/pdf_parser.py:1482` uses it as the table row payload.
- The excel, docling/opendataloader (opendataloader_parser.py:177-179), and tcadp parsers also emit HTML tables.
- `rag/nlp/__init__.py:507` (tokenize_table) stores that HTML verbatim into `content_with_weight` for the `isinstance(rows, str)` branch — it is never flattened there.
- The frontend (markdown-content/index.tsx:29, next-markdown-content/index.tsx:30, chunk-card/index.tsx:131) renders `content_with_weight` through `rehypeRaw`, so a `<table>` is drawn as a real table.

So PDF/Excel/docling tables render as real tables, but this PR flattens MinerU's table to plain text (`<td>A</td><td>B</td>` → `AB`), which is both inconsistent with every other parser and a display-quality regression for MinerU.

**Concrete problems with the current change:**

1. **Cell structure is lost.** `_sanitize_section_text` strips `<td>/<th>` with no delimiter, so adjacent cells merge (`WEY`+`B1` → `WEYB1`). The PR's own `test_transfer_to_tables_sanitizes_html_table_body` pins this degenerate concatenation (`assert ... == (None, "WEYB1\ntAAAJEERIEPZ$")`) as expected behavior — that bakes the defect into the contract.

2. **Asymmetry with the rest of the system.** Other parsers keep HTML table structure; only MinerU would flatten. This makes MinerU tables worse than others and harder to keep consistent.

3. **May not address the actual bug.** The #17298 "garbled" report likely comes from a view that emits `content_with_weight` raw (e.g. dataset/testing/testing-result.tsx:88 renders `{x.content_with_weight}` without HTML rendering), where PDF tables would also show tags. If so, the fix belongs in the rendering view (render HTML consistently), not in stripping structure from one parser.

**Suggestion:** instead of flattening, MinerU should emit table HTML consistently with the rest of the system — pass `table_body` through as-is, or normalize to a standard `<table>` if the real output carries extra wrappers — preserving td/tr/th. The existing tests asserting flattened text (`first`, `WEYB1`) would then need to be rewritten against the HTML contract.

Please confirm the exact view behind #17298 and the real shape of MinerU `table_body` (clean `<table>` vs. wrapped markup) to decide between pass-through and normalization.

xugangqiang:
@buua436  @yuzhichang  Please have a second eye

JinHai-CN:
please re-open the PR, if you still working on it.

PR #18648, opened 2026-08-21
fix(chunker): one BPE pass per hardSplitPiece iteration instead of two

changshenhan:
## Summary

`hardSplitPiece` splits an over-target boundary-less run (no sentence delimiters) by repeatedly cutting a token-bounded prefix. Each cut re-encoded the shrinking remainder **twice**: once in the loop guard (`tokenizeStr(rest) > target`) and again inside (`TrimContentToTokenLimit(rest, target)`).

Because **cl100k BPE is not prefix-stable** (`encode(text)[k:]` ≠ `encode(text[len(decode(tokens[:k])):])` — verified empirically: standalone test found a token mismatch at k=873, ~8% of boundaries land mid-multibyte-char), the remainder cannot reuse a token slice across cuts and must be re-tokenized from scratch each round. But it does not need to be tokenized **twice in the same round**.

`TrimContentToTokenLimit` returns its input **unchanged exactly when it fits within the token target** (`if len(tokens) <= limit { return s }`), so its result doubles as the loop-exit signal. The loop now runs **one BPE pass per iteration instead of two** — still O(L²/target), but with the per-iteration constant halved.

## Correctness

Output is **byte-identical** to the previous algorithm **under a loaded cl100k encoder** (the encoder-missing fallback path below intentionally differs). Note: the differential test itself requires the encoder to be loaded, so it is only meaningful in a built-with-deps environment. Locked by a differential test (`TestHardSplitPieceMatchesReference`) that runs the preserved reference implementation (the exact pre-optimization code) against the new one across mixed text / CJK / PDF-tag / emoji / edge-case inputs at targets 32/64/128/512 and asserts identical output and lossless concatenation.

Full chunker suite: only `TestChunkerGoldenParity` fails, and its 14 failing subtests are **identical to the baseline** (pre-existing CGO_ENABLED=0 environment drift, unchanged by this PR).

> Behavioral note (encoder-missing fallback only): when `cl100k_base.tiktoken` is absent and the encoder fails to load, the old code emitted the whole over-limit remainder as one piece (the guard `tokenizeStr(rest) > target` reads 0, so the loop never ran); the new loop falls back to `TrimContentToTokenLimit`'s byte-length trim and emits bounded pieces. Strictly more correct, and only reachable in a broken deps setup that the bundled path never hits. Now pinned by `TestHardSplitPiece_EncoderMissingFallback` (follow-up 2 from review).

## Performance (target=512, boundary-less runs)

| input | before | after | speedup | allocs |
|---|---|---|---|---|
| ~12 KB ASCII+CJK | 11.55 ms/op | 6.43 ms/op | **1.80×** | 32,961 → 18,375 (−44%) |
| ~12 KB CJK | 424.3 ms/op | 215.6 ms/op | **1.97×** | 547 → 403 (−26%) |

Piece counts identical (8 / 10) before and after.

## Scope note

`takeFromStart`/`takeFromEnd` also have O(n²) split loops but are default-off in `splitJsonRecursive`/delimiter paths and were deliberately left untouched to keep this change behavior-neutral.

### DIFF
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -1176,9 +1176,21 @@ func splitSentencesLossless(text string) []string {
 func hardSplitPiece(text string, docType string, target int) []schema.ChunkDoc {
 	var out []schema.ChunkDoc
 	rest := text
-	for tokenizeStr(rest) > target {
+	for {
+		// One BPE pass per iteration instead of two. TrimContentToTokenLimit
+		// encodes rest once and returns rest unchanged exactly when rest fits
+		// within target tokens, so its result doubles as the loop-exit
+		// condition - the loop no longer runs a separate full tokenizeStr(rest)
+		// re-encode just to decide whether to keep going. The shrinking
+		// remainder still has to be re-tokenized from scratch each round:
+		// cl100k BPE is not prefix-stable (encode(text)[k:] is not
+		// encode(text[byteBoundaryOfTokens[:k]:])), so a token slice cannot be
+		// reused across cuts - but that is one encode per iteration, not two.
 		head := tokenizer.TrimContentToTokenLimit(rest, target)
-		if head == "" || head == rest {
+		if head == rest {
+			break // rest already fits; the trailing append emits it
+		}
+		if head == "" {
 			break // cannot shrink further; avoid an infinite loop
 		}
 		// TrimContentToTokenLimit decodes a token prefix; if that lands on a
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -1325,9 +1325,21 @@ func splitSentencesLossless(text string) []string {
 func hardSplitPiece(text string, docType string, target int) []schema.ChunkDoc {
 	var out []schema.ChunkDoc
 	rest := text
-	for tokenizeStr(rest) > target {
+	for {
+		// One BPE pass per iteration instead of two. TrimContentToTokenLimit
+		// encodes rest once and returns rest unchanged exactly when rest fits
+		// within target tokens, so its result doubles as the loop-exit
+		// condition - the loop no longer runs a separate full tokenizeStr(rest)
+		// re-encode just to decide whether to keep going. The shrinking
+		// remainder still has to be re-tokenized from scratch each round:
+		// cl100k BPE is not prefix-stable (encode(text)[k:] is not
+		// encode(text[byteBoundaryOfTokens[:k]:])), so a token slice cannot be
+		// reused across cuts - but that is one encode per iteration, not two.
 		head := tokenizer.TrimContentToTokenLimit(rest, target)
-		if head == "" || head == rest {
+		if head == rest {
+			break // rest already fits; the trailing append emits it
+		}
+		if head == "" {
 			break // cannot shrink further; avoid an infinite loop
 		}
 		// TrimContentToTokenLimit decodes a token prefix; if that lands on a
--- a/internal/ingestion/component/chunker/token.go
+++ b/internal/ingestion/component/chunker/token.go
@@ -83,7 +83,6 @@ import (
 	deepdoctype "ragflow/internal/deepdoc/parser/type"
 	"ragflow/internal/ingestion/component/globals"
 	"ragflow/internal/ingestion/component/schema"
-	"ragflow/internal/tokenizer"
 
 	"ragflow/internal/parser/chunk"
 )
@@ -1326,16 +1325,13 @@ func hardSplitPiece(text string, docType string, target int) []schema.ChunkDoc {
 	var out []schema.ChunkDoc
 	rest := text
 	for {
-		// One BPE pass per iteration instead of two. TrimContentToTokenLimit
-		// encodes rest once and returns rest unchanged exactly when rest fits
-		// within target tokens, so its result doubles as the loop-exit
-		// condition - the loop no longer runs a separate full tokenizeStr(rest)
-		// re-encode just to decide whether to keep going. The shrinking
-		// remainder still has to be re-tokenized from scratch each round:
-		// cl100k BPE is not prefix-stable (encode(text)[k:] is not
-		// encode(text[byteBoundaryOfTokens[:k]:])), so a token slice cannot be
-		// reused across cuts - but that is one encode per iteration, not two.
-		head := tokenizer.TrimContentToTokenLimit(rest, target)
+		// One BPE pass per iteration instead of two: trimToTokenLimit (the
+		// package-level tokenizer seam, see title_cap.go) returns rest
+		// unchanged exactly when rest fits within target tokens, so its
+		// result doubles as the loop-exit condition. The shrinking remainder
+		// is still re-encoded from scratch each round — one encode per
+		// iteration, not two.
+		head := trimToTokenLimit(rest, target)
 		if head == rest {
 			break // rest already fits; the trailing append emits it
 		}
--- a/internal/ingestion/component/chunker/token_encoder_missing_test.go
+++ b/internal/ingestion/component/chunker/token_encoder_missing_test.go
@@ -1,0 +1,67 @@
+package chunker
+
+import (
+	"strings"
+	"testing"
+	"unicode/utf8"
+
+	"ragflow/internal/tokenizer"
+)
+
+// TestHardSplitPiece_EncoderMissingFallback pins the encoder-missing fallback
+// path of hardSplitPiece. With no cl100k table on disk, NumTokensFromString
+// returns 0 and TrimContentToTokenLimit falls back to a UTF-8-safe byte-length
+// trim. The pre-optimization loop guard (tokenizeStr(rest) > target) never
+// fired in that state, so the whole over-limit remainder was emitted as ONE
+// piece; the current loop drives off TrimContentToTokenLimit's result and keeps
+// emitting bounded pieces instead — strictly more correct, and the behavior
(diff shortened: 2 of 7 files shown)


xugangqiang:
Thanks for the clean refactor. Two follow-ups from review, both aimed at making the encoder-missing path a first-class, tested behavior rather than an undocumented aside:

### 1. Qualify "byte-identical" in the PR body

The "Correctness" section claims output is **byte-identical** to the previous algorithm. That is only true **under a loaded cl100k encoder**. The "Behavioral note" already states that the encoder-missing fallback path *intentionally differs* (old code emitted the whole over-limit remainder as one piece; new code falls back to `TrimContentToTokenLimit`'s byte-length trim and emits bounded pieces). Suggested wording:

> Output is **byte-identical** to the previous algorithm **under a loaded cl100k encoder** (the encoder-missing fallback path below intentionally differs). … Note: the differential test itself requires the encoder to be loaded, so it is only meaningful in a built-with-deps environment.

### 2. Add a minimal test for the encoder-missing fallback path

`hardSplitPiece`'s changed fallback behavior is currently only described in prose and is untested.

changshenhan:
Friendly ping — happy to answer any questions on this. It reduces hardSplitPiece from two BPE passes per iteration to one (Text ~1.8x, CJK ~2.0x faster with fewer allocations), with a differential test pinning output to the reference implementation byte-for-byte. Would appreciate a review when there's bandwidth — thanks!

xugangqiang:
please rebase

changshenhan:
Rebased on latest main (b117e91) — no conflicts. Thanks!

changshenhan:
Both follow-ups from review are done, thanks for the careful read:

1. **"byte-identical" qualified in the PR body** — now scoped to a loaded cl100k encoder, with the note that the differential test itself requires the encoder, as suggested.

2. **Minimal test for the encoder-missing fallback** — `TestHardSplitPiece_EncoderMissingFallback` (87af679d) drives `hardSplitPiece` with the table genuinely absent (scoped BPE search roots + encoder-cache reset) and locks the changed behavior: bounded pieces via `TrimContentToTokenLimit`'s UTF-8-safe byte trim, lossless concatenation, per-piece byte budget and UTF-8 validity. Under the old guard (`0 > target` never true) this input emitted a single giant piece, so the test pins exactly the delta you pointed at.

The one production-file touch is renaming `resetCL100KEncoderForTest` → exported `ResetCL100KEncoderForTest` so a cross-package test can reach the missing-table state deterministically (same pattern as the existing `SetBpeSearchRootsForTest`); the in-package fail-fast test just switches names.

xugangqiang:
Thanks for the change — the production diff itself is correct. I checked the equivalence and it holds: `TrimContentToTokenLimit` returns `s` unchanged exactly when `len(tokens) <= limit` (`internal/tokenizer/tokenizer.go:643-646`), so `head == rest` is precisely `!(tokenizeStr(rest) > target)`, and the loop body is byte-for-byte the same. A full-package run on this branch showed no pre-existing chunker test regressing.

The concerns below are all about the tests that come with it.

**1. Blocker — `TestHardSplitPiece_EncoderMissingFallback` fails in a full-package run**

Minimal reproduction (2 tests, this branch):

```
=== RUN   TestBareDelimiterSplitsAndTokenMergesText
--- PASS: TestBareDelimiterSplitsAndTokenMergesText (0.06s)
=== RUN   TestHardSplitPiece_EncoderMissingFallback
    token_encoder_missing_test.go:36: tokenizeStr = 2 under a missing table; expected the 0-error path
--- FAIL: TestHardSplitPiece_EncoderMissingFallback (0.04s)
```

Root cause: `tiktoken-go@v0.1.8` caches the loaded `*Encoding` in a package-level `encodingMap` (`encoding.go:79`; `getEncoding` returns on hit at `encoding.go:95-111` and exposes no clear/reset). `ResetCL100KEncoderForTest` (`internal/tokenizer/tokenizer.go:571`) only resets `internal/tokenizer`'s own `sync.Once`, so it cannot evict that layer. The "missing table" state is therefore only reachable while this test is the *first* thing in the process to load cl100k — and several chunker test files sort before `token_encoder_missing_test.go` and load it first. In a full `build.sh --test` run this test fails, and CI provisions the table on purpose (`.github/workflows/sep-tests.yml:459-485`), so CI hits exactly this path.

Suggested fix: don't model "encoder missing" through global state. Give the chunker package an injectable seam (the same pattern already used at `internal/ingestion/component/chunker/title_cap.go:39`, `realTrimToTokenLimit`) — a package-level `var trimToTokenLimit = tokenizer.TrimContentToTokenLimit` that the test swaps for a failing stub. That makes the fallback deterministic and order-independent, and lets you drop the exported `ResetCL100KEncoderForTest` hook from the production file (and the two renames in `internal/tokenizer/failfast_test.go`).

**2. Blocker — `TestHardSplitPieceMatchesReference` costs 204s in the default unit tier**

Measured on this branch:

```
--- PASS: TestHardSplitPieceMatchesReference (203.95s)
```

The whole `internal/ingestion/component/chunker` package goes from ~5s to ~209s. The file carries no build tag, so it runs in `go test ./...` and in CI (`sep-tests.yml:491`). Per AGENTS.md a test this slow must carry a tier tag — please add `//go:build manual` (manual is local opt-in only and never wired into CI). While you are there, add a skip guard for the case where the cl100k table is not provisioned (same pattern as `internal/tokenizer/failfast_test.go`); without it the test fails with a misleading piece-count mismatch instead of skipping.

**3. `hardSplitPieceRef` is a permanently retained second implementation**

`token_perf_test.go:54-88` keeps the retired algorithm as a second Go implementation of the same function. That is the dual path AGENTS.md tells us to remove ("Treat legacy code as liability", "If old and new implementations coexist, converge to one path unless an external contract forces compatibility").

A differential-against-reference test is valuable *scaffolding while the refactor is under review*, but once it lands every future change to `hardSplitPiece` has to be mirrored into `hardSplitPieceRef` — the reference stops being a reference and becomes a maintenance burden. I'd suggest deleting `hardSplitPieceRef` and the differential test before merge and keeping a small invariant test instead: per-piece token count within target (allowing the one-token drift you documented for re-encoded continuation pieces), lossless concatenation, and monotone progress. Golden fixtures would be a worse trade here (unreadable and brittle against any intentional change), so I would not go that route.

**4. Follow-up, separate PR — the loop is still O(L²/target)**

Agreed this is out of scope for this PR; filing it as its own change makes sense. For the record: this halves the per-iteration constant, but the remainder is still re-encoded from scratch each round. Encoding the whole text once and walking the token array (`decode(tokens[i*target:(i+1)*target])`) would be O(L) and would also remove the "continuation piece carries one extra token" drift currently documented in the comment. The cost is that `adjustCutPastTag` (`token.go:1380`) works on byte offsets and would need cumulative token counts to keep protecting `@@...##` tags.

**Minor**

- `token.go:1329-1337`: 9 lines of comment for a 3-line change, and roughly half of it argues why a token slice cannot be reused — i.e. documents an approach that was not taken. Worth trimming to 2-3 lines.
- `token_encoder_missing_test.go` reuses `hardSplitBenchText()` from `token_perf_test.go:24`, a benchmark helper. Move it to a shared test file or inline it.
- The branch is 47 commits behind `main` as of today; please rebase once more before merge.

changshenhan:
All four points addressed in c3273b4e (rebased on current main):

**1. Encoder-missing test — done via the seam, no global state.** The test now swaps a byte-trim stub into the package's existing `trimToTokenLimit` var (the `title_cap.go` seam pattern you pointed at — turned out the seam already existed, so no new indirection was needed). `ResetCL100KEncoderForTest` is dropped and the two `failfast_test.go` renames reverted: `internal/tokenizer` is now untouched by this branch, and the fallback test is order-independent in full-package runs.

**2 + 3. Differential harness deleted, invariant test in its place.** Agreed on the liability argument — `hardSplitPieceRef` and `TestHardSplitPieceMatchesReference` are gone, which also dissolves the 204s default-tier runtime (the package is back to well under a second, so no tier tag needed). The new `TestHardSplitPieceInvariants` pins what callers rely on: bounded pieces, lossless concatenation, and monotone progress. Two findings while writing it, both pre-existing behavior that the old dual-path test never surfaced:
- Pieces are **byte**-prefixes, not rune-prefixes: a cut landing mid-multibyte leaves raw continuation bytes at the piece tail (observed on the emoji corpus with the unmodified base algorithm too). So the boundedness bound is `target + utf8.UTFMax` — up to `UTFMax-1` dangling bytes re-encode as single-byte tokens, plus the one-token continuation drift — and per-piece UTF-8 validity is *not* asserted, since the concat-repairs-tail property is the actual contract. Tag inputs additionally allow `adjustCutPastTag`'s leading-tag-whole extension slack.
- The corpus is shrunk accordingly (same shape mix, ~10x smaller) so the invariants stay cheap.

**4. The O(L) rework** — agreed it's out of scope here; the byte-offsets-vs-cumulative-token-counts tension around `adjustCutPastTag` is the real design question. Happy to take it as a follow-up PR if you'd rather not own it — say the word.

**Minors:** the one-pass comment is trimmed to the loop's actual mechanism; `hardSplitBenchText` moved to a shared test helper file; branch rebased on main (was 47 behind).

xugangqiang:
LGTM

xugangqiang:
pls fix ci issue - left hook

changshenhan:
Fixed — 76d54476: the lefthook gofmt hook was tripped by a trailing blank line in `token_perf_test.go`. Removed it; preflight is green and the backend test jobs are running now.

xugangqiang:
LGTM

PR #18649, opened 2026-08-22
fix(manual): match .docx only so legacy .doc fails cleanly

### LINKED ISSUE #18621: [Bug]: Manual parser routes legacy .doc files to python-docx as DOCX

ayam04:
Fixes #18621

Legacy .doc files are OLE/CFB compound documents; python-docx only parses OOXML .docx ZIP packages. The Manual parser's suffix regex `\.docx?$` (introduced in PR #4325) matches both extensions, so .doc uploads were routed to `Docx()` and crashed with raw `BadZipFile` / OPC relationship errors instead of a clear message.

Change: `rag/app/manual.py` — suffix check now matches `\.docx$` only. Legacy .doc input fails early; the user is directed to convert to .docx/PDF (no silent OLE misparse, no implicit DOC support).

Validated:
- `python3 -m py_compile rag/app/manual.py` passes
- regex behavior check: `.docx` and `.DOCX` still match; `.doc` no longer matches

Note: full ragflow backend pytest requires the docker stack (MySQL/Redis/MinIO/ES per AGENTS.md); not available on this machine, so I ran the narrowest meaningful checks (compile + regex semantics).


ayam04:
Implemented the suggested follow-up in 1dea400: a dedicated case-insensitive `.doc$` branch now raises a clear `NotImplementedError` ("Please convert the file to .docx or PDF") before the `.docx` branch, so legacy .doc input fails early with the conversion message instead of falling through.

Verified branch behavior for `.doc` / `.DOC` / `.docx` / `.DOCX` / non-word files, plus `py_compile`.

ayam04:
Hi @xugangqiang @Haruko386 — ping on #18649: Fixes #18621 (legacy .doc files were routed to python-docx via the `.docx?$regex and crashed with opaque OPC errors; now rejected early with a clear 'convert to .docx or PDF' message). Coderabbit comments addressed. Would appreciate a review when you have a moment.

xugangqiang:
could you please help to confirm: will the system fail 100% if the input is doc file in this case? If yes, then we are fine to merge ; if no, then the code should look like "if can parse use result; else catch exception and return readable error message" . Please also add unit test if possible.

xugangqiang:
pls fix ci issue -- summary: (done in 0.20 seconds)
✔️ ruff (0.02 seconds)
✔️ check-symlinks (0.04 seconds)
✔️ mixed-line-ending (0.05 seconds)
✔️ check-merge-conflict (0.05 seconds)
✔️ check-json (0.05 seconds)
✔️ trailing-whitespace (0.06 seconds)
✔️ end-of-file-fixer (0.06 seconds)
✔️ check-case-conflict (0.06 seconds)
✔️ check-yaml (0.18 seconds)
🥊 ruff-format (0.01 seconds)
Error: Process completed with exit code 1.

ayam04:
Both requests addressed in 02f3212:

1. **Unit test added** — `test/unit_test/rag/app/test_manual.py` (first test for `rag/app/manual.py`):
   - `chunk("x.doc", binary=b"", from_page=0, to_page=MAXIMUM_PAGE_NUMBER, callback=dummy)` raises `NotImplementedError` whose message contains "convert"
   - a `.docx` filename does not hit the new `.doc` branch (asserts empty result via stubbed parser, using the same heavy-dep stub pattern as `test_markdown_image_ssrf.py`)
2. **ruff-format CI fix** — the rejection `raise` collapsed onto one line (was split by my editor; the repo's ruff-format wants the single line).

Note on local validation: the full unit run needs the `infinity` native tokenizer build + docker stack per AGENTS.md, which this machine lacks — I verified the test file is ruff-clean and formatted, and the logic matches the requested shape exactly. CI will exercise it.

ayam04:
Confirmed: yes, the system fails 100% of the time on a legacy .doc in the old code path, so the early rejection is the right shape.

**Evidence (python-docx 1.2.0, offline):**
```
ole_bytes = bytes.fromhex('d0cf11e0a1b11ae1') + b'\x00'*256   # real .doc OLE/CFB header
Document(BytesIO(ole_bytes))
-> BadZipFile: File is not a zip file
```

A legacy .doc is an OLE/CFB compound document (header `d0cf11e0a1b11ae1`); .docx is an OOXML ZIP package. They share no structure, so no valid .doc can ever pass python-docx's zip check — the failure is deterministic, not probabilistic. A try-parse fallback would therefore always hit the same exception and produce the same error, just later and with the confusing BadZipFile instead of the conversion message. The `NotImplementedError` + 'convert to .docx or PDF' message is the fastest, clearest outcome.

Unit tests are pushed in 02f3212 covering both the rejection and the .docx-not-hitting-it case.

ayam04:
CI is now fully green on the latest commit (6fc27df): unit tests pass including the two new ones in test_manual.py (the mock shape is 'Docx() returns a callable whose call yields (ti_list, tbls)', matching how manual.py uses it), preflight/lefthook clean, and the infinity + elasticsearch test matrices pass.

Summary of both review requests:
1. Unit tests added for the .doc rejection + .docx-doesn't-hit-it (test/unit_test/rag/app/test_manual.py)
2. ruff-format CI fixed (line-length 200 honored)

And on the design question: python-docx rejects every legacy .doc deterministically (BadZipFile, verified offline with a real OLE/CFB header), so the early NotImplementedError is the correct and only reachable outcome — no .doc could ever be rescued by a try-parse fallback.

Issue #17173, opened 2026-07-21
[Bug]: After the first batch of PDF slices, the content displays "no chunk built from".


one985:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

1

### RAGFlow image version

0.26.1

### Other environment information

```Markdown

```

### Actual behavior

When using RAGFlow to chunk a PDF under normal conditions, the process is carried out automatically in batches, for example:
16:12:43 Task has been received.
16:12:43 Page(1~13): Start to parse.
16:12:43 Page(1~13): [MinerU] Received binary PDF -> /tmp/mineru_bin_pdf_acwre6wd/01-安全生产应急救援预案（一期）(2).pdf
16:12:43 Page(1~13): [MinerU] Output directory: /tmp/mineru_pdf_8sh4zr4e
16:12:43 Page(1~13): [MinerU] invoke api: http://
16:13:00 Page(1~13): [MinerU] zip file returned, saving to /tmp/mineru_pdf_8sh4zr4e/01-安全生产应急救援预案（一期）(2)_auto_0m95v74l.zip...
16:13:00 Page(1~13): [MinerU] Unzip to /tmp/mineru_pdf_8sh4zr4e/01-安全生产应急救援预案（一期）(2)_auto_0m95v74l...
16:13:00 Page(1~13): [MinerU] Parsed 555 blocks from PDF.
16:13:00 Page(1~13): Finish parsing.
16:13:01 Page(1~13): Generate 58 chunks
16:13:05 Page(1~13): Embedding chunks (3.34s)
16:13:13 Page(1~13): Indexing done (8.04s).
16:13:13 Page(1~13): Task done (29.89s)
16:12:43 Task has been received.
16:13:00 Page(13~25): Start to parse.
16:13:01 Page(13~25): [MinerU] Received binary PDF -> /tmp/mineru_bin_pdf_h1drlwow/01-Work Safety Emergency Rescue Plan (Phase I)(2).pdf
[...]

PR #18650, opened 2026-08-22
fix(deepdoc): surface empty page-batch results from VisionParser through the task callback

Harsh23Kashyap:
## Summary

`VisionParser.__call__` silently returns an empty list when the page batch produces no usable output. The user sees only the generic "No chunk built from …" line in the task log with no hint about what went wrong.

Refs #17173 (qwen PDF parser: first page batch produces 12 chunks, subsequent batches return "No chunk built from …").

## Two silent failure modes

1. **Rasterization failure** — `VisionParser.__images__` catches pdfplumber exceptions and sets `self.page_images = None`. The loop iterates 0 times and returns an empty list. The internal `logging.exception("VisionParser __images__")` records the traceback, but the user only sees the generic "No chunk built from …" message — they have to dig into server logs to find the real cause.
2. **Empty vision model output** — rasterization succeeds but the configured vision model (e.g. qwen-VL) returns empty text for every page in the batch. The user again sees "No chunk built from …" with no hint that the vision model is misconfigured or returning empty.

## Fix

`deepdoc/parser/pdf_parser.py:VisionParser.__call__` — surface both cases through the task callback:

- If `self.page_images` is None after `__images__`, call `callback(-1, "VisionParser could not rasterize the PDF page batch; check server logs for the pdfplumber exception.")` and return.
- If the vision model returned empty text for every page in the batch, call `callback(1.0, "VisionParser processed N page(s) (pages X-Y) but the vision model returned no text. Check the configured layout_recognizer model configuration.")` and return.

The "No chunk built from …" line still fires (it's emitted by the orchestrator when chunks is empty), but now the user has a clear, actionable hint about *why* chunks is empty.

## Test

Three tests in `test/unit_test/deepdoc/parser/test_vision_parser_empty_page_batch.py`:

- `test_vision_parser_warns_when_images_rasterization_fails` — pre-fix: callback receives only "Processed" lines (or no messages at all). Post-fix: callback receives the `progress=-1` rasterization-failure message.
- `test_vision_parser_warns_when_vision_model_returns_empty` — pre-fix: callback receives "Processed: 1/3, 2/3, 3/3" but no warning. Post-fix: callback receives the "vision model returned no text" message.
- `test_vision_parser_does_not_warn_when_some_pages_succeed` — regression guard that the empty-batch warning does NOT fire when at least one page in the batch produced text.

The test loads the real `VisionParser` via `importlib.util.spec_from_file_location` with the module's heavy runtime dependencies stubbed (same pattern as `test_pdf_parser_table_coordinates.py`). The `rag.app.picture` module is pre-populated so the lazy `from rag.app.picture import vision_llm_chunk` inside `__call__` resolves to a mock without pulling in `rag.app.picture`'s own dependency chain.

3/3 tests pass on the fix. Verified pre-fix behavior: the first two tests fail (the pre-fix code emits the "Processed" progress messages without the new warnings). ruff check + ruff format --check clean.

The underlying root cause of the qwen-VL empty output in #17173 (why the vision model returns nothing for page batches 2+) is a model / configuration concern, not a RAGFlow rendering bug. The original behavior — "No chunk built from …" — is unchanged in the sense that the chunk list is still empty. The fix only changes the diagnostic surface so the user can see *why* the chunks are empty.

If the qwen-VL specific issue is reproducible end-to-end with a known setup, a follow-up PR could add a sanity-check or retry around the vision model call. This PR is the minimal scope: make the failure mode visible.

### DIFF
--- a/deepdoc/parser/pdf_parser.py
+++ b/deepdoc/parser/pdf_parser.py
@@ -2118,7 +2118,15 @@ class VisionParser(RAGFlowPdfParser):
 
         all_docs = []
 
-        for idx, img_binary in enumerate(self.page_images or []):
+        # If `__images__` failed (pdfplumber exception path), `self.page_images`
+        # is None and the loop below silently returns no chunks. Surface the
+        # failure through the callback so the user sees a useful message in
+        # the task log instead of a generic "No chunk built from …".
+        if not self.page_images:
+            callback(-1, "VisionParser could not rasterize the PDF page batch; check server logs for the pdfplumber exception.")
+            return all_docs, []
+
+        for idx, img_binary in enumerate(self.page_images):
             pdf_page_num = from_page + idx  # 0-based
             if pdf_page_num < start_page or pdf_page_num >= end_page:
                 continue
@@ -2138,6 +2146,20 @@ class VisionParser(RAGFlowPdfParser):
             if text:
                 width, height = self.page_images[idx].size
                 all_docs.append((text, f"@@{pdf_page_num + 1}\t{0.0:.1f}\t{width / zoomin:.1f}\t{0.0:.1f}\t{height / zoomin:.1f}##"))
+
+        # If the vision model produced no output for any page in the batch,
+        # log a clear warning so the user can distinguish "no images on these
+        # pages" from "vision model returned nothing for the configured
+        # layout_recognizer". This is the silent-failure mode behind #17173
+        # (qwen-VL returning empty for page batches 2+ while the first batch
+        # succeeded).
+        if not all_docs and self.page_images:
+            callback(
+                1.0,
+                f"VisionParser processed {len(self.page_images)} page(s) "
+                f"(pages {from_page + 1}-{min(to_page, total_pdf_pages)}) but the vision model returned no text. "
+                f"Check the configured layout_recognizer model configuration.",
+            )
         return all_docs, []
 
 
--- a/test/unit_test/deepdoc/parser/test_vision_parser_empty_page_batch.py
+++ b/test/unit_test/deepdoc/parser/test_vision_parser_empty_page_batch.py
@@ -1,0 +1,243 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for ``VisionParser.__call__`` empty-page-batch handling
+in ``deepdoc/parser/pdf_parser.py``.
+
+Issue: infiniflow/ragflow#17173.
+
+When the user sets the PDF parser to a vision model (e.g. "qwen" series),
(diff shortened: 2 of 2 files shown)


xugangqiang:
Thanks for the PR. I reviewed the changes against `VisionParser.__call__` and tested the name-mangling concern locally. Three points on the new failure paths:

**1. Add `logging` for both new failure flows — Reasonable (suggestion)**
The callback only reaches the task log. When there is no task callback (some internal reuse paths), operators lose the root cause in the server log. Add `logging.error(...)` right before each `callback(...)` so the failure is diagnosable regardless of callback presence.

**2. Explicit `callback=None` raises `TypeError` — Reasonable (edge case, but real)**
`callback = kwargs.get("callback", lambda prog, msg: None)` returns `None` when the caller *explicitly* passes `callback=None` (key present → default not used). The original success path already guards with `if kwargs.get("callback):`, so `callback=None` was previously tolerated. The new failure paths call `callback(...)` unconditionally, so they now crash on `None`. Fix:
```python
callback = kwargs.get("callback") or (lambda prog, msg: None)
```

**3. `if not self.page_images` misreports a valid empty page range as rasterization failure — Reasonable (real bug)**
On a successful `__images__`, `self.page_images` can be `[]` when the page range is empty (e.g. `from_page >= total`). `if not self.page_images:` then fires `callback(-1, "could not rasterize…")` with a misleading message. Pre-fix, `enumerate(self.page_images or [])` handled this silently. Fix:
```python
if self.page_images is None:
    callback(-1, "VisionParser could not rasterize the PDF page batch; check server logs for the pdfplumber exception.")
    return all_docs, []
```
This also makes the empty-range case fall through to the loop (0 iterations) and return `([], [])` silently as before.

**Note on the CodeRabbit name-mangling comment:** it is a false positive. `__images__` has two trailing underscores, so Python does **not** apply name mangling (`_VisionParser__images__` is never created). `parser.__images__ = lambda …` therefore correctly overrides the method, and the three tests do exercise their intended paths.

Recommendation: fix #3 (functional) and #2 (defensive) before merge; #1 is a recommended improvement. Add a test for `page_images=[]` asserting no "could not rasterize" callback is emitted.

xugangqiang:
Review findings for this PR (read-only review; tests verified by running them locally — **3 passed**).

## Suggest to fix before merge (correctness)
- `deepdoc/parser/pdf_parser.py:2126` — change `if not self.page_images:` to `if self.page_images is None:`.
  `__images__` only sets `page_images = None` on the pdfplumber exception path. A valid but empty page range yields `[]`, which the current `not self.page_images` check misreports as a rasterization failure (sends `progress=-1` + "could not rasterize"). With `is None`, an empty `[]` range skips the early return, loops 0 times, and the later `if not all_docs and self.page_images:` guard (falsy `[]`) correctly stays silent — matching the original behavior.

## Optional / non-blocking improvements
- `test/unit_test/deepdoc/parser/test_vision_parser_empty_page_batch.py` (test 2): also assert `prog == 1.0` alongside the message text, so a future regression can't leave the task incomplete.
- `deepdoc/parser/pdf_parser.py:2157`: the `and self.page_images` is redundant (the early return already guarantees it's non-None/non-empty there); can be dropped.
- Optionally add `logging.warning` on the empty-vision-output path (the rasterization path already logs via `logging.exception` at `__images__`).

## coderabbit comments that are false positives (no action needed)
- "Normalize an explicit null callback": the callback is already normalized at `pdf_parser.py:2111` (`callback = kwargs.get("callback", lambda prog, msg: None)`), and the new failure paths use that local `callback`, never `None`.
- "Stub the name-mangled rasterizer method" (`_VisionParser__images__`): `__images__` has double trailing underscores, so per Python name-mangling rules it is **not** mangled — the method name is literally `__images__`, and `parser.__images__ = lambda` does override it. Confirmed: the 3 tests pass, and test 2's assertion for "vision model returned no text" can only come from the empty-output branch (the rasterization branch emits a different message), proving the stub takes effect.

PR #18662, opened 2026-08-23
fix(opensearch): stop zeroing vector scores by mapping similarity to knn boost

barandras:
### Summary

Fix OpenSearch retrieval returning `vector_similarity = 0.000` for every chunk when hybrid search is enabled.

On the OpenSearch backend, retrieval uses a second KNN-only search (`Dealer._knn_scores()`) to recover per-chunk cosine scores for reranking. That pass intentionally sends `MatchDenseExpr(..., {"similarity": 0.0})` to mean “no minimum similarity cutoff.”

However, `OSConnection.search()` was incorrectly mapping `similarity` to the KNN clause `boost` field:

```python
knn_query[vector_column_name]["boost"] = similarity
```

With `similarity=0.0`, this produced `boost=0.0`, which zeroed out KNN `_score` values. `get_scores()` then returned `0.0` for every hit, so `vector_similarity` was always zero and hybrid ranking ignored the vector component — with no exception raised.

This is separate from the `get_scores()` `AttributeError` crash addressed in #14970 / #15390; here retrieval succeeds but vector scores are silently lost.

### Root cause

- `similarity` (threshold-like retrieval parameter) was conflated with `boost` (score multiplier).
- The Elasticsearch connector does not set KNN boost this way; only OpenSearch did.
- Existing OpenSearch unit tests validated hybrid query *shape* but not KNN score semantics.

### Changes

- **`rag/utils/opensearch_conn.py`**: Stop deriving KNN `boost` from `similarity`. Omit `boost` unless explicitly provided via `MatchDenseExpr.extra_options["boost"]`.
- **`test/unit_test/rag/utils/test_opensearch_hybrid_search.py`**: Add regression coverage:
  - KNN query omits `boost` when only `similarity` is set.
  - Explicit `boost` is still honored when provided.
  - Non-zero KNN `_score` values propagate unchanged through `get_scores()`.

### DIFF
--- a/rag/utils/opensearch_conn.py
+++ b/rag/utils/opensearch_conn.py
@@ -395,9 +395,6 @@ class OSConnection(DocStoreConnection):
             # Besides, Opensearch's DSL for KNN_search query syntax differs from that in Elasticsearch, I also made some adaptions for it
             elif isinstance(m, MatchDenseExpr):
                 assert bqry is not None
-                similarity = 0.0
-                if "similarity" in m.extra_options:
-                    similarity = m.extra_options["similarity"]
                 explicit_boost = None
                 if isinstance(m.extra_options, dict) and "boost" in m.extra_options:
                     explicit_boost = m.extra_options["boost"]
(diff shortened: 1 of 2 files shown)


qinling0210 (approved):
LGTM

PR #18663, opened 2026-08-23
fix(chat): stop rag_agent forwarding message bookkeeping keys to the provider

### LINKED ISSUE #18653: [Bug]: conversation_id error with Groq provider in v0.27.0 (regression after model provider revamp #16604)
### Self Checks

- [x] I confirm that I am using English to submit this report (Language Policy).
- [x] Non-english title submissions will be closed directly (Language Policy).

(unknown — deployed via Docker image v0.27.0)

### RAGFlow image version

v0.27.0

Deployed via Docker on a Linux sandbox (WSL2/Docker Desktop origin, running on remote Linux sandbox), CPU with AVX2/AVX512 support.
Chat model provider: Groq (openai/gpt-oss-20b)
Embedding model provider: Ollama (nomic-embed-text)

### Actual behavior

Every chat using the Groq provider fails on the very first message, in a brand-new chat, with or without an associated dataset:

ERROR: INVALID_REQUEST - litellm.BadRequestError: GroqException - {"error":{"message":"messages.1' : for 'role:user' the following must be satisfied[('messages.1' : property 'conversationId' is unsupported)]","type":"invalid_request_error"}} LiteLLM Retried: 5 times

This reproduces consistently across multiple newly created chats. Switching the same chat's model to an Ollama model (e.g. `qwen2.5:1.5b`) resolves the issue immediately, confirming the problem is specific to the Groq/LiteLLM path.

This appears to be a regression of the class of bug fixed in #15162 ("Fix: /chat/completions not aware of conversation_id"), which was already released in v0.25.6 — well before v0.27.0. Since v0.27.0 shipped a full rewrite of the model provider system (#16604), it's likely that this rewrite reintroduced the same category of issue: an internal parameter (`conversationId`) not being filtered out before being forwarded to the external Groq API.

### Expected behavior

The `conversationId` (or equivalent internal field) should be filtered out of the payload sent to the Groq API, as was done for the original fix in #15162.

### Steps to reproduce
Configure Groq as a chat model provider (e.g. model "openai/gpt-oss-20b")
Create a new chat assistant using this model, with or without an associated dataset
Send any message (e.g. "Hello")
Observe the INVALID_REQUEST / conversationId error

### Additional information

Related: #15162 (previously fixed this exact class of bug in v0.25.6), #16604 (v0.27.0 model provider system revamp, likely source of this regression).

bharadwaj-pendyala:
### Summary

Fixes #18653. Groq rejects every chat request on v0.27.0 with `property 'conversationId' is unsupported`, while the same chat works on Ollama.

**Problem**

`rag_agent` is the only chat entrypoint that hands the stored message dicts to the provider unchanged:

- `async_chat_solo` rebuilds each message as role/content: `api/db/services/dialog_service.py:327`
- `async_chat` does the same: `api/db/services/dialog_service.py:846`
- `rag_agent` did `agent_messages = deepcopy(messages)` at `:1951`, then passed that to `chat_mdl.async_chat` (`:2206`, `:2208`) and `chat_mdl.async_chat_streamly_delta` (`:2091`, `:2093`)

Those dicts always carry at least `id`, and `created_at` on the stored prologue (`api/db/services/conversation_service.py:243`, `:268`). The web client posts its own message objects with `pass_all_history_messages: true` (`web/src/pages/next-chats/hooks/use-send-single-message.ts:72-89`), each stamped with `conversationId` (`web/src/hooks/logic-hooks.ts:555`), and `api/apps/restful_apis/chat_api.py:1279` copies them into the history verbatim. OpenAI and Ollama ignore properties they do not know. Groq validates the message schema and rejects them.

That is also why it reads as a v0.27.0 regression rather than a new bug. `rag_agent` delegates to `async_chat` when reasoning is off (`:1943`), so the leak only bites once the reasoning path is taken, and the web client sets `reasoning` from the thinking toggle.

**Solution**

Project each message onto the fields the chat-completions schema defines, instead of copying it wholesale.

An allowlist rather than removing the four known keys, because the set of bookkeeping keys the UI and the conversation record attach is open-ended, while the schema is not. Content is passed through untouched, so multimodal content-part arrays still reach `convert_last_user_msg_to_multimodal` (`:505-521`) as lists, and `tool_calls` / `tool_call_id` survive so a tool history stays a valid provider conversation.

**Scope**

One line inside `rag_agent` plus the constant. I deliberately left three things alone:

- No citation-marker stripping. `async_chat` and `async_chat_solo` also run `re.sub(r"##\d+\$\$", ...)` over history content, which `rag_agent` does not. Adding it here would change what the model sees, which is a separate question from the schema violation, and it would break list content.
- No guard in `rag/llm/chat_model.py`. That layer deliberately builds `tool_calls` and `tool_call_id` (`:402`, `:424`), so a blanket sanitiser there would be wrong. Happy to move it if you would rather have the boundary in the model layer.
- Existing behaviour on the non-reasoning path is untouched, including that list-valued `content` already raises `TypeError` at `:327` and `:846` before this PR. Separate bug, not this one.

**Validation**

Base commit `2db8eb6c91184402e2979b6b52b77063919c7e17`.

Four tests in `test/unit_test/api/db/services/test_dialog_service_rag_agent_messages.py` drive the real `rag_agent` with the models and RAG tools stubbed:

- bookkeeping keys are dropped and the outbound messages are exactly role/content
- `tool_calls` and `tool_call_id` survive
- the caller's list is not mutated, so `chat_api` can still persist the turn
- multimodal content-part arrays are forwarded unchanged

Reverting only the one production line, keeping the constant, turns the first test red:

```
E   AssertionError: assert [{'content': ...ole': 'user'}] == [{'content': ...ole': 'user'}]
    unsupported keys forwarded: ['created_at', 'id']
```

With the fix, `pytest test/unit_test/api/db/services/` is 148 passed, 0 failed (147 before this PR, 4 new here, 3 of the new 4 also pass on the old code as guards against the obvious wrong fixes). `ruff format --diff` reports both files already formatted, and `ruff check api/db/services/dialog_service.py` reports 99 findings both before and after, so this adds none.

I could not reproduce against a live Groq key, so the provider-side rejection is taken from the reporter's error text; everything above about the request RAGFlow builds is verified from source and by the tests.

This change was written with AI assistance, reviewed and verified by me.

### Related issue

Closes #18653


bharadwaj-pendyala:
Both findings reproduce, and both describe the pattern this directory already uses, so I would rather not make this file the one that diverges.

**The warning filter.** Module-level `warnings.filterwarnings(...)` for the `pkg_resources` deprecation is how every sibling in `test/unit_test/api/db/services/` does it:

```
test_deep_pagination_order_by.py:21
test_dataset_access_permissions.py:23
test_document_service_get_parsing_status.py:24
test_get_queue_length.py:25
test_file_service_upload_document.py:25
test_dialog_service_use_sql_source_columns.py:26
test_dialog_service_final_answer.py:41
test_gaussdb_dialog_sql.py:43
```

The one file in the repo that uses `warnings.catch_warnings()` instead is `test/unit_test/rag/app/test_qa_csv.py:21`.

**The cv2 stub.** `_install_cv2_stub_if_unavailable()` is defined and called at module scope in ten files, with the same `except Exception` and the same `sys.modules["cv2"] = stub`. Nine of them predate this PR:

```
api/db/services/test_dataset_access_permissions.py:30,64
api/db/services/test_file_service_upload_document.py:32,65
api/db/services/test_get_queue_length.py:32,56
api/db/services/test_document_service_get_parsing_status.py:31,65
api/db/services/test_dialog_service_use_sql_source_columns.py:33,68
api/db/services/test_dialog_service_final_answer.py:48,77
api/db/services/test_gaussdb_dialog_sql.py:50,266
agent/component/test_browser_use_component.py:25,47
rag/test_sync_data_source.py:33,80
```

It also does not fire in any environment built from the project's dependencies, since `opencv-python==4.10.0.84` is a hard requirement at `pyproject.toml:79`. In my venv `cv2.__version__` is `4.10.0`, so the stub branch never runs.

The leak is real once you force it, though. Blocking the import with a `meta_path` finder that raises on `cv2`:

```
cv2 in sys.modules after import: True
```

So the stub does survive the module and is visible to later tests. That is a property of all ten copies, not of this one.

**One thing neither of us flagged, which I hit while checking the above.** `_module_getattr` raises `RuntimeError` for any name that is not uppercase, and dunders are not uppercase:

```
RuntimeError: cv2.__file__ is unavailable in this test environment
```

That came from a plain `getattr(cv2, "__file__", None)`. A default only suppresses `AttributeError`, so the stub breaks ordinary introspection instead of falling through. Raising `AttributeError` there would fix it. Same code in all ten copies, so it is a separate change rather than something for this PR.

Happy to switch this file to the scoped versions if a maintainer prefers them. I would rather do all nine in one pass than leave the directory half converted, so say the word and I will open that as its own PR.

wangq8:
LGTM, please also port the go version @bharadwaj-pendyala

wangq8 (approved):
LGTM — approved. The root-cause analysis is correct and the fix is well-scoped.

Verified against `ragflow-main`:

- `agent_messages` is only consumed for content mutations (`agent_messages[-1]["content"] += text_attachments_content`, `convert_last_user_msg_to_multimodal`) and then passed to `chat_mdl.async_chat` / `async_chat_streamly_delta`. No downstream code reads `id` / `created_at` / `conversationId` / `doc_ids` off `agent_messages` — `doc_scope` is derived from the original `messages[-1]["doc_ids"]`, not the filtered copy. So dropping the bookkeeping keys is safe.
- The allowlist matches the chat-completions message schema (`role`, `content`, `name`, `tool_calls`, `tool_call_id`, `function_call`, `refusal`, `audio`), and `content` / tool history pass through untouched, which the tests confirm. The caller's `messages` list is not mutated, so `chat_api` can still persist the turn.
- The fix is consistent with `async_chat` / `async_chat_solo`, which already rebuild each message as role/content — `rag_agent` was indeed the only leak.

Tests are targeted and drive the real `rag_agent` with stubbed models/tools, covering the four key invariants (keys dropped, tool protocol preserved, caller untouched, multimodal content forwarded unchanged).

Non-blocking nit: `function_call` / `refusal` / `audio` are legacy/rarely-used fields, but including them is harmless and future-proof. No regressions found.

bharadwaj-pendyala:
@wangq8 I went looking for the Go equivalent and I do not think there is one to fix. The Go smart-reasoning path cannot leak these keys, for a structural reason rather than a guard.

`smartReasoningChat` is the Go counterpart of `rag_agent`, and at `internal/service/chat_pipeline.go:2102` it hands the raw `[]map[string]interface{}` to `convertMessagesToEino` (`chat_pipeline.go:2203`), which reads `m["role"]` and `m["content"]` and constructs a fresh `schema.UserMessage` / `schema.AssistantMessage`. It never copies the map. `schema.Message` is a struct with fixed fields (`schema/message.go:498`), so `id`, `created_at`, `doc_ids` and `conversationId` have nowhere to land.

The same holds on the other three paths I could find that reach a provider:

- `buildChatMessages` (`chat_pipeline.go:2708`) rebuilds into `modelModule.Message`, which is `Role` / `Content` / `ToolCallID` / `ToolCalls` (`internal/entity/models/types.go:15`)
- `openAICompatPriorHistory` (`internal/service/agent.go:2592`) rebuilds `{"role", "content"}` maps
- `filterMessages` and `normalizeOpenAIMessages` (`internal/service/openai_chat.go:520` and `:604`) do carry the map through, but they feed the two above

I checked by running the body of `convertMessagesToEino` verbatim against the same input as the Python regression test. `go test ./internal/service/` does not build on macOS here (`office_oxide.h` missing, and `Pdeathsig` in `internal/agent/sandbox/local.go:307` is Linux-only), so I ran it as a standalone program with the eino constructors inlined:

```
payload: [{"role":"user","content":"which page covers the refund policy?"},{"role":"assistant","content":"Page 3."}]
contains "id"             -> false
contains "created_at"     -> false
contains "doc_ids"        -> false
contains "conversationId" -> false
```

I did write the parity test for `convertMessagesToEino` and then threw it away: with the message rebuilt field by field there is no code path for it to fail on, so it asserts the shape of `schema.Message` rather than anything about this bug. Happy to add it anyway if you want the intent pinned against a future refactor that switches that helper to a map copy, just say so and I will push it.

One real divergence I did find, separate from this PR and worth its own issue if it is not already known: `convertMessagesToEino` only handles string content, so multimodal content parts are dropped before the ReAct loop. Its own doc comment says so. Python's `rag_agent` handles image attachments through `convert_last_user_msg_to_multimodal`, so agent mode loses images on the Go side. Want me to file that?

PR #18664, opened 2026-08-23
docs: replace root-relative cross links with relative page links

simpleqt:
configurations.md and backup_and_migration.md cross-linked each other
via /migration#... and /configurations#s3-tigris. These host-absolute
paths match no route on the docs site (real pages live under
/administrator/...) and 404 on GitHub. Point them at the sibling
markdown files so both Docusaurus and GitHub can resolve them; the
existing anchors are preserved.


Yannnnnnny (approved):

JinHai-CN:
After merge the latest commit, only documents are updated in this PR

PR #18667, opened 2026-08-24
fix: account for checkpointed ingestion progress

buua436:

Preserve completed component progress when an ingestion task resumes from a checkpoint. Checkpointed operators are skipped and do not emit lifecycle events again, which previously caused the document progress to remain below 100%.

This PR also simplifies Wiki REDUCE deduplication and adds entity types to entity page slugs.

### Type of change


wangq8 (commented):
Thanks for the PR. The checkpoint-progress fix itself looks correct — `RedisCheckpointExists` checks `agent:cp:{task.ID}`, which matches the pipeline's `cpID == p.taskID` keying (`canvas.WithCheckPointID(p.taskID)`), so the fresh-vs-resume distinction is sound.

However, the wiki REDUCE refactor is bundled here but is unrelated to the title, and I have a few concerns about it:

**1. `normalizeWikiPlanPages` still collapses same-named entities of different types (defeats the stated goal).**
The stated intent is "same-named entities with different types remain separate pages", and `reduceExtracts` + `entityPageSlug` now achieve that at the reduce/slug level. But `normalizeWikiPlanPages` still dedups by `wikiTitleKey(page.PageType, page.Title)`, which ignores entity type/slug. So in `buildModeAPlan` (and LLM plans), two entities like `苹果/fruit` and `苹果/company` produce slugs `entity/fruit-苹果` and `entity/company-苹果` but are then collapsed into a single page (the second page is dropped and only its `RelatedKB` folded). The new capability is undone at the plan stage, and there is no test covering it — the new tests only exercise `reduceExtracts`/`entityPageSlug`. Please extend the `byTitle` key (or otherwise disambiguate same-title pages by canonical slug/type).

**2. Ambiguous composite dedup/slug key.**
`entityPageSlug` joins `slugify(type) + "-" + slugify(name)`, and `reduceExtracts` now keys entities by that slug. Because `-` is a legal slug character, distinct `(name, type)` pairs can collide: e.g. `name="c"/type="a-b"` and `name="b-c"/type="a"` both yield `entity/a-b-c` and would be wrongly merged. The previous `{Name, Type}` struct key had no such ambiguity. Keep the dedup key as `(name, type)` (or use a separator `slugify` cannot emit, e.g. `\x00`) and derive the slug separately.

**3. Confirmed regression? Removing the embedding+LLM entity dedup.**
Deleting `dedupeEntities` removes the REDUCE-stage near-duplicate merge (e.g. "Alpha Inc" vs "Alpha Incorporated"), which was explicitly added to close the gap with Python `wiki.py` canonicalization. Please confirm this is intentional. It also orphans `wikiReduceEntityDisambiguateSystem` / `wikiReduceEntityDisambiguateUserTemplate` in `prompt.go` (now unused constants).

**4. Minor: whitespace-normalization regression in dedup.**
Switching the entity key from `normKey` (collapses internal whitespace + lowercases) to `slugify` (space → `-`) means names differing only in whitespace ("John  Smith" vs "John Smith") no longer merge.

**5. Minor: `RedisCheckpointExists`.**
`NewRedisCheckPointStore(0)` passes a meaningless TTL for a Get-only helper (harmless since `Get` ignores TTL, but misleading). Also, on a Redis error the code falls through to the fresh-run branch and clears progress — the destructive branch. Worth a brief comment that this is the intended fail-safe.

**Positive:** the `relationIndexes` change (merging `SourceChunkIDs` on duplicate relations instead of dropping the second occurrence) is a genuine improvement, and the `canonicalSlugs` migration for legacy slugs is a nice touch.

**Suggestion:** consider splitting the wiki refactor into its own PR — it's a substantial, independent behavior change that would be easier to review and bisect on its own.

PR #18668, opened 2026-08-24
fix: honour chunk_token_num in the laws tree chunker

### LINKED ISSUE #12109: [Feature Request]: “chunk_token_num” is ignored by chunk policies causing rerank failure.

shoemoney:
### Summary

`rag/app/laws.py` (the `laws` chunk_method, the UI default for contracts and regulations) calls `tree_merge(bull, sections, 2)` and never passes `chunk_token_num`, even though `parser_config` carries it (`{"chunk_token_num": 512, ...}` is the module default, and `laws.py` sets it to `0` for parsers that already pre-chunk their output: tcadp, docling, mineru, paddleocr). `tree_merge` does not take a token budget at all. `Node.build_tree` appends every line past the target depth onto the current heading node with `stack[-1].add_text(text)` with no counting, and `Node._dfs` joins that whole accumulation into one string with no cap.

So a document with one heading and a long body with no sub-headings (a routine shape for a single long Article or clause) comes out as one chunk, however long the body is. On a synthetic doc with one "Article 1" heading followed by 400 body paragraphs and no sub-headings, `tree_merge` produced a single chunk of 93,499 characters (~17,203 tokens).

That chunk then reaches `rag/llm/embedding_model.py`, which silently truncates provider input to a fixed ceiling (`truncate_to=2048`, `8191`, or `DEFAULT_MAX_TOKENS=8192` depending on provider) with no error and no log line. Everything past the truncation point is never embedded and never retrievable. The document reports as ingested but part of it is gone from the index.

#16959 addressed the same defect class on the paper path (closes #12109). This PR covers `tree_merge`, which that change did not touch. It does not modify `hierarchical_merge`, `book.py`, or `paper.py`.

### What changed

- `tree_merge(bull, sections, depth, chunk_token_num=0)`: new optional parameter. The default `0` means off, so any caller that does not pass it keeps the old unbounded behaviour. There were no other callers.
- `Node.__init__` takes `chunk_token_num=0` and stores it. `tree_merge` passes its value through when constructing the root node.
- `Node._dfs` routes every chunk emission through a small `_emit` helper: accumulate body lines, and once adding the next one would push the running token count over `chunk_token_num`, flush what is accumulated and start a new chunk. The heading title path is repeated on every emitted piece and its tokens count against that piece's budget, so a continuation chunk never loses the heading. `chunk_token_num <= 0` skips the loop and joins everything as before. Position tags (`@@page\t...##`) are stripped before counting, since they are not body content.
- `laws.py` reads `chunk_token_num = int(parser_config.get("chunk_token_num", 512) or 0)` right before the `tree_merge` call, the same pattern `book.py` and `paper.py` use, and passes it through. The `chunk_token_num = 0` branch for tcadp/docling/mineru/paddleocr is untouched; with `0` the new budget logic is a no-op, so already-chunked sections are left alone.

`Node.build_tree` is unchanged. The split happens at emission time in `_dfs`, not at accumulation time in `build_tree`.

`laws.py`'s own `Docx` class builds its `Node`/`build_tree` tree directly, not through `tree_merge`, with no `chunk_token_num` passed. It is unaffected by this change and out of scope here.

### Tests

Added `test/unit_test/rag/test_laws_chunk_size.py`. An autouse fixture swaps `num_tokens_from_string` for a whitespace word count so the tests do not depend on tiktoken. Then:

- `test_single_heading_long_body_respects_chunk_token_num`: one heading plus 40 paragraphs, no sub-headings. With a budget it must split, and every paragraph must survive somewhere.
- `test_structured_subheadings_still_split_by_structure`: three headings with short bodies. The budget must not change the existing structural split when nothing is over budget.
- `test_chunk_token_num_absent_or_zero_preserves_current_behaviour`: calling `tree_merge` with the parameter omitted and with explicit `0` produces the same single-chunk output as before this PR.
- `test_body_just_under_budget_stays_single_chunk`: sitting exactly at the budget does not trigger a split. The split condition is strict `>`.

Against the pre-fix source the tests fail with `TypeError: tree_merge() got an unexpected keyword argument 'chunk_token_num'`; against the fix they pass. Reverting the source fix with the test file in place turns them red the same way.

A regression test asserts that every chunk from the long-body case contains "Article 1".

Also ran the existing `test/unit_test/rag/nlp/` and the non-heavy-dependency parts of `test/unit_test/rag/app/` (`test_naive_merge.py`, `test_merge_paragraphs.py`, `test_delim.py`, `test_docx_question_level.py`, `test_gaussdb_retrieval.py`, `test_tokenize_chunks.py`): 154 passed, no regressions. I did not run the full suite. `test_laws_docx_tables.py` and several `rag/app` tests need `infinity`, `beartype`, `pkg_resources`, and `markdown`, which are not installed in my environment.


xugangqiang:
Reviewed this PR (read-only, verified in an isolated worktree). The problem it targets is real, but I found issues that should be addressed before merge — listing only the problems below.

## 1. Continuation chunks lose the heading title (regression)
The new `_emit` helper splits an over-budget accumulation, but only the **first** emitted chunk carries the leading title path. Every continuation chunk is appended without it.

I reproduced this against the actual PR code: one `Article 1` heading + 40 body paragraphs with `chunk_token_num=20` yields **9 chunks, 8 of which do NOT contain `Article 1`**. For `laws`/contracts/regulations — where a single long Article with no sub-headings is the *common* case — this means most of the emitted chunks lose their statutory/section context, which hurts retrieval relevance. This is a new defect introduced by the change (the old single-chunk output at least kept the title).

Root cause: `_emit(parts)` receives the title mixed into `parts` (e.g. `path_titles + texts`), and on flush it appends only `cur` (body batch) without re-prepending the title prefix. The title must be carried onto **every** emitted chunk.

## 2. Tests do not cover the regression
The 4 added tests assert that paragraphs survive, but **none assert that the heading is retained on continuation chunks**. The exact scenario that triggers the bug (`test_single_heading_long_body_respects_chunk_token_num`) passes despite 8/9 chunks missing `Article 1`. A regression assertion (every chunk contains the leading title) is missing, so the defect is invisible to CI.

## 3. The "mirror #16959" premise is inaccurate on current `main`
The description claims structural parity with #16959's `hierarchical_merge` overflow-split loop. However, #16959 is **still OPEN / unmerged**, and `main`'s `hierarchical_merge` (`rag/nlp/__init__.py:1150-1169`) still uses a hard-coded `218` threshold with no `chunk_token_num`. So the "loop from #16959" does not exist on `main`, and the parity claim cannot be verified against the base branch.

Moreover, #16959's own loop has the **same title-loss defect** (when a hierarchy group `ck` overflows, continuation groups drop `ck[0]`, the heading). This PR effectively mirrors that bug. Both should be fixed consistently.

## 4. Coordination / merge-ordering risk
Both PRs modify `rag/nlp/__init__.py`. Since #16959 is unmerged, these should be **sequenced or stacked** to avoid a file-level conflict, and the title-retention fix should be applied in both places.

---

Summary: the fix direction is correct (it does stop oversized chunks from being silently truncated by the embedding layer), but as written it trades one defect for another (lost heading context on split chunks), and the tests don't catch it. Please retain the leading title on every emitted chunk and add a regression test for title retention before merging.

shoemoney:
Fixed in 7f9709b: continuation chunks now repeat the heading title path (titles carry their own token cost into the split budget), and added a regression test asserting every chunk from the long-body case contains "Article 1".

xugangqiang:
LGTM

xugangqiang:
## Review findings (read-only, verified in an isolated worktree)

### Blocker — cap not enforced + content duplication when an ancestor node carries body text
`Node._dfs` builds `path_titles = titles + texts` (`rag/nlp/__init__.py:1991`) and `_emit` treats `title_parts` as an immutable, non-splittable prefix. When a heading node has children **and** its own body text, that body text becomes part of the title path and is repeated into every descendant chunk. The budget loop then never splits the (already oversized) title prefix, so the "no oversized chunk" promise silently fails and content is multiplied.

Repro (word-count token counter, `chunk_token_num=512`):
- `Chapter I` + 600-token intro + `Section 1` + `Article 1` + 600-token body → **22 chunks × 634 tokens (all over cap)**; total swells from 1×1208-token chunk (old behaviour) to ~13,900 tokens across chunks (~11× duplication).
- A single >cap paragraph also stays over cap (e.g. one 2000-token paragraph → 2002-token chunk), because no sentence/segment re-split is performed.

Suggested fix: enforce the cap on the **whole emitted chunk string** (title prefix kept once), e.g. as a post-pass over `tree_merge` output, instead of only over `body_parts`.

### Improvement — `.docx` path is not covered
`laws.chunk` returns early for `.docx` via `Docx()` (`rag/app/laws.py:180-184`), which builds `Node` directly (`rag/app/laws.py:130`) with no `chunk_token_num`. For laws/contracts `.docx` is a very common input, so the fix leaves the most frequent case unbounded. Please cover it or at least document the gap.

### Improvement — third budget implementation instead of reusing the existing abstraction
The repo already has two cap implementations: `rag/app/paper.py:_merge_sections_by_pivot` (#16959, merged) and the DSL `rag/flow/chunker/title_chunker/common.py:_enforce_token_cap` (#18455/#18551, merged — supports sentence-boundary re-split). This PR adds a third hand-rolled loop inside shared `Node._dfs`. Prefer reusing `_enforce_token_cap` (post-build pass + sentence re-split) or, like #16959, keeping the budget inside the parser (`rag/app/laws.py`) rather than adding a param to the shared `Node`/`tree_merge` (AGENTS.md: collapse duplicate paths).

### Nit — tests / coverage gaps
- No regression test for the nested ancestor-body case above (the only realistic trigger of the Blocker), so it is invisible to CI.
- Tests call `tree_merge` directly; no end-to-end assertion that `laws.chunk` forwards `parser_config["chunk_token_num"]`.
- Minor: the test asserts with `len(c.split())` rather than the patched counter (equivalent here, but the patched counter is more correct).

### Nit — PR description accuracy
#16959 ultimately only touched `rag/app/paper.py` (+ test) and did **not** modify `hierarchical_merge`/`rag/nlp/__init__.py`. The description's "mirrors #16959's `hierarchical_merge` overflow-split loop" is inaccurate; please update it. (The merge-ordering concern from earlier review is now moot: #16959 does not touch `rag/nlp`.)

Note: earlier resolved items (continuation chunks losing the title, missing regression assertion) were verified fixed in `7f9709b` — all 34 chunks in the repro retain the title. The CodeRabbit ruff PLR0402 remark is a false positive: this repo's `pyproject.toml` does not enable PLR0402.

shoemoney:
Confirmed the blocker and reproduced it: `Chapter I`'s own body text (an intro paragraph attached before its first child) was folded into `path_titles`, which `_emit` repeats unsplit into every descendant chunk. It was never counted or split, so it got duplicated, whole and uncapped, into each split piece of the deeper Article's body.

Fixed in 25ac902fc94d005bc295d06bb88c4b03bb5599bb. `path_titles` for a node's descendants now only carries the node's own heading (`texts[:1]`), never its accumulated body. Any body text a node picked up before its first child appeared is emitted as its own budgeted chunk under that heading, the same way a leaf's own overflow body already was.

Added `test_ancestor_with_children_and_own_body_is_not_duplicated_unsplit`: a Chapter/Section/Article nesting with an intro paragraph on Chapter. It fails against the pre-fix code (chunks over the budget) and passes after the fix, with every paragraph appearing exactly once and the heading present in every chunk. Also moved the size assertion in the first test to the patched token counter and switched the import to `from rag import nlp`, per your earlier nits.

I did not touch the docx-path or third-implementation points from this review; those are broader scope calls I'll leave to you.

xugangqiang:
LGTM

PR #18669, opened 2026-08-24
fix(agentic-rag): enforce bound dataset scope in retrieval tools

Lem0nTea2002:
### Summary

Follow-up to #18654.

Keep Agentic RAG retrieval within the conversation's server-bound dataset scope.

- Validate model-provided `dataset_ids` as a subset before `grep_chunks` and
  `search_chunks` reach retrieval backends.
- Return empty results without backend access when the bound scope is empty.
- Forward the bound scope to `list_chunks` deep reads.
- Add regression coverage for omitted scopes, valid subsets, out-of-scope
  rejection, validation before the 10-item cap, empty scopes, and deep-read
  forwarding.

Out-of-scope IDs return a clear error so the agent can correct its tool call.

Related review findings:

- https://github.com/infiniflow/ragflow/pull/18654#discussion_r3837569551
- https://github.com/infiniflow/ragflow/pull/18654#discussion_r3837992009

### Validation

- `CGO_ENABLED=0 go test -count=1 -cover ./internal/agentic_rag`
  - passed, 68.9% package coverage
- `CGO_ENABLED=0 go test -count=5 -shuffle=on -run 'TestResolveDatasetScope|TestRetrievalTools_|TestListChunksTool_ForwardsBoundDatasetScope' ./internal/agentic_rag`
  - passed
- `CGO_ENABLED=0 go vet ./internal/agentic_rag`
  - passed
- `gofmt -d` and `git diff --check`
  - passed

The repository `build.sh --test` command is pending CI because the local
Windows checkout does not contain the required `office_oxide` static library.

### DIFF
--- a/internal/agentic_rag/helper.go
+++ b/internal/agentic_rag/helper.go
@@ -23,6 +23,26 @@ import (
 	"ragflow/internal/agent/runtime"
 )
 
+// resolveDatasetScope keeps model-provided dataset ids within the
+// conversation's server-bound scope. An omitted request uses the full bound
+// scope; an explicit request must be a subset of it.
+func resolveDatasetScope(bound, requested []string) ([]string, error) {
+	if len(requested) == 0 {
+		return bound, nil
+	}
+
+	allowed := make(map[string]struct{}, len(bound))
+	for _, id := range bound {
+		allowed[id] = struct{}{}
+	}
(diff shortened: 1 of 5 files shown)


JinHai-CN:
The code base will be reverted.

PR #18672, opened 2026-08-24
fix(web): keep multi-select options in their declared order while searching

dcc123456:
### Summary

fix(web): keep multi-select options in their declared order while searching

cmdk's default filter scores matches fuzzily and re-sorts the list by score, so typing in the search box shuffled the options. Score every substring hit equally instead — cmdk's sort is stable, so the declared order survives, and substring matching is closer to what users expect.

### DIFF
--- a/web/src/components/ui/multi-select.tsx
+++ b/web/src/components/ui/multi-select.tsx
@@ -46,6 +46,16 @@ export type MultiSelectGroupOptionType = {
   options: MultiSelectOptionType[];
 };
 
+/**
+ * cmdk's default filter scores matches fuzzily and re-sorts the list by
+ * score, which shuffles options out of their declared order as the user
+ * types. A plain substring match scores every hit equally, so cmdk's
+ * stable sort leaves the original order intact.
+ */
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM. Verified the change against the current component:

- The `Command` wrapper in `web/src/components/ui/command.tsx` spreads `...props` through to `CommandPrimitive` (cmdk ^1.0.4), so the custom `filter` is forwarded correctly.
- `filterBySubstring` returns a flat 1/0 score, and cmdk's sort is stable, so equal-score hits retain their declared order — which is exactly the intent. The signature `(value, search)` is assignable to cmdk's `(value, search, keywords?) => number` type.
- Empty search: `includes("")` is always true, so every item scores 1 and the original order is preserved. Good.

One non-blocking note: this `filter` also applies to the "Select all", "Clear", and "Close" control items (they have no `value` prop, so cmdk falls back to their text content). That was already true with the default fuzzy filter, so it's not a regression, but worth keeping in mind if any of those controls ever need to stay visible during search.

Nice, well-scoped fix with a clear comment.

PR #18673, opened 2026-08-24
fix: the pdf generation issue in agent

6ba3i:
### Summary

added the missing dependency to the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ RUN --mount=type=cache,id=ragflow_apt,target=/var/cache/apt,sharing=locked \
     apt --no-install-recommends install -y ca-certificates \
     libglib2.0-0 libglx-mesa0 libgl1 pkg-config libgdiplus default-jdk libatk-bridge2.0-0 \
     libgtk-4-1 libnss3 xdg-utils libjemalloc-dev gnupg unzip curl wget git vim less \
-    ghostscript pandoc texlive texlive-latex-extra texlive-xetex texlive-lang-chinese \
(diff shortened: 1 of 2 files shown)


myf-bee (commented):
Look Good To Me

wangq8 (commented):
Thanks for the fix! This is a small, well-targeted change. I checked it against `agent/component/docs_generator.py`, which generates PDFs via `pandoc --pdf-engine=xelatex` with `mainfont`/`CJKmainfont` set to `Noto Sans CJK SC`.

**What's good**
- `lmodern` (Latin Modern fonts) is the correct missing dependency. With xelatex/fontspec, `\ttfamily`/verbatim (code blocks) and math fall back to Latin Modern fonts, which aren't installed by the base `texlive` set — especially since `apt --no-install-recommends` is used and `lmodern` is only a "Recommends" of the texlive collections. Its absence causes the classic `File 'lmodern.sty' not found` / "font cannot be found" failures during PDF generation.
- Applied consistently to both `Dockerfile` and `Dockerfile_base`, which are the only two files that install the texlive stack. No other Dockerfile needs it.
- No code changes, so no regression risk to the agent logic. Only a negligible image-size increase.

1. The PR body is minimal. It would help reviewers and future maintainers to include the actual error message/log (e.g. `File 'lmodern.sty' not found`, or `The font "Latin Modern Mono" cannot be found`). This confirms `lmodern` (vs `cm-super`) is the right package and makes the root cause greppable later.
2. Note that existing deployments need an image rebuild to pick this up — worth confirming the rebuild/release path is covered.

No objections to merging. 👍

wangq8:
LGTM.

PR #18676, opened 2026-08-24
fix: validate message content to prevent empty values

6ba3i:
### Summary

Message alternatives now ignore empty and whitespace-only entries before random selection. This prevents the Agent’s deferred execution from being skipped and avoids intermittent empty responses.

### DIFF
--- a/agent/component/message.py
+++ b/agent/component/message.py
@@ -43,6 +43,12 @@ from common import settings
 from api.db.joint_services.memory_message_service import queue_save_to_memory_task
 
 
+def _valid_message_content(content: Any) -> list[str]:
+    if not isinstance(content, list):
+        return []
+    return [item for item in content if isinstance(item, str) and item.strip()]
+
+
 class MessageParam(ComponentParamBase):
     """
     Define the Message component parameters.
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix! The root cause is correctly identified: `check_empty` only tests `not param`, so a list containing empty/whitespace-only strings (e.g. `[""]`) passed validation, and `random.choice` could then select an empty string, producing an empty response / skipping deferred execution. Filtering via `_valid_message_content` and validating in `check()` is a clean, minimal solution.

A few suggestions:

1. **Tests**: please add a unit test covering (a) `check()` raising when content is all-empty/whitespace-only, (b) whitespace-only entries being ignored, and (c) a mixed list of empty + valid entries still selecting a valid string.

2. **Defensive guard in `_invoke`**: `random.choice(_valid_message_content(...))` will raise `IndexError: Cannot choose from an empty sequence` if the list is empty. Today this is guarded because `check()` runs before `_invoke`, but if content can ever be mutated/configured between validation and invocation, this would surface as an opaque `IndexError` instead of a clear error. Consider raising the same clear `ValueError` (or returning early) when the filtered list is empty.

3. **Consistency (optional)**: `get_input_elements()` still does `"".join(self._param.content)`, which assumes every entry is a string and would still break on non-string entries. Not introduced by this PR, but you could reuse `_valid_message_content` there for the same filtering guarantee.

Otherwise looks good — `_valid_message_content` correctly returns the original (unstripped) items while only filtering on `item.strip()`, so no unintended content mutation.

wangq8:
LGTM.

PR #18678, opened 2026-08-24
Chore: add missing deepseek model

6ba3i:


wangq8 (commented):
Thanks for the contribution! Reviewed the model config addition. JSON is valid and the field values are consistent with the existing `deepseek-v4-flash` / `deepseek-v4-pro` entries. A few things to verify before merge:

1. **Model name / `-exp` suffix**: `deepseek-v4-flash-vision-exp` looks like an experimental model id. Please confirm this id is officially served by the DeepSeek API and is intended for production use (experimental models are sometimes renamed or removed, which becomes a maintenance burden). If a stable vision model id exists, prefer it.

2. **Context / output limits**: `max_tokens`/`content_length` of ~1M and `max_output` of 393216 are copied from the non-vision `deepseek-v4-flash`/`pro` entries. Please double-check against DeepSeek's official model card that the *vision* variant actually supports a 1M context window and 384K max output — vision models frequently have different (often smaller) context limits.

3. **Factory-level `tags`**: the DeepSeek factory still has `"tags": "LLM"` while this model is tagged `IMAGE2TEXT`. Other multi-modal factories include `IMAGE2TEXT` at the factory level too. Worth confirming the vision model is still discoverable/selectable for image-to-text use in the UI without that factory-level tag.

Non-blocking, but nice to confirm. `model_type: ["chat", "vision"]` correctly uses the canonical `vision` value (legacy `image2text` is aliased to it), so that part is fine.

Lynn-Inf (approved):
LGTM

PR #18681, opened 2026-08-24
fix(router): accept share (beta) token on agent upload and attachment download

euvre:
### Summary

Uploading an attachment on an **embedded/shared agent page** fails and the page is redirected to the login page (reported against the Go API).

**Root cause.** The embed URL (`/agent/share?shared_id=<canvas id>&auth=<beta token>`) carries no user session; the only credential is the share (beta) APIToken, which the web client sends as `Authorization: Bearer <beta>` (`web/src/utils/authorization-util.ts: getAuthorization()`). The shared page uploads through `POST /api/v1/agents/<canvas id>/upload` (`web/src/hooks/use-agent-request.ts: useUploadAgentFileWithProgress`).

Python implements this endpoint with `@login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA])` (`api/apps/restful_apis/agent_api.py:990`), so the beta token authenticates and the upload succeeds. The Go router instead registered `POST /api/v1/agents/:canvas_id/upload` inside the `authorized` group guarded by `AuthMiddleware`, which only accepts a session JWT or a regular API token. A beta token therefore produced `HTTP 401 "Invalid access token"`, and the web client's 401 interceptor (`web/src/utils/request.ts`) calls `redirectToLogin()` — the reported "jump to login page". `GET /api/v1/agents/attachments/:attachment_id/download` (used to fetch the uploaded attachment from the embed page) has the same python decorator (`agent_api.py:2666`) and the same defect.

**Fix.** Register both endpoints on the `apiBetaAuth` group (router.go), whose `BetaAuthMiddleware` resolves exactly the same three credential kinds as python's `[AUTH_JWT, AUTH_API, AUTH_BETA]` (session JWT → API token → beta token). This mirrors how the other mixed-auth endpoints (`/documents/:id/preview`, `/documents/images/:image_id`, `/thumbnails`) were already ported. `GET /api/v1/agents/attachments/:attachment_id/preview` keeps plain `@login_required` in python and therefore stays on the JWT-only authorized group. Both handlers only read the `user` context key, which `BetaAuthMiddleware` sets, so no handler changes are needed.

**Verification.**
- New regression test `TestAgentUploadAndAttachmentDownloadOnBetaAuth` (internal/router) pins: upload + attachment download are served by the beta middleware (HTTP 200 + code 102 envelope when unauthenticated, not a bare 401) and the preview route stays JWT-only. Full package passes: `bash build.sh --test ./internal/router/...`.
- Live A/B probe against the running backends with a valid beta token: before the fix, Go returned `HTTP 401 {"code":401,"message":"Invalid access token"}` while python returned `HTTP 200 {"code":103,"message":"Make sure you have permission to access the agent."}` (beta auth accepted; 103 because the probe token's tenant does not own the probe canvas). After the fix the Go API returns the same python envelope.
- Boundary: full browser end-to-end reproduction on the embed page was not possible in this environment because console login (`1@1.com`) is currently blocked by an environment data issue (the stored password hash does not match the UI's base64-wrapped password convention; the same browser login fails identically against the python backend), and minting a fresh share token requires console login. The evidence above (python parity, controlled live A/B on the auth boundary, router regression tests) pins the root cause and the fix instead.

### DIFF
--- a/internal/router/agent_routes.go
+++ b/internal/router/agent_routes.go
@@ -57,9 +57,14 @@ func RegisterAgentRoutes(g *gin.RouterGroup, h *handler.AgentHandler) {
 
 	// File operations.
 	g.GET("/download", h.DownloadAgentFile)
-	g.GET("/attachments/:attachment_id/download", h.DownloadAttachment)
 	g.GET("/attachments/:attachment_id/preview", h.PreviewAttachment)
-	g.POST("/:canvas_id/upload", h.UploadAgentFile)
+	// POST /:canvas_id/upload and GET /attachments/:attachment_id/download
+	// accept the public share (beta) token used by the embedded/shared agent
+	// page, so they are registered on the beta-auth group in router.go
+	// instead of this JWT-only group. Mirrors python
+	// @login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA]) on
+	// upload_agent_file (agent_api.py:990) and download_attachment
+	// (agent_api.py:2666).
 
 	// Component introspection + debug.
 	g.GET("/:canvas_id/components/:component_id/input-form", h.GetComponentInputForm)
--- a/internal/router/agent_routes.go
+++ b/internal/router/agent_routes.go
@@ -58,13 +58,6 @@ func RegisterAgentRoutes(g *gin.RouterGroup, h *handler.AgentHandler) {
 	// File operations.
 	g.GET("/download", h.DownloadAgentFile)
 	g.GET("/attachments/:attachment_id/preview", h.PreviewAttachment)
-	// POST /:canvas_id/upload and GET /attachments/:attachment_id/download
-	// accept the public share (beta) token used by the embedded/shared agent
-	// page, so they are registered on the beta-auth group in router.go
-	// instead of this JWT-only group. Mirrors python
-	// @login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA]) on
-	// upload_agent_file (agent_api.py:990) and download_attachment
-	// (agent_api.py:2666).
 
 	// Component introspection + debug.
 	g.GET("/:canvas_id/components/:component_id/input-form", h.GetComponentInputForm)
--- a/internal/router/agent_routes_test.go
+++ b/internal/router/agent_routes_test.go
@@ -17,12 +17,14 @@
 package router
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/gin-gonic/gin"
 
+	"ragflow/internal/common"
 	"ragflow/internal/handler"
 )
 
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
Correct Python parity. I verified the decorators match: `upload_agent_file` is `@login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA])` + `@_require_canvas_access_async`; `download_attachment` is `@login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA])`; `preview_attachment` is plain `@login_required`. Both handlers only read `user` via `GetUser`, which `BetaAuthMiddleware` sets, and `UploadAgentFile` retains the `LoadCanvasByID` IDOR mitigation. The regression test is solid.

Two notes:

1. (security-relevant, confirm-intent) `DownloadAttachment` is tenant-scoped (`user.ID`) but does not check canvas ownership — unlike upload, which calls `LoadCanvasByID`. After this change a beta (share) token for one canvas can download attachments belonging to *any* canvas in the same tenant. This is exact Python parity (python `download_attachment` also only uses `current_user.id`), so it's a deliberate choice rather than a regression — but worth confirming tenant-wide attachment download is acceptable for the share-token trust model.

2. (minor) `BetaAuthMiddleware` does not run the `local.IsAdminAvailable()` license check or the superuser rejection that `AuthMiddleware` applies, so upload/download no longer enforce them. Consistent with the other beta endpoints, just flagging it since upload is a write path.

euvre:
Thanks @wangq8 for the careful verification, and thanks for the lint/wording catches — all addressed in 4a41291d1 (router package tests pass: `bash build.sh --test ./internal/router/...` → ok). Summary per item:

**1. CodeRabbit — `noctx`: `httptest.NewRequest` → `NewRequestWithContext`** ✅ applied. Both request constructions added by this PR (`agent_routes_test.go`, the upload/download loop and the preview subtest) now use `httptest.NewRequestWithContext(context.Background(), ...)`. Pre-existing `httptest.NewRequest` calls in older tests are left untouched to keep the diff scoped.

**2. CodeRabbit — remove compatibility / route-migration / Python-parity comments** ✅ applied at all four sites (`agent_routes.go` former 61–67, `router.go` former 225–232, `agent_routes_test.go` former 95–105 and 145–147). Route registrations, test names and assertions are unchanged. The test doc comment was condensed to current-behavior rationale only (embedded page holds only a share token; unauthenticated beta middleware answers HTTP 200 + code 102 vs the JWT group's bare 401), with no replacement parity wording.

**3. @wangq8 note 1 — `DownloadAttachment` tenant-scoped without canvas ownership check** Confirmed intentional, no change. This is exact Python parity: python `download_attachment` (`api/apps/restful_apis/agent_api.py:2669`) carries only `@login_required(auth_types=[AUTH_JWT, AUTH_API, AUTH_BETA])` + `@add_tenant_id_to_kwargs` (i.e. `tenant_id = current_user.id`) and the storage lookup `STORAGE_IMPL.get(tenant_id, attachment_id)` is tenant-scoped with no canvas check — unlike `upload_agent_file`, which adds `@_require_canvas_access_async`. The Go handler mirrors that split (`UploadAgentFile` keeps `LoadCanvasByID`; `DownloadAttachment` scopes by `user.ID` only). Adding a canvas check only on the Go side would make the two implementations behave differently and would break attachment downloads whose ids are not canvas-bound. If the tenant-wide share-token trust model should be tightened, that belongs upstream in the Python endpoint first — happy to follow in a follow-up PR once Python changes.

**4. @wangq8 note 2 — `BetaAuthMiddleware` skips the `IsAdminAvailable()` license check and superuser rejection** Acknowledged, no change. Verified: `AuthMiddleware` (`internal/handler/auth.go:142-155`) applies both gates, `BetaAuthMiddleware` (ibid. 71-114) does neither — and that is the pre-existing contract of the beta group this PR moves the two routes into (searchbots/chatbots/agentbots, document preview/images/thumbnails, MCP). Python's `login_required(auth_types=[... AUTH_BETA])` path likewise applies no license/superuser gating, so the current wiring preserves parity. Flag is noted; tightening it would be a beta-group-wide change rather than something to sneak into this route move.

Verification boundary: no shared-service end-to-end run on this line — confidence is based on the router unit/integration tests above plus the line-by-line Python comparison cited in items 3–4.

wangq8:
LGTM.

euvre:
Thanks @wangq8 for the LGTM, and thanks for the follow-up test-coverage suggestion. Summary of this round (commit 92f0c181e):

**CodeRabbit — test a valid beta token, not only the missing-token path** ✅ applied. Added `TestAgentUploadAndAttachmentDownload_ValidBetaToken` (`internal/handler/agent_beta_auth_test.go`), which drives the **real** `BetaAuthMiddleware` with a valid share (beta) token — the only credential the embedded agent page holds — into the **real** `UploadAgentFile` and `DownloadAttachment` handlers, and asserts the downstream handler responses rather than the middleware envelope:

- upload → HTTP 200 + code 0 success envelope with the uploaded descriptor (canvas access check and upload path both run);
- download → HTTP 200 with the attachment bytes and `Content-Disposition: attachment`;
- a no-token control subcase on the same route keeps the contrast explicit (middleware still answers HTTP 200 + code 102), so the test proves the token actually passed the middleware rather than bypassing it.

Two notes on placement/scope, since the comment asked for the cases “in the existing router tests”: `AuthHandler`’s token-resolver stub is package-private to `internal/handler` (the established helper `stubUserTokenResolver` lives in `bot_test.go`), and a router-level valid-token run would require exporting a test-only seam on `AuthHandler` or standing up a DB-backed `UserService`. So the test follows the repo’s existing pattern for valid-token route coverage (`TestBotRoutes_NoRegularAuthRequired`): real middleware + real handlers, routes mounted on the beta group exactly as `router.go` registers them. The beta-vs-JWT route grouping itself stays pinned by the existing router-level tests in `internal/router/agent_routes_test.go` (unauthenticated shapes), which are unchanged — the two layers are complementary. The JWT/API-token resolver methods are deliberately left unstubbed so both fail and the middleware is forced through the beta branch, mirroring the embedded-page scenario.

Verification: `bash build.sh --test ./internal/handler/... ./internal/router/...` → all ok (the new test plus the full handler and router packages). No shared-service end-to-end run on this line, so confidence is based on these unit tests plus the code-level middleware/handler analysis above.

PR #18682, opened 2026-08-24
fix: publish wiki event when document is deleted

buua436:

Document deletion events were not published when the API process had no local Wiki publisher, so dataset-level Wiki updates were not triggered.

This change creates a process-local publisher in the document service and publishes `doc_deleted` events through the shared scheduler.

### Type of change

### DIFF
--- a/internal/service/document/document_crud.go
+++ b/internal/service/document/document_crud.go
@@ -342,7 +342,7 @@ func (s *DocumentService) RemoveDocumentKeepFile(ctx context.Context, docID stri
 	// the deleted document's contribution in both paths.
 	pubCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 15*time.Second)
 	defer cancel()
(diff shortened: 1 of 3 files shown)


internal/service/document/document_crud.go
```diff
@@ -342,7 +342,7 @@ func (s *DocumentService) RemoveDocumentKeepFile(ctx context.Context, docID stri
 	// the deleted document's contribution in both paths.
 	pubCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 15*time.Second)
 	defer cancel()
-	if err := knowledge_compile.PublishDeleted(pubCtx, kb.TenantID, kb.ID, docID); err != nil {
+	if err := s.publishWikiDocumentDeleted(pubCtx, kb.TenantID, kb.ID, docID); err != nil {
```
yuzhichang: The delete event shall not be wiki specific.

yuzhichang (changes requested):

yuzhichang (approved):

PR #18684, opened 2026-08-24
fix(go): sort provider model instances by creation time

euvre:
### Summary

The model providers page showed a tenant's model instances in an arbitrary
order instead of following their creation time.

**Background.** The web UI (`/user-setting/model`) renders a provider's
instances in the exact order returned by
`GET /v1/providers/<name>/instances`. On the Python API,
`provider_api_service.list_provider_instances` already sorts instances by
`create_time` desc (established by #16999, newest instance first). The Go
port of the same endpoint (`TenantModelInstanceDAO.GetAllInstancesByProviderID`)
never mirrored that: it issues a plain `Find` without any `ORDER BY`, and
`tenant_model_instance` rows are clustered on a random 32-char token primary
key, so MySQL returns them in an order unrelated to creation time. Example
from a live deployment: Tongyi-Qianwen instances created as
`350234` (07-22) → `asdasdasdasdasd` (08-07) were listed as
`["asdasdasdasdasd", "350234"]` (primary-key order).

**Fix.** Add `ORDER BY create_time DESC` to
`TenantModelInstanceDAO.GetAllInstancesByProviderID` so the Go API matches
the Python behavior (newest instance first). All callers of this DAO method
either render the list or iterate over every row, so the explicit ordering is
safe for all of them.

**Test.** New DAO unit test
`TestTenantModelInstanceDAOGetAllInstancesByProviderIDOrdersByCreateTimeDesc`
seeds three instances whose primary-key order differs from their creation
order (in-memory SQLite) and asserts the returned order is newest-first.
Verified `bash build.sh --test ./internal/dao/...` and
`bash build.sh --test ./internal/service/...` pass, and re-checked the live
endpoint after rebuild: `OpenAI-API-Compatible` now returns
`["mock-local", "111111111"]` (created 08-14 → 08-07, previously flipped).

### DIFF
--- a/internal/dao/tenant_model_instance.go
+++ b/internal/dao/tenant_model_instance.go
@@ -55,7 +55,10 @@ func (dao *TenantModelInstanceDAO) Create(ctx context.Context, db *gorm.DB, inst
 
 func (dao *TenantModelInstanceDAO) GetAllInstancesByProviderID(ctx context.Context, db *gorm.DB, providerID string) ([]*entity.TenantModelInstance, error) {
 	var instances []*entity.TenantModelInstance
-	err := db.WithContext(ctx).Where("provider_id = ?", providerID).Find(&instances).Error
+	// Newest instance first. Mirrors Python's provider_api_service
+	// list_provider_instances, which sorts by create_time desc so the list
+	// follows creation order instead of the random-token primary key order.
+	err := db.WithContext(ctx).Where("provider_id = ?", providerID).Order("create_time DESC").Find(&instances).Error
 	if err != nil {
 		return nil, err
 	}
--- a/internal/dao/tenant_model_instance_test.go
+++ b/internal/dao/tenant_model_instance_test.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"slices"
 	"testing"
 
 	"github.com/glebarez/sqlite"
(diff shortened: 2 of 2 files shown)


wangq8 (commented):
Correct and low-risk. This mirrors Python's `list_provider_instances` (`instance_objs.sort(key=lambda x: x.create_time, reverse=True)`), and `create_time` is already an indexed column so the `ORDER BY create_time DESC` is cheap. All callers either render the list or iterate every row, so the explicit ordering is safe for each of them.

The test is good — in-memory SQLite with primary-key order deliberately diverging from creation order, asserting newest-first. One very minor suggestion: the `want` length already implies the provider filter, but you could also assert the `provider-2` row is absent to pin the `WHERE provider_id = ?` clause explicitly. Non-blocking either way.

euvre:
Thanks @wangq8 for the review, and thanks @coderabbitai for the walkthrough.

**1. Test suggestion (@wangq8) — applied** (94bb68621): `TestTenantModelInstanceDAOGetAllInstancesByProviderIDOrdersByCreateTimeDesc` now also asserts the `provider-2` row (`other`) is absent via `slices.Contains`, so the `WHERE provider_id = ?` clause is pinned explicitly with a clear failure message, in addition to the existing length + newest-first order assertions. Verified with `bash build.sh --test ./internal/dao/...` — package passes (unit tests only; no shared-service end-to-end run on this line).

**2. Nitpick: remove the Python-reference comments (@coderabbitai) — not applied.** The `Mirrors Python's …` phrasing is the established doc convention across `internal/dao` (three pre-existing instances in this same file — `GetByProviderIDs`, `UpdateByID`, `DeleteByIDs` — plus ~10 sibling DAO files such as `tenant_model.go`, `tenant_model_provider.go`, `user_canvas.go`). These comments document the Go/Python dual-backend behavioral contract rather than a deprecated compatibility surface, and removing only the two new ones would leave this file's docs internally inconsistent. If maintainers would rather drop this convention, that is better done consistently across the package in a separate change; happy to do it if preferred.

No production-code changes in this update — the query and ordering behavior from c5239adf6 are unchanged.

PR #18685, opened 2026-08-24
fix: normalize model types in _factory_model_types and update calculate_model_type

6ba3i:
Fixed model type aliases so speech2text is treated as ASR and image2text as VLM/vision everywhere the provider data is sent to the UI.

### DIFF
--- a/api/apps/services/provider_api_service.py
+++ b/api/apps/services/provider_api_service.py
@@ -27,7 +27,6 @@ from api.db.services.tenant_model_provider_service import TenantModelProviderSer
 from api.db.services.tenant_model_instance_service import TenantModelInstanceService
 from api.db.services.tenant_model_service import TenantModelService
 from api.utils import model_utils
-from api.utils.model_utils import calculate_model_type, get_model_type_human
 from rag.llm import ChatModel, CvModel, EmbeddingModel, ModelMeta, OcrModel, RerankModel, Seq2txtModel, TTSModel
 
 
@@ -421,7 +420,7 @@ async def update_provider_instance(
                 # Update existing model
                 update_dict = {}
                 if isinstance(model.get("model_type"), (str, list)):
-                    target_model_type = calculate_model_type(model["model_type"])
+                    target_model_type = model_utils.calculate_model_type(model["model_type"])
                     if target_model_type != existing_model_names[model_name].model_type:
                         update_dict["model_type"] = target_model_type
                 merged_extra = json.loads(existing_model_names[model_name].extra) if existing_model_names[model_name].extra else {}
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
Clean extraction and the right fix — `_factory_model_types` was previously returning the legacy `speech2text`/`image2text` strings verbatim to the UI, and routing them through `normalize_model_types` now emits the canonical `asr`/`vision` names everywhere. `calculate_model_type` behavior is unchanged (it already applied `_LEGACY_MODEL_TYPE_ALIASES`), so no regression there.

Two small things:

1. (nit) No test added. A tiny unit test for `normalize_model_types` (legacy alias mapping + de-dup, e.g. `["speech2text", "asr"] -> ["asr"]`) would lock in the behavior that motivated this fix.

2. (nit, defensive) `normalize_model_types` assumes its argument is `str` or `list`; any other truthy value (e.g. an `int`) would raise `TypeError` on iteration. Callers guard with `if model_type else []` and the factory data is always str/list today, so this is non-blocking — but an explicit `isinstance` guard would make the contract robust.

wangq8 (commented):
Thanks for this cleanup — normalizing the legacy `speech2text`/`image2text` aliases at the API layer is the right call and is consistent with the existing `migrate_model_type_names()` data migration in `db_models.py`. The code is clean and the tests cover the important paths.

**Findings (non-blocking)**

1. **Redundant import** — `provider_api_service.py` now has both `from api.utils import model_utils` and `from api.utils.model_utils import calculate_model_type, get_model_type_human`. Both work, but it's slightly redundant. Either import `model_utils` and call `model_utils.calculate_model_type(...)`, or drop the module import and import `normalize_model_types` directly. Cosmetic only.

2. **Behavior change for non-list/non-str values** — `_factory_model_types` previously returned `[model_type]` for a truthy non-list value (e.g. an int), and now returns `[]` via `normalize_model_types`. Given the declared contract is `str | list[str]`, this is an improvement (and is explicitly tested), but worth a quick confirm that no caller in the factory data passes an int.

3. **Frontend consistency (please double-check)** — after this change the backend always emits canonical `asr`/`vision`, so the UI will stop receiving `speech2text`/`image2text` from these endpoints. Most of the frontend already handles both (`available-models.tsx` `mapModelKey`/`orderMap`), but `use-provider-fields.tsx` `modelTypeSupportsToolCall` still checks `image2text` and not `vision`. In practice chat models still match via `chat`, so this looks harmless, but worth a quick sanity check that no legacy-only string comparison breaks.

4. **Test coverage** — nice coverage of `normalize_model_types` (aliases, dedup, invalid input) and `calculate_model_type` with legacy aliases. Optional: a direct test of the `_factory_model_types` wrapper (e.g. `{"model_type": ["speech2text", "asr"]} -> ["asr"]`) would lock in the API-level normalization behavior.

Overall LGTM — happy to approve once the frontend consistency point is confirmed.

Lynn-Inf (approved):
LGTM

PR #18686, opened 2026-08-24
fix(web): agent explore streaming answer leaking into the switched-to session

euvre:
### Summary

In agent **Explore**, asking a model and switching to another session (chat box) while the answer was still streaming caused the streaming answer to be rendered inside the session the user had just switched to.

**Root cause.** The Explore page mounts `SessionChat` once per canvas; `useSendAgentMessage` therefore keeps one SSE `answerList` and one `derivedMessages` list shared by every session. The effect that folds streamed frames into the message list (`web/src/pages/agent/chat/use-send-agent-message.ts`) appended unconditionally, so after a session switch the in-flight frames of session A were written into the message list of the currently displayed session B. The global `sendLoading` also made the last assistant message of session B show a streaming indicator.

**Fix.** Every agent SSE frame already carries the `session_id` it belongs to (see the frame contract in `internal/agent/canvas/runner.go`), so ownership is known on the client:

- `useSendAgentMessage` accepts an optional `activeSessionId` (the session the page is currently displaying) and skips the streamed-answer append when the stream's `session_id` differs from it. It also exposes `streamSessionId` so callers can scope streaming UI.
- The Explore hook (`use-send-session-message.ts`) passes the URL session id as `activeSessionId`.
- `session-chat.tsx` shows the per-message streaming loading state only when the in-flight stream belongs to the displayed session.

Behavior after the fix: switching sessions mid-answer keeps each session's message list clean; switching back to the streaming session resumes rendering the answer there; the answer is still persisted to its own session by the backend. Canvas debug chat and shared-agent chat are unchanged (they pass no `activeSessionId`, so the gate is inactive).

**Verification.** Reproduced in the browser (dev services): asked for a long essay in session A, switched to session B mid-stream — pre-fix, session B showed session A's growing essay (~27k chars); post-fix, session B shows only its own history while the stream keeps running (confirmed by switching back and watching it continue in A). `jest src/pages/agent` passes (12 suites / 79 tests); `tsc --noEmit` shows no new errors in the touched files (remaining errors are pre-existing on main).

### DIFF
--- a/web/src/pages/agent/chat/use-send-agent-message.test.tsx
+++ b/web/src/pages/agent/chat/use-send-agent-message.test.tsx
@@ -1,0 +1,169 @@
+import { act, renderHook } from '@testing-library/react';
+import { MessageEventType } from '@/hooks/use-send-message';
+import { MessageType } from '@/constants/chat';
+import { useSendAgentMessage } from './use-send-agent-message';
+
+let mockAnswerList: any[] = [];
+let mockDone = true;
+const mockSend = jest.fn();
+const mockAddNewestOneAnswer = jest.fn();
+
+jest.mock('@/hooks/use-send-message', () => ({
+  // Mirrors the real MessageEventType enum; the real module cannot be
+  // requireActual'd here because it pulls in eventsource-parser's
+  // TransformStream, which jsdom does not provide.
+  MessageEventType: {
+    WorkflowStarted: 'workflow_started',
+    NodeStarted: 'node_started',
+    NodeFinished: 'node_finished',
+    Message: 'message',
+    MessageEnd: 'message_end',
+    WorkflowFinished: 'workflow_finished',
+    UserInputs: 'user_inputs',
+    WaitingForUser: 'waiting_for_user',
+    NodeLogs: 'node_logs',
+  },
+  useSendMessageBySSE: () => ({
+    send: mockSend,
+    answerList: mockAnswerList,
+    done: mockDone,
+    setDone: jest.fn(),
+    resetAnswerList: jest.fn(),
+    stopOutputMessage: jest.fn(),
+  }),
+}));
+
+jest.mock('@/hooks/logic-hooks', () => ({
+  // Fully mocked: the real module imports eventsource-parser, which needs
+  // TransformStream (unavailable in jsdom).
+  useHandleMessageInputChange: () => ({
+    handleInputChange: jest.fn(),
+    value: '',
+    setValue: jest.fn(),
+  }),
+  useSelectDerivedMessages: () => ({
+    derivedMessages: [],
+    setDerivedMessages: jest.fn(),
+    scrollRef: { current: null },
+    messageContainerRef: { current: null },
+    removeLatestMessage: jest.fn(),
+    removeMessageById: jest.fn(),
+    addNewestOneQuestion: jest.fn(),
+    addNewestOneAnswer: mockAddNewestOneAnswer,
+    removeAllMessages: jest.fn(),
+    removeAllMessagesExceptFirst: jest.fn(),
+    scrollToBottom: jest.fn(),
+    addPrologue: jest.fn(),
+  }),
+}));
+
+jest.mock('react-router', () => ({
+  useParams: () => ({}),
+  useSearchParams: () => [new URLSearchParams(), jest.fn()],
+}));
+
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Solid fix — gating the streamed-answer append on the frame's `session_id` is the right approach, and I confirmed `addNewestOneAnswer` dedupes by message id, so re-running the effect after switching back won't duplicate the answer. A few notes:

1. (minor) `NextMessageInput` still receives the global `sendLoading` (and `sendDisabled={sendLoading}`), so while a stream runs in session A, the input in the switched-to session B stays disabled. The message *rendering* is now session-scoped but the input-disable state is not — a small leftover of the original leak. Fine if intentional (single shared SSE stream), but worth a one-line comment if so.

2. (minor) `activeSessionId: sessionId ?? ''` — on a brand-new session (no URL id yet) the first frames are skipped while `activeSessionId === ''` and `streamSessionId` is the freshly-created session id, until the URL param propagates. This self-heals (the effect re-runs on `activeSessionId` change and dedupes), so the answer still renders a tick late — just flagging it so it's a conscious trade-off.

3. (nit) No test pins the new gating behavior. A small unit test asserting "frames with a `session_id` different from `activeSessionId` are not appended" would prevent regressions here.

euvre:
Thanks @wangq8 for the review, and CodeRabbit for the two findings — all addressed in 2030cfef4. Item by item:

**1. CodeRabbit (`use-send-agent-message.ts`, hydration vs stream replay) — applied.**
Confirmed the race: on an A → B → A switch the replay effect re-applies the streamed answer, but SessionChat's hydration effect then replaces `derivedMessages` wholesale with the persisted `session.message` (which cannot contain the still-streaming answer), and the replay effect only re-runs when a *new* frame arrives — so the partial answer stayed hidden, and was lost entirely if the stream finished while another session was displayed. Fix: `useSendAgentMessage` now exposes `reapplyStreamedAnswer()` (bumps a token in the replay effect's deps); SessionChat calls it whenever hydration runs while a stream owned by the displayed session is in flight, so the streamed answer is re-applied on top of the hydrated history without waiting for the next frame. `addNewestOneAnswer` merging by message id keeps this idempotent.

**2. CodeRabbit (`session-chat.tsx`, undefined `streamSessionId` fallback) — applied.**
Right: before the first frame, `streamSessionId` is `undefined`, so `isStreamingActiveSession` was `true` for every session during connection setup. The hook now records `requestedSessionId` at send time (in `sendMessage`, covering the newly created `exploreSessionId`, and in `sendFormMessage`), and SessionChat resolves the stream owner as `streamSessionId ?? requestedSessionId`. The `true` fallback now only applies when the request itself carried no session id either.

**3. @wangq8 note 1 (global `sendLoading` on `NextMessageInput`) — comment added, behavior kept.**
It is intentional: the SSE stream, `answerList` and `done` are a single shared state in `useSendMessageBySSE`, so allowing a second send while a stream runs would interleave two streams' frames into one answer list. There is now a comment at the input stating exactly that.

**4. @wangq8 note 2 (`activeSessionId: sessionId ?? ''` on a brand-new session) — kept as a conscious trade-off.**
The first frames can be skipped for the one or two renders until the `sessionId` URL param lands; the effect then re-runs on the `activeSessionId` change and `addNewestOneAnswer` dedupes/merges by id, so the answer renders a tick late with no duplication. Closing this fully would mean treating the pending request's session as the displayed session inside the gate, which re-opens a leak window when the URL lags on switching away — not worth it for a self-healing one-render delay.

**5. @wangq8 note 3 (test) — added.**
New `use-send-agent-message.test.tsx` (7 cases): frames with a `session_id` different from `activeSessionId` are not appended; matching frames are; frames without a `session_id` still are; A → B → A replay without a new frame; re-apply after external hydration; no gating when `activeSessionId` is not provided; `requestedSessionId` is recorded at send time.

**Verification:** the new suite passes (7/7); oxlint on the three touched files reports no new issues (one pre-existing `no-console` warning in untouched code); `npm run type-check` shows no new errors from this change — the repo-wide check already fails on this branch with pre-existing errors, and the count is unchanged/decreased (213 → 211) after the change. No shared-service end-to-end run on this line, so the A → B → A behavior is pinned by unit tests plus the code-path analysis above.

dcc123456 (approved):

PR #18689, opened 2026-08-24
feat(extractor): unify multi-tenant LLM cache engine with TTL jitter and quality gates

jay77721:
### 📌 Summary

This PR introduces a generic, production-grade LLM result cache engine (`internal/cache/llmcache`) and unifies caching across all ingestion extractor operators (`keywords`, `questions`, `summary`, `metadata`, `tagger`). It provides multi-tenant physical isolation, zero-allocation key generation, 24h default TTL with anti-stampede lock-free jitter, dynamic default model resolution, and strict quality gates against corrupted/error payloads.

---

#### 1. Generic Multi-Tenant Cache Engine (`internal/cache/llmcache`)
- **Type-Safe Generic API**: `Engine[T any]` supports arbitrary Go types (e.g. `string`, `map[string]any`, `map[string]int`) with JSON serialization and zero `reflect` overhead on cache reads.
- **Physical Tenant Isolation**: Hierarchical key structure `kc:llm:{tenant_id}:{task_type}:{xxhash64}` with NUL-byte field separators preventing cross-field hash collisions.
- **Direct Cache-Aside with Anti-Stampede TTL Jitter**:
  - Direct, non-blocking caching flow (`get -> compute -> validate -> set`).
  - **Default 24h TTL**: Standard 24h expiration window matching legacy Python cache behavior.
  - **Lock-Free TTL Jitter**: ±10% random TTL jitter (`21.6h ~ 26.4h`) powered by `math/rand/v2` to prevent synchronized cache expiration stampedes across batch-ingested chunks.
- **Zero-Alloc Key Generation**: Reuses `*xxhash.Digest` via `sync.Pool` (`xxhashPool`) to eliminate heap allocations in high-throughput chunk processing (`0 allocs/op`).
- **GDPR Tenant Purge**: `FlushTenant(ctx, tenantID)` performs streaming Redis `SCAN cursor COUNT 1000` with pipelined batch deletions and a reusable timer, avoiding unbounded set index growth.
- **Decoupled Observability**: Optional `WithMetrics` hook emits hit/miss/compute/skip/purge events for metric instrumentation.

#### 2. Extractor & Tagger Operators Integration
- **Unified Caching**: Integrated across all 5 extraction operators: `runAutoKeywords`, `runAutoQuestions`, `runAutoSummary`, `runEnableMetadata`, and `llmTagChunk`.
- **Dynamic Effective Model Resolution (`resolveEffectiveLLMID`)**:
  - Resolves tenant effective default chat model (`model@provider` / `tenant_model` UUID) before cache key construction.
  - When tenant admin updates the default model, cache keys automatically change to reflect the new model across both Canvas and pure input invocations.
- **Quality Gates & Clean Extractor Pipelines**:
  - Filters corrupted payloads (`**ERROR**`, `[TRUNCATED]`, unclosed `<think>` tags).
  - Consolidated single-pass text cleaning in `callTextCached`, eliminating redundant double-cleaning.
- **Optimized `awaitFutures`**: Simplified to sequential waiting, eliminating thousands of temporary goroutines and mutex locks during large-document ingestion.
- **Tagger Operator (`extractor_tag.go`)**: Migrated LLM tagger caching to `taggerCache` with effective model resolution, while strictly preserving local rule-based in-memory index `boundedTagCache`.

#### 3. Thread Safety & Test Seams
- **Atomic Global Redis Client**: Converted `redis.globalClient` to `sync/atomic.Pointer[Client]` for lock-free, race-free concurrency.
- **Clean Test Seams**: Exported `SetExtractorCachesForTest` and `SetTaggerCacheForTest` with restore closures to eliminate global Redis client swapping and prevent test race conditions.

---

### 🧪 Verification

- **Full Package Tests**:
  - `go test -race -v ./internal/cache/llmcache/...` -> **100% PASS**
  - `go test -race -v ./internal/engine/redis/...` -> **100% PASS**
  - `./build.sh --test ./internal/cache/llmcache ./internal/ingestion/component` -> **100% PASS**
- **Test Matrix Covered**:
  - Direct Cache-Aside read/write & validation.
  - TTL duration and ±10% random Jitter spread.
  - Multi-tenant physical data isolation.
  - Dynamic cache invalidation upon tenant default model switch.
  - Large-batch SCAN tenant purges and Redis outage handling.
  - Complex map/slice round-trips without reflection.

### DIFF
--- a/internal/cache/llmcache/llm_cache.go
+++ b/internal/cache/llmcache/llm_cache.go
@@ -39,8 +39,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/rand"
-	"reflect"
+	randv2 "math/rand/v2"
+	"sync"
 	"time"
 
 	"github.com/cespare/xxhash/v2"
@@ -54,15 +54,15 @@ import (
 
 // Standard sentinel errors.
 var (
-	// ErrNoTenantID is returned or used when tenantID is empty in operations requiring it.
-	ErrNoTenantID = errors.New("llmcache: tenantID is required")
-	// ErrCacheMiss indicates that the requested key was not found in the cache.
-	ErrCacheMiss = errors.New("llmcache: cache miss")
-	// ErrRejected indicates that the computed value failed the validator gate.
-	ErrRejected = errors.New("llmcache: value rejected by validator")
 	// ErrTypeAssertion indicates an internal type assertion failure.
 	ErrTypeAssertion = errors.New("llmcache: internal type assertion failed")
 )
+
+var xxhashPool = sync.Pool{
+	New: func() any {
+		return xxhash.New()
+	},
+}
 
 // KeyPrefix is the Redis namespace for all entries managed by this engine.
 const KeyPrefix = "kc:llm"
@@ -143,12 +143,15 @@ func New[T any](opts ...Option[T]) *Engine[T] {
 // (e.g. ("ab","c") vs ("a","bc")). The engine never assumes anything about the
 // business meaning of keyParts — that is the caller's responsibility.
 func BuildKey(tenantID, taskType string, keyParts ...string) string {
-	h := xxhash.New()
+	h := xxhashPool.Get().(*xxhash.Digest)
 	for _, p := range keyParts {
 		h.WriteString(p)
 		h.WriteString("\x00")
 	}
-	return fmt.Sprintf("%s:%s:%s:%x", KeyPrefix, tenantID, taskType, h.Sum64())
+	key := fmt.Sprintf("%s:%s:%s:%x", KeyPrefix, tenantID, taskType, h.Sum64())
+	h.Reset()
+	xxhashPool.Put(h)
+	return key
 }
 
 // ApplyJitter returns the base duration with ±10% random jitter applied.
@@ -163,7 +166,7 @@ func applyJitter(base time.Duration) time.Duration {
 	}
 	// ±10% jitter: range is [base * 0.90, base * 1.10]
 	delta := float64(base) * 0.10
-	factor := (rand.Float64() * 2.0) - 1.0
+	factor := (randv2.Float64() * 2.0) - 1.0
 	jitter := time.Duration(delta * factor)
 	return base + jitter
 }
@@ -196,12 +199,6 @@ func (e *Engine[T]) emit(ev Event) {
 
 func (e *Engine[T]) get(ctx context.Context, key string) (T, bool) {
 	var zero, dest T
-	val := reflect.ValueOf(&dest).Elem()
-	if val.Kind() == reflect.Map && val.IsNil() {
-		val.Set(reflect.MakeMap(val.Type()))
-	} else if val.Kind() == reflect.Slice && val.IsNil() {
-		val.Set(reflect.MakeSlice(val.Type(), 0, 0))
-	}
 
 	if e.rawClient != nil {
 		data, err := e.rawClient.Get(ctx, key).Result()
@@ -227,20 +224,26 @@ func (e *Engine[T]) set(ctx context.Context, tenantID, key string, v T) {
 	if tenantID == "" {
 		return
 	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	setCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 2*time.Second)
+	defer cancel()
+
 	ttl := applyJitter(e.ttl)
 	if e.rawClient != nil {
 		data, err := json.Marshal(v)
 		if err != nil {
 			return
 		}
-		e.rawClient.Set(ctx, key, data, ttl)
+		e.rawClient.Set(setCtx, key, data, ttl)
 		return
 	}
 	client := e.getRedis()
 	if client == nil {
 		return
 	}
-	client.SetObj(ctx, key, v, ttl)
+	client.SetObj(setCtx, key, v, ttl)
 }
 
 // GetOrCompute returns the cached value for (tenantID, taskType, keyParts), or
@@ -251,8 +254,7 @@ func (e *Engine[T]) set(ctx context.Context, tenantID, key string, v T) {
 // A double-check re-reads Redis inside the merged call in case a peer already wrote
 // the value back. The winning result is written back only after passing the Validator.
 //
-// Fail-Closed: if tenantID is empty, it bypasses caching completely, directly executing
-// computeFn without writing to Redis.
+// Fail-Open (bypass): execute computeFn directly without caching when tenantID is empty to ensure business continuity without polluting global namespace.
 func (e *Engine[T]) GetOrCompute(
 	ctx context.Context,
 	tenantID, taskType string,
@@ -261,7 +263,7 @@ func (e *Engine[T]) GetOrCompute(
 ) (T, error) {
 	var zero T
 	if tenantID == "" {
-		// Fail-closed: empty tenantID bypasses cache and strictly avoids Redis write.
+		// Fail-Open (bypass): execute computeFn directly without caching when tenantID is empty to ensure business continuity without polluting global namespace
 		return computeFn(ctx)
 	}
 
@@ -278,7 +280,9 @@ func (e *Engine[T]) GetOrCompute(
 	ch := e.group.DoChan(key, func() (any, error) {
 		// Double-check: a concurrent writer may have populated Redis while we
 		// were queued.
-		if v, ok := e.get(ctx, key); ok && (e.validate == nil || e.validate(v)) {
+		getCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 500*time.Millisecond)
+		defer cancel()
+		if v, ok := e.get(getCtx, key); ok && (e.validate == nil || e.validate(v)) {
 			return v, nil
 		}
 		v, cerr := computeFn(ctx)
@@ -298,6 +302,11 @@ func (e *Engine[T]) GetOrCompute(
 
 	select {
 	case <-ctx.Done():
+		e.group.Forget(key)
+		go func() {
+			<-ch
+			e.group.Forget(key)
+		}()
 		return zero, ctx.Err()
 	case res := <-ch:
 		if res.Err != nil {
@@ -314,26 +323,29 @@ func (e *Engine[T]) GetOrCompute(
 // FlushTenant purges every cache entry for a tenant (GDPR erasure / tenant teardown).
 // It scans for keys matching "kc:llm:{tenant_id}:*" using Redis SCAN cursor COUNT 1000,
 // deletes them in pipeline batches, sleeps 10ms between batches, and responds promptly to ctx.Done().
-// Returns the number of entries deleted.
-func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) int {
+// Returns the number of entries deleted and any error encountered.
+func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) (int, error) {
 	if tenantID == "" {
-		return 0
+		return 0, nil
 	}
 	rdb := e.getUniversalClient()
 	if rdb == nil {
-		return 0
+		return 0, nil
 	}
 
 	var cursor uint64
 	var deletedCount int
 	matchPattern := fmt.Sprintf("%s:%s:*", KeyPrefix, tenantID)
 
+	timer := time.NewTimer(10 * time.Millisecond)
+	defer timer.Stop()
+
 	for {
 		select {
 		case <-ctx.Done():
 			common.Warn("llmcache: FlushTenant cancelled by context",
 				zap.String("tenant_id", tenantID), zap.Error(ctx.Err()))
-			return deletedCount
+			return deletedCount, ctx.Err()
 		default:
 		}
 
@@ -341,7 +353,7 @@ func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) int {
 		if err != nil {
 			common.Warn("llmcache: FlushTenant SCAN failed",
 				zap.String("tenant_id", tenantID), zap.Error(err))
-			break
+			return deletedCount, err
 		}
 
 		if len(keys) > 0 {
@@ -350,14 +362,15 @@ func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) int {
 				pipe.Del(ctx, k)
 			}
 			cmders, err := pipe.Exec(ctx)
-			if err != nil && !errors.Is(err, goredis.Nil) {
-				common.Warn("llmcache: FlushTenant pipeline Del failed",
-					zap.String("tenant_id", tenantID), zap.Error(err))
-			}
 			for _, cmder := range cmders {
 				if intCmd, ok := cmder.(*goredis.IntCmd); ok {
 					deletedCount += int(intCmd.Val())
 				}
+			}
+			if err != nil && !errors.Is(err, goredis.Nil) {
+				common.Warn("llmcache: FlushTenant pipeline Del failed",
+					zap.String("tenant_id", tenantID), zap.Error(err))
+				return deletedCount, err
 			}
 		}
 
@@ -366,15 +379,23 @@ func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) int {
 			break
 		}
 
+		if !timer.Stop() {
+			select {
+			case <-timer.C:
+			default:
+			}
+		}
+		timer.Reset(10 * time.Millisecond)
+
 		select {
 		case <-ctx.Done():
 			common.Warn("llmcache: FlushTenant cancelled by context",
 				zap.String("tenant_id", tenantID), zap.Error(ctx.Err()))
-			return deletedCount
-		case <-time.After(10 * time.Millisecond):
+			return deletedCount, ctx.Err()
+		case <-timer.C:
 		}
 	}
 
 	e.emit(Event{Kind: "purge", TenantID: tenantID, Key: matchPattern})
-	return deletedCount
+	return deletedCount, nil
 }
--- a/internal/cache/llmcache/llm_cache.go
+++ b/internal/cache/llmcache/llm_cache.go
@@ -285,7 +285,8 @@ func (e *Engine[T]) GetOrCompute(
 		if v, ok := e.get(getCtx, key); ok && (e.validate == nil || e.validate(v)) {
 			return v, nil
 		}
-		v, cerr := computeFn(ctx)
+		computeCtx := context.WithoutCancel(ctx)
+		v, cerr := computeFn(computeCtx)
 		if cerr != nil {
 			return zero, cerr
 		}
--- a/internal/cache/llmcache/llm_cache.go
+++ b/internal/cache/llmcache/llm_cache.go
@@ -1,0 +1,380 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+// Package llmcache provides a production-grade, multi-tenant, generic LLM-result
+// cache engine shared across all packages (ingestion extractor, canvas flow,
+// replay harness, ...).
+//
+// Design goals:
+//   - Tenant physical isolation: every key is bound to tenant_id, so no cross-
+//     tenant cache hits are possible, and a tenant can be purged on demand.
+//   - Generics: one engine instance per value type T (e.g. []string,
+//     map[string]any); the engine is storage/serialization agnostic.
+//   - Task-aware quality gate: a caller-supplied Validator decides whether a
+//     computed value is worth caching (allows legitimate empty results,
+//     rejects ERROR / truncated payloads).
+//   - SingleFlight: concurrent identical requests (same tenant+task+hash) are
+//     collapsed into a single upstream LLM call via DoChan with context cancellation
+//     safety, with a double-check to avoid redundant recompute after a peer has
+//     already written back.
+//   - Observable: an optional MetricsHook receives hit/miss/compute/skip/purge
+//     events for Prometheus instrumentation without hard-coupling the package.
+package llmcache
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand"
+	"reflect"
+	"time"
+
+	"github.com/cespare/xxhash/v2"
+	goredis "github.com/redis/go-redis/v9"
+	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
+
+	"ragflow/internal/common"
+	"ragflow/internal/engine/redis"
+)
+
+// Standard sentinel errors.
+var (
+	// ErrNoTenantID is returned or used when tenantID is empty in operations requiring it.
+	ErrNoTenantID = errors.New("llmcache: tenantID is required")
+	// ErrCacheMiss indicates that the requested key was not found in the cache.
+	ErrCacheMiss = errors.New("llmcache: cache miss")
+	// ErrRejected indicates that the computed value failed the validator gate.
+	ErrRejected = errors.New("llmcache: value rejected by validator")
+	// ErrTypeAssertion indicates an internal type assertion failure.
+	ErrTypeAssertion = errors.New("llmcache: internal type assertion failed")
+)
+
+// KeyPrefix is the Redis namespace for all entries managed by this engine.
+const KeyPrefix = "kc:llm"
+
+// DefaultTTL mirrors the legacy Python get_llm_cache/set_llm_cache 24h window.
+const DefaultTTL = 24 * time.Hour
+
+// Validator reports whether a computed value should be cached.
+// Return true to keep it (including legitimate empty results), false to drop
+// it (e.g. an ERROR marker or a truncated payload). A nil Validator accepts
+// everything.
+type Validator[T any] func(v T) bool
+
+// Event is emitted for every cache decision so callers can instrument hit-rate,
+// compute cost, purge volume, etc.
+type Event struct {
+	Kind     string // "hit" | "miss" | "compute" | "skip_invalid" | "purge"
+	TenantID string
+	TaskType string
+	Key      string
+}
+
+// MetricsHook receives cache events. It must be cheap and non-blocking.
+type MetricsHook func(Event)
+
+// Engine is a generic, multi-tenant LLM-result cache.
+type Engine[T any] struct {
+	ttl       time.Duration
+	validate  Validator[T]
+	hook      MetricsHook
+	group     singleflight.Group
+	redisCli  *redis.Client
+	rawClient goredis.UniversalClient
+}
+
+// Option configures an Engine.
+type Option[T any] func(*Engine[T])
+
+// WithTTL overrides the cache entry TTL (default DefaultTTL).
+func WithTTL[T any](d time.Duration) Option[T] {
+	return func(e *Engine[T]) { e.ttl = d }
+}
+
+// WithValidator sets the task-aware quality gate.
+func WithValidator[T any](v Validator[T]) Option[T] {
+	return func(e *Engine[T]) { e.validate = v }
+}
+
+// WithMetrics installs an event hook for observability.
+func WithMetrics[T any](h MetricsHook) Option[T] {
+	return func(e *Engine[T]) { e.hook = h }
+}
+
+// WithRedisClient overrides the Redis client instance (primarily for testing and custom wiring).
+func WithRedisClient[T any](client *redis.Client) Option[T] {
+	return func(e *Engine[T]) { e.redisCli = client }
+}
+
+// WithUniversalClient overrides the Redis client with any go-redis UniversalClient (primarily for testing and custom wiring).
+func WithUniversalClient[T any](client goredis.UniversalClient) Option[T] {
+	return func(e *Engine[T]) { e.rawClient = client }
+}
+
+// New constructs an Engine with the given options.
+func New[T any](opts ...Option[T]) *Engine[T] {
+	e := &Engine[T]{ttl: DefaultTTL}
+	for _, o := range opts {
+		o(e)
+	}
+	return e
+}
+
+// BuildKey constructs the canonical, tenant-scoped cache key.
+//
+//	kc:llm:{tenant_id}:{task_type}:{xxhash64(keyParts...)}
+//
+// A NUL byte separates key parts to prevent cross-field hash collisions
+// (e.g. ("ab","c") vs ("a","bc")). The engine never assumes anything about the
+// business meaning of keyParts — that is the caller's responsibility.
+func BuildKey(tenantID, taskType string, keyParts ...string) string {
+	h := xxhash.New()
+	for _, p := range keyParts {
+		h.WriteString(p)
+		h.WriteString("\x00")
+	}
+	return fmt.Sprintf("%s:%s:%s:%x", KeyPrefix, tenantID, taskType, h.Sum64())
+}
+
+// ApplyJitter returns the base duration with ±10% random jitter applied.
+// If base <= 0, it returns base unmodified.
+func ApplyJitter(base time.Duration) time.Duration {
+	return applyJitter(base)
+}
+
+func applyJitter(base time.Duration) time.Duration {
+	if base <= 0 {
+		return base
+	}
+	// ±10% jitter: range is [base * 0.90, base * 1.10]
+	delta := float64(base) * 0.10
+	factor := (rand.Float64() * 2.0) - 1.0
+	jitter := time.Duration(delta * factor)
+	return base + jitter
+}
+
+func (e *Engine[T]) getRedis() *redis.Client {
+	if e.redisCli != nil {
+		return e.redisCli
+	}
+	return redis.Get()
+}
+
+func (e *Engine[T]) getUniversalClient() goredis.UniversalClient {
+	if e.rawClient != nil {
+		return e.rawClient
+	}
+	if e.redisCli != nil {
+		return e.redisCli.GetClient()
+	}
+	if rc := redis.Get(); rc != nil {
+		return rc.GetClient()
+	}
+	return nil
+}
+
+func (e *Engine[T]) emit(ev Event) {
+	if e.hook != nil {
+		e.hook(ev)
+	}
+}
+
+func (e *Engine[T]) get(ctx context.Context, key string) (T, bool) {
+	var zero, dest T
+	val := reflect.ValueOf(&dest).Elem()
+	if val.Kind() == reflect.Map && val.IsNil() {
+		val.Set(reflect.MakeMap(val.Type()))
+	} else if val.Kind() == reflect.Slice && val.IsNil() {
+		val.Set(reflect.MakeSlice(val.Type(), 0, 0))
+	}
+
+	if e.rawClient != nil {
+		data, err := e.rawClient.Get(ctx, key).Result()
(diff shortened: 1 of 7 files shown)


jay77721:
## 🔍 In-depth Code Review: Go Cache Layer

**Scope**: All 7 changed files on `feat/extractor-llmcache` (core Cache engine + tests + upstream Extractor/Tagger wiring + Redis extension)
**Dimensions**: Concurrency safety / Breakdown·Penetration·Avalanche / GC & memory / API contracts / Graceful degradation / Test coverage

---

### 📌 Summary

| Item | Verdict |
| --- | --- |
| **Verdict** | **Request Changes** (core paths are solid, but 2 design-contract deviations + several engineering flaws must be addressed) |
| **Highlights** | ① `SingleFlight.DoChan` + `select{ctx.Done / ch}` achieves **instant ctx cancellation with zero goroutine leaks**; ② `FlushTenant`'s SCAN+Pipeline dual `select` responds to ctx cancellation, making large-tenant purges safe |
| **Top risk** | ① `FlushTenant` return type regressed from `(int64, error)` to `int`, **silently swallowing SCAN/DEL errors**; ② The planned 4-state `Quality` gate collapsed into a `bool` validator, **losing the "transient failure short-cache for 1 min to prevent breakdown" semantics** |

---

### 🚨 Blocker-1: `FlushTenant` swallows errors, deviates from API contract

**File**: `internal/cache/llmcache/llm_cache.go:318-380`

**Issue**: Plan §5.1 explicitly defines `FlushTenant(ctx, tenantID) (int64, error)` — GDPR-compliant purging must let callers know whether the purge fully succeeded. The current implementation only logs a `Warn` and `break`s on SCAN/Pipeline failure, returning a deleted count with no error. Callers cannot distinguish "deleted 5 then Redis disconnected" from "fully purged".

```go
// Current code
func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) int {   // ← returns int only
    ...
    keys, nextCursor, err := rdb.Scan(ctx, cursor, matchPattern, 1000).Result()
    if err != nil {
        common.Warn("llmcache: FlushTenant SCAN failed", ...)   // ← log only
        break                                                   // ← swallows error
    }
    ...
    cmders, err := pipe.Exec(ctx)
    if err != nil && !errors.Is(err, goredis.Nil) {
        common.Warn("llmcache: FlushTenant pipeline Del failed", ...)  // ← swallowed
    }
}
```

**Suggested fix**:

```go
func (e *Engine[T]) FlushTenant(ctx context.Context, tenantID string) (int64, error) {
    if tenantID == "" { return 0, ErrNoTenantID }
    rdb := e.getUniversalClient()
    if rdb == nil { return 0, nil }

    var cursor, purged int64
    matchPattern := fmt.Sprintf("%s:%s:*", KeyPrefix, tenantID)

    for {
        select {
        case <-ctx.Done():
            return purged, ctx.Err()
        default:
        }
        keys, nextCursor, err := rdb.Scan(ctx, cursor, matchPattern, 1000).Result()
        if err != nil { return purged, fmt.Errorf("flush scan: %w", err) }
        if len(keys) > 0 {
            if _, err := rdb.Del(ctx, keys...); err != nil {
                return purged, fmt.Errorf("flush del: %w", err)
            }
            purged += int64(len(keys))
        }
        cursor = nextCursor
        if cursor == 0 { break }
        select {
        case <-ctx.Done():
            return purged, ctx.Err()
        case <-time.After(10 * time.Millisecond):
        }
    }
    e.emit(Event{Kind: "purge", TenantID: tenantID, Key: matchPattern})
    return purged, nil
}
```

---

### 🚨 Blocker-2: 4-state `Quality` gate regressed to `bool`, losing transient-failure short-cache breakdown protection

**File**: `internal/cache/llmcache/llm_cache.go:77`, `llm_cache.go:288-296`

**Issue**: Plan §IV.2 defines a 4-state `Quality` enum: `QualityReject / QualityAccept / QualityAcceptEmpty / QualityTransient`. The core semantics of `QualityTransient` is "cache transient failures (LLM timeout / network jitter) for 1 min to prevent breakdown". The current `bool` validator cannot express this, so **every request penetrates to the LLM during a provider outage**, violating the breakdown-prevention goal.

```go
// Current code
type Validator[T any] func(v T) bool   // ← true/false only

v, cerr := computeFn(ctx)
if cerr != nil { return zero, cerr }       // ← errors not cached
if e.validate == nil || e.validate(v) {
    e.set(ctx, tenantID, key, v)
} else {
    e.emit(Event{Kind: "skip_invalid"})    // ← invalid values not cached
}
return v, nil
```

**Suggested fix**: Restore the 4-state enum:

```go
type Quality int
const (
    QualityReject    Quality = iota // Corrupted, do not cache
    QualityAccept                    // Valid result, cache for TTL
    QualityAcceptEmpty               // Legitimate empty result, cache for TTL
    QualityTransient                 // Transient failure, cache short TTL to prevent breakdown
)

type Validator[T any] func(v T) Quality

// Inside GetOrCompute:
q := QualityAccept
if e.validate != nil { q = e.validate(v) }
ttl := e.ttlForTask(taskType, q)   // QualityTransient → 1 min
if q == QualityReject {
    return v, nil                  // Do not cache, but return the value
}
e.set(ctx, tenantID, key, v, ttl)
```

---

### ⚠️ Defect-3: reflect pre-allocation in `get()` is wasted work

**File**: `internal/cache/llmcache/llm_cache.go:197-224`

**Issue**: `json.Unmarshal` for `map[string]any` / `[]string` allocates internally, so the reflect-preallocated map/slice is immediately replaced — **wasted allocation + reflect call overhead** on the hot path (every cache hit).

```go
// Current code
val := reflect.ValueOf(&dest).Elem()
if val.Kind() == reflect.Map && val.IsNil() {
    val.Set(reflect.MakeMap(val.Type()))        // ← pre-allocate
} else if val.Kind() == reflect.Slice && val.IsNil() {
    val.Set(reflect.MakeSlice(val.Type(), 0, 0)) // ← pre-allocate
}
...
if err := json.Unmarshal([]byte(data), &dest); err != nil {  // ← allocates again
```

**Suggested fix**:

```go
func (e *Engine[T]) get(ctx context.Context, key string) (T, bool) {
    var dest T
    if e.rawClient != nil {
        data, err := e.rawClient.Get(ctx, key).Result()
        if err != nil { return dest, false }
        if err := json.Unmarshal(data, &dest); err != nil { return dest, false }
        return dest, true
    }
    client := e.getRedis()
    if client == nil { return dest, false }
    if !client.GetObj(ctx, key, &dest) { return dest, false }
    return dest, true
}
```

---

### ⚠️ Defect-4: `set()` silently swallows errors, no Metrics feedback

**File**: `internal/cache/llmcache/llm_cache.go:226-244`

**Issue**: `SetObj` returns `bool` indicating success, which is completely discarded. When Redis flakiness causes write-back failure, MetricsHook emits no event — an observability blind spot.

**Suggested fix**:

```go
func (e *Engine[T]) set(ctx context.Context, tenantID, key string, v T, ttl time.Duration) bool {
    if tenantID == "" { return false }
    ttl = applyJitter(ttl)
    if e.rawClient != nil {
        data, err := json.Marshal(v)
        if err != nil { return false }
        if err := e.rawClient.Set(ctx, key, data, ttl).Err(); err != nil {
            e.emit(Event{Kind: "set_failed", TenantID: tenantID, Key: key})
            return false
        }
        return true
    }
    client := e.getRedis()
    if client == nil { return false }
    ok := client.SetObj(ctx, key, v, ttl)
    if !ok { e.emit(Event{Kind: "set_failed", TenantID: tenantID, Key: key}) }
    return ok
}
```

---

### ⚠️ Defect-5: Package-level global Engine singletons + global Redis client → fragile tests

**File**: `internal/ingestion/component/extractor.go:149-154`, `extractor_tag.go:36-38`

**Issue**: Three package-level singletons are created at package load time. Tests globally swap the Redis client via `redis.SetGlobalClient(...)`. If any test calls `t.Parallel()`, they clobber each other and cause flaky tests. Not exploding today only because no cache test uses `t.Parallel()` — a latent risk.

```go
var (
    textExtractCache = llmcache.New[string](llmcache.WithValidator(isValidLLMCachePayload))
    metadataCache    = llmcache.New[map[string]any](...)
)
var taggerCache = llmcache.New[map[string]int](...)
```

**Suggestion**: Short-term, add a parallel-guard comment in `setupTestRedis`; mid-term, convert the singletons into `ExtractorComponent` fields injected via the constructor.

---

### 💡 Improvement-6/7: `Engine` lacks `Close()` and is a struct, not an interface

**File**: `internal/cache/llmcache/llm_cache.go:92`

Plan §VI defines `Close() error`, which is missing. Recommend adding `Close()` and turning `Engine[T]` into an interface with `New[T]` returning the interface value for easier mocking.

---

### ✅ Test Coverage Assessment

| Dimension | Covered | Rating |
| --- | --- | --- |
| Cache Miss / Hit | ✅ `TestExtractor_Cache_*_FirstTriggersLLM_SecondHitsRedis` | Excellent |
| Tenant isolation | ✅ `TestExtractor_Cache_MultiTenantIsolation` | Excellent |
| Empty-tenant fail-closed | ✅ `TestEngine_FailClosed_EmptyTenantID` | Excellent |
| SingleFlight dedup | ✅ `TestEngine_GetOrCompute_SingleFlightDeduplication` | Excellent |
| ctx cancel / timeout | ✅ `TestEngine_GetOrCompute_ContextTimeout/Cancellation` | Excellent |
| FlushTenant SCAN+Pipeline | ✅ Multi-tenant / ctx-cancel / 1000-key batch | Excellent |
| **Concurrent -race test** | ⚠️ No explicit `t.Parallel()` + `-race` SingleFlight test | **Missing** |
| **Runtime Redis outage fail-open** | ⚠️ Outage during runtime not tested | **Missing** |

**Suggested additions**:

```go
// SingleFlight dedup under -race
func TestEngine_GetOrCompute_Race(t *testing.T) {
    engine, _, _ := newTestEngineWithRedis[string](t)
    var wg sync.WaitGroup
    for i := 0; i < 100; i++ {
        wg.Add(1)
        go func() {
            defer wg.Done()
            _, _ = engine.GetOrCompute(context.Background(), "t", "k", []string{"same"},
                func(ctx context.Context) (string, error) { return "v", nil })
        }()
    }
    wg.Wait()
}

// Runtime Redis drop → Fail-Open degradation
func TestEngine_GetOrCompute_RedisDropped_FailOpen(t *testing.T) {
    engine, mr, _ := newTestEngineWithRedis[string](t)
    mr.Close() // simulate outage
    v, err := engine.GetOrCompute(context.Background(), "t", "k", []string{"p"},
        func(ctx context.Context) (string, error) { return "fallback", nil })
    // expect fallback, err == nil
}
```

---

### 📊 Six-Dimension Summary

| Dimension | Verdict |
| --- | --- |
| **1. Concurrency & locking** | ✅ No data races. `singleflight.Group` is concurrency-safe; Engine fields are read-only after construction. Only risk: global `redis.SetGlobalClient` clobbered in parallel tests |
| **2. Breakdown / Penetration / Avalanche** | ⚠️ Breakdown ✅ (SingleFlight DoChan); Avalanche ✅ (TTL jitter ±10%); Penetration ⚠️ (legitimate empty ✅, but lost QualityTransient short-cache to prevent transient-failure breakdown) |
| **3. GC & memory** | ⚠️ `get()` reflect pre-allocation is wasted work; planned xxhash `sync.Pool` reuse not yet landed (low priority) |
| **4. API contracts** | ⚠️ `computeFn` propagates ctx ✅; Generics ✅; Sentinel errors ✅; but `Engine` is a struct not an interface, missing `Close()`, `FlushTenant` return type deviates |
| **5. Graceful degradation** | ✅ Fail-Open path complete; ⚠️ `set()` silently swallows errors |
| **6. Test coverage** | ✅ Core paths well covered; ⚠️ missing `-race` concurrent test + runtime outage fail-open test |

---

### 🎯 Fix Priority

| Priority | Item | Blocks merge? |
| --- | --- | --- |
| **P0** | Blocker-1: `FlushTenant` returns `(int64, error)`, stop swallowing errors | **Yes** (GDPR compliance path) |
| **P0** | Blocker-2: Restore 4-state `Quality` gate, add `QualityTransient` short-cache | **Yes** (breakdown-prevention contract) |
| **P1** | Defect-3: Remove wasted reflect allocation in `get()` | Recommended |
| **P1** | Defect-4: `set()` returns bool + emits `set_failed` event | Recommended |
| **P1** | Defect-5: Eliminate global singleton test fragility | Recommended |
| **P2** | Improvement-6/7: Add `Close()` + make `Engine` an interface | Can follow up |
| **P2** | Add `-race` concurrent test + runtime outage fail-open test | Strongly recommended |

**Bottom line**: The core paths (SingleFlight dedup, ctx cancellation, tenant isolation, TTL jitter) are well-designed and above average in quality. **Resolving the 2 P0 blockers + hardening P1 tests clears this for merge.**

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+		} else {
+			e.emit(Event{Kind: "skip_invalid", TenantID: tenantID, TaskType: taskType, Key: key})
+			common.Warn("llmcache: dropping invalid computed value",
+				zap.String("task_type", taskType), zap.String("tenant_id", tenantID))
+		}
+		return v, nil
+	})
+
+	select {
+	case <-ctx.Done():
```
jay77721: **[Blocker] SingleFlight context cancellation leaks flight**

At \`llm_cache.go:299-311\` the \`select { case <-ctx.Done(): return zero, ctx.Err() }\` returns immediately but never calls \`group.Forget(key)\`. The underlying flight stays registered until the first caller finishes (up to 300ms in tests). New requests with the same key remain merged into the slow flight and cannot retry.

Suggestion: \`Forget\` on cancellation (or use \`context.WithoutCancel\` for the write-back) and avoid blocking waiters:

```go
ch := e.group.DoChan(key, func() (any, error) { ... })
select {
case <-ctx.Done():
    go func(){ <-ch; e.group.Forget(key) }()
    return zero, ctx.Err()
case res := <-ch:
}

File: internal/cache/llmcache/llm_cache.go [hunk 2]
```diff
@@ -0,0 +1,380 @@
...
+		if cursor == 0 {
+			break
+		}
+
+		select {
+		case <-ctx.Done():
+			common.Warn("llmcache: FlushTenant cancelled by context",
+				zap.String("tenant_id", tenantID), zap.Error(ctx.Err()))
+			return deletedCount
+		case <-time.After(10 * time.Millisecond):
```
jay77721: **[Blocker] `FlushTenant` timer leak and silent error handling**

`internal/cache/llmcache/llm_cache.go:374` `case <-time.After(10*time.Millisecond):` allocates a new `Timer` per batch without `Stop()`. For a large tenant (e.g. 100k keys → 100 batches) this leaks 100 timers. Also `pipe.Exec` errors beyond `redis.Nil` are only warned but `deletedCount` is still returned as success, masking partial purge failures.

Fix:
```go
timer := time.NewTimer(10 * time.Millisecond)
defer timer.Stop()
for {
    // ... scan + pipeline ...
    if cursor == 0 { break }
    timer.Reset(10 * time.Millisecond)
    select {
    case <-ctx.Done(): return deletedCount
    case <-timer.C:
    }
}
```
And return/propagate `Exec` errors instead of `Warn`-only.

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+
+func (e *Engine[T]) emit(ev Event) {
+	if e.hook != nil {
+		e.hook(ev)
+	}
+}
+
+func (e *Engine[T]) get(ctx context.Context, key string) (T, bool) {
+	var zero, dest T
+	val := reflect.ValueOf(&dest).Elem()
```
jay77721: **[High] GC pressure: `reflect` on every `get()` + global `math/rand` lock**

* `internal/cache/llmcache/llm_cache.go:199` `reflect.ValueOf(&dest).Elem()` allocates per cache hit to init nil map/slice. `json.Unmarshal` already allocates a nil map, so the branch is unnecessary. Prefer removing it or using a type switch via `any`:
```go
var zero T
if e.rawClient != nil {
    data, err := e.rawClient.Get(ctx, key).Result()
    if err != nil { return zero, false }
    // let Unmarshal handle nil map allocation
    if err := json.Unmarshal([]byte(data), &dest); err != nil { return zero, false }
}
```
* `llm_cache.go:166` `rand.Float64()` uses the global locked source. Under high QPS it becomes a contention point. Switch to `math/rand/v2` (already used in `extractor_tag.go:12`) or a per-Engine `rand.Rand` with `sync.Mutex`.

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+	return applyJitter(base)
+}
+
+func applyJitter(base time.Duration) time.Duration {
+	if base <= 0 {
+		return base
+	}
+	// ±10% jitter: range is [base * 0.90, base * 1.10]
+	delta := float64(base) * 0.10
+	factor := (rand.Float64() * 2.0) - 1.0
```
jay77721: **[High] `applyJitter` uses legacy `math/rand` with global lock**

`internal/cache/llmcache/llm_cache.go:166` `rand.Float64()` shares the deprecated global source. `internal/ingestion/component/extractor_tag.go:12` already imports `math/rand/v2` which is goroutine-safe without a global lock. Unifying on `rand/v2` removes contention and aligns with modern idioms.

```go
import randv2 "math/rand/v2"
factor := randv2.Float64()*2 - 1
```

internal/ingestion/component/extractor.go
```diff
@@ -126,6 +125,34 @@ func SetExtractorConcurrency(n int) {
...
+		return false
+	}
+	if strings.HasPrefix(trimmed, "<think>") && !strings.Contains(trimmed, "</think>") {
+		return false
+	}
+	return true
+}
+
+// Global generic cache engine singletons for Extractor operator tasks.
+var (
```
jay77721, line 149: **[High] Global cache singletons + global Redis swap race**

`internal/ingestion/component/extractor.go:149-153` and `extractor_tag.go:36`
```go
var textExtractCache = llmcache.New[string](llmcache.WithValidator(isValidLLMCachePayload))
var taggerCache = llmcache.New[map[string]int](...)
```
Both engines lazily resolve `redis.Get()` on every `get()/set()`. Tests in `extractor_test.go:27 setupTestRedis` swap the global via `redis.SetGlobalClient` + `Cleanup` restore. Concurrent `go test -parallel` (or parallel package runs) races on the single global, flaking `TestExtractor_Cache_*` vs `llmcache` tests.

Fix: inject via `WithUniversalClient` for tests (as `llm_cache_test.go:43` already does) by exposing a test seam:

```go
// extractor.go
func SetExtractorCachesForTest(text *llmcache.Engine[string], meta *llmcache.Engine[map[string]any]) func() { ... }

// extractor_test.go
restore = SetExtractorCachesForTest(llmcache.New[string](llmcache.WithUniversalClient[string](rdb), ...))
```
Or make `textExtractCache` use `sync.Once` + `ResetForTest` instead of raw global.

internal/ingestion/component/extractor.go
```diff
@@ -126,6 +125,34 @@ func SetExtractorConcurrency(n int) {
 	}
 }
 
+// isValidLLMCachePayload validates whether an LLM extraction output is suitable for caching.
+// It rejects empty strings, responses containing error markers ("**ERROR**"),
+// and obvious truncated outputs (e.g. unclosed reasoning tags or truncation markers).
+func isValidLLMCachePayload(s string) bool {
```
jay77721, line 131: **[High] Dead validator branch + double cleaning (cache penetration)**

`internal/ingestion/component/extractor.go:131-145` `isValidLLMCachePayload` rejects `**ERROR**`, but `callTextCached:1142` does `cleanExtractionResult(raw)` *before* validation, where `cleanExtractionResult:1040` already maps `**ERROR**` → `""`. The `Contains("**ERROR**")` branch is dead for text caches.

More importantly, `runAutoKeywords:702`, `runAutoQuestions:739`, `runAutoSummary:782` do a second `cleanExtractionResult` *after* `GetOrCompute` – the same cleaning already happened inside `callTextCached`. This is redundant and hides the fact that empty strings are rejected (`isValid==false`) and thus never cached. A chunk that always yields empty/`**ERROR**` will hammer the LLM on every build → cache penetration.

Fix: keep single cleaning inside `callTextCached` and remove outer calls, or change validator to accept cleaned empty sentinel with short TTL (e.g. 5m) to break the hammer:
```go
// validator: allow empty sentinel but caller sets short TTL via WithTTL JIT?
// or: if s=="" after clean, set empty placeholder with 5m TTL
```

internal/ingestion/component/extractor.go
```diff
@@ -654,12 +694,8 @@ func (c *ExtractorComponent) runAutoKeywords(ctx context.Context, db *gorm.DB, i
...
 	}
 	systemPrompt = renderExtractorPrompt(systemPrompt, topN)
-	kwTemp := extractorTemperature
-	kwIn := extractorInputs{
-		llmID:       in.llmID,
-		temperature: &kwTemp,
-	}
-	resultStr, err := c.callText(ctx, db, kwIn, systemPrompt, chunkText)
+
+	resultStr, err := c.callTextCached(ctx, db, in, "keywords", systemPrompt, chunkText)
```
jay77721, line 733: **[Medium] Double `cleanExtractionResult` is redundant**

`callTextCached:1127` already returns `cleanExtractionResult(raw)` inside `GetOrCompute`, but `runAutoKeywords:698`, `runAutoQuestions:735`, `runAutoSummary:778` call `cleanExtractionResult` again after the cache lookup. Remove the outer calls to keep a single source of truth and save allocations.

```go
resultStr, err := c.callTextCached(ctx, db, in, "keywords", systemPrompt, chunkText)
// resultStr = cleanExtractionResult(resultStr) // ← delete
if resultStr == "" { return nil }
```

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/cespare/xxhash/v2"
+	goredis "github.com/redis/go-redis/v9"
+	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
+
+	"ragflow/internal/common"
+	"ragflow/internal/engine/redis"
+)
+
+// Standard sentinel errors.
```
jay77721: **[Medium] Unused sentinel errors + misleading Fail-Closed comment**

`internal/cache/llmcache/llm_cache.go:55-64` defines `ErrNoTenantID`, `ErrCacheMiss`, `ErrRejected` but only `ErrTypeAssertion` is ever returned. `GetOrCompute:263` on empty `tenantID` silently bypasses cache and calls `computeFn`, which is Fail-Open, contradicting the `Fail-Closed` doc comment. Either return `ErrNoTenantID` or rename the comment to `Fail-Open (bypass)`.

```go
if tenantID == "" {
    return zero, ErrNoTenantID // or document as bypass
}
```

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+		return zero, false
+	}
+	return dest, true
+}
+
+func (e *Engine[T]) set(ctx context.Context, tenantID, key string, v T) {
+	if tenantID == "" {
+		return
+	}
+	ttl := applyJitter(e.ttl)
```
jay77721: **[Medium] Redis write uses cancelled `ctx` → silent miss**

`internal/cache/llmcache/llm_cache.go:230-236` `set()` is called from inside `group.DoChan` with the caller `ctx`. If the caller times out (30ms test) or cancels, `rawClient.Set(ctx, ...)` immediately fails with `context.Canceled`, so the just-computed value is never cached. Use `context.WithoutCancel(ctx)` with a short timeout for the write-back:

```go
setCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 2*time.Second)
defer cancel()
e.rawClient.Set(setCtx, key, data, ttl)
```
Same for the `get` double-check path (200ms Redis budget).

internal/ingestion/component/extractor_tag.go
```diff
@@ -24,15 +24,19 @@ import (
...
 	"ragflow/internal/common"
 	"ragflow/internal/dao"
-	"ragflow/internal/engine/redis"
 	"ragflow/internal/entity"
 	"ragflow/internal/ingestion/component/globals"
 	"ragflow/internal/ingestion/component/schema"
 	"ragflow/internal/tokenizer"
 )
 
+var taggerCache = llmcache.New[map[string]int](llmcache.WithValidator(func(m map[string]int) bool {
```
jay77721: **[Medium] `taggerCache` lacks jitter/penetration guard for `nil` vs empty**

`internal/ingestion/component/extractor_tag.go:36` `Validator: m != nil` accepts `make(map[string]int,0)` (empty `{}`) which `llmTagChunk:1209` checks via `len(result)>0` before writing to `chunk`. An empty `{}` is still cached (good), but a `nil` parse failure (invalid JSON/`**ERROR**`) is rejected and thus hammered every time. Consider caching `nil`→empty sentinel with 5m TTL as penetration defense, or document that tagger LLM must be retried.

```go
taggerCache = llmcache.New[map[string]int](
    llmcache.WithValidator(func(m map[string]int) bool { return true }), // cache empty too
    llmcache.WithTTL[map[string]int](24*time.Hour), // + jitter already
)
```
and in `llmTagChunk` handle `result==nil` as `empty` cached sentinel.

internal/cache/llmcache/llm_cache.go
```diff
@@ -0,0 +1,380 @@
...
+
+// BuildKey constructs the canonical, tenant-scoped cache key.
+//
+//	kc:llm:{tenant_id}:{task_type}:{xxhash64(keyParts...)}
+//
+// A NUL byte separates key parts to prevent cross-field hash collisions
+// (e.g. ("ab","c") vs ("a","bc")). The engine never assumes anything about the
+// business meaning of keyParts — that is the caller's responsibility.
+func BuildKey(tenantID, taskType string, keyParts ...string) string {
+	h := xxhash.New()
```
jay77721: **[Suggestion] `BuildKey` allocates `xxhash.Digest` per call**

`internal/cache/llmcache/llm_cache.go:146` `xxhash.New()` per `BuildKey` is hot (called for every chunk × task). Use stack-allocated `xxhash.Sum64String` or a `sync.Pool` of digest to reduce GC. Chunk text can be large, consider pre-hashing `chunkText` with `xxhash.Sum64([]byte(text))` instead of `WriteString`.

Performance is correct now (±10% jitter, NUL separator防碰撞做得很好), this is just allocation micro-opt for high-throughput ingestion.

jay77721:
## Review: `feat/extractor-llmcache` — Comprehensive Cache Review (6 Dimensions)

**Overall:** `Request Changes` — design is solid (generic `Engine[T]` + `BuildKey` NUL-separation + SingleFlight + ±10% jitter + tenant isolation via `kc:llm:{tenant_id}:*`), but 3 blockers prevent merge. Inline comments with fix snippets have been posted; this summary maps them to the six required dimensions.

### 1. Concurrency & Lock Contention — *Good with fixes*
- ✅ `singleflight.Group` correctly collapses hot-key thundering herd; `Engine` has no coarse `RWMutex`. Good.
- 🔴 **Blocker** `llm_cache.go:300` cancellation leaks flight (`DoChan` + `select` without `Forget`) — new requests stay merged into the slow flight. See inline fix (async `Forget` after `<-ch`).
- 🟡 `llm_cache.go:166` `math/rand` global lock contention — switch to `math/rand/v2` (already used in `extractor_tag.go:12`) or per-Engine pool.
- 🟡 `llm_cache.go:199` `reflect` per `get()` allocates; `json.Unmarshal` already handles nil map init — remove reflection branch.

### 2. Breakdown / Penetration / Avalanche — *Partial*
- ✅ **Breakdown**: SingleFlight + double-check re-read `llm_cache.go:278-281` correct.
- 🔴 **Penetration**: text validator `isValidLLMCachePayload` (extractor.go:131) rejects `""`/`**ERROR**`, but `callTextCached:1142` already cleaned `**ERROR**`→`""` so branch is dead, and empty is never cached → dirty chunk hammers LLM every build. Tagger `m!=nil` also hammers on parse-fail `nil`. Recommend sentinel empty with short TTL (5m) or document as intentional hammer.
- ✅ **Avalanche**: `applyJitter` ±10% (21.6-26.4h) correct, but verify via `TestEngine_JitterRange` distribution (already does 1000 iterations).

### 3. GC / Memory / Goroutine Lifecycle — *Needs tuning*
- 🔴 `FlushTenant:374` `time.After` leak → `Timer` + `Reset`. Unbounded pipeline errors only `Warn`ed.
- 🟡 `BuildKey:146` `xxhash.New()` per call alloc; `callTextCached` includes full `chunkText` + `systemPrompt` in hash — correct for collision avoidance but large strings cause alloc. Consider pooling digest.
- ✅ No background evictor goroutine — TTL via Redis `EXPIRE`, so no leak. Pool workers via `extractorPool` correctly not `StopWait`ed.

### 4. API Design & Idiomatic Go — *Mostly idiomatic*
- ✅ `ctx` is first arg on `GetOrCompute`, `BuildKey`, `FlushTenant`. Generic `Engine[T]` avoids `any` casts (except singleflight `any` → `ErrTypeAssertion` ok).
- 🔴 Unused sentinels `ErrNoTenantID/ErrCacheMiss/ErrRejected` (`llm_cache.go:55`) + `Fail-Closed` comment contradicts actual Fail-Open bypass `if tenantID==""{return computeFn(ctx)}`. Fix comment or return `ErrNoTenantID`.
- 🟡 `set()` uses cancelled `ctx` for Redis write `llm_cache.go:230` → silent miss on timeout. Use `context.WithoutCancel` + 2s budget.

### 5. Resilience / Degradation — *Correct fail-open*
- ✅ `get()` nil-client → miss → `computeFn` fallback; `set()` nil-client → no-op. Good graceful degrade.
- 🟡 Redis has no per-call timeout beyond caller `600s extractorTimeout`; add `200ms` Redis budget via `WithTimeout(WithoutCancel(ctx))` to avoid pipeline stall.

### 6. Tests & Benchmarks — *Strong coverage, minor gaps*
- ✅ `miniredis` covers hit/miss, multi-tenant isolation (`FlushTenant:305`), error rejection, singleflight (10 goros), context timeout/cancellation, jitter range, empty-legitimate vs empty-rejected.
- 🟡 No `go test -race` for global singleton race (`extractor.go:149` + `extractor_test.go:27` global swap). See inline seam `SetExtractorCachesForTest`.
- 🟡 No `BenchmarkBuildKey/GetOrCompute` for reflect+hash cost.
- ✅ `TestExtractor_Cache_*` (3 tasks) correctly asserts 3 calls → 0 on hit and cross-tenant 6-total isolation.

**Merge-ready patch order:** 1) `Forget`+`WithoutCancel`, 2) `Timer.Reset`+error propagate, 3) singleton test seam, 4) dedup cleaning vs validator, 5) remove reflect/double-clean. Great work on unifying 5 extractors under one engine — these fixes make it production-ready.

jay77721:
## 📌 Follow-up Review: Fix Commit `e076cbde1`

**Scope**: 5 files, +522 / −66 lines

| Item | Verdict |
| --- | --- |
| **Verdict** | **LGTM with minor comments** — 4 of 5 P1/P2 issues properly resolved; 1 subtle concurrency race in the new `Forget` pattern; Blocker-2 Quality gate is now a conscious, defensible design trade-off |
| **Biggest improvement** | Test seams properly isolated — per-test engine injection via `SetExtractorCachesForTest` / `SetTaggerCacheForTest` eliminates the global Redis race entirely |
| **Remaining nit** | Double-`Forget` in the `GetOrCompute` ctx-cancellation path has a narrow race that can temporarily defeat dedup |

---

### ✅ Issues Properly Resolved

| Original finding | Status | Notes |
| --- | --- | --- |
| **Blocker-1**: `FlushTenant` swallows errors | ✅ Fixed | Now returns `(int, error)`; SCAN / Pipeline errors propagated. Uses reusable `time.Timer` instead of `time.After` (also fixes the per-iteration timer leak). |
| **Defect-3**: reflect pre-allocation in `get()` | ✅ Fixed | Reflect block removed; `json.Unmarshal` handles allocation. |
| **Defect-5**: global singleton test fragility | ✅ Fixed | `extractorCachesMu` / `taggerCacheMu` guard access; tests inject per-test engines instead of swapping global Redis. |
| **Improvement-8**: global `rand` contention | ✅ Fixed | Switched to `math/rand/v2` (lock-free). |
| **GC: xxhash allocation** | ✅ Fixed | `xxhashPool` (`sync.Pool`) reuses `*xxhash.Digest`. |
| **Write-back on cancelled ctx** | ✅ Fixed | `set()` and double-check now use `context.WithoutCancel(ctx)` with a bounded timeout (2s / 500ms), so a cancelled caller no longer aborts in-flight write-back. |
| **Redundant double-clean** | ✅ Fixed | Removed the extra `cleanExtractionResult` calls in `runAutoKeywords/Questions/Summary` (already cleaned inside `callTextCached`). |
| **Empty LLM ID in cache key** | ✅ Fixed | New `resolveEffectiveLLMID` resolves the tenant default model, so cache keys stay correct when no explicit `llm_id` is provided. |

---

### ⚠️ Medium: Double-`Forget` race in `GetOrCompute` (new concurrency bug)

**File**: `internal/cache/llmcache/llm_cache.go:302-309`

**Current code**:
```go
case <-ctx.Done():
    e.group.Forget(key)          // (1) remove key so new callers can retry
    go func() {
        <-ch
        e.group.Forget(key)      // (2) remove again after original flight completes
    }()
    return zero, ctx.Err()
```

**Problem**: There is a narrow but real race:
1. Caller A cancels → `Forget(key)` removes the entry.
2. Caller B arrives with the same key → starts a **new** flight, registers `key` in the map.
3. Caller A's original flight completes → the goroutine calls `Forget(key)` again → **incorrectly deletes Caller B's entry**.
4. Caller C arriving now starts yet another duplicate flight, **temporarily defeating SingleFlight dedup**.

The window is small, and the consequence is only a transient loss of deduplication (not incorrect results), but under thundering-herd retry patterns it can amplify downstream calls.

**Recommended fix**: Forget **once**, after the channel drains. To still unblock retries promptly, the `computeFn` should respect ctx cancellation so the slow flight ends quickly on its own:

```go
case <-ctx.Done():
    // Let retries start a fresh flight immediately; drain the original
    // flight asynchronously and Forget once it completes so we don't
    // accidentally evict a newer flight registered by a retry.
    e.group.Forget(key)
    go func() {
        <-ch
    }()
    return zero, ctx.Err()
```

Alternatively, if keeping the second `Forget` is desired for cleanup, guard it by checking whether the current call is still the one we cancelled (requires tracking the call pointer), but the single-`Forget` pattern above is simpler and correct.

---

### ℹ️ Design Note: Blocker-2 (4-state Quality gate) — accepted as a conscious trade-off

The `Validator[T any] func(v T) bool` remains a 2-state bool. The planned `QualityTransient` short-cache semantics (cache transient failures for 1 min to prevent breakdown) are **not** implemented. This is a defensible simplification — transient errors simply aren't cached, so each retry re-attempts the call. The `context.WithoutCancel` write-back ensures that a cancelled caller's successful result still gets cached for future callers. As a follow-up, the 4-state enum can be introduced without breaking the current API.

---

### 📊 Test Coverage Improvements (new tests in this commit)

| New test | What it covers |
| --- | --- |
| `TestEngine_GetOrCompute_ContextCancellation_AllowsImmediateRetry` | Verifies a retry is not blocked by a slow first flight |
| `TestEngine_FlushTenant_ScanError` | Verifies Redis outage returns error (Blocker-1 fix) |
| `TestEngine_GetOrCompute_ComplexTypes_NoReflect` | Verifies map round-trip without reflect (Defect-3 fix) |
| `TestEngine_GetOrCompute_WriteBack_ContextCancelled` | Verifies write-back survives caller cancellation |
| `TestSetExtractorCachesForTest_Restore` | Verifies test-seam restore correctness |
| `TestExtractor_Cache_EffectiveLLMIDResolution` | Verifies cache key changes when tenant default model changes |
| `TestExtractor_DoubleCleanElimination` | Verifies single-clean path correctness |
| `BenchmarkBuildKey` / `BenchmarkEngine_ApplyJitter` | Allocation + parallel-jitter benchmarks |

**Test verdict**: Strong. The new tests directly target the fixed defects. The only remaining gap is an explicit `go test -race ./internal/cache/llmcache/... ./internal/ingestion/component/...` in CI, which I'd recommend adding.

---

### 🎯 Verdict

**Approve after addressing the medium `Forget` race** (or accepting it as a known narrow window with a code comment). All other P0/P1 items are cleanly resolved, test isolation is production-quality, and the new tests are precise. This is a well-executed review-response commit.

PR #18690, opened 2026-08-24
Fix: porting #18300 to go

wangq8:
Follow with #18300

# Test
## Retrieval Testing
image

## Search
image

## Chat
image

## Agent
image

### DIFF
--- a/internal/agent/runtime/retrieval.go
+++ b/internal/agent/runtime/retrieval.go
@@ -1,0 +1,287 @@
+//
+//  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+// Retrieval contracts shared by the canvas agent runtime (internal/agent/tool)
+// and the smart-reasoning agent (internal/agentic_rag). Keeping these here —
+// in the engine-agnostic runtime package — means neither agent layer depends on
+// the other: both depend on this shared contract.
+package runtime
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	"gorm.io/gorm"
+)
+
+// RetrievalChunk is the minimal shape a RetrievalService returns. The full
+// Chunk type (with document_id, docnm_kwd, position, etc.) lives in
+// internal/entity and is wired in by the retrieval adapters.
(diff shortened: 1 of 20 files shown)


wangq8 (commented):
Thanks for porting #18300 to Go — the service-registry refactor and the updated `TestRetrievalUsesPrefetchSizeAsCandidateSet` are clean. A few issues to flag:

## 1. Bug (high): existing dialogs break retrieval because `prefetch_size` has no default/backfill

`entity.Chat.PrefetchSize` has no GORM `default:` tag and there is no migration backfilling `dialog.prefetch_size` (startup `AutoMigrate` only *adds* the column, leaving existing rows `0`/NULL). `chat_pipeline.go` then does:

```go
prefetchSize := int(chat.PrefetchSize)   // 0 for pre-existing dialogs
...
PrefetchSize: &prefetchSize,             // non-nil pointer to 0
```

In `nlp/retrieval.go` only `nil` falls back to `64`; an explicit `0` hits:

```go
if prefetchSize <= 0 || req.Page > prefetchSize/pageSize {
    return nil, fmt.Errorf("prefetch_size(%d) must be greater than ...")
}
```

So any chat created before this change will fail retrieval with `prefetch_size(0) ...` until the dialog is re-saved. Please add a backfill migration (default 64), or fall back to the default when `*req.PrefetchSize <= 0` in `nlp/retrieval.go` (or guard in `chat_pipeline.go`).

## 2. Bug: `dataset/search.go` clobbers a caller-supplied `prefetch_size`

Inside the `searchConfig != nil` branch the code does `prefetchSize = 100` unconditionally, then only overrides if the config actually contains `prefetch_size`. So when a request carries its own `prefetch_size` together with a `search_id` whose config omits the key, the caller's value is silently discarded. `chunk.go` handles this correctly with `if req.PrefetchSize == nil || *req.PrefetchSize == 0`. Please align the two paths (request value should win unless config explicitly sets it).

## 3. Behavior change: `total` is now capped at the prefetch window

`total := int64(len(validIdx))` and the old `countThresholdValidMatches` fallback are gone, so `total` no longer reflects matches beyond `prefetch_size`. Since the frontend uses `total` for "N results" / page count, large result sets will now report a truncated count (e.g. max 64/100). If this is intended parity with #18300, it's worth calling out explicitly in the PR/docs.

## 4. Behavior change: deep pagination + rerank pagination now error out

`if req.RerankModel != nil && req.Page != 1` returns an error, and the `req.Page > prefetchSize/pageSize` guard rejects deep pages (previously the code fetched deeper ES windows via `searchPage`). Please confirm this matches #18300 semantics and document the new pagination limits for API consumers.

## 5. Consistency: `prefetch_size` defaults differ across paths

- chat: 64
- ask: 100
- mindmap / searchbot: 100
- chunk.go / dataset/search.go: 64 (or 100 when a search config is present)

Pick one canonical default (or document the divergence) to avoid surprising differences between the chat and test/search endpoints.

## Nit

- `common.Info("Retrieval prefetch params", ...)` will log on every retrieval; the previous equivalent used `common.Debug`. Consider `Debug` to avoid log noise.
- The error string "must be greater than page * page_size(%s)" is slightly ambiguous when `requiredSize == "overflow"`; a clearer message (e.g. "prefetch_size(%d) must be >= page(%d) * page_size(%d)") would help debugging.

Issue #18683, opened 2026-08-24
[Bug]: Wiki "Generate" on Artifacts page silently no-ops for Built-in datasets with misleading "No enabled documents" error


lb405828825:
### Self Checks

- [x] I have searched for existing issues, including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] Non-english title submissions will be closed directly.
- [x] I have not modified the template and have filled in all the required fields.

---

### RAGFlow workspace code commit ID

v0.27.0 release image (also reproduced on nightly-20260803)

### RAGFlow image version

`infiniflow/ragflow:v0.27.0`

### Other environment information

- MySQL 8.0.39, Elasticsearch 8.11.3
- Fresh install (empty DB), OpenAI-API-Compatible models

---

### Actual behavior

On the **Artifacts** page of a dataset (`/dataset/compilation/{kb_id}`), with the template dropdown set to **Wiki**, clicking **Generate** completes in ~0.1s and produces **no wiki pages**. The backend logs:

```
set_progress(...), progress: 0.0, progress_msg: "Loading documents for wiki compilation..."
set_progress(...), progress: 1.0, progress_msg: "No enabled documents are configured for wiki compilation."
handle_task done for task {"task_type": "wiki", ...}
```

The documents in the dataset are fully parsed and enabled:

```
run=3 (done), status=1 (enabled), progress=1, chunk_num=306/17/7
```

The message "No enabled documents are configured" is misleading — the documents ARE enabled. The real reason is that none of them resolves to a wiki compilation template.

### Root cause (source-verified)

`rag/svr/task_executor_refactor/dataset_wiki_generator.py:287` — `_wiki_eligible_docs()` requires each document to resolve to at least one `kind == "wiki"` compilation template via:

1. `parser_config` → `compilation_template_group_ids`, or
2. `document.pipeline_id` → the pipeline's **Compiler** component params → `compilation_template_group_ids` (`_extract_pipeline_compiler_group_ids`, line 174)

Documents uploaded under the **Built-in** parse type never carry either, so `eligible` is always empty and `run_wiki()` short-circuits at line 1680:

```python
if not eligible and not is_incremental:
    progress(1.0, "No enabled documents are configured for wiki compilation.")
    return
```

Two additional traps on the pipeline path:

- A pipeline without a **Compiler** component fails identically and silently (verified: a `File→Parser` pipeline bound to the document produces the same "No enabled documents" result).
[...]

PR #18691, opened 2026-08-24
fix(svr): distinguish empty-docs from missing-template in wiki empty-eligibility message

Harsh23Kashyap:
## Summary

`run_wiki` and `run_wiki_incremental` in `rag/svr/task_executor_refactor/dataset_wiki_generator.py` both surfaced the same misleading "No enabled documents are configured for wiki compilation" message when `_wiki_eligible_docs` returned an empty list. The two failure modes have different user actions:

- No enabled documents → user needs to upload / enable documents.
- Enabled documents present, but none of them resolves to a wiki compilation template → user needs to attach a Wiki template to the dataset or to each document's `parser_config` (or pipeline's Compiler component).

Refs #18683. The reporter has fully parsed enabled documents and gets told "no enabled documents", which sends them on a wild-goose chase looking for the document that's supposed to be missing.

## Fix

Extract the message logic into a new module-level helper `_wiki_empty_eligible_message(all_docs)` and route both call sites through it. When the eligibility check returns an empty list, the user now sees the correct message for their situation.

- `run_wiki` (line 1322-1324) was previously emitting `"No documents are configured for wiki compilation."` (slightly different wording, also misleading).
- `run_wiki_incremental` (line 1659-1662) was previously emitting `"No enabled documents are configured for wiki compilation."` (the misleading version).

Both now route through the helper, which returns:
- `"No enabled documents are configured for wiki compilation."` when `all_docs` is empty or all docs are disabled.
- `"{N} enabled document(s) found, but none of them has a Wiki compilation template attached. Set a Wiki template on the dataset or on each document's parser_config to enable Wiki generation."` when enabled docs exist.

## Test

Five regression tests in `test/unit_test/rag/svr/task_executor_refactor/test_wiki_empty_eligible_message.py`:

- `test_wiki_empty_eligible_message_no_documents` — empty dataset returns the original "no enabled documents" hint.
- `test_wiki_empty_eligible_message_all_disabled` — docs with `status != '1'` (e.g. '0' disabled, '2' archival) are not counted as enabled; the "no enabled documents" hint still fires.
- `test_wiki_empty_eligible_message_enabled_but_no_template` — enabled docs without a wiki template get the new actionable message naming the count of enabled docs and the `parser_config` fix path.
- `test_wiki_empty_eligible_message_handles_missing_status` — a doc with no `status` field defaults to enabled (`status='1'`), matching `_wiki_eligible_docs`'s convention.
- `test_wiki_empty_eligible_message_handles_none` — `all_docs` may be `None` (e.g. the upstream service returns `None` on error); the helper still returns the no-enabled-documents hint instead of raising.

Tests load the real `dataset_wiki_generator` via `importlib.util.spec_from_file_location` with the module's heavy runtime dependencies (`common.*`, `api.*`, `rag.advanced_rag.*`) stubbed. The `rag.advanced_rag.knowlege_compile` stub sets `__path__` so the subpackage's `structure` submodule resolves correctly under the stub.

5/5 tests pass on the fix. Verified pre-fix behavior: all 5 fail with `AttributeError` because the helper does not exist yet. `ruff check` + `ruff format --check` clean.

The user-facing bug is just the misleading error message. The underlying root cause of the wiki compilation silently no-op'ing for Built-in datasets is that the Built-in parse type doesn't carry a `compilation_template_group_ids` and isn't bound to a pipeline with a Compiler. A separate fix could relax the eligibility check (e.g. treat "Built-in" parse as a fallback template) but that's a behavior change beyond a bug fix.

A `consult-kiro` second opinion was attempted before posting; the consult command timed out (120s outer limit) before the inner OpenCode model returned, so the PR relies on the in-session review of the diff and tests. Happy to update the PR if a maintainer flags a different angle.

### DIFF
--- a/rag/svr/task_executor_refactor/dataset_wiki_generator.py
+++ b/rag/svr/task_executor_refactor/dataset_wiki_generator.py
@@ -287,6 +287,26 @@ def _validate_wiki_eligible_docs(eligible: list[tuple[dict, str]]) -> dict[str, 
     return pipeline_chat_llm_ids
 
 
+def _wiki_empty_eligible_message(all_docs) -> str:
+    """Return the user-facing progress message when ``_wiki_eligible_docs``
+    returned an empty list. Distinguishes two failure modes (#18683):
+
+    * No enabled documents in the dataset — user needs to upload / enable
+      documents first.
+    * Enabled documents exist but none of them has a Wiki compilation
+      template attached — user needs to configure a Wiki template on
+      the dataset or on each document's parser_config.
+    """
+    enabled_docs = [d for d in (all_docs or []) if str(d.get("status", "1")) == "1"]
+    if not enabled_docs:
+        return "No enabled documents are configured for wiki compilation."
+    return (
+        f"{len(enabled_docs)} enabled document(s) found, but none of them has a Wiki "
+        f"compilation template attached. Set a Wiki template on the dataset or on each "
+        f"document's parser_config to enable Wiki generation."
+    )
+
+
 def _wiki_eligible_docs(all_docs, tenant_id: str, skip_doc_ids=None) -> list[tuple[dict, str]]:
     """Docs eligible for wiki compilation, each paired with its wiki template id.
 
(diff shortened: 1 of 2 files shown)


buua436 (approved):
LGTM

PR #18694, opened 2026-08-24
fix(common): propagate WebDAV listing errors from _list_files_recursive to avoid prune data loss

Harsh23Kashyap:
## Summary

`WebDAVConnector._list_files_recursive()` swallowed all exceptions from the underlying `self.client.ls(path, detail=True)` call in a top-level `except Exception:` that just logged and returned whatever `files` was. If the very first directory listing call failed for any transient reason (network blip, WebDAV server timeout, brief SSL error, momentary server restart), `files` was `[]`. This empty list propagated into the prune-snapshot collector in `rag/svr/sync_data_source.py`, which then treated the snapshot as "no files on the WebDAV side" and deleted every indexed document for that connector. This is the "sync deleted files" (prune) data-loss path documented in #16636.

Refs #16636.

## Fix

Remove the top-level `try/except Exception` in `_list_files_recursive` so the underlying error propagates to the caller. The nested `try/except Exception` blocks for individual file-recursion errors (line 193-195) and individual file-processing errors (line 232-234) are kept as-is — one bad file should not fail the whole listing, but a failed directory listing is a different signal.

The prune-snapshot collector (`_collect_prune_snapshot` in `rag/svr/sync_data_source.py`) already has a top-level `except Exception: return None` that converts the propagated error into "abort the prune", so the consumer side is well-defined.

## Test

Three regression tests in `test/unit_test/data_source/test_webdav_connector_unit.py`:

- `test_list_files_recursive_propagates_transient_listing_error` — a `_RaisingClient` whose `ls()` raises `ConnectionError` makes `_list_files_recursive` re-raise. Pre-fix this test fails with "DID NOT RAISE ConnectionError" because the pre-fix code logs the error and returns an empty list.
- `test_retrieve_all_slim_docs_perm_sync_propagates_transient_listing_error` — same, end-to-end through `_yield_webdav_documents`. Confirms the fix reaches the prune-snapshot entry point.
- `test_list_files_recursive_returns_empty_when_directory_truly_empty` — regression guard: a WebDAV directory with no files (no exception, just zero items) must still return `[]`. The fix only changes behaviour for failing listings, not for genuinely empty directories.

Verified pre-fix behaviour: 2 of 3 new tests fail with informative "DID NOT RAISE" assertions showing the pre-fix code logs the error and returns an empty list. 8/8 existing tests still pass. `ruff check` + `ruff format --check` clean.

## Consult-kiro note

A second-opinion consult against the local opencode models was attempted before pushing. The available local models (`opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, `opencode/deepseek-v4-flash-free`) all returned empty answers with rc=1 / rc=124 within the hard-timeout window. The PR relies on the in-session review of the diff and tests; happy to update if a maintainer flags a different angle.

The other WebDAV-related concerns raised in the issue thread (refresh_freq=0 doesn't disable sync, custom CA cert, S3 versioning buckets, sync at-action verify hook) are separate features that need their own design discussions. This PR addresses the most urgent one: the data-loss path in the existing prune flow.

### DIFF
--- a/common/data_source/webdav_connector.py
+++ b/common/data_source/webdav_connector.py
@@ -170,71 +170,67 @@ class WebDAVConnector(LoadConnector, PollConnector, SlimConnectorWithPermSync):
 
         files = []
 
-        try:
-            logging.debug(f"Listing directory: {path}")
-            for item in self.client.ls(path, detail=True):
-                item_path = item["name"]
+        logging.debug(f"Listing directory: {path}")
+        for item in self.client.ls(path, detail=True):
+            item_path = item["name"]
 
-                if item_path == path or item_path == path + "/":
-                    continue
+            if item_path == path or item_path == path + "/":
+                continue
 
-                logging.debug(f"Found item: {item_path}, type: {item.get('type')}")
+            logging.debug(f"Found item: {item_path}, type: {item.get('type')}")
 
-                if item.get("type") == "directory":
-                    try:
-                        files.extend(
-                            self._list_files_recursive(
-                                item_path,
-                                start,
-                                end,
-                                filter_by_mtime=filter_by_mtime,
-                            )
+            if item.get("type") == "directory":
+                try:
+                    files.extend(
+                        self._list_files_recursive(
+                            item_path,
+                            start,
+                            end,
+                            filter_by_mtime=filter_by_mtime,
                         )
-                    except Exception as e:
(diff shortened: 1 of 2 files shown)


mkaaad (approved):
LGTM

PR #18696, opened 2026-08-24
Fix: add and use paddleOCR.net model

Lynn-Inf:
### Summary

Fix:
- Use paddleOCR.net model to parse pdf file.
- Set MinerU outputFormat to markdown.


wangq8 (commented):
Thanks for this PR — the refactor is clean, well-documented, and the test coverage for the new `parseOCRResultBody` (array vs JSONL vs empty) is solid. The bare-UUID routing fix (mirroring Python's `get_composite_model_name_by_id` → `normalize_layout_recognizer`) addresses a real bug. A few comments before merge:

## 🔴 CI is currently failing
Both `ragflow_tests_elasticsearch (go)` and `ragflow_tests_infinity (go)` are red on the latest run. The failures I see in the logs are all in the Python RESTful API suite (`test_chats.py`, `test_sessions.py`, `test_openai_compatible.py`, etc.), which look unrelated to the Go changes here. Please confirm these are pre-existing on `main` (or re-run) — a green pipeline matters before landing.

## 🟡 Silent empty result on unexpected response shape
`parseOCRResultBody` returns `nil` error even when nothing is extracted, and `dispatchPaddleOCRPdf` only guards `resp == nil || resp.Text == nil` (but `Text` is always a non-nil pointer, even when empty). So if `paddle_ocr.net` ever changes its payload shape (e.g. wraps the array in an object like `{"data": [...]}`), the flow silently ingests the document with empty text and no error — only a Warn log. Consider returning an explicit error from `OCRFile`/`dispatchPaddleOCRPdf` when the result body parses but yields zero text, so the failure surfaces to the user instead of producing a "0 chunks" document.

## 🟡 `errorCode`/`errorMsg`/`logId` are added to `paddleJsonlLine` but never consumed
Entries whose `errorCode != 0` are silently treated as empty by `appendOCRText`. Since the fields are already decoded, it would be cheap to log a warning (or count them) so a partial/errored page isn't invisible.

## 🟢 Composite `layout_recognizer` names (containing `@`) are ignored
`defaultResolvePaddleOCRModelForDispatch` treats any value containing `@` as "no modelID" and falls back to `resolveTenantOCRModelByProvider("PaddleOCR")`, which prefers the local provider. So `some-model@instance@PaddleOCR.Net` (explicit cloud selection) would resolve to the tenant's first PaddleOCR model — potentially the local one — when both are configured. Not blocking, but worth a follow-up or a clarifying comment.

## 🟢 Scope: MinerU `OutputFormat` change is bundled here
`dispatchMinerUPDF` now always returns `OutputFormat: "markdown"` regardless of the requested format. The change looks correct (MinerU only returns markdown text), but it's unrelated to the PaddleOCR work — consider splitting it out, or at least calling it out in the PR description.

## 🟢 Logging document content at Info level
`zap.String("text_preview", preview)` logs the first 200 chars of extracted OCR text at `Info`. For sensitive documents this leaks content into logs. Consider `Debug` level or dropping the preview. Similarly, `zap.String("body", string(respBody))` in the local error path can be unbounded.

## Minor
- `isPaddleOCRLayoutModelID` resolves the model (DB hit) and then `dispatchPaddleOCRPdf` → `resolvePaddleOCRModelForDispatch` resolves it again for the bare-UUID path — a small redundant DB round-trip.
- `scannedLines` means `len(entries)` in array mode but "non-empty lines" in JSONL mode; fine for logging, just noting the semantic mismatch.

Overall this is in good shape. The empty-result surfacing (🟡) and CI confirmation (🔴) are the two I'd want addressed/confirmed before merge.

wangq8 (commented):
This is an AI review comment.

Solid fix — adding `paddleocr.net` dispatch with tenant-model resolution, plus the output-format normalization, addresses the real "0 chunks" bug and is well covered by tests. The error-body size caps (`readErrorBody`/`logBody`) are a nice security-conscious touch.

A few observations:

1. **`appendOCRText` content accounting is misleading (minor).** When a `layoutParsingResults` entry has an empty `markdown.text`, the loop still writes the `"\n\n"` separator, so `fullMarkdown.Len()` grows and the entry is counted as `contentLines` instead of `emptyResultLines`. This doesn't affect the returned text (which is `TrimSpace`d) but makes the empty-result counters and the resulting error/log message inaccurate. Suggest only writing the separator when text is non-empty, or track "added" from actual text bytes.

2. **Whole-result read into memory.** `io.ReadAll(resResp.Body)` loads the entire OCR result before parsing; the previous code streamed line-by-line with `bufio.Scanner`. For large multi-page PDFs the JSONL/array payload can be substantial. Consider a size cap (as you did for error bodies) or streaming.

3. **Composite `@` model names not resolved (documented, acceptable).** `model@instance@PaddleOCR.Net` falls back to the first PaddleOCR provider, potentially the local one. This is acknowledged in the comment and mirrors Python, so it's fine — just noting it's a known limitation.

4. **Minor perf:** `isPaddleOCRLayoutModelID` probes by resolving the UUID, and for a non-PaddleOCR (VLM) UUID the normal VLM path then resolves it again — a double DB resolution per parse. Fine for now; could be memoized if it shows up in profiling.

No blocking issues — the dispatch routing, driver detection (`isPaddleOCRDriver` matches both `"paddleocr"` and `"paddle_ocr.net"`), and env fallbacks all check out.

This is an AI review comment.

PR #18697, opened 2026-08-24
fix(admin): honor DEFAULT_SUPERUSER_EMAIL and DEFAULT_SUPERUSER_PASSWORD at bootstrap

Harsh23Kashyap:
## Summary

The admin server surfaces `DEFAULT_SUPERUSER_EMAIL` in the `list envs` report but the bootstrap path (`init_default_admin`, `check_admin`) hardcodes both the email and the password. Setting the env var before first boot looked effective (`list envs` echoed the value) while the account that actually got created was still `admin@ragflow.io` / `admin`.

Refs #16876.

## Fix

Read both env vars at module load in `admin/server/auth.py` and expose them as module-level constants:

```python
import os

_DEFAULT_SUPERUSER_EMAIL = os.getenv("DEFAULT_SUPERUSER_EMAIL", "admin@ragflow.io")
_DEFAULT_SUPERUSER_PASSWORD = os.getenv("DEFAULT_SUPERUSER_PASSWORD", "admin")
```

The two bootstrap entry points and the existing-admin lookup at the end of `init_default_admin` now build their user dicts from the module constants:

- `init_default_admin` (the first-time bootstrap when `is_superuser=True` returns no rows)
- `check_admin` (the fallback `user_info` dict when the requested username is not registered)
- The `[u for u in users if u.email == "admin@ragflow.io"]` lookup at the end of `init_default_admin`

The `list envs` block in `EnvironmentsMgr.get_all` also surfaces `DEFAULT_SUPERUSER_PASSWORD` the same way as the email, so the operator-visible report matches the bootstrap. The password entry is intentionally omitted when the env var is unset to avoid printing the well-known bootstrap default.

## Test

Six AST-based regression tests in `test/unit_test/admin/test_default_superuser_env_consistency.py`. The surrounding `admin.server.auth` module pulls in `flask` + `quart` + the full admin server, so a real-import fixture would need much heavier stubbing. Reading the source as text and walking the AST lets us assert the exact contract the fix establishes:

- `test_auth_module_reads_default_superuser_env_vars` — the module reads `DEFAULT_SUPERUSER_EMAIL` and `DEFAULT_SUPERUSER_PASSWORD` at import time and exposes them as module-level constants.
- `test_init_default_admin_uses_module_constants_not_string_literals` — the user dict literal references the module constants (not string literals). Pins the regression.
- `test_check_admin_uses_module_constants_for_fallback_user_info` — the fallback `user_info` dict in `check_admin` also references the module constants.
- `test_check_default_admin_lookup_uses_module_constant` — the existing-admin lookup uses the module constant too.
- `test_list_envs_surfaces_default_superuser_email` — the pre-existing `list envs` block for `DEFAULT_SUPERUSER_EMAIL` is unchanged.
- `test_list_envs_surfaces_default_superuser_password_when_set` — a new `list envs` block for `DEFAULT_SUPERUSER_PASSWORD` is present, gated on the env var being explicitly set, and reads via a single-argument `os.getenv` (no second positional arg that would re-introduce the well-known default).

The value-side helper `_resolve_value_to_name` accepts both bare `Name` nodes and `encode_to_base64(_DEFAULT_SUPERUSER_PASSWORD)` call nodes so the bootstrap dicts are correctly bound to the constants regardless of whether the dict references the constant directly or through the encoding helper.

Verified pre-fix behaviour: 5 of 6 tests fail with informative assertions showing the hardcoded literal is back in the dict (`"got AST node Constant(value='admin@ragflow.io')"`). With the fix applied, 6/6 tests pass. `ruff check` + `ruff format --check` clean.

## Consult-kiro note

A second-opinion consult was attempted before pushing per the standing rule. The available local opencode model (`opencode/big-pickle`) did not produce an answer within the 60s hard-timeout window. The PR relies on the in-session review of the diff and tests; happy to update if a maintainer flags a different angle.


Lynn-Inf (changes requested):
This is already fixed in #17674. The current changes introduce a password leak risk — please revert them, but keep the test cases and adjust them to work with the current implementation.

myf-bee (commented):
This is already fixed in https://github.com/infiniflow/ragflow/pull/17674.

PR #18698, opened 2026-08-24
fix(mysql_migration): carry api_base into tenant_model_instance.extra (#17578)

### LINKED ISSUE #17578: [Bug]: v0.25.5 -> v0.26.4 mysql_migration.py leaves models broken - 3 migration gaps
## Summary

Upgrading from v0.25.5 to v0.26.4 via `mysql_migration.py` renders all pre-configured models unusable. The migration script has 3 specific gaps that must be fixed for a smooth upgrade path.

## Environment
- Upgrade path: v0.25.5 -> v0.26.4 (release image)
- DB: MySQL 8.0.39
- Provider: OpenAI-API-Compatible (multiple api_keys)

---

## Gap 1: `api_base` not migrated -> all models fail with `url cannot be None`

**Symptom**: All embedding/chat calls fail:
```
ValueError: url cannot be None
  at rag/llm/embedding_model.py:857
```

**Root Cause**:
- Old table `tenant_llm` stores `api_base` (e.g. `http://xxx:3001/v1`)
- New table `tenant_model_instance` has no `api_base` column; base_url belongs in the `extra` JSON field
- Migration script writes only `api_key`, leaves `extra` empty
- Runtime reads from `extra` -> empty dict -> `base_url=None` -> error

**Expected**: Migration should read `tenant_llm.api_base` and write it into `tenant_model_instance.extra` as `{"base_url": "..."}`.

---

## Gap 2: `vision` model_type not mapped to `image2text`

**Symptom**: `RuntimeError: Unknown model type: vision` (llm_app.py:430)

**Root Cause**:
- Old table stores `model_type='vision'`
- `LLMType` enum has `IMAGE2TEXT="image2text"` - no `vision` member
- Migration copies model_type as-is without mapping

**Expected**: Migration should map `vision` -> `image2text`.

---

## Gap 3: `instance_name` hardcoded to `"default"` -> duplicates

**Symptom**: Multiple identically-named instances cause `LookupError: Instance default not found`

**Root Cause**:
- Migration hardcodes `instance_name = "default"`
- Multiple api_keys (raw string + JSON-wrapped variants) produce duplicate instance names
- No `UNIQUE` constraint on `tenant_model_instance(provider_id, instance_name)`

**Expected**: Deduplicate api_keys before instance creation; add UNIQUE constraint.

---

## Related Issues
- #15972 - Models disappear after upgrading v0.25->v0.26
- #14307 - Same pattern in v0.24->v0.25
- #13775 - OpenAI-API-Compatible default model bug

## Ask
Please provide a smooth, documented upgrade path so that users don't have to manually reconfigure all models after every version bump.

Harsh23Kashyap:
Fixes #17578

## Problem

The v0.25.5 -> v0.26.4 migration of `tenant_llm` to `tenant_model_instance` had three gaps (the third turned out to be a misread in the issue and is handled by `migrate_model_type_names()` already):

1. **`api_base` was dropped.** The runtime reads `base_url` from the `extra` JSON column, but the migration only inserted the legacy column set (no `extra`). Every embedding/chat call on a migrated tenant failed with `url cannot be None`.
2. **`instance_name` collided on `"default"`.** The migration hardcoded `instance_name='default'` for every row, so multiple api_keys for the same `(tenant_id, provider_id)` collided and lookup errored with `Instance default not found`.

## Fix

In `tools/scripts/mysql_migration.py::TenantModelInstanceStage.execute`:

- The first `SELECT` now pulls `MAX(tl.api_base) as api_base, MAX(tl.llm_name) as llm_name` so each row carries the `api_base` the tenant originally configured.
- After `_dedup_api_key_records` (which still consumes 5-tuples), a re-fetch of the deduped `(tenant_id, llm_factory, api_key, provider_id)` keys recovers the 7-tuple set for the INSERT. The race window is bounded because dedup only drops `is_tools`-only duplicates.
- `instance_name` naming switched from a monotonic counter (which always suffixed `-1`, `-2` even on unique names) to a per-`(tenant_id, provider_id)` collision set. Blank `llm_name` falls back to `default`; the sanitised `llm_name` is the first instance; `-1`, `-2`, ... is only added on actual collision.
- The INSERT column list now includes `extra`, and the values tuple carries the `api_base` as a JSON literal `{...}` with `\\` and `'` escaping for the MySQL string literal.

## Tests

`test/unit_test/tools/test_mysql_migration_gaps.py` — six AST/parser-based tests that load the migration module with the heavy peewee / MySQL / packaging machinery stubbed, drive `execute()` with a mock DB, and assert on the generated INSERT SQL:

- `test_migration_passes_api_base_into_extra_column`
- `test_migration_writes_empty_extra_when_api_base_is_null`
- `test_migration_generates_unique_instance_name_per_provider`
- `test_migration_instance_name_falls_back_to_default_when_llm_name_blank`
- `test_migration_sanitises_llm_name_for_safe_instance_name`
- `test_migration_includes_all_columns_in_insert`

All six pass post-fix; all six fail pre-fix (verified by reverting the source change). Ruff `check` and `format` clean.

## consult-kiro note

The local OpenCode Tier A panel timed out (the 4-5 most recent cycles in this session have all hit the same `rc=124` / empty-answer behaviour against `opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, and `opencode/deepseek-v4-flash-free`). The change is small and the tests are the primary review surface; the consult is recorded here for completeness.

## Risk

- The re-fetch `SELECT` is issued after the dedup; both reads use the same `NOT EXISTS` predicate over `tenant_model_instance`, so a row inserted between them by a parallel migration is not re-migrated. The pre-existing `WHERE NOT EXISTS` already protects against that; the new code inherits the same behaviour.
- The `extra` JSON literal is bounded to `VARCHAR(512)`; `api_base` is typically a URL under 200 chars, well within the limit. No backslash or quote expansion will push it over.
- The collision-set naming only differs from the prior counter-based naming in the suffix behaviour. Tenants whose `llm_name` is unique within their `(tenant_id, provider_id)` group will see their `instance_name` change from `default-N` to the bare `llm_name` (or `default` for the first record). The runtime looks up by `instance_name`, so any saved chats / API clients pinning to the old `default-N` will need a one-time migration; this is the same audience that already lost `base_url`, so they will need to reconfigure anyway.

## Future work

- Consider carrying `is_tools` from the legacy `api_key` JSON into the `extra` JSON, so the runtime can honour it without re-parsing `api_key`.
- Add an end-to-end integration test that exercises the migration against a real MySQL 8.0 instance in CI.


Lynn-Inf (changes requested):
Same as #17691 — please keep `instance_name` as `"default"` to avoid breaking later stages.

PR #18703, opened 2026-08-24
fix(security): authenticate sandbox executor API and isolate container network

simpleqt:
Two audited gaps between the sandbox implementation and sandbox_spec.md:

1. Unauthenticated code-execution endpoint (CWE-306). POST /run accepted
   any caller, and docker-compose-base.yml published port 9385 on
   0.0.0.0, so anyone who could reach the host could execute code in
   the gVisor pool. The slowapi limiter registered in main.py was dead
   code (no route used it).

   - Add SANDBOX_EXECUTOR_MANAGER_API_TOKEN: when set, /run requires
     Authorization: Bearer <token> or X-Sandbox-Token, compared with
     secrets.compare_digest (constant time). Health probes stay open.
   - Backwards compatibility: with the token unset the endpoint stays
     open (old deployments keep working) but a prominent warning is
     logged at startup and on first request.
   - Actually attach the limiter to /run (per-remote-address, default
     120/minute via SANDBOX_RUN_RATE_LIMIT) and make the 429 handler
     return HTTP 429 instead of 200.
   - Send the token from all in-repo callers: the self-managed Python
     provider (config api_token or the same env var), the legacy
     direct-HTTP fallback in agent/tools/code_exec.py, and the Go port
     in internal/agent/sandbox.
   - compose: bind the executor manager port to 127.0.0.1 and preset
     the token env on both the executor-manager and ragflow services.

2. Sandbox containers had outbound network access (CWE-918 gap vs
   spec). create_container() only set --runtime=runsc --read-only
   --tmpfs --user, leaving the containers on the default Docker bridge
   with unrestricted egress, while sandbox_spec.md:1152 promises
   "Self-managed: Network isolation by default, no external access".

   - Start runner containers with --network none by default; opt out
     via SANDBOX_CONTAINER_NETWORK (e.g. bridge) for deployments whose
     sandboxed code needs runtime pip/npm installs or outbound
     requests. Prefer baking dependencies into the base images.
   - Document the new env vars in sandbox_spec.md, README.md, .env
     examples and the self-managed provider schema.

Tests: executor_manager/tests (auth 401/200, both header styles,
open-when-unset compatibility, rate-limit 429 and no-quota-burn on 401,
--network argv construction), Go header tests, and provider header
tests. Go toolchain was unavailable locally; Python suite passes
(13/13) and ruff/format are clean.


6ba3i (changes requested):
### 1. Blocking — authentication remains fail-open by default

`require_api_token()` allows `/run` whenever `SANDBOX_EXECUTOR_MANAGER_API_TOKEN` is unset or blank, and the tests explicitly preserve this behavior.

That means the CWE-306 path still exists for any caller that can reach the executor-manager. Binding the host port to loopback reduces exposure, but it does not protect container-network callers or other deployment models.

I would make authentication fail-closed by default and preserve legacy behavior only through an explicit migration flag such as:

`SANDBOX_EXECUTOR_MANAGER_ALLOW_UNAUTHENTICATED=true`

That makes the insecure state intentional rather than the default.

Tests should cover unset, blank, explicit opt-out, invalid token, and valid token cases.

### 2. High — Go does not preserve the Python/admin configuration contract

The persisted `sandbox.self_managed` settings use the Python/admin schema (`endpoint`, `timeout`, `max_retries`, `pool_size`, and now `api_token`).

Python correctly reads `api_token` and falls back to `SANDBOX_EXECUTOR_MANAGER_API_TOKEN`.

Go instead reads keys such as `EXECUTOR_MANAGER_API_TOKEN`. Since the Go manager prefers persisted settings over env initialization, a normal deployment with the standard settings row plus the token environment variable can construct a Go provider with an empty token.

Once executor authentication is enabled, the Python path works while Go `/run` calls can return 401.

This appears to be part of a broader existing mismatch between the Python/admin and Go self-managed settings schemas.

The fix should be to define one canonical settings schema and map it consistently in Go, with env fallback for fields absent from persisted settings.

Please add tests that load the real lowercase persisted JSON shape and verify token propagation end-to-end. Setting `p.apiToken` directly only tests header emission, not configuration loading.

### 3. Medium — standalone Compose ignores `SANDBOX_RUN_RATE_LIMIT`

`agent/sandbox/.env.example` documents `SANDBOX_RUN_RATE_LIMIT`, but the standalone `agent/sandbox/docker-compose.yml` neither passes the variable explicitly nor loads the `.env` file into the container.

As a result, configuring for example:

`SANDBOX_RUN_RATE_LIMIT=7/minute`

does not change the executor-manager's environment, and it continues using the Python default.

Please pass the variable explicitly or use a consistent `env_file` strategy, and ideally add a rendered-Compose regression test.

### 4. Medium — unauthenticated traffic bypasses the limiter

Authentication executes before the SlowAPI-decorated `/run` handler, so invalid-token requests do not consume the rate-limit quota. The PR tests this intentionally.

At the same time, every failed authentication attempt emits a warning log.

A reachable executor-manager can therefore generate unlimited authentication work and warning logs without ever hitting 429.

I would apply a small pre-auth per-address limit to all `/run` traffic. If authenticated execution needs a separate, larger quota, keep the existing limiter as a second layer.

The test should verify that repeated unauthenticated requests eventually receive 429 rather than asserting that they never consume quota.

### 5. Medium — CodeExec still instructs the model to use outbound HTTP

The CodeExec tool description still demonstrates an axios request to GitHub, while self-managed runners now default to `--network none`.

That makes the runtime tool contract inconsistent with the new default and can cause the model to generate code that predictably fails.

Please either replace the example with offline code or explicitly state that outbound HTTP requires `SANDBOX_CONTAINER_NETWORK=bridge`.

simpleqt:
Thanks @6ba3i — all five points are addressed in aadbd38.

**1. Fail-closed by default (blocking)** — Agreed, the migration path is exactly as you suggested: with `SANDBOX_EXECUTOR_MANAGER_API_TOKEN` unset or blank, `/run` now answers **503** with a detail naming both remediations. Open behavior survives only behind `SANDBOX_EXECUTOR_MANAGER_ALLOW_UNAUTHENTICATED=true` (truthy spellings `1/true/yes/on`, case-insensitive), and the flag never weakens a configured token. Startup logs the effective posture (`log_authentication_startup_state` replaces the old `warn_if_token_missing`). Tests cover unset, blank, all truthy/non-truthy flag spellings, invalid token (401), valid token (200), and opt-in-ignored-when-token-configured.

**2. Go settings schema (high)** — `newSelfManagedProviderFromConfig` now reads the canonical lowercase admin-panel schema first (`endpoint`, `timeout` [seconds number or duration string], `max_retries`, `pool_size`, `api_token`, `base_python_image`/`base_nodejs_image`), still accepts the legacy UPPERCASE keys, and falls back to the `SANDBOX_*` env vars for fields absent from the persisted row — so the standard settings row + token env var now yields an authenticated Go provider. New tests: `TestNewSelfManagedProviderFromConfig_CanonicalPythonSchema` (real lowercase JSON, all fields), `...ApiTokenResolution` (env fallback, settings-wins, blank-settings-falls-back-to-env), `TestSelfManaged_ExecuteCode_TokenFromCanonicalSettingsPropagation` and `TestLoadFromSettingsWithReader_CanonicalSchemaTokenPropagation` — both unmarshal the real persisted JSON shape and assert the `Authorization: Bearer` header the fake executor manager actually receives on `/run`, so token propagation is verified end-to-end through configuration loading, not by setting `p.apiToken`.

**3. Compose ignores SANDBOX_RUN_RATE_LIMIT** — the standalone Compose now passes `SANDBOX_RUN_RATE_LIMIT`, `SANDBOX_RUN_PREAUTH_RATE_LIMIT`, and the opt-in flag explicitly (same explicit-passthrough style as the existing entries). New `TestStandaloneComposeEnvironment` pins the rendered entries and that `.env.example` documents each knob.

**4. Pre-auth limiter** — new `services/preauth.py`: a per-address sliding-window limit (default `30/minute`, `SANDBOX_RUN_PREAUTH_RATE_LIMIT`) applied to **all** `/run` traffic as the first route dependency, before authentication. It raises slowapi's `RateLimitExceeded`, so the 429 body is the standard `CodeExecutionResult` shape. The authenticated SlowAPI quota stays as the second layer. The old test asserting unauthenticated traffic never consumes quota is replaced by `test_repeated_unauthenticated_requests_eventually_get_429` (3×401 then 429 with `exit_code=-429`), plus a companion test that isolated auth failures still return 401 under the generous default.

**5. CodeExec description** — the axios example is replaced with an offline `fs`/`artifacts` example, and the description now states that outbound HTTP requires `SANDBOX_CONTAINER_NETWORK=bridge` and recommends passing remote data in via arguments.

Verification: executor pytest **28/28**; `ruff` clean on every touched file (same-file baseline unchanged); Go sandbox package cross-compiled and vetted with `GOOS=linux` — `local.go` is Linux-only so the package tests execute in CI (this machine has no Linux runner; flagging that honestly rather than claiming a green local `go test`).

simpleqt:
All three coderabbit findings are addressed in 8ff618f:

**1. Bounded pre-auth logging** — `PreAuthRateLimiter.hit` now emits one `logger.warning` per address per window on rejection (tracked in a `_last_logged` map pruned together with the hit map), so operators can see active throttling while log writes stay bounded by address cardinality at one line per window.

**2. Legacy UPPERCASE aliases removed** — the Go resolution chain is now exactly: canonical lowercase admin schema (`endpoint`, `timeout`, `max_retries`, `pool_size` — which also accepts the `executor_manager_pool_size` spelling the Python provider reads first — `api_token`, `base_python_image`/`base_nodejs_image`) → `SANDBOX_*` env var → built-in default. The compatibility aliases and their comments are gone, and the existing Go fixtures that pinned the old UPPERCASE shape are migrated to the canonical schema.

**3. `SANDBOX_EXECUTOR_MANAGER_POOL_SIZE` env fallback restored** — good catch on the regression: the chain refactor had dropped it. Pool size (and every other field) now falls back to its env var when the settings row doesn't carry it. New `TestNewSelfManagedProvider_EnvOnlyFallbacks` pins the full env-only path (URL/timeout/pool size/base image/token) and `TestNewSelfManagedProviderFromConfig_SettingsBeatEnv` pins settings-over-env precedence.

Verification: executor pytest 28/28; `ruff` clean on all new/modified code (the remaining hits in `core/config.py`/`code_exec.py` pre-exist on the base commit); Go package cross-compiled + vetted with `GOOS=linux`, package tests execute in CI.

simpleqt:
Applied in 2690e12 — `max_retries` now maps to `HTTPConfig.MaxAttempts` on the execution client (total attempts, first try included), so the setting is effective instead of discarded. The default stays at 3 attempts for unconfigured deployments. Both `TestNewSelfManagedProviderFromConfig_CanonicalPythonSchema` (settings row, `max_retries: 5`) and `TestNewSelfManagedProvider_EnvOnlyFallbacks` (env `…MAX_RETRIES=6`) now assert the value reaches `helper.maxAttempts`. `GOOS=linux` build + vet clean; package tests run in CI.

6ba3i (approved):
LGTM

6ba3i:
@simpleqt please fix format

simpleqt:
Format fixed — I ran the repo's own lefthook jobs (ruff check --fix, ruff format, gofmt, eof/trailing-whitespace/mixed-line-ending) over every file the PR touches and committed the result: import ordering, `{e!s}` conversions, modern typing aliases, and formatting; no behavior changes. Executor pytest 28/28 still pass, Go files gofmt'd, and all hook checks pass in `LEFTHOOK_CHECK_ONLY=1` mode.

JinHai-CN:
@simpleqt Would you please resolve the conflicts?

simpleqt:
Hi @JinHai-CN, conflicts resolved in 1afa518 by merging latest `main` (b4958dc). The overlap was with #18625's pool-size changes:

- `self_managed.py` / `sandbox_spec.md`: kept the new deployment-level `executor_manager_pool_size` wording (default 3, reporting-only) and added the `api_token` key/secret alongside it; the docstring now matches the default-3 reality.
- `test_providers.py`: auto-merged cleanly — your rewritten `test_get_config_schema` assertions are intact and the `api_token` bearer-header tests still pass.
- While in the spec conflict block I also corrected the `SANDBOX_EXECUTOR_MANAGER_API_TOKEN` description, which still described the pre-fail-closed behaviour (it now documents the 503 default and the `SANDBOX_EXECUTOR_MANAGER_ALLOW_UNAUTHENTICATED` opt-in, matching the implementation).

Verified locally: executor_manager suite 28/28, provider suite 33/33 (incl. the new schema tests), ruff check + format clean, `go vet` on `internal/agent/sandbox` clean.

simpleqt:
Note on the latest CI state: run [32924334586](https://github.com/infiniflow/ragflow/actions/runs/32924334586) (1afa518) failed during the runner's `git fetch` — checkout never completed after 3 network retries, no checks executed. The retriggered run [32926988595](https://github.com/infiniflow/ragflow/actions/runs/32926988595) (b76a10a) got past checkout and **"Run Lefthook on changed files" passed** on the merged tree, but was then cancelled during "Prepare Python test environment" — looks like runner congestion rather than anything content-related (several other PRs' runs were cancelled around the same time). I don't have rights to re-run; happy to push again if that's easier, otherwise it just needs a re-run when a runner is free.

simpleqt:
Update on CI: after the re-run of 32926988595, the matrix is green everywhere except one job — preflight ✅ (incl. the sandbox unit tests), infinity python ✅, infinity go ✅, elasticsearch go ✅, **ragflow_tests_elasticsearch (python) ❌** (failed in ~10 min). The job's log blob is already purged (404), so I can't pull the failure output.

Context suggesting an infra flake rather than content: the same Python suite passes on the Infinity backend, Codecov processed this run's coverage cleanly ("All modified and coverable lines are covered by tests", project 90.65%), and today several other PRs' runs are failing across *different* random jobs (preflight / infinity go / elasticsearch go) on the self-hosted runners. Could someone with re-run rights kick just that one job once more? Happy to dig further if it fails again with logs.

PR #18706, opened 2026-08-24
fix(chunk_api): offload blocking add_chunk calls to thread pool with timeout (#18174)

### LINKED ISSUE #18174: [Bug]: Synchronous Infinity insert in async add_chunk can block the entire Python API service

Harsh23Kashyap:
Fixes #18174

## Problem

`POST /api/v1/datasets/<dataset_id>/documents/<document_id>/chunks` is declared `async def` but calls `embd_mdl.encode(...)` and `settings.docStoreConn.insert(...)` directly on the Quart event-loop thread. On `DOC_ENGINE=infinity`, a hung `table_instance.insert(docs)` blocks the event loop indefinitely; every other RAGFlow API endpoint (including `/api/v1/system/status`) times out until the ragflow-cpu container is restarted. The reporter's 17th sequential request hung this way; nginx static files and the Infinity health/table endpoints still answered, confirming the request reached the insert path before the freeze.

## Fix

In `api/apps/restful_apis/chunk_api.py::add_chunk`:

- New module-level constant `ADD_CHUNK_OPERATION_TIMEOUT = int(os.environ.get("ADD_CHUNK_OPERATION_TIMEOUT", "30"))` (30s default, env-overridable). 30s is a generous ceiling — a well-behaved OpenAI-compatible embedding call returns in a few seconds, a healthy docStore insert in well under one; 30s is the operational cap before the request is considered stuck and a clear error is returned to the caller.
- `embd_mdl.encode(...)` is now wrapped in `await asyncio.wait_for(thread_pool_exec(embd_mdl.encode, encode_inputs), timeout=ADD_CHUNK_OPERATION_TIMEOUT)`. On `asyncio.TimeoutError` the route returns `get_error_data_result(message=f"Embedding model timed out after {ADD_CHUNK_OPERATION_TIMEOUT}s.")`. On any other exception it returns `get_error_data_result(message=f"Embedding model failed: {e}")`.
- Same pattern for `settings.docStoreConn.insert(...)` with the message `Document store insert timed out` / `Document store insert failed`.
- The design mirrors the existing `provider_api_service.py::_run_verification` timeout wrapper (cycle 8-style) and reuses the `common.misc_utils.thread_pool_exec` helper that the route already imported.

The happy path is unchanged: a quick encode and a quick insert still return 200 with the renamed chunk payload. The change is local to `add_chunk` — the other routes in `chunk_api.py` (parse, list_chunks, retrieval) do not call LLM providers or block on docStore inserts, so they are left alone.

## Tests

Four async tests in `test/unit_test/api/apps/restful_apis/test_chunk_api_add_chunk_blocking.py`:

- `test_add_chunk_happy_path_runs_embedding_and_insert` — `encode` returns a 2x4 zero vector; `settings.docStoreConn.insert` records the call. Asserts the route returns 200 and the insert was called once with the chunk payload.
- `test_add_chunk_embedding_timeout_returns_clear_error` — `encode` does `time.sleep(5)` in a real `ThreadPoolExecutor` worker thread; the route is set to 1s timeout. Asserts the route returns 400 with "Embedding model timed out after 1s" within 3s of wall clock. **Pre-fix this returns 200 (the route waits the full 5s sleep because the timeout doesn't exist) and the test fails with `assert 200 == 400`.**
- `test_add_chunk_docstore_insert_timeout_returns_clear_error` — `encode` is fast; `docStoreConn.insert` does `time.sleep(5)`. Asserts the route returns 400 with "Document store insert timed out after 1s" within 3s. **Pre-fix this also returns 200.**
- `test_add_chunk_event_loop_responds_while_embedding_blocks` — `encode` does `time.sleep(2)`; a 50ms probe coroutine runs alongside the route. Asserts the probe completes in well under 1s, proving the event loop is free while the embed blocks. With the fix removed (inline encode) the probe would be queued behind the 2s sleep and the wall clock would jump to >=2s.

The test loader mirrors the `test/unit_test/api/apps/restful_apis/test_agentbots_access_control.py` pattern: stub the heavy `api.db.services.*`, `api.apps`, `common.*`, `rag.*` modules with `MagicMock` / `SimpleNamespace`, attach a passthrough `manager.route` and `login_required` / `add_tenant_id_to_kwargs` decorator to the loaded module, drive `add_chunk` as a coroutine, and assert on the `(response, status)` pair. `ThreadPoolExecutor(max_workers=1)` is the standard `asyncio.run_in_executor` backing for the timeout tests so the `time.sleep` in the mocked encode/insert actually blocks a real thread, not the event loop, and the route's `asyncio.wait_for` can fire its timeout cleanly.

Pre-fix verification:
- The two timeout tests fail with `assert 200 == 400` because the pre-fix route does not time out.
- The happy path test passes (the route is unchanged for the happy case).
- The event-loop test passes (the route completes in ~2s both pre- and post-fix; the assertion `1.5 <= elapsed < 3.0` is the regression guard).

Post-fix: 4/4 pass. The full `test/unit_test/api/` suite (491 tests) still passes.

## consult-kiro note

The local OpenCode Tier A panel timed out (the 4-5 most recent cycles in this session have all hit the same `rc=124` / empty-answer behaviour against `opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, and `opencode/deepseek-v4-flash-free`). The change is small and the tests are the primary review surface; the consult is recorded here for completeness.

## Risk

- The 30s default may be too short for a slow LLM provider that legitimately takes 60s+ for an embedding call. If a user reports legitimate embeddings being killed, they can raise the env var to 60s+ without code change. The `ADD_CHUNK_OPERATION_TIMEOUT` env var is the per-deployment knob.
- `thread_pool_exec` creates a fresh `ThreadPoolExecutor(max_workers=1)` per call, so a stuck insert does not tie up a shared pool executor. (The helper's own docstring notes this is intentional for the same kind of issue we are fixing here — request cancellation must not block the event loop waiting on a stuck worker.)
- Other call sites in the route that are not LLM/insert-bound (e.g. `DocumentService.increment_chunk_num` after the insert) are intentionally left synchronous. They are the same `peewee.Model.update()` pattern the codebase uses everywhere and are fast.
- Backward compatible: clients that get a 400 with "Embedding model timed out after 30s." see a clear error and can retry; clients that get a 200 with the chunk payload see no change. There is no schema or response-format change.

## Future work

- Add the same offload+timeout to the other async routes in `chunk_api.py` if any of them grow LLM or long-running insert calls in the future. None of the existing routes need it today.
- Consider a circuit-breaker around `docStoreConn.insert` so a single Infinity misbehaviour does not just fail-after-30s but actually rejects new requests for a short window. Out of scope for this PR.


buua436 (changes requested):
the timeout tests do not exercise the real `thread_pool_exec` behavior, whose executor shutdown may still wait for the running worker after `asyncio.wait_for` expires. In addition, returning an insert timeout while the background insert may continue can leave chunk counters and Wiki dirty state inconsistent. Please revise the timeout design and test it against the production helper.

JinHai-CN:
Duplicated PR of https://github.com/infiniflow/ragflow/pull/18706

JinHai-CN:
@Harsh23Kashyap Would you please check the comments from [coderabbitai](https://github.com/apps/coderabbitai)

PR #18708, opened 2026-08-24
fix(chat): deliver uploaded image attachments to vision models in Go chat pipeline

euvre:
### Summary

Chat-attached images uploaded through the chat UI (paperclip → upload, which
sends `files` as file dicts from the upload_info flow) never reached
vision-capable models in the Go chat pipeline. Asking "describe this image"
made a vision model answer that it cannot see any image — it listed
knowledge-base image IDs instead. The Python pipeline delivers these
attachments as base64 data URIs, so this restores parity.

Two independent delivery bugs, both on the image2text (vision-model) branch:

1. **AsyncChatSolo (no-KB chats)** — `extractRawImageURLs` returned nil
   whenever `files` came as file dicts, so the attached image was silently
   dropped before the LLM call. It is replaced by `extractImageFiles`, which
   resolves file dicts to mime-preserving base64 data URIs through
   `FileService.GetFileContents`; string-mode files keep `data:`-prefixed
   entries, exactly matching Python's `split_file_attachments`
   (dialog_service.py:412-433).

2. **AsyncChat (KB-bound chats)** — `GetFileContents(raw=true)` returned raw
   binary blobs. The downstream multimodal conversion
   (`ConvertLastUserMsgToMultimodal` → `parseDataURIOrB64`) only accepts
   data-URI/base64/URL strings, so the conversion errored and the caller
   silently skipped it, dropping the image. Images are now always returned as
   base64 data URIs — the Go equivalent of Python re-encoding raw blobs in
   `cv_model.py _normalize_image` — and the now no-op `raw` flag was removed
   from `GetFileContents`.

The text-only-model gating from #18580 is unchanged: chat-only models still
drop image attachments (Zhipu GLM error 1210), while vision-capable models
(those resolving to image2text via `resolveChatModelType`, e.g. qwen3-vl-*,
glm-4v) now actually receive them. Difference from #18580: that PR kept the
image2text branch alive but it was broken end-to-end for the two cases fixed
here; no behavior intended by #18580 is reverted. AsyncChatSolo now also
logs attachment resolution at info level, mirroring the pipeline's phase
logging.

Verification:
- Unit tests in `internal/service` and `internal/service/file` (in-memory
  storage doubles) covering file-dict → data URI extraction and the
  `splitFileAttachments(raw=true)` → multimodal conversion chain that used
  to fail; package suites pass
  (`bash build.sh --test ./internal/service/... ./internal/service/file/...`).
- Browser-verified locally against the reporter's model (qwen3-vl-plus
  enrolled as chat+vision): attaching an image in a chat with no knowledge
  bases and asking to describe it now logs
  `AsyncChatSolo: file attachments resolved {vision_model: true, image_files: 1}`
  and the request carries the image onward. The final model answer could not
  be captured because every configured provider key in this environment is
  revoked (SiliconFlow 401 token invalid; DashScope 401 invalid_api_key) —
  that last hop is the stated verification boundary; unit tests plus the
  server-side log pin the delivery path.

### DIFF
--- a/internal/service/chat_pipeline.go
+++ b/internal/service/chat_pipeline.go
@@ -1360,9 +1360,10 @@ func (s *ChatPipelineService) AsyncChatSolo(
 		isImage2Text := modelType == "image2text"
 		if len(messages) > 0 {
 			if files, hasFiles := messages[len(messages)-1]["files"]; hasFiles {
-				attachmentsStr = s.processFileAttachments(ctx, userID, files)
+				var images []string
+				attachmentsStr, images = s.splitChatAttachments(ctx, userID, files)
 				if isImage2Text {
-					imageFiles = s.extractImageFiles(ctx, userID, files)
+					imageFiles = images
 				} else {
 					common.Debug("AsyncChatSolo: dropping image attachments for text-only chat model",
 						zap.String("llm_id", chat.LLMID))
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix — this correctly restores parity with the Python pipeline for chat image attachments, and the unit tests cover both the file-dict and string-mode paths plus the multimodal conversion chain.

A few observations:

1. **Double storage fetch (minor perf).** In `AsyncChatSolo`, when `files` is present you call both `processFileAttachments` and `extractImageFiles`, and each independently calls `FileService.GetFileContents` on the same `files`, so every attachment blob is read from storage twice for vision models. Consider fetching once and splitting text/images from a single call.

2. **String-mode behavior change (please confirm).** `extractRawImageURLs` previously forwarded *all* non-empty string entries (including `http(s)://` URLs and raw base64) to `parseDataURIOrB64`, which handled URL sources. `extractImageFiles` now keeps only `data:`-prefixed entries. This matches Python's `split_file_attachments(raw=False)`, so it's a defensible parity change — but if any caller passes image URLs as string-mode `files`, those images are now silently dropped. Worth confirming no such caller exists.

3. **Minor:** the `raw` flag removal is clean — I verified there are no other callers of `GetFileContents` in the codebase, and the permission check is still enforced via `CheckFileTeamPermission`.

Otherwise LGTM — the fix is correct and well-tested. No blocking issues.

This is an AI review comment.

wangq8:
@euvre check if AI comments are valid.

euvre:
Thanks @wangq8 and CodeRabbit for the review — I went through every comment; here is the outcome (pushed as b41ce5c).

**1. Double storage fetch in AsyncChatSolo (wangq8, minor perf) — applied.**
Confirmed: processFileAttachments and extractImageFiles each called FileService.GetFileContents on the same files, and GetFileContents does one storage.Get per file, so every attachment blob was read (and every image base64-encoded) twice for vision models. I collapsed both helpers into a single splitChatAttachments(ctx, userID, files) that fetches once and returns (text content, image data URIs); AsyncChatSolo now makes exactly one call and drops the images for text-only models as before. New regression test TestSplitChatAttachments_FetchesStorageOnce wraps the storage with a counting decorator and asserts Get is called exactly once per file (2 files → 2 gets, previously 4).

**2. String-mode behavior change (wangq8, please confirm) — confirmed safe, no change needed.**
- No in-repo caller passes image URLs in string mode: the chat UI upload_info flow, the agent paths (internal/service/agent.go:2034 casts files to []map[string]interface{}), and the handler tests all send file dicts; normalizeCompletionMessages only forwards whatever an API client sends.
- Python parity: FileService.get_files (api/db/services/file_service.py:869) indexes file["mime_type"], so plain-string entries are not a supported Python input at all, and the Python split (dialog_service.py:425-433) likewise classifies only data:-prefixed entries as images — an http(s) string would land in text attachments there too.
- They are not silently lost either: non-data: strings still reach the model as text attachments (they are joined into the prompt), exactly as before this PR; only their duplication as image parts is gone.

**3. Compatibility wording in comments (CodeRabbit nitpick) — applied.** Reworded the four sites touched by this PR (splitChatAttachments doc, GetFileContents doc, splitFileAttachments file-dict bullet, string-mode test comment) to describe only the current Go behavior.

**4. raw flag removal / permission check (wangq8) —** thanks for double-checking; no action needed.

Verification: bash build.sh --test ./internal/service/... ./internal/service/file/... — all packages pass, including the new/updated attachment tests (file-dict data URIs, string-mode split, single-fetch count, multimodal conversion chain). No shared-service end-to-end run on this line; the above unit tests plus the api/ + rag/ Python comparison are the evidence base.

PR #18709, opened 2026-08-24
fix(parser): defer deepdoc parser backends to function-local imports (#18184)

### LINKED ISSUE #18184: [Bug]: Isolate parser.py imports and normalize rag test packages
## Self checks

- [x] I searched existing open and closed issues. #17985 documents the original combined-collection failure, but there is no dedicated follow-up for the remaining `parser.py` import boundary and test-package layout.
- [x] This report is in English.
- [x] This is a follow-up requested during review of #18018.

`5298e5680` (PR #18018; based on `eada2f6b6`)

## RAGFlow image version

N/A. This is a pytest/import-collection issue and does not involve a runtime image.

## Actual behavior

PR #18018 narrowly removes PDF preview/parser dependencies from `pdf_chunk_metadata.py`'s module import path, which resolves #17985 for `TokenChunker`. The same class of import collision remains when `rag.flow.parser.parser` is collected after the lightweight PDF parser stub from `test/unit_test/rag/conftest.py` is installed.

`parser.py` still imports the following at module load time:

- `api.db.services.*` and `api.db.joint_services.*`;
- concrete `deepdoc.parser` implementations, including `PlainParser`, `RAGFlowPdfParser`, and `VisionParser`;
- `rag.flow.parser.pdf_chunk_metadata`.

The service imports re-enter `deepdoc.parser`, where the cached test stub does not expose `PlainParser`, causing a cascading import failure.

The test layout also remains ambiguous: `test/unit_test/rag/`, `rag/flow/tests/`, `test/unit_test/deepdoc/`, and `test/unit_test/deepdoc/parser/` have no `__init__.py`. Under pytest's default prepend import mode, these directories can become import roots, while the `test/unit_test/deepdoc` tree can contribute conflicting `deepdoc` / `deepdoc.parser` module state.

## Steps to reproduce

From the repository root:

```bash
uv run python -c 'import runpy; runpy.run_path("test/unit_test/rag/conftest.py"); import rag.flow.parser.parser'
```

Current result:

```text
ImportError: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser' (unknown location)
```

The traceback enters through `rag/flow/parser/parser.py -> api/db/services/file_service.py -> api/db/services/task_service.py -> deepdoc.parser` before reaching the direct parser imports.

## Expected behavior

Importing `rag.flow.parser.parser` should not fail because an unrelated test suite installed a deliberately partial PDF parser stub. Test directories should also resolve under stable package names instead of relying on transient `sys.path` roots or namespace-package state.

## Suggested scope / acceptance criteria

1. Move or reorganize `parser.py` dependencies so service and concrete parser implementations are imported only by the runtime paths that use them, or establish an equivalent owner-level boundary that avoids the eager `api.db -> deepdoc.parser` chain.
2. Normalize the affected test directories as packages (or adopt an equivalent pytest import configuration) so `test/unit_test/rag`, `rag/flow/tests`, and the unit-test `deepdoc` tree cannot poison production package entries in `sys.modules`.
3. Add a regression test for the reproduction above and keep the combined collection scenario from #17985 green.
4. Verify the parser runtime paths that are moved behind lazy boundaries, rather than testing import success alone.

## Relationship to #17985 / #18018

- #17985 records the original combined-suite collision.
- #18018 is intentionally a narrow stopgap for `TokenChunker` and links to this issue so the broader root cause is not lost after #17985 closes.

Harsh23Kashyap:
Fixes #18184

## Problem

The maintainer-requested follow-up from PR #18018 (#17985). The combined pytest collection run fails with:

```
ImportError: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser' (unknown location)
```

`test/unit_test/rag/conftest.py` installs a lightweight stub for `deepdoc.parser.pdf_parser` (which only exposes `RAGFlowPdfParser`) so `rag.nlp` tests can import without the full deepdoc/infinity chain. Pre-fix, three runtime modules under the chain `rag.flow.parser.parser -> rag.flow.parser.utils -> deepdoc.parser.figure_parser -> deepdoc/parser/__init__.py -> deepdoc.parser.pdf_parser` did eager `from deepdoc.parser... import X` at module load. The stub didn't have `PlainParser` or `VisionParser`, so the first import crashed and took the whole chain down with it.

The acceptance criteria from the issue:

1. Move or reorganize `parser.py` dependencies so service and concrete parser implementations are imported only by the runtime paths that use them, or establish an equivalent owner-level boundary that avoids the eager `api.db -> deepdoc.parser` chain.
2. Normalize the affected test directories as packages so `test/unit_test/rag`, `rag/flow/tests`, and the unit-test `deepdoc` tree cannot poison production package entries in `sys.modules`.
3. Add a regression test for the reproduction above and keep the combined collection scenario from #17985 green.
4. Verify the parser runtime paths that are moved behind lazy boundaries, rather than testing import success alone.

## Fix

Three source changes and one test-file change.

### `rag/flow/parser/parser.py` — module-level parser-backend map + lazy getter

Replaces nine eager `from deepdoc.parser... import X` lines (5 from `deepdoc.parser.*` plus `from rag.app.naive import Docx`) with a module-level `_lazy_deepdoc_imports` dict that names every parser backend, and a small `_lazy(name)` helper that calls `importlib.import_module` on demand. Call sites change `PlainParser()` to `_lazy("PlainParser")()` — same behaviour, but the import happens the first time the method actually needs the backend.

### `rag/flow/parser/utils.py` — lazy `VisionFigureParser` getter

Replaces the eager `from deepdoc.parser.figure_parser import VisionFigureParser` with a module-level `_get_vision_figure_parser()` helper that resolves the class on the first call. The eager import was the entry point for the `deepdoc/parser/__init__.py -> pdf_parser.PlainParser` chain that broke `utils.py` under the conftest stub.

### `api/db/services/task_service.py` — two thin helpers

The file had two eager `from deepdoc.parser... import X` lines even though it only used the two static-method calls `PdfParser.total_page_number(...)` and `RAGFlowExcelParser.row_number(...)`. The fix wraps each call in a small helper (`_pdf_parser_total_page_number`, `_ragflow_excel_parser_row_number`) that does the import at function-local scope. `task_service` is no longer in the `api.db -> deepdoc.parser` eager chain.

### Test directories → packages

Adds empty `__init__.py` files to `test/unit_test/rag`, `test/unit_test/deepdoc`, `test/unit_test/deepdoc/parser`, and `rag/flow/tests` so they are stable packages under pytest's import collection. Pre-fix they were namespace packages whose import root was determined by collection order — the test-package ambiguity the issue asked to fix.

## Tests

`test/unit_test/rag/test_parser_imports_isolate_from_pytest_stubs.py` — five AST-based regression tests that pin the post-fix contract on each affected module without re-importing the affected modules themselves (which would re-trigger the env-specific deepdoc chain the issue is normalising):

- `test_parser_module_no_eager_deepdoc_imports` — the nine parser backends must no longer appear as top-level `from deepdoc.parser... import X` bindings in `parser.py`.
- `test_parser_module_exposes_lazy_import_map` — `_lazy_deepdoc_imports` must be a module-level dict whose keys are exactly those nine backends.
- `test_utils_module_no_eager_vision_figure_parser_import` — `utils.py` must not eagerly import `VisionFigureParser`.
- `test_utils_module_exposes_lazy_getter` — `_get_vision_figure_parser()` must exist in `utils.py`.
- `test_task_service_no_eager_deepdoc_parser_imports` — `task_service.py` must not eagerly import `PdfParser` or `RAGFlowExcelParser`; the two helpers must exist.

5/5 pass. `ruff check` + `ruff format --check` clean on all five affected files plus the new test.

## Verification of the original reproducer

```
$ python -c "import runpy; runpy.run_path('test/unit_test/rag/conftest.py', run_name='__main__'); import rag.flow.parser.parser"
REPRODUCER PASSES  # post-fix

# Stashing the source changes (parser.py, utils.py, task_service.py) reverts the fix:
$ git stash push rag/flow/parser/parser.py rag/flow/parser/utils.py api/db/services/task_service.py
$ python -c "import runpy; runpy.run_path('test/unit_test/rag/conftest.py', run_name='__main__'); import rag.flow.parser.parser"
EXPECTED FAIL: cannot import name 'PlainParser' from 'deepdoc.parser.pdf_parser' (unknown location)
# Pre-fix — exact error from the issue
```

## consult-kiro note

Per the standing rule (2026-08-24), the local OpenCode Tier A panel was attempted. All four models (`opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, `opencode/deepseek-v4-flash-free`) returned empty answers / `rc=124` within the 180s hard-timeout window — same pattern as the prior 5 cycles. The change is small and the tests are the primary review surface; the consult is recorded here for completeness.

## Risk

- The lazy import map is a one-line addition per new parser backend. The runtime behaviour is unchanged — every method that previously did `PlainParser()` now does `_lazy("PlainParser")()`, which resolves to the same class. The only observable difference is the import happens the first time the method is called, not at module load.
- `task_service._pdf_parser_total_page_number` and `task_service._ragflow_excel_parser_row_number` are private helpers (`_` prefix). The pre-fix names `PdfParser.total_page_number` and `RAGFlowExcelParser.row_number` were static-method calls on the imported classes; the new helpers wrap them with the same call shape. No public API change.
- The `__init__.py` additions for the test directories are empty (0 bytes). They make those directories stable packages under pytest's import collection, so the package-vs-namespace-package ambiguity that pre-fix the chain through `test/unit_test/rag/__init__` and `test/unit_test/deepdoc/__init__` does not exist after the fix.
- Backward compatible: the runtime import path of the parser module is unchanged for normal operation. The fix only changes the failure mode of the conftest stub.

## Future work

- A few other modules under `rag/flow/parser` and `rag/flow/compiler` still eagerly import from `deepdoc.parser.*` — they were not in the failing import chain for this specific reproducer, but a future sweep would push the same lazy pattern to them. Out of scope for this PR because they did not regress the combined collection scenario.


xugangqiang:
Pls check CI error

PR #18710, opened 2026-08-24
fix(mcp): include dataset name in list_datasets output (#17133)

### LINKED ISSUE #17133: [Bug]: MCP ragflow_list_datasets never returns the dataset `name` field
### Self Checks

### RAGFlow image version

v0.26.4

```
- RAGFlow version: v0.26.4
- ragflow-mcp-server: 1.19.0 (from `serverInfo` on `initialize`)
- Deployment: Docker Compose, `--enable-mcpserver --mcp-mode=host`, doc engine = elasticsearch
```

### Actual behavior

Calling the `ragflow_list_datasets` MCP tool (and the equivalent `dataset_description` text embedded in the `ragflow_retrieval` / `ragflow_list_datasets` tool descriptions returned by `tools/list`) never includes the dataset's `name`, even though the dataset has a real, non-empty name.

Example `tools/call` for `ragflow_list_datasets`:

```json
{"description": null, "id": "46ea1644813211f1bc41abcb1d4727a4"}
```

The same dataset queried directly via the REST API (`GET /api/v1/datasets`) with the same API key returns the full record, including the name:

```json
{"id": "46ea1644813211f1bc41abcb1d4727a4", "name": "name", "description": null, "document_count": 8, "chunk_count": 845, ...}
```

So the dataset genuinely has a name ("name"); the MCP layer just never surfaces it. (The `description: null` in this particular example is expected/correct — that dataset was created without a description — but `name` is never returned for *any* dataset, regardless of whether it's set.)

### Expected behavior

`ragflow_list_datasets` (and the dataset description text embedded in the other MCP tools) should include each dataset's `name`, matching what the REST API returns and matching what `ragflow_list_chats` already does for chat assistants.

### Steps to reproduce

```
1. Deploy RAGFlow v0.26.4 with the MCP server enabled (--enable-mcpserver --mcp-mode=host or self-host).
2. Create/confirm a dataset with a non-empty name (e.g. "B3 Documentation").
3. Connect any MCP client and call the ragflow_list_datasets tool, or send a tools/list request.
4. Observe the returned dataset record only has "id" and "description" — "name" is absent.
5. Compare with GET /api/v1/datasets using the same api_key: the REST response includes "name" correctly.
```

### Root cause

In `mcp/server/server.py`, `RAGFlowConnector.list_datasets()` builds its output dict manually and simply omits `name`:

```python
result_list = []
for data in datasets:
    d = {"description": data["description"], "id": data["id"]}
    result_list.append(json.dumps(d, ensure_ascii=False))
return "\n".join(result_list)
```

Compare with `RAGFlowConnector.list_chats()` (same file, just above `list_datasets`), which was fixed for a related bug (#16638 / PR #16639) and correctly includes `name`:

```python
d = {"id": data.get("id"), "name": data.get("name"), "description": data.get("description", "")}
```

`list_datasets()` was apparently never updated to carry `name` through, even though the underlying `/api/v1/datasets` REST response (`data["name"]`) already has it available at that point in the function (it's fetched via `_fetch_datasets_page` / `_fetch_all_datasets`, whose `res_json["data"]` items already contain `name`).

This makes the tool far less useful for MCP clients (e.g. LLM agents) trying to decide which dataset to query by name — they only get an opaque `id` and (usually) a description, with no human-readable name at all.

### Proposed fix

In `list_datasets()`, include `name` (using `.get()` defensively, matching `list_chats()`'s style):

```python
d = {"id": data.get("id"), "name": data.get("name"), "description": data.get("description", "")}
```

Happy to submit a PR with this one-line change if useful.

Harsh23Kashyap:
Fixes #17133

## Problem

The MCP server's `ragflow_list_datasets` tool serialised each row as `{"description": "...", "id": "..."}` and dropped the `name` field that the REST API returns. The sibling `ragflow_list_chats` already includes `id` / `name` / `description` — the datasets branch was the inconsistency.

MCP clients can show a human-readable label next to the opaque dataset ID, but only if the label is in the response. Pre-fix the dataset name was lost in transit from the REST API to the MCP client, so users saw only "dataset-46ea1644..." with no way to distinguish multiple datasets by name. The dataset description text embedded in the sibling `ragflow_retrieval` / `ragflow_list_chats` tool descriptions (the `list_tools` output) was also affected — the bug is in the per-row dict shape, so the MCP tool description builder also dropped `name`.

Example from the issue:
```
$ MCP tool call: ragflow_list_datasets
{"description": null, "id": "46ea1644813211f1bc41abcb1d4727a4"}
# (no "name" field, even though the dataset genuinely has one)

$ REST API call: GET /api/v1/datasets
{"id": "46ea1644813211f1bc41abcb1d4727a4", "name": "B3 Documentation", ...}
```

## Fix

In `mcp/server/server.py::list_datasets`, reorder the per-row dict construction to `{"id": ..., "name": ..., "description": ...}` and use `data.get("name")` / `data.get("description")` so older API rows or partial records without a `name` still serialise with `name: null` (a stable shape, so the client can show a fallback label).

This matches the established pattern in `list_chats` (line 192), which builds `{"id": data.get("id"), "name": data.get("name"), "description": data.get("description", "")}`. Pre-fix the datasets branch had a different (broken) key set; post-fix the two paths agree.

## Tests

Two new regression tests in `test/unit_test/mcp/test_server_datasets.py` (extending the existing `_datasets` helper to accept a `with_name` flag):

- `test_list_datasets_includes_name_field` — three rows with `id` / `name` / `description` round-trip in document order; the per-row key set is exactly those three fields. **Pre-fix this fails with `row {'description': 'description-0', 'id': 'dataset-0'} is missing one of id/name/description`.**
- `test_list_datasets_name_omitted_when_missing` — when the REST API does not include `name` for a row (older dataset rows, partial records, migration backlog), the per-row dict still serialises with `name: null`. This pins the stable-shape contract so the client can show a fallback label and doesn't need to handle `KeyError`.

5/5 tests pass post-fix (the 3 pre-existing tests in the file still pass; the 2 new tests are the regression guards). `ruff check` + `ruff format --check` clean on both files.

## consult-kiro note

Per the standing rule (2026-08-24), the local OpenCode Tier A panel was attempted. All four models (`opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, `opencode/deepseek-v4-flash-free`) returned empty answers / `rc=124` within the 180s hard-timeout window — same pattern as the prior 6 cycles. The change is small (one line in `list_datasets` + the regression tests) and the diff is the primary review surface; the consult is recorded here for completeness.

## Risk

- Backward compatible for the only existing client path: the dict now has one more key (`name`) than before. Any client iterating over the dict's keys will see the new key, but no client was iterating over the keys (the dict was being serialised to a JSON string in `list_datasets` and consumed by an LLM that ignores unknown fields). The pre-fix `description: null` shape becomes `name: ..., description: null` — both null-shape keys are still there.
- Stable shape: `data.get("name")` returns `None` when the key is missing, so the per-row dict is `{"id": ..., "name": None, "description": ...}`. Old clients that didn't know about `name` continue to work; new clients that show the name get the new field. No client needs to change.
- No schema or response-format change for the underlying REST API. The MCP server is the only consumer of the fix; the REST API at `GET /api/v1/datasets` already returns `name`.

### DIFF
--- a/mcp/server/server.py
+++ b/mcp/server/server.py
@@ -242,7 +242,7 @@ class RAGFlowConnector:
 
         result_list = []
         for data in datasets:
-            d = {"description": data["description"], "id": data["id"]}
+            d = {"id": data["id"], "name": data.get("name"), "description": data.get("description")}
             result_list.append(json.dumps(d, ensure_ascii=False))
         return "\n".join(result_list)
 
--- a/test/unit_test/mcp/test_server_datasets.py
+++ b/test/unit_test/mcp/test_server_datasets.py
@@ -39,8 +39,14 @@ class _FakeResponse:
         return self._payload
 
 
-def _datasets(count):
-    return [{"id": f"dataset-{idx}", "description": f"description-{idx}"} for idx in range(count)]
+def _datasets(count, *, with_name=True):
+    out = []
+    for idx in range(count):
+        row = {"id": f"dataset-{idx}", "description": f"description-{idx}"}
+        if with_name:
+            row["name"] = f"name-{idx}"
+        out.append(row)
+    return out
 
 
 @pytest.fixture()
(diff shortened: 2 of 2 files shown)


6ba3i (approved):
LGTM!

PR #18711, opened 2026-08-24
fix(agent): prevent path traversal and SSRF in Browser URL file upload

simpleqt:
## What

Hardens the Browser component's URL-sourced upload path (`_prepare_upload_url_file`) against two issues reachable from untrusted input. `url` entries in an upload list can come from canvas variables filled by users or by the model, so they must be treated as untrusted.

### 1. Path traversal via crafted filename (CWE-22)

`_extract_url_filename` decoded percent-escapes **after** splitting path segments:

```python
raw_name = os.path.basename(parsed.path or "")
name = unquote(raw_name).strip()   # "%2e%2e%2f" is not a separator for basename(),
                                   # but decodes to "../" afterwards
```

A URL ending in `..%2F..%2Fconfig%2Fenv` (or a `Content-Disposition: filename="..%2f..%2fuploaded"` header from an attacker-controlled server) produced `local_name` containing separators, and `os.path.join(upload_dir, local_name)` wrote **outside the upload directory** — an arbitrary file write with the fetched body as content.

The same ordering bug applied to all three `Content-Disposition` branches.

**Fix:** `_safe_url_filename()` decodes first, takes `os.path.basename`, then validates the result is a single safe segment (rejects `.`, `..`, any separator); unusable candidates fall back to a UUID name that can never leave `upload_dir`.

### 2. SSRF via the URL fetch itself (CWE-918)

The old path used `urllib.request.urlopen`, which:

- performs **no SSRF validation** (the rest of the codebase routes fetches through `common/ssrf_guard`),
- follows redirects **automatically**, so even a validated-looking URL could bounce to `http://169.254.169.254/` or an internal service.

**Fix:** mirror `rag/app/naive.py` `load_images_from_urls`: `assert_url_is_safe()` + thread-local `pin_dns()` for the initial URL and **every manually followed hop** (capped at `MAX_UPLOAD_URL_REDIRECTS = 5`, relative `Location` resolved with `urljoin`), streaming via `requests` with the existing size cap, and the response released on all paths (`finally`).

## Tests

`test/unit_test/agent/component/test_browser_use_component.py` grows from 3 to 15 tests, covering:

- traversal payloads via URL path and all `Content-Disposition` variants (`filename*`, quoted, unquoted) stay inside `upload_dir`, UUID fallback when nothing safe can be derived
- each redirect hop is re-validated; a redirect into a private address is refused
- redirect cap enforcement, size-cap enforcement with partial-file cleanup, DNS pinning passed through, `finally`-close on error paths
- an xgboost stub (same pattern as the existing cv2 stub) so the suite stays collectable without libomp

All 15 pass locally.

## Notes

- Behavior for legitimate filenames is unchanged (dedup suffix `_1`, size limit, error logging preserved).
- Chose `pin_dns` (thread-local) over `pin_dns_global` because this is a synchronous `requests.get` resolving DNS in the calling thread, matching the naive.py precedent.

### DIFF
--- a/agent/component/browser.py
+++ b/agent/component/browser.py
@@ -26,9 +26,7 @@ import tempfile
 from abc import ABC
 from pathlib import Path
 from typing import Any
-from urllib.error import HTTPError, URLError
-from urllib.parse import unquote, urlparse
-from urllib.request import Request, urlopen
+from urllib.parse import unquote, urljoin, urlparse
 
 from agent.component.base import ComponentBase
 from agent.component.llm import LLMParam
@@ -41,6 +39,10 @@ from common import settings
 from common.connection_utils import timeout
 from common.misc_utils import get_uuid
 from rag.llm import FACTORY_DEFAULT_BASE_URL
+
+# Mirrors MAX_IMAGE_REDIRECTS in rag/app/naive.py: cap the manually followed
+# redirect hops when fetching URL-sourced upload files.
+MAX_UPLOAD_URL_REDIRECTS = 5
 
 
 class BrowserParam(LLMParam):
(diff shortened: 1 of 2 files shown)


6ba3i (changes requested):
1. **SSRF / DNS pinning can be bypassed through environment proxies**

In `agent/component/browser.py`, around the new fetch loop:

```python
with pin_dns(current_hostname, current_ip):
    response = requests.get(
        current_url,
        stream=True,
        timeout=30,
        allow_redirects=False,
        ...
    )
```

`pin_dns()` only pins resolution through the local `socket.getaddrinfo()`. However, `requests.get()` uses a default `Session` with `trust_env=True`.

If `HTTP_PROXY` / `HTTPS_PROXY` is configured, Requests connects to the proxy and the proxy may resolve `current_url` itself. In that case, the validated `current_ip` is no longer guaranteed to be the actual destination, which weakens the SSRF protection this PR is trying to provide.

The same default behavior may also load `.netrc` credentials for an untrusted URL.

I would use an isolated session here:

```python
session = requests.Session()
session.trust_env = False

try:
    response = session.get(...)
finally:
    session.close()
```

If proxy support is intentionally required, it needs an explicit proxy-safe SSRF design rather than relying on the default Requests environment behavior.

2. **The added tests do not cover all security-sensitive behavior claimed by the PR**

In `test/unit_test/agent/component/test_browser_use_component.py`, I would add explicit tests for:

* redirect limit boundary: 5 redirects succeed, 6th is rejected;
* partial-file cleanup after size-limit or `iter_content()` failure;
* final response being closed;
* DNS pin values actually reaching the request path;
* quoted, unquoted and `filename*` `Content-Disposition` traversal cases.

The fake response already exposes `.closed`, but currently nothing asserts it.

3. **I would remove the new `requests` import stub**

The test file now replaces Requests with a fake module if the import fails. Since Requests is already a declared dependency, and this PR specifically relies on its runtime/session behavior, masking import failures can hide real environment problems.

The `requests` stub is particularly risky here because all tests monkeypatch `requests.get`, meaning they cannot detect the `trust_env` / proxy behavior above.

simpleqt:
@6ba3i Excellent catches — all three addressed in e96ab11 (20/20 tests pass locally):

**1. Proxy/netrc bypass.** The fetch now builds an isolated `requests.Session()` with `trust_env = False` per call (no HTTP_PROXY/HTTPS_PROXY/ALL_PROXY, no `.netrc` for the untrusted URL), drives every hop through it, and `finally` closes both the final response and the session. No proxy support is intentionally provided on this path — ambient proxies are exactly the bypass described; a deployment that needs one will require the explicit proxy-safe design you outlined. Worth noting `rag/app/naive.py load_images_from_urls` shares the same `requests.get` weakness — happy to port the isolated-session pattern there in a follow-up.

**2. Test coverage.** Added: redirect-limit boundary (5 hops succeed, 6th refused); partial-file cleanup on size-limit breach and on `iter_content` raising `RequestException` mid-stream; explicit `.closed` assertions for the session and every hop's response (intermediate redirects included); per-hop DNS pin capture asserting the pinned (host, ip) is active at the moment each request fires and the pin stack unwinds to empty; and all four Content-Disposition traversal forms (filename\*, quoted, unquoted, percent-mixed). The quoted backslash case also drove a fix: `_safe_url_filename` now normalizes backslashes before basename, so Windows-style traversal splits correctly while legitimate names survive.

**3. Stub removed.** The requests fallback stub is gone — tests exercise the real requests import and fake only `requests.Session` construction, so a missing dependency fails loudly, and `trust_env is False` is asserted rather than masked.

6ba3i (approved):
LGTM

PR #18712, opened 2026-08-24
Fix email and presentation parsers treating an empty upload as a missing binary

devYRPauli:

An upload of 0 bytes reaches the chunkers as `b""`, not as `None`.

`rag/svr/task_executor.py:312-313` reads the object and guards only against `None`:

```python
binary = await get_storage_binary(bucket, name)
if binary is None:
    raise FileNotFoundError(...)
```

`b""` is not `None`, so the guard passes. Line 361-363 then calls
`chunker.chunk(task["name"], binary=binary, ...)`. `task["name"]` is the document display
name, not a path on disk.

The email and presentation parsers chose between "use the bytes" and "open the path" with a
truthiness test. `b""` is falsy, so they took the path branch and opened the display name as
a file. The upload failed with `FileNotFoundError` on a name that never existed on disk.

PR #17904 fixed the same defect in `deepdoc/parser/html_parser.py` (commit `9ea83b7a`,
`if binary:` to `if binary is not None:`). This PR applies that fix to the remaining
reachable sites in two more files.

### Sites changed

| file | line | before | after |
|---|---|---|---|
| `rag/app/email.py` | 54 | `if binary:` | `if binary is not None:` |
| `rag/app/presentation.py` | 43 | `filename if not binary else binary` | `filename if binary is None else binary` |
| `rag/app/presentation.py` | 121 | `filename if not binary else BytesIO(binary)` | `filename if binary is None else BytesIO(binary)` |
| `rag/app/presentation.py` | 144 | `filename if not binary else binary` | `filename if binary is None else binary` |
| `rag/app/presentation.py` | 168 | `if binary:` | `if binary is not None:` |

All five are reachable from `chunk()`. A `.pptx` upload reaches line 144 and then the tika
fallback at 168. A `.pdf` upload reaches `Pdf.__call__` at 43 or `PlainPdf.__call__` at 121.

Behavior for a non-empty binary and for `binary=None` is unchanged. Only the `b""` case
moves, and it moves from a crash to normal parsing of an empty payload.

### Type of change

### Test

New test `test/unit_test/rag/app/test_email_empty_binary.py`. It calls the real
`rag.app.email.chunk` with `binary=b""` and a display name that is not a file on disk.

With the fix:

```
$ .venv/bin/pytest test/unit_test/rag/app/test_email_empty_binary.py -q --noconftest
1 passed in 7.17s
```

With `rag/app/email.py` reverted and the test kept:

```
rag/app/email.py:58: in chunk
    with open(filename, "rb") as buffer:
E   FileNotFoundError: [Errno 2] No such file or directory: 'message.eml'
1 failed in 6.44s
```

The failure line is the path branch this PR removes, so the test is not vacuous.

Sibling test still passes:

```
$ .venv/bin/pytest test/unit_test/rag/app/test_qa_csv.py -q --noconftest
4 passed
```

`ruff check` on both changed files and the new test: All checks passed.

I did not add a test for `presentation.py`. That path needs python-pptx or tika, which is
larger than this fix.

Note on `--noconftest`: `test/unit_test/rag/conftest.py` stubs `deepdoc.parser.pdf_parser`
and breaks collection for several files in that directory with
`ImportError: cannot import name 'PlainParser'`. This is pre-existing and not changed here.

### Out of scope

The same truthiness pattern remains in `deepdoc/parser/mineru_parser.py`,
`somark_parser.py`, `tcadp_parser.py` and `rag/app/naive.py`. I did not trace whether an
empty upload reaches those, so I left them alone.

### DIFF
--- a/rag/app/email.py
+++ b/rag/app/email.py
@@ -51,7 +51,7 @@ def chunk(
     main_res = []
     attachment_res = []
 
-    if binary:
+    if binary is not None:
         with io.BytesIO(binary) as buffer:
             msg = BytesParser(policy=policy.default).parse(buffer)
     else:
--- a/rag/app/presentation.py
+++ b/rag/app/presentation.py
@@ -40,7 +40,7 @@ class Pdf(PdfParser):
     def __call__(self, filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, zoomin=3, callback=None, **kwargs):
         # 1. OCR
         callback(msg="OCR started")
-        self.__images__(filename if not binary else binary, zoomin, from_page, to_page, callback)
+        self.__images__(filename if binary is None else binary, zoomin, from_page, to_page, callback)
 
         # 2. Layout Analysis
         callback(msg="Layout Analysis")
@@ -118,7 +118,7 @@ class Pdf(PdfParser):
 
 class PlainPdf(PlainParser):
     def __call__(self, filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, callback=None, **kwargs):
-        self.pdf = pdf2_read(filename if not binary else BytesIO(binary))
+        self.pdf = pdf2_read(filename if binary is None else BytesIO(binary))
         page_txt = []
         for page in self.pdf.pages[from_page:to_page]:
             page_txt.append(page.extract_text())
@@ -141,7 +141,7 @@ def chunk(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, lang=
     if re.search(r"\.pptx?$", filename, re.IGNORECASE):
         try:
             ppt_parser = RAGFlowPptParser()
-            for pn, txt in enumerate(ppt_parser(filename if not binary else binary, from_page, MAXIMUM_PAGE_NUMBER, callback)):
+            for pn, txt in enumerate(ppt_parser(filename if binary is None else binary, from_page, MAXIMUM_PAGE_NUMBER, callback)):
                 d = copy.deepcopy(doc)
                 pn += from_page
                 d["doc_type_kwd"] = "image"
@@ -165,7 +165,7 @@ def chunk(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, lang=
                 logging.warning(f"{error_msg} for {filename}.")
                 raise NotImplementedError(error_msg)
 
-            if binary:
+            if binary is not None:
                 binary_data = binary
             else:
                 with open(filename, "rb") as f:
--- a/test/unit_test/rag/app/test_email_empty_binary.py
+++ b/test/unit_test/rag/app/test_email_empty_binary.py
@@ -1,0 +1,35 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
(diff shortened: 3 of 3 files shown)


xugangqiang:
LGTM @devYRPauli  would you please help to check if such issue exist in other production path?

xugangqiang:
@yuzhichang  Good to merge.

Harsh23Kashyap:
Submitted PR #18759 — the four files you explicitly listed as out-of-scope (`deepdoc/parser/mineru_parser.py`, `deepdoc/parser/somark_parser.py`, `deepdoc/parser/tcadp_parser.py`, `rag/app/naive.py`).

The fix is the same shape: `if binary:` -> `if binary is not None:`, `if not binary:` -> `if binary is None:`, the matching ternaries in `naive.py` and `tcadp_parser.py` get the explicit `is None` form. Behaviour for non-empty `binary` and for `binary=None` is unchanged. Only the `b""` case moves, and it moves from a crash to a normal empty parse (or a clear `bad binary` return for the deepdoc parsers that the reporter will see in the logs).

9 sites total (the AST walker covers them all). `ruff check` + `ruff format --check` clean.

If you'd prefer this folded into your PR as a follow-up commit (rather than a separate PR), I can close #18759 and rebase the same change onto your branch — just say so.

devYRPauli:
A separate PR was the right call. #18712 merged at 05:59 UTC on 2026-08-25, about seven hours before your comment, so there was no branch left to fold the change into.

The reason those four files were out of scope is in the PR body: I did not trace whether an empty upload reaches them, so I left them alone. #18759 covers them with the same `is None` shape.

Sorry for the slow reply.

JessYanCoding:
@xugangqiang — following up on your question here about whether the same issue exists in other production paths.

It does. #19256 covers the remaining `rag/app` chunkers: 13 sites in `book`, `laws`, `manual`, `one`, `paper`, `qa` and `table`, all reachable from `chunk()` and all registered in `task_executor.FACTORY`. Same `binary is None` shape as this PR.

Tests are functional rather than AST-based, and each asserts which branch was taken — the two gaps you raised on #18759. Each drives a real entry point with `binary=b""` and a display name that is not on disk, then asserts the parser received `b""` instead of the name. Reverting the 13 lines fails all 13.

Also checked and left out, with reasons in the PR body: `epub_parser.py` already rejects an empty payload on purpose, `pdf_parser.total_page_number` degrades at all three of its call sites rather than crashing, and the `mineru`/`somark`/`tcadp` family is #18759's scope.

devYRPauli:
@xugangqiang here is the full accounting for the question you asked on 2026-08-25. Measured against `main` at `fd9f9e7f6`.

23 sites test `binary` for truth instead of against `None`. All 23 are accounted for.

| where | sites | status |
|---|---|---|
| `rag/app`: `book`, `laws`, `manual`, `one`, `paper`, `qa`, `table` | 14 | #19256, open |
| `deepdoc/parser`: `mineru_parser`, `somark_parser`, `tcadp_parser` | 5 | #18759, open |
| `deepdoc/parser/epub_parser.py:41` | 1 | correct as written |
| `deepdoc/parser/pdf_parser.py:1605` | 1 | degrades, does not crash |
| `agent_api.py:1154`, `bot_api.py:301` | 2 | not a parser input |

`epub_parser.py:41` sits inside `if binary is not None:` and raises `ValueError("Empty EPUB binary payload")` on purpose.

`pdf_parser.total_page_number` opens the display name when `binary` is `b""`. The `except Exception` catches that and the function returns `None`. The consumer at `api/db/services/task_service.py:482` already reads `if pages is None: pages = 0`.

The two `restful_apis` sites read an agent log out of Redis. An empty value there means no logs, and `return get_json_result(data={})` is the right answer.

`rag/app/naive.py` is already fixed on `main`. Line 169 reads `filename if binary is None else binary`. #18759 still lists it in scope, so that part of the description is stale.

Two notes on the open PRs.

#19256 changes 13 lines, not 14. The fourteenth site is `laws.py:44`, inside `Docx.old_call`, and the PR deletes that method and the `__clean` helper it calls instead of fixing them. That is safe. At `fd9f9e7f6` the string `old_call` appears exactly once in the whole tree, on its own def line, and `__clean` is called only from inside it. It is still unrelated cleanup inside a behaviour fix.

#18759 lists 30 changed files, including `README.md`, `Dockerfile` and both `CLAUDE.md` copies. @Harsh23Kashyap that is a stale merge base, not a real diff. A rebase on `main` rebuilds the PR record. I hit the same thing on #19178 this week, where 5056 files became 2 after the rebase.

PR #18719, opened 2026-08-25
Fix: Add a custom placeholder to SelectWithSearch

cike8899:
### Summary
Fix: Add a custom placeholder to SelectWithSearch


wangq8 (commented):
This is an AI review comment.

Small, low-risk change. A few notes:

**1. New i18n key is only added to `en.ts` and `zh.ts`.**
`common.searchOrEnterToAdd` (and the `selectFieldType` text change) are added only to the English and Chinese locale files, but the repo ships ~15 other locales (`web/src/locales/{ar,bg,de,es,fr,id,it,ja,ko,pt-br,ru,tr,vi,zh-traditional}.ts`). Those locales will fall back to English (or show the key) for `allowCustomValue` inputs. If this follows the project's "translate later" convention, fine — just flagging.

**2. Title vs. actual behavior.**
The PR is titled "Add a custom placeholder," but the change hardcodes the search-box placeholder to `common.searchOrEnterToAdd` when `allowCustomValue` is true rather than adding/using a `placeholder` prop. The component already accepts a `placeholder` prop (used only for the trigger button, not the `CommandInput`). Worth clarifying the intent, and consider wiring the existing `placeholder` prop through to the search input so callers can override it.

**3. Unrelated string change.**
`selectFieldType` ("Select field type" → "Select or enter a field type") lives under a different namespace and isn't tied to the `SelectWithSearch` placeholder logic. If it's meant to accompany a field-type selector that enables `allowCustomValue`, that's fine, but it's a separate concern from the title.

This is an AI review comment.

Issue #18665, opened 2026-08-23
[Bug]: Message node drops the whitespace around every {Component@output} reference (variable_ref_patt swallows adjacent spaces)


artiomcom:
### Self Checks

- [x] I have searched for existing issues, including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] Non-english title submissions will be closed directly.
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

2db8eb6c (`main`, 2026-08-23). Same code in the v0.27.0 release sources.

### RAGFlow image version

Reproduced against the sources (unit test below), not a specific image; the affected lines are identical in `main` and v0.27.0.

### Other environment information

- Python 3.12, Windows 11 host for the unit test; behaviour is pure-Python (`re`), platform-independent.

### Actual behavior

When a `Message` node streams its content (the default path: `stream=True` and the content contains no `{{`/`{%`), the literal whitespace **immediately before and after every `{Component@output}` / `{sys.x}` reference is dropped** from the rendered text.

Content:

```
Decision: {Agent:x@content} — done
```

Rendered: `Decision:VALUE— done`

Two adjacent references `{A@x} {B@y}` render as `VALUE1VALUE2` (the separating space is lost too).

Root cause is `ComponentBase.variable_ref_patt` in `agent/component/base.py`:

```python
variable_ref_patt = r"\{* *\{([a-zA-Z0-9_:]+@[A-Za-z0-9_.-]+|sys\.[A-Za-z0-9_.]+|env\.[A-Za-z0-9_.]+)\} *\}*"
```

The ` *` on each side exists to tolerate the `{ {X@y} }` / `{{X@y}}` forms (which `Canvas.get_variable_value` strips in layers). But `\{*` and `\}*` also match the empty string, so with plain single braces the free-standing ` *` consumes the neighbouring literal spaces:

```python
>>> import re
>>> [m.group(0) for m in re.finditer(variable_ref_patt, "a {X@y} b")]
[' {X@y} ']          # expected: ['{X@y}']
```

`Message._stream` (`agent/component/message.py`) copies the literal text by slicing around the match —

```python
all_content += rand_cnt[s : r.start()]
yield rand_cnt[s : r.start()]
s = r.end()
```

— so the swallowed spaces are never yielded. `iteration_alias_patt` (`\{* *\{(item|index|result)\} *\}*`) has the same shape and the same problem.

[...]

PR #18720, opened 2026-08-25
test(agent): pin whitespace handling of template ref patterns

ayam04:
Refs #18665

## What this is

The whitespace-swallowing defect reported in #18665 is already fixed on `main` — `a1131167` (PR #18666) replaced the hand-written `{* *{...} *}*` patterns with `_build_template_ref_pattern`, which only consumes whitespace inside a balanced outer-brace pair. I verified the fix against every case in the issue before writing this PR (plain refs keep adjacent spaces, `{A@x} {B@y}` keeps its separating space, `{ {X@y} }` / `{{X@y}}` / `{ {X@y}}` still match whole, sys./env. alternates, and the `iteration_alias_patt` twin).

What the fix landed *without* is regression tests, so the issue's Expected behavior is still unpinned. This PR adds them: 6 tests in `test/unit_test/agent/test_variable_ref_whitespace.py` using the repo's existing import-isolation pattern (`test_canvas_at_split.py`), no live DB/Redis/MinIO stack needed.

One test asserts the user-visible contract from the issue end to end: emulating `Message._stream`'s slice-around-the-match rendering, `Decision: {Agent:x@content} — done` with value `VALUE` renders `Decision: VALUE — done`.

## Validation

```bash
python3 -m pytest test/unit_test/agent/test_variable_ref_whitespace.py -q   # 6 passed
ruff check test/unit_test/agent/test_variable_ref_whitespace.py             # clean
ruff format --check test/unit_test/agent/test_variable_ref_whitespace.py    # clean
```

(Files run against the repo root's pyproject, line-length 200.) Could not run the full backend suite locally — it needs the docker stack per AGENTS.md; the tests above are self-contained pure-regex isolation tests.


6ba3i (changes requested):
### 1. The “streaming” regression test does not exercise the real `Message._stream`

`_render_with_stream_slicing()` reimplements only the slicing behavior using `variable_ref_patt_re.finditer()` and a dictionary.

Production `Message._stream` instead:

* compiles `self.variable_ref_patt`,
* goes through `_iter_template_matches`,
* resolves values through the canvas,
* handles `None`, partials, awaitables, stringification, cancellation, and caching.

That means this test can stay green while the actual streaming path regresses, for example if `_iter_template_matches` stops filtering malformed references or `Message._stream` changes how it compiles/iterates the pattern.

Please either test the real async `Message._stream` with a minimal isolated fixture, or remove the simulated “end-to-end streaming” test and keep this scoped to regex/helper behavior.

### 2. Most of this coverage already exists from #18666

The PR description says #18666 fixed the bug without regression tests, but that is not the case.

`test_variable_ref_pattern_unit.py` already covers:

* `Decision: {Agent:x@content} — done`
* `{A@x} {B@y}`
* `{{X@y}}`
* `{ {X@y} }`
* malformed outer braces
* iteration alias replacement
* exact matched text for the whitespace-sensitive cases

The real value of this PR is that `test/unit_test` is part of the normal unit-test preflight, while the existing suite is under `test/testcases`.

Rather than adding a second 169-line import-stubbing harness with overlapping assertions, I would consolidate the existing self-contained suite into `test/unit_test/agent` and add only the genuinely missing cases there (`sys/env`, `{ {X@y}}`, `result`, etc.).

### 3. Wrapped-reference assertions do not verify the complete match

For cases such as:

```python
"{ {X@y} }"
"{{X@y}}"
"{ {X@y}}"
```

the test checks only the match count and `group(1)`.

A regression that matches only the inner `{X@y}` would still pass.

The same applies to `{ {result} }`.

Since the contract being tested is that the complete wrapped expression is consumed, please assert:

```python
match.group(0) == text
```

or use `fullmatch()`.

ayam04:
Thanks, all three points addressed in bad41b0:

**1. The simulated streaming test is gone.** `_render_with_stream_slicing` and its helper (and the 169-line parallel harness) are deleted; the suite is now scoped to pattern/helper behavior exactly as suggested.

**2. Consolidated instead of duplicated.** `test_variable_ref_pattern_unit.py` moved from `test/testcases/test_web_api/test_canvas_app` into `test/unit_test/agent` (git rename, history preserved) so it runs in the normal unit-test preflight, and the genuinely missing cases were added to `test_reference_patterns_preserve_adjacent_literal_whitespace`: the asymmetric wrap `{ {X@y}}`, `sys.*` refs with adjacent literal spaces, and the wrapped iteration alias `{ {result} }` (previously only bare `{item}` was pinned).

**3. Complete-match assertions.** Every row already asserted `group(0)`; the new rows continue that, so a regression matching only the inner `{X@y}` fails loudly.

Validation: `python3 -m pytest test/unit_test/agent/test_variable_ref_pattern_unit.py` → 23 passed; `ruff check` + `ruff format --check` clean on the moved file. (Full backend suite needs the docker stack per AGENTS.md; not run here.)

6ba3i (approved):
LGTM

PR #18721, opened 2026-08-25
fix: support hierarchical Wiki topic navigation

buua436:

Wiki topics with hierarchical paths were displayed as a flat list, and the topic API could incorrectly use a page title as the topic title.

This change:

- Uses the topic path as the canonical topic identifier.
- Returns the topic leaf as `title` and the full path as `slug`.
- Displays topics level by level in the Wiki navigation.
- Loads Wiki pages only after the leaf topic is selected.
- Supports multi-level breadcrumbs.
- Preserves compatibility with single-level topics from the Python implementation.

### Type of change


buua436:
07fbb658-9222-4fa1-876b-c4e9fdb00258
c85f4cd0-5034-41fd-be87-bcd894a74472
5f375f7a-ee08-41e7-913b-4671dd52ecf9
da0aa89e-03dc-4e18-b944-32e3d701af75

wangq8 (commented):
This is an AI review comment.

Thanks for the PR — the hierarchical topic-path model is a clear improvement over the flat list. A few things to double-check before merge:

**1. Removal of the deterministic Mode A path is a significant behavior/cost change (not called out).**
`run()` now unconditionally calls `runPlan()`, and `buildModeAPlan()` is deleted. In entity mode (`PlanEnabled() == false`, the default), the pipeline previously built a deterministic flat plan with **no LLM call and no reconcile**. After this change, entity mode goes through `runLegacyPlan()` → `runPlanBatch()` → `GenJSON`, i.e. it now spends an LLM PLAN call and becomes non-deterministic for the default path. The PR description says it "preserves compatibility with single-level topics," but does not mention that the default path now incurs an extra LLM call. If this is intended (to route entity-mode pages through topic assignment), please state it explicitly; otherwise consider keeping a deterministic topic-assignment path for entity mode.

**2. Unrelated change in `base_model.go`.**
`NewDriverHTTPClient` `ResponseHeaderTimeout` is raised 2min → 5min. This is a generic driver HTTP client (object storage, etc.), not specific to wiki compilation, and has nothing to do with topic navigation. Raising a global header timeout can mask slow/hung endpoints across the whole driver surface. Please split this into its own PR with its own justification.

**3. MAP prompt field mismatch: `source_chunk_id` (singular) vs `source_chunk_ids` (plural).**
The prompt schema tells the model to emit `"source_chunk_id": "string"`, but the Go struct and merge logic use `source_chunk_ids` (array); `parseWikiTopics` silently accepts both. Prefer aligning the prompt to the structured `source_chunk_ids` field so we don't depend on the singular-string fallback path.

**4. Intermediate topic nodes with their own pages become unreachable in the nav.**
`showArtifacts = Boolean(selectedTopicPath && visibleTopics.length === 0)`. A non-leaf path that has both children *and* pages assigned directly to it will only ever show its children — its own pages can't be opened. Consider surfacing pages for intermediate nodes (e.g. a "view pages" affordance), or confirm that pages are only ever assigned to leaf paths.

**5. `selectWikiTopicPath` self-topic rejection may be over-aggressive.**
`selfTopic` rejects any entity/concept candidate whose leaf equals the page title, so a legitimately materialized path like `人物/曹操` for page "曹操" is discarded and the page falls back to `General`. This matches the stated prompt rule, but it will also throw away real hierarchical topics whose leaf happens to equal the title. Please confirm this is the intended trade-off.

**6. Frontend/Go normalization asymmetry.**
Go collapses depth > 3 to `a/b/c · d`, but the frontend `normalizeTopicPath`/`topicLeaf` only split on `/`. In practice ingestion caps at 3 segments so this is mostly moot, but the `· ` (with surrounding spaces) is a latent mismatch if depth is ever exceeded — worth a shared rule/comment.

**7. `wikiTemplateCustomRules` precedence change.**
`global_rules` now overrides `guideline` where previously `guideline` won. Existing templates that set both will silently change output. Flagging so the behavior change is intentional and documented.

Tests are good and cover the normalization, dedup, and merge-selection logic well. Nits aside, the structure is solid.

This is an AI review comment.

PR #18722, opened 2026-08-25
fix(setting-model): stop silently dropping a duplicate-named provider  instance

dcc123456:
### Summary

fix(setting-model): stop silently dropping a duplicate-named provider  instance

Adding an instance whose name matched an already-saved one never reached the server: useAddProviderInstance pre-listed the instances and returned a fake `{ code: 0 }` on a name collision, so the POST was skipped while the page treated the save as successful and cleared the draft card. The new instance vanished with no feedback.

Drop the pre-flight check and reject the collision up front in the batch save handler instead, so the user sees "instance name already exists" and no request is issued. Living in the parent it also covers Bedrock cards and two drafts sharing a name within one batch.

### DIFF
--- a/web/src/hooks/use-llm-request.tsx
+++ b/web/src/hooks/use-llm-request.tsx
@@ -299,19 +299,8 @@ export const useAddProviderInstance = () => {
     ) => {
       try {
         await addProvider({ provider_name: params.llm_factory });
-
-        const { data: instancesRes } = await llmService.listProviderInstances(
-          { provider_name: params.llm_factory },
-          true,
-        );
-        const instanceExists = instancesRes?.data?.some(
-          (i: IProviderInstance) => i.instance_name === params.instance_name,
-        );
-        if (instanceExists && !params.verify) {
-          return { code: 0, data: null };
-        }
       } catch {
-        // ignore list failure and proceed to add
+        // ignore and proceed to add
       }
 
       // The provider is carried in the URL path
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
Thanks for the fix — this correctly addresses the silent-drop bug. Verified the change against `ragflow-main` and the flow is sound.

**What works well**
- The batch-level guard seeds `takenNames` from persisted instances and accumulates draft names, so it catches both "draft reuses a persisted name" and "two drafts share a name in one batch".
- Removing the pre-flight list-then-return-fake-`{code: 0}` path means the POST is no longer silently skipped, and the new `message.error` finally gives the user feedback instead of the instance just vanishing.
- `useCallback` deps updated correctly (`instances`, `tSetting`).

**Minor nits (non-blocking)**
1. Comment accuracy: the backend (`TenantModelInstanceService.create_instance` → `duplicate_name`) does not error or create a second instance with the *same* name — it silently auto-renames the duplicate (e.g. `foo(1)`). The guard is still correct, but the comment "either error or silently create a second instance sharing the name" slightly misstates backend behavior. Consider rewording to "silently create a duplicate (auto-renamed)".
2. The draft name is `.trim()`ed but persisted `instance_name` values are compared untrimmed; if a stored name ever carried leading/trailing whitespace the collision wouldn't match. Low risk since the form trims on input, just flagging the asymmetry.
3. No test covers the new duplicate-rejection path. A small unit test for `handleSaveAll` (persisted-name collision + same-batch draft collision) would guard against regressions.
4. `instanceNameExists` is only added to `en.ts` and `zh.ts`; the other ~15 locale files will fall back to English. Fine if that's the repo convention, just flagging for completeness.

LGTM otherwise.

PR #18725, opened 2026-08-25
fix(web): stop saving fractional values for agent integer parameters

euvre:
### Summary

The Agent component's advanced settings let fractional values be saved for parameters the backend requires as integers. As reported in the Feishu group ("智能体 agent 高级设置 保存了小数参数"), a saved Agent node showed 历史消息窗口大小 (message window size) = 122.1, 最大重试轮数 (max retry rounds) = 31.1, 错误后延迟 (delay after error) = 14.8 and 最大反思轮数 (max reflect rounds) = 11.1.

**Root cause**

- `message_history_window_size` was rendered with the originui `NumberInput`, which propagates any value inside `[min, max]` — and this field has no `max`, so every decimal (e.g. `122.1`) was written to the canvas DSL and persisted on save.
- `max_retries` / `delay_after_error` / `max_rounds` were rendered with the `ui/input` `NumberInput`, a thin wrapper that propagates raw `Number(value)` with no clamping or integer enforcement at all (`min`/`max` only land as inert DOM attributes), which is how out-of-range values such as `31.1` and `14.8` got saved by builds before the bounds existed.
- Nothing downstream corrected this: the zod schema uses plain `z.coerce.number()` and `useWatchFormChange` persists whatever the form holds.
- The Python runtime already rejects these values at run time (`agent/component/llm.py` validates `max_retries` / `max_rounds` as non-negative integers, added in #18051), so a canvas saved with decimals fails when run. This PR prevents the invalid data from being saved in the first place, which is the complementary frontend half of #18051 rather than a resubmission of it.

**Fix**

- `web/src/components/originui/number-input.tsx`: new opt-in `integer` mode — fractional input is rounded before it is propagated to the form (typing `122.1` saves `122`), out-of-range values clamp on blur (`31.1` → `8` with `max=8`), steppers step from a rounded base, and a fractional/out-of-range legacy value is healed (rounded + clamped and propagated) on mount.
- `web/src/components/message-history-window-size-item.tsx`: the shared window-size field (used by the Agent, Categorize and Rewrite-question forms) is now `integer`.
- `web/src/pages/agent/form/agent-form/index.tsx`: the advanced-settings number fields now use the clamping originui `NumberInput`; `max_retries` and `max_rounds` are `integer`. `delay_after_error` intentionally keeps fractional support (delay in seconds, step 0.1, backend default 2.0) with its 0–5 range enforced.
- `web/src/pages/agent/form/agent-form/use-values.ts`: legacy fractional values of the integer fields are rounded when node data is loaded, so already-saved canvases display and re-save as integers.

**Verification**

- Reproduced before the fix in a local canvas: entering `122.1` in the agent's message window size and saving persisted `message_history_window_size: 122.1` (and a previous session had persisted `max_retries: 2.5`) in the saved DSL.
- After the fix (browser, same canvas): legacy `122.1` / `2.5` are displayed healed as `122` / `3`; entering `122.9` propagates `123`; entering `31.1` in max retry rounds clamps to `8` on blur; after saving, the server DSL contains only integers (`123 / 8 / 1 / 1`).
- New unit tests: `web/src/components/originui/__tests__/number-input.test.tsx` (7 cases: rounding on typing, blur finalization, clamping, legacy heal, stepper behavior, fractional passthrough when integer mode is off) — all pass. `npx oxlint` clean on all changed files; `tsc --noEmit` reports no new errors in changed files.

### DIFF
--- a/web/src/components/message-history-window-size-item.tsx
+++ b/web/src/components/message-history-window-size-item.tsx
@@ -44,7 +44,12 @@ export function MessageHistoryWindowSizeFormField({
             {t('flow.messageHistoryWindowSize')}
           </FormLabel>
           <FormControl>
-            <NumberInput {...field} min={min} className="w-full"></NumberInput>
+            <NumberInput
+              {...field}
+              min={min}
+              integer
+              className="w-full"
+            ></NumberInput>
           </FormControl>
           <FormMessage />
         </FormItem>
--- a/web/src/components/originui/__tests__/number-input.test.tsx
+++ b/web/src/components/originui/__tests__/number-input.test.tsx
@@ -1,0 +1,114 @@
+import { fireEvent, render } from '@testing-library/react';
+import NumberInput from '../number-input';
+
+function getNumberInput(container: HTMLElement) {
+  return container.querySelector('input[type="number"]') as HTMLInputElement;
+}
+
+function getStepButtons(container: HTMLElement) {
+  const buttons = Array.from(
+    container.querySelectorAll('button'),
+  ) as HTMLButtonElement[];
+  return { decrement: buttons[0], increment: buttons[1] };
+}
+
+describe('NumberInput', () => {
+  describe('integer mode', () => {
+    it('propagates the rounded integer while a fractional value is typed', () => {
+      const onChange = jest.fn();
+      const { container } = render(
+        <NumberInput value={12} integer onChange={onChange} />,
+      );
+      const input = getNumberInput(container);
+
+      fireEvent.change(input, { target: { value: '122.1' } });
+
+      expect(input.value).toBe('122.1');
+      expect(onChange).toHaveBeenCalledWith(122);
+    });
+
+    it('finalizes a fractional value to an integer on blur', () => {
+      const onChange = jest.fn();
+      const { container } = render(
+        <NumberInput value={12} integer onChange={onChange} />,
+      );
+      const input = getNumberInput(container);
+
+      fireEvent.change(input, { target: { value: '122.1' } });
+      fireEvent.blur(input);
+
+      expect(input.value).toBe('122');
+      expect(onChange).toHaveBeenLastCalledWith(122);
+    });
+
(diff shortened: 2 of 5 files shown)


wangq8 (commented):
This is an AI review comment.

Good fix and thorough write-up. The opt-in `integer` mode, the mount-time heal for legacy values, and the added unit tests are all solid, and the round-trip behavior (typing `122.9` → `123`, `31.1` → clamp to `8` on blur) matches the runtime validation in `agent/component/llm.py`. A few notes:

1. `step={0.1}` on `delay_after_error` is now a dead prop. `originui/number-input`'s stepper buttons hardcode ±1 (`handleIncrement`/`handleDecrement`) and never read `step`, and it also hides the native spinner (`.number-input-hide-spin`). Since this field moved from `ui/input` (no custom steppers) to `originui`, the user now gets +/− buttons that jump 2.0 → 3.0 instead of 0.1. Decimals can still be typed manually (no `integer` flag), so nothing is corrupted, but the 0.1-step affordance is lost. Consider supporting `step` in the stepper buttons, or removing the misleading `step={0.1}` prop (or setting `hideIcons` for this field).

2. `toIntegerValues` rounds but does not clamp. A legacy out-of-range value like `max_retries: 31.1` is rounded to `31` in `use-values`, then relies on the `NumberInput` mount-heal to clamp it to `max=8` (via `field.onChange` → dirty → `useWatchFormChange` persist). This works in the agent form, but it is a two-step correction spread across two files. Consider clamping in `toIntegerValues` too so the `defaultValues` fed to `useForm` are already within range and the intent is self-contained.

3. Minor: the heal `useEffect` now lists `onChange` in its deps. With react-hook-form's `field.onChange` this is stable and fine, but any caller passing an inline callback would re-run the heal on each render. Not a blocker, just something to keep in mind.

No regressions or security concerns; the tests are meaningful and pass.

This is an AI review comment.

dcc123456:
It has already been fixed.

PR #18726, opened 2026-08-25
Go: porting #18663

wangq8:


wangq8 (commented):
This is an AI review comment.

Thanks for porting #18663 to Go. The allowlist approach correctly stops bookkeeping keys (`id`, `created_at`, `conversationId`, `doc_ids`) from reaching strict providers, and threading `name` / `tool_calls` / `tool_call_id` / `function_call` / `refusal` / `audio` through `Message` is consistent with the Python `LLM_MESSAGE_FIELDS` set. A few things to address before merge:

1. **Missing tests.** The Python PR ships 4 regression tests; this port adds none. `normalizeLLMMessage` and `modelMessageFromMap` (especially the `tool_calls` type-switch between `[]map[string]interface{}` and `[]interface{}`) are exactly the kind of code that silently regresses. Please add Go unit tests covering: bookkeeping keys dropped, `tool_calls` / `tool_call_id` preserved, caller map not mutated, and multimodal content-part arrays forwarded unchanged.

2. **`content == nil` guard dropped in `buildChatMessages` (chat_pipeline.go).** The old code skipped messages with `content == nil`; the new code only skips `role == ""`. A stored message with a role but nil `content` and no `tool_calls` will now be serialized by `base_model.go` as `"content": null`, which a strict provider (Groq is the very provider this targets) may reject. Suggest keeping the guard but allowing tool-call messages, e.g. skip only when `role == "" || (content == nil && no tool_calls && no tool_call_id)`.

3. **Scope is broader than the Python original.** In Python, `async_chat` / `async_chat_solo` still rebuild messages as role/content only; the allowlist was applied to `rag_agent` alone. Here `AsyncChat` / `AsyncChatSolo` now forward `name` / `tool_calls` / `function_call` / `refusal` / `audio` on all paths, not just the reasoning path. That is arguably a correctness improvement (tool history is no longer dropped), but it is a behavior change beyond the port — please confirm it is intended.

4. **`Message.Name` is typed `interface{}`.** OpenAI's `name` field is a string, and the other scalar field (`ToolCallID`) uses `string`. Using `interface{}` loses type safety and forces the `msg.Name != nil` check in `base_model.go` to rely on interface nil-ness. Prefer `string` (with `omitempty`) or `*string`.

Minor: `modelMessageFromMap` reuses `stringFromMap` from `bot.go`, which is documented for `entity.JSONMap` (prompt_config) reads. It works because `map[string]interface{}` is assignable to `entity.JSONMap`, but a dedicated helper in this file would be clearer.

This is an AI review comment.

PR #18733, opened 2026-08-25
fix(chat): propagate mind map stream failures

Lem0nTea2002:
### Summary

Mind map generation currently discards the asynchronous error channel returned by `ChatStream`. When a provider fails after emitting partial Markdown, the handler parses and returns an incomplete tree as a successful response.

This change consumes the active error channel for both the configured and fallback models, propagates provider failures, and surfaces cancellation or deadline errors after the stream closes. Successful stream behavior remains unchanged.

Follow-up to https://github.com/infiniflow/ragflow/pull/18615#discussion_r3828866157

### Testing

- Added deterministic tests for successful chunk collection.
- Added coverage for partial output followed by an asynchronous provider error.
- Added coverage for cancellation after stream channels close.
- `gofmt` and `git diff --check` passed.
- Local package execution is blocked before test startup by existing Windows native dependencies; Linux CI will provide package-level confirmation.

### DIFF
--- a/internal/handler/mindmap.go
+++ b/internal/handler/mindmap.go
@@ -72,24 +72,37 @@ func runMindMap(ctx context.Context, config mindMapRunConfig) (mindMapNode, erro
 	// composite-name parsing which fails for bare IDs. If the configured
 	// model can't be resolved, fall back to the tenant's default chat model
 	// (mirrors Python's gen_mindmap get_tenant_default_model_by_type).
-	ch, _, streamErr := config.LLM.ChatStream(streamCtx, modelTenantID, modelID, messages, &modelModule.ChatConfig{})
+	ch, streamErrs, streamErr := config.LLM.ChatStream(streamCtx, modelTenantID, modelID, messages, &modelModule.ChatConfig{})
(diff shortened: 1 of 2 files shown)


Lynn-Inf (changes requested):
The change is reasonable, but `streamErrs` and `streamErr` are confusing. Keep the first as `streamErrs` and rename the second to `findChatModelErr`.

wangq8 (commented):
LGTM. Verified against `internal/service/model_chat.go`: `chatStreamWithContext` always closes the error channel (deferred) and buffers it with cap 1, and it closes before the chunk channel, so `<-streamErrs` in `collectMindMapStream` cannot deadlock. Reading the error after the chunk loop (rather than only at stream open) correctly handles partial-output-then-error, and checking `ctx.Err()` surfaces cancellation/deadline — a small improvement over the previous behavior of silently returning partial text on timeout. Tests cover the three relevant paths. Nice follow-up.

Lem0nTea2002:
Thanks for the review. Renamed the synchronous model lookup error to `findChatModelErr` while keeping `streamErrs` for asynchronous stream failures.

PR #18734, opened 2026-08-25
fix(agent): panic and raw error when auto play runs without a default TTS model

euvre:
### Summary

When a tenant has no default TTS model configured (System Model Settings → TTS empty) and an agent Message node enables auto play (自动播放), running the agent failed in an unfriendly way:

- **Go backend**: the run crashed with `runtime error: invalid memory address or nil pointer dereference` and the UI surfaced the raw Go panic stack. The Message component maps the `auto_play` boolean to the built-in engine selector `"gtts"`; the audio dispatch passed it as a model name together with nil provider/instance pointers, and `ModelProviderService.getModelInstanceAndProviderByName` dereferenced the nil `providerName`.
- **Python backend**: `agent/canvas.py` resolved the tenant default TTS model outside any error handling, so `Exception("No default tts model is set.")` aborted the whole run and the raw error text replaced the agent reply.

**Fix**

- `ModelProviderService.AudioSpeech`: when no model id and no provider/instance/model names are given (the auto_play path), resolve the tenant's default TTS model via `GetTenantDefaultModelByType` and synthesize through it — the same call shape the `/api/v1/chat/audio/speech` endpoint uses. A missing default now yields the typed error `no default tts model is set`.
- `internal/agent/audio/tts_dispatch.go`: built-in engine selectors (`gtts`/`edge-tts`/`custom`) and an empty hint no longer masquerade as model names; they fall through to the tenant-default path (mirrors Python `canvas.run` auto_play semantics). The requested format defaults to `mp3`, matching the chat TTS endpoint.
- `getModelInstanceAndProviderByName`: nil name pointers return a clear error instead of panicking; a user without a tenant no longer returns `(nil, nil)`.
- `defaultModelRefs`: read `tenant.tts_id`/`ocr_id` through `ptrStringValue` — both columns are nullable and the previous unconditional dereference panicked exactly in the reported configuration (every default set except TTS).
- `agent/canvas.py` (Python): the default TTS resolution for auto_play is wrapped in try/except; on failure the run logs a warning and continues without audio instead of replacing the answer with the error text.

TTS failures in the Go Message component were already non-fatal (the textual answer is preserved and the failure is recorded under `outputs["audio_error"]`), so after this change an agent without a default TTS model answers normally and simply skips the audio.

**Verification**

- Reproduced the Go panic end-to-end in the browser UI (agent `Begin → Message(auto_play=true)`, tenant without a default TTS model): the run failed with a toast containing the full panic stack at `getModelInstanceAndProviderByName`. After the fix the same agent run returns the text answer with no error.
- Unit tests: `bash build.sh --test ./internal/agent/audio/... ./internal/service/...` passes, including new cases for the pseudo-engine fallback, the no-default-TTS error path, the partial-name guard, and the NULL `tts_id` column.
- Boundary: actual audio synthesis with a configured default TTS model was not exercised end-to-end (no TTS provider available in this environment); that path reuses the same `GetTenantDefaultModelByType` → `driver.AudioSpeech` call shape as the working chat TTS endpoint.

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -656,8 +656,15 @@ class Canvas(Graph):
                 cpn_obj = self.get_component_obj(self.path[i])
                 if cpn_obj.component_name.lower() == "message":
                     if cpn_obj.get_param("auto_play"):
-                        tts_model_config = get_tenant_default_model_by_type(self._tenant_id, LLMType.TTS)
-                        tts_mdl = LLMBundle(self._tenant_id, tts_model_config)
+                        try:
+                            tts_model_config = get_tenant_default_model_by_type(self._tenant_id, LLMType.TTS)
+                            tts_mdl = LLMBundle(self._tenant_id, tts_model_config)
+                        except Exception as e:
+                            # A missing/unresolvable default TTS model must not
+                            # fail the whole run: skip auto play and keep the
+                            # textual answer flowing.
+                            _logger.warning("Auto play skipped: default TTS model is not available: %s", e)
+                            tts_mdl = None
                     if isinstance(cpn_obj.output("content"), partial):
                         _m = ""
                         buff_m = ""
--- a/internal/agent/audio/tts_dispatch.go
+++ b/internal/agent/audio/tts_dispatch.go
@@ -77,8 +77,10 @@ type TTSDispatcher interface {
 //
 //   - ModelProviderRequest.ModelName (from req.Engine)  → modelName
 //     The Engine field is repurposed as a model identifier hint
-//     in the audio package's contract. Empty falls through to the
-//     model's default TTS model.
+//     in the audio package's contract. Built-in engine selectors
+//     ("gtts" / "edge-tts" / "custom") and an empty hint fall
+//     through to the tenant's default TTS model; any other value
+//     is passed as an explicit model name to the by-name lookup.
 //   - Text  → audioContent
 //   - Voice → TTSConfig.Params["voice"]
 //   - Lang  → TTSConfig.Params["lang"]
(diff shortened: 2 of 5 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the thorough fix. The root cause is correctly identified: `defaultModelRefs` unconditionally dereferenced `tenant.TTSID`/`tenant.OCRID` (both nullable `*string` columns), so a tenant with every default set *except* TTS panicked in the auto_play path. The change to `ptrStringValue(...)` is exactly right, and the companion fixes (nil-guard in `getModelInstanceAndProviderByName`, and the pseudo-engine fallback in `tts_dispatch.go`) close the other panic vectors.

Observations:

1. **`userID` → `GetTenantDefaultModelByType` is consistent with the existing convention.** The new branch passes `AudioSpeech`'s `userID` argument directly as the tenant ID. That looks suspicious on its face, but it matches the existing chat TTS endpoint (`internal/handler/chat_audio.go` passes `user.ID`), and in this port the personal tenant's `ID` is set to the user id at creation (`internal/service/user.go`). So no behavior change there — just flagging it since the parameter is named `userID` while the callee expects `tenantID`.

2. **`EngineCustom` semantics (please confirm).** The switch treats `"custom"` as a built-in selector that falls through to the tenant default, same as `"gtts"`/`"edge-tts"`. This mirrors Python's `canvas.run` auto_play (which always uses `get_tenant_default_model_by_type(TTS)`), so it looks correct — but if `"custom"` can ever represent an explicit tenant model name, it would now be silently ignored. Worth a sanity check that nothing passes a literal model name colliding with these selectors.

3. **Minor nit: default `TTSConfig` has no `Format`.** In the new branch, when `modelConfig` is nil it's initialized to `&modelModule.TTSConfig{}` without `Format`. The agent dispatch path always sets `Format: "mp3"`, so it's moot in practice, but initializing `Format: "mp3"` here too would make the branch self-consistent for any future caller that hits it with a nil config.

4. **Tests are solid.** The pseudo-engine fallback, the no-default-TTS error, the partial-name guard, and the NULL `tts_id` case are all covered. The acknowledged gap (no end-to-end synthesis with a *configured* default TTS) is acceptable since that path reuses the same `GetTenantDefaultModelByType → driver.AudioSpeech` shape as the working chat TTS endpoint.

5. **Python side is safe.** `tts_mdl = None` on failure is fine because `tts_mdl` is pre-initialized to `None` (canvas.py:636) and every downstream use is guarded by `if tts_mdl`. The run continues with text-only output, matching the Go Message component's non-fatal behavior.

No blocking issues — nice work.

This is an AI review comment.

Lynn-Inf (approved):
LGTM

PR #18735, opened 2026-08-25
Fix: Add the chat model to the Python version of the dataset configuration page.

cike8899:
### Summary
Fix: Add the chat model to the Python version of the dataset configuration page.

### DIFF
--- a/web/src/components/parse-configuration/graph-rag-form-fields.tsx
+++ b/web/src/components/parse-configuration/graph-rag-form-fields.tsx
@@ -18,7 +18,7 @@ import { FormLayout } from '@/constants/form';
 import { DocumentParserType, GenerateType } from '@/constants/knowledge';
 import { useTranslate } from '@/hooks/common-hooks';
 import { cn } from '@/lib/utils';
-import { useKnowledgeBaseContext } from '@/pages/dataset/contexts/knowledge-base-context';
+import { useOwnerTenantId } from '@/pages/dataset/contexts/knowledge-base-context';
 import { LLMModelItem } from '@/pages/dataset/setting/python/configuration/common-item';
 import { upperFirst } from 'lodash';
 import { useCallback, useMemo } from 'react';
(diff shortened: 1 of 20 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix — this correctly surfaces the indexing/chat model (`parser_config.llm_id`) on the Python dataset configuration page and tidies up the repeated inline `useKnowledgeBaseContext().knowledgeBase?.tenant_id` calls into a single `useOwnerTenantId()` hook.

A few observations:

1. **English locale already present.** `globalIndexModel` / `globalIndexModelTip` already exist in `en.ts` (and most other locales); this PR adds only the `zh.ts` entries, which is correct. One nit: the `zh` tip text ("用于生成自动元数据、自动关键词和自动问题") is narrower than the English ("Used to generate Knowledge graphs, RAPTOR, auto-metadata, auto-keyword and auto-question") — consider aligning the two so Chinese users aren't missing the Knowledge-graph/RAPTOR part.

2. **Duplicate field registration (please confirm).** `GraphRagItems` in `web/src/components/parse-configuration/graph-rag-form-fields.tsx` also renders an `LLMModelItem` bound to `parser_config.llm_id`. As of this PR that component has no importers in `web/src`, so there's no live conflict — but if/when it gets wired into the Python config page alongside the new field in `python/general-form.tsx`, react-hook-form would register the same field name twice. Worth keeping in mind.

3. **Good cleanup.** Hoisting `useKnowledgeBaseContext()` out of JSX attributes into `useOwnerTenantId()` also removes a rules-of-hooks footgun; the `oxlint-disable-next-line react-refresh/only-export-components` comment is appropriate for a hook exported alongside a context.

4. **Unrelated change.** The `serplyApiKeyTip` re-wrap is pure prettier formatting and unrelated to this fix — harmless, but splitting it out keeps the diff focused.

Otherwise LGTM — no blocking issues.

This is an AI review comment.

PR #18736, opened 2026-08-25
fix(web): keep Firefox chat auto-scroll pinned to the bottom

dcc123456:
### Summary

fix(web): keep Firefox chat auto-scroll pinned to the bottom

Firefox drifted off the bottom partway through a streamed answer for three reasons Chrome does not hit. Scroll anchoring shifted `scrollTop` to hold its anchor node still whenever the answer re-laid out, and the scroll handler cannot tell that apart from a user scrolling up, so auto-follow latched off for the rest of the answer. Firefox's fractional `scrollTop` also made sub-pixel jitter read as an upward gesture. And `scrollHeight - clientHeight` is a difference of two rounded values, so it lands just short of a fractional maximum.

Disable anchoring on the transcript, require a gesture's worth of movement before disarming auto-follow, overshoot so the browser clamps to the real bottom, and widen the near-bottom tolerance.

### DIFF
--- a/web/src/hooks/logic-hooks.ts
+++ b/web/src/hooks/logic-hooks.ts
@@ -414,6 +414,12 @@ export const useSpeechWithSse = (url: string = api.chatsTts) => {
 
 //#region chat hooks
 
+// Firefox reports a fractional `scrollTop`, while `scrollHeight` / `clientHeight`
+// are rounded. `scrollToBottom` therefore lands a sub-pixel *below* the position
+// a native clamp had produced, which reads as a decreasing `scrollTop`. Require a
+// real gesture's worth of movement so that jitter is not mistaken for one.
+const UserScrollUpThreshold = 2;
+
 export const useScrollToBottom = (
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
LGTM, well-reasoned. The root-cause analysis (Firefox scroll anchoring + fractional `scrollTop` + rounded `scrollHeight`/`clientHeight`) is correct, and each change targets a distinct part of the problem. Two minor, non-blocking observations:

- Widening the near-bottom tolerance 25→60 makes auto-follow slightly stickier for small upward scrolls: combined with the 2px threshold, a user scrolling up ~40px is still treated as "at bottom" and may get pulled back down. Worth a quick manual check that reading recent lines mid-stream still feels natural.
- `overflowAnchor = 'none'` is set in an effect keyed on `[containerRef]`. If the parent ever swaps the DOM node while keeping the same ref object, the style won't be re-applied (and isn't reverted on unmount). Fine for the current single-container chat usage.

PR #18737, opened 2026-08-25
Refactor: expose knn top_k and knn num_candidaties

wangq8:
Follow on #18300


wangq8 (commented):
This is an AI review comment.

Thanks for this refactor — cleanly renaming `top` → `knn_top_k` and adding `knn_num_candidates`, with good REST backward-compat via `AliasChoices`. I traced the full call graph (all `retrieval()`/`search()`/`get_vector()` callers) and verified the positional-argument fixes in `bot_api.py`/`chunk_api.py` are correct. A few things to consider before merge:

## 1. `retrieval()` drops the `top` kwarg — breaking for out-of-tree callers
`rag/nlp/search.py` removes the `top=1024` parameter entirely (and re-orders the signature so `doc_ids` now follows `vector_similarity_weight`). All in-repo callers were updated, so this is safe today, but `Dealer.retrieval` is a semi-public method also used by forks/plugins (e.g. the enterprise fork, `rag/advanced_rag`, custom tools). Any caller still passing `top=...` will hit a `TypeError: unexpected keyword argument 'top'`. Recommend keeping a deprecated `top` shim that maps to `knn_top_k` (with a `DeprecationWarning`) for one release, or explicitly coordinating the same change in downstream forks.

## 2. `knn_num_candidates` is unbounded and silently clamped
In `dataset_api_service.py`:
```python
knn_top_k = max(1, min(int(req.get("knn_top_k", 1024)), 2048))
knn_num_candidates = int(req.get("knn_num_candidates", 2048))   # no upper bound, no >= knn_top_k check
```
- `knn_top_k` is clamped to `[1, 2048]`, but `knn_num_candidates` has only pydantic `ge=1` and no `le`.
- In `es_conn.py` it's then force-clamped to `max(k, min(num_candidates, 10000))`, so a user passing `knn_num_candidates < knn_top_k` silently gets `k`, and a huge value silently becomes 10000. This is correct for ES (num_candidates must be ≥ k) but gives the caller no feedback.
- Suggest adding `le=10000` (or whatever the engine cap is) to `SearchDatasetReq`/`SearchDatasetsReq` and either validating `knn_num_candidates >= knn_top_k` or documenting the clamp in the docstring.

## 3. `search()` `size` default silently changed from 1024 → 30
```python
-        topk = int(req.get("topk", 1024))
-        ps = int(req.get("size", topk))
+        ps = int(req.get("size", 30))
```
The old default `size` was `topk` (1024); the new default is a hardcoded `30`. I checked all direct callers of `Dealer.search()` (`graphrag/utils.py`, `chunk_api.py`, `dataset_api_service.py`) and `retrieval()` — they all set `size` explicitly, so there's no live regression. But the default semantic changed silently; worth a comment or keeping it derived from `knn_top_k` to avoid a future footgun.

## 4. `knn_num_candidates` only affects Elasticsearch
`infinity_conn.py` explicitly pops `num_candidates` (`matchExpr.extra_options.pop("num_candidates", None)`), so the new parameter is a no-op on Infinity. The API docstring presents it as a general "advanced knn parameter" — worth a one-line note that it's currently ES-only.

## 5. Minor: redundant double-omit in the web layer
`agent/form/retrieval-form/use-values.ts` now returns `omit(formData, 'top_k')`, and `agent/form/tool-form/retrieval-form/index.tsx` wraps it again with `omit(useValues(), 'top_k')`. Harmless, but redundant.

## 6. Missing tests
- No unit test for the new ES `num_candidates` clamp logic in `es_conn.py` (the `max(k, min(...))` branch).
- No test asserting the legacy `top_k` → `knn_top_k` alias actually maps for `SearchDatasetReq`/`SearchDatasetsReq` (e.g. sending `top_k` and checking the dumped dict contains `knn_top_k`).

## Nits
- `num_candidates or top_k * 2` in `get_vector` treats `0` as "use default"; fine given `ge=1`, but `if num_candidates is None` is more explicit.

Verified positives: legacy `top_k` alias works end-to-end (`validation_alias=AliasChoices(...)` → `model_dump(by_alias=True)` emits the field name `knn_top_k`, which is what the service reads); positional-arg call sites in `bot_api.py`/`chunk_api.py` are correctly converted to keyword args; `get_vector` callers that omit `num_candidates` keep their previous ES behavior (`top_k * 2`).

This is an AI review comment.

wangq8 (commented):
This is an AI review comment.

Follow-up review (re-checking after the `fix`/`FIx comments` commits). Nice work — the previous concerns are mostly addressed: `knn_num_candidates` now has `le=10000` + a `model_validator` enforcing `>= knn_top_k`, the ES-only scope is documented, the `size` default change got an explanatory comment, and the `advanced_rag`/`graphrag` callers were updated. A few remaining items:

## 1. GraphRAG methods now correctly `await` the async `get_vector` — please confirm + test
`rag/graphrag/search.py` changes `get_relevant_ents_by_keywords` / `get_relevant_relations_by_txt` from `def` to `async def` and adds `await self.get_vector(...)`. This is actually a **bug fix**: `Dealer.get_vector` is `async`, and the previous code called it without `await`, so `dataStore.search(...)` was receiving a coroutine instead of a `MatchDenseExpr` (which `es_conn` would silently skip via its `isinstance(..., MatchDenseExpr)` check — i.e. the KG vector path wasn't running). This is a real functional fix bundled into a rename refactor; worth calling out in the PR description and ideally a test.

## 2. `get_vector` default `num_candidates=20` no longer scales with `top_k`
```python
async def get_vector(self, txt, emb_mdl, top_k=10, num_candidates=20, similarity=0.1):
```
The v1 version used `num_candidates or top_k * 2`, so the default always tracked `top_k`. With a static `20`, any future `get_vector(txt, mdl, top_k=200)` (no explicit `num_candidates`) gets `num_candidates=20`, which ES then force-clamps up to `k=200` — silently discarding the intended 2x. All current callers pass `num_candidates` explicitly so there's no live bug, but I'd suggest `num_candidates: int | None = None` and computing `num_candidates if num_candidates is not None else top_k * 2` to keep the invariant.

## 3. Legacy `top_k > 2048` now rejects instead of silently clamping
`knn_top_k` gained `le=2048` (correct for the new name), but the legacy alias `top_k` previously had only `ge=1` and the service silently clamped `min(..., 2048)`. A legacy client sending `top_k=3000` used to get 2048; now it gets a 400 validation error. Reasonable strictness improvement, just flagging the backward-compat nuance for the alias.

## 4. `retrieval()` still drops the `top` kwarg (breaking for out-of-tree callers)
In-repo coverage is now complete (including `advanced_rag` and `graphrag`), but `Dealer.retrieval` is semi-public. The enterprise fork and any plugins calling `retrieval(..., top=...)` will hit `TypeError` on merge. Recommend a short-lived `top` shim that maps to `knn_top_k` (with `DeprecationWarning`), or an explicit coordinated rename in `ragflow_enterprise`.

## 5. Still missing tests
- No unit test for the new `model_validator` (`knn_num_candidates >= knn_top_k`).
- No test asserting the `top_k` → `knn_top_k` alias actually maps through `SearchDatasetReq`/`SearchDatasetsReq` (i.e. send `top_k`, assert the dumped dict contains `knn_top_k`).
- No test for the ES `num_candidates` clamp branch (`max(k, min(...))`) in `es_conn.py`.

## Nits
- `web/src/pages/agent/form/tool-form/retrieval-form/index.tsx` does `omit(useValues(), 'top_k')` while `useValues()` already returns `omit(formData, 'top_k')` — redundant double-omit.
- The PR body is empty; a short "### Summary" describing the rename + the GraphRAG async fix would help reviewers.

This is an AI review comment.

PR #18738, opened 2026-08-25
fix(web): fall back a page when the dataset file list empties

dcc123456:
### Summary

fix(web): fall back a page when the dataset file list empties

Deleting every document on the last page left the table stranded on an empty page. Reuse the existing useGoToPreviousPageOnEmpty hook, already applied on the agents/datasets/chats/searches list pages.


wangq8 (commented):
LGTM. Clean reuse of the existing `useGoToPreviousPageOnEmpty` hook, consistent with the agents/datasets/chats/searches pages. `documents` and `loading` are both in scope from `useFetchDocumentList`, and the hook guards `pagination.current > 1`, so an empty dataset on page 1 won't loop. No regressions spotted.

PR #18745, opened 2026-08-25
Porting #18737 to go

wangq8:
Porting #18737 to go
Test: passed.

### DIFF
--- a/internal/agent/tool/retrieval_nlp.go
+++ b/internal/agent/tool/retrieval_nlp.go
@@ -29,8 +29,8 @@
 //   tool.RetrievalRequest.Query      → nlp.RetrievalRequest.Question
 //   tool.RetrievalRequest.DatasetIDs → nlp.RetrievalRequest.KbIDs
 //   tool.RetrievalRequest.TopN       → nlp.RetrievalRequest.PageSize
-//   tool.RetrievalRequest.TopK       → nlp.RetrievalRequest.Top
-//                                       (fallback Top=TopN*4 so rerank
+//   tool.RetrievalRequest.TopK       → nlp.RetrievalRequest.KNNTopK
+//                                       (fallback KNNTopK=TopN*4 so rerank
 //                                        has headroom)
 //   tool.RetrievalRequest.KeywordsSimilarityWeight
 //                                    → nlp.RetrievalRequest.VectorSimilarityWeight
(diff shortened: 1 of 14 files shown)


wangq8 (commented):
This is an AI review comment.

**Summary:** Faithful Go port of #18737. The `Top` → `KNNTopK` rename plus the new `KNNNumCandidates` field is propagated consistently through the agent tool, nlp service, dataset service, handlers, and the ES engine. Validation bounds match the Python pydantic model (`knn_top_k` 1..2048, `knn_num_candidates` 1..10000, `num_candidates >= knn_top_k`), and the new tests cover the important paths. The `num_candidates` plumbing into the ES KNN query in `chunk.go` is correct and safely clamped.

Non-blocking observations / suggestions:

1. **`pageSize` default changed from `topk` to `30`** (`internal/service/nlp/retrieval.go` `Search()`). This is faithful to Python (`ps = int(req.get("size", 30))`), but it's an independent behavior change: any internal caller that omits `PageSize` and previously relied on the large default (1024) will now get 30 results. Worth calling out in the PR description.

2. **`knn_num_candidates` is not re-derived after the dataset-level `search_config.top_k` override.** In `SearchDatasets`, `knnTopK` is re-clamped from `searchConfig["top_k"]` (1..2048), but `knnNumCandidates` (request value or default 2048) is left unchanged. If a request sets a small `knn_num_candidates` (valid against its own `knn_top_k`) while the dataset config raises `top_k`, the effective `num_candidates` can end up `< knn_top_k`. The ES layer (`chunk.go`) silently bumps it via `max(k, min(n, 10000))`, so nothing breaks, but the user's explicit value is silently ignored. This mirrors the Python code (same gap), so it's optional — consider re-clamping `knn_num_candidates = max(knn_num_candidates, knn_top_k)` after the config override for consistency.

3. **Legacy `top_k` error message.** When the user supplies the legacy `top_k` alias, the validation error reads `knn_top_k must be between 1 and 2048`, which can be confusing. Cosmetic — consider mentioning the alias in the message.

4. **Dify path passes `topK` unclamped into `KNNTopK`** (`internal/handler/dify_retrieval_handler.go`). This bypasses the 2048 upper bound enforced on the main dataset API. It's pre-existing behavior (only renamed here), but now that `knn_num_candidates` defaults to 2048, a large `knn_top_k` will be silently clamped/bumped at the ES layer. Worth a follow-up clamp for consistency.

5. **Test coverage gaps (minor):** no case for the legacy `top_k` > 2048 being rejected, and no boundary case `knn_top_k=2048` with default `knn_num_candidates=2048` (equal, valid).

No correctness or security issues found; the rename is complete with no leftover `Top` field references in the retrieval/dataset/nlp paths.

This is an AI review comment.

PR #18747, opened 2026-08-25
Fix(go): PaddleOCR -> PaddleOCR.local, PaddleOCR.net -> PaddleOCR

Lynn-Inf:
### Summary

As title.


wangq8 (changes requested):
Thanks for this rename + dispatch cleanup. I reviewed it against `main` and found one blocking issue plus a migration concern to confirm before merge.

## Blocking: duplicate declarations in `pdf_vision_dispatch.go` (won't compile)

The new block for `resolvePaddleOCRModelForDispatch`, `defaultResolvePaddleOCRModelForDispatch`, and `dispatchPaddleOCRPdf` was **added** but the previous block (still written against the old `PaddleOCR.Net` / `.Net` naming) was left in place. The file now declares each symbol twice, which is a Go compile error (`redeclared in this block`):

- `var resolvePaddleOCRModelForDispatch` → declared at ~line 231 and again at ~line 358
- `func defaultResolvePaddleOCRModelForDispatch` → declared at ~line 233 and again at ~line 360
- `func dispatchPaddleOCRPdf` → declared at ~line 247 and again at ~line 374

Please remove the old block (the one whose comments reference `PaddleOCR.Net` / `.Net`). This is consistent with the `UNSTABLE` merge state reported by GitHub.

## Concern: provider rename is breaking for existing tenants (no migration)

The rename swaps the public provider names:
- cloud `PaddleOCR.Net` → `PaddleOCR`
- local `PaddleOCR` → `PaddleOCR.local`

`CreateModelDriver` is keyed on `provider.Name` (see `model.go:295`), so the factory switch is now consistent with the new `conf/models/*.json` names — that part is correct. However, tenant model configs are persisted by provider name, and `paddleOCRProviderAliases()` now returns only `["PaddleOCR", "PaddleOCR.local"]`, dropping `"PaddleOCR.Net"`. Tenants that already saved the old names (`PaddleOCR.Net` for cloud, or `PaddleOCR` for local) will no longer resolve through either the direct lookup or the alias fallback. If the Go backend is still pre-release and no persisted tenants exist this is fine, but otherwise a migration (or keeping `"PaddleOCR.Net"` in the alias list) is needed.

## Otherwise looks good

- `PaddleOCRConfigFromAPIKey` correctly mirrors Python's `PaddleOCROcrModel` (base_url prefers `paddleocr_base_url` over `paddleocr_api_url`, nested `api_key` key, non-JSON → zero values) and has solid table-driven tests.
- `isPaddleOCRDriver` / `Name()` updates are internally consistent.
- No leftover `PaddleOCRNet` / `PaddleOCR.Net` references remain in `web/`.
- The new `PaddleOCR.local` provider schema + `ListModels` wiring (case-insensitive via `EqualFold`) is reasonable.

Minor: the redundant self-alias in `paddleOCRProviderAliases()` (`"PaddleOCR"` retried after the original `"PaddleOCR"` lookup already failed) is harmless but could be trimmed to just the alternate spelling.

wangq8 (commented):
This is an AI review comment.

Re-reviewed at head `761426022` ("fix: build"). The blocking issue from my previous review is fixed and CI is green, so this is now down to a few confirm/cleanup items.

## Resolved ✅
- Duplicate `resolvePaddleOCRModelForDispatch` / `defaultResolvePaddleOCRModelForDispatch` / `dispatchPaddleOCRPdf` declarations are gone — each symbol is now declared exactly once in `pdf_vision_dispatch.go`, and `ragflow_preflight` + `ragflow_tests_* (go)` pass.

## Still open — provider rename is breaking for existing tenants (please confirm)
The public provider names swap (cloud `PaddleOCR.Net` → `PaddleOCR`, local `PaddleOCR` → `PaddleOCR.local`), but `paddleOCRProviderAliases()` now returns only `["PaddleOCR", "PaddleOCR.local"]` — the old `"PaddleOCR.Net"` spelling is dropped everywhere (Go alias list and the web `LLMFactory` enum). Tenant model configs are persisted by provider name, so any tenant that saved the old names will no longer resolve through the direct lookup or the alias fallback, and the web UI will no longer recognize `llm_factory: "PaddleOCR.Net"`. If the Go backend is pre-release with no persisted tenants this is fine; otherwise we need a migration or read-time legacy-name normalization (keep `"PaddleOCR.Net"` in the alias list, and normalize the web enum lookup).

## Parity gap in `PaddleOCRConfigFromAPIKey`
Python's `PaddleOCROcrModel._resolve_config` accepts both lowercase (`paddleocr_*`) and uppercase (`PADDLEOCR_*`) keys in the api_key JSON (the uppercase form is what env auto-provisioning writes). The new Go helper only reads lowercase keys, so a persisted api_key carrying `PADDLEOCR_BASE_URL` / `PADDLEOCR_ACCESS_TOKEN` / `PADDLEOCR_ALGORITHM` would yield zero values on the Go side. Since both backends may read the same persisted api_key, consider matching the uppercase fallback too.

## `CheckConnection` unwrapping is not scoped to PaddleOCR
The new block in `model_service.go` runs `PaddleOCRConfigFromAPIKey` for every provider's `CheckConnection`, not just PaddleOCR. It's effectively harmless today (only a JSON api_key containing `paddleocr_access_token` would trigger), but scoping it to the PaddleOCR provider name/class would make intent clearer and avoid surprising other JSON-keyed providers.

## Minor: inconsistent empty-text handling
`dispatchPaddleOCRPdf` now returns an error on empty OCR text, but `parsePDFWithPaddleOCR` still silently returns a zero-page result (`pageCount = 0`) on empty text. Pre-existing, but now inconsistent with the dispatch path — worth deciding whether empty → error should apply in both paths (otherwise a scanned PDF with no extractable text is silently treated as an empty-but-successful parse).

## Nitpick
The "mirroring Python's …" comments are fine for traceability; no action needed unless you prefer to drop them.

This is an AI review comment.

wangq8 (commented):
This is an AI review comment.

Reviewed the diff. The rename correctly aligns the Go backend with Python's canonical provider names (`PaddleOCR` = cloud, `PaddleOCR.local` = local; Python's `PaddleOCROcrModel._FACTORY_NAME` is "PaddleOCR") and fixes `parsePDFWithPaddleOCR` to use the async Job API exactly like Python's `PaddleOCRParser.parse_pdf()`. Tests are thorough and pin the api_key JSON contract and the provider rename.

Notes / questions (non-blocking):

1. `CheckConnection` in `internal/service/model_service.go` applies `PaddleOCRConfigFromAPIKey(apiKey)` to every provider, not just PaddleOCR. For a non-PaddleOCR provider whose api_key happens to be a JSON object carrying `paddleocr_access_token` (or a nested `api_key.paddleocr_access_token`), the key/url would be silently rewritten. Please gate this on the provider (e.g. `strings.EqualFold(providerName, "PaddleOCR")`) so it cannot affect other providers.

2. Backward compatibility: renaming the registered provider names orphans any tenant that already persisted the old names under the Go backend — old cloud "PaddleOCR.Net" no longer matches `paddleOCRProviderAliases()` (which dropped "PaddleOCR.Net"), and old local "PaddleOCR" now resolves to the cloud driver. If the Go backend is still pre-GA this is fine, but if it has been deployed you may want a migration/alias for the legacy spellings. Worth confirming.

3. Cosmetic: the error string in `parsePDFWithPaddleOCR` still says "requires paddleocr_base_url or PADDLEOCR_BASE_URL", but the function now also falls back to `PADDLEOCR_API_URL`. Minor message inconsistency.

4. `PaddleOCRConfigFromAPIKey`'s `paddleocr_algorithm` is discarded (`_`) in CheckConnection. Acceptable for a connection check, just noting it.

Otherwise the factory/driver/web/enum/conf renames are consistent (no remaining "PaddleOCR.Net"/`PaddleOCRNet` references; `FindProvider` is case-insensitive so `ListModels` lookup works). LGTM aside from the above.

This is an AI review comment.

PR #18748, opened 2026-08-25
fix(graph): ignore malformed n-hop metadata

blueOf:
### Summary

GraphRAG persists n-hop relation paths in entity chunks. When a legacy or partially written record is missing path/weights, has mismatched lengths, or contains non-numeric weights, the retrieval loop can raise and fail the whole query.

This change validates each persisted path, normalizes finite numeric weights, logs and skips malformed entries, and keeps valid paths available for retrieval.

### Tests

- Targeted GraphRAG unit test file: 99 passed
- Python syntax compilation for the modified files
- git diff --check

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -23,7 +23,7 @@ import pandas as pd
 
 from common.misc_utils import get_uuid
 from rag.graphrag.query_analyze_prompt import PROMPTS
-from rag.graphrag.utils import get_entity_type2samples, get_llm_cache, set_llm_cache, get_relation
+from rag.graphrag.utils import _iter_valid_n_hop_edges, get_entity_type2samples, get_llm_cache, set_llm_cache, get_relation
 from common.token_utils import num_tokens_from_string
 
 from rag.nlp.search import Dealer, index_name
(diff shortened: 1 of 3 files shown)


wangq8:
@blueOf the code under `graphrag` is legacy, and no longer used. now the new `graphrag` is in knowledge compilation.

wangq8 (commented):
This is an AI review comment.

Reviewed the diff. The fix is correct and well-tested.

What it does correctly:
- `_iter_valid_n_hop_edges` defensively validates each persisted n-hop entry (list/dict shape, path length >= 2, weights length == path-1, string entities, numeric finite weights) and skips malformed records instead of letting `nbr["path"]`/`nbr["weights"]` raise and abort the whole query.
- Edge-index semantics are preserved: the yielded `edge_index` comes from `enumerate(zip(path, path[1:], weights))`, matching the original `i in range(len(path)-1)` used for the `sim` denominator `(2 + i)` and the weight lookup.
- Normalizing weights via `float()` also fixes the prior latent crash where a string weight caused `max(pagerank, wts[i])` to raise a TypeError.
- Swapping root `logging.warning` for a module-level `logger` is a nice improvement.
- Tests cover malformed entries, non-list metadata, non-finite/overflow weights, and confirm the warning log does not leak raw persisted values.

Minor (non-blocking):
- Because a single malformed edge triggers `break` and the `for...else` then skips `yield from edges`, a path with one bad edge drops the entire path, including earlier valid edges. That is conservative and strictly better than crashing, but if you ever want to salvage partial paths you'd yield the accumulated valid edges before the break instead.

No regressions or security concerns found. LGTM (not approving).

This is an AI review comment.

PR #18749, opened 2026-08-25
fix(web): keep token chunker settings when saving in One mode

euvre:
### Summary

On the dataset configuration page, switching the Token Chunker delimiter mode to **One** and saving permanently wiped the previously configured values: after a page reload the overlap percent showed `0` and the delimiter list was reduced to the seeded default `\n` (reported in the Feishu group "v1.0 go测试").

Root cause: `transformTokenChunkerParams` (`web/src/pages/agent/utils.ts`) — the shared form→API transform used both by the dataset configuration save path (`transformFormConfigToApi`) and the agent canvas DSL builder — forced `overlapped_percent: 0` and `delimiters: []` whenever `delimiter_mode === 'one'`. The zeroing was introduced with the One-mode form (#14024) as payload cosmetics and predates the dataset page reusing the transform (#17100); every save in One mode therefore overwrote the user's settings in `parser_config`.

The chunker ignores both fields in One mode (`rag/flow/chunker/token_chunker.py` returns early and merges all sections into a single chunk; the Go pipeline keeps the same contract), so persisting the configured values is behavior-neutral at runtime and preserves the user's settings when they switch back to Delimiters mode.

- Stop zeroing `overlapped_percent` / `delimiters` in `transformTokenChunkerParams`; always convert and persist the configured values (unchanged handling for `delimiter` mode, `children_delimiters` gating, and context-window splitting).
- Add `web/src/pages/agent/utils.test.ts` covering both modes.

### Verification

- Reproduced on the Go backend before the fix: PUT payload contained `"delimiter_mode":"one","overlapped_percent":0,"delimiters":[]`; after reload the UI showed overlap `0` and a single `\n` delimiter.
- After the fix, the same flow (set overlap 9 + a custom `###` delimiter → switch to One → save → reload → switch back to Delimiters) shows overlap `9` and delimiters `\n`, `###`; the stored `parser_config` keeps `overlapped_percent: 0.09, delimiters: ["\n","###"], delimiter_mode: "one"`.
- `npx jest src/pages/agent/utils.test.ts` passes; `oxlint` clean on changed files; `tsc --noEmit` reports no errors in the changed files (remaining repo-wide errors are pre-existing in untouched files).

Verification boundary: end-to-end check done against the local Go API + browser UI; real ingestion with the One mode was not run (not needed for this frontend-only persistence fix).


wangq8 (approved):
LGTM. The root-cause analysis is correct: `transformTokenChunkerParams` forced `overlapped_percent: 0` and `delimiters: []` whenever `delimiter_mode === 'one'`, so every save in One mode overwrote the user's settings in `parser_config`.

I verified the runtime-neutrality claim against `rag/flow/chunker/token_chunker.py`: in `delimiter_mode == "one"` it returns early and merges all sections into a single chunk (both the text/markdown path and the JSON path), so `overlapped_percent` and `delimiters` are never consulted. Persisting the configured values is therefore behavior-neutral while preserving user settings for when they switch back to Delimiters mode.

The new tests cover both modes and pin the correct behavior (overlapped_percent conversion and delimiter/children_delimiters gating). Nice small, well-scoped fix with regression coverage.

One minor note (non-blocking): the test uses `toBe(0.09)` / `toBe(0.3)` on the result of `Number(x)/100` — this is safe today since the literal and the computed value resolve to the same IEEE-754 double, but if you ever touch that transform, consider `toBeCloseTo` to avoid float-comparison fragility.

dcc123456 (approved):

wangq8:
@euvre fix CI

euvre:
Hi @wangq8, thanks for the nudge on CI — and @dcc123456, thanks for the approval!

**CI:** I dug into the failing `ragflow_tests_elasticsearch (web)` job and both of yesterday's failures were runner-side infrastructure, not the diff:

- Runs 32933451726 and 32935627791 died during `docker network create` with `Error response from daemon: all predefined address pools have been fully subnetted` — the shared runner daemon had 53 leftover networks, so the job exited **before any test step ran**.
- The earlier preflight failure (run 32926362542) was `git fetch` transport errors (`GnuTLS recv error`, `early EOF`, HTTP/2 CANCEL).
- For reference, the same diff was fully green on Aug 25 (run 32839112636: elasticsearch + infinity + preflight all passed), and the infinity leg kept passing through the flaky window.

Since job re-runs aren't possible from a fork account, I pushed 70e2f9fb3 to retrigger the workflow — run 33047789783 is now fully green: preflight ✅ 27s, elasticsearch ✅ 4m16s, infinity ✅ 5m47s.

**Float assertions** (from the review): agreed, applied in the same commit 70e2f9fb3 — both `overlapped_percent` assertions in `web/src/pages/agent/utils.test.ts` now use `toBeCloseTo` instead of `toBe`, so they stay robust if the transform's rounding ever changes.

Local verification: `npm run type-check` shows no errors in `src/pages/agent/` (the repo currently carries ~210 pre-existing TS errors in unrelated pages, untouched by this PR); the unit tests themselves run in the CI matrix above. To be explicit about the boundary: I did not run any shared-service end-to-end verification locally — the green CI matrix is the end-to-end evidence for this change.

Thanks both for reviewing!

PR #18750, opened 2026-08-25
Fix total as the total, follow Go

wangq8:
Tested: pass

### DIFF
--- a/api/apps/restful_apis/bot_api.py
+++ b/api/apps/restful_apis/bot_api.py
@@ -466,7 +466,6 @@ async def retrieval_test_embedded(tenant_id=None):
             enrich_chunks_with_document_metadata(ranks["chunks"], metadata_fields)
 
         ranks["labels"] = labels
-        ranks["total"] = len(ranks["chunks"])
 
         return get_json_result(data=ranks)
 
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -1587,7 +1587,6 @@ async def search_datasets(tenant_id: str, req: dict):
         except Exception:
             logging.warning("search_datasets KG retrieval failed: datasets=%s tenant=%s", kb_ids, tenant_id, exc_info=True)
     ranks["chunks"] = settings.retriever.retrieval_by_children(ranks["chunks"], tenant_ids)
-    ranks["total"] = len(ranks["chunks"])
 
     for c in ranks["chunks"]:
         c.pop("vector", None)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix. The intent is correct: `ranks["total"]` is supposed to report the total number of matching chunks (before pagination), not the size of the current page. `RetrievalService.retrieval()` in `rag/nlp/search.py` already sets `ranks["total"] = int(filtered_count)`, so the two removed lines were clobbering that value with `len(ranks["chunks"])` — i.e. the page size, plus the KG chunk when `use_kg` is enabled. Removing them restores the correct total.

Two things worth addressing before merging:

1. **Incomplete / inconsistent.** The same pattern still exists in `search()` (`api/apps/services/dataset_api_service.py`, the `ranks["total"] = len(ranks["chunks"])` right after `retrieval_by_children`). That endpoint has the identical bug — it reports page size instead of the true total. For consistency with this change (and with the Go implementation you're following), that line should also be removed.

2. **`total` can now exceed `len(chunks)`.** After this change:
   - In `search_datasets`, `retrieval_by_children` consolidates child chunks into their parent chunk, so the returned `chunks` count can be smaller than `filtered_count`; `total` now reports the pre-consolidation count.
   - In `_retrieval` (bot_api), when `use_kg` is enabled, the KG chunk is `insert(0, ...)`'d into `ranks["chunks"]` but is no longer reflected in `total`.

   These are arguably the correct semantics (total = total matches), but please confirm the SDK/frontend pagination logic doesn't assume `total == len(chunks)` when `page_size >= total`; otherwise pagination UI could misreport.

Minor: consider adding/adjusting a test that asserts `total` equals the pre-pagination match count for a query matching more chunks than `page_size` (and, for `search_datasets`, the child→parent consolidation case).

This is an AI review comment.

wangq8 (commented):
This is an AI review comment.

Thanks for the fix. The intent is correct: `ranks["total"]` is supposed to report the total number of matching chunks (before pagination), not the size of the current page. `RetrievalService.retrieval()` in `rag/nlp/search.py` already sets `ranks["total"] = int(filtered_count)`, so the two removed lines were clobbering that value with `len(ranks["chunks"])` — i.e. the page size, plus the KG chunk when `use_kg` is enabled. Removing them restores the correct total.

Two things worth addressing before merging:

1. **Incomplete / inconsistent.** The same pattern still exists in `search()` (`api/apps/services/dataset_api_service.py`, the `ranks["total"] = len(ranks["chunks"])` right after `retrieval_by_children`). That endpoint has the identical bug — it reports page size instead of the true total. For consistency with this change (and with the Go implementation you're following), that line should also be removed.

2. **`total` can now exceed `len(chunks)`.** After this change:
   - In `search_datasets`, `retrieval_by_children` consolidates child chunks into their parent chunk, so the returned `chunks` count can be smaller than `filtered_count`; `total` now reports the pre-consolidation count.
   - In `_retrieval` (bot_api), when `use_kg` is enabled, the KG chunk is `insert(0, ...)`'d into `ranks["chunks"]` but is no longer reflected in `total`.

   These are arguably the correct semantics (total = total matches), but please confirm the SDK/frontend pagination logic doesn't assume `total == len(chunks)` when `page_size >= total`; otherwise pagination UI could misreport.

Minor: consider adding/adjusting a test that asserts `total` equals the pre-pagination match count for a query matching more chunks than `page_size` (and, for `search_datasets`, the child→parent consolidation case).

This is an AI review comment.

PR #18751, opened 2026-08-25
fix(web): stop multi-model card config from overwriting chat LLM settings

euvre:
### Summary

In the chat multi-model comparison view, each card's model popover lets users pick a model and tune parameters (temperature, top_p, presence/frequency penalty, max_tokens). The "Apply model configs" checkmark on each card called `patchChat` with the card's `llm_id`, `tenant_llm_id` and the whole card form as `llm_setting`, which **overwrote the chat dialog's saved model configuration**. After applying, the original chat settings panel showed the card's comparison values (e.g. temperature 0.1232, top_p 0.1222, max_tokens 128000) instead of the chat's own parameters, and UI-only `temperatureEnabled`/`topPEnabled`/... flags were persisted into the dialog's `llm_setting`. Reported in the Feishu group as: after saving model parameters with several decimal places in multi-model chat, the original chat model parameters are affected.

The patch served no runtime purpose: a multi-model card is a comparison surface, and each card already sends its own `llm_id` and generation parameters with every completion request (`useImperativeHandle` spreads the card form into the request; the chat completion handler accepts per-request `llm_id`/`temperature`/`top_p`/... overrides). It only destroyed the chat's saved configuration.

### Changes

- `web/src/pages/next-chats/chat/chat-box/next-multiple-chat-box.tsx`: remove the per-card "Apply model configs" button, its `handleApplyConfig` dialog patch, and the now-unused imports (`Tooltip*`, `usePatchChat`, `useFindLlmByUuid`, `useWatch`, `isEmpty`/`omit`, `ListCheck`, i18next `t`). The `syncedDialogIdRef` guard that keeps each card's independently selected model is kept, with its comment updated.
- `web/src/locales/*`: remove the dead `chat.applyModelConfigs` key from all 11 locale files.
- `test/playwright/e2e/test_next_apps_chat.py`: drop the e2e step that clicked Apply and asserted the dialog PATCH payload (`llm_id` + `llm_setting`); the following send step still asserts that completion requests carry model-specific fields, which keeps the request-scoped behavior covered.

Difference from the earlier fix in #17866: that PR kept the Apply button and only stopped sibling cards' selections from being clobbered after the patch refetched the dialog. This PR removes the destructive persistence itself, so the multi-model view can no longer mutate the chat's saved model configuration at all.

### Verification

- Reproduced on the running stack (Go API + web dev server): saving baseline chat settings (0.5/0.5/0.3/0.2/2048), then setting 0.1232/0.1222/0.1232/0.1234/128000 on a multi-model card and clicking Apply fired `PATCH /api/v1/chats/{id}` and replaced the dialog's `llm_setting` (confirmed in the request/response bodies).
- After the fix, re-walked the same path: the Apply button is gone, editing card parameters fires no chat write request, and the chat settings still show the original values after leaving the multi-model view.
- `oxlint` clean on the changed component; `tsc --noEmit` reports no errors in any touched file (the remaining repo-wide errors are pre-existing and unrelated).
- The heavy Playwright e2e flow was updated but not executed in this environment; the request-scoped completion behavior it now asserts is unchanged code.


wangq8 (commented):
This is an AI review comment.

Reviewed the diff. This is a clean, well-scoped fix.

What it does correctly:
- Removes the destructive "Apply model configs" path entirely instead of patching around it. Per-card completions already forward their own llm_id/temperature/top_p/etc. per request, so the dialog PATCH only mutated the chat's saved llm_setting and served no runtime purpose.
- The `syncedDialogIdRef` guard is preserved and its comment updated to the new rationale (a refetch must not clobber per-card selection), which is still correct after removing Apply.
- Dead imports (`Tooltip`, `usePatchChat`, `useFindLlmByUuid`, `useWatch`, `isEmpty`/`omit`, `ListCheck`, `t`) and the `chat.applyModelConfigs` locale key are removed across all 11 locale files. I verified against the PR head that no dangling references remain (no `t(...)`, `useWatch`, `usePatchChat`, `useFindLlmByUuid`, `isEmpty`, `omit`, `ListCheck`, or `Tooltip` usages left in the file).
- The e2e flow drops the Apply/PATCH assertion but keeps the send-step assertion that completion requests carry per-card model fields, so the request-scoped behavior stays covered.

No regressions, style, or security concerns found. LGTM (not approving).

This is an AI review comment.

dcc123456:
The direction of the modification is wrong.

PR #18753, opened 2026-08-25
fix(agent): default missing Graph path so stored DSL cannot KeyError

### LINKED ISSUE #18746: [Bug]: Dataflow pipeline crashes with KeyError: 'path' when stored DSL lacks the path key
### Self Checks

2db8eb6c91184402e2979b6b52b77063919c7e17

### RAGFlow image version

v0.27.0

- Deployment: Docker Compose (`docker/docker-compose.yml`, `ragflow-cpu` profile)
- OS type: macOS host (Docker Desktop)
- Task type affected: `dataflow` (Data Pipeline / `category=dataflow_canvas` agent)

### Actual behavior

A Data Pipeline (agent, `category=dataflow_canvas`) that had previously run successfully stopped being runnable at all. Every parse attempt — both a Knowledge Base document parse that references the pipeline and a fresh "Test Run" from the canvas editor — failed immediately (0s) with:

```
[ERROR][Exception]: 'path'
```

This is a bare `KeyError: 'path'`, raised before the pipeline's first component (`File`) ever executes — there's no "File fetched." log line, just "Task has been received." followed directly by the exception.

Root cause: `agent/canvas.py`, `Graph.load()`:

```python
def load(self):
    self.components = self.dsl["components"]
    ...
    self.path = self.dsl["path"]   # <-- strict bracket access
```

`path` is pure runtime execution-state (the list of already-visited component ids), not part of the authored graph. The class already treats it that way everywhere else:

- `Graph.__init__` defaults it to `self.path = []`.
- `rag/flow/pipeline.py`'s `Pipeline.run()` self-initializes it on first use: `if not self.path: self.path.append("File")`.
- A sibling field on the same class, `memory`, is already read defensively: `self.memory = self.dsl.get("memory", [])`.

So a DSL that has never been executed yet — or was re-saved by a code path that doesn't round-trip this backend-only bookkeeping field — legitimately has no `"path"` key. `Graph.load()` shouldn't treat that as fatal.

I confirmed this directly by dumping the affected canvas's stored DSL from the `user_canvas` table:

```json
top-level keys: ["components", "globals", "graph", "messages", "variables"]
```

No `path` key at all — `Graph.load()` raised `KeyError: 'path'` unconditionally for this row, every time, until the key was backfilled.

### Expected behavior

`Graph.load()` should default `path` to `[]` when the DSL doesn't have one yet, the same way `memory` already does, instead of crashing pipeline execution outright.

### Steps to reproduce

1. Create a Data Pipeline / dataflow agent (`category=dataflow_canvas`) with any components (e.g. `File → Parser → Chunker → ...`).
2. Confirm it runs fine via "Test Run" in the canvas editor.
3. Get (or force) the stored DSL to lose its `"path"` key — in our case this happened via the canvas editor's own autosave after a backend reconnect; to reproduce directly without depending on that: `UPDATE user_canvas SET dsl = JSON_REMOVE(dsl, '$.path') WHERE id = '<id>';` (MySQL).
4. Either re-run "Test Run" on that canvas, or parse/re-parse a Knowledge Base document configured to use this pipeline as its parsing method.
5. Observe the task fails instantly with `[ERROR][Exception]: 'path'`.

### Additional information

Task log excerpt (Knowledge Base document parse):

```
16:49:49 Task has been received.
16:49:49 [ERROR][Exception]: 'path'
```

Call path: `rag/svr/task_executor_refactor/dataflow_service.py::DataflowService.run_dataflow()` → `rag/flow/pipeline.py::Pipeline.__init__()` (subclass of `agent/canvas.py::Graph`) → `Graph.load()`:

```
File "agent/canvas.py", line 111, in load
    self.path = self.dsl["path"]
KeyError: 'path'
```

Fix applied and verified locally (one line, matching the existing `memory` field's pattern):

```diff
-        self.path = self.dsl["path"]
+        self.path = self.dsl.get("path", [])
```

After applying this, the same pipeline (and other pipelines whose stored DSL is missing `path`) run again without any other change. Happy to open a PR with this fix if useful.

BetterAndBetterII:
`Graph.load()` required a `path` key on the stored DSL. That field is runtime execution state, not part of the authored graph: `Graph.__init__` already defaults it to `[]`, and the sibling `memory` field is already read with `.get(..., [])`.

A dataflow canvas whose stored DSL only has `components` / `globals` / `graph` / `messages` / `variables` therefore crashed on every Test Run and Knowledge Base parse with `KeyError: 'path'`, before the first component ran (`DataflowService.run_dataflow()` → `Pipeline.__init__()` → `Graph.load()`).

Default missing `path` to `[]`. Existing `path` is unchanged.

Fixes #18746

### DIFF
--- a/agent/canvas.py
+++ b/agent/canvas.py
@@ -108,7 +108,7 @@ class Graph:
         for k, cpn in self.components.items():
             cpn["obj"] = component_class(cpn["obj"]["component_name"])(self, k, component_params[k])
 
-        self.path = self.dsl["path"]
+        self.path = self.dsl.get("path", [])
 
     @staticmethod
     def validate_component_parameters(dsl):
--- a/test/unit_test/agent/test_graph_load_missing_path.py
+++ b/test/unit_test/agent/test_graph_load_missing_path.py
@@ -1,0 +1,205 @@
+#
+#  Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Regression tests for Graph.load() when stored DSL has no `path` key.
+
+Issue #18746: a dataflow canvas whose stored DSL only has
+`components` / `globals` / `graph` / `messages` / `variables` crashed
+on every Test Run and Knowledge Base parse with KeyError: 'path'.
+
+Call path: DataflowService.run_dataflow() -> Pipeline.__init__() ->
+Graph.load(). `path` is runtime execution state, not authored graph,
+and Graph.__init__ already defaults it to []. Canvas.load() already
+reads the sibling `memory` field with .get(..., []).
+"""
+
+from __future__ import annotations
+
+import contextvars
+import importlib.util
+import json
+import sys
+from pathlib import Path
+from types import ModuleType, SimpleNamespace
+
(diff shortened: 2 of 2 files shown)


wangq8:
LGTM.

PR #18759, opened 2026-08-25
fix(parsers): treat empty bytes as a real payload, not a missing file (PR #18712 sibling)

Harsh23Kashyap:
## Problem

A 0-byte upload reaches the chunkers as `b""`, not as `None`. The guard in `rag/svr/task_executor.py:312-313` only rejects `None`:

```python
binary = await get_storage_binary(bucket, name)
if binary is None:
    raise FileNotFoundError(...)
```

`b""` is not `None`, so the guard passes. The chunkers that do `if binary:` then took the file-path branch and opened the display name as a file (which is not a real path on disk) — crashing with `FileNotFoundError` on a name that never existed.

PR #18712 (open as of this writing) fixed this in `rag/app/email.py` and `rag/app/presentation.py` — 5 sites in 2 files. The author explicitly listed the same pattern remaining in **4 more files** and left them out:

> The same truthiness pattern remains in `deepdoc/parser/mineru_parser.py`, `somark_parser.py`, `tcadp_parser.py` and `rag/app/naive.py`. I did not trace whether an empty upload reaches those, so I left them alone.

This PR is that trace + the matching fix.

## Sites changed

| file | line | before | after |
|---|---|---|---|
| `deepdoc/parser/mineru_parser.py` | 1003 | `if binary:` | `if binary is not None:` |
| `deepdoc/parser/somark_parser.py` | 698 | `if binary:` | `if binary is not None:` |
| `deepdoc/parser/tcadp_parser.py` | 268 | `if binary:` | `if binary is not None:` |
| `deepdoc/parser/tcadp_parser.py` | 395 | `binary if binary else filepath` | `binary if binary is not None else filepath` |
| `deepdoc/parser/tcadp_parser.py` | 401 | `if binary:` | `if binary is not None:` |
| `rag/app/naive.py` | 118, 428, 725 | `filename if not binary else binary` | `filename if binary is None else binary` |
| `rag/app/naive.py` | 559 | `Document(filename) if not binary else Document(BytesIO(binary))` | `Document(filename) if binary is None else Document(BytesIO(binary))` |
| `rag/app/naive.py` | 684 | `BytesIO(binary) if binary else open(filename, "rb")` | `BytesIO(binary) if binary is not None else open(filename, "rb")` |
| `rag/app/naive.py` | 713 | `if not binary:` | `if binary is None:` |

All 9 are reachable from `chunk()`. Behaviour for non-empty `binary` and for `binary=None` is unchanged. Only the `b""` case moves, and it moves from a crash to normal parsing of an empty payload.

## Tests

`test/unit_test/rag/app/test_parser_empty_binary_is_not_none.py` — six AST-based regression tests, mirroring the same pattern PR #18712 used for its `test_email_empty_binary.py`:

- `test_module_has_no_truthy_binary_check` (parametrized over the four patched files) — walks the AST and asserts no top-level `if binary:` / `if not binary:` / `X if binary else Y` / `X if not binary else Y` pattern remains. The AST helpers only flag a node when the test is the bare name `binary` or `not binary`; `binary is None` / `binary is not None` are not flagged, so the post-fix code passes cleanly.
- `test_mineru_extract_pdf_takes_binary_branch_on_empty_bytes` — pins the temp-dir `mineru_bin_pdf_` string in the parser body so a future refactor that drops the binary branch trips this test.
- `test_naive_by_deepdoc_filename_picks_binary_when_present` — pins the specific `filename if not binary else binary` / `binary if binary else filename` ternaries in `naive.py` that the original PR's email/presentation tests would not have caught.

Pre-fix verification (stashed source change):
- `somark` and `tcadp` and `naive` AST tests fail with `AssertionError: ... still has the falsy-empty-binary pattern. Found: [(line, '<pattern>')]`.
- The naive.py ternary test fails on both pre-fix ternaries.
- The MinerU body test still passes (it only checks for the temp-dir string, which is present in both pre- and post-fix code).

Post-fix: 6/6 pass.

`ruff check` + `ruff format --check` clean on the four source files plus the new test.

## consult-kiro note

Per the standing rule (2026-08-24), the local OpenCode Tier A panel was attempted. All four models (`opencode/big-pickle`, `opencode/gemini-3.7-flash`, `opencode/gpt-5-nano`, `opencode/deepseek-v4-flash-free`) returned empty answers / `rc=124` within the 180s hard-timeout window — same pattern as the prior 7 cycles. The change is small and the tests are the primary review surface; the consult is recorded here for completeness.

## Risk

- Backward compatible: every site moves from one explicit path to another; the behaviour for non-empty `binary` and for `binary=None` is unchanged. Only the `b""` case moves, and it moves from a crash to a normal empty parse.
- The four patched files are exercised by the existing deepdoc + rag app unit-test suites; the new AST test pins the fix without depending on the live OCR services.
- No public API change. No schema or response-format change.

## Future work

- The same falsy-empty-binary pattern is still in `deepdoc/parser/html_parser.py` and `deepdoc/parser/pdf_parser.py` in some code paths; PR #18712 patched the first one and left the second alone. A future PR could close the loop.


xugangqiang (commented):
## Test coverage gaps

The AST-based regression tests assert the *absence* of truthiness patterns but do not verify the *presence* or *direction* of the replacement. This leaves two holes:

1. **No branch-direction assertion.** A reversed explicit condition like `filename if binary is not None else binary` would pass all six tests while reintroducing the bug. The tests should additionally assert that `binary is not None` / `binary is None` appear at the fixed sites, or (better) that the binary branch is actually selected when `binary=b`.

2. **No functional coverage.** Unlike the parent PR #18712, which called the real `email.chunk(..., binary=b)` and asserted no `FileNotFoundError` on the display name, this PR never invokes a parser with empty bytes. `test_mineru_extract_pdf_takes_binary_branch_on_empty_bytes` only greps for the temp-dir string and passes both pre- and post-fix, so it is not a real regression test.

**Suggestion:** add at least one functional test that calls a patched entry point (e.g. `by_deepdoc`, `Docx.__call__`, or `Pdf.__call__`) with `binary=b` and a non-existent `filename`, and asserts it does not raise `FileNotFoundError` on the display name. This matches the bar set by #18712 and closes the direction-reversal hole that the AST test cannot catch.

xugangqiang (commented):
## Test coverage gaps

The AST-based regression tests assert the *absence* of truthiness patterns but do not verify the *presence* or *direction* of the replacement. This leaves two holes:

1. **No branch-direction assertion.** A reversed explicit condition like `filename if binary is not None else binary` would pass all six tests while reintroducing the bug. The tests should additionally assert that `binary is not None` / `binary is None` appear at the fixed sites, or (better) that the binary branch is actually selected when `binary=b""`.

2. **No functional coverage.** Unlike the parent PR #18712, which called the real `email.chunk(..., binary=b"")` and asserted no `FileNotFoundError` on the display name, this PR never invokes a parser with empty bytes. `test_mineru_extract_pdf_takes_binary_branch_on_empty_bytes` only greps for the temp-dir string and passes both pre- and post-fix, so it is not a real regression test.

**Suggestion:** add at least one functional test that calls a patched entry point (e.g. `by_deepdoc`, `Docx.__call__`, or `Pdf.__call__`) with `binary=b""` and a non-existent `filename`, and asserts it does not raise `FileNotFoundError` on the display name. This matches the bar set by #18712 and closes the direction-reversal hole that the AST test cannot catch.

xugangqiang:
## Test coverage gaps

The AST-based regression tests assert the absence of truthiness patterns but do not verify the presence or direction of the replacement. This leaves two holes:

1. No branch-direction assertion. A reversed explicit condition like `filename if binary is not None else binary` would pass all six tests while reintroducing the bug. The tests should additionally assert that `binary is not None` / `binary is None` appear at the fixed sites, or better, that the binary branch is actually selected when `binary=b""`.

2. No functional coverage. Unlike the parent PR #18712, which called the real `email.chunk(..., binary=b""\)` and asserted no `FileNotFoundError` on the display name, this PR never invokes a parser with empty bytes. `test_mineru_extract_pdf_takes_binary_branch_on_empty_bytes` only greps for the temp-dir string and passes both pre- and post-fix, so it is not a real regression test.

Suggestion: add at least one functional test that calls a patched entry point (e.g. `by_deepdoc`, `Docx.__call__`, or `Pdf.__call__`) with `binary=b""` and a non-existent `filename`, and asserts it does not raise `FileNotFoundError` on the display name. This matches the bar set by #18712 and closes the direction-reversal hole that the AST test cannot catch.

Issue #16978, opened 2026-07-16
[Bug]: PDF cannot insert native images; Mineru does not recognize them.


cuihangrui:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

v0.25.1

### RAGFlow image version

v0.25.1

### Other environment information

```Markdown

```

### Actual behavior

Below is the configuration of my knowledge base.

<img width="751" height="1110" alt="Image" src="https://github.com/user-attachments/assets/15253b41-ece2-455b-bd72-d83df16365d1" />

<img width="772" height="1114" alt="Image" src="https://github.com/user-attachments/assets/59996a1a-cb2a-497e-8cd8-91687ed94d17" />

<img width="752" height="1062" alt="Image" src="https://github.com/user-attachments/assets/3ea2e4c7-ec3b-4ed4-8b46-521dc6cd0615" />
If the uploaded file is an image, the text within it can be properly extracted. However, if the uploaded file is a PDF or DOCX format with images, no information from the images will be recognized.

### Steps to reproduce

```Markdown
If the uploaded file is an image, the text within it can be properly extracted. However, if the uploaded file is a PDF or DOCX format with images, no information from the images will be recognized.
```

cuihangrui:
How to configure the IMAGE2TEXT model? Why can't I find the "IMAGE2TEXT" category in the model management? Is it classified under the llm category? Does it belong to the capabilities of the vl model?
@dosu

zeweihan:
The standalone-image-vs-embedded-image asymmetry is the part worth underlining, because it means the same content has two completely different retrieval fates depending on its container format — and the system reports success in both cases.

[...]

PR #18761, opened 2026-08-25
feat(mineru): observability for embedded image loss in chunk stream (#16978)

Taranum01:
Addresses #16978.

## Problem
Embedded images inside PDF/DOCX documents were silently dropped from the chunk stream when there was no caption, no footnote, and no IMAGE2TEXT VLM description available. Ingestion still returned HTTP 200 with healthy-looking `chunk_num`, but the operative visual content (signature pages, exhibits, redline markup, scanned contracts) never made it to retrieval.

The container-format asymmetry described in the issue — standalone PNG/JPGs go through the Picture chunker with full OCR + VLM; the same image embedded in a PDF/DOCX goes through `_transfer_to_sections` which only lifts captions/footnotes — means operators get no signal that anything was lost.

## What this PR actually does (and doesn't)
**This is observability, not a fix.** Embedded images are still dropped in raw mode (the only mode that drops — naive/manual/paper always emit the image via `_transfer_to_tables`, even with empty text, so they are never "dropped"). The real fix is in #16978 item 2 (merge-granularity alignment), which is correctly out of scope here.

What this PR does add:
- An `image_coverage` stamp at the parser boundary that turns the loss from invisible to inspectable.
- A per-parse summary log so operators can grep / alert on the coverage counters.
- A test that pins the new metric semantics.

## Coverage counters (per parse)
- `images_detected` — every IMAGE block the parser saw.
- `images_chunked` — IMAGE blocks emitted to the chunk stream (raw mode: via `_transfer_to_sections`; app-media mode: via `_transfer_to_tables`).
- `images_described` — IMAGE blocks where a `vlm_description` was actually emitted (text path: embedded in section; table path: appended to the text tuple).
- `images_dropped_no_text` — IMAGE blocks whose text was empty after `_transfer_to_sections` skipped the section-building branch (raw-mode only — app-media modes always emit even with empty text).
- `images_unreadable_resource` — IMAGE blocks whose binary resource couldn't be opened (file missing, corrupted bytes, etc.) even though the text payload was non-empty. Distinguishes "text was there but binary was missing" from "no text at all".

The result is stashed on `self.last_image_coverage` and a final `INFO` summary is emitted with the `vlm_configured` flag.

## Per-mode semantics
- **raw mode** (`parse_method="raw"`): images_detected counts every IMAGE block; images_chunked counts every IMAGE block whose section text was non-empty; images_described counts every IMAGE block whose vlm_description was embedded in the section. images_dropped_no_text applies.
- **app-media mode** (`parse_method` in `{naive, manual, paper}`): images_detected counts every IMAGE block; images_chunked counts every IMAGE block whose binary was readable; images_described counts every IMAGE block whose vlm_description was appended to the tuple. images_dropped_no_text does NOT apply (empty-text blocks still emit a tuple with a placeholder).

## Tests
- 4 existing tests updated for the new `images_unreadable_resource` key.
- 6 new tests cover: captioned image, silently-dropped, VLM-described, mixed batch, app-media modes (naive/manual/paper), leading-newline fallback, init invariant.

40 of my tests pass. The 2 remaining failures are pre-existing upstream `RagTokenizer.set_language` errors, unrelated to this change.

## Review fixes (this revision)
- Moved the `images_described += 1` increment in `_transfer_to_tables` AFTER the readability check so unreadable images aren't double-counted as described.
- Updated the corresponding test to assert the new behavior (`images_described == 1`, `images_unreadable_resource == 1`).
- Removed review-history wording from comments per the AGENTS.md guidance.

### DIFF
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -960,7 +960,7 @@ class MinerUParser(RAGFlowPdfParser):
             else:
                 sections.append((section, self._line_tag(output)))
         # Stash on the instance so callers (and parse_pdf) can inspect coverage
-        # for the most recent parse without changing the existing return shape.
+        # for the most recent parse.
         self.last_image_coverage = image_coverage
         if image_coverage["images_detected"] > 0:
             self.logger.info(
@@ -1015,7 +1015,7 @@ class MinerUParser(RAGFlowPdfParser):
             vlm_description = (output.get("vlm_description") or "").strip()
             if vlm_description:
                 texts.append(vlm_description)
-                self.last_image_coverage["images_described"] += 1
+            has_vlm_description = bool(vlm_description)
 
             image = None
             image_path = output.get("img_path")
@@ -1025,9 +1025,9 @@ class MinerUParser(RAGFlowPdfParser):
                         source.load()
                         image = source.copy()
                 except Exception as e:
-                    self.logger.warning(f"[MinerU] Failed to load image '{image_path}': {e}")
+                    self.logger.debug(f"[MinerU] Failed to load image '{image_path}': {e}")
             if image is None:
-                self.logger.warning("[MinerU] Skip image without a readable resource: %s", image_path)
+                self.logger.debug("[MinerU] Skip image without a readable resource: %s", image_path)
                 # The IMAGE block was detected and counted as chunked, but
                 # Image.open() couldn't read the resource. Distinguish this
                 # from "dropped_no_text" (which means no caption/footnote/
@@ -1036,6 +1036,13 @@ class MinerUParser(RAGFlowPdfParser):
                 self.last_image_coverage["images_chunked"] -= 1
                 self.last_image_coverage["images_unreadable_resource"] += 1
                 continue
+
+            # Only count this image as "described" when it actually
+            # reaches the tables list. An unreadable resource above
+            # (or a vlm_description that survives image-load failure)
+            # shouldn't inflate the metric.
+            if has_vlm_description:
+                self.last_image_coverage["images_described"] += 1
 
             tables.append(((image, texts or [""]), positions))
         return tables
@@ -1175,7 +1182,7 @@ class MinerUParser(RAGFlowPdfParser):
             # vlm_configured flag tells operators whether downstream chunks
             # could have been enriched by a vision model at all. The
             # coverage dict is initialized in __init__ so this access is
-            # unconditional — see issue #16978 review follow-up.
+            # unconditional — see issue #16978.
             coverage = self.last_image_coverage
             coverage["vlm_configured"] = vision_model is not None
             self.logger.info(
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -152,6 +152,16 @@ class MinerUParser(RAGFlowPdfParser):
         self.page_from = 0
         self.page_to = MAXIMUM_PAGE_NUMBER
         self.logger = logging.getLogger(self.__class__.__name__)
+        # Initialize the coverage stamp so callers don't need a
+        # getattr(self, "last_image_coverage", None) fallback — the
+        # attribute exists from construction regardless of whether
+        # parse_pdf has run yet (issue #16978 review follow-up).
+        self.last_image_coverage: dict = {
+            "images_detected": 0,
+            "images_chunked": 0,
+            "images_described": 0,
+            "images_dropped_no_text": 0,
+        }
 
     @staticmethod
     def _is_zipinfo_symlink(member: zipfile.ZipInfo) -> bool:
@@ -847,8 +857,27 @@ class MinerUParser(RAGFlowPdfParser):
     def _transfer_to_sections(self, outputs: list[dict[str, Any]], parse_method: str = None, table_enable: bool = False):
         sections = []
         parse_method = (parse_method or "raw").lower()
+        # Track image coverage so silent loss of embedded PDF/DOCX images is
+        # queryable at the serialization boundary (issue #16978). When an
+        # IMAGE block yields nothing — no caption, no footnote, no VLM
+        # description — the image body is dropped from the chunk stream, so
+        # we surface that gap explicitly instead of letting it disappear.
+        image_coverage = {
+            "images_detected": 0,
+            "images_chunked": 0,
+            "images_dropped_no_text": 0,
+            "images_described": 0,
+        }
         for output in outputs:
             output_type = output.get("type")
+            # Count every IMAGE block we see — including those routed
+            # through naive/manual/paper's separate _transfer_to_tables
+            # path below — so the image_coverage stamp accurately reports
+            # "detected" regardless of parse_method (issue #16978 review
+            # follow-up: previously the increment lived below this skip
+            # and naive/manual/paper reported images_detected=0).
+            if output_type == MinerUContentType.IMAGE:
+                image_coverage["images_detected"] += 1
             # These chunkers consume tables and images separately, so exclude
             # media from their text sections. Raw consumers keep legacy sections.
             # Charts are routed the same way as images: MinerU emits them as
@@ -871,7 +900,14 @@ class MinerUParser(RAGFlowPdfParser):
                     # the chunk so it becomes searchable / retrievable.
                     vlm_description = (output.get("vlm_description") or "").strip()
                     if vlm_description:
+                        image_coverage["images_described"] += 1
                         section = (section.strip("\n") + "\n" + vlm_description).strip("\n") if section.strip() else vlm_description
+                    # Strip any residual newlines so the empty-after-strip case
+                    # ("\n" from empty caption+footnote joined by "\n") is
+                    # correctly counted as dropped at the boundary check below
+                    # (issue #16978).
+                    if not section.strip():
+                        section = ""
                 case MinerUContentType.CHART:
                     section = "".join(output.get("chart_caption", [])) + "\n" + "".join(output.get("chart_footnote", []))
                     vlm_description = (output.get("vlm_description") or "").strip()
@@ -904,8 +940,20 @@ class MinerUParser(RAGFlowPdfParser):
             if output_type == MinerUContentType.TABLE and not table_enable:
                 section = self._sanitize_section_text(section)
             if not section:
-                self.logger.debug("[MinerU] Skip section after sanitization: type=%s", output.get("type"))
+                if output.get("type") == MinerUContentType.IMAGE:
+                    image_coverage["images_dropped_no_text"] += 1
+                    self.logger.warning(
+                        "[MinerU] Dropped embedded image at page_idx=%s with no caption, footnote, or VLM description "
+                        "(img_path=%s). Configure an IMAGE2TEXT vision model to make embedded images retrievable.",
+                        output.get("page_idx"),
+                        output.get("img_path"),
+                    )
+                else:
+                    self.logger.debug("[MinerU] Skip section after sanitization: type=%s", output.get("type"))
                 continue
+
+            if output.get("type") == MinerUContentType.IMAGE:
+                image_coverage["images_chunked"] += 1
 
             if section and parse_method in {"manual", "pipeline"}:
                 sections.append((section, output["type"], self._line_tag(output)))
@@ -913,6 +961,17 @@ class MinerUParser(RAGFlowPdfParser):
                 sections.append((section + self._line_tag(output), output["type"]))
             else:
                 sections.append((section, self._line_tag(output)))
+        # Stash on the instance so callers (and parse_pdf) can inspect coverage
+        # for the most recent parse without changing the existing return shape.
+        self.last_image_coverage = image_coverage
+        if image_coverage["images_detected"] > 0:
+            self.logger.info(
+                "[MinerU] image_coverage detected=%s chunked=%s described=%s dropped_no_text=%s",
+                image_coverage["images_detected"],
+                image_coverage["images_chunked"],
+                image_coverage["images_described"],
+                image_coverage["images_dropped_no_text"],
+            )
         return sections
 
     def _transfer_to_tables(self, outputs: list[dict[str, Any]], table_enable: bool = True):
@@ -1099,6 +1158,22 @@ class MinerUParser(RAGFlowPdfParser):
 
             sections = self._transfer_to_sections(outputs, parse_method, enable_table)
             tables = self._transfer_to_tables(outputs, enable_table) if (parse_method or "raw").lower() in {"naive", "manual", "paper"} else []
+            # Emit an image_coverage stamp so silent loss of embedded PDF/DOCX
+            # images is visible at the parser boundary (issue #16978). The
+            # vlm_configured flag tells operators whether downstream chunks
+            # could have been enriched by a vision model at all. The
+            # coverage dict is initialized in __init__ so this access is
+            # unconditional — see issue #16978 review follow-up.
+            coverage = self.last_image_coverage
+            coverage["vlm_configured"] = vision_model is not None
+            self.logger.info(
+                "[MinerU] image_coverage final detected=%s chunked=%s described=%s dropped_no_text=%s vlm_configured=%s",
+                coverage["images_detected"],
+                coverage["images_chunked"],
+                coverage["images_described"],
+                coverage["images_dropped_no_text"],
+                coverage["vlm_configured"],
+            )
             return sections, tables
         finally:
             if temp_pdf and temp_pdf.exists():
--- a/deepdoc/parser/mineru_parser.py
+++ b/deepdoc/parser/mineru_parser.py
@@ -870,11 +870,11 @@ class MinerUParser(RAGFlowPdfParser):
         }
         for output in outputs:
             output_type = output.get("type")
-            # Count every IMAGE block we see — including those routed
-            # through naive/manual/paper's separate _transfer_to_tables
-            # path below — so the image_coverage stamp accurately reports
-            # "detected" regardless of parse_method.
-            if output_type == MinerUContentType.IMAGE:
+            # Count every visual block we see (IMAGE and CHART — MinerU emits
+            # both as visual content) so the image_coverage stamp reports
+            # "detected" symmetrically with "chunked" / "described" (which
+            # both already include CHART via _transfer_to_tables).
+            if output_type in {MinerUContentType.IMAGE, MinerUContentType.CHART}:
                 image_coverage["images_detected"] += 1
             # These chunkers consume tables and images separately, so exclude
             # media from their text sections. Raw consumers keep legacy sections.
@@ -910,6 +910,7 @@ class MinerUParser(RAGFlowPdfParser):
                     section = "".join(output.get("chart_caption", [])) + "\n" + "".join(output.get("chart_footnote", []))
                     vlm_description = (output.get("vlm_description") or "").strip()
                     if vlm_description:
+                        image_coverage["images_described"] += 1
                         section = (section.strip("\n") + "\n" + vlm_description).strip("\n") if section.strip() else vlm_description
                 case MinerUContentType.EQUATION:
                     section = output.get("text", "")
@@ -938,11 +939,15 @@ class MinerUParser(RAGFlowPdfParser):
             if output_type == MinerUContentType.TABLE and not table_enable:
                 section = self._sanitize_section_text(section)
             if not section:
-                if output.get("type") == MinerUContentType.IMAGE:
+                if output.get("type") in {MinerUContentType.IMAGE, MinerUContentType.CHART}:
                     image_coverage["images_dropped_no_text"] += 1
-                    self.logger.warning(
-                        "[MinerU] Dropped embedded image at page_idx=%s with no caption, footnote, or VLM description "
-                        "(img_path=%s). Configure an IMAGE2TEXT vision model to make embedded images retrievable.",
+                    # DEBUG (not WARNING) because PDFs with many decorative
+                    # images would flood the log; the parse_pdf summary
+                    # already reports the aggregate dropped_no_text count.
+                    self.logger.debug(
+                        "[MinerU] Dropped embedded %s at page_idx=%s with no caption, footnote, or VLM description "
+                        "(img_path=%s). Configure an IMAGE2TEXT vision model to make embedded visuals retrievable.",
+                        "chart" if output.get("type") == MinerUContentType.CHART else "image",
                         output.get("page_idx"),
                         output.get("img_path"),
                     )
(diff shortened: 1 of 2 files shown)


xugangqiang:
## Review findings

### 1. `images_detected` is counted in the wrong place (functional bug)

The increment of `images_detected` lives inside the `match/case MinerUContentType.IMAGE:` block (~line 859), but `naive` / `manual` / `paper` parse methods never reach it — those modes `continue` earlier at line ~839:

```python
if parse_method in {"naive", "manual", "paper"} and (output_type == MinerUContentType.IMAGE or ...):
    continue
```

So for the three modes where images are actually chunked (via `_transfer_to_tables`), the coverage stamp reports `images_detected=0`. This directly contradicts the PR's goal of making loss visible — operators will be misled into thinking those modes have no images.

**Fix:** count `images_detected` *before* the naive/manual/paper skip branch, so every mode reports detected images correctly. `chunked` / `dropped` / `described` can stay where they are.

### 2. Duplicate `return` statement (dead code)

The diff adds a new `return sections, tables` (~line 1121) but leaves the original one (~line 1122) in place, making the second one unreachable dead code. One of them should be removed.

### 3. Test coverage gaps

All 4 new tests use `parse_method="raw"`. None cover `naive` / `manual` / `paper` — exactly the modes where the coverage tracking is currently broken. Also missing:
- a test for the `parse_pdf` final `image_coverage final ...` log line,
- a test for the `vlm_configured` flag,
- a test exercising the `getattr(self, "last_image_coverage", None)` fallback path.

### 4. `last_image_coverage` not initialized in `__init__`

It is only assigned inside `_transfer_to_sections`; `parse_pdf` relies on `getattr(..., None)` as a fallback. Initializing it in `__init__` would make the attribute's existence independent of call ordering.

### 5. On the fix vs. scope

The PR improves observability, but it does not fix issue #16978 itself — embedded images are still dropped, only now with a warning. That is a reasonable first step, but the PR title `fix(mineru): ...` overpromises; `surface` or `observe` would be more accurate.

Taranum01:
Thanks for the thorough review, @xugangqiang. All five findings addressed in this force-push:

## Review #1 — `images_detected` counted in the wrong place
Moved the increment ABOVE the `naive/manual/paper` early-skip branch (around line 850), so every IMAGE block the parser sees is counted regardless of parse_method. Previously the increment lived inside the `MinerUContentType.IMAGE` match-case which was unreachable for naive/manual/paper. New parametrized test `test_transfer_to_sections_counts_images_detected_for_app_media_modes[naive|manual|paper]` pins the new behavior.

## Review #2 — duplicate return statement
Removed the second `return sections, tables` (unreachable dead code).

## Review #3 — test coverage gaps
Added 6 new tests:
- `test_transfer_to_sections_counts_images_detected_for_app_media_modes[naive|manual|paper]` — parametrized over the three app-media modes (the bug zone).
- `test_transfer_to_sections_app_media_mode_with_image_dropped_after_sanitization` — invariants for app-media mode edge cases.
- `test_transfer_to_sections_app_media_image_with_only_vlm_description` — VLM counter still increments.
- `test_mineruparser_initializes_last_image_coverage_in_constructor` — covers the new __init__ path (review #4).
- `test_parse_pdf_emits_image_coverage_final_log_with_vlm_configured_flag` — covers the parse_pdf summary log with vlm_configured=True.
- `test_parse_pdf_image_coverage_final_log_reflects_no_vision_model` — covers vlm_configured=False.

All 38 of my tests pass; the 2 failures (`test_transfer_to_tables_emits_ordered_typed_media`, `test_tokenize_table_uses_payload_type_instead_of_html_content`) are pre-existing upstream `RagTokenizer.set_language` errors, not introduced by this PR.

## Review #4 — `last_image_coverage` not initialized in __init__
Initialized the attribute in `__init__` with the same zeroed-out dict shape the helper fills in. Replaced the `getattr(self, ..., None) or {...}` fallback in parse_pdf with a direct attribute read — the attribute is now guaranteed to exist.

## Review #5 — title overpromises
Renamed the PR title from `fix(mineru): ...` to `feat(mineru): surface silent loss of embedded PDF images via image_coverage stamp`. The commit message body now opens with: *"This is observability, not a fix — embedded images are still dropped in the naive/manual/paper modes. The PR title 'feat' (vs 'fix') and the docstring call out that surface, not repair."* The docstrings on `_transfer_to_sections` and `parse_pdf` already say "surface the gap explicitly instead of letting it disappear".

`git diff upstream/main..HEAD --stat` shows:
```
deepdoc/parser/mineru_parser.py                    |  33 ++++++++--
test/unit_test/deepdoc/parser/test_mineru_parser.py | 238 +++++++++++++++++++++
```

xugangqiang:
Review of the current code (head 5a0d249). The observability direction is reasonable, but I found the following problems that should be fixed before merge:

## 1. `images_described` overcounts unreadable images (functional bug)
In `_transfer_to_tables` (`deepdoc/parser/mineru_parser.py:979`), `images_described` is incremented *before* the image-readability check at line 990:

```python
if vlm_description:
    texts.append(vlm_description)
    self.last_image_coverage["images_described"] += 1   # before readability check
...
if image is None:                                       # line 990
    ... images_chunked -= 1; images_unreadable_resource += 1; continue
```

An IMAGE block that has a `vlm_description` but whose binary resource cannot be opened is counted as `described` yet never emitted to `tables`. This directly undermines the PR's stated goal of accurate loss reporting. CodeRabbit flagged this (outside-diff L976-998) but it is still unaddressed.
**Fix:** move `self.last_image_coverage["images_described"] += 1` to *after* the `if image is None` check, i.e. only when the image is successfully appended to `tables`.

## 2. A test locks in the bug above
`test_transfer_to_tables_counts_unreadable_resource` asserts `images_described == 2`, but image B in that test is unreadable, so the correct value should be `1`. The test asserts buggy behavior. After fixing #1, change the assertion to `images_described == 1`.

## 3. `Closes #16978` is misleading
The PR only adds observability; embedded images are still dropped (and only in `raw` mode — `naive`/`manual`/`paper` always emit the image via `_transfer_to_tables`, even with empty text, so they are never "dropped"). The core issue is not resolved. Please link a dedicated observability issue instead of `Closes #16978`.

## 4. Metric semantics differ by `parse_method`
`images_dropped_no_text` is effectively a `raw`-mode-only metric (app-media modes never drop). Operators may be misled into thinking app-media modes lose images when they don't. Worth documenting the per-mode meaning in the summary log or PR description.

## 5. Fragile shared mutable counter
`self.last_image_coverage` is mutated in two methods (`_transfer_to_sections` and `_transfer_to_tables`) for the same image; double-counting is only avoided by the `parse_method` gate in `parse_pdf`. The test `test_transfer_to_tables_counts_unreadable_resource` manually calls both methods and asserts a confusing `images_chunked == 3` (the author's own comment says "let me double-check the actual decrement"). Consider passing one explicit `coverage` object between the two transfers instead of stashing it on `self`.

## 6. Comment hygiene (per repo AGENTS.md)
- Remove compatibility wording "without changing the existing return shape" (around line 932).
- Remove review-history narration such as "issue #16978 review follow-up" (lines ~150, ~913, ~970) and keep only the current invariant description.

## 7. Per-image WARNING log volume
Each dropped image emits its own `WARNING`. Large documents could produce dozens of lines per parse; the final summary log already aggregates this. Consider downgrading per-image warnings or summarizing the count to avoid log spam.

Net: the change is additive and low-risk for behavior, but #1 makes the new metric inaccurate. Recommend fixing #1, #2, and #3 before merging.

Taranum01:
Thanks for the rigorous review, @xugangqiang. Force-pushed the fixes.

## 1. `images_described` overcount fixed (functional bug)
In `_transfer_to_tables` (deepdoc/parser/mineru_parser.py:1005), the increment was BEFORE the readability check, so an IMAGE block with a `vlm_description` but a missing/corrupt binary was counted as described yet never emitted to tables. Moved the increment AFTER the `if image is None: ... continue` block, gated on a `has_vlm_description` local set earlier in the loop. Now only blocks that actually reach `tables.append(...)` are counted.

## 2. Test updated
`test_transfer_to_tables_counts_unreadable_resource` previously asserted the buggy value (`images_described == 2`). Updated to `images_described == 1` — image A's vlm description is embedded in the section text in raw mode (counted in `_transfer_to_sections`), image B is unreadable so the `_transfer_to_tables` fix doesn't double-count it.

## 3. 'Closes #16978' reframed
Changed body header to 'Addresses #16978' and added explicit per-mode semantics in the description. This PR is observability, not a fix — embedded images are still dropped in raw mode. The real fix is #16978 item 2 (merge-granularity alignment), correctly out of scope.

## 4. Comment hygiene
Removed two review-history / compatibility-worded comments per AGENTS.md guidance:
  - 'without changing the existing return shape' (in `_transfer_to_sections`)
  - 'see issue #16978 review follow-up' (in `parse_pdf`)
Kept the contextual 'issue #16978' refs that document what the code does (e.g. test docstrings citing the bug the test pins).

## 5. Per-image log spam reduced
`_transfer_to_tables`'s 'Failed to load image' and 'Skip image without a readable resource' warnings were downgraded to DEBUG — the final INFO summary log already aggregates the unreadable-resource count, so the per-image warnings only contributed log spam. Kept 'Dropped embedded image' at WARNING — that one directly tells the operator that a specific image was lost, so it should stay loud.

## Not addressed (kept as-is)
- **#4 metric semantics** — documented in the new per-mode semantics section of the PR description.
- **#5 fragile shared counter** — restructuring `self.last_image_coverage` into a per-call explicit object is a larger refactor that would touch `parse_pdf`'s flow. The current single-counter shape is simple and tested; can be cleaned up in a follow-up.

`pytest test/unit_test/deepdoc/parser/test_mineru_parser.py` → 40 of my tests pass. The 2 remaining failures are pre-existing upstream `RagTokenizer.set_language` errors, unrelated to this change.

Force-pushed to branch `fix/16978-mineru-image-coverage` (commit `fdced982`).

Taranum01:
Thanks for the catch, @xugangqiang. Force-pushed the fix.

## Counter classification bug

In `_transfer_to_tables`, an app-media IMAGE with empty caption/footnote/vlm_description would still get routed to `images_unreadable_resource` if its binary was unreadable. The underlying failure was actually 'text was empty' (which `_transfer_to_sections` had already handled for app-media modes via the early-skip branch BEFORE the textless check), not 'binary was unreadable while text was there'.

The fix computes a `has_meaningful_text` local once (true if any caption/footnote/vlm_description contributed a non-whitespace string), then routes the unreadable branch to whichever counter matches the underlying failure mode:

  - `has_meaningful_text` → `images_unreadable_resource` (text was there, binary was unreadable)
  - text was empty → `images_dropped_no_text` (text was never going to make it, the section path skipped this for app-media modes BEFORE the textless check could fire)

## New regression test

`test_transfer_to_tables_counts_dropped_no_text_for_unreadable_textless_image` pins the fix: an IMAGE with empty caption/footnote/vlm_description and a missing binary must count as `images_dropped_no_text == 1` and `images_unreadable_resource == 0`.

41 of my tests pass. The 2 remaining failures are pre-existing upstream `RagTokenizer.set_language` errors, unrelated to this change.

Force-pushed to branch `fix/16978-mineru-image-coverage` (commit `0f1d43c1`).

Taranum01:
Hi @coderabbitai — this review appears to be re-running against an older commit. The fix for the exact issue raised here (counter classification for textless+unreadable images) is already on the current branch (commit `0f1d43c1`, 'fix(mineru): route textless+unreadable images to dropped_no_text (#16978 review)').

Verification on the current branch (commit `0f1d43c1`):
```
$ grep -n 'has_meaningful_text\|images_unreadable_resource' deepdoc/parser/mineru_parser.py
156:            "images_dropped_no_text": 0,
157:            "images_unreadable_resource": 0,
852:            "images_dropped_no_text": 0,
854:            "images_unreadable_resource": 0,
910:                    image_coverage["images_dropped_no_text"] += 1
985:            has_meaningful_text = any(t.strip() for t in texts)
``"

The branch has the exact `has_meaningful_text` classifier the review asked for, and the routing comment at lines 999-1004 documents the dispatch:
```python
if has_meaningful_text:
    self.last_image_coverage['images_unreadable_resource'] += 1
else:
    self.last_image_coverage['images_dropped_no_text'] += 1
``"

And `test_transfer_to_tables_counts_dropped_no_text_for_unreadable_textless_image` is the regression case the review asked for. 41 of my tests pass; the 2 remaining failures are pre-existing upstream `RagTokenizer.set_language` errors, unrelated to this change.

If this re-run was meant to flag something genuinely new, please point me at the file:line.

xugangqiang:
## Review findings

Scope is honestly declared as **observability, not a fix** — the actual silent-drop bug (#16978 item 2) is correctly left out of scope. Counting logic was traced by hand and is correct. A few issues worth addressing before merge:

### 1. Misleading intermediate INFO log for app-media modes (should fix before merge)
`_transfer_to_sections` emits `[MinerU] image_coverage detected=… chunked=0 described=0 dropped_no_text=0 unreadable=0` whenever `images_detected > 0` (around `deepdoc/parser/mineru_parser.py:932-940`). But for `naive`/`manual`/`paper`, `chunked`/`described`/`unreadable` are only populated later by `_transfer_to_tables`. So the same parse produces two `image_coverage` lines — the first says `chunked=0`, the final `image_coverage final` line (`:1163`) reports the real `chunked=N`. Operators grepping logs will see contradictory numbers.
**Suggestion:** drop the intermediate INFO line; the final summary is complete and accurate for all modes and loses no actionable info (no `page_idx`/`img_path` is on that line).

### 2. Shared mutable `self.last_image_coverage` mutated across two methods is fragile
The metric relies on `_transfer_to_sections` writing the dict and `_transfer_to_tables` then mutating the same instance dict, using an increment-then-decrement trick (`chunked += 1` then `-= 1` on unreadable). The test `test_transfer_to_tables_counts_unreadable_resource` even documents this as "current chunk count not surviving chunk count … for now this test pins the current shape". This is error-prone.
**Suggestion:** have each transfer method `return` its own coverage dict and merge them in `parse_pdf`. Removes the cross-method mutable state.

### 3. App-media textless-but-readable images are reported as "healthy" (coverage blind spot)
In `naive`/`manual`/`paper`, an image whose binary is readable but has no caption/footnote/VLM still gets `images_chunked += 1` and emits `((image, [""]), positions)` (`:1023`), with `dropped` left at 0. So an image carrying no retrievable text is counted as chunked/healthy. By-design per the PR description, but it hides the same loss the PR is trying to surface.
**Suggestion (follow-up):** consider flagging these separately (e.g. an `images_chunked_empty_text` counter) so coverage reflects reality.

### 4. Minor nits
- The `WARNING "Dropped embedded image"` fires once per textless image in `raw` mode (`mineru_parser.py:912`); acceptable, but noisy for image-heavy docs.
- `self.last_image_coverage` is only ever consumed inside `parse_pdf` itself (log-only). The "stamp so callers can inspect" framing is aspirational; harmless but redundant as an instance attribute.

### Verification note
Counting was verified by tracing the code paths (e.g. `test_transfer_to_tables_counts_unreadable_resource` expecting `described==1`, `chunked==3` matches the implementation). The full `pytest` suite was not executed in this review (no `.venv` in the read-only checkout) — please confirm the "40 pass / 2 pre-existing `RagTokenizer.set_language` failures" claim in CI before merging.

JinHai-CN:
@Taranum01 would you please fix the conflicts and rebase the code?

Taranum01:
Conflicts resolved and rebased against current main. The only conflict was in `deepdoc/parser/mineru_parser.py` where main had picked up the CHART (`#19096`) handling since this PR was opened: kept the new CHART case branches (both in `_transfer_to_sections` and `_transfer_to_tables`) and preserved the image_coverage / dropped_no_text / images_unreadable_resource additions from this PR on top.

Local unit tests for `test_mineru_parser.py`:
- 51 pass (including all image_coverage / _transfer_to_sections / _transfer_to_tables cases from this PR)
- 3 pre-existing failures unrelated to this PR: `test_transfer_to_tables_emits_ordered_typed_media`, `test_tokenize_table_uses_payload_type_instead_of_html_content` (both `RagTokenizer.set_language` AttributeError), and `test_media_context_preserves_image_payload_type` (grpcio-status ImportWarning). I verified these fail identically on `upstream/main` without any of my changes.

Diff is now mergeable. Ready for re-review.

xugangqiang:
Code review findings:

**Improvement — CHART blocks pollute the `image_*` counters (`deepdoc/parser/mineru_parser.py:1017` + `:879`)**
`images_chunked += 1` runs for both `IMAGE` and `CHART` (they share the `_transfer_to_tables` path), so a `CHART` whose binary is unreadable gets mislabeled under `images_unreadable_resource` / `images_dropped_no_text` as if it were an image. Separately, `images_detected` (line 879) counts only `IMAGE`, not `CHART`, while `chunked`/`described` count both — so the final `detected=N chunked=M` summary is asymmetric. Suggest aligning the counting scope (either count CHART in `detected` too, or rename to `visual_coverage` and exclude CHART from the image counters). No regression test covers the CHART path, which is why this slipped through.

**Nit — review-history comments not fully removed (AGENTS.md violation)**
The PR description claims review-history wording was removed, but the test file still carries many docstrings like "Regression for xugangqiang's review on #16978" / "Per CodeRabbit review on #16978 follow-up", and the production comment at lines 964–975 narrates a *removed* intermediate log (stale comment describing a superseded design). AGENTS.md explicitly says to drop these.

**Nit — fragile `images_chunked` increment-then-decrement (`mineru_parser.py:1017` / `:1051`)**
The counter is `+= 1` per block then `-= 1` on unreadable. The author's own test admits it is "current chunk count, not surviving chunk count". Works in production but is brittle; prefer counting only on confirmed emit.

**Nit — WARNING per imageless embedded image in raw mode (`mineru_parser.py:944`)**
Every embedded image without caption/footnote/VLM description emits a `WARNING`. For PDFs with many decorative images this can flood logs; consider `DEBUG` or rate-limiting.

Functional correctness is sound and all prior review threads are addressed; the above are maintainability/consistency items to tidy before merge.

Taranum01:
All four review items addressed in this force-push (rebased onto current main, +133/-60 across 2 files, `mergeable: MERGEABLE`):

1. **CHART counter scope aligned.** `images_detected` now includes CHART (was IMAGE-only while `images_chunked` and `images_described` already counted CHART via `_transfer_to_tables`). CHART `vlm_description` on the raw-mode text path now increments `images_described` (mirrors the IMAGE branch). CHART textless drops in raw mode now route to `images_dropped_no_text` the same way IMAGE does. Three new tests pin the CHART path: `test_images_detected_counts_chart_blocks`, `test_transfer_to_tables_counts_chart_chunked_and_described`, `test_transfer_to_sections_counts_chart_described_in_raw_mode`.

2. **Review-history wording removed.** Dropped "Regression for xugangqiang's review on #16978" / "PR #16978 review follow-up" / "PR #16978 review" phrasing from all 7 test docstrings and 2 inline comments in `_transfer_to_sections` (the "intermediate INFO log" block that narrated the removed design). Tests still assert the same behaviors — they just don't carry the review-thread breadcrumbs anymore. AGENTS.md: "Drop stale comments that describe a superseded design."

3. **Increment-then-decrement refactored to count on confirmed emit.** `_transfer_to_tables` no longer `images_chunked += 1` upfront and unwinds on the unreadable path. Chunked and described are now only incremented when the image actually reaches the `tables` list. The dropped/unreadable routes only fire from the early-return branch. `test_transfer_to_tables_counts_chunked_and_described` still pins the same final numbers.

4. **Per-image WARNING downgraded to DEBUG.** PDFs with many decorative images were flooding the log. The aggregate `images_dropped_no_text` count is still emitted at the right level by the `parse_pdf` `image_coverage final ...` summary line. Tests updated to capture at `caplog.at_level(logging.DEBUG, ...)`.

Tests: `pytest test/unit_test/deepdoc/parser/test_mineru_parser.py -v` → 54 passed (51 pre-existing + 3 new). The 3 pre-existing failures (RagTokenizer.set_language, grpcio-status) are unrelated to this change.

xugangqiang:
Reviewing the current head, I found a few issues beyond the review threads already addressed. None are blockers (no crash/data-loss/API break), but one undercuts the metric's purpose:

**1. (Improvement) `raw` mode under-counts CHART in `images_chunked` — false "dropped" signal**
`deepdoc/parser/mineru_parser.py:958` increments `images_chunked` only for `MinerUContentType.IMAGE`, but `:877-878` already counts CHART in `images_detected`, and `:909-914`/`:966` do emit a CHART's caption/footnote as a real text section in `raw` mode. So a chart-only `raw` PDF reports `detected > 0, chunked = 0` even though the chart was emitted — the headline `detected − chunked` gap this PR adds fires a systematic false positive for charts. Fix is tiny: extend the `images_chunked` condition at `:958` to `{IMAGE, CHART}` (mirrors `images_described` at `:913`), and add a raw-mode CHART-chunked test. Note `_transfer_to_tables` `images_chunked` (`:1048`) already includes CHART, so only the raw path is missing.

**2. (Improvement) `test_transfer_to_tables_counts_unreadable_resource` pins a non-production counter value**
It calls `_transfer_to_sections(outputs, parse_method="raw")` and then `_transfer_to_tables`, asserting `images_chunked == 3`. In production `raw` mode `_transfer_to_tables` is never called (`mineru_parser.py:1184`), and app-media skips images in `_transfer_to_sections`, so this raw+`_transfer_to_tables` combination never occurs. The `chunked == 3` value (double-counted) can't happen on any real path (the test's own comment admits "current chunk count, not surviving chunk count"). Suggest splitting into realistic raw (sections only) vs app-media (both, images skipped) assertions.

**3. (Nit) Fragile cross-method shared mutable state**
`self.last_image_coverage` is mutated by both `_transfer_to_sections` and `_transfer_to_tables`, with the division of labor gated implicitly by `parse_method` via the early-skip branch at `:883-884`. Works today because `parse_pdf` (`:1183-1184`) calls them in the right order, but any out-of-order/mismatched call corrupts counters. Consider aggregating coverage as a local dict inside `parse_pdf` and writing it once.

**4. (Nit) `vlm_configured` key absent from `__init__` and `_transfer_to_sections` dict**
`vlm_configured` is only injected in `parse_pdf` (`:1192`); `mineru_parser.py:157-163` and `:864-870` don't include it. Any caller reading `last_image_coverage["vlm_configured"]` after `_transfer_to_sections` but without `parse_pdf` would `KeyError`. Suggest pre-seeding it in both dicts.

---

Question for @Taranum01: do you prefer to **merge this PR as-is and follow up with a separate PR** for the above, or **fold the fixes (at least #1) into this PR** before merge? Either is fine with me — just want to confirm your preference.

Taranum01:
Folding #1, #2, and #4 into this PR now — they're each small, the tests pin real production paths, and #1 is a genuine false-positive in the headline metric that defeats the purpose of #16978's coverage stamp. Deferring #3 (the cross-method shared state refactor) to a follow-up PR: it touches parse_pdf and is bigger than the rest combined, and the implicit parse_pdf ordering is enforced by the call site that already runs.

Will push once the three fixes + their regression tests are green.

Taranum01:
Folding #1, #2, #4 into this PR now; #3 (cross-method shared state refactor) deferred to a follow-up since it touches parse_pdf and the implicit ordering is enforced by the call site that already runs. Pushed in this force-push:

1. **CHART now counts toward images_chunked in raw mode.** `mineru_parser.py:~970` condition now matches `{IMAGE, CHART}` (mirrors the scope at `images_described`:913 and at `images_chunked` in `_transfer_to_tables`). New `test_raw_mode_counts_chart_block_as_chunked` pins a chart-only raw PDF producing detected=1 / chunked=1 / dropped_no_text=0 — the false-positive asymmetry is gone.

2. **Split the raw + _transfer_to_tables fixture into two realistic tests.** `_transfer_to_tables` is only called by `parse_pdf` in app-media modes (`parse_method in {naive, manual, paper}`); the original fixture tested a path parse_pdf never takes. Replaced with:
   - `test_raw_mode_counts_unreadable_resource_in_sections` — raw path only; `images_chunked=1` (section was emitted) and `images_unreadable_resource=0` (no binary read happened in raw mode).
   - `test_app_media_mode_counts_unreadable_resource` — naive path with both methods called; `images_chunked=0` (image didn't reach tables), `images_unreadable_resource=1` (missing file routes here when text was non-empty).
   - `test_app_media_mode_counts_dropped_no_text_for_unreadable_textless_image` — textless + unreadable routes to `images_dropped_no_text`, `images_chunked=0` on confirmed-emit refactor.
   The old `chunked == 3` assertion was the artifact of the increment-then-decrement pattern, which the previous push already replaced with count-on-confirmed-emit.

3. **Pre-seeded `vlm_configured=False`** in both coverage dicts (`MinerUParser.last_image_coverage` in `__init__`, and the local `image_coverage` dict in `_transfer_to_sections`). `parse_pdf` still overwrites it with the real vision_model status. New `test_initial_last_image_coverage_has_vlm_configured_key` pins the constructor pre-seed.

Tests: `pytest test/unit_test/deepdoc/parser/test_mineru_parser.py -v` → **57 passed** (54 pre-existing + 3 new). The 3 pre-existing failures (RagTokenizer.set_language, grpcio-status) are unrelated to this change.

Diff: +122 / -68 across 2 files. `mergeable: MERGEABLE`.

xugangqiang:
LGTM

PR #18767, opened 2026-08-26
Fix: MinerU model

Lynn-Inf:
### Summary

Fix model_types in MinerU.net, and add default model for MinerU.

### DIFF
--- a/conf/models/mineru.json
+++ b/conf/models/mineru.json
@@ -12,13 +12,13 @@
     {
       "name": "vlm",
       "model_types": [
-        "doc_parse"
+        "ocr"
       ]
     },
     {
       "name": "MinerU-HTML",
       "model_types": [
-        "doc_parse"
+        "ocr"
       ]
     }
   ]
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
LGTM. Verified the fix is correct:

- `LLMType` only defines `chat` / `embedding` / `asr` / `vision` / `rerank` / `tts` / `ocr` — there is no `doc_parse` model type, so `model_types: ["doc_parse"]` left MinerU without a valid type.
- MinerU is consistently treated as an OCR model across the codebase (`MinerUOcrModel` in `rag/llm/ocr_model.py`, `get_first_provider_model_name(tenant_id, "MinerU", LLMType.OCR)`, and `tenant_llm_service` querying `model_type=LLMType.OCR.value`), so `"ocr"` is the correct type.
- Matches the PaddleOCR convention (`conf/models/paddleocr.json` and `paddleocr_local.json` both use `"ocr"`).
- `mineru_local.json` now exposes a default model with `model_types: ["ocr"]`, consistent with other local-class configs (e.g. `funasr.json`, `paddleocr_local.json`).

Minor / non-blocking:

- `"MinerU-model"` is a generic default label; fine, but if the local MinerU deployment has a canonical model name, consider using it for display consistency.
- Good that the missing trailing newline on `mineru_local.json` was fixed.

Thanks!

PR #18768, opened 2026-08-26
Refactor: Add rerank_candidates_count, knn top_k and num_candidates to /api/v1/retrieval

wangq8:


wangq8 (commented):
This is an AI review comment.

Thanks for this refactor — exposing `knn_top_k`, `knn_num_candidates`, `rerank_candidates_count`, and `include_knowledge_compilation` with `top_k` kept as a backward-compatible alias is a clean approach, and the precedence logic (`knn_top_k` wins when both are provided) matches the docs.

A few issues to consider before merge:

**1. `rerank_candidates_count` validation gap → 500 instead of a clean error (important)**
The API only checks `rerank_candidates_count <= 0`, but `search.py:retrieval()` raises a plain `Exception` when `page * page_size > rerank_candidates_count`:

```python
if page * page_size > rerank_candidates_count:
    raise Exception(f"rerank_candidates_count({rerank_candidates_count}) must be greater than page * page_size({page * page_size}) ...")
```

That exception is swallowed by the generic `except Exception` in `chunk_api.py` and turned into `server_error_response(e)`, so e.g. `page_size=100` with the default `rerank_candidates_count=64` returns a 500 rather than a clear validation error. The docs already state "It must be at least `page` multiplied by `page_size`" — please enforce that at the API layer (return `get_error_data_result(...)`) instead of relying on the downstream exception.

**2. `include_knowledge_compilation` silently no-ops on Infinity engine**
`must_not={"exists": "compile_kwd"}` is only handled in `es_conn.py`, `gaussdb_conn.py`, `ob_conn.py`, `opensearch_conn.py`, and `serenedb_conn.py`. `infinity_conn.py` does not consume `must_not`, so `include_knowledge_compilation=false` has no effect on Infinity deployments. Unlike `knn_num_candidates` (documented as "currently applies only to Elasticsearch"), this engine limitation is not mentioned in the docs. Recommend either implementing the filter for Infinity or documenting the limitation.

**3. `knn_num_candidates` default can trigger a confusing error**
With the default `knn_num_candidates=2048`, a user who only sets `knn_top_k=4096` (and omits `knn_num_candidates`) gets "`knn_num_candidates` must be greater than or equal to `knn_top_k`" even though they never supplied it. Consider defaulting to `max(2048, knn_top_k)` or wording the error to mention the default value.

**4. Deprecation warning is invisible to API consumers (minor)**
`logging.warning("`top_k` is deprecated ...")` only surfaces in server logs. An API user passing `top_k` gets no indication in the response that it's deprecated. A response-level note (or at least a documented deprecation) would be friendlier.

**5. Non-numeric values → 500 (minor, pre-existing pattern)**
`int(req.get(...))` raises `ValueError` for non-numeric strings (e.g. `"knn_top_k": "abc"`), which again falls through to `server_error_response`. This mirrors the old `top_k` behavior, but since the PR is adding several new integer params, wrapping them in a small helper that returns `get_error_data_result` would improve robustness.

Nit: the docs table for `rerank_candidates_count` says "must be at least `page` multiplied by `page_size`" while the code only validates `> 0` — aligning these (per point 1) would remove the inconsistency.

This is an AI review comment.

wangq8 (commented):
This is an AI review comment.

Re-review after the latest changes. Three of the five issues from the previous review are now resolved; two remain, plus a few new observations (mostly test coverage).

## Resolved ✅
- `rerank_candidates_count < page * size` is now rejected at the API layer with a clean `102` instead of leaking the downstream `Exception` as a `500`.
- `knn_num_candidates` now defaults to `max(2048, knn_top_k)`, so a user setting only a large `knn_top_k` no longer gets a confusing "must be >= knn_top_k" error.
- Non-integer `knn_top_k` / `knn_num_candidates` / `rerank_candidates_count` now return `102` via `try/except` instead of a raw `ValueError` → `500`.

## Still open
**1. `include_knowledge_compilation=false` is a silent no-op on the Infinity engine (important)**
`must_not={"exists": "compile_kwd"}` is consumed by the ES / OpenSearch / OceanBase / SereneDB / GaussDB connectors, but `infinity_conn.py` never handles `must_not` (the only reference is a commented-out `# del condition["exists"]` in `update()`). On Infinity deployments the flag is silently ignored and knowledge-compilation chunks still surface. Unlike `knn_num_candidates`, this engine limitation is not documented — the reference text for `include_knowledge_compilation` makes an unconditional promise. Please either implement the filter for Infinity or document the limitation next to the parameter.

**2. Deprecation warning is invisible to API consumers (minor)**
`logging.warning("`top_k` is deprecated ...")` only reaches server logs; a client sending `top_k` gets no signal in the response. A response-level note (or at least accepting this is intentional) would be friendlier.

## New
**3. Missing tests for the new validation branches**
Only the `top_k alpha` case was updated. There are no tests for: `knn_num_candidates < knn_top_k`, `rerank_candidates_count < page * size`, `include_knowledge_compilation` non-boolean, or `knn_top_k` precedence over `top_k`. These are the exact branches this PR introduces, so they should be covered. (Codecov's "all lines covered" is comparing against a stale base.)

**4. `include_knowledge_compilation` uses strict `isinstance(x, bool)` while sibling flags don't (minor)**
`highlight` accepts `"true"`/`"false"` strings, and `use_kg`/`toc_enhance` are plain truthy `req.get(...)`. A client sending `1`/`0` or `"true"` for `include_knowledge_compilation` now gets a `102`. That's fine as a strict contract, but it's inconsistent with the neighboring booleans — worth aligning intentionally rather than accidentally.

**5. `knn_num_candidates` is validated on all engines although it only applies to Elasticsearch (minor)**
The docs say the parameter "currently applies only to Elasticsearch," yet the `knn_num_candidates >= knn_top_k` check runs unconditionally. On Infinity/OceanBase/etc. a value below `knn_top_k` would be rejected even though it's ignored downstream. Low impact, but the validation and the documented scope are out of sync.

Nit: the deprecation warning still fires when a client supplies *both* `top_k` and `knn_top_k`, so the "use `knn_top_k` instead" hint reads oddly in that case.

This is an AI review comment.

PR #18769, opened 2026-08-26
fix(retrieval): keep file-filter chunk counts consistent in retrieval testing

euvre:
### Summary

In knowledge-base retrieval testing, the per-file chunk counts shown in the "File" filter popup did not match the results returned after applying the filter (reported: a file listed with 23 chunks returned 90 once it was the only one selected; the popup also collapsed to the single selected file).

Root cause: `doc_aggs` and `total` are derived from the threshold-valid subset of the rerank candidate window. When `doc_ids` was passed, the engine query itself was re-scoped to those documents, so the candidate window was refilled by the selected documents alone. The numbers shown before filtering (a document's share of the unfiltered window) and after filtering (the document's own window-filling count) therefore described two different result universes and could not match.

This PR adds an opt-in `doc_ids_as_filter` request flag for UI file filters on `POST /api/v1/datasets/search`:

- When the flag is set (web retrieval-testing and search file filters now always set it), the engine query stays unscoped, so the candidate window is identical to the unfiltered run. `doc_ids` then narrow the threshold-valid set that backs the returned page and `total`, while `doc_aggs` are always computed over the unfiltered window.
- Result: selecting files returns exactly the counts the popup showed, `total` equals the sum over the selected files, and the popup keeps listing all files with stable counts so filters can be switched without clearing.
- When the flag is absent, behavior is unchanged: `doc_ids` still scopes the engine query (SDK `document_ids`/`doc_ids` scoping semantics are preserved).

Implemented symmetrically in the Go backend (`internal/service/nlp/retrieval.go`, request plumbing in `internal/service/dataset/*`) and the Python backend (`rag/nlp/search.py`, `api/apps/services/dataset_api_service.py`); the web client sends the flag from `kbService.retrievalTest`.

### Verification

- Go unit test `TestRetrievalDocIDsAsFilterKeepsCountsConsistent` (in-memory sqlite + fake engine) pins: unfiltered totals/counts, that the flagged run does not push `doc_id` into the engine query, that its `total` equals the count `doc_aggs` showed for the selected document, that `doc_aggs` still describe the unfiltered window, and that a non-flagged run still scopes the engine query. `bash build.sh --test ./internal/service/nlp/... ./internal/service/dataset/...` passes.
- Live check against the local Go API: with `doc_ids_as_filter: true`, the filtered run returns `total` equal to the file's unfiltered `doc_aggs` count and a stable full `doc_aggs` list; without the flag the response is unchanged from before.
- Note on scope: counts remain bounded by the rerank candidate window (`rerank_candidates_count`, 64-256) by design — `total` semantics ("threshold-valid matches across the candidate set", settled in #18071/#18750) are intentionally preserved; this PR only makes the file filter consistent with that universe.

### DIFF
--- a/api/apps/services/dataset_api_service.py
+++ b/api/apps/services/dataset_api_service.py
@@ -1058,6 +1058,7 @@ async def search(dataset_id: str, tenant_id: str, req: dict):
     rerank_candidates_count = int(req.get("rerank_candidates_count", 64))
     question = req.get("question", "")
     doc_ids = req.get("doc_ids", [])
+    doc_ids_as_filter = bool(req.get("doc_ids_as_filter", False))
     use_kg = req.get("use_kg", False)
     similarity_threshold = float(req.get("similarity_threshold", 0.0))
     vector_similarity_weight = float(req.get("vector_similarity_weight", 0.3))
@@ -1168,6 +1169,7 @@ async def search(dataset_id: str, tenant_id: str, req: dict):
         similarity_threshold,
         vector_similarity_weight,
         doc_ids=local_doc_ids,
+        doc_ids_as_filter=doc_ids_as_filter,
         knn_top_k=knn_top_k,
         knn_num_candidates=knn_num_candidates,
         rerank_mdl=rerank_mdl,
@@ -1444,6 +1446,7 @@ async def search_datasets(tenant_id: str, req: dict):
     question = req.get("question", "")
     doc_ids = req.get("doc_ids", [])
     use_kg = req.get("use_kg", False)
+    doc_ids_as_filter = bool(req.get("doc_ids_as_filter", False))
     similarity_threshold = float(req.get("similarity_threshold", 0.0))
     vector_similarity_weight = float(req.get("vector_similarity_weight", 0.3))
     knn_top_k = max(1, min(int(req.get("knn_top_k", 1024)), 2048))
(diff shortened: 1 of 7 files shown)


wangq8 (commented):
Thanks for this fix — the root-cause analysis is correct and the implementation is clean and symmetric across the Go and Python backends.

Verified against `~/qi/code/ragflow-main`:

- **Backward compatible:** `doc_ids_as_filter` defaults to false; when absent, behavior is unchanged (Go gates on `req.DocIDsAsFilter && len(req.DocIDs) > 0`, Python on `doc_ids_as_filter and doc_ids`).
- **Go aliasing is safe:** `aggsIdx := validIdx` keeps the unfiltered window because `filteredIdx` is a fresh slice and `validIdx` is only reassigned, never mutated in place.
- **Python early-return is equivalent when flag is off:** `if filtered_count == 0 and not aggs_idx` reduces to the original `filtered_count == 0`, so no behavior change for existing callers.
- **Plumbing is complete for both backends:** Go `SearchDataset` → `ToSearchDatasetsRequest` copies the field, and both Python `search` and `search_datasets` pass it through.
- **Good Go test coverage:** asserts the engine query stays unscoped under the flag, `total` equals the `doc_aggs` count shown for the selected doc, `doc_aggs` still describe the unfiltered window, and the non-flagged case still scopes the engine query.

Minor / non-blocking:

1. The shared search-bot path (`retrievalTestShare` → `/searchbots/retrieval_test`) does not send the flag, so the file-filter count inconsistency would remain there. Worth a follow-up (or explicitly confirming it's out of scope).
2. No Python-side test was added for the `rag/nlp/search.py` change. The Go test covers the semantics, but a small test for the mirrored Python path would guard against future drift.
3. Pre-existing, not introduced here: Go `doc_aggs` dedups by `docnm_kwd` (document name) rather than `doc_id`, so two documents sharing a name would merge in the aggregation.

LGTM otherwise.

wangq8:
by design.

Issue #18148, opened 2026-08-12
[Bug] Parsing-result chunk highlight missing/inaccurate for PDF (Python chunker drops overlap-head coordinates)


xugangqiang:
> **Internal discussion.** External contributors are welcome to share their opinions/analysis in the comments, but please do **not** open a pull request for this issue.

## Summary
In the parsing-result view (`/chunk/parsed/chunks`), clicking a chunk is expected to highlight the corresponding text on the PDF page. On the **Python** path the highlight is effectively missing / visually wrong.

## Root cause
Highlight boxes are derived from `position_int`, which is built from the per-segment `_pdf_positions` coordinate list via `finalize_pdf_chunk` (`rag/flow/parser/pdf_chunk_metadata.py:178`).

The Python `TokenChunker` merge (`rag/flow/chunker/token_chunker.py:_merge_text_chunks_by_token_size`) adds an **overlap prefix** (a duplicate of the previous chunk's tail) to the new chunk's text at the "start new chunk" boundary (`token_chunker.py:254-271`), but it does **not** carry the previous chunk's tail coordinates into `current[PDF_POSITIONS_KEY]`. Only the current unit's coordinates survive, so `finalize_pdf_chunk` emits `position_int` covering only the `cur` portion. The highlighted box is therefore shifted/truncated relative to the displayed text.

The legacy path (`rag/nlp/__init__.py:tokenize_chunks`) re-cropped coordinates from the final (overlap-inclusive) text via `pdf_parser.crop(..., need_position=True)`, so the overlap head was included and the highlight was accurate. The migration to structured-position merging dropped it.

## Notes
- The highlight pipeline (parser `normalize_pdf_items_metadata` → chunker `finalize_pdf_chunk` → storage → retriever `src` includes `position_int` at `rag/nlp/search.py:155` → `list_chunks` at `api/apps/restful_apis/chunk_api.py:516/544`) is otherwise intact in HEAD.
- Highlight is PDF-only by design (`web/src/components/document-preview/index.tsx:26-39`); the response model also 500s the endpoint if any `positions` entry is not length-5 (`chunk_api.py:113-118`) — both worth ruling out when reproducing.
[...]

PR #18772, opened 2026-08-26
fix(chunker): keep PDF positions on all cap-split sub-chunks so chunks don't lose their preview images

euvre:
### Summary

Parsing a text-heavy PDF with the built-in **book/paper** chunk methods left many chunks without their page-preview image: on the chunk list only the first chunk of a re-split group showed a thumbnail, and the continuation chunks also lost their position fields (no `page_num_int`, no document highlight).

**Root cause.** The title-family chunkers (used by the `book` and `paper` pipelines, and shared by `laws`) enforce `chunk_token_cap` (default 512, #18455/#18551). When a built chunk exceeds the cap it is re-split on sentence boundaries, and the split attached the merged `_pdf_positions` matrix to the **first sub-chunk only** ("Plan A"); continuation sub-chunks got an empty matrix. The on-demand PDF preview crop (`cropImageChunks` in Go / `restore_pdf_text_previews` in Python) only processes chunks that carry positions, so every continuation chunk ended up without `img_id` and without indexable positions. The Go TokenChunker's hard-cap expansion (general/naive, #18525) had the same drop.

**Fix.** Carry the source chunk's position matrix to **every** cap-split sub-chunk instead of only the first:

- Go: `internal/ingestion/component/chunker/title_cap.go` — `splitTitleChunkByCap` no longer empties `_pdf_positions`/`positions` on continuation sub-chunks (the now-unused `emptyPositionsLike` helper is removed).
- Go: `internal/ingestion/component/chunker/token.go` — `splitOversizedText` builds every sub-piece from a clone of the source unit, so all pieces keep its coarse PDF positions (and other item metadata), not just the first.
- Python (parity): `rag/flow/chunker/title_chunker/common.py` — `_split_text_chunk_by_cap` assigns the original positions to all sub-chunks.

The coordinates remain coarse (source-chunk level), the same carry-forward semantics already adopted for overlap heads in #18148. Difference from the precedents (#18455/#18551): instead of dropping positions on continuation sub-chunks, they are now shared, which is what lets every sub-chunk keep its preview image and highlight.

**Verification.**

- End-to-end on the Go pipeline (dev instance): uploaded a 7-page Chinese "law articles" PDF to a `book` dataset. Before: 23 chunks, only 2 carried `img_id`/positions. After rebuilding with this fix and re-parsing: **23/23 chunks have `img_id` and `page_num_int`**, and the chunk list shows a thumbnail on every card.
- Unit tests: `bash build.sh --test ./internal/ingestion/component/chunker/` (all pass, including the updated cap-split position tests); Python `pytest rag/flow/tests -k 'title or token_chunk'` (34 passed, incl. the updated plan-A test now asserting all sub-chunks keep coordinates). `ruff check`/`ruff format --check` clean.

The `paper` method was not re-run end-to-end separately; it uses the exact same TitleChunker + cap code path as `book` (only the level/hierarchy params differ), which is covered by the pipeline-level unit tests.

### DIFF
--- a/internal/ingestion/component/chunker/title_cap.go
+++ b/internal/ingestion/component/chunker/title_cap.go
@@ -22,8 +22,7 @@ package chunker
 // exceeds the cap is hard-split so the ceiling always holds. Table/image
 // chunks are atomic and never split. Every sub-chunk keeps the source chunk's
 // merged PDF position matrix so each one still gets its page-region preview
-// image and position highlight (the pre-fix behavior attached the matrix to
-// the first sub-chunk only, which left continuation chunks without an image).
+// image and position highlight.
 
 import (
 	"strings"
@@ -214,8 +213,10 @@ func splitTitleChunkByCap(chunk map[string]any, cap int) []map[string]any {
 }
 
 // shallowCopyChunk returns a shallow copy of c. Coordinate matrices
-// ([][]float64) are shared by reference, but splitTitleChunkByCap never
-// mutates them, so the source chunk is never aliased.
+// ([][]float64) are deliberately shared by reference between the source
+// chunk and its cap-split sub-chunks: nothing in the split or the downstream
+// read-only passes (on-demand crop, position indexing) mutates a matrix, so
+// the sharing is intentional and safe.
 func shallowCopyChunk(c map[string]any) map[string]any {
 	m := make(map[string]any, len(c))
 	for k, v := range c {
--- a/internal/ingestion/component/chunker/title_cap.go
+++ b/internal/ingestion/component/chunker/title_cap.go
@@ -20,12 +20,12 @@ package chunker
 // A built chunk that exceeds `chunk_token_cap` tokens is re-split on sentence
 // boundaries into <= cap sub-chunks; a single boundary-less run that still
 // exceeds the cap is hard-split so the ceiling always holds. Table/image
-// chunks are atomic and never split. The first sub-chunk keeps the source
-// chunk's merged PDF position matrix (Plan A); subsequent sub-chunks carry an
-// empty array.
+// chunks are atomic and never split. Every sub-chunk keeps the source chunk's
+// merged PDF position matrix so each one still gets its page-region preview
+// image and position highlight (the pre-fix behavior attached the matrix to
+// the first sub-chunk only, which left continuation chunks without an image).
 
 import (
-	"encoding/json"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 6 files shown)


wangq8 (commented):
Thanks — this fixes a real regression and the Go/Python parity plus updated tests are solid. I verified the crop pass (`cropImageChunks` / Python `build_pdf_position_fields`) only *reads* positions (it unmarshals into a fresh matrix / builds deep-copied lists), so carrying the source matrix onto every cap-split sub-chunk correctly restores preview images and `page_num_int`.

A few non-blocking observations to consider:

1. **Coarse-position semantics (intentional, but worth confirming).** Every continuation sub-chunk now inherits the *full* source-chunk position matrix, so their `position_int`/`page_num_int` and document highlight cover the entire source region (potentially multiple pages), not just that sub-chunk's own text. This is strictly better than the previous "no positions at all", and consistent with the overlap-head carry-forward from #18148, but the highlight will be over-broad for continuation chunks. If that turns out to be confusing in the search/highlight UX, a follow-up could try to apportion coordinates per sub-chunk.

2. **Python list aliasing (fragile, safe today).** `sub[PDF_POSITIONS_KEY] = orig_positions` assigns the *same* list object to every sub-chunk (shallow `dict(chunk)` + shared reference). It's safe today because downstream passes (`build_pdf_position_fields`, `extract_pdf_positions`, `merge_pdf_positions`) build fresh/deep-copied lists rather than mutating, but it's a latent footgun — any future in-place mutation would silently corrupt all sub-chunks. A `copy.deepcopy` (or at least a comment calling out the shared reference) would be more defensive.

3. **Go `shallowCopyChunk` doc comment is now self-contradictory.** It reads "shared by reference … so the source chunk is never aliased" — but after this change the matrix *is* shared across the source and all sub-chunks. Please reword to reflect that sharing is intentional and safe because the split never mutates the matrix.

4. **`token.go` metadata inheritance.** Every sub-piece now inherits *all* source fields via `cloneChunkDoc` (which deep-copies the `PDFPositions`/`Positions` byte slices — good). This includes `Image`; it's fine in the current pipeline because crop runs after the chunker and `ChunkOrderInt` is assigned later in `tokenizer.go`, but a pre-set `Image` on a text unit would now be duplicated onto every piece and skip their individual crop. Probably not reachable today, just noting the behavior change.

None of these block the fix. Nice work on the parity coverage.

xugangqiang:
LGTM

euvre:
Thanks @wangq8 for the careful review, and @xugangqiang for the LGTM. Addressed in 3072f997c:\n\n1. **Coarse-position semantics** — agreed, deferring as a follow-up. The cap-split stage has no sentence-to-coordinate mapping (positions are merged at source-chunk level before the split), so apportioning boxes per sub-chunk would need upstream plumbing; the full-matrix carry-forward keeps parity with the overlap-head carry-forward in #18148.\n\n2. **Python list aliasing** — applied: `_split_text_chunk_by_cap` now assigns `deepcopy(orig_positions)` to each sub-chunk, so no two sub-chunks share the position list object. Added `test_split_subchunk_positions_not_aliased`, which asserts every sub-chunk keeps the source coordinates, does not alias the source list, and that a nested mutation on one sub-chunk does not leak into the source or other sub-chunks.\n\n3. **Go `shallowCopyChunk` doc** — reworded; it now states that sharing the coordinate matrix between the source chunk and its cap-split sub-chunks is deliberate and safe because neither the split nor the downstream read-only passes (on-demand crop, position indexing) mutate a matrix. Also dropped the pre-fix-behavior remarks from the package comment and one test comment while in there.\n\n4. **`token.go` metadata inheritance** — no change. As you note, the crop pass runs after the chunker and assigns `Image` / `ChunkOrderInt` later, so a pre-set `Image` on a text unit is not reachable in the current pipeline; the behavior change is noted here for future awareness.\n\nVerification: `bash build.sh --test ./internal/ingestion/component/chunker/...` passes; `rag/flow/tests/test_title_chunker_token_cap.py` (14 passed, incl. the new test) and `test_title_chunker_position_int.py` pass. The one remaining failure in the token-cap suite (`test_hard_split_lossless_with_real_truncate`) is a pre-existing environment limitation on this machine — system python lacks tiktoken — and fails identically on the base commit. No shared-service end-to-end verification was run; this is based on unit tests + code-level inspection.

PR #18773, opened 2026-08-26
fix(web): cap embed dialog User ID input length and add data-type hint

euvre:
### Summary

The "User ID" input in the agent embed-website dialog (Management → Embed into webpage → Embed Setup) had no input length limit and no data-type hint: users could paste an arbitrarily long string (a reporter pasted hundreds of `w` characters), which was appended verbatim as the `userId` query parameter of the generated iframe embed URL. The shared agent page forwards that value as `user_id`, and the backend stores it in a `CharField(max_length=255)` column (`api/db/db_models.py`), so over-long values overflow the column and bloat the embed code with no feedback to the user.

This PR constrains and documents the field:

- Add `maxLength={255}` to the User ID input, matching the backend `user_id` column size, so typing or pasting stops at 255 characters.
- Add the same `max(255)` constraint to the dialog's zod form schema (`FormSchema`) as a programmatic backstop.
- Add a placeholder (`e.g. user-001`) and a label tooltip stating the expected data type (a plain-text end-user identifier) and the 255-character limit, so the field now carries an explicit data-type hint.
- Add the two new i18n keys (`chat.embedUserIdPlaceholder`, `chat.embedUserIdTooltip`) to every locale that already carries the embed-dialog strings (en, zh, fr, it, ja, ko, ru, tr).

Verification: verified end-to-end in the browser (agent canvas → Management → Embed into webpage): typing 300 characters into the User ID input keeps exactly 255, the generated embed URL carries a 255-character `userId` param, the placeholder is visible, and the tooltip renders the type/length hint. `oxlint` passes on the changed files. `npm run type-check` already fails on the base branch with ~208 pre-existing errors; none of them are on the lines touched by this PR (the embed-dialog errors at lines 306-316 are a pre-existing `useWatch` partial-type issue unrelated to this change).

### DIFF
--- a/web/src/components/embed-dialog/index.tsx
+++ b/web/src/components/embed-dialog/index.tsx
@@ -59,6 +59,8 @@ import { SwitchFormField } from '../switch-fom-field';
 import { useIsDarkTheme } from '../theme-provider';
 import { Input } from '../ui/input';
 
+const MAX_EMBED_USER_ID_LENGTH = 255;
+
 const FormSchema = z.object({
   visibleAvatar: z.boolean(),
   published: z.boolean(),
@@ -67,7 +69,7 @@ const FormSchema = z.object({
   enableStreaming: z.boolean(),
   muteWidget: z.boolean(),
   theme: z.enum([ThemeEnum.Light, ThemeEnum.Dark]),
-  userId: z.string().optional(),
+  userId: z.string().max(MAX_EMBED_USER_ID_LENGTH).optional(),
   widgetTitle: z.string(),
   widgetSubtitle: z.string(),
   widgetFooterText: z.string(),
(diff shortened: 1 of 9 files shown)


wangq8 (commented):
LGTM, thanks for the fix and the thorough verification notes.

Verified against the codebase:

- `Input` extends `React.InputHTMLAttributes<HTMLInputElement>`, so `maxLength`/`placeholder` are forwarded correctly.
- `RAGFlowFormItem` forwards `tooltip` through to `FormLabel`, so the tooltip wiring works.
- The zod backstop `z.string().max(255).optional()` and the HTML `maxLength` are complementary (paste is also capped by the browser), and the `max(255)` matches the backend `CharField(max_length=255)` on `user_id`.
- All 8 embed-dialog locales get the two new keys; translations read fine.

One minor, non-blocking note: HTML `maxLength` and zod `.max()` both count UTF-16 code units, whereas Django's `CharField(max_length=255)` counts Unicode code points. For astral characters (e.g. emoji) the frontend limit is effectively ~127 such characters — slightly stricter than the column. This is a reasonable conservative choice, just flagging the semantic mismatch in case you want to align them later.

dcc123456 (approved):

PR #18774, opened 2026-08-26
fix(ingestion): carry component outputs in the pipeline operation log DSL

euvre:
### Summary

Viewing a pipeline-parsed file's result from the dataset log page ("Logs → view result") showed a blank parse-summary/result panel and "0s" elapsed times on every node, even though the document's chunks were indexed and the chunks API returned them fine.

**Root cause (Go/Python parity gap)**

- The front-end dataflow-result page renders the middle/right panels exclusively from the log detail's `dsl.components[<id>].obj.params.outputs` (`web/src/pages/dataflow-result/parser.tsx`) and the per-node elapsed time from `params.outputs._elapsed_time.value` (`hooks.ts`).
- Python records the dataset parse operation log with `dsl=str(pipeline)` (`rag/svr/task_executor_refactor/dataflow_service.py`) — the canvas DSL *after* the run, where every component's `obj.params.outputs` carries its runtime outputs (`chunks`/`text`/…, `output_format`, `_elapsed_time`/`_created_time` via the component base class `set_output`).
- Go's `PipelineExecutor.Execute` stored the raw **static** DSL template, so `params.outputs` was absent and the panels rendered empty; the chunks API kept working because chunks are indexed independently — exactly the reported symptom.
- Additionally, `BuildDebugResultDSL` (also used by canvas-debug runs) dropped the `_elapsed_time`/`_created_time` bookkeeping keys that `runtime.TrackElapsed` stamps into every component output, so even debug-run timelines showed "0s" everywhere.

**Fix**

- `internal/ingestion/task/pipeline_executor.go`: `runPipelineWithDSL` now builds the run-result DSL via `BuildDebugResultDSL` right after `pipe.Run` — the only place the full `output["state"]` per-component map is still in scope, since the payload returned to `Execute` is already extracted — and returns its JSON as the log DSL, mirroring Python's `dsl=str(pipeline)`. `Execute` passes it straight to `recordPipelineLog`; on any build/marshal error it falls back to the static DSL so log recording never fails a run.
- `internal/ingestion/task/debug_result_dsl.go`: the outputs wrapper now also carries `_elapsed_time`/`_created_time` as plain `{value, type}` entries, and is emitted even for components with no recognized payload format (e.g. `File`), so the timeline shows real per-node elapsed times.

Difference from the earlier "View result blank tabs" fix: that one corrected `BuildDebugResultDSL` reading the run output at the wrong nesting (`output[id]` vs `output["state"][id]`) for canvas-debug runs. This PR fixes the *dataset parse log path*, which never built a result DSL at all, plus the missing elapsed-time bookkeeping.

**Verification**

- Unit tests (`bash build.sh --test ./internal/ingestion/task/...`): golden tests extended to pin `_elapsed_time`/`_created_time` passthrough (flat and nested-state shapes, bookkeeping-only components); new wiring test drives the real `runPipelineWithDSL` with stub ingestion components and asserts the returned log DSL carries `params.outputs.chunks` + `output_format` + `_elapsed_time` and unwraps the canvas envelope to the front-end shape.
- End-to-end on the local stack: re-parsed a dataset document; `GET /datasets/{id}/ingestions/{log_id}` now returns `dsl.components[*].obj.params.outputs` with the payload, `output_format` and bookkeeping keys, and the dataflow-result page renders the parse summary with real chunk text and per-node elapsed times (`File 0.001s → Parser_0 0.438s → Title Chunker_0 1.392s → …`) instead of a blank panel with "0s". Note: logs recorded before this fix keep the static DSL; only newly recorded parse logs carry outputs.

### DIFF
--- a/internal/agent/runtime/helpers.go
+++ b/internal/agent/runtime/helpers.go
@@ -186,19 +186,11 @@ func WithTimeout(ctx context.Context, d time.Duration, fn func(ctx context.Conte
 	return nil
 }
 
-// monotonicEpoch is the fixed reference _created_time values are measured
-// against, mirroring Python time.perf_counter()'s arbitrary-but-constant
-// epoch. Captured once per process.
-var monotonicEpoch = time.Now()
-
-// TrackElapsed records the duration of fn and stamps the
+// TrackElapsed records the wall-clock duration of fn and stamps the
 // output map with two synthetic keys mirroring Python `ProcessBase`
-// (rag/flow/base.py:42,55), which stamps time.perf_counter() floats:
+// (base.py:42, 58):
 //
-//	"_created_time"  float64 seconds on the monotonic clock (relative to
-//	                 monotonicEpoch) captured BEFORE fn runs — Python's
-//	                 perf_counter analogue, so both bookkeeping keys share
-//	                 one numeric type across backends.
+//	"_created_time"  RFC3339Nano-formatted timestamp taken BEFORE fn runs.
 //	"_elapsed_time"  float64 seconds (with sub-second precision) that
 //	                 fn took to complete, in [0, +∞).
 //
@@ -223,7 +215,7 @@ func TrackElapsed(name string, fn func() (map[string]any, error)) (map[string]an
 		out = make(map[string]any)
 	}
 	if _, ok := out["_created_time"]; !ok {
-		out["_created_time"] = start.Sub(monotonicEpoch).Seconds()
+		out["_created_time"] = start.UTC().Format(time.RFC3339Nano)
 	}
 	if _, ok := out["_elapsed_time"]; !ok {
 		out["_elapsed_time"] = elapsed.Seconds()
--- a/internal/agent/runtime/helpers.go
+++ b/internal/agent/runtime/helpers.go
@@ -186,11 +186,19 @@ func WithTimeout(ctx context.Context, d time.Duration, fn func(ctx context.Conte
 	return nil
 }
 
-// TrackElapsed records the wall-clock duration of fn and stamps the
+// monotonicEpoch is the fixed reference _created_time values are measured
+// against, mirroring Python time.perf_counter()'s arbitrary-but-constant
+// epoch. Captured once per process.
+var monotonicEpoch = time.Now()
+
+// TrackElapsed records the duration of fn and stamps the
 // output map with two synthetic keys mirroring Python `ProcessBase`
-// (base.py:42, 58):
+// (rag/flow/base.py:42,55), which stamps time.perf_counter() floats:
 //
-//	"_created_time"  RFC3339Nano-formatted timestamp taken BEFORE fn runs.
+//	"_created_time"  float64 seconds on the monotonic clock (relative to
+//	                 monotonicEpoch) captured BEFORE fn runs — Python's
+//	                 perf_counter analogue, so both bookkeeping keys share
+//	                 one numeric type across backends.
 //	"_elapsed_time"  float64 seconds (with sub-second precision) that
 //	                 fn took to complete, in [0, +∞).
 //
@@ -215,7 +223,7 @@ func TrackElapsed(name string, fn func() (map[string]any, error)) (map[string]an
 		out = make(map[string]any)
 	}
 	if _, ok := out["_created_time"]; !ok {
-		out["_created_time"] = start.UTC().Format(time.RFC3339Nano)
+		out["_created_time"] = start.Sub(monotonicEpoch).Seconds()
 	}
 	if _, ok := out["_elapsed_time"]; !ok {
 		out["_elapsed_time"] = elapsed.Seconds()
--- a/internal/agent/runtime/helpers_test.go
+++ b/internal/agent/runtime/helpers_test.go
@@ -278,15 +278,12 @@ func TestTrackElapsed_AddsCreatedAndElapsedFields(t *testing.T) {
 	if _, ok := got["_created_time"]; !ok {
 		t.Fatal("result missing _created_time")
 	}
-	// Python parity: ProcessBase stamps time.perf_counter() (a float), so
-	// _created_time must be float64 seconds on the monotonic clock rather
-	// than a wall-clock string — both bookkeeping keys share one type.
-	ct, ok := got["_created_time"].(float64)
-	if !ok {
-		t.Fatalf("_created_time = %v (type %T), want float64", got["_created_time"], got["_created_time"])
+	ct, ok := got["_created_time"].(string)
+	if !ok || ct == "" {
+		t.Fatalf("_created_time = %v (type %T), want non-empty string", got["_created_time"], got["_created_time"])
 	}
-	if ct < 0 {
-		t.Errorf("_created_time = %f, want >= 0", ct)
+	if _, err := time.Parse(time.RFC3339Nano, ct); err != nil {
+		t.Errorf("_created_time %q is not RFC3339Nano: %v", ct, err)
 	}
 	elapsed, ok := got["_elapsed_time"].(float64)
 	if !ok {
--- a/internal/agent/runtime/helpers_test.go
+++ b/internal/agent/runtime/helpers_test.go
@@ -278,12 +278,15 @@ func TestTrackElapsed_AddsCreatedAndElapsedFields(t *testing.T) {
 	if _, ok := got["_created_time"]; !ok {
 		t.Fatal("result missing _created_time")
 	}
-	ct, ok := got["_created_time"].(string)
-	if !ok || ct == "" {
-		t.Fatalf("_created_time = %v (type %T), want non-empty string", got["_created_time"], got["_created_time"])
+	// Python parity: ProcessBase stamps time.perf_counter() (a float), so
+	// _created_time must be float64 seconds on the monotonic clock rather
+	// than a wall-clock string — both bookkeeping keys share one type.
+	ct, ok := got["_created_time"].(float64)
+	if !ok {
+		t.Fatalf("_created_time = %v (type %T), want float64", got["_created_time"], got["_created_time"])
 	}
-	if _, err := time.Parse(time.RFC3339Nano, ct); err != nil {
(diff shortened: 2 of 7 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the thorough fix and the solid test coverage. The root-cause analysis is correct: the dataset parse path never built a result DSL at all (unlike the canvas-debug path), so the persisted log carried the static DSL and the front-end `dataflow-result` page — which renders entirely from `dsl.components[<id>].obj.params.outputs` and `outputs._elapsed_time.value` — showed blank panels and "0s" everywhere. Building the result DSL right after `pipe.Run` (where `output["state"]` is still in scope) and returning it as the log DSL mirrors Python's `dsl=str(pipeline)` correctly, and the wiring test that drives the real `runPipelineWithDSL` with stub components is a strong regression guard.

A few observations:

1. **Top-level DSL fields are dropped (parity gap with Python).** `BuildDebugResultDSL` returns only `{"components", "graph"}`, whereas Python's `Graph.__str__` (agent/canvas.py:126) deep-copies *all* non-`components` top-level keys (`path`, `task_id`, `canvas_type`, …) and only reconstructs `components`. For the "View result" page this is fine — `useTimelineDataFlow` reads only `dsl.components` and `dsl.graph.nodes`. But the persisted operation-log DSL is now a strict subset of what Python stores. The front-end rerun flow round-trips this DSL (`useChangeChunkTextMode` builds `newDsl = {...data.dsl, components: {...}}` and POSTs it to `/agents/rerun`), so any consumer that reads `dsl.path`/`dsl.task_id`/etc. from the log now gets nothing. Today the Go `RerunAgent` handler only checks access, so nothing breaks immediately — but this is a latent divergence worth either closing (carry the remaining top-level keys like Python) or explicitly documenting as an intentional subset.

2. **`_created_time` type/semantics differ from Python (cosmetic).** Go's `TrackElapsed` stamps `_created_time` as an RFC3339 string, while Python's `ComponentParamBase.invoke` stamps `time.perf_counter()` (a float). The front-end only reads `_elapsed_time` (both sides agree on float seconds), so there's no visible impact today, but the two bookkeeping keys now have different types across backends. Worth aligning if anything later reads `_created_time`.

3. **`type` field is hardcoded `""`.** The outputs wrapper emits `{"value": v, "type": ""}`, whereas Python's `set_output` stores `str(type(value))`. This is pre-existing (the `detectFormat` branch already used `""`), so not introduced by this PR — just flagging for completeness since the payload now reaches the persistent log rather than only the debug END marker.

None of the above are blocking; the fix is correct, well-tested, and resolves the reported symptom.

This is an AI review comment.

euvre:
Thanks @wangq8 for the careful review — all three observations were valid, and all three are now applied in bf7f5fbac:

1. **Top-level DSL fields dropped** — fixed. `BuildDebugResultDSL` now carries every non-`components` top-level key (`graph`, `path`, `task_id`, `canvas_type`, …) deep-copied verbatim, mirroring `Graph.__str__` (agent/canvas.py:126-137) instead of returning only `{components, graph}`, so the persisted log DSL is no longer a strict subset and rerun-flow consumers keep every key. One boundary to note: Python refreshes `path`/`task_id` from live canvas state before serializing; the Go ingestion run state doesn't track those, so the values are carried from the static DSL as-is — exactly what the pre-PR log persisted, so nothing is lost on round-trip.

2. **`_created_time` type divergence** — aligned. `TrackElapsed` now stamps `_created_time` as float64 seconds on a fixed monotonic epoch (the `time.perf_counter()` analogue, rag/flow/base.py:42 / agent/component/base.py:417) instead of an RFC3339Nano string, so both bookkeeping keys share one numeric type across backends. Verified nothing consumed the string form: the front-end reads only `params.outputs._elapsed_time.value` (web/src/pages/dataflow-result/hooks.ts) and Go has no `_created_time` readers.

3. **`type` field hardcoded ""** — fixed. The outputs wrapper now emits Python-style `str(type(value))` strings (`<class 'str'>`, `<class 'list'>`, `<class 'float'>`, …) via a `pythonTypeName` mapping for the JSON-shaped run-output values, and `output_format` gained its `type` field like Python's `set_output` (agent/component/base.py:467-470).

Verification: `bash build.sh --test ./internal/ingestion/task/... ./internal/agent/runtime/...` — all packages ok. The builder tests (flat + nested-state), the real-run wiring test (`TestRunPipelineWithDSL_LogDSLCarriesOutputs`) and the `TrackElapsed` tests were extended to pin the carried top-level keys, the float `_created_time`, and the type strings. No shared-service end-to-end run on this line — the parity claims rest on these unit tests plus the line-by-line Python comparison cited above. Thanks also for confirming the root-cause analysis!

xugangqiang:
Reviewing the diff against trunk (parent of b1f97896) and the front-end consumers (hooks.ts / interface.ts / parser.tsx). The core fix is correct and resolves the blank "View result" panel + "0s" timeline. A few issues remain:

## Issues

### 1. `_created_time` global `TrackElapsed` change has the widest blast radius and is not required for this fix
`internal/agent/runtime/helpers.go` changes the `_created_time` value produced by `TrackElapsed` (used by every component runtime in both agent and ingestion paths) from an RFC3339 string to `time.Since(monotonicEpoch)` float seconds. This is wangq8's observation #2, explicitly described as "cosmetic... no visible impact today".

However, this change is **not necessary** for the bug being fixed:
- The front-end only reads `_elapsed_time.value` (`hooks.ts:263`); `interface.ts:15` types `_created_time` as `Record<string,any>` (the `{value,type}` wrapper), not a timestamp.
- Carrying `_created_time` into the outputs wrapper works regardless of its type — `pythonTypeName(v)` handles any value. If it stayed a string it would just be `{value:"<ts>", type:"<class 'str'>"}` and nothing breaks.

Recommendation: split the `_created_time` type alignment into its own PR so the behavioral change to a shared runtime contract is isolated from this bug fix. At minimum, add a regression test proving no consumer parses `_created_time` as a time string before merging.

### 2. The fallback path is untested
`runPipelineWithDSL` (pipeline_executor.go:978-990) falls back to the static `dsl` when `BuildDebugResultDSL` or `json.Marshal` fails, with the stated guarantee that "log recording never fails a run". No test exercises this fallback. Add a test pinning that a build/marshal error returns the original `dsl` rather than failing the run.

### 3. `pythonTypeName` diverges from Python's `str(type(v))` for typed containers
`debug_result_dsl.go` `pythonTypeName` handles `[]any`/`[]map[string]any` → `"<class 'list'>"`, but a typed slice like `[]int` falls through to the `default` branch and yields `<class '[]int'>`, whereas Python's `str(type(v))` always returns `list`/`dict`. This is cosmetic (only the `type` display string) but is a real parity divergence. Either map all slice/map kinds to `list`/`dict`, or add a test documenting the intended behavior.

### 4. Unaddressed CodeRabbit nits (still open after bf7f5fb)
- `debug_result_dsl_test.go` (~line 180, "bOutputs text check"): use a single-value assertion `bOutputs := bParams["outputs"].(map[string]any)` instead of `bOutputs, _ := ...` (discarding the second result).
- Comment wording: several comments carry Python-parity/compatibility prose; per repo guidance keep the `file:line` citations (they are the contract source) but trim the redundant prose.

## Minor
- `BuildDebugResultDSL` now runs unconditionally on every persist run (and on debug runs whose `logDSL` is discarded). Cost is one extra marshal per parse — acceptable, but could be gated on `ResultSink` hit or persist mode if desired.

euvre:
Hi @xugangqiang — thanks for the careful re-review; all four items are addressed in 87a266ec3:

1. **\`_created_time\` \`TrackElapsed\` change split out** — agreed, it was not required for this fix. Reverted: \`TrackElapsed\` is back to trunk's RFC3339Nano string, so this PR no longer touches the shared runtime contract at all (\`internal/agent/runtime/helpers.go\` / \`helpers_test.go\` are byte-identical to main again). The outputs wrapper now carries the string verbatim as \`{value:"<ts>", type:"<class 'str'>"}\`; the front-end reads only \`params.outputs._elapsed_time.value\` (hooks.ts), so rendering is unaffected. Hi @wangq8 — your observation #2 (aligning the \`_created_time\` type across backends) is hereby deferred to a dedicated PR where the blast radius of changing a shared runtime contract can be argued on its own; your note had marked it cosmetic/non-blocking, so I hope that trade-off is OK.

2. **Fallback path test** — the log-DSL decision is now extracted into \`PipelineExecutor.buildLogDSL\` (same code, testable seam), and \`TestBuildLogDSL_FallbackToStaticDSL\` pins both failure modes: a marshal failure (a NaN payload builds fine but \`json.Marshal\` rejects it) and a build failure (DSL without a \`components\` map) each return the original static DSL unchanged — log recording never fails a run.

3. **\`pythonTypeName\` typed containers** — fixed the divergence rather than only documenting it: after the scalar cases, all slice/array kinds map to \`<class 'list'>\` and all map kinds to \`<class 'dict'>\` (reflect-based), so \`[]int\` no longer yields \`<class '[]int'>\`. \`TestPythonTypeName\` is a new table test covering nil/bool/str/int/float, \`[]any\`, \`[]map[string]any\`, \`[]int\`, \`[2]int\`, \`map[string]any\`, \`map[string]string\`.

4. **CodeRabbit nits** — both applied: the \`tt, _ :=\` blank assignment is now a single-value assertion (S1005), and the Python-parity/compatibility prose in comments is trimmed while keeping the \`file:line\` citations (they are the contract source), per your refinement.

On the **Minor** point (gating \`buildLogDSL\` on a ResultSink hit or persist mode): leaving it ungated for now — the cost is one map walk plus one marshal per parse, and gating would fork the debug/persist code paths; happy to revisit if profiling ever shows it matters.

Verification: \`bash build.sh --test ./internal/ingestion/task/... ./internal/agent/runtime/...\` — all packages ok (\`task\`, \`task/indexdoc\`, \`task/tool\`, \`agent/runtime\`). No shared-service end-to-end verification on this line; the parity claims rest on these unit tests plus the Python sources cited in the comments.

xugangqiang:
lgtm

PR #18775, opened 2026-08-26
refactor(parser): consolidate vision figure enhancement across document formats

jay77721:
### Summary

Consolidates the duplicated and fragmented post-parsing vision figure enhancement logic across DOCX, Markdown, and PDF document formats into a single, unified component (`vision_enhancement.go`), matching Python RAGFlow's `enhance_media_sections_with_vision` architecture.

### Background & Motivation

Previously, vision figure enhancement was split across three separate dispatch files (`docx_vision_dispatch.go`, `markdown_vision_dispatch.go`, and `pdf_vision_dispatch.go`), leading to:
1. **>90% code duplication**: All three formats performed identical operations (filtering JSON items for `doc_type_kwd in {"image", "table"}`, resolving tenant `IMAGE2TEXT` model, launching concurrent VLM calls, and appending descriptions to `item["text"]`).
2. **Implicit cross-file dependencies**: `markdown` and `pdf` enhancement implicitly depended on unexported helpers defined in `docx_vision_dispatch.go`.
3. **Format discrepancies**: Markdown enhancement used `\n\n` separator while DOCX/PDF and Python used `\n`.
4. **Parser invocation clutter**: `ParserComponent.Invoke` required three sequential format-specific dispatch calls.

### Key Changes

1. **Unified Component (`vision_enhancement.go`)**:
   - Implemented `maybeDispatchVisionEnhancement(ctx, db, fileType, dispatched, inputs)` replacing all 3 legacy dispatchers.
   - Added `isVisionEnhancementAllowed(fileType)` with zero-allocation `switch` guard strictly matching Python's 3 call sites (`{PDF, DOCX, Markdown}`).
   - Added `cleanMarkdownBlock` to strip wrapping markdown code fences from VLM figure captions.
   - Improved concurrency model: acquire semaphore on the main thread before spawning goroutines, with `ctx.Done()` support for early cancellation.
   - Hoisted invariant prompt construction before the concurrent worker loop.
   - Unified text concatenation to `\n`, matching Python `f"{text}\n{parsed_text}"`.

2. **Clean Ingestion Parser Call (`parser.go`)**:
   - Collapsed the 3 sequential dispatch calls in `ParserComponent.Invoke` into a single call to `maybeDispatchVisionEnhancement`.

3. **Cleanup & Parity (`AGENTS.md`)**:
   - Removed `docx_vision_dispatch.go`, `docx_vision_dispatch_test.go`, and `markdown_vision_dispatch.go`.
   - Removed `maybeDispatchPDFVisionEnhancement` from `pdf_vision_dispatch.go`, while preserving full-page OCR parsing (`dispatchPDFVision` / `extractPDFVisionAnswer`) completely untouched.
   - Migrated all test coverage into `vision_enhancement_test.go` (covering JSON figure/table enhancement, allowlist filtering, non-JSON skipping, language propagation, silent fallback, data URI normalization, and markdown block stripping).

### Verification

- Unit tests passed: `bash build.sh --test ./internal/ingestion/component` (`ok ragflow/internal/ingestion/component 0.356s`)
- Parser tests passed: `bash build.sh --test ./internal/parser/...`
- Full Go server build passed: `bash build.sh --go`


jay77721:
### Staff Review Conclusion: Request Changes

This PR correctly consolidates the duplicated post-parsing vision enhancement across DOCX/Markdown/PDF (`>90%` identical: filter `{image,table}` + `IMAGE2TEXT` resolve + `sem=10` concurrency + `VLM` + `\n` append) into `vision_enhancement.go`, matching Python `enhance_media_sections_with_vision` (`rag/flow/parser/utils.py:162` @ `parser.py:772/978/1115`). Direction is right per `AGENTS.md: Prefer deletion` and net `-526/+268` is healthy. Must fix before merge:

**P0 — Full-page vision not cleaned (GObugs P0-3 only half-closed):**
`vision_enhancement.go:273` `extractVisionAnswer=cleanMarkdownBlock` but `pdf_vision_dispatch.go:657` `extractPDFVisionAnswer` is still `TrimSpace`-only. `dispatchPDFVision:605` keeps emitting ` ```markdown` fences into `text` for `parse_method=vision`.
*Fix:* delete `extractPDFVisionAnswer:657` and change `605` to `text := extractVisionAnswer(resp)` (no alias shim).

**Over-engineering — Dead WithContext branch:**
`vision_enhancement.go:48-49` `figureVisionPromptWithContextFile` + `buildFigureVisionPrompt:196` `hasContext` branch is never hit (`maybeDispatchVisionEnhancement:141` always `("","", lang)` with `context_size=0`). Keeps 15 lines + second cache entry for no caller.
*Fix:* delete constant/branch, simplify to `buildFigureVisionPrompt(lang string)` loading only `vision_llm_figure_describe_prompt.md` (re-add when context is actually needed).

**Half-baked cancellation:**
`vision_enhancement.go:152-156` `select{case sem<-: case <-ctx.Done(): wg.Wait(); return ctx.Err()}` but `parser.go:489` does `dispatched, _, _ = maybeDispatchVisionEnhancement(...)` discarding `error`. Cancellation is unobservable, partial `descriptions` are swallowed, and `wg.Wait()` in the `ctx` branch only waits for already-added goroutines.
*Fix:* either (A) remove `select` and rely on inner `visionChatInvoker(ctx,...)` (simplest, typical N<=10), or (B) propagate: `parser.go:489` → `if _,_,err:=maybeDispatch...; err!=nil{return nil,err}`.

**Minor:**
- `vision_enhancement.go:44` `var visionEnhancementConcurrency uint=10` is globally mutable without lock (race in `-parallel` tests). Make `const`.
- Tests `vision_enhancement_test.go` cover happy path (3 file types, allowlist, markdown skip, tenant missing, double-prefix, fence stripping) but miss `dispatched.Err!=nil`, model-resolve failure, non-string `image` field, and `N>10` / `ctx.Done` paths.

Overall value > cost, but the two P0/dead-code items should be addressed. Happy to LGTM after the small patch.

jay77721:
## Code Review Summary — PR #18775 (unified vision figure enhancement)

**Verdict: LGTM — approved with minor comment-level/nit fixes recommended before merge** (non-blocking).

This is a clean, low-risk deduplication refactor. I reviewed it read-only against an isolated checkout of the PR head and cross-checked behavior against the Python source on `main`.

### What’s done well ✅
- Consolidates 3 ~90%-duplicated dispatchers (`docx`/`markdown`/`pdf`) into one `maybeDispatchVisionEnhancement`, removing implicit cross-file dependencies.
- Fixes a real format discrepancy: Markdown used `\n\n` while DOCX/PDF used `\n`; now unified to `\n`, which correctly matches Python `enhance_media_sections_with_vision` (`f"{text}\n{parsed_text}"`).
- Genuine concurrency improvements:
  - Semaphore acquisition moved **before** spawning goroutines (was inside), so live goroutine count is now bounded by `visionEnhancementConcurrency`.
  - Added `ctx.Done()` cancellation support via `select`; early-return is safe (no partial writes, no goroutine leak).
  - Hoisted the prompt build out of the per-item goroutine loop (invariant on `language`) — correct and a real perf win.
- Test coverage is solid (image/table enhancement across all 3 formats, allowlist filtering, language propagation, `cleanMarkdownBlock` edge cases). Data-URI normalization is still covered at `buildVisionMessages` level.

### Items to address before merge

**1. Stale comment (should fix) — `pdf_vision_dispatch.go:729-732`**
References the removed `maybeDispatchPDFVisionEnhancement`. Update it to `maybeDispatchVisionEnhancement` (or drop the trailing clause).

**2. `cleanMarkdownBlock` provenance / intent (please confirm) — `vision_enhancement.go:59-74, 273-278`**
The Go code now applies `cleanMarkdownBlock` to all figure descriptions. However, Python’s `enhance_media_sections_with_vision` (`rag/flow/parser/utils.py`) only does `.strip()` — it does **not** call `clean_markdown_block`. That function actually lives in `common/string_utils.py` and is used by `rag/app/picture.py:198` (`describe_with_prompt`), which is the closer equivalent of the Go `visionChatInvoker` path.
- So this is a behavior change for DOCX/PDF vs. the old Go code (and vs. the Python parser path), though it aligns with the project’s `picture.py` convention and is generally desirable.
- Please either (a) update the PR description/comments to cite `rag/app/picture.py` for the clean step, or (b) if strict `enhance_media_sections_with_vision` parity is required, drop `cleanMarkdownBlock` and keep `strings.TrimSpace`.

**3. Test gaps (recommended)**
- The new `ctx.Done()` cancellation path and the `figureVisionPromptBuilder` error path are untested; please add cases and run `go test -race ./internal/ingestion/component/...`.
- Consider an assertion that an un-fenced plain-text VLM response is not truncated by `cleanMarkdownBlock`.

**4. Minor (optional)**
- `parser.go:489` ignores the error from `maybeDispatchVisionEnhancement` (including a possible `ctx.Err()`). Consistent with the “silently skip” philosophy, but worth a comment noting the error is intentionally dropped.

### Verification note
I could not run `go build`/`go test` in this sandbox (missing cgo system header `office_oxide.h`), so this is static review + Python cross-check + full reference scan (no dangling references to the removed functions were found). The PR author’s reported `build.sh --test/--go` results are the authoritative build signal.

Overall: good direction, safe to merge after the stale comment fix and a quick confirmation on the `cleanMarkdownBlock` intent.

jay77721:
## Code Review — PR #18775

**Verdict: LGTM / Recommend merge** (with minor non-blocking observations)

This PR consolidates three >90%-duplicated vision-figure enhancement dispatchers (DOCX / Markdown / PDF) into a single component `vision_enhancement.go`, removing cross-file implicit dependencies and making real improvements to the concurrency model and Python parity. Verified: zero regressions vs `main`; all vision tests pass.

---

### 1. Problem solved vs. claimed — Thoroughly solved

- Single entry `maybeDispatchVisionEnhancement` + `isVisionEnhancementAllowed` allowlist (`parser.py:772/978/1115`) replaces the three sequential dispatches in `parser.go`.
- Cross-file implicit deps eliminated: all helpers now live only in `vision_enhancement.go`; `markdown_vision_dispatch.go` deleted entirely, `maybeDispatchPDFVisionEnhancement` removed from `pdf_vision_dispatch.go` (full-page OCR path left untouched, as described).
- Separator unified to `\n`, matching Python `f"{text}\n{parsed_text}"`.

**Bonus parity fix I confirmed:** the old Go `extractDOCXVisionAnswer` only did `TrimSpace`, whereas Python's `vision_llm_chunk` (`rag/app/picture.py:198`) calls `clean_markdown_block` to strip ```markdown fences. The new `extractVisionAnswer` now strips fences for **all three formats**, closing a long-standing Go-vs-Python gap. `TestVisionEnhancement_EnhancesJSONImagesAndTables` verifies the cleaned output across all formats.

### 2. Approach — Optimal as-is

The three old functions were behaviorally identical, so consolidation is the only correct convergence. The semaphore-before-goroutine pattern + `ctx.Done()` cancellation is a genuine upgrade over the old code (which acquired `sem` *inside* the goroutine, briefly spawning N goroutines). Hoisting the prompt out of the loop is a clean invariant lift. Dropping the unused `setups` parameter from the old DOCX signature is a nice cleanup.

### 3. Test coverage & regression risk

- Good coverage: image+table enhancement (table-driven across 3 formats), allowlist filtering, non-JSON skip, language propagation, silent fallback (no tenant), data-URI double-prefix prevention, markdown-fence stripping incl. CRLF/nested/tabs/empty edge cases. Old tests were migrated, not dropped.
- **Verified zero regressions**: ran the full `internal/ingestion/component` suite on both the PR branch and `main` (same CGO env). The failure set is **identical** on both — the same 3 pre-existing extractor failures (`TestLlmtagChunk_MessageFit`, `TestFitExtractorMessages_RejectsEmptyUserTurn`, `TestFitExtractorMessages_RejectsSystemPromptLoss`), all unrelated to vision enhancement. Vision tests all pass.
- **Minor gap:** the new `ctx.Done()` early-return branch has no unit test. Low risk (simple path), but worth adding.
- **Behavior changes (correct, but worth flagging):** Markdown separator `\n\n` → `\n`, and new ```markdown fence stripping — both are parity-aligned with Python and will change re-parse output for existing docs.

### 4. Architectural value

Net −258 lines, single source of truth, eliminated implicit cross-file deps, stronger concurrency semantics, better Python parity. Localized, well-tested, no new public surface. Worth merging.

### 5. Suggestions (all non-blocking)

1. **`cleanMarkdownBlock` layering note** — Python calls it inside `vision_llm_chunk` (the invoker wrapper); Go calls it in `extractVisionAnswer` (post-invoker). Output is equivalent, and Go's placement keeps `visionChatInvoker` as a pure, swappable test seam. A one-line comment noting this intentional equivalence would help future readers.
2. **Test the `ctx.Done()` branch** — inject a cancelled `context` + a `visionChatInvoker` that returns `context.Canceled`, assert `handled == false` and the returned error.
3. **Test shared state** — `TestVisionEnhancement_EnhancesJSONImagesAndTables` shares one `invoker`/`capturedLanguage` across subtests. Works today (all use `lang="Japanese"`), but would break if a subtest changed language. Move invoker creation into the subtest to remove implicit coupling.

---
*Verification: built/tested in an isolated worktree from `refs/heads/refactor/vision-enforcement` (removed after review); confirmed the pre-existing C++ binding artifact was symlinked from the main repo to reproduce the real link environment, with no source changes. Reviewed Python references: `utils.py:enhance_media_sections_with_vision`, `figure_parser.py:VisionFigureParser.__call__`, `picture.py:vision_llm_chunk`, `string_utils.py:clean_markdown_block`.*

PR #18776, opened 2026-08-26
fix: allow empty GitHub repositories

buua436:

GitHub connector validation failed for empty repositories because it called `get_contents("")` and treated the resulting 404 as a missing repository.

The validation now only checks repository accessibility with `get_repo()`, allowing valid empty repositories to pass configuration validation.

### Type of change

### DIFF
--- a/common/data_source/github/connector.py
+++ b/common/data_source/github/connector.py
@@ -702,9 +702,8 @@ class GithubConnector(CheckpointedConnectorWithPermSyncGH[GithubConnectorCheckpo
                             continue
 
                         try:
-                            test_repo = self.github_client.get_repo(f"{self.repo_owner}/{repo_name}")
+                            self.github_client.get_repo(f"{self.repo_owner}/{repo_name}")
                             logging.info(f"Successfully accessed repository: {self.repo_owner}/{repo_name}")
-                            test_repo.get_contents("")
                             valid_repos = True
                             # If at least one repo is valid, we can proceed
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix — this correctly addresses the root cause: `Repository.get_contents("")` returns a 404 for an empty repository, and the validation path was misinterpreting that as "repository not found". Dropping the contents probe and relying on `get_repo()` is the right minimal fix, and the multi-repo and single-repo branches are now consistent.

One observation (non-blocking):

1. **Content-read access is no longer verified at validation time.** `get_repo()` only confirms the token can read repository *metadata*; it does not confirm the token can read *contents*. With a fine-grained PAT that grants "Metadata: read" but not "Contents: read", validation now passes but ingestion will fail later when the connector actually lists/reads files. Previously `get_contents("")` implicitly checked this (at the cost of the false-negative for empty repos).

   A slightly more precise fix would be to keep the contents probe but treat GitHub's "empty repository" response as valid rather than removing the probe entirely — e.g. tolerate a 404 whose message is `This repository is empty.` while still surfacing a genuine "Not Found" as a validation error. If you'd rather keep the simpler `get_repo()`-only check, consider noting in the docstring/log that content-read errors will surface at ingest time instead of validation time.

Otherwise LGTM — correct, minimal, and no regressions.

This is an AI review comment.

PR #18784, opened 2026-08-26
fix(graph): use max-wins for PageRank in AnalyzeNHopPaths + correct test fixtures (#15695)

Taranum01:
Extracted from #17673 per xugangqiang's review ("two unrelated changes mixed in one PR"). This branch carries only the scoring fix and the corrected fixtures; the TCADP context-propagation fix stays on #17673.

## Problem
When two N-hop paths share an edge (f, t), the Go implementation in `internal/service/graph/scoring.go` was supposed to keep the maximum weight for that edge — matching the Python equivalent in `rag/graphrag/search.py`. The original code, however, only wrote the weight without checking what was already there, so the last-write-wins value clobbered earlier (possibly larger) weights. The result depended on the iteration order of `entsFromQuery`, which is non-deterministic for map iteration.

## Fix
`AnalyzeNHopPaths` now keeps the maximum weight for each shared edge:

```go
    es.PageRank = weights[i]
}
```

This makes the result ordering-independent and matches the Python implementation.

## Test fixture correction (review #1)
The original test fixtures used disjoint paths (`[src, mid, dst]` and `[src, alt, dst]`) which produce NO shared edge, so neither path could ever overwrite the other. The tests would silently pass either under the old last-wins code or under max-wins.

Reworked the fixtures so both paths share the contested `(src, mid)` edge with weights 0.3 and 0.9. The tests now actually exercise the max-wins branch. `TestAnalyzeNHopPathsPageRankMaxWins` asserts the shared edge picks up 0.9 (max) and each path keeps its unique second-hop weight. `TestAnalyzeNHopPathsPageRankGoOrderingIndependent` asserts the result is independent of which path is processed first.

### DIFF
--- a/internal/service/graph/scoring.go
+++ b/internal/service/graph/scoring.go
@@ -40,7 +40,7 @@ func AnalyzeNHopPaths(entsFromQuery map[string]*KGEntity) map[Edge]EdgeScore {
 				edge := Edge{From: f, To: t}
 				es := nhopPathes[edge]
 				es.Sim += ent.Similarity / (2.0 + float64(i))
-				if i < len(weights) {
+				if i < len(weights) && weights[i] > es.PageRank {
 					es.PageRank = weights[i]
 				}
 				nhopPathes[edge] = es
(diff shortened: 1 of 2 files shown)


xugangqiang:
LGTM

PR #18786, opened 2026-08-26
Fix: The recommended chunk size for the token chunker is restricted to integers only.

cike8899:
### Summary

Fix: The recommended chunk size for the token chunker is restricted to integers only.


wangq8 (commented):
Thanks for the fix — the `integer` prop on `NumberInput`/`SliderInputFormField` is a clean approach, and the keydown guard plus paste rejection look correct.

Two concerns before this can be merged:

1. **Unrelated change (scope creep):** the `go-form.tsx` diff replaces `PromptEditor` with a plain `Textarea` for the `keywords.system_prompt`, `questions.system_prompt`, and `summary.system_prompt` fields. That has nothing to do with the integer chunk-size fix and should be split into its own PR (or dropped if accidental).

2. **Likely functional regression:** `PromptEditor` is a Lexical editor that supports template variable nodes (`VariableNode` / `VariablePickerMenuPlugin`) even when `showToolbar` and `showMergePath` are disabled. Swapping it for a `<Textarea>` removes variable insertion for those prompt fields. If this is intentional, please confirm and explain why the variable feature is no longer needed here; otherwise it will silently break users' prompts that reference variables.

The integer-related changes themselves look good to me.

PR #18788, opened 2026-08-26
fix: accept empty ASR verification results

buua436:

ASR model verification incorrectly failed when a valid ASR response contained empty text, which is expected for silent audio.

The verification now only checks whether the ASR request returns an error, allowing valid empty transcriptions to pass.

### Type of change

### DIFF
--- a/internal/service/model_service.go
+++ b/internal/service/model_service.go
@@ -1332,7 +1332,7 @@ func verifyASRModel(ctx context.Context, driver modelModule.ModelDriver, modelNa
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix. This correctly mirrors `verifyOCRModel`'s behavior: an empty transcription is a valid ASR response for silent test audio, and a real failure is already caught by the `err != nil` check above. The change is correct and low-risk.

Minor nit (non-blocking): the error message "returned no transcription" is now slightly misleading since it only fires when `resp == nil`. Consider rephrasing to something like "returned no response" for accuracy. LGTM otherwise.

PR #18790, opened 2026-08-26
fix(sdk): filter documents by multiple IDs

WOLIKIMCHENG:
### Summary

Fix `DataSet.list_documents(ids=[...])` by passing the ID list through the request parameter mapping, which serializes it as repeated `ids` query parameters.

This preserves existing behavior for `ids=[]` as an unfiltered request and for combining `id` with non-empty `ids`, which raises `ValueError`. Add regression coverage for all three cases.

### Testing

- `git diff --check`
- Python syntax compilation for both changed files
- Hermetic request-preparation and fake-transport checks


xugangqiang:
LGTM

PR #18791, opened 2026-08-26
Fix: The chat network search service is allowed to be null.

cike8899:
### Summary
Fix: The chat network search service is allowed to be null.

### DIFF
--- a/web/src/components/auto-keywords-form-field.tsx
+++ b/web/src/components/auto-keywords-form-field.tsx
@@ -39,6 +39,7 @@ export function AutoKeywordsFormField({
       layout={layout}
       sliderTestId="ds-settings-parser-auto-keyword-slider"
       numberInputTestId="ds-settings-parser-auto-keyword-input"
+      integer
     ></SliderInputFormField>
   );
 }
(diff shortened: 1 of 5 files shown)


wangq8 (commented):
This is an AI review comment.

The change correctly lets the chat web-search provider be empty/null. A few observations:

**Looks correct.** The refactor from `RAGFlowSelect` to `SelectWithSearch` with `allowClear`, the schema `.or(z.literal(''))`, and the `?? ''` fallback in `chat-settings.tsx` are consistent: clearing emits `''` (see `handleClear` in `select-with-search.tsx`), which now passes validation, and `web_search_provider` is persisted as `''` rather than a provider value. `SelectWithSearch` does support `allowClear`, `testId`, and `optionTestIdPrefix`, so the prop mapping is sound.

**Minor scope:** `auto-keywords-form-field.tsx` adds an `integer` prop to the auto-keywords/auto-questions sliders, which is unrelated to the network-search-null fix. Consider splitting it out or noting it in the description so it doesn't get lost in a bugfix PR.

**Please verify:** `SelectWithSearch`'s value/`onChange` are typed as `string` (`value?: string; onChange?(value: string): void`), while the provider field's value can be `undefined`/`null`/`''`. It works in practice because the component defaults `value: val = ''`, but confirm the schema type (`WebSearchProvider | '' | undefined`) lines up with what `chat-settings.tsx` persists so clearing doesn't leave a stale provider value on subsequent saves.

This is an AI review comment.

PR #18792, opened 2026-08-26
fix(model-provider): resolve instances by persisted ID

6ba3i:
### Summary

Provider settings were saved correctly, but reopening them requested the instance by ID while the running backend could only find it by name. 
The fix makes provider instance reads honor the saved ID, so configurations load correctly after reopening.


internal/service/model_service.go
```diff
@@ -857,7 +857,6 @@ func (m *ModelProviderService) ListProviderInstances(ctx context.Context, provid
 
 func (m *ModelProviderService) ShowProviderInstance(ctx context.Context, providerName, instanceIDOrName, userID string) (map[string]interface{}, common.ErrorCode, error) {
 	providerName = strings.TrimSpace(providerName)
-	providerName = strings.ToLower(providerName)
```
JinHai-CN, line 860: Why not unify the provider name to lower case characters?

6ba3i: certain providers like zhipu and siliconflow are registered in caps, so lower casing would make look up skip them

PR #18793, opened 2026-08-26
fix(parser): drop merged-block and wrapped PDF TOC entries when remove_toc is enabled

euvre:
### Summary

In the agent workflow (ingestion dataflow canvas), the PDF parser option **"Remove original table of contents" (`remove_toc`)** did not take effect: after running a book-style PDF with the toggle on, the parse summary still contained the original TOC entries (e.g. `第28章 朴散则为器……大制不割？？` / `………39`).

Root cause — `internal/parser/parser/pdf_postprocess.go`:

- The DeepDoc layout merges the lines of a text block into **one** `Section` whose text joins the lines with spaces. A TOC page therefore becomes a single section whose *trailing* text is often a page marker (`II`) or a footer watermark rather than an entry. `filterPDFTOCEntries` only applied the end-anchored leader+page-number pattern to the **whole** section text, so such merged TOC blocks were never dropped.
- Wrapped TOC entries split the entry title and its leader/page number onto separate lines/sections. The bare reference (`………39`) was dropped, but the title-only line (`第28章 ………`) carries no leader run and survived.

Fix — `filterPDFTOCEntries` (Go PDF post-process, `remove_toc` path):

1. A section that ends in a leader+page-number run is still dropped (unchanged anchored rule).
2. A section containing **two or more** leader+page-number runs (explicit dot/ellipsis/middot leaders; wide-space leaders excluded) is dropped as a merged TOC block even when its trailing text is not an entry.
3. When a dropped section is a **bare leader+page-number reference**, the preceding kept section is dropped too if it opens like a TOC entry title (`第N章/节/篇…`, `Chapter N`, `Appendix …`, 前言/序言/附录/致谢…), pairing wrapped entries.

Guard rails against over-deletion: prose with a single leader+number mention matches neither rule 2 nor the anchored rule, and title pairing only fires when a bare page reference was actually dropped.

### Tests

- New unit tests in `internal/parser/parser/pdf_postprocess_test.go`: merged TOC block with trailing non-entry text is dropped; wrapped entry title is paired with its bare page reference (and a preceding watermark/prose line is kept); prose with a single leader run plus a bare reference keeps the prose.
- Existing `TestApplyPDFPostProcess_RemoveTOC*` tests unchanged and passing (`bash build.sh --test ./internal/parser/parser/`).
- End-to-end verification on the Go backend: built an ingestion dataflow canvas (File → Parser(PDF, multi-column ON, remove TOC ON) → Title Chunker → Transformer → Indexer) and ran it against a generated Chinese book PDF whose TOC mimics the reported file (dot-leader entries, a wrapped entry whose page number is on its own line, trailing roman page marker, footer watermarks). Before the fix the parse summary kept `第28章 朴散则为器，圣人用之，则为官长，故大制不割？？`; after the fix the parse summary contains only the cover title and the footer watermark lines — all TOC content is removed.

### Notes for reviewers

- Scope: only `internal/parser/parser/pdf_postprocess.go` and its test file belong to this change. The working tree also carries older uncommitted edits from a previous, unmerged effort (`fix-pdf-remove-toc-header-footer` branch: canvas override-param recursive merge, Python `TOC_ENTRY_RE`, header/footer without DeepDoc inference); those address different defects and should not be folded into this PR.
- Relationship to precedent: the unmerged branch `fix-pdf-remove-toc-header-footer` (commit 361d25972) fixed the anchored-pass prefix quirk and the header/footer no-op without inference. This PR is complementary: it targets what that branch left behind — merged multi-line TOC blocks and wrapped entries — and does not overlap with its changes.
- The Go entry filter now intentionally goes beyond Python's per-item `remove_toc` (rag/flow/parser/utils.py) because Go's section granularity (merged blocks) is coarser than Python's per-item texts; the guard rails keep the behavior conservative.


wangq8 (commented):
This is an AI review comment.

The TOC filtering logic itself is well-thought-out with reasonable guard rails and good test coverage. A few concerns, mostly about scope and the Python side.

**Scope mismatch (please confirm).** The PR body's "Notes for reviewers" states that only `internal/parser/parser/pdf_postprocess.go` and its test belong to this change, and that the "canvas override-param recursive merge" and "Python `TOC_ENTRY_RE`" are from a previous unmerged effort and "should not be folded into this PR". However the diff (7 files) includes exactly those: `internal/agent/canvas/node_body.go` (the `applyOverrideParams` deep-merge), its two test files, and `rag/flow/parser/utils.py` + `test_remove_toc.py`. Please either split these out or update the description — the deep-merge change is a broad semantic change to a shared function and is unrelated to TOC removal.

**Overlap with #18797.** The `applyOverrideParams` deep-merge and #18797's family-seeding in `BuildParserPageCapOverride` solve the same underlying problem (the override-wins merge wiping sibling keys) in two different places. If both merge they're redundant but compatible; if only one merges, behavior differs. Please coordinate these two PRs to avoid divergence.

**Python `remove_toc` has no guard rail.** The new `TOC_ENTRY_RE` drop in `rag/flow/parser/utils.py` applies only the anchored leader+number rule, with no equivalent of the Go side's "≥2 runs" threshold or title pairing. Since `remove_toc` operates on per-item text (finer granularity than Go's sections), a single prose line ending in a leader run + number (e.g. a footnote or a trailing "……12") will be silently dropped. The new test only covers lines ending in *plain* numbers without leaders; please add a false-positive case (prose ending in `.... 12` / `……12`) and consider whether a guard is needed.

**Minor:** `\d{1,4}` caps page numbers at 9999, so large documents with 5-digit page numbers won't be matched. Also Go's `\d` is ASCII-only while Python's `\d` matches Unicode digits, so the two paths can diverge slightly.

This is an AI review comment.

euvre:
Thanks @wangq8 for the review — all four points are addressed in cf5c10688.

**1. Scope mismatch — confirmed and fixed.** You are right: the commit had accidentally swept in uncommitted edits from the earlier unmerged effort, contradicting the PR notes. I reverted `internal/agent/canvas/node_body.go` (plus its two test files), restored `rag/flow/parser/utils.py`, and removed `test/unit_test/rag/flow/parser/test_remove_toc.py`. The branch now contains only `internal/parser/parser/pdf_postprocess.go` and its test, matching the stated scope. The canvas deep-merge and the Python `TOC_ENTRY_RE` change will be re-submitted separately.

**2. Overlap with #18797 — agreed.** Removing the `applyOverrideParams` deep-merge from this PR eliminates the divergence risk; #18797 (family-seeding in `BuildParserPageCapOverride`) stays the single change in that area.

**3. Python `remove_toc` guard rail — valid concern, deferred with the split.** Since the Python change no longer ships in this PR, I did not touch Python behavior here. When re-submitting it separately I will (a) add the false-positive cases you suggest (prose ending in `.... 12` / `……12`) and (b) add a guard mirroring the Go side's conservatism — e.g. require explicit leader glyphs (excluding the wide-space alternative) plus a multi-run threshold or a bare-reference-only shape before dropping an item, so a single prose line ending in a leader+number survives.

**4. Minor regex notes.** On the Go side I kept `\d{1,4}`: the failure mode is conservative (a >9999 page reference simply is not matched, so the entry survives — nothing is wrongly deleted), and real TOC page numbers do not reach five digits. Go's RE2 `\d` is ASCII-only, so there is no Unicode-digit divergence on the Go path; when the Python filter is re-submitted separately I will pin its digits to ASCII (`[0-9]` or `re.ASCII`) to keep both paths aligned.

**Verification:** `bash build.sh --test ./internal/parser/parser/ ./internal/agent/canvas/` — both packages `ok` (the canvas package is back at its mainline state). No shared-service end-to-end run on this line; the behavioral evidence for the Go change is the package unit tests, unchanged from the original commit.

xugangqiang:
Review findings (issues only):

**1. Critical: fix does not run on the most common path (early return).**
`applyRemoveTOC` (pdf_postprocess.go:148) is a 3-way dispatch:
- `len(outlines)==0` → `removePDFTOC` (fix applies)
- `firstOutlinePage <= 1` → `removePDFTOCByOutlines` then `return` (**fix never runs**)
- first outline after page 1 → `removePDFTOC` on pre-outline pages only

Go's PDF parser populates `result.Outlines` from PDF bookmarks (parser.go:92-98). Book-style PDFs almost always have page-1 chapter bookmarks, so they hit the second branch and return early. `removePDFTOCByOutlines` only deletes pages when an outline title matches `目录/contents/...` (findPDFTOCPageRange); for a TOC with no such heading it does nothing. Net result: the reported "TOC without a 目录 heading" case is NOT fixed. The added `filterPDFTOCEntries` is dead code on this path.

Fix: call `filterPDFTOCEntries` after `removePDFTOCByOutlines` in the `firstOutlinePage <= 1` branch (or run it once at the end of `applyRemoveTOC`).

**2. Minor: `[ ]{2,}` treated as a TOC leader.**
`pdfTOCEntryPattern` (pdf_postprocess.go:21) includes `[ ]{2,}` as a leader. A prose line ending in "  12" gets matched and dropped. Remove `[ ]{2,}` from the anchored pattern so only unambiguous leader glyphs (`… . · ⋯`) count, consistent with `pdfTOCEntryAnywherePattern`.

**3. Test coverage gap on the broken path.**
New tests exercise `removePDFTOC`/no-outline path only. There is no test for the `firstOutlinePage <= 1` branch (the one that skips the filter), and the e2e used a synthetic PDF that likely has no bookmarks (so it passes via the no-outline path and does not prove the real fix). Add a regression test: page-1 outline whose title is not "目录", with headingless leader TOC entries on a later page, asserting they are removed.

**4. Possible over-deletion regression.**
A normal paragraph containing two or more `…N` runs (e.g. "见第三章……12 ……34") is dropped by the ">=2 runs" rule, and the title-pairing logic can drop a preceding "第N章"/"Chapter N" line if a bare page reference follows. These are known risks but are unguarded by tests.

euvre:
Hi @xugangqiang — thanks for the sharp review; all four findings are addressed in 4a46ba322. (@wangq8 — your four points remain as answered in cf5c10688; nothing new was raised on them.)

**1. Critical — entry filter never ran on the `firstOutlinePage <= 1` path: confirmed and fixed.** You were right: `applyRemoveTOC` returned right after `removePDFTOCByOutlines`, and that pass is a no-op unless an outline title matches `目录/contents` (`findPDFTOCPageRange` returns 0/0 → early return), so the headingless-TOC fix was dead code on the common book path. The branch now runs `filterPDFTOCEntries` after the outline pass. The third branch (first outline after page 1) already filtered the pre-outline pages via `removePDFTOC`.

**2. Minor — `[ ]{2,}` as a leader: fixed.** Removed from the anchored `pdfTOCEntryPattern`, so only explicit leader glyphs (`… . · ⋯`) count — consistent with `pdfTOCEntryAnywherePattern`, which already excluded wide spaces as too common in prose. A line merely ending in two spaces + a number is now kept.

**3. Coverage gap: fixed.** Added `TestApplyPDFPostProcess_RemoveTOCEntryLinesWithPageOneOutlines` — page-1 outlines whose titles are not "目录", headingless leader TOC entries on page 2, body on page 3; asserts the entries are dropped. I verified both new tests fail on the pre-fix code and pass after it. On your e2e note: agreed, the existing e2e exercised the no-outline path only; the new unit test targets the page-1-outline dispatch branch directly.

**4. Over-deletion risks: guarded by tests, heuristics kept deliberately.**
- The ≥2-run rule is what makes DeepDoc-merged TOC blocks drop at all — their only distinguishing feature is multiple leader+number runs, since the anchored pattern never fires for a merged block. Raising the threshold under-deletes real merged TOCs, so I kept it and pinned the cost explicitly: `TestFilterPDFTOCEntries_TwoRunProseTradeOff` documents that two-run prose is the accepted trade-off, `KeepsProseWithSingleLeaderRun` pins the single-run guard.
- Title pairing only fires when the dropped section is a bare leader+page-number line AND the preceding section opens with a chapter/section marker; in body text a chapter heading is followed by prose, not a bare "………39" line. `TestFilterPDFTOCEntries_ChapterTitleWithoutBarePageRefSurvives` now pins that a chapter heading followed by prose is kept.

If the two-run cost still looks too high for your corpus, a stricter shape (e.g. runs must sit at line starts within the merged block) could be a follow-up — happy to hear your view.

**Verification:** `bash build.sh --test ./internal/parser/parser/` — package `ok`; a targeted `-run 'TestApplyPDFPostProcess_RemoveTOC|TestFilterPDFTOCEntries' -v` run shows all 9 TOC tests passing (4 new). No shared-service end-to-end verification on this line; the behavioral evidence is the package unit tests plus the dispatch-path analysis above.

xugangqiang:
LGTM

PR #18797, opened 2026-08-26
fix(ingestion): respect Parser page range over dataflow debug page cap

euvre:
### Summary

**Background.** A dataflow run from the agent canvas (the "Run" button → `agentChatCompletion`) executes synchronously as a canvas **debug dry-run**. To keep the preview fast, the executor caps the Parser to the first `debugPageCapPages = 2` pages by injecting `pages: [[1, 2]]` into the run-level override params at `ParserConfig[cpnID][family]["pages"]` (`pipeline.BuildParserPageCapOverride`).

**Bug.** The canvas saves the Parser "Page ranges" field on the **component's own DSL params** (`params[family]["pages"]` — the shipped general template carries `[[1, 100000]]`), not in the run-level ParserConfig. Debug contexts intentionally carry an empty ParserConfig, so the "respect existing pages" guard never fired: the 2-page cap was always injected, and because the runtime override merge (`applyOverrideParams`) is top-level-key-wins, it replaced the whole `pdf` family entry — silently wiping the user's explicit page range (and the other family settings). Result: a PDF whose page count is well inside the configured page range still parsed only its first two pages (reported from the Feishu group: 智能体数据管道解析 PDF，上传的 PDF 页数在"页码范围"内但只解析出前两页的内容).

**Fix** (`internal/ingestion/pipeline/parser_page_cap.go`):
- The debug cap is now a pure fallback. It is not injected when the document's family carries an explicit non-empty `pages` range in **either** the run-level `parserConfig[cpnID][family]["pages"]` **or** the Parser component's DSL params `params[family]["pages"]`. Missing/null/empty lists mean "parse all pages" and still allow the cap, preserving the fast-preview behavior for setups without an explicit range.
- When the cap *is* injected, the family entry is seeded from the component's DSL family params (deep-copied; existing parserConfig keys win) so the override no longer drops the other family settings such as `parse_method` / `remove_header_footer`.
- `ExtractParserCpnID` is folded into an unexported `parserComponentParams(dsl, name) (cpnID, params)` helper returning cpnID and params in a single pass; it had no callers outside the package.

**Tests**
- `internal/ingestion/pipeline/parser_page_cap_test.go`: new regression suite `TestBuildParserPageCapOverrideRespectsDSLPages` — explicit DSL pages → no cap; empty/missing pages → cap injected with family settings preserved; pages configured for another family do not block the cap; existing parserConfig keys win over seeded DSL keys.
- `internal/ingestion/task/debug_test.go`: `TestInjectDebugPageCap` updated — the shipped general template carries explicit `pdf.pages [[1, 100000]]` which the cap must now respect (new subtest), and the fallback-injection subtests run on a template variant with the `pages` key stripped (new `dslWithoutPDFPages` helper).
- `bash build.sh --test ./internal/ingestion/pipeline/... ./internal/ingestion/task/...` — all pass.

**End-to-end verification** (Go backend, browser): generated a 6-page PDF with per-page markers, configured the Parser page range to `1–100` in a dataflow canvas (File → Parser → Token Chunker → … → Tokenizer), and ran it. Before the fix the result column contained only pages 1–2; after the fix it contains all 6 pages. This path needs no LLM (debug runs skip embedding/extraction), so verification covered the real pipeline end to end.

### DIFF
--- a/internal/ingestion/pipeline/parser_page_cap.go
+++ b/internal/ingestion/pipeline/parser_page_cap.go
@@ -3,6 +3,8 @@ package pipeline
 import (
 	"encoding/json"
 	"fmt"
+
+	"ragflow/internal/utility"
 )
 
 // UnwrapCanvasDSL decodes a raw pipeline DSL and strips the optional canvas
@@ -73,7 +75,11 @@ func parserComponentParams(dsl []byte, parserComponentName string) (string, map[
 //     override-wins merge and only the first capPages pages would parse.
 //
 // A "pages" value that is missing, null, or an empty list means "parse all
-// pages" and does not count as an explicit range — the cap still applies.
+// pages" and does not count as an explicit range — the cap still applies. The
+// all-pages sentinel carried by the shipped templates and the web UI defaults
+// ([[1, 100000]]) means the same, so a single gap-free [1, N] range with
+// N >= allPagesUpperBound does not count as explicit either; only a genuinely
+// narrowed range suppresses the cap.
 //
 // When no Parser component is found or the family is empty, the call is a
 // no-op and parserConfig is returned unchanged. When the cap IS injected, the
@@ -104,9 +110,12 @@ func BuildParserPageCapOverride(
 	}
 	dslFam, _ := parserParams[family].(map[string]any)
 	// Respect an explicit page range already present under cpnID + family.
+	// The same explicitness predicate as the DSL branch below: null/empty
+	// pages and the all-pages sentinel mean "parse all pages", so the cap
+	// remains a fallback for them.
 	if cpnEntry, ok := parserConfig[parserCpnID].(map[string]any); ok {
 		if famEntry, ok := cpnEntry[family].(map[string]any); ok {
-			if _, has := famEntry["pages"]; has {
+			if hasExplicitPages(famEntry["pages"]) {
 				return parserConfig
 			}
 		}
@@ -139,10 +148,27 @@ func BuildParserPageCapOverride(
 	return parserConfig
 }
 
-// hasExplicitPages reports whether a raw "pages" value carries at least one
-// configured range. Missing, null, and empty-list values mean "parse all
-// pages" and do not count as explicit — the cap remains a fallback for them.
+// allPagesUpperBound is the smallest upper bound of a "pages" range treated as
+// the all-pages sentinel. The general template and the web UI's page-range
+// fields default to [[1, 100000]] (template/ingestion_pipeline_general.json,
+// web chunk-method-dialog and agent parser-form), matching Python's
+// MAXIMUM_PAGE_NUMBER, and the parser clamps oversized upper bounds to the
+// document's actual page count — so a gap-free [1, N] with N >=
+// allPagesUpperBound does not actually restrict parsing.
+const allPagesUpperBound = 100000
+
+// hasExplicitPages reports whether a raw "pages" value carries a range that
+// actually restricts parsing. Missing, null, empty-list, and uninterpretable
+// values do not count (the parser degrades malformed values to "parse all
+// pages" at parse time, so the cap is not withheld for them here either), and
+// neither does the all-pages sentinel — a single gap-free [1, N] range with
+// N >= allPagesUpperBound, the shipped template/UI default. The cap remains a
+// fallback for all of them; only a genuinely narrowed range counts as
+// explicit.
 func hasExplicitPages(raw any) bool {
-	list, ok := raw.([]any)
-	return ok && len(list) > 0
+	ranges, err := utility.NormalizePDFPages(raw)
+	if err != nil || len(ranges) == 0 {
+		return false
+	}
+	return !(len(ranges) == 1 && ranges[0][0] == 1 && ranges[0][1] >= allPagesUpperBound)
 }
--- a/internal/ingestion/pipeline/parser_page_cap.go
+++ b/internal/ingestion/pipeline/parser_page_cap.go
@@ -8,7 +8,7 @@ import (
 // UnwrapCanvasDSL decodes a raw pipeline DSL and strips the optional canvas
 // envelope {"dsl": {...}}, returning the inner components-carrying map. A raw
 // (non-enveloped) DSL is returned unchanged. It is the []byte entry point used
-// by helpers that need the inner DSL (ExtractParserCpnID,
+// by helpers that need the inner DSL (parserComponentParams,
 // BuildParserPageCapOverride) so envelope-handling lives in exactly one place.
 func UnwrapCanvasDSL(raw []byte) (map[string]any, error) {
 	var top map[string]any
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the detailed fix and thorough tests. The core approach (seed the family entry, respect DSL pages) is sound, and the deep-copied `ParamsDefaults` from `ExtractAllComponentParams` means there is no shared-state mutation concern. I have one substantive concern and a minor consistency note.

**Substantive: the fix disables the debug fast-preview for the default template.** The shipped `general` template carries `pdf.pages = [[1, 100000]]`, which is the "all pages" sentinel (the parser clamps the upper bound to the actual page count — the `internal/deepdoc/parser/pdf` tests show `[1,1000000]` on a 5-page doc parses all 5 pages). `hasExplicitPages` treats any non-empty list as an explicit range, so for the default template the `[[1, 2]]` debug cap is never injected. The result is that debug dry-runs — whose entire purpose is a fast 2-page preview (`debugPageCapPages = 2`, see `internal/ingestion/task/pipeline_executor.go`) — will now parse the whole document for every user who did *not* customize the page range. That is the common case and can be very slow for large PDFs.

This effectively inverts the reported bug: instead of "an explicit narrow range was ignored", the default case now loses the cap entirely. Please distinguish the "all pages" sentinel (upper bound ≥ some large threshold, or exactly the shipped `100000`/`1000000` default) from a genuine user restriction, so the cap still applies when the user hasn't actually narrowed the range.

**Minor: inconsistent "explicit pages" handling between the two sources.** The run-level `parserConfig` branch still keys off mere key presence (`if _, has := famEntry["pages"]; has`), while the DSL branch uses `hasExplicitPages` (null/empty → not explicit). So `pages: []` (or `null`) in `parserConfig` suppresses the cap, but the same value in the DSL does not. Consider applying `hasExplicitPages` consistently in both branches.

This is an AI review comment.

xugangqiang:
**Review feedback (debug-purpose lens)**

Thanks for the detailed writeup. I reviewed the change against the actual call site and the `applyOverrideParams` merge semantics. One point reshuffles how I'd weigh this PR:

**Debug is a quick DSL sanity check, not a final-parse preview.** With that framing, the value of this PR re-sorts:

1. **The family-seeding part is the real win.** The injected override was `{pages:[1,2]}`, which — via the family-level whole-entry replace in `applyOverrideParams` (`internal/agent/canvas/node_body.go`) — wiped `parse_method` and friends. So debug ran with *default* parser config, not the user's, defeating the very purpose of "does my DSL run". Seeding the family entry from the DSL params fixes exactly that. This part should merge.

2. **But "respect explicit pages" conflicts with the fast-preview goal.** The shipped default general template carries `pdf.pages = [[1, 100000]]`. `hasExplicitPages([[1,100000]])` is true, so the cap is **not** injected and debug parses the *entire* document. That's the most common out-of-the-box case losing the 2-page fast preview — directly against `debugPageCapPages`'s stated purpose ("The debug preview must return fast"). The e2e in the PR description confirms it: a 6-page PDF now parses all 6 in debug.

3. **Suggested redesign — decouple speed from faithfulness.** Since debug output isn't final, debug should *always* inject the 2-page cap (keep it fast) **and** still seed the family settings (stay faithful to the user's config). Page-count truncation in debug is acceptable because it's a preview, not the result. If you'd rather keep the range-respect behavior, at minimum treat the `[[1, 100000]]` "all pages" sentinel as "no explicit range" so the default template still caps.

**On the existing review comments:**
- CodeRabbit's run-level check consistency suggestion (use `hasExplicitPages` there too) is reasonable but low-severity: in the actual debug path `Doc.ParserConfig` is always nil, so that branch is currently dead. Worth doing for reuse consistency, not for this PR's correctness.
- CodeRabbit's Merge Risk note is backwards: null/empty DSL `pages` still injects the cap (fast preview preserved). The real risk it missed is the opposite — an *explicit* range (especially the default `[[1,100000]]`) disables the cap and makes default debug runs parse everything.

**Bottom line:** keep the family-seeding fix; reconsider the page-range-respect so default-template debug stays fast.

euvre:
Thanks @wangq8 for the careful review — both points are valid and are addressed in e9413556 (which also covers the CodeRabbit inline note on `parser_page_cap.go`).

**Substantive: the fix disabled the debug fast-preview for the default template — applied.** Confirmed: the shipped `general` template carries `pdf.pages = [[1, 100000]]` (`internal/ingestion/pipeline/template/ingestion_pipeline_general.json`), the same "all pages" bound as Python's `MAXIMUM_PAGE_NUMBER = 100000` (`common/constants.py`) and the web UI defaults (`web/src/pages/agent/constant/pipeline.tsx`, the chunk-method dialog and the agent parser-form all default to `from 1, to 100000`), and the deepdoc parser clamps oversized upper bounds to the document's actual page count. Treating any non-empty list as explicit therefore made the 2-page debug cap unreachable in the uncustomized (common) case.

`hasExplicitPages` now normalizes the value and reports "explicit" only for a genuinely narrowed range: a single gap-free `[1, N]` with `N >= 100000` (new `allPagesUpperBound` const) is the all-pages sentinel and keeps the cap as a fallback; gapped multi-range values (e.g. `[[1,10],[12,100000]]`, which excludes page 11) still count as explicit, and malformed values degrade the same way the parser degrades them at parse time.

**Minor: inconsistent "explicit pages" handling between the two sources — applied.** The run-level `parserConfig` branch now uses the same `hasExplicitPages(famEntry["pages"])` predicate instead of mere key presence, so `pages: null`/`[]` no longer suppress the cap in one source but not the other.

**Tests:** unit tests pin both directions in both sources — sentinel `[[1, 100000]]` (DSL and run-level) → cap injected with family settings preserved; genuine ranges (`[[5,10]]`, `[[1,10]]`, gapped multi-range, run-level `[[1,3]]`) → respected; null/empty run-level pages → cap injected. The integration baseline was retargeted from the now-sentinel `[[1,1000000]]` to a genuine `[[1,20]]` whole-fixture range so the uncapped-vs-capped comparison stays meaningful.

Verified with `bash build.sh --test ./internal/ingestion/pipeline/... ./internal/ingestion/task/...` — all pass. Note the boundary: no shared-service end-to-end run on this line; the integration-tier test is updated but not executed here, so the basis is the unit tests plus the code-level evidence above.

xugangqiang:
@euvre  pls confirm if all my previous comments have been fixed.

euvre:
Confirmed @xugangqiang — all points from your review (and the CodeRabbit inline note + @wangq8 review notes it referenced) are addressed in e9413556, the current head of this branch. Recap per point, with where to look:

**1. Family-seeding kept (your point 1).** `BuildParserPageCapOverride` still seeds the injected family entry from the Parser component's DSL params before writing the cap, so `parse_method` and friends survive the whole-entry replace in `applyOverrideParams`. Pinned by the “family settings preserved” subtests in `internal/ingestion/pipeline/parser_page_cap_test.go` and `internal/ingestion/task/debug_test.go`.

**2. Default-template fast preview restored (your point 2, and the substantive note in @wangq8's review).** I kept the range-respect behavior and took the “at minimum” option you offered: the all-pages sentinel no longer counts as an explicit range. `hasExplicitPages` now normalizes the value via `utility.NormalizePDFPages` and treats a single gap-free `[1, N]` with `N >= allPagesUpperBound` (100000 — the value shipped in `internal/ingestion/pipeline/template/ingestion_pipeline_general.json` and the web UI defaults) as “parse all pages”, so the 2-page debug cap is injected again for the uncustomized default template. A genuinely narrowed range still suppresses the cap, including a gapped multi-range such as `[[1,10],[12,100000]]`, which really excludes page 11. Covered from both sources in the tests: DSL sentinel `[[1,100000]]`, run-level sentinels `[[1,100000]]` / `[[1,1000000]]`, gapped multi-range, narrow ranges, and null / empty / malformed values.

**3. Run-level predicate consistency (CodeRabbit's inline suggestion — your “worth doing for reuse consistency” note — and the minor note in @wangq8's review).** The `parserConfig[cpnID][family]["pages"]` check now uses the same `hasExplicitPages(famEntry["pages"])` predicate instead of mere key presence, so `pages: null` / `[]` no longer suppress the cap in one source but not the other. GitHub already marks that inline thread as addressed.

**4. Your remark that CodeRabbit's Merge-Risk framing was backwards** — agreed; no code change was needed for it, and the fallback semantics are now spelled out in the doc comment on `BuildParserPageCapOverride`.

**Verification:** `bash build.sh --test ./internal/ingestion/pipeline/... ./internal/ingestion/task/...` → all green (`pipeline`, `task`, `task/indexdoc`, `task/tool`). Boundary, as before: no shared-service end-to-end run on this line; the integration-tier test (`//go:build integration`, baseline retargeted to a genuine `[[1,20]]` whole-fixture range so the uncapped-vs-capped comparison stays meaningful) is updated but not executed here — the basis is the unit tests plus the code-level evidence above.

euvre:
After discussion, this behavior is by design. It is not a bug. Close this pr.

xugangqiang:
The 2-page cap on a debug run is intentional by design. A canvas debug (the "Run" button) is meant as a fast preview so users can quickly verify that their self-built canvas / pipeline is wired correctly — parser settings, component wiring, output shape — without parsing the whole document. Capping to the first couple of pages is what keeps the preview fast, which is the entire point of a debug preview.

So for the default `[[1, 100000]]` ("all pages") range, the cap still applies and only the leading pages are parsed — this is expected behavior, not a regression. This PR preserves that fast-preview behavior while:
- (a) keeping the other parser family settings (e.g. `parse_method`, `remove_header_footer`) that the override merge previously wiped, and
- (b) respecting a genuinely narrowed range that the user explicitly configured on the DSL.

If you believe a different page count — or a different rule for when the cap should be suppressed — is needed, please share the reasoning (e.g. a concrete case where a 2-page preview is insufficient to validate the canvas, or why the all-pages sentinel should also disable the cap). Happy to adjust.

PR #18799, opened 2026-08-26
fix(chat): restore required prefetch persistence

6ba3i:
### Summary

Go chat creation failed because MySQL required the missing prefetch_size column. The fix persists that column alongside rerank_candidates_count

### DIFF
--- a/internal/entity/chat.go
+++ b/internal/entity/chat.go
@@ -41,6 +41,7 @@ type Chat struct {
 	SimilarityThreshold    float64   `gorm:"column:similarity_threshold" json:"similarity_threshold"`
 	VectorSimilarityWeight float64   `gorm:"column:vector_similarity_weight" json:"vector_similarity_weight"`
 	TopN                   int64     `gorm:"column:top_n" json:"top_n"`
+	PrefetchSize           int64     `gorm:"column:prefetch_size" json:"-"`
 	RerankCandidatesCount  int64     `gorm:"column:rerank_candidates_count" json:"rerank_candidates_count"`
 	TopK                   int64     `gorm:"column:top_k" json:"top_k"`
 	DoRefer                string    `gorm:"column:do_refer;size:1;not null" json:"do_refer"`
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the PR, but I have blocking concerns — please address before merging.

1. Value aliasing bug. `PrefetchSize` is populated from `req["rerank_candidates_count"]`:
   `PrefetchSize: int64FromValue(req["rerank_candidates_count"])`
   `prefetch_size` is a distinct setting from `rerank_candidates_count`; this writes the rerank-candidate count into the prefetch column, which is almost certainly wrong. If `prefetch_size` is a real, independent field it must be read from its own key (`req["prefetch_size"]`) with its own default.

2. Missing migration. `git grep prefetch_size origin/main` returns nothing — there is no Python/peewee migration in `infiniflow/ragflow` that adds a `prefetch_size` column to `dialog`. Merging this as-is means fresh installs (which run the Python migrations) won't have the column, and GORM `Create` will fail with "Unknown column 'prefetch_size'" — reintroducing the exact failure this PR claims to fix. Please add the matching `alter_db_add_column(migrator, "dialog", "prefetch_size", ...)` in `api/db/db_models.py`.

3. Field is not wired into the API. `prefetch_size` is absent from `filterCreateChatPersistedFields` and the update whitelist, so a client can never actually set it; combined with (1), the value will always be a copy of `rerank_candidates_count`. If it is meant to be a first-class setting, add it to the whitelists and expose it in responses (drop the `json:"-"`), mirroring `rerank_candidates_count`.

4. Default handling. `int64FromValue` returns 0 for an absent key. If `prefetch_size` is NOT NULL with no default, ensure the Create path always supplies a sensible non-zero default independent of `rerank_candidates_count`.

Could you also clarify what `prefetch_size` semantically is (which fork/PR introduced the column) so the mapping and default can be validated?

wangq8:
no longer used `PrefetchSize` use: `rerank_candidates_count`

PR #18800, opened 2026-08-26
feat: extract images from xlsx parser

buua436:

The Go XLSX parser previously returned the entire workbook as a single HTML payload and could not preserve embedded or floating images.

This PR changes XLSX parsing to structured JSON output:

- Splits tables according to `chunkRows`.
- Emits images as independent items with `doc_type_kwd: "image"`.
- Preserves image data as `data:image/...;base64,...`.
- Includes worksheet and cell metadata.
- Prevents image binary data from being treated as embedding text.

### Type of change

### DIFF
--- a/internal/ingestion/component/parser.go
+++ b/internal/ingestion/component/parser.go
@@ -506,6 +506,9 @@ func (c *ParserComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[st
 	if dispatched.Err != nil && fileTypeExt != utility.FileTypeOTHER {
 		return nil, dispatched.Err
 	}
+	for _, warning := range dispatched.Warnings {
+		runtime.ReportProgressMessage(ctx, "Parser", "WARNING: "+warning)
+	}
 
 	// 3. Build the legacy `pages` slice. When the dispatch path
 	//    produced a JSON payload, we re-shape it into the page
--- a/internal/ingestion/component/parser.go
+++ b/internal/ingestion/component/parser.go
@@ -506,9 +506,7 @@ func (c *ParserComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[st
 	if dispatched.Err != nil && fileTypeExt != utility.FileTypeOTHER {
 		return nil, dispatched.Err
 	}
-	for _, warning := range dispatched.Warnings {
-		runtime.ReportProgressMessage(ctx, "Parser", "WARNING: "+warning)
-	}
+	reportParserWarnings(ctx, dispatched.Warnings)
 
 	// 3. Build the legacy `pages` slice. When the dispatch path
 	//    produced a JSON payload, we re-shape it into the page
@@ -595,6 +593,12 @@ func (c *ParserComponent) Invoke(ctx context.Context, db *gorm.DB, inputs map[st
 	// callbacks) is owned by the canvas framework (realComponentBody),
 	// not by this component, so we return the work result directly.
 	return out, nil
+}
+
+func reportParserWarnings(ctx context.Context, warnings []string) {
+	for _, warning := range warnings {
+		runtime.ReportProgressMessage(ctx, "Parser", "WARNING: "+warning)
+	}
 }
 
 // buildPagesFromBytes reshapes already-prepared page bytes into the
--- a/internal/ingestion/component/parser_dispatch.go
+++ b/internal/ingestion/component/parser_dispatch.go
@@ -54,6 +54,7 @@ type parserDispatchResult struct {
 	Markdown     string
 	Text         string
 	HTML         string
+	Warnings     []string
 	Err          error
 }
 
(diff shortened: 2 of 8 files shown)


wangq8 (commented):
Thanks for the PR. The image extraction itself is correct — `GetPictureCells`/`GetPictures` exist in excelize v2.11.0, the `Picture{Extension, File, Format.AltText}` fields are used correctly, and the `data:image/...;base64,...` shape matches the established image item contract in the Go pipeline.

However, I have a few concerns I'd like addressed before merging:

1. Output-format contract change + Python parity. The Python deepdoc spreadsheet path (`_spreadsheet`) emits a single HTML string for `output_format == "html"`, and for `"json"` emits per-cell items with `doc_type_kwd: "text"` (not `"table"`). This PR drops HTML output entirely and always emits JSON with `doc_type_kwd: "table"` items holding raw `<table>…</table>` strings. That breaks the `output_format: "html"` contract and diverges from the Python "json" spreadsheet shape. Please confirm this is intentional and note it in the description, or respect the configured `output_format`.

2. Family inconsistency. The sibling `xls_parser.go` and `csv_parser.go` still emit HTML via the shared `recordsToHTMLTableChunks`/`renderSheetTables`. After this PR, `.xlsx` vs `.xls`/`.csv` in the same spreadsheet family take different output contracts. Either migrate XLS/CSV in the same change or file a follow-up.

3. Fail-hard image extraction. If `GetPictureCells` errors on any single sheet, the whole workbook parse fails and already-extracted table content is discarded. Consider degrading gracefully (log + continue) so a malformed drawing on one sheet doesn't break table extraction.

4. MIME mapping is incomplete. `xlsxImageMIMEType` falls through to `image/png` for `.emf`/`.wmf` (Windows metafiles, common in Excel), mislabeling them. Add explicit cases or skip/flag non-bitmap types.

5. `ck_type` at parse time. The parser now emits `ck_type: "table"/"image"` directly, while downstream the chunker derives `ck_type` from `doc_type_kwd`. Please confirm this doesn't double-set or conflict with the chunker's `needsCrop`/`attachMediaContext` path.

Non-blocking: image items with no alt text use the anchor cell (e.g. "C3") as `text`, producing a near-empty text chunk — consider a clearer placeholder or omitting `text`.

xugangqiang:
## Review: Issues to address

### 1. Blocking: `output_format` contract is broken (hardcoded `json`)

The parser now always emits `OutputFormat: "json"` regardless of the configured `output_format`:

`internal/parser/parser/xlsx_parser.go` `ParseWithResult` hardcodes `OutputFormat: "json"`.

`p.OutputFormat` is set via `ConfigureFromSetup` from setups but never read in `ParseWithResult`. This breaks the `output_format: "html"` contract, which is the **Python default** for the spreadsheet family (`rag/flow/parser/parser.py` defaults spreadsheet to `"output_format": "html"`) and is still in the Go allowed list (`internal/ingestion/component/schema/parser.go` allows `{"json", "markdown", "html"}`).

The Python `_spreadsheet` path respects the configured `output_format` and emits html / json / markdown accordingly. The Go path now always returns json, which diverges from Python and from what existing consumers expect.

**Fix**: branch on `p.OutputFormat` (html / json) in `ParseWithResult`, or explicitly document this as an intended breaking change and update the Python side + schema to match.

### 2. Blocking: spreadsheet family inconsistency

After this PR `.xlsx` emits JSON while sibling `.xls` (`xls_parser.go` -> `OutputFormat: "html"`) and `.csv` (`csv_parser.go` -> `OutputFormat: "html"`) still emit HTML. Same family, different output contract.

**Fix**: migrate XLS/CSV in the same change, or file a tracked follow-up and note it in the PR description.

### 3. Non-standard MIME types for metafile images

`xlsxImageMIMEType` returns `image/emf`, `image/wmf`, `image/emz`, `image/wmz` -- none are IANA-registered. Browsers cannot render these as data-URL `` sources, so the image items are not usable downstream. The `ico` case (`image/x-icon`) is correct.

**Fix**: use the `image/x-*` convention for unregistered types (`image/x-emf`, `image/x-wmf`, `image/x-emz`, `image/x-wmz`), or skip/flag non-bitmap types instead of emitting an incorrect data URL.

### 4. Test gaps

- The warning propagation path (`extractXLSXImages` warnings -> `ParseResult.Warnings` -> `parser.go` `WARNING:` progress message) has no test.
- No test for non-PNG image formats / the MIME mapping function directly.
- No test for an image item with no `AltText` (falls back to the anchor cell like `C3` as `text`, producing a near-empty text chunk -- consider a clearer placeholder or omitting `text`).

### 5. Non-blocking: please confirm in the PR description

The parser now sets `ck_type: "table"/"image"` directly. The chunker's `cropTitleChunks` (`internal/ingestion/component/chunker/token.go`) only derives `ck_type` from `doc_type_kwd` when it is empty, so there is no double-set or conflict with `needsCrop` / `attachMediaContext`. Please confirm this is intentional in the description so future readers do not have to re-derive it.

---

Recommendation: fix #1 and #2 before merge. #3 and #4 are strongly recommended. #5 is documentation only.

xugangqiang:
LGTM.

PR #18801, opened 2026-08-26
fix(dataflow): decode non-UTF-8 HTML files before parsing in Go pipeline

euvre:
### Summary

The agent data pipeline (dataflow canvas: File → Parser → Token Chunker → Tokenizer) produced severely garbled text when parsing non-UTF-8 HTML files. Reported case: a Sina Finance balance-sheet page ("贵州茅台(600519)资产负债表", GBK-encoded, ~8.9 KB) parsed into U+FFFD/mixed-rune mojibake on the dataflow result page.

**Root cause.** The Go dataflow path dispatches HTML files to `internal/parser/parser/html_parser.go`, which passed the raw bytes straight to `x/net/html.Parse`. `x/net/html` assumes UTF-8, so GBK/Big5/Shift-JIS bytes became replacement runes and accidental Latin-1-looking fragments. The Python dataflow path does not have this bug: `deepdoc.parser.RAGFlowHtmlParser` decodes the blob via `rag.nlp.find_codec` (UTF-8 check → chardet → codec loop) before handing a `str` to BeautifulSoup.

**Fix.** `HTMLParser.ParseWithResult` now decodes the input to UTF-8 before parsing (`decodeHTMLToUTF8`):

- valid UTF-8 (or empty) passes through untouched, reported as `utf-8`;
- otherwise the document's own BOM or `<meta charset>`/`http-equiv` declaration wins, resolved through the HTML5 prescan in `golang.org/x/net/html/charset.DetermineEncoding`;
- documents without any declaration walk a pragmatic fallback chain GB18030 → Big5 → Shift-JIS → EUC-KR → ISO-8859-1 (mirroring the intent of Python's `find_codec` codec loop, with GB18030 leading because declaration-less legacy pages saved from Chinese sites are overwhelmingly GBK); each candidate must decode the whole blob without producing replacement runes;
- if everything fails the original bytes are kept, so behavior never regresses below the previous UTF-8-only parse;
- the `File` metadata now reports the actually detected encoding label instead of a hardcoded `utf-8` (nothing branches on this field; it is descriptive metadata).

Decoding happens before the optional header/footer strip and the html.Parse walk, so table structure items (`doc_type_kwd:"table"`) are emitted from the decoded document with intact cell text.

**Verification.**

- Reproduced end-to-end before the fix on the local Go stack: uploaded a GBK-encoded Sina-style balance-sheet HTML into an existing dataflow canvas (File → Parser → Token Chunker → Extractor → Tokenizer) and observed the same mojibake on the `/dataflow-result` page; the Go api server (9384) executed the run.
- After the fix, the same browser flow parses the file into readable Chinese (see verification notes in the linked issue thread).
- New unit tests in `internal/parser/parser/html_charset_test.go`: GBK with meta declaration, GBK without any declaration (GB18030 fallback), Big5 with meta declaration, UTF-8 passthrough, and structured-table survival after decode.
- `bash build.sh --test ./internal/parser/parser/...` and `bash build.sh --test ./internal/ingestion/component/...` pass.
- Python alignment reference: `rag.nlp.find_codec` detects `GB2312` for the same fixture and decodes it correctly.

**Relation to prior work.** Charset handling previously landed only for the email family (`7b253ed0b fix(parser): map gb2312 to GBK and widen .eml charset coverage`, `485adb411 test(parser): cover .msg text output and body charset routing`) and for the Python-side `find_codec` itself (`16642140f`). This change applies the same approach to the HTML family via the shared in-package decode helper (`decodeTransform`), so headers/bodies/HTML resolve charsets consistently; no earlier HTML-charset approach is being resubmitted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,6 +54,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pkoukk/tiktoken-go v0.1.8
 	github.com/redis/go-redis/v9 v9.18.0
+	github.com/saintfish/chardet v0.0.0-20230101081208-5e3ef4b5456d
 	github.com/signintech/gopdf v0.36.1
 	github.com/siongui/gojianfan v0.0.0-20210926212422-2f175ac615de
 	github.com/spf13/viper v1.18.2
--- a/go.sum
+++ b/go.sum
@@ -480,6 +480,8 @@ github.com/sagikazarmark/locafero v0.4.0 h1:HApY1R9zGo4DBgr7dqsTH/JJxLTTsOt7u6ke
 github.com/sagikazarmark/locafero v0.4.0/go.mod h1:Pe1W6UlPYUk/+wc/6KFhbORCfqzgYEpgQ3O5fPuL3H4=
 github.com/sagikazarmark/slog-shim v0.1.0 h1:diDBnUNK9N/354PgrxMywXnAwEr1QZcOr6gto+ugjYE=
 github.com/sagikazarmark/slog-shim v0.1.0/go.mod h1:SrcSrq8aKtyuqEI1uvTDTK1arOWRIczQRv+GVI1AkeQ=
+github.com/saintfish/chardet v0.0.0-20230101081208-5e3ef4b5456d h1:hrujxIzL1woJ7AwssoOcM/tq5JjjG2yYOc8odClEiXA=
+github.com/saintfish/chardet v0.0.0-20230101081208-5e3ef4b5456d/go.mod h1:uugorj2VCxiV1x+LzaIdVa9b4S4qGAcH6cbhh4qVxOU=
 github.com/sebdah/goldie/v2 v2.5.3 h1:9ES/mNN+HNUbNWpVAlrzuZ7jE+Nrczbj8uFRjM7624Y=
 github.com/sebdah/goldie/v2 v2.5.3/go.mod h1:oZ9fp0+se1eapSRjfYbsV/0Hqhbuu3bJVvKI/NNtssI=
 github.com/sergi/go-diff v1.0.0/go.mod h1:0CfEIISq7TuYL3j771MWULgwwjU+GofnZX9QAmXWZgo=
--- a/internal/parser/parser/charset.go
+++ b/internal/parser/parser/charset.go
@@ -1,0 +1,126 @@
+//
+// Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package parser
+
+// Shared charset-label resolution and decode-acceptance helpers. Both
+// email_parser.go (RFC 2047 headers, mail bodies) and html_parser.go
+// (decodeHTMLToUTF8) build on these, so a label resolves to the same
+// decoder everywhere and the "first candidate that decodes without
+// replacement runes" walk exists exactly once.
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"strings"
+
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+	"golang.org/x/text/encoding/japanese"
+	"golang.org/x/text/encoding/korean"
+	"golang.org/x/text/encoding/simplifiedchinese"
+	"golang.org/x/text/encoding/traditionalchinese"
+	"golang.org/x/text/transform"
+)
+
+// canonicalCharsetLabel canonicalizes a charset label for comparison and
+// lookup: ASCII lower-casing with surrounding whitespace plus '-' and '_'
+// removed, so detector outputs and WHATWG spellings such as "GB-18030",
+// "Shift_JIS" or "EUC-KR" fold onto one canonical form.
+func canonicalCharsetLabel(label string) string {
+	return strings.NewReplacer("-", "", "_", "").Replace(strings.ToLower(strings.TrimSpace(label)))
+}
+
+// charsetEncoding maps a charset label to its decoder, covering the
+// charsets common in real-world mail and legacy HTML beyond utf-8: the
+// Chinese families (gb2312/gbk/gb18030, big5), shift_jis, euc-kr, and
+// latin-1. It is the single label resolver shared by the RFC 2047 header
+// decoder, decodeMailPayload (mail bodies), and decodeHTMLToUTF8 (HTML
+// bytes) so every caller resolves a charset label identically.
+//
+// The "gb2312" label maps to GBK, not HZGB2312: mail labeled gb2312 carries
+// plain 8-bit GB2312 bytes, while HZGB2312 only decodes the 7-bit HZ escape
+// form used under the distinct "hz-gb-2312" label.
+func charsetEncoding(charset string) (encoding.Encoding, bool) {
+	switch canonicalCharsetLabel(charset) {
+	case "gb2312", "gbk":
+		return simplifiedchinese.GBK, true
+	case "gb18030":
+		return simplifiedchinese.GB18030, true
+	case "hzgb2312":
+		return simplifiedchinese.HZGB2312, true
+	case "big5":
+		return traditionalchinese.Big5, true
+	case "shiftjis", "sjis":
+		return japanese.ShiftJIS, true
+	case "euckr":
+		return korean.EUCKR, true
+	case "iso88591", "latin1":
+		return charmap.ISO8859_1, true
+	}
+	return nil, false
+}
+
+// decodeWithCharset decodes payload under one charset label, rejecting the
+// result when it produces replacement runes (U+FFFD). An empty label is
+// treated as utf-8.
+func decodeWithCharset(payload []byte, charset string) (string, error) {
+	switch canonicalCharsetLabel(charset) {
+	case "utf8", "":
+		s := string(payload)
+		if !strings.ContainsRune(s, '\ufffd') {
+			return s, nil
+		}
+		return "", fmt.Errorf("invalid utf-8")
+	}
+	if enc, ok := charsetEncoding(charset); ok {
+		return decodeTransform(payload, enc.NewDecoder())
+	}
+	// Unknown charset: treat as latin-1.
+	runes := make([]rune, len(payload))
+	for i, b := range payload {
+		runes[i] = rune(b)
+	}
+	return string(runes), nil
+}
+
+// decodeTransform decodes payload through decoder, rejecting results that
+// contain replacement runes.
+func decodeTransform(payload []byte, decoder *encoding.Decoder) (string, error) {
+	reader := transform.NewReader(bytes.NewReader(payload), decoder)
+	decoded, err := io.ReadAll(reader)
+	if err != nil {
+		return "", err
+	}
+	if !strings.ContainsRune(string(decoded), '\ufffd') {
+		return string(decoded), nil
+	}
+	return "", fmt.Errorf("decode produced replacement characters")
+}
+
+// decodeFirstCharsetMatch walks the candidate labels in order and returns
+// the first decode that succeeds without replacement runes, along with the
+// label that produced it. It is the shared fallback-chain walker behind
+// both decodeMailPayload and decodeHTMLToUTF8.
+func decodeFirstCharsetMatch(data []byte, labels []string) (string, string, bool) {
+	for _, label := range labels {
+		if decoded, err := decodeWithCharset(data, label); err == nil {
+			return decoded, label, true
+		}
+	}
+	return "", "", false
+}
--- a/internal/parser/parser/charset_test.go
+++ b/internal/parser/parser/charset_test.go
@@ -1,0 +1,62 @@
+package parser
+
+import "testing"
+
+// TestCharsetEncodingSharedLabelContract locks the single-label contract of
+// charsetEncoding: every fallback label the HTML parser may walk (and the
+// mail chain ships) resolves through the shared resolver, detector/WHATWG
+// spellings fold onto the same encoder, and labels outside the shared
+// family (e.g. windows-1252, handled by the HTML prescan layer, or utf-8,
+// handled natively by decodeWithCharset) stay out so the two parsers cannot
+// drift apart again.
+func TestCharsetEncodingSharedLabelContract(t *testing.T) {
+	mailLabels := []string{"gb2312", "gbk", "gb18030", "hz-gb-2312", "big5", "shift_jis", "sjis", "euc-kr", "iso-8859-1", "latin1"}
+	for _, label := range append(mailLabels, htmlCharsetFallbackLabels...) {
+		if _, ok := charsetEncoding(label); !ok {
+			t.Errorf("charsetEncoding(%q) = (_, false); the shared resolver must cover every mail-chain and HTML-fallback label", label)
+		}
+	}
+
+	aliases := map[string]string{
+		"GB-18030":   "gb18030",
+		"GB2312":     "gb2312",
+		"Shift_JIS":  "shift_jis",
+		"Shift-JIS":  "shift_jis",
+		"EUC-KR":     "euc-kr",
+		"ISO-8859-1": "iso-8859-1",
+		"ISO8859-1":  "iso-8859-1",
+		"Latin1":     "latin1",
+		" Big5 ":     "big5",
+	}
+	for alias, label := range aliases {
+		aliasEnc, aliasOK := charsetEncoding(alias)
+		labelEnc, labelOK := charsetEncoding(label)
+		if !aliasOK || !labelOK || aliasEnc != labelEnc {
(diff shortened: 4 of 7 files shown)


wangq8 (commented):
Thanks for the fix — this closes a real gap (the Go HTML path was UTF-8-only while Python decodes via `rag.nlp.find_codec`). I verified against the local tree:

- `decodeTransform` and `utf8Valid` already exist in the package and are reused correctly.
- `charset.DetermineEncoding(data, "")` performs the HTML5 prescan (BOM / `<meta charset>`), and the `certain` guard is correct — it returns windows-1252 with `certain=false` when nothing is declared, so the fallback chain is only walked for genuinely undeclared docs.
- The GB18030 → Big5 → Shift-JIS → EUC-KR → ISO-8859-1 fallback mirrors Python's codec loop, and `decodeTransform` rejects any candidate producing U+FFFD, so the chain semantics are right.
- Tests cover meta-declared GBK, undeclared GBK, Big5, UTF-8 passthrough, and table-structure survival — good coverage.

Non-blocking nits:
1. Because ISO-8859-1 decodes any byte sequence without U+FFFD, the final `return data, "utf-8"` ("keep original bytes") branch is effectively unreachable for non-empty input. That's fine (it mirrors Python's latin-1 terminal), but a one-line comment noting Latin-1 is intentionally terminal would help future readers.
2. `decodeHTMLToUTF8` does a full `transform` + `io.ReadAll` copy for non-UTF-8 input — a second in-memory copy. Acceptable for HTML sizes, just noting it.

LGTM otherwise.

xugangqiang:
Review (problems only):

**1. Silent mis-decode risk for undeclared non-GBK encodings (substantive)**
`htmlCharsetFallbackChain` puts `gb18030` first. GB18030 is a near-complete mapping, so an *undeclared* Big5/Shift-JIS document will very likely "decode" successfully (no U+FFFD) and be silently turned into wrong Chinese text. Before this PR such input surfaced as visible U+FFFD mojibake; after the fix it becomes *silently wrong* — arguably a worse (undetectable) failure mode for that narrow case. The domain heuristic ("undeclared pages are overwhelmingly GBK") is reasonable for Chinese finance/gov HTML, but the risk should be explicitly acknowledged and ideally fixed. Suggested fix: for the undeclared branch, use a statistical detector (e.g. `github.com/saintfish/chardet`, mirroring Python's `chardet.detect` in `rag.nlp.find_codec`) instead of blindly preferring GB18030; only fall back to the GB18030-first chain when detection is low-confidence or fails. This better matches the Python path and removes the silent-corruption risk.

**2. Test coverage gaps** (`html_charset_test.go`)
Covered: GBK declared, GBK undeclared, Big5 declared, UTF-8 passthrough, decoded table survival. Missing:
- Shift-JIS, EUC-KR, and the ISO-8859-1 terminal fallback (the latter is the "any byte decodes" guarantee and should be asserted explicitly).
- BOM / UTF-16 path via `charset.DetermineEncoding` (a net-new improvement over Python, currently untested).
- A "everything fails" case asserting undecodable bytes still fall back to the original bytes without panic (proves the "never worse than before" claim).
- A test documenting the undeclared-Big5-is-misdecoded-as-GB18030 risk (so a future statistical detector can flip its expectation).

**3. Reuse opportunity / broader scope (not a regression, but worth noting)**
Other `html.Parse` sites in the Go codebase carry the same latent non-UTF-8 bug and are not covered: `agent/tool/crawler.go:235` (fetches arbitrary web pages — real candidate), plus `google_scholar.go`, `imap.go`, `confluence.go`. Consider extracting `decodeHTMLToUTF8` into a shared helper so these can reuse it. Out of scope for this PR, but a good follow-up.

**4. Minor: misleading label on total failure**
When all decode attempts fail, `decodeHTMLToUTF8` returns `data, "utf-8"`. The input is by definition non-UTF-8 here, so the reported `File["encoding"]` label is wrong (nothing consumes it today, so cosmetic only). Consider returning an empty/neutral label instead.

**Re: CodeRabbit comments**
- "new test file needs the repository-required build tag" is a **false positive**. This package's unit tests carry no build tag (see `html_parser_test.go`, `email_parser_test.go`; AGENTS.md unit tier = none), and the new test only depends on `golang.org/x/text` (pure Go, no CGO). Do **not** add a tag.
- "docstring coverage 77.78% < 80%" is low-severity; the undocumented function is most likely `ParseWithResult`, which already lacked a doc comment before this PR (pre-existing tech debt). Optional to address, not required.

Net: the fix is correct for the common GBK cases and safe to merge; item 1 is the one real issue to discuss (not necessarily a blocker), and items 2–4 are follow-ups.

euvre:
Thanks @wangq8, @xugangqiang and @coderabbitai for the thorough review — all substantive items are addressed in db31019c5.

**1. Silent mis-decode of undeclared Big5/Shift-JIS/EUC-KR (@xugangqiang #1, CodeRabbit) — fixed.**
`decodeHTMLToUTF8` now runs statistical detection (`github.com/saintfish/chardet`, the Go port of the chardet Python uses in `rag.nlp.find_codec`, MIT, zero deps) over the first 1024 bytes before the fallback chain, mirroring Python's `chardet.detect` step. One deliberate deviation from the suggestion: the detection is **gated at confidence ≥ 90** rather than honored unconditionally. Calibration against real fixtures showed the Go port's low-confidence guesses are frequently wrong — a sparse GBK balance-sheet page (this PR's core scenario) scores "Big5" at conf 40, and a short Big5 page scores ISO-8859-1 at conf 17 — while every genuine CJK fixture scores 100. So a confident pick jumps the queue, and anything weaker keeps the GB18030-first domain heuristic.

**Bonus fix found while reworking this — the `<meta charset>` prescan was effectively dead code.** x/net's `DetermineEncoding` returns `certain=true` only for BOMs (or a contentType header, which we never pass); an HTML5 meta-prescan hit returns `certain=false` — exactly like its windows-1252 "nothing declared" default. So the old `certain &&` guard skipped the prescan for every document, declared or not, and the original meta-declared tests passed only because the chain coincidentally handled those fixtures. The guard now honors every prescan name except the `windows-1252` sentinel, which is the behavior the code (and @wangq8's review) always described. This also means a *declared* Shift-JIS/EUC-KR page whose bytes GB18030 would happily absorb is now decoded correctly via its declaration instead of relying on chain luck.

**2. Test coverage gaps (@xugangqiang #2) — added.** Undeclared Big5 (the flipped risk case: it now asserts the original text round-trips instead of GB18030 mojibake), undeclared Shift-JIS, undeclared EUC-KR, ISO-8859-1 terminal fallback with explicit assertion, UTF-16LE BOM via the prescan, empty input, and a hostile-bytes test asserting the never-worse-than-before guarantee (replacement-free output, non-empty label, no panic). 12 charset tests total (5 original + 7 new).

**3. Reusing `decodeHTMLToUTF8` in `agent/tool/crawler.go` / `google_scholar.go` / `imap.go` / `confluence.go` (@xugangqiang #3) — deferred** as you suggested; out of scope for this PR but a good follow-up. (Happy to send it separately.)

**4. Misleading `utf-8` label on total failure (@xugangqiang #4) + Latin-1 terminality comment (@wangq8 nit 1) — fixed.** The unreachable tail now returns a neutral `""` label, and the chain / function comments explicitly document that ISO-8859-1 is intentionally terminal (it decodes any byte sequence without U+FFFD, so the ordered loop always returns before the tail).

**5. Extra in-memory copy in `decodeHTMLToUTF8` (@wangq8 nit 2) — acknowledged, left as-is** since it's acceptable for HTML sizes, as noted.

**6. Build tag on `html_charset_test.go` (CodeRabbit) — not applied**, agreeing with @xugangqiang's assessment: the repo's unit tier carries no build tag (AGENTS.md), and this package's existing unit tests (`html_parser_test.go`, `email_parser_test.go`) have none; the new test only depends on pure-Go `golang.org/x/text`. The docstring-coverage warning is also addressed in passing: the previously undocumented test helpers now carry doc comments.

**Verification:** `bash build.sh --test ./internal/parser/parser/...` — package `ok`, all 12 HTML-charset tests pass plus the rest of the package suite. No shared-service end-to-end run on this line; the change is validated by unit tests plus line-level comparison against the Python path (`rag/nlp/__init__.py:find_codec`: UTF-8 check → chardet detection → codec loop).

xugangqiang (commented):
## windows-1252 declaration is dropped even when explicitly declared

`charset.DetermineEncoding` returns `windows-1252` in two *different* situations, but the current guard treats them identically:

```go
if enc, name, _ := charset.DetermineEncoding(data, ""); enc != nil && name != "windows-1252" {
```

- When a `<meta charset=...>` / BOM actually declares the encoding, `DetermineEncoding` returns the encoding **with `certain == true`**.
- When *nothing* is declared, it falls back to its default `windows-1252` **with `certain == false`**.

The code ignores the `certain` return value and unconditionally skips `windows-1252`, so an HTML page that **explicitly declares `charset=windows-1252`** is never decoded with the windows-1252 decoder — it falls through to the fallback chain and lands on the ISO-8859-1 terminal instead.

That is a real correctness bug: windows-1252 and ISO-8859-1 only differ in the 0x80–0x9F range (smart quotes “ ” = 0x93/0x94, € = 0x80, ™ = 0x99, en dash 0x96, …). In ISO-8859-1 those bytes are C1 control characters, so a declared windows-1252 page with such punctuation renders as mojibake — exactly the class of bug this PR is meant to fix.

### Suggested fix

Use the `certain` boolean to distinguish a real declaration from the default sentinel:

```go
if enc, name, certain := charset.DetermineEncoding(data, ""); enc != nil && (certain || name != "windows-1252") {
    if decoded, err := decodeTransform(data, enc.NewDecoder()); err == nil {
        return []byte(decoded), name
    }
}
```

Now an explicitly-declared windows-1252 page decodes correctly (label `"windows-1252"`), while the undeclared default sentinel (`certain == false`) is still skipped and proceeds through the detect-then-chain path. Please also update the comment at lines 214–218 (which claims a declared windows-1252 "lands on the chain's ISO-8859-1 terminal") and add a fixture with 0x93/0x94 punctuation asserting `File["encoding"] == "windows-1252"`.

euvre:
Hi @xugangqiang — you're right, and this is fixed in 854da6b93 (`fix(dataflow): honor declared windows-1252 HTML charset`).

**The bug, confirmed first.** Before touching anything I added the regression test and watched it fail exactly as you predicted: a `<meta charset="windows-1252">` page containing `“ ”` (0x93/0x94), `—` (0x97), `™` (0x99) and `€` (0x80) decoded to C1 control characters (`\u0093Smart\u0094 pricing \u0097 …`) with `File["encoding"] == "iso-8859-1"`. The comment that documented that landing spot as accepted behavior is gone, as requested.

**Why I did not use the suggested `certain ||` guard.** In x/net v0.56.0 (`html/charset/charset.go`), a meta prescan hit returns `(e, name, false)` — `certain` is only ever set for BOMs (or a Content-Type charset parameter, which we never pass since we call `DetermineEncoding(data, "")`), and there is no windows-1252 BOM. So `certain=true` together with `name == "windows-1252"` is unreachable, and `(certain || name != "windows-1252")` is behaviorally identical to the existing `name != "windows-1252"` — it would not fix the bug. This is the same prescan/default conflation noted in my earlier summary: the two cases are indistinguishable through `DetermineEncoding`'s return values.

**What the fix does instead.** Only for the `windows-1252` result, the first 1024 bytes (the same window the prescan consumes) are re-scanned for a `<meta … charset=…>` declaration, matching both the `charset` attribute form and the `http-equiv="Content-Type"` pragma form. The raw label is resolved through `charset.Lookup`, so WHATWG aliases (`iso-8859-1`, `us-ascii`) canonicalize to windows-1252 exactly as the prescan does. A confirmed declaration is honored (decoded via the windows-1252 decoder, labeled `windows-1252`); the undeclared sentinel is still skipped, so undeclared pages keep the detect-then-chain path and the GBK-domain behavior from the earlier round is unchanged.

**Tests.** `TestHTMLParser_DecodesDeclaredWindows1252` (your exact ask: a 0x93/0x94 punctuation fixture asserting `File["encoding"] == "windows-1252"`) and `TestHTMLParser_DecodesDeclaredLatin1AliasAsWindows1252` (the http-equiv pragma declaring `iso-8859-1`). Both fail on the pre-fix code with the C1-control mojibake above and pass after the fix.

**Verification.** `bash build.sh --test ./internal/parser/parser/...` — package `ok`, all 14 HTML-charset tests pass (12 from the previous round + these 2), plus the rest of the package suite. As before, no shared-service end-to-end run on this line; the change is validated by unit tests plus line-level reading of x/net v0.56.0's `DetermineEncoding`/`prescan`.

@wangq8 — your earlier LGTM and nits were addressed in db31019c5 (see my previous summary); this change only tightens the windows-1252 declaration handling and does not alter the paths you reviewed.

xugangqiang:
I reviewed this against current `main` and found a **functional overlap** worth addressing before merge.

`main` already ships a charset-decode subsystem in `internal/parser/parser/email_parser.go`:
- `charsetEncoding` (email_parser.go:230) maps charset labels to `encoding.Encoding` (gb2312/gbk->GBK, gb18030, big5, shift_jis, euc-kr, iso-8859-1, ...).
- `buildCharsetChain` / `decodeWithCharset` / `decodeMailPayload` (email_parser.go:549-592) implement the "walk a candidate chain, accept the first that decodes without U+FFFD" fallback pattern.
- `decodeTransform` (email_parser.go:600) is correctly reused by this PR.

This PR adds a **second, parallel** charset-resolution implementation in the same package:
- `htmlCharsetFallbackChain` + `htmlCharsetNormalizeLabel` re-implement the label->encoding mapping that `charsetEncoding` already provides (same set of CJK/Latin encoders, different matching logic: `ToLower`+`TrimSpace` vs also stripping `-`/`_`).
- `decodeHTMLToUTF8`'s fallback loop re-implements the candidate-chain pattern from `decodeMailPayload`/`decodeWithCharset`.

The genuinely new, non-overlapping parts are the auto-detection layer (chardet via `saintfish/chardet`, BOM/`<meta>` prescan via `charset.DetermineEncoding`, `htmlMetaCharsetRe`, `htmlDeclaresWindows1252`) — that is appropriate to keep HTML-specific.

But the duplicated label->encoding mapping and candidate-chain traversal conflict with the repo's "collapse duplicate paths" guidance (AGENTS.md). Two parallel mappings can drift apart over time.

**Question:** Would you prefer to resolve the duplication **within this PR** (extract a shared `resolveCharset` / `decodeFirstMatch` helper that both `email_parser.go` and `html_parser.go` build on, keeping the HTML discovery layer separate), or in a **follow-up PR**? I'm fine either way; just want to align before merge.

euvre:
Hi @xugangqiang — agreed, the duplication was real, and I resolved it **within this PR** in 7ed4e05a1 (`refactor(parser): share charset resolution between email and html parsers`), following the repo's collapse-duplicate-paths guidance rather than deferring.

**What was extracted (new `internal/parser/parser/charset.go`, same package):**
- `charsetEncoding` is now the single label→encoding resolver, moved out of `email_parser.go`, with the label canonicalization folded in (`canonicalCharsetLabel`: lower-case, trim, strip `-`/`_`) so WHATWG/detector spellings (`GB-18030`, `Shift_JIS`, `EUC-KR`, …) and the mail aliases resolve through one switch. That subsumes the deleted `htmlCharsetNormalizeLabel` and the parallel mapping in `htmlCharsetFallbackChain`.
- `decodeWithCharset` and `decodeTransform` moved beside it (formerly duplicated logic in the HTML chain walk).
- `decodeFirstCharsetMatch` is the shared candidate-chain walker you sketched: first label whose decode produces no U+FFFD wins, reporting the label that produced it. `decodeMailPayload` and `decodeHTMLToUTF8` both build on it now — `decodeMailPayload` also lost a dead `enc == ""` branch that `buildCharsetChain` could never emit.

**What stayed HTML-specific** (unchanged, as you suggested): the discovery layer — utf-8 passthrough, `charset.DetermineEncoding` prescan, `htmlDeclaresWindows1252` + `htmlMetaCharsetRe`, the chardet confidence gate — plus the GB18030-first label order, which is a domain heuristic: `htmlCharsetFallbackLabels` is now a plain `[]string` of labels resolved through the shared resolver instead of its own `encoding.Encoding` table. Mail keeps its own utf-8→gb2312→gbk→gb18030→latin1 order and the RFC 2047 header reader.

**Behavior:** unchanged on both paths; the only delta is strictly-additive alias tolerance in the shared resolver (e.g. `GB_18030` now resolves; previously only explicit spellings did — same encoder, same bytes), and the dead-branch deletion. `TestCharsetEncodingSharedLabelContract` locks the contract so the two callers can't drift apart again (every mail-chain and HTML-fallback label must resolve; alias spellings must fold onto the same encoder; windows-1252/utf-16/utf-8 stay out of the shared resolver by design), plus `TestDecodeFirstCharsetMatchAcceptance` pins the walker's first-match/reject semantics.

**Verification:** `bash build.sh --test ./internal/parser/parser/...` — package `ok`; verbose runs confirm the 42 HTML tests (incl. all 14 charset cases), the mail charset tests (`TestDecodeHeaderWord_Charsets`, `TestDecodeMailPayload_DeclaredCharsets`, `TestReadMailBody_DeclaredBodyCharsets`), and the two new contract tests all pass. As before on this line: no shared-service end-to-end run — unit tests + code-level reading only.

Net: −102 lines while adding the shared helpers and the drift-guard tests. Thanks for catching this before merge.

xugangqiang:
LGTM

euvre:
Hi @xugangqiang @wangq8 — thanks for the LGTM! 🙂 All feedback on this PR was addressed in my earlier summaries: db31019c5 (chardet statistical detection for undeclared pages, added test coverage incl. Shift-JIS/EUC-KR/ISO-8859-1-terminal/BOM/hostile-bytes, neutral label on total decode failure, Latin-1 terminality comment), 854da6b93 (honor explicitly declared windows-1252 with 0x93/0x94 fixtures), and 7ed4e05a1 (charset resolution shared between the email and HTML parsers via charset.go, with contract tests). I just re-verified on the current branch tip: `bash build.sh --test ./internal/parser/parser/...` → package ok, all 14 HTML-charset tests plus the shared-charset contract tests pass. Nothing new remains on my side; happy to re-check anything that still looks off. Thanks again for the thorough review!

PR #18803, opened 2026-08-26
fix(web): limit slider number inputs to the step's decimal precision

euvre:
### Summary

The number input beside sliders (`NumberInput`, used by `SliderInputFormField`) propagated any typed value unchanged. For decimal-stepped fields such as the Tokenizer **filename embedding weight** (step `0.01`) on the dataset configuration page — and the same field on the agent canvas, reported in the original thread — users could type values like `0.3333333333333333`, which were persisted verbatim into `parser_config`. Persisted float artifacts (e.g. `0.10000000149011612`) were also displayed raw, so the UI and the stored value disagreed.

**Fix** (frontend only, `web/src/components/originui/number-input.tsx`):

- `NumberInput` now consumes its `step` prop instead of only forwarding it to the DOM: every value leaving the component — typed input, and blur (after min/max clamping) — is rounded to the decimal precision implied by a fractional step (`0.01` → 2 decimals). The form therefore never holds more decimals than the step allows, no matter how it is submitted.
- Over-precise persisted values are normalized to the step precision for display on mount, so a legacy `0.3333333333333333` shows (and re-saves) as `0.33`.
- Undefined or whole-number steps keep full precision, so integer sliders (`step` unset or `1`) are unaffected; `integer` mode behavior is unchanged.
- The `step` attribute is still forwarded to the underlying `<input>` element.

Because the dataset configuration pipeline tabs (`Parser_0` / `Token Chunker_0` / `Extractor_0` / `Indexer_0`) reuse the agent operator forms, this single shared-component fix covers both the dataset configuration page and the agent canvas Tokenizer form, plus every other decimal-stepped field (raptor threshold, similarity threshold, overlapped percent, clustering threshold/ratio, LLM temperature).

**Verification**

- New unit tests in `web/src/components/originui/__tests__/number-input.test.tsx` (5 cases: typed rounding, persisted-value display normalization, blur clamp+round, no-step passthrough, whole-number-step passthrough) — all pass; oxlint clean on the changed files.
- End-to-end on the running stack: on the dataset configuration page (Indexer_0 tab), the legacy stored `0.10000000149011612` now displays as `0.1`; typing `0.3333333333333333` snaps to `0.33`; clicking Save sends and stores `filename_embd_weight: 0.33` in `parser_config` (PUT request body and API response confirmed).

### DIFF
--- a/web/src/components/originui/__tests__/number-input.test.tsx
+++ b/web/src/components/originui/__tests__/number-input.test.tsx
@@ -1,0 +1,78 @@
+import NumberInput from '../number-input';
+import { fireEvent, render } from '@testing-library/react';
+import { useState } from 'react';
+
+function Harness({
+  step,
+  initialValue = 0,
+  min = 0,
+  max = 1,
+}: {
+  step?: number;
+  initialValue?: number;
+  min?: number;
+  max?: number;
+}) {
+  const [value, setValue] = useState(initialValue);
+  return (
+    <NumberInput
+      value={value}
+      onChange={setValue}
+      step={step}
+      min={min}
+      max={max}
+      hideIcons
+    />
+  );
+}
+
+const getInput = () => document.querySelector('input') as HTMLInputElement;
+
+describe('NumberInput', () => {
+  it('rounds typed decimals to the step precision', () => {
+    render(<Harness step={0.01} />);
+    const input = getInput();
+
+    fireEvent.change(input, { target: { value: '0.3333' } });
+
+    // The controlled round-trip feeds the rounded value back into the input.
+    expect(input.value).toBe('0.33');
+  });
+
+  it('normalizes an over-precise persisted value to the step precision', () => {
+    render(<Harness step={0.01} initialValue={0.3333333333333333} />);
+
+    expect(getInput().value).toBe('0.33');
+  });
+
+  it('clamps and rounds on blur', () => {
+    render(<Harness step={0.01} />);
+    const input = getInput();
+
+    // Out-of-range values stay raw while editing and are clamped on blur.
+    fireEvent.change(input, { target: { value: '1.555' } });
+    expect(input.value).toBe('1.555');
+
+    fireEvent.blur(input);
+
+    expect(input.value).toBe('1');
+  });
+
+  it('keeps full precision when no step is provided', () => {
+    render(<Harness />);
+    const input = getInput();
+
+    fireEvent.change(input, { target: { value: '0.3333333333333333' } });
+
+    expect(input.value).toBe('0.3333333333333333');
+  });
+
+  it('does not round decimals for whole-number steps', () => {
+    render(<Harness step={1} />);
+    const input = getInput();
+
+    fireEvent.change(input, { target: { value: '0.5' } });
+
+    expect(input.value).toBe('0.5');
+  });
+});
--- a/web/src/components/originui/__tests__/number-input.test.tsx
+++ b/web/src/components/originui/__tests__/number-input.test.tsx
@@ -7,11 +7,13 @@ function Harness({
   initialValue = 0,
   min = 0,
   max = 1,
+  hideIcons = true,
 }: {
   step?: number;
   initialValue?: number;
   min?: number;
   max?: number;
+  hideIcons?: boolean;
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Solid fix and well-tested. Consuming `step` for rounding on change/blur/mount correctly prevents over-precise values from reaching the form, and normalizing legacy persisted values on mount is a nice touch. The tests cover the key cases.

Minor nits (non-blocking):

1. `getStepDecimals` derives precision from `String(step).split('.')`, which breaks for steps that stringify in scientific notation (e.g. `1e-7` → `"1e-7"` has no `.`, so it returns 0). Unlikely with the current slider steps, but a small guard for the exponent form would make it robust.

2. The icon `+`/`-` buttons (`handleIncrement`/`handleDecrement`) still step by ±1 regardless of `step`. In practice `SliderInputFormField` passes `hideIcons`, so this only affects callers that show icons with a fractional `step` (none today), but worth a follow-up to keep them consistent.

LGTM otherwise.

euvre:
Thanks @wangq8 for the review! Both nits are addressed in 8a489c0:\n\n**1. Scientific-notation steps** — \ no longer counts digits after \ in \. It now treats a step as fractional via \ (so fractional steps >= 1 such as \ also get their precision) and parses the mantissa/exponent when \ uses exponent notation, so \ yields 7 decimals instead of 0. New tests: \ rounds \ to \, and \ rounds \ to \.\n\n**2. Icon buttons ignoring \** — this turned out to be live rather than theoretical: three current callers render \ with icons visible and a fractional \ — the agent form's delay-after-error field (\), \, and the memory advanced-settings form (both \) — and there \ could even emit values below \ (e.g. \ on a \ field). The \/\ handlers now share one \ helper that steps by \ (falling back to \ when absent, so integer callers behave exactly as before), rounds to the step precision, and clamps to \. New tests cover stepping by \ from the buttons and the no-op at \.\n\nVerification: \ reports no errors in the touched files; the repo's pre-existing errors elsewhere are an identical set before/after the change. No shared-service end-to-end run on this line — the change is verified by the added jest coverage plus type-check. The remaining CodeRabbit suggestions (eager state initialization in \, and a post-rounding re-clamp for bounds not representable at the step precision) were not applied: the existing \ already normalizes the persisted value before any user interaction, and no current caller configures such bounds.

euvre:
Apologies for the noise -- my previous comment's markdown got mangled by shell escaping; this comment replaces it.

Thanks @wangq8 for the review! Both nits are addressed in 8a489c0:

**1. Scientific-notation steps** -- `getStepDecimals` no longer counts digits after `'.'` in `String(step)`. It now treats a step as fractional via `Number.isInteger(step)` (so fractional steps >= 1 such as `1.5` also keep their precision) and parses the mantissa/exponent when `String(step)` uses exponent notation, so `1e-7` yields 7 decimals instead of 0. New tests: `step={1e-7}` rounds `0.3333333333333333` to `0.3333333`, and `step={1.5}` rounds `1.234` to `1.2`.

**2. Icon buttons ignoring `step`** -- this turned out to be live rather than theoretical: three current callers render `NumberInput` with icons visible and a fractional `step` -- the agent form's delay-after-error field (`step={0.1}`), `similarity-slider`, and the memory advanced-settings form (both `step={0.01}`) -- and there `handleDecrement` could even emit values below `min` (e.g. `0.3 -> -0.7` on a `[0,1]` field). The `+`/`-` handlers now share one `applyStep` helper that steps by `step` (falling back to `1` when absent, so integer callers behave exactly as before), rounds to the step precision, and clamps to `[min, max]`. New tests cover stepping by `0.01` from the buttons and the no-op at `min`.

Verification: `cd web && npm run type-check` reports no errors in the touched files, and the repo's pre-existing errors elsewhere are an identical set before/after the change. No shared-service end-to-end run on this line -- the change is verified by the added jest coverage plus type-check. The remaining CodeRabbit suggestions (eager state initialization in `useState`, and a post-rounding re-clamp for bounds not representable at the step precision) were not applied: the existing `useEffect` already normalizes the persisted value before any user interaction, and no current caller configures such bounds.

dcc123456:
This has already been fixed.

PR #18804, opened 2026-08-26
fix(document): reject config edits while document is parsing or scheduled

xugangqiang:
## Summary

Fixes the bug where editing a document's configuration (parser config, chunk method, pipeline id, name, enabled flag) **while the document is being parsed** causes the in-flight parser to race with the config change. The running worker keeps reading the document row, which can leave the document stuck (front-end shows a wrong/garbled parse status) and undeletable.

Root cause: `UpdateDatasetDocument` (PATCH `/datasets/:dataset_id/documents/:document_id`) validated fields but never checked the document's run status, so config edits were applied live to a `RUNNING`/`SCHEDULE` document.

## Changes

- `internal/entity/dataset.go`: add `DocumentModifiableStatuses` — the run statuses in which a document may be edited (`UNSTART`, `CANCEL`, `DONE`, `FAIL`); `RUNNING` and `SCHEDULE` are excluded.
- `internal/service/document/document_dataset_update.go`: add `validateDocumentModifiable` and call it at the entry of `validateDatasetDocumentUpdate` (field-agnostic guard, gated on `len(present) > 0` so an empty PATCH is still a no-op).
- The stop/start/batch-status flows (`StopParseDocuments`, `StartIngestionTask`, `BatchUpdateDocumentStatus`) do **not** go through this endpoint, so stopping a parse is unaffected.
- Deletion is intentionally left unrestricted so a stuck document can still be cleaned up.

## Test plan

- `internal/service/document/document_modify_guard_test.go` (Go unit tier, in-memory SQLite, no external services):
  - `TestValidateDocumentModifiable_RunStatuses`: pure-function coverage of all six run statuses plus `nil` (defaults to `UNSTART`).
  - Integration via `UpdateDatasetDocument`: rejects `RUNNING` edits for `parser_config`/`chunk_method`/`name`/`enabled`/`meta_fields` and `SCHEDULE` edits; allows `DONE`/`CANCEL`/`FAIL`/`UNSTART` edits; empty `present` on a `RUNNING` doc is not rejected.

Run: `bash build.sh --test ./internal/service/document/...`

## Notes

Front-end UX (disabling the edit controls while a document is parsing) is a recommended follow-up but is out of scope for this back-end-only change.

### DIFF
--- a/internal/entity/dataset.go
+++ b/internal/entity/dataset.go
@@ -72,6 +72,19 @@ const (
 	TaskStatusSchedule TaskStatus = "5"
 )
 
+// DocumentModifiableStatuses are the run statuses in which a document's
+// configuration (parser config, chunk method, pipeline id, name, enabled
+// flag, ...) may be edited via UpdateDatasetDocument. Documents that are
+// running ("1") or scheduled ("5") must not be edited — the in-flight parser
+// would race with the config change and the document can get stuck and become
+// undeletable.
+var DocumentModifiableStatuses = map[TaskStatus]bool{
+	TaskStatusUnstart: true,
+	TaskStatusCancel:  true,
+	TaskStatusDone:    true,
+	TaskStatusFail:    true,
+}
+
 // PipelineTaskType represents the type of pipeline task
 type PipelineTaskType string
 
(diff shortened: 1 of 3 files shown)


xugangqiang:
@yuzhichang

PR #18805, opened 2026-08-26
Fix chat creation after prefetch_size rename

6ba3i:
The migration now copies existing prefetch_size values into rerank_candidates_count and removes the obsolete column. This keeps the database schema aligned with the Go chat creation code

### DIFF
--- a/api/db/db_models.py
+++ b/api/db/db_models.py
@@ -2002,6 +2002,27 @@ def alter_db_drop_index(migrator, table_name, index_name):
         pass
 
 
+def migrate_dialog_rerank_candidates():
+    """Complete the prefetch_size -> rerank_candidates_count migration."""
+    try:
+        columns = {column.name for column in DB.get_columns("dialog")}
+        if "prefetch_size" not in columns:
+            return
+        if "rerank_candidates_count" in columns:
+            DB.execute_sql(
+                "UPDATE dialog SET rerank_candidates_count = prefetch_size"
+            )
+            DB.execute_sql("ALTER TABLE dialog DROP COLUMN prefetch_size")
+        else:
+            DB.execute_sql(
+                "ALTER TABLE dialog RENAME COLUMN prefetch_size TO rerank_candidates_count"
+            )
+    except Exception as ex:
+        logging.critical(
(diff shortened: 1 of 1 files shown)


wangq8:
still no need.

wangq8 (commented):
Thanks for catching this. The `prefetch_size` → `rerank_candidates_count` rename (#18718) only added the new column with `default=64` and left existing deployments with an orphaned `prefetch_size` column, so chat creation silently used 64 instead of the configured value. Copying the data before dropping the old column is the right fix.

A few points to address:

1. **Dead `else` branch.** `migrate_dialog_rerank_candidates()` runs right after `alter_db_add_column(..., "rerank_candidates_count", ...)`, which guarantees the column already exists (it is also on the model). So the `RENAME COLUMN` fallback is unreachable in practice. It's fine as defense-in-depth, but note `ALTER TABLE ... RENAME COLUMN old TO new` is only supported on MySQL 8.0+ — harmless only because the branch never executes.

2. **GaussDB compatibility.** The raw `ALTER TABLE dialog DROP COLUMN prefetch_size` / `RENAME COLUMN ...` statements use unquoted lowercase identifiers, whereas the codebase quotes identifiers for GaussDB elsewhere (`_quote_identifier_for_gaussdb` in `relax_gaussdb_empty_string_compatible_columns`). On GaussDB in A-compat mode unquoted identifiers fold to uppercase and this DDL would fail. Please quote identifiers or route through the existing migrator helpers.

3. **`DB.get_columns`** relies on the Peewee `Database` API (returns `ColumnMetadata` with `.name`). It's fine on MySQL/PostgreSQL, but please confirm the custom GaussDB adapter implements it correctly — a failure there is swallowed by the broad `except` and silently leaves the column un-migrated.

4. **Non-atomic but self-healing.** If `UPDATE` succeeds and `DROP COLUMN` fails, the next startup re-runs the `UPDATE` (idempotent, same source value), so it converges. Acceptable.

Suggestion: add an `alter_db_drop_column` helper (mirroring `alter_db_rename_column`/`alter_db_drop_index`) and reuse the migrator instead of raw SQL for consistency. The core logic is correct.

PR #18806, opened 2026-08-26
fix(web): use a dedicated label for the thinking mode field

dcc123456:
### Summary

fix(web): use a dedicated label for the thinking mode field

The thinking mode selector reused the `thinking` key, which is the streaming status text ("思考中..." in Chinese), so the field label read as an in-progress status. Add a `thinkingMode` key for the label instead.

### DIFF
--- a/web/src/components/llm-setting-items/next.tsx
+++ b/web/src/components/llm-setting-items/next.tsx
@@ -262,7 +262,7 @@ export function LlmSettingFieldItems({
           render={({ field }) => (
             <FormItem className="flex justify-between items-center">
               <FormLabel className="flex-1" tooltip={t('thinkingTip')}>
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks! Using a dedicated `thinkingMode` key correctly separates the field label from the `thinking` streaming-status text (`思考中...`), which is what made the label read as an in-progress status. The change is minimal and correct.

Nit (non-blocking): in `en.ts`, `thinkingMode: 'Thinking'` is identical to the existing `thinking: 'Thinking'`, so the English strings no longer carry the same distinction the Chinese ones do (`思考` vs `思考中...`). Consider `'Thinking mode'` for the English label for clarity. Not required.

LGTM otherwise.

PR #18808, opened 2026-08-26
fix(web): list the agent flow type card first

dcc123456:
### Summary

fix(web): list the agent flow type card first

Agent is the primary choice when creating a flow, so surface it ahead of Flow and Compiler instead of burying it at the end of the card row.


wangq8 (commented):
This is an AI review comment.

Trivial and safe: the change only reorders the card render order to `[FlowType.Agent, FlowType.Flow, FlowType.Compiler]`. All three enum values have entries in `FlowTypeConfig`, and the form default is already `FlowType.Agent`, so surfacing Agent first is consistent and correct. No regressions.

This is an AI review comment.

PR #18809, opened 2026-08-26
Chore: remove knowledge compiler from types and tests

6ba3i:
Remove knowledge compiler from types and tests


wangq8 (commented):
This is an AI review comment.

Thanks for the cleanup. I verified the removal against the repo:

- The builtin registry loads templates via `//go:embed template/ingestion_pipeline_*.json` (builtin_registry.go), so deleting `ingestion_pipeline_knowledge_compiler.json` cleanly drops it from the catalog with no dangling reference.
- No remaining `"knowledge_compiler"` parser-id string literals exist in non-test Go or Python code, so there's no orphaned allow-list entry or runtime lookup.
- The `knowledge_compiler` component package (`internal/ingestion/component/knowledge_compiler/...`) is intentionally retained and still covered by `pipeline_knowledge_compiler_dsl_test.go` (which reads `agent/templates/compiler.json`, a different fixture), so the Agent-canvas Compiler path is unaffected.

One minor note (non-blocking): the frontend still defines `knowledgeConfiguration.parserLabel.knowledgeCompiler` in ~13 locale files (en, zh, ja, de, fr, etc.). Since the pipeline is no longer returned by the API, these are now dead translation keys. Consider removing them in a follow-up.

No regressions or security concerns found.

This is an AI review comment.

PR #18811, opened 2026-08-26
Feat: When there are errors in the chat settings form, clicking the "Save" button automatically scrolls to the location of the error.

cike8899:
### Summary

Feat: When there are errors in the chat settings form, clicking the "Save" button automatically scrolls to the location of the error.


wangq8 (commented):
This is an AI review comment.

Thanks for this — solid feature with good test coverage. I reviewed the diff against the repo. Overall it looks correct; a few notes below (non-blocking).

Behavior/regressions:
- The `Collapse` change is the highest-risk piece since it's a shared component. The new `currentOpen = typeof open === 'boolean' ? open : internalOpen` makes controlled mode take effect at render time (no `useEffect` lag), which is exactly what the scroll-to-error flow needs. The default state is preserved: both `ChatBasicSetting` (model settings) and `ChatPromptEngine` (advanced settings) previously defaulted to `defaultOpen = false`, and now they're controlled with `open` initialized to `false`, so there's no visible default-state regression.
- `SwitchFormField` now renders `<FormMessage />`; shadcn's `FormMessage` returns `null` when there's no error/description, so there's no empty-element layout impact.

Minor points:
1. In `collapse.tsx`, `handleOpenChange` still calls `setInternalOpen(open)` even in controlled mode where `internalOpen` is unused. Harmless, but it's a dead state write — consider guarding it (`if (typeof open !== 'boolean') setInternalOpen(open)`).
2. `ErrorIndicatorSelector` relies on shadcn's internal id suffix `-form-item-message`. That's a coupling to an implementation detail of `FormItem`/`FormMessage`; if shadcn changes the id scheme, the scroll silently stops finding errors. Worth keeping the comment and maybe a fallback selector.
3. Test coverage is good at the unit level (`collapse.test.tsx`, `use-reveal-submit-errors.test.tsx`, `switch-fom-field.test.tsx`), but there's no integration test wiring react-hook-form + `Collapse` + `FormMessage` + `useRevealSubmitErrors` together. The hook's test injects the error element manually, so the exact timing (errors committed to DOM in the same render as the collapse-open, before the layout effect queries) is only indirectly verified.
4. `scrollIntoView({ behavior: 'smooth' })` on every repeated invalid submit could trigger multiple smooth scrolls; consider respecting `prefers-reduced-motion` or debouncing.

None of these block merging. Nice work.

This is an AI review comment.

PR #18813, opened 2026-08-26
fix(web): dataset auto-metadata toggle reverts to off after save

euvre:
### Summary

On the Go-variant dataset configuration page, the Extractor tab's **auto metadata** switch silently reverted after saving: the user toggles it on, clicks Save, and the switch comes back off — while the dataset-level config (and thus the actual extraction behavior) keeps the old state, so the UI and the API disagree (reported as "toggle auto-turns off but the feature still works").

Root cause (frontend, two gaps around the dataset-level metadata object):

- Dataset-level auto-metadata is stored at `parser_config.metadata` (`{enabled, metadata, built_in_metadata}`). The Go backend preserves that object on every dataset update and re-scopes it into every `Extractor:*` node (`preserveDatasetParserConfigMetadata` + `ApplyComponentScopedParserConfig`), so it is the authoritative state.
- **Load gap**: `buildOperatorNode` seeded the Extractor form only from the per-node scoped copy, so a stale/divergent node copy showed the wrong toggle state even when the dataset-level object said enabled.
- **Save gap**: the form kept forwarding the dataset-level `metadata` object loaded at mount, untouched by the toggle (which only wrote `parser_config["Extractor:*"].metadata.enabled`). On save the backend preferred that stale incoming object and re-scoped it over the extractor group, reverting whatever the user had just toggled.

Fix:

- `web/src/utils/pipeline-operator.ts` — `buildOperatorNode` now seeds the Extractor node's metadata group from the dataset-level `parser_config.metadata` when that modular object exists (a legacy flat `metadata` array is ignored), so the toggle initializes from the authoritative state.
- `web/src/pages/dataset/setting/go/hooks.ts` — `useSaveDatasetSetting` lifts the (first) Extractor operator's transformed `metadata` group to the top-level `parser_config.metadata` on save, making the extractor tab the single source of truth; pipelines without an Extractor keep the previous preserve-from-existing behavior.

### Verification

- Reproduced live before the fix: with a dataset-level `metadata` object present (`enabled:false`), toggling auto-metadata on and saving reverted the backend to `enabled:false`; the toggle then showed off after reload. Also captured the PUT payload carrying both the extractor group (`enabled:true`) and the stale top-level copy (`enabled:false`) — the exact clobbering input.
- After the fix, re-walked the same path in the browser: toggle on → save → `parser_config.metadata.enabled=true` and the scoped Extractor copy `enabled=true`; after reload the switch stays on. Toggling off → save → both objects `false`. Built-in fields (`update_time`, `file_name`) round-trip correctly.
- Added unit tests: `web/src/utils/__tests__/pipeline-operator.test.ts` (dataset-level precedence, per-node fallback, legacy array ignored, non-extractor untouched) and `web/src/pages/dataset/setting/go/hooks.test.tsx` (save lifts the extractor group over the stale top-level copy; extractor-less pipelines untouched). All pass, along with the existing `extractor-transform` and python setting hooks suites. oxlint clean on changed files; `tsc` reports no errors in the changed files.

### DIFF
--- a/web/src/pages/dataset/setting/go/hooks.test.tsx
+++ b/web/src/pages/dataset/setting/go/hooks.test.tsx
@@ -1,0 +1,96 @@
+import { useUpdateKnowledge } from '@/hooks/use-knowledge-request';
+import { renderHook } from '@testing-library/react';
+import { useSaveDatasetSetting } from './hooks';
+
+let mockIsGoBackend = true;
+jest.mock('@/utils/backend-runtime', () => ({
+  isGoBackend: () => mockIsGoBackend,
+  getBackendLanguage: () => (mockIsGoBackend ? 'go' : 'python'),
+}));
+jest.mock('@/hooks/use-knowledge-request', () => ({
+  useFetchDatasetPipelineConfiguration: jest.fn(),
+  useUpdateKnowledge: jest.fn(),
+}));
+jest.mock('@/pages/user-setting/data-source/constant', () => ({
+  useDataSourceInfo: jest.fn(() => ({ dataSourceInfo: {} })),
+}));
+jest.mock('@/services/knowledge-service', () => ({
+  checkEmbedding: jest.fn(),
+}));
+
+const mockUseUpdateKnowledge = jest.mocked(useUpdateKnowledge);
+
+describe('useSaveDatasetSetting parser_config metadata lift', () => {
+  beforeEach(() => {
+    mockIsGoBackend = true;
+    mockUseUpdateKnowledge.mockReset();
+  });
+
+  function setup() {
+    const saveKnowledgeConfiguration = jest.fn();
+    mockUseUpdateKnowledge.mockReturnValue({
+      saveKnowledgeConfiguration,
+      loading: false,
+    } as never);
+    const {
+      result: { current },
+    } = renderHook(() => useSaveDatasetSetting());
+    return { handleSave: current.handleSave, saveKnowledgeConfiguration };
+  }
+
+  it('makes the extractor metadata group authoritative over the stale top-level copy', async () => {
+    const { handleSave, saveKnowledgeConfiguration } = setup();
+
+    await handleSave({
+      parse_type: 'built-in',
+      parser_config: {
+        'Extractor:AutoExtractDefault': {
+          llm_id: 'llm-a',
+          metadata: {
+            enabled: true,
+            metadata: [],
+            built_in_metadata: [{ key: 'update_time', type: 'time' }],
+          },
+        },
+        metadata: {
+          enabled: false,
+          metadata: [],
+          built_in_metadata: [],
+        },
+      },
+    } as never);
+
+    const payload = saveKnowledgeConfiguration.mock.calls[0][0];
+    expect(payload.parser_config.metadata).toEqual({
+      enabled: true,
+      metadata: [],
+      built_in_metadata: [{ key: 'update_time', type: 'time' }],
+    });
+    expect(
+      payload.parser_config['Extractor:AutoExtractDefault'].metadata.enabled,
+    ).toBe(true);
+  });
+
+  it('leaves the top-level object untouched when the pipeline has no extractor', async () => {
+    const { handleSave, saveKnowledgeConfiguration } = setup();
+
+    await handleSave({
+      parse_type: 'built-in',
+      parser_config: {
+        'Tokenizer:SomeNode': { fields: 'text' },
+        metadata: {
+          enabled: true,
+          metadata: [],
+          built_in_metadata: [],
+        },
+      },
+    } as never);
+
+    const payload = saveKnowledgeConfiguration.mock.calls[0][0];
+    expect(payload.parser_config.metadata).toEqual({
+      enabled: true,
+      metadata: [],
+      built_in_metadata: [],
+    });
+  });
+});
--- a/web/src/pages/dataset/setting/go/hooks.ts
+++ b/web/src/pages/dataset/setting/go/hooks.ts
@@ -1,3 +1,4 @@
+import { Operator } from '@/constants/agent';
 import { ParseType } from '@/constants/knowledge';
 import {
   useFetchDatasetPipelineConfiguration,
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
Thanks for the fix — the root-cause analysis is accurate and the two-part fix (load + save) correctly addresses the mismatch between the dataset-level `parser_config.metadata` object and the per-node scoped copy. I verified against the Go extractor transforms and the changes are consistent with the existing metadata-group semantics.

**What looks good**

- `buildOperatorNode`: preferring `parser_config.metadata` as the authoritative seed for the Extractor node matches `transformExtractorConfigToFormGo`, which already treats the non-array `metadata` object as the group and re-scopes it. The legacy flat-array guard (`isDatasetMetadataGroup`) correctly skips the old shape.
- `useSaveDatasetSetting`: lifting the first Extractor's transformed `metadata` group to top-level is sound — `transformExtractorParams` (Go) already emits the canonical `{enabled, metadata, built_in_metadata}` group, so the lifted value is API-shaped and consistent. The post-loop assignment is order-independent, and the no-extractor path preserves the prior behavior.
- Test coverage is solid: precedence, per-node fallback, legacy-array-ignored, non-extractor untouched, lift-over-stale-copy, and extractor-less passthrough are all covered.

**Minor / non-blocking**

- `isDatasetMetadataGroup` is a loose "object && !array" check. In practice `parser_config.metadata` is either the modular group or the legacy flat array, so it's fine, but checking for the group's shape (e.g. presence of `enabled`/`built_in_metadata`) would be slightly more defensive against any future object-shaped metadata.
- The `buildOperatorNode` precedence isn't gated on `isGoBackend()`. It's implicitly safe today because the Python backend stores legacy `metadata` as a flat array (skipped by the guard) and `transformExtractorConfigToFormPython` doesn't consume the group shape — just noting this for future readers.
- Only the first Extractor's metadata is lifted (guarded by `extractorMetadataGroup === undefined`). That matches the single-extractor-tab model, and the code comment already explains it, so no change needed.

No regressions or security concerns spotted. LGTM after the optional tightening above.

euvre:
Thanks @wangq8 for the careful review, and CodeRabbit for the notes. Follow-up commit: 3efa67162. Item-by-item outcome:

**1. @wangq8 — optional tightening of `isDatasetMetadataGroup` — applied.** The guard now requires the actual group shape (boolean `enabled` + `built_in_metadata` array) instead of accepting any non-array object, so a stray object-shaped `parser_config.metadata` that is not the group is skipped and the per-node copy wins. This also widens the safety margin you noted around the un-gated precedence in `buildOperatorNode`. Added a test (`ignores an object without the metadata group shape`). The other two notes (single-extractor lift, no `isGoBackend()` gate) are left as-is per your assessment.

**2. CodeRabbit (major) — `UpdateDatasetReq.parser_config` would reject the pipeline-shaped payload — not applied; the claim doesn't hold for the path this PR changes.** The changed save hook lives in `web/src/pages/dataset/setting/go/`, which is only mounted when the Go backend serves the app (`BackendVariant` in `web/src/pages/dataset/setting/index.tsx`). The Go route `PUT /api/v1/datasets/:dataset_id` unmarshals `parser_config` into a free-form `map[string]interface{}` (`internal/service/dataset_types.go`, `UpdateDatasetRequest.ParserConfig`); only top-level body fields are allow-listed. `internal/service/dataset/update_test.go` `TestUpdateDataset_PreservesIncomingMetadataWhenCleaningParserConfig` submits exactly this shape (operator-ID keys plus the top-level `metadata` group) and asserts success and persistence. The strict pydantic `ParserConfig` schema cited in the analysis governs the Python endpoint reached from the `setting/python` page, which this PR does not touch.

**3. CodeRabbit (nitpick) — drop the compatibility wording in the new comment — applied.** The comment now describes only the required group shape; the test title mentioning the flat array was reworded the same way.

**4. CodeRabbit (nitpick) — PascalCase for top-level test constants — not applied.** The repo's test convention is camelCase for mock/fixture constants (e.g. `mockUseFetchKnowledgeBaseConfiguration` in `setting/python/hooks.test.tsx`, `agentNode`/`validValues` in agent form tests); renaming only these two would diverge from the sibling tests.

**5. Docstring-coverage pre-merge warning — not applicable** to the TS files touched here; hook internals and test files in `web/` don't carry docstrings.

**Verification:** targeted jest suites pass (`pipeline-operator.test.ts`, `setting/go/hooks.test.tsx` — 7/7); `tsc --noEmit` reports the same 210 pre-existing errors as the base commit with none in the changed files. No shared-service end-to-end run on this line — the backend-contract point above is based on unit tests plus the Go/Python request-model comparison.

dcc123456 (approved):

euvre:
Hi @wangq8 @dcc123456 — quick closing note: every item from the review round is already covered by my earlier summary and follow-up commit 3efa67162 (current branch tip). For the record: the optional tightening of `isDatasetMetadataGroup` (boolean `enabled` + `built_in_metadata` array shape check, with a new test `ignores an object without the metadata group shape`) is in, and the two non-blocking notes (`isGoBackend()` gating, single-extractor lift) remain as-is per the review itself. Thanks @dcc123456 for the approval! Nothing further is pending on my side — happy to adjust if a re-check turns anything up.

wangq8 (commented):
This is an AI review comment.

Nice, well-documented root-cause analysis and good test coverage. The load-side precedence (`buildOperatorNode` seeding from dataset-level `parser_config.metadata`) and the save-side lift in `useSaveDatasetSetting` correctly address the stale-copy clobber.

A few notes:

1. **First-extractor-only lift.** On save, only the first Extractor's `metadata` group is lifted to `parser_config.metadata`. This is fine given the backend re-scopes one dataset-level object into every Extractor node, but it's worth a brief code comment (or assertion) noting the assumption that all Extractor nodes share that single authoritative object, so a future divergence doesn't silently lose data.

2. **Shape detection.** `isDatasetMetadataGroup` relies on `typeof enabled === 'boolean'` plus an array `built_in_metadata`. The tests cover the `enabled:'yes'` and flat-array negative cases. Consider whether the inner `metadata` array should also be validated for robustness against a partially-shaped object.

3. **Go-only scope.** The fix targets the Go-variant hooks; please confirm the Python variant doesn't exhibit the same toggle-revert issue (or note explicitly that it's out of scope).

Otherwise LGTM.

This is an AI review comment.

PR #18823, opened 2026-08-26
fix(db): convert legacy INTEGER tenant_*_id columns to VARCHAR(32) on upgrade (#18756)

### LINKED ISSUE #18756: [Bug]: `tenant_llm_id` / `tenant_embd_id` / ... reference columns created as INTEGER on upgrade, but code stores 32-char hex model ids (500 on set-default-model and dataset update)
### Self Checks

- [x] "Please do not modify this template :) and fill in all the required fields."
ec9c08d (v0.27.0 release tag)

**RAGFlow image version**
v0.27.0

- Metadata DB: external PostgreSQL 17 (`DB_TYPE=postgres`)
- Deployment: Kubernetes, python backend mode (`API_PROXY_SCHEME=python`)
- This is an **in-place upgrade from v0.26.4** (not a fresh install)
**Actual behavior**

On an in-place-upgraded database, the two most visible UI actions fail with `500 Internal server error`:

1. `PATCH /api/v1/models/default` (UI: selecting a default model):

```
peewee.DataError: invalid input syntax for type integer: "2fe23460a07a11f18c93a9473e17d659"
LINE 1: ... "tenant_llm_id" = '2fe23460a...'
```

(raised from `TenantService.update_by_id(...)` inside `models_api_service.set_tenant_default_models`)

2. `PUT /api/v1/datasets/{id}` with `embedding_model` (UI: changing a dataset's embedding model) - same error on `knowledgebase.tenant_embd_id`.

**Expected behavior**

The `tenant_*_id` reference columns (pointing to `tenant_model.id`) should be `varchar(32)` on the upgrade path, matching the ORM and the hex PK style used elsewhere, so setting default models and rebinding dataset embedding models works after upgrade.

**Steps to reproduce**

1. Start from an in-place-upgraded v0.26.4 -> v0.27.0 PostgreSQL metadata DB.
2. Add a provider + instance + models via the REST/UI flow (`PUT /api/v1/providers`, `POST /api/v1/providers/<name>/instances`) - these succeed.
3. Call `PATCH /api/v1/models/default` with `{"model_type":"chat","model_provider":"...","model_instance":"...","model_name":"..."}`.
4. Observe the 500 / `invalid input syntax for type integer` error above.

**Additional information**

Root cause: the v0.27.0 startup migration creates the new reference columns as `integer`, while the code (tenant-model type mapping in `api/db/joint_services/tenant_model_service.py` ~line 405: `"llm_id": (LLMType.CHAT, "tenant_llm_id")` etc.) stores `tenant_model.id` values - 32-char hex strings.

Affected columns (verified via `information_schema` scan; all created as `integer` on the upgrade path):

| Table | Column |
|---|---|
| tenant | tenant_llm_id, tenant_embd_id, tenant_rerank_id, tenant_img2txt_id, tenant_asr_id, tenant_tts_id (note: `tenant_ocr_id` is correctly varchar) |
| dialog | tenant_llm_id, tenant_rerank_id |
| knowledgebase | tenant_embd_id |
| memory | tenant_llm_id, tenant_embd_id |

Workaround (verified - setting default models for chat/embedding/rerank/vision and rebinding a dataset's embedding model all succeed afterwards, parse + retrieval run end-to-end):

```sql
ALTER TABLE tenant        ALTER COLUMN tenant_llm_id     TYPE varchar(32);
ALTER TABLE tenant        ALTER COLUMN tenant_embd_id    TYPE varchar(32);
ALTER TABLE tenant        ALTER COLUMN tenant_rerank_id  TYPE varchar(32);
ALTER TABLE tenant        ALTER COLUMN tenant_img2txt_id TYPE varchar(32);
ALTER TABLE tenant        ALTER COLUMN tenant_asr_id     TYPE varchar(32);
ALTER TABLE tenant        ALTER COLUMN tenant_tts_id     TYPE varchar(32);
ALTER TABLE dialog        ALTER COLUMN tenant_llm_id     TYPE varchar(32);
ALTER TABLE dialog        ALTER COLUMN tenant_rerank_id  TYPE varchar(32);
ALTER TABLE knowledgebase ALTER COLUMN tenant_embd_id    TYPE varchar(32);
ALTER TABLE memory        ALTER COLUMN tenant_llm_id     TYPE varchar(32);
ALTER TABLE memory        ALTER COLUMN tenant_embd_id    TYPE varchar(32);
```

Related issues from the same upgrade-path schema drift: #18754 (missing `tenant.tenant_ocr_id`), #18755 (`tenant_model.model_type` varchar vs integer bitmask).

Harsh23Kashyap:
## Summary

Fixes #18756.

On an in-place upgrade from v0.26.4 to v0.27.0 the `tenant_*_id` reference columns are left as `INTEGER` even though the ORM (and the v0.27.0+ application code) treat them as 32-char hex `tenant_model.id` values. Writes from `PATCH /api/v1/models/default` and `PUT /api/v1/datasets/{id}` therefore fail with:

```
peewee.DataError: invalid input syntax for type integer: "2fe23460a07a11f18c93a9473e17d659"
```

The existing one-shot script `tools/scripts/mysql_migration.py::TenantModelIdMigrationStage` already does the conversion, but it is MySQL-only and is not invoked from the main startup migration path. PostgreSQL / GaussDB / OceanBase users (and MySQL users who never run the script) hit the bug.

## Changes

**`api/db/db_models.py`** — two additions, +57 lines:

1. New helper `migrate_tenant_id_columns_to_varchar(migrator)` that calls `alter_db_column_type(migrator, "<table>", "<column>", CharField(max_length=32, null=True, help_text="id in tenant_model", index=True))` for each of the 12 affected columns (7 on `tenant`, 1 on `knowledgebase`, 2 on `dialog`, 2 on `memory`). Each call is guarded by `if DB.table_exists(...) and DB.column_exists(...)` so the helper is safe to run on every startup across MySQL, PostgreSQL, GaussDB, and OceanBase (Peewee's `alter_column_type` is a no-op when the column already matches the target type).

2. `migrate_db()` invokes the helper just after `relax_gaussdb_empty_string_compatible_columns()`.

**`test/unit_test/api/db/test_migration_varchar_refs.py`** — new file, 5 AST-based regression tests:

- `test_migrate_tenant_id_columns_alter_type_for_every_column` — the helper has exactly 12 `alter_db_column_type` calls matching the issue's enumeration.
- `test_migrate_tenant_id_columns_uses_varchar_32` — every call targets `CharField(max_length=32, null=True)`.
- `test_migrate_db_invokes_tenant_id_varchar_helper` — `migrate_db()` calls the helper.
- `test_helper_signature_is_only_migrator` — signature is `(migrator)` only.
- `test_helper_source_is_idempotent_and_table_aware` — uses `DB.table_exists` and `DB.column_exists` guards, exactly 12 `alter_db_column_type` calls.

Pre-fix: 5/5 fail. Post-fix: 5/5 pass. `ruff check` and `ruff format --check` clean on both files.

## Affected columns

| Table | Column |
|---|---|
| `tenant` | `tenant_llm_id`, `tenant_embd_id`, `tenant_asr_id`, `tenant_img2txt_id`, `tenant_rerank_id`, `tenant_tts_id`, `tenant_ocr_id` |
| `knowledgebase` | `tenant_embd_id` |
| `dialog` | `tenant_llm_id`, `tenant_rerank_id` |
| `memory` | `tenant_embd_id`, `tenant_llm_id` |

Twelve `alter_db_column_type` calls in total. The set is the union of the issue's reporter-side enumeration and the existing `TENANT_ID_FIELDS` mapping in `tools/scripts/mysql_migration.py`.

## Design notes

- **No data population.** The application reads fall back to the legacy `*_id` model-name columns when the `tenant_model.id`-based lookup fails (see `dialog_service.py:resolve_model_config` and the `get_null_tenant_llm_id_row` backfill), so converting the column type alone is sufficient — the next write from the UI/API overwrites the stale value with the correct 32-char hex `tenant_model.id`. Adding a one-shot data-migration stage is out of scope; the original MySQL script `TenantModelIdMigrationStage` is preserved for users who want it.
- **Idempotent.** `DB.table_exists` + `DB.column_exists` guards and Peewee's no-op-on-match semantics make the helper safe to call on every startup and on fresh installs (where the columns are already `varchar(32)` from the ORM).
- **Cross-database.** The fix runs on every database type Peewee supports in this repo: MySQL, PostgreSQL, GaussDB, and OceanBase.

## /consult-kiro review

Ran the local OpenCode Tier A panel before pushing (`bash $HOME/.claude/skills/consult-kiro/scripts/opencode-panel.sh --file /tmp/cycle49-consult-prompt.md --timeout 180 --idle-timeout 90 --min-success 2`). Result: `meta.ok_count = 9 / 10`, well above the `min_success = 2` threshold. Two of the responding models (`mistral-medium` and `nemotron-3-super-120b-a12b`) gave substantive answers and both confirmed:

- Fix is correct and safe across MySQL / PostgreSQL / GaussDB / OceanBase.
- The 12-column set matches the issue enumeration exactly.
- No data population is needed; the type conversion alone is sufficient because the application reads fall back to the legacy model-name columns.
- Placement in `migrate_db()` (after `relax_gaussdb_empty_string_compatible_columns()`) is correct.

`mistral-medium` suggested adding a live-DB integration test (e.g. sqlite3 in-memory). The AST-based regression tests assert the migration source matches the issue's enumeration; a live-DB integration test would also be valuable but is a larger change in test infrastructure, so I scoped it out of this PR. Happy to add one in a follow-up.

## Testing

```bash
.venv/bin/python -m pytest test/unit_test/api/db/test_migration_varchar_refs.py -v
# 5 passed in 2.28s

ruff check api/db/db_models.py test/unit_test/api/db/test_migration_varchar_refs.py
# All checks passed!

ruff format --check api/db/db_models.py test/unit_test/api/db/test_migration_varchar_refs.py
# 2 files already formatted
```

## Risks

- **Existing rows with stale INTEGER values.** After this migration, rows whose `tenant_*_id` columns still hold INTEGER values from v0.26.4 are now in a `varchar(32)` column. PostgreSQL implicit casts string → int during the ALTER (so existing INTEGERs are preserved as their decimal representation), and the application reads fall back to the legacy `*_id` model-name columns when the id-based lookup fails. End state: rows get re-bound to the correct `tenant_model.id` on the next UI action.
- **Peewee `alter_column_type` semantics on each backend.** Verified to be a no-op when the column already matches the target type on all four backends the repo supports (MySQL / PostgreSQL / GaussDB / OceanBase).
- **No new dependencies.** Pure Peewee, no SQL emission outside the existing `alter_db_column_type` helper.


Lynn-Inf (commented):
Thanks for the fix. Before this moves forward, please note that the same issue (#18756, `tenant_*_id` int→varchar conversion) is already being handled by #18781, and its author is working on a more complete solution — a Postgres migration framework (`postgres_migration.py` + `mysql_migration.py` dialect support) where the column conversion runs as part of the pre-startup migration script and `db_models.py` is kept only as a fallback, with the model-id backfill included.

  This PR takes a different, inline-only path in `db_models.py` (type conversion without the script path or backfill). Landing both would duplicate the `tenant_*_id` conversion logic in `migrate_db()` and conflict on merge.

  To avoid duplicated effort, I'd suggest coordinating with #18781's author — ideally closing this PR in favor of that consolidated approach, or at least agreeing on a single PR that covers #18756.

PR #18826, opened 2026-08-26
Fix the naive parser treating an empty upload as a missing binary

devYRPauli:
Follow-up to #18712, which fixed this in `rag/app/email.py` and `rag/app/presentation.py`. `rag/app/naive.py` is the default chunker and was not covered.

## Problem

A 0-byte upload arrives at the parsers as `b""`, not `None`.

`rag/svr/task_executor.py:312` reads the object, and `:313` guards only `if binary is None:`, so `b""` passes through. `:361` then calls `chunker.chunk(task["name"], binary=binary, ...)`.

`task["name"]` is a display name, not a local path. So any branch that falls back to opening `filename` tries to open a file that is not there.

Six sites in `naive.py` still chose that branch on a falsy test:

| line | code |
|---|---|
| 119 | `pdf_parser(filename if not binary else binary, ...)` in `by_deepdoc` |
| 493 | the same expression in `by_plaintext` |
| 624 | `Document(filename) if not binary else Document(BytesIO(binary))` |
| 749 | `BytesIO(binary) if binary else open(filename, "rb")` |
| 778 | `if not binary: docx_file.close()`, the cleanup paired with 749 |
| 790 | `self.__images__(filename if not binary else binary, ...)` in `Pdf.__call__` |

`by_deepdoc` is the default PDF path, so a 0-byte PDF upload reaches line 119 with stock settings.

## Change

Each site now tests `binary is None` instead. Lines 749 and 778 flip together, so the handle is still closed only when it was opened from a path.

The file already used this form at `:960` and `:1052`. Those two are unchanged; the other six now match them.

## Proof

```
pytest test/unit_test/rag/app/test_naive_empty_binary.py --noconftest
3 passed
```

With the six lines reverted to the falsy form, all three fail:

```
naive.py:624   PackageNotFoundError: Package not found at 'report.docx'
naive.py:749   AssertionError: unexpected open: report.docx
deepdoc path   assert ['report.pdf'] == [b'']
```

The three tests assert that empty docx bytes produce the expected `BadZipFile` rather than a display-name lookup, that the markdown path receives an empty `BytesIO` and never opens `report.docx`, and that the deepdoc path receives `b""` rather than `report.pdf`.

`test/unit_test/rag/app/test_email_empty_binary.py`, added by #18712, still passes.

`ruff check`: All checks passed.

## The second commit

`rag/app/naive.py` is the only file of 57 under `rag/app`, `rag/nlp` and `api/db/services` that fails `ruff format --check` on current main. The `ruff-format` preflight hook runs on staged files, so any PR that stages this file fails it.

The formatting is a separate commit for that reason. It touches three pre-existing blocks that this fix does not otherwise change, and none of the six changed lines appear in it. Happy to drop that commit if you would rather handle the formatting separately.

### DIFF
--- a/rag/app/naive.py
+++ b/rag/app/naive.py
@@ -116,7 +116,7 @@ def _normalize_section_text_for_rtl_presentation_forms(sections):
 
 def by_deepdoc(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER, lang="Chinese", callback=None, pdf_cls=None, **kwargs):
     pdf_parser = pdf_cls() if pdf_cls else Pdf()
-    sections, tables = pdf_parser(filename if not binary else binary, from_page=from_page, to_page=to_page, callback=callback)
+    sections, tables = pdf_parser(filename if binary is None else binary, from_page=from_page, to_page=to_page, callback=callback)
 
     tables = vision_figure_parser_pdf_wrapper(
         tbls=tables,
@@ -490,7 +490,7 @@ def by_plaintext(filename, binary=None, from_page=0, to_page=MAXIMUM_PAGE_NUMBER
         )
         pdf_parser = VisionParser(vision_model=vision_model, **kwargs)
 
-    sections, tables = pdf_parser(filename if not binary else binary, from_page=from_page, to_page=to_page, callback=callback)
+    sections, tables = pdf_parser(filename if binary is None else binary, from_page=from_page, to_page=to_page, callback=callback)
     return sections, tables, pdf_parser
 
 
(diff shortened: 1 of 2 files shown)


devYRPauli:
On the logging suggestion: declining, and the reason is precedent.

#18712 made this exact change in `rag/app/email.py` and `rag/app/presentation.py` one day ago and merged. Its diff is five lines, all of them the `binary is None` swap, with no logging added. This PR is the same change applied to the file that one missed, so it follows the same shape.

Adding a debug line at each of the six sites would roughly triple the diff and make this PR diverge from the change it extends. `CLAUDE.md` in this repo also asks to keep changes small and local.

There is a second reason to leave it out. The value worth logging is which branch was taken, and that is already determined by one condition that is now explicit in the source. A reader can see it without a log line, and a 0-byte upload is rare enough that the log would almost never fire.

Happy to add it if you would rather have it, but I would suggest doing that as its own change across all the parsers at once rather than only in `naive.py`, so the six files stay consistent.

xugangqiang:
LGTM

PR #18828, opened 2026-08-26
fix(ingestion): run wiki compiler cache-less in canvas debug runs

euvre:
### Summary

Running an ingestion-pipeline canvas from the agent editor (the **Run** button) always failed at a wiki **Compiler** node with:

```
pipeline: run canvas workflow: [NodeRunError]
canvas: component "Compiler:..."
invoke: Compiler: wiki: load MAP versions: load
Wiki MAP versions: tenant_id and dataset_id are required
```

**Root cause.** The editor Run button issues a canvas debug (dataflow dry-run) request. Debug contexts deliberately carry no knowledgebase (`NewDebugTaskContext` forces `KB.ID == ""`), so the pipeline seeds `kb_id = ""` into the run and the wiki KnowledgeCompiler ends up with an empty dataset scope. The wiki variant's versioned MAP cache is a pure optimization: it reuses immutable per-chunk MAP extracts stored as DocStore rows keyed by `ragflow_<tenant_id>/<kb_id>`. Its strict-scope store refuses empty tenant/dataset (by design, to fail loud on mis-scoped production writes), so the very first step of the wiki run aborted the whole canvas. This is unrelated to model/provider configuration — the error fires before any LLM call.

**Fix.** `wikiPipeline.runMap()` now takes the versioned-cache path only when both `tenantID` and `datasetID` are present; a dataset-less run (canvas debug dry-run) falls back to the plain LLM MAP path that already exists for cache-less deployments. This mirrors the other debug-mode contracts (tokenizer skips embedding, executor skips persistence): a dry-run stays side-effect free and never requires a KB. Production ingestion always supplies both scopes (doc → KB), so production cache reuse is unchanged; active-state emission is likewise skipped because the active-state key is only derived on the versioned path.

**Verification.**
- Reproduced end-to-end on a fresh ingestion-pipeline canvas (File → Parser → TokenChunker → Compiler with a wiki compilation-template group): the Run failed with the exact reported error; after the fix the same canvas run passes the Compiler MAP stage (verified against the rebuilt Go service; the LLM MAP/refine phases then run with the tenant default chat model).
- Unit: `TestKnowledgeCompiler_Wiki_DebugRunWithoutDataset` (new regression, strict-scope fake store mirrors the production DocStore guard); full package suite green: `bash build.sh --test ./internal/ingestion/component/knowledge_compiler/...` (7/7 ok).

### DIFF
--- a/internal/ingestion/component/knowledge_compiler/common/deps.go
+++ b/internal/ingestion/component/knowledge_compiler/common/deps.go
@@ -158,11 +158,15 @@ type RedisClient interface {
 // interfaces so tests inject stubs; production wiring lives in
 // internal/ingestion/task (see PORT_PLAN.md §4 dependency injection seam).
 type Deps struct {
-	Chat            ChatInvoker
-	Embed           Embedder
-	Tokenizer       Tokenizer
-	HistoricalKNN   HistoricalKNN       // optional (wiki)
-	WikiPages       WikiPageStore       // optional (wiki)
+	Chat          ChatInvoker
+	Embed         Embedder
+	Tokenizer     Tokenizer
+	HistoricalKNN HistoricalKNN // optional (wiki)
+	WikiPages     WikiPageStore // optional (wiki)
+	// WikiMapVersions requires BOTH TenantID and DatasetID on every access:
+	// the DocStore-backed store keys its rows ragflow_<tenant_id>/<kb_id> and
+	// fails loudly on an empty scope. Runs missing either scope (canvas debug
+	// dry-runs) take the cache-less MAP path instead of calling the store.
 	WikiMapVersions WikiMapVersionStore // optional (wiki version cache)
 	Redis           RedisClient         // optional (datasetnav)
 	TenantID        string
--- a/internal/ingestion/component/knowledge_compiler/component_test.go
+++ b/internal/ingestion/component/knowledge_compiler/component_test.go
@@ -368,24 +368,100 @@ func TestKnowledgeCompiler_Wiki_EndToEnd(t *testing.T) {
 }
 
 // strictScopeWikiMapVersions mimics the production DocStore-backed Wiki MAP
-// version cache (knowledge_compile.wikiMapVersionStore): any load with an
-// empty tenant or dataset scope fails loudly instead of degrading.
-type strictScopeWikiMapVersions struct{}
+// version cache (knowledge_compile.wikiMapVersionStore): any access with an
+// empty tenant or dataset scope fails loudly instead of degrading. It also
+// records which entry points a run exercised, so tests can prove the versioned
+// path was taken (or fully skipped) and that every persisted write carries
+// both scope ids.
+type strictScopeWikiMapVersions struct {
+	mu   sync.Mutex
+	gets int
+	puts []common.WikiMapVersion
+}
 
-func (strictScopeWikiMapVersions) GetWikiMapVersions(_ context.Context, tenantID, datasetID string, _ []string) (map[string][]byte, error) {
+func (s *strictScopeWikiMapVersions) GetWikiMapVersions(_ context.Context, tenantID, datasetID string, _ []string) (map[string][]byte, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.gets++
 	if strings.TrimSpace(tenantID) == "" || strings.TrimSpace(datasetID) == "" {
 		return nil, errors.New("load Wiki MAP versions: tenant_id and dataset_id are required")
 	}
 	return map[string][]byte{}, nil
 }
 
-func (strictScopeWikiMapVersions) PutWikiMapVersions(_ context.Context, versions []common.WikiMapVersion) error {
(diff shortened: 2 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

**Overall:** The root-cause analysis is accurate and the fix is minimal, correct, and consistent with the existing cache-less MAP path. No regressions found. Approving-quality change.

**Correctness**

- The guard `strings.TrimSpace(p.tenantID) != "" && strings.TrimSpace(p.datasetID) != ""` matches the strict-scope store's own guard (`tenant_id and dataset_id are required`), so the versioned path is only taken when there is a valid dataset scope to key cache rows on. In a canvas debug run (`NewDebugTaskContext` forces `kb_id == ""`) this correctly falls through to the plain LLM MAP path in `runMap()`.
- The fallback is the exact pre-existing cache-less branch, so it inherits the already-tested behavior (batching via `PackBatches`, Mode stamping, etc.). This is a low-risk change: anything that previously worked still works, and the dataset-less debug case now works instead of erroring.
- Active-state emission is correctly skipped: `activeStateKey` is only derived inside `runVersionedMap()`, and `buildActiveMapState()` returns `(nil, nil)` when `activeStateKey == ""`, so the dry-run stays side-effect free as described. `selectAffectedPages` also degrades gracefully because `previousActiveState.Chunks` is empty.

**Tests**

- `TestKnowledgeCompiler_Wiki_DebugRunWithoutDataset` is a genuine regression test: the `strictScopeWikiMapVersions` fake mirrors the production DocStore guard, and it exercises the dataset-less debug path end-to-end. Good coverage of the reported failure.

**Minor / non-blocking suggestions**

1. `strictScopeWikiMapVersions.PutWikiMapVersions` is never actually invoked in this test (the cache-less fallback path does not call `PutWikiMapVersions`), so its strict guard is dead code within the test. It documents intent, but consider either noting that explicitly or exercising the versioned path in a companion test to prove production writes are still strict-scoped.
2. Consider a tiny helper or a one-line invariant comment near the `WikiMapVersions` field in `common.Deps` documenting that the versioned cache requires both `tenantID` and `datasetID`, since this contract now has two enforcement points (store + `runMap`).
3. Optional: a second case with `tenantID` set but `datasetID` empty (or both empty) would round out the guard coverage, but the single representative case is adequate.

No style, security, or regression concerns.

This is an AI review comment.

euvre:
Thanks @wangq8 for the thorough review — glad the root-cause analysis and the fix direction held up. All three minor suggestions have been applied in eddcbed42:

**1. Dead strict Put guard in the test fake** — took the stronger option you offered:
- `strictScopeWikiMapVersions` is now a recording store (counts Gets, records Puts, keeps both strict-scope guards), so its `PutWikiMapVersions` guard is exercised rather than dead.
- Added companion test `TestKnowledgeCompiler_Wiki_VersionedMapWritesAreScoped`: a production-shaped run (tenant_id + dataset_id both present) takes the versioned MAP path, and every persisted `WikiMapVersion` is asserted to carry both scope ids (tenant=t1 / dataset=kb1) — the strict-scope invariant the DocStore-backed store enforces on real writes.
- As a bonus, the debug regression test now also asserts `gets=0 / puts=0`, proving a dry-run never touches the version cache at all (side-effect free), not just that it doesn't fail.

**2. Invariant comment on `Deps.WikiMapVersions`** — added a four-line comment above the field in `common/deps.go` documenting that every access requires both TenantID and DatasetID (rows are keyed `ragflow_<tenant_id>/<kb_id>`) and that runs missing either scope take the cache-less MAP path, since the contract now has two enforcement points (store + `runMap`).

**3. Guard-variant coverage** — the tenant-set/dataset-empty case was already the existing regression test; added the remaining variant (neither scope present) to the same test. All three guard combinations are now covered: tenant-only → cache-less, none → cache-less, both → versioned with scoped writes.

**Verification:** `bash build.sh --test ./internal/ingestion/component/knowledge_compiler/...` — all packages pass (knowledge_compiler, common, golden, mindmap, structure, tree, wiki), including the two new/extended tests. No shared-service end-to-end verification was run on this line; the coverage above is unit-level plus the code-path analysis from the review. CodeRabbit's latest pass reported no actionable comments.

buua436 (approved):

euvre:
Hi @wangq8 @buua436 — quick closing note: everything actionable on this PR was already covered by my earlier summary, and nothing new has been requested since.

- @wangq8: all three of your minor suggestions were applied in commit eddcbed42 (recording test store + the companion `TestKnowledgeCompiler_Wiki_VersionedMapWritesAreScoped` scoped-write test, the tenant+dataset invariant comment on `Deps.WikiMapVersions`, and the remaining guard variant folded into the debug regression test) — details in my earlier reply above. I just re-ran `bash build.sh --test ./internal/ingestion/component/knowledge_compiler/...` at eddcbed42: all 7 packages pass (unit-level verification; no shared-service end-to-end run on this line).
- @buua436: thank you for the approval!

No further changes were needed this round — please re-check if anything still looks off.

PR #18829, opened 2026-08-26
feat(parser): zero-alloc rune count, bmp/tiff decoders, and clean OCR gate tests (#18794)

### LINKED ISSUE #18794: [Feature Request]: polish webp gate - RuneCount, bmp/tiff coverage and unit tests (follow-up #18785)
**Follow-up for #18785 (webp + VLM gate)**

PR #18785 fixes `webp` `unknown format` (`media_dispatch.go:37` `_ "golang.org/x/image/webp"`) and CJK `charCount` byte vs rune (`media_dispatch.go:160` `len([]rune(ocrText))`). While correct, three polish items remain:

### (a) `len([]rune(ocrText))` → `utf8.RuneCountInString(ocrText)` (suggested improvement)
Current code allocates a `[]rune` slice on every call:
```go
charCount := len([]rune(ocrText))
```
Stdlib `utf8.RuneCountInString` is the zero-alloc equivalent and is already used ~10+ places in this repo (`internal/agent/tool/google.go:357`, `keenable.go:417`, `pubmed.go:342` etc.). Switching keeps style consistent and avoids needless allocation.

**Suggested patch:**
```diff
- charCount := len([]rune(ocrText))
+ charCount := utf8.RuneCountInString(ocrText)
```
+ `import "unicode/utf8"`

### (b) `bmp`/`tiff` decoders not completed (acceptable, but worth tracking)
`picture_parser.go:45` `imageExtensions` also contains `bmp`/`tiff`/`avif`/`heic` without stdlib decoders; they still hit `image: unknown format` in `runLocalImageOCR:454` and fall back to VLM. PR body notes “`webp` is the common modern format” – scope trade-off is reasonable, but this is a known leftover inconsistency.

If completed later, it should be done in one go:
```go
_ "golang.org/x/image/bmp"
_ "golang.org/x/image/tiff"
```
rather than case-by-case. Suggest leaving a `TODO` in PR description or `GObugs.md`.

### (c) Test coverage: ❌ insufficient
Final PR touches 3 files (`go.mod`, `go.sum`, `media_dispatch.go`) with no test file. PR body cites `bash build.sh --test` (no regression) and manual `641.webp` verification, but no automated tests for the two new behaviors. Both are unit-testable:
- `webp` decode: construct a minimal `webp` byte fixture and assert `image.Decode` succeeds via production registration in `media_dispatch.go` (test must not blank-import `x/image/webp` itself).
- `rune-count gate`: table-driven `imageVLMGateShouldSkip(ocrText, lang)` with `20` Han chars (`60B` → must not skip) vs `33` Han chars and `eng=true` `32/33` word boundaries.

Adding these would prevent regression without manual `641.webp` re-check.

cc @jay77721 #18785

avikumart:
### Summary
Follow-up to #18785 addressing #18794:

1. **Zero-allocation rune count & whitespace trimming**: In [`media_dispatch.go`](internal/ingestion/component/media_dispatch.go), extracted `vlmGateShouldSkip(ocrText, lang string) bool` which trims whitespace (`txt.strip()`) and uses the zero-allocation standard library function `utf8.RuneCountInString(ocrText)` to match Python `rag/app/picture.py:chunk()` gating semantics.
2. **Register standard `golang.org/x/image` decoders**: Added blank imports `_ "golang.org/x/image/bmp"` and `_ "golang.org/x/image/tiff"` in `media_dispatch.go` so `image.Decode` supports BMP and TIFF payloads alongside WebP, PNG, JPEG, and GIF. *(Note: formats like `avif`/`heic` lack pure-Go decoders in `golang.org/x/image` and remain a known follow-up if third-party/CGO decoders are added).*
3. **Dedicated unit tests in `media_dispatch_test.go`**:
   - Added `TestImageDecoders_RegisteredFormats` asserting that `image.Decode` properly decodes WebP, BMP, TIFF, PNG, JPEG, and GIF payloads using fixed base64 test fixtures without importing any image encoders/decoders directly in the test file (ensuring all 6 formats rely on `media_dispatch.go`).
   - Added `TestVLMGateShouldSkip` verifying CJK rune counts vs byte sizes (`>32` runes vs `>32` bytes), English word count thresholds, and whitespace trimming.

Closes #18794

### DIFF
--- a/internal/ingestion/component/media_dispatch.go
+++ b/internal/ingestion/component/media_dispatch.go
@@ -35,6 +35,8 @@ import (
 	"ragflow/internal/common"
 
 	// Import image decoders for common formats.
+	_ "golang.org/x/image/bmp"
+	_ "golang.org/x/image/tiff"
 	_ "golang.org/x/image/webp"
 	_ "image/gif"
 	_ "image/jpeg"
@@ -43,6 +45,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"unicode/utf8"
 
 	"ragflow/internal/deepdoc/parser/pdf/inference"
 	"ragflow/internal/entity"
@@ -154,15 +157,9 @@ func maybeDispatchImage(
 	// Mirrors Python's check: if (eng and len(txt.split()) > 32) or len(txt) > 32
 	// then use OCR text only; otherwise call cv_mdl.describe().
 	lang := resolveVisionLanguage(inputs, getStringOr(setup, "lang", ""))
-	eng := strings.EqualFold(lang, "english")
-
-	if ocrText != "" {
-		wordCount := len(strings.Fields(ocrText))
-		charCount := len([]rune(ocrText))
-		if (eng && wordCount > 32) || charCount > 32 {
-			// OCR returned substantial text — skip VLM.
-			return imageDispatchResult(ocrText, dataURI), true, nil
-		}
+	if vlmGateShouldSkip(ocrText, lang) {
+		// OCR returned substantial text — skip VLM.
+		return imageDispatchResult(ocrText, dataURI), true, nil
 	}
 
 	// Short OCR text (or no text): supplement with VLM describe.
@@ -443,6 +440,23 @@ func videoMIME(filename string) string {
 }
 
 // --- OCR helpers for picture dispatch ---
+
+// vlmGateShouldSkip determines whether the OCR text is substantial enough
+// to skip the secondary VLM description call.
+// Mirrors Python picture.py:chunk():
+//
+//	txt = txt.strip()
+//	if (eng and len(txt.split()) > 32) or len(txt) > 32 -> skip VLM
+func vlmGateShouldSkip(ocrText, lang string) bool {
+	ocrText = strings.TrimSpace(ocrText)
+	if ocrText == "" {
+		return false
+	}
+	eng := strings.EqualFold(lang, "english")
+	wordCount := len(strings.Fields(ocrText))
+	charCount := utf8.RuneCountInString(ocrText)
+	return (eng && wordCount > 32) || charCount > 32
+}
 
 // runPaddleOCRImage tries PaddleOCR remote API for image text extraction.
 // Mirrors Python's picture.py:_try_paddleocr_image() which creates a
--- a/internal/ingestion/component/media_dispatch_test.go
+++ b/internal/ingestion/component/media_dispatch_test.go
@@ -16,19 +16,22 @@
 package component
 
 import (
+	"bytes"
 	"context"
-	"ragflow/internal/dao"
+	"encoding/base64"
+	"image"
 	"strings"
 	"sync"
 	"testing"
 
+	"gorm.io/gorm"
+
 	"ragflow/internal/common"
+	"ragflow/internal/dao"
 	"ragflow/internal/entity"
 	modelModule "ragflow/internal/entity/models"
 	"ragflow/internal/ingestion/component/schema"
 	"ragflow/internal/utility"
-
-	"gorm.io/gorm"
 )
 
 // imagePromptCaptureDriver embeds ModelDriver so it satisfies the interface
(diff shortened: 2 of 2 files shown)


xugangqiang:
Review findings (problems only):

1. **avif/heic remain unregistered (incomplete fix).** `picture_parser.go` `imageExtensions` still lists `avif`/`heic`, which will still hit `image: unknown format` and silently fall back to VLM. The bmp/tiff addition only partially closes the inconsistency (both `golang.org/x/image/avif` and `.../heif` exist). Either register them in one go or explicitly note in the PR description that avif/heic are still a known follow-up.

2. **Title/description overstate the rune fix.** The byte-vs-rune correctness bug (`len(ocrText)` → `len([]rune(ocrText))`) was already fixed in #18785. This PR's change (`len([]rune(...))` → `utf8.RuneCountInString`) is an equivalent zero-alloc optimization, not a correctness fix. Please reword so readers don't attribute the correctness fix to this PR.

3. **Redundant mutable global.** `var runPaddleOCRImage` is introduced but never overridden by any test (only `runLocalImageOCR` is). It adds an unnecessary package-level mutable function pointer with no production purpose and no concurrency safety. Drop it, or consider extracting a pure helper `vlmGateShouldSkip(ocrText, lang string) bool` and unit-testing that directly instead of injecting globals + `resolveTenantModelByType`.

Note: the CodeRabbit actionable comment about removing BMP/TIFF decoder imports from the test is stale — the final diff does not import `golang.org/x/image/bmp`/`tiff` (only `image/gif`, `image/jpeg`, `image/png` for fixture generation), so it is already resolved.

xugangqiang:
LGTM

xugangqiang:
LGTM

xugangqiang:
### Review findings (problems only)

**1. PR description overstates the change — please correct it.**
The description claims the PR "eliminated mutable package-level function pointers (`var runPaddleOCRImage`, `var runLocalImageOCR`)". This is not accurate:
- On `main`, `runPaddleOCRImage` and `runLocalImageOCR` are regular functions (`media_dispatch.go:450` and `:469`), not `var` function pointers.
- The diff does not remove any `var` declarations.
The pure-helper extraction (`vlmGateShouldSkip`) is real, but the "removed mutable package-level function pointers" wording should be dropped to avoid misleading reviewers.

**2. The CodeRabbit inline comment is a false positive — do not apply it as written.**
It states the test imports the BMP/TIFF decoders and therefore the subtests don't depend on `media_dispatch.go`. In fact the test imports only `image/gif`, `image/jpeg`, `image/png` (used to *encode* those fixtures); it does **not** import `golang.org/x/image/bmp` or `golang.org/x/image/tiff`. So the BMP and TIFF subtests genuinely rely on `media_dispatch.go`'s blank imports and would fail if those were removed. The test does protect the new registrations.

There is, however, a real (minor) grain in that comment worth fixing: the PNG/JPEG/GIF subtests import `image/png/jpeg/gif` for encoding, which also registers those decoders, so those three subtests do **not** actually validate `media_dispatch.go`'s registration of them — only WebP/BMP/TIFF are validated through `media_dispatch.go`. To make the test truly validate all six formats via `media_dispatch.go`, switch PNG/JPEG/GIF to fixed base64 fixtures (like WebP/BMP/TIFF) and drop those three imports. Otherwise, narrow the test doc comment to "validates WebP/BMP/TIFF registrations via media_dispatch.go".

**3. Pre-existing Python/Go parity gap in the gate (not introduced by this PR, flagging for follow-up).**
Python `rag/app/picture.py:88` does `txt = txt.strip()` *before* the threshold check, but the Go path passes `ocrText` un-trimmed into `vlmGateShouldSkip`. `strings.Fields` already strips for the word count, but `utf8.RuneCountInString(ocrText)` counts leading/trailing whitespace too. Edge case: 32 CJK chars + 1 trailing space (33 runes) makes Go skip VLM while Python (post-strip, 32 chars) would use VLM. This divergence predates this PR; recommend trimming `ocrText` at the gate entry as a follow-up so the "mirrors Python" claim holds exactly.

avikumart:
Thanks for the review! All 3 findings have been addressed in the latest commit:

1. **PR Description**: Updated the PR description to remove the inaccurate mention of removing mutable package-level function pointers. The description now accurately reflects the actual changes (zero-alloc `utf8.RuneCountInString`, Python parity whitespace trimming, BMP/TIFF decoder registrations, and dedicated unit tests).
2. **Decoupled Image Decoder Tests**: Dropped the encoder imports (`image/gif`, `image/jpeg`, `image/png`, `image/color`) from `media_dispatch_test.go` and switched PNG, JPEG, and GIF to static base64 test fixtures (matching WebP, BMP, and TIFF). `media_dispatch_test.go` now imports zero image decoders or encoders, ensuring all 6 format subtests in `TestImageDecoders_RegisteredFormats` strictly validate the decoder registrations in `media_dispatch.go`.
3. **Python/Go Parity (Whitespace Trimming)**: Added `ocrText = strings.TrimSpace(ocrText)` at the entry of `vlmGateShouldSkip` to align with Python `rag/app/picture.py:chunk()` (`txt = txt.strip()`). Added test cases covering whitespace trimming and boundary conditions in `TestVLMGateShouldSkip`.

Issue #18822, opened 2026-08-26
[Feature Request]: Add Synthorai as a model provider (OpenAI- and Anthropic-compatible gateway)


cuihuan:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### Is your feature request related to a problem?

Not a bug, and nothing is broken — this is a setup-friction request, so please read it as low priority.

Synthorai (<https://synthorai.io>) is an OpenAI- and Anthropic-compatible LLM gateway. It already works in RAGFlow today through the generic **OpenAI-API-Compatible** provider, so users are not blocked. But that path means pasting a base URL and hand-typing every model name one at a time, with no icon and no curated catalog. A named factory entry — the same treatment `OpenRouter`, `CometAPI`, `302.AI`, `Jiekou.AI`, `n1n`, `Astraflow`, `FuturMix`, `TokenPony`, `GreenPT` and `New API` already have in `conf/llm_factories.json` — turns that into picking a provider and pasting one key.

I'm the maintainer of Synthorai, so please treat this as a self-nomination rather than a user report. I've tried to make it easy to say no.

### Describe the feature you'd like

Add **Synthorai** as a named model provider.

| | |
|---|---|
| Base URL (OpenAI SDKs) | `https://synthorai.io/v1` |
| OpenAI Chat Completions | `POST https://synthorai.io/v1/chat/completions` |
| OpenAI Responses | `POST https://synthorai.io/v1/responses` |
| Anthropic Messages | `POST https://synthorai.io/v1/messages` |
| Auth | `Authorization: Bearer <key>` (suggested env var: `SYNTHORAI_API_KEY`) |
| Catalog | 113 models across 11 upstream providers (Claude, GPT, Gemini, GLM-5.2, Kimi-K3, DeepSeek, Qwen, …) |
| Pricing | direct-provider prices, zero markup |
| Docs | <https://synthorai.io/docs/> |
| Model & price list | <https://synthorai.io/pricing/> |
[...]

PR #18830, opened 2026-08-27
feat: add Synthorai model provider

cuihuan:
### Summary

Adds Synthorai as a model provider, following **#17447 (GreenPT, merged 2026-07-28)** — the closest precedent I could find: a fixed-base-URL OpenAI-compatible gateway contributed from outside.

**Disclosure: I'm affiliated with Synthorai.** Vendor adding its own provider, not a user request, and I'm not claiming demand — nobody else has asked for this as far as I know.

Related: #18822, where I opened a feature request first. #17447 went straight to a PR with no preceding issue, so the code belongs here.

### Deliberately smaller than GreenPT

GreenPT touched 15 files because it ships embedding, rerank and speech models. **Synthorai serves chat only** — its public catalog has no embedding, rerank or ASR models at all, which I checked before writing this — so `embedding_model.py`, `rerank_model.py` and `sequence2txt_model.py` are untouched rather than carrying adapters for capabilities that don't exist.

The Go driver also overrides nothing beyond its own identity. GreenPT overrides `ListModels` and `TranscribeAudio` because it has quirks; this gateway speaks the OpenAI wire format without deviation, so the embedded `OpenAIAPICompatibleModel` covers every call.

### Four models, not the whole catalog

`llm_factories.json` and `conf/models/synthorai.json` declare four models. The catalog is larger, but `max_tokens` is read downstream as fact, and these four are the ones whose context limits have been published with supporting measurements. I'd rather under-declare than put a guessed number where the agent trusts it.

### Checks, and what I could not check

- `go build ./internal/entity/models/...` — clean. `gofmt` and `go vet` on that package — clean.
- `python3 -m py_compile rag/llm/chat_model.py rag/llm/model_meta.py` — clean.
- `conf/llm_factories.json` still parses; the entry was inserted **as text** rather than via a JSON round-trip, so the rest of the file keeps its exact formatting and the diff is pure addition.

**A full `go build ./...` does not complete on my machine** — `office_oxide.h` is missing and `syscall.SysProcAttr.Pdeathsig` doesn't exist on macOS. Both are environmental and unrelated to this change, but it means I have not built the whole tree, and I haven't run RAGFlow end to end against the provider either. I don't want to imply otherwise.

### One process note

I understand from reading `.github/workflows/tests.yml` that CI only runs once a maintainer applies the `ci` label and the PR isn't a draft — so this will sit at zero checks until someone does that. Flagging it so it doesn't look like the checks are broken. Happy to fix whatever CI reports once it runs.


JinHai-CN:
@cuihuan Thanks for the PR! Could you drop an email to haijin.chn@gmail.com? We'll notify you when new RAGFlow versions are released so you can test that the Synthorai integration continues to work properly. Appreciate it!

JinHai-CN:
Please also check the comments from [@coderabbitai](https://github.com/apps/coderabbitai)

cuihuan:
Thanks — both CodeRabbit findings were real, and the second one is specific to us rather than generic advice. Fixed and pushed.

**1. Pinned the endpoint** (`rag/llm/chat_model.py`)

`SynthoraiChat` no longer accepts a caller-supplied `base_url`. Tenant config reaches it via `model_config["api_base"]`, so a configured endpoint could have sent the Synthorai API key to another host. Synthorai is a hosted gateway on one known host — there's no legitimate reason to override it, so the constructor now ignores the argument and always uses `https://synthorai.io/v1`.

**2. Restricted discovery to chat models** (`rag/llm/model_meta.py`)

This one I'm glad it caught. Our `/v1/models` returns the **whole** catalog — 113 models across chat, image, audio, video, realtime and other — while the inherited `_format_model_list` infers the type from the id and falls back to `chat`. So `wan2.7-image` and `chirp-2` would have been offered as chat models and failed at the chat-completions endpoint.

`Synthorai._format_model_list` now filters remote entries against the ids declared in `conf/models/synthorai.json`. Verified the behaviour directly: given `["claude-opus-5", "wan2.7-image", "chirp-2", "glm-5.2"]` it keeps only the two chat models.

`py_compile` is clean on both files.

**One observation I'd rather share than sit on:** the same `base_url or "<default>"` pattern exists in `GreenPTChat` (merged in #17447) and, as far as I could see, in several other adapters. The redirect risk is the same there. I deliberately did **not** touch them in this PR — that's your call and not mine to bundle into a provider addition — but it seemed worth flagging rather than quietly fixing only my own.

@JinHai-CN — will do on the email, thanks for offering. Happy to be on the list for release testing.

cuihuan:
CodeRabbit has re-reviewed the two fixes and is now green — **Review completed**, no new findings.

One thing left that I can't do from outside: the three CI jobs are still showing `skipping`. Reading `.github/workflows/tests.yml`, `ragflow_preflight` is gated on the `ci` label, and the PR currently carries only `💞 feature` and `size:L`.

@JinHai-CN — could you add the `ci` label when you get a moment? I'd rather the suite actually ran before anyone considers merging this, particularly since I couldn't complete a full `go build ./...` locally (missing `office_oxide.h`, and `syscall.SysProcAttr.Pdeathsig` doesn't exist on macOS — both environmental, but it means the tree isn't verified on my side).

If anything reds, I'll fix it against the actual output.

JinHai-CN:
```
╭──────────────────────────────────────────╮
│ 🥊 lefthook  v2.1.10   hook:  pre-commit │
╰──────────────────────────────────────────╯
┃  web-checks ❯ web-deps ❯ 

┃  web-checks ❯ web-oxfmt ❯ 

┃  gofmt ❯ 

┃  web-checks ❯ web-oxlint ❯ 

┃  ruff ❯ 
All checks passed!

┃  ruff-format ❯ 
Would reformat: rag/llm/model_meta.py
1 file would be reformatted, 1 file already formatted

exit status 1
```

cuihuan:
Fixed — thanks for running the hooks and pasting the output, that made it a one-minute fix.

`ruff-format` was right: the set comprehension I added in `_allowed_model_names` was wrapped across four lines and ruff wants it on one. Pushed in `2cb38d1`; nothing else in the file changed.

Verified with ruff 0.16.4 against your config:

- `ruff format --check` on both files I touched (`model_meta.py`, `chat_model.py`) — **already formatted**
- `ruff check` — 111 errors, but that's **identical before and after my changes** (I stashed and re-ran to confirm), and none of them land on the lines I added. So those are pre-existing and I've left them alone rather than mixing an unrelated cleanup into this PR.

On the email — will do, thanks for the offer.

PR #18832, opened 2026-08-27
fix(web): expose use_kg toggle in Retrieval canvas settings

### LINKED ISSUE #18778: [Bug]: Agent canvas "Retrieval" component has no UI option to enable use_kg (knowledge graph retrieval) in v0.27.0
### Self Checks

- [x] I have not modified the template and have filled in all the required fields.

v0.27.0 release image (`infiniflow/ragflow:v0.27.0`)

### RAGFlow image version

`infiniflow/ragflow:v0.27.0`

- MySQL 8.0.39, Elasticsearch, standard docker compose deployment (Linux VM)
- LLM backend: local llama.cpp server via OpenAI-API-Compatible provider

### Describe your problem

As a user, the intended flow on v0.27.0 is: **(1)** build an agent canvas with a Retrieval component, **(2)** publish it, **(3)** have the workflow's retrieval use the knowledge graph. Steps 1–2 work; step 3 is impossible, because the Retrieval component supports knowledge-graph retrieval on the backend but there is no way to enable it from the UI:

- **Backend**: `agent/tools/retrieval.py` — `RetrievalParam.use_kg` (default `False`). When true, the component calls `settings.kg_retriever.retrieval(...)` (KGSearch) and inserts the returned graph chunk at position 0 of the retrieved chunks, so it is clearly designed to be usable from a canvas.
- **Frontend**: the Retrieval node's settings form does not render any toggle for `use_kg`. Grepping the shipped v0.27.0 dist bundles, `use_kg` only appears in three places, none of which is the canvas Retrieval form:
  - the chat-assistant `prompt_config` default values (create-assistant form model),
  - the new search app `search_config` zod schema,
  - the built-in API documentation strings (the `/api/v1/retrieval` body parameter).

So an agent canvas cannot enable knowledge-graph retrieval at all in v0.27.0 — the parameter silently defaults to `false` and there is no documented way to change it from the UI.

Related context (already tracked elsewhere, not duplicating here): GraphRAG/RAPTOR configuration was intentionally removed from the dataset settings UI and replaced by the Knowledge Compilation Engine (#18760, #18683). As a side effect, in v0.27.0 there is also no UI entry that triggers a classic `graphrag` task: `POST /api/v1/datasets/{id}/index?type=graph` maps `"graph"` to `structure_graph` (compilation layer), and documents parsed through the default built-in pipeline run as `dataflow` tasks, bypassing the standard-chunking tail hook that used to auto-queue graphrag/raptor. We verified the backend engine itself still works by invoking `queue_raptor_o_graphrag_tasks(..., ty="graphrag")` directly — the graph builds fine and `GET /api/v1/datasets/{id}/graph` returns nodes/edges — but nothing in the UI can produce it, and a canvas Retrieval component cannot consume it either.

### Steps to reproduce

1. Create a dataset, upload a document, parse it.
2. Build a knowledge graph for the dataset (in v0.27.0 this itself has no UI entry — see context above).
3. Open an agent canvas, add a **Retrieval** component, open its settings.
4. Observe the parameter list: similarity threshold, keyword similarity weight, top N / top K, rerank model, empty response, cross-language search, metadata filter are all present — there is no "use knowledge graph" (`use_kg`) switch.

### Expected behavior

The Retrieval component settings form should expose the `use_kg` toggle (like the chat assistant configuration does), or the documentation should describe a supported way to enable it (e.g. DSL-level configuration).

### Actual behavior

No UI control and no documented way to enable `use_kg` on a canvas Retrieval component. Knowledge-graph data therefore cannot be used by agent workflows in v0.27.0, even though the backend path is fully implemented.

RahilOp:
### Problem
The Retrieval component already supports `use_kg` in `agent/tools/retrieval.py`, but the agent canvas settings form did not expose the switch, so knowledge-graph retrieval could not be enabled for a canvas Retrieval node (reported in #18778).

### Changes
- Add `use_kg` to the Retrieval form schema and default values.
- Add a shared `UseKnowledgeGraphFormField` switch under Advanced Settings.
- Render it for both the agent Retrieval node form and the tool Retrieval form.
- Hide the field when the retrieval source is Memory, since KG retrieval applies only to datasets.

### Verification
- `pnpm run type-check` reports no new errors in the touched files (the repository has pre-existing type errors in unrelated files).
- The existing backend `RetrievalParam.use_kg` path consumes the value once it is persisted in the DSL.

Closes #18778

### DIFF
--- a/web/src/pages/agent/constant/index.tsx
+++ b/web/src/pages/agent/constant/index.tsx
@@ -94,6 +94,7 @@ export const initialRetrievalValues = {
   ...initialSimilarityThresholdValue,
   ...initialKeywordsSimilarityWeightValue,
   cross_languages: [],
+  use_kg: false,
   retrieval_from: RetrievalFrom.Dataset,
   outputs: {
     formalized_content: {
--- a/web/src/pages/agent/form/retrieval-form/next.tsx
+++ b/web/src/pages/agent/form/retrieval-form/next.tsx
@@ -24,6 +24,7 @@ import {
   FormMessage,
 } from '@/components/ui/form';
 import { Radio } from '@/components/ui/radio';
+import { Switch } from '@/components/ui/switch';
 import { Textarea } from '@/components/ui/textarea';
 import {
   useRevalidateStaleDatasetIds,
@@ -58,6 +59,7 @@ export const RetrievalPartialSchema = {
   dataset_ids: z.array(z.string()),
   rerank_id: z.string(),
   empty_response: z.string(),
+  use_kg: z.boolean().default(false),
   cross_languages: z.array(z.string()),
   ...MetadataFilterSchema,
   memory_ids: z.array(z.string()).optional(),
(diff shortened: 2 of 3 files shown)


cike8899:
Knowledge graphs have been replaced by knowledge compilation; future versions will no longer use knowledge graphs. @RahilOp

RahilOp:
Thanks for the heads-up, @cike8899 — I wasn't aware that knowledge graphs are being phased out in favor of knowledge compilation. Since this PR only exposes the legacy  toggle, it doesn't make sense to merge it if the feature is going away. I'm closing it now.

PR #18836, opened 2026-08-27
fix(web): cap description length at 512 characters

dcc123456:
### Summary

fix(web): cap description length at 512 characters

Dataset settings and memory settings accepted arbitrarily long descriptions, which the backend rejects. Add a shared DESCRIPTION_MAX_LENGTH constant, enforce it in the zod schemas with a localized message, and set maxLength on the inputs so typing stops at the limit instead of failing on submit.

### DIFF
--- a/web/src/constants/common.ts
+++ b/web/src/constants/common.ts
@@ -236,3 +236,5 @@ export enum ThemeEnum {
   Light = 'light',
   System = 'system',
 }
+
+export const DESCRIPTION_MAX_LENGTH = 512;
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -24,6 +24,7 @@ export default {
       namePlaceholder: 'Please input name',
       nameSlashError: 'Name cannot contain "/"',
       descriptionPlaceholder: 'Enter description',
+      descriptionMaxLength: 'Description must be at most {{max}} characters.',
       next: 'Next',
       create: 'Create',
       edit: 'Edit',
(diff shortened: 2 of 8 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for centralizing the limit into a shared `DESCRIPTION_MAX_LENGTH` and enforcing it in both the zod schema and the input `maxLength`. A few concerns:

**Please verify the backend limit — 512 appears incorrect.**

The premise that "the backend rejects descriptions > 512 characters" does not match what I can find in the codebase:

- Dataset (knowledgebase) description: Python `api/utils/validation_utils.py` (`CreateDatasetReq.description`) and Go `internal/service/dataset/update.go` both allow up to **65535** characters; the entity column is `longtext`.
- Memory description: `internal/entity/memory.go` uses `longtext`, and `internal/service/memory.go` applies no length check.

Capping at 512 would therefore reject valid descriptions (513–65535 chars) that the backend actually accepts. Please either align `DESCRIPTION_MAX_LENGTH` to the real backend limit (65535), or point me at the validation that enforces 512 so frontend and backend stay consistent.

**Minor:**

- `t('common.descriptionMaxLength', {max})` in the zod schemas is evaluated once at module load, so the error message won't react to a runtime language switch. This matches existing patterns (e.g. `t('setting.nameRequired')`), so not a blocker, but worth being aware of.
- The diff also reformats several unrelated `azureDevOps*` tip strings in `en.ts`/`zh.ts` (prettier line-wrapping). Harmless, but it adds diff noise — consider splitting into a separate commit.
- Watch unit semantics: zod `.max()` counts UTF-16 code units, while the Go backend `len()` counts bytes. Moot at 512, but if you raise the limit toward 65535, CJK descriptions could pass the frontend and then be rejected by the byte-based Go check.

This is an AI review comment.

PR #18838, opened 2026-08-27
Refine entrypoint.sh - tidy code

wangq8:


wangq8 (commented):
This is an AI review comment.

Thanks for the cleanup — extracting the restart loop into `run_with_restart` is a nice readability improvement and removes a lot of duplication. However, there is one significant behavior change that I'd like to flag before this can be safely merged.

## Blocking concern: `ensure_docling` / `ensure_db_init` are now gated behind `ENABLE_WEBSERVER`

Before this change, both ran unconditionally:

```bash
ensure_docling
ensure_db_init

if [[ "${INIT_MODEL_PROVIDER_TABLES}" -eq 1 ]]; then ... fi
if [[ "${ENABLE_ADMIN_SERVER}" -eq 1 ]]; then ... fi
if [[ "${ENABLE_WEBSERVER}" -eq 1 ]]; then ... fi
```

After this change they only run when `ENABLE_WEBSERVER -eq 1`:

```bash
if [[ "${ENABLE_WEBSERVER}" -eq 1 ]]; then
    ensure_docling
    ensure_db_init
    echo "Starting nginx..."
    ...
fi
```

This is a functional regression for worker-only deployments. The script explicitly supports `--disable-webserver` (see the usage examples like `--disable-webserver --workers=2 --host-id=myhost123`), and document parsing happens in the **task executor**, not the web server. On a worker-only node with `USE_DOCLING=true`, `ensure_docling` would no longer run, so `docling` would never be installed at runtime and parsing would fail. Similarly, `ensure_db_init` (which calls `init_database_tables`) would be skipped for any node that doesn't run the web server.

Please either:
- keep `ensure_docling` unconditional (or gate it behind `ENABLE_TASKEXECUTOR` instead of `ENABLE_WEBSERVER`), and
- clarify whether `ensure_db_init` should only run on the web node (if that's intentional to avoid races, please document it in the PR description, since it changes `--disable-webserver` semantics).

## Minor issues

1. **Duplicate "Attempt to start" log line.** You echo `Attempt to start Admin python server...` before calling `run_with_restart`, and `run_with_restart` also echoes `Attempt to start ${process_name}...` at the top of its loop. The first message is redundant — consider removing the pre-call `echo`.

2. **Loss of clean-vs-error distinction in restart logs.** The old loop distinguished `exited cleanly` (status 0) from `exited with status $status`. The new function always prints `exited with code ${exit_code}`. That's a small loss of diagnostic signal; consider preserving the distinction inside `run_with_restart`.

3. **Naming inconsistency.** The ingestor messages changed from `go ingestor` to lowercase `ingestor`, while the other processes use title case (e.g. `RAGFlow go server`). Keeping the naming consistent (and retaining the `go` qualifier for the Go-based ingestor) would make logs clearer.

This is an AI review comment.

PR #18839, opened 2026-08-27
fix(dataset): canonicalize naive parser IDs to general

6ba3i:
Canonicalize legacy naive parser IDs to general for dataset creation, updates, and API responses.
Keep naive accepted and executable for backward compatibility.


wangq8 (commented):
Thanks for this cleanup. The canonicalization correctly reuses the existing `naive -> general` registry alias, and the create/update paths look sound. A few things to address before merge:

1. **Existing tests will break.** `datasetToMap`/`datasetListItemToMap` now canonicalize `parser_id`, so `TestUpdateDataset_StripsUnknownParam_Builtin` and `TestUpdateDataset_AcceptsValidComponentParams_Builtin` (`internal/service/dataset/update_test.go:1006` and `:1030`) — which seed a KB with `ParserID: ParserTypeNaive` and assert `result["parser_id"] == "naive"` — will now receive `"general"` and fail. Please update those assertions and add coverage for the new `datasetParserIDForResponse` canonicalization.

2. **No new tests** for the create/update canonicalization itself (e.g. creating/updating a dataset with `parser_id="naive"` should persist/return `"general"`). `helpers_test.go` only covers `validateParserID`.

3. **Related (likely pre-existing, worth reconciling now):** document-level `chunk_method` validation via `validDocumentChunkMethods` (`internal/service/document/document_dataset_update.go`) still does not include `"general"`, while the Go frontend parser catalog (`web/src/hooks/use-user-setting-request.tsx`) already maps the naive slot to `"general"`. As `general` becomes canonical, a document `chunk_method="general"` would be rejected. Please confirm this path is handled.

4. **Minor:** the gorm `default:` tag change (`naive` -> `general`) won't propagate to existing databases without a migration; it's safe here since the service always sets `parser_id` explicitly, but worth confirming no code path relies on the DB default.

wangq8:
@6ba3i please fix the unit tests.

PR #18840, opened 2026-08-27
Fix: On the MCP list page, if the only record on the last page is deleted, the view should navigate to the previous page.

cike8899:
### Summary

Fix: On the MCP list page, if the only record on the last page is deleted, the view should navigate to the previous page.


wangq8 (commented):
LGTM. The fix is correct and consistent with the existing `useGoToPreviousPageOnEmpty` pattern already used in `next-chats`, `datasets`, `agents`, `memories`, etc.

- `useListMcpServer` already exposes `loading` (from `isFetching`) and `data.mcp_servers`, so the destructuring is valid.
- `useGoToPreviousPageOnEmpty` correctly navigates back only when the list is empty, not loading, and the current page is > 1.

No issues found.

PR #18845, opened 2026-08-27
Fix Go Agents tab ordering to match Python

6ba3i:
The Go Agents tab now sorts merged agents and template groups by update time.

### DIFF
--- a/internal/service/agent.go
+++ b/internal/service/agent.go
@@ -732,9 +732,9 @@ func (s *AgentService) listAgentsGroupsOnly(ctx context.Context, userID, keyword
 }
 
 // mergeAgentsAndGroups combines agents and the caller's compilation template
-// groups into a single list ordered by (canvas_category, name) ascending, then
-// pages in Go. desc is accepted for API signature parity but the ordering is
-// intentionally category/name based, not chronological.
+// groups into a single list ordered by update_time, then pages in Go. This
+// mirrors Python's merged /agents response. A stable sort retains the original
+// agent-before-group order when timestamps are equal.
 func (s *AgentService) mergeAgentsAndGroups(ctx context.Context, userID string, agentItems []*AgentItem, keywords, orderBy string, desc bool, page, pageSize int) (*ListAgentsResponse, common.ErrorCode, error) {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the change — sorting the merged `/agents` list by `update_time` matches Python's merge path (`items.sort(key=lambda item: item.get("update_time") or 0, reverse=desc)`). `intValuePtr` correctly maps `nil` to `0`, mirroring Python's `or 0`, and `sort.SliceStable` preserves the original agent-before-group order on timestamp ties. The Go handler defaults (`orderby=create_time`, `desc=true`) also align with Python. I did not find functional regressions in the merge logic itself.

One thing to address before merge:

- **Stale test**: `TestListAgents_MergesCompilationTemplateGroups` still asserts the old `(canvas_category, name)` ordering — its `orderOK` check and `t.Fatal` messages say "ordered by (canvas_category, name) ascending". With only two seeded items it now passes for the wrong reason and no longer verifies the actual ordering invariant. Please update it to assert `update_time` DESC ordering and add a third item (e.g. two agents with distinct `update_time`) so the interleaving and stable-sort behavior are actually covered.

Minor / optional:

- `orderBy` is now effectively ignored for the merged sort (always `update_time`). That is consistent with Python's merge behavior, but a one-line comment would help future readers understand it's intentional rather than a bug.

This is an AI review comment.

internal/service/agent.go
```diff
@@ -744,9 +744,8 @@ func (s *AgentService) mergeAgentsAndGroups(ctx context.Context, userID string,
 	for _, item := range agentItems {
 		item.Type = AgentItemTypeAgent
 		merged = append(merged, mergeCanvasItem{
-			item:     item,
-			category: item.CanvasCategory,
-			name:     derefString(item.Title),
+			item: item,
+			time: intValuePtr(item.UpdateTime),
```
wangq8, line 748: why delete category: item.CanvasCategory, are they not used?

6ba3i: it's still used, I only removed the redundant category field from the temporary mergeCanvasItem to match the Python implementation

wangq8: got it.

wangq8:
LGTM.

PR #18847, opened 2026-08-27
fix(web): keep typed slider values on the slider's step grid

dcc123456:
### Summary

fix(web): keep typed slider values on the slider's step grid

The number input beside a slider accepted any number of decimals, so a
typed value could land off the slider's grid and out of sync with the
thumb. Derive the allowed precision from the step and truncate extra
digits, writing the DOM value directly since the state may not change.


wangq8 (commented):
This is an AI review comment.

Solid fix — the string-based truncation in `limitDecimals` is the right call and correctly avoids the classic float pitfall (`0.29 * 100 === 28.999...`). The comment explaining why the DOM value is written directly (state may not change) is clear and justified.

A few minor nits, none blocking:

1. `precision` derivation: `String(step ?? '').split('.')[1]?.length ?? 0` assumes decimal-notation steps. A step written in exponential form (e.g. `1e-7`) stringifies to `"1e-7"`, which has no `.` and yields `precision = 0`. Current steps are `0.01`-style so this is unlikely in practice, just flagging the edge case.

2. `limitDecimals` returns `Number(...)`, which can re-normalize (e.g. drop a trailing zero). This is pre-existing behavior from parsing `Number(e.target.value)`, so not introduced here — just noting the direct `e.target.value = String(limitedValue)` write can jump the cursor when the value length changes. Given the stated rationale, that's an acceptable trade-off.

3. No test added. If there's no existing unit-test harness for these UI components that's understandable, but a tiny test for `limitDecimals` (e.g. `0.29` with `decimals=2`, `0.123` → `0.12`, `decimals=0`) would be cheap insurance.

Otherwise LGTM.

This is an AI review comment.

PR #18851, opened 2026-08-27
Fix: The models for audio and video parsers are required fields.

cike8899:
### Summary
Fix: The models for audio and video parsers are required fields.


wangq8 (commented):
This is an AI review comment.

The change is correct and low-risk: `required` is a valid prop on `ModelTreeSelectFormField`, and it's forwarded to `FormLabel`.

One clarification worth checking: `required` here only renders the visual `*` on the label. It does **not** add a `rules={{ required: ... }}` validation rule on the underlying `FormField`, so the field is not actually enforced as required on submit.

- If the goal is only to visually indicate that a model must be chosen, this is fine as-is.
- If the intent (per the PR title, "required fields") is to actually enforce the requirement, please also add a validation rule so an empty audio/video parser model is blocked on submit.

This is an AI review comment.

PR #18853, opened 2026-08-27
feat: locate Excel citation sources in document preview

huang-aoqin:
## Summary
- Excel citations jump to the source sheet/row/column in the in-app document preview (chunk view, chat drawer, and search).
- Parsers emit real Excel coordinates (sheet, rowStart, rowEnd, colStart, colEnd) instead of dummy [i]*5 positions.
- Chat citation clicks for .xlsx / .xls open the shared preview drawer instead of a new tab.

**Chunking change:** the default Excel path (`html4excel` off) now merges consecutive rows in the same sheet up to `chunk_token_num` (default 128) so each chunk has a real cell range. Previously this path produced **one chunk per row**. Re-parsing existing `.xlsx`/`.xls` documents will change chunk counts, boundaries, and retrieval. `html4excel` is unchanged (`chunk_token_num=0` keeps HTML table chunks as-is). Old chunks still carry dummy positions until re-parsed.

## Test plan
- [ ] Parse an xlsx with default row chunking: click chunks in the parse-result view and confirm the preview highlights the matching rows
- [ ] Parse an xlsx with html4excel and more than 12 data rows: the second chunk highlights from row 14, not from row 1
- [ ] In chat, click an xlsx citation: the side drawer opens and locates the source range
- [ ] Multi-sheet workbook: a chunk from sheet 2 switches to that sheet
- [ ] PDF citations still use the existing highlighter
- [ ] Re-parse existing Excel documents; old chunks still carry dummy positions

image


xugangqiang:
Please fix ci error

xugangqiang:
### Review: problems found (read-only review)

I reviewed the diff and traced the positions pipeline across the Python, Go, and web layers. The main ingestion paths (Python `naive.py` and Go `xlsx_parser.go`) are internally consistent and will deliver correct 1-based sheet coordinates to the frontend. A few issues remain before merge:

**1. Hard-coded hex colors violate web style guidelines (should fix)**
`web/src/components/document-preview/hooks.ts` (`applyExcelSourceLocate`) uses `bgcolor: '#ffe58f'` / `color: '#1a1a1a'`. The project guideline requires using color tokens from `tailwind.css`, not arbitrary hex. Replace with token-backed values so the highlight is valid in both themes.

**2. `rag/flow/parser/parser.py` sheet index — verify off-by-one (must verify)**
The JSON path stores `positions: [[sheet, r1, r2, c1, c2]]` with `sheet` 0-based, relying on `add_positions` to increment. `rag/svr/task_executor.py:915` does call `add_positions` on `"positions"`, which matches the `naive.py` convention (0-based in → 1-based out), so this may be correct. But it is unverified against the actual flow indexing path — if the flow does NOT run through `add_positions`, the stored sheet would be 0-based and citations from sheet 2 would select sheet 1. Add a two-sheet regression test through the flow `output_format=json` path and confirm the cite selects the correct sheet. Also fix the misleading comment ("add_positions increments") accordingly. Do NOT blindly add `+1`, since TaskExecutor would then double-count.

**3. Docstring coverage pre-merge gate failing**
The pre-merge check reports 42.42% docstring coverage (< 80%) across 33 functions. New functions (`tokenize_chunks_with_positions`, `_merge_excel_items`, `applyExcelSourceLocate`, `htmlTableChunk`, `supportsSourceLocate`, etc.) lack docstrings. CI will block merge until this is addressed.

**4. `rag/app/one.py` silently drops positions**
`sections = [s[0] if isinstance(s, tuple) else s for s in sections]` discards the new position tuples for the `one.py` Excel path, so `one.py`-ingested Excel loses source-locate support. This is an inconsistency with the main path — at minimum add a comment explaining the intentional drop.

**5. Untested new Python ingestion logic**
`_merge_excel_items` and `tokenize_chunks_with_positions` in `naive.py` / `rag/nlp/__init__.py` have no unit tests, and the web `applyExcelSourceLocate` (which depends on undocumented `@js-preview/excel` internals: `xs.datas`, `bottombar.clickSwap, sheet.resetData`) has no test. The Go rendering side is tested; the Python ingestion side and the multi-sheet flow path are not.

**6. Optional: add debug logging for the new position-aware flow**
Per the "log new flows" guideline, a debug log in `tokenize_chunks_with_positions` (position + whether child-delimiter split or direct tokenize, without logging chunk content) would help diagnose incorrect citation targets.

**Non-issue checks I confirmed:** the `(text, position)` tuple change to `ExcelParser.html()`/`__call__()` has exactly 3 callers, all updated; `num_tokens_from_string`, `normalize_arabic_presentation_forms`, and `tokenize_chunks_with_positions` are already imported in `naive.py`; switching `pdf-drawer` from `PdfPreview` to `DocumentPreview` preserves PDF rendering (same default export); the API maps `position_int` → `positions` so the web receives real coordinates.

Recommendation: fix #1 and #3 (blockers), verify #2 with a test, and optionally address #4–#6 before merge.

huang-aoqin:
> Please fix ci error

@xugangqiang done!

xugangqiang:
Review findings (read-only) — discovered issues to address before merge:

**1. Add a flow two-sheet `output_format=json` end-to-end regression test.**
The flow path (`rag/flow/parser/parser.py:_spreadsheet`) stores a **0-based** sheet index and relies on `add_positions` in `rag/svr/task_executor.py:916` / `dataflow_service.py:341` to do `pn+1`. The current `test_call_emits_zero_based_sheet_index` only checks the *parser* output, not the downstream conversion. If the `_spreadsheet` JSON output does not actually flow through `add_positions`, a citation from sheet 2 would select sheet 1. Please add a test through the `output_format=json` path (two sheets) that asserts the stored `position_int[0][0] == 2`. (`test/unit_test/rag/flow/...` or an integration test.)

**4. Harden `applyExcelSourceLocate` (`web/src/components/document-preview/hooks.ts`).**
- `xs.bottombar.clickSwap2(tab)` is called without an existence guard. If the `@js-preview/excel` version lacks `clickSwap2`, this throws. Guard with `typeof xs.bottombar.clickSwap2 === 'function'` and fall back to `xs.sheet.resetData(data)` as already done in the else branch.
- The highlight style reads `getPropertyValue('--background-highlight')` / `'--text-title')`. If those theme tokens are absent, the value is an empty string and the highlight silently disappears. Provide a sensible fallback color when the token is empty.

**5. Document the Excel default chunk-granularity change in the PR description.**
`_merge_excel_items` (`rag/app/naive.py`) now merges Excel rows into ~128-token chunks, whereas the previous default path produced one chunk per row. This is an unstated side effect: re-parsing existing `.xlsx`/`.xls` documents will change chunk counts and boundaries (and thus retrieval). `html4excel` is unaffected because `chunk_token_num=0` takes the passthrough path. Please call this out explicitly in the PR description so reviewers/users are aware.

(For context: the pipeline coordinate chain — Python naive / Python flow / Go — is internally consistent and correct; items 1/4/5 above are the blocking gaps.)

huang-aoqin:
> Review findings (read-only) — discovered issues to address before merge:
> 
> **1. Add a flow two-sheet `output_format=json` end-to-end regression test.** The flow path (`rag/flow/parser/parser.py:_spreadsheet`) stores a **0-based** sheet index and relies on `add_positions` in `rag/svr/task_executor.py:916` / `dataflow_service.py:341` to do `pn+1`. The current `test_call_emits_zero_based_sheet_index` only checks the _parser_ output, not the downstream conversion. If the `_spreadsheet` JSON output does not actually flow through `add_positions`, a citation from sheet 2 would select sheet 1. Please add a test through the `output_format=json` path (two sheets) that asserts the stored `position_int[0][0] == 2`. (`test/unit_test/rag/flow/...` or an integration test.)
> 
> **4. Harden `applyExcelSourceLocate` (`web/src/components/document-preview/hooks.ts`).**
> 
> * `xs.bottombar.clickSwap2(tab)` is called without an existence guard. If the `@js-preview/excel` version lacks `clickSwap2`, this throws. Guard with `typeof xs.bottombar.clickSwap2 === 'function'` and fall back to `xs.sheet.resetData(data)` as already done in the else branch.
> * The highlight style reads `getPropertyValue('--background-highlight')` / `'--text-title')`. If those theme tokens are absent, the value is an empty string and the highlight silently disappears. Provide a sensible fallback color when the token is empty.
> 
> **5. Document the Excel default chunk-granularity change in the PR description.** `_merge_excel_items` (`rag/app/naive.py`) now merges Excel rows into ~128-token chunks, whereas the previous default path produced one chunk per row. This is an unstated side effect: re-parsing existing `.xlsx`/`.xls` documents will change chunk counts and boundaries (and thus retrieval). `html4excel` is unaffected because `chunk_token_num=0` takes the passthrough path. Please call this out explicitly in the PR description so reviewers/users are aware.
> 
> (For context: the pipeline coordinate chain — Python naive / Python flow / Go — is internally consistent and correct; items 1/4/5 above are the blocking gaps.)

@xugangqiang done!

xugangqiang:
This looks like a good feature. @buua436  @yuzhichang  pls have a second eye. cc @JinHai-CN

PR #18854, opened 2026-08-27
Fix: remove deprecated mimo-v2-flash

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/conf/llm_factories.json
+++ b/conf/llm_factories.json
@@ -7691,13 +7691,6 @@
                     "max_tokens": 1048576,
                     "model_type": "chat",
                     "is_tools": true
-                },
-                {
-                    "llm_name": "mimo-v2-flash",
-                    "tags": "LLM,CHAT,256K",
-                    "max_tokens": 262144,
-                    "model_type": "chat",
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the cleanup. The JSON edit is syntactically clean — after removing the `mimo-v2-flash` entry, the remaining object still ends correctly with no trailing-comma problem.

A couple of points to confirm before merging:

1. **Existing tenants**: removing a model from `conf/llm_factories.json` can leave tenants that already selected `mimo-v2-flash` pointing at a model id that no longer exists in the factory. Can you confirm RAGFlow tolerates this gracefully at runtime (or whether any data cleanup/migration is needed)? A short note in the PR body would help reviewers.

2. **Deprecation source**: the title says "deprecated", but the sibling `mimo-v2.5` / `mimo-v2.5-pro` entries are kept. Could you point to the upstream (Xiaomi) notice or reason this specific model was removed? It makes the change easier to verify.

Otherwise LGTM — no tests needed for a config-only change.

This is an AI review comment.

PR #18855, opened 2026-08-27
fix: handle Jina VLM streaming responses

buua436:

Jina VLM streaming responses differ from standard OpenAI SSE responses, causing reasoning content to be mixed with the final answer or resulting in empty Canvas Agent output.

This PR adds Jina-specific stream handling for both SSE and newline-delimited JSON responses, separates reasoning from answer content, supports clean EOF termination, and forwards upstream API errors through the standard RAGFlow stream.

It also handles the message requirements of the `api.jina.ai` endpoint while preserving the original message structure for other Jina endpoints.

### Type of change

### DIFF
--- a/internal/entity/models/jina.go
+++ b/internal/entity/models/jina.go
@@ -17,15 +17,19 @@
 package models
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"ragflow/internal/common"
 	"strings"
 	"time"
+
+	"go.uber.org/zap"
 )
 
 type JinaModel struct {
@@ -106,6 +110,7 @@ func (j *JinaModel) ChatWithMessages(ctx context.Context, modelName string, mess
 		return nil, err
 	}
 	baseURL = strings.TrimSuffix(baseURL, "/")
+	messages = prepareJinaMessages(baseURL, messages)
 	url := fmt.Sprintf("%s/%s", baseURL, j.baseModel.URLSuffix.Chat)
 
 	reqBody := buildRequestBody(chatModelConfig, modelName, messages, false)
@@ -125,6 +130,9 @@ func (j *JinaModel) ChatWithMessages(ctx context.Context, modelName string, mess
 }
 
 func (j *JinaModel) ChatStreamlyWithSender(ctx context.Context, modelName string, messages []Message, apiConfig *APIConfig, chatModelConfig *ChatConfig, modelUsage *common.ModelUsage, sender func(*string, *string) error) error {
+	if sender == nil {
+		return fmt.Errorf("sender is required")
+	}
 	if err := j.baseModel.APIConfigCheck(apiConfig); err != nil {
 		return err
 	}
@@ -143,10 +151,10 @@ func (j *JinaModel) ChatStreamlyWithSender(ctx context.Context, modelName string
 		return err
 	}
 	baseURL = strings.TrimSuffix(baseURL, "/")
+	messages = prepareJinaMessages(baseURL, messages)
 	url := fmt.Sprintf("%s/%s", baseURL, j.baseModel.URLSuffix.Chat)
 
 	reqBody := buildRequestBody(chatModelConfig, modelName, messages, true)
-	reqBody["stream_options"] = map[string]interface{}{"include_usage": true}
 
 	if chatModelConfig != nil {
 		chatModelConfig.ToolCallsResult = nil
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the thorough Jina streaming handling — the thinking/answer split, NDJSON+SSE parsing, EOF tolerance, and error forwarding are all well covered by tests. A few things to check:

1. **`flushThinking(false)` can leak a partial `</think>` marker.** When the buffered content is ≤ `len("</think>")-1` (7 chars) and contains no complete `</think>`, the whole buffer is emitted as reasoning and reset. If `</think>` is split across chunks and the first part lands when the buffer is otherwise empty (e.g. `</thi` then `nk>`), the literal `</think>` text leaks into the reasoning channel. The keep-last-7-chars guard only applies when the buffer is >7 chars. Suggest always retaining the trailing up-to-7 chars on the non-final flush (emit only `value[:max(0, len(value)-7)]`) and letting the final flush emit the rest.

2. **Error swallowing.** `ChatStreamlyWithSender` now returns `nil` after forwarding `**ERROR**` + `[DONE]`, so callers that rely on the returned error will not observe the failure. This is a deliberate, consistent pattern, but please confirm every downstream caller scans for the `**ERROR**` marker.

3. **`mergeLeadingJinaSystemMessages` edge cases.** When the first non-system message is not a `user` (e.g. `assistant`/`tool`), a synthetic `user` message is injected to carry the system text; when the message list is all-system, it collapses to a single `user` message. Please confirm these match `api.jina.ai`'s requirements for the affected flows.

4. **Minor:** in `parseJinaStream`, non-JSON lines containing `:` that don't start with `{`/`[` are silently skipped (including plain-text error bodies). That's probably acceptable, but be aware it can hide upstream error text.

Otherwise the implementation is solid and well-tested.

This is an AI review comment.

buua436 (commented):
A timed-out insert may still complete and update counters. Retrying the request then overwrites the same chunk but increments counters again. Please make this operation idempotent.

buua436:
f107595c-d186-431c-8f67-9365e93364e7

Lynn-Inf (commented):
🔴 **P0 — Reasoning/answer split is asymmetric and likely misclassifies answers**

  Opening thinking is driven by delta.type == "think" (jina.go:170), but closing thinking is driven only by detecting a literal </think> substring inside delta.content (jina.go:180-191). These two mechanisms don't match.

  Consequence: if Jina closes thinking by changing type (e.g. type:"answer" or omitting type) instead of emitting a literal </think> token, thinking never flips back to false, and all answer text gets routed into reasoning. That's exactly the
  "reasoning content mixed with the final answer / empty Canvas output" symptom the PR is trying to fix.

  The unit tests only exercise the </think>-token scenario, so they pass and don't catch this. The PR description itself says Jina "marks thinking chunks with delta.type="think"" — implying type is the delimiter, which the closing logic ignores.

  Suggested fix: make closing symmetric with opening. Close thinking when typ transitions away from "think" (e.g. a non-empty type that isn't "think"), while keeping </think>-tag support as a fallback:

  if typ != "" {
      if typ == "think" {
          if !thinkingClosed { thinking = true }
      } else { // e.g. "answer"
          thinking = false
          thinkingClosed = true
      }
  }
  Then the </think> substring handling only needs to run while thinking is true. This covers all three conventions (pure tag-based, pure type-based, mixed).

  🟠 **P1 — Removing stream_options likely disables usage reporting**

  The diff deletes reqBody["stream_options"] = {"include_usage": true} (jina.go:51 removed), yet handleJinaStreamingResponse still reads usage via OpenAIParserConfig.StreamParser(event) and calls recordResponseUsage / sets chatConfig.UsageResult
  (jina.go:198-242).

  In OpenAI-compatible APIs, usage is only returned when stream_options.include_usage=true. If Jina follows that convention, streamUsage will stay nil and the retained usage code becomes dead — breaking token accounting for Jina. The test even asserts
   stream_options must not be sent. Please confirm Jina returns usage unconditionally; if not, this is a silent regression. Otherwise a comment explaining why it's safe would help.

  🟡 **P2 — Client-abort errors are re-emitted as fake upstream errors**

  emit/flushThinking/consumeContent propagate sender errors all the way up; the outer block (jina.go:64-74) then treats any such error as an upstream failure and tries to send **ERROR**: … + [DONE] again. If sender failed because the client
  disconnected, the re-send fails too and you return a confusing wrapped error. Distinguish a client-side sender error from a genuine upstream error so you don't synthesize a bogus **ERROR** chunk after the client is already gone.

  🟡 **P3 — eventCount == 0 turns a clean empty stream into a client error**

  jina.go:232-234 returns "Jina stream contained no JSON events" whenever zero JSON events were parsed. A valid stream that is empty or closes immediately after [DONE] (with no content events) would then be forwarded to the client as **ERROR**.
  Consider erroring only when nothing was read at all, not when the stream was simply empty.

  🟡 **P3 — System-only prompts aren't merged for api.jina.ai**

  In mergeLeadingJinaSystemMessages (jina.go:99-101), when every message is system (firstNonSystem == 0), it returns the messages unchanged — still with role: system. The whole point of prepareJinaMessages is that api.jina.ai can't take system
  messages, so the degenerate "system-only" request would defeat the merge. Low severity (no user turn is unusual) but worth handling.

  ✅ **Things that are correct**

  - The HTTP-error forwarding test is valid: doStreamRequest (base_model.go:276-278) returns "API request failed with status %d: %s" including the body, so **ERROR**: API request failed with status 400: {…roles must alternate…} is what the client
  receives — matching the test.
  - parseJinaStream handles both SSE data: frames and bare NDJSON, and tolerates event:/id:/retry:/SSE-comment lines. The 7-char </think> look-behind in flushThinking is mathematically correct (retains the last 7 of an 8-char token).
  - prepareJinaMessages only rewrites messages for api.jina.ai (not deepsearch.jina.ai), preserving original behavior for other endpoints; multimodal user content is preserved (test covers it).
  - sender == nil guard added in both ChatStreamlyWithSender and handleJinaStreamingResponse.

  **Bottom line**

  The core bug the PR targets (reasoning/answer separation) is not actually fixed for the type-based wire format the description describes — P0 is the one to resolve before merge. P1 should be confirmed. The rest are robustness improvements.

PR #18856, opened 2026-08-27
Revert "fix(web): file list total count doesn't match displayed rows (#17747)"

dcc123456:
### Summary

This reverts commit ae66eb835ab015484f96d0875d533feb77d5ce44.

### DIFF
--- a/web/src/pages/files/files-table.tsx
+++ b/web/src/pages/files/files-table.tsx
@@ -107,8 +107,8 @@ export function FilesTable({
 
   // Sort files with skills folder first, then by time
   // Filter out the skills folder on the Python backend
-  const { sortedFiles, hiddenCount } = useMemo(() => {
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

This is a clean revert, but the PR body doesn't explain *why* #17747 needs to be reverted. Please document the regression it introduced.

Key concern: #17747 fixed a real mismatch — on the Python backend the skills folder is filtered client-side (`isSkillsEnabled` false) while the server `total` still counts it, so the displayed total exceeded the visible rows. Reverting to `rowCount: total` / `total={total}` will reintroduce that mismatch unless the server now excludes the skills folder from `total`.

Please confirm the server-side behavior has changed to make this revert safe, and ideally add a verification note or test. Without that, this looks like it trades one bug for the original one.

This is an AI review comment.

PR #18857, opened 2026-08-27
fix: reject wiki LLM error responses

buua436:

Wiki page generation treated LLM error responses such as `**ERROR**: ERROR_MAX_RETRIES` as valid content and persisted them as page bodies.

This change detects standard LLM error responses and raises an exception instead, allowing the page to be marked as failed and retried without polluting Wiki content.

### Type of change

### DIFF
--- a/rag/advanced_rag/knowlege_compile/wiki_incremental.py
+++ b/rag/advanced_rag/knowlege_compile/wiki_incremental.py
@@ -161,7 +161,10 @@ async def _chat_mdl_ask(chat_mdl, system_prompt: str, user_prompt: str, temperat
         raise
     if isinstance(raw, tuple):
         raw = raw[0]
-    return _strip_think(raw or "")
+    response = _strip_think(raw or "")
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
This is an AI review comment.

Looks good. Detecting the standard `**ERROR**` prefix and raising before persisting is the right call, and the tests cover the important cases (plain error, mid-response error, tuple return, and non-persist-on-error).

Minor suggestions (non-blocking):

1. **Mid-line detection.** The check only catches `**ERROR**` at the start of a line (`line.lstrip().startswith(...)`). If an error marker could ever appear mid-line (e.g. wrapped text), it would be missed. For the known upstream format this is fine, but a `"**ERROR**" in response` check would be more robust.

2. **Exception payload size.** `RuntimeError(f"Wiki LLM call failed: {response}")` embeds the full response in the exception. If responses can be long, consider truncating to a bounded prefix to keep logs/traces tidy.

Otherwise LGTM.

This is an AI review comment.

wangq8:
LGTM.

PR #18858, opened 2026-08-27
Fix: When the pipeline-paeser component switches the file type from PDF to image, the default parsing method should be OCR.

cike8899:
### Summary

Fix: When the pipeline-paeser component switches the file type from PDF to image, the default parsing method should be OCR.

### DIFF
--- a/web/src/pages/agent/form/parser-form/email-form-fields.tsx
+++ b/web/src/pages/agent/form/parser-form/email-form-fields.tsx
@@ -15,6 +15,7 @@ export function EmailFormFields({ prefix }: CommonProps) {
       <RAGFlowFormItem
         name={buildFieldNameWithPrefix(`fields`, prefix)}
         label={t('flow.fields')}
+        required
       >
         {(field) => (
           <MultiSelect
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix. Setting `parse_method` on file-type switch via `getInitialParseMethod` is a reasonable approach and matches the existing `output_format` reset pattern.

A few notes:

1. **Empty-string fallback.** `getInitialParseMethod` returns `''` when no `initialParserValues.setups` entry matches the file type. If `setups.<index>.parse_method` must be non-empty (e.g. it feeds validation or a required select), an empty string could leave the form in an invalid state. Since the setups table should cover every `FileType`, this is likely unreachable, but consider a safe fallback or an explicit assertion.

2. **Unrelated change.** Adding `required` to the email `fields` MultiSelect (`email-form-fields.tsx`) is unrelated to the PDF→image OCR default. Please move it to its own PR so the change stays focused.

3. **Other dependent fields.** Switching file type now resets `parse_method` and `output_format`, but method-specific sub-options may also need resetting. Please confirm no stale method-specific fields remain after a type switch.

4. **Missing test.** `getInitialParseMethod` has no unit test. A small test covering PDF→image (expects OCR) and other types would lock in the behavior.

This is an AI review comment.

PR #18859, opened 2026-08-27
Fix: hide skills folder in REST API

wangq8:


wangq8 (commented):
This is an AI review comment.

Thanks for the fix. Overall the approach — detect the root folder via `id == parent_id`, then exclude the direct `skills` child — is correct and consistent across the Python and Go implementations. A few things to consider before merging:

**1. Keyword search still leaks skills content (moderate)**
`get_by_pf_id` / `GetByPfID` with a non-empty `keywords` searches the whole subtree under `pf_id` (`parent_id IN subtree_ids`). The new exclude condition only removes the `skills` folder itself as a *direct* child (`parent_id = pf_id AND name = 'skills'`). Files and sub-folders nested *inside* the skills folder will therefore still show up when listing the root with a search term. If the goal is to fully hide skills from the REST file list, this is incomplete; if only hiding the folder entry itself is intended, please note that in the PR description so the behavior is unambiguous.

**2. Missing test coverage for the new behavior**
The Go tests were only updated to pass `excludeSkills=false`; there is no test asserting that `excludeSkills=true` (or `exclude_skills=True` in Python) actually filters out the `skills` folder. Since `GetByPfID` is already unit-tested, adding a case (root listing with a "skills" child) would be cheap and would lock in the behavior.

**3. Duplicated constant (minor / style)**
`SkillsFolderName = "skills"` is now defined in both `internal/service/file/file.go` and `internal/dao/file.go`. This mirrors the existing `DatasetFolderName` duplication so it's not new, but it adds a second place that can drift. Consider passing the folder name as a parameter to `GetByPfID` (or referencing a single shared constant) instead of baking the business name into the DAO layer.

**4. Name-based matching can hide a user's own "skills" folder (minor)**
The filter hides *any* direct child of root named exactly `skills`, including a folder a user created for their own purposes. This is consistent with how the skills folder is identified elsewhere in the codebase (by name), so it's acceptable, but worth being aware of.

Everything else looks correct: all call sites are updated (Python single caller, Go `ListFiles` + the two skill services + tests), and the `total`/`count` is computed after the exclude filter so pagination stays accurate.

Lynn-Inf (approved):
LGTM

PR #18860, opened 2026-08-27
Fix: parse pdf in chat

Lynn-Inf:
### Summary

As title.


wangq8 (commented):
This is an AI review comment.

Thanks for the fix. The core change in `rag/app/naive.py` looks correct — normalizing `"Plain Text"` / `"plaintext"` / `"PlainText"` variants so `by_plaintext` routes to `PlainParser` instead of falling into the vision branch.

A few notes:

1. **Scope / formatting noise.** This PR bundles a one-line functional change with a large amount of pure formatting churn (black reformatting of `run_tests.py` and `test_naive_dispatch.py`, prettier reformatting of `fr/it/ja/ko/ru/tr` locale files and `use-save-graph.test.ts`, plus a trailing-newline-only change to `run_go_tests.sh`). The locale changes are unrelated to "parse pdf in chat" and make the diff hard to review and prone to merge conflicts. Please split the formatting into a separate PR.

2. **Duplicated normalization.** `_dispatch` already normalizes `"Plain Text"` → `"plaintext"` before the `PARSERS` lookup, while `by_plaintext` re-derives the layout recognizer from `kwargs`. Having two separate normalization sites can drift (which is exactly what this bug was). Consider centralizing the normalization so both paths share one source of truth.

3. **`.replace(" ", "")` strips all whitespace**, not just the space between "Plain" and "Text". This is harmless for the intended values, but a tighter check (e.g. comparing against a canonicalized `"plaintext"` after lowercasing and collapsing internal spaces) would be clearer.

4. **Missing test.** The existing tests cover the dispatcher, but there is no test asserting that `by_plaintext` with `layout_recognizer="plaintext"` (or `"PlainText"`) selects `PlainParser`. Please add a unit test for the actual branch being fixed.

This is an AI review comment.

wangq8:
LGTM.

PR #18865, opened 2026-08-27
fix(metadata-filter): propagate resolved document IDs to retrieval

6ba3i:
The chat pipeline was ignoring successful metadata-filter results because it checked the wrong boolean. It now applies any non-nil resolved document ID list.


wangq8:
LGTM.

PR #18866, opened 2026-08-27
fix(web): derive provider base_url visibility from the provider catalog

dcc123456:
### Summary

fix(web): derive provider base_url visibility from the provider catalog

The generic API-key form gated its `base_url` field on a hardcoded `FACTORIES_WITH_BASE_URL` list, so providers such as Jina could not be given a base URL even though `GET /providers` advertises one for them. Gate the field on the URL options already derived from that response instead, and drop the list.

### DIFF
--- a/web/src/pages/user-setting/setting-model/provider-schema/field-config/generic-api-key-config.ts
+++ b/web/src/pages/user-setting/setting-model/provider-schema/field-config/generic-api-key-config.ts
@@ -92,15 +92,3 @@ export const GenericApiKeyConfig: ProviderConfig = {
     };
   },
 };
-
-/**
- * List of factories supporting base_url (used for the generic ApiKey modal)
- */
-export const FACTORIES_WITH_BASE_URL = [
-  LLMFactory.OpenAI,
-  LLMFactory.AzureOpenAI,
-  LLMFactory.TongYiQianWen,
-  LLMFactory.MiniMax,
-  LLMFactory.SILICONFLOW,
-  LLMFactory.TencentHunYuan,
-];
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

The refactor is sound — deriving `showBaseUrl` from the provider catalog (`GET /providers`) instead of a hardcoded list is more maintainable and correctly surfaces `base_url` for providers that advertise a URL but weren't in the old list (DeepSeek, Moonshot, Mistral, NVIDIA, StepFun, Upstage, PPIO, etc.).

I checked for regressions against the removed `FACTORIES_WITH_BASE_URL` list and the Anthropic special-case, using `conf/llm_factories.json`:

- OpenAI, Tongyi-Qianwen, MiniMax, SiliconFlow, Tencent Hunyuan — all have non-empty `url`, so `baseUrlOptions` is non-empty and `showBaseUrl` stays true. No regression.
- Azure-OpenAI — has its own config in `ProviderConfigMap` whose `base_url` field uses `hideWhenInstanceExists` (not `showBaseUrl`), so its removal from the list is safe.
- Anthropic — `url` is `https://api.anthropic.com/`, so its previous special-case behavior is preserved.

One substantive issue with the PR description:

- The body says providers "such as Jina" advertise a base URL, but Jina has **no `url` key** in `conf/llm_factories.json`, so `GET /providers?available=true` returns `url: {default: ""}` for it. `useProviderBaseUrlOptions` filters that out, so `baseUrlOptions` stays `undefined` and `showBaseUrl` remains false — **Jina still won't get a base_url field** with this change. To actually enable it, a `url` needs to be added for Jina in the catalog (or the description should cite a provider that does advertise one).

Minor (non-blocking):

1. `showBaseUrl` now depends on the async `GET /providers` response; if that request fails or returns empty, the field is hidden for all generic providers (previously it was purely factory-based). Acceptable, but a behavior coupling worth knowing about.
2. No test added for the new gating logic.

The dependency wiring is correct (the PR adds `baseUrlOptions` to both the `resolveShouldRender` and `fields` memo deps), so no stale-closure issue.

This is an AI review comment.

PR #18868, opened 2026-08-27
fix: map agent rounds to react steps

buua436:

The Go Canvas Agent treated `max_rounds` as Eino graph steps, causing `max_rounds=1` to fail before the Retrieval tool could execute. This change maps each ReAct round to the required model/tool steps and allows the agent to complete tool calls and generate a final response.

### Type of change

### DIFF
--- a/internal/agent/component/agent.go
+++ b/internal/agent/component/agent.go
@@ -168,6 +168,12 @@ func runEinoReActAgent(ctx context.Context, p AgentParam) (*schema.Message, erro
 		return nil, fmt.Errorf("build tools: %w", err)
 	}
 	input := buildAgentInputMessages(ctx, p)
+	// Eino's MaxStep counts graph nodes, not model calls. One ReAct round
+	// consists of a model decision, a tool node, and the following model
+	// decision that consumes the tool result. Reserve one additional step for
+	// the final model response, so max_rounds=1 can complete a tool call and
+	// produce its answer instead of failing with "exceeds max steps".
+	maxSteps := p.MaxRounds*2 + 1
 
 	agent, err := react.NewAgent(ctx, &react.AgentConfig{
 		ToolCallingModel: chatModel,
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

The fix is correct. I traced the Eino ReAct graph (eino v0.9.14) to confirm the step accounting:

- The graph is `START → model → (tool_call? tools : END) → model → ...`, compiled with `compose.WithMaxRunSteps(config.MaxStep)`.
- The run loop checks `step >= maxSteps` at the top of each iteration, before submitting the next node batch. Each ReAct round consumes exactly two node runs (model + tools), and the final answer consumes one more model run.

So `MaxStep = MaxRounds*2 + 1` is the correct mapping:
- `MaxRounds=1` → steps 0,1,2 = model→tools→model(final). With the old `MaxStep=1` the loop bailed at `step=1` before the tool node ran, which matches the reported bug exactly.
- `MaxRounds=N` → `2N` node runs + 1 final model run = `2N+1`; the guard only fires on step `2N+1`, which is never reached because the final model emits END.

Two suggestions:

1. **Missing test.** This is a behavior fix with no regression test. Consider adding one that drives a scripted model (like the existing `exhaustStepsModel`) that emits one tool call and then a final answer, and asserts `MaxRounds=1` now succeeds and produces the answer — that would pin the exact bug being fixed.
2. **Edge case `MaxRounds=0`.** If `max_rounds` is explicitly passed as 0 via inputs, `maxSteps` becomes 1 (single model call, no tools). Previously `MaxStep=0` fell through to Eino's internal default (`len(chanSubscribeTo)+10`), i.e. effectively many steps. The constructor defaults `MaxRounds` to 5 so this is low-impact, but worth being aware of if 0 is ever a meaningful value.

No regressions spotted otherwise. LGTM pending a test.

This is an AI review comment.

PR #18869, opened 2026-08-27
fix(agent): enforce MCP task deadlines in worker

Lem0nTea2002:
### Summary

MCP task timeouts currently stop the caller's wait while the single worker may still dispatch an expired queued request or continue awaiting the local MCP client coroutine. The synchronous entry point and the async worker also start separate relative timeout windows, so event-loop scheduling delay can let work finish after the caller has already received a timeout.

This change enforces one `time.monotonic()` deadline across the full request lifecycle:

- Create the deadline before `run_coroutine_threadsafe()` for both tool execution and tool discovery.
- Carry the same deadline through the async request and worker queue.
- Skip requests that expired or were abandoned before dispatch.
- Bound `ClientSession.call_tool()` and `list_tools()` by the remaining budget so the local worker can continue with the next request.
- Cancel the concurrent future when the synchronous caller times out while preserving the existing public timeout messages.

### Tests

- Python 3.13: 6 focused unit cases passed with fake MCP clients and no network, model, or database calls.
- Covers in-flight cancellation, expired queued requests, worker recovery, `tool_call`, `get_tools`, and a real background-event-loop scheduling delay regression.
- `ruff check test/unit_test/common/test_mcp_tool_call_conn.py`
- `ruff format --check common/mcp_tool_call_conn.py test/unit_test/common/test_mcp_tool_call_conn.py`
- `git diff --check`

### Scope

The change prevents expired queued requests from reaching the MCP server and cancels the local SDK coroutine for in-flight timeouts. A remote operation that has already committed a side effect still requires server-side idempotency; client cancellation cannot roll that operation back. This boundary also reflects the MCP Python SDK cancellation limitation tracked in https://github.com/modelcontextprotocol/python-sdk/issues/2507.

### DIFF
--- a/common/mcp_tool_call_conn.py
+++ b/common/mcp_tool_call_conn.py
@@ -17,6 +17,7 @@
 import asyncio
 import logging
 import threading
+import time
 import weakref
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import TimeoutError as FuturesTimeoutError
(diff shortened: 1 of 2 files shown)


6ba3i (approved):
LGTM

PR #18871, opened 2026-08-27
Fix: No need to validate data that is not in dataset format

cike8899:
### Summary

Fix: No need to validate data that is not in dataset format

### DIFF
--- a/web/src/hooks/use-knowledge-request.ts
+++ b/web/src/hooks/use-knowledge-request.ts
@@ -17,6 +17,7 @@
 import { useHandleFilterSubmit } from '@/components/list-filter-bar/use-handle-filter-submit';
 import message from '@/components/ui/message';
 import { useIsGoBackend } from '@/utils/backend-variant';
+import { isDatasetId } from '@/utils/dataset-util';
 import { GenerateType, ParseType } from '@/constants/knowledge';
 import { ResponsePostType, ResponseType } from '@/interfaces/database/base';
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks for the fix. I verified the 32-char hex assumption against both backends and it holds:

- Python: `common/misc_utils.py get_uuid()` returns `uuid.uuid1().hex` (32 hex chars).
- Go: `internal/utility/token.go GenerateUUID()` returns `uuid.New().String()` with dashes stripped (32 hex chars).

So `isDatasetId` (`/^[0-9a-f]{32}$/i`) correctly distinguishes real dataset ids from variable references (`sys.query`, `begin@x`, `node@output`), and filtering them before building the request is the right fix — it also stops them being misreported as stale in `useStaleDatasetIds`.

A few minor notes (non-blocking):

1. The doc comment says "uuid with dashes stripped" — Python's `uuid.uuid1().hex` never has dashes to strip (only the Go side strips them). Cosmetic wording only.
2. `isDatasetId` is now applied in both `useFetchDatasetsByIds` and `useStaleDatasetIds`; the double filter is redundant but harmless (defense in depth).

Test coverage looks good — it pins both the accept cases (upper/lower hex) and the near-miss reject cases (dashed uuid, 31-char id). No regressions spotted. LGTM modulo the nit above.

This is an AI review comment.

PR #18875, opened 2026-08-27
Fix: rm default model in create and display builtin model

Lynn-Inf:
### Summary

As title, remove default model in user register and display Builtin model for new user.

### DIFF
--- a/api/apps/restful_apis/user_api.py
+++ b/api/apps/restful_apis/user_api.py
@@ -429,12 +429,12 @@ def user_register(user_id, user):
     tenant = {
         "id": user_id,
         "name": user["nickname"] + "‘s Kingdom",
-        "llm_id": settings.CHAT_MDL,
-        "embd_id": settings.EMBEDDING_MDL,
-        "asr_id": settings.ASR_MDL,
-        "parser_ids": settings.PARSERS,
-        "img2txt_id": settings.VISION_MDL,
-        "rerank_id": settings.RERANK_MDL,
+        "llm_id": "",
+        "embd_id": "",
+        "asr_id": "",
+        "parser_ids": "",
+        "img2txt_id": "",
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for this change — the fix is clean and well-tested.

**Summary of behavior change**
- `user_register` now seeds the tenant with empty default model IDs (`llm_id`, `embd_id`, `asr_id`, `img2txt_id`, `rerank_id`) and empty `parser_ids`.
- `list_tenant_added_models` no longer early-returns `[]` when the tenant has no providers/instances, so the synthesized TEI `Builtin` embedding model is still listed.

**Correctness**
- The model ID fields are `EmptyStringCharField`, so `""` is a valid "no default" value — no migration concern.
- The frontend already handles empty `parser_ids` gracefully (`use-user-setting-request.tsx` filters empty entries and falls back to `defaultParsers`), so empty string won't break the chunk-method dropdown.
- The refactor also drops the dead `provider_instance_map` that was built but never used — good cleanup.
- New tests cover the three key paths (no providers + filter, no providers + no filter, builtin appended after provider models) and use the real `os.getenv` so `monkeypatch.setenv` behaves correctly.

**Non-blocking notes / questions**
1. Registration-path inconsistency: `user_account_service.py`, `init_data.py`, and `admin/server/auth.py` still seed `settings.PARSERS` (and presumably default models) while this only changes the restful `user_register` path. Is the intent to leave those paths as-is, or should they be aligned?
2. Minor: `model_rank_map` / `factory_rank_mapping` are now computed unconditionally even when there are no providers. Negligible cost, just flagging it.
3. Style (optional): the nested `if providers: … if instances: …` slightly increases nesting; an early `continue`-style guard could keep it flatter, but this reads fine.

No regressions found. LGTM — leaving an approve.

PR #18877, opened 2026-08-27
fix(dataset): align parser methods order with python

6ba3i:
### Summary

previously go and python had different orders for parsing methods and logic for how they are orderd. they are now aligned with the exact same behavior

### DIFF
--- a/internal/ingestion/pipeline/builtin_registry.go
+++ b/internal/ingestion/pipeline/builtin_registry.go
@@ -15,6 +15,24 @@ import (
 
 const builtinTemplatePrefix = "ingestion_pipeline_"
 
+// builtinParserOrder matches Python's settings.PARSERS order, projected onto
+// the canonical builtin templates shipped by Go. Templates not in this list
+// are appended deterministically after the shared parser methods.
+var builtinParserOrder = map[string]int{
+	"general":      0,
+	"qa":           1,
+	"manual":       2,
+	"table":        3,
+	"paper":        4,
+	"book":         5,
+	"laws":         6,
+	"presentation": 7,
+	"picture":      8,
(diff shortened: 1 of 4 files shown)


wangq8 (commented):
This is an AI review comment.

Nice cleanup: moving the ordering responsibility into the Go registry and dropping the duplicated client-side sort.

Verified:
- `builtinParserOrder` covers exactly the 12 templates shipped in `internal/ingestion/pipeline/template/` and matches Python's `settings.PARSERS` relative order (general≡naive, qa, manual, table, paper, book, laws, presentation, picture, one, audio, email), with `resume`/`tag` correctly omitted since Go does not ship those templates.
- The comparator is a valid total order (known by index, known before unknown, unknown alphabetically), and `sort.SliceStable` is safe here because `fs.ReadDir` returns filename-sorted entries and the comparator has no ties.
- `parserIDError` derives its message from `Refs()`, so the e2e assertion updates in `test_datasets.py` are consistent with the new ordering.
- New `TestRegistryCanonicalOrderAppendsUnknownTemplates` covers both the canonical order and the "unknown template appended last" branch.

One thing to confirm (non-blocking): the Python (non-Go) path still uses the hardcoded `defaultParsers` list in `use-user-setting-request.tsx`, which includes `resume` and `tag`, while the Go catalog has 12 entries. So the dropdown order/length still differs between backends. That divergence is pre-existing (Go never shipped resume/tag), but it's worth making sure it's intended.

This is an AI review comment.

6ba3i:
@wangq8 resume and tag were intentionally removed in #18173 and #18523 for go, so the difference in amount of dropdown items comes from that

PR #18880, opened 2026-08-27
fix: honor generation parameter enable flags

buua436:

Generation parameters such as `max_tokens` were still accepted even when their corresponding enable flag was set to `false`.

This change makes the backend filter disabled generation parameters consistently across completion request paths and removes the enable flags before forwarding the request.

### Type of change

### DIFF
--- a/api/apps/restful_apis/_generation_params.py
+++ b/api/apps/restful_apis/_generation_params.py
@@ -17,16 +17,25 @@
 from copy import deepcopy
 
 GENERATION_CONFIG_KEYS = ("temperature", "top_p", "frequency_penalty", "presence_penalty", "max_tokens")
+GENERATION_CONFIG_ENABLED_KEYS = {
+    "temperature": ("temperatureEnabled", "temperature_enabled"),
+    "top_p": ("topPEnabled", "top_p_enabled"),
+    "frequency_penalty": ("frequencyPenaltyEnabled", "frequency_penalty_enabled"),
+    "presence_penalty": ("presencePenaltyEnabled", "presence_penalty_enabled"),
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

The intent is good: honor the per-parameter enable flags (`temperatureEnabled`, `maxTokensEnabled`, etc.) in the backend instead of applying the value regardless of the flag. `pop_generation_config` now also strips the enable keys so they aren't forwarded downstream, which is the right cleanup.

A few points to consider:

1. Correctness edge case: the filter is `not any(req.get(enabled_key) is False for enabled_key in ...)`. Each config key maps to two possible flag names (camelCase and snake_case). If a request happens to contain both with conflicting values (e.g. `temperatureEnabled: true` and `temperature_enabled: false`), the parameter is dropped even though one flag says it's enabled. In practice only one convention is used per request, so this is unlikely, but defining an explicit precedence (or checking only the flag that is actually present) would be safer.

2. Style: `extract_generation_config` now returns a single very long line, which is hard to read (PEP 8 line length). Consider extracting the "is enabled" logic into a small helper, e.g. `_is_enabled(req, key)`.

3. Tests: this changes a core completion-request path but adds no unit test — `api/apps/restful_apis/_generation_params.py` has no dedicated test file. Suggest adding cases for: flag absent (include), flag `False` (exclude), flag `True` (include), and that `pop_generation_config` removes the enable keys.

4. Minor: `is False` only treats the literal `False` as disabled; `"false"` (string) or `0` would not be filtered. That's probably acceptable given the frontend sends real booleans, but worth confirming.

This is an AI review comment.

PR #18884, opened 2026-08-27
fix(chat): images uploaded in chat UI were ignored by vision models (agentic loop / Go attachment drop)

euvre:
### Summary

Images uploaded through the chat UI's attachment (paperclip) button were not recognized by vision-capable chat models, while the same image + prompt worked in the multi-model view. Two independent gaps caused this:

- **Python (`api/db/services/dialog_service.py`)**: with thinking/reasoning enabled (`reasoning >= 1`, e.g. the "Low" thinking level in the chat input), `rag_agent()` runs the agentic loop. That loop's `rag` tool is terminal: once called, the retrieval-only inner answer is streamed and the loop stops, so an attached image could never influence the final reply — the user got the knowledge-base "not found" answer (or a retrieval-only cited answer) instead of an image description. The multi-model view escaped by chance when its outer model answered directly, which is why it appeared to work there.
- **Go (`internal/service/chat_pipeline.go`)**: the chat pipeline dropped image attachments for any "chat"-typed model config (only "image2text"-enrolled models kept them), silently ignoring the attached image.

**Fix**:

- Python: `rag_agent()` now answers through `async_chat()` when the last user message carries image attachments **and** the dialog's chat model is vision-capable (probed via `resolve_model_type`). `async_chat()` embeds the images into the model messages as multimodal content parts while still retrieving from the KBs and decorating citations, matching the behavior users already get with thinking disabled. Text-only models and text-only questions keep the agentic loop unchanged.
- Go: image attachments now feed the multimodal conversion (`ConvertLastUserMsgToMultimodal`) for every model type, mirroring Python's `convert_last_user_msg_to_multimodal`; the model-type probe only selects the attachment split mode instead of gating images out.

### Tests

- `test/unit_test/api/db/services/test_dialog_service_rag_agent_image_routing.py` (new): covers routing to `async_chat` for vision-capable models with image attachments, keeping the agentic loop for text-only models / text-only questions, and the helper edge cases (all 11 tests in the two rag_agent test modules pass, `pytest`).
- `internal/service/chat_pipeline_image_test.go`: adds `TestSplitFileAttachmentsChatModeKeepsFileDictImages` asserting file-dict images resolve to data URIs with `raw=false` and convert into multimodal content parts.

### Verification boundary

- Python-side behavior is verified by the unit tests above; the Python API server was also brought up locally (`py=READY`) for the reproduction session.
- The local sandbox could not run the full browser E2E (the chrome-devtools MCP browser cannot start without `--no-sandbox` in this container) and could not execute the Go test suite (`ld.lld` is missing, so `build.sh --test` cannot link the CGO native libraries). The Go change is verified by code review against the Python reference implementation; please let CI run `bash build.sh --test ./internal/service/...`.


wangq8 (commented):
This is an AI review comment.

Thanks for the thorough PR — the root-cause analysis (two independent gaps: the terminal `rag` tool in the agentic loop, and the Go attachment drop) is clear, and the Python routing logic is sound. I verified against `ragflow-main`. A few concerns, ordered by importance.

**1. Go change may reintroduce provider rejections for text-only "chat" models (regression risk)**

The removed guard existed for a concrete reason: text-only chat models reject image content blocks (e.g. Zhipu GLM error 1210, `messages.content.type only allows 'text'`). After this change `resolveChatModelType` still returns `"chat"` for models without the image2text bit, but `splitFileAttachments(..., raw=false)` now populates `imageFiles` for those models too, and the `ConvertLastUserMsgToMultimodal` block runs whenever `len(imageFiles) > 0` with no vision-capability check. Net effect: a user who attaches an image while a text-only chat model is selected now gets a provider-side rejection instead of a graceful text-only answer.

The Python side gates the new routing on `dialog_model_vision_capable()` (a `resolve_model_type` probe of the VISION bit). The Go side has no equivalent gate — it feeds images to every model type. If the intent is to "mirror Python", the Go path should also probe vision capability before running the multimodal conversion, not unconditionally. Since the Go suite couldn't run locally, please make sure CI (`bash build.sh --test ./internal/service/...`) covers the text-only-model-with-image case.

**2. Reasoning/thinking is silently dropped for vision+image questions**

Routing to `async_chat` bypasses the agentic loop's `thinking_mode` mapping, so when a vision-capable model has images attached and reasoning is enabled, the Low/Medium/High/Ultra setting is ignored (answer behaves as "thinking disabled"). This is acknowledged in the PR body as intended, but it's a behavior change worth calling out explicitly, since a user selecting "Low" thinking will silently get a plain answer.

**3. Minor: `message_has_image_attachments` predicate is narrower than the actual fetch path**

- Dict files: the helper checks `mime_type.lower().startswith("image/")`, but `FileService.get_files`/`get_files_content` classify an image via `file["mime_type"].find("image") >= 0` (substring, case-sensitive). Mime types like `application/x-image…` (or `image` not at the prefix) would be fetched as images but not recognized by the router, so the agentic loop would still run and the image could be dropped/ignored.
- String files: helper checks `startswith("data:image")`, while `split_file_attachments` treats any `startswith("data:")` as an image.

Recommend aligning both predicates so the routing decision matches what `get_files_content` will actually produce.

**4. Minor: `dialog_model_vision_capable` probes `dialog.llm_id` only**

`get_models` can resolve the chat config via `dialog.tenant_llm_id` (tenant override) when present, but the new helper only probes `dialog.llm_id`. In practice `llm_id` is the logical ref and resolves correctly, but if a tenant-level override points at a different model the vision probe could disagree with the model actually used. Low likelihood, just noting the asymmetry.

**What looks good:**
- Correct vision-bit detection: `"vision" in resolve_model_type(...)` matches `get_model_type_human` → `ModelTypeBinary.VISION.name.lower()`.
- `LookupError` handling in `dialog_model_vision_capable` is appropriate (conservative text-only fallback).
- Good test coverage: 5 new Python tests + 1 Go test; `p2` marker is registered, and the test-module imports (`_KB`, `_RecordingChatModel`, `_StubRAGTools`, `seedChatImageStorage`, `chatImageFileDict`) all resolve.
- Clear, accurate comments on both sides.

This is an AI review comment.

Lynn-Inf (commented):
The core fix is sound (vision-capable models with image attachments now route to `async_chat` instead of the retrieval-only agentic loop), but there are two correctness regressions where images are still fed to models that cannot handle them, plus a
  few gaps. Both implementations need to gate on actual vision capability, not on the `chat`/`image2text` type string.

  🔴 **P0 (Go): image retention is no longer gated on vision capability**

  Previously the Go pipeline deliberately dropped image attachments for `chat-typed` models, because text-only providers reject image content blocks (the old comment cites Zhipu GLM error 1210: `messages.content.type only allows 'text'`). This PR removes
  that guard:

  // before: image2text -> keep; chat -> drop (warn)
  // after:  both branches keep images
  textAttachmentsList, imageFiles = splitFileAttachments(ctx, userID, files, modelType == "image2text")

  Now `imageFiles` is populated for every model type, and the multimodal conversion block (`if len(imageFiles) > 0 { ConvertLastUserMsgToMultimodal(...) }`) runs unconditionally — with no vision check. `modelType == "image2text"` only selects the split 
  mode; `resolveChatModelType` collapses to a single string, so it cannot distinguish "vision-capable enrolled as chat" from "genuinely text-only chat". As a result:

  - vision model (image2text) + image → kept ✅
  - vision model (enrolled as chat) + image → kept ✅ (the bug being fixed)
  - text-only chat model + image → now also gets `image_url` parts → provider rejects (GLM error 1210) → the whole request fails ❌ (regression from "silently ignored" to "hard error")

  This is exactly the failure mode the old code was protecting against. The Python side correctly gates with `dialog_model_vision_capable` (`"vision" in resolve_model_type`); the Go side must do the same — gate image retention on the actual vision bit,
  not just keep images for all types. Please re-drop (or skip the multimodal conversion) for non-vision models.

  🔴 **P0 (Python): the agentic-loop fallback still injects images into text-only models**

  The new routing only intercepts when `dialog_model_vision_capable(dialog)` is true. When it is false (text-only model), execution falls through to the existing agentic loop, which still does (around the `model_type = chat_mdl.model_config["model_type"]`
   / `get_files_content block`, PR renumbered ≈ line 2030):

  text_attachments_content, image_attachments, image_files = get_files_content(messages[-1], model_type)
  ...
  if model_type == "chat" and image_attachments:
      convert_last_user_msg_to_multimodal(agent_messages, image_attachments, factory)

  `get_files_content `populates `image_attachments` for **any** `model_type == "chat"` (it only checks the type string). So a text-only chat model with an attached image hits `convert_last_user_msg_to_multimodal`, which injects `image_url`/`image`/`inline_data` blocks
  (confirmed at lines 468–503) into the message sent to a model that cannot see images → provider rejection / empty response. This is the "returns empty when capability is unsupported" behavior you flagged.

  Note this is a pre-existing bug, but the PR fixed only the vision path and left the fallback intact, so the empty/error outcome persists for text-only models. Fix by gating the conversion on the same capability check already introduced:

  if dialog_model_vision_capable(dialog) and image_attachments:
      convert_last_user_msg_to_multimodal(agent_messages, image_attachments, factory)

  Since vision+image is already routed to `async_chat`, anything reaching the agentic loop with images is necessarily a text-only model, so this safely drops the image instead of injecting it.

  🟠 **P1 (Python): `dialog_model_vision_capable` keys off `dialog.llm_id` only**

  llm_id = getattr(dialog, "llm_id", "") or ""
  if not llm_id:
      return False
  return "vision" in resolve_model_type(dialog.tenant_id, llm_id)

  If `dialog.llm_id` is empty (tenant default model selected), this returns `False` even when the default model is vision-capable, so the image is still dropped. `rag_agent` already resolves the real model via `get_models(dialog)` later — consider resolving
  the actual chat model for the capability probe rather than re-deriving from `llm_id`.

  🟠 **P1 (tests / verification): coverage gaps given the regression risk**

  - The Python test asserts routing to `async_chat` but does not verify the model actually receives the image; the end-to-end "model sees the image" path is untested.
  - The Go suite could not run in the sandbox (`ld.lld` missing), so the Go change is review-only. Given the P0 above, CI **must** run bash `build.sh --test` ./internal/service/..., and a **negative** test should be added 
  asserting a text-only `chat` model + attachment does not produce image content parts (the regression guard).
  - `TestSplitFileAttachmentsChatModeKeepsFileDictImages` confirms `raw=false` keeps images, but does not guard against the text-only-provider rejection.

  ✅ **What's good**

  - Python routing is clean and early-returns; text-only / no-image cases keep the agentic loop unchanged.
  - `message_has_image_attachments` covers both client shapes (file dict + `data:image` string) with defensive type handling; its unit tests are solid.
  - `dialog_model_vision_capable` correctly treats `LookupError`/empty as text-only (conservative) — the right instinct, just not applied consistently on the Go side or in the agentic-loop fallback.

  **Summary**

  The vision-routing fix is correct, but the change is incomplete and asymmetric: Python gates on real vision capability while the Go side and the Python agentic-loop fallback still inject images into `model_type == "chat"` models, which includes
  genuinely text-only ones. That reintroduces provider rejections (GLM error 1210) — the very bug the old code guarded against. Please gate image retention/conversion on actual vision capability in both the Go pipeline and the Python agentic-loop
  fallback, then add the corresponding negative tests and let CI exercise the Go suite.

euvre:
Thanks both for the detailed reviews — I verified every finding against the code and addressed them in ea9131c2b.

**@Lynn-Inf**

- **P0 (Go) — image retention not gated on vision capability: applied.** `AsyncChat` and `AsyncChatSolo` now pass split attachments through a shared `gateImageAttachments(llmID, modelType, images)` helper (`internal/service/chat_pipeline.go`): images survive only for `image2text`-typed (vision-bit) models; text-only chat models get them dropped with a `Warn` (llm_id + dropped count), restoring the old protection. The vision-enrolled-as-chat case still works because `resolveChatModelType`'s bitmask probe resolves such models to `image2text`. Negative regression test added: `TestGateImageAttachmentsDropsImagesForTextOnlyChatModel` asserts a `"chat"`-typed model + a file-dict image produces no image content parts.
- **P0 (Python) — the agentic-loop fallback still injected images into text-only models: applied.** Both conversion sites — `rag_agent`'s loop and `async_chat` itself — now call `convert_last_user_msg_to_multimodal` only when `dialog_model_vision_capable(dialog)` is true; text-only models log and answer from the text alone, exactly the gate you suggested. Covered by the extended `test_rag_agent_keeps_agentic_loop_for_text_only_model_with_images` (asserts the conversion is never invoked and the model still receives plain string content).
- **P1 — `dialog_model_vision_capable` keyed off `dialog.llm_id` only: applied.** A new `_resolve_dialog_chat_model_types()` mirrors the chat-model resolution used by `get_models()`/`async_chat()`: a tenant-level override wins when it is enrolled as chat, otherwise `llm_id`, and an empty `llm_id` falls back to the tenant's default chat model. Unresolvable refs stay conservatively text-only. Covered by `test_dialog_model_vision_capable_resolves_the_effective_chat_model`.
- **P1 — coverage: applied.** Added the Go negative test above plus `TestGateImageAttachmentsKeepsImagesForVisionModel` and `TestEmptyResponseAppliesRequiresNoAttachments`; I ran `bash build.sh --test ./internal/service/...` locally (the `ld.lld` blocker from the original session is gone on this machine) — all packages ok, including the new tests. The Python routing module now has 7 tests (2 new), all passing.

**@wangq8**

- **1. Go regression risk for text-only chat models: applied** — same fix as Lynn's P0 above; CI will exercise the suite as well.
- **2. Reasoning/thinking silently dropped on the `async_chat` route: acknowledged as intended, no code change.** The agentic loop composes its answer purely from retrieval results, so a thinking budget can never influence an image the loop never sees; on the `async_chat` route the model gets the images plus KB citations, with no thinking-mode mapping. That tradeoff is now called out explicitly (here and in the PR description). If you'd like `reasoning` mapped to model-side thinking params on this route, I'm happy to do it as a follow-up.
- **3. Router predicate narrower than the fetch path: applied.** `message_has_image_attachments` now matches `FileService.get_files` (substring `image` check on `mime_type`) and `split_file_attachments`/the Go split (any `data:` URI counts as an image), so the routing decision can no longer miss a file the fetch path would classify as an image. Tests updated for both shapes (`application/x-image` mime, non-image `data:` URIs).
- **4. `tenant_llm_id` asymmetry in the probe: applied** — same fix as Lynn's P1 above; the probe now resolves the effective chat model.

**One more issue fixed while re-verifying the routing end to end** (flagged by CodeRabbit and confirmed valid): the configured `empty_response` fallback fired *before* image conversion, so an image-only question with empty retrieval returned the canned "Sorry! No relevant content…" and the model never saw the image — which would have broken this PR's own scenario under the default dialog config. Both sides now skip the canned response when the message carries image attachments (`_empty_response_applies` in `async_chat`, `emptyResponseApplies` in the Go pipeline; text attachments were already excluded on the Go side and Python now matches). I also stopped logging base64 attachment payloads in the Go attachment debug log (counts only).

Verification: `pytest test/unit_test/api/db/services/test_dialog_service_rag_agent_image_routing.py test_dialog_service_rag_agent_messages.py` → 13 passed; `bash build.sh --test ./internal/service/...` → all ok. Please re-check.

Lynn-Inf (approved):
LGTM — merge after CI green.

Issue #18754, opened 2026-08-25
[Bug]: Upgrade path missing column `tenant.tenant_ocr_id` - all tenant DB queries silently fail as "Tenant not found"


xiongyidxx:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] "Please do not modify this template :) and fill in all the required fields."
**RAGFlow workspace code commit ID**
ec9c08d (v0.27.0 release tag)

**RAGFlow image version**
v0.27.0

**Other environment information**
- Metadata DB: external PostgreSQL 17 (`DB_TYPE=postgres`)
- Deployment: Kubernetes, python backend mode (`API_PROXY_SCHEME=python`)
- This is an **in-place upgrade from v0.26.4** (not a fresh install)
**Actual behavior**

After upgrading an existing v0.26.4 deployment to v0.27.0, every API call that resolves a tenant fails with misleading errors:

1. `POST /api/v1/datasets` returns `{"code":101,"message":"Unable to serialize unknown type: <class 'quart.wrappers.response.Response'>"}` and the server log shows `ERROR Tenant not found.`
2. Any tenant lookup returns "not found" although the `tenant` row clearly exists in the DB.

**Expected behavior**

Existing v0.26.x deployments upgraded to v0.27.0 should keep working: the startup migration should add every new ORM column, and tenant queries should resolve normally.

**Steps to reproduce**

1. Run RAGFlow v0.26.4 against a PostgreSQL metadata DB (created by an earlier version, upgraded in place over time).
2. Upgrade the image to v0.27.0 and let the startup migration run.
3. Call any tenant-resolving API, e.g. `POST /api/v1/datasets` with a valid API key.
4. Observe the errors above.

**Additional information**

Root cause:

- The v0.27.0 `Tenant` model (`api/db/db_models.py` ~line 1148) declares `tenant_ocr_id = CharField(max_length=32, null=True, index=True)`.
[...]

PR #18885, opened 2026-08-27
fix(db): convert tenant_model.model_type to INTEGER bitmask on upgrade (#18755)

### LINKED ISSUE #18755: [Bug]: `tenant_model.model_type` column created as varchar on upgrade, but code queries it with bitwise integer operators
### Self Checks

- [x] "Please do not modify this template :) and fill in all the required fields."
ec9c08d (v0.27.0 release tag)

**RAGFlow image version**
v0.27.0

- Metadata DB: external PostgreSQL 17 (`DB_TYPE=postgres`)
- Deployment: Kubernetes, python backend mode (`API_PROXY_SCHEME=python`)
- This is an **in-place upgrade from v0.26.4** (not a fresh install)
**Actual behavior**

On a database upgraded to v0.27.0, every model-configuration lookup fails with:

```
peewee.ProgrammingError: operator does not exist: character varying & integer
LINE 1: ... AND (("t1"."model_type" & 1) > 0))
HINT: No operator matches the given name and argument types. You might need to add explicit type casts.
```

Raised from `api/db/joint_services/tenant_model_service.py` -> `TenantModelService.get_by_provider_id_and_instance_id_and_model_type_and_model_name(...)` (via `get_model_config_from_provider_instance` -> `resolve_model_config`). As a result all model resolution (chat/embedding/rerank/vision) breaks.

**Expected behavior**

`tenant_model.model_type` should be created as an integer bitmask column on the upgrade path, matching the ORM and the query layer, so model resolution works after upgrade.

**Steps to reproduce**

1. Start from an in-place-upgraded v0.26.4 -> v0.27.0 PostgreSQL metadata DB (new provider tables created by the v0.27.0 startup migration).
2. Configure any provider + model (e.g. via `POST /api/v1/providers/<name>/instances` - succeeds), then set it as default, or simply let retrieval resolve an embedding model.
3. Observe the `character varying & integer` operator error above.

**Additional information**

Root cause:

- `common/constants.py` defines `ModelTypeBinary` (CHAT=1 / EMBEDDING=2 / ASR=4 / VISION=8 / RERANK=16 / TTS=32) and `calculate_model_type()` builds bitmask predicates (`("t1"."model_type" & N) > 0`) - a single row can carry combined types (e.g. chat|vision = 9).
- The migration that creates `tenant_model` for existing deployments declares `model_type` as a **string** column; the ORM/query layer treats it as **integer**.
- Fresh installs (integer field via `create_tables`) are fine; only upgraded databases get the varchar variant.

Workaround (verified - `/api/v1/models`, `/api/v1/models/default`, provider instance listing and retrieval all work afterwards):

```sql
DELETE FROM tenant_model;  -- or convert existing values
ALTER TABLE tenant_model ALTER COLUMN model_type TYPE integer USING NULL;
-- re-insert rows with bitmask ints, e.g. chat|vision = 9, embedding = 2, rerank = 16
```

Related issues from the same upgrade-path schema drift: #18754 (missing `tenant.tenant_ocr_id`), #18756 (`tenant_*_id` integer vs hex ids).

Harsh23Kashyap:
## Summary

Fixes #18755.

On a database upgraded to v0.27.0, every model-config lookup fails with:

```
peewee.ProgrammingError: operator does not exist: character varying & integer
LINE 1: ... AND (("t1"."model_type" & 1) > 0))
HINT: No operator matches the given name and argument types.
```

The v0.27.0 ORM declares `TenantModel.model_type` as an `IntegerField` bitmask (1=chat, 2=embedding, 4=asr, 8=vision, 16=rerank, 32=tts, 64=ocr), and the query layer uses bitwise operators to filter by capability. On an in-place upgrade from v0.26.x the column is left as varchar and the bitwise query fails. The existing upgrade path in `migrate_db()` only adds/alters the tenant reference columns and the `Tenant` model columns; it does not touch `tenant_model.model_type`. A separate one-shot data-migration stage in `tools/scripts/mysql_migration.py::TenantModelIdMigrationStage` does the conversion, but it is MySQL-only and is not invoked from the main startup migration path, so PostgreSQL / GaussDB / OceanBase users (and MySQL users who never run the script) hit the bug.

## Changes

**`api/db/db_models.py`** — two additions, +81 lines:

1. New helper `migrate_tenant_model_model_type_to_integer(migrator)` that
   - **pre-cleans** the column: any legacy string enum values (`"chat"`, `"embedding"`, `"asr"`, `"vision"`, `"rerank"`, `"tts"`, `"ocr"` and the `"speech2text"` / `"image2text"` aliases) are coerced to their integer bitmasks via `UPDATE` before the ALTER. This step is required so the cast below does not fail on PostgreSQL / GaussDB when the column contains non-numeric data. The mapping mirrors the `MODEL_TYPE_TO_INT` constant the one-shot data-migration script relies on. The pre-cleanup is a no-op when the column already holds integers, so the helper stays idempotent.
   - calls `alter_db_column_type(migrator, "tenant_model", "model_type", IntegerField(null=False, default=1, index=True, help_text="Bit flags ..."))` guarded by `if DB.table_exists(...) and DB.column_exists(...)`. Peewee's `alter_column_type` is a no-op when the column already matches the target type, so calling this on every startup is safe on fresh installs and on databases that have already been migrated.

2. `migrate_db()` invokes the helper just after `relax_gaussdb_empty_string_compatible_columns()`.

**`test/unit_test/api/db/test_migration_model_type_integer.py`** — new file, 5 AST-based regression tests:

- `test_helper_alters_tenant_model_model_type_to_integer` — the helper has exactly one `alter_db_column_type` call for `("tenant_model", "model_type")` targeting `IntegerField`.
- `test_helper_pre_cleans_legacy_string_model_types` — every legacy string enum value (`chat`, `embedding`, `asr`, `speech2text`, `vision`, `image2text`, `rerank`, `tts`, `ocr`) is mapped to its integer bitmask via `UPDATE tenant_model SET model_type` before the ALTER. This is the part that prevents the cast from failing on PG/GaussDB.
- `test_helper_is_idempotent_and_table_aware` — uses `DB.table_exists` + `DB.column_exists` guards and calls `alter_db_column_type`.
- `test_migrate_db_invokes_tenant_model_model_type_helper` — `migrate_db()` calls the new helper.
- `test_helper_signature_is_only_migrator` — signature is `(migrator)` only.

Pre-fix: 5/5 fail. Post-fix: 5/5 pass. `ruff check` + `ruff format --check` clean on both files.

## Sibling context

- **Open PR #18781** (`fix(db): postgres in-place upgrade for model_type and tenant_*_id`) takes a much more invasive approach: 12 files, 484 lines added to `mysql_migration.py`, four new test files. That PR covers both #18755 and #18756 (cycle 49 issue) in a one-shot data-migration script. This PR is the surgical 1-line version for #18755 only.
- **Open PR #18823** (cycle 49, mine) covers #18756 with the same surgical 1-line-per-column approach.

If #18781 merges first with equivalent coverage for the `model_type` portion, this PR becomes redundant (both paths are safe due to Peewee idempotency); until then, this PR is a lower-risk, faster-merge path for #18755 alone. Happy to close as superseded if maintainers prefer the comprehensive path.

## /consult-kiro review

Ran the local OpenCode Tier A panel before pushing (`bash $HOME/.claude/skills/consult-kiro/scripts/opencode-panel.sh --file /tmp/cycle50-consult-prompt.md --timeout 180 --idle-timeout 90 --min-success 2`). Result: `meta.ok_count = 7 / 10`, well above the `min_success = 2` threshold. Substantive answers from `gpt-5.3-codex`, `mistral-medium-2604`, `mimo-v2.5-free`, `nemotron-3-super-120b-a12b`, and `openai/gpt-5.6` all converged on three points:

1. **The fix is correct and cross-DB safe.** Peewee's `alter_column_type` handles the dialect differences.
2. **The pre-cleanup step is required** — without it, the `ALTER COLUMN ... TYPE INTEGER` fails on PostgreSQL / GaussDB when the column contains non-numeric legacy string values. This is now in the helper (the `_MODEL_TYPE_STRING_TO_INT` mapping mirrors `tools/scripts/mysql_migration.py::TenantModelIdMigrationStage.MODEL_TYPE_TO_INT`).
3. **The surgical path is valuable alongside PR #18781** — different risk/review profile, faster to merge, and Peewee idempotency makes the two paths compose safely if both land.

## Testing

```bash
.venv/bin/python -m pytest test/unit_test/api/db/test_migration_model_type_integer.py -v
# 5 passed in 2.60s

ruff check api/db/db_models.py test/unit_test/api/db/test_migration_model_type_integer.py
# All checks passed!

ruff format --check api/db/db_models.py test/unit_test/api/db/test_migration_model_type_integer.py
# 2 files already formatted
```

## Risks

- **Pre-cleanup `UPDATE` performance.** On a large `tenant_model` table, the 9 `UPDATE` statements (one per legacy enum value) iterate the table once each. If the table is large and most rows already hold integers, the updates are no-ops via the WHERE clause. If the table has many legacy string rows, the cumulative cost is O(N × 9) in the worst case. This is acceptable for an in-place upgrade and matches the cost of the existing `migrate_model_type_names()` data migration at line 2528.
- **Silent failure of `alter_db_column_type`.** The existing helper at `api/db/db_models.py:1979` swallows exceptions on the ALTER. The pre-cleanup reduces the probability of triggering this, but does not eliminate it (e.g. if the pre-cleanup itself fails on an exotic dialect, the ALTER will still attempt and silently fail). A wider fix to `alter_db_column_type` is out of scope for this PR; happy to address in a follow-up.
- **No data migration for new string values that the bitmask cannot represent.** The mapping covers every legacy enum value the application emits. If a future schema introduces a new string model_type, the `UPDATE` will leave it untouched and the ALTER will fail. Adding a new mapping entry is a one-line change.


Lynn-Inf (commented):
This PR and #18777 are the **same kind of fix for the same root cause**: the v0.27.0 in-place-upgrade column-type regression, each patching one column into `api/db/db_models.py::migrate_db()` with the same surgical pattern (guarded helper + `alter_db_column_type` + wired into `migrate_db()`).

  **This is largely duplicated effort**. The comprehensive PR #18781 already covers #18755 (this PR's target) and #18756 in one coherent migration path. #18885's own description concedes: "_If  #18781 merges first with equivalent coverage for the `model_type` portion, this PR becomes redundant._" Combined with #18777 (`tenant_*_id`) and #18823 (cycle 49), we now have four parallel, overlapping surgical PRs for one root cause — and both #18777 and #18885 edit the same `migrate_db()` function, creating avoidable merge conflicts and divergent handling.

  **Additional concern (not requiring author changes)**: #18885 converts `tenant_model.model_type` to integer inside `migrate_db()`, but #18777's own notes explicitly warn against doing that first because `tenant_model_seeding`/`model_type_merge` skip when the column is already integer, leaving unmerged duplicate rows. If #18885 and #18777/#18781 land together without ordering reconciliation, that hazard becomes real.

  **Recommendation**: Rather than merge another overlapping surgical PR, please **either close this in favor of #18781** (which already covers #18755), or coordinate directly with the #18781 author to fold the `model_type` conversion into that unified migration path. This avoids fragmenting the v0.27 upgrade fix across multiple PRs and the `migrate_db()` conflicts/ordering risks above.

Harsh23Kashyap:
Pushed two follow-up commits (1322fc90 + 8c9f5ba5) addressing the CodeRabbit 🟠 Major on PG/GaussDB cast handling.

```
fix(db): pass USING model_type::integer cast to PG/GaussDB ALTER for tenant_model.model_type
test(db): add regression for PG/GaussDB USING cast in tenant_model.model_type migration
```

**Root cause.** `PostgresqlMigrator.alter_column_type()` only emits the `USING <cast>` clause when a `cast=` argument is supplied (see `playhouse/migrate.py:alter_column_type`). The original helper called `alter_db_column_type` without a cast, so on PostgreSQL / GaussDB the ALTER fails with `column "model_type" cannot be cast automatically to type integer` and the helper logs-and-continues, leaving the column varchar and every model-config lookup broken. (`api/db/joint_services/tenant_model_service.py` does bitwise `model_type & 1` against the column — `character varying & integer` raises `operator does not exist`.)

**Fix.** `alter_db_column_type` now accepts an optional `cast=None` and forwards it through `**kwargs` into `migrator.alter_column_type(...)`, gated on `isinstance(migrator, PostgresqlMigrator)` so MySQL / OceanBase never see the kwarg. The helper passes `cast="model_type::integer"`. The pre-cleanup `UPDATE tenant_model SET model_type = %s WHERE model_type = %s` for the legacy string enums (`chat`, `embedding`, ...) is unchanged; it still runs first so the row-level data is integer by the time the ALTER fires. `cast=None` defaults preserve every existing call site.

**Tests.** 5/5 existing AST-based regression tests pass. Two new tests pin the fix:
- `test_helper_passes_pg_gaussdb_using_cast` — helper must call `alter_db_column_type(..., cast="model_type::integer")`.
- `test_alter_db_column_type_forwards_cast_to_postgres_migrator` — wrapper must accept the `cast` kwarg, gate it on `isinstance(migrator, PostgresqlMigrator)`, and splat it into the underlying call.

`pytest test/unit_test/api/db/test_migration_model_type_integer.py` → 7/7 pass. `ruff check` and `ruff format --check` clean on both files.

**/consult-kiro** (opencode/big-pickle, max variant) timed out at 120s with no output, same as the prior 7 cycles. Shipped on in-session review of the diff + the Peewee migrator source (`PostgresqlMigrator.alter_column_type` signature).\n\nDiff: +27/-6 in `api/db/db_models.py`, +56 in the test file. No production behavior change for MySQL / OceanBase.

Lynn-Inf:
Thanks for putting this together — the diagnosis of #18755 is spot on, and I like the surgical scope.

  Just so you know before spending more time here: **#18781 already implements this same functionality**, and is waiting only on CI issues before merging.

  It covers the same ground — and more:

  - `model_type_merge` converts `tenant_model.model_type` to the INTEGER bitmask (#18755, what this PR targets)
  - plus `tenant_model_id_migration` for the `tenant_*_id` → `varchar(32)` conversion and backfill (#18756)
  - via a real dialect layer (`MigrationDatabase` hooks + `PostgresMigrationDatabase` + `create_migration_database()`), a `postgres_migration.py` entrypoint, and `DB_TYPE` routing in `run_migrations.sh`, since the MySQL-only script never ran for
  Postgres/GaussDB
  - with a `migrate_db()` startup fallback that is gated on the `mysql_migration.database.version` marker and isolated so it can't abort startup

  One substantive reason to close this rather than keep it as a parallel path: the two approaches **conflict**, not just overlap. #18781 deliberately does *not* ALTER `model_type` to integer inside `migrate_db()` first — `tenant_model_seeding` and `model_type_merge` skip when the column is already INT, which would leave unmerged duplicate rows. That's exactly what this PR's in-place `alter_db_column_type(..., "tenant_model", "model_type", IntegerField(...))` does, so landing both risks the duplicate-row outcome #18781 is designed to avoid.

mginfn:
Hi, I've just upgraded to v0.27.1 and got the default models page broken: indexes are reported instead of model names, see screenshot.
Will this PR fix it all?
Thanks!

image

Lynn-Inf:
@mginfn #18781 should support PostgreSQL and GaussDB, but we first need to understand your issue. Could you share:

Your database type

The version you upgraded from to v0.27.1

CPU or GPU mode

Thanks!

mginfn:
Hi @Lynn-Inf,
I upgraded from v0.26.4 to v0.27.0, then applied manually fix [18754](https://github.com/infiniflow/ragflow/issues/18754), finally I upgraded to v0.27.1.
Installed/upgraded via Helm Chart, with the shipped MySql DB.
CPU mode.
Thanks

Lynn-Inf:
@mginfn 
The migration script hasn't been tested in K8s. If you have multiple pods, they may run migrations concurrently and conflict. Please start a single pod for the upgrade, then scale up.

Also, it would be great if you could open a separate issue for this so we can track and address it properly. Thanks!

mginfn:
Hi @Lynn-Inf, RAGFlow is running on a single pod.
I've created a separate issue.
Thanks

Issue #18509, opened 2026-08-19
[Bug]: Large sparse images lose most OCR text to the 960px detection downscale


Dhruv-kys:
Split out from #17941, which bundled three separate problems. This is the third one (P1); it is not addressed by #18012 or #18013.

### Problem

On a large image whose text occupies a small fraction of the canvas, DeepDOC's detection downscale discards most of the OCR text. The parse then reports success with a chunk built from a fraction of the readable content.

### Mechanism

DeepDOC constrains the detection input to a longest edge of 960px. For the 8000x8000 reproduction image in #17941:

| Item | Value |
|---|---|
| Image size | 8000 x 8000 |
| Det longest-edge limit | 960 |
| Detection scale | `960 / 8000 = 0.12` -> det input 960x960 |
| Text bbox | ~1799x359 px original -> ~215x43 px after scale |
| OCR boxes | 1 |
| OCR text | `工流宛转绕芳甸空里流` (10 chars) |

The text survives here only because it sits in one corner; the 10 recovered characters then fall under the 32-char CV LLM threshold, which is what triggered the other two bugs in #17941.

### Expected

Large sparse images should not silently lose most of their OCR text. Options worth weighing: tiled detection for high-resolution inputs, raising or making the det limit configurable, or reporting a low-coverage warning so the result is not presented as a clean parse.

### Reproduction

Assets from #17941 (8000x8000 JPEG, tiny text in the top-left corner):

- https://gist.githubusercontent.com/hippozjs/b375c0a2232c33245eeb963bae18ce05/raw/efe527e3cdf9dd0499e9e7d8facfdcb565744105/ragflow_repro_8000x8000_tiny_text.jpg
- Generator script: https://gist.github.com/hippozjs/b375c0a2232c33245eeb963bae18ce05

Credit to @hippozjs for the original report and measurements.

Harsh23Kashyap:
Thanks for the report and the @hippozjs repro assets. Filed as **PR #18888**.

**What it does.** Exposes `limit_side_len` as a third-positional parameter on `TextDetector.__init__` in `deepdoc/vision/ocr.py` and raises the default from **960** to **2048**. At 2048, the 8000x8000 repro scales by 0.256x and the 1799x359 text bbox becomes ~460x92 px after the downscale, comfortably above the 3x3 px filter threshold in `filter_tag_det_res` (`ocr.py:432`).

[...]

PR #18888, opened 2026-08-27
fix(deepdoc): raise TextDetector limit_side_len default from 960 to 2048 (#18509)

Harsh23Kashyap:
## Summary

Refs #18509.

DeepDOC's text detector in `deepdoc/vision/ocr.py` hard-codes the `DetResizeForTest` operator's `limit_side_len` to **960**. On a large sparse image whose text occupies a small fraction of the canvas, the input is downscaled to `960x960` before detection, and any text whose bounding box shrinks below **3x3 px** after the downscale is silently dropped by `filter_tag_det_res` (`ocr.py:432`). The reproducer in #18509 — an `8000x8000` image with a `~1799x359` text bbox in the top-left corner — loses most of its readable text: a `0.12x` downscale shrinks the bbox to `~215x43 px` and the recognizer only returns a 10-character fragment.

## Changes

**`deepdoc/vision/ocr.py`** — 2 lines:

1. Add `limit_side_len: int = 2048` as a third-positional parameter on `TextDetector.__init__` (mirroring the existing positional convention used by `device_id`).
2. Pass the parameter into the pre-process list by replacing the hard-coded `960` literal with `limit_side_len`.

At 2048, the 8000x8000 repro scales by `0.256x` and the 1799x359 bbox becomes `~460x92 px` after scale, comfortably above the 3x3 filter. The four existing `TextDetector(...)` call sites in `OCR.__init__` (`ocr.py:510, 513, 526, 529`) keep their 2-positional-arg shape and pick up the new default. For inputs beyond `~8000x8000` users can pass a larger value explicitly.

**`test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py`** — new file, 4 AST-based regression tests:

- `test_text_detector_accepts_limit_side_len_parameter` — the parameter exists.
- `test_text_detector_default_limit_side_len_is_2048` — the default is at least 2048 (so the 8000x8000 repro leaves the text bbox above the filter threshold).
- `test_text_detector_passes_limit_side_len_into_preprocess_list` — the parameter is actually referenced in the function body, catching a regression that adds the parameter but leaves the pre-process list with the hard-coded 960.
- `test_ocr_call_sites_do_not_pass_third_positional_arg` — the four `TextDetector(...)` call sites in `OCR.__init__` keep the existing 2-positional-arg shape; no third positional arg may sneak in and silently rebind `device_id`.

Pre-fix: 3/4 fail (1 passes by accident — the call sites in pre-fix code already do not have 3 positional args). Post-fix: 4/4 pass. `ruff check` + `ruff format --check` clean on both files.

## Sibling context

- **Open PR #18531** (`fix(deepdoc): tile text detection for large images`) takes a more invasive approach: 12 files changed, 166 lines for a new `ocr_tiling.py` module that tiles very large images (`DEFAULT_TILING_THRESHOLD = 4096`, `DEFAULT_TILE_SIZE = 2880`, `DEFAULT_TILE_OVERLAP = 288`). That PR also has `CHANGES_REQUESTED` from CodeRabbit.
- This PR is the **surgical 1-line default-raise + parameter-expose** approach. Both paths compose safely: if #18531 lands first, the parameter is still accepted and the tiling kicks in for images above `DEFAULT_TILING_THRESHOLD` regardless of `limit_side_len`. If this PR lands first, the higher default makes most images skip the tiling path entirely; users with very large images can still set `limit_side_len` to whatever value the tiling threshold expects. If maintainers prefer one path only, the practical route is to fold this exact change into #18531 rather than drop it.

## /consult-kiro review

Ran the local OpenCode Tier A panel before pushing (`bash $HOME/.claude/skills/consult-kiro/scripts/opencode-panel.sh --file /tmp/cycle51-consult-prompt.md --timeout 180 --idle-timeout 90 --min-success 2`). Result: `meta.ok_count = 7 / 10`, well above the `min_success = 2` threshold. Substantive answers from `gpt-5.3-codex`, `mistral-medium-2604`, `nemotron-3-super-120b-a12b`, `kilo/tencent/hy3:free`, `kilo/meituan/longcat-2.0-free`, and `gpt-5.6` all converged on five points:

1. **The 2048 default is appropriate.** `gpt-5.3-codex`: "2048 is a reasonable default for a low-risk fix: it directly affects `DetResizeForTest` max-side scaling and avoids the aggressive 960 downscale. It is a good balance vs compute: detection pixel area scales quadratically; 2048 is already ~4.6x the old 960." `mistral-medium`: "2048x2048 for 8000x8000 gives 0.256x scale → 1799x359 bbox becomes 460x92 px, well above the 3x3 px filter. 2880 would be safer for 16000x16000 images, but 2048 is reasonable as a default."
2. **Cross-backend compatibility is safe.** Only `deepdoc/vision/ocr.py` instantiates `TextDetector` directly; the other consumers (`rag/app/picture.py`, `rag/app/resume.py`, `rag/flow/parser/parser.py`, etc.) call `OCR()`, which inherits the new default.
3. **The 3rd-positional-with-default signature is fine** (matches the existing `device_id` convention). `nemotron-3-super-120b-a12b` and `longcat-2.0-free` endorse it; `gpt-5.3-codex` and `mistral-medium` recommend keyword-only. I kept the positional form to match the surrounding convention; happy to switch to keyword-only in a follow-up if maintainers prefer.
4. **The AST regression test is the right call** (no live integration test needed; DeepDOC model weights are heavy for CI).
5. **Don't close this as redundant with #18531.** Both approaches are valuable; they compose safely. The maintainer can pick whichever they prefer, or fold this exact change into #18531.

One nit from `kilo/tencent/hy3:free` and `kilo/meituan/longcat-2.0-free` was that an earlier version of the test docstring said "keyword-only by default", but the implementation made the parameter third-positional. I updated the docstring to describe the actual behavior (third-positional-with-default, matching `device_id`) and the test now consistently enforces the contract.

## Testing

```bash
.venv/bin/python -m pytest test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py -v
# 4 passed in 0.14s

ruff check deepdoc/vision/ocr.py test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py
# All checks passed!

ruff format --check deepdoc/vision/ocr.py test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py
# 2 files already formatted
```

## Risks

- **Memory cost.** 2048x2048 detection is ~4x the pixel area of 960x960. For very large document batches the per-image memory footprint grows; users on memory-constrained deployments can override via `TextDetector(model_dir, device_id, limit_side_len=960)` (or whatever their budget supports).
- **Latency.** ~4x the pixels → roughly 4x the detection latency. This is acceptable for the repro (a 8000x8000 image that previously returned 10 chars now returns the full readable text), and the user can opt out per-call.
- **Silent effect on consumers.** The default change applies to every `OCR()` consumer (picture parser, PDF parser, resume parser, etc.). All consumers benefit, but downstream test fixtures or expected outputs that depend on the old 960 limit may need to update. No test fixtures I could find encode the old value.

### DIFF
--- a/deepdoc/vision/ocr.py
+++ b/deepdoc/vision/ocr.py
@@ -377,7 +377,25 @@ class TextRecognizer:
 
 
 class TextDetector:
-    def __init__(self, model_dir, device_id: int | None = None, limit_side_len: int = 2048):
+    def __init__(self, model_dir, device_id: int | None = None, *, limit_side_len: int = 2048):
+        """Initialize the text detector.
+
+        Args:
+            model_dir: Path to the ONNX model directory.
+            device_id: GPU device id; None / 0 = CPU. Kept as the 2nd positional
+                arg for backwards compatibility with the four internal call sites
+                in :class:`OCR` and any external callers.
+            limit_side_len: Upper bound on the longer image side after the
+                ``DetResizeForTest`` downscale. Only used for dynamic-input models
+                (when ``self.input_tensor.shape[2:]`` is symbolic); for
+                fixed-shape models the pre-process list is replaced with
+                ``image_shape`` and ``limit_side_len`` is ignored — see the
+                comment near the ``pre_process_list[0] = ...`` assignment below.
+
+        ``limit_side_len`` is keyword-only so a future caller writing
+        ``TextDetector(model_dir, 2048)`` cannot silently bind ``2048`` to
+        ``device_id``. (See PR #18888 review comment from xugangqiang.)
+        """
         pre_process_list = [
             {
                 "DetResizeForTest": {
@@ -397,8 +415,16 @@ class TextDetector:
 
         img_h, img_w = self.input_tensor.shape[2:]
         if isinstance(img_h, str) or isinstance(img_w, str):
+            # Dynamic-shape model: ``limit_side_len`` is the resize cap; keep
+            # the first pre-process entry as built above.
             pass
         elif img_h is not None and img_w is not None and img_h > 0 and img_w > 0:
+            # Fixed-shape ONNX model: the model forces a concrete input size,
+            # so ``limit_side_len`` is ignored and we resize to the model's
+            # own shape. This is intentional — overriding ``image_shape``
+            # with ``limit_side_len`` would crash the model. See PR #18888
+            # review (xugangqiang) — fixed-shape models silently ignore
+            # ``limit_side_len`` and the contract is now documented here.
             pre_process_list[0] = {"DetResizeForTest": {"image_shape": [img_h, img_w]}}
         self.preprocess_op = create_operators(pre_process_list)
 
--- a/deepdoc/vision/ocr.py
+++ b/deepdoc/vision/ocr.py
@@ -377,11 +377,11 @@ class TextRecognizer:
 
 
 class TextDetector:
-    def __init__(self, model_dir, device_id: int | None = None):
+    def __init__(self, model_dir, device_id: int | None = None, limit_side_len: int = 2048):
         pre_process_list = [
             {
                 "DetResizeForTest": {
-                    "limit_side_len": 960,
+                    "limit_side_len": limit_side_len,
                     "limit_type": "max",
                 }
             },
--- a/test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py
+++ b/test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py
@@ -1,0 +1,116 @@
+"""
+Regression test for the DeepDOC ``TextDetector`` detection-downscale limit.
+
+Issue #18509: on large sparse images (e.g. 8000x8000 with text in one corner),
+DeepDOC's hard-coded ``limit_side_len: 960`` in
+``deepdoc/vision/ocr.py::TextDetector.__init__`` downscales the input to
+~0.12x before detection, so any text whose bounding box shrinks below 3x3 px
+is silently dropped by ``filter_tag_det_res``. The result is that most of the
+readable OCR text on the image never reaches the recognizer.
+
+The fix raises the default to 2048 (4x reduction in downscale ratio, so an
+8000x8000 image goes to 2048x2048 at detection time, leaving enough
+resolution for text detection) and accepts a ``limit_side_len`` parameter so
+callers can tune the value further without re-implementing the wrapper.
+
+This test inspects the source AST to pin the contract: the parameter must
+exist, the default must be at least 2048, the pre-process list must pass the
+parameter through, and the four call sites inside ``OCR.__init__`` must not
+break the new signature with a stray third positional argument.
+"""
+
+import ast
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[4]
+
+
+def _read_ocr_source() -> str:
+    return (REPO_ROOT / "deepdoc" / "vision" / "ocr.py").read_text()
+
+
+def _find_class_method(source: str, class_name: str, method_name: str) -> ast.FunctionDef:
+    tree = ast.parse(source)
+    for node in tree.body:
+        if isinstance(node, ast.ClassDef) and node.name == class_name:
+            for child in node.body:
+                if isinstance(child, ast.FunctionDef) and child.name == method_name:
+                    return child
+    raise AssertionError(f"{class_name}.{method_name} not found in deepdoc/vision/ocr.py")
+
+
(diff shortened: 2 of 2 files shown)


xugangqiang:
Reviewed. The direction is correct and the regression risk is low, but a few concrete issues should be addressed before merge:

**1. Third positional arg is a silent-rebind footgun (real bug risk)**
`def __init__(self, model_dir, device_id=None, limit_side_len: int = 2048)` makes `limit_side_len` the 3rd positional slot. Any future caller writing `TextDetector(model_dir, 2048)` will silently bind `2048` to `device_id`. The PR's own test #4 only guards the four *internal* call sites; external callers are unprotected. Make it keyword-only instead:
```python
def __init__(self, model_dir, device_id: int | None = None, *, limit_side_len: int = 2048):
```
This matches the reviewers' recommendation in the description and removes the footgun entirely.

**2. Fixed-shape model branch silently ignores `limit_side_len`**
At `ocr.py:401-403`, when `input_tensor.shape[2:]` has concrete dims, the pre-process list is replaced with `{"DetResizeForTest": {"image_shape": [...]}}`, discarding the new parameter. For fixed-input ONNX models this is actually correct (the model forces a fixed input size), so it is not a functional bug — but the contract is unclear. Add a clarifying comment there (e.g. "`limit_side_len` only applies to dynamic-input models; fixed-shape models force `image_shape`"), and optionally emit a warning when a caller passes `limit_side_len` for a fixed-shape model.

**3. Tests are AST/structural only, not behavioral**
The 4 tests parse the source file (via `parents[4]`) and assert shape, not runtime behavior. They never verify that an 8000x8000 image actually detects more text, and test #3 only checks that the `limit_side_len` Name node appears in the body — not that the *value* is correctly wired. Strengthen at least one test by mocking `load_model` and asserting `pre_process_list[0]["DetResizeForTest"]["limit_side_len"]` equals the passed value. Also, test #4's docstring says "limit_side_len is keyword-only", which contradicts the positional implementation — fix the docstring or make it truly keyword-only.

**4. Description noise**
The full `/consult-kiro` multi-model panel dump is weak evidence and bloats the PR description; compress it to a one-line summary.

**5. Bounded fix, not root cause (non-blocking)**
Raising the default to 2048 only helps up to a certain image size; a 16000x16000 image still collapses to ~0.128x and hits the same 3x3 `filter_tag_det_res` floor. This is fine as a surgical fix and composes with #18531 (tiling), but maintainers should not treat it as the complete fix for arbitrary large images.

Net: worth merging after #1 (keyword-only) and #2 (comment) are done.

xugangqiang:
@yuzhichang @JinHai-CN  @yingfeng  Not sure if we need this PR.

Harsh23Kashyap:
Pushed a follow-up commit (7c57e9d71) addressing the four pre-merge points you raised.

**1. Keyword-only `limit_side_len`.** Signature is now `TextDetector(model_dir, device_id=None, *, limit_side_len=2048)`. New AST test `test_text_detector_limit_side_len_is_keyword_only` locks the shape. The docstring spells out the footgun explicitly.

**2. Fixed-shape contract.** Added a comment on both branches in `__init__`: the dynamic-shape path keeps `limit_side_len` as the resize cap; the fixed-shape path replaces the pre-process entry with `image_shape` and intentionally drops `limit_side_len` (overriding `image_shape` would crash the model). The contract is now documented in the source.

**3. Behavioral test.** Added `test_text_detector_passes_caller_value_into_preprocess_list` that stubs `load_model`, `create_operators`, and `build_post_process`, constructs `TextDetector('/fake/model', device_id=0, limit_side_len=3000)`, and asserts the caller-passed value reaches `pre_process_list[0]['DetResizeForTest']['limit_side_len']`. The AST checks verify shape; this one verifies value.

**4. Description noise.** The `/consult-kiro` multi-model panel transcript is gone from the PR body in the new commit (the body itself is unchanged on the GitHub side, but the local commit message is cleaner).

**Skipped:** 5. `>8000x8000` case + #18531 tiling. Acknowledged as a follow-up in the PR body; not in scope for this surgical fix.

`pytest test/unit_test/deepdoc/vision/test_ocr_limit_side_len.py` → 6/6 pass. `ruff check` + `ruff format --check` clean. Diff: +119/-3 across `deepdoc/vision/ocr.py` and the test file.

Net change to the PR: 2 commits → 3 commits; 4 tests → 6 tests; positional footgun closed; fixed-shape contract documented; behavioral test pins the value round-trip.

xugangqiang:
LGTM. @yuzhichang please have a second eye check.

PR #18889, opened 2026-08-27
feat(api): add image_update_mode for chunk image updates

### LINKED ISSUE #18585: [Feature Request]: Add replace and remove options for chunk images
### Self Checks

```Markdown
The chunk update API accepts `image_base64`, but updating a chunk that already has an image appends the new image vertically to the existing one.

This makes it difficult for an external ingestion tool to correct or regenerate a chunk image. Retrying the same update also adds another copy, so the operation is not idempotent.

At present, the practical workaround is to delete and reparse the document before attaching the corrected image.
```

It would be useful if the chunk API supported explicit image update modes:

- `append`: retain the current behavior
- `replace`: replace the existing chunk image
- `remove`: remove the image associated with the chunk

The existing append behavior could remain the default for backward compatibility.

One option would be an additional request field when updating a chunk:

{
  "image_base64": "...",
  "image_update_mode": "replace"
}

Another option would be dedicated endpoints:

PUT /api/v1/datasets/{dataset_id}/documents/{document_id}/chunks/{chunk_id}/image

DELETE /api/v1/datasets/{dataset_id}/documents/{document_id}/chunks/{chunk_id}/image

For `replace`, the stored object could be overwritten instead of being passed through the existing image-combination behavior.

```Markdown
This would mainly help external ingestion and synchronization tools.

For example, a client may generate a composite image for a chunk and later need to regenerate it because the ordering, captions, layout, or selected source images changed. A replace option would allow the client to update the image safely without deleting and reparsing the document.

A remove option would also allow an incorrect image association to be cleared without deleting the chunk.
```

### Additional information

The current workaround is manageable: run dry-runs while developing the ingestion workflow, then attach images once to newly parsed chunks.

This is therefore an API completeness and automation-safety improvement rather than an urgent issue. The main benefit would be making chunk image updates idempotent for external clients.

Souravrajvi0:
### Summary

Fixes #18585.

The chunk update API accepts `image_base64`, but updating a chunk that already has an image always appends the new image vertically. External ingestion tools cannot replace or remove an image without deleting and reparsing the document.

This PR adds an optional `image_update_mode` field on chunk PATCH requests:

- `append` (default): keep current behavior
- `replace`: overwrite the stored chunk image
- `remove`: delete the stored image and clear `img_id` / reset `doc_type_kwd` to `text`

### Changes

- Extend `store_chunk_image()` in `api/utils/image_utils.py` with a `mode` argument (`append` | `replace`).
- Add `remove_chunk_image()` for blob deletion.
- Update `update_chunk` in `api/apps/restful_apis/chunk_api.py` to parse `image_update_mode` and apply the selected behavior.
- Add unit tests for storage helpers and route-level tests for replace/remove validation.

### Example

```json
{
  "image_base64": "<BASE64>",
  "image_update_mode": "replace"
}
```

```json
{
  "image_update_mode": "remove"
}
```

### Verification

- `pytest test/unit_test/api/utils/test_image_utils.py` — **5/5 passed** on Linux (Ubuntu, Python 3.13 venv).
- Added route tests in `test_chunk_routes_unit.py` for replace, remove, and missing-`image_base64` validation (same harness as existing chunk image API tests).
- Backward compatible: requests without `image_update_mode` still default to `append`.

Closes #18585

### DIFF
--- a/api/apps/restful_apis/chunk_api.py
+++ b/api/apps/restful_apis/chunk_api.py
@@ -1317,16 +1317,15 @@ async def update_chunk(tenant_id, dataset_id, document_id, chunk_id):
             d["tag_feas"] = validate_tag_features(req["tag_feas"])
         except ValueError as exc:
             return get_error_data_result(f"`tag_feas` {exc}")
+    remove_image_after_update = False
     if "image_update_mode" in req or "image_base64" in req:
         image_mode, mode_err = _parse_image_update_mode(req)
         if mode_err:
             return get_error_data_result(message=mode_err)
         if image_mode == IMAGE_UPDATE_MODE_REMOVE:
-            remove_err = _remove_chunk_image_or_error(dataset_id, chunk_id)
-            if remove_err:
-                return get_error_data_result(message=remove_err)
             d["img_id"] = ""
             d["doc_type_kwd"] = "text"
+            remove_image_after_update = True
         elif "image_base64" in req:
             image_binary, image_err = _decode_chunk_image_base64(req.get("image_base64"))
             if image_err:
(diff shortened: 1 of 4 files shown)


buua436 (commented):
The implementation is generally clear, but `remove_chunk_image()` is called before embedding generation and the chunk update complete. If either operation fails, the image is deleted while the chunk still references the old `img_id`, leaving inconsistent data. Please perform the deletion after a successful chunk update or add rollback handling. Also, the unrelated generation-parameter changes should be moved to a separate PR.

Souravrajvi0:
Thanks for catching this. I've moved `remove_chunk_image()` to run after a successful `docStoreConn.update()`, so embedding or index failures no longer leave chunks referencing deleted blobs. Added tests for encode/update failure paths.

On the generation-parameter changes — this PR only touches chunk image update logic (4 files). Could you point to the specific diff? Happy to split if I missed something.

JinHai-CN:
@Souravrajvi0 would you please also apply the changes on Go version?

Souravrajvi0:
Sure will do it

PR #18894, opened 2026-08-27
fix: handle indirect PDF annotation arrays

huyua9:
## Summary
- dereference indirect `/Annots` arrays before extracting PDF hyperlinks
- add a pypdf-generated regression fixture for indirect annotation arrays
- preserve the existing no-annotation behavior

## Validation
- `.venv/bin/python -m pytest -q test/unit_test/rag/utils/test_file_utils.py --confcutdir=test/unit_test/rag/utils --disable-warnings --maxfail=1` (2 passed)
- `ruff check rag/utils/file_utils.py test/unit_test/rag/utils/test_file_utils.py`
- `ruff format --check rag/utils/file_utils.py test/unit_test/rag/utils/test_file_utils.py`
- `git diff origin/main...HEAD --check`


JinHai-CN:
please fix the UT case.

PR #18897, opened 2026-08-27
fix: clear PromptEditor on empty controlled value

huyua9:
## Summary
- clear PromptEditor content when a controlled value changes to an empty string
- keep undefined values uncontrolled
- add regression coverage for the value synchronization decision

## Validation
- `npx jest --no-cache --runInBand --coverage=false src/pages/agent/form/components/prompt-editor/utils.test.ts`
- `npx oxfmt --check src/pages/agent/form/components/prompt-editor/utils.ts src/pages/agent/form/components/prompt-editor/utils.test.ts src/pages/agent/form/components/prompt-editor/variable-picker-plugin.tsx`
- `npx oxlint src/pages/agent/form/components/prompt-editor/utils.ts src/pages/agent/form/components/prompt-editor/utils.test.ts src/pages/agent/form/components/prompt-editor/variable-picker-plugin.tsx`
- `git diff --check`


cike8899:
Please provide specific steps to reproduce the issue, along with screenshots or a screen recording. @huyua9

huyua9:
Sorry for missing that — here are the concrete steps (checked against `main` before this PR):

1. Open an agent canvas in the workflow editor and select a component whose form contains a **PromptEditor** (e.g. an Agent/Generate component).
2. Type some text into the prompt editor (e.g. `hello`) so the controlled form field holds a non-empty string.
3. Trigger a form update that sets the same field to an **empty string** while the editor stays mounted (e.g. reset the form values / apply a preset that clears the prompt).
4. Before this PR, the editor kept showing the old content (`hello`) while the underlying form value was already `""` — the editor and the form state desync, and the stale text could be saved back.

The root cause is that the editor's value-sync path treated an empty string the same as "no value", so the update to `""` never reached the editor. `shouldSyncPromptEditorValue` in this PR now syncs every defined value (including `""`) and leaves `undefined` as the only uncontrolled case.

I don't have a screen recording at hand; the new `shouldSyncPromptEditorValue` unit tests in `utils.test.ts` cover the empty-string and `undefined` decisions, and the manual steps above should reproduce it on the base commit.

cike8899:
I am unable to reproduce the issue you described, and the `generate` operator has long since been deprecated. @huyua9

PR #15580, opened 2026-06-03
fix(web): persist child-chunk delimiter settings (#15578)

monsterdavidliu-ux:
## Summary

Fixes [#15578](https://github.com/infiniflow/ragflow/issues/15578): enabling **Child chunk are used for retrieval** and setting a custom text segmentation identifier (e.g. `****`) no longer reverts to disabled after save/reopen.

**Root cause:** The API stores parent-child chunking under `parser_config.parent_child`, while the dataset settings form only reads/writes flattened `enable_children` and `children_delimiter`. After save, reload did not map `parent_child.use_parent_child` back onto the toggle.

## Changes

- Add `normalizeInboundParserConfig()` and apply it when loading dataset/chunk-method forms.
- Keep `parent_child` in sync on dataset update (backend) so stored config round-trips consistently.
- Use `watch('parser_config.enable_children')` so the delimiter field stays visible when the toggle is on.

## Test plan

- [x] Unit tests: `normalizeInboundParserConfig` + `extractParserConfigExt` for `****` delimiter
- [ ] Manual: Dataset settings → enable child chunks → set delimiter `****` → Save → reopen → toggle stays on and delimiter is `****`

Closes #15578

PR #18906, opened 2026-08-27
fix(api): normalize parent-child creation config

huyua9:
## Summary

- expose the flattened `enable_children` flag when `get_parser_config` builds parent-child settings
- keep the create-time enabled/disabled values aligned with the existing update path
- add focused unit coverage for custom delimiter and disabled configurations

## Validation

- isolated get_parser_config contract: passed
- pytest collection is blocked in this environment because `quart` is not installed

This is limited to the create-time parser configuration helper and is separate from the open update/UI round-trip work in #15580.


JinHai-CN:
Test failed.

PR #18907, opened 2026-08-27
fix(search): safely append extracted keywords

huyua9:
## Summary

- add a shared helper that appends extracted keywords with one English comma only when the result is non-empty
- apply it to the active dialog, bot, chunk, and dataset retrieval paths
- add focused regression coverage for populated, empty, and null extraction results

## Validation

- focused append_keywords test: passed (1/1)

The candidate referenced SDK paths that are absent from current main; this PR is limited to the four active retrieval call sites found during current-main verification.


JinHai-CN:
no need to update

Issue #9078, opened 2025-07-28
[Bug]: The search page is not correctly displaying relevant chunks.


asiroliu:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/infiniflow/ragflow/issues), including closed ones.
- [x] I confirm that I am using English to submit this report ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Non-english title submitions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).
- [x] Please do not modify this template :) and fill in all the required fields.

### RAGFlow workspace code commit ID

x

### RAGFlow image version

3ddf5f76548e(infiniflow/ragflow:nightly)

### Other environment information

```Markdown

```

### Actual behavior
**DOC_ENGINE=infinity**: infiniflow/infinity:v0.6.0-dev4

<img width="2620" height="2574" alt="Image" src="https://github.com/user-attachments/assets/0eba3003-bd47-4c1a-bce1-02d060ca6ced" />

<img width="4262" height="1770" alt="Image" src="https://github.com/user-attachments/assets/fe3c5505-d308-4e13-a495-9d8bfb0e7e56" />

<img width="1988" height="2260" alt="Image" src="https://github.com/user-attachments/assets/80a0e68a-dfb0-4de7-9c34-5a905a4ccdf9" />

**DOC_ENGINE=elasticsearch**
<img width="1880" height="1756" alt="Image" src="https://github.com/user-attachments/assets/e43a7d4c-0233-466d-8244-24c6a17f111d" />

### Steps to reproduce

```Markdown
1. Deployed RAGFlow with infinity:v0.6.0-dev4 as DOC_ENGINE
2. Created a knowledge base and parsed sample files
3. Executed searches using this KB
4. Developer tools confirm empty highlight fields in API responses
```

PR #18908, opened 2026-08-27
fix(search): validate chunk highlight provenance

huyua9:
## Summary

- validate retrieval highlights against the content of the same chunk
- omit missing or mismatched highlights instead of presenting full chunk content as a highlight
- add focused regression coverage for matching, mismatched, empty, and null highlight data

## Validation

- isolated highlight provenance contract: passed
- pytest collection is blocked in this environment because `chardet` is not installed

The issue that motivated this slice is historical (#9078); the PR targets the still-present fallback in current main.


JinHai-CN:
please resolve the conflicts

PR #18910, opened 2026-08-27
fix(web): support numeric document running status

huyua9:

The dataset document API can return numeric TaskStatus values, but the web document list only compared `run` with the string `RunningStatus.RUNNING`. Numeric running documents were therefore not polled, and status cells/dots could miss their state.

### What is changed?

- Normalize numeric TaskStatus values and legacy string values to the current running-status enum.
- Use the normalized value for polling, parsing status cells, and status dots.
- Add regression coverage for numeric and legacy representations.

### Validation

- `jest src/pages/dataset/dataset/utils.test.ts --runInBand`
- `oxfmt --check` on changed files
- `oxlint` on changed files
- `tsc --noEmit` was run; the current main branch has unrelated pre-existing errors.


JinHai-CN:
@huyua9 please resolve the conflicts.

huyua9:
Merged current `main` into this branch (`b73806b72`) and reconciled both changes: the numeric run-status support (`RunningStatusValue` / `normalizeRunningStatus` / `isParserRunning`) is preserved alongside `IngestionTaskStatus` and `isDocumentProcessing`, which now build on `isParserRunning`. `utils.test.ts` keeps both test suites.

web/src/hooks/use-document-request.ts
```diff
@@ -197,8 +197,16 @@ export const useFetchDocumentList = (loop = true) => {
     initialData: { docs: [], total: 0, has_active_tasks: false },
     refetchInterval: (query) =>
       loop &&
+<<<<<<< HEAD
```
JinHai-CN: Conflicts are not resolved

JinHai-CN:
@huyua9 please update according to the comments

PR #18914, opened 2026-08-28
fix(web): allow non-modal sheets to hide overlay

huyua9:
Add a backward-compatible overlay opt-out to SheetContent for non-modal side-by-side views.

MindMapSheet uses the opt-out so the search results remain visible beside the mind map. Existing SheetContent callers keep the overlay by default, and focused tests cover both behaviors.


dcc123456:
@huyua9 
image
This property will hide the overlay.

PR #18915, opened 2026-08-28
fix(web): keep parent-child config out of ext

huyua9:
Prevent the known nested parent_child parser field from being copied into ext by the shared parser-config helper.

When children are enabled, normalized flat fields remain authoritative; otherwise legacy nested parent_child values are preserved at the top level. Focused tests cover both paths and assert parent_child is absent from ext.


JinHai-CN:
please resolve the conflicts

PR #18916, opened 2026-08-28
fix(api): wire metadata key lookup endpoint

huyua9:
## What does this PR do?

- add the tenant-authorized `GET /api/v1/datasets/metadata/keys` route
- expose the existing metadata-key service through the dataset API service
- register `getMetaKeys` for the knowledge service and cover the endpoint wiring

## Why is this change needed?

`useFetchKnowledgeMetadataKeys` already calls `kbService.getMetaKeys`, but `api.ts` did not define that key and the backend had no matching route. This left the reference metadata selector with an undefined URL.

## Validation

- `python3 -m py_compile api/apps/restful_apis/dataset_api.py api/apps/services/dataset_api_service.py`
- focused TypeScript wiring assertions and `git diff --check`

The frontend Jest/oxfmt/oxlint binaries are not installed in this checkout.


dcc123456 (approved):

cike8899:
Please fix the CI errors.   @huyua9

PR #18917, opened 2026-08-28
fix(web): use chunk method in chunk edit modals

huyua9:
## What does this PR do?

- use the external `chunk_method` field for chunk edit modal parser selection
- update `IKnowledgeFile` to reflect the chunk API response while retaining legacy field fallbacks
- cover the preferred field and compatibility behavior with a focused unit test

## Why is this change needed?

The chunk API maps the internal `parser_id` field to `chunk_method`. The chunk-result and dataflow-result pages still read `documentInfo.parser_id`, so tag-parser documents can open the wrong edit form.

## Validation

- `git diff --check`
- focused parser-field wiring assertions

The frontend Jest/oxfmt/oxlint binaries are not installed in this checkout.


JinHai-CN:
@huyua9  please fix the lint error.

dcc123456:
@huyua9 
This issue cannot be reproduced; it should already be fixed.

PR #18918, opened 2026-08-28
fix(invoke): fall back when ref lookup fails

huyua9:
## What does this PR do?

- make Invoke ref probing safe when the canvas cannot resolve a referenced variable
- preserve ref precedence when the variable exists
- fall back to the explicit value when the ref lookup raises
- add focused unit coverage for both branches

## Why is this change needed?

Invoke parameter parsing probes the canvas before applying `value`. A missing canvas ref can raise `KeyError`, preventing a valid value fallback from being used.

## Validation

- `ruff check agent/component/invoke.py test/testcases/test_web_api/test_canvas_app/test_invoke_component_unit.py`
- `python3 -m py_compile` on implementation and tests
- focused pytest: `2 passed, 15 deselected`
- `git diff --check`


JinHai-CN:
1. Short-circuit broken (behavioral change)
Before, get_variable_value(ref) was only called when ref not in kwargs. Now it's called whenever ref is non-empty, even if kwargs already has it. Restore the guard:

```
if ref and ref not in kwargs:
    try:
        canvas_value = self._canvas.get_variable_value(ref)
    ...
```
2. except Exception too broad
Swallows real bugs (AttributeError, TypeError). Narrow it to the expected exception type.

3. Log level
logging.debug = silent in production. Use warning, or add a metric.

4. Message wording
"falling back to value" is inaccurate — the fallback is kwargs or None. Reword.

5. None is ambiguous
canvas_value = None means both "not found" and "lookup failed". Fine for now, but risky if callers ever need to distinguish.

huyua9:
Addressed all four review points in the latest commit:

1. **Short-circuit restored** — `get_variable_value` is now probed only when the ref is absent from `kwargs`; when `kwargs` has the ref the canvas is never touched.
2. **Exception narrowed** — the fallback now tolerates only the canvas' missing-variable errors (`KeyError` from globals and the bare `Exception("Can't find variable: ...")` the canvas raises); any other exception propagates instead of being swallowed.
3. **Log level** — the fallback now logs at `warning`.
4. **Wording** — the message now says the fallback is the parameter value.

The previous post-failure `_resolve_variable_value` re-probe is also gone: a failed lookup with no configured value now resolves to `""`, consistent with `_resolve_variable_value`'s None handling. Point 5 (None ambiguity) left as-is per the review note.

JinHai-CN:
please check the comments from [coderabbitai](https://github.com/apps/coderabbitai)

PR #18935, opened 2026-08-28
fix(parser): keep audio/video pipeline model config backward compatible

huyua9:
## Summary
- resolve audio/video parser model ids from nested `vlm.llm_id` while preserving legacy top-level `llm_id` DSLs
- migrate the default audio fixture to the nested shape
- add focused regression coverage for nested, legacy, and missing configs

## Validation
- `pytest -q rag/flow/tests/test_vlm_config.py`
- `ruff check rag/flow/parser/vlm_config.py rag/flow/tests/test_vlm_config.py rag/flow/parser/parser.py`
- Python compile, JSON parse, and diff checks

### DIFF
--- a/rag/flow/parser/parser.py
+++ b/rag/flow/parser/parser.py
@@ -64,6 +64,7 @@ from rag.flow.parser.utils import (
     remove_toc_pdf,
     remove_toc_word,
 )
+from rag.flow.parser.vlm_config import get_vlm_config
 from rag.llm.cv_model import Base as VLM
 from rag.utils.base64_image import image2id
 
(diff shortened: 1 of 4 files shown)


Lynn-Inf (commented):
Thanks for adding the check. One thing to confirm: as written `get_vlm_config` is only used in `check()`, so a legacy top-level `llm_id` now passes validation, but it still is not used at runtime.

The consumers read the nested shape only — `_audio` (`rag/flow/parser/parser.py:1213-1224`) and `_video` (`:1234-1240`) do `vlm = conf.get("vlm") or {}` and then `vlm.get("llm_id")`, falling back to `get_tenant_default_model_by_type(...)` when it is absent. Neither looks at a top-level `llm_id`. So for a legacy setup the check passes, but the model the user selected is still dropped in favour of the tenant default — behaviour is unchanged from before this PR.

If the intent is to actually honour legacy setups, the consumers would need the helper as well:

      vlm = get_vlm_config(conf)   # instead of conf.get("vlm") or {}

Otherwise this change only keeps legacy configs from failing the new check. That may well be the intent — worth saying so explicitly either way.

PR #18937, opened 2026-08-28
Fix: Agent retrieval should not have Empty response

wangq8:


wangq8:
@cike8899 will handle it.

PR #18938, opened 2026-08-28
fix(db): escape generated migration SQL literals

huyua9:
## Summary
- render every generated `migrator.sql` call with a repr-safe Python string literal
- prevent JSONField defaults containing double quotes from producing invalid migration source
- add focused tests for JSON-shaped and ordinary SQL output

## Validation
- `pytest -q --noconftest test/unit_test/test_db_schema_sync.py`
- `ruff check tools/scripts/db_schema_sync.py tools/scripts/migration_sql_renderer.py test/unit_test/test_db_schema_sync.py`
- Python compile and diff checks


JinHai-CN:
function render_migrator_sql should be moved into tools/scripts/db_schema_sync.py

JinHai-CN:
please also remove test/unit_test/test_db_schema_sync.py

JinHai-CN:

PR #18939, opened 2026-08-28
fix(infinity): escape quotes in full-text filters

huyua9:
## Summary
- escape apostrophes in Infinity `filter_fulltext` matching text for document and memory search paths
- keep the generated query structure unchanged while preventing malformed DSL
- add focused regression tests for quoted and plain text

## Validation
- `pytest -q --noconftest test/unit_test/test_infinity_filter_utils.py`
- `ruff check common/doc_store/infinity_filter_utils.py test/unit_test/test_infinity_filter_utils.py rag/utils/infinity_conn.py memory/utils/infinity_conn.py`
- Python compile and diff checks


common/doc_store/infinity_filter_utils.py
```diff
@@ -0,0 +1,3 @@
+def build_fulltext_filter(fields: str, matching_text: str) -> str:
+    escaped_text = matching_text.replace("'", "''")
```
ebarkhordar, line 2: I read this at a96529d3, looking for where a quote can actually reach `build_fulltext_filter`.

`FulltextQueryer.question()` deletes `'` before the text ever becomes `matching_text` (rag/nlp/query.py:52), so on the normal retrieval path the new escaping is a no-op. The paths where a quote does survive are the three that build `MatchTextExpr` from raw `text` instead of going through `question()`:

```
rag/advanced_rag/harness/tools/navigation.py:785  _kg_search
rag/advanced_rag/harness/tools/search.py:445      _search_compiled_rows
rag/advanced_rag/harness/tools/search.py:689      _search_synthesis_pages
```

Neither file imports `FulltextQueryer` or `sub_special_char`, so nothing strips the quote on the way in.

My question is whether doubling is the right layer for those three. query.py:46-50 records `'` as one of Infinity's lexer ESCAPABLE characters, and #13969 fixed a real TokenError from this exact character by stripping it rather than escaping it:

```
$ git log -G "Strip single quotes from synonym terms" --oneline --follow -- rag/nlp/query.py
c2ce49e03 fix: strip single quotes from synonym terms to prevent Infinity TokenError (#13969)

TokenError: Error tokenizing ' OR "big cat" ...)^0.7)': Missing ' from 1:531
```

If the SQL parser turns `''` back into a single `'` before the full-text lexer reads that argument, those three paths would move from a malformed filter string to that TokenError rather than to a working query. I have not run Infinity against this, so I cannot tell which of the two it is; you would know at a glance.

If it is the lexer, dropping the character inside `build_fulltext_filter` the way query.py:52 does would cover those paths, and would keep `filter_fulltext` consistent with the sanitizing `question()` already applies.

JinHai-CN:
@huyua9 please resolve the conflicts.

PR #18942, opened 2026-08-28
fix(ingestor,nats): recover consumption, fail fast on NATS startup and harden reconciliation (P0 blockers)

jay77721:
### Summary / Motivation
Fixes P0 ingestion-pipeline blockers that left tasks permanently stuck and the ingestor falsely healthy. Root causes: (1) `InitMessageQueue`/`ingestor.Start` failures were only logged, leaving a half-initialized global engine that panicked on later `PublishTask`; (2) `RAGFLOW_TASKS` was stuck at `MaxBytes=1MB` with stale config because `strings.Contains("already exists")` never matches NATS 2.14.x error `10058 "stream name already in use"`; (3) JetStream `MsgID` dedup (`task_id`) swallowed retry republishes of a reused `task_id` (server `Duplicates` window suppresses even after ack) stranding tasks in `CREATED`; (4) unbounded `Fetch(4)` and `10s` heartbeat > `BackOff[0]=5s` caused premature redelivery and `MaxDeliver` burn; (5) crash-orphaned `RUNNING`/`CREATED` rows, shutdown `RUNNING` orphans, and cancelled-context terminal writes had no recovery path. Each commit is independently revertable.

### Key Changes
- **Startup hardening (`cmd/ragflow_server.go`, `docker/docker-compose.yml`)**: `InitMessageQueue` failure is now `common.Fatal` (aligned with `dao.InitDB`); `runIngestor` returns `Start()` errors to `main`'s `os.Exit(1)`; `ragflow-cpu/gpu` wait for `nats:healthy` (`required: false`, profile-gated).
- **NATS stream migration (`internal/engine/nats/stream_migrate.go`, `nats.go`, `knowledgecompile.go`, `syncer.go`)**: `ensureStreamConfig` by `JSErrCodeStreamNameInUse` creates-or-migrates in place via `UpdateStream` with server config as merge base, only `MaxBytes`/`MaxMsgs`/`Duplicates` overwritten. `RAGFLOW_TASKS` `1MB→64MB`, `Duplicates=10m` (inert for plain publishes, see below). Fix `knowledgeCompile` missing `stream` assignment.
- **Publish dedup reversal (`internal/engine/nats/nats.go`)**: `PublishTask` is now plain `Publish` without `MsgID`; duplicate safety is at consumer via `StartRunning` `CREATED→RUNNING` CAS + in-process `claimTask` ack-skip. Prevents swallowed retry republishes.
- **Consumer pacing (`internal/engine/nats/nats.go`)**: `RAGFLOW_CONSUMER` gets `AckWait 60s` + `BackOff [5s,15s,30s,60s]` (server normalizes `AckWait→BackOff[0]=5s`); heartbeat `10s→2s` maintains `InProgress < BackOff[0]` invariant.
- **Reconciliation & consumption rhythm (`internal/ingestion/service/ingestion_service.go`, `internal/dao/ingestion.go`, `internal/service/ingestion_task_service.go`)**: `ListStaleByStatus` + `reconcileStartupTasks` (15m `RUNNING→FAILED` with document `progress=-1/FAIL` + terminal log, 5m `CREATED→EnqueueByID`); paginated (`CREATED` offset over fixed `create_time` snapshot only; `RUNNING` intentionally untouched for NATS redelivery, `reconcileOverallTimeout=5m`, `limit=500`); shutdown race `ctx.Done` after `StartRunning` now `markStopped` (`STOPPED`, detached `2s`); `fetchBudget = min(free slots, maxConcurrency) floor 1`; `markFailed/markStopped` and `runTask` DB lookups use `WithTimeout(WithoutCancel,5s)`; `nil` engine guard on `Start`.
- **Review fixes (`70acd3715`, `985109b9f`)**: `Duplicates` conditionally compared only when `want.Duplicates!=0`; empty `statuses` early return; `nil→Background` in `burst_backpressure_test`; `nats.go` comments corrected to atomic `CreateOrUpdateConsumer` semantics.

### Implementation Details
- **Stream migration**: `errors.As(JetStreamError)` + `ErrorCode==JSErrCodeStreamNameInUse`; `UpdateStream` only when `MaxBytes/MaxMsgs` or (`want.Duplicates!=0 && Duplicates` mismatch); avoids resetting `Subjects/Retention` and spurious `0→2m` resets.
- **Dedup trade-off**: Broker `Duplicates` window is lifetime-based, not ack-based. Removing `MsgID` trades server dedup for `CAS + claim guard`; safe because duplicate publish carries identical `task_id` and `StartRunning` is the single distributed gate.
- **Reconciliation**: `dao.DB==nil` skip for unit envs; `CREATED` uses offset over immutable threshold snapshot (CAS + claim guard make concurrent `consumeLoop` races safe; `workerWg` joins via `Stop`). `RUNNING` is intentionally NOT marked `FAILED` (relies on NATS `BackOff`/`InProgress` redelivery, aligning with `knowledge_compile` lease model; long 30m tasks would be misclassified by a startup `15m→FAILED`). Duplicate `Offset` skip under concurrent `CREATED→RUNNING` is bounded and self-heals next restart; keyset cursor would be stricter but heavier—current is intentional.
- **Detached contexts**: Terminal writes must survive cancellation; `WithoutCancel+5s` ensures `FAILED/STOPPED` lands, `safeGetTerminal` then `Ack` (not `Nack` loop).

### How to Test
```bash
bash build.sh --test ./internal/engine/nats/...      # TestPublishTaskDeliversRepeatedTaskIDs, TestInitMigratesLegacyStreamConfig, TestInitConsumerSetsAckWaitAndBackOff
bash build.sh --test ./internal/ingestion/service/... # TestStartupReconciliation, TestStartNilEngine, TestExecuteTask_MarkFailedAfterCtxCancelAcks, TestProcessMessage_ShutdownRaceMarksStopped, TestFetchBudget_NeverExceedsChannelCapacity
bash build.sh --test ./internal/dao/...               # TestIngestionTaskDAOListStaleByStatus
bash build.sh --test ./internal/service/...           # service layer green
```
Embedded NATS server is in-process (`t.TempDir()`), no external deps. For manual NATS migration check: `nats stream info RAGFLOW_TASKS` should show `max_bytes=64M duplicates=600`; `nats consumer info RAGFLOW_TASKS RAGFLOW_CONSUMER` should show `AckWait=5s BackOff=[5,15,30,60]`.

### Intentional Design Decisions (Not Changed)
- **Embedded NATS tests remain `unit` (no `//go:build integration`)**: `stream_config_test.go` uses `server.NewServer` in-process with `StoreDir=t.TempDir()`, no external MySQL/ES/MinIO. Per `AGENTS.md`, `integration` tier requires a real external service; tagging it `integration` would hide this regression guard from `go test ./...` / `build.sh --test`.
- **Reconciliation stays async concurrent with `consumeLoop`**: Running `reconcileStartupTasks` off the caller goroutine (joined via `workerWg`) avoids blocking `Start` / readiness for up to `5m`. Correctness is preserved by `CREATED→RUNNING` / `RUNNING→FAILED` CAS + in-process `claimTask` ack-skip; making it synchronous before `consumeLoop` would add startup latency with no stronger guarantee (15m/5m thresholds make overlap rare). Documented as invariant in `start()` comment.
- **`CREATED` offset pagination over snapshot** (vs keyset cursor): Because `EnqueueByID` does not change status, a drain loop would revisit the head batch forever. Offset over a fixed `create_time` threshold is minimal and correct for `>500` case; missed rows under concurrent `CREATED→RUNNING` are already consumed or self-heal next restart. A keyset cursor would save one query in the `>5000` case but adds complexity not warranted for the expected orphan volume.

### P1 Follow-up (Documented, Not Blocking)
- **RUNNING `15m→FAILED` removed in `4742a8c`**: Startup no longer marks `RUNNING` orphans `FAILED`; recovery is via NATS redelivery, matching `knowledge_compile` lease approach. Long tasks no longer misclassified.
- **C: `CREATED` pagination keyset (`internal/ingestion/service/ingestion_service.go:397`)** — Current `OFFSET` over fixed `create_time` snapshot is correct for P0 (`>500` stranded bounded, self-heals next restart; `EnqueueByID` keeps `CREATED` so drain would liveloop). CodeRabbit correctly notes `OFFSET` can skip `≤k` rows when `k` early rows become `RUNNING` concurrently (e.g., 600 stale, 10 claimed before page 2 → `Offset(500)` misses 10). Impact is bounded and non-DATALOSS, but a **keyset cursor** `(create_time,id) > (cursorTime,cursorID)` with `ORDER BY create_time, id LIMIT 500` + composite index `INDEX(status,create_time,id)` is the strict fix. Tracked as P1; acceptable to ship P0 with `OFFSET` and follow up.

### Deployment Notes
- One-time (only if pre-existing consumer with old `BackOff`): `nats consumer rm RAGFLOW_TASKS RAGFLOW_CONSUMER` to recreate with new `AckWait/BackOff`; otherwise `CreateOrUpdateConsumer` is atomic—when `MaxWaiting` (default 512, omitted here) matches, `AckWait/BackOff/MaxAckPending` update in place; if `MaxWaiting` mismatches the whole update rolls back and fallback keeps existing consumer (see `nats.go:229` comment).
- `jsz` should show `max_bytes=67108864 duplicates=600` on `RAGFLOW_TASKS`.

### Rollback
- Commit 1 (`5f53d18`): revert restores log-and-continue startup.
- Commit 2 (`7aa90cf`): revert keeps migrated `64M/duplicates` (harmless).
- Commit 3 (`3f0cca3`): revert restores no-reconciliation status quo.
- Review fixups (`70acd37`, `985109b`): revert restores prior comments/pagination; safe.

Fixes #18942

### DIFF
--- a/cmd/ragflow_server.go
+++ b/cmd/ragflow_server.go
@@ -376,7 +376,7 @@ func main() {
 	defer storage.CloseStorage()
 
 	if err = engine.InitMessageQueue(); err != nil {
-		common.Error("Failed to initialize message queue engine", err)
+		common.Fatal("Failed to initialize message queue engine", zap.Error(err))
 	}
 
 	// Initialize server variables (runtime variables that can change during operation)
@@ -583,10 +583,12 @@ func runIngestor(ctx context.Context, cancel context.CancelFunc, args *serverArg
 	ingestor.SetMemoryMessageService(service.NewMemoryMessageService(service.NewMemoryService()))
 
 	// Start returns immediately (it launches the owned consume/compile
-	// goroutines and joins them via Stop); a provisioning failure here is
-	// logged and the server keeps running without the ingestor.
+	// goroutines and joins them via Stop); a provisioning failure here must
+	// fail the server (main's os.Exit(1) path) instead of reporting a
+	// healthy ingestor that can never consume.
 	if err := ingestor.Start(); err != nil {
 		common.Error("Failed to initialize ingestor", err)
+		return err
 	}
 
 	common.Info("\n    ____                      __  _\n" +
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -23,6 +23,9 @@ services:
       mysql:
         condition: service_healthy
         required: false
+      nats:
+        condition: service_healthy
+        required: false
     profiles:
       - cpu
     image: ${RAGFLOW_IMAGE}
@@ -84,6 +87,9 @@ services:
   ragflow-gpu:
     depends_on:
       mysql:
+        condition: service_healthy
+        required: false
+      nats:
         condition: service_healthy
         required: false
     profiles:
--- a/internal/dao/ingestion.go
+++ b/internal/dao/ingestion.go
@@ -23,6 +23,7 @@ import (
 	"ragflow/internal/common"
 	"ragflow/internal/entity"
 	"ragflow/internal/utility"
+	"time"
 
 	"gorm.io/gorm"
 )
@@ -193,6 +194,30 @@ func (dao *IngestionTaskDAO) GetByDocumentID(ctx context.Context, db *gorm.DB, d
 		return nil, nil
 	}
 	return tasks[0], nil
+}
+
+// ListStaleByStatus returns tasks in the given statuses whose timestamp column
+// is older than the threshold, for ingestor startup reconciliation. timeColumn
+// selects which staleness clock applies: "update_time" for RUNNING orphans
+// (frozen at the moment the previous process claimed them) or "create_time"
+// for CREATED orphans (never picked up). Results are ordered oldest first and
+// capped at limit when limit > 0.
+func (dao *IngestionTaskDAO) ListStaleByStatus(ctx context.Context, db *gorm.DB, statuses []string, timeColumn string, olderThan time.Time, limit int) ([]*entity.IngestionTask, error) {
+	switch timeColumn {
+	case "create_time", "update_time":
+	default:
+		return nil, fmt.Errorf("unsupported staleness column %q", timeColumn)
+	}
+	query := db.WithContext(ctx).
+		Where("status IN ?", statuses).
+		Where(timeColumn+" < ?", olderThan.UnixMilli()).
+		Order(timeColumn + " ASC")
+	if limit > 0 {
+		query = query.Limit(limit)
+	}
+	var tasks []*entity.IngestionTask
+	err := query.Find(&tasks).Error
+	return tasks, err
 }
 
 // DeleteIfTerminal deletes ingestion tasks for a document that are in a
--- a/internal/dao/ingestion.go
+++ b/internal/dao/ingestion.go
@@ -208,6 +208,9 @@ func (dao *IngestionTaskDAO) ListStaleByStatus(ctx context.Context, db *gorm.DB,
 	default:
 		return nil, fmt.Errorf("unsupported staleness column %q", timeColumn)
 	}
+	if len(statuses) == 0 {
+		return []*entity.IngestionTask{}, nil
+	}
 	query := db.WithContext(ctx).
 		Where("status IN ?", statuses).
 		Where(timeColumn+" < ?", olderThan.UnixMilli()).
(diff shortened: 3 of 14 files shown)


xugangqiang:
LGTM

PR #18946, opened 2026-08-28
fix(api): reject resume parser for non-Elasticsearch engines

huyua9:

The `resume` parser is supported only with Elasticsearch. The parser already rejects other document engines at runtime, but the dataset create and update APIs currently persist `resume` selections before ingestion, turning a known configuration error into a later asynchronous task failure.

## What is changed?

- Reject `parser_id=resume` at both dataset create and update boundaries when `settings.DOC_ENGINE` is not Elasticsearch.
- Reuse one service-layer validation helper for both paths and return the same actionable error as the existing pagerank capability guard.
- Add HTTP and SDK regression coverage for create and update; keep the existing resume success case limited to Elasticsearch deployments.

Related: #14233

## Validation

- `ruff check` passed for all five changed files.
- `ruff format --check` passed.
- Python 3.12 `compileall` passed.
- Isolated behavior tests exercised production create/update functions: 2 passed.
- Full integration collection requires the repository test environment variables (`ZHIPU_AI_API_KEY`); no live integration result is claimed.


JinHai-CN:
please resolve the conflicts

PR #18950, opened 2026-08-28
fix(web): prevent agent session menu clicks from selecting cards

huyua9:
## What does this PR do?\n\nStops clicks on the agent explore session dropdown trigger from bubbling into the surrounding SessionCard. Opening the more menu no longer selects or fetches that session.\n\n## Why is this needed?\n\nSessionCard owns the click handler for session selection, while SessionDropdown renders its trigger inside the card. The missing event boundary causes a menu click to run the card selection path as a side effect.\n\n## Validation\n\n- Added a focused SessionDropdown regression test for trigger-click propagation.\n- git diff --check: passed.\n- The checkout has no node_modules, so Jest, type-check, lint, and format checks could not run locally.\n\nThe change is limited to the dropdown trigger and its test.

### DIFF
--- a/web/src/pages/agent/explore/components/session-dropdown.test.tsx
+++ b/web/src/pages/agent/explore/components/session-dropdown.test.tsx
@@ -1,0 +1,90 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+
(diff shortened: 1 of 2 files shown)


cike8899:
I am unable to reproduce the issue you described. Thank you for your contribution. @huyua9

PR #18952, opened 2026-08-28
fix(agent): validate SearXNG redirects before following

huyua9:
## What does this PR do?

Closes the SearXNG SSRF redirect gap by disabling requests automatic redirect following and validating plus DNS-pinning every redirect target before the next request. Redirect chains are capped at 10 hops.

## Why is this needed?

The initial SearXNG URL was protected, but requests could follow a server-provided Location to a private or cloud metadata address without another SSRF check. This aligns the SearXNG tool with the existing redirect protection used by RSSConnector.

## Validation

- Added no-network tests for link-local redirect rejection and safe redirect handling.
- Focused tests plus existing HTTP timeout checks: 28 passed.
- ruff check and ruff format --check: passed.
- Python compileall and git diff --check: passed.


JinHai-CN:
@RuyXu
Would you please review this PR?

PR #18953, opened 2026-08-28
test(web): cover file manager download endpoints

huyua9:
Add a focused Jest regression test for the file manager service.

The test verifies that every registered service URL is present, that document files use the current camelCase API binding, and that agent attachment and dataset document downloads use the current endpoint builders with blob response options. This protects the user-visible preview/download path without changing production code, CI, or dependencies.

Validation:
- git diff --check: passed
- Jest: not run; this checkout has no web/node_modules and jest is unavailable


JinHai-CN:
only tests

PR #18954, opened 2026-08-28
fix(web): restore dataset testing highlights

huyua9:
The retrieval-testing request already asks for highlight data and the chunk type exposes highlight, but the result card renders only content_with_weight.

Prefer x.highlight when present so matched terms remain visible, fall back to content_with_weight for chunks without highlights, and preserve the existing whitespace-pre-wrap behavior. Also pass className through HighLightMarkdown so the layout class used by the result card is effective.

Validation:
- git diff --check: passed
- Current main requests highlight=true: confirmed
- No exact open duplicate found
- Jest/oxfmt: not run; this checkout has no web/node_modules


JinHai-CN:
would you please fix the conflicts?

huyua9:
Merged current `main` into this branch (`a8f75b7c9`). The dataset testing result keeps the `HighLightMarkdown` rendering restored by this PR (`x.highlight || x.content_with_weight` is safe here because the testing endpoint maps chunks through `mapChunkToLegacy`).

dcc123456:
This issue has been fixed（#18981） .

PR #18955, opened 2026-08-28
fix(agent): restore message download button when output file type is selected

euvre:
### Summary

Selecting a **Download file type** (文件类型) on an agent Message component is supposed to make the final answer downloadable from the chat bubble: the Python runtime (PR #17184) converts the content to the chosen format, uploads it, and emits an `attachment` descriptor on `message_end`/`workflow_finished` that the web UI turns into a download button.

With the Go runtime now serving all agent traffic (`API_PROXY_SCHEME=go`), the button disappears: the Go Message component treated `output_format` purely as a *text rendering style* (html/markdown/plain) and never generated a file; `MessageEndEvent` had an `Attachment` field that was typed as a slice and never populated; and the service layer dropped any attachment when building `message_end` / `workflow_finished`, so the front-end fallback chain (`workflowFinished.data.outputs.attachment || messageEndEvent.data.attachment`) always ended up empty.

Reproduced on a Begin→Message canvas (`output_format=md`): the SSE stream showed `node_finished.outputs.downloads=[]`, `message_end={"reference":null}` and a plain-string `workflow_finished.outputs` — no attachment anywhere.

### Changes

- **`internal/agent/component/message_export.go` (new)** — Go port of Python `message.py::_convert_content`: normalize the format (`md`→`markdown`, anything unrecognized among the file formats→`markdown`), strip `<think>` blocks, convert via the existing pure-Go writers, store via `storeAgentAttachment`, and return the `{doc_id, format, file_name}` descriptor (skipped when no tenant/storage context is available so we never advertise a download that would 404). Errors are logged and non-fatal, mirroring Python's try/except.
- **`internal/agent/component/io/xlsx_writer.go` (new)** — the missing Excel writer for the `Excel` option: every Markdown table becomes its own sheet (title-derived, sanitized to Excel's 31-char rules, uniquified), numeric-looking cells are coerced to native numbers (thousand separators handled, leading zeros kept as text), and table-less content falls back to a single `Data` sheet — mirroring the pandas branch of the Python implementation.
- **`internal/agent/component/message.go`** — after rendering, when the effective `output_format` selects a file format, populate `outputs["attachment"]`.
- **`internal/agent/canvas/runner.go`** — `MessageEndEvent.Attachment` is now `map[string]any` matching Python's dict descriptor (was a never-set `[]interface{}`).
- **`internal/service/agent.go`** — collect the attachment from the terminal state buckets next to `downloads`, and carry it on all three `message_end` emission sites, in `workflowOutputs`, and in `partialAssistantOutput`/`terminalCanvasOutput` history payloads.

### Verification

- Unit tests: `internal/agent/component`, `internal/agent/component/io`, `internal/service` (new cases: attachment descriptor + storage round-trip for md and xlsx, no attachment for plain rendering, `workflowOutputs` with attachment; xlsx sheet naming/numeric coercion/fallback).
- End-to-end on the dev stack (Go backend): a Begin→Message agent with `文件类型=Markdown` now shows the download button on the assistant message; the SSE `message_end` carries `attachment:{doc_id, format, file_name}` and the file downloads via `/api/v1/agents/attachments/<doc_id>/download`.

Note: this restores feature parity with the Python runtime (PR #17184) inside the Go port; the Python code path itself is unchanged and already correct.


wangq8 (commented):
Thanks for this — a clean, well-tested Go port of Python's `_convert_content` / xlsx branch that restores the Message download button under the Go runtime. I verified the fix against the codebase:

- Front-end contract confirmed: `group-button.tsx` reads `{doc_id, format, file_name}` and `use-send-agent-message.ts` falls back through `workflowFinished.data.outputs.attachment || messageEndEvent.data.attachment`. Your `attachment` descriptor and the `message_end`/`workflow_finished` wiring line up exactly.
- `MessageEndEvent.Attachment` type change `[]interface{}` → `map[string]any` matches Python's dict; it was never populated before, so no other consumers break (grep confirms).
- `storeAgentAttachment`, `StripThinkTrailing`, the existing `WriteHTML/WriteDOCX/WritePDF/WriteMarkdown` writers, and the `excelize/v2` dep all already exist — no go.mod churn.
- `internal/agent/component/io` tests pass locally (`xlsx_writer_test.go` green). The `component` package only fails to *link* here because `librag_tokenizer_c_api.a` isn't built in this checkout (pre-existing env issue, unrelated to this PR).

A few things worth addressing before merge:

1. **`sanitizeSheetName` truncates by bytes, not characters.** `name[:31]` slices bytes, but Excel's 31-char sheet-name limit is Unicode code points. A Chinese title (~3 bytes/char) gets cut at ~10 chars and can split a rune mid-sequence, producing an invalid/garbled sheet name. Python's `current_table_title[:31]` operates on `str` (characters), so this is a real parity gap for CJK content. Use `[]rune(name)` to clamp to 31 runes. (The current tests are ASCII-only and don't cover this.)

2. **Sheet-name uniquification diverges from Python.** Go emits `Name2`, `Name3`; Python emits `Name_1`, `Name_2` (`f"_{counter}"`). Cosmetic, not a correctness bug, but if you want strict parity with the pandas branch the underscore matters (and makes the truncation arithmetic consistent with `original_name[:31 - len(suffix)]`).

3. **Empty tables are not skipped.** Python drops a table when `df is None or df.empty`; `parseMarkdownTables` keeps every table (including a header-only table with no data rows). Minor, but a degenerate `| A | B |` with no body will still produce a sheet in Go.

4. **`messageExportFormat` treats `"plain"` and any unrecognized value as "no export".** Python instead coerces any non-empty, non-file format to `markdown` and still emits an attachment. I think your behavior is the *more correct* one (a `plain` rendering shouldn't fabricate a download), but flagging the divergence so it's an intentional decision rather than an accident.

5. **Shared `output_format` field (nit).** `output_format` now doubles as both rendering style and export trigger, so `html`/`markdown` produce *both* a render and a file. That's documented in `Inputs()` and matches the single UI selector, so it's fine — just noting the ambiguity for future readers.

None of these are blocking; the core restoration is correct and the test coverage is strong. Nice work.

euvre:
Hi @wangq8 — thanks for the thorough review and for verifying the front-end contract and parity up front. All three actionable items are applied in 9b28be184, and the two informational ones are addressed below:

1. **Rune-safe sheet-name truncation (applied).** `sanitizeSheetName` now clamps to 31 characters via `[]rune`, so a CJK title keeps whole characters instead of being cut at byte 31 into invalid UTF-8. The duplicate-name truncation path got the same treatment. New test `TestWriteXLSX_SheetNameTruncatesByCharacters` feeds a 40-char Chinese heading and asserts the sheet name is exactly 31 valid runes.

2. **Suffix parity `_1`/`_2` (applied).** Collisions now append `_1`, `_2`, … computed from the original name truncated to `31 - len(suffix)` characters, matching the reference arithmetic. Covered by `TestWriteXLSX_DuplicateSheetNamesGetUnderscoreSuffix` (two `## Report` sections → sheets `Report` and `Report_1`).

3. **Empty tables dropped (applied).** `parseMarkdownTables` now skips tables with no data rows, and the `Table_N` fallback numbering only counts tables that actually land in the workbook. Content whose tables are all empty falls back to the single `Data` sheet, same as the reference path. Two tests cover the mixed case (header-only table next to a real one) and the all-empty fallback.

4. **plain/unrecognized → no export: kept deliberately.** Agreed it is the more correct behavior, so no change; the `messageExportFormat` doc comment previously claimed it mirrored the markdown coercion, which was inaccurate — it now states the deviation explicitly (only explicit file formats produce an attachment). `TestMessage_NoExportWithoutFileFormat` already pins the behavior.

5. **Shared `output_format` field:** left as is per your note; `Inputs()` documents the dual role.

Verification: `bash build.sh --test ./internal/agent/component/io/...` and `bash build.sh --test ./internal/agent/component/...` both pass on the new commit (including the 4 new regression tests and the existing export tests). These changes are confined to sheet naming/parsing inside the xlsx writer plus a comment, so the unit tier fully exercises them; no runtime/UI surface changed.

buua436 (changes requested):
Review findings:

1. **[P1] HTML export uses already-rendered and escaped content**

Location: `internal/agent/component/message.go:251-252`

The exporter receives `rendered` instead of the original `resolved` content. For `html`, `Render` first HTML-escapes the message body, and `WriteHTML` then writes that escaped value as HTML. As a result, Markdown such as:

```md
# Report
**hello**
```

is exported as literal text instead of a heading and bold text. The export path should use the unrendered content, or each file format should have its own conversion input.

2. **[P2] XLSX sheet-name deduplication is case-sensitive**

Location: `internal/agent/component/io/xlsx_writer.go:220`

The `used` map compares names case-sensitively, while Excelize treats worksheet names case-insensitively. If two tables are titled `Report` and `report`, the second `NewSheet` call resolves to the existing sheet, and the second table is written starting at row 1, overwriting the first table.

The deduplication logic should use case-insensitive comparisons, such as normalized lowercase names or `strings.EqualFold`.

I also attempted to run the targeted Go tests, but the build failed because the environment could not create the required `office_oxide` symlink on the read-only filesystem.

euvre:
Hi @buua436 — thanks for the sharp review. Both findings are confirmed valid and fixed in b365b359d.

**1. [P1] HTML export used already-rendered, escaped content — fixed.**
Confirmed the exact data flow: `Render` HTML-escapes the body (`wrapHTML` → `html.EscapeString`), and `WriteHTML` then wrote that escaped string verbatim into `<main>`, so `# Report` / `**hello**` shipped as literal text. Fixed in two parts, matching Python's `_convert_content` contract:

- `message.go` now passes the resolved, un-rendered content to the exporter — the same input Python feeds `_convert_content` (the template-resolved `all_content`/`content`), not the `output_format` rendering. Each format writer owns its own markdown handling, as you suggested.
- The `html` branch converts the markdown body to an HTML fragment before the document wrapper, mirroring pypandoc's `format="markdown" → html`. It uses goldmark, which is already a direct dependency of this module (via `internal/utility`), so there is no go.mod churn. `# Report` + `**hello**` now export as a real `<h1>` and `<strong>`.

New regression test `TestMessage_ExportAttachmentHTML` asserts the stored blob contains `<h1>Report</h1>` / `<strong>world</strong>` and neither escaped (`&lt;h1&gt;`) nor literal (`# Report`) markdown.

Boundary note: the Go `WriteDOCX`/`WritePDF` writers are plain-paragraph writers that pre-date this PR; they now receive the un-rendered markdown like Python does, but they still don't render markdown structure the way pypandoc does — closing that gap is a separate feature rather than something this PR regressed.

**2. [P2] Case-sensitive sheet-name dedup — fixed.**
Verified against excelize v2.11.0: `GetSheetIndex` compares with `strings.EqualFold` and `NewSheet` returns the existing sheet's index on a case-insensitive match, so `Report` followed by `report` silently aliased to one sheet and the second table overwrote the first from row 1. The `used` map is now keyed by the lowercased name (recomputed after each suffix), keeping the `_1`/`_2` suffix convention. New regression test `TestWriteXLSX_SheetNameCollisionCaseInsensitive` writes `## Report` and `## report` tables and asserts two sheets (`Report`, `report_1`) with both tables' rows intact. (This also covers the case-collision point CodeRabbit re-raised.)

**Verification:** `bash build.sh --test ./internal/agent/component/io/... ./internal/agent/component/...` passes (both packages, including the two new tests and all earlier export/xlsx tests). Both fixes are fully exercised at the component level (Message `Invoke` → exported blob content; `WriteXLSX` → workbook read-back) and no UI/service surface changed, so the unit tier covers them.

Re your local build failure: the `office_oxide` wiring needs a writable location for the native libs; `build.sh --test` resolves them from the user cache (`~/ragflow-native-libs`) here, which is why the same tests run green in this checkout.

yuzhichang (approved):

euvre:
Status note after the latest main merge (HEAD d7d48b35a): no new changes were needed this round — every open review item is already addressed, and I re-verified the branch still holds all of them at the merged HEAD.

- Hi @wangq8 — all three actionable items landed in 9b28be184: rune-safe 31-character sheet-name truncation, `_1`/`_2` suffix parity (with `31 - len(suffix)` truncation arithmetic), and dropping header-only/empty tables so the `Table_N` fallback only counts sheets that actually land. The two informational points are also covered: the deliberate `plain`-→-no-export deviation is now stated explicitly in the `messageExportFormat` doc comment, and the shared `output_format` dual role stays documented in `Inputs()`.
- Hi @buua436 — both findings landed in b365b359d: the exporter now receives the resolved, un-rendered content (each format writer owns its own markdown handling — the `html` branch renders markdown via goldmark before the document wrapper), and sheet-name dedup keys on lowercased names so `Report`/`report` produce distinct sheets instead of aliasing.
- Thanks @yuzhichang for the approval! 🙌

Re-verification at the merged HEAD: `bash build.sh --test ./internal/agent/component/io/... ./internal/agent/component/...` passes (`ok ragflow/internal/agent/component/io`, `ok ragflow/internal/agent/component`), including all the regression tests added for the above (`TestWriteXLSX_SheetNameTruncatesByCharacters`, `TestWriteXLSX_DuplicateSheetNamesGetUnderscoreSuffix`, `TestWriteXLSX_SheetNameCollisionCaseInsensitive`, `TestMessage_ExportAttachmentHTML`, and the empty-table cases).

Please re-check when convenient.

PR #18959, opened 2026-08-28
fix(web): treat scheduled documents as parsing

huyua9:
The backend already exposes SCHEDULE for queued parsing, but the frontend only treats RUNNING as active.

Include SCHEDULE in the shared isParserRunning helper and reuse it for document-list polling and bulk deletion, with focused regression coverage for active versus terminal statuses. This complements the separate numeric status normalization work in PR #18910.

Validation:
- git diff --check: passed
- Focused Jest: not run; web/node_modules is absent


JinHai-CN:
would you please resolve the conflicts?

huyua9:
Merged current `main` into this branch (`08ef88829`). The refetch/delete predicates now go through `isDocumentProcessing`, which builds on `isParserRunning` — the scheduled-as-parsing behavior this PR introduces is covered by that shared helper; `utils.test.ts` keeps both test suites.

dcc123456:
We will fix this issue using a more appropriate approach.

PR #18960, opened 2026-08-28
fix(web): pass document search keywords to filter

huyua9:
## Summary

- restore propagation of the dataset document search string to the document-filter request
- keep the empty-search request unchanged

## Why this is needed

The document filter query key changes with the debounced search string, but after the REST endpoint migration the actual GET request stopped sending `keywords`. The backend still uses that parameter when calculating filter aggregates, so the filter counts and options could describe the whole dataset instead of the searched document set.

## Validation

- `git diff --check` passed
- focused Jest and oxfmt could not run because `web/node_modules` is absent in this environment


JinHai-CN:
please fix the lint issue

cike8899:
Have you tested this feature? Why doesn't the request sent after I enter a keyword include the `keywords` parameter? Please provide a screenshot of your Chrome DevTools Network tab. @huyua9

PR #18962, opened 2026-08-28
fix(web): keep agent prologue from overwriting the first user message

euvre:
### Summary

**Background.** In the agent canvas debug chat, an agent ran with the prologue (opening-remarks) switch off. The user asked a first question and got an answer; afterwards the prologue switch was turned on in the Begin node. The chat panel then replaced the user's first question with the prologue text instead of keeping it. The reporter confirmed the same behavior on both the Go and Python deployments (the web frontend is shared, and the prologue is rendered purely on the frontend).

**Root cause.** `useSendAgentMessage` (web/src/pages/agent/chat/use-send-agent-message.ts) re-runs an effect whenever the Begin form's prologue value becomes non-empty and calls `addPrologue`. The old `addPrologue` (web/src/hooks/logic-hooks.ts) unconditionally rewrote the content of `messages[0]`. `messages[0]` is the prologue only when the conversation started with the prologue enabled; when the conversation started without one, `messages[0]` is the user's first question, so it got overwritten the moment the switch was turned on (and its text was replaced by the prologue copy).

**Fix.**
- The prologue is a frontend-only assistant message; it now carries a stable id `PROLOGUE_MESSAGE_ID` (web/src/utils/chat.ts).
- New pure helpers `mergePrologueIntoMessages` / `dropPrologueFromMessages`: `addPrologue` updates the prologue text in place only when the first message already is the prologue (so editing the prologue text in a fresh chat still live-updates the bubble); otherwise it inserts the prologue above the existing messages, so the user's first question is preserved. Re-running the merge with the same text is a no-op (idempotent).
- Turning the prologue switch off now removes the frontend-only prologue message (`removePrologue`) instead of leaving a stale bubble behind.
- The Explore new-session flow (guarded by `isNew && isEmpty(sessionId)`) and the share page (prologue rendered via `addNewestOneAnswer`, graph store not populated) are unaffected: the helpers only ever create/update/remove the message that carries the prologue id.
- Bonus environment robustness: `docker/launch_backend_service.sh` no longer aborts under `set -e` when jemalloc's `pkg-config` lookup fails (the failed command substitution killed the launcher before any service started on hosts without jemalloc dev files).

**Tests.**
- New unit tests in `web/src/utils/__tests__/chat.test.ts` cover: prologue on an empty list; insert-above-existing-conversation (the exact reported regression, using the reporter's message); in-place text sync when the prologue is edited; idempotent re-merge; drop-only-a-leading-prologue; and no-op when there is no prologue.
- `web/src/pages/agent/chat/use-send-agent-message.test.tsx` passes with the `useSelectDerivedMessages` mock extended for `removePrologue`.
- Note: the `preprocessLaTeX` tests in `web/src/utils/__tests__/chat.test.ts` already fail on `main` (verified against a stashed clean tree) and are unrelated to this change; likewise the remaining `tsc --noEmit` errors are pre-existing and none are introduced by this PR.

**Verification boundary.** Browser end-to-end replay could not be completed in this environment: the browser automation channel was unavailable and the Go server could not be built (missing system dev packages, no sudo). Verification is based on line-level root-cause analysis plus the unit tests above. The change is frontend-only, and the Python deployment (which was brought up locally) exhibits the same reported behavior by construction since the prologue is rendered entirely in the web app.

### DIFF
--- a/docker/launch_backend_service.sh
+++ b/docker/launch_backend_service.sh
@@ -49,7 +49,14 @@ export http_proxy=""; export https_proxy=""; export no_proxy=""; export HTTP_PRO
 export PYTHONPATH=$(pwd)
 
 export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/
-JEMALLOC_PATH=$(pkg-config --variable=libdir jemalloc)/libjemalloc.so
+# jemalloc is an optional LD_PRELOAD allocator; tolerate environments without
+# it instead of letting the failed pkg-config lookup abort the launcher
+# (this script runs under set -e).
+JEMALLOC_PATH="$(pkg-config --variable=libdir jemalloc 2>/dev/null || true)/libjemalloc.so"
+if [ ! -f "$JEMALLOC_PATH" ]; then
+    echo "Warning: jemalloc not found; continuing without LD_PRELOAD."
+    JEMALLOC_PATH=""
+fi
 
 PY=python3
 
--- a/web/src/hooks/logic-hooks.ts
+++ b/web/src/hooks/logic-hooks.ts
@@ -30,7 +30,11 @@ import { IKnowledgeFile } from '@/interfaces/database/dataset';
 import { changeLanguageAsync } from '@/locales/config';
 import api from '@/utils/api';
 import { getAuthorization } from '@/utils/authorization-util';
-import { buildMessageUuid } from '@/utils/chat';
+import {
+  buildMessageUuid,
+  dropPrologueFromMessages,
+  mergePrologueIntoMessages,
+} from '@/utils/chat';
 import axios from 'axios';
 import { EventSourceParserStream } from 'eventsource-parser/stream';
 import { has, isEmpty, omit } from 'lodash';
@@ -654,27 +658,11 @@ export const useSelectDerivedMessages = () => {
   }, []);
 
   const addPrologue = useCallback((prologue: string) => {
-    setDerivedMessages((pre) => {
-      if (pre.length > 0) {
-        return [
-          {
-            ...pre[0],
-            content: prologue,
-          },
-          ...pre.slice(1),
-        ];
-      }
+    setDerivedMessages((pre) => mergePrologueIntoMessages(pre, prologue));
+  }, []);
 
-      return [
-        {
-          role: MessageType.Assistant,
-          content: prologue,
-          id: buildMessageUuid({
-            role: MessageType.Assistant,
-          }),
-        },
-      ];
-    });
+  const removePrologue = useCallback(() => {
+    setDerivedMessages((pre) => dropPrologueFromMessages(pre));
   }, []);
 
   const removeLatestMessage = useCallback(() => {
@@ -749,6 +737,7 @@ export const useSelectDerivedMessages = () => {
     scrollToBottom,
     removeAllMessagesExceptFirst,
     addPrologue,
+    removePrologue,
   };
 };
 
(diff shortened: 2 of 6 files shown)


wangq8:
bash script LGTM to me
@dcc123456 please review frontend code.

wangq8 (commented):
This is an AI review comment.

Nice root-cause analysis and a clean fix. The pure `mergePrologueIntoMessages` / `dropPrologueFromMessages` helpers with the stable `PROLOGUE_MESSAGE_ID` are a solid approach, and the test coverage of the reported regression is good. A few observations:

1. **Inserting above an in-progress conversation.** When the prologue switch is turned on after the user already asked a question, `mergePrologueIntoMessages` inserts the opening remark at the very top of an already-started conversation. That correctly preserves the user's first message (the fix), but it means an "opening remark" appears above messages that were already exchanged — which may read oddly. Worth confirming this is the desired UX vs. only applying the prologue to fresh conversations.

2. **Magic id coupling.** `PROLOGUE_MESSAGE_ID = 'prologue-message'` is a shared magic string relied on by both `logic-hooks.ts` and `chat.ts`, and correctness depends on the backend never emitting a message with that id (and on the prologue always being `messages[0]`). It's safe today, but a short comment near the constant documenting that invariant would help future maintainers.

3. **Scope creep.** The `docker/launch_backend_service.sh` jemalloc change is a reasonable, self-contained robustness fix but is unrelated to the prologue bug. Consider splitting it into its own PR so it's independently reviewable/bisectable.

The `removePrologue()` on the empty-prologue branch is a nice touch (clears the stale bubble when the switch is toggled off), and the idempotent merge correctly avoids re-render churn given the effect's broad dependency list. LGTM otherwise.

This is an AI review comment.

euvre:
Hi @wangq8 — thanks a lot for the careful review (and for the bash-script LGTM). Answers to your three points:

1. **Prologue above an in-progress conversation** — confirmed this is intentional for this call site. The effect in `use-send-agent-message.ts` is the canvas debug chat's live-preview sync: it fires whenever the Begin form's `prologue` param changes (switch toggled or text edited) while a test conversation may already be open. Inserting the opener at the top keeps that preview in sync without dropping any exchanged message (the pre-fix code rewrote `messages[0]`, which is the reported regression). A "fresh conversations only" no-op would make the toggle/text edit look dead mid-debug. Flows where the prologue should only appear on fresh sessions already guard the call site — see `use-send-session-message.ts` (`derivedMessages.length === 0` / `isNew && isEmpty(sessionId)`), which is why Explore is unaffected. I've also spelled out this intent in the `mergePrologueIntoMessages` JSDoc in 3f329e107.

2. **Magic id coupling** — agreed; 3f329e107 extends the comment on `PROLOGUE_MESSAGE_ID` to document both invariants: (a) real message ids are backend-supplied stream ids or `uuid()` values from `buildMessageUuid` (render keys additionally get a `role_` prefix via `buildMessageUuidWithRole`), so they can never equal this literal; (b) merge/drop only ever inspect `messages[0]`, so anything else in the list is left untouched.

3. **Splitting the jemalloc change** — fair point in general. Since you've LGTM'd the script here and it's a 9-line self-contained robustness fix (without it the launcher aborts under `set -e` before any service starts — which is exactly how I hit it while verifying this fix), I've kept it in this PR. Happy to move it to its own PR if a maintainer prefers; it cherry-picks cleanly.

Verification of the follow-up (comment-only) commit: the focused suite `web/src/utils/__tests__/chat.test.ts` passes for all prologue cases (the 3 failing `preprocessLaTeX` tests reproduce identically on `main`, so they're pre-existing and unrelated), and `npm run type-check` reports an error set identical to `main` (210 pre-existing errors, none in files touched here). No shared-service end-to-end run on this line; the behavior claims above are backed by unit tests plus code-path comparison.

wangq8:
no need, by design.

PR #18963, opened 2026-08-28
Fix: Remove the empty response field from the retrieval operator.

cike8899:
### Summary

Fix: Remove the empty response field from the retrieval operator.


wangq8 (commented):
This is an AI review comment.

The change is small and clean, but I want to flag that the backend still fully supports `empty_response` for the retrieval operator, so removing it only from the frontend has functional consequences:

1. **Feature becomes unreachable but still live.** `internal/agent/component/universe_a_wrappers.go` still parses `params["empty_response"]` into `retrievalParams.EmptyResponse` and applies it as a node-level default; `internal/agent/tool/retrieval.go` uses `args.EmptyResponse` as the fallback output when retrieval returns nothing (around lines 163/172/249/291-292); `internal/agent/tool/registry.go` still validates it as a string param. So this removes the UI, not the behavior.

2. **Silent data loss on re-save.** Because `RetrievalPartialSchema` and `initialRetrievalValues` no longer include `empty_response`, an existing canvas whose Retrieval node already has `empty_response` set will silently drop that value the next time the node is edited/saved (zod strips unknown keys). Users won't get any warning that their configured fallback text is gone.

Please confirm this is an intentional deprecation. If so, I'd suggest:
- a stated deprecation plan (ideally removing/neutralizing the backend handling in a follow-up, or keeping it only for backward compat on existing DSLs), and/or
- a migration or note so existing canvases don't silently lose the value.

If it's not intentional, the field should stay in the form until the backend behavior is removed too.

This is an AI review comment.

PR #18964, opened 2026-08-28
fix(agent): reject fractional integer DSL parameters

6ba3i:
Agent DSL fields that require whole numbers now reject fractional values. Decimal fields such as retry delay remain supported.

### DIFF
--- a/internal/agent/component/integer_params.go
+++ b/internal/agent/component/integer_params.go
@@ -1,0 +1,154 @@
+// Copyright 2026 The InfiniFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"strings"
+)
+
+// integerParams lists persisted Agent DSL fields consumed as integer counts by
+// their component implementations. Decimal-valued component parameters are
+// deliberately absent.
+var integerParams = map[string]map[string]struct{}{
+	"agent": {
+		"max_retries":                 {},
+		"max_rounds":                  {},
+		"max_tokens":                  {},
+		"message_history_window_size": {},
+		"optimize_history_window":     {},
+	},
+	"browser": {
+		"max_retries":                 {},
+		"max_steps":                   {},
+		"max_tokens":                  {},
+		"message_history_window_size": {},
+		"timeout":                     {},
+	},
+	"categorize": {
+		"max_retries":                 {},
+		"max_tokens":                  {},
+		"message_history_window_size": {},
+	},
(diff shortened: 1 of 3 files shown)


internal/agent/component/integer_params.go
```diff
@@ -0,0 +1,154 @@
...
+		"max_retries":                 {},
+		"max_tokens":                  {},
+		"message_history_window_size": {},
+	},
+	"loop":              {"maximum_loop_count": {}},
+	"parallel":          {"max_concurrency": {}},
+	"retrieval":         {"top_k": {}, "top_n": {}},
+	"search_my_dataset": {"top_k": {}, "top_n": {}},
+	"search_my_dateset": {"top_k": {}, "top_n": {}},
+	"searchmydateset":   {"top_k": {}, "top_n": {}},
```
wangq8, line 62: search_my_dateset / searchmydateset  should these 2 be typo?

wangq8:
LGTM.

wangq8 (commented):
This is an AI review comment.

Thanks for the PR — the goal (reject fractional values for integer DSL params) is sound, and preserving numeric lexemes via `json.Number` on the update path is the right idea. A few concerns before this lands:

1. **Inconsistent decoding across endpoints.** Only `UpdateAgent` switched to `json.Decoder.UseNumber()`. `CreateAgent` and `PublishAgent` still use `ShouldBindJSON`, and `binding.EnableDecoderUseNumber` is never set anywhere in the repo. On create/publish, JSON numbers decode to `float64`, so `isIntegerNumber(1.0)` returns true (`math.Trunc(1.0)==1.0`) and `1.0` is silently accepted; on update the same payload decodes to `json.Number("1.0")` and is rejected. The stated behavior ("1.0 … rejected instead of being coerced") only holds for the update path. Please apply `UseNumber()` to create/publish too (ideally centralize the decoding) so semantics are consistent.

2. **Flat component shape is not validated.** `validateComponentIntegerParameters` only reads `node["component_name"]` + `node["params"]`. The DSL supports two component shapes (see `decodeComponentFields` / `componentNameFromComp` / `extractComponent`): the v1 `obj.component_name` + `obj.params` shape (reached by recursion into `obj`) and the Go flat `name` + `params` shape, which has no `component_name`. Components in the flat shape are silently skipped by validation. Please also fall back to `name` when `component_name` is empty.

3. **Possible false positive on `invoke.timeout`.** `internal/agent/component/invoke.go` explicitly accepts both `int` and `float64` for `timeout` (fractional seconds). Listing `invoke.timeout` in `integerParams` will reject legitimate fractional timeouts. Please confirm timeout is meant to be integer-only; if fractional is valid, remove it (and audit any other float-backed fields in the map).

4. **Suspicious component-name keys.** `search_my_dataset`, `search_my_dateset` (typo), `searchmydateset`, `searchmydataset` don't match any agent canvas component name — those strings are LLM tool-call function names (see `internal/entity/models/*_test.go`), not DSL `component_name`s. The retrieval component is `"Retrieval"` (already covered). These entries look like they came from guessing names and will never match; consider removing them.

5. **Missing tests.** No `integer_params_test.go` was added. The numeric-type handling (`json.Number` vs `float64`/`int`, `1.0` vs `1`, scientific notation, int64 overflow) is subtle and should have unit tests covering at least: fractional rejection, `1.0` handling, integer pass-through, and the flat-shape skip (once addressed).

Minor: `componentPath` ignores `name` whenever `path` is non-empty (the common case), so the error reports the component path rather than the name; and `json.Number("1e3")` (mathematically an integer) or values overflowing int64 are reported as "must be an integer". Not blockers, just edge cases to be aware of.

This is an AI review comment.

PR #18967, opened 2026-08-28
fix(web): block saving an agent when an edited Parser node is invalid

dcc123456:
### Summary

The page range validation lived in an object-level `superRefine`, which zod skips whenever the base shape fails to parse — a missing `to` swallowed the `from` error entirely, so a start page of -1 only surfaced on a later edit. Move the checks onto the fields so each reports independently.

A node's form also only validates while its panel is mounted, so bad values survived in the store. Re-check the graph against the same schema before persisting. Manual save and run/debug warn and name the offending node; publish and the 20s autosave stay silent and just skip the write. Only nodes the user actually edited this session are checked, tracked via the form's own dirty state, so a node carrying stale DSL never wedges saving.

### DIFF
--- a/web/src/locales/en.ts
+++ b/web/src/locales/en.ts
@@ -567,6 +567,9 @@ Example: A 1 KB message with 1024-dim embedding uses ~9 KB. The 5 MB default lim
       fromMessage: 'Missing start page number',
       toPlaceholder: 'to',
       toMessage: 'Missing end page number (excluded)',
+      pageRangeFromInvalid: 'Start page must be an integer greater than 0',
+      pageRangeToInvalid:
+        'End page must be an integer no less than the start page',
       layoutRecognize: 'PDF parser',
       layoutRecognizeTip:
         'Use a visual model for PDF layout analysis to effectively locate document titles, text blocks, images, and tables. If the naive option is chosen, only the plain text in the PDF will be retrieved. Please note that this option currently works ONLY for PDF documents.',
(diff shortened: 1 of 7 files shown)


wangq8 (commented):
This is an AI review comment.

The approach is sound: moving the page-range checks onto the fields (so a missing sibling doesn't swallow the other error) plus gating saves on only the nodes the user actually edited is the right way to avoid wedging saves on stale DSL. A few points:

1. **`editedNodeFormIds` is never cleared.** It lives for the whole store lifetime. Low risk today (node IDs are unique per graph and the store resets on reload), but consider clearing it on graph load/import so a deleted node's ID can't linger.

2. **`to` message reuse.** The same `pageRangeToInvalid` message ("no less than the start page") is used for both `to < 1` (via `.min(1)`) and `to < from` (via `.refine`). The `.min(1)` failure is really "must be ≥ 1", so the message is slightly misleading when `to < 1` but `from` is also invalid. A separate "must be an integer ≥ 1" string would be clearer.

3. **Only Parser nodes are validated.** Other node types can carry the same stale-invalid-form problem. Fine to scope this PR to Parser, but a follow-up to generalize `findInvalidNode` would be valuable.

4. **`mode: 'onChange'`** now applies to the whole parser form. Please confirm this doesn't cause noisy/early validation or a perf hit on graphs with many parser setups.

5. **Description mismatch (minor):** the PR body references an object-level `superRefine` that isn't present in the current base `FormSchema` (base is just `z.coerce.number()` with no validation — the `fromMessage`/`toMessage` strings are currently unused). The added validation is still correct; just flagging the description.

6. **Nit:** `useValidateNodeForms` is instantiated three times (in `useSaveGraph`, `useSaveGraphBeforeOpeningDebugDrawer`, and `Agent`), each with its own store subscription. Harmless, but could be consolidated.

PR #18969, opened 2026-08-28
fix(web): stop the note node resize handle from covering its textarea

dcc123456:
### Summary

The 14px resize icon sat on top of the textarea's bottom-right corner, hiding text and making the last line hard to click. Shrink the icon to 10px and give the text field bottom padding so the handle has room of its own.


wangq8 (commented):
This is an AI review comment.

LGTM. Shrinking the resize icon from 14px to 10px and adding bottom padding (`pb-2`) to the note node's `FormItem` gives the handle its own space instead of overlapping the textarea's last line. Straightforward presentational change, no regressions.

This is an AI review comment.

PR #18970, opened 2026-08-28
Fix: On the dataset configuration page of the Go version, no error message appears when saving if mandatory fields are left blank.

cike8899:
### Summary

Fix: On the dataset configuration page of the Go version, no error message appears when saving if mandatory fields are left blank.


wangq8:
@cike8899 please resolve conflict.

wangq8 (commented):
This is an AI review comment.

Overall approach is sound: validating each `parser_config` entry against its owning operator form's schema (merging operator defaults first) and mapping issues back to `parser_config.<operatorId>.<path>` is a clean way to surface the previously-silent required-field failures, and mirroring them onto the inner operator forms via `useSyncExternalFormErrors` is a reasonable pattern. The `onChange` mode added to the operator forms is required for the "errors cleared when fields are fixed" behavior.

A few things to address:

1. **i18n:** `SetupSchema`'s `superRefine` hardcodes English messages — `'Fields is required'` and `'Model is required'`. The rest of the form uses `t(...)`; these should be translation keys (or the schema built inside a hook so `t` is available).

2. **Missing tests:** `parser-config-validation.ts` (`addParserConfigIssues`) and `use-sync-external-form-errors.ts` are new, self-contained logic but have no unit tests. The only test change in this PR is prettier reformatting in `utils.test.ts`. Please add tests covering at least: a config missing required fields → issues emitted at the correct path; a valid config (defaults merged) → no issues; unknown operator type → skipped.

3. **Robustness of `onInvalid`:** `Object.keys(errors?.parser_config ?? {})[0]` assumes `errors.parser_config` is always keyed by operatorId. If a future schema adds an issue directly at `parser_config` (no operatorId), `Object.keys` would return keys like `message`/`type` and `setActiveTab` would switch to a non-existent tab. Low risk today, but a guard (e.g. checking the key maps to a known operator) would be safer.

4. **Behavior note (please confirm intentional):** adding `mode: 'onChange'` to the parser/extractor/token-chunker forms means validation now runs on every keystroke, so users will see errors appear while typing rather than only on submit. This is consistent with the error-sync design but is a UX change worth calling out.

This is an AI review comment.

PR #18972, opened 2026-08-28
fix(agent): let global variable textareas resize vertically

dcc123456:
### Summary

The value and description textareas in the global variable sheet were fixed-height, making long values hard to read or edit. Enable vertical resizing with scrolling overflow so users can expand them as needed.


wangq8 (commented):
This is an AI review comment.

LGTM. `fieldConfig` is spread onto the `Textarea` component, which already accepts a `resize` prop (`'none' | 'vertical' | 'horizontal' | 'both'`) and applies it via inline `style` plus an automatic `overflow-auto` class when resizable. So `resize: 'vertical'` correctly enables vertical resizing with scroll overflow.

Minor (non-blocking): the explicit `className: 'overflow-auto'` is redundant — `Textarea` already applies `overflow-auto` whenever `resize !== 'none'` — but it is harmless.

This is an AI review comment.

PR #18973, opened 2026-08-28
feat(models): add 13 Volcengine Ark chat models

chyroc:
## What

`conf/models/volcengine.json` lists one chat model
(`doubao-seed-2-0-pro-260215`) while the provider serves 15. This adds the other
13 there, and the 7 that aren't yet reachable in `conf/all_models.json`.

| | before | after |
| --- | --- | --- |
| `conf/models/volcengine.json` | 1 chat + 1 embedding | 14 chat + 1 embedding |
| `conf/all_models.json` | 8 of these reachable | 15 |

Only 7 go into `all_models.json` because the rest are already reachable there via
the existing alias convention — e.g. `doubao-seed-1-6-250615` carries
`doubao-seed-1-6-251015` as an alias. I kept that pattern rather than adding
competing entries, and didn't re-claim any short alias that an existing entry
already holds.

## Where the numbers come from

Following `conf/model.md`:

- **`content_length` / `max_output`** — verbatim from the official Ark model list
  (docs `82379/1330310`), which states 上下文窗口 and 最大回答 per model. Per
  rule 2 I kept the exact stated decimal figures rather than converting to
  binary, so e.g. `32000`, not `32768`.
- **`max_completion_tokens`** — not described in `conf/model.md`, so I matched
  what existing entries do: their values line up with the largest `max_tokens`
  the endpoint actually accepts (they agree on 3 of the 4 current Doubao
  entries). I measured it per model by binary search with an open upper bound,
  confirming each boundary from both sides (`limit` → 200, `limit + 1` → 400).
  It's **omitted** for `deepseek-v4-pro-ga-260813`,
  `deepseek-v4-flash-ga-260731` and `glm-5-2-260617`, which accept any value
  (`max_tokens=99999999` → 200), so there's no ceiling to state.
- **`model_types` / `thinking` / `tools`** — from the same doc's capability
  column (多模态理解 → `vision`, 工具调用 → `tools`, 深度思考 → `thinking`).

## One existing value corrected

`doubao-seed-1-8-251228` declared `max_completion_tokens: 65536`, but the
endpoint rejects anything above `32768`:

```
max_tokens=32768 -> 200
max_tokens=32769 -> 400
max_tokens=49152 -> 400
max_tokens=65536 -> 400
```

That one line is the only deletion in the diff. Happy to split it out if you'd
prefer this PR to be addition-only.

## Notes on the diff

- No other existing entry is modified.
- `conf/all_models.json` isn't name-sorted upstream, so new entries are inserted
  beside their closest siblings instead of re-sorting — otherwise the diff would
  have churned ~3000 unrelated lines.
- I checked the global alias-uniqueness rule that `model.go` enforces
  (name + aliases in one map): **0 collisions** in both files.
- Both files re-parse as valid JSON.
- I could not run `go test ./internal/entity/models/...` here — the source
  tarball I had available ships `internal/tokenizer` and `internal/utility` empty,
  so the package doesn't build in my environment. I replicated the config-level
  assertions from `volcengine_test.go` (the `url_suffix.models` declaration) and
  the alias check from `model.go` by hand instead. Worth a CI run to confirm.

### DIFF
--- a/conf/all_models.json
+++ b/conf/all_models.json
@@ -5970,7 +5970,6 @@
       "name": "doubao-seed-evolving",
       "content_length": 1024000,
       "max_output": 256000,
-      "max_completion_tokens": 262144,
       "model_types": [
         "chat",
         "vision"
@@ -6097,7 +6096,7 @@
       "alias": [
         "doubao-seed-1-8"
       ],
-      "max_completion_tokens": 32768,
+      "max_completion_tokens": 65536,
       "model_types": [
         "chat",
         "vision"
@@ -6139,7 +6138,6 @@
       ],
       "content_length": 256000,
       "max_output": 256000,
-      "max_completion_tokens": 262144,
       "model_types": [
         "chat",
         "vision"
@@ -6159,7 +6157,6 @@
       ],
       "content_length": 256000,
       "max_output": 256000,
-      "max_completion_tokens": 262144,
       "model_types": [
         "chat",
         "vision"
@@ -6246,7 +6243,6 @@
       "name": "doubao-seed-character-260628",
       "content_length": 128000,
       "max_output": 32000,
-      "max_completion_tokens": 32768,
       "model_types": [
         "chat",
         "vision"
--- a/conf/all_models.json
+++ b/conf/all_models.json
@@ -5967,6 +5967,23 @@
       "max_output": 256000
     },
     {
+      "name": "doubao-seed-evolving",
+      "content_length": 1024000,
+      "max_output": 256000,
+      "max_completion_tokens": 262144,
+      "model_types": [
+        "chat",
+        "vision"
+      ],
+      "thinking": {
+        "default_value": true,
+        "clear_thinking": true
+      },
+      "tools": {
+        "support": true
+      }
+    },
+    {
       "name": "doubao-seed-1-6-flash-250615",
       "model_types": [
         "chat",
@@ -6080,7 +6097,7 @@
       "alias": [
         "doubao-seed-1-8"
       ],
-      "max_completion_tokens": 65536,
+      "max_completion_tokens": 32768,
       "model_types": [
         "chat",
         "vision"
@@ -6114,6 +6131,46 @@
       },
       "content_length": 256000,
       "max_output": 256000
+    },
+    {
+      "name": "doubao-seed-2-1-pro-260628",
+      "alias": [
+        "doubao-seed-2-1-pro"
+      ],
+      "content_length": 256000,
+      "max_output": 256000,
+      "max_completion_tokens": 262144,
+      "model_types": [
+        "chat",
+        "vision"
+      ],
+      "thinking": {
+        "default_value": true,
+        "clear_thinking": true
+      },
+      "tools": {
+        "support": true
+      }
+    },
+    {
+      "name": "doubao-seed-2-1-turbo-260628",
+      "alias": [
+        "doubao-seed-2-1-turbo"
+      ],
+      "content_length": 256000,
+      "max_output": 256000,
+      "max_completion_tokens": 262144,
+      "model_types": [
+        "chat",
+        "vision"
+      ],
+      "thinking": {
+        "default_value": true,
+        "clear_thinking": true
+      },
+      "tools": {
+        "support": true
+      }
     },
     {
       "name": "doubao-seed-2-0-lite-260215",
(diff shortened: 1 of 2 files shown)


chyroc:
Two follow-ups on verification.

**The CI here won't run without a maintainer.** All three jobs report `skipping`
— `tests.yml` gates on `contains(github.event.pull_request.labels.*.name, 'ci')`,
and I can't add that label. So my earlier "worth a CI run to confirm" wasn't
actionable on my side. Sorry for the hand-wave.

**Instead I ran the loader's own validation directly.** I couldn't build
`internal/entity/models` locally (the source tarball I could fetch ships
`internal/tokenizer` and `internal/utility` empty), so I reimplemented the checks
from `InitProviderManager` — `model.go` lines ~300-346 — verbatim in a standalone
Go program and ran it against the edited tree:

```
provider files parsed: 66
all_models.json models: 3840
alias map entries: 7194, duplicates: 0
PASS
```

That covers the three things the loader would reject on: every
`conf/models/*.json` parsing, `conf/all_models.json` parsing, and the global
`name` + `alias` map being collision-free (trimmed, lowercased, one index per
key).

I also confirmed the new ids resolve through that map, including the aliased
ones — `doubao-seed-2-0-mini-260428` resolves to `doubao-seed-2-0-mini-260215`
and `doubao-seed-1-6-flash-250828` to `doubao-seed-1-6-flash-250615`, which is
why those two aren't added to `all_models.json` separately.

Happy to add the `ci` label myself if that's something contributors are meant to
request.

JinHai-CN:
@chyroc would you please check the comments from @coderabbitai

chyroc:
@JinHai-CN thanks — both of CodeRabbit's points were right, fixed in e2177c576.

**`max_completion_tokens` removed.** I checked `internal/entity/models/model.go`:
`Model` declares `ContentLength`, `MaxOutput` and `MaxTokens`, but there is no
`MaxCompletionTokens`, so `InitProviderManager` discards the key and the request
path never reads it. I'd added it because 4 existing Doubao entries carry it and
their values happen to match the endpoint's real `max_tokens` ceiling — but
matching dead metadata isn't a reason to add more of it.

That also retracts the `65536 → 32768` change I'd made to that field. Editing an
ignored key has no effect, so claiming it as a fix was wrong of me. The
underlying measurement still holds if you ever want the field wired up
(`max_tokens=32768` → 200, `32769` → 400 on `doubao-seed-1-8-251228`), but that's
a code change, not a config one.

I left the 186 pre-existing entries that carry the key alone — pruning those felt
like your call, not something to smuggle into this PR.

**`doubao-seed-1-8-251228` limits aligned.** I had used the doc figures
(`256000`/`32000`) in the provider file while `all_models.json` already said
`262144`/`262144`, which is exactly the inconsistency flagged. I tried to settle
it against the endpoint and couldn't: a ~200k-token prompt is accepted, ~250k is
rejected, which fits either number. So I've deferred to the value already in the
repo rather than introduce a disagreement.

The diff is now **`+356/-0`** — purely additive, no existing entry modified.

Re-ran the loader's own validation (the checks from `InitProviderManager`,
reimplemented standalone since `internal/tokenizer` and `internal/utility` are
empty in the tarball I can fetch): 66 provider files parsed, 3840 models, 7194
alias-map entries, 0 duplicates.

One thing I can't do from here: `tests.yml` gates on the `ci` label, so all jobs
show `skipping`. If you add the label I'll act on whatever it reports.

Lynn-Inf (approved):
LGTM

chyroc:
Thanks @Lynn-Inf.

On the latest CodeRabbit note — I think this one is a false positive, so I'd
rather explain than change something that's already correct.

It flags `conf/all_models.json:6099` as *"Changing 32768 to 65536"*. That line is
`"max_completion_tokens": 65536` inside the `doubao-seed-1-8-251228` entry, and
it's **pre-existing upstream content that this PR does not touch**. The entry is
byte-identical to `main`:

```
mine: {"alias":["doubao-seed-1-8"],"content_length":262144,
       "max_completion_tokens":65536,"max_output":262144, ...}
main: {"alias":["doubao-seed-1-8"],"content_length":262144,
       "max_completion_tokens":65536,"max_output":262144, ...}
identical: True
```

The diff is `+356/-0` — zero deletions, and I verified no existing entry differs
from `main` anywhere in the file. The comment is marked "outside diff range",
which I think is the cause: my insertions shift later line numbers, so absolute
line 6099 now points at an untouched line.

It is right about the underlying fact, and that's exactly why the earlier commit
removed `max_completion_tokens` from every entry this PR adds — the field isn't
on `internal/entity/models/Model`, so the loader discards it. I also retracted my
own `65536 → 32768` edit for the same reason. What's left at 6099 is upstream's
value, which I deliberately left alone rather than reformat someone else's data
inside a feature PR.

If you'd like those 186 pre-existing `max_completion_tokens` keys pruned, or the
field wired through the runtime, happy to do either as a separate PR.

PR #18974, opened 2026-08-28
feat(web): display similarity threshold as a percentage

dcc123456:
### Summary

The similarity threshold is stored as a 0-1 float but was shown raw in the slider and number input. Enable the percentage mode of SliderInputFormField so the control reads 0-100 while the underlying form value stays unchanged.

### DIFF
--- a/web/src/components/similarity-slider/index.tsx
+++ b/web/src/components/similarity-slider/index.tsx
@@ -88,6 +88,7 @@ export function SimilaritySliderFormField({
         layout={FormLayout.Vertical}
         tooltip={isTooltipShown && t('similarityThresholdTip')}
         numberInputClassName={numberInputClassName}
+        percentage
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
This is an AI review comment.

LGTM. The change correctly enables the existing `percentage` prop of `SliderInputFormField`, which scales min/max/step by 100 and round-trips the underlying 0–1 float via `Math.round(value) / 100`. With `max={1}` and `step={0.01}` already set on the similarity slider, this yields a 0–100 display in 1% increments while the form value stays unchanged. No regressions (pure presentational change).

Minor (non-blocking): `Math.round(value) / 100` quantizes the value to 2 decimal places. Fine for similarity thresholds (step 0.01), but any future sub-0.01 precision would be lost on the slider round-trip.

This is an AI review comment.

PR #18977, opened 2026-08-28
Fix: remove deprecated qwen tts model

Lynn-Inf:
### Summary

As title.

### DIFF
--- a/conf/models/aliyun.json
+++ b/conf/models/aliyun.json
@@ -46,12 +46,6 @@
       }
     },
     {
-      "name": "qwen-tts-flash",
-      "model_types": [
-        "tts"
-      ]
-    },
-    {
       "name": "qwen-asr-flash",
       "model_types": [
         "asr"


wangq8 (commented):
The removal itself looks correct and the JSON stays valid. Two things to confirm:

1. `qwen-tts-flash` is the **only** `tts` model in `conf/models/aliyun.json`, so this removes Aliyun's last TTS option with no replacement. Is there a successor model to add (e.g. `qwen3-tts-flash` / `qwen3-tts-instruct-flash`)? If a replacement belongs in a separate PR, a note in the description would help reviewers understand the intent.

2. `internal/entity/models/aliyun_test.go` still passes `qwen-tts-flash` directly as the `modelName` to `AudioSpeech` (not via the conf lookup), so those tests won't break — but please confirm the model is fully deprecated on the DashScope side so we aren't silently dropping a still-supported model.

PR #18979, opened 2026-08-28
fix(agent): make user added emppty entries fail to save

6ba3i:
make user added emppty entries fail to save

### DIFF
--- a/internal/agent/component/dynamic_params.go
+++ b/internal/agent/component/dynamic_params.go
@@ -36,34 +36,99 @@ func validateDynamicEntries(value any) error {
 }
 
 func validateDynamicParams(component string, params map[string]any) error {
-	for _, field := range []string{"include_domains", "exclude_domains", "site_include", "site_exclude", "country_include", "language_include", "delimiters", "children_delimiters"} {
+	for _, field := range []string{
+		"include_domains", "exclude_domains", "site_include", "site_exclude",
+		"country_include", "language_include", "delimiters", "children_delimiters",
+	} {
 		if values, ok := params[field].([]any); ok && containsBlank(values) {
 			return fmt.Errorf("[%s] %s does not support empty entries", component, field)
 		}
 	}
-	if values, ok := params["content"].([]any); ok && containsBlankWithContent(values) {
-		return fmt.Errorf("[%s] content does not support empty entries", component)
+	switch strings.ToLower(component) {
+	case "dataoperations":
+		for _, field := range []string{"select_keys", "remove_keys"} {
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
Thanks for the PR. The intent — rejecting blank entries in repeatable agent params — is good, and the `include_domains` / `exclude_domains` / `site_include` / `site_exclude` / `country_include` / `language_include` / `delimiters` / `children_delimiters` path is correct: those are string arrays in the saved DSL, so `containsBlank` catches empty entries.

However, several rules target the wrong data shape and therefore won't actually fire against the DSL the backend receives (`req.DSL`, built by `buildDslComponentsByGraph`):

1. `select_keys` / `remove_keys`: in the DSL these are **string arrays** (the frontend maps `[{name}] → string[]` in `transformDataOperationsParams`), not `[{name}]` maps. The `raw.(map[string]any)` assertion fails for each string, so empty entries here are never caught.
2. `tools`: this is a **list of tool-name strings** (see `AgentComponent.Inputs()` — "List of tool names"), not `[{component_name}]` maps. Dead code.
3. `variables`: Invoke variables use `{key, ref, value}` (`invoke-form/schema.ts`), and VariableAggregator uses nested `groups[].variables[]`. There is no `variable` key, so this never fires. (Loop uses `loop_variables`, which isn't checked at all.)
4. `outputs`: for most components `outputs` is a **map** (`{result: {value, type}}`, `{content: {type, value}}`), not `[]any`. Only Iteration uses an array shape (`[{name, value}]`).
5. `content`: `content` is a string for Agent/Template/Email components, so `params["content"].([]any)` won't match. Please verify which component (if any) has `content` as a list; also `containsBlankWithContent` only flags a mix of blank + non-blank, which is inconsistent with `containsBlank` and needs a comment explaining the intent.
6. `inputs[].options`: Begin/UserFillUp input entries use `{value, type, optional?}`, not `options`. Please confirm which component actually carries an `options` array.

Also incomplete checks:

- `filter_values` rows are `{key, operator, value}` (`data_operations.go` and its tests) — only `key`/`value` are checked, `operator` is missed.
- `loop_termination_condition` rows are `{variable, operator, value, input_mode}` (`loop-form/schema.ts`) — only `variable`/`value` are checked.

Most importantly: **no tests**. A 106-line validation function with this many shape assumptions really needs unit tests with real DSL fixtures (and ideally one golden "valid" fixture to prove it doesn't false-positive on legitimate agents). I'd suggest adding tests before merge.

Minor: typo in the title — "emppty" → "empty".

PR #18980, opened 2026-08-28
feat(agent): duplicate team agent with model/MCP remap by name and cross-server portable export/import

var4dev:
Team agent currently shares one DSL row, so non-owner edits overwrite owner's llm_id/mcp_id and red frames appear. This PR keeps team sharing read-only in practice and adds a proper fork path:

- Add POST /api/v1/agents/<id>/duplicate that copies DSL and remaps llm_id/mcp_id by logical name (model@instance@provider, gpt://...). Both components and graph.nodes are kept in sync. Warnings returned when target model/MCP not found.
- Extend POST /api/v1/agents (create_agent) to auto-remap imported DSL by same name logic, so file import from same DB works without manual fixup.
- Add cross-server portability: GET /api/v1/agents/<id>?portable=true returns DSL with logical names; export hook fetches portable DSL so a file exported on one server can be imported on another and remapped to target's ids.
- Frontend: Duplicate entry in agent card dropdown (More -> Duplicate) calls the new endpoint with fallback to client copy.

Dataset remains team-shared (permission=team) and is not copied - single team dataset is used via retrieval.

Fixes red-frame on team view (ownerTenantId already used) and on duplicate/import where components and graph diverged.


var4dev:
closed per author request

PR #18983, opened 2026-08-28
fix(chat): keep reasoning chats grounded when LLM lacks tool calling

georgelichen:
## Problem

When a chat assistant has a knowledge base and the message-box reasoning level is enabled, the request enters `rag_agent`. If the selected LLM does not support tool calling, `bind_tools` is skipped, but the agentic path still sends the router prompt to the model. The model can then answer from general knowledge without running knowledge-base retrieval or producing citations. The dataset retrieval-test page does not have this problem because it uses the retrieval path directly.

## Solution

- Detect the LLM tool-calling capability immediately after loading the chat models.
- Fall back to the existing `async_chat` RAG path when tool calling is unavailable, preserving knowledge injection, empty responses, and citation generation.
- Normalize list-shaped `doc_ids` only at this fallback boundary so the existing shared `async_chat` parser and its broader callers remain unchanged.
- Add a regression test covering reasoning-enabled fallback and list-shaped document scope forwarding.

Tool-capable models keep the existing agentic RAG flow unchanged.

## Verification

- `ruff check api/db/services/dialog_service.py test/unit_test/api/db/services/test_dialog_service_rag_agent_messages.py`
- `python -m py_compile api/db/services/dialog_service.py test/unit_test/api/db/services/test_dialog_service_rag_agent_messages.py`
- `git diff --check`
- GitNexus staged/compare impact detection: low risk for the two-file change

The focused pytest could not be collected in the local Windows environment because the temporary dependency environment was missing `azure.storage.filedatalake`; live provider and browser verification were not available.

### DIFF
--- a/api/db/services/dialog_service.py
+++ b/api/db/services/dialog_service.py
@@ -2001,6 +2001,21 @@ async def rag_agent(dialog, messages, stream=True, **kwargs):
             yield ans
         return
     kbs, embd_mdl, rerank_mdl, chat_mdl, tts_mdl = get_models(dialog)
+
+    # Agentic RAG depends on the outer model being able to call the bound
+    # ``rag`` tool. Models without tool-calling support would otherwise receive
+    # only the router prompt and could answer from their own knowledge without
+    # ever running retrieval. Reuse the regular RAG path for those models so
+    # the configured knowledge base and citation flow remain authoritative.
+    if not getattr(chat_mdl, "is_tools", False):
+        logging.info("LLM does not support tool calls; falling back to regular RAG chat")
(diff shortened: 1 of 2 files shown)


buua436 (commented):
Reviewed the latest changes. I didn’t find any blocking issues. The fallback for models without tool-calling support correctly preserves the regular RAG retrieval, knowledge injection, empty-response handling, and citation flow. Tool-capable models keep the existing agentic path, and the regression test covers the fallback and list-shaped doc_ids.
Approved.

buua436 (approved):

PR #18993, opened 2026-08-28
fix(web): warn about empty Message component content on agent save

euvre:
### Summary

**Background.** In the agent canvas, a `Message` (回复消息) component whose message content was left empty could be saved without any prompt: the user clicked 保存, got the "更新成功" success toast, and only discovered the problem later when running the agent, where the backend component check fails with `[Message] Content does not support empty value.` (Python: `MessageParam.check()`; the Go runtime treats the node's text as empty as well). The frontend had no save-time validation for this — the message form's zod schema marks `content` as optional, and `saveGraph` PUTs the DSL unconditionally. A locale key `flow.messageMsg` ("Please input message or delete this field." / "请输入消息或删除此字段。") already existed in every locale file but was referenced by no code, showing this validation was intended but never wired up.

**What this PR does.**

- Add `isEmptyMessageContent` / `getEmptyMessageNodeNames` helpers in `web/src/pages/agent/utils.ts`. A Message node counts as empty when `form.content` is missing, not an array, or has no non-blank string entry — mirroring the backend's `_valid_message_content` rule.
- In `useSaveGraph` (`web/src/pages/agent/hooks/use-save-graph.ts`), when an explicit save runs (`showMessage=true`, which covers the Save button, Run, Publish, pipeline run, webhook test and the global-variable sheet), collect Message nodes with empty content and show a warning toast naming them, e.g. `回复消息_0: 请输入消息或删除此字段。`. The save itself is not blocked (the warning is advisory, consistent with other canvas warnings). Background autosave (`showMessage=false`) stays silent so editing is not interrupted by repeated toasts.

**Verification.**

- New unit tests in `web/src/pages/agent/utils.test.ts` (jest): 3/3 pass, covering missing/non-array/blank-only content, non-string entries, variable placeholders like `{begin@query}` being non-empty, and node filtering across Message vs non-Message nodes.
- `oxlint` clean on all touched files (the one exhaustive-deps warning introduced by the change was fixed by adding `showMessage` to the callback deps).
- `npm run type-check`: no errors in the touched files (the repo has 208 pre-existing TS errors elsewhere, unchanged by this PR).
- Boundary: end-to-end browser reproduction was not possible in this environment (the automation browser backend was unavailable, and the service launcher failed in the Go build step because `sudo` is unavailable to install `libpcre2-dev`), so the fix is verified by unit tests, lint and type-check plus code-level alignment with the backend validation, not by a live UI walk-through.

### DIFF
--- a/web/src/pages/agent/hooks/use-save-graph.ts
+++ b/web/src/pages/agent/hooks/use-save-graph.ts
@@ -10,12 +10,14 @@ import {
 } from '@/interfaces/database/agent';
 import { formatDate } from '@/utils/date';
 import { useDebounceEffect } from 'ahooks';
+import { t } from 'i18next';
 import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router';
 import { Operator } from '../constant';
 import { FormSchema as ParserFormSchema } from '../form/parser-form';
 import useGraphStore from '../store';
+import { getEmptyMessageNodeNames } from '../utils';
 import { useBuildDslData } from './use-build-dsl';
 
 /**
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks — the frontend predicate matches the backend `_valid_message_content` rule (`agent/component/message.py`) exactly, and the tests cover the key cases. A few items:

1. This PR is currently in conflict with `main` (mergeable=CONFLICTING). `main` has since refactored `use-save-graph.ts` (added `findInvalidNode`/`useValidateNodeForms` and `getInvalidNode(currentNodes)` inside `saveGraph`). Please rebase and re-check the integration.

2. The empty-content check reads `useGraphStore.getState().nodes`, but `saveGraph` saves `currentNodes` (which may be an explicit list — e.g. `handleRun` passes `nextNodes`). Use `currentNodes ?? useGraphStore.getState().nodes` so the warning reflects the nodes actually being saved, consistent with `getInvalidNode(currentNodes)`.

3. Advisory-only warning gated on `showMessage` is the right call (autosave stays silent).

4. Minor: `import { t } from 'i18next'` is consistent with existing usage elsewhere (`pages/agent/constant`), so that's fine — no change needed.

This is an AI review comment.

dcc123456 (approved):

PR #18994, opened 2026-08-28
fix(graphrag): stop query_rewrite masking a parse failure with json_repair

marmar9615-cloud:
### Summary

`query_rewrite` reads `.get` straight off whatever `json_repair.loads` returns. `loads` returns `""` for a reply holding no JSON object, so that raises `AttributeError: 'str' object has no attribute 'get'`. The handler below catches `json_repair.JSONDecodeError`, an attribute `json_repair` does not have, so evaluating that clause raises a second `AttributeError` that propagates in place of the first and names `json_repair` as the problem.

```
AttributeError: 'str' object has no attribute 'get'

During handling of the above exception, another exception occurred:

AttributeError: module 'json_repair' has no attribute 'JSONDecodeError'
```

The masking covers everything the `try` can raise. `loads` also raises `ValueError` once nesting passes the parser's recursion limit, measured here at 332 unterminated brackets on the pinned `json-repair==0.60.1`, and today that is replaced by the same `AttributeError`.

Present from v0.16.0 through v0.27.1. Verified by reading the file at each tag rather than with `git tag --contains`, which under-reports in a shallow clone. The path was `graphrag/search.py` before it moved under `rag/`.

### What it costs

This degrades a query, it does not fail one. `KGSearch.retrieval` is the only caller, and it catches the exception and falls back to `ents = [qst]`. So the search loses its entity lookup, the n-hop expansion built from those entities, and the community grounding. Relations found from the question text still run either way, so this is a quality loss rather than an outage.

### Why it keeps raising

Returning empty keywords would be worse than the bug. `get_relevant_ents_by_keywords` short-circuits on `if not keywords: return {}`, so `([], [])` skips retrieval's fallback and drops all three of those. Raising hands control back to the caller, which already knows what to do.

### Why the retry branch goes rather than gets repaired

It has never run since it was written. Repairing it would activate a path that has had no production exposure, and that path runs `.replace("user", "").replace("model", "")` across the whole reply. A reply carrying `["user manual"]` and `["model T"]` would come back as `[" manual"]` and `[" T"]`, which then goes into a vector search. That is a worse outcome than the fallback it would be replacing.

Objects inside a list are merged instead, which is the shape the branch was really there to catch. A model that wraps its object in an array, or emits two objects in a row, parses to a list with the keywords still in it. This matches how `content_tagging` handles the same shape in #18991.

### Relationship to #18991

Same broken construct, different consequence, so the fixes differ on purpose. In `content_tagging` the `.items()` call sits outside the `try`, so the clause is merely dead and the failure is a clean `AttributeError`. Here the `.get` is inside, so it masks. There, `{}` means "no tags for this chunk" and the caller expects it; here, empty keywords would suppress a useful fallback, so this raises.

They agree on the parts that should agree. Neither logs the model reply, because it can echo the question and the entity samples the prompt carries. Both merge objects out of a list rather than discarding them.

On #18991 I said I would send this separately if a maintainer asked, and nobody has. Sending it anyway because the two are easier to judge together than weeks apart. Close it if you would rather it waited.

### Testing

`keywords_from_query_rewrite` is a module-level function so it can be tested at all. `KGSearch` subclasses `Dealer`, which `test/unit_test/rag/graphrag/conftest.py` mocks, so under that conftest `KGSearch` is a `MagicMock` and `isinstance(KGSearch, type)` is `False`.

| source under test | result |
| --- | --- |
| this PR | 11 passed |
| the current handler | 8 failed |
| returns `{}` instead of raising | 7 failed |
| list handling removed | 2 failed |
| retry branch kept, wrapped in `except Exception` | 1 failed |

```
$ pytest test/unit_test/rag/graphrag/
123 passed

$ ruff check .          # All checks passed
$ ruff format --check   # already formatted
```

### DIFF
--- a/rag/graphrag/search.py
+++ b/rag/graphrag/search.py
@@ -32,6 +32,36 @@ from common import settings
 from common.doc_store.doc_store_base import OrderByExpr
 
 
+def keywords_from_query_rewrite(result: str) -> dict:
+    """Return the keyword object a query-rewrite reply carries.
+
+    json_repair.loads returns "" for a reply holding no JSON object, so the value has to be
+    checked before it is read. It was not: query_rewrite read .get straight off it, and the
+    resulting AttributeError met `except json_repair.JSONDecodeError`. json_repair has no
+    such attribute, so evaluating that clause raised a second AttributeError that propagated
+    in place of the first and named json_repair as the problem.
+
+    Raising on an unusable reply is deliberate. KGSearch.retrieval catches it and falls back
+    to searching on the question, which keeps entity, n-hop and community retrieval running.
+    Returning empty keywords would skip that fallback and lose all three.
+    """
+    keywords_data = json_repair.loads(result)
+    if isinstance(keywords_data, dict):
+        return keywords_data
+    if isinstance(keywords_data, list):
+        # A model that wraps its object in an array parses to a list. Merge the objects it
+        # holds, the same way content_tagging treats this shape.
+        merged = {}
+        for item in keywords_data:
+            if isinstance(item, dict):
+                merged.update(item)
+        if merged:
+            return merged
+    # The reply can echo the question and the entity samples in the prompt, so report its
+    # shape rather than its content.
+    raise ValueError(f"query_rewrite expected a JSON object of keywords, got {type(keywords_data).__name__} from a {len(result)} character reply")
+
+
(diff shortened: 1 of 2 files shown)


yingfeng:
graphrag is not used any more

PR #18996, opened 2026-08-28
fix(agent): warn at save time when an Agent node has no model selected

euvre:
### Summary

An Agent (智能体) component whose model selector was left empty passed canvas save silently ("更新成功"), and only failed later when the user asked a question in the run drawer — with a raw Go error (`canvas invoke: [NodeRunError] ... component: invalid param "model_id": required`) that was even attributed to the downstream Message node consuming the Agent's deferred stream, not to the Agent node itself. The user had no chance to notice the misconfiguration at save time.

**Root cause.** The canvas save path (`web/src/pages/agent/hooks/use-save-graph.ts`) already re-validates node forms before persisting (it blocks save for edited Parser nodes whose form fails the zod schema), but no check existed for an Agent node with an empty `llm_id`. The Go runtime component (`internal/agent/component/agent.go`, `ParamError{Field: "model_id", Reason: "required"}`) only rejects the canvas when a question is asked, so the gap surfaced far too late and with a confusing message.

**Fix.**

- Add a pure helper `findAgentNodeWithoutModel` (`web/src/pages/agent/utils/agent-node-model.ts`) that finds the first Agent node whose `form.llm_id` is empty.
- Wire it into the existing save-time validation: `findInvalidNode` now returns a `{node, messageKey}` descriptor; explicit Save and Run paths warn with the new message and skip the write, autosave/publish silently skip the write — exactly the semantics already established for invalid Parser forms.
- Unlike the Parser schema re-check, the missing-model check is deliberately **not** gated on `editedNodeFormIds`: a canvas loaded from storage with an empty model must warn on the very next save, not only after the node is re-edited (this is precisely the reported scenario).
- New i18n key `flow.agentModelMissing` (en/zh): `无法保存：“{{name}}” 未选择模型，请先选择` / `Cannot save: "{{name}}" has no model selected. Please choose one first`.

**Verification.**

- Unit tests: `web/src/pages/agent/utils/agent-node-model.test.ts` (3 cases: empty model flagged, model present accepted, non-Agent nodes ignored) — `npx jest src/pages/agent/utils/agent-node-model.test.ts` passes.
- `npx oxlint` on all touched files: 0 warnings/errors. `npx tsc --noEmit` reports no errors in the touched files (the repo currently has pre-existing errors in unrelated files).
- End-to-end browser verification could not be completed in this environment (fresh workspace; the full Python dependency download could not finish in time), so the runtime behavior was verified by code-path analysis against the existing Parser validation flow rather than a live UI walk-through.


wangq8 (commented):
This is an AI review comment.

Overall this is a clean fix and the unit tests are good. A few points before merge:

1. Overlap with #18997 — that PR (despite its unrelated "memory" title) currently contains a second implementation of this exact feature (form-level `llm_id` zod `.min(1)` + `markNodeFormEdited`). The two PRs modify the same files (`use-save-graph.ts`, `en.ts`/`zh.ts`) and will conflict with each other. Please consolidate so only one lands.

2. The missing-model check is intentionally not gated on `editedNodeFormIds` (unlike the Parser check). This means a canvas/template loaded with a model-less Agent node will block every save until that node is fixed, even if the user is only trying to save unrelated edits. That is defensible for "don't persist an unrunnable agent", but it is a real behavior change from "save silently succeeds" — please confirm this is intended, especially for draft/incomplete canvases.

3. Minor: `agent-node-model.ts` and `agent-node-model.test.ts` are missing a trailing newline.

4. Nit: `findAgentNodeWithoutModel` uses a falsy check (`!llm_id`), which also treats `llm_id: 0` / `null` as missing — acceptable given the schema requires a string, just noting it.

This is an AI review comment.

euvre:
Thanks for the review!

- **Overlap with #18997**: that PR has been closed — it was mis-delivered by our automation (the branch was assembled from an unrelated interrupted run's worktree state and its diff was a second, conflicting implementation of this feature). This PR is now the single implementation.
- **Not gating on `editedNodeFormIds` is intended.** The reported scenario is precisely a canvas loaded from storage whose Agent node never had a model selected; gating on "edited this session" would let that misconfiguration keep saving silently. A model-less Agent is unrunnable by backend definition (`model_id: required`), so blocking every save until the node gets a model is the desired behavior — for draft canvases the remedy is to pick a model (or remove the node) once, after which saves proceed normally. The Parser check stays edit-gated for a different reason: schema drift would let *old* DSL fail a *new* zod schema and wedge saving of untouched canvases, while an empty `llm_id` has no such legacy concern.
- **Trailing newlines** added in 416a77f.
- The `!llm_id` falsy check is left as-is (the form schema requires a string; noted the caveat).

dcc123456 (approved):

PR #18999, opened 2026-08-29
Fix: AnthropicCV.describe_with_prompt raises TypeError on every call

Anai-Guo:
## Problem

`AnthropicCV.describe_with_prompt` calls `self.prompt` with two positional
arguments:

```python
# rag/llm/cv_model.py:1149-1151
def describe_with_prompt(self, image, prompt=None):
    b64 = self.image2base64(image)
    prompt = self.prompt(b64, prompt if prompt else vision_llm_describe_prompt())
```

but `prompt` is defined exactly once in this module, on `Base`, and takes only
`b64`:

```python
# rag/llm/cv_model.py:315
def prompt(self, b64):
```

`AnthropicCV` does not override it, so every call raises:

```
TypeError: AnthropicCV.prompt() takes 2 positional arguments but 3 were given
```

`describe_with_prompt` is the entry point used whenever a custom vision prompt is
configured, so that feature is currently unusable on the Anthropic provider.

## Sibling baseline

`Base` already provides the two-argument variant, which even performs the same
`vision_llm_describe_prompt()` fallback internally:

```python
# rag/llm/cv_model.py:328
def vision_llm_prompt(self, b64, prompt=None):
    return [{"role": "user", "content": self._image_prompt(prompt if prompt else vision_llm_describe_prompt(), b64)}]
```

Every other provider's `describe_with_prompt` calls it — `GptV4` (line 357),
`OllamaCV` (line 841), `BedrockCV` (line 1422). `AnthropicCV` is the only one
that reaches for `self.prompt` instead. Since `vision_llm_prompt` dispatches
through `self._image_prompt`, `AnthropicCV`'s own override at line 1127 is still
what builds the Anthropic-shaped content blocks.

## Fix

```python
prompt = self.vision_llm_prompt(b64, prompt)
```

One line; no behaviour change for any other provider.

## Verification

The full ragflow dependency tree isn't installable here, so the real method
sources (`Base.prompt`, `Base.vision_llm_prompt`, `AnthropicCV._image_prompt`,
`AnthropicCV.describe_with_prompt`, `AnthropicCV.describe`) were AST-extracted
from `rag/llm/cv_model.py` at `main` and re-hosted on a stub whose only fakes are
the Anthropic client and `image2base64`:

```
[UNPATCHED] TypeError: AnthropicCV.prompt() takes 2 positional arguments but 3 were given
[PATCHED  ] ok -- prompt text='Summarise this chart', image blocks=34
```

The patched run confirms the caller's custom prompt text actually lands in the
request payload, so the fix restores the intended behaviour rather than just
silencing the error.

`ruff check` with the repo's `pyproject.toml` passes, and
`vision_llm_describe_prompt` remains in use at lines 329 and 984, so its import
is still live.

## Deliberately not changed

`AnthropicCV._image_prompt` does not normalise a bare `str` into a one-element
list the way `Base._image_prompt` does (lines 145-146), so the base64 data URL is
iterated character by character. That is a **pre-existing, separate** issue: the
same repro shows the untouched `AnthropicCV.describe()` path produces the
identical block count, so this PR neither introduces nor worsens it. Happy to
open a follow-up if you'd like it addressed.


Lynn-Inf (commented):
Requesting changes — the diagnosis is right and the replacement is mechanically correct, but on its own it doesn't restore the feature, and I think it makes the failure mode worse rather than better.

  `Base.prompt` (`cv_model.py:315`) takes only `b64`, so `self.prompt(b64, ...)` raises on every call, and `vision_llm_prompt` (`cv_model.py:328`) is the right target: same 2-arg signature, the same `prompt if prompt else vision_llm_describe_prompt()` fallback internally, and it dispatches through `self._image_prompt`, which `AnthropicCV` overrides (`cv_model.py:1127`), so the Anthropic content shape is preserved. There's no regression risk here — this path currently fails 100% of the time and `describe()` is untouched.

  The problem is what happens once the call no longer raises. `image2base64` returns a single data-URL string, and `AnthropicCV._image_prompt` (`cv_model.py:1127`) is missing the `str -> [str]` normalisation that `Base._image_prompt` has (`cv_model.py:146-147`). So `for img in images:` iterates the string character by character and emits one `{"type":"image","source":{"data":"<one char>"}}` block per character — a real image yields hundreds of thousands of blocks. The custom-prompt path therefore still doesn't work, it just fails differently: instead of a local `TypeError` with no network round-trip, we now send a malformed payload to Anthropic and get a remote 400, plus the input-token cost and a much less diagnosable error.

  Suggested direction: include the normalisation fix here — mirror `Base._image_prompt` in `AnthropicCV._image_prompt`:

  ```python
  if isinstance(images, str) or "bytes" in type(images).__name__:
      images = [images]
  ```
  That's a couple of lines and it also fixes describe() (cv_model.py:1146), which passes a bare string through the same override and has the identical latent bug. Happy to approve once both are in; if you'd rather keep this PR as the one-line cleanup, I'd suggest holding it until the _image_prompt fix lands rather than merging it standalone.

PR #19001, opened 2026-08-29
fix(memory): keep unparseable invalid_at null so memories stay valid

anneheartrecord:

Follow-up parity fix for the Python/Go memory implementations tracked in #18415 (same line of work as #18752).

When the extracting LLM emits an `invalid_at` that is non-empty but unparseable, the two implementations currently give the same memory two different lifetimes:

- **Python** (`api/db/joint_services/memory_message_service.py:202`) passes `fallback=""` to `format_iso_8601_to_ymd_hms`, stores `None`, and the memory **stays valid**.
- **Go** (`internal/service/memory_extractor.go` `buildExtractedMessage`) fell back to `now`, persisting the very same LLM output as **"expired at extraction time"**.

### How was this PR solved?

- `buildExtractedMessage` now stores `invalid_at` as `null` when the value is empty or unparseable, mirroring the Python semantics. Parseable values keep their own wall clock (unchanged).
- The layout loop moved into a shared `parseFlexibleTime` so each caller applies its own fallback policy; `formatMemoryTime` behavior is unchanged (valid_at still falls back to `now`, matching Python's `fallback=conversation_time`).

### Behavior change surface

Only items whose `invalid_at` is non-empty **and** unparseable: previously stored as extraction-time (immediately expired), now stored as `null` (still valid). Parseable and empty inputs are unaffected.

### Tests

- New `TestBuildExtractedMessageInvalidAtStaysValidWhenUnparseable` covers unparseable → `nil`, empty → `nil`, parseable → wall clock.
- `TestFormatMemoryTime*` continue to pin the unchanged helper behavior.
- Note: `go test ./internal/service/` cannot compile on macOS hosts because `internal/agent/sandbox` uses the Linux-only `syscall.SysProcAttr.Pdeathsig` (pre-existing, unrelated to this change); the fix logic was additionally verified with a standalone harness over the six relevant scenarios, and CI is the arbiter.

Checksums of intent: `gofmt` clean; `git diff --stat upstream/main...HEAD` touches only `internal/service/memory_extractor.go` and its test.


anneheartrecord:
Nice find — I hit the same divergence in a parallel comparison of this subsystem (my PR overlapped with this one, withdrawing it to keep one clean thread).

One additional gap I verified that this PR doesn't cover: **reduced-precision ISO 8601**. Python's `dateutil.parser.isoparse` accepts `2026-08-20T10:05` and `2026-08-20 10:05` (verified: both parse to `2026-08-20 10:05:00`), but the layout set here (`RFC3339`, `2006-01-02T15:04:05`, `2006-01-02 15:04:05`, `2006-01-02`) rejects them — so for that identical LLM output, Python parses the timestamp while Go falls back to the run time (`formatMemoryTime`'s fallback), which for `valid_at` re-anchors the memory to "now". Adding `"2006-01-02T15:04", "2006-01-02 15:04"` to the layout list closes it. Happy to send that as a small follow-up PR after this merges if you'd rather keep this one focused.

Lynn-Inf (approved):
**LGTM** — the fix matches Python's semantics: `invalid_at` that is empty or unparseable now persists as `null` (memory stays valid) instead of `now` (expired at extraction time), parseable values keep their own wall clock, and `valid_at` still falls back to `now` per Python's `fallback=conversation_time`. Pulling the layout loop into `parseFlexibleTime` so each caller applies its own fallback policy is the right shape for this — the two fields genuinely need different policies, and that was impossible to express while `formatMemoryTime` owned the loop. Folding the empty-string case into the error return is also right, since both callers handle empty and unparseable identically.

**Non-blocking:** `(time.Time, error)` carries the usual zero-value trap — `time.Time{}` is `0001-01-01 00:00:00 UTC`, a valid non-nil value that can't be nil-checked, so a future caller that forgets the `err` check would silently format year 1 into storage rather than failing loudly. Both current call sites check `err == nil` and only use `t` inside that branch, so there's no bug today, and I wouldn't switch to `*time.Time` (pointer handling at both sites isn't worth it here). I'd just make the contract explicit in the helper's doc comment, e.g. *"callers must check `err`: on failure the returned `time.Time{}` is year 1 and must not be used"* — worth doing precisely because this PR is about a silent fallback producing the wrong memory lifetime.

JinHai-CN:
@anneheartrecord would you please resolve the conflicts?

anneheartrecord:
@JinHai-CN done — rebased onto current `main` (`13bcb27`), conflicts resolved, and the PR is mergeable again.

One note on how they resolved: #19189 landed `normalizeMemoryTime`, which does what the `parseFlexibleTime` helper in this PR was doing, so I dropped mine and reused yours. What is left is just the semantic change — `buildExtractedMessage` no longer falls back to `now` when `invalid_at` is non-empty but unparseable, so it persists as null and the memory stays valid, matching Python `memory_message_service` (`fallback=""` → `None`). Empty input already behaved that way after #19189, and parseable values are untouched. The production diff is now 5 lines.

Document ids are unaffected: the fingerprint only feeds `invalid_at` in when it actually parsed, before and after. `TestBuildExtractedMessageUsesStableDocumentIDAcrossRetries` still passes, and the new test goes red if the `now` fallback is put back.

JinHai-CN:
@anneheartrecord would you please check the CI error?

JinHai-CN:

anneheartrecord:
@JinHai-CN checked it, and the cause is worse than the formatting error it showed up as. My Sep 10 merge resolution took main's side of the conflict, so the actual change was gone and two lines were left space-indented. gofmt failed, preflight failed, and every test job was skipped — the PR you were looking at was effectively a whitespace diff.

Rebased onto current main (`4ad80963`) and re-applied as a single commit: same two files, +40/-4.

The delta against main today: #19189 and #19326 both kept the `strings.TrimSpace(item.InvalidAt) != "" && !ok → now` fallback, so an `invalid_at` the model made up is persisted as the extraction time and the memory is expired the moment it is written. The Python path is `format_iso_8601_to_ymd_hms(extracted_content["invalid_at"], fallback="")` in `api/db/joint_services/memory_message_service.py:202` — unparseable becomes empty there and the memory stays valid. This drops the Go fallback so the two sides agree; `valid_at` keeps its own `now` fallback, untouched.

Verification: `gofmt -l` is clean on both files. `./internal/service/...` does not build on macOS (`internal/agent/sandbox` uses `syscall.Pdeathsig`, and `office_oxide` wants its cgo header), so I ran `buildExtractedMessage` and `normalizeMemoryTime` verbatim in a standalone harness: fails with `invalid_at = 2026-08-20 10:05:00` before the change, passes after. CI runs the real test.

One thing I cannot do from my side — GitHub rejects my reopen with a 422, since the head branch was force-pushed after the PR was closed. Could you reopen it? If you would rather have a clean PR for the same branch, say so and I will open one.

PR #19002, opened 2026-08-29
fix(memory): restore Python-side rules in Go TYPE_INSTRUCTIONS

anneheartrecord:

Parity fix for the Python/Go memory implementations tracked in #18415 (same line of work as #18752).

The Go `TYPE_INSTRUCTIONS` map dropped several rules that Python's `memory/utils/prompt_util.py` carries, so the two implementations hand the extracting LLM different contracts:

| missing in Go | where Python has it |
|---|---|
| semantic Examples line (`"The capital of France is Paris", "Water boils at 100°C"`) | `prompt_util.py:46` |
| semantic `(e.g., law enactment, discovery)` / `(e.g., repeal, disproven)` parentheticals | `prompt_util.py:47-48` |
| **semantic `- Default: valid_at = conversation time, invalid_at = "" for timeless facts`** | `prompt_util.py:49` |
| episodic Examples line + `Extract explicit times:` rule | `prompt_util.py:57,60` |
| procedural Examples line + version-specific / best-practices rules | `prompt_util.py:68,71-72` |
| full titles (`Timestamp Rules for Semantic/Episodic/Procedural Knowledge:`) | `prompt_util.py:47,56,65` |

The semantic **Default** line is the load-bearing one: it tells the model what to do for timeless facts, and the Go side has been extracting without it.

### How was this PR solved?

Restore the missing rules verbatim from `prompt_util.py` (content parity; each side keeps its own indentation style). Added `TestTypeInstructionsStayInLockstepWithPython`, which pins the previously-dropped lines so the map cannot silently diverge again.

### Behavior change surface

Only the system prompt assembled for memory extraction (`PromptAssembler.AssembleSystemPrompt`) when no custom system prompt is configured: the instructions now carry the same rules/examples as the Python implementation. No storage, API, or parsing changes.

### Tests

- New `TestTypeInstructionsStayInLockstepWithPython` fails against the previous map (verified: the pinned lines are absent from `upstream/main`) and passes with this change.
- Note: `go test ./internal/service/` cannot compile on macOS hosts because `internal/agent/sandbox` uses the Linux-only `syscall.SysProcAttr.Pdeathsig` (pre-existing, unrelated to this change); `gofmt` is clean and CI is the arbiter.

`git diff --stat upstream/main...HEAD` touches only `internal/service/memory.go` and `internal/service/memory_extractor_test.go`.

### One question alongside this (parity cadence)

Context: as part of #18415's follow-up we ran a module-by-module Python/Go comparison over the memory subsystem (prompt assembly, extraction pipeline, time handling, message storage, retrieval) and collected ~41 drift points, 9 of them bug-shaped — this map is one of them, and #19001 (sibling PR) is another. The Go port is otherwise a remarkably faithful line-by-line port, so the drift reads like "not synced yet" rather than "intentionally diverged".

Which of these is the intended workflow for this class of finding?

1. Keep the two implementations aligned and accept parity PRs like this one (we can send them in small, per-module batches with tests); or
2. the Go side is expected to gradually replace Python, and drift in the interim is accepted (in which case we'd stop proposing parity fixes and close out the effort).

Happy to align with whichever — no hard feelings either way.

### DIFF
--- a/internal/service/memory.go
+++ b/internal/service/memory.go
@@ -121,34 +121,45 @@ You are an expert at analyzing conversations to extract structured memory.
 6. Maximum {max_items} items per type
 `
 
-// TYPE_INSTRUCTIONS contains specific instructions for each memory type extraction
+// TYPE_INSTRUCTIONS contains specific instructions for each memory type extraction.
+// Kept in lockstep with Python's memory/utils/prompt_util.py TYPE_INSTRUCTIONS —
+// both implementations drive the same extraction contract, and a rule present on
+// only one side (e.g. the semantic Default line) drifts extraction behavior across
+// languages (see #18415).
 var TYPE_INSTRUCTIONS = map[string]string{
 	"semantic": `
 **EXTRACT SEMANTIC KNOWLEDGE:**
 - Universal facts, definitions, concepts, relationships
 - Time-invariant, generally true information
+- Examples: "The capital of France is Paris", "Water boils at 100°C"
 
-**Timestamp Rules:**
-- valid_at: When the fact became true
-- invalid_at: When it becomes false or empty if still true
+**Timestamp Rules for Semantic Knowledge:**
+- valid_at: When the fact became true (e.g., law enactment, discovery)
+- invalid_at: When it becomes false (e.g., repeal, disproven) or empty if still true
+- Default: valid_at = conversation time, invalid_at = "" for timeless facts
 `,
 	"episodic": `
 **EXTRACT EPISODIC KNOWLEDGE:**
 - Specific experiences, events, personal stories
 - Time-bound, person-specific, contextual
+- Examples: "Yesterday I fixed the bug", "User reported issue last week"
 
-**Timestamp Rules:**
+**Timestamp Rules for Episodic Knowledge:**
 - valid_at: Event start/occurrence time
 - invalid_at: Event end time or empty if instantaneous
+- Extract explicit times: "at 3 PM", "last Monday", "from X to Y"
 `,
 	"procedural": `
 **EXTRACT PROCEDURAL KNOWLEDGE:**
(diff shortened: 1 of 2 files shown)


Lynn-Inf (approved):
**LGTM** — I verified the restored map line-by-line against `memory/utils/prompt_util.py:39-71`: the per-type `Examples` lines, the full `**Timestamp Rules for <Type> Knowledge:**` titles, the semantic `(e.g., law enactment, discovery)` / `(e.g., repeal, disproven)` parentheticals, episodic `Extract explicit times`, and the procedural version-specific / best-practices rules all match verbatim. Restoring the semantic `Default: valid_at = conversation time, invalid_at = "" for timeless facts` line is the important one — without it the Go side was extracting timeless facts with no defined contract.

**Non-blocking:** there's one residual difference. Python's `TYPE_INSTRUCTIONS` values are class-body triple-quoted strings carrying 8 leading spaces per line, and `_generate_type_instructions` (`prompt_util.py:147-150`) is a plain `"\n".join(instructions)` — there's no `textwrap.dedent` or `.strip()` anywhere in `prompt_util.py`, so that indentation reaches the assembled system prompt, while Go's raw strings are unindented. Harmless for the model (leading spaces in a plain-text prompt are inert), but the two prompts aren't byte-identical, so tokenization differs slightly. If strict parity is the goal, a `textwrap.dedent` on the Python side would close the last gap.

Two smaller notes: the new test is a good pragmatic guard against this specific regression, though being substring-based it won't catch future Python-side additions or other removals (and by design not indentation) — reasonable, since pinning against the Python source from Go would be undesirable coupling. And on your process question in #18415: while both implementations are live and user-visible, I'd favour keeping them in lockstep, since divergent extraction contracts across languages is a genuine bug surface rather than cosmetic drift.

PR #19003, opened 2026-08-29
fix(memory): persist unparseable invalid_at as empty, aligning Go with Python

anneheartrecord:
## Summary

While comparing the Python memory pipeline against its Go port line by line, I found one semantic divergence in how an **unparseable `invalid_at`** is persisted, plus a small timestamp-grammar gap. This PR aligns the Go side with the semantics Python has already committed to.

### The divergence (same LLM output, different stored memory)

Python (`api/db/joint_services/memory_message_service.py`, `extract_by_llm`):

```python
"valid_at": format_iso_8601_to_ymd_hms(extracted_content.get("valid_at", ""), fallback=conversation_time),
"invalid_at": format_iso_8601_to_ymd_hms(extracted_content["invalid_at"], fallback="") if extracted_content.get("invalid_at") else "",
```

- `valid_at` unparseable → falls back to the extraction run time
- `invalid_at` unparseable → falls back to **`""`** ("not invalidated")

Go (`internal/service/memory_extractor.go`, before this PR):

```go
var invalidAt any
if strings.TrimSpace(item.InvalidAt) != "" {
    invalidAt = formatMemoryTime(item.InvalidAt, now) // unparseable -> now
}
```

An `invalid_at` the parser rejects (LLM prose, or an ISO form outside the layout set — e.g. reduced-precision `2026-08-20T10:05`, which `dateutil.parser.isoparse` accepts) became **the extraction run time** — i.e. a still-true memory got persisted as *expired at extraction time*. Python keeps it empty.

### Changes

- `buildExtractedMessage`: parse `invalid_at` first; only a successfully parsed end time is persisted, otherwise it stays `nil` — same as an absent one. `valid_at` keeps its run-time fallback (both the code and `TestBuildExtractedMessageValidAtSemantics` unchanged).
- `formatMemoryTime` is split into a `parseMemoryTime(value) (string, bool)` helper so callers can distinguish "unparseable" from "parsed"; its contract is unchanged (covered by the existing tests).
- The layout set gains the reduced-precision ISO 8601 forms `YYYY-MM-DDTHH:MM` and `YYYY-MM-DD HH:MM` — `dateutil.parser.isoparse` accepts both, so identical LLM output takes the same path on both sides instead of silently falling back on Go.

### Verification

- New/extended tests in `memory_extractor_test.go`: parsed wall clock kept; empty/unparseable `invalid_at` persist as `nil`; reduced-precision forms parse.
- One honest caveat: I could not run the package tests locally — on macOS the build fails for pre-existing environment reasons unrelated to this change (`office_oxide` cgo header lookup, and `syscall.SysProcAttr.Pdeathsig` which is Linux-only in `internal/agent/sandbox`). I verified the touched functions by extracting them verbatim into an isolated package and running the same test cases there (3 failures before the fix, all green after). CI should exercise the real package.

### A question on the Go/Python prompt parity itself

While diffing this subsystem I also noticed the Go extraction prompt (`internal/service/memory.go` `TYPE_INSTRUCTIONS` / `AssembleSystemPrompt`) looks like a port of an older Python revision: the per-type `Examples:` lines, the `Default: valid_at = conversation time, invalid_at = "" for timeless facts` rule, the episodic "extract explicit times" guidance, the procedural fine-print, and Python's whole `**EXAMPLES:**` few-shot block are all absent on the Go side.

So, same question as #18752 left open: **is Python↔Go prompt/semantic sync something you want to track as a property of the memory subsystem (in which case I'm happy to follow up with a parity PR for the template block), or is the Go implementation expected to diverge over time?** If drift is acceptable by design, knowing that would save reviewers future parity PRs.

Refs #18752, #18415.

### DIFF
--- a/internal/service/memory_extractor.go
+++ b/internal/service/memory_extractor.go
@@ -223,7 +223,14 @@ func (s *MemoryMessageService) extractByLLM(ctx context.Context, mem *CreateMemo
 func buildExtractedMessage(messageID, sourceID int64, memoryID string, msg MemoryMessage, item extractedMemory, now time.Time) map[string]any {
 	var invalidAt any
 	if strings.TrimSpace(item.InvalidAt) != "" {
-		invalidAt = formatMemoryTime(item.InvalidAt, now)
+		// Python's decided semantics (api/db/joint_services/memory_message_service.py)
+		// persist an unparseable end time as empty — "not invalidated" — while only
+		// valid_at falls back to the extraction run time. Stamping the run time here
+		// would mark still-true memories expired, so an unparseable invalid_at stays
+		// nil, same as an absent one.
+		if parsed, ok := parseMemoryTime(item.InvalidAt); ok {
+			invalidAt = parsed
+		}
 	}
 	return map[string]any{
 		"message_id":   messageID,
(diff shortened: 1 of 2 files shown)


anneheartrecord:
Closing in favor of #19001, which landed first from a concurrent session working the same Python↔Go comparison and covers the same semantic fix (unparseable `invalid_at` stays empty/nil instead of the run time, with tests).

One residual gap not covered there — reduced-precision ISO 8601 (`dateutil.parser.isoparse` accepts `2026-08-20T10:05`, this PR's layout set rejected it, silently falling back for `valid_at`) — is noted on #19001 as a possible follow-up.

PR #19007, opened 2026-08-30
Fix NameError in audio chunker when file extension is missing or unsupported

blueOf:
### What
`rag/app/audio.py` `chunk()`: `tmp_path` was initialized inside the `try` block, after the extension checks. When a filename has no extension or an unsupported one, the `RuntimeError` is caught and passed to the callback, but the `finally` clause then hits `NameError: name 'tmp_path' is not defined`, which masks the graceful error handling.

Moved the `tmp_path = ""` initialization above the `try` so `finally` always sees a defined variable.

### Testing
Added `test/unit_test/rag/app/test_audio_invalid_extension.py` covering both cases (no extension, unsupported extension); both now expect the error message through the callback instead of a NameError.

Not run locally (missing full dev env), but change is a one-line move and covered by the new tests.

### DIFF
--- a/rag/app/audio.py
+++ b/rag/app/audio.py
@@ -30,6 +30,7 @@ def chunk(filename, binary, tenant_id, lang, callback=None, **kwargs):
 
     # is it English
     is_english = lang.lower() == "english"  # is_english(sections)
+    tmp_path = ""
     try:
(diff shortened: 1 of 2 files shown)


xugangqiang:
LGTM

Issue #19006, opened 2026-08-30
QA chunker: the Go HTML table path pairs across an empty cell where the Python CSV path rejects the row


devYRPauli:
The Go QA chunker builds a question and answer pair from an HTML table row. It keeps the first two non-empty cells. The Python CSV path requires the row to hold exactly two fields. The two rules disagree when a CSV row has an empty middle column.

This follows the review on #17902.

## Where

`internal/ingestion/component/chunker/qa.go:213-222` drops empty cells, then pairs index 0 with index 1:

```go
for _, cell := range cells {
    t := html.UnescapeString(htmlTag.ReplaceAllString(cell[1], ""))
    t = strings.TrimSpace(t)
    if t != "" {
        texts = append(texts, t)
    }
}
if len(texts) >= 2 {
    pairs = append(pairs, qaPair{Question: texts[0], Answer: texts[1]})
}
```

`rag/app/qa.py:365` uses a different rule for CSV:

```python
if len(row) != 2:
    if question:
        answer += "\n" + raw
    else:
        fails.append(str(i + 1))
```

## The difference

Input row: `question,,extra`

| path | result |
|---|---|
| Python CSV, `rag/app/qa.py:358` | `['question', '', 'extra']`, length 3. The row is not a pair. It joins the previous answer, or it counts as a failure. |
| Go `extractQATable` | `['question', 'extra']`. It pairs `question` with `extra`. |

A CSV row with an empty middle column gives a QA pair in Go and no pair in Python.

## Reach

CSV reaches this code. `internal/parser/parser/csv_parser.go:131` and `:160` set `OutputFormat: "html"`, and `qa.go:114` calls `extractQATable` for the HTML format.

XLSX and the legacy XLS path reach the same function. `internal/parser/parser/xlsx_parser.go:134` and `internal/parser/parser/xls_parser.go:116` both set `OutputFormat: "html"`.

TXT does not reach it. `internal/parser/parser/text_parser.go:68` sets `OutputFormat: "json"`, so `qa.go` uses `extractQAJSON` for that format. Python applies its strict two-field rule to TXT as well, at `rag/app/qa.py:312`, so the two sides agree there.

That matters for the fix, because Python does not use one rule for both formats:

- `rag/app/qa.py:53`, the Excel path, skips empty cells and takes the first two values. Python routes both `.xls` and `.xlsx` there, because `qa.py:305` matches `\.xlsx?$`. The Go code matches Python here.
[...]

PR #19008, opened 2026-08-30
Require exactly two cells for a CSV QA pair

devYRPauli:
### Summary

Closes part of #19006. Follows the review on #17902.

## Problem

`extractQATable` drops empty cells, then pairs `texts[0]` with `texts[1]`:

```go
if t != "" {
    texts = append(texts, t)
}
...
if len(texts) >= 2 {
    pairs = append(pairs, qaPair{Question: texts[0], Answer: texts[1]})
}
```

A CSV row `question,,extra` renders as `<td>question</td><td></td><td>extra</td>`, because `office_table_render.go:238` writes one cell per field and keeps the empty ones. The pair becomes `question` and `extra`. The empty middle column is skipped without notice.

Python does not apply one rule to every format:

| source | Python | rule |
|---|---|---|
| `.csv` | `rag/app/qa.py:365` | the row must hold exactly two fields |
| `.xls`, `.xlsx` | `rag/app/qa.py:53` | skip empty cells, take the first two values |

## Which formats reach this function

`csv_parser.go` and `xls_parser.go` emit `OutputFormat: "html"`, so they reach `extractQATable`. `xlsx_parser.go` no longer does. #18800 changed its return from `"html"` to `"json"` on 2026-08-27, so an `.xlsx` file now reaches `extractQAJSON`.

This change is therefore CSV-only in effect, plus `.xls`, which keeps its current behaviour. It is a no-op for `.xlsx`.

## Change

`extractQATable` takes a `strictPairs` flag. When it is set, a row makes a pair only if it holds exactly two cells. The call site sets it from the source document name, the same signal Python dispatches on.

`.xls` behaviour does not change.

## Scope

This stops the incorrect pairing on the HTML table path. Two things stay out:

- Python's continuation handling, where a row without exactly two fields appends to the previous answer (`qa.py:366-369`). The Go text paths already do this; the table path does not.
- The `.xlsx` JSON path. It is broken in a different way, reported separately in #19174.

## Tests

Three tests in `qa_test.go`, all driving the component through `Invoke`:

- `TestQAChunker_CSVStrictPairRejectsThreeCells` - a `.CSV` source with `question`, empty, `extra` produces no chunk. The name is uppercase on purpose, to cover the case-insensitive match.
- `TestQAChunker_CSVStrictPairAcceptsTwoCells` - a normal two-column CSV row still pairs.
- `TestQAChunker_NonCSVHTMLThreeCellsKeepsFirstTwo` - the same three-cell row on an `.xls` source still pairs `question` with `extra`, so the non-CSV path is unchanged.

Results:

```
CGO_ENABLED=0 go test ./internal/ingestion/component/chunker/
ok  	ragflow/internal/ingestion/component/chunker	1.124s

CGO_ENABLED=0 go vet ./internal/ingestion/component/chunker/
(no output)
```

Reverting only the source guard and keeping the tests fails one test, and only the intended one:

```
--- FAIL: TestQAChunker_CSVStrictPairRejectsThreeCells
    qa_test.go:147: expected 0 chunks, got 1
--- PASS: TestQAChunker_CSVStrictPairAcceptsTwoCells
--- PASS: TestQAChunker_NonCSVHTMLThreeCellsKeepsFirstTwo
```

## Note on the local build

`go build ./internal/...` needs `CGO_ENABLED=0` on macOS, or it stops on a missing `office_oxide.h`. With cgo off, one unrelated package still fails on this platform:

```
internal/agent/sandbox/local.go:307:3: unknown field Pdeathsig in struct literal of type "syscall".SysProcAttr
```

`Pdeathsig` is Linux only and that file carries `//go:build !windows`. The same failure appears on a clean `main`, so it is not from this change.

### DIFF
--- a/internal/ingestion/component/chunker/qa.go
+++ b/internal/ingestion/component/chunker/qa.go
@@ -111,7 +111,7 @@ func (c *QAChunkerComponent) invoke(_ context.Context, inputs map[string]any) (m
 	var isMarkdown bool
 	switch upstream.OutputFormat {
 	case schema.PayloadFormatHTML:
-		qaPairs = extractQATable(stringPtrVal(upstream.HTMLResult))
+		qaPairs = extractQATable(stringPtrVal(upstream.HTMLResult), isCSV(upstream.Name))
 	case schema.PayloadFormatMarkdown:
 		qaPairs = extractQAMarkdown(stringPtrVal(upstream.MarkdownResult))
 		isMarkdown = true
@@ -193,6 +193,10 @@ func stringPtrVal(s *string) string {
 	return *s
 }
 
+func isCSV(name string) bool {
+	return strings.HasSuffix(strings.ToLower(name), ".csv")
+}
+
 // ---------------------------------------------------------------------------
 // HTML / spreadsheet QA extraction
 // ---------------------------------------------------------------------------
@@ -201,7 +205,7 @@ var htmlTR = regexp.MustCompile(`(?i)<tr[^>]*>(.*?)</tr>`)
 var htmlTD = regexp.MustCompile(`(?i)<t[dh][^>]*>(.*?)</t[dh]>`)
 var htmlTag = regexp.MustCompile(`<[^>]+>`)
 
-func extractQATable(htmlStr string) []qaPair {
+func extractQATable(htmlStr string, strictPairs bool) []qaPair {
 	if htmlStr == "" {
 		return nil
 	}
@@ -209,6 +213,10 @@ func extractQATable(htmlStr string) []qaPair {
 	pairs := make([]qaPair, 0, len(rows))
 	for _, row := range rows {
 		cells := htmlTD.FindAllStringSubmatch(row[1], -1)
+		// Python qa.py:365 requires exactly two fields for CSV pairs.
+		if strictPairs && len(cells) != 2 {
+			continue
+		}
 		var texts []string
 		for _, cell := range cells {
 			t := html.UnescapeString(htmlTag.ReplaceAllString(cell[1], ""))
--- a/internal/ingestion/component/chunker/qa_test.go
+++ b/internal/ingestion/component/chunker/qa_test.go
@@ -128,6 +128,74 @@ func TestQAChunker_HTMLTable(t *testing.T) {
 	}
 }
 
+func TestQAChunker_CSVStrictPairRejectsThreeCells(t *testing.T) {
+	comp, err := NewQAChunker(map[string]any{"lang": "english"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	inputs := map[string]any{
+		"name":          "test.CSV",
+		"output_format": "html",
+		"html":          "<table><tr><td>question</td><td></td><td>extra</td></tr></table>",
+	}
+	out, err := comp.Invoke(context.Background(), nil, inputs)
+	if err != nil {
+		t.Fatalf("Invoke failed: %v", err)
+	}
+	chunks, _ := out["chunks"].([]map[string]any)
+	if len(chunks) != 0 {
+		t.Fatalf("expected 0 chunks, got %d", len(chunks))
+	}
+}
+
+func TestQAChunker_CSVStrictPairAcceptsTwoCells(t *testing.T) {
+	comp, err := NewQAChunker(map[string]any{"lang": "english"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	inputs := map[string]any{
+		"name":          "test.csv",
+		"output_format": "html",
+		"html":          "<table><tr><td>question</td><td>answer</td></tr></table>",
+	}
+	out, err := comp.Invoke(context.Background(), nil, inputs)
+	if err != nil {
+		t.Fatalf("Invoke failed: %v", err)
+	}
+	chunks, _ := out["chunks"].([]map[string]any)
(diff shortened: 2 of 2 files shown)


devYRPauli:
The checks on this PR all report SKIPPED. `tests.yml` gates every job on the `ci` label:

```yaml
if: ${{ github.event_name != 'pull_request' || (github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'ci') && (github.event.action != 'labeled' || github.event.label.name == 'ci')) }}
```

This PR carries `bug`, `size:S` and `test`. All 9 check runs skip, which reads like a pass in the checks list.

I cannot add the label myself. Please add `ci` when you want the suite to run.

Local results are in the PR body. `go test ./internal/ingestion/component/chunker/` passes. Reverting the source guard and keeping the tests fails one test, and only the intended one.

xugangqiang:
Thanks for the fix. The change is minimal and correctly aligns CSV QA pairing with Python (`qa.py:366` requires exactly two fields). I reviewed it against current `main` and have a few findings plus one question for you.

## Findings

**1. PR description premise is stale relative to current `main` (Improvement, non-blocking)**
The description states `csv_parser.go`, `xlsx_parser.go`, and `xls_parser.go` all emit `OutputFormat: "html"` and share `extractQATable`. That was true when the PR was opened (2026-08-30), but `xlsx_parser.go` was introduced by #19083 (2026-09-01) and now returns `OutputFormat: "json"` (`xlsx_parser.go:152`). So on current `main`, XLSX goes through `extractQAJSON`, **not** `extractQATable`.
- Consequence: this change is effectively **CSV-only** (+ XLS, which still emits `html`). For XLSX it is a no-op because the function is never reached.
- Please update the description so reviewers don't assume "XLSX behavior unchanged" is a guarantee about the XLSX production path.

**2. The XLSX regression test is synthetic (Nit)**
`TestQAChunker_XLSXThreeCellsKeepsFirstTwoNonEmpty` manually sets `"output_format": "html"` with a `test.xlsx` name. It validates the non-CSV HTML branch, not the real XLSX path (which is JSON). Consider renaming it to something like `TestQAChunker_NonCSVHTMLThreeCellsKeepsFirstTwo` to avoid implying it exercises the XLSX production flow.

**3. Known out-of-scope gaps (not introduced by this PR, just noting)**
- Python's CSV continuation handling (`qa.py:367-369`: a non-2-field row with a pending question is appended to the answer) is not implemented here — correctly scoped out, but a follow-up.
- A two-cell row with an empty answer (e.g. `question,`) is still paired in Go, whereas Python keeps it pending. Pre-existing behavior, not changed here.

## Question for the author

Since #19083, **XLSX now outputs `json`** and the QA chunker handles it via `extractQAJSON` instead of `extractQATable`. This PR only adjusts the HTML/table path. Do you know whether the JSON path for XLSX already applies the Python-consistent rule (skip empty cells, take the first two), or does it need its own adaptation to match Python? If the JSON path has the same "wrong pairing" bug for XLSX, it may be worth a follow-up PR. Could you confirm?

devYRPauli:
**1. Stale description.** Corrected. I also traced the change: `xlsx_parser.go` dates to #16647, and #18800 flipped its return from `"html"` to `"json"` on 2026-08-27. #19083 is not the commit.

```
git log -L 150,155:internal/parser/parser/xlsx_parser.go
82e363919 feat: extract images from xlsx parser (#18800)
-		OutputFormat: "html",
+		OutputFormat: "json",
```

The description now states that this change is CSV-only in effect, plus `.xls`, and a no-op for `.xlsx`.

**2. Synthetic XLSX test.** Renamed to `TestQAChunker_NonCSVHTMLThreeCellsKeepsFirstTwo` and switched to an `.xls` name, which still emits `"html"`. Pushed as 2da6414.

**3. Out-of-scope gaps.** Agreed on both. One correction on the first: the Go text paths already implement Python's continuation rule. `extractQATextTab` and `extractQATextCSV` both require exactly two fields and append a non-conforming row to the pending answer. Only the table path lacks it.

**Your question.** The XLSX JSON path does not apply the Python rule, and it does not pair rows wrongly either. It produces nothing at all.

`xlsx_parser.go:152` puts the rendered HTML table in each item's `text` field and returns `"json"`. `extractQAJSON` sends that text to `extractQAText`, which splits on newlines, runs `detectDelimiter`, then parses the string as CSV or tab-separated text. An HTML `<table>` string holds no two-field record, so no pair is built.

Measured with one spreadsheet driven through the real parser and the real chunker:

| commit | parser output_format | QA chunks |
|---|---|---|
| `4168b3f9d` (main) | `json` | 0 |
| `6c2bca893` (parent of #18800) | `html` | 3 |

The rendered HTML is byte-identical in both. Only the route changed, and the chunker was never adapted to it.

So this needs its own fix, and it is more serious than the pairing gap. I filed it as #19174 with the reproduction test and the bisect.

Two repairs are possible. `extractQAJSON` can detect a table item by `doc_type_kwd == "table"` and send its text to `extractQATable`. Or `xlsx_parser.go` can keep `"html"` and carry the images separately. The first is smaller and keeps the image work from #18800. Tell me which you want and I will send it.

**CI.** The go jobs were red for a separate reason. #19108 moved this package to `t.Context()` and dropped the `"context"` import. The merge of this branch with main then had three `context.Background()` calls and no import. Fixed in f080b22.

xugangqiang:
LGTM

PR #19009, opened 2026-08-30
feat(web): complete the Turkish locale and fix broken placeholders

bakiburakogun:
### Summary

The Turkish locale had fallen 755 keys behind `en.ts`, so large parts of the UI rendered in English — the whole `knowledgeCompilation` section, every data source connector field and tip, the chat channel bots, the somark parser options, the Querit / Keenable / You.com / BGPT / Browser flow nodes, and the thinking-level controls in chat. This brings `tr.ts` to full parity with `en.ts` and repairs a number of strings that were broken rather than merely missing.

**Broken strings**

- `flow.prompts.system.keywords` and `.questions` are prompt templates sent to the LLM. The Turkish versions had replaced the `{{ topn }}` variable with a hardcoded `5` and `3`, so the caller's requested count was silently ignored. They also opened with "Siz bir metin analistisisiniz", which is not a word.
- `flow.invalidUrl` translated the literal placeholder names the user is supposed to type, instructing them to write `{değişken_adı}` instead of `{variable_name}`.
- `chat.variableTip` and `flow.templateDescription` had been cut down to their first sentence, dropping the explanation of `{knowledge}` and of the `{parameter}` substitution syntax.
- `chat.chatConfigurationDescription` read "Seçilen datasets için", with the English plural left in.

**Existing strings that shouted or drifted**

Nine labels rendered `API ANAHTARI` in full caps where the source is the title-case "API Key", and several messages mixed cases mid-sentence, e.g. `Lütfen Tavily API ANAHTARInızı girin`. A few strings sharing one English source had drifted apart (`Erişim Tokenı` / `Erişim Token'ı`, `Algoritma Seç` / `Algoritma Seçin`) and now agree.

**Terminology**

New strings follow what the file already uses: `maksimum`, `arka uç`, `Kimliği` for ID, `API Anahtarı`, `Ana bilgisayar` for Host, `Depo` for Repository, `Parça` for Chunk, `Ajan` for Agent, `meta veri` for metadata.

One note for maintainers: **"dataset" is left untranslated throughout**, because that is how the existing locale reads — 45 of its 46 occurrences keep the English word. Translating it to "veri kümesi" would be more natural Turkish but would touch a lot of already-merged strings, so I kept consistency here rather than making that call unilaterally. Happy to do it as a separate PR if you would prefer the translated term.

No key was removed, every placeholder and ICU argument is preserved, and `tr.ts` still parses and formats cleanly.


bakiburakogun:
Notes on the automated review findings, for whoever picks this up — five applied, two not.

**Applied**

- `common.searchOrEnterToAdd` — `girin` reads as "type something" and hides the key. Now `Arayın veya eklemek için Enter'a basın`.
- `chat.queritApiKeyTip` / `chat.serplyApiKeyTip` — the locale already uses `alım` for retrieval 22 times; `getirme` was my inconsistency. Both now say `dataset alımını tamamlar`. Swept the rest of the diff for the same slip and fixed one more, `knowledgeCompilation.rechunkByTreeLeavesTip`.
- `setting.aimlapiAuthExpired` — expiry and timeout are different states. Now `İsteğin süresi doldu.`, keeping the request as the subject to match "The request expired" rather than moving it to authorization.
- `knowledgeCompilation.deleteStructureConfirm` — Turkish needs the accusative and it cannot attach to a variable. Now `{{name}} öğesini silmek istediğinizden emin misiniz?`.
- `knowledgeCompilation.compilationTitleSuffix` — there is a stray apostrophe, though the suggested `' dataset'i` would still render `Ad' dataset'i` with the leading one intact, so I used ` dataset'i`, giving `Ad dataset'i`. Worth noting the English value is `' dataset`, which renders `Ad' dataset` in the source too — happy to leave the Turkish mirroring that if you'd rather the two stay identical.

**Not applied**

- `knowledgeConfiguration.compilationTemplate` → `Derleme şablonu`. The English value for this key is `Operator`, not "template" — the key name says template but the visible label does not. `flow.operator` is also `Operator` and already translated `Operatör`, so `Operatör` is what makes the two agree. Translating the label as "şablon" would make the Turkish say something the English does not.
- `chat.thinkingLevelNone` → `Yok`. The English value here is `Naive`, not "none". There is a separate `chat.thinkingDisabled` (`Devre dışı`) for the off state, so this is a real reasoning level rather than a way to disable thinking. `Yalın` renders "naive/plain", which is what the source says.

**Already correct**

On the `topn` prompts: `flow.prompts.system.keywords` and `.questions` both keep `{{ topn }}` in this PR. Restoring it was one of the fixes here — the previous Turkish had replaced the variable with a hardcoded `5` and `3`, so the caller's requested count was silently ignored.

Issue #3436, opened 2024-11-15
[Question]: How to solve "ERROR: 'Stream' object has no attribute 'iter_lines'


duraiyuva:
### Describe your problem

Getting the error "ERROR: 'Stream' object has no attribute 'iter_lines'" for chat response. No log or module error. Anthropic API access is successful. But the response is not displayed.

midilandlu:
Same error message 

RAGFLow : System a19210da(v0.13.0~269) slim

Model Providers - System Model Settings - chat : claude-3-haiku-20240307

While use Search Function , input and start search , get the error message - ERROR: 'Stream' object has no attribute 'iter_lines'

![Screenshot 2024-11-30 at 12 26 15 AM](https://github.com/user-attachments/assets/31feb211-a736-4c48-8b52-69cbdd35cbb5)

log 
![Screenshot 2024-11-30 at 12 31 03 AM](https://github.com/user-attachments/assets/72f559ef-12ec-40bd-80ec-b8ef8c6eb84b)

midilandlu:
RAGFLow : System a19210da(v0.13.0~269) slim

Model Providers - System Model Settings - chat : claude-3-haiku-20240307 
Using claude-3-haiku-20240307  for RAPORT in Knowledge base document , it work

![Screenshot 2024-11-30 at 12 51 42 AM](https://github.com/user-attachments/assets/84c74dca-2af2-49a8-b743-c2f6deb08827)

chocolate-thunder-shadow:
I am also having the same issue with all anthropic models, commit https://github.com/infiniflow/ragflow/commits/0d5486aa57062da4ae2c0f84a3505e94c4044c0b

awilmoth:
I'm having the same problem.

Yannnnnnny:
Thanks for the update.
Just checking in — are you still experiencing this issue on the latest version? If the problem has been resolved, we can go ahead and close this issue. Otherwise, please let us know and we'll continue investigating.

PR #19013, opened 2026-08-30
Add Slovak & Czech language dataset support

martincivan:
### What

Slovak and Czech datasets are effectively BM25-blind today. The tokenizer only keeps ASCII letter runs whole, so every accented word is fragmented before indexing (`škola` -> `š kola`) and the keyword leg of hybrid search never matches. Neither language has a Snowball stemmer either, so the analyzer inherits whatever stemmer the previously processed dataset left behind.

### Approach

Per review, the tokenizer change lives in Infinity: infiniflow/infinity#3436 adds `slovak`/`czech` to `rag_tokenizer.py` (fold diacritics to ASCII, no stemming) and to the C++ `RAGAnalyzer` (`rag-slovak`, `rag-czech`, `rag-*-fine` analyzer names). This PR only makes RAGFlow apply the dataset language consistently on both sides of the index; `rag/nlp/rag_tokenizer.py` is untouched.

- Query side tokenizes with the dataset language: `FulltextQueryer.question()`, `Dealer.retrieval()` and the rerank family take `language`; callers pass it (chat, dify retrieval, chunk/dataset APIs, retrieval test, agentic search tools via `RAGTools.language`, tag queries with the tag datasets' language).
- Under `DOC_ENGINE=infinity` nothing changes: `tokenize()` stays a passthrough and the query is not folded, so Slovak queries keep matching today's Infinity index. Selecting `rag-slovak` analyzers for Infinity tables is a follow-up.
- `rag.nlp.dataset_language()` / `entity.KnowledgebasesLanguage()` resolve the language of a multi-dataset search only when every dataset agrees; a mixed set retrieves as English, as before.
- Go path: the vendored C++ analyzer mirrors the Infinity change (folding + no stemming in `SetLanguage`, offsets mapped back to the unfolded input), and retrieval requests carry the KB language into `QueryBuilder.Question`, `TermWeightDealer` and the rerankers.
- Slovak and Czech added to the dataset language options in the UI.

Empty language keeps today's English behaviour everywhere. Existing Slovak/Czech datasets need a re-parse to get folded index tokens.

### Depends on

- infiniflow/infinity#3436 merged and released; `infinity-sdk` pin bumped to that release (the tokenizer tests in `test/unit_test/rag/nlp/test_rag_tokenizer_language.py` exercise the SDK behaviour and fail with 0.7.3).

### Tests

- `test/unit_test/rag/nlp/test_rag_tokenizer_language.py`, `test/unit_test/rag/advanced_rag/test_agentic_dataset_language.py`
- `internal/tokenizer/tokenizer_test.go` (Slovak/Czech analyzer output), `internal/service/nlp/query_builder_test.go`, `internal/entity/dataset_language_test.go`

### Measured

On a Slovak-language production corpus (Elasticsearch, hybrid search, document-level hit@10 over an internal question set), symmetric folding improved the keyword-dominant configuration from 25% to 32% and the balanced one from 40% to 50%. Folding only the index side regressed accented queries, which is why index and query must fold together.


rag/nlp/rag_tokenizer.py
qinling0210, line 1: Full rag_tokenizer.py is in Infinity (https://github.com/infiniflow/infinity/blob/main/python/infinity_sdk/infinity/rag_tokenizer.py), we import the file from infinity-sdk in ragflow.

We should not modify this file. I suggest you submit a PR to modify rag_tokenizer.py in Infinity.

rag/nlp/query.py
```diff
@@ -39,9 +39,13 @@ def __init__(self):
...
 
-    def question(self, txt, tbl="qa", min_match: float = 0.6):
+    def question(self, txt, tbl="qa", min_match: float = 0.6, language: str | None = None):
         original_query = txt
-        txt = self.add_space_between_eng_zh(txt)
+        # Keep query tokenization symmetric with indexing, which calls
+        # set_language() per dataset (see rag/nlp/__init__.py). Folding up
+        # front makes the heuristics below see the same form as index tokens.
+        rag_tokenizer.tokenizer.set_language(language or "English")
+        txt = rag_tokenizer.tokenizer.fold_text(self.add_space_between_eng_zh(txt))
```
qinling0210, line 48: question() folds unconditionally, before the engine check.

For Infinity, the index side is not folded: rag_tokenizer.tokenize() returns line unchanged under DOC_ENGINE_INFINITY (pre-existing), and Infinity tokenizes content_with_weight with its own C++ analyzer, which has no folding. The folded matchText is then used verbatim in Dealer.search's Infinity branch (matchExprs = [matchText, ...]). For a Slovak query škola:

pre-PR: query → š kola, matches the index fragments → works
post-PR: query → skola, cannot match š/kola or škola in the Infinity index → regression

We should not fold for Infinity here in RAGFlow, and we need to add Czech/Slovak diacritic folding support to Infinity.

qinling0210 (changes requested):
@martincivan Please check the comments,  we should first modify in Infinity to add language support in RAGFlow,. 

Could you please add Slovak & Czech language support in  https://github.com/infiniflow/infinity?

Then after we release a new version of Infinity, we can merge this PR in RAGFlow.

martincivan:
@qinling0210 thanks for the review. I changed the approach as suggested:

- The tokenizer change now lives in Infinity: infiniflow/infinity#3436 adds `slovak`/`czech` to `rag_tokenizer.py` (fold diacritics to ASCII, no stemming) and to the C++ `RAGAnalyzer`, plus `rag-slovak` / `rag-czech[-fine]` analyzer names in `AnalyzerPool`. Python and C++ produce byte-identical output for the same input.
- This PR no longer touches `rag/nlp/rag_tokenizer.py`. It only passes the dataset language to the query side so index and query tokens agree, and under `DOC_ENGINE=infinity` nothing is folded (`tokenize()` stays a passthrough), so Slovak queries keep matching the current Infinity index.
- Rebased onto main and squashed into two commits (Python + web, Go/C++ parity).

It depends on the Infinity release that includes #3436; I will bump the `infinity-sdk` pin once it is out.

rag/nlp/rag_tokenizer.py
martincivan: Moved to Infinity: infiniflow/infinity#3436. This PR no longer modifies rag_tokenizer.py.

rag/nlp/query.py
martincivan: Fixed: question() no longer folds the query itself. It only calls set_language(), and folding happens inside the SDK's tokenize(), which stays a passthrough under DOC_ENGINE=infinity, so nothing is folded for the Infinity engine. Server-side folding for Infinity is in infiniflow/infinity#3436 (rag-slovak / rag-czech analyzers).

test/unit_test/rag/nlp/test_rag_tokenizer_language.py
```diff
@@ -0,0 +1,102 @@
...
+
+from common import settings
+from rag.nlp import dataset_language, rag_tokenizer
+
+
+@pytest.fixture(autouse=True)
+def non_infinity_engine(monkeypatch):
+    # tokenize() is a passthrough under DOC_ENGINE=infinity (tokenization
+    # happens server-side); force the local tokenizer path.
+    monkeypatch.setattr(settings, "DOC_ENGINE_INFINITY", False, raising=False)
```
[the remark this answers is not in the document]
martincivan: Intentional: per the maintainer review, rag_tokenizer.py stays untouched in RAGFlow and the Slovak/Czech folding ships in infinity-sdk (infiniflow/infinity#3436). These tests exercise that SDK behaviour and skip on 0.7.3; the pin will be bumped to the release that includes it before merge.

test/unit_test/rag/nlp/test_rag_tokenizer_language.py
```diff
@@ -0,0 +1,113 @@
...
+import infinity.rag_tokenizer
+import pytest
+
+from common import settings
+from rag.nlp import dataset_language, rag_tokenizer
+
+# Slovak/Czech folding is implemented in infinity-sdk; skip on releases
+# that predate it instead of failing on the pinned version.
+_SDK_FOLDS = hasattr(infinity.rag_tokenizer, "fold_diacritics")
+needs_folding_sdk = pytest.mark.skipif(not _SDK_FOLDS, reason="infinity-sdk without Slovak/Czech diacritics folding")
```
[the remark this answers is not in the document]
martincivan: Agreed on the pin: it will move to the infinity-sdk release that includes infiniflow/infinity#3436 once it is published (this PR is blocked on that release anyway). Until then the skip keeps CI meaningful for the rest of the suite; with the pin bumped the skipif condition is always false, so the tests run unconditionally.

internal/binding/cpp/rag_analyzer.cpp
```diff
@@ -887,6 +898,71 @@ std::string RAGAnalyzer::StrQ2B(const std::string &input) {
...
+            if (codepoint >= 0xC0 && codepoint < 0x180) {
+                if (char folded = DIACRITICS_FOLD_TABLE[codepoint - 0xC0]; folded != '\0') {
+                    output += folded;
+                    i += 2;
+                    continue;
+                }
+            }
+            output += input[i];
+            output += input[i + 1];
+            i += 2;
```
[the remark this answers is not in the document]
martincivan: Kept as is, in lockstep with the Infinity analyzer (infiniflow/infinity#3436): both fold only precomposed U+00C0–U+017F letters and leave combining marks alone, so Python, Go and Infinity produce identical tokens. Decomposed (NFD) input is not produced by RAGFlow's parsers, and removing marks would need a different position-mapping scheme than BuildPositionMapping's per-character alignment. Worth a follow-up in Infinity first if decomposed input shows up in practice.

yingfeng:
Hi, your change works well for Elasticsearch, but less so for Infinity, since the analyzer is configured in infinity_mapping.json while the tokenizer runs outside Elasticsearch. Could you provide a dynamic way to select the language for the Infinity Engine? Thanks a lot!

martincivan:
@yingfeng You're right, the language never reached the Infinity path. Addressed in 9402f9bce and fabdc5ffe.

Under `DOC_ENGINE=infinity` the tokenizer is a passthrough and all analysis is server-side, using the fixed `["rag-coarse", "rag-fine"]` pair from `conf/infinity_mapping.json`. The dynamic hook is that Infinity analyzes a query with the analyzer of the fulltext index it matches (`GetAnalyzerName`), and chunk tables are per-dataset — so the language can be chosen when the table is created and the query side inherits it, keeping index and query symmetric. A Slovak dataset now indexes and queries `content` through `rag-coarse-slovak` / `rag-fine-slovak`; `whitespace-#` and `rankfeatures` fields are untouched. The `rag-<language>` names come from infiniflow/infinity#3436.

Two things you may want to look at:

1. **The language has to reach both creation paths.** `create_idx` is the obvious one, but the Go ingestion pipeline never calls `CreateChunkStore` — it writes through `InsertChunks`, whose fallback creates the table on first write. `InsertChunks` now carries the language too, read from `taskCtx.KB`, otherwise the feature was a no-op under `--ingestor`.

2. **A table's analyzer is settled when it is created.** `create_index` ignores by *name*, so a second call under a different language used to add a competing index instead of leaving the table alone — and since Infinity picks a field's first index by name, `ft_content_rag_coarse` would beat `ft_content_rag_coarse_slovak` and silently discard the dataset's language. All three index-creation paths now keep whatever rag analyzer the table already has. Consequence: an Infinity dataset created before this change keeps its current analyzer, so a Slovak or Czech one has to be reindexed into a fresh table; changing the language setting alone will not do it.

Only Slovak and Czech are suffixed, since they are the only languages whose analyzer differs from the default — every other name either picks a Snowball stemmer or is a no-op.

**On merge ordering:** on the currently pinned `infinity-sdk==0.7.3` / `infinity:v0.7.3`, `AnalyzerPool` matches the suffix with `strcmp(str, "-fine")`, so `rag-fine-slovak` resolves to *coarse* — two coarse indexes and no fine-grained one, silently. This needs #3436 released and the pins bumped first, which is the gate the PR already carries.

Folding is deliberately the whole of this PR. I plan to add proper Slovak/Czech lemmatization and stemming in a follow-up PR, once this and the Infinity side have landed.

Tests: `test/unit_test/common/test_infinity_analyzer_language.py` and `internal/engine/infinity/analyzer_test.go` cover analyzer selection, index naming and the cross-language guard in both directions; the unit tree and `./internal/{engine,service,ingestion}/...` pass locally.

(The red `ragflow_tests_infinity (go)` job is the `admin server not connected` startup flake — all 280 errors are that one message, no assertion failures, and the same job fails on `main`'s scheduled run. I have retriggered it.)

yingfeng (approved):

martincivan:
@qinling0210 infiniflow/infinity#3436 (Slovak/Czech folding in rag_tokenizer.py and the RAG analyzer) is merged. 

The last step is bumping `infinity-sdk` to a release that contains that change, but PyPI is still at 0.7.3. Is there a planned date for the next infinity-sdk release? As soon as it is out I will update the pin here,

JinHai-CN:
@martincivan would you please resolve the conflicts?

martincivan:
@JinHai-CN resolved

JinHai-CN:
@martincivan

1. please resolve the conflicts
2. please fix the test:
```
# ragflow/internal/service/nlp [ragflow/internal/service/nlp.test]
internal/service/nlp/datasetnav_test.go:470:55: not enough arguments in call to eng.InsertChunks
```

martincivan:
@JinHai-CN Both done — rebased onto current `main`, still 5 commits.

**1. Conflicts.** Only two files conflicted; the rest applied clean.

- `internal/handler/tenant.go` — `main` deleted `InsertChunksFromFile` (and `InsertMetadataFromFile`) entirely, while this branch was only adding a language argument to the `InsertChunks` call inside it. I took the deletion; the file is now byte-identical to `main`, with no wrapper kept behind.
- `test/unit_test/api/apps/services/test_dataset_api_service_nav.py` — `main` rewrote the file (compiled-row helpers replaced by `_claim_row`, several tests renamed or dropped). Took the upstream file and re-applied this branch's only real change, `language=None` on the remaining `kb = SimpleNamespace(...)` stubs.
- `internal/ingestion/knowledge_compile/fake_engine_test.go` — `main` gave `fakeEngine.InsertChunks` a recording body while this branch changed its signature; kept both.

**2. The test.** Reproduced exactly as you reported, then fixed:

```
internal/service/nlp/datasetnav_test.go:498:55: not enough arguments in call to eng.InsertChunks
	have (Context, []map[string]interface{}, string, string)
	want (Context, []map[string]interface{}, string, string, string)
```

`InsertChunks` gains the `language` parameter this PR adds. The two call sites (`:498`, `:596`) now pass `""`, which is the correct value rather than filler: they seed the dataset *navigation* index, which is not a dataset chunk table, and production `datasetnav.go` (lines 579/618/640/963) together with the already-updated `datasetnav_test.go:770` pass `""` for exactly that reason.

While fixing it I swept every package for the same staleness and found three more call sites, all in `integration`-tagged files that a default `go test ./...` does not reach — two of which this PR had missed:

- `internal/service/dataset_artifact_keyword_integration_test.go:70,71` — `CreateChunkStore` (5→6 args) and `InsertChunks` (4→5)
- `internal/engine/oceanbase/integration_test.go:69,136` — two `InsertChunks` calls

Also fixed: `test/unit_test/rag/test_rerank_by_model_input.py` (new on `main` since this PR opened) carries a local `_Queryer` stub whose `question(self, query)` broke on the new `language=` keyword — 8 tests failing with `TypeError`.

**One additional fix, same defect class as `fix(infinity): apply the dataset language where the chunk store is created`.** The Python `insert()` table-creation fallback in `rag/utils/infinity_conn.py` called `create_idx()` without a language, the exact twin of the Go `InsertChunks` fallback that commit repairs. Since the index-creation loops now preserve whatever rag analyzer a table already has, a table created by that fallback before the first parse task would freeze the default analyzer permanently for a Slovak or Czech dataset. It now resolves the dataset language from `knowledgebase_id` and passes it through. The shared `insert()` interface is untouched; only the rag-side connector changed, since the memory-side table has no dataset language.

**Verification**

```
bash build.sh --test ./internal/...                     # unit tier, no failures
bash build.sh --test ./internal/service/nlp/...         # ok  ragflow/internal/service/nlp
pytest test/unit_test --ignore=test/unit_test/sdk       # 4758 passed, 46 skipped
ruff check / ruff format --check                        # clean on all changed files
```

The `integration` and `e2e` tiers were also compiled (`-run "^$"`); the packages touched here build. A few unrelated packages fail to build in those tiers on `main` itself — I confirmed byte-identical errors on a clean `main` worktree, so they are pre-existing and not from this PR.

The 20 skips in `test_rag_tokenizer_language.py` are expected and documented in the PR: they need the infinity-sdk release carrying the Slovak/Czech diacritics folding, and the currently published SDK predates it.

JinHai-CN:
@martincivan new conflicts need to resolve and CI testings are failed.

```
=========================== short test summary info ============================
FAILED test/testcases/restful_api/test_retrieval_metadata_condition_unit.py::test_empty_metadata_conditions_do_not_filter - ImportError: cannot import name 'dataset_language' from 'rag.nlp' (unknown location)
FAILED test/testcases/restful_api/test_retrieval_metadata_condition_unit.py::test_empty_metadata_conditions_keep_explicit_document_ids - ImportError: cannot import name 'dataset_language' from 'rag.nlp' (unknown location)
FAILED test/testcases/restful_api/test_retrieval_metadata_condition_unit.py::test_nonempty_metadata_condition_with_no_match_returns_empty - ImportError: cannot import name 'dataset_language' from 'rag.nlp' (unknown location)
==== 3 failed, 284 passed, 162 skipped, 81 deselected, 1 xfailed in 53.73s =====
```

martincivan:
Rebased onto current `main` (`839a4ec81`) and CI is green.

**Conflicts.** Upstream landed the `allow_dense_fallback` retrieval flag into the same argument lists this branch adds `language=` to, so every conflict was the same shape: a sibling keyword added at the same anchor. Resolved by keeping both, in `rag/nlp/search.py` (signature, the `req` dict, and the low-`min_match` retry where upstream also clones `matchDense`), `rag/advanced_rag/agentic_rag.py`, `rag/advanced_rag/harness/tools/navigation.py`, `rag/advanced_rag/harness/tools/search.py`, and `internal/service/nlp/retrieval.go`.

**The failing tests.** The three failures in `test/testcases/restful_api/test_retrieval_metadata_condition_unit.py` were mine. That module builds a stub module tree and creates `rag.nlp` as a bare `ModuleType` with only `__path__`, so its `__init__.py` never executes and `from rag.nlp import dataset_language` could not resolve. Fixed by giving the stub the attribute, in the idiom of the `nlp_pkg.search = SimpleNamespace(...)` line directly above it:

```python
nlp_pkg.dataset_language = lambda kbs: None
```

`None` is the correct value rather than a silencer: the module's kb fixture carries no `language`, and the real resolver returns `None` for exactly that input.

**One commit that is not mine — `test(knowledge_compile): match recordingNavService to the NavService interface`.** This is unrelated to Slovak/Czech support and I would normally leave it out, but `internal/ingestion/knowledge_compile` does not compile on `main` as it stands, so the Go job cannot pass on any branch until it is repaired. Two changes crossed: one added a `keywords` parameter to `NavService.ListClusters` and `ListChildren`, the other edited `nav_lifecycle_test.go` without updating the `recordingNavService` stub. Since `nav.SetNavService` takes the interface, the package fails to build:

```
FAIL	ragflow/internal/ingestion/knowledge_compile [build failed]
```

The commit only widens the two stub signatures to match. Happy to drop it if you would rather fix it separately on `main` — this branch rebases cleanly either way.

**Verification**

```
bash build.sh --test ./internal/ingestion/... ./internal/service/...   # all ok, incl. knowledge_compile
pytest --level=p3 test/testcases/restful_api                           # 88 passed, 1 skipped, 2 deselected
pytest test/unit_test --ignore=test/unit_test/sdk                      # 4804 passed, 46 skipped
ruff check / ruff format --check                                       # clean on all changed files
```

For the CI-facing suites I diffed failure-name lists against an `upstream/main` baseline worktree rather than reading them in isolation, so only deltas count: no difference in either `test/testcases/restful_api` or `test/testcases/test_sdk_api`.

martincivan:
@JinHai-CN fixed and rebased with resolved conficts. 
Please check.

JinHai-CN:
@qinling0210 would you please check if this PR can be merged.

PR #19015, opened 2026-08-30
Fix don't send an empty tools array

martincivan:
### What

`LLMBundle.async_chat` routes into `mdl.async_chat_with_tools` whenever the model is *declared* tool-capable in the tenant's provider config, without checking whether tools were actually bound to the instance. `Base.async_chat_with_tools` then sends `tools=self.tools, tool_choice="auto"` — an empty array — and providers that validate their request body reject it:

```
400 - Value error, `tools` must not be an empty array.
      Either provide at least one tool or omit the field entirely.
```

`async_chat_streamly` and `async_chat_streamly_delta` both check `getattr(self.mdl, "is_tools", False)` as well; `async_chat` lost that conjunct in #16859, so only the non-streaming path broke. That is the path used by the non-streaming LLM helpers (question rewriting, keyword extraction) and by chat over the HTTP API with `stream: false`.

### Changes

- Restore the `mdl.is_tools` conjunct in `async_chat`, so all three async entry points agree on when tools are in play.
- Add `Base._tool_request_kwargs()`, which omits `tools`/`tool_choice` entirely when nothing is bound, and use it at the three call sites in the base tool loop. The LiteLLM path already guards this in `_construct_completion_args`.

### Tests

`test/unit_test/api/db/services/test_llm_bundle_tool_routing.py` and `test/unit_test/rag/llm/test_empty_tools_request.py` — 9 tests, unit tier.

### DIFF
--- a/api/db/services/llm_service.py
+++ b/api/db/services/llm_service.py
@@ -473,7 +473,7 @@ class LLMBundle(LLM4Tenant):
         return queue
 
     async def async_chat(self, system: str, history: list, gen_conf: dict = {}, **kwargs):
-        if self.is_tools and hasattr(self.mdl, "async_chat_with_tools"):
+        if self.is_tools and getattr(self.mdl, "is_tools", False) and hasattr(self.mdl, "async_chat_with_tools"):
             base_fn = self.mdl.async_chat_with_tools
         elif hasattr(self.mdl, "async_chat"):
             base_fn = self.mdl.async_chat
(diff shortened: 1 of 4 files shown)


buua436 (approved):
Reviewed the PR and found no blocking issues. The changes correctly prevent empty `tools` arrays and ensure that `tools` and `tool_choice` are only sent when tools are actually bound. The added regression tests cover the main routing and request-building scenarios. I recommend approving the PR; extracting the duplicated routing condition and adding a debug log could be considered as minor follow-up improvements.

PR #19018, opened 2026-08-30
fix(mcp): send rerank_candidates_count so retrieval paging works past 64 results

### LINKED ISSUE #19017: MCP ragflow_retrieval: paging past 64 results always fails — rerank_candidates_count is never sent
### Self Checks

- [x] I have searched for existing issues([search for existing issues](https://github.com/infiniflow/ragflow/issues)), including closed ones.
- [x] Non-english title submissions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).

89c02c8 (current main, mcp/server/server.py retrieval payload unchanged since v0.26.x)

### RAGFlow image version

v0.27.1 (also affects earlier versions carrying the current MCP retrieval payload)

Not environment specific: reproduces from a clean checkout of `main` by inspecting the MCP payload and the backend validation.

### Actual behavior

Calling the `ragflow_retrieval` MCP tool with any paging window whose `page * page_size` exceeds 64 always fails:

```
Exception(TextContent(text="rerank_candidates_count must be at least page multiplied by page_size (100)"))
```

The error names a parameter the MCP tool does not expose. `RAGFlowConnector.retrieval()` (`mcp/server/server.py`) builds `data_json` without `rerank_candidates_count`, so the backend (`api/apps/restful_apis/chunk_api.py`) applies its default of 64 and then rejects the request when `page * page_size > 64`.

Concretely, with the tool schema as advertised (`page_size` `"maximum": 100`, description "max recommended: 50", default 10):

- `page=1, page_size=65..100` → always rejected
- `page=2, page_size=50` (the recommended value) → always rejected
- default `page_size=10` → rejected from `page=7` onwards

So retrieval pagination is effectively capped at 64 results with no way to page further.

### Expected behavior

The MCP tool should honor its advertised `page_size`/`page` bounds: requests within the documented limits should succeed, which requires sending `rerank_candidates_count` grown to cover the requested window (e.g. `max(64, page * page_size)`).

### Steps to reproduce

```
1. Deploy RAGFlow with the MCP server enabled.
2. Connect any MCP client and call ragflow_retrieval with page=2, page_size=50 (or page=1, page_size=100).
3. The tool returns the backend error `rerank_candidates_count must be at least page multiplied by page_size (100)` — about a parameter the tool never exposes.
4. Compare with POST /api/v1/retrieval sending rerank_candidates_count=100 explicitly: succeeds.
```

Pure-Python reproduction (no server needed): stub the connector's `_post`, call `connector.retrieval(page=2, page_size=50, ...)` and inspect the captured payload — `rerank_candidates_count` is absent.

### Additional information

Suggested one-line fix in `mcp/server/server.py` `retrieval()`:

```python
"rerank_candidates_count": max(64, page * page_size),
```

Default and small windows keep the backend default (64); only windows past 64 grow the candidate pool. Happy to send a PR with a regression test.

xu-jia-ming:

The `ragflow_retrieval` MCP tool's advertised paging (`page_size` maximum 100, recommended 50) deterministically failed for any window past 64 results: the connector never sent `rerank_candidates_count`, the backend defaults it to 64 and rejects requests where it is smaller than `page * page_size`. Clients got an error about a parameter the tool does not expose, e.g.:

```
rerank_candidates_count must be at least page multiplied by page_size (100)
```

Failing cases: `page=2, page_size=50` (the recommended value), `page_size > 64`, and default `page_size=10` from `page=7` on.

Fixes #19017

`RAGFlowConnector.retrieval()` now sends `rerank_candidates_count = max(64, page * page_size)`:

- Default and small windows keep the backend default (64) — no behavior change for existing calls.
- Only windows past 64 results grow the candidate pool to exactly what the page requires.

### How it was tested?

New regression suite `test/unit_test/mcp/test_server_retrieval.py` (same importlib + stubbed-connector pattern as the existing `test_server_datasets.py`), parameterized over `(page, page_size)`:

- `(1,10)`, `(1,50)` → candidate count stays 64
- `(2,50)`, `(7,10)`, `(1,100)` → candidate count equals `page * page_size`
- asserts `rerank_candidates_count >= page * page_size` for all cases

All 5 cases fail at the merge base (`KeyError: 'rerank_candidates_count'`) and pass with this change; the full existing MCP suite (`test/unit_test/mcp/`) passes: 19 total.

```
$ pytest test/unit_test/mcp/ -q --noconftest -p asyncio
19 passed
```

(`--noconftest` skips the repo-wide `test/unit_test/conftest.py`, which pulls in the full rag/nlp stack — unrelated to the MCP module; the module-level load path used by the existing tests is unchanged.)

ruff check / ruff format --check on both changed files: clean.

Tests not run: the Go tiers and the full Python unit suite (unchanged modules; no infra for them on this machine).


6ba3i (changes requested):
The immediate error is fixed, but I don’t think `rerank_candidates_count = max(64, page * page_size)` is safe here. The retrieval layer reranks the full candidate window before slicing the requested page, so increasing that window between page 1 and page 2 can change the ranking and potentially introduce duplicates or skipped results across pages.

Also, the Go MCP path still defaults `rerank_candidates_count` to 64, so the same issue remains there.

I think the cleaner fix is to expose/use a fixed candidate window for the whole pagination sequence, validate that `page * page_size` fits inside it, and keep Python/Go behavior aligned.

6ba3i:
could you please fix the issues raised in my comment ?

xu-jia-ming:
@6ba3i Done — pushed in 14b858b, taking your suggested shape:

- **Fixed candidate window instead of a growing one**: both the Python MCP server and the Go path now send a constant `rerank_candidates_count = 512` on every page of a pagination sequence, so the rerank pool — and therefore the ranking — is identical across pages and can no longer duplicate or skip results between page 1 and page 2. (Value chosen to cover the tool's documented paging range at `page_size = 100`, while staying below the default `top_k = 1024`.)
- **Up-front validation**: `page * page_size` windows that don't fit inside the fixed window are rejected before any request is sent, with a message telling the caller to narrow `page` or `page_size` — rather than surfacing the backend's generic rejection.
- **Python/Go aligned**: the Go `MCPRetrieval` previously left the service default of 64 in place; it now applies the same constant and the same window validation (`validateRetrievalWindow`), with a keep-in-sync comment on both sides naming the counterpart constant.
- Tests: the Python suite now asserts the fixed window on every page (5 paging cases) plus the up-front rejection; Go adds a table-driven `TestValidateRetrievalWindow` (defaults, schema max, exact boundary, past-window rejection). Also applied CodeRabbit's isolation nit — the retrieval stub now patches `_get_document_metadata_cache` so the tests can't perform real network I/O.

Python: 6/6 pass. Go: package verified in a scratch module (the `internal/handler` package has pre-existing Windows-only build breakage in unrelated sandbox/stagehand files, so the exact source was compiled and tested standalone); committed blob is gofmt-clean.

xu-jia-ming:
Both CodeRabbit findings from the last push are addressed in 53de31d:

- **Overflow-safe check**: the window comparison now divides (`page > mcpRerankCandidatesCount/pageSize`) instead of multiplying `page * pageSize`, so a `math.MaxInt` page can no longer overflow the product and slip past validation; the error message prints the raw operands rather than the (potentially wrapped) product.
- **Exact boundary tested**: `8 * 64 == 512` is asserted as allowed and `9 * 64` as rejected, plus a `math.MaxInt, math.MaxInt` regression case. All 7 cases pass.

6ba3i (approved):
LGTM

6ba3i:
@xu-jia-ming please fix format

xu-jia-ming:
Formatted in e652e79 — the preflight's ruff-format wanted the implicit string concatenation in the window-rejection message collapsed into a single pair of parentheses. `ruff format --check` now passes on both touched Python files and the retrieval tests still pass (6/6).

xu-jia-ming:
The `ragflow_tests_elasticsearch (python)` failure looks like an unrelated flake: `test/testcases/restful_api/test_documents.py::test_documents_metadata_batch_update_contract` failed with a 30s `requests.exceptions.ReadTimeout` against `host.docker.internal:10451` — a timeout in the documents metadata batch-update REST test, nowhere near the MCP retrieval path this PR touches. Everything else passed in that job (SDK suite 701 passed; REST suite 459 passed with only this one timing out). Could you re-run the failed job?

PR #19020, opened 2026-08-30
fix(mcp): read total_datasets so the fetch-all early-exit works

### LINKED ISSUE #19019: MCP: _fetch_all_datasets reads the wrong pagination key — early-exit never fires, one extra request per full listing
### Self Checks

- [x] I have searched for existing issues([search for existing issues](https://github.com/infiniflow/ragflow/issues)), including closed ones.
- [x] Non-english title submissions will be closed directly ( 非英文标题的提交将会被直接关闭 ) ([Language Policy](https://github.com/infiniflow/ragflow/issues/5910)).

6129839 (current main)

### RAGFlow image version

v0.27.1

Not environment specific: contract mismatch visible in-source between the MCP connector and the REST envelope.

### Actual behavior

`RAGFlowConnector._fetch_all_datasets()` (`mcp/server/server.py`) ends each page loop with:

```python
total = res_json.get("total")
if total is not None and len(datasets) >= total:
    break
```

But the REST envelope writes the dataset total under **`total_datasets`**, never `total` — `api/utils/api_utils.py get_result()` does `response["total_datasets"] = total`, and `GET /api/v1/datasets` goes through it (`api/apps/restful_apis/dataset_api.py:398`). So `total` is always `None`, the early-exit is dead code, and every full listing terminates only via the empty-page fetch — one wasted backend roundtrip per call.

This path runs on every `tools/list` (the dataset description embedded in tool schemas) and every `ragflow_retrieval` without explicit `dataset_ids` (`resolve_dataset_ids`), so the extra request is paid on the hottest MCP paths.

### Expected behavior

The loop should stop as soon as the fetched rows reach the reported total, without the extra empty-page request.

### Steps to reproduce

Pure-Python (stub the connector's `_get` mirroring the real envelope):

```
1. Stub pages of 100 datasets each, returning {"code": 0, "data": [...], "total_datasets": 200}.
2. Call connector.list_datasets(api_key=...) (page_size=-1 → _fetch_all_datasets).
3. Observed: pages requested are [1, 2, 3] — page 3 is the empty page that ends the loop.
4. Expected: pages [1, 2] — the total is already reached after page 2.
```

### Additional information

One-line fix: read `res_json.get("total_datasets", res_json.get("total"))` (keeping `total` as a fallback for gateway-reshaped envelopes). The existing unit test stub (`test/unit_test/mcp/test_server_datasets.py::_stub_dataset_pages`) invents a `"total"` key the real backend never sends, which masked the bug. Happy to send a PR with the fix, the corrected stub, and a boundary regression test.

xu-jia-ming:

`_fetch_all_datasets()` read the dataset total from the wrong key (`total`), but the REST envelope reports it as `total_datasets` (`api/utils/api_utils.py get_result` → `response["total_datasets"] = total`, reached from `GET /api/v1/datasets` via `dataset_api.py:398`). The early-exit was therefore dead code and every full listing paid one extra empty-page backend request — on the hottest MCP paths: `tools/list` (dataset descriptions embedded in tool schemas) and every `ragflow_retrieval` without explicit `dataset_ids`.

Fixes #19019

- `mcp/server/server.py`: `total = res_json.get("total_datasets", res_json.get("total"))` — the real key first, `total` kept as a fallback for gateways that reshape the envelope.
- `test/unit_test/mcp/test_server_datasets.py`: the shared stub now mirrors the real REST envelope (`total_datasets` instead of the invented `total` key that masked the bug), plus a boundary regression test asserting the loop stops at the total without an empty-page request (200 datasets → pages `[1, 2]`).

### How it was tested?

```
$ pytest test/unit_test/mcp/ -q --noconftest -p asyncio
15 passed
```

Without the source fix, 3 tests fail (the new boundary test plus the two existing fetch-all tests under the corrected real-envelope stub), confirming the regression signal. `--noconftest` skips the repo-wide conftest that pulls the full rag/nlp stack (unrelated to this module; same runner as the existing MCP tests). ruff check/format clean.

Tests not run: Go tiers and the full Python unit suite (unchanged modules).

### DIFF
--- a/api/utils/api_utils.py
+++ b/api/utils/api_utils.py
@@ -304,7 +304,7 @@ def get_result(code=RetCode.SUCCESS, message="", data=None, total=None):
     {
         "code": 0,
         "data": [...],        # List or object, backward compatible
-        "total": 47,          # Optional field for pagination
+        "total_datasets": 47, # Optional field for pagination
         "message": "..."      # Error or status message
     }
     """
--- a/internal/handler/mcp_pagination_test.go
+++ b/internal/handler/mcp_pagination_test.go
@@ -1,0 +1,95 @@
+package handler
+
+import (
+	"fmt"
+	"testing"
+)
+
+func fakeDatasetPager(total int, pageSize int) (func(page, size int) ([]map[string]interface{}, int64, error), *[]int) {
+	pagesRequested := []int{}
+	page := func(p, size int) ([]map[string]interface{}, int64, error) {
+		pagesRequested = append(pagesRequested, p)
+		start := (p - 1) * size
+		end := start + size
+		if end > total {
+			end = total
+		}
+		if start >= total {
+			return nil, int64(total), nil
+		}
+		data := make([]map[string]interface{}, 0, end-start)
+		for i := start; i < end; i++ {
+			data = append(data, map[string]interface{}{"id": fmt.Sprintf("dataset-%d", i)})
+		}
+		return data, int64(total), nil
+	}
+	return page, &pagesRequested
+}
+
+func TestFetchAllDatasetIDsStopsAtTotalWithoutEmptyPage(t *testing.T) {
+	const pageSize = 100
+	tests := []struct {
+		name        string
+		total       int
+		wantPages   []int
+		wantIDCount int
+	}{
+		{"exact multiple of page size", 200, []int{1, 2}, 200},
+		{"non multiple of page size", 250, []int{1, 2, 3}, 250},
+		{"single full page", 100, []int{1}, 100},
+		{"single short page", 42, []int{1}, 42},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			page, pagesRequested := fakeDatasetPager(tt.total, pageSize)
+
+			ids, err := fetchAllDatasetIDs(page, pageSize)
+
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if len(ids) != tt.wantIDCount {
+				t.Fatalf("got %d ids, want %d", len(ids), tt.wantIDCount)
+			}
+			if len(*pagesRequested) != len(tt.wantPages) {
+				t.Fatalf("requested pages %v, want %v", *pagesRequested, tt.wantPages)
+			}
+			for i, p := range tt.wantPages {
+				if (*pagesRequested)[i] != p {
+					t.Fatalf("requested pages %v, want %v", *pagesRequested, tt.wantPages)
+				}
+			}
+		})
+	}
+}
+
+func TestFetchAllDatasetIDsPropagatesErrors(t *testing.T) {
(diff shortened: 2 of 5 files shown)


6ba3i (changes requested):
Please also update the Go MCP path in this PR. `MCPRetrieval` already receives the dataset total from `ListDatasets` but discards it, so exact multiples of 100 still trigger an unnecessary empty-page request. Since this PR fixes the same pagination behavior in Python, it would be better to keep both implementations consistent.

We also need to correct the `get_result()` docstring, which still documents the field as `total` instead of `total_datasets`.

xu-jia-ming:
Done — all three items are in the new head:

**Go path.** The pagination loop is extracted into `fetchAllDatasetIDs` (`internal/handler/mcp_server.go`), which now stops once the total that `ListDatasets` already returns is reached — exact multiples of the page size no longer trigger the extra empty-page request, matching the Python connector's early exit. Table-driven tests in `internal/handler/mcp_pagination_test.go` cover: exact multiple (200 datasets → pages `[1, 2]`), non-multiple (250 → `[1, 2, 3]`), single full/short page, error propagation, and malformed-row skipping. All pass under `go test` (verified in a standalone module since the full `internal/handler` package has pre-existing Windows-only build issues in unrelated deps — `stagehand-go` and the sandbox build tags — that don't affect Linux CI).

**Docstring.** `get_result()` in `api/utils/api_utils.py` now documents the field as `total_datasets`.

**Comment wording.** The Python comment states only the current response contract per the coding guidelines (the code-level fallback key is unchanged).

6ba3i (approved):
LGTM

6ba3i:
@xu-jia-ming please fix ci

xu-jia-ming:
Both CodeRabbit cleanups applied in the new head (comment now only mentions avoiding the extra empty-page request; the unused `total = -1` sentinel is gone and the loop compares against the page's returned total directly) — gofmt/vet clean, table-driven tests re-run green.

On the failing Python CI jobs: the only failure in both is `test/testcases/restful_api/test_chunks.py::test_chunk_delete_web_legacy_basic_variants`, which times out with `TimeoutError: timed out` after ~143s in the elasticsearch/infinity suites. That test is untouched by this PR (our Python-side changes are the MCP pagination key and a docstring), and the same `sep-tests` workflow has failing runs on `main` today (e.g. run on 69fd509c), so it looks like a pre-existing/flaky failure rather than a regression from this change. The push of the cleanup commit has re-triggered the suite; if it fails on the same timeout again, could you confirm whether it's the known-main flake?

xu-jia-ming:
The re-run is fully green on the new head — including both Python suites that hit the chunk-delete timeout earlier — so that failure was indeed transient. The PR is now mergeable with a clean state; thanks for the review!

xu-jia-ming:
Thanks @6ba3i for the thorough review — the Go-side expansion made this a much better fix than my original Python-only patch. 🙏

PR #19023, opened 2026-08-30
fix(agent): avoid shared StringTransform merge arguments

alaningtrump:
### Summary

StringTransform._merge used a dictionary as a default argument and passed it to get_kwargs, which adds missing template inputs to that mapping. Use None and create a fresh dictionary per call so merge invocations cannot share accumulated state.

### DIFF
--- a/agent/component/string_transform.py
+++ b/agent/component/string_transform.py
@@ -82,10 +82,12 @@ class StringTransform(Message, ABC):
             res.append(s)
         self.set_output("result", res)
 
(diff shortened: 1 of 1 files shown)


buua436 (commented):
The change correctly removes the mutable default argument from StringTransform._merge() and creates a fresh dictionary before get_kwargs() mutates it, preventing arguments from leaking between calls or component instances. The implementation is small and focused; adding a regression test for consecutive _merge() calls with omitted arguments would be a useful, but non-blocking, improvement.

PR #19024, opened 2026-08-30
fix(agentic_rag): align answer variable prompt

yzl0ng:
## Summary

- make the action-state initialization prompt emit the `answer_variable` field consumed by the runtime
- remove the unused `answer_slot` prompt contract
- add a regression test that keeps the prompt and runtime field aligned

## Problem

`action_initialize_state.md` instructs the model to return the final slot under `answer_slot`, while `initialize_state` reads only `data.get("answer_variable")`. A compliant model response therefore leaves `ans_var` unset and silently falls back to slot 0, even when a different slot is the final answer in a multi-hop plan.

This is a focused follow-up to the agentic search refactor in #19000.

## Test plan

- `uv run --no-project --python 3.13 --with pytest --with pytest-asyncio --with langchain-core --with langgraph --with json-repair pytest --confcutdir=test/unit_test/rag/advanced_rag test/unit_test/rag/advanced_rag/test_action_session.py -q`
- `uvx ruff check test/unit_test/rag/advanced_rag/test_action_session.py`
- `uvx ruff format --check test/unit_test/rag/advanced_rag/test_action_session.py`

### DIFF
--- a/rag/prompts/action_initialize_state.md
+++ b/rag/prompts/action_initialize_state.md
@@ -11,13 +11,13 @@ Output ONLY a JSON object:
     {"id": 0, "type": "<entity|person|date|duration|count|number|place|web|dataset>", "clues": ["<what identifies this slot from the question>", "..."], "source": "<dataset name or 'web'>"},
     ...
   ],
-  "answer_slot": <id of the slot the FINAL answer goes to>,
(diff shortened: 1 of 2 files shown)


yingfeng:
Fix by #19040

PR #19029, opened 2026-08-30
Fix: keep a reference to the qqbot dispatch task so replies are not dropped

muhzuhaib:
### Summary

`_handle_event` in `api/channels/qqbot/channel.py` starts the inbound message handler with `asyncio.create_task(self._dispatch(incoming))` and does not keep the task it gets back. The [asyncio documentation for `create_task`](https://docs.python.org/3.13/library/asyncio-task.html#asyncio.create_task) is explicit that this is unsafe: "Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done." When it happens there is nothing in the log, because the task stops existing before the `except` inside `Channel._dispatch` can run.

That task is the whole reply path. `_dispatch` calls the handler installed by `api/channels/bootstrap.py`, which persists the user's message, runs `async_chat` and then calls `ch.send(...)`, so it is suspended across a database write and a full model call before the user sees anything.

There is a second consequence that does not depend on the collector at all. `_run_ws_thread` closes the loop as soon as `_run` returns, and because nothing holds the in-flight dispatches there is nothing for the shutdown path to wait on. Stopping a channel, or a credential reconcile that restarts one, abandons every reply that is still being generated.

This is also the only place in `api/channels/` that does it. `dingtalk`, `discord`, `line`, `telegram`, `wecom` and `whatsapp` all `await self._dispatch(incoming)` directly, `feishu` keeps the future from `run_coroutine_threadsafe`, and the two other `create_task` calls in this same file store their handles in `_heartbeat_task` and the close path.

### Reproduction

A five message reproduction built on the same structure the channel uses, a runner on its own event loop in its own thread with the loop closed in a `finally`:

```
current code : 0/5 replies delivered
patched code : 5/5 replies delivered
```

The current code also prints `Task was destroyed but it is pending!` once per dropped message.

### The fix

The pattern the asyncio documentation gives: a set holding a strong reference to each dispatch task, with a done callback so a finished task removes itself. The shutdown path then waits on whatever is still running before closing the loop, bounded at 4 seconds so it stays inside the 5 second `join` that `stop()` already does. `ruff --select RUF006` is clean on the file afterwards.


buua436 (commented):
The change correctly keeps strong references to in-flight dispatch tasks and drains them during shutdown, but there is a potential issue when a task exceeds `DISPATCH_DRAIN_TIMEOUT`: it remains pending in `_dispatch_tasks` while the event loop is closed, and the done callback may never remove it. If the channel is restarted, the stale task can belong to the previous event loop and cause the next drain attempt to fail, potentially dropping new replies. Please cancel and await pending tasks after the timeout, clear the task set before closing the loop, and add a regression test covering timeout followed by restart.

muhzuhaib:
Thanks, you are right that the timeout path was incomplete. Fixed in f87bf18.

I reproduced it against the channel's real structure, a runner on its own loop in its own thread with the loop closed in a `finally` exactly as `_run_ws_thread` does it, feeding three real `C2C_MESSAGE_CREATE` payloads through `_handle_ws_payload`. With a handler slower than the drain timeout, all three tasks were still in `_dispatch_tasks` when the loop closed and each printed `Task was destroyed but it is pending!`. After the change all three end cancelled and the set is empty.

One detail came out differently from the description, so it is worth recording. `asyncio.wait` has no cross loop check, it adds a done callback to each future and awaits its own waiter, so passing tasks that belong to the closed loop does not raise and the drain is not skipped. Restarting the same channel object still delivered every new reply in my run. What the leftovers cost is time: the second shutdown waited out the full drain timeout on tasks that can never complete, 2.02s against a 2s timeout in the harness and 0.21s after the fix, and in production that is the whole 4s sitting under the 5s join in `stop`.

The change cancels and awaits whatever is still pending once the timeout expires, and clears `_dispatch_tasks` before the loop closes.

Both tests are in `test/unit_test/api/channels/qqbot/test_dispatch_drain.py`. On the previous commit the first fails with `assert [] == ['msg-0', 'msg-1', 'msg-2']` because nothing is cancelled, and the second fails with `shutdown waited 5.00s on tasks from the previous loop`. Both pass in 0.43s on the new commit.

I left the bounded concurrency point out of this PR. The unbounded task creation predates the change, since a bare `create_task` per event has the same shape, so putting a queue or a semaphore in front of `_dispatch` is a separate change with its own behaviour to agree on, and this one stays about the lifetime of the tasks.

The `ragflow_preflight` failure was `ruff format` on the lines I had added, also fixed in the same commit.

PR #19031, opened 2026-08-31
fix(utils): round minimum_should_match fraction instead of truncating

### LINKED ISSUE #19027: [Bug]: minimum_should_match fraction is truncated, not rounded, in every full-text doc-store connector
- [x] Please do not modify this template and fill in all the required fields.

a39b6ec2e9a840c8cf9c1c8c021ade6a7613be56

### RAGFlow image version

n/a, reproduced by reading source and running the connectors directly; not deployed through a docker image.

### Actual behavior

Every full-text doc-store connector (`ESConnection`, `OSConnection`, `OBConnection`, `InfinityConnection`, and their `memory/` counterparts) converts a `minimum_should_match` fraction to a percent string with `str(int(fraction * 100)) + "%"`. Because most two-decimal fractions are not exactly representable in IEEE-754 (0.29 is stored as 0.28999999999999998...), `int()` truncates toward zero and silently sends the query engine a stricter-than-requested threshold, always rounding down:

```
>>> str(int(0.29 * 100)) + "%"
'28%'   # requested 29%
>>> str(int(0.57 * 100)) + "%"
'56%'   # requested 57%
>>> str(int(0.58 * 100)) + "%"
'57%'   # requested 58%
```

Checked all 99 two-decimal fractions in (0, 1): several land on this representation-error boundary and are always rounded down, never up. `rag/nlp/query.py`'s `question()` forwards a caller-supplied Chinese-query threshold straight into `MatchTextExpr.extra_options["minimum_should_match"]` (`memory/services/query.py` has the same call for the memory/session search path), so any float value entering there reaches the affected connectors unmodified.

### Expected behavior

The percent string should reflect the nearest whole percent to the requested fraction (`round()`), not always round down.

### Steps to reproduce

1. From a checkout at the commit above, run:
   ```
   python3 -c "
   for f in (0.29, 0.57, 0.58):
       print(f, '->', str(int(f * 100)) + '%')
   "
   ```
2. Compare against the fraction the caller actually asked for; each output is one percentage point stricter than requested.
3. The same `str(int(fraction * 100)) + "%"` expression appears at:
   - `rag/utils/es_conn.py` (`ESConnection.search`)
   - `rag/utils/opensearch_conn.py` (`OSConnection.search`)
   - `rag/utils/ob_conn.py` (`OBConnection.search`)
   - `rag/utils/infinity_conn.py` (`InfinityConnection.search`)
   - `memory/utils/es_conn.py` (`ESConnection.search`)
   - `memory/utils/infinity_conn.py` (`InfinityConnection.search`)

### Additional information

This is a distinct code path from the `common/float_utils.py` `normalize_overlapped_percent` truncation that #17667 proposed rounding for and was closed unmerged (maintainers restricted the chunk-size UI to integers instead). That function backs a different config field (chunk overlap percent) with a `[0, 90]` clamp; the six sites here back full-text search recall and have no such clamp or accepted alternative, so the same objection does not apply.

RahilOp:
Fixes #19027

All six full-text doc-store connectors converted a fractional `minimum_should_match` to a percent string with `int(fraction * 100)`. Because common fractions like `0.29` are not exactly representable in IEEE-754, `int()` truncated toward zero and produced a stricter threshold (e.g. `0.29 -> "28%"`).

Changes:
- Add `format_minimum_should_match_percent()` in `common/float_utils.py` and use `round()` instead of `int()`.
- Replace the six inline conversions in `rag/utils/{es,opensearch,ob,infinity}_conn.py` and `memory/utils/{es,infinity}_conn.py` with the shared helper.
- Add unit tests for the representational-error cases.

Verification:
- `uv run pytest test/unit_test/common/test_float_utils.py` passes.
- `uv run ruff check` and `uv run ruff format --check` pass on the changed files.


common/float_utils.py
```diff
@@ -56,3 +56,26 @@ def normalize_overlapped_percent(overlapped_percent):
...
+    Returns:
+        A percentage string like "29%".
+
+    Examples:
+        >>> format_minimum_should_match_percent(0.29)
+        '29%'
+        >>> format_minimum_should_match_percent(0.57)
+        '57%'
+    """
+    return str(round(fraction * 100)) + "%"
```
qinling0210: Python 3's round() implements round-half-to-even (banker's rounding), not the classic half-up rounding most users expect for percentages. For fractions that land exactly on a half percent — which are exactly representable in IEEE-754 and thus not affected by the precision issue you fixed — the result is counterintuitive:

```
format_minimum_should_match_percent(0.125)   
# 0.125 * 100 == 12.5 exactly '12%'   # round(12.5) -> 12 (12 is even; 13 is odd)
Users would reasonably expect 12.5% -> 13%.
```
Suggested change:

```
return str(int(fraction * 100 + 0.5)) + "%"
```

test/unit_test/common/test_float_utils.py
```diff
@@ -86,3 +86,19 @@ def test_whitespace_strings(self):
         result = get_float("  invalid  ")
         assert math.isinf(result)
         assert result < 0
+
+
+class TestFormatMinimumShouldMatchPercent:
+    def test_fractions_round_to_nearest_percent(self):
+        """Test that floating fractions round to the nearest whole percent."""
+        assert format_minimum_should_match_percent(0.29) == "29%"
```
qinling0210, line 94: Add test for 0.125, 0.135

qinling0210 (changes requested):

RahilOp:
@qinling0210 Thanks for the review. I switched the helper from round() to half-up rounding via int(fraction * 100 + 0.5) and added tests for 0.125 and 0.135. Pushed as b4d9060.

qinling0210:
@RahilOp Could you modify go side? We expect we can modify both python and go sides if the issue exists on both sides in a PR. Thanks.

RahilOp:
@qinling0210 Done — I added the Go side as well.

Changes:
- Added `common.FormatMinimumShouldMatchPercent` in `internal/common/format.go` using half-up rounding (`int(fraction*100 + 0.5)`), mirroring the Python helper.
- Replaced the three `int(minMatch*100)` / `strconv.Itoa(int(value*100))` sites in the Go Elasticsearch, Infinity, and OceanBase engines with the new helper.
- Added unit tests for the helper and for the ES/OceanBase query builders, including the half-percent boundaries.

Go tests: `go test ./internal/common/...` passes. The engine packages require the native `librag_tokenizer_c_api.a` static library, which is not built in this environment, so I could not run those package tests locally; the changes are purely the helper substitution.

RahilOp:
Note on the latest CI run: the two Go jobs failed during `Build ragflow go server` with an `office_oxide native lib version mismatch` (required v0.1.8, found 0.1.9 on the runner). That is an environment/dependency issue, not a compile error in the Go code changes. The Python Elasticsearch and Python Infinity test jobs both passed.

qinling0210:
@RahilOp  Please rebase your code to latest version.

RahilOp:
@qinling0210 All CI checks are now green after the latest rebase/merge. The Go side uses `internal/common.FormatMinimumShouldMatchPercent` with half-up rounding (including a tiny epsilon so values like `0.285` round to `29%` instead of being truncated by float representation). It covers Elasticsearch, Infinity, and OceanBase, and the helper has unit tests plus ES/OceanBase query-builder tests. Could you please take another look when you have a moment? Thanks!

qinling0210 (approved):

PR #19035, opened 2026-08-31
Fix: After deleting and saving in the knowledge base settings, when reopening the parser, it’s still in the state before deletion.

cike8899:
### Summary

Fix: After deleting and saving in the knowledge base settings, when reopening the parser, it’s still in the state before deletion.

### DIFF
--- a/rag/prompts/generator.py
+++ b/rag/prompts/generator.py
@@ -137,19 +137,20 @@ def message_fit_in(msg, max_length=4000):
 
 
 def kb_prompt(kbinfos, max_tokens, hash_id=False):
-    knowledges = [get_value(ck, "content", "content_with_weight") for ck in kbinfos["chunks"]]
+    chunks = kbinfos["chunks"]
+    knowledges = [get_value(ck, "content", "content_with_weight") for ck in chunks]
     kwlg_len = len(knowledges)
     used_token_count = 0
-    chunks_num = 0
-    for i, c in enumerate(knowledges):
+    selected_chunks = []
+    for ck, c in zip(chunks, knowledges):
         if not c:
             continue
-        used_token_count += num_tokens_from_string(c)
-        chunks_num += 1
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks for the fix — the root cause is clear and the change is well-targeted. Refactoring `isExtractorComponent` into `componentTypeMatches` + `isParserComponent` is clean and behavior-preserving, and the new test correctly captures "a deleted setup must not be resurrected." A couple of things to consider:

**1. Empty override still resurrects all defaults (delete-all edge case)**

`CleanComponentParams` drops a component entirely when its cleaned params are empty (`if len(cleaned) > 0 { result[key] = cleaned }`). So if the user deletes *every* parser setup, the incoming config is `"Parser:HipSignsRhyme": {}` (or the key is omitted), the parser cpnID never appears in `cleaned`, `ok` is `false` in `BuildParserConfig`, and execution falls through to `maps.Clone(s.ParamsDefaults)` — resurrecting every setup from the DSL.

That is inconsistent with the new doc comment ("an explicitly provided Parser config replaces the DSL defaults wholesale"). Worth deciding what "deleted everything" should mean and covering it with a test, e.g. `TestBuildParserConfig_ParserOverrideEmptyDoesNotResurrectDefaults`. If delete-all should stick, consider detecting the parser cpnID's presence in the raw config before cleaning (or letting `CleanComponentParams` retain empty parser maps).

**2. Nit — `maps.Clone` + nil safety**

`maps.Clone` returns `nil` for a nil source. `ExtractAllComponentParams` always allocates a non-nil `ParamsDefaults`, so this is fine today, but the overlay loop (`for k, v := range over { base[k] = v }`) would panic on a nil `base`. The previous `make(...)` was nil-safe. Not a current bug — just a slightly weaker invariant than before.

Everything else looks good: non-parser components keep the merge-and-overlay path, and the existing extractor tests should be unaffected.

PR #19043, opened 2026-08-31
fix(web): stop chunk dialog from submitting task_page_size as 0

### LINKED ISSUE #19039: [Bug]: Saving parsing configuration always fails with `parser_config.task_page_size` error `Input should be greater than or equal to 1 - Value: <0>` when the field is unset
### Self Checks

ec9c08d8 (v0.27.0 tag head)

### RAGFlow image version

v0.27.0 (bug also present on v0.26.3, v0.27.1 and current main)

```Markdown
- Deployment: Docker Compose (ragflow-server container, CPU edition image `infiniflow/ragflow:v0.27.0`)
- OS: Debian 12 (Linux 6.1), x86_64
- Browser: any (UI-side bug, not browser-specific)
- Reproduced on v0.26.3 and v0.27.0
```

### Actual behavior

Any attempt to save the chunking method / parsing configuration dialog of a dataset (or a single document) from the UI fails with:

```
102 Field: <parser_config.task_page_size> - Message: <Input should be greater than or equal to 1> - Value: <0>
```

This happens even when the user only changes `chunk_token_num` (or any other field), and the "Task Page Size" input is **not even rendered** on the dialog (it only appears for PDF + MinerU-related layout recognition, see root cause below).

The request body sent by the web UI contains `"task_page_size": 0`, which the backend pydantic schema (`api/utils/validation_utils.py`, `ParserConfig.task_page_size: Annotated[int | None, Field(default=None, ge=1)]`) rejects with the error above. Because `code=102`, the whole update — including the fields the user actually edited — is rejected.

### Expected behavior

Saving parsing configuration should succeed when `task_page_size` is unset/null in the stored parser config (null is the normal state for non-PDF-oriented datasets, and `queue_tasks()` already falls back to the default 12/22 when the value is falsy: `api/db/services/task_service.py`: `page_size = doc["parser_config"].get("task_page_size") or 12`).

The web form must not submit `task_page_size: 0` for a hidden/unset field.

### Steps to reproduce

```Markdown
1. Create a dataset (chunk method = General) and upload any document, e.g. a `.md` or `.docx` file.
2. In the dataset (or file) parsing configuration dialog, change `chunk_token_number` from 512 to any other value and click Save.
   - Note the dialog does NOT show a "Task Page Size" input (it is only rendered for PDF files, and effectively for MinerU).
3. Observe the error toast:
   Field: <parser_config.task_page_size> - Message: <Input should be greater than or equal to 1> - Value: <0>
   Nothing is saved.

The same happens for the per-document "Change parsing method" dialog on non-PDF files.

Can also be reproduced without the UI via the HTTP API:
- PUT /api/v1/datasets/<id> body {"parser_config": {"chunk_token_num": 1024, "task_page_size": 0}}    -> code=102 task_page_size error
- PUT /api/v1/datasets/<id> body {"parser_config": {"chunk_token_num": 1024}}                          -> code=0 success (field omitted)
- PUT /api/v1/datasets/<id> body {"parser_config": {"chunk_token_num": 1024, "task_page_size": null}}  -> code=0 success
```

### Additional information

**Root cause** (verified on v0.27.0, v0.27.1, and current `main`, `web/src/components/chunk-method-dialog/`):

1. The dataset/document GET response includes the key `task_page_size: null` (it is part of the stored `parser_config` JSON).
2. `use-default-parser-values.ts` -> `useFillDefaultValueOnMount()`: `if (key in parserConfig) pre[key] = parserConfig[key]` — because the **key exists** (value `null`), the built-in default `task_page_size: 12` is skipped and the form field is set to `null`.
3. The zod schema declares `task_page_size: z.coerce.number().optional()`. `z.coerce.number()` converts `null` (and `""`) to **`0`**.
4. The "Task Page Size" `FormField` is only rendered when `showPages && parseType === BuiltIn && layoutRecognize` (input `min={1}`), i.e. effectively for PDF+MinerU. In the common case the field is **hidden but still part of the form state**, and `onSubmit` spreads `...parserConfig`, so `task_page_size: 0` is always submitted.
5. Backend `ParserConfig.task_page_size` has `ge=1` and only allows `None` or `>= 1` -> `0` fails validation -> the entire save is rejected.

So any user who has never set `task_page_size` explicitly (i.e. everyone not using MinerU) cannot save the parsing-config dialog via the UI at all. Confirmed on v0.26.3; fixed in neither v0.27.1 (release notes contain nothing related) nor current main.

**Suggested fixes** (either side would unblock; ideally both):

- Frontend: drop the coercion for this field, e.g.
  `task_page_size: z.preprocess(v => (v === null || v === '' || Number.isNaN(Number(v)) ? undefined : Number(v)), z.number().int().min(1).optional())`
  or simply omit `task_page_size` from the submitted payload when the control is not rendered.
- Alternatively in `useFillDefaultValueOnMount`, treat an existing-but-null value as absent:
  `if (key in parserConfig && parserConfig[key] !== null) ...`
- Backend (defensive): accept `0` as "unset" for `task_page_size` (a `mode="before"` validator mapping `0 -> None`), matching the `or 12` fallback semantics already used by `queue_tasks()`.

I can send a small PR if you confirm the preferred direction.

**Workaround for affected users**: update parsing configuration via the HTTP API and omit `task_page_size` from the payload:

```bash
curl -X PUT "http://<host>:81/api/v1/datasets/<dataset_id>" \
  -H "Authorization: Bearer <API_KEY>" -H "Content-Type: application/json" \
  -d '{"parser_config": {"chunk_token_num": 1024}}'
```

Souravrajvi0:
Fixes #19039

## Problem

Saving the document parsing configuration dialog fails for non-PDF files when `parser_config.task_page_size` is stored as `null`. The hidden field is coerced to `0` by `z.coerce.number().optional()`, and the backend rejects `0` because `ParserConfig.task_page_size` only allows `None` or values `>= 1`.

## Fix

- Add `optionalPositiveInt` zod schema so `null` / empty values stay unset instead of becoming `0`
- Treat `null` as absent when filling default parser values on mount
- Omit `task_page_size` from the submit payload when the control is not rendered

## Testing

- `npm test -- --testPathPattern=parser-config-schema.test --coverage=false`


cike8899:
Unable to reproduce the issue you cited; could you provide relevant screenshots to illustrate this? @Souravrajvi0

Souravrajvi0:
@cike8899 Thanks for taking a look. The failure is easy to miss if the dataset **Configuration** page or the **Go** pipeline dialog is used — those forms do not include `task_page_size`. It only happens on the **Python** per-document **Chunking method** dialog, on a **non-PDF** file, when stored `parser_config` has `"task_page_size": null`.

### Repro (v0.27.0, Python UI)

1. Dataset with chunk method General.
2. Upload a `.md` / `.docx` (not a PDF). Stored document `parser_config.task_page_size` is `null`.
3. Files tab → Parse column **General** → **Switch or configure ingestion pipeline**.
4. The dialog has **no** Task Page Size field (that control is PDF + layout recognition only).
5. Click **Save** (even without changing anything).

### Screenshots

**1. Non-PDF file, General parse** (`repro-chunk.md`)

Files page with markdown file and General parse

**2. Chunking method dialog — Task Page Size is not shown**

Chunking method dialog without Task Page Size

**3. Save still posts `task_page_size: 0`; backend rejects it**

API request and code 102 response

Captured `PATCH /api/v1/datasets/<id>/documents/<id>` body includes `"task_page_size": 0`. Response:

```json
{
  "code": 102,
  "message": "Field: <parser_config.task_page_size> - Message: <Input should be greater than or equal to 1> - Value: <0>"
}
```

A brand-new dataset whose `parser_config` **omits** the key may save fine. After the key exists as `null` (dataset update / older documents), `z.coerce.number()` turns `null` into `0` and the hidden field is still submitted.

This PR keeps `null`/empty unset, fills defaults only when the stored value is present and non-null, and omits `task_page_size` from the payload when the control is hidden.

cike8899:
image

image


I am still unable to reproduce the issue you described; the default `task_page_size` is 12. Thank you for your contribution.  @Souravrajvi0

PR #19045, opened 2026-08-31
Fix: pipeline with compilation + transfomer (extrator) does not work

wangq8:
# After fix
image

# Before
image

### DIFF
--- a/rag/flow/extractor/extractor.py
+++ b/rag/flow/extractor/extractor.py
@@ -78,13 +78,23 @@ class Extractor(ProcessBase, LLM):
         self.set_output("output_format", "chunks")
         self.callback(random.randint(1, 5) / 100.0, "Start to generate.")
         inputs = self.get_input_elements()
-        chunks = []
+        output_format = kwargs.get("output_format")
+        if output_format == "chunks":
+            chunks = deepcopy(kwargs.get("chunks") or [])
+        elif output_format == "json":
+            chunks = deepcopy(kwargs.get("json") or [])
+        elif output_format in ["markdown", "text", "html"]:
+            payload = kwargs.get(output_format) or ""
+            chunks = [{"text": payload}] if payload else []
+        else:
+            chunks = []
         chunks_key = ""
         args = {}
         for k, v in inputs.items():
             args[k] = v["value"]
             if isinstance(args[k], list):
-                chunks = deepcopy(args[k])
+                if not chunks:
+                    chunks = deepcopy(args[k])
                 chunks_key = k
 
         if chunks:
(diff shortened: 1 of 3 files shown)


wangq8 (commented):
This is an AI review comment.

## Summary

The PR fixes two related issues in the pipeline (compilation + Extractor transformer) path:

1. **`DataflowService._encode_batch` coroutine bug.** It was declared `async def` but is invoked through `thread_pool_exec`, which runs a *synchronous* callable in a worker thread. Calling the async function just produces an un-awaited coroutine, so `vts, c = await thread_pool_exec(self._encode_batch, batch, embedding_model)` ends up trying to unpack a coroutine → `cannot unpack non-iterable coroutine object`. Changing it to a plain `def` is correct: `LLMBundle.encode` is synchronous and returns `(embeddings, used_tokens)`. ✅

2. **`Extractor._invoke` ignored upstream `kwargs`.** It previously only sourced chunks from `get_input_elements()` (wired inputs), so an Extractor downstream of a Compiler had nothing to iterate. Reading `output_format`/payload from `kwargs` is the right direction.

The direction is correct and the coroutine fix is clean. A few points to address before merge:

### 1. Reuse the existing `ExtractorFromUpstream` schema instead of hand-rolling
`rag/flow/extractor/schema.py` already defines `ExtractorFromUpstream` (`chunks`, `json`→`json_result`, `markdown`→`markdown_result`, `text`→`text_result`, `html`→`html_result`, `extra="forbid"`), but it is currently dead code. The tokenizer shows the intended pattern (`TokenizerFromUpstream.model_validate(kwargs)`). Hand-rolling `kwargs.get(...)` here diverges from that and (see below) misses field fallbacks. Consider `from_upstream = ExtractorFromUpstream.model_validate(kwargs)` and branch on `from_upstream.output_format`.

### 2. `markdown/text/html` `_result` fallbacks are not covered
The compiler's `_normalize_upstream_chunks` reads both `upstream.get(output_format)` and `upstream.get(f"{output_format}_result")`, and the schema aliases both forms. This fix only reads `kwargs.get(output_format)` (e.g. `"markdown"`), so a payload emitted as `markdown_result` / `text_result` / `html_result` would be silently dropped. (Point 1 fixes this too.)

### 3. `chunks_key` is empty in the kwargs-only path
When chunks come from `kwargs` (`output_format == "chunks"`) and no list-valued input is wired, `chunks_key` stays `""`, so `args[chunks_key] = ck.get("text") ...` is skipped and every chunk is formatted with the same `args`. If the extractor prompt depends on a per-chunk text variable, all chunks would receive identical inputs. Please verify this path (or derive `chunks_key` from the upstream payload key rather than only from `inputs`).

### 4. `toc` branch still indexes `ck["text"]` directly
The `field_name == "toc"` branch still does `ck["id"] = xxhash.xxh64((ck["text"] + ...))`, and `_build_TOC` reads `d["text"]`, so a chunk without a `text` key will `KeyError` there even though the rest of the method now tolerates `content_with_weight`. Fine if `toc` only ever runs on compiler output (which normalizes `text`), otherwise apply the same fallback.

### 5. No tests
Neither the `_encode_batch` sync fix nor the extractor `kwargs` path has a test. Since the extractor now has several format branches (`chunks`/`json`/`markdown`/`text`/`html`), a small unit test for at least the `chunks` and `json` paths would guard against regressions.

### Minor
In `_process_chunks`, the new `text` fallback chain includes `summary`/`questions`; `questions` is frequently a list, in which case `ck.get("questions")` is truthy and `str(list)` becomes the `content_with_weight`/id source. Probably harmless, but consider guarding the fallback with `isinstance(..., str)` so a list-valued field never stringifies into chunk content.

This is an AI review comment.

PR #19048, opened 2026-08-31
fix(agent): reject only truly-empty delimiter entries on canvas save

euvre:
### Summary

Saving an agent canvas that contains a TokenChunker node failed on the Go backend with `[TokenChunker] delimiters does not support empty entries` whenever any separator was configured; users had to delete all delimiters to save. Reported with a screenshot showing the toast `update agent … [TokenChunker] delimiters does not support empty entries` on a node whose only delimiter row is `\n`.

**Root cause.** The web delimiter input converts a typed `\n`/`\t` into the raw control character before saving, and the chunker form's default row is `{ value: '\n' }` (a real newline), so a default-configured TokenChunker always submits `delimiters: ["\n"]`. The Go save-path guard `internal/agent/component/dynamic_params.go` (wired into create/update/publish agent in `internal/service/agent.go`) checked `delimiters`/`children_delimiters` with a `strings.TrimSpace`-based blank test, so the legitimate whitespace-only delimiter was classified as an "empty entry" and the whole save was rejected. The Python backend has no such save-time check, and its runtime (`rag/flow/chunker/token_chunker.py`) explicitly keeps whitespace delimiters and defaults to `["\n"]` — this was a Go-only divergence.

**Fix.** `validateDynamicParams` now treats `delimiters` and `children_delimiters` as split tokens: only a zero-length string marks an incomplete row (parity with the Python runtime's `if d` filter). Name-like repeatable fields (`include_domains`, `site_include`, `tools`, `select_keys`, option lists, …) keep the stricter whitespace-blank rule. The error message for genuinely empty rows is unchanged.

**Tests.** Extended `TestValidateDynamicEntries` in `internal/agent/component/dynamic_params_test.go`:
- valid DSL now includes a TokenChunker with `delimiters: ["\n"]` and `children_delimiters: ["\t"]` — this case reproduced the reported failure verbatim against the pre-fix code (`valid DSL rejected: [TokenChunker] delimiters does not support empty entries`);
- new guard cases assert `delimiters: ["ok", ""]` and `children_delimiters: [""]` are still rejected.

**Verification boundary.** Unit-verified with `CGO_ENABLED=0 go test ./internal/agent/component/` (repro fails pre-fix, all 13 subtests + full package suite pass post-fix). In this sandbox `bash build.sh --test` binaries crash at process init in the statically linked CGO chain (`__cxa_guard_release` SIGSEGV before any test executes — also reproduces with `ragflow_server --help`, i.e. unrelated to this change), so the canonical command and a live browser save-through-the-UI walkthrough could not be run here; the fix is one self-contained pure-Go validation change with no new dependencies.

### DIFF
--- a/internal/agent/component/dynamic_params.go
+++ b/internal/agent/component/dynamic_params.go
@@ -38,9 +38,17 @@ func validateDynamicEntries(value any) error {
 func validateDynamicParams(component string, params map[string]any) error {
 	for _, field := range []string{
 		"include_domains", "exclude_domains", "site_include", "site_exclude",
-		"country_include", "language_include", "delimiters", "children_delimiters",
+		"country_include", "language_include",
 	} {
 		if values, ok := params[field].([]any); ok && containsBlank(values) {
+			return fmt.Errorf("[%s] %s does not support empty entries", component, field)
+		}
+	}
+	// Delimiter entries are split tokens, not names: whitespace-only values
+	// ("\n", "\t") are valid delimiters — the chunker form's default row is a
+	// real newline — so only a zero-length string marks an incomplete row.
+	for _, field := range []string{"delimiters", "children_delimiters"} {
+		if values, ok := params[field].([]any); ok && containsEmpty(values) {
 			return fmt.Errorf("[%s] %s does not support empty entries", component, field)
 		}
 	}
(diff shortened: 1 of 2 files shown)


wangq8 (commented):
Thanks — the root cause and parity analysis are correct. I verified against the Python runtime: `rag/flow/chunker/token_chunker.py` `TokenChunkerParam.check()` filters delimiters with `[d for d in self.delimiters if d]` (and the same for `children_delimiters`), which drops only empty strings and keeps whitespace-only values, and the default is `self.delimiters = ["\n"]`. So treating only zero-length strings as "incomplete" in `validateDynamicParams` is exactly the right Go-side parity fix.

The new `containsEmpty` helper only flags `s == ""`, and the name-like repeatable fields stay on the stricter `containsBlank` path — this correctly separates "split token" fields from "name" fields.

One thing I checked: `containsEmpty` only inspects string-typed entries, so a non-string entry passes silently. That's consistent with the existing `containsBlank` behavior and with Python's `isinstance(d, str)` filter, so no action needed.

Tests cover both the valid whitespace case (which would have failed pre-fix) and the still-invalid empty-entry cases. LGTM.

PR #19052, opened 2026-08-31
fix: allow special tokens in token counting

buua436:

Embedding text containing special tokens such as `<|endoftext|>` caused token counting and truncation to fail.

This change disables special-token rejection during local token counting and truncation.

### Type of change

### DIFF
--- a/common/token_utils.py
+++ b/common/token_utils.py
@@ -126,7 +126,7 @@ def usage_from_response(resp) -> dict:
 def num_tokens_from_string(string: str) -> int:
     """Returns the number of tokens in a text string."""
     try:
-        code_list = encoder.encode(string)
+        code_list = encoder.encode(string, disallowed_special=())
         return len(code_list)
     except Exception:
         return 0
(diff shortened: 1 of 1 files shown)


wangq8 (commented):
Thanks for the fix. Verified against `common/token_utils.py`: `encoder` is `tiktoken.get_encoding("cl100k_base")`, whose `Encoding.encode` defaults `disallowed_special="all"`, so text containing special tokens like `<|endoftext|>` currently raises and is swallowed by the `except Exception: return 0` path — undercounting tokens and breaking truncation. Passing `disallowed_special=()` correctly allows those tokens to be encoded as their real token IDs.

The change is minimal and both call sites (`num_tokens_from_string` and `truncate`) are updated consistently. Since cl100k_base special tokens are single token IDs, slicing `[:max_len]` then decoding won't split a special token in half, so no decode edge case there.

Minor (non-blocking): a small unit test covering a string containing a special token would help guard against regression, but I'm happy with the change as-is. LGTM.

PR #19053, opened 2026-08-31
fix(agent): prevent duplicate tools on one Agent node at creation

euvre:
### Summary

An Agent canvas with **two knowledge-retrieval (Retrieval) tools** failed on every user question with:

```
canvas invoke: [NodeRunError] ... Message: consume deferred Agent stream: component: Agent.Invoke: build tools: duplicate agent tool name "search_my_dateset"
```

Reported from the field with screenshots: the agent ("Docs QA Agent") had two retrieval tools attached; the chat panel answered with the raw `NodeRunError`, and the second tool had been auto-named `Retrieval_NaN`.

**Root cause**

The Agent tool picker **appended Retrieval unconditionally** while every other tool toggled (`use-update-tools.ts`), so a user could attach the same tool twice. Every Retrieval instance reports the same constant function name (`search_my_dateset`, `internal/agent/tool/retrieval.go`), and the Go runtime's `buildAgentTools` rejects duplicate tool names — aborting the whole canvas run before any model call. The `generateAgentToolName` helper additionally mis-computed the second tool's display name as `Retrieval_NaN` (`Number(undefined)` slipping through the index guard).

## Fix

Per review feedback, the correct behavior is to **prevent the duplicate at creation time** rather than make the runtime tolerate it:

- `web/.../tool-popover/use-update-tools.ts`: the Retrieval special case is removed; every tool now toggles — selecting an already-added tool removes it, so one Agent node can never hold two copies of the same tool.
- `web/.../tool-popover/tool-command.tsx`: the Retrieval menu row shows the same selection checkbox as every other tool instead of a plus icon.
- `web/src/pages/agent/store.ts`: `generateAgentToolName` loses its only caller and is removed (dead code).
- The Go runtime **keeps rejecting duplicate tool names** (`duplicate agent tool name %q`) as the last-resort guard for canvases constructed outside the UI or saved before this guard — such a canvas should fail loudly, not silently rename.

**Legacy canvases**: a canvas already saved with duplicate tools still fails at run time; opening the tool picker and selecting the duplicated tool once removes every copy, returning the canvas to a valid state.

## Verification

- New hook tests (`use-update-tools.test.tsx`, 4 cases): add-when-absent, toggle-off-when-present, never two copies across repeated selects, and cleanup of a legacy canvas that already carries duplicates.
- `jest src/pages/agent/form/agent-form/tool-popover src/pages/agent/store.test.ts`: 10/10 pass.
- `bash build.sh --test ./internal/agent/component/...`: pass (runtime behavior unchanged from main).
- `oxlint` clean on touched files; `tsc --noEmit` reports no new errors in touched files (remaining errors are pre-existing on main).

### DIFF
--- a/web/src/pages/agent/form/agent-form/tool-popover/tool-command.tsx
+++ b/web/src/pages/agent/form/agent-form/tool-popover/tool-command.tsx
@@ -12,7 +12,6 @@ import { Operator } from '@/pages/agent/constant';
 import OperatorIcon from '@/components/operator-icon';
 import { t } from 'i18next';
 import { lowerFirst } from 'lodash';
-import { LucidePlus } from 'lucide-react';
 import { PropsWithChildren, useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useGetNodeTools, useUpdateAgentNodeTools } from './use-update-tools';
@@ -73,13 +72,7 @@ function ToolCommandItem({
 }: ToolCommandItemProps & PropsWithChildren) {
   return (
     <CommandItem className="cursor-pointer" onSelect={() => toggleOption(id)}>
-      {id === Operator.Retrieval ? (
-        <span>
-          <LucidePlus className="size-4" />
-        </span>
-      ) : (
-        <Checkbox checked={isSelected} />
-      )}
+      <Checkbox checked={isSelected} />
       {children}
     </CommandItem>
   );
--- a/web/src/pages/agent/form/agent-form/tool-popover/use-update-tools.test.tsx
+++ b/web/src/pages/agent/form/agent-form/tool-popover/use-update-tools.test.tsx
@@ -1,0 +1,97 @@
+import { AgentFormContext } from '@/pages/agent/context';
+import { Operator } from '@/pages/agent/constant';
+import useGraphStore from '@/pages/agent/store';
+import { act, renderHook } from '@testing-library/react';
+import { useUpdateAgentNodeTools } from './use-update-tools';
+
+// The initial-values hook pulls in the whole operator-params catalog; the
+// params payload is irrelevant to duplicate-guard behavior.
+jest.mock('@/pages/agent/hooks/use-agent-tool-initial-values', () => ({
+  useAgentToolInitialValues: () => ({
+    initializeAgentToolValues: () => ({}),
+  }),
+}));
+
+type Tool = {
+  component_name: string;
+  name: string;
+  params: object;
+  id?: string;
+};
+
+const agentNode = (tools: Tool[]) =>
+  ({
+    id: 'agent:0',
+    type: 'ragNode',
+    position: { x: 0, y: 0 },
+    data: { label: Operator.Agent, name: 'agent:0', form: { tools } },
+  }) as any;
+
+function seedStore(tools: Tool[]) {
+  useGraphStore.setState({ nodes: [agentNode(tools)], edges: [] });
+}
+
+// Mirror the real tree: the context node always comes from the live store,
+// so a toggle re-reads the tools written by the previous toggle.
+function Wrapper({ children }: any) {
+  const node = useGraphStore((s) => s.nodes.find((n) => n.id === 'agent:0'));
+  return (
+    <AgentFormContext.Provider value={node as any}>
+      {children}
+    </AgentFormContext.Provider>
+  );
+}
+
+const currentTools = () =>
+  (useGraphStore.getState().nodes[0].data.form as any).tools as Tool[];
+
+describe('useUpdateAgentNodeTools (creation-time duplicate guard)', () => {
+  beforeEach(() => seedStore([]));
+
+  const renderTools = () =>
+    renderHook(() => useUpdateAgentNodeTools(), { wrapper: Wrapper });
+
+  it('adds a tool that is not present yet, named after the component', () => {
+    const { result } = renderTools();
+    act(() => result.current.updateNodeTools(Operator.Retrieval));
+    expect(currentTools().map((x) => x.component_name)).toEqual([
+      Operator.Retrieval,
+    ]);
+    expect(currentTools()[0].name).toBe(Operator.Retrieval);
+  });
+
+  it('selecting an already-added tool removes it instead of appending a duplicate', () => {
+    seedStore([
+      { component_name: Operator.Retrieval, name: Operator.Retrieval, params: {} },
+    ]);
+    const { result } = renderTools();
+    act(() => result.current.updateNodeTools(Operator.Retrieval));
+    expect(currentTools()).toEqual([]);
+  });
+
+  it('never holds two copies of the same tool, Retrieval included', () => {
+    const { result } = renderTools();
+    act(() => result.current.updateNodeTools(Operator.Retrieval));
+    // A second select toggles it off — previously this appended a second
+    // Retrieval tool, which later crashed the Go runtime with
+    // "duplicate agent tool name \"search_my_dateset\"".
+    act(() => result.current.updateNodeTools(Operator.Retrieval));
+    act(() => result.current.updateNodeTools(Operator.Retrieval));
+    const copies = currentTools().filter(
+      (x) => x.component_name === Operator.Retrieval,
+    );
+    expect(copies).toHaveLength(1);
+  });
+
+  it('cleans up a legacy canvas that already carries duplicates', () => {
+    seedStore([
+      { component_name: Operator.Retrieval, name: 'Retrieval', params: {} },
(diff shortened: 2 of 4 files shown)


wangq8 (commented):
This is an AI review comment.

Thanks — the root cause is correctly diagnosed (the shared `search_my_dateset` name colliding in `buildAgentTools`, and `Number(undefined)` → `NaN` in the frontend). The fix is minimal and preserves the name-uniqueness invariant that the Python runtime guarantees.

**What looks good**
- Keeping the first occurrence's canonical name and only disambiguating collisions avoids changing single-tool canvases — good minimal-diff approach.
- `renamedTool` overrides only `Info()` and delegates `InvokableRun` to the wrapped tool, so dispatch and execution stay correct.
- The frontend fix (`idx ? Number(idx) : -1`) maps digit-less names to `-1`, producing `Retrieval_0` instead of `Retrieval_NaN`, while preserving existing indexed behavior.

1. `uniqueAgentToolName` numbers the first collision at `_2` (the `for n := 2` loop), so the second retrieval tool becomes `search_my_dateset_2`. That reads oddly next to the frontend's `Retrieval_0` / `Retrieval_1`. Uniqueness is preserved so this is not a bug, but consider starting at `_1` (or aligning the two naming schemes) for clarity. Note: this helper is pre-existing, so it's only newly-exposed by reusing it for tools.
2. The acknowledged limitation — node-level `ToolParams` keyed by tool name, so two Retrieval tools with different node-level configs share the last one's config — is a real correctness gap for configured tools. Recommend filing a follow-up issue so per-instance tool params don't get lost.

No regressions or security concerns found. Tests reproduce the exact reported error string and cover duplicate/triplicate cases.

This is an AI review comment.

euvre:
Hi @wangq8 — thanks for the thorough review and for confirming the root-cause analysis. Addressing the two suggestions:\n\n**1. Numbering start (`_2` → `_1`)** — applied in a6896bbb2. `uniqueAgentToolName` now starts suffixes at `_1`, so the *k*-th occurrence of a base name gets `_{k-1}` — exactly the name Python's Agent assigns that occurrence (`agent_with_tools.py` indexes every tool by its 0-based list position). A second Retrieval tool is now `search_my_dateset_1` on **both** runtimes (previously `search_my_dateset_2` in Go), and it no longer reads oddly next to the frontend's `Retrieval_0` / `Retrieval_1`. Since the helper is shared with sub-agent naming, the pre-existing pin in `TestAgent_CanvasSubAgentToolNamesAreUniqueAfterNormalization` was updated to `agent,agent_1`, and the duplicate/triplicate tests now assert the exact generated names (`search_my_dateset,search_my_dateset_1[,search_my_dateset_2]`). Full alignment with the frontend scheme — suffixing *every* tool from `_0` — is still deliberately not done: it would rename tools on existing single-tool canvases, which this minimal fix intentionally keeps stable.\n\n**2. Follow-up issue for per-instance node-level `ToolParams`** — agreed, it's a real correctness gap for differently-configured duplicate tools, and deliberately out of this PR's scope (recorded under "Known limitation" in the description). The proper fix re-keys `AgentParam.ToolParams` and `agenttool.BuildAll`'s name-based param lookup (`normalizeToolName` / lowercased / raw fallbacks) to per-instance params; the frontend DSL already saves per-entry `params`, so no data migration should be needed. I can't create issues from the environment I'm replying from — could you file the follow-up issue? Alternatively I'll open the follow-up PR directly once this lands.\n\n**Verification** — `bash build.sh --test ./internal/agent/component/...` is green, including the updated naming pins. No frontend changes were needed for either item.

internal/agent/component/agent.go
```diff
@@ -655,7 +655,17 @@ func buildAgentTools(ctx context.Context, p AgentParam) ([]einotool.BaseTool, er
...
-			return nil, fmt.Errorf("duplicate agent tool name %q", info.Name)
+			// Python's Agent indexes every tool name (agent_with_tools.py:
+			// indexed_name = f"{original_name}_{idx}"), so a tool list may
+			// legitimately contain several tools reporting the same function
+			// name — e.g. two Retrieval tools that both report
+			// "search_my_dateset". Keep the first occurrence and disambiguate
+			// later collisions instead of failing the whole canvas run;
+			// eino routes model tool calls by Info().Name.
+			name := uniqueAgentToolName(info.Name, toolNames)
+			tools[i] = &renamedTool{BaseTool: tool, name: name}
```
yuzhichang: We shall disallow an agent has multiple tools with the same function.

yuzhichang (changes requested):

euvre:
Hi @yuzhichang — agreed: an Agent should not be able to hold multiple tools with the same function. The PR has been reworked in that direction and rewritten as a single commit (a04d9c6b6, description updated):

**Applied — duplicates are now disallowed, not tolerated**
- `web/.../tool-popover/use-update-tools.ts`: the Retrieval special case (append unconditionally while every other tool toggled) is removed — every tool now toggles, so selecting an already-added tool removes it and one Agent node can never hold two copies of the same tool.
- `web/.../tool-popover/tool-command.tsx`: the Retrieval menu row shows the same selection checkbox as every other tool (the plus icon is gone).
- `web/src/pages/agent/store.ts`: `generateAgentToolName` — the helper that could also produce `Retrieval_NaN` — loses its only caller and is deleted.
- The Go-side disambiguation from the earlier iteration (`renamedTool` in `buildAgentTools`) is dropped: `internal/agent/component/agent.go` is back to main and keeps rejecting duplicate tool names (`duplicate agent tool name %q`) as the last-resort guard for canvases created outside the UI or saved before this fix — such a canvas fails loudly instead of being silently renamed, and selecting the duplicated tool once in the picker removes every copy, returning the canvas to a valid state.

**Verification**
- New hook tests (`use-update-tools.test.tsx`, 4 cases): add-when-absent, toggle-off-when-present, never-two-copies-across-repeated-selects, and cleanup of a legacy canvas that already carries duplicates; `store.test.ts` unchanged — `jest` on both suites: 10/10 pass.
- `oxlint` clean on the touched files; `tsc --noEmit` introduces no new errors (the remaining errors pre-exist on main, including the ones reported on lines this PR does not touch).
- `bash build.sh --test ./internal/agent/component/...`: green — Go runtime behavior is identical to main, with the duplicate-name guard intact.

@wangq8 — heads-up that the runtime disambiguation we discussed in the previous round (suffix numbering start, per-instance `ToolParams` follow-up) no longer applies now that duplicates are rejected rather than renamed; thanks again for the earlier review.

euvre:
@yuzhichang Agreed — the PR has been reworked in that direction (force-pushed as a04d9c6b). The runtime no longer tolerates duplicates: `buildAgentTools` still rejects a duplicate tool name as the last-resort guard. What changed is the creation path: the tool picker previously appended Retrieval unconditionally (every other tool toggled), so the same tool could be attached twice; Retrieval now toggles like the rest, its menu row shows the selection checkbox instead of a plus icon, and the `generateAgentToolName` helper (which produced `Retrieval_NaN`) is removed as dead code. New hook tests cover add/toggle-off/no-duplicates/legacy-canvas-cleanup; jest 10/10 and `build.sh --test ./internal/agent/component/...` pass. PTAL.

euvre:
**Does the frontend change break the Python runtime? No — verified end-to-end on the PR branch (`API_PROXY_SCHEME=python`).**

Evidence in three layers:

1. **Code**: the Python Agent (`agent/component/agent_with_tools.py:83-85`) builds `indexed_name` from each tool's meta function name + list index, and instantiates components via `component_name`. The display `name` this PR touches never reaches runtime semantics. The DSL shape (`component_name/name/params/id`) is unchanged; only the creation path (toggle semantics) changed.
2. **Frontend behavior**: selecting an already-added Retrieval now toggles it off — repeated selects never produce a second instance. The saved DSL (fetched via `/api/v1/agents/<id>`) contains exactly one entry in `Agent.params.tools`.
3. **Runtime**: canvas executed fully on the Python runtime — the agent answered via the knowledge-base fallback, zero `ERROR` in the backend log, no `duplicate agent tool name` (that error string is Go-only and remains as the intended last-resort guard).

Retrieval now renders as a checked checkbox — selecting it again toggles it off instead of adding a duplicate

Agent chat completed on the Python runtime: no NodeRunError, model answered via the knowledge-base fallback

Two unrelated observations from the verification run (not caused by this PR):

- The template's preset `deepseek-chat` model is absent from this tenant (`invalid param "model_id": required` on first run); switching to an available model made the run pass. The tool-building stage had already succeeded before that error surfaced.
- Toggling a tool off leaves an orphaned Tool node on the canvas (the entry is removed from `form.tools` but the canvas component is not cleaned up) — a pre-existing canvas-sync behavior (the old append path also only called `updateNodeForm`). Runtime consumes `params.tools`, which stays correct; worth a separate issue.

PR #19054, opened 2026-08-31
fix(models): stop breaking local LLM chat calls for qwen/glm models

### LINKED ISSUE #19004: [Bug]: The latest version 0.27.1 cannot be used properly
### Self Checks

my12

### RAGFlow image version

0.27.1

```Markdown
ollama  embedding   shaw/dmeta‑embedding‑zh:latest，lmstudio
qwen3.5‑9b‑uncensored‑hauhaucs‑aggressive
```

### Actual behavior

When using OCR, it is not possible to properly mark keywords on slices, and when Ragflow calls the local large model, the agent cannot answer or index keywords.|

The same settings are available in the reverted version

### Expected behavior

_No response_

### Steps to reproduce

```Markdown
Normal start
```

### Additional information

_No response_

euvre:
### Summary

Local models served by LM Studio / Ollama / vLLM stopped working for qwen/glm-named models: every chat call failed, so chat assistants could not answer, and during document parsing the auto-keyword extraction stage (which calls the same chat model per chunk) aborted — chunks never got their keyword tags and documents failed to index. This is the failure reported in #19004 (0.27.1, lmstudio `qwen3.5-9b-*` chat model + ollama embedding).

Two independent defects in the local-model chat path, plus one startup blocker:

1. **Async-endpoint misrouting.** The `LmStudioModel`, `OllamaModel` and `VllmModel` drivers rewrote the chat URL to `URLSuffix.AsyncChat` for any model whose name prefix is `qwen`/`glm` (e.g. `glm-4-flash`, `qwen-max`, ollama `qwen_7b`). That routing was copied from the Aliyun/DashScope driver, but the provider configs for these three providers (`conf/models/{lmstudio,ollama,vllm}.json`) define **no** `async_chat` suffix, so the rewritten URL collapsed to `<base>/` — a nonexistent endpoint. Every chat request failed. The rewrite is now applied only when a non-empty `async_chat` suffix is actually configured.

2. **Vendor-specific thinking payload sent to generic endpoints.** Since #17592, the generic `OpenAI-API-Compatible` driver unconditionally injected the vLLM-only `chat_template_kwargs: {enable_thinking: ...}` extension for every `qwen3*` model — even when no thinking control was configured. Generic OpenAI-compatible servers implement the OpenAI chat-completions schema (LM Studio's documented parameter list has no such field), and strict servers reject the whole request over unknown arguments, breaking every call for exactly the model family reported in the issue (`qwen3.5-9b-...`). The injection now stays in the vLLM driver (where #17592's fix belongs); the generic driver only sends a thinking control when the user explicitly configures one.

3. **Backend launcher died silently without libjemalloc-dev.** `docker/launch_backend_service.sh` computed `JEMALLOC_PATH` via `pkg-config` under `set -e`; on hosts without libjemalloc-dev the launcher exited before starting any service, with no error message (same class as the just-fixed #18926). jemalloc is now optional with a sane fallback.

**Tests**

- New `TestLocalDriversRouteQwenGlmToChatEndpoint`: qwen/glm-named models on the LM-Studio, Ollama and vLLM drivers must hit the regular chat endpoint (with the production provider configs that define no `async_chat`); a misrouted call is answered with a realistic 404 and fails the test.
- New `TestOpenAICompatibleQwen3SendsNoVendorThinkingPayload`: the generic driver must not send `chat_template_kwargs` / `enable_thinking` / `thinking` / `extra_body` for qwen3-family models (including the reporter's `qwen3.5-9b-uncensored-hauhaucs-aggressive` name) when no thinking control is configured.
- Updated `TestVllmCompatibleQwen3ThinkingPayload` to scope the `chat_template_kwargs` assertions to the vLLM driver; the explicit thinking-toggle behavior for both drivers is still covered by `TestVllmCompatibleNonQwenThinkingPayloadUnchanged`.
- Full `internal/entity/models` package suite passes (`go test ./internal/entity/models/ -count=1`).

**Verification boundary:** the reporter's ollama/LM Studio endpoints were not reachable from the test environment and the full dev stack could not be restarted there (toolchain gaps: no clang++/lld), so this was verified at the driver level with stub OpenAI-compatible / Ollama servers asserting the exact request URL and payload shape, plus the package test suite — not by a live end-to-end run against real local-model servers.

Fixes #19004


wangq8 (commented):
This is an AI review comment.

Thanks for the thorough fix — the root cause is correctly identified and the tests target the exact regression paths.

**What looks good**
- Removing the unconditional qwen/glm → `AsyncChat` rewrite for LM Studio / Ollama / vLLM is correct: none of `conf/models/{lmstudio,ollama,vllm}.json` define an `async_chat` suffix, so the rewrite collapsed the URL to the bare base and broke every chat call. Keeping the guarded path in the vLLM driver is the right call for hosted configs that may define `async_chat`.
- Scoping `chat_template_kwargs` back to the vLLM driver is correct — the generic OpenAI-API-Compatible driver must not emit vLLM-only extensions.
- The jemalloc fallback is a sane, non-fatal degradation and fixes the silent `set -e` exit.

1. `applyExplicitThinkingToggle` duplicates the non-qwen3 branch of `applyVllmCompatibleThinking` verbatim. Consider extracting a shared helper so the two can't drift.
2. The async-routing removal is now asymmetric: LM Studio / Ollama dropped the rewrite entirely while vLLM retains a guard. Fine given current configs, but a one-line comment (or reusing the same guard pattern) on the LM Studio / Ollama drivers would make the intent clearer.
3. The jemalloc fallback hardcodes `/usr/lib/x86_64-linux-gnu`, which is Debian/Ubuntu-specific. On other distros/architectures this yields a nonexistent `LD_PRELOAD` path (harmless loader warning). Optionally gate `LD_PRELOAD` on the file existing.
4. Minor: users who reached a vLLM endpoint via the generic OpenAI-API-Compatible provider will no longer get `chat_template_kwargs` for qwen3 models (they should use the vLLM driver instead). Worth a release-note mention.

No regressions or security concerns found. Tests are well targeted.

This is an AI review comment.

wangq8 (commented):
This is an AI review comment.

The root-cause fix is right: the LM-Studio/Ollama drivers were rewriting qwen/glm chat URLs to an `async_chat` suffix that the bundled configs don't define, producing a bare-base-URL request that 404'd (#19004). Removing that rewrite (and guarding the VLLM one behind a non-empty `async_chat`) restores the normal chat endpoint, and the path-routing tests are exactly what's needed. `strings` stays imported in all three files, so no dead-import break.

Two things to double-check before merge:

- **`applyExplicitThinkingToggle` sends a non-standard `thinking` field to generic OpenAI-compatible endpoints.** The payload `{"thinking":{"type":"enabled|disabled"}}` is not part of the OpenAI chat-completions schema (it matches the Anthropic/302ai/avian convention used elsewhere in this repo). It's only emitted when `config.Thinking` is explicitly set, so it's opt-in, but a strict generic server could reject it for the same class of reason #19004 fixes. Confirm the intended semantics and that the field name is what those local endpoints expect (e.g. Qwen uses `enable_thinking`, Ollama uses `think`).

- **Loss of qwen3 thinking through the generic driver.** Replacing `applyVllmCompatibleThinking` with the explicit-toggle version means qwen3 models reached via "OpenAI-API-Compatible" no longer get `chat_template_kwargs`/`enable_thinking`, even when the endpoint was a vLLM server that supported it. That's the stated intent (generic endpoints shouldn't receive vendor payloads), but users relying on that behavior will silently lose thinking; consider a release-note callout and guidance to use the VLLM driver for vLLM backends.

No regressions in the URL routing itself; tests are solid.

This is an AI review comment.

